PR #849, opened 2018-07-06
Fio p

atheurer:
This adds fio parameters to result.json, so they can be imported into ES.  This will allow the pbench dashboard to better organize results for comparisons.


agent/bench-scripts/postprocess/fio-postprocess
```diff
@@ -147,6 +138,37 @@ if ( @cpu_busy ) {
...
+if ($fio_results{"client_stats"}) {
+	for my $client (@{ $fio_results{"client_stats"} }) {
+		for my $param (keys $$client{"job options"} ) {
+			if (not $param =~/^write_\w+_log$/) {
+				$benchmark_params{$param} = $$client{"job options"}{$param};
+			}
+		}
+	}
+} else {
+	for my $client (@{ $fio_results{"client_stats"} }) {
```
portante: What is different between this block and the one above?

atheurer: The difference is where the stats are stored.  One is for when using --client with fio, and one is for running a local fio job.  The "client_stats" being the difference.

FWIW, I plan on removing the use of --clients and running fio via ssh like we do with uperf, so this code will get smaller.  I actually would prefer I do both of these changes with the same pull request, so I am thinking of just cancelling this PR.

atheurer: Actually, I do have a mistake there, extra reference to client_stats in the second section.  I will fix...

atheurer:
A new commit has been added to hopefully complete this work.

agent/bench-scripts/postprocess/fio-postprocess
```diff
@@ -147,6 +138,53 @@ if ( @cpu_busy ) {
 	$resource{'cpu_busy'} = \@cpu_busy;
 }
 
+# get the fio benchmark parameters from the fio-result.txt (which is mostly json data)
+my %benchmark_params;
+my $fio_results_scalar = get_json($dir . "/fio-result.txt");
+my %fio_results = %$fio_results_scalar;
+my $job_num = 1;
+my $num_jobs = 1;
+for my $param (keys $fio_results{"global options"} ) {
```
portante: Does "`global options`" with a space in the key name mean that there are more keys in the JSON document which have spaces and other non-alphanumeric characters?

atheurer: No, there will be no option names with spaces.  I am confident of this because I have been working on adding -all- fio options to pbench with a new universal benchmark-option parser that validates against a json option-spec file for each benchmark.  For example, I have the fio json spec file populated with every single option:

<pre>
    "size_KBMG_range" :
        {
        "description" : "a range (x-y) of bytes in units of k/K (1024), m/M (1024^2) or g/G (1024^3): 4k-8k 16k-2m 24m-1G",
        "parameters" : [ "bsrange", "bssplit" ],
        "regex" : "[0-9]+[kbmgKBMG]\\-[0-9]+[kbmgKBMG]",
        "value_transforms" : [ "s/([0-9]+)[gG]/($1*1024).\"M\"/eg", "s/([0-9]+)[mM]/($1*1024).\"K\"/eg" ]
        },
    "generic_string" :
        {
        "description" : "all types of strings",
        "parameters" : [ "buffer_pattern", "cgroup", "clocksource", "continue_on_error", "cpus_allowed_policy",
                         "cpus_allowed", "directory", "exec_postrun", "exec_prerun", "ignore_error", "ioscheduler" ],
        "regex" : ".+"
        },
    "ioengine_types" :
        {
        "description" : "all possible ioengine types",
        "parameters" : [ "ioengine" ],
        "regex" : "^(|p|pv|v)sync$|^pvsync2$|^posixaio$|^mmap$|^^(|net)splice$|^sg$|^null$|^net$|^cpuio$|^rdma$|^e4defrag$|^falloc$|^filecreate$|^external$"
        },
    "io_submit_modes" :
        {
        "description" : "How IO submissions and completions are done",
        "parameters" : [ "io_submit_mode" ],
        "regex" : "^inline$|^offload$"
        },
    "time_smh" :
        {
        "description" : "time in seconds, minutes, or hours: 10s 2m 1h",
        "parameters" : [ "runtime", "steadystate_duration", "steadystate_ramp_time", "steadystate" ],
        "regex" : "^[0-9]+[smh]$"
        },
    "boolean" :
        {
        "description" : "1 for true and 0 for false",
        "parameters" : [ "allow_file_create", "allow_mounted_write", "allrandrepeat", "atomic", "block_error_percentiles",
                         "buffered", "cgroup_nodelete", "clat_percentiles", "create_fsync", "create_only", "create_on_open",
                         "create_serialize", "direct", "disable_bw_measurement", "disable_clat", "disable_lat", "disable_slat",
                         "disk_util", "do_verify", "end_fsync", "error_dump", "experimental_verify", "file_append",
                         "fill_device", "fsync_on_close", "gtod_reduce", "invalidate", "lat_percentiles", "log_max_value",
                         "log_offset", "log_store_compressed", "log_unix_epoch", "overwrite", "per_job_logs", "pre_read",
                         "randrepeat", "rate_ignore_thinktime", "replay_no_stall", "scramble_buffers", "serialize_overlap",
                         "stats", "sync", "trim_verify_zero", "unified_rw_reporting", "unique_filename", "unlink_each_loop",
                         "unlink", "verify_dump", "erify_fatal", "verify_state_load", "verify_state_save" ],
        "regex" : "[0,1]"
        },
    "integer" :
        {
        "description" : "a whole number",
        "parameters" : [ "ba", "buffer_compress_chunk", "buffer_compress_percentage", "bwavgtime", "cgroup_weight", "cpumask",
                         "dedupe_percentage", "fdatasync", "flow_id", "flow_sleep", "flow", "flow_watermark", "fsync", "gia",
                         "gtod_cpu", "hugepage-size", "iodepth_batch_complete_max", "iodepth_batch_complete_min", "iodepth_batch",
                         "iodepth_low", "iodepth", "iopsavgtime", "kb_base", "log_avg_msec", "log_compression", "log_hist_coarseness",
                         "log_hist_msec", "loops", "nice", "nrfiles", "numjobs", "offset_align", "openfiles", "percentage_random",
                         "prioclass", "prio", "rate_cycle", "rate_iops_min", "rate_iops", "rate_min", "rate", "replay_align",
                         "replay_scale", "replay_time_scale", "rwmixread", "rwmixwrite", "significant_figures", "thinktime_blocks",
                         "thinktime_spin", "thinktime", "trim_backlog_batch", "trim_percentage", "uid", "unit_base", "verify_async",
                         "verify_backlog_batch", "verify_interval", "verify_offset", "write_barrier" ],
        "regex" : "[0,1]"
        }
}

</pre>

I am getting somewhat off-topic, but it was convenient to explain it with this data.  The above classifies options into groups based on what kind of value they would be assigned, provides a regex to validate the value, and provide a regex to substitute the value if necessary (let's say you want to always convert 1M to 1024k).  I did not come across a single option that has a space.

In the future, if we did have an option with a space, I plan on adding a "options-transforms" which would use a substitution regex to convert it to "_" or something else.


FWIW, The reference to "global options" is only from fio's native json file, which we do not import in whole.  We only dig down to the global and jobs options and pull those out.

So, I don't think we have anything to worry about w.r.t. spaces.

agent/bench-scripts/postprocess/fio-postprocess
```diff
@@ -155,14 +157,28 @@ if ($fio_results{"client_stats"}) {
...
-		for my $param (keys $$client{"job options"} ) {
+} elsif ($fio_results{"jobs"}) {
+	for my $job (@{ $fio_results{"jobs"} }) {
+		for my $param (keys $$job{"job options"} ) {
 			if (not $param =~/^write_\w+_log$/) {
-				$benchmark_params{$param} = $$client{"job options"}{$param};
+				if ($param =~/^numjobs$/) {
+					$num_jobs = $benchmark_params{$param};
+				} elsif ($param =~/^filename$/) {
+					if ($job_num % 4 == 0) {
```
k-rister: What is the significance of 4 that you are using it to % here?

atheurer: thanks for catching this.  It was supposed to be $numjobs

ndokos:
I would like to delay this until v0.53 (end of August): that will allow us to get the unit tests done, get some more review and testing in and coordinate with @gurbirkalsi 's changes to the dashboard. The main impetus is to get v0.52 with the new indexer out *now* and get it into production. @atheurer, @k-rister: is that OK with you?

k-rister:
Sure.  I'm having issues with fio results in the dashboard that I need to address with @gurbirkalsi anyway.

portante:
Closing in favor of #896, which is a rebase and squash of the commits, with two commits to fix unit test environment issues to enable these changes to test cleanly.

PR #851, opened 2018-07-17
Add a proper log environment for verify-backup-tarballs

portante:
The base problem with `pbench-verify-backup-tarballs` is actually with `pbench-report-status` which fails to work because the directory path to the log file is not created when a script does not invoke, `log_init`.  The first patch addresses this problem.

The second patch adds to the unit tests the contents of the report-status payload itself.

The third patch adds proper logging to the `pbench-verify-backup-tarballs` command so that we don't get a cron job email with the output of stderr and the report status email.


portante:
Thanks @ndokos, I'll wait for @chaitanyaenr's review before merging.

PR #853, opened 2018-07-23
Parallelize archive directory checks; compare file lists

portante:
In an attempt to speed up the verification process we have parallelized
the checks of the archive (primary) and the backup directories.

In addition, instead of comparing the output of the md5sum commands, we
compare the content of the combined md5sum files (which are kept in
sorted order by making sure the list of controllers is sorted before
processing, and leveraging "ls"'s sorted output.

### DIFF
--- a/server/pbench/bin/gold/test-9.2.txt
+++ b/server/pbench/bin/gold/test-9.2.txt
@@ -32,5 +32,5 @@
 +++ test-report-status.log file contents
 /var/tmp/pbench-test-server/test-report-status.log:pbench-verify-backup-tarballs.run-1900-01-01T00:00:00-UTC(unit-test)
 /var/tmp/pbench-test-server/test-report-status.log:* Files that exist only in primary directory - extra files in this list are probably OK: they just have not been backed up yet.
-/var/tmp/pbench-test-server/test-report-status.log:/var/tmp/pbench-test-server/pbench/archive/fs-version-001/controller/pbench-user-benchmark_38_2016-05-18_19:36:32.tar.xz: OK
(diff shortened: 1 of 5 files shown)


portante:
@ndokos, @chaitanyaenr, I was not able to run this on a large data set, as the `pbench-report-status` command failed due to:

```
/opt/pbench-server/bin/pbench-report-status: line 103: /bin/sed: Argument list too long
```

So instead of using `sed`, this PR now has a second commit which provides a JSON payload generator to avoid that problem.

Additionally, it took 3 hours to run on our current data set, which seems to be much better since now the archive and backup are checked at the same time.

Also, I have added the ability to test they rsync behavior so we can verify rsync is doing what we want.

chaitanyaenr:
LGTM except that the unit tests are failing.

tenstormavi:
Agreed with Ravi. Apart from unit tests, it seems to be fine.

portante:
@tenstormavi, @chaitanyaenr, @ndokos: well, this now works.

While I was trying to figure out why the rsync was getting different values, I applied a number of fixes for the test environment to see if there were left over artifacts from various tests impacting the rsync.  It was not the case, but the cleanups found some major inconsistencies which have been fixed.

chaitanyaenr:
@portante it's time to remove the "Do not merge" label I guess?

portante:
Once we merge #855 and #856, I'll rebase this work on top of the new master, splitting the rsync work out to its own PR, leaving this PR to contain only verify-backup-tarballs work.

portante:
Still a WIP: this is a rebase on top of #855, and preparation for a rebase on #856 when it lands.

Then we'll move the "Enable actual use of rsync" commit to its own PR so that this PR can just be about parallelizing the archive directory checks.

portante:
Created #857 to track the enabling of rsync commit for review.  This PR is now just about the behaviors of `pbench-verify-backup-tarballs`.

PR #855, opened 2018-07-26
Replace use of sed with a JSON payload generator

portante:
Ran into:

```
    /opt/pbench-server/bin/pbench-report-status: line 103:        /bin/sed: Argument list too long
```

We use a simple python script to properly generate the JSON payload, escaping the file contents so that we always have valid data.


portante:
Taken from #853, since this is really a bug fix unto itself.

PR #858, opened 2018-07-30
Trafficgen updates

k-rister:


k-rister:
Hmm, not sure what is going on with travis-ci but something seems off.

k-rister:
Thanks @ndokos. I looked at the output and I guess I just scrolled past that without seeing it (multiple times)...

agent/bench-scripts/pbench-trafficgen
```diff
@@ -43,7 +43,7 @@ traffic_generator="trex-txrx" # can be either trex-txrx or moongen-txrx
 trex_use_ht="n"
 trex_use_l2="n"
 benchmark_run_dir=""
-traffic_directions="bidirec"
+traffic_directions="bidirectional"
```
portante: Do we need to be binary compatible, and support `bidirec` to mean `bidirectional`?

k-rister: No.  There are a lot more changes coming like this,  we are going to be changing the entire option structure and don't have any need for backwards compatibility.

portante: Shall we make these changes in 0.52 instead of 0.51 so that you can accumulate a number of them into one release?

k-rister: The option structure changes are not pending real soon, that is more a forward looking viewpoint.  The changes and additions in this PR are needed to deploy new functionality to users sooner rather than later.

portante:
Let's wait for @atheurer to review before we merge.

Do we have any unit tests for `trafficgen`?

k-rister:
This PR was rebased on top of #867 since that is a dependency.

@atheurer  - I've added a first pass at postprocessing the TRex profiler data that is produced by the new trex-txrx-profile traffic generator.  *I think* I've managed to follow your data model but would appreciate a review.

Issue #866, opened 2018-08-02
gen_data fails when timeseries data is not present


k-rister:
When gen_data is called on a large chunk of data, if any piece of that data is missing timeseries data then the code bombs out instead of gracefully handling (ie. skipping) that data.

k-rister:
A fix for this was merged in #858

PR #867, opened 2018-08-02
BenchPostprocess.pm:gen_data(): skip data that is missing timeseries …

k-rister:
…samples

### DIFF
--- a/agent/bench-scripts/postprocess/BenchPostprocess.pm
+++ b/agent/bench-scripts/postprocess/BenchPostprocess.pm
@@ -549,20 +549,22 @@ sub create_graph_hash {
 	my $html_name = $$workload_ref{'parameters'}{'benchmark'}[0]{get_label('benchmark_name_label')};
 	foreach my $metric_type ('throughput', 'latency', 'resource', 'efficiency') {
 		if ($$workload_ref{$metric_type}) {
-		foreach my $metric_name (keys %{ $$workload_ref{$metric_type} }) {
-			for (my $i = 0; $i < scalar @{ $$workload_ref{$metric_type}{$metric_name} }; $i++) {
-				my $series_name = get_uid($$workload_ref{$metric_type}{$metric_name}[$i]{get_label('uid_label')}, \%{ $$workload_ref{$metric_type}{$metric_name}[$i] }); 
-				for (my $j = 0; $j < scalar @{ $$workload_ref{$metric_type}{$metric_name}[$i]{get_label('timeseries_label')} }; $j++ ) {
-					my $timestamp_ms = $$workload_ref{$metric_type}{$metric_name}[$i]{get_label('timeseries_label')}[$j]{get_label('date_label')};
-					if ($timestamp_ms) {
-						my $value = $$workload_ref{$metric_type}{$metric_name}[$i]{get_label('timeseries_label')}[$j]{get_label('value_label')};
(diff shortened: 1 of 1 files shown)


k-rister:
Address #866

portante:
+1 to adding a unit test that highlights the problem and verifies the fix, thanks!

k-rister:
There is no existing code that causes this error.  The was found while working on #858 where we are combining timeseries (new) and non-timeseries (existing) data.  The existing data did not require a call to gen_data() so this was never exposed before.

ndokos:
@k-rister: we are planning to cut the next release in the next couple of days. Will you be able to provide a unit test for this PR before then?

k-rister:
Can we defer the unit test?  There is further work to be done here in the near future -- the profiler that this postprocesses is going to have more stuff added to it and then the postprocessor will have to be taught how to consume that.  Until all work that is completed, the unit test would just be churning with each update and be more of an annoyance than a useful tool.

portante:
Unit tests help us detect breakages due to other changes as well.  Let's work to get a unit test.

k-rister:
Looks like I'm confusing my PRs.  I thought this was the other one (#858) when I did my last response.

I think if we are going to write unit tests for the BenchPostprocess.pm stuff that needs to a more thought out exercise by someone who understands it fully (@atheurer?).  I'm willing to help, just not sure where to get started.

portante:
So we have some bench script tests, we just need a dataset that will cause that new code path to be executed.

Is that easy get such a set?

portante:
@k-rister, in the bench-scripts directory, there is a "[`test-04`](https://github.com/distributed-system-analysis/pbench/blob/master/agent/bench-scripts/unittests#L171)" unit tests which causes `create_graph_hash` to be invoked.  It might be that you could copy that test and adjust the data inputs to run it through that code path.

k-rister:
So I'm running into a bit of a chicken and egg problem and need some advice.  The code in PR #858 is the only place that I know of that exercises the code in question.  Accordingly, I have rebased PR #858 on this PR.  So when working on a unit-test for this PR (#867) the code I need access to doesn't exist yet.

I see 3 options:

1. PR #858 is rebased on existing master and is merged with a broken postprocessor which will be fixed once this PR is completed.
2. This PR (#867) is merged without a unit-test and then a unit-test is added to PR #858.
3. The PRs are somehow merged together.  I'm not real sure how I would go about doing this in practice, it can obviously be done though.

Or some other alternative that I haven't thought of yet.

I think I would lean toward option 2.

@ndokos , @portante -- thoughts?

portante:
@k-rister, seems like option #3 is the best.

Perhaps could:

 * `git remote add upstream git@github.com:distributed-system-analysis/pbench.git`
 * `git checkout <PR #867 branch>`
 * `git branch --set-upstream-to=upstream/master <PR #867 branch>`
 * `git rebase`
 * `git format-patch -1`
 * `git checkout <PR #858 branch>`
 * `git branch --set-upstream-to=upstream/master <PR #858 branch>`
 * `git rebase`
 * `git am 0001-<generated above>.patch`
 * `git rebase -i` (then use `squash` command to merge commits)

The above commands will take you through the process of extracting one commit and merging it with the other one.

k-rister:
It looks like my rebase of PR #858 means it already includes this fix, so I think I can just close this PR and continue on with that one by adding the unit test.

Issue #875, opened 2018-08-13
Bugs in v0.51 server:  pbench-server-prep-shim-001 and pbench-sync-package-tarballs


ndokos:
Deploying v0.51 servers uncovered two bugs:

- pbench-server-prep-shim-001: the script processes tarballs that are sent by a version-001 agent. The latter creates a TODO/tarball.tar.xz link to signal that the tarball is ready to be processed. The shim should remove that link after processing, but it does not. Consequently, it encounters that link on every future iteration and complains about it.

- pbench-sync-package-tarballs: this script is called (remotely on a satellite server) by pbench-sync-satellite. The script set its config file like this:

     export CONFIG=${CONFIG:-/opt/pbench-server/lib/config/pbench-server.cfg}

If CONFIG is not set when the script is invoked, this will do the right thing. Unfortunately, if an agent has ever been installed, the /etc/profile.d that it installed is sourced and it sets CONFIG to point to the agent config file, which breaks things. The above setting needs to be unconditional.

PR #876, opened 2018-08-13
Fixes for two bugs discovered when deploying the server

ndokos:
Fixes issue #875.

Deploying v0.51 servers uncovered two bugs:

- pbench-server-prep-shim-001: the script processes tarballs that are
  sent by a version-001 agent. The latter creates a
  TODO/tarball.tar.xz link to signal that the tarball is ready to be
  processed. The shim should remove that link after processing, but it
  does not. Consequently, it encounters that link on every future
  iteration and complains about it.

  The fix is to remove the link. The only problem that might arise is
  if the link removal fails: in that case, the script complains but
  since it cannot remove it, it will continue encountering it and
  complaining about it.

- pbench-sync-package-tarballs: this script is called (remotely on a
  satellite server) by pbench-sync-satellite. The script set its
  config file like this:

    export CONFIG=${CONFIG:-/opt/pbench-server/lib/config/pbench-server.cfg}

  If CONFIG is not set when the script is invoked, this will do the
  right thing. Unfortunately, if an agent has been installed, the
  /etc/profile.d that it installed is sourced and it sets CONFIG to
  point to the agent config file, which breaks things.

  The fix is to set it unconditionally in pbench-sync-package-tarballs:

    export CONFIG=/opt/pbench-server/lib/config/pbench-server.cfg

### DIFF
--- a/server/pbench/bin/pbench-server-prep-shim-001
+++ b/server/pbench/bin/pbench-server-prep-shim-001
@@ -229,6 +229,19 @@ while read tblink ;do
         continue
     fi
 
+    # Remove the link from the reception area.  rm -f returns the same
+    # exit status as rm, except in the case when $tblink does not
+    # exist.
+    rm -f $tblink
+    sts=$?
+    if [ $sts -ne 0 ] ;then
+        echo "$TS: Error: \"rm -f $tblink\", status $sts" |
+            tee -a $status >&4
+        nerrs=$nerrs+1
+        # this will continue to give errors every time through
+        # so we need a better solution
+    fi
+
     pushd $dest/TODO > /dev/null
     ln -s $dest/$(basename $tb) .
     sts=$?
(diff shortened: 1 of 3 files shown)


server/pbench/bin/pbench-server-prep-shim-001
```diff
@@ -229,6 +229,17 @@ while read tblink ;do
         continue
     fi
 
+    # remove the link from the reception area
+    rm -f $tblink
```
portante: Does `rm -f` return an error status when it fails?

server/pbench/bin/pbench-sync-package-tarballs
```diff
@@ -5,7 +5,7 @@
 # WARNING; Do not put anything in here that will pollute the
 # standard output!
 
-export CONFIG=${CONFIG:-/opt/pbench-server/lib/config/pbench-server.cfg}
+export CONFIG=/opt/pbench-server/lib/config/pbench-server.cfg
```
portante: Can you add a comment that says, "we ensure we are using the server config file" or something to that effect?

portante:
@tenstormavi, @chaitanyaenr, can you review these changes?

Issue #884, opened 2018-08-16
Refactor `es_index()` method of `index-pbench` into its own module


portante:
In preparation for moving the `es_index()` method to its own project, [`py-es-bulk`](https://github.com/distributed-system-analysis/py-es-bulk), we need to refactor that method into its own module first.

portante:
Fixed by #885.

PR #836, opened 2018-05-31
Fix indexing performance and enhance error handling

portante:
The performance problem with indexing is two-fold:

 * we accumulate all generated records into one big list before we start indexing, which is really bad
  * we have observed 10s of GBs of memory used when indexing large amounts of data
  * to fix, we just need to use the streaming bulk methods and improve the error handling
 * in order to ensure error handling for Elasticsearch is good, we need source IDs for all documents ahead of time to make it easier to handle retrying operations when an ES instance is overloaded

Along the way, we have attempted to improve the error handling so that we index as much as possible, report errors better, and report them succinctly.

PR #885, opened 2018-08-17
Refactor `es_index` into its own module

portante:
This layers on top of #836 (which must be merged first), refactoring the indexing code into its own module.

Addresses #884.


portante:
Now that #836 has merged, we can consider merging this PR.

portante:
Do you think it is worth pulling this into the v0.52 release?

ndokos:
Sure - why not? It is a a relatively minor change as far as the working of the code is concerned. And it gets us one step closer to having it completely separated out into its own repo.

PR #890, opened 2018-08-22
Fix indexing work to improve error handling

portante:
We fix the default "read timeout" to be very large so that we don't
timeout prematurely and retry needlessly for template operations.

We have enhanced the error handling so that we don't always assume we
get a response object, and we dump out error messages that are
unexpected for now.

We also add the new es_put_template wrapper which properly handles the
retries.

PR #891, opened 2018-08-23
Use dictionary constructor to avoid sorting issues

portante:


portante:
Needed by #890 so that we don't get sorting errors affecting the update unittests in that PR.

portante:
@ndokos, yes, I'll see about fixing that.

portante:
@ndokos, okay, that should address the remaining uses of dictionaries that are convertible without causing more issues.

server/pbench/bin/index-pbench
```diff
@@ -248,14 +248,14 @@ def convert_to_float(d):
...
-        self.run_metadata = {
-            "runtstamp": ptb.start_run,
-            "runid": ptb.md5sum,
-            "experiment": experiment,
-        }
+        self.run_metadata = _dict_const([
+            ("runtstamp", ptb.start_run),
+            ("runid", ptb.md5sum),
+            ("experiment", experiment)
+        ])
```
ndokos: Is this different from
     _dict_const(runtstamp = ptb.start_run, ...)

? Not sure why you are using a list of tuples here. Is "experiment" causing a problem?

server/pbench/bin/index-pbench
```diff
@@ -1497,10 +1497,10 @@ def hostnames_if_ip_from_sosreport(sos):
...
         pass
 
-    d = {
-        'hostname-f': hostname_f,
-        'hostname-s': hostname_s
-    }
+    d = _dict_const([
+        ('hostname-f', hostname_f),
+        ('hostname-s', hostname_s)
+    ])
```
ndokos: See previous comment.

server/pbench/bin/index-pbench
```diff
@@ -1613,13 +1613,14 @@ def mk_metadata(fname, tb, mdconf, md5sum):
...
-              'md5': md5sum
-    }
+    mddict = _dict_const([
+        ('generated-by', _NAME_),
+        ('generated-by-version', _VERSION_),
+        ('pbench-agent-version', mdconf.get("pbench", "rpm-version")),
+        ('file-date', mtime.strftime("%Y-%m-%d")),
+        ('file-name', fname),
+        ('md5', md5sum)
+    ])
```
ndokos: See previous comment. Are the dashes causing a problem here?

server/pbench/bin/index-pbench
```diff
@@ -1631,7 +1632,7 @@ def mk_run_action(ptb, options, INDEX_PREFIX, opctx):
        the tarball: its name, size, md5, mtime.
        Metadata about the run are *data* to be indexed.
     """
-    source = { '@timestamp': ptb.start_run.replace('_', 'T') }
+    source = _dict_const([('@timestamp', ptb.start_run.replace('_', 'T'))])
```
ndokos: '@' sign causing a problem? I guess I see a pattern...

server/pbench/bin/index-pbench
portante: That syntax would also work here, but the list syntax is usable in more places, but I did not try to be consistent since both approaches are valid.

server/pbench/bin/index-pbench
portante: I don't think you can use a "-" in an identifier, right?

server/pbench/bin/index-pbench
portante: Yes, a "-" in an identifier is not allowed, right?

PR #892, opened 2018-08-24
Ensure UTC for timezone in unittests

portante:
And we make TMPDIR available for all tests.


portante:
As a result of working on #890, if we use `0` time values, they are converted to the timezone in which the unit tests are running.  So this ensures that does not happen.

Prior to PR #890 we don't to do this, but with PR #890 this starts to happen.

chaitanyaenr:
Looks like one of the unit test failed because of some transient travis issues, restarting the job.

PR #896, opened 2018-08-24
Rework fio p

portante:
Supersedes #849.


atheurer:
I need to clone, run this, and fix a couple things.  There was at least one mistake Karl found, and it's been so long, that I don't even understand what the code is doing until I run it myself again :)

atheurer:
There is one big issue we need to resolve, and that is per-job fio parameters. In a nutshell: there are global options and there are job options, and there can be more than one "job" in a fio-job (thanks for using the same name, fio!).  Right now, if you don't specify a custom fio job file when using pbench-fio, the per-job options will always be the same, so recording the benchmark parameters is easy.  If, however, someone uses a custom job file and has either different value for the same option (like one job has rw=read, while another has rw=write), then we have to store these in a way that we can search them later.  For now, I have new code that just documents an option like: rw=a,b,c,,,,n, where a-n are the values use for each job, in order, so an fio job file with 2 job sections, the first with rw=read and the second with rw=write, would generate the following in our result.json: rw=read,write.  @bengland2 @k-rister @acalhoun is this good enough for you?

BTW, has anyone -ever- used the custom job file option with pbench-fio?  I am not sure it even works...

k-rister:
The rw=read,write seems find to me.  I agree that there is potential for very complicated scenarios with fio, especially when the jobs are not all the same.

I used to use the custom job file option, but it has been a few months since I switched back to using the CLI options.  I think I used the custom job file option when I was tweaking parameters that were not directly exposed via options and it worked ok.

atheurer:
Never mind, I am recreating all of this in a are branch and a new PR....

atheurer:
Replacement PR: https://github.com/distributed-system-analysis/pbench/pull/916

PR #899, opened 2018-08-27
Attempt to ensure unit tests pass in all env

portante:
It is not sufficient to use an OrderedDict in all cases to get passing
unittests in different environments, because accessing the list of
files via the file system can also yield a non-alphabetical order of
the tool names.

We just have sort the templates by tool name before we invoke the
template creation method.


portante:
Verified via Matt B. in his environment.

PR #916, opened 2018-09-05
fio-postprocess: more fixup to process fio options

atheurer:
Options from fio-result.txt are converted into
benchmark parameters for result.json.  This fixes
the per-job (in a job) options and concatinates those
together into a single benchmark-parameter, like:
  rw=read,read,read,write
The unittest output files for this are so old...
     -HOW OLD ARE THEY!?!
They are so old, there have been major version updates
of fio benchmark since the output files were added!
So, this update includes fio output that is actually
relevant, like output that includes the fio job options
we need to process -the whole point of this feature.


atheurer:
Supersedes https://github.com/distributed-system-analysis/pbench/pull/896

atheurer:
Re-opening this, as I have found out, trying to add more to this is just a nightmare w.r.t unittests, so let's apply this first (fix for fio parameters) and worry about the other stuff in a different PR.

portante:
@atheurer, is this a PR you still want to get merged?

atheurer:
@portante yes, I want this PR merged

Issue #871, opened 2018-08-08
pprof tool script eats a core


sjug:
The tool-scripts/pprof will consume 1 core for length of test run. This seems to be due to the new while & wait pattern?

strace:
```
...
rt_sigprocmask(SIG_BLOCK, [CHLD], [], 8) = 0
rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
rt_sigprocmask(SIG_BLOCK, [CHLD], [], 8) = 0
rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
rt_sigprocmask(SIG_BLOCK, [CHLD], [], 8) = 0
rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
...
```

portante:
How short is the interval in use?

If the spawned processes complete before the specified interval, e.g. when there is a bug in the command that is used which results in no wait, then the wait loop will consume CPU spinning.

Adding a sleep to a [target time interval](https://stackoverflow.com/questions/645992/bash-sleep-until-a-specific-time-date) [1] would prevent that from happening.

[1] https://stackoverflow.com/questions/645992/bash-sleep-until-a-specific-time-date

sjug:
Fixed by https://github.com/distributed-system-analysis/pbench/pull/919

PR #919, opened 2018-09-06
Ensure pprof executes commands once per interval

portante:
Fixes issue #871.

### DIFF
--- a/agent/tool-scripts/pprof
+++ b/agent/tool-scripts/pprof
@@ -119,6 +119,9 @@ collect_data() {
 get_masters() {
     declare -A master_pids
     while true; do
+        # Calculate the absolute time of the interval
+        current_epoch=$(date +%s.%N)
+        target_epoch=$(( $current_epoch + $interval ))
         for master in "${masters[@]}"; do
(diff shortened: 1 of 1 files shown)


sjug:
I've tried this out and it seems to break pprof collection in the sense that no samples are collected, the result directory is empty and no pprof commands are executed.

portante:
@sjug, thanks, I refactored the code in newly added commit.  Let's see if that helps us debug what is going on.  Removing redundancy is a good place to start when debugging a problem.

agent/tool-scripts/pprof
```diff
@@ -100,77 +99,121 @@ done
...
+    suffix=$4
+    oc get --raw /debug/pprof/heap --server ${server} > ${pbench_tmp}/${target}${suffix}_heap && tool ${tool} -text -seconds=${interval} -output ${PPROF_TMPDIR}/${target}/${datestamp}${suffix}_heap.txt ${pbench_tmp}/${target}${suffix}_heap
+}
+
+_pprof_profile() {
+    datestamp=$1
+    server=$2
+    target=$3
+    suffix=$4
+    oc get --raw /debug/pprof/profile?seconds=${interval} --server ${server} > ${pbench_tmp}/${target}${suffix}_profile && tool ${tool} -text -seconds=${interval} -output ${PPROF_TMPDIR}/${target}/${datestamp}${suffix}_profile.txt ${pbench_tmp}/${target}${suffix}_profile
```
sjug: `${tool_bin} tool ${tool}`

chaitanyaenr:
Verified that the tool is working as expected.

PR #921, opened 2018-09-07
Use ${tool_bin} instead of "go" in pprof

portante:
Final fix for issue #871.


portante:
@sjug, @chaitanyaenr, missed this commit some how.  Can you two review again?

PR #928, opened 2018-09-14
Fix bugs, clean up crontab output, enhance unit tests

ndokos:
Quote variables in conditional tests.

Rearrange the crontab output, so that env variables are list first
in the crontab, before all the rules.

Enhance the unittests script for test-8 (the activate test) to include
the generated crontab in the gold output. Also add start/stop
delimiters.


server/pbench/bin/pbench-server-activate-create-crontab
```diff
@@ -21,31 +21,43 @@ mkdir -p $crontabpath
...
-    local role=$1
-    local host mailfrom mailto
+function crontab_header() {
+    role=$1
+    local mailfrom mailto
+
+    grep -q PATH= $crontab ||
+        echo "PATH=$bindir:$PATH" >> $crontab
+    grep -q CONFIG $crontab ||
+        echo "CONFIG=$CONFIG" >> $crontab
```
portante: For these two statements, can you add a comment explaining what they do?  It seems subtle enough to warrant one.

server/pbench/bin/pbench-server-activate-create-crontab
```diff
@@ -54,13 +66,21 @@ function crontab_normal() {
...
         echo "$role: No mailto specified"
+    fi
 
+    if [[ ! -z "$mailto" ]] ;then
+        grep -q MAILTO $crontab ||
+            echo "MAILTO=$mailto" >> $crontab
     fi
+    if [[ ! -z "$mailfrom" ]] ;then
+        grep -q MAILFROM $crontab ||
+            echo "MAILFROM=$mailfrom" >> $crontab
```
portante: And a comment for these two ...

server/pbench/bin/unittests
```diff
@@ -81,6 +82,10 @@ function _run_activate {
     oldpath=$PATH
     PATH=$_tdir:$PATH
     $_tdir/pbench-server-activate ${_testdir}/opt/pbench-server/lib/config/pbench-server.cfg ${_testdir} >> $_testout
+    echo "--- Finished activate for $test (status=$?)" >> $_testout
+    echo "+++ ${_testdir}/opt/pbench-server/lib/crontab/crontab" >> $_testout
+    grep -v PATH= ${_testdir}/opt/pbench-server/lib/crontab/crontab >> $_testout
```
portante: Can you add a comment for why we ignore the `PATH` variable setting?   One guess would be that it is because it is different between tests.  Could we consider searching for the expected "full" `PATH` variable value to skip that, so that if the path is not what is expected, then `grep` command won't match, allowing the unit tests will fail with that path information?

ndokos: Reworked the mailto/mailfrom settings in the `pbench-server-activate-create-crontab` script and simplified the structure by getting rid of unneeded generality. Reworked the unittests script to simplify it and avoid some pain points. PTAL and let me know.

PR #930, opened 2018-09-17
Add support to run stockpile along with collect sysinfo

### LINKED ISSUE #915: Allow stockpile collection along with sysinfo collection
Pbench collects system information including block, libvirt, kernel config, security mitigations,  sos topology e.t.c after each benchmark run. Adding support to run stockpile along with collect sysinfo will gather info about openstack, openshift, dmidecode, sosreport e.t.c. This way we can capture the cluster state after each benchmark run. A sample use case would be to collect info about the state of the OpenShift cluster after each scale test. 

There's going to be some data duplication if we are running both but we can ignore that for now as the plan is to eventually replace "collect sysinfo" with stockpile once we are sure that the stockpile is collecting everything that "collect sysinfo" does.

chaitanyaenr:
Stockpile is as part of collect sysinfo to gather system information
about controller and registered remotes. Eventually, collect sysinfo
will be replaced with stockpile once it has support to gather all
the info that it's collecting.

Fixes #915

### DIFF
--- a/agent/base
+++ b/agent/base
@@ -266,10 +266,19 @@ function verify_common_bench_script_options {
 
 # generate inventory file with controller and remotes
 function generate_inventory {
-    echo "[controller]"
-    echo $HOSTNAME
-    echo "[remote]"
-    ls $tools_group_dir | awk -F@ '/^remote/ {print $2};'
+	local component=$1
+	if [[ $component == "stockpile" ]]; then
+		echo "[stockpile]"
+		echo $HOSTNAME
+		echo "[all]"
+		echo $HOSTNAME
+		ls $tools_group_dir | awk -F@ '/^remote/ {print $2};'
+	else
+		echo "[controller]"
+		echo $HOSTNAME
+		echo "[remote]"
+		ls $tools_group_dir | awk -F@ '/^remote/ {print $2};'
+	fi
 }
 
 if [[ "${_PBENCH_BENCH_TESTS}" == 1 ]] ;then
--- a/agent/util-scripts/gold/pbench-collect-sysinfo/test-24.txt
+++ b/agent/util-scripts/gold/pbench-collect-sysinfo/test-24.txt
@@ -1,5 +1,5 @@
 +++ Running test-24 pbench-collect-sysinfo --options
-default, none, all, block, libvirt, kernel_config, security_mitigations, sos, topology, ara
+default, none, all, block, libvirt, kernel_config, security_mitigations, sos, topology, ara, stockpile
 --- Finished test-24 pbench-collect-sysinfo (status=0)
 +++ pbench tree state
 /var/tmp/pbench-test-utils/pbench
--- a/agent/util-scripts/pbench-collect-sysinfo
+++ b/agent/util-scripts/pbench-collect-sysinfo
@@ -18,13 +18,13 @@ group=default
 dir="/tmp"
 # array containing all the possible sysinfo options
 sysinfo_opts_default=( block libvirt kernel_config security_mitigations sos topology )
-sysinfo_opts_available=( block libvirt kernel_config security_mitigations sos topology ara )
+sysinfo_opts_available=( block libvirt kernel_config security_mitigations sos topology ara stockpile )
 # get comma separated values
 sysinfo_opts_default_comma_separated=$(IFS=,; echo "${sysinfo_opts_default[*]}")
 sysinfo_opts_available_comma_separated=$(IFS=,; echo "${sysinfo_opts_available[*]}")
 check=false
 
-# Display sysinfo options 
+# Display sysinfo options
 function display_sysinfo_opts {
 	printf "default, none, all"
 	for item in ${sysinfo_opts_available[*]}; do printf ", %s" $item ; done
@@ -106,7 +106,7 @@ fi
 
 # check if the input sysinfo parameter passed by the user is a valid option or not
 if $check; then
-	debug_log "[$script_name]: sysinfo option is set to $sysinfo" 
+	debug_log "[$script_name]: sysinfo option is set to $sysinfo"
 	if [ "$sysinfo" == "all" ] || [ "$sysinfo" == "default" ] || [ "$sysinfo" == "none" ]; then
 		:
 	else
--- a/agent/util-scripts/pbench-sysinfo-dump
+++ b/agent/util-scripts/pbench-sysinfo-dump
@@ -48,7 +48,7 @@ function collect_mitigation_data {
 	# check the generic vulnerabilities files first
 	if [ -d  /sys/devices/system/cpu/vulnerabilities ] ;then
 		grep -Hs . /sys/devices/system/cpu/vulnerabilities/* >> $dir/security-mitigation-data.txt
-	fi 
+	fi
 	# then check the RHEL-specific flag settings files - only
 	# applicable on x86_64
 	if [ -d /sys/kernel/debug/x86 ] ;then
@@ -134,6 +134,42 @@ function collect_ara_data {
 	fi
 }
 
+function collect_stockpile_data {
+	# parse config
+	stockpile_path=$(getconf.py stockpile_path stockpile)
+	stockpile_user=$(getconf.py stockpile_user stockpile)
+	stockpile_local_remote_user=$(getconf.py local_remote_user stockpile)
+	stockpile_host_remote_user=$(getconf.py host_remote_user stockpile)
+	stockpile_output_path=$(getconf.py stockpile_output_path stockpile)
+	stockpile_log=$(getconf.py stockpile_log stockpile)
+
+	if [[ -z "$stockpile_path" ]]; then
+		stockpile_path=$(getconf.py pbench_install_dir pbench-agent)
+	fi
+
+	# set stockpile log file and output paths
+	if [[ -z "$stockpile_log" ]]; then
+                stockpile_log=$dir/stockpile.log
+        fi
+	if [[ -z "$stockpile_output_path" ]]; then
+        	stockpile_output_path=$dir/stockpile.json
+	fi
+
+	# generate inventory and run stockpile
+	generate_inventory stockpile > $INVENTORY
+	debug_log "[$script_name]Collecting stockpile data"
+	stockpile_opts="stockpile_user=$stockpile_user local_remote_user=$stockpile_local_remote_user host_remote_user=$stockpile_host_remote_user stockpile_output_path=$stockpile_output_path"
+	stockpile_playbook="$stockpile_path/stockpile/stockpile.yml"
+	echo "ansible-playbook -vv --extra-vars '"$stockpile_opts"' -i $INVENTORY $stockpile_playbook" > ${benchmark_run_dir}/stockpile.cmd
+	chmod +x ${benchmark_run_dir}/stockpile.cmd
+	debug_log "Running stockpile with options: $stockpile_opts"
+	${benchmark_run_dir}/stockpile.cmd 2>&1 >$stockpile_log
+	if [[ $? != 0 ]]; then
+        	error_log "[$script_name] Stockpile run failed"
+		exit 1
+	fi
+}
+
 for item in ${sysinfo//,/ };do
 	debug_log "[$script_name]: collecting $item"
 	if [[ "$item" == "kernel_config" ]]; then
@@ -150,6 +186,8 @@ for item in ${sysinfo//,/ };do
 		collect_block &
 	elif [[ "$item" == "ara" ]]; then
 		collect_ara_data &> $pbench_log
+	elif [[ "$item" == "stockpile" ]]; then
+		collect_stockpile_data &
 	else
 		error_log "[$script_name]bad sysinfo value"
 	fi
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "agent/stockpile"]
+	path = agent/stockpile
+	url = https://github.com/redhat-performance/stockpile
--- a/agent/bench-scripts/gold/test-benchmark-clis/test-CL.txt
+++ b/agent/bench-scripts/gold/test-benchmark-clis/test-CL.txt
@@ -10,7 +10,7 @@ Bench Script: pbench-cyclictest --help
 		-s --stress                                   run stress process on the same cpu as --cpu
 		--tool-group=str
 		--sysinfo=str,                                str= comma separated values of sysinfo to be collected
-		                                                   available: default, none, all, block, libvirt, kernel_config, security_mitigations, sos, topology, ara
+		                                                   available: default, none, all, block, libvirt, kernel_config, security_mitigations, sos, topology, ara, stockpile
 
 Bench Script: pbench-cyclictest --tool-group=bad --sysinfo=bad
 ------------
@@ -26,7 +26,7 @@ Bench Script: pbench-cyclictest --tool-group=bad --sysinfo=bad
 		-s --stress                                   run stress process on the same cpu as --cpu
 		--tool-group=str
 		--sysinfo=str,                                str= comma separated values of sysinfo to be collected
-		                                                   available: default, none, all, block, libvirt, kernel_config, security_mitigations, sos, topology, ara
+		                                                   available: default, none, all, block, libvirt, kernel_config, security_mitigations, sos, topology, ara, stockpile
 
 Bench Script: pbench-cyclictest --bad-to-the-bone 
 ------------
@@ -42,7 +42,7 @@ pbench-cyclictest --bad-to-the-bone
 		-s --stress                                   run stress process on the same cpu as --cpu
 		--tool-group=str
 		--sysinfo=str,                                str= comma separated values of sysinfo to be collected
-		                                                   available: default, none, all, block, libvirt, kernel_config, security_mitigations, sos, topology, ara
+		                                                   available: default, none, all, block, libvirt, kernel_config, security_mitigations, sos, topology, ara, stockpile
 
 Bench Script: pbench-dbench --help 
 ------------
@@ -77,7 +77,7 @@ Bench Script: pbench-dbench --help
 		                                        part of the label, just use --tool-label-pattern= to tell this script
 		                                        the prefix pattern to use for CPU information.
 		             --sysinfo=str,             str= comma separated values of sysinfo to be collected
-		                                             available: default, none, all, block, libvirt, kernel_config, security_mitigations, sos, topology, ara
+		                                             available: default, none, all, block, libvirt, kernel_config, security_mitigations, sos, topology, ara, stockpile
 
 Bench Script: pbench-dbench --tool-group=bad --sysinfo=bad
 ------------
@@ -116,7 +116,7 @@ Bench Script: pbench-dbench --tool-group=bad --sysinfo=bad
 		                                        part of the label, just use --tool-label-pattern= to tell this script
 		                                        the prefix pattern to use for CPU information.
 		             --sysinfo=str,             str= comma separated values of sysinfo to be collected
-		                                             available: default, none, all, block, libvirt, kernel_config, security_mitigations, sos, topology, ara
+		                                             available: default, none, all, block, libvirt, kernel_config, security_mitigations, sos, topology, ara, stockpile
 
 Bench Script: pbench-dbench --bad-to-the-bone 
 ------------
@@ -155,7 +155,7 @@ pbench-dbench --bad-to-the-bone
 		                                        part of the label, just use --tool-label-pattern= to tell this script
 		                                        the prefix pattern to use for CPU information.
 		             --sysinfo=str,             str= comma separated values of sysinfo to be collected
-		                                             available: default, none, all, block, libvirt, kernel_config, security_mitigations, sos, topology, ara
+		                                             available: default, none, all, block, libvirt, kernel_config, security_mitigations, sos, topology, ara, stockpile
 
 Bench Script: pbench-fio --help 
 ------------
@@ -241,7 +241,7 @@ The following options are available:
 		set the histogram logging interval in seconds (default 10)
 
 	--sysinfo=str            str= comma separated values of sysinfo to be collected
-		available: default, none, all, block, libvirt, kernel_config, security_mitigations, sos, topology, ara
+		available: default, none, all, block, libvirt, kernel_config, security_mitigations, sos, topology, ara, stockpile
 
 Bench Script: pbench-fio --tool-group=bad --sysinfo=bad
 ------------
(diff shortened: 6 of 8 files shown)


agent/util-scripts/pbench-collect-sysinfo
```diff
@@ -242,6 +242,55 @@ if [ $lcl_cnt -gt 0 ]; then
...
+	stockpile_output_path=$benchmark_run_dir/stockpile.json
+fi
+
+# Genrate inventory and run stockpile
+generate_stockpile_inventory > $stockpile_inventory
+echo "Collecting stockpile data"
+debug_log "[$script_name]Collecting stockpile data"
+ansible-playbook -vv --extra-vars "stockpile_user=$stockpile_user local_remote_user=$stockpile_local_remote_user host_remote_user=$stockpile_host_remote_user stockpile_output_path=$stockpile_output_path" -i $stockpile_inventory $stockpile_path/stockpile/stockpile.yml 2>&1 > "$stockpile_log"
+if [[ $? != 0 ]]; then
+	error_log "[$script_name] Stockpile run failed"
```
chaitanyaenr: The benchmark run will not fail even if stockpile fails for now.

agent/util-scripts/pbench-collect-sysinfo
```diff
@@ -242,6 +242,55 @@ if [ $lcl_cnt -gt 0 ]; then
...
+which ansible &>/dev/null
+if [[ $? != 0 ]]; then
+	debug_log "[$script_name] ansibke package is not installed on the system, installing"
+	echo "ansible is not installed on the system, installing"
+	yum install ansible -y &>/dev/null
+fi
+if [[ -d "$stockpile_path/stockpile" ]]; then
+	rm -rf $stockpile_path/stockpile
+fi
+git clone https://github.com/redhat-performance/stockpile.git $stockpile_path/stockpile &>/dev/null
```
chaitanyaenr: Stockpile git repo is cloned for now to get the stockpile code. It would be great if we make a release in stockpile and package it as an rpm. The only problem with this approach is that the stockpile is very fast moving, so we won't get the latest changes.

agent/util-scripts/pbench-collect-sysinfo
```diff
@@ -242,6 +242,55 @@ if [ $lcl_cnt -gt 0 ]; then
...
 wait
 
+stockpile_path=$(getconf.py -C $configfile stockpile_path stockpile)
+stockpile_user=$(getconf.py -C $configfile stockpile_user stockpile)
+stockpile_local_remote_user=$(getconf.py -C $configfile local_remote_user stockpile)
+stockpile_host_remote_user=$(getconf.py -C $configfile host_remote_user stockpile)
+stockpile_output_path=$(getconf.py -C $configfile stockpile_output_path stockpile)
+
+# Install stockpile and its dependencies
+which git &>/dev/null
```
chaitanyaenr: As discussed with @portante and @ndokos, the dependencies should be just installed on the controller, not on all the remotes. We are working on a common approach to handle the dependencies during the runtime.

agent/base
```diff
@@ -230,6 +230,17 @@ function generate_inventory {
 	ls $tools_group_dir | awk -F@ '/^remote/ {print $2};'
 }
 
+# generate inventory file with controller and remotes for stockpile
+export stockpile_inventory=/tmp/stockpile_inventory.$$
```
atheurer: Shouldn't this belong in $pbench_run/tmp?

agent/util-scripts/pbench-collect-sysinfo
```diff
@@ -242,6 +242,52 @@ if [ $lcl_cnt -gt 0 ]; then
...
+# install stockpile and its dependencies
+for package in $stockpile_dependencies; do
+	which $package &>/dev/null
+	if [[ $? != 0 ]]; then
+		debug_log "[$script_name] $package is not installed on the system, installing"
+		echo "$package is not installed on the system, installing"
+		yum install $package -y &>/dev/null
+	fi
+done
+if [[ -d "$stockpile_path/stockpile" ]]; then
```
atheurer: So we now have to git clone every time we need to run stockpile?
And we're putting the repo under /opt/pbench-agent?  I am not sure /opt/pbench-agent is the best place for ephemeral storage.  It would probably be more appropriate in /var/lib/pbench-agent.

atheurer:
@chaitanyaenr can you add the ability to have collect-sysinfo only collect stockpile, in the same way you can specify other collectors, like --sysinfo=stockpile?

atheurer:
@chaitanyaenr FYI, this integration IMO does not work that great when just running 'pbench-collect-sysinfo' on the cmdline.  This script requires a --dir to put have a place to put data into, but the playbook for stockpile does not use this.  So, if you run this script, even with --dir, the stockpile json file ends up in "/".  The Ansible playbook should be using $dir, and cannot assume $benchmark_run_dir is available.

chaitanyaenr:
@atheurer thanks for the review, added stockpile as an option to collect-sysinfo and also modified it to use $dir instead of $benchmark_run_dir. Running pbench-collect-sysinfo on the cmdline should work now.

chaitanyaenr:
@atheurer any more changes needed?

agent/base
```diff
@@ -230,6 +230,17 @@ function generate_inventory {
 	ls $tools_group_dir | awk -F@ '/^remote/ {print $2};'
 }
 
+# generate inventory file with controller and remotes for stockpile
```
portante: This looks like a repeat of an existing function.  Any way we can parameterize this and collapse into one?

agent/util-scripts/pbench-sysinfo-dump
```diff
@@ -123,6 +123,58 @@ function collect_ara_data {
...
+	fi
+	git clone https://github.com/redhat-performance/stockpile.git $stockpile_path/stockpile &>/dev/null
+	if [[ $? != 0 ]]; then
+        	echo "Failed to clone stockpile git repository"
+        	error_log "[$script_name] Failed to clone stockpile repository"
+        	exit 1
+	fi
+
+	# set stockpile log file and output paths
+	if [[ -z "$stockpile_log" ]]; then
```
portante: Where would this reference to `stockpile_log` be satisfied if not from the config file?

agent/config/pbench-agent.cfg.example
```diff
@@ -40,8 +40,16 @@ interval = 30
...
 [config]
 path = %(pbench_install_dir)s/config
 files = pbench-agent.cfg
 
+[stockpile]
+stockpile_user = root 
+local_remote_user = root 
+host_remote_user = root
+stockpile_path = /var/lib/pbench-agent
+stockpile_log =
```
portante: Can you add a comment that `stockpile_log` and `stockpile_output_path` are optional, the default being ... (whatever the default is)?

agent/config/pbench-agent.cfg.example
```diff
@@ -40,8 +40,16 @@ interval = 30
...
 
 [config]
 path = %(pbench_install_dir)s/config
 files = pbench-agent.cfg
 
+[stockpile]
+stockpile_user = root 
+local_remote_user = root 
+host_remote_user = root
+stockpile_path = /var/lib/pbench-agent
```
portante: It looks as if `stockpile_path` is optional in the code, with a hard-coded default of `/var/lib/pbench-agent`.  Can you confirm, and if so, make this a required parameter with a comment that describes why `/var/lib/pbench-agent` is used, and remove the hard-coded default in the code?

agent/config/pbench-agent.cfg.example
```diff
@@ -40,8 +40,16 @@ interval = 30
 # pandas-package = python-pandas
 # Fedora has python2-pandas and python3-pandas
 # pandas-package = python2-pandas
+stockpile_dependencies = git, ansible
```
portante: Why does stockpile depend on `git`?

agent/util-scripts/pbench-sysinfo-dump
```diff
@@ -123,6 +123,58 @@ function collect_ara_data {
...
+	stockpile_output_path=$(getconf.py stockpile_output_path stockpile)
+	stockpile_dependencies=$(getconf.py -l stockpile_dependencies packages)
+
+	# install stockpile and its dependencies
+	for package in $stockpile_dependencies; do
+       		which $package &>/dev/null
+        	if [[ $? != 0 ]]; then
+                	debug_log "[$script_name] $package is not installed on the system, installing"
+                	echo "$package is not installed on the system, installing"
+                	yum install $package -y &>/dev/null
```
portante: For now, stockpile is just an ansible playbook.  Are there other functions of pbench-agent that rely on ansible?  If so, can we avoid installing it here, and just reporting to the user that without ansible installed these functions are not available?

agent/util-scripts/pbench-sysinfo-dump
```diff
@@ -123,6 +123,58 @@ function collect_ara_data {
...
+        	fi
+	done
+	if [[ -z "$stockpile_path" ]]; then
+        	stockpile_path=/var/lib/pbench-agent
```
portante: This seems bad to hard-code a default.

agent/util-scripts/pbench-sysinfo-dump
```diff
@@ -123,6 +123,58 @@ function collect_ara_data {
...
+	fi
+	if [[ -d "$stockpile_path/stockpile" ]]; then
+        	rm -rf $stockpile_path/stockpile
+	fi
+	git clone https://github.com/redhat-performance/stockpile.git $stockpile_path/stockpile &>/dev/null
```
portante: Can we use `curl` instead to fetch a release tarball from github?

chaitanyaenr: We don't have a release for stockpile as of now, it would be nice to have a release and maybe an rpm for a stable release like what we have in pbench.

agent/config/pbench-agent.cfg.example
chaitanyaenr: We don't have stockpile packaged, so we are cloning the repo for now. Eventually we will be packaging and installing a stable version of stockpile.

agent/config/pbench-agent.cfg.example
chaitanyaenr: Done.

chaitanyaenr:
@portante made some changes to address the comments, PTAL.

agent/util-scripts/pbench-sysinfo-dump
```diff
@@ -123,6 +123,58 @@ function collect_ara_data {
 	fi
 }
 
+function collect_stockpile_data {
+	# parse config
+	stockpile_path=$(getconf.py stockpile_path stockpile)
```
mrsiano: is this getconf.py part of pbench utils ?!
also looks like we call it many times, we can do something like single call with arguments and return some sort of json string with with answer per argument ?

agent/util-scripts/pbench-sysinfo-dump
mrsiano: you can add TODO here to handle this, once it's ready.
i'll also open issue for that.

mrsiano:
this work out of the box @chaitanyaenr can you apply the changes please to merge this one.
/ltgm

portante:
@chaitanyaenr, would it be okay if @mrsiano made a PR to address the comments requesting changes, but post that PR against your repo so that you could then merge them, and update this PR with the fixes?

We'd like to get this functionality added sooner than later.

agent/base
chaitanyaenr: Fixed.

agent/config/pbench-agent.cfg.example
chaitanyaenr: You are right @portante , it's optional. Removed the hard-coded default, we use $pbench_run defined in the config if it's not set.

agent/util-scripts/pbench-sysinfo-dump
chaitanyaenr: Fixed

chaitanyaenr:
@portante  sure. Most of the comments have been addressed in the latest commit, we can merge @mrsiano's PR with the fixes and get this in, sorry for the delay.

agent/util-scripts/pbench-sysinfo-dump
portante: Seems best to work with `git submodules` [1](https://git-scm.com/docs/git-submodule).  Instead of cloning here, when we package the `pbench-agent` we should pull a specific version of `stockpile` into the source tree so that it ends up installed in `/opt/pbench-agent/...` somewhere.

agent/util-scripts/pbench-sysinfo-dump
portante: @chaitanyaenr, does this question make sense to you?

agent/util-scripts/pbench-sysinfo-dump
```diff
@@ -139,6 +191,8 @@ for item in ${sysinfo//,/ };do
 		collect_block &
 	elif [[ "$item" == "ara" ]]; then
 		collect_ara_data &> $pbench_log
+	elif [[ "$item" == "stockpile" ]]; then
+		collect_stockpile_data
```
portante: Shouldn't this invocation be put in the background like the other sysinfo options?

agent/config/pbench-agent.cfg.example
portante: We have `check_install_rpm` in the `agent/base` include which should take care of such checks.

But let's change this to assume the stockpile code is "part" of the `pbench-agent` install using `git submodules` [(1)](https://git-scm.com/docs/git-submodule).

Further, it seems best to make the `pbench-agent` RPM depend on ansible now, as we are using it in other places in the code, too.

agent/util-scripts/pbench-sysinfo-dump
chaitanyaenr: Yes, modified it to place the log file in sysinfo dir along with the other sysinfo log files.

portante:
@chaitanyaenr, can you rebase this work and comment on some of the review comments still opened and not resolved?

agent/config/pbench-agent.cfg.example
chaitanyaenr: Agree with making the pbench-agent RPM depend on ansible. Will take a look at git submodules  @portante.

agent/base
```diff
@@ -244,10 +244,19 @@ function verify_common_bench_script_options {
 export INVENTORY=/tmp/inventory.$$
 trap "rm -f $INVENTORY" QUIT INT EXIT
 function generate_inventory {
```
portante: Seems like we should promote `generate_inventory` to a separate script that lives in `util-scripts`, perhaps ... or ... leverage the new Perl Ansible module that @atheurer recently did instead.

Having the trap in the base is not great because it fires whether or not a caller invoked generate_inventory or not.

agent/config/pbench-agent.cfg.example
portante: So now that we are using git sub modules, do we still need this?

agent/util-scripts/pbench-sysinfo-dump
```diff
@@ -123,6 +123,50 @@ function collect_ara_data {
...
+	stockpile_path=$(getconf.py stockpile_path stockpile)
+	stockpile_user=$(getconf.py stockpile_user stockpile)
+	stockpile_local_remote_user=$(getconf.py local_remote_user stockpile)
+	stockpile_host_remote_user=$(getconf.py host_remote_user stockpile)
+	stockpile_output_path=$(getconf.py stockpile_output_path stockpile)
+	stockpile_dependencies=$(getconf.py -l stockpile_dependencies packages)
+	stockpile_log=$(getconf.py stockpile_log stockpile)
+
+	# install stockpile and its dependencies
+	for package in $stockpile_dependencies; do
```
portante: Seems like this can be removed now that we'll be packaging it all together.

agent/util-scripts/pbench-sysinfo-dump
```diff
@@ -123,6 +123,50 @@ function collect_ara_data {
...
+                stockpile_log=$dir/stockpile.log
+        fi
+	if [[ -z "$stockpile_output_path" ]]; then
+        	stockpile_output_path=$dir/stockpile.json
+	fi
+
+	# generate inventory and run stockpile
+	generate_inventory stockpile > $INVENTORY
+	debug_log "[$script_name]Collecting stockpile data"
+	ansible-playbook -vv --extra-vars "stockpile_user=$stockpile_user local_remote_user=$stockpile_local_remote_user host_remote_user=$stockpile_host_remote_user stockpile_output_path=$stockpile_output_path" -i $INVENTORY $stockpile_path/stockpile/stockpile.yml 2>&1 > "$stockpile_log"
```
portante: Can you add a `debug_log` line to capture the `ansible-playbook` command line invocation?  Perhaps write the full command to a file, and then invoke the file?  Like we do in benchmark scripts?

chaitanyaenr:
Sorry for the delay @portante, got buried in ocp 4.0 scale testing, will take a look at the comments and make changes accordingly.

portante:
> /hold

What does that mean?

chaitanyaenr:
@portante missed @aakarshg changes when I rebased. It should be good to go now.

portante:
Still waiting for Nick to review before we merge.

agent/util-scripts/pbench-sysinfo-dump
```diff
@@ -127,6 +127,45 @@ function collect_ara_data {
...
+	# parse config
+	stockpile_path=$(getconf.py stockpile_path stockpile)
+	stockpile_user=$(getconf.py stockpile_user stockpile)
+	stockpile_local_remote_user=$(getconf.py local_remote_user stockpile)
+	stockpile_host_remote_user=$(getconf.py host_remote_user stockpile)
+	stockpile_output_path=$(getconf.py stockpile_output_path stockpile)
+	stockpile_log=$(getconf.py stockpile_log stockpile)
+
+	if [[ -z "$stockpile_path" ]]; then
+		stockpile_path=$(getconf.py pbench_install_dir DEFAULT)
```
ndokos: This does not work: it needs to be `getconf.py pbench_install_dir pbench-agent` or `getconf.py pbench_install_dir stockpile`.

agent/util-scripts/pbench-sysinfo-dump
```diff
@@ -127,6 +127,45 @@ function collect_ara_data {
...
+	fi
+	if [[ -d "$stockpile_path/stockpile" ]]; then
+        	rm -rf $stockpile_path/stockpile
```
ndokos: Why is this removed? The playbook is removed along with the directory and the collection fails.

agent/util-scripts/pbench-sysinfo-dump
```diff
@@ -127,6 +127,45 @@ function collect_ara_data {
...
+	if [[ -z "$stockpile_output_path" ]]; then
+        	stockpile_output_path=$dir/stockpile.json
+	fi
+
+	# generate inventory and run stockpile
+	generate_inventory stockpile > $INVENTORY
+	debug_log "[$script_name]Collecting stockpile data"
+	stockpile_opts="stockpile_user=$stockpile_user local_remote_user=$stockpile_local_remote_user host_remote_user=$stockpile_host_remote_user stockpile_output_path=$stockpile_output_path"
+	stockpile_playbook="$stockpile_path/stockpile/stockpile.yml"
+	echo "ansible-playbook -vv --extra-vars $stockpile_opts -i $INVENTORY $stockpile_playbook" >$stockpile_path/stockpile.cmd
```
ndokos: I think the quoting here is wonky: the `--extra-vars` argument is an unquoted space-separated list of `var=value` pairs. This shoulde be 

    echo "ansible-playbook -vv --extra-vars '"$stockpile_opts"' -i $INVENTORY $stockpile_playbook" >$stockpile_path/stockpile.cmd

Also, I don't think the  command should be in $stockpile_path, which if otherwise unspecified is `/opt/pbench-agent`: that should be considered read-only.

portante:
@chaitanyaenr, have you had a chance to consider @ndokos's comments?

chaitanyaenr:
@portante @ndokos made the changes requested, PTAL. Unit tests related to 'pbench-copy-results' failed, will check. Thanks.

agent/config/pbench-agent.cfg.example
```diff
@@ -7,4 +7,4 @@ pbench_web_server = pbench.example.com
 
 [config]
 path = %(pbench_install_dir)s/config
-files = pbench-agent-default.cfg
+files = pbench-agent.cfg
```
ndokos: This should be pbench-agent-default.cfg, as it was before.

portante: Looks like this change was taken care of with the latest commit.

agent/util-scripts/gold/pbench-collect-sysinfo/test-24.txt
```diff
@@ -1,5 +1,6 @@
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
portante: Seems to have been addressed.

agent/util-scripts/pbench-sysinfo-dump
portante: Seems to have been resolved.

agent/util-scripts/pbench-sysinfo-dump
portante: Looks like this was resolved.

agent/util-scripts/pbench-sysinfo-dump
```diff
@@ -134,6 +134,42 @@ function collect_ara_data {
...
+	# parse config
+	stockpile_path=$(getconf.py stockpile_path stockpile)
+	stockpile_user=$(getconf.py stockpile_user stockpile)
+	stockpile_local_remote_user=$(getconf.py local_remote_user stockpile)
+	stockpile_host_remote_user=$(getconf.py host_remote_user stockpile)
+	stockpile_output_path=$(getconf.py stockpile_output_path stockpile)
+	stockpile_log=$(getconf.py stockpile_log stockpile)
+
+	if [[ -z "$stockpile_path" ]]; then
+		stockpile_path=$(getconf.py pbench_install_dir pbench-agent)
```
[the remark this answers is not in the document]
portante: Rather than wait for an update, I'm gonna merge this and post a PR to fix this last comment.

portante:
@ndokos, can you sign off on this before we merge it?

agent/util-scripts/pbench-sysinfo-dump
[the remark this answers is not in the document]
portante: Resolved by PR #1209.

PR #932, opened 2018-09-18
Replaced argparse with optparse.

### LINKED ISSUE #886: Replace use of `optparse` with `argparse`
In the python world, `optparse` is deprecated, so we should make an effort to move to `argparse`, it's replacement.

chawlanikhil24:
Fixes #886 

Now, optparse handles only the cli options that have been passed, whereas argparse also covers the arguments passed oncli along with options passed. Therefore, the output was quite different from optparse output.

1. paser.parse_args() returns a namespace object that also has the file name that has passed onto the cli, whereas optparse returns a dictionary.
2. Therefore, a minimalistic serialization of the output of parse_args output has been done from Namespace to Dictionary so as to look like optparse output
3. After, the supplying of optparse output, there are multiple invocations of those options in main(). Therefore, it appeared more optimal to serialize the output of argparse.

### DIFF
--- a/server/pbench/bin/index-pbench
+++ b/server/pbench/bin/index-pbench
@@ -14,6 +14,8 @@ from urllib3 import Timeout
 from datetime import datetime, timedelta
 from collections import Counter
 from optparse import OptionParser, make_option
+from argparse import ArgumentParser
+
 try:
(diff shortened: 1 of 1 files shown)


chawlanikhil24:
With reference to issue "Replace use of `optparse` with `argparse` #886" , this PR has been created

PR #933, opened 2018-09-18
Validated argparse with optparse

### LINKED ISSUE #886: Replace use of `optparse` with `argparse`

chawlanikhil24:
Fixes #886

Now, optparse handles only the cli options that have been passed, whereas argparse also covers the arguments passed on cli along with options passed. Therefore, the output was quite different from optparse output.

paser.parse_args() returns a namespace object that also has the file name that has passed onto the cli, whereas optparse returns a Values object.
After, the supplying of optparse output, there are multiple invocations of those options in main(). Therefore, it appeared more optimal to serialize the output of argparse.

### DIFF
--- a/server/pbench/bin/index-pbench
+++ b/server/pbench/bin/index-pbench
@@ -13,7 +13,8 @@ import hashlib, json, glob, csv, tarfile, logging
 from urllib3 import Timeout
 from datetime import datetime, timedelta
 from collections import Counter
-from optparse import OptionParser, make_option
+from argparse import ArgumentParser
+
 try:
     from configparser import SafeConfigParser, Error as ConfigParserError, NoSectionError, NoOptionError
 except ImportError:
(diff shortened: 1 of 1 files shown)


server/pbench/bin/index-pbench
```diff
@@ -2073,25 +2074,24 @@ def main(options, args):
...
     parser.set_defaults(tmpdir = os.environ.get("TMPDIR"))
-
-    (options, args) = parser.parse_args()
-
+    parsed = parser.parse_args()
+    #parse_args returns a namespace object. Therefore this is a minimal serialisation that has been added to process namespaces to normal dictionary
+    parsed = vars(parsed)
+    #argparse also handles the cli argument. Therefore the dictionary already had that element. To make the output argparse similar to optparse this element has been popped
+    args = parsed.pop("path_to_tarball_file")
+    options = parsed
```
ndokos: The unit tests failures are caused by the following: `options` was a `Values` object in optparse, not a dict, so you could refer to an attribute as `options.debug_time_operations`. Now that it is a dict, you have to use index notation: `options['debug_time_operations']` (remember, this is Python, not Javascript ;-) ). But why convert it to a dict in the first place? You could leave it as a Namespace and refer to `parsed.debug_time_operations`.

chawlanikhil24:
@ndokos @portante Updated the PR with the requested changes. PTAL

server/pbench/bin/index-pbench
```diff
@@ -2073,25 +2074,22 @@ def main(options, args):
...
+    parser.add_argument("-U", "--unittest", action="store_true", dest="debug_unittest", help="Run in unittest mode")
+    parser.add_argument("-T", "--time-ops", action="store_true", dest="debug_time_operations", help="Time action making routines")
+    # argparse also handles the cli argument passed to the script, therefore this parser has been added
+    parser.add_argument("path_to_tarball_file",nargs=1,help="specify the path to the tar.gz file")
     parser.set_defaults(cfg_name = os.environ.get("ES_CONFIG_PATH"))
     parser.set_defaults(tmpdir = os.environ.get("TMPDIR"))
-
-    (options, args) = parser.parse_args()
-
+    parsed = parser.parse_args()
```
portante: Seems like if you used `options` instead of `parsed`, you'd be able to reduce these lines to:

```
    options = parser.parse_args()
    args = options.path_to_tarball_file
    del options.path_to_tarball_file
```

And the deletion of `path_to_tarball_file` from options does not seem necessary, which would bring those lines changed down to two.

server/pbench/bin/index-pbench
```diff
@@ -2073,25 +2074,22 @@ def main(options, args):
...
+    parser = ArgumentParser("Usage: index-pbench [--config <path-to-config-file>] [--metadata \"<JSON doc>\"] [--working-directory <dir>] <path-to-tarball>")
+    parser.add_argument("-C", "--config", dest="cfg_name", help="Specify config file")
+    parser.add_argument("-E", "--indexing-errors", dest="indexing_errors", help="Name of a file to write JSON documents that fail to index")
+    parser.add_argument("-D", "--temp-directory", dest="tmpdir", help="Temporary directory to use while indexing")
+    parser.add_argument("-M", "--metadata", dest="metadata_string", help="Specify additional metadata (e.g. for browbeat) as a JSON document string")
+    # options for debuggging and unit testing
+    parser.add_argument("-U", "--unittest", action="store_true", dest="debug_unittest", help="Run in unittest mode")
+    parser.add_argument("-T", "--time-ops", action="store_true", dest="debug_time_operations", help="Time action making routines")
+    # argparse also handles the cli argument passed to the script, therefore this parser has been added
+    parser.add_argument("path_to_tarball_file",nargs=1,help="specify the path to the tar.gz file")
```
ndokos: It's not for this PR, but we eventually might want the flexibility of calling it with more than one tarball.

PR #934, opened 2018-09-20
Scripts to facilitate restoring results from tarballs

ndokos:
WARNING: these are *completely* untested. This is meant as a starting
point for discussion.

These scripts are meant to be interactively used.

Most of the requests we've had so far have been prefix requests, not
individual tarballs. The scripts tarballs-from-prefix and
unpack-tarballs are meant to address this situation:

 - tarballs-from-prefix <controller> <prefix>

   This script takes two arguments, the controller and the prefix and
   returns a list of tarballs for that controller that should be under
   that prefix.  Each element of the list is just a tarball name, not
   a path.

 - unpack-tarballs <controller> <tarball> ...

   This script takes as arguments the controller and a list of tarball
   names (as returned by tarballs-from-prefix). It checks each tarball
   and if conditions are right, it changes its state to TO-UNPACK,
   thereby notifying the pbench-unpack-tarballs script to unpack the
   tarball.

   The intended workflow is:

   unpack-tarballs <controller> $(tarballs-from-prefix <controller> <prefix>)

   But unpack-tarballs can also deal with a bunch of individual
   tarballs (whether they have a prefix or not). If the tarball does
   have an old-style prefix (prefix.<resultname>), it is renamed
   to the new style, <resultname>.prefix before the state of the
   tarball is changed. When pbench-unpack-tarballs processes it,
   it will DTRT wrt the prefix.

### DIFF
--- a/server/pbench/scripts/tarballs-from-prefix
+++ b/server/pbench/scripts/tarballs-from-prefix
@@ -1,0 +1,40 @@
+#! /bin/bash
+
+# we asssume that these scripts are installed in /opt/pbench-server/scripts
+# on the server.
+
+###########################################################################
+# load common things
+opts=$SHELLOPTS
+case $opts in
+    *xtrace*)
+        dir=$(dirname $(which $0))/../bin
+        PROG=$(basename $(which $0))
+        ;;
+    *)
+        dir=$(dirname $0)/../bin
+        PROG=$(basename $0)
+        ;;
+esac
+
+# TOP, ARCHIVE, INCOMING, RESULTS, USERS are all defined by the base file
+. $dir/pbench-base.sh
+###########################################################################
+ 
+controller=$1
+if [ -z "$controller" ] ;then
+   exit 1
+fi
+shift
+prefix=$1
+
+cd $ARCHIVE/$controller || exit 1
+
+# find all the prefix files that contain the given prefix
+# and return the list (slightly munged)
+
+grep -H $prefix .prefix/prefix.* |
+    sed 's/^.prefix\/prefix.//
+         s/:.*/.tar.xz/'
+
+exit 0
--- a/server/pbench/scripts/tarballs-from-prefix
+++ b/server/pbench/scripts/tarballs-from-prefix
@@ -30,11 +30,13 @@ prefix=$1
 
 cd $ARCHIVE/$controller || exit 1
 
-# find all the prefix files that contain the given prefix
-# and return the list (slightly munged)
+# find all the prefix files that contain the given prefix and return
+# the list (slightly munged to contain just the result name) - we
+# assume that the prefix files follow the current convention
+# (<resultname>.prefix) rather than the old one (prefix.<resultname>).
 
-grep -H $prefix .prefix/prefix.* |
-    sed 's/^.prefix\/prefix.//
-         s/:.*/.tar.xz/'
+grep -H $prefix .prefix/*.prefix |
+    sed 's/^.prefix\///
+         s/\.prefix:.*/.tar.xz/'
 
 exit 0
--- a/server/pbench/scripts/tarballs-from-prefix
+++ b/server/pbench/scripts/tarballs-from-prefix
@@ -35,7 +35,8 @@ cd $ARCHIVE/$controller || exit 1
 # assume that the prefix files follow the current convention
 # (<resultname>.prefix) rather than the old one (prefix.<resultname>).
 
-grep -H $prefix .prefix/*.prefix |
+# match the prefix exactly
+grep -H "^${prefix}$" .prefix/*.prefix |
     sed 's/^.prefix\///
          s/\.prefix:.*/.tar.xz/'
 
--- a/server/pbench/scripts/tarballs-from-prefix
+++ b/server/pbench/scripts/tarballs-from-prefix
@@ -35,8 +35,8 @@ cd $ARCHIVE/$controller || exit 1
 # assume that the prefix files follow the current convention
 # (<resultname>.prefix) rather than the old one (prefix.<resultname>).
 
-# match the prefix exactly
-grep -H "^${prefix}$" .prefix/*.prefix |
+# match the prefix exactly and literally, not as a regexp
+grep -xHF "${prefix}" .prefix/*.prefix |
     sed 's/^.prefix\///
          s/\.prefix:.*/.tar.xz/'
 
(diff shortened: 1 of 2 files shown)


server/pbench/scripts/tarballs-from-prefix
```diff
@@ -0,0 +1,40 @@
...
+fi
+shift
+prefix=$1
+
+cd $ARCHIVE/$controller || exit 1
+
+# find all the prefix files that contain the given prefix
+# and return the list (slightly munged)
+
+grep -H $prefix .prefix/prefix.* |
```
portante: Seems like we might want to use `.prefix/*.prefix` instead here, which is the way the latest code creates prefix files.

server/pbench/scripts/unpack-tarballs
```diff
@@ -0,0 +1,92 @@
...
+   exit 1
+fi
+shift
+tarballs="$@"
+
+cd $ARCHIVE/$controller || exit 1
+
+incoming=$INCOMING/$controller
+
+linksrc=MOVE_UNPACKED
```
portante: Should this be `MOVE-UNPACKED`?

server/pbench/scripts/unpack-tarballs
```diff
@@ -0,0 +1,92 @@
...
+    if [ -d "$resultdir" ] ;then
+        # unpacked already
+        echo "skipping $resultdir: unpacked already"
+        continue
+    elif [ -L "$resultdir" ] ;then
+        tmpdir=$(readlink -f "$resultdir")
+        if [ -d "$tmpdir" ] ;then
+            echo "skipping $resultdir: it is a link to $tmpdir"
+            continue
+        fi
```
portante: If it is not a link to `$tmpdir`, what would it be a link to?  Does this indicate a problem that we need to flag?  And if so, should we not do anything with this result?

server/pbench/scripts/unpack-tarballs
```diff
@@ -0,0 +1,92 @@
...
+    fi
+
+    # Let it be processed by pbench-unpack-tarballs
+
+    # Check that the prefix file is named properly
+    # and rename it if necessary (like shim-001 does).
+    if [ -f ".prefix/prefix.$resultname" ] ;then
+        mv .prefix/prefix.$resultname .prefix/$resultname.prefix
```
portante: Seems like we should drop this code in favor of fixing up the `.prefix` directories in a separate script ahead of time.  What do you think?

server/pbench/scripts/unpack-tarballs
```diff
@@ -0,0 +1,92 @@
...
+        # pbench-unpack-tarballs will deal with the prefix now.
+    fi
+    mv $linksrc/$tb $linkdest
```
portante: To be safe, we should probably add a trailing slash to be sure the target is used as a directory.

server/pbench/scripts/tarballs-from-prefix
ndokos: OK

server/pbench/scripts/unpack-tarballs
ndokos: Yes, thanks.

server/pbench/scripts/unpack-tarballs
ndokos: Indeed: since we are assuming that the prefix files have been fixed up already in the grep above, we don't need to fix them up here.

server/pbench/scripts/unpack-tarballs
ndokos: OK

server/pbench/scripts/unpack-tarballs
ndokos: I added some flagging in the most recent commit - let me know what you think.

server/pbench/scripts/tarballs-from-prefix
```diff
@@ -0,0 +1,43 @@
...
+
+cd $ARCHIVE/$controller || exit 1
+
+# find all the prefix files that contain the given prefix and return
+# the list (slightly munged to contain just the result name) - we
+# assume that the prefix files follow the current convention
+# (<resultname>.prefix) rather than the old one (prefix.<resultname>).
+
+# match the prefix exactly
+grep -H "^${prefix}$" .prefix/*.prefix |
```
portante: Should this be `grep -HF` to avoid any patterns in the prefix string messing things up?

server/pbench/scripts/unpack-tarballs
```diff
@@ -0,0 +1,91 @@
...
+    for tb in $tarballs ;do
+        echo $tb >> $tblist
+    done
+fi
+
+while read tb ;do
+    # tb is a simple name: foo.tar.xz, not a path.
+    resultname=${tb%.tar.xz}
+
+    if [ ! -f "$resultname.tar.xz.md5" ] ;then
```
portante: Does this assume that the tar balls given the command line are in the archive controller area and are given as the full path there?

server/pbench/scripts/unpack-tarballs
```diff
@@ -0,0 +1,91 @@
...
+            echo "Cannot happen: $resultdir points to $tmpdir which is not a directory."
+            continue
+        fi
+    else
+        echo "Cannot happen: $resultdir is neither a directory, nor a link."
+        continue
+    fi
+
+    # Let it be processed by pbench-unpack-tarballs
+    mv $linksrc/$tb $linkdest/
```
portante: Don't we have to check that the `$linksrc/$tb` exists?  And wouldn't it be `$linksrc/$resultname`?

server/pbench/scripts/tarballs-from-prefix
ndokos: But of course!

server/pbench/scripts/unpack-tarballs
ndokos: Yes to the first question, no to the second; the link is to the tarball `<resultname>.tar.xz`.

server/pbench/scripts/unpack-tarballs
ndokos: This needs more work: I'm assuming a whole lot that is not necessarily true. Update coming.

server/pbench/scripts/tarballs-from-prefix
ndokos: Is there a way to interpret it as a literal string but still have it anchored?

ndokos: -x seems to work.

PR #935, opened 2018-09-21
Improve unpack tar balls error handling

portante:
This is not a complete fix for error handling by any means, but it does
mean we will detect more errors with symlinks in TO-UNPACK, and clean
up better on some of the error conditions.

We have improved the "find" commands used to fetch all the data about
the tar balls to be processed so that we don't have to use the "du"
command, and we still process all the symlinks, including the bad ones.

We have added a unit test which exercises all the functional code
paths of the pbench-unpack-tarballs code.  It does not yet come close
to getting coverage of all the minor error paths.

We have made changes to those error paths so that we can tell a bit
more about what result tar ball had those errors.

Lastly, if a result tar ball has already been unpacked into the incoming
directory (either the temporary or the real one), we now error out so
that we can track that error condition.


server/pbench/bin/pbench-unpack-tarballs
```diff
@@ -107,79 +119,91 @@ while read size result ;do
...
         nerrs=$nerrs+1
+        # FIXME: quarantine $result
         continue
     fi
 
     mkdir -p $UNPACK_PATH/$hostname
     status=$?
     if [[ $status -ne 0 ]] ;then
-        echo "$TS: mkdir -p $UNPACK_PATH/$hostname failed: code $status" | tee -a $mail_content >&4
+        echo "$TS: mkdir -p $UNPACK_PATH/$hostname failed $result: code $status" | tee -a $mail_content >&4
```
ndokos: "failed for $result"

server/pbench/bin/pbench-unpack-tarballs
```diff
@@ -107,79 +119,91 @@ while read size result ;do
...
-    # solution will involve renaming the unpacked directory appropriately.
-    if [ ${resultname%%.*} == "DUPLICATE__NAME" ] ;then
-        ndups=$ndups+1
-        continue
-    fi
-
-    let start_time=$(date +%s)
+    let start_time=$(timestamp-seconds-since-epoch)
     pushd $UNPACK_PATH/$hostname > /dev/null 2>&1
     tar --extract --no-same-owner --no-overwrite-dir --touch --file="$result"
```
ndokos: Unpack into tmp and check that names match.

server/pbench/bin/pbench-unpack-tarballs
```diff
@@ -192,12 +216,9 @@ while read size result ;do
...
 
-    # Version 001 agents use a prefix file
-    # If there is a prefix file, create a link as specified by the
-    # prefix.  Even if no prefix is specified, we still want to link
-    # from the results/ hierarchy to incoming/.
-    basedir=$(dirname $link)
-    # pbench-dispatch has already moved it to the .prefix subdir
+    # Version 001 agents use a prefix file.  If there is a prefix file,
+    # create a link as specified in the prefix file.  pbench-dispatch
+    # has already moved it to the .prefix subdir
```
ndokos: Is `basedir` defined somewhere else? If not, we still need to define it: it is used on line 222.

server/pbench/bin/pbench-unpack-tarballs
portante: Thanks, I'll fix that.

server/pbench/bin/pbench-unpack-tarballs
portante: Thanks, working on that now ...

server/pbench/bin/pbench-unpack-tarballs
portante: Yes, I think it is defined at [line 139](https://github.com/distributed-system-analysis/pbench/pull/935/files/016d4b7a155be09dc052c65c4c935fa676896528#diff-1e38c99b24d595c1a1a6f1685f2f6b55R139).

PR #938, opened 2018-09-22
Rework pbench-move-unpacked to not use prefix

portante:
We have reworked this code for pbench-move-unpacked to avoid knowing about
the prefix or user links, by just replacing the incoming symlink with the
newly copied directory.

We also add similar error handling and tar ball gathering logic as we have
in pbench-unpack-tarballs.


server/pbench/bin/pbench-move-unpacked
```diff
@@ -64,125 +102,95 @@ function process_tarball {
...
-        # Remove the link first (issue #771).
-        rm -f $RESULTS/$hostname/$prefix$resultname
-        ln -sf $incoming $RESULTS/$hostname/$prefix$resultname
     fi
-        
+
+    rm -f ${incoming}.link
+    status=$?
+    if [[ $status -ne 0 ]] ;then
+        echo "$TS: WARNING, could remove ${incoming}.link: code $status" | tee -a $mail_content >&4
```
ndokos: "could not ..."

PR #941, opened 2018-09-25
Add ability of index-pbench to emit index patterns

portante:
Running `./index-pbench -I` or `./index-pbench --dump-index-patterns` will emit a list of all the index patterns used, and a short description for what they contain.


ndokos:
@portante: PTAL at the code. Current unit tests pass locally. New unit tests for the new functionality need to be added, but maybe we should first agree whether the output is what we want.

portante:
@ndokos, thanks for making this a reality!

@gurbirkalsi, this will now be able to print out the index name patterns in use for a given configuration.

ndokos:
@portante: are you going to squash or do you want me to do it?

PR #950, opened 2018-10-08
Restore normal ln behavior for unit tests

portante:
We also emit the unit test state directories with mode, size, and the
symlink (present), shortening the path to reduce the volume of output.

### DIFF
--- a/server/pbench/bin/unittests
+++ b/server/pbench/bin/unittests
@@ -109,23 +109,34 @@ function _run_allscripts {
     _run pbench-edit-prefixes
 }
 
+function _local_find {
+    # We create our own local find command so that we don't emit the size
+    # information for directories.  This is due to the fact that on different
+    # file systems empty directories, or directories with small numbers of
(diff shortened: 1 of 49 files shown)


portante:
Depends on #949 merging first.

server/pbench/bin/unittests
```diff
@@ -109,23 +109,31 @@ function _run_allscripts {
     _run pbench-edit-prefixes
 }
 
+function _local_find {
+    # We create our own local find command so that we don't emit the size
+    # information for directories.  This is due to the fact that on different
+    # file systems an empty directories, or directories with small numbers of
```
ndokos: "an empty" --> "empty"

server/pbench/bin/unittests
```diff
@@ -109,23 +109,31 @@ function _run_allscripts {
...
+    # files can be handled differently.  E.g. on Ext4 directories have a
+    # minimum size of 4096, while on XFS only after a size do they grow to
```
ndokos: "after a size" ? Not sure what that means.

PR #953, opened 2018-10-12
Stop creating the unnecessary log directory

portante:
We remove the unused and unnecessary log directory creation step since it ends up creating thousands of directories, 1,440 to exact, every day.  This will eventually reek havoc with a file system.

### DIFF
--- a/server/pbench/bin/pbench-sync-package-tarballs
+++ b/server/pbench/bin/pbench-sync-package-tarballs
@@ -57,13 +57,6 @@ mkdir -p "$tmp"
 > $index_content
 
 log_init $(basename $0) $LOGSDIR/$(basename $0)
-logdir=$LOGSDIR/$(basename $0)/$prefix/$TS
-mkdir -p "$logdir"
-rc=$?
-if [[ $rc != 0 ]] ;then
-    echo "$TS: Failed: mkdir -p $logdir" >&4
-    exit 2
-fi
 
 calculate_prefixname () {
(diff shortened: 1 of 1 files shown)


server/pbench/bin/pbench-sync-package-tarballs
```diff
@@ -128,7 +121,7 @@ $PROG: Processed $nprocessed result tar balls, with $nerrs errors
 
 EOF
     cat $tmp/mail.log >> $index_content
-    pbench-report-status --name $PROG --timestamp $TS --type status $index_content
+    $dir/pbench-report-status --name $PROG --timestamp $TS --type status $index_content
```
tenstormavi: Do we need `$dir` for specifically this script only or every other server scripts?

portante: In the context of when `pbench-sync-package-tarballs` is run (it is invoked remotely by `pbench-sync-satellite` when that cron job runs), unless the `pbench` user has `/opt/pbench-server/bin` in its path, `pbench-report-status` won't be found.  Notice that at the beginning of the script, `$dir` is first calculated in order to invoke `pbench-base.sh`.

The full fix on the master branch should arguably be to have `pbench-base.sh` update `PATH` with the location of the server execution environment.  But for this v0.53 branch, it seems sufficient to just reference `$dir/pbench-report-status` as a temporary fix until we finalize v0.54.

Issue #956, opened 2018-10-16
replace fiologparser_hist.py


bengland2:
fiologparser_hist.py is not fast enough to generate cluster-wide latency percentiles as a function of time.  We are planning to replace fiologparser_hist.py with fio-histo-log-pctiles.py (in fio 3.11) to solve this.  This requires some small tweaks to pbench-fio and postprocessing.

portante:
Fixed by #958 / #964.

PR #958, opened 2018-10-16
Fio histo log pctiles

bengland2:
This PR requires that a changed version of fio-histo-log-pctiles.py be used in the pbench-fio RPM.  I'm trying to [get that into upstream fio](https://github.com/axboe/fio/pull/706) now.   It removes pandas dependency and computes cluster-wide response time percentiles over time.   See per-commit comments (click on ...) for details.  I tested it with a 3-VM Ceph cluster on my laptop.


agent/bench-scripts/postprocess/fio-postprocess-viz.py
```diff
@@ -41,12 +41,17 @@ def main(ctx):
     out_files[i].write("#LABEL:%s\n" % columns[i])
 
   with open(join(ctx.DIR, 'hist.csv'), 'r') as csv:
-    csv.readline()
+    while True:
```
portante: Can we turn this into a loop that does not use `while True:`?  Something like:

```
    line = csv.readline()
    while line is not None and not line.startswith('msec'):
        line.csv.readline()
    if line is None:
        print('ERROR: hit end of file without seeing header')
        sys.exit(1)
    # NOTE that "line" contains the header which we ignore
```

agent/bench-scripts/postprocess/fio-postprocess-viz.py
```diff
@@ -41,12 +41,17 @@ def main(ctx):
...
+      l = csv.readline()
+      if l == None:
+          print('ERROR: hit end of file without seeing header')
+          sys.exit(1)
+      if l.startswith('msec'):
```
portante: Since we changed from milliseconds to seconds, why is the header still using `msec`?

agent/bench-scripts/pbench-fio
```diff
@@ -68,9 +68,9 @@ tool_label_pattern="fio-"
 tool_group="default"
 max_key_length=20
 primary_metric="readwrite_IOPS"
-histogram_interval_msec=$(getconf.py histogram-interval-msec pbench-fio)
```
portante: For compatibility, should we do something like:

```
# Fetch the new configuration value in seconds as a string.
histogram_interval_sec_str=$(getconf.py histogram-interval-sec pbench-fio)
if [ -z "$histogram_interval_sec_str" ] ;then
        # Fall back to old configuration value in milliseconds, and convert to seconds.
        histogram_interval_msec_str=$(getconf.py histogram-interval-msec pbench-fio)
        if [ -z "$histogram_interval_msec_str" ] ;then
                let histogram_interval_msec=10000
        else
                # Explicitly convert from string to a number
                let histogram_interval_msec=${histogram_interval_msec_str}*1
        fi
        let histogram_interval_sec=${histogram_interval_msec}/1000
else
        # Explicitly convert from string to a number
        let histogram_interval_sec=${histogram_interval_msec_str}*1
fi
```

agent/bench-scripts/pbench-fio
```diff
@@ -157,15 +157,15 @@ function fio_usage() {
 		printf -- "\t--remote-only\n"
 		printf "\t\trun this on the remotes only\n"
 		printf "\n"
-		printf -- "\t--histogram-interval-msec=<int>\n"
-		printf "\t\tset the histogram logging interval in milliseconds (default $histogram_interval_msec)\n"
+		printf -- "\t--histogram-interval-sec=<int>\n"
```
portante: Can we keep the old parameter option for compatibility?

agent/bench-scripts/pbench-fio
bengland2: we can't really fall back without changing fio-histo-log-pctiles.py because its --time-quantum parameter is in units of seconds, so it doesn't make sense to use milliseconds anymore.  I thought we all agreed (or at least Andrew did) to go with seconds for this parameter because no one was interested in sub-second histogram logs.

agent/bench-scripts/postprocess/fio-postprocess-viz.py
bengland2: good point, but it doesn't really matter right?  Andrew said fio-postprocess-viz.py is going away to be replaced by browbench right?

agent/bench-scripts/pbench-fio
bengland2: fio-histo-log-pctiles.py doesn't support sub-second time quantum, Andrew seemed to think it was ok not to go below 1 second.  I don't think a lot of people were using histogram logging up to now.  

I could keep the old parameter name and just convert it into seconds, and not document it (deprecate it).  Is that good enough.

agent/bench-scripts/pbench-fio
portante: Is the proposal above to fall back to milliseconds or to seconds?  The above appears to propose the final result is in seconds, no?

agent/bench-scripts/pbench-fio
portante: Yes, keeping the old parameter name and just convert it into seconds, and not documenting it (deprecating it), seems fine.

portante:
@bengland2, please consider merging https://github.com/bengland2/pbench/pull/2 into your own local tree and then updating this PR with those changes, squashing them all down to one commit.

portante:
@bengland2, please consider merging https://github.com/bengland2/pbench/pull/2 into your own local tree and then updating this PR with those changes, squashing them all down to one commit.

portante:
A test

portante:
Why?

portante:
Well not quite, was looking for a rebase and merge, sec...

portante:
In the future, let's be sure we get folks to properly review and sign off before one merges anything.  Notice that there was still a review rejection on.  Thanks.

portante:
Note the revert commit was never merged, decided to fix things via PR #964 instead.

Fixes #956.

bengland2:
thanks again Peter.

PR #959, opened 2018-10-16
Index uperf result data properly

ndokos:
WARNING: This is not finished - do not merge!

This commit concerns itself only with the lowest level result.json files (those in
the <iteration>/sample-NN/ directories.

There are problems with timestamps: the timestamp in each document needs to be
converted (I think). Also, the index needs to use the experiment timestamp from
the run doc, not the timestamp from an individual data point.

The unit tests are failing (starting with 7.6). The changes have been tested only in
a small harness outside the main index-pbench script.


server/pbench/bin/index-pbench
```diff
@@ -322,6 +330,168 @@ class ResultData(object):
...
+            if latency:
+                dat = {
+                    "latency_desc": data[i][0],
+                    "throughput_desc": data[i][1]
+                }
+            elif throughput:
+                dat = {"throughput_desc": data[i][0]}
+            else:
+                # impossible?
+                return
```
portante: Can we raise an exception instead?

server/pbench/bin/index-pbench
```diff
@@ -322,6 +330,168 @@ class ResultData(object):
...
+                    # They should be in sync.
+                    lat, xput = res
+                    assert lat['date'] == xput['date']
+                    d.update({"latency_value": lat['value'], "throughput_value": xput['value']})
+                elif throughput:
+                    xput = res[0]
+                    d.update({"throughput_value": xput['value']})
+                else:
+                    # impossible
+                    return
```
portante: Can we raise an exception instead?

server/pbench/bin/index-pbench
```diff
@@ -322,6 +330,168 @@ class ResultData(object):
...
+            yield datum
+        return
+
+    @staticmethod
+    def gen_sources(results):
+        # There should always be a 'parameters' block.
+        # There is no reason for this to be an array, but it is: we
+        # just pick the 0^th (and only element) as the metadata for
+        # the document.
+        md = results['parameters']['benchmark'][0]
```
portante: Can we wrap this in a `try` block to catch badly formed results?

server/pbench/bin/index-pbench
```diff
@@ -287,15 +291,19 @@ class ResultData(object):
                 self.counters['not_valid_json_file'] += 1
                 continue
 
-            convert_to_float(results)
+            # for each file, generate a sequence of documents and yield each one
+            for md, ts, src in gen_sources(results):
```
portante: Do we need the class name here to reference `gen_sources` as a static method?

server/pbench/bin/index-pbench
```diff
@@ -322,6 +330,168 @@ class ResultData(object):
...
+                tseries.append(ts)
+                convert_to_float(tseries)
+                del results['throughput']['Gb_sec'][i]['timeseries']
+                ds = (results['throughput']['Gb_sec'][i],)
+                data.append(ds)
+            # print("data (len, type) = ", len(data), type(data))
+            # pp.pprint(data)
+        else:
+            # impossible?
+            return
```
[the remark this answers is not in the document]
ndokos: And catch it somewhere so we can continue? Or bail out completely? Error handling is certainly lacking.

server/pbench/bin/index-pbench
ndokos: I'll take a stab at it.

server/pbench/bin/index-pbench
ndokos: I don't know if we need to (I have to do some testing of index-pbench with this incorporated and see if it fails), but it's probably a good idea for code readability.

ndokos:
@portante: PTAL

server/pbench/bin/index-pbench
ndokos: And, yes, we *do* need to qualify the name.

server/pbench/bin/index-pbench
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/python3 -u
```
portante: So this forces unbuffered mode, right?  But is that necessary for the final commit?

server/pbench/bin/index-pbench
```diff
@@ -321,6 +368,115 @@ class ResultData(object):
...
+                        datum = datum_wrapper
+                        datum.update({"datum_value": res['value']})
+                        yield res['date'], datum
+
+        return
+
+    # XXX - copied from ToolData - DRY?
+    def convert_timestamp(self, orig_ts):
+        # Convert the given string timestamp, assumed to be a float in
+        # milliseconds since the epoch, to the expected string timestamp format.
```
portante: Can you add to this comment that it is assumed the value is already in the UTC timezone?

server/pbench/bin/index-pbench
```diff
@@ -287,20 +315,39 @@ class ResultData(object):
...
     def make_source(self):
-        """Simple jump method to pick the write source generator based on the
-        handler's prospectus."""
+        """
+        If we ever decide we need to process anything other than JSON files, this
+        is going to become a jump method, similar to ToolData.make_source().
+        ATM, we only handle JSON files.
+        """
         if not self.json_files:
             # If we do not have any json files for this experiment, ignore it.
```
portante: s/experiment/result/?

server/pbench/bin/index-pbench
```diff
@@ -287,20 +315,39 @@ class ResultData(object):
                 self.counters['not_valid_json_file'] += 1
                 continue
 
-            convert_to_float(results)
+            # There should always be a 'parameters' block.
+            # There is no reason for this to be an array, but it is: we
+            # just pick the 0^th (and only element) as the metadata for
+            # the document.
+            assert len(results['parameters']['benchmark'])
+            md = results['parameters']['benchmark'][0]
```
portante: Can we use a `try/except` here instead of an assertion?

server/pbench/bin/index-pbench
```diff
@@ -287,20 +315,39 @@ class ResultData(object):
...
+                       "iteration_number": iter_number,
+                       "sample": sample})
+            benchmark = md['benchmark_name']
+            # for each file, generate a sequence of documents and yield each one
+            for ts, src in ResultData.gen_sample_sources(results, benchmark):
+                # each generated result is a document, accompanied by a timestamp
+                # and some metadata.
+                source = _dict_const()
+                source['@metadata'] = self.run_metadata
+                source['@metadata'].update(md)
```
portante: I believe this `.update(md)` also updates `self.run_metadata`.  Do we want that?

server/pbench/bin/index-pbench
```diff
@@ -254,6 +258,23 @@ class ResultData(object):
             ("experiment", experiment)
         ])
         self.ptb = ptb
+
+        # XXX - copied from ToolData - need DRY
```
portante: +1

server/pbench/bin/index-pbench
```diff
@@ -15,6 +15,9 @@ from datetime import datetime, timedelta
 from collections import Counter
 from argparse import ArgumentParser
 
+import pprint
```
portante: Can we use `json.dump()` instead?

server/pbench/bin/index-pbench
```diff
@@ -321,6 +368,115 @@ class ResultData(object):
...
+            # for each client/server pair (or client?) used in the test.
+            if benchmark == "uperf":
+                latency_sublist = ['usec']
+            elif benchmark == "fio":
+                latency_sublist = ['clat', 'lat', 'slat']
+            else:
+                return
+            for sublist in latency_sublist:
+                latency_list = results['latency'][sublist]
+                for i in range(len(latency_list)):
```
portante: Can we consider using `for idx,latency_el in enumerate(latency_list):` instead of `range(len())`?

server/pbench/bin/index-pbench
```diff
@@ -24,7 +27,7 @@ try:
 except ImportError:
     from elasticsearch import VERSION as es_VERSION, Elasticsearch
 
-from pbench import tstos, es_index, es_put_template
```
portante: Is there a reason why need to make this change now?  If so, can we do it in a separate PR from this one?

server/pbench/bin/index-pbench
```diff
@@ -321,6 +368,115 @@ class ResultData(object):
             datafiles.append(datafile)
         return datafiles
 
+    @staticmethod
+    def expand_template(templ, d):
```
portante: Is this going to fleshed out, or is it just a place holder for expanding the "uid" template field?

server/pbench/bin/index-pbench
```diff
@@ -321,6 +368,115 @@ class ResultData(object):
...
+        return templ
+
+    @staticmethod
+    def make_datum_wrapper(templ):
+        # make a shallow copy
+        d = templ.copy()
+        del d['timeseries']
+        # construct the uid from the template and the values in d
+        # and replace the template with the result
+        d['uid'] = ResultData.expand_template(d['uid'], d)
```
portante: Is this going to be an MD5 sum of the timeseries?  Not sure what this logic is about.

server/pbench/bin/index-pbench
```diff
@@ -321,6 +368,115 @@ class ResultData(object):
...
+        ts = datetime.utcfromtimestamp(ts_float)
+        if self.start_run_ts is not None and ts < self.start_run_ts:
+            self.counters['result_ts_before_start_run_ts'] += 1
+        if self.end_run_ts is not None and ts > self.end_run_ts:
+            self.counters['result_ts_after_end_run_ts'] += 1
+        return ts.strftime("%Y-%m-%dT%H:%M:%S.%f%z")
+
+
+
+def mk_result_data_actions(ptb, options, idx_prefix, idx_patterns, opctx):
```
portante: Can we make this move of the location of this function in a separate commit?  I think you are moving it and changing it, but I'd like to be sure.

portante:
A test

ndokos:
@portante: I added a couple of commits to address the (non)-uniqueness of the source id and also to expand the template (currently, without any error checking). I have not addressed any of the comments in your lates review yet.

server/pbench/bin/index-pbench
ndokos: Instead of pprint? If so, the plan is to get rid of pprint altogether. It's only used for debugging.

server/pbench/bin/index-pbench
ndokos: Yes,

server/pbench/bin/index-pbench
ndokos: Sure.

server/pbench/bin/index-pbench
ndokos: Sure.

server/pbench/bin/index-pbench
ndokos: Fleshed out a bit (but without error checking) in commit e8f89dc.

server/pbench/bin/index-pbench
ndokos: OK

server/pbench/bin/index-pbench
ndokos: We use "experiment" in a number of places and I tend to prefer it here:  I think of it as "we run an experiment and we get results, but this experiment produced no results (at least no result.json files)." If you prefer it, I can change it, but please take a look at the other "experiment" instances and see if we need to change it in more places than just here.

server/pbench/bin/index-pbench
ndokos: Absolutely not. I'll check and fix it.

server/pbench/bin/index-pbench
ndokos: The `uid` element is going to be the expanded template, replacing `%foo%` in the template with the value of `foo` in the dict `d`. The MD5 sum over the whole doc is going to be the source id of the document. See commit  [ 58d8abf ](https://github.com/distributed-system-analysis/pbench/pull/959/commits/58d8abf45fe6bdaffb8f3dd261861866ccd2f3c5) for the source id mod and commit [e8f89dc](https://github.com/distributed-system-analysis/pbench/pull/959/commits/e8f89dc84e170ad6652f6980ae14df55d64a4666) for the template expansion mod.

server/pbench/bin/index-pbench
ndokos: Yes,

ndokos:
@portante: I hope this is OK. PTAL and let me know.

portante:
@ndokos, did you intend to check in the `server/pbench/bin/foo` file?

ndokos:
@portante: still need to add a couple of small unit tests (uperf latency and fio), but test-7.15 is the model for them: uperf throughput test with 5 data points kept, so the gold output includes everything that would be sent to ES. PTAL.

server/pbench/bin/index-pbench
```diff
@@ -321,6 +383,166 @@ class ResultData(object):
...
+                    tseries = xput_el['timeseries']
+                    for res in tseries:
+                        datum = datum_wrapper
+                        datum.update({"value": res['value']})
+                        src_id = _make_source_id(datum)
+                        yield res['date'], datum, src_id
+
+        return
+
+    # XXX - copied from ToolData - DRY?
```
portante: +1 to following DRY principle.

server/pbench/bin/unittests
```diff
@@ -381,7 +381,18 @@ declare -A cmds=(
...
+    # 7.2.X: UnsupportedTarballFormat exceptions:
+    # missing metadata.log
+    [test-7.2.0]="_run_index"
+
+    # badly formed tarball: ./<resultdir> prefix instead of <resultdir>
+    [test-7.2.1]="_run_index"
+
+    # "impossible" - extract tarball into tmpdir but resulting dir does not exist -
+    # will probably require sleight-of-hand to trigger.
+    # [test-7.2.2]="_run_index"
```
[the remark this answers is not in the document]
ndokos: I might if I can find a simple way to trigger it. Otherwise, I'll open an issue.

server/pbench/bin/unittests
```diff
@@ -401,6 +412,16 @@ declare -A cmds=(
...
+    # truncated uperf results data so we can do
+    # a complete comparison of input to output.
+    # uperf throughput (Gb_sec) results
+    [test-7.15]="_run_index"
+
+    # truncated uperf results data so we can do
+    # a complete comparison of input to output.
+    # uperf latency and throughput (usec and trans_sec) results
+    # TBD
+
```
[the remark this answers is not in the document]
ndokos: I'm going to add this test.

server/pbench/bin/index-pbench
ndokos: Did as you suggested to eliminate the duplication with `gen_run_timestamps()`. Also, factored out the `convert_timestamp()` method into an abstract object that both `ResultData` and `ToolData` inherit from, to eliminate *that* duplication.

server/pbench/bin/index-pbench
```diff
@@ -321,6 +383,161 @@ class ResultData(object):
...
         return datafiles
 
+    @staticmethod
+    def expand_template(templ, d):
+        import re
+
+        e = re.compile("%\w*?%")
+        s = templ
+        for m in re.findall(e, s):
+            s = re.sub(e, d[m[1:-1]], s)
```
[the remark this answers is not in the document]
ndokos: Added comments.

ndokos:
@portante: ptal, particularly at the additions to the wrapper. Exception handling reorg has *not* been done.

server/pbench/bin/index-pbench
```diff
@@ -245,15 +270,18 @@ def convert_to_float(d):
     else:
         pass
 
+class ResultData(PbenchData):
 
-class ResultData(object):
     def __init__(self, ptb, experiment, opctx):
```
portante: Can you use `super` here?  Something like:

```
def __init__(self, ptb, experiment, opctx):
    start, end = ptb.gen_run_timestamps()
    super(ResultData, self).__init__(start, end)
    self.run_metadata = _dict_const([
            ("runtstamp", ptb.start_run),
            ("runid", ptb.md5sum),
            ("experiment", experiment)
        ])
    self.ptb = ptb
    self.counters = Counter()
    opctx.append(_dict_const(object="ResultData", counters=self.counters))
    self.json_files = None
```

server/pbench/bin/index-pbench
```diff
@@ -245,15 +270,18 @@ def convert_to_float(d):
...
         self.run_metadata = _dict_const([
             ("runtstamp", ptb.start_run),
             ("runid", ptb.md5sum),
             ("experiment", experiment)
         ])
         self.ptb = ptb
+
+        self.start_run_ts, self.end_run_ts = ptb.gen_run_timestamps()
```
portante: Can you consider removing this in favor of using the super class init method?

server/pbench/bin/index-pbench
```diff
@@ -602,20 +820,9 @@ class ToolData(object):
...
-                self.start_run_ts = None
-        try:
-            self.end_run_ts = datetime.strptime(ptb.end_run, "%Y-%m-%dT%H:%M:%S.%f")
-        except ValueError:
-            try:
-                self.end_run_ts = datetime.strptime(ptb.end_run, "%Y-%m-%dT%H:%M:%S")
-            except ValueError:
-                self.end_run_ts = None
+
+        self.start_run_ts, self.end_run_ts = ptb.gen_run_timestamps()
```
portante: Can you use, `super(ToolData, self).__init__(start, end)` here as well?

server/pbench/bin/index-pbench
```diff
@@ -225,6 +227,29 @@ def es_template(es, options, INDEX_PREFIX, INDEX_VERSION, config, dbg=0):
         sys.stdout.flush()
 
 
+
+###########################################################################
+# Pbench Data abstract class - ToolData and ResultData inherit from it.
+
+class PbenchData(object):
+
+    def __init(self):
```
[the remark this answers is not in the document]
ndokos: I can do this, but see below...

server/pbench/bin/index-pbench
ndokos: Replace with

    super(ResultData, self).__init__(ptb.gen_run_timestamps())

? Is that what you mean?

ndokos: The above does not work. I'm not sure what you want me to do here.

ndokos: I misspelled the constructor in the superclass :-(

portante: @ndokos, here is a PR that shows it can work: https://github.com/ndokos/pbench/pull/8

server/pbench/bin/index-pbench
ndokos: Done

server/pbench/bin/index-pbench
ndokos: Done.

server/pbench/bin/index-pbench
ndokos: Done.

ndokos:
@portante: I think this is ready to squash down before pulling it in. Let me know if you see problems.

portante:
FWIW, was able to run all unit tests successfully on a RHEL 7.5 box.

PR #964, opened 2018-10-24
Fix failing tests from PR #958

portante:


portante:
Finalizes fix to #956.

PR #946, opened 2018-10-05
Enhance pbench server unit tests

portante:
A series of commits to enhance the server unit tests now that server activation and auditing are used.

PR #965, opened 2018-10-24
Server infrastructure enhancements

portante:
These 15 commits represent an attempt to clean up the server code base ahead of a major restructuring to support enhanced unit tests in PR #946.

The goal is to keep these as individual commits, rebased together when merged, so that we can bisect the changes easier if we have to debug a problem.


portante:
Rebased to pick up unit test fixes, please review again @tenstormavi, thanks!

Issue #967, opened 2018-10-24
Perl syntax error in fio-postprocess with newer versions of Perl


k-rister:
Newer versions of Perl do not allow the following syntax:

keys $$job{"job options"}  -- [Code Link](https://github.com/distributed-system-analysis/pbench/blob/master/agent/bench-scripts/postprocess/fio-postprocess#L189) -- [Source Patch](https://github.com/distributed-system-analysis/pbench/commit/e2717ec6e065785b84ed72d7ad5cc9935ab0ae25#diff-158241170fcbc8bd5f6935890df06a30R189)

Which results in:

Experimental keys on scalar is now forbidden at /opt/pbench-agent/bench-scripts/postprocess/fio-postprocess line 189

It needs to be written like this:

keys (%{ $$job{"job options"} })

PR #968, opened 2018-10-24
fio-postprocess: fix Perl syntax bug

k-rister:
- In newer versions of Perl the existing syntax results in the
  following error:

  "Experimental keys on scalar is now forbidden at <file> line <line #>"


k-rister:
Addresses #967

PR #970, opened 2018-10-25
Update list of perl versions we test against

portante:
This is to try to avoid issues like #967.


k-rister:
Can we do something else to address the Travis failures that occurred here?  It seems like we frequently have CI builds fail due to package installation issues that have nothing to do with out code.  While it would be nice to not have to deal with this at all at a minimum it would be nice to detect this type of problem and automatically restart the job (which I just manually did).

k-rister:
The restarted failed again.  Upon closer inspection I'm not sure it will ever pass, it seems to imply that the package that is being requested for install does not exist.

portante:
Retries now pass with flying colors.

k-rister:
Well that is kind of my point, right?  What can we do to make this more reliable?  As is it seems very hit or miss which seems completely counter to it's intent.

ndokos:
Maybe [travis_retry](https://docs.travis-ci.com/user/common-build-problems/#my-builds-are-timing-out) would help with the dependency installation failures.

PR #971, opened 2018-10-31
Remove trailing whitespace from files

portante:
Git periodically complains about trailing white space during merges and
rebase operations.  We used `sed -i 's/[ \t]*$//' "$1"` to run through
the tree and remove the whitespace, and then fixed up `index-pbench` and
`pbench-gen-json-payload` so that they don't generate output that
trailing whitespace.


agent/tool-scripts/postprocess/samples/fio-0/result/reference-result/tools-default/localhost/sar/sar-stdout.txt
```diff
@@ -619,10 +619,10 @@ Linux 4.10.16-100.fc24.x86_64 (alphaville.usersys.redhat.com) 	06/14/17 	_x86_64
 
 13:51:14        BUS  idvendor    idprod  maxpower                manufact                                         product
 13:51:17          3       46d      c52b       196                Logitech                                    USB Receiver
-13:51:17          2      8087      8000         0                                                                        
-13:51:17          4      17ef      1010         0   LENOVO                                        Lenovo ThinkPad Dock   
-13:51:17          1      8087      8008         0                                                                        
-13:51:17          3      17ef      1010         0   LENOVO                                        Lenovo ThinkPad Dock   
+13:51:17          2      8087      8000         0
```
ndokos: Not sure we should do this: this is what sar stdout looked like.

portante:
Not worth fixing up at this point.

PR #972, opened 2018-11-01
Use pbench-index instead of index-pbench

portante:
We drop the `_run_index` method in the unit tests in favor of a new
`debug_unittest` `pbench-index.cfg` file variable to govern the way
unit test mode is invoked in `index-pbench`.  We in turn drop the `-U`
`index-pbench` switch.

We also pass long `pbench-index` arguments to `index-pbench` so that
we can exercise the `-I` switch for dumping the index template names.

All of the existing `test-7.*` state tar balls needed to be updated
to place the tar balls in the archive directory hierarchy properly,
adding state for `test-7.0` so that `index-pbench` will be invoked.

Finally, we trimmed down the size of the state file for test-7.16 by
eliminating the exploded tar ball directory, a `foo.*` file, and a
very large `pbench.log` file.


server/pbench/bin/unittests
```diff
@@ -378,45 +378,49 @@ declare -A cmds=(
...
     [test-6.6]="_run pbench-backup-tarballs"
 
-    # indexing tests
-    [test-7.0]="_run_index -I"
-    [test-7.1]="_run_index"
+    # indexing tests are all test-7.*
+
+    # Special case of running pbench-index with -I to get a dump
+    # of the index patterns in use.
+    [test-7.0]="_run pbench-index -I"
```
ndokos: Should the `_run_index` function be deleted then?

PR #983, opened 2018-11-08
Limit the length of ES error messages

ndokos:
This is just to prevent a large error message from bloating the log
file. It does not solve the underlying problem which is that we sent
a payload that was bigger than the max that ES was willing to accept.

### DIFF
--- a/server/pbench/lib/pbench/__init__.py
+++ b/server/pbench/lib/pbench/__init__.py
@@ -19,6 +19,7 @@ def tstos(ts=None):
 
 _r = SystemRandom()
 _MAX_SLEEP_TIME = 120
+_MAX_ERRMSG_LENGTH = 16384
 
 def calc_backoff_sleep(backoff):
     global _r
@@ -227,8 +228,12 @@ def es_index(es, actions, errorsfp, dbg=0):
             status = resp['status']
         except KeyError as e:
             assert not ok
-            # resp is not of expected form
-            print(resp)
+            # resp is not of expected form.
+            # set it to the complete payload, so that
+            # it can be reported below.
+            resp = resp_payload
+            # Limit the length of the error message.
+            print("es_index: ERROR %r" % (e), file=sys.stderr)
             status = 999
         else:
             assert action['_id'] == resp['_id']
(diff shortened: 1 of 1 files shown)


ndokos:
This needs a unit test before it can be merged.

server/pbench/lib/pbench/__init__.py
```diff
@@ -227,8 +228,9 @@ def actions_tracking_closure(cl_actions):
             status = resp['status']
         except KeyError as e:
             assert not ok
-            # resp is not of expected form
-            print(resp)
+            # resp is not of expected form.
+            # Limit the error message to 16K
+            print(resp[:_MAX_ERRMSG_LENGTH])
```
portante: Do we have to first convert `resp` into a string to do this?  Perhaps we should use `json.dumps` instead?

And not sure we should put the size we are limiting the error message to in the comments.

ndokos: Done in latest.

server/pbench/lib/pbench/__init__.py
```diff
@@ -227,8 +228,10 @@ def actions_tracking_closure(cl_actions):
             status = resp['status']
         except KeyError as e:
             assert not ok
-            # resp is not of expected form
-            print(resp)
+            # resp is not of expected form.
+            # Limit the length of the error message.
+            msg = json.dumps(resp, indent=4)
+            print(msg[:_MAX_ERRMSG_LENGTH])
```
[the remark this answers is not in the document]
ndokos: I'll do that, but just for my benefit: are you worried about a full string copy? If so, I'm pretty sure that that does not happen. It is just a ref.

server/pbench/lib/pbench/__init__.py
```diff
@@ -248,8 +251,10 @@ def actions_tracking_closure(cl_actions):
                 errorsfp.flush()
                 failures += 1
             else:
-                # Retry all other errors
-                print(resp)
+                # Retry all other errors.
+                # Limit the length of the error message.
+                msg = json.dumps(resp, indent=4)
+                print(msg[:_MAX_ERRMSG_LENGTH])
```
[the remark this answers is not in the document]
ndokos: Sure,

ndokos:
@portante: did the changes and squashed, I think it's ready.

server/pbench/lib/pbench/__init__.py
```diff
@@ -227,8 +228,9 @@ def actions_tracking_closure(cl_actions):
             status = resp['status']
         except KeyError as e:
             assert not ok
-            # resp is not of expected form
-            print(resp)
+            # resp is not of expected form.
+            # Limit the length of the error message.
+            print("es_index: WARNING - retrying error %r\n%s" % (e, json.dumps(resp, indent=4)[:_MAX_ERRMSG_LENGTH]), file=sys.stderr)
```
portante: Do we retry these errors?

PR #986, opened 2018-11-09
Extend auditing to incoming, results, and users directory hierarchies

portante:
The server audit is now applied to the incoming, results, and users directory hierarchies.  Any unpacked tar ball should now be comprehensively checked to see that all is in the correct place.

The `test-20` unit test gold file holds an example of an audit report covering all the possible outputs it can emit. Each unit test runs the report as well, and they have been updated accordingly.


portante:
Almost ready for review, but not quite.  I still need to update `test-20` to provide error cases that will hit all the error reporting conditions.

portante:
@ndokos, @tenstormavi, this seems ready for review now.

As you'll read in the updated commit message, `test-20`'s gold file has the output of a report that shows what it covers in an audit.

portante:
Rebased, will merge once the CI tests pass.

portante:
@tenstormavi, @ndokos, okay, I had to rebase on top of the syslog change that was merged, and found a problem with `pbench-report-status` which I fixed in a new commit as part of this PR.  Please review again when you have a chance.

portante:
Okay, I made the changes to the comments, and then added two more commits to get things working under Fedora 29 (Python 3.7 complains about SafeConfigParser with deprecation warnings in the unit tests, and the F29 PATH variable is different causing test-8 to fail).

PR #1011, opened 2018-12-18
pbench-report-status: fix md5 calculation.

ndokos:
Calculate the md5 on the generated JSON document which includes the
timestamp.

Previously, we were calculating it on the input document which did not
include the timestamp: the result was that the md5 could be identical
between different documents (e.g. the same error message generated
during different runs of a script): since the md5 is used as the ES id
for the document, the indexing of the document could fail as a
duplicate.


portante:
This will be the first PR to 0.55.  Then we can backport to 0.54 branch with a follow-on PR.

Issue #1009, opened 2018-12-15
Consider using syslog instead of email as the fallback when ES is not available


portante:
Satellites generally set use-mail-for-status to true since they don't have access to an ES instance generally.  But since we made all the scripts call `pbench-report-status` unconditionally, on satellites configured for mail they send half a dozen emails every minute.

Oops.

portante:
Addressed by #1014.

PR #1014, opened 2018-12-20
Use syslog instead of email when an Elasticsearch instance is not available

portante:
Addresses #1009.

The first commit is large because we sort the JSON payload keys to avoid unit test failures when the order changes.

The second commit implements the change to use syslog via the `logger` command.


server/pbench/bin/pbench-report-status
```diff
@@ -86,36 +75,52 @@ rm -rf $tmp" EXIT QUIT INT
...
 
 server=$(getconf.py --config $IDXCONFIG server Server)
-if [[ -z "$server" ]] ;then
-    echo "$prog: no server defined in [Server] section of $IDXCONFIG" >> ${tmp}/log
-    exit 2
-fi
+index_prefix=$(getconf.py --config $IDXCONFIG index_prefix Settings)
+if [ -z "$server" -o -z "$index_prefix" ] ;then
+    # We don't have an Elasticsearch configuration, use syslog.
+    if [ -z "$server" -o -z "$index_prefix" ] ;then
```
tenstormavi: Seems like this should be `if [ -z "$server" ]`

server/pbench/bin/pbench-report-status
```diff
@@ -86,36 +75,52 @@ rm -rf $tmp" EXIT QUIT INT
...
+        echo "$prog: no server defined in [Server] section of $IDXCONFIG" >> ${tmp}/log
+    fi
+    if [ -z "$server" -o -z "$index_prefix" ] ;then
```
tenstormavi: And this should be `if [ -z "$index_prefix" ] `

server/pbench/bin/pbench-report-status
```diff
@@ -86,36 +75,52 @@ rm -rf $tmp" EXIT QUIT INT
...
+        xz -9 -c ${tmp}/payload > $LOGSDIR/$name/${prog}-payload.${timestamp%-*}.xz
+    else
+        cat ${tmp}/payload | logger --size 4096 --tag $name --priority daemon.info
+    fi
+else
+    # We have an Elasticsearch configuration.
+
+    # Snip off the trailing "-<TZ>" - assumes that <TZ> does not contain a "-"
+    datestamp="${timestamp%-*}"
+    # Snip off the time part of the timestamp to get YYYY-MM
```
ndokos: This threw me: maybe add a comment documenting the format of the timestamp? Also change the comment to say "... the day and time part..." perhaps.

server/pbench/bin/pbench-report-status
```diff
@@ -86,36 +75,52 @@ rm -rf $tmp" EXIT QUIT INT
...
+    index="$index_prefix"."$name"."$datestamp"
+    >${tmp}/http_code
+    >${tmp}/response_body
+    curl --silent --show-error -XPOST -H 'Content-Type: application/json' --output ${tmp}/response_body --write-out "%{http_code}" "http://$server/$index/$doctype/$md5" --data @"${tmp}/payload" > ${tmp}/http_code 2>> ${tmp}/log
+    status=$?
+    if [ ${status} -ne 0 -o "$(cat ${tmp}/http_code)" != "201" ]; then
+        echo "${prog}: curl failed (exit code: \"${status}\", HTTP code: \"$(cat ${tmp}/http_code 2> /dev/null)\")" >> ${tmp}/log
+        if [ -s ${tmp}/response_body ]; then
+            echo "Response body:" >> ${tmp}/log
+            cat ${tmp}/response_body >> ${tmp}/log 2> /dev/null && echo >> ${tmp}/log
```
ndokos: What happens to $tmp/log? Are we just throwing it away?

server/pbench/bin/pbench-report-status
portante: Thanks, nice catch.  I'll fix this.

server/pbench/bin/pbench-report-status
portante: Thanks, nice catch.  I'll fix this.

server/pbench/bin/pbench-report-status
portante: Since this is a lift and shift, it should not be a change in behavior.

As before, the contents of the `${tmp}/log` file are added to the caller's log file; see line 70 above.

tenstormavi:
LGTM

portante:
@ndokos, can you review again?

Issue #700, opened 2017-10-20
pidstat: make default collection interval larger.


ndokos:
@ekuric ran into problems when postprocessing pidstat data after a long run: the pidstat pp grew to 32GB of memory and the system started OOM killing things.

- In the short term, increase the default value of the pidstat collection interval to 30s (@ekuric: is that enough for your purposes?)

- In the longer term, investigate the (in)efficiency of pidstat pp and see what can be done.

mrsiano:
@atheurer @ekuric 
is that possible to build the csv file in stages 
something like:
```
my $lastchunk = '';
my $buffer = '';
open(FILE, $filename) or die "Can't open `$filename': $!";
while (sysread FILE, $buffer, 4096) {
         # do the parsing..
         ...
         # do the gen_data in portions.
         gen_data(\%pidstat, \%graph_type, \%graph_threshold, $dir);
}
```

portante:
Seems like `gen_data` needs to take some kind of on-disk intermediate form that each post-processing in the short term to address the memory consumption.  Having it take in a perl variable containing all of the data in memory is not really scalable.

Also, we need to consider NOT doing this kind of post-processing in the agent at all.  Having the agents just collect the data, and any necessary metadata would make the agents much leaner and less memory and CPU intensive.

The post-processing step can be then done server side with all the right context so that it can be handled efficiently.

Having the agent be responsible for this processing means that we have to update the agent to get the data processing right.  If we keep the agents simple, then we can fix data processing problems like this on the server side without having to require updates to the agent.

atheurer:
Server side works if you are certain you always have a server in your infra (which we do, but other may not).  I am not against server side, but for the short term it should be really easy to make gen_data use a lot less memory, most likely by calling it for each graph instead of passing a hash of all the graphs' data.

atheurer:
I will look at the gen-data memory optimization when I am back from PTO next week.

portante:
Don't we always want to keep the agent's dead simple?

So if we have a simple agent that just collects and does not process data, can't we provide to those that don't have a server side solution a way to do what the server would do locally so that we don't make the agent complicated for all?

portante:
[From an email from @mrsiano]

Regarding PR #1015, I've tried to come up with some POC which will help us to improve post-processing response time and space complexity.

[...]

PR #1015, opened 2019-01-02
Fixes: #700 parse pidstat in chunks.

mrsiano:
When scanning large pidstat-stdout.txt file, we might face OOM kill, because we try to load the entire file into pidstat [1].
in order to avoid that, we should work in chunks, and pidstat
should hold just some buffer and not the entire file.

This patch will hold a buffer of 10,000 lines, and will use gen_data  func to dump into a csv format and reset the pidstat each iteration.

this is a short term solution and it might take longer because we do the parsing phase while scanning.
this way we'll make sure we'll not hit any OOM kills.

[1]
https://github.com/distributed-system-analysis/pbench/blob/19fa2ad408214df97fab06da1d03aea020590299/agent/tool-scripts/postprocess/pidstat-postprocess#L31

### DIFF
--- a/agent/tool-scripts/postprocess/GenData.pm
+++ b/agent/tool-scripts/postprocess/GenData.pm
@@ -42,7 +42,7 @@ sub gen_data {
 	my %counts;
 	my $filename;
 	my $timestamp_ms;
-	
+
 	# Generate the CSV file, and don't filter out anything based on a
 	# threshold, and generate the averages for each series.
 	my $csv;
@@ -61,7 +61,7 @@ sub gen_data {
 			#printf "\tchart = $chart\n";
 
 			$filename = $csv . "/" . $htmlpage. "_" .$chart . ".csv";
-			open(TOOL_CSV, ">$filename") || die "$script: could not open $filename\n";
+			open(TOOL_CSV, ">>$filename") || die "$script: could not open $filename\n";
 
 			# first check whether the timestamps in the different series agree: if so,
 			# we'll generate a normal (type-1) csv file:  each row will have the form
@@ -135,7 +135,7 @@ sub gen_data {
 							if ( not defined $stats{$htmlpage}{$chart}{$series_key}{$timestamp_ms} ) {
 								die "$script: timestamp $timestamp_ms missing for ($htmlpage, $chart, $series_key)";
 							}
-							$value = $stats{$htmlpage}{$chart}{$series_key}{$timestamp_ms}; 
+							$value = $stats{$htmlpage}{$chart}{$series_key}{$timestamp_ms};
 							$totals{$chart}{$series_key} += $value;
 							if ($value > $maxval{$htmlpage}{$chart}{$series_key}) {
 								$maxval{$htmlpage}{$chart}{$series_key} = $value;
@@ -175,7 +175,7 @@ sub gen_data {
 					for $series_key (@series_keys) {
 						# this should not happen any more, but paranoid checking doesn't hurt.
 						if ( defined $stats{$htmlpage}{$chart}{$series_key}{$timestamp_ms} ) {
-							$value = $stats{$htmlpage}{$chart}{$series_key}{$timestamp_ms}; 
+							$value = $stats{$htmlpage}{$chart}{$series_key}{$timestamp_ms};
 							$totals{$chart}{$series_key} += $value;
 							if ($value > $maxval{$htmlpage}{$chart}{$series_key}) {
 								$maxval{$htmlpage}{$chart}{$series_key} = $value;
--- a/agent/tool-scripts/postprocess/pidstat-postprocess
+++ b/agent/tool-scripts/postprocess/pidstat-postprocess
@@ -40,6 +40,8 @@ my $line;
 my $line2;
 my $tid;
 my @stats;
+my $lines_limit = 10000;
+my $lines_scanned = 0;
 
 my @virsh_list;
 my @virsh_vcpus;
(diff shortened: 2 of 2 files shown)


mrsiano:
@atheurer @portante PTAL

agent/tool-scripts/postprocess/pidstat-postprocess
```diff
@@ -149,35 +162,36 @@ while ($line = <PIDSTAT_TXT>) {
...
+        				if (! defined($pidstat{$htmlpage}{$graph}{$pid}{$timestamp_ms})) {
+        					$pidstat{$htmlpage}{$graph}{$pid}{$timestamp_ms} = 0;
+        				}
+        			}
+        		}
+        	}
+        }
+
+        # dump data into csv.
+        gen_data(\%pidstat, \%graph_type, \%graph_threshold, $dir);
```
portante: Won't we need to modify `gen_data()` to add the data to existing `.csv`, `.html`, and `.json` files?

agent/tool-scripts/postprocess/pidstat-postprocess
```diff
@@ -149,35 +162,36 @@ while ($line = <PIDSTAT_TXT>) {
...
-foreach  $htmlpage ( keys %pidstat ) {
-	foreach  $graph ( keys %{ $pidstat{$htmlpage} } ) {
-		foreach $pid ( keys %{ $pidstat{$htmlpage}{$graph} } ) {
-			foreach $timestamp_ms ( sort {$a <=> $b} (keys %timestamps ) ) {
-				if (! defined($pidstat{$htmlpage}{$graph}{$pid}{$timestamp_ms})) {
-					$pidstat{$htmlpage}{$graph}{$pid}{$timestamp_ms} = 0;
-				}
-			}
-		}
+    # parse data every 10,000 lines.
```
portante: Seems like the indentation is using spaces instead of tabs for this file.

agent/tool-scripts/postprocess/pidstat-postprocess
```diff
@@ -149,35 +162,36 @@ while ($line = <PIDSTAT_TXT>) {
...
-$graph_threshold{file_io}{io_reads_KB_sec} = 50;
-$graph_threshold{file_io}{io_writes_KB_sec} = 50;
-$graph_threshold{memory_faults}{minor_faults_sec} = 150;
-$graph_threshold{memory_faults}{major_faults_sec} = 150;
-$graph_threshold{context_switches}{voluntary_switches_sec} = 100;
-$graph_threshold{context_switches}{nonvoluntary_switches_sec} = 100;
-$graph_threshold{memory_usage}{virtual_size} = 100;
-$graph_threshold{memory_usage}{resident_set_size} = 100;
-gen_data(\%pidstat, \%graph_type, \%graph_threshold, $dir);
+close(PIDSTAT_TXT);
```
portante: Do you need to deal with any lines that have been read since the last dump and process them here?

agent/tool-scripts/postprocess/pidstat-postprocess
```diff
@@ -149,35 +162,36 @@ while ($line = <PIDSTAT_TXT>) {
...
-			foreach $timestamp_ms ( sort {$a <=> $b} (keys %timestamps ) ) {
-				if (! defined($pidstat{$htmlpage}{$graph}{$pid}{$timestamp_ms})) {
-					$pidstat{$htmlpage}{$graph}{$pid}{$timestamp_ms} = 0;
-				}
-			}
-		}
+    # parse data every 10,000 lines.
+    if ($lines_scanned % $lines_limit == 0) {
+
+        # TODO: this code shold be multithreaded in order to speed up the process.
```
portante: Can you move this comment to an issue instead?

agent/tool-scripts/postprocess/pidstat-postprocess
mrsiano: I also ask my self that question I guess @atheurer can answer that better.
AFAI look at the code looks like it's just append, since the file name was always be the same.
moreover I've did some simple test, and compare this patch with the original one.
and the csv output was with the same amount of lines.

agent/tool-scripts/postprocess/pidstat-postprocess
mrsiano: looks like the we can setup the $graph_threshold just once.

agent/tool-scripts/postprocess/pidstat-postprocess
mrsiano: sure

agent/tool-scripts/postprocess/pidstat-postprocess
portante: It seems this line, https://github.com/distributed-system-analysis/pbench/blob/master/agent/tool-scripts/postprocess/GenData.pm#L64, uses a single greater-than character ("`>`") in the `open()` call which indicates it will overwrite data.  If it used double greater-than characters ("`>>`") that would indicate it is appending.

mrsiano: done

agent/tool-scripts/postprocess/GenData.pm
```diff
@@ -61,7 +61,7 @@ sub gen_data {
 			#printf "\tchart = $chart\n";
 
 			$filename = $csv . "/" . $htmlpage. "_" .$chart . ".csv";
-			open(TOOL_CSV, ">$filename") || die "$script: could not open $filename\n";
+			open(TOOL_CSV, ">>$filename") || die "$script: could not open $filename\n";
```
portante: Yes, this will append to it for sure.

Can we add a parameter to `gen_data()` so that the existing behavior can be preserved and only when requested we append new data?

atheurer: I think that would be fine, but make it the last option in the function call (to append), so that all other existing calls to gen_data don't have to be updated.

Issue #1018, opened 2019-01-03
postprocessing - generate data in parallel manner


mrsiano:
In continuing to #1015 and #700 
as a short term solution, we try to parse data in chunks in order to avoid any oom kills.
chunking data make the pp execution time to be quite longer.
in order to speed up the process we should call the gen_data by another thread in order to support parallelism.

at the same time we should make sure the pidstat collection is synced when calling gen_data.

@atheurer @portante

portante:
Closing this bug because the post-processing step appears to run all the tool post-processing, see #1023.

PR #1023, opened 2019-01-09
A few cleanups to pbench-postprocess-tools

portante:
Based on a review for #1018, refactor a bit to clean up.

First and foremost, the handling of a missing non-default tool group has been fixed so that we emit an error and do nothing for a bad tool group.  This avoids collecting data under a tool group that does not exist.

We use `remote_hostname` everywhere instead of `remote` in certain places.

We added some comments and removed a level of indentation in one place that was not needed.

### DIFF
--- a/agent/util-scripts/pbench-postprocess-tools
+++ b/agent/util-scripts/pbench-postprocess-tools
@@ -39,7 +39,7 @@ if [ $? -ne 0 ]; then
 	printf -- "\t-d str --dir=str, str = a directory where the $script_name\n"
 	printf -- "\t                        will store and process data\n"
 	printf "\n"
-	printf -- "\t-i str --iteration=num, num = a number representing the\n"
+	printf -- "\t-i num --iteration=num, num = a number representing the\n"
 	printf -- "\t                              iteration data was collected for\n"
 	exit 1
 fi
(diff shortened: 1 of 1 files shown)


agent/util-scripts/pbench-postprocess-tools
```diff
@@ -174,30 +165,32 @@ for this_tool_file in `/bin/ls $tool_group_dir`; do
...
-				if [ ! -z "$pids_to_kill" ] ;then
-					echo -n "killing the following screen sessions for $name: "
-					echo "$screens_to_kill"
-					kill $pids_to_kill
-				fi
-			else
-				$pbench_bin/tool-scripts/$name --$action --iteration=$iteration --group=$group --dir=$dir "${tool_opts[@]}" &
-				pids="$pids $!"
+		elif [ "$action" == "kill" ]; then
+			screens_to_kill=`screen -ls | grep "$screen_name" | awk '{print $1}'`
```
mrsiano: screen dose not need any further rpms?! we may want to use nohup ?!

portante: This has been in place since day 1.  It does not cause us any dependency problems and works well for now.

portante:
@atheurer, @ndokos, can you two take a look and review as well?

portante:
@ndokos, @atheurer, please take a look when you have a chance.  Thanks!

Issue #1027, opened 2019-01-14
We need to gracefully handle closed indices


portante:
Ignoring the fact that we have a bad date stamp in the index name, we need to handle these kinds of errors when an index is enountered that is closed.

```
run-2019-01-01T01:08:02-UTC: EC2::ip-10-0-202-145/pbench-user-benchmark__2018.12.31T15.47.22: success
   2019-01-01T01:10:00-UTC: Finished /pbench/archive/fs-version-001/EC2::ip-10-0-202-145/TO-INDEX/pbench-user-benchmark__2018.12.31T15.47.22.tar.xz (size 43364992)
   2019-01-01T01:10:00-UTC: Starting /pbench/archive/fs-version-001/b06-h31-1029p/TO-INDEX/fio_glusterfs-HDD-25GBper-pod-seq_write_read_2018.12.31T18.55.51.tar.xz (size 74338724)
        done templates (end ts: 2019-01-01T01:11:02-UTC, duration: 2.43s, successes: 3, retries: 0)
es_index: ERROR KeyError('create',)
es_index: WARNING - retrying action
{
    "index": {
        "_index": "dsa.pbench.result-data.1970-01-01",
        "_type": "pbench-result-data",
        "_id": "fba255dc184516b9a52d395cb3253008",
        "status": 403,
        "error": "IndexClosedException[[dsa.pbench.result-data.1970-01-01] closed]"
    }
}
```

portante:
See also #1026.

portante:
Fixed by #1028.

PR #1028, opened 2019-01-15
Fix zero date stamp

### LINKED ISSUE #1026: Indices generated with 1970-01-01 as their data prefix should NOT be allowed
The result data index chose below is, `dsa.pbench.result-data.1970-01-01`, but the date stamp is for data that has a `0` (zero) timestamp.  We should not be generating zero timestamps, and must not allow near-zero timestamps either.

```
run-2019-01-01T01:08:02-UTC: EC2::ip-10-0-202-145/pbench-user-benchmark__2018.12.31T15.47.22: success
   2019-01-01T01:10:00-UTC: Finished /pbench/archive/fs-version-001/EC2::ip-10-0-202-145/TO-INDEX/pbench-user-benchmark__2018.12.31T15.47.22.tar.xz (size 43364992)
   2019-01-01T01:10:00-UTC: Starting /pbench/archive/fs-version-001/b06-h31-1029p/TO-INDEX/fio_glusterfs-HDD-25GBper-pod-seq_write_read_2018.12.31T18.55.51.tar.xz (size 74338724)
        done templates (end ts: 2019-01-01T01:11:02-UTC, duration: 2.43s, successes: 3, retries: 0)
es_index: ERROR KeyError('create',)
es_index: WARNING - retrying action
{
    "index": {
        "_index": "dsa.pbench.result-data.1970-01-01",
        "_type": "pbench-result-data",
        "_id": "fba255dc184516b9a52d395cb3253008",
        "status": 403,
        "error": "IndexClosedException[[dsa.pbench.result-data.1970-01-01] closed]"
    }
}
```

Another issue about handling "`closed`" indices should be opened.

portante:
Begins to fix #1026.


portante:
Also addresses #1027 as well in a separate commit.

server/pbench/bin/index-pbench
```diff
@@ -242,12 +244,24 @@ class PbenchData(object):
         ts_float = float(orig_ts)/1000
         ts = datetime.utcfromtimestamp(ts_float)
         if self.start_run_ts is not None and ts < self.start_run_ts:
```
ndokos: Please add a comment here to explain what is going on.

server/pbench/bin/index-pbench
```diff
@@ -520,15 +534,12 @@ def mk_result_data_actions(ptb, options, idx_prefix, idx_patterns, opctx):
     sources = rd.make_source()
     if not sources:
         return
+    template = idx_patterns["result-data"]["template"]
     for source, source_id in sources:
-        tstamp = rd.convert_timestamp(source['@timestamp'])
-        year, month, day = tstamp.split('T', 1)[0].split('-')[0:3]
+        idx_name = rd.generate_index_name(template, idx_prefix, source)
```
ndokos: We are generating an index name from the timestamp in `source` (in both versions - this is not a change made by this PR): is this correct? We can end up using the 2019-01 index, even though the experiment was run in 2018-12. Should we not be using the run_metadata timestamp to generate the index name? I have a feeling I asked you this before, but I don't remember.

server/pbench/bin/index-pbench
```diff
@@ -2079,19 +2100,18 @@ class PbenchTarBall(object):
     def gen_run_timestamps(self):
         try:
             start_run_ts = datetime.strptime(self.start_run, "%Y-%m-%dT%H:%M:%S.%f")
-        except ValueError:
+        except ValueError as e:
```
ndokos: Should all the `except ValueError` be `except ValueError as e`? For consistency, if for nothing else.

portante: We won't use `e` any where, that was left over from a previous attempt to fix the problem that I abandoned.  I'll be refactoring this code in a bit and so this will be removed.

server/pbench/bin/index-pbench
portante: Well, for some of the code we were using the value from the `@timestamp` field (tool data and result data, I believe) and for other records we were using another value (have to look at run and toc docs to see what they were doing).

This change to use `generate_index_name()` is attempting to always make it self-consistent: we ONLY ever generate the timestamp we use in the index name from the source data's `@timestamp` field.

server/pbench/bin/index-pbench
portante: Sure, something like:

```
The calculated timestamp, `ts`, is earlier that the timestamp of the start of the pbench run itself, `start_run_ts`.  That can occur because of at least 3 different situations: 1. the recorded start run timestamp, `start_run_ts`, is not correct; 2. the `ts` is a valid absolute timestamp that was incorrectly recorded; 3. the `ts` is a valid _relative_ timestamp, measured from the beginning of some unknown time `T0`.

Since we do not have enough information at this point to correct or detect the 1st or 2nd situation, we make an attempt to detect and correct the 3rd situation.

For the 3rd situation, we make an attempt to treat the original timestamp as if it is relative to the start of the run, and calculate a _new_ absolute timestamp from that point.  If that new absolute timestamp is before the end of the run as well, we just use the new absolute timestamp instead of the relative timestamp.  If the new absolute timestamp is beyond the end of the run, it is likely that the timestamp is not relative so we use it as is.
```

Arguably, we should disallow timestamps before and after the run, but that is for another PR to solve.

portante:
@ndokos, okay, based on your feedback, I refactored the code one more time to eliminate some confusion by consolidating the timestamp handling entirely in one routine.  This should remove most of the different ways in which we do things.

server/pbench/bin/index-pbench
```diff
@@ -241,12 +242,43 @@ class PbenchData(object):
...
+            # and correct the 3rd situation.
+            #
+            # For the 3rd situation, we make an attempt to treat the original
+            # timestamp as if it is relative to the start of the run, and
+            # calculate a _new_ absolute timestamp from that point.  If that
+            # new absolute timestamp is before the end of the run as well, we
+            # just use the new absolute timestamp instead of the relative
+            # timestamp.  If the new absolute timestamp is beyond the end of
+            # the run, it is likely that the timestamp is not relative so we
+            # use it as is.
```
portante: The flaw in this logic is that it is determined in isolation each time.  We might want to determine this once and decide to apply it consistently for all related timestamps.

server/pbench/bin/index-pbench
```diff
@@ -241,12 +242,43 @@ class PbenchData(object):
...
+                ts = newts
+        if self.ptb.end_run_ts is not None and ts > self.ptb.end_run_ts:
+            self.counters['ts_after_end_run_ts'] += 1
         return ts.strftime("%Y-%m-%dT%H:%M:%S.%f%z")
 
+    def generate_index_name(self, template, prefix, source, tool=None):
+        """Return a fully formed index name given its template, prefix, source
+        data (for an @timestamp field) and an optional tool name."""
+        year, month, day = source['@timestamp'].split('T', 1)[0].split('-')[0:3]
+        assert (self.year, self.month, self.day) >= (year, month, day)
```
portante: The `start_run` `(year,month,day)` tuple should always be less-than or equal-to the given `@timestamp` `(year,month,day)` tuple, not greater-than or equal-to.

portante:
@ndokos, this is ready for review now.

portante:
Sorry just a few of the commits updated to bump the version number used so that we can track all the changes being made in the wild.

server/pbench/bin/index-pbench
```diff
@@ -265,20 +269,24 @@ class PbenchData(object):
...
             # timestamp.  If the new absolute timestamp is beyond the end of
             # the run, it is likely that the timestamp is not relative so we
-            # use it as is.
+            # raise an error.
             d = timedelta(0, 0, orig_ts * 1000)
-            newts = self.ptb.start_run_ts + d
-            if newts > self.ptb.end_run_ts:
+            ts = self.ptb.start_run_ts + d
+            if ts > self.ptb.end_run_ts:
                 self.counters['ts_before_start_run_ts'] += 1
```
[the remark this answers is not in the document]
portante: It seems to be.

We are counting the fact that the original time stamp given before the start of the run.  We only count this error when the attempt to convert a relative timestamp into an absolute timestamp results in that absolute timestamp being beyond the end of the run (meaning it is NOT a relative timestamp).

portante:
@ndokos, let me know if you agree with that comment.  Once you mark it approved I'll merge it.  Thanks!

ndokos:
Oh, right: I see now.

PR #1030, opened 2019-01-17
Dashboard deploy

portante:
Fix the ansible playbook for installing the development environment, and some enhancements to the deployment playbook based on what we found deploying to production.  We also rename the dashboard URL to `/dashboard`.


gurbirkalsi:
@portante These changes look good. Just tested in my environment and was able to get a deployment to one of my VMs without any issues.

PR #1035, opened 2019-01-21
Reduce pidstat space by 90%

atheurer:
-Skips redundant data in pidstat output at runtime
-Stores data by PID/cmd


agent/tool-scripts/postprocess/pidstat-postprocess
```diff
@@ -70,88 +69,69 @@ if ( (($? >> 8) == 0) && (! -e "/usr/bin/vdsClient") && (-e "/usr/bin/virsh") )
...
-		if ( $pid =~ /(\d+)\s+(\d+)/ ) {
-			$pid = $1;
-			$tid = $2;
-			if ($tid == 0) {
-				next;
-			} else {
-				$pid = $tid;
-				$cmd =~ s/\|_+//;
+my %timestamps;
+my @attributes = qw(usr system guest CPU_PCT CPUID minflt_s majflt_s VSZ RSS MEM_PCT kB_rd_s kB_wr_s iodelay cswch_s nvcswch_s);
```
portante: Is there a way we can keep this list of attributes stored either here or in `pidstat-convert` to avoid the possibility drift in the future?

agent/tool-scripts/postprocess/pidstat-convert
```diff
@@ -0,0 +1,114 @@
...
+		# maintain a file for each of the PIDs instead of having all data in 1 file
+		my $file = $tool_output_dir . "/pids/" . $pid . "/" . $cmd;
+		if ( ! -e $file ) {
+			mkdir($tool_output_dir . "/pids/" . $pid);
+			open($file_handles{$pid}, ">", $file) || die "Could not open file $file";
+			select($file_handles{$pid});
+			$| = 1; # always flush
+			select(STDOUT);
+		}
+		if (exists $file_handles{$pid}) {
```
portante: What does it mean when this is `false`?

Doesn't line 89 mean that if it doesn't `die` there is a file handle in `file_handles{$pid}`?

And do we need to close the old file handle first?

atheurer: Originally I did not have this check, and for only a few pids/filehandles, I would get a "does not exist" for the filehandle, so I added this check.  It's unclear to me why the file-open would work, but I would not get a file handle.  I'll run a few more tests to see if we can find out why.

atheurer: I am not seeing this problem any more, and it may be for two reasons:
-I originally did not have ulimit -n included in the command to run pidstat-convert, and we may have hit the default limit of 1024 max open files
-When I was originally testing this, I cat the output of a previously run, pidstat stdout (3.2GB) to pidstat-convert, so it was not receiving new data every second, but as fast as it could process it.  Maybe there's an issue with opening files that quickly.

atheurer:
I have added the new perl module for sysstat.  We can add common functions for sysstat post-processing there in the future.  The new commits also include capturing sdtout and sdterr for post-processing.  I tested for filehandles again and cannot reproduce the no-filehandle issue, but I have added a warning message in case it happens.  Also, I added the close for all the files at the end.

atheurer:
I will need to add some "gold" files for unittests for sar,mpstat,iostat,pidstat to pass.  These files are $tool-pp-std[out|err].txt

atheurer:
There is some more work I will have to do, as the CSV and average files generated do not match the gold files when running the unittest.  Some of this is due to labels being slightly different, but some of it is the values are not the same, which is a problem.

atheurer:
I have added a few more commits to fix the unittests and to fix a problem in parsing the column data (we missed a column).  The unittests now must have the ./pids dir and its files in order to run, which it does now.  The good thing is all the data the original gold files have is exactly reproduced with the new code.

The unittest only runs pidstat-postprocess.  If we want to unittest pidstat-convert, I think it could be done, but we'll need to do a bit more work to make that happen.

agent/tool-scripts/postprocess/pidstat-convert
```diff
@@ -0,0 +1,114 @@
...
+			my $tid = $2;
+			if ($tgid =~ /^0|\-$/) {
+				$pid = $prev_pid . "-" . $tid;
+				$cmd =~ s/\|__//;
+			} else {
+				$prev_pid = $tgid;
+				next; # Not going to save the per-process stats because we have the per-thread stats
+			}
+		}
+		# filenames have restrictions, and a process name with arguments don't always work as filename
```
ndokos: "..,a process name with arguments does not always work as a filename."

agent/tool-scripts/postprocess/pidstat-convert
```diff
@@ -0,0 +1,114 @@
...
+		
+		$cmd =~ s/%/%%/g; # avoid 'Invalid conversion in printf: "%=" ' errors if cmd contains a %.;
+		$cmd =~ s/\s$//g; # remove the last space pidstat always adds
+		$cmd =~ s/\s/_/g; # convert spaces to _
+		$cmd =~ s/\//_/g; # remove the first _
```
ndokos: This changes / to _ throughout, correct? So the comment needs fixing?

agent/tool-scripts/postprocess/pidstat-convert
```diff
@@ -0,0 +1,114 @@
...
+		$cmd =~ s/\'//g;
+		$cmd =~ s/\"//g;
+		$cmd =~ s/^_//g;
```
ndokos: The comment from above belongs here.

portante:
Can we consider #1046, which is just your code with all the pid files removed for the unit tests?

portante:
Don't close just yet.  I'll move the comments from Nick over to the new PR and address them.

agent/tool-scripts/postprocess/pidstat-convert
```diff
@@ -0,0 +1,114 @@
...
+	}
+}
+use lib "$_test_alt_tools_lib";
+no lib ".";
+use File::Path 'rmtree';
+
+my %previous_stats; # used to detect repeated value
+my %current_stats;
+my @attributes = qw(usr system guest CPU_PCT CPUID minflt_s majflt_s VSZ RSS MEM_PCT kB_rd_s kB_wr_s iodelay cswch_s nvcswch_s);
+my $tool_output_dir = shift; # the one and only argument to this script, the directory where these file shoudl be written
```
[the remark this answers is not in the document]
portante: Will be addressed in the same file in #1046.

agent/tool-scripts/postprocess/pidstat-convert
```diff
@@ -0,0 +1,114 @@
...
+if (not defined $tool_output_dir) {
+	print "Directory to store PIDs is required\n";
+	exit 1;
+}
+my %file_handles;
+if ( -e $tool_output_dir . '/pids') {
+	rmtree($tool_output_dir . '/pids');
+}
+mkdir($tool_output_dir . '/pids');
+# read the STDIN which is expcted to be STDOUT from 'pidstat -l -H -w -u -h -d -r  -p ALL'
```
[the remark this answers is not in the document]
portante: Will be addressed in the same file in #1046.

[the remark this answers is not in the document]
portante: Fix in #1046 PR.

agent/tool-scripts/postprocess/pidstat-convert
```diff
@@ -0,0 +1,114 @@
...
+	###v11 with threads
+        # Time   UID      TGID       TID    %usr %system  %guest    %CPU   CPU  minflt/s  majflt/s     VSZ     RSS   %MEM   kB_rd/s   kB_wr/s kB_ccwr/s iodelay   cswch/s nvcswch/s  Command
+        # 1547852254     0         1         0    0.00    0.00    0.00    0.00     4      0.13      0.00  194520    7572   0.00      1.16     28.14      1.62      89      0.62      0.01  /usr/lib/systemd/systemd --switched-root --system --deserialize 22
+        # 1547852254     0         0         1    0.00    0.00    0.00    0.00     4      0.13      0.00  194520    7572   0.00      0.02      0.00      0.00      89      0.62      0.01  |__/usr/lib/systemd/systemd --switched-root --system --deserialize 22
+        #
+        if ( $line =~ /\s{0,1}(\d+)\s+\d+\s+(\S+|\S+\s+\S+)\s+(\d+\.\d+\s+\d+\.\d+\s+\d+\.\d+\s+\d+\.\d+\s+\d+\s+\d+\.\d+\s+\d+\.\d+\s+\d+\s+\d+\s+\d+\.\d+\s+\d+\.\d+\s+\d+\.\d+\s+\d+\.\d+\s+\d+\s+\d+\.\d+\s+\d+\.\d+\s+)(.*)/ ) {
+		my $time = $1;
+		my $pid = $2;
+		my $values_str = $3;
+		my $cmd = substr $4, 0, 40; # limit the length to 60 chars
```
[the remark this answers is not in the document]
portante: Fix in #1046 PR.

agent/tool-scripts/postprocess/pidstat-convert
portante: Fix in #1046 PR.

agent/tool-scripts/postprocess/pidstat-convert
portante: Fix in #1046 PR.

agent/tool-scripts/postprocess/pidstat-convert
portante: Fix in #1046 PR.

portante:
Closing in favor of #1046.

PR #1036, opened 2019-01-23
Merge index-pbench into pbench-index using all python

portante:
In an attempt to prepare for parallelizing the indexing process, we
merge the existing functionality of `pbench-index` with `index-pbench`,
dropping the `index-pbench` name and retaining the `pbench-index` name.


server/pbench/bin/pbench-index
```diff
@@ -1,247 +1,2671 @@
...
-# index-pbench need to get at it
-export TMPDIR
-
-echo "$TS: starting at $(timestamp)"
-
-typeset -i nidx=0
-indexed=$TMPDIR/$PROG.$TS.indexed
-typeset -i nerrs=0
-erred=$TMPDIR/$PROG.$TS.erred
-typeset -
```
tenstormavi: Is this should be "logger" instead of "_logger" because you have used `logger.*` everywhere else in the code Or am I missing something here?

tenstormavi: like here.

portante: I think we want it to be `_logger` so that we use the convention of variables marked with a leading underscore denoting module local, not to be exported.

run-unittests
```diff
@@ -46,10 +52,11 @@ echo ""
 echo ""
 echo "+++ Agent util-scripts Unit Tests +++"
 echo ""
+tail -n 9999999 -f /var/tmp/agent.util-scripts.out --pid $tpid &
```
ndokos: You don't worry about any other tail process *except* for the server scripts tail process which you wait for. Is that necessary? And if it is, why is it not necessary for all the other tails?

portante: Resolved this removing duplicate code, making a `tail_and_wait` function to do the same thing, use it for all the unit tests that are run, and finally, stop using a background tail and just let it run in the foreground.

portante:
FYI: Builds appear to be failing due to some Ubuntu key server.

server/bin/pbench-index
```diff
@@ -1,259 +1,3563 @@
...
+                datum[identifier] = _dict_const([
+                    # Since they are all the same, we use the first to
+                    # generate the real timestamp.
+                    ('@timestamp', ts_val),
+                    ('@timestamp_original', str(first)),
+                    ('run', self.run_metadata),
+                    ('iteration', self.iteration_metadata),
+                    ('sample', self.sample_metadata),
+                    (self.toolname, _dict_const())
+
```
ndokos: Do we need a signal handler to be able to do this deletion (and maybe others) if the process gets killed?

ndokos:
@portante: not sure why, but when running unittests in parallel in my clone of this branch, I got the following error:

     ...
    PASS - test-16
    ./unittests: line 745: wait: pid 2698 is not a child of this shell
    cat: /var/tmp/pbench-test-server/test-17/output.diff: No such file or directory
    FAIL (sts = 127) - test-17
    PASS - test-18

Running it again to see if it is at all reproducible.

server/bin/pbench-base.sh
```diff
@@ -1,62 +1,47 @@
...
 fi
 
-# Required
-TOP=$(getconf.py pbench-top-dir pbench-server)
-test -d $TOP || doexit "Bad TOP=$TOP"
-
-# Required
-TMP=$(getconf.py pbench-tmp-dir pbench-server)
-test -d $TMP || doexit "Bad TMP=$TMP"
+$dir/pbench-base $PROG > /tmp/pbench-base.sh.$$
```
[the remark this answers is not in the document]
portante: Yeah, need to rework how we invoke server side components entirely to avoid this fragile construct.

portante:
@ndokos, regarding the spurious test error for missing `output.diff` file, the return status of the sub-process, `127` as seen in your report indicates the `wait` command returned the value because the process did not exist to get a real status, as reported earlier with "`pid 2698 is not a child of this shell`".

It is likely something in your environment prevented the test from starting in the background.

server/bin/pbench-index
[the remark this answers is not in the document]
portante: Not sure what you mean by this comment.  It seems like you are trying to report that a line 3291 we return a value of 12 when we get an error during setup which is not expected.  Is that what you mean?

portante: Are you referring to the `os.remove(ie_filename)` at line 3563?

portante:
With this latest update I have addressed the unit test failure in `test-5.2` (making sorted list of tar balls stable), and removed the old `index-pbench` server script.

portante:
Updated to add support for quarantining tar balls.

portante:
@ndokos, @tenstormavi, rebased this on PR #1149, which gives us the `pbench-report-status` written in python, now callable from `pbench-index`.

The last piece of this is to write a wrapper for the server commands so that we can properly invoke the correct python interpreter in all the deployed environments we support, and so that we don't need to call `getconf.py` so often.

server/bin/pbench-base.sh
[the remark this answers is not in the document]
portante: Fixed as of PR #1153.

portante:
Rebased on top of #1153, which takes care of the last comment.  This should be good to go when #1149 and #1153 are merged first.

@ndokos, @tenstormavi, please have a look.

ndokos:
@portante: can you rebase this branch?

ndokos:
This needs rebasing now that #1153 is in but I'm not sure why it says it cannot be rebased safely: when I did the `git pull`, it merged everything automatically with no problems.

 I also get random errors when running `unittests` in parallel - not that there is anything wrong with the unit tests - the error looks like this:
    
    ./unittests: line 760: wait: pid 21500 is not a child of this shell
    cat: /var/tmp/pbench-test-server/test-21/output.diff: No such file or directory

but the only consistency is that it happens in the wait on line 760: the pid and the "failing" test are random. It all works properly with --serial.

portante:
> This needs rebasing now that #1153 is in but I'm not sure why it says it cannot be rebased safely: when I did the git pull, it merged everything automatically with no problems.

@ndokos, rebased.

portante:
I am not seeing the error below when running the unit tests in parallel mode on my server.

```
./unittests: line 760: wait: pid 21500 is not a child of this shell
cat: /var/tmp/pbench-test-server/test-21/output.diff: No such file or directory
```

The `wait` error could happen if a unit tests job was somehow paused for a long enough time that pids were re-used.  I am not sure how that could happen normally.

The second error, `output.diff: No such file or directory`, happens when something cleans up the `/var/tmp/pbench-test-server/` directory hierarchy before the unit test script finishes. This can happen if a unit test run is put in the background, and then suspended, and another unit tests job starts to run.

PR #1037, opened 2019-01-23
Query prefix field in run document for summary visualization

gurbirkalsi:
The dashboard redux service will query the prefix value for a run before building the url to fetch result.json docs from the production environment.

This also improves the index selection workflow by removing ranges and allowing users to select specific indices to query run documents. This required a restructuring of redux services by moving specific dispatched actions to the `global` service in order to provide data at the appropriate scope throughout the dashboard.

### DIFF
--- a/web-server/v0.4/src/models/global.js
+++ b/web-server/v0.4/src/models/global.js
@@ -1,4 +1,4 @@
-import { queryDatastoreConfig } from '../services/global';
+import { queryDatastoreConfig, queryMonthIndices } from '../services/global';
 
 export default {
   namespace: 'global',
@@ -6,6 +6,9 @@ export default {
   state: {
     collapsed: false,
     datastoreConfig: {},
+    indices: [],
+    selectedIndices: [],
+    selectorIndices: ['0'],
   },
 
   effects: {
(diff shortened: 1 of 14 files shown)


web-server/v0.4/src/models/global.js
```diff
@@ -17,6 +20,37 @@ export default {
         payload: response,
       });
     },
+    *fetchMonthIndices({ payload }, { call, put }) {
+      let response = yield call(queryMonthIndices, payload);
+
+      let indices = [];
+      response.map(index => {
+        if (index.index.includes('dsa.pbench.run')) {
```
portante: Use of `dsa.pbench` here should probably be replaced with the prefix value from the configuration file.

gurbirkalsi: @portante The global redux model now takes advantage of the datastoreConfig file to build the prefix for reference when filtering month indices.

gurbirkalsi:
[fix for querying prefix fields in run documents](https://trello.com/c/fhlFM6d0/64-fix-for-querying-prefix-fields-in-run-documents)

PR #1040, opened 2019-01-25
Reduce overhead for local dev by using node environment vars

gurbirkalsi:
Reference the node environment status in order to configure
appropriate endpoints and public paths for development and
production environments.

### DIFF
--- a/web-server/v0.4/README.md
+++ b/web-server/v0.4/README.md
@@ -40,7 +40,7 @@ Pbench Dashboard is a web-based platform for consuming indexed performance bench
 
 Assets placed in the `public` directory are copied to the `dist` directory for reference in the generated `index.html` file during the build process.
 
-Assets placed in the `src/assets/` directory are only referenced within components or layouts definitions and is packaged in the generated `***.js` file during the build process.
+Assets placed in the `src/assets/` directory are only referenced within component or layout definitions and are packaged in the generated `***.js` file during the build process.
 
 
 ## Installation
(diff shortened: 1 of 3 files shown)


gurbirkalsi:
@portante This PR has been deployed on my UI VM as well in order to test differences between the dev and prod environments.

web-server/v0.4/README.md
```diff
@@ -61,13 +61,7 @@ This will automatically open the application on [http://localhost:8000](http://l
...
 ## Local Development
 
-Both the production and development builds of the dashboard require specific configurations in order to run on their respective environment. For local development, modify the following configuration definitions:
-
-`.webpackrc.js`
-
-```Javascript
-publicPath: '/'
-```
+Both the production and development builds of the dashboard require specific configurations in order to run on their respective environment. For local development, modify the following configuration definition:
```
portante: Does the last sentence in this comment need to change now?

web-server/v0.4/src/services/global.js
```diff
@@ -1,8 +1,20 @@
...
-  return request('http://localhost:8000/dev.config.json', {
+  // Note that window.location.pathname should already have a trailing slash.
+  let configEndpoint = '';
+  let environment = process.env || 'development';
+  
+  switch(environment) {
+    case 'development':
+      configEndpoint = 'http://localhost:8000/dev.config.json'
+      break
+    case 'production':
```
portante: Should we be using a simple `if` statement here instead of a `switch` statement?  What happens if somebody sets `process.env` to "`staging`"?

portante: Could you add a comment there to that effect?

But the work "production" does not sound like it fits and seems confusing.  Why not just have one variable called, "`configEndpoint`"?  What does the current approach let you do different?

web-server/v0.4/README.md
portante: We before you had lines 65-70 in place, but now you removed them leaving the last line reading, "For local development, modify the following configuration definition:", but nothing appears to be there.

web-server/v0.4/src/services/global.js
portante: Great, that helps.  Thanks.

gurbirkalsi:
@portante Let me know if the updated changes resolves the prior conversations.

gurbirkalsi:
[reduce overhead for local dev using node environment variables](https://trello.com/c/MouhNbnJ/65-reduce-overhead-for-local-dev-using-node-environment-variables)

PR #1041, opened 2019-01-25
Adds  full PDF download feature to comparison page

Anisha1234:
This PR will add following feature;

- Will add feature to download the comparison page of pbench dashboard as PDF.
- Uses jsPDF and html2canvas library for the PDF conversion.

This PR consists of two parts
1. Download whole page as PDF:
The [1st commit](https://github.com/distributed-system-analysis/pbench/pull/1041/commits/d061a20d38618757eebff3944c1cdfa73d5b0b9d) consists of working code for downloading whole comparision as PDF.

2. Download user selected divisions only:

- The [2nd commit](https://github.com/distributed-system-analysis/pbench/pull/1041/commits/86e85b0282b845ad1e9e80618bb8a3c12ac697a8) consists requested changes and styling of checkbox with addition of description box for selected divisions.

- Due to problem in integrating promise to the functions, the feature of downloading custom selected divisions as PDF is still under progress.

Working Process:
The description box shows the divisions selected to be downloaded (by default: **all**). On changing the checkbox, the divisions will be selected and on pressing share button, the combined PDF of the selected divisions will be downloaded. 
main1


web-server/v0.4/src/routes/Dashboard/RunComparison.js
```diff
@@ -686,15 +736,22 @@ export default class RunComparison extends ReactJS.Component {
...
-              id={table}
-              title={() => <h5>{table}</h5>}
-              columns={columns}
-              dataSource={tableData[table]}
-              expandedRowRender={expandedRowRender}
-              pagination={{ pageSize: 20 }}
-            />
+            <div id={'table' + table}>
+              <input
+                type="checkbox"
```
portante: So will this be displayed as an actual checkbox or will it be displayed as a simple button that can be pressed?

web-server/v0.4/src/routes/Dashboard/RunComparison.js
```diff
@@ -649,7 +694,12 @@ export default class RunComparison extends ReactJS.Component {
                   (Object.keys(timeseriesDropdown).length > 0) ? (
                     <div style={{ marginBottom: 32 }}>
                       {Object.keys(tableData).map(table => (
-                        <div>
+                        <div id={'alltable' + table}>
+                          <input
+                            type="checkbox"
```
portante: So will this be displayed as an actual checkbox or will it be displayed as a simple button that can be pressed?

web-server/v0.4/src/routes/Dashboard/RunComparison.js
```diff
@@ -533,7 +573,12 @@ export default class RunComparison extends ReactJS.Component {
                 <div style={{ margin: 16 }}>
                   <Spin spinning={this.state.loading}>
                     {Object.keys(clusteredGraphData).map(cluster => (
-                      <div>
+                      <div id={'cluster' + cluster}>
+                        <input
+                          type="checkbox"
```
portante: So will this be displayed as an actual checkbox or will it be displayed as a simple button that can be pressed?

web-server/v0.4/src/routes/Dashboard/RunComparison.js
```diff
@@ -334,30 +334,70 @@ export default class RunComparison extends ReactJS.Component {
...
+          doc.addPage();
+          doc = doc.addImage(imgData, 'PNG', 0, position, imgWidth, imgHeight);
+          heightLeft -= pageHeight;
+        }
+        doc.save();
+      });
     });
   };
 
+  selectDiv = (type, cluster) => {
```
portante: Can you add some comments explaining what the different code paths mean?

web-server/v0.4/src/routes/Dashboard/RunComparison.js
```diff
@@ -334,30 +334,70 @@ export default class RunComparison extends ReactJS.Component {
     let end = this.state.checked.length;
```
portante: Do we need this variable, `end`?  It does not appear to be used anywhere.

web-server/v0.4/src/routes/Dashboard/RunComparison.js
```diff
@@ -334,30 +334,70 @@ export default class RunComparison extends ReactJS.Component {
...
-      position = fullIHeight + 5;
-      fullIHeight = fullIHeight + imgHeight;
-      var heightLeft = fullIHeight;
-      doc = doc.addImage(imgData, 'PNG', 0, position, imgWidth, imgHeight);
-      heightLeft -= pageHeight;
-      while (heightLeft >= 0) {
-        position = heightLeft - fullIHeight;
-        console.log('position renew : ' + position);
-        doc.addPage();
+    var requests = this.state.checked.map((check, index) => {
```
portante: Why do we need to use the `requests` variable?

web-server/v0.4/src/routes/Dashboard/RunComparison.js
Anisha1234: @portante  To select the custom divisions, I am planning to have checkboxe. We can hide them initially and show them once the share button is pressed.

web-server/v0.4/src/routes/Dashboard/RunComparison.js
Anisha1234: To select the custom divisions, I am planning to have checkbox. Once the divisions are selected, we can download the pdf with clicking on share button. 
screenshot from 2019-01-28 14-54-39

web-server/v0.4/src/routes/Dashboard/RunComparison.js
Anisha1234: Yes, have added few comments in current commit.

gurbirkalsi:
@Anisha1234 This looks great so far! A few UX suggestions: 

- Renaming the `Share` button on the details section to `Export` in order to make it clear to the user that they will be exporting the page as a PDF. 

- Adding a description field to the export form in order to enable the user to enter information regarding the selected runs, methodology, intended heuristics, etc. 

- Making the file name field optional and programmatically constructing the file name using the run names or perhaps a timestamp indicating when they exported the PDF (`moment.js` library is helpful for this). 

- Grouping the render options in a card and enabling all options by default for quick exporting of the entire run comparison page. 

- Using the `Spin` component to show a loading indicator in the modal while the PDF is being built once the user has click on the save button. 

I've created a quick mockup of the export modal given these UX changes using CodeSandbox: https://codesandbox.io/s/v37qvxvpvl

Let me know if these suggestions sound reasonable and if they make sense for the workflow we are targeting here!





v37qvxvpvl_-_codesandbox

web-server/v0.4/src/routes/Dashboard/RunComparison.js
```diff
@@ -38,6 +56,16 @@ export default class RunComparison extends ReactJS.Component {
       graphKeys: [],
       tableData: [],
       loading: false,
+      generatingPdf: false,
+      modal: {
```
gurbirkalsi: The state items on lines 60 and 65 are initialized as objects but are only referenced by their children keys. We can flatten these objects and reference their children keys to improve performance and readability. 

Currently, the `RunComparison` component creates a new item within the component's state as the children keys are not referenced by their parent key. This creates two sets of the same state item: e.g.
```
pdf: {
  pdfHeader: '',
  confirmLoading: false
},
pdfHeader: ''
```

web-server/v0.4/src/routes/Dashboard/RunComparison.js
```diff
@@ -38,6 +56,16 @@ export default class RunComparison extends ReactJS.Component {
...
+        visible: false,
+      },
+      checked: ['details', 'summary', 'table', 'timeseries'],
```
gurbirkalsi: What do you think about pulling the behaviors of `checked` and `defaultChecked` together? Since we initialize the checked array with the default values for pdf creation, we can reference this array in the export modal for both cases where all values are selected and where the user makes changes to the selection.

web-server/v0.4/src/routes/Dashboard/RunComparison.js
```diff
@@ -38,6 +56,16 @@ export default class RunComparison extends ReactJS.Component {
...
+      defaultChecked: [],
+      pdf: {
+        pdfHeader: ' ',
+        confirmLoading: false,
+      },
```
gurbirkalsi: We'll also want add a state item for `pdfName`.

web-server/v0.4/src/routes/Dashboard/RunComparison.js
gurbirkalsi: Can you please update the state names to reflect its intended use case within the code? E.g. `selectedComponents` provides further insight into what is being selected over `checked`.

web-server/v0.4/src/routes/Dashboard/RunComparison.js
```diff
@@ -303,6 +331,126 @@ export default class RunComparison extends ReactJS.Component {
     this.setState({ timeseriesDropdownSelected: timeseriesDropdownSelected });
   };
 
+  savePDF = async () => {
```
gurbirkalsi: Since this function call takes advantage of Promises, we can remove the reference to `async` as the usage of Promises ensures an asynchronous behavior in the function call.

web-server/v0.4/src/routes/Dashboard/RunComparison.js
```diff
@@ -303,6 +331,126 @@ export default class RunComparison extends ReactJS.Component {
...
+    /* maps through the state array containing the IDs and saves the pdf */
+    var doc = new jsPDF('p', 'mm', [1050, 1475]);
+    var position = 5;
+    var fullIHeight = 0;
+    var promises = this.state.defaultChecked.map(async (check, index) => {
```
gurbirkalsi: Instead of using a `map` function here, we can use a `forEach` function which avoids return a new array during each call. For more on the differences between `map` and `forEach`: 

[https://codeburst.io/javascript-map-vs-foreach-f38111822c0f](https://codeburst.io/javascript-map-vs-foreach-f38111822c0f)

web-server/v0.4/src/routes/Dashboard/RunComparison.js
```diff
@@ -303,6 +331,126 @@ export default class RunComparison extends ReactJS.Component {
...
+        canvas.getContext('2d');
+        var imgData = canvas.toDataURL();
+        var imgWidth = 345;
+        var pageHeight = 510;
+        var imgHeight = (canvas.height * imgWidth) / canvas.width;
+        position = fullIHeight + 5;
+        fullIHeight = fullIHeight + imgHeight;
+        var heightLeft = fullIHeight;
+        doc = doc.addImage(imgData, 'PNG', 10, position, imgWidth, imgHeight);
+        doc.text(this.state.pdfHeader, 172, 10, 'center');
```
gurbirkalsi: Since we are calling `doc.text` in a loop, we are writing the text to the document for how many items we are iterating over the selected items array. This can be moved outside of the loop to reduce the number of times we write to the pdf document.

web-server/v0.4/src/routes/Dashboard/RunComparison.js
```diff
@@ -303,6 +331,126 @@ export default class RunComparison extends ReactJS.Component {
...
+        heightLeft -= pageHeight;
+        while (heightLeft >= 0) {
```
gurbirkalsi: You introduced a while statement here to offset the position of the document while adding an image; however, it might make sense to remove this while statement as you assign the `heightLeft` just before the loop.

web-server/v0.4/src/routes/Dashboard/RunComparison.js
```diff
@@ -385,15 +534,67 @@ export default class RunComparison extends ReactJS.Component {
               </Button>
             </div>
           </Description>
+          <Description term="Slected Division">
```
gurbirkalsi: Since we reference the selected components in the export modal, we can remove this data from the description component.

gurbirkalsi:
You'll notice that the performance of the web application is slow when updating the pdf title and description in the export modal. We may want to consider updating the state when the user has clicked the "Save" button in the modal. We can accomplish this using the `<Form>` component.

web-server/v0.4/src/routes/Dashboard/RunComparison.js
Anisha1234: > The state items on lines 60 and 65 are initialized as objects but are only referenced by their children keys. We can flatten these objects and reference their children keys to improve performance and readability.
> 
> Currently, the `RunComparison` component creates a new item within the component's state as the children keys are not referenced by their parent key. This creates two sets of the same state item

I will resolve this issue in the next commit.

web-server/v0.4/src/routes/Dashboard/RunComparison.js
Anisha1234: > What do you think about pulling the behaviors of `checked` and `defaultChecked` together? Since we initialize the checked array with the default values for pdf creation, we can reference this array in the export modal for both cases where all values are selected and where the user makes changes to the selection.

Actually what I am trying to do is, when the user selects all the divisions, I am considering the id of the whole page, so that while downloading the PDF, the system don't need to go through loop. And by default also(when no division is selected or all divisions selected), it will download whole page when clicked on save. I am renaming both the  arrays to defaultComponents (All the divisions selected by default) and selectedComponents(selected divisions after change of switch).

web-server/v0.4/src/routes/Dashboard/RunComparison.js
Anisha1234: By using forEach function, I am getting `"undefined is not iterable (cannot read property Symbol(Symbol.iterator))"` error in promise.all

web-server/v0.4/src/routes/Dashboard/RunComparison.js
Anisha1234: Added.

web-server/v0.4/src/routes/Dashboard/RunComparison.js
Anisha1234: Resolved.

web-server/v0.4/src/routes/Dashboard/RunComparison.js
Anisha1234: Resolved

web-server/v0.4/src/routes/Dashboard/RunComparison.js
gurbirkalsi:
```
savePDF = () => {
    const { defaultChecked, pdfHeader, pdfName } = this.state;
    /* maps through the state array containing the IDs and saves the pdf */

    document.getElementById("table").offsetHeight;
    var doc = new jsPDF('p', 'mm', [1050, 1475], true);
    doc.text(pdfHeader, 172, 10, 'center');
    var position = 5;
    var fullIHeight = 0;
    let selectedElements = [] ;

    defaultChecked.forEach((element) => {
      selectedElements.push(html2canvas(document.querySelector(`#${  element}`)).then((canvas, index) => {
        doc.setPage(index);
        canvas.getContext('2d');
        let imgData = canvas.toDataURL();
        let imgWidth = 345;
        let pageHeight = 510;
        let imgHeight = (canvas.height * imgWidth) / canvas.width;
        position = fullIHeight + 5;
        fullIHeight += imgHeight;
        let heightLeft = fullIHeight;
        doc = doc.addImage(imgData, 'JPEG', 10, position, imgWidth, imgHeight);
        heightLeft -= pageHeight;
        {/*while (heightLeft >= 0) {
          position = heightLeft - fullIHeight;
          doc.addPage();
          doc = doc.addImage(imgData, 'JPEG', 10, position, imgWidth, imgHeight);
          heightLeft -= pageHeight;
        }*/}
      }))
    })

    Promise.all(selectedElements).then(() => {
      doc.save(`${pdfName  }.pdf`);
      this.setState({
        visible: false,
        generatingPdf: false,
      });
    });
  };
```

Anisha1234:
> You'll notice that the performance of the web application is slow when updating the pdf title and description in the export modal. We may want to consider updating the state when the user has clicked the "Save" button in the modal. We can accomplish this using the `<Form>` component.
> You'll notice that the performance of the web application is slow when updating the pdf title and description in the export modal. We may want to consider updating the state when the user has clicked the "Save" button in the modal. We can accomplish this using the `<Form>` component.

@gurbirkalsi I have fixed the react state for pdf title and description. Now the states are getting updated when user clicks on "save" button. I have also added PDF compression feature which reduced PDF size to 200KB-350KB.

web-server/v0.4/src/routes/Dashboard/RunComparison.js
```diff
@@ -391,9 +527,48 @@ export default class RunComparison extends ReactJS.Component {
...
+          onOk={this.handleOk}
+          okText={'Save'}
+          onCancel={this.handleCancel}
+        >
+          <Spin spinning={this.state.generatingPdf}>
+            <Form>
+              <Form.Item
+                colon={false}
+                label="File Name"
+                extra="(Optional) Selected results will be used if left blank."
```
gurbirkalsi: We render a timestamp instead of the selected results here, can we update this statement to reflect that?

web-server/v0.4/src/routes/Dashboard/RunComparison.js
```diff
@@ -391,9 +527,48 @@ export default class RunComparison extends ReactJS.Component {
...
+              >
+                <Input addonAfter={'.pdf'} id="pdfName" />
+              </Form.Item>
+              <Form.Item colon={false} label="Description">
```
gurbirkalsi: Lets also add an `extra` param here to let the user know the description field is also optional.

gurbirkalsi:
@Anisha1234 This PR looks solid and is ready to be merged once we make those small changes above. PDF generation seemed to perform well on my local build; however, we might want to address dynamic generation in a future PR. I'll create a separate issue to track this - great work Anisha!

PR #1045, opened 2019-01-25
Fix pbench-move-results to only push valid results

### LINKED ISSUE #1039: pbench-move/copy-results should verify directory name matches "name" field in metadata.log
We have seen evidence of folks renaming the directories of pbench runs in `/var/lib/pbench-agent` to a new name from what it was originally before calling `pbench-move/copy-results`.  By doing this, the pbench UI will not be able to construct the URLs to the data because the new directory name is not indexed or stored with the pbench data.

The `pbench-move/copy-results` script should report an error and NOT move/copy this directory since it won't be indexed properly.

For example, somebody might run `pbench-fio` without using the `--config` parameter:
```
$ pbench-fio <various parameters>
```

Resulting in a result directory created in `/var/lib/pbench-agent/fio__2019.01.14T18.41.43` like:
```
$ ls -l /var/lib/pbench-agent/
total 212
-rw-r--r--. 1 root root 209974 Jan 14 18:14 pbench.log
drwxr-xr-x. 2 root root   4096 Jan 14 18:14 fio__2019.01.14T18.41.43
drwxr-xr-x. 2 root root   4096 Jan 14 18:14 tmp
drwxr-xr-x. 2 root root    131 Dec  4 23:49 tools-default
```

And then they might rename that directory to a name that has some meaning to them:

```
$ mv /var/lib/pbench-agent/fio__2019.01.14T18.41.43 /var/lib/pbench-agent/190114.1841-pbfio-randwrite-1i-4Kbs-4iod-10m_1x200-preAll-xfsDef
```

And then they invoke `pbench-move-results` which results in a tar ball being created named: `190114.1841-pbfio-randwrite-1i-4Kbs-4iod-10m_1x200-preAll-xfsDef.tar.xz`.

But the `metadata.log` file in the top-level directory of that result has the `name` field still specified as `fio__2019.01.14T18.41.43`, and an empty `config` field.

If the user invoked `pbench-fio` using `--config`:
```
$ pbench-fio --config "pbfio-randwrite-1i-4Kbs-4iod-10m_1x200-preAll-xfsDef" <various-paramaters>
```

Then `pbench-fio` would create the result directory for the user with the config name embedded in it:

```
$ ls -l /var/lib/pbench-agent/
total 212
-rw-r--r--. 1 root root 209974 Jan 14 18:14 pbench.log
drwxr-xr-x. 2 root root   4096 Jan 14 18:14 fio_pbfio-randwrite-1i-4Kbs-4iod-10m_1x200-preAll-xfsDef_2019.01.14T18.41.43
drwxr-xr-x. 2 root root   4096 Jan 14 18:14 tmp
drwxr-xr-x. 2 root root    131 Dec  4 23:49 tools-default
```

portante:
A valid result is defined to be where the name of the result directory
in /var/lib/pbench-agent is the same as the value in the metadata.log
file under the "name" option of the "pbench" section.

Fixes #1039.


ndokos:
@portante: I think we missed the cp -r ==> cp -a change in agent/utils-scripts/unittests. I don't see it in the master branch, but I'm not quite sure what happened.

PR #1046, opened 2019-01-26
Reduce pidstat space by 90%

portante:
- Skips redundant data in pidstat output at runtime
 - Stores data by PID/cmd
 - Create a perl module for sysstat processing
   - include a subroutine to return all pidstat attributes for v11.2

See original at #1035.

### DIFF
--- a/agent/tool-scripts/postprocess/pidstat-convert
+++ b/agent/tool-scripts/postprocess/pidstat-convert
@@ -1,0 +1,118 @@
+#!/usr/bin/perl
+
+# Pipe the output of a pidstat command to this script to:
+# - create a directory for each PID
+# - create a file named same as the process in the PID directory
+#   - output the per PID data to the PID file
+#   - if the values repeat in subsequent sample, omit the value
+
+use strict;
+use warnings;
+
+# Check for an alternate tools library path for testing
+my $_test_alt_tools_lib;
+BEGIN {
+        my $_pbench_tspp_dir = $ENV{'pbench_tspp_dir'};
+	$_test_alt_tools_lib=$ENV{_TEST_ALTERNATE_TOOLS_LIBRARY};
+	if (not defined $_test_alt_tools_lib or not -d $_test_alt_tools_lib) {
+                $_test_alt_tools_lib = "$_pbench_tspp_dir";
+	}
+}
+use lib "$_test_alt_tools_lib";
+no lib ".";
+use File::Path qw(rmtree make_path);
+use File::Basename;
+use SysStat qw(get_pidstat_attributes);
+
+my %previous_stats; # used to detect repeated value
+my %current_stats;
+my @attributes = get_pidstat_attributes();
+my $tool_output_dir = shift; # the one and only argument to this script, the directory where these files should be written
+if (not defined $tool_output_dir) {
+	print "Directory to store PIDs is required\n";
+	exit 1;
+}
+my %file_handles;
+if ( -e $tool_output_dir . '/pids') {
+	rmtree($tool_output_dir . '/pids');
+}
+mkdir($tool_output_dir . '/pids');
+# read the STDIN which is expected to be STDOUT from 'pidstat -l -H -w -u -h -d -r -p ALL'
+my $prev_pid;
+while (my $line = <>) {
+	# example STDIN
+	#
+	###v11 no threads
+        ## Time        UID       PID    %usr %system  %guest   %CPU   CPU  minflt/s  majflt/s     VSZ     RSS   %MEM   kB_rd/s   kB_wr/s kB_ccwr/s iodelay   cswch/s nvcswch/s  Command
+        # 1547780606     0      2368    0.00    0.00    0.00    0.00     4      0.00      0.00  112808    4300   0.00      0.00      0.00      0.00       0      0.00      0.00  /usr/sbin/sshd -D
+        # 1547780606     0      2373    0.00    0.00    0.00    0.00     0      0.00      0.00  228772    7940   0.00      0.00      0.00      0.00       0      0.00      0.00  /usr/sbin/rsyslogd -n
+        # 1547780606     0      2380    0.00    0.00    0.00    0.00     7      0.00      0.00 1221772   15360   0.01      0.00      0.00      0.00       0      0.00      0.00  /usr/sbin/libvirtd
+        
+	###v11 with threads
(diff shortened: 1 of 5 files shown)


portante:
@atheurer, this is a rebase and squash of your commit, but invoking `pidstat-convert` as part of the unit tests infrastructure instead of providing all of those `/pids/*` files.  This might be worth considering given that exercise of `pidstat-convert` and the lower number of files in the commit to consider.

portante:
Waiting for @ndokos's review.

agent/tool-scripts/postprocess/pidstat-convert
```diff
@@ -0,0 +1,118 @@
...
+        # 1547780606     0      2380    0.00    0.00    0.00    0.00     7      0.00      0.00 1221772   15360   0.01      0.00      0.00      0.00       0      0.00      0.00  /usr/sbin/libvirtd
+        
+	###v11 with threads
+        # Time   UID      TGID       TID    %usr %system  %guest    %CPU   CPU  minflt/s  majflt/s     VSZ     RSS   %MEM   kB_rd/s   kB_wr/s kB_ccwr/s iodelay   cswch/s nvcswch/s  Command
+        # 1547852254     0         1         0    0.00    0.00    0.00    0.00     4      0.13      0.00  194520    7572   0.00      1.16     28.14      1.62      89      0.62      0.01  /usr/lib/systemd/systemd --switched-root --system --deserialize 22
+        # 1547852254     0         0         1    0.00    0.00    0.00    0.00     4      0.13      0.00  194520    7572   0.00      0.02      0.00      0.00      89      0.62      0.01  |__/usr/lib/systemd/systemd --switched-root --system --deserialize 22
+        #
+        if ( $line =~ /\s{0,1}(\d+)\s+\d+\s+(\S+|\S+\s+\S+)\s+(\d+\.\d+\s+\d+\.\d+\s+\d+\.\d+\s+\d+\.\d+\s+\d+\s+\d+\.\d+\s+\d+\.\d+\s+\d+\s+\d+\s+\d+\.\d+\s+\d+\.\d+\s+\d+\.\d+\s+\d+\.\d+\s+\d+\s+\d+\.\d+\s+\d+\.\d+\s+)(.*)/ ) {
```
mrsiano: we have nicer \ readable way to do this formatting in perl?

agent/tool-scripts/postprocess/pidstat-convert
```diff
@@ -0,0 +1,118 @@
...
+			# keep a copy of the real cmd, without the "/" conversions
+			printf { $file_handles{$pid_cmd} } "%s\n", $cmd;
+			select($file_handles{$pid_cmd});
+			$| = 1; # always flush
+			select(STDOUT);
+		}
+		if (exists $file_handles{$pid_cmd}) {
+			my $output = "";
+			my $changed = 0;
+			# save space by not writing duplicate values from subsequent samples
```
mrsiano: isn't it avoid dup by just appending to a file (since we clear the buffer each iter)?

agent/tool-scripts/postprocess/pidstat-convert
atheurer: In that regex we have repeated patterns like "\s+(\d+\.\d+)", and would have liked to add something like (\s+(\d+\.\d+)){4} to represent \s+(\d+\.\d+)\s+(\d+\.\d+)\s+(\d+\.\d+)\s+(\d+\.\d+), but I was not able to make that work because we can't have nested ()'s and the ()'s save to a variable $1 or $2 and so on, depending on their position.  Maybe someone more well versed in regex can suggest something

agent/tool-scripts/postprocess/pidstat-postprocess
```diff
@@ -70,88 +70,82 @@ if ( (($? >> 8) == 0) && (! -e "/usr/bin/vdsClient") && (-e "/usr/bin/virsh") )
...
-		$timestamps{$timestamp_ms}++;
+my %timestamps;
+my @attributes = get_pidstat_attributes();
+my $pids_dir = $dir . "/pids";
+if (opendir(my $pids_dh, $pids_dir) || die "Could not open $pids_dir") {
+        for my $pid (grep(/\d+/, readdir($pids_dh))) {
+		my $pid_dir = $pids_dir . "/" . $pid;
+		opendir(my $pid_dh, $pid_dir) || die "Could not open $pid_dir";
+		# A PID may be recycled over time as 1 process exits and another later is created,
+		# and therfore more than 1 process could have the same PID
```
ndokos: "threrefore"

agent/tool-scripts/postprocess/pidstat-postprocess
```diff
@@ -70,88 +70,82 @@ if ( (($? >> 8) == 0) && (! -e "/usr/bin/vdsClient") && (-e "/usr/bin/virsh") )
...
+					$cmd = $line;
+					chomp $cmd;
+					# convert qemu to a more readable label with VM name
+					if ( $cmd =~ /qemu/ and $cmd =~ /name\s(\S+).*/ ) {
+						$vm_name = $1;
+						$cmd = "KVMguest--" . $vm_name;
+					}
+					# modify to be appropriate as a label for a series in a graph
+					$cmd =~ s/%/%%/g; # avoid 'Invalid conversion in printf: "%=" ' errors if cmd contains a %.;
+					$cmd =~ s/[\'|\"]//g; # quotes cause problems everywhere
```
ndokos: This gets rid of '|' chars as well: is that intended?

atheurer: No, it was not.

agent/tool-scripts/postprocess/pidstat-postprocess
portante: Fixed in upcoming update...

agent/tool-scripts/postprocess/pidstat-postprocess
portante: But it seems like it is a good thing none-the-less.  So let's leave it there.

portante:
Fixed the last comment, shall we merge, @ndokos, @atheurer?

agent/tool-scripts/postprocess/pidstat-postprocess
ndokos: In that case, the comment should be fixed.

PR #1047, opened 2019-01-26
Ensure pbench-move/copy-results only works with metadata.log

portante:
Going forward, if a pbench result does not have a metadata.log file
present, then we are not going to try to handle that result. This
will avoid problems that arise from not being able to index the data
from the result to use in various user interfaces (Kibana, Grafana,
and the pbench UI itself).


portante:
@ndokos, @atheurer, please take a look, thanks!

atheurer:
Seems reasonable to me

Issue #1043, opened 2019-01-25
Modify pbench-user-benchmark to automatically run user's script piped into pbench-tool-trigger


portante:
Today, if a user wants to trigger tools, they have to know to run their script piped to `pbench-tool-trigger`.

Instead, `pbench-user-benchmark` could detect that tool triggers are registered and NOT start/stop tools directly, but instead pipe the output of the user's script to `pbench-tool-trigger` for them.

portante:
Fixed by #1049.

PR #1049, opened 2019-01-27
Fix pbench-tool-trigger to work as expected

portante:
We also fix pbench-specjbb2005 to use its new behavior, while enhancing
that benchmark to work in a non-destructive way by copying necessary
files into the benchmark run directory in /var/lib/pbench-agent.

We also enhance pbench-user-benchmark to accept a --use-tool-triggers
parameter that directs the script to properly invoke tool triggers
instead of traditional start/stop tools.

Along the way we fix up a number of unit tests, and the infrastructure
for those unit tests.


portante:
@atheurer, @ndokos, I broke this PR into three commits, the first two deal with fixes to the `bench-scripts` and `utils-scripts` test environment that don't involve changes to the behavior of `pbench-tool-trigger`.  This should make it a bit easier to see what changes are being made specifically for fixing triggers.

portante:
Will merge once @ndokos gets a chance to review as well.

agent/util-scripts/pbench-tool-trigger
```diff
@@ -39,31 +49,46 @@ while (<TRIG>) {
...
 	chomp;
 	($group, $start, $stop) = split(/:/);
-	printf "[$script]found trigger: group:$group start-trigger:$start stop-trigger:$stop\n";
-	$start_t{$group} = $start;
-	$started{$group} = 0;
-	$stop_t{$group} = $stop;
+	if ($group ne $tool_group) {
+		next;
+	}
+	if ($start eq "" || $stop eq "") {
```
ndokos: Do we need to check the trigger strings for colons? Or document a limitation?

agent/bench-scripts/pbench-specjbb2005
```diff
@@ -170,87 +179,207 @@ while true; do
...
+	exit 1
 fi
 
-if [ "$stop_warehouses" == "default" ]; then
-	stop_warehouses=`echo "$nr_cpus *2 /$nr_jvms" | bc`
+if [ "${nr_jvms}" == "node" ]; then
+	# Find out number of numa nodes
+	nr_jvms=$(ls -1d /sys/devices/system/node/node* | wc -l)
+fi
+if [ "${nr_jvms}" -eq "${nr_jvms}" ] 2>/dev/null ; then
```
ndokos: I was a bit baffled by this: I guess it (and the subsequent ones) tests that the variable's value is a number. If so, I think it deserves a comment. If not, it certainly deserves a comment!

portante: Sure I'll add comments.

agent/util-scripts/pbench-tool-trigger
portante: Posted an updated to catch it when registering.

portante:
Fixes #1049.

portante:
Fixes #1043.

agent/util-scripts/unittests
```diff
@@ -255,6 +257,13 @@ declare -A options=(
     [test-31]="--prefix=foo/bar --user=ndk"
     [test-32]="--help"
     [test-34]="--help"
+    [test-35]="--group=default --start-trigger=\"START:DEFAULT\" --stop-trigger=\"STOP DEFAULT\""
+    [test-35]="--group=default --start-trigger=\"START DEFAULT\" --stop-trigger=\"STOP:DEFAULT\""
+)
+
+declare -A expected_status=(
+    [test-05]=1
+    [test-06]=1
```
ndokos: Need to add test-35 with an expected status of 1 here I believe.

portante: Fixed.

agent/bench-scripts/pbench-specjbb2005
portante: Added comments.

PR #1051, opened 2019-01-27
CDM and pbench-run-benchmark

atheurer:
This includes two new major features which should be considered Beta:

1) `pbench-run-benchmark`, a universal benchmark script which uses config files to describe how a benchmark should be run.  This eliminates a benchmark specific script for execution, and only needs the config file and a post-processing file.  Support for `fio` and `trafficgen` is included in this PR.

2) Common Data Model for `pbench-run-benchmark`, where JSON documents suitable for Elasticsearch and adhering to this spec: https://github.com/atheurer/CommonDataModel.git

There are other features implemented in support of the two above:
- Library for CDM in perl
- Library for SSH/SCP in Perl, implemented with Ansible
- Library for base functions in Perl

If users want to beta test these features, after installation, run:
```
# pbench-run-benchmark fio --defaults=basic
```


portante:
@atheurer, this is reporting on GitHub as having 198 separate commits.  When I tried to rebase this on `master` it failed with conflicts.

It looks like the list of commits is repeated somehow.  I am going to break them out into two unique sets and post separate PRs for each so that we can review and tease this out.

portante:
Let's merge once @ndokos has a chance to review.

PR #1054, opened 2019-01-29
Use createBrowserHistory to manage session history

gurbirkalsi:
Removes the # between the host name and dashboard routes
by using the HTML5 History API.


gurbirkalsi:
The legacy browsers the module lists are known to be edge cases. For a list of supported browsers that can enable this API, see this link: https://caniuse.com/#feat=history

PR #1056, opened 2019-01-30
Re-factor pbench-uperf to reduce ssh/scp, log server

portante:
This is a re-factoring of `pbench-uperf` to create all the remote client/server
files locally and then copy them remotely via a tar ball, substantially
reducing the number of `ssh`/`scp` commands per host to one per client and
server during the test.

Further, when the server is run, the server `stderr`/`stdout` is captured to
a log file and collected with the run to help debug problems.

Additional error checking was added along the way as well in a number
of places.


portante:
Also created #1057 with just the changes to capture the uperf server logs.

atheurer:
@portante FWIW, we'll have Ansible/rsync available once we write our bash library.  I could accelerate that work if you want.

PR #1057, opened 2019-01-30
Capture uperf-server logs for debugging

portante:
We also now run the `pbench-uperf` client and server with the `-v` flag
to get verbose output.


portante:
@atheurer, what do you think about always running `uperf` with the `-v` flag?  It was really useful to figure out what was happening with a client that appeared not to be able to connect to the server.

portante:
@atheurer, can you officially "approve" this request as well?

atheurer:
I do t think I can do that unless I am an official reviewer. Can you add me and then I can approve

portante:
I think you are already added.  There should be a review button in the top right of your screen.

portante:
Just merging anyways.

Issue #1059, opened 2019-01-31
Using pbench-user-benchmark causes user's script to behavior incorrectly when stderr is redirected to a file


portante:
By default `pbench-user-benchmark` redirects both `stdout` and `stderr` to a file.  For certain scripts that expect access to the console via `stderr` this behavior of `pbench-user-benchmark` causes them to fail.

Consider adding an option like `--no-stderr-redirect` to `pbench-user-benchmark` to allow those use cases to work well.

portante:
Fixed by #1061.

PR #1061, opened 2019-01-31
Add --no-stderr-capture flag to pbench-user-benchmark

portante:
Sometimes users will have to run a program that can't tolerate capturing
stderr to a file for whatever reason.  The new flag, `--no-stderr-capture`,
gives the users control over that behavior.


portante:
Fixes #1059.

PR #1064, opened 2019-02-02
Fix perf pp tool

### LINKED ISSUE #1058: Processing perf tool data on a system with "lots" of CPUs take too long
Observed a simple 10 second pbench-user-benchmark run on a 208 CPU box (4 socket, 26 cores, HT enabled) where the processing of the perf data into per-CPU tar files tool close to 3 minutes (maybe more).

portante:
Fixes #1058.

We address the problem of generating per-CPU data taking too long by forking each CPU data generation into its own process to run at whatever level parallelism the host can afford.

We also:

 * Fix up the `tool-scripts/postprocess/gold/index.html` with the missing `.html` files
 * Fix the `jstack`/`jmap` unit tests to include the generated `.html` files
 * Address some minor comment cleanups in `perf-postprocess`
 * Provide a minimal fix to parallelize the perf per-CPU report generation
 * Provide a more complete refactoring of the `perf` tool to address both reduce the use of `pushd`/`popd` pairs

### DIFF
--- a/agent/base
+++ b/agent/base
@@ -48,12 +48,12 @@ TS_FORMAT="%FT%H:%M:%S"
 
 if [[ -z "$_PBENCH_BENCH_TESTS" ]]; then
     function timestamp {
-        # use ns in the timestamp
-        echo "$(date --utc +"${TS_FORMAT}.%N")"
+	# use ns in the timestamp
+	echo "$(date --utc +"${TS_FORMAT}.%N")"
     }
 else
     function timestamp {
-        echo "1900-01-01T00:00:00.000000"
+	echo "1900-01-01T00:00:00.000000"
     }
 fi
 # contains the base functions needed to use pbench
@@ -77,7 +77,7 @@ function error_log {
 function debug_log {
     debug_date=$(timestamp)
     if [ "$PBENCH_debug_mode" != "0" ]; then
-        echo "[debug][$debug_date] $*"
+	echo "[debug][$debug_date] $*"
     fi
     echo "[debug][$debug_date] $*" >> $pbench_log
 }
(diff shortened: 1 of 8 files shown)


atheurer:
FWIW, we used to fork to do this, but some users ran into memory exhaustion, and that's why we went back to single-threaded processing.  I am all for the forking if we move this post-processing to pbench-server.

portante:
@atheurer, @ndokos, okay this is ready for review.  It reworks the perf tooling to use "`perf archive`" and drops the per-CPU processing entirely since one can now pull the `perf.data` file locally and process it.

atheurer:
I am curious what the size of the archive is. Also, do we compress perf.data already?

atheurer:
@portante I would prefer the 'perf'archive' happen I the post-process phase, as I am not sure how much cpu that uses, and I don't think there's any loss of information my waiting until PP phase

portante:
> I am curious what the size of the archive is.

I have seen 16 MB files for the `.bz2` tar balls created by `perf archive`, but I would assume that could grow with many more symbols.

> Also, do we compress perf.data already?

Yes.

portante:
> I would prefer the `perf archive` happen I the post-process phase, as I am not sure how much cpu that uses, and I don't think there's any loss of information my waiting until PP phase.

I debated this while working on this change.

We need a "finalize tool collection" phase which is not really post-processing, but is the gathering of all the extra data needed for post-processing with the preparation for pulling data remotely for post-processing.

The current post-processing phase is a tad over-loaded.  The post-processing step is required to run on each of the hosts where the start/stop tools ran, performs transformations of the collected data on those hosts using host specific context in a few cases (`perf` being one of them), and then copies all the data back from the tool host to the controller host.

This archive step is not a data transformation at all, it is essentially configuration data needed for transformations.  It seemed putting this at the stop tools point is the best place for non-post-processing data collection.

But as you point out, if that becomes CPU intensive it could affect a benchmark.

Until we have a "finalize tool collection" phase I'll move it to the post-processing step.

atheurer:
I agree with you on the finalize part, but I wonder when it should be done: after a sample is complete, after an iteration is complete , or after all iterations are complete.  Ideally we would just wait to the end, as long as we don’t run into space issues.

portante:
Addressed the unit test failures.

atheurer:
OK, everything looks good to me

PR #1065, opened 2019-02-02
Fix mpstat dates

### LINKED ISSUE #1055: Handle timestamp rollover in mpstat tool
Below is an example of a timestamp rolling over from the last hour of the day to the next day.  We need to handle this correctly.

```
==> sample2/tools-default/svt_master_1_etcd_1:b06-h31-1029p/mpstat/mpstat-stdout.txt <==
Linux 3.10.0-957.1.3.el7.x86_64 (b06-h31-1029p.rdu.openstack.engineering.redhat.com)    01/09/19        _x86_64_        (64 CPU)

23:59:39     CPU    %usr   %nice    %sys %iowait    %irq   %soft  %steal  %guest  %gnice   %idle
00:01:39     all    0.34    0.00    0.29    0.00    0.00    0.01    0.00    0.00    0.00   99.35
```

portante:
Fixes #1055.

We address this issue by providing two commits.  The first adds a new unit test, `mpstat-1`, which contains an example mpstat output where the very first header is the last timestamp of the day, and the next timestamp in the data is in the following day, but 24 hour clock has wrapped.

The second commit shows how when we fix the timestamp handling we arrive at the correct timestamps.


agent/tool-scripts/postprocess/mpstat-postprocess
```diff
@@ -41,39 +41,31 @@ my %graph_type;
...
-	if ( $prev_time eq "" ) { # if this is the first time
-		$prev_time = $this_time;
-	}
-
 	# time format: HH:MM:SS
 	# 24-hour format,*no* AM/PM designation
-	my $hourminsec = $this_time;
-	$hourminsec =~ /(\d+):\d+:\d+\s?(\w\w)?/;
+	my $hourminsec = $_[0];
+	$hourminsec =~ /(\d+):(\d+):(\d+)/;
```
ndokos: Why did the original code worry about AM/PM designations? Why does the new code *not* have to worry about that?

portante: I don't know why it did because the line matching for timestamps did not appear to include a match for AM/PM.

ndokos:
Although I've approved these changes, I would like @atheurer to take a look and make sure that the current format is OK.

PR #1066, opened 2019-02-04
Native integration of jschart with pbench dashboard

gurbirkalsi:
Utilizes and depends on v1.0.4 of jschart for rendering
formatted json objects containing timeseries data.

Includes styling enhancements for responsiveness with
dashboard run comparison page.


portante:
@gurbirkalsi, can you post some before/after screen shots as well?

gurbirkalsi:
The .less file was removed because we are using inline JSX styling throughout the dashboard. I will be removing the other .less files tied to routes in a separate PR. Inline JSX styling allows us to take advantage of conditional styling especially in the case of React state changes.

gurbirkalsi:
Before jschart integration (using `timeseries.js` dependency): 

pdash

gurbirkalsi:
After jschart integration: 

pbench_dashboard

gurbirkalsi:
I've also removed the dependency being used for rendering the previous timeseries charts under `src/utils/timeseries.js` in addition to the co-dependencies under `package.json`

PR #1069, opened 2019-02-06
Parallelize the unit tests

portante:
When the server "`unittests`" command is run by default, all the tests will
run serially.  When given the option `--parallel`, then each unit test
will be run at the same time, with the output sorted by test order and
displayed when they are all finished.

Additionally, the outer "`run-unittests`" script will run each sub-system
unit tests in parallel, asking for server unit tests to run in parallel.

### DIFF
--- a/run-unittests
+++ b/run-unittests
@@ -3,10 +3,20 @@
 dir=$(dirname $0)
 let sts=0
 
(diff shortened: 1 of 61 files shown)


ndokos:
Two minutes of wall clock time on my laptop, 14m user + 1.5m system, so about a 7-fold gain - nice.

PR #1072, opened 2019-02-09
Metric fix for CDM: host instead of hostname

atheurer:
We had the wrong field name used (hostname) when creating metric documents


portante:
It seems the release 0.56 is in the can already.  We'd like to release 0.57 really close on its heals with some other fixes for indexing.  So let's just consider this for 0.57.

atheurer:
Great, thanks.

ndokos:
It did actually make into 0.56. Changed the milestone accordingly.

Issue #1034, opened 2019-01-21
network error from PDash on pbench-fio test result


bengland2:
I selected one of Elko's pbench test datasets , search for "b06-h31" and you'll find it, here's a screenshot
![screenshot from 2019-01-21 12-55-24](https://user-images.githubusercontent.com/1717107/51491331-0b0c0c80-1d7c-11e9-98d1-a1906f6b3f92.png)

And when I press the "Compare Results" button, I get a message box with title "network error", followed by:

"The selected iteration's resources are too large to parse in the browser. Please try another result."

I couldn't get the screenshot because the message box goes away.   I'm not sure why I'm getting this error.  Is it the wrong pbench version?  Is it that there isn't a clear specification for what is to be compared?  Or is it really just too much data?

gurbirkalsi:
@bengland2 Hi Ben, I tried searching for that result you selected under `/results` on production and wasn't able to find the result indexed under production - it was only indexed with elasticsearch. Since we don't index result.json file data in elasticsearch, we query them from the production server in order to build the comparison. 

I should update the error message to reflect that the result.json file was not found as a result of the 404.

gurbirkalsi:
@bengland2 Please disregard the last comment. This is an issue where the dashboard is not using the the prefix value for constructing the url to query against the production server. 

I'll be filing a PR shortly to resolve this issue so that we can fetch result.json data using the correct constructed url to production. For more details, please see the following discussion: https://github.com/distributed-system-analysis/pbench/issues/1034

gurbirkalsi:
I've filed a resolution for this issue at #1037

portante:
The fix from #1037 addresses the issue with pbench results have a "prefix" specified.  But once that issue is solved, the next issue brought up here is that the `result.json` files are too big to load into the browser.

We need to fix the server-side indexing to provide JSON documents in Elasticsearch that the dashboard can query without the need to ingest large JSON documents to find it.

gurbirkalsi:
Parsed Iteration JSON: https://gist.github.com/gurbirkalsi/23ab49de69666b4224625044eaa3e692

gurbirkalsi:
Parsed Column JSON: https://gist.github.com/gurbirkalsi/2d538d90424f87dcfe69e600ee3ed1bd

portante:
Seems to be fixed by #1073.

PR #1073, opened 2019-02-09
Update index document models to fix data problems

portante:
This is a significant refactoring of the index mappings for all document
types in order to fix problems with missing fields, adjust how
data is indexed for usability, correcting data types, fixing field name
conflicts, while updating various fields using proper tokenizers and/or
analyzers to ensure their data is indexed in a more natural or usable
way.

The driving factor for this work was the fact that there were no
result-data templates being created, and what we had in the code base
was completely wrong.  We replaced the existing but unused result-data
mapping with one which matches how documents are generated as of this
commit.  We add code to ensure that all templates are constructed and
pushed.

We add `_meta` version fields to each mapping in the templates. This
version is used in two ways:

  * We now query to see if the currently installed version is older,
    that is, has a version number less than the current version, and if
    so update it accordingly
  * As the version number input in index names

As a result of adding the version, the index names have been changed to
add a version number that matches the mapping in the template.  This
lets us easily re-index data using the new mapping, while allowing the
consumers of the index to use the new and old version indices at the
same time opportunistically. We now enforce the index prefix setting to
not contain the period character, since that separates the various
parts of the index name, <prefix>.<version>.<name>.<YYYY-MM-DD>.

While doing this work, we found a bug in discovering the tool data
files in the tar ball.  Sometimes the full hostname is being used
instead of the short hostname.  So now we look for tool data by short
and long hostname.

We turn off the use of the _all field for all document types, and add a
`default_field` setting for each type so that unspecified searches don't
touch every field.  This required us to add a "settings" infrastructure
per index name.

We fix the indexing for mpstat to emit floats, prevent duplicate mpstat
data, have prometheus-metrics emit fields in its namespace, have
proc-vmstat index similar to the way the proc-vmstat-postprocess works
(although we index both the original value and the computed rate), add
proper indexing of the proc-interrupts data, and generally correct many
data types.

In order to ensure and test all these changes, we re-worked the unit
test mock infrastructure to work with less specific interactions in
our normal code, moving it to its own module, and adding the ability
to compare generated JSON documents against the mappings to flag
problems. We also moved more Elasticsearch infrastructure into the
pbench library module.

We add a unit test, 7.17, to verify vmstat tool data indexing works.

Since we are re-working toc-entries in this effort, we take the
opportunity to add missing fields for directories (time stamp and mode),
ensuring all empty directories are indexed as well, and add the mtime
field for files. This is in preparation for viewing tar ball contents
via the JSON documents only.

When we get BadDate exceptions from processing the pbench tar ball we
will exit with that status.  But when we encounter these errors while
dealing with result and tool data, we have already indexed the run
document, and other tool data, so ending early is not too useful.  We
now continue to index so that we can report all the errors, indexing
what we can.

We add an "@idx" field to every document to record the "index" or the
offset of that document in the original source so that we can go back
and look up the original document a bit easier.  For result data, that
is the array offset for a particular object in the original JSON document,
while for tool data that is the row in a .csv file, or which timestamp
in a text file.

The uid calculations for result data have been fixed to replace
"controller_host" references with the controller field of the run
document.

In order to reduce a bit of space, result-data documents have had all
but their identifying fields stripped and placed into a parent "sample"
document.  This significantly reduces the size of the JSON documents
for each result value.  The parent sample document contains all of the
metadata from the benchmark.parameters section of the original
result.json file.

We have dropped the use of @metadata fields for all but the run
documents, where that namespace is reserved for metadata about the tar
ball which is rarely used. We then add three namespaces for result-data
and tool-data documents: run, iteration, and sample. This is used
consistently across all document types.


portante:
This is still a work-in-progress as I need to properly update all the templates.

But the big change is already in place: versioned index names.

@gurbirkalsi, we'll want to see about a change to the pbench dashboard to accommodate this.

portante:
I still have two items to work on for this effort, but this should be reviewable as is.  We might want to refactor this into smaller change sets, but for now it seems best to review in its current form.

portante:
Don't merge this until I have had a chance to run this in a live test environment.

.gitignore
```diff
@@ -2,3 +2,4 @@ doc/*.html
 __pycache__
 *.swp
 unittests.log
+*~
```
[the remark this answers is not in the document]
ndokos: Come to think of it, where does '*.swp' and 'unittests.log' get created?

server/lib/mappings/result-data.json
```diff
@@ -1,590 +1,189 @@
...
-                      "primary_metric": {
-                        "type": "string"
-                      },
-                      "benchmark_name": {
-                        "type": "string"
-                      },
-                      "max_stddevpct": {
-                        "type": "long"
-                      }
+                    "benchmark_name": {
```
portante: It seems we need to move some of this data to benchmark specific data sub-fields.

server/lib/mappings/result-data.json
```diff
@@ -1,590 +1,189 @@
...
-              },
-              "parameters": {
+            "@timestamp": {
+                "type": "date",
+                "format": "dateOptionalTime"
+            },
+            "timestamp_original": {
+                "type": "long"
+            },
+            "@metadata": {
```
portante: We should consider making iteration summary documents, which are "parents" to sample summary documents, which are in turn "parents" to individual sample documents.  Then for each sample document just have enough metadata to identify its "coordinates" (so to speak) in the universe of data, but leaving all the detailed metadata in the summary documents.

The thinking is that folks will browse the summary documents, but will only view the actual sample graphs when they know exactly which sample(s) to view from the summaries.

.gitignore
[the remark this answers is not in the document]
portante: For `unittests.log`, I often capture my unit test output to a log file of that name, so I figured ignoring it would avoid my accidentally checking it in.

As for `*.swp`, those show up when editing via the "other" editor (that is not named emacs).  When they show up as files that are not tracked, again, it is an attempt to be sure they don't get checked in accidentally.

Finally, the `*~` occurs from emacs backup files.

Does ignoring these three items cause a problem?

server/lib/pbench/mock.py
```diff
@@ -0,0 +1,221 @@
...
+class MockElasticsearch(object):
+    """A simple mock for the V1 Elasticsearch client object. We really just
+    duck-type it, only providing names for attributes we use.  It is not complete
+    by any stretch of the imagination.
+    """
+    def __init__(self, hosts, max_retries=None):
+        self.hosts = hosts
+        self.max_retries = max_retries
+        self.mpt = _MockPutTemplate()
+        self.msb = _MockStreamingBulk(15, self.mpt)
```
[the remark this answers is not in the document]
portante: Sure.

server/lib/pbench/mock.py
```diff
@@ -0,0 +1,221 @@
...
+                resp[action['_op_type']]['status'] = 200
+                ok = True
+            yield ok, resp
+        self.report()
+
+    def validate_type(self, action):
+        """Crude approach to validating the constructed dictionaries ahead of being
+        converted to JSON.
+
+        We first collect all the mappings that posted to templates, then
```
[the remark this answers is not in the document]
portante: Ah, yes, ...

.gitignore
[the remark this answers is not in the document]
ndokos: I don't know how I was misreading the '*~" one, but it finally dawned on me on the way home tonight - d'oh! 
No problem -  I was just wondering (and being thick as a brick about it ...)

server/lib/mappings/run.json
```diff
@@ -43,12 +50,59 @@
...
+                    "hostname-s": {
+                        "type": "string",
+                        "index": "not_analyzed"
+                    },
                     "label": {
                         "type": "string",
                         "index": "not_analyzed"
                     },
                     "tools": {
-                        "type": "nested"
```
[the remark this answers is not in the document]
portante: Not a problem.  Its parent `host_tools_info` is already nested.

server/bin/gold/test-7.17.txt
```diff
@@ -0,0 +1,2858 @@
...
+                "id": "13c4e0016fb79b8402a1ff86e0e6c413",
+                "iterations": "1",
+                "name": "pbench-user-benchmark_example-vmstat_2018.10.24T14.38.18",
+                "prefix": "test_glusterfs-storage-block",
+                "script": "pbench-user-benchmark",
+                "start_run": "2018-10-24T14:38:18.299722",
+                "tarball-dirname": "pbench-user-benchmark_example-vmstat_2018.10.24T14.38.18",
+                "tarball-toc-prefix": "pbench-user-benchmark_example-vmstat_2018.10.24T14.38.18"
+            },
+            "sosreports": []
```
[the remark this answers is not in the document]
portante: Fixed is next update.

server/lib/pbench/mock.py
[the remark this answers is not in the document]
portante: Removed the comment.  The description I was thinking of writing was actually written for the `_check_fields` function.

server/lib/pbench/mock.py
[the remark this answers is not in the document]
portante: Added a comment but did not add a meaningful name.  It is not the kind of constant we want to change because it impacts the unit test output (explained in the comment).  Is that sufficient?

portante:
This commit is intended to provide the proper result-data that will allow the UI to consume it from Elasticsearch, giving us a chance to properly fix #1034.

portante:
Depends on #1089, #1091, and #1092, and should be rebased on top of #1090, all of which were pulled out of this series of commits.

server/lib/mappings/result-data.json
portante: Moved all "benchmark" data, that taken from the `parameters.benchmark[0]` array entry to the "benchmark" namespace.

server/lib/mappings/result-data.json
portante: We emit a sample document, which is the parent of all the result data.  The shrinks the size of each result data document considerably.

portante:
@ndokos, I have rebased on this PR, and updated it to add support for indexing the satellite information, as well as the on-disk controller name to handle satellite tar balls.  I'll be posting a PR shortly which will update the UI to use that new name.

ndokos:
Reviewed everything *except* `index-pbench`: so far, so good. This needs rebasing (I got two easily resolved conflicts) and a couple of the gold files (7.0.1 and 7.17) need changes to accommodate the introduction of `pbench-server-activate-setup-web-server`.

portante:
@ndokos, I have already rebased the code, and added support for properly indexing satellite tar balls so that the UI can reference them properly.

portante:
@gurbirkalsi, the second commit shows how we can use the new `@metadata.controller_dir` field to construct URL references that will work for all tar balls that provide it.

server/bin/index-pbench
```diff
@@ -296,60 +206,69 @@ class PbenchData(object):
...
-}
-
-
 class ResultData(PbenchData):
+    """Benchmark result data: JSON files only.
+
+    Each pbench tar ball contains the result of running a benchmark (pbench-*
+    bench script or pbench-user-benchmark script wrapping a user provided
+    script), where N iterations of that benchmark were run, with each
+    iteration have N samples.
```
ndokos: having M samples.

server/bin/index-pbench
```diff
@@ -382,145 +299,272 @@ class ResultData(PbenchData):
         gen = self._make_source_json()
         return gen
 
+    # Table of data for supported benchmarks and what took look for in the
```
ndokos: took -> to

server/bin/index-pbench
```diff
@@ -362,12 +281,10 @@ class ResultData(PbenchData):
...
         return a list of dicts containing their metadata.
         """
-        paths = [x for x in ptb.tb.getnames() if x.find("result.json") >= 0 and ptb.tb.getmember(x).isfile()]
+        paths = [x for x in ptb.tb.getnames() if (os.path.basename(x) == "result.json") and ptb.tb.getmember(x).isfile()]
         datafiles = []
         for p in paths:
-            fname = os.path.basename(p)
-            datafile = _dict_const(path=p, basename=fname)
-            datafiles.append(datafile)
+            datafiles.append(os.path.dirname(p))
```
ndokos: So this is a list of directories where a "result.json" is found? Should the name of the function be changed?

server/bin/index-pbench
```diff
@@ -534,50 +578,135 @@ class ResultData(PbenchData):
...
+                                    ( 'name', sample_md['name'] ),
+                                    ( '@idx', sample_md['@idx'] ),
+                                    ( 'uid', sample_md['uid'] ),
+                                    ( 'measurement_type', sample_md['measurement_type'] ),
+                                    ( 'measurement_units', sample_md['measurement_units'] )
+                                ])
+                            sample_md['start'] = start_ts
+                            sample_md['end'] = end_ts
+                            # Now we can emit the sample document knowing the
+                            # ID of it's iteration parent document.
```
ndokos: its

server/bin/index-pbench
```diff
@@ -534,50 +578,135 @@ class ResultData(PbenchData):
...
-        )
-        yield action
+    for source, source_id, parent_id, doc_type in sources:
+        try:
+            idx_name = rd.generate_index_name('result-data', source)
+        except BadDate:
+            # We don't raise this exception because we are already well into
+            # indexing so much data that there is no point in stopping
+            # now. The source of the exception has already counted the
+            # exception it in its operational context.
```
ndokos: Extraneous "it".

server/bin/index-pbench
```diff
@@ -534,50 +578,135 @@ class ResultData(PbenchData):
...
+                                # Sample documents inherit the run and
+                                # iteration data of its parent.
+                                ( 'run', iteration['run'] ),
+                                ( 'iteration', iteration['iteration'] ),
+                                ( 'benchmark', iteration['benchmark'] ),
+                                ( 'sample', sample_md )
+                            ])
+                            sample_id = _make_source_id(source)
+                            yield source, None, 'sample'
+                            # Now we can yield each entry of the timeseries
```
ndokos: Tricky! I missed this the first time around :-)

server/bin/index-pbench
```diff
@@ -1145,51 +1400,81 @@ class ToolData(PbenchData):
...
+        }
+    }
+
+    def _stdout_keyval(self, file_object, converter):
+        """Process a line of a stdout key/value pair output file, building up the
+        record (dict) by adding each key/value pair found, associating them
+        with the previously encountered timestamp.  A record is yielded as one
+        JSON object when indexed.
 
         The format for files considered by this method are as follows:
```
ndokos: are  --> is

server/bin/index-pbench
```diff
@@ -1215,39 +1500,223 @@ class ToolData(PbenchData):
...
+                record[self.toolname]['rate'] = rate = _dict_const()
+            elif ts_orig is None:
+                # We have not encountered a timestamp yet, so ignore
+                # all lines until the first timestamp.
+                continue
+            else:
+                key, value = line.strip().split(' ')
+                parts = key.split('_', 1)
+                if len(parts) == 1:
+                    if remaps is not None:
```
ndokos: What is going on with `remaps`? I don't understand.

server/bin/index-pbench
```diff
@@ -2095,112 +2486,299 @@ class PbenchTarBall(object):
...
+        for m in self.members:
+            if m.name == metadata_log_path:
+                metadata_log_found = True
+            sampled_prefix = m.name.split(os.path.sep)[0]
+            if sampled_prefix != self.dirname:
+                # All members of the tarball should have self.dirname as its
+                # prefix.
+                raise UnsupportedTarballFormat(
+                    "{} - tarball member dir prefix should be \"{}\", but is"
+                    " \"{}\" instead for \"{}\".".format(
```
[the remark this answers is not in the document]
portante: Fixed.

server/bin/index-pbench
```diff
@@ -1,52 +1,68 @@
 #!/usr/bin/env python3
 # -*- mode: python -*-
 
-# prereqs:
-# config file: specified in CONFIG "Indexing" section
-# mapping file:
+"""Pbench indexing drriver, responsible for indexing a single pbench tar ball
```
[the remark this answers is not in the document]
portante: Fixed.

server/bin/index-pbench
```diff
@@ -1,52 +1,68 @@
...
-_VERSION_ = "0.7.0.0"
-_NAME_    = "index-pbench"
-_DEBUG    = 0
+from pbench import tstos, get_es, es_index, es_put_template, PbenchConfig, BadConfig, get_pbench_logger
+
+
+_NAME_ = "index-pbench"
+_PARENT_ = "pbench-index"
+
+# This is the version of this python code. We use the version number to mean
```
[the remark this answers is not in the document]
portante: I remove the blank line in the comment so that it is all one paragraph, and reworked the text to kill the run-on sentence.  Since all the other comments go before the item that talk about, it seems this makes it more uniform keeping it before `_VERSION_` is defined.

server/bin/index-pbench
```diff
@@ -1,52 +1,68 @@
...
+# the following:
+#   <major>.<minor>.<patch>
+#     major: change when:
+#       - moving between supported versions of Elasticsearch
+#       - incompatible changes occur with index name structure
+#       - when structure of index mappings change
+#     minor: change when:
+#       - code structure is changed
+#       - additional indexes are added without change existing indices or
```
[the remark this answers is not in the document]
portante: Fixed.

server/bin/index-pbench
portante: Fixed.

server/bin/index-pbench
portante: Fixed.

server/bin/index-pbench
portante: Fixed.

server/bin/index-pbench
portante: Fixed.

server/bin/index-pbench
portante: Fixed.

server/bin/index-pbench
```diff
@@ -805,29 +990,54 @@ _known_handlers = {
...
         ]
-    }
+    },
+    'prometheus-metrics': {
+        '@prospectus': {
+            'handling': 'json',
+            'method': 'json'
+        }
+    },
+    # The following tools should be processed vi JSON data generated from the
```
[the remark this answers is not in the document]
portante: Fixed.

server/bin/index-pbench
portante: Fixed.

server/bin/index-pbench
portante: I've added some comments to help explain this a bit.

server/bin/index-pbench
```diff
@@ -1386,10 +1867,10 @@ class ToolData(PbenchData):
                             break
                     else:
                         # Ignore .csv files for which we don't have a handler
-                        #_logger.warning("no .csv handler for '{}' ({}, {})\n",
+                        #self.logger.warning("no .csv handler for '{}' ({}, {})\n",
                         #        alias_name, fname, p)
                         continue
-            assert handler_rec is not None
+            assert handler_rec is not None, "Logic bomb! hander_rec is None"
```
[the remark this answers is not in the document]
portante: Fixed.

server/bin/index-pbench
```diff
@@ -2095,112 +2486,299 @@ class PbenchTarBall(object):
...
+        self.extracted_root = self.idxctx.options.tmpdir
+        self.tb.extractall(path=self.extracted_root)
         if not os.path.isdir(os.path.join(self.extracted_root, self.dirname)):
             self.delete_extracted()
             raise UnsupportedTarballFormat(
                 "{} - extracted tarball but can't get to the directory \"{}\".".format(
                     self.tbname,
                     os.path.join(self.extracted_root, self.dirname)))
+        # Open the MD5 file of the tarball and read the MD5 sum from it.
+        md5sum = open("%s.md5" % (self.tbname)).read().split()[0]
```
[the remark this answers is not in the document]
portante: Yes, which means it will raise an exception killing the PbenchTarBall() object construction at line 3157, resulting an exception that will be caught at line 3218, reported as an unexpected exception.  The `pbench-dispatch` could is supposed to have validated that we have a `.md5` file along side this tar ball, so it seems appropriate to leave it as a course-grain exception handling path.

server/bin/index-pbench
```diff
@@ -2095,112 +2486,299 @@ class PbenchTarBall(object):
...
+        # comparing date to the start run value, and adjusting it
+        # accordingly.  Since these are supposed to be close, we calculate
+        # an offset rounded to a half-hour and add it to the local
+        # date. We then convert back to an ISO format date.
+        offset = date_ts - self.start_run_ts
+        res = round(((float(offset.seconds) / 60) / 60) + (offset.days * 24), 1)
+        date_ts -= timedelta(0, int(res * 60 * 60), 0)
+        date = date_ts.isoformat()
+        # The pbench tar balls metadata.log file has two sections, "pbench"
+        # and "run" that we merge together a fix up.  We first pull the
```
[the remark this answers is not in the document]
portante: Fixed.

server/bin/index-pbench
```diff
@@ -617,60 +746,67 @@ _known_handlers = {
         'patterns': [
             {
                 'pattern': re.compile(r'disk_IOPS\.csv'),
-                'class': 'disk',
+                'class': None,
```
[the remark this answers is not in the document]
portante: Sure, working on it ...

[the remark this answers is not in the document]
portante: Updated.

portante:
@ndokos, updated with your latest round of comments fixed.

server/bin/index-pbench
```diff
@@ -1145,53 +1437,106 @@ class ToolData(PbenchData):
...
+    # has stat "pgrefill" with a value, the other has stat "pgrefill" as an
+    # object with substats "foo" and "bar":
+    #     { "pgrefill": 1000 }
+    #     { "pgrefill": { "foo": 20, "bar": 10 } }
+    # Both of those documents can't land in the same index because the
+    # mappings for "pgrefill" are radically different.
+    #
+    # So this table gives us a place to rename the key without a substat to
+    # avoid the conflict.
+    _remaps = {
```
ndokos: Thanks for the comment! That makes things much clearer.

server/bin/index-pbench
```diff
@@ -1145,53 +1437,106 @@ class ToolData(PbenchData):
...
+                yield datum, source_id
+                idx += 1
         return
 
-    def _make_source_stdout(self):
-        """
-        Read the given set of files one at a time, emitting a record
-        for each set of key/value pairs associated with a timestamp.
+    # For some tools, proc-vmstat being the first case we encounter this,
+    # depending on the version tool run and the version of the OS on which the
```
ndokos: the version of the tool run...

server/bin/index-pbench
```diff
@@ -1145,53 +1437,106 @@ class ToolData(PbenchData):
...
+    # tool is run, the key names emitted can change version to version.  When
+    # they do, sometimes this can cause a mapping conflict definition of the
```
ndokos: "definition" should be excised, maybe?

server/bin/index-pbench
```diff
@@ -2095,112 +2550,299 @@ class PbenchTarBall(object):
...
+        ])
 
-        This is offered as a service once here when processing a tar ball
-        so that any code needing the start or end of a run has a timestamp
-        in a form they can use that is consistent with all the other uses.
+    def gen_toc(self):
+        """Generate (t)able (o)f (c)ontents JSON documents for this tarball.
+
+        There should be one JSON document emitted per directory entry found in
+        the tarball. Each non-directory entry found is add to the given
```
ndokos: added

server/bin/index-pbench
```diff
@@ -2095,112 +2550,299 @@ class PbenchTarBall(object):
...
+        prefix_l = len(self.dirname)
+        # Note we have to fully populate the dictionary by processing all the
+        # members before we can yield the generated sources for each
+        # directory.
+        toc_dirs = _dict_const()
+        for m in self.members:
+            # Always strip the prefix
+            name = m.name[prefix_l:]
+            if m.isdir():
+                dname = name if name.endswith(os.path.sep) else "{}/".format(name)
```
ndokos: For consistency, use '/' in both places or os.path.sep in both.

server/bin/index-pbench
```diff
@@ -2095,112 +2550,299 @@ class PbenchTarBall(object):
...
-                    return dt, dt.isoformat()
+                    ftype = self._mode_table[m.type]
+                except KeyError:
+                    ftype = "unk"
+                fentry['type'] = ftype
+                if m.issym():
+                    fentry['linkpath'] = m.linkpath
+                dname = os.path.dirname(name)
+                if not dname.endswith(os.path.sep):
+                    dname = "{}/".format(dname)
```
ndokos: Ditto.

server/bin/index-pbench
```diff
@@ -2212,79 +2854,306 @@ class PbenchTarBall(object):
...
-            raise Exception('Missing start_run or end_run elements')
-        return mdfconfig
-
 
-def get_es_hosts(config):
+class IdxContext(object):
     """
-    Return list of dicts (a single dict for now) -
-    that's what ES is expecting.
+    The general indexing options, including confiuration and other external
```
ndokos: configuration

server/bin/index-pbench
```diff
@@ -2293,128 +3162,102 @@ def main(options, args):
...
         # dictionary is passed along to each generator so that it can add its
         # context for error handling to the list.
-        actions = make_all_actions(
-                ptb, options, INDEX_PREFIX, _index_patterns, opctx)
+        actions = make_all_actions(ptb, idxctx)
 
-        # Create the various index templates
-        if options.debug_time_operations: _ts("es_template")
-        es_template(es, options, INDEX_PREFIX, INDEX_VERSION, config, dbg=_DEBUG)
+        # Now that we are ready to being the actual indexing step, ensure we
```
ndokos: begin

server/bin/index-pbench
```diff
@@ -2470,10 +3316,12 @@ if __name__ == '__main__':
     parser.set_defaults(tmpdir = os.environ.get("TMPDIR"))
     parsed = parser.parse_args()
 
-    if parsed.dump_index_patterns:
+    if parsed.dump_index_patterns or parsed.dump_templates:
```
ndokos: -I and -Q are mutually exclusive in the implementation: that should be good enough, but I just wanted to make sure that you are OK with that.

gurbirkalsi:
>@gurbirkalsi, the second commit shows how we can use the new `@metadata.controller_dir` field to construct URL references that will work for all tar balls that provide it.

@portante what is the satellite field referring to? 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

portante:
@gurbirkalsi, the concept of a satellite pbench server is used to implement a "pull" mechanism for pbench results instead of every client "pushing" to the central server.  We have a satellite server in AWS (given the satellite name of "EC2") outside the Red Hat firewall that external clients can push to.  Then the internal pbench server "pulls" from that satellite pbench server instance.

All controller names from the satellite pbench server are prefixed with "`<satellite name>::`" on the central pbench server so that the on-disk file system conflicts are avoided.

For example, suppose we have a pbench satellite server named "GOO".  If pbench controller is named "`10.0.1.10`" externally, on disk, the controller name would be "`10.0.1.10`", and so a `result.json` file might be referenced as, `http://pbench.aws.example.com/incoming/10.0.1.10/pbench-user-benchmark_fooconfig_2019.03.08T10.10.10/result.json`.  Now if the internal server also has a controller named "`10.0.1.10`", then when the data from the remote satellite is pulled locally, they would be placed under the same controller.

Instead, we use the satellite name as a prefix when we copy the controller locally.  So in the above example, to reference the same `result.json` file, you'd have to do: `http://pbench.internal.com/incoming/GOO::10.0.1.10/pbench-user-benchmark_fooconfig_2019.03.08T10.10.10/result.json`, where we now have added the "`GOO::`" prefix to it.

The new `controller_dir` field always contains the on-disk name of the controller directory so that the UI can properly construct the path without knowing about the satellite prefix.

server/bin/index-pbench
```diff
@@ -2212,79 +2854,306 @@ class PbenchTarBall(object):
...
+        if len(keys) != 1:
+            raise MappingFileError("Invalid mapping file: %s" % mapping_fn)
+        return keys[0], mapping[keys[0]]
+
+    def _load_templates(self):
+        # where to find the mappings
+        MAPPING_DIR = os.path.join(
+            os.path.dirname(os.path.dirname(os.path.abspath(sys.argv[0]))),
+            'lib', 'mappings')
+        # where to find the mappings
```
ndokos: settings

server/bin/index-pbench
portante: Fixed.

server/bin/index-pbench
portante: Fixed.

server/bin/index-pbench
portante: Fixed.

server/bin/index-pbench
portante: Fixed.

server/bin/index-pbench
portante: Fixed.

server/bin/index-pbench
portante: Fixed.

server/bin/index-pbench
portante: Fixed.

server/bin/index-pbench
portante: Well, nothing enforces the mutual exclusion, other than `-I` is processed first, and then `-Q` is checked for.

server/bin/index-pbench
portante: Fixed.

portante:
@gurbirkalsi, @ndokos, all comments addressed to my knowledge.

server/bin/index-pbench
ndokos: The problem is that if you specify both, only the -I will be active, because in main(), we deal with that first and then return. So I misspoke when I said "muitually exclusive": it's just that "-I -Q" or "-Q -I" is equivalent to "-I".

server/bin/index-pbench
```diff
@@ -296,231 +205,365 @@ class PbenchData(object):
...
-}
-
-
 class ResultData(PbenchData):
+    """Benchmark result data: JSON files only.
+
+    Each pbench tar ball contains the result of running a benchmark (pbench-*
+    bench script or pbench-user-benchmark script wrapping a user provided
+    script), where N iterations of that benchmark were run, with each
+    iteration have M samples.
```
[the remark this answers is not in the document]
portante: Fixed.

server/bin/index-pbench
```diff
@@ -534,79 +577,201 @@ class ResultData(PbenchData):
...
-_known_handlers = {
+# The handler record dictionaries in "patterns" list contain a number of
+# fields:
+#   required:
+#   --------
+#     "pattern":   a regular expression string pattern to match files tool
+#                  data files
+#   optional:
+#   --------
+#     "converter":    they data type use for the value, "int" or "float"
```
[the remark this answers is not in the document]
portante: Fixed.

server/bin/index-pbench
```diff
@@ -534,79 +577,201 @@ class ResultData(PbenchData):
...
+#     "class":        a given tool might have different "classes" of data
+#                     recorded, e.g. CPU, disk, memory, process, etc.
+#     "metric":       the metric name for the data in a file, e.g. all
+#                     data in a .csv file might be for IOPS (read and
+#                     write).
+#     "colpat":       for some .csv files, each column header encodes the
+#                     identifier of the entity the metric data is about, and
+#                     optionally has sub-fields; this field contains the
+#                     regular expression pattern for the column header
+#     "subfields":    there are some columns that record the metric data is
```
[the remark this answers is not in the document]
portante: Fixed.

server/bin/index-pbench
```diff
@@ -805,29 +1026,55 @@ _known_handlers = {
...
-    }
+    },
+    'prometheus-metrics': {
+        '@prospectus': {
+            'handling': 'json',
+            'method': 'json'
+        }
+    },
+    # The following tools should be processed via JSON data generated from the
+    # sysstat tool suite itself, which currently don't handle.
```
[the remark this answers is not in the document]
portante: Fixed.

server/bin/index-pbench
portante: Right, it seems we agree on the existing behavior.  Are you look for a change to this `if` statement to make it check for one and then the other, or just an acknowledgement that it is okay to leave it as is?

server/bin/index-pbench
```diff
@@ -534,79 +577,200 @@ class ResultData(PbenchData):
...
 #
-# The "colpat" field, or column pattern, describes as a regular expression the
-# constituent components of the .csv file column header.  Currently, the "id"
-# and "subfield" match groups are the only two expected groups, with
-# "subfield" considered optional.
+# For the "method" field, we support:
+#   "individual":         used with "csv" handling, indicates the rows from
+#                         each individual .csv file form one document with
+#                         the columns as field names
+#   "unifiy":             used with "csv" handling, indicates the rows from
```
ndokos: "unify"

server/bin/index-pbench
ndokos: The latter.

server/bin/index-pbench
portante: Fixed.

server/bin/index-pbench
portante: Great, for posterity, just noting that we talked and we'll leave it as is for now.

portante:
Applied latest typ-o fix.

portante:
Changes look good on staging, so once we travis job is re-run I'll merge it.  Thanks for the reviews everyone!

ndokos:
@portante: I restarted two jobs that failed spuriously: one seems to have passed, I guess the other failed again and you restarted it?

portante:
@ndokos, I restarted that last one that failed.

PR #976, opened 2018-11-02
Rewrite pbench-backup-tarballs and pbench-verify-backup-tarballs to support S3

tenstormavi:
Rewrite pbench-backup-tarballs and pbench-verify-backup-tarballs to support S3.

We are enhancing the backup strategy to support 2 backup destinations: a local NFS mount directory, and an S3 object store.

The approach taken is to re-write `pbench-backup-tarballs.sh` and `pbench-verify-backup-tarballs.sh` in Python (renamed to `.py` files).

The backup operation now only checks the `TO-BACKUP` state directory for new results that need to be backed up, and figures out which of those results need to be copied to the NFS mount and / or the S3 object store.

The verify backup operation ensures the local and NFS tar ball directories have no bit-rot (verifies the `.md5` stored checksums match the recalculated checksum of each tar ball), and then reports any tar balls that are not backed up in the NFS or S3 object store.

We add a new configuration section for the S3 object store in the pbench-server config file.  And we add a new module for handling the S3 configuration information and S3 interface behavior.

Fixes #923.

PR #1089, opened 2019-02-21
Add pbench python logger

portante:


portante:
Foundational work for #976 and #1036.

PR #1090, opened 2019-02-21
Remove spaces from field names

portante:
Unfortunately, we had to update the samples by hand since the new code
expects them removed.


portante:
Taken from #1073, the server is currently fixing up field names with spaces in them that it knows about, but this PR attempts to fix this at the source.

portante:
@atheurer, good question, I'll check and see.

portante:
Yes, I'll post a PR to fix those references to handle both.

portante:
PR #1098 fixes the UI to handle field names with both "closest sample" and "closest_sample".

Issue #979, opened 2018-11-05
Fix pbench-index/index-pbench to remove unpacked tar ball hierarchy after it finishes indexing


portante:
It appears that `pbench-index` or `index-pbench` is unpacking all the tar balls it has to work with but not cleaning up the unpacked directory as it goes.  This leads to an explosion of disk usage that is unnecessary, and with large tar balls can be problematic.

ndokos:
That happens (or *should* happen) through the TMPDIR env variable: index-pbench unpacks the tarball in there to avoid keeping it all in memory IIRC. But pbench-index *should* be cleaning that up through a trap.

portante:
Doesn't that `trap` execute when all the tar balls have been processed?  So if there are 100 tar balls to be indexed, won't all 100 get unpacked before they are all cleaned up by the 1 `trap` at the end?

ndokos:
Yes, that's right: we'll have to get `index-pbench` to delete the unpacked tarball at the end.

portante:
This will be addressed by #1073.

portante:
We also need to ensure that extracted read-only files don't prevent removal of extracted tar ball hierarchy.

portante:
Fixed as of #1091.

PR #1091, opened 2019-02-22
Ensure indexing properly cleans up extracted tar balls

portante:
Prior to this change, if the indexing job has more than one tar ball to
index, index-pbench will extract each tar ball it works on, leaving the
extracted contents to be cleaned up by pbench-index. This works well for
one or two tar balls, but if we have a large set of tar balls to work
through this can be problematic.

This change makes it so that index-pbench properly cleans up the
extracted tar ball before exiting.  We modify pbench-index to give a
dedicated sub-directory to index-pbench so that we can detect problems.
We also modify the unit test harness to detect when any server script
leaves data around in its given TMPDIR.

### DIFF
--- a/server/bin/pbench-index
+++ b/server/bin/pbench-index
@@ -70,33 +70,32 @@ linksrc=TO-INDEX
 linkdest=INDEXED
 linkerrdest=WONT-INDEX
 
-# get the list of files we'll be operating on
-#list=$(ls $ARCHIVE/*/$linksrc/*.tar.xz 2>/dev/null)
-# Better: avoids long command line problems and sorts by size
-TMPDIR=$TMP/pbench-index.$$
-# index-pbench need to get at it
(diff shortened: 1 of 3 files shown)


portante:
Fixes #979.

server/bin/pbench-index
```diff
@@ -148,6 +151,9 @@ while read size result ;do
         > $errors_json
     fi
 
+    # Exit if index-pbench fails to clean up properly.
```
ndokos: Why exit? Why not continue with  the next tarball? After all, we are going to wait a minute and do the next tarball when pbench-index is run again in any case - do we expect any difference?

portante: So it seems bad in principle to keep processing data when we know we have encountered a problem.  Since this code can't control how it is invoked, we need to solve the larger problem with the cron jobs another way.

Issue #982, opened 2018-11-07
Need to speed up `index-pbench` as much as possible


portante:
Possible approaches to speeding up `index-pbench`:

 * process each tool data file in a separate parallel sub-process
 * review and profile the code to find bottle-necks and hot-spots
 * parallelize `pbench-index` to run the indexing of each tar ball in parallel

Other ideas?

portante:
Partial improvement comes with #1092.

portante:
As of #1306, we have separate indexing jobs for run/toc/result data and tool data, allowing the tool data to be indexed at its own rate.

portante:
While we closed the work in PR #1381, there are some ideas there that we should consider well for the future.

Defining a fast algorithm for generating a unique, repeatable ID from the data is worth following up on when this work is picked up again.

PR #1092, opened 2019-02-22
Use the already extracted sosreports

portante:
Since index-pbench already extracts the entire pbench tar ball into a
temporary directory, instead of using the tarfile object's "extractfile"
method, which creates an in-memory copy of the file, use the extract
copy.


portante:
Part of the fix for #982.

PR #1096, opened 2019-02-22
Update base URL paths for controller and search routes

gurbirkalsi:
Previously this was causing issues in routing and retrieving the
dashboard config file due to missing route name or slash.


gurbirkalsi:
These changes have been tested and deployed on my UI VM under `/dashboard/`

PR #1098, opened 2019-02-24
Fix UI 'closest sample' references

portante:
Handle removed spaces in result.json field names, which is only "closest sample"
for now.

Enables the UI to work with PR #1090.


portante:
@gurbirkalsi, this is ready to be reviewed now.

PR #1100, opened 2019-02-25
Don't require pbench-server 'environment' option

portante:
The environment option is really optional, so don't bother enforcing its
presence.

For the others, make sure we just raise BadConfig() exception as we
don't need a stack trace.


portante:
Pulled from the first commit of #1073.

PR #1105, opened 2019-02-27
Ansible playbook for installing key and config files

ndokos:


agent/ansible/agent-config/pbench-agent-config.yml
```diff
@@ -0,0 +1,13 @@
...
+- name: install pbench.repo
+  hosts: servers
+  remote_user: root
+  become: yes
+  become_user: root
+
+  roles:
+      - pbench-agent-install-key
+      - pbench-agent-install-config
+
```
portante: Can we remove the extra blank lines?

ndokos: Done

PR #1107, opened 2019-02-28
Some minor behavior changes for field names and use of incoming URL

portante:
* Handle both new and old field names (requires PR #1073 for new field names)
   * Adding some error handling when we don't get our expected data
 * Strip trailing slashes in the configured URLs in case somebody makes a mistake
 * Instead of using the configured prefix, use the `incoming` URL which does not need the `run.prefix` value

### DIFF
--- a/web-server/v0.4/src/pages/Dashboard/RunComparison.js
+++ b/web-server/v0.4/src/pages/Dashboard/RunComparison.js
@@ -221,7 +221,7 @@ export default class RunComparison extends ReactJS.Component {
           iterationRequests.push(
             axios.get(
               datastoreConfig.results +
-              '/results/' +
+              '/incoming/' +
               encodeURIComponent(
                 clusteredIterations[primaryMetric][cluster][iteration].controller_name
               ) +
--- a/web-server/v0.4/README.md
+++ b/web-server/v0.4/README.md
@@ -1,5 +1,6 @@
 ## Foreword
 
+
 Pbench Dashboard is a web-based platform for consuming indexed performance benchmark data. The platform provides a consolidated view of benchmark data within tables, charts, and other powerful visualizations. Users are able to quickly navigate through benchmark data and tune analytics through comparison tools for in-depth analysis.
 
 ## Scaffolding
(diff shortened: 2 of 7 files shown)


gurbirkalsi:
Do we need to update the field name to `run.start_run` at line 38 under `services/dashboard.js` for the pre-v1 run mapping version? Currently the last modified column in the controller view contains null data due to this.

portante:
@gurbirkalsi, yes that is a good catch.  I'll update the commit shortly.

portante:
@gurbirkalsi, fixed, please take a look.

gurbirkalsi:
We'll also want to add the `runs_preV1` field name to the controllers query to maintain the ordered list of controllers by their last modified date. 

https://github.com/distributed-system-analysis/pbench/blob/fd429caf9a5af2602fbb18580915797639aeba06/web-server/v0.4/src/services/dashboard.js#L32

portante:
@gurbirkalsi, fixed the aggregation to order by both sub-aggregations, and then fixed the "Results" page to that sorting by name works properly (ascending was descending and vice-versa) and that sorting by start time worked properly.

web-server/v0.4/src/pages/Dashboard/RunComparison.js
```diff
@@ -221,7 +221,7 @@ export default class RunComparison extends ReactJS.Component {
           iterationRequests.push(
             axios.get(
               datastoreConfig.results +
-              '/results/' +
+              '/incoming/' +
```
gurbirkalsi: The incoming endpoint fetches empty response data for the timeseries graphs under the 'All' tab in the run comparison page. This ends up rendering empty timeseries charts within the tab. Do we have the result.json files for sample data at this endpoint?

portante: @gurbirkalsi is this for all result data?  The `/results/` path ends up being a link to the `/incoming/` path, so if you have a particular example I can debug what is going on.

gurbirkalsi: For some reason my browser was blocking the resources earlier - it worked once I reset my CORS extension.

gurbirkalsi:
Thanks for making these changes @portante - this PR looks solid and is ready for a merge! :rocket:

PR #1108, opened 2019-02-28
Add bpftrace to pbench-agent-tools

sourabhtk37:
## Todo
- [x] find where bpftrace is installed: `/usr/bin/` or `/usr/local/bin`
- [x] Check for required packages and kernel headers for bpftrace to run 
---- 
This is for tracking purpose. I used the existing jmap tool to base this. Need your inputs/reviews on implementing the post-processing and other features needed for pbench. 

Here's the sample stdout from a run, which uses a basic bpftrace program, [syscount.bt](https://github.com/iovisor/bpftrace/blob/master/tools/syscount.bt): 
```
# cat /var/lib/pbench-agent/iozone__2019.02.26T16.23.27/1/reference-result/tools-default/localhost/bpftrace/bpftrace-stdout.txt
Attaching 3 probes...
Counting syscalls... Hit Ctrl-C to end.
 
Top 10 syscalls IDs:
@syscall[202]: 82215
@syscall[186]: 157832
@syscall[232]: 176516
@syscall[38]: 183350
@syscall[7]: 228544
@syscall[20]: 253303
@syscall[8]: 284114
@syscall[1]: 311635
@syscall[0]: 465967
@syscall[47]: 682795
 
Top 10 processes:
@process[grep]: 72991
@process[cut]: 75418
@process[Chrome_IOThread]: 89515
@process[Chrome_ChildIOT]: 100151
@process[code]: 167427
@process[sh]: 358588
@process[chromium-browse]: 365207
@process[xprop]: 622420
@process[Xorg]: 669920
@process[iozone]: 921651
```

Signed-off-by: T K Sourab <sourabhtk37@gmail.com>

### DIFF
--- a/agent/tool-scripts/bpftrace
+++ b/agent/tool-scripts/bpftrace
@@ -1,0 +1,172 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: t; sh-basic-offset: 8; sh-indentation: 8; sh-indent-for-case-alt: + -*-
+
+# Example usage: pbench-register-tool --name=bpftrace -- --script=sample_prog.bt
+# Which becomes, after processing: /usr/bin/bpftrace sample_prog.bt
+
+script_path=$(dirname $0)
+script_name=$(basename $0)
+pbench_bin="$(realpath -e ${script_path}/..)"
+
+# source the base script
+. "$pbench_bin"/base
+
+# Pbench tool scripts must provide the following functions
+# 1) Install the tool
+# 2) Start data collection
+# 3) Stop data collection
+# 4) post-process the data
+
+# Defaults
+tool=bpftrace
+tool_bin=/usr/bin/$tool
+group=default
+iteration=""
+dir=""
+script=""
+mode=""
+
+function usage {
+	printf -- "The following options are available: \n\n"
+	printf -- "\t--install                 install this pbench tool\n"
+	printf -- "\n"
+	printf -- "\t--start|stop|postprocess  start/stop/postprocess the data collection\n"
+	printf -- "\t--iteration=int           the iteration (required)\n"
+	printf -- "\t--group=str               the pbench tool group (optional)\n"
+	printf -- "\t--dir=str                 directory to store data collection (required)\n"
+	printf -- "\t--script=str              path to bpftrace script (required w/ --start)\n"
+}
+
+opts=$(getopt -q -o dgi --longoptions "dir:,script:,group:,iteration:,start,stop,install,postprocess" -n "getopt.sh" -- "$@")
+if [[ $? -ne 0 ]]; then
+	printf -- "\n" >$2
+	printf -- "$script_name: you specified an invalid option\n\n" >$2
+	usage
+	exit 1
+fi
+eval set -- "$opts"
+while true; do
+	case "$1" in
+	--install)
+		mode="install"
+		shift
+		;;
+	--start)
+		mode="start"
+		shift
+		;;
+	--stop)
+		mode="stop"
+		shift
+		;;
+	--postprocess)
+		mode="postprocess"
+		shift
+		;;
+	-d | --dir)
+		shift
+		if [[ -n "$1" ]]; then
+			dir="$1"
+			shift
+		fi
+		;;
+	-g | --group)
+		shift
+		if [[ -n "$1" ]]; then
+			group="$1"
+			shift
+		fi
+		;;
+	-i | --iteration)
+		shift
+		if [[ -n "$1" ]]; then
+			iteration="$1"
+			shift
+		fi
+		;;
+	--script)
+		shift
+		if [[ -n "$1" ]]; then
+			script=$1
+			shift
+		fi
+		;;
+	--)
+		shift
+		break
+		;;
+	*)
+		printf -- "$script_name:Error: Invalid command line options\n\n" >&2
+		usage
+		exit 1
+	esac
+done
+
+if [[ -z $mode ]]; then
+	printf -- "$script_name:Error: one of the following options is required, --install|--start|--stop|--postprocess\n\n" >&2
+	usage
+	exit 1
+fi
+
+if [[ "$mode" != "install" ]]; then
+	if [[ -z $dir || ! -d "$dir" ]]; then
+		printf -- "$script_name:Error: --dir argument is required with a valid directory\n\n" >&2
+		usage
+		exit 1
+	fi
+	if [[ -z $iteration ]]; then
+		printf -- "$script_name:Error: --iteration argument is required\n\n" >&2
+		usage
+		exit 1
+	fi
+fi
+
+if [[ "$mode" = "start" ]]; then
+	if [[ -z $script ]]; then
+		printf -- "$script_name:Error: --script argument is required for --start\n\n" >&2
+		usage
+		exit 1
+	else
+		script=$(echo $script | sed 's/^"\(.*\)"$/\1/')
+	fi
+fi
+
+tool_dir="$dir/tools-$group"
+tool_output_dir=$tool_dir/$tool # all tools keep data in their tool specific dir
+tool_pid_file=$pbench_tmp/$group.$iteration.$tool.pid
+
+case "$mode" in
+install)
+	check_install_rpm bpftrace || exit 1
+	;;
+start)
+	tool_cmd="$tool_bin $script"
+	tool_cmd_file="$tool_output_dir/$tool.cmd"
+	tool_stdout_file=$tool_output_dir/$tool-stdout.txt
+	tool_stderr_file=$tool_output_dir/$tool-stderr.txt
+	mkdir -p $tool_output_dir
+	if [[ $? -ne 0 ]]; then
+		error_log "$script_name: failed to create tool output directory, $tool_output_dir" >&2
+		exit 1
+	fi
+	echo "$tool_cmd" >$tool_cmd_file
+	debug_log "$script_name: running $tool_cmd"
+	$tool_cmd >"$tool_stdout_file" 2>"$tool_stderr_file" &
+	echo $! >$tool_pid_file
+	wait
+	;;
+stop)
+	if [[ -s "$tool_pid_file" ]]; then
+		pid=$(cat "$tool_pid_file")
+		debug_log "stopping $script_name"
+		kill -SIGINT $pid && /bin/rm "$tool_pid_file"
+	else
+		warn_log "[$script_name]: tool is not running, nothing to kill"
+	fi
+	;;
+postprocess)
+	;;
+
+esac
+
+exit 0


atheurer:
1) Can you provide examples of how the bpftrace tool was registered and how you started/stopped the tool (using a pbench benchmark or running pbench-start-tools or pbench-stop-tools)?

2) Does syscount.bt require compiling before using?  If so, please see how this is handled for systemtap for pbench (the stap code is compiled during --install, which what pbench-register-tool calls), as bpf might need to do this in a similar way. 

3) The --install will need to check for required packages (not cmds) and install if not found (using check_install_rpm)

agent/tool-scripts/bpftrace
```diff
@@ -0,0 +1,123 @@
...
+		error_log "$tool_bin not found, exiting"
+		exit 1
+	fi
+	;;
+	start)
+	mkdir -p $tool_output_dir
+	echo "$script_name: running $tool_cmd"
+	tool_stdout_file=$tool_output_dir/$tool-stdout.txt
+	tool_stderr_file=$tool_output_dir/$tool-stderr.txt
+	tool_cmd_file="$tool_output_dir/$tool.cmd"
```
portante: Where is this used?

Seems like the pattern for other tools is to emit the full command line that will be used for the tool into the `$tool_cmd_file`.

agent/tool-scripts/bpftrace
```diff
@@ -0,0 +1,123 @@
...
+    echo "Running"
+	$tool_cmd  > "$tool_stdout_file" 2> "$tool_stderr_file" &
+    wait
+	;;
+	stop)
+    echo "Killing the process"
+	kill -SIGINT $(pidof bpftrace)
```
portante: Instead of using `pidof`, at line 115, can you capture the pid of the tool into a file, like:

```
$tool_cmd  > "$tool_stdout_file" 2> "$tool_stderr_file" &
echo ${!} > ${pbench_tmp}/${tool}.pid
wait
```

And then kill it using:

```
kill -SIGINT $(cat ${pbench_tmp}/${tool}.pid)
```

Or something like the above that works.  Can you find an example of this pattern in the `sar` tool?

agent/tool-scripts/bpftrace
```diff
@@ -0,0 +1,123 @@
...
+
+function usage() {
+	echo "
+	USAGE:
+	$script_name: you specified an invalid option
+	The following options are available:
+	--install,			          install this perf tool
+	--start|stop|postprocess	start/stop/post-process the data collection
+	--dir=str			            directory to store data collection (required)
+    --bpf_prog
```
portante: Seems like there is a tabs vs spaces problem with this file.  Can you be sure to use tabs and not spaces when indenting?

agent/tool-scripts/bpftrace
```diff
@@ -0,0 +1,123 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: t; sh-basic-offset: 8; sh-indentation: 8; sh-indent-for-case-alt: + -*-
+# The following script runs a bpftrace program with the program supplied as argument in pbench-register-tool
+# pbench-register-tool --name=bpftrace -- --bpf_prog=sample_prog.bt
```
sourabhtk37: @atheurer 1. Here's how I registered the tool.

agent/tool-scripts/bpftrace
```diff
@@ -0,0 +1,123 @@
...
+	fi
+	;;
+	start)
+	mkdir -p $tool_output_dir
+	echo "$script_name: running $tool_cmd"
+	tool_stdout_file=$tool_output_dir/$tool-stdout.txt
+	tool_stderr_file=$tool_output_dir/$tool-stderr.txt
+	tool_cmd_file="$tool_output_dir/$tool.cmd"
+    echo "Running"
+	$tool_cmd  > "$tool_stdout_file" 2> "$tool_stderr_file" &
```
sourabhtk37: @atheurer  1. Started the tool by appending forming the command in line 97 and running it here. The tool is stopped by simply sending interrupt signal to the process in line 119.

sourabhtk37:
@atheurer 2. The syscount.bt doesn't need compiling into the kernel, bpftrace using libbpf dynamically loads it into the kernel. 
3. Yes, I'm working on that. Need to make sure kernel configs required to run is enabled and other packages required for running are also installed. Thanks.

agent/tool-scripts/bpftrace
sourabhtk37: Yeah, need to save the command into the file.

agent/tool-scripts/bpftrace
sourabhtk37: What's purpose would that serve?

agent/tool-scripts/bpftrace
```diff
@@ -20,104 +23,105 @@ pbench_bin="`cd ${script_path}/..; /bin/pwd`"
...
 	if [ ! -e $tool_bin ]; then
 		error_log "$tool_bin not found, exiting"
 		exit 1
 	fi
 	;;
-	start)
+start)
 	mkdir -p $tool_output_dir
+	echo "$tool_cmd" >$tool_cmd_file
 	echo "$script_name: running $tool_cmd"
```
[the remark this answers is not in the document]
sourabhtk37: Do you mean line 115? @ndokos

agent/tool-scripts/bpftrace
```diff
@@ -20,104 +23,105 @@ pbench_bin="`cd ${script_path}/..; /bin/pwd`"
...
-	tool_stdout_file=$tool_output_dir/$tool-stdout.txt
-	tool_stderr_file=$tool_output_dir/$tool-stderr.txt
-	tool_cmd_file="$tool_output_dir/$tool.cmd"
-    echo "Running"
-	$tool_cmd  > "$tool_stdout_file" 2> "$tool_stderr_file" &
-    wait
+	echo "Running"
```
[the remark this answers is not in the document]
sourabhtk37: Yeah, thanks.

atheurer:
> @atheurer 2. The syscount.bt doesn't need compiling into the kernel, bpftrace using libbpf dynamically loads it into the kernel.

Yes, I understand it can do this, but my concern is that libbpf doing this at tool-start time might require a lot of [cpu] resources, or might require a little bit of time to do this.  As a comparison, for systemtap, this is why we "pre" compile the script when we register the tool, instead of just doing it automatically at start-tools.  Maybe this is not possible with BPF, but I guess we can see how it behaves with the way you have it now, and then see later if it becomes an issue.

agent/tool-scripts/bpftrace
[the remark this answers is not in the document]
ndokos: Yes.

sourabhtk37:
- [ ] Move bpftrace script to remote system

agent/tool-scripts/bpftrace
```diff
@@ -0,0 +1,140 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: t; sh-basic-offset: 8; sh-indentation: 8; sh-indent-for-case-alt: + -*-
+# The following script runs a bpftrace program with the program supplied as argument in pbench-register-tool
```
portante: s/as argument/as an argument/

agent/tool-scripts/bpftrace
```diff
@@ -0,0 +1,140 @@
...
+# pbench-register-tool --name=bpftrace -- --script=sample_prog.bt
```
portante: Can you indent this 2 or 4 spaces to denote this as an example command?

And do I understand this correctly that the original bpftrace command would look like: `bpftrace --script=sample_prog.bt`?

If that is the case, can you add that to the comment?

agent/tool-scripts/bpftrace
```diff
@@ -0,0 +1,140 @@
...
+# 3) Stop data collection
+# 4) post-process the data
+
+# Defaults
+tool=bpftrace
+tool_bin=/usr/bin/$tool
+group=default
+dir="/tmp"
+
+opts=$(getopt -q -o idp --longoptions "dir:,script:,group:,iteration:,start,stop,install,postprocess" -n "getopt.sh" -- "$@")
```
portante: s/dgi/idp/

The single character options below seem only to be `-i`, `-g`, and `-d`.

agent/tool-scripts/bpftrace
```diff
@@ -0,0 +1,140 @@
...
+	--)
+		shift
+		break
+		;;
+	esac
+done
+
+# Check for location of bpftrace binary
+if [[ ! -e "/usr/bin/$tool" ]]; then
+	if [[ ! -e "/usr/bin/$tool" ]]; then
```
portante: Why do we need to check for the same tool path twice?

agent/tool-scripts/bpftrace
```diff
@@ -0,0 +1,140 @@
...
+		error_log "$tool_bin not found, exiting"
+		exit 1
+	else
+		tool_bin=/usr/local/bin/$tool
```
portante: Why are we looking for the tool file in `/usr/local/bin` as well?  We should probably only look for it in the location the bpftrace RPM installs it.

agent/tool-scripts/bpftrace
```diff
@@ -0,0 +1,140 @@
...
+tool_dir="$dir/tools-$group"
+tool_output_dir=$tool_dir/$tool # all tools keep data in their tool specific dir
+tool_cmd_file="$tool_output_dir/$tool.cmd"
+tool_pid_file=$pbench_tmp/$group.$iteration.$tool.pid
+tool_stdout_file=$tool_output_dir/$tool-stdout.txt
+tool_stderr_file=$tool_output_dir/$tool-stderr.txt
+tool_cmd="$tool_bin $script"
+case "$mode" in
+install)
+	if [[ ! -e $tool_bin ]]; then
```
portante: Can we use `check_install_rpm bpftrace` instead?  See other tool files for an example.

agent/tool-scripts/bpftrace
```diff
@@ -0,0 +1,140 @@
...
+			shift
+		fi
+		;;
+	--script)
+		shift
+		if [ -n "$1" ]; then
+			script=$(echo $1 | sed 's/^"\(.*\)"$/\1/')
+			shift
+		else
+			echo "you did not provide a script"
```
portante: Will this logic hit this `else` statement if `--script` is not given?  Seems like we need to check for all the required parameters after this while loop that processes all the options.

agent/tool-scripts/bpftrace
sourabhtk37: The original command looks like `bpftrace sample_prog.bt`.

portante: Okay, so won't `pbench-register-tool --name=bpftrace -- --script=sample_prog.bt` record the tool to run as, `bpftrace --script=sample_prog.bt`?

sourabhtk37: Nope. `tool_cmd="$tool_bin $script"` Will create the command after parsing, resulting in `bpftrace sample_prog.bt`

portante: Ah, right, the tool we are registering is the `pbench/tool-scripts/bpftrace`, which takes the `--script` argument.  That then invokes `/usr/bin/bpftrace sample_prog.bt`.  So let's try to capture that in the comments to help make it clear.

sourabhtk37: Added comments for clarity.

agent/tool-scripts/bpftrace
sourabhtk37: I have added check for `--script`

agent/tool-scripts/bpftrace
```diff
@@ -0,0 +1,137 @@
...
+script_present=0
+
+opts=$(getopt -q -o dgi --longoptions "dir:,script:,group:,iteration:,start,stop,install,postprocess" -n "getopt.sh" -- "$@")
+if [ $? -ne 0 ]; then
+	printf "\n"
+	printf "$script_name: you specified an invalid option\n\n"
+	printf "The following options are available: \n\n"
+	printf -- "\t--install,			install this perf tool\n"
+	printf "\n"
+	printf -- "\t--start|stop|post-process	start/stop/post-process the data collection\n"
```
portante: s/post-process/postprocess/g

Note that `--post-process` won't be matched at line 59, only `--postprocess`.

agent/tool-scripts/bpftrace
```diff
@@ -0,0 +1,137 @@
...
+		fi
+		;;
+	--script)
+		shift
+		script_present=1
+		if [ -n "$1" ]; then
+			script=$(echo $1 | sed 's/^"\(.*\)"$/\1/')
+			shift
+		else
+			echo "you did not provide a script"
```
portante: Seems like we don't need this `else` branch now that you added the check at line 102.

agent/tool-scripts/bpftrace
```diff
@@ -0,0 +1,137 @@
...
+	-i | --iteration)
+		shift
+		if [ -n "$1" ]; then
+			iteration="$1"
+			shift
+		fi
+		;;
+	--script)
+		shift
+		script_present=1
```
portante: Seems like we don't need this variable.  Can't we just check for a non-empty `$script` instead?

agent/tool-scripts/bpftrace
```diff
@@ -0,0 +1,137 @@
...
+# 2) Start data collection
+# 3) Stop data collection
+# 4) post-process the data
+
+# Defaults
+tool=bpftrace
+tool_bin=/usr/bin/$tool
+group=default
+dir="/tmp"
+script_present=0
```
portante: Can we just initialize the `script` variable to an empty string instead of using `script_present`?

agent/tool-scripts/bpftrace
```diff
@@ -0,0 +1,137 @@
...
+		fi
+		;;
+	--)
+		shift
+		break
+		;;
+	esac
+done
+
+if [ $script_present -eq 0 ]; then
```
portante: Can we check for `$script` being an empty string instead of using a second variable?

agent/tool-scripts/bpftrace
```diff
@@ -0,0 +1,137 @@
...
+tool_pid_file=$pbench_tmp/$group.$iteration.$tool.pid
+tool_stdout_file=$tool_output_dir/$tool-stdout.txt
+tool_stderr_file=$tool_output_dir/$tool-stderr.txt
+tool_cmd="$tool_bin $script"
+case "$mode" in
+install)
+	check_install_rpm bpftrace || exit 1
+	;;
+start)
+	mkdir -p $tool_output_dir
```
portante: Can you check that the `$tool_output_dir` was created successfully?

Perhaps add:
```
if [[ $? -ne 0 ]]; then
    error_log "$script_name: failed to create tool output directory, $tool_output_dir"
    exit 1
fi
```

agent/tool-scripts/bpftrace
sourabhtk37: Good catch.

agent/tool-scripts/bpftrace
sourabhtk37: I think we need this, There are two cases: 1) where `--script=` is supplied with no value 2) and `--script` option is not provided at all. 
So:
1) `pbench-register-tool --name=bpftrace -- --script=`
2) `pbench-register-tool --name=bpftrace --`

agent/tool-scripts/bpftrace
sourabhtk37: checking for non-empty `$script`

agent/tool-scripts/bpftrace
sourabhtk37: Done.

agent/tool-scripts/bpftrace
portante: You are correct, @sourabhtk37, but other arguments need the same checking.  So rather than just checking for `--script` I have posted a PR against your branch in your git repo that suggests a number of changes to enhance the command line processing, error handling, and logic of what is required when and what is optional.

portante:
Can you also rebase and squash this commits?

PR #1111, opened 2019-03-02
Ansible playbook for server post-installation

ndokos:
The  firt commit consists of the playbook and roles, as well as additions and modifications to the pbench-server-activate-* scripts.

The second consists of a change to the unittests script: it now installs the web-server bits, mimicking the installation of same by the server package itself. This changes the output of *all* the unit tests.

### DIFF
--- a/server/ansible/server-config/roles/pbench-server-activate/tasks/main.yml
+++ b/server/ansible/server-config/roles/pbench-server-activate/tasks/main.yml
@@ -1,0 +1,3 @@
+---
+- name: activate the server configuration
+  shell: PATH=/opt/pbench-server/bin:$PATH pbench-server-activate {{ configdest }}/pbench-server.cfg > /tmp/foo
--- a/server/ansible/server-config/roles/pbench-server-activate/tasks/main.yml
+++ b/server/ansible/server-config/roles/pbench-server-activate/tasks/main.yml
@@ -1,3 +1,8 @@
 ---
+- name: ensure httpd is installed
+  package:
+      name: httpd
+      state: present
+
 - name: activate the server configuration
-  shell: PATH=/opt/pbench-server/bin:$PATH pbench-server-activate {{ configdest }}/pbench-server.cfg > /tmp/foo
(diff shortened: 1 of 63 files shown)


server/bin/pbench-server-activate-start-httpd
```diff
@@ -45,4 +45,4 @@ ln -s \
         ${pbench_users_dir} \
         ${pbench_static_dir} \
         ${docroot} 2>/dev/null
-exit $?
+exit 0
```
portante: Why do we need to do this?

portante: Seems like we should be capturing the status of the systemctl commands and firewall-cmds as well, failing when any of them don't work, including the creation of the links.

server/ansible/server-config/roles/pbench-server-activate/tasks/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+- name: activate the server configuration
+  shell: PATH=/opt/pbench-server/bin:$PATH pbench-server-activate {{ configdest }}/pbench-server.cfg > /tmp/foo
```
portante: If we really don't need the output from `pbench-server-activate` then prehaps `/dev/null` would be better?

Perhaps instead we should consider having `pbench-server-activate` save all this output to `/var/log/pbench-server-activate.log`?

server/bin/pbench-server-activate-start-httpd
ndokos: The `ln -s` fails if the links exist already, so this would fail every time after the first time.

I'll move the firewall stuff into the playbook, but the only httpd module I found was for enabling/disabling apache modules. Do you know of anything else?

I'll rework this.

server/ansible/server-config/roles/pbench-server-activate/tasks/main.yml
ndokos: Oh - right. I forgot about that: that was debugging. I'll fix.

PR #1112, opened 2019-03-02
Remove use of deprecated ant design's onRowClick

portante:

PR #1113, opened 2019-03-04
Use only query iterations

portante:
Use one common code for fetching and parsing iteration data (`queryIterations` and `parseIterationData`).

We also fix the sorting to behave correctly on all the pages by sharing the same sort method.

We reduce the number of calls to `setState()` to avoid repeated renderings of the page unnecessarily.

Depends on PR #1112.

### DIFF
--- a/web-server/v0.4/src/pages/Dashboard/Summary.js
+++ b/web-server/v0.4/src/pages/Dashboard/Summary.js
@@ -1,9 +1,10 @@
 import ReactJS from 'react';
 import { connect } from 'dva';
 import { Select, Spin, Tag, Table, Button, Card, notification } from 'antd';
-import axios from 'axios';
 import cloneDeep from 'lodash/cloneDeep';
 import PageHeaderLayout from '../../layouts/PageHeaderLayout';
+import { queryIterations } from '../../services/dashboard';
+import { parseIterationData } from '../../utils/parse';
 
 const tabList = [
   {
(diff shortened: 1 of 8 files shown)


gurbirkalsi:
Hi @portante, I get the following error when clicking into a summary page from the results list, going back to the list of results, and then clicking on another result in the results list:

```
TypeError: Cannot read property 'columns' of undefined
Summary.render
src/pages/Dashboard/Summary.js:331
  328 | }
  329 | 
  330 | var responseDataCopy = {};
> 331 | responseDataCopy['columns'] = cloneDeep(responseData[0].columns);
      | ^  332 | responseDataCopy['iterations'] = cloneDeep(responseData[0].iterations);
  333 | responseDataCopy['resultName'] = cloneDeep(responseData[0].resultName);
```
This is most likely happening as the render function is being executed while the iteration data is being parsed. Can we have a check here in the render function to make sure the `responseData` variable contains data?

web-server/v0.4/src/pages/Dashboard/Summary.js
```diff
@@ -4,6 +4,8 @@ import { Select, Spin, Tag, Table, Button, Card, notification } from 'antd';
 import axios from 'axios';
```
gurbirkalsi: Can we remove this import since axios is not being referenced in this component?

portante: Yes, I'll remove that.

portante:
@gurbirkalsi, thanks, I'll see why the current if doesn't catch that case.

portante:
@gurbirkalsi, fixed the bad columns reference, can you try again?

web-server/v0.4/src/pages/Dashboard/Summary.js
```diff
@@ -713,179 +394,175 @@ class Summary extends ReactJS.Component {
...
+          <Table columns={columns} dataSource={tocTree} defaultExpandAllRows />
+        </Card>
+      ),
+    };
+
+    return (
+      <div style={{ display: 'flex', flexDirection: 'column' }}>
+        <div>
+          <PageHeaderLayout
+            title={selectedResults['run.name']}
```
gurbirkalsi: The name of the run is not being rendered as we are not referring to the first run item in the `selectedResults` array.

portante: @gurbirkalsi, fixed.

Issue #1116, opened 2019-03-05
Remove "each child in a list should have a unique key prop" console error/log message


portante:
```
index.js:1 Warning: Each child in a list should have a unique "key" prop.

Check the render method of `Summary`. See https://fb.me/react-warning-keys for more information.
    in Option (created by Summary)
    in Summary (created by Connect(Summary))
    in Connect(Summary) (created by Connect(Connect(Summary)))
    in Connect(Connect(Summary)) (created by Route)
    in Route (created by RouteWithProps)
    in RouteWithProps (created by Route)
    in Switch (created by Route)
    in div (created by Basic)
    in Basic (created by Context.Consumer)
    in Adapter (created by BasicLayout)
    in div (created by BasicLayout)
    in BasicLayout (created by Context.Consumer)
    in Adapter (created by BasicLayout)
    in div (created by BasicLayout)
    in BasicLayout (created by Context.Consumer)
    in Adapter (created by BasicLayout)
    in div (created by ContainerQuery)
    in ContainerQuery (created by BasicLayout)
    in DocumentTitle (created by SideEffect(DocumentTitle))
    in SideEffect(DocumentTitle) (created by BasicLayout)
    in BasicLayout (created by Connect(BasicLayout))
    in Connect(BasicLayout) (created by Route)
    in Route (created by RouteWithProps)
    in RouteWithProps (created by RouterWrapper)
    in Switch (created by RouterWrapper)
    in Router (created by ConnectedRouter)
    in ConnectedRouter (created by RouterWrapper)
    in RouterWrapper
    in Provider (created by DvaContainer)
    in DvaContainer
```

portante:
Fixed by #1118.

aquibbaig:
This needs to be fixed in the `searchList` page too. If we don't have specific keys what about using [uuidv4](https://www.npmjs.com/package/uuid)

PR #1118, opened 2019-03-05
Resolve "each child in a list should have a unique key prop" console error/log message

gurbirkalsi:
Rendered lists of children elements should contain a unique "key" prop


portante:
Fixes #1116.

PR #1119, opened 2019-03-05
pbench-run-benchmark & co: fix whitespace

atheurer:


agent/bench-scripts/pbench-gen-iterations
```diff
@@ -207,7 +207,6 @@ sub build_iteration_cmds {
             $cmdline .= " --" . $arg . "=" . $simplex_params{$arg};
         }
         return ($cmdline);
```
ndokos: Along with the white space, a closing brace was lost.

portante: Where is the lost closing brace supposed to go?

atheurer: Lost bracket goes in line 208, and it has been restored.

PR #1124, opened 2019-03-08
Handle @metadata.controller_dir properly

portante:
... and the other optional parameters as well.

We also fix the other fetch of run.prefix which could be undefined.


portante:
@ndokos, this fix is needed for working with existing indexed data.  I'd like to get this into 0.57 for completeness, but not necessary.

@gurbirkalsi, please take a look when you have a chance.

PR #1130, opened 2019-03-13
Destructure page components for reducing component rerenders.

gurbirkalsi:
This first pass in an effort to reduce page component rerenders represents one of three steps to enhance component performance:

1. Destructure page components by reorganizing logic into subcomponents (SearchBar, MonthSelect, RowSelection, Table, etc).
2. Group Redux actions and reducers to reorganize information needed for dashboard entry points that require page initializing (Controllers and Search)
3. Compare previous props and state values to prevent page rerenders using shouldComponentUpdate() or converting components to PureComponent

### DIFF
--- a/web-server/v0.4/src/models/dashboard.js
+++ b/web-server/v0.4/src/models/dashboard.js
@@ -64,7 +64,7 @@ export default {
         console.log('Unsuccessful run data query, no hits on matching documents: ' + error);
         hits = [];
       }
-      hits.map(result => {
+      hits.map((result, index) => {
         let name, controller, id, start, end;
         try {
           name = result.fields['run.name'][0];
@@ -86,7 +86,7 @@ export default {
           return;
         }
         let record = {
-          key: name,
+          key: index,
           startUnixTimestamp: Date.parse(start),
           ['run.name']: name,
           ['run.controller']: controller,
--- a/web-server/v0.4/src/pages/Dashboard/Controllers.js
+++ b/web-server/v0.4/src/pages/Dashboard/Controllers.js
@@ -1,9 +1,12 @@
-import { Component } from 'react';
+import React, { Component } from 'react';
 import { connect } from 'dva';
 import { routerRedux } from 'dva/router';
-import { Card, Table, Input, Button, Icon, Form, Select, notification } from 'antd';
-const FormItem = Form.Item;
+import { Card, Form } from 'antd';
+
 import PageHeaderLayout from '../../layouts/PageHeaderLayout';
+import SearchBar from '../../components/SearchBar';
+import MonthSelect from '../../components/MonthSelect';
+import Table from '../../components/Table';
 import { compareByAlph } from '../../utils/utils';
 
 @connect(({ global, dashboard, loading }) => ({
@@ -11,9 +14,10 @@ import { compareByAlph } from '../../utils/utils';
   indices: global.indices,
   selectedIndices: global.selectedIndices,
   datastoreConfig: global.datastoreConfig,
-  loadingControllers: loading.effects['dashboard/fetchControllers'],
-  loadingIndices: loading.effects['global/fetchMonthIndices'],
-  loadingConfig: loading.effects['global/fetchDatastoreConfig'],
+  loadingControllers:
+    loading.effects['dashboard/fetchControllers'] ||
+    loading.effects['global/fetchMonthIndices'] ||
+    loading.effects['global/fetchDatastoreConfig'],
 }))
 export default class Controllers extends Component {
   constructor(props) {
@@ -21,8 +25,6 @@ export default class Controllers extends Component {
 
     this.state = {
       controllers: [],
-      selectedIndicesUpdated: false,
-      searchText: '',
     };
   }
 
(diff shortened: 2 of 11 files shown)


portante:
@gurbirkalsi, looks like this requires a rebase.

web-server/v0.4/src/models/dashboard.js
```diff
@@ -55,60 +55,68 @@ export default {
...
+        let id;
+        let start;
+        let end;
         try {
           name = result.fields['run.name'][0];
           controller = result.fields['run.controller'][0];
           id = result.fields['run.id'][0];
           start =
-            typeof result.fields['run.start'] != 'undefined'
+            typeof result.fields['run.start'] !== 'undefined'
```
portante: Why are we adding the extra equal sign?  Doesn't `typeof` return an actual string?  If so, aren't we already properly comparing two strings?

web-server/v0.4/src/models/dashboard.js
```diff
@@ -167,25 +175,25 @@ export default {
     *updateSelectedController({ payload }, { select, put }) {
       yield put({
         type: 'modifySelectedController',
-        payload: payload,
+        payload,
```
portante: Is this still valid?  Isn't this supposed to be an object of the `payload` field that is `put()`?

web-server/v0.4/src/services/dashboard.js
```diff
@@ -134,8 +134,8 @@ export async function queryIterations(params) {
 
   let iterationRequests = [];
   selectedResults.map(result => {
-    const controller_dir = result['@metadata.controller_dir'];
-    if (controller_dir === undefined) {
+    let controller_dir = result['@metadata.controller_dir'];
```
portante: This was already fixed, so a rebase will get the proper fix already committed.

web-server/v0.4/src/pages/Dashboard/Summary.js
```diff
@@ -574,7 +495,7 @@ class Summary extends ReactJS.Component {
             title={selectedResults[0]['run.name']}
             content={
               <Tag color="blue" key={selectedController}>
-                {'controller: ' + selectedController}
+                {`controller: ${  selectedController}`}
```
portante: Formatting?

web-server/v0.4/src/models/dashboard.js
gurbirkalsi: The extra equal operator (`===`) provides type checking vs the regular counterpart (`==`) that provides type coercion. In general, it is good practice to use type-safe equality operators to avoid edge case scenarios such as these which evaluate to `true`: 
```
[] == ![]
8 == "08"
```

web-server/v0.4/src/models/dashboard.js
gurbirkalsi: This is a feature of ECMAScript 6 that allows you to define object literal methods and properties in a shorthand form for a cleaner syntax. This requires the key name to match the name of the assigned variable.

web-server/v0.4/src/services/dashboard.js
gurbirkalsi: Sounds good! I'll go ahead and rebase in order to resolve this.

web-server/v0.4/src/pages/Dashboard/Summary.js
gurbirkalsi: This is another feature of ES 6 (ECMAScript 6) that allows you to use template string literals instead of string concatenation. Template string literals should only be used where interpolation is happening or new lines are being created. I'll remove the space before `selectedController` so we can achieve a cleaner syntax here.

gurbirkalsi:
Thanks Peter. I certainly agree with your sentiment that the organization for changes can be improved here to show the precise enhancements. I'll be sure to follow a concise and targeted organization going forward so we have a clear understanding of the changes. 

I'll reorganize this PR to reflect only the intended enhancements stated in the description for destructering components and submit a separate PR for ES 6 upgrades. My intention for using ES 6 features is to slowly adopt Airbnb's JavaScript style guide as the new baseline standard for this project as they provide strict rules to improve on the readability and performance of JavaScript.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

gurbirkalsi:
Unfortunately, those warnings still don't disappear after adding the peerDependency definitions in our `package.json` file. After looking at react-fittext's repo, they have not released a version of their library that supports React 16 which is the version of React being used in the dashboard. However, this is a dependency for antd components which is why we receive this warning.

web-server/v0.4/src/pages/Dashboard/Controllers.js
```diff
@@ -44,13 +46,9 @@ export default class Controllers extends Component {
...
-    })
-      .then(() => {
-        this.fetchMonthIndices();
-      })
-      .catch(err => {
-        console.log(err);
-      });
+    }).then(() => {
+      this.fetchMonthIndices();
+    });
```
portante: Why are we dropping the `.catch()` clause which will log the error to the console?  Are we handling this some place else now?

gurbirkalsi: I didn't realize this was dropped and will check this back in. Eventually, we'll want to have a global error reducer that captures networking errors given the status of the request ("request", "success", "error") and logs them with Redux.

Issue #1070, opened 2019-02-06
Update pbench-sysstat, pbench-uperf, pbench-fio to latest upstream.


ndokos:
Current versions used in pbench vs. what's available upstream at the current time:

- [ ] `fio`: 3.12 vs 3.12
- [ ] `uperf`: 1.0.4 vs 1.0.6
- [ ] `sysstat`: 11.2.0 vs 12.0.3

COPR builds of `uperf` and `sysstat` have succeeded on everything but F29 - the F29 failures are under investigation.

portante:
For `sysstat` upgrade from 11.2.0 to 12.0.3, we have the following issues:

 * `pidstat` output now includes two new columns: `%wait` and `iodelay`
 * `iostat` output now includes additional columns and the columns have been rearranged

An investigation of `sar` output is underway.  `mpstat` output has not changed.

portante:
First, the `sysstat` support was updated with PR #1133 in `pbench-agent` `v0.58`.

Since then, we now support minimum versions of `fio` and `uperf`.

However, we do not maintain packaging mechanisms for `fio` and `uperf` in this repository, so we can close this.

PR #1133, opened 2019-03-13
Use sysstat 12.0.3

portante:
Partially fix for #1070.  We provide a series of patches to update the pbench-sysstat package to 12.0.3, changing the `sar` tool to require that version, and update the unit tests for post-processing.

We take care of some house-keeping clean up in the first two commits, and add the subsequent version change work in the following ones.


atheurer:
@portante should I assume these changes are not changing any of the tool's output to JSON?

portante:
Yes, for now I am trying to see how hard it is to make the existing unit tests succeed with out changes.

portante:
@atheurer, it would be good if you could provide early feedback on `pidstat-` and `iostat-postprocess` changes.  I have not started the sar changes yet.

The `iostat` changes are a bit more involved, but still fairly straightforward (units change from 512 byte sectors to kB).

portante:
@atheurer, I added the iostat postprocessing support for CDM to the "iostat support" commit, and then added two follow-on commits for adding unit tests.

We can't make the unit tests work because each time you run the generation of a CDM document the UUID changes for the JSON.  We don't do that when indexing on the pbench server today, we always generate a reproducible UID so that we can validate the document contents match the UID.  If we could do the same here, the second commit proposes storing the JSON documents in one file per device (avoids 46,000+ files in one of the unit tests).

Something to consider.

When you are ready, I'll move these CDM unit tests to another commit, and we can merge this PR.

Thanks!

atheurer:
Also, forgot to mention, I think we should defer the iostat-postprocess-cdm unittest until we come up with a better method.  We can address that when I write the postprocess-cdm for sar, mpstat, and pidstat.

portante:
@atheurer, I have pulled out the unit tests for `iostat-postprocess-cdm`, leaving the actual fixes in place.  Once the TravisCI tests finish running, I'll merge this.

PR #1134, opened 2019-03-14
Move to Ubuntu Xenial in TravisCI w py36

portante:


portante:
@ndokos, this makes the unit tests much more stable, they passed without having to restart any jobs.

Perhaps worth even taking into 0.57.

PR #1135, opened 2019-03-14
Remove 'selectorIndices' logic for simple month selection

gurbirkalsi:
Two arrays were being used to track the value and position
of a selected month. This is now referencing only the month
value and references the month string as a key within the
`<Select.Option>` child component.

### DIFF
--- a/web-server/v0.4/src/pages/Search/SearchList.js
+++ b/web-server/v0.4/src/pages/Search/SearchList.js
@@ -12,7 +12,6 @@ const Option = Select.Option;
   fields: search.fields,
   selectedFields: search.selectedFields,
   selectedIndices: global.selectedIndices,
-  selectorIndices: global.selectorIndices,
   indices: global.indices,
   datastoreConfig: global.datastoreConfig,
   loadingMapping: loading.effects['search/fetchIndexMapping'],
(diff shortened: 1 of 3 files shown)


web-server/v0.4/src/pages/Search/SearchList.js
```diff
@@ -210,12 +200,12 @@ export default class SearchList extends Component {
...
                     style={{ width: '100%' }}
                     placeholder="Select index"
-                    value={selectorIndices}
+                    value={selectedIndices}
                     onChange={this.updateSelectedIndices}
                     tokenSeparators={[',']}
                   >
-                    {indices.map((index, i) => {
-                      return <Select.Option key={i}>{index}</Select.Option>;
+                    {indices.map((month) => {
```
portante: This looks kinda similar to the code in the Controllers page, any possibility of refactoring this out into a component?

gurbirkalsi: @portante Yes we should reuse the `MonthSelect` component I've defined in PR #1130 to replace this Select reference. I'll update that PR with those changes to SearchList.js.

gurbirkalsi: We may want to merge this commit first so we can rebase PR #1130 on top of the modified Redux logic before I make that change.

PR #1136, opened 2019-03-14
Reference selectedResults as an array from SearchList

gurbirkalsi:
Converts the selected result to an array for parsing
within the Summary page component. Previously this was
passed as an object which caused an error.

This follows the same workflow presented from the row
onClick method in the Results component.

### DIFF
--- a/web-server/v0.4/src/pages/Search/SearchList.js
+++ b/web-server/v0.4/src/pages/Search/SearchList.js
@@ -135,16 +135,16 @@ export default class SearchList extends Component {
     });
   };
 
-  retrieveResults = params => {
+  retrieveResults = selectedResults => {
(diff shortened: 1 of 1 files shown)


web-server/v0.4/src/pages/Search/SearchList.js
```diff
@@ -140,7 +140,7 @@ export default class SearchList extends Component {
 
     dispatch({
       type: 'dashboard/updateSelectedController',
-      payload: params['run.controller'],
+      payload: params[0]['run.controller'],
```
portante: Can we also rename `params` to `selectedResults`?

gurbirkalsi: I've updated the reference here.

PR #1139, opened 2019-03-15
Use tail for all unit tests

portante:


portante:
@ndokos, pulled out commit into its own PR from #1036.

portante:
@ndokos, so this is a tail of one file, right?  Why would you see that here?

ndokos:
The inotify stuff? No idea. Somebody must be using inotify watchers as if there is no tomorrow, on my laptop. That was an FYI: I suspect it's not going to show up anywhere else.

portante:
@ndokos, thanks, not sure what to do about inotify issues.  That seems like it is just with your environment.

We have been using back-grounded jobs since before this commit.  Are you requesting that we fix that problem with this commit?

I am just wondering what is the specific change you are requesting (as your review is marked "requested changes".

ndokos:
The inotify stuff was FYI only: I'm not suggesting that anything should be done about it.
The interrupted run might be worth doing something about: that's why I marked it "Requested changes". Howeveer, since this is mostly Travis-CI where the VM will die, it might not be worth the bother. I leave it to your discretion.

portante:
@ndokos, since this condition of not handling ctrl-c within `run-unittests` existed prior to this PR, I have filed https://github.com/distributed-system-analysis/pbench/issues/1141 to track separately.  I'll merge this in the mean time.

PR #1142, opened 2019-03-18
Integrate redux persist for persisting and rehydrating redux store

gurbirkalsi:
Redux Persist takes the state object and saves it to the persisted
storage (LocalStorage) for reference throughout multiple sessions.

This adds a configuration for setting up a base PersistGate in order
to capture state changes to the Redux store.

### DIFF
--- a/web-server/v0.4/src/pages/Dashboard/Controllers.js
+++ b/web-server/v0.4/src/pages/Dashboard/Controllers.js
@@ -24,20 +24,21 @@ export default class Controllers extends Component {
     super(props);
 
     this.state = {
-      controllers: [],
+      controllers: this.props.controllers
     };
   }
 
   componentDidMount() {
-    this.queryDatastoreConfig();
+    const { controllers, selectedIndices, indices } = this.props;
+
+    if (controllers.length === 0 || selectedIndices.length === 0 || indices.length === 0) {
+      this.queryDatastoreConfig();
+    }
   }
 
-  componentDidUpdate(prevProps) {
-    const { controllers } = this.props;
-    const prevControllers = prevProps.controllers;
-
-    if (controllers !== prevControllers) {
-      this.setState({ controllers });
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.controllers !== this.props.controllers) {
+      this.setState({ controllers: nextProps.controllers });
     }
   }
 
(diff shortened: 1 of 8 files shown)


web-server/v0.4/src/pages/Dashboard/Controllers.js
```diff
@@ -27,7 +27,13 @@ export default class Controllers extends Component {
   }
 
   componentDidMount() {
-    this.queryDatastoreConfig();
+    const { controllers, selectedIndices, indices } = this.props;
+
+    if (controllers.length === 0 || selectedIndices.length === 0 || indices.length === 0) {
+      this.queryDatastoreConfig();
+    } else {
+      this.setState({ controllers });
```
portante: Why do we set the state to the same state it was here?  If we need this, can you add a comment explaining why it is necessary?

gurbirkalsi: We set controllers to the state here in the case where the user has spun up the dashboard session again because the component needs to remount. Since the data has persisted from the last session, we don't need to make the necessary queries to initialize the view; however, our Table component in this page component depends on accessing controllers from the state rather than the props for search manipulation.

portante: Okay, that seems suspect.  Do we know the user has spun up the dashboard again?  Or are we just doing it because it "could" be the case they have spun it up again?

If we know for sure, that is one thing, but if this is just to cover the "could" case, can't we use additional data from the environment to tell for sure?

gurbirkalsi: We know the user has spun up the dashboard again or for the first time because this logic is executed in the `componentDidMount()` method which is invoked after a component is mounted to the DOM tree. Since the `Controllers` component is an entry point and defined as the base url `/dashboard`, we can check the following:

1. the `Controllers` component has remounted (logic within `componentDidMount()` lifecycle method is referenced)
2. the necessary component props data has persisted when restarting the session

gurbirkalsi: I'm including a cleaner solution for handling the handoff from the controllers prop to the controllers state data. Essentially, we'll initialize the state data with the controllers prop in the constructor of the component and use `componentWillReceiveProps()` to make necessary updates to the state data when the controllers prop changes. This way we avoid update the state in `componentDidMount()` and remove an "extra check" we have to perform during the persisted lifecycle.

PR #1148, opened 2019-03-22
Add env var for server unit test parallelism

portante:
We also ensure that for now TravisCI runs server tests serially to avoid
not emitting output for more than ten minutes triggering the TravisCI auto
kill feature.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,28 @@
 dist: xenial
 language: perl
 perl:
- - "5.28"
- - "5.26"
- - "5.24"
- - "5.22"
- - "5.20"
- - "5.18"
- - "5.16"
- - "5.10"
+  - "5.28"
+  - "5.26"
+  - "5.24"
+  - "5.22"
+  - "5.20"
+  - "5.18"
+  - "5.16"
+  - "5.10"
 before_install:
- - sudo apt-get update
+  - sudo apt-get update
 install:
- - sudo apt-get install python python-pip bc libjson-perl libswitch-perl realpath
- - sudo pip install configtools elasticsearch
- - sudo apt-get install python-software-properties
- - sudo add-apt-repository ppa:fkrull/deadsnakes -y
- - sudo apt-get update
- - sudo apt-get install python3.6 --force-yes -y
- - sudo wget -O /tmp/get-pip.py https://bootstrap.pypa.io/get-pip.py
- - sudo python3.6 /tmp/get-pip.py
- - sudo pip install 'configtools<0.4.0' elasticsearch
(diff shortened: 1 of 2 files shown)


portante:
Merged without review for now.  We can always back it out if necessary.

.travis.yml
```diff
@@ -1,26 +1,28 @@
...
+  - sudo apt-get install python-software-properties
+  - sudo add-apt-repository ppa:fkrull/deadsnakes -y
+  - sudo apt-get update
+  - sudo apt-get install python3.6 --force-yes -y
+  - sudo wget -O /tmp/get-pip.py https://bootstrap.pypa.io/get-pip.py
+  - sudo python3.6 /tmp/get-pip.py
+  - sudo pip install 'configtools<0.4.0' elasticsearch
+  - sudo ln -sf python3.6 /usr/bin/python3
+env:
+  - PBENCH_UNITTEST_SERVER_MODE=serial
```
ndokos: The default mode is "serial", so this is unnecessary - right? Is this belt-and-suspenders insurance?

portante:
This is to ensure the default mode *for unit tests* is "`serial`".

So first, `server/bin/unittests` defaults to "`serial`" today.  Using this environment variable ensures that if the default for `server/bin/unittests` changes from "`serial`" to "`parallel`", the unit tests will still run with mode "`serial`".  We need mode "`serial`" to prevent too many parallel jobs running preventing output from being generated beyond the 10 minute default timeout.

Somebody can still modify the top-level `run-unittests` script to add `--parallel` if they want, but such a change would then kill the Travis CI jobs again, since the command line option overrides the built-in default and the environment variable.

The purpose of this is so that a developer can set *their* working server unit tests mode to "`parallel`" without affecting the Travis CI behavior with a change to the top-level `run-unittests`.

.travis.yml
portante: This is to ensure the default mode *for unit tests* is "`serial`".

So first, `server/bin/unittests` defaults to "`serial`" today.  Using this environment variable ensures that if the default for `server/bin/unittests` changes from "`serial`" to "`parallel`", the unit tests will still run with mode "`serial`".  We need mode "`serial`" to prevent too many parallel jobs running preventing output from being generated beyond the 10 minute default timeout.

Somebody can still modify the top-level `run-unittests` script to add `--parallel` if they want, but such a change would then kill the Travis CI jobs again, since the command line option overrides the built-in default and the environment variable.

The purpose of this is so that a developer can set *their* working server unit tests mode to "`parallel`" without affecting the Travis CI behavior with a change to the top-level `run-unittests`.

PR #1150, opened 2019-03-28
Cephfs fio

bengland2:
This PR should make pbench-fio work with shared filesystems (where multiple hosts/pods access the same files).  In this situation, fio needs to generate filenames unique to the host and job number.

There is a new templates/fio-shared-fs.job file.  I hesitate for now to change filename=$target to directory=$target in the templates/fio.job file because this could break some other use case.  Specifically filename=$target is probably more appropriate for block devices/partitions being fed to pbench-fio (where fio can't make up the pathname as it sees fit).   So I created a separate templates/fio-shared-fs.job that just changes "filename=$target" to "directory=$target" , and you can specify this alternate job file template using the --job-file= option.

See the documentation change and commit comments for more explanation.

This PR has had limited testing.   I would like to see it tested a bit more before merging, but wanted people to have a look at it and see what they thought.

### DIFF
--- a/agent/bench-scripts/pbench-fio
+++ b/agent/bench-scripts/pbench-fio
@@ -567,22 +567,34 @@ function fio_run_job() {
 	pbench-start-tools --group=$tool_group --iteration=$iteration --dir=$benchmark_results_dir
 	local client_opts=""
 
-	if [ ! -z "$client_file" ] ;then
-		typeset -i nclients=$(wc -l $client_file | cut -d ' ' -f 1)
-		client_opts="--client=$client_file --max-jobs=$nclients"
-	elif [ ! -z "$clients" ]; then
+	# if a client-file parameter was passed, then clients is already 
+	# initialized to its contents
+
+	if [ ! -z "$clients" ]; then
 		local max_jobs=0
+		local nclients=0
+
+		if [ ! -z $numjobs ]; then
+			# so that a separate filename will be generated
+			# for each (host, job-number) pair
+			let max_jobs=$numjobs
+		else
+		       for t in `echo $targets | sed -e s/,/" "/g` ; do
+			       let max_jobs=$max_jobs+1
+		       done
+		fi
+
+		# make the fio command shorter by always using a client file
+		# regardless of how clients were specified
+		# we always save the client list in the benchmark results directory
+
+		client_file=$benchmark_results_dir/pbench-clients.list
 		for client in `echo $clients | sed -e s/,/" "/g`; do
-			client_opts="$client_opts --client=$client"
-			if [ ! -z $numjobs ]; then
-				let max_jobs=$numjobs
-			else
-			       for t in `echo $targets | sed -e s/,/" "/g` ; do
-				       let max_jobs=$max_jobs+1
-			       done
-			fi
-		done
-		client_opts="$client_opts --max-jobs=$max_jobs"
+			echo $client
+		done > $client_file
+		let nclients=`wc -l < $client_file`
+		let max_jobs=$max_jobs*$nclients
+		client_opts="$client_opts --client=$client_file --max-jobs=$max_jobs"
 	fi
 
 	# create a command file and keep it with the results for debugging later, or user can run outside of pbench
@@ -593,7 +605,7 @@ function fio_run_job() {
 	$benchmark_results_dir/fio.cmd >$benchmark_results_dir/fio-result.txt
 	popd >/dev/null
 	pbench-stop-tools --group=$tool_group --iteration=$iteration --dir=$benchmark_results_dir
-
+ 
 	histogram_cmd="fio-histo-log-pctiles.py --time-quantum $histogram_interval_sec --percentiles 0 50 90 95 99 100 -- "
 	if [ ! -z "$clients" ]; then
 		for client in `echo $clients | sed -e s/,/" "/g`; do
--- a/agent/bench-scripts/templates/fio-shared-fs.job
+++ b/agent/bench-scripts/templates/fio-shared-fs.job
@@ -1,0 +1,33 @@
+# Syntax:
+#   $target : expands to the fio target file directory
+#   $@ : expands to the value of the option as passed to
+#        pbench-fio
+#
+# For python2 without the latest configparser installed, lines
+# cannot be indented.
+
+[global]
+bs = $@
+ioengine = libaio
+iodepth = 32
+direct = 1
+sync = 0
+time_based = 1
+clocksource = gettimeofday
+ramp_time = 5
+write_bw_log = fio
+write_iops_log = fio
+write_lat_log = fio
+log_avg_msec = 1000
+write_hist_log = fio
+log_hist_msec = 10000
+# log_hist_coarseness = 4 # 76 bins
+
+[TheJob]
+runtime = $runtime
+directory = $target
+rw = $@
+size = $size
+# if you want to specify more than 1 process per fio client host/pod,
+# uncomment this next line and add --numjobs=K to your pbench-fio command
+#numjobs = $numjobs
--- a/agent/bench-scripts/pbench-fio.md
+++ b/agent/bench-scripts/pbench-fio.md
@@ -89,6 +89,22 @@ and then randomly read/write only an insignificant fraction of the entire set of
 yet still get representative throughput and latency numbers.  
 The same is true for using fio for random I/O on block devices.  
 
+## shared filesystems
+
+A slightly different set of parameters is needed for pbench-fio to work with shared filesystems (where same set of files
+is shared by multiple clients/mountpoints).  For this to work, you must do 2 things:
+
+* specify --target=my-directory 
+* specify --jobfile=/opt/pbench-agent/bench-scripts/templates/fio-shared-fs.job  
+
+**my-directory** is the shared filesystem mountpoint or some subdirectory of it
+
+The job file specification tells fio to target a directory instead of a file.  fio --client
+will then generate unique filenames for every pair of (host, job-number) combinations
+so that no 2 fio jobs will access the same pathname.  See [fio
+documentation](https://fio.readthedocs.io/en/latest/fio_doc.html#client-server) for details.
+(at present the combination of numjobs > 1 with fio-shared-fs.job has not been tested)
+
 # data persistence
 
 Performance of writes is usually only interesting if we know that the writes are persistent (will be visible after a client or server abrupt reset (i.e. crash or power-cycle).  Otherwise you may be just writing to RAM.  For sequential writes, this can be accomplished with the fio jobfile parameter **fsync_on_close=1** . For random writes, this can be accomplished with the parameter **--sync=1** .  Technically O_DIRECT open flag does not guarantee persistence of writes, and only specifies that writes bypass buffer cache.  For example, it is possible to do an O_DIRECT write and have it not be persistent if it reaches block device hardware but was never actually written to persistent storage there, and yes this can happen.  O_SYNC means that the driver both issues the write and blocks until the write is guaranteed to persist.  
@@ -154,9 +170,13 @@ A successful test should show both high throughput and acceptable latency.  For 
 
 # measuring latency percentiles
 
-Usually when a latency requirement is defined in a SLA (service-level agreement) for a cluster, the implication is that *at any point in time* during the operation of the cluster, the Nth latency percentile will not exceed X seconds.  However, fio does not output that information directly in its normal logs - instead it outputs latency percentiles measured over the duration of the entire test run.  For short test this may be sufficient, but for longer tests or tests designed to measure *changes* in latency, you need to capture the variation in latency percentiles over time.   The new fio histogram logging feature helps you do that.  It uses the histogram data structures internal to every fio process, emitting this histogram to a log file at a user-specified interval.  A post-processing tool, [fiologparser-hist.py](https://github.com/axboe/fio/blob/master/tools/hist/fiologparser_hist.py), then reads in these histogram logs and outputs latency percentiles over time.   pbench's postprocessing tool runs it and then graphs the results.  The job file parameters **write_hist_log=h** and **hist_log_msec** control the pathnames and interval for histogram logging.
+Usually when a latency requirement is defined in a SLA (service-level agreement) for a cluster, the implication is that *at any point in time* during the operation of the cluster, the Nth latency percentile will not exceed X seconds.  However, fio does not output that information directly in its normal logs - instead it outputs latency percentiles measured over the duration of the entire test run.  For short test this may be sufficient, but for longer tests or tests designed to measure *changes* in latency, you need to capture the variation in latency percentiles over time.   The new fio histogram logging feature helps you do that.  It uses the histogram data structures internal to every fio process, emitting this histogram to a log file at a user-specified interval.  A post-processing tool, [fio-histo-log-pctiles.py](https://github.com/axboe/fio/blob/master/tools/hist/fio-histo-log-pctiles.py), then reads in these histogram logs and outputs latency percentiles over time.   pbench's postprocessing tool runs it and then graphs the results.  Documentation is [here](https://github.com/axboe/fio/blob/master/doc/fio-histo-log-pctiles.pdf).  The job file parameters **write_hist_log=h** and **hist_log_msec** control the pathnames and interval for histogram logging.
 
-# example of OpenStack Cinder volumes
+# examples
+
+Here are some common use cases for distributed fio.
+
+## example of OpenStack Cinder volumes
 
 Advice: do everything on smallest possible scale to start with, until you get your input files to work, then scale it out.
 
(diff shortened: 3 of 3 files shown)


agent/bench-scripts/templates/fio-shared-fs.job
```diff
@@ -0,0 +1,35 @@
...
+# log_hist_coarseness = 4 # 76 bins
+
+[TheJob]
+directory = $target
+rw = $@
+size = 4096M
+numjobs = 1
+
+#rate_iops=10
+#directory=/tmp/fio
```
portante: Can we remove the last two commented-out lines?

agent/bench-scripts/templates/fio-shared-fs.job
```diff
@@ -0,0 +1,35 @@
+# Syntax:
+#   $target : expands to the fio target file
```
portante: Should this read, `expands to the fio target directory`?

agent/bench-scripts/pbench-fio
```diff
@@ -575,7 +575,9 @@ function fio_run_job() {
 		for client in `echo $clients | sed -e s/,/" "/g`; do
 			client_opts="$client_opts --client=$client"
 			if [ ! -z $numjobs ]; then
-				let max_jobs=$numjobs
+                # so that a separate filename will be generated
+                # for each (host, job-number) pair
+				let max_jobs=$max_jobs+$numjobs
```
portante: Isn't `numjobs` a command line parameter?  If it is, won't it default to `1` but potentially some other value that is not one?  If so, how will it work in that case?

agent/bench-scripts/pbench-fio
```diff
@@ -575,7 +575,9 @@ function fio_run_job() {
 		for client in `echo $clients | sed -e s/,/" "/g`; do
 			client_opts="$client_opts --client=$client"
 			if [ ! -z $numjobs ]; then
-				let max_jobs=$numjobs
+                # so that a separate filename will be generated
```
portante: Can you say more as to why this works?

So we'll have multiple jobs run per client?

agent/bench-scripts/templates/fio.job
```diff
@@ -24,7 +24,7 @@ write_hist_log = fio
 log_hist_msec = 10000
 # log_hist_coarseness = 4 # 76 bins
 
-[job-$target]
+[TheJob]
```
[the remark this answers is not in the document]
k-rister: This change concerns me since I run multi-job against a single host and am therefor given multiple jobs with unique names in the current scheme.  For example:

[global]
bs=4k
runtime=120
ioengine=libaio
iodepth=12
direct=1
sync=0
time_based=1
clocksource=gettimeofday
ramp_time=5
write_bw_log=fio
write_iops_log=fio
write_lat_log=fio
log_avg_msec=1000
write_hist_log=fio
log_hist_msec=10000
norandommap=1

[job-/dev/disk/by-id/nvme-INTEL_SSDPED1K375GA_PHKS8020006Z375AGN]
filename=/dev/disk/by-id/nvme-INTEL_SSDPED1K375GA_PHKS8020006Z375AGN
rw=read
size=10g
numjobs=1

[job-/dev/disk/by-id/nvme-INTEL_SSDPED1K375GA_PHKS802000W8375AGN]
filename=/dev/disk/by-id/nvme-INTEL_SSDPED1K375GA_PHKS802000W8375AGN
rw=read
size=10g
numjobs=1

This yields uniquely identifying information in the JSON results that would be missing without unique job names.  Why does that need to be removed?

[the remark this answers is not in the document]
bengland2: Karl, so it sounds like we could take out the change to fio.job.  This is exactly why I created a separate fio-shared-fs.job, because I didn't want to disrupt existing usage of fio.job, only I didn't realized significance of "[fio-$target]" to you.  It does have some weird behavior associated with it, but as long as you don't put slashes in the target I guess it would be ok in fio.job.  But for fio-shared-fs.job, the target isn't a particular block device or set of block devices, it is a directory, and there can be only 1, so we really don't want to have "[fio-$target]" there.

agent/bench-scripts/templates/fio-shared-fs.job
bengland2: yes

agent/bench-scripts/pbench-fio
bengland2: it defaults to 1 unless you override it.   It's the number of jobs per host, effectively.  See [fio documentation](https://fio.readthedocs.io/en/latest/fio_doc.html) and search for definition there.

agent/bench-scripts/pbench-fio
bengland2: max_jobs is set to 0 at the top of the loop.  So if we define numjobs in the fio command (i.e. to be something other than 1), then max_jobs will be set to numjobs, which would be wrong - we want max_jobs to be set to the total number of fio processes.  Again see fio documentation for --max-jobs.  By adding numjobs to maxjobs for each client, we arrive at the correct value for max jobs, which is the total number of fio processes in the run.  Because of "directory=$target" line, fio will auto-generate one filename for each (host, job-number) pair.

This is essentially the same logic as what happens if numjobs is *not* defined (3 lines down) - in that case, we add 1 to max_jobs for each target in the target list, but the end result is the same - max_jobs is defined to be the total number of fio processes running in the test.

agent/bench-scripts/pbench-fio
portante: So are you then looking for `$max_jobs` to be the total number of jobs running across all hosts?

Is so, why not just count up the number of hosts and multiply by `$numjobs`?

bengland2: You could do that, but I was trying to make it easy for readers to see the similarity between the two cases, which are a) no numjobs parameter defined (defaulting to 1), but targets are defined, and b) only 1 target is defined, but numjobs is defined to be > 1 by maximizing the common code between them,   In either case, max_jobs is increased for each client host in the same way.

portante:
@atheurer, can you weigh in here when you have a chance?

portante:
@bengland2, are you still interested in solving this problem?

agent/bench-scripts/templates/fio.job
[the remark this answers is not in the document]
bengland2: I did this in latest commit to this PR.

agent/bench-scripts/templates/fio-shared-fs.job
bengland2: done in latest commit.

agent/bench-scripts/templates/fio-shared-fs.job
bengland2: done in latest commit.

agent/bench-scripts/pbench-fio
bengland2: @portante it's not clear why I should change this.

portante: This might be a long post, but seems worth to look at this in detail.  The code fragment in question is lines [570](https://github.com/distributed-system-analysis/pbench/pull/1150/files#diff-aa16cf8d4d3d68ccbdda0a744df85bdcR578) - [588](https://github.com/distributed-system-analysis/pbench/pull/1150/files#diff-aa16cf8d4d3d68ccbdda0a744df85bdcR588).

The sole purpose of this code fragment appears to be to fill in the `$client_opts` variable for use when the `fio.cmd` file is created at line [591](https://github.com/distributed-system-analysis/pbench/pull/1150/files#diff-aa16cf8d4d3d68ccbdda0a744df85bdcR591).  The variable is only filled in if there is a `$client_file` parameter or a `$clients` parameter.

If either parameter exists, in addition to adding the `--client` arguments, a `--max-jobs` argument is added.  The calculation of the value to use for the `--max-jobs` parameter is needs to be sorted out.

In the case where a `$client_file` parameter exists, the `$client_file` is given as the argument to one `--client=` parameter, and the value given to the `--max-jobs=` parameter is simply calculated to be the number of clients specified in the `$client_file`, regardless of the value of `$numjobs`.

In the case where a `$clients` parameter exists, one `--client=` parameter is added with a single client value pulled from the `$clients` list.  As each client value is considered, the value to use for the `--max-jobs=` parameter is calculated in two different ways:

 1. When `$numjobs` exists, that value is added to `$max_jobs` each time, effectively (# of clients) * (`$numjobs`)
 2. When `$numjobs` does not exist, the number of `$targets` is used as the value of `$numjobs`, which is effectively (# of clients) * (`$numjobs`)

It seems like this code fails to properly account for `$client_file` as a result.

Perhaps we can consider the following instead for this logic:

```
if [ ! -z $numjobs ]; then
    # so that a separate filename will be generated
    # for each (host, job-number) pair
    local local_numjobs=$numjobs
else
    local local_numjobs=0
    for t in `echo $targets | sed -e s/,/" "/g` ; do
        let local_numjobs=$local_numjobs+1
    done
fi
if [ ! -z "$client_file" ] ;then
    typeset -i nclients=$(wc -l $client_file | cut -d ' ' -f 1)
    local max_jobs=$nclients*$local_numjobs
    client_opts="--client=$client_file --max-jobs=$max_jobs"
elif [ ! -z "$clients" ]; then
    local nclients=0
    for client in `echo $clients | sed -e s/,/" "/g`; do
        client_opts="$client_opts --client=$client"
        let nclients=$nclients+1
    done
    local max_jobs=$nclients*$local_numjobs
    client_opts="$client_opts --max-jobs=$max_jobs"
fi
```

Finally, the question remains if this calculation is correct when **_not_** using the `directory` mode in the job file.

bengland2: used tabs instead of spaces to make it more readable.      I disagree with, or perhaps don't fully understand, your analysis, but I can't change it as you requested.    Your change goes from 1 line of code to 20, but it's not clear to me what it buys you.    My goal is not to fix pbench-fio for everyone, only to add distributed filesystem testing capability.  This 1-line change does that.

bengland2:
So I resubmitted the commits in the PR to condense them, make them easier to review.  The change to the bench/agent-scripts/templates/fio.job is gone.  This is absolutely the smallest PR I could produce at this point.   Let's review it together.

agent/bench-scripts/pbench-fio
bengland2: specifically, if there is only 1 client, the code reduces to exactly what it was before the change, since max_jobs is initialized to zero.

portante: > used tabs instead of spaces to make it more readable

The point of the code fragment was to make sure it fit compactly in the window.  Using 8 space tabs in the original file would be great for the comment lines, as well.

> it's not clear to me what it buys you

First, if you use a clients file instead of listing clients on the `pbench-fio` command line, it does not seem your change would work.

The point of the 20 lines of code change is to make sure the same behavior is consistent across the different ways clients are provided without repeating code (DRY principle).

Second, it seems that this behavior is tied to using `directory` in a job file, so the outstanding question that sill has to be answered is if this change works when NOT use `directory`.

agent/bench-scripts/pbench-fio
portante: Outdated comment, see discussion in more recent code comments.

portante:
Resolution from a recent discussion:

- [ ] @bengland2 to verify change works with existing fio clients

- [ ] Adding to documentation that shared job file is not tested with --numjobs parameter

bengland2:
Here's an example of what happens with pbench-fio commands tested with cephfs on my 3-VM cluster. 
  All commands use this prefix:

```
p="pbench-fio --sysinfo=none --samples=3 --max-failures=0 -c 192.168.121.64,192.168.121.112,192.168.121.158 --job-file=fio-shared-fs.job  -b 4 -t read -s 1G"
```

With command "

$p -d /mnt/cephfs --numjobs=1

```
[bengland@bene-laptop ceph-ansible]$ a all -a 'ls -l /mnt/cephfs'
ceph1 | CHANGED | rc=0 >>
total 3145728
-rw-r--r-- 1 root root 1073741824 May 21 18:02 192.168.121.112.TheJob.0.0
-rw-r--r-- 1 root root 1073741824 May 21 18:02 192.168.121.158.TheJob.0.0
-rw-r--r-- 1 root root 1073741824 May 21 17:20 192.168.121.64.TheJob.0.0

ceph3 | CHANGED | rc=0 >>
total 3145728
-rw-r--r-- 1 root root 1073741824 May 21 18:02 192.168.121.112.TheJob.0.0
-rw-r--r-- 1 root root 1073741824 May 21 18:02 192.168.121.158.TheJob.0.0
-rw-r--r-- 1 root root 1073741824 May 21 17:20 192.168.121.64.TheJob.0.0

...
```
Perfect.  However, with --numjobs=2 instead of 1, this happens:

```
[root@bene-laptop sample2]# more fio-result.txt 
<ceph2> fio: option <��`> outside of [] job section
<ceph1> fio: option <��`> outside of [] job section
<ceph3> fio: option <��`> outside of [] job section
```
and fio.job file is empty in the sample2/ directory.

same with this.   

$p  -d /mnt/cephfs --targets=f1,f2

but if I remove directory= from the job file and use this command:

$p targets=/mnt/cephfs/f1,/mnt/cephfs/f2

I get these files:

```
[bengland@bene-laptop ceph-ansible]$ a all -a 'ls -l /mnt/cephfs/*'
ceph2 | CHANGED | rc=0 >>
-rw-r--r-- 1 root root 1073741824 May 21 18:24 /mnt/cephfs/f2

ceph3 | CHANGED | rc=0 >>
-rw-r--r-- 1 root root 1073741824 May 21 18:24 /mnt/cephfs/f2

ceph1 | CHANGED | rc=0 >>
-rw-r--r-- 1 root root 1073741824 May 21 18:24 /mnt/cephfs/f2
```

This isn't right.   I expected /mnt/cephfs/192.168.121.{64,112,158}.f1.{0,1}  for a total of 6 files and 6 jobs.

bengland2:
I've tested pbench-fio with numjobs > 1 and now it works, because I got fio-shared-fs.job fixed.  I verified that it could create multiple fio job files per host this way.  So I think this commit is ready!    It took me a while because I uncovered a Ceph bug while working on it, but that in itself was useful.

bengland2:
ok, egg on my face, I tested a few extra boundary conditions (i.e. --client-file).  Some of this is related to what Peter was talking about.    I moved as much code outside the per-client loop as possible, the body of the loop is 1 line now.  Now it is working  for a variety of inputs and boundary conditions, IMO.

agent/bench-scripts/templates/fio.job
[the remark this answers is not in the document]
portante: This has been resolved by `fio.job` no longer being changed.

agent/bench-scripts/pbench-fio
```diff
@@ -567,22 +567,34 @@ function fio_run_job() {
...
+		       for t in `echo $targets | sed -e s/,/" "/g` ; do
+			       let max_jobs=$max_jobs+1
+		       done
+		fi
+
+		# make the fio command shorter by always using a client file
+		# regardless of how clients were specified
+		# we always save the client list in the benchmark results directory
+
+		client_file=$benchmark_results_dir/pbench-clients.list
```
portante: If `$client_file` is already defined, won't this override that global variable?  If so, can we add a comment saying that we are explicitly doing that?

Also, why re-write the file with the same contents?

portante:
@bengland2, I filed a PR against your github pbench repo with unit test fixes: https://github.com/bengland2/pbench/pull/3

agent/bench-scripts/pbench-fio
bengland2: "won't this override that global variable":  I suppose you would have preferred if I defined a new variable.   Context: I was trying to unify the handling of --client and --client-file.    When --client-file is parsed, the "clients" variable is initialized with the contents of the file.   So from then on we just deal with the "clients" variable.    But I didn't like running fio with each client passed individually on the command line, this makes fio command *very* long in the case of a large-scale test, so I decided to create a file containing the list of clients in the benchmark directory (so it is saved with test results), and do fio --client=$client_file instead.    It's very important that the contents of client_file be saved in the results directory since otherwise we have no record of this input to fio.

"why rewrite the file": So you wanted to copy the pathname in $client_file to $benchmark_results_dir/pbench-clients.list instead, but that directory didn't exist when the --client-file option was parsed, and the way I'm doing it correctly uses the contents of that file and saves the list of client host/ips in the results directory, so why not?

"rebase since it won't merge as is": perhaps it won't because of the regression test failures, but prior to today github was telling me that it was ok to merge it.    I'll look at the test failures and see what's going on today.  I'm reluctant to rebase (because this requires a complete retest) unless there is a good reason for it (i.e. git can't merge the code), again I'll take a look and see if that is the case.

bengland2:
@portante , it says right above the comment box that "This branch has no conflicts with the base branch -- Merging can be performed automatically".  It does say that the travis CI build failed, so I'll take a look at your PR and merge your stuff into my PR, hopefully this will resolve it.

agent/bench-scripts/pbench-fio
bengland2: I think the new code is consistent with the different ways clients are provided.  Also, I tested with an fio run on RBD block devices not using directory in a job file, and that correctly ran as well.  So I believe this is resolved.

portante:
@bengland2, not sure yet why GitHub won't allow the rebase and merge yet, but when I reviewed your code I checked it out, set the tracking branch to `upstream/master` and issued a `git rebase` to resolve the conflicts.

While reviewing this code last night I found a larger problem with `pbench-fio`.  It turns out that we should not be using `$targets` in the calculation because when a `--job-mode` of serial is used, the total number of targets is not considered.

I'll be posting a rather larger PR that contains multiple commits that incrementally fix a series of small problems and cleanups that might make all this easier.

bengland2:
so [this test](https://travis-ci.org/distributed-system-analysis/pbench/jobs/539211238) passed and it says this branch has no conflicts with the base branch, so can we merge it?  This doesn't mean we can't make pbench-fio better in the future, but we shouldn't hold it back because of problems that were in the code before the PR.

portante:
It still says I cannot rebase and merge this. Can you rebase to latest master first?

bengland2:
@portante I rebased, but this test failed, "Agent unit tests FAILED".  It's not obvious to me that the test failure was caused by my code.   Since the last one succeeded, I think the rebasing was what broke it.  I also squashed the commits FYI.

portante:
@bengland2, it looks like the unit tests files did not make it across the rebase. I'll take a look in a bit.

portante:
Closing in favor of #1224, where this code has been merged into those larger fixes (covers proper use of `$targets`), and unit tests are working along side additional unit tests pulled from `pbench-fio.md` documentation.

PR #1153, opened 2019-04-02
Refactor server commands to use single pbench-base

portante:
Prior to this commit, there we had two sources of truth for reading and defining pbench environment variables from the server config files:

  1. the original `pbench-base.sh`
  2. the new Python3 `PbenchConfig` class in `server/lib/pbench/__init__.py`

With future work coming where Bash scripts will be re-written in Python3, this duplication of truth will eventually bite us.  To avoid problems going forward, we are moving to one source of truth for pbench server environment variables: the Python3 `PbenchConfig` class

Since we mostly use Bash at this point, we create a new Python3 program, `pbench-base.py`, which is responsible for reading in the server config file and placing all the common variables found into the environment, and then exec'ing the target shell script.

To do this, we first create a "trampoline" program, `pbench-trampoline`, which is responsible for properly invoking `pbench-base.py`.  All server commands have been renamed to `*.sh` or `*.py`, as appropriate, and then their original names are symlink'd to `pbench-trampoline`.  This trampoline script serves two functions:

  1. It properly determines in one place the right way to invoke Python3 (python3 command where available, SCL everywhere else)
  2. It determines which server commands should invoke `pbench-base.py` and which should invoke Python3 directly

We still retain the original `pbench-base.sh` script, but it no longer defines any environment variables, it just defines helper functions for server bash scripts.

There are two existing Python programs which are invoked directly from the trampoline, both renamed to *.py:

  1. `pbench-report-status`
  2. `pbench-satellite-state-change` (re-written for Python3)

We had to modify the mock'd `ssh` command script to unset the `CONFIG` environment variable to properly mimic the normal ssh environment where on the remote host no environment variables from the invoking host are present.

We used to explicitly export the `CONFIG` environment variable in `pbench-sync-package-tarballs` since it is being invoked remotely. This is no longer necessary, as `pbench-base.py` now calculates the default location for the `CONFIG` file location using the same algorithm that `pbench-sync-package-tarballs` did.  This also allows us to remove the explicit definition of the `CONFIG` variables from the crontabs in the future.


portante:
@ndokos, rebased on the latest master, should be good for review now.

server/bin/pbench-index.sh
```diff
@@ -0,0 +1,231 @@
...
+
+    # make sure that all the relevant state directories exist
+    mk_dirs $hostname
+
+    # Ensure we can create the TMPDIR used for extracted tar balls each time
+    # we call index-pbench to help ensure everything operates cleanly and we
+    # only have one tar ball extracted at a time.
+    mkdir $TMPDIR || doexit "Bad $TMPDIR"
+
+    python3 ${PROGLOC}/bin/index-pbench --config ${PROGLOC}/lib/config/pbench-server.cfg -E $errors_json ${@} $link
```
ndokos: How do you know there is a python3? What if this is an SCL system?

ndokos: NM - I think I got it.

PR #1159, opened 2019-04-03
Launch run comparison across multiple controllers using search view

gurbirkalsi:
This PR modifies redux store logic for tracking a single selected controller or multiple selected controllers for visualization throughout the pbench dashboard. In addition, this PR improves the search view UX to apply or reset filters once the user has made their selection on the filters pane.

Users are now able to select multiple results from their search query and launch a run to run comparison that may have originated from more than one controller.

### DIFF
--- a/web-server/v0.4/src/models/search.js
+++ b/web-server/v0.4/src/models/search.js
@@ -53,13 +53,16 @@ export default {
         let searchResults = {};
         searchResults['resultCount'] = response.hits.total;
         let parsedResults = [];
-        response.hits.hits.map(result => {
+        response.hits.hits.map((result) => {
           let parsedResult = {};
           selectedFields.map(field => {
             parsedResult[field] = result._source[field.split('.')[0]][field.split('.')[1]];
           });
           if (typeof result._source['run']['prefix'] != 'undefined') {
             parsedResult['run.prefix'] = result._source['run']['prefix'];
+          }
+          if (typeof result._source['run']['id'] !== 'undefined') {
+            parsedResult['key'] = result._source['run']['id'];
           }
           parsedResults.push(parsedResult);
         });
--- a/web-server/v0.4/src/pages/ComparisonSelect/index.js
+++ b/web-server/v0.4/src/pages/ComparisonSelect/index.js
@@ -9,7 +9,7 @@ import { parseIterationData } from '../../utils/parse';
 import { queryIterations } from '../../services/dashboard';
 
 @connect(({ global, dashboard, loading }) => ({
-  selectedController: dashboard.selectedController,
+  selectedControllers: dashboard.selectedControllers,
   selectedResults: dashboard.selectedResults,
   iterations: dashboard.iterations,
   results: dashboard.results,
@@ -100,7 +100,7 @@ class ComparisonSelect extends ReactJS.Component {
 
   compareIterations = params => {
     const { configData } = this.state;
-    const { results, selectedController, selectedResults } = this.props;
+    const { results, selectedControllers, selectedResults } = this.props;
     const { dispatch } = this.props;
     const configCategories = Object.keys(configData);
 
@@ -125,7 +125,6 @@ class ComparisonSelect extends ReactJS.Component {
           configCategories: configCategories,
           configData: configData,
           results: results,
-          controller: selectedController,
           selectedResults: selectedResults,
         },
       })
@@ -134,7 +133,7 @@ class ComparisonSelect extends ReactJS.Component {
 
   compareAllIterations = () => {
     const { responseDataAll, configData } = this.state;
-    const { results, selectedController, selectedResults } = this.props;
+    const { results, selectedControllers, selectedResults } = this.props;
     const { dispatch } = this.props;
     const configCategories = Object.keys(configData);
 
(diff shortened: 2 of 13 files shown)


web-server/v0.4/src/pages/ComparisonSelect/index.js
```diff
@@ -266,7 +264,7 @@ class ComparisonSelect extends ReactJS.Component {
     }
 
     return (
-      <PageHeaderLayout title={selectedController}>
+      <PageHeaderLayout title={Array.isArray(selectedControllers) ? selectedControllers.join(', ') : selectedControllers}>
```
portante: When is it not an array?

gurbirkalsi: selectedControllers is not an array when the user follows the workflow of selecting a controller from the controllers view before navigating through the rest of the dashboard. We initialize selectedControllers as an array in the dashboard Redux model and overwrite it as a string when this case is met.

portante: Can we change that so that it is always an array, even if it is just one element?

portante:
And can you add unit tests to verify these changes?

web-server/v0.4/src/pages/ComparisonSelect/index.js
gurbirkalsi: I've updated the data model to expect an array at every point we render the selected controllers.

gurbirkalsi:
@portante I've updated the unit tests for testing the rendering of multiple selected controllers with mock props.

web-server/v0.4/src/models/search.js
```diff
@@ -53,11 +53,12 @@ export default {
         let searchResults = {};
         searchResults['resultCount'] = response.hits.total;
         let parsedResults = [];
-        response.hits.hits.map(result => {
+        response.hits.hits.map((result, index) => {
           let parsedResult = {};
           selectedFields.map(field => {
             parsedResult[field] = result._source[field.split('.')[0]][field.split('.')[1]];
           });
+          parsedResult['key'] = index;
```
portante: Can you assert that we have an actual string for the index name here?

web-server/v0.4/src/models/search.js
```diff
@@ -53,11 +53,12 @@ export default {
         let searchResults = {};
         searchResults['resultCount'] = response.hits.total;
         let parsedResults = [];
-        response.hits.hits.map(result => {
+        response.hits.hits.map((result, index) => {
```
portante: If `response.hits.hits` does not have a `result` and an `index` field, what happens here?

web-server/v0.4/src/pages/ComparisonSelect/index.js
```diff
@@ -266,7 +264,7 @@ class ComparisonSelect extends ReactJS.Component {
     }
 
     return (
-      <PageHeaderLayout title={selectedController}>
+      <PageHeaderLayout title={selectedControllers.join(', ')}>
```
portante: How many controllers can be selected?  If there are many, how useful is the title?

Could we instead shorten the list with an elipsis or something?

web-server/v0.4/src/models/search.js
gurbirkalsi: The `map()` method takes the array of hits and will always return the value which is designated as `result` in this case in addition to the array `index` currently being iterated on. If `response.hits.hits` is an empty array, the `map()` method would be skipped and as a result the reducer would contain its default value for search results. This would then render an empty table in the search view showing a "No data found" message in the table which could mean the user had a bad query.

web-server/v0.4/src/models/search.js
gurbirkalsi: I've updated the key assignment to take the result id if the field exists. Since we have stable IDs for rendering items, we can remove the index field referenced by the `map()` method.

web-server/v0.4/src/pages/ComparisonSelect/index.js
gurbirkalsi: The user can select as many controllers as they choose and the PageHeaderLayout component will wrap the title appropriately. The title may be useful to confirm the controllers the user selected but we should eventually display a controller tag next to each result selected. 

Did you see a different behavior with many controllers selected while testing?

web-server/v0.4/src/models/search.js
portante: Can you open an issue to check for errors when response has no `hits`, and/or when it does not have a  `result` or `index` field?

web-server/v0.4/src/pages/ComparisonSelect/index.js
portante: I did not try many controllers, but it would be good to know what the expected behavior is and make sure this is worth doing.

portante: Perhaps we should open an issue to track this?

gurbirkalsi: I've opened issue #1166 to track this behavior.

web-server/v0.4/src/models/search.js
gurbirkalsi: I've opened issue #1167 to track this behavior. Ideally, this would encompass all cases where we would potentially continue with post processing after receiving either empty data or a network error.

PR #1163, opened 2019-04-10
Runbench cdmv4

atheurer:
This PR includes an update to CDM, sample query scripts for CDM, and a postprocess-only mode for pbench-run-benchmark, which we can use to defer postprocessing on the server (once that is automated, we can disable postprocessing when running under pbench-agent).

A follow-up PR will be made to sync up to Sysstat v12


atheurer:
I am adding do-not-merge label until this is ready.  Still going to target 0.59, but I want to get all sysstat cdm processing in this PR before merging.

PR #1165, opened 2019-04-10
Improve dashboard filtering UX

gurbirkalsi:
Improve filtering UX with new component for managing filter state and interaction with Table component.

Moves parsing logic to utility definition for use with
filtering updates.

### DIFF
--- a/web-server/v0.4/src/utils/parse.js
+++ b/web-server/v0.4/src/utils/parse.js
@@ -1,3 +1,6 @@
+import cloneDeep from 'lodash/cloneDeep';
+import isMatch from 'lodash/isMatch';
+
 const containsKey = (columns, item) => {
   var contains = false;
   for (var column in columns) {
@@ -81,17 +84,8 @@ export const parseIterationData = results => {
 
     var columns = [
       {
-        title: '#',
-        dataIndex: 'iteration_number',
-        fixed: 'left',
-        width: 50,
-        key: 'iteration_number',
-        sorter: (a, b) => a.iteration_number - b.iteration_number,
-      },
-      {
         title: 'Iteration Name',
         dataIndex: 'iteration_name',
-        fixed: 'left',
         width: 150,
         key: 'iteration_name',
       },
@@ -174,17 +168,17 @@ export const parseIterationData = results => {
             if (port_val != undefined && !ports.includes(port_val)) {
               ports.push(port_val);
             }
-            var columnTitle =
+            var columnHost =
               'client_hostname:' +
               curr_iter_type[iterationNetwork][iterationData].client_hostname +
               '-server_hostname:' +
               curr_iter_type[iterationNetwork][iterationData].server_hostname +
               '-server_port:' +
               port_val;
-            var _columnPrefix = iterationType + '-' + iterationNetwork + '-' + columnTitle + '-';
-            var columnMean = _columnPrefix + 'mean';
-            var columnStdDev = _columnPrefix + 'stddevpct';
-            var columnSample = _columnPrefix + 'closestsample';
+            var columnKey = iterationType + '-' + iterationNetwork + '-' + columnHost + '-';
+            var columnMean = columnKey + 'mean';
+            var columnStdDev = columnKey + 'stddevpct';
+            var columnSample = columnKey + 'closestsample';
             if (constructChildCol) {
               if (columns[parentColumnIndex]['children'] == undefined) {
                 var childColumnIndex = 0;
@@ -195,14 +189,14 @@ export const parseIterationData = results => {
                 );
               }
               var child_column = columns[parentColumnIndex].children[childColumnIndex];
-              if (!containsIteration(child_column, columnTitle)) {
+              if (!containsIteration(child_column, columnHost)) {
                 if (child_column['children'] == undefined) {
-                  child_column['children'] = [{ title: columnTitle, dataIndex: columnTitle }];
+                  child_column['children'] = [{ title: columnHost, dataIndex: columnKey }];
                 } else {
-                  child_column['children'].push({ title: columnTitle });
+                  child_column['children'].push({ title: columnHost });
                 }
               }
-              var dataChildColumnIndex = getColumnIndex(child_column['children'], columnTitle);
+              var dataChildColumnIndex = getColumnIndex(child_column['children'], columnHost);
               if (dataChildColumnIndex == undefined) {
                 dataChildColumnIndex = 0;
               }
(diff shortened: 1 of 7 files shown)


gurbirkalsi:
@portante I've reverted the formatting changes that somehow got into the commit. I must have used Prettier to format while working on these changes. We can consider implementing a pre-commit hook to lint the staged files before pushing.

web-server/v0.4/src/utils/parse.js
```diff
@@ -290,3 +283,38 @@ export const parseIterationData = results => {
...
+  for (var response in responseDataCopy) {
+    const responseIterations = responseDataCopy[response].iterations;
+    let selectedConfigLength = Object.keys(selectedConfig).length;
+    if (selectedConfigLength > 0) {
+      const filteredResponseData = [];
+      for (let iteration in responseIterations) {
+        let found = [];
+        for (let config in selectedConfig) {
+          if (
+            (selectedConfig[config] !== undefined) &
```
portante: s/&/&&/ ?

web-server/v0.4/src/utils/parse.js
```diff
@@ -290,3 +283,38 @@ export const parseIterationData = results => {
...
+            (selectedConfig[config] == responseIterations[iteration][config])
+          ) {
+            found.push(true);
```
portante: Since we are only using `found.length` below which is a count, should we just use a counter instead of an array here?

web-server/v0.4/src/utils/parse.js
```diff
@@ -290,3 +283,38 @@ export const parseIterationData = results => {
...
+    responseDataCopy[response].columns = cloneDeep(responseData[response].columns);
+    responseDataCopy[response].iterations = cloneDeep(responseData[response].iterations);
+    responseDataCopy[response].resultName = cloneDeep(responseData[response].resultName);
+  }
+
+  for (var response in responseDataCopy) {
+    const responseIterations = responseDataCopy[response].iterations;
+    let selectedConfigLength = Object.keys(selectedConfig).length;
+    if (selectedConfigLength > 0) {
+      const filteredResponseData = [];
```
portante: Perhaps consider renaming to `filteredResponseIterationData`?

web-server/v0.4/src/utils/parse.js
```diff
@@ -290,3 +283,38 @@ export const parseIterationData = results => {
...
+  for (var response in responseData) {
+    responseDataCopy[response] = [];
+    responseDataCopy[response].columns = cloneDeep(responseData[response].columns);
+    responseDataCopy[response].iterations = cloneDeep(responseData[response].iterations);
+    responseDataCopy[response].resultName = cloneDeep(responseData[response].resultName);
+  }
+
+  for (var response in responseDataCopy) {
+    const responseIterations = responseDataCopy[response].iterations;
+    let selectedConfigLength = Object.keys(selectedConfig).length;
```
portante: The calculation of the length of keys for `selectedConfig` can be done once before the `for` loop at line 296, right?

portante: Then we can also move line 299 outside of the `for` loop as well since we don't do anything if there are not selected configs, right?

web-server/v0.4/src/utils/parse.js
```diff
@@ -197,7 +190,7 @@ export const parseIterationData = results => {
               var child_column = columns[parentColumnIndex].children[childColumnIndex];
               if (!containsIteration(child_column, columnTitle)) {
                 if (child_column['children'] == undefined) {
-                  child_column['children'] = [{ title: columnTitle, dataIndex: columnTitle }];
+                  child_column['children'] = [{ title: columnTitle, dataIndex: _columnPrefix }];
```
portante: Why does `_columnPrefix` work here?  Seems like that is just a local variable used to avoid duplicate code up at line 177.

web-server/v0.4/src/utils/parse.js
gurbirkalsi: Good catch, we should be using && here.

web-server/v0.4/src/utils/parse.js
gurbirkalsi: Yes! Using a counter here will improve performance and save memory being allocated.

web-server/v0.4/src/utils/parse.js
gurbirkalsi: I agree

web-server/v0.4/src/utils/parse.js
gurbirkalsi: Yes, and we can initialize `selectedConfigLength` as a const since it is never reassigned. We can also move our `if (selectedConfigLength > 0)` logic outside of the `for` loop.

web-server/v0.4/src/utils/parse.js
gurbirkalsi: We need `_columnPrefix` here as opposed to `columnTitle` as there were console warnings for potential duplicate keys given a dataIndex of `columnTitle`. By using the whole column prefix, we ensure that table cells have unique dataIndex values by referencing the iteration primary metric and network type they belong to.

gurbirkalsi: Not sure why we have an underscore in `_columnPrefix` but I'll modify that so we maintain the camel case conventions we have throughout here.

portante: I believe I wrote that code to indicate that it was a local temporary variable not intended to be used elsewhere.  It was done to avoid repeating code multiple times (DRY principle).

So why that would help still is a mystery.

gurbirkalsi: It could be possible that a recent update to antd's library modified the behavior of tables and required unique dataIndex values which is why we were not receiving those warnings at that time. Here's the warning for reference: 

`Warning: Encountered two children with the same key, "client_hostname:100.100.100.138-server_hostname:100.100.100.140-server_port:20010". Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.`

You'll see here we only use `columnTitle` which only contains hostname information whereas the `columnPrefix` contains both the information in `columnTitle` and the iteration type and network. This causes potential issues for iteration types and networks that are indexed with similar hostname data.

portante: Okay, so we should rename `columnPrefix` since it is not just a prefix, but a unique identifier.

gurbirkalsi: I've renamed the reference here to `columnKey` which references the `columnHost` information we collect. I've also included a cleaner solution for maintaining immutability of result iterations data when filtering and improved the naming of references for better oversight into what is being copied and filtered vs. the original source reference.

PR #1170, opened 2019-04-16
Restructure selected dashboard components in redux store

gurbirkalsi:
For handling persistence of the redux store, selected configuration
data needs to be stored in the global reducer while the dashboard
namespace is blacklisted to avoid local caching.

Previously, caching of the dashboard namespace was causing conflicts
with an edge case where server side data was being updated but was not
reflected in persisted user sessions.

### DIFF
--- a/web-server/v0.4/src/app.js
+++ b/web-server/v0.4/src/app.js
@@ -2,10 +2,17 @@ import { createStore } from 'redux';
 import { persistStore, persistReducer } from 'redux-persist';
 import storage from 'redux-persist/lib/storage';
 
+/*
+* redux persist configuration for saving state object to persisted storage.
+*
+* `dashboard` and `search` are blacklisted from the saved state as the namespaces 
+* persist data that is constantly updated on the server side. 
+*/
 const persistConfig = {
   timeout: 1000,
   key: 'root',
   storage,
+  blacklist: ['dashboard', 'search'],
 };
 
 const persistEnhancer = () => createStore => (reducer, initialState, enhancer) => {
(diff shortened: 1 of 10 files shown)


web-server/v0.4/src/app.js
```diff
@@ -6,6 +6,7 @@ const persistConfig = {
   timeout: 1000,
   key: 'root',
   storage,
+  blacklist: ['dashboard'],
```
portante: What does it mean to "blacklist" exactly?  Can you add a comment describing why the `dashboard` data is black listed here?

web-server/v0.4/src/models/global.js
```diff
@@ -8,6 +8,8 @@ export default {
     datastoreConfig: {},
     indices: [],
     selectedIndices: [],
+    selectedResults: [],
```
portante: Does having selected indices and results in the global redux store mean that it will be shared between all pages that import global?  And if so, how might that affect behaviors between pages?

gurbirkalsi: That's correct - `global` is another redux namespace used to track information that is shared globally throughout the dashboard. For example, we use the `selectedResults` state in the Results, ComparisonSelect, and RunComparison page components. This doesn't change the behaviors of pages as the page components are now just pulling this information from a separate key in the redux store.

web-server/v0.4/src/app.js
gurbirkalsi: blacklisting allows us to prevent certain redux store namespaces from being persisted. In this case, we want to avoid persisting the dashboard namespace as server side information is constantly changing and we avoid pulling new data if the information has already been persisted.

gurbirkalsi: I've also added the search namespace to this config as the search view renders information that is dependent on server side information being modified or updated.

PR #1175, opened 2019-04-19
Matrix visualization for increase/decrease in percentage for comparison cluster

Anisha1234:
- This PR includes visualization for describing the percent differences across generated clusters for a run comparison. 
- This will enhance the summary bar graph displayed showing mean values for each cluster.

main7

### DIFF
--- a/web-server/v0.4/src/pages/RunComparison/index.js
+++ b/web-server/v0.4/src/pages/RunComparison/index.js
@@ -15,6 +15,8 @@ import {
   Input,
   Switch,
   message,
+  Collapse,
+  Icon,
 } from 'antd';
 import { ResponsiveBar } from '@nivo/bar';
 import DescriptionList from 'ant-design-pro/lib/DescriptionList';
@@ -22,12 +24,13 @@ import jsPDF from 'jspdf';
 import html2canvas from 'html2canvas';
 import moment from 'moment';
 import PageHeaderLayout from '../../layouts/PageHeaderLayout';
-import { generateIterationClusters } from '../../utils/parse';
+import { generateIterationClusters, getComparisonColumn } from '../../utils/parse';
 import TimeseriesGraph from '@/components/TimeseriesGraph';
 
 const { Description } = DescriptionList;
 const { TabPane } = Tabs;
 const FormItem = Form.Item;
+const { Panel } = Collapse;
 
 const columns = [
   {
@@ -102,6 +105,7 @@ class RunComparison extends React.Component {
       selectedComponents: [],
       pdfHeader: '',
       pdfName: '',
+      column: [],
     };
   }
 
@@ -131,6 +135,10 @@ class RunComparison extends React.Component {
         this.setState({ loadingClusters: false });
         this.setState({
           ...timeseriesData,
+        });
+        const column = getComparisonColumn(iterationClusters.maxIterationLength);
+        this.setState({
+          column,
         });
       });
     });
@@ -266,6 +274,7 @@ class RunComparison extends React.Component {
       loadingPdf,
       exportModalVisible,
       loadingClusters,
+      column,
     } = this.state;
     const { selectedControllers, selectedResults, iterationParams } = this.props;
 
@@ -509,9 +518,26 @@ class RunComparison extends React.Component {
                         />
                       </Col>
                     </Row>
+                    <Collapse
+                      bordered
+                      defaultActiveKey={['1']}
+                      expandIcon={({ isActive }) => (
+                        <Icon type="caret-right" rotate={isActive ? 90 : 0} />
+                      )}
+                    >
+                      <Panel header="Percent Differences Table" key="1">
+                        <Table
+                          columns={column}
+                          dataSource={clusteredGraphData[cluster]}
+                          pagination={{ pageSize: 50 }}
+                          scroll={{ x: 1500, y: 240 }}
+                          bordered
+                        />
+                      </Panel>
+                    </Collapse>
                     <br
                       style={{
-                        borderTopWidth: 1,
+                        borderTopWidth: 3,
                         borderStyle: 'solid',
                         borderColor: '#8c8b8b',
                       }}
--- a/web-server/v0.4/src/utils/parse.js
+++ b/web-server/v0.4/src/utils/parse.js
@@ -1,4 +1,6 @@
 import _ from 'lodash';
+import { Icon } from 'antd';
+import React from 'react';
 
 export const parseIterationData = results => {
   const iterations = [];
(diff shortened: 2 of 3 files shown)


portante:
@Anisha1234, we have removed the "`[WIP]`" text from the title of this PR since we have a "`WIP`" label that we use now instead.

Can you rebase  this PR and squash the commits down to one as well?
 
We have moved this PR to the 0.60 release since we are working to push 0.59 out the door shortly.

Anisha1234:
> @Anisha1234, we have removed the "`[WIP]`" text from the title of this PR since we have a "`WIP`" label that we use now instead.
> 
> Can you rebase this PR and squash the commits down to one as well?
> 
> We have moved this PR to the 0.60 release since we are working to push 0.59 out the door shortly.

@portante Ihave rebased and squashed the commits down to one. Moreover the work is ready to be reviewed so removed "WIP" label.

gurbirkalsi:
@Anisha1234 I'm seeing NaN values being rendered as part of the percent differences, can you take a look at this error? 
Pbench_Dashboard

gurbirkalsi:
Can we consider making the percent differences table toggle-able in order to keep the initial rendering of the run comparison page focused on the clustering graphs and matching configurations?

web-server/v0.4/src/pages/RunComparison/index.js
```diff
@@ -295,9 +297,73 @@ class RunComparison extends React.Component {
       loadingPdf,
       exportModalVisible,
       loadingClusters,
+      iterationNo,
     } = this.state;
     const { selectedControllers, selectedResults, iterationParams } = this.props;
 
+    const children = [];
```
gurbirkalsi: This logic for generating the column data should be moved to `parse.js` similar to how we invoke `onGenerateIterationClusters` in the event of a cluster change on line 179.

Anisha1234:
@gurbirkalsi I have fixed the suggested changes. Please have a look at it and let me know if any further changes are required.

web-server/v0.4/src/utils/parse.js
```diff
@@ -299,3 +306,65 @@ export const generateIterationClusters = (config, iterations) => {
 
   return parseClusteredIterations(primaryMetricIterations, clusterLabels, selectedConfig);
 };
+
+export const getComparisionColumn = (record, iterationNo) => {
```
gurbirkalsi: It seems we have a spelling error here - did you mean `getComparisonColumn`?

web-server/v0.4/src/utils/parse.js
```diff
@@ -224,6 +227,10 @@ export const parseClusteredIterations = (clusteredIterations, clusterLabels, sel
               return false;
             })
           ];
+        const percentage =
+          ((clusterObject[iteration] - clusterObject[0]) / clusterObject[iteration]) * 100;
```
gurbirkalsi: The calculation of percent difference is missing one aspect of the formula where the average of both numbers is used as the divisor. The formula for calculating percent difference is as follows:
`|x1 - x2|/((x1 + x2)/2) * 100`

gurbirkalsi:
@Anisha1234 the percent differences don't seem to be rendered here - are you able to reproduce this on your environment? Pbench_Dashboard

Anisha1234:
> @Anisha1234 the percent differences don't seem to be rendered here - are you able to reproduce this on your environment? Pbench_Dashboard

@gurbirkalsi  I have fixed the issue. It was due to the calculation of no if iterations.

web-server/v0.4/src/utils/parse.js
```diff
@@ -300,3 +316,74 @@ export const generateIterationClusters = (config, iterations) => {
 
   return parseClusteredIterations(primaryMetricIterations, clusterLabels, selectedConfig);
 };
+
+export const getComparisonColumn = (record, iterationNo) => {
+  const children = [];
+  for (let i = 0; i < iterationNo; i += 1) {
```
gurbirkalsi: Can we consider using `forEach()` here to maintain the ES6 usage throughout the codebase? For more on `forEach()`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

web-server/v0.4/src/utils/parse.js
```diff
@@ -300,3 +316,74 @@ export const generateIterationClusters = (config, iterations) => {
...
+    children.push({
+      title: `Iteration - ${i}`,
+      dataIndex: '0',
```
gurbirkalsi: In general, we should set unique values here for `dataIndex` and `key` since we are iterating over the length of iterations for a cluster.

web-server/v0.4/src/utils/parse.js
```diff
@@ -300,3 +316,74 @@ export const generateIterationClusters = (config, iterations) => {
...
+                    theme="filled"
+                    style={{ marginLeft: '3%', color: 'green' }}
+                  />
+                </div>
+              );
+            }
+            if (text === 0) {
+              return `${text}%`;
+            }
+            if (text === undefined) {
```
gurbirkalsi: Can this edge case be caught during the parsing phase so we don't need to account for this during the rendering phase?

web-server/v0.4/src/utils/parse.js
```diff
@@ -300,3 +316,74 @@ export const generateIterationClusters = (config, iterations) => {
...
+                  <Icon
+                    type="caret-up"
+                    theme="filled"
+                    style={{ marginLeft: '3%', color: 'green' }}
+                  />
+                </div>
+              );
+            }
+            if (text === 0) {
+              return `${text}%`;
```
gurbirkalsi: What if we rendered a dash icon here when the percentage is equal to 0 to show this is a baseline representation of the percent difference?

web-server/v0.4/src/utils/parse.js
```diff
@@ -207,6 +209,7 @@ export const parseClusteredIterations = (clusteredIterations, clusterLabels, sel
   const clusteredGraphData = [];
   const graphKeys = [];
   const tableData = [];
+  let noOfIteration = 0;
```
gurbirkalsi: Can we consider naming this variable `maxClusterLength` to better describe this value store as the max number of iterations found out of all clusters?

web-server/v0.4/src/utils/parse.js
Anisha1234: @gurbirkalsi , the forEach() method executes a provided function for each array element. And here we have to iterate over an integer number. So should we consider using recursion here?

gurbirkalsi: Sorry, I thought `iterationNo` was an array here - this looks good!

gurbirkalsi:
@Anisha1234 Once we resolve the build error, we should be ready to merge this PR- great work Anisha!

portante:
Removing the WIP label as this has been reviewed once already.

web-server/v0.4/package.json
```diff
@@ -118,4 +118,4 @@
     "last 2 versions",
     "not ie <= 10"
   ]
-}
+}
```
portante: Why did we remove the trailing newline of this file in this commit?

PR #1178, opened 2019-04-25
Integrate Husky for pre-commit git hook management via package.json

gurbirkalsi:
Husky automatically creates git hook scripts as part of `.git/hooks`
when the developer runs `yarn` to install package dependencies. Husky
will trigger `npm run lint-staged` which runs both prettier and linting
commands configured by their respective `.prettierrc` and `.eslintrc`
files.

### DIFF
--- a/web-server/v0.4/package.json
+++ b/web-server/v0.4/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "pbench_dashbaord",
+  "name": "pbench_dashboard",
   "version": "2.0.0",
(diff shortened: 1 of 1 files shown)


web-server/v0.4/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "pbench_dashbaord",
+  "name": "pbench_dashboard",
```
gurbirkalsi: @portante Seems we had a spelling issue which I've also resolved with this PR.

PR #1181, opened 2019-04-25
Integrate umi-request for handling network request errors

gurbirkalsi:
umi-request combines fetch and axios libraries to provide features
such as caching, timeout, and error handling.

`request.js` now implements a custom error handler to extend the default
handler shipped with umi-request for redirecting the user to the
appropriate exception pages.

### DIFF
--- a/web-server/v0.4/src/utils/request.js
+++ b/web-server/v0.4/src/utils/request.js
@@ -1,7 +1,6 @@
-import fetch from 'dva/fetch';
+import { extend } from 'umi-request';
 import { notification } from 'antd';
-import { routerRedux } from 'dva/router';
-import store from '../global';
+import router from 'umi/router';
 
 const codeMessage = {
   200: 'The server successfully returned the requested data. ',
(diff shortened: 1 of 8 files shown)


web-server/v0.4/src/utils/request.js
```diff
@@ -20,79 +19,32 @@ const codeMessage = {
...
+  if (status === 403) {
+    router.push('/exception/403');
+    return;
+  }
+  if (status <= 504 && status >= 500) {
+    router.push('/exception/500');
+    return;
   }
+  if (status >= 404 && status < 422) {
+    router.push('/exception/404');
```
portante: So if we get a 404 for a `result.json` file, will this cause the user to go to another page?

gurbirkalsi: Correct; however, this doesn't seem to be the most ideal UX in this case. Ideally, we should redirect the user back to their previous "state" in the dashboard workflow. For example, if the user receives a 404 for a `result.json` file when navigating to the Summary page, they would be redirected back to the Results page. Since we receive this error within the component lifecycle of `componentDidMount()`, the user will see a loading indicator until the asynchronous network request has been completed and the request error handler is able to redirect the user back. 

I've included this solution in a separate commit and also removed `axios` as a dependency as `umi-request` is now the default networking library used throughout the dashboard.

portante: Have you thought about a simple pop-up window with the errors, with some context?

portante:
@gurbirkalsi, can you get the unit tests working before we review officially?

gurbirkalsi:
Yes! There seems to be an issue thrown by the jest test suite which is caused by importing `umi/router`. I've tried adjusting the pattern matching for reading the node module; however, the issue is persisting on the Travis CI build whereas it is passing during the local test.

web-server/v0.4/src/utils/request.js
gurbirkalsi: Dashboard_-_Pbench_Dashboard

Yes! This is a pop-up window shown before redirecting the user back to their previous state within the dashboard session.

portante:
@gurbirkalsi, seems like this needs a rebase now that other changes were merged first.

portante:
@gurbirkalsi, something is failing in the unit tests.  Five of them failed because of external ubuntu repos not being available, but the last 3 are failing in the unit tests.  Can you check it out?

gurbirkalsi:
It seems requests to the npm registry are timing out - we'll need to add a longer timeout during the install phase (`--network-timeout 1000000`)

PR #1185, opened 2019-04-30
Trafficgen updates

k-rister:


agent/bench-scripts/postprocess/trafficgen-postprocess-cdm
```diff
@@ -23,11 +23,12 @@ my $rundir = shift(@ARGV);
 my $base_bench_dir = shift(@ARGV);
 
 my @trial_metrics = (
-    { 'key' => 'result',
+    { 'key' => 'flubberbubbles',
```
atheurer: @k-rister is this a technical term?

agent/bench-scripts/postprocess/trafficgen-postprocess-cdm
```diff
@@ -96,6 +96,270 @@ my @trial_stats_device_metrics = (
...
     );
 
+#    { 'key' => '',
+#      'subkey' => '',
+#      'field' => '',
+#      'class' => '',
+#      'type' => '',
+#      'name_format' => '',
+#      'extra_field' => '',
+#      'cumulative' => 0 },
```
atheurer: Can you remove the above?

k-rister: Sure

agent/bench-scripts/postprocess/trafficgen-postprocess-cdm
k-rister: Why of course.  Actually, this particular metric was an exception/one-off so I tagged it with something weird so it would never match.

k-rister: Note, this will probably change if we come up with the different metric value documents.  The difference is this is the pass/fail value as opposed to a double.

PR #1188, opened 2019-05-02
e2e testing for controller, results, and search pages using Puppeteer

gurbirkalsi:
Puppeteer is a Node library that provides an API to control the browser.
e2e testing covers all potiential user workflows related to a page
component and expectations for how page components should function.

Functionality provided by third party libraries are not covered within
these tests as we expect library maintainers to provide coverage for
functionality.

Before implementing e2e testing for the comparisonselect and
runcomparison page components, we require error handling for results
selected that are not supported by our comparison methodology (ex.
user-benchmark type). Once the error handler is integrated, we can write
generalized e2e tests using server data.

### DIFF
--- a/web-server/v0.4/README.md
+++ b/web-server/v0.4/README.md
@@ -22,6 +22,7 @@ Pbench Dashboard is a web-based platform for consuming indexed performance bench
 │   ├── pages                       # app page components and templates
 │   │   └── document.ejs            # HTML entry
 │   ├── services                    # redux services
+│   ├── e2e                         # e2e test definitions
 │   ├── utils                       # utility scripts
 │   ├── app.js                      # app theme configuration
 │   ├── global.js                   # global imports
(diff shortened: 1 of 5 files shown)


gurbirkalsi:
@portante I've included a pre-push hook for triggering e2e tests as a simpler and cleaner solution for launching tests that interact with our internal components (elasticsearch, results server, backend).

web-server/v0.4/unittests
```diff
@@ -6,6 +6,11 @@ if [[ $? -ne 0 ]]; then
   echo "yarn failed!" >&2
   exit 1
 fi
+yarn start &
+if [[ $? -ne 0 ]]; then
```
portante: Not sure this makes sense.  I don't think the value of `$?` is set to anything as the `yarn start` is just immediately put in the background because of the trailing `&` on the line.  You might want to check `$!` to see that it got a pid value, and that the pid is of the `yarn start`.

It is your intention to create the `yarn start` process as a background job?

gurbirkalsi: Correct, however this is not required now that we have moved e2e testing to the pre-push hook. Beforehand, `yarn start` was started as a background job to allow the e2e tests to run on the local environment while the local dev server runs.

gurbirkalsi: I've removed this block in the most recent commit.

PR #1193, opened 2019-05-07
Apply prettier configuration to all js source files

gurbirkalsi:
This PR establishes baseline styling conventions enforced by configurations defined in `.prettierrc` and automatically runs via the pre-commit hook integration with Husky. As a result of this integration, we should not see styling changes in future PRs as they are automatically written to staged files once the user has passed the JavaScript linting hook.


portante:
@gurbirkalsi, also needs a rebase due to other PRs being merged first.

PR #1196, opened 2019-05-10
Simplify server config files.

### LINKED ISSUE #1195: Simplify pbench-server-default.cfg by eliminating some definitions
`pbench-server-default.cfg` contains definitions for

- pbench-incoming-dir
- pbench-results-dir
- pbench-users-dir
- pbench-static-dir

that are all relative to `<pbench-top-dir>/public_html`. It might be the case that these directories can be all in unrelated areas of a file system or even on different file systems, but practically speaking that is not the case (and it's not clear if the code depends on them being together).

I propose that we elevate the above structure to a requirement, eliminate the above from the config file and define them in terms of `<pbench-top-dir>` wherever they occur.

They actually occur mostly in the activate scripts which we are trying to eliminate in favor of ansible playbooks. The underlying reason for wanting the tighter coupling is to make the ansible role that creates this structure simpler.

ndokos:
Fixes #1195.

Eliminate and consolidate variables in the server config files. See the individual commit for specific changes.

### DIFF
--- a/server/bin/pbench-server-activate-create-results-dir-structure
+++ b/server/bin/pbench-server-activate-create-results-dir-structure
@@ -20,36 +20,6 @@ if [ -z "$archive_dir" ] ;then
     exit 4
 fi
 
-pbench_html_dir=$(getconf.py pbench-html-dir pbench-server)
-if [ -z "$pbench_html_dir" ] ;then
-    echo "$prog: \"pbench-html-dir\" is undefined in section \"pbench-server\" of config file." >&2
-    exit 5
-fi
-
-pbench_incoming_dir=$(getconf.py pbench-incoming-dir pbench-server)
-if [ -z "$pbench_incoming_dir" ] ;then
-    echo "$prog: \"pbench-incoming-dir\" is undefined in section \"pbench-server\" of config file." >&2
-    exit 6
-fi
-
-pbench_results_dir=$(getconf.py pbench-results-dir pbench-server)
-if [ -z "$pbench_results_dir" ] ;then
-    echo "$prog: \"pbench-results-dir\" is undefined in section \"pbench-server\" of config file." >&2
-    exit 7
-fi
-
-pbench_users_dir=$(getconf.py pbench-users-dir pbench-server)
-if [ -z "$pbench_users_dir" ] ;then
-    echo "$prog: \"pbench-users-dir\" is undefined in section \"pbench-server\" of config file." >&2
-    exit 8
-fi
-
-pbench_static_dir=$(getconf.py pbench-static-dir pbench-server)
-if [ -z "$pbench_static_dir" ] ;then
-    echo "$prog: \"pbench-static-dir\" is undefined in section \"pbench-server\" of config file." >&2
-    exit 9
-fi
-
 logsdir=$(getconf.py pbench-logs-dir pbench-server)
 if [ -z "$logsdir" ] ;then
     echo "$prog: \"pbench-logs-dir\" is undefined in section \"pbench-server\" of config file." >&2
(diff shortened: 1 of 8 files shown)


server/bin/pbench-server-activate-create-results-dir-structure
```diff
@@ -90,14 +60,19 @@ typeset -i nerrs=0
 
 # it's OK if the directories exist already.
 mkdir -p ${archive_dir}
+
+pbench_incoming_dir=${pbench_dir}/public_html/incoming
```
portante: We seem to use `pbench_` prefixes to denote variables pulled directly from the config file.  Do we want to keep this nomenclature?

server/bin/pbench-server-activate-create-results-dir-structure
```diff
@@ -90,14 +60,19 @@ typeset -i nerrs=0
...
+pbench_results_dir=${pbench_dir}/public_html/results
```
portante: It would probably be good to consider continuing the DRY principle and define a variable for the `public_html` path rather than repeating that each time.

server/bin/pbench-server-activate-create-results-dir-structure
ndokos: I think that's too much trouble for this script: it already uses the pbench_ prefixed names and changing all of them is just a source of errors that I would rather avoid. Besides, this script will be going away after PR#1099 is merged.

server/bin/pbench-server-activate-create-results-dir-structure
ndokos: Theoretically, yes, but this script is going away soon. This PR is just an intermediate step to the ansible playbooks and, as a consequence, obsolescence for all the activate scripts (with the create_crontab exception),

server/bin/pbench-server-activate-create-results-dir-structure
portante: Not sure, but it seems these variables still exist in #1199.

server/bin/pbench-server-activate-create-results-dir-structure
portante: That would be great, but I still see these variables in #1199.

server/bin/pbench-server-activate-create-results-dir-structure
ndokos: I don't think so. I just did a global search for them and did not find them. I'm not sure whether I did it correctly, but I was trying to follow your instructions about making #1199 dependent on this PR. The commits in #1199 are a superset of the commits in this PR. So maybe you are looking at the "common" commits?

server/bin/pbench-server-activate-create-results-dir-structure
ndokos: See above for a possible explanation. If that's not correct, can you point out where you see them?

ndokos:
It depends on whether v0.60 will be an agent-only release or an agent-and-server release. But in any case, #1199 is a superset of this one (at least, the list of commits I see there is a superset of the list of commits I see here): that's a result of making #1199 dependent on this one. So if we just close this one, will all the commits in #1199 survive or will it blow up?

The other question is the decommissioning of the activate scripts (with the exception of create_crontab): neither PR currently takes care of that. I was assuming we would do it in a subsequent PR, once the ansible scripts are deemed good enough to replace the internal RPM.

server/bin/pbench-server-activate-create-results-dir-structure
portante: > So maybe you are looking at the "common" commits?

I see now that the 3 commits in this PR are the exact same commits in PR #1199.  I think what through me was:

> Besides, this script will be going away after PR #1199 is merged.

I read that to mean that PR #1199 will remove these, but it doesn't.

I'll drop this asking for this change, but really hope we won't get bit by the mixing nomenclature since this is going to live for longer than expected.

server/bin/pbench-server-activate-create-results-dir-structure
portante: Yes, my comment above was a result of not noticing that PR #1199 contains these exact same commits.

portante:
> So if we just close this one, will all the commits in #1199 survive or will it blow up?

All the commits in #1199 will survive as they are copied.  If we merge this PR, when you rebase #1199 on `master` the three commits that are the same in this PR and PR #1199 will disappear from #1199.

> I was assuming we would do it in a subsequent PR, once the ansible scripts are deemed good enough to replace the internal RPM.

This is what concerns me.  If we were to merge this PR and PR #1199, we'll have two ways of doing things.

Can we consider making this PR only about the agent side config changes, and making PR #1199 be about the server side config changes along with the _removal_ of the activate scripts in favor of the ansible playbooks?

ndokos:
I can certainly just do the agent changes here and put the server changes in #1199. Not getting rid of the activate scripts immediately however was a feature: having a second way to install the server if the ansible scrips crashed and burned badly enough, actually gives me peace of mind (even though it's giving you heartburn :-) ) Even though the ansible scripts have been tested on RHEL7 and the Fedoras, there are some issues (mostly undocumented dependencies: Avinash was testing it and ran iinto a bunch of them - he will open (or maybe already has - I haven't checked) an issue on that); I wanted the backup method in case something worse happens.

ndokos:
THis PR is mistitled: there are NO changes to the agent side at all. So it can all be incorporated into #1199 as is.

PR #1201, opened 2019-05-14
Sysstat fixes and CDM

atheurer:
- speed-up of sar and mpstat postprocess
- generate CDM for sar, pidstat, mpstat and iostat
- fixes for missing sar data
- fixes for log_cdm_metric_sample/gen_cdm_metric_data

PR #1204, opened 2019-05-17
Refactor agent to use envi for perl module paths

portante:
We had too much logic to support the unit tests in all the various
perl modules making it complicated to support. We strip all that
out, replacing those mechanisms with just environment variables
provided by the "base" script.

This simplifies much of the code and gives us a way to add other
perl module paths in the future.

We define the new `pbench_lib_dir` environment variable that the
CDM spaces uses, adding it to the "base" script.

We also fixed up the logging mechanisms to emit all the arguments
given on the call, this caused a change to the pbench-dbench test-02
output.

We also fix up the pbench-agent.cfg.example and pbench-agent-default.cfg
files so that all of the defaults exist in the default config file. We
then modified all references to the "pbench_install_dir" in the config
file to be "install-dir" to be clear we weren't try to reference th
environment variable.


portante:
This commit is a foundational fix to enable PR #1201 to work properly in all environments unit tests will run.

portante:
Almost done, but not quite.

My development system has an `/opt/pbench-agent` directory, so that allows the unit tests to pass.  I have to fix the unit tests so that they use their own.  The `bench-scripts` and `util-scripts` tests need to be fixed.

portante:
Okay, with a bit more work, we have verified that the unit tests run properly in a fresh environment.

atheurer:
OK, once this PR is merged, I take it I'll need to adjust my commits to avoid a merge conflict?

atheurer:
I am seeing some errors when running benchmarks:

# pbench-uperf --test-types=stream --instances=1 --samples=1 --protocols=tcp
  File "/usr/bin/getconf.py", line 65, in <module>
    status = main(conf, args, opts)
  File "/usr/bin/getconf.py", line 31, in main
    configtools.print_list(configtools.get_list(conf.get(sec, option)), sep)
  File "/usr/lib64/python2.7/ConfigParser.py", line 623, in get
    return self._interpolate(section, option, value, d)
  File "/usr/lib64/python2.7/ConfigParser.py", line 691, in _interpolate
    self._interpolate_some(option, L, rawval, section, vars, 1)
  File "/usr/lib64/python2.7/ConfigParser.py", line 723, in _interpolate_some
    option, section, rest, var)
ConfigParser.InterpolationMissingOptionError: Bad value substitution:
	section: [pbench-agent]
	option : install-dir
	key    : pbench_install_dir
	rawval :

Warning: Permanently added '127.0.0.1' (ECDSA) to the list of known hosts.
  File "/usr/bin/getconf.py", line 65, in <module>
    status = main(conf, args, opts)
  File "/usr/bin/getconf.py", line 31, in main
    configtools.print_list(configtools.get_list(conf.get(sec, option)), sep)
  File "/usr/lib64/python2.7/ConfigParser.py", line 623, in get
    return self._interpolate(section, option, value, d)
  File "/usr/lib64/python2.7/ConfigParser.py", line 691, in _interpolate
    self._interpolate_some(option, L, rawval, section, vars, 1)
  File "/usr/lib64/python2.7/ConfigParser.py", line 723, in _interpolate_some
    option, section, rest, var)
ConfigParser.InterpolationMissingOptionError: Bad value substitution:
	section: [pbench-agent]
	option : install-dir
	key    : pbench_install_dir
	rawval :

atheurer:
Also, I am not sure we get these env vars, like $pbench_lib_dir unless every pbench script source the pbench base file.  Not sure how we will do that in the perl scripts.

portante:
@atheurer, so first we need to make sure you have the `base` script installed from your tree, and have updated the `pbench-agent.cfg` and `pbench-agent-default.cfg` to be what is in master, keeping relevant changes specific to your environment.

Then, we'll have to wrap `pbench-run-benchmark` with a bash script that sources the `base` script, and then invoke the original `pbench-run-benchmark` (renamed `pbench-run-benchmark.pl`?).

atheurer:
@portante there are more than one perl script, so adding bash wrappers is going to add potentially many files.  Can we just update pbench's profile.sh to include the essential vars we need from base?

atheurer:
...or just source the base script in pbench's profile.sh script

portante:
@atheurer, what are the other perl scripts that are directly invoked by the user?  Are there that many?

portante:
Here is a simple wrapper that "should" work: https://github.com/distributed-system-analysis/pbench/pull/1206

atheurer:
There is also pbench-gen-iterations and pbench-run-benchmark-sample.  Either of these are runnable by the user.  There is basically any script in our pbench path which is not bash which will end up having this problem.  If we can just get the essential env vars in the pbench profile (or simply source base in that profile), then none of this wrapper stuff needs to happen.

atheurer:
This script does not work because it does not include the subdir:
<pre>
/opt/pbench-agent/bench-scripts/pbench-run-benchmark: line 7: /opt/pbench-agent/pbench-run-benchmark.pl: No such file or directory</pre>

I would much prefer we just source the base script with backticks in the beginning of the perl script, or rely on the env var being defined.  We can have both really: if the var is undefined, then the script soucres the base immediately followed by echoing the var to stdout, which is captured by perl and assigned to the var.  A subroutine like:

<pre>
sub get_pbench_var()
    my $var_name = $1
    return env{$var_name} if defined env{$var_name};
    value=`$pbench_install_dir/base; echo \$$var_name`;
    return $value;
} </pre>

atheurer:
@portante Actually, since we have a perl PbenchBase.pm, we should just be defining all of the same things that the bash base script does, but in perl in that library.  Any other language we use or will use should have the same procedure.  For our perl scripts, we only need to know how to load PbenchBase.pm,  which is typically <pre>dirname($0) . "/../lib"</pre>.  Once we have that, we have subroutines we can call to get paths like pbench_run_dir and pbench_inst_dir.

portante:
@atheurer, since this is a closed/merged PR, can we move this discussion to PR #1206?

PR #1206, opened 2019-05-20
Wrapper for pbench-run-benchmark to get ENV

portante:


portante:
Moving the discussion about how to enable perl-based command line tools from #1204 to here:

 * from https://github.com/distributed-system-analysis/pbench/pull/1204#issuecomment-494120053

> There is basically any script in our pbench path which is not bash which will end up having this problem.

Is it really the case that we want to enable every script to be runnable from the command line directly?  Aren't most of the scripts in the new approach to `pbench-run-benchmark` called from that script directly?  There might be only one or two that we could consider allowing the user's to call directly, right?  Do we really want to repeat the support problems we have with `pbench-start`/`stop-tools` with the new bench-scripts side?

> If we can just get the essential env vars in the pbench profile (or simply source `base` in that profile), then none of this wrapper stuff needs to happen.

While this is one approach, adding more reliance on a profile being set correctly in the root account makes it more difficult to maintain the `pbench-agent`.  Won't we have an easier time with less requirements on the user's profile being setup correctly?  Most applications know where to find their default configuration in `/etc/sysconfig`, and the like.  Seems like using the profile is not the way to go.

 * https://github.com/distributed-system-analysis/pbench/pull/1204#issuecomment-494124590

I have an update to this PR that should fix this problem, will post shortly.

* https://github.com/distributed-system-analysis/pbench/pull/1204#issuecomment-494153295

Working to adhere to the DRY principle is a huge boon to maintainability.  Having perl and bash code that do the same thing repeating the same logic will add more work for us as maintainers to keep working properly over time.  Since the `agent/base` script defines environment variables after processing the configuration file, we can keep the logic there in one place, rather than duplicating it.

The update to this PR should address this concern as well.

atheurer:
> Moving the discussion about how to enable perl-based command line tools from #1204 to here:
> 
> * from [#1204 (comment)](https://github.com/distributed-system-analysis/pbench/pull/1204#issuecomment-494120053)
> 
> > There is basically any script in our pbench path which is not bash which will end up having this problem.
> 
> Is it really the case that we want to enable every script to be runnable from the command line directly? 

Having the ability to run a specific program to get only the particular function you need is a good idea, and emulates the philosophy of UNIX.

> Aren't most of the scripts in the new approach to `pbench-run-benchmark` called from that script directly? There might be only one or two that we could consider allowing the user's to call directly, right? Do we really want to repeat the support problems we have with `pbench-start`/`stop-tools` with the new bench-scripts side?

Yes, normally pbench-run-benchmark is the one to cal these scripts, but the ability to debug a specific function is accomplished by calling a script directly.  While pbench start/stop tools may be not familiar to users, I certainly would not want to do something to make calling them from the user not possible or requiring a wrapper script.

> 
> > If we can just get the essential env vars in the pbench profile (or simply source `base` in that profile), then none of this wrapper stuff needs to happen.
> 
> While this is one approach, adding more reliance on a profile being set correctly in the root account makes it more difficult to maintain the `pbench-agent`. Won't we have an easier time with less requirements on the user's profile being setup correctly? Most applications know where to find their default configuration in `/etc/sysconfig`, and the like. Seems like using the profile is not the way to go.

Having a working profile seems like an all or nothing situation, and adding the essential vars (2 of them?) seems like a trivial thing to do.  However, if maintaining this profile is such a burden, then let's get rid of this profile, and put all of our executable programs somewhere in the default path.

I think the real problem is just how complicated agent/base has become to set up just a few vars.  We now have layers on layers just to get some vars.  I see we still have configtools to get some vars, but configtools sometimes has some reliance on env vars, which were set from agent/base, but that also calls cofigtools to define this env var (pbench_install_dir).    @ndokos I'd like your opinion on the situation.  Personally, I would be perfectly fine with a /etc/sysconfig/pbench-agent, as long as it contains  *only* simple var assignments, like "pbench_run=/var/lib/pbench-agent" -that can be interpreted by any language without the need for wrapper scripts.  Agent/base can then be used only for defining shared functions for other pbench bash scripts.

> 
> * [#1204 (comment)](https://github.com/distributed-system-analysis/pbench/pull/1204#issuecomment-494124590)
> 
> I have an update to this PR that should fix this problem, will post shortly.
> 
> * [#1204 (comment)](https://github.com/distributed-system-analysis/pbench/pull/1204#issuecomment-494153295)
> 
> Working to adhere to the DRY principle is a huge boon to maintainability. Having perl and bash code that do the same thing repeating the same logic will add more work for us as maintainers to keep working properly over time. Since the `agent/base` script defines environment variables after processing the configuration file, we can keep the logic there in one place, rather than duplicating it.
> 
> The update to this PR should address this concern as well.

portante:
> Having the ability to run a specific program to get only the particular function you need is a good idea, and emulates the philosophy of UNIX.

Not sure how exposing the internals of pbench is emulating the philosophy of UNIX. Isn't that like saying I have to expose all the component parts of `ps`, or `ls`?

We all seem to agree that the pbench-agent needs to done it's job well, but does it need to be so flexible all of its internals are runnable by all users and supported?  Why not say that `pbench-run-benchmark` is the interface, and if you need to go deeper contribute to the code to do what is needed?

> Yes, normally pbench-run-benchmark is the one to cal these scripts, but the ability to debug a specific function is accomplished by calling a script directly.

If we are talking about developers, then can't somebody who is developing pbench can run `base` directly and debug it that way?  Why do we need to expose the internals of pbench to everybody using it?

> Having a working profile seems like an all or nothing situation, and adding the essential vars (2 of them?) seems like a trivial thing to do.

The profile is working as it is to define the `CONFIG` variable.  Adding two more that are copies of data in the agent configuration file does not follow the DRY principle, and will likely cause problems down the road.

> However, if maintaining this profile is such a burden, then let's get rid of this profile, and put all of our executable programs somewhere in the default path.

I agree, removing the profile entirely would be great.  We don't use a profile on the server side, and it works well.

But doing that right now is not in the scope of this work and will cause too much churn for getting this all working.

> I think the real problem is just how complicated agent/base has become to set up just a few vars.

Moving to a model using something like `/etc/sysconfig/pbench-agent` would be fine to do.  Let's consider that as future work so that we don't create too much change on top of what we already have here.

This change keeps things in one place, and provides the environment for `pbench-run-benchmark`.  Developers can know to source `base` when they want to work, or we can just create simple symlinks to `pbench-trampoline` for those scripts that want `base` always sourced.

atheurer:
> > Having the ability to run a specific program to get only the particular function you need is a good idea, and emulates the philosophy of UNIX.
> 
> Not sure how exposing the internals of pbench is emulating the philosophy of UNIX. Isn't that like saying I have to expose all the component parts of `ps`, or `ls`?

What I am saying is that one pbench program is like running "ps" and  another is like running "ls", and the user may want to run these on their own, or we combine these components in a script like pbench-run-benchmark.
> 
> We all seem to agree that the pbench-agent needs to done it's job well, but does it need to be so flexible all of its internals are runnable by all users and supported? Why not say that `pbench-run-benchmark` is the interface, and if you need to go deeper contribute to the code to do what is needed?
> 
> > Yes, normally pbench-run-benchmark is the one to cal these scripts, but the ability to debug a specific function is accomplished by calling a script directly.
> 
> If we are talking about developers, then can't somebody who is developing pbench can run `base` directly and debug it that way? Why do we need to expose the internals of pbench to everybody using it?

At this point given the circumstances and the difference of opinion we have, I'd rather just require the user run base themselves before running any non-bash script, as I find the idea of a wrapper script utterly ridiculous.

If we can get to a point where there is no /etc/profile.d/pbench.sh , no configtools, and a /etc/sysconfig/pbench-agent, and therefore no reliance on a base for getting basic pbench vars, then the problem goes away entirely.

PR #1209, opened 2019-05-23
Reference pbench_install_dir enviroment variable

portante:
We reference the `pbench_install_dir` environment variable instead of
fetching it from the config file directly since `base` provides it.

We also fix a few tabs vs spaces issues, and add the "item" which was
in error when looping through the list of sysinfo items to execute.


portante:
Final fix related to #930.

PR #1211, opened 2019-05-23
Wrap jschart with React component for handling reloading/resetting

gurbirkalsi:
`TimeseriesGraph` defines jschart as a React component and handles data
visualization according to the React lifecycle of the component.

Initial rendering occurs when the component mounts and reloading occurs
when new data, series names, or axis series data is passed to the
component.

### DIFF
--- a/web-server/v0.4/package.json
+++ b/web-server/v0.4/package.json
@@ -31,7 +31,7 @@
     "dva-core": "^1.1.0",
     "enquire-js": "^0.2.1",
     "html2canvas": "^1.0.0-alpha.12",
-    "jschart": "^1.0.5",
+    "jschart": "^1.0.7",
     "jspdf": "^1.5.3",
     "lodash": "^4.17.10",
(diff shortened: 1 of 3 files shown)


k-rister:
On a related topic, are the current dynamic features sufficient for what is needed in the dashboard?  What I have implemented so far allows the data series to be reloaded in a chart, but properties of the chart itself (chart title, axis mode, axis titles, stacked vs. line, etc.) are not dynamic -- is that something that is needed?

gurbirkalsi:
@k-rister These recent changes are great for allowing reloading of data series in charts in the dashboard; however, we should strive to allow reloading and configurability of all chart properties. This would be especially useful in the run comparison page where we are constantly using the same chart ID to display multiple cluster timeseries graphs.

k-rister:
@gurbirkalsi Ok, that is consistent with what I was thinking.  I just wanted to make sure we were on the same page before I spent more time on it.  Thanks!

PR #1207, opened 2019-05-21
Sysstat pr3

atheurer:
replaces https://github.com/distributed-system-analysis/pbench/pull/1201

PR #1212, opened 2019-05-24
Add collection of online CPU status for mpstat

portante:
And only use the `online-cpus.txt` if present, since stop-tools
now collects it.

We add a unit test which properly excludes CPUs based on the
presence of the `online-cpus.txt`.

We also include some commit updates based review of PR #1207.


atheurer:
Anyway, looks good to me, and I back when I tested this new caching of online cpu info, I saw a 16x difference in mpstat postprocessing performance.

portante:
> we now no longer allow for checks on the local system for an online cpu at post-process script execution?

We have just moved it to `pbench-stop-tools` via the `sar` tools file.  That captures the online CPU state in a file the mpstat always looks for.

> we just assume a CPU is online if the online-cpus file is not present.

Yes

> Hopefully this is not a problem when re-post-processing older data [which does not include the online-cpus file]

So if we reprocess pre-PR-#1212-collected `mpstat` data with this change we won't have the online-cpus.txt file collected, so this code will assume the CPU is online.

We could actually detect that, because the column totals will be less than 100%.

> I am hoping mpstat does not (and has not for a long time) had that behavior

I think it still does.  So that is why `pbench-stop-tools` is now collecting it all the time.  It won't help if a CPU goes offline after `pbench-start-tools` is called and before `pbench-stop-tools` is called, or if it goes offline and back online during a start/stop tools pair.  To handle this we'd have to sum all the columns for a given CPU to see if it is substantially less than 100%.

The new change should not be a different behavior than what we have today because we are still consulting the online CPU states at the end of `pbench-stop-tools`.

PR #1220, opened 2019-05-29
Restore creation of pbench_run directory

### LINKED ISSUE #1219: The "pbench_run" directory is not created automatically
On a freshly installed system, if one accepts all the defaults, and just copies the `pbench-agent.cfg.example` file to `pbench-agent.cfg`, setting the proper host names, then the `/var/lib/pbench-agent` directory is not created for the user.

portante:
A recent commit, 7d57f5e2 (PR #1204), introduced an error check for the `pbench_run` directory's existence when the automatic creation of that directory is expected.

Restore the automatic directory creation.

Fixes #1219.

### DIFF
--- a/agent/base
+++ b/agent/base
@@ -14,17 +14,12 @@ if [[ -z "$pbench_run" ]]; then
     if [[ -z "$pbench_run" ]]; then
 	pbench_run=/var/lib/pbench-agent
     fi
+    export pbench_run
+else
     if [[ ! -d $pbench_run ]]; then
-        echo "[ERROR] pbench run directory, ${pbench_run}, does not exist" >&2
+        echo "[ERROR] the provided pbench run directory, ${pbench_run}, does not exist" >&2
         exit 1
     fi
-    export pbench_run
-fi
-
-# if $pbench_run does not exist, but we had an old /var/lib/pbench directory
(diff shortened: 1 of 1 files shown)


agent/base
```diff
@@ -14,17 +14,12 @@ if [[ -z "$pbench_run" ]]; then
     if [[ -z "$pbench_run" ]]; then
 	pbench_run=/var/lib/pbench-agent
     fi
+    export pbench_run
+else
     if [[ ! -d $pbench_run ]]; then
-        echo "[ERROR] pbench run directory, ${pbench_run}, does not exist" >&2
+        echo "[ERROR] the provided pbench run directory, ${pbench_run}, does not exist" >&2
```
ndokos: That does not create the directory: it exits with an error if it does not exist. So that contradicts what the commit message says, right?

portante: No, but since this is now in an else clause, we only consider the missing `$pbench_run` directory as an error if a caller of `base` set the environment variable to a directory that does not exist.

The only caller that does that is `unittests`.

Now, if `$pbench_run` is not provided as an environment variable, then we ensure by line 17 above that `$pbench_run` has a value, and line 28 below will create it.

PR #1222, opened 2019-05-30
pbench-run-benchmark monitor

k-rister:
When pbench-run-benchmark launches a process via PbenchAnsible.pm the output of that process is buffered within Ansible until the process exits.  In order to be able to monitor a benchmark while it is running these patches introduce a monitor script that is used as a wrapper around the benchmark process.  This wrapper takes special care to preserve the distinctly separate STDERR and STDOUT data streams while allowing the user to attach to a screen session on a/the client (named pbench-output-monitor) to view the output in real time with identical separation of output streams.

In the monitor, each benchmark process has it's STDERR and STDOUT redirected to named pipes via tee.  Using tee ensures that the output is preserved in the Ansible workflow as is currently done.  A bash process is laucned inside of the screen session which simply uses cat to extract the output from the named pipes while preserving the STDERR/STDOUT relationship.

Since it is possible that a given client may be executing multiple copies of a benchmark, special care is taken to ensure that the screen session which handles the output monitor(s) is launched properly.  A flock call is used to ensure that only the first invocation of the monitor creates the screen session; any subsequent call to the output monitor fails to obtain the flock so it attaches to the existing screen session rather than creating a new one.  This allows all invocations that are being monitored to be contained within a single screen session for easier navigation and monitoring.  There is potential for a race between the first invocation and any subsequent invocation of the screen session so a small polling loop is introduced (with 0.1 second granularity to minimize polling overhead) to ensure that the screen session's socket file exists prior to attempting to attach additional sessions.

### DIFF
--- a/agent/bench-scripts/pbench-run-benchmark-sample
+++ b/agent/bench-scripts/pbench-run-benchmark-sample
@@ -17,7 +17,8 @@ use JSON;
 use File::Temp;
 use Data::Dumper;
 use PbenchBase       qw(get_json_file put_json_file get_benchmark_names get_clients
-                        get_params get_pbench_bench_config_dir load_benchmark_config);
+                        get_params get_pbench_bench_config_dir load_benchmark_config
+                        get_pbench_install_dir );
 use PbenchAnsible    qw(ssh_hosts ping_hosts copy_files_to_hosts copy_files_from_hosts
                         remove_files_from_hosts remove_dir_from_hosts create_dir_hosts
                         sync_dir_from_hosts verify_success yum_install_hosts);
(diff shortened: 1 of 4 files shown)


portante:
@k-rister, @atheurer, is this something we want to have reviewed and merged?  Or is this just a work-in-progress item?

atheurer:
I will review today

k-rister:
I believe it is good to go, I've been using it and it works for me.

portante:
FWIW: this is still "tech-preview".  We have no unit tests for this functionality like we have for other agent side commands.  To graduate this to fully supported will require integration into TavisCI unit tests.

PR #1224, opened 2019-05-31
Refactor pbench-fio to enable easier unit testing, fix bugs, and clean up a bit for easier maintenance

portante:
While reviewing a few other PRs (#1150 and #1216) it became clear that unit tests where difficult to write for `pbench-fio`.  In addition to unit test updates, and related changes, we noticed some bugs handling multiple `--targets`.

The two of the commits merge in the work from #1150 and #1216 as well (each squashed to one commit for those two PRs).


portante:
Rebased and updated to adding missing commit that I had not completed in my local workspace.

portante:
Still need to add some more unit tests to cover the changes made.

portante:
@ndokos, I'll merge when you are ready, and then we can perform some validation tests.  Perhaps @bengland2 can help to verify as well?

agent/bench-scripts/pbench-fio
```diff
@@ -489,29 +489,34 @@ function fio_device_check() {
 
 function fio_create_jobfile() {
 	local fio_job_file="${13}"
```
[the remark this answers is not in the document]
portante: This is historical, and felt that it was not worth changing, but I did add a comment to help number the arguments to make it easier to verify.

agent/bench-scripts/pbench-fio.md
```diff
@@ -89,6 +89,22 @@ and then randomly read/write only an insignificant fraction of the entire set of
 yet still get representative throughput and latency numbers.  
 The same is true for using fio for random I/O on block devices.  
 
+## shared filesystems
+
+A slightly different set of parameters is needed for pbench-fio to work with shared filesystems (where same set of files
+is shared by multiple clients/mountpoints).  For this to work, you must do 2 things:
+
+* specify --target=my-directory 
+* specify --jobfile=/opt/pbench-agent/bench-scripts/templates/fio-shared-fs.job
```
atheurer: @portante this needs to be --job-file to match the options processing

atheurer: Also, if --file-size or -s is not used, the fio job will not work [when using the fio-shared-fs.job], so we should probably add a third requirements here:

specify --file-size=size-in-MB

portante: Fixed by #1256.

k-rister: Should we elaborate on why this is?  Is the idea that the user must specify a file size that is <= to the available space divided by the number of clients?

portante: Well, now that I looked at how the fio job file is being generated, the `fs-shared-fs.job` does not contain a default `size` value.  I'll post a PR that we could consider.  We might not need to add this file size if we provide a default.

atheurer: Yes, adding a default size would be good, if we already do this for the regular job file

portante: @k-rister, let's review #1257 first, and then see what we should have for documentation.

PR #1230, opened 2019-06-07
pbench-move-results: Quote variables in conditional expressions

### LINKED ISSUE #1229: weird "host unreachable" message from pbench-move-results
When I tried to move results to the pbench server, I got a error message that did not make sense, might want to fix this to just print out "host unreachable" instead of raw partial HTML page.
```
[root@e24-h09-740xd dmcrypt]# /opt/pbench-agent/util-scripts/pbench-move-results --user bengland@redhat.com --prefix "pre-test-check"
/opt/pbench-agent/util-scripts/pbench-move-results: line 118: [: too many arguments
[error][2019-06-07T15:46:28.820583734] ERROR: results host unreachable: <!DOCTYPE html>
<html>
<head>
<title>Pbench Server down for service</title>
<style>
    body {
        width
```
 This is on RHEL8 GA
```
[root@e24-h09-740xd dmcrypt]# subscription-manager repos --list-enabled
+----------------------------------------------------------+
    Available Repositories in /etc/yum.repos.d/redhat.repo
+----------------------------------------------------------+
Repo ID:   rhel-8-for-x86_64-baseos-rpms
Repo Name: Red Hat Enterprise Linux 8 for x86_64 - BaseOS (RPMs)
Repo URL:  https://cdn.redhat.com/content/dist/rhel8/$releasever/x86_64/baseos/os
Enabled:   1
```
 with pbench versions here:
```
[root@e24-h09-740xd dmcrypt]# rpm -qa | grep pbench
pbench-agent-0.60-1gc9e89f8f.noarch
pbench-sysstat-12.0.3-1.x86_64
pbench-agent-internal-0.60-1.noarch
pbench-fio-3.12-1.x86_64
```

ndokos:
Fixes #1229

### DIFF
--- a/agent/util-scripts/pbench-move-results
+++ b/agent/util-scripts/pbench-move-results
@@ -17,7 +17,7 @@ function usage() {
 
 # Process options and arguments
 opts=$(getopt -q -o u:p:xSh --longoptions "user:,prefix:,xz-single-threaded,show-server,help" -n "getopt.sh" -- "$@");
-if [ $? -ne 0 ]; then
+if [[ $? -ne 0 ]]; then
     printf "\n"
     printf "%s\n" $*
     printf "$script_name: you specified an invalid option\n\n"
@@ -35,14 +35,14 @@ while true; do
     case "$1" in
 	-u|--user)
 	    shift;
-            if [ -n "$1" ]; then
+            if [[ -n "$1" ]]; then
                 user="$1"
                 shift;
             fi
             ;;
         -p|--prefix)
             shift;
-            if [ -n "$1" ]; then
+            if [[ -n "$1" ]]; then
                 prefix="$1"
                 shift;
             fi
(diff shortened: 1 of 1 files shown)


ndokos:
Going to [[ wholesale is not correct: in particular when == and != are used, that's a pattern matching expression, so we are trading one kind of danger for another in those cases. I can change them in other cases, but I'm not sure what it'll gain us: we cannot make it an automatic rule - we'll have to think about each case.

ndokos:
We can standardize on [[ for expessions of the form `[[ -f $foo ]]` where the variable can be quoted optionally (the quoting is mandatory for [). Is that acceptable?

portante:
> We can standardize on [[ for expessions of the form [[ -f $foo ]] where the variable can be quoted optionally (the quoting is mandatory for [). Is that acceptable?

Yes, that seems like a good start.

We just need to be clear what to use when and be sure why.  The benefits of being consistent will help maintainability.

ndokos:
I actually changed all of them to [[...]] tests, but verified that RHS operands are quoted, so the potential pattern-matching tests do not cause spurious matches.

PR #1231, opened 2019-06-07
Fix uperf runtime padding

### LINKED ISSUE #1227: pbench-uperf default runtime is not padded by 6 seconds as happens when --runtime is specified
If you invoke `pbench-uperf` using `--runtime=60` and compare that run to one invoked without using `--runtime`, the first will actually run `uperf` for 66 seconds, while the second will only run for 60 seconds.

portante:
Fixes an inconsistency between using:

 * `pbench-uperf`
 * `pbench-uperf --runtime=60`

The first will run a test for 60 seconds, the second will add 6 seconds of padding and run it for 66 seconds.

Fixes #1227.

### DIFF
--- a/agent/bench-scripts/pbench-uperf
+++ b/agent/bench-scripts/pbench-uperf
@@ -294,8 +294,7 @@ while true; do
 		-r|--runtime)
 		shift
 		if [ -n "$1" ]; then
-			# runtime is padded by 6 seconds to allow 3 seconds ramp up and ramp down
-			runtime=`echo "$1 + 6" | bc`
+			runtime="$1"
 			shift
 		fi
 		;;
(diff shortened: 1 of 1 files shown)


agent/bench-scripts/pbench-uperf
```diff
@@ -425,6 +424,9 @@ if [ "$postprocess_only" != "y" ]; then
 	fi
 fi
 
+# runtime is padded by 6 seconds to allow 3 seconds ramp up and ramp down
+runtime_padded=`echo "${runtime} + 6" | bc`
```
ndokos: runtime_padded=$((${runtime}+6))

perhaps?

portante:
@atheurer, can you take a look at this commit and make sure you are on board with it?

portante:
@atheurer, should we continue to wait for you?  Or should we merge and review a different way?

PR #1236, opened 2019-06-10
Integrate Explore page for discovering and launching shared dashboard sessions

gurbirkalsi:
Users may launch collected dashboard sessions from the shared
session table which includes details regarding session IDs, created
times, and description.

This page adds a new datastoreConfig field for the GraphQL server.

### DIFF
--- a/web-server/v0.4/README.md
+++ b/web-server/v0.4/README.md
@@ -76,6 +76,7 @@ export default {
   '/dev/datastoreConfig': {
       "elasticsearch": "http://elasticsearch.example.com",
       "results": "http://results.example.com",
+      "graphql": "http://graphql.example.com",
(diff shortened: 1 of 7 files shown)


web-server/v0.4/README.md
```diff
@@ -76,6 +76,7 @@ export default {
   '/dev/datastoreConfig': {
       "elasticsearch": "http://elasticsearch.example.com",
       "results": "http://results.example.com",
+      "server": "http://server.example.com",
```
portante: Did you consider naming this `graphql`, or something like that, instead of using the generic name of `server`?

gurbirkalsi: Yes, `server` is too generic here - I've replaced this with `graphql`.

PR #1239, opened 2019-06-12
Filter iteration hostname & ports by `all` keyword on component update

gurbirkalsi:
`TableFilterSelection` listens for port changes and finds the aggregate hostname and port selection for default display in the comparison select page.

### DIFF
--- a/web-server/v0.4/src/utils/parse.js
+++ b/web-server/v0.4/src/utils/parse.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-param-reassign */
 import _ from 'lodash';
 
 export const parseIterationData = results => {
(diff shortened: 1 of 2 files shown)


web-server/v0.4/src/utils/parse.js
```diff
@@ -77,9 +78,17 @@ export const parseIterationData = results => {
...
+              ? `client_hostname:${hostMetadata.client_hostname}`
+              : ``;
+            const serverHostname = hostMetadata.server_hostname
+              ? `server_hostname:${hostMetadata.server_hostname}`
+              : ``;
+            const serverPort = hostMetadata.server_port
+              ? `server_port:${hostMetadata.server_port}`
+              : ``;
+
+            const columnHost = `${clientHostname}${serverHostname}${serverPort}`;
```
portante: Why does it work to mash up these names without a separator?

web-server/v0.4/src/utils/parse.js
```diff
@@ -77,9 +78,17 @@ export const parseIterationData = results => {
...
-            }-server_port:${hostMetadata.server_port}`;
+            const clientHostname = hostMetadata.client_hostname
+              ? `client_hostname:${hostMetadata.client_hostname}`
+              : ``;
+            const serverHostname = hostMetadata.server_hostname
+              ? `server_hostname:${hostMetadata.server_hostname}`
+              : ``;
+            const serverPort = hostMetadata.server_port
+              ? `server_port:${hostMetadata.server_port}`
+              : ``;
```
portante: If all of the above resolve to empty strings, how should this function behave?

web-server/v0.4/src/utils/parse.js
gurbirkalsi: This works because `columnHost` is also assigned as the dataIndex for the table. However, we should include separators here for better readability.

web-server/v0.4/src/utils/parse.js
gurbirkalsi: An empty column cell would be rendered in this case. We make an assumption that each iteration has a hostname or port assigned to it but in the case there isn't, we make use of the `columnPrefix` as the dataIndex used for that column render.

PR #1199, opened 2019-05-11
Server installation ansible playbooks and roles

ndokos:
These playbooks implement a complete install and setup
of one or more pbench servers.

In addition to installing the pbench-server RPM and the config file
for each server, they do all the work that the pbench-server-activate-*
scripts do.

PR #1242, opened 2019-06-14
Rework pbench-agent ansible playbooks

ndokos:
Rework the ansible playbooks to make them as similar as possible
to the server ansible playbooks (PR #1199).

There are three playbooks: the main one is `pbench-agent-install.yml`
which installs the COPR repos, uses them to install the `pbench-agent`
RPM from COPR and then copies the config file and key file for the
environment requested through the `cenv` command line variable (if
this is not set, the default is "production").

Usage:

     ansible-playbook --user=root -i /path/to/inventory.file \
                      pbench-agent-install.yml \
                      [ --extra-vars='{cenv: <environment>}' ]

Two other playbooks use subsets of the defined roles:
`pbench-repo-install.yml` installs the COPR repos only. You can then
install the `pbench-agent` RPM manually.  In that case, the third
playbook can be used to install the config and key files afterwards.

An example inventory file, `Inventory/pbench-agentf.hosts.example`, is
provided. The second section of the file contains the definition of
various variables that are used by the playbooks.  These are set to
dummy values here, so a site admin should modify this file for a
particular installation and make it available to the users for that
site.

Users should then copy and modify the example file, without touching
the variables section: they should just add their hosts to the first
section (possibly augmented with host variable definitions,
e.g. ansible_python_interpreter, that might be need to work around
problems on some distros).


portante:
When you rebase to pick up #1246 to fix the TravisCI failures, can you also squash these two commits together?

agent/ansible/Inventory/pbench-agent.hosts.example
```diff
@@ -0,0 +1,17 @@
+# Users should copy the site-specific file and modify *only* the
+# first section by adding their hosts. They may also need to
+# provide host-specific variables to deal with problems (e.g ATM,
+# a F30 host needs to have the `ansible_python_interpreter' variable
+# to /usr/bin/python3 in order for things to work - see host3 below
```
portante: s/to/set to/

agent/ansible/roles/pbench-repo-install/vars/main.yml
```diff
@@ -0,0 +1,44 @@
+---
+baseurl: "{{ 'http://coprbe.devel.redhat.com/results/ndokos/pbench/rhel-8.dev-$basearch/'
```
portante: Can we remove the internal URL for RHEL 8 from all this?

agent/ansible/roles/pbench-repo-install/vars/main.yml
```diff
@@ -0,0 +1,44 @@
...
+                 if ansible_distribution == 'RedHat' and ansible_distribution_major_version == '8'
+                 else
+             'https://copr-be.cloud.fedoraproject.org/results/ndokos/pbench/epel-7-$basearch/'
```
portante: Can we factor out the `https://copr-be.cloud.fedoraproject.org/results/ndokos` as a prefix in a variable?

agent/ansible/roles/pbench-repo-install/templates/etc/yum.repos.d/pbench-test.repo.j2
```diff
@@ -0,0 +1,11 @@
+###########################################################################
+# External COPR repo
```
portante: Can we remove the notion of having a test vs production set of repos and just have one play book that works using a variable that can be changed for test vs production?

PR #1243, opened 2019-06-18
Reference selectedIterations via global redux store namespace

gurbirkalsi:


gurbirkalsi:
@Anisha1234 Can you confirm if this PR resolves the issues we observed with sharing session urls that redirect to the run comparison page?

PR #1246, opened 2019-06-25
Resolve linting errors

gurbirkalsi:


portante:
The check was *not* pending, it had passed but was not updating.  So I merged this PR.

PR #1254, opened 2019-07-09
pbench-agent now relies on Python 3 instead of Python 2

portante:
The pbench-agent now relies on Python 3 instead of Python 2.

To make this happen, we had to ensure all of our Python 2 executables we executed in the unit test suite.  The first two commits provide those changes, while the last commit actually makes the change to Python 3.

We had to first add test infrastructure so that tool "`datalog`s" can be tested, and then enhance the `pbench-fio` test infrastructure to execute the code path where our `fio-postprocess-viz.py` was executed, and the resulting generated HTML captured.


portante:
Each separate commit should pass unit tests successfully.

agent/tool-scripts/datalog/prometheus-metrics-datalog
```diff
@@ -15,39 +16,64 @@ key = sys.argv[7]
 if (cert == "None") or (key == "None"):
     command = "prom2json http://%s:%d/metrics" % (host, port)
 else:
-    command = "prom2json %s --cert=%s --key=%s https://%s:%d/metrics" % (
-            "--accept-invalid-cert" if selfsigned == "true" else "", cert, key, host, port)
+    command = "prom2json%s --cert=%s --key=%s https://%s:%d/metrics" % (
+            " --accept-invalid-cert" if selfsigned == "true" else "", cert,
+            key, host, port)
+
+_sleep_iterations = 0
```
ndokos: Can we squirrel this section away in a function to indicate that we are mocking `sleep` and `time`? Or at least add a comment to that effect?

agent/tool-scripts/datalog/unittests
```diff
@@ -0,0 +1,90 @@
...
+    [prometheus-metrics]="example.host.com 3 tool_log 9042 cert false key"
+)
+
+function testit {
+    printf -- "Testing ${2}-datalog under setup ${1} ...\n"
+
+    local _execdir=${_testroot}/${1}
+    mkdir ${_execdir}
+
+    # Make sure we use PYTHONUNBUFFERED is set as an environment variable
```
ndokos: Delete `we use`?

agent/tool-scripts/datalog/unittests
```diff
@@ -0,0 +1,90 @@
...
+    # so that data is written immediately to output files.
+    _PBENCH_UNIT_TEST=3 _testlog=${_execdir}/test-execution.log PYTHONUNBUFFERED=True pbench_lib_dir=$(pwd)/../../lib ./${2}-datalog ${options[${2}]} > ${_execdir}/stdout 2> ${_execdir}/stderr < /dev/null
```
ndokos: Why `_PBENCH_UNIT_TEST=3`? Does the `3` have special meaning?

agent/tool-scripts/datalog/unittests
```diff
@@ -0,0 +1,90 @@
...
+    sts=${?}
+    if [ ${sts} -eq 0 ]; then
+        rm -rf ${_execdir}
+    fi
+    return ${sts}
+}
+
+res=0
+
+for i in $(ls -1 gold/) ; do
```
ndokos: Should we depend on what is in `gold`? Wouldn't it better in some sense to have a list of things specified here and complain if the gold file is missing? I think that's how we do it in all the other unit tests.

agent/tool-scripts/prometheus-metrics
```diff
@@ -191,7 +191,7 @@ case "$mode" in
 			debug_log "[$script_name] key is not defined for $host, assuming that the endpoint doesn't need authentication"
 			key="None"
 		fi
-		tool_cmd='GOPATH=$pbench_install_dir/.go PATH=$PATH:$GOPATH/bin $script_path/datalog/$tool-datalog $host $interval $tool_log $port $cert $selfsigned $key'
+		tool_cmd='PYTHONUNBUFFERED=True GOPATH=$pbench_install_dir/.go PATH=$PATH:$GOPATH/bin $script_path/datalog/$tool-datalog $host $interval $tool_log $port $cert $selfsigned $key'
```
ndokos: Maybe add the same comment as in `unittests` here to explain why `PYTHONUNBUFFERED` is needed?

agent/tool-scripts/datalog/prometheus-metrics-datalog
portante: I added various comments to describe what was happening.  Moving this to a function or a class seemed like syntactic sugar given it is only used in this one place.

agent/tool-scripts/datalog/unittests
portante: Fixed.

agent/tool-scripts/datalog/unittests
portante: Yes, added a comment describing its meaning.

agent/tool-scripts/datalog/unittests
portante: This infrastructure was copied from the `tool-scripts/postprocess` infrastructure where that used the `samples` directory instead.  Since this new datalog infrastructure does not have a `samples` directory, and since we always have "gold" files, it seemed reasonable to use that.

agent/tool-scripts/prometheus-metrics
portante: Added a comment, and updated the commit message to record the explicit fix.

agent/bench-scripts/test-bin/bm
```diff
@@ -4,3 +4,18 @@
...
+
+if [[ "${_tst_benchmark}" == "pbench-fio" ]]; then
+    if [[ "${1%%=*}" == "--client" ]]; then 
+        client_file=${1##*=}
+        for client in $(cat ${client_file}); do
+            if [[ "${client%%.*}" == "hist" ]]; then 
+                # When we are mocking out a benchmark command for pbench-fio,
+                # for each client which begins with "hist" create an empty
+                # fio latency histogram log file so that the invocation of
+                # out latency visualizations can be exercised.
```
ndokos: out --> our

agent/bench-scripts/postprocess/fio-postprocess-viz.py
```diff
@@ -1,12 +1,8 @@
...
-from __future__ import print_function
+#!/usr/bin/env python3
 
 import sys
 from os.path import join, basename
-try:
-    from configparser import SafeConfigParser
-except ImportError:
-    from ConfigParser import SafeConfigParser
+from configparser import SafeConfigParser
```
ndokos: Running the unit tests on F29, I get a warning from this line (in this and every other file that contains it, so many unit tests fail) - here's a typical example:
````
  +++ Running test-31 pbench-copy-results --prefix=foo/bar --user=ndk
+ /var/tmp/pbench-test-utils/opt/pbench-agent/util-scripts/pbench-add-metalog-option:20: DeprecationWarning: The SafeConfigParser class has been renamed to ConfigParser in Python 3.2. This alias will be removed in future versions. Use ConfigParser directly instead.
+   config = SafeConfigParser()
+ /var/tmp/pbench-test-utils/opt/pbench-agent/util-scripts/pbench-add-metalog-option:20: DeprecationWarning: The SafeConfigParser class has been renamed to ConfigParser in Python 3.2. This alias will be removed in future versions. Use ConfigParser directly instead.
+   config = SafeConfigParser()
  tar --create --force-local "pbench-user-benchmark_ndk-test-1_2018.05.23T03.21.32" | xz -T0 > "/var/tmp/pbench-test-utils/pbench/tmp/pbench-copy-results.NNNNN/testhost/pbench-user-benchmark_ndk-test-1_2018.05.23T03.21.32.tar.xz" 
  --- Finished test-31 pbench-copy-results (status=0)

````

File: agent/bench-scripts/gold/pbench-user-benchmark/test-12.txt [hunk 8]
```diff
@@ -4,15 +4,15 @@ ERROR:root:current position: 51
 Traceback (most recent call last):
   File "/var/tmp/pbench-test-bench/opt/pbench-agent/bench-scripts/postprocess/user-benchmark-wrapper", line XXX, in <module>
     data = json.load(data_file)
-  File "/usr/lib/python2.7/json/__init__.py", line XXX, in load
-    **kw)
-  File "/usr/lib/python2.7/json/__init__.py", line XXX, in loads
+  File "/usr/lib/python3.6/json/__init__.py", line XXX, in load
```
ndokos: Missed this before: this fails for me because I'm running 3.7.

portante: Fixed this by removing the exception entirely.  And did a better conversion to Python 3 while cleaning up the code in `user-benchmark-wrapper`.

PR #1256, opened 2019-07-11
Fix up pbench-fio documentation

portante:
Fixes comments posted in #1224.


portante:
Let's consider these documentation changes after we consider #1257.

portante:
This change is superseded by #1257.

PR #1260, opened 2019-07-15
Implement multi upload API for object size more than 5 GB in size.

tenstormavi:


server/bin/pbench-backup-tarballs.py
```diff
@@ -248,17 +252,53 @@ def backup_to_s3(s3_obj, logger, controller_path, controller, tb, tar, resultnam
...
-        # moved aside and we don't retry, but we save the links for eventual
-        # stashing into S3. Verification will complain about these.
-        logger.warning("Tar ball, {}, too large to upload", tar)
-        return Status.TOO_LARGE
-
     md5_base64_value = (base64.b64encode(
         bytes.fromhex(archive_md5_hex_value))).decode()
+    tb_size = s3_obj.connector.getsize(tar)
+    if tb_size > (5 * GB):
+        multi_upload_config = TransferConfig(
```
ndokos: This all should go into an `s3Connector` method.

server/bin/pbench-backup-tarballs.py
```diff
@@ -248,17 +252,53 @@ def backup_to_s3(s3_obj, logger, controller_path, controller, tb, tar, resultnam
...
+                                                    ExtraArgs = {'Metadata': { 'MD5': local_multipart_etag}})
+                except ClientError as e:
+                    logger.error(
+                        "Multi-upload to s3 failed, client error: {}".format(e)
+                    )
+                    return Status.FAIL
+                else:
+                    # compare the multi etag value uploaded in metadata field # with s3 etag for data integrity.
+                    try:
+                        obj = s3_obj.connector.get_object(Bucket='{}'.format(
```
ndokos: I hope `get_object` just gets  a header, not the whole thing (we don't want to transfer the whole tarball back). Is that right?

server/bin/pbench-backup-tarballs.py
```diff
@@ -248,17 +252,53 @@ def backup_to_s3(s3_obj, logger, controller_path, controller, tb, tar, resultnam
...
+        # calculate multi etag value
+        local_multipart_etag = calculate_multipart_etag(tar, chunk_size)
+        try:
+            with open(tar, "rb") as f:
+                try:
+                    s3_obj.connector.upload_fileobj(f,
+                                                    Bucket=s3_obj.bucket_name,
+                                                    Key=s3_resultname,
+                                                    Config = multi_upload_config,
+                                                    ExtraArgs = {'Metadata': { 'MD5': local_multipart_etag}})
```
ndokos: I think there should be two pieces of metadata: the md5sum and the calculated etag.

server/bin/pbench-backup-tarballs.py
```diff
@@ -248,17 +252,53 @@ def backup_to_s3(s3_obj, logger, controller_path, controller, tb, tar, resultnam
...
+                        "Multi-upload to s3 failed, client error: {}".format(e)
+                    )
+                    return Status.FAIL
+                else:
+                    # compare the multi etag value uploaded in metadata field # with s3 etag for data integrity.
+                    try:
+                        obj = s3_obj.connector.get_object(Bucket='{}'.format(
+                            s3_obj.bucket_name), Key='{}'.format(s3_resultname))
+                    except Exception:
+                        logger.exception("Unable to get the etag from S3: {}".format(s3_resultname))
```
ndokos: The message is wrong: it should say `get_object failed`.

server/lib/s3backup/__init__.py
```diff
@@ -42,6 +42,13 @@ def put_object(self, Bucket, Key, Body, ContentMD5):
         return self.s3connector.put_object(Bucket=Bucket, Key=Key,
                 Body=Body, ContentMD5=ContentMD5)
 
+    def upload_fileobj(self, read_content, Bucket, Key, Config, ExtraArgs):
```
ndokos: If you make the above a method in `s3Connector` then these two routines are unnecessary.

server/lib/s3backup/__init__.py
```diff
@@ -112,6 +119,22 @@ def put_object(self, Bucket=None, Key=None, Body=None, ContentMD5=None):
             with open('{}/{}/{}'.format(self.path, self.bucket_name, Key), 'wb') as f:
                 f.write(Body.read())
 
+    def upload_fileobj(self, read_content=None, Bucket=None, Key=None, Config=None, ExtraArgs=None):
```
ndokos: ... and maybe these two as well. but we'll have to take a closer look at the unit tests.

server/lib/s3backup/__init__.py
```diff
@@ -158,3 +181,26 @@ def __init__(self, config):
...
+            data = fp.read(chunk_size)
+            if not data:
+                break
+            md5s.append(hashlib.md5(data))
+
+    if len(md5s) > 1:
+        digests = b"".join(m.digest() for m in md5s)
+        new_md5 = hashlib.md5(digests)
+        new_etag = '%s-%s' % (new_md5.hexdigest(),len(md5s))
+    elif len(md5s) == 1:
```
ndokos: This should be a `cannot happen`, correct? We *must* have `len(md5s) > 1`.

server/bin/pbench-backup-tarballs.py
tenstormavi: yes, it only gets a header.

server/bin/pbench-backup-tarballs.py
```diff
@@ -19,7 +19,7 @@
 
 # The link source and destination for this operation of this script.
 _linksrc = "TO-BACKUP"
-_linketagfail = "ETAG-FAILED"
+_linkdestfail = "BACKED-UP-FAILED"
```
ndokos: BACKUP-FAILED sounds better.

server/lib/pbench/__init__.py
```diff
@@ -235,7 +235,7 @@ def __init__(self, cfg_name):
                 " TO-LINK" \
                 " TO-INDEX INDEXED WONT-INDEX" \
                 " TO-COPY-SOS COPIED-SOS" \
-                " TO-BACKUP BACKED-UP ETAG-FAILED" \
+                " TO-BACKUP BACKED-UP BACKED-UP-FAILED" \
```
ndokos: BACKUP-FAILED

server/lib/s3backup/__init__.py
```diff
@@ -160,9 +159,9 @@ def put_tarball(self, Name=None, Body=None, Size=0, ContentMD5=None, Bucket=None
                             Bucket=self.bucket_name, Key=Key)
                         self.logger.error("Multi-upload to s3 failed:"
                                           " {key},etag doesn't match,"
-                                          " so moving it to ETAG_FAILURE for inspection".
+                                          " so moving it to BACKED-UP-FAILED for inspection".
```
ndokos: BACKUP-FAILED
But this is a somewhat presumptuous message, because `put_tarball` does *not know* what its called will do with the failure. It's better to limit the logging here to what `put_tarball` knows went wrong and leave details like what happens to the tarball to the caller to log for itself (if necessary).

server/bin/pbench-backup-tarballs.py
```diff
@@ -102,12 +95,13 @@ def sanity_check(lb_obj, s3_obj, config, logger):
...
+            'The QUARANTINE path {}, resolves to {}'
+            'which is not a directory'.format(qdir, os.path.realpath(qdir)))
         lb_obj = None
 
     # make sure the S3 bucket exists
     try:
-        s3_obj.connector.head_bucket(Bucket='{}'.format(s3_obj.bucket_name))
+        s3_obj.head_bucket(s3_obj.bucket_name)
     except Exception:
         logger.exception(
```
ndokos: Let's get rid of the exception traceback both here and on line 79: ;log them as errors instead (and fix the gold files) - the traceback is not telling us anything useful.

server/bin/gold/test-19.txt
```diff
@@ -0,0 +1,287 @@
++++ Running pbench-move-unpacked
```
portante: Didn't `pbench-move-unpacked` get removed?  Is there a need to add it back?

server/bin/pbench-verify-backup-tarballs.py
```diff
@@ -308,7 +308,7 @@ def main():
             report.write("{}.{}({})\n".format(prog, config.TS, config.PBENCH_ENV))
 
             try:
-                # Check the data integrity in ARCHIVE (Question 1).
+                # Check the data integrity in ARCHIVE.
```
portante: Why remove "Question 1"?  These are here because the algorithm described at the top of the module asked those questions, and the comment records where that question is answered.

If you don't want them, can you find another way to connect the module comment to the code?

Issue #1179, opened 2019-04-25
trap command to handle `pbench-metadata-log` is missing at the end of every benchmark run, except `pbench-user-benchmark`


tenstormavi:
while working on issue #1085, I found that except `pbench-user-benchmark` every other benchmark script has no trap command to handle the `pbench-metadata-log` at the end of every run. I think it should be there.

portante:
Fixed by #1261.

PR #1261, opened 2019-07-15
remove .iterations files created by bench mark scripts

### LINKED ISSUE #1085: pbench-agent bench mark scripts create .iteration files in /var/lib/pbench-agent/*.iterations
Those files should either be removed or added to the run directory.

tenstormavi:
Fixes #1085 #1179 
pbench-agent bench mark scripts create .iteration files in /var/lib/pbench-agent/*.iterations. These files should be removed at the end of every bench mark script. And solves the missing trap command to handle `pbench-metadata-log` at the end of every benchmark run.


agent/bench-scripts/pbench-user-benchmark
```diff
@@ -160,7 +160,7 @@ function record_iteration {
 }
 record_iteration ${iteration} ${iteration_name} ${benchmark_bin}
 # on abnormal exit, make sure that the metadata log exists and is complete.
-trap "pbench-metadata-log --group=$tool_group --dir=$benchmark_run_dir int" INT QUIT
+trap "pbench-metadata-log --group=$tool_group --dir=$benchmark_run_dir int; benchmark_iterations" INT QUIT
```
portante: So this `trap` appears to do two things:

 1. Runs the command `pbench-metadata-log --group=$tool_group --dir=$benchmark_run_dir int`
 1. Tries to run `benchmark_iterations`, which does not appear to be a command or a variable

It seems that it would be useful to just move `$benchmark_iterations` from `$pbench_tmp` to `$pbench_run` so that it is preserved in the data set.  It would be useful for debugging.

agent/bench-scripts/pbench-dbench
```diff
@@ -318,6 +318,8 @@ max_key_length=0
 export benchmark config
 pbench-collect-sysinfo --group=$tool_group --dir=$benchmark_run_dir --sysinfo=$sysinfo beg
 pbench-metadata-log --group=$tool_group --dir=$benchmark_run_dir beg
+# on abnormal exit, make sure that the metadata log exists and is complete.
+trap "pbench-metadata-log --group=$tool_group --dir=$benchmark_run_dir int; mv $benchmark_iterations $pbench_run" INT QUIT
```
portante: Can we consider changing the definition of `${benchmark_iterations}` so that it is kept only in `${pbench_run}`?

ndokos: @portante: I think it should be `${benchmark_run_dir}`, not `${pbench_run}` - correct? Something like this:


    benchmark_iterations="${benchmark_run_dir}/${benchmark_fullname}.iterations"

Do we want the file name to contain the full benchmark name, or to be simply `iterations`?

portante: Seems like having it named `${benchmark_run_dir}/.iterations` would keep it from conflicting but make sure it is not prominent.

agent/util-scripts/gold/pbench-metadata-log/test-22.txt
```diff
@@ -46,7 +46,6 @@ iter3
 /var/tmp/pbench-test-utils/pbench/metadata.log:config = 
 /var/tmp/pbench-test-utils/pbench/metadata.log:date = 1900-01-01T00:00:00
 /var/tmp/pbench-test-utils/pbench/metadata.log:rpm-version = 
-/var/tmp/pbench-test-utils/pbench/metadata.log:iterations = iter1, iter2, iter3
```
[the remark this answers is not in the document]
portante: Yes, `pbench-metadata-log` can't find the `.iteration` file because it is still located in the `tmp/` directory.  If we just move `samples/pbench-metadata-log/test-22/pbench/tmp/pbench.iterations` to `samples/pbench-metadata-log/test-22/pbench/.iterations`, the unit test should pass without the removal of line 49.

portante:
And can you also rebase and squash down these commits to one?  Thanks!

agent/util-scripts/gold/pbench-metadata-log/test-22.txt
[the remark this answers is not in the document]
portante: This is still not right, it appears.  Line 49 in the original is being removed.  It should stay.  It appears the `.iterations` file has been moved to the wrong location.  See the  comment above for the correct place.

PR #1266, opened 2019-07-15
Update the version of the pbench-fio package required to 3.12.

ndokos:


ndokos:
There was a second `pbench-agent-default.cfg` file that was used for (almost) all installations. That file is going away, so the change in this PR is necessary in order for things to continue to work.

portante:
As we talked this was working in all our environments because the RPMs we built from this source code has their own `pbench-agent.cfg` files with the correct `pbench-fio` version of `3.12` in them.

Issue #1275, opened 2019-07-20
TravisCI builds seem to fail due to "Cannot find module 'eslint-config-prettier/@typescript-eslint'"


portante:
All TravsiCI builds seem to be failing because of the following:

```
$ umi test ./src/components ./src/pages --verbose --maxWorkers=1 --runInBand
✖  error     Error: Cannot find module 'eslint-config-prettier/@typescript-eslint' 
Referenced from: 
    at ModuleResolver.resolve (/home/travis/build/distributed-system-analysis/pbench/web-server/v0.4/node_modules/eslint/lib/util/module-resolver.js:72:19)
    at resolve (/home/travis/build/distributed-system-analysis/pbench/web-server/v0.4/node_modules/eslint/lib/config/config-file.js:507:28)
    at load (/home/travis/build/distributed-system-analysis/pbench/web-server/v0.4/node_modules/eslint/lib/config/config-file.js:579:26)
    at configExtends.reduceRight (/home/travis/build/distributed-system-analysis/pbench/web-server/v0.4/node_modules/eslint/lib/config/config-file.js:453:36)
    at Array.reduceRight (<anonymous>)
    at applyExtends (/home/travis/build/distributed-system-analysis/pbench/web-server/v0.4/node_modules/eslint/lib/config/config-file.js:431:26)
    at Object.loadObject (/home/travis/build/distributed-system-analysis/pbench/web-server/v0.4/node_modules/eslint/lib/config/config-file.js:566:35)
    at new Config (/home/travis/build/distributed-system-analysis/pbench/web-server/v0.4/node_modules/eslint/lib/config.js:85:46)
    at new CLIEngine (/home/travis/build/distributed-system-analysis/pbench/web-server/v0.4/node_modules/eslint/lib/cli-engine.js:466:23)
    at Object.<anonymous> (/home/travis/build/distributed-system-analysis/pbench/web-server/v0.4/node_modules/sylvanas/lib/eslintJs.js:6:16)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
yarn unit tests failed!
```

portante:
Fixed by #1284.

PR #1279, opened 2019-07-22
Add path analyzer for indexing ToC data

portante:
Uses the "`path_hierarchy`" tokenizer for the directory and link fields of the table-of-contents JSON documents. The tokenizer creates multiple tokens out of paths, e.g. the path "`/home/foo/bar/file.txt`" would end up with four tokens when indexed into Elasticsearch [1]:

 * `/home`
 * `/home/foo`
 * `/home/foo/bar`
 * `/home/foo/bar/file.txt`

[1] https://www.elastic.co/guide/en/elasticsearch/reference/1.7/analysis-pathhierarchy-tokenizer.html

### DIFF
--- a/server/bin/gold/test-7.0.1.txt
+++ b/server/bin/gold/test-7.0.1.txt
@@ -581,13 +581,13 @@ Template: pbench-unittests.v1.run
                     "type": "date"
                 },
                 "directory": {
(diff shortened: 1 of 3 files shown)


portante:
Merging despite TravisCI failures since they are the result of known issues documented with #1275.

PR #1289, opened 2019-07-30
Decompose pbench-move-results into "make" and "move" components

portante:
We now have separate commands for "making a result tar ball" and for "moving" a result tar ball", `pbench-make|move-result-tb`. By breaking this command up into its loop components, we can use those components for other purposes, e.g. allowing a caller to invoke "make" and "move" in different environments.

One such intended use is to drive server side testing where an existing set of tar balls can be "moved" repeatedly to verify a staging environment.

### DIFF
--- a/agent/util-scripts/pbench-move-results
+++ b/agent/util-scripts/pbench-move-results
@@ -1,9 +1,9 @@
 #!/bin/bash
 # -*- mode: shell-script; indent-tabs-mode: t; sh-basic-offset: 4; sh-indentation: 4; sh-indent-for-case-alt: + -*-
 
-script_path=`dirname $0`
-script_name=`basename $0`
-pbench_bin="`cd ${script_path}/..; /bin/pwd`"
+script_path="$(dirname ${0})"
+script_name="$(basename ${0})"
+pbench_bin="$(realpath -e ${script_path}/..)"
 
 # source the base script
 . "$pbench_bin"/base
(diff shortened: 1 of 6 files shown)


agent/util-scripts/pbench-move-results
```diff
@@ -115,170 +116,85 @@ fi
...
+    debug_log "the following ssh command failed: \"ssh -q -i ${pbench_bin}/id_rsa ${ssh_opts} ${results_repo} exit\""
     exit 1
 fi
 
 let runs_copied=0
 let failures=0
 
-tmp=${pbench_tmp}/${script_name}.$$
-trap "rm -rf $tmp" EXIT INT QUIT
+tmp=${pbench_tmp}/${script_name}.${$}
```
ndokos: I have to say that this looks strange :-)

agent/util-scripts/pbench-move-results
```diff
@@ -115,170 +116,85 @@ fi
...
+trap "rm -rf ${tmp}" EXIT INT QUIT
 
-mkdir -p $tmp/$controller
-sts=$?
-if [[ $sts -ne 0 ]] ;then
-    error_log "Failed: \"mkdir -p $tmp/$controller\", status $sts"
+mkdir -p ${tmp}/${controller}
+sts=${?}
```
ndokos: ... and so does this.

agent/util-scripts/pbench-move-result-tb
```diff
@@ -0,0 +1,85 @@
...
+if [[ ! -f "${tarball}" ]]; then
+    error_log "ERROR: tar ball does not exist, ${tarball}"
+    exit 1
+fi
+if [[ ! -f "${tarball}.md5.check" ]]; then
+    error_log "ERROR: tar ball's .md5.check does not exist, ${tarball}.md5.check"
+    exit 1
+fi
+controller_dir=$(dirname ${tarball})
+cnt=$(find ${controller_dir} -type f 2> /dev/null | wc -l 2> /dev/null)
```
ndokos: Any errors in removing a previous result's artifacts will cause cascading errors for all subsequent tarballs: are we OK with that? Maybe we should relax the check to be just for this tarball: `find ....-name "${tarball}.*"` ?

portante: Addressed this by checking the error status of the `rm` commands and the `touch` command.

PR #1291, opened 2019-08-01
Problems while testing v0.63 release candidate

ndokos:
This PR addresses a problem in bench-scripts/unittests and associated problems. There are three commits:

 - the first commit fixes the script and modifies some gold files to accommodate those fixes. That uncovers the failures that should have been there, but were not, when PR#1261 was tested.
- the second commit fixes  pbench-uperf and pbench-user-benchmark so that the unit tests all pass.
- the third commit applies the same fix to all the other bench scripts (which have no unit test coverage).

### DIFF
--- a/agent/bench-scripts/test-bin/mock-cmd
+++ b/agent/bench-scripts/test-bin/mock-cmd
@@ -1,3 +1,16 @@
 #!/bin/bash
 
 echo "$0 $*" >> $_testlog
+
+prog=$(basename "$0")
+args=( "$@" )
+lastidx=$((${#args[@]} - 1))
+# avoid "bad array subscript" error
+if [[ "$lastidx" -lt 0 ]] ;then
+   exit 0
+fi
+
+last=${args[$lastidx]}
+if [[ "$prog" == "pbench-metadata-log" ]] && [[ "$last" == "beg" ]];then
+    echo [pbench] > ${benchmark_run_dir}/metadata.log
+fi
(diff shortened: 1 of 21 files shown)


agent/bench-scripts/test-bin/mock-cmd
```diff
@@ -1,3 +1,9 @@
 #!/bin/bash
 
 echo "$0 $*" >> $_testlog
+
+prog=$(basename "$0")
+last=$(echo $* | sed 's/.* //')
```
portante: Can we do something like the following to avoid using `sed` (causing an extra fork):
```
args=( "$@" )
lastidx=$((${#args[@]} - 1))
last=${args[$lastidx]}
```

agent/bench-scripts/unittests
```diff
@@ -130,10 +130,11 @@ function _dump_logs {
     _dump_file ${1} user-benchmark-summary.json
     _dump_file ${1} user-benchmark.cmd
     (
-    cd ${_testdir}/${1}
-    for f in $(find . -name results.html 2> /dev/null | sort); do
-        _dump_file ${1} ${f}
-    done
+     if cd ${_testdir}/${1} > /dev/null 2>&1 ;then
```
portante: Seems like it works, but indentation appears to be off.

portante:
Also, when referencing a PR in commit messages or comments, can you use the form "PR #1261" instead of "PR#1261"?  The space before the pound sign (`#`) let's GitHub automatically create URL links to the PR or issue.

PR #1293, opened 2019-08-01
PbenchBase.pm: add required metadata log for benchmark run name

k-rister:
- Without this metadata log entry you cannot use pbench-move-results
  to upload the data.


portante:
@atheurer, this looks good.  I'll merge it if you can sign off as well.

PR #1295, opened 2019-08-02
Fix name of path_tokenizer

portante:
PR #1279 added the name of the "path tokenizer" as `path-tokenizer`, but was referencing it as `path_tokenizer`.  This change corrects that.


portante:
Now depends on #1294.

Issue #963, opened 2018-10-22
Switch to using CentOS 7 containers for unit tests


portante:
See https://stackoverflow.com/questions/32535195/how-to-run-tests-on-centos-7-with-travis-ci

webbnh:
At this point, CentOS 7 is no longer interesting, and we're using Jenkins rather than Travis...so, I'm thinking that this Issue is now obsolete.

PR #1298, opened 2019-08-07
Cdm benchmarks

atheurer:
This should not affect any production-level code in pbench, just "beta" code for pbench-run-benchmark and CDM.

This is a bunch of updated for pbench-run-benchmark & CDM, hopefully to get us much closer to more people trying out pbench-run-benchmark.  Notable changes:
-CDM
  -updated to use cdmv4dev
-ES Queries:
  - new pbench perl library for Elastic Search queries
  - sample query scripts
-pbench-run-benchmark
  - by default generate the standard result html/json, not the CDM, so results just work on pbench-web


atheurer:
@portante @ndokos rightnow this requires perl-UUID, which I have discovered comes from EPEL, so I will have to research alternatives.  There is a "epel-perl" package that does come from rhel-7-optional, so maybe that will work, but I'll have to see how it integrates with Perl.

portante:
Can we squash this commits down to one?  Or perhaps to 2 or 3?

Also, any way we can get the unit tests to pass as well?

atheurer:
> Can we get the unit tests to pass? _Seems_ like that is indicating these changes affect the non-CDM side.

The Perl module problem for UUID is precisely why those unittests don't pass.  This is why I need to change it to a module which we know we can support easily (not requiring EPEL), but also the TravisCI will need to change to support having that module available when unittests are run.  I know nothing about Travis, so I need you or Nick to let me know if using the perl module "perl-uuid" is possible in Travis.  I don't know if this module is installed via the same package name in Travis as it is in RHEL.  This is how it's named for RHEL:

<pre>
Installing:
 uuid-perl         x86_64                  1.6.2-26.el7          rhel-7-optional                   29 k</pre>

atheurer:
Now squashed to a single commit

portante:
I'll see if I can find the TravisCI magic for that.

portante:
This might be a good point to bump the priority of containerizing the unit tests, issue #963.

atheurer:
I believe the Ubuntu equivalent for uuid-perl is libossp-uuid-perl:

https://launchpad.net/ubuntu/+source/ossp-uuid

atheurer:
I have added code for conditional loading of perl module for CDM (and therefore Data::UUID) and it looks like the unit tests are passing now.

atheurer:
> I think this PR needs to be rebased after some recent changes went in.

It has been rebased as of today

k-rister:
Hmm, if it has been rebased on master then why isn't the code #1303 present in pbench-run-benchmark?

atheurer:
> Hmm, if it has been rebased on master then why isn't the code from #1303 present in pbench-run-benchmark?

I see it in my branch: https://github.com/atheurer/pbench/blob/0b3cb6b0fef098d2ee9c1188e9c549fbfa3adae1/agent/bench-scripts/pbench-run-benchmark#L280

k-rister:
Ok, I see it now too.  I must have messed up when I updated the branch.

PR #1300, opened 2019-08-08
Fixes for unpack tarballs

portante:
First, we remove `pbench-move-unpacked`. It is no longer used in any of our deployments, and it is a hassle to keep around the code. Removing it now is less painful than trying to remove it later when other code changes may be necessary.

Second, we now fix the order in which operations performed. The indexing and the copying of sosreports now follows unpacking of tar balls. This removes a race condition where `pbench-copy-sosreports` may miss seeing sosreports because the tar ball is not unpacked. It also gives us the opportunity to leverage the already unpacked tar balls for the indexing step (planned for a future commit).

This commit depends on PRs #1294 and #1295.


server/bin/state/config-satellite/pbench-server.cfg
```diff
@@ -11,6 +11,7 @@ pbench-top-dir = %(unittest-dir)s/pbench-satellite
 pbench-local-dir = %(unittest-dir)s/pbench-satellite-local
 # A satellite just unpacks and syncs
 dispatch-states = TO-UNPACK, TO-SYNC
+unpacked-states = UNPACKED
```
ndokos: UNPACKED is the standard link desitnation for pbench-unpack-tarballs, I think `unpacked-states` should be all the others that depend on the unpacked tarball being there: TO-INDEX etc.

portante: Good catch.

server/lib/config/pbench-server-default.cfg
```diff
@@ -83,10 +83,10 @@ use-mail-for-status = no
 # When a tarball comes in to this server, what processing should it
 # undergo?  Production-like servers do at least the first two, and
 # possibly more, of the following dispatch states:
-#dispatch-states = TO-UNPACK, TO-INDEX, TO-COPY-SOS, TO-BACKUP
+dispatch-states = TO-UNPACK, TO-INDEX, TO-COPY-SOS, TO-BACKUP
 
 # Satellite servers typically do at least this dispatch state:
```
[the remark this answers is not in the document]
portante: Fixed.

server/bin/state/config-satellite/pbench-server.cfg
portante: Fixed.

PR #1301, opened 2019-08-08
Remove switch

tc-wilson:
Removed the use of perl Switch in fio-postprocess


portante:
@atheurer, just being safe here, mind reviewing before we merge?

portante:
@tc-wilson, @atheurer, do you think we can finalize this work for 0.64 soon?

PR #1303, opened 2019-08-09
update the closest samples index handling

k-rister:
- The logic that finds the closest samples index for an iteration
  needs to be updated to properly handle different indexing schemes.
  For example, the old pbench style of indexing starts at 1
  (ie. sample1) while the new pbench-run-benchmark style of indexing
  starts at 0 (ie. sample0).

- Update the hyperlinks that are generated for the HTML output from
  generate-benchmark-summary to use the $iteration_name_format to
  properly recreate the iteration name.

### DIFF
--- a/agent/bench-scripts/pbench-run-benchmark
+++ b/agent/bench-scripts/pbench-run-benchmark
@@ -277,6 +277,14 @@ while (scalar @param_sets > 0) {
 	$iteration_label =~ s/^\s+//;
 	$iteration_label =~ s/\s/_/g;
 
+	# if the logic for determining a label for the iteration has
+	# yielded an empty string just use the benchmark name -- this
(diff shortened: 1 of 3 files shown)


agent/bench-scripts/pbench-run-benchmark
```diff
@@ -277,6 +277,14 @@ while (scalar @param_sets > 0) {
 	$iteration_label =~ s/^\s+//;
 	$iteration_label =~ s/\s/_/g;
 
+	# if the logic for determaining a label for the iteration has
```
portante: s/determaining/determining/

k-rister: done

k-rister: done

PR #1172, opened 2019-04-17
Comparison tool shareable URL generator 

Anisha1234:
Will fix #844 

In order to enable productivity features within pbench dashboard and eventually benchpress dashboard in the future — our platform requires an efficient backend for supporting dynamic storage and retrieval of user data. 
This platform consists of a Postgres database exposed via a GraphQL API server. This API server is powered by Prisma in order to enable high performance querying, a GUI client for visualizing queries, and a realtime event system for database event tracking.

The backend environment is currently being tracked at the following GitHub repo:
https://github.com/gurbirkalsi/pbench_dashboard_backend

This PR involves:

- UI for shareable URL generation
- Saving whole redux store data to Postgres database
- Retrival and Re-rendering of entire redux component with the fetched data.

PR #1308, opened 2019-08-12
Reference application data via static `appName` field in `app.js`

gurbirkalsi:
Separates user selected data and datastore data into separate stores of
`global` and `datastore`. The `datastore` Redux model is now blacklisted
to support constant updates to server side data.

### DIFF
--- a/web-server/v0.4/src/app.js
+++ b/web-server/v0.4/src/app.js
@@ -7,13 +7,13 @@ import storage from 'redux-persist/lib/storage';
  * `dashboard` and `search` are blacklisted from the saved state as the namespaces
  * persist data that is constantly updated on the server side.
  */
-const appPath = window.location.pathname.split('/').pop();
+const appName = 'dashboard';
 
 const persistConfig = {
   timeout: 1000,
-  key: appPath !== '' ? appPath : 'root',
+  key: appName,
   storage,
-  blacklist: ['dashboard', 'search'],
+  blacklist: ['dashboard', 'search', 'datastore'],
 };
 
 const persistEnhancer = () => createStore => (reducer, initialState, enhancer) => {
--- a/web-server/v0.4/src/app.js
+++ b/web-server/v0.4/src/app.js
@@ -1,18 +1,18 @@
 import { persistStore, persistReducer } from 'redux-persist';
 import storage from 'redux-persist/lib/storage';
 
+import { getAppPath } from './utils/utils';
+
 /*
  * redux persist configuration for saving state object to persisted storage.
  *
  * `dashboard` and `search` are blacklisted from the saved state as the namespaces
  * persist data that is constantly updated on the server side.
  */
-const pathKey = window.location.pathname.split('/')[1];
-const appPath = pathKey || 'dashboard';
 
 const persistConfig = {
   timeout: 1000,
-  key: appPath,
+  key: getAppPath(),
   storage,
   blacklist: ['dashboard', 'search', 'datastore'],
 };
--- a/web-server/v0.4/src/app.js
+++ b/web-server/v0.4/src/app.js
@@ -7,11 +7,12 @@ import storage from 'redux-persist/lib/storage';
  * `dashboard` and `search` are blacklisted from the saved state as the namespaces
  * persist data that is constantly updated on the server side.
  */
-const appName = 'dashboard';
+const pathKey = window.location.pathname.split('/')[1];
+const appPath = pathKey || 'dashboard';
 
 const persistConfig = {
   timeout: 1000,
-  key: appName,
+  key: appPath,
   storage,
   blacklist: ['dashboard', 'search', 'datastore'],
 };
--- a/web-server/v0.4/README.md
+++ b/web-server/v0.4/README.md
@@ -83,6 +83,21 @@ export default {
 }
 ```
 
+## Storage Config
+
+Pbench Dashboard stores application data using local browser storage given an application key defined in `app.js`. To deploy multiple instances of the dashboard on the same domain, change the `appName` field to a unique value for each deployment. 
+
+```JavaScript
+const appName = 'dashboard';
+
+const persistConfig = {
+  timeout: 1000,
+  key: appName,
+  storage,
+  blacklist: ['dashboard', 'search'],
+};
+```
+
 ## Build
 
 Build Application
--- a/web-server/v0.4/src/models/datastore.js
+++ b/web-server/v0.4/src/models/datastore.js
@@ -40,6 +40,10 @@ export default {
         type: 'getMonthIndices',
         payload: indices,
       });
+      yield put({
+        type: 'global/updateSelectedIndices',
+        payload: [indices[0]],
+      });
     },
   },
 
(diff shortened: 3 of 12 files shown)


gurbirkalsi:
@Anisha1234 Please let me know if these changes resolve the issues we observed with recreating user sessions in PR #1172.

web-server/v0.4/src/app.js
```diff
@@ -7,13 +7,13 @@ import storage from 'redux-persist/lib/storage';
  * `dashboard` and `search` are blacklisted from the saved state as the namespaces
  * persist data that is constantly updated on the server side.
  */
-const appPath = window.location.pathname.split('/').pop();
```
portante: This means we can't load different versions of the dashboard, right?

I'd like to be able to handle data migrations by loading two different configurations.

gurbirkalsi: This does allow the deployment of multiple versions of the dashboard; however, the `appName` field needs to be changed for each deployment so each deployment reads from their respective local storage key. I've added details for accomplishing this with the changes to the `README.md`.

portante: Not sure, but that looks like a code change before you build the application.  Why not have the `appName` read from the on-disk configuration file with everything else?

gurbirkalsi: I agree! Should be introduce a new configuration file for app specific configuration fields? We're only referencing datastore specific fields in the current `config.json.j2` file.

portante: We can introduce a new configuration file, but that would just be like using the window location as well, since each app would have its own path that way.

Why do we have to remove the `window.location.pathname.split('/').pop()` method?

gurbirkalsi: Referencing the last slash separated path item in the url causes multiple local Redux stores to be created since there are many routes within the app (especially in the case of URL session generation where we create a unique ID per session). 

We can simply reference the first path item to avoid this behavior - this will also maintain the naming logic when we deploy the app.

Anisha1234:
@gurbirkalsi @portante I tested the changes of this PR on top of my sharable URL generation PR #1172 . It is throwing an error on every page except the dashboard page.  

1. Summary page: "run.name of undefined" 
2. Results page: "hits of undefined"
(when the data returned from the GraphQL query contains that field)
3. Comparision-select page: Rendering blank page
4. Run-comparison: Rendering blank page

I am not sure if this behaviour is because it is not being able to rehydrate the whole redux session from the shared config.

Anisha1234:
@gurbirkalsi @portante I tested the changes of this PR again my PR #1172 for sharable URL generation. I found an error on every page except the landing page(dashboard component). The error says "`result.iterationData.forEach is not a function"` . When I checked the data, I found an HTML string in the place of **iterationData**. I am not sure if it is because of the rehydration of the redux store.

gurbirkalsi:
@Anisha1234 Can you please share the full contents of the logs emitted to the dev console from this error?

Anisha1234:
@gurbirkalsi this is the sample iterationData before sending it to server.
```
iterationData: {
0:
iteration_data:
parameters: {benchmark: Array(1)}
throughput: {Gb_sec: Array(2)}
__proto__: Object
iteration_name: "tcp_stream-64B-1i"
iteration_number: 1
__proto__: Object}
```

And this is what is the output from parse.js after fetching the data from server
```
{iterationData: "<!DOCTYPE html><html lang="en"><head>↵↵<link rel="…<script src="/umi.js"></script>↵↵↵↵</body></html>", controllerName: "undefined", resultName: "incoming", tableId: 0}
controllerName: "undefined"
iterationData: "<!DOCTYPE html><html lang="en"><head>↵↵<link rel="stylesheet" href="/umi.css">↵↵  <meta charset="UTF-8">↵  <meta http-equiv="X-UA-Compatible" content="IE=edge">↵  <meta name="viewport" content="width=device-width, initial-scale=1">↵  <title>Pbench Dashboard</title>↵  <link rel="icon" href="/favicon.ico" type="image/x-icon">↵<script>↵  window.routerBase = "/dashboard";↵</script>↵</head>↵↵<body>↵  <div id="root"></div>↵<script src="/umi.js"></script>↵↵↵↵</body></html>"
resultName: "incoming"
tableId: 0
__proto__: Object}
```
Here is the complete error log
```
uncaught at _callee3 at _callee3 
 at _callee6 
 at takeEvery(dashboard/fetchIterations, _callee) 
 at _callee 
 at _callee 
 Error: TypeError: result.iterationData.forEach is not a function
    at http://localhost:8000/umi.js:212963:26
    at Array.forEach (<anonymous>)
    at parseIterationData (http://localhost:8000/umi.js:212949:11)
    at _callee6$ (http://localhost:8000/umi.js:207973:71)
    at tryCatch (http://localhost:8000/umi.js:195964:40)
    at Generator.invoke [as _invoke] (http://localhost:8000/umi.js:196190:22)
    at Generator.prototype.<computed> [as next] (http://localhost:8000/umi.js:196016:21)
    at next (http://localhost:8000/umi.js:184387:27)
    at currCb (http://localhost:8000/umi.js:184464:7)
    at http://localhost:8000/umi.js:71246:17
    at http://localhost:8000/umi.js:70266:26
    at onError (http://localhost:8000/umi.js:71247:11)
    at sagaWithCatch$ (http://localhost:8000/umi.js:70792:13)
    at tryCatch (http://localhost:8000/umi.js:195964:40)
    at Generator.invoke [as _invoke] (http://localhost:8000/umi.js:196190:22)
    at Generator.prototype.<computed> [as throw] (http://localhost:8000/umi.js:196016:21)
    at next (http://localhost:8000/umi.js:184365:32)
    at Object.currCb [as cont] (http://localhost:8000/umi.js:184464:7)
    at end (http://localhost:8000/umi.js:184432:23)
    at abort (http://localhost:8000/umi.js:184155:5)
    at Object.task.cont (http://localhost:8000/umi.js:184168:9)
    at next (http://localhost:8000/umi.js:184404:16)
    at currCb (http://localhost:8000/umi.js:184464:7)
```

portante:
@gurbirkalsi, can you take a look at what Anisha has posted?

gurbirkalsi:
@portante Anisha and I discussed potential root causes for this issue yesterday during our sync. She is going to reset her localStorage cache and confirm the data we're posting to the graphql instance is correct. The changes to the redux model namespaces here are likely causing differences in data here.

Anisha1234:
> reset her localStorage cache

@gurbirkalsi I tested after clearing my localStorage cache but the problem still persists.

Anisha1234:
> @portante Anisha and I discussed potential root causes for this issue yesterday during our sync. She is going to reset her localStorage cache and confirm the data we're posting to the graphql instance is correct. The changes to the redux model namespaces here are likely causing differences in data here.

@gurbirkalsi I tried clearing my localStorage, but the problem still seems to persist.

gurbirkalsi:
@Anisha1234 I wasn't able to reproduce the issue you reported above with your updated PR; however, url generation is now working with the latest change I've made to retrieve the dashboard app path for Redux Persist. 

You should reference the `getAppPath()` utility function in the dashboard Redux model and concat the `persist` key that is included at the library level:

```javascript
window.localStorage.setItem(`persist:${getAppPath()}`, parsedConfig);
```

portante:
Let's merge this commit as is, and further debug on top of master.

Issue #1264, opened 2019-07-15
We need to add tracking information to the pbench-server for what was indexed when and by who


portante:
We recently had a problem determining what was indexing data into a given Elasticsearch instance.

We need to track in the metadata of then run document the following additional pieces of information (for starters):

 * the hostname, pid, and thread ID of the indexer
 * the git version of the code base used
 * the RPM version of the code base deployed
 * a timestamp of when the RUN document was generated by the indexer (since the @timestamp field is derived from the start time of the pbench run itself, which is most likely in the past)

portante:
We should also a log entries reporting when:

- run data is indexed (start, stop)
- result data is indexed (start, stop)
- which iteration of how many are being processed

portante:
To elaborate a bit further, the problem is when one performs a query to Elasticsearch and gets back data that does not look right.

We encountered a problem recently where the indexing changed unexpectedly and we could not tell what or who or how things had changed because we had no indication in the JSON data.  There was no trace of what host sent what JSON document at what time, just the resulting JSON payload transformed from the pbench data.

While the JSON data is *usually* good, having a signature on each JSON document would allow us to look up what entity actually performed the indexing, which would include the actual time of indexing, the hostname, the pid, gid, uid, version (if available), and git commit hash (if available).  In the face of problems, this will be invaluable.

Further, this tracking will let us build a dashboard of the report-status data to that we can see how well things are working operationally.

PR #1309, opened 2019-08-14
WIP: enhanced report status for tracking

portante:
Enhancements to use a new class, `Report()`, to construct all data related to the operation of a particular server-side script.  That indexed JSON document's ID is used at the contents of the `@generated-by` field for all indexed pbench data.  This allows us to put arbitrary information about the process performing the indexing, without adding unneeded metadata to all indexed documents, while allowing us to track down the origin of the indexer.


portante:
This needs much work to get all the unit tests to pass, but seems promising.

k-rister:
> Enhancements to use a new class, `Report()`, to construct all data related to the operation of a particular server-side script. That indexed JSON document's ID is used at the contents of the `@generated-by` field for all indexed pbench data. This allows us to put arbitrary information about the process performing the indexing, without adding unneeded metadata to all indexed documents, while allowing us to track down the origin of the indexer.

For those of us that aren't real familiar with this side of things can you explain the comment "allowing us to track down the origin of the indexer"?  I'm confused as to why that would not be obvious, but I probably don't understand how it actually works.

portante:
First, see issue #1264.

The second reason for this change is that we are changing the server indexing to break up indexing of pbench data into two steps (see merged PR #1306):

 1. Index run, table-of-contents, & result data
 2. Index tool data

In the past, we placed the small amount of tracking data in the run document, and since all other documents have the ID of the run document with which they are associated, we could use that run document to identify what entity performed the indexing.

Now that we are breaking up the indexing steps, we could potentially have two different entities indexing the tool data and the run data.

So by use the report status data as the tracking ID for all indexed documents, and enhancing all pbench server side processes to generate a report with the proper tracking information, we can run separate services entirely while still tracking who did what, when and how.

k-rister:
Isn't all indexing being performed in a controlled process on the pbench server?  I'm not understanding how you wouldn't know what entity performed the indexing.

Note, I'm not trying to argue against what you are doing in any way, I'm just trying to understand how things are working.  It almost reads like we have results being indexed from unknown locations (servers?) by unknown processes -- but if that were the case then this additional information that you are adding wouldn't be present anyway.  Or am I really misunderstanding something?

portante:
> Isn't all indexing being performed in a controlled process on the pbench server?

Good question.

It is a controlled process on the pbench server.  But when one queries for results it is often not at the time that the results were indexed.

So let's just consider run documents, since there is just one per tar ball, where a run document describes all the information about a pbench result tar ball.  The `@timestamp` field in the JSON document for a run uses the time at which the pbench-agent recorded the start of the benchmark; let's call that time `t0`.  The time at which that pbench result was indexed is at time `t1`.  And finally the time at which a query is done which pulls that JSON document is at time `t2`.

At time `t2`, if something does not look right with the JSON data, how would we determine what entity with what version of the software indexed the run JSON document (and by indexed we mean built the JSON document from the result tar ball and then sent it to Elasticsearch for indexing)?

The JSON document is indexed using an `@timestamp` of `t0`, not `t1`, since we want people to look up their data based on when they ran a particular benchmark.

Now on the pbench server, we have some operating system provided data that records what version of a pbench-server RPM was installed at what time, but because we don't store the time at which the JSON document was indexed, `t1`, one has to look at other logs to find out at what time that document *could* have been indexed so that we could correlate that time with the version of the software that was installed.

This process of correlation is tedious at best, and error prone, because in some cases the logs for installs are not complete, and the logs for the indexing steps are not complete, and/or there are errors in the logs.

This commit provides a separate JSON document that describes all the information about the entity that is executing in the `server-reports` index, and that ID is then attached to all JSON documents that are indexed.  This allows us to get that ID from a suspect document, and track down the exact information on who did what, when & how.

> It almost reads like we have results being indexed from unknown locations (servers?) by unknown processes -- but if that were the case then this additional information that you are adding wouldn't be present anyway.

Right, if we see data in an index that *doesn't* have the `@generated-by` ID, then we know it is not from our environment.  If it does have an `@generated-by` ID, and we can't find that ID, then we know something is broken with this indexing process.  If we *can* find that ID, then we can track down the exact time that entity changed its software version, on what host, from what user, etc.

riya-17:
@portante  it looks good to me

PR #1310, opened 2019-08-15
pbench-*-tools: fixup some iteration related issues

k-rister:
- First, the -i option is documented but was not actually processed
  (--iteration was).

- Second, when killing the tools for a tool group the iteration number
  is not relevant so remove it from the equation.  We always want the
  tools from the specified tool group to be killed no matter what the
  current iteration is.  This also makes iteration numbering schemes
  irrelevant when killing tools.


portante:
@atheurer, any love for this PR?

PR #1137, opened 2019-03-14
Initial dashboard backend GraphQL infrastructure

gurbirkalsi:
Integrates Prisma as a GraphQL API server in order to
provide backend functionality through defined dashboard
datamodels.

PR #1316, opened 2019-08-28
Update README for using dashboard playbooks

portante:
We also add a convenience link to the `config.json.j2` to make it easier to use the scripts, and add an ignore the generated `dist` directories.

Further, we now ignore patch files from git format-patch, and ignore `run-unittests.log` files.

### DIFF
--- a/contrib/ansible/dashboard/README.md
+++ b/contrib/ansible/dashboard/README.md
@@ -1,15 +1,24 @@
-## Ansible playbook for pbench dashboard
-This will ease installation, using pbench dashboard.
+## Ansible playbooks for installing and deploying the pbench dashboard
+This will ease installation, and deployment of the pbench dashboard.
 
 ## Required
 - Ansible needs to be installed on the host where you want to run this playbook
-- Python
-- JSON file containing the following key values: "elasticsearch", "production", "run_index", "prefix"
+- An inventory file containing the following key values defined:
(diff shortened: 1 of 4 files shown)


contrib/ansible/dashboard/README.md
```diff
@@ -1,15 +1,23 @@
...
-This will ease installation, using pbench dashboard.
+## Ansible playbooks for installing and deploying the pbench dashboard
+This will ease installation, and deployment of the pbench dashboard.
 
 ## Required
 - Ansible needs to be installed on the host where you want to run this playbook
-- Python
-- JSON file containing the following key values: "elasticsearch", "production", "run_index", "prefix"
+- An inventory file containing the following key values defined:
+  - "`elasticsearch`", "`production`", "`run_index`", "`prefix`"
```
gurbirkalsi: We should also include the new key called `graphql` for referencing the Prisma environment deployed along with pbench dashboard.

gurbirkalsi: I don't believe we included this key in `config.js.j2` as part of PR #1137 but we should consider including it with this PR.

portante: Okay, I'll fix that, thanks

portante: Fixed.

PR #1319, opened 2019-08-28
Updates to support uperf in run-benchmark

atheurer:
-support clients and servers in pbench-run-benchmark
-new uperf scripts: xml generator, client/server wrappers
 and postprocess script
-uperf.json in ./config/benchmark
-updates in pbench-output-monitor to support killing
 remote server processes
-updates in PbenchAnsible.pm to support ssh to endpoints
 with different directories

### DIFF
--- a/agent/bench-scripts/pbench-run-benchmark-sample
+++ b/agent/bench-scripts/pbench-run-benchmark-sample
@@ -19,7 +19,9 @@ use PbenchBase       qw(get_json_file put_json_file get_benchmark_names get_clie
                         get_pbench_install_dir );
 use PbenchAnsible    qw(ssh_hosts ping_hosts copy_files_to_hosts copy_files_from_hosts
                         remove_files_from_hosts remove_dir_from_hosts create_dir_hosts
-                        sync_dir_from_hosts verify_success yum_install_hosts);
+                        sync_dir_from_hosts verify_success yum_install_hosts
+                        copy_files_to_hostdirs create_hostdirs ssh_hostdirs
+                        sync_from_hostdirs remove_from_hostdirs kill_ssh);
 use PbenchCDM        qw(create_bench_iter_sample_doc);
 
 my $pbench_bench_config_dir = get_pbench_bench_config_dir;
@@ -59,12 +61,21 @@ if ($sample_dir =~ /.+([0-9])+$/) {
 }
 $sample{'sample'}{'num'} = $sample_num;
 
-# build a hash of clients and servers to be used in almost all the steps bewlow
+# build a hash of clients and servers to be used in almost all the steps below
 my %hosts;
 for my $type ("client", "server") {
+    my $count = 0;
     if (exists $iteration{'run'}{'bench'}{$type . 's'} and $iteration{'run'}{'bench'}{$type . 's'}) {
-        my @hosts = split(/ /, $iteration{'run'}{'bench'}{$type . 's'});
-        $hosts{$type} = \@hosts;
+        my @host_records;
+        for my $host (split(/\s+/, $iteration{'run'}{'bench'}{$type . 's'})) {
+            my %record = ('hostname' => $host,
+                          'type' => $type,
+                          'count' => $count,
+                          'dir' => $sample_dir . "-" . $type . $count );
+            push(@host_records, \%record);
+            $count++;
+        }
+        $hosts{$type} = \@host_records;
     }
 }
 
@@ -78,16 +89,11 @@ if (! $pp_only) {
     # pre-benchmark-execution:
     # - verify access to clients and servers by attempting to create the directories
     for my $type (keys %hosts) {
-        if (scalar @{ $hosts{$type} }) {
-            if ($specs{$type} and ${specs}{$type}{"copy-to"}) {
-                my $result;
-                my $rc;
-                my @files = @{ ${specs}{$type}{"copy-to"} };
-                # the end of the dir name gets "-client" or "-server" just in case this host is
-                # also a controller or both client and server
-                $result = create_dir_hosts(\@{ $hosts{$type} }, $sample_dir . "-" .  $type,
-                                           $base_dir);
-            }
+        if (scalar @{ $hosts{$type} }) { # the list of hosts for client or server
+            my $result;
+            # the end of the dir name gets "-client" or "-server" just in case this host is
+            # also a controller or both client and server
+            $result = create_hostdirs(\@{ $hosts{$type} }, $base_dir);
         }
     }
     # - verify required software on clients and server (todo)
@@ -110,7 +116,8 @@ if (! $pp_only) {
         my $rc = $?;
         open(my $fh, ">" . $sample_dir . "/preprocess-output.txt");
         if ($rc != 0) {
-            printf "Controller pre-script failed, exiting\n";
+            printf "Controller pre-script failed, output:\n\n";
+            print $preprocess_cmd . "\n";
             exit 1;
         }
     }
@@ -123,16 +130,15 @@ if (! $pp_only) {
                 my @files = @{ ${specs}{$type}{"copy-to"} };
                 # the end of the dir name gets "-client" or "-server" just in case this host is
                 # also a controller or both client and server
-                $result = copy_files_to_hosts(\@{ $hosts{$type} }, \@files, $sample_dir . "-" .
-                                              $type, $base_dir);
+                $result = copy_files_to_hostdirs(\@{ $hosts{$type} }, \@files, $base_dir);
             }
-	    if ($specs{$type} and ${specs}{$type}{"bin"}) {
-		# since there is an executable that is going to be used lets copy over the output
-		# wrapper in advance
-		my $result;
-		my @files = ( $local_output_wrapper_path ."/" . $output_wrapper_name );
-		$result = copy_files_to_hosts(\@{ $hosts{$type} }, \@files, $remote_output_wrapper_path, $base_dir);
-	    }
+	        if ($specs{$type} and ${specs}{$type}{"bin"}) {
+		        # since there is an executable that is going to be used lets copy over the output
+		        # wrapper in advance
+		        my $result;
+		        my @files = ( $local_output_wrapper_path . "/" . $output_wrapper_name );
+		        $result = copy_files_to_hosts(\@{ $hosts{$type} }, \@files, $remote_output_wrapper_path, $base_dir);
+	        }
         }
     }
     # - execute pre-benchmark scripts on clients and servers (todo)
@@ -143,8 +149,10 @@ if (! $pp_only) {
             
     # benchmark-execution:
     print "starting benchmark\n";
+    my $server_ssh_logdir; # Saving this logdir, in order to kill this stuff after the test
     for my $type (reverse sort keys %hosts) { # reverse so "server" is first
         if ($hosts{$type} and scalar @{ $hosts{$type} }) {
+            printf "launching %s(s)\n", $type;
             my $cmd;
             if ($specs{$type}{"bin"}) {
                 $cmd = $remote_output_wrapper_path . "/" . $output_wrapper_name. " " .
@@ -164,12 +172,27 @@ if (! $pp_only) {
                     $cmd = eval "\$_='$cmd'; $r; return scalar \$_";
                 }
             }
-            my $result = ssh_hosts(\@{ $hosts{$type} }, $cmd, $sample_dir . "-" . $type, $base_dir);
+            if ($type eq "server") {
+                # Servers need to be backgrounded so we can proceed to running
+                # the clients.
+                $server_ssh_logdir = ssh_hostdirs(\@{ $hosts{$type} }, $cmd, $base_dir, 1);
+            } else {
+                ssh_hostdirs(\@{ $hosts{$type} }, $cmd, $base_dir, 0);
+            }
+            if ($type eq 'server') {
+                my $seconds = scalar @{ $hosts{$type} };
+                printf "Waiting %d seconds for servers to be ready\n", $seconds;
+                sleep($seconds);
+            }
         }
+    }
+    if (defined $server_ssh_logdir) {
+        kill_ssh($server_ssh_logdir); # Servers can now be shut down
     }
 
     # stop the tools:
     print "stopping tools\n";
+    print "pbench-stop-tools --iteration=0 --group=" . $tool_group . " --dir=" . $sample_dir . "\n";
     system("pbench-stop-tools --iteration=0 --group=" . $tool_group . " --dir=" . $sample_dir);
 
     # post-process the tools:
@@ -182,14 +205,10 @@ if (! $pp_only) {
     for my $type (keys %hosts) {
         if ($hosts{$type} and scalar @{ $hosts{$type} }) {
             mkdir($sample_dir . "/" . $type . "s");
-            my $sync_result = sync_dir_from_hosts(\@{ $hosts{$type} },
-                                                  $sample_dir . "-" . $type . "/", #copy-from
+            my $sync_result = sync_from_hostdirs(\@{ $hosts{$type} },
                                                   $sample_dir . "/" . $type . "s", #copy-to
                                                   $base_dir);
-                                                  # adding the "/" ensures the base dir
-                                                  # name is not copied
-            my $remove_result = remove_dir_from_hosts(\@{ $hosts{$type} },
-                                                      $sample_dir . "-" . $type,
+            my $remove_result = remove_from_hostdirs(\@{ $hosts{$type} },
                                                       $base_dir);
         }
     }
@@ -221,7 +240,8 @@ if (exists $specs{"controller"}{"post-script"} and $specs{"controller"}{"post-sc
     open(my $fh_output, ">" . $sample_dir . "/postprocess-output.txt");
     printf { $fh_output } "%s\n", $postprocess_output;
     if ($rc != 0) {
-        printf "Controller post-script failed, exiting\n";
+        printf "Controller post-script failed, output below:\n\n";
+        print $postprocess_output . "\n";
         exit 1;
     }
 } else {
--- a/agent/bench-scripts/postprocess/uperf-client-launcher
+++ b/agent/bench-scripts/postprocess/uperf-client-launcher
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+# Include uperf envronment variables so it is clear what uperf options were used
+uperf_vars=""
+while [ ! -z "$1" ]; do
+	arg=`echo $1 | awk -F= '{print $1}' | sed -e s/^--// -e 's/-/_/g'`
+	val=`echo $1 | awk -F= '{print $2}'`
+	if [ $arg == "message_size" ]; then
+		uperf_vars="$uperf_vars wsize=$val rsize=$val"
+	elif [ $arg == "remotehost" ]; then
+		let field=$pbench_count+1
+		remotehost=`echo $val | cut -d, -f$field`
+		uperf_vars="$uperf_vars remotehost=$remotehost"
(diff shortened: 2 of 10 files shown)


agent/bench-scripts/postprocess/uperf-prepare-xmlfile
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: t; sh-basic-offset: 8; sh-indentation: 8; sh-indent-for-case-alt: + -*-
+#echo "<profile name=\"protocol:\$protocol-test_type:\$test_type-wsize:\$wsize-threads:\$threads\">"
```
portante: Is this for debugging?

agent/bench-scripts/postprocess/uperf-prepare-xmlfile
```diff
@@ -0,0 +1,56 @@
...
+
+echo "$@" >pre-script-args.txt
+test_type=stream
+while [ ! -z $1 ]; do
+	arg=`echo $1 | awk -F= '{print $1}'`
+	if [ $arg == "--test-type" ]; then
+		test_type=`echo $1 | awk -F= '{print $2}'`
```
portante: Why isn't this: `echo $2 | ...`?

agent/bench-scripts/postprocess/uperf-prepare-xmlfile
```diff
@@ -0,0 +1,56 @@
...
+		break
+	fi
+	shift
+done
+	
+file=$pbench_tmp/uperf.xml
```
portante: Why hard code the file here instead of just printer to stdout and letting the caller redirect it to where it wants it?

agent/bench-scripts/postprocess/uperf-prepare-xmlfile
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: t; sh-basic-offset: 8; sh-indentation: 8; sh-indent-for-case-alt: + -*-
+#echo "<profile name=\"protocol:\$protocol-test_type:\$test_type-wsize:\$wsize-threads:\$threads\">"
+
+echo "$@" >pre-script-args.txt
```
portante: Is this for debugging?  We only seem to look for `--test_type` and ignore the rest.

agent/bench-scripts/postprocess/uperf-prepare-xmlfile
```diff
@@ -0,0 +1,56 @@
...
+	echo '    <transaction duration="$runtime">' >>$file
+	echo '      <flowop type="read" options="count=16 size=$rsize"/>' >>$file
+	echo '    </transaction>' >>$file
+	echo '    <transaction iterations="1">' >>$file
+	echo '      <flowop type="disconnect"/>' >>$file
+	echo '    </transaction>' >>$file
+	;;
+esac
+echo '  </group>' >>$file
+echo '</profile>' >>$file
```
portante: Seems like we can remove some redundancy here, fixing a missing `</transactions>` in the mean time, perhaps consider:

```
echo '<?xml version="1.0"?>' >$file
echo '<profile name="netperf">' >>$file
echo '  <group nthreads="$instances">' >>$file
case $test_type in
    *)
        flow_type="connect"
	;;
	maerts|bidirec)
        flow_type="accept"
	;;
esac
echo '    <transaction iterations="1">' >>$file
echo "      <flowop type=\"${flow_type}\" options=\"remotehost=\$remotehost protocol=\$protocol\"/>" >>$file
echo '    </transaction>' >>$file
echo '    <transaction duration="$runtime">' >>$file
case $test_type in
	rr)
	echo '      <flowop type="write" options="size=$wsize"/>' >>$file
	echo '      <flowop type="read"  options="size=$rsize"/>' >>$file
	;;
	stream|bidirec)
	echo '      <flowop type="write" options="count=16 size=$wsize"/>' >>$file
	;;
	maerts|bidirec)
	echo '      <flowop type="read" options="count=16 size=$rsize"/>' >>$file
	;;
esac
echo '    </transaction>' >>$file
echo '    <transaction iterations="1">' >>$file
echo '      <flowop type="disconnect"/>' >>$file
echo '    </transaction>' >>$file
echo '  </group>' >>$file
echo '</profile>' >>$file
```

agent/bench-scripts/pbench-run-benchmark-sample
```diff
@@ -123,16 +129,15 @@ if (! $pp_only) {
...
+                $result = copy_files_to_hostdirs(\@{ $hosts{$type} }, \@files, $base_dir);
             }
-	    if ($specs{$type} and ${specs}{$type}{"bin"}) {
-		# since there is an executable that is going to be used lets copy over the output
-		# wrapper in advance
-		my $result;
-		my @files = ( $local_output_wrapper_path ."/" . $output_wrapper_name );
-		$result = copy_files_to_hosts(\@{ $hosts{$type} }, \@files, $remote_output_wrapper_path, $base_dir);
-	    }
+	        if ($specs{$type} and ${specs}{$type}{"bin"}) {
```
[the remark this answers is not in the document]
portante: The whitespace formatting still doesn't look quite right, does it?

agent/bench-scripts/pbench-run-benchmark-sample
```diff
@@ -164,12 +171,27 @@ if (! $pp_only) {
...
+                # Servers need to be backgrounded so we can proceed to running
+                # the clients.
+                $server_ssh_logdir = ssh_hostdirs(\@{ $hosts{$type} }, $cmd, $base_dir, 1);
+            } else {
+                ssh_hostdirs(\@{ $hosts{$type} }, $cmd, $base_dir, 0);
+            }
+            if ($type eq 'server') {
+                my $seconds = scalar @{ $hosts{$type} };
+                printf "Waiting %d seconds for servers to be ready\n", $seconds;
+                sleep($seconds);
```
portante: This seems really fragile.  Can we use a redis server here to coordinate servers and clients a little cleaner?

agent/config/benchmark/uperf.json
```diff
@@ -0,0 +1,71 @@
...
+    "post-script" : " ",
+    "copy-to" : [ "/var/lib/pbench-agent/tmp/uperf.xml" ],
+    "bin" : "/opt/pbench-agent/bench-scripts/postprocess/uperf-client-launcher",
+    "param_regex" : []
+  },
+  "server" : { 
+    "packages" : [ "pbench-uperf-1.0.4-1" ],
+    "pre-script" : " ",
+    "post-script" : " ",
+    "bin" : "/opt/pbench-agent/bench-scripts/postprocess/uperf-server-launcher",
```
portante: Can we just assume all `bin` keys have values that are relative to the `pbench_install_dir` instead of hard-coding the paths?  Same for the `pre|post-script` keys above?

agent/config/benchmark/uperf.json
```diff
@@ -0,0 +1,71 @@
...
+        "positive_integer" : {
+          "description" : "a whole number greater than 0",
+          "arguments" : [ "instances" ],
+          "value_regex" : "[1-9][0-9]*"
+          }
+        }
+      }
+    },
+  "client" : { 
+    "packages" : [ "pbench-uperf-1.0.4-1" ],
```
portante: Wouldn't we want the client and server to alway share the same package version?  If not, then let's keep it that way.  But if they alway should, seems like `packages` should be specified once to avoid mistakes in the JSON, and shared by `client` and `server`.

agent/config/benchmark/uperf.json
```diff
@@ -0,0 +1,71 @@
...
+    "pre-script" : " ",
+    "post-script" : " ",
+    "copy-to" : [ "/var/lib/pbench-agent/tmp/uperf.xml" ],
```
portante: Why do we need to encode the `copy-to` location to be a temporary directory?  This seems like an execution time value that shouldn't be in the JSON, but not sure about it.

agent/config/benchmark/uperf.json
k-rister: I think we had a vision here where it would be possible for users to provide their own benchmark configuration files (ie. before something is upstream in pbench) so being able to reference ad hoc locations would be required.  In extreme cases this would allow someone to provide an implementation of a benchmark that we don't support ourselves, but in other cases it may just streamline the process of updating what we do support.

For example, for trafficgen we would do a normal (fully integrated) implementation for pbench-run-benchmark in pbench.  In parallel, we would have the source of those files included in our actual trafficgen repository (ie. the config JSON, the postprocessor, etc.).  When we update trafficgen (usually asynchronously to pbench updates) we would update the files in the trafficgen repo so that a user could take advantage of whatever the changes are prior to getting an updated version of pbench.  Once those updates are in trafficgen they would be synchronized (most likely with path updates) into pbench for a future pbench release.

If we follow this model it pushes the maintenance of these files (which are heavily tied to the benchmark) into the benchmark itself.

agent/config/benchmark/uperf.json
k-rister: I may be speaking for Andrew here, but I think the point is to move all benchmark specific logic like this into the JSON configuration file so that the generic scripts (pbench-run-benchmark, pbench-gen-iterations, pbench-run-benchmark-sample) can just use these well know field names to access the necessary information to perform well known steps like copying a config file to a client system.

There is very little (if any) execution time scripting that would be benchmark specific and be able to handle these types of things for us.  The major exception to that is postprocessing which obviously tends to be very specific to the various output(s) that a benchmark produces.

agent/bench-scripts/postprocess/uperf-server-launcher
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+echo "This is the uperf server launcher"
+echo "args: $@"
+
+port=`echo "20000 + 3 * $pbench_count" | bc`
+#existing_pid=`
```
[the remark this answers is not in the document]
atheurer: Yes, the #existing_pid needs to go

agent/bench-scripts/postprocess/uperf-server-launcher
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+echo "This is the uperf server launcher"
+echo "args: $@"
```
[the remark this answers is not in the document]
atheurer: This is just debug code that we can take out

agent/bench-scripts/postprocess/uperf-client-launcher
```diff
@@ -0,0 +1,27 @@
...
+	else
+		uperf_vars="$uperf_vars $arg=$val"
+	fi
+	shift
+done
+uperf_vars=`echo $uperf_vars | sed -e 's/^ //'`
+port=`echo "20000 + 3 * $pbench_count" | bc`
+echo "uperf_vars: [$uperf_vars]"
+printf "%s\n" "#!/bin/bash" >uperf-client.cmd
+printf "%s %s\n" "$uperf_vars" "/usr/local/bin/uperf -v -m ./uperf.xml -x -a -i 1 -P $port >uperf-client-stdout.txt 2>uperf-client-stderr.txt" >>uperf-client.cmd
```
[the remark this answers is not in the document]
atheurer: I tend to use 1 printf or echo per line of output, but I suppose I can consolidate it.

agent/bench-scripts/postprocess/uperf-server-launcher
```diff
@@ -0,0 +1,12 @@
...
+
+port=`echo "20000 + 3 * $pbench_count" | bc`
+#existing_pid=`
+echo "/usr/local/bin/uperf -v -s -P $port >server-stdout.txt 2>server-stderr.txt" >server.cmd
+echo "exit $?" >>server.cmd
+chmod +x server.cmd
+./server.cmd & echo "$!" >server.pid
```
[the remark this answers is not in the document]
atheurer: The point of the current method is to allow someone to re-run server.cmd if they need to debug this.  We do this a lot in pbench now, and I was continuing that method.  If we don't want to continue that method, we can do what you describe.  The capture of the server pid is actually no longer necessary anyway, and we can remove that.

agent/bench-scripts/postprocess/uperf-client-launcher
```diff
@@ -0,0 +1,27 @@
...
+chmod +x ./uperf-client.cmd
+./uperf-client.cmd
```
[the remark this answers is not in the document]
atheurer: If we don't use this method, we can't easily track what variables were used that uperf.xml needs.  So, if we get an error like "malformed xml", it's not obvious what environment variables were missing that caused the malformed xml.

agent/bench-scripts/pbench-run-benchmark-sample
atheurer: I suppose one could add a more robust mechanism, but I think that is an enhancement for a later PR.

agent/bench-scripts/postprocess/uperf-prepare-xmlfile
atheurer: No, it's just old code that needs to be removed.

agent/bench-scripts/postprocess/uperf-prepare-xmlfile
atheurer: Why would we assume all controller pre-scripts' output be redirected to a file in a specific place?  The caller is pbench-run-benchmark-sample -if we want per-benchmark control over where this files gets redirected to, now we have to add even more configuration in the benchmark.json so pbench-run-benchmark-sample knows where to redirect this.  Then , what if a benchmark needs the pre-script to generate 2 files, or 3?

agent/bench-scripts/postprocess/uperf-prepare-xmlfile
atheurer: Yes, it can be removed

agent/bench-scripts/postprocess/uperf-prepare-xmlfile
atheurer: This might change the way bidirec works, and I would prefer to optimize this later once we know we have the everything else working fine.

agent/config/benchmark/uperf.json
atheurer: It cannot be relative to pbench_install_dir because the "bin" could be literally anything, and is sometimes the actual benchmark binary itself, like fio.  If a benchmark absolutely has to have a wrapper script for the client or server, it will likely reside somewhere under pbench_install_dir.  To fix absolute paths that could eventually be wrong (like pbench_install_dir later changes to somethign other than /opt/pbenc-agent), we have a issue (https://github.com/distributed-system-analysis/pbench/issues/1320) in the run-benchmark project to allow the use of common pbench vars in the benchmark.json, so, for the uperf example, one could use: bin: $pbench_install_dir/postprocess/uperf-server-launcher.  For fio, since it does not use a wrapper script, should have either bin: "fio" or bin: "/usr/local/bin/fio" if we are certain that pbench-fio installed to /usr/local/bin/

agent/config/benchmark/uperf.json
atheurer: Yes, packages could go to a non-server non-client part of the json.  I think I added it per server and client "just in case", but it that looks like we'll never need that, we can eliminate it.

agent/config/benchmark/uperf.json
atheurer: The copy-to: is actually specifying the *source* of the files that need to be copied, which could be from anywhere.  See (https://github.com/distributed-system-analysis/pbench/issues/1320) for not hardcoding things like /var/lib/pbench-agent/tmp/ and using a pbench_var.  

The destination for copy-to is fixed and the remote sample directory for each client.

agent/bench-scripts/postprocess/uperf-server-launcher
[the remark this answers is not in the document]
k-rister: I would prefer to not get rid of the ability to easily/quickly relaunch the commands when doing deep dive analysis/debug.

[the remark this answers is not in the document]
portante: I can understand the need to capture the command that is run, but since the command is only different here by `$port` it seems it is a bit of an overkill just to rerun the command given all the mechanism involved to do so (create a file, make sure it exits with a status value (which it would by default I believe), and then make it executable, and then run it).  Validating the arguments and then executing the command with an `exec` seems easy enough.

Regardless, it would be great to make sure we check the arguments properly.

agent/bench-scripts/postprocess/uperf-client-launcher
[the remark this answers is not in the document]
portante: Recording the arguments to the `uperf-client-launcher` would be the same, which you already emit when the script first starts?

Over all the mechanism seems fragile to pass arguments, rebundle them a new way to then pass them along again.

No need to hold this up on this discussion.

agent/config/benchmark/uperf.json
portante: So then this begs the question of the location of the "launchers".  Seems like they are specific to uperf, so perhaps they should be in a separate location from the `post-process` directory to indicate they are uperf specific?  Perhaps a "contrib" area?

Perhaps we should have the `pbench-uperf` package provide those interfaces as an alternative so that the pbench-agent does not contain any benchmark specific mechanisms for `run-benchmark`.

In general this seems confusing to me and the interfaces seem tied directly to the implementation of `uperf` and `fio`.  Perhaps if we had an example "generic" benchmark to highlight the pbench-run-benchmark related interfaces separate from those of `uperf` and `fio` implementations that would make this clearer?

agent/config/benchmark/uperf.json
portante: Can you explain more?  If the destination is fixed, why does it need to be in the JSON specification?  What entity consumes the value of `copy-to`?

atheurer: The destination is fixed, but the names of the files to "copy-to" the clients/servers are not.  This is where a benchmark specifies those files.  Each benchmark may want different files copied from the agent to the clients/servers (fio.job for fio, uperf.xml for uperf, etc)  It is up to pbench-run-benchmark-sample to provide this functionality to the benchmark, as long as the benchmark json provides the source files it needs.  Please see https://github.com/distributed-system-analysis/pbench/blob/5f9773d83e05968e64d93aec7ea14385fde07f1b/agent/bench-scripts/pbench-run-benchmark-sample#L123

agent/config/benchmark/uperf.json
atheurer: Contrib in my mind would indicate this is optional, which it is not.  We don't keep $benchmark-postprocess files with pbench-$benchmark.RPM, so it does not make sense to include other benchmark helper scripts in those RPMs.  Those RPMs IMO are for the underlying benchmark only and nothing else.  Placing these scripts in the postprocess directory is probably not the most ideal place, but it was a convenient one.  They should go in a pbench directory that is not in the path (not bench-scripts and not tool-scripts), but should be part of pbench-agent because they are essential to running the benchmark.  Perhaps they should go in $pbench_install_dir/util-scripts/.

If we were following a more conforming installation, we would probably be placing almost all of this in /usr/libexec, and have the [eventually with pbench-run-benchmark] very few user-runnable scripts in /usr/bin/

k-rister: I'm going to work on implementing some other simple workloads this week just to get some more stuff in there.  I feel like we tackled three of the more complicated workloads (fio, uperf, trafficgen) first which was really good for making sure we covered the various nuances of a complicated workload but it does seem like it has made the barrier to entry, or understanding, a bit high.

Right now I'm thinking things like kernbench and some RT workloads which are pretty simplistic in nature, although the devil is in the details as usual.

atheurer:
I am putting this work on pause and therefore closing this PR

portante:
Just because the work is on pause, why close it?  Are you not intending to submit this work at all?

PR #1322, opened 2019-08-29
Pbench base

atheurer:
(this shows 2 commits, but first one is for uperf-run-benchmark, since it is not in master yet)

Eliminates need to source base script before running pbench-run-benchmark

### DIFF
--- a/agent/bench-scripts/pbench-gen-iterations
+++ b/agent/bench-scripts/pbench-gen-iterations
@@ -80,13 +80,16 @@
 
 use strict;
 use warnings;
-use lib $ENV{'pbench_lib_dir'};
+my $pbench_install_dir;
+BEGIN {
+    chomp($pbench_install_dir = `getconf.py pbench_install_dir pbench-agent`);
+    }
+use lib $pbench_install_dir . "/lib";
 use Text::ParseWords;
 use JSON;
 use Data::Dumper;
-use PbenchBase qw(get_json_file get_benchmark_names
-                  get_pbench_bench_config_dir load_benchmark_config);
-my $pbench_bench_config_dir = get_pbench_bench_config_dir;
+use PbenchBase qw(get_json_file get_benchmark_names load_benchmark_config);
+my $pbench_bench_config_dir = $pbench_install_dir . "/config/benchmark";
 
 # Return an array of all the valid arguments for a specific benchmark
 sub get_benchmark_args {
--- a/agent/bench-scripts/pbench-run-benchmark
+++ b/agent/bench-scripts/pbench-run-benchmark
@@ -15,14 +15,16 @@
 use strict;
 use warnings;
 use File::Basename;
-my $pbench_install_dir = $ENV{'pbench_install_dir'};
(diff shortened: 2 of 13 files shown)


portante:
Doesn't this add more called to `getconf.py` instead of just having one in `base`?  This then spreads out code changes to many more modules for removing this later.

Could we stick will sourcing `base` for now and then solve later by not pushing `getconf.py` calls to so many more places?

atheurer:
> Doesn't this add more called to `getconf.py` instead of just having one in `base`? This then spreads out code changes to many more modules for removing this later.
> 
> Could we stick will sourcing `base` for now and then solve later by not pushing `getconf.py` calls to so many more places?

I did not realize making calls to getconf.py was a problem.  I assumed any of our scripts were welcome to do so.  There are other scripts which do this, like pbench-fio, pbench-register-tool-set, pbench-sysinfo-dump, pbench-agent-config-activate, pbench-edit-prefix, pbench-move-results, pbench-agent-config-ssh-key, and pbench-make-result-tb.  Are you inferring that we should get rid of all these calls as well?  If so, I don't understand the motivation.  I would prefer we make calls to getconf.py for exactly what we need instead of calling a base script which may do a lot of work which we don't need, depending on what script is calling it.  If we really do want to get rid of using getconf.py, can we just get a standard /etc/sysconfig/pbench-agent file like many other programs have?

k-rister:
I second Andrew's questions.

At the very least, don't we need language specific alternatives for sourcing `base` to cover the Python and Perl stuff?

atheurer:
The only alternative I can see is sourcing the base script with backticks in perl, combined with a call to "env", then capture that output.  We can then define any environment variables we see in perl that start with "pbench_".

atheurer:
Going to close this in favor of another method

portante:
FWIW, I'd love to see a PR to switch all the environment variables currently provided by `base` moved from the pbench-agent.cfg to `/etc/sysconfig/pbench-agent` file.  That would be a great first step.

If somebody wants to do a PR to remove `getconf.py` and the agent configuration file replacing it entirely with a fully populated `/etc/sysconfig/pbench-agent` file, that would be welcome, too, but that is very large undertaking, and there are conversion issues from user's existing `pbench-agent.cfg` files that would need to be handled during the deployment.

portante:
Not to try to argue a point further, but to record the thinking and current state of what we doing today ...

See also #1204, where we changed most of the perl modules to use the `pbench_install_dir` environment variable.

Looking at `base` it fetches 5 common pieces of data from the config file:

 * `pbench_run` === `$(getconf.py pbench_run pbench-agent)`
 * `pbench_log` === `$(getconf.py pbench_log pbench-agent)`
 * `pbench_install_dir` === `$(getconf.py install-dir pbench-agent)`
 * `ssh_opts` === `$(getconf.py ssh_opts results)`
 * `scp_opts` === `$(getconf.py scp_opts results)`

All five of those pieces of data are used very often throughout the life of a benchmark run, so having them fetched once in `base` is a big win.

PR #1324, opened 2019-08-29
Fix mocking of python's time.time() method

portante:
The method of mocking out various methods and classes for running unit tests for the python modules is a bit fragile.  Recently, we deployed the latest code changes before this commit to a staging server and nothing worked.  This was due to the fact that, as is expected when not mocking properly for unit tests, the production code paths were incorrect.

Most of these changes are cleanups from when we split out the `indexer.py` module from the `pbench-index.py` command.

Here is a short list:

  * Modify pbench-base.sh to mock a time stamp using 1970 instead of 1900 to be consistent with seconds-since-the-epoch

  * Moved mocking of the `time.time` method to one place, the `__init__.py` module, where we now import `time.time` as `_time`, so that `PbenchConfig` is now the one place time is mocked by replacing the `_time` method; to make this work, all other references to `time.time` are replaced with references to `pbench._time`; this is subtle in that we have to `import pbench` in other modules so that we reference the `_time` method that gets mocked

  * We remove the hard-coded mock'd timestamp strings where possible

  * Update the `pbench.tstos()` method to format timestamps using the `UTC` timezone string in the output, and invoking the `_time()` method if no timestamp is given to pick up a mocked timestamp, while using the same code path between production and unit tests

  * Add an import of the `logging` module to `lib/bin/indexer.py`

  * Moved the `_calc_backoff_sleep` method from `__init__.py` to the `indexer.py` module where it belongs, adding a `_sleep_w_backoff()` method to simplify the code and provide for a `_sleep` method that can be mocked properly in the future

  * Import the `pbench` module as a whole in `report.py` to be consistent with the way the module is imported by the `indexer.py` module


portante:
@ndokos, rebased on top of the other merges, resolving conflicts, and then updated the commit message to reflect your edits above.  I'll merge once all the TravisCI tests pass.

PR #1327, opened 2019-09-03
Process "base" file in pbench-run-benchmark & co. scripts

atheurer:
-Use a BEGIN code block to get pbench environment variables
 before loading pbench-provided perl modules
-Calls to other pbench scripts are prepended with sourcing "base"

### DIFF
--- a/agent/bench-scripts/pbench-gen-iterations
+++ b/agent/bench-scripts/pbench-gen-iterations
@@ -80,13 +80,20 @@
 
 use strict;
 use warnings;
-use lib $ENV{'pbench_lib_dir'};
+use File::Basename;
+my %pbench_env;
+BEGIN {
+    chomp(my $script_path = dirname($0));
+    chomp(my $dir = `cd ${script_path}/..; /bin/pwd`);
+    foreach (grep(/^pbench\S+=/, `. $dir/base; env`)) {
+    $pbench_env{$1} = $2 if (/^pbench_(\S+)=(.+)/);
+    }
+}
+use lib $pbench_env{'lib_dir'};
 use Text::ParseWords;
 use JSON;
 use Data::Dumper;
-use PbenchBase qw(get_json_file get_benchmark_names
-                  get_pbench_bench_config_dir load_benchmark_config);
-my $pbench_bench_config_dir = get_pbench_bench_config_dir;
+use PbenchBase qw(get_json_file get_benchmark_names load_benchmark_config);
 
 # Return an array of all the valid arguments for a specific benchmark
 sub get_benchmark_args {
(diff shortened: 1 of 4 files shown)


agent/bench-scripts/pbench-gen-iterations
```diff
@@ -80,13 +80,20 @@
 
 use strict;
 use warnings;
-use lib $ENV{'pbench_lib_dir'};
+use File::Basename;
+my %pbench_env;
+BEGIN {
```
portante: Instead of having this `BEGIN` stanza copied four different times in four different files, why not just have `pbench-run-benchmark` be a simple `bash` script that first sources `base` like the rest of pbench, and then have it invoke `pbench-run-benchmark.pl`?  Then we don't have any duplicate code, and we avoid extra mechanism by making the environment variables always available.

And if we dropped all the `getconf.py` code, won't we have to setup the environment variables before executing a perl script anyways, which means we'll have this duplicate code everywhere still?

Seems like we all agree duplicate code is bad and can become a problem when it is updated in one place and not another?

atheurer: Why, because It's not just pbench-run-benchmark, we would need a wrapper script for every non-bash script that anyone might run at some point, including pbench-gen-iterations, pbench-run-benchmark-sample, and who knows what other non-bash scripts we might have now or in the future.  So how is creating at least three wrapper scripts better than having a BEGIN {... statement in the four perl scripts, which, BTW, is not all that different from every bash script we currently have which has:

<pre>
1 script_path=`dirname $0`
2 script_name=`basename $0`
4 pbench_bin="`cd ${script_path}/..; /bin/pwd`"
5
6 # source the base script
7 . "$pbench_bin"/base
</pre>

That is in **45** scripts in pbench.  Where is the outrage?  There isn't any because it's not a big deal, and neither is this in 4 scripts:

<pre>
1 BEGIN {
2    chomp(my $script_path = dirname($0));
3    chomp(my $dir = `cd ${script_path}/..; /bin/pwd`);
4    foreach (grep(/^pbench\S+=/, `. $dir/base; env`)) {
5    $pbench_env{$1} = $2 if (/^pbench_(\S+)=(.+)/);
6    }
7 }</pre>

But that's not all; each wrapper script has to have code in it to set up the environment, which is exactly the same as the top example here.  Then we have multiple wrapper scripts with... more duplicated code, so we're right back at square 1 with now having to maintain a bunch of wrapper scripts and still have duplicated 7 lines of code.

portante:
Why was this PR closed?  I have not had a chance to respond to your comments, which you have valid points.  Can we re-open so that we can work through this?

PR #1332, opened 2019-09-05
Centrally log pbench-server

### LINKED ISSUE #310: Centrally log pbench-server operations
Instead of only using local log files, consider sending (structured) logs to a central server.

riya-17:
Fixes #310

The PR is regarding sending the logs to central server instead of using local log files.
To store the logs on Local log files FileHandler was used. I replaced the File Handler with the sysLogHandler to store logs on server.
To store logs on server rsyslog is required. It forwards log messages in an IP network with the help of configuration file in which we can define the IP network of where we want to send the data.

### DIFF
--- a/server/bin/gold/test-26.3.txt
+++ b/server/bin/gold/test-26.3.txt
@@ -1,0 +1,112 @@
++++ Running test_logger_type.py
+BadConfig exception was raised
+--- Finished test_logger_type.py (status=1)
++++ Running unit test audit
+pbench-audit-server: No option 'logger_host' in section: 'logging' (config file /var/tmp/pbench-test-server/test-26.3/opt/pbench-server/lib/config/pbench-server.cfg)
+--- Finished unit test audit (status=1)
++++ var/www/html tree state (/var/tmp/pbench-test-server/test-26.3/var-www-html)
+lrwxrwxrwx         65 incoming -> /var/tmp/pbench-test-server/test-26.3/pbench/public_html/incoming
+drwxrwxr-x          - pbench-results-host-info.versioned
+lrwxrwxrwx         38 pbench-results-host-info.versioned/pbench-results-host-info.URL002 -> pbench-results-host-info.URL002.active
+-rw-rw-r--        120 pbench-results-host-info.versioned/pbench-results-host-info.URL002.active
+-rw-rw-r--         95 pbench-results-host-info.versioned/pbench-results-host-info.URL002.maint
+lrwxrwxrwx         64 results -> /var/tmp/pbench-test-server/test-26.3/pbench/public_html/results
+lrwxrwxrwx         63 static -> /var/tmp/pbench-test-server/test-26.3/pbench/public_html/static
+lrwxrwxrwx         62 users -> /var/tmp/pbench-test-server/test-26.3/pbench/public_html/users
+--- var/www/html tree state
++++ results host info (/var/tmp/pbench-test-server/test-26.3/var-www-html/pbench-results-host-info.versioned)
+/var/tmp/pbench-test-server/test-26.3/var-www-html/pbench-results-host-info.versioned/pbench-results-host-info.URL002.active:pbench@pbench.example.com:/var/tmp/pbench-test-server/test-26.3/pbench-local/pbench-move-results-receive/fs-version-002
+/var/tmp/pbench-test-server/test-26.3/var-www-html/pbench-results-host-info.versioned/pbench-results-host-info.URL002.maint:MESSAGE===System Under Maintenance - please retry at a later time (unit-test-user@example.com)
+--- results host info
++++ var/www/html-satellite tree state (/var/tmp/pbench-test-server/test-26.3/var-www-html-satellite)
+lrwxrwxrwx         75 incoming -> /var/tmp/pbench-test-server/test-26.3/pbench-satellite/public_html/incoming
+drwxrwxr-x          - pbench-results-host-info.versioned
+lrwxrwxrwx         38 pbench-results-host-info.versioned/pbench-results-host-info.URL002 -> pbench-results-host-info.URL002.active
+-rw-rw-r--        140 pbench-results-host-info.versioned/pbench-results-host-info.URL002.active
+-rw-rw-r--         95 pbench-results-host-info.versioned/pbench-results-host-info.URL002.maint
+lrwxrwxrwx         74 results -> /var/tmp/pbench-test-server/test-26.3/pbench-satellite/public_html/results
+lrwxrwxrwx         73 static -> /var/tmp/pbench-test-server/test-26.3/pbench-satellite/public_html/static
+lrwxrwxrwx         72 users -> /var/tmp/pbench-test-server/test-26.3/pbench-satellite/public_html/users
+--- var/www/html-satellite tree state
++++ results host info (/var/tmp/pbench-test-server/test-26.3/var-www-html-satellite/pbench-results-host-info.versioned)
+/var/tmp/pbench-test-server/test-26.3/var-www-html-satellite/pbench-results-host-info.versioned/pbench-results-host-info.URL002.active:pbench@pbench-satellite.example.com:/var/tmp/pbench-test-server/test-26.3/pbench-satellite-local/pbench-move-results-receive/fs-version-002
+/var/tmp/pbench-test-server/test-26.3/var-www-html-satellite/pbench-results-host-info.versioned/pbench-results-host-info.URL002.maint:MESSAGE===System Under Maintenance - please retry at a later time (unit-test-user@example.com)
+--- results host info
++++ pbench tree state (/var/tmp/pbench-test-server/test-26.3/pbench)
+drwxrwxr-x          - archive
+drwxrwxr-x          - archive/fs-version-001
+drwxrwxr-x          - public_html
+drwxrwxr-x          - public_html/incoming
+drwxrwxr-x          - public_html/results
+drwxrwxr-x          - public_html/static
+drwxrwxr-x          - public_html/static/css
+drwxrwxr-x          - public_html/static/css/v0.2
+drwxrwxr-x          - public_html/static/css/v0.2/css
+-rw-rw-r--        308 public_html/static/css/v0.2/css/pbench_utils.css
+drwxrwxr-x          - public_html/static/css/v0.3
+drwxrwxr-x          - public_html/static/css/v0.3/css
+-rw-rw-r--      11798 public_html/static/css/v0.3/css/LICENSE.TXT
+-rw-rw-r--       3663 public_html/static/css/v0.3/css/jschart.css
+drwxrwxr-x          - public_html/static/js
+drwxrwxr-x          - public_html/static/js/v0.2
+drwxrwxr-x          - public_html/static/js/v0.2/js
+-rw-rw-r--       9415 public_html/static/js/v0.2/js/app.js
+-rw-rw-r--       5556 public_html/static/js/v0.2/js/pbench_utils.js
+drwxrwxr-x          - public_html/static/js/v0.3
+drwxrwxr-x          - public_html/static/js/v0.3/js
+-rw-rw-r--      11798 public_html/static/js/v0.3/js/LICENSE.TXT
+-rw-rw-r--     143934 public_html/static/js/v0.3/js/jschart.js
+drwxrwxr-x          - public_html/users
+--- pbench tree state
++++ pbench-local tree state (/var/tmp/pbench-test-server/test-26.3/pbench-local)
+drwxrwxr-x          - logs
+drwxrwxr-x          - pbench-move-results-receive
+drwxrwxr-x          - pbench-move-results-receive/fs-version-002
+drwxrwxr-x          - quarantine
+drwxrwxr-x          - quarantine/duplicates-002
+drwxrwxr-x          - quarantine/errors-002
+drwxrwxr-x          - quarantine/md5-002
+drwxrwxr-x          - tmp
+--- pbench-local tree state
++++ pbench-satellite tree state (/var/tmp/pbench-test-server/test-26.3/pbench-satellite)
+drwxrwxr-x          - archive
+drwxrwxr-x          - archive/fs-version-001
+drwxrwxr-x          - public_html
+drwxrwxr-x          - public_html/incoming
+drwxrwxr-x          - public_html/results
+drwxrwxr-x          - public_html/static
+drwxrwxr-x          - public_html/static/css
+drwxrwxr-x          - public_html/static/css/v0.2
+drwxrwxr-x          - public_html/static/css/v0.2/css
+-rw-rw-r--        308 public_html/static/css/v0.2/css/pbench_utils.css
+drwxrwxr-x          - public_html/static/css/v0.3
+drwxrwxr-x          - public_html/static/css/v0.3/css
+-rw-rw-r--      11798 public_html/static/css/v0.3/css/LICENSE.TXT
+-rw-rw-r--       3663 public_html/static/css/v0.3/css/jschart.css
+drwxrwxr-x          - public_html/static/js
+drwxrwxr-x          - public_html/static/js/v0.2
+drwxrwxr-x          - public_html/static/js/v0.2/js
+-rw-rw-r--       9415 public_html/static/js/v0.2/js/app.js
+-rw-rw-r--       5556 public_html/static/js/v0.2/js/pbench_utils.js
+drwxrwxr-x          - public_html/static/js/v0.3
+drwxrwxr-x          - public_html/static/js/v0.3/js
+-rw-rw-r--      11798 public_html/static/js/v0.3/js/LICENSE.TXT
+-rw-rw-r--     143934 public_html/static/js/v0.3/js/jschart.js
+drwxrwxr-x          - public_html/users
+--- pbench-satellite tree state
++++ pbench-satellite-local tree state (/var/tmp/pbench-test-server/test-26.3/pbench-satellite-local)
+drwxrwxr-x          - logs
+drwxrwxr-x          - pbench-move-results-receive
+drwxrwxr-x          - pbench-move-results-receive/fs-version-002
+drwxrwxr-x          - quarantine
+drwxrwxr-x          - quarantine/duplicates-002
+drwxrwxr-x          - quarantine/errors-002
+drwxrwxr-x          - quarantine/md5-002
+drwxrwxr-x          - tmp
+--- pbench-satellite-local tree state
++++ pbench log file contents
+++++ pbench-local/logs
+---- pbench-local/logs
+++++ pbench-satellite-local/logs
+---- pbench-satellite-local/logs
+--- pbench log file contents
--- a/server/bin/state/test-26.1.config/pbench-server.cfg
+++ b/server/bin/state/test-26.1.config/pbench-server.cfg
@@ -1,9 +1,7 @@
 install-dir = %(unittest-dir)s/opt/pbench-server
 
-###########################################################################
-# crontab roles
-[pbench-server-backup]
-bucket_name = SPECIAL_BUCKET
+[logging]
+logger_type = file
 
 ###########################################################################
 # The rest will come from the global state config file and the default config file.
--- a/server/bin/state/test-26.2.config/pbench-server.cfg
+++ b/server/bin/state/test-26.2.config/pbench-server.cfg
@@ -1,9 +1,7 @@
 install-dir = %(unittest-dir)s/opt/pbench-server
 
-###########################################################################
-# crontab roles
-[pbench-server-backup]
-bucket_name = SPECIAL_BUCKET
+[logging]
+logger_type = devlog
 
 ###########################################################################
 # The rest will come from the global state config file and the default config file.
--- a/server/bin/test-bin/test_logger_type.py
+++ b/server/bin/test-bin/test_logger_type.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+# -*- mode: python -*-
+
+import os, sys
+import logging
+
+from pbench import PbenchConfig, BadConfig, get_pbench_logger
+
+_NAME_ = "pbench-logger-test"
+cfg_name = os.environ["CONFIG"]
+logdir = os.environ["LOGSDIR"]
+
+log_files = {
+    "file": "file.log",
+    "devlog": "devlog.log",
+    "hostport": "hostport.log"
+    }
+
+log_msgs = {
+    "file": "logger_type=file in file file.log",
+    "devlog": "logger_type=devlog in file devlog.log",
+    "hostport": "logger_type=hostport in file hostport.log"
+}
+
+def mock_the_handler(logger, logger_type, fname):
+    
+    # Assumption: only one Handler is present.
+    hdlr = logger.logger.handlers[0]
+    logger.logger.removeHandler(hdlr)
+
+    # logger.logger is used: the first logger is used to format 
+    # the logs with the help of _styleAdapter and the second is 
+    # used to log the messages
+    fh = logging.FileHandler(os.path.join(logdir,fname))
+    fh.setLevel(logging.DEBUG)
+    logger.logger.addHandler(fh)
+    
+    return logger
+
+def test_pbench_logger():
+
+    config = PbenchConfig(cfg_name)
+    logger = get_pbench_logger(_NAME_, config)
+
+    logger_type = config.get("logging", "logger_type")
+
+    logger = mock_the_handler(logger, logger_type, log_files[logger_type])
+    logger.debug(log_msgs[logger_type])
+    
+    if os.path.isfile(os.path.join(logdir,log_files[logger_type])):
+        with open(os.path.join(logdir,log_files[logger_type]), 'r') as f:
+            assert f.read()[:-1] == log_msgs[logger_type], "Mismatch: the file did not contain the expected message."
+
+if __name__ == "__main__":
+    try:
+        test_pbench_logger()
+    except BadConfig as bd:
+        print("BadConfig exception was raised")
+        sys.exit(1)
--- a/server/bin/unittests
+++ b/server/bin/unittests
@@ -49,6 +49,7 @@ function _run {
     else
         echo "+++ Running ${tname} ${@}" >> ${_testout}
     fi
+    export PYTHONPATH=${_testopt}/lib:$PYTHONPATH
     ${tname} ${@} >> ${_testout} 2>&1
     echo "--- Finished ${tname} (status=${?})" >> ${_testout}
 }
@@ -302,6 +303,7 @@ function _setup_state {
 
     # Expected location of the final configuration files
     export CONFIG=$_testopt/lib/config/pbench-server.cfg
+    export LOGSDIR=$_testdir_local/logs
     export SATCONFIG=$_testopt_sat/lib/config/pbench-server.cfg
 
     # The activate invocations are supposed to work without CONFIG being set,
@@ -314,6 +316,12 @@ function _setup_state {
     _state_config="${_tdir}/state/${1}.config/pbench-server.cfg"
     if [ ! -e ${_state_config} ]; then
         _state_config="${_tdir}/state/config/pbench-server.cfg"
+    else
+        # the testcase-specific config file includes the global state config file
+        # but under the name "state-pbench-server.cfg" to prevent conflicts. It also
+        # gets the default section below, so we have to include the install-dir line in it
+        # and delete it from the global state config file.
+        sed 1d "${_tdir}/state/config/pbench-server.cfg" > ${_testopt}/lib/config/state-pbench-server.cfg
     fi
     > ${_testtmp}/pbench-server.cfg
     printf "[DEFAULT]\n"                   >> ${_testtmp}/pbench-server.cfg
@@ -340,6 +348,12 @@ function _setup_state {
     _state_config="${_tdir}/state/${1}.config-satellite/pbench-server.cfg"
     if [ ! -e ${_state_config} ]; then
         _state_config="${_tdir}/state/config-satellite/pbench-server.cfg"
+    else
+        # the testcase-specific config file includes the global state config file
+        # but under the name "state-pbench-server.cfg" to prevent conflicts. It also
+        # gets the default section below, so we have to include the install-dir line in it
+        # and delete it from the global state config file.
+        sed 1d "${_tdir}/state/config/pbench-server.cfg" > ${_testopt}/lib/config/state-pbench-server.cfg
     fi
     > ${_testtmp}/pbench-server.cfg
     printf "[DEFAULT]\n"                   >> ${_testtmp}/pbench-server.cfg
@@ -462,8 +476,10 @@ function _reset_state {
 
     export PATH=${_orig_PATH}
     unset CONFIG
+    unset LOGSDIR
     unset SATCONFIG
 
+    rm -f ${_testopt}/lib/config/state-pbench-server.cfg
     rm -f $_testactout
     rm -f $_testlog
     rm -f $_testactlog
@@ -678,6 +694,19 @@ declare -A cmds=(
     # Tests that the behavior of the find command works as we expect to
     # ensure that buckets for pbench-unpack-tarballs work properly.
     [test-25]="_run test-find-behavior"
+
+    # Test to Log messages on File
+    [test-26.1]="_run test_logger_type.py"
+
+    # Test to Log messages on devlog
+    [test-26.2]="_run test_logger_type.py" 
+
+    # Test to check error when logger_port and logger_host are not provided with hostport
+    [test-26.3]="_run test_logger_type.py"
+
+    # Test to Log messages on hostport with logger-host and logger_port
+    [test-26.4]="_run test_logger_type.py"  
+
 )
 all_tests_sorted=$(for x in ${!cmds[@]} ;do echo $x ;done | sed 's/\./-/' | sort -n -t '-' -k 3 | sort -n -t '-' -k 2 --stable | sed 's/\(.*-[0-9]\+\)-\([0-9]\+\)/\1.\2/')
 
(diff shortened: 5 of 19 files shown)


server/lib/pbench/__init__.py
```diff
@@ -153,7 +153,8 @@ def get_pbench_logger(caller, config):
         except FileExistsError:
             # directory already exists, ignore
             pass
-        fh = logging.FileHandler(os.path.join(logdir, "{}.log".format(caller)))
+        #fh = logging.FileHandler(os.path.join(logdir, "{}.log".format(caller)))
+        fh = SysLogHandler(address=('localhost', 514), facility=19, socktype = socket.SOCK_STREAM)
```
portante: Great!

Can you also verify that just using `/dev/log` also works?  See ["How to Configure Logging to Syslog in Python"](https://stackoverflow.com/questions/3968669/how-to-configure-logging-to-syslog-in-python).

And then we need to make the use of the `FileHandler` and `SyslogHandler` configurable so that for the unit test environment we can use `FileHandler` and all other environments use `SyslogHandler`.

riya-17: @portante made the 1st change i.e. localhost to /dev/log . Will work on second part

server/lib/pbench/__init__.py
```diff
@@ -153,7 +153,8 @@ def get_pbench_logger(caller, config):
         except FileExistsError:
             # directory already exists, ignore
             pass
-        fh = logging.FileHandler(os.path.join(logdir, "{}.log".format(caller)))
+        #fh = logging.FileHandler(os.path.join(logdir, "{}.log".format(caller)))
+        fh = SysLogHandler(address='/dev/log', facility=6, socktype = socket.SOCK_DGRAM)
```
portante: We need options here, something like:

```
    if config.logger_type == "file":
        handler = logging.FileHandler(os.path.join(logdir, "{}.log".format(caller)))
    elif config.logger_type == "devlog":
        handler = logging.SyslogHandler(address='/dev/log', facility=6)
    elif config.logger_type == "hostport":
        handler = logging.SysLogHandler(address=(config.logger_host, config.logger_port), facility=6, socktype = socket.SOCK_DGRAM)
    else:
        raise Exception("Unsupported logger type")
```

And change the references to `fh` to `handler` since we no longer always have a "(f)ile (h)andler".

Then change the `PbenchConfig()` class to look for the "`logger_type`", defaulting it to "`devlog`".

riya-17: @portante Please may I know. Why do we need different logging Handlers? We were supposed to have FileHandler for unittests and SysLogHandler for the rest.

ndokos: I believe it's for flexibility: /dev/log allows logging to the rsyslog running on the server (which may in turn decide to send the logs to a remote server), whereas the logger_host/logger_port allows us to bypass the rsyslog on the server and send directly to a remote rsyslog server.

riya-17: So, I understand that we need to distinguish between the three types of files. Just wanted to ask if there is a predefined way in pbench to do this or we need to create this functionality. Also, if we are going to create this then is there going to be a specific way to create this or we can do this using a hack.
I tried to find a way to get log type using:
- I have been through log files but couldn't find anything that will specify the type of log files, not even in the log messages.
- tried to find any function or attribute inside PbenchConfig but no help there.

riya-17:
@portante @ndokos Can you please review the PR

server/lib/pbench/__init__.py
```diff
@@ -216,6 +224,18 @@ def __init__(self, cfg_name):
             self.COMMIT_ID = self.conf.get("pbench-server", "commit_id")
         except NoOptionError:
             self.COMMIT_ID = "(unknown)"
+        
+        try:
+            self.logger_type = self.conf.get("logging", "logger")
+        except (NoOptionError, NoSectionError):
+            self.logger_type = "devlog"
+
+        if self.logger_type == "hostport":
```
portante: Consider moving this under an `else:` clause of the `try:` block of line 228, something like:

```
try:
    self.logger_type = self.conf.get("logging", "logger")
except (NoOptionError, NoSectionError):
    self.logger_type = "devlog"
else:
    if self.logger_type == "hostport":
        try:
            self.logger_host = self.conf.get("logging", "logger_host")
            self.logger_port = self.conf.get("logging", "logger_port")
        except (NoOptionError) as exc:
            # Note we don't have to check for a `NoSectionError` here because we have already successfully fetched the `logger` from the `logging` section above.
            raise BadConfig(exc)
```

riya-17: Ok sure

riya-17:
@portante @ndokos PR is updated as requested

server/lib/pbench/__init__.py
```diff
@@ -153,16 +154,25 @@ def get_pbench_logger(caller, config):
...
             pass
-        fh = logging.FileHandler(os.path.join(logdir, "{}.log".format(caller)))
-        fh.setLevel(logging.DEBUG)
+
+        if config.logger_type == "file":
+            handler = logging.FileHandler(os.path.join(logdir, "{}.log".format(caller)))
+        elif config.logger_type == "devlog":
+            handler = logging.SyslogHandler(address='/dev/log', facility=logging.handlers.SysLogHandler.LOG_DAEMON)
+        elif config.logger_type == "hostport":
+            handler = logging.SysLogHandler(address=(config.logger_host, int(config.logger_port)))
```
ndokos: I think we want to use TCP, so you will need to add a `socktype=socket.SOCK_STREAM ` here,  since the default is UDP (see https://docs.python.org/3/library/logging.handlers.html#sysloghandler).

riya-17: ok sure

riya-17:
@ndokos Updated the requested changes

psuriset:
Looks good to me.

server/lib/pbench/__init__.py
ndokos: Well, famous last words - running some experiments with SOCK_STREAM has uncovered problems: apparently the Python logging module's implementation of the SOCK_STREAM interface is buggy.

So please revert this change: delete the `socktype` option and just let it default to UDP.

server/lib/pbench/__init__.py
```diff
@@ -153,16 +154,25 @@ def get_pbench_logger(caller, config):
         except FileExistsError:
             # directory already exists, ignore
             pass
-        fh = logging.FileHandler(os.path.join(logdir, "{}.log".format(caller)))
-        fh.setLevel(logging.DEBUG)
+
+        if config.logger_type == "file":
+            handler = logging.FileHandler(os.path.join(logdir, "{}.log".format(caller)))
+        elif config.logger_type == "devlog":
+            handler = logging.SyslogHandler(address='/dev/log', facility=logging.handlers.SysLogHandler.LOG_DAEMON, socktype=socket.SOCK_STREAM)
```
ndokos: The SOCK_STREAM implementation seems to be buggy. Just delete the `socktype` option both here and on line 163.

ndokos: Should the facility be really set here? Why not leave it out and let it default to LOG_USER? I'm not sure I understand why this case is different. Can you explain?

riya-17: This helps to log messages to /var/log/daemon.log . This can be removed according to requirement.

server/lib/pbench/__init__.py
portante: @riya-17, let's add a comment, here, and in `pbench-server-default.cfg`, that the `hostport` logger type is UDP based logging.

riya-17:
@portante yes sure we can schedule a meeting with Nick. We have planned something like creating a test Kibana environment and then see whether the log messages are going or not and comparing them, So, we are open to ideas and would love to discuss

server/bin/pytests/test_26.py
```diff
@@ -0,0 +1,54 @@
...
+
+    logger_type = config.get("logging", "logger_type")
+    log = mock_the_handler(logger, logger_type, files[logger_type])
+    log.debug(messages[logger_type])
+    
+    try:
+        if os.path.isfile(os.path.join(logdir,files[logger_type])):
+            with open(os.path.join(logdir,files[logger_type]), 'r') as f:
+                assert f.read()[:-1] == messages[logger_type]
+            os.remove(os.path.join(logdir,files[logger_type]))
```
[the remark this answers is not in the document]
ndokos: In fact, it would probably be a good idea to get the contents of the file into the output of the test, so we can compare it against the gold file.

server/bin/unittests
```diff
@@ -674,6 +688,9 @@ declare -A cmds=(
 
     # Tests config file that specifies two shims versions
     [test-24]="_run_activate"
+
+    # Test to Log messages on server Centrally
+    [test-26]="_run_pytest test_26"
```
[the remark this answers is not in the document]
riya-17: Test-25 is added by peter in his #1400 PR

server/bin/pytests/test_26.py
```diff
@@ -0,0 +1,54 @@
...
+    except NoFileError:
+        print("NoFileError: File Does Not exist")
```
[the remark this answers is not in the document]
riya-17: If the file is not generated

server/bin/pytests/test_26.py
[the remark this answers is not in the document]
riya-17: every time a different file is created as file, devlog, hostport have different filename, but gold file will only be able to see one file and will throw error for the other 2. We can do like, we can create all files but compare the content of only mentioned file, this can save the test.

server/bin/pytests/test_26.py
```diff
@@ -0,0 +1,54 @@
...
+    logger = get_pbench_logger(_NAME_, config)
+
+    logger_type = config.get("logging", "logger_type")
+    log = mock_the_handler(logger, logger_type, files[logger_type])
+    log.debug(messages[logger_type])
+    
+    try:
+        if os.path.isfile(os.path.join(logdir,files[logger_type])):
+            with open(os.path.join(logdir,files[logger_type]), 'r') as f:
+                assert f.read()[:-1] == messages[logger_type]
```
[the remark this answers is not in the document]
riya-17: ok, i will see

server/bin/pytests/test_26.py
```diff
@@ -0,0 +1,54 @@
...
+    
+    return logger
+
+def test_pbench_logger():
+
+    config = PbenchConfig(cfg_name)
+    logger = get_pbench_logger(_NAME_, config)
+
+    logger_type = config.get("logging", "logger_type")
+    log = mock_the_handler(logger, logger_type, files[logger_type])
```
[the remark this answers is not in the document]
riya-17: ok

server/bin/pytests/test_26.py
```diff
@@ -0,0 +1,54 @@
...
+
+messages = {
+    "file": "logger_type=file in file foo",
+    "devlog": "logger_type=devlog in file bar",
+    "hostport": "logger_type=hostport in file baz"
+}
+
+def mock_the_handler(logger, logger_type, fname):
+    
+    hdlr = logger.logger.handlers[0]
```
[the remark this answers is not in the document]
riya-17: Ok, I will add the comments

server/bin/pytests/test_26.py
```diff
@@ -0,0 +1,54 @@
...
+import logging
+import pytest
+
+from pbench import PbenchConfig, BadConfig, get_pbench_logger
+
+_NAME_ = "pbench-logger-test"
+cfg_name = os.environ.get("CONFIG")
+logdir = "/var/tmp/pbench-test-server/test-26/pbench-local/logs"
+
+files = {
```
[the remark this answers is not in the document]
riya-17: ok

server/bin/pytests/test_26.py
```diff
@@ -0,0 +1,54 @@
...
+
+import os, sys
+import logging
+import pytest
+
+from pbench import PbenchConfig, BadConfig, get_pbench_logger
+
+_NAME_ = "pbench-logger-test"
+cfg_name = os.environ.get("CONFIG")
+logdir = "/var/tmp/pbench-test-server/test-26/pbench-local/logs"
```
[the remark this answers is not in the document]
riya-17: ok

server/bin/pytests/test_26.py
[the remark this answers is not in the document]
riya-17: Should I also add the exception, as if there is no handler it will cause error

server/bin/pytests/test_26.py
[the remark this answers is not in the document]
riya-17: Yes, I found this assert as the best without any problem

server/bin/unittests
```diff
@@ -59,6 +59,12 @@ function _run_indexing {
     _run pbench-index --tool-data
 }
 
+function _run_pytest {
+    lname=${1}
```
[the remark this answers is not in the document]
riya-17: Sorry that was by mistake. I changed it to 'tname'

server/bin/unittests
```diff
@@ -59,6 +59,12 @@ function _run_indexing {
...
+    export PYTHONPATH=${_testopt}/lib:$PYTHONPATH
+    pytest ${_testopt}/pytests/${lname}.py
```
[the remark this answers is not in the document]
riya-17: sure

server/bin/unittests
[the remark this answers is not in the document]
ndokos: OK - planning ahead is good!

server/bin/unittests
```diff
@@ -674,6 +690,9 @@ declare -A cmds=(
 
     # Tests config file that specifies two shims versions
     [test-24]="_run_activate"
+
+    # Test to Log messages on server Centrally
+    [test-26]="_run_pytest test_26"
```
ndokos: Sw we will have test-26.{1,2,3} but all three will run `test_logger_type`.

server/bin/pytests/test_26.py
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
riya-17: ok

server/bin/pytests/test_26.py
```diff
@@ -0,0 +1,59 @@
...
+
+log_msgs = {
+    "file": "logger_type=file in file foo",
+    "devlog": "logger_type=devlog in file bar",
+    "hostport": "logger_type=hostport in file baz"
+}
+
+def mock_the_handler(logger, logger_type, fname):
+    
+    # only one Handler is present.
```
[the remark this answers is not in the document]
riya-17: ok

server/bin/pytests/test_26.py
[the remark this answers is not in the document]
riya-17: ok

server/bin/unittests
```diff
@@ -445,7 +460,7 @@ function _run_test {
     _save_tree
     _dump_logs
     _verify_output ${testname}
-    _reset_state ${testname}
+    #_reset_state ${testname}
```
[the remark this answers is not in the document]
riya-17: yaa

server/bin/pytests/test_logger_type.py
```diff
@@ -0,0 +1,58 @@
...
+
+def test_pbench_logger():
+
+    config = PbenchConfig(cfg_name)
+    logger = get_pbench_logger(_NAME_, config)
+
+    logger_type = config.get("pbench-server", "logger_type")
+
+    if(logger_type=='file' or logger_type=='devlog' or logger_type=='hostport'):
+        logger.info("Valid logger type")
```
ndokos: We should not do this: it will use the actual logger which might send log messages to some remote service, every time somebody runs the unit test. Just print it.

server/bin/unittests
```diff
@@ -59,6 +59,12 @@ function _run_indexing {
     _run pbench-index --tool-data
 }
 
+function _run_pytest {
+    tname=${1}
+    export PYTHONPATH=${_testopt}/lib:$PYTHONPATH
+    pytest ${_testopt}/pytests/${tname}.py
```
ndokos: 1) Wrap the run with `echo ...` similar to what `_run` does.
2) Redirect the output of the test to `${_testout}` similar to what `_run` does.
3) Run pytest with -q to quiet down its output. The normal output prints out python versions which we don't want in the output file: they are going to cause spurious failures if somebody is using a slightly different version of python or pytest.

server/lib/config/pbench-server-default.cfg
```diff
@@ -110,6 +110,12 @@ webserver = %(default-host)s
 host-info-path-prefix = pbench-results-host-info.versioned/pbench-results-host-info.URL
 host_info_url = http://%(webserver)s/%(host-info-path-prefix)s
 
+# hostport logger type is UDP based logging.
```
ndokos: ".... uses UDP-based logging."

server/lib/pbench/__init__.py
```diff
@@ -216,6 +224,18 @@ def __init__(self, cfg_name):
...
             self.COMMIT_ID = "(unknown)"
+        
+        try:
+            self.logger_type = self.conf.get("logging", "logger")
+        except (NoOptionError, NoSectionError):
+            self.logger_type = "devlog"
+        else:
+            if self.logger_type == "hostport":
+                try:
+                    self.logger_host = self.conf.get("logging", "logger_host")
```
ndokos: AFAICT, there is no definition of `logger_host` in any config file, even if `logger_type` is set to `hostport`.  Yet this exception does not seem to trigger (I don't see any such output in the gold file e.g.) and 26.3 passes. This needs investigating: we should get  (or manufacture) an exception and the test should *NOT* pass.

server/bin/pytests/test_logger_type.py
```diff
@@ -0,0 +1,58 @@
...
+    logger.logger.addHandler(fh)
+    
+    return logger
+
+def test_pbench_logger():
+
+    config = PbenchConfig(cfg_name)
+    logger = get_pbench_logger(_NAME_, config)
+
+    logger_type = config.get("pbench-server", "logger_type")
```
ndokos: I think this is wrong: the logger_type is supposed to be defined in the `logging` section. See `lib/pbench/__init.py__`. Please check and fix if needed.

server/bin/state/test-26.1.config/pbench-server.cfg
```diff
@@ -0,0 +1,55 @@
...
+commit_id = unit-test
+environment = unit-test
+admin-email = unit-test-user@example.com
+pbench-top-dir = %(unittest-dir)s/pbench
+pbench-local-dir = %(unittest-dir)s/pbench-local
+pbench-backup-dir = %(pbench-local-dir)s/archive.backup
+# Add role for sync'ing with satellites
+roles = pbench-prep, pbench-results, pbench-backup, pbench-sync-satellites
+debug_unittest = True
+logger_type = file
```
ndokos: If `logger_type` is indeed supposed to be defined in the `logging` section, then this (and similar for the other two config files) will need to be moved to such a section.

server/lib/pbench/__init__.py
ndokos: That is, it should not pass if logger_type is set to hostport but logger_host or logger_port (or both) are missing.

server/bin/pytests/test_logger_type.py
ndokos: ... and the option in the config file that PbenchConfig is looking for is `logger`, not `logger_type`.
This all needs straightening out.

server/lib/pbench/__init__.py
```diff
@@ -229,6 +249,7 @@ def mocked_time():
                 return 0
             global _time
             _time = mocked_time
+            self.logger_type = "file"
```
ndokos: This gave me conniptions. Please get rid of it!

server/lib/config/pbench-server-default.cfg
```diff
@@ -110,6 +110,12 @@ webserver = %(default-host)s
 host-info-path-prefix = pbench-results-host-info.versioned/pbench-results-host-info.URL
 host_info_url = http://%(webserver)s/%(host-info-path-prefix)s
 
+# hostport logger type uses UDP-based logging.
+[logging]
+logger_type = file
+logger_host = localhost
```
ndokos: We don't want logger_host and logger_port defined at all *UNLESS* logger_type is set to hostport.

server/lib/pbench/__init__.py
```diff
@@ -7,7 +7,7 @@
 import time
 import json
 import errno
-import logging
+import logging.handlers
```
ndokos: import logging
     from logging import handlers

     logging,get_logger(...)
     handlers.SysLogHandler(...)

server/lib/pbench/__init__.py
```diff
@@ -153,16 +154,25 @@ def get_pbench_logger(caller, config):
         except FileExistsError:
             # directory already exists, ignore
             pass
-        fh = logging.FileHandler(os.path.join(logdir, "{}.log".format(caller)))
-        fh.setLevel(logging.DEBUG)
+
+        if config.logger_type == "file":
+            handler = logging.FileHandler(os.path.join(logdir, "{}.log".format(caller)))
+        elif config.logger_type == "devlog":
+            handler = logging.handlers.SysLogHandler(address='/dev/log')
```
ndokos: If we import `handlers` as I suggested above:

      from logging import handlers

then this becomes:

     handler = handlers.SysLogHandler(....)

server/lib/pbench/__init__.py
```diff
@@ -153,16 +154,25 @@ def get_pbench_logger(caller, config):
...
+        elif config.logger_type == "hostport":         #hostport logger type is UDP based logging
+            handler = logging.handlers.SysLogHandler(address=(config.logger_host, int(config.logger_port)))
```
ndokos: Ditto.

server/lib/pbench/__init__.py
ndokos: The upshot is that we need two unit tests for `hostport`: one where we don't define logger_host/logger_port in the (test) config file),  in which case we expect an exception here; and one where they are both defined (logger_host will have to be set to localhost in that case, otherwise a socket() call will fail - DAMHIKT), in which case, the pytest should do its thing.

server/bin/pytests/test_logger_type.py
```diff
@@ -0,0 +1,58 @@
...
+    return logger
+
+def test_pbench_logger():
+
+    config = PbenchConfig(cfg_name)
+    logger = get_pbench_logger(_NAME_, config)
+
+    logger_type = config.get("logging", "logger_type")
+
+    if(logger_type=='file' or logger_type=='devlog' or logger_type=='hostport'):
```
ndokos: ... or (shorter and a bit more readable):

          if logger_type in ['file', 'devlog', 'hostport']:
                ....
or even:

          if logger_type in log_files.keys():
               ....

although this last one is probably too cryptic.

server/bin/unittests
```diff
@@ -59,6 +59,19 @@ function _run_indexing {
...
+function _run_pytest {
+    tname=${1}
+    export PYTHONPATH=${_testopt}/lib:$PYTHONPATH
+    shift
+    if [[ -z "${@}" ]]; then
+        echo "+++ Running ${tname}" >> ${_testout}
+    else
+        echo "+++ Running ${tname} ${@}" >> ${_testout}
+    fi
+    pytest -q ${_testopt}/pytests/${tname}.py
```
ndokos: This needs to be redirected to the _testout file, but configparser needs to be modifed so that the SafeConfigParser deprecation warnings disappear.

ndokos: If you add the following line in `_run_pytest` then the deprecation warning is eliminated, so we can redirect the output and not worry about them:

    export PYTHONWARNINGS="ignore::DeprecationWarning:configparser" 

That's a temporary step that allows you to get the "correct" output in the gold files (i.e. without the deprecation warnings), but still allows reviewers to run the unit tests without having to modify anything on their side.

In the longer term, I'll make the change in configtools and release an updated RPM (bumping up the version in the process, so that when people upgrade pbench-agent or pbench-server, they will also update their configtools installation). At some point in the future, after everybody has upgraded, we can then get rid of the `PYTHONWARNINGS` export in a subsequent PR.

riya-17: sure

ndokos:
The unit test failure is in the agent tests: Travis cannot find a repository. That needs to be fixed, but it is not part of this PR (which is server-side only).

server/bin/unittests
```diff
@@ -678,6 +700,19 @@ declare -A cmds=(
     # Tests that the behavior of the find command works as we expect to
     # ensure that buckets for pbench-unpack-tarballs work properly.
     [test-25]="_run test-find-behavior"
+
+    # Test to Log messages on File
+    [test-26.1]="_run_pytest test_logger_type"
```
portante: Pytest should not need external server configurations to get different behaviors.  With Pytest we can provide all the inputs the unit test needs.  Can we encode all this in the test python code?

ndokos: Isn't the point of the test to specify the logging method *in the config file* and then make sure that the message ends up in the expected place?

portante: Yes, but you can make the config file object in Pytest.  The config file parsing and all that is secondary.

server/lib/pbench/__init__.py
```diff
@@ -229,6 +250,7 @@ def mocked_time():
                 return 0
             global _time
             _time = mocked_time
+            #self.logger_type = "file"
```
portante: Can you remove this line entirely?  Commented-out lines with out comments that describe why it is there can end up reducing the clarity of the code when somebody else tries to read and understand what it does.

server/bin/gold/test-26.3.txt
```diff
@@ -0,0 +1,112 @@
++++ Running test_logger_type.py
+BadConfig exception was raised
+--- Finished test_logger_type.py (status=0)
```
ndokos: We probably should exit with status 1 here.

riya-17: Yaa

server/bin/unittests
```diff
@@ -49,6 +49,8 @@ function _run {
     else
         echo "+++ Running ${tname} ${@}" >> ${_testout}
     fi
+    export PYTHONPATH=${_testopt}/lib:$PYTHONPATH
+    export PYTHONWARNINGS="ignore::DeprecationWarning:configparser"
```
portante: So we need this because ... ?

server/lib/config/pbench-server-default.cfg
```diff
@@ -122,6 +122,12 @@ webserver = %(default-host)s
 host-info-path-prefix = pbench-results-host-info.versioned/pbench-results-host-info.URL
 host_info_url = http://%(webserver)s/%(host-info-path-prefix)s
 
+# hostport logger type uses UDP-based logging.
```
portante: Can you move this comment to the line before `logger_host`?

server/lib/config/pbench-server-default.cfg
```diff
@@ -122,6 +122,12 @@ webserver = %(default-host)s
...
+[logging]
+logger_type = file
```
portante: Can we make the default `devlog`?  And then fix all the other unit tests to use `file` as necessary?

server/bin/test-bin/test_logger_type.py
```diff
@@ -0,0 +1,59 @@
...
+    
+    if os.path.isfile(os.path.join(logdir,log_files[logger_type])):
+        with open(os.path.join(logdir,log_files[logger_type]), 'r') as f:
+            assert f.read()[:-1] == log_msgs[logger_type], "Mismatch: the file did not contain the expected message."
+
+if __name__ == "__main__":
+    try:
+        test_pbench_logger()
+    except BadConfig as bd:
+        print("BadConfig exception was raised")
```
portante: Can you make the message print additional information?  Something like:

```
print("BadConfig exception was raised: {:r}".format(bd))
```

riya-17: Printing the additional information may cause error as it will include python versions which will create problem due to different version in different machines.

server/bin/unittests
riya-17: Because they are removing the older version of configparser which is throwing warning that, "change your configparser into the latest one if you are using the older version else it will create problem in the future"

server/lib/config/pbench-server-default.cfg
riya-17: sure

server/lib/config/pbench-server-default.cfg
riya-17: Changing the default to devlog doesn't make any difference to unit-tests as they have their own config files.

server/bin/unittests
portante: @ndokos, does https://github.com/distributed-system-analysis/configtools/pull/16 fix this?

ndokos:
@gurbirkalsi, @Anisha1234 :the agent and server unit tests pass but the dashboard unit tests are failing again. Can you please take a look?

server/bin/state/test-26.1.config/pbench-server.cfg
```diff
@@ -0,0 +1,10 @@
+install-dir = %(unittest-dir)s/opt/pbench-server
+
+[logging]
+logger_type = file
+
+###########################################################################
+# The rest will come from the default config file.
```
ndokos: The rest will come from the global state file and the default config file.

server/bin/state/test-26.2.config/pbench-server.cfg
```diff
@@ -0,0 +1,10 @@
+install-dir = %(unittest-dir)s/opt/pbench-server
+
+[logging]
+logger_type = devlog
+
+###########################################################################
+# The rest will come from the default config file.
```
ndokos: Ditto.

server/bin/unittests
ndokos: Yes, it does.

server/bin/unittests
```diff
@@ -314,6 +316,12 @@ function _setup_state {
     _state_config="${_tdir}/state/${1}.config/pbench-server.cfg"
     if [ ! -e ${_state_config} ]; then
         _state_config="${_tdir}/state/config/pbench-server.cfg"
+    else
+        # the testcase-specific config file includes the global state config file
+        # but under the name "state-pbench-server.cfg" to prevent conflicts. It also
+        # gets the defaul section below, so we have to include the install-dir line in it
```
portante: s/defaul/default

server/bin/unittests
```diff
@@ -314,6 +316,12 @@ function _setup_state {
...
+        # and delete it from the global state config file.
+       sed 1d "${_tdir}/state/config/pbench-server.cfg" > ${_testtmp}/state-pbench-server.cfg
```
portante: Seems like the `sed` command is indented 3 spaces instead of 4 ...

Also, why not just copy this state file to the same location as the default config file?

```
        sed 1d "${_tdir}/state/config/pbench-server.cfg" > ${_testopt}/lib/config/state-pbench-server.cfg
```

And then clean it up explicitly below when the test is done?
```

File: server/bin/unittests [hunk 44]
```diff
@@ -324,7 +332,7 @@ function _setup_state {
     echo "$_testopt/bin/pbench-server-config-activate ${_testtmp}/pbench-server.cfg" >> $_testactout
     $_testopt/bin/pbench-server-config-activate ${_testtmp}/pbench-server.cfg >> $_testactout
     rc=$?
-    rm ${_testtmp}/pbench-server.cfg
+    rm ${_testtmp}/*pbench-server.cfg
```
portante: If we move it to the default location, we don't need this removal here.  And, in fact, doesn't removing it here mean that during test runs tests won't have the config file in place?

server/bin/unittests
```diff
@@ -340,6 +348,12 @@ function _setup_state {
     _state_config="${_tdir}/state/${1}.config-satellite/pbench-server.cfg"
     if [ ! -e ${_state_config} ]; then
         _state_config="${_tdir}/state/config-satellite/pbench-server.cfg"
+    else
+        # the testcase-specific config file includes the global state config file
+        # but under the name "state-pbench-server.cfg" to prevent conflicts. It also
+        # gets the defaul section below, so we have to include the install-dir line in it
```
portante: s/defaul/default

Same comments as above.

server/bin/unittests
```diff
@@ -350,7 +364,7 @@ function _setup_state {
     echo "$_testopt_sat/bin/pbench-server-config-activate ${_testtmp}/pbench-server.cfg" >> $_testactout
     $_testopt_sat/bin/pbench-server-config-activate ${_testtmp}/pbench-server.cfg >> $_testactout
     rc=$?
-    rm ${_testtmp}/pbench-server.cfg
+    rm ${_testtmp}/*pbench-server.cfg
```
portante: Same comment as above.

server/bin/unittests
```diff
@@ -462,6 +476,7 @@ function _reset_state {
 
     export PATH=${_orig_PATH}
     unset CONFIG
+    unset LOGSDIR
```
portante: We also need to remove the state config file we moved to standard location below.

server/bin/unittests
ndokos: No - once the (top-level) config file is assembled from its various pieces, `pbench-server-config-activate` copies it *and all other config files it finds in the same directory* to the destination directory, so the config file in the tmp directory can safely be deleted.

server/bin/unittests
ndokos: We blow away the whole directory, so there is no need to deal with individual files, is there?

portante: We want to make sure we clean up after ourselves in the unit test harness so that we can tell when the server leaves unexpected files around due to its operation.

server/bin/unittests
```diff
@@ -477,6 +478,7 @@ function _reset_state {
     export PATH=${_orig_PATH}
     unset CONFIG
     unset LOGSDIR
+    unset STATECONFIG
```
ndokos: This does nto remove the file: it just unsets the env variable you defined on l. 307 (which btw is not used anywhere, so there no reason to define it).

ndokos: It's the file that's (possibly) created on l.325 that need to be removed in the section below: you need to do

      rm -f ${_testopt}/lib/config/state-pbench-server.cfg

This will work whether the file exists or not: the -f suppresses the error message if it does not exist.

server/bin/test-bin/test_logger_type.py
portante: I agree that might be true if you were emitting a stack track, but in this case it should just be the text of the exception that was raised.

Issue #1330, opened 2019-09-03
Internal users of npm and yarn must switch to internal Nexus proxy server


gurbirkalsi:
Pbench dashboard will require this migration to handle downloading JS packages from the internal Nexus proxy registry. This configuration will have to be applied to the dev, staging, and production environments altogether as each environment is either developed or built within the internal network.

We can initialize `.yarnrc` and `.npmrc` files to handle the routing to the Nexus server and reinstall/redeploy the dashboard with those loaded configurations. 

This migration probably won't apply to the jschart node module as it is not dependent on being developed/built under internal network. We don't query visualization data at any point that requires a secure connection to the internal network.

gurbirkalsi:
Issue resolved with merging of PR #1334

PR #1334, opened 2019-09-05
Update README to detail option of downloading packages from private registry

gurbirkalsi:


gurbirkalsi:
Resolves Issue #1330

gurbirkalsi:
I agree! That would leave less room for potential conflict. I'll update the names of these files and reference them as examples in the README.

PR #1335, opened 2019-09-06
Fixes for indexing, add support for indexing trafficgen data

portante:
Reworked the handling of tools so that we have one method for handling it and we avoid throwing stack traces as a result of missing tool data.

Change a few "error" log entries emitted by the indexer to "warnings" since they don't represent an error with how the indexing is working.

Fixed a missing `counters` attribute of a `PbenchTemplate` and added a few more uses of those counters.

Provided better names for all the "operational context" counters so that we know where in the indexing of the data the counters were created, while only printing out counters that have data.

Finally, we can now index result.json files generated by trafficgen.  We have added a sample trafficgen tar ball which contains minimal tool data, and a fairly reasonable amount of result data.  We have updated the
method of indexing to relax the names of the result arrays in the JSON so that we can accommodate all the various data arrays from trafficgen.

This means that what we used to call "`measurement_units`" is now called "`measurement_title`".  We also add a "`measurement_idx`" field to be sure we track how many arrays we are processing, even though it is usually only one.  Further, have fixed up the `mock` module to handle more kinds of data types seen in the data sets we index.


server/bin/gold/test-7.15.txt
```diff
@@ -122,8 +122,8 @@ len(actions) = 30
                     }
                 },
                 {
-                    "hostname": "flat8_40gb",
-                    "hostname-f": "flat8_40gb",
+                    "hostname": "master_40gb",
+                    "hostname-f": "master_40gb",
```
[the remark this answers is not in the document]
portante: Well, it turns out that this change resulted from refactoring the code to use the `get_hosts()` method everywhere.

server/lib/pbench/indexer.py
```diff
@@ -1089,14 +1129,31 @@ def _handle_iteration(self, iter_data, iter_name, iter_number, result_json):
...
             bm_md['uid'] = ResultData.expand_template(bm_md['uid_tmpl'], bm_md, run=self.run_metadata)
 
+        try:
+            bm_md['trafficgen_uid_tmpl'] = bm_md['trafficgen_uid']
+        except KeyError:
+            # Ignore a missing 'trafficgen_uid' field
+            pass
+        else:
+            bm_md['trafficgen_uid'] = ResultData.expand_template(bm_md['trafficgen_uid_tmpl'], bm_md, run=self.run_metadata)
+
```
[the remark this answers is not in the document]
portante: Well because we don't have an infrastructure for generic transformations.  I'd like to solve that via future Scribe integration work, but seems over-kill to do it here given this is only one UID field that needs to be fixed up.

Is that acceptable?

server/lib/pbench/indexer.py
```diff
@@ -1195,112 +1252,108 @@ def gen_sources(results, sublists, iteration, cvt_ts):
...
+                            ( '@timestamp_original', start['date'] ),
+                            # Sample documents inherit the run and
+                            # iteration data of its parent.
+                            ( 'run', iteration['run'] ),
+                            ( 'iteration', iteration['iteration'] ),
+                            ( 'benchmark', iteration['benchmark'] ),
+                            ( 'sample', sample_md )
+                        ])
+                        sample_id = PbenchData.make_source_id(source)
+                        yield source, None, 'sample'
```
ndokos: I don't understand this: why None? where is the ID of the parent iteration doc?

server/lib/pbench/indexer.py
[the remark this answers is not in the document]
ndokos: I would appreciate at least a comment, explaining the problem.

server/lib/pbench/indexer.py
portante: This is the same code as before.  Are the comments at line 1310 and 1311 sufficient?

Basically, the "`source`"  document we are yielding here *is* the parent document used for all the samples that are generated below, see line 1356, where "`sample_id`" is referenced.

ndokos: OK - thanks: I get it now - I think the comment is clear enough, but it didn't penetrate the dense fog in my skull at the time I read it.

portante:
@ndokos, I believe I have resolved all of your comments.  Please take another look, thanks!

server/lib/pbench/indexer.py
```diff
@@ -3527,21 +3527,31 @@ def gen_toc(self):
...
-                    name, os.path.sep)
-                if dname in toc_dirs:
+                if path == "/" or path == "":
+                    dpath = "/"
+                    name = None
+                    path_els = []
+                else:
+                    dpath = path[:-1] if path.endswith(os.path.sep) else path
+                    name = os.path.basename(dpath)
+                    path_els = dpath.split(os.path.sep)[1:-1]
```
ndokos: Would it make sense to have a check somewhere that `dpath` always starts with an `os.path.sep`? The split result would be strange if that were not the case.

ndokos: I just noticed a typo in the commit message: 

"measurement_units" is not called "measurement_title": not-->now

portante: > Would it make sense to have a check somewhere that `dpath` always starts with an `os.path.sep`? The split result would be strange if that were not the case.

Not sure that is necessary.  Before we get to this point of processing the TOC entries, the `PbenchTarBall` constructor at line 3650 has already verified that all members have the proper prefix.  With that, when we strip the `prefix_l` length off the member's `name`, we should always have a leading `os.path.sep`, except in the case of the top-level directory, which we handle at line 3532.

PR #1338, opened 2019-09-10
Make s3 backup optional.

ndokos:
If the S3 service is inaccessible, the backup does not succeed and
keeps retrying. That may be reasonable if the S3 service is up and
running, but there may be situations where we want to bypass it.

Commenting out the [pbench-server-backup] section (or just the
endpoint_url option in that section) in the server config file now
allows the rest of the backup to succeed (including moving the
TO-BACKUP link to the final BACKED-UP state). That allows us to do
local backups, irrespective of the S3 backup service.

test-6.13 implements the commenting out of the endpoint_url option;
test-6.14 implements the commenting out of the [pbench-server-backup]
section (the tar ball is a symlink to the 6.13 tarball).


server/lib/s3backup/__init__.py
```diff
@@ -64,6 +63,11 @@ def __init__(self, config, logger):
         else:
             self.connector = S3Connector(config)
 
+        try:
+            self.bucket_name = self.connector.bucket_name
+        except Exception:
+            pass
```
portante: Can we add a comment explaining why we can `pass` here and ignore the exception?

server/lib/s3backup/__init__.py
```diff
@@ -206,19 +210,25 @@ def getsize(self, tb):
...
-            'pbench-server-backup', 'access_key_id')
-        self.secret_access_key = config.get(
-            'pbench-server-backup', 'secret_access_key')
-        self.s3client = boto3.client('s3',
-                                     aws_access_key_id='{}'.format(
-                                         self.access_key_id),
-                                     aws_secret_access_key='{}'.format(
-                                         self.secret_access_key),
-                                     endpoint_url='{}'.format(self.endpoint_url))
-        self.transfer_config = None
```
portante: We no longer need the `transfer_config` attribute?

server/lib/s3backup/__init__.py
```diff
@@ -334,6 +350,7 @@ def upload_fileobj(self, Body=None, Bucket=None, Key=None, Config=None, ExtraArg
                                        ExtraArgs['Metadata']['MD5']))
 
     def head_bucket(self, Bucket=None):
+
```
portante: Is this addition intentional?

server/bin/pbench-backup-tarballs.py
```diff
@@ -105,6 +105,10 @@ def sanity_check(lb_obj, s3_obj, config, logger):
     # make sure the S3 bucket exists
     try:
         s3_obj.head_bucket(s3_obj.bucket_name)
+    except AttributeError:
+        logger.error(
+            "Bucket: not defined in config file")
```
portante: Can you make this message a `warning` and add to it saying that S3 backups are not enabled as a result?

server/bin/state/test-6.13.config/pbench-server.cfg
```diff
@@ -0,0 +1,54 @@
+install-dir = %(unittest-dir)s/opt/pbench-server
```
portante: What is the difference between the `test-6.13.config/pbench-server.cfg` file and the `test-6.14` one?  And further, can this be symlinks to existing files that are the same?

server/lib/s3backup/__init__.py
ndokos: We do not.

server/lib/s3backup/__init__.py
ndokos: Yes, I'll do that.

server/lib/s3backup/__init__.py
ndokos: I don't think so. I'll revert it.

server/bin/pbench-backup-tarballs.py
ndokos: Sure.

server/bin/pbench-backup-tarballs.py
```diff
@@ -105,9 +105,13 @@ def sanity_check(lb_obj, s3_obj, config, logger):
...
         s3_obj.head_bucket(s3_obj.bucket_name)
+    except AttributeError:
+        logger.warning(
+            "Bucket: not defined in config file - S3 backup is disabled.")
+        s3_obj = None
     except Exception:
-        logger.error(
-            "Bucket: {} does not exist or you have no access".format(s3_obj.bucket_name))
+        logger.warning(
+            "Bucket: {} is not accesssible - S3 backup is disabled".format(s3_obj.bucket_name))
```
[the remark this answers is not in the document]
ndokos: What, you don't speak parseltongue?

server/lib/s3backup/__init__.py
```diff
@@ -64,6 +63,25 @@ def __init__(self, config, logger):
...
+        # not defined at all (which is possible when we deliberately
+        # turn off the S3 backup by deleting the relevant section in
+        # the config file), we would get an AttributeError exception
+        # here - but we don't want to abort the application: it might
+        # still want to continue with local backups, so we ignore the
+        # exception: the application will check it.
+        try:
+            self.bucket_name = self.connector.bucket_name
+        except AttributeError:
+            pass
```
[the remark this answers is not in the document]
ndokos: I was too lazy to change the check. Done now.

PR #1341, opened 2019-09-10
Upgrade Prisma service to v1.34 to enable storage of large string values

gurbirkalsi:
Prisma was reporting the following issue during url generation for run comparison pages: `Error: Value for field config is too long.` This error was preventing pbench dashboard from saving config data to our postgres db.

Upgrading and redeploying the Prisma service with the latest version and migrating the datamodel schema to support syntax changes as part of the version bump has resolved this error.

### DIFF
--- a/web-server/prisma/datamodel.prisma
+++ b/web-server/prisma/datamodel.prisma
@@ -1,6 +1,7 @@
 type Url {
-  id: ID! @unique
-  createdAt: DateTime!
+  id: ID! @id
+  createdAt: DateTime! @createdAt
+  updatedAt: DateTime! @updatedAt
   config: String!
   description: String
 }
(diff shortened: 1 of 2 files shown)


web-server/prisma/datamodel.prisma
```diff
@@ -1,6 +1,7 @@
 type Url {
-  id: ID! @unique
-  createdAt: DateTime!
+  id: ID! @id
```
portante: Can you explain why this needs to change?  It would be great to have this in a comment if possible.

gurbirkalsi: v1.32 of Prisma introduced a new datamodel syntax that was a breaking change. The `id`, `createdAt`, and `updatedAt` fields now allow custom naming for the field while still maintaining the behaviors for the field type as a generator. For example, the following key and value would be valid under this schema: `myId: ID! @id` 

For more on this change: https://www.prisma.io/blog/datamodel-v11-lrzqy1f56c90#use-any-field-as-id-createdat-or-updatedat

PR #1342, opened 2019-09-11
Eliminates shims option from the [pbench-prep] section of pbench-server-default.cfg

### LINKED ISSUE #1307: Get rid of `shims` option in pbench-server-default.cfg
The `shims` option in the `[pbench-prep]` section of pbench-server-default.cfg should be eliminated. The only user is `pbench-server-activate-create-crontab` and it can easily do what it needs by using the `pbench-move-results-receive-versions` from the `[pbench-server]' section.

Anisha1234:
Fixes #1307 
The fix eliminates `shims` option in the `[pbench-prep]` section of `pbench-server-default.cfg` and adress the use of it in `pbench-server-activate-create-crontab` with the help of
`pbench-move-results-receive-versions` from the `[pbench-server]' section.

### DIFF
--- a/server/bin/pbench-server-activate-create-crontab
+++ b/server/bin/pbench-server-activate-create-crontab
@@ -143,11 +143,11 @@ for role in $roles ;do
             done
             ;;
         pbench-prep)
-            shims=$(getconf.py -l shims $role)
+            versions=$(getconf.py -l pbench-move-results-receive-versions pbench-server)
             tasks=$(getconf.py -l tasks $role)
-            for shim in $shims ;do
+            for version in $versions ;do
                 for task in $tasks; do
-                    crontab_with_substitutions $shim $task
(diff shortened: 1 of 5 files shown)


server/bin/pbench-server-activate-create-crontab
```diff
@@ -143,7 +143,8 @@ for role in $roles ;do
             done
             ;;
         pbench-prep)
-            shims=$(getconf.py -l shims $role)
+            version=$(getconf.py -l pbench-move-results-receive-versions pbench-server)
```
ndokos: ATM, there is only one version, but it is conceivable that we might end up with two in the future (the same way that we had two in the past: there is some unforeseen problem that requires us to introduce a second one). This is not going to work in that case

portante: Looks like this has been resolved.

PR #1343, opened 2019-09-12
Adds unit test for TimeseriesGraph component

Anisha1234:
Code Coverage Report:
```
File                             |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line 
components/TimeseriesGraph       |      100 |       100 |      100 |      100 |              |
  index.js                       |      100 |       100 |      100 |      100 |      
```  
Coverage reports are currently retrieved in Jest by running the following command:` umi test ./src/components ./src/pages --verbose --maxWorkers=1 --runInBand "--coverage"`

### DIFF
--- a/web-server/v0.4/src/components/TimeseriesGraph/index.test.js
+++ b/web-server/v0.4/src/components/TimeseriesGraph/index.test.js
@@ -1,0 +1,69 @@
+import React from 'react';
+import { shallow, mount } from 'enzyme';
+import TimeseriesGraph from './index';
+
+const mockProps = {
+  dataSeriesNames: ['name1', 'time', 'name2'],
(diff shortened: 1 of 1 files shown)


Anisha1234:
> Can you squash these commits down to one?
Sure @portante

Anisha1234:
@gurbirkalsi I am already using `setprops` in `componentDidUpdate` and also tried to compare the current props with `prevprops` but still not being able to revoke the conditional statement here. Because `jschart` is a primitive function over here, I am not being able to `spyOn` in as well. Is there any other approach I can go for?

portante:
Based on conversations of this PR, we still need to resolve some issues with it, so we have added the `WIP` label, and are moving it out of the 0.65 release.

PR #1344, opened 2019-09-13
Enhance ansible dashboard README

portante:
We are updating the `README.md` for the dashboard to fix up a few incorrect references to variable names expected by the `config.json.j2` template.  We also adjust the sample `inventory` file.

Lastly, we ignore the `.yarnrc` and `.npmrc` files to avoid folks inadvertently checking them into the code base.

### DIFF
--- a/contrib/ansible/dashboard/inventory
+++ b/contrib/ansible/dashboard/inventory
@@ -5,13 +5,13 @@ user-server
 [servers:vars]
 results_url="http://{{ inventory_hostname }}"
 run_index="run.example"
+graphql_url="http://{{ inventory_hostname }}"
 
 [ui-server]
 ui-server.example.com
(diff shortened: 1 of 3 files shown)


contrib/ansible/dashboard/inventory
```diff
@@ -5,13 +5,13 @@ user-server
 [servers:vars]
 results_url="http://{{ inventory_hostname }}"
 run_index="run.example"
+graphql_url="http://{{ inventory_hostname }}:4000"
```
gurbirkalsi: @portante Why do we specify a hardcoded port of `4000` here?

portante: It is because this is just an example, but I'll take it out.

PR #1350, opened 2019-09-17
pbench-unpack-tarballs and pbench-dispatch: merge and rewrite in python

### LINKED ISSUE #1315: pbench-unpack-tarballs and pbench-dispatch: merge and rewrite in python
Some history: originally there was only one way that users could send tarballs to the server. The tarballs ended up in a directory that would be periodically examined by `pbench-dispatch` which would then proceed to copy the tarballs to the `archive` area and would then "set state" for each tarball to indicate to other scripts what processing should be done on the tarball (e.g. unpacking the tarball, indexing the tarball, backing up the tarball etc). The state is set by creating symbolic links in designated directories. Those directories are scanned by the corresponding script to find all the tarballs that it needs to process. Once the script processes a tarball, it switches the state of the tarball by moving the link to another of the designated directories (or to a directory that nobody scans: that is a final state for the tarball). Here's a state transition diagram (boxes with rounded corners are non-final states; boxes with square corners are final states):

img

The original method of sending results to the server was somewhat fragile, so we added a second more robust method (but we kept the old method for backwards compatibility: the original method is finally going away with PR #1302). That required `pbench-dispatch` to check two reception areas, rather than one, but otherwise the processing was the same.

The original motivation for `pbench-dispatch` was so that a tarball could be dispatched and worked on simultaneously by a number of scripts: `pbench-unpack-tarball` could unpack it, while at the same time `pbench-index` could be indexing it into ES and `pbench-backup-tarballs` could be backing it up. `pbench-dispatch` did that by creating multiple symlinks, one in each designated directory for the desired script.

Now some scripts can work on the tarball as is (e.g. backup), while others have to unpack it first (e.g unpack -duh - and index). It didn't make sense to unpack it twice, so PRs #1300 and #1306 changed pbench-index to take an already unpacked tarball. That meant that `pbench-dispatch` could not set the state of the tarball to TO-UNPACK **and** TO-INDEX: it could only set the state to TO-UNPACK and TO-BACKUP since they both operate on the tarball and can do so independently of each other. The `pbench-unpack-tarballs` script now has to set the state to TO-INDEX (and any others that depend on the fact that the tarball has been unpacked).

The state transition diagram now looks like this:

img

So there is not much need for `pbench-dispatch`: `pbench-unpack-tarballs` can set the state for any other scripts (other than itself) that operate on the tarball itself  and do not need it unpacked and then unpack the tarballs and set the state for all the scripts that need the tarball to be unpacked. Note that the following diagram is not a pure state transition diagram: it combines a couple of stages of processing with the two state transition diagrams, one for packed tarballs and the other for unpacked ones:

img

The combined script can then be rewritten in python. It's our long term goal to have all the server scripts in python, so we can take advantage of commonalities. The biggest gain in the short run is that, once issue #310 is addressed, we will have a common logging framework that all the python scripts (including this one) can use.

Eventually, we will want to reimplement the state transitions in a more robust form. First, we can write a small abstraction layer that all the scripts will use for their state transition needs, but still using the symlink method underneath, Once we get that layer right, we can rewrite everything underneath it to use a different mechanism (maybe a database and a message passing system but that remains TBD for now). Also eventually, we might decide to merge more scripts together in order to reduce the total number, But both of those will be left for future issues.

riya-17:
Fixes #1315 

This PR deals with Merging the pbench-unpack-tarballs and pbench-dispatch codes and rewriting them in python.
pbench-dispatch deals with sending the tarballs from archive area to their respective state directories.
pbench-unpack-tarballs work on unpacking the tarballs and setting states of state directories.

*This PR needs unittest

### DIFF
--- a/server/bin/pbench-dispatch-unpack.py
+++ b/server/bin/pbench-dispatch-unpack.py
@@ -1,0 +1,348 @@
+#!/usr/bin/env python3
+# -*- mode: python -*-
+
+
+""" 
+# When tarballs are copied to archive area by shim. Then pbench dispatch 
+# set state to know what processing needs to be done to each tarball. it is done by 
+# setting symlinks in designated directories
+
+# those directories are scanned by corresponding scripts
+
+# the work that pbench-dispatch is left for is only to move symlinks from TODO to 
+# TO-UNPACK and TO-BACKUP
+
+# then pbench-unpack-tarball will shift to UNPACKED, TO_INDEX, and any other
+# pbench-unpack script can set the state for first the state which need tarballs itself
+# and then set the state for the one which need unpacked tarballs 
+"""
+
+import os
+import sys
+import glob
+import hashlib
+import shutil
+import tempfile
+import tarfile
+import re
+from contextlib import closing
+from time import time
+from enum import Enum
+from argparse import ArgumentParser
+from s3backup import S3Config, Entry
+
+from pbench import PbenchConfig, BadConfig, get_pbench_logger, md5sum, quarantine
+from pbench.report import Report
+
+_NAME_ = "pbench-dispatch-unpack-tarballs"
+
+class Results(object):
+    def __init__(self, ntotal=0, ntb=0, nerrs=0,
+                 ndups=0, ninfo=0):
+        self.ntotal = ntotal
+        self.ntb = ntb
+        self.nerrs = nerrs
+        self.ndups = ndups
+        self.ninfo = ninfo
+
+def copy_symlinks(state_list, linkdestlist, controller_path, link):
+
+    link_basename = os.path.basename(link)
+
+    totsuc=toterr=0
+
+    for state in linkdestlist:
+        pathh = os.path.join(controller_path,state,"")
+        os.makedirs(pathh, exist_ok=True)
+        os.symlink(link, os.path.join(pathh,link_basename))      
+        if os.path.islink(os.path.join(pathh,link_basename)):
+            totsuc = totsuc + 1
+        else:
+            logger.info("Cannot create {} link to {}".format(link, state))
+            toterr = toterr + 1
+
+    state_list.extend([totsuc, toterr])
+
+    return state_list
+
+def check_md5(config, archive_md5, logger, tb, link):
+
+    qdir = config.QDIR
+
+    # check md5 file exist and it is a regular file
+    if os.path.exists(archive_md5) and os.path.isfile(archive_md5):
+        pass
+    else:
+        # md5 file does not exist or it is not a regular file
+        quarantine(qdir, logger, tb)
+        logger.error(
+            "Quarantine: {}, {} does not exist or it is not a regular file".format(tb, archive_md5))
+        return 0
+
+    # read the md5sum from md5 file
+    try:
+        with open(archive_md5) as f:
+            archive_md5_hex_value = f.readline().split(" ")[0]
+    except Exception:
+        # Could not read file.
+        quarantine(qdir, logger, tb)
+        logger.exception(
+            "Quarantine: {}, Could not read {}".format(tb, archive_md5))
+        return 0
+
+    # match md5sum of the tb to its md5 file
+    try:
+        archive_tar_hex_value = md5sum(link)
+    except Exception:
+        # Could not read file.
+        quarantine(qdir, logger, tb)
+        logger.exception(
+            "Quarantine: {}, Could not read {}".format(tb, link))
+        return 0
+    
+    if archive_tar_hex_value != archive_md5_hex_value:
+        return 1
+
+def dispatch_unpack(config, linkdestlist, unpack_linkdestlist, logger):
+
+    archive = config.ARCHIVE
+    qdir = config.QDIR
+    incoming = config.INCOMING
+    linksrc = "TODO"
+    linkdest = "UNPACKED"
+
+    tblist = glob.iglob(os.path.join(archive,'*',linksrc, '*.tar.xz'))
+
+    ntotal=ntb=nerrs=ndups=ninfo=0
+
+    for tb in tblist:
+
+        ntotal=ntotal+1
+        link = os.path.realpath(tb)
+        archive_md5 = ("{}.md5".format(link))
+
+        md5_res = check_md5(config, archive_md5, logger, tb, link)
+
+        if md5_res == 0:
+            continue
+
+        linksrc_path = os.path.dirname(tb) 
+        tb_linksrc = os.path.basename(linksrc_path)
+        
+        controller_path = os.path.dirname(linksrc_path)
+        controller = os.path.basename(controller_path)
+        controller_dir = os.path.dirname(controller_path)
+
+        resultname = os.path.basename(tb)
+        resultname = resultname[:-7]
+
+        if (linksrc != tb_linksrc):
+            # All is NOT well: we expect $linksrc as the parent directory name
+            # of the symlink tb name.
+            logger.error("Fatal - unexpected linksrc for tb")
+
+        if( archive != controller_dir):
+            # The controller's parent is not $ARCHIVE!
+            logger.error("FATAL - unexpected archive directory for tb")
+
+        if(len(link) == 0):
+            logger.info("Symlink target for tb does not exist")
+            nerrs = nerrs+1
+            quarantine(qdir, logger, tb)
+            continue
+
+        if(resultname[:15] == ("DUPLICATE__NAME")):
+            ndups = ndups +1
+            quarantine(qdir, logger, tb)
+            if os.path.isfile(tb):
+                logger.info("Cannot remove tarball links")
+            continue
+
+        if md5_res == 1:
+            nerrs=nerrs+1
+            quarantine(qdir, logger, tb)
+            logger.error(
+                "Quarantine: {}, md5sum of {} does not match with its md5 file {}".format(tb, link, archive_md5))
+            if os.path.isfile(tb):
+                logger.info("Cannot remove tarball links")
+            continue
+
+        income = incoming
+
+        income=os.path.join(income,controller,resultname)
+        os.makedirs(income)
+        if not os.path.exists(income):
+            nerrs=nerrs+1
+            continue
+
+        income_unpack = income+".unpack"
+        os.makedirs(income_unpack)
+        if not os.path.exists(income_unpack):
+            logger.info("'mkdir {}.unpack' failed for {}".format(resultname, tb))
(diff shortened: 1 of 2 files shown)


server/bin/pbench-dispatch.py
```diff
@@ -0,0 +1,405 @@
+#!/usr/bin/env python3
+# -*- mode: python -*-
+
+
+""" 
+# check directory where tarballs are there
```
ndokos: "... where tarballs are received."

server/bin/pbench-dispatch.py
```diff
@@ -0,0 +1,405 @@
...
+
+# copy tarballs to archive area
+
+# set state to know what processing needs to be done to each tarball. it is done by 
+# setting symlinks in designated directories
+
+# those directories scanned by corresponding scripts
```
ndokos: "... are scanned..."

server/bin/pbench-dispatch.py
ndokos: But I don't think that's right: the shim does that and moves tarballs to the archive. This script figures out what it has to do by scanning the TODO directories in the archive area.

server/bin/pbench-dispatch.py
```diff
@@ -0,0 +1,405 @@
+#!/usr/bin/env python3
+# -*- mode: python -*-
+
+
+""" 
+# check directory where tarballs are there
+
+# copy tarballs to archive area
```
ndokos: No, the shim does that, I believe.

server/bin/pbench-dispatch.py
```diff
@@ -0,0 +1,405 @@
...
+
+# set state to know what processing needs to be done to each tarball. it is done by 
+# setting symlinks in designated directories
+
+# those directories scanned by corresponding scripts
+
+# the work that pbench-dispatch is left for is only to move symlinks from TODO to 
+# TO-UNPACK and TO-BACKUP
+
```
ndokos: Maybe we should get rid of TO-UNPACK: see below.

server/bin/pbench-dispatch.py
```diff
@@ -0,0 +1,405 @@
...
+# then pbench-unpack-tarball will shift to UNPACKED, TO_INDEX, and any other
+# pbench-unpack script can set the state for first the state which need tarballs itself
+# and then set the state for the one which need unpacked tarballs 
+"""
```
ndokos: I'd rename the script to `pbench-dispatch-unpack.py` and not refer to the old names any more. Something like this perhaps:

> This script scans the archive for tar balls in TODO state. For each one, it adds it to the state directories specified in the config option "dispatch-states", unpacks the tarball and after unpacking it also adds it to the state directories specified in the config option "unpack-states". That way, other scripts that operate on the tarball itself can start work immediately: those scripts are the ones that check the state directories specified in "dispatch-states" (currently, the only such state is TO-BACKUP); scripts that depend on the tarball being unpacked, will have to wait until this script unpacks the tar ball and places it in the proper states before they can process it (currently, these are the TO-INDEX and TO-COPY-SOSREPORTS states). For accounting purposes, we also set the state to UNPACKED: a final state which does not trigger any further processing.
```

File: server/bin/pbench-dispatch.py [hunk 6]
```diff
@@ -0,0 +1,405 @@
...
+    subj="$PROG.$TS($PBENCH_ENV) - w/ $nerrs errors"
+    print("subj -----------------------", subj)
+
+    logger.info('end-{}'.format(config.TS))
+
+    return 0
+
+if __name__ == '__main__':
+    status = main()
+    sys.exit(status)
```
ndokos: Missing newline.

server/bin/pbench-dispatch.py
```diff
@@ -0,0 +1,405 @@
...
+        config = PbenchConfig(cfg_name)
+    except BadConfig as e:
+        print("{}: {}".format(_NAME_, e), file=sys.stderr)
+        return 1
+
+    logger = get_pbench_logger(_NAME_, config)
+
+    qdir = config.QDIR = config.get('pbench-server', 'pbench-quarantine-dir')
+
+    lb_obj = LocalConfigObject(config)
```
ndokos: In lines 344-349  you don't need lb_obj. As mentioned above, you can get rid of it in the sanity_check() function and after that, it is not used at all.

server/bin/pbench-dispatch.py
```diff
@@ -0,0 +1,405 @@
...
+
+    lb_obj = sanity_check(lb_obj, config, logger)
+
+    if lb_obj is None:
+        return 3
+
+    linksrc = "TODO"
+    linkedestlist = config.conf.get("pbench-server", "dispatch-states")
+    linkedestlist = linkedestlist.split(", ")
```
ndokos: See the note for line 377 below: it applies here as well.

server/bin/pbench-dispatch.py
```diff
@@ -0,0 +1,405 @@
...
+                             counts.ndups,
+                             counts.ninfo))
+
+    logger.info(result_string)
+
+    subj="$PROG.$TS($PBENCH_ENV) - w/ $nerrs errors"
+    print("subj -----------------------", subj)
+
+    logger.info('end-{}'.format(config.TS))
+
```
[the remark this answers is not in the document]
riya-17: I will take care of that

server/bin/pbench-dispatch.py
```diff
@@ -0,0 +1,405 @@
...
+    if lb_obj is None:
+        return 3
+
+    linksrc = "TODO"
+    linkedestlist = config.conf.get("pbench-server", "dispatch-states")
+    linkedestlist = linkedestlist.split(", ")
+
+    if(linkedestlist == None):
+        print(config.timestamp(), "status : config file error: either no dispatch-states defined or a typo")
```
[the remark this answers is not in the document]
riya-17: ok

server/bin/pbench-dispatch.py
```diff
@@ -0,0 +1,405 @@
...
+                     " Warnings: {}"
+                     .format(counts.ntotal,
+                             counts.ntb,
+                             counts.nerrs,
+                             counts.ndups,
+                             counts.ninfo))
+
+    logger.info(result_string)
+
+    subj="$PROG.$TS($PBENCH_ENV) - w/ $nerrs errors"
```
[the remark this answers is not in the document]
riya-17: later line is the remnant but not the first one.

server/bin/pbench-dispatch.py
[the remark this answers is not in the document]
riya-17: sure

server/bin/pbench-dispatch.py
```diff
@@ -0,0 +1,405 @@
...
+    ntotal = ntb = nerrs = ndups = ninfo = 0
+
+    for tarball in tarlist:
+
+        ntotal = ntotal+1
+
+        res_values = []
+        res_values = to_Get_Values(tarball, res_values)
+
+        linksrc_path, tb_linksrc, controller_path, controller, controller_dir, resultname, link= res_values[0], res_values[1], res_values[2], res_values[3], res_values[4], res_values[5], res_values[6]
```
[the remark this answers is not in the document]
riya-17: I thought there is a large no. of initialization so I built another function. Ok, I will do that

server/bin/pbench-dispatch-unpack.py
```diff
@@ -0,0 +1,348 @@
+#!/usr/bin/env python3
+# -*- mode: python -*-
+
+
+""" 
+# When tarballs are copied to archive area by shim. Then pbench dispatch
```
zulcss, line 6: '#' is not needed

server/bin/pbench-dispatch-unpack.py
```diff
@@ -0,0 +1,348 @@
...
+    def __init__(self, ntotal=0, ntb=0, nerrs=0,
+                 ndups=0, ninfo=0):
+        self.ntotal = ntotal
+        self.ntb = ntb
+        self.nerrs = nerrs
+        self.ndups = ndups
+        self.ninfo = ninfo
+
+def copy_symlinks(state_list, linkdestlist, controller_path, link):
+
```
zulcss, line 49: remove newline

server/bin/pbench-dispatch-unpack.py
```diff
@@ -0,0 +1,348 @@
...
+        else:
+            logger.info("Cannot create {} link to {}".format(link, state))
+            toterr = toterr + 1
+
+    state_list.extend([totsuc, toterr])
+
+    return state_list
+
+def check_md5(config, archive_md5, logger, tb, link):
+
```
zulcss, line 69: ditto

server/bin/pbench-dispatch-unpack.py
```diff
@@ -0,0 +1,348 @@
...
+        quarantine(qdir, logger, tb)
+        logger.exception(
+            "Quarantine: {}, Could not read {}".format(tb, link))
+        return 0
+    
+    if archive_tar_hex_value != archive_md5_hex_value:
+        return 1
+
+def dispatch_unpack(config, linkdestlist, unpack_linkdestlist, logger):
+
```
zulcss, line 107: ditto

server/bin/pbench-dispatch-unpack.py
```diff
@@ -0,0 +1,348 @@
...
+import os
+import sys
+import glob
+import hashlib
+import shutil
+import tempfile
+import tarfile
+import re
+from contextlib import closing
```
zulcss, line 28: Remove not being used

server/bin/pbench-dispatch-unpack.py
```diff
@@ -0,0 +1,348 @@
...
+from time import time
```
zulcss, line 29: you should be able to get by with doing
"from pbench import _time"

server/bin/pbench-dispatch-unpack.py
```diff
@@ -0,0 +1,348 @@
...
+from enum import Enum
```
zulcss, line 30: Remove not being used

server/bin/pbench-dispatch-unpack.py
```diff
@@ -0,0 +1,348 @@
...
+        self.ninfo = ninfo
+
+def copy_symlinks(state_list, linkdestlist, controller_path, link):
+
+    link_basename = os.path.basename(link)
+
+    totsuc=toterr=0
+
+    for state in linkdestlist:
+        pathh = os.path.join(controller_path,state,"")
```
zulcss, line 55: path should be good enough

Issue #249, opened 2016-05-19
Clean up dataset load errors in jschart


k-rister:
When a dataset load error is encountered in jschart an obnoxious "Load Errors" message is prominently displayed in the chart area.

Would like to improve upon this by being more subtle with the error message while giving the user more information about what is wrong (probably add more details to the javascript console including the dataset name).

![foo](https://cloud.githubusercontent.com/assets/15911116/15396297/13ca0ddc-1da1-11e6-894d-802ba3821a7f.png)

k-rister:
Looking into this is revealing some other subtle bugs with empty datasets that need to be fixed as well.

k-rister:
I have a fix that handles this more gracefully.  Datasets that have load errors are now flagged as invalid and hidden.  In the table invalid datasets are marked with a yellow highlight and pinned at the top so that they are clearly visible to the user.  The large error message in the chart body is gone and there are log entries in the Javascript console.  It looks like this now:

![invalid-dataset](https://cloud.githubusercontent.com/assets/15911116/15681036/75aa5946-271d-11e6-8279-43c851c01ad2.png)

While invalid datasets will exist in both the table and the legend many of the traditional user initiated actions (highlighting, unhiding, etc.) are ignored.

portante:
@k-rister, really nice!

Issue #217, opened 2016-04-20
Ensure that the crontab on the server uses the specified temp directory.


ndokos:
That's a config file issue but we need to verify that the temp directory argument is in the crontab file.

portante:
We have seen some evidence of a "sed" temp file in the archive directory.  We should make sure server side scripts that use sed have a temp directory properly set for those scripts.

We found some evidence that the `crontab` file needs to have the configured temporary directory properly specified so that the scripts pick that up.

PR #200, opened 2016-03-17
Modify the sar script (and its relatives) to handle user-defined opti…

ndokos:
…ons.

We now honor any options that are passed in when the tool is registered,
e.g. with

 pbench-register-tool --name=pidstat -- --options="-R -U"

However the postprocessing is skipped, since it is virtually certain
that the output is going to change and the postprocessing script is
going to break. It is expected that a user will provide his/her own
postprocessing in that case.

See issue #168.

Issue #173, opened 2016-02-17
turbostat tool fails when run inside a KVM guest


jeremyeder:
turbostat tool is part of the default tool-set.  But when pbench is used inside a KVM guest turbostat cannot/will not run.

Actually it's not a huge issue because it only shows up as a "cant find a turbostat pid to kill" when user-benchmark ends.

```
# dmidecode|grep KVM
	Product Name: KVM
# turbostat -i1
turbostat: /dev/cpu/7/msr offset 0xe2 read failed: Input/output error
```

turbostat datalog needs to be conditionalized to only run when not in KVM.

What's funny is that turbostat runs OK inside Xen/EC2, but I don't know if we can trust the data.

```
# dmidecode|grep Product
	Product Name: HVM domU
root@ip-172-31-2-226: ~ # turbostat sleep 3
     CPU TSC_MHz     SMI  CPU%c1  CPU%c3  CPU%c6  CPU%c7 Pkg%pc2 PkgWatt RAMWatt   PKG_%   RAM_%
       -    2394       0  100.00    0.00    0.00    0.00    0.00   51.44    8.25    0.00    0.00
       0    2394       0  100.00    0.00    0.00    0.00    0.00   51.44    8.25    0.00    0.00
       1    2394       0  100.00
3.009719 sec
```

atheurer:
Let's add some kind of test for turbostat --install, so when register-tool is called, it will not be added when that turbostat --install call fails

jeremyeder:
#175

ndokos:
PR #175 has been merged. Closing this issue.

PR #1354, opened 2019-09-23
cleanup-log-directory

### LINKED ISSUE #975: Fix sync-satellite to clean up log directories when no errors
Currently sync-satellite leaves the timestamped log directory in place after it runs, which will lead to many timestamped directories being created.

riya-17:
Fixes #975 

This PR works on cleaning the logs directory if there are no errors.

### DIFF
--- a/server/bin/pbench-sync-satellite.sh
+++ b/server/bin/pbench-sync-satellite.sh
@@ -77,6 +77,8 @@ ssh to the $remote_host was sucessfull for changing the state of tarballs,
 but failed to remove local "${state_change_log}".
 EOF
             pbench-report-status --name ${PROG} --pid ${$} --timestamp $(timestamp) --type error ${index_content}
+        else
+            rm ${logdir}/mv.log
         fi
     fi
     return $status
@@ -250,7 +252,19 @@ for host in $hosts ;do
         for x in $state_list; do
             echo "$host/TO-SYNC/$x" >> ${state_change_log}
         done
+        rm $logdir/$host/ok-checks.log
     fi
+
+    if [[ ! -s $logdir/$host/fail-checks.log ]] ; then
+        rm $logdir/$host/fail-checks.log
+        grep -q -F -v 'OK' $logdir/$host/md5-checks.log
+        if [[ $? -ne 0 ]]; then
+            rm $logdir/$host/md5-checks.log
+        fi
+    fi
+
+    rmdir ${logdir}/${host} 2>/dev/null
+
     popd > /dev/null 2>&4
 done
 
--- a/server/bin/gold/test-11.txt
+++ b/server/bin/gold/test-11.txt
@@ -137,8 +137,6 @@ drwxrwxr-x          - logs/pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC
 drwxrwxr-x          - logs/pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/controller
 -rw-rw-r--         60 logs/pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/controller/fail-checks.log
 -rw-rw-r--         96 logs/pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/controller/md5-checks.log
--rw-rw-r--         36 logs/pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/controller/ok-checks.log
--rw-rw-r--          0 logs/pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/mv.log
 -rw-rw-r--        218 logs/pbench-sync-satellite/pbench-sync-satellite.error
 -rw-rw-r--       2641 logs/pbench-sync-satellite/pbench-sync-satellite.log
 drwxrwxr-x          - pbench-move-results-receive
@@ -212,11 +210,6 @@ pbench-user-benchmark_38_2016-05-18_19:36:32.tar.xz: FAILED
 fio__2016-08-16_22:03:11.tar.xz: OK
 pbench-user-benchmark_38_2016-05-18_19:36:32.tar.xz: FAILED
 ----- pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/controller/md5-checks.log
-+++++ pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/controller/ok-checks.log
-fio__2016-08-16_22:03:11.tar.xz: OK
------ pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/controller/ok-checks.log
-+++++ pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/mv.log
------ pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/mv.log
 +++++ pbench-sync-satellite/pbench-sync-satellite.error
 pbench-sync-satellite: completing satellite state changes ... (/var/tmp/pbench-test-server/test-11/pbench-local/logs/pbench-sync-satellite/ONE/change_state.log)
 pbench-sync-satellite: completed satellite state changes
--- a/server/bin/gold/test-5.2.txt
+++ b/server/bin/gold/test-5.2.txt
@@ -641,20 +641,6 @@ drwxrwxr-x          - logs/pbench-server-prep-shim-002
 -rw-rw-r--        943 logs/pbench-server-prep-shim-002/pbench-server-prep-shim-002.log
 drwxrwxr-x          - logs/pbench-sync-satellite
 drwxrwxr-x          - logs/pbench-sync-satellite/ONE
-drwxrwxr-x          - logs/pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC
-drwxrwxr-x          - logs/pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/controllerA
--rw-rw-r--          0 logs/pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/controllerA/fail-checks.log
--rw-rw-r--         47 logs/pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/controllerA/md5-checks.log
--rw-rw-r--         47 logs/pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/controllerA/ok-checks.log
-drwxrwxr-x          - logs/pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/controllerB
--rw-rw-r--          0 logs/pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/controllerB/fail-checks.log
--rw-rw-r--         47 logs/pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/controllerB/md5-checks.log
--rw-rw-r--         47 logs/pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/controllerB/ok-checks.log
-drwxrwxr-x          - logs/pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/controllerC
--rw-rw-r--          0 logs/pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/controllerC/fail-checks.log
--rw-rw-r--         54 logs/pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/controllerC/md5-checks.log
--rw-rw-r--         54 logs/pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/controllerC/ok-checks.log
--rw-rw-r--          0 logs/pbench-sync-satellite/ONE/run-1970-01-01T00:00:00-UTC/mv.log
 -rw-rw-r--        219 logs/pbench-sync-satellite/pbench-sync-satellite.error
 -rw-rw-r--       2586 logs/pbench-sync-satellite/pbench-sync-satellite.log
 drwxrwxr-x          - logs/pbench-unpack-tarballs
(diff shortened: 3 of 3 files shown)


server/bin/pbench-sync-satellite.sh
```diff
@@ -266,6 +266,13 @@ if [ -s ${state_change_log} ] ;then
     fi
 fi
 
+if [ $nerrs == 0 ] ; then
+    for entry in "${logdir_for_remote}"/*
```
portante: Typically, we get the list of files like this and store it in a file, and then read from the file.  Can you look at other examples in the code that follow that pattern?

server/bin/pbench-sync-satellite.sh
```diff
@@ -266,6 +266,13 @@ if [ -s ${state_change_log} ] ;then
...
+    do
```
portante: We have typically been keeping the `do` on the same line as the `for` with a semi-colon (`;`) separating them.

server/bin/pbench-sync-satellite.sh
```diff
@@ -266,6 +266,13 @@ if [ -s ${state_change_log} ] ;then
     fi
 fi
 
+if [ $nerrs == 0 ] ; then
```
portante: Can you use the double-bracket form here?  Something like, `if [[ ${nerrs} == 0 ]]; then`?

server/bin/pbench-sync-satellite.sh
riya-17: Sure, will look into it

server/bin/pbench-sync-satellite.sh
riya-17: ok

server/bin/pbench-sync-satellite.sh
riya-17: Sure :)

server/bin/pbench-sync-satellite.sh
riya-17: @portante May I know why we are copying them first and then deleting instead of deleting the directly?

portante: @riya-17, basically line 270 in bash causes the shell to expand all the files that match the pattern, `${logdir_for_remote}/*`, which can potentially be very large.  The shell has a limit to the number of characters in the expansion.  Also, to aid debugging, writing the list of items to be deleted can help us debug the behaviors live.

server/bin/pbench-sync-satellite.sh
```diff
@@ -266,6 +266,13 @@ if [ -s ${state_change_log} ] ;then
...
+    for entry in "${logdir_for_remote}"/*
+    do
+        rm -rf $entry
```
portante: Upon further review of the existing code, I don't think we want to add this blanket recursive `rm` command.

If we review the code, we should see that the `${state_change_log}` is a file that lives in the `${logdir_for_remote}/` directory, and only removed if we were successfully able to remotely change the state on the satellite.  If `pbench-sync-satellite.sh` fails to remotely change the state, we need to leave this file here so that the next execution will attempt to finish before proceeding.

The only other contents of the `${logdir_for_remote}` is the `${logdir}`, which is set to `${logdir_for_remote}/${TS}`.  We already have a `trap` which tries to clean that directory up, so if that is not happening, we need to debug why first.

server/bin/pbench-sync-satellite.sh
riya-17: Ok, got it. Thank you

server/bin/pbench-sync-satellite.sh
riya-17: @portante If pbench-sync-satellite.sh fails to change the state on the satellite than the error will be thrown and the removal will not take place. So, there will be no need for executing this script. It happened to be executed if every task is fulfilled successfully.

riya-17:
@ndokos Please Take a Look

server/bin/pbench-sync-satellite.sh
```diff
@@ -266,6 +266,10 @@ if [ -s ${state_change_log} ] ;then
     fi
 fi
 
+if [[ ${nerrs} == 0 ]] ; then
+    find "${logdir_for_remote}"/* -name "*-UTC" -type d -print0|xargs -0 rm -r --
```
portante: This does not seem right still.

The main loop for the satellite is at [line #148](https://github.com/distributed-system-analysis/pbench/pull/1354/files#diff-2b119fd9bfaa54cda0c0fbdad8ae2257R148) where it loops over the list of hosts (really, pbench "`controller`" names), and executes a sequence of tasks.

At [line #173](https://github.com/distributed-system-analysis/pbench/pull/1354/files#diff-2b119fd9bfaa54cda0c0fbdad8ae2257R173), we create a directory in the `${logdir}` (which is just `${logdir_for_remote}/${TS}`) for the given `host`, and we proceed to create 3 (three) files in that directory:

 1. `md5-checks.log`, @ [line #200](https://github.com/distributed-system-analysis/pbench/pull/1354/files#diff-2b119fd9bfaa54cda0c0fbdad8ae2257R200)
 2. `ok-checks.log`, @ [line #203](https://github.com/distributed-system-analysis/pbench/pull/1354/files#diff-2b119fd9bfaa54cda0c0fbdad8ae2257R203)
 3. `fail-checks.log`, # [line #217](https://github.com/distributed-system-analysis/pbench/pull/1354/files#diff-2b119fd9bfaa54cda0c0fbdad8ae2257R217)

Then we can see that at [line #249](https://github.com/distributed-system-analysis/pbench/pull/1354/files#diff-2b119fd9bfaa54cda0c0fbdad8ae2257R249) we essentially copy the contents of `ok-checks.log` into the `${state_change_log}` file.

So we should be able to do the following to clean things up better:

 1. Always remove the `ok-checks.log` file at the end of each iteration
 2. Remove the `fail-checks.log` file if it is empty
 3. Add a third `grep` of the `md5-checks.log` file to look for all the lines NOT containing `OK` or `FAILED`, and removing `md5-checks.log` if there were no matching lines
 4. Add an `rmdir ${logdir}/${host}` which will remove the directory for the given host if there were no errors

Can you try that out instead?

riya-17: Ok.. sure

riya-17:
@ndokos @portante Please take a Look

riya-17:
@ndokos PTAL

server/bin/pbench-sync-satellite.sh
```diff
@@ -250,10 +250,30 @@ for host in $hosts ;do
...
             echo "$host/TO-SYNC/$x" >> ${state_change_log}
         done
+        rm $logdir/$host/ok-checks.log
     fi
+
+    if [[ ! -s $logdir/$host/fail-checks.log ]] ; then
+        rm $logdir/$host/fail-checks.log
+    fi
+
+    if [[ -z $(grep -E -v 'OK|FAILED' $logdir/$host/md5-checks.log) ]] ; then
```
portante: If processing all the tar balls for this `${host}` is successful, then `fail-checks.log` will be empty because [line 217](https://github.com/distributed-system-analysis/pbench/pull/1354/files#diff-2b119fd9bfaa54cda0c0fbdad8ae2257R217) did not find any lines with "`FAILED`" in the `md5-checks.log` file.  Given that logic, then the only time [line 260](https://github.com/distributed-system-analysis/pbench/pull/1354/files#diff-2b119fd9bfaa54cda0c0fbdad8ae2257R260) will evaluate to `True` is when the `md5sum` command on [line 200](https://github.com/distributed-system-analysis/pbench/pull/1354/files#diff-2b119fd9bfaa54cda0c0fbdad8ae2257R200) emits text lines that do not have "`OK`" or "`FAILED`" on the line.

Rather than have a separate check, since the only time we would want to remove that file is if there are no "`FAILED`" checks, can we consider moving this logic to the same `then` clause of [line 257](https://github.com/distributed-system-analysis/pbench/pull/1354/files#diff-2b119fd9bfaa54cda0c0fbdad8ae2257R257)? .

server/bin/pbench-sync-satellite.sh
```diff
@@ -250,10 +250,30 @@ for host in $hosts ;do
...
 done
 
+if [[ -s ${logdir}/mv.log ]] ; then
+    rm ${logdir}/mv.log
+fi
+
+rmdir ${logdir}/${host} 2>/dev/null
+sts = ${?}
+if [[ ${sts}!=0 ]]; then
+    echo "$PROG: rmidir failed to remove ${logdir}/${host} - status ${sts}" &>4
```
portante: s/rmidir/rmdir

portante: s/&>4/>&4/

server/bin/pbench-sync-satellite.sh
```diff
@@ -250,10 +250,30 @@ for host in $hosts ;do
...
+    fi
+
+    if [[ -z $(grep -E -v 'OK|FAILED' $logdir/$host/md5-checks.log) ]] ; then
+        rm $logdir/$host/md5-checks.log
+    fi
+
     popd > /dev/null 2>&4
 done
 
+if [[ -s ${logdir}/mv.log ]] ; then
```
portante: Can we consider moving this check and removal of `mv.log` to the else block of [line 69](https://github.com/distributed-system-analysis/pbench/pull/1354/files#diff-2b119fd9bfaa54cda0c0fbdad8ae2257R69)?

server/bin/pbench-sync-satellite.sh
ndokos: The check here is too complicated: run grep and get its output and then check if the output is the empty string. `grep` has an option for checking if there is a match or not, without producing any output: `grep -q -E -v 'OK|FAIL' <somefile>` returns success if `<somefile>` contains something other than OK or FAIL, failure otherwise. So the test above can be written more simply and more efficiently like this:

    if grep -q -E -v 'OK|FAIL' $logdir/$host/md5-checks.log ;then
         ....

That returns as soon as the answer is known instead of going through the whole file, it avoids having to gather the output of `grep` intro a string in order to check whether it is the empty string and it avoids the explicit test. But the main thing it achieves is simplicity and readability.

riya-17:
@ndokos @portante PTAL

server/bin/pbench-sync-satellite.sh
```diff
@@ -250,10 +253,26 @@ for host in $hosts ;do
...
             echo "$host/TO-SYNC/$x" >> ${state_change_log}
         done
+        rm $logdir/$host/ok-checks.log
+    fi
+
+    if [[ ! -s $logdir/$host/fail-checks.log ]] ; then
+        rm $logdir/$host/fail-checks.log
+    fi
+
+    if [[ -z $(grep -q -E -v 'OK|FAILED' $logdir/$host/md5-checks.log) ]] ; then
```
ndokos: Please reread my comment about how to simplify this: adding -q is *NOT* the only thing you have to do; in fact, since it eliminates the output, the -z test will *always* succeed: that is wrong.

server/bin/pbench-sync-satellite.sh
```diff
@@ -250,10 +253,26 @@ for host in $hosts ;do
...
+        rm $logdir/$host/md5-checks.log
     fi
+
     popd > /dev/null 2>&4
 done
 
+rmdir ${logdir}/${host} 2>/dev/null
+sts = ${?}
+if [[ ${sts}!=0 ]]; then
+    echo "$PROG: rmdir failed to remove ${logdir}/${host} - status ${sts}" s/&>4/>&4/
```
portante: The characters at the end of the line are not valid `bash` syntax: `s/&>4/>&4/`.

The previous comment where that was written was intended to be short hand for substituting "`&>4`" with "`>&4`", the latter being the correct `bash` shell syntax for writing the text of the `echo` command to file descriptor `#4`.

server/bin/pbench-sync-satellite.sh
riya-17: @ndokos it Exits immediately with zero status if any match is found, even if an error was detected. So, it is not possible to use -q option.

server/bin/pbench-sync-satellite.sh
```diff
@@ -250,10 +253,26 @@ for host in $hosts ;do
...
+        rm $logdir/$host/ok-checks.log
+    fi
+
+    if [[ ! -s $logdir/$host/fail-checks.log ]] ; then
+        rm $logdir/$host/fail-checks.log
+    fi
+
+    if [[ -z $(grep -E -v -m 1 'OK|FAILED' $logdir/$host/md5-checks.log) ]] ; then
+        rm $logdir/$host/md5-checks.log
     fi
```
portante:
```
    if [[ ! -s $logdir/$host/fail-checks.log ]] ; then
        rm $logdir/$host/fail-checks.log
        # Look for all lines in md5-checks.log that DO NOT match "OK", where
        # "grep" will return 0 when it does find such lines, and non-zero when
        # it does not find such lines.
        grep -q -F -v 'OK' $logdir/$host/md5-checks.log
        if [[ $? -ne 0 ]]; then
            # We can safely remove md5-checks.log because grep DID NOT find
            # any lines other than those that contained "OK".
            rm $logdir/$host/md5-checks.log
        fi
    fi
```

server/bin/pbench-sync-satellite.sh
```diff
@@ -250,10 +253,26 @@ for host in $hosts ;do
...
+
     popd > /dev/null 2>&4
 done
 
+rmdir ${logdir}/${host} 2>/dev/null
```
portante: Let's move this inside the `for host in ${hosts}; do` loop before line 267.

portante: And just the `rmdir` needs to move, the check for success or failure can be dropped because failure just means the directory contains log files we want to keep.

riya-17:
@ndokos @portante PTAL

server/bin/pbench-sync-satellite.sh
```diff
@@ -78,6 +78,9 @@ but failed to remove local "${state_change_log}".
 EOF
             pbench-report-status --name ${PROG} --pid ${$} --timestamp $(timestamp) --type error ${index_content}
         fi
+        if [[ -s ${logdir}/mv.log ]] ; then
```
portante: Lines 81 thru 83, are not quite right, it seems.  It currently reads as, "if we have a `${logdir}/mv.log` of some non-zero size, remove it."  This means that if the log file is empty, we won't remove it, and so we won't be able to remove the `${logdir}` directory.

Instead, if we change lines 80 thru 83 to be:

```
        else
            # The removal of ${state_change_log} was successful, so we can attempt to remove
            # the ${logdir}mv.log file.
            rm ${logdir}/mv.log
        fi
```

ndokos:
@gurbirkalsi: The e2e tests failed. I restarted the build just in case it was a fluke.

PR #1359, opened 2019-09-24
Reference @metadata.controller_dir when querying result.json files

gurbirkalsi:
**Summary**

Enforces the use of the `@metadata.controller_dir` field across run comparison initiation points (Result and Search views). Selecting a result row from the Result page now finds the result `key` property for building the `selectedResults` data structure within the `dashboard` Redux model.

### DIFF
--- a/web-server/v0.4/mock/api.js
+++ b/web-server/v0.4/mock/api.js
@@ -74,6 +74,11 @@ export const mockMappings = {
               user: { type: 'string', index: 'not_analyzed' },
             },
           },
+          '@metadata': {
+            properties: {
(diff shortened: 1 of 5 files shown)


portante:
@gurbirkalsi, looks like the e2e tests failed.  Can you take a look?

gurbirkalsi:
I've included the enhancements to the Search page component e2e tests as a separate commit. We're now using pattern matching against the mocked API data for verifying workflows within this page component.

portante:
Will merge once the TravisCI completes successfully.

PR #1360, opened 2019-09-25
Fix audit server code to handle --* controller names

portante:


server/bin/pbench-audit-server.sh
```diff
@@ -689,7 +689,7 @@ function verify_archive {
         rm -f ${tarballs}
 
         if [ -s ${lclreport} ]; then
-            printf "\nController: $(basename ${controller})\n"
+            printf "\nController: $(basename -- ${controller})\n"
```
ndokos: I did a grep of (dir|base)name and counted 13 instances. I count only 10 fixes above. Can you check?

PR #1369, opened 2019-09-26
Short-term fix for missing pb-m-r user/prefix args

portante:
As of PR https://github.com/distributed-system-analysis/pbench/pull/1289, specifically line [177](https://github.com/distributed-system-analysis/pbench/pull/1289/files?file-filters%5B%5D=No+extension#diff-d7362c40121612a2669161b65d94590aR177), any invocation of `pbench-move|copy-results` that did not use both `--user` and `--prefix` always received an unwanted "`user`", and sometimes got the wrong "`--prefix`".

 * If you did not use either `--user` or `--prefix`, then your "`user`" was being set to "`--prefix`"

 * If you used only `--user`, then your "`user`" was correct, but your "`prefix`" was being set to "`--xy-singled-threaded`"

 * If you used only `--prefix`, then your "`user`" was being set to "`--prefix`", and your "`prefix`" was being ignored

If you used both `--user` and `--prefix` then your "`user`" and "`prefix`" were being set as expected.

This PR corrects the invocation of `pbench-make-result-tb` from `pbench-move-results`.


agent/util-scripts/pbench-make-result-tb
```diff
@@ -29,50 +35,57 @@ prefix=""
...
             fi
+            shift;
             ;;
         --prefix)
-            shift;
             if [[ -n "${1}" ]]; then
                 prefix="${1}"
-                shift;
             fi
+            shift;
```
ndokos: I don't understand: why are the first two treated differently from the next two? It seems to me that the treatment should be the same in all four cases.

agent/util-scripts/pbench-move-results
```diff
@@ -174,7 +174,7 @@ for dir in `/bin/ls -ort -d */ | awk '{print $8}' | grep -v "^tools-" | grep -v
     # contain the tarball and the md5 file (as ${tb}.tar.xz.md5.check); pass
     # that information on to pbench-make-result-tb.
     result_dir=$(basename ${dir})
-    result_tb_name=$(pbench-make-result-tb --result-dir ${result_dir} --target-dir ${tmp}/${controller} --user ${user} --prefix ${prefix} --xz-single-threaded ${xz_single_threaded})
+    result_tb_name=$(pbench-make-result-tb --result-dir ${result_dir} --target-dir ${tmp}/${controller} --user "${user}" --prefix "${prefix}" --xz-single-threaded ${xz_single_threaded})
```
ndokos: Why are the rest of the variables not quoted?

agent/util-scripts/pbench-make-result-tb
portante: Because `--user` and `--prefix` are optional values provided by our caller, `pbench-move-results`, where-as the other parameters are required values for operation.

agent/util-scripts/pbench-move-results
portante: I'll add the quotes, but they are not necessary because `pbench-move-results` will always have a value for them.  I was trying to keep the changes to the absolute minimum.

portante: Added.

PR #1379, opened 2019-10-01
Fix cli args

### LINKED ISSUE #1358: Parsing command line options for agent CLI commands does not handle missing arguments
The way we currently handle parsing of pbench-agent CLI command line options fails to catch cases where options are missing requiring arguments, but followed by another option.

For example, `pbench-move-results --user --prefix=foo --xz-single-threaded` will result in the `user` argument being set to "`--prefix=foo`".

portante:
Fixes #1358.

Broken into separate commits to make it easier to review.

### DIFF
--- a/agent/bench-scripts/pbench-cyclictest
+++ b/agent/bench-scripts/pbench-cyclictest
@@ -47,7 +47,7 @@ function usage {
 }
 
 # Process options and arguments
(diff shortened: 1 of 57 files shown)


portante:
I started working on this further and find that fixing this is very tedious because we have so much repeated code (see the discussions in #1206, #1322, #1327).

We really need to solve the core "DRY principle violation" problem before solving this.  With the removal of such repeated code we would have a chance of properly testing these changes.

portante:
@zulcss, this is a PR the highlights the need for the re-write in Python 3.

PR #1381, opened 2019-10-02
Do not use PBenchData.make_source_id() for generating source IDs.

ndokos:
This is a partial implementation of the above for proc-interrupts data
*only*. It is meant as a point of discussion only. If the approach
seems valid, I will extend it to the rest.


ndokos:
BTW, this is completely untested at the moment (it will fail unit tests of course).

ndokos:
Extending it to nested dictionaries should be straightforward, but I'm not sure about the last question: MD5 is non-commutative, so we'd have to calculate it in *exactly the same order* as we did the first time, so that might be problematic.

portante:
> Extending it to nested dictionaries should be straightforward, but I'm not sure about the last question: MD5 is non-commutative, so we'd have to calculate it in _exactly the same order_ as we did the first time - that might be problematic.

Right, we need to define the algorithm and document it.

> There are two questions I think: if we have a dictionary and later on reconstruct a dictionary with the same components, when we traverse the latter dictionary will we get all the elementary components in the same order?

That has to be part of the algorithm definition for generating a source ID: all key, value pairs are processed in sorted order of the keys (defined with the proper `LOCALE` as well for sort semantics).

>  and if we round trip a dictionary through `json.dumps()` to ES and back again, will we get the "same" dictionary (i.e. traversing the dictionary will give us the same elementary components in the same order)?

Not sure what you mean by this.  If we have a proper definition of the order of how keys are processed, and how values are transformed (numbers left as is, strings/bytes properly UTF-8 encoded), then won't that make it repeatable?

PR #1382, opened 2019-10-04
Fix proc int indexing

portante:
This is just an FYI about what we can change to improve indexing, provides a hacked together, `indexer-driver.py` which will let us run indexing in a more targeted manner.

This is not intended to be merged.


portante:
No longer needed.

PR #1384, opened 2019-10-08
pbench-trafficgen: point to the updated trafficgen repository location/name

k-rister:
- The repository name was changed a long time ago but the previous
  name continued to work for backwards compability.  Recently this has
  caused some confusion so go ahead and update the code to be
  consistent with reality.


k-rister:
The E2E test is failing due to some TravisCI/Ubuntu error...I'll keep retrying it I guess...

portante:
Looks good; would merge, but wanted to make sure @atheurer signed off before I merged since he is a requested reviewer.

PR #1388, opened 2019-10-17
Preserve search state to enable multi-controller search

gurbirkalsi:
**Summary**

Selected runs within the Search page component are now "preserved" within the component state when the user executes a search query and later "committed" when the user initiates a run comparison or executes a new search query. 

In addition, the search filters aggregated via the Elasticsearch mappings have been moved under the search input in order to provide enhanced viewability of the search results table.


gurbirkalsi:
Resolves issue #1388

PR #1390, opened 2019-10-22
metalog-%-error

### LINKED ISSUE #1329: pbench-add-metalog-option blows up when the option is a '%'
```
pbench-run-benchmark trafficgen <other options> --rate-unit=%
...
  File "/opt/pbench-agent/util-scripts/pbench-add-metalog-option", line 32, in <module>
    status = main(lfile, section, option)
  File "/opt/pbench-agent/util-scripts/pbench-add-metalog-option", line 22, in main
    config.set(section, option, ', '.join(sin.split()))
  File "/usr/lib64/python3.6/configparser.py", line 1193, in set
    super().set(section, option, value)
  File "/usr/lib64/python3.6/configparser.py", line 894, in set
    value)
  File "/usr/lib64/python3.6/configparser.py", line 402, in before_set
    "position %d" % (value, tmp_value.find('%')))
ValueError: invalid interpolation syntax in '%' at position 0
```

The `%` is a valid value for trafficgen's --rate-unit parameter.

riya-17:
Fixes #1329 

The problem is that configparser Interpolation (https://docs.python.org/3/library/configparser.html#configparser.InterpolationError) uses `%` as a marker for the interpolation construct `%(option)s` which replaces the construct by the value of the option. Configparser Interpolation does not accept the `%` symbol in any other context unless it is escaped as a double percent sign. So when we write a value with '%', ex: "30%" in any config file, this error in thrown. To fix this error, whenever the user enters a value that includes a  `%` char,  it will be converted into `%%` (double percent sign) which interpolation treats as a literal single percent sign.

### DIFF
--- a/agent/util-scripts/gold/eval/test-48.txt
+++ b/agent/util-scripts/gold/eval/test-48.txt
@@ -1,0 +1,15 @@
++++ Running test-48 eval pbench-add-metalog-option /var/tmp/pbench-test-utils/pbench/metadata.log run percent < /var/tmp/pbench-test-utils/pbench/tmp/foo.txt
+--- Finished test-48 eval (status=0)
++++ pbench tree state
+/var/tmp/pbench-test-utils/pbench
+/var/tmp/pbench-test-utils/pbench/metadata.log
+/var/tmp/pbench-test-utils/pbench/tmp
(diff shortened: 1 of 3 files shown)


ndokos:
Looks good. Can you please add a simple unit test for this functionality?

PR #1394, opened 2019-10-29
add timestamp for tarball

### LINKED ISSUE #1391: Enhance pbench-make-result-tb to record a creation timestamp (UTC) in the metadata.log file
It would be very helpful to the server going forward to track the timestamp of when the tar ball was created in UTC.  We don't want to rely on knowing the timezone and then using the timestamps in the tar ball.  By recording the timestamp explicitly in the `metadata.log` file we remove the guess-work on the server side.

riya-17:
Fixes #1391 

Adds timestamp for the time tarball is created

### DIFF
--- a/agent/util-scripts/gold/pbench-copy-results/test-31.txt
+++ b/agent/util-scripts/gold/pbench-copy-results/test-31.txt
@@ -322,6 +322,7 @@ config = ndk-test-1
 date = 2018-05-23T03:21:32
 rpm-version = 0.50-1g799ea025
 iterations = 1, 1
+tar-ball-creation-timestamp = 1900-01-01T00:00:00.000000
 
 [tools]
 hosts = alphaville
--- a/agent/util-scripts/pbench-make-result-tb
+++ b/agent/util-scripts/pbench-make-result-tb
@@ -162,6 +162,9 @@ result_size=$(du -sb ${pbench_run_name} | awk '{print $1}')
 debug_log "preparing to tar up ${result_size} bytes of data from ${full_result_dir}"
 printf -- "$result_size" | pbench-add-metalog-option ${mdlog} run raw_size
 
+ts=$(timestamp)
+printf -- "%s" "$ts" | pbench-add-metalog-option ${mdlog} pbench tar-ball-creation-timestamp
+
 tarball="${target_dir}/${pbench_run_name}.tar.xz"
 if [[ "${xz_single_threaded}" == "1" ]]; then
     tar_cmd="tar --create --force-local -xz \"${pbench_run_name}\""
--- a/agent/util-scripts/gold/pbench-copy-results/test-39.txt
+++ b/agent/util-scripts/gold/pbench-copy-results/test-39.txt
@@ -322,6 +322,7 @@ config = ndk-test-1
 date = 2019-09-27T14:21:31
 rpm-version = 0.50-1g799ea025
 iterations = 1, 1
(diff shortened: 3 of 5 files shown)


agent/util-scripts/pbench-make-result-tb
```diff
@@ -162,6 +162,14 @@ result_size=$(du -sb ${pbench_run_name} | awk '{print $1}')
...
 printf -- "$result_size" | pbench-add-metalog-option ${mdlog} run raw_size
 
+times=$(date -u "+%Y-%m-%dT%H:%M:%S-%Z")
+if [[ -z "$_PBENCH_BENCH_TESTS" ]]; then
+	# use ns in the timestamp
+	printf "${times}" | pbench-add-metalog-option ${mdlog} pbench timestamp
+else
+	printf "1900-01-01T00:00:00.000000" | pbench-add-metalog-option ${mdlog} pbench timestamp
+fi
+
```
ndokos: No, we don't want to put *any* knowledge about unit tests in the scripts themselves. All the scripts source the base script: the base script is the *only one* that knows about unit tests, just so that no other script needs to know. All that *this* script needs to do is to use the `timestamp` function that is defined in the base script: that function will return a fake timestamp if running in the unit test regime or a real one otherwise - this script does not have to worry about that:

       ts="$(timestamp)"
       printf -- "%s" "$ts"  | pbench-add-metalog-option "${mdlog}" pbench timestamp

is all that's needed here. BTW, you want to use the `printf -- "%s" ....` idiom, so that we avoid troubles with % signs (see your other PRs :-) ).

agent/util-scripts/gold/pbench-copy-results/test-31.txt
```diff
@@ -322,6 +322,7 @@ config = ndk-test-1
 date = 2018-05-23T03:21:32
 rpm-version = 0.50-1g799ea025
 iterations = 1, 1
+timestamp = 1900-01-01T00:00:00.000000
```
ndokos: I think we should use a name for the option that is less ambiguous than `timestamp`: I suggest `tar-ball-creation-timestamp`.

agent/util-scripts/pbench-make-result-tb
riya-17: @ndokos I have tried this earlier but it is not giving any output when run.
```timestamp =                                 ```

File: agent/util-scripts/pbench-make-result-tb [hunk 3]
```diff
@@ -162,13 +162,8 @@ result_size=$(du -sb ${pbench_run_name} | awk '{print $1}')
...
 
-times=$(date -u "+%Y-%m-%dT%H:%M:%S-%Z")
-if [[ -z "$_PBENCH_BENCH_TESTS" ]]; then
-	# use ns in the timestamp
-	printf "${times}" | pbench-add-metalog-option ${mdlog} pbench timestamp
-else
-	printf "1900-01-01T00:00:00.000000" | pbench-add-metalog-option ${mdlog} pbench timestamp
-fi
+ts=$(timestamp)
+	printf "$ts" | pbench-add-metalog-option ${mdlog} tar-ball-creation-timestamp
```
ndokos:
```
printf -- "%s" "$ts}" |  ....
```
and no indentation is needed or desired.

ndokos: You forgot the section name (the second argument to pbench-add-metalog-option).

PR #1395, opened 2019-10-29
Fix pbench-make-result-tb errors on arguments containing '%'.

### LINKED ISSUE #1392: `pbench-make-result-tb` errors when arguments contain %.
% signs are legal in filenames and they are legal as options to some commands (e.g trafficgen - see PR #1390 and the associated issue #1329). PR #1390 is fixing the issue for options added to the metadata log through `pbench-add-metalog-option`. We we were going to test that by passing a `--prefix` or a `--user` option containing a % sign to `pbench-move/copy-results`, but `pbench-make-result-tb` chokes on that. The reason is that it uses

    printf -- "${prefix} | pbench-add-metalog-option ${mdlog} run prefix

and similar, but if the argument to `printf` contains a '%', that's an error:

> $ printf -- "30%"
bash: printf: `%': missing format character

We should change the `printf` calls to

    printf -- "%s" "${prefix}" | pbench-add-metalog-option ....

In addition, there are problems if the result name contains a '%' (e.g. if the user uses a `--config` option with a value containing a '%'). A similar remedy will fix that case as well, but one gotcha is
when the value to be printed contains a newline, e.g. the penultimate line of `pbench-make-result-tb`:

    printf -- "${tarball}\n"

which should be rewritten as:

   printf -- "%s\n" "${tarball}"

Finally, a change will be needed to `pbench-metadata-log` to make sure that when it adds the result name to the `metadata.log` file, it escapes any '%' signs, else the configparser will choke (as it did in issue #1329).

riya-17:
fixes #1392

 When pbench-make-result-tb is passed arguments or option values containing '%', it signals an error. This PR fixes that, so that such arguments or option values are handled properly.


ndokos:
And as you reminded me in issue #1392, `pbench-metadata-log` also needs a change.

agent/util-scripts/pbench-make-result-tb
```diff
@@ -10,11 +10,11 @@ pbench_bin="$(realpath -e ${script_path}/..)"
 
 function usage() {
     printf "usage:\n"
-    printf "${script_name} --result-dir=<pbench results dir> --target-dir=<where to put tar ball> [--help] [--user=<user>] [--prefix=<path>] [--xz-single-threaded=<0|1>]\n"
+    printf "%s" "${script_name} --result-dir=<pbench results dir> --target-dir=<where to put tar ball> [--help] [--user=<user>] [--prefix=<path>] [--xz-single-threaded=<0|1>]\n"
```
portante: The idea behind `printf` in general, even though we have been sloppy with its use up until now, is that the format string contains all the text that is constant with `%` modifiers which describe where the variables land among the text.

So in this case, we should really use (combining both `printf `lines):

```
printf "usage:\n%s --result-dir=<pbench results dir> --target-dir=<where to put tar ball> [--help] [--user=<user>] [--prefix=<path>] [--xz-single-threaded=<0|1>]\n" "${script_name}"
```

Note the `%s` holds the place for the variable `script_name` to be expanded, where we quote the variable to make sure spaces in the variable contents are kept together.

agent/util-scripts/pbench-make-result-tb
```diff
@@ -10,11 +10,11 @@ pbench_bin="$(realpath -e ${script_path}/..)"
...
 }
 
 function missing_arg() {
-    printf "\n${script_name}: ${1} is missing its argument\n\n" >&2
+    printf "%s" "${script_name}: ${1} is missing its argument\n\n" >&2
```
portante: As above, here we have two arguments, so it would look like:

```
printf "\n%s: %s is missing its argument\n\n" "${script_name}" "${1}" >&2
```

agent/util-scripts/pbench-make-result-tb
```diff
@@ -9,12 +9,10 @@ pbench_bin="$(realpath -e ${script_path}/..)"
...
 
 function usage() {
-    printf "usage:\n"
-    printf "${script_name} --result-dir=<pbench results dir> --target-dir=<where to put tar ball> [--help] [--user=<user>] [--prefix=<path>] [--xz-single-threaded=<0|1>]\n"
-}
+	printf "usage:\n%s --result-dir=<pbench results dir> --target-dir=<where to put tar ball> [--help] [--user=<user>] [--prefix=<path>] [--xz-single-threaded=<0|1>]\n" "${script_name}"}
 
 function missing_arg() {
-    printf "\n${script_name}: ${1} is missing its argument\n\n" >&2
+	printf "\n%s: %s is missing its argument\n\n" "${script_name}" "${1}" >&2
```
portante: Tabs vs spaces here ...

agent/util-scripts/pbench-make-result-tb
```diff
@@ -150,25 +148,25 @@ fi
...
+printf -- "%s" "$result_size" | pbench-add-metalog-option ${mdlog} run raw_size
 
 tarball="${target_dir}/${pbench_run_name}.tar.xz"
 if [[ "${xz_single_threaded}" == "1" ]]; then
     tar_cmd="tar --create --force-local -xz \"${pbench_run_name}\""
 else
     tar_cmd="tar --create --force-local \"${pbench_run_name}\" | xz -T0"
 fi
-printf -- "${tar_cmd} > \"${tarball}\"\n" >&2
+printf -- "%s" "${tar_cmd} > \"${tarball}\"\n" >&2
```
portante: It seems this should be something like:

```
printf -- "%s > \"%s\"\n" "${tar_cmd}" "${tarball}" >&2
```

This might be the reason why the unit tests a failing in TravisCI, though I am not sure.

riya-17: Ok, I will look into it

agent/util-scripts/pbench-make-result-tb
```diff
@@ -9,10 +9,8 @@ pbench_bin="$(realpath -e ${script_path}/..)"
 . "$pbench_bin"/base
 
 function usage() {
-    printf "usage:\n"
-    printf "${script_name} --result-dir=<pbench results dir> --target-dir=<where to put tar ball> [--help] [--user=<user>] [--prefix=<path>] [--xz-single-threaded=<0|1>]\n"
+    printf "usage:\n%s --result-dir=<pbench results dir> --target-dir=<where to put tar ball> [--help] [--user=<user>] [--prefix=<path>] [--xz-single-threaded=<0|1>]\n" ${script_name}
```
ndokos: printf -- "....%s...." "${script_name}"

agent/util-scripts/pbench-make-result-tb
```diff
@@ -86,8 +84,8 @@ while true; do
 	    break;
 	    ;;
 	*)
-	    printf "\n${script_name}: you specified an invalid option or an unexpected argument: \"${opt}\"\n\n" >&2
-	    usage >&2
+ 	    printf "\n%s: you specified an invalid option or an unexpected argument: \"${opt}\"\n\n" "${script_name}" >&2    
+		usage >&2
```
portante: Still seems to have a tabs vs spaces issue with lines 87 and 88.

agent/util-scripts/pbench-make-result-tb
```diff
@@ -9,12 +9,10 @@ pbench_bin="$(realpath -e ${script_path}/..)"
...
 
 function usage() {
-    printf "usage:\n"
-    printf "${script_name} --result-dir=<pbench results dir> --target-dir=<where to put tar ball> [--help] [--user=<user>] [--prefix=<path>] [--xz-single-threaded=<0|1>]\n"
+    printf -- "usage:\n%s --result-dir=<pbench results dir> --target-dir=<where to put tar ball> [--help] [--user=<user>] [--prefix=<path>] [--xz-single-threaded=<0|1>]\n" "${script_name}"
 }
-
 function missing_arg() {
-    printf "\n${script_name}: ${1} is missing its argument\n\n" >&2
+    printf -- "\n:%s %s is missing its argument\n\n" "${script_name}" "${1}" >&2
```
portante: s/:%s %s/%s: %s/

agent/util-scripts/pbench-make-result-tb
```diff
@@ -23,7 +21,7 @@ function missing_arg() {
 opts=$(getopt -q -o h --longoptions "user:,prefix:,result-dir:,target-dir:,xz-single-threaded:,help" -n "getopt.sh" -- "${@}")
 sts=${?}
 if [[ ${sts} -ne 0 ]]; then
-    printf "\n${script_name}: you specified an invalid option or an unexpected argument\n\n" >&2
+    printf -- "\n:%s you specified an invalid option or an unexpected argument\n\n" "${script_name}" >&2
```
portante: s/:%s/%s:/

agent/util-scripts/pbench-metadata-log
```diff
@@ -153,12 +153,15 @@ function metadata_log_start {
     local ts=$3
     local log=$dir/metadata.log
 
+    name=$(basename $dir | sed 's/%/%%/')
+    config=$(echo $config | sed 's/%/%%/')
```
portante: Can we just create the entire `metadata.log` file as a temporary file and then use `sed 's/%/%%/g'` to process the entire file?

Note the trailing "`g`" on the `sed` is necessary.

riya-17: Ok, I will try it

ndokos:
You seem to have conflict markers left in `agent/util-scripts/unittests` around line 262. That's why the unit tests failed. Please fix.

agent/util-scripts/unittests
```diff
@@ -327,6 +333,12 @@ declare -A options=(
     [test-45]="--name=mpstat --no-install --remotes=@${_testdir}/tmp/remotes.lis"
     [test-46]="--name=mpstat --no-install --remotes=@${_testdir}/tmp/remotes.lis"
     [test-47]="--name=mpstat --no-install --remotes=@${_testdir}/tmp/remotes.lis --labels=labelOne,labelTwo"
+<<<<<<< HEAD
+=======
+    [test-48]="--name=mpstat --no-install --remotes=@${_testdir}/tmp/remotes.lis --labels=labelOne,labelTwo"
+    [test-48]="pbench-add-metalog-option ${_testdir}/metadata.log run percent < ${_testdir}/tmp/foo.txt"
+    [test-49]="--prefix=foo/bar3% --user=ndk20%"
+>>>>>>> fix-pbench-make-result-tb-errors
```
ndokos: You need to clean this up.

agent/util-scripts/unittests
```diff
@@ -259,7 +259,13 @@ declare -A tools=(
     [test-45]="pbench-register-tool"
     [test-46]="pbench-register-tool"
     [test-47]="pbench-register-tool"
+<<<<<<< HEAD
     [test-48]="test-add-metalog-%-option"
+=======
+    [test-48]="eval"
+    [test-49]="pbench-copy-results"
+    [test-50]="test-pbench-metadata-log"
+>>>>>>> fix-pbench-make-result-tb-errors
```
ndokos: You need to clean this up.

agent/util-scripts/pbench-metadata-log
```diff
@@ -237,4 +237,6 @@ case $1 in
         exit 2
 esac
 
+sed -i 's/%/%%/g' $dir/metadata.log
+
```
ndokos: That is not going to work: the `metadata_log_end` function uses `pbench-add-metalog-option` which already escapes % signs. This is going to escape them *twice*.

riya-17: Should I add that only in metadata-log-beg option? Or I should I keep it as it was?

riya-17: It is working fine with 'metadata-log-beg' option.

ndokos: I think the best thing to do is to modify the `metadata_log_start` routine: produce a  file  the way it is done currently, but store it in a *temporary* file; then at the very end, pass it through `sed` and store it in the final file. BTW, it's better to create it in a temp file *first*, rather than storing it in the final file and doing `sed -i` on that: that way the file is produced atomically, so there is less chance of somebody looking at it in its intermediate state (and getting interpolation errors).

ndokos:
@portante: PTAL

agent/util-scripts/samples/pbench-copy-results/test-49/opt/pbench-agent/id_rsa
```diff
@@ -0,0 +1 @@
+foo
```
portante: Is there a reason why we need this `id_rsa` file for this test?  Seems like this would be required by `ssh` but we really don't need it since we don't invoke `ssh`.  At least, that it what one would think for unit tests.

agent/util-scripts/gold/pbench-metadata-log/test-03.txt
```diff
@@ -68,33 +79,31 @@ turbostat: --interval=3
...
+/var/tmp/pbench-test-utils/test-execution.log:/var/tmp/pbench-test-utils/pbench/tmp/pmdlog.NNNNN/dhcp31-174/iostat:foo
+/var/tmp/pbench-test-utils/test-execution.log:/var/tmp/pbench-test-utils/pbench/tmp/pmdlog.NNNNN/dhcp31-174/mpstat:bar
+/var/tmp/pbench-test-utils/test-execution.log:/var/tmp/pbench-test-utils/pbench/tmp/pmdlog.NNNNN/dhcp31-174/perf:baz
+/var/tmp/pbench-test-utils/test-execution.log:/var/tmp/pbench-test-utils/pbench/tmp/pmdlog.NNNNN/dhcp31-174/pidstat:hunoz
+/var/tmp/pbench-test-utils/test-execution.log:/var/tmp/pbench-test-utils/pbench/tmp/pmdlog.NNNNN/dhcp31-174/proc-interrupts:hukerz
+/var/tmp/pbench-test-utils/test-execution.log:/var/tmp/pbench-test-utils/pbench/tmp/pmdlog.NNNNN/dhcp31-174/proc-vmstat:foobar
+/var/tmp/pbench-test-utils/test-execution.log:/var/tmp/pbench-test-utils/pbench/tmp/pmdlog.NNNNN/dhcp31-174/sar:foo
+/var/tmp/pbench-test-utils/test-execution.log:/var/tmp/pbench-test-utils/pbench/tmp/pmdlog.NNNNN/dhcp31-174/turbostat:bar
```
portante: Why do we lose `.../dhcp31-174/tools-default/sar` directory munged on this line?  Seems like a good fix, but would love to know why this was fixed.

agent/util-scripts/gold/pbench-metadata-log/test-03.txt
```diff
@@ -68,33 +79,31 @@ turbostat: --interval=3
...
+/var/tmp/pbench-test-utils/test-execution.log:/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/scp -r perf104:/var/tmp/pbench-test-utils/pbench/tools-default /var/tmp/pbench-test-utils/pbench/tmp/pmdlog.NNNNN/perf104
+/var/tmp/pbench-test-utils/test-execution.log:/var/tmp/pbench-test-utils/pbench/tmp/pmdlog.NNNNN/perf104/iostat:foo
+/var/tmp/pbench-test-utils/test-execution.log:/var/tmp/pbench-test-utils/pbench/tmp/pmdlog.NNNNN/perf104/mpstat:bar
+/var/tmp/pbench-test-utils/test-execution.log:/var/tmp/pbench-test-utils/pbench/tmp/pmdlog.NNNNN/perf104/perf:baz
+/var/tmp/pbench-test-utils/test-execution.log:/var/tmp/pbench-test-utils/pbench/tmp/pmdlog.NNNNN/perf104/pidstat:hunoz
+/var/tmp/pbench-test-utils/test-execution.log:/var/tmp/pbench-test-utils/pbench/tmp/pmdlog.NNNNN/perf104/proc-interrupts:hukerz
+/var/tmp/pbench-test-utils/test-execution.log:/var/tmp/pbench-test-utils/pbench/tmp/pmdlog.NNNNN/perf104/proc-vmstat:foobar
+/var/tmp/pbench-test-utils/test-execution.log:/var/tmp/pbench-test-utils/pbench/tmp/pmdlog.NNNNN/perf104/sar:foo
+/var/tmp/pbench-test-utils/test-execution.log:/var/tmp/pbench-test-utils/pbench/tmp/pmdlog.NNNNN/perf104/turbostat:bar
```
portante: Ditto here as well for `.../perf104/tools-default/iostat` ...

agent/util-scripts/gold/pbench-metadata-log/test-03.txt
```diff
@@ -1,9 +1,18 @@
 +++ Running test-03 pbench-metadata-log --dir=/var/tmp/pbench-test-utils/pbench/tmp beg
+/bin/ls: cannot access '/var/tmp/pbench-test-utils/pbench/tmp/pmdlog.NNNNN/dhcp31-174/tools-default': No such file or directory
```
portante: Where is this output coming from and is it necessary?

agent/util-scripts/gold/pbench-metadata-log/test-03.txt
```diff
@@ -1,9 +1,18 @@
...
+/bin/ls: cannot access '/var/tmp/pbench-test-utils/pbench/tmp/pmdlog.NNNNN/perf104/tools-default': No such file or directory
 --- Finished test-03 pbench-metadata-log (status=0)
 +++ pbench tree state
 /var/tmp/pbench-test-utils/pbench
 /var/tmp/pbench-test-utils/pbench/tmp
 /var/tmp/pbench-test-utils/pbench/tmp/metadata.log
+/var/tmp/pbench-test-utils/pbench/tmp/pmdlog
```
portante: Where in the code is this `pmdlog` directory being created?  Seems like a bug.

agent/util-scripts/unittests
portante: Seems like this was cleaned up.

agent/util-scripts/unittests
portante: Seems like this was cleaned up.

agent/util-scripts/pbench-metadata-log
portante: Seems like the recommended approach was taken.

agent/util-scripts/samples/pbench-copy-results/test-49/opt/pbench-agent/id_rsa
ndokos: pbench-copy-results checks and fails if it does not exist.

agent/util-scripts/gold/pbench-metadata-log/test-03.txt
ndokos: That's a bug in the mocked scp. I'll post a fix on @riya-17's branch.

agent/util-scripts/gold/pbench-metadata-log/test-03.txt
ndokos: Ditto.

agent/util-scripts/gold/pbench-metadata-log/test-03.txt
ndokos: Same bug: the mocked scp was creating the tools files in the wrong directory.

agent/util-scripts/gold/pbench-metadata-log/test-03.txt
ndokos: It was not: it was copied in from the sample directory (!?!). I'll post a fix in @riya-17's branch.

portante:
@riya-17, @ndokos, can we review and merge https://github.com/distributed-system-analysis/pbench/pull/1412 before we continue on this one?  It would seem a separate commit makes it easier to focus on what this commit is doing specifically.

riya-17:
@portante Nick introduced some changes in this PR as well and some of the changes are similar to your PR #1412 and some are contradictory. Can you look at the changes in this PR?

ndokos:
Yes, I think that's the best course. @riya-17 and I just discussed it: we'll wait until #1412 is merged and this PR willl be rebased on top of that.

agent/util-scripts/test-bin/scp
```diff
@@ -6,6 +6,10 @@ echo "${0} ${*}" >> $_testlog
 src="${@:$(( ${#} - 1 ))}"
 # Fetch the last argument as the "destination" of the copy.
 dst="${@:$#}"
+echo "$0 $*" >> $_testlog
```
ndokos: Why is this file part of this PR at all? Part of the reason for merging #1412 before this one was to get rid of the need to change this file.

portante: This seems like a bad merge.

ndokos:
The basic problem for test-03 is the change that we made to pbench-metadata-log to make the temporary directory unique by tacking on the process id. Tests 31, 32, 33, 39, 41 do not need any mods AFAICT, so you should not have them in this PR at all. Test-49 is a new one that you added for this PR, but the current gold version is wrong.

riya-17:
I will remove the PR sent by @ndokos completely and then rebase the changes

ndokos:
That's correct: most of those changes are now part of #1412.

riya-17:
@portante @ndokos I have brought the PR to its initial state before Nick's PR to my Branch. Please review and let me know from here what needs to be done.

ndokos:
Well, it's not in the initial state before you merged my PR: it still contains changes to scp and a bunch of gold files that it should not contain.

It should only contain:

- the changes to pbench-make-result-tb and pbench-metadata-log
- the changes to unittests
- the samples/pbench-copy-results/test-49 directory hierarchy (with id_rsa and metadata.log files in the appropriate places).
- the test-pbench-metadata-log script for test-50.
- two new gold files for test-49 and test-50.
- one modified gold file for test-03 (because of including the pid in the name of the temp directory).

That's it.

PR #1396, opened 2019-10-30
Remove redundant fetch of MD5s from verify backup

portante:
To remove the redundant fetch of the MD5 values we added those methods to
the BackupObject class which let's us store the accummulated values in the
object for use later.

We also enhance the report formating so that we get separartion between
sub-sections of the report, and a summary of the timings for all the
phases.


portante:
Reverted this commit because it was breaking the build, and I still did not have a chance to apply some fixes.

PR #1399, opened 2019-10-30
Audit now ignores incoming `*.unpack` dirs

portante:
The audit server code now ignores incoming `*.unpack` dirs which have valid tar balls in the `$ARCHIVE` hierarchy, and reports those directories which do not.


server/bin/pbench-audit-server.sh
```diff
@@ -201,11 +201,14 @@ function verify_incoming {
         tarball_dirs="${workdir}/tarballdirs"
         empty_tarball_dirs="${workdir}/emptytarballdirs"
         > ${empty_tarball_dirs}.unsorted
+        unpacking_tarball_dirs="${workdir}/unpackingtarballdirs"
+        > ${unpacking_tarball_dirs}
```
ndokos: Shouldn't  the `tarball_dirs` and `tarball_links` files be initialized to empty as well?

portante: One can do that, but `-fprintf` of `find` either creates, or overwrites the file given.

ndokos: But then why do it for two, and not for the other two? Let `-fprintf` handle it for all of them or (redundantly I realize) initialize all of them. It's the inconsistency that bugs me.

portante: Fixed.

portante:
@ndokos, all comments addressed, please take another look.

server/bin/pbench-audit-server.sh
```diff
@@ -199,55 +199,77 @@ function verify_incoming {
...
 
         invalid_tb_dirs="${workdir}/invalidtbdirs"
-        > ${invalid_tb_dirs}.unsorted
+        > ${invalid_tb_dirs}
         while read tb ; do
             if [[ -r $ARCHIVE/${controller}/${tb}.tar.xz ]]; then
                 continue
             fi
-            printf "\t\t${tb}\n" >> ${invalid_tb_dirs}.unsorted
+            printf "\t\t${tb}\n" >> ${invalid_tb_dirs}
```
ndokos: It'd be more efficient to redirect the while loop to the file, rather than doing it every time through the loop.

server/bin/pbench-audit-server.sh
```diff
@@ -199,55 +199,77 @@ function verify_incoming {
...
+
+        invalid_unpacking_dirs="${workdir}/invalidunpacking"
+        > ${invalid_unpacking_dirs}
+        while read tb_u ; do
+            tb=${tb_u%*.unpack}
+            if [[ -r $ARCHIVE/${controller}/${tb}.tar.xz ]]; then
+                continue
+            fi
+            printf "\t\t${tb_u}\n" >> ${invalid_unpacking_dirs}
+        done < ${unpacking_tarball_dirs}
```
ndokos: Same thing here.

server/bin/pbench-audit-server.sh
```diff
@@ -199,55 +199,77 @@ function verify_incoming {
...
+        rm ${invalid_unpacking_dirs}
 
         invalid_tb_dir_links="${workdir}/invalidtbdirlinks"
-        > ${invalid_tb_dir_links}.unsorted
+        > ${invalid_tb_dir_links}
         while read tb link ; do
             # The link in the incoming directory does not point to
             # the expected directory in the unpack hierarchy.
-            printf "\t\t${tb}\n" >> ${invalid_tb_dir_links}.unsorted
+            printf "\t\t${tb}\n" >> ${invalid_tb_dir_links}
```
ndokos: Same thing here.

server/bin/pbench-audit-server.sh
portante: Done.

server/bin/pbench-audit-server.sh
portante: Done.

server/bin/pbench-audit-server.sh
portante: Done.

ndokos:
I thought there were three while loops that I flagged but I now see only two. Did I hallucinate the third one?

portante:
> I thought there were three while loops that I flagged but I now see only two. Did I hallucinate the third one?

No, it was there, but looking at the code, I removed it by just using the `find` command output.  There was no need to process that data because we no longer use symlinks in `incoming`.

PR #1400, opened 2019-10-31
Enhanced pbench-unpack-tarballs to operate on buckets of tar balls by size, reverse sorted by date

### LINKED ISSUE #1398: Refine algorithm for unpacking tar balls
### Background

The current method of storing tar balls leverages a file system hierarchy.  The state of what needs to happen next for those tar balls is encoded via `symlinks` on the file system.  Since the symbolic links are in separate directories, we have to crawl through the hierarchy of directories to find which tar balls to act on next.  This is by nature a polling mechanism, and it takes some amount of time to accumulate all of the tar ball names to operate on for a given method.

This is true for ALL pbench server side methods and operations.

Once we accumulate a list of tar balls to operate on, unless the method stops and re-polls for tar balls to work on, new tar balls that method will have to wait until the method is restarted.  Typically, this is waiting for the cron job restart.

For unpacking tar balls, this is _typically_ not a big problem because pbench result tar balls arrive quite in frequently relative to the length of time it takes to unpack them.  The unpacking algorithm sorts all the tar balls it finds that need unpacking by their size, and works from the small to the largest.

However, in cases where we do get many result tar balls at the same time, or where we have to re-unpack a set of tar balls, this method leads to tar balls being processed infrequently due to many large tar balls clogging the queue.

### Proposal

1. Have unpack tar balls periodically re-poll for the list of tar balls to service
  * It should keep a target window for re-polling, which should be a minimum of 1 minute (current highest resolution offered by cron jobs), and with a configured maximum (perhaps 5 minutes)
  * It should keep a running estimate of how long it takes to unpack tar balls on average, and re-poll if that average unpack time is longer than the target window
2. Have multiple instances of unpack tar balls work for different tar ball sizes
  * This will prevent larger tar balls from starving out smaller ones when the unpack time is beyond the desired time
3. Have unpack tar balls operate from newest to oldest within its size category

portante:
Fixes #1398.

We now have four different jobs by default for processing tar ball sizes.  Based on historical pbench result sizes, we set default buckets to capture the first 90%, then next 5% (95th percentile), then the next 4% (99th percentile), and then finally the last 1% (100th percentile).

The algorithm of unpack tar balls now stops processing the list of tar balls to unpack if the time taken to unpack tar balls so far is twice as long as it takes to find tar balls to unpack.  If the time taken to find tar balls takes less than one minute, we just use 60 sec as the minimum since that is the shortest possible interval of time between cronjob invocations.

Further, tar balls are processed now from newest to oldest, so by restarting periodically, more recent results will get unpacked before older results within a size bucket.

Removed the redundant `test-5.1.setup` file replacing it with a symlink to `test-5.setup` (updating `test-5.2.setup` to link to the same).

This PR contains 4 commits to make it easier to review:

1. The first commit introduces a unit test to ensure `find` works as expected independent of these changes
2. The second commit refactors the code a bit to address Bourne Shell style `[ ]` conditionals, where we now uniformly use Bash style `[[ ]]` conditionals, address duplicate code found during reviews, and other clean-ups
3. The third commit refactors parts of the code into functions without other additional behavioral changes to make it easier to review
4. The fourth commit applies the new restart algorithm

### DIFF
--- a/server/bin/pbench-unpack-tarballs.sh
+++ b/server/bin/pbench-unpack-tarballs.sh
@@ -38,53 +38,48 @@
 
 
 # load common things
-. $dir/pbench-base.sh
+. ${dir}/pbench-base.sh
 
 # check that all the directories exist
-test -d $ARCHIVE || doexit "Bad ARCHIVE=$ARCHIVE"
-test -d $INCOMING || doexit "Bad INCOMING=$INCOMING"
-test -d $RESULTS || doexit "Bad RESULTS=$RESULTS"
-test -d $USERS || doexit "Bad USERS=$USERS"
+test -d ${ARCHIVE} || doexit "Bad ARCHIVE=${ARCHIVE}"
+test -d ${INCOMING} || doexit "Bad INCOMING=${INCOMING}"
+test -d ${RESULTS} || doexit "Bad RESULTS=${RESULTS}"
+test -d ${USERS} || doexit "Bad USERS=${USERS}"
 
-# make sure only one copy is running.
-# Use 'flock -n $LOCKFILE /home/pbench/bin/pbench-unpack-tarballs' in the
-# crontab to ensure that only one copy is running. The script itself
-# does not use any locking.
-
-# the link source and destination(s) for this script
+# The link source and destination(s) for this script.
 linksrc=TO-UNPACK
 linkdest=UNPACKED
 linkdestlist=$(getconf.py -l unpacked-states pbench-server)
 
-tmp=$TMP/$PROG.$$
-trap 'rm -rf $tmp' EXIT INT QUIT
+tmp=${TMP}/${PROG}.${$}
+trap 'rm -rf ${tmp}' EXIT INT QUIT
 
-mkdir -p $tmp || doexit "Failed to create $tmp"
+mkdir -p ${tmp} || doexit "Failed to create ${tmp}"
 
-log_init $PROG
+log_init ${PROG}
 
-echo $TS
+echo ${TS}
 
 # Accumulate errors and logs in files for reporting at the end.
-mail_content=$tmp/mail_content.log
-index_content=$tmp/index_mail_contents
+mail_content=${tmp}/mail_content.log
+index_content=${tmp}/index_mail_contents
 
 # get the list of files we'll be operating on - sort them by size
-list=$tmp/$PROG.list
+list=${tmp}/${PROG}.list
 
-# Find all the links in all the $ARCHIVE/<controller>/$linksrc
+# Find all the links in all the ${ARCHIVE}/<controller>/${linksrc}
 # directories, emitting a list of their full paths with the size
 # in bytes of the file the link points to, and then sort them so
 # that we process the smallest tar balls first.
 > ${list}.unsorted
-# First we find all the $linksrc directories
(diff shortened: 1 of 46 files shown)


server/bin/pbench-unpack-tarballs.sh
```diff
@@ -110,45 +112,52 @@ function gen_work_list() {
...
-            find -L $linksrc_dir -type l -name '*.tar.xz' -printf "%s %p\n" 2>/dev/null >> ${list}.unsorted
+            find -L $linksrc_dir -type l -name '*.tar.xz' ! -name 'DUPLICATE__NAME*' -printf "%CY-%Cm-%CdT%CT %s %p\n" 2>/dev/null >> ${list}.unsorted
         fi
     done
-    sort -n ${list}.unsorted > ${list}
+    sort -k 1 -r ${list}.unsorted > ${list}
     rm -f ${list}.unsorted
+    # Pad by one minute, the default smallest cronjob interval.
+    let max_seconds=${SECONDS}+60
+    return $(( ${max_seconds} * 2 ))
```
ndokos: So if the loop and the sort takes 5 seconds, this routine will return 2*(5+60) = 130 seconds of work time. Is that what you want or did you mean to say 2*5+60 = 70 seconds? Also, shouldn't we scale the work time differently depending on the bucket? Large tarballs take roughly the same time to be found and sorted, but they take longer to be unpacked.

ndokos: If we sort by time, and always  work from newer to older, isn't there a danger of starving the older tarballs?

server/bin/pbench-unpack-tarballs.sh
```diff
@@ -38,295 +38,361 @@
...
+        resultname=${resultname%.tar.xz}
+
+        link=$(readlink -e ${result})
+        if [[ -z "${link}" ]]; then
+            echo "${TS}: symlink target for ${result} does not exist" | tee -a ${mail_content} >&4
+            nerrs=${nerrs}+1
+            hostname=$(basename $(dirname $(dirname ${result})))
+            mkdir -p ${ARCHIVE}/${hostname}/${linkerr}
+            move_symlink ${hostname} ${resultname} ${linksrc} ${linkerr} || doexit "Error handling failed for symlink"
+            if [[ ${SECONDS} -ge ${max_seconds} ]]; then break; fi
```
riya-17: Will this condition vary after each if condition?

riya-17: ok. Thank you :)

server/bin/pbench-unpack-tarballs.sh
portante: As we discussed face-to-face, we want the unpack loop to re-evaluate the list of tar balls to unpack if `${max_seconds}` has passed since it started unpacking tar balls.  The goal is to have a gross approximation of unpacking the most recent tar balls as soon as possible.  

The shortest amount of time in the current method that we could notice and unpack a tar ball is 1 minute, the minimum cronjob interval.   But this code does not use 1 minute as the maximum so that for the smaller tar balls we can process more of the existing set before we find a new set.  For the larger ones it doesn't really matter, as there are fewer of them.

As for starvation, yes this can starve out older tar balls, but that happens today but in reverse: older tar balls can starve out newer ones for many hours or days in the right situation.  Solving starvation properly is a much larger change, and out of scope for this change.

PR #1403, opened 2019-11-04
Rewritten metalog-%-error testcase

### LINKED ISSUE #1329: pbench-add-metalog-option blows up when the option is a '%'

### LINKED ISSUE #1392: `pbench-make-result-tb` errors when arguments contain %.

riya-17:
Fixes #1329

The problem is that configparser Interpolation (https://docs.python.org/3/library/configparser.html#configparser.InterpolationError) uses % as a marker for the interpolation construct %(option)s which replaces the construct by the value of the option. Configparser Interpolation does not accept the % symbol in any other context unless it is escaped as a double percent sign. So when we write a value with '%', ex: "30%" in any config file, this error in thrown. To fix this error, whenever the user enters a value that includes a % char, it will be converted into %% (double percent sign) which interpolation treats as a literal single percent sign.

This PR deals with the testcase of the problem

### DIFF
--- a/agent/util-scripts/gold/test-add-metalog-%-option/test-48.txt
+++ b/agent/util-scripts/gold/test-add-metalog-%-option/test-48.txt
@@ -1,5 +1,5 @@
-+++ Running test-48 eval pbench-add-metalog-option /var/tmp/pbench-test-utils/pbench/metadata.log run percent < /var/tmp/pbench-test-utils/pbench/tmp/foo.txt
---- Finished test-48 eval (status=0)
++++ Running test-48 test-add-metalog-%-option 
+--- Finished test-48 test-add-metalog-%-option (status=0)
 +++ pbench tree state
 /var/tmp/pbench-test-utils/pbench
 /var/tmp/pbench-test-utils/pbench/metadata.log
--- a/agent/util-scripts/test-bin/test-add-metalog-%-option
+++ b/agent/util-scripts/test-bin/test-add-metalog-%-option
@@ -1,0 +1,4 @@
+#! /bin/bash
+
+echo "30%" | pbench-add-metalog-option ${_testdir}/metadata.log run percent
+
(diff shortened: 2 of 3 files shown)


portante:
@gurbirkalsi, can you take a look to see what the e2e tests are failing?

PR #1404, opened 2019-11-04
Fix failing unit tests on some Fedora systems

portante:
It looks like newer versions of pass process associative arrays in different orders.  So we consistently use a numeric indexing scheme to avoid that behavior, allowing the unit tests to pass in all cases.

We also add the line number where a remotes file is not formatted as expected.

### DIFF
--- a/agent/util-scripts/pbench-register-tool
+++ b/agent/util-scripts/pbench-register-tool
@@ -152,7 +152,8 @@ if [[ ${remotes_arg::1} == "@" ]]; then
 		IFS=',' read -ra parts <<< "${line}"
 		if [[ ${#parts[@]} == 0 ]]; then
 			# Ignore empty lines
-			:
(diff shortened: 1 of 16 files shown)


portante:
Just e2e test failures, we can ignore for now because these changes are not to the UI side of the tree.

agent/util-scripts/pbench-register-tool
```diff
@@ -316,7 +318,8 @@ if [[ ${do_install} == 1 ]]; then
 else
 	isntall_opt="--no-install"
```
ndokos: Typo: install_opt

portante: Thanks, fixed!

portante:
Unit tests are passing, the e2e tests are failing, but this code change does not change anything on the UI side so we can ignore for now.

PR #1405, opened 2019-11-04
Remove redundant fetch of MD5s from verify backup

portante:
This is a re-submit of this patch (7328bc648) after fixing the test environment on Fedora systems (a9a007ca6).

To remove the redundant fetch of the MD5 values we added those methods to the BackupObject class which let's us store the accumulated values in the object for use later.

We also enhance the report formatting so that we get separation between sub-sections of the report, and a summary of the timings for all the phases.


portante:
Unit tests are passing, only the e2e tests are failing; will merge regardless because these changes don't affect the UI.

PR #1406, opened 2019-11-04
Trigger e2e tests after dashboard environment initializes 

gurbirkalsi:
**Summary**

Currently, e2e tests will occasionally fail if the dashboard environment has not successfully been initialized. As the tests listen for the dashboard running on a port to trigger e2e tests, we need a smarter way to confirm the environment is ready. This is currently being accomplished with a hard coded sleep timer in between commands.

### DIFF
--- a/web-server/v0.4/e2etests
+++ b/web-server/v0.4/e2etests
@@ -17,17 +17,7 @@ if [[ $? -ne 0 ]]; then
   echo "yarn failed!" >&2
   exit 1
 fi
-yarn start &
-if [[ $? -ne 0 ]]; then
-  echo "yarn failed!" >&2
-  exit 1
-fi
-sleep 60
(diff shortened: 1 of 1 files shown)


gurbirkalsi:
@portante Sorry, I had meant to mark this PR as WIP. We need to find a way to sequentially trigger `yarn start` and then `yarn test` so our e2e tests run once the development environment has successfully initialized. Are you familiar with techniques that would allow this within a bash script?

portante:
@gurbirkalsi, funny I just spent the last few hours trying to work something out using bash, but because the `yarn start` does not listen for SIGTERM or SIGINT, it just hangs.

So I have updated #1408 with a fix for the e2e tests as well.

portante:
Closing in favor for #1409.

PR #1409, opened 2019-11-05
Refine e2e tests to clean up background web server

portante:
We enhance the `e2etests` script to keep track of the `yarn start` running in the background and make sure it is stopped before the script exits.

We also fix the yarn version we use to `1.19.1`.

Finally, we update the `README.md` to record the requirement of having `google-chrome-stable` installed for the E2E tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,7 +64,8 @@ jobs:
       before_install:
         # Enable user namespace cloning
         - "sysctl kernel.unprivileged_userns_clone=1"
-        # Remove /usr/local/bin/yarn to avoid conflicts with installed version
+        # Remove /usr/local/bin/yarn to avoid conflicts with the "old" version
+        # provided by the TravisCI language "node_js".
         - "sudo rm /usr/local/bin/yarn"
       install:
         - sudo apt-key adv --fetch-keys http://dl.yarnpkg.com/debian/pubkey.gpg
--- a/web-server/v0.4/e2etests
+++ b/web-server/v0.4/e2etests
@@ -1,6 +1,11 @@
 #!/bin/bash
 
-cd $(dirname $0)
+# We change our working directory to the same location as this script is
+# found so that yarn can work with the package.json file found in the same
+# directory.
+cd $(dirname ${0})
+
+# Setup the mock'd data store configuration.
 cat > mock/datastoreConfig.js <<EOF
 export default {
   '/dev/datastoreConfig': {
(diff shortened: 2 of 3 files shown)


portante:
@ndokos, please take a look at the bash hackery employed ... (eek).

portante:
Okay, this seems to be working well now for E2E tests.  I have this working on my Fedora 30 box, and the e2e tests seems to pass under TravisCI.

ndokos:
Famous last words :-) It's still failing...

portante:
@ndokos, I have been experimenting with how to detect "yarn start" is running correctly in the background, as it appears the Ubuntu environment gets a different version of yarn / umi than what we get under Fedora.

portante:
Retargeted this PR to just provide some remaining clean up work, add some comments, and document `google-chrome-stable` requirement.

web-server/v0.4/e2etests
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 
-cd $(dirname $0)
+# FIXME - Why do we have to CD to the directory where this script is
```
gurbirkalsi: We need to be able to reference `package.json` when invoking yarn commands in this script.

.travis.yml
```diff
@@ -64,7 +64,8 @@ jobs:
       before_install:
         # Enable user namespace cloning
         - "sysctl kernel.unprivileged_userns_clone=1"
-        # Remove /usr/local/bin/yarn to avoid conflicts with installed version
+        # Remove /usr/local/bin/yarn to avoid conflicts with the "old" version
+        # provided by the TravisCI language "node_js".
```
ndokos: "language" ?

web-server/v0.4/e2etests
portante: Updated the comment removing the `FIXME` and replacing it with the reason.

web-server/v0.4/e2etests
```diff
@@ -12,16 +17,22 @@ export default {
   },
 };
 EOF
+# Make sure the yarn environment is all up-to-date, or perform the
+# install ahead of actually running the tests.
+printf -- "Running \"yarn install\" command ...\n"
```
ndokos: Should the next line be `yarn install  --network-timeout ...`?

web-server/v0.4/e2etests
```diff
@@ -12,16 +17,22 @@ export default {
...
-if [[ $? -ne 0 ]]; then
-  echo "yarn failed!" >&2
-  exit 1
+yarn_start_pid=${!}
+if [[ -z "${yarn_start_pid}" ]]; then
+    printf -- "yarn e2e tests failed! (could not execute \"yarn start\" in background)\n" >&2
+    exit 1
 fi
 # Wait checking once a second for the web-server to show up on port
 # 8000 at the expected URL.
```
ndokos: Maybe add an explanation of what happens if the curl fails indefinitely?

.travis.yml
portante: Yes, see https://docs.travis-ci.com/user/languages/javascript-with-nodejs/.

web-server/v0.4/e2etests
portante: Yes, I can add `install` as that is the default behavior.

portante: Fixed.

web-server/v0.4/e2etests
portante: Somebody will have to stop the `e2etest` script and debug why it happened.  In TravisCI the job will just time out and get cleaned up.  The key here is we can't start running the E2E tests until we detect the web server is up and running.  So there is no point doing anything further.

There is a comment below at line 51 what states this.  Is that sufficient?

ndokos: I'd rather have the explanation as a comment, but I won't insist.

PR #1411, opened 2019-11-06
Revamp TravisCI jobs to break out testing areas

portante:
This is a series of commits to address some issues with the TravsiCI.

The biggest change is that we are revamping the jobs into agent, server, and UI unit tests, and UI e2e tests.  This makes what we are testing more targeted and the setup more streamlined.

We also take the opportunity to "disable" the E2E tests until we can figure out what is happening with them instead of having many failed jobs to review.

We wait for server URL instead of sleeping in the E2E tests.


portante:
This should restore our TravisCI jobs for the time being.

portante:
@ndokos, @gurbirkalsi, just updated with what we think is working E2E tests.  If these tests pass, let's review to double check and then merge.

portante:
Okay, famous last words.  This is still not working for the E2E tests, now for a different reason:

```
This dependency was not found:
* @tmp/history in ./src/pages/.umi/dva.js
```

I have also fixed up the unit tests so they work with the latest version of Enzyme that was being installed when we run in TravisCI and on clean development environments.

PR #1412, opened 2019-11-06
Fix use of scp/ssh options and mock scp cleanly

portante:
The scp mock command now only mocks local copies, which are used when copying
remote tool data locally.  The other copies are only recorded.

There was a wayward tmp directory in the sample data which should not be there.

Finally, we fix the various scp commands to use the right options.

### DIFF
--- a/agent/util-scripts/test-bin/scp
+++ b/agent/util-scripts/test-bin/scp
@@ -1,22 +1,35 @@
 #!/bin/bash
 
-echo "$0 $*" >> $_testlog
+echo "${0} ${*}" >> $_testlog
 
-# this gets the last argument
-tools=${@:$#}
+# Fetch the second-to-last argument as the "source" of the copy.
+src="${@:$(( ${#} - 1 ))}"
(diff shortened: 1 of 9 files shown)


agent/util-scripts/test-bin/scp
```diff
@@ -1,22 +1,35 @@
...
-echo hukerz > $tools/proc-interrupts
-echo foobar > $tools/proc-vmstat
-echo foo > $tools/sar
-echo bar > $tools/turbostat
-for x in $(find $tools -type f | sort); do
-    echo -n $x: ; cat $x
-done >> $_testlog
+# Pretend that we copied a tools-<dir> from the remote.
 
+# Grab the target directory being copyied from the source argument
```
ndokos: copyied --> copied

portante: Fixed.

portante:
Unit tests are succeeding, but the E2E tests are failing, so we can ignore since this is agent side work.

PR #1413, opened 2019-11-07
Upgrade umi and umi plugins to latest version 

gurbirkalsi:
**Summary**

Upgrading umi and umi plugins to their latest versions ensures compatibility between dependencies. This change also references the `persistStore` via the constructor of `BasicLayout` to ensure the `PersistGate` component opens on initial load.


gurbirkalsi:
@portante Hopefully this upgrade resolves the dependency issues we are observing in TravisCI

PR #1419, opened 2019-11-21
fio-postprocess: bug fix for mixed i/o data handling

k-rister:
- Previously, fio-postprocess tracked mixed i/o (ie. simultaneous
  reads and writes on the same test) separately using the 'rw_label'.
  This would create two separate records in the samples array with the
  same timestamp but different i/o types and values.  The problem is
  that the rest of the postprocessing stack is not written to handle
  data in this way and does not expect there to be multiple samples in
  the array with the same timestamp.  This lead to data corruption in
  multiple places and levels of the postprocessing stack.

- Fix the bug by intelligently combining the data where it's
  relationship is understood and removing the concept of separate i/o
  types and values.

### DIFF
--- a/agent/bench-scripts/postprocess/fio-postprocess
+++ b/agent/bench-scripts/postprocess/fio-postprocess
@@ -257,16 +257,47 @@ foreach $client_hostname_dir (@client_hostname_dirs) {
 					my $timestamp_ms = $1;
 					my $value = $2;
 					my $rwtype = $3;
-					my %sample = (get_label('date_label') => int $timestamp_ms, get_label('value_label') => int $value, get_label('rw_label') => int $rwtype);
+					my %sample;
 					my $index = value_exists(get_label('date_label'), $timestamp_ms, \@samples);
-					# duplicate timestamp *AND* duplicate rwtype
-					if ($index >= 0 && $samples[$index]{get_label('rw_label')} == $rwtype) {
-						printf "warning: timestamp already exists with this RW type, skipping this sample (file = %s timestamp = %d, value = %d, rwtype = %d)\n",
-						    "$dir/clients/$client_hostname_dir/$log_file", $timestamp_ms, $value, $rwtype;
+					if ($index == -1) {
+					    %sample = ( get_label('date_label') => int $timestamp_ms,
+							'data' => { $rwtype => { 'values' => [ int $value ] } },
+							'type' => $log_type );
+					    push(@samples, \%sample);
 					} else {
-						push(@samples, \%sample);
+					    if (exists($samples[$index]{'data'}{$rwtype})) {
+						push(@{$samples[$index]{'data'}{$rwtype}{'values'}}, int $value);
+					    } else {
+						$samples[$index]{'data'}{$rwtype} = { 'values' => [ int $value ] };
+					    }
 					}
 				}
+			}
+			for (my $i=0; $i < scalar @samples; $i++) {
+			    $samples[$i]{get_label('value_label')} = 0;
+
+			    foreach my $key (keys %{$samples[$i]{'data'}}) {
+				$samples[$i]{'data'}{$key}{'mean'} = 0;
+
+				if (scalar @{$samples[$i]{'data'}{$key}{'values'}} > 1) {
+				    printf "[WARNING] %s: timestamp %s for rwtype %d found multiple times, values [%s] will be averaged\n", $log_file, $samples[$i]{get_label('date_label')}, $key, join(', ', @{$samples[$i]{'data'}{$key}{'values'}});
+				}
+
+				for (my $x=0; $x < scalar @{$samples[$i]{'data'}{$key}{'values'}}; $x++) {
+				    $samples[$i]{'data'}{$key}{'mean'} += $samples[$i]{'data'}{$key}{'values'}[$x];
+				}
+
+				$samples[$i]{'data'}{$key}{'mean'} /= scalar @{$samples[$i]{'data'}{$key}{'values'}};
+
+				$samples[$i]{get_label('value_label')} += $samples[$i]{'data'}{$key}{'mean'};
+			    }
+
+			    if ($samples[$i]{'type'} =~ /lat$/) {
+				$samples[$i]{get_label('value_label')} /= scalar(keys(%{$samples[$i]{'data'}}));
+			    }
+
+			    delete $samples[$i]{'type'};
+			    delete $samples[$i]{'data'};
 			}
 			$nr_log_files++;
 			close(IN_FILE);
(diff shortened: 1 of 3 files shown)


agent/bench-scripts/postprocess/fio-postprocess
```diff
@@ -257,17 +257,25 @@ foreach $client_hostname_dir (@client_hostname_dirs) {
 					my $timestamp_ms = $1;
 					my $value = $2;
 					my $rwtype = $3;
-					my %sample = (get_label('date_label') => int $timestamp_ms, get_label('value_label') => int $value, get_label('rw_label') => int $rwtype);
+					my %sample;
 					my $index = value_exists(get_label('date_label'), $timestamp_ms, \@samples);
-					# duplicate timestamp *AND* duplicate rwtype
-					if ($index >= 0 && $samples[$index]{get_label('rw_label')} == $rwtype) {
-						printf "warning: timestamp already exists with this RW type, skipping this sample (file = %s timestamp = %d, value = %d, rwtype = %d)\n",
```
atheurer: One thing occurred to me about this warning: I added this because we observed (sadly quite often) that fio would report an operation twice with the same timestamp but different values.  I am a little concerned that we are losing the ability to identify this problem.  Also, I am wondering, if there are more than 1 sample for the same time-stamp [and same rw-type] are we handling this properly?  My understanding is that below we are just adding to the value (line 266).  If we encounter >1 sample with same timestamp and same rwtype, we could be adding those two values.  That is going to be the wrong interpretation.  We either need to drop duplicate samples [of the same rw-type and timestamp] or average all of those samples' values into 1 value.  We should only combine values when the rw-types are different.  So I think you will still need to track $rwtype while samples are being processed, but the sample itself does not have to have a $rwtype.

k-rister: Ugh.  What do you think is more appropriate, ignoring the values for repeated timestamp+rwtype combinations (the previous behavior) or averaging the multiple values for the same combo?

atheurer: I suppose more appropriate is averaging.

portante: Perhaps we can add a flag to that number to indicate it is an average of two values having the same timestamp?

k-rister:
Posted some updated code that handles the multiple timestamp+rwtype case.  The new code dumps output like this:

```
[WARNING] fio_clat.1.log: timestamp 1571584936425 for rwtype 0 found multiple times, values [34482, 36000] will be averaged
[WARNING] fio_clat.2.log: timestamp 1571584936425 for rwtype 0 found multiple times, values [32166, 43000] will be averaged
[WARNING] fio_clat.2.log: timestamp 1571584936425 for rwtype 1 found multiple times, values [35403, 30000] will be averaged
[WARNING] fio_clat.3.log: timestamp 1571584936425 for rwtype 0 found multiple times, values [31544, 29000] will be averaged
[WARNING] fio_lat.1.log: timestamp 1571584936425 for rwtype 0 found multiple times, values [40965, 46000] will be averaged
[WARNING] fio_lat.2.log: timestamp 1571584936425 for rwtype 1 found multiple times, values [41910, 36000] will be averaged
[WARNING] fio_lat.2.log: timestamp 1571584936425 for rwtype 0 found multiple times, values [38297, 48000] will be averaged
[WARNING] fio_lat.3.log: timestamp 1571584936425 for rwtype 0 found multiple times, values [37936, 33000] will be averaged
```

Once we arrive at a consensus on implementation I'll clean up the commits (collapse them) and fix any new unit-test issues.

agent/bench-scripts/postprocess/fio-postprocess
k-rister: I don't think adding a flag will work without a fairly complete modification of the entire postprocessing stack to surface new properties like that.  Further, as it turns out this is a fairly common issue so I don't think it is the kind of thing that will happen once and awhile and you want to know.  It happens all the time.  My new approach does provide more verbose logging that can be used to investigate the values if the user desires.

portante:
@k-rister, can you share some examples of the raw data that ends up being a duplicate?  I am curious, thanks.

k-rister:
> @k-rister, can you share some examples of the raw data that ends up being a duplicate? I am curious, thanks.

@portante Here is an example from the end of a completion latency log.  I've alternated bold/not-bold records with identical timestamps and as you can see at the very end of the file it dumped out the same timestamp 4 times instead of 2, twice with each rwtype (3rd column):


$ tail -n 16  9__bs\=16k_iodepth\=4_iodepth_batch_complete_max\=4/sample4/clients/localhost/fio_clat.2.log
**1571584930425, 35529, 1, 0
1571584930425, 32340, 0, 0**
1571584931425, 35322, 1, 0
1571584931425, 32106, 0, 0
**1571584932425, 32320, 0, 0
1571584932425, 35591, 1, 0**
1571584933425, 35336, 1, 0
1571584933425, 32161, 0, 0
**1571584934425, 35341, 1, 0
1571584934425, 32166, 0, 0**
1571584935425, 35405, 1, 0
1571584935425, 32258, 0, 0
**1571584936425, 32166, 0, 0
1571584936425, 35403, 1, 0
1571584936425, 43000, 0, 0
1571584936425, 30000, 1, 0**

portante:
> @portante Here is an example from the end of a completion latency log. I've alternated bold/not-bold records with identical timestamps and as you can see at the very end of the file it dumped out the same timestamp 4 times instead of 2, twice with each rwtype (3rd column):

Is it always at the end?  Or does this duplication happen elsewhere?

If always at the end, seems like those values can just be ignored.

k-rister:
> Is it always at the end? Or does this duplication happen elsewhere?

It appears to always be at the end on the particular data that I have in front of me, but I'm not sure that should be considered representative.  Maybe @atheurer can enlighten us on where he was seen it.  Either way, the proposed changes would handle it cleanly and log the behavior.

> 
> If always at the end, seems like those values can just be ignored.

At this point I've written the logic to handle it in all cases and it should also be able to handle any $rwtype with the updates in a clean manner.  There are other ops that fio can do (such as `trim`) that we haven't really tested but I would surmise that it would show up with yet a different value (than 0 or 1) but this code would aggregate any op type while average multiple values for any particular op type.

atheurer:
> > @portante Here is an example from the end of a completion latency log. I've alternated bold/not-bold records with identical timestamps and as you can see at the very end of the file it dumped out the same timestamp 4 times instead of 2, twice with each rwtype (3rd column):
> 
> Is it always at the end? Or does this duplication happen elsewhere?
> 
> If always at the end, seems like those values can just be ignored.

I don't have an example handy, but I am pretty sure I have seen it not at the end but somewhere in the middle.

portante:
@k-rister, unit tests failing at`Testing bench-scripts/postprocess/fio-postprocess under sample fio-0`:

```
+ Argument "1/8" isn't numeric in division (/) at ../../bench-scripts/postprocess/fio-postprocess line 296, <IN_FILE> line 16.
```

portante:
@k-rister, can you collapse this to two commits, the `pbench-run-benchmark` fixes and the `fio-postprocess` fixes?  Thanks!

PR #1422, opened 2019-12-06
Enhancement of the server audit script to detect unnecessary files

### LINKED ISSUE #1191: Error checking in the shim scripts
`pbench-server-prep-shim-002` loops over md5 files in its reception area: it will not look for anything else, so if something else ends up in the reception area (e.g. a partially copied tarball without an md5 file, or an md5.check file that never got promoted to an md5 file), it will stay there forever. Somebody needs to check and clean these remnants out.

Anisha1234:
Will fix #1191 

Work Implemented:

1. Check for any other files except type _`.tar.xz`_ or _`.tar.xz.md5`_
2. Check for tars not having md5 files
3. Check md5sum of the files.

### DIFF
--- a/server/bin/pbench-audit-server.sh
+++ b/server/bin/pbench-audit-server.sh
@@ -65,6 +65,7 @@ archive_report=$workdir/archive_report
 incoming_report=$workdir/incoming_report
 results_report=$workdir/results_report
 users_report=$workdir/users_report
+reception_report=$workdir/reception_report
 index_content=$workdir/index_content
 bad_controllers=$workdir/badcontrollers
 controllers=$workdir/controllers
@@ -78,6 +79,12 @@ tarballs=$workdir/tarballs
 linkdirs=$workdir/linkdirs
 empty=$workdir/empty
 users=$workdir/users
+reception_dir_prefix=$(getconf.py pbench-receive-dir-prefix pbench-server)
+if [ -z "$reception_dir_prefix" ] ;then
+    echo "Failed: \"getconf.py pbench-receive-dir-prefix pbench-server\"" 
+    exit 2
+fi
+reception_dir=${reception_dir_prefix}-002
 
 # Make sure the directory exists
 mkdir -p $workdir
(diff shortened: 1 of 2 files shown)


portante:
@Anisha1234, when work is ready for review, please be sure to also remove the `WIP` label.

If the work is truly still a "work-in-progress", then leave the label, but add a comment referencing the reviewers to take a look even thought it is still a `WIP`.

server/bin/pbench-audit-server.sh
```diff
@@ -78,6 +79,12 @@ tarballs=$workdir/tarballs
 linkdirs=$workdir/linkdirs
 empty=$workdir/empty
 users=$workdir/users
+reception_dir_prefix=$(getconf.py pbench-receive-dir-prefix pbench-server)
+if [ -z "$reception_dir_prefix" ] ;then
```
portante: Please use `[[ ]]` instead of the single version.

server/bin/pbench-audit-server.sh
```diff
@@ -711,6 +718,23 @@ function verify_archive {
     return $cnt
 }
 
+function verify_reception {
```
portante: No sure this code belongs in the audit server, since the reception area is very dynamic, with multiple actors adding files remotely, and the "shim" moving them to the archive area in a non-atomic way.

Can you review the shim code and see where we could add this kind of report there?

Anisha1234:
> @Anisha1234, when work is ready for review, please be sure to also remove the `WIP` label.
> 
> If the work is truly still a "work-in-progress", then leave the label, but add a comment referencing the reviewers to take a look even though it is still a `WIP`.

@portante I have removed the label. Had missed it somehow.

server/bin/pbench-audit-server.sh
Anisha1234: @portante Sure

portante:
> @portante I have removed the label. Had missed it somehow.

Can you also edit the description to remove the reference to this being a work-in-progress?

Going forward, we should probably just write the description to what we want the PR to end up being, and use the `WIP` label until the code changes catch up with what is described.

portante:
See comment in #1191.

PR #1423, opened 2019-12-06
Implements the feature to delete created sessions

Anisha1234:
Implements the feature to delete user-created sessions in Explore.js page
Screenshot (1)


web-server/v0.4/src/pages/Explore/index.js
```diff
@@ -85,16 +96,38 @@ class Explore extends Component {
...
+            dataIndex: '',
+            key: 'start',
+            render: (text, record) => (
+              <Popconfirm
+                title="Delete the dashboard session?"
+                cancelText="No"
+                okText="Yes"
+                onConfirm={() => this.deleteSharedSession(record)}
+              >
+                <Icon type="delete" theme="twoTone" style={{ textAlign: 'right' }} />
```
gurbirkalsi: What are your thoughts on making the `Delete Session` action a separate column as opposed to a nested column with `<Button type="link">Delete Session</Button>` as the rendered element? This approach would help us avoid rendering an additional column header within the view.

web-server/v0.4/src/pages/Explore/index.js
```diff
@@ -85,16 +96,38 @@ class Explore extends Component {
...
+                onConfirm={() => this.startSharedSession(record)}
+              >
+                <Button type="link">Start Session</Button>
+              </Popconfirm>
+            ),
+          },
+          {
+            title: 'Delete Session',
+            dataIndex: '',
+            key: 'start',
```
gurbirkalsi: We should specify a unique `key` value for this column as well.

web-server/v0.4/src/services/explore.js
```diff
@@ -19,3 +19,26 @@ export default async function querySharedSessions(params) {
     },
   });
 }
+export async function deleteSharedSessions(params) {
+  // eslint-disable-next-line no-unused-vars
```
gurbirkalsi: We can remove this commented line since all of the declared vars are being used in the next line.

Anisha1234:
> Can you add a unit test for these changes, and possibly an e2e test?

Sure @portante

web-server/v0.4/src/models/explore.js
```diff
@@ -11,19 +11,15 @@ export default {
   effects: {
     *fetchSharedSessions({ payload }, { call, put }) {
       const response = yield call(querySharedSessions, payload);
-
+      console.log(response);
```
portante: Why do we need this log entry on the console?

Anisha1234: Fixed. Left it by mistake.

web-server/v0.4/package.json
```diff
@@ -50,7 +50,9 @@
     "react-document-title": "^2.0.3",
     "react-dom": "^16.8.3",
     "react-fittext": "^1.0.0",
+    "react-redux": "^7.1.3",
```
gurbirkalsi: Is there new functionality being offered by upgrading/including these dependencies? If so, we should make sure this doesn't break existing functionality in the dashboard.

web-server/v0.4/src/pages/Explore/index.js
```diff
@@ -1,7 +1,7 @@
-import React, { Component } from 'react';
+import React from 'react';
```
gurbirkalsi: We should consider enforcing the use of a destructured reference to `Component` so we can say `class Explore extends Component {` on line 14 to maintain consistent component definition structure throughout the codebase.

web-server/v0.4/src/pages/Explore/index.test.js
```diff
@@ -0,0 +1,29 @@
...
+
+const mockProps = {};
+
+const mockDispatch = jest.fn();
+configure({ adapter: new Adapter() });
+const wrapper = shallow(<Explore.WrappedComponent dispatch={mockDispatch} {...mockProps} />, {
+  disableLifecycleMethods: true,
+});
+
+describe('test Summary page component', () => {
```
gurbirkalsi: We should specify `Explore` page component within this test description.

web-server/v0.4/package.json
Anisha1234: Removed!

web-server/v0.4/src/pages/Explore/index.js
Anisha1234: Done!

web-server/v0.4/src/pages/Explore/index.test.js
Anisha1234: Corrected!

portante:
@Anisha1234, can you rebase this PR?  It is not mergable as it stands now.  But it looks good to go.  Thanks!

Issue #1424, opened 2019-12-06
Render Pagination component in Table component


gurbirkalsi:
#### Summary

The `Table` component renders 10 rows of data by default. In order to allow users to process larger amounts of data at once, a page size changer component is required to enable dynamic rerendering of available rows of data. 

ant-design includes a component for handling page size changing within the `Table` component out of the box called `Pagination` (https://ant.design/components/pagination). This component will be implemented within the `Table` component definition and pass the user selected row size to the `Table` component for reference. 

#### Goals
1. Implement `Pagination` component under the `Table` component definition. 

2. Implement unit tests for verifying the `Pagination` component renders the selected number of rows. 

3. Implement e2e tests for verifying correct `Pagination` component behavior within the dashboard workflow.

PR #1429, opened 2019-12-09
Enable pagination for Table component

aquibbaig:
**Feat-** Allow users to search through various controllers in the dashboard, n(10/20/30) items at a time. 
**Fixes-** #1424

### DIFF
--- a/web-server/v0.4/mock/api.js
+++ b/web-server/v0.4/mock/api.js
@@ -1,22 +1,22 @@
 import config from './datastoreConfig';
+import constants from '../config/constants';
 
-export const mockControllerAggregation = {
+// Generate controllers as per max page size options
+const maxTableSize = parseInt(constants.tableSizeOptions.pop(), 10);
+let generatedBuckets = new Array(maxTableSize).fill({});
+generatedBuckets = generatedBuckets.map((val, index) => {
+  const key = index + 1;
+  return {
+    key: `controller_${key}`,
+    doc_count: key,
+    runs_prev1: { value: null },
+    runs: { value: key, value_as_string: key.toString() },
+  };
+});
+export const generateMockControllerAggregation = {
   aggregations: {
     controllers: {
-      buckets: [
-        {
-          key: 'a_test_controller',
-          doc_count: 1,
-          runs_preV1: { value: null },
-          runs: { value: 1, value_as_string: '1111-11-11' },
-        },
-        {
-          key: 'b_test_controller',
-          doc_count: 2,
-          runs_preV1: { value: null },
-          runs: { value: 2, value_as_string: '2222-22-22' },
-        },
-      ],
+      buckets: generatedBuckets,
     },
   },
 };
--- a/web-server/v0.4/src/components/Table/index.js
+++ b/web-server/v0.4/src/components/Table/index.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import { Table as AntdTable } from 'antd';
+import constants from '../../../config/constants';
 
 const Table = props => {
   const { dataSource, columns, loading, onRow, ...childProps } = props;
@@ -14,6 +15,11 @@ const Table = props => {
       onRow={onRow}
       scroll={{
         x: true,
+      }}
+      pagination={{
+        defaultPageSize: 10,
+        showSizeChanger: true,
+        pageSizeOptions: constants.tableSizeOptions,
       }}
       {...childProps}
     />
--- a/web-server/v0.4/src/e2e/controllers.e2e.js
+++ b/web-server/v0.4/src/e2e/controllers.e2e.js
@@ -1,5 +1,5 @@
 import puppeteer from 'puppeteer';
-import { mockControllerAggregation, mockIndices } from '../../mock/api';
+import { generateMockControllerAggregation, mockIndices } from '../../mock/api';
 
 let browser;
 let page;
@@ -18,7 +18,7 @@ beforeAll(async () => {
         status: 200,
         contentType: 'application/json',
         headers: { 'Access-Control-Allow-Origin': '*' },
-        body: JSON.stringify(mockControllerAggregation),
+        body: JSON.stringify(generateMockControllerAggregation),
       });
     } else if (request.method() === 'GET' && request.url().includes('indices')) {
       request.respond({
@@ -45,7 +45,7 @@ describe('controller page component', () => {
       const testController = await page.$eval('.ant-table-row', elem =>
         elem.getAttribute('data-row-key')
       );
-      expect(testController).toBe('a_test_controller');
+      expect(testController).toBe('controller_1');
       done();
     },
     30000
@@ -59,7 +59,7 @@ describe('controller page component', () => {
     await page.type('.ant-input', testController);
     await page.click('.ant-input-search-button');
     testController = await page.$eval('.ant-table-row', elem => elem.getAttribute('data-row-key'));
-    expect(testController).toBe('a_test_controller');
+    expect(testController).toBe('controller_1');
   });
 
   test('should reset search results', async () => {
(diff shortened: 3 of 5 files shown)


aquibbaig:
@gurbirkalsi @Anisha1234  Needs a minor tweak this component to allow user-defined "tableRows" per page. Shall that be a better idea?

gurbirkalsi:
+1 to @Anisha1234's suggestion on implementing this at a component level vs. page component level. We should consider adding this to our `Table` component definition that extends ant-design's `Table` out of the box. This way, we'll avoid having to pass down the `pagination` prop to every instance of the `Table` component. https://github.com/distributed-system-analysis/pbench/blob/40b5afc64fe3391b45a558be062988dae4660e9c/web-server/v0.4/src/components/Table/index.js#L8-L19

aquibbaig:
Updated!

aquibbaig:
@gurbirkalsi @Anisha1234 Should we limit this pagination component to 5hits/page too?

aquibbaig:
**TODO:** We have 2 mock controllers, so essentially clicking on any pagination item doesn't mean anything in the test environment

aquibbaig:
@gurbirkalsi I refactored the mock `api.js` file to support 26 records cause there are 26 alphabets to support our ascending/descending tests. I think this needs some review to make it cleaner

web-server/v0.4/mock/api.js
```diff
@@ -1,64 +1,60 @@
...
-          'run.config': ['test_size_2'],
-        },
+// Generates mock results
+export const generateMockResults = () => {
+  const mockResultsSchema = {
+    hits: {
+      hits: [],
+    },
+  };
+  for (let i = 0; i <= 25; i += 1) {
```
portante: Running from 0 to 25 inclusively gives us 26 results.  It helps to add a comment that that is intentional.  Alternatively, you can have the loop start at 1 and go to 26, `for (let i = 1; i <= 26; i += 1)` or from zero, `for (let i = 0; i < 26; i += 1)`, either way the `26` is in the loop to make it easier to understand 26 items are expected.

If we agree to have for options as described below, we might want to have the mock instance derive the number of items to generate from the code itself instead of hard-coding the value of `26`.

portante: This comment applies to the loop at line 12 as well.

aquibbaig:
@portante Thanks for the review on this PR. However, we generated 26 items exactly because of the fact that we assert things to come out a certain way which proves the test(E.g. sorting ascending would result in a_test_controller with 100% accuracy, as we have 26 different alphabets). For this reason, we rule out all the third party libraries such as `faker` or `chance.js` to generate these items for us. We can extend pagination to have 50, 100 options if we need to. But we don't need to test that because if it is tested for 20 and 30, it is automatically tested for any number of records.

Also, regarding the last part, could you please explain what is meant by the below lines. It would be really helpful to design a solution ASAP
> If we agree to have for options as described below, we might want to have the mock instance derive the number of items to generate from the code itself instead of hard-coding the value of 26.

portante:
> However, we generated 26 items exactly because of the fact that we assert things to come out a certain way which proves the test (E.g. sorting ascending would result in a_test_controller with 100% accuracy, as we have 26 different alphabets).

Understood.  Please consider the comment made in review further, not from the point of view of having 26 letters in the alphabet, but from the point of view of using hard-coded values in code that are not well commented, or don't lend themselves to a clear understanding of what is happening.

> For this reason, we rule out all the third party libraries such as `faker` or `chance.js` to generate these items for us. We can extend pagination to have 50, 100 options if we need to. But we don't need to test that because if it is tested for 20 and 30, it is automatically tested for any number of records.

If we offer 4 options to choose from instead of 3 options, wouldn't we want to verify in our tests that all for options are made available at the right times (i.e. we don't offer 100 when we have < 50 items, or 50 when we have less than 20 items, etc.

> Also, regarding the last part, could you please explain what is meant by the below lines. It would be really helpful to design a solution ASAP
> 
> > If we agree to have for options as described below, we might want to have the mock instance derive the number of items to generate from the code itself instead of hard-coding the value of 26.

The notion of the mock generating the list of items in two places with a count of 26 is based on a hard-coded constant.  Instead, we should consider having the mock code to generate items that will exercise options for pagination based what those operations are.  For example, to ensure all four options for pagination work for 10, 20, 50, 100, the mock code would inspect the array of values, `[ 10, 20, 50, 100 ]`, and use the last value to determine how many items to generate to ensure we test all the cases.

web-server/v0.4/mock/api.js
```diff
@@ -1,64 +1,64 @@
...
-    index: `${prefix}.0000-00-00`,
-  },
-  {
-    index: `${prefix}.0000-00-01`,
-  },
-];
+// Generates mock indices
+export const generateMockIndices = () => {
+  const indexSchema = [];
+  for (let i = 0; i < 26; i += 1) {
```
portante: Nice, this makes it clear that we are generating 26 items.

Now we need to not hard-code 26 and derive it from the total number of items we need for the pagination schemes we support.

web-server/v0.4/mock/api.js
```diff
@@ -1,64 +1,64 @@
...
-          'run.config': ['test_size_2'],
-        },
+// Generates mock results
+export const generateMockResults = () => {
+  const mockResultsSchema = {
+    hits: {
+      hits: [],
+    },
+  };
+  for (let i = 0; i < 26; i += 1) {
```
portante: Further, since we use the number 26 in hard-coded in two places, having it be driven by a constant or derived from the pagination scheme would make this code much more robust.

aquibbaig: @portante Thanks for your update on this Pull Request. I have updated the number of items to `[10, 20, 50, 100]` now.

I agree that we should not hardcode values as such and generate controllers from the mock. I believe we hardcoded values because the mock objects, ``controllers, indices, results` follow a specific schema as such and there are boundaries on which we can generate these values, for example, we have prefixes to indices and so on. It's a bit tricky to set this up. What I'd like to suggest is for us to have a json api consisting of an array of objects which we can include over here. It would be really helpful if you can share your approach here if I'm missing something. Thanks!

aquibbaig:
Also, the tests are gonna fail for now due to values changed in the array!

web-server/v0.4/mock/api.js
aquibbaig: @portante I believe you want it to function like an API, for example: If we select say 100 items, this value `100` is passed on to the loop and it generates 100 items then and there. So, basically the `mock.js` has functions that act as API endpoints

web-server/v0.4/mock/api.js
```diff
@@ -1,64 +1,114 @@
...
+  const factor = Math.floor(i / 26 + 1);
+  const sequence =
+    i < 26
+      ? `${(i + 10).toString(36)}_${factor.toString()}`
+      : `${((i % 26) + 10).toString(36)}_${factor.toString()}`;
+  return sequence;
+};
+
+/**
+  Genrate mock controllers
```
portante: s/Genrate/Generate/

web-server/v0.4/src/e2e/controllers.e2e.js
```diff
@@ -169,4 +169,43 @@ describe('controller page component', () => {
     },
     30000
   );
+
```
portante: Seems like we don't need to keep this code commented out.  Is there a reason to keep it?

aquibbaig:
@portante @gurbirkalsi We are taking the last element on the pagination option, which ensures that we use the maximum possible table size and we make sure that way that adequate amount of items are generated all the time.

web-server/v0.4/src/components/Table/index.js
```diff
@@ -15,6 +15,11 @@ const Table = props => {
       scroll={{
         x: true,
       }}
+      pagination={{
+        defaultPageSize: 10,
+        showSizeChanger: true,
+        pageSizeOptions: ['10', '20', '50', '100'],
```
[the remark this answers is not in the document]
aquibbaig: @portante fixed!

aquibbaig:
@portante @gurbirkalsi  Thanks for the review on this PR, merge when green!

PR #1430, opened 2019-12-09
Log messages using logging functions in bash scripts.

### LINKED ISSUE #1421: Replace log_init/log_finish with use of the "logger" command
Let's remove the `log_init` / `log_finish` methods and replace with the use of the `logger` command.

riya-17:
Fixes #1421

Server bash scripts used "echo" and output redirection to log info and error messages to local files.
This PR does two things:

- It defines two functions, "log_info" and "log_error", in pbench-base.sh.
- It replaces uses of "echo" for logging purposes with one or the other of these two functions.

Note that the messages have not been changed at all and neither have the destinations: after the
change, the same messages are logged to the same places. This means that there is *no* change to
the gold files used for unit testing.

In a future PR, we are planning to change the logging functions in pbench-base.sh, so that they
follow the policy that is already used by server python scripts: logging to a centralized log
service. That is going to involve changes to pbench-base.sh and virtually all the gold files,
but it will involve *no* changes to any of the bash scripts.

### DIFF
--- a/server/bin/pbench-base.sh
+++ b/server/bin/pbench-base.sh
@@ -138,3 +138,19 @@ function quarantine () {
         fi
     done
 }
+
+function log_info {
+    if [[ -z "${2}" ]]; then
+        printf -- "%b\n" "${1}"
+    else
+        printf -- "%b\n" "${1}" | tee -a "${2}"
+    fi
+}
+
+function log_error {
+    if [[ -z "${2}" ]]; then
+        printf -- "%b\n" "${1}" >&4
+    else
+        printf -- "%b\n" "${1}" | tee -a "${2}" >&4
+    fi
+}
--- a/server/bin/pbench-copy-sosreports.sh
+++ b/server/bin/pbench-copy-sosreports.sh
@@ -64,7 +64,7 @@ if [ "$SOSREPORTDEST" == "" ]; then
     SOSREPORTDEST=$_sosrpt_user@$_sosrpt_host:$_sosrpt_dir/
 fi
 
-echo "$TS: starting at $(timestamp)"
+log_info "$TS: starting at $(timestamp)"
 
 # get the list of files we'll be operating on
 list=$(ls $ARCHIVE/*/$linksrc/*.tar.xz 2>/dev/null)
@@ -81,7 +81,7 @@ index_content=$TMPDIR/index_mail_contents
 for result in $list ;do
     link=$(readlink -e $result)
     if [ ! -f $link ] ;then
-        echo "$TS: $link does not exist" >&4
+        log_error "$TS: $link does not exist"
         nerrs=$nerrs+1
         continue
     fi
@@ -113,7 +113,7 @@ for result in $list ;do
         # sosreport should be RO
         chmod 444 $x
         if [ ! -f $x.md5 ] ;then
-            echo "$TS: FAILED: no MD5 file found for $x" >&4
+            log_error "$TS: FAILED: no MD5 file found for $x"
             nerrs=$nerrs+1
             continue
         fi
@@ -121,7 +121,7 @@ for result in $list ;do
         md5=$(md5sum $x)
         md5=${md5%% *}
         if [ "$md5" != $(cat $x.md5) ] ;then
-            echo "$TS: FAILED: MD5 does not match for  $x" >&4
+            log_error "$TS: FAILED: MD5 does not match for  $x"
             nerrs=$nerrs+1
             continue
         fi
@@ -129,7 +129,7 @@ for result in $list ;do
         cp "$x" "$x".md5 $TMPDIR/
         sts=$?
         if [[ $sts -ne 0 ]] ;then
-            echo "$TS: FAILED: cp "$x" "$x".md5 $TMPDIR/ - code $sts" >&4
+            log_error "$TS: FAILED: cp "$x" "$x".md5 $TMPDIR/ - code $sts"
             nerrs=$nerrs+1
             continue
         fi
@@ -137,15 +137,15 @@ for result in $list ;do
     done
 
     if [[  $nsr == 0 ]] ;then
-        echo "$TS: No sosreports found for $result"
+        log_info "$TS: No sosreports found for $result"
     else
         pushd $incoming > /dev/null 2>&4
         cmd="/usr/bin/rsync -av $TMPDIR/ $SOSREPORTDEST"
-        echo "$TS: $cmd"
+        log_info "$TS: $cmd"
         $cmd
         sts=$?
         if [[ $sts -ne 0 ]] ;then
-            echo "$TS: FAILED: $cmd - code $sts" >&4
+            log_error "$TS: FAILED: $cmd - code $sts"
             nerrs=$nerrs+1
             continue
         fi
@@ -156,18 +156,18 @@ for result in $list ;do
     mv $result $(echo $result | sed "s/$linksrc/$linkdest/")
     sts=$?
     if [[ $sts -ne 0 ]] ;then
-        echo "$TS: Cannot move $result link from $linksrc to $linkdest: code $sts" >&4
+        log_error "$TS: Cannot move $result link from $linksrc to $linkdest: code $sts"
         nerrs=$nerrs+1
         continue
     fi
 
     # log the success
-    echo "$TS: $hostname/$resultname: processed $nsr sosreports for $result"
+    log_info "$TS: $hostname/$resultname: processed $nsr sosreports for $result"
     ntotal=$ntotal+$nsr
     nresults=$nresults+1
 done
 
-echo "$TS: ending at $(timestamp), processed $ntotal sosreports for $nresults results directories with $nerrs errors"
+log_info "$TS: ending at $(timestamp), processed $ntotal sosreports for $nresults results directories with $nerrs errors"
 
 log_finish
 
--- a/server/bin/pbench-edit-prefixes.sh
+++ b/server/bin/pbench-edit-prefixes.sh
@@ -35,7 +35,7 @@ log_init $PROG
 # the link source for this script
 linksrc=TO-LINK
 
-echo $TS
+log_info "$TS"
 
 # get the list of files we'll be operating on
 list=$(ls $ARCHIVE/*/$linksrc/*.tar.xz 2>/dev/null)
@@ -45,7 +45,7 @@ typeset -i nep=0
 for result in $list ;do
     link=$(readlink -e $result)
     if [ ! -f $link ] ;then
-        echo "$TS: $link does not exist" >&4
+        log_error "$TS: $link does not exist" 
         continue
     fi
 
@@ -62,14 +62,14 @@ for result in $list ;do
 
     # find the prefix file and execute it - all the checking was done on the client
     prefix=$ARCHIVE/$hostname/.prefix/prefix.${resultname%.tar.xz}
-    echo $prefix
+    log_info $prefix
     cmds=$(cat $prefix)
-    echo $cmds
+    log_info $cmds
     pushd $RESULTS/$hostname > /dev/null 2>&4
     eval "$cmds"
     status=$?
     if [ $status -ne 0 ] ;then
-        echo "$TS: eval $cmds failed - code $status" >&4
+        log_error "$TS: eval $cmds failed - code $status" 
         continue
     fi
     popd > /dev/null 2>&4
@@ -78,16 +78,16 @@ for result in $list ;do
     rm $result
     status=$?
     if [ $status -ne 0 ] ;then
-        echo "$TS: Cannot remove $result link from $linksrc: code $status" >&4
+        log_error "$TS: Cannot remove $result link from $linksrc: code $status" 
         continue
     fi
 
     # log the success
-    echo "$TS: $hostname/$resultname: success - processed $result"
+    log_info "$TS: $hostname/$resultname: success - processed $result"
     nep=$nep+1
 done
 
-echo "$TS: Processed $nep edit-prefix requests"
+log_info "$TS: Processed $nep edit-prefix requests"
 
 log_finish
 exit 0
--- a/server/bin/pbench-satellite-cleanup.sh
+++ b/server/bin/pbench-satellite-cleanup.sh
@@ -13,7 +13,7 @@ trap "rm -rf $tmp" EXIT
 
 log_init $PROG
 
-echo "$TS: $PROG starting"
+log_info "$TS: $PROG starting"
 
 tarballs=$(cd $ARCHIVE > /dev/null; find . -path '*/TO-DELETE/*.tar.xz' -printf '%P\n' | sort)
 hosts="$(for host in $tarballs ;do echo "${host%%/*}" ;done | sort -u )"
@@ -61,8 +61,7 @@ for host in $hosts; do
                 # Note that if we do remove the tarball but fail to change the
                 # state, this error will persist until the state is changed
                 # manually.
-                echo "$TS: Failed to remove $ARCHIVE/$host/$x, code: $rc" |
-                        tee -a $mail_content >&4
+                log_error "$TS: Failed to remove $ARCHIVE/$host/$x, code: $rc" "${mail_content}"
                 ntberrs=$ntberrs+1
                 popd > /dev/null 2>&4
                 continue
@@ -73,8 +72,7 @@ for host in $hosts; do
             rm $x.md5
             rc=$?
             if [ $rc != 0 ]; then
-                echo "$TS: Failed to remove $ARCHIVE/$host/$x.md5, code: $rc" |
-                        tee -a $mail_content >&4
+                log_error "$TS: Failed to remove $ARCHIVE/$host/$x.md5, code: $rc" "${mail_content}"
                 nmd5errs=$nmd5errs+1
             fi
         fi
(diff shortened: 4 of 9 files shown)


server/bin/pbench-unpack-tarballs.sh
```diff
@@ -136,7 +136,8 @@ function  move_symlink() {
     mv ${ARCHIVE}/${hostname}/${linksrc}/${resultname}.tar.xz ${ARCHIVE}/${hostname}/${linkdest}/${resultname}.tar.xz
     status=${?}
     if [[ ${status} -ne 0 ]]; then
-        echo "${TS}: Cannot move symlink ${ARCHIVE}/${hostname}/${resultname}.tar.xz from ${linksrc} to ${linkdest}: code ${status}" | tee -a ${mail_content} >&4
+        log_error "${TS}: Cannot move symlink ${ARCHIVE}/${hostname}/${resultname}.tar.xz from ${linksrc} to ${linkdest}: code ${status}"
+        status_append "${TS}: Cannot move symlink ${ARCHIVE}/${hostname}/${resultname}.tar.xz from ${linksrc} to ${linkdest}: code ${status}" ${mail_content}
```
portante: Perhaps `status_append` could also issue the log so that you don't repeat code unnecessarily?

riya-17:
Hey @portante  @ndokos please review and tell whether it is on right track or it still needs enhancements

server/bin/pbench-unpack-tarballs.sh
riya-17: There are also some messages that logs only as log_error like:
https://github.com/distributed-system-analysis/pbench/blob/40b5afc64fe3391b45a558be062988dae4660e9c/server/bin/pbench-server-prep-shim-002.sh#L106

and there are messages that logs only in status_update like:
https://github.com/distributed-system-analysis/pbench/blob/40b5afc64fe3391b45a558be062988dae4660e9c/server/bin/pbench-server-prep-shim-002.sh#L200

riya-17: ok

server/bin/pbench-base.sh
```diff
@@ -138,3 +138,23 @@ function quarantine () {
         fi
     done
 }
+
+function log_info {
+    echo $1
+}
+
+function log_error {
```
portante: Can we consider the following for `log_error` instead, so that we don't need `status_append_and_log`?

```
function log_error {
    if [[ -z "${2}" ]]; then
        printf -- "%s\n" "${1}" >&4
    else
        printf -- "%s\n" "${1}" | tee -a "${2}" >&4
    fi
}
```

It is more mechanism to use `printf` instead of `echo`, but it avoids the implied behavior of `echo` with new lines.

This also lets us use `log_error` to specify a file to also add the line to, like:

```
log_error "my error message" "${mail_content}"
```

Where "`my error message`" is printed to `>&4` and to the file name in `${mail_content}`.

server/bin/pbench-unpack-tarballs.sh
```diff
@@ -136,7 +136,7 @@ function  move_symlink() {
     mv ${ARCHIVE}/${hostname}/${linksrc}/${resultname}.tar.xz ${ARCHIVE}/${hostname}/${linkdest}/${resultname}.tar.xz
     status=${?}
     if [[ ${status} -ne 0 ]]; then
-        echo "${TS}: Cannot move symlink ${ARCHIVE}/${hostname}/${resultname}.tar.xz from ${linksrc} to ${linkdest}: code ${status}" | tee -a ${mail_content} >&4
+        status_append_and_log "${TS}: Cannot move symlink ${ARCHIVE}/${hostname}/${resultname}.tar.xz from ${linksrc} to ${linkdest}: code ${status}" ${mail_content} error
```
portante: With the suggestion to use `log_error` as above, this would become:

```
log_error "${TS}: Cannot move symlink ${ARCHIVE}/${hostname}/${resultname}.tar.xz from ${linksrc} to ${linkdest}: code ${status}" "${mail_content}"
```

server/bin/pbench-base.sh
ndokos: `log_info` will have to get the same treatment.

ndokos: Not sure why github won't let me reply to the two comments that ask whether we log the two messages today: the answer is we do log them today; they are sent to stdout within the log_init/log_finish section, so they are redirected to the log file.

server/bin/pbench-unpack-tarballs.sh
```diff
@@ -354,7 +355,7 @@ function do_work() {
         let end_time=$(timestamp-seconds-since-epoch)
         let duration=end_time-start_time
         # log the success
-        echo "${TS}: ${hostname}/${resultname}: success - elapsed time (secs): ${duration} - size (bytes): ${size}"
+        log_info "${TS}: ${hostname}/${resultname}: success - elapsed time (secs): ${duration} - size (bytes): ${size}"
```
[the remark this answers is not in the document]
ndokos: No, but I think it should.

server/bin/pbench-unpack-tarballs.sh
```diff
@@ -382,7 +383,7 @@ while true; do
     do_work ${max_seconds} < ${list}
 done
 
-echo "${TS}: Processed ${ntb} tarballs"
+log_info "${TS}: Processed ${ntb} tarballs"
```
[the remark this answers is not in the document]
ndokos: No, but I think it should. Why not make the change in this commit? There's never going to be a better time.

server/bin/pbench-base.sh
portante: Are there any places we have informational logs saved to the report content?

server/bin/pbench-unpack-tarballs.sh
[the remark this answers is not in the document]
portante: Can we debate that in another PR that makes those changes?  Here, it would be great if her changes can be made without any changes to the unit test output.

server/bin/pbench-unpack-tarballs.sh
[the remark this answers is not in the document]
portante: Can we debate the content of the report in another PR so that this PR can have no changes to the test output?

server/bin/pbench-unpack-tarballs.sh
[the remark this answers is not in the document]
ndokos: OK - we can l;eave it for another PR.

server/bin/pbench-unpack-tarballs.sh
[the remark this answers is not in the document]
ndokos: OK - another PR it is.

server/bin/pbench-base.sh
ndokos: Not in `pench-unpack-tarballs` - maybe there are some in the other scripts, but we can leave `log_info` simple, until and unless we need to complicate it.

riya-17:
> So this looks good: we go to stage 2 where all the other scripts are changed in the same way. Agreed?

yes

server/bin/pbench-copy-sosreports.sh
```diff
@@ -183,3 +183,4 @@ EOF
 pbench-report-status --name ${PROG} --pid ${$} --timestamp $(timestamp) --type status ${index_content}
 
 exit 0
+
```
portante: Added an extra line ...

server/bin/pbench-edit-prefixes.sh
```diff
@@ -78,16 +78,17 @@ for result in $list ;do
...
+    log_info "$TS: $hostname/$resultname: success - processed $result"
     nep=$nep+1
 done
 
-echo "$TS: Processed $nep edit-prefix requests"
+log_info "$TS: Processed $nep edit-prefix requests"
 
 log_finish
 exit 0
+
```
portante: Added an extra line ...

riya-17:
@portante @ndokos pbench-audit-server.sh still needs changes, I will work on that script. Please review rest of the scripts and tell whether is it good to move to next step?

riya-17:
@ndokos @portante Please review the PR

server/bin/pbench-audit-server.sh
```diff
@@ -727,9 +727,11 @@ if [[ $? -ne 0 ]]; then
 fi
 eTS=$(timestamp)
 if [[ -s ${archive_report} ]]; then
-    printf "\nstart-${sTS}: archive hierarchy: $ARCHIVE\n" | tee -a ${report}
+    log_info "" "${report}"
+    log_info "start-${sTS}: archive hierarchy: $ARCHIVE" "${report}"
```
ndokos: Instead of calling `log_info` twice (once with an empty arg), can't you pass the string with `\n` to `log_info`? I really dislike all the multiple calls here and below.

server/bin/pbench-base.sh
```diff
@@ -139,8 +139,16 @@ function quarantine () {
...
+#    printf -- "%s\n" "${1}"
+#}
+
 function log_info {
-    printf -- "%s\n" "${1}"
+    if [[ -z "${2}" ]]; then
+        printf -- "%s\n" "${1}"
+    else
+        printf -- "%s\n" "${1}" | tee -a "${2}"
+    fi
```
ndokos: OK - I see you *did* make the change :-)

server/bin/pbench-audit-server.sh
```diff
@@ -727,26 +730,22 @@ if [[ $? -ne 0 ]]; then
 fi
 eTS=$(timestamp)
 if [[ -s ${archive_report} ]]; then
-    log_info "" "${report}"
-    log_info "start-${sTS}: archive hierarchy: $ARCHIVE" "${report}"
+    log_info "${SingleNewline}start-${sTS}: archive hierarchy: $ARCHIVE" "${report}"
```
ndokos: That's better, but I'd still like to see if we can leave the strings alone and still  have everything work.

server/bin/pbench-sync-satellite.sh
```diff
@@ -270,21 +267,21 @@ done
...
 fi
 
 let end_time=$(timestamp-seconds-since-epoch)
 let duration=end_time-start_time
-echo "$TS: end - $(timestamp)" | tee -a $mail_content
-echo "$TS: duration (secs): $duration" | tee -a $mail_content
-echo "$TS: Total $nprocessed files processed, with $nfailed_md5 md5 failures and $nerrs errors" | tee -a $mail_content
+log_info "$TS: end - $(timestamp)" "${mail_content}"
+log_info "$TS: duration (secs): $duration" "${mail_content}"
+log_info "$TS: Total $nprocessed files processed, with $nfailed_md5 md5 failures and $nerrs errors" "${mail_content}"
```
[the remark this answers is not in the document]
ndokos: Yes, you did!

server/bin/pbench-audit-server.sh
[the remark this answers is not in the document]
ndokos: .. which you did do :-)

ndokos: Try changing the %s in the printf format specification in log_info to %b and see if that causes problems. %b is supposed to expand backslash escape sequences. Read the bash manpage, the section on printf, for details.

server/bin/pbench-audit-server.sh
ndokos: Using %b instead of %s should solve this problem I think.

server/bin/pbench-sync-satellite.sh
```diff
@@ -91,7 +91,7 @@ index_content=$tmp/index_mail_contents
 > $index_content
```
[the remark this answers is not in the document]
riya-17: Yaa, ${index_content} didn't get used

server/bin/pbench-sync-satellite.sh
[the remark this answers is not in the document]
riya-17: Should I make the changes for this in this PR only?

server/bin/pbench-satellite-cleanup.sh
```diff
@@ -159,3 +152,4 @@ cat $mail_content >> $index_content
 pbench-report-status --name ${PROG} --pid ${$} --timestamp $(timestamp) --type status ${index_content}
```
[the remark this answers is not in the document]
riya-17: Is it? Should I make the changes for this in this PR only?

server/bin/pbench-audit-server.sh
```diff
@@ -727,26 +727,22 @@ if [[ $? -ne 0 ]]; then
...
-    log_info "end-${eTS}: archive hierarchy: $ARCHIVE" "${report}"
+    log_info "\nend-${eTS}: archive hierarchy: $ARCHIVE" "${report}"
 fi
 
 sTS=$(timestamp)
 verify_controllers $INCOMING > ${incoming_report} 2>&1
 if [[ $? -ne 0 ]]; then
     let ret=ret+1
 fi
+Newline=$'\n\n'
```
ndokos: This is unnecessary now, no?

server/bin/pbench-base.sh
```diff
@@ -145,9 +145,9 @@ function quarantine () {
 
 function log_info {
     if [[ -z "${2}" ]]; then
-        printf -- "%s\n" "${1}"
+        printf -- "%b\n" "${1}"
     else
-        printf -- "%s\n" "${1}" | tee -a "${2}"
+        printf -- "%b\n" "${1}" | tee -a "${2}"
```
ndokos: We should probably use '%b' in `log_error` as well. At some point, somebody is going to use '\n' in an error message.

server/bin/pbench-audit-server.sh
riya-17: ok

server/bin/pbench-audit-server.sh
riya-17: yes

server/bin/pbench-base.sh
riya-17: ok, sure

server/bin/pbench-satellite-cleanup.sh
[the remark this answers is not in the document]
riya-17: @ndokos In every shell script ${index_content} is only used with pbench-report-status. So, I think this doesn't need to be changed.

server/bin/pbench-sync-satellite.sh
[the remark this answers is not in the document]
ndokos: Actually, it is being used at the very end of the script to add a header to the report,  and it is also used in the function `do_remote_sat_state_change`, so we'll leave it alone.

server/bin/pbench-satellite-cleanup.sh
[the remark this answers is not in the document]
ndokos: You are right: index_content gets initialized with a header and then the accumulated messages in mail_content are appended to it. No need to change anything.

riya-17:
@ndokos @portante Please review the PR

> Apart from resolving a conversation by adding the issue # for the missing log_finish, this all looks good to me.

#1443

server/bin/pbench-audit-server.sh
```diff
@@ -727,9 +727,9 @@ if [[ $? -ne 0 ]]; then
 fi
 eTS=$(timestamp)
 if [[ -s ${archive_report} ]]; then
-    printf "\nstart-${sTS}: archive hierarchy: $ARCHIVE\n" | tee -a ${report}
+    log_info "\nstart-${sTS}: archive hierarchy: $ARCHIVE" "${report}"
```
portante: Are we sure we want to do this kind of change for the audit server code?

This means that new-line characters will get into the log statements.  If we could capture the message without the new-line formatting characters, log that, and then add it to the report with the new-line characters that would be ideal.

However, that may not be worth doing.

Could we consider removing the use of the `tee` command, and just having the `printf` and `cat` commands output to the `${report}`, and then add log_info statements to before the various `$(timestamp)` calls?

server/bin/pbench-sync-package-tarballs.sh
```diff
@@ -59,7 +59,7 @@ calculate_md5_prefix $tar_list > $tmp/files-to-sync
 tar cf - -T $tmp/files-to-sync >&100
 status=$?
 if [[ $status != 0 ]] ;then
-    echo "$TS: Failed: tar --create -T $tmp/files-to-sync" >&4
+    log_error "$TS: Failed: tar --create -T $tmp/files-to-sync"
```
portante: Can we consider using `log_exit` here?

riya-17: ok

server/bin/pbench-audit-server.sh
riya-17: ok

ndokos:
Now that #1440 has been merged, you should rebase this: that will allow the unit tests to pass.

server/bin/pbench-audit-server.sh
riya-17: @portante according to your suggestion I tried printf with cat but it doesn't seems to work. The approaches that I used are these that seems to work a little, other are way apart but each leads to change in gold files:
printf "message" > ${report} | cat ${report}
printf "message" > ${report} | log_info | cat ${report}

ndokos: II would be something like this:

     printf  "message" >>  ${report}
     cat ${archive_report}  >> ${report}
     log_info "some other message"

Basically, the detailed information ends up in the report; the log only has a brief indication of what happened (in a single line, so that it's easier to scan).

ndokos: @portante:  It seems that it's not quite clear yet what we should do with audit, so it might make sense to leave it out of this PR and consider it separately - would it make sense to leave the audit code unchanged in this PR?

portante: @ndokos, that seems like a reasonable approach.

riya-17: @ndokos the suggestion that you provided I tried that as well earlier but that is leading to the change in gold file as well.

Ok. I will leave it unchanged

riya-17:
Done

PR #1433, opened 2019-12-11
Replace `which` cmd with `command -v`

### LINKED ISSUE #1420: Replace "which" in bash scripts with "command -v"
bash has a `command -v` builtin that can replace the `which` command (which is not always available).

sourabhtk37:
- Removed installation of which from Dockerfiles

Signed-off-by: T K Sourab <tsourab@redhat.com>

Fixes #1420

### DIFF
--- a/server/bin/pbench-remote-sync-package-tarballs
+++ b/server/bin/pbench-remote-sync-package-tarballs
@@ -15,6 +15,12 @@ case $opts in
         PROG=$(basename $0)
         ;;
 esac
+
+if [[ -z "${dir}" || -z "${PROG}" ]]; then
+    echo "INTERNAL ERROR: bad environment encountered" >&2
+    exit 1
+fi
+
 case $# in
     2)
         :
@@ -25,10 +31,6 @@ case $# in
         ;;
 esac
 
-if [[ -z "${dir}" || -z "${PROG}" ]]; then
-    echo "INTERNAL ERROR: bad environment encountered" >&2
-    exit 1
-fi
 
 satellite_config=$1
 output_tar_ball=$2
--- a/server/bin/pbench-remote-sync-package-tarballs
+++ b/server/bin/pbench-remote-sync-package-tarballs
@@ -7,8 +7,8 @@
 opts=$SHELLOPTS
 case $opts in
     *xtrace*)
-        dir=$(dirname $(which $0))
-        PROG=$(basename $(which $0))
+        dir=$(dirname $(command -v $0))
+        PROG=$(basename $(command -v $0))
         ;;
     *)
         dir=$(dirname $0)
--- a/server/bin/pbench-remote-sync-package-tarballs
+++ b/server/bin/pbench-remote-sync-package-tarballs
@@ -20,10 +20,15 @@ case $# in
         :
         ;;
     *)
-        echo "Usage: $PROG <satellite-config> <output-tar-ball>"
+        echo "Usage: $PROG <satellite-config> <output-tar-ball>" >&2
         exit 1
         ;;
 esac
+
+if [[ -z "${dir}" || -z "${PROG}" ]]; then
+    echo "INTERNAL ERROR: bad environment encountered" >&2
+    exit 1
+fi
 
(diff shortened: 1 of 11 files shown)


server/bin/pbench-remote-sync-package-tarballs
```diff
@@ -7,8 +7,8 @@
 opts=$SHELLOPTS
 case $opts in
     *xtrace*)
-        dir=$(dirname $(which $0))
-        PROG=$(basename $(which $0))
+        dir=$(dirname $(command -v $0))
+        PROG=$(basename $(command -v $0))
```
ndokos: I don't think this case is tested by *any* unit tests. Can you make sure that it works properly? Manually if necessary, but ideally by adding a unit test.

portante: Perhaps we can just consider adding an assertion that `${dir}` and `${PROG}` are properly defined by line 18.  If we do that, then we'll save ourselves a lot of headaches, and we won't necessarily need a unit test for that.  What do you think, @ndokos?

server/bin/pbench-server-set-result-state
```diff
@@ -5,8 +5,8 @@
 opts=$SHELLOPTS
 case $opts in
     *xtrace*)
-        dir=$(dirname $(which $0))
-        PROG=$(basename $(which $0))
+        dir=$(dirname $(command -v $0))
+        PROG=$(basename $(command -v $0))
```
portante: Same comment as with `pbench-remote-sync-package-tarballs` regarding an assertion being added after line 15 below.

server/bin/pbench-remote-sync-package-tarballs
ndokos: That'd be great.

sourabhtk37: Do you mean to check whether the var are empty?

portante: Yes, so something like the following immediately following line 26:

```
if [[ -z "${dir}" || -z "${PROG}" ]]; then
    echo "INTERNAL ERROR - bad environment encountered" >&2
    exit 1
fi
```

portante: And if you could also fix the `echo` statement on line 23 to send its output to stderr using, "`echo "Usage: ..." >&2`", that would be great.

server/bin/pbench-server-set-result-state
portante: And if you could add the same `case $# in/esac` statement as it exists in `pbench-remote-sync-package-tarballs`, that would be even better.

server/bin/pbench-server-set-result-state
```diff
@@ -5,14 +5,28 @@
...
+    2)
+        :
+        ;;
+    *)
+        echo "Usage: $PROG <satellite-config> <output-tar-ball>" >&2
+        exit 1
+        ;;
+esac
+
+if [[ -z "${dir}" || -z "${PROG}" ]]; then
```
portante: Shouldn't this be checked after line 15 above since `$PROG` is referenced in the above `case` construct?

server/bin/pbench-remote-sync-package-tarballs
```diff
@@ -20,11 +20,16 @@ case $# in
         :
         ;;
     *)
-        echo "Usage: $PROG <satellite-config> <output-tar-ball>"
+        echo "Usage: $PROG <satellite-config> <output-tar-ball>" >&2
         exit 1
         ;;
 esac
 
+if [[ -z "${dir}" || -z "${PROG}" ]]; then
```
portante: Shouldn't this be checked after line 17 above since `$PROG` is referenced in the above `case` construct?

portante:
@ndokos, this looks good now.  Can you review so we can merge this sooner than later to drop the need for `which`?

portante:
Too quick on the merge, would have liked this to have been squashed to one commit.  Sorry I missed that.

PR #1437, opened 2019-12-18
Add Marker 'Bad-Hostname' field for the sosreport tarball

### LINKED ISSUE #1263: Bad sosreport data should not stop all indexing of pbench data
Today if we can't find the `hostname` or `hostname_-F` files in a sosreport tar ball, the entire indexing step for that tar ball will be killed.

Instead, we should add a marker of some sort indicating a bad `hostname` field for the sosreport, so that we can re-index them later.

riya-17:
Fixes #1263 

Added a Marker to the sosreport tar ball if it can't find ```hostname or hostname_-F```

### DIFF
--- a/server/lib/pbench/indexer.py
+++ b/server/lib/pbench/indexer.py
@@ -3109,7 +3109,7 @@ def hostnames_if_ip_from_sosreport(sos_file_name):
         try:
             hostname_f = str(sostb.extractfile(hostname_f_file[0]).read(), 'iso8859-1')[:-1]
         except IOError as e:
-            raise SosreportHostname("Failure to fetch a hostname-f from the sosreport")
+            return (1, "Failure to fetch a hostname-f from the sosreport")
         if hostname_f == 'hostname: Name or service not known':
             hostname_f = ""
     else:
@@ -3119,12 +3119,12 @@ def hostnames_if_ip_from_sosreport(sos_file_name):
         try:
             hostname_s = str(sostb.extractfile(hostname_s_file[0]).read(), 'iso8859-1')[:-1]
         except IOError as e:
-            raise SosreportHostname("Failure to fetch a hostname from the sosreport")
+            return (1, "Failure to fetch a hostname from the sosreport")
     else:
         hostname_s = ""
 
     if not hostname_f and not hostname_s:
-        raise SosreportHostname("We do not have a hostname recorded in the sosreport")
+        return (1, "We do not have a hostname recorded in the sosreport")
 
     # import pdb; pdb.set_trace()
     if hostname_f and hostname_s:
@@ -3144,7 +3144,7 @@ def hostnames_if_ip_from_sosreport(sos_file_name):
         elif hostname_s != "localhost":
             hostname_f = hostname_s
         else:
-            raise SosreportHostname("Can't reconcile short and full hostname")
+            return (1, "Can't reconcile short and full hostname")
 
     elif not hostname_f and hostname_s:
         # The sosreport did not include, or failed to properly collect, the
@@ -3157,7 +3157,7 @@ def hostnames_if_ip_from_sosreport(sos_file_name):
         hostname_s = hostname_f.split('.')[0]
     else:
         # both undefined
-        raise SosreportHostname("Hostname undefined (both short and long)")
+        return (1, "Hostname undefined (both short and long)")
 
     if hostname_f == "localhost" and hostname_s != "localhost":
         hostname_f = hostname_s
@@ -3165,7 +3165,7 @@ def hostnames_if_ip_from_sosreport(sos_file_name):
     elif hostname_f != "localhost" and hostname_s == "localhost":
         hostname_s = hostname_f.split('.')[0]
     elif hostname_f == "localhost" and hostname_s == "localhost":
-        raise SosreportHostname("The sosreport did not collect a hostname other than 'localhost'")
+        return (1, "The sosreport did not collect a hostname other than 'localhost'")
     else:
         pass
 
@@ -3184,7 +3184,7 @@ def hostnames_if_ip_from_sosreport(sos_file_name):
         ip_files = [x for x in ipfiles if x.find('sos_commands/networking/ip_address') >= 0]
         if ip_files:
             d.update(if_ip_from_sosreport(sostb.extractfile(ip_files[0])))
-    return d
+    return (0, d)
 
 def mk_sosreports(tb, extracted_root, logger):
     logger.debug("start")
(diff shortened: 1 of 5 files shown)


server/bin/pbench-copy-sosreports.sh
```diff
@@ -138,6 +138,7 @@ for result in $list ;do
 
     if [[  $nsr == 0 ]] ;then
         echo "$TS: No sosreports found for $result"
+	res="${result}.Bad-Hostname"
```
portante: Not sure what the intention here is, but if it is to make a change on disk for the result tar ball name, that won't really work.

server/lib/pbench/indexer.py
```diff
@@ -3157,15 +3157,15 @@ def hostnames_if_ip_from_sosreport(sos_file_name):
...
+        return _dict_const([('message',"Hostname undefined (both short and long)")])
 
     if hostname_f == "localhost" and hostname_s != "localhost":
         hostname_f = hostname_s
         hostname_s = hostname_f.split('.')[0]
     elif hostname_f != "localhost" and hostname_s == "localhost":
         hostname_s = hostname_f.split('.')[0]
     elif hostname_f == "localhost" and hostname_s == "localhost":
-        raise SosreportHostname("The sosreport did not collect a hostname other than 'localhost'")
+        return _dict_const([('message',"The sosreport did not collect a hostname other than 'localhost'")])
```
ndokos: It's probably better to change the return type of the function to a tuple: (status, payload). The status tells us if everything went OK (in which case the payload is the host_md dict as happens now), or if an unrecoverable error occurred when processing this sosreport (in which case, the payload can be a description of the problem). But in either case, the payload is going to be sent to ES and indexed.

server/lib/pbench/indexer.py
```diff
@@ -3204,6 +3204,11 @@ def mk_sosreports(tb, extracted_root, logger):
             logger.warning("Failed to fetch .md5 of sosreport {}: {}", sos, e)
             continue
         host_md = hostnames_if_ip_from_sosreport(os.path.join(extracted_root, sos))
+        if 'message' in host_md:
+            src = os.path.join(extracted_root, sos)
+            if os.path.exists(src):
+                os.rename(src,src+".BadHostname")
+            raise SosreportHostname(host_md['message'])
```
ndokos: Here, we'll check the status returned: if OK, we'll proceed as before and add the constructed dict to the sosreport list; if the returned status is an error, we'll munge the payload returned (the description of the problem) into a different dictionary (we'll have to decide the exact format) and *also* add the dictionary to the sosreportlist.

The next step is to investigate the *caller(s)* of mk_sosreports: they will get back a list of dicts but some of those will be from *failed* sosreports, so they will need to handle those dicts differently from the current normal handling. But the dicts (good or bad) will be indexed nevertheless, because we handle the bad cases as well as the good cases: we do not raise an exception the way we do today.

ndokos: BTW, we don't want to raise an exception in l3211: that would defeat the purpose. Also, renaming the file is ineffective, because it's a temporary directory: once the indexer is done, the temp directory is going to disappear without a trace.

server/lib/pbench/indexer.py
riya-17: ok

server/lib/pbench/indexer.py
riya-17: ok

server/lib/pbench/indexer.py
```diff
@@ -3109,7 +3109,7 @@ def hostnames_if_ip_from_sosreport(sos_file_name):
         try:
             hostname_f = str(sostb.extractfile(hostname_f_file[0]).read(), 'iso8859-1')[:-1]
         except IOError as e:
-            raise SosreportHostname("Failure to fetch a hostname-f from the sosreport")
+            return ('error',"Failure to fetch a hostname-f from the sosreport")
```
ndokos: The return code part should not be a string: you can always define an [enumeration](https://docs.python.org/3/library/enum.html?highlight=enumeration) for the various  values and return that instead.

server/lib/pbench/indexer.py
```diff
@@ -3205,12 +3205,14 @@ def mk_sosreports(tb, extracted_root, logger):
             continue
         host_md = hostnames_if_ip_from_sosreport(os.path.join(extracted_root, sos))
         # get hostname (short and FQDN) from sosreport
-        d = _dict_const()
-        d['name'] = sos
-        d['md5'] = md5_val
-        d.update(host_md)
-        sosreportlist.append(d)
-
+        if 'OK' in host_md:
```
ndokos: `host_md` is now a tuple, so check with

       if host_md[0] == OK:

where OK is one value of the enumeration.

server/lib/pbench/indexer.py
```diff
@@ -3205,12 +3205,14 @@ def mk_sosreports(tb, extracted_root, logger):
...
+            d = _dict_const()
+            d['name'] = sos
+            d['md5'] = md5_val
+            d.update(host_md)
```
ndokos: Uh, `host_md` is now a tuplel, so you got to select the right element of it to update `d` with.

server/lib/pbench/indexer.py
```diff
@@ -3205,12 +3205,14 @@ def mk_sosreports(tb, extracted_root, logger):
...
+            sosreportlist.append(d)
+        else:
+            sosreportlist.append(host_md)
```
ndokos: This is the error case: what we do here has to be coordinated with what we want to store in ES so that the dashboard will make sense of it.

riya-17: Will change the value of it after discussion. Currently gave a value which I thought would be suitable.

server/lib/pbench/indexer.py
```diff
@@ -3203,14 +3203,17 @@ def mk_sosreports(tb, extracted_root, logger):
...
         d['name'] = sos
         d['md5'] = md5_val
-        d.update(host_md)
-        sosreportlist.append(d)
-
+        if ret_val[0] == 0:
+            d.update(ret_val[1])
+            sosreportlist.append(d)
+        else:
+            d['sos-report-error'] = ret_val[1]
```
ndokos: I'd have spelled it 'sosreport-error', but it's OK either way.  

You  can move the `sosreportlist.append(d) ` after the if-statement: both branches do the same thing.

riya-17: sure

ndokos:
BTW, the approval is for the code changes only. The unit tests still need to be addressed before merging, but I want to test this on staging as well.

server/lib/pbench/indexer.py
```diff
@@ -3109,7 +3109,7 @@ def hostnames_if_ip_from_sosreport(sos_file_name):
         try:
             hostname_f = str(sostb.extractfile(hostname_f_file[0]).read(), 'iso8859-1')[:-1]
         except IOError as e:
-            raise SosreportHostname("Failure to fetch a hostname-f from the sosreport")
+            return (1,"Failure to fetch a hostname-f from the sosreport")
```
portante: Typically, we make sure a space follows a comma.

riya-17:
@portante Should I add a log message in the unit test 7.4 and 7.5 itself if there is any hostname error or should I include a different unit test?

ndokos:
Dashboard unit tests and e2e tests are failing but agent/server unit tests are passing. @gurbirkalsi , @Anisha1234 , @aquibbaig : can you please check?

portante:
Please rebase now that the dashboards tests have been fixed in #1490 (merged).

PR #1440, opened 2019-12-20
Use coreutils for Ubuntu Eoan release in Travis-CI

portante:


portante:
@ndokos, this looks like a simple and effective fix, can you quickly review?  If you can't before the Christmas break, I'll just merge it.

PR #1449, opened 2019-12-24
Replace 'echo "" >&4; log_finish; exit n' with log_exit

### LINKED ISSUE #1446: Replace echo;log_finish;exit with log_exit
Related to work performed in #1444, let's replace the use of the pattern, `echo "" >&4; log_finish; exit 1` with the `log_exit` method from `pbench_base.sh`.

riya-17:
Fixes #1446 

It replace the use of the pattern, echo "" >&4; log_finish; exit 1 with the log_exit method from pbench_base.sh.

### DIFF
--- a/server/bin/pbench-base.sh
+++ b/server/bin/pbench-base.sh
@@ -102,9 +102,13 @@ function log_finish {
 }
 
 function log_exit {
-    echo "$PROG: $1" >&4
+    echo "${PROG}: ${1}" >&4
     log_finish
-    exit 1
+    if [[ -z "${2}" ]]; then
+        exit 1
+    else
+        exit ${2}
+    fi
 }
 
 # Function used by the shims to quarantine problematic tarballs.  It
@@ -120,9 +124,7 @@ function quarantine () {
     sts=$?
     if [ $sts -ne 0 ] ;then
         # log error
-        echo "$TS: quarantine $dest $files: \"mkdir -p $dest/\" failed with status $sts" >&4
-        log_finish
-        exit 101
+        log_exit "$TS: quarantine $dest $files: \"mkdir -p $dest/\" failed with status $sts" 101
     fi
     for afile in ${files} ;do
         if [ ! -e $afile -a ! -L $afile ] ;then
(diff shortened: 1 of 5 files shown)


portante:
A `grep -RF -C 3 log_finish .` in `server/bin` of the tree seems to reveal the following files have that pattern:

 * `pbench-sync-package-tarballs.sh`
 * `pbench-dispatch.sh`
 * `pbench-server-prep-shim-002.sh`
 * `pbench-sync-satellite.sh`

riya-17:
> Are these _all_ the places where the pattern occurs? It seems that there should be more.
> 
> E.g. lines 99 and 107 in pbench-server-prep-shim-002.sh should probably get the treatment.

Yes there are more, I asked for the same in the issue. Because they have different exit code.

riya-17:
> A `grep -RF -C 3 log_finish .` in `server/bin` of the tree seems to reveal the following files have that pattern:
> 
>     * `pbench-sync-package-tarballs.sh`
> 
>     * `pbench-dispatch.sh`
> 
>     * `pbench-server-prep-shim-002.sh`
> 
>     * `pbench-sync-satellite.sh`

Thanks peter for pointing out the file names

server/bin/pbench-base.sh
```diff
@@ -102,9 +102,13 @@ function log_finish {
 }
 
 function log_exit {
-    echo "$PROG: $1" >&4
+    echo "$PROG: ${1}" >&4
```
portante: Can you also use `${PROG}`?

server/bin/pbench-base.sh
```diff
@@ -120,9 +124,7 @@ function quarantine () {
     sts=$?
     if [ $sts -ne 0 ] ;then
         # log error
-        echo "$TS: quarantine $dest $files: \"mkdir -p $dest/\" failed with status $sts" >&4
-        log_finish
-        exit 101
+        log_exit "$TS: quarantine $dest $files: \"mkdir -p $dest/\" failed with status $sts" "101"
```
portante: I don't think you need quotes around the `101` status code.

server/bin/pbench-server-prep-shim-002.sh
```diff
@@ -80,9 +80,7 @@ trap 'rm -rf $tmp' EXIT INT QUIT
 mkdir -p $tmp
 sts=$?
 if [ $sts != 0 ]; then
-    echo "Failed: \"mkdir -p $tmp\", status $sts" >> $errlog
-    log_finish
-    exit 4
+    log_exit "Failed: \"mkdir -p $tmp\", status $sts" >> $errlog
```
portante: It seems the original code was wrong here.  It should have being doing `echo "Failed: ..." >&4` since we have issued the `log_init` already at line 74.

So then this should probably become:

```
log_exit "Failed: \"mkdir -p $tmp\", status $sts" 4
```

PR #1452, opened 2020-01-06
Update to the new context API from legacy context

### LINKED ISSUE #1451: Migrate from legacy context to the new Context API
In the dashboard, we currently use the legacy context API which is to be deprecated soon as per the [official docs](https://reactjs.org/docs/legacy-context.html), which says quote
> Note:
The legacy context API will be removed in a future major version. Use the new context API introduced with version 16.3. The legacy API will continue working for all 16.x releases.

It's safe to configure the new context API for this reason.

aquibbaig:
An initial commit on the way of rendering breadcrumbs which are also passed from the `context`.
Fixes #1451

### DIFF
--- a/web-server/v0.4/src/components/PageHeader/index.js
+++ b/web-server/v0.4/src/components/PageHeader/index.js
@@ -1,8 +1,8 @@
 import React, { PureComponent, createElement } from 'react';
-import PropTypes from 'prop-types';
 import pathToRegexp from 'path-to-regexp';
 import { Breadcrumb, Tabs, Tag } from 'antd';
 import classNames from 'classnames';
+import { BreadcrumbContext } from '@/layouts/BasicLayout';
 import styles from './index.less';
 import urlToList from '@/components/_utils/pathTools';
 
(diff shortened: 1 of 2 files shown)


aquibbaig:
@gurbirkalsi @portante Not sure why the build fails, running `npm test` locally works for me. Also, this particular error has been filed as a bug in the official umi repository and can be found [in this link](https://github.com/umijs/umi/issues/3585).

portante:
@aquibbaig, is there some other syntactic change to the code that prevents `router.js` from loading?  Are there further logs from the test failures that might indicate that?

aquibbaig:
@portante There are some syntactic errors on the console, yes. But those are for any other branch as well, which pass the test. This seems rather strange behaviour.

aquibbaig:
Can you try running a `umi test` or `npm run test` against this branch and any other? To check if everything works perfectly fine.

aquibbaig:
@portante Said by @gurbirkalsi, the context provider may mess up with the default redux provider, for which we are facing such issues. Waiting for further comments on this Pull request as such.

gurbirkalsi:
@aquibbaig Can we consider implementing breadcrumbs using Redux as a medium for passing location props as we discussed earlier? If we find that approach isn't suitable, lets reopen this PR and find a way to update our context API to support that workflow.

PR #1453, opened 2020-01-06
Adds ability to write a description for a session in the dashboard

### LINKED ISSUE #1427: Add ability to write a description for a session in the dashboard
Can we had a prominent UI for adding a description to a saved session in the Dashboard?  It would be great if we prompt the user to add one if they don't give one by default.

Anisha1234:
Fixes #1427 
Implements UI for adding a description to a saved session in the Dashboard.
editDescriptipn

### DIFF
--- a/web-server/v0.4/src/pages/Explore/index.js
+++ b/web-server/v0.4/src/pages/Explore/index.js
@@ -21,6 +21,7 @@ class Explore extends Component {
       sharedSessions: props.sharedSessions,
       visible: false,
       selectedID: '',
+      editedDesc: '',
     };
   }
 
@@ -74,6 +75,12 @@ class Explore extends Component {
     });
   };
 
+  getEditedDesc = value => {
+    this.setState({
+      editedDesc: value,
+    });
+  };
+
   showModal = id => {
     this.setState({
       visible: true,
@@ -82,9 +89,8 @@ class Explore extends Component {
   };
 
   handleSave = () => {
-    const { selectedID } = this.state;
-    const { value } = document.getElementById('editedInput');
-    this.editDescription(selectedID, value);
+    const { selectedID, editedDesc } = this.state;
+    this.editDescription(selectedID, editedDesc);
     this.setState({
       visible: false,
     });
@@ -165,7 +171,11 @@ class Explore extends Component {
             <Button key="back" onClick={this.handleCancel}>
               Cancel
             </Button>,
-            <Button key="submit" type="primary" onClick={this.handleSave}>
+            <Button
+              key="submit"
+              type="primary"
+              onClick={this.getEditedDesc(document.getElementById('editedInput'))}
+            >
               Save
             </Button>,
           ]}
--- a/web-server/v0.4/src/pages/Explore/index.js
+++ b/web-server/v0.4/src/pages/Explore/index.js
@@ -1,10 +1,12 @@
 import React, { Component } from 'react';
 import { connect } from 'dva';
 import { routerRedux } from 'dva/router';
-import { Card, Button, Popconfirm } from 'antd';
+import { Card, Button, Popconfirm, Icon, Form, Modal, Input } from 'antd';
 
 import PageHeaderLayout from '../../layouts/PageHeaderLayout';
 import Table from '@/components/Table';
+
+const { TextArea } = Input;
 
 @connect(({ datastore, explore, loading }) => ({
   sharedSessions: explore.sharedSessions,
@@ -17,6 +19,8 @@ class Explore extends Component {
 
     this.state = {
       sharedSessions: props.sharedSessions,
+      visible: false,
+      selectedID: '',
     };
   }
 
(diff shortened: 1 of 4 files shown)


web-server/v0.4/src/services/explore.js
```diff
@@ -19,3 +19,29 @@ export default async function querySharedSessions(params) {
...
+    data: {
+      query: `
+        mutation($description: String!,$id: ID!) {
+        updateUrl(
+          data: {description: $description}
+          where: {id: $id})
+        {
+          id
+          config
+          description
```
portante: Why do we need a separate method/function for fetching the description when the `querySharedSessions` method/function already returns the description?

portante: Or is this not query, but a write method?  If so, then we need to rename this from `query` to `write` or something to show/document the intended behavior.

Anisha1234: @portante we have been using the `query` keyword in every function which is used for modifying/fetching data in the `service` folder. Should I update the name of the method to `updateDescription` instead?

portante: Yes, that would help, and also add a comment that describes the intended behavior of the method/function.

Anisha1234: @portante @gurbirkalsi I have updated the name of the method to highlight the intended method but not sure if it is going with the naming conventions used in the dashboard. Please let me know if it needs to be changed.

web-server/v0.4/src/services/explore.js
```diff
@@ -19,3 +19,29 @@ export default async function querySharedSessions(params) {
     },
   });
 }
+
+export async function updateDescription(params) {
```
Anisha1234: The `updateDescription` function here take the edited value of description from the `user` and updates it in our database. Later after updating the database, we are using `querySharedSessions` method which is referred as`fetchSharedSessions` method in `pages/Explore/index.js` to re-display the newly updated data.

web-server/v0.4/src/services/explore.js
Anisha1234: > Yes, that would help, and also add a comment that describes the intended behaviour of the method/function.

@portante Done!

web-server/v0.4/src/services/explore.js
portante: Can you add the above as a comment in the code itself?

web-server/v0.4/src/services/explore.js
```diff
@@ -19,3 +19,35 @@ export default async function querySharedSessions(params) {
     },
   });
 }
+
+/* The updateDescription function here take the edited value of description from the user 
+and updates it in the database. */
+
+/* Later after updating the database, we are using querySharedSessions method which is referred as
```
portante: This comment contains references to files which can change over time.  I'd refran from referring to the external file structure in comments.

Instead, consider dropping this entirely and instead adding a comment to `querySharedSessions` about what it does.

Anisha1234: Updated single line comments without any file reference.

web-server/v0.4/src/pages/Explore/index.test.js
```diff
@@ -0,0 +1,29 @@
...
+
+const mockProps = {};
+
+const mockDispatch = jest.fn();
+configure({ adapter: new Adapter() });
+const wrapper = shallow(<Explore.WrappedComponent dispatch={mockDispatch} {...mockProps} />, {
+  disableLifecycleMethods: true,
+});
+
+describe('test Summary page component', () => {
```
gurbirkalsi: We should specify the `Explore` page component here in the test description.

web-server/v0.4/src/pages/Explore/index.js
```diff
@@ -60,8 +64,40 @@ class Explore extends Component {
...
+  showModal = id => {
+    this.setState({
+      visible: true,
+      selectedID: id,
+    });
+  };
+
+  handleSave = () => {
+    const { selectedID } = this.state;
+    const { value } = document.getElementById('editedInput');
```
gurbirkalsi: Can we consider using the React API to modify user input using the component state here?

web-server/v0.4/src/services/explore.js
```diff
@@ -1,6 +1,7 @@
 import request from '../utils/request';
 
-export default async function querySharedSessions(params) {
+// queies all the available shared sessions from the database to display
```
gurbirkalsi: We have a spelling error on this comment.

web-server/v0.4/src/pages/Explore/index.test.js
Anisha1234: Corrected!

web-server/v0.4/src/services/explore.js
Anisha1234: Resolved!

PR #1457, opened 2020-01-13
Merge Configtools into Pbench

riya-17:
Fixes  19(https://github.com/distributed-system-analysis/configtools/issues/19)

I had two ideas to work out this issue: 

- first was to create a separate configtools folder and then make the files to communicate.

- Second mix them with the existing code according to the hierarchy they follow.

So, I tried to implement the 2nd one. Please let me know if this approach is better or the otherwise.


agent/bench-scripts/unittests
```diff
@@ -72,7 +72,7 @@ function remove_path {
 PATH=$(remove_path /opt/pbench-agent/bench-scripts ${PATH})
 PATH=$(remove_path /opt/pbench-agent/util-scripts ${PATH})
 # Allows us to intercept scp, ssh, rsync, etc.
-export PATH=${_testopt}/unittest-scripts:${_testopt}/util-scripts:${_testopt}/bench-scripts:${PATH}
+export PATH=${_testopt}/unittest-scripts:${_testopt}/util-scripts:${_testopt}/bench-scripts:${PWD}/../../bin:${PATH}
```
ndokos: Using $PWD is going to be unreliable I think: for the unittests, we'll have to copy the new hierarchy (@portante suggested `common/bin` and `common/lib`, rather than the bare `bin` and `lib` I had suggested in a previous comment) into the test directory. PATH should be set from there, not from $PWD.

agent/profile
```diff
@@ -24,14 +24,13 @@ then
     echo "                Setup has failed - could not determine the proper environment prefix: \"$prefix\"" >&2
     echo "                You will need to fix it before working with pbench." >&2
 else
-    if [[ -f /opt/configtools/bin/getconf.py ]]
+    if [[ -f $PWD/../bin/getconf.py ]]
```
ndokos: Here too, $PWD is going to be problematic. We should use `$prefix` instead. We can assume that the RPM (or whatever takes its place in the future) creates a `common` hierarchy in /opt/pbench-agent (i.e. where $prefix points) and then we can add `$prefix/common/bin` to PATH.

agent/run-unittests
```diff
@@ -28,6 +28,7 @@ function tail_and_wait {
 }
 
 trap "kill -KILL -$$ > /dev/null 2>&1" INT TERM QUIT
+export PATH=$PWD/../bin:$PATH
```
ndokos: Ditto - although I'm not sure what to use here. But I dont think you should define PATH here: we don't use getconf.py explicitly here. It is only used implicitly by the various `unittests` scripts, so if *they* set the PATH appropriately, then this script does not need to worry about it.

agent/bench-scripts/unittests
riya-17: ok

agent/run-unittests
riya-17: resolved

agent/profile
riya-17: ok

agent/bench-scripts/unittests
```diff
@@ -42,6 +42,10 @@ for script in ${_tdir}/../util-scripts/pbench-* ; do
     cp ${script} ${_testopt}/util-scripts/
     let res=res+${?}
 done
+cp -a $_tdir/../../bin/getconf.py $_testopt/common/bin
```
ndokos: Probably better to copy `$_tdir/../..bin/*` just in case we add more things there in the future.

agent/bench-scripts/unittests
```diff
@@ -42,6 +42,10 @@ for script in ${_tdir}/../util-scripts/pbench-* ; do
...
+let res=res+${?}
+cp -a $_tdir/../../lib/configtools/__init__.py $_testopt/common/lib/configtools
```
ndokos: Here do  `cp -a $_tdir/../../lib/* $_testopt/common/lib/` - again, just to make sure that when we add additional libraries, we'll copy them all. This also implies that when you do the mkdir on line 27, you only need to create `$_testopt/common/lib` - the rest of the hierarchy will be created by the `cp -a`.

agent/bench-scripts/unittests
```diff
@@ -72,7 +76,7 @@ function remove_path {
 PATH=$(remove_path /opt/pbench-agent/bench-scripts ${PATH})
 PATH=$(remove_path /opt/pbench-agent/util-scripts ${PATH})
 # Allows us to intercept scp, ssh, rsync, etc.
-export PATH=${_testopt}/unittest-scripts:${_testopt}/util-scripts:${_testopt}/bench-scripts:${PATH}
+export PATH=${_testopt}/unittest-scripts:${_testopt}/util-scripts:${_testopt}/bench-scripts:${_testopt}/common/bin:${PATH}
```
ndokos: Don't we need an enhanced PYTHONPATH?

agent/profile
```diff
@@ -24,13 +24,7 @@ then
...
     echo "                You will need to fix it before working with pbench." >&2
 else
-    if [[ -f /opt/configtools/bin/getconf.py ]]
-    then
-	pathins /opt/configtools/bin
-	# no need to set PYTHONPATH if configtools is installed
-    else
-	pathins $prefix/configtools/bin
-    fi
+    pathins $prefix/common/bin
```
ndokos: Don't we need an enhanced PYTHONPATH?

agent/util-scripts/unittests
```diff
@@ -21,12 +21,16 @@ export _testdir=${_testroot}/pbench
...
+mkdir -p ${_testopt}/config ${_testopt}/util-scripts ${_testopt}/tool-scripts ${_testopt}/lib ${_testopt}/unittest-scripts ${_testopt}/common/bin ${_testopt}/common/lib/configtools
 let res=res+${?}
 cp -L ${_tdir}/../base ${_testopt}/
 let res=res+${?}
 cp -rL ${_tdir}/../config/* ${_testopt}/config/
 let res=res+${?}
+cp -a $_tdir/../../bin/getconf.py $_testopt/common/bin
+let res=res+${?}
+cp -a $_tdir/../../lib/configtools/__init__.py $_testopt/common/lib/configtools
+let res=res+${?}
```
ndokos: All the comments in `bench-scripts/unittests` are applicable here too.

server/bin/pbench-trampoline
```diff
@@ -14,7 +14,7 @@ else
     printf -- "${_basename}: pbench-trampoline needs python3, either directly or through SCL\n" >&2
     exit 1
 fi
-export PYTHONPATH=$(dirname ${_dirname})/lib
+export PYTHONPATH=$(dirname ${_dirname})/lib:$(dirname ${_dirname})/common/bin:$(dirname ${_dirname})/common/lib/configtools:$PYTHONPATH
```
ndokos: So maybe we don't need an enhanced PYTHONPATH in the unittests scripts and the base file: this takes care of it.

agent/profile
ndokos: Maybe not  here - it *is* enhanced in the trampoline code.

agent/bench-scripts/unittests
ndokos: Maybe not  here - it *is* enhanced in the trampoline code.

server/bin/unittests
```diff
@@ -255,6 +259,10 @@ function _setup_state {
     let res=res+$?
     cp -a $_tdir/{job_pool.sh,unittests,pbench*} $_testopt_sat/bin
     let res=res+$?
+    cp -a $_tdir/../../bin/getconf.py $_testopt_com/bin
+    let res=res+$?
+    cp -a $_tdir/../../lib/configtools/__init__.py $_testopt_com/lib/configtools
+    let res=res+$?
```
ndokos: See comments in `bench-scripts/unittests`.

agent/profile
ndokos: ... except the trampoline code is on the server, so maybe we *do* need it on the agent - I feel dizzy :-)

ndokos:
BTW, after the unit tests pass locally, in order for Travis CI to test this properly, you should modify `.travis.yml` to *NOT* install configtools.

agent/util-scripts/unittests
riya-17: ok

server/bin/pbench-trampoline
riya-17: ok

server/bin/unittests
riya-17: ok

agent/profile
ndokos: I think you do need PYTHONPATH - if you don't think so, can you explain why not?

agent/bench-scripts/unittests
ndokos: I think you do need PYTHONPATH - if you don't think so, can you explain why not?

agent/util-scripts/unittests
```diff
@@ -86,7 +90,7 @@ function remove_path {
 PATH=$(remove_path /opt/pbench-agent/bench-scripts ${PATH})
 PATH=$(remove_path /opt/pbench-agent/util-scripts ${PATH})
 # Allows us to intercept scp, ssh, rsync, etc.
-export PATH=${_testopt}/unittest-scripts:${_testopt}/util-scripts:${_testopt}/tool-scripts:${PATH}
+export PATH=${_testopt}/unittest-scripts:${_testopt}/util-scripts:${_testopt}/tool-scripts:${_testopt}/common/bin:${PATH}
```
ndokos: PYTHONPATH?

server/bin/unittests
```diff
@@ -299,7 +307,7 @@ function _setup_state {
     # All the "real" scripts are found at $_testopt/bin, the mock scripts
     # are found in $_testopt/unittest-scripts.
     _orig_PATH=$PATH
-    export PATH=$_testopt/unittest-scripts:$_testopt/bin:$PATH
+    export PATH=$_testopt/unittest-scripts:$_testopt/bin:$_testopt_com/bin:$PATH
```
ndokos: PYTHONPATH?

.travis.yml
```diff
@@ -24,7 +23,7 @@ jobs:
       before_install:
         - sudo apt-get update
       install:
-        - pip install 'configtools<0.4.0' elasticsearch sh boto3
+        - pip install urllib3 elasticsearch sh boto3
```
ndokos: Why is urllib3 now necessary?

riya-17: Travis is throwing error without it and asked to install

agent/profile
```diff
@@ -24,14 +24,9 @@ then
...
 
+    export PYTHONPATH=${_testopt}/common/lib:${PYTHONPATH}
```
ndokos: Oops! This won't work here: you need to use ${prefix}.

.travis.yml
portante: This seems weird.  Can you capture the error and report it back here?

riya-17:
@portante the error is not getting produced now. Can go without urllib3.

Issue #1425, opened 2019-12-06
Render Breadcrumbs component in PageHeader component


gurbirkalsi:
#### Summary
pbench dashboard utilizes UmiJS for generating routing configurations in the `pages` directory. In order to improve navigation throughout the platform, breadcrumbs are required for routing users to specific phases of the pbench dashboard workflow. For example, the user can navigate back to the result of their choice instead of having to start from the initial stage of the workflow by selecting a controller. 

ant-design includes a component for handling breadcrumb navigation out of the box called `Breadcrumbs` (https://ant.design/components/breadcrumb) for rendering available routes given a data structure that contains the necessary information. This location data is stored within the Redux state and is made available by connectors to the Redux state within page components. 

`Breadcrumbs` component will be included in the `PageHeader` component as a static global UI element and reference location data via props passed down from the parent page component (e.g. `Controllers`, `Results`, etc).

#### Goals
1. Implement ant-design `Breadcrumb` component within `PageHeader` component for rendering available location data routes

2. Implement unit tests for verifying location data is valid and user clicks to the `Breadcrumb` component route users to the appropriate page

3. Implement e2e tests for verifying correct behavior for breadcrumbs within the dashboard workflow.

PR #1461, opened 2020-01-16
Render breadcrumbs 

aquibbaig:
Previously, we were using `routes` in PageHeader component, however they were always undefined because they were not passed at the first place from top components. 

Also, there was a small bug where we placed a root pathname '/' which was preventing breadcrumbs from rendering in the first place.
That renders breadcrumbs for the pages present in the sidebar, hence is a step towards fixing #1425.

### DIFF
--- a/web-server/v0.4/src/common/menu.js
+++ b/web-server/v0.4/src/common/menu.js
@@ -5,6 +5,28 @@ const menuData = [
     name: 'Dashboard',
     icon: 'dashboard',
     path: '/',
+    routes: [
+      {
+        name: 'Results',
+        path: '/results',
+        routes: [
+          {
+            name: 'Summary',
+            path: '/summary',
+          },
+          {
+            name: 'Comparison Select',
+            path: '/comparison-select',
+            routes: [
+              {
+                name: 'Comparison',
+                path: '/comparison',
+              },
+            ],
+          },
+        ],
+      },
+    ],
   },
   {
     name: 'Search',
@@ -18,7 +40,7 @@ const menuData = [
   },
 ];
 
-function formatter(data, parentPath = '/', parentAuthority) {
+function formatter(data, parentPath = '', parentAuthority) {
   return data.map(item => {
     let { path } = item;
     if (!isUrl(path)) {
@@ -29,8 +51,8 @@ function formatter(data, parentPath = '/', parentAuthority) {
       path,
       authority: item.authority || parentAuthority,
     };
-    if (item.children) {
-      result.children = formatter(item.children, `${parentPath}${item.path}/`, item.authority);
+    if (item.routes) {
+      result.routes = formatter(item.routes, `${parentPath}${item.path}`, item.authority);
     }
     return result;
   });
--- a/web-server/v0.4/src/components/PageHeader/index.js
+++ b/web-server/v0.4/src/components/PageHeader/index.js
@@ -1,23 +1,29 @@
 import React, { PureComponent, createElement } from 'react';
 import PropTypes from 'prop-types';
-import pathToRegexp from 'path-to-regexp';
 import { Breadcrumb, Tabs, Tag } from 'antd';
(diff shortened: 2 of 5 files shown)


aquibbaig:
Refactored existing methods

web-server/v0.4/src/components/PageHeader/index.js
```diff
@@ -1,23 +1,49 @@
...
+  if (breadcrumbNameMap[url]) {
+    breadcrumb = breadcrumbNameMap[url];
+  } else {
     Object.keys(breadcrumbNameMap).forEach(item => {
-      if (pathToRegexp(item).test(url)) {
-        breadcrumb = breadcrumbNameMap[item];
+      if (item.includes(url)) {
+        // Returns the route if final path correlates with current path
+        if (
+          item.split('/')[item.split('/').length - 1] === url.split('/')[url.split('/').length - 1]
```
portante: So this appears to execute the split 4 times instead of just twice.

Consider something like:

```
let item_comps = item.split('/');
let url_comps = url.split('/');
if (item_comps[item_comps.length - 1] === url_comps[url_comps.length - 1]) {
...
```

web-server/v0.4/src/components/PageHeader/index.js
```diff
@@ -1,23 +1,49 @@
...
     });
   }
-  return breadcrumb || {};
-}
+  return breadcrumb;
+};
+
+// Returns the matching path key from breadcrumbNameMap
+// based upon current location
+const getPathKey = (breadcrumbNameMap, url) => {
```
portante: Can we refactor this method and the above method to share code? They seem almost identical.

web-server/v0.4/src/common/menu.js
```diff
@@ -5,6 +5,18 @@ const menuData = [
     name: 'Dashboard',
     icon: 'dashboard',
     path: '/',
+    routes: [
```
gurbirkalsi: This is a great start, we'll also want to add routes for the remaining pages in the dashboard:

```javascript
{
  name: 'Results',
  path: '/results',
  routes: [
    {
      name: 'Summary',
      path: '/summary',
    },
    {
      name: 'Comparison Select',
      path: '/comparison-select',
      routes: [
        {
          name: 'Comparison',
          path: '/comparison'
        }
      ]
    }
  ],
}
```

web-server/v0.4/src/components/PageHeader/index.js
gurbirkalsi: +1, we should be able to call `getPathKey()` in `getBreadcrumb()`

web-server/v0.4/src/components/PageHeader/index.js
```diff
@@ -96,12 +122,14 @@ export default class PageHeader extends PureComponent {
 
   conversionFromLocation = (routerLocation, breadcrumbNameMap) => {
     const { breadcrumbSeparator, linkElement = 'a' } = this.props;
-    // Convert the url to an array
-    const pathSnippets = urlToList(routerLocation.pathname);
+    const pathKeyFromRoutes = getPathKey(breadcrumbNameMap, routerLocation.pathname);
+    // PathSnaippets store the user navigation history
```
gurbirkalsi: Spelling

web-server/v0.4/src/components/PageHeader/index.js
aquibbaig: Sounds great!

web-server/v0.4/src/components/PageHeader/index.js
aquibbaig: Absolutely!

aquibbaig:
@gurbirkalsi Thanks for the review. Seems like there was a bug in getting the link from url's which was taking us to routes which are `undefined`. I have fixed it by adding a util function to get a link from the urls.

PR #1466, opened 2020-01-24
fix maximum depth error on Explore page

### LINKED ISSUE #1465: Explore page throws "Maximum update depth exceeded."
## Error trace
Screenshot from 2020-01-24 12-12-51
```
Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
```

## File trace - `src/pages/explore/index.js, Line 214`
```
we set state on getEditedDesc,
onClick={this.getEditedDesc(document.getElementById('editedInput'))}
```
This causes state to be set and render the component, again gets 'editedInput' from DOM, setsState, renders and so on, going into an infinite loop.

aquibbaig:
Fixes #1465

### DIFF
--- a/web-server/v0.4/src/pages/Explore/index.js
+++ b/web-server/v0.4/src/pages/Explore/index.js
@@ -211,7 +211,7 @@ class Explore extends Component {
             <Button
               key="submit"
               type="primary"
-              onClick={this.getEditedDesc(document.getElementById('editedInput'))}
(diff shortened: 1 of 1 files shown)


aquibbaig:
@gurbirkalsi Rebased and updated!

PR #1468, opened 2020-01-24
import link, fix side bar collapse issue

### LINKED ISSUE #1467: Mobile view throws "Link is not imported" and SideBar doesn't collapse
`Link` component is undefined inside `GlobalHeader` component and since it appears only in mobile view it is difficult to trace. 
Also, the `Sidebar` doesn't collapse on click.

aquibbaig:
Fixes #1467

### DIFF
--- a/web-server/v0.4/src/components/SiderMenu/index.js
+++ b/web-server/v0.4/src/components/SiderMenu/index.js
@@ -9,7 +9,6 @@ const SiderMenuWrapper = props => {
     <DrawerMenu
       getContainer={null}
       level={null}
-      handleChild={<i className="drawer-handle-icon" />}
       onHandleClick={() => {
         props.onCollapse(!collapsed);
       }}
--- a/web-server/v0.4/src/layouts/BasicLayout.js
+++ b/web-server/v0.4/src/layouts/BasicLayout.js
@@ -224,7 +224,7 @@ class BasicLayout extends React.PureComponent {
           // you will be forced to jump to the 403 interface without permission
           Authorized={Authorized}
           menuData={getMenuData()}
-          collapsed={collapsed !== true}
+          collapsed={collapsed}
           // eslint-disable-next-line no-restricted-globals
           location={location}
           isMobile={mb}
@@ -235,7 +235,7 @@ class BasicLayout extends React.PureComponent {
             <GlobalHeader
               logo={logo}
               fetchingNotices={fetchingNotices}
-              collapsed={collapsed}
+              collapsed={collapsed !== true}
               datastoreConfig={datastoreConfig}
               savingSession={savingSession}
               sessionBannerVisible={sessionBannerVisible}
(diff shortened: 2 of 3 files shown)


web-server/v0.4/src/components/SiderMenu/index.js
```diff
@@ -9,9 +9,9 @@ const SiderMenuWrapper = props => {
     <DrawerMenu
       getContainer={null}
       level={null}
-      handleChild={<i className="drawer-handle-icon" />}
+      handlechild={<i className="drawer-handle-icon" />}
       onHandleClick={() => {
-        props.onCollapse(!collapsed);
+        props.onCollapse(collapsed);
```
portante: Can you add a comment for why this is "NOT collapsed"?

web-server/v0.4/src/components/SiderMenu/index.js
```diff
@@ -9,9 +9,9 @@ const SiderMenuWrapper = props => {
     <DrawerMenu
       getContainer={null}
       level={null}
-      handleChild={<i className="drawer-handle-icon" />}
+      handlechild={<i className="drawer-handle-icon" />}
```
portante: This seems like an interface inconsistency with the `DrawerMenu`.  Can you confirm?  The GitHub repo does not seem to have any references to `handlechild`.

aquibbaig: Yes, absolutely. There is no such method there and we do not use it in our code as well. So removing that.

web-server/v0.4/src/components/SiderMenu/index.js
aquibbaig: There were certain inconsistencies in our code regarding props passed from BasicLayout on to the SiderMenu. I have made some minor refactors to it which makes it easy to predict the collapse behaviour now!

aquibbaig:
@portante Thanks for the review. Fixed!!

aquibbaig:
@gurbirkalsi Rebased, updated

web-server/v0.4/src/layouts/BasicLayout.js
```diff
@@ -224,7 +224,7 @@ class BasicLayout extends React.PureComponent {
           // you will be forced to jump to the 403 interface without permission
           Authorized={Authorized}
           menuData={getMenuData()}
-          collapsed={collapsed !== true}
+          collapsed={collapsed}
```
gurbirkalsi: Can we consider reverting the default collapsed state to already collapsed? This allows for more space to view tables and data throughout the platform.

aquibbaig: @gurbirkalsi testing this out I can find that `collapsed` variable is misleading and passing `!collapsed` actually sets default state to collapsed,  unlike the other one

aquibbaig: Wait a minute, Let me test it out again though

aquibbaig: @gurbirkalsi You're right, my bad!

gurbirkalsi:
@aquibbaig It looks like the default state for `collapsed` is still being set to false. Can you submit another PR to address this issue?

PR #1478, opened 2020-01-29
pbench-run-benchmark linpack support

k-rister:


k-rister:
DO NOT MERGE, not yet complete.

k-rister:
This PR is focused on add support for running linpack via pbench-run-benchmark.  During the course of implementation several minor issues with the pbench-run-benchmark infrastructure have been encountered and fixes for those are included.

k-rister:
Most notably, there is no postprocessing yet for the linpack results which still needs to be done.

k-rister:
I think this is good to go now, the CI failures are something weird with the dashboard which this doesn't touch.

portante:
@k-rister, can you rebase this patch set and squash them down to a smaller number of related commits?

Also PR #1490 seems to have fixed the failing CI tests for the most part, so a rebase would be great.  I'll review in the mean time.

agent/bench-scripts/pbench-run-benchmark
```diff
@@ -360,8 +369,11 @@ while (scalar @param_sets > 0) {
         $iteration_id++;
     }
     if ($pp_only) {
-        print BULK_FH "echo Waiting for $nr_pp_jobs post-processing jobs to finish\n";
-        print BULK_FH "wait\n";
+	if ($params{'postprocess-mode'} eq 'cdm') {
```
portante: Seems like a tabs vs spaces formatting disagreement here.

agent/bench-scripts/pbench-run-benchmark-sample
```diff
@@ -216,12 +220,15 @@ if (exists $specs{"controller"}{"post-script"} and $specs{"controller"}{"post-sc
...
     printf { $fh_cmd } "%s\n", $postprocess_cmd;
     close($fh_cmd);
-    my $postprocess_output = `. $pp_cmd_file`;
-    my $rc = $?;
-    open(my $fh_output, ">" . $sample_dir . "/postprocess-output.txt");
-    printf { $fh_output } "%s\n", $postprocess_output;
+    my $postprocess_output = `. $pp_cmd_file 2>&1`;
+    my $rc = $? >> 8;
+    if (open(my $fh_output, ">" . $sample_dir . "/postprocess-output.txt")) {
+	printf { $fh_output } "%s\n\n", $postprocess_output;
```
portante: Seems like tabs vs spaces formatting issues here.

agent/bench-scripts/pbench-run-benchmark
k-rister: I'm not an expert in this area...but isn't this type of editor behavior supposed to be controlled by these lines at the top of the files:

```
#!/usr/bin/perl
# -*- mode: perl; indent-tabs-mode: t; perl-indent-level: 4 -*-
# vim: autoindent tabstop=4 shiftwidth=4 expandtab softtabstop=4 filetype=perl
```

If so, does that mean these not correct?

agent/bench-scripts/pbench-run-benchmark-sample
k-rister: See above...

k-rister:
I rebased the code and squashed the linpack related commits together.  I left the bug fixes as individual commits since that seems appropriate.

agent/bench-scripts/pbench-run-benchmark
portante: It looks like we have a fight going on with tabs and spaces, where we have tabs in the file, but have `expandtab` in the `vim` settings, but don't have that specified in the `emacs` settings.

I have opened issue #1500 to propose a path forward to fix this.

agent/bench-scripts/pbench-run-benchmark-sample
portante: See #1500.

PR #1480, opened 2020-01-30
Fix pagination e2e bug

### LINKED ISSUE #1482: Pagination e2e tests inconsistent across instances
As we can see, the e2e tests for the pagination component are failing across some test instances due to inconsistent behavior of the page with respect to the pagination element. What we need to do is that we need to synchronize the behavior to be exact same and expected across all the tests

aquibbaig:
I had a look at [this](https://github.com/puppeteer/puppeteer/issues/1805) issue in the official puppeteer docs. To fix this, they use `waitFor()`. I have used `waitForSelector()` method to wait till the element is visible to the page, also `waitFor()` before selecting rows. Fixes #1482.

### DIFF
--- a/web-server/v0.4/src/e2e/controllers.e2e.js
+++ b/web-server/v0.4/src/e2e/controllers.e2e.js
@@ -176,22 +176,37 @@ describe('controller page component', () => {
   });
 
   test('should display 20 controllers on preference', async () => {
+    await page.waitForSelector(
+      '.ant-pagination-options > .ant-pagination-options-size-changer > .ant-select-selection > .ant-select-selection__rendered > .ant-select-selection-selected-value'
+    );
     await page.click('.ant-select-selection-selected-value');
-    await page.click('.ant-select-dropdown-menu-item:nth-child(2)');
+    await page.waitForSelector('.ant-select-dropdown-menu-item:nth-child(2)');
+    const elementToClick = await page.$('.ant-select-dropdown-menu-item:nth-child(2)');
+    await page.evaluate(el => el.click(), elementToClick);
     const rows = await page.$$('.ant-table-row');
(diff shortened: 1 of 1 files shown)


aquibbaig:
Thanks @gurbirkalsi for recommending `puppeteer-selector`

mnmehta:
Hi Aquib, I have been running into this issue and tried your fix, but I don't think the fix is working. Since the issue only happens sporadically, the best way I have found of really stressing this is to take the last 3 dropdown tests in the e2e/controllers.e2e.js file (i.e. the 20,50,100 dropdowns) and replicate 100 more copies of those lines in the file, so that the test will do 100 attempts of the 20,50,100 dropdown tests. You will find that at least a few will fail, even after waiting for selectors and injecting your delays. (By the way you have no delay after the 50 dropdown, only after the 20 and 100 dropdowns, not sure why). I'm not sure what the underlying issue is, I'm wondering if there is a bug in the puppetteer/chrome code where it is really selecting the wrong dropdown for some reason.

aquibbaig:
@mnmehta Thanks for the insight! And good catch, the wait for selectors was missed by mistake for `50 items`, so let me add that right away. And let me figure out another way of doing this and yes, seems to me this is a bug.

aquibbaig:
@mnmehta Let me know if this fix works. I used the reference from [here](https://github.com/puppeteer/puppeteer/issues/1484).

aquibbaig:
@portante @gurbirkalsi @mnmehta  I tested this against 80-90 cases and all of them pass for now. Please have a check from your end as well. That'd be great!

mnmehta:
Great work @aquibbaig, 300 back to back dropdown selections work now with no failures, before this I would usually see 7-10 failures.

PR #1481, opened 2020-01-31
Fix cleanup of child procs in web-server e2etests

mnmehta:
The existing cleanup was not killing the web server, makes the 2nd test in a back to back test hang.


mnmehta:
Will resubmit this PR, it seems to be pulling along my old PR for some reason,

PR #1489, opened 2020-02-06
S3 backup large object fixes

### LINKED ISSUE #1415: Fix the S3 backup for large objects.
PR #1260 implemented a new backup method to an S3 service, but there are problems with sending large objects (> 5GB) to it. PR #1338 made S3 backup optional to work around this issue, but we need to solve it.

ndokos:
Fixes #1415

- The S3Connector call to the s3client.upload_fileobj() method was using
  named params for everything, but the first three should be positional
  (body, bucket, key),

- The ETag field of the object returned by get_object() is a string that
  is wrapped in double quotes, whereas the calculated ETag is a string
  without quotes. Fortunately, the calculation is correct: apart from
  the double quotes, the two strings are identical.

- The mock S3 connector required some changes: the unit tests were not
  passing because of "large object" failures. It turns out that the
  double quote disease had spread to several places (in particular
  to the function that is called locally to calculate the ETag of
  a tarball, which was inconsistent).

  The current setup is as follows:
  + When we calculate locally the ETag of a tarball, we do *NOT* include
    double quotes in the returned value.
  + In the mock case, we simulate multi-upload by writing an ETag file
    when we "upload" the tarball - we add the double quotes in that file
    to simulate as precisely as possible what S3 would do in the real case.
  + In the mock case, the mock get_object() returns the double-quoted ETag
    that it reads from the previously written file, just as the real get_object()
    returns a double-quoted ETag from S3.
  + So when we do a get_object(), the returned ETag field is double-quoted (both
    in the real and in the mock case), so we *strip* the double quotes before
    comparing it to the locally calculated value.

Also added some additional debug logging.


portante:
Please rebase now that the dashboards tests have been fixed in #1490 (merged).

grafuls:
Reviewed, lgtm.

portante:
@grafuls, we rely on the review mechanism provided by GitHub.  So if you click on the "Files Changed" tab, you can then click the "Review changes" button and "Approve" the commit to finalize your review.

server/bin/pbench-backup-tarballs.py
```diff
@@ -212,6 +213,7 @@ def backup_to_local(lb_obj, logger, controller_path, controller, tb,
             else:
                 tar_done = True
 
+        logger.debug("End local backup of {}.".format(tar))
         if md5_done and tar_done:
             logger.info(
                 "Locally backed-up sucessfully: {tarball}".format(tarball=os.path.join(controller, resultname)))
```
portante: Can you also change this wording to something along the lines of what you added in the other log statements?

Perhaps: `"Local backup of {tarball} successful".format(tarball=...)`

server/bin/test-bin/test-find-behavior
```diff
@@ -62,4 +62,5 @@ declare -A ub=(
 for bucket in small medium large huge; do
     printf "\nBucket: %s -- %s <= size < %s\n" "${bucket}" "${lb[${bucket}]}" "${ub[${bucket}]}"
     sort ${bucket}.lis
+    rm ${bucket}.lis
```
portante: Yeah!

server/bin/test-bin/test-find-behavior
```diff
@@ -62,4 +62,5 @@ declare -A ub=(
 for bucket in small medium large huge; do
     printf "\nBucket: %s -- %s <= size < %s\n" "${bucket}" "${lb[${bucket}]}" "${ub[${bucket}]}"
     sort ${bucket}.lis
```
riya-17: why do we need to sort it before removing?

portante: Mostly because of how our unit tests work.  The tests are based on "gold" file comparisons, which means if the order of those buckets change in how they are processed, then the gold file comparisons will fail.

We sort all the buckets so that they maintain the same order independent of the environment in which the tests run.

PR #1490, opened 2020-02-06
Upgrade and lock eslint dependencies for consistent linting across environments 

gurbirkalsi:
**Summary**

Previously, eslint dependencies were using semantic versioning for handling package installation via `package.json`. By locking dependency versions, we can ensure that the correct version will be referenced in any environment. This change also includes fixes for the `import/order` error presented from the linter process by reordering import statements for affected .js files.


web-server/v0.4/package.json
```diff
@@ -11,7 +11,7 @@
     "site": "umi-api-doc static && gh-pages -d dist",
     "analyze": "cross-env ANALYZE=true umi build",
     "lint:style": "stylelint \"src/**/*.less\" --syntax less",
-    "lint": "eslint --ext .js src && npm run lint:style",
+    "lint": "eslint --ext .js src mock && npm run lint:style",
```
gurbirkalsi: These changes now also run the linter against the `mock` directory

PR #1496, opened 2020-02-10
Add basic Python structure

zulcss:
Add basic python infrastructure like a setup.py, tox.ini
(for running unitests, lint checkers, etc) and external packaging
requirements. This will make life easier for end users, allow pbench to
be installable via pypi, and make packager's life easier in the long
run.

Also included is a pbench library directory so we can eventually share
code between pbench-server and pbench-agent.

This is the first step in creating a unified python3 based interface
for pbench-agent.

### DIFF
--- a/pbench/tests/base.py
+++ b/pbench/tests/base.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+
+# Copyright 2010-2011 OpenStack Foundation
+# Copyright (c) 2013 Hewlett-Packard Development Company, L.P.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+from oslotest import base
+
+
+class TestCase(base.BaseTestCase):
+
+    """Test case base class for all unit tests."""
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,34 @@
+[tox]
+minversion = 2.0
+envlist = py36,pep8
+skipsdist = True
+
+[testenv]
+usedevelop = True
+setenv =
+   PYTHONWARNINGS=default::DeprecationWarning
+   OS_STDOUT_CAPTURE=1
+   OS_STDERR_CAPTURE=1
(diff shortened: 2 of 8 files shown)


zulcss:
Well there is a couple of ways that this can be done, I think:

* same namespace for server and agent - Share common code between agent and server, binaries are taken care of via entry-points in setup.py.

pro - keep existing CI infrastructure. simpler packaging
con - lots of code churn, with moving python bits to a common namespace,  possible tech debt

* Breaking up server, agent, dashboard into own repos/code bases - server, agent, and dashboard become separate projects

pro - cleaner code bases, easier to remvoe tech debt,
cons - making CI a bit more complicated, possible RPM dependency hell/dependency hell can be mittigated with using python packages, and/or containers

My preference would be breaking out server, agent, and dashboard into separate packages.

ndokos:
I agree that we should break it up into separate repos (and separate packages, although the latter is the case already at the RPM level for agent and server  - not sure how the dashboard is packaged): there is very little common code between agent and server (the only thing I can think of is the configtools library, which is just one small file). I don't think it would affect dependencies one way or the other (but maybe I'm misunderstanding something) and the CI would actually get simpler: breakages in server unit tests would not affect the agent or dashboard and similarly for the other cases: today, we are occasionally blocked from merging something in one component because the unit tests for another component are failing; the CI would also be quicker. Finally, making a server or an agent RPM would be quicker, since only  the relevant repo would need to be cloned: today, I get everything and end up throwing away much of it  when making an RPM.

portante:
@zulcss, @ndokos, can we move the repo break-up discussion to #1497?

pbench/tests/base.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+
+# Copyright 2010-2011 OpenStack Foundation
+# Copyright (c) 2013 Hewlett-Packard Development Company, L.P.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
```
grafuls: Do we need this here? setup.py states we are using GPLv3

tox.ini
```diff
@@ -0,0 +1,34 @@
...
+   OS_STDOUT_CAPTURE=1
+   OS_STDERR_CAPTURE=1
+   OS_TEST_TIMEOUT=60
+deps = -r{toxinidir}/requirements.txt
+       -r{toxinidir}/test-requirements.txt
+commands = stestr run {posargs}
+
+[testenv:lower-constraints]
+basepython = python3
+deps = -c{toxinidir}/lower-constraints.txt
```
grafuls: don't we need this file too? lower-constraints.txt

zulcss:
Better commit coming, please ignore this branch

PR #12, opened 2015-07-20
This adds stddev results to pbench_fio

atheurer:
This is implmented in very similar was as pbench_uperf:
By default 5 samples are taken to compute stddev

PR #1499, opened 2020-02-11
More S3 large object fixes.

### LINKED ISSUE #1415: Fix the S3 backup for large objects.

ndokos:
Fixes #1415.

This PR is a continuation of the already merged #1489.

There have been substantial changes since the last review. Comments have been addressed. The PR has been reorganized into multiple commits:

- a commit to implement the delayed formatting of the logging messages.
- changes to gold files for non-backup and non-veriify tests, caused by additional logging in the backup programs and library.
- changes to test-5.2 because of additional logging (split out into a separate commit because the changes are more extensive).
- changes to the gold files for the backup tests (test-6*) both because of logging and because of fixes to the backup programs and library (including the mock S3 implementation). A state file was changed as well.
- changes to the gold files for the verify tests (test-9.*)  because of logging changes and the  fixes to the backup programs and library (including the mock S3 implementation). These tests also had bugs and there are changed to just about all the state files.
- Fixes to tthe backup programs and library (including the mock S3 implementation).
- README file added describing  some details of the S3 API and the S3 mock implementation.

This last commit is probably the best starting place after the (more or less trivial) first three commits have been examined.


ndokos:
Not sure why, but @grafuls does not appear in the potential reviewer list.

server/bin/unittests
```diff
@@ -594,8 +597,8 @@ declare -A cmds=(
     # service.
 
     # check for successful local backup if S3 is disabled
-    [test-6.13]="_run pbench-backup-tarballs"
     [test-6.14]="_run pbench-backup-tarballs"
+    [test-6.15]="_run pbench-backup-tarballs"
```
portante: Why not just add the new test as `test-6.15` and avoid all the churn?

server/bin/pbench-backup-tarballs.py
```diff
@@ -46,15 +46,15 @@ def __init__(self, ntotal=0, nbackup_success=0, nbackup_fail=0,
 def sanity_check(lb_obj, s3_obj, config, logger):
     # make sure archive is present
     archive = config.ARCHIVE
+    archivepath= os.path.realpath(archive)
 
-    if not os.path.realpath(archive):
-        logger.error(
-            'The ARCHIVE directory {}, does not resolve to a real location'.format(archive))
+    if not archivepath:
+        logger.error('The ARCHIVE directory {}, does not resolve to a real location', archive)
```
grafuls: Just nitpicking here but for the sake of readability I would rather use the python3 f-strings.
logger.error(f'The ARCHIVE directory {archive}, does not resolve to a real location')

server/bin/pbench-backup-tarballs.py
```diff
@@ -240,18 +242,26 @@ def backup_to_s3(s3_obj, logger, controller_path, controller, tb, tar, resultnam
     except Exception:
         s3_md5 = None
     else:
-        s3_md5 = obj['ETag'].strip("\"")
+        if obj['Metadata']:
```
grafuls: What if there is no metadata key on the obj? You might get an unhandled KeyError exception.
Would use the get method instead which returns None if not there, which will then evaluate to False. 

obj.get('Metadata')

server/lib/s3backup/__init__.py
```diff
@@ -154,7 +154,7 @@ def put_tarball(self, Name=None, Body=None, Size=0, ContentMD5=None, Bucket=None
                     })
             except ClientError as e:
                 self.logger.error(
```
grafuls: Here you could use self.logger.exception to print out the stack trace.

server/bin/pbench-backup-tarballs.py
grafuls: Now I see from the s3backup/__init__.py that the Metadata key would always be there, still if an empty dict is return from the get, this will also evaluate to False.

server/bin/pbench-backup-tarballs.py
portante: If the logging level is set such that the message won't be emitted, won't using f-strings incurs all the formatting without being useful?

As a rule, we are trying to avoid pre-formatted strings in log messages.

server/lib/s3backup/__init__.py
portante: We are actively trying to avoid stack traces in the log messages when handling "known" error conditions.  This is why we just print the nature of the exception in the error message.  In other places where we don't expect failures, we dump the whole exception (e.g. line 133).

server/bin/pbench-backup-tarballs.py
```diff
@@ -240,18 +242,26 @@ def backup_to_s3(s3_obj, logger, controller_path, controller, tb, tar, resultnam
     except Exception:
         s3_md5 = None
```
portante: If we get an Exception here, is that normal?  We might want to consider logging something.

server/bin/pbench-backup-tarballs.py
portante: Seems best to be defensive here, and verify we got back a `dict` with a `Metadata` member.

ndokos: The code has been moved into a function in the s3backup module and appropriate defensive measures have been added.

server/bin/pbench-backup-tarballs.py
ndokos: Yes and no: this is actually fixed by the last commit in the series. If the key is not found, that should be normal (but it was not, hence the commit). However, there may be other exceptions that are not normal.

ndokos:
This needs squashing, but I believe that it's ready for review.

ndokos:
I introduced a problem in the "normal" (small object) case that is now fixed. 

I have  reorganized and squashed commits:

- The first commit introduces additional logging.
- The second commit fixes the unit tests to take the additional logging into account.
- The next three commits change gold files: first, existing gold files associated with `pbench-backup-tarballs` (test-6.*), then existing gold files associated with `pbench-verify-backup-tarballs` (test-9.*), and finally new gold files for both programs to handle the large object case. I'm going to squash these three to a single commit before merging, but I left them separate for review purposes.
- The next commit has all the code changes ot `pbench-backup-tarballs`, `pbench-verify-backup-tarballs` and the `s3backlup` module (including the mocked S3 implementation).
- The final commit includes just a README file, describing the mock S3 implementation and providing examples of the real implementation to illustrate points, It should probably be read before trying to figure out what is happening in the unit tests.

PTAL.

server/bin/pbench-backup-tarballs.py
```diff
@@ -254,10 +256,13 @@ def backup_to_s3(s3_obj, logger, controller_path, controller, tb, tar, resultnam
         else:
             logger.error(
                 "The tarball {} is already present in S3 bucket, but with different MD5", s3_resultname)
+            # logger.debug(
```
portante: Not sure we need this commented out.

server/bin/pbench-verify-backup-tarballs.py
```diff
@@ -120,10 +120,12 @@ def s3_entry_list_creation(self):
             while True:
```
portante: Can we use a variable here?  Infinite loops can be a bit nerve racking to look at at time.

server/bin/pbench-verify-backup-tarballs.py
```diff
@@ -120,10 +120,12 @@ def s3_entry_list_creation(self):
...
                 resp = self.s3_config_obj.list_objects(**kwargs)
                 for obj in resp['Contents']:
-                    md5_returned = obj['ETag'].strip("\"")
+                    md5_returned = self.s3_config_obj.header_md5(obj)
                     self.content_list.append(Entry(obj['Key'], md5_returned))
+                self.logger.debug('list_objects: got {} objects', len(resp['Contents']))
                 try:
                     kwargs['ContinuationToken'] = resp['NextContinuationToken']
+                    self.logger.debug('list_objects: continuing...')
```
portante: If we use a variable to exit the loop, then we can calculate the that variable and use that variable to collapse the debug statement at line 128 and 125 so that we don't emit two of them.

server/bin/pbench-verify-backup-tarballs.py
```diff
@@ -398,12 +411,14 @@ def main():
...
             reportfp.seek(0)
 
+            logger.debug('Sending report: start')
             report = Report(config, _NAME_)
             report.init_report_template()
             try:
                 report.post_status(config.timestamp(), "status", reportfp.name)
             except Exception:
                 pass
+            logger.debug('Sending report: end')
```
portante: We have added a lot of debug statements.  Is debug the default logging level by default?

server/lib/s3backup/__init__.py
```diff
@@ -189,6 +183,35 @@ def head_bucket(self, Bucket=None):
...
+    # by get_tarball_header().
+    @staticmethod
+    def s3_md5(tbh):
+        if 'Metadata' in tbh and tbh['Metadata']:
+            if 'md5' in tbh['Metadata']:
+                return tbh['Metadata']['md5']
+            else:
+                return None
+        else:
+            return tbh['ETag'].strip('"')
```
portante: Is an `ETag` value the same as a MD5 value?  This seems confusing.

server/lib/s3backup/__init__.py
```diff
@@ -189,6 +183,35 @@ def head_bucket(self, Bucket=None):
     def list_objects(self, **kwargs):
         return self.connector.list_objects(**kwargs)
 
+    # response decoding - tbh is the header returned
+    # by get_tarball_header().
```
portante: In some places, line at line 199 below, the comment for the method follows the function declaration line.  In this case it is before.  Seems like it should follow the function definition.  What do you think?

server/lib/s3backup/__init__.py
```diff
@@ -189,6 +183,35 @@ def head_bucket(self, Bucket=None):
...
+    @staticmethod
+    def s3_md5(tbh):
```
portante: How many places is `s3_md5` called from?  Seems like it is just one, so it is worth having a method for it?

server/lib/s3backup/__init__.py
```diff
@@ -246,6 +267,30 @@ def __init__(self, config):
...
                                              self.secret_access_key),
                                          endpoint_url='{}'.format(self.endpoint_url))
+        self.logger = logger
+
+    @staticmethod
+    def calculate_multipart_etag(tb, chunk_size):
+        md5s = []
+
+        with open(tb, 'rb') as fp:
+            while True:
```
portante: Why not have a for loop?

```
for data in fp.read(chunk_size):
    md5s.append(hashlib.md5(data))
```

server/lib/s3backup/__init__.py
```diff
@@ -246,6 +267,30 @@ def __init__(self, config):
...
+                    break
+                md5s.append(hashlib.md5(data))
+
+        if len(md5s) > 1:
+            digests = b"".join(m.digest() for m in md5s)
+            new_md5 = hashlib.md5(digests)
+            new_etag = '{}-{}'.format(new_md5.hexdigest(), len(md5s))
+        elif len(md5s) == 1:
+            # file smaller than chunk size
+            new_etag = '{}'.format(md5s[0].hexdigest())
```
portante: Isn't a `.hexdigest()` already a string object?  So this could be written as:

```
new_etag = md5s[0].hexdigest()
```

server/bin/gold/test-9.2.txt
```diff
@@ -198,13 +199,22 @@ end-1970-01-01T00:00:00-UTC: archive hierarchy: /var/tmp/pbench-test-server/test
...
 1970-01-01T00:00:00.000000 DEBUG pbench-verify-backup-tarballs.pbench-verify-backup-tarballs main -- Finished S3 list (<Status.SUCCESS: 10>)
 1970-01-01T00:00:00.000000 DEBUG pbench-verify-backup-tarballs.pbench-verify-backup-tarballs main -- Checking MD5 signatures of archive
 1970-01-01T00:00:00.000000 DEBUG pbench-verify-backup-tarballs.pbench-verify-backup-tarballs main -- Finished checking MD5 signatures of archive
 1970-01-01T00:00:00.000000 DEBUG pbench-verify-backup-tarballs.pbench-verify-backup-tarballs main -- Checking MD5 signatures of local backup
 1970-01-01T00:00:00.000000 DEBUG pbench-verify-backup-tarballs.pbench-verify-backup-tarballs main -- Finished checking MD5 signatures of local backup
+1970-01-01T00:00:00.000000 DEBUG pbench-verify-backup-tarballs.pbench-verify-backup-tarballs main -- Comparing ARCHIVE with BACKUP: start
+1970-01-01T00:00:00.000000 DEBUG pbench-verify-backup-tarballs.pbench-verify-backup-tarballs compare_entry_lists -- controller/fio__2016-08-16_22:03:11.tar.xz: present in ARCHIVE but not in BACKUP#012
```
portante: I wonder how the trailing `#012` was emitted.

server/bin/gold/test-9.3.txt
```diff
@@ -198,13 +199,21 @@ end-1970-01-01T00:00:00-UTC: archive hierarchy: /var/tmp/pbench-test-server/test
...
 1970-01-01T00:00:00.000000 DEBUG pbench-verify-backup-tarballs.pbench-verify-backup-tarballs main -- Finished S3 list (<Status.SUCCESS: 10>)
 1970-01-01T00:00:00.000000 DEBUG pbench-verify-backup-tarballs.pbench-verify-backup-tarballs main -- Checking MD5 signatures of archive
 1970-01-01T00:00:00.000000 DEBUG pbench-verify-backup-tarballs.pbench-verify-backup-tarballs main -- Finished checking MD5 signatures of archive
 1970-01-01T00:00:00.000000 DEBUG pbench-verify-backup-tarballs.pbench-verify-backup-tarballs main -- Checking MD5 signatures of local backup
 1970-01-01T00:00:00.000000 DEBUG pbench-verify-backup-tarballs.pbench-verify-backup-tarballs main -- Finished checking MD5 signatures of local backup
+1970-01-01T00:00:00.000000 DEBUG pbench-verify-backup-tarballs.pbench-verify-backup-tarballs main -- Comparing ARCHIVE with BACKUP: start
+1970-01-01T00:00:00.000000 DEBUG pbench-verify-backup-tarballs.pbench-verify-backup-tarballs compare_entry_lists -- controller/fio__2016-08-16_22:03:11.tar.xz: present in BACKUP but not in ARCHIVE#012
```
portante: Also has a trailing `#012`.

server/lib/s3backup/README
```diff
@@ -0,0 +1,238 @@
...
+=list_objects()= to treat the bucket as a large bucket, so the first
+time it gets called (without a =ContinuationToken=), it returns the
+first entry of however many it finds in the S3 "store", and sets the
+=IsTruncated= and =NextContinuationToken= fields in the response. The
+second time it is called (with a =ContinuationToken=), it returns the
+rest of the entries without a =NextContinuationToken=. It does not
+check for the "proper" continuation token (and it recalculates things
+inefficiently, but we don't worry about that).
+
+** Unit tests
```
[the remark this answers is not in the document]
ndokos: This was actually my scratch pad to jot down what each test was supposed to do. It really does not belong to the README file, so I should have deleted it, particularly since there are some questions that I did not answer. Do you think it's useful enough to put into unitttests, or should I just delete it?

server/bin/pbench-backup-tarballs.py
ndokos: Deleted.

server/bin/pbench-verify-backup-tarballs.py
ndokos: Sure.

server/bin/pbench-verify-backup-tarballs.py
ndokos: Yes, it is. We probably want it at DEBUG for unit tests and staging, but we can turn it down to INFO for production. One possibility is to specify it in the config file (a general default and perhaps individual defaults for each program).

server/lib/s3backup/__init__.py
ndokos: Two: pbench-backup-tarballs.py and header_md5() in the S3Config class. And even if there was only one, I'd want to hide what it does :-)

server/lib/s3backup/__init__.py
ndokos: Uhh - let the record show an embarassed silence...

server/lib/s3backup/__init__.py
ndokos: Yes.

server/bin/gold/test-9.2.txt
ndokos: I don't know - I'll check,

server/bin/gold/test-9.3.txt
ndokos: I'll check.

server/lib/s3backup/__init__.py
ndokos: Not always: it depends on whether it's a large object or not.

server/lib/s3backup/__init__.py
ndokos: OK.

server/lib/s3backup/__init__.py
ndokos: I moved the comment but expanded it also: let me know what you think of it.

server/bin/gold/test-9.2.txt
ndokos: See lines, 205, 212, 221, 232 and 240 in pbench_veriy_backup_tarballs.py:compare_entry_lists(): a string is constructed with a '\n` and then written to the report as well as logger.debug()ed.

server/bin/pbench-verify-backup-tarballs.py
ndokos: OK - thanks.

server/lib/s3backup/README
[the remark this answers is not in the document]
ndokos: I deleted the section: I don't think it contained anything useful to move to unittests.

server/bin/pbench-backup-tarballs.py
```diff
@@ -237,12 +236,15 @@ def backup_to_s3(s3_obj, logger, controller_path, controller, tb, tar, resultnam
...
     # Check if the result already present in s3 or not
     try:
-        obj = s3_obj.get_tarball_header(Bucket=s3_obj.bucket_name,
+        tbh = s3_obj.get_tarball_header(Bucket=s3_obj.bucket_name,
                                         Key='{}'.format(s3_resultname))
-    except Exception:
+    except NoSuchKey:
+        s3_md5 = None
+    except Exception as e:
+        logger.error(e)
```
portante: This is not a new code path, but it does seem like a logic error.  Why would a general `Exception` cause us to consider the tar ball missing from the S3 instance?  Seems like for this tar ball, we'd only want to report the error talking to S3, and move to the next tar ball to consider.

server/lib/s3backup/__init__.py
```diff
@@ -88,38 +92,23 @@ def getsize(self, tar):
...
-            new_md5 = hashlib.md5(digests)
-            new_etag = '{}-{}'.format(new_md5.hexdigest(), len(md5s))
-        elif len(md5s) == 1:
-            # file smaller than chunk size
-            new_etag = '{}'.format(md5s[0].hexdigest())
+        try:
+            resp = self.connector.get_object(Bucket=Bucket, Key=Key)
+        except ClientError as e:
+            if e.response['Error']['Code'] == 'NoSuchKey':
+                raise NoSuchKey
```
portante: Should we add an `else` with a raise?

ndokos:
@portante: I added two new unit tests to test other errors and modifed the mock get_object() to generate those errors. One thing I don't understand is that I had to modify the error logging  in those case in `backup_to_s3` to use `str(e)` explicitly (see [here](https://github.com/ndokos/pbench/blob/6b69bd300d426b4404057ab491597be1d06fc12f/server/bin/pbench-backup-tarballs.py#L244) ) otherwise I would get tracebacks.

server/bin/pbench-backup-tarballs.py
```diff
@@ -230,38 +231,42 @@ def backup_to_s3(s3_obj, logger, controller_path, controller, tb, tar, resultnam
...
     try:
-        obj = s3_obj.get_tarball_header(Bucket=s3_obj.bucket_name,
-                                        Key='{}'.format(s3_resultname))
-    except Exception:
+        tbh = s3_obj.get_tarball_header(Bucket=s3_obj.bucket_name,
+                                        Key=s3_resultname)
+    except NoSuchKey:
         s3_md5 = None
+    except Exception as e:
+        logger.error(str(e))
```
portante: Could we add an error message that gives some context, like: `logger.error("Error fetching tar ball header from S3: {}", e)`?

server/lib/s3backup/__init__.py
```diff
@@ -22,6 +22,13 @@ class Status(Enum):
     SUCCESS = 0
     FAIL = 1
 
+class NoSuchKey(Exception):
+    pass
+
+# a generic exception for testing - only raised by the mocked get_object()
+# so that get_tarball_header() can be tested.
+class Get_Object_Exception(Exception):
```
portante: This exception appears to be only used by the "mock" code, can you consider renaming it to `MockGetObjectException`?  And also move the comment as a doc string?

ndokos:
Once this passes, I'm going to squash it down to two commits.

Issue #1352, opened 2019-09-18
Need to index existence of tar ball even if it cannot be indexed properly.


ndokos:
ATM, we give up completely on tar balls that don't meet some minimal criteria (e.g. existence and goodness of metadata.log file, good start run date, good hostnames in sosreports etc). But we want to use ES as the authority for the data that are available, not just the data that are good enough to be indexed. So we need to modify pbench-index.py to index at least a minimal run document that will allow us to get at the tarball, even if it fails to index properly.

riya-17:
Isn't this issue has already been addressed?

dbutenhof:
No, it hasn't; we do the tarball consistency checks and throw an exception without doing any indexing if it doesn't pass.

We do create a `Dataset` ... but so far the data queries don't report a dataset that has a `Dataset` but no Elasticsearch presence. We _could_ change that, e.g. by doing a `Dataset` query for controller/name match that's in `QUARANTINE` state (which such a dataset would be) in addition to the Elasticsearch query. I'd considered using my `Dataset` DB table to look up controllers in the first place rather than Elasticsearch, when I added users; I didn't mostly because we use ES because of the fast replicated index and search, and bypassing that felt like going in the wrong direction.

But a more direct interpretation of this issue would be that we create an Elasticsearch `run-data` document, at least, even if we quarantine the tarball because it's "bad". That suggests doing the consistency checks in stages, and erroring out only when we get to an indexing stage we can't handle (for example, pulling out the iteration data)... but presumably returning a new error counter for "partially indexed" tarballs which would be reported and rolled into the `server-report` document.

PR #1502, opened 2020-02-13
fixing of tarballs and errors occured in tar ball due to which a Tarball is unable to get indexed properly

riya-17:
Fixes issue #1352 
This commit deals with BadDate error.

### DIFF
--- a/server/lib/pbench/indexer.py
+++ b/server/lib/pbench/indexer.py
@@ -769,15 +769,13 @@ class PbenchData(object):
             orig_ts_float = float(orig_ts)
         except Exception as e:
             self.counters['ts_not_epoch_millis_float'] += 1
-            raise BadDate("{!r} is not a float in milliseconds since the"
-                    " epoch: {}".format(orig_ts, e))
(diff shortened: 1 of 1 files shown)


riya-17:
> Can you explain more about what you are trying to do here? It seems that there is a particular error that you want to solve, but not sure how the commit message matches the changes that I see in the code.
> 
> I have marked it with the label, `WIP`, for now.

I am trying to deal with with BadDate Error. removing the raised errors and replacing with a tuple which shows the error being raised with the help of 0 and 1. Planning to add a new field into the indexer to display that error with a default timestamp "1970-01-01T00:00:00.000000". 

The changes that are being displayed in pbench-index.py are to display those error which cannot be dealt in indexing. I have commented them out at present but will be building a function to add them in indexing.

portante:
@riya-17, @ndokos, in the future, for a PR like this one (one where a fix is being performed related to another issue), having a clear problem statement in the description of the PR, where we describe what the problem is, how it is manifested, and how the code changes in the commit addresses that problem, would be helpful.

PR #1505, opened 2020-02-17
Pbench REST for tarballs transfer

grafuls:
First prototype for sending tar files via rest api in chunks.
Added pbench-server as a systemctl service. (Requires changes to RPM spec)

feat: pbench-server REST API 
feat: agent side of transfer REST
feat: moved util-scripts to own utils dir 
feat: moved read in chunks and post to copy_results_tb
reverted util-scripts changes
fixed revert
feat: refactored pbench-make-result-tb to python
feat: converted copy-results-tb 
fix: formatting on logger


grafuls:
These 2 tests suits can be run via:
CONFIG=agent/config/pbench-agent-default.cfg python -m pytest agent/utils/tests
CONFIG=server/lib/config/pbench-server-default.cfg python -m pytest server/utils/tests

server/utils/__init__.py
```diff
@@ -0,0 +1 @@
+
```
portante: Nit: not sure we need a blank line in this file.

server/utils/config.py
```diff
@@ -0,0 +1,5 @@
+
+import configtools
+
+opts, _ = configtools.parse_args()
```
portante: What do we gain making the `opts` and `conf` be module level variables?

agent/utils/add_metalog_option.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python3
```
portante: Could we consider making an `agent/lib/pbench` module path, and placing all these "utils" there as modules, and then have actual commands in `agent/util-scripts` invoke them?

server/utils/config.py
grafuls: Just separation of concerns for code readability and potential use on additional libraries.

agent/utils/add_metalog_option.py
grafuls: That wouldn't be a problem, though `agent/lib` seems to be just perl scripts, didn't want to mix those.

grafuls:
What's going on here? PR seems to have one less commit as in my branch [here](https://github.com/distributed-system-analysis/pbench/compare/master...grafuls:pbench-server).
Also the travis job which was successful doesn't appear to have been updated on this PR.

server/utils/config.py
portante: Could we consider merging this convenience function into the `lib/configtools` module entirely, adding something like the following there?

```
class _SimpleOpts(object):
    def __init__(self, conf_file=None):
        self.filename = conf_file
def get_conf(conf_file=None):
    conf, _ = init(_SimpleOpts(conf_file), '_PBENCH_SERVER_CONFIG')
    return conf
```

This way we use the sample server config environment variable, and we avoid parsing options intended for the `getconf.py` interfaces, and unit tests can then pass an explicit configuration file for testing.

portante: Another option for the server side is to just use the `server/lib/pbench/__init__.py:PbenchConfig` object itself.

Is there a reason why that wouldn't work?

agent/utils/config.py
```diff
@@ -0,0 +1,5 @@
+
+import configtools
+
+opts, _ = configtools.parse_args()
```
portante: Same comments apply here as made in `server/utils/config.py`.

server/utils/pbench_server.py
```diff
@@ -0,0 +1,136 @@
...
+
+                f.write(chunk)
+
+        resp = jsonify(dict(message='File successfully uploaded'))
+        resp.status_code = 201
+        return resp
+
+
+api.add_resource(Upload, f"{app.config['REST_URI']}/upload")
+api.add_resource(Check, f"{app.config['REST_URI']}/check")
```
portante: Can we merge the `Check` method into the `Upload` method by having the `Upload` method require a header that contains a MD5 value that can be used to verify the uploaded data?

grafuls:
> Some general comments on the server side (that's all I've looked at so far):
> 
>     * I think we should put all the python modules under server/lib. That way adding that to PYTHONPATH gets us everything.
DONE
> 
>     * the systemd directory should also go under lib.
DONE
> 
>     * you should not add a top-dir definition in the default config file: that is to be done only by the private config file corresponding to the server.
DONE
> 
>     * rest_log uses an absolute path - if you need it, it should probably be defined in terms of the install-dir which also goes in the private config file (unit tests use a different setting than actual server installations).
Moved to use $(pbench-logs-dir)
> 
>     * there is a logger defined in server/lib/pbench - can you use that instead of defining your own?
DONE
> 
>     * there is a PbenchConfig class ins server/lib/pbench that can also be used instead of rolling your own.
DONE
> 
> 
> There are examples of the setup for these last two in the server/bin/*.py files.

agent/lib/pbench/tests/fixtures/config/pbench-agent.cfg
```diff
@@ -0,0 +1,57 @@
+[DEFAULT]
+pbench_install_dir = /opt/pbench-agent
+pbench_results_redirector = pbench.results.example.com
+pbench_web_server = pbench.example.com
+
+version = 002
+
+[config]
+path = %(pbench_install_dir)s/config
+files = pbench-agent-default.cfg
```
ndokos: Just include lines 1-10 in your specific config file but change the pbench_install_dir somehow to reflect the place where the tests are run.

ndokos: So the pbench_install_dir should be changed to a path *relative to wherever the pbench-agent.cfg is*.
E.g. suppose I'm sitting at  the top level (where .travis.yml is), the pbench_agent_default.cfg file is `./agent/config/pbench_agent_default.cfg` relative to me and the pbench agent config file is `./agent/lib/pbench/tests/fixtures/config/pbench_agent.cfg` relative to me. Then you should set

     pbench_install_dir = ../../../../..

i.e. go back up to the ./agent directory five levels up before going down to the `%(pbench_install_dir)s/config` subdir to find the default config file.

You should be able to test this with:

     ./agent/util-scritps/getconf.py -L

which should give a list of the two relevant config files with full pathnames.

.travis.yml
```diff
@@ -29,6 +29,33 @@ jobs:
...
+      python:
+        - "3.6"
+      virtualenv:
+        system_site_packages: true
+      env:
+        - CONFIG=./agent/lib/pbench/tests/fixtures/config/pbench-agent.cfg
+        - PYTHONPATH=$PYTHONPATH:$TRAVIS_BUILD_DIR/agent/lib
+      install:
+        - pip install requests responses pytest configtools werkzeug
+        - sudo apt-get install python3-rpm
```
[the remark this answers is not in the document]
grafuls: travis machine uses debian and this specific package is not available via pypy

agent/lib/pbench/make_result_tb.py
```diff
@@ -0,0 +1,111 @@
...
+import tarfile
+import datetime
+from pathlib import Path
+
+import configtools
+from agent.lib.pbench import config_agent
+
+from agent.lib.pbench.add_metalog_option import add_metalog_option
+
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
grafuls: moved it dunder init instead.

server/utils/pbench_server.py
grafuls: I'd prefer to keep this 2 separated, the hash check is a lengthy process as well as the upload, would be more prone to timeouts, plus I prefer the granularity of knowing what part of the upload went wrong.

.travis.yml
```diff
@@ -27,6 +27,33 @@ jobs:
         - pip install elasticsearch sh boto3
       script: ./server/bin/unittests
 
+    - stage: "Agent Pytests"
```
portante, line 30: Be sure to review https://github.com/distributed-system-analysis/pbench/pull/1583 which is going to either make this easier for you to change, or harder because of merge conflicts. =)

agent/lib/pbench/__init__.py
```diff
@@ -0,0 +1,15 @@
+import logging
+import sys
+
+import configtools
+
+opts, _ = configtools.parse_args()
```
portante, line 6: Can we move this to module function like we discussed before?  Having this setup when the module is imported makes it difficult to unit test effectively, and is not needed right when the module is loaded.

agent/lib/pbench/add_metalog_option.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python3
+
+# Usage: pbench-add-metalog-option <metadata log file> <section> <option>
```
portante, line 3: Does this module "become" the `pbench-add-metalog-option` CLI interface?

server/lib/pbench_server/pbench_server.py
```diff
@@ -0,0 +1,130 @@
...
+        with open(full_path, "wb") as f:
+            chunk_size = 4096
+            logger.debug("Writing chunks")
+
+            while True:
+                chunk = request.stream.read(chunk_size)
+                if len(chunk) == 0:
+                    break
+
+                f.write(chunk)
```
portante, line 118: Why can't we calculate the MD5 check sum of the file as we are reading it off the wire, like https://github.com/distributed-system-analysis/pbench/blob/f32e5b4eeae9f99789ecc824b7bfca6aed509535/agent/util-scripts/pbench-tool-data-sink#L65?

Then we could get rid of the `/check` end point.

.travis.yml
grafuls: That patch is using tox to run the already existing unit test scripts but I am using pytest framework instead. I can modify this definitions to use tox to run my pytests.

agent/lib/pbench/add_metalog_option.py
grafuls: That's the intention yes. This can be renamed to that or we can have the command aliased or create a dynamic link.

agent/lib/pbench/copy_result_tb.py
```diff
@@ -0,0 +1,88 @@
...
+            sys.exit(1)
+
+    def check_sum(self):
+        headers = {
+            "filename": secure_filename(os.path.join(self.controller, self.tarball))
+        }
+        try:
+            response = requests.get(self.check_url, headers=headers)
+            self.logger.info("File uploaded successfully")
+        except Exception:
```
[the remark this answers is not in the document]
grafuls: the logger.exception method prints out the stack trace as well.

agent/lib/pbench/make_result_tb.py
```diff
@@ -0,0 +1,126 @@
...
+            self.logger.debug(
+                "The %s/metadata.log file seems to be missing", pbench_run_name
+            )
+            sys.exit(0)
+
+        opts, _ = configtools.parse_args()
+        opts.filename = md_log
+        conf_md, _ = configtools.init(opts, None)
+        md_config = conf_md["pbench"]
+        res_name = md_config.get("name")
```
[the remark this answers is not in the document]
grafuls: this for the metadata log not included on the config.

agent/lib/pbench/add_metalog_option.py
```diff
@@ -0,0 +1,34 @@
...
+# iterations: 1-iter, 2-iter, 3-iter
+#
+# where the iterations are in the <iterations.file>, one iteration per line.
+import os
+import sys
+from configparser import ConfigParser, NoSectionError
+
+
+def add_metalog_option(log_file, section, option, value):
+    if not os.path.exists(log_file):
```
portante, line 17: Typically, the "pythonic" way to approach this is something like:

```
try:
    config.read(log_file)
except OSError as exc:
    if exc.errno == errno.ENOENT:
        raise Exception(f"Log file does not exist: {log_file}")
    raise
```

This lets us avoid extra tests and code to support those tests.
```

PR #1501, opened 2020-02-13
User authentication

aquibbaig:
Initial PR for User Authentication feature on the dashboard

PR #1506, opened 2020-02-19
User authentication design

aquibbaig:
#1501 was an early implementation for design logic but it was nowhere near to a proper design flow, hence closing that and reopening with this PR for User authentication. This design is in direction to the one suggested by Mary and @gurbirkalsi


aquibbaig:
Reopening in the dashboard repo.

PR #1507, opened 2020-02-19
WIP: Convert agent scripts to use python3

zulcss:


pbench/cmd/cli.py
```diff
@@ -0,0 +1,39 @@
...
+from pbr import version
+
+
+class Main(app.App):
+
+    log = logging.getLogger(__name__)
+
+    def __init__(self):
+        super(Main, self).__init__(
+            description="A Benchmarking and Performance Analysis Framework",
```
portante: Is there anyway we can pull this from `setup.py`?

Issue #1498, opened 2020-02-11
Make python code pep8/flake8 compliant


zulcss:
Lots of non-flake8 python code especially on the server backend. It would be nice if its compliant and gating in the CI.

portante:
@zulcss, can you check to see if we have completed this work in the `br69` branch?  I am guessing we have.

portante:
Now that we merged the `br69` branch onto `master` this work is done.

PR #1511, opened 2020-02-20
Fix pep8 warnings

zulcss:
Fix pep8 linting errors and warnings, how this was fixed was the following:

- Installed pycodestyle locally
- Run it against the tree, fix the errors

I used autopep8 to automatically fix most of the pep8 errors for me. I verified the fixes before submitting this PR.

This PR fixes Issue #1498


grafuls:
We could consider having 2 pre-commit hooks to do the formatting and the checking with [black](https://github.com/psf/black) as a linter and flake8 as the checker.

ndokos:
IMO, all the white space changes should be done in one single commit (and probably in another PR): mixing white space fixes with other fixes is always confusing, so if we could do all the non-white space stuff (which I assume is more serious) first and then do white space on its own.

And before we do the WS stuff, we'll all have to install the proper pre-commit hooks, right? Or can that be done centrally? If not, we'll need to educate ourselves (that may be a real or a royal we, but I certainly need the education) and we'll need to educate everybody else about them.

PR #1513, opened 2020-02-21
sleep

portante:
A simple experiment to see where Ubuntu's TravisCI environment provides the `sleep` command.


portante:
The Ubuntu TravisCI environment provides `/bin/sleep`.

PR #1514, opened 2020-02-21
Don't emit messages directly to the tty

portante:
The `pbench-fio` benchmark script was emitting a message directly to the `tty` when the `fio` version could not be determined.  Instead we add it to the `pbench.log` and emit it on `stderr`.

We also define the `PYTHONPATH` once in the server unit tests, and make the python markers in the `getconf.py` modules uniform with other files.


portante:
> You left the `/common/lib` part in PYTHONPATH for future use? We don't have that any longer (but maybe @zulcss is adding it in the other PRs?)

It was part of the original code, so I did not change that.  A `grep -RF "common/lib"` shows that references to that path a riddled through the server code, which probably means this change is another cleanup item.

PR #1472, opened 2020-01-27
rename CONFIG to their respective(agent or server side) environment var 

riya-17:
Fixes issue #139 

This PR deals with renaming the CONFIG environment variable to their
respective environment(i.e. server, agent) variable like _PBENCH_SERVER_CONFIG
for the server side and _PBENCH_AGENT_CONFIG for the agent side.

PR #1515, opened 2020-02-25
Refactoring of server ansible playbooks and fixes from the configtools reorg

ndokos:
There are two commits in this PR:

- The first commit refactors the playbooks to move site-specific variables out of the playbooks and into a group_vars file, accessible from the inventory file. There is now a configfiles variable that is a list of config files to install (in addition to the default file that is part of the distribution): we make use of that for the S3 backup information. We also allow multiple repos now: it makes life easier for developers.

-  The second commit deals with the fallout from the configtools reorg (PR #1472 and PR #1457). We need to define PATH, PYTHONPATH and _PBENCH_SERVER_CONFIG in the environment in order to be able to execute some commands (getconf.py in particular). We also need to use scl to execute these commands on RHEL7 systems.

### DIFF
--- a/server/ansible/roles/pbench-repo-install/templates/etc/yum.repos.d/pbench-server.repo.j2
+++ b/server/ansible/roles/pbench-repo-install/templates/etc/yum.repos.d/pbench-server.repo.j2
@@ -1,11 +1,13 @@
-###########################################################################
-# pbench "official" repo
-[copr-pbench-server]
-name=Copr repo for pbench-server
-baseurl={{ baseurl }}
+{% for repo in repos %}
+
+[copr-{{ repo.tag }}]
+name=Copr {{ repo.tag }} repo for pbench-server
+baseurl={{ repo.baseurl }}
 skip_if_unavailable=True
-gpgcheck= {{ gpgcheck }}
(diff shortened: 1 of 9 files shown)


server/ansible/roles/pbench-repo-install/templates/etc/yum.repos.d/pbench-server.repo.j2
```diff
@@ -1,11 +1,16 @@
...
-gpgkey={{ gpgkey }}
-enabled=1
+gpgcheck= {{ repo.gpgcheck }}
+gpgkey={{ repo.gpgkey }}
+enabled={{ repo.enabled }}
 enabled_metadata=1
 skip_if_unavailable=1
+
+{% endfor %}
+
```
portante: Just some trailing whitespace.

portante:
Added @zulcss and @grafuls to review for awareness; would be great if one other person also reviewed.

ndokos:
This needs to be squashed and the description needs to be rewritten, but I think it's ready to be reviewed.

ndokos:
I squashed it into two commits and rewrote the commit messages and the PR description. PTAL.

PR #1516, opened 2020-02-25
Add function to index reason of failed tarballs

riya-17:
This PR deals with generating a report to save the reason of a tarball failure.

### DIFF
--- a/server/bin/pbench-index.py
+++ b/server/bin/pbench-index.py
@@ -299,26 +299,29 @@ def main(options, name):
                 except UnsupportedTarballFormat as e:
                     idxctx.logger.warning("Unsupported tar ball format: {}", e)
                     tb_res = 4
+                    indexing_failure(os.path.realpath(tb),"UnsupportedTarballFormat", tb_res)
                 except BadDate as e:
                     idxctx.logger.warning("Bad Date: {!r}", e)
                     tb_res = 5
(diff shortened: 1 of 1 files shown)


server/bin/pbench-index.py
```diff
@@ -454,6 +457,9 @@ def main(options, name):
 
     return res
 
+def indexing_failure(tb, error, tb_res):
```
ndokos: This is going to be fleshed out similarly to `pbench-report-status`. We'll need to use a different index and a different document type (TBD), so we'll need to tweak it in order to be able to reuse it here.

server/bin/pbench-index.py
```diff
@@ -299,26 +299,29 @@ def main(options, name):
                 except UnsupportedTarballFormat as e:
                     idxctx.logger.warning("Unsupported tar ball format: {}", e)
                     tb_res = 4
+                    indexing_failure(os.path.realpath(tb),"UnsupportedTarballFormat", tb_res)
```
ndokos: There should probably be a single call to `indexing_failure()` in a `finally` section: all we need to do is save `e` to a more permanent place (`saved_e` perhaps - initialized to `None`). Then in the `finally` section, we check `saved_e` and if not `None`, we call `indexing_failure()`.

portante:
Can you rebase this on the `br69` branch?

Issue #1512, opened 2020-02-20
Modify the indexer of the pbench server to index pbench-run-benchmark data


portante:
The current pbench indexer does not consider `pbench-run-benchmark` runs.  We need to fix that which should be fairly straightforward to do.

PR #1517, opened 2020-02-26
Correctly invoke pbench-metadata-log 'beg'

portante:
Both invocations of `pbench-metadata-log` need the name of the benchmark.  Without the `$benchmark` environment variable being set, `pbench-metadata-log` fails to create the required metadata in the top-level `metadata.log` file.  Without this metadata, server-side indexing will fail.

See also #1512.

### DIFF
--- a/agent/lib/PbenchBase.pm
+++ b/agent/lib/PbenchBase.pm
@@ -171,8 +171,9 @@ sub load_benchmark_config {
 
 sub metadata_log_begin_run {
     my $benchmark_run_dir = shift;
+    my $benchmark_name = shift;
     my $group = shift;
-    system("pbench-metadata-log --group=" . $group . " --dir=" . $benchmark_run_dir . " beg");
+    system("benchmark=" . $benchmark_name . " pbench-metadata-log --group=" . $group . " --dir=" . $benchmark_run_dir . " beg");
 }
 
 sub metadata_log_end_run {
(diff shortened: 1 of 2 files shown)


agent/lib/PbenchBase.pm
```diff
@@ -171,8 +171,9 @@ sub load_benchmark_config {
 
 sub metadata_log_begin_run {
     my $benchmark_run_dir = shift;
+    my $benchmark_name = shift;
     my $group = shift;
-    system("pbench-metadata-log --group=" . $group . " --dir=" . $benchmark_run_dir . " beg");
+    system("benchmark=" . $benchmark_name . " pbench-metadata-log --group=" . $group . " --dir=" . $benchmark_run_dir . " end");
```
k-rister: Are we sure this is correct?  The subroutine is `metadata_log_begin_run` so using `beg` as the argument seems more logical than `end` to me, but I can't profess to fully understanding the metadata log usage/behavior.

portante: Yes, copy-paste-error on my part.  Thanks for catching that!

portante:
@k-rister, does this look right to you now?

k-rister:
I'm going to try and test this today; I should be able to apply this patch on top of an existing pbench install and do some pbench-run-benchmark stuff, upload the results, and see it indexed in the dashboard, correct (in theory anyway)?

ndokos:
I don;t think it will be indexed: the indexer won't know how to index it yet.  But the change will make it so that the tarball is not immediately rejected because of "Bad metadata log file". And it will provide a test case for the work that I think @portante is planning to do on that in the next couple of days.

PR #1521, opened 2020-02-28
Enhance indexing to handle pbench-run-benchmark

portante:
Partial fix for #1512.

The current way the indexer recognizes what to index or not is via the `script` option of the `[run]` section in a `metadata.log` in a pbench result.  We currently only accept `fio`, `uperf`, and `trafficgen`, the names that `pbench-(fio|uperf|trafficgen)` emit today as the `$benchmark` environment variable value.

However, when `pbench-run-benchmark` executes, the value it uses for the `$benchmark` enviroment variable is the existing `pbench-*` counterparts.

We add a test for a `fio` benchmark run via `pbench-run-benchmark` to verify that.

This particular data set also uses the "user" feature, which identified that we were missing a mapping for that field in `result-data-sample`, and `tool-data-*` mappings.

This commit also provides:

 * the proper handling and mappings for the `raw_size` and `tar-ball-creation-timestamp` fields in the `@metadata` field of a `run` document
 * the ability to index pbench tar balls which don't have any tool data
 * the ability to index linpack workloads
 * a fix for bad assertions encountered in certain cases
 * adds a new `parent` field to address issue #1524


server/lib/pbench/indexer.py
```diff
@@ -1322,33 +1355,83 @@ def gen_sources(obj, results, iteration, cvt_ts):
...
+                                sample_md['end'] = end_ts
+                        else:
+                            # FIXME: Hack
+                            start_ts = obj.ptb.start_run
+                            start = end = None
                         # Now we can emit the sample document knowing the
                         # ID of its iteration parent document.
                         source = _dict_const([
+                            # We endow the source document for the sample with
+                            # the timestamp of the first elment of all the
```
ndokos: element

server/lib/pbench/indexer.py
```diff
@@ -1254,6 +1300,19 @@ def make_sample_wrapper(result_type, title, result_el_idx, result_el):
...
         result_el['uid'] = ResultData.expand_template(result_el['uid_tmpl'], result_el)
+        try:
+            mean_val = float(result_el['mean'])
+        except KeyError:
+            # We don't have a mean field, so just ignore.
+            pass
+        except ValueError:
+            # The mean field is not a proper float value, so move it to the
+            # side to avoid indexing errors.
+            result_el['mean_raw'] = result_el['mean']
```
ndokos: Should this be in the mapping?

portante: Yes, good catch.

server/lib/pbench/indexer.py
portante: Fixed.

Issue #1523, opened 2020-03-02
Rewrite pbench-server-prep-shim-002 as python3 program


riya-17:
We need to rewrite pbench-server-prep-shim-002 in Python3

portante:
Resolved by #1525.

PR #1525, opened 2020-03-03
pbench-server-prep-shim-002 into python3 program

riya-17:
Fixes issue #1523 

Rewriting Pbench-server-prep-shim-002 into python program


server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,225 @@
...
+    if not os.path.isdir(qdir):
+        logger.error("Failed: {} does not exist, or is not a directory".format(qdir))
+        return 2
+
+    receive_dir_prefix = config.get('pbench-server','pbench-receive-dir-prefix')
+    if len(receive_dir_prefix)==0:
+        logger.error("Failed: getconf.py pbench-receive-dir-prefix pbench-server")
+        return 2
+
+    receive_dir= receive_dir_prefix+"-002"
```
grafuls: It is a good practice not to use `+` for sting concatenation, use f-string instead:
f"{receive_dir_prefix}-002"

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,225 @@
...
+        if tar.endswith('.tar.xz.md5'):
+            list_check.append(tar)
+
+    list_check.sort()
+    logger.info(config.TS)
+
+    ntotal = ntbs = nerrs = nquarantined = ndups = 0
+
+    for tbmd5 in list_check:
+        ntotal = ntotal+1
```
grafuls: ntotal += 1

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,225 @@
...
+        tb = tbmd5[0:-4]
+        tbdir = os.path.dirname(tb)
+        result_name = os.path.basename(tb)
+        resultname = result_name[0:-7]
+
+        controller = os.path.basename(tbdir)
+        dest = os.path.join(config.ARCHIVE, controller)
+
+        if os.path.isfile(os.path.join(dest, result_name)) and os.path.isfile(os.path.join(dest, os.path.basename(tbmd5))) and os.path.exists(os.path.join(dest, result_name)) and os.path.exists(os.path.join(dest, os.path.basename(tbmd5))):
```
grafuls: You'll have a flake8 issue here for having a line longer than 119 chars.

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,225 @@
...
+            logger.error("{}: Duplicate: {} duplicate name".format(config.TS, tb))
+            quarantine(os.path.join(duplicates,controller), tb, tbmd5)
+            ndups = ndups+1
```
grafuls: ndups += 1

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,225 @@
...
+            continue
+
+        # read the md5sum from md5 file
+        try:
+            with open(tbmd5) as f:
+                archive_md5_hex_value = f.readline().split(" ")[0]
+        except Exception:
+            logger.error("Quarantine: Could not read {}".format(tb))
+            quarantine(os.path.join(qdir_md5,controller), tb, tbmd5)
+            nquarantined = nquarantined+1
```
grafuls: nquarantined += 1

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,225 @@
...
+            continue
+
+        # match md5sum of the tarball to its md5 file
+        try:
+            archive_tar_hex_value = md5sum(tb)
+        except Exception:
+            logger.error("Quarantine: Could not read {}".format(tb))
+            quarantine(os.path.join(qdir_md5,controller), tb, tbmd5)
+            nquarantined = nquarantined+1
```
grafuls: nquarantined += 1

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,225 @@
...
+            continue
+
+        if archive_tar_hex_value != archive_md5_hex_value:
+            logger.error("{}: Quarantined: {} failed MD5 check".format(config.TS, tb))
+            quarantine(os.path.join(qdir_md5,controller), tb, tbmd5)
+            nquarantined = nquarantined+1
```
grafuls: nquarantined += 1

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,225 @@
...
+        if not os.path.isdir(os.path.join(dest,'TODO')):
+            try:
+                os.makedirs(os.path.join(dest,'TODO'))
+            except FileExistsError:
+                # directory already exists, ignore
+                pass
+            except Exception:
+                logger.error("{}: Error in creating TODO directory. {}".format(config.TS, e))
+                quarantine(os.path.join(errors,controller), tb, tbmd5)
+                nerrs = nerrs+1
```
grafuls: nerrs += 1

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,225 @@
...
+                continue
+
+        try:
+            shutil.copy(tb, dest)
+            shutil.copy(tbmd5, dest)
+        except Exception:
+            logger.error("{}: Error in copying tarball files to Destination path. {}".format(config.TS, e))
```
grafuls: you can use logger.exception which will print out the `e` variable without you needing to reference it.

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,225 @@
...
+            try:
+                shutil.rmtree(os.path.join(dest, result_name))
+                shutil.rmtree(os.path.join(dest, os.path.basename(tbmd5)))
+            except Exception as e:
+                logger.error("{}: Warning: cleanup of copy failure failed itself. {}".format(config.TS, e))
```
grafuls: same as above

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,225 @@
...
+            quarantine(os.path.join(errors,controller), tb, tbmd5)
+            nerrs = nerrs+1
```
grafuls: +=

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,225 @@
...
+            continue
+        
+        try:
+            os.remove(tb)
+            os.remove(tbmd5)
+        except Exception as e:
+            logger.error("{}: Warning: cleanup of successful copy operation failed. {}".format(config.TS, e))
```
grafuls: same as above

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,225 @@
...
+        
+        try:
+            os.symlink(os.path.join(dest,result_name), os.path.join(dest,'TODO', result_name))
+        except Exception as e:
+            logger.error("{}: Error in creation of symlink. {}".format(config.TS, e))
```
grafuls: same as above

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,225 @@
...
+            quarantine(os.path.join(errors,controller), os.path.exists(os.path.join(dest,tb)), os.path.exists(os.path.join(dest,tbmd5)))
+            nerrs = nerrs+1
```
grafuls: +=

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,225 @@
...
+            continue
+        
+        ntbs = ntbs+1
```
grafuls: +=

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,225 @@
...
+    with tempfile.NamedTemporaryFile(mode='w+t', dir=config.TMP) as reportfp:
+        reportfp.write("{}\n".format(result_string))
+        reportfp.seek(0)
+
+        report = Report(config, _NAME_)
+        report.init_report_template()
+        try:
+            report.post_status(config.TS, "status", reportfp.name)
+        except Exception:
+            pass
```
grafuls: How do you handle non successful posts?

server/bin/pbench-server-prep-shim-002.py
riya-17: ok

server/bin/pbench-server-prep-shim-002.py
riya-17: ok

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,228 @@
...
+        print("{}: {} (config file {})".format(_NAME_, e, cfg_name), file=sys.stderr)
+        return 1
+
+    if not os.path.exists(config.ARCHIVE):
+        print("{}: Bad ARCHIVE={}".format(_NAME_, config.ARCHIVE), file=sys.stderr)
+        return 1
+
+    logger = get_pbench_logger(_NAME_, config)
+
+    qdir = config.QDIR = config.get('pbench-server', 'pbench-quarantine-dir')
```
zulcss, line 42: Can you remove config.QDIR, its not being used anywhere.

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,246 @@
...
+        except Exception:
+            logger.exception(
+                "os.mkdir: Unable to create errors destination directory: {}", errors
+            )
+            return 1
+
+    list_check = list()
+    receive_dir_check = glob.iglob(os.path.join(receive_dir, "**"), recursive=True)
+    for tar in receive_dir_check:
+        if tar.endswith(".tar.xz.md5"):
```
portante, line 106: Can we put this pattern right into the `iglob()` call?

Something like:

```
list_check = glob.glob(os.path.join(receive_dir, "**", "*.tar.xz.md5", recursive=True)
```

Which would replace lines 103 thru 107.

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,246 @@
...
+            list_check.append(tar)
+
+    list_check.sort()
+    # logger.info(config.TS)
```
portante, line 110: Can you remove? Not worth keeping as a comment.

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,246 @@
...
+        print(f"{_NAME_}: ERROR: No config file specified; set CONFIG")
+        return 2
+
+    try:
+        config = PbenchConfig(cfg_name)
+    except BadConfig as e:
+        print(f"{_NAME_}: {e} (config file {cfg_name})", file=sys.stderr)
+        return 1
+
+    if not os.path.exists(config.ARCHIVE):
```
portante, line 32: Can we use Pathlib for `ARCHIVE`, "receive-dir", etc.?

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,246 @@
...
+        result_name = os.path.basename(tb)
+
+        controller = os.path.basename(tbdir)
+        dest = os.path.join(config.ARCHIVE, controller)
+
+        if (
+            os.path.isfile(os.path.join(dest, result_name))
+            and os.path.isfile(os.path.join(dest, os.path.basename(tbmd5)))
+            and os.path.exists(os.path.join(dest, result_name))
+            and os.path.exists(os.path.join(dest, os.path.basename(tbmd5)))
```
portante, line 127: Do we need to both check with `isfile()` and with `exists`?  If we use Pathlib, we might be able to leverage that class's methods to make this a bit cleaner.

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,246 @@
...
+
+        try:
+            shutil.copy(tb, dest)
+            shutil.copy(tbmd5, dest)
+        except Exception:
+            logger.error(
+                "{}: Error in copying tarball files to Destination path.", config.TS
+            )
+            try:
+                shutil.rmtree(os.path.join(dest, result_name))
```
portante, line 179: Not sure you need `shutil.rmtree` since this should be one file, right?

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,292 @@
...
+            logger.error(
+                "{}: Warning: cleanup of successful copy operation failed.", config.TS
+            )
+
+        try:
+            os.symlink((dest / result_name), (dest / "TODO" / result_name))
+        except Exception:
+            logger.error("{}: Error in creation of symlink.", config.TS)
+            quarantine(
+                (errors / controller), (dest / tb).exists(), (dest, tbmd5).exists(),
```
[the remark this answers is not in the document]
riya-17: No it doesn't seems like a glitch

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,331 @@
...
+        # resultname: get the basename foo.tar.xz and then strip the .tar.xz
+        resultname = tb.name
+
+        controller = tbdir.name
+        dest = archive / controller
+
+        if (
+            (dest / resultname).is_file()
+            and (dest / tbmd5.name).is_file()
+            and (dest / resultname).exists()
```
portante, line 132: Are not this two `.exists` checks redundant?

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,331 @@
...
+        try:
+            shutil.copy2(tbmd5, dest)
+        except Exception:
+            logger.error(
+                "{}: Error in copying .md5 file to Destination path.", config.TS
+            )
+            try:
+                os.remove(dest / tbmd5.name)
+            except Exception:
+                logger.error(
```
portante, line 172: If the copy failed, then it is possible the destination file does not exist.  So this error should only be logged if the exception is not "file not found".

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,331 @@
...
+        # take a bit longer.  If it fails, the file will NOT be at the
+        # destination.
+        try:
+            shutil.move(str(tb), str(dest))
+        except Exception:
+            logger.error(
+                "{}: Error in moving tarball file to Destination path.", config.TS
+            )
+            try:
+                os.remove(dest / resultname)
```
portante, line 191: Same comment as above, for failure conditions, the destination tar ball may not exist, so we should only log this exception if it is not "file not found".

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,292 @@
...
+                    "{}: Warning: cleanup of copy failure failed itself.", config.TS
+                )
+            quarantine((errors / controller), tb, tbmd5)
+            nerrs += 1
+            continue
+
+        try:
+            os.remove(tb)
+            os.remove(tbmd5)
+        except Exception:
```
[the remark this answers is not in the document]
portante: I agree, this seems to be one of the downsides of using a file system.

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,331 @@
...
+        return 2
+
+    try:
+        config = PbenchConfig(cfg_name)
+    except BadConfig as e:
+        print(f"{_NAME_}: {e} (config file {cfg_name})", file=sys.stderr)
+        return 1
+
+    archive = Path(config.ARCHIVE)
+    archive_p = archive.resolve()
```
portante, line 253: See comments from #1667 on using `Path` objects to resolve to a proper directory.

server/bin/pbench-server-prep-shim-002.py
riya-17: No, one of them is for tarball and another one is for .md5

portante: So the redundancy is not about the two `.exists()` checks for each file, but for having an `.is_file()` and a `.exists()` check for the same file.  How can you have `.is_file()` return `True` and `.exists()` return `False`?  A file can't be a file if it doesn't exist, right?

Seems like just checking that both are files also checks for existence.

server/bin/pbench-server-prep-shim-002.py
riya-17: ok

server/bin/pbench-server-prep-shim-002.py
riya-17: Seems fair

server/bin/pbench-server-prep-shim-002.py
riya-17: ok

server/bin/pbench-server-prep-shim-002.py
riya-17: Ohh.. Ok

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,327 @@
...
+            nerrs += 1
+            continue
+
+        # Now that we have successfully moved the tar ball and its .md5 to the
+        # destination, we can remove the original .md5 file.
+        try:
+            os.remove(tbmd5)
+        except Exception:
+            logger.error(
+                "{}: Warning: cleanup of successful copy operation failed.", config.TS
```
portante, line 203: Can you add the exception string to the logged error message, something like:

```
try:
    os.remove(tbmd5)
except Exception as exc:
    logger.error(
        "{}: Warning: cleanup ... failed: '{}'", config.TS, exc
    )
```

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,327 @@
...
+            )
+
+        try:
+            os.symlink((dest / resultname), (dest / "TODO" / resultname))
+        except Exception:
+            logger.error("{}: Error in creation of symlink.", config.TS)
```
portante, line 209: Can you also add the error message here, like above?

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,327 @@
...
+        )
+        return 2
+
+    try:
+        config = PbenchServerConfig(cfg_name)
+    except BadConfig as e:
+        print(f"{_NAME_}: {e} (config file {cfg_name})", file=sys.stderr)
+        return 1
+
+    archive = Path(config.ARCHIVE)
```
portante, line 249: Seems like `config.ARCHIVE` is already a `Path()` object (see https://github.com/distributed-system-analysis/pbench/blob/br69/lib/pbench/server/__init__.py#L59).  we can just assign it to `archive` for convenience.

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,327 @@
...
+    except FileNotFoundError:
+        print(f"{_NAME_}: Bad ARCHIVE={config.ARCHIVE}", file=sys.stderr)
+        return 1
+    else:
+        if not archivepath.is_dir():
+            print(f"{_NAME_}: Bad ARCHIVE={config.ARCHIVE}", file=sys.stderr)
+            return 1
+
+    try:
+        os.mkdir(os.path.join(config.LOGSDIR, "pbench-server-prep-shim-002"))
```
portante, line 261: It doesn't seem like we need to do this, see https://github.com/distributed-system-analysis/pbench/blob/br69/lib/pbench/server/__init__.py#L49, as it is already checked to see if it exists.

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,327 @@
...
+    except FileExistsError:
+        # directory already exists, ignore
+        pass
+    except Exception:
+        print("os.mkdir: Unable to create destination directory")
+        return 1
+
+    errorfile = Path(
```
portante, line 269: Can you use the `open()` method of the `Path` object in a with statement to ensure it gets closed at each return statement?

```
with Path(config.LOGSDIR, "psps-002", "psps-002.error").open() as error_f:
    qdir, receive_dir = fetch_config_val(config, error_f)
    if qdir is None and receive_dir is None:
        return 2
    ...
    counts = process_tb(
        config, logger, receive_dir, qdir_md5, duplicates, errors, error_f
    )
```

server/bin/pbench-server-prep-shim-002.py
riya-17: It doesn't seem to get created unless we call 'get_pbench_logger'
And if I call 'get_pbench_logger'  first then in some cases it is creating pbench-server-prep-shim-002.log file which doesn't need to be created at all.

riya-17: yaa

riya-17: it seems like it gets created with 'get_pbench_logger'
If I am calling 'get_pbench-logger' before the creation of errorfile than it creates 'pbench-servre-prep-shim-002.log' which doesn't seems to get created at all in test-13 and test-15.

portante:
@zulcss, you have requested changes, can you re-review to make sure you are good with this PR?

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,318 @@
...
+    with tempfile.NamedTemporaryFile(mode="w+t", dir=config.TMP) as reportfp:
+        reportfp.write(f"{counts.nstatus}{result_string}\n")
+        reportfp.seek(0)
+
+        report = Report(config, _NAME_)
+        report.init_report_template()
+        try:
+            report.post_status(config.timestamp(), "status", reportfp.name)
+        except Exception:
+            logger.warning("Report post Unsuccesful")
```
portante, line 310: One slight tweak here, can you capture the error encountered?

E.g.

```
try:
    report.post_status( ... )
except Exception as exc:
    logger.warning("Report post unsuccessful: '{}'", exc)
```

ndokos:
You'll need to rebase this on current master.

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,307 @@
...
+    return qdir
+
+
+def md5_check(tb, tbmd5, logger):
+    # read the md5sum from md5 file
+    try:
+        with tbmd5.open() as f:
+            archive_md5_hex_value = f.readline().split(" ")[0]
+    except Exception:
+        logger.exception("Quarantine: Could not read {}", tbmd5)
```
portante, line 88: If we get an exception, the `archive_md5_hex_value` variable will be undefined when we return at line 96.

Could we raise an exception that will be caught and handled at the time `md5_check` is called?

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,307 @@
...
+        controller = tbdir.name
+        dest = archive / controller
+
+        if all([(dest / resultname).is_file(), (dest / tbmd5.name).is_file()]):
+            errorfile.write(f"{config.TS}: Duplicate: {tb} duplicate name\n")
+            quarantine((duplicates / controller), logger, tb, tbmd5)
+            ndups += 1
+            continue
+
+        archive_tar_hex_value, archive_md5_hex_value = md5_check(tb, tbmd5, logger)
```
portante, line 137: Here we should consider catching an exception raised from `md5_check`.

server/bin/pbench-server-prep-shim-002.py
portante: An alternative to raising an exception is to keep logging it, but make sure we set `archive_md5_hex_value` to `None` so the value is defined when returned at line 96.

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -0,0 +1,308 @@
...
+            archive_md5_hex_value = f.readline().split(" ")[0]
+    except Exception:
+        archive_md5_hex_value = None
+        logger.exception("Quarantine: Could not read {}", tbmd5)
+
+    # get hex value of the tarball's md5sum
+    try:
+        archive_tar_hex_value = md5sum(tb)
+    except Exception:
+        logger.exception("Quarantine: Could not read {}", tb)
```
portante, line 95: Same thing applies here, `archive_tar_hex_value` needs a value in order to avoid an another exception at line 97 below.

server/bin/pbench-server-prep-shim-002.py
portante: Now that we are going to get `None` values from `md5_check()`, we also need to check if we are receiving `None` values and if so, take the error path at line 140.

portante:
Thanks @zulcss, we'll wait for @ndokos's review before merging.

portante:
@riya-17, one last rebase and then we'll merge.  No further reviews are needed.

PR #1529, opened 2020-03-06
WIP - python changes

zulcss:
This brranch contains a bunch of things at once:
- python3 infra 
- unittests
- Start of pbench-cli

This branch is used to tract work for all the python changes that are going on right now.

### DIFF
--- a/agent/bench-scripts/unittests
+++ b/agent/bench-scripts/unittests
@@ -1,4 +1,5 @@
 #!/bin/bash
+
 export LANG=C
 export LC_ALL=C
 export _tdir=$(dirname $(readlink -f ${0}))
@@ -20,6 +21,7 @@ export _testlog=${_testroot}/test-execution.log
 # This directory location is equivalent to the $pbench_run (or
 # /var/lib/pbench-agent).
 export _testdir=${_testroot}/pbench-agent
+export _testvenv=${_testroot}/venv
 
 # Copy bench-scripts execution environment to _testroot
 export _testopt=${_testroot}/opt/pbench-agent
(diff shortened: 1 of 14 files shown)


agent/bench-scripts/unittests
```diff
@@ -75,6 +91,8 @@ PATH=$(remove_path /opt/pbench-agent/util-scripts ${PATH})
 export PATH=${_testopt}/unittest-scripts:${_testopt}/util-scripts:${_testopt}/bench-scripts:${PATH}
 export PYTHONPATH=${_testopt}/common/lib:${PYTHONPATH}
 
+_install_venv
```
portante: Can we consider moving this out of the normal unit test script path and into a script designed for TravisCI so that we don't have to path the install price every time we run the unit tests?

It seems that we'll have virtual environments we'll be working in normally.  But not all of us currently work in virtual environments.

server/bin/unittests
```diff
@@ -17,13 +20,31 @@ export _tdir=$(dirname $(readlink -f $0))
...
+export _testvenv=${_testbase}/venv
+
 rm -rf ${_testbase}
 mkdir ${_testbase}
 if [[ ! -d $_testbase ]]; then
     echo "ERROR: failed to create test base directory, \"$_testbase\"" >&2
     exit 1
 fi
 
+function _install_venv {
```
portante: And it seems that the fact that this function is almost identical in every unit test script indicates we should factor it out to a single place (DRY principle) and only for the TravisCI environment.

A separate method for creating a `virtualenv` for ones working environment for develop could leverage the same procedure, or be kept in our `CONTRIBUTION` documentation.

zulcss:
As discussed with Peter more complete branch coming

PR #1535, opened 2020-03-10
Fix .pre-commit-config.yaml YAML signature

portante:


.pre-commit-config.yaml
```diff
@@ -1,4 +1,4 @@
-___
+---
 # black complains loudly that fio-histo-log-pctile.py is not a python file
 # because it isnt, so ignore it
```
ndokos: isn't

zulcss: +1 from other than the spelling typo

portante: Fixed.

Issue #1520, opened 2020-02-28
Tracking the life-cycle of tar balls in the pbench-server


portante:
Let's consider what metadata about tar balls that we have today.

A move-results provides
 * Name
 * Time the tar ball was pushed to the server
 * Size
 * Controller Directory Name
 * MD5 file with an MD5 value
 * Timestamp in the tar ball name

The first server side entity to touch a tar ball is the [pbench-server-prep-shim-002.sh](https://github.com/distributed-system-analysis/pbench/blob/v0.67/server/bin/pbench-server-prep-shim-002.sh).

What does it do?
 * Quarantines duplicates at [line 130](https://github.com/distributed-system-analysis/pbench/blob/v0.67/server/bin/pbench-server-prep-shim-002.sh#L130)
 * Quarantines bad MD5s at [line 141](https://github.com/distributed-system-analysis/pbench/blob/v0.67/server/bin/pbench-server-prep-shim-002.sh#L141)
 * Creates controller directory in `$ARCHIVE` at [line 147](https://github.com/distributed-system-analysis/pbench/blob/v0.67/server/bin/pbench-server-prep-shim-002.sh#L147)
 * Copies TB and MD5 files to controller directory at [line 156](https://github.com/distributed-system-analysis/pbench/blob/v0.67/server/bin/pbench-server-prep-shim-002.sh#L156)
   * FIXME: Should be using `cp -a` to preserve the timestamps
   * FIXME: Should also be validating the timestamp in the tar ball name

At this point in time, a JSON document can be constructed to be indexed into Elasticsearch to track this tar ball.  The index name might be, `<prefix>.tar-ball-state`, no date in index name, as records will be modifiable.

A JSON might look like the following, using an Elasticsearch document type of "existence":

```
{
    "name": "pbench-user-benchmark__2019.08.24T22.50.36.tar.xz",
    "size": 382000000,
    "creation_ts": "2019.08.24T23.31.42.839105",
    "md5": "50a083010...0024012",
    "controller": "EC2::ats-sm-swx9r-workload-a-7ncrq",
    "tarball_ts": "2019.08.24T22.50.36.000000"
}
```

Then we'd have an operations index, `<prefix>.tar-ball-operations.YYYY-MM`, containing immutable JSON documents recording what happened to a tar ball.

```
{
    "name": "pbench-user-benchmark__2019.08.24T22.50.36.tar.xz",
    "controller": "",
    "operator": "pbench-disptach",
[...]

PR #1542, opened 2020-03-16
Tracking tar balls status in the pbench-server 

riya-17:
Fixes issue #1520

This PR deals with tracking the life of the tarballs whether they are PASSED or FAILED and the stages they have been through or failed on like DISPATCH, INDEX,etc.


riya-17:
This PR partially covers the intended workflow. Opening another PR as it has got a lot of issues regarding implementation and idea.

PR #1543, opened 2020-03-16
Index pbench-user-benchmark data

portante:
We start by registering one known user benchmark, "Hammerdb-tpcc", and build the framework around that.  This work introduces a refactoring of the indexing code to make python objects for iterations and samples to help reuse code a bit more.

We only handle iteration lists in the metadata.log file which are comma-space (`, `) separated, since some iteration names can contain commas.  When only using comma separated iteration names, we can't tell where an iteration ends and where one begins.  It then follows that iteration names cannot include spaces.

We increased the use of the `members` field, avoiding reprocessing the tar ball when looking for iterations, samples, and other files.

We use compiled regular expressions where possible.

We also fix the `@timestamp_original` to use a string mapping for consistency since not all timestamps are integers, some are date strings.


server/lib/pbench/indexer.py
```diff
@@ -893,17 +897,294 @@ def get_result_json_dirs(ptb):
...
+                                "Logic bomb! bad ubm driver sample_md_cols")
+                        return None
+                _dict_const(
+                    name=name, version=ver.strip(),
+                    description=blob.strip(),
+                    driver=ubm_driver)
+        return user_benchmark
+
+    def _make_user_benchmark_json(self):
+        """Fine all of the user-benchmark-result.csv files, and build up sample
```
ndokos: Find

server/lib/pbench/indexer.py
```diff
@@ -893,17 +897,294 @@ def get_result_json_dirs(ptb):
...
+                for sample_md_col in ubm_driver['sample_md_cols']:
+                    if sample_md_col not in cols:
+                        idxctx.logger.error(
+                                "Logic bomb! bad ubm driver sample_md_cols")
+                        return None
+                _dict_const(
+                    name=name, version=ver.strip(),
+                    description=blob.strip(),
+                    driver=ubm_driver)
+        return user_benchmark
```
npalaska: Few Suggestions: 
1. According to the docString, this function is used to determine if this is a known user benchmark so it can be more readable if it returns boolean.
2. Wouldn't this function always return `None` because user_benchmark is set to None (on line 979) but never gets reassigned?
3. I think there are a lot of if statements with similar logger messages and they all return None so it can be rewritten to be more concise. Maybe `set` operations can be useful something like
```
cols = set(ubm_driver['columns'].keys())
cols_to_check = set(list(ubm_driver['val_col']) + list(ubm_driver['start_ts_col']) + list(ubm_driver['end_ts_col']) + ubm_driver['benchmark_md_cols'] + ubm_driver['sample_md_cols'] )
set_diff = cols_to_check - cols  # since we want to check if all the items in cols_to_check present in cols
if len(set_diff) >= 1:
  idxctx.logger.error("Logic bomb! bad ubm driver: {}".format(' '.join(map(str, set_diff))))
  return None
```
The above code may not log the field names such as `val_col ` or `start_ts_col` etc. but the logged column names can be traced back to these fields.

server/lib/pbench/indexer.py
portante: Fixed.

server/lib/pbench/indexer.py
portante: Reworked this code.

portante:
@mnmehta, `pbench-user-benchmark` will be indexed by the server when this lands.

portante:
This is ready for review, but I added the "do not merge" label because we need to make sure a few other PRs land in the 0.68 and 0.69 releases before this one lands.  We want to have all the pbench-agent side fixes in before we take this code.

agent/bench-scripts/pbench-fio
```diff
@@ -797,7 +797,6 @@ function fio_run_benchmark() {
 						for sample in `seq 1 $nr_samples`; do
 							# each attempt at a test config requires multiple samples to get stddev
 							benchmark_results_dir="$iteration_dir/sample$sample"
-							benchmark_tools_dir="$benchmark_results_dir/tools-$tool_group"
 							if [ "$postprocess_only" = "n" ]; then
```
zulcss: I saw this in #1553 can this be dropped?

server/lib/pbench/indexer.py
```diff
@@ -1018,20 +1017,22 @@ def __init__(self, ptb):
...
         Fetch the list of directories containing result.json files for this
         experiment; return a list directory path names.
         """
         paths = [
-            x
-            for x in ptb.tb.getnames()
-            if (os.path.basename(x) == "result.json") and ptb.tb.getmember(x).isfile()
+            x.name
+            for x in ptb.members
+            if x.isfile() and (os.path.basename(x.name) == "result.json")
```
zulcss: missing colon?

server/lib/pbench/indexer.py
```diff
@@ -3317,8 +3681,99 @@ def hostnames_if_ip_from_sosreport(sos_file_name):
     return (0, d)
 
 
+class BadIterationName(Exception):
```
zulcss: Can this be grouped with the other classes that raise an exception, it starts around line 93.

server/lib/pbench/indexer.py
```diff
@@ -3317,8 +3681,99 @@ def hostnames_if_ip_from_sosreport(sos_file_name):
...
+                    # record the iteration number explicitly since the name is
+                    # the iteration number.
+                    self.number = 1
+                else:
+                    raise BadIterationName(self.name)
+        self.path = os.path.join(ptb.extracted_root, ptb.dirname, self.name)
+        self.md = iter_dict
+
+
+class BadSampleName(Exception):
```
zulcss: ditto

server/lib/pbench/indexer.py
```diff
@@ -3524,36 +3979,80 @@ def __init__(self, idxctx, tbarg, tmpdir, extracted_root):
         # It's what ties all of the relevant documents together.
         self.run_metadata["id"] = md5sum
 
+    def gen_files_by_partial_path(self, path):
+        """Generator for all files in the tar ball which match the given path
+        pattern.
+        """
+        for member in self.members:
+            if member.isfile() and member.name.find(path) >= 0:
```
zulcss: This looks fine, but can I suggest we use pathlib here? https://docs.python.org/3/library/pathlib.html

server/lib/pbench/indexer.py
```diff
@@ -3524,36 +3979,80 @@ def __init__(self, idxctx, tbarg, tmpdir, extracted_root):
...
             iterations_str = self.run_metadata["iterations"]
         except Exception:
-            # TBD - trawl through tb with some heuristics
+            # Since we don't have the iterations list in the metadata, we run
+            # through the tar ball members looking for directories that are
+            # most likely iterations.
             iterations = []
-            for path in self.tb.getnames():
-                path_els = path.split("/")
+            for member in self.members:
```
zulcss: pathlib would make life easier here as well.

server/lib/pbench/indexer.py
```diff
@@ -3708,9 +4211,9 @@ def mk_sosreports(self):
         self.idxctx.logger.debug("start")
 
         sosreports = [
-            x
-            for x in self.tb.getnames()
-            if x.find("sosreport") >= 0 and x.endswith(".md5")
+            x.name
```
zulcss: ditto

agent/bench-scripts/pbench-fio
portante: Yes, now that that PR is merged, this is dropped.

server/lib/pbench/indexer.py
portante: This is a list comprehension, so no colon needed.

server/lib/pbench/indexer.py
portante: Sure.

server/lib/pbench/indexer.py
portante: k

server/lib/pbench/indexer.py
portante: Seems like we don't gain much but syntactic changes.  Am I missing something?

server/lib/pbench/indexer.py
portante: Here `pathlib` might be helpful, since we could do `path_els = Path(member.name).parts`.

server/lib/pbench/indexer.py
portante: Not sure how `pathlib` will help with `.find("sosreport")` here, or with `.endswith(".md5")`.

server/lib/pbench/indexer.py
```diff
@@ -3524,36 +3979,80 @@ def __init__(self, idxctx, tbarg, tmpdir, extracted_root):
...
+        """
         samples = []
-        for path in self.tb.getnames():
-            if path.find("{}/".format(iteration)) < 0:
+        for member in self.members:
+            if not member.isdir():
+                # We are only looking for directories
                 continue
-            path_els = path.split("/")
+            if member.name.find("{}/".format(iteration.name)) < 0:
```
portante: Consider using an f-string here.

server/lib/pbench/indexer.py
portante: Let's track this in issue #1584 and make this a deliberate change in a separate commit.

server/lib/pbench/indexer.py
portante: Let's track this in issue #1584 and make this a deliberate change in a separate commit.

server/lib/pbench/indexer.py
portante: Let's track this in issue #1584 and make this a deliberate change in a separate commit.

server/lib/pbench/indexer.py
```diff
@@ -1040,27 +1057,380 @@ def get_result_json_dirs(ptb):
...
+                        ("@idx", 0),
+                        ("measurement_idx", 0),
+                        ("measurement_title", val_col),
+                        ("measurement_type", "number"),
+                        ("name", sample_obj.name),
+                    ]
+                )
+                sample_md = dict()
+                sample_md.update(base_sample_md)
+                sample_md["closest_sample"] = sample_obj.number
```
ndokos, line 1328: Is this because we assume that there is only one sample? Otherwise, how do we know that this is the closest sample? Or is there some comparison somewhere that I missed?

server/lib/pbench/indexer.py
```diff
@@ -1040,27 +1057,380 @@ def get_result_json_dirs(ptb):
...
+                )
+                source = _dict_const(
+                    [
+                        ("@timestamp", start_time),
+                        ("@timestamp_original", str(row[start_ts_col])),
+                        ("iteration", iteration_md),
+                        ("result", result),
+                        (
+                            "run",
+                            _dict_const(
```
ndokos, line 1378: OK, this answers a previous question.

server/lib/pbench/indexer.py
```diff
@@ -3317,8 +3697,83 @@ def hostnames_if_ip_from_sosreport(sos_file_name):
...
+
+    def __init__(self, ptb, name):
+        self._ptb = ptb
+        iter_dict = dict()
+        try:
+            items = ptb.mdconf.items(f"iterations/{name}")
+        except NoSectionError:
+            if name == "1":
+                # Old pbench-user-benchmark implementations used an iteration
+                # name of "1" for the on-disk directory, since they only every
```
ndokos, line 3716: ever

server/bin/gold/test-7.24.txt
```diff
@@ -0,0 +1,1071 @@
...
+                },
+                "name": "Hammerdb-tpcc",
+                "version": "HammerDB-3.2"
+            },
+            "iteration": {
+                "name": "1-10U",
+                "number": 1
+            },
+            "run": {
+                "config": "Maridb_tuned_TP_HTon_40P_256Gmem_with_csv",
```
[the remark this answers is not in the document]
portante: Yes, that is the name of the tar ball in the state file, `test-7.24.tar.xz`.

server/lib/pbench/indexer.py
```diff
@@ -1040,27 +1057,380 @@ def get_result_json_dirs(ptb):
...
+                "Error reading user benchmark name from {}, {}, user"
+                " benchmark data won't be indexed",
+                ubm_name,
+                exc,
+            )
+            return None
+        user_benchmark = None
+        try:
+            name, ver = first_line.split(" ", 1)
+        except ValueError:
```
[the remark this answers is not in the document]
portante:
```
$ python3
Python 3.6.5 (default, Aug 28 2018, 14:01:20)
[GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.2)] on darwin
>>> first_line = "foo"
>>> name, ver = first_line.split(" ", 1)
  File "<stdin>", line 1, in <module>
ValueError: not enough values to unpack (expected 2, got 1)
```

server/lib/pbench/indexer.py
```diff
@@ -1040,27 +1057,380 @@ def get_result_json_dirs(ptb):
...
+                        "Cpus": "40",
+                        "Memory": "263826620",
+                        "StorageType": "NVME"
+                    },
+                },
+                "iteration": {
+                    "name": "10U-NVME",
+                    "number": 1
+                },
+                "run": {
```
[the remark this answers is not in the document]
portante: Not sure what you mean by "higher up".

Typically we have "run" data, "iteration" data, and "sample" data, where run data changes only between runs, iteration data only between iterations, and sample data only between samples.

This is just the "run" data we add to all sample documents.

server/lib/pbench/indexer.py
```diff
@@ -1040,27 +1057,380 @@ def get_result_json_dirs(ptb):
...
+                "@timestamp_original": "2020.01.31.14:33:38",
+                "iteration": {
+                    "name": "10U-NVME",
+                    "number": 1
+                },
+                "result": {
+                    "@idx": 0,
+                    "value": 917474
+                },
+                "run": {
```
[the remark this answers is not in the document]
portante: > Is this run section the same as in the result data sample document?

No.  See the gold file for [test-7.24.txt](https://github.com/distributed-system-analysis/pbench/blob/591ab5d9d9d6c6d0eab5912792fb0341265e5fcf/server/bin/gold/test-7.24.txt), line [529](https://github.com/distributed-system-analysis/pbench/blob/591ab5d9d9d6c6d0eab5912792fb0341265e5fcf/server/bin/gold/test-7.24.txt#L529) vs line [571](https://github.com/distributed-system-analysis/pbench/blob/591ab5d9d9d6c6d0eab5912792fb0341265e5fcf/server/bin/gold/test-7.24.txt#L571).  You can see that for the result data _sample_ document we have verbose metadata.  For the result data document we only have identifiers.

> BTW, how should one think of the name "result data sample document"? I'm having some difficulty parsing it. Is there one such for many result data documents?

Yes, result data documents have all the identifying metadata data required at each level: run, iteration, and sample, but applied to every individual number emitted in the time-series.  This allows one to query for number time-series data that matches a given set of IDs.

Result data sample documents have the rich metadata that accompanies each level: run, iteration, and sample, but there is only one such document generated per run/iteration/sample combination.

server/lib/pbench/indexer.py
portante: Yes, right now, `pbench-user-benchmark` only generates one sample per iteration.

server/lib/pbench/indexer.py
portante: Fixed.

server/bin/gold/test-7.24.txt
```diff
@@ -0,0 +1,1071 @@
...
+        "_source": {
+            "@timestamp": "2020-02-06T10:29:26.000000",
+            "@timestamp_original": "2020.02.06.10:29:26",
+            "iteration": {
+                "name": "1-10U",
+                "number": 1
+            },
+            "result": {
+                "@idx": 0,
+                "value": 382109
```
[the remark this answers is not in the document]
portante: Not sure what you mean by this question.  This value should be the `val_col` from the table we use (line [1,104 of indexer.py](https://github.com/distributed-system-analysis/pbench/blob/37564305bb1157b2cc5f8487952c8d82985e995f/server/lib/pbench/indexer.py#L1104)).  We should be emitting a value as requested by the driver data for this benchmark.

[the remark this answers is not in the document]
portante: I have verified that the values place here are the `Tpm` value from the `.csv` files being indexed in the test.

server/lib/pbench/indexer.py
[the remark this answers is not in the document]
ndokos: Thanks!

server/lib/pbench/indexer.py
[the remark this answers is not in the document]
ndokos: Thanks!

server/lib/pbench/indexer.py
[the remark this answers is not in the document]
ndokos: I meant  higher up in the tree structure that the document encodes. I think I understand what's going on.

server/bin/gold/test-7.24.txt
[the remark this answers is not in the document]
ndokos: Yes, that's what I meant. Thanks!

portante:
Thanks @ndokos, just one more quick sign-off as I have merged the various commits down to two, one for the small server-side change, and the other being the user-benchmark indexing work.

PR #1547, opened 2020-03-18
Add single quotes to perf tool options

### LINKED ISSUE #1544: perf tool problem
A simple pbench-user-benchmark run produces this error:

```
[root@alphaville nick]# pbench-user-benchmark --config=v0.68-candidate -- sleep 120
Running sleep 120 for iteration 1-default
iostat: stopping
mpstat: stopping
perf: stopping
perf: tool is not running, nothing to kill
perf: post-processing following stop
/opt/pbench-agent/tool-scripts/perf: line 818: pushd: /var/lib/pbench-agent/pbench-user-benchmark_v0.68-candidate_2020.03.17T00.38.41/1-default/sample1/tools-default/perf: No such file or directory
/opt/pbench-agent/tool-scripts/perf: line 819: /var/lib/pbench-agent/pbench-user-benchmark_v0.68-candidate_2020.03.17T00.38.41/1-default/sample1/tools-default/perf/perf-stop-postprocess.out: No such file or directory
/opt/pbench-agent/tool-scripts/perf: line 823: popd: directory stack empty
pidstat: stopping
mpstat: post-processing following stop
iostat: no post-processing available following stop
proc-interrupts: stopping
proc-vmstat: stopping
...
```

The perf tool needs to initialize `tool_output_dir` in the stop section: it currently defines it only in the start section. Alternatively, the definition should be pulled out into common code.

Also the tool script is now chattier than it was: we probably need to redirect that output somewhere other than stdout.

portante:
Fixes #1544.

### DIFF
--- a/agent/util-scripts/pbench-register-tool-set
+++ b/agent/util-scripts/pbench-register-tool-set
@@ -161,7 +161,7 @@ for tool_name in ${the_tool_set}; do
 done
 # low overhead perf
 if [[ ${reg_perf} -ne 0 ]]; then
-	pbench-register-tool --name=perf${group_arg}${install_arg}${remotes_arg}${labels_arg} -- --record-opts="record -a --freq=100"
+	pbench-register-tool --name=perf${group_arg}${install_arg}${remotes_arg}${labels_arg} -- --record-opts="'record -a --freq=100'"
 	rc=${?}
 	if [[ ${rc} != 0 ]] ;then
 		nerrs=${nerrs}+1
--- a/agent/util-scripts/gold/pbench-register-tool-set/test-11.1.txt
+++ b/agent/util-scripts/gold/pbench-register-tool-set/test-11.1.txt
@@ -21,7 +21,7 @@ perf tool is now registered in group default
 === /var/tmp/pbench-test-utils/pbench/tools-default/mpstat:
 --interval=3
 === /var/tmp/pbench-test-utils/pbench/tools-default/perf:
---record-opts=record -a --freq=100
+--record-opts='record -a --freq=100'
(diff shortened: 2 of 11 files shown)


agent/util-scripts/pbench-register-tool-set
```diff
@@ -161,7 +161,7 @@ for tool_name in ${the_tool_set}; do
 done
 # low overhead perf
 if [[ ${reg_perf} -ne 0 ]]; then
-	pbench-register-tool --name=perf${group_arg}${install_arg}${remotes_arg}${labels_arg} -- --record-opts="record -a --freq=100"
+	pbench-register-tool --name=perf${group_arg}${install_arg}${remotes_arg}${labels_arg} -- --record-opts="'record -a --freq=100'"
```
grafuls: Double quotes here appear to be redundant.

portante: Looks are deceiving.  Try taking the double quotes out to see if the unit tests pass?  When you do, because `bash` processes the single quotes, the stored string no longer has any quotes, and so it can't be passed correctly to the tool when it is invoked.

ndokos: Someone should write an article (or perhaps a book :-) ) on quoting problems in bash: "The Joy of Quoting" - guaranteed best-seller...

ndokos: If you happen to have a copy of Kernighan & Pike's "Unix Programming Environment", read p.128. There is a PDF online at https://scis.uohyd.ac.in/~apcs/itw/UNIXProgrammingEnvironment.pdf - but you have to go to p.140 of the PDF to find p.128 of the book.

PR #1549, opened 2020-03-19
Use exec instead of fork for tools were possible

portante:
This allows us to handle special cases for stopping tools correctly, while avoiding extra code to do so.

We also add missing package names for a few tools (addresses #1550).


agent/util-scripts/testit
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+echo "\$*='${*}'"
+echo "\$@='${@}'"
```
ndokos: perl has nothing over bash in unreadability...

portante: oops ... fixed.

PR #1552, opened 2020-03-21
Fix pbench-specjbb2005's iteration handling

portante:
The `pbench-metadata-log` `end` operation looks for a `.iterations` file.  Without that file, we won't index the iteration names properly.


zulcss:
lgtm

PR #1557, opened 2020-03-22
Fix uperf XML for bidirec test type

### LINKED ISSUE #1510: pbench-uperf bidirec test-type does not actually run bidirectional
The code to generate the uperf XML uses a case statement which assumes multiple entries in the case statement can be executed in 1 pass (executing multiple entries is required for bidirec).  This is not the behavior with any recent version of bash (and maybe all versions).  Therefore the generated XML will only include one direction of traffic.

A new script dedicated to generating uperf XML (which does not use a case statement) can be found here:
https://github.com/atheurer/pbench/blob/uperf-run-benchmark/agent/bench-scripts/postprocess/uperf-prepare-xmlfile

This is for use with pbench-run-benchmark, but it should be easily adapted to work with pbench-uperf.  However, It uses uperf environment variables for many parameters, so the actual invocation of uperf (client) requires something like:
<pre> "remotehost=192.168.0.2 protocol=tcp wsize=8192 runtime=30 ./uperf"</pre>

If the env var thing is not desirable, then the script could be modified (or contents embedded in pbench-uperf) to write the hard-coded values for params like is it done today.

Another option is to deprecate "bidirec" and have users double the number of clients/server assignments but reverse order of the host assignment:

stream:
<pre>--clients=a --servers=b --test-type=stream</pre>

bidirec:
<pre>--clients=a,b --servers=b,a --test-type=stream</pre>

This is probably the most ideal because I am pretty sure, even if the XML is fixed, the post-processing is also broken because it only looks at one transaction type (Txn2).  It also requires almost no code changes :)

portante:
Fixes #1510 by emitting one group for write and another read.

### DIFF
--- a/agent/bench-scripts/pbench-uperf
+++ b/agent/bench-scripts/pbench-uperf
@@ -1,5 +1,5 @@
 #!/bin/bash
-# -*- mode: shell-script; indent-tabs-mode: t; sh-basic-offset: 8; sh-indentation: 8; sh-indent-for-case-alt: + -*-
+# -*- mode: shell-script; indent-tabs-mode: t; sh-basic-offset: 8; sh-indentation: 8; tab-width: 8 -*-
 
 # This is a script to run the uperf benchmark
 # Author: Andrew Theurer
@@ -124,9 +124,9 @@ function gen_xml {
 
 	echo "<?xml version=\"1.0\"?>"
 	echo "<profile name=\"$proto-$test_type-${size}B-${nthr}i\">"
-	echo "  <group nthreads=\"$nthr\">"
 	case $test_type in
-		rr)
+	rr)
+		echo "  <group nthreads=\"$nthr\">"
 		echo "    <transaction iterations=\"1\">"
 		echo "      <flowop type=\"connect\" options=\"remotehost=$h protocol=$proto\"/>"
 		echo "    </transaction>"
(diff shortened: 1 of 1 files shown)


agent/bench-scripts/pbench-uperf
```diff
@@ -147,8 +149,10 @@ function gen_xml {
 		echo "    <transaction iterations=\"1\">"
 		echo "      <flowop type=\"disconnect\" />"
 		echo "    </transaction>"
-		;;
-		maerts|bidirec)
+		echo "  </group>"
+		;;&
```
k-rister: According to @atheurer the behavior or `;;&` is not consistent/portable across different versions of Bash.

portante: Would you agree that given we are moving to Python 3 in the future for all this, and that our version of bash for our target environments is constrained to Fedora derived distributions, that this seems like the simplest and most straightforward change for the v0.68 release?

k-rister: All I know is that I've been told this isn't portable.  I have no idea what the set of Bash versions that this does and does not work on is, but choosing to do something that you know will break somewhere seems far from "simplest and most straight forward".

portante: Good points.

From what I have read [1], the `;;&` syntax is not POSIX, is available in `zsh` and `ksh` (derived from there), but according to the `bash` docs [2] it is well documented in `bash`.

Seems like this is a safe syntax to use with bash. 

[1] https://unix.stackexchange.com/questions/75354/can-bash-case-statements-cascade
[2] https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Conditional-Constructs

PR #1248, opened 2019-06-28
First pass implementation of the "Tool Meister"

portante:
### First pass implementation of the "Tool Meister"

See #1245.

The goal of the "Tool Meister" is to encapsulate the starting and stopping of tools into a wrapper daemon which is started once on each node for the duration of a benchmark script.  Instead of the start/stop tools scripts using SSH to start/stop tools on local or remote hosts, a Redis Server is use to communicate with all the started Tool Meisters which execute the tool start/stop operations as a result of messages they receive using Redis's publish/subscribe pattern.

As a result of this encapsulation, we provide a separation of the orchestration of running tools remotely from how they are started and stopped.  We have a CLI interface to start and stop the "Tool Meisters", which is a model that other orchestration implementations can follow (`pbench-tool-meister-start`, `pbench-tool-meister-stop`).  We also have a CLI interface for communicating with the "Tool Meisters" to send the tool `start` & `stop` messages.  However, this commit does not provide a Python3 API for that communication.

The Redis server location is passed as a set of parameters (host & port) to the Tool Meister instance, along with the name of a "key" in the Redis server which contains that Tool Meister's initial operating instructions for the duration of the benchmark script's execution:

  * What Redis pub/sub channel to use
  * What tool group describing the tools to use and their options

The Tool Meister then runs through a simple two phase life-cycle for tools until it is told to "`terminate`": "`start`" the registered tools on this host, and "`stop`" the registered tools on this host.

The initial expected phase is "`start`", where it waits to be told when to start its tools running from a published message on the "tool meister" channel. Once it starts one or more tools in the background via `screen`, it waits for a "`stop`" message to invoke the running tools' `stop` action.

This start/stop cycle is no different from the previous way tools were started and stopped, except that the start and stop operations no longer involve `ssh` operations to remote hosts.

Each `start` and `stop` message sent to the Tool Meisters is accompanied by two parameters: the tool `group` of the registered tool set (only used to ensure the context of the message is correct), and a path to a `directory` on the host (the controller) driving the benchmark where all the tool data will be collected.

Since the benchmark script ensures the directory is unique for each set of tool data collected (iteration / sample / host), the Tool Meister running on the same host as the controller just writes its collected tool data in that given directory.

[...]

PR #1563, opened 2020-03-24
Modify the name of the screen log directory to include the hostname.

ndokos:
That prevents multiple hosts from stomping on the same files.


agent/util-scripts/pbench-postprocess-tools
```diff
@@ -155,7 +155,7 @@ for this_tool_file in `/bin/ls $tool_group_dir`; do
 		debug_log "[${script_name}] ${pbench_bin}/tool-scripts/${name} --${action} --dir=${tool_output_dir} ${tool_opts[@]}"
 		if [ "$action" == "start" ]; then
 			# using screen to avoid tty issues and guarantee tool is backgrounded
-			screen_dir="${tool_output_dir}/.screen.d/${group}-${name}"
+			screen_dir="${tool_output_dir}/.screen.${hostname}.d/${group}-${name}"
```
portante: Is this `hostname` variable going to have the same value as the `hostname` variable used when the directory is moved?

portante: And won't the `.screen.${hostname}.d` not be copied by both the local `mv` and remote `tar` command use a simple `*` wildcard and not referencing the directory?

Can't wait for the Tool Meister (#1248) to fix this.

portante:
Can we consider #1564 instead?

PR #1565, opened 2020-03-25
pbench-trafficgen: harden post processing

k-rister:
- Capture the output from the post processing scripts

- Exit if process-iteration-samples generates an unexpected return
  code

### DIFF
--- a/agent/bench-scripts/pbench-trafficgen
+++ b/agent/bench-scripts/pbench-trafficgen
@@ -1137,14 +1137,18 @@ function inner_most_loop() {
 						fi
 						echo "$script_path/postprocess/$benchmark_name-postprocess \"$benchmark_results_dir\" \"$iteration\" \"$max_loss_pct\" \"$trafficgen_version\" \"$tool_label_pattern\" \"$tool_group\"" >"$benchmark_results_dir/$benchmark_name-postprocess.cmd"
 						chmod +x "$benchmark_results_dir/$benchmark_name-postprocess.cmd"
-						$benchmark_results_dir/$benchmark_name-postprocess.cmd || rc=1
+						$benchmark_results_dir/$benchmark_name-postprocess.cmd 2>&1 | tee $benchmark_results_dir/$benchmark_name-postprocess.out || rc=1
 					fi
 				done
 				if [ $rc -eq 0 ]; then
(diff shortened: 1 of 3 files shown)


k-rister:
We have anecdotal reports that `pbench-trafficgen` can get in an infinite loop where it keeps retrying a iteration/sample and we believe the reason is that `process-iteration-samples` is exiting uncleanly (ie. dying).  Investigation so far has shown that `trafficgen` itself is successfully completing the binary search so it is unclear why `process-iteration-samples` is having problems handling the resulting data.

Because this typically happens in protracted executions of `pbench-trafficgen` (over night, over the weekend, etc.) we have been unable to capture what the error is to fix it.  This proposed change modifies `pbench-trafficgen` to capture more output from the post processing phase and exit if an unexpected return code from `process-iteration-samples` is observed so that the root problem can be debugged.  If `process-iteration-samples` exits uncleanly then it is likely, if not impossible, for `pbench-trafficgen` to properly continue so aborting appears to be the wisest choice so that the root cause can be determined.

k-rister:
I remembered some more details on the problem.  We are usually running with multiple samples and each sample generates a significant amount of output.  Since `process-iteration-samples` only runs at the end of the iteration it is generally the case that when we manually detect that this problem is happening that the samples that have been retried since `process-iteration-samples` executed have completely recycled the output buffer and whatever output that `process-iteration-samples` generated has been lost.

k-rister:
This appears to have caught the following error in production performance testing:

```
Operation "ne": no method found,
        left argument in overloaded package JSON::PP::Boolean,
        right argument in overloaded package JSON::PP::Boolean at /opt/pbench-agent/bench-scripts/postprocess/process-iteration-samples line 239.
Error: values do not match
key: claimed_dev_pairs values: ARRAY(0x27b3be8) ARRAY(0x27e7f98)
[error][2020-03-26T14:38:58.482710809] process-iteration-samples generated an unknown error, aborting
```

k-rister:
I pushed a proposed fix to this problem which forces the variables being compared to always be treated as strings.  I'm guessing that we are seeing some non-deterministic behavior from Perl with respect to what types it believes a variable is because this problem was only being seen sporadically (it was iterating through the list of benchmark parameters which were not changing and it was in the second iteration that it failed -- as opposed to the first).  I suppose this is a loosely typed language issue...

k-rister:
I guess once we agree on the changes I will have to fixup the unit tests since this is creating a new file that contains the output of the `process-iteration-samples` invocation.

PR #1569, opened 2020-03-25
Change the fio version that pbench will use

ndokos:
Bump it to 3.18.


portante:
@k-rister, @atheurer, tagging you two for awareness.  We think this is going to be fine, but please let us know if you have some concerns with using 3.18 of `fio`.

k-rister:
I'm already using 3.19 so I don't have any objections.

PR #1571, opened 2020-03-25
Fix pidstat and turbostat datalogs

portante:


agent/tool-scripts/datalog/pidstat-datalog
```diff
@@ -45,7 +45,7 @@ options="${5}"
 default_options="-l -H -w -u -h -d -r ${threads_arg} -p ALL ${patterns_arg}"
 
 if [[ -z "${_tool_bin}" ]]; then
-	_tool_bin="/usr/local/bin/mpstat"
+	_tool_bin="/usr/local/bin/pidstat"
```
ndokos: Blanch - how did I miss that?

ndokos: Is there no unit test for this?

portante: I kept the original implementation characteristic where we have a hard-coded path to `/usr/local/bin` for the `sysstat` tools, as provided by the `pbench-sysstat` RPM.

But that is really silly, we should be invoking those tools from the `PATH` with `-V` looking for the version we expect.  I'll fix that for the Tool Meister release in 0.69.

PR #1583, opened 2020-03-30
Run unittests with tox

zulcss:
Tox aims to automate and standardize testing in Python. It is part of a
larger vision of easing the packaging, testing, and release process of
pbench.

Here, we setup tox environments for lint, py3, server, and agent to
allow users and developers unit tests and linting. We automate the
installation of required python dependencies through requirements.txt
and test-requirements.txt

Updated the .travis.yml to use tox.

Also updated contributors guide to include how to install tox, and to
run the unittests.


.travis.yml
```diff
@@ -1,52 +1,21 @@
...
+   - TOXENV=server
+before_install:
+  - sudo apt-get update
+install:
+    # Perl JSON and JSON-XS required
+    - sudo apt-get install libjson-perl libjson-xs-perl
+    - travis_retry pip install tox
+script:
+   - tox
+   - test -d .tox/$TOXENV/log && cat .tox/$TOXENV/log/*.log || true
```
portante, line 18: How do you ensure the "false" side of the test actually returns a failing status?

.travis.yml
```diff
@@ -1,52 +1,21 @@
+language: python
```
portante, line 1: Can you run yaml lint on this?

.travis.yml
zulcss: Actually thinking about it and looking at the documentation again it should no longer be needed.

zulcss:
Done

zulcss:
Fixed

PR #1593, opened 2020-04-03
Ansible playbook and roles for pbench-agent installation.

ndokos:
Current problems:

- remote root required.


portante:
@ndokos, is this still a work-in-progress?  I'd like to wrap up 0.69 if we can.

PR #1602, opened 2020-04-13
Fix pbench-trafficgen's use of pbench-tool-trigger

### LINKED ISSUE #1601: pbench-tool-trigger does not support the pbench-trafficgen use case correctly
The `pbench-trafficgen` benchmark uses `pbench-tool-trigger` in a unique way, compared to `pbench-user-benchmark` and `pbench-specjbb2005`.

The current interface to `pbench-tool-trigger`, a perl script, takes four (4) arguments:

 1. an iteration number
 2. a benchmark "run" directory (not a result directory)
 3. a tool group name
 4. a sample number (optional, defaults to `1`)

The first argument, the "iteration" number, is used to create iteration directories to contain tool data collected between the `start` and `stop` signatures.  As each start/stop pair in encountered, the iteration number is incremented by `pbench-tool-trigger`.

While `pbench-user-benchmark` and `pbench-specjbb2005` both pass a number for the iteration argument, `pbench-trafficgen` passes the name of an iteration directory.  This is because for the trafficgen use case, the start/stop markers only ever appear once in the output being processed, so the increment of the iteration number never occurs.

The reason why this worked prior to the [v0.68](https://github.com/distributed-system-analysis/pbench/releases/tag/v0.68) release is because `pbench-tool-trigger` never modified the iteration name.  However, starting with PR #1475, we unified the way the iteration names were being create by `pbench-user-benchmark` and `pbench-specjbb2005` by adding a `-default` to the iteration name.

We either need to modify `pbench-tool-trigger` to be hold NOT to add `-default` in the `pbench-trafficgen` use case, or modify `pbench-trafficgen` to let `pbench-tool-trigger` handle the iteration names.

portante:
Fixes #1601.

Basically, we pass the iteration number to `pbench-tool-trigger` as expected, so that `pbench-start-tools` will then create the expected hierarchy, and then `pbench-trafficgen` moves the collected tool data into the target result directory hierarchy.


portante:
@k-rister, I have added an additional test for `pbench-trafficgen`, test-39, which mocks out various things to make it work.  It is really ugly.  I had to change pbench-trafficgen a bit to squeeze in there a bit to make this all work.

agent/bench-scripts/pbench-trafficgen
```diff
@@ -693,144 +775,119 @@ fi
...
-							cmd+="${active_device_index}"
-							active_device_state=0
-							if [ ${active_devices_processed} -lt ${total_active_devices} ]; then
-								cmd+=","
-							fi
-						fi
-					else
-						error_log "couldn't find an active device [${active_device}] in the list of devices [${devices}], exiting"
-						exit 1
+		if [ $traffic_generator == "trex-txrx" -o $traffic_generator == "trex-txrx-profile" ]; then
```
ndokos, line 836: Can't this `else` be deleted and the whole `if` outdented?

PR #1606, opened 2020-04-16
Move to support uperf-1.0.7 and fio-3.19 RPM names

### LINKED ISSUE #1604: Track PRs filed against uperf to add support for -P and -R options
We filed two PRs against uperf, https://github.com/uperf/uperf/pull/20 and https://github.com/uperf/uperf/pull/21, which add the `-P` and `-R` options, and also fix up the uperf version and build maintenance.

If and when those two PRs land we should update the pbench-agent to use that new version of uperf.

portante:
Fixes #1604.

This changes the expected name of the `uperf` RPM from `pbench-uperf` to just `uperf`, while bumping the expected version to `1.0.7`.

We also add a commit for renaming the `pbench-fio` RPM to just `fio`.

### DIFF
--- a/agent/bench-scripts/pbench-uperf
+++ b/agent/bench-scripts/pbench-uperf
@@ -24,12 +24,16 @@ pbench_bin="`cd ${script_path}/..; /bin/pwd`"
 # source the base script
 . "$pbench_bin"/base
 
-benchmark_rpm=$script_name
 benchmark="uperf"
+benchmark_rpm=${benchmark}
 if [[ -z "$benchmark_bin" ]]; then
     benchmark_bin=/usr/local/bin/$benchmark
 fi
-ver=1.0.4
+ver="$(getconf.py version uperf)"
+if [[ -z "${ver}" ]]; then
+        error_log "${script_name}: package version is missing in config file"
+        exit 1
+fi
 
 # Every bench-script follows a similar sequence:
 # 1) process bench script arguments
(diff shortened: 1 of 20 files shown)


agent/bench-scripts/pbench-uperf
```diff
@@ -24,12 +24,12 @@ pbench_bin="`cd ${script_path}/..; /bin/pwd`"
...
 . "$pbench_bin"/base
 
-benchmark_rpm=$script_name
 benchmark="uperf"
+benchmark_rpm=${benchmark}
 if [[ -z "$benchmark_bin" ]]; then
     benchmark_bin=/usr/local/bin/$benchmark
 fi
-ver=1.0.4
+ver=1.0.7
```
ndokos, line 32: We should probably do this in pbench-agent-default.cfg, the same way we do fio.

agent/bench-scripts/pbench-uperf
```diff
@@ -182,10 +182,10 @@ function stop_server {
 }
 
 function install_uperf {
-	if check_install_rpm $benchmark_rpm $ver; then
-		debug_log "[$script_name]$benchmark_rpm is installed"
+	if check_install_rpm ${benchmark_rpm} ${ver}; then
+		debug_log "[$script_name] ${benchmark_rpm} is installed"
 	else
-		error_log "[$script_name]$benchmark_rpm installation failed, exiting"
+		error_log "[$script_name] ${benchmark_rpm} installation failed, exiting"
```
ndokos, line 188: We don't install any longer, so the message is wrong.

agent/bench-scripts/pbench-uperf
portante: Done.

agent/bench-scripts/pbench-uperf
portante: Fixed.

agent/bench-scripts/pbench-uperf
```diff
@@ -24,12 +24,16 @@ pbench_bin="`cd ${script_path}/..; /bin/pwd`"
...
 . "$pbench_bin"/base
 
-benchmark_rpm=$script_name
 benchmark="uperf"
+benchmark_rpm=${benchmark}
 if [[ -z "$benchmark_bin" ]]; then
     benchmark_bin=/usr/local/bin/$benchmark
 fi
-ver=1.0.4
+ver="$(getconf.py version uperf)"
```
zulcss, line 32: This will have to be updated to use pbench-config after br69 branch moves into master

PR #1616, opened 2020-04-24
Capture ${HOME}/.ssh/config if available

portante:


agent/util-scripts/pbench-metadata-log
```diff
@@ -151,6 +151,14 @@ function metadata_log_start {
     local dir=$1
     local group=$2
     local ts=$3
+
+    # Opportunistically capture the caller's SSH configuration file in
+    # case they have host specific configurations they might want to
+    # consider in the future when reviewing historical data.  The host
+    # names used in the config file might be very different from what
+    # tools capture in their output.
+    cp ${HOME}/.ssh/config ${dir}/ssh.config > /dev/null 2>&1
```
mnmehta, line 160: While we're at it, perhaps also capture /etc/ssh/ssh_config and /etc/ssh/ssh_config.d/*?

portante: Done.

agent/util-scripts/pbench-metadata-log
```diff
@@ -151,6 +151,16 @@ function metadata_log_start {
     local dir=$1
     local group=$2
     local ts=$3
+
+    # Opportunistically capture the caller's SSH configuration in case they
+    # have host specific configurations they might want to consider in the
+    # future when reviewing historical data.  The host names used in the
+    # config file might be very different from what tools capture in their
+    # output.
+    cp -L  ${HOME}/.ssh/config   ${dir}/ssh.config > /dev/null 2>&1
```
ndokos, line 160: Just a few ruminations pointing out the obvious (these are actually questions that should be added to the issue - I'll do that but I'll leave them here as well for now):

I wonder if there could be anything in .ssh/config that should be scrubbed.
Also, until we get the agent to run under a non-root user id, this is going to be the root .ssh/config.

portante:
@ndokos, I don't think anything has changed, but just need one last approval before we can merge.

PR #1617, opened 2020-04-24
Use mv instead of cp to avoid a data copy

portante:
If the reception area and the archive area are on the same volume, using a `cp` command will ALWAYS copy the data. But if we use a `mv` command then we can avoid a data copy when the reception and archive areas share a volume.  When the reception and archive areas are on separate volumes, a `mv` will just do a "copy-n-delete".

We change the algorithm to first copy the `.md5` file to the destination so that we have a well defined set of clean steps when the tar ball copy fails.

### DIFF
--- a/server/bin/pbench-server-prep-shim-002.sh
+++ b/server/bin/pbench-server-prep-shim-002.sh
@@ -153,23 +153,47 @@ while read tbmd5 ;do
         continue
     fi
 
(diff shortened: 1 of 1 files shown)


portante:
Not sure what is up with GitHub, but this Travis-CI [job](https://travis-ci.org/github/distributed-system-analysis/pbench/builds/678875181) for this PR passed.

PR #1623, opened 2020-04-27
Improves handling of ssh error in sync-satellite logs

### LINKED ISSUE #1609: Improve handling of sync satellite logs
If we have connection problems to the remote satellite host, we see the following in the `.log` file:

```
run-2020-02-20T06:35:01-UTC: start - 2020-02-20T06:35:02-UTC
ssh: connect to host pbench-satellite.example.com port 22: No route to host
```

And a single line added to the `.error` file:

```
pbench-sync-satellite: pbench-sync-satellite: pbench-remote-sync-package-tarballs: failed.
```

Let's consider capturing the `ssh` error message into a file, and adding that to a `report_status` call.

We should also consider remove the output in the error file entirely, adding it to the improved `report_status` call.

riya-17:
Fixes #1609

### DIFF
--- a/server/bin/pbench-base.sh
+++ b/server/bin/pbench-base.sh
@@ -102,7 +102,12 @@ function log_finish {
 }
 
 function log_exit {
-    echo "${PROG}: ${1}" >&4
+    local _msg="${PROG}: ${1}"
+    if [[ -z "${3}" ]]; then
+        printf -- "%s\n" "${_msg}" >&4
+    else
+        printf -- "%s\n" "${_msg}" | tee -a "${3}" >&4
+    fi
     log_finish
     if [[ -z "${2}" ]]; then
         exit 1
--- a/server/bin/pbench-remote-sync-package-tarballs
+++ b/server/bin/pbench-remote-sync-package-tarballs
@@ -22,11 +22,11 @@ if [[ -z "${dir}" || -z "${PROG}" ]]; then
 fi
 
 case $# in
-    2)
+    3)
         :
         ;;
     *)
-        echo "Usage: $PROG <satellite-config> <output-tar-ball>" >&2
+        echo "Usage: $PROG <satellite-config> <output-tar-ball> <ssh-error>" >&2
         exit 1
         ;;
 esac
@@ -34,6 +34,7 @@ esac
 
 satellite_config=$1
 output_tar_ball=$2
+errors=$3
 shift 2
 
 remote_host=$(getconf.py satellite-host $satellite_config)
@@ -47,4 +48,4 @@ if [ -z "${remote_opt}" ]; then
     exit 1
 fi
 
-ssh ${remote_host} "${remote_opt}/bin/pbench-sync-package-tarballs" > ${output_tar_ball}
+ssh ${remote_host} "${remote_opt}/bin/pbench-sync-package-tarballs" > ${output_tar_ball}  2> ${errors}
--- a/server/bin/pbench-sync-satellite.sh
+++ b/server/bin/pbench-sync-satellite.sh
@@ -118,11 +118,13 @@ typeset -i nprocessed=0
 typeset -i nfailed_md5=0
 typeset -i nerrs=0
 
+syncerr=${tmp}/syncerrors
+
 # Fetch all the tarballs from remote host's archive
-pbench-remote-sync-package-tarballs ${satellite_config} ${tmp}/satellite.${remote_prefix}.tar
+pbench-remote-sync-package-tarballs ${satellite_config} ${tmp}/satellite.${remote_prefix}.tar ${syncerr}
 rc=$?
 if [[ $rc != 0 ]] ;then
-    log_exit "$PROG: pbench-remote-sync-package-tarballs: failed." 2
+    log_exit "FAILED: $(cat ${syncerr})" 2 "${mail_content}"
 fi
 
 if [ -s $tmp/satellite.$remote_prefix.tar ]; then
--- a/server/bin/test-bin/ssh
+++ b/server/bin/test-bin/ssh
@@ -9,5 +9,8 @@ if [ "${remote_host}" = "pbench-satellite.example.com" ] ;then
     # environment from leaking into the "remote" server environment.
     unset _PBENCH_SERVER_CONFIG
     $*
+elif [ "${remote_host}" == "pbench-satellite-error.example.com" ] ;then
+    echo "ssh: connect to host ${remote_host}: No route to host" >&2
+    exit 1
 fi
 exit 0
--- a/server/bin/gold/test-28.txt
+++ b/server/bin/gold/test-28.txt
@@ -1,0 +1,163 @@
++++ Running pbench-sync-satellite satellite-one
+--- Finished pbench-sync-satellite (status=2)
++++ Running unit test audit
+Template:  pbench-unittests.v3.server-reports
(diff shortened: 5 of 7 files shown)


server/bin/pbench-remote-sync-package-tarballs
```diff
@@ -34,6 +34,8 @@ esac
 
 satellite_config=$1
 output_tar_ball=$2
+tmp=$(dirname $(dirname ${output_tar_ball}))
```
portante, line 37: Not sure we can do this, too much knowledge of what the caller is doing in order for this to work safely in all cases.

server/bin/pbench-remote-sync-package-tarballs
```diff
@@ -47,4 +49,4 @@ if [ -z "${remote_opt}" ]; then
     exit 1
 fi
 
-ssh ${remote_host} "${remote_opt}/bin/pbench-sync-package-tarballs" > ${output_tar_ball}
+ssh ${remote_host} "${remote_opt}/bin/pbench-sync-package-tarballs" | tee ${ssh_error} ${output_tar_ball} >/dev/null
```
portante, line 52: This is going to break the current method of use for `pbench-remote-sync-package-tarballs`.

The way this works today is by creating a tar ball on the remote side via `tar -cf -`, where the tar ball contents are sent to `stdout` which this script captures to a file given by the caller of `pbench-remote-sync-package-tarballs`.

I think all we have to do is capture `stderr` to a file separately, like:

```
ssh ${remote_host} "${remote_opt}/bin/pbench-sync-package-tarballs" > ${output_tar_ball} 2> ${errors}
```

The `${errors}` file could be specified as a third argument to `pbench-remote-sync-package-tarballs`.

.gitignore
```diff
@@ -10,3 +10,9 @@ dist
 .yarnrc
 *.egg-info
 .tox
+pip-wheel-metadata
```
ndokos, line 13: Aren't these changes supposed to be here? Shouldn't they disappear when rebasing on the proper underlying branch (be that master or br69)?

.travis.yml
```diff
@@ -7,6 +7,8 @@ env:
   - TOXENV=lint
   - TOXENV=agent
   - TOXENV=server
+  - TOXENV=py3-server
```
ndokos, line 10: Ditto.

server/bin/pbench-sync-satellite.sh
```diff
@@ -119,10 +119,10 @@ typeset -i nfailed_md5=0
 typeset -i nerrs=0
 
 # Fetch all the tarballs from remote host's archive
-pbench-remote-sync-package-tarballs ${satellite_config} ${tmp}/satellite.${remote_prefix}.tar
+pbench-remote-sync-package-tarballs ${satellite_config} ${tmp}/satellite.${remote_prefix}.tar ${tmp}/errors
 rc=$?
 if [[ $rc != 0 ]] ;then
-    log_exit "$PROG: pbench-remote-sync-package-tarballs: failed." 2
+    log_error "FAILED: $(cat ${tmp}/errors)" "${mail_content}"
```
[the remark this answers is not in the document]
riya-17: No nothing specific

.gitignore
riya-17: They are here as it has been rebase on br69 but later changed to master. So I had to rebase it again on master

server/bin/pbench-sync-satellite.sh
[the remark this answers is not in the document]
riya-17: @portante log_exit only take numeric argument. To log this message into reports we need ${mail_content} so, I think we should use log_error. What do you say?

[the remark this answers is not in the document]
portante: Or modify `log_exit` to accept `${mail_content}` as its third argument so it can behave as `log_error` does.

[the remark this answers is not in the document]
riya-17: ok

server/bin/pbench-base.sh
```diff
@@ -107,7 +107,12 @@ function log_exit {
     if [[ -z "${2}" ]]; then
         exit 1
     else
-        exit ${2}
+	if [[ -z "${3}" ]]; then
+	    exit ${2}
+    	else
+	    printf -- "%b\n" "${1}" | tee -a "${3}"
```
portante, line 113: Indentation problems, tabs vs spaces perhaps?

It seems we might want to consider line 105 as well.  What `log_exit` used to do was send the string, `${PROG}: ${1}` to file descriptor 4.   Now we want that same string also sent to `${3}`, if that argument exists.

So you could consider something like:

```
local _msg="${PROG}: ${1}"
if [[ -z "${3}" ]]; then
    printf -- "%s\n" "${_msg}" >&4
else
    printf -- "%s\n" "${_msg}" | tee -a "${3}" >&4
fi
log_finish
if [[ -z "${2}" ]]; then
    exit 1
else
    exit ${2}
fi
```

server/bin/test-bin/ssh
```diff
@@ -9,5 +9,8 @@ if [ "${remote_host}" = "pbench-satellite.example.com" ] ;then
     # environment from leaking into the "remote" server environment.
     unset _PBENCH_SERVER_CONFIG
     $*
+else
```
portante, line 12: I think this mock might be used by other tests eventually.  You might want to check for a specific `${remote_host}` value that would trigger the error.

riya-17: I did use the value 'pbench-satellite-error.example.com' I assigned that in the config file

portante: Yes, but it would seem you need to check for that value to take this action only when that value is provided.  That way, another test won't get tripped up by that behavior.

riya-17: ok

server/bin/pbench-base.sh
```diff
@@ -102,12 +102,17 @@ function log_finish {
 }
 
 function log_exit {
-    echo "${PROG}: ${1}" >&4
+    local _msg="${PROG}: ${1}"
+    if [[ -z "${3}" ]]; then
+	    printf -- "%s\n" "${_msg}" >&4
```
portante, line 107: 4 spaces should be used for each level of indentation.

server/bin/pbench-base.sh
```diff
@@ -102,12 +102,17 @@ function log_finish {
...
+    else
+	    printf -- "%s\n" "${_msg}" | tee -a "${3}" >&4
```
portante, line 109: 4 spaces should be used for each level of indentation.

server/bin/pbench-base.sh
```diff
@@ -102,12 +102,17 @@ function log_finish {
...
+    fi
     log_finish
     if [[ -z "${2}" ]]; then
-        exit 1
+	    exit 1
```
portante, line 113: 4 spaces should be used for each level of indentation.

server/bin/pbench-base.sh
```diff
@@ -102,12 +102,17 @@ function log_finish {
...
     else
-        exit ${2}
+	    exit ${2}
```
portante, line 115: 4 spaces should be used for each level of indentation.

riya-17:
@portante indentation really created a problem, tab gives 8 tabspace while using spaces produced error in flake8, so it was kind of cycle. I have rebased and added added the

Issue #1622, opened 2020-04-25
Improvements needed in handling iteration name lists


portante:
On staging we encountered the following error:

```
2020-04-23T02:47:02.321174 ERROR 44552 139794332383040 pbench-index-tool-data.pbench-index main 345 -- Other indexing error: iteration0,iteration1,iteration2
Traceback (most recent call last):
  File "/opt/rh/rh-python36/root/usr/lib64/python3.6/configparser.py", line 846, in items
    d.update(self._sections[section])
KeyError: 'iterations/iteration0,iteration1,iteration2'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/pbench-server/lib/pbench/indexer.py", line 3712, in __init__
    items = ptb.mdconf.items(f"iterations/{name}")
  File "/opt/rh/rh-python36/root/usr/lib64/python3.6/configparser.py", line 849, in items
    raise NoSectionError(section)
configparser.NoSectionError: No section: 'iterations/iteration0,iteration1,iteration2'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/pbench-server/bin/pbench-index.py", line 320, in main
    idxctx.es, actions, fp, idxctx.logger, idxctx._dbg
  File "/opt/pbench-server/lib/pbench/indexer.py", line 761, in es_index
    for ok, resp_payload in streaming_bulk_generator:
  File "/opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/elasticsearch1/helpers/__init__.py", line 159, in streaming_bulk
    for bulk_actions in _chunk_actions(actions, chunk_size, max_chunk_bytes, client.transport.serializer):
  File "/opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/elasticsearch1/helpers/__init__.py", line 53, in _chunk_actions
    for action, data in actions:
  File "/opt/pbench-server/lib/pbench/indexer.py", line 713, in actions_tracking_closure
    for cl_action in cl_actions:
  File "/opt/pbench-server/lib/pbench/indexer.py", line 4510, in mk_tool_data_actions
    for td in self.mk_tool_data():
  File "/opt/pbench-server/lib/pbench/indexer.py", line 4486, in mk_tool_data
    iterations = self.get_iterations()
  File "/opt/pbench-server/lib/pbench/indexer.py", line 4042, in get_iterations
    iter_objs.append(Iteration(self, iteration))
[...]

Issue #1621, opened 2020-04-25
pbench-indexer "AssertionError" with "actions_retry_deque" is not empty


portante:
Saw this error on staging, where the `es_index` action re-try `deque` (Python 3 queue object) is not empty when it should be:

```
2020-04-23T16:09:12.967620 ERROR 1574 139925273302848 pbench-sync-satellite.report post_status 232 -- Failed to post status, name = pbench-sync-satellite, timestamp = 2020-04-23T16:09:02-UTC, doctype = status, file_to_index = /srv/pbench-local/tmp/pbench-sync-satellite.1248/index_mail_contents
Traceback (most recent call last):
  File "/opt/pbench-server/lib/pbench/report.py", line 210, in post_status
    self.es, _es_payload_gen(payload_gen), sys.stderr, self.logger
  File "/opt/pbench-server/lib/pbench/indexer.py", line 857, in es_index
    assert len(actions_retry_deque) == 0
AssertionError
```

portante:
The assertion should be removed and replaced with an error log, that also provides more contextual information.

portante:
Let's consider moving this to https://github.com/distributed-system-analysis/py-es-bulk/ when we change to depend on it (see https://github.com/distributed-system-analysis/pbench/issues/1755).

riya-17:
@portante I think this issue has been taken care of in PR #1626, we can close it.

portante:
PR #1626 does not really allow us to close this issue because it only solves the problem of the error reporting.  The reason the assertion was seen in the first place still needs to be investigated.

riya-17:
@portante "2020-04-23T16:09:12.967620 ERROR 1574 139925273302848 pbench-sync-satellite.report post_status 232 -- Failed to post status, name = pbench-sync-satellite, timestamp = 2020-04-23T16:09:02-UTC, doctype = status, file_to_index = /srv/pbench-local/tmp/pbench-sync-satellite.1248/index_mail_contents" 
what does 232 meant in this line?

PR #1536, opened 2020-03-11
Refactor indexer module functions into class methods

portante:
This is a code refactoring of the indexer module to reduce the number of module level functions that effectively operate as class methods. That is, all the functions took a `PbenchTarBall` object via `ptb` variables.

As a result of this clean up, further cleanup within classes can be performed without cascading effects to all the function/method interfaces.

No functionality was changed.

This change also tightens down the `flake8` checks and proposes fixes for both the agent and server side to address problems found by `flake8`.

Of note, `configtools` has had its missing `expand_range` and `parse_range` code restored.

PR #1626, opened 2020-04-28
Address indexing related behaviors seen in the wild

### LINKED ISSUE #1620: pbench-indexer "AttributeError" when searching by IP address for host names
The following error was seen in our staging server:

```
2020-04-24T00:56:44.961121 ERROR 43498 140639375042368 pbench-index.indexer mk_tool_info 4307 -- mk_tool_info
  File "/opt/pbench-server/lib/pbench/indexer.py", line 4267, in mk_tool_info
    full_hostname = get_hostname_f_from_sos_d(sos_d, ip=host)
  File "/opt/pbench-server/lib/pbench/indexer.py", line 3516, in get_hostname_f_from_sos_d
    return search_by_ip(sos_d, ip)
  File "/opt/pbench-server/lib/pbench/indexer.py", line 3504, in search_by_ip
    if ip in d.values():
AttributeError: 'str' object has no attribute 'values'
```

This will happen when we have to search for host names by IP address.

portante:
Two commits, one to address some `black` formatting ugliness, and a second commit to address problems encountered while indexing in the wild.

First, we have seen the `action_retry_deque` assertion fire, which was fairly useless as it was because no context information was given. So we address that by turning that assertion into a logging "error", adding some additional context information to help (helps debug #1621).

Second, we add the tar ball being indexed to error and exception log messages to make it more convenient to track down the problem when debugging (helps tracking down when #1622 occurs).

Third, there is a problem deriving host names given IP addresses, so we fixed the code to make sure the logic is correct and works in all cases. We also added a new test, `test-7.25`, to verify the behavior (fixes #1620).

### DIFF
--- a/server/lib/pbench/indexer.py
+++ b/server/lib/pbench/indexer.py
@@ -553,14 +553,10 @@ class PbenchTemplates(object):
             ts_val = source["@timestamp"]
         except KeyError:
             self.counters["ts_missing_at_timestamp"] += 1
-            raise BadDate(
-                "missing @timestamp in a source document:" " {!r}".format(source)
-            )
+            raise BadDate(f"missing @timestamp in a source document: {source!r}")
         except TypeError as e:
             self.counters["bad_source"] += 1
-            raise Exception(
-                "Failed to generate index name, {!r}, source:" " {!r}".format(e, source)
-            )
+            raise Exception(f"Failed to generate index name, {e}, source: {source!r}")
         year, month, day = ts_val.split("T", 1)[0].split("-")[0:3]
         return template.format(
             prefix=self.idx_prefix,
@@ -712,12 +708,12 @@ def es_index(es, actions, errorsfp, logger, _dbg=0):
     def actions_tracking_closure(cl_actions):
         for cl_action in cl_actions:
             for field in ("_id", "_index", "_type"):
-                assert field in cl_action, "Action missing '{}' field:" " {!r}".format(
-                    field, cl_action
-                )
+                assert (
+                    field in cl_action
+                ), f"Action missing '{field}' field: {cl_action!r}"
             assert _op_type == cl_action["_op_type"], (
                 "Unexpected _op_type"
-                " value '{}' in action {!r}".format(cl_action["_op_type"], cl_action)
+                f""" value '{cl_action["_op_type"]}' in action {cl_action!r}"""
             )
 
             actions_deque.append((0, cl_action))  # Append to the right side ...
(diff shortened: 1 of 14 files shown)


server/lib/pbench/indexer.py
```diff
@@ -3483,22 +3523,38 @@ def valid_ip(address):
...
 def search_by_ip(sos_d_list, ip):
     # import pdb; pdb.set_trace()
     for sos_d in sos_d_list:
+        try:
+            host_f = sos_d["hostname-f"]
+        except KeyError:
+            continue
         for l in sos_d.values():
-            if isinstance(l, list):
+            if not isinstance(l, list):
```
ndokos, line 3551: Ugh! How stupid...

server/lib/pbench/indexer.py
```diff
@@ -3483,22 +3523,38 @@ def valid_ip(address):
...
                 continue
             for d in l:
-                if isinstance(d, dict):
+                if not isinstance(d, dict):
```
ndokos, line 3554: How many bugs can one pack in 4 lines of code?

server/lib/pbench/indexer.py
portante: @ndokos, yes, I know, ..., I broke it in PR #1536.  :(

server/lib/pbench/indexer.py
portante: A lot when you make a bad code change (PR #1536) and we don't have a unit test to catch the bug.

ndokos: I thought I had introduced those bugs from day 1 and I was being self-deprecating - next time, I'll check before opening my mouth :-)

server/lib/pbench/indexer.py
```diff
@@ -759,10 +759,10 @@ def actions_tracking_closure(cl_actions):
         try:
             resp = resp_payload[_op_type]
         except KeyError as e:
-            assert not ok, "{!r}".format(ok)
-            assert e.args[0] == _op_type, "e.args = {!r}, _op_type = {!r}".format(
-                e.args, _op_type
-            )
+            assert not ok, f"ok = {ok!r}, e = {e!r}"
```
ndokos, line 762: So these assertions remain because they have not triggered yet and they are not expected to trigger ever?

portante: Yes.

server/lib/pbench/indexer.py
```diff
@@ -4584,7 +4654,7 @@ def mk_result_data_actions(self):
                 else:
                     # Only the parent result data documents hold the tracking IDs.
                     source["@generated-by"] = self.idxctx.get_tracking_id()
-                count += 0
+                count += 1
```
portante, line 4657: @ndokos, doh!  Can't believe we missed this!

ndokos: It wasn't near the lamp post :-) - but I saw the changes in all the gold files and was wondering.

PR #1631, opened 2020-05-01
Add new "cull unpacked tar balls" service

### LINKED ISSUE #1610: Create new pbench server service to trim expired unpacked tar ball directories
We need a service to periodically remove any unpacked tar balls that are older than a specified age (in days).

This cron job should:

 * run once a day
 * first remove all "users" and "results" symlinks (handling prefix hiearchy)
 * move incoming top-level directory to a delete "directory" hierarchy
 * initiate the directory hierarchy deletions sequentially

We should also fix `pbench-unpack-tarballs.sh` to not unpack tar balls older than some time window.

portante:
Fixes #1610.

We now run a culling service once a day to find all unpacked tar ball directories whose age is beyond the configured maximum.  We also have changed the unpacked tar balls service to only unpack tar balls less than the configured maximum age.

One can keep an already unpacked tar ball by creating a `.__pbench_keep__` file in the top level of the unpacked tar ball directory.

A report is generated much like the following:

```
Culled 3 unpacked tar balls (0 errors) in 0.00 secs

Actions Taken:
  - controller-no-prefixes/tarball_culled_1970.01.01T00.00.00 (0 errors, 0.00 secs)
      $ rm results/controller-no-prefixes/tarball_culled_1970.01.01T00.00.00
      $ mv incoming/controller-no-prefixes/tarball_culled_1970.01.01T00.00.00 incoming/controller-no-prefixes/.delete.tarball_culled_1970.01.01T00.00.00
      $ rmtree incoming/controller-no-prefixes/.delete.tarball_culled_1970.01.01T00.00.00
  - controller-prefixes/tarball_culled-w-prefix_1970.01.01T00.00.00 (0 errors, 0.00 secs)
      $ rm results/controller-prefixes/pre0/pre1/pre2/tarball_culled-w-prefix_1970.01.01T00.00.00
      $ mv incoming/controller-prefixes/tarball_culled-w-prefix_1970.01.01T00.00.00 incoming/controller-prefixes/.delete.tarball_culled-w-prefix_1970.01.01T00.00.00
      $ rmtree incoming/controller-prefixes/.delete.tarball_culled-w-prefix_1970.01.01T00.00.00
  - controller-prefixes/tarball_culled-w-userA_1970.01.01T00.00.00 (0 errors, 0.00 secs)
      $ rm results/controller-prefixes/path0/path1/tarball_culled-w-userA_1970.01.01T00.00.00
      $ rm users/userA/controller-prefixes/path0/path1/tarball_culled-w-userA_1970.01.01T00.00.00
      $ mv incoming/controller-prefixes/tarball_culled-w-userA_1970.01.01T00.00.00 incoming/controller-prefixes/.delete.tarball_culled-w-userA_1970.01.01T00.00.00
      $ rmtree incoming/controller-prefixes/.delete.tarball_culled-w-userA_1970.01.01T00.00.00
```


portante:
@riya-17, @zulcss, @ndokos: FYI, this is the placeholder PR for the work to cull unpacked tar balls.

server/bin/pbench-unpack-tarballs.sh
```diff
@@ -161,6 +161,18 @@ function do_work() {
             continue
         fi
 
+        pbench-check-tb-age ${link}
```
portante, line 164: @ndokos, isn't the `pbench-check-tb-age` script used here?

portante:
This work depends on #1638 in order to reduce the total amount of change in the tests.

portante:
I have a basic test which verifies it works.  I just need to add tar balls which have prefixes and users to make sure those links are removed.

server/bin/pbench-check-tb-age.py
```diff
@@ -0,0 +1,168 @@
...
+)
+tb_pat = re.compile(tb_pat_r)
+
+
+def main(options):
+    if not options.tb_path:
+        print(
+            f"{_NAME_}: ERROR: No tar ball path specified", file=sys.stderr,
+        )
+        return 2
```
zulcss, line 38: It would be nice if the return codes are documented.

server/bin/pbench-check-tb-age.py
```diff
@@ -0,0 +1,168 @@
...
+    tb_path = os.path.realpath(options.tb_path)
+    tb_name = os.path.basename(tb_path)
+
+    if not options.cfg_name:
+        print(
+            f"{_NAME_}: ERROR: No config file specified; set"
+            " _PBENCH_SERVER_CONFIG env variable",
+            file=sys.stderr,
+        )
+        return 3
```
zulcss, line 48: ditto

server/bin/pbench-check-tb-age.py
```diff
@@ -0,0 +1,168 @@
...
+
+    try:
+        config = PbenchConfig(options.cfg_name)
+    except BadConfig as e:
+        print(f"{_NAME_}: {e}", file=sys.stderr)
+        return 4
```
zulcss, line 54: ditto

server/bin/pbench-check-tb-age.py
```diff
@@ -0,0 +1,168 @@
...
+
+    archive = config.ARCHIVE
+    archive_p = os.path.realpath(archive)
+
+    if not archive_p:
+        print(
+            f"The configured ARCHIVE directory, {archive}, does not exist",
+            file=sys.stderr,
+        )
+        return 5
```
zulcss, line 64: ditto

server/bin/pbench-check-tb-age.py
```diff
@@ -0,0 +1,168 @@
...
+
+    if not os.path.isdir(archive_p):
+        print(
+            f"The configured ARCHIVE directory, {archive}," " is not a valid directory",
+            file=sys.stderr,
+        )
+        return 6
```
zulcss, line 71: ditto

server/bin/pbench-check-tb-age.py
```diff
@@ -0,0 +1,168 @@
...
+
+    incoming = config.INCOMING
+    incoming_p = os.path.realpath(incoming)
+
+    if not incoming_p:
+        print(
+            f"The configured INCOMING directory, {incoming}, does not exist",
+            file=sys.stderr,
+        )
+        return 7
```
zulcss, line 81: ditto

server/bin/pbench-check-tb-age.py
```diff
@@ -0,0 +1,168 @@
...
+        )
+        return 8
+
+    # Fetch the configured maximum number of days a tar can remain "unpacked"
+    # in the INCOMING tree.
+    try:
+        max_unpacked_age = config.conf.get("pbench-server", "max-unpacked-age")
+    except NoOptionError as e:
+        print(f"{e}", file=sys.stderr)
+        return 9
```
zulcss, line 97: ditto

server/bin/pbench-check-tb-age.py
```diff
@@ -0,0 +1,168 @@
...
+    try:
+        max_unpacked_age = int(max_unpacked_age)
+    except Exception:
+        print(f"Bad maximum unpacked age, {max_unpacked_age}", file=sys.stderr)
+        return 10
```
zulcss, line 102: ditto

server/bin/pbench-check-tb-age.py
```diff
@@ -0,0 +1,168 @@
...
+
+    match = tb_pat.fullmatch(tb_name)
+    if not match:
+        print(f"Unrecognized tar ball name format, {tb_name}", file=sys.stderr)
+        return 11
```
zulcss, line 107: ditto

server/bin/pbench-check-tb-age.py
```diff
@@ -0,0 +1,168 @@
...
+
+    if not tb_path.startswith(archive_p):
+        print(f"Given tar ball, {tb_path}, not from the ARCHIVE tree", file=sys.stderr)
+        return 12
```
zulcss, line 111: ditto

server/bin/pbench-check-tb-age.py
```diff
@@ -0,0 +1,168 @@
...
+
+    if not os.path.exists(tb_path):
+        print(
+            f"Given tar ball, {tb_path}, does not seem to exist in the ARCHIVE tree",
+            file=sys.stderr,
+        )
+        return 13
```
zulcss, line 118: ditto

server/bin/pbench-check-tb-age.py
```diff
@@ -0,0 +1,168 @@
...
+
+    if not os.path.isdir(archive_p):
+        print(
+            f"The configured ARCHIVE directory, {archive}," " is not a valid directory",
+            file=sys.stderr,
+        )
+        return 6
+
+    incoming = config.INCOMING
+    incoming_p = os.path.realpath(incoming)
```
zulcss, line 74: This can be incorporated in one line

server/bin/pbench-cull-unpacked-tarballs.py
```diff
@@ -0,0 +1,362 @@
...
+from configparser import ConfigParser, NoOptionError
+
+from pbench import (
+    PbenchConfig,
+    BadConfig,
+    get_pbench_logger,
+)
+from pbench.indexer import _STD_DATETIME_FMT
+
+# from pbench.report import Report
```
zulcss, line 36: remove

server/bin/pbench-cull-unpacked-tarballs.py
```diff
@@ -0,0 +1,362 @@
...
+
+    # Now that all RESULTS symbolic links, and any USERS symbolic links to the
+    # incoming directory have been removed, we can rename the directory (so it
+    # is no longer visible to the web server) and then delete it.
+    if dry_run:
+        print(f"rm -r {tb_incoming_dir}")
+    else:
+        deleten = f".delete.{os.path.basename(tb_incoming_dir)}"
+        dirn = os.path.dirname(tb_incoming_dir)
+        del_path = os.path.join(dirn, deleten)
```
zulcss, line 142: consider use to try: finally:

server/bin/pbench-cull-unpacked-tarballs.py
portante: Actually, this is a place holder for a report that I have not implemented ...

server/bin/pbench-cull-unpacked-tarballs.py
portante: Not sure how `try/finally` would apply here.  If the `os.rename` fails, there is no action that I can see that we always want performed in the finally, same is true for the `shutil.rmtree` call.

Can you given an example of what you mean?

server/bin/pbench-check-tb-age.py
portante: Are you thinking of using a `;` between statements?

server/bin/pbench-check-tb-age.py
portante: The emitted error is the documentation for the return code.  We could add a function documentation string, but then we'd be duplicating the strings in the error messages.

What do you suggest?

portante:
Okay folks, we now have a unit tests for the culling behavior, `test-27`, and it also generates a report now.

server/bin/pbench-cull-unpacked-tarballs.py
portante: Latest commit now generates a report.

server/bin/pbench-check-tb-age.py
zulcss: incoming = os.path.realpath(config.INCOMING)

portante: Isn't `incoming` used in the `print` statement below?

server/bin/pbench-unpack-tarballs.sh
ndokos: But of course!

server/bin/pbench-cull-unpacked-tarballs.py
```diff
@@ -0,0 +1,457 @@
...
+    state.
+
+    Returns 0 on success, > 0 if any errors encountered.
+    """
+    duration = 0
+    if not dry_run:
+        logger.info("Began removing unpacked tar ball directory, '{}'", tb_incoming_dir)
+        start = pbench._time()
+
+    # Walk the results hierarchy for the tar ball's controller to find all
```
ndokos, line 129: Can there be more than one? Isn't that one the specified prefix? Walking the hierarchy is probably warranted to make sure that nothing is left behind, but it should be possible to retrieve the prefix (either from the prefix file in the archive or from the metadata.log) and target that only - right?

server/bin/pbench-cull-unpacked-tarballs.py
```diff
@@ -0,0 +1,457 @@
...
+            controller_name,
+            results_p,
+            users_p,
+            logger,
+            options.dry_run,
+        )
+        unpacked_dir_name = os.path.basename(tb_incoming_dir)
+        act_name = os.path.join(controller_name, unpacked_dir_name)
+        actions_taken.append((_actions_taken, act_name, errors, duration))
+        if errors > 0:
```
ndokos, line 385: So if one unpacked tar ball fails to be removed, then we abort the whole thing? Why not try them all?

server/bin/pbench-cull-unpacked-tarballs.py
```diff
@@ -0,0 +1,457 @@
...
+        duration = end - start
+        total = len(actions_taken)
+        print(
+            f"Culled {total:d} unpacked tar balls ({errors:d} errors)"
+            f" in {duration:0.2f} secs",
+            file=tfp,
+        )
+        if total > 0:
+            print(f"\nActions Taken:", file=tfp)
+        for (act_l, act_n, act_e, act_d) in actions_taken:
```
ndokos, line 412: It might be nice to document the structure of `actions_taken` somewhere: I had to jump around to figure it out and if somebody modifies this script in the future, I think they will appreciate the guidance!

server/bin/pbench-unpack-tarballs.sh
portante: We might need it, but do we need it enough right now to delay this work further?  Seems like we can take this in, and add that behavior when/if that need arises.  What do you think?

server/bin/pbench-check-tb-age.py
portante: Still not sure what end will that serve.  Can you provide a scenario for that?  Usually, like errno values, you keep the association of number to error string independent of the code using them, so that more than one piece of code can use the same numbers without having to duplicate the strings.

In this case, there is only one place where these codes have meaning.  The error codes do get used in the `INFO` log message produced by `pbench-unpack-tarballs.sh`.  In that case, the consumer of the codes is the human being reading the log output, to which they also have access to the error message emitted by `pbench-check-tb-age`.

Let's discuss further if there is another situation that needs to be considered.

server/bin/pbench-cull-unpacked-tarballs.py
portante: It certainly seems like there can be only one link.  But given this age of all the existing unpacked tar ball and successive versions of server software, this seemed like the safest way to ensure all things are cleaned up.

server/bin/pbench-cull-unpacked-tarballs.py
portante: Because we are deleting data which could be catastrophic if the errors are due to a bad configuration.  This is very conservative for sure, but seems like the best approach until we get enough experience with the errors we encounter.  We can always adjust this later when we gain familiarity with how this new service will run.

server/bin/pbench-cull-unpacked-tarballs.py
portante: I'll fix this up using a class.

server/bin/pbench-cull-unpacked-tarballs.py
```diff
@@ -0,0 +1,457 @@
...
+keep around unpacked tar balls for 30 days.
+
+An unpacked tar ball directory can be kept around longer than the maximum
+configured age by creating a ".__pbench_keep__" file in the top level of that
+directory.
+
+The culling of unpacked tar balls occurrs once a day. Each unpacked tar ball
+found in the ${INCOMING} directory hierarchy is checked against the configured
+maximum age, and removed (along with its ${RESULTS} and ${USERS} hierarchy
+links.
```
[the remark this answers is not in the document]
portante: Fixed.

server/bin/pbench-check-tb-age.py
```diff
@@ -0,0 +1,168 @@
...
+    if config._ref_datetime is not None:
+        try:
+            curr_dt = config._ref_datetime
+        except Exception:
+            # Ignore bad dates from test environment.
+            curr_dt = datetime.utcnow()
+    else:
+        curr_dt = datetime.utcnow()
+
+    # First phase is to find all the tar balls which are beyond the max
```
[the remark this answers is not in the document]
portante: Fixed.

server/bin/pbench-unpack-tarballs.sh
ndokos: I'm fine with that. We probably should open an issue after this gets merged.

server/bin/pbench-cull-unpacked-tarballs.py
ndokos: Agreed.

server/bin/pbench-cull-unpacked-tarballs.py
ndokos: OK

server/bin/pbench-unpack-tarballs.sh
portante: I'll do that; leaving this conversation unresolved to help me track that.

portante: Filed https://github.com/distributed-system-analysis/pbench/issues/1652.

PR #1638, opened 2020-05-05
Fix all date stamps to be 1970 based

portante:
This change is a pain to make, but was mostly accomplished by a few `sed` commands:

```
sed -i 's/1900-01-01/1970-01-01/g' gold/*.txt
sed -i 's/1900.01.01/1970.01.01/g' gold/*.txt
sed -i 's/1970-01-01T00:00:00.000000/1970-01-01T00:00:42.000000/g' gold/*.txt
sed -i 's/"@timestamp": "1970-01-01T00:00:00",/"@timestamp": "1970-01-01T00:00:42",/g' gold/test-*.txt
sed -i 's/1970-01-01T00:00:00-UTC/1970-01-01T00:00:42-UTC/g' gold/test-*.txt
```

Beyond that we had to update the base state used by tests 5.2, 7.`{1..7}`, 7.18, 16, 17, 20, and 22 by hand to remove "YYYY" and "1900" setup values.

The code changes are localized to:

 * `server/bin/pbench-base.sh`
 * `server/lib/pbench/__init__.py`

The rest of the changes are `gold` and `state` files.


portante:
Bummer, another test ordering problem with test-16, checking ...

portante:
I'll be fixing the instabilities in the tests in a separate commit, and once that is reviewed and merge, will revisit this one.

portante:
See #1650.

server/bin/gold/test-7.4.txt
```diff
@@ -83,12 +83,12 @@ len(actions) = 5
             "@generated-by": "70015f01dedbadbeefc105edcafedead",
             "@metadata": {
                 "controller_dir": "alphaville",
-                "file-date": "2019-03-07T02:10:59",
-                "file-name": "/var/tmp/pbench-test-server/test-7.4/pbench/archive/fs-version-001/alphaville/test-7.4.tar.xz",
-                "file-size": 1484,
-                "md5": "90c3fceba53c859d87ad66edfcc1154e",
+                "file-date": "2020-05-05T02:38:54",
+                "file-name": "/var/tmp/pbench-test-server/test-7.4/pbench/archive/fs-version-001/alphaville/test_7.4_2015.09.21T15.31.08.tar.xz",
```
ndokos, line 87: Just to test my understanding: the renaming of the tarball to include the date is *NOT* necessary for this PR. You are doing it because you are looking ahead to the culling PR - correct?

portante: Correct.

None of this is necessary by itself.  But the amount of change that results swamps future changes.

So by making all these test environment changes here, we NO other code changes, we isolate the massive amount of change to make things easier in the future.

ndokos: Do you care that e.g. in the 5.2 tar ball, there are internal tar balls with dates like 1970-01-01T00:00:00 (i.e. no 42)? I presume these will not cause problems but just wanted to point them out.

portante: I think most, if not all the tar balls, don't have 42 on them.  The 42 was for the log file entries.  I may not have been perfectly consistent on that.

PR #1650, opened 2020-05-06
Stabilize test output for unpack tar balls

portante:
The new unpack tar balls algorithm no longer sorts by size, but by modification time from newest to oldest.  We update the code comments to reflect the current behavior (which was using "change time") and fix the code to actually use "modification time" (the `%Tk` `find` printf format).


portante:
@zulcss, @riya-17, can you use the official "review" button to approve?

PR #1660, opened 2020-05-13
Add roles to tweak the firewall and add the pbench user.

ndokos:
Add a role to allow http/https traffic through the firewall and a role
to create the pbench user.  Import the roles and invoke them from the
pbench-server-config role.

Cleanup and renaming of some variables - the following variables are
assumed to be defined in the inventory file:

- pbench_repo_url_prefix: specifies the URL prefix of the repo where the
  pbench-server RPM is found. It is only used by the pbench-repo-install
  role.

- pbench_config_url: specifies the URL where the config files for specific
  server environments are found.

- pbench_config_files: specifies a list of config files that should be
  installed on a particular host.

TBD: there is a problem with this. I want to define a list of repos
in the pbench-repo-install role, but I want to be able to override it.
Putting the list in pbench-repo-install/vars/main.yml and overriding it
through an inventory group_vars does not seem to work. This needs to be
resolved.


portante:
Linting errors should be fixed by: #1661.

ndokos:
Rebased

PR #1666, opened 2020-05-15
Add server maint role, drop tool indexing

portante:
We add the notion of a maintenance role for the audit service, and drop the tool data indexing by default.

We also drop the default "vos" user for sosreports and use the default user.  This let's us remove a few unnecessary config files and their symlinks from the unit tests.

We also take this opportunity to make sure all the configuration files in the tree a properly sync'd with each other.


server/bin/state/config-satellite/pbench-server.cfg
```diff
@@ -27,12 +28,11 @@ logger_type = file
...
 # crontab roles
-#
-# For a satellite server we typically don't need to index results or handle
-# sosreports, so we override the tasks to drop those two.
 [pbench-results]
-# The results role on a satellite does not need to copy sosreports or index,
-# but we do need to make sure we run sync-cleanup.
+# The results role on our example satellite does not need to copy sosreports,
+# index data, or cull unpacked tar balls, but we do need to make sure we run
+# sync-cleanup.  As as satellite, we also only need one process unpacking tar
```
ndokos, line 34: As as -> as a
one process -> at most one process

portante:
Given we had two previous approvals on this PR, but that a small set of comment fixes obviated those reviews, and that we need to get these changes in ahead of a production push, I am going to merge this PR anyways.

Issue #1584, opened 2020-03-31
Use Python 3's "pathlib" where possible


portante:
During the review of #1543, it was pointed out that we could do much better with how we handle file paths by use Python 3's `[pathlib](https://docs.python.org/3.6/library/pathlib.html)`.

Track this in a separate issue so that somebody can take this up as a focused effort instead of trying to add it quickly to an existing PR.

portante:
Fixed by #1667.

PR #1667, opened 2020-05-15
Use pathlib where possible

riya-17:
Fixes issue #1584

### DIFF
--- a/lib/pbench/server/__init__.py
+++ b/lib/pbench/server/__init__.py
@@ -38,32 +38,24 @@ class PbenchServerConfig(PbenchConfig):
         super().__init__(cfg_name)
 
         # Now fetch some default common pbench settings that are required.
-        try:
-            self.TOP = Path(self.conf.get("pbench-server", "pbench-top-dir"))
-            if not self.TOP.is_dir():
-                raise BadConfig(f"Bad TOP={self.TOP}")
-            self.TMP = Path(self.conf.get("pbench-server", "pbench-tmp-dir"))
-            if not self.TMP.is_dir():
-                raise BadConfig(f"Bad TMP={self.TMP}")
-            self.LOGSDIR = Path(self.conf.get("pbench-server", "pbench-logs-dir"))
-            if not self.LOGSDIR.is_dir():
-                raise BadConfig(f"Bad LOGSDIR={self.LOGSDIR}")
-            self.BINDIR = Path(self.conf.get("pbench-server", "script-dir"))
-            if not self.BINDIR.is_dir():
-                raise BadConfig(f"Bad BINDIR={self.BINDIR}")
-            self.LIBDIR = Path(self.conf.get("pbench-server", "lib-dir"))
-            if not self.LIBDIR.is_dir():
-                raise BadConfig(f"Bad LIBDIR={self.LIBDIR}")
-            # the scripts may use this to send status messages
-            self.mail_recipients = self.conf.get("pbench-server", "mailto")
-            self.ARCHIVE = Path(self.conf.get("pbench-server", "pbench-archive-dir"))
-        except (NoOptionError, NoSectionError) as exc:
-            raise BadConfig(str(exc))
-        else:
-            self.INCOMING = self.TOP / "public_html" / "incoming"
-            # this is where the symlink forest is going to go
-            self.RESULTS = self.TOP / "public_html" / "results"
-            self.USERS = self.TOP / "public_html" / "users"
+        self.TOP = self._get_valid_dir_option("TOP", "pbench-server", "pbench-top-dir")
+        self.TMP = self._get_valid_dir_option("TMP", "pbench-server", "pbench-tmp-dir")
+        self.LOGSDIR = self._get_valid_dir_option(
+            "LOGSDIR", "pbench-server", "pbench-logs-dir"
+        )
+        self.BINDIR = self._get_valid_dir_option(
+            "BINDIR", "pbench-server", "script-dir"
+        )
+        self.LIBDIR = self._get_valid_dir_option("LIBDIR", "pbench-server", "lib-dir")
+        self.mail_recipients = self.conf.get("pbench-server", "mailto")
+        self.ARCHIVE = self._get_valid_dir_option(
+            "ARCHIVE", "pbench-server", "pbench-archive-dir"
+        )
+
+        self.INCOMING = self.TOP / "public_html" / "incoming"
+        # this is where the symlink forest is going to go
+        self.RESULTS = self.TOP / "public_html" / "results"
+        self.USERS = self.TOP / "public_html" / "users"
 
         try:
             self.PBENCH_ENV = self.conf.get("pbench-server", "environment")
@@ -124,6 +116,82 @@ class PbenchServerConfig(PbenchConfig):
             + " "
             + " ".join([f"WONT-INDEX.{i:d}" for i in range(1, 12)])
         )
+
+    def _get_conf(self, section, option):
+        """
+        _get_conf - get the option from the section, raising
+        BadConfig Error if it is empty, return the option value as a string.
+        """
+        try:
+            option_val = self.conf.get(section, option)
+        except (NoOptionError, NoSectionError) as exc:
+            raise BadConfig(str(exc))
+        else:
+            if not option_val:
+                raise BadConfig(f"option {option} in section {section} is empty")
+
+        return option_val
+
+    def get_conf(self, name, section, option, logger):
+        """
+        get_conf - get the option from the section, raising BadConfig Error
+        if it is empty, return the option value as a string.
+        """
+        try:
+            option_val = self._get_conf(section, option)
+        except BadConfig as exc:
+            logger.error("Bad {}= '({})'", name, exc)
+            return None
+
+        return option_val
+
+    def _get_valid_dir_option(self, req_val, section, option):
+        """_get_valid_dir_option - get the directory option from the
+        given section, raising BadConfig Error if the path is not resolved
+        or is not a directory, returning a Path directory object.
+        """
+        dir_val = self._get_conf(section, option)
+        dir_path = self._get_valid_path(req_val, dir_val, None)
+        if not dir_path:
+            raise BadConfig(f"Bad {req_val}={dir_val}")
+
+        return dir_path
+
+    def _get_valid_path(self, req_val, dir_val, logger):
+        """_get_valid_path - get realpath from the given path, raising
+        Error if the path is not resolved or is not a directory,
+        returning a Path directory object.
+        """
+        try:
+            dir_path = Path(dir_val).resolve(strict=True)
+        except FileNotFoundError:
+            if logger:
+                logger.error(
+                    "The {} directory, '{}', does not resolve to a real location",
+                    req_val,
+                    dir_val,
+                )
+            return None
+        else:
+            if not dir_path.is_dir():
+                if logger:
+                    logger.error(
+                        "The {} directory, does not resolve to a directory ('{}')",
+                        req_val,
+                        dir_path,
+                    )
+                return None
+        return dir_path
+
+    def get_valid_dir_option(self, req_val, dir_val, logger):
+        """get_valid_dir_option - get the directory path raising an Error
+        if it is not a directory, returning a Path directory object.
+        """
+        dir_path = self._get_valid_path(req_val, dir_val, logger)
+        if not dir_path:
+            return None
+
+        return dir_path
 
     def timestamp(self):
         """
--- a/server/bin/gold/test-1.txt
+++ b/server/bin/gold/test-1.txt
@@ -1,36 +1,41 @@
 +++ Running pbench-server-prep-shim-002
-pbench-server-prep-shim-002: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001
+pbench-server-prep-shim-002: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001 (config file /var/tmp/pbench-test-server/test-1/opt/pbench-server/lib/config/pbench-server.cfg)
 --- Finished pbench-server-prep-shim-002 (status=1)
 +++ Running pbench-sync-satellite satellite-one
-pbench-sync-satellite: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001
+pbench-sync-satellite: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001 (config file /var/tmp/pbench-test-server/test-1/opt/pbench-server/lib/config/pbench-server.cfg)
 --- Finished pbench-sync-satellite (status=1)
 +++ Running pbench-dispatch
-pbench-dispatch: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001
+pbench-dispatch: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001 (config file /var/tmp/pbench-test-server/test-1/opt/pbench-server/lib/config/pbench-server.cfg)
 --- Finished pbench-dispatch (status=1)
 +++ Running pbench-unpack-tarballs small
-pbench-unpack-tarballs: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001
+pbench-unpack-tarballs: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001 (config file /var/tmp/pbench-test-server/test-1/opt/pbench-server/lib/config/pbench-server.cfg)
 --- Finished pbench-unpack-tarballs (status=1)
 +++ Running pbench-copy-sosreports
-pbench-copy-sosreports: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001
+pbench-copy-sosreports: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001 (config file /var/tmp/pbench-test-server/test-1/opt/pbench-server/lib/config/pbench-server.cfg)
 --- Finished pbench-copy-sosreports (status=1)
 +++ Running pbench-index
+pbench-index: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001
 --- Finished pbench-index (status=3)
 +++ Running pbench-index --tool-data
+pbench-index-tool-data: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001
 --- Finished pbench-index (status=3)
 +++ Running pbench-clean-up-dangling-results-links
-pbench-clean-up-dangling-results-links: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001
+pbench-clean-up-dangling-results-links: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001 (config file /var/tmp/pbench-test-server/test-1/opt/pbench-server/lib/config/pbench-server.cfg)
 --- Finished pbench-clean-up-dangling-results-links (status=1)
 +++ Running pbench-cull-unpacked-tarballs
---- Finished pbench-cull-unpacked-tarballs (status=4)
+pbench-cull-unpacked-tarballs: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001
+--- Finished pbench-cull-unpacked-tarballs (status=2)
 +++ Running pbench-backup-tarballs
---- Finished pbench-backup-tarballs (status=3)
+pbench-backup-tarballs: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001
+--- Finished pbench-backup-tarballs (status=1)
 +++ Running pbench-verify-backup-tarballs
+pbench-verify-backup-tarballs: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001
 --- Finished pbench-verify-backup-tarballs (status=1)
 +++ Running pbench-satellite-cleanup
-pbench-satellite-cleanup: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench-satellite/archive/fs-version-001
+pbench-satellite-cleanup: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench-satellite/archive/fs-version-001 (config file /var/tmp/pbench-test-server/test-1/opt/pbench-server-satellite/lib/config/pbench-server.cfg)
 --- Finished pbench-satellite-cleanup (status=1)
 +++ Running unit test audit
-pbench-audit-server: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001
+pbench-audit-server: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001 (config file /var/tmp/pbench-test-server/test-1/opt/pbench-server/lib/config/pbench-server.cfg)
 --- Finished unit test audit (status=1)
 +++ var/www/html tree state (/var/tmp/pbench-test-server/test-1/var-www-html)
 lrwxrwxrwx         62 incoming -> /var/tmp/pbench-test-server/test-1/pbench/public_html/incoming
@@ -88,16 +93,6 @@ drwxrwxr-x          - public_html/users
 --- pbench tree state
 +++ pbench-local tree state (/var/tmp/pbench-test-server/test-1/pbench-local)
 drwxrwxr-x          - logs
-drwxrwxr-x          - logs/pbench-backup-tarballs
--rw-rw-r--        280 logs/pbench-backup-tarballs/pbench-backup-tarballs.log
-drwxrwxr-x          - logs/pbench-cull-unpacked-tarballs
--rw-rw-r--        226 logs/pbench-cull-unpacked-tarballs/pbench-cull-unpacked-tarballs.log
-drwxrwxr-x          - logs/pbench-index
-drwxrwxr-x          - logs/pbench-index-tool-data
--rw-rw-r--        178 logs/pbench-index-tool-data/pbench-index-tool-data.log
--rw-rw-r--        158 logs/pbench-index/pbench-index.log
-drwxrwxr-x          - logs/pbench-verify-backup-tarballs
--rw-rw-r--        241 logs/pbench-verify-backup-tarballs/pbench-verify-backup-tarballs.log
 drwxrwxr-x          - pbench-move-results-receive
 drwxrwxr-x          - pbench-move-results-receive/fs-version-002
 drwxrwxr-x          - quarantine
@@ -144,21 +139,6 @@ drwxrwxr-x          - tmp
 --- pbench-satellite-local tree state
 +++ pbench log file contents
 ++++ pbench-local/logs
-+++++ pbench-backup-tarballs/pbench-backup-tarballs.log
-1970-01-01T00:00:42.000000 ERROR pbench-backup-tarballs.pbench-backup-tarballs sanity_check -- The ARCHIVE directory /var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001, does not resolve /var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001 to a directory
------ pbench-backup-tarballs/pbench-backup-tarballs.log
-+++++ pbench-cull-unpacked-tarballs/pbench-cull-unpacked-tarballs.log
-1970-01-01T00:00:42.000000 ERROR pbench-cull-unpacked-tarballs.pbench-cull-unpacked-tarballs main -- The configured ARCHIVE directory, /var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001, is not a valid directory
------ pbench-cull-unpacked-tarballs/pbench-cull-unpacked-tarballs.log
-+++++ pbench-index-tool-data/pbench-index-tool-data.log
-1970-01-01T00:00:42.000000 ERROR pbench-index-tool-data.pbench-index main -- pbench-index-tool-data: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001
------ pbench-index-tool-data/pbench-index-tool-data.log
-+++++ pbench-index/pbench-index.log
-1970-01-01T00:00:42.000000 ERROR pbench-index.pbench-index main -- pbench-index: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001
------ pbench-index/pbench-index.log
-+++++ pbench-verify-backup-tarballs/pbench-verify-backup-tarballs.log
-1970-01-01T00:00:42.000000 ERROR pbench-verify-backup-tarballs.pbench-verify-backup-tarballs main -- The setting for ARCHIVE in the config file is /var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001, but that is not a directory
------ pbench-verify-backup-tarballs/pbench-verify-backup-tarballs.log
 ---- pbench-local/logs
 ++++ pbench-satellite-local/logs
 ---- pbench-satellite-local/logs
--- a/server/bin/gold/test-2.txt
+++ b/server/bin/gold/test-2.txt
@@ -101,7 +101,7 @@ pbench-copy-sosreports: Bad INCOMING=/var/tmp/pbench-test-server/test-2/pbench/p
 +++ Running pbench-clean-up-dangling-results-links
 --- Finished pbench-clean-up-dangling-results-links (status=0)
 +++ Running pbench-cull-unpacked-tarballs
---- Finished pbench-cull-unpacked-tarballs (status=4)
+--- Finished pbench-cull-unpacked-tarballs (status=3)
 +++ Running pbench-backup-tarballs
 Template:  pbench-unittests.v3.server-reports
 Index:  pbench-unittests.v3.server-reports.1970-01 1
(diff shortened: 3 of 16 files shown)


riya-17:
@portante I am not facing any error locally it is the lint errors that are failing. I formatted according to flake8 and black as well

server/bin/pbench-backup-tarballs.py
```diff
@@ -179,44 +180,42 @@ def backup_to_local(
...
-        if os.path.exists(backup_md5) and os.path.isfile(backup_md5):
+        if backup_md5.exists() and backup_md5.is_file():
             pass
         else:
             # backup md5 file does not exist or it is not a regular file
             logger.error(
-                "{md5} does not exist or it is not a regular file".format(
-                    md5=backup_md5
-                )
+                "{} does not exist or it is not a regular file".format(backup_md5)
```
portante, line 193: This is an example where because it is logging, we should fix this to be:

```
logging.error("{} does not exist or it is not a regular file", backup_md5)
```

server/bin/pbench-backup-tarballs.py
```diff
@@ -179,44 +180,42 @@ def backup_to_local(
...
 
         # read backup md5 file
         try:
-            with open(backup_md5) as f:
+            with backup_md5.open() as f:
                 backup_md5_hex_value = f.readline().split(" ")[0]
         except Exception:
             # Could not read file
-            logger.exception("Could not read file {md5}".format(md5=backup_md5))
+            logger.exception("Could not read file {}".format(backup_md5))
```
portante, line 203: Consider:

```
logger.exception("Could not read file {}", backup_md5)
```

server/bin/pbench-backup-tarballs.py
```diff
@@ -251,12 +250,10 @@ def backup_to_local(
                 )
 
                 # remove the copied md5 file from backup
-                bmd5_file = os.path.join(
-                    backup_controller_path, "{}.md5".format(resultname)
-                )
-                if os.path.exists(bmd5_file):
+                bmd5_file = backup_controller_path / "{}.md5".format(resultname)
```
portante, line 253: Can you also change this to use: `f"{resultname}.md5"`?

server/bin/pbench-backup-tarballs.py
```diff
@@ -251,12 +250,10 @@ def backup_to_local(
...
+                if bmd5_file.exists():
                     try:
-                        os.remove(bmd5_file)
+                        bmd5_file.remove(bmd5_file)
                     except Exception:
                         logger.exception("Unable to remove: {}".format(bmd5_file))
```
portante, line 258: Consider changing this to:

```
logger.exception("Unable to remove: {}", bmd5_file)
```

server/bin/pbench-backup-tarballs.py
```diff
@@ -409,9 +406,9 @@ def backup_data(lb_obj, s3_obj, config, logger):
             )
             continue
 
-        resultname = os.path.basename(tar)
-        controller_path = os.path.dirname(tar)
-        controller = os.path.basename(controller_path)
+        resultname = Path(tar).name
```
portante, line 409: Let's create one `Path(tar)` object, and then use the `.name` and `.parent` methods instead of instantiating it twice.

server/bin/pbench-check-tb-age.py
```diff
@@ -143,10 +144,8 @@ def main(options):
     if timediff.days > max_unpacked_age:
         # Finally, make one last check to see if this tar ball
         # directory should be kept regardless of aging out.
-        controller_p = os.path.basename(os.path.dirname(tb_path))
-        if os.path.isfile(
-            os.path.join(incoming_p, controller_p, tb_name, ".__pbench_keep__")
-        ):
+        controller_p = Path(Path(tb_path).parent).name
```
portante, line 147: Doesn't `Path(tb_path).parent` given you a `Path()` object already?  If so, why not just do, `Path(tb_path).parent.name`?

server/bin/pbench-check-tb-age.py
```diff
@@ -143,10 +144,8 @@ def main(options):
...
+        if Path.is_file(Path(incoming_p, controller_p, tb_name, ".__pbench_keep__")):
```
portante, line 148: Seems like you could do: `Path(incoming_p, controller_p, tb_name, ".__pbench_keep__").is_file()`?

server/bin/pbench-check-tb-age.py
```diff
@@ -37,7 +38,7 @@ def main(options):
         )
         return 2
     tb_path = os.path.realpath(options.tb_path)
```
portante, line 40: Can you consider making `tb_path` be a `Path()` object, perhaps via `Path(os.path.realpath(options.tb_path))`?  It seems like you'll find that one change will make it more natural to use `pathlib` here.

server/bin/pbench-backup-tarballs.py
riya-17: I changed it into this as it must have been fixed in other PR and is not reflected here due to master and br69 difference. It was the error which lint was facing so I resolved it.

server/bin/pbench-backup-tarballs.py
riya-17: ditto

server/bin/pbench-backup-tarballs.py
riya-17: ditto

server/bin/pbench-backup-tarballs.py
riya-17: ditto

server/bin/pbench-check-tb-age.py
riya-17: ok

server/bin/pbench-check-tb-age.py
riya-17: Yaa

server/bin/pbench-check-tb-age.py
riya-17: Sure

server/bin/pbench-check-tb-age.py
riya-17: Creating it a path object was creating problem. I will see

server/bin/pbench-backup-tarballs.py
riya-17: ok

server/bin/pbench-backup-tarballs.py
```diff
@@ -179,44 +180,42 @@ def backup_to_local(
         return Status.FAIL
 
     # Check if tarball exists in local backup
-    backup_tar = os.path.join(backup_controller_path, resultname)
-    if os.path.exists(backup_tar) and os.path.isfile(backup_tar):
-        backup_md5 = os.path.join(backup_controller_path, "{}.md5".format(resultname))
+    backup_tar = backup_controller_path / resultname
+    if backup_tar.exists() and backup_tar.is_file():
+        backup_md5 = backup_controller_path / "{}.md5".format(resultname)
```
portante, line 185: s/`"{}.md5".format(resultname)`/`f"{resultname}.md5"`/

server/bin/pbench-backup-tarballs.py
portante: Before we continue reviewing, let's rebase this code on br69 and then we'll continue the review.

server/bin/pbench-verify-backup-tarballs.py
```diff
@@ -91,8 +92,8 @@ def fs_entry_list_creation(self):
         self.missing_list = []
         self.error_list = []
         for tar in tarlist:
-            result_name = os.path.basename(tar)
-            controller = os.path.basename(os.path.dirname(tar))
+            result_name = Path(tar).name
```
portante, line 95: Let's consider making `tar` and `Path()` object first instead of creating it multiple times.

server/bin/pbench-check-tb-age.py
```diff
@@ -107,11 +108,11 @@ def main(options):
         print(f"Unrecognized tar ball name format, {tb_name}", file=sys.stderr)
         return 11
 
-    if not tb_path.startswith(archive_p):
+    if not str(tb_path).startswith(archive_p):
```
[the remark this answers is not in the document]
riya-17: tb_path.parents will still be a posix object due to which this statement won't work.

riya-17:
[Link to doc](https://docs.google.com/document/d/1vf9Jl4l3_LUIL7qSOfLWTIYaciCYb6V24-wxZoNvj9E/edit?usp=sharing) which explains why os.path is still in use at some places

server/bin/pbench-backup-tarballs.py
```diff
@@ -79,14 +80,16 @@ def sanity_check(lb_obj, s3_obj, config, logger):
...
+        try:
+            backuppath = backup.resolve(strict=True)
+        except FileNotFoundError:
             logger.error(
                 "The BACKUP directory {}, does not resolve to a real location", backup
             )
             lb_obj = None
 
-        if not os.path.isdir(backup):
+        if not backup.is_dir():
```
ndokos, line 92: Shouldn't this be `backuppath.is_dir()`?

server/bin/pbench-backup-tarballs.py
```diff
@@ -108,14 +111,16 @@ def sanity_check(lb_obj, s3_obj, config, logger):
...
+    try:
+        qdirpath = qdir.resolve(strict=True)
+    except FileNotFoundError:
         logger.error(
             "The QUARANTINE directory {}, does not resolve to a real location", qdir
         )
         lb_obj = None
 
-    if not os.path.isdir(qdir):
+    if not qdir.is_dir():
```
ndokos, line 123: qdirpath?

server/bin/pbench-cull-unpacked-tarballs.py
```diff
@@ -428,8 +425,8 @@ def main(options):
             logger,
             options.dry_run,
         )
-        unpacked_dir_name = os.path.basename(tb_incoming_dir)
-        act_name = os.path.join(controller_name, unpacked_dir_name)
+        unpacked_dir_name = Path(tb_incoming_dir).name
+        act_name = Path(controller_name, unpacked_dir_name)
```
ndokos, line 429: Nit: it's a path object, not a name.

server/bin/pbench-verify-backup-tarballs.py
```diff
@@ -288,8 +290,8 @@ def main():
 
     logger = get_pbench_logger(_NAME_, config)
 
-    archive = config.ARCHIVE
-    if not os.path.isdir(archive):
+    archive = Path(config.ARCHIVE)
+    if not archive.is_dir():
```
ndokos, line 294: Does `archive` and `backup` need to be resolved?

server/bin/pbench-index.py
```diff
@@ -152,16 +153,16 @@ def main(options, name):
...
-        INCOMING_rp = os.path.realpath(idxctx.config.INCOMING)
-        if not os.path.isdir(INCOMING_rp):
+        INCOMING_rp = Path(idxctx.config.INCOMING).resolve()
+        if not INCOMING_rp.is_dir():
             idxctx.logger.error("{}: Bad INCOMING={}", name, idxctx.config.INCOMING)
             res = 3
-        qdir = idxctx.config.get("pbench-server", "pbench-quarantine-dir")
-        if not os.path.isdir(qdir):
+        qdir = Path(idxctx.config.get("pbench-server", "pbench-quarantine-dir"))
+        if not qdir.is_dir():
```
ndokos, line 165: resolve it?

server/bin/pbench-index.py
```diff
@@ -274,26 +275,24 @@ def main(options, name):
...
+            ie_filename = Path(tmpdir, f"{name}.{idxctx.TS}.indexing-errors.json")
 
             for size, controller, tb in tarballs:
                 # Sanity check source tar ball path
-                linksrc_dir = os.path.dirname(tb)
-                linksrc_dirname = os.path.basename(linksrc_dir)
-                assert linksrc_dirname == linksrc, (
+                linksrc_dir = Path(tb).parent
+                linksrc_dirname = linksrc_dir.name
+                assert str(linksrc_dirname) == linksrc, (
```
ndokos, line 295: Ins't `linksrc_dirname` a string already?

server/bin/pbench-index.py
portante: Perhaps `if not qdir.resolve(strict=True).is_dir()`, and capture the error raised by `.resolve()` when the path fails?

server/bin/pbench-backup-tarballs.py
riya-17: I used backup as it was used earlier as well.

server/bin/pbench-cull-unpacked-tarballs.py
riya-17: but controller_name isn't a Path obj due to which I will have to create Path object

server/bin/pbench-backup-tarballs.py
ndokos: I think the general rule is to check whether the fully resolved target is a directory.  So I think the previous checking of `backup` is a bug.

server/bin/pbench-cull-unpacked-tarballs.py
ndokos: As I said it's a naming nit: if you call something `foo_name`, the implication is that it is the name of the object `foo` (i.e. a string); but `act_name` is not a name, it's not a string; it's a Path object, so maybe call it `act_path`.

server/bin/pbench-base.py
```diff
@@ -28,18 +29,16 @@
...
 if not parsed.cfg_name:
     # pbench-base.py is not always invoked with -C or --config or the _PBENCH_SERVER_CONFIG
     # environment variable set.  Since we really need access to the config
     # file to operate, and we know the relative location of that config file,
     # we check to see if that exists before declaring a problem.
-    config_name = os.path.join(
-        os.path.dirname(_dir), "lib", "config", "pbench-server.cfg"
-    )
-    if not os.path.exists(config_name):
+    config_name = Path(_dir.parent, "lib", "config", "pbench-server.cfg")
```
ndokos, line 40: I would feel better if `config_name` was consistenly either a string or a path: it's a path in this branch, it's a string in the else branch. I tested that `PbenchConfig()` is happy with either a string or a path, but we should be consistent (and if you decide to make it a path in all cases, then rename it to reflect that choice).

riya-17: ok sure

server/bin/pbench-base.py
```diff
@@ -28,31 +29,29 @@
...
+    if not config_path.exists:
         print(
             f"{_prog}: No config file specified: set _PBENCH_SERVER_CONFIG env variable or use"
             f" --config <file> on the command line",
             file=sys.stderr,
         )
         sys.exit(1)
 else:
-    config_name = parsed.cfg_name
+    config_path = parsed.cfg_name
```
ndokos, line 49: Make this into a Path.

server/bin/pbench-backup-tarballs.py
```diff
@@ -49,16 +49,16 @@ def __init__(
 
 def sanity_check(lb_obj, s3_obj, config, logger):
     # make sure archive is present
-    archive = config.ARCHIVE
-    archivepath = os.path.realpath(archive)
+    archive = Path(config.ARCHIVE)
+    archivepath = archive.resolve()
 
     if not archivepath:
```
portante, line 55: I believe this is a pre-existing bug, as `realpath` does not return an empty string if the given path does not exist.

With `pathlib`, we have an opportunity to fix that, with something like:

```
archive = Path(config.ARCHIVE)
try:
    archivepath = archive.resolve(strict=True)
except FileNotFound:
    logger.error("The ARCHIVE directory, '{}', does not resolve to a real location", archive)
else:
    if not archivepath.is_dir():
        logger.error("The ARCHIVE directory, '{}', does not resolve to a directory ('{}')", archive, archivepath)
```

riya-17: Ok sure

server/bin/pbench-verify-backup-tarballs.py
```diff
@@ -290,38 +290,41 @@ def main():
...
+        if not archivepath.is_dir():
+            logger.error(
+                "The setting for ARCHIVE in the config file is {}, but that is"
+                " not a directory",
+                archive,
+            )
+            return 1
 
     # add a BACKUP field to the config object
     config.BACKUP = backup = config.conf.get("pbench-server", "pbench-backup-dir")
```
ndokos, line 311: The error logged below is for this line (line 311), so we'll need a try/except here and log that error in the exception.

server/bin/pbench-verify-backup-tarballs.py
```diff
@@ -290,38 +290,41 @@ def main():
...
-    backuppath = Path(backup).resolve()
-    if not backuppath:
+    try:
+        backuppath = Path(backup).resolve(strict=True)
+    except FileNotFoundError:
```
ndokos, line 314: This exception will need a different errror logged (""The backup directory {}, does not resolve to a real location").

server/bin/pbench-backup-tarballs.py
```diff
@@ -158,11 +162,11 @@ def backup_to_local(
         # for other errors where we still want to backup in S3.
         return Status.FAIL
 
-    backup_controller_path = os.path.join(lb_obj.backup_dir, controller)
+    backup_controller_path = Path(lb_obj.backup_dir, controller)
 
     # make sure the controller is present in local backup directory
     try:
-        os.mkdir(backup_controller_path)
+        backup_controller_path.mkdir()
```
portante, line 169: Consider using the `exist_ok` parameter set to True to avoid the need for the `try/except` block.

server/bin/pbench-base.py
```diff
@@ -28,31 +29,29 @@
...
     # pbench-base.py is not always invoked with -C or --config or the _PBENCH_SERVER_CONFIG
     # environment variable set.  Since we really need access to the config
     # file to operate, and we know the relative location of that config file,
     # we check to see if that exists before declaring a problem.
-    config_name = os.path.join(
-        os.path.dirname(_dir), "lib", "config", "pbench-server.cfg"
-    )
-    if not os.path.exists(config_name):
+    config_path = Path(_dir.parent, "lib", "config", "pbench-server.cfg")
+    if not config_path.exists:
```
portante, line 41: Seems like you need to invoke the `exists` method by adding the `()`?

server/bin/pbench-index.py
```diff
@@ -183,28 +184,28 @@ def main(options, name):
         tb_glob = os.path.join(ARCHIVE_rp, "*", linksrc, "*.tar.xz")
         for tb in glob.iglob(tb_glob):
             try:
-                rp = os.path.realpath(tb)
+                rp = Path(tb).resolve()
```
portante, line 187: Seems like we need `strict=True` argument here.

server/bin/pbench-verify-backup-tarballs.py
portante: +1

If we get an empty string from the `config.conf.get()` invocation, then the `Path()` constructor will just use the current working directory to make the `Path` object.  So we need to check that we actually got a value from the config file first.

We might mean we have the same bug in the `PbenchServerConfig` class constructor.

server/bin/pbench-verify-backup-tarballs.py
portante: +1

lib/pbench/server/__init__.py
```diff
@@ -39,24 +45,38 @@ def __init__(self, cfg_name):
 
         # Now fetch some default common pbench settings that are required.
         try:
-            self.TOP = Path(self.conf.get("pbench-server", "pbench-top-dir"))
+            self.TOP = Path(
+                validate_config_val(self.conf.get("pbench-server", "pbench-top-dir"))
```
portante, line 49: Nice change.

Perhaps we can leverage this new method to reduce the overall "mechanisms" by making `validate_config_val` a method of the `PbenchConfig` base class, and then having it use `self.conf.get` directly.

Something like:

```
def _get_conf(self, section, option):
    """_get_conf - get the option from the section, raising ValueError if it is empty, return the option value as a string."""
    option_val = self.conf.get(section, option)
    if not option_val:
        raise ValueError(f"option {option} in section {section} is empty")
    return option_val
```

And while we're at it, we should DRY this out, adding `_get_valid_dir_option` method to the `PbenchConfig` base class as well:

```
def _get_valid_dir_option(self, section, option):
    """_get_valid_dir_option - get the directory option from the given section, raising a ValueError if it is not a directory, returning a Path directory object."""
    dir_val = self._get_conf(section, option)
    dir_path = Path(dir_val)
    if not dir_path.is_dir():
        raise ValueError(dir_val)
    return dir_path
```

Then we can use it like:

```
try:
    self.TOP = self._get_valid_dir_option("pbench-server", "pbench-tmp-dir")
except ValueError as exc:
    raise BadConfig(f"Bad TOP={exc}")
```

server/bin/pbench-backup-tarballs.py
```diff
@@ -108,20 +110,22 @@ def sanity_check(lb_obj, s3_obj, config, logger):
...
-        logger.error(
-            "The QUARANTINE path {}, resolves to {}" "which is not a directory",
-            qdir,
-            qdirpath,
-        )
-        lb_obj = None
+    else:
+        if not qdirpath.is_dir():
+            logger.error(
+                "The QUARANTINE path {}, resolves to {}" "which is not a directory",
```
portante, line 124: s/" "//

server/bin/pbench-index.py
```diff
@@ -153,16 +154,16 @@ def main(options, name):
 
     res = 0
     try:
-        ARCHIVE_rp = os.path.realpath(idxctx.config.ARCHIVE)
-        if not os.path.isdir(ARCHIVE_rp):
+        ARCHIVE_rp = Path(idxctx.config.ARCHIVE).resolve()
```
portante, line 157: I think we want to use `strict=True` option here.  We might want to make a helper method which takes a configuration value, turns it into a Path object, resolves it, and validates it is a directory, since we do that often.

riya-17: > It might be possible to abstract the repetitive checks for each directory into a utility function that can be imported and used by every program here (although there are some special cases: print vs logger.error and some checking is file checking, rather than directory checking).

Then should we consider the idea given by Nick?

lib/pbench/server/__init__.py
```diff
@@ -145,6 +124,27 @@ def mocked_time():
...
+            raise ValueError(f"option {option} in section {section} is empty")
+        return option_val
+
+    def _get_valid_dir_option(self, req_val, section, option):
+        """_get_valid_dir_option - get the directory option
+        from the given section, raising a ValueError if it
+        is not a directory, returning a Path directory object.
+        """
+        dir_val = self._get_conf(section, option)
+        dir_path = Path(dir_val)
```
ndokos, line 143: This will need to be resolved before checking.

lib/pbench/server/__init__.py
```diff
@@ -125,6 +124,27 @@ def mocked_time():
...
+        if not dir_path.is_dir():
+            raise BadConfig(f"Bad {req_val}={dir_path}")
+        return dir_path
+
```
ndokos, line 147: Add an externally callable method to do parts of what _get_valid_dir_option() does.

server/bin/pbench-index.py
portante: What is the idea given by Nick?

server/bin/pbench-backup-tarballs.py
```diff
@@ -48,28 +48,10 @@ def __init__(
...
-            archive,
-            archivepath,
-        )
-        return None, None
-
     # make sure the local backup directory is present
     backup = config.BACKUP
 
-    if len(backup) == 0:
+    if not backup:
```
ndokos, line 54: This should be an if./else: if backup is not defined then set lb_obj to None, but don't do the rest of that stuff that depends on a valid backup. So the try/except should be inside an else.

server/bin/pbench-backup-tarballs.py
```diff
@@ -102,25 +74,15 @@ def sanity_check(lb_obj, s3_obj, config, logger):
     # make sure the quarantine directory is present
     qdir = config.QDIR
 
-    if len(qdir) == 0:
+    if not qdir:
```
ndokos, line 77: Similarly, this should be an if/else.

lib/pbench/server/__init__.py
```diff
@@ -38,32 +38,24 @@ def __init__(self, cfg_name):
...
+        self.BINDIR = self._get_valid_dir_option(
+            "BINDIR", "pbench-server", "script-dir"
+        )
+        self.LIBDIR = self._get_valid_dir_option("LIBDIR", "pbench-server", "lib-dir")
+        self.mail_recipients = self.conf.get("pbench-server", "mailto")
+        self.ARCHIVE = self._get_valid_dir_option(
+            "ARCHIVE", "pbench-server", "pbench-archive-dir"
+        )
+
+        self.INCOMING = self.TOP / "public_html" / "incoming"
```
ndokos, line 55: We can't use get_valid_dir_option() here because we don't necessarily have a logger yet?

server/bin/pbench-cull-unpacked-tarballs.py
```diff
@@ -201,11 +199,12 @@ def remove_unpacked(tb_incoming_dir, controller_name, results, users, logger, dr
     # Now that all RESULTS symbolic links and any USERS symbolic links to the
     # incoming directory have been removed, we can rename the directory (so it
     # is no longer visible to the web server) and then delete it.
-    deleten = f".delete.{os.path.basename(tb_incoming_dir)}"
-    dirn = os.path.dirname(tb_incoming_dir)
-    del_path = os.path.join(dirn, deleten)
+    tb_incoming_dir = Path(tb_incoming_dir)
```
ndokos, line 202: @ndokos:  please check this code!

server/bin/pbench-cull-unpacked-tarballs.py
```diff
@@ -332,58 +329,23 @@ def main(options):
...
+    archivepath = config.ARCHIVE
 
-    incoming = config.INCOMING
-    incoming_p = os.path.realpath(incoming)
-
-    if not incoming_p:
-        logger.error("The configured INCOMING directory, {}, does not exist", incoming)
+    incoming = Path(config.INCOMING)
+    incomingpath = config.get_valid_dir_option("INCOMING", incoming, logger)
+    if incomingpath == 1:
```
ndokos, line 336: Return value -> None allows this to be simplified as in backuppath in pbench-backup-tarballs.py

server/bin/pbench-cull-unpacked-tarballs.py
```diff
@@ -332,58 +329,23 @@ def main(options):
...
-        return 4
-
-    results = config.RESULTS
-    results_p = os.path.realpath(results)
-
-    if not results_p:
-        logger.error("The configured RESULTS directory, {}, does not exist", results)
+    results = Path(config.RESULTS)
+    resultspath = config.get_valid_dir_option("RESULTS", results, logger)
+    if resultspath == 1:
```
ndokos, line 341: Ditto

server/bin/pbench-cull-unpacked-tarballs.py
```diff
@@ -332,58 +329,23 @@ def main(options):
...
-        return 4
-
-    users = config.USERS
-    users_p = os.path.realpath(users)
-
-    if not users_p:
-        logger.error("The configured USERS directory, {}, does not exist", users)
+    users = Path(config.USERS)
+    userspath = config.get_valid_dir_option("USERS", users, logger)
+    if userspath == 1:
```
ndokos, line 346: Ditto

server/bin/pbench-index.py
```diff
@@ -152,24 +153,34 @@ def main(options, name):
     linkerrdest = "WONT-INDEX"
 
     res = 0
+
+    ARCHIVE_rp = Path(idxctx.config.ARCHIVE)
+    ARCHIVE_path = idxctx.config.get_valid_dir_option(
+        "ARCHIVE", ARCHIVE_rp, idxctx.logger
+    )
+    if ARCHIVE_path == 1:
```
ndokos, line 161: Already checked - so can be simplified.

server/bin/pbench-index.py
```diff
@@ -152,24 +153,34 @@ def main(options, name):
...
+        res = 3
+
+    INCOMING_rp = Path(idxctx.config.INCOMING)
+    INCOMING_path = idxctx.config.get_valid_dir_option(
+        "INCOMING", INCOMING_rp, idxctx.logger
+    )
+    if INCOMING_path == 1:
```
ndokos, line 168: retval should be None on error.

server/bin/pbench-index.py
```diff
@@ -152,24 +153,34 @@ def main(options, name):
...
         qdir = idxctx.config.get("pbench-server", "pbench-quarantine-dir")
-        if not os.path.isdir(qdir):
-            idxctx.logger.error(
-                "{}: {} does not exist, or is not a directory", name, qdir
-            )
-            res = 3
-    except Exception:
-        idxctx.logger.exception("{}: Unexpected setup error", name)
-        res = 12
+    except (NoOptionError, NoSectionError) as exc:
```
ndokos, line 173: Maybe define an interface for checking this too.

server/bin/pbench-index.py
```diff
@@ -275,25 +286,23 @@ def main(options, name):
...
-            indexed = os.path.join(tmpdir, f"{name}.{idxctx.TS}.indexed")
-            erred = os.path.join(tmpdir, f"{name}.{idxctx.TS}.erred")
-            skipped = os.path.join(tmpdir, f"{name}.{idxctx.TS}.skipped")
-            ie_filename = os.path.join(
-                tmpdir, f"{name}.{idxctx.TS}.indexing-errors.json"
-            )
+            indexed = Path(tmpdir, f"{name}.{idxctx.TS}.indexed")
+            erred = Path(tmpdir, f"{name}.{idxctx.TS}.erred")
+            skipped = Path(tmpdir, f"{name}.{idxctx.TS}.skipped")
+            ie_filename = Path(tmpdir, f"{name}.{idxctx.TS}.indexing-errors.json")
```
ndokos, line 300: Only it's not a filename any longer, so the variable name should be changed.

server/bin/gold/test-2.txt
```diff
@@ -101,7 +101,7 @@ pbench-copy-sosreports: Bad INCOMING=/var/tmp/pbench-test-server/test-2/pbench/p
 +++ Running pbench-clean-up-dangling-results-links
 --- Finished pbench-clean-up-dangling-results-links (status=0)
 +++ Running pbench-cull-unpacked-tarballs
---- Finished pbench-cull-unpacked-tarballs (status=4)
+--- Finished pbench-cull-unpacked-tarballs (status=3)
```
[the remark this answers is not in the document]
riya-17: At some places there are different status for realpath error and directory not found error due to which this change has occured.

lib/pbench/server/__init__.py
riya-17: because we need self.TOP for this option

server/bin/gold/test-1.txt
```diff
@@ -1,36 +1,41 @@
...
 +++ Running pbench-unpack-tarballs small
-pbench-unpack-tarballs: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001
+pbench-unpack-tarballs: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001 (config file /var/tmp/pbench-test-server/test-1/opt/pbench-server/lib/config/pbench-server.cfg)
 --- Finished pbench-unpack-tarballs (status=1)
 +++ Running pbench-copy-sosreports
-pbench-copy-sosreports: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001
+pbench-copy-sosreports: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001 (config file /var/tmp/pbench-test-server/test-1/opt/pbench-server/lib/config/pbench-server.cfg)
 --- Finished pbench-copy-sosreports (status=1)
 +++ Running pbench-index
+pbench-index: Bad ARCHIVE=/var/tmp/pbench-test-server/test-1/pbench/archive/fs-version-001
```
portante, line 17: Some of the changes here include the full path to the configuration file, which is a good change.  Is there a reason why this one does not?

server/bin/gold/test-6.1.txt
```diff
@@ -107,9 +106,6 @@ drwxrwxr-x          - tmp
 --- pbench-satellite-local tree state
 +++ pbench log file contents
 ++++ pbench-local/logs
-+++++ pbench-backup-tarballs/pbench-backup-tarballs.log
```
portante, line 110: Is there a reason why this test output was removed?

server/bin/gold/test-6.2.txt
```diff
@@ -107,9 +106,6 @@ drwxrwxr-x          - tmp
 --- pbench-satellite-local tree state
 +++ pbench log file contents
 ++++ pbench-local/logs
-+++++ pbench-backup-tarballs/pbench-backup-tarballs.log
```
portante, line 110: Something seems wrong with this test output being removed.

lib/pbench/server/__init__.py
```diff
@@ -125,6 +117,77 @@ def mocked_time():
...
+        try:
+            dir_path = Path(dir_val).resolve(strict=True)
+        except FileNotFoundError:
+            raise BadConfig(f"Bad {req_val}={dir_val}")
+        else:
+            if not dir_path.is_dir():
+                raise BadConfig(f"Bad {req_val}={dir_path}")
+        return dir_path
+
+    def get_conf(self, name, section, option, logger):
```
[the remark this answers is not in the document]
riya-17: Yaa, that will be a good idea

server/bin/gold/test-1.txt
riya-17: I have no idea why some of them does not output that. I will look into it.

server/bin/gold/test-6.1.txt
riya-17: Earlier ARCHIVE was not being checked in pbench/server/__init__.py , they were being checked in each script separately. Now ARCHIVE being checked in pbench/server/__init__.py, so there are no logs that are being produced in that script due to which .log file is not created and removed from logs. Same log have been created in line-2 of same test due to ARCHIVE being checked in pbench/server/__init__.py

server/bin/gold/test-6.2.txt
riya-17: Earlier ARCHIVE was not being checked in pbench/server/init.py , they were being checked in each script separately. Now ARCHIVE being checked in pbench/server/init.py, so there are no logs that are being produced in that script due to which .log file is not created and removed from logs. Same log have been created in line-2 of same test due to ARCHIVE being checked in pbench/server/init.py

riya-17:
@portante for handling py3-server test cases should I use sys.exit(1) when the BadConfig Error is raised

server/bin/gold/test-1.txt
riya-17: this (config file ...) statement is created by pbench-base.py script

lib/pbench/server/__init__.py
```diff
@@ -125,6 +117,72 @@ def mocked_time():
...
+                    "The {} directory, '{}', does not resolve to a real location",
+                    req_val,
+                    dir_val,
+                )
+            return None
+        else:
+            if not dir_path.is_dir():
+                if logger:
+                    logger.error(
+                        "The ARCHIVE directory, '{}', does not resolve to a directory ('{}')",
```
ndokos, line 178: This should not refer to ARCHIVE: it should be a generic {} filled in from an argument.

server/bin/pbench-backup-tarballs.py
```diff
@@ -327,25 +286,29 @@ def backup_data(lb_obj, s3_obj, config, logger):
...
+                "Quarantine: {}, {} does not exist or it is not a regular file",
+                tb,
+                tar,
             )
             continue
 
-        archive_md5 = f"{tar}.md5"
-
+        archive_md5 = Path(f"{tar}.md5")
         # check md5 file exist and it is a regular file
```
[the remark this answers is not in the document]
riya-17: seems like it is being checked in 311 line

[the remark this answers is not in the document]
ndokos: I meant: fix the comment to read "check that the md5 file exists ..."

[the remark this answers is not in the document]
riya-17: Did that.. I misunderstood earlier but corrected it :+1:

PR #1673, opened 2020-05-21
Refactor setuptools installation

zulcss:
Keeping in mind that pbench will become modular in the future. Simplify the pbench python namespace so that it acts more of a "library" layer betwen agent and server. That is it wont install any cli but it provides methods and classes in order to run the server or agent.

Also refactor the tests infrastructure so that we can split it out into functional tests and unit tests so that we can eventually replace the current shell scripts with a python/tox/pytest infrastructure.


zulcss:
> Looks good, but can you squash to one commit with the same subject line as the PR line?

Done

lib/pbench/cli/getconf.py
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/python3
```
portante, line 1: Why do we still have `getconf.py`?

portante: Never mind, this is the module that `pbench-config` generated command actually uses.

PR #1676, opened 2020-05-21
Modify the config install action to work with ssh

ndokos:
Instead of having the remotes use getURL to fetch their config files
(which is not always possible, e.g. if the remote is outside the
firewall), fetch the config files locally and store them under a
temporary directory, named partially after the inventory hostname, and
then copy them using ssh to the remotes.

The assumption is that the ansible control machine is inside the
firewall, so it can use getURL to fetch the config files, but it can
also ssh to the remotes to copy the file(s), even if (some of) the
remotes are outside the firewall.


portante:
@ndokos, can you rebase and then we'll merge?

ndokos:
Rebased.

PR #1682, opened 2020-05-26
Fix user-benchmark data indexing for dashboard

portante:
The pbench dashboard expects a few additional fields that were not originally provided when indexing user-benchmark data.

### DIFF
--- a/server/bin/gold/test-7.24.txt
+++ b/server/bin/gold/test-7.24.txt
@@ -543,11 +543,16 @@ len(actions) = 16
                 },
                 "closest_sample": 1,
                 "end": "2020-02-06T10:46:32.000000",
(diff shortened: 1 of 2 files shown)


gurbirkalsi:
@portante The dashboard looks for a specific role value of `aggregate` and doesn't handle n/a values. Does n/a mean there was only one client for the sample?

portante:
Bummer, I'll change the value to aggregate.

portante:
> @portante The dashboard looks for a specific role value of `aggregate` and doesn't handle n/a values. Does n/a mean there was only one client for the sample?

Fixed by https://github.com/distributed-system-analysis/pbench/pull/1685.

PR #1684, opened 2020-05-27
Fix pbench-uperf default client handling

portante:
When `--clients` is not specified the `uperf_identifier` variable was not constructed the same way in the three code blocks that required it.

This is an example of what happens when code is replicated multiple times (not following the DRY principle).  This change does not try to solve the DRY problem, but restores the proper functionality.  A more invasive commit will be needed to apply the DRY principle.


portante:
@ndokos, can you re-apply your review?  I had to rebase this PR.

PR #1686, opened 2020-05-28
Add agent, server and web-server Makefiles

ndokos:
The agent and web-server Makefiles provide an install target.

For the server Makefile, there are two targets: `install' and `install-build-artifacts'.  The first one can be used interactively to do a local server installation. The second one is only to be used when building an RPM: it installs SHA1 and SEQNO files that are not generally available: they are only produced during an RPM build.

They also include a `clean' target for debugging convenience: `make clean' removes the corresponding local installation.

The installation location is controlled by the DESTDIR variable (by default, `/opt/pbench-server' and `/opt/pbench-web-server' respectively).  The default can be overridden on the command line.

In addition, we add a `package.json' file for the web-server npm install (and remove the patch that added a different package.json file to the server installation - it is not clear why that was a patch).

We also make VERSION a file local to web-server/, not a link. That follows the trend set by PR #1670. The top-level VERSION file is then useless and we get rid of it.

Finally, there is an opportunistic fix for a bug that reared its ugly head when testing the above:

pbench-server-activate-create-crontab role: fix PATH

Add the server bin/ path at the beginning of PATH in the execution environment instead of at the end: that avoids unpleasant surprises if e.g. the agent is already installed.


portante:
Let's use `Draft` mode for the PR instead of the `WIP` label.

ndokos:
Fixed the unit test failures and squashed, I still need to rewrite the description.

PR #1691, opened 2020-06-04
Fix server ansible roles.

ndokos:
pbench-clean-yum-cache role: (new)
Add a pbench-clean-yum-cache role: the ansible package module is not
always successful in updating to the latest package, particularly on RHEL7
boxes. This is an attempt to help it do that, but it remains to be seen how
effective it is.

pbench-server-install role:
Import the pbench-clean-yum-cache role so that it is executed before we try
to install/update packages.

pbench-server-user role:
Add some  SELinux settings to the /home/<pbench>/.ssh directory.

Fix comments and add pbench_repo_url_prefix variable to the inventory example
file.

### DIFF
--- a/server/ansible/roles/pbench-clean-yum-cache/tasks/main.yml
+++ b/server/ansible/roles/pbench-clean-yum-cache/tasks/main.yml
@@ -1,0 +1,4 @@
+---
+- name: Clean yum cache
+  command: yum clean all
+
--- a/server/ansible/roles/pbench-clean-yum-cache/tasks/main.yml
+++ b/server/ansible/roles/pbench-clean-yum-cache/tasks/main.yml
@@ -2,3 +2,9 @@
 - name: Clean yum cache
   command: yum clean all
 
+- name: Delete /var/yum/cache directory
+  file:
+    path: /var/yum/cache
+    state: absent
+
+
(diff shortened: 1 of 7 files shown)


server/ansible/roles/pbench-clean-yum-cache/tasks/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+- name: Clean yum cache
+  command: yum clean all
+
```
portante, line 4: Can we also add `rm -rf /var/cache/yum`?

Probably with the "directory" / "absent" file module?

server/ansible/roles/pbench-server-user/tasks/main.yml
```diff
@@ -27,3 +27,12 @@
     owner: "{{ pbench_user }}"
     group: "{{ pbench_group }}"
     state: directory
+
+- name: set SELinux context on /home/<pbench>/.ssh
+  sefcontext:
```
portante, line 32: `s/sefcontext/setcontext/`?

server/ansible/roles/pbench-clean-yum-cache/tasks/main.yml
ndokos: Sure.

server/ansible/roles/pbench-server-user/tasks/main.yml
ndokos: No, the module is called [sefcontext](https://docs.ansible.com/ansible/latest/modules/sefcontext_module.html).

portante:
TravisCI job passed (https://travis-ci.org/github/distributed-system-analysis/pbench/builds/695184688).

PR #1692, opened 2020-06-06
More refactoring for logging and unit tests

portante:
Ahead of trying to land PR #1248, we need to make a few other substantive changes to the code base to avoid unrelated changes to the Tool Meister work.

There are 7 commits in this PR, kept separately on purpose, to allow for easier review.

 * Refactoring to introduce `PbenchAgentConfig` class

   Leverages more shared code, we have created a `PbenchAgentConfig` class which is a sub-class of the common `PbenchConfig` class, allowing us to share more code between the server and agent environments.

   Having the `PbenchConfig` class shared with the agent side as well allows us to introduce the logging infrastructure, previously only available to the server side, to the agent side as well.

   Along the way, we address a number of issues in the Python3 code found while re-working unit tests, including the proper use of temporary directories.

 * Implement `PbenchConfig` base class

   We restructure the `PbenchConfig` class to be a base class, and move the pbench server specific components to a `PbenchServerConfig` sub-class.

   We also remove the `ServerConfig` implementation, and fix up the pytest environment so that it users the real `pbench-server-default.cfg` file.

 * Move server side logging to common

   The logging mechanism for the server is not really code specific to the server.  We want to use it for the agent, so we are just doing a code move first.

 * Run agent side test directories individually

   This let's us use `tox -e util-scripts -- test-51` so that we can run individual tests in each directory where we have unit tests.

 * Remove explicit sub-classing of `object`

   Sub-classing of `object` in Python3 is now implicitly done for all classes.

   As a result of this change, we also fix up `flake8` errors for two `datalog` Python3 modules.

 * Change `pbench-dbench` to stop using `/tmp`

 * Add `py3-functional` to `run-unittests`


portante:
This PR pulls two commits out of PR #1248 so that they can be considered independently.

pbench/agent/results.py
```diff
@@ -3,56 +3,50 @@
...
 from pbench.cli.agent.commands.log import add_metalog_option
 from pbench.common import configtools
 
 
-class MakeResultTb:
-    def __init__(self, result_dir, target_dir, user, prefix, _logger=None):
-        self.result_dir = result_dir
-        self.target_dir = target_dir
+class MakeResultTb(object):
+    def __init__(self, result_dir, target_dir, user, prefix, config, logger):
```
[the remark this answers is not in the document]
portante: I am not a fan of global variables in general, an especially not in libraries.  We want library methods to use the logger of their caller, not their own internal logger.

pbench/agent/__init__.py
```diff
@@ -1,7 +1,91 @@
...
+    """
+
+    def __init__(self, cfg_name):
+        super().__init__(cfg_name)
+
+        try:
+            # Provide a few convenience attributes.
+            self.agent = self.conf["pbench-agent"]
+            self.results = self.conf["results"]
+            # Now fetch some default common pbench settings that are required.
```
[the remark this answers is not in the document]
portante: Sure, changing this to use `@property` here would be good to do, but not sure there is really a gain with efficiency because the config class is not long lived in agent code with lots of accesses.

For the `pbench-config` use case it of pulling one properly out of a config file, we should address that efficiency in other ways.  One of which could be to export all commonly used configuration values as environment variables once in `agent/base`, another would be to ignore this until the python re-write completes.

portante:
I have rebased this PR on Chuck's latest commit to the `br69` branch, PR #1673.  This was a rather complicated rebase due to a number of conflicts and code moves.  I have also applied and/or responded to all code review comments up to this point.

lib/pbench/common/constants.py
```diff
@@ -1,3 +1,8 @@
 import os
 
 AGENT_PATH = os.path.join(os.path.dirname(os.path.realpath(__file__)), "../../agent")
```
portante, line 3: Unused, remove.

lib/pbench/test/unit/server/common.py
```diff
@@ -1,7 +1,2 @@
 temp_dir = "/tmp/"
```
portante, line 1: Unused, remove.

CONTRIBUTION.md
```diff
@@ -1,5 +1,6 @@
 # CONTRIBUTION GUIDELINES FOR PBENCH
 
+
```
portante, line 3: Remove unnecessary blank line addition.

lib/pbench/common/constants.py
portante: Fixed.

lib/pbench/test/unit/server/common.py
portante: Fixed.

CONTRIBUTION.md
portante: Fixed.

lib/pbench/__init__.py
```diff
@@ -17,7 +17,7 @@
 _STD_DATETIME_FMT = "%Y-%m-%dT%H:%M:%S.%f"
 
 
-class PbenchConfig(object):
+class PbenchConfig:
```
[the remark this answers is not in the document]
portante: See https://github.com/distributed-system-analysis/pbench/pull/1692/commits/467bf57f41d16e553c7289fb7f4e1fac0d4b3e9c.  There was a review comment asking to remove the addition of `(object)` to class definitions since in Python3 you no longer need to sub-class `object` explicitly.

lib/pbench/agent/results.py
```diff
@@ -90,70 +101,94 @@ def make_result_tb(self):
...
+                tarball.unlink()
+            except OSError as exc:
+                if exc.errno != errno.ENOENT:
+                    self.logger.error("error removing failed tar ball, {}", tarball)
             sys.exit(1)
 
         self.make_md5sum(tarball)
 
-        return tarball
+        # The contract with the caller it to just return the full path to the
```
ndokos, line 122: it --> is

lib/pbench/agent/results.py
```diff
@@ -90,70 +101,94 @@ def make_result_tb(self):
...
+            )
+            sys.exit(1)
 
-    def post_file(self, file):
-        filename = secure_filename(os.path.join(self.path, file))
-        with open(f"{os.path.join(self.path, file)}.md5.check", "r") as _check:
-            md5sum = _check.read()
+        with open(self.tarball_md5, "r") as md5fp:
+            md5sum = md5fp.read()
+        filename = secure_filename(str(self.tarball))
```
ndokos, line 189: What does this protect us from?

portante: This came from the original code, but near as I can tell, `secure_filename()` munges a full POSIX file name path into a single file name so that the `PUT` request can't cause the server to access an unwanted path.

For example, if the tar ball path is `/var/tmp/send/tarball.tar.xz`, the "secure" filename sent over the wire in the PUT headers would be `var_tmp_send_tarball.tar.xz`.

lib/pbench/agent/results.py
portante: Fixed.

PR #1702, opened 2020-06-22
Rework pbench-agent ansible playbooks

ndokos:
Rework the ansible playbooks to make them as similar as possible to
the server ansible playbooks (PR#1199). In particular, fix the config
and ssh key install to allow hosts outside the firewall to be
configured (as long as the ansible host is inside).

Usage:

     ansible-playbook --user=root -i /path/to/inventory.file \
                      pbench-agent-install.yml \
                      [ --extra-vars='{cenv: <environment>}' ]

Two other playbooks use subsets of the defined roles:
pbench-repo-install.yml installs the COPR repos only. You can then
install the pbench-agent RPM manually.  In that case, the third
playbook can be used to install the config and key files afterwards.

An example inventory file, Inventory/pbench-agent.hosts.example, is
provided. The second section of the file contains the definition of
various variables that are used by the playbooks.  These are set to
dummy values here, so a site admin should modify this file for a
particular installation and make it available to the users for that
site.

Users should then copy and mofify the example file, without touching
the variables section: they should just add their hosts to the first
section (possibly augmented with host varirable definitions,
e.g. ansible_python_interpreter, that might be need to work around
problems on some distros).


agent/ansible/roles/pbench-agent-config-install/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+pbench_agent_install_dir: /opt/pbench-server
```
portante, line 2: s/server/agent/?

agent/ansible/roles/pbench-agent-config-install/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
+---
+# Install config file for pbench agent or server (the inventory file
```
portante, line 2: Can we just talk about the agent here instead of mentioning the server as well?

agent/ansible/roles/pbench-agent-key-install/tasks/main.yml
```diff
@@ -0,0 +1,37 @@
+---
+# Install key file for pbench agent 
+
+- name: create temp dir to store the config locally
```
portante, line 4: s/to store the config/to store the key/

agent/ansible/roles/pbench-agent-key-install/tasks/main.yml
```diff
@@ -0,0 +1,37 @@
...
+    prefix: "{{ inventory_hostname }}-"
+  register: tempdir_1
+
+- name: relax perms
+  delegate_to: localhost
+  file:
+    path: "{{ tempdir_1.path }}"
+    mode: 0755
+
+- name: install the config file(s) locally
```
portante, line 17: s/config/key/

PR #1705, opened 2020-06-24
Fixes to ansible playbooks/roles

ndokos:
Fixes to agent and server playbooks/roles/sample inventiory files.
    
 - recognize CentOS
 - use pbench_configuration_environment in the sample inventory file and set its value to either a host-specific value or the configured default ('production')
 - stop command from complaining when running `yum clean`

Commits:
1. agent ansible fixes (see above)
2. server ansible fixes (see above)
3. agent ansible: install both pbench-agent and pbench-sysstat

### DIFF
--- a/server/ansible/roles/pbench-clean-yum-cache/tasks/main.yml
+++ b/server/ansible/roles/pbench-clean-yum-cache/tasks/main.yml
@@ -1,6 +1,8 @@
 ---
 - name: Clean yum cache
   command: yum clean all
+  args:
+    warn: False
 
 - name: Delete /var/yum/cache directory
   file:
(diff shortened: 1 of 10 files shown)


server/ansible/roles/pbench-clean-yum-cache/tasks/main.yml
```diff
@@ -1,10 +1,12 @@
 ---
```
portante, line 1: Is there a way to make this role common or perhaps see if there is a galaxy role already that does this?

ndokos: I don't know if there is one already, but I am planning to make it common when moving this stuff up to Galaxy. There is a bunch of duplication between server and agent that is going to go away then.

PR #1710, opened 2020-06-25
Pbench rpm

ndokos:
Pull the RPM making machinery for the agent into the upstream repo.

```
cd agent/rpm
make rpm
```
should work <s>(but it still uses the roundabout upstream-tarball ways - getting rid of that is TBD)</s>. THere are still some ugly hacks, but it's pretty close: the `agent/Makefile` provides an install target that installs to DESTDIR (by default `/opt/pbench-agent`), The spec file uses this target (modifying DESTDIR) in order to build the RPM. The `agent/rpm/Makefile` provides `srpm` and `rpm` targets to build locally and `copr` and `copr-test` targets to build on COPR (you need an account on COPR and you need to set the env variable COPR_USER to the account name).

That's the good. The bad is that the server needs the same treatment and it has not had it yet. The ugly is the ad-hoc way of dealing with `pbench-config` and the python libraries.

### DIFF
--- a/agent/rpm/pbench-agent.spec
+++ b/agent/rpm/pbench-agent.spec
@@ -1,0 +1,127 @@
+Name:           pbench-agent
+Version:     0.60
+%define gdist gc9e89f8f
+Release:     1%{?gdist}%{!?gdist:}
+Summary:        The pbench harness
+
+License:        GPLv3+
+URL:            http://perf1.perf.lab.eng.bos.redhat.com/atheurer/pbench
+Source0:        pbench-agent-%{version}.tar.gz
+Buildarch:      noarch
+
+
+%if 0%{?rhel} == 6
+%define turbostatpkg cpupowerutils
+%else
+%define turbostatpkg kernel-tools
+%endif
+
+%if 0%{?rhel} == 7
+Requires:  scl-utils, rh-python36
+%endif
+
+%if 0%{?rhel} == 8 || 0%{?fedora} >= 29
+Requires:  python3-pip
+%endif
+
+%if 0%{?fedora} == 0
+# NOT fedora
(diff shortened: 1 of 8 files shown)


agent/rpm/pbench-agent.spec
```diff
@@ -0,0 +1,127 @@
...
+Summary:        The pbench harness
+
+License:        GPLv3+
+URL:            http://perf1.perf.lab.eng.bos.redhat.com/atheurer/pbench
+Source0:        pbench-agent-%{version}.tar.gz
+Buildarch:      noarch
+
+
+%if 0%{?rhel} == 6
+%define turbostatpkg cpupowerutils
```
portante, line 14: These are tool RPM dependencies, we need to remove them from the agent RPM dependency list.

agent/rpm/pbench-agent.spec
```diff
@@ -0,0 +1,127 @@
...
+# NOT fedora
+%define perljsonxs pbench-perl-JSON-XS
+%else
+%define perljsonxs perl-JSON-XS
+%endif
+
+
+Requires:  bzip2, tar, xz, screen
+Requires:  perl, perl-JSON, %{perljsonxs}, perl-Time-HiRes, perl-Data-UUID
+Requires:  net-tools, numactl, perf, psmisc, bc, sos, %{turbostatpkg}
```
portante, line 37: It seem `net-tools`, `numactl`, `perf`, and `%{turbostatpkg}` are for tools, which we need to remove from the agent dependency, but `sos`, `bc`, and `psmisc` are required for agent functioning.

We'll have to make sure the datalogs for those tools have the proper checks in them.

portante (changes requested):
We really should just merge this PR so that we can work through the building process iteratively allowing others to participate.

PR #1722, opened 2020-07-15
Tracking the life-cycle of tarballs

riya-17:
Resolves issue #1520 

It deals with Tracking the status of each Tarball. Whether the tarball has been processed successfully or failed and if it is failed then what led to its failure.


riya-17:
@portante we can add them to quarantine method but that will raise the concern regarding the reason for moving to quarantine. That needs to be gathered as well.

dbutenhof:
### First, some meta-comments

At the time this was last active, I was still figuring out all this stuff and didn't really have any comments, although there was something that tickled the back of my mind.

This has come back to me, however, as I start to think about how to manage user associations for tarballs. We need to track the authenticated user ID from intake (PUT to the Pbench server) through to indexing when we need to add that user ID to the ES documents. In theory we could do this with another companion file like how we manage the MD5, but we really want to get rid of all that and track state in an easily accessible and searchable DB and/or message stream and get rid of all the soft links and assorted file system cookie crumbs we scatter now.

So I thought about a tracking class, which it seemed overlapped uncomfortably with this... but with a different focus. And I worried about how to reconcile this.

And then while I was hiking, I realized that these are complementary, not conflicting. I think this directly reflects on the comments  Nick made in review back in September. I think that, in the end, this is not a "live state tracker" such as I'm envisioning, but rather a persistent permanent _audit log_ showing sequenced and timestamped records of every Pbench touch on each dataset.

That is, each audit log entry records a single action/decision made by Pbench regarding a tarball. It doesn't matter if there are multiple potential states, because each decision is indelibly recorded at the time its made. We can pull them together into a report if necessary, but we don't need to be able to do this "real time". (I think we'll want an Audit view in the dashboard where we can track the lifecycle of any tarball; e.g., "hey, I pushed a tarball and it's not there"; "so, the audit log shows that you pushed it a year ago, and looked at it once, you got expiration warnings starting 6 months ago, and 3 months ago we deleted it according to the published retention policy.")

I think the model coded here is a good start on building such an audit log.

### Specific comments

I agree with Pete that there are too many calls; scattered calls make it easy to miss things.

I think we want something like an `AuditLog` object; when we start any action on a tarball (intake, unpacking, "dispatch", indexing, perhaps even _access_ once we hide our unpacked files behind an API, so we can track "last access" as input to expiration algorithms), we create an `AuditLog` object, which will identify the operation we're starting, the specific tarball, and will get a timestamp. When we _complete_ the operation, we "finalize" that `AuditLog` object, with completion status; it'll add a new timestamp, and write the permanent audit log document.

I think it's TBD whether we actually write a pair of audit entries for start/finalize; there are advantages in tracking to writing a pair (e.g., if there's a catastrophic failure in the middle), but writing a single entry (probably under a `finally`) is more compact. Since this is fully encapsulated, the caller doesn't need to care whether we write one or two, or where/when/how. This reduces the audit log to two well-defined touch points for each operation. (Plus making it a lot easier to add to existing code paths.)

I also think we want a new ES index for the audit log; the focus here is different from the existing server report documents, and the fields should be different. We should design a new audit template that includes the fields we want to record.

riya-17:
I agree with this as well that there are too many scattered calls which kinda makes it really difficult for smooth collection of data. The idea about an AuditLog object is really intriguing and would love to discuss it further.

portante:
Future work tracked from issue #1520.

PR #1725, opened 2020-07-21
Pbench server API

FuqingWang:
Created a new API in Pbench server which serves as a middle ware for dashboard ElasticSeach calls.
The server APIs pulls data from ElasticSearch based on the arguments passed in from front end.
And the dashboard will in the future, call server APIs instead to get data from ElasticSearch.


portante:
Replaced by #1726.

PR #1726, opened 2020-07-22
Pbench Server API

FuqingWang:
Created a two APIs in Pbench server which serves as a middle ware for dashboard Elasticseach and GraphQL calls.
The server APIs pulls data from Elasticsearch based on the arguments passed in from front end requests.
And the dashboard will in the future, call server APIs instead to get data instead of making explicit queries.
Note: elasticsearch and graphql section needs to be specified in pbench-server.cfg

### DIFF
--- a/lib/pbench/server/api/__init__.py
+++ b/lib/pbench/server/api/__init__.py
@@ -1,9 +1,10 @@
 import os
 import sys
 import hashlib
+import requests
 
 from pathlib import Path
-from flask import request, jsonify, Flask
+from flask import request, jsonify, Flask, make_response
 from flask_restful import Resource, abort, Api
 from werkzeug.utils import secure_filename
 
@@ -28,6 +29,8 @@ def allowed_file(filename):
 def register_endpoints(api, app):
     api.add_resource(Upload, f"{app.config['REST_URI']}/upload")
     api.add_resource(HostInfo, f"{app.config['REST_URI']}/host_info")
+    api.add_resource(Elasticsearch, f"{app.config['REST_URI']}/elasticsearch")
+    api.add_resource(GraphQL, f"{app.config['REST_URI']}/graphql")
 
 
 def create_app():
@@ -52,8 +55,9 @@ def create_app():
     api = Api(app)
 
     app.logger = get_pbench_logger(__name__, config)
-
     app.config_server = config.conf["pbench-server"]
+    app.config_elasticsearch = config.conf["elasticsearch"]
+    app.config_graphql = config.conf["graphql"]
 
     prdp = app.config_server.get("pbench-receive-dir-prefix")
     try:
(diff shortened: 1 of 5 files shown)


lib/pbench/server/api/__init__.py
```diff
@@ -70,6 +72,50 @@ def create_app():
...
+class Download(Resource):
+    def post(self):
+        global app
+        try:
+            # query ElasticSearch
+            json_data = request.get_json()
+            if "payload" in json_data:
+                response = requests.post(json_data["url"], json=json_data["payload"])
+            else:
+                response = requests.get(json_data["url"])
```
portante, line 97: Can we make it so that the URL does not include the host name of the Elasticsearch instance, but instead have the server provide that, but the client just provides the path and arguments components?

lib/pbench/server/api/__init__.py
```diff
@@ -70,6 +72,55 @@ def create_app():
...
+
+def _corsify_actual_response(response):
+    response.headers.add("Access-Control-Allow-Origin", "*")
+    return response
+
+
+class Download(Resource):
+    def post(self):
+        global app
+        elasticsearch = "http://elasticsearch.perf.lab.eng.bos.redhat.com:9280"
```
portante, line 91: This needs to be read from the configuration file.  See https://github.com/distributed-system-analysis/pbench/blob/master/server/lib/config/pbench-server.cfg.example#L33.

portante:
@FuqingWang, any update on this work?

lib/pbench/server/api/__init__.py
```diff
@@ -70,6 +72,55 @@ def create_app():
...
+            if "payload" in json_data:
+                response = requests.post(url, json=json_data["payload"])
+            else:
+                response = requests.get(url)
+
+            # construct response object
+            response = make_response(response.text)
+            response = _corsify_actual_response(response)
+        except Exception:
+            app.logger.debug("There was something wrong with the POST request.")
```
npalaska, line 108: Hi, Is it possible if we can add more details why the post request failed, maybe we can log the response code like 500, or 401/403. That will help us debug later. This quickstart might help in writing exceptions https://requests.readthedocs.io/en/latest/user/quickstart/#errors-and-exceptions

FuqingWang:
Yes, updating later today.

lib/pbench/server/api/__init__.py
portante: Perhaps substitute `debug` with `exception`.

FuqingWang:
I have substituted `debug` using `exception` for the exception clauses. 
Also I have changed `Download` to `Elasticsearch` as resource name to have a more explicit meaning.

lib/pbench/server/api/__init__.py
```diff
@@ -75,6 +79,103 @@ def create_app():
...
+        try:
+            # query GraphQL
+            response = requests.post(graphQL, json=json_data)
+
+            # construct response object
+            response = make_response(response.text)
+            response = _corsify_actual_response(response)
+        except Exception:
+            app.logger.exception("There was something wrong with the POST request.")
+            abort(500, message="There was something wrong with your graphql request")
```
dbutenhof, line 114: Hmm; is this really an "internal error" or should we do better; for example, perhaps requesting a retry if the GraphQL post showed a connection error or timeout? (I know, decoding errors and figuring out what's recoverable and how can be pretty difficult and sometimes nondeterministic; but adding a fixed 500 error rarely makes anyone happy.)

lib/pbench/server/api/__init__.py
```diff
@@ -75,6 +79,103 @@ def create_app():
...
+        return response
+
+    # From the client side, an OPTIONS request is initiated before every POST to request CORS.
+    # On the server side, the server must specify the response headers to grant CORS accordingly.
+
+    def options(self):
+        try:
+            response = _build_cors_preflight_response()
+        except Exception:
+            app.logger.exception("There was something wrong with the OPTIONS request.")
```
dbutenhof, line 125: Can this be decoded better? Seems any exception here is likely to indicate an improbable Flask problem with make_response or with the created response object itself, not the client's request.

lib/pbench/server/api/__init__.py
```diff
@@ -75,6 +79,103 @@ def create_app():
...
+        return response
+
+
+class Elasticsearch(Resource):
+    def post(self):
+        global app
+        elasticsearch = app.config_elasticsearch.get("server")
+        json_data = request.get_json()
+        if not json_data:
+            app.logger.debug("Invalid json object")
```
dbutenhof, line 137: Maybe useful to include the request string that couldn't be JSON-ified? It's always possible the client data can't be reconstructed.

lib/pbench/server/api/__init__.py
```diff
@@ -75,6 +79,103 @@ def create_app():
...
+            abort(400, message="Invalid json object in request")
+
+        if not json_data["indices"]:
+            app.logger.debug("Missing path")
+            abort(400, message="Missing path in request")
```
dbutenhof, line 142: Here it seems we should be identifying the missing "indices" property name.

lib/pbench/server/api/__init__.py
```diff
@@ -75,6 +79,103 @@ def create_app():
...
+
+        try:
+            # query ElasticSearch
+            json_data = request.get_json()
```
dbutenhof, line 146: Why do this a second time?

lib/pbench/server/api/__init__.py
```diff
@@ -75,6 +79,103 @@ def create_app():
...
+                response = requests.post(url, json=json_data["payload"])
+            else:
+                response = requests.get(url)
+
+            # construct response object
+            response = make_response(response.text)
+            response = _corsify_actual_response(response)
+        except Exception:
+            app.logger.exception("There was something wrong with the POST request.")
+            abort(500, message="There was something wrong with your download request")
```
dbutenhof, line 162: Especially since we're constructing the URI here, and deciding whether to POST or GET, I think we really need more information here to be sure we understand what went wrong if something happens. Maybe at least that we "POST to url xyzzy" or "GET url xyzzy". And the Elasticsearch HTTP response would be great, too.

lib/pbench/server/api/__init__.py
```diff
@@ -75,6 +79,103 @@ def create_app():
...
+    # On the server side, the server must specify the response headers to grant CORS accordingly.
+
+    def options(self):
+        try:
+            response = _build_cors_preflight_response()
+        except Exception:
+            app.logger.exception("There was something wrong with the OPTIONS request.")
+            abort(400, message="There was something wrong with your request")
+        response.status_code = 200
+        return response
```
dbutenhof, line 176: This seems tedious repetition. I might even suggest a new small superclass to encapsulate common behavior between your two classes; that way future modifications (including improving error reports?) only need to be made in one place.

npalaska:
I can work on the remaining part of this PR I guess. I believe it just need some improvement in code structure. However, as we are thinking to move away from scp'ing' the tarball results from agent to server and instead using an HTTP post request to do the same, we need a corresponding api in server so the agent can perform the post request. Can that change go with this PR or should we create another PR? 
@portante / @ndokos / @dbutenhof

ndokos:
This PR is sitting on top of PR #1505 which does implement the HTTP POST request method on the server  to receive results from the agent. The final link I believe is the mods to the agent: `pbench-move-results` will need to ditch the current ssh/scp method and learn how to use the new HTTP POST method. The actual sending is done by the script `pbench-copy-result-tb` which is called from `pbench-move-results`, so it's only `pbench-copy-result-tb` that needs to be changed (to the best of my knowledge).

npalaska:
Ahh Thanks @ndokos I remember reviewing 1505 a while ago. The upload class is already serving as an endpoint for the http post request

portante:
Closing this PR in favor of #1769.

lib/pbench/server/api/__init__.py
npalaska: We are logging `error` level now along with he actual url with a request object so that should capture what couldn't JSON-ified

lib/pbench/server/api/__init__.py
npalaska: Good point, I can use this idea in the new PR to make logging more informative #1769

PR #1732, opened 2020-07-24
Fix tabs vs spaces in agent/base

portante:
While working on #1248, the tabs vs spaces issue in the file was adding extra changes in the diffs.  This PR pulls those changes out separately to reduce them.

This commit should only contain whitespace changes (tabs to spaces).

### DIFF
--- a/agent/base
+++ b/agent/base
@@ -1,5 +1,5 @@
 #!/bin/bash
-# -*- mode: shell-script; indent-tabs-mode: t; sh-basic-offset: 4; sh-indentation: 4; tab-width: 8 -*-
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; sh-indentation: 4; tab-width: 8 -*-
 
 # base: contains the base functions and environment variables needed to use
 #       the pbench-agent.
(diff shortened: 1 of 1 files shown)


ndokos:
Unit tests fail for unrelated reasons: the datalog unit test failure is fixed by #1733. The cull-tarballs one is not fixed yet.

portante:
> Why is `indent-tabs-mode` set to `t` at the top of the file? That's probably what contributed to this problem. I'll do a different PR and set it to nil.

I made this change, thanks.

agent/base
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-# -*- mode: shell-script; indent-tabs-mode: t; sh-basic-offset: 4; sh-indentation: 4; tab-width: 8 -*-
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; sh-indentation: 4; tab-width: 8 -*-
```
ndokos, line 2: That's going to work if the file is edited with emacs (which is all I was thinking about when I made the comment), but it is not going to save us from the vims of this world...

portante: There is a `vi` based string we could add at the bottom to enforce it the same way, but people use other editors besides `vi` and `emacs`.

So, yes, this is a short term fix, and since `base` is going away when we switch to all Python3 in the future, this change seems enough for now.

PR #1733, opened 2020-07-24
Fixes datalog-cpuacct directory order

riya-17:
This PR deals with fixing datalog-cpuacct directory sequencing by sorting directories. I have also checked all the scripts in datalog directory and they do not seem to suffer from this problem.


portante:
@riya-17, let's have you rebase this commit, now that #1734 is merged.  That should allow this PR to pass all Travis CI tests so that we can merge it normally.

PR #1734, opened 2020-07-24
Sort the output of pbench-cull-unpacked-tarballs.py

ndokos:
Two aspects:

- When unit testing, force the generator and sort the list before processing.

- For the report, always sort the actions (even when not unit testing). It keeps
  the code cleaner and it does not cost much.


portante:
Let's merge even though the tests failed.  The failure is known and will be fixed by #1733.

PR #1745, opened 2020-08-06
Rename pbench-agent.cfg.example to pbench-agent.cfg

ndokos:
That's enough to allow a pbench-agent RPM installation to do everything *except* send results to a server.


agent/util-scripts/unittests
```diff
@@ -216,7 +216,7 @@ function _setup_state {
     tst=${1}
     tool=${2}
 
-    sed "s;/opt/pbench-agent;${_testopt};" ${_testopt}/config/pbench-agent.cfg.example > ${_testopt}/config/pbench-agent.cfg
+    sed -i "s;/opt/pbench-agent;${_testopt};" ${_testopt}/config/pbench-agent.cfg
```
portante, line 219: As seen in the failing unit tests, how does the `pbench-agent.cfg` file get created in the `${_testopt}` installation directory?

ndokos: See https://github.com/distributed-system-analysis/pbench/blob/master/agent/util-scripts/unittests#L28 - I thought this should have copied it, but apparently it does not  - checking...

portante:
@ndokos, any reason not to merge this?

PR #1746, opened 2020-08-10
Clean up assorted issues with /etc/profile.d/pbench

RobertKrawitz:
* The current code clobbers the `prefix` and `pathins` function if the user happens to have them set.
* The current code doesn't check for an empty PYTHONPATH (or much less likely PATH) and hence incorrectly adds the (dynamic) current working directory to either if they're empty.
* The current call to path insertion is not quoted and will behave incorrectly if the prefix contains whitespace.

Checked with shellcheck and manually sourcing it twice to ensure that it's idempotent.

### DIFF
--- a/agent/profile
+++ b/agent/profile
@@ -1,35 +1,43 @@
 # -*- mode: shell-script -*-
-prefix=/opt/pbench-agent
 
-if [[ ! -f ${prefix}/profile ]]
-then
-    if [[ -f ${PWD}/profile ]]
-    then
-	prefix=$PWD
-    else
-	# last ditch attempt
-	prefix=$TOP
-    fi
-fi
-
-pathins() {
-    if [ -d "$1" ] && [[ ":$PATH:" != *":$1:"* ]]; then
-        PATH="$1:$PATH"
+function _pbench_pathins() {
+    if [[ -d "$1" ]] && [[ ":${PATH:-}:" != *":$1:"* ]] ; then
+        PATH="$1${PATH:+:$PATH}"
     fi
 }
 
-if [[ ! -f ${prefix}/util-scripts/pbench-register-tool-set ]]
-then
-    echo "*** WARNING *** Something wrong in the pbench base profile:" >&2
-    echo "                Setup has failed - could not determine the proper environment prefix: \"$prefix\"" >&2
-    echo "                You will need to fix it before working with pbench." >&2
-else
-    export PYTHONPATH=${prefix}/lib:${PYTHONPATH}
-    export _PBENCH_AGENT_CONFIG=${prefix}/config/pbench-agent.cfg
+function _pbench_setup() {
+    local prefix=/opt/pbench-agent
 
-    pathins ${prefix}/bench-scripts
-    pathins ${prefix}/util-scripts
-fi
+    if [[ ! -f ${prefix}/profile ]]
+    then
+	if [[ -f ${PWD}/profile ]]
+	then
+	    prefix=$PWD
+	else
+	    # last ditch attempt
+	    prefix=$TOP
+	fi
+    fi
 
-unset prefix
-unset -f pathins
+    if [[ ! -f ${prefix}/util-scripts/pbench-register-tool-set ]]
+    then
+	echo "*** WARNING *** Something wrong in the pbench base profile:" >&2
+	echo "                Setup has failed - could not determine the proper environment prefix: \"$prefix\"" >&2
+	echo "                You will need to fix it before working with pbench." >&2
+    else
+	local pdir=${prefix}/lib
+	if [[  ":${PYTHONPATH:-}:" != *":${pdir}:"* ]] ; then
+	    export PYTHONPATH=$pdir${PYTHONPATH:+:$PYTHONPATH}
+	fi
+	export _PBENCH_AGENT_CONFIG=${prefix}/config/pbench-agent.cfg
+
+	_pbench_pathins "${prefix}/bench-scripts"
+	_pbench_pathins "${prefix}/util-scripts"
+    fi
+}
+
+_pbench_setup
+
+unset -f _pbench_pathins
+unset -f _pbench_setup


agent/profile
```diff
@@ -1,35 +1,44 @@
...
+	echo "*** WARNING *** Something wrong in the pbench base profile:" >&2
+	echo "                Setup has failed - could not determine the proper environment prefix: \"$prefix\"" >&2
+	echo "                You will need to fix it before working with pbench." >&2
+    else
+	local pdir=${prefix}/lib
+	if [[ -z "$PYTHONPATH" && $PYTHONPATH != *":$pdir:"* && $PYTHONPATH != "$pdir"* && $PYTHONPATH != *"$pdir" ]] ; then
+	    export PYTHONPATH=${prefix}/lib${PYTHONPATH:+:$PYTHONPATH}
+	fi
+	export _PBENCH_AGENT_CONFIG=${prefix}/config/pbench-agent.cfg
+	export PBENCH_ROOT=$prefix
```
portante, line 34: Unless we are going to use `PBENCH_ROOT` other places in the code and add it to the documentation in this commit, exporting it doesn't seem like a good idea.  We already have the `pbench_install_dir` enviroment variable created by `/opt/pbench-agent/base`.

agent/profile
```diff
@@ -1,35 +1,44 @@
...
-    if [ -d "$1" ] && [[ ":$PATH:" != *":$1:"* ]]; then
-        PATH="$1:$PATH"
+    if [[ ! -f ${prefix}/util-scripts/pbench-register-tool-set ]]
+    then
+	echo "*** WARNING *** Something wrong in the pbench base profile:" >&2
+	echo "                Setup has failed - could not determine the proper environment prefix: \"$prefix\"" >&2
+	echo "                You will need to fix it before working with pbench." >&2
+    else
+	local pdir=${prefix}/lib
+	if [[ -z "$PYTHONPATH" && $PYTHONPATH != *":$pdir:"* && $PYTHONPATH != "$pdir"* && $PYTHONPATH != *"$pdir" ]] ; then
```
portante, line 30: Won't this only set `PYTHONPATH` if it does not already exist?

And, can't we consolidate this `if` line to: `if [[ ":${PYTHONPATH}:" != *":${qdir}:"* ]]; then`?

RobertKrawitz: Yeah, that logic is wrong.

agent/profile
RobertKrawitz: This is so external tools that need to know where pbench is installed can get at it.

agent/profile
```diff
@@ -1,35 +1,44 @@
 # -*- mode: shell-script -*-
-prefix=/opt/pbench-agent
 
-if [[ ! -f ${prefix}/profile ]]
-then
-    if [[ -f ${PWD}/profile ]]
+function _pbench_pathins() {
+    if [ -d "$1" ] && [[ $PATH != *":$1:"* && $PATH != "$1"* && $PATH != *"$1" ]] ; then
```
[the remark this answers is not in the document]
RobertKrawitz: On further thought, it does.  I'll revert this part.

agent/profile
```diff
@@ -1,35 +1,44 @@
 # -*- mode: shell-script -*-
-prefix=/opt/pbench-agent
 
-if [[ ! -f ${prefix}/profile ]]
-then
-    if [[ -f ${PWD}/profile ]]
+function _pbench_pathins() {
+    if [ -d "$1" ] && [[ ":$PATH:" != *"$1:"* ]] ; then
```
portante, line 4: Seems to be missing a leading `:`, see original.

agent/profile
```diff
@@ -1,35 +1,44 @@
...
-    if [ -d "$1" ] && [[ ":$PATH:" != *":$1:"* ]]; then
-        PATH="$1:$PATH"
+    if [[ ! -f ${prefix}/util-scripts/pbench-register-tool-set ]]
+    then
+	echo "*** WARNING *** Something wrong in the pbench base profile:" >&2
+	echo "                Setup has failed - could not determine the proper environment prefix: \"$prefix\"" >&2
+	echo "                You will need to fix it before working with pbench." >&2
+    else
+	local pdir=${prefix}/lib
+	if [[ -z "$PYTHONPATH" || ":$PYTHONPATH:" != *"$1:"* ]]
```
portante, line 30: Still missing a leading `:`, and it does not seem the `-z` check is necessary.  If `PYTHONPATH` is not set or empty, then the second check will also catch that allowing the `export` to happen.

agent/profile
portante: It is really worth having two ways to get the same information, given that it is well known that when multiple ways are given to get information invariably one method breaks over time?

We have been working to DRY our the code base, which is quite hard to do.  We should avoid adding to it if we can.

So perhaps we can consider documenting that external programs use a method like we find in [`base`]( https://github.com/distributed-system-analysis/pbench/blob/master/agent/base#L58), or change `base` to drop setting `pbench_install_dir` and instead expect it to be set via the `profile`.

Either way, for compatibility, let's use `pbench_install_dir` instead of `PBENCH_ROOT`.

RobertKrawitz:
Added a change so that an empty PYTHONPATH won't fail if `set -u` is in effect.

agent/profile
portante: Any thoughts on this?  I don't think we should be adding this variable in light of what is already in `base`.

agent/profile
```diff
@@ -1,35 +1,44 @@
 # -*- mode: shell-script -*-
-prefix=/opt/pbench-agent
 
-if [[ ! -f ${prefix}/profile ]]
-then
-    if [[ -f ${PWD}/profile ]]
+function _pbench_pathins() {
+    if [ -d "$1" ] && [[ ":${PATH:-}:" != *"$1:"* ]] ; then
```
portante, line 4: Seems to be missing a leading `:`.

portante:
@RobertKrawitz, where are we at with this PR?  Is this something that you still want to get into Pbench?

agent/profile
RobertKrawitz: I'd like to have a pointer to the pbench root visible outside, but if it's a real problem I can pull that one out.

RobertKrawitz:
> @RobertKrawitz, where are we at with this PR? Is this something that you still want to get into Pbench?

Yes I would; sorry, it dropped off my radar.  Pushing a fix for the missing colon.

portante (changes requested):
Seems okay, consider merging https://github.com/RobertKrawitz/pbench/pull/1 to finalize.

RobertKrawitz:
Done.  I was actually going to make some of those changes, but the old code had those constructs.

agent/profile
portante: Can you update this commit message and PR description to remove mention of the `PBENCH_ROOT` variable?  And open an issue for adding such an environment variable in the future?  I think it is a good idea, but I'd rather make the change along with the Python3 refactoring we are doing (see PR #1770, for example).

portante (changes requested):
Looks good!

Let's rebase this code, squash the commits to one (1), update the PR description and commit message to remove mention of `PBENCH_ROOT`, file an issue for that for future work with Python3 refactoring, and then we'll merge this.  Thanks!

portante (changes requested):
Almost, just need to update the commit message to remove references to `PBENCH_ROOT`.

RobertKrawitz:
Sorry about that, fixed now.

portante (dismissed):
Looks good.

agent/profile
portante: Tracking this work in #1794

agent/profile
```diff
@@ -1,35 +1,43 @@
...
-    if [ -d "$1" ] && [[ ":$PATH:" != *":$1:"* ]]; then
-        PATH="$1:$PATH"
+    if [[ ! -f ${prefix}/util-scripts/pbench-register-tool-set ]]
+    then
+	echo "*** WARNING *** Something wrong in the pbench base profile:" >&2
+	echo "                Setup has failed - could not determine the proper environment prefix: \"$prefix\"" >&2
+	echo "                You will need to fix it before working with pbench." >&2
+    else
+	local pdir=${prefix}/lib
+	if [[  ":${PYTHONPATH:-}:" != *"$pdir:"* ]] ; then
```
ndokos, line 30: Why is this different from the `PATH` test above? The PATH one has `:$1:` and this one has `$pdir:` with no initial colon. Also `${pdir}` with braces?

ndokos (changes requested):

agent/profile
RobertKrawitz: I thought I had fixed both of these...guess not.

ndokos (approved):

portante:
Not sure what happened, but opened up https://github.com/RobertKrawitz/pbench/pull/2.

portante (changes requested):
Can we take in another PR to finalize the cleanup?

See https://github.com/RobertKrawitz/pbench/pull/2.

portante (approved):

ndokos (approved):
LGTM

PR #1748, opened 2020-08-11
Require fio 3.21 and later

### LINKED ISSUE #1728: fio: Fedora RPM 3.19-3 does not install `fio-hist-log-pctiles.py`
Installing the `fio` RPM on Fedora 32 satisfies the dependency that `pbench-fio` is looking for, but there are failures in post-processing, because `fio-histo-log-pctiles.py` is *not* installed.

portante:
The `fio` 3.20 version now properly installs `fio-histo-log-pctiles.py` via `make install`.  The Fedora RPM build of `fio` for the 3.21-1 version is the first build to include that update.

Fixes #1728.

### DIFF
--- a/agent/bench-scripts/tests/pbench-fio/test-04.txt
+++ b/agent/bench-scripts/tests/pbench-fio/test-04.txt
@@ -446,8 +446,8 @@ fio job complete
 /var/tmp/pbench-test-bench/pbench-agent/tools-default/sar
 --- pbench tree state
 +++ pbench.log file contents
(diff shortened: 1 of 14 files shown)


portante:
We do not have any automation to build a container with the right packages installed to verify such behaviors.  This is one of the areas I hope we can fix soon.

This bug was found by folks installing the `fio` built by Fedora instead of the one we build, and seeing that it did not include the right files.  It was fixed upstream a while back, but the packaging updates as recorded in #1728 took some time.

PR #1752, opened 2020-08-13
Give default values to most of the variables.

ndokos:
Only `pbench_key_url` and `pbench_config_url` are left to be defined by the user (probably in the inventory file). The rest are defaulted but can be overridden.

Comment out the defaulted values in the example inventory file and add explanatory comments.

### DIFF
--- a/agent/ansible/Inventory/pbench-agent.hosts.example
+++ b/agent/ansible/Inventory/pbench-agent.hosts.example
@@ -12,20 +12,41 @@ host3       ansible_python_interpreter=/usr/bin/python3
 # This section contains site-specific variables. An admin should
 # modify these settings and provide the resulting file as a site-specific
 # example. Once that is done, users should *NOT* have to modify this section at all.
+
+# All but two of the variables below are given default values that should serve most
+# users, but they can be overridden explicitly by uncommenting them and giving them
+# appropriate values.
+
+# The two variables pbench_key_url and pbench_config_url are *not* defaulted: they
+# are specific to a particular environment and they *have* to be defined (here or
+# elsewhere) by the user.
+
 [servers:vars]
-pbench_repo_url_prefix = https://copr-be.cloud.fedoraproject.org/results/<EXAMPLE_USER>
+# We assume that the COPR convention of repo naming is followed. Obviously if the repo
+# is in COPR that assumption holds, but if you decide to use a different repo, you have
+# to follow the convention (or change the pbench-agent-install role).
+# The convention is that the repo name consists of the prefix defined here (or in the
+# defaults of the pbench-agent-install role), followed by a distro designation:
+#     fedora-NN   for Fedora distros
+#     epel-7      for RHEL7 and CentOS 7 distros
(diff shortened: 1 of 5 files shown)


agent/ansible/Inventory/pbench-agent.hosts.example
```diff
@@ -12,20 +12,26 @@ host3       ansible_python_interpreter=/usr/bin/python3
 # This section contains site-specific variables. An admin should
 # modify these settings and provide the resulting file as a site-specific
 # example. Once that is done, users should *NOT* have to modify this section at all.
+
+# All but two of the variables below are given default values that should server most
```
portante, line 16: s/server/serve/

agent/ansible/roles/pbench-agent-config/tasks/main.yml
```diff
@@ -1,5 +1,17 @@
 ---
 # pbench agent configuration
+- debug:
```
portante, line 3: Do these `debug` tasks run only when some verbosity level of ansible is requested?

agent/ansible/roles/pbench-repo-install/defaults/main.yml
```diff
@@ -1,4 +1,6 @@
 ---
+pbench_repo_url_prefix: https://copr-be.cloud.fedoraproject.org/results/ndokos
```
portante, line 2: Perhaps make the user in this URL a variable defaulting to `ndokos` so that it can be conveniently changed?

agent/ansible/roles/pbench-agent-config/tasks/main.yml
ndokos: Yes, it requires `-v` - else they are skipped.

agent/ansible/roles/pbench-repo-install/defaults/main.yml
ndokos: I debated this with myself (I'm not sure who won though...) I'd rather treat this as an atomic piece of information: it can be overridden as a whole in the inventory file, but  the default value relieves most people from  having to define it at all. So the onus is put on people who use their own repo (COPR or otherwise) to define it approprirately. WDYT?

portante: Perhaps consider:

```
fedoraproject_username: ndokos
pbench_repo_url_prefix: https://copr-be.cloud.fedoraproject.org/results/{{ fedoraproject_username }}
```

That way they can override the `pbench_repo_url_prefix`, or if they are using Fedora Project themselves, they can just change the user name used.

ndokos: Sounds good - I'll push a change.

ndokos:
If this last commit looks OK, I'll squash/rebase for final review and merging.

agent/ansible/Inventory/pbench-agent.hosts.example
ndokos: Fixed.

PR #1768, opened 2020-08-31
Lock versions of flake8 and black to known-good

dbutenhof:
Simple change to preclude unit testing from using the new 20.8b1 version of black, which introduces reformatting differences from 19.10b0 against which we've set expectations. I'm locking flake8 to the current version for the same reason. TBD whether we should try to lock other tools as well, but this should suffice for right now.


.gitignore
```diff
@@ -18,3 +18,5 @@ ChangeLog
 .pytest_cache/
 .venv
 .vscode
+Pipfile
+Pipfile.lock
```
npalaska, line 22: Minor: We can probably add a new line character at the end of gitignore

PR #1769, opened 2020-09-01
Pbench Server APIs for ES and GraphQL

npalaska:
This work is an extension of Fuqing's PR https://github.com/distributed-system-analysis/pbench/pull/1726
Description from Original PR:

> Created a two APIs in Pbench server which serves as a middle ware for dashboard Elasticseach and GraphQL calls.
> The server APIs pulls data from Elasticsearch based on the arguments passed in from front end requests.
> And the dashboard will in the future, call server APIs instead to get data instead of making explicit queries.
> Note: elasticsearch and graphql section needs to be specified in pbench-server.cfg with host and port info

### DIFF
--- a/lib/pbench/server/api/__init__.py
+++ b/lib/pbench/server/api/__init__.py
@@ -1,9 +1,15 @@
+"""
+Pbench Server API module -
+Provides middleware for remote clients, either the associated Pbench
+Dashboard or any number of pbench agent users.
+"""
+
 import os
 import sys
 import hashlib
+from pathlib import Path
 import requests
 
-from pathlib import Path
 from flask import request, jsonify, Flask, make_response
 from flask_restful import Resource, abort, Api
 from werkzeug.utils import secure_filename
@@ -18,6 +24,7 @@ app = None
 
 
 def allowed_file(filename):
+    """Check if the file has the correct extension."""
     try:
         fn = filename.rsplit(".", 1)[1].lower()
     except IndexError:
@@ -27,6 +34,9 @@ def allowed_file(filename):
 
 
 def register_endpoints(api, app):
+    """Register flask endpoints with the corresponding resource classes
+    to make the APIs active."""
+
     api.add_resource(Upload, f"{app.config['REST_URI']}/upload")
     api.add_resource(HostInfo, f"{app.config['REST_URI']}/host_info")
     api.add_resource(Elasticsearch, f"{app.config['REST_URI']}/elasticsearch")
@@ -34,6 +44,8 @@ def register_endpoints(api, app):
 
 
 def create_app():
+    """Create Flask app with defined resource endpoints."""
+
     global app
 
     cfg_name = os.environ.get("_PBENCH_SERVER_CONFIG")
@@ -60,10 +72,18 @@ def create_app():
     app.config_graphql = config.conf["graphql"]
 
     prdp = app.config_server.get("pbench-receive-dir-prefix")
+    if not prdp:
+        app.logger.error("Missing config variable for pbench-receive-dir-prefix")
+        sys.exit(1)
     try:
         upload_directory = Path(f"{prdp}-002").resolve(strict=True)
+    except FileNotFoundError:
+        app.logger.exception("pbench-receive-dir-prefix does not exist on the host")
+        sys.exit(1)
     except Exception:
-        app.logger.error("Missing config variable for pbench-receive-dir-prefix")
+        app.logger.exception(
+            "Exception occurred during setting up the upload directory on the host"
+        )
         sys.exit(1)
     else:
         app.upload_directory = upload_directory
@@ -81,7 +101,7 @@ def create_app():
 
 def _build_cors_preflight_response():
     response = make_response()
-    response.headers.add("Access-Control-Allow-Origin", "*")
+    response = _corsify_actual_response(response)
     response.headers.add("Access-Control-Allow-Headers", "*")
     response.headers.add("Access-Control-Allow-Methods", "*")
     return response
@@ -92,15 +112,39 @@ def _corsify_actual_response(response):
     return response
 
 
-class GraphQL(Resource):
+class CorsOptionsRequest:
+    """
+    From the client side, an OPTIONS request is initiated before every POST to request CORS.
+    On the server side, the server must specify the response headers to grant CORS accordingly.
+    """
+
+    def __init__(self, options_exception_msg):
+        self.options_exception_msg = options_exception_msg
+
+    def options(self):
+        try:
+            response = _build_cors_preflight_response()
+        except Exception:
+            app.logger.exception(self.options_exception_msg)
+            abort(400, message=self.options_exception_msg)
+        response.status_code = 200
+        return response
+
+
+class GraphQL(Resource, CorsOptionsRequest):
+    """GraphQL API for post request via server."""
+
+    def __init__(self):
+        CorsOptionsRequest.__init__(self, "Bad options request for the GraphQL query")
+
     def post(self):
         global app
-        graphQL = app.config_graphql.get("server")
-        json_data = request.get_json()
+        graphQL = f"{app.config_graphql.get('host')}:{app.config_graphql.get('port')}"
+        json_data = request.get_json(silent=True)
 
         if not json_data:
-            app.logger.debug("Invalid query object")
-            abort(400, message="Invalid query object in request")
+            app.logger.error("GraphQL: Invalid json object %s", request.url)
+            abort(400, message="GraphQL: Invalid json object in request")
 
         try:
             # query GraphQL
@@ -109,69 +153,98 @@ class GraphQL(Resource):
             # construct response object
             response = make_response(response.text)
             response = _corsify_actual_response(response)
+
+        except requests.exceptions.ConnectionError:
+            app.logger.exception("Connection refused during the GraphQL post request")
+            abort(500, message="Network problem, could not post to GraphQL Endpoint")
+        except requests.exceptions.Timeout:
+            app.logger.exception("Connection timed out during the GraphQL post request")
+            abort(
+                500, message="Connection timed out, could not post to GraphQL Endpoint"
+            )
+        except requests.exceptions.InvalidURL:
+            app.logger.exception("Invalid url during the GraphQL post request")
+            abort(
+                500, message="Invalid GraphQL url, could not complete the post request"
+            )
         except Exception:
-            app.logger.exception("There was something wrong with the POST request.")
-            abort(500, message="There was something wrong with your graphql request")
+            app.logger.exception("Exception occurred during the GraphQL post request")
+            abort(500, message="Could not post to GraphQL endpoint")
+
         response.status_code = 200
         return response
 
-    # From the client side, an OPTIONS request is initiated before every POST to request CORS.
-    # On the server side, the server must specify the response headers to grant CORS accordingly.
 
-    def options(self):
-        try:
-            response = _build_cors_preflight_response()
-        except Exception:
-            app.logger.exception("There was something wrong with the OPTIONS request.")
-            abort(400, message="There was something wrong with your request")
-        response.status_code = 200
-        return response
+class Elasticsearch(Resource, CorsOptionsRequest):
+    """Elasticsearch API for post request via server."""
 
+    def __init__(self):
+        CorsOptionsRequest.__init__(
+            self, "Bad options request for the Elasticsearch query"
+        )
 
-class Elasticsearch(Resource):
     def post(self):
         global app
-        elasticsearch = app.config_elasticsearch.get("server")
-        json_data = request.get_json()
+        elasticsearch = f"{app.config_elasticsearch.get('host')}:{app.config_elasticsearch.get('port')}"
+        json_data = request.get_json(silent=True)
         if not json_data:
-            app.logger.debug("Invalid json object")
-            abort(400, message="Invalid json object in request")
+            app.logger.error(
+                "Elasticsearch: Invalid json object. Query: %s", request.url
+            )
+            abort(400, message="Elasticsearch: Invalid json object in request")
 
         if not json_data["indices"]:
-            app.logger.debug("Missing path")
-            abort(400, message="Missing path in request")
+            app.logger.error("Elasticsearch: Missing indices path in the post request")
+            abort(400, message="Missing indices path in the Elasticsearch request")
 
         try:
-            # query ElasticSearch
-            json_data = request.get_json()
+            # query Elasticsearch
             if "params" in json_data:
-                url = f"{elasticsearch}{json_data['indices']}?{json_data['params']}"
+                url = f"{elasticsearch}/{json_data['indices']}?{json_data['params']}"
             else:
-                url = f"{elasticsearch}{json_data['indices']}"
+                url = f"{elasticsearch}/{json_data['indices']}"
 
             if "payload" in json_data:
                 response = requests.post(url, json=json_data["payload"])
             else:
+                app.logger.info(
+                    "No payload found in Elasticsearch post request json data"
+                )
                 response = requests.get(url)
 
             # construct response object
             response = make_response(response.text)
             response = _corsify_actual_response(response)
-        except Exception:
-            app.logger.exception("There was something wrong with the POST request.")
-            abort(500, message="There was something wrong with your download request")
-        response.status_code = 200
-        return response
 
-    # From the client side, an OPTIONS request is initiated before every POST to request CORS.
-    # On the server side, the server must specify the response headers to grant CORS accordingly.
-
-    def options(self):
-        try:
-            response = _build_cors_preflight_response()
+        except requests.exceptions.ConnectionError:
+            app.logger.exception(
+                "Connection refused during the Elasticsearch post request"
+            )
+            abort(
+                500, message="Network problem, could not post to Elasticsearch Endpoint"
+            )
+        except requests.exceptions.Timeout:
+            app.logger.exception(
+                "Connection timed out during the Elasticsearch post request"
+            )
+            abort(
+                500,
+                message="Connection timed out, could not post to Elasticsearch Endpoint",
+            )
+        except requests.exceptions.InvalidURL:
+            app.logger.exception("Invalid url during the Elasticsearch post request")
+            abort(
+                500,
+                message="Invalid Elasticsearch url, could not complete the post request",
+            )
         except Exception:
-            app.logger.exception("There was something wrong with the OPTIONS request.")
-            abort(400, message="There was something wrong with your request")
+            app.logger.exception(
+                "Exception occurred during the Elasticsearch post request"
+            )
+            abort(
+                500, message="Could not post to Elasticsearch endpoint",
+            )
+
         response.status_code = 200
         return response
 
@@ -187,9 +260,7 @@ class HostInfo(Resource):
                 )
             )
         except Exception:
-            app.logger.exception(
-                "There was something wrong constructing the host info."
-            )
+            app.logger.exception("There was something wrong constructing the host info")
             abort(500, message="There was something wrong with your request")
         response.status_code = 200
         return response
--- a/lib/pbench/server/api/__init__.py
+++ b/lib/pbench/server/api/__init__.py
@@ -1,9 +1,10 @@
 import os
 import sys
 import hashlib
+import requests
 
 from pathlib import Path
-from flask import request, jsonify, Flask
+from flask import request, jsonify, Flask, make_response
 from flask_restful import Resource, abort, Api
 from werkzeug.utils import secure_filename
 
@@ -28,6 +29,8 @@ def allowed_file(filename):
 def register_endpoints(api, app):
     api.add_resource(Upload, f"{app.config['REST_URI']}/upload")
     api.add_resource(HostInfo, f"{app.config['REST_URI']}/host_info")
+    api.add_resource(Elasticsearch, f"{app.config['REST_URI']}/elasticsearch")
+    api.add_resource(GraphQL, f"{app.config['REST_URI']}/graphql")
 
 
 def create_app():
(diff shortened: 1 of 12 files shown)


npalaska:
@portante the py3-server tests are passing locally but for some reason the lint command is failing with this error 
```ERROR: InvocationError for command /home/travis/build/distributed-system-analysis/pbench/.tox/lint/bin/black --check . (exited with code 1)```
I am not sure but could this be related to https://github.com/distributed-system-analysis/pbench/pull/1768 The PR deals with the version of the black which is where the problem seems to be. Any suggestions?

**Comment by dbutenhof** [comment]
I submitted; so you should be able to merge my tox.ini changes and clear the black error.

File: lib/pbench/server/api/__init__.py [hunk 1]
```diff
@@ -92,15 +92,36 @@ def _corsify_actual_response(response):
...
+
     def post(self):
         global app
         graphQL = app.config_graphql.get("server")
         json_data = request.get_json()
 
         if not json_data:
-            app.logger.debug("Invalid query object")
-            abort(400, message="Invalid query object in request")
+            app.logger.error("GraphQL: Invalid json object \n "
```
dbutenhof, line 122: Do we really want to be logging with explicit newlines? This can make a text log file harder to skim and to process with tools. I don't recall seeing this pattern elsewhere yet, though I may not have noticed. And if you were dumping a JSON object, I could see some value (or at least inevitability), but it seems unnecessary for a URI.

lib/pbench/server/api/__init__.py
```diff
@@ -92,15 +92,36 @@ def _corsify_actual_response(response):
     return response
 
 
-class GraphQL(Resource):
+class CorsOptionsRequest:
+    def __init__(self, options_exception_msg):
```
dbutenhof, line 96: The way your classes are set up and named, the error message could have just used the class name in a template string without the extra parameter, but it doesn't really hurt.

lib/pbench/server/api/__init__.py
```diff
@@ -92,15 +92,36 @@ def _corsify_actual_response(response):
...
+        except Exception as e:
+            app.logger.exception(self.options_exception_msg + " Raised: %s", e.__class__)
+            abort(400, message=self.options_exception_msg)
+        response.status_code = 200
+        return response
+
+
+class GraphQL(Resource, CorsOptionsRequest):
+    def __init__(self):
+        super(GraphQL, self).__init__("Bad options request for the Graphql query")
```
dbutenhof, line 114: Probably should use "GraphQL" instead of "Graphql".

lib/pbench/server/api/__init__.py
```diff
@@ -109,41 +130,40 @@ def post(self):
...
-            app.logger.exception("There was something wrong with the OPTIONS request.")
-            abort(400, message="There was something wrong with your request")
         response.status_code = 200
         return response
 
 
-class Elasticsearch(Resource):
+class Elasticsearch(Resource, CorsOptionsRequest):
+    def __init__(self):
+        super(Elasticsearch, self).__init__("Bad options request for the ElasticSearch query")
```
dbutenhof, line 153: Elasticsearch is not PascalCase, just Capitalized!

portante:
FWIW, we refer to "Pbench" itself with just a capital "P", lowercase "bench".

lib/pbench/server/api/__init__.py
npalaska: make sense, but flask request.url does contain the actual json query https://flask.palletsprojects.com/en/1.1.x/api/#incoming-request-data

dbutenhof: Well, you could log the `base_url` to keep it short; but according to that doc page, `uri` isn't JSON payload, just the full user URL including the base URI and any query string, so it shouldn't be that bad.

dbutenhof:
Looks like you have *real* `black` formatting errors:

```
would reformat /home/travis/build/distributed-system-analysis/pbench/lib/pbench/server/api/__init__.py
would reformat /home/travis/build/distributed-system-analysis/pbench/lib/pbench/test/unit/server/test_requests.py
```

I guess you need to run those files through black and push them out.

dbutenhof (dismissed):
Black is happy, I'm happy. Seems like Pete's request is resolved, too.

lib/pbench/server/api/__init__.py
```diff
@@ -1,9 +1,16 @@
+"""
+Simple server APIs for redirecting other services call via Pbench server
```
portante, line 2: Trailing periods on sentences?

lib/pbench/server/api/__init__.py
```diff
@@ -26,11 +34,18 @@ def allowed_file(filename):
...
+    to make the APIs active"""
+
     api.add_resource(Upload, f"{app.config['REST_URI']}/upload")
     api.add_resource(HostInfo, f"{app.config['REST_URI']}/host_info")
+    api.add_resource(Elasticsearch, f"{app.config['REST_URI']}/elasticsearch")
+    api.add_resource(GraphQL, f"{app.config['REST_URI']}/graphql")
 
 
 def create_app():
+    """ Create Flask app with defined resource endpoints"""
```
portante, line 47: Nit: leading space ...

lib/pbench/server/api/__init__.py
```diff
@@ -52,14 +67,17 @@ def create_app():
...
+    app.config_graphql = config.conf["graphql"]
 
     prdp = app.config_server.get("pbench-receive-dir-prefix")
     try:
         upload_directory = Path(f"{prdp}-002").resolve(strict=True)
-    except Exception:
-        app.logger.error("Missing config variable for pbench-receive-dir-prefix")
+    except Exception as e:
+        app.logger.error(
+            "Missing config variable for pbench-receive-dir-prefix, %s", e.__class__
```
portante, line 79: Seems like we should check if `prdp` received an empty value from the configuration, and emit the missing configuration variable log message without the exception class.

Then, we could resolve the `Path`, catching `FileNotFoundError` if the path is missing, and then using `app.logger.exception()` for any other exception raised.

Thoughts?

lib/pbench/server/api/__init__.py
```diff
@@ -75,6 +93,163 @@ def create_app():
...
+        response.status_code = 200
+        return response
+
+
+class GraphQL(Resource, CorsOptionsRequest):
+    """ GraphQL API for post request via server """
+
+    def __init__(self):
+        CorsOptionsRequest.__init__(
+            self, "Bad options request for the Elasticsearch query"
```
portante, line 135: Is this `Elasticsearch` or `GraphQL`?

lib/pbench/server/api/__init__.py
```diff
@@ -75,6 +93,163 @@ def create_app():
...
+                "Elasticsearch: Invalid json object \n" "Query: %s", request.url
+            )
+            abort(400, message="Elasticsearch: Invalid json object in request")
+
+        if not json_data["indices"]:
+            app.logger.error("Elasticsearch: Missing indices path in the post request")
+            abort(400, message="Missing indices path in the Elasticsearch request")
+
+        try:
+            # query ElasticSearch
```
portante, line 203: s/ElasticSearch/Elasticsearch

lib/pbench/server/api/__init__.py
```diff
@@ -75,6 +93,163 @@ def create_app():
...
+            # construct response object
+            response = make_response(response.text)
+            response = _corsify_actual_response(response)
+
+        except requests.exceptions.ConnectionError:
+            app.logger.exception(
+                "Connection refused during the Elasticsearch post request"
+            )
+            abort(
+                500, message="Network problem, could not post to elasticSearch Endpoint"
```
portante, line 223: s/elasticSearch/Elasticsearch/

lib/pbench/test/unit/config/pbench-server.cfg
```diff
@@ -4,6 +4,12 @@ install-dir = ./opt/pbench-server
 [pbench-server]
 pbench-top-dir = ./srv/pbench
 
+[elasticsearch]
+server = http://elasticsearch.perf.lab.eng.bos.redhat.com:9280
```
portante, line 8: The test environment should not be referencing internal instances.  Let's remove these and find another way to do this.

lib/pbench/test/unit/server/conftest.py
```diff
@@ -12,6 +12,12 @@
 [pbench-server]
 pbench-top-dir = {TMP}/srv/pbench
 
+[elasticsearch]
+server = http://elasticsearch.perf.lab.eng.bos.redhat.com:9280
```
portante, line 16: The test environment should not be referencing internal instances.  Let's remove these and find another way to do this.

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -19,6 +19,48 @@ def test_host_info(client, pytestconfig):
         assert response.json.get("message") == expected_message
 
 
+class TestElasticSearch:
```
portante, line 22: s/ElasticSearch/Elasticsearch

portante (changes requested):
Looks good, but we can't reference the internal hosts in public code.

lib/pbench/server/api/__init__.py
npalaska: oops, good catch

lib/pbench/test/unit/config/pbench-server.cfg
npalaska: I agree, I will look into this

lib/pbench/server/api/__init__.py
npalaska: Indeed, the exception can be narrowed down before we can call the broader Exception class

server/lib/config/pbench-server-default.cfg
```diff
@@ -126,6 +126,12 @@ lowerbound = 820
 # index_prefix =
 # bulk_action_count =
```
npalaska, line 128: It is mentioned on line 123 that `# NOTE: No defaults are provided for the "Indexing" section deliberately.` what is the reason that we don't want to include the dummy elastic endpoint in default config file that can be overwritten from the main config file? cc: @portante @ndokos 

I did this to remove the reference to actual elastic/graphql instance for unit testing, so we can just check the api workflow without actually hitting the endpoints. If this is fine I can actually reuse the Indexing config parameter instead of separate elasticsearch variable.
Any thought?

lib/pbench/server/api/__init__.py
```diff
@@ -1,9 +1,16 @@
+"""
+Simple server APIs for redirecting other services call via Pbench server.
+This serves as a middleware for dashboard Elasticseach and GraphQL calls
+as well as Pbench agent move results call.
+"""
```
portante, line 5: Could we consider a rewording of this a bit?

```
"""Pbench Server API module - Provides middleware for remote clients, either the associated Pbench Dashboard or any number of pbench agent users.
```

Seems like as a module documentation string, we should avoid listing which APIs are provided since they are likely to change.

lib/pbench/server/api/__init__.py
```diff
@@ -75,6 +99,158 @@ def create_app():
     return app
 
 
+def _build_cors_preflight_response():
+    response = make_response()
+    response.headers.add("Access-Control-Allow-Origin", "*")
```
portante, line 104: Perhaps call `_corsify_actual_response(response)` here?

server/lib/config/pbench-server-default.cfg
```diff
@@ -126,6 +126,12 @@ lowerbound = 820
...
 
+[elasticsearch]
+server = elasticsearch.example.com
```
portante, line 130: For Elasticsearch we really need `host` and `port`.  Can we remove the `server`, `host`, and `port` options from the `Indexing` section, and move them to the `elasticsearch` section as `host` and `port` only.  Then fix up the `indexer.py` to reference both sections?

portante (changes requested):
Good progress!

server/lib/config/pbench-server-default.cfg
npalaska: So you mean something like this 
``` 
[Indexing]
index_prefix =
bulk_action_count =

[elasticsearch]
host = 
port = 
```
and then referencing these 2 endpoints in indexer.py's `_get_es_hosts` to get the ES host?

npalaska: It actually make sense if we have 2 separate sections in our server config file `[Indexing]` and `[elasticsearch]` where elasticsearch section holds host and port specific information and `Indexing` section holds index specific information such as index_prefix and all. However, it means we might have to update the pbench-server.cfg file

lib/pbench/server/indexer.py
```diff
@@ -539,22 +539,34 @@ def _get_es_hosts(config, logger):
...
         logger.warning(
-            "Need an [Indexing] section with host and port defined"
-            " in {} configuration file",
+            "Do not have [elasticsearch] section with host and port defined"
+            " in {} configuration file, checking the Indexing section",
             " ".join(config.files),
         )
-        return None
+        try:
+            URL = config.get("Indexing", "server")
```
portante, line 551: Seems like we can just drop the reference to the `server` option of the `Indexing` section entirely.

server/lib/config/pbench-server-default.cfg
portante: > what is the reason that we don't want to include the dummy elastic endpoint in the default config file that can be overwritten from the main config file?

Mostly because we want to avoid providing default for services that have to be explicitly setup.

> I did this to remove the reference to actual elastic/graphql instance of unit testing ...

I think you mean that you override the configuration file in the unit tests, right?

The concern here is to make sure we try talking to services for Elasticsearch or GraphQL which don't actually exist.  DNS lookups for what we consider to be non-existent host names could actually resolve in some environments.

Let's also drop the reference in the comments to the `server` options in the `Indexing` section.  Thanks!

lib/pbench/server/api/__init__.py
```diff
@@ -85,11 +261,12 @@ def get(self):
...
             )
-        except Exception:
+        except Exception as exc:
             app.logger.exception(
-                "There was something wrong constructing the host info."
+                "There was something wrong constructing the host info. %s",
+                exc.__class__,
             )
-            abort(500, message="There was something wrong with your request")
+            abort(400, message="There was something wrong with your request")
```
portante, line 269: Why this change from a `500` to a `400` error?  A `400` error code typically means the server is functioning fine, but encountered some problem with the request payload itself.  A `500` error typically means that the server failed to handle the request properly.

In this case `jsonify` should not raise any errors.  If it does, it indicates something is wrong with the server operation, not with their request payload.

portante (changes requested):
Looking good, just a bit more ...

portante:
@npalaska, can you run through the unresolved comments in #1726 to make sure they are addressed in this PR, and resolve them with a note confirming they have been fixed here?  Thanks!

npalaska:
@portante I have mostly addressed it but I will recheck just to make sure, thanks

lib/pbench/server/api/__init__.py
npalaska: Ahh, make sense

lib/pbench/server/api/__init__.py
```diff
@@ -211,6 +211,9 @@ def post(self):
             if "payload" in json_data:
                 response = requests.post(url, json=json_data["payload"])
             else:
+                app.logger.info(
+                    "No playload found in elasricsearch post request json data"
```
dbutenhof, line 215: "elasricsearch" -> "Elasticsearch" ;-)

lib/pbench/server/api/__init__.py
```diff
@@ -212,7 +212,7 @@ def post(self):
                 response = requests.post(url, json=json_data["payload"])
             else:
                 app.logger.info(
-                    "No playload found in elasricsearch post request json data"
+                    "No playload found in elasticsearch post request json data"
```
dbutenhof, line 215: Technically I think it should still be capitalized; however since this is a log rather than an external message, it's not critical.

dbutenhof (approved):

lib/pbench/server/api/__init__.py
```diff
@@ -75,6 +99,160 @@ def create_app():
...
+
+    def __init__(self, options_exception_msg):
+        self.options_exception_msg = options_exception_msg
+
+    def options(self):
+        try:
+            response = _build_cors_preflight_response()
+        except Exception as e:
+            app.logger.exception(
+                self.options_exception_msg + " Raised: %s", e.__class__
```
portante, line 129: No need reach into the exception class to get the name, just use `self.options_exception_msg` here.

lib/pbench/server/api/__init__.py
portante: And for consistency with the message at line 202.

lib/pbench/server/api/__init__.py
```diff
@@ -85,9 +263,10 @@ def get(self):
                     f":{app.config_server.get('pbench-receive-dir-prefix')}-002"
                 )
             )
-        except Exception:
+        except Exception as exc:
             app.logger.exception(
-                "There was something wrong constructing the host info."
+                "There was something wrong constructing the host info. %s",
+                exc.__class__,
```
portante, line 269: No need to use exception's class here.

lib/pbench/server/indexer.py
```diff
@@ -539,22 +539,26 @@ def _get_es_hosts(config, logger):
...
+            "Do not have [elasticsearch] section with host and port defined"
+            " in {} configuration file.",
             " ".join(config.files),
         )
         return None
     except NoOptionError:
         try:
-            host = config.get("Indexing", "host")
-            port = config.get("Indexing", "port")
+            URL = config.get("elasticsearch", "server")
```
portante, line 553: We no longer need this backwards compatibility step.

portante (changes requested):
Almost, just a couple of things to resolve.

dbutenhof (approved):
Looks pretty good.

lib/pbench/server/indexer.py
```diff
@@ -539,22 +539,15 @@ def _get_es_hosts(config, logger):
...
     try:
-        URL = config.get("Indexing", "server")
-    except NoSectionError:
+        host = config.get("elasticsearch", "host")
+        port = config.get("elasticsearch", "port")
+    except (NoSectionError, NoOptionError):
         logger.warning(
-            "Need an [Indexing] section with host and port defined"
-            " in {} configuration file",
+            "Do not have [elasticsearch] section with host and port defined"
```
portante, line 546: Perhaps consider, "Failed to find an [elasticsearch] section ..."

portante (changes requested):
One small change to consider...

dbutenhof (approved):

lib/pbench/server/api/__init__.py
portante: This particular comment thread seems to have been resolved.

lib/pbench/server/api/__init__.py
```diff
@@ -75,6 +99,156 @@ def create_app():
...
+class GraphQL(Resource, CorsOptionsRequest):
+    """GraphQL API for post request via server."""
+
+    def __init__(self):
+        CorsOptionsRequest.__init__(self, "Bad options request for the GraphQL query")
+
+    def post(self):
+        global app
+        graphQL = f"{app.config_graphql.get('host')}:{app.config_graphql.get('port')}"
+        json_data = request.get_json()
```
portante, line 143: According to the `flask` [documentation](https://flask.palletsprojects.com/en/1.1.x/api/#flask.Request.on_json_loading_failed), a `BadRequest` error will be raised if the JSON fails to parse.  Don't we have to catch that error and handle it properly?

lib/pbench/server/api/__init__.py
```diff
@@ -75,6 +99,156 @@ def create_app():
...
+
+    def __init__(self):
+        CorsOptionsRequest.__init__(
+            self, "Bad options request for the Elasticsearch query"
+        )
+
+    def post(self):
+        global app
+        elasticsearch = f"{app.config_elasticsearch.get('host')}:{app.config_elasticsearch.get('port')}"
+        json_data = request.get_json()
```
portante, line 189: See other comment above about handling `BadRequest` exceptions here...

portante (changes requested):
Looks good, which one exception (pun intended): we need to consider how to handle `BadRequest` exceptions from JSON parsing in `flask.request.get_json()`.

riya-17 (approved):

lib/pbench/server/api/__init__.py
dbutenhof: Maybe there should be some `flask` api annotations around this? That could centralize the behavior instead of duplicating the `except`/`abort` code across the various APIs...

https://flask-restplus.readthedocs.io/en/stable/errors.html

npalaska: Yes, however, I chose to pass the parameter `silent=True` to `get_json`, it will silence the `BadRequest` exception and return None instead, if it fails to parse the json. According to documentation ` on_json_loading_failed()` will only get called if the mimetype fails to indicate if its a json. So most of the time if its not a json it will return None, and by passing `silent=True` the parsing exception will also return None. therefore both conditions can be handled by checking if return value is None or not

dbutenhof (dismissed):

lib/pbench/server/api/__init__.py
portante: Not sure I follow ... is it true that as it stands right now, `get_json()` will never raise a `BadRequest` exception?

npalaska: Yes, if parsing fails `get_json(silent=True)` will return None instead of calling `on_json_loading_failed()` (`on_json_loading_failed()` raises the BadRequest if it couldt return a value) [doc](https://flask.palletsprojects.com/en/1.1.x/api/#flask.Request.get_json)

portante (approved):
Looks good!

portante (changes requested):
Actually, it needs a final rebase before we can merge it ...

portante:
@npalaska, can we get a rebase and squash of the commits?

portante (approved):
Ship it!

dbutenhof (approved):

FuqingWang:
Just curious, what is the reason of keeping those config files under server/bin/state/? They look identical to me.

portante:
Some of them might be identical, but most or subtly different.  They are there for the unit tests.

PR #1770, opened 2020-09-01
Toolmeister changes

zulcss:
- Split out CLI bits and library bits
- Make everything installable via setuptools
- Use the click library when needed.
- Use a BaseCommand class amongst the toolmeister scripts.
- Consolidate code
- Remove unused code
- Use jinja for configuration file


portante:
@zulcss, any update on this work?

zulcss:
> @zulcss, any update on this work?

Should be good to be reviewed now.

zulcss:
> Can you rebase this work? I just applied a few fixes from @Maxusmusti 

Rebased with mater

lib/pbench/agent/base.py
```diff
@@ -0,0 +1,97 @@
+import abc
+import datetime
+import os
+import socket
+import sys
+from pathlib import Path
+
+import click
+import six
```
portante, line 9: Let's consider removing the use of `six` to detect Python2 environments.

It is not clear we want to embed knowledge of what we require for Python3 into the code.

lib/pbench/agent/base.py
```diff
@@ -0,0 +1,97 @@
...
+        self.context = context
+
+        if six.PY2:
+            click.secho("python3 required, either directly or through SCL", err="red")
+            sys.exit(1)
+        self.config = self.context.config
+
+        # very first thing to do is figure out which pbench we are
+        self.pbench_run = Path(os.environ.get("pbench_run", self.config.pbench_run))
+        if not self.pbench_run.exists():
```
portante, line 25: Is what the previous code did?

lib/pbench/agent/base.py
```diff
@@ -0,0 +1,97 @@
...
+            click.secho(
+                f"[ERROR] the provided pbench run directory, {self.pbench_run}, does not exist",
+                err="red",
+            )
+            sys.exit(1)
+
+        # the pbench temporary directory is always relative to the $pbench_run
+        # directory
+        self.pbench_tmp = Path(os.environ.get("pbench_tmp", (self.pbench_run / "tmp")))
+        self.pbench_tmp.mkdir(parents=True, exist_ok=True)
```
portante, line 35: Won't this raise an exception if the directory path fails to create?

Further, we should consider removing `parents=True` since we verified it exists just above.

lib/pbench/agent/base.py
```diff
@@ -0,0 +1,97 @@
...
+        if not self.pbench_tmp.exists():
+            click.secho(
+                f"[ERROR] unable to create TMP dir, {self.pbench_tmp}", err="red"
+            )
+            sys.exit(1)
+        self.pbench_log = Path(os.environ.get("pbench_log", self.config.pbench_log))
+        self.pbench_install_dir = Path(
```
portante, line 42: Shouldn't this, and the check below, have already been done when constructing a `PbenchAgentConfig()` object?

lib/pbench/agent/base.py
```diff
@@ -0,0 +1,97 @@
...
+            os.environ.get("pbench_install_dir", self.config.pbench_install_dir)
+        )
+        if not self.pbench_install_dir.exists():
+            click.secho(
+                f"[ERROR] pbench installation directory, {self.pbench_install_dir}, does not exist",
+                err="red",
+            )
+            sys.exit(1)
+        self.pbench_bspp_dir = os.environ.get(
+            "pbench_bspp_dir", (self.pbench_install_dir / "bench-scripts/postprocess")
```
portante, line 52: s/`(self.pbench_install_dir / "bench-scripts/postprocess")`/`(self.pbench_install_dir / "bench-scripts" / "postprocess")`/.

lib/pbench/cli/agent/__init__.py
```diff
@@ -0,0 +1,10 @@
+import click
```
portante, line 1: Can you add a module documentation string describing why this module is needed?

lib/pbench/agent/meister/start.py
```diff
@@ -0,0 +1,438 @@
+import errno
+from jinja2 import Environment, FileSystemLoader
```
portante, line 2: Pep8 seems to request that straight `import`s are kept in one section, and `from * import`s are in another.

portante (changes requested):
Thanks, Chuck, good start!

It seems we now have a mix between the use of `tool-meister` and `meister` where the latter is assumed to imply the "tool-meister".  It might worth considering using `tool_meister` under `agent`.

The `tool-meister` daemon and the `tool-data-sink` daemon don't really need to be `click`'d do they?  Their core functions are not callable by any other subsystem, and they really are not part of the Pbench Agent CLI.  It seems best to keep them as simple `argv` command line invoked daemons since the code takes care of invoking them correctly.  Some of the shared code should be moved to a library, but we really need to consider this well.

lib/pbench/agent/meister/start.py
```diff
@@ -0,0 +1,438 @@
...
+            retcode = os.spawnl(
+                os.P_WAIT,
+                data_sink_path,
+                data_sink,
+                "localhost",
+                str(self.redis_port),
+                tds_param_key,
+            )
+        except Exception:
+            logger.exception("failed to create pbench data sink, daemonized")
```
dbutenhof, line 176: What is _daemonized_ supposed to indicate here?

dbutenhof (commented):
I like the hierarchies; I think I'll need another pass to really grok the flow, but a few minor notations along the way...

lib/pbench/agent/meister/start.py
```diff
@@ -99,10 +95,10 @@ def start(self):
                 return 1
 
         try:
-            timeout = time.time() + REDIS_MAX_WAIT
+            timeout = time.time() + 60
```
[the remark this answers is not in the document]
portante: Seems like a bad merge?

lib/pbench/agent/meister/meister.py
```diff
@@ -0,0 +1,177 @@
...
+            return 3
+
+        try:
+            params_raw = redis_server.get(self.context.param_key)
+            if params_raw is None:
+                logger.error(
+                    'Parameter key, "%s" does not exist.', self.context.param_key
+                )
+                return 4
+            logger.info("params_key (%s): %r", self.context.param_key, params_raw)
```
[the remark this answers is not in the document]
portante: This is an artifact of the base logging system used by the server not shared by the agent side.

We should probably fix that in a separate commit.

lib/pbench/agent/meister/start.py
portante: Is this message a hold over from the existing code?  See https://github.com/distributed-system-analysis/pbench/blob/master/agent/util-scripts/pbench-tool-meister-start#L368

lib/pbench/agent/meister/__init__.py
```diff
@@ -0,0 +1,800 @@
...
+        ) = ret_val
+        self._running_tools = dict()
+        self._rs = redis_server
+        logger.debug("pubsub")
+        self._pubsub = self._rs.pubsub()
+        logger.debug("subscribe %s", self._channel)
+        self._pubsub.subscribe(self._channel)
+        logger.debug("listen")
+        self._chan = self._pubsub.listen()
+        logger.debug("done listening")
```
npalaska, line 258: If we are going keep these dubug logs, can we make them a little elaborative or add more info.

lib/pbench/agent/meister/__init__.py
```diff
@@ -0,0 +1,800 @@
...
+        }
+        self._valid_actions = frozenset(["start", "stop", "send", "terminate"])
+        for key in self._state_trans.keys():
+            assert (
+                key in self._valid_actions
+            ), f"INTERNAL ERROR: invalid state transition entry, '{key}'"
+            assert self._state_trans[key]["next"] in self._valid_states, (
+                "INTERNAL ERROR: invalid state transition 'next' entry for"
+                f" '{key}', '{self._state_trans[key]['next']}'"
+            )
```
npalaska, line 286: I think since the values of `self._state_trans`, `self._valid_actions`, and `self._valid_states`  variables are not changing, this for loop is gonna be always true so why do we need to assert that?

lib/pbench/agent/meister/base.py
```diff
@@ -0,0 +1,15 @@
+from pbench.agent.base import BaseCommand
+from pbench.agent.meister.client import ClientMixIn
+from pbench.agent.meister.meister import MeisterMixIn
+from pbench.agent.meister.start import StartMixIn
+from pbench.agent.meister.stop import StopMixIn
+
+
+class MeisterCommand(BaseCommand, ClientMixIn, MeisterMixIn, StartMixIn, StopMixIn):
+    def __init__(self, context):
+        super(MeisterCommand, self).__init__(context)
```
npalaska, line 10: Do we need this the python 2 way? We can use shorter version of super with python 3, I think something like `super().__init__(context)`

agent/util-scripts/unittests
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
-export LANG=C
-export LC_ALL=C
+export LANG=C.UTF-8
+export LC_ALL=C.UTF-8
```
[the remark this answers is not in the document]
zulcss: Otherwise click would get a traceback:

https://click.palletsprojects.com/en/5.x/python3/#python-3-surrogate-handling

lib/pbench/agent/utils.py
```diff
@@ -0,0 +1,60 @@
...
+    else:
+        log = logging.getLogger(name)
+
+    # Make sh logging a bit less verbose
+    logging.getLogger("sh").setLevel(logging.WARNING)
+
+    if (
+        os.environ.get("_PBENCH_TOOL_MEISTER_CLIENT_LOG_LEVEL") == "debug"
+        or os.environ.get("_PBENCH_TOOL_MEISTER_START_LOG_LEVEL") == "debug"
+        or os.environ.get("_PBENCH_TOOL_MEISTER_START_LOG_LEVEL") == "debug"
```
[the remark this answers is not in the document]
zulcss: Yes, unit tests are a bitch

lib/pbench/agent/meister/start.py
zulcss: straight copy with toolmeister

lib/pbench/agent/meister/__init__.py
zulcss: straight copy of toolmeister

lib/pbench/agent/meister/__init__.py
zulcss: straight copy of toolmeister

lib/pbench/agent/meister/__init__.py
```diff
@@ -0,0 +1,800 @@
...
+        self.pbench_bin = pbench_bin
+        ret_val = self.fetch_params(params)
+        (
+            self._benchmark_run_dir,
+            self._channel,
+            self._controller,
+            self._group,
+            self._hostname,
+            self._tools,
+        ) = ret_val
```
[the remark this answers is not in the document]
zulcss: This is straight out of toolmeister, this just a copy and probably should be addressed in another branch.

lib/pbench/cli/agent/__init__.py
zulcss: done

lib/pbench/agent/meister/start.py
zulcss: done

lib/pbench/agent/base.py
zulcss: done

lib/pbench/agent/base.py
zulcss: done

setup.cfg
```diff
@@ -26,5 +26,13 @@ console_scripts =
    pbench-server = pbench.cli.server.shell:main
    pbench-config = pbench.cli.getconf:main
 
+   # backwards compatibility
```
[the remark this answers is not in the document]
zulcss: done

lib/pbench/agent/meister/templates/redis.template
```diff
@@ -0,0 +1,8 @@
+bind {{ hostnames }}
```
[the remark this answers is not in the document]
zulcss: Its past as a list so jinja will behave correctly

zulcss:
> Thanks, Chuck, good start!
> 
> It seems we now have a mix between the use of `tool-meister` and `meister` where the latter is assumed to imply the "tool-meister". It might worth considering using `tool_meister` under `agent`.

Done

> The `tool-meister` daemon and the `tool-data-sink` daemon don't really need to be `click`'d do they? Their core functions are not callable by any other subsystem, and they really are not part of the Pbench Agent CLI. It seems best to keep them as simple `argv` command line invoked daemons since the code takes care of invoking them correctly. Some of the shared code should be moved to a library, but we really need to consider this well.

I considered that, but its using click for consistency and we can re-use the same arguments between the toolmeister and same methods.

lib/pbench/agent/meister/meister.py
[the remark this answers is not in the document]
zulcss: see above

portante:
> I considered that, but its using click for consistency and we can re-use the same arguments between the toolmeister and same methods.

So what we would all be agreeing to is that:

 1. The "Tool Data Sink" and the "Tool Meister" daemon processes are now invocable from the command line just like other pbench agent CLI interfaces, even though we are not going to support them invoked via the CLI
 2. It is worth moving all the code for the "Tool Data Sink" and the "Tool Meister" to a new location to satisfy the `click` infrastructure

The "Tool Data Sink" and the "Tool Meister" will likely be containerized in their own images, but we really can't afford to support them a part of the CLI.

Let's consider this well and only make code changes where really necessary.  In this case, we seem to be adding a lot of mechanism and moving a lot of code for very little gain.

dbutenhof:
> So what we would all be agreeing to is that:
> 
> 1. The "Tool Data Sink" and the "Tool Meister" daemon processes are now invocable from the command line just like other pbench agent CLI interfaces, even though we are not going to support them invoked via the CLI

That seems a bit hyperbolic ... tool meister is already launched as a CLI with arguments; this is just standardizing the definition and processing of the arguments into the new framework. Doing so in no way implies that we're "supporting" general direct user invocation.

(Actually, when the Pythonization is complete, there will be no need/justification for a CLI interface here and the tools and agent should simply instantiate a ToolMeister object directly, making this ... eventually ... irrelevant.)

> 2. It is worth moving all the code for the "Tool Data Sink" and the "Tool Meister" to a new location to satisfy the click infrastructure

That's a better question. I get the idea of leveraging the common code for common arguments. Not sure whether that's worthwhile in this case, though.

> The "Tool Data Sink" and the "Tool Meister" will likely be containerized in their own images, but we really can't afford to support them a part of the CLI.

Agreed; but, again, changing the existing CLI from direct simple argument handling to @zulcss's new parsing framework doesn't really change that.

> Let's consider this well and only make code changes where really necessary. In this case, we seem to be adding a lot of mechanism and moving a lot of code for very little gain.

That's definitely true.

portante:
> > So what we would all be agreeing to is that:
> > 
> > 1. The "Tool Data Sink" and the "Tool Meister" daemon processes are now invocable from the command line just like other pbench agent CLI interfaces, even though we are not going to support them invoked via the CLI
> 
> That seems a bit hyperbolic ... tool meister is already launched as a CLI with arguments;

See PR https://github.com/distributed-system-analysis/pbench/pull/1809, which demonstrates that the "CLI" is only an artifact of implementation, and that with very little code changes the CLI can be removed entirely for the Tool Data Sink, and replaced with a minimal SSH-only wrapper for the Tool Meisters.

> this is just standardizing the definition and processing of the arguments into the new framework. Doing so in no way implies that we're "supporting" general direct user invocation.

While that is true, it is not clear that we want to do that, given both the Tool Data Sink and the Tool Meister are not standard command line interfaces we want users to have access to.

> 
> (Actually, when the Pythonization is complete, there will be no need/justification for a CLI interface here and the tools and agent should simply instantiate a ToolMeister object directly, making this ... eventually ... irrelevant.)

It seems that PR https://github.com/distributed-system-analysis/pbench/pull/1809 supports that conclusion as well.

> 
> > The "Tool Data Sink" and the "Tool Meister" will likely be containerized in their own images, but we really can't afford to support them a part of the CLI.
> 
> Agreed; but, again, changing the existing CLI from direct simple argument handling to @zulcss's new parsing framework doesn't really change that.

Doesn't it add unnecessary mechanism for which their operation is not required, and places both the tool data sink and the tool meister daemon into the CLI command hierarchy?  Do we really want anybody who installs the pbench agent to have the ability to directly invoke those two daemons directly?  The original intention was that `pbench-tool-meister-start` handles all that is required for their operation, hiding details and interfaces from the user.

lib/pbench/agent/base.py
zulcss: Yes

zulcss:
This has been deprecated into smaller bits.

PR #1774, opened 2020-09-03
ovs-hw-offload flows support in openvswitch-datalog

pradiptapks:
Currently the pbench `openvswitch-datalog` tool-script support collecting metrics from standard openvswitch and ovs-dpdk layer.
As openvswitch feature extends with TC hardware offload support and during the network performance test with trafficgen is required to review the flows which dump in SMART NIC (e.g., Mellanox Connectx5).

With minor code change in `openvswitch-datalog` script, we are able to collect the OVS hardware offload follows from DUT node. Since ovs-hw-offload in OpenStack 16.x is [GA](https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/16.1/html-single/network_functions_virtualization_planning_and_configuration_guide/index#sect-configuring-hw-offload), so the changes `openvswitch.datalog` can collect flows periodically for further analysis.

### DIFF
--- a/agent/tool-scripts/datalog/openvswitch-datalog
+++ b/agent/tool-scripts/datalog/openvswitch-datalog
@@ -52,7 +52,9 @@ rxq_mapping_file=${tool_output_dir}/pmd-rxq-show.txt
 ovs-appctl dpif-netdev/pmd-rxq-show > ${rxq_mapping_file}
 
 stats_file=${tool_output_dir}/ovs-stats.txt
+offload_stats_file=${tool_output_dir}/ovs-offload-stats.txt
 > ${stats_file}
+> ${offload_stats_file}
 if [[ ! -z "${bridges}" ]]; then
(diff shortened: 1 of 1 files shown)


portante:
@pradiptapks, in the future, don't open a new PR in order to make changes to an existing one.  Please use `git commit --amend` to make update your branch, and use `git push <your github repo> +<your branch>`, where the leading `+` indicates a "force push".

portante (changes requested):
Seems like you need to rebase this commit, and deal with a tabs vs spaces issue in the datalog that was changed.

What would it take to add support for post-processing the hardware offload stats in the `openvswitch-postprocess` Perl script?

portante:
We are not going to continue this approach of adding more data gathering code directly in Pbench.  Instead, the use of PCP or Prometheus data collectors is encouraged.

PR #1778, opened 2020-09-03
Tool Metadata Abstraction

Maxusmusti:
IN PROGRESS:
Working on tool metadata abstraction


agent/tool-scripts/meta.json
```diff
@@ -0,0 +1,48 @@
+{
+	"transient":{
```
dbutenhof, line 2: Interestingly, I'd imagined it "inside out" with ["node-exporter": {"type":"persistent"}, "blktrace":{"type":"transient"}, ...]

Yours is better as you can 'tool in meta.transient.keys()' which makes a lot more sense, and we can still easily define properties for each.

agent/util-scripts/pbench-tool-meister
```diff
@@ -1152,8 +1153,16 @@ def main(argv):
             #   a. handle graceful termination (TERM, INT, QUIT)
             #   b. log operational state (HUP maybe?)
 
+            #ADDING METADATA GRAB HERE
+            meta_raw = redis_server.get("tool-metadata")
+            if params_raw is None:
+                logger.error('Metadata was never loaded.')
+            meta_str = meta_raw.decode("utf-8")
+            tool_metadata = json.loads(meta_str)
+            logger.info(f"Metadata: {tool_metadata}")
```
dbutenhof, line 1162: I'm not sure this is the only place we'll want access to the metadata. Maybe something like a ToolMetadata object that can be constructed either from the JSON or from the Redis document, and passed around as context with methods to abstract the representation?

dbutenhof (commented):
Cool: quick prototyping.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -342,6 +342,12 @@ def main(argv):
         ), f"bad channel: {resp!r}"
         assert resp["data"] == 1, f"bad data: {resp!r}"
 
+    # 2.5. Add tool metadata json to redis
+    #FIXME: ABSOLUTE PATH IS BAD (also add try/catch)
+    with open('/opt/pbench-agent/tool-scripts/meta.json') as json_file:
```
npalaska, line 347: We can add the path in the pbench-agent config file I guess

agent/tool-scripts/meta.json
portante: We can dynamically create the convenient way to access the data, too.

agent/util-scripts/pbench-tool-meister
portante: Both the on-disk JSON document and the in-Redis JSON document should be the same, so an entity constructing the `ToolMetadata` object can just read from on or the other to then construct the object.

agent/util-scripts/pbench-tool-meister-start
portante: We know the path, we just need the installation directory, to that access the path relative to that:

```
with Path(config.install_dir, "tool-scripts", "meta.json").open("r") as json_file:
    ...
```

agent/tool-scripts/meta.json
dbutenhof: We absolutely want all behavior to be encapsulated behind a class API, with nobody outside the class (or singleton, since I generally dislike static members) knowing anything about the JSON schema.

Maxusmusti:
Work continued in PR #1787

PR #1782, opened 2020-09-04
Obsolete exception SosreportHostname & TarError

### LINKED ISSUE #1763: pbench-index.py: obsolete error codes
The pbench-index.py  error codes need some updating: on[ ll.90 and 91](https://github.com/distributed-system-analysis/pbench/blob/80c8d8e2fa8d99066bf9c637286f7729a5935b73/server/bin/pbench-index.py#L90), the error codes 10 and 11 are no longer generated, so the comments referring to them can be removed.

riya-17:
Fixes #1763

SosreportHostname & TarError exceptions are not being used anymore.
This PR deals with getting rid of it.


dbutenhof (commented):
Interestingly, this (and the success of unit testing) suggests that there were no unit tests that cared about the changed return values from pbench-index.py. That sounds completely wrong in a number of ways.

server/bin/pbench-index.py
```diff
@@ -220,7 +216,7 @@ def main(options, name):
         idxctx.logger.exception(
             "Unexpected error encountered generating list" " of tar balls to process"
         )
-        return 12
+        return 10
```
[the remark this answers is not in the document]
riya-17: Sure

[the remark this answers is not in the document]
riya-17: It really doesn't seem to change any test case on changing the error code.

server/bin/pbench-index.py
```diff
@@ -124,7 +129,7 @@ def main(options, name):
         return 2
     except BadConfig as e:
         print(f"{name}: {e}", file=sys.stderr)
-        return 3
+        return BAD_CFG
     except JsonFileError as e:
         print(f"{name}: {e}", file=sys.stderr)
         return 8
```
dbutenhof, line 135: Thanks; but 0, 1, 2, and 8 are still returned as literals.

dbutenhof (commented):
It'd be great if we had unit tests that ensured each of the detected conditions was properly reported; maybe it's not a big deal since I don't think any external code really cares... it's just run by a crontab entry.

server/bin/pbench-index.py
```diff
@@ -87,9 +85,7 @@ def main(options, name):
             7 - Bad metadata.log file encountered
             8 - Error reading a mapping file for Elasticsearch templates
             9 - Error creating one of the Elasticsearch templates
-           10 - Bad hostname in a sosreport
-           11 - Failure unpacking the tar ball
-           12 - generic error, needs to be investigated and can be retried
+            10 - generic error, needs to be investigated and can be retried
```
portante, line 88: We don't want to change the meaning of the generic error, because we track this value on-disk currently.  Best to leave this as `12`, and just remove references to `10` and `11`.

portante: And perhaps instead of encoding this is comments, we turn this into a dictionary, where we use the constants you define below as keys to a dictionary containing the strings they map to for the exception.  Then we can reference those strings algorithmically when building log messages from the exceptions.

server/bin/pbench-index.py
```diff
@@ -114,7 +119,7 @@ def main(options, name):
             " use --config <file> on the command line",
             file=sys.stderr,
         )
-        return 2
+        return BAD_CFG
```
portante, line 122: We need to be careful here because we are now changing the return value.  Let's make sure we don't change any return values as part of this change.

server/bin/pbench-index.py
portante: Yes, let's get those literals changed as well.

server/bin/pbench-index.py
```diff
@@ -428,7 +425,7 @@ def main(options, name):
                         f"Logic Bomb!  Unexpected tar ball handling "
                         f"result status {tb_res:d} for tar ball {tb}"
                     )
-                elif tb_res >= 4 or res <= 11:
```
portante, line 431: And we might want to add a method to encapsulate what is ignorable.

Also, there appears to be a bug here, s/`res <= 9`/`tb_res < GENERIC_ERROR`.

portante (changes requested):

server/bin/pbench-index.py
riya-17: ok

server/bin/pbench-index.py
riya-17: ok

server/bin/pbench-index.py
riya-17: ok

dbutenhof: And Pete's comment got me back to thinking about this...

This could also be a decent use for an Enum, or an IntEnum (which would allow exiting with the integer equivalent). (Although Python enum isn't as flexible -- or as easily abused? -- as Java enum!)

For example,

```
class ExitCode(IntEnum):
    OK = 0                            # Successful completion
    OP_ERROR = 1             # Operational error while indexing
    CONFIG_FILE = 2          # Configuration file not specified
    BAD_CFG = 2                 # Bad configuration file
```
In fact, if I was doing this in Java I might even extend the enum constructor to include a default message... not so much because we need it, but because it would move the documentation commentary into actual code. And it'd be easy to add an "ignorable" property, too. It's not clear that's feasible in Python, though.

An alternative that would combine all the meta-data might be something like this:

```
class ErrorCode:
    def __init__(self, name, value, ignorable, message):
        self.name = name
        self.value = value
        self.ignorable = ignorable
        self.message = message

class Errors:
    def __init__(self, *codes):
        self.errors = dict()
        for code in codes:
            self.errors[error.name] = error

    def __getitem__(self, key):
        return self.errors[key]

error_codes = Errors(
    ErrorCode('OK', 0, True, 'Successful completion'),
    ErrorCode('OP_ERROR', 1, False, 'Operational error during indexing'),
    [...]
    )
```

riya-17: seems like a good approach

server/bin/pbench-index.py
riya-17: @portante  Sorry, I couldn't understand this "method to encapsulate what is ignorable"

server/bin/pbench-index.py
```diff
@@ -331,30 +376,30 @@ def main(options, name):
                             idxctx.es, actions, fp, idxctx.logger, idxctx._dbg
                         )
                 except UnsupportedTarballFormat as e:
-                    idxctx.logger.warning("Unsupported tar ball format: {}", e)
-                    tb_res = 4
+                    idxctx.logger.warning(
+                        "{}: {}", error_codes["TB_META_ABSENT"].message, e
+                    )
+                    tb_res = error_codes["TB_META_ABSENT"].value
```
dbutenhof, line 382: Interesting. While this is a faithful representation of the previous error code here, I'm not sure that it was really accurate. The description references the metadata file, but the code in the try block seems most likely to fail attempting to open the tarball or maybe constructing the generator. Does it even reference the metadata file?

Pete, should this maybe really have been "problem reading tar file" rather than referring to metadata in the first place? It seems that, regardless of the error code, the old log message may have been more meaningful because it really doesn't match with the comment description of error 4.

I guess I kinda opened this can of worms... but maybe we can at least straighten out the worms  a bit before we close it again.

server/bin/pbench-index.py
```diff
@@ -428,7 +473,7 @@ def main(options, name):
                         f"Logic Bomb!  Unexpected tar ball handling "
                         f"result status {tb_res:d} for tar ball {tb}"
                     )
-                elif tb_res >= 4 or res <= 11:
+                elif tb_res >= 4 or res < error_codes["GENERIC_ERROR"].value:
```
dbutenhof, line 476: This is a weird expression. I don't really see how it even makes sense; but this is the purpose for which I'd envisioned the "ignorable" property on the error object, in response to Pete's earlier comment.

If we can't figure out how to leverage that here, maybe the attribute was a bad idea.

dbutenhof (commented):
@portante ... that "ignore" code is confusing. It also seems that the original comment for error code 4 was simply wrong, leading to a possibly misleading message now that it's been built into code. (See, that's one great thing about moving "documentation" from comments into code; it becomes visible and testable.)

server/bin/pbench-index.py
```diff
@@ -161,19 +203,19 @@ def main(options, name):
         "INCOMING", INCOMING_rp, idxctx.logger
     )
     if not INCOMING_path:
-        res = 3
+        res = error_codes["BAD_CFG"].value
```
npalaska, line 206: Here instead of assigning a value to `res`, should we assign an ErrorCode object itself to it something like `res = error_codes["BAD_CFG]`. That way we can later check if the res is ignorable or what is the value of res when we are returning it. So on line 222 we can return it as `return res.value` instead of `return res`. It will also help on line 476 where we need to check whether `res` or `tb_res` is ignorable or not by just doing `if res.ignorable or tb_res.ignorable`

server/bin/pbench-index.py
```diff
@@ -114,20 +156,20 @@ def main(options, name):
             " use --config <file> on the command line",
             file=sys.stderr,
         )
-        return 2
+        return error_codes["CFG_ERROR"].value
```
npalaska, line 159: Nit: should we change the plural name `error_codes` -> `error_code` as we always refer to one error code everytime

server/bin/pbench-index.py
npalaska: I agree with Dave, maybe instead of assigning a value to `res` or `tb_res` we can keep them as ErrorCode object so we can check if they are ignorable or not something like `if res.ignorable or tb_res.ignorable`. Please check my previous comment

server/bin/pbench-index.py
riya-17: got the idea now. Thanks!

server/bin/pbench-index.py
riya-17: sure

server/bin/pbench-index.py
riya-17: "ignorable" property just checks whether the code is ignorable or not while here we are dealing with error codes. We need to specify a range or error code for which a particular message assigned to it is logged. If I use ignorable here it might be true for multiple values and we might end up logging the wrong error message. like it might be true for GENERIC values as well but we do not deal with GENERIC values here.

server/bin/pbench-index.py
riya-17: @portante can you please clear this confusion

server/bin/pbench-index.py
```diff
@@ -423,23 +468,29 @@ def main(options, name):
                     rename_tb_link(
                         tb, Path(controller_path, f"{linkerrdest}.1"), idxctx.logger,
                     )
-                elif tb_res in (2, 3):
+                elif tb_res.value in (2, 3):
```
dbutenhof, line 471: We need to think about what to call it; but it'd be cleaner to encapsulate this check (along with the >=4 or <12 below) into the metadata. Perhaps what we need is something other than just "ignorable". I don't fully understand the real semantic intent of this code ... which is exactly why I don't like it, and would like to replace it with something that's more clear as well as cleanly encapsulated.

2 and 3 are "configuration data" errors (missing or bad), but it's not clear we ever actually use those for `tb_res` in the first place, nor what it would mean since `tb_res` deals with the tarball.

I still can't imagine the purpose of checking `tb_res >=4 or res <= 11`, and that bothers me; encapsulating the decisions would make this much more clear. Perhaps Pete and/or Nick can recall the original intent, and we can design something that makes sense, is clear to a reader, and maintainable???

server/bin/pbench-index.py
```diff
@@ -423,23 +468,29 @@ def main(options, name):
...
                     assert False, (
                         f"Logic Bomb!  Unexpected tar ball handling "
-                        f"result status {tb_res:d} for tar ball {tb}"
+                        f"result status {tb_res.value:d} for tar ball {tb}"
                     )
-                elif tb_res >= 4 or res <= 11:
+                elif (
+                    tb_res.value >= 4
+                    or tb_res.value < error_code["GENERIC_ERROR"].value
+                ):
```
dbutenhof, line 479: This has changed the original code, regardless of the inscrutability of the original intent. The original was `tb_res >= 4 or res <= 11` and the new code checks both conditions against `tb_res`.

dbutenhof (commented):
If Pete or Nick can't remember the original intent of the "special conditions" for res and tb_res, I might try to build my own table and see if I can recreate why it might have made sense; right now it all seems fairly arbitrary. (Sigh.)

server/bin/pbench-index.py
portante: If we have a `class` that describes the return codes from indexing tar balls, then we can add a method to that class that will report if a given return value is in the "ignorable" range.

server/bin/gold/test-5.2.txt
```diff
@@ -976,31 +976,31 @@ run-1970-01-01T00:00:42-UTC: Processed 10 result tar balls, 7 successfully (0 pa
 1970-01-01T00:00:42.000000 DEBUG pbench-index.pbench-index main -- start processing list of tar balls
 1970-01-01T00:00:42.000000 INFO pbench-index.pbench-index main -- Starting /var/tmp/pbench-test-server/test-5.2/pbench/archive/fs-version-001/controller-b-with-prefixes/TO-INDEX/tarball-0_1970.01.01T00.42.00.tar.xz (size 212)
 1970-01-01T00:00:42.000000 DEBUG pbench-index.pbench-index main -- open tar ball
-1970-01-01T00:00:42.000000 WARNING pbench-index.pbench-index main -- Unsupported tar ball format: /var/tmp/pbench-test-server/test-5.2/pbench/archive/fs-version-001/controller-b-with-prefixes/tarball-0_1970.01.01T00.42.00.tar.xz - tar ball is missing "tarball-0_1970.01.01T00.42.00/metadata.log".
+1970-01-01T00:00:42.000000 WARNING pbench-index.pbench-index main -- Tar ball does not contain a metadata.log file: /var/tmp/pbench-test-server/test-5.2/pbench/archive/fs-version-001/controller-b-with-prefixes/tarball-0_1970.01.01T00.42.00.tar.xz - tar ball is missing "tarball-0_1970.01.01T00.42.00/metadata.log".
```
portante, line 979: Is there a reason for this message change?  Does it convey additional information?  It seems like it is repeating information in the message, where the old message didn't.

server/bin/gold/test-7.1.txt
```diff
@@ -303,7 +303,7 @@ drwxrwxr-x          - tmp
 1970-01-01T00:00:42.000000 DEBUG pbench-index.pbench-index main -- start processing list of tar balls
 1970-01-01T00:00:42.000000 INFO pbench-index.pbench-index main -- Starting /var/tmp/pbench-test-server/test-7.1/pbench/archive/fs-version-001/controller/TO-INDEX/test_7.1_1970.01.01T00.00.00.tar.xz (size 1256)
 1970-01-01T00:00:42.000000 DEBUG pbench-index.pbench-index main -- open tar ball
-1970-01-01T00:00:42.000000 WARNING pbench-index.pbench-index main -- The metadata.log file is curdled in tar ball: /var/tmp/pbench-test-server/test-7.1/pbench/archive/fs-version-001/controller/test_7.1_1970.01.01T00.00.00.tar.xz - error fetching required metadata.log fields, "No section: 'run'"
+1970-01-01T00:00:42.000000 WARNING pbench-index.pbench-index main -- Bad metadata.log file encountered: /var/tmp/pbench-test-server/test-7.1/pbench/archive/fs-version-001/controller/test_7.1_1970.01.01T00.00.00.tar.xz - error fetching required metadata.log fields, "No section: 'run'"
```
portante, line 306: Is there a reason to change this message?

server/bin/pbench-index.py
portante: Not sure exactly how to answer this, but perhaps by looking at the original code it will make sense.

In the original code, prior to line 406, the variable `tb_res` was set to an integer value by the `try`/`except`/`else` block from lines 306 - 372.   Then the `if` / `elif` / `else` block here from lines 406 - 448 attempts to decide what actions to take based on the numeric error code value.  This `if` block attempts to deal with the error codes in numerical order:  `0` success (lines 407 - 416), `1` indexing error (lines 418 - 425), `2` & `3` logic bomb! (lines 427 - 430; this should not happen because the `try`/`except`/`else` block from lines 306 - 372 does not assign `tb_res` to a value of `2` or `3`), `4` thru `11` quietly move the tar ball to a error destination without logging anything (due to the fact that those errors are logged elsewhere), `else` all other errors logged and moved to the default error destination.

The danger with encapsulation here is that we'll be adding a lot of mechanism to the code which is all inline and not separated across modules.  So we want to be careful we don't over engineer this fix.

server/bin/pbench-index.py
portante: I think the original code had a bug, I don't see `res` as a variable that we need to check in this context.

portante (dismissed):

server/bin/gold/test-5.2.txt
riya-17: No, Earlier messages were written in logger directly now we are using error code comments

server/bin/gold/test-7.1.txt
riya-17: No, Earlier messages were written in logger directly now we are using error code comments

server/bin/pbench-index.py
riya-17: Yaa I thought it's a bug so I replaced

server/bin/pbench-index.py
```diff
@@ -468,23 +468,26 @@ def main(options, name):
...
                     )
-                elif tb_res in (2, 3):
+                elif tb_res.value in (2, 3):
                     assert False, (
                         f"Logic Bomb!  Unexpected tar ball handling "
-                        f"result status {tb_res:d} for tar ball {tb}"
+                        f"result status {tb_res.value:d} for tar ball {tb}"
                     )
-                elif tb_res >= 4 or res < error_codes["GENERIC_ERROR"].value:
+                elif tb_res.value >= 4 or tb_res.value < 12:
```
dbutenhof, line 476: Now this really makes no sense; 12 is always `>= 4` and [0, 1, 2, 3] are `< 12` so I'm pretty sure this doesn't do whatever we want it to do. Should this be `and` so we ignore [4, 5, 6, 7, 8, 9] errors? That lines up with Pete's explanation earlier that we're ignoring the "tarball format" errors assuming they've already been logged.

So [1, 2, 3, 8, 9, 12] are _internal_ errors (that shouldn't or can't happen), while [4, 5, 6, 7] are errors in the tarball format. Maybe it makes sense to silently ignore tarballs that are clearly bad, while reporting the internal errors and potentially retrying. This seems to somewhat tie to the `and` alternative. If that was the intent, then I think the solution would be an Error object property defining each error as "internal" or "tarball_error", and then this would be `elif tb_res.tarball_error:`

Then again, the range includes "internal" indexing/template errors [8,9] which Pete suggests are simply included here because they're already logged. (Although I don't see any errors detected that _aren't_ logged above, so I'm not convinced that tracks.) We already assume this was broken originally in checking both `res` and `tb_res`, so it's not hard to believe that this should really be `>= 4 and < 8` if it comes to that.

And the way the data is structured now, the explicit `ignorable` property could be dropped and we could just have

```
def ignorable(self):
    return self.value == 0
```

Or maybe even call it _success_ to be more clear.

server/bin/pbench-index.py
```diff
@@ -45,6 +43,28 @@
...
+        self.errors = dict()
+        for code in codes:
+            self.errors[code.name] = code
+
+    def __getitem__(self, key):
+        return self.errors[key]
+
+
+def success(res):
+    return res.value == 0
```
dbutenhof, line 65: I would make this a method of `ErrorCode`, so that you can use `res.success()` ... it's a property of the Error, not an unrelated function.

Another possibility would be for `ErrorCode`'s `__init__` to create an instance property `success` like

```
    self.success = value == 0
```

and then you could just use `res.success`

lib/pbench/server/indexing_tarballs.py
```diff
@@ -275,37 +277,37 @@ def sigquit_handler(*args):
...
+                            idxctx.logger.warning(
+                                "{}: {!r}", error_code["BAD_DATE"].message, e
+                            )
+                            tb_res = error_code["BAD_DATE"]
                         except FileNotFoundError as e:
-                            idxctx.logger.warning("No such file: {}", e)
-                            tb_res = 6
-                        except BadMDLogFormat as e:
                             idxctx.logger.warning(
-                                "The metadata.log file is curdled in" " tar ball: {}", e
```
dbutenhof, line 288: The new error message is much better, although we apparently no longer identify the tarball with the problem, which isn't good.

dbutenhof (commented):
I was looking this over "again" and realized I'd neglected to post several comments from a few days back...

lib/pbench/test/unit/server/test_indexing_tarballs.py
```diff
@@ -0,0 +1,54 @@
...
+        except Exception as exc:
+            raise FileNotFoundError(str(exc))
+
+    def get_data(self):
+        date = None
+        try:
+            date = self._data["_source"]
+        except Exception as e:
+            print("date error ", e)
+        return date
```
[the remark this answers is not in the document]
webbnh: Is this code all gone, now?  (I cannot find it.)

I wanted to point out that catching all exceptions at line 15 and then raising `FileNotFoundError` is probably not a good idea, as the real exception might be that the file is present but contains invalid JSON.  Including the stringization of the original exception as the message for the resulting exception helps, but, if what you want to do is to report the original exception, then you probably don't need to catch it here (it will report itself...).  Alternatively, you can use the `raise FileNotFoundError from exc` idiom, but, again, it's a tad confusing/misleading to be reporting a file-not-found issue if the root cause is bad file contents.  (A more generic exception, like `RuntimeError`, might be a better choice.)

Similarly, the `try` block at line 20 is a bit of a "big hammer".  There are only a few things which could go wrong at line 21 (i.e., `_data` is not a `dict` (e.g., it's `None`), or it doesn't contain the `_source` key), they are things which can be detected locally and easily, and testing for them explicitly would enable more descriptive diagnostics.

However, if this code has been removed, then I suppose it's moot.  :-)

lib/pbench/server/indexing_tarballs.py
```diff
@@ -50,7 +48,7 @@ def _count_lines(fname):
 class Index:
     """class used to collect tarballs and index them"""
 
-    def __init__(self, name, options, idxctx, incoming):
+    def __init__(self, name, options, idxctx, incoming, error_code):
```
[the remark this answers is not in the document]
riya-17: ok

lib/pbench/test/unit/server/test_indexing_tarballs.py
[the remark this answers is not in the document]
riya-17: I just thought of adding tests at the end and also I was a lot confused about them so will add a different commit for them

lib/pbench/server/indexing_tarballs.py
riya-17: I think it is a Github issue that it isn't aligning the right message with that particular exception.

lib/pbench/test/unit/server/test_pbench_index.py
```diff
@@ -0,0 +1,67 @@
+import json
+import pytest
+from server.bin.pbench_index import main
+
+
+class options(object):
+    def __init__(self):
+        self.cfg_name = (
+            "/home/rriya/Git/pbench/server/bin/state/config/pbench-server.cfg"
```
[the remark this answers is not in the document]
riya-17: yeah I was just trying to make it work

lib/pbench/test/unit/server/test_pbench_index.py
```diff
@@ -0,0 +1,67 @@
+import json
+import pytest
+from server.bin.pbench_index import main
+
+
+class options(object):
```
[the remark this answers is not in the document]
riya-17: ok

lib/pbench/server/indexing_tarballs.py
```diff
@@ -116,26 +116,26 @@ def collect_tb(self, ARCHIVE_rp, qdir):
             raise
         except Exception:
             idxctx.logger.exception(
-                "Unexpected error encountered generating list"
-                " of tar balls to process"
+                "{} generating list" " of tar balls to process",
```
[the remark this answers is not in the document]
riya-17: yeah sure

lib/pbench/test/unit/server/test_indexing_tarballs.py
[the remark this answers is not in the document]
riya-17: I will take care of these points and improve that test

lib/pbench/server/indexing_tarballs.py
```diff
@@ -517,3 +535,35 @@ def sigquit_handler(*args):
...
+            ),
+            ErrorCode(
+                "TEMPLATE_CREATION_ERROR",
+                9,
+                False,
+                "Error creating one of the Elasticsearch templates",
+            ),
+            ErrorCode("GENERIC_ERROR", 12, False, "Unexpected error encountered"),
+        )
+        return error_code
```
dbutenhof, line 569: I think that the `error_code(s)` list should be constructed once as it was before, just within the scope of the module. (Or even as a Class property) rather than constructing it dynamically in multiple places with this method. They don't change!

lib/pbench/server/indexing_tarballs.py
dbutenhof: Yes, apparently the tarball name in the output comes from the exception...

lib/pbench/server/indexing_tarballs.py
```diff
@@ -320,7 +340,11 @@ def sigquit_handler(*args):
                                 failures,
                                 retries,
                             )
-                            tb_res = 1 if failures > 0 else 0
+                            tb_res = (
+                                error_code["OP_ERROR"]
+                                if failures > 0
+                                else error_code["OK"]
+                            )
```
webbnh, line 347: For your consideration...since it's only the error which is the conditional,
```
                            tb_res = error_code["OP_ERROR" if failures > 0 else "OK"]
```

lib/pbench/server/indexing_tarballs.py
```diff
@@ -364,7 +388,7 @@ def sigquit_handler(*args):
                         # `linkerrdest` for later retry.
                         controller_path = linksrc_dir.parent
 
-                        if tb_res == 0:
+                        if tb_res == error_code["OK"]:
```
webbnh, line 391: Comparing objects/user-defined types with `==` makes me a little squirmy...I always have to go look up what it means.  If this is really what you want, I would recommend providing an explicit method on the object to do it.  Alternatively, if you accept Dave's suggestion to statically define `error_code`, then you can use `is` instead of `==`, which makes things clear.

server/bin/pbench-index.py
```diff
@@ -61,9 +61,7 @@ def main(options, name):
             7 - Bad metadata.log file encountered
             8 - Error reading a mapping file for Elasticsearch templates
             9 - Error creating one of the Elasticsearch templates
-           10 - Bad hostname in a sosreport
-           11 - Failure unpacking the tar ball
-           12 - generic error, needs to be investigated and can be retried
+            12 - generic error, needs to be investigated and can be retried
```
webbnh, line 64: This comment should presumably be removed from here:  the definitions of the error codes are no longer here, so the detailed documentation of them should not be here, either -- it should be with the definitions.  (A reference to where those definitions can be found might be good, but there will be nothing which causes this comment to be updated when the actual definitions are changed, so having this here is potentially worse than useless.)

webbnh (approved):
It looks like a nice improvement.

I have a few coding/style suggestions and one doc request.  And, I second Dave's suggestion to use a single, statically-defined set of error codes.

lib/pbench/server/indexing_tarballs.py
```diff
@@ -275,37 +295,37 @@ def sigquit_handler(*args):
...
+                                "{}: {}", error_code["TB_META_ABSENT"].message, e
+                            )
+                            tb_res = error_code["TB_META_ABSENT"]
                         except BadDate as e:
-                            idxctx.logger.warning("Bad Date: {!r}", e)
-                            tb_res = 5
+                            idxctx.logger.warning(
+                                "{}: {!r}", error_code["BAD_DATE"].message, e
+                            )
+                            tb_res = error_code["BAD_DATE"]
```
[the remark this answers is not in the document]
riya-17: > Any reason why you omitted the `!r` from line 299 when you used it in line 304? (Unless there is a good reason -- like it produces a different result which is important -- you should be consistent.)

I have added them wherever I could. rest of them are producing different output so I will just omit them

lib/pbench/server/indexing_tarballs.py
riya-17: fixed

lib/pbench/server/indexing_tarballs.py
```diff
@@ -36,6 +34,25 @@ class SigTermException(Exception):
...
+        self.success = value == 0
+        self.tarball_error = tarball_error
+        self.message = message
+
+
+class Errors:
+    def __init__(self, *codes):
+        self.errors = dict()
+        for code in codes:
+            self.errors[code.name] = code
```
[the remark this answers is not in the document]
riya-17: fixed

lib/pbench/server/indexing_tarballs.py
```diff
@@ -116,26 +134,26 @@ def collect_tb(self, ARCHIVE_rp, qdir):
...
             # tuple to return the status and return value
-            return (12, [])
+            return (error_code["GENERIC_ERROR"].value, [])
         else:
             if not tarballs:
                 idxctx.logger.info("No tar balls found that need processing")
                 return (0, [])
 
-        return (0, sorted(tarballs))
+        return (error_code["OK"].value, sorted(tarballs))
```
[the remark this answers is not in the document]
riya-17: done

lib/pbench/server/indexing_tarballs.py
```diff
@@ -147,8 +165,10 @@ def process_tb(self, tarballs):
             idxctx.logger.debug("update_templates [start]")
             idxctx.templates.update_templates(idxctx.es)
         except TemplateError as e:
-            idxctx.logger.error("update_templates [end], error {}", repr(e))
-            res = 9
+            idxctx.logger.error(
+                "{}, {}", error_code["TEMPLATE_CREATION_ERROR"].message, repr(e)
```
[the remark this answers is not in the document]
riya-17: ok

lib/pbench/server/indexing_tarballs.py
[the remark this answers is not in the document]
riya-17: fixed

server/bin/pbench-index.py
riya-17: done

lib/pbench/server/indexing_tarballs.py
riya-17: fixed

lib/pbench/server/indexing_tarballs.py
riya-17: done

lib/pbench/server/indexing_tarballs.py
```diff
@@ -201,6 +201,11 @@ def collect_tb(self, ARCHIVE_rp, qdir):
 
         return (error_code["OK"].value, sorted(tarballs))
 
+    def emit_error(self, logger, error, exception):
+        ec = self.error_code[error]
+        logger("{}: {}", ec.message, exception)
+        return ec
```
dbutenhof, line 207: Losing the actual method name (reporting them all as `emit_error`) bothers me a little, but it's outweighed by the nice consistency of this method. Also, I can't really say it's all that much worse reporting every log message as being from `emit_error` vs `process_tb` -- which really isn't all that much more helpful anyway.

I'm not quite sure how I feel about using `logger` as the local name for a reference to a specific logger method; it's not as clear as I'd like. And in fact, raising all of them to `logger.exception` would take care of the former problem by including the stack trace to show where it was actually triggered ... although I'm not sure the extra logging volume is really worthwhile.

In short, I'm just thinking aloud. You could probably come up with a formal parameter name that would make the usage more clear, like maybe `logger_method`, but I guess that's not a terribly big deal (nor an enormous improvement).

dbutenhof (dismissed):
I have a mild concern about naming and the error log context, but I don't feel strongly enough (nor have a sufficiently better suggestion) to push for another change.

webbnh (changes requested):
Looks pretty good, but I have a couple of small requests.

lib/pbench/server/indexing_tarballs.py
webbnh: If we bought into always using `.exception()` for the logging, we could probably eliminate the `logger` parameter altogether, as it's available as `self.idxctx.logger`.

server/bin/gold/test-5.2.txt
```diff
@@ -968,31 +968,31 @@ run-1970-01-01T00:00:42-UTC: Processed 10 result tar balls, 7 successfully (0 pa
 1970-01-01T00:00:42.000000 DEBUG pbench-index.indexing_tarballs process_tb -- start processing list of tar balls
 1970-01-01T00:00:42.000000 INFO pbench-index.indexing_tarballs process_tb -- Starting /var/tmp/pbench-test-server/test-5.2/pbench/archive/fs-version-001/controller-b-with-prefixes/TO-INDEX/tarball-0_1970.01.01T00.42.00.tar.xz (size 212)
 1970-01-01T00:00:42.000000 DEBUG pbench-index.indexing_tarballs process_tb -- open tar ball
-1970-01-01T00:00:42.000000 WARNING pbench-index.indexing_tarballs process_tb -- Tar ball does not contain a metadata.log file: /var/tmp/pbench-test-server/test-5.2/pbench/archive/fs-version-001/controller-b-with-prefixes/tarball-0_1970.01.01T00.42.00.tar.xz - tar ball is missing "tarball-0_1970.01.01T00.42.00/metadata.log".
+1970-01-01T00:00:42.000000 WARNING pbench-index.indexing_tarballs emit_error -- Tar ball does not contain a metadata.log file: /var/tmp/pbench-test-server/test-5.2/pbench/archive/fs-version-001/controller-b-with-prefixes/tarball-0_1970.01.01T00.42.00.tar.xz - tar ball is missing "tarball-0_1970.01.01T00.42.00/metadata.log".
```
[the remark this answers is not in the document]
riya-17: We can take a poll @ndokos @dbutenhof @portante @npalaska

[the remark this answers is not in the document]
riya-17: It doesn't concern me

lib/pbench/server/indexing_tarballs.py
```diff
@@ -65,7 +63,65 @@ def _count_lines(fname):
...
+       Return Values (now a sub-set of the original status codes above):
+         0 - Successfully processed all tar balls (errors processing tar
+             balls are reported in the logs and in index status reports)
+         1 - Failed to process one or more tar balls for unknown reasons
+             (see logs)
+         2 - Missing configuration file
+         3 - Invalid configuration file
+         8 - Unable to load and process expected mapping files
+         9 - Unable to update index templates in configured Elasticsearch
+             instance
```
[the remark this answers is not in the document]
riya-17: ok

lib/pbench/server/indexing_tarballs.py
```diff
@@ -65,7 +63,65 @@ def _count_lines(fname):
...
+    """
+
+    error_code = Errors(
+        ErrorCode("OK", 0, None, "Successful completion"),
+        ErrorCode("OP_ERROR", 1, False, "Operational error while indexing"),
+        ErrorCode("CFG_ERROR", 2, False, 'Configuration file not specified"'),
```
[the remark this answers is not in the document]
riya-17: yeah

lib/pbench/server/indexing_tarballs.py
```diff
@@ -166,7 +226,7 @@ def process_tb(self, tarballs):
             idxctx.templates.update_templates(idxctx.es)
         except TemplateError as e:
             idxctx.logger.error(
-                "{}, {}", error_code["TEMPLATE_CREATION_ERROR"].message, repr(e)
+                "{}, {!r}", error_code["TEMPLATE_CREATION_ERROR"].message, e
             )
             res = error_code["TEMPLATE_CREATION_ERROR"]
```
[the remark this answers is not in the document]
riya-17: It will.. yeah they produce different results

lib/pbench/server/indexing_tarballs.py
riya-17: but won't that change the mogging details? means now they are warnings after this they will be exceptions

server/bin/gold/test-5.2.txt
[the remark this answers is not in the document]
dbutenhof: Yes, this was the main point of my comment on the `emit_error` method; we're changing the logged method name; and while it's a visible difference, I'm not convinced it's worth the effort to fix. In theory, we could unwind the stack trace a level and log the caller frame rather than the current frame, but I don't see that added complication being of much value in this case.

I wouldn't want to elevate everything to exception, because it means everything becomes ERROR.

lib/pbench/server/indexing_tarballs.py
```diff
@@ -116,26 +163,30 @@ def collect_tb(self, ARCHIVE_rp, qdir):
...
+            return (error_code["GENERIC_ERROR"].value, [])
         else:
             if not tarballs:
                 idxctx.logger.info("No tar balls found that need processing")
-                return (0, [])
 
-        return (0, sorted(tarballs))
+        return (error_code["OK"].value, sorted(tarballs))
+
+    def emit_error(self, logger_method, error, exception):
```
dbutenhof, line 177: There should really be a docstring here. Explain the purpose, the interpretation of `logger_method` (though the name change definitely helps), and it is probably worth pointing out that all log messages will originate from this frame rather than from the caller.

lib/pbench/server/indexing_tarballs.py
webbnh: Yes; and that's the "price" for, as I said, "buying into" this change.

So much of what we do involves trade-offs, and this is a good example:  in order to gain the benefits of consistency and uniformity (and simplicity), we have to get over the pain of the change, put up with the "exception" output instead of the warning output, and cope with the fact that exception is now the only option (for things which use this function).  Perhaps the lack of flexibility and the other costs aren't worth it, in which case we should go with what you have.  (I'm not really in a good position to judge....)

server/bin/gold/test-5.2.txt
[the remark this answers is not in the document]
webbnh: Yeah, adding the stack unwinding thing seems like more effort and complication than is warranted for this situation.

Given that changing from warning to exception results in a change in semantics, we should probably go with what we have.  (Yeah, we could invent some additional mechanism, but, again, I don't think it's warranted.)

webbnh (dismissed):
Once there's a suitable docstring for `emit_error()`, I think it's good to go.

npalaska:
Looks good to me, but needs a rebase

lib/pbench/server/indexing_tarballs.py
dbutenhof: What I had in mind was something a bit more complete. As an example:

```
"""Helper method to write a log message in a standard format from an error code

    Args
        logger_method -- Reference to a method of a Python logger object, like idxctx.logger.warning
        error -- An error code name from the Errors collection, like "OK"
        exception -- the original exception leading to the error

    Returns
        The numeric value of the error code

    Although all log messages will appear to have originated from this method, the origin can easily be
    identified from the error code value, and this interface provides simplicity and consistency.
"""

File: server/bin/pbench-index.py [hunk 36]
```diff
@@ -44,38 +44,8 @@ def main(options, name):
...
-             balls are reported in the logs and in index status reports)
-         1 - Failed to process one or more tar balls for unknown reasons
-             (see logs)
-         2 - Missing configuration file
-         3 - Invalid configuration file
-         8 - Unable to load and process expected mapping files
-         9 - Unable to update index templates in configured Elasticsearch
-             instance
+       Status codes used by the error below are defined from the list present
+       in indexing-tarballs.
```
dbutenhof, line 48: I'm not sure this is of much use anymore. One has to read the code to see that it's referencing the `error_code` attribute of the created `Index` object, and we don't (and probably don't want to) provide any more details in this context.

We need to either expand this to be useful _without_ reading the code, or we might as well delete it. Since code importing a module (or creating an object) usually isn't responsible for documenting the details of that imported code, I'm inclined to just remove this.

webbnh: I think this is my fault.  My earlier suggestion was to remove the commend but that leaving a "reference to where those definitions can be found might be good".  I repent...let's just remove it.  (Sorry for the bad lead, Riya!)

lib/pbench/server/indexing_tarballs.py
webbnh: I concur...but I'm always suspicious when the comment is longer than the code that it's documenting.

webbnh (approved):
Once Dave's issues are resolved, I'm good with it.

server/bin/pbench-index.py
riya-17: it was a good suggestion but maybe didn't work here.

riya-17:
@dbutenhof @webbnh I have addressed both comments. Can you please take a look

webbnh (approved):
So close!  Just one small bit of text to fix (I think...).

lib/pbench/server/indexing_tarballs.py
```diff
@@ -117,26 +166,45 @@ def collect_tb(self):
...
+        """Helper method to write a log message in a standard format from an error code
+
+            Args
+                logger_method -- Reference to a method of a Python logger object,
+                                like idxctx.logger.warning
+                error -- An error code name from the Errors collection, like "OK"
+                exception -- the original exception leading to the error
+
+            Returns
+                The numeric value of the error code
```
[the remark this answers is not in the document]
riya-17: No it isn't :(. I will correct it. Sorry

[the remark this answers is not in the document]
webbnh: All's well that ends.  ;-)

npalaska (approved):

dbutenhof (approved):

PR #1787, opened 2020-09-08
Proper Tool Metadata Abstraction

Maxusmusti:
Complete tool metadata abstraction work!
(Still code review changes to iron out)
On new branch, same as PR #1778


Maxusmusti:
Created a more proper and up-to-date branch for this work.

npalaska (commented):
Its still a draft but I left few comments

agent/tool-scripts/meta.json
```diff
@@ -0,0 +1,48 @@
+{
+	"transient":{
+		"blktrace": null,
```
npalaska, line 3: What does a null represent here?

agent/util-scripts/pbench-tool-meister
```diff
@@ -361,9 +361,10 @@ class ToolMeister(object):
         else:
             return benchmark_run_dir, channel, controller, group, hostname, tools
 
-    def __init__(self, pbench_bin, params, redis_server, logger):
+    def __init__(self, pbench_bin, params, redis_server, tool_metadata, logger):
```
npalaska, line 364: if we can get tool_metadata from redis_server `redis_server.get("tool-metadata")` why do we need to pass it as a separate parameter?

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -340,6 +340,13 @@ def main(argv):
         ), f"bad channel: {resp!r}"
         assert resp["data"] == 1, f"bad data: {resp!r}"
 
+    # 2.5. Add tool metadata json to redis
+    #FIXME: ABSOLUTE PATH IS BAD (also add try/catch)
+    with Path("/opt/pbench-agent", "tool-scripts", "meta.json").open("r") as json_file:
+        #with open('/opt/pbench-agent/tool-scripts/meta.json') as json_file:
```
npalaska, line 346: we can remove this commented line I guess

agent/util-scripts/pbench-tool-meister
Maxusmusti: This will be changed in favor of a proper metadata class, currently just a relic of the old PR

agent/tool-scripts/meta.json
Maxusmusti: Once we want to add additional tool info/properties, those nulls will be replaced. For now null simply represents no additional data

portante:
@Maxusmusti, is this really ready for review?  Typically, if a PR is in `Draft` mode we don't want to officially request code reviews.  However, if it is ready to review, then let's take it out of `Draft` mode.

portante:
Also what necessitated a new PR?  Typically, there are steps one can take with `git` to fix up the proper branch to use, and/or, fix up the GitHub PR mechanism for the pull request against the new branch.

Maxusmusti:
The reason I had assigned reviewers was simply for consistency (added the same people who were marked as reviewers on the old PR). Also I was (and still am) unaware of a means by which one switches the branch that a PR originates from, so I made a fresh PR for the more properly named and updated branch (old one had some issues).

agent/util-scripts/pbench-tool-meister
```diff
@@ -67,6 +67,50 @@ tar_path = None
...
+            return self.data
+
+        if self.mode == "json":
+            with self.json.open("r") as json_file:
+                metadata = json.load(json_file)
+                self.data = metadata
+        elif self.mode == "redis":
+            meta_raw = self.redis_server.get("tool-metadata")
+            if meta_raw is None:
+                print('Metadata was never loaded.')
```
dbutenhof, line 96: Seems like this should be a log rather than just a print. Maybe even an exception, since we hopefully aren't going to try asking for redis data if it hasn't been loaded yet.

agent/util-scripts/pbench-tool-meister
```diff
@@ -81,35 +82,44 @@ class ToolMetadata:
...
         self.data = None
 
-    #TODO: ADD BETTER ERROR CHECKING TO THIS METHOD
     def getFullData(self):
         if self.data:
             return self.data
 
         if self.mode == "json":
+            if not os.path.isfile(self.json):
+                self.logger.exception('There is no tool-scripts/meta.json in given install dir')
```
dbutenhof, line 91: `logger.exception` wants to be called from an exception handler context; since this isn't, it should really be `logger.error` instead.

agent/util-scripts/pbench-tool-meister
```diff
@@ -81,35 +82,44 @@ class ToolMetadata:
...
+                return None
             with self.json.open("r") as json_file:
                 metadata = json.load(json_file)
                 self.data = metadata
         elif self.mode == "redis":
             meta_raw = self.redis_server.get("tool-metadata")
             if meta_raw is None:
-                print('Metadata was never loaded.')
+                self.logger.exception('Metadata was never loaded into redis')
```
dbutenhof, line 99: `self.logger.error` ???

agent/util-scripts/pbench-tool-meister
```diff
@@ -81,35 +82,44 @@ class ToolMetadata:
...
             metadata = json.loads(meta_str)
             self.data = metadata
         return self.data
 
-    def getPersistentTools(self):
+    def __dataCheck(self):
         if not self.data:
-            self.getFullData()
+            if not self.getFullData():
+                self.logger.exception(f"Unable to access data through {self.mode}")
```
dbutenhof, line 109: `self.logger.error` ???

agent/tool-scripts/meta.json
```diff
@@ -0,0 +1,48 @@
...
+		"systemtap": null,
+		"tcpdump": null,
+		"turbostat": null,
+		"user-tool": null,
+		"virsh-migrate": null,
+		"vmstat": null
+	},
+
+	"persistent":{
+		"node-exporter": "prometheus",
```
dbutenhof, line 45: We could work around this later by inspecting the type, but to allow for future expansion, maybe starting with a record would be more flexible? E.g., `"node-exporter": {"source": "prometheus"}` ???

agent/util-scripts/pbench-tool-data-sink
```diff
@@ -22,6 +22,10 @@ import subprocess
 import sys
 import tempfile
 
+#import importlib
+#toolmeister = importlib.import_module("pbench-tool-meister")
+from modules import metaclass
```
dbutenhof, line 27: Did you mean to use this import?

agent/util-scripts/pbench-register-tool
```diff
@@ -116,7 +116,7 @@ function usage() {
 	printf -- "\tdenoted by a leading hash, or pound (\"#\"), character.\n"
 	printf -- "\nAvailable tools:\n"
 	local tool=""
-	for tool in $(find ${pbench_bin}/tool-scripts -maxdepth 1 ! -type d ! -name '*README*' ! -name base-tool ! -name unittests -printf "%P\n" 2> /dev/null | sort); do
+	for tool in $(find ${pbench_bin}/tool-scripts -maxdepth 1 ! -type d ! -name '*README*' ! -name base-tool ! -name meta.json ! -name unittests -printf "%P\n" 2> /dev/null | sort); do
```
portante, line 119: We want to eventually replace this `for` loop with an internal CLI call which will emit the list of tools available as derived from the `meta.json` file.

Is this something we should do in this PR or wait until later?  I would like to see this done as part of this PR, but open to other opinions.

agent/util-scripts/pbench-tool-data-sink
portante: I wonder what `flake8` reports for this.

agent/util-scripts/pbench-tool-meister
```diff
@@ -67,7 +68,6 @@ tar_path = None
 # FIXME: The client response channel should be in a shared constants module.
 client_channel = "tool-meister-client"
 
-
```
portante, line 70: Running `flake8` should flag this missing line...

agent/util-scripts/pbench-tool-meister
```diff
@@ -363,7 +363,8 @@ class ToolMeister(object):
 
     def __init__(self, pbench_bin, params, redis_server, logger):
         self.logger = logger
-        self.persist_tools = ["node-exporter", "dcgm"]
+        self.tool_metadata = metaclass.ToolMetadata(self.logger, redis_server=redis_server)
+        self.persist_tools = list(self.tool_metadata.getPersistentTools().keys())
```
portante, line 367: Why not have `getPersistentTools()` return a `list`?  Seems like too much knowledge of the implementation leaking out here.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -340,6 +341,12 @@ def main(argv):
         ), f"bad channel: {resp!r}"
         assert resp["data"] == 1, f"bad data: {resp!r}"
 
+    # 2.5. Add tool metadata json to redis
+    tm_start_path = pathlib.Path(__file__).parent.absolute()
```
portante, line 345: Consider adding a `try/except` block here to capture errors from opening the file, reading from the file, and loading the JSON data.

Also, it seems `absolute` is not documented, can you use `resolve(strict=True)` instead?

portante (changes requested):
As this is a `Draft`, it goes without saying this is not ready for merging, but I'd like to know what the intent is for using `metaclass` module?

Maxusmusti:
The metaclass module is a way for any agent script to import the ToolMetadata class and anything related in the future. Was not sure how else to make it usable from both the TM and TDS, and was considering adding functionality to use it in TM-start as well

portante (changes requested):
Still looking for resolution on `list` vs `keys()` interface.

Maxusmusti:
The reason I return dicts is because (mainly just for the tds right now), it is necessary to have the collector and port data as well for each persistent tool. If we want, I could create separate functionality so that the ToolMetadata returns a list for the getTools, and has a separate function getToolProperties where you pass in the tool name and receive those values.

agent/util-scripts/modules/metaclass.py
```diff
@@ -0,0 +1,64 @@
+from pathlib import Path
+import json
+import os
+
+
+class ToolMetadata:
+    def __init__(self, logger, redis_server=None, install_path=None):
```
portante, line 7: Can we consider an interface like:

```
def __init__(self, mode, context, logger):
    self.logger = logger
    if mode not in ("redis", "json"):
        raise Exception()
    if not context:
        raise Exception()
    self.mode = mode
    if mode == "redis":
        self.redis_server = context
        self.json_file = None
    else:
        self.redis_server = None
        json_path = Path(context, "tool-scripts", "meta.json")
        try:
            self.json_file = json_path.resolve(strict=True)
        except FileNotFoundError:
            raise Exception(f"missing {json_path}")
        except Exception:
            raise
    self.data = None
```

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -340,6 +340,12 @@ def main(argv):
         ), f"bad channel: {resp!r}"
         assert resp["data"] == 1, f"bad data: {resp!r}"
 
+    # 2.5. Add tool metadata json to redis
```
portante, line 343: Here we want to use the new class, right?

portante (changes requested):

agent/util-scripts/modules/metaclass.py
```diff
@@ -0,0 +1,98 @@
+from pathlib import Path
+import json
+import os
+
+
+class ToolMetadata:
+    def __init__(self, mode, context, logger):
+        self.logger = logger
+        if mode not in ("redis", "json"):
+            raise Exception()
```
dbutenhof, line 10: The Exception() objects could have descriptions; though I'm not sure it's worth subclassing Exception it might be a bit cleaner.

agent/util-scripts/modules/metaclass.py
```diff
@@ -0,0 +1,98 @@
...
+        if not context:
+            raise Exception()
+        self.mode = mode
+        if mode == "redis":
+            self.redis_server = context
+            self.json_file = None
+        else:
+            if not self.mode == "json":
```
dbutenhof, line 18: If `mode not in ("redis", "json"):` above makes this unnecessary. Or, alternatively, if we really want to say "either redis or it doesn't matter and we read json anyway" then maybe it's the earlier check that's unnecessary! (But I'd prefer to keep the Exception and just drop this redundant check. I don't really like the implications, anyway.)

agent/util-scripts/modules/metaclass.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def __dataCheck(self):
+        if not self.data:
+            if not self.getFullData():
+                self.logger.error(f"Unable to access data through {self.mode}")
+                return 0
+        return 1
+
+    def getPersistentTools(self):
+        if self.__dataCheck():
```
dbutenhof, line 65: Instead of sprinkling the load around, it might make more sense to just do the load at the end of the constructor, and throw an exception on failure; we're always going to need it, and that way we don't need to make sure to check everywhere else.

Maxusmusti:
Everything should be ready to move out of draft and into review now, but need to figure out what's still going wrong with the unit tests...if possible would like to look at it Monday after scrum. All of them pass except the functionals with this error still:

`!   File "/var/tmp/pbench-test-utils/opt/pbench-agent/util-scripts/pbench-tool-meister-start", line 33, in <module>
!     from modules import metaclass
! ModuleNotFoundError: No module named 'modules'`

Tried appending to python path but still didn't work.

portante:
@Maxusmusti, see https://github.com/Maxusmusti/pbench/pull/4 for a possible fix.

agent/util-scripts/pbench-tool-data-sink
```diff
@@ -136,10 +139,8 @@ class BaseCollector:
 
 
 class PromCollector(BaseCollector):
-    allowed_tools = {"node-exporter": "9100", "dcgm": "8000"}
-
-    def __init__(self, benchmark_run_dir, host_tools_dict, logger):
-        super().__init__(benchmark_run_dir, host_tools_dict, logger)
+    def __init__(self, benchmark_run_dir, host_tools_dict, logger, tool_metadata):
+        super().__init__(benchmark_run_dir, host_tools_dict, logger, tool_metadata)
         self.volume = self.benchmark_run_dir + "/prom_vol"
```
portante, line 144: Can you convert this to use `Pathlib`, something like `self.volume = self.benchmark_run_dir / "prom_vol"`?

portante: And perhaps would could consider a directory hierarchy like:

```
 - <benchmark_run_dir>
   |- tools-<group>
      |- prometheus
      |- pcp
```

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -340,6 +342,13 @@ def main(argv):
...
+    tm_start_path = Path(
+        Path(__file__).resolve(), "..", "..", "tool-scripts", "meta.json"
+    ).resolve(strict=True)
```
portante, line 348: Can you we add a `try/except` to provide a command line error that is a bit more user-friendly?

portante (changes requested):
Coming along nicely ...

lib/pbench/agent/modules/metaclass.py
```diff
@@ -40,27 +44,32 @@ def getFullData(self):
...
             with self.json.open("r") as json_file:
                 metadata = json.load(json_file)
-                self.data = metadata
         elif self.mode == "redis":
             try:
                 meta_raw = self.redis_server.get("tool-metadata")
                 if meta_raw is None:
-                    self.logger.error("Metadata was never loaded into redis")
+                    if self.initialized:
+                        self.logger.error("Metadata has not been loaded into redis yet")
```
dbutenhof, line 52: This construct bothers me a bit. If we start up an agent tool meister expecting to pull Redis data, and the initial `__getInitialData` in the constructor fails, it'll fail silently, returning `None`; and a subsequent call will simply log that Metadata hasn't been loaded into redis... which isn't strictly true, nor useful.

I assume that the intent was to catch the case where we construct with JSON, and the object gets switched to `redis` mode without calling `loadIntoRedis` ... that seems like a rare edge case, and could be avoided by restricting the mode change to a method that changed mode, loaded the JSON into redis, and failed visibly if anything went wrong.

Maxusmusti:
Fixed all code review suggestions (except the "BadConstruction" Exception class naming, open to suggestions on that). All that's left is to make pbench-register-tools (bash version) pull tool list from meta.json, then should be complete.

agent/util-scripts/pbench-tool-data-sink
```diff
@@ -419,11 +422,11 @@ class ToolDataSink(Bottle):
                 persistent_tools = []
                 transient_tools = []
                 for tool_name in tools.keys():
-                    if tool_name in PromCollector.allowed_tools:
+                    if tool_name in self.tool_metadata.getPersistentTools():
                         persistent_tools.append(tool_name)
                     elif tool_name in BaseCollector.allowed_tools:
                         noop_tools.append(tool_name)
-                    else:
+                    elif tool_name in self.tool_metadata.getTransientTools():
```
dbutenhof, line 429: I like the extra validation of the `elif` instead of the old `else` ... but this raises a possible error case where `tool_name` is something else. We're no longer defaulting that to "transient" as we did before. Not entirely sure we should care since it's an unlikely error; but it might be worth at least a warning log entry "just in case".

agent/util-scripts/pbench-register-tool
```diff
@@ -116,9 +116,8 @@ function usage() {
 	printf -- "\tdenoted by a leading hash, or pound (\"#\"), character.\n"
 	printf -- "\nAvailable tools:\n"
 	local tool=""
-	for tool in $(find ${pbench_bin}/tool-scripts -maxdepth 1 ! -type d ! -name '*README*' ! -name base-tool ! -name unittests -printf "%P\n" 2> /dev/null | sort); do
-		printf -- "\t${tool}\n"
-	done
+        awk '/null/ {print $1}' ${pbench_bin}/tool-scripts/meta.json | sed 's/"/\t/' | sed 's/"://'
+        awk '/collector/ {print $1}' ${pbench_bin}/tool-scripts/meta.json | sed 's/"/\t/' | sed 's/":/ (persistent)/'
```
dbutenhof, line 120: This is awfully dependent on details, like the fact that we don't have any persistent tool attributes.

This makes me really miss the neat little `jq` utility... although it'd be easy enough if `bench-register-tool` was converted to Python.

```
jq 'keys[] as $k|.[$k]|keys[]' meta.json
```

dbutenhof (approved):
Wishful thinking ...

agent/util-scripts/pbench-register-tool
Maxusmusti: I was originally going to use jq, but it would require additional installation and I wasn't sure if we wanted that. This solution is essentially a placeholder until Chuck's work puts the script into python, in which case I can simply use the ToolMetadata class to retrieve and print tool names. Until then, it should hold true that these properties will be stable for transient/persistent tools, but also open to alternate approaches (I'm not the most well-versed in bash)

dbutenhof: > I was originally going to use jq, 

Yeah, it was more a side comment than a criticism; converting to Python is a better solution than adding an extra package dependency. I just couldn't resist!

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -340,6 +342,18 @@ def main(argv):
         ), f"bad channel: {resp!r}"
         assert resp["data"] == 1, f"bad data: {resp!r}"
 
+    # 2.5. Add tool metadata json to redis
+    try:
+        tm_start_path = Path(
+            Path(__file__).resolve(), "..", ".."
```
portante, line 348: We need to consider pulling the installation directory path from a `PbenchAgentConfiguration` object instead.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -340,6 +342,18 @@ def main(argv):
...
+        ).resolve(strict=True)
+    except FileNotFoundError:
+        raise Exception(f"missing {tm_start_path}")
```
portante, line 351: If we can't resolve the location of the installation directory, it would seem that `tm_start_path` would be defined.  Perhaps we could log the exception as an error, like: `logger.error("Unable to determine the proper installation directory, expected directory %s not found", install_dir)`?

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -340,6 +342,18 @@ def main(argv):
...
+    except Exception:
+        raise
```
portante, line 353: Instead of raising an exception, let's consider using logger.exception with a suitable message.

lib/pbench/agent/modules/metaclass.py
```diff
@@ -3,21 +3,27 @@
...
+        return repr(self.msg)
+
+
 class ToolMetadata:
     def __init__(self, mode, context, logger):
+        self.initialized = False
         self.logger = logger
         if mode not in ("redis", "json"):
-            raise Exception()
+            raise (BadConstruction("Mode was not 'json' or 'redis'"))
```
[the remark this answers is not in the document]
portante: This is an internal error of sorts, as only the code would know what to pass here.  So perhaps just an assertion?  `assert mode in ("redis", "json"), f"Logic bomb!  Unexpected mode, {mode}, encountered constructing tool meta data"`

portante (changes requested):
Coming along ...

agent/util-scripts/pbench-tool-data-sink
dbutenhof: @portante, not really sure what you mean by that. I don't see any relevant errors. We're just looping through the selected tools from Redis, and trying to categorize each into persistent/transient. There should never be any selected tools that aren't in the list. However, if there are, the "old" code (in an earlier commit) silently categorized them as transient; the new code won't categorize the tool at all. Which for example means that a "missing" sole transient tool would trigger the `elif not transient_tools:` below and won't expect an upload to the sink.

I like the `elif` rather than the `else` but if it does somehow happen through some weird skew or error, it'd be nice to see some indication, either via a log or maybe even an assert.

agent/util-scripts/pbench-register-tool
```diff
@@ -116,9 +116,8 @@ function usage() {
 	printf -- "\tdenoted by a leading hash, or pound (\"#\"), character.\n"
 	printf -- "\nAvailable tools:\n"
 	local tool=""
-	for tool in $(find ${pbench_bin}/tool-scripts -maxdepth 1 ! -type d ! -name '*README*' ! -name base-tool ! -name meta.json ! -name unittests -printf "%P\n" 2> /dev/null | sort); do
-		printf -- "\t${tool}\n"
-	done
+	cat ${pbench_bin}/tool-scripts/meta.json | python3 -c 'from __future__ import print_function; import sys, json; [print("\t" + tool) for tool in json.load(sys.stdin)["transient"].keys()]'
+	cat ${pbench_bin}/tool-scripts/meta.json | python3 -c 'from __future__ import print_function; import sys, json; [print("\t" + tool + " (persistent)") for tool in json.load(sys.stdin)["persistent"].keys()]'
```
dbutenhof, line 120:
```
python3 -c "import sys, json; meta = json.load(open(sys.argv[1])); [print(f'\t{tool}') for tool in (*meta['transient'].keys(), *meta['persistent'].keys()) ]" ${pbench_bin}/tool-scripts/meta.json
```

agent/util-scripts/pbench-tool-data-sink
```diff
@@ -221,21 +221,20 @@ class PromCollector(BaseCollector):
...
+            "--remove-files",
+            "--exclude",
+            "prometheus/prometheus_data.tar.gz",
             "-zcvf",
-            f"{self.benchmark_run_dir}/tools-{self.tool_group}/prometheus/prometheus_data.tar.gz",
+            f"{self.volume}/prometheus_data.tar.gz",
             "-C",
-            f"{self.benchmark_run_dir}/",
-            "prom_vol",
+            f"{self.volume}/../",
```
dbutenhof, line 232: This might be a bit easier to read if you had a `self.tool_group_dir` because seeing this construct makes me want to pop back up to find the original definition to see exactly what the parent directory is. Also, although it's a technicality in our case since it's unlikely we'd ever support Windows, it's cleaner to keep Python code free of OS dependencies, which is one reason why `Path() / 'directory'` is better than ` + '/directory'`

agent/util-scripts/pbench-tool-data-sink
```diff
@@ -299,8 +298,7 @@ class ToolDataSink(Bottle):
         self.web_server_thread = None
 
     def run(self):
-        """run - Start the Bottle web server running and the watcher thread.
-        """
+        """run - Start the Bottle web server running and the watcher thread."""
```
dbutenhof, line 301: The human brain is a weird thing. I skimmed past this without paying attention, and my brain told me it was something about a bottle and watching the water run...

OK, maybe it's just me. ;-)

portante (approved):
Ship it!

portante (changes requested):
Actually, while it looks good, it needs a rebase on `tool-meister` because I had rebased `tool-meister` on `master`.

While you are at it, could you also review the outstanding comments that are not marked "resolved" and make sure they have addressed with a change or with a comment explaining no change was made, and then mark them as "resolved"?

Thanks!

agent/util-scripts/pbench-tool-meister-start
Maxusmusti: Is this an existing class or something else to be implemented?

Maxusmusti:
Will be making the last of the changes now, only thing I am unsure about is the PbenchAgentConfiguration object. Does that already exist somewhere (aka the PbenchAgentConfig class that exists in the pbench.agent module), or is that something else you would like implemented?

agent/util-scripts/pbench-tool-meister-start
portante: Existing, see https://github.com/distributed-system-analysis/pbench/blob/tool-meister/lib/pbench/agent/__init__.py#L14.

Maxusmusti:
Nevermind, I figured out how to use the PbenchAgentConfig class

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -353,6 +356,17 @@ def main(argv):
         ), f"bad channel: {resp!r}"
         assert resp["data"] == 1, f"bad data: {resp!r}"
 
+    # 2.5. Add tool metadata json to redis
+    try:
+        inst_dir = PbenchAgentConfig(os.environ["_PBENCH_AGENT_CONFIG"]).pbench_install_dir
+        tm_start_path = Path(inst_dir).resolve(strict=True)
+    except FileNotFoundError:
+        logger.error("Unable to determine proper installation directory, {inst_dir} not found")
```
portante, line 364: Here I think you need to `return 1` or some error number value to indicate the `pbench-tool-meister-start` command was not successful.

portante: One more point, when using the python logging sub-system, we typically don't want to fully format strings ahead of time.  The logging sub-system gives us a way to do that at run time to avoid the formatting if the log message is not going to be emitted.

In this case it is not a big deal because we are typically always going to emit the message, but the discipline is worth maintaining to avoid making that mistake with `.info` or `.debug` log statements.

Consider: `logger.error("Unable to determine proper installation directory, '%s' not found", inst_dir)`

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -353,6 +356,17 @@ def main(argv):
...
+    except Exception as exc:
+        loger.exception(f"Failed to resolve expected installation: {exc}")
```
portante, line 366: `s/loger/logger/`

And consider using a message like the following without emitting the exception stack trace, like:

```
logger.error("Unexpected error encountered resolving installation directory: '%s'", exc)
```

lib/pbench/agent/modules/metaclass.py
```diff
@@ -0,0 +1,106 @@
+from pathlib import Path
+import json
+import os
+
+
+class BadConstruction(Exception):
```
portante, line 6: All you need in Python for sub-classing `Exception` is:

```
class BadConstruction(Exception):
    pass
```

The constructor and string conversion is handled by default.

portante: Can we use the name for the class as something like, `ToolMetadataExc`?

lib/pbench/agent/modules/metaclass.py
```diff
@@ -0,0 +1,106 @@
...
+
+class ToolMetadata:
+    def __init__(self, mode, context, logger):
+        self.logger = logger
+        assert mode in (
+            "redis",
+            "json",
+        ), f"Logic bomb! Unexpected mode, {mode}, encountered constructing tool meta data"
+        if not context:
+            raise (BadConstruction("No proper context given"))
```
portante, line 22: Perhaps change to `assert context, f"Logic bomb!  No context given on ToolMetadata object construction"`?

lib/pbench/agent/modules/metaclass.py
```diff
@@ -0,0 +1,106 @@
...
+        if mode == "redis":
+            self.redis_server = context
+            self.json_file = None
+        else:
+            self.redis_server = None
+            json_path = Path(context, "tool-scripts", "meta.json")
+            try:
+                self.json = json_path.resolve(strict=True)
+            except FileNotFoundError:
+                raise Exception(f"missing {json_path}")
```
portante, line 33: Perhaps `s/Exception/ToolMetadataExc`?

lib/pbench/agent/modules/metaclass.py
```diff
@@ -0,0 +1,106 @@
...
+        elif self.mode == "redis":
+            try:
+                meta_raw = self.redis_server.get("tool-metadata")
+                if meta_raw is None:
+                    self.logger.error("Metadata has not been loaded into redis yet")
+                    return None
+                meta_str = meta_raw.decode("utf-8")
+                metadata = json.loads(meta_str)
+            except Exception:
+                self.logger.error("Failure to reach redis server")
```
portante, line 56: Seems like we should refine this a bit to tease out different failure cases.  Perhaps something like:

```
try:
    meta_raw = self.redis_server.get("tool-metadata")
except Exception:
    self.logger.exception("Failure to fetch tool metadata from the Redis server")
    raise
else:
    if meta_raw is None:
        self.logger.error("Metadata has not been loaded into Redis server yet")
        return None
try:
    metadata = json.loads(meta_raw.decode("utf-8"))
except Exception as exc:
    self.logger.error("Bad metadata loaded into Redis server, '%s', json=%r", exc, meta_raw)
    return None

File: lib/pbench/agent/modules/metaclass.py [hunk 37]
```diff
@@ -0,0 +1,106 @@
...
+            return self.data["persistent"][tool]
+        elif tool in self.data["transient"].keys():
+            return self.data["transient"][tool]
+
+    def loadIntoRedis(self, info):
+        if self.mode == "redis":
+            try:
+                self.json = Path(info).resolve(strict=True)
+            except FileNotFoundError:
+                raise Exception(f"missing {info}")
```
portante, line 93: Perhaps `s/Exception/ToolMetadataExc/`?

portante (changes requested):
Still reviewing ...

agent/util-scripts/pbench-tool-meister-start
portante: Further, I am not entirely sure, but the code for `PbenchConfig` and `PbenchAgentConfig` classes don't appear to explicitly raise a `FileNotFoundError`, but neither to they prevent that error from being raised for their operation.

It might be better to consider handling the exception separately, something like:

```
try:
    inst_dir = PbenchAgentConfig(...).pbench_install_dir
except BadConfig as exc:
    logger.error("%s", exc)
    return 1
except Exception:
    logger.error("Unexpected error encountered logging pbench agent configuration", exc)
    return 1

try:
    tm_start_path = Path(inst_dir).resolve(strict=True)
except ...
```

portante (changes requested):
Finished this pass of reviews on the code.

portante:
PR #1821 should provide a fix for the failing unit tests.

portante:
@Maxusmusti, now that #1821 is merge into `master`, and the `tool-meister` branch has been rebased on `master`, let's rebase this PR and see if what test failures remain.

Maxusmusti:
Not sure why the flood of commits, but all the changes are exactly as they should be (should definitely use "squash and merge")

agent/util-scripts/gold/pbench-register-tool/test-46.txt
```diff
@@ -19,7 +19,6 @@ Available tools:
 	blktrace
 	bpftrace
 	cpuacct
-	dcgm
```
dbutenhof, line 22: This actually made me think: should we sort the list of keys so dcgm and node-exporter don't get moved to the end of the list?

dbutenhof (approved):

agent/util-scripts/gold/pbench-register-tool/test-46.txt
portante: @dbutenhof, just do confirm in writing, do you think we can do that later, or would you want to see the sorting done in this PR?

portante (changes requested):
Looks good!  Just needs a rebase and a squash.

agent/util-scripts/gold/pbench-register-tool/test-46.txt
dbutenhof: @portante @Maxusmusti I approved the PR, which I figured would be clear enough; but for the record, yes, sure. It's a pretty trivial UX issue; it'd be more consistent and maybe a bit more clear, but I doubt most people really pay attention to the full list of tools in the usage.

Maxusmusti:
It should already be rebased on the tool-meister branch, unless there was another new merge today. Should I squash the last chunk of updates and force push to metadata, or would it be better to squash through the "squash and merge" option

agent/util-scripts/gold/pbench-register-tool/test-46.txt
Maxusmusti: I can easily change that once Chuck's work has been integrated, I'll be sure to sort in the future

Maxusmusti:
Ok now it's ACTUALLY rebased, it seems as though I had botched it earlier

portante (approved):

PR #1793, opened 2020-09-15
Making a server RPM

ndokos:
The process is:
    
        cd .../server/rpm
        make copr-test
    
This builds RPMs on everything except RHEL7: the RHEL7 COPR build  complains about rh-python36 being filtered out by exclude filtering.  However, a build on a RHEL7 VM succeeds.
    
The new approach:
    
 - run `setup install' to generate a PKG-INFO and the pbench-config and pbench-server scripts.
    
 - run `make install-all' to a temp directory in the server to get all  the server bits - that installs enough of the python infrastructure to allow a `setup install' in the spec file.
    
 - run `make install' to the same temp directory in the web-server to get  all the web-server bits.
    
 - create a tarball in the temp directory.
   
 - make an SRPM and send it to COPR for all the builds.
    
 - the spec file now does a `setup install' to get the python bits and  a `make install' to get everything else.
    
 - the ansible roles have been enhanced to deal with the pbench-server systemd set-up.


server/rpm/Makefile
```diff
@@ -0,0 +1,76 @@
+# Making a pbench-agent RPM requires a few steps
```
portante, line 1: s/pbench-agent/pbench-server/

server/rpm/pbench-server.spec
```diff
@@ -0,0 +1,241 @@
...
+
+(cd ./web-server;
+ %{__make} clone DESTDIR=%{?buildroot}/%{static} INSTALL="%{__install} -p")
+
+cp -a %{?buildroot}/%{static}/package.json %{?buildroot}/%{installdir}
+
+# clean up unneeded files
+cd %{buildroot}/%{installdir}
+rm -rf Makefile html/static/Makefile \
+   lib/python3.7/site-packages/pbench/agent \
```
portante, line 84: Should this be `server` instead of `agent`?

server/rpm/pbench-server.spec
```diff
@@ -0,0 +1,241 @@
...
+/%{installdir}/bin/pbench-sync-satellite.sh
+/%{installdir}/bin/pbench-trampoline
+/%{installdir}/bin/pbench-unpack-tarballs.sh
+/%{installdir}/bin/pbench-verify-backup-tarballs.py
+/%{installdir}/bin/pbench-check-tb-age
+/%{installdir}/bin/pbench-check-tb-age.py
+/%{installdir}/bin/pbench-cull-unpacked-tarballs
+/%{installdir}/bin/pbench-cull-unpacked-tarballs.py
+
+/%{installdir}/lib/python3.7/site-packages/pbench/common/__init__.py
```
portante, line 204: Shouldn't this be `python3.6`?

server/rpm/pbench-server.spec
```diff
@@ -0,0 +1,241 @@
...
+/%{installdir}/lib/python3.7/site-packages/pbench/common/__pycache__/__init__.cpython-37.pyc
```
portante, line 205: And why are we installing `__pycache__` files?

server/rpm/seqno
```diff
@@ -0,0 +1 @@
+91
```
portante, line 1: Is `server/rpm/seqno` really a file we want to commit to the tree?

portante (changes requested):
Looks good, just a couple of fix-ups.

server/rpm/pbench-server.spec
ndokos: No - basically running `setup.py` installs a bunch of extra things and this code gets rid of them - hack, cough.

server/rpm/pbench-server.spec
ndokos: Ah, yes - it's 3.7 on my Fedora box, but it needs to be generalized to work everywhere.

server/rpm/pbench-server.spec
ndokos: I'll get rid of them.

server/rpm/seqno
ndokos: No.

server/rpm/Makefile
ndokos: OK

ndokos:
The approach is wrong: I cannot run setup.py in the RPM Makefile; I have to run it in the install phase of the spec file.

server/rpm/pbench-server.spec
```diff
@@ -0,0 +1,241 @@
...
+Requires:  policycoreutils-python-utils
+%endif
+
+Requires: npm
+
+# installdir has to agree with the definition of install-dir in
+# pbench-server.cfg, but we can't go out and pluck it from there,
+# because we don't know where the config file is. Note that we omit
+# the initial / - it is added in every use below.  IMO, that's more
+# readable since it appear in the middle of the path in all cases,
```
portante, line 44: s/appear/appears

server/rpm/pbench-server.spec
```diff
@@ -0,0 +1,241 @@
...
+# *except* in the %files section (and one instance in the %post
+# and %postun sections).
+
+%define installdir opt/pbench-server
```
portante, line 48: Should we just document that the pbench-server (and for that matter, the pbench-agent) RPM installs have the spec file hard-coded to be `/opt/pbench-server`?  If so, then we DO know where the configuration file is, `/opt/pbench-server/lib/config/pbench-server.cfg`.

portante (changes requested):
A few more comments...

server/Makefile
```diff
@@ -55,33 +54,43 @@ binlinks = \
...
+install-bin: install-binbase install-bintools install-binscripts
 
 install-binbase:
 	for f in ${binbase} ;do \
 	    ${COPY} bin/$$f ${BINDIR} \
 	;done
 
 install-bintools:
-	for f in ${bintools} ;do \
+	-for f in ${bintools} ;do \
```
portante, line 93: Why ignore success/failure here?

portante (changes requested):
Looks good; still looking for past comments to be addressed somehow.

server/rpm/pbench-server.spec
ndokos: Six-of-one ... I'm changing my mind about it every time you raise the issue. Although this is the default, we *can* change it on the command line, by modifying the `rpmbuild` invocation in the Makefile to `rpmbuild --define 'installdir foo/bar'`, and that *might* be useful when testing; practically speaking we have never availed ourselves of this capability. The ansible playbooks alsio set the default value  to  `/opt/pbench-server`,  but it too can be overridden (IIRC) e.g. in the inventory. So if we want to change it, we need to change it in two places.

So we can certainly say in the doc that the installation directory is `/opt/pbench-server` and say nothing about changing it. Or we can say the default installation directory is `/opt/pbench-server` and tell how to change it if necessary. Either is fine, but it does not imply any code changes - right?

server/Makefile
ndokos: `pbench-config` and `pbench-server` may not exist: they are created by the setup. I can split them off into a separate target.

ndokos: Actually, that may be a remnant of an aborted change: let me check.

ndokos: Yup, it was a remnant: fixed.

server/rpm/pbench-server.spec
portante: Seems like the answer is to modify the ansible playbooks to assume (and verify) that the pbench-server RPM installation will always be in `/opt/pbench-server`, flagging a problem when the `install-dir` prefix does not match.

Another option is to remove `install-dir` from the pbench-server.cfg file, and derive the `install-dir` value in the ansible playbooks from an environment variable.  Then have the environment variable set and recorded by the RPM installation.

Regardless, seems like we should not tackle this in this PR.  Perhaps you could open an issue to track for the future?

portante: Filed #1803 to track.

portante (changes requested):
Looks good, we just want to squash to a single commit when ready.

server/rpm/pbench-server.spec
```diff
@@ -183,6 +189,14 @@ fi
 /%{installdir}/lib/pbench/server/utils.py
 /%{installdir}/lib/pbench/server/api/__init__.py
 /%{installdir}/lib/pbench/server/s3backup/__init__.py
+/%{installdir}/lib/pbench/cli/__init__.py
+/%{installdir}/lib/pbench/cli/agent/__init__.py
+/%{installdir}/lib/pbench/cli/agent/commands/__init__.py
+/%{installdir}/lib/pbench/cli/agent/commands/log/__init__.py
+/%{installdir}/lib/pbench/cli/agent/commands/results/__init__.py
+/%{installdir}/lib/pbench/cli/agent/options.py
```
dbutenhof, line 197: Do we need these "agent" files for the server?

server/Makefile
```diff
@@ -109,13 +109,20 @@ install-lib:
 	${COPY} -r lib ${DESTDIR}
 
 # copy enough to be able to run setup again in the destination directory
-install-server-lib:
+install-server-lib:
```
portante, line 112: Trailing whitespace added ...

server/rpm/pbench-server.spec
```diff
@@ -110,31 +108,33 @@ rm -rf Makefile %{static}/Makefile \
...
-pip3 install elasticsearch1 urllib3 boto3 ujson
+echo "python3 -m pip install -r requirements.txt"
+python3 -m pip install -r requirements.txt
 %endif
 
 # on Fedora, we need to pip-install elasticsearch1: only the latest
 # version is packaged as an RPM in the Fedora repos.
 %if 0%{?fedora}
-pip3 install elasticsearch1 urllib3 boto3 ujson
+echo "python3 -m pip install -r requirements.txt"
```
portante, line 136: Any way to make this be `%if 0%{?rhel} == 8 or 0%{?fedora}`?

server/rpm/pbench-server.spec
```diff
@@ -166,6 +166,9 @@ fi
 /%{installdir}/%{static}/VERSION
 /%{installdir}/%{static}/package.json
 /%{installdir}/package.json
+/%{installdir}/requirements.txt
```
portante, line 169: Since we only need these files during the `%post` phase, can we remove then at the end of the `%post` phase and then not include them here?

portante (changes requested):
Looks good, a few comments to consider.

server/rpm/pbench-server.spec
```diff
@@ -110,31 +104,33 @@ rm -rf Makefile %{static}/Makefile \
...
 
 %post
 
 chown -R pbench.pbench /%{installdir}
 
+cd /%{installdir}
+
 # on RHEL7 we use scl to get python3
 %if 0%{?rhel} == 7
 # install python3-elasticsearch
```
dbutenhof, line 118: I guess this comment really wasn't accurate before since you listed boto3 and ujson as well as just elasticsearch1; it's even less accurate now that you're referencing `requirements.txt`. Maybe just "install Python dependencies"?

server/rpm/pbench-server.spec
ndokos: No - they are gone.

server/rpm/pbench-server.spec
ndokos: I've left them separate so far because in some case in the past, rhel8 and fedora needed different treatment. I'll squash them at the very end, if they remain the same until that point.

server/rpm/pbench-server.spec
ndokos: Not sure - I think they are installed even if the post phase removes them, so it might complain about files installed but not enumerated (or whatever the complaint is). I'll try it out and see.

server/rpm/pbench-server.spec
ndokos: Yes, I'll go through and scrub them.

server/VERSION
```diff
@@ -1,2 +1 @@
-0.69.2
-
+0.69.4
```
dbutenhof, line 1: Possibly doesn't matter that there's no newline at the end; but GitHub is complaining, and some other utilities might as well.

ndokos: Fixed.

portante (changes requested):
Seems okay, still in `Draft` mode, are there more changes coming?

PR #1796, opened 2020-09-16
Switch indexer.py to pyesbulk package

dbutenhof:
This is the basic change to shift `indexer.py` to using Pete's external pyesbulk Python package rather than directly interacting with Elasticsearch.

This is going to fail unit testing due to the lack of a pyesbulk wheel (`pyesbulk-1.0.0-py3-none-any.whl` for example). Just for reference (and reproducibility) I've left the `--find-links` in `requirements.txt` that allowed tox to locate a locally built copy of the wheel; that'll be removed when no longer necessary, before I promote this PR to reviewable.


portante (changes requested):
Great start!  I'll work on packaging py-es-bulk.

portante:
> Great start! I'll work on packaging py-es-bulk.

Published: https://pypi.org/project/pyesbulk/

dbutenhof:
BTW, good for review; but don't merge because I still haven't been able to stand up enough of a "modern" server environment to actually test this.

dbutenhof:
Testing on a VM shows that pbench-index, using the new pyesbulk package, successfully creates the proper prefixed templates and index documents. Seems like we're ready to go.

portante (changes requested):
Looks good!  Can you squash down to one commit?

portante (approved):
Ship it!

npalaska (approved):
Looks fine to me

Issue #1776, opened 2020-09-03
indexer.py cleanup


dbutenhof:
Notes from Pete's walkthrough, which can probably be resolved all at once. (Or, if not, we can spin off separate issues.)

- [ ] _make_source_json (about line 1521) can be streamlined. It iterates searching for a results.json file; but we only want the one at the top level.

- [ ] ResultData.gen_sources is an @staticmethod but is passed `self`... it should be a normal object method

- [ ] en_sources uses literal list [“latency”, “resource”, “throughput”] … could arguably be constants and constant list for better clarity

- [ ] Improve comment at 1805 “Generate a sequence of JSON documents…”

PR #1797, opened 2020-09-17
indexer.py cleanup 

npalaska:
This PR tries to address some or all the indexer.py cleanup issues mentioned in the https://github.com/distributed-system-analysis/pbench/issues/1776

### DIFF
--- a/lib/pbench/server/indexer.py
+++ b/lib/pbench/server/indexer.py
@@ -1560,7 +1560,7 @@ class ResultData(PbenchData):
             # FIXME: this is a "traffigen" specific field; we take care of
             # this here because we don't have a mechanism for source-data-
             # specific transformations external to the code.
-            bm_md["uid_tmpl"] = bm_md["trafficgen_uid"]
+            bm_md["trafficgen_uid_tmpl"] = bm_md["trafficgen_uid"]
         except KeyError:
             # Ignore a missing 'trafficgen_uid' field
             pass
@@ -1568,8 +1568,8 @@ class ResultData(PbenchData):
             # Now that we have a proper "template" field, generate the actual
             # UID from the template using the existing metadata we have
             # collected.
-            bm_md["uid"] = ResultData.expand_template(
-                bm_md["uid_tmpl"], bm_md, run=self.run_metadata
+            bm_md["trafficgen_uid"] = ResultData.expand_template(
+                bm_md["trafficgen_uid_tmpl"], bm_md, run=self.run_metadata
             )
 
         iteration = _dict_const(
--- a/lib/pbench/server/indexer.py
+++ b/lib/pbench/server/indexer.py
@@ -4,7 +4,6 @@ result tar balls.
 
 import copy
 import csv
-import glob
 import hashlib
 import json
 import logging
@@ -1327,10 +1326,10 @@ class ResultData(PbenchData):
         """
         result_json = Path(
             self.ptb.extracted_root, self.ptb.dirname, "result.json"
-        ).resolve(strict=True)
+        )
         try:
             # Read the file and interpret it as a JSON document.
-            with open(result_json) as fp:
+            with result_json.open() as fp:
                 results = json.load(fp)
         except Exception as e:
             self.logger.warning(
--- a/lib/pbench/server/indexer.py
+++ b/lib/pbench/server/indexer.py
@@ -15,6 +15,7 @@ import socket
 import sys
 import tarfile
 import errno
+from pathlib import Path
 from collections import Counter
 from configparser import ConfigParser
 from configparser import Error as ConfigParserError
@@ -141,9 +142,9 @@ class PbenchTemplates:
 
     def __init__(self, basepath, idx_prefix, logger, known_tool_handlers=None, _dbg=0):
         # Where to find the mappings
-        MAPPING_DIR = os.path.join(os.path.dirname(basepath), "lib", "mappings")
+        MAPPING_DIR = Path(os.path.dirname(basepath), "lib", "mappings")
         # Where to find the settings
-        SETTING_DIR = os.path.join(os.path.dirname(basepath), "lib", "settings")
+        SETTING_DIR = Path(os.path.dirname(basepath), "lib", "settings")
 
         self.versions = {}
         self.templates = {}
@@ -154,7 +155,7 @@ class PbenchTemplates:
 
         # Pbench report status mapping and settings.
         server_reports_mappings = {}
-        mfile = os.path.join(MAPPING_DIR, "server-reports.json")
+        mfile = Path(MAPPING_DIR, "server-reports.json")
         key, mapping = self._fetch_mapping(mfile)
         try:
             idxver = mapping["_meta"]["version"]
@@ -170,7 +171,7 @@ class PbenchTemplates:
             )
         server_reports_mappings[key] = mapping
         server_reports_settings = self._load_json(
-            os.path.join(SETTING_DIR, "server-reports.json")
+            Path(SETTING_DIR, "server-reports.json")
         )
 
         ip = self.index_patterns["server-reports"]
@@ -194,7 +195,7 @@ class PbenchTemplates:
         # only one settings file below.
         run_mappings = {}
         idxver = None
-        for mapping_fn in glob.iglob(os.path.join(MAPPING_DIR, "run*.json")):
+        for mapping_fn in MAPPING_DIR.glob("run*.json"):
             key, mapping = self._fetch_mapping(mapping_fn)
             try:
                 idxver_val = mapping["_meta"]["version"]
@@ -218,7 +219,7 @@ class PbenchTemplates:
                     )
                 )
             run_mappings[key] = mapping
-        run_settings = self._load_json(os.path.join(SETTING_DIR, "run.json"))
+        run_settings = self._load_json(Path(SETTING_DIR, "run.json"))
 
         ip = self.index_patterns["run-data"]
         idxname = ip["idxname"]
@@ -243,7 +244,7 @@ class PbenchTemplates:
 
         # Next we load all the result-data mappings and settings.
         result_mappings = {}
-        mfile = os.path.join(MAPPING_DIR, "result-data.json")
+        mfile = Path(MAPPING_DIR, "result-data.json")
         key, mapping = self._fetch_mapping(mfile)
         try:
             idxver = mapping["_meta"]["version"]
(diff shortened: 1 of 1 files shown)


lib/pbench/server/indexer.py
```diff
@@ -447,8 +447,7 @@ def dump_templates(self):
         sys.stdout.flush()
 
     def update_templates(self, es, target_name=None):
-        """Push the various Elasticsearch index templates required by pbench.
-        """
+        """Push the various Elasticsearch index templates required by pbench."""
```
dbutenhof, line 450: This line might be the one that's too long for black's approval? Try reformatting manually with `black .` (which will check and fix any nonconforming file).

lib/pbench/server/indexer.py
```diff
@@ -1536,121 +1539,113 @@ def _make_source_json(self):
...
+        result_json = os.path.join(
+            self.ptb.extracted_root, self.ptb.dirname, "result.json"
+        )
+        try:
+            # Read the file and interpret it as a JSON document.
+            with open(result_json) as fp:
+                results = json.load(fp)
+        except Exception as e:
+            self.logger.warning(
+                "result-data-indexing: encountered invalid JSON file," " {}: {:r} ({})",
```
dbutenhof, line 1551: Or this might be where `black` is unhappy. First off, if you're joining the continuation string onto the same line, you don't need to terminate the first string, which would save you 3 characters. Not sure that's enough, though.

lib/pbench/server/indexer.py
```diff
@@ -1536,121 +1539,113 @@ def _make_source_json(self):
...
+                iter_name = iteration["iteration_name"]
+                iter_data = iteration["iteration_data"]
+            except KeyError:
                 self.logger.warning(
-                    "result-data-indexing: encountered unexpected"
-                    " JSON file format, %s ({})",
+                    "result-data-indexing: could not find"
+                    " iteration data in JSON file, {} ({})",
                     result_json,
                     self.ptb._tbctx,
```
dbutenhof, line 1580: I shouldn't worry about this since it's unusual and weird, and trying to identify a specific broken iteration would be awkward... but somehow it makes me uncomfortable. So instead of offering specific criticism or correction, I'll simply make myself feel better by sharing that it makes me vaguely uncomfortable. ;-)

dbutenhof (commented):
Looks like a good cleanup; too bad `black` isn't happy.

lib/pbench/server/indexer.py
npalaska: Weirdly this reformat happened after I ran `Black indexer.py`

lib/pbench/server/indexer.py
npalaska: yeah weird I am not sure why black is throwing error in the unittest, when in fact it says nothing to reformat when I run the indexer.py file with black

lib/pbench/server/indexer.py
dbutenhof: What version of black are you running interactively? Various versions have sometimes incompatible ideas of what's "good", which is why I locked our lint version of black to 19.10b0.

lib/pbench/server/indexer.py
```diff
@@ -1536,121 +1537,113 @@ def _make_source_json(self):
...
-                # sample result.json - we ignore each sample result.json since
-                # each sample is contained in the experiment level
-                # result.json.
-                continue
-            elif dirname != self.ptb.dirname:
-                # iteration result.json - there is no value in processing the
-                # result.json file at the iteration level because everything
-                # it contains is also in the experiment level's result.json.
-                continue
+        result_json = os.path.join(
```
portante, line 1540: Consider using `Pathlib` here.

portante (changes requested):
Perhaps make an initial commit which reformats using `black`, and then perform your changes on top?

lib/pbench/server/indexer.py
npalaska: ahh good catch thanks @dbutenhof I have black version 20.8b1 inside the virtual env but had 19.10b0 outside venv, should have updated it inside venv, yikes

lib/pbench/server/indexer.py
npalaska: I wonder if I should replace all the `os.path` with `pathlib.Path`?

npalaska:
@portante or @dbutenhof can anyone of you help me recall what we discussed about point 5 of the issue https://github.com/distributed-system-analysis/pbench/issues/1776 I am sort of confused what cleanup need to be done there

dbutenhof:
> @portante or @dbutenhof can anyone of you help me recall what we discussed about point 5 of the issue #1776 I am sort of confused what cleanup need to be done there

Way down in `_handle_iteration`,

```
        try:
            # Attempt to copy the `trafficgen_uid` field to a renamed field
            # adding the "_tmpl" suffix to indicate it is just a template for
            # forming the trafficgen UID value.
            #
            # FIXME: this is a "traffigen" specific field; we take care of
            # this here because we don't have a mechanism for source-data-
            # specific transformations external to the code.
            bm_md["trafficgen_uid_tmpl"] = bm_md["trafficgen_uid"]
        except KeyError:
            # Ignore a missing 'trafficgen_uid' field
            pass
        else:
            # Now that we have a proper "template" field, generate the actual
            # UID from the template using the existing metadata we have
            # collected.
            bm_md["trafficgen_uid"] = ResultData.expand_template(
                bm_md["trafficgen_uid_tmpl"], bm_md, run=self.run_metadata
            )
```

Pete commented that we shouldn't be doing that; and we should really be mapping `"trafficgen_uid"` to `"uid_tmpl"` and then creating `"uid`" instead of `"trafficgen_uid"`.

I wondered if there might be upstream complications to this change, if someone's depending on the old mapping; but I wanted to record Pete's comments so we could dredge them out later. I guess this is that time.

npalaska:
Thanks @dbutenhof , Yes I thought about that when I read the point 5, wouldnt it break something if we change the template

lib/pbench/server/indexer.py
```diff
@@ -1774,16 +1774,16 @@ def _handle_iteration(self, iter_data, iter_name, iter_number, result_json):
             # FIXME: this is a "traffigen" specific field; we take care of
             # this here because we don't have a mechanism for source-data-
             # specific transformations external to the code.
-            bm_md["trafficgen_uid_tmpl"] = bm_md["trafficgen_uid"]
+            bm_md["uid_tmpl"] = bm_md["trafficgen_uid"]
```
npalaska, line 1777: Does this mean we can remove the `trafficgen_uid` field from json?

dbutenhof: @portante can we get an official experienced interpretation here? This seems to be what you were driving for initially with your comment during the presentation; but how sure are we this won't break anyone down the line?

portante: I think we should address this kind of change as a dedicated PR.  @npalaska, can you file an issue to track this?

lib/pbench/server/indexer.py
```diff
@@ -1536,116 +1539,109 @@ def _make_source_json(self):
...
-                # result.json.
-                continue
-            elif dirname != self.ptb.dirname:
-                # iteration result.json - there is no value in processing the
-                # result.json file at the iteration level because everything
-                # it contains is also in the experiment level's result.json.
-                continue
+        result_json = pathlib.Path(
+            self.ptb.extracted_root, self.ptb.dirname, "result.json"
+        ).resolve(strict=True)
```
portante, line 1544: Here you don't need `.resolve(strict=True)` because you are going to `open` it below, while implies a "resolve" operation.

lib/pbench/server/indexer.py
```diff
@@ -1536,116 +1539,109 @@ def _make_source_json(self):
...
+        try:
+            # Read the file and interpret it as a JSON document.
+            with open(result_json) as fp:
```
portante, line 1547: Now that `result_json` is a `Pathlib` object, perhaps you could use the `open()` method, like `with result_json.open() as fp:`?

portante (changes requested):
Let's move the `trafficgen_uid` change to another commit, since that one changes the data model while the other ones don't.

lib/pbench/server/indexer.py
npalaska: Will do

lib/pbench/server/indexer.py
```diff
@@ -15,6 +15,7 @@
 import sys
 import tarfile
 import errno
+import pathlib
```
riya-17, line 18: We can import "from pathlib import Path" this way we can avoid using "pathlib.Path" in code and we can directly use just "Path"

lib/pbench/server/indexer.py
```diff
@@ -1536,116 +1539,109 @@ def _make_source_json(self):
...
-                # sample result.json - we ignore each sample result.json since
-                # each sample is contained in the experiment level
-                # result.json.
-                continue
-            elif dirname != self.ptb.dirname:
-                # iteration result.json - there is no value in processing the
-                # result.json file at the iteration level because everything
-                # it contains is also in the experiment level's result.json.
-                continue
+        result_json = pathlib.Path(
```
riya-17, line 1542: instead of "pathlib.Path" we can use "Path"

lib/pbench/server/indexer.py
```diff
@@ -1536,116 +1539,109 @@ def _make_source_json(self):
...
-                            " {}: {} ({})",
-                            iteration["iteration_name_format"],
-                            result_json,
-                            exc,
-                            self.ptb._tbctx,
-                        )
-                        self.counters["bad_iteration_name_fmt"] += 1
-                        continue
-                    iter_dir = os.path.join(self.ptb.extracted_root, dirname, iter_name)
+                iter_dir = os.path.join(
```
riya-17, line 1605: We can use "Path" here and channge conditions accordingly.

lib/pbench/server/indexer.py
```diff
@@ -1536,116 +1539,109 @@ def _make_source_json(self):
...
+                    self.ptb.extracted_root, self.ptb.dirname, iter_name
+                )
+                if not os.path.isdir(iter_dir):
```
riya-17, line 1608: here we can consider `if not iter_dir.is_dir()`

lib/pbench/server/indexer.py
npalaska: filed an issue to track https://github.com/distributed-system-analysis/pbench/issues/1852

npalaska:
Closing this PR as it's not required anymore

Issue #1364, opened 2019-09-26
Script to add the VERSION-BUILD-COMMIT_ID string to config file on installation.


ndokos:
The RPM build process provides files in the installation root called VERSION, SEQNO, SHA1 that contain the strings out of which we build the complete version string (e.g. v0.64-1ga1b2c3). We need this string to appear in the indexed data in order to be able to trace back problems. To do this, we'll add a script that will be invoked as part of the server "activation" process; the script will construct the string and add it to the config file as the value of the `commit_id` option. The indexer and/or the pbench-report-status script will then be able to retrieve it from there.

portante:
@ndokos, did we already address this issue in a PR somewhere?

portante:
@ndokos, any update on the status of this issue?

ndokos:
Yes, that is done by the current server RPM making machinery: see the [Makefile](https://github.com/distributed-system-analysis/pbench/blob/c69094876cb587c0c7a64af62cfdc0bef0e2721f/server/rpm/Makefile#L66) which saves it to a SHA1 file that gets installed on the server (that is also done by the <=0.69 RPM making machinery that is in the internal repo). Then the [ansible playbook](https://github.com/distributed-system-analysis/pbench/blob/rpm-work/server/ansible/roles/pbench-server-add-commit-id/tasks/main.yml) copies it it into the server config file.

portante:
This was done in PR #1793.

PR #1805, opened 2020-09-17
General method of fetching git commit ID

portante:


fetch-curr-commit-id
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env python3
+
+# See https://stackoverflow.com/questions/14989858/get-the-current-git-hash-in-a-python-script
+
+import git
+
+repo = git.Repo(search_parent_directories=True)
+print(f"{repo.head.object.hexsha}"[:9])
```
dbutenhof, line 8: 9 is a fairly typical default for `core.abbrev`, but it's really meant to be used like `short_sha = repo.git.rev_parse(sha, short=True)` where `--short` will return the shortest _unique_ abbreviation of __at least__ `core.abbrev` characters.

I'm not really sure of the scaling thresholds, but Linus recommends 12 characters, and I'm not sure it's a great idea to hardcode the truncation. (Even though 9 is almost certainly sufficient for uniqueness in our repo for all time, I'd feel better letting git decide what's unique rather than making an arbitrary cut.)

dbutenhof (commented):
So where will this get used? And I suspect we should do something to ensure that gitpython is installed in the context where we depend on this since it's likely not present by default. (It wasn't on my RHEL8 or on my Fedora 32 VM.)

fetch-curr-commit-id
portante: Good points, I'll fix.

portante:
> So where will this get used?

As a result of working on #1735, we needed a way to consider what the current git commit ID is for the code base being built.

However, that PR ended up just using the git commit ID that is in the RPM version string.

Since I had started to develop this method, I thought I'd post this as a PR so that we could avail ourselves of this code as necessary.

> And I suspect we should do something to ensure that gitpython is installed in the context where we depend on this since it's likely not present by default. (It wasn't on my RHEL8 or on my Fedora 32 VM.)

Yes, I need to add it to `requirements.txt`.

portante:
Might be part of a solution for #1364.

fetch-curr-commit-id
portante: Fixed.

portante (commented):
Ready for review!

requirements.txt
```diff
@@ -1,14 +1,15 @@
 ansible
-elasticsearch
-sh
 boto3
-requests
-werkzeug
+bottle
```
dbutenhof, line 3: Sorted and everything. I'm a bit sad that I hadn't done that. Well, cool.

dbutenhof (approved):

riya-17 (approved):

ndokos (commented):
Better late than never: LGTM

PR #1809, opened 2020-09-21
Minimal attempt to remove tool meister commands

portante:
Here we replace the `pbench-tool-data-sink` command line interface entirely replacing it with a `fork()` and invocation of the Tool Data Sink library code directly.  We then do the same for the `pbench-tool-meister` command.

In both cases the original commands are copied as-is to the library side of the tree with no modifications.  The new `pbench-tool-meister` command is a very simple trampoline command file required for our SSH tool meister default orchestration.

This demonstrates that neither the Tool Data Sink or the Tool Meister require a CLI interface for their operation as they only need a simple set of arguments passed to them.

### DIFF
--- a/agent/util-scripts/gold/test-client-tool-meister/test-53.txt
+++ b/agent/util-scripts/gold/test-client-tool-meister/test-53.txt
@@ -72,7 +72,7 @@
 [info][1900-01-01T00:00:00.000000] "iostat" tool is now registered for host "testhost.example.com" in group "default"
 --- pbench.log file contents
 +++ mock-run/tm/pbench-tool-data-sink.err file contents
-Bottle v#.##.## server starting up (using DataSinkWsgiServer(handler_class=<class '__main__.DataSinkWsgiServer.__init__.<locals>.DataSinkWsgiRequestHandler'>))...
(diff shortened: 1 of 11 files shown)


dbutenhof (dismissed):
I really don't think any of this has direct bearing on the discussion in Chuck's PR, but I like it. :+1:

portante:
> I really don't think any of this has direct bearing on the discussion in Chuck's PR, but I like it. 👍

Can you explain why it doesn't?

dbutenhof:
> > I really don't think any of this has direct bearing on the discussion in Chuck's PR, but I like it. +1
>
> Can you explain why it doesn't?

You know what... never mind. I find myself sinking into the role of "devil's advocate" here and while it was an interesting philosophical diversion originally, I really don't see much point to continuing it.

You're right that this approach doesn't require "parsing" of command line options, or even sequencing multiple arguments; although you're still doing that in the "ssh launch shell", which makes it technically possible for anyone to manually invoke it from the shell... but that's really not a big deal and I'll stop there. ;-)

dbutenhof (dismissed):

dbutenhof (dismissed):

dbutenhof (approved):

zulcss (approved):
Not totally happy but this can be reviewed latter.

PR #1810, opened 2020-09-21
Convert shell scripts to python

zulcss:
Converted the following scripts to python:

- pbench-list-triggers 
- pbench-register-tool-trigger 
- pbench-send-tools 
- pbench-start-tools 
- pbench-stop-tools 
- pbench-kill-tools 
- pbench-avg-stddev
- pbench-log-timestamp 
- pbench-cleanup
- pbench-agent-config-activate 
- pbench-agent-config-ssh-key 
- pbench-clear-tools
- pbench-list-tools 
- pbench-register-tool 
 - pbench-register-tool-set 
  - pbench-add-metalog-option 
  - pbench-postprocess-tools 
  - pbench-collect-sysinfo


portante (changes requested):
How does this work relate to PR #1770?  Is it a replacement?  Does it layer on top of that work?  Or does it already include that work?

Could we consider one commit for all the non-shell script conversion work, and then keep the individual commits for each shell script conversion?

zulcss:
> How does this work relate to PR #1770? Is it a replacement? Does it layer on top of that work? Or does it already include that work?
> 
> Could we consider one commit for all the non-shell script conversion work, and then keep the individual commits for each shell script conversion?

This commit builds on top of 1770, yeah i can merge these together.

dbutenhof (commented):
OK, enough for today. Some comments follow... More later.

dbutenhof (changes requested):

Issue #1751, opened 2020-08-12
When using tcpdump with 2 args like --interface=br1 --packets=1000 2nd arg is ignored


mnmehta:
https://github.com/distributed-system-analysis/pbench/blob/e9a3b3fa2d54ee11940e5e4b014e1fe48936a65b/agent/util-scripts/pbench-postprocess-tools#L166

@David Gray looks like you ran into a bug, if you specify something like --interface=br1 --packets=1000 the 2nd arg is being ignored, and gets the default value of 500. Your quick fix is to edit /opt/pbench-agent/util-scripts/pbench-postprocess-tool around line 166 replace "printf --" with "echo -e". @Peter Portante it looks like  printf -- --interface=br1 --packets=100 will only print the first arg, it does not print the 2nd. You can probably fix it by putting more escaped quotes on around tool_opts, or the echo -e also seems to work.

portante:
Fixed by #1813.

PR #1812, opened 2020-09-22
Fix tool argument handling (v0.69)

portante:
Fixes issue #1751, where tools with multiple arguments only have the first argument passed to the tool command.

This PR is only for the `b0.69` branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ dist
 *.egg-info
 .tox
 .eggs
+Pipfile
+Pipfile.lock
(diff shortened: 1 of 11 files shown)


ndokos (dismissed):
LGTM -  do we need to turn off black for this? The tests are failing because of it.

portante:
> Do we need to turn off black for this? The tests are failing because of it.

I back-ported PR #1768 to this branch, which should fix that problem.  The agent and server tests also failed, but they seem to be an ordering problem we have not locked down in the unit tests.

portante:
@ndokos, so I've had to add a couple more commits to address the test failures.  Seems like we have a working environment now.

zulcss (approved):
lgtm

ndokos (approved):

PR #1813, opened 2020-09-22
Fix tool argument handling

portante:
Fixes issue #1751, where tools with multiple arguments only have the first argument passed to the tool command.

When we load the registered tools into Redis, we process the tool options to make sure they are handled as a single string on the tool command line.

A separate PR for the `b0.69` branch exists as well, PR #1812.

### DIFF
--- a/agent/util-scripts/gold/test-client-tool-meister/test-53.txt
+++ b/agent/util-scripts/gold/test-client-tool-meister/test-53.txt
@@ -66,9 +66,9 @@
 --options=forty-two
 --- pbench tree state
 +++ pbench.log file contents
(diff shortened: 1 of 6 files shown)


Maxusmusti (commented):
The code looks good and seems to work, but there appears to be a discrepancy between the altered debug lines in util-scripts tests 53, 56, 57 and the travis environment.

ndokos (approved):
LGTM - but why are the tests failing? Maybe declare the `tool_opts` array as an indexed array with `declare  -a` would fix that?

zulcss (approved):
lgtm

Maxusmusti (approved):

ndokos (approved):

PR #1815, opened 2020-09-23
Only emit a warning message on missing debugfs (v0.69)

### LINKED ISSUE #1811: The perf-datalog is incorrectly requiring that debugfs be mounted before it will work
Starting with commit #53dba5f45 via PR #1538 (originally #1460), see also issue #1459), the `perf-datalog` is incorrectly requiring that `debugfs` be mounted before it will work.

Prior to that commit the code was (in the old `agent/tool-scripts/perf-tool`):

    if ! cat /proc/mounts | grep -q debugfs; then	
        mount -t debugfs none /sys/kernel/debug	
    fi

The new code now does (in the `agent/tool-scripts/datalog/perf-datalog`):

    if ! grep -q debugfs /proc/mounts; then
        printf -- "%s: required debugfs not mounted\n" "${PROG}" >&2
        exit 1
    fi

This should be changed to a warning to the user that `debugfs` is not mounted, and the `perf` data collected will not be as informational.

Note that PR #1538 is **_correctly_** not trying to mount `debugfs` itself.

portante:
Fixes #1811 for the `b0.69` branch.

See also #1814.

### DIFF
--- a/agent/tool-scripts/datalog/perf-datalog
+++ b/agent/tool-scripts/datalog/perf-datalog
@@ -17,8 +17,7 @@ if [[ ${?} -ne 0 ]]; then
 fi
 
 if ! grep -q debugfs /proc/mounts; then
-        printf -- "%s: required debugfs not mounted\n" "${PROG}" >&2
-        exit 1
+        printf -- "%s: debugfs not mounted, kernel perf data may be incomplete\n" "${PROG}" >&2
 fi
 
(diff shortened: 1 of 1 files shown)


portante:
FYI - This is built on top of PR #1812 since that has fixes to the test environment we need.

dbutenhof (approved):

ndokos (approved):

PR #1816, opened 2020-09-23
Perhaps useful script for updating gold files

portante:
When tests fail due to gold files being out of date, the `update-gold` script contributed here makes it easy to pull the newly generated gold file in as the committed gold file for the test.

Note well that one can take in unwanted change with this file, so beware.

### DIFF
--- a/contrib/development/update-gold
+++ b/contrib/development/update-gold
@@ -1,0 +1,56 @@
+#! /bin/bash
+
+cur_dir=${PWD}
+if [[ ! -x ${cur_dir}/unittests ]]; then
+    printf -- "ERROR: ${cur_dir} is not where a 'unittests' executable lives.\n" >&2
+    exit 1
+fi
+
+export _GF="/tmp/gold_files.$$"
+
+trap "rm -f ${_GF}" INT EXIT
+
+function _do_update {
+    > ${_GF}
+    ls -1 gold/test-${1}.txt gold/${1}.txt gold/*/test-${1}.txt gold/*/${1}.txt tests/*/test-${1}.txt tests/*/${1}.txt > ${_GF} 2> /dev/null
+    cnt="$(wc -l ${_GF} | awk '{print $1}')"
+    if [[ ${cnt} < 1 ]]; then
+        printf -- "ERROR: ${1} is not a recognized test.\n" >&2
+        exit 1
+    fi
+    if [[ ${cnt} > 1 ]]; then
+        printf -- "ERROR: ${1} is not a unique test name.\n" >&2
+        exit 1
+    fi
+
+    gold_file="$(cat ${_GF})"
+    _path=$(dirname "${gold_file}")
+    _name=$(basename "${gold_file}")
+    _name=${_name%.*}
+
+    dir_name="$(basename "${cur_dir}")"
+    if [[ "${dir_name}" == "bench-scripts" ]]; then
+        _tstdir=bench
+        _suffix="_output.txt"
+    elif [[ "${dir_name}" == "util-scripts" ]]; then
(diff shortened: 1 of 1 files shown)


contrib/development/update-gold
```diff
@@ -0,0 +1,56 @@
...
+    exit 1
+fi
+
+export _GF="/tmp/gold_files.$$"
+
+trap "rm -f ${_GF}" INT EXIT
+
+function _do_update {
+    > ${_GF}
+    ls -1 gold/test-${1}.txt gold/${1}.txt gold/*/test-${1}.txt gold/*/${1}.txt tests/*/test-${1}.txt tests/*/${1}.txt > ${_GF} 2> /dev/null
```
dbutenhof, line 15: Without being completely certain exactly how and where your ${1} pattern might appear in nested directory filenames... would this be more efficient as a `find gold -name \*${1}.txt` ???

contrib/development/update-gold
```diff
@@ -0,0 +1,56 @@
...
+    if [[ "${dir_name}" == "bench-scripts" ]]; then
+        _tstdir=bench
+        _suffix="_output.txt"
+    elif [[ "${dir_name}" == "util-scripts" ]]; then
+        _tstdir=utils
+        _suffix="_output.txt"
+    elif [[ "${dir_name}" == "bin" ]]; then
+        _tstdir=server
+        _suffix="/output.txt"
+    else
```
dbutenhof, line 41: Seems it might be better to normalize the naming conventions across the three, to avoid the suffix and various search naming patterns; but this is less change, so, fine. ;-)

contrib/development/update-gold
```diff
@@ -0,0 +1,56 @@
...
+        printf -- "ERROR: unrecognized test environment: ${cur_dir}\n" >&2
+        exit 1
+    fi
+    _test_out="/var/tmp/pbench-test-${_tstdir}/${_name}${_suffix}"
+
+    if [[ -e ${_test_out} ]]; then
+        mv ${_test_out} ${_path}/${_name}.txt && git add ${_path}/${_name}.txt && rm -f $(dirname ${_test_out})/output.diff $(dirname ${_test_out})/result.txt && rmdir $(dirname ${_test_out}) 2> /dev/null
```
dbutenhof, line 48: Do we really need to delete the previous run test results? That "hides the evidence" and should be cleaned up anyway by a subsequent run, right? Wouldn't it be better to leave it for further examination if necessary?

dbutenhof (approved):
Minor procedural comments and questions... but it looks fine, and seems useful.

ndokos (dismissed):

contrib/development/update-gold
portante: It might be.  I originally had this working for one kind of unit test structure, but then we kept making additional changes to and I just added the other names and locations.

contrib/development/update-gold
portante: Yes, for sure, but not sure we want to invest in this infrastructure but instead invest in the `pytest` based work coming.

contrib/development/update-gold
portante: Sure, but that helps me track what I have worked on and reviewed vs what is left to consider.

So if I make a change, and 5 tests fail, reviewing the diff of all five is a pain all at once.  Instead, I can review one of the files, `update-gold` on it, and then see what the others are left without rechecking them.

It is just the way I operate, so we can change this in the future.

dbutenhof (approved):

dbutenhof (dismissed):

dbutenhof (approved):

riya-17 (approved):

PR #1586, opened 2020-04-01
Pbench Documentation Site Design

anishaswain:
Fixes #1447 

The site is designed using PatternFly 4 with some slight CSS modifications. 
[Pbench website development guide
](https://docs.google.com/document/d/1xAlOZyzFU0xH_4v3QdLxCIiWnUp-79-2i9FEMJo4Rvc/edit?usp=sharing)

Wire Frame: 
[pbench-wires.pdf](https://github.com/distributed-system-analysis/pbench-dashboard/files/4354348/pbench-wires.pdf)

Demo: https://anisha1234.github.io/pbench/index.html

Screenshot

PR #1820, opened 2020-09-26
Fixes and foundational elements for PatternFly switch in our documentation

portante:
In order to prepare for our switch to use PatternFly project styling for the pbench documentation site, we need boat-load of files for the PatternFly infrastructure.

The first commit cleans up existing files that are not needed, and uses relative references for existing pages, while the second commit is a dump of all the new PatternFly infrastructure files added to the branch, but not being used yet.


portante:
By merging this first, we can substantially reduce the changes to be reviewed in #1586.

anishaswain (approved):
Checked with the lading page. Seems to render perfectly. So LGTM

PR #1821, opened 2020-09-28
Fix bench/util-scripts unit tests env overrides

portante:
Both the `bench-scripts` and the `util-scripts` unit test environments were overriding the various `pbench_*` environment variables defined in `base` with definitions provided ahead of time.  While this works, and the `base` script allows for it, it hid a few minor problems with the way some tests used sample data vs config data.

By adding an explicity definition of `pbench_run` in the generated `pbench-agent.cfg` files, we avoid the need to override many of the `pbench_*` environment variables.


portante:
This should allow tests leveraging the `PbenchAgentConfig` python object to work properly as well.

ndokos (approved):
LGTM.

Maxusmusti (approved):
Looks good to me as well. Sorry for delay, I have classes 12-8 MW

portante:
> Sorry for delay, I have classes 12-8 MW

No problem, don't think twice about it!

PR #1822, opened 2020-09-29
v0.69 branch RPM, Ansible, & Container backports 

portante:
Updates to the `b0.69` branch to enable RPM builds, updated Ansible playbooks, and container image builds.


agent/rpm/Makefile
```diff
@@ -0,0 +1,63 @@
...
+	if [ -d ${CWD}/patches ] ;then cp ${CWD}/patches/* ${RPMSRC} ;fi
+
+.PHONY: tarball
+tarball:
+	git submodule init
+	git submodule update
+	make -C .. DESTDIR=${TMPDIR}/${prog}-${VERSION}/agent
+	echo "${sha1}" > ${TMPDIR}/${prog}-${VERSION}/agent/SHA1
+	echo "${seqno}" > ${TMPDIR}/${prog}-${VERSION}/agent/SEQNO
+	tar zcf ${RPMSRC}/pbench-agent-${VERSION}.tar.gz -C ${TMPDIR} ${prog}-${VERSION}
```
ndokos, line 58: There are a couple of different possilbe mechanisms in building the tar ball, so I just want to put down some thoughts:

- We don't use the `exclude.pats` file here - instead, the explicit lists of files in the agent/Makefile ensure that the right files end up in the tarball. So we could get rid of the `exclude.pats` file completely (but see next bullet).

- Arguably, we should get rid of the extensive lists and add more things to `exclude.pats` - but there are problems: in my experiments, the current entries  in `exclude.pats`  don't work: e.g. `gold/` does not get rid of the `gold` subdirectories. `gold/*` gets rid of everything in the `gold/` subdirectory, but leaves the (empty) directory as part of the tarball. Something like `*/gold` works, except if the directory is at the top level. So it's not as easy and automatic as I would like it to be.

ndokos (commented):
Looks good - but the proof of the pudding is in the RPM making/image making/installing, which I have not tried yet. I left a couple of comments for possible changes but I don't think they are strictly necessary. I'll go test it before I approve it.

agent/rpm/Makefile
portante: For the `b0.69` branch let's just keep the explicit list in the `Makefile`, but for the `master` branch, I think restructuring the code so that it is easy to exclude non-installable items would be best; e.g., making a `tests` directory which has all the necessary infrastructure for tests separate from the rest of the code.

ndokos:
Submitted https://github.com/portante/pbench/pull/13 against your branch.

ndokos:
Minimal testing of the RPM I built (pbench-list-tools, pbench-clear-tools, pbench-register-tool-set, pbench-user-benchmark, pbench-move-results) showed no problems.

agent/ansible/bin/yamlconf
```diff
@@ -0,0 +1,31 @@
+#! /usr/bin/env python
```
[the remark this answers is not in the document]
portante: For now, I just dropped the need to install `yamlconf` at all.

agent/rpm/Makefile
portante: Removed the need for the `exclude.pats` entirely.

ndokos (approved):
Nice and simple.

Issue #1801, opened 2020-09-17
Enable the pbench-server running in a WSGI environment


portante:
The pbench-server starts up as a development server listening on http://127.0.0.1:8001/ - we'll need to get a WSGI environment for it.

See #1799.

PR #1823, opened 2020-09-30
Add gunicorn wsgi server configuration

npalaska:
- Runs the pbench-server systemd service with wsgi gunicorn server
- Add example configuration for Nginx 
- Addresses the issue https://github.com/distributed-system-analysis/pbench/issues/1801

### DIFF
--- a/lib/pbench/cli/server/shell.py
+++ b/lib/pbench/cli/server/shell.py
@@ -1,8 +1,23 @@
 #!/bin/env python3
 
 from pbench.server.api import create_app
+import subprocess
 
 
 def main():
     app = create_app()
-    app.run(debug=True, port=app.config["PORT"])
+    port = app.config["PORT"]
+    host = app.config["BIND_HOST"]
+    workers = app.config["WORKERS"]
+    subprocess.run(
+        [
+            "gunicorn",
+            "--workers",
+            str(workers),
+            "--pid",
+            "/run/pbench-server/gunicorn.pid",
+            "--bind",
+            f"{str(host)}:{str(port)}",
+            "pbench.server.api:create_app()",
+        ]
+    )
--- a/lib/pbench/server/api/__init__.py
+++ b/lib/pbench/server/api/__init__.py
@@ -93,6 +93,8 @@ def create_app():
     app.config["MAX_CONTENT_LENGTH"] = app.config_server.get("rest_max_content_length")
     app.config["REST_URI"] = app.config_server.get("rest_uri")
     app.config["LOG"] = app.config_server.get("rest_log")
+    app.config["BIND_HOST"] = app.config_server.get("bind_host")
+    app.config["WORKERS"] = app.config_server.get("workers")
 
     register_endpoints(api, app)
 
--- a/server/lib/config/nginx.conf.example
+++ b/server/lib/config/nginx.conf.example
@@ -1,0 +1,28 @@
+# Nginx setup to work with wsgi
+# Begin by opening up Nginx’s default configuration file:
+# Open up a server block just above the other server {} block that is already in the file
+
+http {
+    . . .
+
+    include /etc/nginx/conf.d/*.conf;
+
+    server {
+    }
+
+    server {
+        listen 80 default_server;
+
+        . . .
+
+# Add the following configuration inside the new server block
+
+server {
+    listen <port>;
+    server_name <Nginx_domain_or_IP>;
+
+    location / {
+        proxy_pass http://<pbench-server:8001>;
+        # proxy_pass http://127.0.0.1:8001; if running on same host as pbench-server
+    }
+}
--- a/server/lib/systemd/pbench-server.service.default
+++ b/server/lib/systemd/pbench-server.service.default
@@ -10,8 +10,11 @@ After=network.target httpd.service
 
 [Service]
 Type = simple
+PIDFile = /run/pbench-server/gunicorn.pid
+ExecStartPre = /bin/mkdir /run/pbench-server
 ExecStart = /opt/pbench-server/bin/pbench-server
-ExecStop = /usr/bin/pkill pbench-server
+ExecStop = /bin/kill -s TERM $MAINPID
+ExecStopPost = /bin/rm -rf /run/pbench-server
 Restart = always
 StartLimitInterval = 60
 StartLimitBurst = 10
--- a/server/lib/systemd/pbench-server.service.rhel7
+++ b/server/lib/systemd/pbench-server.service.rhel7
@@ -10,8 +10,11 @@ After=network.target httpd.service
 
 [Service]
 Type = simple
-ExecStart = /usr/bin/scl enable rh-python36 -- /opt/pbench-server/bin/pbench-server
-ExecStop = /usr/bin/pkill pbench-server
+PIDFile = /run/pbench-server/gunicorn.pid
+ExecStartPre = /bin/mkdir /run/pbench-server
+ExecStart = /usr/bin/scl enable rh-python36 /opt/pbench-server/bin/pbench-server
+ExecStop = /bin/kill -s TERM $MAINPID
+ExecStopPost = /bin/rm -rf /run/pbench-server
 Restart = always
 StartLimitInterval = 60
 StartLimitBurst = 10
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,6 +7,7 @@ colorlog
 elasticsearch1==1.10.0
 flask
 flask-restful
+gunicorn
 pyesbulk
 python-daemon
 python-pidfile
--- a/server/lib/config/pbench-server-default.cfg
+++ b/server/lib/config/pbench-server-default.cfg
@@ -68,6 +68,10 @@ rest_max_content_length = 100 * 1024 * 1024
 rest_uri = /api/v%(rest_version)s
 rest_log = %(pbench-logs-dir)s/pbench-server.log
 
+# WSGI gunicorn specific configs
+workers = 3
+bind_host = 0.0.0.0
+
 # WARNING - the pbench-server.cfg file should provide a definition of
 # pbench-backup-dir, e.g.:
 #     pbench-backup-dir = %(pbench-local-dir)s/archive.backup


npalaska:
We also need to open a port on our pbench-server for an Nginx host to have access (while also making sure other IP addresses can not directly access pbench-server).
Can we do something like this `firewall-cmd --zone=internal --add-source=<nginx_ip> --add-port=8001/tcp`?
cc: @ndokos @portante

server/lib/systemd/pbench-server.service.rhel7
```diff
@@ -9,8 +9,10 @@ Documentation = https://github.com/distributed-system-analysis/pbench
 After=network.target httpd.service
 
 [Service]
+user = $(/opt/pbench-server/bin/pbench-config -l user pbench-server)
```
portante, line 12: How does `pbench-config` get the `_PBENCH_SERVER_CONFIG` environment variable defined?

server/lib/systemd/pbench-server.service.rhel7
```diff
@@ -9,8 +9,10 @@ Documentation = https://github.com/distributed-system-analysis/pbench
...
+group = $(/opt/pbench-server/bin/pbench-config -l group pbench-server)
 Type = simple
-ExecStart = /usr/bin/scl enable rh-python36 -- /opt/pbench-server/bin/pbench-server
+ExecStart = /usr/bin/scl enable rh-python36 -- gunicorn --workers 3 --bind 0.0.0.0:8001 "pbench.server.api:create_app()"
```
portante, line 15: Perhaps we could consider hiding all the details of how to start a `pbench-server` instance inside the `pbench-server` script itself?  Then we could have that script handle the new for using SCL vs non-SCL, and fetch configuration options for the number of workers and the host/port to bind to.

portante (changes requested):
Can you look into hiding the invocation method for `pbench-server`?

server/lib/systemd/pbench-server.service.rhel7
npalaska: `pbench-config` script internally tries to read the environment variable  `_PBENCH_SERVER_CONFIG` It needs to be exported first I think

server/lib/systemd/pbench-server.service.rhel7
npalaska: Yes, in the recent commit I am starting the gunicorn server via pbench-server script, however I didnt notice that we now have 2 systemd service files one for rhel7 and one default. rhel7 systemd service file runs with scl and default runs without scl and the decision which one to run is made via one of the ansible roles. Does your comment implies that we get rid of 2 service file and have only 1 and decide whether to run with scl via pbench-server script?

lib/pbench/cli/server/shell.py
```diff
@@ -1,8 +1,13 @@
...
+import subprocess
 
 
 def main():
     app = create_app()
-    app.run(debug=True, port=app.config["PORT"])
+    port = app.config["PORT"]
+    host = app.config["GUNICORN_HOST"]
+    workers = app.config["GUNICORN_WORKERS"]
+    subprocess.run(["gunicorn", "--workers", str(workers), "--bind",
```
portante, line 12: Looks good.  Just wondering if the gunicorn writes its own pid to a file that we can that use to kill the specific gunicorn pid properly, see https://stackoverflow.com/questions/14604653/how-to-stop-gunicorn-properly.

portante (changes requested):
Looks good, just need to resolve some names and how to stop it properly.

lib/pbench/server/api/__init__.py
```diff
@@ -69,6 +69,8 @@ def create_app():
     app.config["MAX_CONTENT_LENGTH"] = app.config_server.get("rest_max_content_length")
     app.config["REST_URI"] = app.config_server.get("rest_uri")
     app.config["LOG"] = app.config_server.get("rest_log")
+    app.config["GUNICORN_HOST"] = app.config_server.get("gunicorn_host")
```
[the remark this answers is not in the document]
npalaska: I Agree

lib/pbench/cli/server/shell.py
npalaska: Thanks for the link, good idea we can write the gunicorn pid to a file and close it when necessary

server/lib/systemd/pbench-server.service.rhel7
npalaska: After little digging it seems like the environment variable we set in the systemd service file with either `ENVIRONMENT` or `EnvironmentFIle` parameter can only be used with `Exec*` parameter  (like in `ExecStart` or `ExecStop`). I think I can just use the user and group as `pbench` since the service runs as `pbench` user. Let me know if thats fine

portante: So we might want to merge the server ansible playbooks and roles first.  That way, the service file can be a jinja2 template which we render with the variables the playbooks and roles determine.

portante (changes requested):
Looks good, but doesn't the other `systemd/pbench-server.service` need to be modified in the same way?

npalaska:
Yes I will update the default pbench-server service file accordingly, I just wanted to get some feedback first

lib/pbench/cli/server/shell.py
```diff
@@ -1,8 +1,13 @@
...
+import subprocess
 
 
 def main():
     app = create_app()
-    app.run(debug=True, port=app.config["PORT"])
+    port = app.config["PORT"]
+    host = app.config["BIND_HOST"]
+    workers = app.config["WORKERS"]
+    subprocess.run(["gunicorn", "--workers", str(workers), "--pid", "/opt/pbench-server/gunicorn.pid",
```
portante, line 12: This apparently would be reformatted via `black`, as the lint job in Travis CI is failing.

server/lib/systemd/pbench-server.service.default
```diff
@@ -9,9 +9,13 @@ Documentation = https://github.com/distributed-system-analysis/pbench
 After=network.target httpd.service
 
 [Service]
+user = pbench
+group = pbench
 Type = simple
+PIDFile = /opt/pbench-server/gunicorn.pid
 ExecStart = /opt/pbench-server/bin/pbench-server
-ExecStop = /usr/bin/pkill pbench-server
+ExecStop = /bin/kill -s TERM $MAINPID
```
portante, line 17: What entity sets `MAINPID` to be the value stored in `/opt/pbench-server/gunicorn.pid`?

server/lib/systemd/pbench-server.service.default
```diff
@@ -9,9 +9,13 @@ Documentation = https://github.com/distributed-system-analysis/pbench
 After=network.target httpd.service
 
 [Service]
+user = pbench
+group = pbench
 Type = simple
+PIDFile = /opt/pbench-server/gunicorn.pid
```
portante, line 15: Can we move this `.pid` file to `/run/pbench-server.gunicorn.pid`?  Using `/run` is a bit more "standard", and prefixing it with `pbench-server` would avoid potential conflicts from other sub-systems.

portante (changes requested):
Looking good, just a few more things to consider.

server/lib/systemd/pbench-server.service.default
npalaska: `MAINPID` gets auto set via what we store inside `PIDFile` parameter of service file

server/lib/systemd/pbench-server.service.default
npalaska: sounds good

npalaska: Interesting, I cant run pbench-server service as a pbench user and try to read a file from `/run` directory (or we need to give `pbench` user an access to `run` directory) or alternatively run pbench-server service as a root user (without specifying user and group in service file just like what we have now)

portante (approved):
Looks good!

portante (changes requested):
Actually, needs a rebase first...

portante (approved):
Ship it!

lib/pbench/cli/server/shell.py
```diff
@@ -1,8 +1,23 @@
...
+    port = app.config["PORT"]
+    host = app.config["BIND_HOST"]
+    workers = app.config["WORKERS"]
+    subprocess.run(
+        [
+            "gunicorn",
+            "--workers",
+            str(workers),
+            "--pid",
+            "/run/pbench-server.gunicorn.pid",
```
ndokos, line 18: I think the convention is to have a subdirectory under `/run'  for the subsystem (although there are certainly exceptions, e.g. `sshd`), so `/run/pbench-server/gunicorn.pid`. Not sure if that requires additional configuration though.

server/lib/config/nginx.conf.example
```diff
@@ -0,0 +1,27 @@
+# Nginx setup need to done separately on staging and production.
```
ndokos, line 1: Make it into a template and add an ansible role to do the setup?

server/lib/systemd/pbench-server.service.default
```diff
@@ -10,8 +10,10 @@ After=network.target httpd.service
 
 [Service]
 Type = simple
+PIDFile = /run/pbench-server.gunicorn.pid
```
ndokos, line 13: If it changes in shell.py, it needs to change here as well.

server/lib/systemd/pbench-server.service.default
```diff
@@ -10,8 +10,10 @@ After=network.target httpd.service
...
 ExecStart = /opt/pbench-server/bin/pbench-server
-ExecStop = /usr/bin/pkill pbench-server
+ExecStop = /bin/kill -s TERM $MAINPID
+ExecStopPost = /bin/rm -f /run/pbench-server.gunicorn.pid
```
ndokos, line 16: ... and here. Maybe it should be a config item.

server/lib/systemd/pbench-server.service.rhel7
```diff
@@ -10,8 +10,10 @@ After=network.target httpd.service
 
 [Service]
 Type = simple
-ExecStart = /usr/bin/scl enable rh-python36 -- /opt/pbench-server/bin/pbench-server
-ExecStop = /usr/bin/pkill pbench-server
+PIDFile = /run/pbench-server.gunicorn.pid
```
ndokos, line 13: I probably should have made this a template as well and avoid having separate files for RHEL7 and everything else.

ndokos (changes requested):

lib/pbench/cli/server/shell.py
npalaska: yeah its probably doable via service file we can make an `ExecPre` parameter to check or create the directory `/run/pbench-server`. I will try to test it

server/lib/config/nginx.conf.example
npalaska: Oh now that I think about it I think I need to update this. @ndokos I dont think we need separate Nginx setup on staging or production (since we only have one Nginx instance running and we front staging and production pbench server by this one instance, correct me if I am wrong @portante ). So all we need to do is update that nginx config so it can route the calls to appropriate servers (if its not done yet). So we dont need any Nginx setup on server itself per se

dbutenhof: I think the real question here, for distribution, is whether we're going to require anyone standing up a Pbench server to supply and configure their own NGINX instance (in which case we should at least produce a sample configuration file) or whether we offer (perhaps optionally) to stand up a dedicated local NGINX instance for installations that wouldn't otherwise have one....

npalaska: @dbutenhof yes, and this example config will let someone do exactly that, that is configure their own NGINX instance with their pbench server

server/lib/systemd/pbench-server.service.rhel7
npalaska: I can probably do it using the ansible python api [ansible python api](https://docs.ansible.com/ansible/latest/dev_guide/developing_api.html#) (making a decision on whether to use scl or not by checking the ansible distribution) but I havent found a way to get ansible variables like `ansible_distribution_major_version` or `ansible_distribution` via python api

server/lib/systemd/pbench-server.service.default
```diff
@@ -10,10 +10,11 @@ After=network.target httpd.service
 
 [Service]
 Type = simple
-PIDFile = /run/pbench-server.gunicorn.pid
+PIDFile = /run/pbench-server/gunicorn.pid
+ExecStartPre = /bin/mkdir /run/pbench-server
 ExecStart = /opt/pbench-server/bin/pbench-server
 ExecStop = /bin/kill -s TERM $MAINPID
-ExecStopPost = /bin/rm -f /run/pbench-server.gunicorn.pid
+ExecStopPost = /bin/rm -rf /run/pbench-server/gunicorn.pid
```
ndokos, line 17: If we ever add anything else into the directory, this should probably wipe the whole directory.

server/lib/systemd/pbench-server.service.rhel7
```diff
@@ -10,10 +10,11 @@ After=network.target httpd.service
 
 [Service]
 Type = simple
-PIDFile = /run/pbench-server.gunicorn.pid
+PIDFile = /run/pbench-server/gunicorn.pid
+ExecStartPre = /bin/mkdir /run/pbench-server
 ExecStart = /usr/bin/scl enable rh-python36 /opt/pbench-server/bin/pbench-server
 ExecStop = /bin/kill -s TERM $MAINPID
-ExecStopPost = /bin/rm -f /run/pbench-server.gunicorn.pid
+ExecStopPost = /bin/rm -rf /run/pbench-server/gunicorn.pid
```
ndokos, line 17: Ditto: This should wipe the whole directory.

ndokos (changes requested):

server/lib/systemd/pbench-server.service.rhel7
ndokos: I'd rather it be done at installation time. No need to complicate this part of the code.

server/lib/systemd/pbench-server.service.default
ndokos: Is this still a problem? Do we have to create the /run/pbench-server directory beforehand with the appropriate owner/group?

ndokos (approved):

portante (approved):
Ship it!

server/lib/config/nginx.conf.example
ndokos: So that first comment line should go, right?

ndokos (changes requested):
Just one more tiny nit!

portante (approved):
Ship it!

ndokos (approved):

PR #1825, opened 2020-09-30
Base CLI setup

zulcss:
This branch sets up the base CLI agent bits so we can work on top of it:

- Setup BaseCommand class which is basically the base script that was done in shell script.
- Add setup_logigng method so that we can log errors and communicate with the user.
- Remove unused code
- Use CliContext class to create a click object. We create a click object so we can save the state across different CLI commands.

### DIFF
--- a/lib/pbench/agent/base.py
+++ b/lib/pbench/agent/base.py
@@ -1,0 +1,106 @@
+import abc
+import datetime
+import os
+import pathlib
+import socket
+import sys
+
+import click
+
+from pbench.agent import PbenchAgentConfig, fs
+
+
+class BaseCommand(metaclass=abc.ABCMeta):
+    """A base class used to define the command interface."""
+
+    def __init__(self, context):
+        self.context = context
+
+        self.config = PbenchAgentConfig(self.context.config)
+
+        self.pbench_run = self.get_path(os.environ.get("pbench_run", None))
+        if self.pbench_run is None:
+            self.pbench_run = self.config.pbench_run
+        if not self.pbench_run.exists():
+            click.secho(
+                f"[ERROR] the provided pbench run directory, {self.pbench_run}, does not exist"
+            )
+            sys.exit(1)
+
+        # the pbench temporary directory is always relative to the $pbench_run
+        # directory
+        self.pbench_tmp = self.pbench_run / "tmp"
+        try:
+            fs.safe_mkdir(self.pbench_tmp)
+        except Exception:
+            click.secho(f"[ERROR] unable to create TMP dir, {self.pbench_tmp}")
+            sys.exit(1)
+
+        # log file - N.B. not a directory
+        self.pbench_log = self.get_path(os.environ.get("pbench_log", None))
+        if self.pbench_log is None:
+            self.pbench_log = self.config.pbench_log
+
+        self.pbench_install_dir = self.get_path(
+            os.environ.get("pbench_install_dir", None)
+        )
+        if self.pbench_install_dir is None:
+            self.pbench_install_dir = self.config.pbench_install_dir
+        if not self.pbench_install_dir.exists():
+            click.secho(
+                f"[ERROR] pbench installation directory, {self.pbench_install_dir}, does not exist"
+            )
+            sys.exit(1)
+
+        self.pbench_bspp_dir = self.pbench_install_dir / "bench-scripts" / "postprocess"
+        self.pbench_lib_dir = self.pbench_install_dir / "lib"
+
+        self.ssh_opts = self.config.ssh_opts
+        os.environ["ssh_opts"] = self.ssh_opts
+
+        self.scp_opts = self.config.scp_opts
+        os.environ["scp_opts"] = self.scp_opts
+
+        os.environ["_pbench_debug_mode"] = "0"
+        if os.environ.get("_PBENCH_UNIT_TESTS"):
+            self.date = "1900-01-01T00:00:00"
+            self.date_suffix = "1900.01.01T00.00.00"
+            self.hostname = "testhost"
+            self.full_hostname = "testhost.example.com"
+        else:
+            self.date = datetime.datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%s")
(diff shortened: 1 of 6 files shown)


lib/pbench/agent/base.py
```diff
@@ -0,0 +1,104 @@
...
+    def execute(self):
+        """
+        This is the main method of the application
+        """
+        pass
+
+    def get_path(self, path):
+        """Converts a string path into a pathlib object"""
+        if path:
+            if not isinstance(path, pathlib.PurePath):
```
dbutenhof, line 101: Shouldn't there be an `else` returning `path`??? Or maybe just remove the `else:` on the `if path:` and return the input `path` if it either was already a Path object, or was None?

lib/pbench/agent/fs.py
```diff
@@ -0,0 +1,23 @@
...
+        shutil.rmtree(directory)
+
+
+def safe_mkdir(directory, clean=False):
+    """Safely create a directory.
+    Ensures a directory is present.  If it's not there, it is created.  If it is, it's a no-op. If
+    clean is True, ensures the directory is empty.
+    """
+    if clean:
+        safe_rmtree(directory)
```
dbutenhof, line 18: "ensures the directory is empty" ... ha. But I guess at the command level, deleting an empty directory and re-creating it shouldn't be a big deal.

lib/pbench/agent/utils.py
```diff
@@ -1,41 +1,60 @@
...
-            logger.error("%s is not writable", pbench_run)
-            sys.exit(1)
-        pbench_tmp = pathlib.Path(pbench_run, "tmp")
-        if not pbench_run.exists():
-            # the pbench temporary directory is always relative to pbench run
-            pbench_tmp.mkdir(parents=True, exists_ok=True)
+    if (
+        os.environ.get("_PBENCH_TOOL_MEISTER_CLIENT_LOG_LEVEL") == "debug"
+        or os.environ.get("_PBENCH_TOOL_MEISTER_START_LOG_LEVEL") == "debug"
+        or os.environ.get("_PBENCH_TOOL_MEISTER_START_LOG_LEVEL") == "debug"
```
dbutenhof, line 24: Same comments as before.

1. Although I suspect that fine tuning log levels is rarely all that useful, you're overriding that option by bundling them together.
2. You're checking `_PBENCH_TOOL_MEISTER_START_LOG_LEVEL` twice.

lib/pbench/cli/agent/__init__.py
```diff
@@ -0,0 +1,20 @@
+"""
+Create a click context object that holds the state of the agent
+invocation. The CliContext will keep track of passed parameters,
+what command created it, which resources need to be cleaned up,
+and etc.
+We create an empty object at the begining and populate the object
+with configuration, group names, at the beginning of the agent
+exectuion.
```
dbutenhof, line 8: "execution"

lib/pbench/cli/agent/__init__.py
```diff
@@ -0,0 +1,20 @@
+"""
+Create a click context object that holds the state of the agent
+invocation. The CliContext will keep track of passed parameters,
+what command created it, which resources need to be cleaned up,
+and etc.
+We create an empty object at the begining and populate the object
```
dbutenhof, line 6: An empty line between paragraphs would make it easier to read.

lib/pbench/agent/fs.py
zulcss: Done

lib/pbench/cli/agent/__init__.py
zulcss: Done

lib/pbench/cli/agent/__init__.py
zulcss: Done

lib/pbench/agent/utils.py
zulcss: Fixed

lib/pbench/agent/base.py
```diff
@@ -97,8 +97,7 @@ def execute(self):
 
     def get_path(self, path):
         """Converts a string path into a pathlib object"""
-        if path:
-            if not isinstance(path, pathlib.PurePath):
-                return pathlib.Path(path)
+        if not isinstance(path, pathlib.PurePath):
+            return pathlib.Path(path)
```
dbutenhof, line 101: But if `path` is `None`, `pathlib.PurePath` is going to be unhappy. Maybe that's OK if we know it can never be None; but you were checking explicitly before.

zulcss: Yeah fixed

dbutenhof (dismissed):

lib/pbench/agent/base.py
```diff
@@ -7,9 +16,89 @@ class BaseCommand(metaclass=abc.ABCMeta):
...
+        if not self.pbench_run.exists():
+            click.secho(
+                f"[ERROR] the provided pbench run directory, {self.pbench_run}, does not exist"
+            )
+            sys.exit(1)
+
+        # the pbench temporary directory is always relative to the $pbench_run
+        # directory
+        self.pbench_tmp = self.pbench_run / "tmp"
+        fs.safe_mkdir(self.pbench_tmp)
```
portante, line 33: Can we make `safe_mkdir` ensure the directory is either created or it raises an exception?

And it looks like it already does raise an exception which does not appear to be caught here.

portante (changes requested):
Can we address the exception handling in `safe_mkdir` and then rebase/squash to one commit?

lib/pbench/agent/base.py
zulcss: Done

portante (changes requested):
Looks good!  Let's get this rebased and a second review so we can merge.

portante (changes requested):
Can you squash to one commit?  Thanks!

portante (approved):
Ship it!

dbutenhof (approved):

PR #1827, opened 2020-09-30
Pbench agent config activate

zulcss:
Convert pbench-agent-config-activate to python


ndokos:
I think this (and most of the pbench-server-activate-* scripts, with the notable exception of pbench-server-activate-create-crontab) are obsolete: they were only used by the internal RPMs for installing key/config files. WIth the advent of ansible playbooks, these are not needed any longer.

portante (changes requested):
Needs a rebase and should wait for PR #1829 to land.

lib/pbench/agent/base.py
```diff
@@ -0,0 +1,105 @@
+import abc
```
ndokos, line 1: This file is in the tree now I think, so I guess this needs rebasing?

lib/pbench/agent/fs.py
```diff
@@ -0,0 +1,35 @@
...
+        safe_rmtree(directory)
+    try:
+        os.makedirs(directory)
+    except OSError as e:
+        if e.errno != errno.EEXIST:
+            raise
+
+
+def safe_copy(source, dest):
+    """Copy a file from source destination to a directory"""
```
ndokos, line 29: "from source to destination directory"

ndokos (changes requested):
A comment needs some attention, but I think it needs a rebase.

lib/pbench/test/functional/agent/cli/commands/test_config.py
```diff
@@ -0,0 +1,53 @@
...
+
+    conf = opt_pbench / "config"
+    conf.mkdir(parents=True)
+    command = ["pbench-agent-config-activate", str(config)]
+    err, out, exitcode = pytest.helpers.capture(command)
+    assert cfg.exists() is True
+    assert exitcode == 0
+
+
+def test_pench_config_ssh_key(
```
dbutenhof, line 42: Minor typo: `test_p[b]ench_config_ssh_key`

lib/pbench/test/functional/agent/cli/commands/test_config.py
```diff
@@ -0,0 +1,53 @@
...
+
+    config = tmp_path / "pbench-agent.cfg"
+    config.write_text(pbench_conf)
+    cfg = opt_pbench / "config" / "pbench-agent.cfg"
+
+    conf = opt_pbench / "config"
+    conf.mkdir(parents=True)
+    command = ["pbench-agent-config-activate", str(config)]
+    err, out, exitcode = pytest.helpers.capture(command)
+    assert cfg.exists() is True
```
dbutenhof, line 38: Hmm... to be truly paranoid, I might be inclined to verify `assert cfg.exists() is False` before invoking the command. (Easier than deleting it since we don't expect it.)

Shouldn't ever matter as long as we start with a clean test environment...

dbutenhof (approved):

portante (changes requested):
Let's wait on this PR.  We need to remove the notion of user and group completely from the pbench-agent.  See the work started in PR #1997.

ndokos (changes requested):
I like all the deletions, but why do we need python implementations of these things? I believe they are obsolete (they were used *only* by the defunct `pbench-agent-internal` RPM). 

Am I missing something?

portante:
> I like all the deletions, but why do we need python implementations of these things? I believe they are obsolete (they were used _only_ by the defunct `pbench-agent-internal` RPM).
> 
> Am I missing something?

Aren't these interfaces used by the ansible roles?

ndokos:
No - the [ansible role](https://github.com/distributed-system-analysis/pbench/blob/master/agent/ansible/pbench/agent/roles/pbench_agent_files_install/tasks/main.yml) installs the files directly.

portante:
> No - the [ansible role](https://github.com/distributed-system-analysis/pbench/blob/master/agent/ansible/pbench/agent/roles/pbench_agent_files_install/tasks/main.yml) installs the files directly.

Filed PR #2013 to remove entirely.

portante:
Closed by PR #2013.

PR #1828, opened 2020-09-30
Delete pbench-cleanup

ndokos:
It is not used by anything and is no longer useful (if it ever was).


portante (approved):
The only downside to making this change is that some people using the pbench-agent RPMs that have been released might use `pbench-cleanup` since it is in the documentation, and has been in our pbench agent's PATH for a while.

dbutenhof:
> The only downside to making this change is that some people using the pbench-agent RPMs that have been released might use `pbench-cleanup` since it is in the documentation, and has been in our pbench agent's PATH for a while.

The alternative would be deprecating it and deleting it later; would you prefer to do that vs just documenting that it didn't do anything interesting and is gone?

dbutenhof (dismissed):

portante (changes requested):
Actually, I think it is worth to deprecate it now, and then remove it in the following release.  The discipline alone is worth doing so that nobody is surprised by CLI changes.

dbutenhof:
> Actually, I think it is worth to deprecate it now, and then remove it in the following release. The discipline alone is worth doing so that nobody is surprised by CLI changes.

Which means we need a documentation process to remember to release note the deprecation; but that's not a bad thing.

portante:
> Which means we need a documentation process to remember to release note the deprecation; but that's not a bad thing.

Let's file an issue where we can add topics for the release notes, and then perhaps @ndokos can modify this PR to have `pbench-cleanup` emit a deprecation warning message when used?

ndokos:
Yes, this needs unit test fixing as well, so it might as well wait until 0.71. I'll do another PR to add a deprecation warning.

ndokos:
I just didn't want @zulcss to waste his time converting obsolete scripts.

dbutenhof:
I created issue #1836 to track 0.70 Agent release notes, starting with a deprecation note for this.

I linked the issue, but I removed it as we don't really want the issue to close. Is there a way to link an issue with a PR that won't close it?

portante:
> I created issue #1836 to track 0.70 Agent release notes, starting with a deprecation note for this.

Great!

> Is there a way to link an issue with a PR that won't close it?

I just edited the description to use a "check-box" style list, and then added the PR reference to the list entry.  Does that work?

dbutenhof:
> I just edited the description to use a "check-box" style list, and then added the PR reference to the list entry. Does that work?

I thought of using checkboxes, but I figured someone would transfer the list into release notes all at once and close it; but at worst it does no harm.

And, yeah, sure; I put a note here with the issue number, and recording PR numbers with each release note item in the issue is probably a good idea anyway.

riya-17 (approved):

ndokos:
The deprecation notice was added in PR #1864.

PR #1829, opened 2020-09-30
Remove agent and server activate scripts

ndokos:
The only exception is `pbench-server-activate-create-crontab` which is still used by the server ansible playbooks.


portante (changes requested):
Can we consider merging this to the `rpm-work` branch since that is where the Ansible roles and playbooks are for the server at this point?  Then we can merge that whole branch to `master` when it is ready.  What do folks think?

dbutenhof (dismissed):
Simplification is always good; why not just get this in and move on from there?

ndokos:
Yes, I jumped the gun: there are unit test failures that need to be dealt with and there is no reason to push this into 0.70.

riya-17 (dismissed):

npalaska (approved):
This is nice.

portante:
Can you rebase this work and make sure it works properly and then re-request reviews?  I placed it in `Draft` mode until then.

portante:
Replaced by work merge in PR #3100.

PR #1830, opened 2020-10-01
Lock in use of Perl 5.30 for Travis CI

portante:
We move to using Ubuntu "focal" which defaults to Perl 5.30.0.  As a result, we have to update a series of tool post-processing unit tests with changes that result from slightly different floating point output.


portante (commented):
A eye-ball check of the new floating point values seems reasonable, though I don't pretend to understand why such a change occurred.

.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: python
-dist: bionic
+dist: focal
```
dbutenhof, line 2: So I'm guessing it's the "dist: focal" that's "locking" Perl from the bionic 5.26 to focal's 5.30. A bit of background in a git log comment would help here. The issue isn't of much use since it merely points out FP value changes due to the Perl version upgrade without much context.

dbutenhof (approved):
The FP value changes seem to be pretty minor roundoff differences, but it's weird.

.travis.yml
portante: The commit message had the comment, the PR description did not.  Fixed.

ndokos (approved):

zulcss:
lgtm

PR #1831, opened 2020-10-01
Sort modules in *requirements.txt

portante:
Keeping them sorted makes it a bit easier to avoid adding a duplicate and avoids additions happening randomly.

There is a concern that pip might not handle dependency order properly, so we'll have to watch for that.


riya-17 (dismissed):

zulcss:
Why?

portante:
> Why?

Updated the commit message with a log comment explaining why, and also updated the PR description.

dbutenhof (dismissed):
Apparently, there's no dependency-order-safe solution for requirements.txt; because `pip` _builds_ all packages before _installing_ any. There has been concern expressed about packages that optionally use others, and wanting the dependency installed first so the consumer will choose to use it.

It simply can't be done with `requirements.txt` and the only solution is to separately `pip install` each package in the order you want. We could do that instead of using `-r requirements.txt` but unless we know of specific dependencies that concern us, (which I believe we don't), then there seems to be no point in worrying about it.

I like the alphabetical list until/unless we find a reason to drop `requirements.txt` and use some other mechanism.

dbutenhof (approved):

npalaska:
Another thing we could consider is breaking the requirements file for agent and server. Going forward there may be quite a few differences in the packages needed on server and agent hosts. The same requirement file may create a potential bug/conflict

npalaska (approved):

PR #1833, opened 2020-10-01
Deprecated

zulcss:
Remove a bunch of perl elasticsearch scripts since most people are using kibana and I dont think they are being used by anyone. Also no unittests.


dbutenhof (approved):
I'm up for getting rid of anything Perl, but I have no idea whether anyone might actually be depending on these or whether we should care so I'm not sure my vote is meaningful. ;-)

portante (changes requested):
First, the term deprecated is typically used to denote interfaces which are marked for removal in the future, or which won't be supported in some capacity.  Given this PR is outright removing these files, we should give at least one release, if not two, for the removal of these interfaces.

Second, these interfaces are used for the historical work on the CDM, and until we have had a chance to make a thorough review of their use and what we can to do with the data model, it is best if we keep them in place as is.

zulcss:
Closing based on last comment

PR #1835, opened 2020-10-01
CODE REVIEW CHANGES

Maxusmusti:
These are the changes specified for tool-meister branch

### DIFF
--- a/lib/pbench/agent/tool_meister.py
+++ b/lib/pbench/agent/tool_meister.py
@@ -59,7 +59,7 @@ import pidfile
 import redis
 
 from pbench.server.utils import md5sum
-from pbench.agent.modules import metaclass
+import pbench.agent.toolmetadata as toolmetadata
 
 
 # Path to external tar executable.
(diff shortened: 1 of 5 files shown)


lib/pbench/agent/tool_meister.py
```diff
@@ -122,21 +125,22 @@ def start(self):
                 + "/bindings/common"
             )
 
+            script_path = "/samples/scripts/dcgm_prometheus.py"
```
dbutenhof, line 128: No reason not to prepend `self.install_path` here, since it's added in all three uses of the subpath...

lib/pbench/agent/toolmetadata.py
```diff
@@ -65,6 +65,7 @@ def __getInitialData(self):
         return metadata
 
     def __dataCheck(self):
+        """Check for existing/loadable data, return 1 if true and 0 if false"""
```
dbutenhof, line 68: Ah, that makes more sense; but maybe it should return True/False instead of integers?

We really need to be better about using description strings, in general; especially with untyped parameters, having a description can be useful to figure out what to expect.

dbutenhof (approved):

dbutenhof (approved):
Nice.

lib/pbench/agent/tool_meister.py
ndokos: I took the point of the comment to be that you could say
```
   script_path = self.install_path + "/samples/scripts/dcgm_prometheus.py"
```
and just use `script_path` below. 

And maybe use `pathlib` for path component manipulations...

ndokos (changes requested):
LGTM  - one minor comment.

npalaska (approved):

portante (approved):
Ship it!

PR #1839, opened 2020-10-02
Post-Merge Review Changes

Maxusmusti:
Code review changes requested for persistent tooling additions

### DIFF
--- a/agent/tool-scripts/dcgm
+++ b/agent/tool-scripts/dcgm
@@ -1,6 +1,11 @@
 #!/usr/bin/python3
 # -*- mode: python -*-
 
+"""Place-holder "dcgm" persistent tool, providing a help message like the
+other existing transient tools, and issuing a warning indicating this script
+is not actually executed directly.
+"""
+
 import sys
 import os
 import logging
@@ -15,7 +20,9 @@ sh.setFormatter(shf)
 logger.addHandler(sh)
 
 if len(sys.argv) != 2 or sys.argv[1] != "--help":
-    logger.info("This script is deprecated, please run it with --help for info on registering the tool.")
+    logger.info(
+        "This script is deprecated, please run it with --help for info on registering the tool."
+    )
     logger.info("Run /opt/pbench-agent/tool-scripts/dcgm --help for more info.")
     exit(0)
 
--- a/agent/tool-scripts/node-exporter
+++ b/agent/tool-scripts/node-exporter
@@ -1,6 +1,11 @@
 #!/usr/bin/python3
 # -*- mode: python -*-
 
+"""Place-holder "node-exporter" persistent tool, providing a help message like the
+other existing transient tools, and issuing a warning indicating this script
+is not actually executed directly.
+"""
+
 import os
 import sys
 import logging
--- a/lib/pbench/agent/tool_data_sink.py
+++ b/lib/pbench/agent/tool_data_sink.py
@@ -16,7 +16,6 @@ import hashlib
 import json
 import logging
 import os
-import socket
 import subprocess
 import sys
 import tempfile
@@ -26,6 +25,7 @@ from http import HTTPStatus
 from pathlib import Path
 from threading import Thread, Lock, Condition
 from wsgiref.simple_server import WSGIRequestHandler, make_server
+from jinja2 import Environment, FileSystemLoader
 
 import daemon
 import pidfile
@@ -34,6 +34,7 @@ import redis
 from bottle import Bottle, ServerAdapter, request, abort
 
 import pbench.agent.toolmetadata as toolmetadata
+from pbench.agent import PbenchAgentConfig
 
 
 # Read in 64 KB chunks off the wire for HTTP PUT requests.
@@ -119,7 +120,41 @@ class DataSinkWsgiServer(ServerAdapter):
         self._server.shutdown()
 
 
+def _create_from_template(template, context, logger):
+    """Helper method used to generate the contents of a file given a Jinja
+    template and its required context.
+
+    Returns a string representing the contents of a file, or None if the
+    rendering from the template failed.
+    """
+    try:
+        inst_dir = PbenchAgentConfig(
+            os.environ["_PBENCH_AGENT_CONFIG"]
+        ).pbench_install_dir
+    except Exception as exc:
+        logger.error(
+            "Unexpected error encountered logging pbench agent configuration: '%s'",
+            exc,
+        )
+        return None
+
+    template_dir = Environment(
+        autoescape=False,
+        loader=FileSystemLoader(os.path.join(inst_dir, "templates")),
+        trim_blocks=False,
+        lstrip_blocks=False,
+    )
+    try:
+        filled = template_dir.get_template(template).render(context)
+        return filled
+    except Exception as exc:
+        logger.error("File creation failed: '%s'", exc)
+        return None
+
+
 class BaseCollector:
+    """Abstract class for persistent tool data collectors"""
+
     allowed_tools = {"noop-collector": None}
 
     def __init__(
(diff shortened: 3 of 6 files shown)


dbutenhof (approved):
Still not sure it really makes sense to add new deprecated scripts where `--help` documents one option (`--inst`), and using that option will circularly refer you to the `--help` which documents the "required" option you just found doesn't work. But ... at least the module doc string now clearly says it's a pointless diversion, so we're good. ;-)

lib/pbench/agent/tool_data_sink.py
```diff
@@ -134,21 +136,33 @@ def terminate(self):
...
+        if self.host_tools_dict and find_executable("podman"):
             self.abort_launch = False
         else:
+            if not find_executable("podman"):
+                self.logger.error(
+                    "Podman is not installed on this system (required by some registered tools, aborting launch"
+                )
             return 0
 
         config = open("prometheus.yml", "w")
```
portante, line 168: Consider re-writing this `config` creation sequence using a `with` statement.

Alternatively, you might want create the prometheus.yml file as a Jinja2 template, construct the list of host IP, tool, and port as a python list of tuples, and then invoke Jinja2 to render it to a new file.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -134,21 +136,33 @@ def terminate(self):
...
+    """Persistent tool data collector for tools compatible with Prometheus"""
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.volume = self.tool_group_dir / "prometheus"
 
     def launch(self):
 
-        if self.host_tools_dict:
+        if self.host_tools_dict and find_executable("podman"):
```
portante, line 159: Not sure you need this detection here, just the check for `podman`.  Down below, in the for loop, just count how many you added to the prometheus file, and then exit on line 197.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -190,9 +204,7 @@ def launch(self):
         prom_pull.wait()
 
         os.mkdir(self.volume)
```
portante, line 206: We need to `try`/`except` around this directory creation.

portante (changes requested):
Can you also add `try`/`except` clauses around the sub-process creations?

Now that we have a meta.json file, we don't need the place-holder files for `dcgm` and `node-exporter` anymore, do we?

lib/pbench/agent/tool_data_sink.py
```diff
@@ -154,6 +156,31 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.volume = self.tool_group_dir / "prometheus"
 
+    def __create_yml(self, template, context):
```
dbutenhof, line 159: A more specific name would be hand here; also this is maybe an `_` internal method but not really an `__` method.

This is almost a generic "apply Jinja template" method; there's certainly no inherent dependency here on it being a yml file. Maybe this should be elevated to a global util method; but at the very least a name implying something more like "expand Jinja template" would be more informative.

Plus, a description string particularly to explain the "context".

lib/pbench/agent/tool_data_sink.py
```diff
@@ -154,6 +156,31 @@ def __init__(self, *args, **kwargs):
...
+            )
+            return 1
+
+        template_dir = Environment(
+            autoescape=False,
+            loader=FileSystemLoader(os.path.join(inst_dir, "templates")),
+            trim_blocks=False,
+            lstrip_blocks=False,
+        )
+        context = {"tools": context}
```
dbutenhof, line 181: It's easy for someone to see `context` as a parameter to this call and as a parameter to the Jinja call and assume they're the same. It'd be more clear not to reassign the variable here.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -165,14 +192,10 @@ def launch(self):
...
 
-        config.write("global:\n  scrape_interval: 1s\n  evaluation_interval: 1s\n\n")
-        # config.write("alerting:\n  alertmanagers:\n  - static_configs:\n    - targets:\n\nrule_files:\n\n")
-        config.write(
-            "scrape_configs:\n  - job_name: 'prometheus'\n    static_configs:\n    - targets: ['localhost:9090']\n\n"
-        )
-
+        tool_context = []
         for host in self.host_tools_dict:
             if host.startswith("local"):
```
ndokos, line 200: Is this supposed to distinguish between "localhost" and everybody else? Or is there a possibility of some other host having a name that starts wtih "local"? If the former, you should compare it with "localhost", not just checking a prefix. If the latter, maybe a comment is indicated here - otherwise, I need to remember what host_tools_dict contains (which I probably should, but I don't :-) ).

Maxusmusti:
We still need the placeholder node-exporter and dcgm scripts in order to replicate the standardized "--help" option all tools have off of their scripts (At least until all of that is changed). For now people can use the expected method to discover tool options

ndokos (changes requested):

lib/pbench/agent/tool_data_sink.py
Maxusmusti: Yeah, this was in order to correct for the case where a user's formal hostname is something like "localhost.localdomain" on an unnamed vm. In order to still be usable, we would need ip as hostname. Not sure if we support this system-wide, though, but for now the support here doesn't hurt anything, will add a comment

lib/pbench/agent/tool_data_sink.py
```diff
@@ -240,7 +259,12 @@ def launch(self):
...
         ]
-        self.run = subprocess.Popen(args, stdout=prom_logs, stderr=prom_logs)
+        try:
+            self.run = subprocess.Popen(args, stdout=prom_logs, stderr=prom_logs)
+        except Exception as exc:
+            self.logger.error("Podman run process failed: '%s'", exc)
+            self.abort_launch = True
+            return 0
 
         prom_logs.close()
```
dbutenhof, line 269: Actually, here's another place where a `with` would be handy, because we're skipping the close on any exception.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -156,64 +189,40 @@ def __init__(self, *args, **kwargs):
...
-                tool_dict = {}
-                port = self.tool_metadata.getProperties(tool)["port"]
-                tool_dict["hostname"] = host_ip + "_" + tool
-                tool_dict["hostport"] = host_ip + ":" + port
-                tool_context.append(tool_dict)
-        yml = self.__create_yml("prometheus.yml", tool_context)
-        config.write(yml)
-        config.close()
+                self.abort_launch = True
+            config.close()
```
dbutenhof, line 225: Not necessary with `with`.

agent/tool-scripts/dcgm
```diff
@@ -1,6 +1,13 @@
 #!/usr/bin/python3
 # -*- mode: python -*-
 
+"""
+This script was originally created to fit dcgm into the transient tool model
```
portante, line 5: Can we consider re-writing this message to talk about the present and exclude the past?  Something like:

```
"""Place-holder "dcgm" persistent tool, providing a help message like the
other existing transient tools, and issuing a warning indicating this script
is not actually executed directly.
"""
```

agent/tool-scripts/node-exporter
```diff
@@ -1,6 +1,13 @@
 #!/usr/bin/python3
 # -*- mode: python -*-
 
+"""
+This script was originally created to fit node-exporter into the transient tool model
```
portante, line 5: Similar comment as noted in the `dcgm` tool script.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -112,7 +114,42 @@ def stop(self):
             self._server.shutdown()
 
 
+def _create_from_template(template, context, logger):
+    """Helper method used to create a file from Jinja template given context
```
portante, line 118: Consider a re-wordling like:

```
"""Helper method used to generate the contents of a file given a Jinja
template and its required context.

Returns a string representing the contents of a file, or None if the
rendering from the template failed.
"""
```
```

File: lib/pbench/agent/tool_data_sink.py [hunk 12]
```diff
@@ -134,49 +171,58 @@ def terminate(self):
...
-            for tool in self.host_tools_dict[host]:
-                port = self.tool_metadata.getProperties(tool)["port"]
-                config.write(
-                    "  - job_name: '{}_{}'\n    static_configs:\n    - targets: ['{}:{}']\n\n".format(
-                        host_ip, tool, host_ip, port
-                    )
-                )
-
-        config.close()
+                self.abort_launch = True
```
portante, line 224: Can you provide a more detailed description of why we need `abort_launch`?  It seems like it is because the `terminate` method needs to know if it should take action or not.  Why not just check for `self.run` is not `None`?

portante (changes requested):

lib/pbench/agent/tool_data_sink.py
```diff
@@ -125,90 +161,101 @@ def __init__(
...
-                        host_ip, tool, host_ip, port
-                    )
+        tool_context = []
+        with open("prometheus.yml", "w") as config:
+            for host in self.host_tools_dict:
+                # Used to get ip_addr of unnamed hosts like 'localhost.localdomain'
+                if host.startswith("local"):
+                    s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+                    s.connect(("8.8.8.8", 80))
+                    host_ip = str(s.getsockname()[0])
```
portante, line 204: Are these three lines (202-204) solely to get the current host IP address?

portante (approved):
Looks good to me!

lib/pbench/agent/tool_data_sink.py
Maxusmusti: Yeah, other more common methods had failed to work universally, so this was to  ensure uniform results

ndokos (approved):

lib/pbench/agent/tool_data_sink.py
portante: I realize this is in the original code, but I don't think we should keep this anymore.  This is not going to work reliably on hosts with a route that matches the hardcoded `8.8.8.8`.  You'd then be talking through some networking card only to get back to the local host, or worse it gets the wrong IP address where the node-exporter is not listening.

Networking should be properly setup so that prometheus should be able to open a TCP connection to IP resolved from the given name, to then use the HTTP protocol to fetch data from the OpenMetrics end-point.

If a user does not have a reliable name, then they should register that host by IP address and not by name.

portante (changes requested):
Back-tracking ... let's get that IP address resolution out.

portante (changes requested):

dbutenhof (approved):

dbutenhof (dismissed):

portante (changes requested):
Looks good, please rebase these commits and squash down to one commit, thanks!

Maxusmusti:
Rebased and squashed into a single commit

portante (approved):
Ship it!

npalaska (approved):

ndokos (approved):

PR #1842, opened 2020-10-05
Setup base classes

zulcss:
Stub in base classes for various CLI utils that we will
be replacing shell scripts with python scripts.

Since there are alot of different shell scripts to convert
I grouped the various commands into sub-classes from the BaseCommand
since alot of the shell scripts are re-using the same code as well.
So it made sense to me.

### DIFF
--- a/lib/pbench/cli/agent/core.py
+++ b/lib/pbench/cli/agent/core.py
@@ -1,0 +1,50 @@
+"""
+Base classes for various agent cli commands
+
+Base classes are used so that code can be shared between the same type
+of commands without duplicating the same code everywhere.
+"""
+
+from pbench.agent.base import BaseCommand
+
+
+class ToolCommand(BaseCommand):
+    """Common tool command Class"""
+
+    def __init__(self, context):
+        super().__init__(context)
+
+
+class LogCommand(BaseCommand):
+    """Common logging command class"""
+
+    def __init__(self, context):
+        super().__init__(context)
+
+
+class SysInfoCommand(BaseCommand):
+    """Common sysinfo command class"""
+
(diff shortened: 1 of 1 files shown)


dbutenhof (commented):
Is this just a placeholder to help with parallel development? The class names seem to have appeared before (aside from things like `SysInfoCommand` vs `SysinfoCommand`), but always pulling in mix-ins, and with extended behavior.

I think maybe I like having them centralized, although I might prefer having them separated within a single directory.

If these classes remain essentially just renames of the base command class with no extended behaviors, then this plethora seems like unnecessary confusion... but if the classes are going to gain more than trivial behavior, this is a lot of classes to be crammed into a single file.

In any case, this "colo" class style can complicate navigation by losing any obvious link from class to source file.

Maybe even if you could just extend the git comment to provide more context on your intent here?

lib/pbench/cli/agent/core.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class ToolCommand(BaseCommand):
+    """Common tool command Class"""
+
+    def __init__(self, context):
+        super().__init__(context)
+
+
+class LogCommand(BaseCommand):
```
portante, line 18: Is there a reason that we need all the command classes in the same file?  Why not have a `base.py` and a `tool.py`, `log.py`, etc?  If there is a good reason to have them all in the same file, great; however, it is not obvious why they need to be.

Having them in separate files means that we can add and convert commands as necessary without merge conflicts as we work.

portante (changes requested):

zulcss:
This isnt needed anymore

Issue #1840, opened 2020-10-02
md5sum of log.tar.xz does not match with the one store in log.tar.xz.md5 used for unit testing tarfile upload api


npalaska:
- md5sum of `lib/pbench/test/unit/server/fixtures/upload/log.tar.xz` does not match with one stored in `lib/pbench/test/unit/server/fixtures/upload/log.tar.xz.md5`. They should match for the unit testing
- Unit test of the upload api is not actually posting any data in the post request but just writing contents to a file based on the filename passed in the request header 
- Unit test fails for if we store the right md5sum (`md5sum log.tar.xz`) in .md5 file (I think its because we are not passing any data in the request)

portante:
Fixed as of the combination of #1869 and #1843.

PR #324, opened 2016-08-12
collect ansible_facts from pbench runs

chaitanyaenr:

PR #1843, opened 2020-10-05
Reformat the upload api test and fix stored md5sum of log.tar.xz used for unit testing

npalaska:
- Tries to address issue https://github.com/distributed-system-analysis/pbench/issues/1840
- Replicate the actual upload api behavior in the unit test by sending binary data in post request
- Fix md5sum of the compressed log file used for unit testing upload api


lib/pbench/test/unit/server/test_requests.py
```diff
@@ -90,13 +90,17 @@ def test_bad_extension_upload(client, bad_extension):
...
+            data = f.read()
 
-        with open(f"{filename}.md5") as md5sum_check:
+        with open(f"{datafile}.md5") as md5sum_check:
             md5sum = md5sum_check.read()
 
+        filename = "log.tar.xz"
         response = client.post(
             f"{client.config['REST_URI']}/upload",
+            data=data,
```
portante, line 103: Can `data` be the file object?  I seem to recall that you can have the `post` method read the file.

portante (changes requested):
So this change does not affect the unit tests?

dbutenhof (dismissed):

lib/pbench/test/unit/server/test_requests.py
npalaska: yes good point I think I can do `data = open(datafile, "rb")`

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -90,13 +91,16 @@ def test_bad_extension_upload(client, bad_extension):
...
-        filename = "./lib/pbench/test/unit/server/fixtures/upload/log.tar.xz"
+        filename = "log.tar.xz"
+        datafile = Path("./lib/pbench/test/unit/server/fixtures/upload/", filename)
+        data = open(datafile, "rb")
 
-        with open(f"{filename}.md5") as md5sum_check:
+        with open(f"{datafile}.md5") as md5sum_check:
             md5sum = md5sum_check.read()
 
         response = client.post(
```
portante, line 101: Consider using:

```
with open(datafile, "rb") as data_fp:
    response = client.post(
        f"{client.config['REST_URI']}/upload",
        data=data_fp,
        headers={"filename": filename, "md5sum": md5sum},
    )
```

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -90,13 +91,16 @@ def test_bad_extension_upload(client, bad_extension):
 
     @staticmethod
     def test_upload(client, pytestconfig):
-        filename = "./lib/pbench/test/unit/server/fixtures/upload/log.tar.xz"
+        filename = "log.tar.xz"
+        datafile = Path("./lib/pbench/test/unit/server/fixtures/upload/", filename)
```
ndokos, line 95: Do we have to have this hardwired?

portante (changes requested):
Looking really good, perhaps we could consider `with` statement to avoid the possibility of stale open file handles on error.

ndokos (changes requested):

lib/pbench/test/unit/server/test_requests.py
npalaska: This is just for the one unit test, we dont store it anywhere in config file at the moment

portante (approved):
Looks good.  Seems bad, though, that this change does not make the test fail.  Shouldn't the empty payload cause the API to return an error and the normal payload to not return an error?

dbutenhof (dismissed):

portante (changes requested):
And it needs a rebase ...

npalaska:
I too was surprised how the test was not failing, however, the unit test used to fails for any other `tar.xz` file I tried to use, then realized we were not actually sending any data in the post request unit test it was just reading the file contents from the filename sent in the header (where we are sending complete filepath as a name)

portante:
> I too was surprised how the test was not failing, however, the unit test used to fails for any other `tar.xz` file I tried to use, then realized we were not actually sending any data in the post request unit test it was just reading the file contents from the filename sent in the header (where we are sending complete filepath as a name)

Right, so this test sends the contents of a file to the server, and expects to get a success status response.  Why was it successful before this change when the contents were empty?  Do we have a bad error handling case?

npalaska:
I dont think so, I am guessing its because of the fact the we were storing an md5sum of an empty file in `log.tar.xz.md5` file, and this is the hash we were sending in the header of the post request, which upon receiving on the server side matches (since its an empty stream).

portante:
> I dont think so, I am guessing its because of the fact the we were storing an md5sum of an empty file in `log.tar.xz.md5` file, and this is the hash we were sending in the header of the post request, which upon receiving on the server side matches (since its an empty stream).

Okay, so the `content_length` header should also match the bytes received.  See what we do in the Tool Meister: https://github.com/distributed-system-analysis/pbench/blob/master/lib/pbench/agent/tool_data_sink.py#L671, https://github.com/distributed-system-analysis/pbench/blob/master/lib/pbench/agent/tool_data_sink.py#L720

But also, since the Tool Meister client and server are completely under our control, we don't need to double check everything.  But for the pbench server, the API code has to double check that (it might be already):

- We don't accept a `content-length` header of 0
- The `content-length` header cannot specify a size greater than some maximum
- If the total bytes received is greater than the `content-length` header, we return a 400 error as soon as we receive the first byte beyond the specified `content-length`
- Set a timeout for receiving all the specified data (i.e. when the bytes received is less than the specified `content-length`)
  - This puts a bound on slow-loris attacks
- Detect when a client closes the connection before we receive all the data
- The MD5 of the body should match the MD5 of the `md5sum` header
- Accept the payload if it is all received in good standing but we are unable to send the success response

npalaska:
Okay, the API request currently only checks the file extension, and whether all the required headers are present (`md5sum` and `filename`) it does not check the content length, I will add that condition. Regarding the maximum content-length argument, how do we determine that? (As a second thought we can add that in our Nginx configuration to not accept anything bigger than some size like 2 GB or something)

portante:
> Okay, the API request currently only checks the file extension, and whether all the required headers are present (`md5sum` and `filename`) it does not check the content length, I will add that condition.

+1

> Regarding the maximum content-length argument, how do we determine that?

See the Tool Data Sink code which accumulates the total bytes read and then checks it (see the line number references above).

> (As a second thought we can add that in our Nginx configuration to not accept anything bigger than some size like 2 GB or something)

That is an optimization, but we should make sure the base API also implements a maximum.

npalaska:
Sounds good, changes incoming

lib/pbench/server/api/__init__.py
```diff
@@ -286,7 +286,27 @@ def post(self):
...
+        except ValueError:
+            abort(400, message="Invalid content-length header, not an integer")
+        except Exception:
+            abort(400, message="Missing required content-length header")
+        else:
+            if content_length > _MAX_UPLOAD_DATA_SIZE:
+                abort(
+                    400,
+                    message="Content object too large, keep it at 1 GB"
+                    f" ({content_length:d}) and  under",
```
portante, line 300: Extra space; you might want to use the [`humanize`](https://pypi.org/project/humanize/) module to avoid embedding the constant of `1 GB` in the abort message string.

lib/pbench/server/api/__init__.py
```diff
@@ -286,7 +286,27 @@ def post(self):
...
+                )
+            elif content_length == 0:
+                abort(
+                    400,
+                    message="Content object is empty, please re-check the file contents",
```
portante, line 305: Consider the wording, "Payload was empty, nothing was uploaded".

portante (changes requested):
Coming along!

We also need to write the `.md5` file contents on success.

lib/pbench/server/api/__init__.py
dbutenhof: Also, worth keeping in mind here that we're checking the Content-Size header, not the actual payload: so while the caller claimed a length of 0, that doesn't actually mean the "object is empty", only that they said it was.

lib/pbench/server/api/__init__.py
```diff
@@ -305,9 +326,23 @@ def post(self):
             app.logger.exception("There was something wrong uploading {}", filename)
             abort(500, message=f"There was something wrong uploading {filename}")
 
-        if hash_md5.hexdigest() != md5sum:
-            abort(400, message=f"md5sum check failed for {filename}")
+        message = None
+        if bytes_received > content_length:
+            message = f"More bytes received than the specified content length {content_length}"
```
dbutenhof, line 331: We also want to complain if we get too _few_ bytes.

npalaska: Ahh right

lib/pbench/server/api/__init__.py
npalaska: Small concern: We will have to add this module in the requirement file, is it fine to add one more dependency just for this line? (unless we plan to use the module in other part of the code)

lib/pbench/server/api/__init__.py
```diff
@@ -272,37 +274,44 @@ class Upload(Resource):
...
             if content_length > _MAX_UPLOAD_DATA_SIZE:
+                app.logger.debug(
+                    "Tarfile upload: Content-Length exceeded maximum upload allowed size"
+                )
                 abort(
                     400,
-                    message="Content object too large, keep it at 1 GB"
-                    f" ({content_length:d}) and  under",
+                    message=f"Content object too large, keep it at {humanize.naturalsize(_MAX_UPLOAD_DATA_SIZE)}"
+                    f" ({content_length:d}) and under",
```
portante, line 309: Looks good, but perhaps a slight re-ordering to consider:

```
message=f"Payload body too large, {content_length:d} bytes, maximum size should be less than or equal to {humanize.naturalsize(_MAX_UPLOAD_DATA_SIZE)}"
```

lib/pbench/server/api/__init__.py
```diff
@@ -323,21 +332,30 @@ def post(self):
...
-            message = f"md5sum check failed for {filename}"
+            app.logger.debug("Tarfile upload: md5sum check failed")
+            message = f"md5sum check failed for {filename}, upload failed"
 
         if message:
             try:
                 os.remove(full_path)
             except Exception as exc:
-                self.logger.warning(
+                app.logger.debug(
```
portante, line 358: Is there a reason to change this to a `debug` message?  Seems like any failure to clean up should be flagged as a warning as it could indicate other problems.

lib/pbench/server/api/__init__.py
```diff
@@ -323,21 +332,30 @@ def post(self):
...
+                "Tarfile upload: There was something wrong uploading {}", filename
+            )
             abort(500, message=f"There was something wrong uploading {filename}")
 
         message = None
-        if bytes_received > content_length:
-            message = f"More bytes received than the specified content length {content_length}"
+        if bytes_received != content_length:
+            app.logger.debug(
+                "Tarfile upload: Bytes received does not match with content length header value"
```
portante, line 343: Perhaps worth emitting the upload file name?

lib/pbench/server/api/__init__.py
```diff
@@ -272,37 +274,44 @@ class Upload(Resource):
...
         global app
         if not request.headers.get("filename"):
-            app.logger.debug("Missed filename in header")
+            app.logger.debug("Tarfile upload: Missed filename in header")
             abort(400, message="Missing filename header in request")
         filename = secure_filename(request.headers.get("filename"))
 
         if not request.headers.get("md5sum"):
-            app.logger.debug("Missed md5sum in header")
+            app.logger.debug("Tarfile upload: Missed md5sum in header")
```
portante, line 282: Can we add the `filename` header value to all the logging statements?

portante (changes requested):
Looks good, just a few tweaks to the logging and messaging to consider.

lib/pbench/server/api/__init__.py
```diff
@@ -272,21 +274,48 @@ class Upload(Resource):
     def post(self):
         global app
         if not request.headers.get("filename"):
-            app.logger.debug("Missed filename in header")
+            app.logger.debug("Tarfile upload: Missed filename in header")
```
dbutenhof, line 277: Oh, you drew attention to these; I'd skimmed over them. "Missed filename" sounds awkward here. Something more like "Tarfile upload POST operation requires a `filename` header to name the uploaded file." ???

lib/pbench/server/api/__init__.py
```diff
@@ -272,21 +274,48 @@ class Upload(Resource):
...
+                )
+                abort(
+                    400,
+                    message=f"Content object too large, keep it at {humanize.naturalsize(_MAX_UPLOAD_DATA_SIZE)}"
+                    f" ({content_length:d}) and under",
+                )
+            elif content_length == 0:
+                app.logger.debug("Tarfile upload: Content-Length header value is 0")
+                abort(
+                    400, message="Payload Content-Length empty, nothing was uploaded",
```
dbutenhof, line 314: Again, this isn't really true; "nothing was uploaded" implies that we think they didn't upload data in the payload, but at this point we're only examining the header; this isn't going to help them figure out what they got wrong. The debug log message is actually more useful than the HTTP response.

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -91,7 +94,7 @@ def test_bad_extension_upload(client, bad_extension):
 
     @staticmethod
     def test_empty_upload(client, pytestconfig):
-        expected_message = "Payload Content-Length empty, nothing was uploaded"
+        expected_message = "Upload failed, Content-Length received in header is 0"
```
dbutenhof, line 97: It'd probably be nice to have test cases for Content-Length missing, and > MAX as well.

We should also have cases for the uploaded file MD5 not matching, and for a file length != Content-Length.

This could probably be done later, but it'd be better to go in with validation that it all works.

dbutenhof (commented):
Not 100% tested, but this is good.

lib/pbench/test/unit/server/test_requests.py
npalaska: indeed, I am looking into how we can add a custom `Content_Length` header value in flask test client, the request automatically overrides anything we send as value for `Content-Length`. Thats why even in `test_empty_upload` I had to create an empty `tar.xz` file in flask tmp directory and use that to upload. I coundn't simply pass `Content-Length: 0` in header of the request

dbutenhof (dismissed):

lib/pbench/server/api/__init__.py
```diff
@@ -21,6 +22,7 @@
 ALLOWED_EXTENSIONS = {"xz"}
 
 app = None
+_MAX_UPLOAD_DATA_SIZE = 2 ** 30
```
portante, line 25: Can we consider making this a configuration parameter instead of a hard-coded value?

lib/pbench/test/unit/server/test_requests.py
portante: Can you file an issue to track adding this kind of testing later?

portante (changes requested):
One more change, looks really good otherwise:  just think we should use a configuration parameter for the max upload size instead of a hard-coded value.

lib/pbench/server/api/__init__.py
npalaska: Just noticed we already have `rest_max_content_length = 100 * 1024 * 1024` in our server config file. I will just use that (its approx 1 GB)

lib/pbench/server/api/__init__.py
```diff
@@ -313,14 +312,15 @@ def post(self):
             )
             abort(400, message="Missing required content-length header")
         else:
-            if content_length > _MAX_UPLOAD_DATA_SIZE:
+            max_upload_size = int(eval(app.config["MAX_CONTENT_LENGTH"]))
```
portante, line 315: Not sure about `eval` here, why do we need it?

npalaska: Max content length value is stored as `100 * 1024 * 1024` in config file

lib/pbench/test/unit/server/test_requests.py
npalaska: Created an issue to track https://github.com/distributed-system-analysis/pbench/issues/1851

lib/pbench/server/api/__init__.py
portante: Not sure we should support arbitrary strings in the config file.  Let's consider modifying the config file to contain a "humanized" string, and see if that humanize python module can help us read it.

portante (changes requested):
Needs a rebase as well.

lib/pbench/server/api/__init__.py
npalaska: I think the `humanize` module is one directional meaning from bit/bytes to human readable format. https://python-humanize.readthedocs.io/en/latest/filesize/, I will write our own small function to convert from human readable to bytes and we can use humanized string in our config file

lib/pbench/server/utils.py
```diff
@@ -29,6 +29,30 @@ def rename_tb_link(tb, dest, logger):
...
 
 
+def filesize_bytes(size, logger):
+    size_name = ["B", "KB", "MB", "GB", "TB"]
+    try:
+        size = size.split()
+        if len(size) == 1:
+            size = size[0]
+            for i, c in enumerate(size):
+                print(i, c)
```
portante, line 39: Is this `print` statement necessary?

portante (changes requested):
Looks good, just need to resolve the need for that `print` statement.

lib/pbench/server/utils.py
npalaska: oops

dbutenhof (approved):

lib/pbench/server/utils.py
```diff
@@ -29,6 +29,29 @@ def rename_tb_link(tb, dest, logger):
         raise
 
 
+def filesize_bytes(size, logger):
```
portante, line 32: Not sure this works as it stands.  I wrote a little test to drive all possible combinations of input to see what would happen.  Here is the code fixed, with a few changes and a hacky test harness:

```
def filesize_bytes(size):
    size = size.strip()
    size_name = ["B", "KB", "MB", "GB", "TB"]
    try:
        parts = size.split(' ', 1)
        if len(parts) == 1:
            if size.isdigit():
                num = int(size)
                unit = 0
            else:
                for i, c in enumerate(size):
                    if not c.isdigit():
                        break
                print(f"size[:{i}] = {size[:i]}")
                num = int(size[:i])
                unit = size[i:].strip()
                print(f"num = {num}, unit = {unit}")
        else:
            num = int(parts[0])
            unit = parts[1].strip()

        if unit:
            idx = size_name.index(unit.upper())
        else:
            idx = 0
        factor = 1024 ** idx
    except Exception as exc:
        raise Exception("Invalid file size value encountered, '%s': %s", size, exc)
    else:
        return num * factor

sizes = [
    "  10  ",
]

for i in ["B", "b", "KB", "Kb", "kB", "kb", "MB", "GB", "TB"]:
    matrix = [
    f"10{i}",
    f"10 {i}",
    f"10   {i}",
    f"10{i}   ",
    f"10 {i}   ",
    ]
    sizes.extend(matrix)

for size in sizes:
    res = filesize_bytes(size)
    try:
        print(f"string = {size}, integer = {res:d}")
    except Exception as exc:
        print(f"string = {size}, exception = {exc}")
```

portante (changes requested):
Need some fixes to the `filesize_bytes` method.

lib/pbench/server/utils.py
dbutenhof: There ought to be a real pytest, for sure. But you were nastier in testing this than I was. Good for you. ;-)

npalaska: Interesting, thanks for the test @portante, yeah it wasnt very robust so added a comment in our config file `# max allowed size for tarfile upload, acceptable format {X[unit] or X [unit]}`

lib/pbench/server/api/__init__.py
```diff
@@ -272,21 +273,66 @@ class Upload(Resource):
...
+                f"Tarfile upload: Invalid content-length header, not an integer for file {filename}"
+            )
+            abort(400, message="Invalid content-length header, not an integer")
+        except Exception:
+            app.logger.debug(
+                f"Tarfile upload: No Content-Length header value found for file {filename}"
+            )
+            abort(400, message="Missing required content-length header")
+        else:
+            max_upload_size = filesize_bytes(app.config["MAX_CONTENT_LENGTH"])
```
portante, line 315: Should we consider moving the conversion of the `MAX_CONTENT_LENGTH` value to a proper integer once when we load the server configuration?  Seems like a bad config file value would go unnoticed until the server started running and uploads started flowing in.

lib/pbench/server/utils.py
```diff
@@ -29,6 +29,36 @@ def rename_tb_link(tb, dest, logger):
         raise
 
 
+def filesize_bytes(size):
```
portante, line 32: Consider reviewing and merging https://github.com/npalaska/pbench/pull/1.

portante (changes requested):
Looking good, two more tweaks to consider.

lib/pbench/server/api/__init__.py
npalaska: Indeed, I think that would be a good idea, I will move the conversion in `create_app`

dbutenhof (commented):
I like the robustness of Pete's enhancement; but also the ability to specify an integer without units. (Presumably, in real life, a _large_ integer.) So I'll second the suggestion to accept his change.

portante (changes requested):
Looks good, but still need that move to the `create_app` for the file size interpretation.

lib/pbench/server/api/__init__.py
```diff
@@ -96,6 +99,7 @@ def create_app():
     app.config["BIND_HOST"] = app.config_server.get("bind_host")
     app.config["WORKERS"] = app.config_server.get("workers")
 
+    _MAX_UPLOAD_SIZE = filesize_bytes(app.config["MAX_CONTENT_LENGTH"])
```
dbutenhof, line 102: Why not just set `app.config["MAX_CONTENT_LENGTH"]` so the server can get the converted value through the normal config dict mechanism rather than introducing a global? Just combine the conversion with the access in line 96 up above. There's no reason to be storing the original text.

npalaska: We use the "humanized" value of MAX_CONTENT_LENGTH on line #324 in the abort message. We will have to rehumanize the bytes value for the abort message thats why decided to use the constant variable. However, if you think its better to drop global variable I can do that

dbutenhof: You could split it into two keys, though, and still keep it in the dict.

```
app.config["MAX_CONTENT_LENGTH_INPUT"] = app.config_server.get("rest_content_max_length")
app.config["MAX_CONTENT_LENGTH"] = filesize_bytes(app.config["MAX_CONTENT_LENGTH_INPUT"])
```
However, there's also the question of whether we want to "optimally humanize" the value we use or just spit back the config file value they typed. E.g., is "1024 KB" reported as "1024 KB" or as "1 GB"??? In the latter case we'd want to re-humanize the binary value anyway.

npalaska: yeah I think we can add config key to hold bytes number. Regarding your second concern, I believe its manageable since the default config file is in our control

dbutenhof: Pete's fixes for spacing and lack of units could be taken as an argument against the "in our control" reasoning!

But again, I don't object to storing and emitting the original text; I just don't like the global. However, since we're only emitting the text version on a failure path, I don't see the cost of re-humanizing as an issue, either, and there might be some advantages to "normalizing" the notation.

npalaska: I agree, lesser the globals better it is

lib/pbench/server/api/__init__.py
```diff
@@ -296,18 +344,42 @@ def post(self):
...
+        if bytes_received != content_length:
+            app.logger.debug(
+                f"Tarfile upload: Bytes received does not match with content length header value for file {filename}"
+            )
+            message = (
+                f"Bytes received ({bytes_received}) does not match with content length header"
+                f" ({content_length}), upload failed"
+            )
 
+        if hash_md5.hexdigest() != md5sum:
```
portante, line 369: Shouldn't this be `elif`?

portante (changes requested):
Looks really good, just one last question...

lib/pbench/server/api/__init__.py
npalaska: Ahh you are right, thanks for pointing

dbutenhof (approved):
That looks to be seasoned just right. I think it's done!

portante (approved):
Ship it!

ndokos (approved):

PR #1845, opened 2020-10-06
Forward port of b0.69 branch agent RPM work and autogeneration of pbench-config

portante:


agent/Makefile
```diff
@@ -213,83 +224,71 @@ tool-postprocess = \
...
-# SHA1 and SEQNO - these are used when building an RPM only
-# so we provide a target for the spec file to invoke. This
-# is *NOT* meant to be invoked interactively.
-install-build-artifacts:
-	${COPY} SHA1 ${DESTDIR}
-	${COPY} SEQNO ${DESTDIR}
+install-ansible: install-destdir
+	${INSTALL} ${INSTALLOPTS} ${ANSIBLEDIR}
+	cd ansible; \
+	${COPY} ara ${ANSIBLEDIR}
```
dbutenhof, line 291: Only the `ara` directory? Huh.

dbutenhof (approved):

agent/Makefile
ndokos: Yes, we generally don't need any ansible stuff to be installed with two exceptions: the ara collector and pbench-run-benchmark (which is stored elsewhere).

ndokos (approved):

PR #1855, opened 2020-10-07
Ensure Tool Data Sink internal server object exists

### LINKED ISSUE #1853: util-scripts unittests, test-51, randomly fails sometime
As observed while debugging an `agent/util-scripts` unit test hang on `test-51`, there appears to be a race condition in the Tool Data Sink code where the main thread has called the `stop` method of the server object before its internal server object was created.

portante:
Close a race condition where the thread creating the internal Bottle server has not had a chance to create the object before the main thread attempts to invoke it's `stop` method.

Fixes #1853.

### DIFF
--- a/lib/pbench/agent/tool_data_sink.py
+++ b/lib/pbench/agent/tool_data_sink.py
@@ -98,18 +98,25 @@ class DataSinkWsgiServer(ServerAdapter):
 
         self.options["handler_class"] = DataSinkWsgiRequestHandler
         self._server = None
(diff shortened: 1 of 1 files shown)


ndokos (approved):

dbutenhof (approved):
Thought maybe other consumers of self._server could benefit from the CV, but apparently it's not referenced anywhere else, so I guess this is good.

Maxusmusti (approved):

PR #1848, opened 2020-10-06
Add missing quote

ndokos:
Fixes #1846

Also quote a couple of unquoted instances of ${sos_ver} on the same line for
uniformity's sake.

Issue #1846, opened 2020-10-06
pbench-sysinfo-dump syntax + logic bugs


Maxusmusti:
opt/pbench-agent/util-scripts/pbench-sysinfo-dump: line 100: conditional binary operator expected
/opt/pbench-agent/util-scripts/pbench-sysinfo-dump: line 94: expected )
/opt/pbench-agent/util-scripts/pbench-sysinfo-dump: line 94: expected )
/opt/pbench-agent/util-scripts/pbench-sysinfo-dump: line 94: expected )
/opt/pbench-agent/util-scripts/pbench-sysinfo-dump: line 100: syntax error near _modules="date'
/opt/pbench-agent/util-scripts/pbench-sysinfo-dump: line 100: 		_modules="date host release"'

Once these are fixed, still another logic issue afterwards causing the final step to fail

ndokos:
There is a quote missing on line 94.

portante:
Yes, this was something I was supposed to fix a while back and dropped the ball.  Thanks for picking this up, @ndokos, and for filing an issue for it, @Maxusmusti!

PR #1856, opened 2020-10-07
Fix generation of sosreport command

portante:
The fixes to pbench-sysinfo-dump in PR #1819 introduced a problem where the command line options were not properly spaced from each other, causing the sosreport command to fail.

See also issue #1846 and PR #1848.

### DIFF
--- a/agent/util-scripts/pbench-sysinfo-dump
+++ b/agent/util-scripts/pbench-sysinfo-dump
@@ -123,7 +123,7 @@ function collect_sos {
 	for mod in ${_modules}; do
 		printf -- " -o %s" "${mod}" >> ${_cmd}
 	done
-	printf -- "--batch%s --tmp-dir=\"%s\" --name \"%s\"" "${_quiet}" "${dir}" "${_name}" >> ${_cmd}
+	printf -- " --batch%s --tmp-dir=\"%s\" --name \"%s\"\n" "${_quiet}" "${dir}" "${_name}" >> ${_cmd}
 	chmod +x ${_cmd}
 
 	debug_log "[${script_name}]collecting sosreport"


agent/util-scripts/pbench-sysinfo-dump
```diff
@@ -123,7 +123,7 @@ function collect_sos {
 	for mod in ${_modules}; do
 		printf -- " -o %s" "${mod}" >> ${_cmd}
 	done
-	printf -- "--batch%s --tmp-dir=\"%s\" --name \"%s\"" "${_quiet}" "${dir}" "${_name}" >> ${_cmd}
+	printf -- " --batch%s --tmp-dir=\"%s\" --name \"%s\"\n" "${_quiet}" "${dir}" "${_name}" >> ${_cmd}
```
ndokos, line 126: I get the space but what's with the `\n`? That looks wrong to me.

ndokos (changes requested):

agent/util-scripts/pbench-sysinfo-dump
ndokos: Never mind...

ndokos (dismissed):

agent/util-scripts/pbench-sysinfo-dump
dbutenhof: Oh, yeah; I puzzled over that for a moment, too; but it's the end of the command, so terminating the line makes sense.

dbutenhof (dismissed):

ndokos (approved):

Maxusmusti (dismissed):

dbutenhof (approved):

ndokos (approved):

PR #1858, opened 2020-10-07
Test fixes required from previous commit

portante:
Not sure exactly why, but it appears the previous commit, the server RPM work, changes things so that these two tests now fail.


ndokos (approved):
Argh, yes - we are calling pbench-config which needs PYTHONPATH. Sorry I missed it.

dbutenhof (approved):

PR #1863, opened 2020-10-08
Switch legacy unit tests to use C.UTF-8 locale

portante:
Upcoming work to convert our infrastructure to Python3 will be leveraging python modules which require the `C.UTF-8` locale.

To ensure that such a switch does not affect the existing code, we make that change here in isolation to prove that out.


dbutenhof (approved):
Assuming the tests pass, this looks straightforward.

ndokos (approved):

dbutenhof (approved):

PR #1864, opened 2020-10-08
Deprecate pbench-cleanup & add tests for clear-results

portante:
Add a deprecation warning when `pbench-cleanup` is invoked, and change it to just invoke `pbench-clear-results`.  We then add two tests to verify their behavior.

Finally, we change `pbench-clear-results` to explicitly move the generated directory paths instead of relying on a wild-card (dangerous!).

### DIFF
--- a/agent/util-scripts/gold/pbench-cleanup/test-64.txt
+++ b/agent/util-scripts/gold/pbench-cleanup/test-64.txt
@@ -1,0 +1,15 @@
++++ Running test-64 pbench-cleanup 
+[warn][1900-01-01T00:00:00.000000] pbench-cleanup deprecated, will be removed in future release in favor of pbench-clear-results
+--- Finished test-64 pbench-cleanup (status=0)
++++ pbench tree state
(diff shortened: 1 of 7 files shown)


ndokos (dismissed):

riya-17 (approved):

zulcss (approved):
lgtm

dbutenhof (approved):
The extra precautions in `pbench-clear-results` are interesting; was there a specific reason for the added paranoia, or did it just seem like a good idea at the time? (Not complaining, just curious.)

portante:
> The extra precautions in `pbench-clear-results` are interesting; was there a specific reason for the added paranoia, or did it just seem like a good idea at the time? (Not complaining, just curious.)

This is very old code, and since most of the time `pbench-clear-results` is run by the `root` user, it is always dangerous to use `rm -rf *`.  So, yes, out of paranoia, and a healthy sense of caution, changing to remove wildcards and use an explicit path seemed prudent.

PR #1865, opened 2020-10-09
Eliminate scl and rh-python36 from the server spec file and the ansible roles

ndokos:
All the special code associated with SCL on RHEL7 is gone: henceforth
we are going to be using "native" python3 on every platform.

There are two commits: the first one deals with the spec file and the
second deals with the ansible roles.

In addition:

- the spec file has been turned into a jinja2 template, mimicking the
  pbench-agent spec file.

- the rpm/Makefile prepares a tar ball by doing a `python setup.py
  install' into a temp directory, and then installing all the other
  bits in the same temp directory, copying (mostly) what's necessary
  (some things - e.g. the nginx.conf.example file - end up getting
  copied as part of a directory copy, but then are deleted afterwards
  since they are not necessary - that's simpler than enumerating all
  of the files that need to be copied). What's in the temp directory is what
  ends up in the tarball.

- the spec file does a cp -a for the server and another one for the
  web-server, again mimicking the pbench-agent spec file.
    
- now that pbench-server.spec is a jinja2 template, we can get rid of
      utils/set-version-release. Also get rid of exclude.pats since we
      don't use it, move utils/rpm.mk up the rpm directory, get rid of the
      utils subdirectory and fix the include of rpm.mk in the Makefile.
    
 Fixes:
  - fixed misspelled "install-binlinkss"
  - fixed scl remnant in Makefile
  - added python3 requirement
  - removed python setup build requirement
  - create the rpmbuild/SPEC and rmbuild/SOURCES directories if
  necessary - the others get created by rpmbuild if necessary


ndokos:
I still need to look at the other dependencies (that policykit stuff), built it on all platforms and test it.

server/Makefile
```diff
@@ -63,16 +66,15 @@ setupfiles = \
 # targets
 .PHONY: install \
 	install-dirs \
-	install-bin install-binbase install-bintools install-binscripts \
+	install-bin install-binbase install-bintools install-binlinkss \
```
portante, line 69: One `s`?

portante (changes requested):
One extra `s`, otherwise looks good.

dbutenhof (commented):
I was confused by the volume of merge changes to skim through, and then momentarily by the fact that the agent Makefile and spec are still using `scl` (which I understand). Did you miss the server rpm Makefile, or do you intend to still support building the RHEL 7 server RPM on 7.6?

server/rpm/utils/set-version-release
```diff
@@ -0,0 +1,41 @@
+#! /bin/bash
```
portante, line 1: Can't we get rid of this now that we are using Jinja2?

portante (changes requested):
Seems like this needs a rebase, still reporting 5 commits.

server/rpm/Makefile
```diff
@@ -0,0 +1,74 @@
...
+
+VERSION = $(shell cat ${SERVER}/VERSION)
+
+DISTRO = $(shell uname -r | python -c 'import sys; print(sys.stdin.read().split(".")[-2])')
+
+ifeq (${DISTRO},el7)
+PYTHON = scl enable rh-python36 -- python3
+else
+PYTHON = python3
+endif
```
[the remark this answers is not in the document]
ndokos: No, just an oversight: fixed.

server/rpm/utils/set-version-release
ndokos: Right - will do.

server/Makefile
ndokos: Fixed

riya-17 (commented):
looks good

dbutenhof (approved):

server/rpm/Makefile
```diff
@@ -14,11 +14,7 @@ VERSION = $(shell cat ${SERVER}/VERSION)
 
 DISTRO = $(shell uname -r | python -c 'import sys; print(sys.stdin.read().split(".")[-2])')
 
-ifeq (${DISTRO},el7)
-PYTHON = scl enable rh-python36 -- python3
-else
 PYTHON = python3
-endif
 
 RPMSRC = ${HOME}/rpmbuild/SOURCES
```
dbutenhof, line 19: What would really be cool (although technically orthogonal to this PR) would be dependencies to create the `~/rpmbuild` directories we need....

dbutenhof (approved):
Cool.

server/rpm/Makefile
ndokos: Snuck that in. No extra dependencies: instead, when the spec file is created, we create it if necessary, as well as the `rpmbuild/SOURCES` subdir ; `rpmbuild` seems to take care of the rest on its own.

npalaska (approved):

server/rpm/Makefile
dbutenhof: Ah; I missed it the first time. Thanks!

portante (approved):
Ship it!

Issue #1860, opened 2020-10-08
Find the correct way to write/upload md5sum of tarfile uploaded using the upload api call


npalaska:
- Right now the code does not upload/write md5sum of the tarfile being uploaded (it only requires the raw md5sum in the header request) 
- We need a way to correctly write/upload md5sum otherwise uploaded tarfile goes into the quarantine directory 
- We also need to fix the upload directory path, we need to append the controller name to the `pbench-receive-dir` path before uploading so the tarfile can land in a proper directory

PR #1869, opened 2020-10-09
Replace POST /upload to PUT /upload/ctrl/<controller> and write md5sum on disk on server

npalaska:
Addresses the issue https://github.com/distributed-system-analysis/pbench/issues/1860
Specifically it makes the following changes
- Replaces the md5sum header name to standard `Content-MD5`
- First writes the md5sum on the disk before writing the tarfile content
- Writes the tarfile content with tempfile before actually writing on disk
- Replaces upload POST request with PUT request and remove hostname from header req
- The new upload URI will be /upload/ctrl/<string:controller_name>
- The controller name will be specified by the user when building the URI

### DIFF
--- a/lib/pbench/server/api/__init__.py
+++ b/lib/pbench/server/api/__init__.py
@@ -10,6 +10,7 @@ import hashlib
 from pathlib import Path
 import requests
 import humanize
+import tempfile
 
 from flask import request, jsonify, Flask, make_response
 from flask_restful import Resource, abort, Api
@@ -39,7 +40,9 @@ def register_endpoints(api, app):
     """Register flask endpoints with the corresponding resource classes
     to make the APIs active."""
 
-    api.add_resource(Upload, f"{app.config['REST_URI']}/upload")
+    api.add_resource(
+        Upload, f"{app.config['REST_URI']}/upload/ctrl/<string:controller>"
+    )
     api.add_resource(HostInfo, f"{app.config['REST_URI']}/host_info")
     api.add_resource(Elasticsearch, f"{app.config['REST_URI']}/elasticsearch")
     api.add_resource(GraphQL, f"{app.config['REST_URI']}/graphql")
@@ -273,7 +276,7 @@ class HostInfo(Resource):
 
 
 class Upload(Resource):
-    def post(self):
+    def put(self, controller):
         global app
         if not request.headers.get("filename"):
             app.logger.debug(
@@ -285,7 +288,7 @@ class Upload(Resource):
             )
         filename = secure_filename(request.headers.get("filename"))
 
-        if not request.headers.get("md5sum"):
+        if not request.headers.get("Content-MD5"):
             app.logger.debug(
                 f"Tarfile upload: Post operation failed due to missing md5sum header for file {filename}"
             )
@@ -293,7 +296,7 @@ class Upload(Resource):
                 400,
                 message="Missing md5sum header, POST operation requires md5sum of an uploaded file in header",
             )
-        md5sum = request.headers.get("md5sum")
+        md5sum = request.headers.get("Content-MD5")
 
         app.logger.debug("Receiving file: {}", filename)
         if not allowed_file(filename):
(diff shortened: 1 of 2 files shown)


npalaska:
Added `do not merge` because I am seeing an error, where upon uploading the tarfile it goes into `WONT-UNPACK` directory for some reason with the following error
```
find: ‘/srv/pbench/public_html/incoming/npalaska-fedora32-test/pbench-user-benchmark_test2_2020.10.08T21.51.59.unpack/pbench-user-benchm
ark_test2_2020.10.08T21.51.59’: No such file or directory
chmod: missing operand after ‘ugo+rx’
``` 
Trying to figure out why this might be happening

lib/pbench/server/api/__init__.py
```diff
@@ -295,6 +295,16 @@ def post(self):
             )
         md5sum = request.headers.get("md5sum")
 
+        if not request.headers.get("hostname"):
```
dbutenhof, line 298: I'd suggest a more specific extension header for this, like maybe `X-Pbench-Controller-Host` so there's no confusion.

For that matter, we should probably be using the standard `Content-MD5` instead of our own `md5sum`.

dbutenhof (commented):
Maybe there's a mismatch between the actual hierarchy in the tar file vs what the unpacker is expecting? I.e., maybe your headers don't match the reality?

npalaska:
Yo are right @dbutenhof that was the part of mistake from my side while testing. Ran couple more tests and it worked fine now

lib/pbench/server/api/__init__.py
npalaska: hmm, I didnt know there is standard header for md5, thats good

portante: Also, Flask should allow us to declare endpoints which have variable names in the URL.  For example, today at [line 42](https://github.com/distributed-system-analysis/pbench/blob/73c20444d14ccd03e163dad4fcee25b0edd12081/lib/pbench/server/api/__init__.py#L42) we just declare the `Upload` end-point URL to be `/upload`.  But as we [read in the docs](https://flask-restful.readthedocs.io/en/latest/api.html#flask_restful.Api.add_resource), we can instead declare the resource as:

```
api.add_resource(Upload, f"{app.config['REST_URI']}/upload/ctrl/<controller>")
api.add_resource(Upload, f"{app.config['REST_URI']}/upload/user/<user>")
```

And then the `controller` and `user` variables will be passed when the `post()` method is invoked.  That avoids using another header argument, allowing us to easily track behaviors in the logs.  Further, today we need to use `controller`, but tomorrow we'll be doing this by `user` (the controller information is in the tar ball and will be derived from it instead of required for placement).

Also, we should consider using `PUT` instead of `POST`, as I think we want the idempotency of the operation (see [REST - PUT vs POST](https://restfulapi.net/rest-put-vs-post/)).

portante (changes requested):
Partial review ...

lib/pbench/server/api/__init__.py
npalaska: Should I address the `md5sum` header change comment in this PR or a separate one (I can do it here its a small change but another thought says its different change than this issue)

lib/pbench/server/api/__init__.py
```diff
@@ -380,6 +392,11 @@ def post(self):
                     exc,
                 )
             abort(400, message=message)
+
+        # Write the checksum as .md5 file
+        with open(f"{full_path}.md5", "w") as f:
+            f.write(f"{md5sum}  {filename}\n")
```
portante, line 398: We have have settled on in the exist SSH based protocol, and it seems it would apply here as well, is that the following steps are taken to ensure we never see a tar ball without its `.md5` file:

1. Write `.tar.xz` file to a hidden temporary file (see [Tool Data Sink, line 744](https://github.com/distributed-system-analysis/pbench/blob/master/lib/pbench/agent/tool_data_sink.py#L744)) in the same directory it will eventually live
2. After all bits have been received and written to disk, verify the expected MD5 value matches header
3. Write MD5 value to `.tar.xz.md5` first
  - On failure to write, return a 500 error, temporary hidden file will evaporate
4. Now that we have successfully written the `.tar.xz.md5` file, we can `link()` the hidden temporary file to its final name
  - See Tool Data Sink line [787](https://github.com/distributed-system-analysis/pbench/blob/master/lib/pbench/agent/tool_data_sink.py#L787)

portante (changes requested):

lib/pbench/server/api/__init__.py
npalaska: @portante I am bit confused about PUT vs POST, 
the doc says this for PUT

> Use PUT when you want to modify a singular resource which is already a part of resources collection. PUT replaces the resource in its entirety

and for POST

>  Use POST when you want to add a child resource under resources collection.

What we are doing is actually adding a new tarfile under the hostname directory not replacing/ modifying the existing resource/ tarfile. Isnt what we are doing is closer to the POST operation?

npalaska: However, if our resource is controller directory then yes we are modifying the resources then we are closer to the PUT request. I am just confused whether our resource is controller or a tarfile

npalaska: But I think what you said also make sense, we actually PUT a tarfile inside a controller directory

npalaska: @portante interesting read https://stackoverflow.com/questions/56240547/should-http-put-create-a-resource-if-it-does-not-exist. Seems like RFC 2616 is obsolete now

npalaska: okay that make sense

lib/pbench/server/api/__init__.py
npalaska: Ahh yeah writing data to tmpfile first is better

lib/pbench/server/api/__init__.py
dbutenhof: I really like the idea of putting the controller (and maybe host) path in the URL.

There's a lot of ambiguity in `PUT` vs `POST` for object creation. Idempotency is definitely an issue, because `PUT` usually specifies the specific resource (which means you already need to know the URI) whereas POST generally creates a resource and returns the URI (e.g., via the `Location` response header... which we may want to consider anyway).

I'm not sure there's really much ground here for "purity" arguments anyway. And there's no reason we can't allow retrying a POST as long as there's a discernible URI from the arguments and payload of the request ... which in this case, I believe we have.

portante: Regarding creating a resource, the pbench server semantics today are that the client determines the URI of the resource (albeit, indirectly).  The `controller`/`<tar ball>` name forms the URI.  If two different clients send results today to the pbench server where it happens that they both have controllers with the same name and both provide the same tar ball name, then one will be accepted and one will be quarantined.

I think we should keep the same semantic we have today given we are not changing that functionality with this commit.  To that end, the updated RFCs recommends `PUT` for situations where the resource URI is known ahead of time.

Further, both [OpenStack Swift] (https://docs.openstack.org/api-ref/object-store/) and [Amazon S3](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPOST.html) object stores allow for PUT operations.  They also both support `POST` operations, OpenStack Swift with slight semantic differences, and S3 considering it just another way to send the same data.

We could decide to enhance this PR to support both mechanisms to do the same thing.

However, I'd rather only support `PUT` today, so that we can support for `POST` later if we decide we want to take away control of the URI construction from the client when the proper back-end infrastructure plumbed out.

dbutenhof (dismissed):
That is looking pretty good.

Although the Path/file operations should ensure we're catching this in some form, I'm thinking that maybe some explicit unit test cases to validate clean abort behavior when given illegal filename/controller strings might not be a bad idea.

lib/pbench/server/api/__init__.py
```diff
@@ -333,53 +336,82 @@ def post(self):
...
-                app.logger.warning(
-                    "Failed to remove compressed %s when trying to clean up: %s",
-                    full_path,
-                    exc,
+                with md5_full_path.open("w") as md5fp:
+                    md5fp.write(f"{md5sum} {filename}\n")
+            except Exception:
+                try:
+                    os.remove(md5_full_path)
+                except Exception as exc:
```
portante, line 389: We should ignore `FileNotFoundError` exceptions since the creation of the `.md5` file might not have happened.

portante (changes requested):
One small exception handling bug to fix, but looks good otherwise.

We'll also need a PR to change the existing client-side URL construction.

npalaska:
@portante Where do we use the post request currently in client side?

portante:
> @portante Where do we use the post request currently in client side?

I think here: https://github.com/distributed-system-analysis/pbench/blob/master/lib/pbench/cli/agent/commands/results/__init__.py

portante (approved):
Ship it!

dbutenhof (approved):

PR #1873, opened 2020-10-13
Keep only one systemd service file

npalaska:
Address bugs found during the server installation on RDU2 intlab-008 (Oct 13/2020 2:00 – 3:00pm EST)


dbutenhof (commented):
Don't we need a change to keep the installer/configuration from trying to use the `pbench-server.service.rhel7`??

And if the `requests` problem was just using `pip` instead of `pip3` on RHEL7, why didn't we have problems with the other dependencies? Odd.

ndokos (changes requested):
I'm nto sure that we want all of these in the same PR. And the `pbench-server.service.default` file should pprobably be renamed to `pbench-server.service`. Plus the ansible role will need changes for that. Can you take care of the service files in this PR? The RPM/Ansible changes should probably go in through the rpm-branch.

npalaska:
@ndokos this PR is is against RPM-work branch itself

npalaska:
@ndokos only the service file related changes are now addressed in this PR

portante (approved):

ndokos:
`pip` was a red herring: the spec file does `python3 -m pip -r requirements.txt` which is equivalent to `pip3 install -r requirements.txt`, so no change is needed in that.

npalaska:
yes I reverted those changes

ndokos (approved):

PR #1874, opened 2020-10-13
Fixes for server installation failures: RPM and ansible

ndokos:
- The server Makefile installs setup.py, setup.cfg and requirements.txt.
- The spec file includes the above.
- The ansible pbench-server-install-config-file role was missing a
  a `become=no` in the `delete local temp dir' task.
- The ansible pbench-server-systemd-service-role does not distinguish
  between centos/rhel-7 and everything else: everybody does the same
  thing.

### DIFF
--- a/server/Makefile
+++ b/server/Makefile
@@ -39,6 +39,7 @@ install-dirs:
 
 install-python3-setup: install-bin install-lib
 	mkdir -p ${DESTDIR}/python3
+	${COPY} ../requirements.txt ${DESTDIR}
 	(cd ..; SKIP_GENERATE_AUTHORS=1 SKIP_WRITE_GIT_CHANGELOG=1 python3 setup.py install --prefix=${DESTDIR}/python3)
 	${COPY} ${DESTDIR}/python3/bin/pbench-config ${DESTDIR}/python3/bin/pbench-server ${BINDIR}/
 	${RM} -r ${DESTDIR}/python3
--- a/server/rpm/pbench-server.spec.j2
+++ b/server/rpm/pbench-server.spec.j2
@@ -94,6 +94,7 @@ fi
 /%{installdir}/%{static}/VERSION
 /%{installdir}/%{static}/package.json
 /%{installdir}/package.json
+/%{installdir}/requirements.txt
 
 /%{installdir}/lib/config/pbench-server-satellite.cfg.example
 /%{installdir}/lib/config/pbench-server.cfg.example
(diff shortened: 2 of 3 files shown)


server/ansible/roles/pbench-server-systemd-service/vars/main.yml
```diff
@@ -1,6 +1,2 @@
 ---
-pbench_server_service_file_suffix: "{{ 'rhel7' if ansible_distribution_major_version == '7' and 
-                                                  (ansible_distribution == 'RedHat' or ansible_distribution == 'CentOS') 
-                                       else 'default' }}"
-
-pbench_server_service_file:  /opt/pbench-server/lib/systemd/pbench-server.service.{{ pbench_server_service_file_suffix }}
+pbench_server_service_file:  /opt/pbench-server/lib/systemd/pbench-server.service
```
npalaska, line 2: This change is in conflict with the the PR https://github.com/distributed-system-analysis/pbench/pull/1873,  since this a service file related change I addressed it in the PR otherwise these 2 PR will become defendant on each other

server/Makefile
```diff
@@ -39,6 +39,7 @@ install-dirs:
 
 install-python3-setup: install-bin install-lib
 	mkdir -p ${DESTDIR}/python3
+	${COPY} ../setup.py ../setup.cfg ../requirements.txt ${DESTDIR}
```
portante, line 42: Why do we need `setup.py` and `setup.cfg` in the source tree?  All we need to do is `pip3 install -r requirements.txt`, rigth?

server/ansible/roles/pbench-server-systemd-service/vars/main.yml
portante: We'll land #1873 first, so when #1874 is rebased this change will disappear from this commit.

server/rpm/pbench-server.spec.j2
```diff
@@ -94,6 +94,9 @@ fi
 /%{installdir}/%{static}/VERSION
 /%{installdir}/%{static}/package.json
 /%{installdir}/package.json
+/%{installdir}/setup.py
```
portante, line 97: Not sure why we need `setup.py` and `setup.cfg`.

portante (changes requested):

server/rpm/pbench-server.spec.j2
ndokos: I'm just confused - fixed.

server/ansible/roles/pbench-server-systemd-service/vars/main.yml
ndokos: Reverted.

portante (approved):

PR #1875, opened 2020-10-13
Update agent cli to reflect server upload http changes

npalaska:
Agent cli was using post request to `/upload` endpoint, however it should now be updated PUT request to `/upload/ctrl/<controller>` endpoint

### DIFF
--- a/lib/pbench/agent/results.py
+++ b/lib/pbench/agent/results.py
@@ -153,7 +153,7 @@ class MakeResultTb:
 class CopyResultTb:
     chunk_size = 4096
 
-    def __init__(self, tarball, config, logger):
+    def __init__(self, controller, tarball, config, logger):
         if not os.path.exists(tarball):
             logger.error("tarball does not exist, '{}'", tarball)
             sys.exit(1)
@@ -165,7 +165,7 @@ class CopyResultTb:
         self.tarball_md5 = Path(tarball_md5)
         self.logger = logger
         server_rest_url = config.results.get("server_rest_url")
-        self.upload_url = f"{server_rest_url}/upload"
+        self.upload_url = f"{server_rest_url}/upload/ctrl/{controller}"
 
     def read_in_chunks(self, file_object):
         data = file_object.read(self.chunk_size)
@@ -187,10 +187,10 @@ class CopyResultTb:
         with open(self.tarball_md5, "r") as md5fp:
             md5sum = md5fp.read()
         filename = secure_filename(str(self.tarball))
-        headers = {"filename": filename, "md5sum": md5sum}
+        headers = {"filename": filename, "Content-MD5": md5sum}
         with self.tarball.open("rb") as f:
             try:
-                response = requests.post(
+                response = requests.put(
                     self.upload_url, data=self.read_in_chunks(f), headers=headers
                 )
                 self.logger.info("File uploaded successfully")
--- a/lib/pbench/cli/agent/commands/results/__init__.py
+++ b/lib/pbench/cli/agent/commands/results/__init__.py
@@ -1,6 +1,7 @@
 import os
 import sys
 import requests
+import tempfile
 
 from pbench.common.logger import get_pbench_logger
 from pbench.agent import PbenchAgentConfig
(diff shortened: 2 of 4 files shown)


lib/pbench/agent/results.py
```diff
@@ -165,7 +165,7 @@ def __init__(self, tarball, config, logger):
         self.tarball_md5 = Path(tarball_md5)
         self.logger = logger
         server_rest_url = config.results.get("server_rest_url")
-        self.upload_url = f"{server_rest_url}/upload"
+        self.upload_url = f"{server_rest_url}/upload/ctrl/controller"
```
portante, line 168: This appears to be a "fixed", or hard-coded, controller named, `controller`.

The current `bash` implementation starts in `pbench-move-results`, where at [line 85](https://github.com/distributed-system-analysis/pbench/blob/master/agent/util-scripts/pbench-move-results#L85), it sets the `controller` to be the `full_hostname` environment variable, as provided by the [`base`](https://github.com/distributed-system-analysis/pbench/blob/master/agent/base#L128) script.  Then `pbench-move-results` constructs a temporary directory at [line 172](https://github.com/distributed-system-analysis/pbench/blob/master/agent/util-scripts/pbench-move-results#L172), which is where it places the constructed tar ball from `pbench-make-result-tb` (see [line 198](https://github.com/distributed-system-analysis/pbench/blob/master/agent/util-scripts/pbench-move-results#L198)).

The old code did it that way so that it could use one (1) `scp` command (see [line 68](https://github.com/distributed-system-analysis/pbench/blob/master/agent/util-scripts/pbench-copy-result-tb#L68) of `pbench-copy-result-tb`), followed by one (1) `ssh` command to verify the bits on the far side.  This sequence allowed us to copy the `.tar.xz` and the `.tar.xz.md5` file in one `scp` command, also providing the `controller` name (the name of the directory in which the tar ball is contained) without making multiple calls to `scp` or `ssh` beyond the two required (copy the data, verify the MD5 sum holds).

In this new HTTP/S PUT API, we place the `controller` in the URI, the expected MD5 checksum value in the `Content-MD5` header, and `.tar.xz` contents in the payload body.

We almost have all the pieces in place, except that a `controller` needs to be placed into the URI the client forms.

It looks like we would want to use the `controller` found at [line 14](https://github.com/distributed-system-analysis/pbench/blob/master/lib/pbench/cli/agent/commands/results/__init__.py#L14) of the new Python3-based `pbench move/copy results` command(s), and pass that into the `CopyResultTb()` constructor so that it can form the URI properly at [line 168](https://github.com/distributed-system-analysis/pbench/blob/master/lib/pbench/agent/results.py#L168).

portante (changes requested):

lib/pbench/agent/results.py
npalaska: Thanks @portante for the detailed explanation. It make sense to update the constructor of the `CopyResultTb` to take `controller` name since its client defined now.

lib/pbench/cli/agent/commands/results/__init__.py
```diff
@@ -49,15 +50,9 @@ def move_results(ctx, _user, _prefix, _show_server):
...
-    # FIXME: use tempfile
-    temp_dir = os.path.join(
-        config.pbench_tmp,
-        ".".join([os.path.basename(__file__), str(os.getpid())]),
-        controller,
-    )
     try:
-        os.makedirs(temp_dir)
-    except OSError:
+        temp_dir = tempfile.mkdtemp()
```
portante, line 54: This appears to lose the use of `config.pbench_tmp`.  Perhaps we could just do `tempfile.mkdtemp(dir=config.pbench_tmp, prefix="pbench-move-results.")`?

portante (changes requested):
Looking good, just a couple of final touches.

dbutenhof (approved):
Looks good, although I like Pete's suggestions...

dbutenhof (dismissed):
<pedantic_mode>
I won't leave this as a "formal comment" on the current code; but as a gentle reminder for all of us, it'd be nice to try to get in the habit of writing description strings for classes and methods to help future readers easily absorb what's going on.
</pedantic_mode>

portante (approved):
Ship it!

portante (changes requested):
Well, actually, we need a rebase first.

portante (approved):

riya-17 (approved):

dbutenhof (approved):

PR #1880, opened 2020-10-14
fix pbench-server pip install

portante:
In addition to fixing the pip3 install invocation, we also:

 - Ignore local seqno files in the server/rpm tree
 - Add Python3 required modules to the post installation step of the pbench-agent
 - Rearrange pbench-server spec file a bit to reduce steps where possible


npalaska (approved):
Looks good

dbutenhof (approved):
Ship it!

server/rpm/pbench-server.spec.j2
```diff
@@ -9,16 +9,16 @@ Source0:        pbench-server-%{version}.tar.gz
 Buildarch:      noarch
 
 
-Requires: python3
+Requires: python3 python3-devel gcc
```
ndokos, line 12: This was required for compiling psutil, so in the end it may not be needed on the server, depending on how things shake out with the requirements files, but better safe than sorry.

ndokos (approved):
LGTM. We may optimize things a bit after the requirements.txt shake-out - mañana.

PR #1884, opened 2020-10-14
Pbench list triggers

zulcss:
Pythonize pbench-list-triggers


lib/pbench/cli/agent/commands/triggers/list.py
```diff
@@ -0,0 +1,80 @@
...
+            if len(self.groups) == 0:
+                self.logger.error("%s: error fetching list of tool groups", self.name)
+                return 1
+
+            for group in self.groups:
+                trigger = self.gen_tools_group_dir(group) / "__trigger__"
+                if trigger.exists():
+                    with open(trigger) as f:
+                        data = f.read().splitlines()
+                    print("%s: %s" % (group, "  ".join(data)))
```
dbutenhof, line 42: In the `bash` version, the second argument is just a quoted `$(cat __trigger__)` so the records will remain newline separated. You're separating them with whitespace instead, apparently. Why isn't this breaking a test? Or are all the test cases single lines?

dbutenhof (commented):
One "bash transcription" variance...

lib/pbench/cli/agent/commands/triggers/list.py
dbutenhof: OK, so if a tool trigger is a single line with `<start>:<end>` then maybe translating the `$(cat ...)` with `.splitlines()` and `.join()` isn't the best transliteration of the bash idiom?

Perhaps simply printing the first/only line and logging if there's more than one would be cleaner/more explicit?

dbutenhof (dismissed):

lib/pbench/cli/agent/commands/triggers/list.py
```diff
@@ -0,0 +1,83 @@
...
+                    '%s: bad tool group specified, "%s"\n',
+                    self.name,
+                    self.context.group,
+                )
+                return 1
+            trigger = self.tool_group_dir / "__trigger__"
+            if trigger.exists():
+                with open(trigger) as f:
+                    data = f.read()
+                    print(data.strip())
```
npalaska, line 59: shouldn't this be `log` instead of `print`?

npalaska (commented):
Can we remove the PR from draft if ready?

lib/pbench/cli/agent/commands/triggers/list.py
dbutenhof: This is the output of the list command; so, no. ;-)

portante (changes requested):
Needs a rebase ...

dbutenhof (approved):

portante (approved):

agent/Makefile
```diff
@@ -40,7 +40,6 @@ util-scripts = \
 	pbench-init-tools \
 	pbench-kill-tools \
 	pbench-list-tools \
-	pbench-list-triggers \
```
portante, line 43: While we want to remove it from here, we need to add it below just like `pbench-config`.

portante (changes requested):
Close ...

dbutenhof:
> * 1 for Unused import

Had a feeling this was going to happen; you left the `import logging` again.

dbutenhof (approved):

dbutenhof (dismissed):

portante (changes requested):
Looks good.

The tests are failing because the new agent side logging is always creating a log file even when no logs are emitted. I think we can just update the unit test gold files and move on.

dbutenhof (approved):

dbutenhof (approved):

dbutenhof (approved):

lib/pbench/cli/agent/commands/triggers/list.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def execute(self):
+        if not self.pbench_run.exists():
+            # Silently exit if we don't have a pbench run directory.
+            return 1
+
+        if not self.context.group:
+            # list tool triggers in all groups
+            if len(self.groups) == 0:
+                self.logger.error("%s: error fetching list of tool groups", self.name)
```
portante, line 31: Can we add a set of unit tests for this which will test the `--help` and executing it once?  I am concerned that the logging needs to use `{}` style formatting instead of `%s`.

portante (changes requested):

portante (approved):
Nice!

dbutenhof (approved):

PR #1885, opened 2020-10-14
Fix breakage

ndokos:
- We need to create the %{installdir} before copying  ./server/* to it.
- We move the package.json file from %{installdir}/html/static to
  %{installdir}/, so we need to get rid of the former in the %files
  list.


dbutenhof (approved):

npalaska (approved):

portante (approved):
Ship it!

PR #1894, opened 2020-10-20
Stop invoking screen directly in unit tests

portante:
We fix the mock of `screen` for the unit tests so that it is no longer being invoked.  We also modify the Tool Meister code to properly find the `screen` command on the `PATH` instead of invoking it as a hard-coded path.  By removing the invocation of `screen` directly we reduce the requirements for the test environment.

Finally, we fix the `util-scripts` `test-53`, `test-56`, and `test-57` so that the order of execution does not break the test randomly.


agent/util-scripts/gold/test-client-tool-meister/test-53.txt
```diff
@@ -132,4 +120,8 @@ INFO pbench-tool-meister main -- Remove pid file ... (tm-default-testhost.exampl
 /var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/pidof -x iostat
 /var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/pidof -x mpstat
 /var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/pidof -x mpstat
```
ndokos, line 122: I thought the `test-bin/pidof` was deleted above: so how come we have it running here still?

ndokos: nm - it just got moved.

ndokos (approved):

dbutenhof (approved):

PR #1900, opened 2020-10-21
Fix agent rpm builds

portante:


agent/rpm/Makefile
```diff
@@ -53,8 +53,7 @@ patches:
 .PHONY: tarball
 tarball:
 	mkdir -p ${RPMSRC}
-	git submodule init
-	git submodule update
+	(cd ${AGENT}/../; git submodule init; git submodule update)
```
npalaska, line 56: just a question, does this change the working directory and do we need to cd back to the rpm directory?

zulcss (commented):
lgtm

agent/rpm/Makefile
portante: The closing parentheses should prevent the changing of the default working directory from escaping just that command.

portante:
> lgtm

Can you officially approve it to allow a merge?

zulcss (approved):

agent/rpm/Makefile
ndokos, line 56: Why does it matter? You can do `git submodule init/update` from anywhere in the tree.

ndokos (changes requested):

agent/rpm/Makefile
portante: Apparently not on a RHEL 7 box ...

npalaska (approved):

dbutenhof (approved):

Maxusmusti (approved):

agent/rpm/Makefile
ndokos: This is a shell mechanism: executing a command inside parens does it in a subshell, so changing the directory in the subshell does not affect the parent shell at all.

ndokos: I am skeptical: I'll go test.

ndokos (approved):

PR #1904, opened 2020-10-22
Design fixes for Documentation page

anishaswain:
Fixes:
- On hovers, need to turn to the light teal colour (#1BC1C9), not the dark 
- Several instances of Overpass font converted to Work Sans
- There was too much space around the cards (they were not left-aligned with the rest of the page content), and the three cards do NOT need to be sitting on an additional shadow card.
- The cards should all auto-adjust to be the height of the tallest card, making them all the same height 
- The three recent meetings cards were too far apart, so switched the spacing to `margin: var( --pf-global--spacer--md);`
- The `quicklinks` should be left-aligned. 
- The dropdowns for the cards (Pbench agent, Pbench server, and Dashboard) should NOT use all caps -- use sentence case.
- There was a bit of an issue with the interaction of the dropdown on the cards as well. When the card is expanded, the icon would be down, and it needs to swap to pointing to the right on closing.


portante (approved):

portante:
@anishaswain, can you rebase this PR?

portante (changes requested):
While this looks ready to go, it needs a rebase.  Thanks!

npalaska (approved):
Needs squashing I guess

index.css
```diff
@@ -103,6 +102,11 @@ li.PbenchQueItem::after {
    background-image: url(./images/WebServer.png);
  }
 
+ .div4{
```
dbutenhof, line 105: Doesn't "div4" seem like a fairly generic name for a very specific background pattern?

anishaswain:
closing in reference to #1917

PR #1906, opened 2020-10-22
Enhance agent container builds

portante:
We tackle three things with this commit.  This first is that we don't explicitly install `rh-python36`, we simply allow the `pbench-agent` RPM dependencies to drive that installation.

The second change is to make sure we build the name of the user for the COPR repo into the repo name to help debug where RPMs were pulled from at build time.

The last change is to ensure the proper package manager's cache (yum vs dnf) is removed depending on the distribution used.


ndokos (approved):
Looks reasonable.

riya-17 (approved):

dbutenhof (approved):

PR #1910, opened 2020-10-23
Backport container build enhancements to `b0.69`

portante:
There have been a number of enhancements to how we build containers that need to be available for building the `b0.69` branch.  We were successful using these changes to build the `v0.69.4-agent` release using them.

We had to remove the `atomic-openshift-clients` RPM dependency from the "tool" image since that repo is really per OpenShift version.

### DIFF
--- a/agent/containers/images/Dockerfile.layered.j2
+++ b/agent/containers/images/Dockerfile.layered.j2
@@ -7,4 +7,4 @@ FROM pbench-agent-base-{{ distro }}:{{ tag }}
 #        Kubernetes or RHV environments.
 RUN {% if distro == 'centos-7' %}yum{% else %}dnf{% endif %} install -y {% if distro == 'centos-8' %}--enablerepo PowerTools {% endif %}--enablerepo copr-pbench {{ rpms }} && \
     {% if distro == 'centos-7' %}yum{% else %}dnf{% endif %} -y clean all && \
-    rm -rf /var/cache/yum
+    rm -rf /var/cache/{% if distro == 'centos-7' %}yum{% else %}dnf{% endif %}
--- a/agent/containers/images/Makefile
+++ b/agent/containers/images/Makefile
@@ -12,6 +12,12 @@ USER = ndokos
 # using an environment variable as appropriate.
 URL_PREFIX = https://copr-be.cloud.fedoraproject.org/results/${USER}
 
+# By default we use non-"test" COPR repos named "pbench".  We expect
+# test COPR repos to have a suffix added, typically "test", so that
+# the final repo name would be "pbench-test".
+TEST =
+_TEST_SUFFIX = $(if $(TEST),-$(TEST),"")
+
 # By default we use the pbench Quay.io organization for the image
 # repository.  You can override this default using an environment
 # variable as appropriate.
@@ -23,7 +29,7 @@ _REPO_TEMPLATE = ../../ansible/pbench/agent/roles/pbench_repo_install/templates/
 
 # The list of RPMs which provide the various tools we offer.
 # Not intended to be overridden with an environment variable.
-_TOOL_RPMS = pbench-sysstat perf blktrace bpftrace numactl pcp-system-tools strace kernel-tools tcpdump
+_TOOL_RPMS = blktrace bpftrace cpupowerutils golang kernel-tools libvirt-client nmap-ncat numactl pbench-sysstat pcp-system-tools perf procps-ng strace tcpdump trace-cmd
 
 # The list of RPMs for the default workloads we offer.
 # Not intended to be overridden with an environment variable.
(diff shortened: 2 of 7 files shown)


agent/containers/images/Dockerfile.layered.j2
```diff
@@ -7,4 +7,4 @@ FROM pbench-agent-base-{{ distro }}:{{ tag }}
 #        Kubernetes or RHV environments.
 RUN {% if distro == 'centos-7' %}yum{% else %}dnf{% endif %} install -y {% if distro == 'centos-8' %}--enablerepo PowerTools {% endif %}--enablerepo copr-pbench {{ rpms }} && \
     {% if distro == 'centos-7' %}yum{% else %}dnf{% endif %} -y clean all && \
-    rm -rf /var/cache/yum
+    rm -rf /var/cache/{% if distro == 'centos-7' %}yum{% else %}dnf{% endif %}
```
dbutenhof, line 10: I wondered when I first saw this in the original PR why you didn't use `{{ pkgmgr }}` ... I figured it was probably good because this is technically a different context and while the `/var/cache/` directories apparently happen to match, it might be best not to depend on that.

However, I notice that in Dockerfile.base, you _did_ use `{{ pkgmgr }}` in the same context; it'd be good to be consistent, one way or the other.

(And yes, I know this is not the sort of comment one likes to see in a backport, which is of course why I noticed it now and not earlier!)

dbutenhof (approved):

agent/containers/images/Dockerfile.layered.j2
portante: We can't use `pkgmgr` for the other jinja2 invocations because it is not defined.  Only the base images have it defined explicitly.  I need to move the template generation into the image build script so that I don't need that in the `Makefile`.

portante: This is the kind of thing I'd like to fix on `master`, but not backport.

agent/containers/images/Makefile
```diff
@@ -32,108 +38,115 @@ _WORKLOAD_RPMS = fio uperf
 # Not intended to be overridden with an environment variable.
 _ALL_RPMS = ${_TOOL_RPMS} ${_WORKLOAD_RPMS}
 
+_DISTROS = centos-8 centos-7 fedora-32 fedora-31
+
 # Default we only build images the following distributions:
```
ndokos, line 43: This is both ungrammatical and misplaced: it should be above l.41.

agent/containers/images/Makefile
```diff
@@ -32,108 +38,115 @@ _WORKLOAD_RPMS = fio uperf
...
+push-major-minor: $(foreach distro, ${_DISTROS}, ${distro}-push-major-minor)
 
 #+
 # For the following rule patterns, the "%" represents the "distribution" name,
 # as derived from the "all" target's *-distro list.
+#
+# The string matching the "%" is called the "stem", in GNU Make parlance.  The
+# "$*" references are replaced with that stem value.
+#
+# See https://www.gnu.org/software/make/manual/make.html#Automatic-Variables
```
ndokos, line 82: Thank you!

agent/containers/images/Makefile
```diff
@@ -157,13 +170,13 @@ fedora-31-base.Dockerfile: Dockerfile.base.j2 fedora-31-pbench.repo
 
 # Rule pattern dependencies on non-patterned targets have to be set up
 # separately for some reason.
-%.repo: ${_REPO_TEMPLATE}
+%.repo: ${_REPO_TEMPLATE}
```
ndokos, line 173: Whitespace added?

ndokos (changes requested):

agent/containers/images/Makefile
portante: Fixed.

agent/containers/images/Makefile
portante: Fixed.

agent/containers/images/Makefile
```diff
@@ -32,108 +38,115 @@ _WORKLOAD_RPMS = fio uperf
 # Not intended to be overridden with an environment variable.
 _ALL_RPMS = ${_TOOL_RPMS} ${_WORKLOAD_RPMS}
 
-# Default we only build images the following distributions:
-all: centos-8-all-tagged centos-7-all-tagged fedora-32-all-tagged fedora-31-all-tagged
+# By default we only build images the following distributions:
```
ndokos, line 41: ... for the following ...

ndokos (changes requested):

agent/containers/images/Makefile
portante: Fixed.

ndokos (commented):
Late to the party but LGTM.

Issue #1790, opened 2020-09-10
Add support for Elasticsearch V7


portante:
We need to change our data models to support Elasticsearch V7 and later.

At the least we need to make sure we use one document type per index.

We also may have to ensure we don't use more than 1,000 fields per index.

Opportunistically:

- [ ] Leverage index life-cycle management (max index size, max doc count, max age)

- [ ] Leverage document-level security

portante:
Implemented as of #1914.

PR #1914, opened 2020-10-23
Move to Elasticsearch V7 for the pbench-server

dbutenhof:
These changes move Pbench from Elasticsearch V1 to Elasticsearch V7, and should be compatible with Elasticsearch V8.

* `indexer.py` changes for Elasticsearch V7
* Some miscellaneous fixes to `api/__init__.py` including getting rid of the annoying `global api`
* A new web service `queryControllers` that replaces the current dashboard `services/dashboard.ps` function `queryControllers`, including the logic to formulate the list of Elasticsearch indices corresponding to the input date range. This API (`/api/v1/controllers/list`) returns a JSON payload equivalent to the current Dashboard `src/models/dashboard.js` method `fetchControllers`.
* Unit testing for `queryControllers`, including payload post-processing.

Some specific Elasticsearch 7 changes:

* Removal of the document name in mapping documents; this was deprecated in ES6, and will no longer be allowed in ES8
* Restriction of indices to a single document type required splitting the "run" and "result-data" indices each into two separate indices each containing a single document type: "run-data", "run-toc", "result-data-sample" and "result-data".
* The "_parent" parent/child reference system is also deprecated: it required that parent and child documents be within the same index, in any case, which is no longer possible. Although it seems the dashboard does not currently rely on this anyway, I've replaced these metadata links with explicit schema properties referencing the parent document ID, "result_data_sample_parent" and "run_data_parent".
* Removal of the "_type" property in indexing actions
* The version of all indices has been increased by 1.
* Instead of requiring "_type" in captured actions, the unit test Elasticsearch mock object now reports an error if one is seen.

Note that these changes have rippled through all of the server `gold` output files. Most of the changes are straightforward according to the specific changes above.

It might however be valuable to briefly discuss some non-obvious behavior of the Elasticsearch mock that has potentially confusing impact on a number of the tests (particularly in the `test-7.x` range). It keeps track of the index names associated with each bulk index action. It counts the number of times each of those index names occur. In order to limit the sometimes massive output, it _analyses and records_ only the first __15__ documents on each index. Because this change splits apart two of the largest indices, this means that the number of reported actions for tests may increase by up to 30. Each test reports the total count for each index, plus the number of captured (and analyzed) indices. For example,

```
*** 73,81 ****
  Template:  pbench-unittests.v5.result-data
  Template:  pbench-unittests.v5.result-data-sample
  Template:  pbench-unittests.v6.run-data
! Index:  pbench-unittests.v5.result-data.2019-06-21 1810
! Index:  pbench-unittests.v6.run-data.2019-06 57
! len(actions) = 30
  [
      {
          "_id": "5ca1ab1e70015f100dedfab1ed0ff1ce",
--- 73,84 ----
  Template:  pbench-unittests.v5.result-data
  Template:  pbench-unittests.v5.result-data-sample
  Template:  pbench-unittests.v6.run-data
! Template:  pbench-unittests.v6.run-toc
! Index:  pbench-unittests.v5.result-data-sample.2019-06-21 10
! Index:  pbench-unittests.v5.result-data.2019-06-21 1800
! Index:  pbench-unittests.v6.run-data.2019-06 1
! Index:  pbench-unittests.v6.run-toc.2019-06 56
! len(actions) = 41
  [
      {
          "_id": "5ca1ab1e70015f100dedfab1ed0ff1ce",
```

The above shows that the previous gold file has 1810 `result-data` documents and 57 `run-data` documents. Therefore, the `len(actions)` reported is 30, including 15 documents from each.

In contrast, the new results show 10 documents from `result-data-sample`, 1800 from `result-data` (the child of `result-data-sample`), 1 from `run-data`, and 56 from `run-toc` (the child of `run-data`). Therefore, the new `len(actions)` is 15 from each of `result-data` and `run-toc` (30), plus all 10 from `result-data-sample` and 1 from `run-data` for a new total of 41.

The `test-7.16` is an intriguing special case. It was deliberately provoking errors in the scanner that checks action source document keys against the document mapping files, by forcing the `indexer.py` logic to rename conflicting key names collected across multiple result data source files. These keys, `clients` and `servers` are renamed with `<tool>_` with increasing underscores until uniqueness is achieved. The Elasticsearch mock reports messages that these renamed keys are not in the mapping files, which is of course correct and expected. These occur in the `result-data-sample` documents, and are part of the base `"benchmark"` payload that's reproduced across all sample documents: however, because we previously collected only 15 documents from the index shared by `result-data-sample` and `result-data`, we'd captured only one `result-data-sample` and filled the remaining 14 slots with `result-data` child documents. In the new version, we capture all four `result-data-sample` documents... and therefore the reported errors are duplicated four times.

This is just a brief summary of the enormous "fun" I had in triaging the unit test differences, but I thought I'd share.

To do: (not exhaustive)

- [x] Unit tests

- [ ] Need a clean way to make server configuration open the configured backend port (e.g., something like adding `- 8001` to the current `pbench-server-firewall` ansible role's list of protocols to enable ... except we need to parse the actual configured port out of `pbench-server.cfg`

- [x] I need to decide (i.e., I'm seeking input) on the final response payload for `queryControllers` ... right now I'm just returning the raw Elasticsearch JSON. This makes the new endpoint a plug-and-play replacement for the current service, but I'd rather streamline it and replace the `dashboard.js` model code as well.

- [ ] I'm tempted to push the other Flask Response classes out into separate files as well to keep the sprawl under control; opinions?

See issue #1790.


lib/pbench/server/api/__init__.py
```diff
@@ -214,42 +225,43 @@ def post(self):
...
+                    headers={
+                        "Content-Type": "application/json",
+                        "Accept": "application/json",
+                    },
+                    json=json_data["payload"],
                 )
-                response = requests.get(url)
+            else:
+                logger.info("No payload found in Elasticsearch request, using GET")
+                response = requests.get(url, headers={"Accept": "application/json"},)
```
npalaska, line 238: Now that I think about it we can probably have a separate definition for `get` method ?

npalaska (commented):
Looks good at the first glance, I will do detailed review later. I agree we can break Flask Response classes out into separate files (in fact we can have separate class for each flask class I think)

lib/pbench/server/api/__init__.py
dbutenhof: When we're done, the pass-through Elasticsearch and GraphQL APIs should be gone anyway in favor of native Pbench APIs that target dashboard semantics directly, so I'm not sure this is worth worrying about too much right now.

npalaska: yep and with `querycontroller` we would have a targeted API so we can build the payload inside the server, then we can have multiple such targeted endpoints and we can remove the elasticsearch endpoint from the server

lib/pbench/server/api/resources/query_controllers.py
```diff
@@ -0,0 +1,173 @@
...
+            [string]: A comma-separated list of month-qualified index names
+        """
+        startDate = parser.parse(start)
+        endDate = parser.parse(end)
+        referenceDate = deepcopy(startDate)
+        monthResults = list()
+        queryString = ""
+        oneMonth = relativedelta.relativedelta(months=1)
+        if startDate and endDate:
+            while referenceDate < endDate:
```
npalaska, line 51: Can we use a for loop here so we dont need to keep the copy of referenceDate there is a [rrule](http://labix.org/python-dateutil#head-470fa22b2db72000d7abe698a5783a46b0731b57) module in `python-dateutil` that we can use to get the range of months for example 
```
for dt in rrule.rrule(rrule.MONTHLY, dtstart=start, until=end):
    print(dt)
```

lib/pbench/server/api/resources/query_controllers.py
```diff
@@ -0,0 +1,173 @@
...
+            abort(
+                500,
+                message="Connection timed out, could not post to Elasticsearch Endpoint",
+            )
+        except requests.exceptions.InvalidURL:
+            logger.exception("Invalid url during the Elasticsearch post request")
+            abort(
+                500,
+                message="Invalid Elasticsearch url, could not complete the post request",
+            )
```
npalaska, line 160: Minor, but we can add `HTTPError` exception as well if the HTTP request to Elasticsearch returned an unsuccessful status code [requests exceptions](https://requests.readthedocs.io/en/latest/user/quickstart/#errors-and-exceptions)

dbutenhof: I stole most of this from the existing Elasticsearch query; but that's probably a good idea.

lib/pbench/server/api/resources/query_controllers.py
dbutenhof: Cool. I did not know that.

dbutenhof:
FYI: I got the server unit tests working, which was a pain, and I pushed a new commit to record the mess.

I'll act on Nikhil's comments, and add new `queryController` unit testing next...

lib/pbench/server/api/resources/query_controllers.py
```diff
@@ -103,42 +103,61 @@ def post(self):
         Content-Type:   application/json
         Accept:         application/json
 
-        Returns the Elasticsearch hits:hits:_source list.
-
-        {
-
-        }
+        Return payload is a summary of the "aggregations" propery of the
```
portante, line 106: `s/propery/property/`

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -170,3 +172,204 @@ def test_upload(client, pytestconfig, caplog):
...
+        assert res_json[0]["results"] == 2
+        assert res_json[0]["last_modified_value"] == 1.59847315581e12
+        assert res_json[0]["last_modified_string"] == "2020-08-26T20:19:15.810Z"
+        assert res_json[1]["key"] == "unittest-controller2"
+        assert res_json[1]["controller"] == "unittest-controller2"
+        assert res_json[1]["results"] == 1
+        assert res_json[1]["last_modified_value"] == 1.6
+        assert res_json[1]["last_modified_string"] == "2020-09-26T20:19:15.810Z"
+
+    @pytest.mark.parametrize(
```
portante, line 354: Very cool!

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -170,3 +172,204 @@ def test_upload(client, pytestconfig, caplog):
...
+
+    def build_index(self, client, dates):
+        """
+        build_index Build the index list for query
+
+        Args:
+            dates (iterable): list of date strings
+        """
+        prefix = client.config["PREFIX"]
+        idx = prefix + ".v6.run-data."
```
portante, line 228: `v6` seems like a hard-coded value, is there a way we could avoid this without too much hassle?

lib/pbench/server/api/__init__.py
```diff
@@ -21,6 +21,8 @@
 from pbench.common.exceptions import BadConfig
 from pbench.server.utils import filesize_bytes
 
+from pbench.server.api.resources.query_controllers import QueryControllers
+
 ALLOWED_EXTENSIONS = {"xz"}
 
 app = None
```
portante, line 28: Can this global variable be removed?

lib/pbench/server/api/resources/query_controllers.py
```diff
@@ -0,0 +1,273 @@
...
+    """
+
+    def __init__(self, app, api):
+        self.app = app
+        self.api = api
+
+    @staticmethod
+    def _genMonthUriRange(prefix, index, start, end):
+        """
+        _genMonthUriRange Construct a comma-separated list of index names
```
portante, line 20: Since we are going to Elasticsearch V7, we should consider using the index life-cycle management, where we fill an index with data, and have it auto roll-over to another index name for the next one.

Then we just use a date range query across all the indices.  At least that is what the latest Kibana does.  We might want to consider this instead of building a long list of indices and handling the roll-over that way.

lib/pbench/server/api/resources/query_controllers.py
```diff
@@ -0,0 +1,273 @@
...
+        The Elasticsearch URL string is also dependent upon the configured
+        "prefix". This is a Pbench artifact that allows multiple Pbench
+        instances to share a single Elasticsearch server, by segregating
+        their data into unique namespaces. Each Pbench index name is
+        qualified by the index_prefix of the Pbench server that originally
+        indexed the data. The prefix comes from the pbench-server.cfg
+        file for the server instance. For example, the official staging
+        server index_prefix is "staging-pbench" so that data is separate
+        from the production server even though they share a single instance
+        of Elasticsearch.
```
portante, line 88: While the sentiment of this last sentence is correct, the specifics are not quite as described.  It might better to document this hypotheically, ... "For example. a staging server `index_prefix` might be `staging-pbench`, while a production server `index_prefix` might be `prod-pbench`, so that data from the two pbench-server instances does not conflict in the same Elasticsearch instance."   ... Or something like that.

portante (changes requested):
Wow, this is great!

We should talk about [index life-cycle management](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-lifecycle-management.html) to see if there are any gains to be had with that.

lib/pbench/server/indexer.py
```diff
@@ -1168,7 +1113,6 @@ def _make_user_benchmark_json(self):
                     ]
                 )
                 # Yield the result-data document.
-                _parent = _id
                 _id = PbenchData.make_source_id(source, _parent=_parent)
```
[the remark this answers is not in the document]
dbutenhof: Probably. I almost did that, decided to put it off due to the complication of the nested generators, and then my stack overloaded and I completely lost it.

lib/pbench/test/unit/server/test_requests.py
dbutenhof: Yeah; it's hardcoded in `query_controllers.py`, too: check out the large `TODO` at line 200.

The Dashboard handles this as a config file option. While we could do that too, ideally I'd like to get it from the template files when we first load them... e.g., store the version for each "root" index name in a persistent DB (Redis?). That'll be relatively straightforward when we go to microservices.

Another option would be for `indexer.py` to write a JSON file somewhere e.g. under {pbench-top-dir} to store the mapping, which other components could reference. In any case, unless you're going to strong-arm my conscience right now, my thought was to create an issue and defer that.

.gitignore
```diff
@@ -19,3 +19,5 @@ pip-wheel-metadata
 .vscode
 Pipfile
 Pipfile.lock
+AUTHORS
```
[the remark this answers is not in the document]
dbutenhof: This may be because I didn't re-evaluate after merging with recent changes; but I built RPMs several times, and git was complaining about AUTHORS and ChangeLog so I just added them. I guess I'll try again and confirm... and... yeah, I deleted `AUTHORS` and `ChangeLog` and built a new COPR server RPM. The files failed to reappear, so I'll drop these back out.

lib/pbench/server/indexer.py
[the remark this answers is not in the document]
dbutenhof: Done.

lib/pbench/server/api/__init__.py
dbutenhof: Um, yes; but doing so revealed a few interesting artifacts I'd missed. So... good catch.

lib/pbench/server/api/resources/query_controllers.py
dbutenhof: A battle for another day.

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -57,19 +52,29 @@ def test_empty_url_path(client, caplog):
...
-            status=400,
+    def test_bad_request(client, caplog, requests_mock):
+        requests_mock.post(
+            "http://elasticsearch.example.com:7080/some_invalid_url", status_code=400
         )
+        # mocked_responses.add(
+        #     responses.POST,
+        #     "http://elasticsearch.example.com:7080/some_invalid_url",
+        #     status=400,
+        # )
```
npalaska, line 63: commented code?

pytest.ini
```diff
@@ -1,4 +1,4 @@
 [pytest]
-opts = -s
+addopts = -s
```
npalaska, line 2: whats the difference betwen` opts` and `addopts`?

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -170,3 +175,204 @@ def test_upload(client, pytestconfig, caplog):
...
+                "controllers": {
+                    "doc_count_error_upper_bound": 0,
+                    "sum_other_doc_count": 0,
+                    "buckets": [
+                        {
+                            "key": "unittest-controller1",
+                            "doc_count": 2,
+                            "runs": {
+                                "value": 1.59847315581e12,
+                                "value_as_string": "2020-08-26T20:19:15.810Z",
```
npalaska, line 325: why is `value_as_string` a datetime vale?

npalaska (commented):
Looks good overall, I just think we need to break down the api `__init__` file into multiple modules so easier to mange each APIs and its cleaner and have init file just do the high level stuff than defining the APIs itself. I did addressed this in my PR since I was encountering some circular imports because of the global variables

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -4,18 +4,11 @@
...
-import responses
 
 from pathlib import Path
 from werkzeug.utils import secure_filename
 
 
-@pytest.fixture
-def mocked_responses():
-    with responses.RequestsMock() as rsps:
-        yield rsps
```
dbutenhof, line 14: I tried to change my mocks to use this model, but I wasn't able to find any way to get my exception tests to pass. My guess, although I couldn't find enough context to prove it, is that the `RequestsMock` overloading of `body` instead of having a separated `exc` argument, was interfering with the payload in the tested API's `post`, and generating type errors.

I couldn't find any way to fix this, so instead of converted your one mock to use the `requests_mock` package, instead. In theory, I think we could therefore drop `responses` from the `pbench-devel` container build, but I decided not to be that aggressive. There doesn't appear to be an RPM for `requests_mock`, either, so we'll have to rely on the pip install inside the container.

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -59,19 +52,29 @@ def test_empty_url_path(client, caplog):
...
         assert response.json is None
-        assert len(caplog.records) == 0
+
+        # This is a bit awkward, but the requests_mock throws in its own
+        # DEBUG log record to announce the POST; so allow it. (But don't
+        # fail if it's missing.)
+        if len(caplog.records) > 0:
+            assert len(caplog.records) == 1
+            assert caplog.records[0].levelname == "DEBUG"
+            assert caplog.records[0].name == "requests_mock.adapter"
```
dbutenhof, line 77: And the mock requests for some reason is generating its own `DEBUG` log here; so I just accounted for it. (Sigh.)

lib/pbench/test/unit/server/test_requests.py
dbutenhof: Ah; that was Pete's old mock REST code. I commented it instead of deleting it, and then forgot about it.

pytest.ini
dbutenhof: Pragmatically, `addopts` is documented and appears to work, while (at least in my configuration) `opts` generates warnings we may have just been ignoring.

lib/pbench/test/unit/server/test_requests.py
dbutenhof: Because it's the highest value of the selected aggregate  `"aggs": {"runs": {"max": {"field": "run.start"}}}` in a set of controllers found by the search query. Elasticsearch represents the result in both binary and string form, and the existing Dashboard model picks those out so I needed to put them in my sample response.

dbutenhof:
> Looks good overall, I just think we need to break down the api `__init__` file into multiple modules so easier to mange each APIs and its cleaner and have init file just do the high level stuff than defining the APIs itself. I did addressed this in my PR since I was encountering some circular imports because of the global variables

Yes; I started to do that with `query_controllers.py`; but ideally we should not only unbundle the other resources but also farm out the URL registration. `Flask` has mechanisms to do that, but I decided that the point of this PR was really the Elasticsearch 7, and I coded my one new API more as a prototype to lead the way and demonstrate what we need to do. Major restructuring of `/api/__init__.py` should be done in its own PR. Maybe before yours, if you think that would help you to avoid pulling API-isms into the low level support classes and vice versa.

lib/pbench/server/api/__init__.py
```diff
@@ -142,41 +159,43 @@ def options(self):
...
-            f"http://{app.config_graphql.get('host')}:{app.config_graphql.get('port')}"
-        )
+        logger = self.app.logger
+        gql = self.app.config_graphql
+        graphQL = f"http://{gql.get('host')}:{gql.get('port')}"
         json_data = request.get_json(silent=True)
 
         if not json_data:
-            app.logger.warning("GraphQL: Invalid json object {}", request.url)
+            logger.warning("GraphQL: Invalid json object %s", request.url)
```
portante, line 175: Shouldn't this be `{}` still?

lib/pbench/server/api/resources/query_controllers.py
```diff
@@ -0,0 +1,273 @@
...
+            index ([string]): The desired monthly index root
+            start ([datetime.datetime]): The start time
+            end ([datetime.datetime]): The end time
+
+        Returns:
+            [string]: A comma-separated list of month-qualified index names
+        """
+        monthResults = list()
+        queryString = ""
+        for m in rrule.rrule(rrule.MONTHLY, dtstart=start, until=end):
```
portante, line 45: Does `rrule` validate that `start` and `end` are valid dates to begin with?

lib/pbench/server/api/resources/query_controllers.py
```diff
@@ -0,0 +1,273 @@
...
+            monthResults.append(m.strftime("%Y-%m"))
+        for monthValue in monthResults:
+            if index == "v4.result-data.":
```
portante, line 48: This is kinda intimate knowledge of the index name pattern.  Could we pass an argument to direct the addition of the `-*` suffix instead?

lib/pbench/server/api/resources/query_controllers.py
```diff
@@ -0,0 +1,273 @@
...
+                500,
+                message=f"Invalid Elasticsearch url {uri}, could not complete the post request",
+            )
+        except Exception as e:
+            logger.exception(
+                "Exception {!r} occurred during the Elasticsearch post request", e
+            )
+            abort(500, message=f"Could not post to Elasticsearch: {e!r}")
+        else:
+            es_json = es_response.json()
```
portante, line 256: It seems like we're going to need to wrap this entire structure with a `try/except` block, not just for lines 260 - 267 below.

lib/pbench/server/api/resources/query_controllers.py
```diff
@@ -0,0 +1,273 @@
...
+            logger.info("{} hits found", len(es_json["hits"]["hits"]))
```
portante, line 257: We might want to add some context value to match the POST log early with this log message.  When we are searching through lots of logs on the server, it will help to match them up.

lib/pbench/server/api/resources/query_controllers.py
portante: Don't we need to verify the response code as well? If we receive a 400 error, wouldn't we want to deal with that properly instead of 500 error logged?

lib/pbench/server/indexer.py
```diff
@@ -46,9 +47,11 @@
 from pbench.server import tstos
 
 try:
-    from elasticsearch1 import Elasticsearch
+    from elasticsearch import Elasticsearch, VERSION
+
+    assert VERSION[0] == 7, "Pbench currently requires Elasticsearch V7.x"
```
portante, line 52: Yay!

portante (changes requested):
Looking really good!  I think we just want to fix a few things, and then merge this down to one commit.  This is great!

lib/pbench/server/api/__init__.py
dbutenhof: This got pulled in by a merge, I think, and I missed the `%s`.

lib/pbench/server/api/resources/query_controllers.py
dbutenhof: I really need to work out a better model for handling index versioning; this check comes directly from the Javascript. It's not even immediately relevant since it's not used (yet) for result data. This really should key off the index_patterns[root_name]['template_pat'] data, dynamically... but we don't have that until I work out a way to process and cache the templates.

I left this more as a template and marker than anything else; but maybe I should at least add a TODO here so it doesn't get lost.

lib/pbench/server/api/resources/query_controllers.py
dbutenhof: I was assuming that we'd never get a non-200 return without an exception. Is that not true?

Plus, aside from 408, which ought to be a Timeout exception, I'm not sure how we could handle an ES 400 error except as a 500 ... either we failed to properly validate our parameters from the client, or we fabricated the ES URI or payload incorrectly. Do you have any specific recommendations?

And, yeah; I should have put the `es_response.json()` inside the `try` in case there's no JSON response at all.

lib/pbench/server/api/resources/query_controllers.py
dbutenhof: I had thought at one point that an info log for each call might be useful; we can get the API log from httpd, but that doesn't have any payload data if we really want to record that. The tradeoff is that we can get a lot of API hits. I've always gone back and forth on that tradeoff; this time I decided not to. But then I went and added this `info` during debugging, and decided not to remove it, so we're getting one for each (successful) call anyway. Would two be so much worse? ;-)

lib/pbench/server/api/resources/query_controllers.py
dbutenhof: `__getMonthUriRange` takes `datetime.datetime` objects, not raw strings; validation is done in the caller:

```
try:
    start = parser.parse(start_arg)
    end = parser.parse(end_arg)
except Exception as e:
    [...] abort(400, ...)
```

lib/pbench/server/api/resources/query_controllers.py
dbutenhof: Actually, there's already a `info` level message after parsing the input succeeds:

```
        logger.info(
            "QueryControllers POST for user {}, prefix {}, session {}: ({} - {})",
            user,
            prefix,
            session_token,
            start,
            end,
        )
```

lib/pbench/server/api/resources/query_controllers.py
dbutenhof: > Don't we need to verify the response code as well? If we receive a 400 error, wouldn't we want to deal with that properly instead of 500 error logged?

This turns out to be one of those unexpectedly prescient questions, Pete. I investigated the `requests` documentation to see if it's possible to return an error without raising an exception.

And, um, it turns out that `requests` __never__ raises an exception. Unless you tell it to. I verified this by breaking my ES query and tickling the modified server to get an ES 400 response. It went right through my `else:`.

Turns out there's a `raise_for_status` method on the response object that'll raise the appropriate exception if the response isn't a 200 range success `status_code`. I've added that following my call as well as in the pass-through ES and GQL queries!

dbutenhof:
I squashed down to 2 commits, keeping the review comment fixes separate for convenience.

dbutenhof:
And ... that broke Pete's `test_bad_request` because the ElasticSearch resource object wasn't handling HTTPError and was returning a generic 500 status.

dbutenhof:
I couldn't figure out why it failed again: I forgot to _stage_ my fix before I pushed the update. (Sigh.) Rushing always makes things go slower, doesn't it?

portante (changes requested):
Needs a rebase, but looks good!

portante (changes requested):
Looks like unit failures still...

dbutenhof:
I published `pyesbulk` 2.1.1, and rebased; the tests are clean. Whew.

portante (dismissed):

npalaska (approved):
looks good

dbutenhof:
The `pbench/common/logger.py` traceback is fixed.

This seems to have been an old problem that something new I pulled in (possibly `requests_mock`) is now triggering in my `query_controllers.py` context. The `_PbenchLogFormatter.format()` method tries to format UTC time from a current value of `record.asctime` if the property exists. However, if it does exist, it should be a string, which causes `datetime.utcfromtimestamp()` to throw `AttributeError` ... and it does, but only in this context.

Because the `asctime` property is reserved and defined, but not populated until it hits a formatter, this suggests that we've never before run with a competing formatter that added `asctime`, and we've always taken the `AttributeError` and generated a fresh timestamp. Now, something else is populating `asctime`, with a string, and we're taking an unhandled `TypeError` instead, which was generating the traceback.

My fix is to format `record.created`, which is the `time.time()` value captured when the `LogRecord` was created. I left the `AttributeError` handler, although `record.created` should always exist ... so I doubt we'll hit it anymore.

lib/pbench/common/logger.py
```diff
@@ -88,7 +88,7 @@ def format(self, record):
         record.message = record.getMessage()
         if self._fmt.find("{asctime}") >= 0:
             try:
-                record.asctime = datetime.utcfromtimestamp(record.asctime).isoformat()
+                record.asctime = datetime.utcfromtimestamp(record.created).isoformat()
```
portante, line 91: Can we make this change in a separate PR?  I am not sure about this one, and would like to consider this on its own.

portante (changes requested):

lib/pbench/common/logger.py
dbutenhof: Sure; mostly I wanted to be sure it really fixed the problem in Jenkins, since that's behaving differently from my local environment. Since the tests were succeeding before despite the traceback, I guess it's safe to back that out for this PR.

dbutenhof:
Note that the code now should be identical to the code before I tried to fix the logger. We're getting the weird traceback in the pytest runs, but it's succeeding despite the uncaught exception.

portante:
> Note that the code now should be identical to the code before I tried to fix the logger. We're getting the weird traceback in the pytest runs, but it's succeeding despite the uncaught exception.

Great, we can track the issue in #2003.

riya-17 (approved):

riya-17 (commented):
looks good

aquibbaig (commented):
@dbutenhof Great work! :rocket:

npalaska (dismissed):

portante (approved):

ndokos (approved):

PR #1915, opened 2020-10-24
Remove SCL from the agent side

portante:
Remove all references to "SCL" throughout the code, relying on packaging and an installation procedures to provide a proper Python 3 environment.

We use Python 3.6 explicitly for RHEL 8 / CentOS 8 to avoid packages conflicts which don't work with one or the other (e.g. python3-requests is an RPM that only works with Python 3.6, not Python 3.8).

We have refined the RPM list as a result for this work, and dropped the code to generate different "final" `requirements.txt` files based on the platform.  This is because a `pip3 install` will only install modules that are not already found on the `PYTHONPATH`.  RPMs which satisfy the module requirements will prevent `pip3` from installing it locally.

Further, dnf caches repo states in weird ways, so when building "all" the images, we use a new `all-tags` target which helps ensure we get the same tags for all images.  We want individual tags per image in general, since we don't want to force somebody to have to build everything.  But *when* building everything we need the consistency.

Finally, we add the `all-dockerfiles` convenience target to build all the `.repo` and `.Dockerfile` files that would be used.


dbutenhof (approved):
Lots of deleted code; I like it.

Maxusmusti (approved):

portante:
While these changes are nice, they are not sufficient based on testing on RHEL 7, RHEL 8, and various containers.

dbutenhof:
> While these changes are nice, they are not sufficient based on testing on RHEL 7, RHEL 8, and various containers.

SCL: the never-ending gift.

agent/rpm/pbench-agent.spec.j2
```diff
@@ -9,36 +9,45 @@ Source0:        pbench-agent-%{version}.tar.gz
 Buildarch:      noarch
 
 %if 0%{?rhel} == 7
-Requires:  scl-utils, rh-python38, rh-python38-python-pip, rh-redis5
+Requires:  python3, python3-pip
+# For RHEL boxen we need the python development environment in order for the
+# pip3 installs for some modules to be successful.
+Requires:  gcc python3-devel
```
ndokos, line 15: Is this required  only for RHEL7?

agent/containers/images/Dockerfile.base.j2
```diff
@@ -8,9 +8,10 @@ COPY ./{{ pbench_repo_file }} /etc/yum.repos.d/pbench.repo
 # ... and make sure we have a proper pbench-agent.cfg file in place;
 # ... and finally, ensure the proper pbench-agent environment variables are set up.
 RUN \
-{% if distro_image.startswith('centos') %}
-    {{ pkgmgr }} install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-{{ distro_image.split(':', 1)[1] }}.noarch.rpm{% if distro_image == 'centos:7' %} centos-release-scl-rh centos-release-scl{% endif %} && \
+{% if distro_image == 'centos:8' %}
+    {{ pkgmgr }} module -y enable python36 && \
+    {{ pkgmgr }} module -y disable python38 && \
```
dbutenhof, line 13: Huh. Interesting. My RHEL 8 laptop shows Python36 [d][e] and Python38 [d]. Never even knew about "modules" before.

agent/rpm/python-module-mapping.table
```diff
@@ -1,13 +1,13 @@
...
-werkzeug         python3-werkzeug   -
+Module           Fedora (31/32)     EL 7      EL 8               Notes
+===============  =================  ========  =================  =====
+ansible          n/a                -         -                  The "ansible" package is available, but the Python modules are not installed in the Python3 site-packages, so we still have to pip3 install that module.
+bottle           python3-bottle     -         -
+cffi             python3-cffi       -         python3-cffi
+click            python3-click      -         python3-click
+colorlog         python3-colorlog   -         -
+python-daemon    python3-daemon     -         -                  Typically needs the "lockfile" and "docutils" packages installed as well
+python-pidfile   -                  -         -                  Neither Fedora or RHEL package the pidfile module, needs "psutil" which is typicaly packaged
```
ndokos, line 9: "... nor RHEL"

dbutenhof (dismissed):
I thought we had access to the module mapping table down to just `gen_final_requirements` which implies its no longer used at all. Or is it just left for documentation? (Which isn't a bad thing, but files out of the executable code tend to get stale quickly.)

agent/rpm/pbench-agent.spec.j2
```diff
@@ -63,41 +72,25 @@ mkdir -p %{buildroot}/%{installdir}
...
-%endif
-
 # Install python dependencies
-%if 0%{?rhel} == 7
-scl enable rh-python38 -- bash -c "pip3 --no-cache-dir install --prefix=/%{installdir} -r /%{installdir}/requirements.scl" > /%{installdir}/pip3-install.log
-%endif
-%if 0%{?rhel} == 8 || 0%{?fedora} >= 31
-pip3 --no-cache-dir install --prefix=/%{installdir} -r /%{installdir}/requirements.f3x > /%{installdir}/pip3-install.log
-%endif
+pip3 --no-cache-dir install --prefix=/%{installdir} -r /%{installdir}/requirements.txt > /%{installdir}/pip3-install.log
```
ndokos, line 81: I need a memory jog: we manufacture different `requirements.txt` files for the different distros, based on the mapping table - correct? And that change has already been merged - correct?

ndokos (changes requested):

agent/rpm/pbench-agent.spec.j2
dbutenhof: Ah, yes; I think that's correct, and I'd forgotten it. This has gone through so many permutations, my brain is still bouncing.

ndokos: I'm glad I'm not the only one :-)

agent/containers/images/Dockerfile.base.j2
portante: Yeah, this is wacky.

agent/rpm/pbench-agent.spec.j2
portante: Yes.  :(

Basically, `python-pidfile` has a dependency on `psutils`.  To `pip3 install` `psutils` you need `gcc` and the Python3 development environment in order to build `psutils`.  Since the Fedoras and RHEL 8 both have a `psutils` RPM, we don't need those there, but for RHEL 7 we do.

agent/rpm/pbench-agent.spec.j2
portante: I have addressed this in the commit updated commit message and in the description.

agent/rpm/python-module-mapping.table
portante: Fixed.

dbutenhof:
> We have refined the RPM list as a result for this work, and dropped the code to generate different "final" `requirements.txt` files based on the platform. This is because a `pip3 install` will only install modules that are not already found on the `PYTHONPATH`. RPMs with satisfy the module requirements will prevent `pip3` from installing it locally.

Typo in last sentence: "RPMs __which__ satisfy ..."

dbutenhof (dismissed):

dbutenhof (approved):

portante:
RPM from this commit, 13d83abd8, are available from: https://copr.fedorainfracloud.org/coprs/portante/pbench-test/builds/, see pbench-agent-0.71.0-31g13d83abd8.

portante:
Container images will be available at https://quay.io/organization/pbench with tag 13d83abd8.

Maxusmusti:
I can try installing rpm on RHEL7 to test it out at around 8pm

Maxusmusti (dismissed):
Worked flawlessly on RHEL7, tested with both traditional and persistent tools (except for a new major bug I have noticed, but unrelated to this PR [filed in PR #1916])

agent/containers/images/verify-tags
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+if [[ -z "${1}" ]]; then
+    printf -- "No tags files given to process!\n" >&2
+    exit 1
+fi
+
+# Use the first argument to the as the check-sum to verify against.
```
ndokos, line 8: Delete "to the"?

agent/rpm/pbench-agent.spec.j2
ndokos: Sigh...

ndokos (changes requested):
A few nits. BTW thanks for the requirements.txt explanation: it had not sunk in before.

agent/containers/images/verify-tags
portante: Fixed.

portante:
RPMs from this commit, 000e0e7, are available from: https://copr.fedorainfracloud.org/coprs/portante/pbench-test/builds/, see pbench-agent-0.71.0-36g000e0e70c.  And images will be built and pushed to https://quay.io/organization/pbench with tag 000e0e7.

Maxusmusti (approved):
Re-tested rpm, works without any bugs now, nice

dbutenhof (approved):
"What a long, strange trip it's been..."

ndokos (approved):

PR #1916, opened 2020-10-27
Fix for missing templates directory

ndokos:
@Maxusmusti ran into this problem just now: the agent/templates directory (and the prometheus.yml file in it) was not installed.

### DIFF
--- a/agent/rpm/pbench-agent.spec.j2
+++ b/agent/rpm/pbench-agent.spec.j2
@@ -114,6 +114,7 @@ fi
 
 /%{installdir}/lib
 /%{installdir}/ansible
+/%{installdir}/templates
 /%{installdir}/base
 
 /%{installdir}/VERSION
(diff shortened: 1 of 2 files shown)


Maxusmusti (approved):

agent/rpm/pbench-agent.spec.j2
```diff
@@ -123,6 +123,7 @@ fi
 /%{installdir}/requirements.txt
 /%{installdir}/requirements.f3x
 /%{installdir}/requirements.scl
+/%{installdir}/templates/prometheus.yml
```
portante, line 126: Just like `lib`, `ansible`, `stockpile`, `util-scripts`, etc. you don't need to explicitly mention `prometheus.yml`, since when we add others, we'll want those automatically included.

The only reason you need to mention a file individually is if you want to set permissions on it different then the presiding defaults.

portante (changes requested):

portante (changes requested):
Seems like we still are specifying one file when we don't have to.  Is there something I am missing?

ndokos:
Sorry - forgot to use -f and the push didn't go through. Should be OK now.

npalaska (approved):

riya-17 (approved):

dbutenhof (approved):
OK; although these seem to have already been sucked in to Pete's RPM/container #1915.

ndokos:
These changes have been incorporated into #1915.

portante (approved):

portante:
Reopened to make sure it was resolved by #1915, closing after confirming.

PR #1932, opened 2020-10-29
Add our full unit test suite

portante:


dbutenhof (commented):
Cool!

zulcss:
*cough* https://docs.openstack.org/infra/jenkins-job-builder/

portante:
> _cough_ https://docs.openstack.org/infra/jenkins-job-builder/

Based on a quick read of that site it does not seem to help us much.  We are trying to setup our test environment in a container, and then get Jenkins to run the various tests via `podman run` using that container.

If there is a way to describe that through this builder, that would be great, but this looks very much like an openstack.org specific utility.

dbutenhof (approved):
Wow. I played a bit with a pipeline for `pyesbulk` and your new container will help. Clearly you've learned some more about the command environment for the pipeline `sh` commands, too, which is great.

portante:
Merging this onto the `jenkins` branch so that I can create a proper PR to propose merging to `master`.

portante:
The unit tests are failing because the server's `test-26.2` is failing due to the containerized environment not having a `/dev/log` device to test against.  I'll be proposing a refactoring of that test to avoid requiring `/dev/log`.

PR #1934, opened 2020-10-31
Use `_pbench_` prefixed env var for host names

portante:
Instead of accepting `hostname` and `full_hostname` as pre-defined environment variables, we rename them to be `_pbench_hostname` (contains `$(hostname -s)`) and `_pbench_full_hostname` (contains `$(hostname -f)`).  This prevents an inadvertent environment variables of the same name from changing the behavior of the agent.

We also consistently use those environment variables where possible to avoid disagreements on the host name used.

Some of the Python 3 code changed, and so `flake8` detected a problem.  Why it wasn't detected before is a mystery, but we fix it here as well.

We also take the opportunity, in separate commits, to stop embedding unit test logic in `agent/base`, and make sure we explicitly ask for the full hostname using `hostname -f`.

### DIFF
--- a/agent/base
+++ b/agent/base
@@ -104,7 +104,7 @@ if [[ -z "${hostname}" ]]; then
     export hostname=$(hostname -s)
 fi
 if [[ -z "${full_hostname}" ]]; then
-    export full_hostname=$(hostname)
+    export full_hostname=$(hostname -f)
 fi
 
 if [[ -z "$ssh_opts" ]]; then
--- a/agent/base
+++ b/agent/base
@@ -59,16 +59,10 @@ export pbench_bspp_dir pbench_lib_dir
 
 TS_FORMAT="%FT%H:%M:%S"
 
-if [[ -z "${_PBENCH_UNIT_TESTS}" ]]; then
-    function timestamp {
-        # use ns in the timestamp
-        echo "$(date --utc +"${TS_FORMAT}.%N")"
-    }
-else
-    function timestamp {
-        echo "1900-01-01T00:00:00.000000"
-    }
-fi
+function timestamp {
+    # use ns in the timestamp
+    echo "$(date --utc +"${TS_FORMAT}.%N")"
+}
 
 function log {
     echo "[info][$(timestamp)] $*" >> $pbench_log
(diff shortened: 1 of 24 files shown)


agent/base.ut.post
```diff
@@ -0,0 +1,21 @@
...
+function timestamp {
+    echo "1900-01-01T00:00:00.000000"
+}
+
+export date="1900-01-01T00:00:00"
+export date_suffix="1900.01.01T00.00.00"
+
+function check_enable_copr {
+    echo $1
+    #$2 is the version which will vary
```
dbutenhof, line 16: I don't get the relevance of this comment, although I get that it was in the original. Should we just dump this... because it's really more confusing than helpful.

agent/bench-scripts/pbench-iozone
```diff
@@ -218,7 +218,7 @@ pbench-start-tools --group=$tool_group --dir=$benchmark_results_dir
 for file_size in $(echo $file_sizes | sed -e s/,/" "/g); do
 	for record_size in $(echo $record_sizes | sed -e s/,/" "/g); do
 		for run in $(seq 1 ${number_of_runs}); do
-			$benchmark_bin -i0 -i1 -i2 -i3 -i4 -i5 -i6 -i7 -i8 -i9 -i10 -i11 -i12  -c -e -f $targets -s ${file_size}g -r ${record_size}m >> $benchmark_results_dir/$(hostname)_$run.iozone
+			$benchmark_bin -i0 -i1 -i2 -i3 -i4 -i5 -i6 -i7 -i8 -i9 -i10 -i11 -i12  -c -e -f $targets -s ${file_size}g -r ${record_size}m >> $benchmark_results_dir/$(hostname -f)_$run.iozone
```
dbutenhof, line 221: Shouldn't we have `base` in scope here? So why isn't this using the environment variable? (Other than that it wasn't before?)

dbutenhof (dismissed):
Seems OK, although a few minor questions and comments...

agent/base
```diff
@@ -156,33 +137,25 @@ function get_redhat_version() {
...
+function check_enable_copr {
+    local copr_user=$1
+    local copr_name=$2
 
-        if is_fedora ; then
-            dnf copr enable ${copr_user}/${copr_name}
+    if is_fedora ; then
+        dnf copr enable ${copr_user}/${copr_name}
+    else
+        if is_redhat ; then
```
ndokos, line 147: Does this cover CentOS as well?

agent/base
```diff
@@ -278,12 +251,12 @@ function generate_inventory {
     printf -- ${preamble} "${hostname}"
     for dirent in $(/bin/ls -1 ${tool_group_dir}); do
         if [[ "${dirent}" == "__trigger__" ]]; then
-		# Ignore trigger files
-		continue
-	elif [[ ! -d ${tool_group_dir}/${dirent} ]]; then
-		# Ignore spurious files
-		continue
-	fi
+            # Ignore trigger files
```
ndokos, line 254: Does the FIXME on l.242 need to be addressed?

ndokos (changes requested):

agent/base
portante: Good question, filed #1950 to track separately.

agent/base
portante: Thanks, removed the comment at line 242.

ndokos (dismissed):
The PR description needs to be updated I think to delete the stuff about flake8/black which has been moved out of this PR. Otherwise, LGTM.

agent/base.ut.post
```diff
@@ -0,0 +1,24 @@
...
+
+function timestamp {
+    echo "1900-01-01T00:00:00.000000"
+}
+
+export date="1900-01-01T00:00:00"
+export date_suffix="1900.01.01T00.00.00"
+
+function check_enable_copr {
+    # For the unit test environment, we only echo the first arguemnt passed to
```
dbutenhof, line 15: I hate to start an "arguemnt" ... but the word here is misspelled.

The revised comment, however, has the marked advantage of being actually useful! Thanks.

dbutenhof (approved):
Minor typo, which can be fixed later if you're not making another change anyway.

dbutenhof (approved):

ndokos (approved):

PR #1930, opened 2020-10-29
Test use of new Fedora 32 image for linting

portante:

PR #1936, opened 2020-10-31
Jenkins integration using Fedora 32 container

portante:
The approach we are taking is fairly straightforward:

 1. Build a container based on Fedora 32

    - Now that Fedora 33 is released, 32 will be the lowest version we support
    - We add all the RPM dependencies we need to execute from the `pbench-agent.spec` file, and those we have discovered along the way
    - We build the image and publish it to Quay.io as:
        `https://quay.io/pbench/pbench-devel:$(cat jenkins/branch.name)`
       Where `jenkins/branch.name` stores the branch name of that branch.
    - The intention is that each branch would potentially have its own `utils/development.Dockerfile` to meet the needs of that branch

 2. Create a new `jenkins` directory

    The Jenkins pipeline description is kept in `jenkins/Pipeline.gy`, the file extenion is `.gy` for "Groovy", the language used for the pipeline.

    We create a `jenkins/run` command that encapsulates how to run unit tests via `podman`. This let's us hide the particular use of podman syntax from the Jenkins pipeline, allowing us to easily execute from the command line when testing the pipeline outside of Jenkins.

    We also create a `jenkins/tox` command to give us a chance to perform the necessary setup to run `tox` inside the container.

 3. The `jenkins/Pipeline.gy` describes two stages, the first being linting and pytest-based tests, via `tox -e jenkins-pytestse`, the second being the run of all the legacy unit tests in parallel via `tox -e jenkins-unittests`.

    For both stages, we use the new switch, `--current-env` to force `tox` to use the containers environment instead of trying to `pip3 install` everything.  This is a big scaling step to avoid lots of overhead trying to install everything into the container each time tests are run.  Instead, we just build the container with everything we need once, leading to a much faster stage run times.

    Further we have created the new `run-unittests` and `run-pytests` just for these stages, where `run-unittests` runs the tests in parallel with-in the tox environment.

 4. Other changes

    - We fix the `agent/tool-scripts/unittests` to deal with the fact that `find -L ${dir} -perm 755` does not work in a container using overlay-fs: the find command should be considering the permissions on the file ultimately found following symlinks, but seems to be applying the permissions check to the symlink itself
    - We mock out the `cat` command for the server unit tests so that we can force an error reading a prefix file.  Apparently, under container environments, just `chmod`ing a file so that it is unreadable is not sufficient to cause an error.  So we use the mock'd version to accomplish the same effect.
    - We have added the `jinja2` module to the list of required Python 3 modules for the agent since it depends on that directly; and we add it to the `pbench-agent.spec.j2` file as well
    - We restore the dependency on the `procps-ng` since the various tool-scripts use `pidof`, which is our code dependency not a particular tool's dependency
    - We modify the default `tox` install command to explicitly use a known environment install location so that individual tests can take advantage of that (this is how `jenkins/run-pytests` gets the correct environment using `--current-env`).
    - We modify the `agent` & `server` `test-requirements.txt` files to add the `coverage` module which is required when using the `pytest-cov` module

 5. One can easily test how things will run in Jenkins via:

```
$ export WORKSPACE_TMP=/tmp/workspace
$ jenkins/run jenkins/tox --current-env -r -e jenkins-pytests
$ jenkins/run jenkins/tox --current-env -r -e jenkins-unittests
```

Concerns:

 - How do we maintain this list of RPMs for the Dockerfile separate from the `pbench-agent.spec` and `pbench-server.spec` files, and the container image Dockerfiles we maintain for both agent and server?


portante:
This final PR is a combination of the following PRs, #1929, #1930, #1931, and #1932, finalized in an official PR that can be reviewed, and includes a description of the changes made for the use of Jenkins.

jenkins/development.Dockerfile
```diff
@@ -0,0 +1,30 @@
+# Fedora 32 pbench developement image
```
dbutenhof, line 1: "development"

dbutenhof (approved):
Wow; lots of layers built up since the start.

jenkins/development.Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+# We maintain this image with all the necessary user-space setup required to
+# run the various unit tests and functional tests in a common development
+# environment.
+#
+# The image is published to https://quay.io/pbench/development and "tagged"
+# using the branch name in which it works: `master`, `b0.70`, `b0.71`, etc.
+#
+# NOTE WELL: This image has no entry point.
+#
+# It is intended to be built on top of with additional entry point for running
```
ndokos, line 12: This raises my passive voice hackles :-) Maybe something like "The intention is that an image with an entry point for running a test (or a suite of tests) is built on top of this" if you still want passive voice - or "You should build an image on top of this, with an entry point for running a test (or a suite of tests) in pbench".

ndokos (changes requested):

agent/tool-scripts/datalog/prometheus-metrics-datalog
```diff
@@ -249,11 +282,18 @@ while True:
...
+            " start: %d, duration: %d, prom2json: %d]; please consider"
+            " changing the interval to at least %d seconds",
+            PROG,
+            drifted,
+            interval,
+            time_stamp,
+            duration,
+            end_cmd_time_stamp - time_stamp,
+            int(math.ceil(avg_duration * 1.5)),
+        )
```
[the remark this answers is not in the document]
ndokos: NM - the commit message explains.

agent/tool-scripts/unittests
```diff
@@ -134,13 +155,12 @@ if [[ ${?} -ne 0 ]]; then
...
-        continue
-    fi
-    create_mock "${toolname}"
-done
+# Mock out GOLANG setups
+mkdir -p ${_testroot}/go/bin
+if [[ ${?} -ne 0 ]]; then
+    echo "ERROR: failed to create ${_testroot}/go/bin" >&2
+    exit 1
+fi
```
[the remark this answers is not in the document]
ndokos: NM - the PR description explains.

agent/tool-scripts/unittests
```diff
@@ -134,13 +155,12 @@ if [[ ${?} -ne 0 ]]; then
...
 fi
 
-for tool in $(find -L ${_ts} -maxdepth 1 -perm 775); do
-    toolname="$(basename ${tool})"
-    if [[ "${toolname}" == "base-tool" ]]; then
-        continue
-    fi
-    create_mock "${toolname}"
-done
+# Mock out GOLANG setups
```
npalaska, line 158: just a question, why do we need to do a GOLAND setup for tool-scripts unittest?

portante:
Pulled a number of changes out into PRs #1946, #1947, and #1948.  We'll want to get those PRs merged first before we reviewing this further.  As such I converted this to a `Draft` for now.

agent/tool-scripts/unittests
portante: > just a question, why do we need to do a GOLANG setup for tool-scripts unittest?

We have tools that use `GOLANG`,`pprof`, for example.

jenkins/development.Dockerfile
portante: Fixed.

jenkins/development.Dockerfile
portante: Thanks, fixed.

portante (commented):
Updated to remove the `prometheus-metrics-datalog` changes, pushing out into PR #1949.

lib/pbench/cli/agent/commands/results/__init__.py
```diff
@@ -58,42 +56,42 @@ def move_results(ctx, _user, _prefix, _show_server):
         logger.error("Failed to create temporary directory")
         sys.exit(1)
 
-    dirs = [
-        _dir
-        for _dir in next(os.walk(config.pbench_run))[1]
-        if not _dir.startswith("tools-") and not _dir.startswith("tmp")
-    ]
+    runs_copied = 0
+    failures = 0
```
dbutenhof, line 60: Probably pointless to repeat it here, but I don't see any use of `failures` ... so it seems that any copy failures go unreported? If that's deliberate, we should drop `failures` from the return tuple and just return the success count.

lib/pbench/test/unit/common/test_config.py
```diff
@@ -0,0 +1,68 @@
...
+        with pytest.raises(AttributeError):
+            print(f"{config.logger_host!r}")
+        with pytest.raises(AttributeError):
+            print(f"{config.logger_port!r}")
+        assert "42" == config.get(
+            "other", "foobar"
+        ), "Failed to fetch 'foobar' from 'DEFAULT' section"
+        assert "43" == config.get(
+            "other", "barfoo"
+        ), "Failed to fethc 'barfoo' from 'other' section"
```
dbutenhof, line 41: "Failed to fetch" ... ???

lib/pbench/test/unit/common/test_logging.py
```diff
@@ -0,0 +1,144 @@
...
+        assert _handlers[fname] == self.logger.logger.handlers[0]
+        assert (
+            _handlers[fname].__class__.__name__ == "SysLogHandler"
+        ), f"Unexpected handler set, {_handlers[fname].__class__.__name__!r}"
+        assert (
+            _handlers[fname].address[0] == "host.example.com"
+            and _handlers[fname].address[1] == 42
+        ), f"Unexpected handler address set, {_handlers[fname].address!r}"
+
+    @staticmethod
```
dbutenhof, line 125: Is there a non-obvious reason why this test case is static while the others are not?

lib/pbench/cli/agent/commands/results/__init__.py
portante: Answered in PR #1947.

lib/pbench/test/unit/common/test_logging.py
portante: I'll be changing this to a non-static method, but the reason it is different is that all the other tests rely on the `config` fixture at line 17 to setup `self.config`.  I'll fix the code to stop using `self.logger` since it is not necessary.  This method creates its own config.

server/bin/test-bin/cat
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+if [[ ! -z "${1}" && "$(basename ${1})" == "tarball-bad-prefix-file_1970.01.01T00.00.00.prefix" ]]; then
+    exit 1
+fi
```
dbutenhof, line 5: I must have missed something, but the context and significance of this isn't obvious. Maybe a comment?

dbutenhof (approved):

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -164,7 +164,7 @@ def wait_for_subs(chan, expected_tms, logger):
         except Exception:
             logger.warning("data payload in message not UTF-8, '%r'", json_str)
             continue
-        logger.debug('channel payload, "%r"', json_str)
+        logger.debug('w. channel payload, "%r"', json_str)
```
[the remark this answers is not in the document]
portante: Agreed, and the `repr` output already has quotes around it, so I'll remove the extra double quotes.

dbutenhof (dismissed):
Still not really thrilled by the "enumerated list" style debug logs, but it's not a big deal. Let's ship it.

ndokos (changes requested):

dbutenhof (dismissed):

jenkins/run
```diff
@@ -0,0 +1,63 @@
...
+    exit 1
+fi
+
+if [[ -z "${1}" ]]; then
+    printf -- "ERROR - missing required arguments for what to run\n" >&2
+    exit 1
+fi
+
+# FIXME: for now we hard-code the image tag to be jenkins
+_branch_name="jenkins"
```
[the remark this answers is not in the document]
portante: Fixed, I think, but it is ugly.

agent/tool-scripts/unittests
```diff
@@ -126,6 +130,23 @@ function create_mock {
...
+    if [[ "${toolname}" == "README" ]]; then
+        continue
+    fi
+    if [[ "${toolname}" == "perf.README" ]]; then
+        continue
+    fi
+    create_mock "${toolname}"
+done
+
 # Special mock just for systemstap
```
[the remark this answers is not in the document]
portante: Fixed.

ndokos (dismissed):

portante:
I've squashed the commits to 2, in advanced of merging.

jenkins/development.Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+        python3-GitPython \
+        python3-boto3 \
+        python3-botocore \
+        python3-bottle \
+        python3-cffi \
+        python3-click \
+        python3-colorlog \
+        python3-coverage \
+        python3-daemon \
+        python3-elasticsearch \
```
dbutenhof, line 67: Is this really `elasticsearch` vs `elasticsearch1`? Nice, going forward, but means we're currently pip isntalling `elasticsearch1`??

dbutenhof (approved):

jenkins/development.Dockerfile
portante: Yes, I can't find an RPM for the `elasticsearch1` in Fedora.

So we pip install `python-pidfile`, `pyesbulk`, and `elasticsearch1`.

agent/util-scripts/unittests
```diff
@@ -111,14 +113,22 @@ function _run {
...
 
     popd > /dev/null 2>&1
     if [[ ${?} -ne 0 ]]; then
         exit 1
     fi
 
+    if [[ "${sortit}" == "sort" ]]; then
+        sort ${_runout} >> ${_testout}
+    else
+        cat ${_runout} >> ${_testout}
```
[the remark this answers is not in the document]
portante: So `${_testout}` is appended to at line 114 above, and I think it will have other output in it as well.  So if we do a `mv` at line 128 we'd lose that output, if I am reading the code correctly.

ndokos:
I approved this already, but just wanted to mention that the experiments that @portante suggested, that is:

```
WORKSPACE_TMP=/tmp/workspace jenkins/run jenkins/tox --current-env -r -e jenkins-pytests
WORKSPACE_TMP=/tmp/workspace jenkins/run jenkins/tox --current-env -r -e jenkins-unittests
```
worked,  with the caveat that you cannot run them in a git worktree: it's got to be a vanilla branch.

ndokos (approved):

PR #1939, opened 2020-11-02
Setting SELinux Labels correctly in prep-shim

### LINKED ISSUE #1928: SELinux labels are not set correctly in the new prep-shim script
When the prep-shim script copies tar balls and MD5s into the archive directory (and maybe also when it creates a new controller directory), it need to set the selinux label correctly by doing the equivalent of `restorecon -R /path/to/dir`.

Apparently the label was inherited properly from the `fs-version-001` directory in the old shell script, but we seem to have lost that with the new python script.

That needs to be fixed to allow access to tar balls by users, using the `pbench-tarballs` URL.

riya-17:
Fixes #1928 

We not only need to preserve the labels but also all the metadata of the tarball.

- [x] restorecon - help to preserve all metadata

### DIFF
--- a/server/bin/gold/test-5.2.txt
+++ b/server/bin/gold/test-5.2.txt
@@ -1093,6 +1093,14 @@ run-1970-01-01T00:00:42-UTC: pbench-satellite-cleanup ends: Total 3 tarballs cle
 ---- pbench-satellite-local/logs
 --- pbench log file contents
 +++ test-execution.log file contents
+selinux.restorecon('/var/tmp/pbench-test-server/test-5.2/pbench/archive/fs-version-001/controller-b-with-prefixes/tarball-0_1970.01.01T00.42.00.tar.xz', recursive=False, verbose=False, force=False)
+selinux.restorecon('/var/tmp/pbench-test-server/test-5.2/pbench/archive/fs-version-001/controller-b-with-prefixes/tarball-0_1970.01.01T00.42.00.tar.xz.md5', recursive=False, verbose=False, force=False)
+selinux.restorecon('/var/tmp/pbench-test-server/test-5.2/pbench/archive/fs-version-001/controller-b-with-prefixes/tarball-w-dot-prefix_1970.01.01T00.42.00.tar.xz', recursive=False, verbose=False, force=False)
+selinux.restorecon('/var/tmp/pbench-test-server/test-5.2/pbench/archive/fs-version-001/controller-b-with-prefixes/tarball-w-dot-prefix_1970.01.01T00.42.00.tar.xz.md5', recursive=False, verbose=False, force=False)
+selinux.restorecon('/var/tmp/pbench-test-server/test-5.2/pbench/archive/fs-version-001/controller-b-with-prefixes/tarball-w-prefix-dot_1970.01.01T00.42.00.tar.xz', recursive=False, verbose=False, force=False)
+selinux.restorecon('/var/tmp/pbench-test-server/test-5.2/pbench/archive/fs-version-001/controller-b-with-prefixes/tarball-w-prefix-dot_1970.01.01T00.42.00.tar.xz.md5', recursive=False, verbose=False, force=False)
+selinux.restorecon('/var/tmp/pbench-test-server/test-5.2/pbench/archive/fs-version-001/controller-g-normal/tarball-normal_1970.01.01T00.42.00.tar.xz', recursive=False, verbose=False, force=False)
+selinux.restorecon('/var/tmp/pbench-test-server/test-5.2/pbench/archive/fs-version-001/controller-g-normal/tarball-normal_1970.01.01T00.42.00.tar.xz.md5', recursive=False, verbose=False, force=False)
 ssh pbench-satellite.example.com /var/tmp/pbench-test-server/test-5.2/opt/pbench-server-satellite/bin/pbench-sync-package-tarballs
 ssh pbench-satellite.example.com /var/tmp/pbench-test-server/test-5.2/opt/pbench-server-satellite/bin/pbench-satellite-state-change /var/tmp/pbench-test-server/test-5.2/pbench-satellite/archive/fs-version-001
 --- test-execution.log file contents
--- a/server/bin/pbench-server-prep-shim-002.py
+++ b/server/bin/pbench-server-prep-shim-002.py
@@ -12,6 +12,7 @@ import os
 import sys
 import glob
 import shutil
+import selinux
 import tempfile
 
 from pathlib import Path
@@ -201,6 +202,14 @@ def process_tb(config, logger, receive_dir, qdir_md5, duplicates, errors):
             quarantine((errors / controller), logger, tb, tbmd5)
             nerrs += 1
             continue
+
+        # Restore the SELinux context properly
+        try:
+            selinux.restorecon(dest / tb.name)
+            selinux.restorecon(dest / tbmd5.name)
+        except Exception as e:
+            # log it but do not abort
+            logger.error("{}: Error: 'restorecon {}', {}", config.TS, dest / tb.name, e)
 
         # Now that we have successfully moved the tar ball and its .md5 to the
         # destination, we can remove the original .md5 file.
--- a/server/bin/unittests
+++ b/server/bin/unittests
@@ -108,7 +108,10 @@ function _run_allscripts {
     _run pbench-backup-tarballs
     _run pbench-verify-backup-tarballs
     # Invoke pbench-satellite-cleanup in the context of the satellite
-    PATH=${_testopt_sat}/unittest-scripts:${_testopt_sat}/bin:${_orig_PATH} _PBENCH_SERVER_CONFIG=${SATCONFIG} _run pbench-satellite-cleanup
+    PATH=${_testopt_sat}/unittest-scripts:${_testopt_sat}/bin:${_orig_PATH} \
+        PYTHONPATH=${_testopt_sat}/unittest-lib:${_orig_PYTHONPATH} \
+        _PBENCH_SERVER_CONFIG=${SATCONFIG} \
+        _run pbench-satellite-cleanup
 }
 
 function _local_find {
@@ -248,7 +251,15 @@ function _setup_state {
     let res=res+$?
     cp -a $_tdir/test-bin/* $_testopt/unittest-scripts/
     let res=res+$?
-    cp $_tdir/test-bin/* $_testopt_sat/unittest-scripts/
+    cp -a $_tdir/test-bin/* $_testopt_sat/unittest-scripts/
+    let res=res+$?
+    mkdir -p ${_testopt}/unittest-lib/
+    let res=res+$?
+    mkdir -p ${_testopt_sat}/unittest-lib/
+    let res=res+$?
+    cp -a ${_tdir}/test-lib/* ${_testopt}/unittest-lib/
+    let res=res+$?
+    cp -a ${_tdir}/test-lib/* ${_testopt_sat}/unittest-lib/
     let res=res+$?
     mkdir -p $_testopt/bin
     let res=res+$?
@@ -314,7 +325,9 @@ function _setup_state {
     # All the "real" scripts are found at $_testopt/bin, the mock scripts
     # are found in $_testopt/unittest-scripts.
     _orig_PATH=$PATH
+    _orig_PYTHONPATH=$PYTHONPATH
     export PATH=$_testopt/unittest-scripts:$_testopt/bin:$PATH
+    export PYTHONPATH=$_testopt/unittest-lib:$PYTHONPATH
 
     # Expected location of the final configuration files
     export _PBENCH_SERVER_CONFIG=$_testopt/lib/config/pbench-server.cfg
@@ -489,6 +502,7 @@ function _reset_state {
     fi
 
     export PATH=${_orig_PATH}
+    export PYTHONPATH=${_orig_PYTHONPATH}
     unset _PBENCH_SERVER_CONFIG
     unset SATCONFIG
 
--- a/server/rpm/pbench-server.spec.j2
+++ b/server/rpm/pbench-server.spec.j2
@@ -15,8 +15,10 @@ Requires: python3 python3-devel gcc
 Requires:       policycoreutils
 %if 0%{?rhel} != 7
 Requires: policycoreutils-python-utils
+Requires: libselinux-python3
 %else
 Requires: policycoreutils-python
+Requires: python3-libselinux
 %endif
 
 Requires: npm
--- a/jenkins/development.Dockerfile
+++ b/jenkins/development.Dockerfile
@@ -74,6 +74,7 @@ RUN \
         python3-humanize \
         python3-itsdangerous \
         python3-jinja2 \
+        python3-libselinux \
         python3-mock \
         python3-pip \
         python3-psutil \
--- a/server/bin/test-lib/selinux.py
+++ b/server/bin/test-lib/selinux.py
@@ -1,0 +1,23 @@
+# This is a mock file for selinux.restorecon test cases
+import os
+
+
+def restorecon(path, recursive=False, verbose=False, force=False):
+    """ Restore SELinux context on a given path
+
+    Arguments:
+    path -- The pathname for the file or directory to be relabeled.
+
+    Keyword arguments:
+    recursive -- Change files and directories file labels recursively (default False)
+    verbose -- Show changes in file labels (default False)
+    force -- Force reset of context to match file_context for customizable files,
+    and the default file context, changing the user, role, range portion  as well
+    as the type (default False)
+    """
+    with open(os.environ["_testlog"], "a") as ofp:
+        ofp.write(
+            f"selinux.restorecon({str(path)!r}, recursive={recursive!r}, verbose={verbose!r}, force={force!r})\n"
+        )
+
+    return 0


server/bin/pbench-server-prep-shim-002.py
```diff
@@ -166,7 +166,7 @@ def process_tb(config, logger, receive_dir, qdir_md5, duplicates, errors):
         # First, copy the small .md5 file to the destination. That way, if
         # that operation fails it will fail quickly since the file is small.
         try:
-            shutil.copy2(tbmd5, dest)
+            shutil.copy(tbmd5, dest)
```
ndokos, line 169: Does the `shutil.move()` call on l.190 do the right thing for the tarball? IIUC, it does not and we need to do something there. I wonder if the thing to do is leave the `shutil.copy2()` alone (again IIUC, that's the equivalent of `cp -a` in the shell) and do explicit `restorecon` calls (using the `selinux` module from the `libselinux-python3` RPM package) to set the SELinux context on the tar ball and md5 files.

ndokos (changes requested):

server/bin/pbench-server-prep-shim-002.py
riya-17: seems fair

riya-17: How should I import SELinux in code? when I try the python3 command line, it is working after importing. In code, it is saying no module named "selinux".

riya-17: Do we need to define in PATH?

ndokos: If you have the `libeslinux-python3` RPM installed, you should have an `selinux` python3 module in `/usr/lib64/python3.6/site-packages/selinux` (on RHEL7). Then you should be able to
```
import selinux
...
selinux.restorecon(<path>)
...
```
Does that not work?

ndokos (commented):
I think that *should* do it. Do we have unit tests for checking? Do the current unit tests pass locally?

You will need to rebase too.

riya-17:
> I think that _should_ do it. Do we have unit tests for checking? Do the current unit tests pass locally?
> 
> You will need to rebase too.

No, I tested locally but we do not have unittests.  Yaa current unit tests are passing locally.
Yeah I will rebase it

server/bin/pbench-server-prep-shim-002.py
riya-17: I am using RHEL8 and It is failing test cases.

riya-17:
Facing this error in RHEL8
`! Traceback (most recent call last):
!   File "/var/tmp/pbench-test-server/test-0.1/opt/pbench-server/bin/pbench-server-prep-shim-002.py", line 15, in <module>
!     import selinux
! ModuleNotFoundError: No module named 'selinux'`

server/bin/pbench-server-prep-shim-002.py
```diff
@@ -167,6 +168,7 @@ def process_tb(config, logger, receive_dir, qdir_md5, duplicates, errors):
         # that operation fails it will fail quickly since the file is small.
         try:
             shutil.copy2(tbmd5, dest)
+            selinux.restorecon(dest / tbmd5.name)
```
dbutenhof, line 171: According to my reading, this basically sets default context for the target file. Is it ever possible that the original file in the reception area might have any non-default context? If so, we're throwing it away here.

riya-17: No idea about that. @ndokos

portante: If you install this on a VM can we try out a few scenarios to see if this works properly?  Can you devise a test to expose this behavior?

portante (changes requested):

dbutenhof:
It's unfortunate that you apparently couldn't get the selinux Python module to work for you. That's a mystery... and it seems a shame to revert to just spawning the command because that's a lot less efficient. (Presuming of course that the Python module wasn't already doing that!)

In any case, we definitely need to try to confirm proper behavior with a unit test.

ndokos (changes requested):
Why is the module insufficient? There is no version of it for the SCL environment AFAICT, but there are packages for the native python3 environment:

- F31 (and I presume F32 and F33): `python3-libselinux.x86_64                                                                         2.9-6.fc31                                                                          @updates`
- RHEL7: `libselinux-python3.x86_64             2.5-15.el7             @rhel-7-server-rpms`
- RHEL8: `python3-libselinux.x86_64       2.9-3.el8          @anaconda` (plus there is an update available: `python3-libselinux.x86_64       2.9-4.el8_3        rhel-8-for-x86_64-baseos-rpms`

So apart from the screwy naming on RHEL7, there should be no problem in using the `selinux` python module in all platforms. If there are problems, let's discuss in the scrum.

dbutenhof (changes requested):
Actually, the containerized Jenkins tests are failing because `selinux` isn't installed. I don't think it was already in the `requirements.txt` file, so you probably need to add it.

portante (changes requested):
Seems fine, but needs a rebase.

@riya-17, how does the test output compare with `test-5.2.txt` on the `b0.69` branch?

server/requirements.txt
```diff
@@ -7,3 +7,4 @@ gunicorn
 humanize
 pyesbulk>=2.0.1
 requests
+libselinux
```
dbutenhof, line 10: I think this should be `selinux`; at least, `pip` sees that module, and it can be imported, whereas attempting to `pip install libselinux` locally gives me the same problem as the Jenkins tests.

dbutenhof (changes requested):
```
[31mERROR: Could not find a version that satisfies the requirement libselinux (from -r /home/pbench/server/requirements.txt (line 10)) (from versions: none)[0m
[31mERROR: No matching distribution found for libselinux (from -r /home/pbench/server/requirements.txt (line 10))[0m
```

portante:
Let's get this rebased first and address the software dependency requirements before we pull it back out of `Draft` mode.

riya-17:
> Why is the module insufficient? There is no version of it for the SCL environment AFAICT, but there are packages for the native python3 environment:
> 
>     * F31 (and I presume F32 and F33): `python3-libselinux.x86_64                                                                         2.9-6.fc31                                                                          @updates`
> 
>     * RHEL7: `libselinux-python3.x86_64             2.5-15.el7             @rhel-7-server-rpms`
> 
>     * RHEL8: `python3-libselinux.x86_64       2.9-3.el8          @anaconda` (plus there is an update available: `python3-libselinux.x86_64       2.9-4.el8_3        rhel-8-for-x86_64-baseos-rpms`
> 
> 
> So apart from the screwy naming on RHEL7, there should be no problem in using the `selinux` python module in all platforms. If there are problems, let's discuss in the scrum.

These are the software dependency requirements as mentioned by @ndokos and these are only available with dnf install

ndokos (changes requested):
IIUC, we need to test this on a RHEL7 VM *and* a RHEL8 VM to make sure that the tests pass with the given dependencies. And we also need to add some tests for the specific behaviors we are looking for (in particular, that a tar ball that is copied into the archive has the `httpd_<mumble>` label. Is there anything else to do here?

server/test-requirements.txt
```diff
@@ -6,3 +6,4 @@ pytest>=4.6.3, < 5
 pytest-helpers-namespace>=2019.1.8
 pytest-mock>=1.10.4
 requests_mock
+selinux
```
portante, line 9: You don't need this in `test-requirements.txt` if you already have it in `requirements.txt`.

server/bin/gold/test-5.2.txt
```diff
@@ -1,4 +1,12 @@
...
+Warning no default label for /var/tmp/pbench-test-server/test-5.2/pbench/archive/fs-version-001/controller-b-with-prefixes/tarball-0_1970.01.01T00.42.00.tar.xz.md5
+Warning no default label for /var/tmp/pbench-test-server/test-5.2/pbench/archive/fs-version-001/controller-b-with-prefixes/tarball-w-dot-prefix_1970.01.01T00.42.00.tar.xz
+Warning no default label for /var/tmp/pbench-test-server/test-5.2/pbench/archive/fs-version-001/controller-b-with-prefixes/tarball-w-dot-prefix_1970.01.01T00.42.00.tar.xz.md5
+Warning no default label for /var/tmp/pbench-test-server/test-5.2/pbench/archive/fs-version-001/controller-b-with-prefixes/tarball-w-prefix-dot_1970.01.01T00.42.00.tar.xz
+Warning no default label for /var/tmp/pbench-test-server/test-5.2/pbench/archive/fs-version-001/controller-b-with-prefixes/tarball-w-prefix-dot_1970.01.01T00.42.00.tar.xz.md5
+Warning no default label for /var/tmp/pbench-test-server/test-5.2/pbench/archive/fs-version-001/controller-g-normal/tarball-normal_1970.01.01T00.42.00.tar.xz
+Warning no default label for /var/tmp/pbench-test-server/test-5.2/pbench/archive/fs-version-001/controller-g-normal/tarball-normal_1970.01.01T00.42.00.tar.xz.md5
```
portante, line 9: Who generates this output?  If this is coming from the selinux module itself, we'll have to get a hold of how it is produced.

portante (changes requested):

server/test-requirements.txt
riya-17: oh ok... I will change that

server/bin/gold/test-5.2.txt
riya-17: yeah It is coming from the SELinux module I will see how it is getting produced.

server/requirements.txt
riya-17: yes, I changed that

server/bin/gold/test-5.2.txt
riya-17: "libselinux: selinux_restorecon: only log no default label warning for caller-supplied pathname", I will look how we can disable this warning

portante (changes requested):
Okay this looks good, but the unit testing needs to be addressed still.

In particular, we need to "mock out" the actual SELinux operations, since for unit tests we only need to verify that we make the proper API call with the correct arguments.

I'd suggest that we `import` a mocked out module of the SELinux module in `pbench-server-prep-shim-002.py` when running under the unit tests, where that mock to records the argument pass to `selinux.restorecon()`

However, we also need to check that we have a proper label for the argument.  So perhaps we can wrap the `selinux.restorecon()` call to capture `stdout` and `stderr` and report our own failure if something was captured?

riya-17:
> Okay this looks good, but the unit testing needs to be addressed still.
> 
> In particular, we need to "mock out" the actual SELinux operations, since for unit tests we only need to verify that we make the proper API call with the correct arguments.
> 
> I'd suggest that we `import` a mocked out module of the SELinux module in `pbench-server-prep-shim-002.py` when running under the unit tests, where that mock to records the argument pass to `selinux.restorecon()`
> 
> However, we also need to check that we have a proper label for the argument. So perhaps we can wrap the `selinux.restorecon()` call to capture `stdout` and `stderr` and report our own failure if something was captured?

yeah, unit testing is giving us a hard time so working on that. I will create a unit test for verifying the behavior.

lib/pbench/server/selinux.py
```diff
@@ -0,0 +1,24 @@
...
+    force -- Force reset of context to match file_context for customizable files,
+    and the default file context, changing the user, role, range portion  as well
+    as the type (default False)
+    """
+    restorecon_flags = False
+    status = selinux_restorecon(os.path.expanduser(path), restorecon_flags)
+    return status
+
+
+def selinux_restorecon(pathname, restorecon_flags):
```
portante, line 23: When not running the unit tests, how does this work?

riya-17: yeah, that's what I am working on.. just pushed the work.

riya-17: just wanted to see whether it is working on Jenkins majorly

ndokos (dismissed):

ndokos:
Needs rebasing.

server/bin/test-bin/selinux.py
```diff
@@ -0,0 +1,17 @@
...
+    Arguments:
+    path -- The pathname for the file or directory to be relabeled.
+
+    Keyword arguments:
+    recursive -- Change files and directories file labels recursively (default False)
+    verbose -- Show changes in file labels (default False)
+    force -- Force reset of context to match file_context for customizable files,
+    and the default file context, changing the user, role, range portion  as well
+    as the type (default False)
+    """
```
portante, line 16: Much like we do in https://github.com/distributed-system-analysis/pbench/blob/main/agent/util-scripts/test-bin/mpstat, can you write the argument names and values to the test-execution file?

Something like:

```
import sys
import os

with open(os.environ["_testlog"], "a") as ofp:
    ofp.write(f"{path!r}, recursive={recursive!r}, verbose={verbose!r}, force={force!r}\n")

sys.exit(0)
```

server/bin/test-bin/selinux.py
```diff
@@ -0,0 +1,17 @@
+# This is a mock file for selinux.restorecon test cases
```
portante, line 1: Great idea for this mock module.  Can we create a new directory, `test-lib`, as a sister directory to `test-bin`, moving this module there, and adding it to the `PYTHONPATH` from there in `unittests`?  I don't think it needs to be installed in the test environment, but it probably should have an empty `__init__.py` to be a proper python3 module.

portante (commented):
Looks good, just need execution output added, and the module moved.

server/bin/test-bin/selinux.py
riya-17: ok

server/bin/test-bin/selinux.py
riya-17: Ok sure will do that

server/bin/unittests
```diff
@@ -250,6 +250,10 @@ function _setup_state {
     let res=res+$?
     cp $_tdir/test-bin/* $_testopt_sat/unittest-scripts/
     let res=res+$?
+    cp -a $_tdir/test-lib/* $_testopt/unittest-scripts/
```
portante, line 253: No sure this should be in `unittest-scripts`; can we consider something like:

```
mkdir ${_testopt}/unittest-lib
cp -a ${_tdir}/test-lib/* ${_testopt}/unittest-lib/
...
```

portante (changes requested):

server/bin/unittests
riya-17: yeah sure

ndokos (dismissed):

server/bin/pbench-server-prep-shim-002.py
portante: I believe this comment is now resolved.

server/test-requirements.txt
portante: I see that we have resolved this by using an RPM in the development container.

We also need to then add this in the pbench-server RPM `.spec` file.

But, we should also have a python3 library module requirement in the `server/requirements.txt` file.  Can that be restored?

server/bin/gold/test-5.2.txt
portante: I believe this has been resolved.

lib/pbench/server/selinux.py
portante: I believe this has been resolved.

portante (changes requested):
This looks really good.  Just one more topic to address:

We need to add the required RPM to the pbench-server RPM .spec file.

And, we should also have a python3 library module requirement in the server/requirements.txt file.

riya-17:
> This looks really good. Just one more topic to address:
> 
> We need to add the required RPM to the pbench-server RPM .spec file.
> 
> And, we should also have a python3 library module requirement in the server/requirements.txt file.

Hey, are you talking about the spec file at location pbench/server/rpm/pbench-server.spec.j2?

adding the python3 library module in requirements.txt file might create problem coz it will then consider the pip installed module rather than rpm module which our code should not prefer as there are no restorecons available in that

portante:
> Hey, are you talking about the spec file at location pbench/server/rpm/pbench-server.spec.j2?

Yes.

> Adding the python3 library module in the `requirements.txt` file might create a problem coz it will then consider the pip installed module rather than the rpm module which our code should not prefer as there are no properly build SELinux modules available with that especially for fedora 32.

Ah, right, this is the problem that the RPM provides the Python3 module, but it is not available as a pip3 installable module as well.

I'd add a comment to the `server/requirements.txt` file to state that we don't have a pip3 module for selinux, but that it is provided by the selinux RPM for the various distributions we support.

server/rpm/pbench-server.spec.j2
```diff
@@ -65,6 +65,9 @@ cd /%{installdir}
 rm -rf node_modules
 npm install
 
+# install selinux modules under /%{installdir}
+dnf install python3-libselinux
```
portante, line 69: Actually, since we are in an RPM `.spec` file, we can add a `Requires: python3-libselinux` line to replace line 21.

portante (changes requested):

server/rpm/pbench-server.spec.j2
riya-17: ohh ok.. sorry I was not sure how I can add

portante (approved):

ndokos (dismissed):

server/bin/unittests
```diff
@@ -315,6 +323,7 @@ function _setup_state {
     # are found in $_testopt/unittest-scripts.
     _orig_PATH=$PATH
     export PATH=$_testopt/unittest-scripts:$_testopt/bin:$PATH
+    export PYTHONPATH=$_testopt/unittest-lib:$PYTHONPATH
```
dbutenhof, line 326: You copy into separate directories for "server" and "satellite", but there's a single `PYTHONPATH` modification. Is that really what we want? Is there any need for separate Python overlay libraries? Is doesn't appear there is right now, so maybe reserve that for later and just make the single copy you're using?

dbutenhof (approved):

npalaska (dismissed):

server/bin/unittests
portante: At the risk of belaboring this patch, even though it does still need a rebase, this seems like a good catch that we should address.

We handle the `PATH` for the satellite script invocation context separately at: https://github.com/distributed-system-analysis/pbench/blob/3219eea0d1fce179cd1d71ef3b6a1227ec57c842/server/bin/unittests#L111

So we should similarly setup `PYTHONPATH` to add the satellite test libs there as well.

portante (changes requested):
... and a rebase ...

riya-17:
> ... and a rebase ...

Done and fixed

dbutenhof (approved):

server/bin/unittests
```diff
@@ -109,6 +109,7 @@ function _run_allscripts {
     _run pbench-verify-backup-tarballs
     # Invoke pbench-satellite-cleanup in the context of the satellite
     PATH=${_testopt_sat}/unittest-scripts:${_testopt_sat}/bin:${_orig_PATH} _PBENCH_SERVER_CONFIG=${SATCONFIG} _run pbench-satellite-cleanup
+    PYTHONPATH=${_testopt_sat}/unittest-lib:${_orig_PYTHONPATH}
```
portante, line 112: This is subtle:  the `PATH` definition on line 111 above is actually using `bash`'s "temporary environment variable export" feature, where the updated `PATH` assignment is only enabled for the duration of the `_run` invocation on line 111.

In order for this to work, we want to move this `PYTHONPATH` setting following that `PATH`.

Try this out in `bash` to see how the two variables used by `ls` are only set for the duration of the command:

```
$ unset LS_COLORS
$ unset TIME_STYLE
$ ls -l
$ TIME_STYLE=full-iso LS_COLORS=always ls -l
$ echo "TIME_STYLE='${TIME_STYLE}'; LS_COLORS='${LS_COLORS}'"
$ dircolors    # restores your terminal ls colors
```

portante (changes requested):

server/bin/unittests
dbutenhof: Good catch; the `PATH` assignment was long enough that I missed that there was a command at the end of the line.

So this `PYTHONPATH` needs to be moved up to 111 along with `PATH` to have effect on `_run`.

ndokos: Maybe break it into three (or more) lines with `\` to escape the unwanted newlines?

```
PATH=...... \
PYTHONPATH=.... \
_PBENCH_SERVER_CONFIG=.... \
command
```

ndokos (commented):
Just a comment on @portante's change request.

server/bin/unittests
riya-17: I didn't know that. Thank you.  Fixed!

portante (approved):

ndokos (approved):

dbutenhof (approved):

portante (approved):

PR #1940, opened 2020-11-02
Adding signal Handlers in pbench-index.py

### LINKED ISSUE #1681: Consider the addition of signal handling to the pbench indexer
It would be great if we handled signals a bit more gracefully when indexing pbench data.

One possible set of behaviors to consider:

- `SIGHUP`: re-evaluate the list of tar balls to index *after* indexing of the current tar ball has finished
- `SIGINT`: stop indexing the current tar ball, log the interruption, leave the tar ball in `TO-INDEX` or `TO-INDEX-TOOL`, and move on to the next tar ball
- `SIGTERM`: stop indexing the current tar ball, log the interruption, generate a final report of all actions taken, exit

riya-17:
Fixes #1681 

- [x]  SIGINT - Interrupt the current tar ball being indexed, and proceed to the next one, if any
- [x]  SIGTERM - Stop processing tar balls immediately and exit gracefully
- [x]  SIGQUIT - Gracefully stop processing tar ball

### DIFF
--- a/server/bin/pbench-index.py
+++ b/server/bin/pbench-index.py
@@ -9,6 +9,7 @@ into the configured Elasticsearch V1 instance.
 import sys
 import os
 import glob
+import signal
 import tarfile
 import tempfile
 from pathlib import Path
@@ -34,6 +35,22 @@ from pbench.server.indexer import (
 )
 from pbench.server.report import Report
 from pbench.server.utils import rename_tb_link, quarantine
+
+
+class SigIntException(Exception):
+    pass
+
+
+def sigint_handler(*args):
+    raise SigIntException()
+
+
+class SigTermException(Exception):
+    pass
+
+
+def sigterm_handler(*args):
+    raise SigTermException()
 
 
 # Internal debugging flag.
@@ -102,6 +119,35 @@ def main(options, name):
          8 - Unable to load and process expected mapping files
          9 - Unable to update index templates in configured Elasticsearch
              instance
+
+        Signal Handlers used to establish different patterns for the three
+        behaviors:
+
+        1. Gracefully stop processing tar balls
+            - SIGQUIT
+            - The current tar ball is indexed until completion, but no other
+              tar balls are processed.
+            - Handler Behavior:
+                - Sets a flag that causes the code flow to break out of the
+                  for loop.
+                - Does not raise an exception.
+
+        2. Interrupt the current tar ball being indexed, and proceed to the
+           next one, if any
+            - SIGINT
+            - Handler Behavior:
+                - try/except/finally placed immediately around the es_index()
+                  call so that the signal handler will only be established for
+                  the duration of the call.
+                - Raises an exception caught by above try/except/finally.
+                - The finally clause would take down the signal handler.
+
+        3. Stop processing tar balls immediately and exit gracefully
+            - SIGTERM
+            - Handler Behavior:
+                - Raises an exception caught be a new, outer-most, try/except
+                  block that does not have a finally clause (as you don't want
+                  any code execution in the finally block).
     """
     _name_suf = "-tool-data" if options.index_tool_data else ""
     _name_re = "-re" if options.re_index else ""
@@ -216,6 +262,10 @@ def main(options, name):
                 continue
             else:
                 tarballs.append((size, controller, tb))
+    except SigTermException:
+        # Re-raise a SIGTERM to avoid it being lumped in with general
+        # exception handling below.
+        raise
     except Exception:
         idxctx.logger.exception(
             "Unexpected error encountered generating list" " of tar balls to process"
@@ -241,6 +291,10 @@ def main(options, name):
     except TemplateError as e:
         idxctx.logger.error("update_templates [end], error {}", repr(e))
         res = 9
+    except SigTermException:
+        # Re-raise a SIGTERM to avoid it being lumped in with general
+        # exception handling below.
+        raise
     except Exception:
         idxctx.logger.exception(
             "update_templates [end]: Unexpected template" " processing error"
@@ -269,6 +323,10 @@ def main(options, name):
     # documents.
     try:
         tracking_id = report.post_status(tstos(idxctx.time()), "start")
+    except SigTermException:
+        # Re-raise a SIGTERM to avoid it being lumped in with general
+        # exception handling below.
+        raise
     except Exception:
         idxctx.logger.error("Failed to post initial report status")
         return 12
@@ -291,6 +349,16 @@ def main(options, name):
             erred = Path(tmpdir, f"{name}.{idxctx.TS}.erred")
             skipped = Path(tmpdir, f"{name}.{idxctx.TS}.skipped")
             ie_filepath = Path(tmpdir, f"{name}.{idxctx.TS}.indexing-errors.json")
+
+            # We use a list object here so that when we close over this
+            # variable in the handler, the list object will be closed over,
+            # but not its contents.
+            sigquit_interrupt = [False]
+
+            def sigquit_handler(*args):
+                sigquit_interrupt[0] = True
+
+            signal.signal(signal.SIGQUIT, sigquit_handler)
 
             for size, controller, tb in tarballs:
                 # Sanity check source tar ball path
@@ -327,9 +395,19 @@ def main(options, name):
                     # can't/won't be retried.
                     with ie_filepath.open(mode="w") as fp:
                         idxctx.logger.debug("begin indexing")
-                        es_res = es_index(
-                            idxctx.es, actions, fp, idxctx.logger, idxctx._dbg
-                        )
+                        try:
+                            signal.signal(signal.SIGINT, sigint_handler)
+                            es_res = es_index(
+                                idxctx.es, actions, fp, idxctx.logger, idxctx._dbg
+                            )
+                        except SigIntException:
+                            idxctx.logger.exception(
+                                "Indexing interrupted by SIGINT, continuing to next tarball"
+                            )
+                            continue
+                        finally:
+                            # Turn off the SIGINT handler when not indexing.
+                            signal.signal(signal.SIGINT, signal.SIG_IGN)
                 except UnsupportedTarballFormat as e:
                     idxctx.logger.warning("Unsupported tar ball format: {}", e)
                     tb_res = 4
@@ -352,6 +430,11 @@ def main(options, name):
                         "Can't unpack tar ball into {}: {}", ptb.extracted_root, e
                     )
                     tb_res = 11
+                except SigTermException:
+                    idxctx.logger.exception(
+                        "Indexing interrupted by SIGTERM, terminating"
+                    )
+                    break
                 except Exception as e:
                     idxctx.logger.exception("Other indexing error: {}", e)
                     tb_res = 12
@@ -375,6 +458,10 @@ def main(options, name):
                 except _filenotfounderror:
                     # Above operation never made it to actual indexing, ignore.
                     pass
+                except SigTermException:
+                    # Re-raise a SIGTERM to avoid it being lumped in with
+                    # general exception handling below.
+                    raise
                 except Exception:
                     idxctx.logger.exception(
                         "Unexpected error handling" " indexing errors file: {}",
@@ -395,6 +482,10 @@ def main(options, name):
                     # Unconditionally remove the indexing errors file.
                     try:
                         os.remove(ie_filepath)
+                    except SigTermException:
+                        # Re-raise a SIGTERM to avoid it being lumped in with
+                        # general exception handling below.
+                        raise
                     except Exception:
                         pass
                 # Distinguish failure cases, so we can retry the indexing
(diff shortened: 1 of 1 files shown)


server/bin/pbench-index.py
```diff
@@ -36,6 +37,28 @@
...
 
+class CustomError1(Exception):
+    pass
+
+
+class CustomError2(Exception):
+    pass
+
+
+def handler1(signum, frame):
```
ndokos, line 48: I would name the handlers and the exceptions they raise less generically: `sigint_handler` would raise `SigIntException` (it's not an error so why add `Error` to the name?). and `sigterm_handler` with `SigTermException`.

server/bin/pbench-index.py
```diff
@@ -36,6 +37,28 @@
...
+
+
+def handler2(signum, frame):
+    raise CustomError2()
+
+
+# Establish signal handler for INT, SIGHUP, TERM.
+# signal.signal(signal.SIGHUP, handler)
+signal.signal(signal.SIGINT, handler1)
+signal.signal(signal.SIGTERM, handler2)
```
ndokos, line 59: What happens if a SIGINT or SIGTERM arrives outside of the `try/except` where the exception is caught? I don't know that anything bad would happen, but I'd like to make sure that you have thought about that particular situation.

server/bin/pbench-index.py
```diff
@@ -330,6 +353,12 @@ def main(options, name):
                         es_res = es_index(
                             idxctx.es, actions, fp, idxctx.logger, idxctx._dbg
                         )
+                except CustomError1:
+                    idxctx.logger.exception("Continuing with next TarBall")
+                    continue
+                except CustomError2:
+                    idxctx.logger.exception("Breaking out of the loop")
+                    break
```
ndokos, line 361: Assuming I've got it right, this goes to the `finally` on l.485: is that correct?

ndokos (changes requested):

server/bin/pbench-index.py
riya-17: yeah

server/bin/pbench-index.py
riya-17: At the time of 'continue' i.e. SIGINT. It will just continue with the next iteration. 
At the time of 'break' i.e. SIGTERM, it will execute the else in l.482 and the finally in l.485.

server/bin/pbench-index.py
dbutenhof: Thanks, Nick; I actually said the same thing, but I just discovered this morning that I got distracted and never submitted my comments yesterday!

server/bin/pbench-index.py
```diff
@@ -36,6 +37,28 @@
...
+class SigTermException(Exception):
+    pass
+
+
+def sigint_handler(signum, frame):
+    raise SigIntException
+
+
+def sigterm_handler(signum, frame):
+    raise SigTermException()
```
ndokos, line 53: Parens or no parens? Make up your mind :-)

ndokos (changes requested):

server/bin/pbench-index.py
ndokos: Well, the signal handlers get activated early, so if a signal is delivered to the process when we are *NOT* inside the for-loop where the exceptions are handled, the handler will be called and the exception will be generated, but there will be no exception handler to handle it, so presumably the program will terminate abnormally. You can simulate that that's what happens by putting a sleep() at some place before the loop is entered and then delivering a signal to the process. I *think* that that's OK, but I'm wondering if it would be preferable to activate the signal handlers just before the for-loop, rather than at the beginning of the program. Just thinking aloud here: it's not clear to me that one way is "better" than the other.

server/bin/pbench-index.py
ndokos: Welcome to the club! Too many tabs open in too many browser windows...

server/bin/pbench-index.py
dbutenhof: Ha; I missed that. Yeah, I assume we probably want to actually construct exception objects in both case.

server/bin/pbench-index.py
riya-17: yes, I like the idea to declare signal handlers just before the for-loop. Although I am also not sure which would be better, I will try to get deeper into it.

server/bin/pbench-index.py
riya-17: parens

server/bin/pbench-index.py
```diff
@@ -330,6 +353,12 @@ def main(options, name):
                         es_res = es_index(
                             idxctx.es, actions, fp, idxctx.logger, idxctx._dbg
                         )
+                except SigIntException:
+                    idxctx.logger.exception("Continuing with next TarBall")
+                    continue
+                except SigTermException:
+                    idxctx.logger.exception("Breaking out of the loop")
```
dbutenhof, line 360: These should specifically call out the reason for the continue/break: for example,

`Indexing interrupted by SIGINT, continuing to next tarball`
`Indexing interrupted by SIGTERM, terminating`

And eventually, for `SIGHUP`, something like
`Indexing interrupted by SIGHUP, re-evaluating work`

riya-17: ok sure

server/bin/pbench-index.py
riya-17: @ndokos I tested it. It is working fine

server/bin/pbench-index.py
```diff
@@ -292,6 +309,11 @@ def main(options, name):
             skipped = Path(tmpdir, f"{name}.{idxctx.TS}.skipped")
             ie_filepath = Path(tmpdir, f"{name}.{idxctx.TS}.indexing-errors.json")
 
+            # Establish signal handler for INT, SIGHUP, TERM.
+            # signal.signal(signal.SIGHUP, handler)
+            signal.signal(signal.SIGINT, sigint_handler)
+            signal.signal(signal.SIGTERM, sigterm_handler)
```
portante, line 315: Once you establish these signal handlers, they are enabled for the rest of the time the process is running.  This means that we can get weird exceptions raised in contexts which we don't catch them.

Let's consider other implementation alternatives.

portante (changes requested):
Perhaps we can consider different establishment patterns for the four behaviors we might be looking for:

 1. Gracefully stop processing tar balls
    - The current tar ball is indexed until completion, but no other tar balls are processed
    - `SIGQUIT`
    - Handler Behavior:
        - Sets a flag that causes the code flow to `break` out of the `for` loop at line 317
        - Does not raise an exception
 1. Interrupt the current tar ball being indexed, and proceed to the next one, if any
    - `SIGINT`
    - Handler Behavior:
        - `try/except/finally` placed immediately around the `es_index()` call so that the signal handler will only be established for the duration of the call
        - Raises an exception caught by above `try/except/finally`
        - The `finally` clause would take down the signal handler
 1. Report the current state of indexing
    - Current tar ball being processed
    - How many remaining
    - How many have been completed
    - Errors encountered
    - `SIGHUP`
    - Handler Behavior:
        - Kicks a special thread to log indexing state data
        - No exception raised
 1. Stop processing tar balls immediately and exit gracefully
    - `SIGTERM`
    - Handler Behavior:
        - Raises an exception caught be a new, outer-most, `try/except` block that does not have a `finally` clause (as you don't want any code execution in the `finally` block)

riya-17:
@portante The difference between SIGQUIT and SIGTERM is SIGQUIT will keep processing the current tarball and after completion, it will break, On the other hand, SIGTERM must break immediately whenever it is called.
Am I understanding it correctly?

server/bin/pbench-index.py
```diff
@@ -327,9 +354,24 @@ def main(options, name):
...
+                                idxctx.es, actions, fp, idxctx.logger, idxctx._dbg
+                            )
+                        except SigIntException:
+                            idxctx.logger.exception(
+                                "Indexing interrupted by SIGINT, continuing to next tarball"
+                            )
+                            continue
+                        finally:
+                            # take down signal handler SIGINT
+                            pass
```
portante, line 369: Here you need to make a call to `signal.signal()` to remove the signal handler.

portante (changes requested):
Looks good, just a few more changes called out, and can we capture the description I wrote for how the signal handlers are intended to work and add that as a comment in the code?

server/bin/pbench-index.py
riya-17: there is already a try block in l.308. won't that work?

riya-17:
@portante  `The handler for SIGTERM raises an exception, which can happen any time after signal.signal() returns. But no handler is established yet. So it seems we need to catch this exception in the try/except block surrounding line 320.`
there is already a try block in l.308. won't that work?

portante:
> @portante The difference between SIGQUIT and SIGTERM is SIGQUIT will keep processing the current tarball and after completion, it will break, On the other hand, SIGTERM must break immediately whenever it is called.
> Am I understanding it correctly?

Yes, I think so.

server/bin/pbench-index.py
portante: See PR https://github.com/riya-17/pbench/pull/4.

portante (changes requested):
See PR https://github.com/riya-17/pbench/pull/4, and it needs a rebase.

portante (changes requested):
Can you rebase and squash these commits to one?

portante (dismissed):

server/bin/pbench-index.py
```diff
@@ -102,6 +119,45 @@ def main(options, name):
...
+
+        3. Report the current state of indexing
+            - SIGHUP
+            - Current tar ball being processed
+            - How many remaining
+            - How many have been completed
+            - Errors encountered
+            - Handler Behavior:
+                - Kicks a special thread to log indexing state data
+                - No exception raised
```
dbutenhof, line 153: Since we're not actually doing this, we shouldn't be documenting it.

portante: @riya-17, can you file an issue to track the above instead?

riya-17: yeah sure

portante (changes requested):
Needs a rebase before it can be merged.

dbutenhof (approved):

portante (approved):

portante:
@ndokos, can you review?  You still have a change-requested on this PR.

portante (changes requested):
@riya-17, needs another rebase.

server/bin/pbench-index.py
```diff
@@ -102,6 +119,35 @@ def main(options, name):
          8 - Unable to load and process expected mapping files
          9 - Unable to update index templates in configured Elasticsearch
              instance
+
+        Signal Handlers used to establish different patterns for the four
```
ndokos, line 123: What are the four behaviors? I only see three.

server/bin/pbench-index.py
```diff
@@ -292,6 +350,16 @@ def main(options, name):
             skipped = Path(tmpdir, f"{name}.{idxctx.TS}.skipped")
             ie_filepath = Path(tmpdir, f"{name}.{idxctx.TS}.indexing-errors.json")
 
+            # We use a list object here so that when we close over this
```
ndokos, line 353: I don't understand this: why can't we use just a boolean?

server/bin/pbench-index.py
ndokos, line 123: What are the four behaviors? I see only three.

server/bin/pbench-index.py
ndokos, line 353: I don't understand this: why isn't a boolean enough?

ndokos (changes requested):
Looks good overall. Off-by-one error in a comment (?) and one implementation question.

server/bin/pbench-index.py
portante: When you "close" over a variable in a function definition, the value of that variable remains as it was for each invocation of that function when that function is not called in the same scope.  Here is an example:

```
import sys

def setup_closure(value_i, value_l):
    def func_closure():
        print(f"value_i = {value_i!r}; value_l = {value_l!r}")
    return func_closure

def main(argv):
    value_i = 42
    value_l = [ 42 ]

    def func_closure():
        print(f"value_i = {value_i!r}; value_l = {value_l!r}")

    # The next two calls work as expected because the scope
    # of value_i and value_l[0] are the same.
    func_closure()
    value_i += 1
    value_l[0] += 1
    func_closure()

    cl = setup_closure(value_i, value_l)

    # The next two calls don't work as expected because the
    # scope of value_i is not the same.
    cl()
    value_i += 1
    value_l[0] += 1
    cl()

    return 0

if __name__ == "__main__":
    sys.exit(main(sys.argv))
```

portante (changes requested):
... and needs another rebase.

portante (approved):

dbutenhof (approved):

portante:
@ndokos, you still have a "change requested" posted against this PR.  Is there anything else you think @riya-17 needs to do?

ndokos (approved):

PR #1942, opened 2020-11-02
Contribute generate-pbench-timeseries-graphs from `.csv` files in a pbench result hierarchy

RobertKrawitz:
Generate time series graphs from pbench .csv files.

*generate-pbench-timeseries-graphs* -- generate graphs from .csv files from Pbench raw output

Usage (for now): process-timeseries [files...] | bash

Lots of TODOs:

1. Command line arguments for image size, max # of data sets, what else?
2. Actually feed the output into gnuplot rather than just sending it to stdout
3. Optionally sort data sets by max value or max average
4. Combine multiple samples into 1.
   This will require that generate-pbench-timeseries-graphs enumerate the files
   (or accept the file list on stdin rather than command line) because the command
   line won't be able to accept all of the files in a data set.
5. Handle .csv files from perf and any other tools I've missed.
6. Allow scaling of Y and possibly X axis.
7. Allow command line override of per-file settings.
8. Allow additional gnuplot options
9. Error checking
10. Make data max/minvals available to graph generator so that it can set
   the axis bounds in gnuplot correctly.
11->Graham's number ...

### DIFF
--- a/contrib/analysis/generate-pbench-timeseries-graphs
+++ b/contrib/analysis/generate-pbench-timeseries-graphs
@@ -585,6 +585,13 @@ def read_file(filename):
                         base_ts = float(vals[0])
                     coldata[0].append((float(vals[0]) - base_ts) / timestamp_divisor)
                     for i in range(1, len(vals)):
+                        # Not ideal, but if there's a missing value,
+                        # probably the best we can do
+                        try:
+                            val = float(vals[i])
+                        except:
+                            eprint("Missing value in %s at row %d, col %d" % (filename, rows_read, i))
+                            val = 0.0
                         coldata[i].append(val)
                         abssum.append(abs(val))
                         if val > maxval[i]:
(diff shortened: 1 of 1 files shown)


RobertKrawitz:
Doesn't seem to be allowing me to edit the description, but this is out there for purpose of discussion at this time.

portante (commented):
Is there a possibility of doing this in Python 3 instead of Perl?  We are actively moving away from Perl where possible.

RobertKrawitz:
Control-meta-*sigh*...

dbutenhof:
> Control-meta-_sigh_...

Ha. It's funny; I used to be a Perl fan, and actively avoided the space-sensitive indentation model of Python. I finally got "rescued" and now Perl looks really awkward to me.

We really want to avoid adding more Perl dependencies to Pbench. On the other hand, as a "proof of concept" and reference source, this is a great step.

I also really like your comment about generating graphics for a paper or presentation; we should keep that in mind as we extend the dashboard and backend.

portante:
@RobertKrawitz, this looks good.  Do you want to pull it out of `Draft` mode and we can merge it?

portante (approved):

portante:
Flubbed this merge.  I failed to notice that it wasn't rebased, and had 7 commits.  So I rebased and collapsed the commits to one in ad3e4d390.

Issue #1938, opened 2020-11-02
Splitting pbench-index.py functionality into different modules


riya-17:
At present there is a single main function in pbench-index.py due to which dealing with unit-tests is quite difficult. Splitting it in different functions like:

- initialising main
- collection of tarballs
- processing tarballs

PR #1943, opened 2020-11-03
Splitting pbench-index code into modules

riya-17:
Addresses issue #1938

### DIFF
--- a/lib/pbench/server/indexing_tarballs.py
+++ b/lib/pbench/server/indexing_tarballs.py
@@ -1,0 +1,510 @@
+"""Initialising Indexing class"""
+
+import os
+import glob
+import signal
+import tarfile
+import tempfile
+from pathlib import Path
+
+from pbench.common.exceptions import (
+    BadDate,
+    UnsupportedTarballFormat,
+    SosreportHostname,
+    BadMDLogFormat,
+    TemplateError,
+)
+from pbench.server import tstos
+from pbench.server.indexer import (
+    PbenchTarBall,
+    es_index,
+    VERSION,
+)
+from pbench.server.report import Report
+from pbench.server.utils import rename_tb_link, quarantine
+
+
+class SigIntException(Exception):
+    pass
+
+
+def sigint_handler(*args):
+    raise SigIntException()
+
+
+class SigTermException(Exception):
+    pass
+
+
+def _count_lines(fname):
+    """Simple method to count the lines of a file.
+    """
+    try:
+        with open(fname, "r") as fp:
+            cnt = sum(1 for line in fp)
+    except FileNotFoundError:
+        cnt = 0
+    return cnt
+
+
+class Index:
+    """class used to collect tarballs and index them"""
+
+    def __init__(self, name, options, idxctx, incoming):
+
+        self.options = options
+        _re_idx = "RE-" if options.re_index else ""
+        if options.index_tool_data:
+            # The link source and destination for the operation of this script
+            # when it only indexes tool data.
+            self.linksrc = "TO-INDEX-TOOL"
+            self.linkdest = "INDEXED"
+        else:
+            # The link source and destination for the operation of this script
+            # when it indexes run, table-of-contents, and result data.
+            self.linksrc = f"TO-{_re_idx}INDEX"
+            self.linkdest = "TO-INDEX-TOOL"
+        # We only ever use a symlink'd error destination for indexing
+        # problems.
+        self.linkerrdest = "WONT-INDEX"
+        self.idxctx = idxctx
+        self.incoming = incoming
+        self.name = name
+
+    def collect_tb(self, ARCHIVE_rp, qdir):
+        """ Collect tarballs that needs indexing"""
+
+        # find -L $ARCHIVE/*/$linksrc -name '*.tar.xz' -printf "%s\t%p\n" 2>/dev/null | sort -n > $list
+        tarballs = []
+        idxctx = self.idxctx
+        try:
+            tb_glob = os.path.join(ARCHIVE_rp, "*", self.linksrc, "*.tar.xz")
+            for tb in glob.iglob(tb_glob):
+                try:
+                    rp = Path(tb).resolve(strict=True)
+                except OSError:
+                    idxctx.logger.warning("{} does not resolve to a real path", tb)
+                    quarantine(qdir, idxctx.logger, tb)
+                    continue
+                controller_path = rp.parent
+                controller = controller_path.name
+                archive_path = controller_path.parent
+                if str(archive_path) != str(ARCHIVE_rp):
+                    idxctx.logger.warning(
+                        "For tar ball {}, original home is not {}", tb, ARCHIVE_rp
+                    )
+                    quarantine(qdir, idxctx.logger, tb)
+                    continue
+                if not Path(f"{rp}.md5").is_file():
+                    idxctx.logger.warning("Missing .md5 file for {}", tb)
+                    quarantine(qdir, idxctx.logger, tb)
+                    # Audit should pick up missing .md5 file in ARCHIVE directory.
+                    continue
+                try:
+                    # get size
+                    size = rp.stat().st_size
+                except OSError:
+                    idxctx.logger.warning("Could not fetch tar ball size for {}", tb)
+                    quarantine(qdir, idxctx.logger, tb)
+                    # Audit should pick up missing .md5 file in ARCHIVE directory.
+                    continue
+                else:
+                    tarballs.append((size, controller, tb))
+        except SigTermException:
+            # Re-raise a SIGTERM to avoid it being lumped in with general
+            # exception handling below.
+            raise
+        except Exception:
+            idxctx.logger.exception(
+                "Unexpected error encountered generating list"
+                " of tar balls to process"
+            )
+            # tuple to return the status and return value
+            return (12, [])
+        else:
+            if not tarballs:
+                idxctx.logger.info("No tar balls found that need processing")
+                return (0, [])
+
+        return (0, sorted(tarballs))
+
+    def process_tb(self, tarballs):
+        """Process Tarballs For Indexing and creates report
+
+            "tarballs" - List of tarball, it is the second value of
+                the tuple returned by collect_tb() """
+
+        # We always process the smallest tar balls first.
+        idxctx = self.idxctx
+
+        # At this point, tarballs contains a list of tar balls sorted by size
+        # that were available as symlinks in the various 'linksrc' directories.
+        idxctx.logger.debug("Preparing to index {:d} tar balls", len(tarballs))
+
+        try:
+            # Now that we are ready to begin the actual indexing step, ensure we
+            # have the proper index templates in place.
+            idxctx.logger.debug("update_templates [start]")
+            idxctx.templates.update_templates(idxctx.es)
+        except TemplateError as e:
+            idxctx.logger.error("update_templates [end], error {}", repr(e))
+            res = 9
+        except SigTermException:
+            # Re-raise a SIGTERM to avoid it being lumped in with general
+            # exception handling below.
+            raise
+        except Exception:
+            idxctx.logger.exception(
+                "update_templates [end]: Unexpected template" " processing error"
+            )
+            res = 12
+        else:
+            idxctx.logger.debug("update_templates [end]")
+            res = 0
+
+        if res != 0:
+            # Exit early if we encounter any errors.
+            return res
+
+        report = Report(
+            idxctx.config,
+            self.name,
+            es=idxctx.es,
+            pid=idxctx.getpid(),
+            group_id=idxctx.getgid(),
+            user_id=idxctx.getuid(),
+            hostname=idxctx.gethostname(),
+            version=VERSION,
+            templates=idxctx.templates,
+        )
+        # We use the "start" report ID as the tracking ID for all indexed
+        # documents.
+        try:
+            tracking_id = report.post_status(tstos(idxctx.time()), "start")
+        except SigTermException:
+            # Re-raise a SIGTERM to avoid it being lumped in with general
+            # exception handling below.
+            raise
+        except Exception:
+            idxctx.logger.error("Failed to post initial report status")
+            return 12
+        else:
+            idxctx.set_tracking_id(tracking_id)
+
+        with tempfile.TemporaryDirectory(
+            prefix=f"{self.name}.", dir=idxctx.config.TMP
+        ) as tmpdir:
+            idxctx.logger.debug("start processing list of tar balls")
+            tb_list = Path(tmpdir, f"{self.name}.{idxctx.TS}.list")
+            try:
+                with tb_list.open(mode="w") as lfp:
+                    # Write out all the tar balls we are processing so external
+                    # viewers can follow along from home.
+                    for size, controller, tb in tarballs:
+                        print(f"{size:20d} {controller} {tb}", file=lfp)
+
+                indexed = Path(tmpdir, f"{self.name}.{idxctx.TS}.indexed")
+                erred = Path(tmpdir, f"{self.name}.{idxctx.TS}.erred")
+                skipped = Path(tmpdir, f"{self.name}.{idxctx.TS}.skipped")
+                ie_filepath = Path(
+                    tmpdir, f"{self.name}.{idxctx.TS}.indexing-errors.json"
+                )
+
+                # We use a list object here so that when we close over this
+                # variable in the handler, the list object will be closed over,
+                # but not its contents.
+                sigquit_interrupt = [False]
+
+                def sigquit_handler(*args):
+                    sigquit_interrupt[0] = True
+
+                signal.signal(signal.SIGQUIT, sigquit_handler)
+
+                try:
+                    for size, controller, tb in tarballs:
+                        # Sanity check source tar ball path
+                        linksrc_dir = Path(tb).parent
+                        linksrc_dirname = linksrc_dir.name
+                        assert linksrc_dirname == self.linksrc, (
+                            f"Logic bomb!  tar ball "
+                            f"path {tb} does not contain {self.linksrc}"
+                        )
+
+                        idxctx.logger.info("Starting {} (size {:d})", tb, size)
+
+                        ptb = None
+                        try:
+                            # "Open" the tar ball represented by the tar ball object
+                            idxctx.logger.debug("open tar ball")
+                            ptb = PbenchTarBall(
+                                idxctx,
+                                os.path.realpath(tb),
+                                tmpdir,
+                                Path(self.incoming, controller),
+                            )
+
+                            # Construct the generator for emitting all actions.  The
+                            # `idxctx` dictionary is passed along to each generator so
+                            # that it can add its context for error handling to the
+                            # list.
+                            idxctx.logger.debug("generator setup")
+                            if self.options.index_tool_data:
+                                actions = ptb.mk_tool_data_actions()
+                            else:
+                                actions = ptb.make_all_actions()
+
+                            # File name for containing all indexing errors that
+                            # can't/won't be retried.
+                            with ie_filepath.open(mode="w") as fp:
+                                idxctx.logger.debug("begin indexing")
+                                try:
+                                    signal.signal(signal.SIGINT, sigint_handler)
+                                    es_res = es_index(
+                                        idxctx.es,
+                                        actions,
+                                        fp,
+                                        idxctx.logger,
+                                        idxctx._dbg,
+                                    )
+                                except SigIntException:
+                                    idxctx.logger.exception(
+                                        "Indexing interrupted by SIGINT, continuing to next tarball"
+                                    )
+                                    continue
+                                finally:
+                                    # Turn off the SIGINT handler when not indexing.
+                                    signal.signal(signal.SIGINT, signal.SIG_IGN)
+                        except UnsupportedTarballFormat as e:
+                            idxctx.logger.warning("Unsupported tar ball format: {}", e)
+                            tb_res = 4
+                        except BadDate as e:
+                            idxctx.logger.warning("Bad Date: {!r}", e)
+                            tb_res = 5
+                        except FileNotFoundError as e:
+                            idxctx.logger.warning("No such file: {}", e)
+                            tb_res = 6
+                        except BadMDLogFormat as e:
+                            idxctx.logger.warning(
+                                "The metadata.log file is curdled in" " tar ball: {}", e
+                            )
+                            tb_res = 7
+                        except SosreportHostname as e:
+                            idxctx.logger.warning("Bad hostname in sosreport: {}", e)
+                            tb_res = 10
+                        except tarfile.TarError as e:
+                            idxctx.logger.error(
+                                "Can't unpack tar ball into {}: {}",
+                                ptb.extracted_root,
+                                e,
+                            )
+                            tb_res = 11
+                        except SigTermException:
+                            idxctx.logger.exception(
+                                "Indexing interrupted by SIGTERM, terminating"
+                            )
+                            break
+                        except Exception as e:
+                            idxctx.logger.exception("Other indexing error: {}", e)
+                            tb_res = 12
+                        else:
+                            beg, end, successes, duplicates, failures, retries = es_res
+                            idxctx.logger.info(
+                                "done indexing (start ts: {}, end ts: {}, duration:"
+                                " {:.2f}s, successes: {:d}, duplicates: {:d},"
+                                " failures: {:d}, retries: {:d})",
+                                tstos(beg),
+                                tstos(end),
+                                end - beg,
+                                successes,
+                                duplicates,
+                                failures,
+                                retries,
+                            )
+                            tb_res = 1 if failures > 0 else 0
+                        try:
+                            ie_len = ie_filepath.stat().st_size
+                        except FileNotFoundError:
+                            # Above operation never made it to actual indexing, ignore.
+                            pass
+                        except SigTermException:
+                            # Re-raise a SIGTERM to avoid it being lumped in with
+                            # general exception handling below.
+                            raise
+                        except Exception:
+                            idxctx.logger.exception(
+                                "Unexpected error handling" " indexing errors file: {}",
+                                ie_filepath,
+                            )
+                        else:
+                            # Success fetching indexing error file size.
+                            if ie_len > len(tb) + 1:
+                                try:
+                                    report.post_status(
+                                        tstos(end), "errors", ie_filepath
+                                    )
+                                except Exception:
+                                    idxctx.logger.exception(
+                                        "Unexpected error issuing"
+                                        " report status with errors: {}",
+                                        ie_filepath,
+                                    )
+                        finally:
+                            # Unconditionally remove the indexing errors file.
+                            try:
+                                os.remove(ie_filepath)
+                            except SigTermException:
+                                # Re-raise a SIGTERM to avoid it being lumped in with
+                                # general exception handling below.
+                                raise
+                            except Exception:
+                                pass
+                        # Distinguish failure cases, so we can retry the indexing
+                        # easily if possible.  Different `linkerrdest` directories for
+                        # different failures; the rest are going to end up in
+                        # `linkerrdest` for later retry.
+                        controller_path = linksrc_dir.parent
+
+                        if tb_res == 0:
+                            idxctx.logger.info(
+                                "{}: {}/{}: success",
+                                idxctx.TS,
+                                controller_path.name,
+                                os.path.basename(tb),
+                            )
+                            # Success
+                            with indexed.open(mode="a") as fp:
+                                print(tb, file=fp)
+                            rename_tb_link(
+                                tb, Path(controller_path, self.linkdest), idxctx.logger
+                            )
+                        elif tb_res == 1:
+                            idxctx.logger.warning(
+                                "{}: index failures encountered on {}", idxctx.TS, tb
+                            )
+                            with erred.open(mode="a") as fp:
+                                print(tb, file=fp)
+                            rename_tb_link(
+                                tb,
+                                Path(controller_path, f"{self.linkerrdest}.1"),
+                                idxctx.logger,
+                            )
+                        elif tb_res in (2, 3):
+                            assert False, (
+                                f"Logic Bomb!  Unexpected tar ball handling "
+                                f"result status {tb_res:d} for tar ball {tb}"
+                            )
+                        elif tb_res >= 4 and tb_res <= 11:
+                            # # Quietly skip these errors
+                            with skipped.open(mode="a") as fp:
+                                print(tb, file=fp)
+                            rename_tb_link(
+                                tb,
+                                Path(controller_path, f"{self.linkerrdest}.{tb_res:d}"),
+                                idxctx.logger,
+                            )
+                        else:
+                            idxctx.logger.error(
+                                "{}: index error {:d} encountered on {}",
+                                idxctx.TS,
+                                tb_res,
+                                tb,
+                            )
+                            with erred.open(mode="a") as fp:
+                                print(tb, file=fp)
+                            rename_tb_link(
+                                tb,
+                                Path(controller_path, self.linkerrdest),
+                                idxctx.logger,
+                            )
+                        idxctx.logger.info(
+                            "Finished{} {} (size {:d})",
+                            "[SIGQUIT]" if sigquit_interrupt[0] else "",
+                            tb,
+                            size,
+                        )
+
+                        if sigquit_interrupt[0]:
+                            break
+                except SigTermException:
+                    idxctx.logger.exception(
+                        "Indexing interrupted by SIGQUIT, stop processing tarballs"
+                    )
+                finally:
+                    # Turn off the SIGQUIT handler when not indexing.
+                    signal.signal(signal.SIGQUIT, signal.SIG_IGN)
+            except SigTermException:
+                # Re-raise a SIGTERM to avoid it being lumped in with general
+                # exception handling below.
+                raise
+            except Exception:
+                idxctx.logger.exception("Unexpected setup error")
+                res = 12
+            else:
+                # No exceptions while processing tar ball, success.
+                res = 0
+            finally:
+                if idxctx:
+                    idxctx.dump_opctx()
+                idxctx.logger.debug("stopped processing list of tar balls")
+
+                idx = _count_lines(indexed)
+                skp = _count_lines(skipped)
+                err = _count_lines(erred)
+
+                idxctx.logger.info(
+                    "{}.{}: indexed {:d} (skipped {:d}) results," " {:d} errors",
+                    self.name,
+                    idxctx.TS,
+                    idx,
+                    skp,
+                    err,
+                )
+
+                if err > 0:
+                    if skp > 0:
+                        subj = (
+                            f"{self.name}.{idxctx.TS} - Indexed {idx:d} results, skipped {skp:d}"
+                            f" results, w/ {err:d} errors"
+                        )
+                    else:
+                        subj = (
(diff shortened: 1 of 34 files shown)


server/bin/pbench-index.py
```diff
@@ -102,130 +89,39 @@ def main(options, name):
...
-            idxctx.logger.info("No tar balls found that need processing")
-            return 0
+        with open(fname, "r") as fp:
+            cnt = sum(1 for line in fp)
+    except _filenotfounderror:
+        cnt = 0
+    return cnt
 
+
+def process_tb(options, idxctx, sd_obj, name, INCOMING_rp, tarballs):
```
dbutenhof, line 124: The "collect" and "process" functions could be abstracted into a class with shared instance data.

server/bin/pbench-index.py
```diff
@@ -102,130 +89,39 @@ def main(options, name):
...
-        qdir_path = idxctx.config.get_valid_dir_option(
-            "QDIR", Path(qdir), idxctx.logger
-        )
-        if not qdir_path:
-            res = 3
-
-    if res != 0:
-        # Exit early if we encounter any errors.
-        return res
+class LinkSrcDestObj:
```
dbutenhof, line 95: This is just setting up some fixed instance data based on the options; it would make more sense abstracted into the initialization of an `Index` Class instance that would also encapsulate your `process` and `collect` functions. That would get the "orchestration" and linkage data between the two out of `main`. Essentially, `main` would just process the options, instantiate an object, and invoke it to do the work.

server/bin/pbench-index.py
```diff
@@ -517,6 +418,131 @@ def main(options, name):
     return res
 
 
+def collect_tb(idxctx, sd_obj, ARCHIVE_rp, qdir):
+    # find -L $ARCHIVE/*/$linksrc -name '*.tar.xz' -printf "%s\t%p\n" 2>/dev/null | sort -n > $list
+    tarballs = []
+    try:
+        tb_glob = os.path.join(ARCHIVE_rp, "*", sd_obj.linksrc, "*.tar.xz")
+        for tb in glob.iglob(tb_glob):
```
npalaska, line 426: I guess we can replace `os.path` with `pathlib.Path` to be more consistent 
here something like
```
tb_glob = pathlib.Path(ARCHIVE_rp, "*", sd_obj.linksrc)
for tb in tb_glob.rglob("*.tar.xz")
```

riya-17: it seems Path.glob doesn't deal with "*" in between their path due to which it is creating error "(ARCHIVE_rp, "*", sd_obj.linksrc)" like here

portante: Yes, not sure you can use `pathlib` here effectively.

portante (changes requested):
The goal of `pbench-index.py` is to:

 - Find all the tar balls that need to be processed

 - Process each tar ball found, indexing it into Elasticsearch

We can start to move this code to a more service oriented architecture.  Let's consider making a fifo queue for which the "find the tar balls" generator pushes them into the queue.  Then we can have N indexers which handle the indexing by pulling one entry off the queue and processing it.

Thoughts?

riya-17:
@portante I believe deque won't be the best choice as we are going over the list of tarballs twice. Also, I wanted to ask what does 'N indexers' meant?

portante:
@riya-17, where do we go over the list of tar balls twice?

riya-17:
> @riya-17, where do we go over the list of tar balls twice?

First in line l.182 for printing and another in line l.192 for processing.

dbutenhof:
> > @riya-17, where do we go over the list of tar balls twice?
> 
> First in line l.182 for printing and another in line l.192 for processing.

It's debatable whether we really need the printing loop; it's not going to the log, and the server processes run in the background so I'm not sure anyone would ever be looking at this anyway. If we want to retain that, we could print when we stick items onto the FIFO in the first place rather than iterating over the discovered set twice. (Though we might want to consider an `info` log for each tarball we index anyway, in case we ever need to trace the history... although eventually we want a persistent state change transaction record showing all the steps for each tarball, and we can regenerate the full life history any time we want; you already made a first step towards tracking tarball lifecycle in another PR.)

The big thing about moving to a "FIFO" sort of model is that eventually something (either some microservice like the current dispatcher, or simply the web server PUT) will validate the received tarball/md5 and send a message (Redis, Rabbit, Kafka, or whatever) which will be handled by one or more indexer microservice instances off the queue.

Right now it's all inside the `pbench-index` command, but decoupling the intake processing from the index work in our object design will move us closer to the model we eventually want.

riya-17:
> > > @riya-17, where do we go over the list of tar balls twice?
> > 
> > 
> > First in line l.182 for printing and another in line l.192 for processing.
> 
> It's debatable whether we really need the printing loop; it's not going to the log, and the server processes run in the background so I'm not sure anyone would ever be looking at this anyway. If we want to retain that, we could print when we stick items onto the FIFO in the first place rather than iterating over the discovered set twice. (Though we might want to consider an `info` log for each tarball we index anyway, in case we ever need to trace the history... although eventually we want a persistent state change transaction record showing all the steps for each tarball, and we can regenerate the full life history any time we want; you already made a first step towards tracking tarball lifecycle in another PR.)
> 
> The big thing about moving to a "FIFO" sort of model is that eventually something (either some microservice like the current dispatcher, or simply the web server PUT) will validate the received tarball/md5 and send a message (Redis, Rabbit, Kafka, or whatever) which will be handled by one or more indexer microservice instances off the queue.
> 
> Right now it's all inside the `pbench-index` command, but decoupling the intake processing from the index work in our object design will move us closer to the model we eventually want.

@dbutenhof yeah, it does make sense but we are also sorting the tarballs, and sorting a deque (sorted(tarballs_deque)) converts it into a list. Either I will have to write a deque sorting algorithm for sorting deque particulrly or will have to work with the list itself.

Which do you think will be a better option?

aquibbaig (approved):
LGTM!

anishaswain (dismissed):

server/bin/pbench-index.py
```diff
@@ -137,20 +159,7 @@ def main(options, name):
...
-        linkdest = "INDEXED"
-    else:
-        # The link source and destination for the operation of this script
-        # when it indexes run, table-of-contents, and result data.
-        linksrc = f"TO-{_re_idx}INDEX"
-        linkdest = "TO-INDEX-TOOL"
-    # We only ever use a symlink'd error destination for indexing
-    # problems.
-    linkerrdest = "WONT-INDEX"
+    # sd_obj = LinkSrcDestObj(options)
```
dbutenhof, line 162: Did you intend to leave this commented? If so, please explain why. We shouldn't be submitting commented code without a persuasive TODO comment explaining why it's commented and under what conditions we expect it to be restored.

server/bin/pbench-index.py
```diff
@@ -181,393 +190,16 @@ def main(options, name):
...
-            "update_templates [end]: Unexpected template" " processing error"
-        )
-        res = 12
-    else:
-        idxctx.logger.debug("update_templates [end]")
-        res = 0
-
-    if res != 0:
-        # Exit early if we encounter any errors.
+    if isinstance(res, int):
```
dbutenhof, line 196: Overloading the return value really isn't a good strategy for errors. I think we should probably raise exceptions that are handled by the driver code (here) and decoded to defined exit codes if that's really necessary.

In retrospect, I wonder if the distinct exit codes (aside from the 0/success or nonzero/failure meaningful to the shell) really serve any useful purpose.

We don't need to fix that completely in this PR, but I would like to fix the return value overloading we have here. One simple alternative would be to return a tuple of `(status, tarball_list)` which would be for example `(2, [])` on a failure or `(0, [...])` on success.

dbutenhof (changes requested):

server/bin/pbench-index.py
portante: +1 to using a tuple return value.

With regard to status code values, we used those codes to debug why the Tool Meister was not running problem in an environment, so they can be very useful.

dbutenhof: Distinguishing various error conditions is certainly important, but relying on log messages (particularly if they have useful parameters) seems a lot better than trying to backtrack and guess from integer return values without additional context. And if we have log messages, seems to me that the individual integer codes are at best redundant?

portante: Yes, until you don't have logs from the system where the code ran.  =)

server/bin/pbench-index.py
```diff
@@ -181,393 +188,15 @@ def main(options, name):
...
-        "-R",
-        "--re-index",
-        action="store_true",
-        dest="re_index",
-        default=False,
-        help="Perform re-indexing of previously indexed data",
-    )
-    parsed = parser.parse_args()
-    status = main(parsed, run_name)
+    status = main()
```
portante, line 201: Originally, `main()` was a callable method that was not attached to how it got command line parameters.  I had `options` and `run_name` arguments.  It was done this way so that `main()` could be called without outside of a CLI context.

Once you pull the command line parsing code out of it, and place it back at the runtime module level, there is not much left to `main()`.  The behaviors of how tar balls are gathered for indexing are embedded in the new `Index` class.  While this moves code around to reduce the size of `pbench-index.py`, it is not clear how much this helps the situation.

The previous code was really about creating a generator pipeline, of sorts, for a given tar ball, and handing that generator to the es_index() method to send in bulk to Elasticsearch.

We might want to consider making a generator pattern for how tar balls to be indexed are gathered/found/produced, and pass that off to a consumer that deals with them the same way we do today.

We can then make the current method of how they are processed today, and add additional methods that perhaps send them to multiple workers for processing, etc.

portante (changes requested):
See comments in line, this also needs to be rebased.

server/bin/pbench-index.py
riya-17: What if I just split the collecting tarballs function from the whole main module? and keep the other functions intact

server/bin/pbench-index.py
```diff
@@ -116,6 +81,10 @@ def main(options, name):
         )
         return 2
 
+    _name_suf = "-tool-data" if options.index_tool_data else ""
```
portante, line 84: This code block should not be moved, the `name` field is referenced in the `print` statement on line 77 above.

server/bin/pbench-index.py
```diff
@@ -8,55 +8,24 @@
...
-    try:
-        with open(fname, "r") as fp:
-            cnt = sum(1 for line in fp)
-    except _filenotfounderror:
-        cnt = 0
-    return cnt
-
 
 def main(options, name):
+
```
portante, line 28: Not needed.

lib/pbench/server/indexing_tarballs.py
```diff
@@ -0,0 +1,416 @@
...
+                idxctx.logger.info("No tar balls found that need processing")
+                return (1, 0)
+
+        return (0, tarballs)
+
+    def process_tb(self, tarballs):
+        """Process Tarballs For Indexing and creates report"""
+
+        # We always process the smallest tar balls first.
+        tarballs = sorted(tarballs)
```
portante, line 117: Seems like `collect_tb` should handling the sorting.

server/bin/pbench-index.py
```diff
@@ -181,344 +135,15 @@ def main(options, name):
...
-                        f"{name}.{idxctx.TS} - Indexed {idx:d} results, w/ {err:d}"
-                        " errors"
-                    )
-            else:
-                if skp > 0:
-                    subj = f"{name}.{idxctx.TS} - Indexed {idx:d} results, skipped {skp:d} results"
-                else:
-                    subj = f"{name}.{idxctx.TS} - Indexed {idx:d} results"
+    idx = Index(name, options, idxctx, INCOMING_rp)
+    status, value = idx.collect_tb(ARCHIVE_rp, qdir)
```
portante, line 139: The name `value` for the second element of the returned tuple seems a bit, shall we say, "under-descriptive".  ;). The return value of `collect_tb` appears to be a success/failure number, and a list of tar balls.  So why not rename `value` to be `tarballs`?

server/bin/pbench-index.py
```diff
@@ -181,344 +135,15 @@ def main(options, name):
...
-                    print("\nSkipped Results\n===============", file=fp)
-                    with skipped.open() as sfp:
-                        for line in sorted(sfp):
-                            print(line.strip(), file=fp)
-            try:
-                report.post_status(tstos(idxctx.time()), "status", report_fname)
-            except Exception:
-                pass
+    if not status:
+        return idx.process_tb(value)
```
portante, line 142: Here it makes sense to have `value` renamed to be `tarballs`.  And instead of `not status`, consider being more explicit with `status == 0` since that is the success value returned in `status`.

Further, instead of having two `return` statements, and introducing the bug below where `value` is returned incorrectly, why not have the return value of `process_tb(tarballs)` assigned to `status` and just `return status` below?

server/bin/pbench-index.py
```diff
@@ -181,344 +135,15 @@ def main(options, name):
...
 
-    return res
+    return value
```
portante, line 144: See above comment, perhaps should just be `return status`?

portante (changes requested):

server/bin/pbench-index.py
riya-17: yeah

server/bin/pbench-index.py
riya-17: Fixed

lib/pbench/server/indexing_tarballs.py
riya-17: yeah sure

server/bin/pbench-index.py
riya-17: I thought it will cause confusion if it will not be a list of tarballs. Like there are numeric return values 12 and 0

server/bin/pbench-index.py
riya-17: The status is created by us on the basis that whether there are tarballs or not and the value is the return code

portante: Can you show in the code with a comment what contents of value will be when this return statement is executed?

server/bin/pbench-index.py
portante: When is `value` not a list of tar balls?

portante (changes requested):
Please address the `lgtm.com` bot comments, thanks!

server/bin/pbench-index.py
portante: That means it is an empty list of tarballs, still the same thing.  If we are returning `None` instead of an empty list, we should change that to an empty list.

server/bin/pbench-index.py
riya-17: There can be 3  return codes 0, 9, and 12. We have already shown what each return code meant in the main function docstring. Should I show it again with a comment about three of them?

riya-17:
```
except Exception:
    idxctx.logger.exception(
        "Unexpected error encountered generating list"
        " of tar balls to process"
    )   
    # tuple to return the status and return value
    return (1, 12) 
else:
    if not tarballs:
        idxctx.logger.info("No tar balls found that need processing")
        return (1, 0)

return (0, sorted(tarballs))
```

if I  proceed with this code. Here status is the first value and the second is the error code or tarball list. which can be dealt with like the following:
```
status, value = idx.collect_tb(ARCHIVE_rp, qdir)

if status == 0:
        value = idx.process_tb(value)

return value
```
if I rename value to tarballs then it doesn't make much sense as the return value.

if I move with a different approach:

```
except Exception:
    idxctx.logger.exception(
        "Unexpected error encountered generating list"
        " of tar balls to process"
    )   
    # tuple to return the status and return value
    return (1, 12) 
else:
    if not tarballs:
        idxctx.logger.info("No tar balls found that need processing")
        return (0, 0)

return (0, sorted(tarballs))
```

then I will have to check whether the tarballs list is empty or not and also the error code will be returned as 1 if the status conditions are not passed instead of 12
```
status, tarballs = idx.collect_tb(ARCHIVE_rp, qdir)

if status == 0:
    if len(tarballs):
        status = idx.process_tb(tarballs)

    return status
```
Am I missing something?
I believe reaching out to a better name should resolve this other than value and tarballs

lib/pbench/server/indexing_tarballs.py
```diff
@@ -0,0 +1,416 @@
...
+            idxctx.logger.exception(
+                "Unexpected error encountered generating list"
+                " of tar balls to process"
+            )
+            # tuple to return the status and return value
+            return (1, 12)
+        else:
+            if not tarballs:
+                idxctx.logger.info("No tar balls found that need processing")
+                return (1, 0)
```
portante, line 109: Change this to `return (1, [])`, the second tuple always being a list, empty or otherwise.

lib/pbench/server/indexing_tarballs.py
```diff
@@ -0,0 +1,416 @@
...
+                    continue
+                else:
+                    tarballs.append((size, controller, tb))
+        except Exception:
+            idxctx.logger.exception(
+                "Unexpected error encountered generating list"
+                " of tar balls to process"
+            )
+            # tuple to return the status and return value
+            return (1, 12)
```
portante, line 105: Change this to `return (1, [])`, the second tuple always being a list, empty or otherwise.

server/bin/pbench-index.py
portante: Looks like this change has not been reverted ...

server/bin/pbench-index.py
portante: Doesn't seem to be fixed ...

server/bin/pbench-index.py
portante: See the comments in `lib/pbench/server/indexing_tarballs.py`, if we change `collect_tb` to always return a status code and a list tuple, then we do the above code.  Also you don't need to do `if len(tarballs):`, you can achieve the same with `if tarballs:`, an empty list is `False`, and non-empty list is `True`.

portante (changes requested):

lib/pbench/server/indexing_tarballs.py
riya-17: yeah but if there are no tarballs then it is not an error. it exits with exit status as 0(success)

portante: Sorry, `return (0, [])`.

lib/pbench/server/indexing_tarballs.py
```diff
@@ -0,0 +1,415 @@
...
+                    elif tb_res >= 4 or res <= 11:
+                        # # Quietly skip these errors
+                        with skipped.open(mode="a") as fp:
+                            print(tb, file=fp)
+                        rename_tb_link(
+                            tb,
+                            Path(controller_path, f"{self.linkerrdest}.{tb_res:d}"),
+                            idxctx.logger,
+                        )
+                    else:
```
dbutenhof, line 335: LGTM seems to be diagnosing this as unreachable. We've got `if 0 elif 1 elif in (2,3) elif >= 4` ... so, yeah, that covers everything. We can't ever hit the `else`. So what did we really intend here, again? I really do not like the error number logic in this module; it's unclear and poorly documented.

You didn't create the problem, so maybe it doesn't need to be fixed right now; but we need to look at this again.

riya-17: yeah sure

server/bin/pbench-index.py
```diff
@@ -181,344 +134,16 @@ def main(options, name):
...
-                    print("\nSkipped Results\n===============", file=fp)
-                    with skipped.open() as sfp:
-                        for line in sorted(sfp):
-                            print(line.strip(), file=fp)
-            try:
-                report.post_status(tstos(idxctx.time()), "status", report_fname)
-            except Exception:
-                pass
+    if status == 0:
+        if tarballs:
```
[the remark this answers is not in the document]
riya-17: sure

lib/pbench/server/indexing_tarballs.py
portante: Well, line 326 should be `elif tb_res >= 4 and tb_res <= 11:`, then this makes sense.

portante (changes requested):

lib/pbench/server/indexing_tarballs.py
portante: @riya-17, `and` instead of `or`.

portante (changes requested):

dbutenhof (changes requested):
You apparently have merge conflicts to resolve in `pbench-index.py`.

portante (changes requested):
Looks good, just need to address conflicts.

portante (approved):

server/bin/pbench-index.py
riya-17: Sorry, forgot to push the changes

server/bin/pbench-index.py
riya-17: Sorry, forgot to push the changes

server/bin/pbench-index.py
riya-17: ok

lib/pbench/server/indexing_tarballs.py
riya-17: ok

lib/pbench/server/indexing_tarballs.py
riya-17: ok

portante (changes requested):
Just noticed the "LGTM.com" unreachable code warning.  Can you address that?

portante (approved):
Sorry about that, read the wrong `LGTM.com` comment.  All set.

dbutenhof (dismissed):
Of course, taking a step back from the actual changes, the main benefit here is improved ability to do modular pytest unit testing, which I'd like to see attempted since that's the best proof that the modular packaging is "right".

riya-17:
> Of course, taking a step back from the actual changes, the main benefit here is improved ability to do modular pytest unit testing, which I'd like to see attempted since that's the best proof that the modular packaging is "right".

I am also hoping for the best and aiming for that

riya-17:
> Sorry about that, read the wrong `LGTM.com` comment. All set.

I got confused as well earlier but then I realized that it is reporting about a resolved issue

lib/pbench/server/indexing_tarballs.py
riya-17: yes that does resolve it

lib/pbench/server/indexing_tarballs.py
```diff
@@ -0,0 +1,415 @@
...
+            return (12, [])
+        else:
+            if not tarballs:
+                idxctx.logger.info("No tar balls found that need processing")
+                return (0, [])
+
+        return (0, sorted(tarballs))
+
+    def process_tb(self, tarballs):
+        """Process Tarballs For Indexing and creates report"""
```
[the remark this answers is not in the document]
riya-17: sure

lib/pbench/server/indexing_tarballs.py
```diff
@@ -0,0 +1,492 @@
...
+
+class SigIntException(Exception):
+    pass
+
+
+def sigint_handler(*args):
+    raise SigIntException()
+
+
+class SigTermException(Exception):
```
ndokos, line 35: Do we need this here? There is a SigTermException class defined in `pbench-index.py`, together with a signal handler that raises it. So I'm not sure that we need to handle the SigTermException in this module at all, in which case we do not need to define the class. Is this correct?

lib/pbench/server/indexing_tarballs.py
```diff
@@ -0,0 +1,492 @@
...
+                    # get size
+                    size = rp.stat().st_size
+                except OSError:
+                    idxctx.logger.warning("Could not fetch tar ball size for {}", tb)
+                    quarantine(qdir, idxctx.logger, tb)
+                    # Audit should pick up missing .md5 file in ARCHIVE directory.
+                    continue
+                else:
+                    tarballs.append((size, controller, tb))
+        except SigTermException:
```
ndokos, line 113: Do we need to handle SigTermException here?

lib/pbench/server/indexing_tarballs.py
```diff
@@ -0,0 +1,492 @@
...
+
+        try:
+            # Now that we are ready to begin the actual indexing step, ensure we
+            # have the proper index templates in place.
+            idxctx.logger.debug("update_templates [start]")
+            idxctx.templates.update_templates(idxctx.es)
+        except TemplateError as e:
+            idxctx.logger.error("update_templates [end], error {}", repr(e))
+            res = 9
+        except SigTermException:
```
ndokos, line 152: Do we need to handle SigTermException here?

lib/pbench/server/indexing_tarballs.py
```diff
@@ -0,0 +1,492 @@
...
+            user_id=idxctx.getuid(),
+            hostname=idxctx.gethostname(),
+            version=VERSION,
+            templates=idxctx.templates,
+        )
+        # We use the "start" report ID as the tracking ID for all indexed
+        # documents.
+        try:
+            tracking_id = report.post_status(tstos(idxctx.time()), "start")
+        except SigTermException:
```
ndokos, line 184: Do we need to handle SigTermException here?

lib/pbench/server/indexing_tarballs.py
```diff
@@ -0,0 +1,492 @@
...
+                        tb_res = 7
+                    except SosreportHostname as e:
+                        idxctx.logger.warning("Bad hostname in sosreport: {}", e)
+                        tb_res = 10
+                    except tarfile.TarError as e:
+                        idxctx.logger.error(
+                            "Can't unpack tar ball into {}: {}", ptb.extracted_root, e
+                        )
+                        tb_res = 11
+                    except SigTermException:
```
ndokos, line 294: Do we need to handle SigTermException here? Although the log message will have to be generated elsewhere (presumbaly in `pbench-index.py` when the exception is handled).

lib/pbench/server/indexing_tarballs.py
```diff
@@ -0,0 +1,492 @@
...
+                            failures,
+                            retries,
+                        )
+                        tb_res = 1 if failures > 0 else 0
+                    try:
+                        ie_len = ie_filepath.stat().st_size
+                    except FileNotFoundError:
+                        # Above operation never made it to actual indexing, ignore.
+                        pass
+                    except SigTermException:
```
ndokos, line 322: Do we need to handle SigTermException here?

lib/pbench/server/indexing_tarballs.py
```diff
@@ -0,0 +1,492 @@
...
+                                idxctx.logger.exception(
+                                    "Unexpected error issuing"
+                                    " report status with errors: {}",
+                                    ie_filepath,
+                                )
+                    finally:
+                        # Unconditionally remove the indexing errors file.
+                        try:
+                            os.remove(ie_filepath)
+                        except SigTermException:
```
ndokos, line 346: Do we need to handle SigTermException here?

lib/pbench/server/indexing_tarballs.py
```diff
@@ -0,0 +1,492 @@
...
+                    idxctx.logger.info(
+                        "Finished{} {} (size {:d})",
+                        "[SIGQUIT]" if sigquit_interrupt[0] else "",
+                        tb,
+                        size,
+                    )
+
+                    if sigquit_interrupt[0]:
+                        break
+            except SigTermException:
```
ndokos, line 417: Do we need to handle SigTermException here?

lib/pbench/server/indexing_tarballs.py
```diff
@@ -0,0 +1,492 @@
...
+                            for line in sorted(efp):
+                                print(line.strip(), file=fp)
+                    if skp > 0:
+                        print("\nSkipped Results\n===============", file=fp)
+                        with skipped.open() as sfp:
+                            for line in sorted(sfp):
+                                print(line.strip(), file=fp)
+                try:
+                    report.post_status(tstos(idxctx.time()), "status", report_fname)
+                except SigTermException:
```
ndokos, line 485: Do we need to handle SigTermException here?

ndokos (changes requested):
I'd like to hear @portante's and @dbutenhof 's opinion on this before doing anything about it.

lib/pbench/server/indexing_tarballs.py
riya-17: yes. As the exceptions are being raised in this file it is expected by them.

lib/pbench/server/indexing_tarballs.py
riya-17: SigTermException are being placed at these locations to handle Signal handling at all places

lib/pbench/server/indexing_tarballs.py
riya-17: Mainly it is required here as we are handling sigquit in this block

lib/pbench/server/indexing_tarballs.py
riya-17: just to be cautious

lib/pbench/server/indexing_tarballs.py
riya-17: Doesn't seem like right here

lib/pbench/server/indexing_tarballs.py
riya-17: yes this is the most external try/except block so to be cautious

lib/pbench/server/indexing_tarballs.py
riya-17: yes, to not get confused with general exception

lib/pbench/server/indexing_tarballs.py
riya-17: yes, to not get confused with general exception

server/bin/pbench-index.py
```diff
@@ -676,17 +246,5 @@ def sigquit_handler(*args):
...
     )
     parsed = parser.parse_args()
-    try:
-        # The SIGTERM handler is established around main() to make it easier
-        # to handle it cleanly once established.  We also make sure both
-        # SIGQUIT and SIGINT are ignored until we are ready to deal with them.
-        signal.signal(signal.SIGTERM, sigterm_handler)
-        signal.signal(signal.SIGINT, signal.SIG_IGN)
-        signal.signal(signal.SIGQUIT, signal.SIG_IGN)
-        status = main(parsed, run_name)
```
portante, line 686: Is there a reason this was moved to the new module?  Don't we still need the `SIGTERM` handling here?

lib/pbench/server/indexing_tarballs.py
portante: Good catch, Nick, I totally missed this subtly.

We should probably keep this exception definition here, and `import` it into the `pbench-index.py` module instead.

lib/pbench/server/indexing_tarballs.py
portante: @ndokos, see the comment on lines 114 and 115.  If there is a better way to word that comment to make this requirement clear, let me know.

lib/pbench/server/indexing_tarballs.py
portante: @ndokos, see the comment on lines 153 and 154. If there is a better way to word that comment to make this requirement clear, let me know.

lib/pbench/server/indexing_tarballs.py
portante: @ndokos, see the comment on lines 185 and 186. If there is a better way to word that comment to make this requirement clear, let me know.

lib/pbench/server/indexing_tarballs.py
portante: @ndokos, see the comment on lines 185 and 186. If there is a better way to word that comment to make this requirement clear, let me know; regardless, we should probably add a command like that here.

lib/pbench/server/indexing_tarballs.py
portante: @ndokos, see the comment on lines 323 and 324. If there is a better way to word that comment to make this requirement clear, let me know.

lib/pbench/server/indexing_tarballs.py
portante: @ndokos, see the comment on lines 347 and 348. If there is a better way to word that comment to make this requirement clear, let me know.

lib/pbench/server/indexing_tarballs.py
portante: @ndokos, see the comment on lines 418 and 419. If there is a better way to word that comment to make this requirement clear, let me know.

lib/pbench/server/indexing_tarballs.py
portante: @ndokos, see the comment on lines 486 and 487. If there is a better way to word that comment to make this requirement clear, let me know.

portante (changes requested):
We need to restore the `SigTermException` handling in the `pbench-index.py` module as before, importing that exception from this module now.

lib/pbench/server/indexing_tarballs.py
riya-17: Oh that, did not cross my mind. yeah we can do that

server/bin/pbench-index.py
riya-17: Function handling of process_tb and collect_tb takes place in main(), if the error is raised in process_tb and collect_tb it needs to be handled in main(). I didn't think we need sigtern_handler here can you please clarify.

portante: This is the one place where we can handle `SIGTERM`.  Not sure how the module restructuring provides for this in the new code.

portante (changes requested):
Needs a rebase and still has not addressed the signal handling issues.

server/bin/pbench-index.py
riya-17: Got the point

riya-17:
> Needs a rebase and still has not addressed the signal handling issues.

Sorry, forgot to push the changes

portante (approved):

lib/pbench/server/indexing_tarballs.py
```diff
@@ -0,0 +1,492 @@
...
+
+                # We use a list object here so that when we close over this
+                # variable in the handler, the list object will be closed over,
+                # but not its contents.
+                sigquit_interrupt = [False]
+
+                def sigquit_handler(*args):
+                    sigquit_interrupt[0] = True
+
+                signal.signal(signal.SIGQUIT, sigquit_handler)
```
dbutenhof, line 221: Are we not disabling this signal handler anywhere? I don't see it. It probably doesn't really matter, as it only sets a flag and once we've seen the flag we'll never look again, but it would be cleaner (and certainly more modular) to disable the handler before we leave the `process_tb` scope.

portante: Good catch, seems like a gap left from the previous work.  It should be set to `SIG_IGN` by adding a `try/finally` block around the `for` loop starting on line 223.

portante (changes requested):

portante (changes requested):
Needs a rebase on the renamed branch `main`, see https://github.com/distributed-system-analysis/pbench/discussions/2080#discussioncomment-318728.

lib/pbench/server/indexing_tarballs.py
riya-17: ok will update that

portante (approved):

ndokos (dismissed):

ndokos:
Needs rebasing.

dbutenhof (dismissed):

portante (changes requested):
Needs a rebase, but once rebased, we can merge.

portante (approved):

dbutenhof (approved):

PR #1944, opened 2020-11-03
Fix warning while running unit tests

zulcss:
+ ./prometheus-metrics-datalog:108: SyntaxWarning: "is" with a literal. Did you mean "=="?
+   if pid is 0:


dbutenhof (approved):

portante (changes requested):
But that will also cause `black` and `flake8` failures.  I have a fix for that in #1936.

portante (approved):
Nevermind, seems the linting passes...

PR #1946, opened 2020-11-04
Address some undesirable tox behaviors

portante:
The first item is we now use `/tmp/${USER}/tox` as the working directory (`{toxworkdir}`) to avoid touching the local source tree.  Along with this change, we also tell `black` to store its cache in the `lint` environment via the `XDG_CACHE_HOME` environment variable.

Further, we pass along the `PY_COLORS`, `NO_COLORS`, and `TERM` environment variables so that `tox` runs are well behaved in non-interactive terminal environments.

Still further, we remove the `envars` section, moving those into the proper `setenv` section.

Finally, we address some indentation issues in `tox.ini` itself in a separate commit.

PR #1947, opened 2020-11-04
Fix agent side `test_move_results`

portante:
This PR is dependent on PR #1946.  The first 3 commits are from that PR.

### DIFF
--- a/lib/pbench/cli/agent/commands/results/__init__.py
+++ b/lib/pbench/cli/agent/commands/results/__init__.py
@@ -1,11 +1,12 @@
 import os
-import sys
 import requests
+import shutil
+import sys
 import tempfile
 
-from pbench.common.logger import get_pbench_logger
 from pbench.agent import PbenchAgentConfig
 from pbench.agent.results import MakeResultTb, CopyResultTb
+from pbench.common.logger import get_pbench_logger
 
 
 def move_results(ctx, _user, _prefix, _show_server):
@@ -47,9 +48,6 @@ def move_results(ctx, _user, _prefix, _show_server):
         )
         sys.exit(1)
 
-    runs_copied = 0
-    failures = 0
-
     try:
         temp_dir = tempfile.mkdtemp(
             dir=config.pbench_tmp, prefix="pbench-move-results."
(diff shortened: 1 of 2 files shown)


portante (commented):
This PR will be ready for review when #1946 is merged.  We'll want to get this reviewed ahead of #1948.

lib/pbench/cli/agent/commands/results/__init__.py
```diff
@@ -58,42 +56,42 @@ def move_results(ctx, _user, _prefix, _show_server):
         logger.error("Failed to create temporary directory")
         sys.exit(1)
 
-    dirs = [
-        _dir
-        for _dir in next(os.walk(config.pbench_run))[1]
-        if not _dir.startswith("tools-") and not _dir.startswith("tmp")
-    ]
+    runs_copied = 0
+    failures = 0
```
dbutenhof, line 60: Are you even using `failures` anymore? You dropped the `crt.copy_result_tb()` return, which used to trigger incrementing the value. Now we're just ignoring failure to copy?

lib/pbench/test/unit/agent/task/test_move_results.py
```diff
@@ -1,15 +1,43 @@
...
+
+[run]
+controller = agent.example.com
+start_run = YYYY-MM-DDTHH:MM:SS.000000000
+end_run = YYYY-MM-DDTHH:MM:SS.000000000
+
+[iterations/1]
+iteration_name = 1
+user_script = sleep
+"""
```
dbutenhof, line 34: This could be much better formatted, and maybe commented. The leading `"""` is not immediately obvious, and putting the first line of the config file on the same line doesn't help readability!

portante: > This could be much better formatted, and maybe commented.

How shall we format this differently?  Can we propose a formatting that does not differ from what the agent typically generates?

> The leading `"""` is not immediately obvious, and putting the first line of the config file on the same line doesn't help readability!

This is a template that is attempting to mimic the contents of an actual `metadata.log` file as constructed by the agent.  To that end, the first line of the config file has to be next to the `"""` to avoid an extra new line in the file, which is not how the agent constructs the file.  Further, the agent doesn't put comments in the file, and while that shouldn't be a problem, that is different from what the agent code does (though, maybe the agent should emit comments to help be understand the section contents).  The trailing `"""` has to be on a separate line so that the file will have a trailing newline.

I'll fix the `script`, `config`, and `date` fields to be consistent with the `run_name` variable.

lib/pbench/cli/agent/commands/results/__init__.py
portante: Yes, the test uses `failures`, the main code that calls this method is broken in many other ways, but I was not going to try to fix that brokenness as well.

And we don't ignore copy failures.  All of this code calls `sys.exit()`, which is a HUGE mistake, and really my mistake because I code reviewed this and failed to flag this as an issue.  So rather than try to fix all that, I was trying to focus on just fixing the test to actually take action.

The return value from `crt.copy_result_tb` is always None.  So you could never run a successful test here.

And the return value from `mrt.make_result_tb` could only be a `string`, which means the `if result_tb_name` was always `True`.

We'll have to tackle the `sys.exit()` issue, which is what is leading to this mess in another PR.

lib/pbench/test/unit/agent/task/test_move_results.py
dbutenhof: OK, I'll grant you the leading newline.

But I wasn't talking about a comment inside the config file quote, but rather before the assignment.

npalaska: Can this be added to unit test fixtures directory,? We do store the log files that we use for the unit tests in `fixtures` directory like `lib/pbench/test/unit/agent/fixtures/move_results/logs/metadata.log`

ndokos (approved):

lib/pbench/cli/agent/commands/results/__init__.py
npalaska: I can file an issue for better handling `sys.exit()`  to track if you have not filed yet

lib/pbench/test/unit/agent/task/test_move_results.py
portante: > Can this be added to unit test fixtures directory,? We do store the log files that we use for the unit tests in `fixtures` directory like `lib/pbench/test/unit/agent/fixtures/move_results/logs/metadata.log`

Yes, another avenue to take would be to make a fixture of a mocked out pbench run directory with just a `metadata.log` file in it.  Then the test would just pass it to the `move_results` method.  I opted not to do that thinking that we'll want to write more tests to verify the behavior of `move_results`, where this methods allows us to taylor the contents a bit.

npalaska (dismissed):

ndokos (dismissed):

dbutenhof (dismissed):

npalaska (dismissed):

ndokos (dismissed):

lib/pbench/cli/agent/commands/results/__init__.py
npalaska: filed and issue https://github.com/distributed-system-analysis/pbench/issues/1955

dbutenhof (dismissed):

dbutenhof (approved):

ndokos (approved):

portante:
Merging without a successful TravisCI since the Jenkins run in #1936 succeeded, and that PR depends on this PR.

PR #1948, opened 2020-11-04
Address common logging between agent and server

portante:
This is a significant re-work of the common logging module shared between the agent and server side code bases.  It addresses testing issues on the server side, moving all of the logging tests (formerly tests `test-26.*`) to `pytest` based tests, and implements `pytest` based tests for the `PbenchConfig` base class, and the `common/logger` module.

Below is an itemized list, of sorts, covering related changes made and why:

 - Added a `[logging]` section to the `pbench-agent-default.cfg` so that we can begin to use the common logging infrastructure properly throughout the agent Python 3 code base

 - Added the new `log_dir` field to the `PbenchConfig` base class so that the logger module can use it; we default the value to `None` if not present in the configuration file so that the Agent and Server "config" sub-classes know the value was not provided in the configuration file.

   The `PbenchAgentConfig` class sets this new `log_dir` field to the parent directory of the `pbench_log` configuration value to avoid having to change the configuration files on the aside, helping compatibility.

   The `PbenchServerConfig` class sets this new `log_dir` field to the `pbench-logs-dir` configuration file option if `log_dir` is not set.  However, if `log_dir` IS set, it overrides the value of the attribute `LOGSDIR`.

   All the Python 3 code on the server side has been changed to only use the `log_dir` attribute, dropping the use of `LOGSDIR`.  The `LOGSDIR` attribute is still present to provide it to the `bash` scripts which reference the `${LOGSDIR}` environment variable.  This also includes `server/bin/unittests` which was setting `LOGSDIR` but it was never actually being used.

 - We have removed the `pbench-logs-dir` attribute from the example `pbench-server.cfg` file, since the value is properly defaulted in the `pbench-server-default.cfg` file.

 - The Agent and Server log handling is a bit different, with the server side using per-caller directories, and the agent sharing on directory for all callers.  To enable this behavior we add the `log_using_caller_directory` field to the `PbenchConfig` object so that the `PbenchServerConfig` sub-class can request the behavior.

 - To further the removal of "unit test knowledge" from the main code base, we add an undocumented `log_format` field used by the unit tests to ensure timestamps and other unwanted values don't appear in the log stream.

 - Corrected a use of `pathlib` in `lib/pbench/agent/results.py`

 - We remove the unused `LOG` setting, and the `rest_log` configuration file value, since the app server uses the default logging setup for the server side code which is `devlog` (Python 3 `SysLogHandler`).

 - `lib/pbench/test/unit/agent/cli/__init__.py` was removed as it was an empty module and directory

 - We move the `lib/pbench/test/unit/config/pbench-agent.invalid.cfg` test file to the `agent/config` sub-directory, and we removed the unused `lib/pbench/test/unit/config/pbench-agent.cfg` and `lib/pbench/test/unit/config/pbench-server.cfg` test configuration files.

 - For consistency with the non-unit test code, we create pbench logger objects using `pbench` as the caller's name in the unit tests.

This PR is built on top of PR #1947 and #1946, with the first 3 commits coming from those PRs.


portante:
I'll pull this out of draft mode once the two dependent PRs, #1946 and #1947, have been merged.

zulcss (approved):
lgtm

lib/pbench/cli/agent/commands/results/__init__.py
```diff
@@ -58,42 +56,42 @@ def move_results(ctx, _user, _prefix, _show_server):
         logger.error("Failed to create temporary directory")
         sys.exit(1)
 
-    dirs = [
-        _dir
-        for _dir in next(os.walk(config.pbench_run))[1]
-        if not _dir.startswith("tools-") and not _dir.startswith("tmp")
-    ]
+    runs_copied = 0
+    failures = 0
```
dbutenhof, line 60: For the third time, I see no way here for `failures` to be incremented since you dropped checking the return value from `copy_result_tb()` ... what's the intent here? If we're really going to stop caring, let's not return it in the tuple rather than just saying `0`.

lib/pbench/common/logger.py
```diff
@@ -108,13 +112,14 @@ def format(self, record):
         return msg
 
 
-# Used to track the individual FileHandler's created by callers of
+# Used to track the individual logging Handler's created by callers of
```
dbutenhof, line 115: I don't think "handler" should be capitalized here.

lib/pbench/test/unit/agent/task/test_move_results.py
```diff
@@ -1,15 +1,43 @@
...
+import logging
+import os
 import responses
 
+from pathlib import Path
 from pbench.cli.agent.commands.results import move_results
 from pbench.test.unit.agent.task.common import MockDatetime
 
 
+mdlog_tmpl = """[pbench]
```
dbutenhof, line 11: This section would be a lot easier to "grok" at a glance with a comment; and maybe even just a line break after the opening `"""` to make the bracketing more obvious.

lib/pbench/test/unit/common/test_config.py
```diff
@@ -0,0 +1,68 @@
...
+        assert config.TZ == "UTC", f"Unexpected TZ value, {config.TZ!r}"
+        assert (
+            config.log_fmt is None
+        ), f"Unexpected log format value, {config.log_fmt!r}"
+        assert (
+            config.default_logging_level == "INFO"
+        ), f"Unexpected default logging level, {config.default_logging_level!r}"
+        assert (
+            config.log_using_caller_directory is False
+        ), f"Unexpected 'log using coller directory' boolean, {config.log_using_caller_directory!r}"
```
dbutenhof, line 27: Maybe "log using caller directory"??? ;-)

lib/pbench/test/unit/common/test_config.py
```diff
@@ -0,0 +1,68 @@
...
+        with pytest.raises(AttributeError):
+            print(f"{config.logger_host!r}")
+        with pytest.raises(AttributeError):
+            print(f"{config.logger_port!r}")
+        assert "42" == config.get(
+            "other", "foobar"
+        ), "Failed to fetch 'foobar' from 'DEFAULT' section"
+        assert "43" == config.get(
+            "other", "barfoo"
+        ), "Failed to fethc 'barfoo' from 'other' section"
```
dbutenhof, line 41: And our pal "fethc" again (darn, I had to try three times to misspell that...)

lib/pbench/common/logger.py
```diff
@@ -129,29 +134,32 @@ def get_pbench_logger(caller, config):
...
+            else:
+                log_dir = Path(config.log_dir)
             try:
-                logdir.mkdir()
+                log_dir.mkdir()
             except FileExistsError:
                 # directory already exists, ignore
                 pass
-            handler = logging.FileHandler(logdir / f"{caller}.log")
+            handler = logging.FileHandler(log_dir / f"{caller}.log")
```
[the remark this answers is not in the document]
ndokos: NM - I see now...

lib/pbench/server/report.py
```diff
@@ -180,7 +180,7 @@ def post_status(self, timestamp, doctype, file_to_index=None):
                     fname = "report-status-payload.{}.{}.xz".format(
                         self.name, timestamp_noutc
                     )
-                    fpath = os.path.join(self.config.LOGSDIR, self.name, fname)
+                    fpath = self.config.log_dir / self.name / fname
```
ndokos, line 183: Where did `self.config.log_dir` become  a Path object? I thought it's a string.

ndokos (changes requested):

lib/pbench/cli/agent/commands/results/__init__.py
portante: See comments in #1947.

lib/pbench/test/unit/agent/task/test_move_results.py
portante: See comments in #1947.

lib/pbench/server/report.py
portante: Yes, you are correct, the config objects `log_dir` field is always a string.  I've fixed the code to correct the inconsistent use.

dbutenhof (dismissed):

lib/pbench/server/report.py
```diff
@@ -180,7 +182,7 @@ def post_status(self, timestamp, doctype, file_to_index=None):
                     fname = "report-status-payload.{}.{}.xz".format(
                         self.name, timestamp_noutc
                     )
-                    fpath = os.path.join(self.config.LOGSDIR, self.name, fname)
+                    fpath = Path(self.config.log_dir) / self.name / fname
```
ndokos, line 185: I think I had a comment somewhere earlier that `self.config.log_dir` is a string - if I misled you into  "correcting" this, I apologize, but I now think it *is* a Path (l.140ff in pbench/common/logger.py). But a Path of a  Path is a no-op, so it does not hurt.

ndokos (approved):

lib/pbench/server/report.py
portante: I had thought you pointed out a real problem.

In all the places where `config.log_dir` is used and assigned, it should be a `string`.  So the code above should be correct: `assert isinstance(self.config.log_dir, string)`.

I'll go review the logger to ensure it is not assigning a `Path` object to `self.config.log_dir`.

portante: I am not seeing where we assign a `PbenchConfig` object to have a `log_dir` field with a `Path` value.  Let me know if you find otherwise.

dbutenhof (approved):

dbutenhof (approved):

lib/pbench/server/report.py
ndokos: Yes, at the place I pointed to, only a local variable `log_dir` is Path-ified. It *was* a real problem :-)

ndokos (approved):

lib/pbench/test/unit/agent/task/test_copy_result_tb.py
```diff
@@ -16,7 +16,7 @@ class TestCopyResults:
     def config_and_logger(self):
         # Setup the configuration and logger
         self.config = PbenchAgentConfig(os.environ["_PBENCH_AGENT_CONFIG"])
-        self.logger = get_pbench_logger("unittest", self.config)
+        self.logger = get_pbench_logger("pbench", self.config)
```
npalaska, line 19: same as above

lib/pbench/test/unit/common/test_config.py
```diff
@@ -0,0 +1,68 @@
...
+_config_path_prefix = Path("lib/pbench/test/unit/common/config")
+
+
+class TestPbenchConfig:
+    def test_empty_config(self):
+        config = PbenchConfig(_config_path_prefix / "pbench.cfg")
+        assert config.TZ == "UTC", f"Unexpected TZ value, {config.TZ!r}"
+        assert (
+            config.log_fmt is None
+        ), f"Unexpected log format value, {config.log_fmt!r}"
```
npalaska, line 21: Question, whats `!r` does?

lib/pbench/test/unit/common/test_logging.py
```diff
@@ -0,0 +1,140 @@
...
+            suffix=".d", prefix="pbench-common-unit-tests."
+        ) as TMP:
+            assert Path(TMP).exists()
+            self.config.log_dir = str(Path(TMP) / "log-dir")
+            assert not Path(
+                self.config.log_dir
+            ).exists(), (
+                f"Test logging directory unexpectedly exists, {self.config.log_dir}"
+            )
+            Path(self.config.log_dir).mkdir()
```
npalaska, line 73: Not sure but `exist_ok=False` as a default parameter to `mkdir()` would raise `FileExistsError` if target dir for some reason exists so we dont have to check it explicitly above.

npalaska (approved):
Just few comments, but overall looks good

lib/pbench/test/unit/common/test_config.py
dbutenhof: It uses Python's `repr()` instead of `str()`.

lib/pbench/cli/agent/commands/results/__init__.py
```diff
@@ -1,16 +1,17 @@
...
-from pbench.common.logger import get_pbench_logger
 from pbench.agent import PbenchAgentConfig
 from pbench.agent.results import MakeResultTb, CopyResultTb
+from pbench.common.logger import get_pbench_logger
 
 
 def move_results(ctx, _user, _prefix, _show_server):
     config = PbenchAgentConfig(ctx["args"]["config"])
-    logger = get_pbench_logger("pbench-move-results", config)
+    logger = get_pbench_logger("pbench", config)
```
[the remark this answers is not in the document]
portante: Isn't this the agent side code, where we don't use separate log files per caller, but one "`pbench.log`" file (by default in `/var/lib/pbench-agent`)?

lib/pbench/test/unit/common/test_logging.py
portante: Since this is a test, the log directory should not be there.  If it is, then it should raise an error.

lib/pbench/cli/agent/commands/results/__init__.py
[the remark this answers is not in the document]
npalaska: Ahh yes

portante:
Merging without a successful TravisCI since the Jenkins run in #1936 succeeded, and that PR depends on this PR.

PR #1956, opened 2020-11-05
PCP Update

Maxusmusti:
Finally got everything into a good enough state that it's worth putting up a draft.
NOTE: there is still a lot of work left for me to do here, like...
 - Cleaner integration with tool metadata
 - Further abstracting and optimizing solution to work within persistent tool system rather than separately
 - Fix unit test compatibility

I will continue updating this draft PR regularly

ANOTHER NOTE: The first commit in this draft needs credit to be given to Ashwin @not4win as a lot of the code was from his original work, but I'm not sure how to do that...

### DIFF
--- a/agent/tool-scripts/meta.json
+++ b/agent/tool-scripts/meta.json
@@ -43,6 +43,7 @@
 
 	"persistent":{
 		"node-exporter": {"collector": "prometheus", "port": "9100"},
-		"dcgm": {"collector": "prometheus", "port": "8000"}
+		"dcgm": {"collector": "prometheus", "port": "8000"},
+		"pcptool": {"collector": "pcp", "port": "44321"}
 	}
 }
--- a/agent/util-scripts/pbench-tool-meister-start
+++ b/agent/util-scripts/pbench-tool-meister-start
@@ -405,12 +405,8 @@ def main(argv):
 
     # 3. Start the tool-data-sink process
     #   - leave a PID file for the tool data sink process
-    host_tools_dict = dict()
-    for host in tool_group.hostnames.keys():
-        host_tools_dict[host] = tool_group.get_tools(host)
-
     tds_param_key = "tds-{}".format(group)
-    tds = dict(channel=channel, benchmark_run_dir=benchmark_run_dir, group=group, host_tools_dict=host_tools_dict)
+    tds = dict(channel=channel, benchmark_run_dir=benchmark_run_dir, group=group)
     try:
         redis_server.set(tds_param_key, json.dumps(tds, sort_keys=True))
     except Exception:
--- a/agent/util-scripts/pbench-tool-meister-start
+++ b/agent/util-scripts/pbench-tool-meister-start
@@ -405,8 +405,12 @@ def main(argv):
 
     # 3. Start the tool-data-sink process
     #   - leave a PID file for the tool data sink process
+    host_tools_dict = dict()
+    for host in tool_group.hostnames.keys():
+        host_tools_dict[host] = tool_group.get_tools(host)
+
     tds_param_key = "tds-{}".format(group)
-    tds = dict(channel=channel, benchmark_run_dir=benchmark_run_dir, group=group)
+    tds = dict(channel=channel, benchmark_run_dir=benchmark_run_dir, group=group, host_tools_dict=host_tools_dict)
     try:
         redis_server.set(tds_param_key, json.dumps(tds, sort_keys=True))
     except Exception:
--- a/lib/pbench/agent/tool_data_sink.py
+++ b/lib/pbench/agent/tool_data_sink.py
@@ -36,6 +36,9 @@ from bottle import Bottle, ServerAdapter, request, abort
 import pbench.agent.toolmetadata as toolmetadata
 from pbench.agent import PbenchAgentConfig
 
+from pcp import pmapi
+from pcp.pmapi import pmContext as PCP
+import cpmapi as api
 
 # Read in 64 KB chunks off the wire for HTTP PUT requests.
 _BUFFER_SIZE = 65536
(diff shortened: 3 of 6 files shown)


agent/tool-scripts/meta.json
```diff
@@ -43,6 +43,7 @@
 
 	"persistent":{
 		"node-exporter": {"collector": "prometheus", "port": "9100"},
-		"dcgm": {"collector": "prometheus", "port": "8000"}
+		"dcgm": {"collector": "prometheus", "port": "8000"},
+		"pcptool": {"collector": "pcp", "port": "44321"}
```
portante, line 47: Let's drop the existing `pcp` tool entirely, and just replace it with this one.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -395,8 +395,12 @@ def main(argv):
 
     # 3. Start the tool-data-sink process
     #   - leave a PID file for the tool data sink process
+    host_tools_dict = dict()
+    for host in tool_group.hostnames.keys():
+        host_tools_dict[host] = tool_group.get_tools(host)
+
     tds_param_key = "tds-{}".format(group)
-    tds = dict(channel=channel, benchmark_run_dir=benchmark_run_dir, group=group)
+    tds = dict(channel=channel, benchmark_run_dir=benchmark_run_dir, group=group, host_tools_dict=host_tools_dict)
```
portante, line 403: Why do we need this?

lib/pbench/agent/tool_meister.py
```diff
@@ -61,6 +61,42 @@
...
+from ansible.playbook.play import Play
+from ansible.plugins.callback import CallbackBase
+from ansible.vars.manager import VariableManager
+from ansible import context
+
+# ansible constants
+ANSIBLE_CONF = "/etc/ansible/ansible.cfg"
+ANSIBLE_ROLES_PATH = "/etc/ansible/roles"
+PLAYBOOK_PATH = ""
+TOOLS_PMDA_MAPPING = {
```
portante, line 78: Can we consider making this a class variable of `PCPTools`?

agent/util-scripts/pbench-tool-meister-start
Maxusmusti: We don't, it was part of the original work, I plan on removing it today.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -186,6 +190,408 @@ def terminate(self):
...
+            self.logger.info("Reading the pcp mapping file")
+            self.logger.debug("Reading file:%s", json_fn)
+            with open(os.path.join(json_fn)) as json_file:
+                self.mapping = json.loads(json_file.read())
+        except Exception as e:
+            self.logger.critical(
+                "Not able to load the mapping file\n:%s", e, exc_info=True
+            )
+
+    def start_logging(self):
```
dbutenhof, line 257: `start_logger` and `start_logging` are a bit too close for comfort given they're doing radically different things. Can we improve the names here?

lib/pbench/agent/tool_data_sink.py
```diff
@@ -186,6 +190,408 @@ def terminate(self):
...
+        for idx, host in enumerate(hosts):
+            res = ""
+            try:
+                self.logger.debug("checking connection on host:%s", host)
+                res = pmapi.pmContext(api.PM_CONTEXT_HOST, host + ":44321")
+            except Exception:
+                res = None
+
+            if res is not None:
+                self.logger.info("Port 44321 is open in:{}".format(host))
```
dbutenhof, line 315: Since we're probably always going to be logging `info` level at least, this isn't a big deal; but it's odd that you're pre-formatting the message string here rather than passing arguments and letting the logger format when/if necessary.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -186,6 +190,408 @@ def terminate(self):
...
+                self.hosts.append(
+                    host
+                )  # logging only on those hosts with port 44321 active
+            else:
+                self.logger.error("Port 44321 is not open in:{}".format(host))
+                self.logger.error("Ignoring host:%s", host)
```
dbutenhof, line 321: And here we have pre-formatted and post-formatted messages side by side?

lib/pbench/agent/tool_data_sink.py
```diff
@@ -186,6 +190,408 @@ def terminate(self):
...
+
+    def build_pmlog_configs(self, hosts):
+        """ Build pmlog config files for each host. Returns true/false depending on success or failure in building files """
```
dbutenhof, line 324: The wrapping of the docstring could be improved!

lib/pbench/agent/tool_data_sink.py
```diff
@@ -186,6 +190,408 @@ def terminate(self):
...
+        self.logger.info("building pmlogger config files")
+
+        for index, host in enumerate(hosts):
+            # make a folder with name self.benchmark_run_dir/temp/host
```
dbutenhof, line 328: According to the code below, it's `.temp`. It also looks a bit odd having the variable name inline like that. Maybe `with name {self.benchmark_run_dir}/temp/host` to set it off?

lib/pbench/agent/tool_data_sink.py
```diff
@@ -186,6 +190,408 @@ def terminate(self):
...
+            try:
+                os.mkdir(".temp")
+                self.logger.debug(
+                    "Created folder :{}".format(
+                        os.path.join(self.benchmark_run_dir, ".temp")
+                    )
+                )
```
dbutenhof, line 335: Hopefully we _won't_ routinely log `DEBUG` level, so these should definitely all be post-formatted, not pre-formatted.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -186,6 +190,408 @@ def terminate(self):
...
+                stdout, stderr = out.communicate()
+                self.logger.debug("stdout output:%s", stdout.decode())
+
+                if (
+                    stdout.decode().strip()
+                    != f'Creating config file "{host}.config" using default settings ...\n.'
+                ):
+                    # pmlogger failed to stop. Log it
+                    self.logger.error("not able to build pmlogconf file")
```
dbutenhof, line 396: We can't check the exit status? Judging success by matching output strings is always risky.

(Not to mention that this module has now adding template string formatting to %s formatting and `.format` formatting, making the full sweep of Python string formatting styles and making the whole thing look just a bit more messy. I'm assuming this is an inherited issue, but it'd be nice to clean it up.)

lib/pbench/agent/tool_data_sink.py
```diff
@@ -186,6 +190,408 @@ def terminate(self):
...
+                    shell=True,
+                )
+                stdout, stderr = out.communicate()
+                self.logger.debug("stdout output:%s", stdout.decode())
+
+                if (
+                    stdout.decode().strip()
+                    != f'Creating config file "{host}.config" using default settings ...\n.'
+                ):
+                    # pmlogger failed to stop. Log it
```
dbutenhof, line 395: "Failed to stop"? Were we trying to stop it???

lib/pbench/agent/tool_data_sink.py
```diff
@@ -186,6 +190,408 @@ def terminate(self):
...
+                    self.logger.error("not able to build pmlogconf file")
+                else:
+                    self.logger.info("built config file for host:{}".format(host))
+
+                self.pmlog_config_files.append("{}.config".format(host))
+            except Exception as e:
+                # log exception as we are not able to stop logger
```
dbutenhof, line 402: Again we failed to stop it when trying to write a config file. I don't really know how `pmlogconf` works, although the man page I found didn't suggest it does any more than writing a formatted config file.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -186,6 +190,408 @@ def terminate(self):
...
+                self.logger.error(
+                    "not able to create a logger conf file\n", e, exc_info=True
```
dbutenhof, line 404: `exc_info=True` causes the logger to add exception information, so I think the `e` argument is unused since there's no substitution placeholder in the message string.

This pattern is repeated several times below.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -186,6 +190,408 @@ def terminate(self):
...
+                res = None
+
+            if res is not None:
+                self.logger.info("Port 44321 is open in:{}".format(host))
+                self.hosts.append(
+                    host
+                )  # logging only on those hosts with port 44321 active
+            else:
+                self.logger.error("Port 44321 is not open in:{}".format(host))
+                self.logger.error("Ignoring host:%s", host)
```
dbutenhof, line 543: This shouldn't be repeated twice. Why not promote it to `BaseCollector` (or a new intermediary PCP class) so they can share?

lib/pbench/agent/tool_meister.py
```diff
@@ -73,6 +112,245 @@ class ToolException(Exception):
...
+            "output from command:%s is:%s\n and stderr:%s",
+            command,
+            stdout.decode(),
+            stderr,
+        )
+
+        if output_err is not None:
+            self.logger.error("not able to download role. error:%s", output_err)
+
+    def start(self):
```
dbutenhof, line 235: There seems to be a lot of duplication between `start` and `stop` that could probably be factored out into a common method to simplify maintenance.

dbutenhof (commented):
Stacking tool meister on top of Ansible just seems... "off". This sort of cross-node orchestration of tool start/stop seems to be exactly what tool meister was designed to do in the first place. Unfortunate if we need to stack like this to manage PCP.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -186,6 +190,408 @@ def terminate(self):
...
+            self.logger.critical(
+                "Not able to load the mapping file\n:%s", e, exc_info=True
+            )
+
+    def start_logging(self):
+        """ invoke pmlogger and start logging using pmlogger_check -c """
+        # code to start logging
+        self.logger.info("starting logger")
+
+        command = PCP.pmGetConfig("PCP_BINADM_DIR") + "/pmlogger_check -c {}".format(
```
portante, line 262: As disked verbally today, let's work on containerizing the `pmlogger` much like the way we handle Prometheus.

lib/pbench/agent/tool_meister.py
```diff
@@ -73,6 +112,245 @@ class ToolException(Exception):
...
+        )
+
+        # find if the required role exists or not
+        for line in output:
+            if "performancecopilot.metrics.pcp" in line:
+                # if it exists, return. Dont have to proceed
+                return
+
+        # given role does not exist. We have to download it from ansible galaxy
+        command = "ansible-galaxy collection install performancecopilot.metrics"
```
portante, line 195: As we talked about today verbally, let's drop the use of Ansible to install and start PCP on local/remote systems.  Philosophically, the pbench-agent should not be in the business of install software on hosts, or changing the host configuration to suite its needs.  Instead let's build a container for `pmcd` which encapsulates what we need, and do the same for `node-exporter` as well.

portante (changes requested):

portante:
Replaced by #1986.

PR #1959, opened 2020-11-06
Rework pass-thru API implementation a bit

portante:
This all started because logging was not quite set up correctly for the server side API tests.  We use the `caplog` fixture to figure out what is being logged and then added `file` based logging by default so that logs would not leak through to the local system logs via the default `devlog` server configuration during a test run.

In getting the tests to work, we adjust how exceptions are handled so that emitted messages are not so broad.  We use 5xx status codes properly and quiet the logging where possible.

When an internal error is hit, we only report "INTERNAL ERROR" in the response, but log the details.

### DIFF
--- a/lib/pbench/server/api/__init__.py
+++ b/lib/pbench/server/api/__init__.py
@@ -147,39 +147,41 @@ class GraphQL(Resource, CorsOptionsRequest):
 
     def post(self):
         global app
-        graphQL = f"{app.config_graphql.get('host')}:{app.config_graphql.get('port')}"
+        graphQL = (
+            f"http://{app.config_graphql.get('host')}:{app.config_graphql.get('port')}"
+        )
         json_data = request.get_json(silent=True)
 
         if not json_data:
-            app.logger.error("GraphQL: Invalid json object %s", request.url)
+            app.logger.warning("GraphQL: Invalid json object {}", request.url)
             abort(400, message="GraphQL: Invalid json object in request")
 
         try:
             # query GraphQL
-            response = requests.post(graphQL, json=json_data)
-
-            # construct response object
-            response = make_response(response.text)
-            response = _corsify_actual_response(response)
-
+            gql_response = requests.post(graphQL, json=json_data)
         except requests.exceptions.ConnectionError:
             app.logger.exception("Connection refused during the GraphQL post request")
-            abort(500, message="Network problem, could not post to GraphQL Endpoint")
+            abort(502, message="Network problem, could not post to GraphQL Endpoint")
         except requests.exceptions.Timeout:
             app.logger.exception("Connection timed out during the GraphQL post request")
             abort(
-                500, message="Connection timed out, could not post to GraphQL Endpoint"
-            )
-        except requests.exceptions.InvalidURL:
-            app.logger.exception("Invalid url during the GraphQL post request")
-            abort(
-                500, message="Invalid GraphQL url, could not complete the post request"
+                504, message="Connection timed out, could not post to GraphQL Endpoint"
             )
         except Exception:
             app.logger.exception("Exception occurred during the GraphQL post request")
-            abort(500, message="Could not post to GraphQL endpoint")
+            abort(500, message="INTERNAL ERROR")
 
-        response.status_code = 200
+        try:
+            # construct response object
+            raw_response = make_response(gql_response.text)
+            response = _corsify_actual_response(raw_response)
+        except Exception:
+            app.logger.exception("Exception occurred GraphQL response construction")
+            abort(
+                500, message="INTERNAL ERROR",
+            )
+
+        response.status_code = gql_response.status_code
         return response
 
 
@@ -193,16 +195,18 @@ class Elasticsearch(Resource, CorsOptionsRequest):
 
     def post(self):
         global app
-        elasticsearch = f"{app.config_elasticsearch.get('host')}:{app.config_elasticsearch.get('port')}"
+        elasticsearch = f"http://{app.config_elasticsearch.get('host')}:{app.config_elasticsearch.get('port')}"
         json_data = request.get_json(silent=True)
         if not json_data:
-            app.logger.error(
-                "Elasticsearch: Invalid json object. Query: %s", request.url
+            app.logger.warning(
+                "Elasticsearch: Invalid json object. Query: {}", request.url
             )
             abort(400, message="Elasticsearch: Invalid json object in request")
 
         if not json_data["indices"]:
-            app.logger.error("Elasticsearch: Missing indices path in the post request")
+            app.logger.warning(
+                "Elasticsearch: Missing indices path in the post request"
+            )
             abort(400, message="Missing indices path in the Elasticsearch request")
 
         try:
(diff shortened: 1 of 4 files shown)


lib/pbench/server/api/__init__.py
```diff
@@ -147,39 +147,41 @@ def __init__(self):
...
         global app
-        graphQL = f"{app.config_graphql.get('host')}:{app.config_graphql.get('port')}"
+        graphQL = (
+            f"http://{app.config_graphql.get('host')}:{app.config_graphql.get('port')}"
+        )
         json_data = request.get_json(silent=True)
 
         if not json_data:
-            app.logger.error("GraphQL: Invalid json object %s", request.url)
+            app.logger.warning("GraphQL: Invalid json object {}", request.url)
```
dbutenhof, line 156: I always think of `warning` and `error` as representing __our__ problem, in which case using them to indicate failures of external servers or user input is misleading. I sometimes wish there was a good additional category for these, but I usually end up going with `info` so it's clear that we saw something "interesting" but it's not our problem to fix. (Though of course it might indicate a problem we can avoid or mitigate or detect more effectively... presuming we include enough context.)

lib/pbench/server/api/__init__.py
```diff
@@ -147,39 +147,41 @@ def __init__(self):
...
-
-            # construct response object
-            response = make_response(response.text)
-            response = _corsify_actual_response(response)
-
+            gql_response = requests.post(graphQL, json=json_data)
         except requests.exceptions.ConnectionError:
             app.logger.exception("Connection refused during the GraphQL post request")
-            abort(500, message="Network problem, could not post to GraphQL Endpoint")
+            abort(503, message="Network problem, could not post to GraphQL Endpoint")
```
dbutenhof, line 164: 504 would probably be more correct here as it assigns the root problem to an "upstream" server rather than to us.

lib/pbench/server/api/__init__.py
```diff
@@ -147,39 +147,41 @@ def __init__(self):
...
         except requests.exceptions.Timeout:
             app.logger.exception("Connection timed out during the GraphQL post request")
             abort(
-                500, message="Connection timed out, could not post to GraphQL Endpoint"
-            )
-        except requests.exceptions.InvalidURL:
-            app.logger.exception("Invalid url during the GraphQL post request")
-            abort(
-                500, message="Invalid GraphQL url, could not complete the post request"
+                504, message="Connection timed out, could not post to GraphQL Endpoint"
```
dbutenhof, line 168: 408 would seem better here.

lib/pbench/server/api/__init__.py
```diff
@@ -147,39 +147,41 @@ def __init__(self):
...
             )
         except Exception:
             app.logger.exception("Exception occurred during the GraphQL post request")
-            abort(500, message="Could not post to GraphQL endpoint")
+            abort(500, message="INTERNAL ERROR")
```
dbutenhof, line 172: Yeah, I like the "INTERNAL_ERROR" simplification.

lib/pbench/server/api/__init__.py
```diff
@@ -213,47 +217,48 @@ def post(self):
...
+                503, message="Network problem, could not post to Elasticsearch Endpoint"
             )
         except requests.exceptions.Timeout:
             app.logger.exception(
                 "Connection timed out during the Elasticsearch post request"
             )
             abort(
-                500,
+                504,
                 message="Connection timed out, could not post to Elasticsearch Endpoint",
```
dbutenhof, line 239: Same...

lib/pbench/server/api/__init__.py
portante: Yes, `503` is not right, but `504` is `Gateway Timeout`, perhaps `502`, `Bad Gateway`?  https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

lib/pbench/server/api/__init__.py
portante: But `504` is `Gateway Timeout`, which is what happened here.  The API server is the "gateway", and it timed out trying to talk to the configured Elasticsearch instance.

lib/pbench/server/api/__init__.py
portante: Doesn't a `408` mean the _client_ timed out sending a request to the API server, not the API server being a client that timed out sending a request to GraphQL?  It still seems `504` `Gateway Timeout` is the API telling the client that "your request was okay, we received it, but we timed out talking to the GraphQL instance with which we are configured."

lib/pbench/server/api/__init__.py
portante: I agree with `error` representing _our_ problem, for sure.  We have been using warning to indicate a broad category of problems that include server operations and client behaviors that should be flagged.

Using `WARNING` makes it easier to pull out from all the `INFO`s, but perhaps we should introduce our own "level", `NOTICE`, which is much like the `syslog` `NOTICE` level, were we can reserve `WARNING` and above for our problems, and `NOTICE` could be used for this kind of in between reporting.

If enough of us like the `NOTICE` level, then we'll have to do some additional work to plumb that.  For this commit, I'd like to stick with `WARNING` so that we can get these behavioral changes to land.

lib/pbench/server/api/__init__.py
dbutenhof: Ah; actually, 408 means the _server_ timed out waiting for the _client_ ... which I suppose means it got the start of a request but not all of it.

So, yeah, I guess 504 is probably the "nearly bestest" fit.

lib/pbench/server/api/__init__.py
dbutenhof: That's fine; `NOTICE` sounds like a good idea, but I can deal with your scheme in the meantime.

dbutenhof (approved):

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -1,14 +1,22 @@
...
+import responses
+
 from pathlib import Path
+from werkzeug.utils import secure_filename
+
+
+@pytest.fixture
+def mocked_responses():
+    with responses.RequestsMock() as rsps:
+        yield rsps
```
dbutenhof, line 14: Hmm; I'm using the `requests_mock` fixture from `requests` for the same purpose, and I built up a `query_helper` fixture around it to help with setting up the mock, making the call, and checking results so I could easily pop off good/bad tests.

https://github.com/distributed-system-analysis/pbench/pull/1914/commits/14650fcac67f0225dfe4215306df6aef4fbde0ad#diff-94940849d430f436eeb18310e1a882f5c318d16637dc5097057d9eca1d2172ad (although it's gotten a bit thicker in the more recent commits).

I could probably just swap in your fixture, but I have no idea what the tradeoffs really are.

I had considered adding mock tests to the passthrough APIs, but decided to put that off, luckily, since you're doing it.

dbutenhof (approved):

lib/pbench/test/unit/server/test_requests.py
portante: From what I understand, `responses` handles all that work, but I might be missing some subtlety of that work.

npalaska (approved):
Looks good

PR #1960, opened 2020-11-06
Restore the SELinux context of the tarball after the move

ndokos:
Commit c3ad67c101e394784924f14c5bfceb8d239b636f changed the way we get the tarball from the reception area to the archive, from `cp` to `mv` in order to avoid a data copy. But `cp` set the SELinux context from the parent directory, which made the tar ball accessible through HTTP; `mv` does not do that.

Add an explicit `restorecon` to remedy that and fix test-5.2 for the new output.

### DIFF
--- a/server/bin/pbench-server-prep-shim-002.sh
+++ b/server/bin/pbench-server-prep-shim-002.sh
@@ -187,6 +187,13 @@ while read tbmd5 ;do
         (( nerrs++ ))
         continue
     fi
+    # mv does not restore the SELinux context properly, so we do it by hand
+    restorecon ${dest}/${tb}
+    sts=${?}
+    if [[ ${sts} -ne 0 ]]; then
+        # log it but do not abort
+        log_error "${TS}: Error: \"restorecon ${dest}/${tb}\", status ${sts}" "${status}"
+    fi
 
     # Now that we have successfully moved the tar ball and its .md5 to the
     # destination, we can remove the original .md5 file.
--- a/server/bin/gold/test-5.2.txt
+++ b/server/bin/gold/test-5.2.txt
@@ -1103,6 +1103,10 @@ run-1970-01-01T00:00:42-UTC: pbench-satellite-cleanup ends: Total 3 tarballs cle
 ---- pbench-satellite-local/logs
 --- pbench log file contents
 +++ test-execution.log file contents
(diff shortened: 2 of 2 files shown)


server/bin/pbench-server-prep-shim-002.sh
```diff
@@ -187,6 +187,13 @@ while read tbmd5 ;do
         (( nerrs++ ))
         continue
     fi
+    # mv does not restore the SELinux context properly, so we do it by hand
+    restorecon ${dest}/${tb}
```
portante, line 191: Don't we also have to issue a `restorecon` when the `${dest}` is created?  Perhaps we should just issue one `restorecon -R ${dest}`

portante (changes requested):

riya-17 (dismissed):

server/bin/pbench-server-prep-shim-002.sh
ndokos: The `mkdir` creating the directory does set the context properly (assuming that the parent directory is set correctly which the installation playbook takes care of). So yes, we can, but I don't think it's buying us anything.

ndokos:
I guess I shouldn't have changed test-29.0: ubuntu seems to do things differently from fedora (?), so travis failed.

server/bin/pbench-server-prep-shim-002.sh
portante: If you are confident that you the directory is all set, then let's go with it as it is.  We just need to fix the unit tests, which failed in TravisCI for some odd reason.

portante (changes requested):
Looks good, just need to resolve the unit test failures.

server/bin/pbench-server-prep-shim-002.sh
ndokos: Yes, I'm pretty sure it's OK. I undid the test-29.0 change which is what failed in travis (of course, it's now failing on my machine - sigh...)

ndokos:
I don't know how to kick off travis again - and what (if anything) do we need to do about jenkins? Does the branch settings need tweaking?

portante (approved):

portante:
I merged because Riya had already approved.

PR #1967, opened 2020-11-10
Add more fixtures for agent unit tests.

zulcss:
Add more pytest fixtures to assist in the creation of agent unit
tests the python way.

### DIFF
--- a/lib/pbench/test/functional/conftest.py
+++ b/lib/pbench/test/functional/conftest.py
@@ -1,4 +1,6 @@
+import configparser
 import os
+import shutil
 import subprocess
 
 import pytest
@@ -13,6 +15,9 @@ def setup(request, pytestconfig):
 
 @pytest.helpers.register
 def capture(command):
+    """Capture the command output and return the stdout, stderr,
+    and errorcode.
+    """
     proc = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
     out, err = proc.communicate()
     return out, err, proc.returncode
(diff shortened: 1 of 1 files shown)


dbutenhof (commented):
Seems seriously odd that we have pytest code under `lib/pbench/test/functional/` as well as the neighboring `lib/pbench/test/unit`. They're all unit tests, and the only "functional" unit test appears to be `test_pbench_config.py`

Huh.

lib/pbench/test/functional/conftest.py
```diff
@@ -13,13 +15,48 @@ def setup(request, pytestconfig):
...
+@pytest.fixture
+def pbench_install_dir(tmpdir):
+    """Create an agent install directory for a unittest"""
+    pbench_install_dir = tmpdir / "opt/pbench-agent"
+    os.makedirs(pbench_install_dir)
+    yield pbench_install_dir
+
+
+@pytest.fixture
+def pbench_agent_config(tmpdir, pbench_run, pbench_install_dir, pytestconfig):
```
portante, line 43: We might want to consider reusing what we have, or at least including this in the refactoring: https://github.com/distributed-system-analysis/pbench/blob/master/lib/pbench/test/unit/agent/conftest.py#L8

portante (commented):
Can you comment on how we might be able to re-use what we have?

portante:
> Seems seriously odd that we have pytest code under `lib/pbench/test/functional/` as well as the neighboring `lib/pbench/test/unit`. They're all unit tests, and the only "functional" unit test appears to be `test_pbench_config.py`

Could this be resolved by writing a definition of what we mean by "functional" verses "unit" and then reviewing to see which tests fall into which categories?

PR #1971, opened 2020-11-11
fixes accordion cards functionality

### LINKED ISSUE #1970: The accordion cards under https://distributed-system-analysis.github.io/pbench/start.html#install don't work.
Clicking on the `>` does not open them.

anishaswain:
fixes #1970

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -48,7 +48,7 @@ a{
 }
 
 ul{
-  list-style: inside;
+  list-style: none;
 }
 
 a:hover{
--- a/js/accordion.js
+++ b/js/accordion.js
@@ -1,6 +1,7 @@
 var acc = document.getElementsByClassName("accordion");
+window.addEventListener('load', (event) => {
 	for (i = 0; i < acc.length; i++) {
-		acc[i].addEventListener("click", function() {
+		acc[i].addEventListener("click", function () {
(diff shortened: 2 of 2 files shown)


css/style.css
```diff
@@ -48,7 +48,7 @@ a{
 }
 
 ul{
-  list-style: inside;
+  list-style: none;
```
ndokos, line 51: I presume this is the fix. I don't understand it but it works!

ndokos (approved):
Works for me.

css/style.css
dbutenhof: I was actually kinda wondering that, too; my guess that it was partly setting the click handlers on a page load event; but if the change in the `ul` style is responsible, I have no idea how.

dbutenhof (approved):

css/style.css
anishaswain: @ndokos @dbutenhof I saw another design bug while fixing the accordion issue. The navbar was having list-style dots in them. So fixed them here. I thought of adding a comment but then felt adding a comment for only online in CSS might be inconsistent.

dbutenhof: Thanks, Anisha! That makes sense.

portante (approved):

PR #1975, opened 2020-11-13
Correct the `PYTHONPATH` for Jenkins jobs

portante:
We need to add the `lib/python3.*` and `lib64/python3.*` paths.

### DIFF
--- a/jenkins/python-setup.sh
+++ b/jenkins/python-setup.sh
@@ -1,0 +1,14 @@
+_prefix="${1}"
+
+pip3 install --no-cache-dir --prefix="${_prefix}" -r ${progdir}/agent/requirements.txt -r ${progdir}/server/requirements.txt -r ${progdir}/agent/test-requirements.txt -r ${progdir}/server/test-requirements.txt
+
+_pdir=${_prefix}/bin
+if [[ ":${PATH:-}:" != *":${_pdir}:"* ]]; then
+    export PATH=${_pdir}${PATH:+:${PATH}}
+fi
+
+for _pdir in $(ls -1d ${_prefix}/lib*/python3.*/site-packages 2> /dev/null) $(head -n 1 ${_prefix}/lib*/python3.*/site-packages/pbench.egg-link 2> /dev/null); do
+    if [[ ":${PYTHONPATH:-}:" != *":${_pdir}:"* ]]; then
+        export PYTHONPATH=${_pdir}${PYTHONPATH:+:${PYTHONPATH}}
+    fi
+done
--- a/jenkins/run-pytests
+++ b/jenkins/run-pytests
@@ -3,24 +3,7 @@
 prog="$(basename "${0}")"
 progdir="$(realpath -e $(dirname "${0}")/..)"
 
-_prefix="${1}"
-
-pip3 install --no-cache-dir --prefix="${_prefix}" -r ${progdir}/agent/requirements.txt -r ${progdir}/server/requirements.txt -r ${progdir}/agent/test-requirements.txt -r ${progdir}/server/test-requirements.txt
-
-_pdir=${_prefix}/bin
-if [[ ":${PATH:-}:" != *":${_pdir}:"* ]]; then
-    export PATH=${_pdir}${PATH:+:${PATH}}
-fi
-
-_pdir="$(ls -1d ${_prefix}/lib/python3.*/site-packages)"
-if [[ ":${PYTHONPATH:-}:" != *":${_pdir}:"* ]]; then
-    export PYTHONPATH=${_pdir}${PYTHONPATH:+:${PYTHONPATH}}
-fi
-
-_pdir="$(head -n 1 ${_prefix}/lib/python3.*/site-packages/pbench.egg-link)"
-if [[ ":${PYTHONPATH:-}:" != *":${_pdir}:"* ]]; then
-    export PYTHONPATH=${_pdir}${PYTHONPATH:+:${PYTHONPATH}}
-fi
+source $(dirname "${0}")/python-setup.sh
 
 printf -- "black --check ...\n"
 black --check .
--- a/jenkins/run-unittests
+++ b/jenkins/run-unittests
@@ -3,24 +3,7 @@
 prog="$(basename "${0}")"
 progdir="$(realpath -e $(dirname "${0}")/..)"
 
-_prefix="${1}"
-
-pip3 install --no-cache-dir --prefix="${_prefix}" -r ${progdir}/agent/requirements.txt -r ${progdir}/server/requirements.txt
-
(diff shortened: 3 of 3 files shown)


npalaska (approved):
Question: Do we need to do this because some packages are installed inside lib64 directory by default? Is this a package specific thing like why do they end up in different directory?

portante:
> Do we need to do this because some packages are installed inside lib64 directory by default?

Yes.

> Is this a package specific thing like why do they end up in different directory?

Not sure what you mean, but for a long time now Linux distributions have maintained the notion of 32-bit vs 64-bit binaries.  Typically, when a package is pure Python, or when it provides a 32-bit binary interface, it is installed in the `lib/` hierarchy.  When the package provides a 64-bit binary interface, it is installed in the `lib64/` hierarchy.

Two of the packages that we use today, `psutil` and `werkzeug` provide their modules in the `lib64/` hierarchy.  Other packages, like `SQLAlchemy` for example, also provide their modules in the `lib64/` hierarchy.

So we have to add the both the `lib/` and `lib64/` paths to our `PYTHONPATH` setup.

dbutenhof (approved):
Looks fine; I think I might have chosen a more meaningful name than "common.sh" for the script that sets PATH and PYTHONPATH, but that's a nit.

portante:
> I think I might have chosen a more meaningful name than "common.sh" for the script that sets PATH and PYTHONPATH, but that's a nit.

Naming is hard, @dbutenhof!

What do you think of `python-setup.sh`?

dbutenhof:
> > I think I might have chosen a more meaningful name than "common.sh" for the script that sets PATH and PYTHONPATH, but that's a nit.
> 
> Naming is hard, @dbutenhof!
> 
> What do you think of `python-setup.sh`?

You're setting `PATH` and `PYTHONPATH` for the container script, so something like `set-path.sh` might work?

portante:
> You're setting `PATH` and `PYTHONPATH` for the container script, so something like `set-path.sh` might work?

Isn't it more than just setting up `PATH` and `PYTHONPATH`?  It has to first `pip3 install` all the requirements, and then the `bin`, `lib`, and `lib64` directories have to be added to the appropriate `*PATH` variable.

dbutenhof:
> > You're setting `PATH` and `PYTHONPATH` for the container script, so something like `set-path.sh` might work?
> 
> Isn't it more than just setting up `PATH` and `PYTHONPATH`? It has to first `pip3 install` all the requirements, and then the `bin`, `lib`, and `lib64` directories have to be added to the appropriate `*PATH` variable.

Oops; had to go back and look again. Yeah, somehow I left with the impression you'd moved/expanded the path logic.

So, sure, `python-setup.sh` is fine. Or just not worry about it and merge as it.

dbutenhof (approved):
And ... the new name looks fine. Let's go.

npalaska (approved):

jenkins/python-setup.sh
```diff
@@ -0,0 +1,14 @@
+_prefix="${1}"
+
+pip3 install --no-cache-dir --prefix="${_prefix}" -r ${progdir}/agent/requirements.txt -r ${progdir}/server/requirements.txt -r ${progdir}/agent/test-requirements.txt -r ${progdir}/server/test-requirements.txt
+
+_pdir=${_prefix}/bin
+if [[ ":${PATH:-}:" != *":${_pdir}:"* ]]; then
+    export PATH=${_pdir}${PATH:+:${PATH}}
+fi
+
+for _pdir in $(ls -1d ${_prefix}/lib*/python3.*/site-packages 2> /dev/null) $(head -n 1 ${_prefix}/lib*/python3.*/site-packages/pbench.egg-link 2> /dev/null); do
```
ndokos, line 10: @RobertKrawitz had some comments on this in chat:

```
1) Why are you using ls -1d rather than simply globbing the expression?  If the concern is no files present, you can use shopt -s nullglob.

2) This construct is vulnerable to any spaces in the file names.  A safer way to do this would be

while read -r _pdir ; do
...
done <<< "$(ls -1d ... ; head -n 1 ...)"
```

ndokos (changes requested):
Added @RobertKrawitz's comment for review.

jenkins/python-setup.sh
```diff
@@ -0,0 +1,14 @@
+_prefix="${1}"
+
+pip3 install --no-cache-dir --prefix="${_prefix}" -r ${progdir}/agent/requirements.txt -r ${progdir}/server/requirements.txt -r ${progdir}/agent/test-requirements.txt -r ${progdir}/server/test-requirements.txt
```
zulcss, line 3: If you installed all the python rpm dependencies, python-click, etc? Then why are you installing them again?

zulcss: Also why are we not using tox?

portante: > If you installed all the python rpm dependencies, python-click, etc? Then why are you installing them again?

We are not installing them again.  So first, we many module dependencies which are not entirely satisfied by RPMs. At the time of this writing, that would be `elasticsearch1`, `pyesbulk`, and `pidilfe` (`python-pidfile` is the PyPI module).  By using `pip3 install` a second time, in combination with `--current-env` for `tox`, we avoid a very lengthy install process for the Jenkins tests.  For more information, see 2695e8e2a436cb0eaaa2099f231144dffad0c227 (PR #1936).

jenkins/python-setup.sh
portante: @ndokos, @RobertKrawitz, can python modules have spaces in their names?

portante: And this code was taken from https://github.com/distributed-system-analysis/pbench/blob/master/agent/profile#L27, so perhaps we could have somebody post a more thorough fixing to address the use of `ls -1d`?

jenkins/python-setup.sh
portante: > Also why are we not using tox?

We are using `tox`, with the `--current-env`.  This is how we have tests running under 3 minutes.  Otherwise, we have much longer setup times for our tests.

How long does `tox` take to install modules in each environment when no modules are previously available?

RobertKrawitz:
Module names can't have spaces in them, but you can't guarantee that this isn't being installed in a directory whose path to the root doesn't contain a space.

portante:
> Module names can't have spaces in them, but you can't guarantee that this isn't being installed in a directory whose path to the root doesn't contain a space.

Yes, we can.  The `jenkins` jobs run in a container where the `${_prefix}` variable is `/home/pbench`.

ndokos (approved):

jenkins/python-setup.sh
portante: Tracked as issue #1979.

PR #1978, opened 2020-11-16
Refactor testing

zulcss:
Simplify unit tests


dbutenhof (dismissed):
I kinda liked having separate agent/server tox targets for `py3-server` and `py3-agent`; although it's not that big a deal to just use `pytest` directly so I won't complain.

jenkins/run-pytests
```diff
@@ -22,9 +22,7 @@ if [[ ":${PYTHONPATH:-}:" != *":${_pdir}:"* ]]; then
...
 fi
 
-printf -- "black --check ...\n"
-black --check .
-printf -- "\nflake8 --verbose ...\n"
-flake8 --verbose .
-printf -- "\n"
-pytest ./lib/pbench/test/unit/common ./lib/pbench/test/unit/agent ./lib/pbench/test/unit/server ./lib/pbench/test/functional
+printf -- "lint check ... \n"
+tox -e lint
```
portante, line 26: This is already run under tox, see `tox.ini` for the `jenkins-pytests` environment.  See PR #1936.

portante (changes requested):
I agree with @dbutenhof on this, why do we need to merge the agent and server python tests into one `tox` environment?

Also, the jenkins jobs run under their own environments.

PR #1981, opened 2020-11-17
Add support for sysinfo via Tool Meister

portante:
Add support for collecting system configuration information via the Tool Meister sub-system.

The `pbench-collect-sysinfo` CLI interface has been modified to send a message to the Tool Meisters to have them collect all the requested system configuration information.  Much like tool data is collected and sent back to the Tool Data Sink from each Tool Meister, the system information is collected by each Tool Meister and sent back to the Tool Data sink to be stored in the requested location.

This means that `pbench-collect-sysinfo` no longer uses `ssh` to collect and copy the system information from remote hosts.  As it is for tools running on the local host, the local Tool Meister write the collected data into the local volume directly without sending it through the Tool Data Sink.

This change allows us to delete the `pbench-remote-sysinfo-dump` trampoline interface since the Tool Meister takes care of remote operations.

Each benchmark-script has been modified to invoke the `pbench-collect-sysinfo` following the `pbench-tool-meister-start` call, and before the `pbench-tool-meister-stop` call.

We fixed a small bug for `pbench-trafficgen` which was overwriting the "beginning" (`.../sysinfo/beg`) collected system information when collecting the "ending" (`.../sysinfo/end`) information.

The first 2 commits are small, isolated, changes to stop sub-classing `object` in Python 3 (now the default), and to clean up an instance of using f-strings in logging statements.


agent/bench-scripts/pbench-cyclictest
```diff
@@ -141,15 +141,15 @@ printf "\n" >>$benchmark_summary_txt_file
...
-pbench-collect-sysinfo --group=$tool_group --dir=$benchmark_run_dir --sysinfo=$sysinfo beg
-
 # Start the tool meisters on each registered local/remote host
 pbench-tool-meister-start "${tool_group}"
 if [[ ${?} != 0 ]]; then
 	error_log "[${script_name}]: failed to start the tool meisters."
 	exit 1
 fi
 
+pbench-collect-sysinfo --group=$tool_group --dir=$benchmark_run_dir --sysinfo=$sysinfo beg
```
dbutenhof, line 151: Now that Tool Meister is gathering sysinfo ... is there any reason we wouldn't always want to gather sysinfo immediately after starting and immediately before stopping each Tool Meister client? That is... could we remove the use of the special command and simply do it implicitly in the Tool Meister workflow?

lib/pbench/agent/tool_meister.py
```diff
@@ -474,26 +470,28 @@ def _get_data(self):
                 json_str = payload["data"].decode("utf-8")
                 data = json.loads(json_str)
             except Exception:
-                self.logger.warning("data payload in message not JSON, %r", json_str)
+                msg = f"data payload in message not JSON, {json_str!r}"
+                self.logger.warning(msg)
+                self._send_client_status(msg)
```
dbutenhof, line 475: I know this is only two lines; but it's repeated a lot, and refactoring into a "log and reply" or something would help to ensure that we don't accidentally miss some places ... and the future changes in the protocol only need to be made one place. (It'd also reduce your recurring three lines to one, since `msg` would become just the argument to the call.)

lib/pbench/agent/tool_meister.py
```diff
@@ -1014,6 +1023,82 @@ def end_tools(self, data):
...
+                # We found local tools specified in the tools group directory,
+                # so we should collect sysinfo data locally.
+                pbench-sysinfo-dump "${sysinfo_path}" "${sysinfo}" "${label}" &
+        else
+                # We have a remote directory.
+                gather_remote_sysinfo_data ${hostdir} ${label} &
+        fi
+done
+wait
+"""
```
dbutenhof, line 1075: We're keeping the entire shell script in the description string??? Do you think that's useful?

lib/pbench/agent/tool_meister.py
```diff
@@ -1014,6 +1023,82 @@ def end_tools(self, data):
...
+            self._send_client_status(msg)
+            return 1
+
+        args = data["args"]
+        if not args:
+            msg = f"No sysinfo arguments given"
+            self._send_client_status(msg)
+            return 1
+
+        # FIXME - run the sysinfo collection
```
dbutenhof, line 1090: Ah; or did you just leave the entire bash script because you haven't actually translated it yet?

lib/pbench/agent/tool_meister.py
```diff
@@ -1014,6 +1023,82 @@ def end_tools(self, data):
...
+        failures = 0
+
+        self.logger.debug("sysinfo: %r", args)
+
+        if failures > 0:
+            msg = f"{failures} collecting sysinfo data"
+            self._send_client_status(msg)
+        else:
+            self._send_client_status("success")
```
dbutenhof, line 1099: I like this sysinfo implementation: simple, easy to test! (OK, not very informative, but that's an acceptable tradeoff!)

lib/pbench/agent/toolmetadata.py
```diff
@@ -69,7 +69,7 @@ def __dataCheck(self):
         if not self.data:
             self.data == self.__getInitialData()
             if not self.data:
-                self.logger.error(f"Unable to access data through {self.mode}")
+                self.logger.error("Unable to access data through %s", self.mode)
```
dbutenhof, line 72: With multiple pre-formatted `info` and `warning` log messages above, for an `error` you change to post-formatted? Interesting choice! (No, I get that you're using the previous strings twice and there's no need for this to be pre-formatted; there's also no particular value in post-formatting since we're never going to disable `error` logs.)

I guess I'm just saying that I notice the break in pattern here. ;-)

dbutenhof (commented):
Ah; the foundation looks good. Time to build the addition!

lib/pbench/agent/tool_meister.py
portante: No, but this is just a `Draft` PR.  =)

agent/bench-scripts/pbench-cyclictest
portante: Yes, that is definitely a possibility.

We could move the default set of `sysinfo` collections performed into the `pbench-agent.cfg` file, and then when `pbench-tool-meister-start` begins, it could load those values from the global configuration.  The command line options for `--sysinfo` could be passed as an argument to `pbench-tool-meister-start` and then submitted with the base Tool Meister information posted in Redis for each instance.

Then `pbench-tool-meister-start` could initiate the client message instead of having each benchmark handle this flow.  This seems appealing to do in this commit.  Let me first get end-to-end collection of the system information first, and then I'll see about collapsing as described.

lib/pbench/agent/tool_meister.py
portante: So I have implemented the collection ... but have not sent it remotely yet.

agent/bench-scripts/pbench-trafficgen
```diff
@@ -1308,12 +1308,12 @@ fi
 
 pbench-metadata-log --group=$tool_group --dir=$benchmark_run_dir end
 
+pbench-collect-sysinfo --group=$tool_group --dir=$benchmark_run_dir --sysinfo=$sysinfo end
```
portante, line 1311: Note that we also fix the erroneous `beg` to be `end` with this change.

agent/bench-scripts/pbench-user-benchmark
```diff
@@ -355,17 +355,17 @@ for rdir in ${result_dirs}; do
 	ln -s "$(basename "${rdir}")" "$(dirname "${rdir}")/reference-result"
 done
 
+${script_path}/postprocess/user-benchmark-wrapper "${benchmark_run_dir}" "${total_duration}"
```
portante, line 358: In all the other `bench-scripts` we invoke `pbench-metadata-log`, then `pbench-collect-sysinfo`, and then `pbench-tool-meister-stop`.  This change corrects `pbench-user-benchmark` to do the same.

However, we used to run the `user-benchmark-wrapper` after `pbench-metadata-log` and then `pbench-tool-meister-stop`, immediately before `pbench-collect-sysinfo`.  Preserving that sequence was not possible, so we now run the `user-benchmark-wrapper` before we run the finalization sequence.

I don't think that will be a problem, but it seemed running it after everything would make the user wait longer for their optional script to run than it did before.  This way it runs as soon as possible.

lib/pbench/agent/tool_meister.py
```diff
@@ -431,7 +430,7 @@ def __init__(self, pbench_bin, params, redis_server, logger):
         # tools for storing their collected data.
         self._tool_dir = None
         # The temporary directory to use for capturing all tool data.
-        self._tmp_dir = os.environ.get("_PBENCH_TOOL_MEISTER_TMP", "/var/tmp")
+        self._tmp_dir = os.environ["pbench_tmp"]
```
portante, line 433: This prevents us from using `/var/tmp` during the test runs, and brings the Tool Meister in-line with the other code using the `${pbench_run}/tmp` directory.

portante (commented):
@RobertKrawitz, this adds support for collecting system information via Tool Meister, so now we no longer use `ssh` for any of that data collection.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -497,7 +502,10 @@ def _wait_for_all_data(self):
         Waiting is a no-op for all states except 'send'.  In the 'send' state,
         we are expecting to hear from all registered tool meisters.
```
dbutenhof, line 503: Expand for `sysinfo`???

lib/pbench/agent/tool_data_sink.py
```diff
@@ -497,7 +502,10 @@ def _wait_for_all_data(self):
...
         """
-        assert self.state == "send", f"expected state 'send' not '{self.state}'"
+        assert self.state in (
+            "send",
+            "sysinfo",
```
dbutenhof, line 507: I see at least three instances of the ("send", "sysinfo") tuple; I wonder if a class `send_data_states` property might not be a bad idea, to encapsulate this? Or even an `is_expecting_data()` method?

dbutenhof (approved):
A few minor comments, but looks good.

Maxusmusti (changes requested):
Sorry for delay, looks good, I agree with Dave that there may be a better way to handle the ("send, "sysinfo") tuples in the future, but not the most important thing.

Only issue I encountered was that it seems to be impossible to build rpms off this branch (would require changes to Makefile). This is the point it currently gets stuck:
`
cp -a cdm-get-iterations pbench-add-metalog-option pbench-agent-config-activate pbench-agent-config-ssh-key pbench-avg-stddev pbench-cleanup pbench-clear-results pbench-collect-sysinfo pbench-copy-results pbench-copy-result-tb pbench-end-tools pbench-get-iteration-metrics pbench-get-metric-data pbench-get-primary-metric pbench-import-cdm pbench-init-tools pbench-kill-tools pbench-list-tools pbench-log-timestamp pbench-make-result-tb pbench-metadata-log pbench-move-results pbench-output-monitor pbench-postprocess-tools pbench-postprocess-tools-cdm pbench-register-tool pbench-register-tool-set pbench-register-tool-trigger pbench-remote-sysinfo-dump pbench-send-tools pbench-start-tools pbench-stop-tools pbench-sysinfo-dump pbench-tool-meister-client pbench-tool-meister-start pbench-tool-meister-stop pbench-tool-trigger README require-rpm tool-meister /tmp/opt/pbench-agent-0.71.0/agent/util-scripts
cp: cannot stat 'pbench-remote-sysinfo-dump': No such file or directory
`

agent/bench-scripts/pbench-cyclictest
portante: Filed issue #1998 to track this work.

dbutenhof (dismissed):
Looks good!

lib/pbench/agent/tool_data_sink.py
portante: Fixed.

lib/pbench/agent/tool_data_sink.py
portante: Fixed.

portante:
@dbutenhof, @Maxusmusti, I've collapsed the code review comments commit with the others to maintain the three separate comments, and I've updated the 3rd commits description, as well as this PR description.

ndokos (approved):
We are getting rid of `pbench-remote-sysinfo-dump`? Wha's the world coming to?

Maxusmusti (approved):
Built and tested, it all works!

PR #1982, opened 2020-11-17
Harden against spaces in file names

portante:
When using the `bash` `for <var> in $(ls -1d ...); do` expressions, the shell will assign the variable values not only split by new-lines, but also by spaces.  So if a file name or directory in the path has a space in the name, the `ls -1d` output will not be interpreted correctly.

See also PR #1975.

### DIFF
--- a/agent/profile
+++ b/agent/profile
@@ -24,11 +24,13 @@ function _pbench_setup() {
 	echo "                You will need to fix it before working with pbench." >&2
     else
 	local pdir
+	local NL=$'\n'
 	while read -r pdir; do
 	    if [[ ":${PYTHONPATH:-}:" != *":${pdir}:"* ]]; then
-		export PYTHONPATH=${pdir}${PYTHONPATH:+:${PYTHONPATH}}
+		PYTHONPATH=${pdir}${PYTHONPATH:+:${PYTHONPATH}}
 	    fi
-	done <<< "${prefix}/lib $(ls -1d ${prefix}/lib*/python3.*/site-packages)"
+	done <<< "${prefix}/lib${NL}$(ls -1d ${prefix}/lib*/python3.*/site-packages)"
+	export PYTHONPATH
 	export _PBENCH_AGENT_CONFIG=${prefix}/config/pbench-agent.cfg
 
 	_pbench_pathins "${prefix}/bench-scripts"
--- a/agent/profile
+++ b/agent/profile
@@ -24,11 +24,11 @@ function _pbench_setup() {
 	echo "                You will need to fix it before working with pbench." >&2
     else
 	local pdir
-	for pdir in ${prefix}/lib $(ls -1d ${prefix}/lib*/python3.*/site-packages); do
+	while read -r pdir; do
 	    if [[ ":${PYTHONPATH:-}:" != *":${pdir}:"* ]]; then
 		export PYTHONPATH=${pdir}${PYTHONPATH:+:${PYTHONPATH}}
 	    fi
-	done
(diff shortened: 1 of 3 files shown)


portante:
@RobertKrawitz, please take a look.

dbutenhof (approved):
Wow; awkward and a bit ugly (especially when it turns into a pipeline), but feeding the loop line by line should solve the problem.

dbutenhof (dismissed):

RobertKrawitz:
(I've just gotten burned often enough by word splitting that I do whatever I can to avoid it.)

dbutenhof (approved):

portante:
@RobertKrawitz, okay, one last minor fix for empty variables, and we seem to have a working environment.

dbutenhof (approved):

agent/profile
```diff
@@ -24,11 +24,13 @@ function _pbench_setup() {
...
 	local pdir
-	for pdir in ${prefix}/lib $(ls -1d ${prefix}/lib*/python3.*/site-packages); do
+	local NL=$'\n'
+	while read -r pdir; do
 	    if [[ ":${PYTHONPATH:-}:" != *":${pdir}:"* ]]; then
-		export PYTHONPATH=${pdir}${PYTHONPATH:+:${PYTHONPATH}}
+		PYTHONPATH=${pdir}${PYTHONPATH:+:${PYTHONPATH}}
 	    fi
-	done
+	done <<< "${prefix}/lib${NL}$(ls -1d ${prefix}/lib*/python3.*/site-packages)"
```
ndokos, line 32: I think this would be clearer if you created a file, putting the new lines where you want and reading from the file. As it stands, I think that every time I get to this point, I will stumble.

jenkins/python-setup.sh
```diff
@@ -7,8 +7,10 @@ if [[ ":${PATH:-}:" != *":${_pdir}:"* ]]; then
...
 
-for _pdir in $(ls -1d ${_prefix}/lib*/python3.*/site-packages 2> /dev/null) $(head -n 1 ${_prefix}/lib*/python3.*/site-packages/pbench.egg-link 2> /dev/null); do
+NL=$'\n'
+while read -r _pdir; do
     if [[ ":${PYTHONPATH:-}:" != *":${_pdir}:"* ]]; then
-        export PYTHONPATH=${_pdir}${PYTHONPATH:+:${PYTHONPATH}}
+        PYTHONPATH=${_pdir}${PYTHONPATH:+:${PYTHONPATH}}
     fi
-done
+done <<< "$(ls -1d ${_prefix}/lib*/python3.*/site-packages 2> /dev/null)${NL}$(head -n 1 ${_prefix}/lib*/python3.*/site-packages/pbench.egg-link 2> /dev/null)"
```
ndokos, line 15: Ugh.

ndokos (changes requested):

dbutenhof (approved):
Ugly, for sure; but I'm not sure anything else is going to be less ugly!

agent/profile
portante: I'd like to avoid using files.  In the profile setup especially.

dbutenhof (approved):

riya-17 (commented):
looks good

dbutenhof (approved):

portante:
@ndokos, do you think it is worth taking this change into the code base?  I can take it or leave it.

ndokos (approved):

dbutenhof (approved):

PR #1984, opened 2020-11-17
More restorecon-related fixes

ndokos:
This is yet another continuation to PR #1973 and #1960.

This fixes the tar ball and MD5 paths (which *were* caught by
unit test 5.2, except that I was asleep at the wheel).

It also adds /usr/sbin to the PATH: when the script is executed
under cron, that is not (always?) part of the inherited PATH
and the script complains that restorecon cannot be found.


server/bin/pbench-server-prep-shim-002.sh
```diff
@@ -187,8 +192,10 @@ while read tbmd5 ;do
         (( nerrs++ ))
         continue
     fi
+
     # mv, as well as cp -a, does not restore the SELinux context properly, so we do it by hand
-    restorecon ${dest}/${tb} ${dest}/${tb}.md5
+    tbname=$(basename ${tb})
```
portante, line 197: Can't we use the `resultname` variable here instead of creating it again?

portante (changes requested):

server/bin/pbench-server-prep-shim-002.sh
```diff
@@ -11,6 +11,11 @@
 # load common things
 . $dir/pbench-base.sh
 
+# need /usr/sbin in the PATH for restorecon
+if [[ ! ":$PATH:" =~ ":/usr/sbin:" ]]; then
+   PATH=$PATH:/usr/sbin; export PATH
+fi
+
```
dbutenhof, line 18: I guess the local add is fine if we never run into anything else from /usr/sbin. Although if selinux pollutes anything else, it might be cleaner to add it to the default path.

server/bin/pbench-server-prep-shim-002.sh
ndokos: I tried that at first and then I decided to go with the `tbname` (`resultname` has the `.tar.xz` stripped off, so we need to put it back on here, and I thought that was uglier), but I don't have a strong feeling one way or the other.

server/bin/pbench-server-prep-shim-002.sh
ndokos: I considered adding it to the trampoline script but in the end I thought it would be better to localize it as much as possible. Given that the shell script is going away in 0.70.0 (and we have to worry about the selinux python package, rather than the restorecon executable), I think I'll leave it as is.

server/bin/pbench-server-prep-shim-002.sh
ndokos: Changed this and also fixed the log message.

portante (dismissed):

dbutenhof (approved):
That makes sense.

portante:
@ndokos, looks like we need to either mock out `restorecon` in the unit tests under TravsiCI, or we need to install the Ubuntu package for `restorecon`.

ndokos:
Aargh - the unit tests failed because I inadvertently added a `1` , making restorecon always return an error :-) This last push just fixed the typo.

portante (approved):

dbutenhof (approved):

Issue #1955, opened 2020-11-05
We need a better mechanism to handle sys.exit() in lib/pbench/cli/agent/commands/results/__init__.py


npalaska:
Currently there are lot of instances in `lib/pbench/cli/agent/commands/results/__init__.py` where we use `sys.exit()` which leads to some undesirable behaviors.  We anyway should avoid directly using `sys.exit()` as much as possible. See the comment [#1947](https://github.com/distributed-system-analysis/pbench/pull/1947#discussion_r517441807)

riya-17:
@portante Isn't this issue been resolved with #2189 PR? Or is there some more work that needs to be done?

portante:
More work, I'd start by reviewing all the code in the library that calls `sys.exit()`.

PR #1985, opened 2020-11-17
Refactor server API implementation

npalaska:
This PR attempts to refactor some of the server API implementations

- Remove global variable `app` 
- Keep `create_app` functionality clean and high level.
- Modularize the API implementations
  - Create Separate api module for upload, elasticsearch and graphql
- Remove custom implementation of CORS 
  - Use `flask_cors` module to add  CORS decorator
- Use the loaded PbenchServerConfig object instead of reassigning them to app.config

### DIFF
--- a/lib/pbench/cli/server/shell.py
+++ b/lib/pbench/cli/server/shell.py
@@ -1,14 +1,37 @@
 #!/bin/env python3
 
-from pbench.server.api import create_app
 import subprocess
+import sys
+
+from configparser import NoSectionError, NoOptionError
+
+from pbench.common.exceptions import BadConfig, ConfigFileNotSpecified
+from pbench.server.api import create_app, get_server_config
+
+
+def app():
+    try:
+        server_config = get_server_config()
+    except (ConfigFileNotSpecified, BadConfig) as e:
+        print(e)
+        sys.exit(1)
+    return create_app(server_config)
 
 
 def main():
-    app = create_app()
-    port = app.config["PORT"]
-    host = app.config["BIND_HOST"]
-    workers = app.config["WORKERS"]
+    try:
+        server_config = get_server_config()
+    except (ConfigFileNotSpecified, BadConfig) as e:
+        print(e)
+        sys.exit(1)
+    try:
+        port = server_config.get("pbench-server", "rest_port")
+        host = server_config.get("pbench-server", "bind_host")
+        workers = server_config.get("pbench-server", "workers")
+    except (NoOptionError, NoSectionError) as e:
+        print(f"{__name__}: ERROR: {e.__traceback__}")
+        sys.exit(1)
+
     subprocess.run(
         [
             "gunicorn",
@@ -18,6 +41,6 @@ def main():
             "/run/pbench-server/gunicorn.pid",
             "--bind",
             f"{str(host)}:{str(port)}",
-            "pbench.server.api:create_app()",
+            f"pbench.cli.server.shell:app())",
         ]
     )
--- a/lib/pbench/server/__init__.py
+++ b/lib/pbench/server/__init__.py
@@ -61,7 +61,6 @@ class PbenchServerConfig(PbenchConfig):
             "BINDIR", "pbench-server", "script-dir"
         )
         self.LIBDIR = self._get_valid_dir_option("LIBDIR", "pbench-server", "lib-dir")
-        self.mail_recipients = self.conf.get("pbench-server", "mailto")
         self.ARCHIVE = self._get_valid_dir_option(
             "ARCHIVE", "pbench-server", "pbench-archive-dir"
         )
@@ -81,25 +80,6 @@ class PbenchServerConfig(PbenchConfig):
         except NoOptionError:
             self.COMMIT_ID = "(unknown)"
 
-        try:
-            self._unittests = self.conf.get("pbench-server", "debug_unittest")
-        except Exception:
-            self._unittests = False
-        else:
-            self._unittests = bool(self._unittests)
-
-        try:
-            self.elasticsearch = f"{self.conf.get('elasticsearch', 'host')}:{self.conf.get('elasticsearch', 'port')}"
-        except (NoOptionError, NoSectionError):
-            self.elasticsearch = ""
-
-        try:
-            self.graphql = (
-                f"{self.conf.get('graphql', 'host')}:{self.conf.get('graphql', 'port')}"
-            )
-        except (NoOptionError, NoSectionError):
-            self.graphql = ""
-
         if self._unittests:
 
             def mocked_time():
@@ -107,14 +87,6 @@ class PbenchServerConfig(PbenchConfig):
 
             global _time
             _time = mocked_time
-
-            try:
-                ref_dt_str = self.conf.get("pbench-server", "debug_ref_datetime")
-            except Exception:
-                ref_dt_str = "1970-01-02T00:00:00.000000"
-            self._ref_datetime = datetime.strptime(ref_dt_str, _STD_DATETIME_FMT)
-        else:
-            self._ref_datetime = None
 
         # Constants
 
@@ -142,6 +114,41 @@ class PbenchServerConfig(PbenchConfig):
             + " "
             + " ".join([f"WONT-INDEX.{i:d}" for i in range(1, 12)])
         )
+
+    @property
+    def server_config(self):
+        return self.conf["pbench-server"]
+
+    @property
+    def mail_recipients(self):
+        try:
+            return self.conf.get("pbench-server", "mailto")
+        except (NoOptionError, NoSectionError):
+            return ""
+
+    @property
+    def _unittests(self):
+        try:
+            unittests = self.conf.get("pbench-server", "debug_unittest")
+        except (NoOptionError, NoSectionError):
+            return False
+        else:
+            return bool(unittests)
+
+    @property
+    def _ref_datetime(self):
+        if self._unittests:
+            try:
+                ref_dt_str = self.conf.get("pbench-server", "debug_ref_datetime")
+            except Exception:
+                ref_dt_str = "1970-01-02T00:00:00.000000"
+            return datetime.strptime(ref_dt_str, _STD_DATETIME_FMT)
+        else:
+            return None
+
+    @property
+    def rest_uri(self):
+        return self._get_conf("pbench-server", "rest_uri")
 
     def _get_conf(self, section, option):
         """
--- a/lib/pbench/server/api/__init__.py
+++ b/lib/pbench/server/api/__init__.py
@@ -5,442 +5,76 @@ Dashboard or any number of pbench agent users.
 """
 
 import os
-import sys
-import hashlib
-from pathlib import Path
-import requests
-import humanize
-import tempfile
 
-from flask import request, jsonify, Flask, make_response
-from flask_restful import Resource, abort, Api
-from werkzeug.utils import secure_filename
+from flask import Flask
+from flask_restful import Api
 
-from pbench.common.logger import get_pbench_logger
 from pbench.server import PbenchServerConfig
-from pbench.common.exceptions import BadConfig
-from pbench.server.utils import filesize_bytes
-
-from pbench.server.api.resources.query_controllers import QueryControllers
-
-ALLOWED_EXTENSIONS = {"xz"}
-
-
-def allowed_file(filename):
-    """Check if the file has the correct extension."""
-    try:
-        fn = filename.rsplit(".", 1)[1].lower()
-    except IndexError:
-        return False
-    allowed = "." in filename and fn in ALLOWED_EXTENSIONS
-    return allowed
+from pbench.common.exceptions import BadConfig, ConfigFileNotSpecified
+from pbench.server.api.resources.upload_api import Upload, HostInfo
+from pbench.server.api.resources.graphql_api import GraphQL
+from pbench.common.logger import get_pbench_logger
+from pbench.server.api.resources.query_apis.elasticsearch_api import Elasticsearch
+from pbench.server.api.resources.query_apis.query_controllers import QueryControllers
 
 
-def register_endpoints(api, app):
+def register_endpoints(api, app, config):
     """Register flask endpoints with the corresponding resource classes
     to make the APIs active."""
 
-    base_uri = app.config["REST_URI"]
+    base_uri = config.rest_uri
     app.logger.info("Registering service endpoints with base URI {}", base_uri)
+
     api.add_resource(
         Upload,
         f"{base_uri}/upload/ctrl/<string:controller>",
-        resource_class_args=(app, api),
+        resource_class_args=(config, app.logger),
     )
+    api.add_resource(
+        HostInfo, f"{base_uri}/host_info", resource_class_args=(config, app.logger),
+    )
+    api.add_resource(
+        Elasticsearch,
+        f"{base_uri}/elasticsearch",
+        resource_class_args=(config, app.logger),
+    )
+    api.add_resource(
+        GraphQL, f"{base_uri}/graphql", resource_class_args=(config, app.logger),
+    )
+
     api.add_resource(
         QueryControllers,
         f"{base_uri}/controllers/list",
-        resource_class_args=(app, api),
+        resource_class_args=(config, app.logger),
     )
-    api.add_resource(HostInfo, f"{base_uri}/host_info", resource_class_args=(app, api))
-    api.add_resource(
-        Elasticsearch, f"{base_uri}/elasticsearch", resource_class_args=(app, api)
-    )
-    api.add_resource(GraphQL, f"{base_uri}/graphql", resource_class_args=(app, api))
 
 
-def create_app():
-    """Create Flask app with defined resource endpoints."""
-
+def get_server_config():
     cfg_name = os.environ.get("_PBENCH_SERVER_CONFIG")
     if not cfg_name:
-        print(
-            f"{__name__}: ERROR: No config file specified; set"
-            " _PBENCH_SERVER_CONFIG",
-            file=sys.stderr,
+        raise ConfigFileNotSpecified(
+            f"{__name__}: ERROR: No config file specified; set" " _PBENCH_SERVER_CONFIG"
         )
-        sys.exit(1)
 
     try:
-        config = PbenchServerConfig(cfg_name)
+        return PbenchServerConfig(cfg_name)
     except BadConfig as e:
-        print(f"{__name__}: {e} (config file {cfg_name})", file=sys.stderr)
-        sys.exit(1)
+        raise Exception(f"{__name__}: {e} (config file {cfg_name})").with_traceback(
+            e.__traceback__
+        )
 
-    app = Flask(__name__)
+
+def create_app(server_config):
+    """Create Flask app with defined resource endpoints."""
+
+    app = Flask("api-server")
     api = Api(app)
 
-    app.logger = get_pbench_logger(__name__, config)
-    app.config_server = config.conf["pbench-server"]
-    app.config_elasticsearch = config.conf["elasticsearch"]
-    app.config_graphql = config.conf["graphql"]
+    app.logger = get_pbench_logger(__name__, server_config)
 
-    prdp = app.config_server.get("pbench-receive-dir-prefix")
-    if not prdp:
-        app.logger.error("Missing config variable for pbench-receive-dir-prefix")
-        sys.exit(1)
-    try:
-        upload_directory = Path(f"{prdp}-002").resolve(strict=True)
-    except FileNotFoundError:
-        app.logger.exception("pbench-receive-dir-prefix does not exist on the host")
-        sys.exit(1)
-    except Exception:
-        app.logger.exception(
-            "Exception occurred during setting up the upload directory on the host"
-        )
-        sys.exit(1)
-    else:
-        app.upload_directory = upload_directory
+    app.config["DEBUG"] = False
+    app.config["TESTING"] = False
 
-    app.config["ES_URL"] = (
-        "http://"
-        + app.config_elasticsearch.get("host")
-        + ":"
-        + app.config_elasticsearch.get("port")
-    )
-    app.config["PORT"] = app.config_server.get("rest_port")
-    app.config["VERSION"] = app.config_server.get("rest_version")
-    app.config["MAX_CONTENT_LENGTH"] = filesize_bytes(
-        app.config_server.get("rest_max_content_length")
-    )
-    app.config["PREFIX"] = config.conf["Indexing"].get("index_prefix")
-    app.config["REST_URI"] = app.config_server.get("rest_uri")
-    app.config["BIND_HOST"] = app.config_server.get("bind_host")
-    app.config["WORKERS"] = app.config_server.get("workers")
-
-    register_endpoints(api, app)
+    register_endpoints(api, app, server_config)
 
     return app
-
-
-def _build_cors_preflight_response():
-    response = make_response()
-    response = _corsify_actual_response(response)
-    response.headers.add("Access-Control-Allow-Headers", "*")
-    response.headers.add("Access-Control-Allow-Methods", "*")
-    return response
-
-
-def _corsify_actual_response(response):
-    response.headers.add("Access-Control-Allow-Origin", "*")
-    return response
-
-
-class CorsOptionsRequest:
-    """
-    From the client side, an OPTIONS request is initiated before every POST to request CORS.
-    On the server side, the server must specify the response headers to grant CORS accordingly.
-    """
-
-    def __init__(self, app, options_exception_msg):
-        self.app = app
-        self.options_exception_msg = options_exception_msg
-
-    def options(self):
-        try:
-            response = _build_cors_preflight_response()
-        except Exception:
-            self.app.logger.exception(self.options_exception_msg)
-            abort(400, message=self.options_exception_msg)
-        response.status_code = 200
-        return response
-
-
-class GraphQL(Resource, CorsOptionsRequest):
-    """GraphQL API for post request via server."""
-
-    def __init__(self, app, api):
-        CorsOptionsRequest.__init__(
-            self, app, "Bad options request for the GraphQL query"
-        )
-        self.api = api
-
-    def post(self):
-        logger = self.app.logger
-        gql = self.app.config_graphql
-        graphQL = f"http://{gql.get('host')}:{gql.get('port')}"
-        json_data = request.get_json(silent=True)
-
-        if not json_data:
-            logger.warning("GraphQL: Invalid json object {}", request.url)
-            abort(400, message="GraphQL: Invalid json object in request")
-
-        try:
-            # query GraphQL
-            gql_response = requests.post(graphQL, json=json_data)
-            gql_response.raise_for_status()
-        except requests.exceptions.ConnectionError:
-            logger.exception("Connection refused during the GraphQL post request")
-            abort(502, message="Network problem, could not post to GraphQL Endpoint")
-        except requests.exceptions.Timeout:
-            logger.exception("Connection timed out during the GraphQL post request")
-            abort(
-                504, message="Connection timed out, could not post to GraphQL Endpoint"
-            )
-        except Exception:
-            logger.exception("Exception occurred during the GraphQL post request")
-            abort(500, message="INTERNAL ERROR")
-
-        try:
-            # construct response object
-            raw_response = make_response(gql_response.text)
-            response = _corsify_actual_response(raw_response)
-        except Exception:
-            logger.exception("Exception occurred in GraphQL response construction")
-            abort(
-                500, message="INTERNAL ERROR",
-            )
-
-        response.status_code = gql_response.status_code
-        return response
-
-
-class Elasticsearch(Resource, CorsOptionsRequest):
-    """Elasticsearch API for post request via server."""
-
-    def __init__(self, app, api):
-        CorsOptionsRequest.__init__(
-            self, app, "Bad options request for the Elasticsearch query"
-        )
-        self.api = api
-
-    def post(self):
-        logger = self.app.logger
-        es = self.app.config_elasticsearch
-        elasticsearch = f"http://{es.get('host')}:{es.get('port')}"
-        json_data = request.get_json(silent=True)
-        if not json_data:
-            logger.warning("Elasticsearch: Invalid json object. Query: {}", request.url)
-            abort(400, message="Elasticsearch: Invalid json object in request")
-
-        if not json_data["indices"]:
-            logger.warning("Elasticsearch: Missing indices path in the post request")
-            abort(400, message="Missing indices path in the Elasticsearch request")
-
-        try:
-            # query Elasticsearch
-            if "params" in json_data:
-                url = f"{elasticsearch}/{json_data['indices']}?{json_data['params']}"
-            else:
-                url = f"{elasticsearch}/{json_data['indices']}"
-
-            if "payload" in json_data:
-                es_response = requests.post(url, json=json_data["payload"])
-            else:
-                logger.debug("No payload found in Elasticsearch post request json data")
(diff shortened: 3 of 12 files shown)


lib/pbench/server/api/config.py
```diff
@@ -0,0 +1,109 @@
...
+
+    @property
+    def ELASTICSEARCH_URI(self):
+        return f"http://{self.CONFIG.conf['elasticsearch'].get('host')}:{self.CONFIG.conf['elasticsearch'].get('port')}"
+
+    @property
+    def GRAPHQL_URI(self):
+        return f"http://{self.CONFIG.conf['graphql'].get('host')}:{self.CONFIG.conf['graphql'].get('port')}"
+
+    # Add prod specific db config
```
dbutenhof, line 99: For the others, it's arguable whether it makes more sense to pre-fetch and return local properties vs grabbing them from the composed config object. In these, however, I think that returning local properties instead of doing the string formatting on each call would make more sense.

lib/pbench/server/api/__init__.py
```diff
@@ -48,59 +49,16 @@ def register_endpoints(api, app):
     api.add_resource(GraphQL, f"{app.config['REST_URI']}/graphql")
 
 
-def create_app():
+def create_app(server_config=ProdServerConfig):
```
dbutenhof, line 52: I'd rather see you pass in an _object_ rather than a _class_ that gets instantiated here. That just feels awkward; and passing an object will provide more flexibility for testing, too.

lib/pbench/server/api/config.py
npalaska: yeah I could define a property to get elasticsearch_host and elasticsearch_port separately but then where do we need these properties as separate things, each time we ask for es host and port its because we want to form a URI so I thought I would just return the URI because thats what most of the time we need

lib/pbench/server/api/__init__.py
npalaska: you mean something like `create_app(server_config=ProdServerConfig()):` but then what do you mean by instantiating it? The config classes ideally wouldn't have any methods just constant variables

portante (changes requested):
Can you update the description for this PR with more details on what problem this is solving?

It appears to be creating a object to which properties are derived from the loaded configuration.  Can't we just add those properties to the existing `PbenchServerConfig` class?

lib/pbench/server/api/__init__.py
```diff
@@ -48,59 +49,16 @@ def register_endpoints(api, app):
...
-
-    app.config["PORT"] = app.config_server.get("rest_port")
-    app.config["VERSION"] = app.config_server.get("rest_version")
-    app.config["MAX_CONTENT_LENGTH"] = filesize_bytes(
-        app.config_server.get("rest_max_content_length")
-    )
-    app.config["REST_URI"] = app.config_server.get("rest_uri")
-    app.config["BIND_HOST"] = app.config_server.get("bind_host")
-    app.config["WORKERS"] = app.config_server.get("workers")
+    app.config.from_object(server_config())
```
dbutenhof, line 60: `server_config()` here is instantiating an object of your class; I just think that constructor should be called from the client, passing in the object rather than a reference to the class.

lib/pbench/server/api/config.py
dbutenhof: No, what I meant is that it'd be (slightly) more efficient and a bit cleaner to construct this string in `__init__` and simply return the constructed property here rather than re-fetching the config options and rebuilding the template each time.

lib/pbench/server/api/__init__.py
```diff
@@ -63,44 +57,35 @@ def create_app():
...
     try:
-        config = PbenchServerConfig(cfg_name)
+        server_config = PbenchServerConfig(cfg_name)
     except BadConfig as e:
         print(f"{__name__}: {e} (config file {cfg_name})", file=sys.stderr)
         sys.exit(1)
 
-    app = Flask(__name__)
-    api = Api(app)
+    app.config["SERVER_CONFIG"] = server_config.server_config
```
portante, line 65: Why are we going through the trouble of still using `app.config`?  Why not just reference the `server_config` instead?

portante (changes requested):

lib/pbench/server/api/__init__.py
npalaska: I was following the Flask config handling https://flask.palletsprojects.com/en/1.1.x/config/ where it says 

> Independent of how you load your config, there is a config object available which holds the loaded configuration values: The config attribute of the Flask object. This is the place where Flask itself puts certain configuration values and also where extensions can put their configuration values. But this is also where you can have your own configuration.

Flask `Config` is a subclass of a dictionary which comes with some pre-configured values and if we use extensions such as for example `flask_bcrypt` then we would need a `SECRET_KEY` or some key to bcrypt the password hash, this key can be stored as a config `key` (now it can also be just retrieved with PbenchServerConfig class, however some extensions can read the default value if the key is set). The advantage I see is that we dont have to pass the entire server_config object to other modules to get certain values (we just pass the app instance) and also for the testing when we import the `create_app` to make the `app_client` we can assign the `app_client.config = app.config` and therefore we so we dont need to make another serverconfig object

npalaska: but on the second thought I agree it adds an extra redundancy for our use case

npalaska:
I have updated the PR description and name, since its now doing more than just app config changes. I took this PR as an opportunity to modularize the api implementation and some cleanups as mentioned in the description. I cherry-picked these changes (specifically modularize commit) from pbench_user implementation PR and added here since these changes shouldn't be part of that PR.

lib/pbench/server/__init__.py
```diff
@@ -143,6 +117,101 @@ def mocked_time():
             + " ".join([f"WONT-INDEX.{i:d}" for i in range(1, 12)])
         )
 
+    @property
+    def server_config(self):
+        return self.conf["pbench-server"]
+
+    @property
+    def logger(self):
```
portante, line 125: This seems to be adding a new `logger` attribute (property) to a `PbenchServerConfig` class.  All the existing callers create their own loggers from the server config.  Can we have the API instance do the same?

portante (changes requested):

lib/pbench/server/__init__.py
npalaska: Addressed.

lib/pbench/server/api/resources/graphql_api.py
```diff
@@ -0,0 +1,50 @@
...
+            self.logger.exception("Exception occurred during the GraphQL post request")
+            abort(500, message="INTERNAL ERROR")
+
+        try:
+            # construct response object
+            response = make_response(gql_response.text)
+        except Exception:
+            self.logger.exception("Exception occurred GraphQL response construction")
+            abort(
+                500, message="INTERNAL ERROR",
```
portante, line 46: I wonder why this formats differently from line 38?

npalaska: Ahh I thinks its because of the `,` at the end, good catch

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -0,0 +1,192 @@
...
+from flask import request, jsonify
+from werkzeug.utils import secure_filename
+
+ALLOWED_EXTENSIONS = {"xz"}
+
+
+class HostInfo(Resource):
+    def __init__(self, server_config, logger, api):
+        self.server_config = server_config
+        self.api = api
```
portante, line 16: It does not appear that `api` is actually used by this class.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -0,0 +1,192 @@
...
+                dict(
+                    message=f"{self.server_config.get('user')}@{self.server_config.get('host')}"
+                    f":{self.server_config.get('pbench-receive-dir-prefix')}-002"
+                )
+            )
+        except Exception:
+            self.logger.exception(
+                "There was something wrong constructing the host info"
+            )
+            abort(500, message="There was something wrong with your request")
```
portante, line 31: Feel free to create an issue since this is not really your code changes ...

Perhaps just report back to the client, "`INTERNAL ERROR`"?

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -0,0 +1,192 @@
...
+        self.server_config = server_config
+        self.api = api
+        self.logger = logger
+
+    def get(self):
+        try:
+            response = jsonify(
+                dict(
+                    message=f"{self.server_config.get('user')}@{self.server_config.get('host')}"
+                    f":{self.server_config.get('pbench-receive-dir-prefix')}-002"
```
portante, line 24: Instead of fetching this each time, and getting the configuration error each time, perhaps we could fetch these values one in the `HostInfo` constructor, and raising the exception there if they are missing, and then just reference the constructed response.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -0,0 +1,192 @@
...
+            )
+            abort(500, message="There was something wrong with your request")
+        response.status_code = 200
+        return response
+
+
+class Upload(Resource):
+    def __init__(self, upload_directory, max_content_length, logger, api):
+        self.upload_directory = upload_directory
+        self.max_content_length = max_content_length
```
portante, line 39: This item is specific to the `Upload` class, so perhaps we should pass the configuration object so that we don't need to endow knowledge of what the `Upload` class needs outside of the class itself.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -0,0 +1,192 @@
...
+        self.logger = logger
+        self.api = api
```
portante, line 41: Not sure where `api` is used.

lib/pbench/server/api/resources/elasticsearch_api.py
```diff
@@ -0,0 +1,81 @@
...
+from flask import request, make_response
+from flask_cors import cross_origin
+
+
+class Elasticsearch(Resource):
+    """Elasticsearch API for post request via server."""
+
+    def __init__(self, elasticsearch_uri, logger, api):
+        self.elasticsearch = elasticsearch_uri
+        self.api = api
```
portante, line 12: Not sure where `api` is used.

lib/pbench/server/api/resources/elasticsearch_api.py
```diff
@@ -0,0 +1,81 @@
...
+from flask_restful import Resource, abort
+from flask import request, make_response
+from flask_cors import cross_origin
+
+
+class Elasticsearch(Resource):
+    """Elasticsearch API for post request via server."""
+
+    def __init__(self, elasticsearch_uri, logger, api):
+        self.elasticsearch = elasticsearch_uri
```
portante, line 11: This looks like a specific configuration that the `Elasticsearch` class could pull out of the configuration object so that knowledge of this class remains with the class.

lib/pbench/server/api/__init__.py
```diff
@@ -63,361 +60,24 @@ def create_app():
...
+        return PbenchServerConfig(cfg_name)
     except BadConfig as e:
         print(f"{__name__}: {e} (config file {cfg_name})", file=sys.stderr)
         sys.exit(1)
 
+
+def create_app():
+    """Create Flask app with defined resource endpoints."""
+
     app = Flask(__name__)
```
portante, line 72: Is `__name__` going to resolve to "`api`" or to "`__init__`"?  If the former, great; if the latter, then let's not use `__name__` but perhaps `api-server`, or something like that.

lib/pbench/server/__init__.py
```diff
@@ -143,6 +117,97 @@ def mocked_time():
...
+            try:
+                ref_dt_str = self.conf.get("pbench-server", "debug_ref_datetime")
+            except Exception:
+                ref_dt_str = "1970-01-02T00:00:00.000000"
+            return datetime.strptime(ref_dt_str, _STD_DATETIME_FMT)
+        else:
+            return None
+
+    @property
+    def upload_directory(self):
```
portante, line 166: Can we move this to the `Upload` constructor class instead?

lib/pbench/server/__init__.py
```diff
@@ -143,6 +117,97 @@ def mocked_time():
...
+
+    @property
+    def elasticsearch(self):
+        try:
+            return f"http://{self.conf.get('elasticsearch', 'host')}:{self.conf.get('elasticsearch', 'port')}"
+        except (NoOptionError, NoSectionError):
+            return ""
+
+    @property
+    def graphql(self):
```
portante, line 132: Can we move this to the `GraphQL` class constructor instead?

lib/pbench/server/__init__.py
```diff
@@ -143,6 +117,97 @@ def mocked_time():
             + " ".join([f"WONT-INDEX.{i:d}" for i in range(1, 12)])
         )
 
+    @property
+    def server_config(self):
+        return self.conf["pbench-server"]
+
+    @property
+    def elasticsearch(self):
```
portante, line 125: Can we move this to the `Elasticsearch` constructor instead?

lib/pbench/server/__init__.py
```diff
@@ -143,6 +117,97 @@ def mocked_time():
...
+    @property
+    def port(self):
+        return self._get_conf("pbench-server", "rest_port")
+
+    @property
+    def version(self):
+        return self._get_conf("pbench-server", "rest_version")
+
+    @property
+    def max_content_length(self):
```
portante, line 194: Can we move this to the `Upload` constructor instead?

portante (changes requested):
Not sure about all these new properties, as they appear to be very specific to one case, used once, and not general enough to be useful as properties.

lib/pbench/server/api/resources/elasticsearch_api.py
npalaska: yeah I agree, I did that before where I was passing whole config object and elasticsearch class will fetch the required parameters but then I changed it to something like give me only what I want framework, but you put a good point that this is a specific configuration for the Elasticsearch class and let it fetch what it wants

lib/pbench/server/api/__init__.py
npalaska: I am not 100% sure but I think it will resolve to the module name. ref: https://flask.palletsprojects.com/en/1.1.x/tutorial/factory/ so in our case `pbench.server.api`

> `__name__` is the name of the current Python module. The app needs to know where it’s located to set up some paths, and `__name__` is a convenient way to tell it that.

lib/pbench/server/api/resources/graphql_api.py
```diff
@@ -0,0 +1,48 @@
...
+        self.api = api
+        self.logger = logger
+
+    @cross_origin()
+    def post(self):
+        json_data = request.get_json(silent=True)
+
+        if not json_data:
+            self.logger.warning("GraphQL: Invalid json object {}", request.url)
+            abort(400, message="GraphQL: Invalid json object in request")
```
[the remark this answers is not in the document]
npalaska: I tried to address it here

lib/pbench/cli/server/shell.py
```diff
@@ -1,14 +1,18 @@
 #!/bin/env python3
 
-from pbench.server.api import create_app
+from pbench.server.api import create_app, get_server_config
 import subprocess
 
 
+server_config = get_server_config()
+application = create_app(server_config)
```
portante, line 8: Seems like these should not be global variables but local only to `main()`.

lib/pbench/server/api/__init__.py
portante: Try this out:

```
$ cat this.py
class Foo:
    def __init__(self):
        self.myname = "foo"

    def bar(self):
        print(self.myname, __name__)


def this():
    print(__name__)

print(__name__)
this()
Foo().bar()
$ python3 this.py
__main__
__main__
foo __main__
```

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -190,3 +194,25 @@ def allowed_file(filename):
...
         allowed = "." in filename and fn in ALLOWED_EXTENSIONS
         return allowed
+
+    @property
+    def upload_directory(self):
+        prdp = self.config.get_conf(
+            __name__, "pbench-server", "pbench-receive-dir-prefix", self.logger
+        )
+        try:
+            return Path(f"{prdp}-002").resolve(strict=True)
```
portante, line 204: Since we are returning a `Path` object here, line 106 can be changed to `path = self.upload_directory / controller`.

lib/pbench/test/unit/server/conftest.py
```diff
@@ -83,8 +83,9 @@ def client(pytestconfig, monkeypatch):
 
     from pbench.server.api import create_app, get_server_config
```
portante, line 84: Shouldn't this be at the top of the file with all the other imports?

portante (changes requested):

lib/pbench/cli/server/shell.py
npalaska: I had to do this because the `gunicorn` expect the format to be `{module}:flask_app_instance` to run the app. If I define it inside the main function then I have to do something like `pbench.cli.server.shell:main()` where `main` function need to return the flask_app instance. Another alternative would be to create another file say `wsgi.py` to hold the application and `shell.py` would import it and then gunicorn can call `pbench.cli.server.wsgi:application()` where application function in `wsgi.py` would return the app instance

lib/pbench/test/unit/server/conftest.py
npalaska: ahh yeah

lib/pbench/server/api/__init__.py
npalaska: the yes but we dont directly execute the `__init__` file, we execute the file via gunicorn. to do that we import `create_app` in `shell.py` file `from pbench.server.api import create_app` and if I execute shell.py file I get `pbench.server.api` when I do `print(__name__)` inside `create_app()`

lib/pbench/server/api/__init__.py
```diff
@@ -6,53 +6,43 @@
...
-    api.add_resource(Elasticsearch, f"{app.config['REST_URI']}/elasticsearch")
-    api.add_resource(GraphQL, f"{app.config['REST_URI']}/graphql")
-
-
-def create_app():
-    """Create Flask app with defined resource endpoints."""
 
-    global app
 
+def get_server_config():
```
dbutenhof, line 45: Seems like this should probably be in the shell, and simply passed in the config object; no need for shell to import this from the api just to construct the object and pass it back. There's nothing in here that's specific to the API implementation.

lib/pbench/server/api/resources/elasticsearch_api.py
```diff
@@ -0,0 +1,82 @@
...
+            else:
+                url = f"{self.elasticsearch}/{json_data['indices']}"
+
+            if "payload" in json_data:
+                es_response = requests.post(url, json=json_data["payload"])
+            else:
+                self.logger.debug(
+                    "No payload found in Elasticsearch post request json data"
+                )
+                es_response = requests.get(url)
```
dbutenhof, line 46: You need to add the `es_response.raise_for_status()` here to activate the exception transformation. That's in my PR; however since you've moved everything your merge is going to be messy and it could easily be overlooked!

lib/pbench/server/api/__init__.py
portante: Right, so that is the module name, but we want a logger for the instance of the service.  I am not sure we should name the logger based on the module.  At least, in other parts of the server code, we name the logger explicitly, see https://github.com/distributed-system-analysis/pbench/blob/master/lib/pbench/server/indexer.py#L4415.

lib/pbench/server/api/resources/graphql_api.py
```diff
@@ -0,0 +1,51 @@
...
+
+        if not json_data:
+            message = "Invalid json object"
+            self.logger.warning(f"{message}: {request.url}")
+            abort(400, message=message)
+
+        try:
+            # query GraphQL
+            gql_response = requests.post(self.graphql, json=json_data)
+        except requests.exceptions.ConnectionError:
```
dbutenhof, line 29: `qql_response.raise_for_status()`

lib/pbench/test/unit/server/conftest.py
```diff
@@ -12,13 +11,13 @@
 [pbench-server]
 pbench-top-dir = {TMP}/srv/pbench
 
-[elasticsearch]
-host = elasticsearch.example.com
-port = 7080
+[elasticsearch]
```
dbutenhof, line 14: Did you mean to add trailing whitespace on these lines? How'd it get there?

dbutenhof (commented):
And of course you need to rebase with my changes, which is going to be enormous fun. :-)

lib/pbench/test/unit/server/conftest.py
npalaska: Ahh yeah I remove them, I was playing around with it and it got added

lib/pbench/server/api/resources/elasticsearch_api.py
npalaska: Thanks for the heads up

npalaska:
> And of course you need to rebase with my changes, which is going to be enormous fun. :-)

I just rebased and indeed it is going to be lot of fun

lib/pbench/server/api/__init__.py
npalaska: Ahh I see, thanks for the clarification

lib/pbench/server/api/__init__.py
npalaska: That's correct but we also need to import this function in `conftest.py` to build the flask client app for pytests and thats why I kept it here. (I can probably put it in the utils file but little skeptical about having to import `PbenchConfigClass` in the utils file coz in the future as we add more stuff and write utility functions we might encounter circular import scenarios. Correct me if I am wrong but util files shouldnt import important classes right?)

lib/pbench/server/__init__.py
```diff
@@ -143,6 +115,55 @@ def mocked_time():
...
+        except (NoOptionError, NoSectionError):
+            return ""
+
+    @property
+    def index_prefix(self):
+        try:
+            return self.conf.get("Indexing", "index_prefix")
+        except (NoOptionError, NoSectionError):
+            return ""
+
```
npalaska, line 135: I believe we might need these properties as we add more es queries

portante: The question is would we have code written in another module that talks to Elasticsearch NOT use the `elasticsearch_api` module?

The query_controllers module also talks to Elasticsearch, but why not make one module which handles the communication for both?

lib/pbench/server/__init__.py
```diff
@@ -143,6 +115,55 @@ def mocked_time():
             + " ".join([f"WONT-INDEX.{i:d}" for i in range(1, 12)])
         )
 
+    @property
+    def server_config(self):
+        return self.conf["pbench-server"]
+
+    @property
+    def es_url(self):
```
portante, line 123: These seems pretty specific to the Elasticsearch module we have.

lib/pbench/cli/server/shell.py
portante: Right, but why not declare only `application` as `global`, but set its value to `None`.  Then in `main()`, fetch the server configuration, create the application instance, assigning it to the `application` global variable, and then invoke `gunicorn`?

This prevents the application and configuration from being loaded at module load time, so that we can unit test on demand, clearing out the value of the global variable only.

portante (changes requested):
Getting there, I think we can resolve the need for global state fairly easily, and address the separation of concerns with respect to the configuration properties and be ready to merge.

lib/pbench/server/api/__init__.py
```diff
@@ -12,11 +12,11 @@
 
 from pbench.server import PbenchServerConfig
 from pbench.common.exceptions import BadConfig
-from pbench.server.api.resources.elasticsearch_api import Elasticsearch
 from pbench.server.api.resources.upload_api import Upload, HostInfo
 from pbench.server.api.resources.graphql_api import GraphQL
-from pbench.server.api.resources.query_controllers import QueryControllers
 from pbench.common.logger import get_pbench_logger
+from pbench.server.api.resources.query_apis.elasticsearch_api import Elasticsearch
+from pbench.server.api.resources.query_apis.query_controllers import QueryControllers
```
portante, line 19: Any particular reason for this change?

portante (approved):

portante (changes requested):
Actually, needs a rebase ...

lib/pbench/server/api/__init__.py
npalaska: I just thought we could keep the elasticsearch related query apis in one place so its organized and easy to remove and/or add more apis or make changes that would reflect on all the es apis

portante: Typically we have been following the PEP 8 on [Imports](https://www.python.org/dev/peps/pep-0008/#imports):

> Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.
>
> Imports should be grouped in the following order:
>
> * Standard library imports.
> * Related third party imports.
> * Local application/library specific imports.
> You should put a blank line between each group of imports.

lib/pbench/server/api/__init__.py
```diff
@@ -72,375 +61,23 @@ def create_app():
         sys.exit(1)
```
portante, line 61: If we are going to turn this into a utility function, we want to remove the calls to `sys.exit()`, and raise an exception that the outer layer can call `sys.exit()`.  See also #1955.

lib/pbench/server/api/__init__.py
portante: If we can avoid it, in principle, we don't want to modify the code just to accommodate tests.  Let's have the test environment accommodate the code.

portante (changes requested):
Let's address these `sys.exit()` move.  It was fine in `create_app()` before because that was the outer later, but burying `sys.exit()` calls can be tricky to deal with long term.

lib/pbench/test/unit/server/conftest.py
```diff
@@ -83,12 +82,22 @@ def teardown():
...
     monkeypatch.setenv("_PBENCH_SERVER_CONFIG", cfg_file)
-    app = create_app()
-    app.testing = True
+
+    server_config = get_server_config()
+    return server_config
+
+
+@pytest.fixture
+def client(pytestconfig, monkeypatch, server_config):
```
dbutenhof, line 94: Since you refactored the only `monkeypatch` reference into the `server_config` fixture, I don't think you need `monkeypatch` here anymore. Nor probably `pytestconfig` for that matter...

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -28,36 +29,32 @@ def test_host_info(client, pytestconfig, caplog):
 
 class TestElasticsearch:
```
dbutenhof, line 30: We should consider breaking these test classes off into separate files associated with each of the new resource classes.

(I'm not really pointing a finger at you, since I was the first to break off the new `QueryControllers` and I was lazy about this... but we should still think about doing it!)

lib/pbench/test/unit/server/conftest.py
npalaska: yes you are right its redundant

lib/pbench/server/api/__init__.py
npalaska: yeah good point, will do

lib/pbench/test/unit/server/test_requests.py
npalaska: yep I was thinking about that, will just go ahead then

lib/pbench/test/unit/server/conftest.py
```diff
@@ -2,8 +2,7 @@
 import tempfile
 import pytest
 from pathlib import Path
-
-from pbench.server.api import create_app
+from pbench.server.api import create_app, get_server_config
```
portante, line 5: Not something to change here, but PEP 8 specifies that the groups of the different "kinds" of imports (sorted) are separated by a space.  But since `black` and `flake8` don't really catch that we can live with it.

So really the imports should look like:

```
import pytest
import tempfile

from pathlib import Path

from pbench.server.api import create_app, get_server_config
```

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+        return config.get("elasticsearch", "host")
+    except (NoOptionError, NoSectionError):
+        return ""
+
+
+def get_es_port(config):
+    try:
+        return config.get("elasticsearch", "port")
+    except (NoOptionError, NoSectionError):
+        return ""
```
dbutenhof, line 19: Are these actually used anywhere other than `get_es_url`?? It doesn't seem to me that they ought to be; in which case when don't we just elevate the URL to an `@property` on the config object in the first place and skip all this?

portante (approved):

lib/pbench/test/unit/server/test_requests.py
dbutenhof: Good start; we can unbundle the rest later. (And I suspect we'll want to generalize the `query_helper` fixture to mock behavior on the GraphQL and Elasticsearch pass-through APIs as well.)

dbutenhof (approved):

lib/pbench/test/unit/server/conftest.py
npalaska: I see, got it

lib/pbench/server/api/resources/query_apis/__init__.py
npalaska: These are elasticsearch properties and are very specific to the Elasticsearch modules we have and thus I removed it from `PbenchServerConfig` `@properties` and added it here as a common helper function that can be used by number of future query modules. `es_host` and `es_port` are not used independently anywhere else (at least yet) but having a separate function helps to know whether the port information is missing or the host information from the config file.

portante (changes requested):
Needs a rebase ...

lib/pbench/cli/server/shell.py
npalaska: Actually while testing now, I realized this would break. We run the gunicorn using the python subprocess and thus it create a separate gunicorn process and for that process the `application` would still remain `None` because it wont pick the same instance of application that we tried to set in the `main`

lib/pbench/cli/server/shell.py
```diff
@@ -1,14 +1,34 @@
...
+import sys
+
+from configparser import NoSectionError, NoOptionError
+
+from pbench.common.exceptions import BadConfig, ConfigFileNotSpecified
+from pbench.server.api import create_app, get_server_config
+
+
+try:
+    server_config = get_server_config()
```
portante, line 13: Do they really need to be the same server config object?

Why not have:

```
def app():
    return create_app(get_server_config())

def main():
    server_config = get_server_config()
    try:
        ...
```

portante (changes requested):

lib/pbench/cli/server/shell.py
npalaska: In that case we would be creating two server config object instances where one instance is used just for getting gunicorn host port and workers information, is it fine?

portante: Personally, I'd rather that than the global variable.  That global variable is created a module load time, which is not ideal and creates an inflexible interface.

dbutenhof (approved):

portante (changes requested):
Looks good!  Let's rebase and we can merge.

portante (dismissed):

dbutenhof (dismissed):

dbutenhof (approved):

portante (approved):

PR #1988, opened 2020-11-18
Fix spec file and ansible roles

ndokos:
When I simplified the spec file back in June, I went a bit too far
and lost the js files that we need for visualization. Some of these have
been added back: the rest are needed for the (very) old web-server version
only.

The firewall role added http and https services to the "permanent" setup
but neglected to reload the rules, so they were not valid for the current
boot (although they would work fine after a reboot :-) ).

The pbench-tarballs link from /var/www/html to the archive was not created
by the playbook (it was added manually to the production server). It is
now added by the playbook.

### DIFF
--- a/server/rpm/pbench-server.spec.j2
+++ b/server/rpm/pbench-server.spec.j2
@@ -62,11 +62,11 @@ rm -rf %{buildroot}
 mkdir -p %{buildroot}/%{installdir}
 cp -a ./server/* %{buildroot}/%{installdir}
 
-mkdir -p %{buildroot}/%{installdir}/html/static
-cp -a ./web-server/* %{buildroot}/%{installdir}/html/static
+mkdir -p %{buildroot}/%{installdir}/%{static}
+cp -a ./web-server/* %{buildroot}/%{installdir}/%{static}
 
 # for the npm install below
-mv %{buildroot}/%{installdir}/html/static/package.json %{buildroot}/%{installdir}
+mv %{buildroot}/%{installdir}/%{static}/package.json %{buildroot}/%{installdir}
 
 %post
 # Install python dependencies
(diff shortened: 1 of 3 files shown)


ndokos:
This should be ported to master and b0.70 as well. I'll open an issue and fix them up later.

portante (approved):

ndokos:
Rebased.

server/rpm/pbench-server.spec.j2
```diff
@@ -84,6 +84,17 @@ cd /%{installdir}
...
 
+# this only handles v0.3
+# AFAIK, we don't need all the other modules that package.json includes.
+jslist="node_modules/d3/d3.min.js
+        node_modules/d3-queue/build/d3-queue.min.js
+        node_modules/save-svg-as-png/lib/saveSvgAsPng.js"
+
+# copy them to the appropriate directories for v0.3
+for x in ${jslist} ;do
+    cp $x /%{installdir}/%{static}/js/v0.3
```
dbutenhof, line 95: Does there need to be an explicit create of the `/js/v0.3` directory?

ndokos: Looking at it more closely: it does not need an explicit create because the `cp -a` of the web-server stuff creates that subdirectory, but it is hidden, so it needs at least a comment (and maybe an explicit mkdir).

ndokos: Added both a comment and an explicit `mkdir -p` which *should* be a no-op.

dbutenhof: Thanks; the clarification is good.

dbutenhof (approved):

portante (approved):

Issue #1945, opened 2020-11-03
Add a simple "pbench set controller" API to change the name of the controller used on the pbench server


portante:
There are times when the user want to track the pbench result tar ball collected by the agent using a name other than the one determined by `hostname -f` and the `${full_hostname}` variable (see also #1933, and #1935).

RobertKrawitz:
Yup, this is what I need for bench-army-knife.

There should be a way to determine whether this API is available.

PR #1990, opened 2020-11-18
Add `--controller` option to `pbench-move-results`

portante:
This addresses the sentiment in issue #1945.

Adding a new switch, `--controller` to the `pbench-move-results` allows a user to set the controller name when pushed to the remote server, while preserving the controller name used originally.  We also take the opportunity to record the original short and long hostname values explicitly, and the host's IP address (output of `hostname -i`).

We also add support for the `PBENCH_CONTROLLER` environment variable for those that don't want to use the switch (similar to `--user`).

We also:

 * Fix `pbench-move-results` to take no actions remotely if there are no result directories to be moved
 * Fix the tabs vs spaces formatting in the three files related to `pbench-move-results`
 * Add a number of unit tests back that were left dangling, writing a check to make sure this doesn't happen in the future
 * Add additional unit tests to verify the new behaviors


portante:
@RobertKrawitz, see if this is an interface you can use.

agent/base.ut.pre
```diff
@@ -10,3 +10,6 @@ fi
 if [[ -z "${_pbench_full_hostname}" ]]; then
     export _pbench_full_hostname="testhost.example.com"
 fi
+if [[ -z "${_pbench_hostname_ip}" ]]; then
+    export _pbench_hostname_ip="42.42.42.42"
```
dbutenhof, line 14: While that's cute, I wonder if something that's not routed like 172.0.0.0 might be "safer"?

dbutenhof (approved):
Some good stuff in here aside from `--controller`, like the check for `id_rsa` and the sweep for "dangling" gold files.

One minor concern about the choice of garbage IPv4 as a "real" IP ... but looks good.

ndokos (dismissed):

agent/base.ut.pre
portante: Fixed.

agent/util-scripts/pbench-move-results
```diff
@@ -72,19 +80,39 @@ while true; do
...
+cd ${pbench_run} >/dev/null
+if [[ ${?} -ne 0 ]]; then
     error_log "ERROR: pbench local results directory does not exist: ${pbench_run}"
     exit 1
 fi
 
-if [[ ! -f "${pbench_bin}/id_rsa" ]]; then
-    error_log "ERROR: ${pbench_bin}/id_rsa required for moving results to archive host"
+tmp=${pbench_tmp}/${script_name}.${$}
+trap "rm -rf ${tmp}" EXIT INT QUIT
```
RobertKrawitz, line 91: Also trap SIGTERM.  Or do you want the directory to stay around if someone manually runs `kill`?

RobertKrawitz (changes requested):
I also notice an empty file  ...copy-results/test-34/pbench/pbench-user-benchmark_ndk-test-1_2019.01.01T12.00.42/.running; is that supposed to be in the PR?

RobertKrawitz:
Just a couple of little things, otherwise lgtm.

portante:
> I also notice an empty file ...copy-results/test-34/pbench/pbench-user-benchmark_ndk-test-1_2019.01.01T12.00.42/.running; is that supposed to be in the PR?

Yes, that is to verify that running benchmarks are not send to the server.

agent/base.ut.pre
```diff
@@ -10,3 +10,6 @@ fi
 if [[ -z "${_pbench_full_hostname}" ]]; then
     export _pbench_full_hostname="testhost.example.com"
 fi
+if [[ -z "${_pbench_hostname_ip}" ]]; then
+    export _pbench_hostname_ip="172.16.42.42"
```
dbutenhof, line 14: Ah, compromise; good work.

dbutenhof (approved):

dbutenhof (dismissed):

ndokos (approved):

npalaska (approved):
looks good

dbutenhof (approved):

Issue #1989, opened 2020-11-18
Port PR#1988 to master and b0.70


ndokos:
PR #1988 fixes a couple of bugs in the server RPM spec file and a couple of ansible roles. Those fixes are needed in master and in the b0.70 branch as well.

ndokos:
Fixed by #1994

PR #1994, opened 2020-11-18
Fix spec file and ansible roles

ndokos:
This is a forward port of PR #1988 (which was against the b0.69 branch) to master branch.

When I simplified the spec file back in June, I went a bit too far
and lost the js files that we need for visualization. Some of these have
been added back: the rest are needed for the (very) old web-server version
only.

The firewall role added http and https services to the "permanent" setup
but neglected to reload the rules, so they were not valid for the current
boot (although they would work fine after a reboot :-) ).

The pbench-tarballs link from /var/www/html to the archive was not created
by the playbook (it was added manually to the production server). It is
now added by the playbook.

### DIFF
--- a/server/ansible/roles/pbench-server-firewall/tasks/main.yml
+++ b/server/ansible/roles/pbench-server-firewall/tasks/main.yml
@@ -3,6 +3,7 @@
   firewalld:
     service: "{{ item }}"
     permanent: yes
+    immediate: yes
     state: enabled
   with_items:
     - http
--- a/server/ansible/roles/pbench-server-activate-httpd-setup-restart/tasks/main.yml
+++ b/server/ansible/roles/pbench-server-activate-httpd-setup-restart/tasks/main.yml
@@ -12,6 +12,14 @@
   notify:
     - restart apache
 
+- name: "Create pbench-tarballs link in doc root"
+  file:
+    src: "{{ pbench_dir }}/archive/fs-version-001"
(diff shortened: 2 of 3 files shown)


portante (approved):

riya-17 (dismissed):

portante (changes requested):
@ndokos, needs a rebase ...

portante (approved):

portante (changes requested):
@ndokos, needs a rebase due to other merges ...

ndokos:
Rebased.

portante (approved):

server/ansible/roles/pbench-server-firewall/tasks/main.yml
```diff
@@ -3,6 +3,7 @@
   firewalld:
     service: "{{ item }}"
     permanent: yes
+    immediate: yes
     state: enabled
   with_items:
     - http
```
dbutenhof, line 9: Just as a note (doesn't need to be worked out now), but we're going to need to punch a hole for the API port too, currently 8001 by default -- but we should create a fact to pull this from the target config file.

I also wonder whether it'd be better to set up a proxy configuration that will let us split the /api URI path from other accesses and drive everything off http/https without using a separate port, which would make more sense.

(Further note, as we move forward to user authentication, we'll probably want to drop http entirely.)

dbutenhof (approved):

ndokos:
Fixes #1989.

PR #1996, opened 2020-11-18
Fix SELinux labels on the pbench directory structure

ndokos:
We need to use `semanage` to set the correct labels on the top level
pbench directory and its children: most of them get `default_t` but
`archive` and `public_html` and their children get `httpd_sys_content_t`.

The sefcontext ansible module is needed for this: it's available from
Galaxy:

    ansible-galaxy collection install community.general

The rest of the tasks create the directory structure (I deleted the
setype: setting since it does not survive a restorecon - I now assume
that it is just doing the equivalent of a chcon). At the end we run a
recursive restorecon to set the labels on everything.

### DIFF
--- a/server/ansible/roles/pbench-server-activate-create-results-dir-structure/tasks/main.yml
+++ b/server/ansible/roles/pbench-server-activate-create-results-dir-structure/tasks/main.yml
@@ -1,6 +1,43 @@
 ---
 # activate the server - create results dir structure
 
+# The sefcontext module is available from Galaxy with
+#    ansibble-galacy collection install community.general
+#
+# We have two tasks to set the SELinux context with semanage on the
+# top level directory (typically /pbench or /srv/pbench) and its
+# subdirectories: they get a default_t type.  We then override the
+# type for the `archive' and `public_html' subdirs (and their subdirs)
+# which must be available through HTTP: they get a httpd_sys_content_t
+# type.
+
+# The rest of the tasks create the directory structure under the top-level
+# directory. Then at the end, we run a restorecon to give every directory
+# its proper label.
+
+- name: add semanage fcontexts - top level dir gets default_t and subdirs inherit
+  community.general.sefcontext:
+    target: "{{ pbench_dir }}(/.*)?"
+    setype: default_t
+    state: present
+
+- name: add semanage fcontexts - public_html and archive get httpd_sys_content_t
+  community.general.sefcontext:
+    target: "{{ pbench_dir }}/{{ item}}(/.*)?"
+    setype: httpd_sys_content_t
+    state: present
+  with_items:
+    - public_html
+    - archive
+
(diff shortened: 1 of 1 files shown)


server/ansible/roles/pbench-server-activate-create-results-dir-structure/tasks/main.yml
```diff
@@ -51,3 +87,7 @@
     state: directory
   with_items:
     - "002"
+
+- name: Apply new SELinux file context to filesystem
+  command:
+    cmd: "restorecon -irv {{ pbench_dir }}"
```
portante, line 93: The only problem with this command is for existing installs which lots and lots of data already present.  This will take eons in some settings.

Maybe what we can do is store the top of the hierarchies, and then start some background command to finish the task?

Not sure exactly how to handle this.

portante (changes requested):

server/ansible/roles/pbench-server-activate-create-results-dir-structure/tasks/main.yml
dbutenhof: Yes; retrofitting is going to be a problem.

But it's not like we have a lot of deployed servers that need fixing. Maybe we just individually set the directories the installer creates, and for now rely on manually fixing the rest of the tree? This will all go away when/if we move completely to S3 storage.

ndokos: I can do a non-recursive `restorecon` on the top level directories. We depend on inheritance in any case to set the SELinux context. If we start things right. then it should just work(TM).

Instead of worrying about any potential mislabeling here, we can add the check to the server audit and if that ever detects a problem, we fix it by hand. Assuming that't it's a rare (or never) occurrence, that should be OK. If it happens more often, then we add a relabeling cron job.

WDYT?

server/ansible/roles/pbench-server-activate-create-results-dir-structure/tasks/main.yml
```diff
@@ -1,14 +1,50 @@
 ---
 # activate the server - create results dir structure
 
+# The sefcontext module is available from Galaxy with
+#    ansibble-galacy collection install community.general
```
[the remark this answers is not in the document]
ndokos: Thanks! Fixed.

server/ansible/roles/pbench-server-activate-create-results-dir-structure/tasks/main.yml
ndokos: I got rid of the -r. The auditing and fixing are not addressed here. I'll open an issue.

portante (approved):

dbutenhof (approved):

PR #1997, opened 2020-11-19
Remove the pbench user and group from the agent

portante:
The pbench agent currently runs as root, and we don't really have the notion of
separate user for the agent ... yet.

### DIFF
--- a/agent/ansible/pbench/agent/roles/pbench_agent_config/README.md
+++ b/agent/ansible/pbench/agent/roles/pbench_agent_config/README.md
@@ -21,8 +21,6 @@ These are the variables that are defined by default and do not
 generally need to be modified:
 
 - pbench_agent_install_dir: the installation directory for the pbench agent (default: /opt/pbench-agent)
-- pbench_owner: the owner of all the files in the installation directory (default: pbench)
-- pbench_group: the group of all the files in the installation directory (default: pbench)
 - pbench_config_files: the list of config files to be installed (default '["pbench-agent.cfg"]')
 - pbench_config_dest: the location where the config file(s) are to be installed (default: {{ pbench_agent_install_dir }}/config)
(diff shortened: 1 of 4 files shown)


ndokos (commented):
Four questions:

- Why is this a problem?
- Aren't the user and group created properly? 
- After this, all the installed stuff will be owned by root?
- If so, isn't that a step in the wrong direction, since we are trying to get away from root running everything? It is usual practice, I think, to have a separate (non-root) user owning the package's contents.

portante:
> Four questions:
> 
> * Why is this a problem?

Because we apparently have issues with the user and group not being properly applied in all cases, and rather than debugging that problem, this PR proposes we remove them entirely since they are not needed in any way.

> * Aren't the user and group created properly?

From the reports we are receiving, no.

> * After this, all the installed stuff will be owned by root?

Yes.

> * If so, isn't that a step in the wrong direction, since we are trying to get away from root running everything?  It is usual practice, I think, to have a separate (non-root) user owning the package's contents.

Not sure about that practice, as most RPMs install their code all over the place as root.  Then the user can use them.

Our problem is that we are half way between two worlds.

The software should be installed in the proper location as `root`, but the configuration information should be user specific.   We have the software installed as user `pbench`, but the configuration information is specific to `root`.

The first step to fixing this whole problem is to stop using a "user", installing the software as `root`, acknowledge that today the pbench-agent only works for `root`, and then subsequently fix the installation so that the configuration is per user.

dbutenhof (approved):
Seems right; most packages install globally under `/opt` as root __or__ under a user's `~/.local` for private use, not as some arbitrary user that's neither there nor here.

npalaska (approved):

ndokos (approved):

PR #1999, opened 2020-11-19
Add coverage support

zulcss:
Add the ability to run pytest coverage report.

### DIFF
--- a/jenkins/run-pytests
+++ b/jenkins/run-pytests
@@ -23,5 +23,11 @@ pytest_sts=${?}
 if [[ ${pytest_sts} -ne 0 ]]; then
     printf -- "FAILED: pytest ...\n" >&2
 fi
-final_sts=$(( black_sts + flake_sts + pytest_sts))
+printf -- "\nPython coverage ...\n"
+coverage report
+coverage_sts=${?}
+if [[ ${coverage_sts} -ne 0 ]]; then
+    printf -- "FAILED: coverage .\n" >&2
+fi
+final_sts=$(( black_sts + flake_sts + pytest_sts + coverage_sts))
 exit ${final_sts}
--- a/tox.ini
+++ b/tox.ini
@@ -34,6 +34,7 @@ whitelist_externals =
     find
     env
 
+
 [testenv:server]
 description = Runs all non-Python3-based server unit/functional tests
 passenv = PBENCH_UNITTEST_SERVER_MODE
@@ -90,6 +91,7 @@ commands =
     pytest ./lib/pbench/test/unit/common
     pytest ./lib/pbench/test/unit/server 
     pytest ./lib/pbench/test/functional
+    coverage report --include "lib/pbench/server/*,lib/pbench/common/*"
 
 [testenv:py3-agent]
 description = Runs all Python3-based agent unit and functional tests
@@ -99,7 +101,8 @@ deps =
 commands =
     pytest ./lib/pbench/test/unit/common
     pytest ./lib/pbench/test/unit/agent
-    pytest -s ./lib/pbench/test/functional
+    pytest -s ./lib/pbench/test/functional/agent
+    coverage report --include "lib/pbench/agent/*,lib/pbench/cli/*,lib/pbench/common/*"
 
 [testenv:lint]
 description = Runs all linting tasks
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,5 @@ pip-wheel-metadata
 .vscode
 Pipfile
 Pipfile.lock
+.coverage
+cov.xml


jenkins/run-pytests
```diff
@@ -5,9 +5,12 @@ progdir="$(realpath -e $(dirname "${0}")/..)"
 
 source $(dirname "${0}")/python-setup.sh
 
+printf -- "unit test coverage ...\n"
+pytest --cov --cov-append --cov-report=term-missing
```
portante, line 9: Can you merge these options with line 15?

tox.ini
```diff
@@ -110,6 +112,16 @@ deps = -r{toxinidir}/lint-requirements.txt
...
 usedevelop = false
 
+[testenv:coverage]
+description = Generate HTML coverage report
+deps =
+    -r{toxinidir}/agent/requirements.txt
+    -r{toxinidir}/server/test-requirements.txt
+commands =
+   coverage erase
+   pytest --cov --cov-append --cov-report=term-missing .
```
portante, line 122: Can we add these to `py3-server` and `py3-agent`?

If we are going to have a separate `tox` environment, can you run the agent and the server tests explicitly?  And fix the requirements?

portante (changes requested):
Great!

dbutenhof (commented):
I don't get why we're not just reporting coverage on our normal test runs; doing it as a separate detached pass doesn't seem efficient or effective.

tox.ini
zulcss: Because you wouldnt want to gate on python coverage because there is alot of coverage missing. When we get to that point then we can think of doing that.

dbutenhof: We can't just REPORT it without GATING success on it?

jenkins/run-pytests
```diff
@@ -10,4 +10,10 @@ black --check .
 printf -- "\nflake8 --verbose ...\n"
 flake8 --verbose .
 printf -- "\n"
-pytest ./lib/pbench/test/unit/common ./lib/pbench/test/unit/agent ./lib/pbench/test/unit/server ./lib/pbench/test/functional
+coverage erage
```
portante, line 13: `s/erage/erase/`

tox.ini
portante: I just tried the following change, and the py3-server tox environment passes, but the py3-agent fails because the tests are actually broken (command-line arguments of pytest are being used by the test itself).

The change below dumps the coverage reports into separate files:
```
[pportant@gandalf pbench]$ git diff
     1  diff --git a/tox.ini b/tox.ini
     2  index 07ed8ea89..40634f77a 100644
     3  --- a/tox.ini
     4  +++ b/tox.ini
     5  @@ -87,9 +87,13 @@ deps =
     6       -r{toxinidir}/server/requirements.txt
     7       -r{toxinidir}/server/test-requirements.txt
     8   commands =
     9  -    pytest ./lib/pbench/test/unit/common
    10  -    pytest ./lib/pbench/test/unit/server
    11  -    pytest ./lib/pbench/test/functional
    12  +    pytest --cov --cov-append \
    13  +      --cov-report html:cov_html \
    14  +      --cov-report xml:cov.xml \
    15  +      --cov-report annotate:cov_annotated \
    16  +      ./lib/pbench/test/unit/common \
    17  +      ./lib/pbench/test/unit/server \
    18  +      ./lib/pbench/test/functional
    19
    20   [testenv:py3-agent]
    21   description = Runs all Python3-based agent unit and functional tests
    22  @@ -97,9 +101,13 @@ deps =
    23       -r{toxinidir}/agent/requirements.txt
    24       -r{toxinidir}/agent/test-requirements.txt
    25   commands =
    26  -    pytest ./lib/pbench/test/unit/common
    27  -    pytest ./lib/pbench/test/unit/agent
    28  -    pytest -s ./lib/pbench/test/functional
    29  +    pytest --cov --cov-append \
    30  +      --cov-report html:cov_html \
    31  +      --cov-report xml:cov.xml \
    32  +      --cov-report annotate:cov_annotated \
    33  +      ./lib/pbench/test/unit/common \
    34  +      ./lib/pbench/test/unit/agent \
    35  +      ./lib/pbench/test/functional
    36
    37   [testenv:lint]
    38   description = Runs all linting tasks
```

portante (changes requested):
Seems like we can hide the reports.  They do take a bit longer to run, but seems worth it.  What do folks think?

portante:
See also https://github.com/nedbat/coveragepy/pull/1070, where Ned is using the [codecov](https://github.com/marketplace/codecov) site.

jenkins/run-pytests
```diff
@@ -25,3 +25,9 @@ if [[ ${pytest_sts} -ne 0 ]]; then
 fi
 final_sts=$(( black_sts + flake_sts + pytest_sts))
 exit ${final_sts}
+printf -- "\nPython coverage ...\n"
```
portante, line 28: Not sure how this runs with the `exit` at line 27...

jenkins/run-pytests
```diff
@@ -25,3 +25,9 @@ if [[ ${pytest_sts} -ne 0 ]]; then
...
+pytest --cov=pbench
```
portante, line 29: Why not add `--cov=pbench` to line 21 where we specify an `coverage.html` file to be created so as not to pollute the logs.  Then we can modify Jenkins to capture the `coverage.html` file as an artifact of the build.

tox.ini
```diff
@@ -90,6 +92,13 @@ commands =
     pytest ./lib/pbench/test/unit/common
     pytest ./lib/pbench/test/unit/server 
     pytest ./lib/pbench/test/functional
+    pytest --cov --cov-append \
+        --cov-report html:cov_html \
+        --cov-report xml:cov.xml \
+        --cov-report annotate:cov_annotated \
```
portante, line 98: This might be too much, sorry for suggesting it.  I adds `.cov` files to everything ...

tox.ini
```diff
@@ -110,6 +126,16 @@ deps = -r{toxinidir}/lint-requirements.txt
 skip_install = true
 usedevelop = false
 
+[testenv:coverage]
+description = Generate HTML coverage report
+deps =
+    -r{toxinidir}/agent/requirements.txt
```
portante, line 132: Shouldn't we have all the agent and server requirement files loaded?  For example:

```
deps =
    -r{toxinidir}/agent/requirements.txt
    -r{toxinidir}/agent/test-requirements.txt
    -r{toxinidir}/server/requirements.txt
    -r{toxinidir}/server/test-requirements.txt
```

portante (changes requested):

dbutenhof (commented):
Aside from what Pete already caught, this is looking pretty good. No additional comments.

jenkins/run-pytests
```diff
@@ -23,5 +23,11 @@ pytest_sts=${?}
 if [[ ${pytest_sts} -ne 0 ]]; then
     printf -- "FAILED: pytest ...\n" >&2
 fi
-final_sts=$(( black_sts + flake_sts + pytest_sts))
+printf -- "\nPython coverage ...\n"
+coverage report
```
portante, line 27: If we don't have `pytest --cov`, or the like, above, from what coverage data will this report be generated?

portante (changes requested):

jenkins/run-pytests
zulcss: From the entire code base, we are running all the unit tests anyways.

portante: How do we run the unit tests with coverage if we are not using `pytest --cov`?  See also PR https://github.com/zulcss/pbench/pull/3.

tox.ini
```diff
@@ -13,6 +13,7 @@ envlist = lint
           bench-scripts
           py3-server
           server
+          coverage
```
portante, line 16: Where is the definition of the `coverage` tox environment?

portante (changes requested):

tox.ini
```diff
@@ -90,6 +91,7 @@ commands =
     pytest ./lib/pbench/test/unit/common
     pytest ./lib/pbench/test/unit/server 
     pytest ./lib/pbench/test/functional
```
portante, line 93: Seems like we need this to be `pytest ./lib/pbench/test/functional/server`.

tox.ini
```diff
@@ -99,7 +101,8 @@ deps =
 commands =
     pytest ./lib/pbench/test/unit/common
     pytest ./lib/pbench/test/unit/agent
-    pytest -s ./lib/pbench/test/functional
+    pytest -s ./lib/pbench/test/functional/agent
```
portante, line 104: Why do we need the `-s` option?

tox.ini
```diff
@@ -99,7 +101,8 @@ deps =
...
+    coverage report --include "lib/pbench/agent/*,lib/pbench/cli/*,lib/pbench/common/*"
```
portante, line 105: In both cases, the above `coverage` commands don't work:

```
py3-agent run-test: commands[3] | coverage report --include 'lib/pbench/agent/*,lib/pbench/cli/*,lib/pbench/common/*'
No data to report.
```

portante (changes requested):

portante:
PR #2050 supersedes this PR.

PR #2002, opened 2020-11-19
Phase 2 of generate-pbench-timeseries-graphs:

RobertKrawitz:
* Extract .csv files from tarballs (must be local right now)

* Optional specify files on command line

* Add -m/-M for file name patterns that must/must not match

* Add -L max label length

* Add -O output directory

* Place .png files above .csv directory

* Fix a number of errors in the pattern list

* Skip non-tool .csv files more cleanly

### DIFF
--- a/contrib/analysis/generate-pbench-timeseries-graphs
+++ b/contrib/analysis/generate-pbench-timeseries-graphs
@@ -37,268 +37,287 @@
 # 11->Graham's number ...
 
 from __future__ import print_function
-import re
-import sys
+from pathlib import Path
 import argparse
 import math
+import re
+import subprocess
+import sys
+import tarfile
 
 parser = argparse.ArgumentParser(description='Generate graphs from raw Pbench tarball')
-parser.add_argument('-s', "--size", help='Graph size (x,y)', metavar='Size', default='1280,1024')
-parser.add_argument('-w', '--linewidth', help='Line width', metavar='Width', default=2, type=int)
-parser.add_argument('--max-elements', help='Maximum elements to plot', metavar='Max Elements', default=0, type=int)
+parser.add_argument('-s', "--size", help='Graph size (x,y)', metavar='Graph size (x,y)', default='1280,1024')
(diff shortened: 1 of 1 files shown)


portante (commented):
Looks good, let me know when its ready for merging by pulling it out of `Draft` mode.

contrib/analysis/generate-pbench-timeseries-graphs
```diff
@@ -59,258 +59,264 @@ parser.add_argument('files', metavar='file', type=str, nargs="*", help="files")
 args = parser.parse_args()
 
 pattern_map = {
-    re.compile('disk_IOPS.csv$') : {
+    re.compile('disk_IOPS\\.csv$') : {
```
portante, line 62: Using "r" strings might help here, see https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals

portante: We also have code like this already in the pbench server: https://github.com/distributed-system-analysis/pbench/blob/master/lib/pbench/server/indexer.py#L1847

portante (commented):
Looking good ...

contrib/analysis/generate-pbench-timeseries-graphs
RobertKrawitz: That code looks close to what I want.  I'd also like to have a title, graph type (or is that implied in any of the other graphs?), and optional axis labels.  But it would certainly make sense to reuse it.

RobertKrawitz: And I fixed up the regexps.

RobertKrawitz:
@portante would you prefer that I collapse the commits?

portante (changes requested):
Collapse to one commit and we'll merge.  Thanks!

RobertKrawitz:
Collapsed.

portante (approved):

PR #2004, opened 2020-11-20
Fix the Python 3 logger to handle time properly

### LINKED ISSUE #2003: `_PbenchLogFormatter.format()` doesn't handle existing `asctime` property.
This seems to have been an old problem that something new I pulled in (possibly requests_mock) is now triggering in my `query_controllers.py` context. The `_PbenchLogFormatter.format()` method tries to format UTC time from a current value of `record.asctime` if the property exists. However, if it does exist, it should be a string, which causes `datetime.utcfromtimestamp()` to throw `TypeError` ... and as this is unhandled, Python displays traceback.

Because the `asctime` property is reserved and defined, but not populated until it hits a formatter, this suggests that we've never before run with a competing formatter that added `asctime`, and we've always taken the `AttributeError` and generated a fresh timestamp. Now, something else is populating `asctime`, with a string, and we're taking the `TypeError` instead,.

The most likely fix would seem to be to format using `record.created`, which is the `time.time()` value captured when the `LogRecord` was created.

portante:
While working PR #1914, we found that our log formatter was not handling records that already had a formatted `asctime` field.  When we compared the code against what the default Python 3 (3.6 version used, the lowest version we support), we found that we were not handling this properly.

This change updates the code to more closely match the behavior of the 3.6 logger using the logging interfaces as intended.  Well, with one exception, we ignore the specified `datafmt` and always use UTC time in the ISO format.

Fixes #2003.

### DIFF
--- a/lib/pbench/common/logger.py
+++ b/lib/pbench/common/logger.py
@@ -81,17 +81,21 @@ class _PbenchLogFormatter(logging.Formatter):
     def __init__(self, fmt=None, datefmt=None, style="{", max_line_length=0):
         super().__init__(fmt=fmt, datefmt=datefmt, style=style)
         self.max_line_length = max_line_length
+        self.converter = datetime.utcfromtimestamp
+
+    def formatTime(self, record, datefmt=None):
+        """
+        Return the creation time of the specified LogRecord as formatted text.
+        """
+        return self.converter(record.created).isoformat()
 
     def format(self, record):
         # Included from Python's logging.Formatter and then altered slightly to
         # replace \n with #012
         record.message = record.getMessage()
-        if self._fmt.find("{asctime}") >= 0:
-            try:
-                record.asctime = datetime.utcfromtimestamp(record.asctime).isoformat()
-            except AttributeError:
-                record.asctime = datetime.now().isoformat()
-        msg = (self._fmt.format(**record.__dict__)).replace("\n", "#012")
+        if self.usesTime():
+            record.asctime = self.formatTime(record, self.datefmt)
+        msg = self.formatMessage(record).replace("\n", "#012")
         if record.exc_info:
             # Cache the traceback text to avoid converting it multiple times
             # (it's constant anyway)
@@ -103,6 +107,10 @@ class _PbenchLogFormatter(logging.Formatter):
             if not msg.endswith("#012"):
                 msg = msg + "#012"
             msg = msg + record.exc_text
+        if record.stack_info:
+            if not msg.endswith("#012"):
+                msg = msg + "#012"
+            msg = msg + self.formatStack(record.stack_info).replace("\n", "#012")
         if self.max_line_length > 0 and len(msg) > self.max_line_length:
             if self.max_line_length < 7:
                 msg = msg[: self.max_line_length]


portante:
This is marked as a `Draft` because I had a weird failure in my local test that seems unrelated, but need to investigate.

lib/pbench/common/logger.py
```diff
@@ -81,17 +81,21 @@ class _PbenchLogFormatter(logging.Formatter):
     def __init__(self, fmt=None, datefmt=None, style="{", max_line_length=0):
         super().__init__(fmt=fmt, datefmt=datefmt, style=style)
         self.max_line_length = max_line_length
+        self.converter = datetime.utcfromtimestamp
+
+    def formatTime(self, record, datefmt=None):
+        """
+        Return the creation time of the specified LogRecord as formatted text.
+        """
+        return self.converter(record.created).isoformat()
```
dbutenhof, line 90: This is a bit interesting as we're still bypassing the `datefmt` setting and forcing `isoformat()`. I'm not objecting, since I think that's what we want to see... but I I'm also not sure this is really any cleaner than simply doing the formatting inline in `format()`. Also, is the default `converter` supposed to do UTC, or are we now generating local time? The documentation isn't entirely clear on that point; but, again, this isn't necessarily equivalent to what we had before. Should we be overriding `converter` also, as I mentioned in #2003??

dbutenhof (approved):
Approve with minor concerns: this fixes the bug and I think does what we want, but you changed to use `formatTime` because it's "cleaner and more standard" but we're still only going part of the way and I'm not sure it's really "cleaner" than doing it inline because of that. Should we go further?

lib/pbench/common/logger.py
portante: > This is a bit interesting as we're still bypassing the `datefmt` setting and forcing `isoformat()`. I'm not objecting, since I think that's what we want to see... but I I'm also not sure this is really any cleaner than simply doing the formatting inline in `format()`.

It is not cleaner or simpler, but it is quite comparable.  If you were to diff the two routines side-by-side, the differences are in the newline character handling, adding the `.replace()` calls, and in the trimming of the maximum message size.

> Also, is the default `converter` supposed to do UTC, or are we now generating local time? The documentation isn't entirely clear on that point; but, again, this isn't necessarily equivalent to what we had before. Should we be overriding `converter` also, as I mentioned in #2003??

I am not sure I follow these questions.  Aren't we now overriding the `converter` at line 84?

portante (commented):
Answered review questions in-line.

lib/pbench/common/logger.py
dbutenhof: > > This is a bit interesting as we're still bypassing the `datefmt` setting and forcing `isoformat()`. I'm not objecting, since I think that's what we want to see... but I I'm also not sure this is really any cleaner than simply doing the formatting inline in `format()`.
> 
> It is not cleaner or simpler, but it is quite comparable. If you were to diff the two routines side-by-side, the > I am not sure I follow these questions. Aren't we now overriding the `converter` at line 84?

Well, ouch; I apparently completely missed that line when I was reading it through!

OK, so we've got our UTC; we're still ignoring `datefmt` and forcing ISO time format. It's less general, but I'm not particularly concerned about that. I'm satisfied.

ndokos (approved):

PR #2008, opened 2020-11-25
Add support for overriding TM bind hostname

portante:
We add support for a caller to override the bind host name used by the Redis Server and Tool Data Sink using the `PBENCH_TM_BIND_HOSTNAME` environment variable.

We also ensure the list of hostnames used by the Redis Server do not map to the same IP address.

### DIFF
--- a/agent/util-scripts/pbench-tool-meister-start
+++ b/agent/util-scripts/pbench-tool-meister-start
@@ -22,6 +22,7 @@ import json
 import logging
 import os
 import signal
+import socket
 import sys
 import time
 
@@ -288,7 +289,7 @@ def main(argv):
         full_hostname = os.environ["_pbench_full_hostname"]
     except Exception:
         logger.exception("failed to fetch parameters from the environment")
-        return 1
+        return 2
     else:
         tm_dir = Path(benchmark_run_dir, "tm")
         try:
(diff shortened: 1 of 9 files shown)


portante:
@RobertKrawitz, I think this should solve the bind host problem we debugged.

ndokos (approved):

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -288,28 +289,68 @@ def main(argv):
...
     else:
-        hostnames = f"localhost {full_hostname}"
+        # Add to the list of host names the Redis server will use.
+        hostnames_l.append("localhost")
+    try:
+        tm_bind_hostname_ip = socket.gethostbyname(tm_bind_hostname)
+    except socket.error:
+        # The given Tool Meister host name does not map to an IP address, so
+        # we can't use it.
+        tm_bind_hostname_ip = None
```
dbutenhof, line 328: The assignment isn't necessary as `tm_bind_hostname_ip` isn't referenced outside the `else`. LGTM is complaining. However, a log might be handy here? In any case, you should fix the LGTM alert.

dbutenhof (changes requested):

agent/util-scripts/pbench-tool-meister-start
npalaska, line 328: I think this line wont do anything since we dont have a `finally` block

npalaska (approved):

dbutenhof (dismissed):

dbutenhof (approved):

dbutenhof (approved):

ndokos (approved):

PR #2005, opened 2020-11-21
Merge coll sysinfo metadata log

portante:
Merge collect sysinfo and metadata log
====

Merge `pbench-collect-sysinfo` and `pbench-metadata-log` operations and behaviors into the Tool Meister framework.

To make it easier to review, this PR was broken into 4 commits:

 * All the changes to the `agent/bench-scripts` directory without unit tests (14 files)
 * All the changes to the `agent/util-scripts` and `lib/pbench/agent` hierarchies (23 files)
 * UML Sequence Diagram updates (1 file)
 * All testing related updates (100+ files)

Once reviewed, we merged them back down to a single commit.

## Description of combined changes

The last set of `ssh` commands issued by the pbench-agent that are not involved in orchestrating the Tool Meisters, outside of those issued by the `bench-scripts`, are for `hostname -s` data collection on all the registered remote tool hosts.

The Tool Meisters are now given the responsibility for collecting information about their environment, including the version of the pbench agent, and a more comprehensive collection of host name information, returning that with their startup acknowledgement payload.  This change also addresses a previous bug where registered hosts labels were not being used in the on-disk data capture directories as they were before.

This collected data set is now recorded by the Tool Data Sink, after restructuring the action flow so that clients send action playloads to the Tool Data Sink, which then forwards to the Tool Meisters.  This gives the Tool Data Sink the ability to collect all the Tool Meister information and write it to disk before acknowledging to the client that startup is complete.

We add an entirely new section in the `metadata.log` file which is per-host/per-tool, recording the tool options, and the output from the tool install checks (if any).  This section supercedes the previous tool options in the per-host sections, though we leave those options there for compatibility for now.

We add graceful handling of bad PUT requests from Tool Meisters to the Tool Data Sink.

With this refactoring, we also collapse all the "sysinfo", "init", and "end" actions into the respective `pbench-tool-meister-start` and `pbench-tool-meister-stop` interfaces to simplify the CLI behaviors.

To make this possible, we add a new `Client` API used by the CLI interfaces `pbench-tool-meister-start` and `pbench-tool-meister-stop`.

We have also undertaken a major refactoring of `pbench-sysinfo-dump` to remove the dependency on `base` for environment variables.  As a result of this work, we drop the stockpile configuration entirely, as it is no longer required.

We have enhanced the Tool Meister tests to use persistent tools, and enhanced the unit test framework to properly clean up the Tool Meister test environment when tests fail.

[...]

PR #2014, opened 2020-12-02
Fix logging for Tool Meisters

portante:
The debug-ability of the Tool Meisters running remotely is very difficult without a way to properly capture log output from them.  We take three major steps to address this.

The first step is to not use a Python logger until we will (or would have otherwise) daemonized the process.  This allows us to emit error messages on stderr, which will be exposed in the callers context.  Once we are daemonized, `stdout` and `stderr` are detached from the caller, so logging to some place other than `stdout`/`stderr` is the best option from that point forward.

The second step is to recognize that logging to a file on a remote host is useless unless the operator can get a hold of that log file. However, remote Tool Meisters use a temporary directory, so writing to a file will just be lost when the temporary directory is removed by the process of shutting down the Tool Meister.  Instead, we now log to `/dev/log` when daemonized.  We then add the ability to run the Tool Meister without daemonizing so that in a container we can log to `stdout`/`stderr` to leverage the normal container logging mechanism.

As a result of making this change, we introduce a method to request the Tool Meister not be daemonized, and re-organize the code a bit so that we can take advantage of those two environments.

The third step is to add the ability for each Tool Meister to send logs back to the Tool Data Sink via Redis, so that the Tool Data Sink can collect them into a single log file making it easier to see what happened remotely.  If the Redis server logging fails, then the operator can fall back to looking for Tool Meister logs in the container logs or in the remote host's normal syslog sink.

Finally, we address dropped connections to the Redis server so that both the Tool Meister and the Tool Data Sink handle that a bit more gracefully.

----

The unit tests for the Tool Meister code (`util-scripts/tests-53,6,7`) are a bit flakey with this code because the Tool Meisters and the Tool Data Sink are not synchronized.  With PR #2005 we have the Tool Meisters terminated by the Tool Data Sink so the flakiness goes away.

### DIFF
--- a/agent/util-scripts/gold/test-client-tool-meister/test-56.txt
+++ b/agent/util-scripts/gold/test-client-tool-meister/test-56.txt
@@ -158,17 +158,14 @@ Collecting system information
 /var/tmp/pbench-test-utils/pbench/mock-run/tm/redis.conf
 /var/tmp/pbench-test-utils/pbench/mock-run/tm/redis.log
 /var/tmp/pbench-test-utils/pbench/mock-run/tm/tm-lite-remote_a.example.com.err
-/var/tmp/pbench-test-utils/pbench/mock-run/tm/tm-lite-remote_a.example.com.log
 /var/tmp/pbench-test-utils/pbench/mock-run/tm/tm-lite-remote_a.example.com.out
 /var/tmp/pbench-test-utils/pbench/mock-run/tm/tm-lite-remote_b.example.com.err
-/var/tmp/pbench-test-utils/pbench/mock-run/tm/tm-lite-remote_b.example.com.log
 /var/tmp/pbench-test-utils/pbench/mock-run/tm/tm-lite-remote_b.example.com.out
 /var/tmp/pbench-test-utils/pbench/mock-run/tm/tm-lite-remote_c.example.com.err
-/var/tmp/pbench-test-utils/pbench/mock-run/tm/tm-lite-remote_c.example.com.log
 /var/tmp/pbench-test-utils/pbench/mock-run/tm/tm-lite-remote_c.example.com.out
 /var/tmp/pbench-test-utils/pbench/mock-run/tm/tm-lite-testhost.example.com.err
-/var/tmp/pbench-test-utils/pbench/mock-run/tm/tm-lite-testhost.example.com.log
 /var/tmp/pbench-test-utils/pbench/mock-run/tm/tm-lite-testhost.example.com.out
+/var/tmp/pbench-test-utils/pbench/mock-run/tm/tm.logs
 /var/tmp/pbench-test-utils/pbench/mock-run/tools-lite
 /var/tmp/pbench-test-utils/pbench/pbench.log
 /var/tmp/pbench-test-utils/pbench/tmp
@@ -232,7 +229,6 @@ INFO pbench-tool-data-sink log_request -- 127.0.0.1 - - "PUT /tool-data/4a200d14
 INFO pbench-tool-data-sink log_request -- 127.0.0.1 - - "PUT /tool-data/9e243daecdc01b52d608090bb1c49d95/remote_a.example.com HTTP/1.1" 200 0
 INFO pbench-tool-data-sink log_request -- 127.0.0.1 - - "PUT /tool-data/9e243daecdc01b52d608090bb1c49d95/remote_b.example.com HTTP/1.1" 200 0
 INFO pbench-tool-data-sink log_request -- 127.0.0.1 - - "PUT /tool-data/9e243daecdc01b52d608090bb1c49d95/remote_c.example.com HTTP/1.1" 200 0
-INFO pbench-tool-data-sink main -- Remove pid file ... (pbench-tool-data-sink.pid)
 INFO pbench-tool-data-sink run -- Bottle web server exited
 INFO pbench-tool-data-sink run -- Running Bottle web server ...
 --- mock-run/tm/pbench-tool-data-sink.log file contents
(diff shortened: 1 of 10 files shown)


RobertKrawitz:
Keep in mind that `/dev/log` may not exist in a container.  Running it without daemonizing seems like a good approach, as does sending logs back to the tool data sink.

portante:
> Keep in mind that `/dev/log` may not exist in a container. Running it without daemonizing seems like a good approach, as does sending logs back to the tool data sink.

Right, the goal is that the daemonize mode would only be used by the CLI when not using containers.  That is where the logging using "syslog".  The daemonize mode implies console logging so that logs would flow to the container log capture mechanism, and "syslog" would not be used.

In either case, we also try to send logs back to the Tool Data Sink so that the behaviors are captured as part of the performance data in case they are useful later to understand what data was collected.

RobertKrawitz:
> > Keep in mind that `/dev/log` may not exist in a container. Running it without daemonizing seems like a good approach, as does sending logs back to the tool data sink.
> 
> Right, the goal is that the daemonize mode would only be used by the CLI when not using containers. That is where the logging using "syslog". The daemonize mode implies console logging so that logs would flow to the container log capture mechanism, and "syslog" would not be used.

Sounds perfect.  Thanks!

agent/util-scripts/gold/test-client-tool-meister/test-56.txt
```diff
@@ -341,11 +322,70 @@ INFO pbench-tool-meister wait -- waiting for stop mpstat
...
 INFO pbench-tool-meister sysinfo -- pbench-sysinfo-dump -- /var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/pbench-sysinfo-dump /var/tmp/pbench-test-utils/pbench/mock-run/sysinfo/end block,security_mitigations,sos 
 INFO pbench-tool-meister sysinfo -- testhost.example.com: sysinfo send (no-op) lite /var/tmp/pbench-test-utils/pbench/mock-run/sysinfo/end
-INFO pbench-tool-meister main -- terminating
-INFO pbench-tool-meister main -- Remove pid file ... (tm-lite-testhost.example.com.pid)
---- mock-run/tm/tm-lite-testhost.example.com.log file contents
+INFO pbench-tool-meister driver -- terminating
+--- mock-run/tm/tm-lite-testhost.example.com.err file contents
 +++ mock-run/tm/tm-lite-testhost.example.com.out file contents
 --- mock-run/tm/tm-lite-testhost.example.com.out file contents
++++ mock-run/tm/tm.logs file contents
```
portante, line 329: @RobertKrawitz, this is how the combined Tool Meister logs will look in the `${benchmark_run_dir}/tm/tm.logs` file.

The registered host name is emitted with a sequence number so that the log file can be searched / sorted as appropriate.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -568,6 +570,7 @@ def main(argv):
                             " daemonized; return code: %d",
                             retcode,
                         )
+                        failures += 1
```
portante, line 573: Bug found when handling errors starting the Tool Meister daemon locally.

agent/util-scripts/unittests
```diff
@@ -519,9 +519,17 @@ function sort_testlog {
...
+}
+
 function sort_tdslog {
-    sort ${_testdir}/mock-run/tm/pbench-tool-data-sink.log > ${_testdir}/mock-run/tm/pbench-tool-data-sink.log.sorted
-    mv ${_testdir}/mock-run/tm/pbench-tool-data-sink.log.sorted ${_testdir}/mock-run/tm/pbench-tool-data-sink.log
+    sort_log_file ${_testdir}/mock-run/tm/pbench-tool-data-sink.log
+}
+
+function sort_tmlogs {
+    sort_log_file ${_testdir}/mock-run/tm/tm.logs
```
portante, line 532: The sequence number emitted in the Tool Meister logs file, zero padded, makes this work well.

lib/pbench/cli/server/shell.py
```diff
@@ -36,11 +36,11 @@ def main():
...
             "--workers",
-            str(workers),
+            workers,
             "--pid",
             "/run/pbench-server/gunicorn.pid",
             "--bind",
-            f"{str(host)}:{str(port)}",
-            f"pbench.cli.server.shell:app())",
+            f"{host}:{port}",
+            "pbench.cli.server.shell:app())",
```
portante, line 44: A local `tox -e lint` failed for me due to `flake8` catching the empty f-string.  Not sure how this made it past all the commit checks.  Made this a separate commit that we could take in another PR.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -999,12 +1071,6 @@ def main(argv):
                 logger.exception("ERROR - failed to start the tool data sink")
         except Exception:
             logger.exception("ERROR - failed to start the tool data sink")
-        finally:
```
portante, line 1002: This was never necessary as the `pidctx` object takes care of this.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -414,10 +476,21 @@ def _cleanup(self):
...
-        self._pubsub.unsubscribe()
-        self.logger.debug("pubsub close")
-        self._pubsub.close()
+        if self._pubsub:
+            self.logger.debug("unsubscribe client")
+            self._pubsub.unsubscribe()
+            self.logger.debug("pubsub close client")
+            self._pubsub.close()
+        try:
+            self.tm_log_capture_thread.join()
```
portante, line 485: Why join with the logging thread first?  Because we unsubscribe as part of termination, which shutdowns the channel cleanly so that we only take the heavy hammer approach during whacky exception handling.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -398,10 +457,13 @@ def execute(self):
                         self.state_change(data)
         except self.Terminate as exc:
```
portante, line 458: We could consider dropping the `Terminate` exception in favor of unsubscribing from the channel and handling without a warning issue at line 430.

lib/pbench/agent/tool_meister.py
```diff
@@ -50,23 +50,59 @@
...
+        self.counter = 0
+        self.errors = 0
+
+    def emit(self, record):
+        """Publish record to redis logging channel
+        """
+        try:
+            formatted_record = self.format(record)
+            self.redis_client.publish(
+                self.channel, f"{self.hostname} {self.counter:04d} {formatted_record}"
```
portante, line 99: Zero padded counter makes combined log messages a bit more sortable (at least until you have 10,000 or more).

lib/pbench/agent/tool_meister.py
```diff
@@ -463,11 +512,19 @@ def _get_data(self):
...
-                self.logger.exception("Error fetching 'next' data off channel")
-                continue
+            except redis.ConnectionError as exc:
+                try:
+                    msg = exc.args[0]
+                except Exception:
+                    msg = ""
+                if msg.startswith(SERVER_CLOSED_CONNECTION_ERROR):
+                    # Redis disconnected, shutdown as gracefully as possible.
+                    # FIXME: Consider adding connection drop error handling, retry loop
```
portante, line 522: We use `next(self._chan)` to give us a chance to try to handle re-tries in the future.

lib/pbench/agent/tool_meister.py
```diff
@@ -1137,6 +1193,179 @@ def sysinfo(self, data):
...
+    """
+    if logger is None:
+        logger = get_logger(PROG)
+
+    # Add a logging handler to send logs back to the Redis server, with each
+    # log entry prepended with the given hostname parameter.
+    rh = RedisHandler(
+        channel="tm-logging", hostname=params["hostname"], redis_client=redis_server
+    )
+    if os.environ.get("_PBENCH_TOOL_MEISTER_LOG_LEVEL") == "debug":
```
portante, line 1243: Okay, this is ugly.  We call `get_logger()` from two places to handle daemonized vs non-daemonized.  I admit, I did not spend my time thinking about how to make this cleaner.

agent/util-scripts/gold/test-client-tool-meister/test-56.txt
RobertKrawitz: :+1: The sequence number is very helpful.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -362,23 +377,69 @@ def run(self):
...
 
+    def tm_log_capture(self):
+        """tm_log_capture - capture all logs written by local and remote Tool
+        Meisters through the Redis server into one file.
+        """
+        tm_log_file = self.benchmark_run_dir / "tm" / "tm.logs"
+        with tm_log_file.open("w") as fp:
+            try:
+                for payload in self._tm_logging_chan:
+                    self.logger.debug("payload: %r", payload)
```
dbutenhof, line 388: Irrelevant side gripe, but it really bugs me that we're using libc printf `%` substitution on the agent side and `{}` substitution on the server side, with the implication that it's basically impossible to log correctly in common code.

Besides, using `%` just feels like going back to C, and we haven't been playing in that playground for a long time.

lib/pbench/agent/tool_meister.py
```diff
@@ -1137,6 +1196,183 @@ def sysinfo(self, data):
         return failures
 
 
+def get_logger(PROG, daemon=False):
+    # If in Unit Test environment, just log to console
+    # If in non-unit test environment:
+    #   if daemonized, log to syslog and log back to Redis
+    #   if not daemonized, log to console and log back to Redis
+    # We assume
```
dbutenhof, line 1204: Doc string??

dbutenhof (dismissed):

lib/pbench/agent/tool_data_sink.py
portante: I agree.

But fixing that is more invasive than would be worth doing as part of another PR.  So I'd like to get these Tool Meister changes working before I tackle that.

lib/pbench/agent/tool_meister.py
portante: Made into a doc-string.

lib/pbench/agent/tool_data_sink.py
portante: See issue https://github.com/distributed-system-analysis/pbench/issues/2025.

dbutenhof (approved):

dbutenhof (approved):

zulcss (approved):
lgtm

lib/pbench/agent/tool_data_sink.py
portante: Done.

PR #2015, opened 2020-12-02
Prometheus quick fix

Maxusmusti:
The quay org had a name change from "prom" to "prometheus"

### DIFF
--- a/lib/pbench/agent/tool_data_sink.py
+++ b/lib/pbench/agent/tool_data_sink.py
@@ -221,7 +221,7 @@ class PromCollector(BaseCollector):
                 )
                 return 0
 
-            args = ["podman", "pull", "prom/prometheus"]
+            args = ["podman", "pull", "quay.io/prometheus/prometheus"]
             try:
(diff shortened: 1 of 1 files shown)


lib/pbench/agent/tool_data_sink.py
```diff
@@ -221,7 +221,7 @@ def launch(self):
                 )
                 return 0
 
-            args = ["podman", "pull", "prom/prometheus"]
+            args = ["podman", "pull", "quay.io/prometheus/prometheus"]
```
dbutenhof, line 224: In scrum I think Pete had suggested that we consider moving this to a config option, which would allow easily substituting another build of Prometheus tools if necessary for some special purpose. That might be worth considering; but I don't think it needs to gate the "quick fix".

dbutenhof (approved):

lib/pbench/agent/tool_data_sink.py
portante: Tracked in issue #2016.

portante (approved):

PR #2020, opened 2020-12-04
Fix server spec file

dbutenhof:
The reorganized API resource files need to be represented properly in the RPM spec file.

There are several flake8/black violations elsewhere that somehow got past presubmit checks that I needed to fix to get this in; I know Pete has already fixed at least one of them.


npalaska (dismissed):
black was complaining about single vs double quotes?

ndokos (approved):

portante (approved):

PR #2022, opened 2020-12-04
Fix bad jenkins "pytests" runs

portante:
And as a result, we fix the issues caught by these fixes.


dbutenhof (approved):
Cool!

jenkins/run
```diff
@@ -54,6 +54,9 @@ if [[ ${?} -ne 0 ]]; then
     exit 1
 fi
 
+find . -name __pycache__ -exec rm -r {} \; > /dev/null 2>&1
```
npalaska, line 57: should we do `-name __pycache__ -or -name '*.pyc'` I find sometime I need to clear .pyc files to get the pytest working, but if they also get clear when we clear the cache thats fine

portante: Where do you find the `.pyc` files?

npalaska: Ahh interesting I didnt notice they all fall under `__pycache__` directory, cool

npalaska (approved):

ndokos (approved):

PR #2024, opened 2020-12-04
Add endpoint configuration query

dbutenhof:
The advertised endpoints are self-configuring based on reverse-proxy configuration as long as the reverse-proxy service advertises the configured external host address either via `Forward: [...];host=<name>[;...]` or by `X-Forwarded-Host: <host>[,...]` (note that in the latter case, the first hostname given is presumed to be the first and most appropriate proxy).

The following sample is generated by a `GET <host>/api/v1/endpoints` with the header `X-Forwarded-Host: pbench.perf.lab.eng.bos.redhat.com:8902`, for example:

```
{
  "api": {
    "controllers_list": "http://pbench.perf.lab.eng.bos.redhat.com:8902/api/v1/api/v1/controllers/list",
    "controllers_months": "http://pbench.perf.lab.eng.bos.redhat.com:8902/api/v1/api/v1/controllers/months",
    "elasticsearch": "http://pbench.perf.lab.eng.bos.redhat.com:8902/api/v1/api/v1/elasticsearch",
    "endpoints": "http://pbench.perf.lab.eng.bos.redhat.com:8902/api/v1/api/v1/endpoints",
    "graphql": "http://pbench.perf.lab.eng.bos.redhat.com:8902/api/v1/api/v1/graphql",
    "host_info": "http://pbench.perf.lab.eng.bos.redhat.com:8902/api/v1/api/v1/host_info",
    "login": "http://pbench.perf.lab.eng.bos.redhat.com:8902/api/v1/api/v1/login",
    "logout": "http://pbench.perf.lab.eng.bos.redhat.com:8902/api/v1/api/v1/logout",
    "register": "http://pbench.perf.lab.eng.bos.redhat.com:8902/api/v1/api/v1/register",
    "upload_ctrl": "http://pbench.perf.lab.eng.bos.redhat.com:8902/api/v1/api/v1/upload/ctrl/",
    "user": "http://pbench.perf.lab.eng.bos.redhat.com:8902/api/v1/api/v1/user/"
  },
  "identification": "Pbench server 0.71.0-11gdb60e9279",
  "indices": {
    "result_data_index": "drb.v5.result-data.",
    "result_index": "drb.v5.result-data-sample.",
    "run_index": "drb.v6.run-data.",
    "run_toc_index": "drb.v6.run-toc."
  }
}
```

This requires that the dashboard do a `fetch` back through the `origin` address with suffix `/api/v1/endpoints`, and stash the resulting JSON somewhere. I have a prototype dashboard branch where I've simply replaced the `<script>` in `document.ejs` with a more dynamic version that amounts to effectively the same HTTP and Javascript operations performed to load the static `endpoints.js` used by the current code:

```
  <!-- runtime endpoints config -->
  <script type="text/javascript">
    var response = await fetch(window.origin + "/api/v1/endpoints");
    window.endpoints = await response.json();
  </script>
```

(Though whether this is the "best" way to do it may be another question, it works... of course with appropriate changes to the endpoint references through the code, and I haven't yet gotten the e2e tests debugged enough to push a draft PR.)

_NOTE_

The `user` endpoint in Flask ends with a parameter template string; I decided to remove that from the endpoint, but leave the trailing `/` so the dashboard can simply append the username on the URI. The simple parsing here won't work if we end up with a template string that's not at the end, but that's agile.


dbutenhof:
I realize I missed another meta-comment. I materialized "api" as the "root" path to the Pbench server APIs, but perhaps a better and more expandable (maybe even more approachable) mechanism would be to make that a dict of all the supported APIs rather than requiring the client to know them and paste paths together. E.g.,

```
    "api": {
        "queryControllers": {"uri": "dhcp31-170.perf.lab.eng.bos.redhat.com:8001/api/v1/controllers/list"},
        "queryResults": {"uri": "dhcp31-170.perf.lab.eng.bos.redhat.com:8001/api/v1/controllers/runs"},
        [...]
    }
```

I.e., the "api" dict key would match the dashboard service level names. The use of a nested dict with "uri" allows expansion to additional metadata, for example a description string or even a full JSON schema definition of request and response payloads if we eventually want to go that far.

portante (changes requested):
Instead of a separate API endpoint, why not render the web page served to the browser with the information embedded in the rendered page?

dbutenhof:
> Instead of a separate API endpoint, why not render the web page served to the browser with the information embedded in the rendered page?

One big reason is that this is part of the server, not the dashboard. It can be used by curl, or Postman, or any other client to learn the server configuration.

Now, we still need the client to know how to do that first query to find the configuration, and that needs to be known externally somehow. That's where your suggestion might come in, on the dashboard side.

It also occurs to me that a GET <server>:8001 probably should return this information rather than requiring the client to know <server>:8001/api/v1/endpoints. Then if we do a "v2" API the default unqualified endpoint would return the latest version info.

Maybe it should even identify itself as "server": "Pbench server 0.71-xxx" or some such.

gurbirkalsi:
@dbutenhof With the endpoint configuration being integrated into the dashboard binary, is this endpoint going to serve as a starting point for the deployment process to populate `config/endpoints.js` before running `yarn build` and eventually moving it to the target server?

portante:
> @dbutenhof With the endpoint configuration being integrated into the dashboard binary, is this server endpoint going to be a starting point for the deployment process to populate `config/endpoints.js` before running `yarn build` and eventually moving it to the target server?

We need to really consider what it means to embed configuration information into the dashboard binary.

Ideally, the dashboard code should deployable to any server environment, and once loaded into the browser, it figures out its configuration from the server from which it was loaded.  Having the feature that a given dashboard binary could be built for staging, tested there, and then deployed to production with out any changes to the binary is highly desirable.

So let's say the starting URL for the dashboard code is:  `https://pbench.example.com/dashboard/`.  Then the browser will be served the `index.html` file from the disk of the server at `pbench.example.com:/var/www/html/dashboard/index.html`.  The `.html` will reference relative paths for the Javascript components, favicons, images, etc.  We can also configure the web-server to respond to an `https://pbench.example.com/api` or `https://pbench.example.com/dashboard/api` where the browser can then fetch the endpoint configuration information, just it does the CSS, fonts, and other static media.

gurbirkalsi:
I've addressed these requirements in https://github.com/distributed-system-analysis/pbench-dashboard/pull/110. Please let me know if these changes meet the requirements for deploying the dashboard binary with a standalone config.

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,99 @@
...
+        __init__ Construct the API resource
+
+        Args:
+            config (PbenchServerConfig): server config values
+            logger (Logger): message logging
+        """
+        self.logger = logger
+        port = config.get("pbench-server", "rest_port")
+        host = config.get("pbench-server", "host")
+        self.host = f"{host}"
```
portante, line 25: Can this be `str(host)` instead?

portante (dismissed):

lib/pbench/server/api/resources/endpoint_configure.py
dbutenhof: Sure; why not?

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,99 @@
...
+        __init__ Construct the API resource
+
+        Args:
+            config (PbenchServerConfig): server config values
+            logger (Logger): message logging
+        """
+        self.logger = logger
+        port = config.get("pbench-server", "rest_port")
+        host = config.get("pbench-server", "host")
+        self.host = f"{str(host)}"
```
portante, line 25: Wait, why not just, `self.host = str(host)`?

portante (approved):

lib/pbench/server/api/resources/endpoint_configure.py
dbutenhof: Because ... I just kinda made the change on autopilot without really thinking? It does look just a teeny bit... weird, doesn't it.

portante (approved):

dbutenhof:
Ha; just realized I messed this up a bit; another version coming soon.

portante (dismissed):

dbutenhof:
NOTE: The big missing piece here is a way to configure a server for an external proxy like our production/staging environment. There's also a related bug in fabricating the `results` "API" link.

The server only inherently knows its own host address and the configured server port. Now, we support (by default) an Apache reverse proxy to make the API accessible via http/https instead of specifying a new port. (Note that `results` is just normal Apache access to local files through `<pbench-root>/public_html/results` and doesn't need a special port, although in the future we'll be moving this to a real server API so we don't need to run Apache locally or even have a local filesystem to server.)

When you're running with an external proxy, however, we don't want to direct the dashboard to the local host/port(s) at all, so we need new `pbench-server.cfg` options to describe the external proxy configuration here. Something like:

```
[ReverseProxy]
host = pbench.example.com
http_port = 8901
api_port = 8902
```

The endpoint configuration API would then express all the endpoints in terms of that configured proxy address rather than using the local host and port. (Note that unless we _also_ run the local Apache reverse-proxy, the NGINX configuration will need to route the /api URI path to the local server port, and other accesses to local http; whereas if we do keep the local proxy, we don't need to open the server port and NGINX would always route to http/https.)

Note that for this to work we also need Ansible configuration script changes to distinguish local Apache reverse-proxy from remote (e.g., NGINX) reverse-proxy. In the former case, as now, we need to configure Apache local reverse-proxy; in the latter case, we need to instead open the local API port in the firewall. For now, I guess we always need to configure Apache locally; if we drop the Apache `public_html` results access and move to a real API, we could drop that and completely drop Apache configuration from the "remote proxy" configuration flow. I started documenting some of this work in issue #2099 a while ago.

lib/pbench/test/unit/server/test_endpoint_configure.py
```diff
@@ -0,0 +1,42 @@
...
+        """
+        host = server_config.get("pbench-server", "proxy_host")
+        uri_prefix = server_config.rest_uri
+        uri = f"{host}{uri_prefix}"
+        prefix = get_index_prefix(server_config)
+        expected_results = {
+            "metadata": {
+                "identification": f"Pbench server {server_config.COMMIT_ID}",
+                "prefix": prefix,
+                "run_index": f"{prefix}.v6.run-data.",
```
portante, line 25: Not sure this is correct for the dashboard today, as it is currently concatenating the prefix and the specific index where needed.  In the future, when we drop direct Elasticsearch queries from the dashboard, we'll be able to drop the need for this index names.  But for now, we can avoid additional code changes in the dashboard until then by not adding the prefix in the index names.

server/lib/config/pbench-server-default.cfg
```diff
@@ -29,6 +29,20 @@ mailto=%(admin-email)s
 mailfrom=%(user)s@%(host)s
 commit_id=unknown
 
+# Configuration to support remote reverse-proxy configuration (e.g., NGINX)
+# in front of Pbench server.
+#
+# This will be advertised to front-end clients asking for endpoint configuration.
+# By default, this is the http port (80) on the server host, which will be
```
portante, line 36: Should this talk about being a host name / port pair?

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,106 @@
...
+    def __init__(self, config, logger):
+        """
+        __init__ Construct the API resource
+
+        Args:
+            config (PbenchServerConfig): server config values
+            logger (Logger): message logging
+        """
+        self.logger = logger
+        host = config.get("pbench-server", "proxy_host")
```
portante, line 23: This means it is always required to have this in the configuration file.  Could we make this optional?

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,106 @@
...
+        self.host = str(host)
```
portante, line 24: Why do we need to use `str()` here on the value fetched from the configuration file?

portante (changes requested):

server/lib/config/pbench-server-default.cfg
dbutenhof: It's in the examples I gave; but you're right that I didn't say so directly up front, and that would be better... especially since I omitted the `:80` in the default value. (I didn't really think it would help to include that, but perhaps it would?)

lib/pbench/server/api/resources/endpoint_configure.py
dbutenhof: I put it in the default file so I'd have a place to explain it; that'd be harder to do if it was omitted entirely, and it didn't seem to have any real downside. Do you object?

lib/pbench/test/unit/server/test_endpoint_configure.py
dbutenhof: The dashboard today doesn't use this; I have another branch where I started modifying the dashboard code to load the endpoints from this API, and of course it will directly reference the appropriate properties from the loaded JSON instead of constructing them as it does today.

lib/pbench/server/api/resources/endpoint_configure.py
dbutenhof: You've mentioned that before, but I never saw a need. Is the config package really not returning the string?

dbutenhof: > Why do we need to use `str()` here on the value fetched from the configuration file?

Actually, I'm not sure. That slipped in somehow. I assume the config package returns the string; but I think that in another PR you'd encouraged using `str()` and I got paranoid.

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,105 @@
...
+    def __init__(self, config, logger):
+        """
+        __init__ Construct the API resource
+
+        Args:
+            config (PbenchServerConfig): server config values
+            logger (Logger): message logging
+        """
+        self.logger = logger
+        self.host = config.get("pbench-server", "proxy_host")
```
webbnh, line 23: The `"proxy_host"` argument is a key, not a default value...right?  What gets returned if the key is missing (or blank)?  Should there be some sanity checking of the config file values?

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,105 @@
...
+        uri_prefix = config.rest_uri
+        self.uri = f"{self.host}{uri_prefix}"
```
webbnh, line 25: Might it be worthwhile to have a little code which ensures that there is exactly the right amount of punctuation between `self.host` and `uri_prefix`? (I.e., there's supposed to be exactly one `/` there -- right? -- provided by the prefix and _not_ the host?...but, what if it's missing from one or there's and extra from the other?)

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,105 @@
...
+        """
+        Return server configuration information required by the UI
+        dashboard. This includes
+
+        metadata: Information about the server configuration
+            prefix: The indexing prefix applied by the server to partition a
+                shared Elasticsearch cluster index namespace.
+            result_index: The "root" index name for Pbench result data,
+                qualified by the current index version and prefix. In the
+                current ES schema, this is "v5.result-data-sample."
```
webbnh, line 39: You put periods at the end of the description at lines 35-36 and elsewhere below, but for descriptions ending in `."` you omitted the trailing periods.  This is conventional typography for prose, but it creates an ambiguity in this context.  Thus, I would recommend either omitting the periods from all the descriptions (many of them are not full sentences, after all) or supplying periods on all of them (which will make it clear that the text inside the quotes is literal).

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,105 @@
...
+                is currently direct HTTP access through Apache public_html, and
+                not an "API" as such; we will need to adopt this as a real
+                server API when Pbench moves from local filesystem to S3 storage
+                schemas.
+            elasticsearch: The Pbench pass-through URI for the Elasticsearch
+                cluster. This will eventually be superceded by the native
+                Pbench APIs, though it might remain accessible with special
+                user/group privileges to support special cases?
+            graphql: The GraphQL frontend on postgreSQL currently used by the
+                dashboard user mocks. This will be superceded and decprecated
```
webbnh, line 62: `decprecated ` ==> `deprecated`

server/lib/config/pbench-server.cfg.example
```diff
@@ -1,6 +1,7 @@
 [DEFAULT]
 # The values here override those in pbench-server-default.cfg.
-# An installation will want to override at least the following:
+# An installation must override at least the "example.com"
+# host names.
 default-host = pbench.example.com
 install-dir = /opt/pbench-server
```
webbnh, line 6: Should this entire section or lines 5 and/or 6 be commented out...with the expectation that anyone who wants to override `pbench-server-default.cfg` will uncomment them and change their values to something other than the example placeholders?

Possibly ditto the other `host` and/or `port` lines, below?

server/rpm/pbench-server.spec.j2
```diff
@@ -131,6 +131,7 @@ fi
 /%{installdir}/lib/pbench/server/api/__init__.py
 /%{installdir}/lib/pbench/server/api/resources/graphql_api.py
 /%{installdir}/lib/pbench/server/api/resources/upload_api.py
+/%{installdir}/lib/pbench/server/api/resources/endpoint_configure.py
```
webbnh, line 134: Huh.  Another new file...when are we going to switch to the directory syntax?

webbnh (approved):
Looks OK to me.

I found a misspelling and a couple of typographical ambiguities; I have some questions about argument (sanity) checking; and there's a suggestion and a goad...but nothing gating.

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,105 @@
...
+        move the dashboard away from direct access to the backend DB servers.
+        The alternative would be to expose the direct Elasticsearch and GraphQL
+        URIs here.
+        """
+        try:
+            endpoints = {
+                "metadata": {
+                    "identification": f"Pbench server {self.commit_id}",
+                    "prefix": self.prefix,
+                    "run_index": f"{self.prefix}.v6.run-data.",
```
portante, line 85: Won't this change require a change in the dashboard for this to work correctly?  And why provide the `prefix` at all?

portante (changes requested):

lib/pbench/server/api/resources/endpoint_configure.py
dbutenhof: I thought I answered this in another context; but since I'm going through individual file comments now...

Yeah, of course it will. Everything in this PR is designed to provoke a change in the dashboard. In particular, this means that the dashboard no longer needs to know how version and prefix and root index name go together to construct an Elasticsearch index. That sounds like a worthy improvement to me.

server/lib/config/pbench-server.cfg.example
dbutenhof: This file strides between two different styles but doesn't really conform to either.

1. We could build it as a documented template of all of the config options you _could_ set or change, with values commented out. I think Apache does this with httpd.
2. We could keep it stripped down and clean with only the config options a site really _has_ to specify, like `default-host` ... in which case commenting them just means they need to remove the `#` when they edit it and it doesn't have much value.

We kinda do both here. For example, `install-dir` had better be in `pbench-server-default.cfg` because that's the standard default. They'd change it only if they do something really weird; and either it shouldn't be here at all or it should be commented out. The `default-host` is something they have to set.

There are no default, for example, for the `[Indexing]` section below, nor any guidelines for selecting the `bulk_action_count` (or even what it means). Arguably, that _should_ be a default, and should be documented.

I hope you won't choose to drag me down this rat hole right now, but it's something we really ought to record and think about as part of turning our project into a product.

lib/pbench/server/api/resources/endpoint_configure.py
webbnh: I'm just concerned about the case where "them" is "us" (or "me"!).  :-D  (It seems like we lost a bunch of time in our most recent Jam Session due to having the wrong number of slashes...as I recall, we were installing a reverse proxy and, apparently, we didn't quite know what we were doing....)

I understand and concur with resisting feature creep.  I s'pose I'll open an issue.

lib/pbench/server/api/resources/endpoint_configure.py
webbnh: Doing it piecemeal would move us forward in steps.  ;-)

(In any case, now we have #2125.)

server/lib/config/pbench-server.cfg.example
webbnh: No dragging.  (See above where I approved the PR!)

However, it does sound like you would like to see an open an issue for this....  ;-)

server/rpm/pbench-server.spec.j2
webbnh: Can't wait!  :-D

lib/pbench/server/api/resources/endpoint_configure.py
webbnh: Right.  And, given that, do we still need line 84 here (since it's incorporated in line 85)?

dbutenhof: I was being paranoid. But of course you're right. I'll remove it.

server/lib/config/pbench-server.cfg.example
dbutenhof: Yes, and thanks.

lib/pbench/server/api/resources/endpoint_configure.py
dbutenhof: That typo was on the NGINX side, so better handling of errors in our config file wouldn't have helped... I'm also not sure if it was inherently wrong to have the trailing slash or if it just conflicted with something else in Pete's NGINX config.

Still, again, the point is valid, and giving help even for one-time configuration is always a good thing. But I would definitely prefer that we attack that strategically with some deliberation.

webbnh (approved):

lib/pbench/server/api/resources/endpoint_configure.py
portante: Perhaps we could use Python 3's `urllib` module?  See https://stackoverflow.com/questions/15799696/how-to-build-urls-in-python

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,102 @@
...
+                by native Pbench user management APIs
+            queryControllers: Return information about the run documents that
+                were created within a specified range of months.
+            queryMonthIndices: Return the YYYY-mm date strings for all ES
+                vx.run-data.* indices, in descending order
+
+        TODO: We need an internal mechanism to track the active versions of the
+        various Elasticsearch template documents. We're hardcoding them here and
+        in other APIs. We should consider persisting an equivalent of the
+        mapping table built in "indexer.py" for use across the server APIs.
```
portante, line 70: We want to move away from the Dashboard having to "know" how to construct Elasticsearch queries, so would this be really necessary once we add APIs for all the explicit queries made?

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,102 @@
...
+
+        Meta TODO: We're giving the pass-through APIs for Elasticsearch and
+        GraphQL here, which implies adoption of Nikhil's and Fuqing's work to
+        move the dashboard away from direct access to the backend DB servers.
+        The alternative would be to expose the direct Elasticsearch and GraphQL
+        URIs here.
```
portante, line 76: Now that https://github.com/distributed-system-analysis/pbench-dashboard/pull/95 has been merged, do we still need this comment?

server/lib/config/pbench-server-default.cfg
```diff
@@ -29,6 +29,28 @@ mailto=%(admin-email)s
...
+#
+# To configure Pbench to advertise a remote reverse-proxy, set "proxy_host=host:port"
+# in pbench-server.cfg. Note that the local %(bind_port) must then be opened in
+# firewalld, and the reverse proxy must pass %(rest_uri) traffic on the proxy
+# port to %(host):%(bind_port) and other traffic to %(host):80. For example,
+# NGINX should have a "location /api" with at least "proxy_pass http://<host>:<bind_port>"
+# and a "location /" with at least "proxy_pass http://<host>:80" if the Pbench host
+# Apache isn't configured to listen on a non-default port. (Further instruction
+# in reverse proxy configuration is beyond the scope of this mega-comment.)
+proxy_host=%(host)s
```
portante, line 52: I think we can drop this entirely.

The reverse proxy is configured to pass along in a header the original URL that it accepted the request on, so that the backend server, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host.

lib/pbench/server/api/resources/endpoint_configure.py
portante: We can use `X-Forwarded-For` to find the original target URL, and then add our endpoints relative to that.  See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host.

portante (changes requested):

lib/pbench/server/api/resources/endpoint_configure.py
dbutenhof: I've used `urllib` before ... I somehow didn't even think of it here because the construction is simple, but I like that suggestion.

As for the `X-Forwarded-For` ... I had wondered if there was a header like that but kept forgetting to investigate. Yeah, that's great, and I'll have to experiment with it. I like the idea of just configuring all the endpoints based on how the browser got to the configuration API. No header means local proxy, otherwise we use the header value.

lib/pbench/server/api/resources/endpoint_configure.py
dbutenhof: Yes, after we've replaced all the APIs there should be no need for the dashboard to know anything about index names, which will be nice.

lib/pbench/server/api/resources/endpoint_configure.py
dbutenhof: Part of it, until the dashboard drops GraphQL ... but yes, Nikhil's work is now in so that simplifies the comment a bit.

server/lib/config/pbench-server-default.cfg
dbutenhof: Sounds good to me.

lib/pbench/server/api/resources/endpoint_configure.py
dbutenhof: Nope; it'll be gone.

server/lib/config/pbench-server-default.cfg
dbutenhof: Interestingly, there appears to be a newer `Forwarded` header (with a somewhat more obtuse syntax), combining the non-standard `X-Forwarded-For`, `X-Forwarded-Host`, and `X-Forwarded-Proto`.

Maybe I'll try to support either. I'll probably have to try to set up a remote reverse-proxy to my VM to test it. Maybe we can set up another port on perf44??

dbutenhof: > Maybe I'll try to support either. I'll probably have to try to set up a remote reverse-proxy to my VM to test it. Maybe we can set up another port on perf44??

And ... of course I don't need that because I can fake any headers I want through Postman.

lib/pbench/server/api/resources/endpoint_configure.py
dbutenhof: The urllib ... "limitations" didn't make it quite as useful as I'd hoped. It's really not well suited to gluing together sub-paths.

The new commit switches completely to dynamically detecting a reverse proxy by using either `X-Forwarded-Host` or `Forwarded` headers, and there are unit tests to check the handling.

There are some debug log statements here I'll remove, and I need to think about how to deal with the templated user URI, so I'm not quite ready for another round of reviews ... but I wanted to push a checkpoint to show where it it.

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,145 @@
...
+                    "endpoints": f"{uri}/endpoints",
+                    "graphql": f"{uri}/graphql",
+                    "controller_list": f"{uri}/controllers/list",
+                    "controller_months": f"{uri}/controllers/months",
+                    "dataset_list": f"{uri}/datasets/list",
+                    "dataset": f"{uri}/datasets",
+                    "register": f"{uri}/register",
+                    "login": f"{uri}/login",
+                    "logout": f"{uri}/logout",
+                    "user": f"{uri}/user/<string:username>",
```
portante, line 136: This is cool, but doesn't this mean that if a new endpoint is added, it needs to be manually added here?  I wonder if there is a way to dynamically derive this list from Flask somehow?  Perhaps in another PR?

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,145 @@
...
+            if m:
+                origin = m.group("host")
+                self.logger.info("Forwarded: {}, ({})", header, origin)
+        if not origin:
+            header = request.headers.get("X-Forwarded-Host")
+            if header:
+                m = self.x_forward_pattern.search(header)
+                if m:
+                    origin = m.group("host")
+                    self.logger.info("X-Forwarded-Host: {}, ({})", header, origin)
```
portante, line 103: Can we consider making this `.info()` message, and its sister at line 96 `.debug()`?

portante (changes requested):
Lookin' good ...

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,145 @@
...
+        """
+        __init__ Construct the API resource
+
+        Args:
+            config (PbenchServerConfig): server config values
+            logger (Logger): message logging
+
+        Report the server configuration to a web client; web access can be
+        redirected through an external reverse proxy like NGINX by setting
+        the "proxy_host" configuration setting in the "pbench-server"
```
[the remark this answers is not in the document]
dbutenhof: Yep; I missed that, and a few others. I already fixed a few of them after pushing this commit.

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,145 @@
...
+        server uses a local Apache reverse proxy routing through the HTTP
+        port (80), but this can be changed to any host name and port. All
+        server endpoints will be reported with respect to that address.
+        """
+        self.logger = logger
+        self.host = config.get("pbench-server", "host")
+        self.uri_prefix = config.rest_uri
+        self.prefix = get_index_prefix(config)
+        self.commit_id = config.COMMIT_ID
+        self.forward_pattern = re.compile(r";\s*host\s*=\s*(?P<host>[^;\s]*)")
```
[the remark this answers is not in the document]
dbutenhof: I didn't bother because it's effectively a singleton. That might look a bit cleaner, however, so I suppose I can do that anyway.

lib/pbench/server/api/resources/endpoint_configure.py
dbutenhof: They were there for debugging when I pushed queries, and I'd meant to remove them. I do want some way to reconstruct the forwarding headers in case something goes wrong, and I want that to be info so we can debug from a log rather than trying to reproduce it.

lib/pbench/server/api/resources/endpoint_configure.py
dbutenhof: As long as we still have direct Elasticsearch queries in the dashboard code, it needs to know the index names. We can remove the indices section entirely once we've gotten rid of them all, but there are still a few very messy ones I haven't even gotten to attempting yet.

I suppose we could alternately rejigger the `elasticsearch` endpoint so we could fill in index name templates in the URI rather than the dashboard needing to know; but that's potentially a lot more complicated, and this seemed like a more straightforward replacement of the current static configuration.

lib/pbench/server/api/resources/endpoint_configure.py
dbutenhof: Yes. In theory I think it's possible to traverse the endpoints in Flask. I started using the names of the existing dashboard service methods, which couldn't be derived. Now... well, it's an interesting idea. After Webb pushed me far enough past the tipping point, I renamed them to be better aligned with the endpoints. Aside from a couple of minor variations, I would have just taken the URI path past the `/api/v1/` and replaced the slashes with "_", so it's not impossible. Maybe I'll even give it a try...

lib/pbench/server/api/resources/endpoint_configure.py
portante: Why not just add this information to the final `.info()` message below and drop these entirely?

lib/pbench/server/api/resources/endpoint_configure.py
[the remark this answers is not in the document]
portante: FWIW, I am not generally in favor of singletons.  From "[What I think is good and bad](https://nedbatchelder.com/blog/202005/what_i_think_is_good_and_bad.html)", Ned Batchelder's blog: "singletons are hidden global state, and global state is bad".

lib/pbench/server/api/resources/endpoint_configure.py
portante: I am still skeptical as I don't think the current dashboard code works that way, and if you make this change now, it will required more dashboard changes to have it work differently.  See https://github.com/distributed-system-analysis/pbench-dashboard/blob/main/public/endpoints.js#L3.

portante (changes requested):

lib/pbench/server/api/resources/endpoint_configure.py
dbutenhof: What it means for the dashboard code is that the "root" index name is no longer patched together from the prefix and the hardcoded "base" name; the dashboard gets the whole thing from here and doesn't need to know how to construct it. For now, it still knows how to append the time series suffix; that (along with any need to know anything about the index names) will go away once all of the Elasticsearch queries have been migrated into the server.

The real difference is here, in `moment_constants.js`:

```
    if (index === endpoints.result_index) {
      queryString += `${endpoints.prefix + index + monthValue}-*,`;
    } else {
      queryString += `${endpoints.prefix + index + monthValue},`;
    }
```

vs the new

```
    if (index === endpoints.indices.result_index) {
      queryString += `${index + monthValue}-*,`;
    } else {
      queryString += `${index + monthValue},`;
    }
```

I can see I really need to devote some time to figuring out my e2e mocking problems so I can post a draft dashboard PR with the companion code.

portante (approved):

portante (approved):

webbnh (approved):
Looks good, but I found a couple of nits, and I have a bunch of pointed questions.

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,166 @@
...
+
+class EndpointConfig(Resource):
+    """
+    EndpointConfig API resource: this supports dynamic dashboard configuration
+    from the Pbench server rather than constructing a disconnected dashboard
+    config file.
+    """
+
+    forward_pattern = re.compile(r";\s*host\s*=\s*(?P<host>[^;\s]*)")
+    x_forward_pattern = re.compile(r"^\s*(?P<host>[^;\s,]*)")
```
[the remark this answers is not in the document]
dbutenhof: I had a reason in my head when I did it, I think; but you're right.

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,166 @@
...
+    param_template = re.compile(r"<[\w_\d]+:[\d_\w]+>")
```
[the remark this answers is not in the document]
dbutenhof: Ha; I used to use a regex dialect that didn't. Sometimes it still pops out.

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def __init__(self, config, logger):
+        """
+        __init__ Construct the API resource
+
+        Args:
+            config (PbenchServerConfig): server config values
+            logger (Logger): message logging
```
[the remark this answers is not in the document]
dbutenhof: Yes, I've started adding type hints everywhere, but I started this PR last year and I never considered that when I went back to it.

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,166 @@
...
+        self.uri_prefix = config.rest_uri
+        self.prefix = get_index_prefix(config)
+        self.commit_id = config.COMMIT_ID
+
+    def get(self):
+        """
+        Return server configuration information required by web clients
+        including the Pbench dashboard UI. This includes:
+
+        metadata: Information about the server configuration
```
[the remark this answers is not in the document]
dbutenhof: It actually is, but the comment is stale. Sigh. I missed that one.

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,166 @@
...
+        various Elasticsearch template documents. We're hardcoding them here and
+        in other APIs. We should consider persisting an equivalent of the
+        mapping table built in "indexer.py" for use across the server APIs.
+
+        TODO: We provide Elasticsearch index root names here, which the dashboard
+        code needs to perform the queries we've not yet replaced with server-side
+        implementations. The entire "indices" section can be removed once that is
+        resolved.
+        """
+        self.logger.debug("Received these headers: {!r}", request.headers)
```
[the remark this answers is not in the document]
dbutenhof: The log will include the module, method, and line. However, I removed this one anyway in favor of a single `INFO` message at the end to show what we found and where we got it:

```
INFO 18373 140143886407488 pbench.server.api.endpoint_configure get 107 -- Advertising endpoints relative to Forwarded: ; host  = 
```

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,166 @@
...
+        if header:
+            m = self.forward_pattern.search(header)
+            if m:
+                origin = m.group("host")
+                host_source = "Forwarded"
+                host_value = header
+        if not origin:
+            header = request.headers.get("X-Forwarded-Host")
+            if header:
+                m = self.x_forward_pattern.search(header)
```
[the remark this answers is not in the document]
dbutenhof: True; I copied and edited and didn't think to change it. The pattern is anchored, but `match` would be clearer.

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,166 @@
...
+        # Iterate through the Flask endpoints to add a description for each.
+        for rule in current_app.url_map.iter_rules():
+            url = str(rule.rule)
+
+            # Ignore anything that doesn't use our API prefix, because it's
+            # not in our API.
+            if url.startswith(self.uri_prefix):
+                # If the URI is parameterized with a Flask "<type:name>"
+                # template string, we don't want to report it, so we remove
+                # it from the URI. We derive an API name by converting the
```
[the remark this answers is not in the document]
dbutenhof: The client has to remove/substitute the templates then and I figured that wasn't very generic. I didn't really want to make it so Javascript-specific, either, and Javascript will only do that sort of substitution in literal template strings (backticked), so far as I can tell; although it's just a `.replace` away, it's not quite that neat.

Anyway, just letting it append the username to end end of the URI seemed easiest, at least for now.

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,166 @@
...
+                    "run_toc_index": f"{self.prefix}.v6.run-toc.",
+                    "result_index": f"{self.prefix}.v5.result-data-sample.",
+                    "result_data_index": f"{self.prefix}.v5.result-data.",
+                },
+                "api": apis,
+            }
+            response = jsonify(endpoints)
+        except Exception:
+            self.logger.exception("Something went wrong constructing the endpoint info")
+            abort(500, message="INTERNAL ERROR")
```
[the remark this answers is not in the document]
dbutenhof: Yeah; apparently Flask will do an automatic 500 on an unhandled exception. The documentation doesn't make clear what the response body will be, but we can also override this so it's probably worth cleaning this up through all of the resource APIs.

lib/pbench/server/api/resources/endpoint_configure.py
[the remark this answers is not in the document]
portante: Was this fixed?

lib/pbench/server/api/resources/endpoint_configure.py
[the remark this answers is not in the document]
portante: So what should the regex end up looking like?  What does it mean to use brackets with and two different orders for the bracket contents?

lib/pbench/server/api/resources/endpoint_configure.py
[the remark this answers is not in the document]
portante: So seems like we agree that we don't need to add type hints to this PR?

To that end, I've opened https://github.com/distributed-system-analysis/pbench/discussions/2152.

lib/pbench/server/api/resources/endpoint_configure.py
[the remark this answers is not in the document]
portante: Will be this resolved in this PR?

lib/pbench/server/api/resources/endpoint_configure.py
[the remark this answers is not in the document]
portante: So is this resolved?  It seems like this is a code change that will be made?

lib/pbench/server/api/resources/endpoint_configure.py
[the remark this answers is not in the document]
portante: Are you planning making this change in this PR?

lib/pbench/server/api/resources/endpoint_configure.py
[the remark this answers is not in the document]
portante: I second the idea of not making this Javascript specific.

portante (changes requested):
Also needs a rebase.

lib/pbench/server/api/resources/endpoint_configure.py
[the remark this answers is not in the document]
dbutenhof: I downgraded this one to `debug` rather than removing it; I think there's enough context later for routine triage.

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -0,0 +1,166 @@
...
+                if m:
+                    origin = m.group("host")
+                    host_source = "X-Forwarded-Host"
+                    host_value = header
+        if not origin:
+            origin = self.host
+        host = f"http://{origin}"
+        self.logger.info(
+            "Advertising endpoints relative to {}: {}", host_source, host_value
+        )
```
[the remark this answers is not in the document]
dbutenhof: Sure; one more pass.

lib/pbench/server/api/resources/endpoint_configure.py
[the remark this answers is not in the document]
dbutenhof: I added an issue to investigate this, #2149 ... I'd rather not attack that in this context.

lib/pbench/server/api/resources/endpoint_configure.py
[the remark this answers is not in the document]
webbnh: @portante, the ordering of characters and character classes within a bracket expression is not significant, so the two bracket expressions should be equivalent.

I believe that the regex should be `r"<\w+:\w+>"` (but I haven't looked at Dave's updates, yet).

lib/pbench/server/api/resources/endpoint_configure.py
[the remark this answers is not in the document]
webbnh: We don't _have_ to, but it would be better to have the hints in the code than the comment....  :-)

lib/pbench/server/api/resources/endpoint_configure.py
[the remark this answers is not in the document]
webbnh: I guess I was proposing two things:

- make the URI available as a template
- make it easy for the client to fill in the template (which is why I offered JS syntax)

It doesn't have to be Javascript format, but I'll point out that that format also works for shell, so it's not exactly JS-specific.  ;-)  We could use `%s` instead of `${...}`, but I opted for the latter because it provides semantic information.

It's true that the client will likely need to have some foreknowledge in order to fill in the template, but it will need to have even closer coupling to produce the URI if we just lop it off at the first parameter.  Providing a template removes the need for some of that coupling.

Apparently there is a (proposed) IETF [standard for URI templates](https://tools.ietf.org/html/rfc6570).

webbnh (approved):
Looks good.  I just have one question.

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -1,10 +1,12 @@
...
-from flask.globals import current_app
+from logging import Logger
 
+from flask.globals import current_app
 from flask_restful import Resource, abort
 from flask import request, jsonify
 from urllib.parse import urljoin
 
+from pbench.server import PbenchServerConfig
 from pbench.server.api.resources.query_apis import get_index_prefix
```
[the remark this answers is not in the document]
portante: Typically what PEP 8 talks about, which is alphabetical order grouped by `import`s first, `from` next, where standard library are grouped, then external libraries, then project imports.

lib/pbench/server/api/resources/endpoint_configure.py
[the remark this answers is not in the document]
portante: Let's track in issue #2154.

portante (approved):

PR #2027, opened 2020-12-08
Pythonize pbench-log-timestamp

zulcss:
Replace shell script with python script and add unit test.

### DIFF
--- a/lib/pbench/cli/agent/commands/timestamp.py
+++ b/lib/pbench/cli/agent/commands/timestamp.py
@@ -1,0 +1,13 @@
+"""pbench-log-timestamp"""
+
+import time
+
+import click
+
+
+@click.command()
+def main():
+    """ Add timestamp to user input"""
+    stdin = click.get_text_stream("stdin").read()
+    for line in stdin.splitlines():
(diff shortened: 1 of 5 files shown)


portante:
@zulcss did you mean to close this?

lib/pbench/cli/agent/commands/timestamp.py
```diff
@@ -0,0 +1,13 @@
...
+
+import time
+
+import click
+
+
+@click.command()
+def main():
+    """ Add timestamp to user input"""
+    stdin = click.get_text_stream("stdin").read()
```
portante, line 11: This needs to be a generator so that `stdin` can be a pipe and each line read from the pipe is timestamped and written out.

Further, I think we'll want to make sure it is not using buffered I/O so that folks can inspect the output easily as it is generated.

portante (changes requested):
Needs to work with streams of data from a PIPE, and also needs a rebase.

riya-17 (commented):
LGTM

portante:
I'll own driving this change into the repository.

webbnh:
At this point, I think it's safe to say that this PR will never be merged.

PR #2032, opened 2020-12-11
Fix py move-results to use `ConfigParser`

portante:
The Python 3 based "move-results" code was using the old `configtools` module instead of just using `ConfigParser` directly.  This would end up causing failures in the unit tests because it was calling the `parse_args` method of `configtools` to first parse the command line arguments.  But in principle, we never want to access the command line arguments while running unit tests.  The unit test should provide them for the method under test to properly ensure the correct behavior is tested.

We simply replace the use of `configtools` with `ConfigParser` directly, which ends up simplifying the code as well.

Further, using configtools just for the result directory `metadata.log` is over kill, because `configtools` was there to support multiple configuration files, and there is only one `metadata.log` file.

We remove the `SETUP` and `TEARDOWN` messages since they are really debugging statements and have no business being there for normal test operation.

### DIFF
--- a/lib/pbench/agent/results.py
+++ b/lib/pbench/agent/results.py
@@ -5,11 +5,11 @@ import requests
 import sys
 import tarfile
 
+from configparser import ConfigParser
 from pathlib import Path
+
 from werkzeug.utils import secure_filename
 
-from pbench.cli.agent.commands.log import add_metalog_option
-from pbench.common import configtools
 from pbench.common.utils import md5sum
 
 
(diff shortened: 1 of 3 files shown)


riya-17 (approved):
LGTM

lib/pbench/agent/results.py
```diff
@@ -62,18 +62,17 @@ def make_result_tb(self):
...
         except FileNotFoundError:
             self.logger.debug(
                 "The {}/metadata.log file seems to be missing", pbench_run_name
             )
             sys.exit(0)
-
-        opts, _ = configtools.parse_args()
-        opts.filename = md_log
-        conf_md, _ = configtools.init(opts, None)
-        md_config = conf_md["pbench"]
```
dbutenhof, line 76: Well, OK; that's definitely squirrelly. Parse the standard Pbench command arguments, then punch the config object to point at the metadata.log instead of pbench-agent.cfg... clever, but smelly code.

dbutenhof (approved):

PR #2033, opened 2020-12-11
CORS fixes and a new API

dbutenhof:
In experimentation with a dashboard on my live server, I found that the current use of `@cross_origin` isn't actually working; I suspect this is because documentation suggests it needs to follow `@app.route(...)`, which we don't use.

Instead, change to a global CORS configuration triggered by wrapping the `app` object in a `CORS` object. This demonstrably works, and should keep us out of trouble for a while. (Note that, ideally, we want to set up Apache `mod_wsgi` to allow redirecting the default http and eventually https port to our app for `/api...` paths, which would mean our dashboard wouldn't need CORS -- very good from a security point of view -- though we may still want some sort of CORS support to allow integration of our server with other browser apps that might not be able to load their code from our server.)

In the process of my experimentation to get the dashboard started on 0.71 with Elasticsearch V7, I also wrote a second native server API to replace `fetchMonthIndices` ... because that and the existing `queryControllers` are required to get the dashboard up to the initial controller display. (I can't get further without Elasticsearch CORS, on which I'm still waiting.)

Rather than pursuing additional native APIs in the short term, I plan to rely on getting the RDU2 Elasticsearch V7 server configured for CORS and work directly on updating necessary dashboard queries. However I currently have dashboard code that uses my `queryMonthIndices` and `queryControllers` APIs, and when I get to the point of posting a dashboard PR I'll probably leave it that way as a milestone.

### DIFF
--- a/lib/pbench/server/api/resources/query_apis/query_month_indices.py
+++ b/lib/pbench/server/api/resources/query_apis/query_month_indices.py
@@ -1,0 +1,95 @@
+from flask import jsonify
+from flask_restful import Resource, abort
+import requests
+
+from pbench.server.api.resources.query_apis import get_es_url, get_index_prefix
+
+
+class QueryMonthIndices(Resource):
+    """
+    Abstracted Pbench API to get date-bounded controller data.
+    """
+
+    def __init__(self, config, logger):
+        self.logger = logger
+        self.es_url = get_es_url(config)
+        self.prefix = get_index_prefix(config)
+
+    def get(self):
+        """
+        GET to detect the month suffixes existing for run data indices.
+
+        Required headers include
+
+        Content-Type:   application/json
+        Accept:         application/json
+
+        The return payload is a list of "YYYY-mm" date strings corresponding
+        to the months in which tarballs were indexed into the appropriate
+        `run-data` index. (E.g., `drb.v6.run-data.2020-11`). Note that this
+        list is in DESCENDING order, so the earliest date is last.
+
+        NOTE: No authorization or input payload is required for this API.
+
+        NOTE: This is the format currently constructed by the Pbench
+        dashboard `src/model/datastore.js` fetchMonthIndices method, which
+        becomes part of the Redux state.
+
+        [
+            "2020-12",
+            "2020-11",
+            "2020-04"
+        ]
+        """
+        self.logger.info(
+            "QueryMonthIndices GET for prefix {}", self.prefix,
+        )
+
+        uri = f"{self.es_url}/_aliases"
+        try:
+            # query Elasticsearch
+            es_response = requests.get(uri, headers={"Accept": "application/json"})
+            es_response.raise_for_status()
+        except requests.exceptions.HTTPError as e:
+            self.logger.exception("HTTP error {} from Elasticsearch post request", e)
+            abort(502, message="INTERNAL ERROR")
+        except requests.exceptions.ConnectionError:
+            self.logger.exception(
+                "Connection refused during the Elasticsearch post request"
+            )
+            abort(502, message="Network problem, could not post to Elasticsearch")
+        except requests.exceptions.Timeout:
+            self.logger.exception(
+                "Connection timed out during the Elasticsearch post request"
+            )
+            abort(504, message="Connection timed out, could not post to Elasticsearch")
+        except requests.exceptions.InvalidURL:
+            self.logger.exception(
+                "Invalid url {} during the Elasticsearch post request", uri
+            )
+            abort(500, message="INTERNAL ERROR")
+        except Exception as e:
+            self.logger.exception(
+                "Exception {!r} occurred during the Elasticsearch post request", e
+            )
+            abort(500, message="INTERNAL ERROR")
+        else:
+            months = []
+            target = f"{self.prefix}.v6.run-data."
+            try:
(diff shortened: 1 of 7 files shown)


lib/pbench/server/api/resources/query_apis/query_month_indices.py
```diff
@@ -0,0 +1,101 @@
...
+        )
+
+        uri = f"{self.es_url}/_aliases"
+        try:
+            # query Elasticsearch
+            es_response = requests.get(uri, headers={"Accept": "application/json"})
+            es_response.raise_for_status()
+        except requests.exceptions.HTTPError as e:
+            self.logger.exception("HTTP error {} from Elasticsearch post request", e)
+            abort(500, message=f"HTTP error {e} from Elasticsearch")
```
portante, line 56: Not sure we should return a `500` error here.  This seems like a normal operation of this server.  Perhaps returning `502` instead?  Not sure we should pass along the exception, either.

lib/pbench/server/api/resources/query_apis/query_month_indices.py
```diff
@@ -0,0 +1,101 @@
...
+            abort(
+                504, message="Connection timed out, could not post to Elasticsearch",
+            )
+        except requests.exceptions.InvalidURL:
+            self.logger.exception(
+                "Invalid url {} during the Elasticsearch post request", uri
+            )
+            abort(
+                500,
+                message=f"Invalid Elasticsearch url {uri}, could not complete the post request",
```
portante, line 75: We have been following the principle of just returning `INTERNAL ERROR` for 500s.  This is a configuration problem for sure.

lib/pbench/server/api/resources/query_apis/query_month_indices.py
```diff
@@ -0,0 +1,101 @@
...
+                    if target in index:
+                        months.append(index.split(".")[-1])
+                # The dashboard converts the strings to int (removing the '-')
+                # and does a numeric sort; however since we always have a full
+                # "YYYY-mm" date, this isn't necessary.
+                months.sort(reverse=True)
+                self.logger.info("found months {!r}", months)
+            except KeyError:
+                self.logger.exception("ES response not formatted as expected")
+                abort(500, message="Elasticsearch response is odd")
```
portante, line 98: `INTERNAL ERROR`.

portante (changes requested):

lib/pbench/server/api/resources/query_apis/query_month_indices.py
dbutenhof: It's an interesting question; it really depends on what Elasticsearch is complaining about. We could just pass on whatever it said; I'm not sure if there's a useful payload on the HTTPError exception, and the basic niceness of `raise_for_status()` is not having to decode the HTTP error codes.

This is, in any case, just a copy-and-paste from our other ES interface code, so solving this in general across all of them probably makes sense. I can try to do that here, or we can deal with it separately.

lib/pbench/server/api/resources/query_apis/query_month_indices.py
dbutenhof: I'd say "migrating towards it"; but, yeah, we came up with that after I'd merged `query_controllers.py`, which is the origin of this code.

lib/pbench/server/api/resources/query_apis/query_month_indices.py
portante: Separately.

Let's just return an "INTERNAL ERROR" for now, we have the full exception in the logs.  It is unclear what the dashboard would do with the full message any ways.

dbutenhof: The Elasticsearch passthrough actually returns `es_response.status_code` ... that makes perfect sense for the passthrough, since the caller is directly responsible for any syntactic errors in the request payload. I added the same logic to the GraphQL passthrough as it makes equal sense there.

It's less clear for these, as Pbench is responsible for any syntactic errors, and we shouldn't be letting that happen. I'll go with 502 and `INTERNAL ERROR` for now, and make sure we're consistent.

portante (approved):

lib/pbench/server/api/resources/query_apis/query_month_indices.py
aquibbaig: The dashboard contains a component to handle errors with appropriate messages. This component can be made use of. https://github.com/distributed-system-analysis/pbench-dashboard/blob/master/src/pages/Exception/triggerException.js

aquibbaig: However currently, there is no way to include "messages" to be displayed along with this component. Just the http_status_code.

aquibbaig (approved):
LGTM

riya-17 (dismissed):

portante (changes requested):
Looks like it needs a rebase ...

portante (approved):

portante (approved):

npalaska (approved):
Looks fine, but I think we should probably update the flask-cors package name in the requirements file its `flask_cors` rn but should be `flask-cors` https://flask-cors.readthedocs.io/en/latest/

PR #2037, opened 2020-12-13
Address lgtm.com errors

portante:
See https://lgtm.com/projects/g/distributed-system-analysis/pbench/alerts/?mode=list&severity=error

### DIFF
--- a/agent/bench-scripts/templates/make-fio-jobfile.py
+++ b/agent/bench-scripts/templates/make-fio-jobfile.py
@@ -38,6 +38,7 @@ def replace_all(dct, old, new):
         if isinstance(dct[k], dict):
             dct[k] = replace_all(dct[k], old, new)
         elif dct[k] is not None:
+            assert isinstance(dct[k], str)
             old_val = dct[k]
             del dct[k]
             k = k.replace(old, new)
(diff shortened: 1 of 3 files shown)


riya-17 (approved):
LGTM

agent/bench-scripts/templates/make-fio-jobfile.py
```diff
@@ -51,7 +51,7 @@ def replace_val(dct, magic, delta):
     keys = list(dct.keys())
     for k in keys:
         if isinstance(dct[k], dict):
-            dct[k] = replace_val(dct[k], delta)
+            dct[k] = replace_val(dct[k], magic, delta)
         else:
             if (dct[k] is not None) and magic in dct[k]:
                 dct[k] = dct[k].replace(magic, delta[k])
```
dbutenhof, line 57: So the value of `dct[k]` is either a `dict` or a `list`?? It'd be more clear to have an `elif isinstance(dct[k], list)` and maybe an error `else` to be sure; but granting that's a hard convention unlikely to change (?), a comment might be good to explain precisely why we know/assume this.

dbutenhof (approved):
One comment for possible hardening and/or documentation, but all three lines are obviously changed for the better and two of them are entirely self-explanatory.

agent/bench-scripts/templates/make-fio-jobfile.py
portante: > So the value of `dct[k]` is either a `dict` or a `list`??

Not sure this is true.  You can have a `dict` or a `string` and the `in` operator will return `True` or `False`:

```
[pportant@gandalf pbench]$ python3
Python 3.8.6 (default, Sep 25 2020, 00:00:00)
[GCC 10.2.1 20200723 (Red Hat 10.2.1-1)] on linux
>>> foo = "is this in that string"
>>> "this" in foo
True
>>>
```

> It'd be more clear to have an `elif isinstance(dct[k], list)` and maybe an error `else` to be sure; but granting that's a hard convention unlikely to change (?), a comment might be good to explain precisely why we know/assume this.

I have added assertions to both methods to "document" the expected behavior.  And I verified that assertions do not fire via the existing unit tests we have (adding a `not` to the assertion causes the tests to fail).

ndokos (approved):

npalaska (approved):

Issue #1849, opened 2020-10-06
Consider the use of a queue for jobs to be done on the pbench server


portante:
We might want to consider [Kafka](https://kafka.apache.org/) or [python-rq](https://python-rq.org/) (based Redis).

npalaska:
Hey @portante can you explain this issue a little bit, do you mean all the cronjobs supposed to be in the queue structure or is regarding tracking jobs to be done on one tarfile after it gets uploaded?

dbutenhof:
This refers to the use of Redis pub/sub for Tool Meister communication, in particular.

At some point when we microcomponent-ize the server and get rid of the simplistic crontab launch intervals, we'll also want to use messaging to coordinate activities; it might not be a bad idea to choose something now that we can grow into later. Redis pub/sub may be OK while things are simple, (and encapsulating our current inline Redis use into a class, either our own or a package like the one referenced, might help with maintainability); but RabbitMQ or Kafka may give us more options in the long run.

portante:
Well, it was not my intention to refer to the Tool Meister, this is really about the pbench server cronjobs.

PR #2040, opened 2020-12-16
When installing the server, set the host-info state to maintenance

ndokos:
Currently, when installing the server, we set the `host-info` state to `active`, which allows `pbench-move`/`copy-results` to start sending tar balls to the server.

This PR sets the `host-info` state to `maintenance`.

It also adds playbooks and roles to explicitly enable and disable the server. The workflow now becomes

```
ansible-playbook ... pbench-server-disable.yml

ansible-playbook ... pbench-server-install.yml
# check crontab
ansible-playbook ... pbench-server-enable.yml
```

Unfortunately, the disable role has no easy way to make sure that the queue is drained: it can check the reception area to make sure it's empty, but various cron jobs may still be running processing tar balls (or might run in the next few minutes): it could check all the state directories to make sure they don't contain any tar ball symbolic links, but that seems a bit fragile (and perhaps expensive). So, for now at least, we punt and leave it up to the administrator to check and make sure that, after the disablement, all the tar balls have been processed before shutting off the `crontab` manually with

```
crontab -u pbench -r
```

Another possibility is to shut off the crontab anyway, and wait until all currently running `cron` jobs are done before proceeding. There might still be some tar balls in a "to-process" state, but they will be processed after the server is enabled again.

This is what's implemented here: enabling the server turns the `crontab` on and disabling the server turns the `crontab` off. The administrator still has to wait, checking that the various pbench processes that might be running have actually finished, before proceeding.

The crontab enablement/disablement is performed by separate roles: it's only the playbooks that impose the unitarity of enablement or disablement. After review, the roles are now protected, so that they only trigger when the roles are run as the "pbench" user: they are no-ops otherwise.

Privilege escalation in all the playbooks is now done through the `remote_user` variable. It is `root` for the installation playbook, but it is `pbench` for the enablement and disablement playbooks.

### DIFF
--- a/server/ansible/pbench-server-disable.yml
+++ b/server/ansible/pbench-server-disable.yml
@@ -1,0 +1,12 @@
+---
+- name: disable pbench-server
+  hosts: servers
+  remote_user: pbench
+
+  vars:
+    pbench_configuration_environment: "{{ cenv | default('production') }}"
+
+  roles:
+    - pbench-server-vars
+    - { role: pbench-server-set-state, pbench_server_state: off}
+    - pbench-server-crontab-disable
--- a/server/ansible/pbench-server-enable.yml
+++ b/server/ansible/pbench-server-enable.yml
@@ -1,0 +1,12 @@
+---
+- name: enable pbench-server
+  hosts: servers
+  remote_user: pbench
+
+  vars:
+    pbench_configuration_environment: "{{ cenv | default('production') }}"
+
+  roles:
+    - pbench-server-vars
+    - { role: pbench-server-set-state, pbench_server_state: on}
+    - pbench-server-crontab-enable
--- a/server/ansible/roles/pbench-repo-install/defaults/main.yml
+++ b/server/ansible/roles/pbench-repo-install/defaults/main.yml
@@ -1,8 +1,10 @@
 ---
-test_repo_enabled: 0
+fedoraproject_username: ndokos
+pbench_repo_url_prefix: https://copr-be.cloud.fedoraproject.org/results/{{ fedoraproject_username }}
 
 repos:
   - tag: pbench
+    user: "{{ fedoraproject_username }}"
     baseurl: "{{ pbench_repo_url_prefix }}/pbench/{{ distrodir }}"
     gpgkey: "{{ pbench_repo_url_prefix }}/pbench/pubkey.gpg"
     gpgcheck: 1
--- a/server/ansible/roles/pbench-server-activate-setup-results-host-info/tasks/main.yml
+++ b/server/ansible/roles/pbench-server-activate-setup-results-host-info/tasks/main.yml
@@ -1,4 +1,8 @@
 ---
+# Example: /var/www/html/pbench-results-host-info.versioned/pbench-results-host-info.URL002
+#          ^-----------^ ^-------------------pbench_host_info_prefix-------------------^
+#                ^
+#     httpd_document_root_dir
 - name: "make sure the directory exists"
   file:
     path: "{{ httpd_document_root_dir }}/{{ pbench_host_info_prefix | dirname }}"
@@ -27,10 +31,11 @@
   with_items:
     - "002"
 
-- name: "set up links to the active files"
+- name: "set up links to the maintenance files"
   file:
     path: "{{ httpd_document_root_dir }}/{{ pbench_host_info_prefix }}{{ item }}"
-    src:  "{{ pbench_host_info_prefix | basename }}{{ item }}.active"
+    src:  "{{ pbench_host_info_prefix | basename }}{{ item }}.maint"
     state: link
+    force: yes
   with_items:
     - "002"
--- a/server/ansible/roles/pbench-server-crontab-disable/tasks/main.yml
+++ b/server/ansible/roles/pbench-server-crontab-disable/tasks/main.yml
@@ -1,0 +1,10 @@
+---
+- name: "save the current crontab before disabling"
+  shell:
+    cmd: "crontab -l > {{ pbench_server_install_dir }}/lib/crontab/crontab.back.{{ ansible_date_time.iso8601 }}"
+  when: ansible_user_id == pbench_user
+
+- name: "disable crontab"
+  command:
+    cmd: crontab -r
+  when: ansible_user_id == pbench_user
--- a/server/ansible/roles/pbench-server-crontab-enable/tasks/main.yml
+++ b/server/ansible/roles/pbench-server-crontab-enable/tasks/main.yml
@@ -1,0 +1,5 @@
+---
+- name: "enable crontab for user 'pbench' - this role is supposed to be run as that user, *not* as root"
+  command:
+    cmd: crontab {{ pbench_server_install_dir }}/lib/crontab/crontab
+  when: ansible_user_id == pbench_user
--- a/server/ansible/roles/pbench-server-set-state/tasks/main.yml
+++ b/server/ansible/roles/pbench-server-set-state/tasks/main.yml
@@ -1,0 +1,18 @@
+---
+# Example: /var/www/html/pbench-results-host-info.versioned/pbench-results-host-info.URL002
+#          ^-----------^ ^-------------------pbench_host_info_prefix-------------------^
+#                ^
+#     httpd_document_root_dir
+- name: "debug"
+  debug:
+    msg: State is {{ pbench_server_state }}
+    verbosity: 1
+
+- name: "set up links to the appropriate file(s), depending on desired state (on or off)"
+  file:
+    path: "{{ httpd_document_root_dir }}/{{ pbench_host_info_prefix }}{{ item }}"
+    src:  "{{ pbench_host_info_prefix | basename }}{{ item }}.{{'active' if pbench_server_state else 'maint' }}"
+    state: link
+    force: yes
+  with_items:
+    - "002"
--- a/server/ansible/pbench-repo-install.yml
+++ b/server/ansible/pbench-repo-install.yml
@@ -3,8 +3,6 @@
 - name: install pbench.repo
   hosts: servers
   remote_user: root
-  become: yes
-  become_user: root
 
   roles:
     - pbench-repo-install
--- a/server/ansible/pbench-server-install.yml
+++ b/server/ansible/pbench-server-install.yml
@@ -2,8 +2,6 @@
 - name: install pbench-server
   hosts: servers
   remote_user: root
-  become: yes
-  become_user: root
 
   vars:
     pbench_configuration_environment: "{{ cenv | default('production') }}"


portante:
> empty, but various cron jobs may still be running processing tar balle

`s/balle/balls/`

server/ansible/roles/pbench-server-disable/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
...
+- name: "set up links to the maintenance files"
+  file:
+    path: "{{ httpd_document_root_dir }}/{{ pbench_host_info_prefix }}{{ item }}"
+    src:  "{{ pbench_host_info_prefix | basename }}{{ item }}.maint"
+    state: link
+  with_items:
+    - "002"
+
+# we should wait until the queue is drained before disabling crontab
+# but there is no easy way to do that: pung and do it by hand?
```
dbutenhof, line 15: I somehow like "pung" (though I have no idea what it means), but maybe you meant "punt"?? ;-)

server/ansible/roles/pbench-server-disable/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
...
+
+# - name: "disable crontab"
+#   command:
+#     cmd: crontan -u pbench -r
```
dbutenhof, line 19: `crontab`

dbutenhof (commented):
It's hard when our state is so scattered; maybe this is a problem we need to solve within a real server architecture.

We could certainly search all the state directories for links (we do this in several places already, after all), and look for active `pbench` processes... but it might be hard to coordinate reliably.

server/ansible/roles/pbench-server-disable/tasks/main.yml
ndokos: Fixed

server/ansible/roles/pbench-server-disable/tasks/main.yml
ndokos: Fixed

ndokos:
WDYT of this?

>Another possibility is to shut off the crontab anyway, and wait until
all currently running cron jobs are done before proceeding. There might
still be some tar balls in a to-process state, but they will be processed
after the server is enabled again.

It still punts, leaving some responsibility on the administrator, but basically they can just wait for `ps ... | grep pbench` to show no processes. The problem is that there might be an arbitrarily long delay in some of the processing of tar balls that were sent just before the disabling of the server.

server/ansible/roles/pbench-server-disable/tasks/main.yml
ndokos: Those darned off-by-1 errors

server/ansible/roles/pbench-server-disable/tasks/main.yml
ndokos: Another off-by-1 error - I refer you to my sig :)

portante:
> It still punts, leaving some responsibility on the administrator, but basically they can just wait for `ps ... | grep pbench` to show no processes. The problem is that there might be an arbitrarily long delay in some of the processing of tar balls that were sent just before the disabling of the server.

That is okay for now.  As we refactor the server code and insert a message passing system with queues of work to be done, we'll be able to add the ability to shutdown the pipeline and restart it from a saved state without waiting for tons of work to complete.

The `SIGQUIT` work for signal handling that Riya is doing in PR #1940 will also be part of that.

server/ansible/pbench-server-disable.yml
```diff
@@ -0,0 +1,13 @@
+---
+- name: disable pbench-server
+  hosts: servers
+  remote_user: root
```
portante, line 4: Shouldn't this be doable only using the `pbench` user?

server/ansible/roles/pbench-server-enable/tasks/main.yml
```diff
@@ -0,0 +1,17 @@
...
+- name: "set up links to the active files"
+  file:
+    path: "{{ httpd_document_root_dir }}/{{ pbench_host_info_prefix }}{{ item }}"
+    src:  "{{ pbench_host_info_prefix | basename }}{{ item }}.active"
+    state: link
+    force: yes
+  with_items:
+    - "002"
+
+- name: "enable cron jobs"
```
portante, line 15: Do we want to let the admin do this when they are ready?

server/ansible/roles/pbench-server-enable/tasks/main.yml
```diff
@@ -0,0 +1,17 @@
+---
+# Example: /var/www/html/pbench-results-host-info.versioned/pbench-results-host-info.URL002
+#          ^-----------^ ^-------------------pbench_host_info_prefix-------------------^
+#                ^
+#     httpd_document_root_dir
+- name: "set up links to the active files"
```
portante, line 6: This looks almost identical to the "disable" role.  Could we remove the duplication here?

portante (dismissed):

npalaska (commented):
yes I like this idea of keeping the server in maintenance state while installing, I think this will also be lot more streamlined when we come to addressing the issue https://github.com/distributed-system-analysis/pbench/issues/1849  use of a queue for jobs instead of using cronjobs.

server/ansible/pbench-server-disable.yml
ndokos: Yes. It does require a separate ssh-copy-id by the admin, but the playbooks should use the minimal privilege possible.

server/ansible/pbench-server-enable.yml
```diff
@@ -0,0 +1,13 @@
+---
+- name: enable pbench-server
+  hosts: servers
+  remote_user: root
```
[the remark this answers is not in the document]
ndokos: Yup - fixed.

server/ansible/roles/pbench-server-enable/tasks/main.yml
ndokos: Not sure: shouldn't enabling the server be a single step? E.g. the doc would say "run the pbench-server-enable playbook, after the crontab has been checked." Are you proposing a two-step enablement? Or something else?

server/ansible/roles/pbench-server-enable/tasks/main.yml
ndokos: Thinking about this...

server/ansible/roles/pbench-server-enable/tasks/main.yml
ndokos: I made the crontab steps to be separate roles but the enable/display playbook still run both roles. It would be easy to just create separate playbooks, but I'm not sure I see the point.

server/ansible/roles/pbench-server-enable/tasks/main.yml
ndokos: Fixed.

portante (changes requested):

riya-17 (dismissed):

portante (changes requested):
Needs a rebase now ...

ndokos:
Rebased.

portante (approved):

server/ansible/roles/pbench-repo-install/defaults/main.yml
```diff
@@ -1,8 +1,10 @@
 ---
-test_repo_enabled: 0
+fedoraproject_username: ndokos
+pbench_repo_url_prefix: https://copr-be.cloud.fedoraproject.org/results/{{ fedoraproject_username }}
```
webbnh, line 3: @ndokos, @portante, can one of you help me out, here?  It seems weird to have an individual developer's ID in this code.  I presume it's because the repo will have been built under Nick's ID, and, therefore, for any of us to access it, we'll need to reference it appropriately.  But, looking at the options, it looks like other projects are doing their builds under a project ID (at least, I assume that that is what the repos with the `@`-IDs are...).  Should we be following a similar approach?

server/ansible/roles/pbench-repo-install/defaults/main.yml
```diff
@@ -1,8 +1,10 @@
...
 
 repos:
   - tag: pbench
+    user: "{{ fedoraproject_username }}"
     baseurl: "{{ pbench_repo_url_prefix }}/pbench/{{ distrodir }}"
     gpgkey: "{{ pbench_repo_url_prefix }}/pbench/pubkey.gpg"
     gpgcheck: 1
```
webbnh, line 10: Continuing with my previous comment, why is this being done under Nick's ID?  Shouldn't this be `pbench` or some similar project ID?  (What would have to be done to make that work?)  I wouldn't want to have to change this code if Nick's account were to go away for some reason or if someone else takes over his duties.

server/ansible/roles/pbench-server-activate-setup-results-host-info/tasks/main.yml
```diff
@@ -27,10 +31,11 @@
   with_items:
     - "002"
 
-- name: "set up links to the active files"
+- name: "set up links to the maintenance files"
   file:
     path: "{{ httpd_document_root_dir }}/{{ pbench_host_info_prefix }}{{ item }}"
-    src:  "{{ pbench_host_info_prefix | basename }}{{ item }}.active"
+    src:  "{{ pbench_host_info_prefix | basename }}{{ item }}.maint"
```
webbnh, line 37: So, is this creating a symlink to the file created at lines 25-28?

I note that that file is set up without specifying a `mode:` or `group:` (unlike lines 19-20)...that seems like it could be a problem (i.e., if the message set up at line 27 is supposed to be widely accessible...).

server/ansible/roles/pbench-server-crontab-disable/tasks/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+- name: "disable crontab"
+  command:
+    cmd: crontab -r
```
webbnh, line 4: Shouldn't we be specifying a user name (either with a `-u` switch on the command or in a `user:` entry in the YAML)?

server/ansible/roles/pbench-server-crontab-enable/tasks/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+- name: "enable crontab"
+  command:
+    cmd: crontab {{ pbench_server_install_dir }}/lib/crontab/crontab
```
webbnh, line 4: Shouldn't we be specifying a user name (either with a -u switch on the command or in a user: entry in the YAML)?

server/ansible/roles/pbench-server-set-state/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
...
+#     httpd_document_root_dir
+- name: "debug"
+  debug:
+    msg: State is {{ pbench_server_state }}
+    verbosity: 1
+
+- name: "set up links to the appropriate file(s), depending on desired state (or or off)"
+  file:
+    path: "{{ httpd_document_root_dir }}/{{ pbench_host_info_prefix }}{{ item }}"
+    src:  "{{ pbench_host_info_prefix | basename }}{{ item }}.{{'active' if pbench_server_state else 'maint' }}"
```
webbnh, line 14: Given that you're adding this code, do you want to retain lines 34-41 in `pbench-server-activate-setup-results-host-info/tasks/main.yml`?

webbnh (dismissed):
Assuming that you're comfortable with your answers to my questions below, I have no problems with this change.

server/ansible/roles/pbench-repo-install/defaults/main.yml
ndokos: COPR does not let us create an org account, so using my ID has been the fallback. That will go away once the primary mode of installation is a container image. Not sure what the @-ID repos you mention are? What options are you looking at?

server/ansible/roles/pbench-server-crontab-disable/tasks/main.yml
ndokos: Probably: the role assumes that it is run under the "correct" identity - it is used by the disable playbook which specifies  `remote_user: pbench` explicitly, but that's a hidden coupling.

server/ansible/roles/pbench-server-crontab-enable/tasks/main.yml
ndokos: Yes.

dbutenhof (dismissed):

server/ansible/roles/pbench-repo-install/defaults/main.yml
webbnh: https://copr-be.cloud.fedoraproject.org/results/

server/ansible/roles/pbench-server-crontab-disable/tasks/main.yml
webbnh: My principal concern is that `crontab -r` is kind of a big hammer, and I don't want it to land on anyone's thumb.

ndokos: We could run the playbook as root and use -u with the appropriate user, but we were trying to get away from running them as root if we don't have to. So maybe check the id of the user running it and only do it if it is the "right" user?

server/ansible/roles/pbench-repo-install/defaults/main.yml
portante: @webbnh, can you file an issue to investigate the use of project builds in Fedora Project dot Org?

server/ansible/roles/pbench-server-activate-setup-results-host-info/tasks/main.yml
portante: Typically this is not a problem because symlinks are created with wide open permissions by default.

@webbnh, can you file an issue to track this separately?

server/ansible/roles/pbench-server-crontab-disable/tasks/main.yml
portante: I have never been comfortable with automated installs of crontabs via Ansible.

It is probably worth issuing a `crontab -l` first to save the existing crontab locally before performing a `crontab -r` here.

portante (changes requested):
It think it is worth having the crontab disable role save the existing state of the crontab locally so that it is not lost.

portante (approved):

portante (approved):

server/ansible/roles/pbench-server-crontab-disable/tasks/main.yml
webbnh: Yes, I was afraid that `crontab -u` was a privileged command, but I couldn't easily prove that.  However, adding the `-u` alone isn't much of a protection.  Backing up the previous contents is a much better solution.

webbnh: > maybe check the id of the user running it and only do it if it is the "right" user?

I'm a big fan of assertion checks, and of detecting inconsistencies and preventing problems, particularly ones which are hard to diagnose after the fact.  So, I would be in favor of adding a check here, or at least an alert if we're not in the right context.  On the other hand, I'm not closely familiar with the operating environment here -- if it's structurally "impossible" to be running under the wrong username, then we can settle for just the strategy of creating a backup (in addition to enabling recovery, that will leave a trace of who actually ran the command and when).

npalaska (approved):

server/ansible/roles/pbench-server-crontab-disable/tasks/main.yml
webbnh: Bump (so this thread is attached to my review).

webbnh (dismissed):
I approve this PR, unless you want to add the assertion check that we discuss below.

portante (changes requested):
Needs a rebase, and if you could squash the commits to one, thanks!

server/ansible/roles/pbench-server-set-state/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
...
+# Example: /var/www/html/pbench-results-host-info.versioned/pbench-results-host-info.URL002
+#          ^-----------^ ^-------------------pbench_host_info_prefix-------------------^
+#                ^
+#     httpd_document_root_dir
+- name: "debug"
+  debug:
+    msg: State is {{ pbench_server_state }}
+    verbosity: 1
+
+- name: "set up links to the appropriate file(s), depending on desired state (or or off)"
```
riya-17, line 11: I think here is a Typo mistake s/or/on

riya-17 (approved):

ndokos:
Rebased and squashed.

server/ansible/roles/pbench-server-set-state/tasks/main.yml
ndokos: Fixed.

server/ansible/roles/pbench-server-crontab-disable/tasks/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+- name: "save the current crontab before disabling"
+  shell:
+    cmd: "crontab -l > {{ pbench_server_install_dir }}/lib/crontab/crontab.back.{{ ansible_date_time.iso8601 }}"
+  when: ansible_user_id == pbench_user
+
+- name: "disable crontab"
+  command:
+    cmd: crontab -r
+  when: ansible_user_id == pbench_user
```
dbutenhof, line 10: Q: what happens here when `ansible_user_id` != `pbench_user`?? Does the whole role just silently do nothing?

I suppose the Ansible task output will say that it was skipped, but I wonder if anyone will actually notice... ;-)

dbutenhof (approved):

portante (approved):

Issue #1878, opened 2020-10-14
pprof-datalog test is flaky, sometimes fails to remove the ansible tmp files


portante:
Sometimes the `pprof` datalog command will fail with:

```
Testing pprof-datalog under setup pprof ...
diff -cr tests/pprof/gold/stdout /var/tmp/pbench-test-tools-datalog/pprof/stdout
*** tests/pprof/gold/stdout     2020-10-09 22:05:47.899240259 +0000
--- /var/tmp/pbench-test-tools-datalog/pprof/stdout     2020-10-14 01:42:56.475782472 +0000
***************
*** 0 ****
--- 1 ----
+ Unable to remove temporary file [Errno 2] No such file or directory: '${HOME}/.ansible/tmp/ansible-local-3618583w6b6x_gj'
FAIL - pprof
```

This seems to be due to a race condition between the child processes and the parent cleaning up the ansible context.

portante:
Fixed by #2653.

portante:
This could not have been fixed by #2653 since that is a fix for the `oc` tool, not the `pprof` tool.

portante:
Fixed by #2041.

Issue #2026, opened 2020-12-07
Looks like we end up leaving turbostat processes lying around from our test environment.


portante:
```
[root@gprfc058 var]# lsof | grep -F /var/tmp
turbostat  5503          root  cwd       DIR              253,3          6   33578606 /var/tmp/tm.default.5379.mx8bhjk9/gprfc058/turbostat (deleted)
turbostat  5503          root    2w      REG              253,3          0   33578610 /var/tmp/tm.default.5379.mx8bhjk9/gprfc058/turbostat/turbostat-stderr.txt (deleted)
pbench-lo  5504          root  cwd       DIR              253,3          6   33578606 /var/tmp/tm.default.5379.mx8bhjk9/gprfc058/turbostat (deleted)
pbench-lo  5504          root    1w      REG              253,3 3146674072   33578609 /var/tmp/tm.default.5379.mx8bhjk9/gprfc058/turbostat/turbostat-stdout.txt (deleted)
pbench-lo  5504          root    2w      REG              253,3          0   33578610 /var/tmp/tm.default.5379.mx8bhjk9/gprfc058/turbostat/turbostat-stderr.txt (deleted)
turbostat  7775          root  cwd       DIR              253,3          6         83 /var/tmp/tm.default.7648.85xdyfjn/gprfc058/turbostat (deleted)
turbostat  7775          root    2w      REG              253,3          0         89 /var/tmp/tm.default.7648.85xdyfjn/gprfc058/turbostat/turbostat-stderr.txt (deleted)
pbench-lo  7776          root  cwd       DIR              253,3          6         83 /var/tmp/tm.default.7648.85xdyfjn/gprfc058/turbostat (deleted)
pbench-lo  7776          root    1w      REG              253,3 3146226692         88 /var/tmp/tm.default.7648.85xdyfjn/gprfc058/turbostat/turbostat-stdout.txt (deleted)
pbench-lo  7776          root    2w      REG              253,3          0         89 /var/tmp/tm.default.7648.85xdyfjn/gprfc058/turbostat/turbostat-stderr.txt (deleted)
turbostat  9535          root  cwd       DIR              253,3          6   50331914 /var/tmp/tm.default.9411.6p6z50r1/gprfc058/turbostat (deleted)
turbostat  9535          root    2w      REG              253,3          0   50331918 /var/tmp/tm.default.9411.6p6z50r1/gprfc058/turbostat/turbostat-stderr.txt (deleted)
[...]

PR #2041, opened 2020-12-16
Stabilize pprof test, correct perf record usage

portante:
We lump these five commits together since they greatly increase the stability of the code, and / or are small enough that it was not worth submitting a separate PR.  All of these issues came from work on PR #2005.

The first commit made a minor correction to the `fio-shared-fs.job` file to quiet warnings in the RPM sausage making.

The second commit stabilizes the `pprod-datalog` unit tests removing the random failures.

The third commit addresses a bug with `pbench-register-tool-set`, where it was incorrectly registering the `perf` tool with the `record` command in the options.

The fourth commit removes an unnecessary python dependency between the tools and the `pbench` module.

The fifth commit addresses issue #2026, fixing the `turbostat` and `pidstat` tools.


agent/tool-scripts/datalog/pprof-datalog
```diff
@@ -56,44 +57,23 @@ else:
...
 
-try:
-    from ansible.parsing.dataloader import DataLoader
-    from ansible.inventory.manager import InventoryManager
-except ImportError as exc:
-    logging.error(
-        "[%s]: required Ansible python3 modules not available, '%s'", PROG, exc
-    )
-    sys.exit(1)
+import subprocess  # noqa E402
```
dbutenhof, line 60: Why not just pop all the imports up to the top where they belong? They don't seem to be truly "conditional" since there are only abnormal error exits before them.

dbutenhof (dismissed):
The motivation for peeling out the `pprof-master-n-nodes` code and translation into JSON isn't obvious, but I'm not objecting...

portante:
> The motivation for peeling out the `pprof-master-n-nodes` code and translation into JSON isn't obvious, but I'm not objecting...

Does the commit message from that first commit make it clearer?

----
**Stablize the pprof unit tests**

This is really ugly but the solution is simple: run the ansible module work in a sub-process so that the wacky way ansible handles closing down its context does not run afoul of child processes.

Basically, by running the gathering of the master and nodes lists in a single sub-process initially, the ansible module can clean up its state in one process.  The other pprof children will never have loaded any ansible modules, thus avoiding this conflict.

agent/tool-scripts/datalog/pprof-datalog
portante: I have been desperately trying to avoid making changes to these tool datalogs where possible when the changes are cosmetic.  These tools are not heavily used, and very hard to verify they work correctly, and I'd like to eventually drop them entirely.

Historically, this code was contributed by somebody who wanted this data collected by the pbench-agent in a Kubernetes environment, but then they have stopped using the pbench-agent, and this is now "dying on the vine."

Unfortunately, the test for this code keeps failing randomly, impacting development cycles.  So the changes here are trying to be the bare minimum to avoid other bad behaviors.

dbutenhof:
> > The motivation for peeling out the `pprof-master-n-nodes` code and translation into JSON isn't obvious, but I'm not objecting...
> 
> Does the commit message from that first commit make it clearer?


Yeah, it actually does; I just hadn't gone through the individual commits, so I didn't see that.

riya-17 (approved):
Looks Good

dbutenhof (approved):

portante:
Fixes #1878.

PR #2050, opened 2020-12-17
Add coverage support

portante:
Add the ability to run pytest coverage report.

Jenkins integration take from https://mdyzma.github.io/2017/10/14/python-app-and-jenkins/.

Signed-off-by: Charles Short <chucks@redhat.com>

PR #2052, opened 2020-12-18
Add coverage support

portante:
Add the ability to run pytest coverage report.

This is yet another redo of PR #2050 because that was merged in haste with an additional unwanted commit.

### DIFF
--- a/jenkins/python-setup.sh
+++ b/jenkins/python-setup.sh
@@ -1,6 +1,6 @@
 _prefix="${1}"
 
-pip3 install --no-cache-dir --prefix="${_prefix}" -r ${progdir}/lint-requirements.txt -r ${progdir}/agent/requirements.txt -r ${progdir}/server/requirements.txt -r ${progdir}/agent/test-requirements.txt -r ${progdir}/server/test-requirements.txt
+pip3 install --no-cache-dir --progress-bar off --no-color --prefix="${_prefix}" -r ${progdir}/lint-requirements.txt -r ${progdir}/agent/requirements.txt -r ${progdir}/server/requirements.txt -r ${progdir}/agent/test-requirements.txt -r ${progdir}/server/test-requirements.txt
 
 _pdir=${_prefix}/bin
 if [[ ":${PATH:-}:" != *":${_pdir}:"* ]]; then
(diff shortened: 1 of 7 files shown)


ndokos (approved):

jenkins/run-pytests
```diff
@@ -18,10 +20,15 @@ if [[ ${flake_sts} -ne 0 ]]; then
     printf -- "FAILED: flake8 --verbose .\n" >&2
 fi
 printf -- "\n"
-pytest ./lib/pbench/test/unit/common ./lib/pbench/test/unit/agent ./lib/pbench/test/unit/server ./lib/pbench/test/functional
+pytest --cov --cov-append \
+    --cov-report xml:/tmp/${USER}/cov/cov.xml \
```
dbutenhof, line 24: We have the same inconsistent usage of `/tmp/${USER}` vs `{envdir}`... did you revert that deliberately?

portante: I don't think so, but maybe this was a good thing to have made this mistake because it gives me a chance to explain this behavior.

The non-`jenkins-` `pytest` runs are expected to be run by users, so the report generated is an `html` report, intended to be more human readable than the XML.  The Jenkins runs, on the other hand, need the XML to upload for JUnit integration in the Jenkins dashboard, and there is no facility for capturing the `html` directory hierarchy.

So in `tox.ini` `py3-agent` and `py3-server` capture separate coverage for each of them as an `html` report.  The Jenkins job runs both the agent and the server code together, and generates one XML report for uploading to the Jenkins instance.

jenkins/python-setup.sh
```diff
@@ -1,6 +1,6 @@
 _prefix="${1}"
 
-pip3 install --no-cache-dir --prefix="${_prefix}" -r ${progdir}/lint-requirements.txt -r ${progdir}/agent/requirements.txt -r ${progdir}/server/requirements.txt -r ${progdir}/agent/test-requirements.txt -r ${progdir}/server/test-requirements.txt
+pip3 install --no-cache-dir --progress-bar off --no-color --prefix="${_prefix}" -r ${progdir}/lint-requirements.txt -r ${progdir}/agent/requirements.txt -r ${progdir}/server/requirements.txt -r ${progdir}/agent/test-requirements.txt -r ${progdir}/server/test-requirements.txt
```
portante, line 3: This is a desparate attempt to try to remove the escape sequences when running in the container, but for the life of me I can't figure out why the still appear.

ndokos (approved):

dbutenhof (approved):

PR #2053, opened 2020-12-18
Modifications to template data types for dashboard

dbutenhof:
These changes to the document templates allow dashboard queries to sort or aggregate on these fields. One field changes from `"text"` to `"keyword"` (because there seems to be unlikely to be any value to `@timestamp_original` in full text search), whereas another uses the Elasticsearch ability to double-map a field as both `"text"` and `"keyword"` so that the description string could be hit by a full text search and can also be used for sorting.

I've re-indexed my test tarballs with these changes and verified that my modified dashboard code works now.

### DIFF
--- a/server/bin/gold/test-7.0.1.txt
+++ b/server/bin/gold/test-7.0.1.txt
@@ -2335,7 +2335,8 @@ Template: pbench-unittests.v5.result-data
                 "type": "date"
             },
             "@timestamp_original": {
-                "type": "text"
+                "index": false,
+                "type": "keyword"
             },
             "iteration": {
                 "properties": {
@@ -2445,7 +2446,8 @@ Template: pbench-unittests.v5.result-data-sample
                 "type": "date"
             },
             "@timestamp_original": {
-                "type": "text"
+                "index": false,
+                "type": "keyword"
             },
             "benchmark": {
                 "properties": {
@@ -3072,6 +3074,11 @@ Template: pbench-unittests.v5.result-data-sample
                         "type": "long"
                     },
                     "measurement_title": {
+                        "fields": {
+                            "raw": {
+                                "type": "keyword"
+                            }
+                        },
                         "type": "text"
                     },
                     "measurement_type": {
--- a/server/lib/mappings/result-data-sample.json
+++ b/server/lib/mappings/result-data-sample.json
@@ -9,7 +9,8 @@
             "format": "dateOptionalTime"
         },
         "@timestamp_original": {
-            "type": "text"
+            "type": "keyword",
+            "index": false
         },
         "@generated-by": {
             "type": "keyword"
@@ -73,7 +74,12 @@
                     "type": "long"
                 },
                 "measurement_title": {
-                    "type": "text"
+                    "type": "text",
+                    "fields": {
+                        "raw": {
+                            "type": "keyword"
+                        }
+                    }
                 },
                 "category": {
                     "type": "keyword"
--- a/server/lib/mappings/result-data.json
+++ b/server/lib/mappings/result-data.json
@@ -9,7 +9,8 @@
             "format": "dateOptionalTime"
         },
         "@timestamp_original": {
-            "type": "text"
+            "type": "keyword",
+            "index": false
         },
         "result_data_sample_parent": {
             "type": "keyword"


server/lib/mappings/result-data.json
```diff
@@ -9,7 +9,7 @@
             "format": "dateOptionalTime"
         },
         "@timestamp_original": {
-            "type": "text"
+            "type": "keyword"
```
portante, line 12: I don't think we want to use `keyword` here for what amounts to random values.  Is the dashboard using this field?  If so, that is probably a bug.

server/lib/mappings/result-data-sample.json
```diff
@@ -73,7 +73,12 @@
                     "type": "long"
                 },
                 "measurement_title": {
-                    "type": "text"
+                    "type": "text",
+                    "fields": {
+                        "measurement_title_kw": {
```
portante, line 78: Doesn't this make it so that the reference is `...measurement_title.measurement_title_kw`?  If so, could we consider something like:

```
"measurement_title": {
    "type": "keyword",
    "fields": {
        "raw": {
            "type": "text"
        }
    }
}
```

portante (changes requested):

server/lib/mappings/result-data.json
dbutenhof: This decision was more of a "we don't need full text search support" for this. "keyword" isn't tokenized, which seemed like a win if a small one. I suppose "@timestamp_original": {"type": "keyword", "index": false} might be a step further.

server/lib/mappings/result-data-sample.json
dbutenhof: It actually does; my brain apparently atrophied during the break, because the usage I _remembered_ seeing when I researched this way back last year was that ES didn't scope the name. I can vividly recall an example. But I can't find it now. The subfield name __is__ scoped, and thus you're right that simply naming it `"raw"` makes more sense.

server/lib/mappings/result-data-sample.json
```diff
@@ -73,7 +73,12 @@
                     "type": "long"
                 },
                 "measurement_title": {
-                    "type": "text"
+                    "type": "text",
+                    "fields": {
+                        "raw": {
```
portante, line 78: Perhaps `raw` indicates something besides a `keyword`?  What about changing the field to be a `keyword` and adding a `text` sub-field which is of type `text`?

portante (changes requested):

server/lib/mappings/result-data-sample.json
dbutenhof: We need a `keyword` type to sort/aggregate, and this is used for aggregation. Now, whether we really _also_ need to support full text search on this is a separate question. The search query is a bit more free-form, but "title" just suggests something that should be text searchable to me.

Which field is on the top really doesn't make a difference, logically.

server/bin/gold/test-7.0.1.txt
```diff
@@ -2335,7 +2335,7 @@ Template: pbench-unittests.v5.result-data
                 "type": "date"
             },
             "@timestamp_original": {
-                "type": "text"
+                "type": "keyword"
```
portante, line 2338: Did we want to add the `"index": "false"` here?

portante (changes requested):

ndokos (dismissed):

server/bin/gold/test-7.0.1.txt
dbutenhof: I don't know if there's any significant benefit, and I'd have to make another testing round to be sure it doesn't break anything; but we could do that if someone feels strongly about it.

portante: It would be good to remove the indexes for this since this field will be in many documents.

dbutenhof: Likely true of other fields as well; we really should do a full review. I'll plan on setting something up soon after I get this pass stabilized.

But, for now, sure; I'll try marking '@timestamp_original' as "index": false here and in `result-data.json` as well.

portante (approved):

riya-17 (approved):

PR #2056, opened 2020-12-21
Backport f33 centos8 cont fixes

portante:
This is a back port of commits that are in PR #2059 for the `b0.69` branch.

We enable support for Fedora 33 container builds, and fixes for `locale` issues on CentOS 8, along with the apparent rename of the `powertools` repo (used to be `PowerTools`).

### DIFF
--- a/agent/containers/images/Makefile
+++ b/agent/containers/images/Makefile
@@ -39,9 +39,9 @@ _WORKLOAD_RPMS = fio uperf
 _ALL_RPMS = ${_TOOL_RPMS} ${_WORKLOAD_RPMS}
 
 # By default we only build images for the following distributions:
-_DISTROS = centos-8 centos-7 fedora-32 fedora-31
+_DISTROS = centos-8 centos-7 fedora-33 fedora-32
 
-all: $(foreach distro, ${_DISTROS}, ${distro}-all-tagged)
+all: all-tags $(foreach distro, ${_DISTROS}, ${distro}-all-tagged)
 
 #+
 # Tagging targets
@@ -49,6 +49,12 @@ all: $(foreach distro, ${_DISTROS}, ${distro}-all-tagged)
 
 # Add the "latest" tag to the local images.
 tag-latest: $(foreach distro, ${_DISTROS}, ${distro}-tag-latest)
+
+# Add the "beta" tag to the local images.
+tag-beta: $(foreach distro, ${_DISTROS}, ${distro}-tag-beta)
+
+# Add the "alpha" tag to the local images.
+tag-alpha: $(foreach distro, ${_DISTROS}, ${distro}-tag-alpha)
 
 # Add the "v<Major>-latest" tag to the local images.
 tag-major: $(foreach distro, ${_DISTROS}, ${distro}-tag-major)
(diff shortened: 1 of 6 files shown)


agent/containers/images/verify-tags
```diff
@@ -0,0 +1,23 @@
...
+fi
+
+# Use the first argument as the check-sum to verify against.
+cksum="$(cksum ${1} | awk '{print $1}')"
+shift
+
+ret_val=0
+while [[ ! -z "${1}" ]]; do
+    new_cksum="$(cksum ${1} | awk '{print $1}')"
+    shift
```
riya-17, line 15: @portante I am unable to understand how cksum in ln.9 is different from new_cksum in ln.14

riya-17 (commented):
looks good

agent/containers/images/verify-tags
portante: The `shift` built-in of `bash` removes the first command line argument and "shifts" the remaining so that the previously second argument is now the first, and so on.

agent/containers/images/repo.yml.j2
```diff
@@ -1,6 +1,6 @@
 ---
 repos:
-  - tag: pbench
+  - tag: pbench{{ test_suffix }}
```
ndokos, line 3: Where does this fit in? Is there some role that got modified and should be part of this PR?

ndokos (changes requested):

agent/containers/images/repo.yml.j2
portante: See line 215 of the `Makefile`.

agent/containers/images/Makefile
```diff
@@ -39,9 +39,9 @@ _WORKLOAD_RPMS = fio uperf
 _ALL_RPMS = ${_TOOL_RPMS} ${_WORKLOAD_RPMS}
 
 # By default we only build images for the following distributions:
-_DISTROS = centos-8 centos-7 fedora-32 fedora-31
+_DISTROS = centos-8 centos-7 fedora-33 fedora-32
```
dbutenhof, line 42: Hmm. Not really a big deal here since building container images for the agent is recent and not widely distributed; but in general the idea of changing the support matrix for an old release (0.69) seems ... odd.

agent/containers/images/verify-tags
dbutenhof: I had to read this twice to figure out the concern and the explanation. `verify-tags` takes a list of file arguments, calculates the checksum of the first file, and compares all subsequent file checksums against that value. Without searching for and analyzing the uses of this script it's not clear exactly why we're doing that or what it means, which I think was the point of @riya-17's comment.

Could we have some more clear documentation?

agent/containers/images/Makefile
portante: We have not choice in the matter.  COPR no longer allows us to build Fedora 31 RPMs.  It is end-of-life.  We can setup our own build system for past versions, but that is well beyond the scope of this PR.

agent/containers/images/verify-tags
portante: This is a backport of existing code in `master` so if we need to make changes to this script, we'll have to do it there first.

As a matter of behavior, `verify-tags` (called from the `Makefile` @ [line 178](https://github.com/portante/pbench/blob/backport-f33-centos8-cont-fixes/agent/containers/images/Makefile#L178)) verifies that the generated `*-tags.lis` files are all identical.  If they are not identical, it means we have different versions of software built across the release targets.

Each COPR build target ("Fedora 33, Fedora 32, EPEL 8, EPEL 7") has a separate RPM repository in COPR.  Each `<target>-tags.lis` file is generated by inspecting that repository for that target (see `Makefile` @ [line 173](https://github.com/portante/pbench/blob/backport-f33-centos8-cont-fixes/agent/containers/images/Makefile#L173)).

riya-17 (approved):

dbutenhof (approved):

ndokos (approved):

dbutenhof (approved):

PR #2062, opened 2021-01-09
pbench-fio: compute localhost latency profiles

### LINKED ISSUE #2061: pbench-fio latency histograms only get generated when passing --client
To understand, inter alia, public cloud storage performance under etcd load (and thus whether or not it can handle the load of being used on an Openshift controller node) we need to track the 99th percentile latency for fdatasync() calls.  In trying to simulate that we discovered that the latency data wasn't being graphed, and in a chat with @portante he found that the code is there to do it but surprisingly only runs if the benchmark is run using --client.  The data is perfectly valid for a single host, so there's no reason for that restriction.

portante:
Fixes #2061.

### DIFF
--- a/agent/bench-scripts/pbench-fio
+++ b/agent/bench-scripts/pbench-fio
@@ -677,7 +677,7 @@ function fio_run_job() {
 	pbench-send-tools --group=$tool_group --dir=$benchmark_results_dir
 	pbench-postprocess-tools --group=$tool_group --dir=$benchmark_results_dir
 
-	local histogram_params="--time-quantum ${histogram_interval_sec} --percentiles 0 50 90 95 99 100"
+	local histogram_params="--time-quantum ${histogram_interval_sec} --percentiles 0 50 90 95 99 99.5 100"
 	local histogram_cmd="fio-histo-log-pctiles.py ${histogram_params} -- "
 
 	# First move all fio log files to the appropriate client directory.
--- a/agent/bench-scripts/pbench-fio
+++ b/agent/bench-scripts/pbench-fio
@@ -630,8 +630,6 @@ function fio_run_job() {
 				exit 1
 			fi
 		done
-	else
-		mkdir -p $benchmark_results_dir/clients/localhost
 	fi
 
 	local client_opts=""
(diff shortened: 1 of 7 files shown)


agent/bench-scripts/pbench-fio
```diff
@@ -720,6 +721,18 @@ function fio_run_job() {
...
 	else
 		/bin/mv $benchmark_results_dir/*.log $benchmark_results_dir/clients/localhost/ 2> /dev/null
+		# compute localhost latency percentiles
+		if grep -q "^log_hist_msec" $fio_job_file; then
+			does_wildcard_exist $benchmark_results_dir/clients/localhost/'fio_clat_hist.*.log*'
+			if [[ $? -eq 0 ]]; then
+				mkdir -p $benchmark_results_dir/clients/localhost/hist
+				# compute per-client latency percentiles
+				$histogram_cmd $benchmark_results_dir/clients/localhost/fio_clat_hist.*.log* >$benchmark_results_dir/clients/localhost/hist/hist.csv
+				$script_path/postprocess/$benchmark-postprocess-viz.py --job-file=$fio_job_file ${histogram_params} -- $benchmark_results_dir/clients/localhost/hist
```
dbutenhof, line 731: There seem to be unnecessary differences from the clients code above. Specifically, why put the results into `clients/localhost/hist` instead of `hist`? Without that difference it seems that this post-processing step would be identical with or without `clients` and could just be factored out of the `if/else`.

agent/bench-scripts/tests/pbench-fio/test-04.txt
```diff
@@ -24,14 +24,19 @@ size=4096M
 numjobs=1
 
 running fio job: /var/tmp/pbench-test-bench/pbench-agent/fio_test-04_1900.01.01T00.00.00/1-rw-4KiB/fio.job (sample1)
+[warn][1900-01-01T00:00:00.000000] log_hist_msec specified in job file but failed to find any log files (.../clients/localhost/fio_clat_hist.*.log*) to process for histograms
```
dbutenhof, line 27: So awkward to put together all the pieces of the test command structure to figure out whether these gold file changes are good. I'm assuming this must indicate a single-host fio where we were previously not looking for these logs; but there seem to be rather a lot of them. And are there (single host) test cases were we do have the logs to validate that they're processed now?

agent/bench-scripts/pbench-fio
portante: Yeah, it's ugly.  I'll attempt a refactor to unify this behavior.

agent/bench-scripts/tests/pbench-fio/test-04.txt
portante: This was an attempt to make a minimal change without opening a can of worms.  Unfortunately, it failed.

By making this change, almost all of our fio tests have latency profiles that were not being computed, and so we have this deluge of changes in the test environment.

frival:
From our earlier conversation:
+1 The only thing that makes me a little sad is that it goes from 99 to 100 without an intervening step.  In RGW land we found 99.5 to be a good leading indicator that any additional blip could cause 99 to hiccup.  I won't complain here though as 99% (ha, the irony) of use cases won't need it and the rare ones that do should still be able to get it from the raw data IIUC.

portante (commented):
Ready for review, added support for 99.5 percentile handling.

dbutenhof (approved):

portante:
@frival, we'll merge this once you sign off that this change provides what you are looking for, and we get another reviewer for this change.

zulcss (approved):
lgtm

dbutenhof (approved):

dbutenhof (approved):

npalaska (approved):

riya-17 (approved):

agent/bench-scripts/pbench-fio
```diff
@@ -679,47 +677,62 @@ function fio_run_job() {
...
-			/bin/mv $benchmark_results_dir/*.log.$client $benchmark_results_dir/clients/$client/ 2> /dev/null
+	local histogram_params="--time-quantum ${histogram_interval_sec} --percentiles 0 50 90 95 99 99.5 100"
+	local histogram_cmd="fio-histo-log-pctiles.py ${histogram_params} -- "
+
+	# First move all fio log files to the appropriate client directory.
+	if [ ! -z "${clients}" ]; then
+		local client
+		for client in `echo ${clients} | sed -e s/,/" "/g`; do
+			mkdir -p ${benchmark_results_dir}/clients/${client}
+			/bin/mv ${benchmark_results_dir}/*.log.${client} ${benchmark_results_dir}/clients/${client}/ 2> /dev/null
```
ndokos, line 688: Why are we using the full path to `mv` here and below? Guessing unit-test related: mocking `mv` does not affect these?

ndokos (approved):

ndokos (approved):

agent/bench-scripts/pbench-fio
portante: I am guessing, but not sure entirely.  I tried not to change too much here.

PR #2065, opened 2021-01-12
Treat files with suffix .tgz, .tar, or .tar.<something> as tarfiles

RobertKrawitz:
Don't require use of --tarfile to specify tarfiles.  --tarfile is still useful if there's a tarfile that doesn't have a canonical name.


portante (approved):

portante:
See also #2066.

PR #2068, opened 2021-01-12
Add API reverse proxy configuration to default Ansible Pbench deployment scripts

dbutenhof:
The Pbench server API listens on port 8001; we want to be able to access the API externally through normal HTTP/HTTPS.

With an external reverse proxy gateway (e.g., NGINX) this can be accomplished by redirecting (e.g.) `pbench.example.com:/api/` to the actual server at port 8001 and opening the 8001/tcp port in the server firewall.

Our default Ansible deployment mechanism configures a local Apache which is directly accessed from outside. This PR adds an Apache reverse proxy to allow external API access on port 80.

With these changes, the Pbench gunicorn wsgi server will still listen on port 8001, but this port won't need to be opened externally as Apache will proxy `/api/` to port 8001. This means that all external accesses will go to port 80 (http) or (when we get there) 433 (https).

In the short term this affects the `pbench_server` configuration in the dashboard `endpoints.js`. Once #2024 (server-side endpoint configuration and metadata) is merged, I plan to change the dashboard to use that, simply `fetch`-ing that from `window.origin` and assigning it to `window.endpoints` within the launch html page, removing the need to manage a local static `endpoints.js` entirely.

### DIFF
--- a/lib/pbench/cli/server/shell.py
+++ b/lib/pbench/cli/server/shell.py
@@ -25,8 +25,8 @@ def main():
         print(e)
         sys.exit(1)
     try:
-        port = str(server_config.get("pbench-server", "rest_port"))
         host = str(server_config.get("pbench-server", "bind_host"))
+        port = str(server_config.get("pbench-server", "bind_port"))
         workers = str(server_config.get("pbench-server", "workers"))
     except (NoOptionError, NoSectionError) as e:
         print(f"{__name__}: ERROR: {e.__traceback__}")
--- a/server/ansible/roles/pbench-server-activate-httpd-setup-restart/tasks/templates/etc/httpd/conf.d/pbench.conf.j2
+++ b/server/ansible/roles/pbench-server-activate-httpd-setup-restart/tasks/templates/etc/httpd/conf.d/pbench.conf.j2
@@ -1,0 +1,6 @@
+<VirtualHost *:80>
+    ProxyPreserveHost On
+    ProxyPass /api/ http://{{ pbench_host }}:{{ bind_port }}/api/
+    ProxyPassReverse /api/ http://{{ pbench_host }}:{{ bind_port }}/api/
+    ProxyPass / !
+</VirtualHost>
--- a/server/ansible/roles/pbench-server-vars/tasks/main.yml
+++ b/server/ansible/roles/pbench-server-vars/tasks/main.yml
@@ -185,6 +185,22 @@
     msg: "{{ pbench_host }}"
     verbosity: 1
 
+# pbench server port
+- name: fetch the pbench server API port
+  shell: "pbench-config bind_port pbench-server"
+  register: cmd_output
+  environment:
+    _PBENCH_SERVER_CONFIG: "{{ pbench_server_install_dir }}/lib/config/pbench-server.cfg"
+    PATH: "{{ pbench_server_install_dir }}/bin:/usr/bin"
+    PYTHONPATH: "{{ pbench_server_install_dir }}/lib"
+
+- set_fact:
+    bind_port: "{{ cmd_output.stdout_lines[0].strip() }}"
+
+- debug:
+    msg: "{{ bind_port }}"
+    verbosity: 1
+
 # pbench contact email address
 - name: fetch the contact email address
   shell: "pbench-config admin-email pbench-server"
--- a/server/ansible/roles/pbench-server-activate-httpd-setup-restart/tasks/main.yml
+++ b/server/ansible/roles/pbench-server-activate-httpd-setup-restart/tasks/main.yml
@@ -20,6 +20,23 @@
   notify:
     - restart apache
 
+- name: "Install the pbench server proxy config file"
+  template:
+    src: etc/httpd/conf.d/pbench.conf.j2
+    dest: /etc/httpd/conf.d/pbench.conf
+    owner: root
+    group: root
+    mode: 0644
+  notify:
+    - restart apache
+
+- name: Allow application HTTP connections
+  # command: "setsebool -P httpd_can_network_connect on"
+  seboolean:
+    name: httpd_can_network_connect
+    persistent: yes
+    state: yes
+
 - name: "Make sure rsync is installed. Used below."
   package:
     name: rsync
--- a/server/lib/config/pbench-server-default.cfg
+++ b/server/lib/config/pbench-server-default.cfg
@@ -62,7 +62,7 @@ pbench-receive-dir-prefix = %(pbench-local-dir)s/pbench-move-results-receive/fs-
 pbench-quarantine-dir = %(pbench-local-dir)s/quarantine
 
 # pbench-server rest api variables
-rest_port = 8001
+bind_port = 8001
 rest_version = 1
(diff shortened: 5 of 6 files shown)


lib/pbench/cli/server/shell.py
```diff
@@ -40,7 +39,7 @@ def main():
             "--pid",
             "/run/pbench-server/gunicorn.pid",
             "--bind",
-            f"{host}:{port}",
+            f"localhost:${port}",
```
npalaska, line 42: Does `localhost` works? The bind host used to be 0.0.0.0

dbutenhof: That's the point; gunicorn is only listening on localhost now, relying on Apache to act as gateway from the external network. From the outside, you only see port 80. (Or, when we get to https, 433.) Pbench and gunicorn don't care; it's all localhost to us. Kinda like being inside a container; which, eventually, we probably will be.

portante: Well, this requires us to run Apache locally to reverse proxy it, but we run nginx remotely to reverse proxy services.  So ideally we'd like to be able to specify the bind IP and port # together that should be used by the server.

server/ansible/roles/pbench-server-activate-httpd-setup-restart/tasks/templates/etc/httpd/conf.d/pbench.conf.j2
```diff
@@ -0,0 +1,6 @@
+<VirtualHost *:80>
+    ProxyPreserveHost On
+    ProxyPass /api/ http://127.0.0.1:{{ pbench_server_port }}/api/
```
portante, line 3: Perhaps use `{{ pbench_server_bind_name_or_ip }}` here instead of hard-coding `127.0.0.1`?

server/ansible/roles/pbench-server-vars/tasks/main.yml
```diff
@@ -185,6 +185,22 @@
     msg: "{{ pbench_host }}"
     verbosity: 1
 
+# pbench server port
+- name: fetch the pbench server API port
+  shell: "pbench-config rest_port pbench-server"
```
portante, line 190: Is the port being used by the code besides this ansible role?  If not, seems like this should just be an ansible variable that is required.  If it is going to be used, then it is not clear why we need to fetch this value from the configuration file?  It would have to be set by the administrator deploying that config file anyways, right?

portante (changes requested):

lib/pbench/cli/server/shell.py
dbutenhof: Yes; the root point here was that the main html document loaded for the dashboard can query its own origin on port 80 to hit the configuration API.

But, yes, I hadn't thought about your current remote NGINX proxy configuration, which can be configured to do the same thing... but would definitely require opening the API port.

OK, so I'll revert this change; to work in the local Apache proxy mode, you just set `pbench-server.cfg` to listen on `localhost` instead of hardcoding it here. For a remote proxy, you can still specify the hostname and port and manually open the firewall.

That'll work, too.

I suppose a bigger question is whether it's OK to create the Apache proxy config file in the server httpd activate Ansible role. Under the current production/staging server deployment we're running Apache locally and serving the dashboard even though we're using the remote NGINX proxy; would someone reasonably expect to use the standard Ansible config script here? Maybe... in which case I need to either split it into a separate playbook, or make it conditional on something in the inventory file.

server/ansible/roles/pbench-server-vars/tasks/main.yml
dbutenhof: It's used by the gunicorn launch.

server/ansible/roles/pbench-server-activate-httpd-setup-restart/tasks/templates/etc/httpd/conf.d/pbench.conf.j2
dbutenhof: It wouldn't have mattered here; but since it has to match with the gunicorn launch, if I revert that to the server IP/host I'll have to do it here, too.

dbutenhof:
Note that on my test VM, this resolves to `<hostname>:8001` in the `pbench.conf` file (fqdn for my VM) and `0.0.0.0:8001` in the `shell.py` binding for gunicorn. This is working correctly with no non-default ports open in the firewall... which is actually a bit unexpected. But it also means that the Pbench server responds on an external connection, if we open port 8001, and that should satisfy the needs of an external proxy configuration.

portante (changes requested):
Looks good, can you update the actual commit message of commit itself to have the description that is in the PR?  And also address that description so that it does not say that it requires a local Apache to work?

portante:
@dbutenhof was the commit message updated?  I am not able to see it.  I see the PR description is updated.

portante (changes requested):
Perhaps update the commit message itself?

dbutenhof:
> @dbutenhof was the commit message updated? I am not able to see it. I see the PR description is updated.

Yeah, OK; I only changed what's accessible on GitHub yesterday. But, yeah; the permanent git log is important. See if you like this better.

portante (dismissed):

ndokos (commented):
There is some cognitive dissonance in the `rest_port/pbench_server_port` dichotomy. I wonder if they should be made identical?

dbutenhof:
> There is some cognitive dissonance in the `rest_port/pbench_server_port` dichotomy. I wonder if they should be made identical?

Yes, naming is inconsistent; or maybe, situationally consistent. In the config file, it could be pbench-server / port just as we have elasticsearch / port and graphql / port. In the Ansible situation, I thought it was better to use the `pbench_server` prefix to be reasonably consistent with the other facts. (Although we've got inconsistent there, too with just `pbench_host` and even `mailaddr`, so I suppose I could have called it `rest_port` without being inconsistently inconsistent...)

So what would you like me to do?

portante:
> So what would you like me to do?

Consider changing all this to `bind_host` and `bind_port`.

ndokos:
Maybe what @portante suggested - but maybe just leave it as is in this PR;  open an issue, gather up inconsistenies like this and when we get critical mass, do one PR to fix them.

ndokos (dismissed):

dbutenhof:
> > So what would you like me to do?
> 
> Consider changing all this to `bind_host` and `bind_port`.

I'd really rather not blow up a simple change by sweeping through additional code to make changes like that. Changing `rest_port` to `bind_port` probably isn't that big a deal, though it'll break any 0.71 cfg files that might be deployed for testing (e.g., Nikhil).

In the `pbench.conf.j2` Apache template, it's still going to be `{{ pbench_host }}:{{ bind_port }}` because we're proxying to the `[pbench-server] host` (DNS fqdn or IPv4), not the `bind_host` (all-interface wildcard `0.0.0.0`).

The only real improvement in that would be `bind_host`:`bind_port` in `shell.py`. And, OK, that's moderately nice, so I'm tempted to do it anyway, along with using `bind_port` for the Ansible fact to keep it connected to the config setting.

portante (approved):

ndokos (approved):

PR #2070, opened 2021-01-14
Allow specifying different ports for each client instance by means of

RobertKrawitz:
the --unique-ports option.  It's also possible to specify <client>:<port>
if it's necessary to support specific ports.  This is needed for
conveniently supporting multiple pods with OpenShift.

### DIFF
--- a/agent/bench-scripts/pbench-fio
+++ b/agent/bench-scripts/pbench-fio
@@ -82,6 +82,9 @@ if [ -z "$histogram_interval_sec" ] ;then
 	histogram_interval_sec=10
 fi
 sysinfo="default"
+declare -a client_names=()
+declare -A client_ports=()
+unique_ports=0
 
 function usage {
 	printf "The following options are available:\n"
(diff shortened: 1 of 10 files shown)


dbutenhof (commented):
Nice idea

docs/guides/UserGuide.rst
```diff
@@ -206,6 +206,9 @@ from the fio man page.
 +---------------+-------------------------------------------------------------------------------+
 | --job-file    | if you need to go beyond the recognized options, you can use a fio job file.  |
 +---------------+-------------------------------------------------------------------------------+
+| --unique-ports| use different ports for each client (needed if e. g. multiple clients on one  |
```
portante, line 209: Isn't `e. g.` ususally written without the space?  E.g.: `e.g.`?

portante (changes requested):
Looks good, just wondering how the existing unit tests run.  Try `tox -e bench-scripts`.

portante (approved):
Ship it!

dbutenhof (approved):
One part of me says, wow, there are an awful lot of `for client in "${client_names[@]}" ; do` in here; I wonder if some of those loops could be coalesced.

But maybe not. ;-)

portante (approved):

agent/bench-scripts/pbench-fio
```diff
@@ -603,28 +630,28 @@ function fio_run_job() {
 
 	if [ ! -z "$clients" ]; then
 		debug_log "creating directories on the clients"
-		for client in `echo $clients | sed -e s/,/" "/g`; do
+		for client in "${client_names[@]}"; do
 			ssh $ssh_opts $client mkdir -p $benchmark_results_dir &
 		done
 		wait
 		debug_log "opening port 8765 on firewall on the clients"
```
ndokos, line 637: The debug message needs to be fixed.

agent/bench-scripts/pbench-fio
```diff
@@ -603,28 +630,28 @@ function fio_run_job() {
...
-		for client in `echo $clients | sed -e s/,/" "/g`; do
+		for client in "${client_names[@]}"; do
 			ssh $ssh_opts $client "killall fio >/dev/null 2>&1" &
 		done
 		wait
 		debug_log "starting new fio process on the clients"
-		for client in `echo $clients | sed -e s/,/" "/g`; do
-			ssh $ssh_opts $client "pushd $benchmark_results_dir >/dev/null; screen -dmS fio-server bash -c ''$benchmark_bin' --server 2>&1 >client-result.txt'"
+		for client in "${client_names[@]}"; do
+			ssh $ssh_opts $client "pushd $benchmark_results_dir >/dev/null; screen -dmS fio-server bash -c ''$benchmark_bin' --server${client_ports[$client]:+=,${client_ports[$client]}} 2>&1 >client-result.txt'"
```
ndokos, line 649: IIUC, if the array entry exists and has value say 1234, then this evaluates to `--server=,1234` -  why is the comma present? Should it be there?

ndokos (commented):
If the questions raised are real concerns, then I'll open an issue to address them.

agent/bench-scripts/pbench-fio
portante: According to the [documentation](https://fio.readthedocs.io/en/latest/fio_doc.html#client-server), that `--server=,1234` means, "Start a fio server, listening on all interfaces on port 1234."

RobertKrawitz: Yes, this is correct.

agent/bench-scripts/pbench-fio
RobertKrawitz: Yes; the ports can be different, though, so the question is what the message should be.  One message, of the form "opening fio server port on the clients" or a separate message for each client?

portante: Perhaps something like:

```
if (( unique_ports )); then
    uniq_ports_text=" (unique ports per client)"
else
    uniq_ports_text=""
fi
debug_log "opening ports in firewalls of the clients${uniq_ports_text}"
```

agent/bench-scripts/pbench-fio
ndokos: OK, thanks!

portante:
Adjusted commit on branch `main`: https://github.com/distributed-system-analysis/pbench/commit/5294efeaca43feb6e81788fc4bc6173e9c849705.

PR #2075, opened 2021-01-25
Added source code for live-metric-visualizer and prom-graf-visualizer

Maxusmusti:
Available at:
 - quay.io/pbench/live-metric-visualizer
 - quay.io/pbench/prom-graf-visualizer


agent/containers/images/visualizers/live-metric-visualizer/run.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+python3 prom_ds.py &
```
portante, line 2: Why does this need to be run in the background?

agent/containers/images/visualizers/live-metric-visualizer/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM fedora:latest
```
portante, line 1: This looks quite similar to the `prom-graf-visualizer` container.  Any way we can reduce the overlap?  For instance, perhaps have `prom-graf-visualizer` build its image `FROM live-metric-visualizer`?

agent/containers/images/visualizers/prom-graf-visualizer/run.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+/prometheus-${PROM_VERSION}.linux-amd64/prometheus --config.file=/prometheus.yml --storage.tsdb.path=/data </dev/null &>/dev/null &
```
portante, line 2: Perhaps have just one `run.sh` script which operates differently based on the environment variable settings?

agent/containers/images/visualizers/prom-graf-visualizer/prom_ds.py
```diff
@@ -0,0 +1,65 @@
+import json
```
portante, line 1: Can we merge this `prom_ds.py` with the other one and use switches or environment variables to control the desired behavior?

agent/containers/images/visualizers/prom-graf-visualizer/combo.json
```diff
@@ -0,0 +1,14536 @@
+{"dashboard":
```
portante, line 1: Are these three `prom-graf-visualizer/*.json` files the same as the `live-metric-visualizer/*.json` files?

agent/containers/images/visualizers/live-metric-visualizer/prom_ds.py
```diff
@@ -0,0 +1,92 @@
...
+]
+process = subprocess.run(args, stdout=tokenholder)
+tokenholder.close()
+
+tokenholder = open("key.txt", "r")
+token_raw = tokenholder.readline()
+tokenholder.close()
+token_dict = json.loads(token_raw)
+print(token_dict["key"])
+token = token_dict["key"]
```
dbutenhof, line 35: Doesn't seem there's a need for `process` since you don't use it again.

I'm a little curious, too, why you resorted to `curl` and all that work to JSONify stdout, when you used `requests` below??

portante (changes requested):

agent/containers/images/visualizers/live-metric-visualizer/prom_ds.py
```diff
@@ -0,0 +1,92 @@
...
+
+headers = {"Content-Type": "application/json", "Authorization": f"Bearer {token}"}
+
+prom_source_url = f"{graf_base}api/datasources"
+message = {
```
dbutenhof, line 40: A bit odd to name the dict representing a `POST` payload as "message"??

agent/containers/images/visualizers/live-metric-visualizer/prom_ds.py
```diff
@@ -0,0 +1,92 @@
...
+    "url": "http://localhost:9090",
+    "access": "proxy",
+    "basicAuth": False,
+}
+response = requests.post(prom_source_url, headers=headers, json=message)
+
+if response.status_code == 200:
+    print(json.loads(response.content.decode("utf-8")))
+else:
+    print(json.loads(response.content.decode("utf-8")))
```
dbutenhof, line 52: The `if` isn't useful as you're repeating exactly the same `print` for either a success response or an error document.

agent/containers/images/visualizers/live-metric-visualizer/prom_ds.py
```diff
@@ -0,0 +1,92 @@
+import json
+import requests
+import subprocess
+import time
+
+graf_base = "http://localhost:3000/"
```
dbutenhof, line 6: I'd like to see a description of why you created this and what it does so everyone doesn't necessarily have to read the entire program to find out.

The name `prom_ds.py` also doesn't help much; a name more like `configure_grafana.py` would at least give a potential reader an idea of what they're going to see.

agent/containers/images/visualizers/prom-graf-visualizer/combo.json
dbutenhof: I was thinking the same thing. They have exactly the same sizes, so I think so.

Can the be factored down a level so they're not duplicated?

agent/containers/images/visualizers/prom-graf-visualizer/prom_ds.py
```diff
@@ -0,0 +1,65 @@
...
+import requests
+import subprocess
+import time
+
+graf_base = "http://localhost:3000/"
```
dbutenhof, line 6: Again, an inline description of function and purpose, combined with a more specific name, would help this file a lot. It does at least seem to be distinct from its counterpart.

agent/containers/images/visualizers/live-metric-visualizer/run.sh
dbutenhof: That I thought was somewhere between "weird" and "clever" ... maybe both; it just drops the configuration command into the background, and then takes over the parent process with the server that the configuration command is looking for. So once its started, the background command does its thing and goes away.

Again, I think some descriptive comments would help.

Maxusmusti:
Fixed everything except having one image build off of the other. Wanted to first ask if there was any real benefit to keeping them separate vs building one off the other (like being able to use one run command for installing grafana and prometheus for fewer layers / smaller image size) or if having one build off the other would be more efficient.

dbutenhof (commented):
> Fixed everything except having one image build off of the other. Wanted to first ask if there was any real benefit to keeping them separate vs building one off the other (like being able to use one run command for installing grafana and prometheus for fewer layers / smaller image size) or if having one build off the other would be more efficient.

Without analyzing every line, it does seem that the "prometheus" container is basically the "grafana" container plus prometheus, which suggests that layering would be an improvement at least for maintenance going forward. On the other hand, if the prometheus bits aren't large, maybe just doing one `Dockerfile` makes sense. (If the overhead is mostly runtime when building the container, since you're generally going to be building both at the same time anyway it might not matter.)

agent/containers/images/visualizers/grafana_setup.py
```diff
@@ -0,0 +1,103 @@
...
+]
+subprocess.run(args, stdout=tokenholder)
+tokenholder.close()
+
+tokenholder = open("key.txt", "r")
+token_raw = tokenholder.readline()
+tokenholder.close()
+token_dict = json.loads(token_raw)
+print(token_dict["key"])
+token = token_dict["key"]
```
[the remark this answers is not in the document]
portante: @Maxusmusti, there is also a "pycurl" module if you'd like to use that.

portante (changes requested):
If you were to do a side-by-side comparison of the two Dockerfiles, could you find a common subset of the two, make that one Dockerfile, and the have the second be what is left over?

agent/containers/images/visualizers/run.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# Run collector (if needed), grafana setup, and grafana in parallel as grafana/collector are continuous processes
+# and the grafana setup needs to be run while grafana is up
+${COLLECTOR} &
+python3 grafana_setup.py &
+exec grafana-server                                         \
```
portante, line 7: Seems like we could fold this line into `grafana_setup.py`, along with the collector, and call it `run.py` instead.

We'd just have to `subprocess.run()` the `grafana-server` command first from the new `run.py` in the background, and then continue as normal.  We could also invoke the collector there too.

This would really tighten up this code.

portante (changes requested):
Looking really good, just a bit more python to tighten this up and we're good.

agent/containers/images/visualizers/DockerfilePostProm
```diff
@@ -0,0 +1,18 @@
+# To be ran from parent 'visualize' directory: podman build -t <name> -f DockerfilePostProm .
```
dbutenhof, line 1: "To be run" ... although in this context, something like "Use this with the following command" might be more clear. Also, I assume you mean "'visualizers' directory" ... and "parent" seems superfluous and possibly confusing since the file is _in_ that directory.

agent/containers/images/visualizers/README.md
```diff
@@ -0,0 +1,28 @@
...
+- Build locally with podman build -t <name> -f DockerfileLive .
+- Available at quay.io/pbench/live-metric-visualizer
+- A tool available to visualize Prometheus and PCP data live during a Pbench benchmark run
+
+- Directions:
+  -   `podman pull pbench/live-metric-visualizer`
+  -   `podman run --network host pbench/live-metric-visualizer`
+
+- NOTES:
+  - Only the grafana server with preconfigured data sources for prometheus and PCP, as well as preloaded dashboards
```
dbutenhof, line 11: When you start with "Only" and then list several things and continue with "as well as", the "Only" becomes a bit of a misdirection. A positive and direct statement would be better.

"The container includes a Grafana server with preconfigured data sources and dashboards for Prometheus and PCP.", for example.

agent/containers/images/visualizers/grafana_setup.py
```diff
@@ -0,0 +1,99 @@
...
+Chooses what to upload/enable based off VIS_TYPE environment variable.
+
+If VIS_TYPE is 'live' (default for live-metric-visualizer):
+    - Node exporter/DCGM dashboards will be uploaded
+    - Prometheus data source will be configured
+    - Grafana-pcp plugin will be enabled
+    - PCP redis and vector datasources will be configured
+    - PCP default dashboards will be enabled
+If VIS_TYPE is 'prom' (default for prom-graf-visualizer):
+    - Node exporter/DCGM dashboards will be uploaded
```
dbutenhof, line 12: You say "will be uploaded", while in the other two lists "PCP default dashboards" are "enabled". I'm assuming this means that the PCP dashboards are indeed present but not enabled by default whereas node exporter and dcgm have to be added; however, (unless that doesn't actually "enable" them) it might be better to say "uploaded and enabled" so there's no doubt.

agent/containers/images/visualizers/DockerfileLive
```diff
@@ -0,0 +1,37 @@
+# To be ran from parent 'visualize' directory: podman build -t <name> -f live-metric-visualizer/Dockerfile .
```
dbutenhof, line 1: "To be run" ... although in this context, something like "Use this with the following command" might be more clear. Also, I assume you mean "'visualizers' directory" ... and "parent" seems superfluous and possibly confusing since the file is _in_ that directory.

portante (approved):

agent/containers/images/visualizers/run.py
```diff
@@ -0,0 +1,128 @@
...
+
+import os
+import json
+import requests
+import time
+import subprocess
+
+# Start the collector (if needed)
+collector_cmd = os.environ["COLLECTOR"]
+if not collector_cmd == ":":
```
dbutenhof, line 30: The `":"` seems like a bit of a "magic number" that's not really necessary here (you could have omitted the environment variable or set it to `""`), and a bit of documentation at least in the README might be in order here, but I won't worry about it too much. Maybe consider cleaning this up when you add the PCP visualizer path?

dbutenhof (approved):

agent/containers/images/visualizers/run.py
portante: Filed https://github.com/distributed-system-analysis/pbench/issues/2078 to track.

portante:
Adjusted commit on `main` at: https://github.com/distributed-system-analysis/pbench/commit/d4390906f48964ef64c0b44f5f9ad16ae493b6a5.

PR #2076, opened 2021-01-26
Containerization of TM and TDS

portante:
Add support for running the Tool Meister(s) and the Tool Data Sink as containerized services.

We provide separate images for Tool Meister and Tool Data Sink containers. Each container has a entry point command at `tool-meister/tool-*-ep`, which reads environment variables provided by the container specification and passes them as parameters to the respective Tool Meister and Tool Data Sink commands.

We have modified the Tool Meister and Tool Data Sink code to continuously try to connect to the specified Redis server to find their parameter key.

We have changed the Tool Data Sink to only daemonize itself when requested.

The Tool Data Sink was changed to recognize the difference between the internal (to the container) and external namespaces.  The driver must use a `.path` file in the `${pbench_run}` directory when working with the Tool Data Sink so that they can pass external references to the Tool Data Sink container via the APIs, allowing it to map them to internal references in the container.

We have endowed the `pbench-tool-meister-start` and `-stop` commands with an additional parameter, `--redis-server=<host>:<port>`, which allows the user to specify their own Redis server. This implies that the user will also be handling the orchestration of the Tool Data Sink and Tool Meister containers. Besides the `--redis-server` parameter, we also accept an environment variable, `PBENCH_REDIS_SERVER=<host>:<port>`.

The `pbench-tool-meister-start` command will now timeout waiting for the TDS to show up. We now wait a maximum time of one minute for the Tool Data Sink to show up ready for service.  Prior to this change the start-up process would wait indefinitely.  Since we added a new `ReturnCode` for this timeout case, we also fix up the other uses of `ReturnCode` in the Tool Group loading failure case.

Prior to the creation of the Tool Meister sub-system, collection of the configuration (or system) information was performed on a best-effort basis.  A benchmark script would not report "failure" if the `pbench-collect-sysinfo` command failed.  This commit changes both `pbench-tool-meister-start` and `-stop` to behave as before.  The code for initializing and ending the persistent tools has been moved out from underneath a conditional to be clear that its operation is not optional.

In addition to the above major changes we also:

 * In `pbench-tool-meister-stop` we have moved the steps taken to gracefully shutdown the local Tool Data Sink, Tool Meister, and Redis server to its own method to aid readability of the code

 * We use the `error_log` and `warn_log` methods were appropriate to restore / add output to the `pbench.log` file

 * Correct how Tool Data Sink handles `EADDRINUSE`

   We now capture exceptions raise in the WSGI thread so that we can report the success or failure of that thread starting up.  This also removes the hacky timed wait that was there.

   While we are at, we also remove the hacky timed wait for the log capture thread.

 * Only run WSGI server if successfully created

 * Add unit tests for new BenchmarkRunDir class, and the DataSinkWsgiServer class

 * Fix how OSError is forwarded from WSGI thread

 * Rename constants `*_port` to `def_*_port`

 * Move the `ToolGroup` class to its own module

   This allows reuse by the other commands as well, allowing us to DRY out that code a bit as well.

----

There are four (5) commits to this PR that we'll maintain when we merge:

 1. Correction to the agent image building `Makefile` to work as advertised in the `README.md`
 2. The combined (from 20+ separate commits) set of changes as described above
 3. Use of PCP "bintray" repos when building the container images
 4. Properly exposing PCP ports and corrections to the `firewalld` service files
 5. Incorporation of @Maxusmusti work from PR https://github.com/portante/pbench/pull/15


portante:
RPMs for these latest changes at https://copr.fedorainfracloud.org/coprs/portante/pbench-test/, `v0.71.0-126ga8640e45b`.

portante:
RPMs for these latest changes at https://copr.fedorainfracloud.org/coprs/portante/pbench-test/, `v0.71.0-130ga9672a901`.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -550,6 +566,116 @@ def terminate(self):
...
+        # match.
+        #
+        # E.g.:
+        #  $ benchmark_run_dir = /home/<USER>/run-dir
+        #  $ cat /home/<USER>/run-dir/.path
+        #  /home/<USER>/run-dir
+        #  $ podman run --volume /home/<USER>/run-dir:/var/lib/pbench-agent\
+        #    pbench-agent-tool-data-sink bash
+        #  [ abcdefg /]$ cat /var/lib/pbench-agent/.path
+        #  /home/<USER>/run-dir
```
portante, line 657: This example is wrong, this should be:

```
        # E.g.:
        #  $ pbench_run="/home/<USER>/run-dir"
        #  $ benchmark_run_dir="${pbench_run}/script_config_<date>"
        #  $ cat ${pbench_run}/.path
        #  /home/<USER>/run-dir
        #  $ podman run --volume /home/<USER>/run-dir:/var/lib/pbench-agent\
        #    pbench-agent-tool-data-sink bash
        #  [ abcdefg /]$ cat /var/lib/pbench-agent/.path
        #  /home/<USER>/run-dir
```

lib/pbench/agent/tool_data_sink.py
```diff
@@ -550,6 +566,116 @@ def terminate(self):
...
+        try:
+            benchmark_run_dir_lcl = self._ext_benchmark_run_dir.resolve(strict=True)
+        except Exception:
+            # Might be in a container; let's first construct the
+            # internal-to-the-container benchmark run directory.
+            benchmark_run_dir_lcl = (
+                self._int_pbench_run / self._ext_benchmark_run_dir.name
+            )
+            try:
+                dot_path = (self._int_pbench_run / ".path").read_text().strip()
```
portante, line 667: `s/dot_path/dot_path_contents/`

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -27,23 +76,34 @@ import socket
...
 from distutils.spawn import find_executable
 from pathlib import Path
 
 import redis
 
-import pbench.agent.toolmetadata as toolmetadata
-
+from pbench.agent import PbenchAgentConfig
+from pbench.agent.constants import (
+    redis_port as def_redis_port,
```
portante, line 87: Consider changing `redis_port` in the constants module to `def_redis_port`.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -27,23 +76,34 @@ import socket
...
+    cli_tm_channel_prefix,
+    tm_channel_suffix_to_client,
+    tm_channel_suffix_from_client,
+    tm_channel_suffix_to_logging,
+)
+from pbench.agent.redis import RedisChannelSubscriber
 from pbench.agent.tool_data_sink import main as tds_main
 from pbench.agent.tool_meister import main as tm_main
```
portante, line 95: Consider renaming `main` in both of the above modules to something "else".

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -55,26 +115,33 @@ pidfile {tm_dir}/redis_{redis_port:d}.pid
...
 """
 
-# FIXME: this should be moved to a shared area
-channel = "tool-meister-chan"
-
-# Maximum time to wait for the Redis server to respond.
-REDIS_MAX_WAIT = 60
-
 
 class ToolGroup:
```
portante, line 119: Consider moving to its own module.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -151,60 +219,98 @@ class ToolGroup:
                 tools[tool] = host_opts
         return tools
 
+    def get_label(self, host):
+        """get_label - given a target host, return the label associated with
+        that host.
+        """
+        try:
+            label = self.labels[host]
```
portante, line 227: How about:

```
return self.labels.get(host, "")
```

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -527,150 +616,333 @@ def main(argv):
...
+        default=None,
+        help="The list of system information items to be collected.",
+    )
+    parser.add_argument(
+        "--redis-server",
+        dest="redis_server",
+        default=os.environ.get("PBENCH_REDIS_SERVER", None),
+        help=(
+            "Use an existing Redis server specified by <hostname>:<port>;"
+            " implies an existing Tool Data Sink and Tool Meisters as well."
```
portante, line 940: Add help text that specifies that if a `--redis-server` is specified, it is assumed the caller has created and provisioned that Redis server, and the Tool Data Sink and Tool Meisters as well.

portante: Add a message about the environment variable as well.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -272,44 +384,102 @@ def main(argv):
...
+    # Verify all the command line arguments
+    group = cli_params.tool_group
     try:
+        # Load the tool group data
         tool_group = ToolGroup(group)
-    except Exception:
-        logger.exception("failed to load tool group data")
+    except BadToolGroup as exc:
+        logger.error(str(exc))
         return 1
```
portante, line 398: Use `ReturnCode.something`?

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -272,44 +384,102 @@ def main(argv):
...
+    except Exception:
+        logger.exception("failed to load tool group data for '%s'", group)
+        return ReturnCode.BADTOOLGROUP
+    else:
+        if not tool_group.hostnames:
+            # If a tool group has no tools registered, then there will be no
+            # host names on which to start Tool Meisters.
```
portante, line 405: Consider adding a warning or at least some message to say no tools are registered on hosts, so no Tool Meisters will be started.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -272,44 +384,102 @@ def main(argv):
...
-    # See if anybody told us to use certain options with SSH commands.
-    ssh_opts = os.environ.get("ssh_opts", "")
+    sysinfo, bad_l = cli_verify_sysinfo(cli_params.sysinfo)
+    if bad_l:
+        logger.error('invalid sysinfo option(s), "{}"', ",".join(bad_l))
 
+    # Load the tool metadata
     try:
-        benchmark_run_dir = os.environ["benchmark_run_dir"]
+        inst_dir = PbenchAgentConfig(
```
portante, line 414: Why do we need this?  Can't we use the `pbench_bin` environment variable?

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -364,160 +534,79 @@ def main(argv):
...
-        "<host replace me>",
-        f"{tool_meister_cmd}-remote",
-        tm_bind_hostname,
-        str(redis_port),
-        "<tm param key>",
-        "yes",
-    ]
-    ssh_pids = []
+    # +
+    # 3. Push the loaded tool group data and metadata into the Redis server
```
portante, line 606: This should be some kind of API offered in the Python client module.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -527,150 +616,333 @@ def main(argv):
...
+        except Exception:
+            logger.exception("failed to create pbench data sink, daemonized")
+            return kill_redis_server(redis_pid, ReturnCode.TDSFORKFAILED)
         else:
-            exit_status = os.WEXITSTATUS(_exit_status)
-            if pid != exit_pid:
+            # Wait for logging channel to be up and ready before we start the
+            # local and remote Tool Meisters.
+            num_present = 0
+            while num_present == 0:
```
portante, line 708: This should be done with a timeout.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -527,150 +616,333 @@ def main(argv):
...
-        else:
-            logger.debug("publish() = %r", ret)
-        ret_val = kill_redis_server(redis_pid)
-    elif successes > 0:
-        # If any successes, then we need to wait for them to show up as
-        # subscribers.
-        logger.debug(
-            "e. waiting for all successfully spawned SSH processes"
-            " to show up as subscribers"
+            return kill_redis_server(redis_pid, ReturnCode.TMNOSUCCESSES)
```
portante, line 850: Consider having one place to kill the Redis server and then return the intended code perhaps stored in the exception object.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -527,150 +616,333 @@ def main(argv):
...
+    # +
+    # 6. Wait for the TDS to send a message reporting that it, and all the
+    # TMs, started.
+    # -
+
+    # If any successes, then we need to wait for them to show up as
+    # subscribers.
+    logger.debug(
+        "6. waiting for all successfully created Tool Meister processes"
+        " to show up as subscribers"
```
portante, line 866: We should make this part of the API for a "start" operation.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -527,150 +616,333 @@ def main(argv):
...
+        if sysinfo:
+            sysinfo_path = benchmark_run_dir / "sysinfo" / "beg"
+            try:
+                sysinfo_path.mkdir(parents=True)
+            except Exception:
+                logger.error(
+                    "Unable to create sysinfo-dump directory base path: {}",
+                    sysinfo_path,
+                )
+                ret_val = ReturnCode.EXCSYSINFODIR
```
portante, line 894: Why not just `return` directly here?

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -527,150 +616,333 @@ def main(argv):
...
+            else:
+                logger.debug("7. Collecting system information")
+                info_log("Collecting system information")
+                ret_val = client.publish(group, sysinfo_path, "sysinfo", sysinfo)
+                ret_val = (
+                    ReturnCode.SUCCESS if ret_val == 0 else ReturnCode.SYSINFOFAILED
+                )
```
portante, line 901: Why not return directly here as well?

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -527,150 +616,333 @@ def main(argv):
...
+
+        if ret_val == ReturnCode.SUCCESS:
```
portante, line 903: Well, if we decide to return directly above (which we arguably should), why do we need this test?

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -527,150 +616,333 @@ def main(argv):
...
+            tool_dir = benchmark_run_dir / f"tools-{group}"
+            try:
+                tool_dir.mkdir(exist_ok=True)
+            except Exception as exc:
+                logger.error(
+                    'failed to create tool output directory, "{}": {}', tool_dir, exc
+                )
+                ret_val = ReturnCode.EXCTOOLGROUPDIR
```
portante, line 911: Why not return directly here?

agent/util-scripts/pbench-tool-meister-start
portante: This seems like a good use of "import as" functionality.  The "main" function of each of those modules on their own doesn't conflict.  The modules are named differently, so we could just `import pbench.agent.tool_data_sink`, and then then reference `pbench.agent.tool_data_sink.main` instead.  But that is long.  So these two uses of "as" are nice and convenient for where the code is used.

agent/util-scripts/pbench-tool-meister-start
portante: Done.

agent/util-scripts/pbench-tool-meister-start
portante: Done.

agent/util-scripts/pbench-tool-meister-start
portante: Done.

agent/util-scripts/pbench-tool-meister-start
portante: Done.

agent/util-scripts/pbench-tool-meister-start
portante: Done.

agent/util-scripts/pbench-tool-meister-start
portante: Fixed.

agent/util-scripts/pbench-tool-meister-start
portante: Fixed.

agent/util-scripts/pbench-tool-meister-start
portante: This should not feed into the final status of the command, as collecting system information was allowed to fail without affecting the benchmark test run.

agent/util-scripts/pbench-tool-meister-start
portante: Fixed.

lib/pbench/test/unit/agent/test_tool_data_sink.py
```diff
@@ -284,13 +288,194 @@ def mock_make_server(host, port, app, *args, **kwargs):
...
+        threads that we need to test.  We list them here using "MainThr" as the
+        name of the "main thread" which _creates_ the WSGI thread, and "WsgiThr"
+        as the name of the created WSGI thread invoking the "run" method.
+
+        References:
+            .wait() called at
+                line 184 (in .stop() method)
+                line 833 (in __enter__ method)
+            .stop() called at
+                line 886 (in __exit__ method)
```
dbutenhof, line 310: I don't think including line numbers here is a good idea; they'll change, and get out of sync, and these notes will become more confusing than helpful. How many times does `stop()` or `__enter_()` call `wait()` anyway? If there are multiple calls, or if you really feel the reader needs a specific reference, calling out line comments like `# Unit test reference 1` might be a more stable way to accomplish that.

lib/pbench/test/unit/agent/test_tool_data_sink.py
```diff
@@ -284,13 +288,194 @@ def mock_make_server(host, port, app, *args, **kwargs):
...
+            wsgi_server.stop()
+
+        hostnames = dict(
+            A="host.example.com",
+            B="host.example.com",
+            C="oserror.example.com",
+            D="oserror.example.com",
+            E="exception.example.com",
+            F="exception.example.com",
+        )
```
dbutenhof, line 386: It's kinda cute that you're using the hostname to send messages to the WSGI server, but I think a comment here documenting how the messages are encoded ("starts with `exception`") and what they mean ("raise at this operation") would eliminate some potential future confusion.

dbutenhof (commented):
I skimmed through most of the "relatively recent" commits to catch up on what's been going on. Just a few minor comments along the way...

lib/pbench/test/unit/agent/test_tool_data_sink.py
webbnh: I had no idea that `example.com` is a [reserved domain](https://www.iana.org/domains/reserved)!  :-)

(Top-level domains `.test` and `.invalid` are [also available](https://tools.ietf.org/html/rfc2606) for these uses.)

Maxusmusti (changes requested):
Everything worked beautifully for `TAG="051e4b8f3"`, including visualization of live metrics. Everything started, stopped, connected, etc. as expected. 

Unfortunately, issues arose when using `TAG="alpha"`. The tool meister would crash upon receiving the param-key (I got it once to say it was because no mapping existed for various tools, but on every other run it crashed before any logs were viewable). When using an older working tm, the tds would crash after. I'm not sure what changed between the two image builds, since there were no changes to the repo, but something broke in the changes made for the "alpha" images.

portante:
> Everything worked beautifully for `TAG="051e4b8f3"`, including visualization of live metrics. Everything started, stopped, connected, etc. as expected.

Great!

> Unfortunately, issues arose when using `TAG="alpha"`. The tool meister would crash upon receiving the param-key (I got it once to say it was because no mapping existed for various tools, but on every other run it crashed before any logs were viewable). When using an older working tm, the tds would crash after. I'm not sure what changed between the two image builds, since there were no changes to the repo, but something broke in the changes made for the "alpha" images.

Thankfully that is image is just old and out of date and we can re-tag `051e4b8f3` as `alpha`.

lib/pbench/test/unit/agent/test_tool_data_sink.py
portante: Fixed.

lib/pbench/test/unit/agent/test_tool_data_sink.py
portante: Fixed.

portante:
This code is ready for a final review.  All comments have been addressed that I can see.

lib/pbench/agent/tool_meister.py
```diff
@@ -999,13 +1003,17 @@ def start_tools(self, data):
 
         # Name of the temporary tool data directory to use when invoking
         # tools.  This is a local temporary directory when the Tool Meister is
-        # remote from the pbench controller.
-        if self._controller == self._hostname:
+        # remote from the pbench controller.  When the Tool Meister is run in
+        # a container the "directory" parameter will not map into its
+        # namespace, so we always considered containerized Tool Meisters as
```
portante, line 1008: `s/considered/consider/`

lib/pbench/agent/tool_meister.py
```diff
@@ -1841,7 +1854,7 @@ def main(argv):
             f"{PROG}: Unable to fetch and decode parameter key, '{param_key}': {exc}",
             file=sys.stderr,
         )
-        return 7
+        return 6
```
portante, line 1857: Need a `ReturnCode` class.

dbutenhof (approved):

portante:
@Maxusmusti, I see that you have a "change requested" against this PR, but I am not seeing what needs to be changed.  Can you let me know what needs to be fixed?  Thanks!

Maxusmusti:
> @Maxusmusti, I see that you have a "change requested" against this PR, but I am not seeing what needs to be changed. Can you let me know what needs to be fixed? Thanks!

That was an old one, after a rebase I will retest the latest image builds + a traditional run from this branch and approve

webbnh:
> All comments have been addressed that I can see.

I see 5 unresolved "conversations" started by @portante (on 15 February)....

portante:
> > All comments have been addressed that I can see.
> 
> I see 5 unresolved "conversations" started by @portante (on 15 February)....

Let the other unresolved conversations in PR #2005, I was thinking that the 5 unresolved conversations here represent potential future work.

portante:
Rebased to avoid merge conflicts, squashed review comments commits.

webbnh:
> I was thinking that the 5 unresolved conversations here represent potential future work.

OK.  Will anyone come back here and see them, or should someone open issues for them?

dbutenhof (approved):
I apparently forgot to mark a few as Viewed, or you actually changed them. A quick scan didn't show anything obviously wrong.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -646,186 +552,202 @@ def main(_prog, cli_params):
...
+                    return kill_redis_server(redis_pid, ReturnCode.TDSLOGPUBFAILED)
+                else:
+                    if num_present == 0:
+                        if time.time() > timeout:
+                            logger.error(
+                                "The Tool Data Sink failed to start within one minute"
+                            )
+                            return kill_redis_server(
+                                redis_pid, ReturnCode.TDSSTARTUPTIMEOUT
+                            )
```
webbnh, line 614: If this triggers, then we know that the TDS got far enough to daemonize itself but that it died or got stuck before it could report in.  Shouldn't we at least look and see if it created a PID file that we can use to kill it if it's stuck?  (This seems like a piece of utility code which `pbench-tool-meister-stop` could share.)

agent/util-scripts/pbench-tool-meister-stop
```diff
@@ -42,6 +38,88 @@ def is_running(pid):
...
+        tds_pid_file = benchmark_run_dir / "tm" / "pbench-tool-data-sink.pid"
+        try:
+            pid_str = tds_pid_file.read_text()
+        except OSError as exc:
+            if exc.errno != errno.ENOENT:
+                raise
+        else:
+            tds_pid = int(pid_str)
+            logger.debug("waiting for tool-data-sink (%d) to exit", tds_pid)
+            wait_for_pid(tds_pid)
```
webbnh, line 68: Should there be a timeout here (after which we "help it along", with a `kill(SIGKILL)` or similar)?

agent/util-scripts/pbench-tool-meister-stop
```diff
@@ -42,6 +38,88 @@ def is_running(pid):
...
+        except OSError as exc:
+            if exc.errno != errno.ENOENT:
+                raise
+        else:
+            ltm_pid = int(pid_str)
+            logger.debug("waiting for local tool-meister (%d) to exit", ltm_pid)
+            wait_for_pid(ltm_pid)
+    except Exception:
+        logger.exception("Exception encountered waiting for local tool-meister")
+        ret_val = 1
```
webbnh, line 88: This looks ripe for refactoring, given how similar it is to the previous and subsequent blocks.

agent/util-scripts/pbench-tool-meister-stop
```diff
@@ -42,6 +38,88 @@ def is_running(pid):
...
+                else:
+                    if time.time() > timeout:
+                        try:
+                            os.kill(redis_server_pid, signal.SIGKILL)
+                        except ProcessLookupError:
+                            pid_exists = False
+                        except Exception:
+                            raise
+                    else:
+                        time.sleep(0.1)
```
webbnh, line 115: I think we don't want this `else` here:  I think we want to do the sleep whether we send the `SIGKILL` or not.  It may take some time for the `SIGKILL` to have the desired effect, and we don't want to immediately return to the top of the loop and send the `SIGTERM` again (_and_ we probably would rather not exit the loop until the target is actually dead).  So, if we execute the sleep either way, I think that that will do exactly the right thing.

agent/util-scripts/pbench-tool-meister-stop
```diff
@@ -103,20 +181,38 @@ def main(_prog, cli_params):
...
+        if len(parts) != 2:
+            logger.error("Bad Redis server specified, '%s'", cli_params.redis_server)
+            return 1
+        try:
+            redis_port = int(parts[1])
+        except Exception:
+            logger.error("Bad Redis port specified, '%s'", cli_params.redis_server)
+            return 1
+        else:
+            redis_host = parts[0]
```
webbnh, line 211: Do we want to support ":55555" as a shorthand for "localhost:55555"?

webbnh: Y'know, this code should be refactored so that we get the same support for `-start`, `-stop`, and `-client`....

agent/util-scripts/pbench-tool-meister-stop
```diff
@@ -103,20 +181,38 @@ def main(_prog, cli_params):
         full_hostname = os.environ["_pbench_full_hostname"]
         benchmark_run_dir = Path(os.environ["benchmark_run_dir"]).resolve(strict=True)
     except Exception:
-        logger.exception("failed to fetch parameters from the environment")
+        logger.exception("failed to fetch required parameters from the environment")
         return 1
```
webbnh, line 185: `-start` returns a variety of error codes...why doesn't `-stop` return the same ones for the same cases?  Moreover, there are two distinct flavors of "failure":  the kind where the tool couldn't even try to do its job (bad tool group, bad command arguments, et al.) and the kind where it tried but something prevented it from working.  It's worth having the status return distinguish at least those two cases.

webbnh (commented):
Mostly nits with a few questions and a couple of suggestions.

More to come (I'm _almost_ done).

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1683,89 +2059,44 @@ def main(argv):
...
+            pbench_bin,
+            pbench_run,
+            hostname,
+            tar_path,
+            cp_path,
+            param_key,
+            params,
+            optional_md,
+        )
+    return ret_val
```
ndokos, line 2102: You could avoid duplicating the arg list with something like this:

```
if daemonize == "yes:
   func = daemon
else:
   func = driver

return func(
      PROG,
       ...)
```

ndokos (changes requested):
Some typos and some suggestions for refactoring (which can be ignored).

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1609,71 +1831,225 @@ def put_document(self, data_ctx, hostname):
...
+        print("External 'cp' executable not found", file=sys.stderr)
         return 2
 
+    try:
+        pbench_run = os.environ["pbench_run"]
+    except KeyError:
+        print(
+            "Unable to fetch pbench_run environment variable", file=sys.stderr,
+        )
+        return 3
```
webbnh, line 2030: I think we need to set up a 12-step program to get us off of using exceptions.  How about this, instead?
```
    if "pbench_run" not in os.environ:
        print(
            "Unable to fetch pbench_run environment variable", file=sys.stderr,
        )
        return 3
    pbench_run = os.environ["pbench_run"]
```
or, perhaps better,
```
    pbench_run = os.environ.get("pbench_run")
    if pbench_run is None:
        print(
            "Unable to fetch pbench_run environment variable", file=sys.stderr,
        )
        return 3
```

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1609,71 +1831,225 @@ def put_document(self, data_ctx, hostname):
...
+        return 5
 
     try:
-        params_raw = redis_server.get(param_key)
-        if params_raw is None:
-            logger.error('Parameter key, "%s" does not exist.', param_key)
-            return 5
-        logger.debug("params_key (%s): %r", param_key, params_raw)
-        params_str = params_raw.decode("utf-8")
+        # Loop waiting for the key to show up.
```
webbnh, line 2051: "Wait for the parameter message to show up."  (We have the key, and there's no loop here -- _how_ we wait is an implementation detail of the function that we're about to call....)

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1683,89 +2059,44 @@ def main(argv):
...
         #                  "benchmark_run_dir": "/loo/goo" }'
         params = json.loads(params_str)
-        channel_prefix = params["channel_prefix"]
-        benchmark_run_dir = Path(params["benchmark_run_dir"]).resolve(strict=True)
-        bind_hostname = params["bind_hostname"]
-        tool_group = params["group"]
-        tool_trigger = params["tool_trigger"]
-        tools = params["tools"]
-        tool_metadata = ToolMetadata.tool_md_from_dict(params["tool_metadata"])
+        ToolDataSink.fetch_params(params, pbench_run)
```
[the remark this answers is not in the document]
webbnh: If what you're trying to do is to validate the "benchmark_run_dir" (or the tool metadata), why not do that explicitly?

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1683,89 +2059,44 @@ def main(argv):
...
+            redis_port,
+            pbench_bin,
+            pbench_run,
+            hostname,
+            tar_path,
+            cp_path,
+            param_key,
+            params,
+            optional_md,
+        )
```
webbnh, line 2101: I'm not exactly sure why, but these giant parameter lists really bug me, and having them appear twice in a row makes it worse.  How about this, which gets rid of one of them?

```
    start_func = daemon if daemonize else driver
    ret_val = start_func(
            PROG,
            redis_server,
            redis_host,
            redis_port,
            pbench_bin,
            pbench_run,
            hostname,
            tar_path,
            cp_path,
            param_key,
            params,
            optional_md,
        )
```

webbnh (commented):
I still have two more files to go....

Here's another pile of nits, pointed questions, and suggestions.

agent/util-scripts/pbench-tool-meister-stop
```diff
@@ -42,6 +38,88 @@ def is_running(pid):
...
+            if exc.errno != errno.ENOENT:
+                raise
+        else:
+            ltm_pid = int(pid_str)
+            logger.debug("waiting for local tool-meister (%d) to exit", ltm_pid)
+            wait_for_pid(ltm_pid)
+    except Exception:
+        logger.exception("Exception encountered waiting for local tool-meister")
+        ret_val = 1
+
```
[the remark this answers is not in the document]
webbnh: I think the key is the "waiting for local tool-meister", part.

lib/pbench/agent/tool_group.py
```diff
@@ -0,0 +1,158 @@
...
+        self.group = group
+
+        # __trigger__
+        try:
+            _trigger = (self.tg_dir / "__trigger__").read_text()
+        except OSError as ex:
+            if ex.errno != errno.ENOENT:
+                raise
+            # Ignore missing trigger file
+            self.trigger = None
```
webbnh, line 85: Could this be better expressed as the following?

```
        except FileNotFoundError:
            # Ignore missing trigger file
            self.trigger = None
```
Other `OSError`'s will simply propagate past.

lib/pbench/agent/tool_group.py
```diff
@@ -0,0 +1,158 @@
...
+        # names and parameters for each tool
+        self.hostnames = {}
+        self.labels = {}
+        for hdirent in os.listdir(self.tg_dir):
+            if hdirent == "__trigger__":
+                # Ignore handled above
+                continue
+            if not (self.tg_dir / hdirent).is_dir():
+                # Ignore wayward non-directory files
+                continue
```
webbnh, line 106: If `__trigger__` is a non-directory file, we don't need to handle it at line 101 because it would be skipped at line 104.

lib/pbench/agent/tool_group.py
```diff
@@ -0,0 +1,158 @@
...
+            # We assume this directory is a hostname.
+            host = hdirent
```
webbnh, line 108: Why bother with two variables pointing at the host/host directory entry?  (It seems like it creates an opportunity for confusion or error without much benefit.)

lib/pbench/agent/tool_group.py
```diff
@@ -0,0 +1,158 @@
...
+            if host not in self.hostnames:
+                self.hostnames[host] = {}
```
webbnh, line 110: It seems like `host` would never be in `self.hostnames`...what am I missing?

Also, is this attribute ever used?

lib/pbench/agent/tool_group.py
webbnh: Ditto for `tool` at line 122.

lib/pbench/agent/tool_group.py
```diff
@@ -0,0 +1,158 @@
...
+                    (self.tg_dir / host / tool).read_text().split("\n")
+                )
+                tool_opts_lines = []
+                for line_raw in tool_opts_raw_lines:
+                    line = line_raw.strip()
+                    if not line:
+                        # Ignore blank lines
+                        continue
+                    tool_opts_lines.append(line)
+                tool_opts = " ".join(tool_opts_lines)
```
webbnh, line 133: I think you can do this in two code lines:  read the text, and do a regex substitution which replaces the newlines with blanks while removing blank lines, e.g.:
```
                tool_opts_text = (self.tg_dir / host / tool).read_text().strip()
                tool_opts = re.sub(r'\n\s*', ' ', tool_opts_text)
```

lib/pbench/agent/tool_group.py
```diff
@@ -0,0 +1,158 @@
...
+        """
+        tools = dict()
+        for tool, opts in self.toolnames.items():
+            try:
+                host_opts = opts[host]
+            except KeyError:
+                # This host does not have this tool registered, ignore.
+                pass
+            else:
+                tools[tool] = host_opts
```
webbnh, line 151: How about replacing that big, ugly `try` block with this?
```
            # If the host does not have this tool registered, then ignore it.
            if host in opts:
                tools[tool] = opts[host]
```

lib/pbench/agent/tool_meister.py
```diff
@@ -396,66 +400,23 @@ class DcgmTool(PersistentTool):
...
-                    str(self.install_path / "bindings"),
-                    str(self.install_path / "bindings" / "common"),
-                ]
-                unit_tests = bool(os.environ.get("_PBENCH_UNIT_TESTS"))
-                prev_path = os.environ.get("PYTHONPATH", "")
-                if prev_path and not unit_tests:
-                    new_path_l.append(prev_path)
-                self.env = os.environ.copy()
-                self.env["PYTHONPATH"] = ":".join(new_path_l)
+            self.args = [executable]
```
webbnh, line 413: Go for it!  Crush this function down to a single line!  (Well, OK, three, but, still!)
```
        self.args = [executable] if executable is not None else None
```

lib/pbench/agent/tool_meister.py
```diff
@@ -1755,10 +1720,17 @@ def main(argv):
...
+        if not redis_host or not redis_port or not param_key:
+            print(f"{PROG}: Invalid arguments: {argv!r}", file=sys.stderr)
+            return 1
     try:
         daemonize = argv[4]
     except IndexError:
         daemonize = "no"
+    else:
+        if not daemonize:
+            daemonize = "no"
```
webbnh, line 1733: My earlier comments on the analogous code in the TDS also apply here:  consider using a proper boolean for `daemonize` and consider using a simple length test on `argv` rather than a full-blown `try` block.

lib/pbench/agent/tool_meister.py
```diff
@@ -1823,14 +1795,12 @@ def main(argv):
...
+    _PROG = PROG if daemonize != "yes" else None
     try:
-        params_raw = redis_server.get(param_key)
-        if params_raw is None:
-            print(
-                f'{PROG}: Parameter key, "{param_key}" does not exist.', file=sys.stderr
-            )
-            return 6
-        params_str = params_raw.decode("utf-8")
+        # Loop waiting for the key to show up.
```
webbnh, line 1800: See my comment on this code comment text in the TDS code.

webbnh (approved):
Done at last!  Nevertheless, here's another pile of nits, pointed questions, and suggestions.

There's nothing (that I remember) which _must_ be addressed before merging, but I hope that you'll consider making some changes.

portante:
This rebase addresses the test failures from the merged `dcgm` update work.

Another set of commits will follow addressing comments as appropriate.

agent/util-scripts/pbench-tool-meister-start
portante: Tracked in https://github.com/distributed-system-analysis/pbench/issues/2163.

agent/util-scripts/pbench-tool-meister-start
portante: Tracked in https://github.com/distributed-system-analysis/pbench/issues/2164.

agent/util-scripts/pbench-tool-meister-start
portante: Tracked in https://github.com/distributed-system-analysis/pbench/issues/2165.

agent/util-scripts/pbench-tool-meister-start
portante: Tracked in https://github.com/distributed-system-analysis/pbench/issues/2166.

agent/util-scripts/pbench-tool-meister-start
portante: Tracked in https://github.com/distributed-system-analysis/pbench/issues/2167.

lib/pbench/agent/tool_meister.py
portante: Tracked in https://github.com/distributed-system-analysis/pbench/issues/2168.

agent/containers/images/Dockerfile.layered.j2
```diff
@@ -1,10 +1,14 @@
...
+{% if kind in ('tools', 'all') %}
+COPY ./{{ distro }}-pcp.repo /etc/yum.repos.d/pcp.repo
+{% endif %}
+
 # Install all the RPMs required for this image.
 #
 # FIXME: this is not exhaustive, it does not include RPMs to support
 #        Kubernetes or RHV environments.
-RUN {% if distro == 'centos-7' %}yum{% else %}dnf{% endif %} install -y {% if distro == 'centos-8' %}--enablerepo powertools {% endif %}{{ rpms }} && \
+RUN {% if distro == 'centos-7' %}yum{% else %}dnf{% endif %} install -y --setopt=tsflags=nodocs {% if distro == 'centos-8' %}--enablerepo powertools {% endif %}{% if kind in ('tools', 'all') %}--enablerepo pcp {% endif %}{{ rpms }} && \
```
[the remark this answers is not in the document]
portante: > Does `centos-8` use `dnf`?

Yes, Fedora, CentOS 8+, and RHEL 8+ are all `dnf` based.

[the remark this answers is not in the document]
portante: > I don't know if you're interested ...

Tracked in https://github.com/distributed-system-analysis/pbench/issues/2169.

agent/util-scripts/pbench-tool-meister-client
```diff
@@ -68,9 +68,20 @@ def main(argv):
...
             return 0
 
+    redis_server = os.environ.get("PBENCH_REDIS_SERVER", f"localhost:{def_redis_port}")
+    parts = redis_server.split(":", 1)
+    if len(parts) != 2:
+        raise Exception("Bad Redis server specified, '{redis_server}'")
+    try:
+        redis_port = int(parts[1])
+    except Exception:
+        raise Exception("Bad Redis port specified, '{redis_server}'")
```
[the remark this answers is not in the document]
portante: Let's see if exception string changes will be acceptable.

agent/util-scripts/pbench-tool-meister-client
```diff
@@ -68,9 +68,20 @@ def main(argv):
...
+    else:
+        redis_host = parts[0]
```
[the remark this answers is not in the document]
portante: Tracked in https://github.com/distributed-system-analysis/pbench/issues/2170.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -411,15 +295,17 @@ def main(_prog, cli_params):
 
     # Load the tool metadata
     try:
-        inst_dir = PbenchAgentConfig(
-            os.environ["_PBENCH_AGENT_CONFIG"]
-        ).pbench_install_dir
-    except BadConfig as exc:
-        logger.error("%s", exc)
+        inst_dir = os.environ["pbench_install_dir"]
```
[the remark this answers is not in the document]
portante: No because these are well defined environment variables provided by `agent/base`, which we've published.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -574,11 +480,11 @@ def main(_prog, cli_params):
...
         to_client_channel = f"{cli_tm_channel_prefix}-{tm_channel_suffix_to_client}"
-        redis_server = redis.Redis(host="localhost", port=redis_port, db=0)
+        redis_server = redis.Redis(host=redis_host, port=redis_port, db=0)
         to_client_chan = RedisChannelSubscriber(redis_server, to_client_channel)
     except Exception as exc:
         logger.error(
-            "Unable to connect to redis server, %s:%d: %r", "localhost", redis_port, exc
+            "Unable to connect to redis server, %s:%d: %r", redis_host, redis_port, exc
         )
         return kill_redis_server(redis_pid, ReturnCode.REDISCHANFAILED)
```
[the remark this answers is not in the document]
portante: Tracked in https://github.com/distributed-system-analysis/pbench/issues/2171.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -835,12 +757,16 @@ def main(_prog, cli_params):
...
         " to show up as subscribers"
     )
     ret_val = wait_for_tds(to_client_chan, logger)
     if ret_val != 0:
-        return kill_redis_server(redis_pid, ReturnCode.TDSWAITFAILURE)
+        if cli_params.redis_server is None:
+            # We created the Redis server, so we should clean it up.
+            return kill_redis_server(redis_pid, ReturnCode.TDSWAITFAILURE)
+        else:
+            return ReturnCode.TDSWAITFAILURE
```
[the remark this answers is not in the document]
portante: Tracked in https://github.com/distributed-system-analysis/pbench/issues/2172.

agent/util-scripts/pbench-tool-meister-stop
```diff
@@ -42,6 +38,88 @@ def is_running(pid):
     return True
 
 
+def wait_for_pid(pid):
+    """wait_for_pid - wait for a process to actually stop running.
+    """
+    while is_running(pid):
+        time.sleep(0.1)
```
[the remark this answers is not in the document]
portante: See https://pubs.opengroup.org/onlinepubs/9699919799/functions/wait.html.

agent/util-scripts/pbench-tool-meister-start
portante: https://github.com/distributed-system-analysis/pbench/issues/2173

agent/util-scripts/pbench-tool-meister-stop
portante: Tracked in https://github.com/distributed-system-analysis/pbench/issues/2174.

agent/util-scripts/pbench-tool-meister-stop
portante: Tracked in https://github.com/distributed-system-analysis/pbench/issues/2175.

agent/util-scripts/pbench-tool-meister-stop
portante: > Do we want to support ":55555" as a shorthand for "localhost:55555"?

Tracked in #2170.

portante: > Y'know, this code should be refactored so that we get the same support for `-start`, `-stop`, and `-client`....

Tracked in https://github.com/distributed-system-analysis/pbench/issues/2176.

agent/util-scripts/pbench-tool-meister-stop
portante: Tracked in https://github.com/distributed-system-analysis/pbench/issues/2177.

agent/util-scripts/pbench-tool-meister-stop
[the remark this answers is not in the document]
portante: I have a PR coming which is going to address this with a refactoring of how we handle specifying the Redis server argument.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1683,89 +2059,44 @@ def main(argv):
...
-                    bind_hostname,
-                    tds_port,
-                )
-            else:
-                logger.exception("ERROR - failed to start the tool data sink")
-        except Exception:
-            logger.exception("ERROR - failed to start the tool data sink")
-
-    return 0
+    if daemonize == "yes":
```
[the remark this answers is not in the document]
portante: In general I try to avoid booleans, this case though `yes` means `True` and anything else means `False`.  Changed the code to reflect that.

agent/util-scripts/tool-meister/pbench-tool-data-sink
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env python3
+
+"""Simple command-line wrapper to keep the tool data sink from being in the
+CLI command set, while still allowing it to be invoked by the container entry
+point.
+"""
```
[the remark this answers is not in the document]
portante: > I don't see how the existence of this script prevents the TDS from being in the CLI command set

When the pbench-agent RPM is installed, `/opt/pbench-agent/util-scripts/tool-meister` will not be in the `PATH`.

> nor do I see how this script is necessary to allow the tool to be invoked by the container entry point.

Not sure there is much value collapsing all these files.  The `pbench-tool-data-sink` and `pbench-tool-meister` executables are the Python3 shell wrappers to invoke the respective main classes.  The `*-ep` are the specific entry points behaviors for each of the containers, and the `pbench-tool-meister-remote` is what is invoked when running Tool Meisters via SSH.

All of these are separate concerns.

agent/util-scripts/tool-meister/tool-data-sink-ep
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+_dir="$(dirname ${0})"
+
+source /etc/profile.d/pbench-agent.sh
+source /opt/pbench-agent/base
+# Instruct the Tool Data Sink not to daemonize.
+/opt/pbench-agent/util-scripts/tool-meister/pbench-tool-data-sink "${REDIS_HOST}" "${REDIS_PORT}" "${PARAM_KEY}" no
```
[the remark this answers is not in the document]
portante: Unfortunately, `_dir` and the `source` invocations mimic the environment that all the pbench-agent `util-scripts` expect.  So that needs to be provided by these files.  And I'd rather keep these concerns here in a `bash` file than in a `Dockerfile` to avoid restrictions.

portante:
Working my way through code review comments / feedback.  I am aggressively resolving comments, adding tracking issues to defer as much code change to future PRs.

agent/util-scripts/unittests
```diff
@@ -528,7 +528,7 @@ function sort_log_file {
 }
 
 function sort_tdslog {
-    sort_log_file ${_testdir}/mock-run/tm/pbench-tool-data-sink.log
+    sort_log_file ${_testdir}/mock-run/tm/pbench-tool-data-sink.err
```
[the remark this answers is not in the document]
portante: Yes, the logger was changed to use "console" logging which emits on `stderr`; see line 1851 of `lib/pbench/agent/tool_data_sink.py`.

lib/pbench/agent/redis.py
```diff
@@ -220,3 +221,46 @@ def emit(self, record):
...
             self.counter += 1
+
+
+def wait_for_conn_and_key(
+    redis_server, key, PROG=None, redis_host=None, redis_port=None
+):
+    """wait_for_conn_and_key - convenience method of both the Tool Meister and
+    the Tool Data Sink to startup and wait for an initial connection to the
+    Redis server, and for the expected key to show up.
+    """
```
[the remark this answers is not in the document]
portante: Not quite.  The Redis server can be present, but the name of the key we were given on the command line may not have been `set` in the Redis server yet.

lib/pbench/agent/redis.py
```diff
@@ -220,3 +221,46 @@ def emit(self, record):
...
+    payload = None
+    while payload is None:
+        try:
+            payload = redis_server.get(key)
+        except redis.ConnectionError:
+            if connected is None:
+                if PROG is not None:
+                    print(
+                        f"{PROG}: waiting to connect to redis server {redis_host}:{redis_port}"
+                    )
```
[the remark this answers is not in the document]
portante: No, that is really a bug.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -550,6 +610,117 @@ def terminate(self):
...
+        # it'll be using the default benchmark run directory.  If the
+        # benchmark_run_dir parameter is valid, there will be a file
+        # called ".path" in the default benchmark run directory which will
+        # match.
+        #
+        # E.g.:
+        #  $ pbench_run="/home/<USER>/run-dir"
+        #  $ benchmark_run_dir="${pbench_run}/script_config_<date>"
+        #  $ cat ${pbench_run}/.path
+        #  /home/<USER>/run-dir
```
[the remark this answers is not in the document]
portante: Was trying to show the output instead.

lib/pbench/agent/redis.py
```diff
@@ -220,3 +221,46 @@ def emit(self, record):
...
+                connected = False
+            elif connected:
+                if PROG is not None:
+                    print(
+                        f"{PROG}: disconnected from redis server {redis_host}:{redis_port}"
+                    )
+                connected = False
```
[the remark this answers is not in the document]
portante: When this code runs, the `redis_server` object may not have actually connected to the Redis server.  So if we `.get()` and empty value back from that call we know we have established a connection to the Redis server, but the "key" is not present.  So we move from `None` to `True` for `connected`.  Then we make repeated calls, once a second, via `.get()` to see when the key shows up.  If a `ConnectionError` is raised while doing that, we might loose the connection, but we just flag that state and keep waiting.

The problem that I see is that we don't have a maximum time to wait for a connection and a key to show up.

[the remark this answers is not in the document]
portante: Tracking in https://github.com/distributed-system-analysis/pbench/issues/2178.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -124,25 +124,78 @@ def log_request(self, code="-", size="-"):
...
-        with self._lock:
-            self._server = server
-            self._cv.notify()
-        self._logger.debug("Running tool data sink WSGI server ...")
-        self._server.serve_forever()
+        try:
+            server = make_server(self.host, self.port, app, **self.options)
+        except OSError as exc:
+            assert exc.errno != 0, "Logic bomb!  OSError exception with no errno value"
+            with self._lock:
```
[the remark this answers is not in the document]
portante: We don't bother logging the exception because we are passing the exception string and the `errno` value to the waiting thread, which will handle the behavior from there.  We only log the general exception because we don't know what kind of state the caller will be in.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -550,6 +610,117 @@ def terminate(self):
...
+        #  $ benchmark_run_dir="${pbench_run}/script_config_<date>"
+        #  $ cat ${pbench_run}/.path
+        #  /home/<USER>/run-dir
+        #  $ podman run --volume ${pbench_run}:/var/lib/pbench-agent \
+        #    pbench-agent-tool-data-sink bash
+        #  [ abcdefg /]$ cat /var/lib/pbench-agent/.path
+        #  /home/<USER>/run-dir
+        try:
+            benchmark_run_dir_lcl = self._ext_benchmark_run_dir.resolve(strict=True)
+        except Exception:
```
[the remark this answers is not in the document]
portante: Perhaps, but we raise a `ToolDataSinkError()` eventually which has the original exception string in it.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -550,6 +610,117 @@ def terminate(self):
...
+            )
+            dot_path = self._int_pbench_run / ".path"
+            try:
+                dot_path_contents = dot_path.read_text().strip()
+            except Exception as exc:
+                # Failed to read ".path" contents, give up.
+                raise ToolDataSinkError(
+                    f"Run directory parameter, '{ext_benchmark_run_dir}', must"
+                    f" be an existing directory ('{self._ext_pbench_run}/"
+                    f".path' not found, '{exc}').",
```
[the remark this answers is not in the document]
portante: I am not sure how much nuanced exception handling buys us here, because we are reporting the string of the original exception.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -550,6 +610,117 @@ def terminate(self):
...
+                    raise ToolDataSinkError(
+                        f"Run directory parameter, '{ext_benchmark_run_dir}',"
+                        " must be an existing directory (.path contents"
+                        f" mismatch, .path='{dot_path_contents}' !="
+                        f" '{self._ext_pbench_run}').",
+                    )
+        else:
+            # We can access the benchmark_run_dir directly, no need to
+            # consider contents of ".path" file.
+            pass
```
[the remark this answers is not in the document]
portante: The clause is documentation.  If we can read the external benchmark run directory, we are in the same file system name space as the entity that created the Tool Data Sink process.  If we can't read the external benchmark run directory, then we construct an internal run directory path to the same leaf name as the given external benchmark run directory.  If that directory exists, and it has a `.path` in which contains the same name as the external benchmark run directory we were given, then we know we have access to the same directory.  If we don't have that confirmation, we don't want to write any data anywhere, we must exit.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1609,71 +1831,225 @@ def put_document(self, data_ctx, hostname):
...
         redis_port = argv[2]
         param_key = argv[3]
     except IndexError as e:
-        logger.error("Invalid arguments: %s", e)
+        print(f"{PROG}: Invalid arguments: {e}", file=sys.stderr)
         return 1
+    else:
+        if not redis_host or not redis_port or not param_key:
+            print(f"{PROG}: Invalid arguments: {argv!r}", file=sys.stderr)
+            return 1
```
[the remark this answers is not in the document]
portante: It saves one line by adding a test that is not needed.  See https://stackoverflow.com/questions/7604636/better-to-try-something-and-catch-the-exception-or-test-if-its-possible-first.

lib/pbench/agent/tool_data_sink.py
portante: See https://stackoverflow.com/questions/7604636/better-to-try-something-and-catch-the-exception-or-test-if-its-possible-first.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1609,71 +1831,225 @@ def put_document(self, data_ctx, hostname):
...
+    pbench_run,
+    hostname,
+    tar_path,
+    cp_path,
+    param_key,
+    params,
+    optional_md,
+):
+    # Disconnect any existing connections to the Redis server.
+    redis_server.connection_pool.disconnect()
```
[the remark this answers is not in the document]
portante: It can't by design, as `daemon()` is only ever called by `main()`, which always has `redis_server` defined as a `redis.Redis()` instance.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1609,71 +1831,225 @@ def put_document(self, data_ctx, hostname):
...
+        ) as tds_app:
+            tds_app.execute()
+    except OSError as exc:
+        if exc.errno == errno.EADDRINUSE:
+            logger.error(
+                "ERROR - tool data sink failed to start, %s:%s already in use",
+                params["bind_hostname"],
+                def_tds_port,
+            )
+            ret_val = 8
```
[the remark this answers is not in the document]
portante: Tracked in #2168.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -558,20 +729,41 @@ class ToolDataSink(Bottle):
...
+            benchmark_run_dir = BenchmarkRunDir(_benchmark_run_dir, pbench_run)
+            return (
+                benchmark_run_dir,
+                bind_hostname,
+                channel_prefix,
+                tool_group,
+                tool_metadata,
+                tool_trigger,
+                tools,
+            )
```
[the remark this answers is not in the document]
portante: Tracked in #2179.

lib/pbench/agent/tool_data_sink.py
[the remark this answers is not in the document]
portante: Tracked in #2179.

lib/pbench/agent/tool_group.py
```diff
@@ -0,0 +1,158 @@
+import errno
+import os
+
+from pathlib import Path
+
+
+class BadToolGroup(Exception):
```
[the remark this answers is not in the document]
portante: Tracked in #2180.

lib/pbench/agent/tool_group.py
portante: https://stackoverflow.com/questions/7604636/better-to-try-something-and-catch-the-exception-or-test-if-its-possible-first

lib/pbench/agent/tool_group.py
portante: Sure, but it seemed much simpler to compare a string first then to create a whole object and then execute a file system operation.

lib/pbench/agent/tool_group.py
portante: The comments are there to describe that we are changing from consider an element of the list as directory entry to the pbench abstraction of a host name or a tool name.

portante:
Okay folks, I believe I have addressed all the code review comments made that are worth making in this PR still.  I have filed a host of issues tracking future work that I felt should be done in a separate piece of work.

The code review comments are made in separate commits.  Once they have been reviewed, I'll squash them down before merging.

Maxusmusti (dismissed):
Tested traditionally (usual setup) and containerized, with visualization. All worked as expected.

agent/util-scripts/pbench-tool-meister-stop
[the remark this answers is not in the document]
webbnh: Could you be more specific?  All I could find in that reference was a recommendation that portable applications _use_ `waitpid()` (i.e., rather than `wait()`).

lib/pbench/agent/tool_data_sink.py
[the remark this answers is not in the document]
webbnh: Why would you avoid a boolean in a cases where there are only two possible values?

lib/pbench/agent/tool_data_sink.py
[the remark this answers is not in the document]
webbnh: My point was that it appears that we know what the contents of the `.path` file should be -- namely, it should match the contents of the `pbench_run` environment variable; so, rather than showing what the contents of the file should be, we can just put the corrent contents into it...thereby removing a point of failure.

lib/pbench/agent/tool_data_sink.py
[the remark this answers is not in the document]
webbnh: The purpose of lines 671-700 appears to be to set the value of `benchmark_run_dir_lcl`, and that's done before consulting `.path`.  Everything from 679 onwards is testing the value of `.path`, but that value is not actually used -- by that point `benchmark_run_dir_lcl` has _already been set_.

So, it's not obvious how there would be any problem if we were to delete the last 20 lines of this block.  I understand that you're using the outermost `else` block for documentation purposes, but I don't think that it's worth taking up that space.  And, furthermore, I don't see why we need to worry about the contents of `.path`, since we don't seem to be basing anything on it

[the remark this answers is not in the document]
webbnh: So, you're trying to guard against the case where our guess at where to write the results from inside the container exists and is writable but is not actually the place where the user wants us to write?

lib/pbench/agent/tool_data_sink.py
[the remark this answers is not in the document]
webbnh: It saves two lines, and it replaces with a simple `if` test all of the infrastructure required to capture the exception context (to wit, the `try` statement).  And, I think the result is better, more readable code, if for no other reason than it removes much of the indentation which requires the reader to track the scopes that the execution enters and leaves.

lib/pbench/agent/tool_data_sink.py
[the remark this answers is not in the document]
webbnh: I chatted with Peter and he conveyed to me the level of pain and suffering that he experienced while developing the containerization.  This convinced me that the `.path` mechanism is worthwhile, in order to safeguard our users from a similar experience.  We can work on making it more friendly for the user to engage with, but it's worth keeping.

lib/pbench/agent/tool_data_sink.py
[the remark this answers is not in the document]
webbnh: As we discussed in realtime, in this example use case description, we have two sources for a common truth -- we have the `pbench_run` variable which is used in the `podman` invocation, and we have the contents of the `.path` file which is used inside the container.  Since they both need to contain the same value, we should suggest that one be derived from the other (either by using the command I suggested above to generate the `.path` file from the variable's value, or by using the contents of the file to define the variable or using it directly on the `podman` command line).

lib/pbench/agent/tool_data_sink.py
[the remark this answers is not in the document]
webbnh: OK, I'm impressed:  apparently Python has found some way to make `try` blocks really cheap (at least on X86_64) -- my suggested `if` construct is marginally _more_ expensive in the "normal" case.  (It's considerably faster in the exception case, but I don't think we're particularly interested in that, here.)

So, I guess this belongs in the "molehill" category rather than the "mountain" one.

ndokos (dismissed):

lib/pbench/agent/tool_data_sink.py
webbnh: I may have to take back my comment about the 12-step program and getting off of exceptions:  while I still think using `try` blocks makes the code harder to read, they definitely seem to make it perform _better_, which is counterintuitive, given what I know about _implementing_ exception handling....

lib/pbench/agent/tool_group.py
webbnh: Off the cuff, I would say that that SO article doesn't apply here, because having no tool registered isn't really an exceptional case (e.g., it's not an error, and the code provides for it as a part of normal operation).

However, from a performance standpoint, Python once again produces surprising (to me...) results.  In the non-exception case, my suggested code runs a few percent slower than the `try` block.  I would surmise that that is because it's doing the lookup twice (once for the `in` operation and once for the fetch); however, removing that redundancy using the following code:
```
     tmp = opts.get(host)
     if tmp is not None:
         tools[tool] = tmp
```
actually *slowed* the execution down by a significant amount!

And, in the case where the key is not present, while the `try` block is much more expensive than the others (as I expected...), the `in` operator is considerably faster than the `.get()` operation!

So, clearly there are some flaws in my understanding of how this stuff actually works.  :-(

lib/pbench/agent/tool_group.py
webbnh: That's fair.  But, consider that you're comparing every directory entry against "\_\_trigger\_\_" vs. doing a single directory lookup when `hdirent` happens to be "\_\_trigger\_\_".  Based on my earlier learnings today,  wouldn't want to try to guess which one costs more, but I expect that it has something to do with the total number of directory entries...so the bigger the directory, the more likely it is to be a win to _not_ have the special case.

agent/util-scripts/pbench-tool-meister-client
```diff
@@ -47,21 +47,26 @@ def main(argv):
     try:
         group = argv[1]
     except IndexError:
-        raise Exception("Missing group argument")
+        logger.error("Missing group argument")
```
webbnh, line 50: Just to confirm...you want the diagnostics sent to the _log_ and not to the user?  (Or, is it the case that, in this context, the log is `stderr`?)

agent/util-scripts/pbench-tool-meister-client
```diff
@@ -71,11 +76,13 @@ def main(argv):
     redis_server = os.environ.get("PBENCH_REDIS_SERVER", f"localhost:{def_redis_port}")
     parts = redis_server.split(":", 1)
     if len(parts) != 2:
-        raise Exception("Bad Redis server specified, '{redis_server}'")
+        logger.error("Bad Redis server specified, {!r}", redis_server)
```
webbnh, line 79: Why the `!r` directive?  `redis_server` is a `string`, right?  Is this just a trick for adding quotation marks to the output?  (You didn't use this trick for `action` at line 65....)

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -301,34 +300,25 @@ def main(_prog, cli_params):
             "The required 'pbench_install_dir' environment variable appears to be missing"
         )
         return ReturnCode.BADAGENTCONFIG
-    except Exception as exc:
+    try:
+        tm_start_path = Path(inst_dir).resolve(strict=True)
```
webbnh, line 304: This `try` block used to be in the `else` clause of the previous `try` block.  While I'm generally happy to see these `else` clauses removed, I'm suspicious that it was done by accident, here (i.e., there is still one at line 316).  Did you mean to do this?

lib/pbench/agent/tool_data_sink.py
```diff
@@ -640,6 +642,12 @@ class BenchmarkRunDir:
     handle that seamlessly.
     """
 
+    class Exists(Exception):
+        pass
+
+    class Prefix(Exception):
+        pass
```
webbnh, line 649: It would be nice to have some clues as to the intended semantics of these exceptions.  As it stands, these are generic wrappers of a generic exception, without any message text or even code comments.

At a minimum, they should have explanatory code comments.  Beyond that, though, they could be derived from `NotADirectoryError` or `FileNotFoundError`.  Or, you could declare them to be instances rather than classes, and provide a message, like
```
Exists = ValueError("Benchmark run directory does not exist")
```

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1007,9 +1015,9 @@ def record_tms(self, tms):
         """record_tms - record the Tool Meister data and metadata returned from
         the startup acknowledgement messages collected in "tms".
 
-        The first thing we have to do is setup self._tm_tracking properly,
-        adding which tools are no-ops, transient, and persistent, and properly
-        record the initial "posted" state.
+        The first thing we have to do is to determine which tools are no-ops,
+        no-ops, transient, and persistent, and properly record the initial
```
webbnh, line 1019: You have "no-ops" listed twice.

webbnh (approved):
I've got a small question, a couple of confirmations, a nit, and a request for some enhancement (add some code comments, if nothing more).  But, none are of such weight that we must hold up the merge...it's up to you.

agent/util-scripts/tool-meister/tool-data-sink-ep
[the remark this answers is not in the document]
webbnh: But...does `pbench-tool-data-sink` actually use `_dir` or anything that the `source` invocations provide?  If not, then we _could_ get by without them here...in which case, we __could__ invoke `pbench-tool-data-sink` directly using the `Dockerfile` instead of adding the clutter and complication of a wrapper script that we don't need.

agent/util-scripts/tool-meister/pbench-tool-data-sink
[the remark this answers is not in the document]
webbnh: > When the `pbench-agent` RPM is installed, `/opt/pbench-agent/util-scripts/tool-meister` will not be in the PATH.

Wouldn't that be just as true if we didn't create this file?  (That is, if you really want to create this file, perhaps the header comment should say something different:  pointing out that this script is used as the container entry point is good; mentioning that it is not and should not be a part of the CLI command set is good; but, that part about how this script prevents that from happening is confusing, at least to me.)

> Not sure there is much value collapsing all these files.

The value is in reducing the number of moving parts.  The invocation of the TDS is so straightforward that it really doesn't need a script wrapped around it (it really is just a Python command, AFAICS).  Omitting the script means one fewer file to be included in the RPM, installed in the right place, and otherwise tracked and debugged.  If the script actually provided some functionality (or any other significant positive feature) I would say go with it, but it's just a forwarding wrapper, and we don't need it here (the conclusions for the other scripts might be different, but I doubt it).

lib/pbench/agent/redis.py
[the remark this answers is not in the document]
webbnh: I think we're having a problem with terminology....

When a publisher sends a message, it associates with that message a key, right?  Correspondingly, when a subscriber requests a message, it does so by specifying a key, right?  The result is that the subscriber receives only messages associated with the specified key...which means that the TM/TDS waits here not for the _key_ but rather for a __message__ associated with that key to show up.  No?

agent/util-scripts/pbench-tool-meister-client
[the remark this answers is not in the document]
webbnh: The code is reporting the unparsed "redis server" argument, but it's objecting here to just the port; thus, I was thinking that it might provide better insight if we reported specifically what the code thought the port value was, rather than the unparsed input.

agent/util-scripts/pbench-tool-meister-client
portante: Sort of, it renders the `repr` version of the object, so that if it is not a string, we see that in the exception string.

agent/util-scripts/pbench-tool-meister-client
portante: I believe in this case the log is `stderr`.  See https://docs.python.org/3/library/logging.handlers.html#logging.StreamHandler.

agent/util-scripts/pbench-tool-meister-start
portante: Yes this was intentional.  I removed the `else:` for `inst_dir`, but should have also removed the `else:` for `tm_start_path`.  I believe this is further refactored in PR #2181.

lib/pbench/agent/tool_data_sink.py
portante: Tracking in #2184.

portante:
Rebased and resolved all code review comments.  Unfortunately, I hit the bug solved by PR #2185, which delayed the verification of the tests (`py3-agent` and `py3-server`) tox tests (any pytest based test) would fail.

Let's get PR #2185 to land first, and then this one.

agent/util-scripts/pbench-tool-meister-client
webbnh: `os.environ.get()` in this specific case returns a `string`, so `repr`, as far as I can tell, just adds single quotes...which is kind of cute, but just a little weird.

portante: So I want to maintain the discipline of using `repr` in these cases so that when things change (link when I changed from having a `string` to a `Path` and I got errors) the `repr` would have shown me that, but forced "stringification" means that I never see the actual difference.

portante:
Okay, now that we landed #2185, this is ready again for final review.

dbutenhof (dismissed):

portante:
Had to rebase to pick up PR #2186.

dbutenhof (approved):

webbnh (approved):
@portante, merge this quick before someone merges something else!

PR #2083, opened 2021-01-29
Run pbench-server systemd service as a pbench user

npalaska:
- Run pbench-server service as a pbench user
- Remove some `ExecStartPre` commands and use `RuntimeDirectory` to create a sub-directory inside `/run` for storing gunicorn pid

### DIFF
--- a/server/lib/systemd/pbench-server.service
+++ b/server/lib/systemd/pbench-server.service
@@ -10,12 +10,12 @@ After=network.target httpd.service
 
 [Service]
 Type = simple
+User = pbench
+Group = pbench
+RuntimeDirectory = pbench-server
 PIDFile = /run/pbench-server/gunicorn.pid
-ExecStartPre = /bin/mkdir /run/pbench-server
-ExecStartPre = /bin/chown pbench:pbench /run/pbench-server
 ExecStart = /opt/pbench-server/bin/pbench-server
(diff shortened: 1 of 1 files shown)


server/lib/systemd/pbench-server.service
```diff
@@ -10,12 +10,11 @@ After=network.target httpd.service
 
 [Service]
 Type = simple
+User = pbench
```
dbutenhof, line 13: I wonder if you should specify `Group = pbench` as well? Aside from that, this looks good!

dbutenhof (approved):

server/lib/systemd/pbench-server.service
npalaska: yeah we can add `Group = pbench` as well, I was just not sure if user pbench should be part of group pbench or some other group. Usually user by default is in the same group as its name so adding a `Group=` is fine.

portante (changes requested):
Looks good, let's get the `group` setting in for `pbench`.

server/lib/systemd/pbench-server.service
npalaska: fixed

portante (approved):

dbutenhof (approved):

PR #2087, opened 2021-01-31
Fix turbostat for real this time

portante:
Okay, commit f74a4c3b6 tried to fix the termination of the `turbostat` command but failed to miss one critical change: the `safe_kill` call needs the name of the process to be killed.  That previous commit changed the behavior such that the process name is now "turbostat", but we were still looking for "turbostat-datalog".

Along the way, fix `base-tool` to properly handle relative path arguments which facilitate easier manual testing.

See also PR #2041.

### DIFF
--- a/agent/tool-scripts/base-tool
+++ b/agent/tool-scripts/base-tool
@@ -512,7 +512,6 @@ else
 fi
 
 tool_output_dir="${tool_dir}/${tool_name}" # all tools keep data in their tool specific dir
-tool_pid_file="${tool_output_dir}/${tool_name}.pid"
 
 function safe_kill() {
 	# safe kill: check for strange situations and deal with them.  If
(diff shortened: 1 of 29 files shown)


dbutenhof (dismissed):

agent/tool-scripts/base-tool
```diff
@@ -711,19 +710,21 @@ start)
 
 	if [[ ! -z "${options}" ]]; then
 		# Record any options for later post-processing use.
-		echo "${options}" > ${tool_output_dir}/${tool_name}.options
+		echo "${options}" > ./${tool_name}.options
```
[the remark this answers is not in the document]
portante: Well, I changed it so that I can test it easily by hand.  One should be able to do:

```
cd /var/tmp/
mkdir test
tool-scripts/turbostat --start --dir=./test
```

But relative paths were, as they say, always relative.  So `base-tool` would look to write files in `/var/tmp/test/turbostat/test/turbostat`.

agent/tool-scripts/base-tool
```diff
@@ -711,19 +710,21 @@ start)
...
 	fi
 
-	tool_cmd_file="${tool_output_dir}/${tool_name}.cmd"
+	tool_cmd_file="./${tool_name}.cmd"
 	if [[ -z "${gopath}" ]]; then
 		if [[ "${tool_package_name}" == "golang" ]]; then
 			printf -- "%s: golang based tools require the --gopath option\n" "${tool_name}" >&2
+			popd > /dev/null
```
ndokos, line 720: Does it make a difference doing a `popd` before exiting? There are a few more instances on ll. 727 and 738 as well.

ndokos (changes requested):

agent/tool-scripts/base-tool
portante: I don't think so, but wanted to be make sure we explicitly left the context.  I am not in favor of assuming things work a certain way.

ndokos (approved):

dbutenhof (approved):

PR #2074, opened 2021-01-22
Build state tracker

dbutenhof:
This is an infrastructure mechanism  to track the state and metadata of a dataset (tarball) through the stages of the Pbench server. While the principle motive is to provide a way to persist the username who authenticated the `PUT` upload of the tarball through to indexing, it provides additional value including checking state consistency throughout the pipeline. (E.g., in the audit report generator; which will be done as a follow-on.)

Define a SQLAlchemy DB class `Dataset` which defines persistent state for a Pbench dataset including most importantly the dataset's username (which is not otherwise recorded until we index) and the dataset's current state so that we can track the progress of a dataset through the Pbench server pipeline.

A new dataset starts in `UPLOADING` state and will progress through the steps as we perform operations on it. We're defining both "in progress" -ing steps such as `INDEXING` as well as completion steps (which are ready for the next operation) such as `INDEXED`.

I've also defined a `Metadata` model to accommodate information that doesn't fit into a strictly sequenced state model; in particular, I've defined `REINDEX` and `ARCHIVED` boolean metadata keys; the former used by `pbench-reindex` to mark a dataset for re-indexing, (it will be cleared by `pbench-index` after re-indexing completes), and the latter used by `pbench-backup-tarballs` to mark that a dataset has been backed up. Note that nobody's yet _reading_ or validating either metadata key.

The server `unittest` script has been changed to set up the Sqlite DB in the config file, and also to dump the contents of the database as part of the directory/file dump following a test _if the database file exists, if the Dataset table exists, and if there is at least one row in the table_. The dump includes both the basic `Dataset` row and any `Metadata` rows associated with the dataset.

There are a lot of unit test changes; they're hopefully pretty obvious.

PR #2089, opened 2021-02-04
Ensure pbench develop install occurs for tests

portante:
In order for the Jenkins jobs to run correctly, we need to ensure that the `pbench` python3 module is installed, including the CLI binaries that need to be on the `PATH`.

We correct the version of `flake8` to be the one available in Fedora 32 (`3.7.7`), and add the missing RPMs to the development container to avoid unnecessary pip3 installs.  We also create a separate `docs/requirements.txt` file.

Finally, we update the branch name file to properly reference `main` so that we use the right image tag the development container.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,2 +1,3 @@
 [run]
+parallel = True
 data_file = /tmp/${USER}/cov/coverage.db
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,0 +1,1 @@
+sphinx-panels
--- a/jenkins/branch.name
+++ b/jenkins/branch.name
@@ -1,1 +1,1 @@
-master
+main
--- a/jenkins/development.Dockerfile
+++ b/jenkins/development.Dockerfile
@@ -67,6 +67,7 @@ RUN \
         python3-elasticsearch \
         python3-flake8 \
         python3-flask \
+        python3-flask-cors \
         python3-flask-restful \
         python3-gitdb \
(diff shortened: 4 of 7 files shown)


portante:
NOTE: I have not rebuilt the development container yet.  Once we all agree on these changes, we can take that step.

dbutenhof (dismissed):

ndokos (approved):
Looks reasonable, but I have not run anything. How do I test this? Do we need unit tests for the testing framework ("meta-unittests")?

dbutenhof:
> Looks reasonable, but I have not run anything. How do I test this? Do we need unit tests for the testing framework ("meta-unittests")?

Yeah, the test is that when "we" (Pete) builds this new container and pushes it to quay.io, this PR (and others) can have their Jenkins builds re-triggered and ought to work. ;-)

portante:
> Looks reasonable, but I have not run anything. How do I test this?

Good question.  This is trying to get it working again for PR #2074.  So as long as it passes on `main` still, we're good.

The easiest way to test this is to checkout the branch and run the same command that is in `jenkins/Pipeline.gy`. 

> Do we need unit tests for the testing framework ("meta-unittests")?

Not a bad idea.  Not sure our team has the bandwidth to take that on.  =)

portante:
Right now the coverage side is failing for some reason which I have to figure out.

dbutenhof (approved):

ndokos (approved):

PR #1937, opened 2020-11-02
Initial implementation of Pbench user model and associated APIs in server

npalaska:
Initial framework to build Pbench user authentication:

This implements 6 basic user APIs 
1. Register User:
    * Handles Pbench User registration via JSON request 
    * `POST` `/v1/register`
        ```
        json={
            "username": "username",
            "password": "password",
            "firstName": first_name,
            "lastName": "last_name",
            "email": "user@domain.com"
        }
        ```
2. Login User
    * Handles user login and returns a valid pbench auth token
    * `POST` `/v1/login`
        ```
        json={
            "username": "username",
            "password": "password"
        }
        ```
    * Returns a valid pbench auth token for accessing other APIs.
    * User is allowed to create multiple login requests and thus generating multiple auth tokens,
    however, in the future there will be limit on the number of auth token a user can generate.
    * We plan to maintain an `active_tokens` table in our db.
    * Each auth token has its own expiry and is associated with a user id.
    * Any subsequent API request will only pass if the auth token in the header is not expired and is present in `active_tokens` table.
    
3. Logout User
    * Handles the user logout mechanism.
    * `POST` `/v1/logout`
        ```
        headers={ 
        Authorization:   Bearer <Pbench authentication token (user received upon login)>
        }
        ```
    * Deletes the authentication token provided in the headers from the `active_tokens` table.
    * Once logged out user, can not use the same auth token for other API access.
4. Get User
    * Returns the user's self information that was registered, the username must be provided in the url
    * `GET` `/v1/user/<string:username>`
        ```
        headers={ 
        Authorization:   Bearer <Pbench authentication token (user received upon login)>
        }
        ```
    * If the `Authorization` header does not belong to the username provided in the url, we reject the request unless the `Authorization` token belongs to the admin user.
5. Delete User
    * An API for a user to delete an account from the pbench database.
    * `DELETE` `/v1/user/<string:username>"` headers={ 
        Authorization:   Bearer <Pbench authentication token (user received upon login)>
    }
    * A user can only perform the `DELETE` action on another account if the presented auth token belongs to an admin user.
    * If the `Authorization` header does not belong to the username provided in the url, we reject the request unless the `Authorization` token belongs to the admin user.
6. Update User
    * An API for updating the User registration fields, the username must be provided in the url
    * `PUT` `/v1/user/<string:username>`
        ```
        Example Json:
        {
            "first_name": "new_name",
            "password": "password",
            ...
        }
        headers={ 
[...]

PR #2101, opened 2021-02-10
Userdb

dbutenhof:
Add "user" ownership to Pbench Elasticsearch documents.

Extend the ES document templates with a new "phrase" defining the owning user and "access level" for the document. The default is user "public" and access "public" ... the owner is inherited from the Dataset (note that this PR layers on top of #2074), and a non-default user will have "private" access by default. (This is shown in several of the unit test gold files, via a new optional argument in the '.db' state files.)


.coveragerc
```diff
@@ -1,3 +1,4 @@
 [run]
 parallel = True
 data_file = /tmp/${USER}/cov/coverage.db
+omit = */site-packages/*
```
portante, line 4: Can you move this to a separate PR entirely?

.gitignore
```diff
@@ -7,6 +7,7 @@ build
 dist
 *~
 ????-*.patch
+.env
```
portante, line 10: We use .env files in some of the legacy unit tests on the agent side, perhaps make this for just the server side?

portante (commented):
Just a couple of quick comments.

.coveragerc
dbutenhof: #2102

.gitignore
portante: E.g.: https://github.com/distributed-system-analysis/pbench/blob/main/agent/bench-scripts/tests/pbench-trafficgen/test-39.env

portante: I think you are right so ignore more comment.

server/rpm/pbench-server.spec.j2
```diff
@@ -136,6 +136,8 @@ fi
 /%{installdir}/lib/pbench/server/api/resources/query_apis/__init__.py
 /%{installdir}/lib/pbench/server/api/resources/query_apis/elasticsearch_api.py
 /%{installdir}/lib/pbench/server/api/resources/query_apis/query_controllers.py
+/%{installdir}/lib/pbench/server/api/resources/query_apis/query_results.py
+/%{installdir}/lib/pbench/server/api/resources/query_apis/query_result.py
```
webbnh, line 140: I thought I heard Peter describing how we could specify a directory in this list instead of individual files...is that something that we should be pursuing here (if not wholesale, then piecewise covering changes)?

webbnh (commented):
Sorry if these comments are scattered or confusing:  part way through this review, you pushed another commit; and then, I realized that much of what I was reviewing belonged to #2074....  So, if one or more of my comments are misdirected, kindly take them as referring to the other review....

I didn't find anything monumental...just the usual collection of nits and stuff.

lib/pbench/server/api/resources/query_apis/query_result.py
```diff
@@ -0,0 +1,257 @@
...
+                    ' schema; use "Authorization: Bearer <JWT-token>"'
+                )
+                abort(401, message="invalid user authorization")
+
+        try:
+            start = parser.parse(start_arg).replace(day=1)
+            end = parser.parse(end_arg).replace(day=1)
+        except Exception as e:
+            self.logger.info(
+                "Invalid start or end time string: {}, {}: {}", start_arg, end_arg, e
```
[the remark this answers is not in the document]
webbnh: If you preinitialize `start` and `end`, you could check their values in the exception handling code to figure out which one triggered the exception.

lib/pbench/server/api/resources/query_apis/query_result.py
```diff
@@ -0,0 +1,257 @@
...
+                result = {
+                    "runMetadata": run_metadata,
+                    "hostTools": src["host_tools_info"],
+                }
+            except KeyError:
+                self.logger.exception("ES response not formatted as expected")
+                abort(500, message="INTERNAL ERROR")
+            else:
+                # construct response object
+                return jsonify(result)
```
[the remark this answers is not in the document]
webbnh: > it would probably be better to move the return statement inside the try block, since that's the scope of result.

Well, OK, I s'pose the contents of the `else` clause are arguably part of the `try` scope.  And, moving the `return` inside the `try` block risks an exception raised by the `jsonify()` call being misconstrued.  But, I still think it would be better to get rid of the `else` somehow.  ;-)

lib/pbench/server/api/__init__.py
```diff
@@ -53,6 +56,16 @@ def register_endpoints(api, app, config):
...
+    api.add_resource(
+        QueryResults,
+        f"{base_uri}/datasets/list",
+        resource_class_args=(config, app.logger),
+    )
+    api.add_resource(
+        QueryResult,
+        f"{base_uri}/datasets/detail",
+        resource_class_args=(config, app.logger),
+    )
```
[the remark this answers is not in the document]
dbutenhof: Yes, you've mentioned that. I would normally agree, but I'm torn on this one. The names I used are taken from the current dashboard code and are somewhat simple and meaningful (in that context), and continuing to use those tokens in the code eases the transition from old to new.

Besides, naming things is the hardest part of coding, and I'd have to come up with a completely new naming scheme that makes sense, changing the existing APIs as well as these new ones. (Of course, I've been very tempted to re-think my URI naming convention anyway, and doing so together would make sense; which doesn't necessarily mean I want to deal with that right now... I will, however, take your repeated nudges under consideration.)

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -31,3 +33,45 @@ def get_index_prefix(config):
...
+
+    # TODO: hardcoding the index here is risky. We need a framework to
+    # help the web services understand index versions and template
+    # formats, probably by building a persistent database from the
+    # index template documents at startup. This is TBD.
+    for monthValue in monthResults:
+        if index == "v4.result-data.":
+            queryString += f"{prefix + index + monthValue}-*,"
+        else:
+            queryString += f"{prefix + index + monthValue},"
```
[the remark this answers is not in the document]
dbutenhof: Well, I _did_ write it, just not in this PR.

More importantly, the code comment is a completely different angle on this, in that hardcoding the index prefix requires additional maintenance over time when we change the version of indices. Theoretically, I would be doing that, for example, in this PR as I'm changing the Elasticsearch document templates. I'm "cheating" for convenience as we haven't released the previous changes I made for Elasticsearch V7, and another version update seems pointless.

Currently, the indexer reads the template documents from disk, and generates a map of index names (e.g., `result-data` or `run`) to version (e.g., in this example, `v4`). I'd like to move that map into persistent storage accessible throughout the server so that we could refer her to the current version of the `result-data` index document template.

And, yes, that mapping could also encode (and I believe does) whether the date-series of the index is monthly (`run`) or finer granularity (`result-data`) requiring the trailing `-*`, making this logic much more robust. Separating out the template is a very minor simplification that doesn't much interest me right now, in context; it does nothing to address the real concerns with this code. ;-)

lib/pbench/server/api/resources/query_apis/query_result.py
```diff
@@ -0,0 +1,257 @@
...
+    directly, but rather allows building a set of Elasticsearch index
+    names that will contain the data. (Note that the dashboard uses the
+    same date range it used to find controllers; however it could make this
+    more efficient by using only the year and month in which the result
+    was generated.)
+
+    Note that although the name "queryResult" is retained for consistency
+    with the existing dashboard UI code, the word "result" is a bit
+    misleading in that this deals with Pbench "run" documents, not Pbench
+    "results" documents. It's a follow-on to queryResults, which returns
```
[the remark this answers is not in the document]
dbutenhof: Again, this is explaining the historical link; not that you don't have a point, but I'm still reluctant to break away from the dashboard conventions. Yeah, keep pushing, and I'll go over the cliff... I'm not that firmly grounded, but I'm not giving up quite that easily, either.

lib/pbench/server/api/resources/query_apis/query_result.py
```diff
@@ -0,0 +1,257 @@
...
+
+        try:
+            user = json_data["user"]
+            name = json_data["name"]
+            start_arg = json_data["start"]
+            end_arg = json_data["end"]
+        except KeyError:
+            keys = [k for k in ("user", "name", "start", "end") if k not in json_data]
+            self.logger.info("Missing required JSON keys {}", ",".join(keys))
+            abort(400, message=f"Missing request data: {','.join(keys)}")
```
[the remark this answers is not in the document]
dbutenhof: I've gone back and forth between proactive and optimistic error handling; Pete's strong preference for optimistic code and "wasting time" for diagnosis in post-error handling has definitely affected my style recently.

Of course I could simplify by just reporting the actual `KeyError` and not worrying about any other missing keys, but this seemed more useful since I hate reporting errors one at a time and pushing the client to iterate. (Another "optimistic but inclusive" strategy would be to use separate `try` blocks but join the results before reporting; but that just stretches the code out and makes it more awkward to read for an edge error case, which didn't seem justified.)

lib/pbench/server/api/resources/query_apis/query_result.py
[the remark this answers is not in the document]
dbutenhof: Yes, I could; I didn't really see any need for that level of error detail, so I didn't bother. Plus, again, if `start_arg` was bad, it'd leave me without any idea whether `end_arg` was bad without adding extra code for an edge case. Worthwhile? At the time, anyway, I thought not.

lib/pbench/server/api/resources/query_apis/query_result.py
```diff
@@ -0,0 +1,257 @@
...
+                headers={
+                    "Content-Type": "application/json",
+                    "Accept": "application/json",
+                },
+                json=payload,
+            )
+            es_response.raise_for_status()
+        except requests.exceptions.HTTPError as e:
+            self.logger.exception("HTTP error {} from Elasticsearch post request", e)
+            abort(502, message="INTERNAL ERROR")
```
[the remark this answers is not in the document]
dbutenhof: None of these are "more specific": they're all direct descendants of `RequestException`. I suppose I could have alphabetized them... perhaps I would have, had I thought of it at the time.

lib/pbench/server/api/resources/query_apis/query_result.py
```diff
@@ -0,0 +1,257 @@
...
+            abort(500, message="INTERNAL ERROR")
+        else:
+            run_metadata = {}
+            try:
+                es_json = es_response.json()
+                hits = es_json["hits"]["hits"]
+
+                # NOTE: we're expecting just one. We're matching by just the
+                # dataset name, which ought to be unique.
+                self.logger.info("{} datasets found", len(hits))
```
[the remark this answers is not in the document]
dbutenhof: If there are no hits, `hits` is an empty array, with length 0, so that's not an issue. If it's missing, something went seriously wrong with the Elasticsearch query, which will be caught as a `KeyError` below.

I really wanted to confirm that the query had finished. Since I cloned the body, I pretty much left this alone although you're right that logging "hits" in this case isn't of great value unless something weird happened.

lib/pbench/server/api/resources/query_apis/query_result.py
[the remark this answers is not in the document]
dbutenhof: Yes, I know you hate `try/else`, but I've gotten used to it; and I do think it's useful here to keep the result access clearly "within scope" but outside the REST call failure domain.

lib/pbench/server/api/resources/query_apis/query_results.py
```diff
@@ -0,0 +1,289 @@
...
+        time, and a starting UNIX timestamp.
+
+        TODO: We probably need to return all *unowned* runs when called
+        without a session token or user parameter. We need to define precisely
+        how this should work.
+
+        NOTE: This is the format currently constructed by the Pbench
+        dashboard `src/model/dashboard.js` fetchResults method, which
+        becomes part of the Redux state. (Note that the "key" of the sequence
+        is the same as the "run.name".)
```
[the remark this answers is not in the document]
dbutenhof: I'm not sure what you mean by "size of the ES indices" since this is about post-processing the ES query results into what the dashboard code expects.

If you're talking about the duplicated `"key"` value, that's a Javascript thing; a React component that includes an array of data will get lint-dinged if there's no `key`. But it has nothing to do with Elasticsearch...

lib/pbench/server/api/resources/query_apis/query_results.py
```diff
@@ -0,0 +1,289 @@
...
+                "run.controller": "dhcp31-187.perf.lab.eng.bos.redhat.com",
+                "run.start": "2020-04-29T12:49:13.560620",
+                "run.end": "2020-04-29T13:30:04.918704"
+            }
+        ]
+        """
+        json_data = request.get_json(silent=True)
+        if not json_data:
+            self.logger.info("Invalid JSON object. Query: {}", request.url)
+            abort(400, message="Missing request payload")
```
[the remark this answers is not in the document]
dbutenhof: Technically, yes. I suppose I could distinguish based on whether the non-Jsonified payload is empty. Or maybe I should drop the `silent=True` and hope the `BadRequest` exception is helpful... although the documentation I've read doesn't provide much basis for hope.

Frankly it didn't seem worth worrying about at the time; while diagnosing client programming errors non-destructively is important, I sometimes have mixed feelings about the level of detail we should feel obligated to provide when someone does something stupid. ;-)

lib/pbench/server/api/resources/query_apis/query_results.py
```diff
@@ -0,0 +1,289 @@
...
+            self.logger.info(
+                "Invalid start or end time string: {}, {}: {}", start_arg, end_arg, e
+            )
+            abort(400, message="Invalid start or end time string")
+
+        # We have nothing to authorize against yet, but print the specified
+        # user and session token to prove we got them (and so the variables
+        # aren't diagnosed as "unused")
+        self.logger.info(
+            "QueryControllers POST for user {}, prefix {}, session {}: ({}: {} - {})",
```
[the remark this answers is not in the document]
dbutenhof: Yes, I have considered trying to factor out commonality as pairs of "assemble an Elasticsearch payload" and "post-process the Elasticsearch response" factories.

While I've considered it several times, I've never yet taken time to actually do it, nor felt any great pressure to worry about it yet.

lib/pbench/server/api/resources/query_apis/query_results.py
```diff
@@ -0,0 +1,289 @@
...
+                        d["run.prefix"] = run["prefix"]
+                    if "@metadata" in src:
+                        meta = src["@metadata"]
+                        if "controller_dir" in meta:
+                            d["@metadata.controller_dir"] = meta["controller_dir"]
+                        if "satellite" in meta:
+                            d["@metadata.satellite"] = meta["satellite"]
+                    datasets.append(d)
+            except KeyError:
+                self.logger.exception("ES response not formatted as expected")
```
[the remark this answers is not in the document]
dbutenhof: True.

lib/pbench/server/db/database.py
```diff
@@ -0,0 +1,61 @@
...
+
+    @staticmethod
+    def get_engine_uri(config, logger):
+        try:
+            psql = config.get("Postgres", "db_uri")
+            return psql
+        except (NoSectionError, NoOptionError):
+            if logger:
+                logger.error("Failed to find [Postgres] db_uri in configuration file.")
+            sys.exit(1)
```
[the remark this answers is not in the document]
dbutenhof: I'm not entertaining comments on `database.py` as it's almost entirely stolen from Nikhil's #1937. I may consider further cleanups after I rebase on top of his merged code, but right now that's just more merge conflict pain for me.

lib/pbench/server/db/database.py
```diff
@@ -0,0 +1,61 @@
...
+    def init_db(server_config, logger):
+        # Attach the logger to the base class for models to find
+        if not hasattr(Database.Base, "logger"):
+            Database.Base.logger = logger
+
+        # IMPORTANT:
+        # Make sure all the models are imported before this function is called
+        # so that they will be registered properly on the metadata. Otherwise
+        # metadata will not have any tables and create_all functionality will
+        # do nothing.
```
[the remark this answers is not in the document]
dbutenhof: Ditto.

lib/pbench/server/db/database.py
```diff
@@ -0,0 +1,61 @@
...
+        try:
+            Database.Base.query = Database.db_session.query_property()
+            engine = Database.init_engine(server_config, logger)
+            Database.Base.metadata.create_all(bind=engine)
+            Database.db_session.configure(bind=engine)
```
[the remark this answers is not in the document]
dbutenhof: Double-ditto.

server/rpm/pbench-server.spec.j2
dbutenhof: And ... we definitely should do it, but I'd rather not add that basic change into one of these PRs. Someone (hint, hint) could do that independently, and the results could be merged into these ...

lib/pbench/server/api/__init__.py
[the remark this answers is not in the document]
webbnh: If I understand things correctly, the dashboard code is completely separate (e.g., it's in another repo) and it interacts with this code via a RESTful API which uses URIs completely unlike these class names...so I'm having trouble seeing the importance of matching the names with its identifiers.  Stacked against that is the potential pratfall of inadvertently adding or omitting that trailing letter and having nobody notice until late in the process....

lib/pbench/server/api/resources/query_apis/__init__.py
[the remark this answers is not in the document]
webbnh: I'm not a fan of using four lines of code where two will do, especially if the two-line approach moves the code closer to making it easier or more straightforward to format the index name externally.

lib/pbench/server/api/resources/query_apis/query_result.py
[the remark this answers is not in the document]
webbnh: I gave you a final nudge above...we'll see if you fall.  :-D

lib/pbench/server/api/resources/query_apis/query_result.py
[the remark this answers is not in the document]
webbnh: I certainly appreciate deferring the error-handling and diagnostic code until you need it (the phrase "fast path" leaps to mind for some reason... ;-) ), and I support reporting _all_ the errors instead of making the user come back to find the next one, but, knowing the way that code changes over time, I think it should be reporting the error that it actually found, here.  Now, if you want to report _additional_ errors, that's fine, but the assumption that, because we caught a `KeyError`, it must be one of the keys in `json_data` seems a bit of a stretch, especially as the code evolves in the future.

lib/pbench/server/api/resources/query_apis/query_result.py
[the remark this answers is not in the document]
webbnh: As long as there aren't multiple "kinds" of `HTTPError` (where we might want to handle one of them differently), I concur that it doesn't matter.

And, yes, in the absence of some sort of hierarchical relationship, I concur that having them in order (such as alphabetical) is a good thing.

lib/pbench/server/api/resources/query_apis/query_results.py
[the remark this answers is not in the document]
webbnh: Sorry -- I forgot where we were.  If we're doubling up values in the object being sent to JS, that's OK (presumably there won't be so much that it impacts network performance).

My concern was that we were doubling up values in the ES documents:  it doesn't hurt much there until you have millions and millions of documents, and then it starts to seriously impact the on-disk representation of the ES index.

lib/pbench/server/api/resources/query_apis/query_results.py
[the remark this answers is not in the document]
webbnh: I would agree with you, except for the fact that on occasion I've been the one doing the stupid something, and it's awfully nice if the error message makes it obvious what you've done.  (If we can avoid having to run a debugger, it's a total win.)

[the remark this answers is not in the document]
webbnh: (Imagine that it's you running Postman, and you've lost a closing brace -- you don't want to be told that the payload is missing, or even invalid -- you want to be told that the problem is in line `N`, column `C` of the payload....)

lib/pbench/server/api/resources/query_apis/query_results.py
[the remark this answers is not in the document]
webbnh: Well...you probably now have _three_ copies of the hard-coded ES URI fragment...it sure would be nice to have at least _that_ in one place.

lib/pbench/server/db/database.py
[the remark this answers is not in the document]
webbnh: Are you saying that I should be giving input on #1937?

server/rpm/pbench-server.spec.j2
webbnh: I was just wondering whether, instead of adding two files here, you could just add their directory?  (Less change and all that.... ;-) )

lib/pbench/server/db/database.py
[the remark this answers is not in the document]
dbutenhof: > Are you saying that I should be giving input on #1937?

Well, yes; although at this point I'd almost rather have it here or on #2074, get Nikhil's changes in quickly, and decide how to clean it up later. I'm just not going to _do_ anything about them here or in #2074 until after I've rebased on top of #1937.

server/bin/pbench-import-datasets.py
```diff
@@ -0,0 +1,185 @@
...
+"""Command line interface to the Dataset state mechanism.
+
+This serves two purposes:
+
+1. It allows access to the state tracker for bash scripts;
+2. It's used to pre-set database state for the `gold` unit tests
+
+Therefore, while this can eventually be removed, we need to resolve both
+of those requirements first.
+"""
```
dbutenhof, line 13: Ugh; forgot to update this when I copied the state manager. TBD.

portante (changes requested):
> Add "user" ownership to Pbench Elasticsearch documents.

Seems like you are now using `owner`.  Perhaps you could update the PR description and the commit message in the individual commits?

> The default is user "public" and access "public" ... the owner is inherited from the Dataset (note that this PR layers on top of #2074), and a non-default user will have "private" access by default. (This is shown in several of the unit test gold files, via a new optional argument in the '.db' state files.)

Can we consider having this PR make a change to the assuming that there is a `public` user?  I think we don't really want the notion of a default user, right?  All data should be owned by a real user.

More to come ...

lib/pbench/server/api/resources/query_apis/query_controller_list.py
```diff
@@ -16,52 +21,10 @@ def __init__(self, config, logger):
...
-
     def post(self):
         """
-        POST for Pbench controller datasets with start times within the
-        specified search range, and owned by a specified user:
+        POST for Pbench controllers with datasets within the specified date
+        range, and owned by a specified user:
 
         {
             "user": "email-tag",
```
portante, line 30: Shouldn't this parameter be hidden now for this API?

The current use in the dashboard is to view all controllers.  So we want to query for all "public" data and all data from the current user if there is one.

lib/pbench/server/api/resources/query_apis/query_dataset.py
```diff
@@ -0,0 +1,268 @@
...
+        self.es_url = get_es_url(config)
+        self.prefix = get_index_prefix(config)
+
+    def post(self):
+        """
+        POST to get details for a specific Pbench dataset within the
+        specified date range, and owned by a specified user:
+
+        {
+            "user": "email-tag",
```
portante, line 45: Why are we requiring that the API invocation provide a "user" argument?

Why not make this parameter optional to add a constraint on which data is visible.  But we should not use that parameter for access control.

Instead, we have to get the effective user of the query, and make sure the query results in data that is "public" or from a user that matches this user.

lib/pbench/server/api/resources/query_apis/query_dataset_list.py
```diff
@@ -0,0 +1,290 @@
...
+        self.es_url = get_es_url(config)
+        self.prefix = get_index_prefix(config)
+
+    def post(self):
+        """
+        POST for Pbench datasets from a specified controller within the
+        specified date range, and owned by a specified user:
+
+        {
+            "user": "email-tag",
```
portante, line 39: Why are we requiring that the API invocation provide a "user" argument?

Why not make this parameter optional to add a constraint on which data is visible.  But we should not use that parameter for access control.

Instead, we have to get the effective user of the query, and make sure the query results in data that is "public" or from a user that matches this user.

portante (changes requested):
Can you provide examples of where and how the dashboard would use these new APIs?

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -2,9 +2,28 @@
 Helper functions to get the elasticsearch specific configurations
 """
 
+from dateutil import rrule
+
 from configparser import NoSectionError, NoOptionError
 
 
+def get_user_term(user: str):
```
npalaska, line 10: Is the user_term a username?

lib/pbench/server/api/resources/query_apis/query_controller_list.py
```diff
@@ -107,9 +72,14 @@ def post(self):
...
-        without a session token or user parameter. We need to define precisely
-        how this should work.
+        TODO: Do we automatically return all "public" controllers, or should
+        this be a separate parameter? E.g., when "I" log in, do I see all my
+        own datasets *plus* all published/public controllers owned by others,
+        or should that be a separate option (e.g., a checkbox on the view?)
+
+        TODO: Similarly, a query from an unlogged-in session in the dashboard
+        would presumably show datasets with "user": "public"; but does it show
+        all datasets with "access": "public" as well?
```
npalaska, line 82: Wouldn't all the dataset with "public" user have "access" public as well? how are they 2 different things?

lib/pbench/server/api/resources/query_apis/query_dataset.py
```diff
@@ -0,0 +1,268 @@
...
+        qualified by the index_prefix of the Pbench server that originally
+        indexed the data. The prefix comes from the pbench-server.cfg
+        file for the server instance. For example, production, "staging",
+        and development server instances might all share one Elasticsearch
+        cluster by choosing unique prefixes to generate unique index
+        names.
+
+        Required headers include
+
+        X-auth-token:   Pbench session authorization token authorized to access
```
npalaska, line 83: This should be bearer `Authorization`

lib/pbench/server/api/resources/query_apis/query_dataset.py
npalaska, line 45: Interesting, this could be a username as well right?

lib/pbench/server/api/resources/query_apis/query_dataset.py
```diff
@@ -0,0 +1,268 @@
...
+            }
+        ]
+        """
+        json_data = request.get_json(silent=True)
+        if not json_data:
+            self.logger.info("Invalid JSON object. Query: {}", request.url)
+            abort(400, message="Missing request payload")
+
+        try:
+            user = json_data["user"]
```
npalaska, line 130: If we have an authorization header included we could deduce the username and email field from it and a request json payload does not necessarily need user information.

lib/pbench/server/api/resources/query_apis/query_dataset_list.py
```diff
@@ -0,0 +1,290 @@
...
+        qualified by the index_prefix of the Pbench server that originally
+        indexed the data. The prefix comes from the pbench-server.cfg
+        file for the server instance. For example, production, "staging",
+        and development server instances might all share one Elasticsearch
+        cluster by choosing unique prefixes to generate unique index
+        names.
+
+        Required headers include
+
+        X-auth-token:   Pbench session authorization token authorized to access
```
npalaska, line 78: `Authorization` ?

npalaska (commented):
Looks good, just couple of questions

lib/pbench/server/api/resources/query_apis/query_dataset.py
dbutenhof: The comment is perhaps stale, although I'm not sure we've entirely resolved the issue.

The dashboard has never had a "username", only "email". You coded our identity manager with the more conventional username + email style, but I'd written this comment very early on and followed the email convention used in the dashboard.

We really ought to discuss exactly what we're going to do; what user metadata we maintain along with the primary access key, conventions for updating email, etc. For now, yeah, this is whatever token is used to uniquely reference a specific user ... or `None` for public data.

dbutenhof: I'm still using the `json_data["user"]` reference here, but I really should change it to `json_data.get("user")` so that a missing `"user"` will simply return as `None` and trigger the query for published data just like an explicit `"user": None`.

And, yes, I could also default it from the authorization token payload, but this still hasn't been hooked up to the authorization decorator. My thought had been to do that in a separate pass.

There's still the basic question of whether we want a logged-in user to be able to see only _their own_ owned data, rather than also all public data. The `"user"` parameter allows disambiguation, and relying on the authorization token I think implies we need a specific parameter if we want to do that. E.g., if I don't specify `"user"` and I'm authorized, do I see just my data, or also all public data? What about an authorized user specifying `"user": None`? Do we default, or take the specification of "anti-user" (public).

lib/pbench/server/api/resources/query_apis/query_dataset.py
dbutenhof: Yep, another stale comment. I want to refactor all this to reduce duplication; I've been struggling to put that off until I'm done here.

lib/pbench/server/api/resources/query_apis/query_dataset.py
dbutenhof: Yes, we can deduce from the token, but I didn't have that when I wrote this, and we still haven't wired in your authorization decorators. This was partly a way to move forward without all that; but we also need to deal with whether an authorized user sees just their own data or/also public data, and `"user"` provides a hook to override the default. I also wanted to open the possibility for an admin user to see any user's data.

server/lib/mappings/result-data-sample.json
```diff
@@ -15,6 +15,16 @@
         "@generated-by": {
             "type": "keyword"
         },
+        "authorization": {
```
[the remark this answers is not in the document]
dbutenhof: We somewhat decrease any documentation value of the files if we supplement them too much when we read them, though that certainly would reduce the redundancy. But, yes, the `indexer.py` template reader could certainly do that.

lib/pbench/server/api/resources/query_apis/query_dataset.py
```diff
@@ -0,0 +1,268 @@
...
+            abort(500, message="INTERNAL ERROR")
+        else:
+            run_metadata = {}
+            try:
+                es_json = es_response.json()
+                hits = es_json["hits"]["hits"]
+
+                # NOTE: we're expecting just one. We're matching by just the
+                # dataset name, which ought to be unique.
+                self.logger.info("{} datasets found", len(hits))
```
webbnh, line 251: Since `len(hits)` should always be 1, this seems like a really boring log message...I suggest only logging if the length is _not_ 1 (and use a warning or something more severe).

lib/pbench/server/api/resources/query_apis/query_dataset_list.py
webbnh: And "email-tag" should become "username" or similar.

lib/pbench/server/api/resources/query_apis/query_dataset_list.py
```diff
@@ -0,0 +1,290 @@
...
+                "run.controller": "dhcp31-187.perf.lab.eng.bos.redhat.com",
+                "run.start": "2020-04-29T12:49:13.560620",
+                "run.end": "2020-04-29T13:30:04.918704"
+            }
+        ]
+        """
+        json_data = request.get_json(silent=True)
+        if not json_data:
+            self.logger.info("Invalid JSON object. Query: {}", request.url)
+            abort(400, message="Missing request payload")
```
webbnh, line 112: "Missing or bad..."

lib/pbench/server/api/resources/query_apis/query_dataset_list.py
```diff
@@ -0,0 +1,290 @@
...
+            self.logger.info(
+                "Invalid start or end time string: {}, {}: {}", start_arg, end_arg, e
+            )
+            abort(400, message="Invalid start or end time string")
+
+        # We have nothing to authorize against yet, but print the specified
+        # user and session token to prove we got them (and so the variables
+        # aren't diagnosed as "unused")
+        self.logger.info(
+            "QueryControllers POST for user {}, prefix {}, session {}: ({}: {} - {})",
```
webbnh, line 159: "QueryControllers" ==> "QueryDatasetList"

lib/pbench/server/api/resources/query_apis/query_dataset_list.py
```diff
@@ -0,0 +1,290 @@
...
+            abort(504, message="Connection timed out, could not post to Elasticsearch")
+        except requests.exceptions.InvalidURL:
+            self.logger.exception(
+                "Invalid url {} during the Elasticsearch post request", uri
+            )
+            abort(500, message="INTERNAL ERROR")
+        except Exception as e:
+            self.logger.exception(
+                "Exception {!r} occurred during the Elasticsearch post request", e
+            )
```
webbnh, line 242: Including the exception object in the message seems redundant.

lib/pbench/server/api/resources/query_apis/query_controller_list.py
portante: Seems like we really need clarity on the use case from the Dashboard.

lib/pbench/server/api/resources/query_apis/query_dataset_list.py
```diff
@@ -0,0 +1,290 @@
...
+                    "run.controller",
+                    "run.start",
+                    "run.end",
+                    "run.name",
+                    "run.config",
+                    "run.prefix",
+                    "run.id",
+                ]
+            },
+            "sort": {"run.end": {"order": "desc"}},
```
webbnh, line 182: Why do we sort by ending timestamp when we add a field to the results (line 273) with the _beginning_ timestamp?

lib/pbench/server/api/resources/query_apis/query_dataset_list.py
```diff
@@ -0,0 +1,290 @@
...
+                        timestamp = parser.parse(run["start"]).utcfromtimestamp()
+                    except Exception as e:
+                        self.logger.info(
+                            "Can't parse start time {} to integer timestamp: {}",
+                            run["start"],
+                            e,
+                        )
+                        # fall back to the end timestamp sort key (why?)
+                        # TODO: Should this just abort instead?
+                        timestamp = dataset["sort"][0]
```
webbnh, line 271: I imagine we've selected the sort key as a fallback because we know it's "good"...if it weren't then ES would have rejected the document at ingestion...right??

However, if the start time is bad, why should we "abort"?  (And, what would that mean?...drop this data set, fail the whole request, or what?)

lib/pbench/server/database/models/tracker.py
```diff
@@ -300,16 +300,14 @@ class Dataset(Database.Base):
...
     id = Column(Integer, primary_key=True, autoincrement=True)
-    user = Column(String(255), unique=False, nullable=False, default=ANTI_USER)
+    owner = Column(String(255), unique=False, nullable=True, default=None)
     controller = Column(String(255), unique=False, nullable=False)
     name = Column(String(255), unique=False, nullable=False)
 
     # FIXME:
-    # Ideally, `md5` would not be `nullable`: this allows the
-    # pbench-server-prep-shim-002 utility to construct a Dataset object
+    # Ideally, `md5` would not be `nullable`: but allowing it means that
```
webbnh, line 309: s/:/,/

lib/pbench/server/indexing_tarballs.py
```diff
@@ -345,6 +340,18 @@ def sighup_handler(*args):
                                 idxctx.logger.warn(
                                     "Unable to advance dataset state: {}", str(e)
                                 )
+                            else:
+                                username = dataset.owner
```
webbnh, line 344: Does the code after this point do the right thing if there is an exception raised by `Dataset.attach()`?  I imagine that `username` reads as `None` (which is probably OK), but what does it mean for `dataset` to be undefined?

lib/pbench/server/indexing_tarballs.py
```diff
@@ -319,15 +319,10 @@ def sighup_handler(*args):
                         idxctx.logger.info("Starting {} (size {:d})", tb, size)
                         dataset = None
```
webbnh, line 320: What's the scope of `dataset`?  It appears to be assigned at line 327 and read at line 344...if that's correct, then we don't need this initialization here, especially if we move the read from the `else` to the inside of `try` block.

webbnh (changes requested):
Most of my concerns are fixes to comment or diagnostics text.

I do think it would be a good idea to encapsulate the ES schema version (part of the index prefix) in a common definition and reference it symbolically.

server/lib/mappings/result-data-sample.json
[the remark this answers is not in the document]
dbutenhof: I thought about this again and got tempted; although it was a bit less clean than I'd imagined, it's done.

lib/pbench/server/api/resources/query_apis/datasets_detail.py
```diff
@@ -0,0 +1,266 @@
...
+        "host_tools_info" nested array.
+
+        TODO: We probably need to return all *unowned* runs when called
+        without a session token or user parameter. We need to define precisely
+        how this should work.
+
+        [
+            {
+                "runMetadata": {
+                    "file-name": "/pbench/archive/fs-version-001/dhcp31-187.perf.lab.eng.bos.redhat.com/fio_rhel8_kvm_perf43_preallocfull_nvme_run4_iothread_isolcpus_2020.04.29T12.49.13.tar.xz",
```
portante, line 103: Not sure we really want to expose internal to Red Hat names.

lib/pbench/server/api/resources/query_apis/datasets_detail.py
```diff
@@ -0,0 +1,266 @@
...
+    get_user_term,
+)
+
+
+class DatasetsDetail(Resource):
+    """
+    Abstracted Pbench API to get detailed data for a particular
+    dataset by name.
+
+    This API a follow-on to datasets_list, which returns the names of datasets
```
portante, line 22: `s/This API a follow-on/This API is a follow-on/`

lib/pbench/server/api/resources/query_apis/datasets_detail.py
```diff
@@ -0,0 +1,266 @@
...
+from pbench.server import PbenchServerConfig
+from pbench.server.api.resources.query_apis import (
+    get_es_url,
+    get_index_prefix,
+    gen_month_range,
+    get_user_term,
+)
+
+
+class DatasetsDetail(Resource):
```
portante, line 17: Bummer, the state tracking uses the name dataset, a tar ball with metadata.  It is unfortunate that we are introducing a new name.  Could we consider renaming `Dataset` to `Tarball`?  Big can of worms, but I really feel "dataset" is too abstract.  It is really an opaque object that is known to be a tar ball, with metadata about it.

lib/pbench/server/api/resources/query_apis/datasets_detail.py
```diff
@@ -0,0 +1,266 @@
...
+        assuming the session user is authorized to view "user"s data. This
+        will be used to constrain the Elasticsearch query.
+
+        TODO: When we have authorization infrastructure, we'll need to
+        check that "session user" has rights to view "user" data. We might
+        also default a missing "user" JSON field with the authorization
+        token's user. (Or with the default user "public" if there's no
+        token, indicating there's no logged-in user.)
+
+        "name" is the name of a Pbench agent dataset (tarball).
```
portante, line 66: So a dataset is a tar ball?  Why introduce a new name?

lib/pbench/server/api/resources/query_apis/datasets_detail.py
```diff
@@ -0,0 +1,266 @@
...
+        qualified by the index_prefix of the Pbench server that originally
+        indexed the data. The prefix comes from the pbench-server.cfg
+        file for the server instance. For example, production, "staging",
+        and development server instances might all share one Elasticsearch
+        cluster by choosing unique prefixes to generate unique index
+        names.
+
+        Required headers include
+
+        Authorization:  Pbench authorization token authorized to access
```
portante, line 84: Perhaps word this as: `Pbench authorization token granting the caller access to the "user"'s data.`

lib/pbench/server/api/resources/query_apis/datasets_detail.py
```diff
@@ -0,0 +1,266 @@
...
+
+        NOTE: This is the format currently constructed by the Pbench
+        dashboard `src/model/dashboard.js` fetchResults effect. The
+        "runMetadata" property is a union of the run document's "@metadata
+        and "run" sub-documents, while "hostTools" is the run document's
+        "host_tools_info" nested array.
+
+        TODO: We probably need to return all *unowned* runs when called
+        without a session token or user parameter. We need to define precisely
+        how this should work.
```
portante, line 98: It really seems we want to make all tar balls owned by somebody explicitly.

lib/pbench/server/api/resources/query_apis/datasets_detail.py
```diff
@@ -0,0 +1,266 @@
...
+            self.logger.info("Invalid JSON object. Query: {}", request.url)
+            abort(400, message="Invalid request payload")
+
+        try:
+            user = json_data["user"]
+            name = json_data["name"]
+            start_arg = json_data["start"]
+            end_arg = json_data["end"]
+        except KeyError:
+            keys = [k for k in ("user", "name", "start", "end") if k not in json_data]
```
portante, line 131: Since we are going through the effort to generate the list of keys as a list, only to immediately format them as a string, perhaps we can do that once and just reference the string?

lib/pbench/server/api/resources/query_apis/datasets_detail.py
```diff
@@ -0,0 +1,266 @@
...
+        # often.
+        #
+        # So ... do we need to change the code here each time? The old
+        # dashboard config is hardcoded for each version, so one option is
+        # to add a version for each index to the pbench-server.cfg file.
+        #
+        # When we have a full live Python server, persisting the template
+        # information in Redis or postgreSQL probably makes sense; maybe
+        # we just hardcode until then?
+        uri_fragment = gen_month_range(self.prefix, ".v6.run-data.", start, end)
```
portante, line 200: Are we now telling the dashboard these strings via the config APIs?  Seems like we don't want to hard code this.

lib/pbench/server/api/resources/query_apis/datasets_detail.py
```diff
@@ -0,0 +1,266 @@
...
+
+        uri = f"{self.es_url}/{uri_fragment}/_search"
+        try:
+            # query Elasticsearch
+            es_response = requests.post(
+                uri,
+                params={"ignore_unavailable": "true"},
```
portante, line 207: Isn't the `params` argument a string?  This is a bug that I think @anishaswain hit.

lib/pbench/server/indexing_tarballs.py
portante: This then begs the question of supporting no-owner tar balls.  Is this something we want to continue to support with this commit?  I don't remember what was decided about that.

lib/pbench/server/api/resources/query_apis/datasets_list.py
```diff
@@ -0,0 +1,296 @@
...
+        qualified by the index_prefix of the Pbench server that originally
+        indexed the data. The prefix comes from the pbench-server.cfg
+        file for the server instance. For example, production, "staging",
+        and development server instances might all share one Elasticsearch
+        cluster by choosing unique prefixes to generate unique index
+        names.
+
+        Required headers include
+
+        Authorization:  Pbench session authorization token authorized to access
```
portante, line 86: See comment in `datasets_detail.py` of the same...

lib/pbench/server/api/resources/query_apis/datasets_list.py
```diff
@@ -0,0 +1,296 @@
...
+        is the same as the "run.name"; the requirement for "key" is placed by
+        Redux/Javascript, and is inherited here as we're currently replacing
+        the dashboard queries as they were.)
+
+        [
+            {
+                "key": "fio_rhel8_kvm_perf43_preallocfull_nvme_run4_iothread_isolcpus_2020.04.29T12.49.13",
+                "startUnixTimestamp": 1588178953561,
+                "run.name": "fio_rhel8_kvm_perf43_preallocfull_nvme_run4_iothread_isolcpus_2020.04.29T12.49.13",
+                "run.controller": "dhcp31-187.perf.lab.eng.bos.redhat.com",
```
portante, line 112: I don't think we want to expose internal to Red Hat names.

lib/pbench/server/api/resources/query_apis/datasets_list.py
```diff
@@ -0,0 +1,296 @@
...
+            self.logger.info("Invalid JSON object. Query: {}", request.url)
+            abort(400, message="Invalid request payload")
+
+        try:
+            user = json_data["user"]
+            controller = json_data["controller"]
+            start_arg = json_data["start"]
+            end_arg = json_data["end"]
+        except KeyError:
+            keys = [
```
portante, line 129: Similar comment to the one is `datasets_detail.py` ...

lib/pbench/server/api/resources/query_apis/datasets_list.py
```diff
@@ -0,0 +1,296 @@
...
+        # often.
+        #
+        # So ... do we need to change the code here each time? The old
+        # dashboard config is hardcoded for each version, so one option is
+        # to add a version for each index to the pbench-server.cfg file.
+        #
+        # When we have a full live Python server, persisting the template
+        # information in Redis or postgreSQL probably makes sense; maybe
+        # we just hardcode until then?
+        uri_fragment = gen_month_range(self.prefix, ".v6.run-data.", start, end)
```
portante, line 214: Similar comment to `datasets_detail.py` ...

portante (changes requested):
Coming along!

server/bin/pbench-state-manager.py
```diff
@@ -108,6 +111,9 @@ def main(options):
         dest="path",
         help="Specify a tarball filename (from which controller and name will be derived)",
     )
+    parser.add_argument(
+        "--user", dest="user", help="Specify the owning username for the dataset",
```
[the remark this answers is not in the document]
dbutenhof: Yes, I've allowed myself to be overly influenced by the backpressure of our CLI unit test system. I might as well just drop the concept of unowned datasets.

In fact, now that Nikhil's changes are actually in, I really should _validate_ the user parameter against the Users table instead of just taking it blindly. Ha; I hadn't even considered that until now.

dbutenhof (commented):
So, next commit will be making the username mandatory and fixing a few comments...

lib/pbench/server/indexing_tarballs.py
dbutenhof: I thought I'd answered this before, but apparently not. The initialization to `None` (which is now at line 320) ensures that the reference in the `finally` way down at line 428 is safe regardless of which exception paths we might take. That's where I advance the dataset state to `INDEXED` on success or `QUARANTINED` on failure, assuming we found the dataset in the first place. (And of course we always should.)

server/bin/unittests
```diff
@@ -279,10 +279,12 @@ function _setup_state {
     let res=res+$?
     mkdir -p $_testopt_sat/bin
     let res=res+$?
-    mkdir -p $_testopt/common/lib
+    mkdir -p $_testopt/common/
     let res=res+$?
     cp -a $_tdir/{unittests,pbench*} $_testopt/bin
     let res=res+$?
+    cp -a ${_tdir}/../../lib ${_testopt}/common/
```
webbnh, line 286: Rather than going "up" from `_tdir`, why not come "down" from `_gitdir`?  (And, for the record, I think there is a better way to establish `_gitdir` than what is used at lines 16-17 -- `git rev-parse --show-toplevel`.)

server/lib/mappings/result-data-sample.json
```diff
@@ -15,16 +15,6 @@
...
-        "authorization": {
-            "properties": {
-                "owner": {
-                    "type": "keyword"
-                },
-                "access": {
-                    "type": "keyword"
-                }
-            }
-        },
```
webbnh: Why is it good/better to add these programmatically, after the fact than to do it statically in the various files?  And, why doesn't that logic apply to the other mappings, et al. defined in these files?

webbnh (dismissed):
Looks good to me...I just have a few questions, etc.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -117,7 +117,9 @@ def put(self, controller):
 
         # Create a tracking dataset object; it'll begin in UPLOADING state
         try:
-            dataset = Dataset(controller=controller, path=tar_full_path, md5=md5sum)
+            dataset = Dataset(
+                owner="pbench", controller=controller, path=tar_full_path, md5=md5sum
+            )
```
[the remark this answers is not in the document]
dbutenhof: I think that really just obscures the temporary hack... however there should at least be a clear `FIXME` comment, and I guess I rushed through this door too quickly.

server/bin/pbench-base.sh
```diff
@@ -137,7 +137,9 @@ function quarantine () {
         fi
         # Create/update state record if we're not moving a .md5 file
         if [ ${afile} != ${afile%.tar.xz} ] ;then
-            pbench-state-manager.py -c --path="${afile}" --state=quarantined
+            # TODO: When we drop ssh intake, we'll always have a user
+            # by now.
+            pbench-state-manager --create=quarantine --path="${afile}" --state=quarantined
```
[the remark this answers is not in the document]
dbutenhof: Ha; true, that was probably not the best possible example.

server/bin/pbench-state-manager.py
```diff
@@ -50,14 +51,37 @@ def main(options):
...
+                # FIXME: I don't want to be creating the user here or
+                # dealing with a non-existing user. The unittest setup
+                # should create the test users we want ahead of time
+                # using a pbench-user-manager command and we should
+                # depend on having them here! The following is a hack
+                # until that command exists.
+                #
+                # The desired behavior would be to remove this try and
+                # except and allow UnknownUser to be handled below with
+                # an error message and termination.
```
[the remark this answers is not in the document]
dbutenhof: Normally I would want the `UnknownUser` to terminate with a message. Until I have a separate utility to create users, I wanted to create the `User` here, which implies catching the exception and then acting. There'd be no point in a `try` block "below" as we don't want to handle other exceptions (e.g., if something goes wrong with creating the `User` or `Dataset`) because we want them to terminate with a message.

Maybe I'm not sure what you're getting at?

server/bin/pbench-state-manager.py
```diff
@@ -80,7 +104,9 @@ def main(options):
         doit = Dataset.create if options.create else Dataset.attach
 
         # Find the specified dataset, and update the state
-        doit(**args)
+        attached = doit(**args)
+        if not options.create and not attached:
+            print(f"{_NAME_}: dataset")
```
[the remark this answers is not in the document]
dbutenhof: That's actually weird, and I have no idea where that came from. Certainly wasn't me. ;-)

Presumably the idea was that attach might return `None`; but while it used to, it no longer does. This may be left over from a very early version of this change, and I just never paid any attention to it before. I'll remove it!

server/bin/unittests
dbutenhof: Mostly because I copied and pasted without trying to trace through the whole hierarchy of environment variables that the script constructs. I'm also not entirely sure this is even necessary; I added it when my tests initially had trouble finding the DB models, after tearing out some hair I can't really afford to lose... but then I got smarter and blew away the tox virtual environment and forced it to reload, and that worked. I'd actually meant to try removing this to see if it still worked, but I forgot. (Though logically this copy is consistent with the general setup strategy and might be worth doing anyway.)

server/lib/mappings/result-data-sample.json
dbutenhof: The idea (proposed by Pete in an earlier review cycle) is that this is common to all our Elasticsearch documents; e.g., if we support customized document format plugins we'll still want these to support our access control. Adding them after the fact guarantees they're consistent across all documents, and never omitted.

There really isn't anything else that's common across them all, although a few items are common across several, or similar. This seemed like a good convention to me.

lib/pbench/server/api/auth.py
```diff
@@ -15,6 +27,44 @@ def set_logger(logger):
...
+        have a stable "userid" field that we use for indexing... in which case
+        this should translate "username" into "userid" for internal use. For
+        now, just return the original username if it was found.
+
+        FIXME: I thought this would be an ArgumentParser "type" so we could
+        do validation during parsing. That's awkward because we need to finish
+        parsing the --config parameter in order to access the config and logger
+        objects required to initialize database access, which needs to be done
+        before we can make a query. Can we work around this without too much
+        mess?
```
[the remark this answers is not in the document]
webbnh: If you do the Click thing, then what I think you're looking for is [the "eager" command option](https://click.palletsprojects.com/en/7.x/options/#callbacks-and-eager-options).

lib/pbench/server/api/auth.py
```diff
@@ -15,6 +27,44 @@ def set_logger(logger):
...
+
+        Args:
+            :name: The username field of a registered user
+
+        Raises:
+            ValueError: The username doesn't correspond to a registered user
+            TypeError: Some other error occurred looking for the user
+
+        Returns:
+            The specified username if it's valid; does not return on failure
```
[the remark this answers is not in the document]
webbnh: Oh, it definitely makes sense, but talking about not returning under the "Returns" heading puts you into the null solution space remarkably quickly.  ;-)

lib/pbench/server/api/resources/query_apis/datasets_detail.py
```diff
@@ -11,45 +14,46 @@
...
-    directly, but rather allows building a set of Elasticsearch index
-    names that will contain the data. (Note that the dashboard uses the
-    same date range it used to find controllers; however it could make this
-    more efficient by using only the year and month in which the result
-    was generated.)
+    dataset by name.
+
+    Note that datasets_list returns the names of datasets for a controller
+    and then datasets_detail returns details from a specific dataset's run
+    document.
```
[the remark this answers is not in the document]
webbnh: I'm not sure how to find a module -- I know how to find a _file_, but, for Python, those end in `.py`....  ;-)  And, as for finding classes, my IDE knows how to do that at the click of a button!

lib/pbench/server/api/resources/upload_api.py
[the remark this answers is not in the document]
webbnh: My idea was that the changes made to line now 121 would be "permanent" and the "temporary hack" would be positioned elsewhere, where, at the appropriate moment, it would be replaced wholesale.

server/bin/pbench-state-manager.py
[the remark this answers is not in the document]
webbnh: It's pretty trivial.  I would propose the same code, but different sequencing.  I wouldn't put the temporary creation of the `User` embedded in the argument handling code; rather, I would place it later in the function, where it can be set off better from the surrounding structures and more easily nuked at the appropriate time in the future.

There's no reason why what you have isn't workable, it's just that the same construct is embedded a few levels deeper and it interrupts the readability of the surrounding structure.  But, on the assumption that this will all go away in a few weeks, it probably doesn't matter.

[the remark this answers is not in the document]
dbutenhof: I'm inclined to keep this as it is; if I deferred the `User` creation I'd have to remember that the validation (which I do want to be part of the "argument handling") failed so I can use that later, which would be a broader effect on the logic.

lib/pbench/server/api/resources/query_apis/month_indices.py
```diff
@@ -27,23 +27,13 @@ def __init__(self, config: PbenchServerConfig, logger: Logger):
 
     def get(self):
         """
-        Report the month suffixes existing for run data indices.
+        Report the month suffixes for run data indices.
 
-        Required headers include
+        NOTE: No authorization or input parameters are required for this API.
```
webbnh, line 32: This feels odd.  It seems like an information leak.

For instance, it's conceivable that someone could reach unwarranted conclusions about Red Hat business by observing when we produce performance results and when we don't.

I don't know whether we should actually worry about this, since the resolution of the results is so low (a single benchmark result would be enough to cause its production month to be returned in the list) and, once we support a significant number of clients, there will never be a month when there are _no_ results, at which point there probably won't be anything to be gained from the information returned by this request (although, at that point, perhaps this request will no longer even be necessary...).

webbnh (approved):
Just a couple of small items which can probably be deferred or ignored.

lib/pbench/server/api/resources/query_apis/controllers_list.py
```diff
@@ -33,53 +33,38 @@ def __init__(self, config: PbenchServerConfig, logger: Logger):
 
     def post(self):
         """
-        POST for Pbench controllers with datasets within the specified date
-        range, and owned by a specified user:
+        POST to search for Pbench controller names which have registered
+        datasets within a specified date range and which are either owned
+        by a specified username, or have been made publicly accessible.
```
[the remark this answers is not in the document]
dbutenhof: `get_user_term` will generate a search for `"access": "public"` rather than `"owner": "<user>"` if the username is `None`.

lib/pbench/server/api/resources/query_apis/month_indices.py
dbutenhof: If we're going to support the normal starting dashboard page without login, this needs to work because it's how the dashboard sets up the date picker. The alternative is to allow no access at all to a user who's not logged in. (Something I'd personally prefer... but it would take a lot more work on the dashboard side to _only_ support our currently hypothetical share URIs for sessions that aren't logged in.)

webbnh: > If we're going to support the normal starting dashboard page without login, this needs to work

Fair enough.

> The alternative is to allow no access at all to a user who's not logged in. (Something I'd personally prefer... [...])

I'm not opposed to that, but I understand that "we" think that it would be a Good Thing™ not to have that hurdle for viewers.

> it would take a lot more work on the dashboard side to only support our currently hypothetical share URIs for sessions that aren't logged in.

I'm _hoping_ that the hypothetical share URIs are basically orthogonal to the other issue.  That is, I expect that the share URIs are basically a content manifest; whether the requestor can access that content is a separate question.  (And (I think) that question gets answered on the server side, not the client side -- the client just makes the request(s) and displays what the server returns, which might be a combination of content and access errors -- and this is true regardless of whether the user is logged in or not...but, maybe I'm missing something.)

dbutenhof: > I expect that the share URIs are basically a content manifest; whether the requestor can access that content is a separate question. (And (I think) that question gets answered on the server side, not the client side -- the client just makes the request(s) and displays what the server returns, which might be a combination of content and access errors -- and this is true regardless of whether the user is logged in or not...but, maybe I'm missing something.)

That's how I see it, too. We'll have to see how to get there from where we are.

riya-17 (approved):
Looks Good

PR #2114, opened 2021-02-17
Add SIGHUP handler to pbench indexer

### LINKED ISSUE #2060: Add SIGHUP signal handler to the pbench indexer
SIGHUP: re-evaluate the list of tar balls to index after indexing of the current tar ball has finished

- Report the current state of indexing
            - SIGHUP
            - Current tar ball being processed
            - How many remaining
            - How many have been completed
            - Errors encountered
            - Handler Behavior:
            1. Kicks a special thread to log indexing state data
            2. No exception raised

riya-17:
Fixes #2060

Re-evaluate the list of tar balls to index after
indexing of the current tar ball has finished.

Report the current state of indexing
- SIGHUP
- Current tar ball being processed
- How many remaining
- How many have been completed

### DIFF
--- a/lib/pbench/server/indexing_tarballs.py
+++ b/lib/pbench/server/indexing_tarballs.py
@@ -50,7 +50,7 @@ def _count_lines(fname):
 class Index:
     """class used to collect tarballs and index them"""
 
-    def __init__(self, name, options, idxctx, incoming):
+    def __init__(self, name, options, idxctx, incoming, archive, qdir):
 
         self.options = options
         _re_idx = "RE-" if options.re_index else ""
@@ -70,34 +70,36 @@ class Index:
         self.idxctx = idxctx
         self.incoming = incoming
         self.name = name
+        self.archive = archive
+        self.qdir = qdir
 
-    def collect_tb(self, ARCHIVE_rp, qdir):
+    def collect_tb(self):
         """ Collect tarballs that needs indexing"""
 
         # find -L $ARCHIVE/*/$linksrc -name '*.tar.xz' -printf "%s\t%p\n" 2>/dev/null | sort -n > $list
         tarballs = []
         idxctx = self.idxctx
         try:
-            tb_glob = os.path.join(ARCHIVE_rp, "*", self.linksrc, "*.tar.xz")
+            tb_glob = os.path.join(self.archive, "*", self.linksrc, "*.tar.xz")
             for tb in glob.iglob(tb_glob):
                 try:
                     rp = Path(tb).resolve(strict=True)
                 except OSError:
                     idxctx.logger.warning("{} does not resolve to a real path", tb)
-                    quarantine(qdir, idxctx.logger, tb)
+                    quarantine(self.qdir, idxctx.logger, tb)
                     continue
                 controller_path = rp.parent
                 controller = controller_path.name
                 archive_path = controller_path.parent
-                if str(archive_path) != str(ARCHIVE_rp):
+                if str(archive_path) != str(self.archive):
                     idxctx.logger.warning(
-                        "For tar ball {}, original home is not {}", tb, ARCHIVE_rp
+                        "For tar ball {}, original home is not {}", tb, self.archive
                     )
-                    quarantine(qdir, idxctx.logger, tb)
+                    quarantine(self.qdir, idxctx.logger, tb)
                     continue
                 if not Path(f"{rp}.md5").is_file():
                     idxctx.logger.warning("Missing .md5 file for {}", tb)
-                    quarantine(qdir, idxctx.logger, tb)
+                    quarantine(self.qdir, idxctx.logger, tb)
                     # Audit should pick up missing .md5 file in ARCHIVE directory.
                     continue
                 try:
@@ -105,7 +107,7 @@ class Index:
                     size = rp.stat().st_size
                 except OSError:
                     idxctx.logger.warning("Could not fetch tar ball size for {}", tb)
-                    quarantine(qdir, idxctx.logger, tb)
+                    quarantine(self.qdir, idxctx.logger, tb)
                     # Audit should pick up missing .md5 file in ARCHIVE directory.
                     continue
                 else:
--- a/lib/pbench/server/indexing_tarballs.py
+++ b/lib/pbench/server/indexing_tarballs.py
@@ -6,6 +6,7 @@ import signal
 import tarfile
 import tempfile
 from pathlib import Path
+from collections import deque
 
 from pbench.common.exceptions import (
     BadDate,
@@ -128,7 +129,7 @@ class Index:
                 idxctx.logger.info("No tar balls found that need processing")
                 return (0, [])
 
-        return (0, sorted(tarballs))
+        return (0, tarballs)
 
     def process_tb(self, tarballs):
         """Process Tarballs For Indexing and creates report
@@ -139,9 +140,11 @@ class Index:
         # We always process the smallest tar balls first.
         idxctx = self.idxctx
 
+        tb_deque = deque(sorted(tarballs))
+
         # At this point, tarballs contains a list of tar balls sorted by size
         # that were available as symlinks in the various 'linksrc' directories.
-        idxctx.logger.debug("Preparing to index {:d} tar balls", len(tarballs))
+        idxctx.logger.debug("Preparing to index {:d} tar balls", len(tb_deque))
 
         try:
             # Now that we are ready to begin the actual indexing step, ensure we
@@ -220,13 +223,22 @@ class Index:
                 def sigquit_handler(*args):
                     sigquit_interrupt[0] = True
 
+                sighup_interrupt = [False]
+
+                def sighup_handler(*args):
+                    sighup_interrupt[0] = True
+
                 signal.signal(signal.SIGQUIT, sigquit_handler)
+                signal.signal(signal.SIGHUP, sighup_handler)
+                count_processed_tb = 0
 
                 try:
-                    for size, controller, tb in tarballs:
+                    while len(tb_deque) > 0:
+                        size, controller, tb = tb_deque.popleft()
                         # Sanity check source tar ball path
                         linksrc_dir = Path(tb).parent
                         linksrc_dirname = linksrc_dir.name
+                        count_processed_tb += 1
                         assert linksrc_dirname == self.linksrc, (
                             f"Logic bomb!  tar ball "
                             f"path {tb} does not contain {self.linksrc}"
@@ -427,13 +439,33 @@ class Index:
 
                         if sigquit_interrupt[0]:
                             break
+                        if sighup_interrupt[0]:
+                            status, new_tb = self.collect_tb()
+                            if status == 0:
+                                if not set(new_tb).issuperset(tb_deque):
+                                    idxctx.logger.info(
+                                        "Tarballs supposed to be in 'TO-INDEX' are no longer present",
+                                        set(tb_deque).difference(new_tb),
+                                    )
+                                tb_deque = deque(sorted(new_tb))
+                            idxctx.logger.info(
+                                "SIGHUP status (Current tar ball indexed: ({}), Remaining: {}, Completed: {}, Errors_encountered: {}, Status: {})",
+                                Path(tb).name,
+                                len(tb_deque),
+                                count_processed_tb,
+                                _count_lines(erred),
+                                tb_res,
+                            )
+                            sighup_interrupt[0] = False
+                            continue
                 except SigTermException:
                     idxctx.logger.exception(
                         "Indexing interrupted by SIGQUIT, stop processing tarballs"
                     )
                 finally:
-                    # Turn off the SIGQUIT handler when not indexing.
+                    # Turn off the SIGQUIT and SIGHUP handler when not indexing.
                     signal.signal(signal.SIGQUIT, signal.SIG_IGN)
+                    signal.signal(signal.SIGHUP, signal.SIG_IGN)
             except SigTermException:
                 # Re-raise a SIGTERM to avoid it being lumped in with general
                 # exception handling below.
--- a/server/bin/pbench-index.py
+++ b/server/bin/pbench-index.py
@@ -170,9 +170,9 @@ def main(options, name):
 
     idxctx.logger.debug("{}.{}: starting", name, idxctx.TS)
 
-    idx = Index(name, options, idxctx, INCOMING_rp)
+    idx = Index(name, options, idxctx, INCOMING_rp, ARCHIVE_rp, qdir)
 
-    status, tarballs = idx.collect_tb(ARCHIVE_rp, qdir)
+    status, tarballs = idx.collect_tb()
     if status == 0 and tarballs:
         status = idx.process_tb(tarballs)
 
--- a/server/bin/pbench-index.py
+++ b/server/bin/pbench-index.py
@@ -105,6 +105,17 @@ def main(options, name):
                 - Raises an exception caught be a new, outer-most, try/except
                   block that does not have a finally clause (as you don't want
                   any code execution in the finally block).
+
+        4. Re-evaluate the list of tar balls to index after indexing of the
+            current tar ball has finished
+            - SIGHUP
+            - Report the current state of indexing
+                - Current tar ball being processed
+                - Count of Remaining tarballs
+                - Count of Completed tarballs
+                - No. of Errors encountered
+            - Handler Behavior:
+                - No exception raised
     """
 
     _name_suf = "-tool-data" if options.index_tool_data else ""
@@ -170,11 +181,11 @@ def main(options, name):
 
     idxctx.logger.debug("{}.{}: starting", name, idxctx.TS)
 
-    idx = Index(name, options, idxctx, INCOMING_rp, ARCHIVE_rp, qdir)
+    index_obj = Index(name, options, idxctx, INCOMING_rp, ARCHIVE_rp, qdir)
 
-    status, tarballs = idx.collect_tb()
+    status, tarballs = index_obj.collect_tb()
     if status == 0 and tarballs:
-        status = idx.process_tb(tarballs)
+        status = index_obj.process_tb(tarballs)
 
     return status
 
@@ -236,6 +247,7 @@ if __name__ == "__main__":
         signal.signal(signal.SIGTERM, sigterm_handler)
         signal.signal(signal.SIGINT, signal.SIG_IGN)
         signal.signal(signal.SIGQUIT, signal.SIG_IGN)
+        signal.signal(signal.SIGHUP, signal.SIG_IGN)
         status = main(parsed, run_name)
     except SigTermException:
         # If a SIGTERM escapes the main indexing function, silently set our


lib/pbench/server/indexing_tarballs.py
```diff
@@ -50,7 +50,7 @@ def _count_lines(fname):
 class Index:
     """class used to collect tarballs and index them"""
 
-    def __init__(self, name, options, idxctx, incoming):
+    def __init__(self, name, options, idxctx, incoming, archive, qdir):
```
portante, line 53: Can you isolate these changes to add the `archive` and `qdir` parameters in a separate commit from the signal handling changes?

portante (changes requested):
Good start.

Looks like you need to resolve a "LGTM.com" check that failed.

Also, the original intention of SIGHUP was to simply report the state of indexing.  While I like the idea of having SIGHUP also cause the indexer to re-evaluate the list of things to index, the goal was to have it immediately report the state of the indexing via log output while the indexing was on-going.

To do that, we'd need the interrupt handling to "kick" a thread which would perform the log output based on known state.

lib/pbench/server/indexing_tarballs.py
riya-17: Sure and will work on the reviiew

dbutenhof: Also, add SIGHUP to the pbench-index `main` docstring with the other exceptions!

riya-17: ok

lib/pbench/server/indexing_tarballs.py
```diff
@@ -425,13 +435,28 @@ def sigquit_handler(*args):
 
                         if sigquit_interrupt[0]:
                             break
+                        if sighup_interrupt[0]:
+                            status, tarballs = self.collect_tb()
+                            if status != 0:
+                                idxctx.logger.info(
+                                    "Unexpected error while recollecting tarballs"
```
dbutenhof, line 442: "recollecting" sounds odd here, and the normal sense of the word doesn't fit. How about "refreshing list of tarballs to index"??

Also, we continue on despite the failure... but on an error `collect_tb()` returns an empty tarball list, so you've forgotten whatever we started with.

I think you should try `status, new_tb = self.collect_tb()` instead, and assign `tarballs = new_tb` only if it succeeded.

Also, how does the iteration over `tarballs` work when you're completely replacing the list with a new list? Does Python handle this the way we want? (I don't know offhand.)

Finally, you're reporting how many we've done and how many remain based on "idx" relative to the number of elements in `tarballs` ... but `collect_tb()` won't find the datasets we've already indexed in the loop, so `tarballs` will be the number _remaining_ from the original list (unless more have been added). So, for example, if we find 10, index 5, and then get a `SIGHUP`, our `idx` is 5, and `collect_tb()` returns the 5 we haven't indexed yet... so your message will report that we've indexed 5 and there are 0 remaining (5 - 5). I think this logic needs to be re-worked.

While the "update the list" is an intriguing idea, remember that the indexer will be triggered again in a short time, so I'm not sure how useful that really is. The consistency issues would go away if you dropped that and just continued to iterate through the original list.

Eventually (and hopefully not that far off) the indexer will be demand-driven by messages that a new dataset is available.

server/bin/pbench-index.py
```diff
@@ -105,6 +105,17 @@ def main(options, name):
...
                   block that does not have a finally clause (as you don't want
                   any code execution in the finally block).
+        4. Re-evaluate the list of tar balls to index after indexing of the
+            current tar ball has finished
+            - SIGHUP
+            - Report the current state of indexing
+                - Current tar ball being processed
+                - How many remaining
+                - How many have been completed
+                - Errors encountered
```
dbutenhof, line 115: Reporting errors encountered might be useful, but you're not doing that.

server/bin/pbench-index.py
```diff
@@ -105,6 +105,17 @@ def main(options, name):
...
+            - Handler Behavior:
+                1. Kicks a special thread to log indexing state data
```
dbutenhof, line 117: The output isn't being done in a thread; you're generating it inline. Which I think is fine, since this isn't a lot of code. Although I think the original idea was that instead of synchronously looking for an interrupt the next time around the loop at this one spot, the interrupt handler would create a thread that would log the immediate status in parallel, regardless of where in the loop we are. (And note that trying to refresh `tarballs` in that context would be even more problematic than it is here.)

dbutenhof (changes requested):
I'm not sure this will even work as written... replacing a list you're iterating over is a potentially serious problem. There are also some inconsistencies in behavior and descriptions.

server/bin/pbench-index.py
```diff
@@ -170,9 +181,9 @@ def main(options, name):
 
     idxctx.logger.debug("{}.{}: starting", name, idxctx.TS)
 
-    idx = Index(name, options, idxctx, INCOMING_rp)
+    idx = Index(name, options, idxctx, INCOMING_rp, ARCHIVE_rp, qdir)
 
-    status, tarballs = idx.collect_tb(ARCHIVE_rp, qdir)
+    status, tarballs = idx.collect_tb()
```
webbnh, line 186: Reassigning `tarballs` here may not do what you hope (see the "Note" at the end of [this section](https://docs.python.org/3/reference/compound_stmts.html#the-for-statement)).

If you really want to implement this functionality, you should probably use a "queue" model, where you start with a list of tarballs and destroy the list as you process it (e.g., use a `while tarballs` instead of the `for`, and remove the tarball from the list before processing it).

webbnh (approved):
These changes look OK to me, but beware of the issue I noted.

server/bin/pbench-index.py
webbnh: I think doing it "in-line" is far better than trying to use a thread:

- Python threads don't actually offer much concurrency -- Python bytecode execution is done under a global lock, so a thread executing "Python code" only executes in parallel with other threads which _aren't_ running "Python code" (like stuff doing I/O or executing an external library such as NumPy which is written in C).
- Signal delivery is apparently also synchronous with respect to bytecode execution, so the notification is going to wait until "a good time", anyway.  (I.e., it might not actually interrupt anything.)
- The in-line approach is _much_ simpler and easier to understand and maintain.
- The execution of the interrupt is not performance-critical.
- Adding threads makes the update of the `tarballs` list much more complicated.

server/bin/pbench-index.py
```diff
@@ -105,6 +105,17 @@ def main(options, name):
                 - Raises an exception caught be a new, outer-most, try/except
                   block that does not have a finally clause (as you don't want
                   any code execution in the finally block).
+        4. Re-evaluate the list of tar balls to index after indexing of the
```
webbnh, line 108: Looks like there should be a blank line before this section, as there are for the other sections.

lib/pbench/server/indexing_tarballs.py
riya-17: > "recollecting" sounds odd here, and the normal sense of the word doesn't fit. How about "refreshing list of tarballs to index"??

I will change it to this or webb's  suggestion is also a good one



> Also, we continue on despite the failure... but on an error `collect_tb()` returns an empty tarball list, so you've forgotten whatever we started with.

I thought outside for loop will take care of that by completing the iterations, breaking it here might create unwanted terminations



> I think you should try `status, new_tb = self.collect_tb()` instead, and assign `tarballs = new_tb` only if it succeeded.

yeah sure I thought reassigning directly will make more sense



> Also, how does the iteration over `tarballs` work when you're completely replacing the list with a new list? Does Python handle this the way we want? (I don't know offhand.)

when the signal for the SIGHUP handler is called then it completes processing of the current tarball and then checks whether the signal is interrupted. Unindexed tarballs are all present in the "To-Index" folder so it will contain all the new tarballs that just came in and also all the tarballs that would get processed after this iteration. So, if we re-check that folder then it would contain all the tarballs that need to be indexed else they will get duplicated. like we are processing [1, 2, 3, 4] and [1, 2] are indexed, "To-index" folder will contain [3, 4]  as well as new tarballs that came in [5, 6]. So, we can just check tarballs in the "to-index" folder and replace it with an existing list which is 'tarballs'.



> Finally, you're reporting how many we've done and how many remain based on "idx" relative to the number of elements in `tarballs` ... but `collect_tb()` won't find the datasets we've already indexed in the loop, so `tarballs` will be the number _remaining_ from the original list (unless more have been added). So, for example, if we find 10, index 5, and then get a `SIGHUP`, our `idx` is 5, and `collect_tb()` returns the 5 we haven't indexed yet... so your message will report that we've indexed 5 and there are 0 remaining (5 - 5). I think this logic needs to be re-worked.

yeah I didn't think it thoroughly will work on it

> While the "update the list" is an intriguing idea, remember that the indexer will be triggered again in a short time, so I'm not sure how useful that really is. The consistency issues would go away if you dropped that and just continued to iterate through the original list.

What would be best?

riya-17: > @dbutenhof, perhaps a hyphen would help -- "re-collecting"? :-D
> 
> As for reassigning `tarballs`, see my comment below on `pbench-index.py` line 186.

yeah sure

server/bin/pbench-index.py
riya-17: I will add that

server/bin/pbench-index.py
riya-17: yeah, I have thought about the queue earlier but there were different reasons for not using that here. I will look into it again.

server/bin/pbench-index.py
riya-17: yeah sure

server/bin/pbench-index.py
riya-17: @dbutenhof which error should I report? like there is also a report generation at the end of the indexing that also reports no. of errors. So, should I just report the no. of errors the same ways or should I give details as well? also if there will be an error at the time of collection of tarballs inside sighup we can report that as well.

lib/pbench/server/indexing_tarballs.py
```diff
@@ -425,13 +441,33 @@ def sigquit_handler(*args):
...
+                        if sighup_interrupt[0]:
+                            status, new_tb = self.collect_tb()
+                            if status == 0 and new_tb:
+                                # FIXME: add into deque
+                                # We have two choices to add into deque
+                                # Either sort the whole deque which might lead into starvation of
+                                # tarball or just extend the deque with new list
+
+                                # Here I will be using the second approach and we can discuss further
+                                tb_deque.extend(new_tb)
```
dbutenhof, line 453: Won't this add duplicates if we still haven't processed all the original tarballs?

webbnh (changes requested):
It looks pretty good.  I've noted a few small things which should probably be fixed before merging.  The only real issue is updating the tarball list after a SIGHUP -- I think the list should be replaced, not augmented.

lib/pbench/server/indexing_tarballs.py
ndokos: We currently sort them by size: there have been instances of large tarballs taking "forever" to index, so dealing with the small ones first helped a bit. Of course, sorting this batch doesn't do anything about subsequent tarballs and even with SIGHUP adding new tarballs and re-sorting, that will not be solved: if a large tarball starts being indexed, everybody else (including the SIGHUP handling) will have to wait. OTOH, given a steady supply of tarballs and a twitchy SIGHUP finger, it's possible that  a large tarball will be starved. But no matter what we do with the current implementation, this will continue to be a problem. 

So I think, we hold the SIGHUP handling up to (or down to) the same standard: sort the current set of tarballs and process them in that order. Scheduling them in parallel, possilby with different queues handling different size classes, is something to think about. But we all know that the heuristics break down in particular circurmstances.

Scheduling is hard (I have to add that to my sig...)

ndokos: The only thing that we might want to check for sanity's sake is that the new list is a superset of what is left of the old list. I'm not sure that's particularly important  and I don't think we should stop if that is not the case, but having an error message listing any "lost" tarballs might be worth having, even though it should never trigger.

ndokos (commented):
A couple of comment inline.

server/bin/pbench-index.py
riya-17: oh, I must have missed the comment ;P

server/bin/pbench-index.py
riya-17: ok sure

lib/pbench/server/indexing_tarballs.py
```diff
@@ -218,13 +224,23 @@ def process_tb(self, tarballs):
...
+                    sighup_interrupt[0] = True
+
                 signal.signal(signal.SIGQUIT, sigquit_handler)
+                signal.signal(signal.SIGHUP, sighup_handler)
+                idx = 0
 
                 try:
-                    for size, controller, tb in tarballs:
+                    while len(tb_deque) > 0:
+                        # print(f"size = {size}, controller = {controller}, tb = {tb}")
```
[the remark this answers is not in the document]
riya-17: yeah sure

lib/pbench/server/indexing_tarballs.py
```diff
@@ -137,9 +140,12 @@ def process_tb(self, tarballs):
         # We always process the smallest tar balls first.
         idxctx = self.idxctx
 
+        tb_deque = deque()
+        [tb_deque.append(i) for i in sorted(tarballs)]
```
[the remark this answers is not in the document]
riya-17: yeah this will be better

lib/pbench/server/indexing_tarballs.py
riya-17: oh, I must have forgotten about replacing the list it just gets out of my mind. Checking for the superset and logging the lost tarballs is a good idea I will add that.

lib/pbench/server/indexing_tarballs.py
```diff
@@ -425,13 +441,33 @@ def sigquit_handler(*args):
...
+
+                                # Here I will be using the second approach and we can discuss further
+                                tb_deque.extend(new_tb)
+                            idxctx.logger.info(
+                                "SIGHUP status (Current tar ball indexed: ({}), Remaining: {}, Completed: {}, Errors_encountered: {})",
+                                tb,
+                                len(tb_deque),
+                                idx,
+                                _count_lines(erred),
+                            )
```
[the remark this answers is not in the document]
riya-17: ok

riya-17:
@ndokos @dbutenhof @webbnh I have addressed all the comments. Please take a look

lib/pbench/server/indexing_tarballs.py
```diff
@@ -425,13 +439,33 @@ def sigquit_handler(*args):
...
                         if sigquit_interrupt[0]:
                             break
+                        if sighup_interrupt[0]:
+                            status, new_tb = self.collect_tb()
+                            if status == 0 and new_tb:
+                                if not set(new_tb).issuperset(tb_deque):
+                                    idxctx.logger.info(
+                                        "Unexpected Situation: ",
+                                        set(tb_deque).difference(new_tb),
+                                    )
```
dbutenhof, line 449: I don't expect this to happen, and I'm not sure we need another iteration of the PR to fix it, but a little more context might have been useful if this message ever _does_ appear. Some of the tarballs we thought we still had to process are no longer showing as `TO-INDEX`, and that's puzzling. Simply listing the "missing" tarballs makes this message a bit more mysterious than it needs to be.

lib/pbench/server/indexing_tarballs.py
```diff
@@ -218,13 +223,22 @@ def process_tb(self, tarballs):
...
                     sigquit_interrupt[0] = True
 
+                sighup_interrupt = [False]
+
+                def sighup_handler(*args):
+                    sighup_interrupt[0] = True
+
                 signal.signal(signal.SIGQUIT, sigquit_handler)
+                signal.signal(signal.SIGHUP, sighup_handler)
+                idx = 0
```
dbutenhof, line 233: It occurs to me that `idx` isn't a particularly meaningful name anymore. It's a "count of processed tarballs". But while it might be nice to make this more clear should you make some other change, I don't want to push you to make another commit for this small detail.

server/bin/pbench-index.py
```diff
@@ -170,9 +181,9 @@ def main(options, name):
 
     idxctx.logger.debug("{}.{}: starting", name, idxctx.TS)
 
-    idx = Index(name, options, idxctx, INCOMING_rp)
+    idx = Index(name, options, idxctx, INCOMING_rp, ARCHIVE_rp, qdir)
```
dbutenhof, line 184: This isn't really a great use of `idx` either. Generally, short and abbreviated names should be used for things like "throwaway" loop variables, with longer and more meaningful names for variable of more importance and wider scope.

I suppose the fact that I'm paying attention to minor details on this review pass could be because I'm not seeing any important issues.

Again, not anything that needs to be resolved right now, but something to consider if you have to make another change for some other reason.

dbutenhof (approved):

lib/pbench/server/indexing_tarballs.py
riya-17: I was quite not sure about the log message I should put. So. I will just change that

lib/pbench/server/indexing_tarballs.py
```diff
@@ -425,13 +439,33 @@ def sigquit_handler(*args):
 
                         if sigquit_interrupt[0]:
                             break
+                        if sighup_interrupt[0]:
+                            status, new_tb = self.collect_tb()
+                            if status == 0 and new_tb:
```
webbnh, line 444: Here's an awkward question:  if `new_tb` evaluates to `False`, then we won't do the superset test and, more importantly, we won't update `tb_deque`.  The principal reason why `new_tb` would evaluate to `False` (accompanied by a "success" status) would be if someone went in and deleted all of the two-be-indexed datasets, presumably in an attempt to abort indexing.  So, what is going to happen next is that the tarball open (or something similar) will fail...which seems ugly.

This is a long-winded setup to ask, why are we testing `new_tb` at line 444?  Wouldn't it be better to just proceed based on the value of `status`?  (If `new_tb` turns out to be empty, won't proceeding make happen what we want to have happen?)

(Riya, I leave it up to you to decide how to address this -- you could fix it in this PR, do it in another, or just capture it as an issue.)

dbutenhof (approved):

webbnh (dismissed):
Looks good.

Unfortunately, I found an issue...but maybe it doesn't need to be dealt with in _this_ PR.

lib/pbench/server/indexing_tarballs.py
riya-17: > Here's an awkward question: if `new_tb` evaluates to `False`, then we won't do the superset test and, more importantly, we won't update `tb_deque`. The principal reason why `new_tb` would evaluate to `False` (accompanied by a "success" status) would be if someone went in and deleted all of the two-be-indexed datasets, presumably in an attempt to abort indexing. So, what is going to happen next is that the tarball open (or something similar) will fail...which seems ugly.

@webbnh The reason that new_tb would evaluate as false is when there aren't any tarballs that need indexing("TO-INDEX" folder is empty). Maybe l.128 through l.130 can clear something up.


>


>
> This is a long-winded setup to ask, why are we testing `new_tb` at line 444? Wouldn't it be better to just proceed based on the value of `status`? (If `new_tb` turns out to be empty, won't proceeding make happen what we want to have happen?)

we could have proceeded just with status value but if there would have been no tarballs then all the checkings and replacing inside condition will all just not make sense.
if 'new_tb' will turn out to be empty then 'tb_deque' will run as it should and will end as soon as 'tb_deque' is empty.

Am I missing something?

webbnh: If we're running this code, then `tb_deque` had at least one entry in it -- let's assume that it had several -- and something sent a `SIGHUP` which made us call `collect_tb()`.  Now, for the sake of argument, suppose that the TO-INDEX folder is now empty (somehow...), which means that `new_tb` will be empty, but, for the sake of argument, suppose that `tb_deque` is _not_ empty (because we've indexed only the first item of the several).

At this point, we have an interesting choice.  We _could_ update `tb_deque` to be empty, which I think makes good sense, because there is presumably nothing in the TO-INDEX folder, now.  This will (I think) allow indexing to conclude in an orderly fashion.  The alternative, as currently coded, is to ignore the update, do nothing to `tb_deque`, and, on the next iteration of the loop, discover that the next tarball is nowhere to be found.

> Am I missing something?

Sleep?  :-D

webbnh: (Or, am _I_ missing something??  :-) )

riya-17: It is quite an exceptional condition but there is no harm in making it work. As, either both(tb_deque and new_tb) should be empty or both should be containing the same elements.

> Sleep? :-D

haha

riya-17: but yeah there is one more thing it will also add the somehow deleted tarballs in the missing log after not passing the superset condition.

dbutenhof (approved):

lib/pbench/server/indexing_tarballs.py
riya-17: fixed

webbnh (approved):

lib/pbench/server/indexing_tarballs.py
webbnh: > it will also add the somehow deleted tarballs in the missing log after not passing the superset condition.

And that's exactly what I think we want:  the tarballs were there when we started, and they are not there now, and they didn't get indexed, and that's weird and worth logging.

PR #2116, opened 2021-02-17
Fix unused variables from lgtm bot

RobertKrawitz:
Fix some errors from lgtm bot and update todo list

### DIFF
--- a/contrib/analysis/generate-pbench-timeseries-graphs
+++ b/contrib/analysis/generate-pbench-timeseries-graphs
@@ -20,21 +20,21 @@
 #
 # Lots of TODOs:
 #
-# 1) Command line arguments for image size, max # of data sets, what else?
-# 2) Actually feed the output into gnuplot rather than just sending it to stdout
-# 3) Optionally sort data sets by max value or max average
-# 4) Combine multiple samples into 1.
-#    This will require that generate-pbench-timeseries-graphs enumerate the files
-#    (or accept the file list on stdin rather than command line) because the command
-#    line won't be able to accept all of the files in a data set.
-# 5) Handle .csv files from perf and any other tools I've missed.
-# 6) Allow scaling of Y and possibly X axis.
-# 7) Allow command line override of per-file settings.
-# 8) Allow additional gnuplot options
-# 9) Error checking
-# 10) Make data max/minvals available to graph generator so that it can set
-#    the axis bounds in gnuplot correctly.
-# 11->Graham's number ...
+# DONE 1) Command line arguments for image size, max # of data sets, what else?
+#      2) Actually feed the output into gnuplot rather than just sending it to stdout
+#      3) Optionally sort data sets by max value or max average
+#      4) Combine multiple samples into 1.
+#         This will require that generate-pbench-timeseries-graphs enumerate the files
+#         (or accept the file list on stdin rather than command line) because the command
+#         line won't be able to accept all of the files in a data set.
+#      5) Handle .csv files from perf and any other tools I've missed.
+#      6) Allow scaling of Y and possibly X axis.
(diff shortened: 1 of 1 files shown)


contrib/analysis/generate-pbench-timeseries-graphs
```diff
@@ -58,6 +58,17 @@ parser.add_argument('-n', "--dont-generate-graphs", action='store_true', help="D
 parser.add_argument('files', metavar='file', type=str, nargs="*", help="files")
 args = parser.parse_args()
 
+match_res = []
+nomatch_re = None
+initialized_matches = False
```
portante, line 63: Is this variable ever used anywhere else?

portante (changes requested):
Looks good, just one question ...

contrib/analysis/generate-pbench-timeseries-graphs
webbnh: Not that I could find....  It seems pointless, now.

contrib/analysis/generate-pbench-timeseries-graphs
```diff
@@ -58,6 +58,17 @@ parser.add_argument('-n', "--dont-generate-graphs", action='store_true', help="D
...
+if args.match:
+    for pattern in args.match:
+        match_res.append(re.compile(pattern))
+if args.dont_match:
+    nomatch = "|"
+    nomatch = "(%s)" % nomatch.join(args.dont_match)
+    nomatch_re = re.compile(nomatch)
```
webbnh, line 70: Just out of curiosity, why are we using different idioms for the positive and negative matches?

contrib/analysis/generate-pbench-timeseries-graphs
RobertKrawitz: Yeah, I suppose not any more.  It was intended for lazy evaluation, but if I'm going to statically evaluate these lists, no need for it at all.

contrib/analysis/generate-pbench-timeseries-graphs
RobertKrawitz: The intended use case here is to easily process selected data from selected iterations, i. e. constraints from different levels of the tree.  For example, if I want to process pidstat data from iteration 10, I can use

```
-m /10- -m /pidstat
```
If we use the union of all matches, it becomes very awkward to express something like this.  It's a bit awkward if we want to look at iterations 10 and 11, but it's not too bad.  For example, if I want to process pidstat and proc-interrupt data from iterations 10 and 11, it becomes

```
-m '/(10|11)-' -m '/(pidstat|proc-interrupt)'
```
If I want to filter out certain patterns, such as per-CPU, per-IRQ, or per-PID data, then I could use that syntax, or I could use multiple non-match expressions.  But it's not clear to me what the use case for "don't process filenames matching all --dont-match expressions" would be.

```
-m /(10|11)-' -m '(pidstat|proc-interrupt)' -M /pids/ -M by-irq_IRQ
```

webbnh: I concur that generalized match/search syntax is a challenge.  The more comprehensive CLI implementations that I've seen apply the match/don't-match filters in sequence (e.g., a sequence of `-m`'s would be a conjunction, subsequent `-M`'s would filter the initial result, and then subsequent `-m`'s would form a conjunction and select from the filtered result, and so on), but that seems like an awful lot of work, and I'm not sure that it covers all cases.

Regardless, what I was actually asking here was, why are matches represented as a list of individual REs, while don't-matches are combined into a single RE?  (Is not what is good for the goose sauce for the gander?)

webbnh (approved):

portante (approved):

PR #2123, opened 2021-02-25
Fix .coveragerc omit

portante:
This was pulled from PR #2076.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,6 @@
 [run]
 parallel = True
 data_file = /tmp/${USER}/cov/coverage.db
+
+[run]
 omit = */site-packages/*


.coveragerc
```diff
@@ -1,4 +1,6 @@
 [run]
 parallel = True
 data_file = /tmp/${USER}/cov/coverage.db
+
+[report]
```
dbutenhof, line 5: That's ... interesting. While this may be slightly more costly, I doubt it really matters. Essentially, according to my understanding of the documentation, the difference ought to be that my original change doesn't _collect_ coverage metrics for the site-specific Python libraries, whereas your version collects the data but doesn't report it.

Does this actually make a real difference in the run?

dbutenhof (dismissed):
I guess I don't object, if this helps or you just like it this way; however, I don't understand the rationale, so a better explanation would be nice.

.coveragerc
portante: Corrected.

portante (commented):
Now updated to use the correct section name.

.coveragerc
```diff
@@ -1,4 +1,6 @@
 [run]
 parallel = True
 data_file = /tmp/${USER}/cov/coverage.db
+
+[run]
```
portante, line 5: Okay, not sure what I was thinking, but this is already taken care of.

PR #2124, opened 2021-03-01
Streamline server RPM spec file

dbutenhof:
Although the Agent spec file efficiently references directories to be packaged hierarchically, the Server spec file has a full list of the individual packaged files. This is unnecessarily awkward to maintain, and error prone.

The challenge here was some experimentation (the documentation is terrible) to figure out the order of precedence across the `%attr `and `%defattr` with directories and contained individual files. I added some comments in the spec file to clarify.

### DIFF
--- a/server/rpm/pbench-server.spec.j2
+++ b/server/rpm/pbench-server.spec.j2
@@ -101,101 +101,22 @@ fi
 %posttrans
 
 %files
+
 %defattr(644, pbench, pbench, 755)
 /%{installdir}/VERSION
 /%{installdir}/SEQNO
 /%{installdir}/SHA1
-/%{installdir}/%{static}/VERSION
 /%{installdir}/package.json
 /%{installdir}/requirements.txt
 
-/%{installdir}/lib/config/pbench-server-satellite.cfg.example
-/%{installdir}/lib/config/pbench-server.cfg.example
-/%{installdir}/lib/config/pbench-server-default.cfg
+# pbench-base.sh %attr overrides previous /bin %attr
+%attr(755, pbench, pbench) /%{installdir}/bin
+%attr(644, pbench, pbench) /%{installdir}/bin/pbench-base.sh
 
-/%{installdir}/lib/crontab
-/%{installdir}/lib/mappings
-/%{installdir}/lib/settings
-
-/%{installdir}/lib/pbench/common/configtools.py
-/%{installdir}/lib/pbench/common/__init__.py
-/%{installdir}/lib/pbench/common/conf.py
-/%{installdir}/lib/pbench/common/constants.py
-/%{installdir}/lib/pbench/common/exceptions.py
-/%{installdir}/lib/pbench/common/logger.py
-/%{installdir}/lib/pbench/common/utils.py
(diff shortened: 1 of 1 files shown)


dbutenhof:
FYI, it's probably worth mentioning that I ran through several installs on my test VM to verify the behavior; the files are there, and the permissions look right. The RPM is available as `0.71.0-3gcca4629b9` on my `dbutenhof/pbench-test` on COPR.

server/rpm/pbench-server.spec.j2
```diff
@@ -101,101 +101,22 @@ fi
 %posttrans
 
 %files
+
 %defattr(644, pbench, pbench, 755)
```
webbnh, line 105: What is the purpose/difference between the two mode-mask arguments to the `%defattr` directive?

webbnh (approved):
Seems good.

You don't want to use the directory-reference approach for the root directory, `/%{installdir}/`, as well?  (Is there some hidden requirement/benefit to listing those files explicitly?)  Would doing so allow you to omit listing anything else (except for the `/%{installdir}/bin` exception, the exception to that exception, and the other exception)?

server/rpm/pbench-server.spec.j2
dbutenhof: The second one is for directories; which is why there's no equivalent on `%attr`.

dbutenhof:
> Seems good.
> 
> You don't want to use the directory-reference approach for the root directory, `/%{installdir}/`, as well? (Is there some hidden requirement/benefit to listing those files explicitly?) Would doing so allow you to omit listing anything else (except for the `/%{installdir}/bin` exception, the exception to that exception, and the other exception)?

I briefly considered it; it'd mean trying to break in the middle of the assembly process and make sure there's nothing else that we *don't* want. I also thought that calling out the "installer config" type files explicitly was probably useful documentation.

webbnh:
Works for me.

ndokos (approved):

riya-17:
@dbutenhof So, how are they fetching files now or how is it working?

dbutenhof:
> @dbutenhof So, how are they fetching files now or how is it working?

The RPM builder has always had the capability to process an entire directory tree with a single line in the `%files` section; the agent spec file was already using that, but the server for some reason wasn't.

The only tricky part here (requiring several build/install cycles to debug) was working out how to specify a directory tree with one `%attr` (or %defattr) set while overriding that for a specific file within that tree.

I submitted multiple builds to COPR and installed them on my VM to check the results.

riya-17:
> > @dbutenhof So, how are they fetching files now or how is it working?
> 
> The RPM builder has always had the capability to process an entire directory tree with a single line in the `%files` section; the agent spec file was already using that, but the server for some reason wasn't.
> 
> The only tricky part here (requiring several build/install cycles to debug) was working out how to specify a directory tree with one `%attr` (or %defattr) set while overriding that for a specific file within that tree.
> 
> I submitted multiple builds to COPR and installed them on my VM to check the results.

Ohh That's how.. Thanks

PR #2133, opened 2021-03-05
Fix es url formation when params exist in json data

npalaska:
During the discussion with @gurbirkalsi on es queries from dashboard we found a bug in server elasticsearch api endpoint about how we format search url when `ignore_unavailable` key is present in the params of the json data.

If params present in the json data we were formatting it like this 
`"{self.elasticsearch}/{json_data['indices']}?{json_data['params']}"` 
instead it should be
`"{self.elasticsearch}/{json_data['indices']}?ignore_unavailable={json_data['params']['ignore_unavailable']}"`

### DIFF
--- a/lib/pbench/server/api/resources/query_apis/elasticsearch_api.py
+++ b/lib/pbench/server/api/resources/query_apis/elasticsearch_api.py
@@ -25,12 +25,10 @@ class Elasticsearch(Resource):
 
         try:
             # query Elasticsearch
+            url = f"{self.elasticsearch}/{json_data['indices']}"
             if "params" in json_data:
-                url = (
-                    f"{self.elasticsearch}/{json_data['indices']}?{json_data['params']}"
-                )
-            else:
-                url = f"{self.elasticsearch}/{json_data['indices']}"
+                if "ignore_unavailable" in json_data["params"]:
+                    url = f"{url}?ignore_unavailable={json_data['params']['ignore_unavailable']}"
 
             if "payload" in json_data:
                 es_response = requests.post(url, json=json_data["payload"])


lib/pbench/server/api/resources/query_apis/elasticsearch_api.py
```diff
@@ -25,12 +25,10 @@ def post(self):
...
-                url = (
-                    f"{self.elasticsearch}/{json_data['indices']}?{json_data['params']}"
-                )
-            else:
-                url = f"{self.elasticsearch}/{json_data['indices']}"
+                if "ignore_unavailable" in json_data["params"]:
+                    url = f"{url}?ignore_unavailable={json_data['params']['ignore_unavailable']}"
 
             if "payload" in json_data:
                 es_response = requests.post(url, json=json_data["payload"])
```
dbutenhof, line 34: A better (cleaner, and more general) mechanism might be to pass through a distinct `params` to the server API to use in the `requests` POST directly. That way we don't need to worry about `?` vs `&`, and so forth... which is the point of separating the query parameters in the requests interfaces in the first place. Rather than trying to hardcode them in the URI string...

dbutenhof (approved):
As a point fix, fine; a more flexible approach is probably better in the long run.

lib/pbench/server/api/resources/query_apis/elasticsearch_api.py
npalaska: yes I thought about that but this is an open es endpoint disguised as `<server>/elasticsearch` to hide the actual endpoint, we anyway not gonna have this endpoint in the long run. Right now dashboard does some es queries directly to es and to support that for a while this is just a hot fix for correctly formatting the es URI

dbutenhof: I'm actually not sure we don't want to continue support for this endpoint. This would enable more generic open-ended ES queries for clients than we support via canned "endpoint" queries for advanced exploration, without opening direct ES server access. We'll have to figure out how to balance that against user-based security, but I'd thought about something like a "direct ES" Role that we could allow for specific users.

Anyway; again, as a point fix, this is fine, and we can worry about how to generalize/support something like this going forward in a separate context.

dbutenhof (approved):

gurbirkalsi (approved):

PR #2135, opened 2021-03-15
OpenStack SOS plugins support in pbench-collect-sysinfo

pradiptapks:
Added OpenStack SOS plugin support in `pbench-sysinfo-dump` and the plugin run based on the condition check of  `/etc/rhosp-release` in the remote nodes.


dbutenhof (dismissed):
This lists "dependency" modules (including podman, kvm) that aren't specifically "OpenStack", and that could lead to confusion later; however short of a complete new modular organization to detect specific plugins individually, I don't have any meaningful alternative to suggest and adding these seems useful.

pradiptapks:
@dbutenhof podman and kvm are still required as OSP container services managed by podman and kvm for OpenStack compute nodes.

agent/util-scripts/pbench-sysinfo-dump
```diff
@@ -116,6 +116,11 @@ function collect_sos {
 		# ref for plugin tuned: https://github.com/sosreport/sos/commit/c55c58cf2dabf93c924c839c8ed045c18e31ba1c
 		_modules="${_modules} tuned"
 	fi
+        #OpenStack plugin
+        OSP=/etc/rhosp-release
+        if [[ -f "$OSP" ]] ; then
+                _modules="${_modules} openstack_ansible openstack_aodh openstack_ceilometer openstack_cinder openstack_glance openstack_heat openstack_horizon openstack_instack openstack_ironic openstack_keystone openstack_manila openstack_neutron openstack_nova openstack_novajoin openstack_octavia openstack_placement openstack_swift os_net_config mssql cman ceph openvswitch ovn_central ovn_host pacemaker podman rabbitmq kernelrt ceph ceph-ansible kvm corosync"
```
webbnh, line 122: @pradiptapks, is `ceph` supposed to be in the list twice?

@dbutenhof, should we be considering doing duplicate-removals on this list, or does it actually matter (i.e., will sosreport handle that adequately)?

portante: We can't do plugins this way, because they are very tied to the version of the OS.  Each one has to be check against the versions of `sosreport` that we execute against (RHEL 7.7, .8, .9, RHEL 8.*, RHEL 9, Fedora 32 & 33, CentOS 7 & 8, CentOS Streams?).

portante (changes requested):
Unfortunately, `sosreport` modules are tough to manage.

We have to be careful that they don't balloon the size of the data collected (no logs, etc.).

We really need a mechanism for the user to specify the list of modules they want so that they can own what is collected.

agent/util-scripts/pbench-sysinfo-dump
pradiptapks: @webbnh I removed `ceph_ansible` as `ceph` is sufficient to collect the log information.

@portante OpenStack parsers in `sosreport` are common to RHEL7.x, 8.x, CentOS, Fedora. As `podman` has introduced in RHEL8.x, I wonder whether it does backwards support below version 3. May I know which version of so start support podman, so that I can change the condition.

agent/util-scripts/pbench-sysinfo-dump
```diff
@@ -116,6 +116,13 @@ function collect_sos {
 		# ref for plugin tuned: https://github.com/sosreport/sos/commit/c55c58cf2dabf93c924c839c8ed045c18e31ba1c
 		_modules="${_modules} tuned"
 	fi
+	if [[ "${sos_ver}" -ge 3 || "${sos_ver}" -le 3 ]]; then
+		# sos-plugins for OpenStack cloud
+		OSP=/etc/rhosp-release
+		if [[ -f "$OSP" ]] ; then
+			_modules="${_modules} openstack_ansible openstack_aodh openstack_ceilometer openstack_cinder openstack_glance openstack_heat openstack_horizon openstack_instack openstack_ironic openstack_keystone openstack_manila openstack_neutron openstack_nova openstack_novajoin openstack_octavia openstack_placement openstack_swift os_net_config mssql cman ceph openvswitch ovn_central ovn_host pacemaker podman rabbitmq kernelrt ceph kvm corosync"
```
webbnh, line 123: I don't know whether it actually matters, but I still see `ceph` in this list twice -- once after `cman` and once after `kernelrt`.

PR #2137, opened 2021-03-16
Add pbench-generate-token tool

webbnh:
This pull request is a series of commits culminating in the addition of a simple tool which takes a username/password pair and generates a token for use with the pbench server's REST interface.  The targeted use case for the token is to allow Riya's overhaul of the `pbench-move-results` tool to make an authenticated access to the server.

Most of the commits in this pull request are small, point changes to things which I encountered in the process of implementing the new tool:

- Fix docstring typos in `cli/agent/__init__.py` (no functional changes).
- Tweak Click options for agent config command line option, to make it behave more gracefully.
- Tweak the setup option names to replace hyphens with underscores, since my installation was generating warnings.

The new tool accepts a username and password.  These can be specified on the command line for convenience, but, of course, putting the password on the command line in clear text will likely allow it to "leak" to log files or to other users on the system.  If the username and/or password is not specified on the command line, the tool will prompt for them.  The response to the password prompt is not echoed, for security reasons.

The tool makes a `PUT` request to the pbench server creating a "login" and receiving a token from the response.  If the request is successful, the token is printed to `stdout` and the tool exits with a success (zero) status.  The output can be captured either by redirecting `stdout` to a file or by copying-and-pasting the terminal output.  If the request is unsuccessful, the tool exits with an error status with a diagnostic sent to `stderr`.

The token value can then be stored in a "secret" which is made available to other tools such as `pbench-move-results` in their respective execution environments, or it can be specified in clear-text, because it is an ephemeral value.

This implementation of `pbench-generate-token` provides latent support for allowing the user to specify the duration for which the token is valid.  The current server implementation provides a fixed duration, so in effect any value specified to this tool is ignored.  However, in the future, once we have support for role-based authorization in the server, it will be practical to provide variable duration tokens, at which point the option in this tool will become useful.

[This PR now includes unit tests and is ready for formal review; comments provided during the draft phase have been incorporated into their respective commits, and the branch has been rebased on the end of `main`.]

### DIFF
--- a/lib/pbench/cli/agent/commands/generate_token.py
+++ b/lib/pbench/cli/agent/commands/generate_token.py
@@ -1,0 +1,80 @@
+"""pbench-generate-token"""
+
+import click
+import requests
+
+from pbench.agent.base import BaseCommand
+from pbench.cli.agent import pass_cli_context
+from pbench.cli.agent.options import common_options
+
+
+class GenerateToken(BaseCommand):
+    def __init__(self, context):
+        super().__init__(context)
+
+    def execute(self):
+        """Generate a token used to access the pbench server RESTful API"""
+
+        server = self.config.agent.parser.get("results", "server_rest_url")
+        headers = {"Content-Type": "application/json", "Accept": "application/json"}
+        # TODO:  The server does not currently accept the 'token_duration' key,
+        #  so it just gets ignored, for now.
+        payload = {
+            "username": self.context.username,
+            "password": self.context.password,
+            "token_duration": self.context.token_duration,
+        }
+        try:
+            response = requests.post(f"{server}/login", headers=headers, json=payload)
+        except requests.exceptions.ConnectionError as exc:
+            if hasattr(exc.args[0], "reason"):
+                raise exc.args[0].reason
+            raise
+
+        payload = response.json()
+        if response.ok:
+            click.echo(payload["auth_token"])
+            return 0
+
+        click.echo(
+            payload["message"] if "message" in payload else response.reason, err=True
+        )
+        return 1
+
+
+@click.command()
+@common_options
+@click.option(
+    "--username",
+    prompt=True,
+    required=True,
+    help="pbench server account username (will prompt if unspecified)",
+)
+@click.option(
+    "--password",
+    prompt=True,
+    hide_input=True,
+    required=True,
+    help="pbench server account password (will prompt if unspecified)",
+)
+@click.option(
+    "--token-duration",
+    type=int,
+    required=False,
+    default=3600,
+    show_default=True,
+    help="number of seconds",
+)
+@pass_cli_context
+def main(context, username, password, token_duration):
+    context.username = username
+    context.password = password
+    context.token_duration = token_duration
+
+    try:
+        rv = GenerateToken(context).execute()
+    except Exception as exc:
+        click.echo(exc, err=True)
+        rv = 1
+
+    click.get_current_context().exit(rv)
--- a/lib/pbench/cli/agent/options.py
+++ b/lib/pbench/cli/agent/options.py
@@ -19,13 +19,13 @@ def _pbench_agent_config(f):
     return click.option(
         "-C",
         "--config",
+        required=True,
         envvar="_PBENCH_AGENT_CONFIG",
-        type=click.Path(exists=True),
+        type=click.Path(exists=True, readable=True),
         callback=callback,
         expose_value=False,
         help=(
-            "Path to a pbench-agent config. If provided pbench will load "
-            "this config file first. By default is looking for config in "
-            "'_PBENCH_AGENT_CONFIG' envrionment variable."
+            "Path to a pbench-agent configuration file (defaults to the "
+            "'_PBENCH_AGENT_CONFIG' environment variable, if defined)"
         ),
     )(f)
--- a/lib/pbench/test/unit/agent/task/test_generate_token.py
+++ b/lib/pbench/test/unit/agent/task/test_generate_token.py
@@ -1,0 +1,167 @@
+from click.testing import CliRunner
+import requests
+import responses
+
+from pbench.cli.agent.commands import generate_token
+
+
+class TestGenerateToken:
+
+    USER_SWITCH = "--username"
+    PSWD_SWITCH = "--password"
+    USER_PROMPT = "Username: "
+    PSWD_PROMPT = "Password: "
+    USER_TEXT = "test_user"
+    PSWD_TEXT = "password"
+    TOKEN_TEXT = "what is a token but 139 characters of gibberish"
+    URL = "http://pbench.example.com/api/v1"
+    ENDPOINT = "/login"
+
+    @staticmethod
+    def add_success_mock_response():
+        responses.add(
+            responses.POST,
+            TestGenerateToken.URL + TestGenerateToken.ENDPOINT,
+            status=200,
+            json={"auth_token": TestGenerateToken.TOKEN_TEXT},
+        )
+
+    @staticmethod
+    def add_badlogin_mock_response():
+        responses.add(
+            responses.POST,
+            TestGenerateToken.URL + TestGenerateToken.ENDPOINT,
+            status=403,
+            json={"message": "Bad login"},
+        )
+
+    @staticmethod
+    def add_connectionerr_mock_response():
+        responses.add(
+            responses.POST,
+            TestGenerateToken.URL + TestGenerateToken.ENDPOINT,
+            body=requests.exceptions.ConnectionError(
+                "<urllib3.connection.HTTPConnection object at 0x1080854c0>: "
+                "Failed to establish a new connection: [Errno 8] "
+                "nodename nor servname provided, or not known"
+            ),
+        )
+
+    @staticmethod
+    @responses.activate
+    def test_help(pytestconfig):
+        runner = CliRunner(mix_stderr=False)
+        result = runner.invoke(generate_token.main, ["--help"])
+        assert result.exit_code == 0
+        assert str(result.stdout).startswith("Usage:")
+        assert not result.stderr_bytes
+
+    @staticmethod
+    @responses.activate
+    def test_args_both(valid_config, pytestconfig):
+        TestGenerateToken.add_success_mock_response()
+        runner = CliRunner(mix_stderr=False)
+        result = runner.invoke(
+            generate_token.main,
+            args=[
+                TestGenerateToken.USER_SWITCH,
+                TestGenerateToken.USER_TEXT,
+                TestGenerateToken.PSWD_SWITCH,
+                TestGenerateToken.PSWD_TEXT,
+            ],
+        )
+        assert result.exit_code == 0
+        assert result.stdout == f"{TestGenerateToken.TOKEN_TEXT}\n"
+        assert not result.stderr_bytes
+
+    @staticmethod
+    @responses.activate
+    def test_args_username(valid_config, pytestconfig):
+        TestGenerateToken.add_success_mock_response()
+        runner = CliRunner(mix_stderr=False)
+        result = runner.invoke(
+            generate_token.main,
+            args=[TestGenerateToken.USER_SWITCH, TestGenerateToken.USER_TEXT],
+            input=f"{TestGenerateToken.PSWD_TEXT}\n",
+        )
+        assert result.exit_code == 0
+        assert (
+            result.stdout
+            == f"{TestGenerateToken.PSWD_PROMPT}\n"
+            + f"{TestGenerateToken.TOKEN_TEXT}\n"
+        )
+        assert not result.stderr_bytes
+
+    @staticmethod
+    @responses.activate
+    def test_args_password(valid_config, pytestconfig):
+        TestGenerateToken.add_success_mock_response()
+        runner = CliRunner(mix_stderr=False)
+        result = runner.invoke(
+            generate_token.main,
+            args=[TestGenerateToken.PSWD_SWITCH, TestGenerateToken.PSWD_TEXT],
+            input=f"{TestGenerateToken.USER_TEXT}\n",
+        )
+        assert result.exit_code == 0
+        assert (
+            result.stdout
+            == f"{TestGenerateToken.USER_PROMPT}{TestGenerateToken.USER_TEXT}\n"
+            + f"{TestGenerateToken.TOKEN_TEXT}\n"
+        )
+        assert not result.stderr_bytes
+
+    @staticmethod
+    @responses.activate
+    def test_args_none(valid_config, pytestconfig):
+        TestGenerateToken.add_success_mock_response()
+        runner = CliRunner(mix_stderr=False)
+        result = runner.invoke(
+            generate_token.main,
+            args=[],
+            input=f"{TestGenerateToken.USER_TEXT}\n{TestGenerateToken.PSWD_TEXT}\n",
+        )
+        assert result.exit_code == 0
+        assert (
+            result.stdout
+            == f"{TestGenerateToken.USER_PROMPT}{TestGenerateToken.USER_TEXT}\n"
+            + f"{TestGenerateToken.PSWD_PROMPT}\n"
+            + f"{TestGenerateToken.TOKEN_TEXT}\n"
+        )
+        assert not result.stderr_bytes
+
+    @staticmethod
+    @responses.activate
+    def test_bad_login(valid_config, pytestconfig):
+        TestGenerateToken.add_badlogin_mock_response()
+        runner = CliRunner(mix_stderr=False)
+        result = runner.invoke(
+            generate_token.main,
+            args=[
+                TestGenerateToken.USER_SWITCH,
+                TestGenerateToken.USER_TEXT,
+                TestGenerateToken.PSWD_SWITCH,
+                TestGenerateToken.PSWD_TEXT,
+            ],
+        )
+        assert result.exit_code == 1
+        assert not result.stdout
+        assert result.stderr == "Bad login\n"
+
+    @staticmethod
+    @responses.activate
+    def test_connection_failed(valid_config, pytestconfig):
+
+        TestGenerateToken.add_connectionerr_mock_response()
+        runner = CliRunner(mix_stderr=False)
+        result = runner.invoke(
+            generate_token.main,
+            args=[
+                TestGenerateToken.USER_SWITCH,
+                TestGenerateToken.USER_TEXT,
+                TestGenerateToken.PSWD_SWITCH,
+                TestGenerateToken.PSWD_TEXT,
+            ],
+        )
+        assert result.exit_code == 1
+        assert not result.stdout
+        assert str(result.stderr).find("Failed to establish a new connection") != -1
--- a/lib/pbench/cli/agent/__init__.py
+++ b/lib/pbench/cli/agent/__init__.py
@@ -4,7 +4,7 @@ invocation. The CliContext will keep track of passed parameters,
 what command created it, which resources need to be cleaned up,
 and etc.
 
-We create an empty object at the begining and populate the object
+We create an empty object at the beginning and populate the object
 with configuration, group names, at the beginning of the agent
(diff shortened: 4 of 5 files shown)


lib/pbench/cli/agent/commands/generate_token.py
```diff
@@ -0,0 +1,71 @@
...
+    def __init__(self, context):
+        super().__init__(context)
+
+    def execute(self):
+        """Generate a token used to access the pbench server RESTful API"""
+
+        # FIXME:  Should we be logging anything here?
+        #
+        #  self.logger.warn(
+        #     "%s deprecated, will be removed in future release in favor of pbench-clear-results",
```
portante, line 22: Seems unnecessary?

portante (commented):
Great start.  Is there a PR with the server side API changes as well?

riya-17 (commented):
Looks Good

.gitignore
```diff
@@ -20,3 +20,4 @@ pip-wheel-metadata
 .vscode
 Pipfile
 Pipfile.lock
+venv/
```
dbutenhof, line 23: I don't think there's any likely harm in essentially globally reserving "venv" throughout the tree since nobody's likely to use it casually; but maybe this should be anchored to the base directory as `/venv/` just in case? (This could probably apply to most of the others too.) I did that, for example, with `/.env`, but didn't try to retrofit the other entries.

And we _do_ already have `.venv` and we probably don't want to spawn too many variations.

lib/pbench/cli/agent/commands/generate_token.py
dbutenhof: Funny, but the whole commented logger entry from a completely different context bothers me more than the thought of whether we should log here.

I agree with Pete; we should probably log any errors or unusual behavior, but I'm not sure I see any need to log every attempt at generating a PUT token.

lib/pbench/cli/agent/commands/generate_token.py
```diff
@@ -0,0 +1,71 @@
...
+        #     self.name,
+        # )
+
+        server = self.config.agent.parser.get("results", "server_rest_url")
+        headers = {"Content-Type": "application/json", "Authorization": None}
```
dbutenhof, line 27: I think I'd rather omit the `Authorization` header for login rather than supplying one but omitting the value. That just leads to questions about how that's handled in the code that make me want to go re-read Nikhil's login for no real purpose in this context. Let's try to avoid red herrings like that.

And if we're going to be "properly proper" here, maybe an `Accept: application/json` would be good; our current APIs will do that regardless, but establishing good habits is rarely a bad thing.

lib/pbench/cli/agent/commands/generate_token.py
```diff
@@ -0,0 +1,71 @@
...
+        payload = {
+            "username": self.context.username,
+            "password": self.context.password,
+            "token_duration": self.context.token_duration
```
dbutenhof, line 31: `"token_duration"` isn't a parameter of Nikhil's login code. Did you change that? I don't think we want users to specify their own token duration, even if we're bounding it with our own limitations. I think we probably do want to add a "role" request in some form, and here we'd be asking for "PUT-only", and the role might potentially imply a duration (with more restricted roles potentially having a longer duration).

lib/pbench/cli/agent/commands/generate_token.py
```diff
@@ -0,0 +1,71 @@
...
+        }
+        try:
+            response = requests.post(
+                f"{server}/login", headers=headers, json=payload
+            )
+        except requests.exceptions.ConnectionError as exc:
+            raise exc.args[0].reason
+
+        payload = json.loads(response.text)
```
dbutenhof, line 40: You can use `response.json()` here instead of calling the `json` module explicitly, which feels a little cleaner.

lib/pbench/cli/agent/commands/generate_token.py
```diff
@@ -0,0 +1,71 @@
...
+        if response.status_code == 200:
```
dbutenhof, line 41: We don't really expect any success but 200, however this is one of the reasons I decided to use `response.raise_for_status()` in my Elasticsearch queries; `requests` follows the HTTP status RFC in determining "success" or "failure" and raises appropriate exceptions for errors. Modularity-wise, I like deferring that decision to "the designated expert" instead of duplicating the logic through our code... especially with simplifyinginterpretations like "200 is success" that might trip us up later.

lib/pbench/cli/agent/options.py
```diff
@@ -19,13 +19,13 @@ def callback(ctx, param, value):
...
-        type=click.Path(exists=True),
+        type=click.Path(exists=True, readable=True),
         callback=callback,
         expose_value=False,
         help=(
-            "Path to a pbench-agent config. If provided pbench will load "
-            "this config file first. By default is looking for config in "
-            "'_PBENCH_AGENT_CONFIG' envrionment variable."
+            "Path to a pbench-agent configuration file (may be specified "
+            "via the '_PBENCH_AGENT_CONFIG' environment variable, instead)"
```
dbutenhof, line 29: I like "defaults to", which I think more clearly implies that the environment variable only matters if you don't use `--config|-C`.

dbutenhof (commented):
Cool. Of course I have a few comments, questions, and suggestions ...

dbutenhof:
Oh; and you have a formatting problem in your new code:

```
black --check ...
[1mwould reformat /home/pbench/lib/pbench/cli/agent/commands/generate_token.py[0m
[1mOh no! 💥 💔 💥[0m
[1m1 file would be reformatted[0m, 114 files would be left unchanged.[0m
FAILED: black --check .
```

You should install and run `black` ... either use `black --diff .` (at the base of your tree) to find and report errors and then fix them manually, or just `black .` and let it reformat them for you. (You might as well let it have its way, since your PR build will fail if it's unhappy anyway.)

You should also routinely run `flake8 .`, which is also mandatory.

lib/pbench/cli/agent/commands/generate_token.py
webbnh: I believe that we are going to want to allow the user of this tool to be able to specify the token duration.  Obviously, it will be up to the server to decide what to do about the request (e.g., impose a cap, ignore it altogether, etc.), so I figured it wouldn't hurt (much) to go ahead an implement it in the client end.  (That is, I fully expect that specifying this value will have no effect initially.)  Given that the server doesn't actually _require_ a token, yet, we're in a realm of "total flexibility"; as we converge upon the final implementation, I'm hoping we can fold flexible duration into the mix.

AWS S3 offers capability analogous to this, allowing the client to specify how long a pre-signed URL is good for.  I can envision use cases where the client needs only minutes, and I can envision other cases where the client will want the token to be good for as long as the server will tolerate -- I don't see any reason why we cannot and should not allow the user to choose.

Your point about allowing the request to specify a role is an excellent one!  Once we have a better handle on what roles look like, we can extend this tool.  (We had originally envisioned this tool's request as receiving a long-duration `PUT`-only token, but we currently have no way to effect that, without adding to the server API, and I would rather have that extension provide support for roles rather than custom support for move-results.)  For now, I think that we should have this tool execute a standard login and receive a standard token with a standard duration, with the server simply ignoring the request for a different token duration.

lib/pbench/cli/agent/commands/generate_token.py
webbnh: Nice!  Done.

lib/pbench/cli/agent/commands/generate_token.py
webbnh: I concur, although I'm not sure whether, in general `response` is a better expert than the calling code...nevertheless, in _this_ case, I think it is, so I've replaced the test with `response.ok`.

lib/pbench/cli/agent/options.py
webbnh: Tweaked.

webbnh:
> you have a formatting problem in your new code

Not a surprise.  :-D

I wasn't sure what tools would be run for me (e.g., LGTM) vs. ones I needed to run myself.

While we're on the topic...what does 

> continuous-integration/jenkins/pr-merge — This commit cannot be built

mean, and what do I need to do to address that?

lib/pbench/cli/agent/commands/generate_token.py
npalaska, line 27: The `login` does not require an Authorization header

lib/pbench/cli/agent/commands/generate_token.py
npalaska: yeah `token_duration` is not a part of user defined parameter only username and password

lib/pbench/cli/agent/commands/generate_token.py
```diff
@@ -0,0 +1,71 @@
...
+
+@click.command()
+@common_options
+@click.option('--username', prompt=True, required=True,
+              help='pbench server account username '
+                   '(will prompt if unspecified)')
+@click.option('--password', prompt=True, hide_input=True, required=True,
+              help='pbench server account password '
+                   '(will prompt if unspecified)')
+@click.option('--token-duration', type=int, required=False, default=3600,
```
npalaska, line 57: As things stand right now, the token duration is fixated on the server side and we specify it in the server config file

npalaska (commented):
This looks good, we just need to consider the use of `token_duration`

dbutenhof:
> While we're on the topic...what does
> 
> > continuous-integration/jenkins/pr-merge — This commit cannot be built
> 
> mean, and what do I need to do to address that?

Jenkins is running the same container environment we were just talking about. If you open the "Details" link you'll find a console log on the left edge of the Jenkins job page, so you can see the full text output. The `black` report I quoted was taken from there, and it's the reason your "commit cannot be built".

.gitignore
dbutenhof: Most of them I suspect should be anchored, but there's always the chance that some will occur at lower levels of the directory tree. Leaving them unanchored is "flexible" but risks "real files" being ignored by git unintentionally. I'm not sure that's a realistic concern with any of the names we're excluding at this point, but going forward I'd prefer that we try to make new patterns as specific as appropriate rather than just leave them unanchored because it's easy.

lib/pbench/cli/agent/commands/generate_token.py
dbutenhof: I'm not sure I can ever see someone realistically saying "I know I won't want to upload data after more than 10 minutes so let's limit the PUT token I'm using for my cluster deployment tests to 10 minutes". I can't imagine anyone specifying less than the maximum, really, because "things happen" and going back to the well for a second bucket in the middle of putting out the fire is a pain nobody really wants.

But at the very least, I'd prefer to see a TODO comment here explaining that this is currently meaningless but is added as a stake in the ground for possible later experimentation. (Though I think it's just as easy to remove it now and put it or something else in later when we figure out what that is. An issue to record current thoughts on that future seems like a more effective substitute for non-operative code.)

lib/pbench/cli/agent/commands/generate_token.py
webbnh: Yeah, line 22 is just a remnant from my cookie-cuttering this module from something else.  I would have removed it, but I thought that we might be wanting to log _something_...maybe.

As the code stands, I don't think it currently detects "unusual behavior".  I'll grant that maybe we don't want to log successful logins (that's an audit function that we aren't currently pursuing).  Do we actually want to log failures in the tool?

So, I'm thinking there is nothing to log here.  I'm thinking that any logging would/should be done in the server, instead.  Anyone disagree?

webbnh:
> If you open the "Details" link you'll find a console log on the left edge of the Jenkins job page, so you can see the full text output.

Thanks, @dbutenhof.  Aside from the Black complaint, there is also this:

> ERROR: InvocationError for command /usr/bin/bash -c './jenkins/run-pytests /var/tmp/tox/jenkins-pytests' (exited with code 1)

Do I need to do something to address that?

lib/pbench/cli/agent/commands/generate_token.py
webbnh: Fixed.

lib/pbench/cli/agent/commands/generate_token.py
webbnh: Agreed.  (When the time is right, we can consider changing that, but I don't think we need to until we have some sort of roles implementation.)

lib/pbench/cli/agent/commands/generate_token.py
webbnh: If the server had a broader exposure than the RD intranet, I suspect a user wouldn't want their tokens to last any longer than needed; on the flip side, I can easily imagine that different users would have different requirements for how long they need their tokens to live.  So, I "one size" solution is unlikely to fit "all".

I can certainly add a TODO and open an issue.  While it's not hard to remove and re-add this capability, I'm inclined to leave it against the day when the server starts supporting it.  (I really don't think that the tokens intended to support human use of the dashboard are going to work very well for scripted transfer of benchmark results.)

webbnh: `TODO` added; issue #2139 opened.

.gitignore
webbnh: I anchored the ones which were obvious from my checkout.

lib/pbench/cli/agent/commands/generate_token.py
dbutenhof: The function here is pretty simple. While I did add some logging in my `pbench-state-manager.py` tool (which is mostly for the unit tests), the way I handled errors was to simply write the `str()` of any exception to `stderr` as a failure message and exit with `1`. That might well be good enough here, too, even though this is more user-facing.

dbutenhof:
> > If you open the "Details" link you'll find a console log on the left edge of the Jenkins job page, so you can see the full text output.
> 
> Thanks, @dbutenhof. Aside from the Black complaint, there is also this:
> 
> > ERROR: InvocationError for command /usr/bin/bash -c './jenkins/run-pytests /var/tmp/tox/jenkins-pytests' (exited with code 1)
> 
> Do I need to do something to address that?

Pretty sure that's just because due to the `black` error, the script exited with failure. I didn't see anything else that stood out.

.flake8
```diff
@@ -1,6 +1,6 @@
 [flake8]
 doctests = True
-exclude = .venv,.git,.tox,dist,doc,*lib/python*,*egg,build,*web-server/*,
+exclude = .venv,venv,.git,.tox,dist,doc,*lib/python*,*egg,build,*web-server/*,
```
portante, line 3: Can we pull this out into another PR?

.gitignore
```diff
@@ -1,4 +1,4 @@
-doc/*.html
+/doc/*.html
```
portante, line 1: Can we pull these changes out into another PR?

portante (changes requested):

.flake8
webbnh: Certainly....

.gitignore
webbnh: ...do you want them in the same PR as each other?

lib/pbench/cli/agent/commands/generate_token.py
```diff
@@ -0,0 +1,78 @@
...
+import requests
+
+from pbench.agent.base import BaseCommand
+from pbench.cli.agent import pass_cli_context
+from pbench.cli.agent.options import common_options
+
+
+class GenerateToken(BaseCommand):
+    def __init__(self, context):
+        super().__init__(context)
```
portante, line 13: It is not clear you need this.

portante (changes requested):
Looking good ...

.flake8
webbnh: Instead of pulling this change to a new PR, I rejiggered my work environment to use `.venv` instead of `venv`.

.gitignore
webbnh: Since I don't need to change this file to match my work environment, I'm rescinding all of these changes.

.gitignore
webbnh: I opted instead to change my work environment rather than changing this file, so I've rescinded these changes.

lib/pbench/cli/agent/commands/generate_token.py
webbnh: I was using another of Click-based `pbench` command as a template, so I just did what that one did...and blindly followed inheritance (in both senses).  We _do_ want the `PbenchAgentContext` set up (I think...), but I suspect that you are right that the superclass is doing more than we need.  I have a work item queued to rationalize all of the command line tool infrastructure -- this should get shaken out through that.

lib/pbench/test/unit/agent/task/test_generate_token.py
```diff
@@ -0,0 +1,167 @@
...
+
+class TestGenerateToken:
+
+    USER_SWITCH = "--username"
+    PSWD_SWITCH = "--password"
+    USER_PROMPT = "Username: "
+    PSWD_PROMPT = "Password: "
+    USER_TEXT = "test_user"
+    PSWD_TEXT = "password"
+    TOKEN_TEXT = "what is a token but 139 characters of gibberish"
```
dbutenhof, line 16: Therefore, by definition, this `TOKEN_TEXT` is not a token???

lib/pbench/test/unit/agent/task/test_generate_token.py
```diff
@@ -0,0 +1,167 @@
...
+            status=403,
+            json={"message": "Bad login"},
+        )
+
+    @staticmethod
+    def add_connectionerr_mock_response():
+        responses.add(
+            responses.POST,
+            TestGenerateToken.URL + TestGenerateToken.ENDPOINT,
+            body=requests.exceptions.ConnectionError(
```
dbutenhof, line 43: But no linked `reason` exception?

lib/pbench/cli/agent/options.py
```diff
@@ -19,13 +19,13 @@ def callback(ctx, param, value):
     return click.option(
         "-C",
         "--config",
+        required=True,
```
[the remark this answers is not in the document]
dbutenhof: Indeed; I lost my context and somehow thought I was reviewing new code.

dbutenhof (approved):

lib/pbench/test/unit/agent/task/test_generate_token.py
webbnh: I thought about trying to come up with a phrase containing exactly 139 characters, but I found something better to do....

[If your comment was something other than a jest, please un-resolve this "conversation" and elaborate.  :-) ]

lib/pbench/test/unit/agent/task/test_generate_token.py
webbnh: Yeah...I thought about trying to dive into the nuances of the construction of `requests.exceptions.ConnectionError` -- I did give it a reasonable try -- but after awhile it seemed like it was more effort than was required for a mock.

The escapade did, however, lead to the to conclusion that part of the error reporting in the tool code around dealing with exceptions from the request was somewhat fragile (it is more robust, now), so it was worthwhile in that respect.

portante (approved):
Awesome!

portante:
@webbnh, we missed one problem with this merged code: when building an RPM we don't include the new command.  We need to add it for the RPM in `agent/Makefile`.

webbnh:
@portante, my apologies -- clearly my testing regimen is incomplete.

However, Click is meant to be self-discovering, so having the tools installed via a static Makefile seems like it might be swimming upstream.  We should discuss how we want to move forward for the Click-based tools.  (For testing, I installed it via `pip install --editable .`, so I imagine that there is some similar command that we would want the Makefile to run, rather than doing it via an explicit or bulk copy operation.)

webbnh:
Opened #2146 to track this.

portante:
> so I imagine ...

You imagine correctly ... we already do that in the Makefile.  But the problem is that RPM building needs the final files all tar'd up into a RPM source tar ball.

PR #2140, opened 2021-03-18
Ignore annotated coverage files

portante:


.gitignore
```diff
@@ -20,3 +20,4 @@ pip-wheel-metadata
 .vscode
 Pipfile
 Pipfile.lock
+*,cover
```
dbutenhof, line 23: Do these show up anywhere, or should they be anchored to the base directory as `/*,cover`??

dbutenhof (approved):

.gitignore
portante: They show up where the `.py` is when you ask for them using `annotate` command.

dbutenhof: Interesting; I always just pull up the annotated html version in the browser, but that could be convenient. I'll have to try it.

dbutenhof: Ah; OK, that's nice to know, but I like the html version better. (Probably just the color, though!)

image

vs 

image

ndokos (approved):

PR #2141, opened 2021-03-18
Update `pbench-trafficgen` script usage

pradiptapks:
**Update:**

1. Removed the existing usage details from the `pbench-trafficgen` script as it not flexible to upgrade the changes.
2. Created a separate `usage` directory with `pbench-trafficgen-usage.txt`.
3. Updated the latest supported trafficgen options using common examples.
4. Update the `usage` function in pbench-trafficgen to script to use `pbench-trafficgen-usage.txt`.
5. The idea is to create a seprate usage file to maintain the usage details separately when the trafficgen options get enhanced.


portante (changes requested):
In general this drops the ability to change defaults from showing up in the usage text.  Have you considered using something like Jinja2 to render the usage from a template?

webbnh:
Having the usage content in a file separate from the sources makes me uncomfortable:  it's too easy for the file to get lost, damaged, or mismatched (to say nothing of the downsides of making the contents static).

Nevertheless, I'm sympathetic to the problem of having one tool attempting to document the interface to another.  Perhaps there is some way that pbench could invoke trafficgen to obtain the usage information for the currently-installed version and then incorporate that into its own usage?  Alternatively, the pbench usage output could instead include instructions on how the user could obtain that information.

portante:
Perhaps the problem with the `pbench-trafficgen` usage is that it needs to know directly about some of the parameters, and not others.  So it has many of the parameters duplicated from trafficgen itself.

There was an attempt to have `pbench-run-benchmark` avoid that, but that script is not well supported, as the folks who wrote that have moved on.

Can we enumerate what parameters have changed?  And then perhaps use the typical `--` switch to indicate arbitrary parameters to be passed directly to the trafficgen?

webbnh:
> Removed the existing usage details from the pbench-trafficgen script as it not flexible to upgrade the changes.

Unfortunately, the result of this change isn't really any more flexible than the previous -- it just creates a different file to edit when there are changes.

What we need to do is to decouple the wrapper script from the tool.  To the extent that the wrapper doesn't need to intercept or override any of the options to the tool, Peter's suggestion of using `--` to separate wrapper options from tool options is an excellent idea.  But, that doesn't address the change proposed in this PR.  To move @pradiptapks change forward, I think we need to remove the `trafficgen` options from the `pbench-trafficgen` usage text (and revert the extraction of them to a separate file).  If that seems insufficiently friendly, we can add some text which describes to the user how to get `trafficgen` to offer its own usage information.

portante:
TrafficGen is no longer supported starting in v0.71 and later.

PR #2143, opened 2021-03-18
Remove use of podman by Tool Meister subsystem

portante:
Instead of having the Tool Data Sink leverage `podman` for running a `prometheus` container, we assume that the `prometheus2` RPM [1] and PCP RPMs are installed along side the `pbench-agent`.

For non-containerized environments, while the pbench-agent RPM does not have a direct dependency on the `prometheus2` or PCP RPMs, the Tool Meister infrastructure will only try to use it if a persistent tool is registered requiring it.

The "tools" container image builds the `prometheus2`, `node_exporter`, and PCP RPMs into the image to provide that functionality.

This prevents the pulling images at Tool Meister start-up so that our callers can avoid unwanted / unexpected behaviors in their benchmarking environments.

For PCP, removing the use of a `systemd` based PCP container means we have to now handle creating the N `pmlogger` processes, one for every `pmcd` executed by Tool Meisters.  We also have to ensure the pmcds are running before invoking `pmlogger` by using the `pmcd_wait` command.

We provide an updated "pmcd" and "pmlogger" configuration files that have been tested by hand to show they work even in non-root user environments.

The `pmproxy` service is now listening on port `44566` instead of the default of `44322`.

We also fix the `pmlogger` command to hard-code a 3 second interval for now.

The Tool Meister PCP tool needs to know the installation directory in order to find the proper templates to use with `pmcd`.

[1] distributed-system-analysis/prometheus-rpm#1

----

"Correct the working directory for all tools"

We modify the test harness to emit a `<tool>.file` in the current working directory when the tool runs to help us detect when we don't have the execution environment correct.  In live testing, lots of files were being littered in random places.

This led to the discovery that the mock'd `dcgm_prometheus.py` file was not correct.

We also cleaned up the mocks we had in place to reduce the number of individual mocks and remove symlinks to the `bench-scripts` tree.

This work also addresses the working directory used by remote Tool Meisters.

----

"Handle tool shutdown without finite limits"

In case the tools we use are not well behaved, the Tool Meisters need to terminate the sub-processes they create after a short period of time.

We setup a hard-coded 30 maximum timeout for any tool, persistent or transient, to shut down, emitting informational messages every 5 seconds while waiting.  After 30 seconds, the sub-process is "killed" and an error message is emitted.

We also lower the time out in the unit tests since the commands should not take very long at all.

----

Add `firewalld` service files for `node_exporter`, `pmcd`, `prometheus`, `redis`, and the Tool Data Sink.


portante (commented):
Looks ready for review.  This commit is the first fundamental commit originally from PR #2076.

portante:
Well now it is ready.  We need the addition five commits to deal with problems found while testing.

dbutenhof (dismissed):
I think this looks OK. I have some concerns about peeling off each commit into a separate PR, though. This is a great way to do development, one small slice at a time, and we haven't been very good at that. However, taking a large body of code that's been reviewed and stripping off pieces that need to be individually reviewed _again_ isn't really an advantage to anyone. It's more work, not less. Maybe we can just finish this up and resolve to try to be a bit better in the future? ;-)

portante:
> I think this looks OK. I have some concerns about peeling off each commit into a separate PR, though. This is a great way to do development, one small slice at a time, and we haven't been very good at that. However, taking a large body of code that's been reviewed and stripping off pieces that need to be individually reviewed _again_ isn't really an advantage to anyone. It's more work, not less. Maybe we can just finish this up and resolve to try to be a bit better in the future? ;-)

It would be great just merge all of PR #2076 as is.  Is that what you are asking?  Has it been reviewed by everyone, all 20+ commits?  If so, great.

Maxusmusti (dismissed):
Tested with same setup as #2005, everything functioned as expected

portante:
Rebased, just need 2 thumbs up and we can merge.

dbutenhof (dismissed):

Maxusmusti (dismissed):
thumb up

portante:
Merged after squashing commits.  Thanks all!

Issue #2130, opened 2021-03-04
On logout we should return the success irrespective of internal failures 


npalaska:
I'm inclined to think that, from a security perspective (and, to some extent from a UX perspective), we want to make it smooth and easy to log out.  Conversely, being picky about logging out risks leaking security state and information.  So, if we get a logout request, I think the response (to the user) should almost always be "success".  If we want to log something else internally, that's fine, but sending errors on logout (for anything other than the case(s) where the logout fails and the user is still logged in...), doesn't serve the user very well.

_Originally posted by @webbnh in https://github.com/distributed-system-analysis/pbench/pull/1937#discussion_r586834770_

PR #2144, opened 2021-03-18
Fix minor bug in the login workflow

npalaska:
Right now, if the authorization header is present during the login request (An attempt at creating another login token) we check the generated auth token with the one in the header. This check is redundant since we already check for the `integrityError` on authorization tokens. 
This PR: 
- Removes the redundant auth token check when Authorization header is present during login. 
- Unit test for an attempt at updating the auth_token externally and confirmed that auth_token can not be updated externally.

### DIFF
--- a/lib/pbench/server/api/resources/users_api.py
+++ b/lib/pbench/server/api/resources/users_api.py
@@ -35,11 +35,12 @@ class RegisterUser(Resource):
             Content-Type:   application/json
             Accept:         application/json
 
-        :return: JSON Payload
-        if we succeed to add a user entry in the database, the returned response_object will look like the following:
-            response_object = {
-                    "message": "Successfully registered."/"failure message",
-                }
+        :return:
+            Success: 201 with empty payload
+            Failure: <status_Code>,
+                    response_object = {
+                        "message": "failure message"
+                    }
         To get the auth token user has to perform the login action
         """
         # get the post data
@@ -114,13 +115,7 @@ class RegisterUser(Resource):
             self.logger.info(
                 "New user registered, username: {}, email: {}", username, email_id
             )
-
-            response_object = {
-                "message": "Successfully registered.",
-            }
-            response = jsonify(response_object)
-            response.status_code = 201
-            return make_response(response, 201)
+            return "", 201
         except EmailNotValidError:
             self.logger.warning("Invalid email {}", email_id)
             abort(400, message=f"Invalid email: {email_id}")
@@ -142,7 +137,6 @@ class Login(Resource):
             "pbench-server", "token_expiration_duration"
         )
 
-    @Auth.token_auth.login_required(optional=True)
     def post(self):
         """
         Post request for logging in user.
@@ -160,11 +154,15 @@ class Login(Resource):
             Accept:         application/json
 
         :return: JSON Payload
-        if we succeed to decrypt the password hash, the returned response_object will include the auth_token
-            response_object = {
-                    "message": "Successfully logged in."/"failure message",
-                    "auth_token": "<authorization_token>", # Will not present if failed
-                }
+            Success: 200,
+                    response_object = {
+                        "auth_token": "<authorization_token>"
+                        "username": <username>
+                    }
+            Failure: <status_Code>,
+                    response_object = {
+                        "message": "failure message"
+                    }
         """
         # get the post data
         post_data = request.get_json()
@@ -217,17 +215,6 @@ class Login(Resource):
                 500, message="INTERNAL ERROR",
             )
 
-        # check if the user is already logged in, in case the request has a an authorization header included
-        # We would still proceed to login and return a new auth token to the user
-        if user == self.auth.token_auth.current_user():
-            self.logger.warning("User already logged in and trying to re-login")
-            if self.auth.token_auth.get_auth()["token"] == auth_token:
-                # If a user attempts to re-login immediately within a millisecond,
-                # the new auth token will be same as the one present in the header
-                # Therefore we will not fulfill the re-login request
-                self.logger.info("Attempted immediate re-login")
-                abort(403, message="Retry login after some time")
-
         # Add the new auth token to the database for later access
         try:
             token = ActiveTokens(token=auth_token)
@@ -250,7 +237,6 @@ class Login(Resource):
             abort(500, message="INTERNAL ERROR")
 
         response_object = {
-            "message": "Successfully logged in.",
             "auth_token": auth_token,
             "username": username,
         }
@@ -276,10 +262,12 @@ class Logout(Resource):
         Required headers include
             Authorization:   Bearer <Pbench authentication token (user received upon login)>
 
-        :return: JSON Payload
-            response_object = {
-                    "message": "Successfully logged out."/"failure message",
-                }
+        :return:
+            Success: 200 with empty payload
+            Failure: <status_Code>,
+                    response_object = {
+                        "message": "failure message"
+                    }
         """
         auth_token = self.auth.token_auth.get_auth()["token"]
         user = self.auth.token_auth.current_user()
@@ -293,10 +281,7 @@ class Logout(Resource):
             self.logger.exception("Exception occurred during deleting the user entry")
             abort(500, message="INTERNAL ERROR")
 
-        response_object = {
-            "message": "Successfully logged out.",
-        }
-        return make_response(jsonify(response_object), 200)
+        return "", 200
 
 
 class UserAPI(Resource):
@@ -321,15 +306,17 @@ class UserAPI(Resource):
             Authorization:  Bearer Pbench_auth_token (user received upon login)
 
         :return: JSON Payload
-            response_object = {
-                    "message": "Success"/"failure message",
-                    "data": {
+            Success: 200,
+                    response_object = {
                         "username": <username>,
                         "first_name": <firstName>,
                         "last_name": <lastName>,
-                        "registered_on": registered_on,
+                        "registered_on": <registered_on>,
                     }
-                }
+            Failure: <status_Code>,
+                    response_object = {
+                        "message": "failure message"
+                    }
         """
         try:
             user, verified = self.auth.verify_user(username)
@@ -339,18 +326,12 @@ class UserAPI(Resource):
 
         # TODO: Check if the user has the right privileges
         if verified:
-            response_object = {
-                "status": "success",
-                "data": user.get_json(),
-            }
+            response_object = user.get_json()
             return make_response(jsonify(response_object), 200)
         elif user.is_admin():
             try:
                 target_user = User.query(username=username)
-                response_object = {
-                    "message": "success",
-                    "data": target_user.get_json(),
-                }
+                response_object = target_user.get_json()
                 return make_response(jsonify(response_object), 200)
             except Exception:
                 self.logger.exception(
@@ -359,7 +340,7 @@ class UserAPI(Resource):
                 abort(500, message="INTERNAL ERROR")
         else:
             self.logger.warning(
-                "User {} is not authorized to delete user {}.", user.username, username,
+                "User {} is not authorized to get user {}.", user.username, username,
             )
             abort(
                 403,
@@ -387,15 +368,17 @@ class UserAPI(Resource):
             Authorization:  Bearer Pbench_auth_token (user received upon login)
 
         :return: JSON Payload
-            response_object = {
-                    "message": "Success"/"failure message",
-                    "data": {
+            Success: 200,
+                    response_object = {
                         "username": <username>,
                         "first_name": <firstName>,
                         "last_name": <lastName>,
-                        "registered_on": registered_on,
+                        "registered_on": <registered_on>,
                     }
-                }
+            Failure: <status_Code>,
+                    response_object = {
+                        "message": "failure message"
+                    }
         """
         post_data = request.get_json()
         if not post_data:
@@ -445,10 +428,7 @@ class UserAPI(Resource):
             self.logger.exception("Exception occurred during updating user object")
             abort(500, message="INTERNAL ERROR")
 
-        response_object = {
-            "message": "success",
-            "data": user.get_json(),
-        }
+        response_object = user.get_json()
         return make_response(jsonify(response_object), 200)
 
     @Auth.token_auth.login_required()
@@ -463,10 +443,12 @@ class UserAPI(Resource):
             Accept:         application/json
             Authorization:   Bearer Pbench_auth_token (user received upon login)
 
-        :return: JSON Payload
-            response_object = {
-                    "message": "Successfully deleted."/"failure message",
-                }
+        :return:
+            Success: 200 with empty payload
+            Failure: <status_Code>,
+                    response_object = {
+                        "message": "failure message"
+                    }
         """
         try:
             user, verified = self.auth.verify_user(username)
@@ -498,7 +480,4 @@ class UserAPI(Resource):
                 abort(403, message="Admin user can not be deleted")
             self.logger.info("User entry deleted for user with username {}", username)
 
-        response_object = {
-            "message": "Successfully deleted.",
-        }
-        return make_response(jsonify(response_object), 200)
+        return "", 200
--- a/lib/pbench/test/unit/server/test_user_auth.py
+++ b/lib/pbench/test/unit/server/test_user_auth.py
@@ -42,8 +42,6 @@ class TestUserAuthentication:
                 email="user@domain.com",
                 password="12345",
             )
-            data = response.json
-            assert data["message"] == "Successfully registered."
             assert response.content_type == "application/json"
             assert response.status_code, 201
 
@@ -123,12 +121,10 @@ class TestUserAuthentication:
                 email="user@domain.com",
                 password="12345",
             )
-            data_register = resp_register.json
-            assert data_register["message"] == "Successfully registered."
+            assert resp_register.status_code == 201
             # registered user login
             response = login_user(client, server_config, "user", "12345")
             data = response.json
-            assert data["message"] == "Successfully logged in."
             assert data["auth_token"]
             assert data["username"] == "user"
             assert response.content_type == "application/json"
@@ -148,12 +144,11 @@ class TestUserAuthentication:
                 email="user@domain.com",
                 password="12345",
             )
-            data_register = resp_register.json
-            assert data_register["message"] == "Successfully registered."
+            assert resp_register.status_code == 201
+
             # registered user login
             response = login_user(client, server_config, "user", "12345")
             data = response.json
-            assert data["message"] == "Successfully logged in."
             assert data["auth_token"]
             assert response.content_type == "application/json"
             assert response.status_code == 200
@@ -165,8 +160,6 @@ class TestUserAuthentication:
                 headers=dict(Authorization="Bearer " + data["auth_token"]),
                 json={"username": "user", "password": "12345"},
             )
-            data = response.json
-            assert data["message"] == "Successfully logged in."
             assert response.status_code == 200
 
             # Re-login without auth header
@@ -175,8 +168,6 @@ class TestUserAuthentication:
                 f"{server_config.rest_uri}/login",
                 json={"username": "user", "password": "12345"},
             )
-            data = response.json
-            assert data["message"] == "Successfully logged in."
             assert response.status_code == 200
 
     @staticmethod
@@ -193,8 +184,8 @@ class TestUserAuthentication:
                 email="user@domain.com",
                 password="12345",
             )
-            data_register = resp_register.json
-            assert data_register["message"] == "Successfully registered."
+            assert resp_register.status_code == 201
+
             # registered user login
             response = login_user(client, server_config, "user", "123456")
             data = response.json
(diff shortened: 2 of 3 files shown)


lib/pbench/server/api/resources/users_api.py
```diff
@@ -217,16 +217,13 @@ def post(self):
...
-        if user == self.auth.token_auth.current_user():
-            self.logger.warning("User already logged in and trying to re-login")
-            if self.auth.token_auth.get_auth()["token"] == auth_token:
-                # If a user attempts to re-login immediately within a millisecond,
-                # the new auth token will be same as the one present in the header
-                # Therefore we will not fulfill the re-login request
-                self.logger.info("Attempted immediate re-login")
-                abort(403, message="Retry login after some time")
+        if self.auth.token_auth.current_user():
+            self.logger.warning(
```
portante, line 223: Does this really warrant a warning?  Perhaps just an informational message.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -217,16 +217,13 @@ def post(self):
...
+                "User already logged in and trying to re-login. Username: {}",
+                {user.username},
```
portante, line 225: Do you mean to make `user.username` into a `dict`?  Seems like you'd just want to use `user.username`.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -217,16 +217,13 @@ def post(self):
...
         # We would still proceed to login and return a new auth token to the user
-        if user == self.auth.token_auth.current_user():
-            self.logger.warning("User already logged in and trying to re-login")
-            if self.auth.token_auth.get_auth()["token"] == auth_token:
-                # If a user attempts to re-login immediately within a millisecond,
-                # the new auth token will be same as the one present in the header
-                # Therefore we will not fulfill the re-login request
-                self.logger.info("Attempted immediate re-login")
-                abort(403, message="Retry login after some time")
+        if self.auth.token_auth.current_user():
```
portante, line 222: Can you explain more about how this situation could arise?

Doesn't this mean the `POST /login HTTP/1.1` request would have an `Authorization` header on it?  If so, why wouldn't we just take that header and use that token once the username and password are verified?

lib/pbench/test/unit/server/test_user_auth.py
```diff
@@ -298,6 +298,33 @@ def test_update_user(client, server_config):
             assert response.status_code == 200
             assert data_login["message"] == "Successfully logged in."
 
+    @staticmethod
+    def test_external_token_update(client, server_config):
```
portante, line 302: Does this test cover the case where a warning would be issued?

portante (changes requested):

lib/pbench/server/api/resources/users_api.py
npalaska: Agree

lib/pbench/server/api/resources/users_api.py
npalaska: Ah, fixed

lib/pbench/server/api/resources/users_api.py
npalaska: Yes this situation would arise if `Authorization` header is present in the request. When you say use that token do you mean returning that same token? If so, yeah that's probably one way to handle it so we don't create another token if there already is valid token in header. However, we discussed earlier that we would just return the new token if the user ask for one assuming user knows what they are doing. Here is the link to the discussion https://github.com/distributed-system-analysis/pbench/pull/1937#discussion_r545920589

lib/pbench/test/unit/server/test_user_auth.py
npalaska: No, this one is for covering case where user is attempting to update the auth_token externally. There already is a separate test case for testing the re-login with auth header that is `test_user_relogin` function

portante: How does that test verify the new behavior here?

lib/pbench/server/api/resources/users_api.py
portante: Is it really the case that a user who knows what they are doing would ever re-login after logging in?  Wouldn't they have to somehow re-typing the URL and all that.  This seems like a case we really shouldn't be handling.

npalaska: Yeah this is just an edge case scenario, potential behavior from a user. 
There are 2 options: 
- We could return the same token 
- Just ignore the auth header (if present) and proceed with a request

Do you prefer the first one?

lib/pbench/test/unit/server/test_user_auth.py
npalaska: There isn't a new behavior per se in the login, I just removed the redundant auth token check when Authorization header is present during login request. 

This test is just proof check that we shouldn't be allowing the external auth_token update

lib/pbench/server/api/resources/users_api.py
dbutenhof: My take has always been that we should ignore the unnecessary Authorization header and simply generate a new token. That led to confusion because, if they do this in relatively quick succession, they could get a duplicate token.

My main concern is for testing/debugging, where it's common to run server commands though curl or Postman, or other test scripts; and less common to run a full workflow including logout. I would like to be able to handle these cases cleanly; I've never worked with a system that didn't.

The edge case that I think is unlikely is a workflow that repeats login without a logout _and_ presents previous Authorization tokens on later login attempts. I suspect that if that happens it's probably fine to return the Authorization token rather than trying to generate a new one. This won't help with two "unauthorized" login attempts in succession, and that's what I would like to support.

I don't want to query and return a live pre-existing token for the user; because during testing we might (hopefully will) run concurrent functional tests, probably under the same user, and each needs to have their own authorization session; we don't want the first one that logs out to invalidate the others. And even if they log in at the same time, they need unique authorization tokens that can each be stored and retrieved. We definitely need to support multiple concurrent authorization sessions from the same user.

lib/pbench/test/unit/server/test_user_auth.py
```diff
@@ -298,6 +298,33 @@ def test_update_user(client, server_config):
...
+                client,
+                server_config,
+                username="username",
+                firstname="firstname",
+                lastname="lastName",
+                email="user@domain.com",
+                password="12345",
+            )
+            data_register = resp_register.json
+            assert data_register["message"] == "Successfully registered."
```
dbutenhof, line 315: For the record, I still don't like including these `"message"` tags in successful responses. On errors, an error document with a more specific and informative message than the standard HTTP response code is good; I'd prefer that a successful response just be the resource we need to return without embellishment/complication. And saying "Successfully registered" doesn't add anything useful over the 200 status. Presumably the caller knows they were registering a user.

npalaska: Indeed but sometime we dont have any resource to return e.g register, logout, delete operations. In these cases should we just return the status code without anything else? (It does actually make sense because sending message string along with status code is a bit redundant). I initially did this to have consistent responses from all the operations because on login, get, put operations we return jsonified resource along with a status code so on operations that does not have anything to return just jsonify simple message and return along with a status code. However, I can certainly just do `return ' ', 201/200` if that make sense.

dbutenhof: Nothing says you need to return a body if there's nothing to return.

Normal REST convention for POST/PUT is to return the resource (in the body) or a reference to it (usually via `Location` header with a URI that you can `GET`). While that's not a bad thing to consider, I'm not sure it's critically important at this point. No reason for successful logout or a delete to return anything, because the HTTP status is enough.

lib/pbench/server/api/resources/users_api.py
webbnh: At the risk of muddying things, I'd like to come at this from a different angle.  The duration of these tokens is expected/intended to be relatively short, which means that a long-session user is going to need either a way to "refresh" a token or to "re-login" without, in either case, having logged out first.  So, from the server's perspective, what is this request going to look like?  (And, on the client side, what is going to trigger the refresh/relogin operation?)

When a user logs out, we can simply mark the token as being "expired in need of refresh".  If there happens to be a second session using the same username, then having the first session logged out will be no different than having had the second session's token expire naturally -- in either case, the client will need to refresh the token, and it should be able to handle that implicitly.

With that as the groundwork, I don't think it matters whether we hand out unique tokens for every login by a given user.  At our whim, we can keep several tokens going, or we can hand back the same one (perhaps with an updated expiration) every time.

(What am I missing, here?)

webbnh: Oh, one additional thought:  a login request without an Authorization token would be a "new session", while a login request _with_ a token could be a "refresh"...if the distinction is important.  If the distinction is not important, then, I'm with Dave, we should just ignore the token if it shows up on a login request.

webbnh (approved):

lib/pbench/server/api/resources/users_api.py
dbutenhof: That actually would be one way to "refresh" a session, although there's an intriguing article @aquibbaig discovered that instead recommends that login return a "role based" refresh token usable _only_ for refreshing a session, which when presented to a special refresh API will return a new authorization token. (Sounds complicated, still not 100% sure I'm at peace with this idea.)

In any case, this is all out of scope for our current user/authentication system which creates long-life tokens that can be used until (relatively far out) expiration or explicit deletion via logout.

webbnh: I believe that our tokens are [currently configured with a duration of 60 minutes](https://github.com/distributed-system-analysis/pbench/blob/e12844cadb6e8d7195f7db55943bba82e522e1f2/server/lib/config/pbench-server-default.cfg#L33) (which is longer than I expected, but too short for a serious session).

I haven't read Aquib's article yet (I started to look at it when he first pointed it out, but it looked long and I moved on to other things); nevertheless, I have to say that it's not immediately obvious how having a second token which allows you to renew the first is better than having just one....

npalaska: Yes @webbnh the value for the token expiration is really arbitrary right now. We haven't came up with a fixed and more reliable number yet but its definitely going to be lot longer than that. 

About the refresh token API, the article is interesting and after digging a little bit more I came across quite a few article that seem to be suggesting a similar approach but its indeed a little complicated to understand in one read. Anyway, this topic is out of the scope of this PR.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -114,13 +114,7 @@ def post(self):
...
                 "New user registered, username: {}, email: {}", username, email_id
             )
-
-            response_object = {
-                "message": "Successfully registered.",
-            }
-            response = jsonify(response_object)
-            response.status_code = 201
-            return make_response(response, 201)
+            return "", 201
```
dbutenhof, line 117: I continually debate how completely we should try to follow REST conventions; most of the rules are fairly loose.

However, while I'd noticed before that you were using 201 here, for some reason this change just made me think about it more carefully. It's normally strongly recommended that a 201 return a `Location` response header giving the URI the client can use to retrieve the new resource ... in this case the user. Some recommend returning the resource as payload (_as if_ they'd done a `GET` already), or a human-readable HTML description.

Maybe not something we need to worry about right here, but maybe we should consider using `Location` somewhere along here for resources we create?

lib/pbench/server/api/resources/users_api.py
dbutenhof: Yeah, we need to decide how we're going to manage authorization sessions with the dashboard and other potential future clients (including our own functional test environment). We don't have to decide that right now.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -220,9 +214,9 @@ def post(self):
         # Log if the request has an authorization header included and user is still trying to create a new login
         # We would still proceed to login and return a new auth token to the user
         if self.auth.token_auth.current_user():
-            self.logger.warning(
+            self.logger.info(
```
dbutenhof, line 217: Not a big deal; but, technically, the user could already be logged in and this wouldn't trigger ... unless they decide to present the previous login token. That makes the message wording potentially slightly misleading.

dbutenhof (approved):

lib/pbench/server/api/resources/users_api.py
```diff
@@ -114,13 +114,7 @@ def post(self):
             self.logger.info(
```
portante, line 114: Not that you have changed this here, but the logging and return should be outside the `try` block, or in an `else` clause.

lib/pbench/server/api/resources/users_api.py
portante: Seems like a good thing to track in an issue to follow up in another PR?

lib/pbench/server/api/resources/users_api.py
portante: Agreed.

And we need to know what an admin would do with this message.  Why does somebody need to know that they tried to re-login?

lib/pbench/server/api/resources/users_api.py
portante: Forgive me, but I still don't understand this situation.  Can you explain in the comment in the code how this can occur more clearly, and why we need to log an informational message?

portante (changes requested):

lib/pbench/server/api/resources/users_api.py
npalaska: Opened https://github.com/distributed-system-analysis/pbench/issues/2145

lib/pbench/server/api/resources/users_api.py
npalaska: Ah, I originally put this log message here when I was doing some functional testing on user authentications. I am not sure how useful this log statement would be when we have actual users. The log statement only gets invoked when the Authorization header is present in the request. If the user already has a valid token but not presented in the request, this log statement wont be invoked (so its a bit misleading). I can completely take out this message if we dont need it.

lib/pbench/server/api/resources/users_api.py
dbutenhof: We definitely do not want to make a query to see if the user is already logged in; that's irrelevant... and logging such an occurrence would be wasteful and confusing. It's perfectly reasonable, and probably common, for a user to have multiple logins; e.g., from the dashboard and from `pbench-generate-token`, or think of a functional test environment where a "test" user runs multiple tests in parallel.

If we want an admin to be able to track logins, we need administrative tools, something we're currently missing. I can see someone wanting to see how many sessions are active, how many users are logged in, and stuff like that. We should have a robust audit logging system that allows an admin to see who logged in (and out) when, as well as to trace the life cycle of all our resources (including users, datasets). None of that is part of this PR, nor should be!

npalaska: Fixed

dbutenhof (dismissed):

lib/pbench/server/api/resources/users_api.py
npalaska: I removed the logging statement since it was not really needed and a little misleading

portante (dismissed):

lib/pbench/server/api/resources/users_api.py
```diff
@@ -348,7 +327,6 @@ def get(self, username):
             try:
                 target_user = User.query(username=username)
                 response_object = {
-                    "message": "success",
                     "data": target_user.get_json(),
```
dbutenhof, line 330: Since we're only returning the User JSON, hiding it behind a `"data"` element seems obtuse; can we just return the object here?

dbutenhof (changes requested):

lib/pbench/server/api/resources/users_api.py
npalaska: sure

dbutenhof (dismissed):
Looks like maybe you didn't fix the indentation when you removed the `"data"` element? But I'm not sure I want another cycle to fix it???

npalaska:
yeah I realized that after the push, I fixed the formatting in the comments along with rebase

lib/pbench/server/api/resources/users_api.py
```diff
@@ -321,15 +307,17 @@ def get(self, username):
             Authorization:  Bearer Pbench_auth_token (user received upon login)
 
         :return: JSON Payload
-            response_object = {
-                    "message": "Success"/"failure message",
-                    "data": {
+            Success: 200,
+                    response_object = {
```
dbutenhof, line 311: I'm confused. I still don't have "hide whitespace changes" enabled in GitHub, but it looks like you _increased_ the indentation on line 311 to match the now-overly-indented lines 312 to 314, whereas I'd have expected the opposite.

It's a docstring comment; I'm not overly concerned unless GitHub is lying to me...

dbutenhof (approved):

lib/pbench/server/api/resources/users_api.py
```diff
@@ -276,10 +263,12 @@ def post(self):
         Required headers include
             Authorization:   Bearer <Pbench authentication token (user received upon login)>
```
webbnh, line 264: Does the logout code behave well if it is invoked with an expired but otherwise valid token?

My instinct is that we shouldn't be too picky about the supplied token:  all that we really want to do here is to delete it from our database, so if it was in the database and we managed to delete it, I'd say that we were successful, regardless of the current validity of the token.  If we first validate the token (other than by trying to delete it), then I think we create more problems than we solve (e.g., if it is expired, we're presumably going to return an error to the client and leave the expired token in the database to be cleaned up later...each of which is an undesirable outcome under the circumstances).

lib/pbench/server/api/resources/users_api.py
```diff
@@ -339,18 +327,12 @@ def get(self, username):
 
         # TODO: Check if the user has the right privileges
         if verified:
-            response_object = {
-                "status": "success",
-                "data": user.get_json(),
-            }
+            response_object = user.get_json()
             return make_response(jsonify(response_object), 200)
         elif user.is_admin():
```
webbnh, line 332: @npalaska, I realize that it's out of scope for this PR, but could you please correct the text at line 344 -- I think it should refer to "get", not "delete"....

lib/pbench/test/unit/server/test_user_auth.py
```diff
@@ -290,13 +279,37 @@ def test_update_user(client, server_config):
...
+            response = login_user(client, server_config, "username", "12345")
+            assert response.status_code == 200
+            data_login = response.json
+
+            response = client.put(
+                f"{server_config.rest_uri}/user/username",
+                json={"auth_tokens": "external_auth_token"},
+                headers=dict(Authorization="Bearer " + data_login["auth_token"]),
+            )
+            assert response.status_code == 400
```
webbnh, line 312: Would it be worthwhile to test the error message text?  (E.g., to confirm that it failed for the reason that it was supposed to fail for?)

webbnh (dismissed):
Looks good.  I have a couple of sort of policy questions, but they could be handled in a subsequent PR.  I did have a question relevant to this PR on a unit test, and I tripped over a...let's call it a "typo" which would be nice to have fixed if we can tolerate slipping it in, here.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -217,17 +216,6 @@ def post(self):
                 500, message="INTERNAL ERROR",
             )
 
-        # check if the user is already logged in, in case the request has a an authorization header included
-        # We would still proceed to login and return a new auth token to the user
-        if user == self.auth.token_auth.current_user():
```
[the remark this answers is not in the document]
npalaska: yeah since we have removed this check, optional token requirement on line 140 isn't achieving anything

lib/pbench/server/api/resources/users_api.py
npalaska: If client uses an expired token on logout we will return an error, however, if we detect the token is expired we immediately delete it from our database (see line 104 auth.py). Logout endpoint is protected with `login_required` decorator therefore if client uses the expired token we won't even proceed with a request but at the same time it will also get deleted from the database (Same for any endpoint that is strictly protected). 

However, I have already recorded this as an issue to track https://github.com/distributed-system-analysis/pbench/issues/2130.

lib/pbench/test/unit/server/test_user_auth.py
npalaska: We could do that

dbutenhof (approved):

webbnh (approved):

lib/pbench/server/api/resources/users_api.py
webbnh: Oh, no...check out line 397....  (How did we miss these in #1937??)

:-)

webbnh (approved):
GitHub didn't note anything changed in this latest round from the previous round, and neither did I.  So, I've no reason not to approve.

However, I am left wondering whether maybe something you intended to commit didn't make it in (like a change to `users_api.py` line 397, for instance?).

dbutenhof (approved):

portante (changes requested):

dbutenhof (approved):

webbnh (approved):
I don't see any changes...was this just a rebase?

webbnh (dismissed):
I don't see any changes...was this just a rebase?

npalaska:
yeah just a rebase

portante (changes requested):
One more rebase needed ...

portante (approved):

webbnh (approved):

PR #2151, opened 2021-03-25
Update the agent Makefile for pbench-generate-token and other Click-based tools

webbnh:
This PR is for a pair of small changes to the Agent makefile which prepares the sources for the RPM build.

The Click-based tools are actually wrapper scripts generated from information in the `setup` configuration during the installation.  They are then individually copied to the destination directory.  Rather than promulgating this practice, this PR collects the Click-based tools into a list and then performs a single, bulk copy of the list.

And, this change adds `pbench-generate-token` to the list.

Note, this change also adds indentation to the rules' command lines which are continued from the previous line.  (I'm not a fan of these -- I would prefer to see them combined into a single command -- but I didn't want to make that level of change without a better understanding of why they were set up that way.)

### DIFF
--- a/agent/Makefile
+++ b/agent/Makefile
@@ -53,6 +53,18 @@ util-scripts = \
 	require-rpm \
 	tool-meister
 
+# Scripts based on the Python Click package, which are generated during
+# installation.
+click-scripts = \
+	pbench-generate-token \
+	pbench-cleanup \
+	pbench-clear-results \
+	pbench-clear-tools \
+	pbench-config \
+	pbench-list-tools \
+	pbench-list-triggers \
+	pbench-register-tool-trigger
+
 bench-scripts = \
 	pbench-cyclictest \
 	pbench-dbench \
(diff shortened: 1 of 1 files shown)


portante (approved):

agent/Makefile
```diff
@@ -53,6 +53,18 @@ util-scripts = \
 	require-rpm \
 	tool-meister
 
+# Scripts based on the Python Click package, which are generated during
+# installation.
+click-scripts = \
+	pbench-generate-token \
```
dbutenhof, line 59: Ha; they're all in alphabetical order except yours. ;-)

agent/Makefile
```diff
@@ -241,48 +253,42 @@ install-python3-setup: install-util-scripts install-lib
...
 	mkdir -p ${DESTDIR}/python3
 	(cd ..; SKIP_GENERATE_AUTHORS=1 SKIP_WRITE_GIT_CHANGELOG=1 python3 setup.py install --prefix=${DESTDIR}/python3)
-	${COPY} ${DESTDIR}/python3/bin/pbench-config ${UTILDIR}/
-	${COPY} ${DESTDIR}/python3/bin/pbench-clear-tools ${UTILDIR}/
-	${COPY} ${DESTDIR}/python3/bin/pbench-list-triggers ${UTILDIR}/
-	${COPY} ${DESTDIR}/python3/bin/pbench-clear-results ${UTILDIR}/
-	${COPY} ${DESTDIR}/python3/bin/pbench-cleanup ${UTILDIR}/
-	${COPY} ${DESTDIR}/python3/bin/pbench-list-tools ${UTILDIR}/
-	${COPY} ${DESTDIR}/python3/bin/pbench-register-tool-trigger ${UTILDIR}/
+	${COPY} $(addprefix ${DESTDIR}/python3/bin/, ${click-scripts}) ${UTILDIR}/
```
dbutenhof, line 256: Well that's a neat trick.

dbutenhof (approved):

agent/Makefile
webbnh: Dang it!

agent/Makefile
webbnh: Yeah, `make` has all sorts of cool widgets.

agent/Makefile
portante: We can fix this is another commit.

PR #2153, opened 2021-03-26
Improve diagnostic for connection error in pbench-generate-token

### LINKED ISSUE #2150: Need the host name on a failure to connect to pbench server
> [root@8a2ad2ec52b6 rpm]# /opt/pbench-agent/util-scripts/pbench-generate-token 
Username: wscales
Password: 
<urllib3.connection.HTTPConnection object at 0x7fbec7ac19d0>: Failed to establish a new connection: [Errno -2] Name or service not known

webbnh:
This pull request consists of a small commit which tries to produce a better diagnostic when the `pbench-generate-token` tool is unable to connect to the pbench server.

The `requests` package `ConnectionError` exception text is not very friendly, so we catch that exception and raise our own exception which provides a nicer message (not quite as nice as I would like, but better than what was there).

Old message:
```
<urllib3.connection.HTTPConnection object at 0x10a661100>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known
```

New message:
```
Cannot connect to 'http://pbench.example.com/api/v1/login'
```

Resolves #2150.

### DIFF
--- a/lib/pbench/cli/agent/commands/generate_token.py
+++ b/lib/pbench/cli/agent/commands/generate_token.py
@@ -24,12 +24,11 @@ class GenerateToken(BaseCommand):
             "password": self.context.password,
             "token_duration": self.context.token_duration,
         }
+        uri = f"{server}/login"
         try:
-            response = requests.post(f"{server}/login", headers=headers, json=payload)
+            response = requests.post(uri, headers=headers, json=payload)
         except requests.exceptions.ConnectionError as exc:
-            if hasattr(exc.args[0], "reason"):
-                raise exc.args[0].reason
-            raise
+            raise RuntimeError(f"Cannot connect to '{uri}'") from exc
 
         payload = response.json()
         if response.ok:
--- a/lib/pbench/test/unit/agent/task/test_generate_token.py
+++ b/lib/pbench/test/unit/agent/task/test_generate_token.py
@@ -164,4 +164,6 @@ class TestGenerateToken:
         )
         assert result.exit_code == 1
         assert not result.stdout
-        assert str(result.stderr).find("Failed to establish a new connection") != -1
+        assert -1 != str(result.stderr).find(
+            "Cannot connect to 'http://pbench.example.com/api/v1/login'"
+        )


lib/pbench/cli/agent/commands/generate_token.py
```diff
@@ -27,9 +27,7 @@ def execute(self):
         try:
             response = requests.post(f"{server}/login", headers=headers, json=payload)
         except requests.exceptions.ConnectionError as exc:
-            if hasattr(exc.args[0], "reason"):
-                raise exc.args[0].reason
-            raise
+            raise RuntimeError(f"Cannot connect to {server}") from exc
```
dbutenhof, line 30: Definitely an improvement, although I'd prefer to either show the full URI (including the `/login`) or else drop the path entirely and show just the host part (`http://pbench.example.com`). Nobody _actually_ connects to `http://pbench.example.com/api/v1` and you won't get any useful result if you try.

portante (changes requested):
Looks good, would also be great to have a unit test that verifies the error handling.

webbnh:
There _is_ a unit test which verifies the error handling...and...it's failing....  :-}

I shall enhance the code and update the test.

lib/pbench/cli/agent/commands/generate_token.py
ndokos: But I should be able to test by using the full URI (with the `/login` part), right? If so, I think the complete URI should be reported. Or if that's not the case,  the "thing" that I could cut and paste into a browser to check the outcome.

ndokos (commented):
Comment inline.

lib/pbench/cli/agent/commands/generate_token.py
webbnh: It now prints the full URI in the error message (see the PR Description).

dbutenhof (approved):
Looks good, but you've angered the lint spirits:

```
black --check ...
[1mwould reformat /home/pbench/lib/pbench/test/unit/agent/task/test_generate_token.py[0m
[1mOh no! 💥 💔 💥[0m
[1m1 file would be reformatted[0m, 121 files would be left unchanged.[0m
FAILED: black --check .
```

webbnh:
Old news.

portante (approved):

PR #2181, opened 2021-03-29
Refactor tool-meister-start orchestration

portante:
We needed a way to specify what IP address and port to use when the Redis server and Tool Data Sink (really its WSGI server) bind for their respective services, while also being able to specify what address and port the Tool Meister(s) should connect to in environments where those address and ports are different (software-defined-networking, or NAT'd, or some other reason).

We solve this need by adding a new option, `--orchestrate`, along with a new `--tool-data-sink` option, and adjust the behaviors a bit.

The new `--orchestrate` option accepts two keywords, `create` and `existing`, which allow the caller to direct tool-meister-start to either be responsible for creating the Redis server, Tool Data Sink, and Tool Meister instances, or use existing instances.  The `--redis-server` (or the `PBENCH_REDIS_SERVER` environment variable) and `--tool-data-sink` (or the `PBENCH_TOOL_DATA_SINK` environment variable) options give the caller the ability to specify the IP/port combinations to use in either operating case.

When `--orchestrate=existing` is specified, the `--redis-server` and `--tool-data-sink` options (or environment variables) must be specified.

The format for specifying the IP/ports to use via the `--redis-server` and `--tool-data-sink` options can be one of the following:

  * `<host ip>:<port>`: used both for listening and connections
  * `<listen host ip>:<port>;<connection host ip>:<port>`: the 1st host/port pair is used for listening for connections, the 2nd host/port pair is used to specify how to make connections to the instances

The defaults are:
  * Redis: bind & connect, `${_pbench_full_hostname}:17001`
    * Also binds to `localhost`
  * TDS:   bind & connect, `${_pbench_full_hostname}:8080`

Prior to this change, `pbench-tool-meister-start` had one very long `main` method.  This refactoring breaks out the orchestration pieces required to start the Redis server, Tool Data Sink, and the Tool Meisters, into their own methods and classes, reducing `main` to something a bit more reasonable.

As part of this work, a new `BaseServer` abstraction, sub-classed for the `RedisServer` and `ToolDataSink`, now handles how the Redis server and Tool Data Sink instance tracking works, either when they are created by tool-meister-start, or created ahead of time by the caller.


portante:
@RobertKrawitz, okay here is the fixed code to handle specifying the listen (binding) vs connection IP/ports for both the Tool Data Sink and the Redis Server as created by pbench-tool-meister-start.

portante:
I need to test this in a real container, still, but it is a start.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +202,621 @@ def waitpid(pid):
...
+
+    envi_var = None
+    def_port = None
+    bad_port_code = None
+
+    class Err(ReturnCodeErr):
+        pass
+
+    def __init__(self):
+        # The pid of the managed server
```
ndokos, line 360: What about it?

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +202,621 @@ def waitpid(pid):
...
+        self.host = None
+        self.port = None
+        self.bind_host = None
+        self.bind_port = None
+        self._managed = None
```
ndokos, line 365: What is the type of `self.managed`?

ndokos: From the two examples, the answer is `boolean`.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +202,621 @@ def waitpid(pid):
...
+                host_ip is not None
+            ), f"Logic Bomb!  socket.gethostbyname('{self.host}') returned None"
+            # By default, to talk to the Redis server locally, use the
+            # specified host name.
+            self.local_host = self.host
+
+        bind_hostnames_l = [self.bind_host]
+        # Determine if we can also use "localhost" to talk to the Redis server.
+        if self.host != self.bind_host:
+            # Somebody went through the trouble of telling use to bind to one
```
ndokos, line 661: "us"

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +202,621 @@ def waitpid(pid):
...
+            # NOTE: we don't bother trying to determine multiple bind hosts.
+
+            # Take advantage of the bind IP to have local connections use the
+            # local IP address; hardcoded value avoids setups where "localhost"
+            # is not setup (go figure).
+            self.local_host = "127.0.0.1"
+        else:
+            # See if we can safely add "localhost" to the bind host name.  This
+            # check is necessary because sometimes callers might have already
+            # specified a name that maps to 127.0.0.1, and Redis will through
```
ndokos, line 674: "throw"

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +202,621 @@ def waitpid(pid):
...
+            ) from exc
+        else:
+            if retcode != 0:
+                raise self.Err(
+                    f"failed to create redis server, daemonized; return code: {retcode:d}",
+                    ReturnCode.REDISFAILED,
+                )
+
+    def kill(self, ret_val):
+        """kill_redis_server - given a redis server PID file, attempt to KILL the
```
ndokos, line 733: "kill"? or "kill_redis_server"?

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +202,621 @@ def waitpid(pid):
...
+        Redis server.
+
+        Returns "1" if successfully KILL'd; "2" if it encounters an error reading
```
ndokos, line 736: Symbolic names for retcodes?

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -749,19 +1117,17 @@ def main(_prog, cli_params):
...
+    # ret_val = wait_for_tds(to_client_chan, logger)
+    ret_val = tool_data_sink.wait(to_client_chan, logger)
     if ret_val != 0:
-        if cli_params.redis_server is None:
-            # We created the Redis server, so we should clean it up.
-            return kill_redis_server(redis_pid, ReturnCode.TDSWAITFAILURE)
-        else:
-            return ReturnCode.TDSWAITFAILURE
+        # Clean up the Redis server if we created it.
+        return redis_server.kill(ReturnCode.TDSWAITFAILURE)
```
ndokos, line 1124: Shouldn't this be under an `is_managed()` check?

ndokos (changes requested):
Light first pass: typos and some questions. Need to go through `main()` in more detail though.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -58,12 +72,26 @@ steps are taken as a normal client:
    1. Collect any requested system information ("sysinfo" action)
    2. Start any persistent tools running ("init" action)
 
-We then leave running, locally, a Redis server, a Tool Data Sink process,
-and any local or remote Tool Meisters.
+If requested, we then leave running, locally, a Redis server, a Tool Data Sink
+process, and any local and/or remote Tool Meisters.
```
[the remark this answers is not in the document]
Maxusmusti: Especially because it seems like we create and leave them running locally when the user **doesn't** add any request (`--redis-server` or env var)

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -1,33 +1,47 @@
...
+sub-system.
 
-Responsible for orchestrating the Redis Server, Tool Data Sink (TDS), and Tool
-Meisters (TMs):
+If the user does not specify the `--redis-server` command line parameter (or
+the `PBENCH_REDIS_SERVER` environment variable is not present, or is an empty
+string), tool-meister-start is responsible for orchestrating the creation of a
+Redis Server, Tool Data Sink (TDS), and Tool Meisters (TMs).  If the user does
+specify `--redis-server` (or environment variable), tool-meister-start assumes
+the user has already instantiated a TDS and appropriate TMs, and begins
```
Maxusmusti, line 12: Isn't there a `--redis-server` and a `--tool-data-sink` option though? This seems to imply that using `--redis-server` means that redis, tds, and tms have all already been setup by the user. If I just use the --redis-server option, is it then assuming the tds is running locally (`--redis-server` doesn't require `--tool-data-sink` according to usage)? I feel like if I was using only that option, I would expect it to only assume I set up a redis server myself. Also why does --tool-data-sink then require `--redis-server`, would it not go off that same assumption that it used when only `--redis-server` was used (for the tds)? 
The point is, I think with the newly added level of abstraction in this file, it could probably support more than "all or nothing" for pre-set-up components. Is there a reason why we can't have `--redis-server` only mean that the user already has a redis server at `<hostname>:<port>`, have `--tool-data-sink` only mean that the user has a tds already at `<hostname>:<port>`, and maybe a `--tm premade` option to indicate that the user already has tms running on the registered hostnames? They would then be able to mix and match according to their needs (maybe they have an existing redis server they want to use and need us to make tds/tms, or maybe they are using their own tds/tm containers but want us to provide a redis-server).

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +202,621 @@ def waitpid(pid):
...
+                        retcode,
+                    )
+        except Exception as exc:
+            raise self.Err(
+                "failed to create tool data sink, daemonized", ReturnCode.TDSFORKFAILED
+            ) from exc
+
+        # Wait for logging channel to be up and ready before we start the
+        # local and remote Tool Meisters.
+        timeout = time.time() + _TDS_STARTUP_TIMEOUT
```
Maxusmusti, line 521: Maybe after 15-20 seconds of waiting add a message saying "Waiting for tool-data-sink..." or something similar, as I would probably assume something was broken or had crashed well before waiting a minute for start-up, since it usually takes a max of 2-3s. That also gives the user some time to start up their tds and save the run if they forgot to run it before using the `tool-meister-start` script.

Maxusmusti (changes requested):
Looks awesome so far, just a couple of design questions

agent/util-scripts/gold/pbench-tool-meister-start/test-54.txt
```diff
@@ -13,6 +15,9 @@ optional arguments:
                         Use an existing Redis server specified by
                         <hostname>:<port>; implies an existing Tool Data Sink
                         and Tool Meisters as well.
+  --tool-data-sink TOOL_DATA_SINK
+                        Use an existing Tool Data Sink specified by
+                        <hostname>:<port>; requires --redis-server as well.
```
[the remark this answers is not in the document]
portante: No, as it stands today, this is just about how to connect to an existing instance, not how to configure a new instance ourselves.

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
portante: We are either acting as an orchestrator of the Redis server, the Tool Data Sink, and all the Tool Meisters, or we are not, and somebody else created all of them ahead of time.

agent/util-scripts/pbench-tool-meister-start
portante: Removed.

agent/util-scripts/pbench-tool-meister-start
portante: Yes.  We start off a `None` indicating it is not set.  Notice that the assertions that use it assert `True`, not a boolean result value.

agent/util-scripts/pbench-tool-meister-start
portante: Fixed.

agent/util-scripts/pbench-tool-meister-start
portante: While this isn't a change from before, I'll see how to make that happen.

agent/util-scripts/pbench-tool-meister-start
portante: Added flexibility means both added logic and added testing to handle all the cases.

At this time, it seems best to support only one or the other.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +204,635 @@ def waitpid(pid):
...
+                )
+            try:
+                self.port = int(parts[1])
+            except Exception as exc:
+                raise self.Err(
+                    f"Bad Tool Data Sink WSGI port specified, {spec!r}",
+                    self.bad_port_ret_code,
+                ) from exc
+            else:
+                self.host = parts[0]
```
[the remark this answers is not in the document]
webbnh: Ditto for the Redis server.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +204,635 @@ def waitpid(pid):
...
+            # This instance will be managed by us.
+            self._managed = True
+        self.local_host = None
+        self.pid_file = None
+
+    def start(self, tm_dir, full_hostname, logger):
+        """start_redis - configure and start a Redis server.
+
+        Returns a tuple of the Redis server host name, the host name to be used to
+        connect to it locally, and its port.
```
webbnh, line 631: Not in any obvious way (i.e., I can't find a `return` statement in this function...).

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +204,635 @@ def waitpid(pid):
...
+            raise self.Err(
+                f"{self.host} does not map to an IP address", ReturnCode.NOIP
+            ) from exc
+        else:
+            assert (
+                host_ip is not None
+            ), f"Logic Bomb!  socket.gethostbyname('{self.host}') returned None"
+            # By default, to talk to the Redis server locally, use the
+            # specified host name.
+            self.local_host = self.host
```
webbnh, line 663: I don't think this assignment is ever used -- I think it gets overwritten first.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +204,635 @@ def waitpid(pid):
...
+        assert (
+            self._managed is True
+            and self.host is not None
+            and self.port is not None
+            and self.bind_host is not None
+            and self.bind_port is not None
+        ), f"Logic bomb!  Unexpected state: {self!r}"
+
+        try:
+            raw_pid = self.pid_file.read_text()
```
webbnh, line 758: About this assertion....  If we're not managed, then we should not have gotten past line 747; this function doesn't appear to reference any of the other state tested here; but, it _does_ touch `self.pid_file`, so...perhaps it should test that, instead?

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +204,635 @@ def waitpid(pid):
...
+
+    def kill(self, ret_val):
+        """kill - attempt to KILL the running Redis server.
+
+        This method is a no-op if the server instance isn't managed by us.
+
+        Returns ReturnCode "enum" via the "kill" return code method.
+        """
+        if self._managed is False:
+            return ret_val
```
webbnh, line 747: It seems like this should be calling `self.managed()` instead of referencing the attribute directly -- doing so would give you the chance to address the `self._managed is None` case without disturbing the code, here.

webbnh (commented):
More to come....

agent/util-scripts/pbench-tool-meister-start
portante: When I originally started to read this comment, I leaped to a different conclusion almost immediately from the very first question:  Why have command line options (and associated environment variables) and then have a second set of environment variables to change how the orchestration is performed?

As a result, I am thinking of changing the interface to be that there is a new switch, `--orchestrate` which accepts the keyword options: `create` and `existing`.  When `--orchestrate=create` is used, it directs `pbench-tool-meister-start` to create its own instance of a Redis server, a Tool Data Sink, and create both the local and remote Tool Meister instances.  When `--orchestrate=existing` is used, it directs `pbench-tool-meister-start` to expect the caller to have performed the orchestration, and then to expect to be told how to contact the Redis server and Tool Data Sink via the `--redis-server` and `--tool-data-sink` options.  When `--orchestrate` is not present, it would default to `create` to allow for the current behavior of `agent/bench-scripts` to work without further changes.

For example, to support:
 * The current behavior of our `agent/bench-scripts`, no changes
 * Using IP/port combinations different from the defaults, the user would define `PBENCH_REDIS_SERVER` and `PBENCH_TOOL_DATA_SINK` to have the required values (changing the `--redis-server` and `--tool-data-sink` default values given)
 * A Kubernetes environment, one would specify `--redis-server=<...>` & `--tool-data-sink=<...>` (or as environment variables) without `--orchestrate`, and `pbench-tool-meister-start` would be contact them as directed

This would allow for one way to specify the IP/port combinations (command line options and environment variables), and then direct the behavior requested explicitly.  Further, the keyword options make it easier to extend the behavior in the future if desired.

Thoughts?

webbnh: I'm amused:  last week, I wrote up a lengthy reply to Mustafa's comment, but then it got attached to the rest of my review (thanks, Github...), and, by the time I was ready to post it, you (@portante) had already replied, so I just deleted it.  I was going to suggest (among other things) an option not unlike what you outline here.

Since we're intending only two modes of support, it does seem like we should have a binary set of options.  The interface should make it clear to each of the user and the tool which mode we're supposed to be in, and then the parsing and validation should be able to be handled in an orderly fashion.

A switch like `--orchestrate` seems good, but I would avoid providing keyword values to it, unless you seriously think we'll be adding additional values.  I suggest we stick with a binary choice, like `--orchestrate` and `--no-orchestrate` (particularly if our command parser of choice supports negatable options).  Obviously, omitting the option would yield the appropriate default (that being, presumably, having Pbench create the processes).

That said, you seem to be looking to make the `--orchestrate` switch basically orthogonal to the other switches.  (That seems good....)  So, if `--no-orchestrate` is specified, then the agent will _not_ create the processes; otherwise (e.g., if the switch is omitted), it _will_ create them all.  And, then, `--redis-server` and `--tool-data-sync` (etc.) are used to specify where the processes can be contacted, _regardless_ of who created them.

Sounds good.

The only testable constraint (that comes to my mind) is, if `--no-orchestrate` is specified, then both `--redis-server` and `--tool-data-sync` _must_ be specified; otherwise, we have applicable defaults for all cases (I think).

portante: > The only testable constraint (that comes to my mind) is, if `--no-orchestrate` is specified, then both `--redis-server` and `--tool-data-sync` _must_ be specified; otherwise, we have applicable defaults for all cases (I think).

Agreed.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +204,635 @@ def waitpid(pid):
...
+    Raises a StartTmsErr on failure.
+
+    NOTE: all local and remote Tool Meisters are started even if failures
+    occur for some; this allows the user to see logs for all the individual
+    failures.
+    """
+    failures = 0
+    successes = 0
+    # NOTE: it is assumed that the location of the pbench-tool-meister command
+    # is the same on the local host as it is on any remote host.
```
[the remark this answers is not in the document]
portante: Is is really a risk?  If we have a consistent installation defined in the RPM `.spec` file for the agent, and we have a container image installation that is well defined, then most of the "risky" cases are covered.  If somebody goes beyond on those two and creates an entirely different installation methodology, is that something we really want to support without a change to the code, where adding this to the `PATH` places this CLI interface along side other CLI interfaces we support?

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +204,635 @@ def waitpid(pid):
...
+    tool_meister_cmd = exec_dir / "tool-meister" / "pbench-tool-meister"
+    ssh_cmd = "ssh"
+    ssh_path = find_executable(ssh_cmd)
+    base_args = [
+        ssh_cmd,
+    ]
```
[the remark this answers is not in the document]
portante: Not sure why the [original code](https://github.com/distributed-system-analysis/pbench/pull/2181/files?file-filters%5B%5D=No+extension#diff-28fc2747e424d2014f9e3b3dd3e834315fa6db5a6079f62185988250ed197f69L621) was formatted that way, but I can certainly address this here.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +204,635 @@ def waitpid(pid):
...
+    tms = dict()
+    tm_count = 0
+    for host in tool_group.hostnames.keys():
+        tm_count += 1
+        tm_param_key = f"tm-{tool_group.group}-{host}"
+        if host == full_hostname:
+            logger.debug("5a. starting localhost tool meister")
+            try:
+                pid = os.fork()
```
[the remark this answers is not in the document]
portante: First, I don't really want to change this process in this commit, as this code is really just a "lift-n-shift" to move the behavior to a new method.

As for the history, this was a very early decision made in the Tool Meister development where it seemed to gain us no real advantage since we still have to wait for the combined set of processes, and the `subprocess` interfaces don't ease the logic for that.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +204,635 @@ def waitpid(pid):
...
+    if failures > 0:
+        raise StartTmsErr(
+            "failures encountered creating tool miesters", ReturnCode.TMFAILURES
+        )
+    if successes == 0:
+        raise StartTmsErr(
+            "unable to successfully start any tool meisters,"
+            " but encountered no failures either",
+            ReturnCode.TMNOSUCCESSES,
+        )
```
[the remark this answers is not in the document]
portante: Yes, the condition at line 335 would catch this.  We should just `assert len(tool_group.hostnames) > 0`, and assert `tm_count == len(tool_group.hostnames) and tm_count == (successes + failures)`.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +204,635 @@ def waitpid(pid):
...
+                        ) from exc
+            host = host_port_parts[0] if host_port_parts[0] else full_hostname
+            pairs.append((host, port))
+
+        self.bind_host, self.bind_port = pairs[0]
+        if len(pairs) == 2:
+            # Separate bind/connecting ip:port
+            self.host, self.port = pairs[1]
+        else:
+            assert len(pairs) == 1, "Logic bomb!  unexpected pairs, {pairs!r}"
```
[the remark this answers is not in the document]
portante: Well it is using `split(";", 1)`, which means `parts` at line 417 will only ever have at most 2 "parts" to process.  So `len(pairs)` can only be `2`, `1`, `0`, which is why we are asserting it is `1`.  If we get `0` for pairs, that means `envi_val` is an empty string, which can only happen if `full_hostname` is an empty string.  But the check for `full_hostname` at line 961 should take care of that.  So this assertion seems to be sufficient.

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
portante: Addressed.

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
portante: Addressed.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +204,635 @@ def waitpid(pid):
...
+    def __init__(self):
+        # Host / port pair for how to connect to the given instance of a
+        # "server".
+        self.host = None
+        self.port = None
+        # (Optional) Host / port pair for how a server will "bind" to offer its
+        # services.
+        self.bind_host = None
+        self.bind_port = None
+        # Whether or not this is instance is managed by the execution of
```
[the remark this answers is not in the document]
portante: Fixed.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +204,635 @@ def waitpid(pid):
...
+        self._managed = None
+
+    def __repr__(self):
+        return f"{self.host}:{self.port} / {self.bind_host}:{self.bind_port} - {self._managed}"
+
+    def __str__(self):
+        return f"{self.host}:{self.port}"
+
+    def managed(self):
+        return self._managed
```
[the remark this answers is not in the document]
portante: Addressed.

agent/util-scripts/pbench-tool-meister-start
dbutenhof: I kinda like the `--orchestrate` binary. I'm not really crazy about `--no-orchestrate` but the twist here is that the default probably needs to be the historic "orchestrate" behavior of start/stop, so we're likely stuck with it. In any case, I like that better than the `--orchestrate=<keyword>` alternative.

Then again, this still implies "all or nothing" and even though you're not supporting anything else right now, we may well end up moving towards a "split orchestration" model where someone will expect to point us at an existing Redis server but let us orchestrate the data sync and tool meisters because that's purely internal Pbench-ese.

This makes me think we might be better off without `--orchestrate` at all, and simply rely on `--redis-server` to override the internal Redis orchestration and `--tool-data-sync` to override that orchestration. Right now we could require either no overrides or all overrides because you don't internally support splits, but that could easily be relaxed later. Why do we need a second related option? The default becomes "full orchestration because you haven't specified existing endpoints", which is historic behavior.

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
dbutenhof: True; but then maybe the `NOTE` is just stirring up useless confusion and could be removed? Or even just say "Pbench agent must be installed using RPM or equivalent on all nodes".

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +204,635 @@ def waitpid(pid):
...
+        assert self.envi_var is not None, "Logic bomb!  No envi var set"
+        assert self.def_port is not None, "Logic bomb!  No default port set"
+
+        envi_val = os.environ.get(self.envi_var, full_hostname)
+        parts = envi_val.split(";", 1)
+        pairs = []
+        for part in parts:
+            host_port_parts = part.rsplit(":", 1)
+            if len(host_port_parts) == 1:
+                port = self.def_port
```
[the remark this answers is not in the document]
dbutenhof: I like that idea.

>         (host, port) = matches.group(1, 2)

And for what it's worth, I'm a fan of using named capture groups: it makes both the Regex and the code using it a lot more easily readable.

agent/util-scripts/pbench-tool-meister-start
portante: What we currently have does that, but it uses 2 different sets of environment variables to make that happen, and requires that both `--tool-data-sink` and `--redis-server` be used.

The new proposal allows us to operate with no additional options in the default case, and use only one pair of environment variables by adding the `--orchestrate=<keyword>` switch (which avoids adding a second `--no-orchestrate`.

I'll add a commit that implements the new method so that we can decide.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -290,7 +878,37 @@ def main(_prog, cli_params):
...
+    ):
+        logger.error(
+            "both --redis-server and --tool-data-sink must be specified if either is used"
+        )
+        return ReturnCode.MISSINGPARAMS
+
+    try:
+        redis_server = RedisServer(cli_params.redis_server)
+    except RedisServer.Err as exc:
+        logger.error("{exc}")
```
webbnh, line 895: Does this line do what you want it to do?

I would have thought that you would want one of these, instead:
```
        logger.error(f"{exc}")
```
or
```
        logger.error(str(exc))
```
or maybe
```
        logger.error(exc)
```
or something similar hitting an method or attribute on `exc`.

Ditto line 901.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -322,29 +940,48 @@ def main(_prog, cli_params):
...
     # Load and verify required and optional environment variables.
     try:
-        benchmark_run_dir = Path(os.environ["benchmark_run_dir"]).resolve(strict=True)
+        benchmark_run_dir_val = os.environ["benchmark_run_dir"]
         hostname = os.environ["_pbench_hostname"]
         full_hostname = os.environ["_pbench_full_hostname"]
-    except Exception:
-        logger.exception("failed to fetch parameters from the environment")
+    except KeyError as exc:
+        logger.error("failed to fetch required environment variable, '%s'", exc.args[0])
```
webbnh, line 947: Nice!  :-)

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -357,109 +994,19 @@ def main(_prog, cli_params):
...
-        except ValueError:
-            logger.error("Bad Redis port specified, '%s'", cli_params.redis_server)
-            return ReturnCode.BADREDISPORT
-        else:
-            redis_host = parts[0]
+            redis_server.start(tm_dir, full_hostname, logger)
+        except redis_server.Err as exc:
+            logger.error(f"Failed to start a local Redis server: '{exc}'")
+            # return ReturnCode.EXCREDISCREATION
+            return exc.return_code
```
webbnh, line 1009: Why is there commented-out code, here?  I recommend either implementing it or removing it (or adding an appropriately explanatory comment with a `FIXME` or `TODO`).

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -470,31 +1017,34 @@ def main(_prog, cli_params):
     # listen for responses from the Tool Data Sink.
     try:
         to_client_channel = f"{cli_tm_channel_prefix}-{tm_channel_suffix_to_client}"
-        redis_server = redis.Redis(host=redis_host, port=redis_port, db=0)
-        to_client_chan = RedisChannelSubscriber(redis_server, to_client_channel)
+        redis_client = redis.Redis(host=redis_server.host, port=redis_server.port, db=0)
+        to_client_chan = RedisChannelSubscriber(redis_client, to_client_channel)
```
webbnh, line 1021: Having a variable named `to_client_chan` is too similar to the previously named `to_client_channel`.  I recommend renaming `to_client_channel` to `to_client_channel_name` (or, perhaps, just `channel_name`, or even `name`, since it's in a highly-restricted scope).

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -531,216 +1082,74 @@ def main(_prog, cli_params):
...
 
     # +
-    # 4. Start the Tool Data Sink process
+    # Step 4. - Start the Tool Data Sink process (optional)
     # -
 
-    if cli_params.redis_server is None:
-        # FIXME: if only one host is registered, and that host is the same as this
-        # controller, then don't bother starting the Tool Data Sink.
+    if redis_server.managed():
```
webbnh, line 1096: Should this be checking `tool_data_sink.managed()` instead?  If not, should line 1035 be checking `redis_server.managed()` instead?

(I'm nervous about having two independent checks which are supposed to match each other....)

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -749,19 +1158,17 @@ def main(_prog, cli_params):
         "6. waiting for all successfully created Tool Meister processes"
         " to show up as subscribers"
     )
-    ret_val = wait_for_tds(to_client_chan, logger)
+    # ret_val = wait_for_tds(to_client_chan, logger)
+    ret_val = tool_data_sink.wait(to_client_chan, logger)
```
webbnh, line 1162: Can this commented-out code be removed?

webbnh (changes requested):
Done!  (For this round....)

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
webbnh: If the RPM really controls where the command gets installed (and there isn't an option to `dnf`, et al. to say, "No, install it over here."), then I suppose there is no particular risk...and no need for the `NOTE`.  :-)

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
webbnh: Re "lift-n-shift", got it!

As for the rest of it, I think I disagree, but I'm not inclined to debate it in this PR.

agent/util-scripts/pbench-tool-meister-start
webbnh: If we have appropriate command line parser support, then `--orchestrate` and `--no-orchestrate` are opposite senses of _the same option_ (e.g., if both forms appear on the same command line, it will either be automagically caught as an error or the second occurrence will override the first -- either way, from an implementation perspective, it's a __single__ option).

dbutenhof: Apparently `ArgumentParser` doesn't handle negation of keyword options, and doing so is therefore something of a hack. Another option would be simply `--orchestrate=[yes|no]` ... or we could make it a choice keyword of `["none", "all", "redis", "tms", "tds"]` ;-)

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
webbnh: > `len(pairs)` can only be `2`, `1`, `0`

I concur.  (Sorry that I got a little fanciful in my comment.)

> If we get `0` for pairs, that means `envi_val` is an empty string, which can only happen if `full_hostname` is an empty string.

What happens if `os.environ[self.envi_var]` is an empty string (i.e., the user defined the environment variable to be an empty string)?  I think that means that `envi_val` will be an empty string, which means that `len(pairs)` will be zero, which means we'll blow an assertion due to a user error.  I'm opposed to that....  :-)

agent/util-scripts/pbench-tool-meister-start
webbnh: Yeah, if we don't have support for negatable command line options, then the road gets bumpier.  The problem is that we want the _absence_ of the option to be the same as the __positive__ assertion, which means that we need the _presence_ of the option to be the __negative__ assertion, which means we need to name the option `--no-orchestrate` or some positive synonym for it (and lets forget about implementing an `--orchestrate` option).

I don't think that `--orchestrate=[yes|no]` is significantly better than `--orchestrate=[create|existing]`.  (And, @dbutenhof , your more flexible suggestion left out `"tms-and-tds"`, et al.  ;-) )

I'll go look at Peter's new  commit when its ready, but perhaps something like `--external_orchestration` would work (if it is present, we don't create the processes; if it's absent, we do).

dbutenhof: > I don't think that `--orchestrate=[yes|no]` is significantly better than `--orchestrate=[create|existing]`. (And, @dbutenhof , your more flexible suggestion left out `"tms-and-tds"`, et al. ;-) )

Actually, my unstated assumption was that `ArgumentParser` could support a _list_ of those keyword values; although after glancing at the documentation I'm not certain that it can. But, if it could, the default would be `all` (traditional behavior) while a fully externally-orchestrated environment would be `none`. And if we aren't immediately going to support mix-and-match, then the other values are irrelevant for now.

Whether this is an improvement or just a side comment... well, yeah, that's not so clear.

portante: It support that with `choices`, see the latest commit, which I believe makes this a bit more compact.

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
portante: At this time I really don't want to commit to testing all the combinations of who starts what.  I'd rather support just two: `pbench-tool-meister-start` handles them all or it doesn't.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -58,12 +72,26 @@ steps are taken as a normal client:
...
+  - _pbench_hostname
+  - _pbench_full_hostname
+
+There are 4 environment variables optionally used to provide metadata about
+the benchmark execution environment:
+
+  - benchmark
+  - config
+  - date
+  - ssh-opts
```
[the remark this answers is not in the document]
portante: Tracked in issue #2203.

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
portante: This seems much more complicated as a regex than its worth.  We could consider the use of the `ipaddress` module.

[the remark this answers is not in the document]
webbnh: The regex does the "recognition" (i.e., superficial validation) as well as the parsing in one swell foop, so, in the sum of things, I don't think the regex is really more complicated.  Yeah, it's scary looking, but we can declare the regex at the top of the class where no one will have to look at it.  ;-)

I would certainly support using `ipaddress` as well, but I didn't get the impression that it would handle a FQDN -- I thought it only did numerics.  If it's a full replacement for the regex, then I'm on-board.

portante:
Okay, since some of the code-review comments were asking about validating host names and IP addresses, and we had issue #2201 come up during this window of work, I decided to add a commit to solve #2201 first.  It is now the first commit before any other changes.

I'll follow that up with another refactoring so that the `pbench-tool-meister-start` host name processing can leverage it as well.

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
portante: Added a separate initial commit for general hostname validation, then added an additional commit to use it in the new `pbench-tool-meister-start` code.

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
portante: Since we are no longer using an environment variable now, I believe this issue is resolved.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -58,12 +72,26 @@ steps are taken as a normal client:
    1. Collect any requested system information ("sysinfo" action)
    2. Start any persistent tools running ("init" action)
 
-We then leave running, locally, a Redis server, a Tool Data Sink process,
-and any local or remote Tool Meisters.
+If requested, we then leave running, locally, a Redis server, a Tool Data Sink
+process, and any local and/or remote Tool Meisters.
 
 The pbench-tool-meister-stop command will take care of (gracefully) stopping
 all of these processes, locally or remotely.
```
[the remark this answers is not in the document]
portante: Removed.

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
portante: Reworded this entire section.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +204,635 @@ def waitpid(pid):
...
+    """BaseServer - abstract base class for common code shared between the
+    ToolDataSink and RedisServer classes.
+    """
+
+    envi_var = None
+    def_port = None
+    bad_port_code = None
+
+    class Err(ReturnCodeErr):
+        pass
```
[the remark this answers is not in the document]
portante: One can also used the `help()` to see the entire derived sub-classing behavior.  I am inclined not to add a redefinition of a constructor for readability.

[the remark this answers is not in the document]
webbnh: How does `help()` work in GitHub?  ;-)

Seriously, though -- `help()` is very cool (I didn't know about that!), but it only works if I have a running interactive Python instance with the appropriate modules imported into it.  What I was hoping for is declaration of the c'tor here with appropriate type hints, which will help the reader and which will potentially help future developers (to the extent that type hints are actually functional).

But, we can defer this to another time.

dbutenhof (commented):
The levels of `validate-hostname` / `validate_hostname` are a bit convoluted, but I like the idea.

I see Webb pointed out a few problems that need to be resolved; but pending that, it's looking good.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -507,12 +1001,15 @@ def main(_prog, cli_params):
         # host names instead.
```
portante, line 1001: Will remove this `FIXME`, tracking in #2205.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +204,635 @@ def waitpid(pid):
...
+    has already taken the steps to create a Redis server, and fired up a Tool
+    Data Sink, and all the Tool Meisters they require.  All we have to do is
+    talk to the Tool Data Sink and Tool Meisters via the given Redis server.
+
+    When a `redis_server` is not provided, we are responsible for creating and
+    orchestrating the setup of a Redis server, a Tool Data Sink, and all the
+    required Tool Meisters.  There are 4 environment variables which influence
+    this setup:
+
+        * PBENCH_REDIS_BIND_HOSTNAME - "<hostname>[:port]", host name and optional
```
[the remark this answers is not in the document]
portante: Removed entirely.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +204,635 @@ def waitpid(pid):
...
+                                       port number to use for the Redis server
+                                       to bind to; can be different from the
+                                       PBENCH_REDIS_HOSTNAME to allow for external
+                                       connections through address translation
+                                       schemes (NAT or SDN); if not defined,
+                                       then PBENCH_REDIS_HOSTNAME is used, and
+                                       if that is not defined, then the value
+                                       of the `_pbench_full_hostname` variable
+                                       is used
+        * PBENCH_REDIS_HOSTNAME      - "<hostname>[:port]", host name the Tool
```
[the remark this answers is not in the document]
portante: Removed entirely.

lib/pbench/agent/utils.py
```diff
@@ -188,3 +190,40 @@ def collect_local_info(pbench_bin):
         hostdata[arg] = cp.stdout.strip() if cp.stdout is not None else ""
 
     return (version, seqno, sha1, hostdata)
+
+
+# From https://stackoverflow.com/questions/106179/regular-expression-to-match-dns-hostname-or-ip-address
+_allowed = re.compile(
+    r"^(([a-zA-Z]|[a-zA-Z][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z]|[A-Za-z][A-Za-z0-9\-]*[A-Za-z0-9])$"
+)
```
webbnh, line 198: I have several suggestions, here:

- Add `flags=re.IGNORECASE` to the `re.compile()` call so that you don't need to include both letter ranges.
- Use `re.fullmatch()` and omit the anchors.
- Use non-capturing groups.
- Based on the way that you are actually using this RE, you don't need to look for dots in it.
- RFC 1123 states:  "the restriction on the first character is relaxed to allow either a letter or a digit" which has the side effect of allowing your algorithm to match either dotted-decimal quartets or FQDNs.

So, I think you can simplify this to just:
```
_allowed = re.compile(r"[A-Z0-9](?:[A-Z0-9\-]{0,61}[A-Z0-9])?", flags=re.IGNORECASE)
```
That is, a case-blind comparison seeking a letter-or-digit, optionally followed by a sequence of zero to 61 letter-or-digit-or-hyphens followed by a letter-or-digit.

lib/pbench/agent/utils.py
```diff
@@ -188,3 +190,40 @@ def collect_local_info(pbench_bin):
...
+    """
+    if len(host_name) > 255:
+        return 1
+
+    if host_name[-1] == ".":
+        # Strip trailing '.' to allow comprehension below to work in all cases
+        _host_name = host_name[:-1]
+    else:
+        _host_name = host_name
+    if all(_allowed.match(x) for x in _host_name.split(".")):
```
webbnh, line 219: As I mentioned above, I recommend `fullmatch()` rather than `match()`.

lib/pbench/agent/utils.py
```diff
@@ -188,3 +190,40 @@ def collect_local_info(pbench_bin):
...
+        return 0
+
+    # It is not a valid host name, but could be a valid IP address.
+    try:
+        ipaddress.ip_address(host_name)
+    except ValueError:
+        ret_val = 1
+    else:
+        ret_val = 0
+    return ret_val
```
webbnh, line 229: If you accept my suggested RE above, this code should never be presented with any valid IPv4 addresses.  However, it does validate IPv6 ones, right?  It would be worth updating the code comment to that effect.

lib/pbench/test/unit/agent/test_validate_hostname.py
```diff
@@ -0,0 +1,13 @@
...
+from pbench.agent.utils import validate_hostname
+
+
+def test_validate_hostname():
+    assert validate_hostname("test") == 0
+    assert validate_hostname("test.example.com") == 0
+    assert validate_hostname("tes_t.example.com") == 1
+    assert validate_hostname("--run-label__") == 1
+    assert validate_hostname("127.0.0.1") == 0
+    assert validate_hostname("1270.0.0.1") == 1
```
webbnh, line 13: No IPv6 love?

webbnh (changes requested):
This just reviews the first commit with the support for validating hostnames.

I think that this is great stuff (although I do have the usual pile of nits and suggestions), but I think that this change should be split into a separate PR.

webbnh:
> The levels of validate-hostname / validate_hostname are a bit convoluted, but I like the idea.

I had similar thoughts, but I decided to just go with it.  :-)

lib/pbench/agent/utils.py
```diff
@@ -188,3 +190,40 @@ def collect_local_info(pbench_bin):
...
+    names are not resolved to IP addresses, and IP addresses are also accepted.
+
+    Algorithm taken from: https://stackoverflow.com/questions/2532053/validate-a-hostname-string
+
+    Returns 0 on success, 1 on failure.
+    """
+    if len(host_name) > 255:
+        return 1
+
+    if host_name[-1] == ".":
```
webbnh, line 214: This code should guard against zero-length strings in the "is the last character a dot" test.  I don't know whether you want to catch the `IndexError` or add another clause to the existing length test.

portante:
> ... I think that this change should be split into a separate PR.

Filed draft PR #2207.

agent/util-scripts/pbench-tool-meister-start
portante: Removed.

agent/util-scripts/pbench-tool-meister-start
portante: No, that is a bug, nice catch.

agent/util-scripts/pbench-tool-meister-start
portante: Reworked.

agent/util-scripts/pbench-tool-meister-start
portante: Reworked.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +204,635 @@ def waitpid(pid):
...
+            self._managed = False
+        else:
+            self.port = None
+            self.host = None
+            # This instance will be managed by us.
+            self._managed = True
+
+    def __str__(self):
+        return f"{self.host}:{self.port}"
+
```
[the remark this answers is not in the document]
portante: Reworked.

agent/base
```diff
@@ -100,14 +100,32 @@ fi
...
+        error_log "Invalid ${2}: '${1}'"
+        exit 1
+    fi
+}
+
 if [[ -z "${_pbench_hostname}" ]]; then
     export _pbench_hostname=$(hostname -s)
+else
+    validate_hostname "${_pbench_hostname}" "_pbench_hostname"
 fi
```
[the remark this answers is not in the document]
portante: I'll address in PR #2207.

agent/util-scripts/pbench-move-results
```diff
@@ -80,6 +80,20 @@ while true; do
...
+	error_log "Missing controller name (should be \"hostname -f\" value)"
+	exit 1
+    fi
+else
+    validate-hostname "${controller}"
+    if [[ ${?} -ne 0 ]]; then
+	error_log "Invalid controller specified: '${controller}'"
+	exit 1
+    fi
+fi
```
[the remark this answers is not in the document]
portante: I'll address the validation in PR #2207.

As for the log message, that is the same log as message as before, and then intention is not to report what _is_ the host name, but the intended semantic value of what `_pbench_full_hostname` should be.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -1215,22 +1110,44 @@ if __name__ == "__main__":
...
+        dest="orchestrate",
+        default="create",
+        choices=("create", "existing"),
+        help=(
+            "The `create` keyword directs the command to create the various"
+            " instances of the Redis server, Tool Data Sink, and Tool"
+            " Meisters, while the `existing` keyword directs the command to"
+            " use existing instances of all three."
+        ),
+    )
```
webbnh, line 1147: I still think that this would be better as a binary (True/False) option rather than a keyword "choice".  If we call it something like `--external-orchestration` and give it a default value of false, then, unless the switch is specified (that is, for all cases where we're not doing external orchestration), pbench will manage the processes.

If we're really in love with having a keyword interface for the UI (i.e., because we honestly think that we might expand beyond two choices), then that's fine, but, in this proposed implementation, we shouldn't be propagating the keyword-ness beyond the command line processing:  after that point, we should have the state captured in a boolean.  Having tests like `if orchestrate == "create"` is a bad idea, because it's too easy to misspell the keyword (or to be inconsistent about which of the choices we reference); and, having different instances of the test in various parts of the code is just asking for problems.

agent/util-scripts/validate-ipaddress
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python3
+
+import ipaddress
+import sys
```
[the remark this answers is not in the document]
portante: I'll address in PR #2207.

agent/util-scripts/validate-ipaddress
```diff
@@ -0,0 +1,18 @@
...
+
+try:
+    ipaddress.ip_address(sys.argv[1])
+except ValueError:
+    exit_val = 1
+except IndexError:
+    print("Missing IP address argument", file=sys.stderr)
+    exit_val = 1
```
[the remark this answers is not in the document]
portante: I'll address in PR #2207.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -1215,22 +1110,44 @@ if __name__ == "__main__":
...
     parser.add_argument(
         "--redis-server",
         dest="redis_server",
         default=os.environ.get("PBENCH_REDIS_SERVER", None),
         help=(
-            "Use an existing Redis server specified by <hostname>:<port>;"
-            " implies an existing Tool Data Sink and Tool Meisters as well."
+            "Specifies the IP/port to use for the Redis server - if not"
+            " present, the defaults are used; can be present in two forms:"
```
webbnh, line 1154: Should this be explaining what the defaults _are_?  (This is code, so it _could_ actually print out the values.)

Ditto line 1145.

webbnh: Rather than "can be present", how about something like, "the specified value can have either of".

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -1215,22 +1110,44 @@ if __name__ == "__main__":
...
+            " `<bind host>:<port>;<host>:<port>`, a semi-colon separated"
+            " IP/port specified for both how the Redis server will bind"
+            " itself, and how clients will connect; `<host>:<port>`, the"
+            " IP/port combination is used both for binding and connecting"
+            " (NOTE: binding is ignored with --orchestrate=existing)"
```
webbnh, line 1159: Perhaps replace "ignored" with "not used"?  (I'll grant it's a subtle distinction, but "ignored" implies that we use something else instead....)

Ditto line 1150.

lib/pbench/agent/utils.py
portante: I'll address in PR #2207.

lib/pbench/agent/utils.py
portante: I'll address in PR #2207.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -994,13 +821,71 @@ def main(_prog, cli_params):
...
+                "both --redis-server and --tool-data-sink must be specified"
+                " if --orchestrate=existing is used"
+            )
+            return ReturnCode.MISSINGPARAMS
+        ssh_cmd, ssh_path = None, None
+
+    try:
+        redis_server = RedisServer(cli_params.redis_server, full_hostname)
+    except RedisServer.Err as exc:
+        logger.error("{exc}")
```
webbnh, line 875: Does this code still need fixing?  (Ditto line 852.)

webbnh (commented):
Mostly nits'n'stuff.  But, I think it would be better if we replaced the keyword-based orchestration option with a simple binary one.

lib/pbench/agent/utils.py
portante: I'll address in PR #2207.

lib/pbench/agent/utils.py
portante: I'll address in PR #2207.

portante:
Rebased on top of PR #2207.

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
portante: Moved to PR #2207.

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
portante: At line 405 we check for an empty `spec` and use the "default host name" (`def_host_name`) provided, which is already validated.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -773,6 +790,21 @@ def main(_prog, cli_params):
...
         return ReturnCode.MISSINGREQENVS
+    if not full_hostname or not hostname:
+        logger.error(
+            "_pbench_hostname ('%s') and _pbench_full_hostname ('%s')"
+            " environment variables are required to represent the respective"
+            " hostname strings",
+            hostname,
+            full_hostname,
+        )
+        return ReturnCode.MISSINGHOSTNAMEENVS
```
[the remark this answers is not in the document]
portante: Not sure it is worth combining all this. We want the user to know that we expect both `_pbench_hostname` and `_pbench_full_hostname` defined.  They could certainly infer that from encountering both error messages by executing things twice, but this log message explains the requirement.  The validation individually explains the requirement that valid host names be used.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -448,50 +414,27 @@ class BaseServer:
             assert len(pairs) == 1, "Logic bomb!  unexpected pairs, {pairs!r}"
             self.host, self.port = pairs[0]
 
+    def __repr__(self):
+        return f"{self.host}:{self.port} / {self.bind_host}:{self.bind_port}"
+
+    def __str__(self):
+        return f"{self.host}:{self.port}"
```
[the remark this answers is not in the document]
portante: I'll look at the use of these, which I think is non existent.

agent/util-scripts/pbench-tool-meister-start
portante: Philosophically, I'd like to avoid using boolean for command line options and use enums instead.

I can make an enum of the choices and use that for the tests.

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
webbnh: So, `logger.error()` is going to send this to `stderr` or somewhere else where the end-user will see it?  (I keep falling under the misconception that "the log" gets squirreled away where we can ask for it, but it's otherwise mostly out of sight; I was trying to save code because we already know most of what this message is trying to say -- we just want the `%s` values, which I think we would get from the code later on.)

From what you say here, it seems like it would be worthwhile to enhance this:  report on each individually -- but report them both if they are both bad -- and go ahead and reference the `hostname -s` and `hostname -f` commands specifically, like you did elsewhere.

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
webbnh: I couldn't find them, but I didn't look very hard....

agent/util-scripts/pbench-tool-meister-start
webbnh: Yeah...at some point, I would love to have you explain why booleans are to be avoided!  To me they are easy to understand and hard to screw up...at least when the choice is truly binary.

Anyway, if you really want to avoid a "switch", then just encapsulate all the `orchestrate == "create"` tests in a function which returns a boolean value, and I'll shut up.

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
portante: Only used in assertion reporting.  Dropped the `__str__`.

agent/util-scripts/pbench-tool-meister-start
portante: > Yeah...at some point, I would love to have you explain why booleans are to be avoided!

See https://www.teamten.com/lawrence/programming/prefer-enums-over-booleans.html, https://betterprogramming.pub/dont-use-boolean-arguments-use-enums-c7cd7ab1876a, https://nedbatchelder.com/blog/200507/booleans_suck.html regarding.

> just encapsulate all the orchestrate == "create" tests in a function which returns a boolean value

Good idea.

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
portante: See line 763 where we use the `StreamHandler` for the logger.

dbutenhof (approved):

agent/util-scripts/pbench-tool-meister-start
webbnh: Thanks for the references -- those were interesting reading.

Unfortunately, the last one torpedoes his own argument at the end where he raises my exact concerns over _not_ using a boolean.  The second one was also only marginally convincing:  clearly, if you are representing a series of "states", an enum is a better choice; however, that approach suffers when the states are not properly sequenced or distinct, and it is still subject to the evolution problem outlined at the end of the last reference, where you have to reevaluate all of your existing logic when you add the third (or even other) states.  I found the first reference to be the most useful, but, in actuality, his principal argument is just a strawman (I sure that it was a real mistake, but the mistake was not in using a boolean but rather in the understanding of what the boolean meant).

So, it seems like that concerns with booleans are:

1. Lack of extensibility
2. Lack of clarity as to what the two values mean
3. Lack of readability of function invocations
4. Lack of type safety

Lack of extensibility is not necessarily a "problem" -- in fact, it can be a "feature".  It is a problem when the interface is poorly designed or conceptualized; if the concept being represented truly is a binary choice, then using a boolean _supports_ the design and _avoids_ problems in the code.  (This is my principal reason for recommending them, here.)

Lack of clarity also reflects a problem in the design or conceptualization, not in the datatype choice.  Enums can come to the rescue here by renaming "True" and "False" to something interface-specific, but this issue can also be addressed in other ways (such as by naming the interface elements properly or by providing explicit testing functions rather than by testing the attribute directly).  (So, I don't see this alone as a reason to give up the rigor of boolean operation.)

Lack of readability depends on the language and coding style.  In a language like Python which supports keyword arguments, by defining good parameter names and by using them in function invocations, the intention of the caller is clear.  Obviously, using enums would do the name thing, but at the cost of additional mechanism.  (So, for us, working in Python, this is not a reason to give up booleans.)

Lack of type safety would be a good argument, but, if we assume that we're using keyword arguments, there's no possibility of mixing up the parameters.  (So, enums buys us nothing here.)

Therefore, to me, it comes down to trade-off between extensibility and rigorous implementation.  If what we're describing or implementing really has only two states, either has or doesn't have a particular characteristic, or is otherwise represented by a binary choice, then a boolean is unquestionably the right approach -- it just cannot be screwed up, providing we use good engineering (without which enums won't save us...).  On the other hand, if what we're describing or implementing could reasonably be expected to evolve a third state or is otherwise a little squishy, then an enum would be a better choice.

Either way, "rolling our own enums" (e.g., using strings) is a bad idea:  it has type un-safety built right in, and it suffers from all of the growth problems.  We can make that approach more robust, by encapsulating it behind a functional interface for setting and testing the values, but, really, if that's where we're going, we should just switch to an enum.  (Perhaps in the next PR....)

dbutenhof: I think the big concern here is that the evolution from strict boolean (yes we orchestrate, no we don't) is not an enum, but rather a set of enums (or else a proliferation of confusing enums). That is, we want to orchestrate Redis, or the data sync, or the tool meisters; or data sync and tool meisters, or Redis and tool sync but not tool meisters (a plausible case if tool meisters are fired up on cluster pods but the external controller wants us to take care of the data sync they'll use).... etc.

I'm not sure we've really thought out all of the use cases and tradeoffs... which plays back into why Peter didn't want to take on more than the boolean segregation now.

webbnh: Understood, and I concur...and, I think you just explained why `--orchestrate` should _not_ take a keyword/choice:  in a conceivable future, the question of whether or not Pbench should create the Redis server will be orthogonal to the question of whether or not it should create the data sink or the tool meisters, so trying to represent all those (overlapping) states with a single enum/option will likely be problematic, while representing them with a series of booleans will be clear.  (Yes, using booleans _would_ result in the possibility of invalid combinations, but these are easily detected during command line parsing; this is balanced against the challenge of enumerating all of the possible valid combinations and documenting the "name" of each of them.)

So, I don't think it's wrong for `--orchestrate` (or, rather, something like `--external-orchestration`) to be a boolean:  at the moment, it would mean "all or nothing"; in the future, we could split off other components to their own switches, or add semantics to the existing switches which describe the other components.

agent/base
```diff
@@ -100,15 +100,30 @@ fi
 # don't use colons and dashes in the date suffix
 export date_suffix=$(date --date ${date} --utc +"%Y.%m.%dT%H.%M.%S")
 
+function validate_hostname {
+    validate-hostname "${1}"
+    if [[ ${?} -ne 0 ]]; then
```
webbnh, line 105: Not a suggestion (necessarily), just a question:  why use this idiom over the following?
```
    if ! validate-hostname "${1}"; then
```

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -432,15 +431,14 @@ class BaseServer:
...
             assert len(pairs) == 1, "Logic bomb!  unexpected pairs, {pairs!r}"
             self.host, self.port = pairs[0]
+            self._repr = f"{self.name} - {self.host}:{self.port}"
 
     def __repr__(self):
-        return f"{self.host}:{self.port} / {self.bind_host}:{self.bind_port}"
-
-    def __str__(self):
-        return f"{self.host}:{self.port}"
+        return self._repr
```
webbnh, line 441: :-D

I was thinking that the code would decide what to print dynamically, if/when it was called upon to actually print something, rather than use an attribute....  c'est la vie.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -850,35 +854,36 @@ def main(_prog, cli_params):
         ssh_opts=ssh_opts,
     )
 
-    orchestrate = cli_params.orchestrate
-    if orchestrate == "create":
+    _orchestrate = cli_params.orchestrate
```
webbnh, line 857: Not a big deal, but why bother instantiating `_orchestrate` when what we care about is the value of `orchestrate`?

I would have been tempted to create a helper function which accepts `cli_params` and returns the value for `orchestrate` -- it could do all the asserting and testing that it wants, and this code would just work with the result.  (And, it would encapsulate the details of the command line parameter far away....)

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -1151,12 +1156,14 @@ if __name__ == "__main__":
         default=os.environ.get("PBENCH_REDIS_SERVER", None),
         help=(
             "Specifies the IP/port to use for the Redis server - if not"
-            " present, the defaults are used; can be present in two forms:"
+            " present, the defaults are used,"
+            f" ${{_pbench_full_hostname}}:{def_redis_port};"
+            " the specified value can take either of two forms:"
```
webbnh, line 1161: Nice addition, but the syntax is cumbersome.  I would have suggested isolating the f-string to just the bit which needs formatting:
```
            "Specifies the IP/port to use for the Redis server - if not"
            " present, the defaults are used, ${_pbench_full_hostname}:"
            f"{def_redis_port};"
            " the specified value can take either of two forms:"
```

webbnh (approved):
Ship it!

(Unless you feel compelled to incorporate my suggestions....)

agent/base
portante: Creature of habit, my lack of deep understanding of `bash`, and I have always disliked executing commands in `if` statements.  That may be why I would like to convert all this to Python3 ... ;)

agent/util-scripts/pbench-tool-meister-start
portante: Filed issue #2211 to track.

agent/util-scripts/pbench-tool-meister-start
portante: It seems silly to dynamically create the string when it is really fixed a creation time.

portante (commented):
Okay, I have rebased this commit and squashed all the review comments down into the main commit, updated the commit message to match the PR description and code behaviors.

Should be good to go for final sign off after the required rebase.

portante:
Okay, not quite done, need PR #2212 to land first so we can properly build RPMs and containers with these changes.

portante:
@Maxusmusti, @ndokos, this PR still have 2 changes requested from the two of you.  Is that still the case?

agent/base
webbnh: I find that referencing `${?}` tends to be fragile (e.g., the temptation to insert debugging statements between the command which sets it and the code which tests it is overpowering at times, and, of course, doing so is catastrophic), while the idiom of immediately capturing the value in a local variable to prevent that problem tends to be "wordy" and cumbersome.  So, in many cases, invoking the command _in_ the `if` statement is just the right thing.

Don't think of it as "executing a command" -- think of it as making a subroutine call.  ;-)

agent/util-scripts/pbench-tool-meister-start
webbnh: It's mostly a space vs. speed trade-off -- why create and store the value if it's never actually needed?  If it's only needed on an error or diagnostic path, why pay the extra cost in the normal case?

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -1,33 +1,52 @@
...
 
    1. Loading tool group data for the requested tool group
-   2. Starting a Redis server and creating the Redis channel for the TDS to
-      talk to the client
+      - This is the first step regardless so that the tool information can be
+        validated, the number of TMs and their marching orders can be
+        enumerated, and when orchestrating the TMs, the list of where those
+        TMs are requested to run can be determined
+   2. [optional] Starting a Redis server and creating the Redis channel for
+      the TDS to talk to the client
```
webbnh, line 28: Only the "starting" is optional, right?  Creating the channel is done regardless of who started the server, right?  (Why does this make me nervous....)

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -809,13 +1135,48 @@ if __name__ == "__main__":
         default=None,
         help="The list of system information items to be collected.",
     )
+    parser.add_argument(
+        "--orchestrate",
+        dest="orchestrate",
+        default="create",
+        choices=("create", "existing"),
+        help=(
+            "The `create` keyword directs the command to create the various"
```
ndokos, line 1144: The message needs to mention what the default is: apparently `--help` does not say anything about the default value (see e.g. the top of test-54.txt).

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -809,13 +1135,48 @@ if __name__ == "__main__":
...
+        "--tool-data-sink",
+        dest="tool_data_sink",
+        default=os.environ.get("PBENCH_TOOL_DATA_SINK", None),
+        help=(
+            "Specifies the IP/port to use for the Tool Data Sink - if not"
+            " present, the defaults are used, ${_pbench_full_hostname}:"
+            f"{def_wsgi_port};"
+            " the specified value can take either of two forms:"
+            " `<bind host>:<port>;<host>:<port>`, a semi-colon separated"
+            " IP/port specified for both how the Redis server will bind"
```
ndokos, line 1176: Should this be `Tool Data Sink` rather than `Redis server`?

ndokos (changes requested):

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +226,530 @@ def waitpid(pid):
...
+                except ValueError as exc:
+                    if host_port_parts[1] == "":
+                        port = self.def_port
+                    else:
+                        raise self.Err(
+                            f"Bad port specified for {self.name} in '{spec}'",
+                            self.bad_port_ret_code,
+                        ) from exc
+            host = host_port_parts[0] if host_port_parts[0] else def_host_name
+            if validate_hostname(host) != 0:
```
webbnh, line 423: If `host` contains an IPv6 address, it will (presumably) be surrounded by square brackets which need to be removed prior to validating it.

webbnh (approved):
I assume that you are going to open an issue for the IPv6 address-with-port format, rather than add it to this PR, so, modulo the documentation nit below (and Nick's concerns), I think this is good to go.

portante (commented):
Okay, ready for another pass.  I had to rebase anyways, and these changes were straightforward to make.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -420,6 +420,11 @@ class BaseServer:
                             self.bad_port_ret_code,
                         ) from exc
             host = host_port_parts[0] if host_port_parts[0] else def_host_name
+            if host[0] == "[" and host[-1] == "]":
+                # Brackets are invalid for a host name, but might be used when
+                # specifying a port with an IPv6 address, strip them to before
```
dbutenhof, line 425: Typo: "...strip them before..."

Although I would prefer to handle bracket consistency checks as part of the validation; `[pbench.example.com]` and `[10.0.0.1]` are not valid addresses, but `[fe00::0]` is ...

Still, as a hack to avoid collapsing with valid IPv6 addresses, this is a decent first step. ;-)

dbutenhof (dismissed):
Yeah, we need to clean up the IPv6 addressing, but since we're not using IPv6 I guess that can be deferred. (Do we even have IPv6 in RDU2 or BOS labs?)

Maxusmusti (dismissed):
Passed usual local/remote node testing with expected results

agent/util-scripts/pbench-tool-meister-start
portante: Supporting brackets is weird.  The `ipaddress` module does not accept the brackets, and I am pretty sure that specifying a host name with brackets is not accepted for other interfaces.  So we have to strip the brackets no matter what.

And yes, IPv6 is supported in our lab infrastructure, and it turns out that we use it locally for communicating with our GraphQL instance.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +226,535 @@ def waitpid(pid):
...
+        "<tm param key>",
+        "yes",  # Yes, request the tool meister daemonize itself
+    ]
+    tms = dict()
+    tm_count = 0
+    for host in tool_group.hostnames.keys():
+        tm_count += 1
+        tm_param_key = f"tm-{tool_group.group}-{host}"
+        if host == full_hostname:
+            logger.debug("6a. starting localhost tool meister")
```
webbnh, line 277: Why "6a" instead of just "6" like line 1035?

webbnh (approved):
Looks great, let's ship it!  (Please don't make me review the entire thing again....)

What?  Oh, them?...no, pay no attention to those comments.....

dbutenhof (approved):
I say go.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,13 +226,535 @@ def waitpid(pid):
...
+
+
+def start_tms_via_ssh(
+    exec_dir: Path,
+    ssh_cmd: str,
+    ssh_path: Path,
+    tool_group: str,
+    ssh_opts: str,
+    full_hostname: str,
+    redis_server,
```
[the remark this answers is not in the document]
portante: Yes, and I did not go back to refactor this to so that the classes can be defined before the hint is used.

agent/util-scripts/pbench-tool-meister-start
portante: It was `5a` before ...

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
portante: Tracked in #2217.

ndokos (approved):

PR #2186, opened 2021-03-30
Add dependencies to pbench-devel required to build RPMs

webbnh:
This PR modifies the Dockerfile for the `pbench-devel` container to add a couple of missing dependencies required to build RPMs for the agent and server.

PR #2187, opened 2021-03-30
Unify agent and server RPM builds

webbnh:
Previous to this PR, the agent and server each had separate makefiles which included separate subordinate `rpm.mk` files, but overall, the contents were nearly identical.  This PR creates a _common_ `rpm.mk` file, and strips the agent and server makefiles down to little more than an `include` directive.

There were a few subtle changes in behavior as a result, and so I'm keeping this PR as a "draft" until the reviewers are comfortable with the results (and until the change to `jenkins/development.Dockerfile`, which is being reviewed in #2186, is merged.)

I tested by comparing the output of the RPM builds without these changes and to the output with these changes; the differences were all reasonable and expected.  And, I was able to install the resulting RPMs in the `pbench-devel` container.

Some of the items to pay particular attention to:

- The `rpm-clean` target now removes the directories which the `rpm-dirs` target creates, rather than just their contents.
- The `rpm-dirs` target is now a dependency of all the targets which require one or more of the directories which it creates.  This means that this target no longer needs to be built explicitly, and it means that the other targets no longer need to explicitly create those directories.
- The `patches` target now references the current directory as '`.`' rather than via a calculated absolute reference.  This is simpler and hopefully safe.
- Previously, the server build modified `pbench-config` and `pbench-server` to change the script interpreter to `/usr/bin/env python3`.  This seems unnecessary (e.g., if we need it for those two, why don't we need it for _all_ of the tool scripts?), so I've commented it out in the new, common code.  Please let me know whether I should enable or remove it (and the `FIXME`...).
- The code now uses `:=` instead of `=` for all variables which are set to expressions which reference values outside the makefile, such as external files or shell call-outs, so that the expressions are evaluated only once.
- The build now updates the Git submodules unconditionally and recursively.  The recursion isn't required for the agent, and the step isn't required at all for the server, but it shouldn't hurt to do so, and it will help "future-proof" the build.  Also, the command is issued from the current directory instead of from the source root, but my understanding is that it should work properly regardless of where it is issued from, so long as the current directory is in a Git checkout.
- I don't think the `clean-sha1` target worked before this change; should I remove it or fix it?  (If the latter, can you suggest how?)
- Do we want to retain the `veryclean` target?  (I don't think it does anything more than the `clean` target does.)


utils/rpm.mk
```diff
@@ -0,0 +1,103 @@
...
+.PHONY: rpm-dirs
+rpm-dirs:
+	mkdir -p $(addprefix ${HOME}/rpmbuild/,${RPMDIRS})
+
+.PHONY: submodules
+submodules:
+	git submodule update --init --recursive
+
+# FIXME:  Do we still need the sed command to correct for Python3?
+#	  (Previous, only the server was doing it, and only for two files.)
```
dbutenhof, line 67: Good question. I remember seeing this and commenting before; I don't remember the answer. The idea of editing files in the RPM builder always bugged me...

utils/rpm.mk
```diff
@@ -0,0 +1,103 @@
...
+_copr_user = ${USER}
+endif
+
+COPR_TARGETS = copr copr-test copr-interim copr-index copr-inotify copr-dashboard
+.PHONY: ${COPR_TARGETS}
+${COPR_TARGETS}: $(RPMSRPM)/$(prog)-$(version)-$(seqno)$(sha1).src.rpm
+	copr-cli build $(_copr_user)/$(subst copr,pbench,$@) $(RPMSRPM)/$(prog)-$(VERSION)-$(seqno)g$(sha1).src.rpm
+
+.PHONY: veryclean
+veryclean:: clean rpm-clean
```
dbutenhof, line 92: I'd never even noticed the "veryclean" target, and it does indeed seem rather pointless given the structure.

I like your "cleaner clean" and "veryclean" becomes redundant. But what happened to "dot-clean"? I didn't attempt to trace where those files came from, and whether they're still created...

dbutenhof (commented):
Minor comments and questions...

utils/rpm.mk
```diff
@@ -0,0 +1,103 @@
...
+
+$(RPMSRPM)/$(prog)-$(version)-$(seqno)$(sha1).src.rpm: srpm
+
+ifdef COPR_USER
+_copr_user = ${COPR_USER}
+else
+_copr_user = ${USER}
+endif
+
+COPR_TARGETS = copr copr-test copr-interim copr-index copr-inotify copr-dashboard
```
ndokos, line 86: Only the first two are valid - the rest are obsolete.

webbnh, line 86: Reduce this list to just the first two.

utils/rpm.mk
portante: Turns out that we can use RPM to make that happen with the correct Python RPM building support.

webbnh: Apparently the RPM-building tool itself does it, and this is supposed to avoid conflicts with that....

In any case, Nick is going to try to figure out whether we can dispense with this, now.

utils/rpm.mk
```diff
@@ -0,0 +1,103 @@
+# Common definitions for make an RPM, used by both the Agent and the Server.
```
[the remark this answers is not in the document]
webbnh: I think it was a typo; I've fixed it.

utils/rpm.mk
webbnh: According to the [GNU `make` documentation](https://www.gnu.org/software/make/manual/html_node/Standard-Targets.html#Standard-Targets), "standard" targets are `clean`, `distclean`, `mostlyclean`, and `maintainer-clean`, but [it also mentions](https://www.gnu.org/software/make/manual/html_node/Goals.html#index-clean-_0028standard-target_0029) as "typical" `realclean` and `clobber`.  I'm familiar with `clean`, `distclean`, and `clobber` -- I've never encountered the others.

In any case, after talking with Nick, it sounds like my re-implementation of `rpm-clean` is a little too thorough, so I think I'll replace `distclean` with `clobber` and tone down `rpm-clean` and leave `clean` pointing at it.

utils/rpm.mk
webbnh: Done.

webbnh: Fixed.

utils/rpm.mk
webbnh: COPR offered no objections, so I'm going to remove this.

utils/rpm.mk
```diff
@@ -0,0 +1,92 @@
...
+
+.PHONY: spec
+spec: rpm-dirs ${prog}.spec.j2
+	if [ -e ./seqno ] ;then expr ${seqno} + 1 > ./seqno ;fi
+	jinja2 ${prog}.spec.j2 -D version=${VERSION} -D gdist=g${sha1} -D seqno=${seqno} > ${RPMSPEC}/${prog}.spec
+	rpmlint ${RPMSPEC}/${prog}.spec
+
+.PHONY: patches
+patches: rpm-dirs
+	if [ -d ./patches ] ;then cp ./patches/* ${RPMSRC} ;fi
```
portante, line 49: I realize you probably didn't write this, but for safety, can you add a trailing slash to ensure `cp` considers `${RPMSRC}` to be a directory?

portante (approved):
Approving, but if you feel like making a small change, that would be great.

ndokos (dismissed):
LGTM - modulo Peter's comment on the `cp`.

ndokos (approved):

portante (approved):

ndokos (approved):

PR #2188, opened 2021-03-30
Fix logout response workflow

npalaska:
Fixes logout workflow as mentioned in the issue https://github.com/distributed-system-analysis/pbench/issues/2130

Right now the logout workflow returns `401` if an invalid `Authorization` token is presented in a request, however, in order to make logout smooth for the user and not being picky about logging out is less prone to security issues. We almost always want to return success on logout if the token is expired or is already removed from our database and no longer tracked.

### DIFF
--- a/lib/pbench/server/api/auth.py
+++ b/lib/pbench/server/api/auth.py
@@ -79,7 +79,7 @@ class Auth:
                 )
                 abort(
                     401,
-                    message="Malformed Authorization header, request auth needs bearer token: Bearer <session_token>",
+                    message="Malformed Authorization header, request needs bearer token: Bearer <session_token>",
                 )
             return auth_token
 
--- a/lib/pbench/server/api/resources/users_api.py
+++ b/lib/pbench/server/api/resources/users_api.py
@@ -253,7 +253,6 @@ class Logout(Resource):
         self.logger = logger
         self.auth = auth
 
-    @Auth.token_auth.login_required()
     def post(self):
         """
         post request for logging out a user for the current auth token.
@@ -269,17 +268,23 @@ class Logout(Resource):
                         "message": "failure message"
                     }
         """
-        auth_token = self.auth.token_auth.get_auth()["token"]
-        user = self.auth.token_auth.current_user()
+        auth_token = self.auth.get_auth_token(self.logger)
+        user = Auth.verify_auth(auth_token=auth_token)
 
-        try:
-            ActiveTokens.delete(auth_token)
-            self.logger.info(
-                "Auth token entry deleted for user with username {}", user.email
-            )
-        except Exception:
-            self.logger.exception("Exception occurred during deleting the user entry")
-            abort(500, message="INTERNAL ERROR")
+        # "None" user represents that either the token is not present in our database or it is an expired token.
+        # Expired token is already deleted by now if we reach here.
+        # In either case we will return suceess to a client, since the same token can not be used again.
+
+        if user:
+            try:
+                ActiveTokens.delete(auth_token)
+            except Exception:
+                self.logger.exception("Exception occurred deleting an auth token")
+                abort(500, message="INTERNAL ERROR")
+            else:
+                self.logger.debug("User {} logged out", user.username)
+        else:
+            self.logger.debug("User logout with invalid or expired token")
 
         return "", 200
 
--- a/lib/pbench/test/unit/server/test_user_auth.py
+++ b/lib/pbench/test/unit/server/test_user_auth.py
@@ -398,13 +398,13 @@ class TestUserAuthentication:
             )
             assert logout_response.status_code == 200
 
-            # invalid token logout
+            # Logout using invalid token
+            # Expect 200 on response, since the invalid token can not be used anymore
             response = client.post(
                 f"{server_config.rest_uri}/logout",
                 headers=dict(Authorization="Bearer " + data_login["auth_token"]),
             )
-            data = response.json
-            assert data is None
+            assert response.status_code == 200
 
     @staticmethod
     def test_delete_user(client, server_config):


lib/pbench/server/api/auth.py
```diff
@@ -79,7 +79,7 @@ def get_auth_token(self, logger):
                 )
                 abort(
                     401,
-                    message="Malformed Authorization header, request auth needs bearer token: Bearer <session_token>",
+                    message="Malformed Authorization header, request authorization header needs bearer token: Bearer <session_token>",
```
portante, line 82: Perhaps just remove `authorization header` since those two exact words are written before `request`.  Avoids invoking the Department of the Redundancy Department.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -269,14 +268,24 @@ def post(self):
...
-        user = self.auth.token_auth.current_user()
+        auth_token = self.auth.get_auth_token(self.logger)
+        user = Auth.verify_auth(auth_token=auth_token)
+
+        # "None" user represents that either the token is not present in our database or it is an expired token.
+        # Expired token is already deleted by now if we reach here.
+        # In either case we will return suceess to a client, since the same token can not be used again.
 
         try:
             ActiveTokens.delete(auth_token)
```
portante, line 279: Is an exception raised when a previously deleted `auth_token` is used?

lib/pbench/server/api/resources/users_api.py
```diff
@@ -269,14 +268,24 @@ def post(self):
...
-            self.logger.info(
-                "Auth token entry deleted for user with username {}", user.email
-            )
+            # Do we need to log every time someone logs out
+            if user:
+                self.logger.info(
```
portante, line 282: Seems like this could be `debug()`.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -269,14 +268,24 @@ def post(self):
...
+        # "None" user represents that either the token is not present in our database or it is an expired token.
+        # Expired token is already deleted by now if we reach here.
+        # In either case we will return suceess to a client, since the same token can not be used again.
 
         try:
             ActiveTokens.delete(auth_token)
-            self.logger.info(
-                "Auth token entry deleted for user with username {}", user.email
-            )
+            # Do we need to log every time someone logs out
```
dbutenhof, line 280: We'll want an audit log system to track basically everything; if we had one, I suppose maybe it'd be better not to waste "debug log" space with stuff like this. For now, this is our only way to track it, and I'd say the entry is worthwhile.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -269,14 +268,24 @@ def post(self):
...
+            if user:
+                self.logger.info(
+                    "Auth token entry deleted for user with email {}", user.email
+                )
+            else:
+                self.logger.info(
```
portante, line 286: We might want to introduce a counter for certain behaviors instead of logging them every time.  Not sure this is an informational, but at least `.debug()`.

portante (changes requested):

lib/pbench/server/api/resources/users_api.py
```diff
@@ -269,14 +268,24 @@ def post(self):
...
 
         try:
             ActiveTokens.delete(auth_token)
-            self.logger.info(
-                "Auth token entry deleted for user with username {}", user.email
-            )
+            # Do we need to log every time someone logs out
+            if user:
+                self.logger.info(
+                    "Auth token entry deleted for user with email {}", user.email
```
dbutenhof, line 283: ... but I think I'd prefer directly saying that the user logged out rather than reporting in terms of the auth token.

Also, reporting the username is probably more useful right now. If we drop that in favor of email (of which I'm not in favor ;-) ), we'll figure out what to do then.

lib/pbench/test/unit/server/test_user_auth.py
```diff
@@ -398,13 +398,13 @@ def test_invalid_logout(client, server_config):
             )
             assert logout_response.status_code == 200
 
-            # invalid token logout
+            # Logout using invalid token
+            # Expect 200 on response, since the invalid token can not be used anymore
```
dbutenhof, line 402: Of course if it was really an "invalid token" then it could never be used in the first place. More like, either the token was garbage or it's been deleted, and either way cannot be used to authorize access.

lib/pbench/server/api/resources/users_api.py
portante: On the server side username is fine to report in the logs.  The use of "email" to login is a user-interface thing, not an implementation thing.

lib/pbench/server/api/resources/users_api.py
npalaska: No, deleting a row that doesn't exists is not an error or exception.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -279,11 +279,9 @@ def post(self):
             ActiveTokens.delete(auth_token)
             # Do we need to log every time someone logs out
```
dbutenhof, line 280: This comment should probably be deleted now.

dbutenhof (approved):

lib/pbench/server/api/resources/users_api.py
```diff
@@ -269,14 +268,22 @@ def post(self):
                         "message": "failure message"
                     }
         """
-        auth_token = self.auth.token_auth.get_auth()["token"]
-        user = self.auth.token_auth.current_user()
+        auth_token = self.auth.get_auth_token(self.logger)
```
portante, line 271: So this only fetches the `auth_token` from the API call, right?  It does not look the token up in the database?

lib/pbench/server/api/resources/users_api.py
```diff
@@ -269,14 +268,22 @@ def post(self):
...
+        user = Auth.verify_auth(auth_token=auth_token)
```
portante, line 272: This is the call that actually looks up the token in the database to find the user associated with the token?

lib/pbench/server/api/resources/users_api.py
portante: So why are we executing a database operation to remove the token if we found that it is not associated with the user?

Should this be:

```
if user:
    try:
        ActiveTokens.delete(auth_token)
    except Exception:
        self.logger.exception("Exception ...")
        abort(500, ...)
    else:
        self.logger.debug("User {} logged out", user.username)
else:
    self.logger.debug("User logout with invalid or expired token")

return "", 200
```

lib/pbench/server/api/resources/users_api.py
```diff
@@ -269,14 +268,22 @@ def post(self):
...
-            )
+            # Do we need to log every time someone logs out
+            if user:
+                self.logger.debug("User {} logged out", user.username)
+            else:
+                self.logger.debug(
+                    "Logout: Token is not present in the db or user tried an expired token"
+                )
         except Exception:
             self.logger.exception("Exception occurred during deleting the user entry")
```
portante, line 288: So perhaps we should word this something like, `Exception occurred deleting an auth token`?

portante (changes requested):

lib/pbench/server/api/resources/users_api.py
dbutenhof: Agreed; it may be a silent no-op to SQLAlchemy, but it's going to be a constant semantic confusion to anyone reading the code and there's no point in calling SQLAlchemy for an expensive no-op when we know there's nothing to do.

lib/pbench/server/api/resources/users_api.py
npalaska: Yes

lib/pbench/server/api/resources/users_api.py
npalaska: That is a good point

dbutenhof (dismissed):

portante (dismissed):

portante (approved):

dbutenhof (approved):

PR #2196, opened 2021-04-01
Pcp transient main

Maxusmusti:
Original: https://github.com/portante/pbench/pull/14
 - Moved to be on top of main branch now that PR #2076 has been merged
 - Adds a transient pcp-tool that creates local pmloggers and pmcds on remote nodes, and starts/stops/sends alongside transient tools, rather than permanently running in background.
 - Tool is currently fully functional according to both traditional and containerized tests.


portante:
@Maxusmusti, let's squash these commits down to one now, and perhaps drop the repeated comment from PR #2076 in favor of just one commit message describing this work.

lib/pbench/agent/tool_meister.py
```diff
@@ -898,6 +1016,8 @@ def init_tools(self, data):
         for name, tool_opts in sorted(self._tools.items()):
             try:
                 tklass = self._tool_name_class_mappings[name]
+                if name not in self.persistent_tool_names:
```
portante, line 1019: I'd move this out of the `try` block as the first action taken on the `name`, since we don't need to find a persistent tool class mapping if it is not a persistent tool.

portante (changes requested):
Good change, not sure about how to solve the duplication of PCP commands from the persistent tools, we may to use a "mix-in" class (see https://www.residentmar.io/2019/07/07/python-mixins.html) to address that, and I am also not sure about the `-transient` suffix everywhere, seems like that could be in the tool metadata somehow, but would have to think about it.  Just one change to consider inline ...

agent/util-scripts/get-internal-tool
```diff
@@ -0,0 +1,32 @@
...
+import json
+from pathlib import Path
+
+if len(sys.argv) != 4:
+    print("Usage: get-internal-tool <metadata_location> <tool_name> <tool_type>")
+    sys.exit()
+
+meta_loc = Path(sys.argv[1])
+name = sys.argv[2]
+tool_type = sys.argv[3]
```
dbutenhof, line 14: I realize this is a small and relatively simple tool; but it'd be easier to read and maintain with `ArgumentParser` all the same.

agent/util-scripts/get-internal-tool
```diff
@@ -0,0 +1,32 @@
...
+
+try:
+    meta = json.load(open(meta_loc / "meta.json"))
+except Exception as exc:
+    print(f"Failed to load tool metadata: {exc}")
+    sys.exit()
```
dbutenhof, line 20: Might be cleaner to exit with 1 for failures, even if the caller doesn't care now..

And, yeah, more complicated; but a trampoline that calls a method and does a single `sys.exit` with the return status would make it flow better.

agent/util-scripts/pbench-register-tool
```diff
@@ -175,6 +175,16 @@ while true; do
 		shift
 		_do_test_labels=1
 		;;
+	--persistent)
+		shift
+		oldname=${name}
```
dbutenhof, line 180: The name `oldname` ... This is the original value of the `--name`, which you're changing to select a particular class of tool using that name, but which you still want to report at the end. How about `display_name` or `original_name`??

agent/util-scripts/pbench-register-tool
```diff
@@ -125,7 +125,7 @@ function usage() {
 }
 
 # Process options and arguments
-opts=$(getopt -q -o hn:g:r:l: --longoptions "help,name:,group:,no-install,remote:,remotes:,label:,labels:,test-labels" -n "getopt.sh" -- "${@}")
+opts=$(getopt -q -o hn:g:r:l: --longoptions "help,name:,group:,no-install,remote:,remotes:,label:,labels:,test-labels, persistent, transient" -n "getopt.sh" -- "${@}")
```
dbutenhof, line 128: You should add your new options to the `usage` block above.

agent/util-scripts/pbench-register-tool
```diff
@@ -175,6 +175,16 @@ while true; do
 		shift
 		_do_test_labels=1
 		;;
+	--persistent)
+		shift
+		oldname=${name}
+		name=$(${pbench_bin}/util-scripts/get-internal-tool ${pbench_bin}/tool-scripts ${name} persistent)
+		;;
+	--transient)
```
dbutenhof, line 183: What happens if they don't specify `--persistent` or `--transient` for a "base" toolname that can be in either? Are we not handling that case at all? Should we default to one or the other, or complain that the tool class needs to be specified?

Also, what happens if someone specifies a persistent-only tool with --transient, or vice versa? Does the user get a meaningful and useful error message?

Maxusmusti: By default, the tools "base name" represents the original tool type. Since the pcp tool is default persistent, simply doing a register for name=pcp will register the internal pcp tool (persistent). If pcp is registered with --persistent, it will check if "pcp" is a persistent tool already, and register it just the same. If it wasn't, then it would check for an internal "pcp-persistent", and register that internally. For pcp --transient it checks for a transient base tool called "pcp", fails to find it, then checks if there is a "pcp-transient". 

If mpstat was registered with nothing or --transient, it would be registered as "mpstat" internally, a default transient tool. If it was attempted with --persistent right now, it would fail to find a persistent "mpstat", then fail to find an internal "mpstat-persistent", and ultimately print a message saying "Could not find mpstat with type persistent... (insert rest of failure-to-register message + usage/tool options)". Eventually though (according to future plans), an "mpstat-persistent" will be made and then that option will work.

agent/util-scripts/pbench-register-tool
```diff
@@ -202,7 +205,11 @@ if [[ -z ${name} ]]; then
 	exit 1
 fi
 if [[ ! -x "${pbench_bin}/tool-scripts/${name}" ]]; then
```
dbutenhof, line 207: I suppose this will catch the case where `name` is an error message, but it won't look pretty.

lib/pbench/agent/tool_meister.py
```diff
@@ -586,6 +703,7 @@ def fetch_params(params):
         "dcgm": DcgmTool,
         "node-exporter": NodeExporterTool,
         "pcp": PcpTool,
+        "pcp-transient": PcpTransTool,
```
webbnh, line 706: Can you fix (or remove!) the count in line 699?  :-D

webbnh (changes requested):
Looks generally good.  I found a few things for you to consider addressing.

The biggest one is, what is the deal with `agent/tool-scripts/pcp-transient`?  It claims to be deprecated but looks like new code?  Should it be omitted from this PR, or is it supposed to be a placeholder for a future tool (in which case, "deprecated" is the wrong word!)?

agent/tool-scripts/pcp-transient
```diff
@@ -0,0 +1,17 @@
...
+PROG = os.path.basename(sys.argv[0])
+
+if len(sys.argv) != 2 or sys.argv[1] != "--help":
+    print(f"{PROG}: This script is deprecated, please run it with --help for info on registering the tool.", file=sys.stderr)
+    print(f"{PROG}: Run {os.path.abspath(__file__)} --help for more info.", file=sys.stderr)
+    sys.exit(1)
+
+if sys.argv[1] == "--help":
+    print(f"{PROG}: Options:", file=sys.stdout)
+    print(f"{PROG}: No tool-specific options currently", file=sys.stdout)
```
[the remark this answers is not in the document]
Maxusmusti: This is unfortunately a relic of the old tool system. Tools used to have bash scripts in this folder that contained instructions for their start, stop, postprocess, etc. Now that is all encapsulated within the tool meister (for new tools), but the tool registration system uses the existence of those files to check whether a tool can be registered. Additionally, it will always recommend you run `<path>/tool-scripts/<tool_name> --help` to see all of the tool-specific options. These confusing messages actually mirror the functionality (and exact text output) of almost every other existing tool currently, so I'll open up a new issue to fix this poorly-aging system.

agent/util-scripts/pbench-register-tool
```diff
@@ -192,7 +211,11 @@ if [[ -z ${name} ]]; then
...
-	error_log "Could not find ${name} in ${pbench_bin}/tool-scripts/: has this tool been integrated into the pbench-agent?"
+	if [[ ${name} == "error" && "${display_name}" != "" ]]; then
+		error_log "Could not find ${display_name} with type ${type} in ${pbench_bin}/tool-scripts/: has the right type been specified?"
+	else
+		error_log "Could not find ${name} in ${pbench_bin}/tool-scripts/: has this tool been integrated into the pbench-agent?"
+	fi
```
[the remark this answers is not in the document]
Maxusmusti: Removed displayname check, but the question at the end was to match the existing implementation. I would be down to remove them though, no preference either way. @portante thoughts?

agent/util-scripts/pbench-register-tool
```diff
@@ -115,6 +115,7 @@ function usage() {
 	printf -- "\tof the file should contain a host name, optionally followed by a label\n"
 	printf -- "\tseparated by a comma (\",\"); empty lines are ignored, and comments are\n"
 	printf -- "\tdenoted by a leading hash, or pound (\"#\"), character.\n"
+	printf -- "\n\t--persistent or --transient can be used to specify tool run type.\n"
 	printf -- "\nAvailable tools:\n"
 	local tool=""
 	python3 -c "import sys, json; meta = json.load(open(sys.argv[1])); [print(f'\t{tool}') for tool in (*meta['transient'].keys(), *meta['persistent'].keys()) ]" ${pbench_bin}/tool-scripts/meta.json
```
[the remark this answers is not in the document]
Maxusmusti: This would mix all the persistent and transient tools together, though to be fair, there is currently nothing good indicating what is what, so I agree a change is needed. I could change it to output:

Available Tools:
-Transient:
    <tool>
    <tool>
-Persistent:
    <tool>
    <tool>
    
    or something along those lines... @portante thoughts?

agent/tool-scripts/pcp-transient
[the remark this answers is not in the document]
Maxusmusti: Addressed in: https://github.com/distributed-system-analysis/pbench/issues/2204

agent/util-scripts/pbench-register-tool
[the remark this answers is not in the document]
webbnh: I would be good with either approach.

(I thought about suggesting adding "headers", but didn't want to muddy the waters....)

[FWIW, I think Peter is on hiatus wrt reviewing Server-side code, so, if you really want his opinion, I think you'll need to ping him harder than just an @mention.]

If you would prefer to add headers rather than sorting the lists together, I say go for it.

agent/util-scripts/pbench-register-tool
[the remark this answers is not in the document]
webbnh: If it matches the existing stuff, I won't put up too much stink (at the moment...).

Let's move along.

lib/pbench/agent/tool_meister.py
```diff
@@ -328,7 +328,12 @@ def __init__(self, name, tool_opts, logger=None, **kwargs):
         self.pmcd_process = None
         self.pmlogger_process = None
         if self.tool_dir:
-            self.tool_dir.mkdir()
+            try:
+                self.tool_dir.mkdir()
+            except Exception as exc:
+                self.logger.error(
+                    "Failed to create tool directory '%s': '%s'", self.tool_dir, exc
+                )
```
webbnh, line 336: Is there a scenario in which this directory would already exist, in which case we might not want to bomb out, here?  (E.g., if I create the tool, then remove it, then try to create it again?)  Also, this directory is/could be on the remote system where the benchmark workload will run (right?), so it might be inconvenient to delete this directory once it has been created.

This is all a long-winded way of asking whether we should be specifying `exists_ok=True` on the `mkdir()` call....

webbnh (approved):
Looks good.  I say "Ship it!".

I do have one question, but don't feel like you have to gate the merge on fixing it.

webbnh:
...but, you need a rebase....

lib/pbench/agent/tool_meister.py
Maxusmusti: I don't think a scenario currently exists, as other samples/iterations will be under different parent directories, and the whole run will be under a timestamped directory, so no room for re-use currently. Also no way for it to (intentionally at least) attempt to create multiple times for a single sample/iteration. There could potentially be optimizations in the future that involve reusing the created dir though, so perhaps useful for the future.

webbnh (dismissed):

agent/util-scripts/pbench-register-tool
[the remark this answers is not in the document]
dbutenhof: > [FWIW, I think Peter is on hiatus wrt reviewing Server-side code, so, if you really want his opinion, I think you'll need to ping him harder than just an @mention.]

You knew this, right? For posterity, this is agent code.

> If you would prefer to add headers rather than sorting the lists together, I say go for it.

Yeah, making this sort of display unambiguously clear and straightforward isn't easy; I'm OK with leaving it as it is.

agent/util-scripts/pbench-register-tool
```diff
@@ -175,6 +176,24 @@ while true; do
...
+		fi
+		type="persistent"
+		;;
+	--transient)
+		shift
+		display_name=${name}
+		name=$(${pbench_bin}/util-scripts/get-internal-tool ${pbench_bin}/tool-scripts ${name} transient)
+		if [[ ${name} == "" ]]; then
+			name="internal-tool-not-found"
+		fi
```
dbutenhof, line 194: Well, OK; unwieldy, and I always generally dislike repurposing a variable for error handling; but, fine.

dbutenhof (approved):

agent/util-scripts/pbench-register-tool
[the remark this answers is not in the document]
webbnh: > For posterity, this is agent code.

LOL...so it is!  I've been doing so much back-and-forth between reviews that I forgot where I was.  :-(

agent/util-scripts/pbench-register-tool
Maxusmusti: Oh wait a second, that should've also been replaced with a `-z "${name}"` in the condition...

Maxusmusti:
I noticed a quick fix to be made, and also decided to add in the headers for a clearer message.

dbutenhof (dismissed):

dbutenhof:
Of course you actually need to rebase since I just merged #2206 (sorry ;-) )

portante (approved):
This code is pushing us to the edge of the usefulness of `pbench-register-tool` in `bash`.

dbutenhof (approved):

agent/util-scripts/pbench-register-tool
```diff
@@ -115,17 +115,18 @@ function usage() {
 	printf -- "\tof the file should contain a host name, optionally followed by a label\n"
 	printf -- "\tseparated by a comma (\",\"); empty lines are ignored, and comments are\n"
 	printf -- "\tdenoted by a leading hash, or pound (\"#\"), character.\n"
+	printf -- "\n\t--persistent or --transient can be used to specify tool run type.\n"
 	printf -- "\nAvailable tools:\n"
 	local tool=""
-	python3 -c "import sys, json; meta = json.load(open(sys.argv[1])); [print(f'\t{tool}') for tool in (*meta['transient'].keys(), *meta['persistent'].keys()) ]" ${pbench_bin}/tool-scripts/meta.json
+	python3 -c "import sys, json; meta = json.load(open(sys.argv[1])); print('-Transient:'); [print(f'\t{tool}') for tool in meta['transient'].keys()]; print('-Persistent:'); [print(f'\t{tool}') for tool in meta['persistent'].keys()]" ${pbench_bin}/tool-scripts/meta.json
```
webbnh, line 121: I don't want to hold up this PR for this, but, if you find some other reason to make another change....

I really like the headers, but having them preceded by a single hyphen is (as Dave would say) jarring.  You could use the options as headers, in which case they should be preceded by _two_ hyphens (and down-cased).  Or, you could just use spaces.

Also, you can produce the output reasonably straightforwardly with a single `print()` statement:
```
 print('  Transient:', *[f'\t{tool}' for tool in meta['transient'].keys()], '  Persistent:', *[f'\t{tool}' for tool in meta['persistent'].keys()], sep='\n')
```

webbnh (approved):
Looks good -- I like the headers.

One comment and suggestion, but it doesn't need to gate the merge.

portante (changes requested):
Needs a rebase and a squash ... looks good otherwise (though addressing @webbnh's last comment might be worth it.

Maxusmusti:
Rebased, squashed, and added suggested change. Also, the makefile needed to have "validate-hostname" and "validate-ipaddress" from the latest commit on main.

dbutenhof (approved):

portante:
> Rebased, squashed, and added suggested change. Also, the makefile needed to have "validate-hostname" and "validate-ipaddress" from the latest commit on main.

Had filed https://github.com/distributed-system-analysis/pbench/pull/2212.

Maxusmusti:
Just rebased on the validate-* fix

portante (approved):

dbutenhof (approved):

Issue #2183, opened 2021-03-29
CLI user management tool


dbutenhof:
We could use a CLI user management tool. (Note we may get this for free with integration into a support identity manager, but in the interim this would be convenient for testing and administration.)

This would allow listing, creating, modifying, and deleting users from the server command line.

Among other things, our setup procedure might use this to create the "admin" user on installation.

webbnh:
@npalaska, I suggest that you ask Nikhil -- he's looking into that.

webbnh:
Presumably, we'll have/want a GUI for this -- a web page which lists users and offers various ways of manipulating them.  But, as Dave says, we'll presumably want a CLI version as well, for all the reasons that he lists.

But, our user management capabilities will be greatly dependent upon our implementation.  If we end up integrating RH SSO, then the server-side support for user management will have to be able to respond to requests from the GUI or the CLI by interacting with SSO.  If we go with some other identity management system (Auth0, Cognito, you-name-it), then what we can do and how we do it will be dictated by the system's capabilities.

But, for the moment, we can presume that we'll be able to

- list existing users
- delete users
- create users
- modify users' characteristics and capabilities (whatever those are and/or evolve to be)

I expect that what Dave is envisioning is another simple tool like `pbench-generate-token` which grabs a few items from the user and makes a request to the server REST API.  Presumably it will need a token...which creates an interesting chicken-and-egg problem for creating the "admin" user on installation, but maybe we can find a clever way around that (like, if the server determines that are no existing users, then it's willing to accept the first request even without a valid token, especially if that request is to create the admin user...).

(And, yes, I know that @npalaska _is_ Nikhil... :-D )

PR #2197, opened 2021-04-01
Add pbench create user CLI

npalaska:
Addresses `create_user` from issue https://github.com/distributed-system-analysis/pbench/issues/2183

This PR lets us create pbench user from command line. 

The `pbench_create_user` tool accepts a username, password, email, first_name, last_name as a required parameters. These can be specified on the command line. However, if any of the parameter is not specified on the command line, the tool will prompt for them. The response to the password prompt is not echoed. The tool also optionally accept `--exist-ok` option, and if specified it will ignore the error if the user already exists and return success.

### DIFF
--- a/lib/pbench/cli/server/options.py
+++ b/lib/pbench/cli/server/options.py
@@ -1,0 +1,39 @@
+import click
+from typing import Callable
+
+from pbench.cli.server import CliContext
+
+
+def common_options(in_f: Callable) -> Callable:
+    """
+    This function can be used as a decorator for multiple server side CLI utilities,
+    where it receives a click command decorated function as an argument and sets some
+    common configuration options on the CLI, in this case pbench server configuration
+    :param in_f: click command decorated function
+    :return: function with common click options set
+    """
+    out_f = _pbench_server_config(in_f)
+    return out_f
+
+
+def _pbench_server_config(f: Callable) -> Callable:
+    """Option for server configuration"""
+
+    def callback(ctx, param, value):
+        clictx = ctx.ensure_object(CliContext)
+        clictx.config = value
+        return value
+
+    return click.option(
+        "-C",
+        "--config",
+        required=True,
+        envvar="_PBENCH_SERVER_CONFIG",
+        type=click.Path(exists=True, readable=True),
+        callback=callback,
+        expose_value=False,
+        help=(
+            "Path to a pbench-server configuration file (defaults to the "
+            "'_PBENCH_SERVER_CONFIG' environment variable, if defined)"
+        ),
+    )(f)
--- a/lib/pbench/cli/server/user_create.py
+++ b/lib/pbench/cli/server/user_create.py
@@ -1,0 +1,84 @@
+import click
+from pbench import BadConfig
+from pbench.server import PbenchServerConfig
+from pbench.cli.server import pass_cli_context
+from pbench.cli.server.options import common_options
+from pbench.common.logger import get_pbench_logger
+from pbench.server.database.models.users import User
+from pbench.server.database.database import Database
+
+_NAME_ = "pbench-user-create"
+
+
+class UserCreate:
+    def __init__(self, context):
+        self.context = context
+
+    def execute(self):
+        config = PbenchServerConfig(self.context.config)
+
+        logger = get_pbench_logger(_NAME_, config)
+
+        # We're going to need the Postgres DB to track dataset state, so setup
+        # DB access.
+        Database.init_db(config, logger)
+
+        user = User(
+            username=self.context.username,
+            password=self.context.password,
+            first_name=self.context.first_name,
+            last_name=self.context.last_name,
+            email=self.context.email,
+        )
+        user.add()
+        click.echo(f"User {self.context.username} registered")
+
+
+@click.command()
+@common_options
+@click.option(
+    "--username",
+    prompt=True,
+    required=True,
+    help="pbench server account username (will prompt if unspecified)",
+)
+@click.option(
+    "--password",
+    prompt=True,
+    hide_input=True,
+    required=True,
+    help="pbench server account password (will prompt if unspecified)",
+)
+@click.option(
+    "--email",
+    prompt=True,
+    required=True,
+    help="pbench server account email (will prompt if unspecified)",
+)
+@click.option(
+    "--first-name",
+    prompt=True,
+    required=True,
+    help="pbench server account first name (will prompt if unspecified)",
+)
+@click.option(
+    "--last-name",
+    prompt=True,
+    required=True,
+    help="pbench server account last name (will prompt if unspecified)",
+)
+@pass_cli_context
+def main(context, username, password, email, first_name, last_name):
+    context.username = username
+    context.password = password
+    context.email = email
+    context.first_name = first_name
+    context.last_name = last_name
+
+    try:
+        rv = UserCreate(context).execute()
+    except Exception as exc:
+        click.echo(exc, err=True)
+        rv = 2 if isinstance(exc, BadConfig) else 1
+
+    click.get_current_context().exit(rv)
--- a/server/Makefile
+++ b/server/Makefile
@@ -19,6 +19,9 @@ INSTALL = install
 #INSTALLOPTS = --mode 755 --directory --owner=${OWNER} --group=${GROUP}
 INSTALLOPTS = --directory
 
+click-scripts = \
+	pbench-user-create
+
 # targets
 .PHONY: install \
 	install-dirs \
@@ -41,7 +44,7 @@ install-python3-setup: install-bin install-lib
 	mkdir -p ${DESTDIR}/python3
 	${COPY} requirements.txt ${DESTDIR}
 	(cd ..; SKIP_GENERATE_AUTHORS=1 SKIP_WRITE_GIT_CHANGELOG=1 python3 setup.py install --prefix=${DESTDIR}/python3)
-	${COPY} ${DESTDIR}/python3/bin/pbench-config ${DESTDIR}/python3/bin/pbench-server ${BINDIR}/
+	${COPY} $(addprefix ${DESTDIR}/python3/bin/, pbench-config pbench-server ${click-scripts}) ${BINDIR}/
 	${RM} -r ${DESTDIR}/python3
 	${COPY} ../lib/pbench ${LIBDIR}/
 	${RM} -r $$(find ${LIBDIR} -name __pycache__) ${LIBDIR}/pbench/test ${LIBDIR}/pbench/agent ${LIBDIR}/pbench/cli/agent
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,6 +32,7 @@ console_scripts =
    pbench-list-triggers = pbench.cli.agent.commands.triggers.list:main
    pbench-register-tool-trigger = pbench.cli.agent.commands.triggers.register:main
    pbench-server = pbench.cli.server.shell:main
+   pbench-user-create = pbench.cli.server.user_create:main
 
 [tools:pytest]
 testpaths = lib/pbench/test
--- a/lib/pbench/cli/server/__init__.py
+++ b/lib/pbench/cli/server/__init__.py
@@ -1,0 +1,21 @@
+"""
+Create a click context object that holds the state of the server
+invocation. The CliContext will keep track of passed parameters,
+what command created it, which resources need to be cleaned up,
+and etc.
+
+We create an empty object at the beginning and populate the object
+with configuration, group names, at the beginning of the server
+execution.
+"""
+
+import click
+
+
+class CliContext:
+    """Initialize an empty click object"""
+
+    pass
+
+
+pass_cli_context = click.make_pass_decorator(CliContext, ensure=True)
--- a/lib/pbench/test/unit/server/test_user_create.py
+++ b/lib/pbench/test/unit/server/test_user_create.py
@@ -1,0 +1,77 @@
+from click.testing import CliRunner
+import responses
+
+from pbench.cli.server import user_create
+
+
+class TestCreateUser:
+    USER_SWITCH = "--username"
+    PSWD_SWITCH = "--password"
+    PSWD_PROMPT = "Password: "
+    EMAIL_SWITCH = "--email"
+    FIRST_NAME_SWITCH = "--first-name"
+    LAST_NAME_SWITCH = "--last-name"
+    USER_TEXT = "test_user"
+    PSWD_TEXT = "password"
+    EMAIL_TEXT = "test@domain.com"
+    FIRST_NAME_TEXT = "First"
+    LAST_NAME_TEXT = "Last"
+    URL = "http://pbench.example.com/api/v1"
+    ENDPOINT = "/register"
+
+    @staticmethod
+    @responses.activate
+    def test_help(pytestconfig):
+        runner = CliRunner(mix_stderr=False)
+        result = runner.invoke(user_create.main, ["--help"])
+        assert result.exit_code == 0
+        assert str(result.stdout).startswith("Usage:")
+        assert not result.stderr_bytes
+
+    @staticmethod
+    @responses.activate
+    def test_args(server_config, pytestconfig):
+        runner = CliRunner(mix_stderr=False)
+        result = runner.invoke(
+            user_create.main,
+            args=[
+                TestCreateUser.USER_SWITCH,
+                TestCreateUser.USER_TEXT,
+                TestCreateUser.PSWD_SWITCH,
+                TestCreateUser.PSWD_TEXT,
+                TestCreateUser.EMAIL_SWITCH,
+                TestCreateUser.EMAIL_TEXT,
+                TestCreateUser.FIRST_NAME_SWITCH,
+                TestCreateUser.FIRST_NAME_TEXT,
+                TestCreateUser.LAST_NAME_SWITCH,
+                TestCreateUser.LAST_NAME_TEXT,
+            ],
+        )
+        assert result.stdout_bytes == b"User test_user registered\n"
+        assert result.exit_code == 0
+        assert not result.stderr_bytes
+
+    @staticmethod
+    @responses.activate
+    def test_valid_user_registration(server_config, pytestconfig):
+        runner = CliRunner(mix_stderr=False)
+        result = runner.invoke(
+            user_create.main,
+            args=[
+                TestCreateUser.USER_SWITCH,
+                TestCreateUser.USER_TEXT,
+                TestCreateUser.EMAIL_SWITCH,
+                TestCreateUser.EMAIL_TEXT,
+                TestCreateUser.FIRST_NAME_SWITCH,
+                TestCreateUser.FIRST_NAME_TEXT,
+                TestCreateUser.LAST_NAME_SWITCH,
+                TestCreateUser.LAST_NAME_TEXT,
+            ],
+            input=f"{TestCreateUser.PSWD_TEXT}\n",
+        )
+        assert result.exit_code == 0
+        assert (
+            result.stdout
+            == f"{TestCreateUser.PSWD_PROMPT}\n" + "User test_user registered\n"
+        )
+        assert not result.stderr_bytes


lib/pbench/cli/agent/commands/pbench_create_user.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+class PbenchCreateUser(BaseCommand):
+    def __init__(self, context):
+        super().__init__(context)
+
+    def execute(self):
+        """Creates a user in the pbench user database"""
+
+        server = self.config.agent.parser.get("results", "server_rest_url")
```
dbutenhof, line 16: This is interesting; I don't think we want this to be set up as an "agent" utility, with the `pbench-agent.cfg` setting for `server_rest_url`, which requires a `pbench-agent` installation. I think we want to base this on the `pbench-server.cfg` configuration and `rest_uri`. In theory we could allow users to self-register through the CLI, and conceivably that could make this also useful as part of the agent installation, but it's mostly a server administration tool. (And, selfishly, this does me no good for solving my server-side gold unittest problem if it depends on an agent configuration environment!)

It might not be a bad idea to try to normalize the server configuration settings between the agent and server side config files and possibly even move this to a `common` tool set that's not restricted to either server or agent. For simplicity, though, I suggest that we start out as a server management utility using server configuration.

dbutenhof (commented):
This is a great start, and it'd solve my immediate requirement for unit testing except that you've made it dependent on the `pbench-agent.cfg` environment.

I'm glad to see you're using the click infrastructure; and while I'd considered trying that for my utilities, I wasn't sure how much of the infrastructure Chuck set up would translate to the server side, and decided I didn't feel like experimenting. At first I thought you'd done that, and then realized you're doing this as an agent-side utility.

I'd originally imagined something more like a `pbench-user-manager` command that would allow a server administrator to create, edit, and delete users from the CLI. This is more complicated, and obviously would require authorization as admin or user in order to change or delete a user.

Maybe doing this simple "register a user" agent utility is a reasonable place to start, even though it doesn't help with the problem that prompted me to push for dealing with this issue, but we definitely need it accessible on the server side as well... assuming we even want it on the agent side at all.

Also, I believe that for this to install properly you need to add the click path into `setup.cfg` with the other click tools. Right now, that includes both agent and server paths. (And, yeah, I guess we _can_ use click on the server side, because the server shell already does so... I'd forgotten that.)

webbnh (changes requested):
I think there are several, significant, unresolved architectural issues here -- see my comment about `exists_ok`, leading to the discussion of whether this is a server-side or agent-side tool (I think it's neither -- I think it's a "client-side" tool!).

lib/pbench/cli/agent/commands/pbench_create_user.py
```diff
@@ -0,0 +1,101 @@
...
+        except requests.exceptions.ConnectionError as exc:
+            raise RuntimeError(f"Cannot connect to '{uri}'") from exc
+
+        if response.status_code == 201:
+            click.echo(f"User {self.context.username} registered")
+            return 0
+
+        if self.context.exists_ok and response.status_code == 403:
+            click.echo(f"User {self.context.username} already registered")
+            return 0
```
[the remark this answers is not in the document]
dbutenhof: In terms of transient hackery, `exists_ok` is about the _gold_ unit tests and test isolation, with the idea that a test's `.db` companion file can specify users and the `unittests` script will create them. One option is for the test script to start out creating a certain set of users and unit tests are required to use (only) those; however as a completely isolated per-test setup, the "create if doesn't exist" made more sense. I suppose another option would be for the script to do `pbench-create-user` (which reminds me, all the other command names are `-` separated; this one is `_` separated, and should probably be changed) and have it simply return the HTTP status for errors and the script would ignore an "already exists" error.

`pbench-move-results` is definitely an agent command, as that's where performance results are generated, and it doesn't have much meaning elsewhere. I suppose that `pbench-generate-token` is potentially more nuanced, particularly when we start supporting token roles. It could hardcode a "PUT-only" token, or it could be evolved with an option to specify the token role in which case it could be used to generate a token for user delete/modify by the owner or admin.

In theory, `pbench-create-user` could be a convenience if someone is generating data and hasn't already gone to the UI to register; I'm not sure that's such a big deal at this point, but it's a possibility.

General note in terms of naming ... if we're going to move to a single master command with sub-commands, like `git`, it might make more sense to adopt going forward a __pbench__-_verb_-_noun_ naming style that will collapse more naturally into that schema. So maybe we want `pbench-user-create` and `pbench-token-generate` and even `pbench-results-move`? Of course that changes most all of our command names, which might be excessive...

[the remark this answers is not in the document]
webbnh: > In terms of transient hackery

Is it transient?  Do you see the need for this (e.g., in the unit tests) as going away sometime soon?  Perhaps I'm naive, but I expect that we'll be testing this functionality pretty much indefinitely.

> all the other command names are - separated; this one is _ separated, and should probably be changed

Yes, it should certainly be changed!  My understanding of the Click support is that, once everything is in place, we'll be able to invoke commands either with hyphens (`pbench-user-create`) or spaces (as subcommands, `pbench user create`).  Underscores would be parsed completely differently.

> another option would be for the script to do pbench-create-user [...] the script would ignore an "already exists" error.

That would certainly be my preference:  that keeps the tool more focused and better separates policy from mechanism.

> `pbench-move-results` is definitely an agent command

:-)  Spoken like a server-side guy.  :-D

> I suppose that `pbench-generate-token` is potentially more nuanced

Yes, I'm expecting that, given our seeming hunger for command-line tools, it's applicability is only going to grow.

> In theory, pbench-create-user could be a convenience if someone is generating data and hasn't already gone to the UI to register; I'm not sure that's such a big deal at this point, but it's a possibility.

I would be inclined to push back on that:  for human users, creating a username is a one-time task, the first step toward examining data (they may or may not decide to generate their own data to examine) -- if they don't eventually use the GUI (to set up sharing, if they don't view their data themselves), then there is no point to uploading any data, so they might as well fire up the GUI first and create their username.

> if we're going to move to a single master command with sub-commands, like git, it might make more sense to adopt going forward a pbench-verb-noun naming style

Yes, I concur.  (I should have done that with `pbench-generate-token`, but I followed the existing practice instead.)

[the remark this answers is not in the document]
npalaska: > But, that leads to a couple more questions. First, how do we know that a `403` response means "user already exists"? I'm not even sure that we should be seeing a `403`, because that status implies that the requestor has been authenticated, and this request doesn't involve that. I realize that we don't want to leak or otherwise create a way for an assailant to probe for existing users, but I'm not sure how to avoid it and still enable the `exists_ok` concept. (Now, if the request were coming from an authenticated `admin` user on behalf of a third party, life would be very different!). And, my other question centers around the fact that "already exists" implies/suggests that the existing user _matches_ the one in the creation request (same first/last name and email, in addition to the same username); so, really, before deciding whether the `403` is an error or not, we would need to fetch the existing information for the requested user, and decide whether it matches (which, of course, would require authentication, either using the supplied credentials or by running this thing as an admin in the first place).

There is no authentication layer involved when registering a user, meaning a user doesn't need to provide any auth token in the request. So at the moment there is no way to know if the requestor is an admin or someone else when they are just registering a user. We know `403` means a user already exists because thats the status code we defined for user already exist conditions (with varying messages such as duplicate username or email, because these are the only 2 unique constraints in our db). To fetch a user information the request needs pbench authentication token, which is not a required header for create user api. What I can do is I think should just return the message the api returned.

> And, that leads into a discussion touching on whether this is a "server-side" tool, an "agent-side" tool, or something else entirely. If it's a server-side tool, I'm not sure that it needs to be using the client REST interface -- it could be using some "internal" channel which we can trust without too much authentication. On the other hand, it doesn't seem like it makes sense as an agent-side tool -- maybe this is my lack of understanding -- but it doesn't seem like the _agent_ should be involved with things like user management -- that's a server concept used by clients (like the Dashboard and `pbench-move-results` which, admittedly, is an agent tool, but it's interfacing with the server to perform a user function). So, it feels like the proposed tool belongs to a new category..."client-tools". (And, presumably, `pbench-generate-token` should move into this category, and possibly `pbench-move-results`, as well.) And, once we figure out which category it belongs in, then we can figure out how to configure it to get access to the server.

The reason I choose this to be an agent side tool is because on the spectrum of agent -> server I felt its slightly more closer to agent side since most of the time the tool is more useful and convenient from outside the server machine to create a user via cli. (I considered anything outside server is agent but I agree client is more suitable terminology) We dont have the client category yet. 

And as @dbutenhof mentioned the `exists_ok` option is really for the gold unit tests and test isolation. It just ignores if there is an already existing `username` or `email` (403) and returns success without actually creating a new user. Every other error is reported back to the user.

@dbutenhof should I move this utility under server side?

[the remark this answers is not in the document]
webbnh: > There is no authentication layer involved when registering a user, meaning a user doesn't need to provide any auth token in the request. So at the moment there is no way to know if the requestor is an admin or someone else when they are just registering a user. We know 403 means a user already exists because thats the status code we defined for user already exist conditions (with varying messages such as duplicate username or email, because these are the only 2 unique constraints in our db). To fetch a user information the request needs pbench authentication token, which is not a required header for create user api. What I can do is I think should just return the message the api returned.

I agree.  So, there are various issues (some of which are or may be outside the scope of this PR):

- As currently constructed, the `/register` API endpoint leaks security information which could potentially be abused (i.e., it would allow an attacker to probe for existent usernames).  Is this or should this be a concern?  (Should we be returning something other than `403` and/or should we be less explicit with our error messages?)
- Do we want to continue to offer self-registration, or should we switch to a model which requires users to be created by a suitably-privileged, existent account?  (This creates annoying problems of its own, but most of them go away once we integrate with an external authorization facility such as LDAP or RH SSO, and it delivers us from having to cope with unauthenticated accesses.)
- For this PR, we _could_ consider validating the user-exists scenario by trying to use the supplied credentials to execute a login:  if it works, then, yes, indeed, the user does already exists and we're good to go; if it doesn't, then the user that they are trying to create already exists in an incompatible way, things aren't going to work out well, and we should report an error.

> The reason I choose this to be an agent side tool is because on the spectrum of agent -> server I felt its slightly more closer to agent side since most of the time the tool is more useful and convenient from outside the server machine to create a user via cli. (I considered anything outside server is agent but I agree client is more suitable terminology) We dont have the client category yet.

Again, I agree.  In the grand scheme of things, it seems much more likely that end-users will have access to a machine with the agent installed than that they will have access to a machine with the server installed.  So, if we want this tool to be useful to the end-user, it should not be a server-side tool.  However, what we want this tool for initially is for testing the server...which means that either it should be a server-side tool (not exactly desirable), or we need to have the agent available when we test the server (pretty much undesirable), or we need to get on with creating a third category (better than "not exactly desirable"?).

> It just ignores if there is an already existing username or email (403) and returns success without actually creating a new user.

Understood.  However, this could be dealt with in the testing setup instead of in the tool....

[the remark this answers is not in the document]
npalaska: > * As currently constructed, the `/register` API endpoint leaks security information which could potentially be abused (i.e., it would allow an attacker to probe for existent usernames).  Is this or should this be a concern?  (Should we be returning something other than `403` and/or should we be less explicit with our error messages?)

Yes I guess we can be little less explicit with our return messages, however, idea behind these return messages like (`username/email already in use`) is to just give some help to the user to understand what went wrong and then he/she can correct the request with appropriate values. Without any error/help the user would keep guessing what might have went wrong and after couple of failed attempts the user might just leave the room and not come back (we dont want that to happen, we do want user to quickly amend the request without much hassle)


> * Do we want to continue to offer self-registration, or should we switch to a model which requires users to be created by a suitably-privileged, existent account?  (This creates annoying problems of its own, but most of them go away once we integrate with an external authorization facility such as LDAP or RH SSO, and it delivers us from having to cope with unauthenticated accesses.)

This needs more discussion on how to do it and explore our options before I can dive into it. There are certainly few ways to write SSO login scripts with flask and that including all the major SSO providers (gmail, github, redhat etc). I guess we can't just rely on RH SSO since other people who do not have redhat account might want to use the product. However, we certainly need our user database as well.


> * For this PR, we _could_ consider validating the user-exists scenario by trying to use the supplied credentials to execute a login:  if it works, then, yes, indeed, the user does already exist and we're good to go; if it doesn't, then the user that they are trying to create already exists in an incompatible way, things aren't going to work out well, and we should report an error.

I am not sure but I do think this could be an overkill for such a simple utility tool, where we have to hit two rest endpoints in order to create a user that is login and logout, if we login we create pbench token in our database and then logout to remove that token (Potentially 2 database queries). Since this is a server side utility it is assumed that whoever wants to use the cli to create user should have server installed already and so we could just simply pass on the error message that rest endpoint reported.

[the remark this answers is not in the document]
dbutenhof: Chained conversations in GitHub are awkward to unwind!

> Is it transient? Do you see the need for this (e.g., in the unit tests) as going away sometime soon? Perhaps I'm naive, but I expect that we'll be testing this functionality pretty much indefinitely.

@webbnh  I see the "gold" CLI-level unit tests going away, which would remove the need for this. Not right away, unfortunately.

> The reason I choose this to be an agent side tool is because on the spectrum of agent -> server I felt its slightly more closer to agent side since most of the time the tool is more useful and convenient from outside the server machine to create a user via cli. (I considered anything outside server is agent but I agree client is more suitable terminology) We dont have the client category yet.

@npalaska As @webbnh commented, registration is a one-time occurrence per-user and I just don't see how it makes much sense to offer this as part of the agent RPM/container. And, more immediately, since the sense of urgency that caused me to push you on this was primarily motivated by the need to create users for the CLI unit tests in #2101 ... this PR, as written, does me no good at all as the server unit tests can't depend on an agent CLI environment!

Plus, again, I'd originally conceived this as a generic user management CLI tool primarily for an administrator to use as part of the server installation.

However, I can definitely see some benefit to Webb's vision of a separate "client" package that might be compatible with either server or agent, based on HTTP/S communication with an installed server... that could certainly include token generation, user management, and possibly also move/copy results. (The later based on the argument that using the agent for benchmark orchestration and results consolidation and local analysis could be a sensible use case without any interaction at all with a Pbench server, and move/copy results is only necessary if you _are_ using a server.)

(Although in practice I'm not sure we want to require installing a second package to enable this, or go to the effort right now to create such a package.)

> Yes I guess we can be little less explicit with our return messages, however, idea behind these return messages like (username/email already in use) is to just give some help to the user to understand what went wrong and then he/she can correct the request with appropriate values. Without any error/help the user would keep guessing what might have went wrong and after couple of failed attempts the user might just leave the room and not come back (we dont want that to happen, we do want user to quickly amend the request without much hassle)

@npalaska This whole topic get complicated; so many tradeoffs to consider. Some web sites I've used support "self registration" but don't allow you to just enter your info and submit... instead, you start with an email; it sends you a registration link to ensure that your email is valid, and presumably has already checked that there's no existing account with that email. The email link gets you to a secondary registration page where you get to choose a username and password now that the email has been validated. This sort of scheme makes it a bit harder to deploy an account sniffer, without requiring an admin to intercede for every account request. There are plenty of other ideas out there, too.

Maybe for now a simple 400 error for any sort of error would at least make the job a bit harder for a sniffer? Neither 401 nor 403 are really appropriate here, anyway, since no authorization is involved in the first place.

lib/pbench/cli/server/pbench_create_user.py
```diff
@@ -32,18 +39,21 @@ def execute(self):
...
 
 
 @click.command()
-@common_options
+@click.option(
+    "--config-file",
+    required=False,
+    default=os.environ.get("_PBENCH_SERVER_CONFIG"),
+    help="Specify the config file",
+)
```
dbutenhof, line 57: Hmm; I like having the `@common_options` here. One option is to simply clone and edit `lib/cli/agent/options.py` to `lib/cli/server` so that we can use it on the server side as well. A better but messier option would be to refactor this to lift it outside of the `agent`/`server` subtree and share it for both (though having the separate config classes makes this a bit tricker we could raise most of the code with reference to subclass helpers supplied by either side.)

lib/pbench/cli/server/pbench_create_user.py
```diff
@@ -83,8 +93,18 @@ def execute(self):
     required=False,
     help="If user already exists and this option is specified, the error will not be propagated",
 )
```
dbutenhof: I sat here staring at this input box for a while figuring out what I wanted to say. I'm at least half inclined to agree with Webb, drop this option, and simply have the `unittests` script ignore a 403 failure. It's not quite that simple.

- You're not currently returning any indication of the actual HTTP status; but maybe it's OK to simply ignore all errors
- More importantly, though, I'm not entirely sure any of this will even help with my problem. The gold tests capture `stdout` and `stderr` which includes the `Logger` output. So unless the error path has _no_ difference from the success path, the gold unit tests will fail anyway if the sequence in the current run differs from the prototype run.

I think I may need to give up on the idea of making this dynamic within the per-test setup, and fall back to having the `unittests` script create a pre-defined set of users regardless of which users are in the test `.db` state file. (Or maybe ignoring errors is fine after all as long as the order and set of dynamic usernames doesn't vary from run to run...)

lib/pbench/cli/agent/commands/pbench_create_user.py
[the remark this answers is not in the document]
webbnh: > idea behind these return messages like (username/email already in use) is to just give some help to the user to understand what went wrong

Nikhil, I'm right there with you:  I prefer errors to be specific and clear, because, at least initially, _I_ will be the user, and I'll appreciate knowing what's going on when something goes wrong.  But, in the case of reporting back to an unknown, un-credentialed, unauthorized request, we are supposed to be very circumspect about what we give away.  Generally speaking, it's supposed to be "one bit of resolution" -- success or failure, and not much more.  @dbutenhof's suggestion of using `400` is probably appropriate (I don't like it much, but I think it's correct).  If we want to support the user better, then they need to log in first so that we can account for their activities and our responses, and that gets us out of the area of self-registration.

> after couple of failed attempts the user might just leave the room and not come back

I think you're right to be sensitive to this.  However, to support that sort of user, we can use the GUI, and we can implement multiple stages to registration, as Dave outlined.  For the case in this PR, I'm not worried about the user walking away, because I expect that someone using a command line tool to create a registration has an intense interest in making it succeed, and they probably aren't intending to register only a single time (for that, the GUI is likely an easier approach for the user).

> I guess we can't just rely on RH SSO since other people who do not have redhat account might want to use the product.

I think that that depends on the deployment.  For instance, if we are offering our service Red Hat-wide, then RH SSO would probably be just the thing.  For instance, Quay is adopting this model, supporting both internal and external users via RH SSO, with no alternative authentication mechanism.  Obviously, if we want to support other deployments, we would presumably have to support other authentication mechanisms, as well.

> I am not sure but I do think this could be an overkill for such a simple utility tool, where we have to hit two rest endpoints in order to create a user (login and logout), if we login we create pbench token in our database and then logout to remove that token (Potentially 2 database queries).

I'm assuming that it would actually be as many as three endpoints:  attempt to create; when that fails, attempt to log in; if that succeeds, log out.  In human-scale terms, I don't think the user will be able to distinguish one request from three (think about how many requests the browser has to make to load the typical, modern, JS-based web page!).  And, it only makes multiple requests in the create-failure case (and it only makes the third request in the true-duplicate case).  But, this would allow the tool to securely determine whether the request is for a duplicate username.

As for whether it is overkill, yeah, it probably is.  In particular, it's not doing anything that the calling script couldn't do.  That is, the calling script could attempt the creation; if it fails, it could attempt to generate a token; if _that_ fails then there is a real error; otherwise, the failed creation is a non-error, and we're good to go.  And, all this works with just binary answers from the tools.

> Since this is a server side utility it is assumed that whoever wants to use the cli to create user should have server installed already and so we could just simply pass on the error message that rest endpoint reported.

The issue is not so much the behavior of this tool but rather the behaviors of the server required to support this tool.  That is, any REST API which this tool can use _could_ also be used by a piece of malicious software sitting elsewhere on the network.  So, if the server is giving effusive responses to allow this tool to support its user, then the attacker is also going to have that same level of help and support.

I would be inclined to pursue the idea that this tool is a server-side utility and therefore is co-located with the server.  That would enable a _private_ communication with the server which could be trusted without authentication.  For instance, we could provide a listener on a local port which doesn't accept connections from other hosts.  Or, we could give the REST endpoint different behaviors (e.g., different diagnostics, different capabilities, etc.) depending on whether the request comes from the local host or from a remote host.  If we know from context that we can trust the requestor, then lots of these issues go away.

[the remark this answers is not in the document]
webbnh: > I see the "gold" CLI-level unit tests going away, which would remove the need for this. Not right away, unfortunately.

@dbutenhof, understood.  However, once the "gold" unit tests are gone, won't we still have unit tests which need to have usernames in order to perform their function?  Or, would we do it all with mocks?  (Still, for _functional_ testing, we would want to be able to have usernames, and being able to create them automatically for the tests would be handy, so I don't really see the need for this tool going away.)

> in practice I'm not sure we want to require installing a second package to enable this, or go to the effort right now to create such a package.

Don't bother me with trivialities!  :-D

(You may well be right, though; still, we have to know where we want to go before we can decide how and when we want to go there.)

lib/pbench/cli/server/pbench_create_user.py
webbnh: I like the idea have having a uniform user interface across all of our tools, to the extent that we can achieve it.  The best way to accomplish that is by using common code.

However, for the `-C`/`--config-file` switch, I don't think the interface is actually common:  for agent-side tools, `-C` should specify an _agent_ config file (and, if the switch is missing, it should consult the `_PBENCH_AGENT_CONFIG` environment variable for the value), whereas for server-side tools, `-C` should specify a _server_ config file (and, if the switch is missing, it should probably consult some other environment variable).  (And, if it's a "client-side" tool, it should probably have a third behavior variant, which, possibly, devolves to one or both of the previous....)

So, I guess we just go with the clone approach.  :-p

lib/pbench/test/unit/server/test_pbench_create_user.py
```diff
@@ -46,12 +47,15 @@ def test_help(pytestconfig):
 
     @staticmethod
     @responses.activate
-    def test_args(valid_config, pytestconfig):
+    def test_args(pytestconfig):
+        cfg_file = pytestconfig.cache.get("_PBENCH_SERVER_CONFIG", None)
```
webbnh, line 51: On the agent side, this is handled via the `valid_config` Pytest fixture...you don't want to set up something analogous for the server side and avoid the duplicate code?

webbnh (changes requested):
I think you should rework the `--config-file` option.

lib/pbench/cli/server/pbench_create_user.py
npalaska: If I exit the command with http status code, would that solve the problem?

dbutenhof: > If I exit the command with http status code, would that solve the problem?

Not if we also fix it to return 400 for all errors ;-)

But, actually, I had another thought while writing this that I should have seen a while back... this utility isn't going to help me at all for the unit tests in any case, because we don't actually have a running server. We'd need a utility that constructed a `User` model object and added it to the database directly, like I'm doing for datasets in my `pbench-state-manager` and `pbench-import-datasets` utilities.

This REST-based utility makes sense for a production environment if we really want to be able to manage usersfrom outside of the server instance. I'm not convinced that's a strong user case, since we ought to have a user management console inside the dashboard UI. Regardless, I think we _at least_ (maybe _also_) need a local "setup" utility to create the canned `admin` user on install, and _that_ likely shouldn't rely on a fully configured running server ... so the direct DB access would be better, and is also what I need for the unit test environment.

So with this in mind, I've lost interest in the `--exists-ok` option entirely; I can't use it, as this utility is currently constructed, regardless of return values. So let's remove it. Or re-cast this entire utility to work directly on the DB rather than through REST, which would make it appropriate for server installation and CLI unit tests.

And while a CLI/REST interface may be useful in the long run, I'm actually not seeing a strong use case for it right now at all...

webbnh: > let's [...] re-cast this entire utility to work directly on the DB rather than through REST, which would make it appropriate for server installation and CLI unit tests.

That approach would get my vote for the reasons stated in the last paragraph.

npalaska: I think that make sense, I will make the changes to this cli utility to directly work with the database rather than using  the REST layer since its deployed on the same instance as a server.

lib/pbench/cli/server/pbench_create_user.py
```diff
@@ -32,18 +40,21 @@ def execute(self):
...
 
 
 @click.command()
-@common_options
+@click.option(
+    "--config-file",
+    required=False,
+    default=os.environ.get("_PBENCH_SERVER_CONFIG"),
+    help="Specify the config file",
+)
```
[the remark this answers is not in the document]
npalaska: fixed

lib/pbench/cli/server/pbench_create_user.py
```diff
@@ -83,8 +94,18 @@ def execute(self):
...
+    context, config_file, username, password, email, first_name, last_name, exists_ok
+):
+    if not config_file:
+        click.echo(
+            f"{_NAME_}: ERROR: No config file specified; set"
+            " _PBENCH_SERVER_CONFIG env variable",
+            err=True,
+        )
+        sys.exit(1)
+
```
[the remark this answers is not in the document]
npalaska: Yes, fixed. I cloned the `options.py` file in the server and edited it

lib/pbench/test/unit/server/test_pbench_create_user.py
npalaska: Fixed

lib/pbench/cli/server/pbench_create_user.py
```diff
@@ -1,19 +1,27 @@
...
-    def execute(self):
-        """Creates a user in the pbench user database"""
+    def execute(self, config_file):
+        try:
+            config = PbenchServerConfig(config_file)
+            host = str(config.get("pbench-server", "host"))
+            server = f"http://{host}{config.rest_uri}"
+        except BadConfig as e:
+            click.echo(f"{_NAME_}: {e}", err=True)
+            return 2
```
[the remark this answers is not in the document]
npalaska: Fixed

lib/pbench/cli/server/pbench_create_user.py
npalaska: Fixed it, by cloning the options.py file from agent implementation and making the appropriate edits

lib/pbench/cli/server/options.py
```diff
@@ -0,0 +1,31 @@
+import click
+
+from pbench.cli.server import CliContext
+
+
+def common_options(f):
+    f = _pbench_server_config(f)
+    return f
```
webbnh, line 8: This is kind of a triviality, but this code is poor style/practice:  there is no good reason for the input of the function to be overwritten with the output.  (We're not running on a 80386 CPU with such limited resources that we have to help the compiler reuse processor registers!)

However, more significantly, there is no clue here as to what the input to the function is supposed to be, nor what the output is supposed to be, nor what the overall purpose of this function is.  Since this is the external interface to this module, it would be appropriate to have a docstring.

And, since this is new code, it would be good to use Python's type hints for the input and return value (ditto for `_pbench_server_config()`, as well).

Most people aren't familiar with the details of how Python's function decorators work, so it might be worth being a little bit verbose when pointing out that this function receives a function as an argument and returns another function (which will call the input) as its output.

lib/pbench/cli/server/pbench_create_user.py
```diff
@@ -0,0 +1,94 @@
...
+            user = User(
+                username=self.context.username,
+                password=self.context.password,
+                first_name=self.context.first_name,
+                last_name=self.context.last_name,
+                email=self.context.email,
+            )
+            user.add()
+            click.echo(f"User {self.context.username} registered")
+            return 0
```
webbnh, line 39: Given that you report all errors via exceptions, you don't need to return any value on success (i.e., you can just omit this line here and let the caller generate the exit status value).

lib/pbench/cli/server/pbench_create_user.py
```diff
@@ -0,0 +1,94 @@
...
+    context.last_name = last_name
+
+    try:
+        rv = PbenchCreateUser(context).execute()
+    except BadConfig as exc:
+        click.echo(exc, err=True)
+        rv = 2
+    except Exception as exc:
+        click.echo(exc, err=True)
+        rv = 1
```
webbnh, line 92: Alternatively:
```
    except Exception as exc:
        click.echo(exc, err=True)
        rv = 2 if isinstance(exc, BadConfig) else 1
```
This provides a single, common path for all errors (e.g., a single, common `echo()` call), and, in the future, `BadConfig` can be replaced by a tuple of to-be-specially-handled exceptions when we decide that we want other cases to produce an exit status of `2` rathe than `1`.

webbnh (changes requested):
Looks great, now!

I do have one substantial set of requests, though:  for what is effectively new code, I think we should be using Python type hints, and `common_options()` should have a good docstring.  (And, of course, a couple of nits/suggestions.)

lib/pbench/cli/server/pbench_create_user.py
```diff
@@ -0,0 +1,94 @@
+import click
```
dbutenhof, line 1: Again, I think this file should be `pbench-user-create`, at least. We don't want to start using underscores, and we should be moving towards the verb-noun structure to make click subcommand structuring look cleaner.

However, I also note that all of Chuck's commands just have basic object names instead of full command names, which are provided by the `setup.cfg` file `console_scripts` variable as in `pbench-server = pbench.cli.server.shell:main`.

My only concern is that I'm not entirely sure how to employ this sort of command in the cli unit tests since it's going to require a setup build.

lib/pbench/cli/server/pbench_create_user.py
```diff
@@ -0,0 +1,94 @@
...
+
+class PbenchCreateUser:
+    def __init__(self, context):
+        self.context = context
+
+    def execute(self):
+        try:
+            config = PbenchServerConfig(self.context.config)
+        except BadConfig:
+            raise
```
dbutenhof, line 21: I guess I won't push to reproduce the entire agent command support hierarchy here, but I think the agent-side `BaseCommand` (in `lib/pbench/agent/base.by`, which centralizes this sort of setup in one place) is a good idea, and we should be duplicating that on the server side and inheriting from that for our utility commands instead of duplicating common code like this.

And, yeah, once we get this working I should retrofit my utility commands to fit. But I'll leave it to you whether to take that step now or wait for someone to refactor it later... ;-)

lib/pbench/cli/server/pbench_create_user.py
webbnh: As you observed, @dbutenhof, the _implementation_ of the command is separated from the details of how it is invoked (which, to a large extent, are specified in the `setup`).  This means among other things, that the module name can include underscores where the command verb uses hyphens (and, I think that this is actually conventional).

Nevertheless, I concur that we should make an effort to move toward noun-verb structure, although, again, that doesn't _necessarily_ have to be reflected in the implementation (but, it would probably be better if it were).

As for unit tests, if a setup build is inconvenient, you can probably invoke the command's implementation directly, or you can run it the same way that Nikhil's unit tests for the command itself invoke it, via a [Click "runner"](https://click.palletsprojects.com/en/7.x/api/#testing).

npalaska: Another thing, having hyphens instead of underscore is not a valid python identifier, I can not seem to import it into another modules like tests. It is also mentioned in the PEP8 guide that we should be using underscore if at all necessary https://www.python.org/dev/peps/pep-0008/#package-and-module-names

dbutenhof: Yes, if you're importing from the module, you're probably right. But we don't use the `pbench` prefix on any other internal modules, and it's redundant anyway given the `pbench.cli.server` path. So how about just `user_create.py` and map the CLI `pbench-user-create` through `setup.cfg` as the others are done.

lib/pbench/cli/server/options.py
```diff
@@ -1,14 +1,22 @@
...
 from pbench.cli.server import CliContext
 
 
-def common_options(f):
-    f = _pbench_server_config(f)
-    return f
+def common_options(in_f: Callable) -> Callable:
+    """
+    This function can be used as a decorator for multiple server side CLI utilities,
+    where it recieves a click command decorated function as an argument and sets some
```
webbnh, line 10: s/recieves/receives/

lib/pbench/cli/server/pbench_user_create.py
```diff
@@ -86,9 +85,6 @@ def main(context, username, password, email, first_name, last_name):
         rv = PbenchCreateUser(context).execute()
     except BadConfig as exc:
         click.echo(exc, err=True)
-        rv = 2
-    except Exception as exc:
-        click.echo(exc, err=True)
-        rv = 1
+        rv = 2 if isinstance(exc, BadConfig) else 1
```
webbnh: The `except` clause (at new line 86) needs to catch `Exception` (or something more broad than `BadConfig`).

webbnh (changes requested):
So close!  Just one small issue (and one nit).

webbnh (dismissed):
Good to go.

I don't want to gate this commit, but...if you do make any further changes, more use of type hints would be appropriate (e.g., in the definitions of `main()`, `CreateUser.__init__()`, and the `callback()` in `_pbench_server_config()`).

dbutenhof:
Well this is weird; something you did, or something you took in on the rebase, broke the agent unit tests:

```
lib/pbench/test/functional/test_pbench_config.py FFF
lib/pbench/test/functional/agent/cli/commands/test_cleanup.py FF
lib/pbench/test/functional/agent/cli/commands/results/test_clear_results.py FF
lib/pbench/test/functional/agent/cli/commands/tools/test_clear_tools.py FFFFFFFF
lib/pbench/test/functional/agent/cli/commands/tools/test_tools_list.py FF
lib/pbench/test/functional/agent/cli/commands/triggers/test_list_triggers.py FF
lib/pbench/test/functional/agent/cli/commands/triggers/test_register_tool_trigger.py FFFF
```

npalaska:
> Well this is weird; something you did, or something you took in on the rebase, broke the agent unit tests:
> 
> ```
> lib/pbench/test/functional/test_pbench_config.py FFF
> lib/pbench/test/functional/agent/cli/commands/test_cleanup.py FF
> lib/pbench/test/functional/agent/cli/commands/results/test_clear_results.py FF
> lib/pbench/test/functional/agent/cli/commands/tools/test_clear_tools.py FFFFFFFF
> lib/pbench/test/functional/agent/cli/commands/tools/test_tools_list.py FF
> lib/pbench/test/functional/agent/cli/commands/triggers/test_list_triggers.py FF
> lib/pbench/test/functional/agent/cli/commands/triggers/test_register_tool_trigger.py FFFF
> ```

Yeah started after the rebase, looking through this PR commit to see if any changes might have caused it

npalaska:
Noticed that adding `pbench-user-create = pbench.cli.server.user_create::main` in `setup.cfg` breaks unit tests agent CLI

webbnh (approved):

riya-17 (dismissed):

lib/pbench/cli/server/user_create.py
```diff
@@ -0,0 +1,90 @@
...
+from pbench.cli.server import pass_cli_context
+from pbench.cli.server.options import common_options
+from pbench.common.logger import get_pbench_logger
+from pbench.server.database.models.users import User
+from pbench.server.database.database import Database
+
+_NAME_ = "pbench-user-create"
+
+
+class CreateUser:
```
dbutenhof, line 13: Ha; it's a little jarring that module `user_create.py` defines a class called `CreateUser`, but that's a nit at this point.

lib/pbench/cli/server/user_create.py
```diff
@@ -0,0 +1,90 @@
...
+                username=self.context.username,
+                password=self.context.password,
+                first_name=self.context.first_name,
+                last_name=self.context.last_name,
+                email=self.context.email,
+            )
+            user.add()
+            click.echo(f"User {self.context.username} registered")
+        except Exception:
+            raise
```
dbutenhof, line 40: This seems pointless; the whole try should be removed if we're not even logging the exception and simply re-raising it.

dbutenhof (approved):
You need a rebase, at least, but it's looking good. ;-)

setup.cfg
```diff
@@ -32,6 +32,7 @@ console_scripts =
    pbench-list-triggers = pbench.cli.agent.commands.triggers.list:main
    pbench-register-tool-trigger = pbench.cli.agent.commands.triggers.register:main
    pbench-server = pbench.cli.server.shell:main
+   pbench-user-create = pbench.cli.server.user_create:main
```
portante, line 35: We also have to modify the `server/Makefile` to copy the new `pbench-user-create` command.

portante (changes requested):
Looks good.

Can you remove that unnecessary `try/except`, rename `CreateUser` to `UserCreate`, and add the command the `server/Makefile`?

lib/pbench/cli/server/user_create.py
webbnh: Ditto at lines 20-21.

setup.cfg
webbnh: We should probably follow the lead of the `agent/Makefile` for this -- it uses a list of tools to copy, so adding new tools just means adding an entry to the list.  (Currently, I think there is only one tool (`pbench-config`) so there hasn't been a need for a "list", but now that there are two....)

dbutenhof: Thanks, @webbnh; I had thought to make a similar comment after comparing the two Makefiles, but forgot to come back.

webbnh (approved):
~Looks good.~ 👎

server/Makefile
```diff
@@ -41,7 +44,7 @@ install-python3-setup: install-bin install-lib
 	mkdir -p ${DESTDIR}/python3
 	${COPY} requirements.txt ${DESTDIR}
 	(cd ..; SKIP_GENERATE_AUTHORS=1 SKIP_WRITE_GIT_CHANGELOG=1 python3 setup.py install --prefix=${DESTDIR}/python3)
-	${COPY} ${DESTDIR}/python3/bin/pbench-config ${DESTDIR}/python3/bin/pbench-server ${BINDIR}/
+	${COPY} ${DESTDIR}/python3/bin/pbench-config ${DESTDIR}/python3/bin/pbench-server ${click-scripts} ${BINDIR}/
```
webbnh, line 47: I don't think that this will actually work:  I don't think `pbench-user-create` is in the current working directory when this copy command is executed.

In the `agent/Makefile`, we prefix the contents of the list with the appropriate directory path:
```
$(addprefix ${DESTDIR}/python3/bin/, ${click-scripts})
```
I expect that you'll need to do that here.

webbnh (changes requested):
Um...I have to take back my Approval -- see comment.

setup.cfg
webbnh: > Currently, I think there is only one tool (`pbench-config`) so there hasn't been a need for a "list", but now that there are two....

My bad:  `pbench-config` has not yet been Clickinated.  But, under the assumption that there will indeed be future Click-based scripts as part of the server installation, let's keep the list.

webbnh (dismissed):
Cool!

server/Makefile
```diff
@@ -44,7 +44,7 @@ install-python3-setup: install-bin install-lib
 	mkdir -p ${DESTDIR}/python3
 	${COPY} requirements.txt ${DESTDIR}
 	(cd ..; SKIP_GENERATE_AUTHORS=1 SKIP_WRITE_GIT_CHANGELOG=1 python3 setup.py install --prefix=${DESTDIR}/python3)
-	${COPY} ${DESTDIR}/python3/bin/pbench-config ${DESTDIR}/python3/bin/pbench-server ${click-scripts} ${BINDIR}/
+	${COPY} $(addprefix ${DESTDIR}/python3/bin/, pbench-config pbench-server ${click-scripts}) ${BINDIR}/
```
dbutenhof, line 47: Technically, `pbench-server` _is_ click, so it might make more sense to put it in the list, but I think we can go with this for now.

dbutenhof (approved):

portante (approved):

dbutenhof (dismissed):

portante (changes requested):
Apparently still needs a rebase...

portante (approved):

portante (changes requested):
And another rebase ...

webbnh (dismissed):

portante (approved):

dbutenhof (approved):

PR #2206, opened 2021-04-08
Refactor Elasticsearch template management

dbutenhof:
This is "step 1" ... I've moved the template management class out of indexer.py into a separate module. There is no change
in behavior, although the movement changes the Logger path (from "<cmd>.indexer" to "<cmd>.templates") which results in many gold file changes.

Step 2 will be to implement a Postgresql table to track the template metadata and integrate with this class; then we can share one source of information between the indexer and the server APIs. I'll do step 2 as a separate PR to keep each simple (or at least "simpler").


dbutenhof:
Ugh; rebased with userdb at the last moment, and thought I was OK because gold files didn't show conflicts; but I broke it by missing a change in template handling.

portante (dismissed):

portante (dismissed):

dbutenhof:
@portante you're being very proactive here after I declared an error and shifted it back to draft! (Is there a way to cancel review requests? I've never seen a way.)

I pushed a fix, and looking at it realized that there was a bug in the userdb code that I propagated. Something of an inefficiency given the structure of the code, but wrong all the same; I'll push another amend after the tests finish, and we'll see if anyone notices what I fixed. ;-)

npalaska (approved):

riya-17 (approved):

Issue #2201, opened 2021-04-06
The pbench server should not allow controller directories with leading dashes ('--' or '-')


portante:
Today the pbench server allows leading dashes for controller names.  We had a controller name created with `--run-label_` which is obviously incorrect.

portante:
See also #2488.

PR #2207, opened 2021-04-08
Add hostname and IP address validation for env

portante:
From issue #2201, it became apparent that we weren't validating hostnames, which led to weird, but valid, directory names for
controllers being created on the server.  One way this occurs is when the `_pbench_hostname` and or `_pbench_full_hostname` environment variables are defined with invalid values.

We add the new commands `validate-hostname` and `validate-ipaddress` to ensure the definition of those names are correct.

In the process of doing that, we found that we were using invalid host names in two of our unit tests.

----

Derived from PR #2181.

### DIFF
--- a/agent/util-scripts/pbench-register-tool
+++ b/agent/util-scripts/pbench-register-tool
@@ -278,6 +278,19 @@ else
 	fi
 fi
 
+invalids=0
+for (( i=0; ${i} < ${#remotes_A[@]}; i++ )); do
+	remote="${remotes_A[${i}]}"
+	validate-hostname "${remote}"
+	if [[ ${?} -ne 0 ]]; then
+		error_log "Invalid remote specified: '${remote}'"
+		(( invalids++ ))
+	fi
+done
+if [[ ${invalids} -gt 0 ]]; then
+	exit 1
+fi
+
 if [[ ${_do_test_labels} -ne 0 ]]; then
 	# Used by pbench-register-tool-set to avoid duplicating logic, we
 	# have been asked to exit early successfully if all argument
--- a/lib/pbench/agent/utils.py
+++ b/lib/pbench/agent/utils.py
@@ -1,5 +1,7 @@
+import ipaddress
 import logging
 import os
+import re
 import subprocess
 import sys
 
@@ -188,3 +190,43 @@ def collect_local_info(pbench_bin):
         hostdata[arg] = cp.stdout.strip() if cp.stdout is not None else ""
 
     return (version, seqno, sha1, hostdata)
+
+
+# Derived from https://stackoverflow.com/questions/106179/regular-expression-to-match-dns-hostname-or-ip-address
+# with a few modifications: take advantage of ignoring case; use non-capturing
+# groups to improve efficiency; take advantage of RFC 1123 modification to RFC
+# 952 relaxing first character to be letter or digit, with the repetition moved
+# to the last group to alleviate backtracking.  Or in other words, a case-blind
+# comparison seeking a letter-or-digit, optionally followed by a sequence of 0
+# to 61 letter-or-digit-or-hyphens followed by a letter-or-digit, follwed by 0
+# or more instances of that same pattern preceded by a period.
+_allowed = re.compile(
+    r"[A-Z0-9](?:[A-Z0-9\-]{0,61}[A-Z0-9])?(?:\.[A-Z0-9](?:[A-Z0-9\-]{0,61}[A-Z0-9])?)*",
+    flags=re.IGNORECASE,
+)
+
+
+def validate_hostname(host_name: str):
+    """validate_hostname - validate the given hostname uses the proper syntax.
+
+    A host name that follows RFC 952 (amended by RFC 1123) is accepted only.
+    Host names are not resolved to IP addresses, and IP addresses are also
+    accepted.
+
+    Algorithm taken from: https://stackoverflow.com/questions/2532053/validate-a-hostname-string
+
+    Returns 0 on success, 1 on failure.
+    """
+    if not host_name or len(host_name) > 255:
+        return 1
+
+    if _allowed.fullmatch(host_name):
+        return 0
+
+    # It is not a valid host name, but could be a valid IP address.
+    try:
+        ipaddress.ip_address(host_name)
+    except ValueError:
+        return 1
+
+    return 0
--- a/lib/pbench/test/unit/agent/test_validate_hostname.py
+++ b/lib/pbench/test/unit/agent/test_validate_hostname.py
@@ -1,0 +1,15 @@
+"""Tests for validate_hostname
+"""
+
+from pbench.agent.utils import validate_hostname
+
+
+def test_validate_hostname():
+    assert validate_hostname("test") == 0
+    assert validate_hostname("test.example.com") == 0
+    assert validate_hostname("tes_t.example.com") == 1
+    assert validate_hostname("test-.example.com") == 1
+    assert validate_hostname("--run-label__") == 1
+    assert validate_hostname("127.0.0.1") == 0
+    assert validate_hostname("1270.0.0.1") == 0
+    assert validate_hostname("2001:0db8:85a3:0000:0000:8a2e:0370:7334") == 0
--- a/agent/base
+++ b/agent/base
@@ -100,14 +100,29 @@ fi
 # don't use colons and dashes in the date suffix
 export date_suffix=$(date --date ${date} --utc +"%Y.%m.%dT%H.%M.%S")
 
+function validate_hostname {
+    validate-hostname "${1}"
+    if [[ ${?} -ne 0 ]]; then
+        error_log "Invalid ${2}: '${1}'"
+        exit 1
+    fi
+}
+
 if [[ -z "${_pbench_hostname}" ]]; then
     export _pbench_hostname=$(hostname -s)
 fi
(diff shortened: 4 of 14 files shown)


lib/pbench/agent/utils.py
```diff
@@ -188,3 +190,40 @@ def collect_local_info(pbench_bin):
...
+)
+
+
+def validate_hostname(host_name):
+    """validate_hostname - validate the given hostname uses the proper syntax.
+
+    A host name that follows RFC 952 (amended by RFC 1123) is accepted only.  Host
+    names are not resolved to IP addresses, and IP addresses are also accepted.
+
+    Algorithm taken from: https://stackoverflow.com/questions/2532053/validate-a-hostname-string
```
ndokos, line 207: This says there is a 63-octet limit per component: should you be checking for that?

ndokos (changes requested):

lib/pbench/agent/utils.py
```diff
@@ -188,3 +190,40 @@ def collect_local_info(pbench_bin):
...
+    """
+    if len(host_name) > 255:
+        return 1
+
+    if host_name[-1] == ".":
+        # Strip trailing '.' to allow comprehension below to work in all cases
+        _host_name = host_name[:-1]
+    else:
+        _host_name = host_name
+    if all(_allowed.match(x) for x in _host_name.split(".")):
```
ndokos, line 219: Wait - `_allowed` is a regexp for the whole thing, not for each component, right? This checks that each `.`-separated component matches, right? Even if it works, it's inefficient, no?

ndokos (changes requested):

lib/pbench/agent/utils.py
portante: Fixed.

lib/pbench/agent/utils.py
portante: We can be a little loose with the hostname checking, this solves the most egregious issues.  We could validate 63 characters between dots, but it won't cause us any headaches.

ndokos (dismissed):

agent/util-scripts/pbench-register-tool
```diff
@@ -278,6 +278,19 @@ else
 	fi
 fi
 
+invalids=0
+for (( i=0; ${i} < ${#remotes_A[@]}; i++ )); do
+	remote="${remotes_A[${i}]}"
+	validate-hostname "${remote}"
```
portante, line 284: Note that we don't use the `validate_remote` function from `base` because we want to report on all the remotes that are bad before exiting.

portante (commented):
This should be good to go.  I have addressed all the review comments from PR #2181 that were for this code.

lib/pbench/agent/utils.py
```diff
@@ -188,3 +190,33 @@ def collect_local_info(pbench_bin):
...
     return (version, seqno, sha1, hostdata)
+
+
+# From https://stackoverflow.com/questions/106179/regular-expression-to-match-dns-hostname-or-ip-address
+_allowed = re.compile(
+    r"(([a-zA-Z]|[a-zA-Z][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z]|[A-Za-z][A-Za-z0-9\-]*[A-Za-z0-9])"
+)
+
+
+def validate_hostname(host_name: str):
```
portante, line 201: The one change made here with this update is to add type hint.

portante (commented):
Okay, now it is ready for re-review.

lib/pbench/agent/utils.py
```diff
@@ -188,3 +190,33 @@ def collect_local_info(pbench_bin):
         hostdata[arg] = cp.stdout.strip() if cp.stdout is not None else ""
 
     return (version, seqno, sha1, hostdata)
+
+
+# From https://stackoverflow.com/questions/106179/regular-expression-to-match-dns-hostname-or-ip-address
+_allowed = re.compile(
+    r"(([a-zA-Z]|[a-zA-Z][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z]|[A-Za-z][A-Za-z0-9\-]*[A-Za-z0-9])"
+)
```
webbnh, line 198: I have several suggestions, here:

- Add `flags=re.IGNORECASE` to the `re.compile()` call so that you don't need to include both letter ranges, which makes the expression considerably shorter.
- Use non-capturing groups, which improves efficiency, since we don't actually use the matches.
- RFC 1123 states, "the restriction on the first character is relaxed to allow either a letter or a digit," which has the side effect of allowing your algorithm to match either dotted-decimal quartets or FQDNs.
- So, I think you can render this as:
```
_allowed = re.compile(
    r"[A-Z0-9](?:[A-Z0-9\-]{0,61}[A-Z0-9])?(?:\.[A-Z0-9](?:[A-Z0-9\-]{0,61}[A-Z0-9])?)*",
    flags=re.IGNORECASE
)
```
That is, a case-blind comparison seeking a letter-or-digit, optionally followed by a sequence of zero to 61 letter-or-digit-or-hyphens followed by a letter-or-digit, followed by zero or instances of that same pattern preceded by a dot.

Note that I moved the repetition from the first group to the last: this alleviates backtracking when the RegEx engine doesn't find a match.

webbnh (changes requested):
It looks like you got most of the items that I raised in #2181; I've transferred the remaining two.

lib/pbench/agent/utils.py
portante: Slow on the uptake.  I'll modify this.

lib/pbench/test/unit/agent/test_validate_hostname.py
```diff
@@ -0,0 +1,13 @@
...
+from pbench.agent.utils import validate_hostname
+
+
+def test_validate_hostname():
+    assert validate_hostname("test") == 0
+    assert validate_hostname("test.example.com") == 0
+    assert validate_hostname("tes_t.example.com") == 1
+    assert validate_hostname("--run-label__") == 1
+    assert validate_hostname("127.0.0.1") == 0
+    assert validate_hostname("1270.0.0.1") == 1
```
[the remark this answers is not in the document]
portante: I've added a test case.

lib/pbench/test/unit/agent/test_validate_hostname.py
```diff
@@ -0,0 +1,15 @@
+"""Tests for validate_hostname
+"""
+
+from pbench.agent.utils import validate_hostname
+
+
+def test_validate_hostname():
+    assert validate_hostname("test") == 0
+    assert validate_hostname("test.example.com") == 0
+    assert validate_hostname("tes_t.example.com") == 1
```
dbutenhof, line 10: Funny; I had never actually realized (or at least didn't remember having known ;-) ) that "_" wasn't a valid hostname character... but I confirmed with the RFC. So, huh.

One comment on the tests, however: you should use some capital letters. Hostnames are case-insensitive, but we should be able to accept uppercase and it wouldn't be a bad thing to test that explicitly. Capital hex letters in the IPv6 wouldn't be bad, too.

dbutenhof (approved):

lib/pbench/test/unit/agent/test_validate_hostname.py
ndokos: Yes, I didn't remember it either, but they did that right :-) I hate typing underscores... But they probably didn't do it to satisfy me: does the RFC contain a rationale?

ndokos (approved):

lib/pbench/test/unit/agent/test_validate_hostname.py
webbnh: I could not find a rationale.  I think they started with what they already had for names in ARPAnet (which was a flat, global namespace), and made it hierarchical by adding dot-separated domains with similar naming conventions.

lib/pbench/test/unit/agent/test_validate_hostname.py
```diff
@@ -0,0 +1,15 @@
...
+    assert validate_hostname("test-.example.com") == 1
+    assert validate_hostname("--run-label__") == 1
+    assert validate_hostname("127.0.0.1") == 0
+    assert validate_hostname("1270.0.0.1") == 0
+    assert validate_hostname("2001:0db8:85a3:0000:0000:8a2e:0370:7334") == 0
```
webbnh, line 15: I think there might be a small crack here that we need to fill in, either in this PR or in #2181 (or, if necessary, in a follow-on....).

I believe that the conventional way to specify an IPv6 address with a port is to put the node portion in square brackets:  e.g., `[::1]:8080`.

So, when someone tries `pbench-tool-meister-start --redis-server [::1]:17001`, the code is going to grab the brackets as part of the hostname.  However, `ipaddress.ip_address()` doesn't accept them.

So, we'll either need `validate_hostname()` to accept (and remove) them, or we'll need need its various callers (`BaseServer.__init__()` and others) to recognize and strip them.

webbnh (approved):
Looks good, but before you merge we need to decide how to dispose of the issue below (see my comment).

lib/pbench/test/unit/agent/test_validate_hostname.py
portante: We don't any support for port numbers being specified in the current interfaces as of this PR, so it'll have to be addressed in PR #2181.

PR #2208, opened 2021-04-09
Jaeger persistent tool

nikiglen:
This creates a JaegerTool persistent tool subclass, mapped as "jaeger-agent"


lib/pbench/agent/tool_meister.py
```diff
@@ -586,6 +619,7 @@ def fetch_params(params):
         "dcgm": DcgmTool,
         "node-exporter": NodeExporterTool,
         "pcp": PcpTool,
+        "jaeger-agent": JaegerTool,
```
portante, line 622: This should be `jaeger` to match what PR #2209 is using.

lib/pbench/agent/tool_meister.py
```diff
@@ -467,6 +467,39 @@ def install(self):
...
+                else:
+                    self.collector_hosts_ports = opt[12:]
+        if self.collector_hosts_ports is None:
+            self.logger.error("No host:port provided for jaeger collector(s)")
+
+        executable = find_executable("jaeger-agent")
+        if executable:
+            self.args = [
+                executable,
+                ("--reporter.grpc.host-port=%s" % self.collector_hosts_ports),
```
portante, line 491: Consider dropping `--collector` option to simplify, and use the "tool data sink" host name (which we have to find the right field name) and a default port, using Python3 f-strings: `f"--reporter.grpc.host-port={self.tds_host_name}:{self.def_jaeger_port}"`

portante:
@nikiglen, can you take a look at the unit tests failures?  If you run them locally, you should be able to reproduce.

PR #2213, opened 2021-04-13
Add role column in user table

npalaska:
- Adds an optional role column to the user table in order to distinguish certain users as admins. 
- role column takes a string value indicating the type of user, for example it can be set as `administrator` to indicate the admin user.
- role can only be set via CLI and option is not available via REST API

### DIFF
--- a/lib/pbench/cli/server/user_create.py
+++ b/lib/pbench/cli/server/user_create.py
@@ -1,11 +1,12 @@
 import click
+
 from pbench import BadConfig
-from pbench.server import PbenchServerConfig
 from pbench.cli.server import pass_cli_context
 from pbench.cli.server.options import common_options
 from pbench.common.logger import get_pbench_logger
-from pbench.server.database.models.users import User
+from pbench.server import PbenchServerConfig
 from pbench.server.database.database import Database
+from pbench.server.database.models.users import Roles, User
 
 _NAME_ = "pbench-user-create"
 
@@ -29,9 +30,14 @@ class UserCreate:
             first_name=self.context.first_name,
             last_name=self.context.last_name,
             email=self.context.email,
+            role=self.context.role if self.context.role else "",
         )
+
         user.add()
-        click.echo(f"User {self.context.username} registered")
+        if user.is_admin():
+            click.echo(f"Admin user {self.context.username} registered")
+        else:
+            click.echo(f"User {self.context.username} registered")
 
 
 @click.command()
@@ -67,13 +73,28 @@ class UserCreate:
     required=True,
     help="pbench server account last name (will prompt if unspecified)",
 )
+@click.option(
+    "--role",
+    type=click.Choice([role.name for role in Roles], case_sensitive=False),
+    required=False,
+    help="Optional role of the user such as Admin",
+)
 @pass_cli_context
-def main(context, username, password, email, first_name, last_name):
+def main(
+    context: object,
+    username: str,
+    password: str,
+    email: str,
+    first_name: str,
+    last_name: str,
+    role: str,
+) -> None:
     context.username = username
     context.password = password
     context.email = email
     context.first_name = first_name
     context.last_name = last_name
+    context.role = role
 
     try:
         rv = UserCreate(context).execute()
--- a/lib/pbench/server/database/models/users.py
+++ b/lib/pbench/server/database/models/users.py
@@ -1,9 +1,16 @@
 import datetime
-from flask_bcrypt import generate_password_hash
+import enum
+
 from email_validator import validate_email
-from sqlalchemy import Column, Integer, String, DateTime, LargeBinary
-from pbench.server.database.database import Database
+from flask_bcrypt import generate_password_hash
+from sqlalchemy import Column, DateTime, Enum, Integer, LargeBinary, String
 from sqlalchemy.orm import relationship, validates
+
+from pbench.server.database.database import Database
+
+
+class Roles(enum.Enum):
+    ADMIN = 1
 
 
 class User(Database.Base):
@@ -18,6 +25,7 @@ class User(Database.Base):
     password = Column(LargeBinary(128), nullable=False)
     registered_on = Column(DateTime, nullable=False, default=datetime.datetime.now())
     email = Column(String(255), unique=True, nullable=False)
+    role = Column(Enum(Roles), unique=False, nullable=True)
     auth_tokens = relationship("ActiveTokens", backref="users")
 
     def __str__(self):
@@ -66,6 +74,13 @@ class User(Database.Base):
             Database.db_session.rollback()
             raise
 
+    @validates("role")
+    def evaluate_role(self, key, value):
+        try:
+            return Roles[value.upper()]
+        except KeyError:
+            return None
+
     @validates("password")
     def evaluate_password(self, key, value):
         return generate_password_hash(value)
@@ -112,15 +127,11 @@ class User(Database.Base):
             raise
 
     def is_admin(self):
-        # TODO: Add notion of Admin user
-        """this method would always return false for now until we add a notion of Admin user/group.
-        Once we know the admin credentials this method can check against those credentials to determine
-        whether the user is privileged to do more.
-
+        """This method checks whether the given user has an admin role.
         This can be extended to groups as well for example a user belonging to certain group has only those
         privileges that are assigned to the group.
         """
-        return False
+        return self.role is Roles.ADMIN
 
     @staticmethod
     def is_admin_username(username):
--- a/lib/pbench/test/unit/server/test_user_create.py
+++ b/lib/pbench/test/unit/server/test_user_create.py
@@ -11,6 +11,7 @@ class TestCreateUser:
     EMAIL_SWITCH = "--email"
     FIRST_NAME_SWITCH = "--first-name"
     LAST_NAME_SWITCH = "--last-name"
+    ROLE_SWITCH = "--role"
     USER_TEXT = "test_user"
     PSWD_TEXT = "password"
     EMAIL_TEXT = "test@domain.com"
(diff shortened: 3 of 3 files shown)


lib/pbench/server/database/models/users.py
```diff
@@ -18,6 +18,7 @@ class User(Database.Base):
     password = Column(LargeBinary(128), nullable=False)
     registered_on = Column(DateTime, nullable=False, default=datetime.datetime.now())
     email = Column(String(255), unique=True, nullable=False)
+    entitlement = Column(String(255), unique=False, nullable=True)
```
portante, line 21: These feels a lot like a "role".  Is there a reason we can't just call it "role" where we only support one role, "administrator"?

lib/pbench/server/database/models/users.py
```diff
@@ -112,14 +113,15 @@ def delete(username):
...
     def is_admin(self):
-        # TODO: Add notion of Admin user
         """this method would always return false for now until we add a notion of Admin user/group.
         Once we know the admin credentials this method can check against those credentials to determine
         whether the user is privileged to do more.
 
         This can be extended to groups as well for example a user belonging to certain group has only those
         privileges that are assigned to the group.
         """
+        if self.entitlement in ["admin", "administrator"]:
```
portante, line 123: I don't think we need to support two names for the same role.  The CLI and/or UI can map between convenience names and the one internal role name.

portante (changes requested):

lib/pbench/server/database/models/users.py
npalaska: Are you suggesting changing the column to a boolean and restricting the value to either true/false for administrator? that's one possibility, but having a string column we can quickly adopt to supporting multiple roles if we want to.

webbnh: I think he's suggesting changing the name from "entitlement" to "role"....  :-)

lib/pbench/server/database/models/users.py
dbutenhof: Funny, because I said the same thing ... and now I realize that I somehow didn't publish my comments. Sigh.

dbutenhof (changes requested):
I think we _will_ want to make this accessible through the `PUT /users/<id>` API, although it should be changeable only if the authorizing user itself has `"admin"` entitlement.

I suspect you're leaving the CLI open because this can be a tool to create the initial default admin user... but maybe we should allow `--entitlement=admin` on create only if the _user_ is `"admin"`?? Then again, this can be useful for testing and we generally wouldn't allow user login on the server, so maybe it's fine.

webbnh (changes requested):
Looks good.  But, I have a few requests:

- change all references to "entitlement" to be "role"
- implement a single, common definition of the role "name(s)", perhaps as part of an enum, so that it's a closed set or fixed association.
- use the defined role names to validate the input and/or user attribute

webbnh:
> I suspect you're leaving the CLI open because this can be a tool to create the initial default admin user... but maybe we should allow --entitlement=admin on create only if the user is "admin"?? Then again, this can be useful for testing and we generally wouldn't allow user login on the server, so maybe it's fine.

@dbutenhof, I don't think the CLI access comes with any notion of a current user.  We _could_ expand it to require a token, but that would put us squarely back into the chicken-and-egg problem when creating the initial admin user.  Rather, I think we're counting on the fact that this tool can only be used in an environment where it has direct access to the database and therefore can be and is necessarily considered to be privileged without further authentication or authorization.

As for the REST API, I have no problem with expanding the methods for user creation and modification to allow them to modify the target's role...provided that the token indicates that the requestor has the required role, but that feels to me like a different PR.

dbutenhof:
> @dbutenhof, I don't think the CLI access comes with any notion of a current user. We _could_ expand it to require a token, but that would put us squarely back into the chicken-and-egg problem when creating the initial admin user. Rather, I think we're counting on the fact that this tool can only be used in an environment where it has direct access to the database and therefore can be and is necessarily considered to be privileged without further authentication or authorization.

No, what I meant was that we don't allow setting the "admin" role when creating any user that's not named `"admin"`, for now. We will probably want to ease that restriction later, but for simplicity I'm tempted to say we should start out that way.

> As for the REST API, I have no problem with expanding the methods for user creation and modification to allow them to modify the target's role...provided that the token indicates that the requestor has the required role, but that feels to me like a different PR.

Indeed, and that's fine; I'm just saying we're going to need it.

lib/pbench/server/database/models/users.py
```diff
@@ -66,6 +72,12 @@ def add(self):
             Database.db_session.rollback()
             raise
 
+    @validates("role")
+    def evaluate_role(self, key, value):
+        if value.lower() in ["admin", "administrator"]:
+            return Roles(1)
```
dbutenhof, line 78: It'd be better to use `Roles[value.upper()]` to convert the string directly; and, again, I really don't like having two separate keywords here to translate to one role. Let's just stick with "admin" which maps to Roles.ADMIN. (Enums should generally be all-caps.)

lib/pbench/server/database/models/users.py
```diff
@@ -112,14 +124,12 @@ def delete(username):
...
-        # TODO: Add notion of Admin user
-        """this method would always return false for now until we add a notion of Admin user/group.
-        Once we know the admin credentials this method can check against those credentials to determine
-        whether the user is privileged to do more.
-
+        """This method checks whether the given user has an admin role.
         This can be extended to groups as well for example a user belonging to certain group has only those
         privileges that are assigned to the group.
         """
+        if self.role == Roles(1):
```
dbutenhof, line 131: `Roles.ADMIN` is a lot easier to read, without having to find the ENUM to figure out which role is `1`. Don't use ENUMs as integers!

lib/pbench/server/database/models/users.py
```diff
@@ -1,11 +1,16 @@
...
 from flask_bcrypt import generate_password_hash
 from email_validator import validate_email
-from sqlalchemy import Column, Integer, String, DateTime, LargeBinary
+from sqlalchemy import Column, Integer, String, DateTime, LargeBinary, Enum
 from pbench.server.database.database import Database
 from sqlalchemy.orm import relationship, validates
 
 
+class Roles(enum.Enum):
+    Admin = 1
```
dbutenhof, line 11: `ADMIN`

dbutenhof (changes requested):

lib/pbench/cli/server/user_create.py
```diff
@@ -67,13 +70,20 @@ def execute(self):
     required=True,
     help="pbench server account last name (will prompt if unspecified)",
 )
+@click.option(
+    "--entitlement",
+    required=False,
+    default=None,
+    help="Optional entitlement of the user such as admin, if not specified defaults to None",
```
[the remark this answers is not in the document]
dbutenhof: But this still includes "defaults to None" in the help string, so what does it look like with the addition of `show_default`?

Also, I still would prefer this to be a keyword rather than just a string; e.g.,

```
   type=click.Choice(['ADMIN'], case_sensitive=False))
```

For bonus points, create the choice list directly from the ENUM rather than hardcoding it...

lib/pbench/server/database/models/users.py
```diff
@@ -112,14 +125,12 @@ def delete(username):
...
-        Once we know the admin credentials this method can check against those credentials to determine
-        whether the user is privileged to do more.
-
+        """This method checks whether the given user has an admin role.
         This can be extended to groups as well for example a user belonging to certain group has only those
         privileges that are assigned to the group.
         """
+        if self.role == Roles.ADMIN:
+            return True
         return False
```
dbutenhof, line 134: Or how about simply `return self.role == Roles.ADMIN`??

Also, you can add attributes to an ENUM, making admin access an attribute of a role rather than identical to the name of the role, which would allow `return self.role.isAdmin()`... ;-) (e.g., look at my tracker.py `States` enum.)

lib/pbench/test/unit/server/test_user_create.py
```diff
@@ -91,12 +92,39 @@ def test_admin_user_creation(server_config, pytestconfig):
...
+                TestCreateUser.LAST_NAME_TEXT,
+                TestCreateUser.ROLE_SWITCH,
+                "Default",
+            ],
+            input=f"{TestCreateUser.PSWD_TEXT}\n",
+        )
+        assert result.exit_code == 0
         assert (
             result.stdout
             == f"{TestCreateUser.PSWD_PROMPT}\n" + "User test_user registered\n"
```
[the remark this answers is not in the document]
npalaska: Why should it fail? we could just ignore the `role` value and create a user with `None` as role value. Which is what this test case is doing, if we specify anything other than defined values in our enum class (at the moment just `ADMIN`), we would just ignore the parameter but create the user where `role` column is not set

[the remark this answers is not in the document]
webbnh: If as a user I type
```
$ pbench-create-user --username wscales --password hspwd  --role admn --email wscales@redhat.com --first-name Webb --last-name Scales
```
and the command appears to work, then I'm gonna be pissed when I log in and find that I don't have an admin role because my typo was quietly ignored.

Now, if you make `--role` a "choice", then Click should reject that command with an error, and I will see it, fix my typo, and try again.  But, we should have a unit test which proves that Click does that for us....

In general, silent failures (or silently ignoring explicitly specified options) is a Bad Thing™, and it's a Really Bad Thing™ when users are involved.

lib/pbench/cli/server/user_create.py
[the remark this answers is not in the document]
dbutenhof: "defaults to None" is still an awkward phrase, especially with "None" capitalized. If you don't specify a role, there are no roles enabled. "None" is not a role, nor are we presenting the possibility of specifying `--role=None`.

Further, your help already begins with "Optional"...

Oh, and it says "entitlement", which is obsolete. ;-)

dbutenhof (dismissed):

lib/pbench/server/database/models/users.py
```diff
@@ -112,15 +127,11 @@ def delete(username):
...
-        """this method would always return false for now until we add a notion of Admin user/group.
-        Once we know the admin credentials this method can check against those credentials to determine
-        whether the user is privileged to do more.
-
+        """This method checks whether the given user has an admin role.
         This can be extended to groups as well for example a user belonging to certain group has only those
         privileges that are assigned to the group.
         """
-        return False
+        return self.role == Roles.ADMIN
```
webbnh, line 134: Enums are supposed to be [compared by identity](https://docs.python.org/3/library/enum.html#comparisons) (i.e., using the `is` operator), but apparently the equality operator is defined.

lib/pbench/cli/server/user_create.py
```diff
@@ -30,8 +31,14 @@ def execute(self):
             last_name=self.context.last_name,
             email=self.context.email,
         )
+        if self.context.role:
+            user.role = self.context.role
```
webbnh, line 35: Since `role` is an attribute/column of `User`, shouldn't it be supplied to the c'tor?  It _is_ nullable, so I suppose it's a question of style, but something like this might be appropriate:
```
        user = User(
            username=self.context.username,
            password=self.context.password,
            first_name=self.context.first_name,
            last_name=self.context.last_name,
            email=self.context.email,
            role=self.context.role if self.context.role else None,
        )
```
I am prone to over-using the "ternary" (or conditional) operator, but I do like having all of the attributes provided at construction -- it tends to help keep them together.

webbnh (dismissed):
Looks good to go.  (I have two things to offer for your consideration.)

dbutenhof (approved):

portante (dismissed):

npalaska:
Rebasing all the commits now

dbutenhof (approved):

portante (approved):

lib/pbench/cli/server/user_create.py
```diff
@@ -29,9 +30,14 @@ def execute(self):
             first_name=self.context.first_name,
             last_name=self.context.last_name,
             email=self.context.email,
+            role=self.context.role if self.context.role else "",
```
webbnh, line 33: I'm happy to take your word for it, but just to confirm:  an empty string is the appropriate value for `role` if it is unspecified?  We want it to be represented in the database as "null" -- is that how we attain that?  (I would have thought that the appropriate value would be 'None`, but I'm not that familiar with SQLAlchemy....)

webbnh (approved):
Good to go, if you can confirm my question.

lib/pbench/cli/server/user_create.py
npalaska: If you look at line 80 of `lib/pbench/server/database/models/users.py` file (where we `validate` role column value) the return statement is like this `return Roles[value.upper()]` where we convert the role value to upper case and catch the `keyError` if no such a key is present in our Roles enum class. However, if the value passed is `None` it raises the attribute exception like NoneType object has no attribute upper(), that wont be the case if we pass an empty string as a role value (It will be saved in our table as None anyway because no empty key is present in our enum class). Now indeed it can also be fixed by catching a generic `Exception` but I thought if we can catch a specific error instead of generic exception that would be better.

portante: @webbnh are you good with this answer?

webbnh: I forgot about the field validator.

So, if the role string is not a valid role enum, then the field validator is going to return `None`, and the database entry will quietly be set to "null", which is what we want to have happen here.  We just need to be cognizant of this silent failure mode and be aware that we're relying on the command line processing to catch and report it.

OK, merge it!

PR #2219, opened 2021-04-15
Don't validate params for `pbench-fio` install

portante:
Now that we have tools only registered on the local host, `pbench-fio` cannot validate the tool group on remote hosts when it has been given the `--install` parameter.

We also remove `--install` from `pbench-fio` usage text since it is really an internal interface.

### DIFF
--- a/agent/bench-scripts/pbench-fio
+++ b/agent/bench-scripts/pbench-fio
@@ -15,7 +15,7 @@ benchmark_rpm=${benchmark}
 export benchmark_run_dir=""
 # allow unit tests to override
 if [[ -z "${benchmark_bin}" ]]; then
-        benchmark_bin="$(resolve_benchmark_bin "${benchmark}")"
+	benchmark_bin="$(resolve_benchmark_bin "${benchmark}")"
 	if [[ -z "${benchmark_bin}" ]]; then
 		error_log "[${script_name}] ${benchmark} executable not found on PATH=${PATH}"
 		exit 1
@@ -164,9 +164,6 @@ function usage {
 	printf -- "\t--max-failures=<int>\n"
 	printf "\t\tthe maximum number of failures to get below stddev\n"
 	printf "\n"
-	printf -- "\t--install\n"
-	printf "\t\tinstall only (default is False)\n"
-	printf "\n"
 	printf -- "\t--histogram-interval-sec=<int>\n"
 	printf "\t\tset the histogram logging interval in seconds (default $histogram_interval_sec)\n"
 	printf "\n"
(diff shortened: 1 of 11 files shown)


agent/bench-scripts/pbench-fio
```diff
@@ -449,6 +460,12 @@ function fio_process_options() {
 			;;
 		esac
 	done
+
+	if [[ "${install_only}" == "y" ]]; then
+		fio_local_check_install
+		exit ${?}
+	fi
```
dbutenhof, line 467: What confuses me a bit here is what seems to be an implicit assumption (based on collating the problem description and code changes) that `${install_only}` means "agent" vs "controller". Is that really guaranteed? Is there a way the environment dependency/assumption could be more explicit and robust?

portante: It really doesn't matter much, from what I can tell.

Right now, the only use of `--install` is from `pbench-fio` when it remotely invokes itself on the list of clients.

If somebody adds `--install` then `pbench-fio` will just check locally an exit immediately.

I have removed it from the `usage` output.

agent/bench-scripts/tests/pbench-fio/test-03.txt
```diff
@@ -0,0 +1,16 @@
...
+/var/tmp/pbench-test-bench/pbench-agent/tmp
+/var/tmp/pbench-test-bench/pbench-agent/tools-v1-default
+/var/tmp/pbench-test-bench/pbench-agent/tools-v1-default/__trigger__
+/var/tmp/pbench-test-bench/pbench-agent/tools-v1-default/testhost.example.com
+/var/tmp/pbench-test-bench/pbench-agent/tools-v1-default/testhost.example.com/mpstat
+/var/tmp/pbench-test-bench/pbench-agent/tools-v1-default/testhost.example.com/sar
+--- pbench tree state
++++ pbench.log file contents
+[debug][1900-01-01T00:00:00.000000] [check_install_rpm] fio-3.21 is installed
+[debug][1900-01-01T00:00:00.000000] [pbench-fio] fio-3.21 is installed
```
webbnh, line 15: Is there a test for the "is not installed" path?  (Should there be?)

webbnh (approved):

agent/bench-scripts/tests/pbench-fio/test-03.txt
portante: Our current unit tests don't check for this, but we now test that case for `require-rm` in PR #2210.

ndokos (dismissed):

portante (commented):
Rebased and squashed to one commit.

dbutenhof (approved):

ndokos (approved):

webbnh (approved):

PR #2222, opened 2021-04-15
Fix use of `PBENCH_REDIS_SERVER` by both `pbench-tool-meister-client` and `-stop`

portante:
When we implemented PR #2181, we failed to properly test for the bind/connect style of specification in our tests, and missed the fact that we didn't update `pbench-tool-meister-client` and `pbench-tool-meister-stop` to share the same code as is used by `pbench-tool-meister-start`.

There were 5 initial commits in this PR which should were squashed to one once they were reviewed:

 1. Refactor to share a `BaseReturnCode` class in `agent/utils.py`
 2. Refactor the `BaseServer` class into the `agent/utils.py` module
 3. Add tests that demonstrate things break
 4. Fix `pbench-tool-meister-stop`
 5. Fix `pbench-tool-meister-client`

This refactoring made it easier to see the progression of changes, facilitating a hopefully easier review.

The final review comment follows below.

----

Prior to this commit, `pbench-tool-meister-start` was the only CLI interface which was using the two forms for specifying the Redis server and Tool Data Sink host/port for binding and connecting.  Both the CLI commands for stopping the Tool Meister sub-system and communicating with it incorrectly expected the presence of the `PBENCH_REDIS_SERVER` to only specify a `<host>:<port>`, and also indicate that instance was not managed by CLI commands on this controller.

This commit corrects that where both `pbench-tool-meister-client` and `pbench-tool-meister-stop` now properly handle both forms of host/port specification, and `-stop` determines whether or not the CLI is in charge of orchestrating the server components by looking for the local pid files.

This was accomplished via a series of refactorings to share code across those three CLI commands:

   * Add and use `BaseReturnCode` in `lib/pbench/agent/utils.py`
   * Move BaseServer to `.../agent/utils.py`
   * Add a `RedisServer` class for `pbench-tool-meister-stop` to take advantage of the common parameter handling of the Redis server specification
   * In `pbench-tool-meister-client`, we use the new `RedisServerCommon` class shared with `-start` and `-stop`
   * We fix the new `BaseServer.kill()` to exit immediately after posting a `SIGKILL`
   * Add unit tests for all common code moved to `.../agent/utils.py`

### DIFF
--- a/agent/util-scripts/pbench-tool-meister-client
+++ b/agent/util-scripts/pbench-tool-meister-client
@@ -10,15 +10,12 @@ import logging
 import os
 import sys
 
-from pbench.agent.constants import (
-    cli_tm_allowed_actions,
-    cli_tm_channel_prefix,
-    def_redis_port,
-)
+from pbench.agent.constants import cli_tm_allowed_actions, cli_tm_channel_prefix
 from pbench.agent.tool_meister_client import Client
+from pbench.agent.utils import RedisServerCommon
 
 
-def main(argv):
+def main(argv: list) -> int:
     """Main program for the Tool Meister client CLI.  The command line
     arguments are:
 
@@ -73,23 +70,16 @@ def main(argv):
             # properly shut down.
             return 0
 
-    redis_server = os.environ.get("PBENCH_REDIS_SERVER", f"localhost:{def_redis_port}")
-    parts = redis_server.split(":", 1)
-    if len(parts) != 2:
-        logger.error("Bad Redis server specified, {!r}", redis_server)
-        return 1
+    redis_server_env = os.environ.get("PBENCH_REDIS_SERVER", "")
     try:
-        redis_port = int(parts[1])
-    except Exception:
-        logger.error("Bad port for Redis server specified in {!r}", redis_server)
-        return 1
-    else:
-        redis_host = parts[0]
+        redis_server = RedisServerCommon(redis_server_env, "localhost")
+    except RedisServerCommon.Err as exc:
+        logger.error(str(exc))
+        return exc.return_code
 
-    # The Redis server is always running on the local host with the CLI.
     with Client(
-        redis_host=redis_host,
-        redis_port=redis_port,
+        redis_host=redis_server.host,
+        redis_port=redis_server.port,
         channel_prefix=cli_tm_channel_prefix,
         logger=logger,
     ) as client:
--- a/agent/util-scripts/pbench-tool-meister-start
+++ b/agent/util-scripts/pbench-tool-meister-start
@@ -97,12 +97,10 @@ the benchmark execution environment:
   - ssh-opts
 """
 
-import errno
 import json
 import logging
 import os
 import shlex
-import signal
 import socket
 import sys
 import time
@@ -128,9 +126,12 @@ from pbench.agent.tool_meister import main as tm_main
 from pbench.agent.tool_meister_client import Client
 from pbench.agent.toolmetadata import ToolMetadata
 from pbench.agent.utils import (
+    BaseReturnCode,
+    BaseServer,
     cli_verify_sysinfo,
     error_log,
     info_log,
+    RedisServerCommon,
     validate_hostname,
 )
 
@@ -140,12 +141,11 @@ from pbench.agent.utils import (
 _TDS_STARTUP_TIMEOUT = 60
 
 
-class ReturnCode:
+class ReturnCode(BaseReturnCode):
     """ReturnCode - symbolic return codes for the main program of
     pbench-tool-meister-start.
     """
 
-    SUCCESS = 0
     BADTOOLGROUP = 1
     BADAGENTCONFIG = 2
     MISSINGINSTALLDIR = 4
@@ -182,37 +182,8 @@ class ReturnCode:
     BADFULLHOSTNAME = 37
     BADHOSTNAME = 38
 
-    # Kill sub-codes
-    KILL_SUCCESS = 0
-    KILL_READEXC = 1
-    KILL_BADPID = 2
-    KILL_PIDNOTFOUND = 3
-    KILL_KILLERR = 4
-    KILL_KILLEXC = 5
 
-    class Err(RuntimeError):
-        """Err - exception definition to capture return code as an attribute.
-        """
-
-        def __init__(self, message: str, return_code: int):
-            """Adds a return_code attribute to capture an integer representing
-            the return code a caller can pass along.
-            """
-            super().__init__(message)
-            self.return_code = return_code
-
-    @staticmethod
-    def kill_ret_code(kill_code: int, ret_val: int):
-        """kill_ret_code - return an integer return code made up of the given
-        kill code and a return value.
-
-        A kill code of 0 and return value of 42 is returned as 42.
-        A kill code of 5 and return value of 52 is returned as 542.
-        """
-        return (kill_code * 100) + ret_val
-
-
-def _waitpid(pid: int):
+def _waitpid(pid: int) -> int:
     """Wrapper for os.waitpid()
 
     Returns the exit status of the given process ID.
@@ -241,10 +212,10 @@ def start_tms_via_ssh(
     tool_group: str,
     ssh_opts: str,
     full_hostname: str,
-    redis_server,
+    redis_server: RedisServerCommon,
     redis_client: redis.Redis,
     logger: logging.Logger,
-):
+) -> None:
     """start_tms_via_ssh - orchestrate the creation of local and remote Tool
     Meister instances using ssh for those that are remote.
 
@@ -364,88 +335,6 @@ def start_tms_via_ssh(
         )
 
 
-class BaseServer:
-    """BaseServer - abstract base class for common code shared between the
-    ToolDataSink and RedisServer classes.
-    """
-
-    def_port = None
-    bad_port_code = None
-    bad_host_code = None
-    name = None
-
-    class Err(ReturnCode.Err):
-        """BaseServer.Err - derived from ReturnCode.Err, specifically raised by
-        BaseServer and its derived classes.
-        """
-
-        pass
-
-    def __init__(self, spec: str, def_host_name: str):
-        """__init__ - from the given IP/port specification given, determine the
-        IP:port for binding (listening) and the IP:port for connecting.
-
-        The IP/port specification can be given in one of two forms:
-
-          - `<ip>:<port>'
-            * where the same ip address and port are used for binding and
-              connecting
-          - `<bind ip>:<port>;<connect ip>:<port>`
-            * where a semi-colon separates the bind ip/port from the connecting
-              ip/port
-
-        In either case, a missing port (bare colon, optional) indicates the
-        default port should be used. If no IP address is given, the default
-        host name is used.
-
-        No attempt is made to verify that the IP address resolves, or that it
-        is reachable, though we do check they are syntactically valid.
-        """
-        _spec = spec if spec else def_host_name
-        parts = _spec.split(";", 1)
-        pairs = []
-        for part in parts:
-            host_port_parts = part.rsplit(":", 1)
-            if len(host_port_parts) == 1:
-                port = self.def_port
-            else:
-                try:
-                    port = int(host_port_parts[1])
-                except ValueError as exc:
-                    if host_port_parts[1] == "":
-                        port = self.def_port
-                    else:
-                        raise self.Err(
-                            f"Bad port specified for {self.name} in '{spec}'",
-                            self.bad_port_ret_code,
-                        ) from exc
-            host = host_port_parts[0] if host_port_parts[0] else def_host_name
-            if host[0] == "[" and host[-1] == "]":
-                # Brackets are invalid for a host name, but might be used when
-                # specifying a port with an IPv6 address, strip them before we
-                # validate the host name.
-                host = host[1:-1]
-            if validate_hostname(host) != 0:
-                raise self.Err(
-                    f"Bad host specified for {self.name} in '{spec}'",
-                    self.bad_host_ret_code,
-                )
-            pairs.append((host, port))
-
-        self.bind_host, self.bind_port = pairs[0]
-        if len(pairs) == 2:
-            # Separate bind/connecting ip:port
-            self.host, self.port = pairs[1]
-            self._repr = f"{self.name} - {self.bind_host}:{self.bind_port} / {self.host}:{self.port}"
-        else:
-            assert len(pairs) == 1, "Logic bomb!  unexpected pairs, {pairs!r}"
-            self.host, self.port = pairs[0]
-            self._repr = f"{self.name} - {self.host}:{self.port}"
-
-    def __repr__(self):
-        return self._repr
-
-
 class ToolDataSink(BaseServer):
     """ToolDataSink - an encapsulation of the handling of the Tool Data Sink
     specification and methods to optionally create and manage an instance.
@@ -461,10 +350,10 @@ class ToolDataSink(BaseServer):
         exec_dir: Path,
         full_hostname: str,
         tds_param_key: str,
-        redis_server,
+        redis_server: RedisServerCommon,
         redis_client: redis.Redis,
         logger: logging.Logger,
-    ):
+    ) -> None:
         assert (
             self.host is not None
             and self.port is not None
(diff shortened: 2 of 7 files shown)


lib/pbench/agent/utils.py
```diff
@@ -14,6 +17,181 @@
...
+    """
+
+    # Common success code is zero.
+    SUCCESS = 0
+
+    # Common kill sub-codes
+    KILL_SUCCESS = 0
+    KILL_READEXC = 1
+    KILL_BADPID = 2
+    KILL_PIDNOTFOUND = 3
```
portante, line 32: Looks like this can be removed.

portante (commented):
Looks good for initial review.

agent/util-scripts/pbench-tool-meister-stop
```diff
@@ -45,8 +44,35 @@ def wait_for_pid(pid):
...
+        super().__init__(spec, def_host_name)
+        try:
+            self.pid_file = (
+                benchmark_run_dir / "tm" / f"redis_{def_redis_port:d}.pid"
+            ).resolve(strict=True)
+        except FileNotFoundError:
+            self.pid_file = None
+
+    def locally_managed(self):
+        return True if self.pid_file is not None else False
```
dbutenhof, line 67: Why not just `return self.pid_file is not None`??

(You could just `return self.pid_file` but I do like the idea of converting the file path to a boolean here...)

dbutenhof (dismissed):
I think this looks OK...

agent/util-scripts/pbench-tool-meister-stop
portante: Ah, yeah, good question.  I'll fix.

lib/pbench/agent/utils.py
```diff
@@ -14,6 +17,181 @@
...
+                # specifying a port with an IPv6 address, strip them before we
+                # validate the host name.
+                host = host[1:-1]
+            if validate_hostname(host) != 0:
+                raise self.Err(
+                    f"Bad host specified for {self.name} in '{spec}'",
+                    self.bad_host_ret_code,
+                )
+            pairs.append((host, port))
+
```
ndokos, line 125: Is it worth making this a function and adding an all-possible-permutations test for it?

ndokos (changes requested):

agent/util-scripts/pbench-tool-meister-stop
webbnh: Extending @dbutenhof's suggestion:
```
    def locally_managed(self):
        return bool(self.pid_file)
```

:-)

webbnh (approved):
Looks OK.  (In fact, there are some nice improvements there!)

I'm not crazy about the parallel definitions of `RedisServer`, but I can guess why you did that.  It seems like it would be better to provide a single implementation which includes the capabilities which each consumer needs, although in at least one case it seems like you could get by on the base class alone.

And, of course, there is the usual collection of nits and suggestions.

RobertKrawitz:
+1
Verified against bench-army-knife.

agent/util-scripts/pbench-tool-meister-stop
portante: Because this would return `True` for any `Path` object, even if that object is not associated with a path.

lib/pbench/agent/utils.py
portante: Yes, good idea, I should add proper unit tests for this refactored code.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -140,12 +140,11 @@ from pbench.agent.utils import (
 _TDS_STARTUP_TIMEOUT = 60
 
 
-class ReturnCode:
+class ReturnCode(BaseReturnCode):
```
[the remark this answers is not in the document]
portante: > I'm unclear on how we're going to avoid name collisions on values

Good question, I'll investigate; one possibility is to employ a unit test to compare the base class codes against the sub-classes to ensure no overlap.

> how we avoid unintentional value collisions

The way the `ReturnCode` "kill" works is it shifts the codes by 1,000.

agent/util-scripts/pbench-tool-meister-stop
```diff
@@ -89,37 +115,15 @@ def graceful_shutdown(
...
-                else:
-                    if time.time() > timeout:
-                        try:
-                            os.kill(redis_server_pid, signal.SIGKILL)
-                        except ProcessLookupError:
-                            pid_exists = False
-                        except Exception:
-                            raise
-                    time.sleep(0.1)
+        redis_server.kill(1)
```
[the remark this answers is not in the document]
portante: The `kill()` method of `BaseServer` takes a return value to pass along on success, or change to a `KILL_` value.

agent/util-scripts/pbench-tool-meister-client
```diff
@@ -16,6 +16,18 @@ from pbench.agent.constants import (
...
+
+class RedisServer(BaseServer):
+    """RedisServer - an encapsulation of the handling of the Redis server
+    specification.
+    """
+
+    def_port = def_redis_port
+    bad_port_ret_code = 1
+    bad_host_ret_code = 1
+    name = "Redis server"
```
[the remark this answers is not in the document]
portante: > Why isn't the definition of `RedisServer` common and shared with `pbench-tool-meister-start` and `-stop`?

Basically because it _seemed_ it would result in more code for very little gain.

At the time I did not want to embark on solving the return codes for `-stop` and `-client` while doing this work, but just wanted the shared constructor behavior.

Thank you for prompting me to look at this a bit more, as I think we can solve this elegantly.

agent/util-scripts/pbench-tool-meister-stop
webbnh: `bool(self.pid_file)` is roughly equivalent to `self.pid_file not in [0, False, None, "", (), [], {}]`, so it's actually a superset of `self.pid_file is not None`.

That is, if `self.pid_file` is a `Path` object (or anything else which is not `None`), your code will return `True` as well, just like mine does.

There's only a difference if `self.pid_file` somehow evaluates to a "Falsey" value other than `None`.

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
webbnh: OK, so the `KILL_` values are a different "type".  (One wonders if they shouldn't be in their own `IntEnum`.)

agent/util-scripts/pbench-tool-meister-stop
```diff
@@ -18,10 +18,15 @@ import time
 from argparse import ArgumentParser, Namespace
 from pathlib import Path
 
-from pbench.agent.constants import def_redis_port, cli_tm_channel_prefix
+from pbench.agent.constants import cli_tm_channel_prefix
```
webbnh, line 21: `pbench-tool-meister-start` references `def_redis_port` in its help text.  Should `pbench-tool-meister-stop` be doing the same?  Or, better yet, should the two tools be sharing their argument definitions for the pieces which should be uniform between them?

agent/util-scripts/pbench-tool-meister-stop
```diff
@@ -45,8 +49,34 @@ def wait_for_pid(pid):
...
+                benchmark_run_dir / "tm" / f"redis_{self.def_port:d}.pid"
+            ).resolve(strict=True)
+        except FileNotFoundError:
+            pass
+
+    def locally_managed(self):
+        return self.pid_file is not None
+
+
 def graceful_shutdown(
```
webbnh, line 74: There's no description (nor type hint) for the return value from this `graceful_shutdown()` (nor for `locally_managed()`, but that one is easier to infer).

webbnh (approved):
Looks good (the refactor of `RedisServerCommon` is a solid improvement, although whether the server is locally managed or not seems like it would be of broader interest, but it's OK as is).

Two items for your consideration.

lib/pbench/agent/utils.py
portante: Done.

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
portante: Tracking in #2235.

portante (commented):
Okay, added a new commit with the unit tests.  Will rebase before merging.

portante (commented):
Okay, missed a few things, and fixed LGTM comment ...

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -244,7 +215,7 @@ def start_tms_via_ssh(
     redis_server,
```
ndokos, line 215: No type hint here? Isn't it a `RedisServerCommon`? Or maybe a `RedisServer`? Is it n ordering problem (the class is not defined yet)?

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -464,7 +353,7 @@ class ToolDataSink(BaseServer):
         redis_server,
```
ndokos, line 353: Type hint?

ndokos (commented):
A couple of comments, plus I still need to review `test_utils.py`.

agent/util-scripts/pbench-tool-meister-start
portante: Yes, ordering problem and I did not want to blow up the code with a major code move as part of the PR.

agent/util-scripts/pbench-tool-meister-start
portante: Yes, ordering problem and I did not want to blow up the code with a major code move as part of the PR.

lib/pbench/test/unit/agent/test_utils.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def test_constructor(self):
+        with pytest.raises(AssertionError):
+            bs = OurServer("", "")
+
+        for arg1, arg2 in [("", "localhost"), ("localhost", "localhost")]:
+            bs = OurServer(arg1, arg2)
+            assert bs.pid_file is None
+            assert bs.host == "localhost"
+            assert bs.port == 4242
```
webbnh, line 32: Shouldn't the RHS be `OurServer.def_port` rather than a copy of the magic number?

Likewise, shouldn't the RHS in line 41 be `OurServer.bad_host_ret_code`?

Etc.

webbnh: Or, if you think accessing the class attributes is too "risky", then it seems like you should be using symbolic constants for them defined earlier and used to initialize the class.

lib/pbench/test/unit/agent/test_utils.py
```diff
@@ -0,0 +1,210 @@
...
+        ret = bs.kill(42)
+        assert ret == 42
+
+        class MockPath:
+            def __init__(self, kind):
+                self.kind = kind
+
+            def read_text(self):
+                if self.kind == "os":
+                    raise OSError("fake oserror")
```
webbnh, line 118: That doesn't look much like an [`OSError`](https://docs.python.org/3/library/exceptions.html#OSError) -- it should have an `errno` code as it's first argument, I think.

webbnh: Rather than having the c'tor accept a string as an argument (type hints would be handy, BTW), how about having it accept the exception object to raise?  That would decouple and simplify things a bit.

lib/pbench/test/unit/agent/test_utils.py
```diff
@@ -0,0 +1,210 @@
...
+                return
+
+        class MockKill:
+            behaviors = {
+                "1001": (True, True, False, False, False, False),
+                "1002": (False, False, False, False, False, False),
+                "1003": (True, True, False, True, False, False),
+                "1004": (True, False, False, True, False, False),
+                "1005": (True, False, False, False, True, False),
+                "1006": (True, False, False, False, False, True),
```
webbnh, line 148: The third column ("killed by term"?) is always `False`.  Are you missing a test case, or should this column be omitted from the table?

webbnh (approved):
Looks OK...unless you want to address any of the items below.

lib/pbench/agent/utils.py
```diff
@@ -149,11 +152,10 @@ def kill(self, ret_val: int):
 
         try:
             raw_pid = self.pid_file.read_text()
-        except OSError as exc:
-            if exc.errno != errno.ENOENT:
-                return BaseReturnCode.kill_ret_code(
-                    BaseReturnCode.KILL_READEXC, ret_val
-                )
+        except FileNotFoundError:
```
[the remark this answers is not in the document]
portante: Of the code being changed in this commit, I have made this modification as a separate commit for those files that were touched.  I am not going to make this change for any other module.

agent/util-scripts/pbench-tool-meister-stop
```diff
@@ -77,7 +78,7 @@ def graceful_shutdown(
     group: str,
     redis_server: RedisServer,
     logger: logging.Logger,
-):
+) -> int:
```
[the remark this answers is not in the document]
portante: Fixed in a separate commit.

agent/util-scripts/pbench-tool-meister-start
portante: Fixed in a separate commit.

lib/pbench/test/unit/agent/test_utils.py
portante: Fixed as a separate commit.

lib/pbench/test/unit/agent/test_utils.py
portante: The first case was incorrect, fixed in a separate commit.

lib/pbench/test/unit/agent/test_utils.py
portante: Fixed as a separate commit.

portante (commented):
There are 16 commits in this PR, which we'll want to squash before merging.

I have set this to `Draft` mode until we deem the code fit, then I'll squash them all down and we can finalize the approvals.

lib/pbench/test/unit/agent/test_utils.py
```diff
@@ -29,20 +29,24 @@ def test_constructor(self):
             bs = OurServer(arg1, arg2)
             assert bs.pid_file is None
             assert bs.host == "localhost"
-            assert bs.port == 4242
+            assert bs.port == OurServer.def_port
             assert bs.bind_host == bs.host
             assert bs.bind_port == bs.port
-            assert repr(bs) == "forty-two - localhost:4242"
+            assert repr(bs) == f"forty-two - localhost:{OurServer.def_port:d}"
```
webbnh, line 35: Since you made the port value dynamic, do you want to make the server name and/or the hostname dynamic?

Alternatively, if this is only used for debugging, do we really want to be testing it at all?

lib/pbench/test/unit/agent/test_utils.py
```diff
@@ -29,20 +29,24 @@ def test_constructor(self):
...
 
         _def = "localhost"
 
         with pytest.raises(OurServer.Err) as excinfo:
             bs = OurServer("bad_host.example.com", _def)
-        assert excinfo.value.return_code == 43, f"{excinfo.value!r}"
+        assert (
+            excinfo.value.return_code == OurServer.bad_host_ret_code
+        ), f"{excinfo.value!r}"
```
webbnh, line 43: I hate to ask, but...does `excinfo` have a `value` attribute?  If so, what is it?

(And, is `f"{excinfo.value!r}"` better than `repr(excinfo.value)`?)

Ditto line 49.

webbnh (approved):
If your intent is to squash this branch into a single commit on the end of the `main` branch, I believe that you can do so using the button below after you get the approvals, without having to do an external squash and the subsequent round of re-approvals....

In any case, I deem the code to be fit, unless either of the comments below persuades you otherwise.

lib/pbench/test/unit/agent/test_utils.py
portante: `value` is the exception object that was raised.  Go figure.

lib/pbench/test/unit/agent/test_utils.py
portante: I don't think so.  Using the `OurServer.def_port` class variable is fine because to avoid further perpetuating a constant, but `localhost` is so common that it doesn't gain us much making it dynamic.

ndokos (approved):

webbnh (approved):
GitHub sez nothing has changed since my last review, so it must be good to go!

PR #2226, opened 2021-04-17
Ensure we always pull jenkins devel images

portante:
We need to use the `podman` option `--pull=always` to ensure we properly fetch the updated `pbench-devel` image when running our CI jobs.

The default behavior is `missing`, which means only pull the image from the repository if it is not found on the local host.


dbutenhof (approved):
I guess "always" is better than "never" in this case; apparently it can't check with the repository and pull if `latest` is attached to a more recent container? Weird.

npalaska (approved):

PR #2227, opened 2021-04-17
Only measure coverage in `lib/pbench`

portante:
We tell `coverage` to explicitly check all modules in the `lib/pbench` tree only.  Since we can only measure code coverage when using `pytest`, anyways, we won't see any measurements of modules outside of the `lib/pbench` tree that we care about, so using this option is a simple way to make sure we have a clean data set that considers all of our measurable code.

This change ensures that all of the modules under `lib/pbench` are considered when generating coverage reports, instead of only those modules that were touched by tests.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,4 @@
 [run]
 parallel = True
 data_file = /tmp/${USER}/cov/coverage.db
-omit = */site-packages/*
+source = lib/pbench


portante:
This really shows how much we are missing from our coverage in unit tests.

.coveragerc
```diff
@@ -1,4 +1,9 @@
 [run]
 parallel = True
 data_file = /tmp/${USER}/cov/coverage.db
-omit = */site-packages/*
+source =
+    lib/pbench
+omit =
+    */site-packages/*
+    /tmp/${USER}/tox/*/bin/bottle.py
```
dbutenhof, line 8: This seems oddly specific; e.g., note that `gunicorn` and `alembic` which we use on the server side are here, too. Should we be (_can_ we?) ignore everything here except `pbench_*`??

.coveragerc
```diff
@@ -1,4 +1,9 @@
...
+    /tmp/${USER}/tox/*/lib/python3*
```
dbutenhof, line 9: What else do you see here? I don't have an agent tox environment fired up, but `ls /tmp/dbutenho/tox/*/lib/python3*/` only shows me `site-packages`, which is already omitted above.

portante: I had another part of the tree with a `distutils` hierarchy.

.coveragerc
portante: This is the only one that showed up in my environment.  I am not sure how to only include `pbench_*` names while excluding others.

.coveragerc
webbnh, line 9: I could use some help with the context, here.

Before, we were testing everything under the repo root, and now we're going to test only the stuff under `lib/pbench`?  (That seems like a narrowing of focus -- how does it now find **more** stuff which is untested?)

Starting at `lib/pbench`, how does it end up looking in places like `/tmp` such that we need to exclude them?  Does it have other (implicit) start locations?  Is there something in the code that we're having it test which reaches out to `/tmp`?  (E.g., should we be excluding `lib/pbench/test` -- we don't want to include the test coverage machinery in the code being coverage tested...right?)

webbnh: >  does not include command line interface files without `.py` extension.

I missed the directive for this...is it on the command line or in some other configuration file?  (Or, is this part of the `site-packages` exclusion?)

ndokos: > I could use some help with the context, here.

Me too - starting with how to run the thing... I'm almost certainly missing a bunch of things, but grepping through the tree looking for `coverage` finds precious little. Help!

ndokos (commented):
How do I run `coverage` in the tree?

.coveragerc
portante: I've added comments here to see if this helps make things clearer.

.coveragerc
portante: I've added comments here to record why we need these omissions.

portante: > Before, we were testing everything under the repo root, and now we're going to test only the stuff under `lib/pbench`? (That seems like a narrowing of focus -- how does it now find **more** stuff which is untested?)

The `source` statement does not mean a narrowing of what is considered for coverage, but the opposite: it directs the coverage.py utility to inspect all the modules found under that tree and record what was found, in addition to modules executed (and not in the list of omissions), so that one can see what was covered or not for the entire library, not just those modules that happen to have some part of the tested.

> Starting at `lib/pbench`, how does it end up looking in places like `/tmp` such that we need to exclude them? Does it have other (implicit) start locations? Is there something in the code that we're having it test which reaches out to `/tmp`? (E.g., should we be excluding `lib/pbench/test` -- we don't want to include the test coverage machinery in the code being coverage tested...right?)

This is mostly because of how `tox` is directed to use `/tmp/${USER}` a [line 6](https://github.com/distributed-system-analysis/pbench/blob/main/tox.ini#L6) of `tox.ini` (when `TOXWORKDIR` is not pre-defined).  So all pip installs land under that `/tmp/${USER}` hierarchy.

> > does not include command line interface files without `.py` extension.
> 
> I missed the directive for this...is it on the command line or in some other configuration file? (Or, is this part of the `site-packages` exclusion?)
> 
> Or is it because the tool has [always only measured files with the `.py` extension](https://coverage.readthedocs.io/en/coverage-5.0/source.html#execution)?

It is not clear this is about what coverage measures or not: in the absence of any `source`, `include`, or `omit` directives, coverage will record functions executed in any python code it is found.

This is about what additional data surrounding the executable code is included in addition to what is executed, and what should be ignored or not.

So when you specify `source` and a directory, it only looks for `*.py` files in that hierarchy.  If you want to explicitly include files without a `.py` extension, then you have to specify them individually IIF they are not executed.

portante: > > I could use some help with the context, here.
> 
> Me too - starting with how to run the thing... I'm almost certainly missing a bunch of things, but grepping through the tree looking for `coverage` finds precious little. Help!

Coverage data is always collected in `/tmp/${USER}/cov` for a given `tox` environment (if that environment executes python3 code).

To see that coverage, you have to either review the generated HTML files (see below), or run the `coverage` command in the top-level of the pbench source tree (where the `.coverage` file is located).

You can find the output of the `tox` tests what we are doing today, e.g. here is the output from a `tox -e py3-agent`:

```
py3-agent run-test: commands[0] | pytest --cov --cov-append --cov-report html:/tmp/${USER}/tox/py3-agent/cov/html ./lib/pbench/test/unit/common ./lib/pbench/test/unit/agent ./lib/pbench/test/functional
```

And near the end of the `tox` output you'll see:

```
----------- coverage: platform linux, python 3.8.6-final-0 -----------
Coverage HTML written to dir /tmp/${USER}/tox/py3-agent/cov/html
```

.coveragerc
portante: As @webbnh points out, `source` (unfortunately) causes all other python code to be ignored.  So the `omit` is no longer needed.

.coveragerc
portante: When the `source` directive is not used, we see `distutils` as well as `site-packages`.  When using `source` we don't see anything now.

portante: Regarding @webbnh's comments, yes, you are correct that `source` narrows the modules measured so that the `omit` statements are not needed.

portante (commented):
Much simpler, revised the commit message.

portante:
> How do I run `coverage` in the tree?

Coverage data is always collected in `/tmp/${USER}/cov` for a given `tox` environment (if that environment executes python3 code).

To see that coverage, you have to either review the generated HTML files (see below), or run the `coverage` command in the top-level of the pbench source tree (where the `.coverage` file is located).

You can find the output of the `tox` tests what we are doing today, e.g. here is the output from a `tox -e py3-agent`:

```
py3-agent run-test: commands[0] | pytest --cov --cov-append --cov-report html:/tmp/${USER}/tox/py3-agent/cov/html ./lib/pbench/test/unit/common ./lib/pbench/test/unit/agent ./lib/pbench/test/functional
```

And near the end of the `tox` output you'll see:

```
----------- coverage: platform linux, python 3.8.6-final-0 -----------
Coverage HTML written to dir /tmp/${USER}/tox/py3-agent/cov/html
```

ndokos (approved):

ndokos:
Thanks for the instructions! Very helpful. I should have grepped for `--cov`, instead of `coverage`.

ndokos:
`lib/pbench` contains both client and server bits, so when I run py3-agent tests, the resulting coverage has numbers for both, so the agent numbers are skewed. SImilarly, py3-server includes agent numbers, We may have to reorganize the lib directory to decouple them from each other or else use a separate source option for each of the tests (if that is possible).

portante:
> `lib/pbench` contains both client and server bits, so when I run py3-agent tests, the resulting coverage has numbers for both, so the agent numbers are skewed. SImilarly, py3-server includes agent numbers, We may have to reorganize the lib directory to decouple them from each other or else use a separate source option for each of the tests (if that is possible).

You use `--omit` on the report to on see specific sub-sets: `coverage report --omit=lib/pbench/server/*,lib/pbench/cli/server/*,lib/pbench/test/*`.

dbutenhof (approved):

PR #2228, opened 2021-04-17
Share mocks between persistent tools

portante:
This is work pulled over from the `comp-150` branch.

### DIFF
--- a/agent/util-scripts/test-bin/dcgm-exporter
+++ b/agent/util-scripts/test-bin/dcgm-exporter
@@ -1,17 +1,1 @@
-#!/usr/bin/env python2
-
-import os
-import sys
-import time
-
-with open(os.environ["_testlog"], "a") as ofp:
-    args = " ".join(sys.argv)
-    ofp.write("%s\n" % (args,))
-
-with open("dcgm.file", "a+") as ofp:
-    args = " ".join(sys.argv)
-    ofp.write("%s\n" % (args,))
-
-time.sleep(9999)
-
-sys.exit(0)
+mock-cmd
--- a/agent/util-scripts/test-bin/mock-cmd
+++ b/agent/util-scripts/test-bin/mock-cmd
@@ -6,9 +6,8 @@ import time
 
 from pathlib import Path
 
-
+args = " ".join(sys.argv)
 with open(os.environ["_testlog"], "a") as ofp:
-    args = " ".join(sys.argv)
     ofp.write(f"{args}\n")
 
 _prog = os.path.basename(sys.argv[0])
@@ -17,15 +16,15 @@ _prog = os.path.basename(sys.argv[0])
 # ensure we don't run programs where they were not supposed to run.
 if _prog not in ("pbench-tool-meister-client", "yum", "scp"):
     with open(f"{_prog}.file", "a+") as ofp:
-        args = " ".join(sys.argv)
         ofp.write(f"{args}\n")
 
+_sleepers = ["dcgm-exporter", "node_exporter", "pmcd"]
 _sleep = 0
 if _prog == "pbench-sysinfo-dump":
     tgtdir = Path(sys.argv[1])
     sysinfo = sys.argv[2]
     (tgtdir / "contents.lis").write_text(sysinfo)
-elif _prog == "node_exporter" or (_prog == "prometheus" and sys.argv[1] == "run"):
+elif _prog in _sleepers or (_prog == "prometheus" and sys.argv[1] == "run"):
     _sleep = 9999
 
 if _sleep > 0:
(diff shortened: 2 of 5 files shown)


agent/util-scripts/test-bin/node_exporter
```diff
@@ -1 +1 @@
-mock-cmd
\ No newline at end of file
+dcgm-exporter
```
dbutenhof, line 1: Hmm, not to lob random monkey wrenches or anything; but I could see some case for renaming the mock from `dgcm-exporter` to something more generic given these changes.

dbutenhof (dismissed):

npalaska (approved):

agent/util-scripts/test-bin/dcgm-exporter
```diff
@@ -1,16 +1,16 @@
...
-with open(os.environ["_testlog"], "a") as ofp:
-    args = " ".join(sys.argv)
-    ofp.write("%s\n" % (args,))
+_prog = os.path.basename(sys.argv[0])
 
-with open("dcgm.file", "a+") as ofp:
+with open(os.environ["_testlog"], "a") as tfp, open(f"{_prog}.file", "a+") as ofp:
     args = " ".join(sys.argv)
-    ofp.write("%s\n" % (args,))
+    line = f"{args}\n"
```
webbnh, line 11: Seems like you could do without `args`:
```
    line = " ".join(sys.argv) + "\n"
```

agent/util-scripts/test-bin/node_exporter
webbnh: Since the currently-named `dgcm-exporter` is used by both `node_exporter` and `pmcd`, it _does_ seem like it ought to have a more generic name.

My question is, why are we maintaining `agent/util-scripts/test-bin/dcgm-exporter` separate from `agent/util-scripts/test-bin/mock-cmd`?  They seem similar enough that they could be merged, and then all of this magic would be in one place (and we wouldn't have to worry about naming the second one of these).

webbnh (approved):
Looks OK but it seems like `agent/util-scripts/test-bin/dcgm-exporter` and `agent/util-scripts/test-bin/mock-cmd` could/should be merged.  (And, there's another nit/suggestion.)

ndokos (changes requested):
OK - I'll lob the monkey wrench: seems to me that the comments of @dbutenhof and @webbnh should be addressed before this is merged. In particular, I don't understand why they don't all funnel to `mock-cmd`.

agent/util-scripts/test-bin/dcgm-exporter
portante: Yes, that is a possibility.

portante (commented):
Okay, re-worked this change based on code review feedback.

agent/util-scripts/test-bin/mock-cmd
```diff
@@ -17,15 +16,15 @@ _prog = os.path.basename(sys.argv[0])
...
         ofp.write(f"{args}\n")
 
+_sleepers = ["dcgm-exporter", "node_exporter", "pmcd"]
 _sleep = 0
 if _prog == "pbench-sysinfo-dump":
     tgtdir = Path(sys.argv[1])
     sysinfo = sys.argv[2]
     (tgtdir / "contents.lis").write_text(sysinfo)
-elif _prog == "node_exporter" or (_prog == "prometheus" and sys.argv[1] == "run"):
+elif _prog in _sleepers or (_prog == "prometheus" and sys.argv[1] == "run"):
```
dbutenhof, line 27: Ugh; but I suppose the special case of `argv[1]` for prometheus precludes including it in `_sleepers`. It's still ugly!

dbutenhof (approved):

agent/util-scripts/test-bin/mock-cmd
portante: This is why I wanted a separate "mock" for the persistent tools, but with folks wanting to share it more, we get this kind of morass.

Hopefully, the more we convert `bash` to Python3, the more this kind of construct will go away.

portante:
Rebased ...

dbutenhof (approved):

agent/util-scripts/test-bin/mock-cmd
ndokos: Sorry - I didn't realize prometheus was a specilal case: I thought it *was* a persistent tool.

ndokos (approved):

agent/util-scripts/test-bin/mock-cmd
portante: No problem!

webbnh: If it provides better aesthetics, we can have `_prog in _sleepers` in a separate `elif` from the Prometheus special case check.

In my opinion, the `_sleep` variable isn't buying us anything -- that is, if you don't divide Prometheus from the other sleepers, you can get rid of `_sleep` and just invoke `time.sleep()` in the "then" clause.  (I can see why you might want the invocation to be separate, but until we get a case other than `9999`, it doesn't buy us anything.)

So:
```
_sleepers = ["dcgm-exporter", "node_exporter", "pmcd", "prometheus"]
if _prog == "pbench-sysinfo-dump":
    tgtdir = Path(sys.argv[1])
    sysinfo = sys.argv[2]
    (tgtdir / "contents.lis").write_text(sysinfo)
elif _prog == "prometheus" and sys.argv[1] != "run")
    pass
elif _prog in _sleepers:
    time.sleep(9999)

sys.exit(0)
```

portante: Track in issue #2237.

PR #2232, opened 2021-04-20
Add Authorization header to upload PUT API

npalaska:
- Adds a strict Bearer `Authorization` header requirement to upload API
- Dataset owner's username is extracted from the authorization header

### DIFF
--- a/lib/pbench/server/api/resources/upload_api.py
+++ b/lib/pbench/server/api/resources/upload_api.py
@@ -1,16 +1,17 @@
-import os
-import humanize
-import tempfile
 import hashlib
-from pathlib import Path
+import os
+import tempfile
 from http import HTTPStatus
+from pathlib import Path
 
+import humanize
+from flask import jsonify, request
 from flask_restful import Resource, abort
-from flask import request, jsonify
 from werkzeug.utils import secure_filename
-from pbench.server.utils import filesize_bytes
-from pbench.server.database.models.tracker import Dataset, States
 
+from pbench.server.api.auth import Auth
+from pbench.server.database.models.tracker import Dataset, DatasetDuplicate, States
+from pbench.server.utils import filesize_bytes
 
 ALLOWED_EXTENSIONS = {"xz"}
 
@@ -48,10 +49,14 @@ class Upload(Resource):
             )
         )
 
+    @Auth.token_auth.login_required()
     def put(self, controller):
 
-        # FIXME: This should be assigned from the decoded authorization token
-        username = "pbench"
+        try:
+            username = Auth.token_auth.current_user().username
+        except Exception:
+            self.logger.exception("Tarfile upload: Exception verifying the username")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
         if not request.headers.get("filename"):
             self.logger.debug(
@@ -73,7 +78,6 @@ class Upload(Resource):
             )
         md5sum = request.headers.get("Content-MD5")
 
-        self.logger.debug("Receiving file: {}", filename)
         if not self.allowed_file(filename):
             self.logger.debug(
                 f"Tarfile upload: Bad file extension received for file {filename}"
@@ -117,16 +121,28 @@ class Upload(Resource):
         md5_full_path = Path(path, f"{filename}.md5")
         bytes_received = 0
 
-        # TODO: Need real user from PUT!
-
         # Create a tracking dataset object; it'll begin in UPLOADING state
         try:
             dataset = Dataset(
                 owner=username, controller=controller, path=tar_full_path, md5=md5sum
             )
             dataset.add()
+        except DatasetDuplicate:
+            self.logger.info("Dataset already exists {}", filename)
+            response = jsonify(dict(message="Dataset already exists"))
+            response.status_code = 200
+            return response
         except Exception:
             self.logger.exception("unable to create dataset for {}", filename)
+            abort(
+                HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR",
+            )
+
+        if tar_full_path.is_file() or md5_full_path.is_file():
+            self.logger.error(
+                "Dataset or corresponding md5 file already present on the disc, {}",
+                filename,
+            )
             abort(
                 HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR",
             )
--- a/lib/pbench/test/unit/server/test_requests.py
+++ b/lib/pbench/test/unit/server/test_requests.py
@@ -1,10 +1,32 @@
-import os
 import socket
-import pytest
-
 from pathlib import Path
+
+import pytest
 from werkzeug.utils import secure_filename
+
 from pbench.server.database.models.tracker import Dataset, States
+from pbench.test.unit.server.test_user_auth import login_user, register_user
+
+
+def get_pbench_token(client, server_config):
+    # First create a user
+    response = register_user(
+        client,
+        server_config,
+        username="user",
+        firstname="firstname",
+        lastname="lastName",
+        email="user@domain.com",
+        password="12345",
+    )
+    assert response.status_code, 201
+
+    # Login user to get valid pbench token
+    response = login_user(client, server_config, "user", "12345")
+    assert response.status_code == 200
+    data = response.json
+    assert data["auth_token"]
+    return data["auth_token"]
 
 
 class TestHostInfo:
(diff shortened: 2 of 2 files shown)


lib/pbench/server/api/resources/upload_api.py
```diff
@@ -48,10 +50,15 @@ def __init__(self, config, logger):
...
     def put(self, controller):
 
-        # FIXME: This should be assigned from the decoded authorization token
-        username = "pbench"
+        try:
+            user = User.query(id=Auth.token_auth.current_user().id)
+        except Exception:
+            self.logger.exception("Tarfile upload: Exception querying the user table")
+            abort(500, message="INTERNAL ERROR")
+        username = user.username
```
dbutenhof, line 61: I'm tempted to say we should take this opportunity to switch to registering the dataset under User.id here; however that'll break gold tests, and needs to be accompanied by an equivalent change to `pbench-state-manager`; so maybe it's better to defer that.

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -77,97 +100,130 @@ def test_json_object(client, caplog, server_config):
 
 class TestUpload:
     @staticmethod
-    def test_missing_filename_header_upload(client, caplog, server_config):
-        expected_message = (
-            "Missing filename header, "
-            "POST operation requires a filename header to name the uploaded file"
-        )
+    def test_missing_authorization_header(client, caplog, server_config):
```
dbutenhof, line 103: Should we throw in a bad authorization header, too? In theory that's tested elsewhere, and the test case you added ensures that PUT has the decorator, so that may be redundant. Are redundant security tests bad? ;-)

dbutenhof (approved):

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -48,10 +50,15 @@ def __init__(self, config, logger):
...
         )
 
+    @Auth.token_auth.login_required()
     def put(self, controller):
 
-        # FIXME: This should be assigned from the decoded authorization token
-        username = "pbench"
+        try:
+            user = User.query(id=Auth.token_auth.current_user().id)
+        except Exception:
```
portante, line 58: So line 53 already handles making a response with a 403 if the token is not valid, and a 401 if it is missing?

Do we always want to report a 500 error here?  Or are there conditions that are not considered an internal error that should be separated out?

portante (changes requested):

lib/pbench/server/api/resources/upload_api.py
npalaska: > So line 53 already handles making a response with a 403 if the token is not valid, and a 401 if it is missing?

Yes if the Authorization header doesn't pass we won't execute the PUT request. However, it returns 401 in both conditions.

> Do we always want to report a 500 error here? Or are there conditions that are not considered an internal error that should be separated out?

Now that you mentioned, I realized we need one more condition to check that is `if user is None`. Other than that if the exception does occur here while querying that means we could not connect to the db or db session is broken or something to do with sqlalchemy db operation and that would be an internal error

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -160,6 +160,26 @@ def test_bad_extension_upload(client, bad_extension, caplog, server_config):
...
 
+    @staticmethod
+    def test_invalid_authorization_upload(client, caplog, server_config):
+        with client:
+            auth_token = get_pbench_token(client, server_config)
+            # Log the token out
+            response = client.post(
+                f"{server_config.rest_uri}/logout",
+                headers=dict(Authorization="Bearer " + auth_token),
+            )
```
dbutenhof, line 171: Ha; I'd envisioned just making up a random "almost certainly bad" token, not actually logging in and then logging out again to get a guaranteed "good format but invalid" token.

I think _both_ cases are likely worth testing. Though maybe the other is more appropriate for your earlier basic authorization tests and may already be covered.

dbutenhof (dismissed):

lib/pbench/test/unit/server/test_requests.py
webbnh: We've got

- filename
  - missing
  - bad extension
- Auth token
  - missing
  - invalidated
- MD5 Hash
  - missing

It seems like there are at least two more cases that we could cover:
- a malformed Auth token
- a mismatched MD5 hash

webbnh (commented):
Looks good but for a few nits and possibly a couple of missing test cases.

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -77,97 +100,150 @@ def test_json_object(client, caplog, server_config):
...
-    def test_missing_filename_header_upload(client, caplog, server_config):
-        expected_message = (
-            "Missing filename header, "
-            "POST operation requires a filename header to name the uploaded file"
-        )
+    def test_missing_authorization_header(client, caplog, server_config):
         response = client.put(
-            f"{server_config.rest_uri}/upload/ctrl/{socket.gethostname()}"
+            f"{server_config.rest_uri}/upload/ctrl/{socket.gethostname()}",
+            headers={"filename": "f.tar.xz"},
```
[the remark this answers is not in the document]
npalaska: yeah `dict()` doesn't accept any key that has hyphen in it like `Content-MD5` thats why I used dictionary comprehension syntax, I guess I should use it everywhere instead of `Dict`

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -124,20 +121,32 @@ def put(self, controller):
...
         try:
             dataset = Dataset(
                 owner=username, controller=controller, path=tar_full_path, md5=md5sum
             )
             dataset.add()
+        except DatasetDuplicate:
+            self.logger.info("Dataset already exists {}", filename)
+            abort(
+                200, message="Dataset already exists",
+            )
```
dbutenhof, line 134: Hmm. An abort with an error message and success status is a bit weird. You're raising an HttpException which gets caught by Flask to cause the early return.

I think ultimately this does what we want, but I think that just doing a `return` here with the response we want would be cleaner and less likely to be messed up by something like a generic error handler later.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -124,20 +121,32 @@ def put(self, controller):
...
         except Exception:
             self.logger.exception("unable to create dataset for {}", filename)
             abort(
                 HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR",
             )
 
+        if tar_full_path.is_file() and md5_full_path.is_file():
+            self.logger.error(
+                "Dataset or corresponding md5 file already present on the disc, {}",
```
dbutenhof, line 143: What you've coded is "tarball __and__ MD5 file" not "or". I wonder why?

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -96,6 +95,18 @@ def test_missing_authorization_header(client, caplog, server_config):
             headers={"filename": "f.tar.xz"},
         )
         assert response.status_code == 401
+        for record in caplog.records:
+            assert record.levelname not in ("ERROR", "CRITICAL")
+
+    @staticmethod
+    def test_malformed_authorization_header(client, caplog, server_config):
```
dbutenhof, line 102: This is really "bad token" not "malformed"... the header format is correct. Though a "malformed" test, e.g., `Authorization: Bear cub` might not be a bad idea. too.

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -220,10 +259,10 @@ def test_upload(client, pytestconfig, caplog, server_config):
             assert response.status_code == 201, repr(response)
             sfilename = secure_filename(filename)
             tmp_d = pytestconfig.cache.get("TMP", None)
-            receive_dir = os.path.join(
+            receive_dir = Path(
                 tmp_d, "srv", "pbench", "pbench-move-results-receive", "fs-version-002"
             )
-            assert os.path.exists(receive_dir), (
+            assert (
```
dbutenhof, line 265: ?? What are you actually trying to assert here???

lib/pbench/server/api/resources/upload_api.py
npalaska: Ah yeah it should be `or`

lib/pbench/server/api/resources/upload_api.py
npalaska: Yes I also think return would be more appropriate choice here

lib/pbench/test/unit/server/test_requests.py
npalaska: I removed the check `os.path.exists` here as I am using python `Path` to create `receive_dir` and it already offers an existence check.

lib/pbench/test/unit/server/test_requests.py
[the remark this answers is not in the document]
webbnh: `dict()` will accept an "iterable" (or a mapping) in lieu of keyword arguments, and, apparently, you can even mix them, so
```
            headers=dict([("filename", "f.tar.xz")]),
```
or
```
            headers=dict((("filename", "f.tar.xz"),))
```
_would_ work here, but I find the comprehension syntax to be much more flexible and compact, so I would recommend it.

webbnh (approved):
Other than the one assertion we pointed out below, it looks good to me.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -48,10 +49,14 @@ def __init__(self, config, logger):
...
+    @Auth.token_auth.login_required()
     def put(self, controller):
 
-        # FIXME: This should be assigned from the decoded authorization token
-        username = "pbench"
+        try:
+            username = Auth.token_auth.current_user().username
+        except Exception:
+            self.logger.exception("Tarfile upload: Exception verifying the username")
+            abort(500, message="INTERNAL ERROR")
```
portante, line 59: You might as well consider using `HTTPStatus.INTERNAL_SERVER_ERROR` here if you come back to do any further edits.

lib/pbench/test/unit/server/test_requests.py
dbutenhof: My nit is bigger than your nit? Or something like that? Or whatever...

I suppose the point is that the name refers to the authorization header, not the token, while the test has a bad token but (from strict HTTP protocol point of view) the header is correctly formatted. Both cases are worth testing. Neither case is specific to the PUT API.

dbutenhof (approved):

portante (approved):
I am inclined to think this is ready for merging, despite some test feedback.  Perhaps we should track that feedback in issues to be addressed in the future?

lib/pbench/test/unit/server/test_requests.py
npalaska: @dbutenhof you mean testing with header other than a `Bearer` protocol, yeah these cases are tested under `test_user_auth.py` file

npalaska:
Sure I will open an issue to tack from Dave's comment

lib/pbench/test/unit/server/test_requests.py
portante: Tracked in #2246.

lib/pbench/test/unit/server/test_requests.py
portante: Tracked in #2246.

lib/pbench/test/unit/server/test_requests.py
portante: Tracked in #2246.

PR #2234, opened 2021-04-20
Enable minimum RPM version support

### LINKED ISSUE #2142: Need to allow pbench-fio to match newer versions rather than exact
Right now `pbench-fio` will only run if it detects `fio` version 3.21 is installed - any other version and `pbench-fio` exits with an error.  However, the pbench-agent can coexist with more than one fixed major/minor version and we'll likely run into situations where newer major (and possibly minor) versions of RHEL ship a newer `fio` than the exact match.  Most likely a simple `>=` rather than `=` test would be sufficient; hopefully there doesn't need to be a list of known-broken versions as that would be ... unfortunate.

portante:
Fixes #2142.

We need to support a minimum version of an RPM in addition to an exact version.  We use the `rpmdev-vercmp` from the `rpmdevtools` RPM in order to perform the matching.

----

This is a proper squash of the borked merge of PR #2210.


contrib/agent/bench-scripts/pbench-mpt
```diff
@@ -197,8 +197,8 @@ function mpt_install() {
   # TODO: since msg-perf-tool is installed via COPR, check what is the
   # recommended approach
 
-  check_enable_copr orpiske litestomp
-  check_enable_copr orpiske msg-perf-tool
+  #check_enable_copr orpiske litestomp
+  #check_enable_copr orpiske msg-perf-tool
 
 	if check_install_rpm $benchmark_rpm $ver; then
 		debug_log "[$script_name]$benchmark_rpm $ver is installed"
```
ndokos, line 204: (This refers to line 206 but github won't let me add it there) `error_log`? and wrong message.

agent/util-scripts/require-rpm
```diff
@@ -1,20 +1,88 @@
...
+# match parameter can be one of: 'any', 'equ', 'gtr', 'gte'.  The default value
+# for match is 'any' when no version is provided, and 'equ' when a version is
+# provided.
+#
+# If a match is found, emits the RPM EVR (epoch, version, release) found that
+# matches on stdout. If a match is not found, emits the RPM name and version
+# arguments provided on stdout.
+#
+# Exits with 0 (success) if a match is found, 1 if a match is not found, 2 if
+# an error occured during operation; 3 if the 3rd argument, match, is not
```
ndokos, line 18: "occurred"

ndokos (changes requested):
I guess I missed these in all the previous reviews :-(

ndokos (dismissed):

agent/util-scripts/test-bin/rpm
```diff
@@ -0,0 +1,51 @@
...
+
+args="${@}"
+echo "${0} ${args}" >> ${_testlog}
+
+# We are mocking out how the "rpm" command works by using certain names to
+# drive behaviors.  When not testing require-rpm itself, the special package
+# names are not used, and we assume success and exit with 0.  When the special
+# package names are used, we behave as described for each.
+
+_rpm="${3}"
```
webbnh, line 11: It seems like it would be worthwhile to sanity check that `${1}` is `--query` and that `${2}` is `--queryformat="%{EVR}"` (or similar).  It would serve to document what the first two arguments are, and might prevent confusion if this script gets invoked in something other than a query context.

agent/util-scripts/test-bin/test-require-rpm
```diff
@@ -0,0 +1,62 @@
...
+if [[ ${?} -ne 0 ]]; then
+    printf -- "test-require-rpm: failed to create temporary directory, %s\n" "${_tmp}" >&2
+    exit 1
+fi
+
+_prefix="Verify require-rpm with"
+function run_req_rpm {
+    _desc="${1}"
+    shift
+    printf -- "\n%s %s\n" "${_prefix}" "${_desc}"
```
webbnh, line 23: Just out of curiosity, why do you use `printf` here, instead of, say, `echo` like line 25, et al.?
```
    echo -e "\n${_prefix} ${_desc}\n"
```

agent/util-scripts/unittests
```diff
@@ -31,7 +31,7 @@ mv ${_testopt}/config/pbench-agent.cfg ${_testopt}/config/pbench-agent.cfg.orig
 let res=res+${?}
 cp -a ${_tdir}/../templates/* ${_testopt}/templates/
 let res=res+${?}
-cp -a ${_tdir}/{require-rpm,validate-*,pbench-*} ${_testopt}/util-scripts/
+cp -a ${_tdir}/{pbench-*,require-rpm,validate-*} ${_testopt}/util-scripts/
```
webbnh, line 34: Did you change anything here other than the order (which is not significant...right?)?

webbnh (approved):
Looks OK to merge...unless you want to fix any of the items below.

agent/util-scripts/test-bin/test-require-rpm
portante: When using `bash` I try to maintain a habit of not using `echo` and use `printf`.  In many cases it is no different, but it is easier to enhance the format string than to get the variable expansion correct as the complexity grows.

Or maybe because I really miss working with C (not joking, I always have a hard time with not knowing exactly what side effects statements have).

agent/util-scripts/unittests
portante: So the reason why this is changed is because of the iterative process used to develop these commits that occurred in parallel with PR #2212.

Initially I started out by mocking `require-rpm` for most tests, and then testing `require-rpm` directly, but then realized that I could just mock out `rpm` and make all the tests use `require-rpm`.  So before I rebased this PR (or really PR #2210) on PR #2212, when I removed `require-rpm` we just had `pbench-*` left over, so I drop the `{}` construct all together.

When I merged with PR #2212, I had to restore the `{}` construct, and then when I realized I only have to mock `rpm`, I added it back in sorted order.

portante:
Added @webbnh 's code review feedback as another commit to make it easier to review.

agent/util-scripts/require-rpm
```diff
@@ -1,20 +1,88 @@
...
+			fi
+		else
+			rc=3
+			echo "require-rpm: unrecognized 'match' argument, '${_match}'" >&2
+			break
+		fi
+	done
+	if [[ ${rc} -eq 1 ]]; then
+		echo "${_rpm}${_version}"
+	fi
```
webbnh, line 86: The indentation here seems to be using (hard) tabs; is that appropriate?  (I think we were using indents of for spaces.)

agent/util-scripts/test-bin/rpm
```diff
@@ -0,0 +1,60 @@
...
+# We are mocking out how the "rpm" command works by using certain names to
+# drive behaviors.  When not testing require-rpm itself, the special package
+# names are not used, and we assume success and exit with 0.  When the special
+# package names are used, we behave as described for each.
+
+if [[ "${1}" != "--query" ]]; then
+    printf -- "rpm: unexpected 1st argument, '%s'\n" "${1}" >&2
+    exit 101
+fi
+if [[ "${2}" != "--queryformat=%" ]]; then
```
webbnh, line 15: I hate to ask this, but does the comparison at line 15 actually work?  I was expecting `{EVR}` after the `%`, and I don't think `!=` does partial matches....

webbnh (changes requested):
Thanks for committing this last round separately (I wish I'd noticed before doing my review.... :-} ).

It looks like Mr. Jenkins is not happy with you (perhaps the problem is the one that I mention last, below).  The other two comments are your call.

agent/util-scripts/require-rpm
portante: Yes, the original code was 4 space indented.  I'll fix.

agent/util-scripts/test-bin/rpm
portante: I am glad you asked!  That has showed up a hole in our testing, where I did not use `require-rpm` directly with the `bench-scripts`, so we were not testing it properly.

portante: Filed as a separate commit with the fix.

agent/util-scripts/require-rpm
portante: Filed as a separate commits with the fix.

agent/tool-scripts/unittests
```diff
@@ -76,25 +76,21 @@ if [[ ${?} -ne 0 ]]; then
...
-    exit 1
-fi
-printf -- "#!/bin/bash\n\necho \"\$(basename \${0}) \${@}\" >> \${_testlog}\nexit 0\n" > ${_ts}/postprocess/mock-cmd
-chmod 775 ${_ts}/postprocess/mock-cmd
-if [[ ${?} -ne 0 ]]; then
-    echo "ERROR: failed to create ${_ts}/postprocess/mock-cmd" >&2
-    exit 1
-fi
+create_single_mock "echo \"\$(basename \${0}) \${@}\" >> \${_testlog}; sleep 0.1" ${_ts}/datalog/mock-cmd
+create_single_mock "echo \"\$(basename \${0}) \${@}\" >> \${_testlog}" ${_ts}/postprocess/mock-cmd
```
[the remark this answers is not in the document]
portante: Filed as a separate commit with the fix.

dbutenhof (approved):

agent/tool-scripts/unittests
```diff
@@ -76,26 +76,21 @@ if [[ ${?} -ne 0 ]]; then
...
+    contents=${1}
+    target=${2}
+    printf -- "#!/bin/bash\n\n%s\nexit 0\n" "${contents}" > ${target} && chmod 775 ${target}
+    if [[ ${?} -ne 0 ]]; then
+        printf -- "ERROR: failed to create %s\n" "${target}" >&2
+        exit 1
+    fi
+}
+
+create_single_mock 'echo "${1}${2:+-${2}}"' ${_optdir}/util-scripts/require-rpm
```
webbnh, line 90: This is very nice, now.

agent/bench-scripts/tests/pbench-fio/test-03.txt
```diff
@@ -15,5 +15,5 @@
 [debug][1900-01-01T00:00:00.000000] [pbench-fio] fio-3.21 is installed
 --- pbench.log file contents
 +++ test-execution.log file contents
-/var/tmp/pbench-test-bench/opt/pbench-agent/unittest-scripts/require-rpm fio 3.21 gte
+/var/tmp/pbench-test-bench/opt/pbench-agent/unittest-scripts/rpm --query --queryformat=%{EVR} fio
```
webbnh, line 18: This log now shows the mock for `rpm` instead of the mock for `require-rpm` because the test is now using the "real" `require-rpm` instead of the mock'd one and that invokes the mock'd `rpm`?

webbnh (approved):
Looks good.  My one substantial comment would probably best as a follow-on PR.

agent/util-scripts/test-bin/rpm
```diff
@@ -8,6 +8,15 @@ echo "${0} ${args}" >> ${_testlog}
...
 # package names are used, we behave as described for each.
 
+if [[ "${1}" != "--query" ]]; then
+    printf -- "rpm: unexpected 1st argument, '%s'\n" "${1}" >&2
+    exit 101
+fi
+if [[ "${2}" != "--queryformat=%{EVR}" ]]; then
+    printf -- "rpm: unexpected 2nd argument, '%s'\n" "${2}" >&2
+    exit 101
+fi
```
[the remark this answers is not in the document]
portante: Tracked in #2239.

PR #2236, opened 2021-04-21
Allow pbench-server installation directory to be overridden

ndokos:
Define it as a default value for the ansible roles used during server
installation, but allow it to be overridden either on the command
line (with --extra-vars) or from an inventory file.


portante (dismissed):

portante:
@ndokos, shouldn't this PR just be closed now that the server side ansible roles and playbooks have been removed?

portante:
Since the Pbench Server is now containerized this work is no longer needed.

Issue #2245, opened 2021-04-26
The pbench-agent fails to install on RHEL 7/CentOS 7


portante:
The pbench-agent no longer installs on RHEL 7 / CentOS 7 due to the following errors:

```
# yum install pbench-agent
Loaded plugins: langpacks, product-id, search-disabled-repos, subscription-manager
Resolving Dependencies
--> Running transaction check
---> Package pbench-agent.noarch 0:0.71.0-158g2624b973d will be installed
--> Finished Dependency Resolution

Dependencies Resolved

===========================================================================================================================================================================
 Package                              Arch                           Version                                       Repository                                         Size
===========================================================================================================================================================================
Installing:
 pbench-agent                         noarch                         0.71.0-158g2624b973d                          copr-pbench-test-portante                         354 k

Transaction Summary
===========================================================================================================================================================================
Install  1 Package

Total download size: 354 k
Installed size: 1.4 M
Is this ok [y/d/N]: y
Downloading packages:
pbench-agent-0.71.0-158g2624b973d.noarch.rpm                                                                                                        | 354 kB  00:00:00
Running transaction check
Running transaction test
Transaction test succeeded
Running transaction
  Installing : pbench-agent-0.71.0-158g2624b973d.noarch                                                                                                                1/1
WARNING: Running pip install with root privileges is generally not a good idea. Try `pip3 install --user` instead.
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-xzhn6hb8/cryptography/
  Verifying  : pbench-agent-0.71.0-158g2624b973d.noarch                                                                                                                1/1

Installed:
  pbench-agent.noarch 0:0.71.0-158g2624b973d

Complete!
```

The `WARNING` above is not a problem, but the `failed with error code 1` line is a problem, and should have killed the installation.

You can see the full error reported in `/opt/pbench-agent/pip3-install.log`:

```
Collecting cryptography (from ansible-base<2.11,>=2.10.8->ansible->-r /opt/pbench-agent/requirements.txt (line 1))
  Downloading https://files.pythonhosted.org/packages/9b/77/461087a514d2e8ece1c975d8216bc03f7048e6090c5166bc34115afdaa53/cryptography-3.4.7.tar.gz (546kB)
    Complete output from command python setup.py egg_info:

            =============================DEBUG ASSISTANCE==========================
            If you are seeing an error here please try the following to
            successfully install cryptography:

[...]

PR #2247, opened 2021-04-26
Update PCP repo for container builds

portante:
For some reason the bintray.com repos have stopped working, giving us '401 Forbidden' errors.  It appears the new repos are using the jFrog Artifactory instance for PCP.

We also drop the 5.2.2 version requirement for PCP.

----

This PR is based on PR #2248 in order to make both testable.

PR #2248, opened 2021-04-26
Fix CentOS 7 / RHEL 7 pbench-agent installations

portante:
It looks like we need the latest version of `pip` to install all of the module dependencies we have.  For all but CentOS 7 / RHEL 7 we have the correct version.  So for those environments we now have to install the latest version of `pip3` directly in order for us to have our Python3 dependencies satisfied.

The days of our RHEL 7 support are numbered.

Fixes https://github.com/distributed-system-analysis/pbench/issues/2245.

### DIFF
--- a/agent/rpm/pbench-agent.spec.j2
+++ b/agent/rpm/pbench-agent.spec.j2
@@ -76,6 +76,21 @@ cp -a agent/* %{?buildroot}/%{installdir}/
 if pip3 show configtools > /dev/null 2>&1 ;then pip3 uninstall -y configtools ;fi
 
 %post
+%if 0%{?rhel} == 7
+# This is a very ugly work-around for RHEL 7 / CentOS 7 environments. The
+# version of `pip3` that is available in those environments no longer works
+# to install the latest Python3 modules, which we need. So we have to fetch
+# our own copy of `pip3`, as described at:
+#   https://pip.pypa.io/en/latest/installing/#installing-with-get-pip-py 
+# We then install it separately from the existing verion and set `PATH` and
+# `PYTHONPATH` so that we use this updated version when we install our module
+# dependencies below.
+curl -X GET -o /%{installdir}/get-pip.py https://bootstrap.pypa.io/get-pip.py > /%{installdir}/pip3-update.log 2>&1
+python3 /%{installdir}/get-pip.py --prefix=/%{installdir} >> /%{installdir}/pip3-update.log 2>&1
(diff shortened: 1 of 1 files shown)


dbutenhof:
> The days of RHEL 7 are numbered.

"Numbered"? Or "few"? Because when we're talking about major OS releases, it seems that "numbered" is usually more of a "countable infinity" thing rather than "2 or 3". ;-)

agent/rpm/pbench-agent.spec.j2
```diff
@@ -76,6 +76,12 @@ cp -a agent/* %{?buildroot}/%{installdir}/
 if pip3 show configtools > /dev/null 2>&1 ;then pip3 uninstall -y configtools ;fi
 
 %post
+%if 0%{?rhel} == 7
+curl -X GET -o /%{installdir}/get-pip.py https://bootstrap.pypa.io/get-pip.py > /%{installdir}/pip3-update.log 2>&1
+python3 /%{installdir}/get-pip.py --prefix=/%{installdir} >> /%{installdir}/pip3-update.log 2>&1
+export PYTHONPATH=/%{installdir}/lib/python3.6/site-packages:${PYTHONPATH}
+export PATH=/%{installdir}/bin:${PATH}
+%endif
```
dbutenhof, line 84: This seems to be the same as in #2247; I assume your intention is to get this in first, but the duplication still seems a bit odd. This all looks straightforward, but the origin of this `get-pip.py` seems fairly obscure.

Would you like to add a comment referencing https://pip.pypa.io/en/latest/installing/#installing-with-get-pip-py to provide context? (And assurance that this is "official"?)

dbutenhof (approved):

agent/rpm/pbench-agent.spec.j2
portante: > This seems to be the same as in #2247; I assume your intention is to get this in first, but the duplication still seems a bit odd.

This is intentional.  I need both fixes in order to actually build and test things are working, but didn't want to lump them into the same PR since I'd like to expose them for consideration separately.

> This all looks straightforward, but the origin of this `get-pip.py` seems fairly obscure.
> 
> Would you like to add a comment referencing https://pip.pypa.io/en/latest/installing/#installing-with-get-pip-py to provide context? (And assurance that this is "official"?)

Sure, I've updated the comment.

dbutenhof (approved):

Maxusmusti (approved):
Tested on RHEL 7 (for controller and remote hosts), working as expected

PR #2251, opened 2021-04-27
Add Authorization header requirement on all the es APIs

npalaska:
Adds a strict Bearer Authorization requirement on all the elasticsearch Post/Get APIs

### DIFF
--- a/lib/pbench/server/api/auth.py
+++ b/lib/pbench/server/api/auth.py
@@ -79,6 +79,27 @@ class Auth:
         except Exception as e:
             Auth.logger.exception(f"{__name__}: ERROR: {e.__traceback__}")
 
+    def verify_user(self, target_username: str) -> User:
+        """
+        Check whether the requested target user is the owner of the authorization
+        token provided to the API.
+
+        We're returning a User instance corresponding to the target username provided.
+        If the request is not authenticated (current user is None) we return None as
+        we won't be able to verify the target user. However, if the current user is
+        an admin, we return the instance of the target user by querying the target
+        username provided.
+        """
+        current_user = Auth.token_auth.current_user()
+        if not current_user:
+            return None
+        if current_user.username == target_username:
+            return current_user
+        if current_user.is_admin():
+            target_user = User.query(username=target_username)
+            return target_user
+        return None
+
     def get_auth_token(self, logger):
         # get auth token
         auth_header = request.headers.get("Authorization")
--- a/lib/pbench/server/api/resources/query_apis/__init__.py
+++ b/lib/pbench/server/api/resources/query_apis/__init__.py
@@ -22,7 +22,21 @@ class SchemaError(TypeError):
     Generic base class for errors in processing a JSON schema.
     """
 
-    pass
+    def __init__(self, status: int = 400):
+        self.http_status = status
+
+
+class UnverifiedUser(SchemaError):
+    """
+    Unverified attempt to access other user data.
+    """
+
+    def __init__(self, username: str):
+        self.username = username
+        super().__init__(status=401)
+
+    def __str__(self):
+        return f"{self.username} can not be verified"
 
 
 class InvalidRequestPayload(SchemaError):
@@ -41,6 +55,7 @@ class MissingParameters(SchemaError):
     """
 
     def __init__(self, keys: List[AnyStr]):
+        super().__init__()
         self.keys = keys
 
     def __str__(self):
@@ -61,6 +76,7 @@ class ConversionError(SchemaError):
             expected_type: The expected type
             actual_type: The actual type
         """
+        super().__init__()
         self.value = value
         self.expected_type = expected_type
         self.actual_type = actual_type
@@ -115,9 +131,12 @@ def convert_username(value: str) -> str:
         internal username representation
     """
     try:
-        return Auth.validate_user(value)
+        user = Auth().verify_user(value)
     except Exception:
         raise ConversionError(value, "username", type(value).__name__)
+    if not user:
+        raise UnverifiedUser(value)
+    return str(user.id)
 
 
 def convert_json(value: dict) -> dict:
@@ -499,6 +518,7 @@ class ElasticBase(Resource):
             )
             abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
+    @Auth.token_auth.login_required(optional=True)
     def post(self):
         """
         Handle a Pbench server POST operation that will involve a call to the
@@ -512,10 +532,26 @@ class ElasticBase(Resource):
         validation/conversion where necessary. Missing "required" parameters
         and any parameter with a null value are rejected. Parameters that
         are not in the class schema are ignored but logged.
+
+        If the request does not contain the user field, it will be interpreted
+        as a public dataset query.
         """
         json_data = request.get_json(silent=True)
+        if json_data and "user" in json_data:
+            if not json_data["user"]:
+                # There is an empty user field present in the json data;
+                # therefore, this request will be interpreted as a
+                # public dataset query.  Remove it before validating the
+                # remaining json data.
+                del json_data["user"]
+            elif Auth.token_auth.current_user() is None:
+                # Only logged-in users are allowed to query non-public data.
+                abort(HTTPStatus.FORBIDDEN, message="Not Authorized")
         try:
             new_data = self.schema.validate(json_data)
+        except UnverifiedUser as e:
+            self.logger.warning("{}", str(e))
+            abort(HTTPStatus.FORBIDDEN, message=str(e))
         except SchemaError as e:
             self.logger.warning(
                 "{}: {} on {!r}", self.__class__.__name__, str(e), json_data
@@ -523,6 +559,7 @@ class ElasticBase(Resource):
             abort(HTTPStatus.BAD_REQUEST, message=str(e))
         return self._call(requests.post, new_data)
 
+    @Auth.token_auth.login_required(optional=True)
     def get(self):
         """
         Handle a GET operation involving a call to the server's Elasticsearch
--- a/lib/pbench/server/api/resources/query_apis/controllers_list.py
+++ b/lib/pbench/server/api/resources/query_apis/controllers_list.py
@@ -22,7 +22,7 @@ class ControllersList(ElasticBase):
             config,
             logger,
             Schema(
-                Parameter("user", ParamType.USER, required=True),
+                Parameter("user", ParamType.USER, required=False),
                 Parameter("start", ParamType.DATE, required=True),
                 Parameter("end", ParamType.DATE, required=True),
             ),
@@ -58,9 +58,9 @@ class ControllersList(ElasticBase):
             "start" and "end" are datetime objects representing a set of Elasticsearch
                 run document indices in which to search.
         """
-        user = json_data["user"]
-        start = json_data["start"]
-        end = json_data["end"]
+        user = json_data.get("user")
+        start = json_data.get("start")
+        end = json_data.get("end")
 
         # We need to pass string dates as part of the Elasticsearch query; we
         # use the unconverted strings passed by the caller rather than the
--- a/lib/pbench/test/unit/server/conftest.py
+++ b/lib/pbench/test/unit/server/conftest.py
@@ -1,4 +1,5 @@
 import datetime
+from http import HTTPStatus
 import os
 import pytest
 import shutil
@@ -8,6 +9,7 @@ from posix import stat_result
 from stat import ST_MTIME
 
 from pbench.server.api import create_app, get_server_config
+from pbench.test.unit.server.test_user_auth import login_user, register_user
 from pbench.server.api.auth import Auth
 from pbench.server.database.database import Database
 from pbench.server.database.models.template import Template
@@ -221,3 +223,42 @@ def find_template(monkeypatch, fake_mtime):
     with monkeypatch.context() as m:
         m.setattr(Template, "find", fake_find)
         yield
+
+
+@pytest.fixture
+def pbench_token(client, server_config):
+    # First create a user
+    response = register_user(
+        client,
+        server_config,
+        username="drb",
+        firstname="firstname",
+        lastname="lastName",
+        email="user@domain.com",
+        password="12345",
+    )
+    assert response.status_code == HTTPStatus.CREATED
+
+    # Login user to get valid pbench token
+    response = login_user(client, server_config, "drb", "12345")
+    assert response.status_code == HTTPStatus.OK
+    data = response.json
+    assert data["auth_token"]
+    return data["auth_token"]
+
+
+@pytest.fixture(params=("valid", "invalid", "empty"))
+def build_auth_header(request, server_config, pbench_token, client):
+    header = (
+        {} if request.param == "empty" else {"Authorization": "Bearer " + pbench_token}
+    )
+
+    if request.param == "invalid":
+        # Create an invalid token by logging the user out
+        response = client.post(
+            f"{server_config.rest_uri}/logout",
+            headers=dict(Authorization="Bearer " + pbench_token),
+        )
+        assert response.status_code == HTTPStatus.OK
+
+    return {"header": header, "header_param": request.param}
--- a/lib/pbench/test/unit/server/query_apis/conftest.py
+++ b/lib/pbench/test/unit/server/query_apis/conftest.py
@@ -1,3 +1,4 @@
+from http import HTTPStatus
 from typing import Any, Dict
 import pytest
 import requests
@@ -6,7 +7,7 @@ import responses
 
 @pytest.fixture
 @responses.activate
-def query_api(client, server_config):
+def query_api(client, server_config, pbench_token):
     """
     Help controller queries that want to interact with a mocked
     Elasticsearch service.
@@ -29,17 +30,20 @@ def query_api(client, server_config):
         payload: Dict[str, Any],
         expected_index: str,
         expected_status: str,
+        headers: dict = {},
         **kwargs,
     ) -> requests.Response:
         host = server_config.get("elasticsearch", "host")
         port = server_config.get("elasticsearch", "port")
         es_url = f"http://{host}:{port}{expected_index}{es_uri}"
         with responses.RequestsMock() as rsp:
-            rsp.add(responses.POST, es_url, **kwargs)
+            # We need to set up mocks for the Server's call to Elasticsearch,
+            # which will only be made if the request is not forbidden nor unauthorized.
+            if expected_status not in [HTTPStatus.FORBIDDEN, HTTPStatus.UNAUTHORIZED]:
+                rsp.add(responses.POST, es_url, **kwargs)
             response = client.post(
-                f"{server_config.rest_uri}{pbench_uri}", json=payload
+                f"{server_config.rest_uri}{pbench_uri}", headers=headers, json=payload,
             )
-            assert len(rsp.calls) == 1
         assert response.status_code == expected_status
         return response
 
--- a/lib/pbench/test/unit/server/query_apis/test_controllers_list.py
+++ b/lib/pbench/test/unit/server/query_apis/test_controllers_list.py
@@ -14,7 +14,7 @@ class TestControllersList:
 
     def build_index(self, server_config, dates):
         """
-        build_index Build the index list for query
+        Build the index list for query
 
         Args:
             dates (iterable): list of date strings
@@ -25,11 +25,49 @@ class TestControllersList:
             index += f"{idx}{d},"
         return index
 
-    def test_missing_json_object(self, client, server_config):
+    def test_non_accessible_user_data(self, client, server_config, pbench_token):
         """
-        test_missing_json_object Test behavior when no JSON payload is given
+        Test behavior when Authorization header does not have access to other user's data
         """
-        response = client.post(f"{server_config.rest_uri}/controllers/list")
+        # The pbench_token fixture logs in as user "drb"
+        # Trying to access the data belong to the user "pp"
+        response = client.post(
+            f"{server_config.rest_uri}/controllers/list",
+            headers={"Authorization": "Bearer " + pbench_token},
+            json={"user": "pp", "start": "2020-08", "end": "2020-10"},
+        )
+        assert response.status_code == HTTPStatus.FORBIDDEN
+
+    @pytest.mark.parametrize(
+        "user", ("drb", "pp"),
+    )
+    def test_accessing_user_data_with_invalid_token(
+        self, client, server_config, pbench_token, user
+    ):
+        """
+        Test behavior when Authorization header does not have access to other user's data
+        """
+        # valid token logout
+        response = client.post(
+            f"{server_config.rest_uri}/logout",
+            headers=dict(Authorization="Bearer " + pbench_token),
+        )
+        assert response.status_code == HTTPStatus.OK
+        response = client.post(
+            f"{server_config.rest_uri}/controllers/list",
+            headers={"Authorization": "Bearer " + pbench_token},
+            json={"user": user, "start": "2020-08", "end": "2020-10"},
+        )
+        assert response.status_code == HTTPStatus.FORBIDDEN
+
+    def test_missing_json_object(self, client, server_config, pbench_token):
+        """
+        Test behavior when no JSON payload is given
+        """
+        response = client.post(
+            f"{server_config.rest_uri}/controllers/list",
+            headers={"Authorization": "Bearer " + pbench_token},
+        )
         assert response.status_code == HTTPStatus.BAD_REQUEST
         assert response.json.get("message") == "Invalid request payload"
 
@@ -41,32 +79,36 @@ class TestControllersList:
             {"end": "2020"},
             {"user": "x", "start": "2020"},
             {"user": "x", "end": "2020"},
-            {"start": "2020", "end": "2020"},
+            {"some_additional_key": "test"},
         ),
     )
-    def test_missing_keys(self, client, server_config, keys, user_ok):
+    def test_missing_keys(self, client, server_config, keys, user_ok, pbench_token):
         """
-        test_missing_keys Test behavior when JSON payload does not contain
-        all required keys.
+        Test behavior when JSON payload does not contain all required keys.
 
-        Note that "user", "start", and "end" are all required;
+        Note that "start", and "end" are required whereas "user" is not mandatory;
         however, Pbench will silently ignore any additional keys that are
         specified.
        """
-        response = client.post(f"{server_config.rest_uri}/controllers/list", json=keys)
+        response = client.post(
+            f"{server_config.rest_uri}/controllers/list",
+            headers={"Authorization": "Bearer " + pbench_token},
+            json=keys,
+        )
         assert response.status_code == HTTPStatus.BAD_REQUEST
-        missing = [k for k in ("user", "start", "end") if k not in keys]
+        missing = [k for k in ("start", "end") if k not in keys]
         assert (
             response.json.get("message")
             == f"Missing required parameters: {','.join(missing)}"
         )
 
-    def test_bad_dates(self, client, server_config, user_ok):
+    def test_bad_dates(self, client, server_config, user_ok, pbench_token):
         """
-        test_bad_dates Test behavior when a bad date string is given
+        Test behavior when a bad date string is given
         """
         response = client.post(
             f"{server_config.rest_uri}/controllers/list",
+            headers={"Authorization": "Bearer " + pbench_token},
             json={"user": "drb", "start": "2020-12", "end": "2020-19"},
         )
         assert response.status_code == HTTPStatus.BAD_REQUEST
@@ -75,16 +117,32 @@ class TestControllersList:
             == "Value '2020-19' (str) cannot be parsed as a date/time string"
         )
 
-    def test_query(self, server_config, query_api, user_ok, find_template):
+    @pytest.mark.parametrize(
+        "user", ("drb", "", "no_user", None),
+    )
+    def test_query(
+        self,
+        client,
+        server_config,
+        query_api,
+        user_ok,
+        find_template,
+        build_auth_header,
+        user,
+    ):
         """
-        test_query Check the construction of Elasticsearch query URI
-        and filtering of the response body.
+        Check the construction of Elasticsearch query URI and filtering of the response body.
+        The test will run once with each parameter supplied from the local parameterization,
+        and, for each of those, three times with different values of the build_auth_header fixture.
         """
         json = {
-            "user": "drb",
+            "user": user,
             "start": "2020-08",
             "end": "2020-10",
         }
+        if user == "no_user":
+            json.pop("user", None)
+
         response_payload = {
             "took": 1,
             "timed_out": False,
@@ -121,34 +179,57 @@ class TestControllersList:
         }
 
         index = self.build_index(server_config, ("2020-08", "2020-09", "2020-10"))
+        # If we're not asking about a particular user, or if the user
+        # field is to be omitted altogether, or if we have a valid
+        # token, then the request should succeed.
+        if (
+            not user
+            or user == "no_user"
+            or build_auth_header["header_param"] == "valid"
+        ):
+            expected_status = HTTPStatus.OK
+        else:
+            expected_status = HTTPStatus.FORBIDDEN
+
         response = query_api(
             "/controllers/list",
             "/_search?ignore_unavailable=true",
             json,
             index,
-            HTTPStatus.OK,
-            status=HTTPStatus.OK,
+            expected_status,
+            headers=build_auth_header["header"],
+            status=expected_status,
             json=response_payload,
         )
-        res_json = response.json
-        assert isinstance(res_json, list)
-        assert len(res_json) == 2
-        assert res_json[0]["key"] == "unittest-controller1"
-        assert res_json[0]["controller"] == "unittest-controller1"
-        assert res_json[0]["results"] == 2
-        assert res_json[0]["last_modified_value"] == 1.59847315581e12
-        assert res_json[0]["last_modified_string"] == "2020-08-26T20:19:15.810Z"
-        assert res_json[1]["key"] == "unittest-controller2"
-        assert res_json[1]["controller"] == "unittest-controller2"
-        assert res_json[1]["results"] == 1
-        assert res_json[1]["last_modified_value"] == 1.6
-        assert res_json[1]["last_modified_string"] == "2020-09-26T20:19:15.810Z"
+        assert response.status_code == expected_status
+        if response.status_code == HTTPStatus.OK:
+            res_json = response.json
+            assert isinstance(res_json, list)
+            assert len(res_json) == 2
+            assert res_json[0]["key"] == "unittest-controller1"
+            assert res_json[0]["controller"] == "unittest-controller1"
+            assert res_json[0]["results"] == 2
+            assert res_json[0]["last_modified_value"] == 1.59847315581e12
+            assert res_json[0]["last_modified_string"] == "2020-08-26T20:19:15.810Z"
+            assert res_json[1]["key"] == "unittest-controller2"
+            assert res_json[1]["controller"] == "unittest-controller2"
+            assert res_json[1]["results"] == 1
+            assert res_json[1]["last_modified_value"] == 1.6
+            assert res_json[1]["last_modified_string"] == "2020-09-26T20:19:15.810Z"
 
     def test_empty_query(
-        self, client, server_config, query_api, user_ok, find_template
+        self,
+        client,
+        server_config,
+        query_api,
+        user_ok,
+        find_template,
+        build_auth_header,
     ):
         """
         Check proper handling of a query resulting in no Elasticsearch matches.
+        The test will run thrice with different values of the build_auth_header
+        fixture.
         """
         json = {
             "user": "drb",
@@ -166,17 +247,24 @@ class TestControllersList:
             },
         }
 
+        expected_status = HTTPStatus.OK
+        if build_auth_header["header_param"] != "valid":
+            expected_status = HTTPStatus.FORBIDDEN
+
         index = self.build_index(server_config, ("2020-08", "2020-09", "2020-10"))
         response = query_api(
             "/controllers/list",
             "/_search?ignore_unavailable=true",
             json,
             index,
-            HTTPStatus.OK,
+            expected_status,
+            headers=build_auth_header["header"],
             status=HTTPStatus.OK,
             json=response_payload,
         )
-        assert response.json == []
+        assert response.status_code == expected_status
+        if response.status_code == HTTPStatus.OK:
+            assert response.json == []
 
     @pytest.mark.parametrize(
         "exceptions",
@@ -204,7 +292,7 @@ class TestControllersList:
         Check that an exception in calling Elasticsearch is reported correctly.
         """
         json = {
-            "user": "drb",
+            "user": "",
             "start": "2020-08",
             "end": "2020-08",
         }
@@ -225,7 +313,7 @@ class TestControllersList:
         response.raise_for_status() and Pbench handlers.
         """
         json = {
-            "user": "drb",
+            "user": "",
             "start": "2020-08",
             "end": "2020-08",
         }
--- a/lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
+++ b/lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
@@ -14,7 +14,7 @@ class TestDatasetsDetail:
 
     def build_index(self, server_config, dates):
         """
-        build_index Build the index list for query
+        Build the index list for query
 
         Args:
             dates (iterable): list of date strings
@@ -25,11 +25,54 @@ class TestDatasetsDetail:
             index += f"{idx}{d},"
         return index
 
-    def test_missing_json_object(self, client, server_config):
+    def test_non_accessible_user_data(self, client, server_config, pbench_token):
         """
-        test_missing_json_object Test behavior when no JSON payload is given
+        Test behavior when Authorization header does not have access to other user's data
         """
-        response = client.post(f"{server_config.rest_uri}/datasets/detail")
+        # The pbench_token fixture logs in as user "drb"
+        # Trying to access the data belong to the user "pp"
+        response = client.post(
+            f"{server_config.rest_uri}/datasets/detail",
+            headers={"Authorization": "Bearer " + pbench_token},
+            json={
+                "user": "pp",
+                "name": "dataset_name",
+                "start": "2020-08",
+                "end": "2020-10",
+            },
+        )
+        assert response.status_code == HTTPStatus.FORBIDDEN
+
+    @pytest.mark.parametrize(
+        "user", ("drb", "pp"),
+    )
+    def test_accessing_data_with_invalid_token(
+        self, client, server_config, pbench_token, user
+    ):
+        """
+        Test behavior when Authorization header does not have access to other user's data
+        """
+        # valid token logout
+        response = client.post(
+            f"{server_config.rest_uri}/logout",
+            headers=dict(Authorization="Bearer " + pbench_token),
+        )
+        assert response.status_code == HTTPStatus.OK
+        response = client.post(
+            f"{server_config.rest_uri}/datasets/detail",
+            headers={"Authorization": "Bearer " + pbench_token},
+            json={"user": user, "start": "2020-08", "end": "2020-10"},
+        )
+        assert response.status_code == HTTPStatus.FORBIDDEN
+
+    def test_missing_json_object(self, client, server_config, pbench_token):
+        """
+        Test behavior when no JSON payload is given
+        """
+        response = client.post(
+            f"{server_config.rest_uri}/datasets/detail",
+            headers={"Authorization": "Bearer " + pbench_token},
+        )
         assert response.status_code == HTTPStatus.BAD_REQUEST
         assert response.json.get("message") == "Invalid request payload"
 
@@ -43,32 +86,37 @@ class TestDatasetsDetail:
             {"user": "x", "start": "2020"},
             {"user": "x", "end": "2020"},
             {"user": "x", "name": "y", "start": "2021"},
-            {"start": "2020", "end": "2020"},
+            {"some_additional_key": "test"},
         ),
     )
-    def test_missing_keys(self, client, server_config, keys, user_ok):
+    def test_missing_keys(self, client, server_config, keys, user_ok, pbench_token):
         """
-        test_missing_keys Test behavior when JSON payload does not contain
+        Test behavior when JSON payload does not contain
         all required keys.
 
-        Note that "user", "name", "start", and "end" are all required;
+        Note that "name", "start", and "end" are required whereas "user" is not mandatory;
         however, Pbench will silently ignore any additional keys that are
         specified.
        """
-        response = client.post(f"{server_config.rest_uri}/datasets/detail", json=keys)
+        response = client.post(
+            f"{server_config.rest_uri}/datasets/detail",
+            headers={"Authorization": "Bearer " + pbench_token},
+            json=keys,
+        )
         assert response.status_code == HTTPStatus.BAD_REQUEST
-        missing = [k for k in ("user", "name", "start", "end") if k not in keys]
+        missing = [k for k in ("name", "start", "end") if k not in keys]
         assert (
             response.json.get("message")
             == f"Missing required parameters: {','.join(missing)}"
         )
 
-    def test_bad_dates(self, client, server_config, user_ok):
+    def test_bad_dates(self, client, server_config, user_ok, pbench_token):
         """
-        test_bad_dates Test behavior when a bad date string is given
+        Test behavior when a bad date string is given
         """
         response = client.post(
             f"{server_config.rest_uri}/datasets/detail",
+            headers={"Authorization": "Bearer " + pbench_token},
             json={
                 "user": "drb",
                 "name": "footest",
@@ -82,17 +130,33 @@ class TestDatasetsDetail:
             == "Value '2020-19' (str) cannot be parsed as a date/time string"
         )
 
-    def test_query(self, server_config, query_api, user_ok, find_template):
+    @pytest.mark.parametrize(
+        "user", ("drb", "", "no_user", None),
+    )
+    def test_query(
+        self,
+        client,
+        server_config,
+        query_api,
+        user_ok,
+        find_template,
+        build_auth_header,
+        user,
+    ):
         """
-        test_query Check the construction of Elasticsearch query URI
-        and filtering of the response body.
+        Check the construction of Elasticsearch query URI and filtering of the response body.
+        The test will run once with each parameter supplied from the local parameterization,
+        and, for each of those, three times with different values of the build_auth_header fixture.
         """
         json = {
-            "user": "drb",
+            "user": user,
             "name": "fio_rhel8_kvm_perf43_preallocfull_nvme_run4_iothread_isolcpus_2020.04.29T12.49.13",
             "start": "2020-08",
             "end": "2020-10",
         }
+        if user == "no_user":
+            json.pop("user", None)
+
         response_payload = {
             "took": 112,
             "timed_out": False,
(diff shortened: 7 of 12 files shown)


lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -493,8 +502,12 @@ def post(self):
             # be interpreted as formatting commands.
             self.logger.warning("{}", str(e))
             abort(400, message=str(e))
+        except UnVerifiedUser as e:
+            self.logger.warning("{}", str(e))
+            abort(401, message=str(e))
         return self._call(requests.post, new_data)
 
+    @Auth.token_auth.login_required()
```
dbutenhof, line 510: Uh oh. `/controllers/months` wasn't intended to be authenticated. There's nothing user-specific here, and it needs to be able to return the range of index months for unauthenticated calls (public data).

However, this raises an interesting question, because perhaps it _should_ be authenticated rather than allowing just anyone to determine the full range of in-use run index months? E.g., maybe instead of just looking for all the indices and filtering out the date-stamped run indices with the right prefix (as it currently does), we _could_ run a more complex query to find all run indices which have run documents owned by the client or all run indices which have public documents. This, for example, wouldn't give a recent user a range of years in which they wouldn't possibly own any data. Hmm.

However, with the change to allow for an unauthorized query (for public data), let's leave it this way. I'll pull out the comment about user-constrained month ranges (which I think is a good idea anyway) into an issue and we'll worry about that aspect later.

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -470,6 +478,7 @@ def _call(self, method: Callable, json_data: Dict[AnyStr, Any]):
             )
             abort(500, message="INTERNAL ERROR")
 
+    @Auth.token_auth.login_required()
```
dbutenhof, line 481: Authorization has to be _optional_ because a client without a login needs to be able to query for public data.

You could further finesse this by requiring `"user": None` in this case, and defaulting a missing `"user"` parameter in the JSON to `None`. (Note, that's why I referenced the subclass `assembly` methods in the email.)

This goes for the `get` as well.

dbutenhof (changes requested):
This would probably be good enough for the jam session, as it lets us do authenticated queries under a registered user. (And we don't yet have a way to actually "publish" public data anyway.)

However for this to be merged, we need to allow queries from unauthenticated clients, so the authorization can't be required.

lib/pbench/server/api/resources/query_apis/__init__.py
npalaska: Yeah I will make the authentication optional

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -104,12 +104,20 @@ def convert_username(value: str) -> str:
...
     """
     try:
-        return Auth.validate_user(value)
+        user, verified = Auth().verify_user(value)
+        if verified or user.is_admin():
+            return value
+        else:
+            raise UnVerifiedUser(value)
+    except UnVerifiedUser:
+        raise
```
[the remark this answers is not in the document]
dbutenhof: Yeah, I failed to notice that you're returning 401, and you're right that's appropriate.

So a separate handler may be necessary; I think `SchemaError` still fits, however; although it might be a peer of the `ConversionError` subclass rather than a subclass of `ConversionError`. (It is designed to be reported by the schema processing, after all.)

Of course, since the whole Schema is designed to process the JSON parameters to queries, another option would be to expand `SchemaError` (and subclasses) with an HTTP return status; then you could just handle `SchemaError` and abort with the appropriate status.

E.g., a keyword constructor parameter, like

```
class SchemaError(Exception):
   def __init__(self, status: int = 400):
        self.http_status = status
        ...

class UnverifiedUser(SchemaError):
    def __init__(self, user: str):
        super().__init__(status = 401)
        ...
```

lib/pbench/server/api/resources/query_apis/__init__.py
npalaska: Okay I think it would be better if we can somehow separate the public data queries from the queries that does need a user authentication, that way we dont have to manage how the "user" parameter is formatted in the json request in order to understand if it is a public dataset query or an actual user dataset query. For example we can have an endpoint `/controllers/list` that requires a "user" in the json request and an authorization header that belongs to the user in the query. We can also have an endpoint `/controllers/list/public` that does not accept the "user" parameter in the json request, only "start" and "end" and this endpoint does not require an authorization header but internally we do the exact same thing as a class `ControllersList` with user as `None`

This way I feel its more cleaner and probably less prone to security vulnerability (we wont proceed to query the request if the endpoint is strictly protected rather than allowing the optional authorization header). And also maybe it will be easirt to adopt when we have role based authentication

dbutenhof (changes requested):
Sorry, but we absolutely don't want to duplicate API endpoints based on the authorization context!

lib/pbench/server/api/resources/query_apis/__init__.py
dbutenhof: The endpoint shouldn't change, _especially_ based on roles, because the client shouldn't have to know the details of the security token it holds; the server takes care of that. The client's visibility to data changes, but not how they (attempt to) access it. I really don't like this idea of fragmenting APIs; "that way lies madness". (Well, at least a major pain in the butt.)

The authorization token tells the server ("out of band", essentially) what you're _allowed_ to access; the `"user"` parameter tells it what you _want_ to access. If the two scopes overlap, you get what you want; if not, you get a security violation. As I said, we can be "smart" and default a missing `"user"` based on the actual security domain (`"user": ""` or `"user": None` for missing authorization token, or `"user": "<logged-in-user>"` with a token) because that's a convenient and meaningful default... essentially we don't make them specify `"user"` unless they want something unusual, which gets into privilege and roles).

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -104,14 +121,21 @@ def convert_username(value: str) -> str:
...
-        raise ConversionError(value, "username", type(value).__name__)
+        user, verified = Auth().verify_user(value)
+        if verified or user.is_admin():
+            return value
+        else:
+            raise UnVerifiedUser(value)
+    except Exception as exc:
+        raise ConversionError(
+            value, "username", type(value).__name__
+        ) if not isinstance(exc, UnVerifiedUser) else UnVerifiedUser(value)
```
portante, line 138: Why not just do:
```
    try:
        user, verified = Auth().verify_user(value)
        if verified or user.is_admin():
            return value
        else:
            raise UnVerifiedUser(value)
    except UnVerifiedUser:
        raise
    except Exception as exc:
        raise ConversionError(
            value, "username", type(value).__name__
        )
```

or even

```
    try:
        user, verified = Auth().verify_user(value)
    except Exception as exc:
        raise ConversionError(
            value, "username", type(value).__name__
        )
    if verified or user.is_admin():
        return value
    else:
        raise UnVerifiedUser(value)
```

lib/pbench/server/api/resources/query_apis/controllers_list.py
```diff
@@ -27,6 +28,14 @@ def __init__(self, config: PbenchServerConfig, logger: Logger):
             ),
         )
 
+    @Auth.token_auth.login_required()
+    def post(self):
+        return super().post()
+
+    @Auth.token_auth.login_required()
+    def get(self):
+        return super().get()
```
portante, line 37: Can you add a comment for why you have to do this?  I am assuming it is because the default `post()` and `get()` methods don't have `login_required`?

lib/pbench/server/api/resources/query_apis/controllers_list.py
```diff
@@ -142,3 +151,23 @@ def postprocess(self, es_json: Dict[AnyStr, Any]) -> Dict[AnyStr, Any]:
...
+            config,
+            logger,
+            Schema(
+                Parameter("start", ParamType.DATE, required=True),
+                Parameter("end", ParamType.DATE, required=True),
+            ),
+        )
+
+    def assemble(self, json_data: Dict[AnyStr, Any]) -> Dict[AnyStr, Any]:
+        json_data["user"] = None
```
portante, line 172: Why is the user added here as `None`?  Can you add a comment explaining why we have to do this?

lib/pbench/server/api/resources/query_apis/controllers_list.py
portante: Same comments for the other modules where this occurs.

lib/pbench/server/api/resources/query_apis/controllers_list.py
portante: Same comments for the other modules where this occurs.

portante (changes requested):

webbnh (changes requested):
There are several things which should be fixed prior to merging:

- Various stale/cut-and-paste-o'd docstrings (mostly in the tests, but also one in the code);
- A couple of misplaced/unnecessary bits of code;
- A broken assertion in one of the unit tests; and,
- The unit test monkey-patching of `Auth.verify_user()` looks broken.

The unit tests could use some polishing and perhaps some augmentation.

Also, I second Dave's objection to the duplicate endpoints.

lib/pbench/server/api/resources/query_apis/__init__.py
portante: @dbutenhof, can you add a reference to that tracking issue here?

lib/pbench/server/api/resources/query_apis/__init__.py
portante: @webbnh, can you extract this as a discussion topic?  I'd like to expose this thinking more broadly than just a code review comment.

webbnh: Done:  #2277

lib/pbench/server/api/resources/query_apis/controllers_list.py
npalaska: Fixed and added the comment

lib/pbench/server/api/resources/query_apis/controllers_list.py
npalaska: Fixed

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -29,6 +43,7 @@ class InvalidRequestPayload(SchemaError):
     """
 
     def __str__(self):
+        super().__init__()
         return "Invalid request payload"
```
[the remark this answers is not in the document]
npalaska: Fixed

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -104,14 +121,21 @@ def convert_username(value: str) -> str:
...
 
     Returns:
         internal username representation
     """
     try:
-        return Auth.validate_user(value)
-    except Exception:
-        raise ConversionError(value, "username", type(value).__name__)
+        user, verified = Auth().verify_user(value)
+        if verified or user.is_admin():
```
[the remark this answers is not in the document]
npalaska: Yes, if an admin user is doing a request for some other user, the verified status would be false (as the username wont match) but the requesting user is an admin

lib/pbench/test/unit/server/test_controllers_list.py
```diff
@@ -60,11 +62,44 @@ def build_index(self, server_config, dates):
             index += f"{idx}{d},"
         return index
 
-    def test_missing_json_object(self, client, server_config):
+    def test_missing_auth_header(self, client, server_config):
         """
-        test_missing_json_object Test behavior when no JSON payload is given
+        test_missing_json_object Test behavior when no Authorization header is given
```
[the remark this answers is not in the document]
npalaska: I removed this test since the authentication is optional now

lib/pbench/test/unit/server/test_controllers_list.py
```diff
@@ -60,11 +62,44 @@ def build_index(self, server_config, dates):
...
         """
         response = client.post(f"{server_config.rest_uri}/controllers/list")
```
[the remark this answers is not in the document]
npalaska: I removed this test since the authentication is optional now

lib/pbench/test/unit/server/test_controllers_list.py
```diff
@@ -60,11 +62,44 @@ def build_index(self, server_config, dates):
...
+            headers={"Authorization": pbench_token},
+        )
+        assert response.status_code == 401
+
+    def test_non_accessible_user_data(self, client, server_config, pbench_token):
+        """
+        test_missing_json_object Test behavior when Authorization header does not have access to other user's data
+        """
+        # Auth token belongs to the user "drb"
+        # Trying to access the data belong to the user "pp"
```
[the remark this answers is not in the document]
npalaska: Right, and I should mention this in the comment that the auth token belongs to the "drb" user because the pytest fixture `pbench_token` creates it for the "drb" user. 

This test is just testing whether a different user's auth token can access some other user data (doesn't matter if the data for the other user is present or not, we just won't proceed with the request), test to see if the user can access their own data via auth token is performed in other functions such as `test_query` if thats what you mean

lib/pbench/test/unit/server/test_controllers_list.py
```diff
@@ -88,20 +123,25 @@ def test_missing_keys(self, client, server_config, keys, user_ok):
         however, Pbench will silently ignore any additional keys that are
         specified.
```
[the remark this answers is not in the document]
npalaska: fixed

lib/pbench/test/unit/server/test_controllers_list.py
```diff
@@ -110,7 +150,7 @@ def test_bad_dates(self, client, server_config, user_ok):
             == "Value '2020-19' (str) cannot be parsed as a date/time string"
         )
 
-    def test_query(self, client, server_config, query_helper, user_ok):
+    def test_query(self, client, server_config, query_helper, user_ok, pbench_token):
```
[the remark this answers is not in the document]
npalaska: Fixed

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -48,30 +48,36 @@ def test_host_info(client, pytestconfig, caplog, server_config):
 
 class TestElasticsearch:
     @staticmethod
-    def test_json_object(client, caplog, server_config):
-        response = client.post(f"{server_config.rest_uri}/elasticsearch")
+    def test_json_object(client, caplog, server_config, pbench_token):
```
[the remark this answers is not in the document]
npalaska: The Authentication token is optional now

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -109,33 +115,31 @@ def test_malformed_authorization_header(client, caplog, server_config):
             assert record.levelname not in ("ERROR", "CRITICAL")
 
     @staticmethod
-    def test_missing_filename_header_upload(client, caplog, server_config):
+    def test_missing_filename_header_upload(
+        client, caplog, server_config, pbench_token
+    ):
         with client:
```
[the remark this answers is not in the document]
npalaska: That's a good point

lib/pbench/test/unit/server/test_schema.py
```diff
@@ -35,11 +36,11 @@ def test_enum(self):
...
     )
-    def test_successful_conversions(self, test, monkeypatch):
-        def ok(user: str) -> str:
-            return user
+    def test_successful_conversions(self, client, test, monkeypatch, pbench_token):
+        def ok(user: User, verified: bool) -> str:
+            return user, verified
 
-        monkeypatch.setattr(Auth, "validate_user", ok)
+        monkeypatch.setattr(Auth, "verify_user", ok)
```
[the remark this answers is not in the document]
npalaska: `verify_user` is an object method of the Auth class so I can't set the monkeypatch attribute as `monkeypatch.setattr(Auth(), "verify_user", ok)`. if `verify_user` is a static method then it would have been fine passing just one parameter, but since its an object method I have to pass 2 parameters (I think it considers `self` as 1st parameter).


But I think you pointed out a valid mistake, the function should return a User and a bool (in this case always `True`) and I should make the analogous change on line 62

So I think the right way to do this would be
```
def ok(user: User, username: str) -> (User, bool):
    return user, True

monkeypatch.setattr(Auth, "verify_user", ok)
```

lib/pbench/server/api/resources/query_apis/__init__.py
npalaska: I made the authentication optional

lib/pbench/server/api/resources/query_apis/__init__.py
npalaska: Fixed

npalaska:
Removed the duplicate endpoints based on the public dataset and authentication context

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
```diff
@@ -25,11 +25,27 @@ def build_index(self, server_config, dates):
             index += f"{idx}{d},"
         return index
 
-    def test_missing_json_object(self, client, server_config):
+    def test_non_accessible_user_data(self, client, server_config, pbench_token):
```
webbnh, line 28: There is a matrix, here:
requestor | target | result
-- | -- | --
valid (e.g., `drb`) | valid (e.g., `drb`) | ✅ 
valid (e.g., `drb`) | invalid (e.g., `pp`) | ❌ 
valid (e.g., `drb`) | public (e.g., `"user": ""`) | ✅ 
valid (e.g., `drb`) | public (e.g., `"user"` field omitted) | ✅ 
invalid (e.g., `pp`) | valid (e.g., `drb`) | ❌ 
invalid (e.g., `pp`) | invalid (e.g., `pp`) | ❌ 
invalid (e.g., `pp`) | public (e.g., `"user": ""`) | ❓ 
invalid (e.g., `pp`) | public (e.g., `"user"` field omitted) | ❓ 
none (e.g., no token) | valid (e.g., `drb`) | ❌ 
none (e.g., no token) | invalid (e.g., `pp`) | ❌ 
none (e.g., no token) | public (e.g., `"user": ""`) | ✅ 
none (e.g., no token) | public (e.g., `"user"` field omitted) | ✅ 

How many of these cases do we have tests for?  (We might be able to do this with a single, parameterized test; or we could do it with a few of them.)

(Re the ❓ entries:  I don't know whether these should succeed or fail -- we need to decide...do we treat an invalid user/token/login differently from no-user/token/login?)

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -115,9 +129,13 @@ def convert_username(value: str) -> str:
         internal username representation
     """
     try:
-        return Auth.validate_user(value)
+        user, verified = Auth().verify_user(value)
     except Exception:
         raise ConversionError(value, "username", type(value).__name__)
+    if verified or (user and user.is_admin()):
+        return value
```
dbutenhof, line 136: You've changed the semantics of the conversion; `validate_user` returns `str(user.id)`, which is the Elasticsearch owner key, __not__ the username.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
```diff
@@ -198,7 +219,7 @@ def test_empty_query(
         ),
     )
     def test_http_exception(
-        self, server_config, query_api, exceptions, user_ok, find_template
+        self, server_config, query_api, exceptions, user_ok, find_template, pbench_token
```
webbnh, line 222: I don't think you need to add `pbench_token` here -- this test doesn't seem to reference it, and the fixtures that need it have it baked in.

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -514,15 +533,33 @@ def post(self):
...
+        if Auth.token_auth.current_user() is None:
+            if json_data and "user" in json_data:
+                abort(HTTPStatus.FORBIDDEN, message="Not Authorized")
         try:
             new_data = self.schema.validate(json_data)
+            # After the json data is validated make sure the user parameter is present
+            # in new_data for an endpoint which needs a "user" parameter.
+            # If the request does not contain the user field, it will be interpreted as
+            # a public dataaset query. Therefore after validating the json data we will
+            # manually add the user field as None to let the ensemle method understand that this
```
dbutenhof, line 545: "assemble" not "ensemble" ;-)

And note for other reviewers: I've already changed this in a forthcoming PR (which includes "publish" mechanism) so we won't have to inject a "user" key to satisfy the user term builder. So consider this a temporary bridge to slightly minimize my rebase conflicts!

lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
```diff
@@ -25,11 +25,32 @@ def build_index(self, server_config, dates):
             index += f"{idx}{d},"
         return index
 
-    def test_missing_json_object(self, client, server_config):
+    def test_non_accessible_user_data(self, client, server_config, pbench_token):
+        """
+        test_missing_json_object Test behavior when Authorization header does not have access to other user's data
+        """
+        # Auth token belongs to the user "drb" because the pytest fixture pbench_token creates it for the user "drb".
+        # Trying to access the data belong to the user "pp"
```
webbnh, line 33: See all three comments in `TestControllersList:test_non_accessible_user_data()` -- they apply to this function, too.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
```diff
@@ -25,11 +25,27 @@ def build_index(self, server_config, dates):
             index += f"{idx}{d},"
         return index
 
-    def test_missing_json_object(self, client, server_config):
+    def test_non_accessible_user_data(self, client, server_config, pbench_token):
+        """
+        test_missing_json_object Test behavior when Authorization header does not have access to other user's data
+        """
+        # Auth token belongs to the user "drb" because the pytest fixture pbench_token creates it for the user "drb".
```
dbutenhof, line 32: That's a bit long and convoluted. How about "The pbench_token fixture logs in as user "drb""

lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
```diff
@@ -55,20 +75,25 @@ def test_missing_keys(self, client, server_config, keys, user_ok):
         however, Pbench will silently ignore any additional keys that are
         specified.
```
webbnh, line 76: The code doesn't test this assertion.  (The change made to `TestControllersList:test_missing_keys()` should be made here, too.)

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
```diff
@@ -25,11 +25,27 @@ def build_index(self, server_config, dates):
             index += f"{idx}{d},"
         return index
 
-    def test_missing_json_object(self, client, server_config):
+    def test_non_accessible_user_data(self, client, server_config, pbench_token):
+        """
+        test_missing_json_object Test behavior when Authorization header does not have access to other user's data
```
dbutenhof, line 30: I'm trying to stop repeating the function name in the docstring. It wastes space, and it can easily get out of date during refactoring. It's silly that the template even has it.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
```diff
@@ -41,32 +57,37 @@ def test_missing_json_object(self, client, server_config):
...
         ),
     )
-    def test_missing_keys(self, client, server_config, keys, user_ok):
+    def test_missing_keys(self, client, server_config, keys, user_ok, pbench_token):
         """
         test_missing_keys Test behavior when JSON payload does not contain
         all required keys.
 
-        Note that "user", "start", and "end" are all required;
+        Note that "start", and "end" are required where as "user" is not mandatory;
```
dbutenhof, line 68: "whereas" is one word. I would have overlooked this, but I suspect Webb would point it out if I don't first. ;-)

lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
```diff
@@ -55,20 +75,25 @@ def test_missing_keys(self, client, server_config, keys, user_ok):
...
        """
-        response = client.post(f"{server_config.rest_uri}/datasets/detail", json=keys)
+        response = client.post(
+            f"{server_config.rest_uri}/datasets/detail",
+            headers={"Authorization": "Bearer " + pbench_token},
+            json=keys,
+        )
         assert response.status_code == HTTPStatus.BAD_REQUEST
-        missing = [k for k in ("user", "name", "start", "end") if k not in keys]
+        missing = [k for k in ("name", "start", "end") if k not in keys]
```
webbnh, line 84: This code doesn't match the comment at line 74.

lib/pbench/server/api/auth.py
```diff
@@ -76,6 +76,20 @@ def get_secret_key(self):
...
+        querying the Usermodel with the current user
+        :param username:
+        :return: User (UserModel instance), verified status (boolean)
+        """
+        user = self.token_auth.current_user()
+        # check if the current username matches with the one provided
+        verified = user is not None and user.username == username
+        Auth.logger.warning("verified status of user '{}' is '{}'", username, verified)
+
+        return user, verified
```
dbutenhof, line 91: And ... as we've started to veer in this direction... do we want a named tuple here to improve documentation?

lib/pbench/test/unit/server/query_apis/test_datasets_list.py
```diff
@@ -25,11 +25,32 @@ def build_index(self, server_config, dates):
             index += f"{idx}{d},"
         return index
 
-    def test_missing_json_object(self, client, server_config):
+    def test_non_accessible_user_data(self, client, server_config, pbench_token):
+        """
+        test_missing_json_object Test behavior when Authorization header does not have access to other user's data
+        """
+        # Auth token belongs to the user "drb" because the pytest fixture pbench_token creates it for the user "drb".
+        # Trying to access the data belong to the user "pp"
```
webbnh, line 33: See all three comments in `TestControllersList:test_non_accessible_user_data()` -- they apply to this function, too.

lib/pbench/test/unit/server/query_apis/test_datasets_list.py
```diff
@@ -55,20 +75,25 @@ def test_missing_keys(self, client, server_config, keys):
...
        """
-        response = client.post(f"{server_config.rest_uri}/datasets/list", json=keys)
+        response = client.post(
+            f"{server_config.rest_uri}/datasets/list",
+            headers={"Authorization": "Bearer " + pbench_token},
+            json=keys,
+        )
         assert response.status_code == HTTPStatus.BAD_REQUEST
-        missing = [k for k in ("user", "controller", "start", "end") if k not in keys]
+        missing = [k for k in ("controller", "start", "end") if k not in keys]
```
webbnh, line 84: See both comments in `TestDatasetsDetail:test_missing_keys()` -- they apply here, too.

dbutenhof (changes requested):
I'm assuming your GOLD tests would have failed due to the change in document ownership from `user.id` to `value` (username). Interesting that GitHub is still waiting on our dead Jenkins servers ...

lib/pbench/server/api/auth.py
```diff
@@ -76,6 +76,20 @@ def get_secret_key(self):
         except Exception as e:
             Auth.logger.exception(f"{__name__}: ERROR: {e.__traceback__}")
 
+    def verify_user(self, username):
```
webbnh, line 79: Could you add type hints, please?

lib/pbench/test/unit/server/test_schema.py
```diff
@@ -35,11 +36,11 @@ def test_enum(self):
...
     )
-    def test_successful_conversions(self, test, monkeypatch):
-        def ok(user: str) -> str:
-            return user
+    def test_successful_conversions(self, client, test, monkeypatch):
+        def ok(user: User, username: str) -> (User, bool):
+            return user, True
 
-        monkeypatch.setattr(Auth, "validate_user", ok)
+        monkeypatch.setattr(Auth, "verify_user", ok)
```
webbnh, line 43: I'm fairly certain that the signature of `ok()` is supposed to match exactly the signature of `verify_user()`; thus, the first parameter needs to be an `Auth`, not a `User`.  And, this means that the mock has to come up with the appropriate `User` value somehow (presumably, from `test[1]`).

And, the values at line 36 look like user "names" when they should be user "IDs".

lib/pbench/test/unit/server/test_schema.py
```diff
@@ -58,7 +59,7 @@ def test_failed_conversions(self, test, monkeypatch):
         def not_ok(user: str) -> str:
             raise UnknownUser()
 
-        monkeypatch.setattr(Auth, "validate_user", not_ok)
+        monkeypatch.setattr(Auth, "verify_user", not_ok)
```
webbnh, line 62: As I pointed out in my previous comment, the signature for `not_ok()` is incorrect -- it should match `verify_user()`.

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -49,30 +49,37 @@ def test_host_info(client, pytestconfig, caplog, server_config):
 
 class TestElasticsearch:
     @staticmethod
-    def test_json_object(client, caplog, server_config):
-        response = client.post(f"{server_config.rest_uri}/elasticsearch")
+    def test_json_object(client, caplog, server_config, pbench_token):
```
webbnh, line 52: This test seems misnamed:  shouldn't it be "test _missing_ json object"?

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -127,174 +134,165 @@ def test_malformed_authorization_header(self, client, caplog, server_config):
...
-            assert response.status_code == HTTPStatus.BAD_REQUEST
-            assert response.json.get("message") == expected_message
-            self.verify_logs(caplog)
+        expected_message = "Missing required Content-MD5 header"
+        response = client.put(
+            self.gen_uri(server_config), headers=self.gen_headers(pbench_token),
+        )
+        assert response.status_code == HTTPStatus.BAD_REQUEST
+        assert response.json.get("message") == expected_message
+        self.verify_logs(caplog)
```
webbnh, line 159: This is suspicious:  we need a token in order to test some bad requests but not others?  :-(

webbnh (changes requested):
Issues:
 - Please fix the LGTM complaints (I realize that you didn't cause them, but they are easy to fix and in code which you are changing...and, if you don't fix them, I don't know when they will be flagged again...).
 - There are several problems (including a bug, I think) in `convert_username()`.
 - We don't appear to be imposing any kind of authorization on the ES GET requests.
 - I think we're missing a bunch of unit test scenarios.

And, there's a bunch of other, smaller stuff below, including stuff that you didn't actually address from my previous review (you marked the conversations resolved when they weren't...).

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
webbnh: I caught it, got distracted, and then failed to get back to it.  😀

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -115,9 +129,13 @@ def convert_username(value: str) -> str:
...
     """
     try:
-        return Auth.validate_user(value)
+        user, verified = Auth().verify_user(value)
     except Exception:
         raise ConversionError(value, "username", type(value).__name__)
+    if verified or (user and user.is_admin()):
+        return value
+    else:
+        raise UnVerifiedUser(value)
```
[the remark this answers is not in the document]
npalaska: Yes I agree the boolean verified variable was redundant  and I refactored the verify_user function a bit, it now only returns an instance of the `User` If the request is not authenticated we return None as we won't be able to verify the target user. However, if the current user is an admin, we can return the instance of the target user by querying the target username provided.

lib/pbench/test/unit/server/test_requests.py
npalaska: What you mean?

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -514,15 +533,33 @@ def post(self):
...
+            self.logger.warning("{}", str(e))
+            abort(HTTPStatus.FORBIDDEN, message=str(e))
         except SchemaError as e:
             self.logger.warning(
                 "{}: {} on {!r}", self.__class__.__name__, str(e), json_data
             )
             abort(HTTPStatus.BAD_REQUEST, message=str(e))
         return self._call(requests.post, new_data)
 
+    @Auth.token_auth.login_required(optional=True)
```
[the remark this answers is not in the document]
npalaska: Yes, That's a valid point but the only GET request we can do right now is `month_indices` API call and that does not need a user parameter. Only POST API call need a user parameter. I added the `optional=True` thinking we might need it when we add more APIs that needs a user parameter on the GET request. So you are right at the moment we dont need this decorator on the GET request.

lib/pbench/test/unit/server/test_requests.py
webbnh: In `test_missing_controller_header_upload()`, we specify an `"Authorization"` header in the PUT request; in `test_missing_md5sum_header_upload()`, we don't.

I understand that we don't want to test every combination of present and absent headers, but it would make the testing of missing headers more comprehensible if every test presented all the headers except for the one whose absence it is testing.

To the greatest extent possible, unit testing should not be implemented based on "whitebox knowledge" of the CUT.  The unit tests should control the inputs to the function or unit being tested, including behavior driven by its dependencies (i.e., by mocking subroutine calls), and they should avoid relying on implementation details within the function (like the order in which headers are checked).  That sort of reliance results in fragile test scenarios which have to be modified when the CUT is modified, which undermines the robustness of the testing -- we want the freedom to modify the CUT without modifying the tests, so that the tests will prove that the new code is still correct.  It's true that unit testing does require significant knowledge of the implementation -- this is required in order to supply appropriate mocks and to ensure that all paths through the CUT are exercised -- but we need to approach this in such a way as to avoid too close a marriage between the tests and the CUT.

npalaska: Actually we are using `gen_headers` utility method to generate headers in `test_missing_md5sum_header_upload` it only takes the `pbench_token` as a parameter and adds it to the set of header values (controller and authorization). We can't use that method in `test_missing_controller_header_upload` because it by defaults add the controller to the header list.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
npalaska: Most of these tests are present now but I have some confusion
requestor | target | result |Tested
-- | -- | -- | --
valid (e.g., `drb`) | valid (e.g., `drb`) | ✅ |✅
valid (e.g., `drb`) | invalid (e.g., `pp`) | ❌ |✅
valid (e.g., `drb`) | public (e.g., `"user": ""`) | ✅ |✅
valid (e.g., `drb`) | public (e.g., `"user"` field omitted) | ✅ |✅
invalid (e.g., `pp`) | valid (e.g., `drb`) | ❌ |✅ (vice-versa as 2nd case)
invalid (e.g., `pp`) | invalid (e.g., `pp`) | ❌ |(by invalid requester do you mean unauthenticated?)
invalid (e.g., `pp`) | public (e.g., `"user": ""`) | ❓| =
invalid (e.g., `pp`) | public (e.g., `"user"` field omitted) | ❓|=
none (e.g., no token) | valid (e.g., `drb`) | ❌ |(Does invalid and unauthenticated user mean the same thing?)
none (e.g., no token) | invalid (e.g., `pp`) | ❌| =
none (e.g., no token) | public (e.g., `"user": ""`) | ✅ |=
none (e.g., no token) | public (e.g., `"user"` field omitted) | ✅ |=

dbutenhof (approved):
Looking good!

dbutenhof (approved):

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
webbnh: > by invalid requester do you mean unauthenticated?

> Does invalid and unauthenticated user mean the same thing?

By "invalid", I think what I meant was a request specifying a token which is invalid.  I believe that this case is distinct from a request which specifies no token (and, obviously, it's distinct from a request specifying a valid token).

webbnh (commented):
I ran out of time to finish, so I'm sharing what I've got so far; I'll resume on Monday.

lib/pbench/server/api/auth.py
```diff
@@ -76,19 +76,25 @@ def get_secret_key(self):
...
+        If the request is not authenticated (current user is None) we return None as
+        we won't be able to verify the target user. However, if the current user is
+        an admin, we return the instance of the target user by querying the target
+        username provided.
+        """
+        current_user = Auth.token_auth.current_user()
+        if not current_user:
+            return None
+        if current_user and current_user.username == target_username:
+            return current_user
```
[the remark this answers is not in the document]
npalaska: Ah, forgot to refactor this change

webbnh (changes requested):
Looks good; just needs a little more polish.

lib/pbench/test/unit/server/test_requests.py
webbnh: I think it would be better if the code did not use `gen_headers()` for the test scenarios which are omitting the various headers.  The function is handy for the other cases, where the headers are more or less a given, but, for the cases where we are explicitly testing what happens when one of them is missing, I recommend that you just construct the header object explicitly -- it's nearly trivial to do (we could almost get by without using `gen_headers()` in the other functions...), and doing it that way makes it obvious to the reader what is and isn't being specified.  And then you can further simplify `gen_headers()` by getting rid of the default value for the hash argument and the conditional code, because it will always be constructing all of the headers.

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -55,6 +55,7 @@ class MissingParameters(SchemaError):
     """
 
     def __init__(self, keys: List[AnyStr]):
+        super().__init__()
```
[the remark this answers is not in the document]
npalaska: I think so, missing parameter is of type `bad_request` and also It was `400` before this change

lib/pbench/test/unit/server/test_requests.py
npalaska: Yeah that make sense

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
npalaska: Got it, added more unittests for couple of missing scenarios from the table

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
```diff
@@ -38,6 +38,33 @@ def test_non_accessible_user_data(self, client, server_config, pbench_token):
...
+    ):
+        """
+        Test behavior when Authorization header does not have access to other user's data
+        """
+        # valid token logout
+        response = client.post(
+            f"{server_config.rest_uri}/logout",
+            headers=dict(Authorization="Bearer " + pbench_token),
+        )
+        assert response.status_code == 200
```
webbnh, line 52: You should use `HTTPStatus.OK` instead of the numeric literal here (and in datasets-list and datasets-details).

webbnh (changes requested):
Getting there!

I think we can use a nested parameterization approach to testing the combinations of valid/invalid/omitted token against valid/invalid/public/omitted targets which should simplify things somewhat.

There are two more test cases to consider.

And there are a couple of other small things in the unit tests.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
npalaska: I believe this case is already covered in `test_non_accessible_user_data`

lib/pbench/test/unit/server/query_apis/conftest.py
```diff
@@ -29,17 +29,25 @@ def query_api(
...
         with responses.RequestsMock() as rsp:
             rsp.add(responses.POST, es_url, **kwargs)
-            response = client.post(
-                f"{server_config.rest_uri}{pbench_uri}", json=payload
-            )
-            assert len(rsp.calls) == 1
+            if authorization:
+                response = client.post(
+                    f"{server_config.rest_uri}{pbench_uri}",
+                    headers={"Authorization": "Bearer " + pbench_token},
```
dbutenhof, line 43: You could instead just set `headers = {}` and insert the `authorization` header if necessary, which should allow a single `post` rather than repeating it.

lib/pbench/test/unit/server/query_apis/test_datasets_list.py
```diff
@@ -191,6 +218,31 @@ def test_query(
...
+            assert len(res_json.keys()) == 1
+            data = res_json["dhcp31-187.example.com"]
+            assert isinstance(data, list)
+            assert len(data) == 1
+
+        else:
+            response = client.post(
+                f"{server_config.rest_uri}/datasets/list", json=json,
+            )
+            assert response.status_code == HTTPStatus.FORBIDDEN
```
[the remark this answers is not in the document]
npalaska: okay I refactored a little bit so `authentication` can be received as a parameterized argument but I still need `if-else` condition because pytest gonna create all the combinations of `user` and `authentication` field e.g. `(user: "", authentication: True), (user: "", authentication: False), (user: None, authentication: True), ...` and all these tests should be `HTTPStatus.OK` except `(user: "drb", authentication: False)` this should be `401`

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
webbnh: Cool.  (Do we have a test which shows that administrative access succeeds?  If not, no biggie -- it's early days for that, yet.)

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
```diff
@@ -176,51 +184,39 @@ def test_query(
...
+            assert res_json[0]["key"] == "unittest-controller1"
+            assert res_json[0]["controller"] == "unittest-controller1"
+            assert res_json[0]["results"] == 2
+            assert res_json[0]["last_modified_value"] == 1.59847315581e12
+            assert res_json[0]["last_modified_string"] == "2020-08-26T20:19:15.810Z"
+            assert res_json[1]["key"] == "unittest-controller2"
+            assert res_json[1]["controller"] == "unittest-controller2"
+            assert res_json[1]["results"] == 1
+            assert res_json[1]["last_modified_value"] == 1.6
+            assert res_json[1]["last_modified_string"] == "2020-09-26T20:19:15.810Z"
```
webbnh: I'm still confused as to why there is a conditional here.

For one thing, I'm concerned that the conditional is "too smart".  But, that aside, it seems to me that we should be invoking the query unconditionally (with variations controlled by `user` and `authentication`/`authorization`) and checking for the expected results.  We should be able to set the expected results (at least as far as the `response.status_code` value) using the `parameterize` decorators.  (The big block of assertions for the success case can be left substantially as is, but invoked only when the response is `OK`.)

What I'm stuck on here is that I think that the CUT should be invoked in the _same way_, modulo the individual parameterizations, and I don't feel like this code is making that happen.  In particular, we shouldn't be modifying how the query is sent based on the inputs -- it's only the expected results which should vary.  (I hope this makes sense....)

lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
```diff
@@ -138,8 +138,16 @@ def test_bad_dates(self, client, server_config, user_ok, pbench_token):
     @pytest.mark.parametrize(
         "user", ("drb", "", "no_user", None),
     )
+    @pytest.mark.parametrize("authentication", (True, False))
```
webbnh, line 141: Shouldn't there three cases, not two:  valid authentication/authorization, invalid auth, and missing auth?
(We need to chat about the difference between "authentication" and "authorization"....)

webbnh (changes requested):
I think we're getting close.  However, I'm nervous about the bifurcated approach that the unit tests are still using.  I think the unit tests should be doing the same thing each time and looking for the appropriate result, so, any conditionalization that we need should be on the _results_ side, not on the call side.

lib/pbench/test/unit/server/query_apis/test_datasets_list.py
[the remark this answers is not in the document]
webbnh: You should be able to add a parameter for the expected result to the parameterization (to the outer one, if the expected result is independent of the inner one, or to the inner one possibly incorporating the value of the outer one if it's available in that context, or, failing those, you can generate the expected result in the function body, based on the two parameterization values).  It's perfectly reasonable to have an `if-else` construct for determining the expected result; what I'm objecting to is having a conditional affecting how the CUT is invoked (`client.post()` vs. `query_api()`) -- the parameterization should take care of that.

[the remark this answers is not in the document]
npalaska: We can certainly add expected result as a parameterization but it doesn't add much value, because in this test the only time the expected result would be different than the `OK` is when we encounter this scenario `(user: "drb", authentication: False)`. Also the `query_api` internally does `client.post()` so they are not different per se. The only difference is that `query_api` builds a mock response with an expected index and other user provided parameter such as response payload, status, etc on top of the POST. Even if I use the `query_api` for `(user: "drb", authentication: False)` I have to use different set of parameters such as I can't pass the response payload (as there wont be any payload in the response since its 403) and also doing `rsp.add(responses.POST, es_url, **kwargs)` in `query_api` results in an error because in our definition of `post` (lib/pbench/server/api/resources/query_apis/__init__.py) we abort early if we encounter 403 and so wont proceed to execute `python.request`.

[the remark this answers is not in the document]
webbnh: > they are not different per se.

It's the "per se" part that I'm stuck on.  The fact that they are taking distinct paths to generate the result poses problems:  as the code evolves, it's possible for cracks to open between the two paths where things go untested, and it's hard for the reader to discern what coverage the two paths generate.  It's better if we can avoid multiple paths on the driver side and have them only in the part which analyzes the returns.

Regarding parameterizing the expected value, it's a toss up:  having the expected values laid out in a tabular format has nice power, but, if there's only one scenario out of the dozen which produces a different result, then perhaps it would be better to special-case the generation of that value.

With regard to the rest of it, I think it boils down to having the test code express its expectations appropriately.  That is, the expectations are sort of a bundle:  we expect a certain status return, we expect a certain message and/or other returned payload, and we expect certain behavior with respect to requests made by the CUT.  Thus, if we set up a response to the server's request, and then we get a 403 (which we expect), we should also expect that the response that we queued will be unconsumed -- isn't there a way to test/assert that the response is unconsumed (and thereby consume/remove it) without resulting in a test failure?

lib/pbench/test/unit/server/query_apis/conftest.py
```diff
@@ -29,17 +29,21 @@ def query_api(
...
         es_url = f"http://{host}:{port}{expected_index}{es_uri}"
         with responses.RequestsMock() as rsp:
             rsp.add(responses.POST, es_url, **kwargs)
+            if authorization:
+                headers = {"Authorization": "Bearer " + pbench_token}
+            else:
+                headers = {}
             response = client.post(
-                f"{server_config.rest_uri}{pbench_uri}", json=payload
+                f"{server_config.rest_uri}{pbench_uri}", headers=headers, json=payload,
```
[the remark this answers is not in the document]
npalaska: Ah, a little tricky one though, most of the request need an authentication token in the header since the request json we build has a valid "user" field in it. So I will have to use `pbench_token` fixture repetitively in lot of tests instead of using just once in `query_api`. Will that be a desirable?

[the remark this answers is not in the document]
webbnh: IIUC, the value of a text fixture is [cached by PyTest](https://docs.pytest.org/en/6.2.x/fixture.html#fixtures-can-be-requested-more-than-once-per-test-return-values-are-cached), so I don't think it should be a problem to use `pbench_token` repeatedly.

Making the default be the most-used value is appealing, but what is probably more important is to use the parameter in the way which makes it easiest to understand the tests.  For that, I'm inclined to suggest that omitting the parameter should imply omitting the value (i.e., the default value should be `{}`).  The fact that all the other invocations then have to supply the token is not necessarily a bad thing -- that makes it explicit what they are doing, which is generally good for understanding.

lib/pbench/test/unit/server/query_apis/test_datasets_list.py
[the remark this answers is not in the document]
webbnh: > isn't there a way to test/assert that the response is unconsumed (and thereby consume/remove it) without resulting in a test failure?

`query_api()` uses the `responses.RequestsMock()` as a context manager, which is nice in terms of encapsulation, but inconvenient in terms of implementing assertions in the calling test code.  The object could be instead be implemented as a persistent object or even as [a PyTest fixture](https://pypi.org/project/responses/#responses-as-a-pytest-fixture), which would allow the test to check whether the responses have been consumed or not.  (Also, there are `reset()` and `remove()` methods which could be used to clean up after the test, and there is an option on the c'tor to suppress the unused requests assertion...so there are various options.)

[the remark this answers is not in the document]
npalaska: Okay for some reason github just didn't load this comment (or my browser was loading the old page from cache). Anyway I added a conditional check to see if the expected status is `FORBIDDEN` or `UNAUTHORIZED` since these are the 2 conditions when we would abort early. So if the expected status is not one of these 2, we would add the response body to the mock otherwise we skip it 
```
if expected_status not in [HTTPStatus.FORBIDDEN, HTTPStatus.UNAUTHORIZED]:
    rsp.add(responses.POST, es_url, **kwargs)
```

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
npalaska: No the test does not check for the admin access scenario, but that can be handled in another PR. I believe for all the tests files we need to add this admin access test since we now have the cli implemented.

webbnh: Anything that we discover during review that we're not going to address in the pull request should probably have an issue opened to track it.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
```diff
@@ -240,17 +236,24 @@ def test_empty_query(
             },
         }
 
+        expected_status = HTTPStatus.OK
+        if "user" in json and json["user"] and build_auth_header[1] != "valid":
+            expected_status = HTTPStatus.FORBIDDEN
```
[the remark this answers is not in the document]
webbnh: Ditto `test_empty_query()` in `lib/pbench/test/unit/server/query_apis/test_datasets_detail.py`.

lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
```diff
@@ -323,7 +329,7 @@ def test_nonunique_query(
         Check the handling of a query that returns too much data.
         """
         json = {
-            "user": "drb",
+            "user": "",
             "name": "fio",
             "start": "2020-08",
             "end": "2020-10",
```
webbnh, line 335: Since the mock provides the response, how much of a realistic payload do we actually need?

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
```diff
@@ -278,7 +281,7 @@ def test_http_exception(
         Check that an exception in calling Elasticsearch is reported correctly.
         """
         json = {
-            "user": "drb",
+            "user": "",
             "start": "2020-08",
             "end": "2020-08",
         }
```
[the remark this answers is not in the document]
webbnh: Ditto for `lib/pbench/test/unit/server/query_apis/test_datasets_detail.py`.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
```diff
@@ -299,7 +302,7 @@ def test_http_error(self, server_config, query_api, errors, user_ok, find_templa
         response.raise_for_status() and Pbench handlers.
         """
         json = {
-            "user": "drb",
+            "user": "",
             "start": "2020-08",
             "end": "2020-08",
         }
```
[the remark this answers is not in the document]
webbnh: Ditto for `lib/pbench/test/unit/server/query_apis/test_datasets_detail.py`.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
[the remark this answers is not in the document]
webbnh: Ditto for `lib/pbench/test/unit/server/query_apis/test_datasets_list.py`.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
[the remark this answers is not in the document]
webbnh: Ditto for `lib/pbench/test/unit/server/query_apis/test_datasets_list.py`.

webbnh (changes requested):
It looks good; it just needs some polish -- a few code comments would be good, there are a couple spots where the refactoring could use some "fitting" (unnecessary code which can be removed), and some smaller stuff.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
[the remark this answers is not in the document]
npalaska: Yes since we are not using the user paramterization

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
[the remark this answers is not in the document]
npalaska: Yes we do need to set up some payload, I mean thats how the definition of `/controllers/list` endpoint is set (and other endpoints like /details and /dataset/list). If we omit the payload the, the post request can not be completed, I get an assertion error `AssertionError: Not all requests have been executed` but if you mean by omitting a user field that's certainly possible.

lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
npalaska: They need to be realistic, we have validator check on each of these json fields and if they don't match with how its define in our code then we respond back as a bad request

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
npalaska: Created https://github.com/distributed-system-analysis/pbench/issues/2361

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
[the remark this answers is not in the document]
webbnh: OK, so we need to send enough of a payload to get by all the API's parameter checking and up to the point where it makes the request to ES.  I support your effort here to make the parameter as generic as possible.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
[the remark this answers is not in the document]
webbnh: [Answered](https://github.com/distributed-system-analysis/pbench/pull/2251#discussion_r656372613).

lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
webbnh: [Answered](https://github.com/distributed-system-analysis/pbench/pull/2251#discussion_r656372613).

webbnh (dismissed):
Ship it!

lib/pbench/server/api/auth.py
```diff
@@ -76,6 +76,27 @@ def get_secret_key(self):
         except Exception as e:
             Auth.logger.exception(f"{__name__}: ERROR: {e.__traceback__}")
 
+    def verify_user(self, target_username: str) -> User:
+        """
+        Check if the provided username belongs to the current user by
+        querying the User model with the current user
```
portante, line 82: If you come back to this file, missing trailing period.

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -22,7 +22,21 @@ class SchemaError(TypeError):
...
     """
 
-    pass
+    def __init__(self, status: int = 400):
+        self.http_status = status
+
+
+class UnverifiedUser(SchemaError):
+    """
+    Unverified Attempt to access other user data.
```
portante, line 31: If you revisit this file, `s/Attempt/attempt/`.

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -514,15 +534,41 @@ def post(self):
         are not in the class schema are ignored but logged.
         """
         json_data = request.get_json(silent=True)
+        if json_data and "user" in json_data:
```
portante, line 537: This assumes `json_data` is a `dict()`.  Is that always the case?

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -514,15 +534,41 @@ def post(self):
...
+            # parameter. If the request does not contain the user field,
+            # it will be interpreted as a public dataset query. Therefore
+            # after validating the json data we will manually add the user
+            # field as None to let the assemble method understand that this
+            # query is for the public dataset.
+            if (
+                "user" in self.schema.parameters.keys()
+                and "user" not in new_data.keys()
+            ):
+                new_data["user"] = None
```
portante, line 560: This seems odd.  Why do we have to do this when we just removed it above?

portante (dismissed):
Just a few comments, and are we planning on squashing this to one commit?

If we do, can we make sure the commit summary is 50 characters or less, and the body is a proper summary of the changes?

lib/pbench/server/api/resources/query_apis/__init__.py
npalaska: That is because the `assemble` method of the query APIs need the "user" field explicitly set as `None` for querying a public dataset. Therefore after validating the rest of the json data we add it back as a `None`.  The reason I removed it before validating is because the user field was present but empty and validator would raise it as an exception since its not a valid user field however we interpret it as a public dataset query.

But you raised a valid point we dont need to do this, the assemble method can do `user = json_data.get("user")` instead of `user = json_data["user"]` and that should take care of setting the local `user` variable as None.

lib/pbench/server/api/resources/query_apis/__init__.py
npalaska: Yeah I mean flask's `request.get_json` parses the data as JSON objects as key/value pairs, just like a Python dictionary. So its always going to be a dictionary I guess

npalaska:
Yes once I get 2 approval I will squash the commits

npalaska:
By the way @portante the issue you linked will not be addressed in this PR, that was an issue I created from one of the discussion point in this PR

webbnh (dismissed):

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -512,17 +532,34 @@ def post(self):
...
+        if json_data and "user" in json_data:
+            if not json_data["user"]:
+                # There is an empty user field present in the json data;
+                # therefore, this request will be interpreted as a
+                # public dataset query.  Remove it before validating the
+                # remaining json data.
+                del json_data["user"]
+            elif Auth.token_auth.current_user() is None:
+                # Only logged-in users are allowed to query non-public data.
+                abort(HTTPStatus.FORBIDDEN, message="Not Authorized")
```
dbutenhof, line 549: The difference between "missing" and "empty" and between defaulting to "public" vs "mine" really belongs in the validator, not out here. I'll be replacing this anyway as I make the distinction more explicit; but it occurs to me that this is "entangled" as the validator will need to know the authorization user in order to produce the proper internal translation in all cases. Therefore it might make sense to clean this up and pass the authorized user down into the validator (line 551), moving this special-case code into the username validation.

However aside from this "subtle hint" I won't push on that for this PR. :-)

lib/pbench/server/api/resources/query_apis/controllers_list.py
```diff
@@ -58,9 +58,9 @@ def assemble(self, json_data: Dict[AnyStr, Any]) -> Dict[AnyStr, Any]:
             "start" and "end" are datetime objects representing a set of Elasticsearch
                 run document indices in which to search.
         """
-        user = json_data["user"]
-        start = json_data["start"]
-        end = json_data["end"]
+        user = json_data.get("user")
+        start = json_data.get("start")
+        end = json_data.get("end")
```
dbutenhof, line 63: The change from array form to `.get` isn't necessary here since we know they'll exist; but, sure, the consistency doesn't hurt either since missing parameters will have already been validated.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
```diff
@@ -121,31 +179,48 @@ def test_query(self, server_config, query_api, user_ok, find_template):
...
+        self,
+        client,
+        server_config,
+        query_api,
+        user_ok,
+        find_template,
+        build_auth_header,
     ):
         """
         Check proper handling of a query resulting in no Elasticsearch matches.
```
dbutenhof, line 226: Building the repetition into the `build_auth_header` fixture makes universal documentation a bit awkward: however I like how you did that above, and it might not be a bad idea to add the "three times for ..." here, as well, so anyone debugging is aware of what's happening.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
```diff
@@ -121,31 +179,48 @@ def test_query(self, server_config, query_api, user_ok, find_template):
         }
 
         index = self.build_index(server_config, ("2020-08", "2020-09", "2020-10"))
+        # If we're not asking about a particular user, or if the user
+        # field is to be omitted altogether, or if we have a valid
+        # token, then the request should succeed.
+        if not user or user == "no_user" or build_auth_header[1] == "valid":
```
dbutenhof, line 185: Having the `build_auth_header` fixture return a named tuple might make this interface contract just a bit easier to follow. (Or even just a simple `dict`.)

dbutenhof (approved):
Suggestions and nits follow, but this looks pretty good.

lib/pbench/server/api/resources/query_apis/controllers_list.py
npalaska: yes I changed these for the consistency

lib/pbench/server/api/resources/query_apis/__init__.py
npalaska: Okay yeah if its going to get replaced, I would just keep it as it is for now :)

npalaska:
Rebased all the commits with the following 2 minor changes introduced 
- Added some code comments as Dave suggested in his latest review
- `build_auth_header` now returns a dictionary

lib/pbench/server/api/auth.py
```diff
@@ -79,6 +79,27 @@ def get_secret_key(self):
         except Exception as e:
             Auth.logger.exception(f"{__name__}: ERROR: {e.__traceback__}")
 
+    def verify_user(self, target_username: str) -> User:
+        """
+        Check whether the requested target user is the owner of the authorization
+        token provided to the API.
```
webbnh, line 85: I really don't think that that is what this function does -- the `User` returned by this function might have nothing to do with the authorization token if the the token grants Admin access.

What I think this function does is return the `User` associated with the target username, iff the requesting user is authorized to access it.

lib/pbench/test/unit/server/conftest.py
```diff
@@ -1,4 +1,5 @@
 import datetime
+from http import HTTPStatus
 import os
 import pytest
 import shutil
```
webbnh, line 5: I think we are supposed to be grouping the bare `import`'s together and the `from` imports together, not interspersing them.  (So, line 2 should move to between lines 6 and 7.)

webbnh (approved):
It was hard to see what you changed, because the latest changes were squashed into the previous ones.

Having `build_auth_header()` return a `dict` is much nicer!  Although, since we don't need it to be mutable, a `namedtuple` (or `NamedTuple`) would have been better.

And below are a couple of nits I stumbled over.

lib/pbench/server/api/auth.py
```diff
@@ -79,6 +79,27 @@ def get_secret_key(self):
...
+
+        We're returning a User instance corresponding to the target username provided.
+        If the request is not authenticated (current user is None) we return None as
+        we won't be able to verify the target user. However, if the current user is
+        an admin, we return the instance of the target user by querying the target
+        username provided.
```
dbutenhof, line 91: I think Webb's take above would be an improvement on this rather dense paragraph! But I'm not going to gate this PR on any improvement here.

lib/pbench/server/api/auth.py
dbutenhof: That's a good summary, and actually suggests a change in the method name would probably be good. More like `get_target_user` or something?

dbutenhof (approved):
It's probably going to take me a day to rebase on top of this, but I think we're down to small enough nits that we ought to get this in.

PR #2253, opened 2021-04-28
Quiet Redis server connections by default

### LINKED ISSUE #2211: Add periodic message when Tool Data Sink is waiting to connect
Maybe after 15-20 seconds of waiting add a message saying "Waiting for tool-data-sink..." or something similar, as I would probably assume something was broken or had crashed well before waiting a minute for start-up, since it usually takes a max of 2-3s. That also gives the user some time to start up their tds and save the run if they forgot to run it before using the `tool-meister-start` script.

_Originally posted by @Maxusmusti in https://github.com/distributed-system-analysis/pbench/pull/2181#discussion_r605451278_

portante:
We only emit messages on connection failures, on re-established connections, or when we encounter a missing key, and will emit a summary message if it took more than one attempt to connect.

Also fixes #2211.

### DIFF
--- a/lib/pbench/agent/redis.py
+++ b/lib/pbench/agent/redis.py
@@ -222,39 +222,77 @@ class RedisHandler(logging.Handler):
             self.counter += 1
 
 
-def wait_for_conn_and_key(redis_server, key, prog, redis_host, redis_port):
+def wait_for_conn_and_key(redis_server: redis.Redis, key: str, prog: str) -> str:
     """wait_for_conn_and_key - convenience method of both the Tool Meister and
     the Tool Data Sink to startup and wait for an initial connection to the
     Redis server, and for the expected key to show up.
+
+    Arguments:
+
+        redis_server: a Redis client object
+        key: the key name as a string
+        prog: the program name of our caller
+
+    Returns the payload value of the key as a string.
+
+    If successful on the first connection attempt, and the value of the key is
+    retrieved, no messages are issued on stdout.  If the key does not exist yet,
+    we'll report the missing key every 10 seconds from the last successful
+    connection.
+
+    If on the first connection attempt the connection to the Redis server fails,
+    we emit a message to that effect, once.  After that, we'll report each time
+    we are disconnected or re-connected to the Redis server.
     """
     # Loop waiting for the key to show up.
     connected = None
     payload = None
+    attempts = 0
+    errors = 0
+    missing = 0
     while payload is None:
+        attempts += 1
         try:
             payload = redis_server.get(key)
         except redis.ConnectionError:
+            errors += 1
+            # Reset the missing count if we get disconnected
+            missing = 0
             if connected is None:
+                # Only emit this message once for the initial attempt.
                 print(
-                    f"{prog}: waiting to connect to redis server {redis_host}:{redis_port}",
+                    f"{prog}: waiting to connect to Redis server {redis_server}",
                     flush=True,
                 )
                 connected = False
             elif connected:
+                # We always report disconnections.
                 print(
-                    f"{prog}: disconnected from redis server {redis_host}:{redis_port}",
+                    f"{prog}: disconnected from Redis server {redis_server}",
                     flush=True,
                 )
                 connected = False
             time.sleep(1)
         else:
-            if not connected:
+            if connected is False:
+                # We always report re-connections (connections was not None).
                 print(
-                    f"{prog}: connected to redis server {redis_host}:{redis_port}",
-                    flush=True,
+                    f"{prog}: connected to Redis server {redis_server}", flush=True,
                 )
-                connected = True
+            connected = True
             if payload is None:
-                print(f'{prog}: key, "{key}" does not exist yet', flush=True)
+                missing += 1
+                if (missing % 10) == 0:
+                    # Only emit missing key notice every 10 seconds
+                    print(f"{prog}: key '{key}' does not exist yet", flush=True)
                 time.sleep(1)
+
+    if attempts > 1:
+        # Always report stats if it took multiple attempts.
+        print(
+            f"{prog}: connected to Redis server after {attempts:d}"
+            f" attempts (with {errors:d} error(s))",
+            flush=True,
+        )
+
     return payload.decode("utf-8")
--- a/lib/pbench/agent/redis.py
+++ b/lib/pbench/agent/redis.py
@@ -250,13 +250,13 @@ def wait_for_conn_and_key(redis_server: redis.Redis, key: str, prog: str) -> str
     attempts = 0
     errors = 0
     missing = 0
-    while payload is None:
+    while True:
         attempts += 1
         try:
             payload = redis_server.get(key)
         except redis.ConnectionError:
             errors += 1
-            # Reset the missing count if we get disconnected
+            # Reset the missing count since we got disconnected
             missing = 0
             if connected is None:
                 # Only emit this message once for the initial attempt.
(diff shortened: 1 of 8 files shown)


portante (commented):
Arguably should add a unit test for this change.

lib/pbench/agent/redis.py
```diff
@@ -222,39 +222,77 @@ def emit(self, record):
...
                 print(
-                    f"{prog}: connected to redis server {redis_host}:{redis_port}",
-                    flush=True,
+                    f"{prog}: connected to Redis server {redis_server}", flush=True,
                 )
-                connected = True
+            connected = True
             if payload is None:
-                print(f'{prog}: key, "{key}" does not exist yet', flush=True)
+                missing += 1
```
webbnh, line 284: Consider doing the increment after the test:  this will (among other things) result in producing a message immediately if there is no message waiting when the connection is successful.  (If the connection is initially unsuccessful, we get an immediate message; otherwise, we have to wait 10 seconds before we find out that the connection worked but there's nothing to read.)

lib/pbench/agent/redis.py
```diff
@@ -222,39 +222,77 @@ def emit(self, record):
...
                 connected = False
             elif connected:
+                # We always report disconnections.
                 print(
-                    f"{prog}: disconnected from redis server {redis_host}:{redis_port}",
+                    f"{prog}: disconnected from Redis server {redis_server}",
                     flush=True,
                 )
                 connected = False
             time.sleep(1)
```
webbnh, line 275: Do you really want to maintain a separate delay for the non-connected retry (here) and the connected retry (at line 288)?  I can imagine how that could be useful, but, since they are both hard-coded to use the same interval, I would suggest combining them simply by using a single, common call at the bottom of the loop.

webbnh (approved):
Looks good, but, yes, some unit testing would be nice -- unfortunately, there are a _lot_ of paths to test (I count five on the first iteration, and then the number expands combinatorially after that).

Some possible paths:
- Connected first try and received payload (no message)
- Connected first try; received payload on a subsequent try (no message)
- Connected first try; received payload after the tenth try (one message)
- Connected first try; received payload after the twentieth try (two messages)
- Connected on second try and received payload (two messages)
- Connected on second try; received payload on a subsequent try (two messages)
- Connected on second try; received payload after the tenth try (three messages)
- Connected first try, received no payload, then disconnected, then reconnected, then received payload (three messages)

In addition to checking the messages, I would suggest mocking the `sleep()` call to allow you to count the number of calls and to allow time to "pass faster".  ;-)

lib/pbench/agent/redis.py
dbutenhof: Related: it's assumed in the code that 10 delays equals one second (the message says so), so rather than an anonymous inline `1` we'd be better off with a constant with a meaningful name like `ONE_SECOND` to make the semantic assumption clear.

lib/pbench/agent/redis.py
```diff
@@ -222,39 +222,77 @@ def emit(self, record):
...
+                    f"{prog}: connected to Redis server {redis_server}", flush=True,
                 )
-                connected = True
+            connected = True
             if payload is None:
-                print(f'{prog}: key, "{key}" does not exist yet', flush=True)
+                missing += 1
+                if (missing % 10) == 0:
+                    # Only emit missing key notice every 10 seconds
+                    print(f"{prog}: key '{key}' does not exist yet", flush=True)
```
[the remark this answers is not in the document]
dbutenhof: I can see periodic messages if we remain disconnected for a period of time, but this could get fairly noisy especially during startup deployment. Maybe once a minute (or even more) rather than every 10 seconds?

dbutenhof (dismissed):

lib/pbench/agent/redis.py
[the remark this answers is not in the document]
portante: > Why do periodic prints when we're waiting only if we're connected? Isn't a "heartbeat" worthwhile even if we're disconnected?

> I can see periodic messages if we remain disconnected for a period of time, but this could get fairly noisy especially during startup deployment. Maybe once a minute (or even more) rather than every 10 seconds?

This has to do with what I found for behavior when trying to orchestrate the individual components for the Tool Meister sub-system.

When `pbench-tool-meister-start` is responsible for the orchestration, it very carefully and deliberately starts things up in a certain order: Redis server first (confirming it is there and working), Tool Data Sink second (confirming it is there and listening), and then all the Tool Meisters.

In this scenario, the scheme we are following for `wait_for_conn_and_key()` will very likely succeed on the first attempt.  So the complexity here is not due to that case.

When something else is responsible for starting things up, we can have any number of scenarios, e.g. TMs first, TDS 2nd, Redis server last; TDS first, Redis server 2nd, TMs last; 1 TM first, Redis server next, TDS next, etc.

However, the subtlety is that something has to load the keys into the Redis server before the TDS and TMs can begin their operation.  So losing a connection to the Redis server before the TDS and TMs receive their marching orders is in consequential.  As long as it comes back, and they eventually find their keys, we're good.

What we can't tolerate (yet) is getting our marching orders and then having the Redis server disappear.

This sequence allowed me to bring a Redis server up by hand, add keys slowly, and in different orders so that when I screwed up I could just reset the Redis server before it all got going.  And the messages let me know that once I have a Redis server up and running, I had better populate it.

lib/pbench/agent/redis.py
portante: > Related: it's assumed in the code that 10 delays equals one second (the message says so)

10 delays equals one second?  Can you help me see where the message says that?

>, so rather than an anonymous inline `1` we'd be better off with a constant with a meaningful name like `ONE_SECOND` to make the semantic assumption clear.

If we were sleeping for a tenth of a second, I could see making a constant, ... perhaps, but I am not sure the all the mechanism to do that is worth the additional clarity.

lib/pbench/agent/redis.py
```diff
@@ -222,39 +222,76 @@ def emit(self, record):
...
+
+        redis_server: a Redis client object
+        key: the key name as a string
+        prog: the program name of our caller
+
+    Returns the payload value of the key as a string.
+
+    If successful on the first connection attempt, and the value of the key is
+    retrieved, no messages are issued on stdout.  If the key does not exist yet,
+    we'll report the missing key every 10 seconds from the last successful
```
dbutenhof, line 240: This line is the answer to a question you asked elsewhere: "we'll report the missing key every 10 seconds". So, OK, we're each half right; the "code" does make assumptions about the time-span, but it's not _executable_ code nor visible to the end user. Then again, an outdated comment is less likely to trigger a unit test failure and get noticed when something changes, but can still cause confusion to future readers. ;-)

lib/pbench/agent/redis.py
```diff
@@ -222,39 +222,76 @@ def emit(self, record):
...
                 )
-                connected = True
-            if payload is None:
-                print(f'{prog}: key, "{key}" does not exist yet', flush=True)
-                time.sleep(1)
+            if payload is not None:
+                break
+            connected = True
+            if (missing % 10) == 0:
+                # Only emit missing key notice every 10 seconds
```
dbutenhof, line 284: Here, too, we enshrine the time-span assumption in a comment.

dbutenhof (approved):

dbutenhof (approved):

lib/pbench/agent/redis.py
```diff
@@ -222,39 +222,83 @@ def emit(self, record):
...
+            connected = True
+            if (missing % _MISSING_KEY_REPORT_INTERVAL) == 0:
+                # Only emit missing key notice periodically
+                print(f"{prog}: no payload for key '{key}', yet", flush=True)
+            missing += 1
+        time.sleep(_CONNECTION_RETRY_INTERVAL)
+
+    if attempts > 1:
+        # Always report stats if it took multiple attempts.
+        print(
```
webbnh, line 298: So, should these messages be _printed_?...or should they be sent to some sort of logger which might direct them to the terminal?

That is, code might be called via an API from a second-party application, in which case they might not appreciate random messages being spit out....

webbnh (approved):
Looks nice.  However, I have one, small, sticky question below.

lib/pbench/agent/redis.py
portante: Really good question.

The two entities using this are the Tool Meister and the Tool Data Sink.  There are two modes of operation of those services today: containerized or non-containerized.

In the non-containerized case we "daemonize" to run in the background, but we only setup the logging after we daemonize to avoid inadvertently closing the logging `fd`.

So we use `print()`.  We could just hand a function here instead so that they could use a logger.

PR #2255, opened 2021-04-28
Add a template DB table and integrate with index template management

dbutenhof:
This is separated into staged commits which may or may not make review easier.

1. Refactor PbenchTemplates
    
    Add a new Template DB model that will allow persisting our Elasticsearch templates to be shared between indexing and the server APIs.

    This also refactors the templates.py class which supports indexing, with the goal of integrating with the new DB table although the base commit stops short of implementing the integration to allow reviewing the basic refactoring separately.

1. Improve encapsulation

    Push some TemplateFile methods down into JsonFile and a new JsonToolFile subclass for cleaner encapsulation.

1. Integrate DB into template management
    
    The resolve() method now attempts to look up the root index name in the templates DB table. If found, it'll check modification timestamps to determine whether the on disk JSON (currently just resolved and stat-ed, not read) is more recent. If not, the cached DB information is used to create a TemplateFile object rather than reading from disk. If the DB object doesn't exist, it's created; if it exists but is older, it's updated.

1. The server Elasticsearch query APIs now use the Template DB to determine the index version and pattern rather than hardcoding these with the root index name.

### DIFF
--- a/lib/pbench/server/database/database.py
+++ b/lib/pbench/server/database/database.py
@@ -14,15 +14,16 @@ class Database:
             psql = config.get("Postgres", "db_uri")
             return psql
         except (NoSectionError, NoOptionError):
-            logger.error(
-                "Failed to find an [Postgres] section" " in configuration file."
-            )
+            if logger:
+                logger.error(
+                    "Failed to find a [Postgres] section in configuration file."
+                )
             return None
 
     @staticmethod
     def init_db(server_config, logger):
         # Attach the logger to the base class for models to find
-        if not hasattr(Database.Base, "logger"):
+        if logger and not hasattr(Database.Base, "logger"):
             Database.Base.logger = logger
 
         # WARNING:
--- a/lib/pbench/server/database/models/template.py
+++ b/lib/pbench/server/database/models/template.py
@@ -1,0 +1,236 @@
+import datetime
+from pathlib import Path
+from sqlalchemy import Column, Integer, String, event
+from sqlalchemy.exc import IntegrityError, SQLAlchemyError
+from sqlalchemy.sql.sqltypes import DateTime, JSON
+
+from pbench.server.database.database import Database
+
+
+class TemplateError(Exception):
+    """
+    This is a base class for errors reported by the Template class. It is
+    never raised directly, but may be used in "except" clauses.
+    """
+
+    pass
+
+
+class TemplateSqlError(TemplateError):
+    """
+    SQLAlchemy errors reported through Template operations.
+
+    The exception will identify the base name of the template index,
+    along with the operation being attempted; the __cause__ will specify the
+    original SQLAlchemy exception.
+    """
+
+    def __init__(self, operation: str, name: str):
+        self.operation = operation
+        self.name = name
+
+    def __str__(self) -> str:
+        return f"Error {self.operation} index {self.name!r}"
+
+
+class TemplateFileMissing(TemplateError):
+    """
+    Template requires a file name.
+    """
+
+    def __init__(self, name: str):
+        self.name = name
+
+    def __str__(self) -> str:
+        return f"Template {self.name!r} is missing required file"
+
+
+class TemplateNotFound(TemplateError):
+    """
+    Attempt to find a Template that doesn't exist.
+    """
+
+    def __init__(self, name: str):
+        self.name = name
+
+    def __str__(self) -> str:
+        return f"No template {self.name!r}"
+
+
+class TemplateDuplicate(TemplateError):
+    """
+    Attempt to commit a duplicate Template.
+    """
+
+    def __init__(self, name: str):
+        self.name = name
+
+    def __str__(self) -> str:
+        return f"Duplicate template {self.name!r}"
+
+
+class TemplateMissingParameter(TemplateError):
+    """
+    Attempt to commit a Template with missing parameters.
+    """
+
+    def __init__(self, name: str, cause: str):
+        self.name = name
+        self.cause = cause
+
+    def __str__(self) -> str:
+        return f"Missing required parameters in {self.name!r}: {self.cause}"
+
+
+class Template(Database.Base):
+    """
+    Identify a Pbench Elasticsearch document template
+
+    Columns:
+        id              Generated unique ID of table row
+        name            Index name key (e.g., "fio")
+        idxname         Base index name (e.g., "tool-data-fio")
+        template_name   The Elasticsearch template name
+        file            The source JSON mapping file
+        mtime           Template file modification timestamp
+        template_pattern The template for the Elasticsearch index name
+        index_template  The full index name template "p.v.i.y-m[-d]"
+        settings        The JSON settings payload
+        mappings        The JSON mappings payload
+        version         The template version metadata
+    """
+
+    __tablename__ = "templates"
+
+    id = Column(Integer, primary_key=True, autoincrement=True)
+    name = Column(String(255), unique=True, nullable=False)
+    idxname = Column(String(255), unique=True, nullable=False)
+    template_name = Column(String(255), unique=True, nullable=False)
+    file = Column(String(255), unique=False, nullable=False)
+    mtime = Column(DateTime, unique=False, nullable=False)
+    template_pattern = Column(String(255), unique=False, nullable=False)
+    index_template = Column(String(225), unique=False, nullable=False)
+    settings = Column(JSON, unique=False, nullable=False)
+    mappings = Column(JSON, unique=False, nullable=False)
+    version = Column(String(255), unique=False, nullable=False)
+
+    @staticmethod
+    def create(**kwargs) -> "Template":
+        """
+        A simple factory method to construct a new Template object and
+        add it to the database.
+
+        Args:
+            kwargs: any of the column names defined above
+
+        Returns:
+            A new Template object initialized with the keyword parameters.
+        """
+        template = Template(**kwargs)
+        template.add()
+        return template
+
+    @staticmethod
+    def find(name: str) -> "Template":
+        """
+        Return a Template object with the specified base name. For
+        example, find("run-data").
+
+        Args:
+            name: Base index name
+
+        Raises:
+            TemplateSqlError: problem interacting with Database
+            TemplateNotFound: the specified template doesn't exist
+
+        Returns:
+            Template: a template object with the specified base name
+        """
+        try:
+            template = Database.db_session.query(Template).filter_by(name=name).first()
+        except SQLAlchemyError as e:
+            Template.logger.warning("Error looking for {}: {}", name, str(e))
+            raise TemplateSqlError("finding", name)
+
+        if template is None:
+            raise TemplateNotFound(name)
+        return template
+
+    def __str__(self) -> str:
+        """
+        Return a string representation of the template
+
+        Returns:
+            string: Representation of the template
+        """
+        return f"{self.name}: {self.index_template}"
+
+    def _decode(self, exception: IntegrityError) -> Exception:
+        """
+        Decode a SQLAlchemy IntegrityError to look for a recognizable UNIQUE
+        or NOT NULL constraint violation. Return the original exception if
+        it doesn't match.
+
+        Args:
+            exception: An IntegrityError to decode
+
+        Returns:
+            a more specific exception, or the original if decoding fails
+        """
+        # Postgres engine returns (code, message) but sqlite3 engine only
+        # returns (message); so always take the last element.
+        cause = exception.orig.args[-1]
+        if cause.find("UNIQUE constraint") != -1:
+            Template.logger.warning("Duplicate template {!r}: {}", self.name, cause)
+            return TemplateDuplicate(self.name)
+        elif cause.find("NOT NULL constraint") != -1:
+            Template.logger.warning("Missing parameter {!r}, {}", self.name, cause)
+            return TemplateMissingParameter(self.name, cause)
+        Template.logger.exception("Other integrity error {}", cause)
+        return exception
+
+    def add(self):
+        """
+        Add the Template object to the database
+        """
+        try:
+            Database.db_session.add(self)
+            Database.db_session.commit()
+        except IntegrityError as e:
+            Database.db_session.rollback()
+            raise self._decode(e)
+        except Exception:
+            self.logger.exception("Can't add {} to DB", str(self))
+            Database.db_session.rollback()
+            raise TemplateSqlError("adding", self.name)
+
+    def update(self):
+        """
+        Update the database row with the modified version of the
+        Template object.
+        """
+        try:
+            Database.db_session.commit()
+        except IntegrityError as e:
+            Database.db_session.rollback()
+            raise self._decode(e)
+        except Exception:
+            self.logger.error("Can't update {} in DB", str(self))
+            Database.db_session.rollback()
+            raise TemplateSqlError("updating", self.name)
+
+
+@event.listens_for(Template, "init")
+def check_required(target, args, kwargs):
+    """
+    Listen for an init event on Template to validate that a filename was
+    specified; and automatically capture the file's modification timestamp
+    if it wasn't given.
+    """
+    if "file" not in kwargs:
+        raise TemplateFileMissing(kwargs["name"])
+
+    if "mtime" not in kwargs:
+        kwargs["mtime"] = datetime.datetime.fromtimestamp(
+            Path(kwargs["file"]).stat().st_mtime
+        )
--- a/lib/pbench/server/templates.py
+++ b/lib/pbench/server/templates.py
@@ -1,13 +1,15 @@
 import copy
-import glob
 import json
-import os
-import pyesbulk
 import re
 import sys
 
 from collections import Counter
+from datetime import datetime
 from pathlib import Path
+from typing import Any, AnyStr, Dict
+
+import pyesbulk
+from sqlalchemy.sql.sqltypes import JSON
 
 from pbench.common.exceptions import (
     BadDate,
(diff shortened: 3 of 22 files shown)


lib/pbench/server/database/database.py
```diff
@@ -14,15 +14,16 @@ def get_engine_uri(config, logger):
             psql = config.get("Postgres", "db_uri")
             return psql
         except (NoSectionError, NoOptionError):
-            logger.error(
-                "Failed to find an [Postgres] section" " in configuration file."
-            )
+            if logger:
+                logger.error(
+                    "Failed to find a [Postgres] section in configuration file."
+                )
```
dbutenhof, line 20: This is a bit of a hack; with a logger enabled, `pbench-report-status` was creating empty log files, which caused a lot of unnecessary gold file churn. I hacked the code up to allow initializing the DB without a logger just to keep it quiet.

webbnh: For a user command, logging to `stderr` (if no other logger was set up) would be good (and, probably better than no logging at all...).

dbutenhof: Yeah; this isn't a "user command"; it's a bridge command for bash scripts to reach the Python ES code to write server report documents, and those bash scripts run in "server" cron jobs. If we have a working server in the first place, our `crontab` configuration will ensure that we have a proper config file, or all this is pointless. So it didn't seem worth the effort to deal with the special case here.

lib/pbench/server/templates.py
```diff
@@ -301,12 +454,13 @@ def dump_idx_patterns(self):
         sys.stdout.flush()
 
     def dump_templates(self):
-        template_names = [name for name in self.templates]
+        templates = {t.template_name: t for t in self.templates.values()}
+        template_names = [name for name in templates]
         template_names.sort()
         for name in template_names:
```
webbnh, line 646: Or
```
        for name in sorted(templates.keys()):
```

webbnh: Ditto at lines 476-480.

lib/pbench/test/unit/server/test_template.py
```diff
@@ -0,0 +1,88 @@
...
+
+@pytest.fixture(autouse=True)
+def fake_mtime(monkeypatch):
+    """
+    Template's init event listener provides the file's modification date to
+    support template version control. For unit testing, mock the stat results
+    to appear at a fixed time.
+
+    Args:
+        monkeypatch ([type]): [description]
```
webbnh: Looks like this argument description is incomplete.

lib/pbench/test/unit/server/test_template.py
```diff
@@ -0,0 +1,88 @@
...
+        return stat_result(f)
+
+    with monkeypatch.context() as m:
+        m.setattr(Path, "stat", fake_stat)
+        yield
+
+
+class TestTemplate:
+    def test_construct(self):
+        """ Test dataset contructor
```
webbnh: s/contructor/constructor/

lib/pbench/test/unit/server/test_template.py
```diff
@@ -0,0 +1,88 @@
...
+            template_name="tname",
+            file="run.json",
+            index_template="run.{year}-{month}",
+            settings={"none": False},
+            mappings={"properties": None},
+            version=5,
+        )
+        template.add()
+        assert template.name == "run"
+        assert template.mtime == datetime.datetime(2021, 1, 29, 0, 0, 0)
```
webbnh: You should have a symbolic name for `datetime.datetime(2021, 1, 29, 0, 0, 0)` since you use it in more than one place and expect it to have the same value.

webbnh (commented):
OK, I've looked at the first commit....

lib/pbench/server/templates.py
```diff
@@ -256,13 +146,276 @@ def __init__(self, basepath, idx_prefix, logger, known_tool_handlers=None, _dbg=
...
             "desc": "Daily tool data for all tools land in indices"
             " named by tool; e.g. prefix.v0.tool-data-iostat.YYYY-MM-DD",
         },
     }
 
+    _fpat = re.compile(r"tool-data-frag-(?P<toolname>.+)\.json")
+
+    def __init__(
+        self,
+        prefix: str = None,
```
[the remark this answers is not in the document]
webbnh: I think I saw a spot where the code was omitting the `prefix` argument, so a default value is needed, but shouldn't it be `""` rather than `None`?

lib/pbench/server/database/models/template.py
```diff
@@ -0,0 +1,183 @@
...
+    template_name = Column(String(255), unique=True, nullable=False)
+    file = Column(String(255), unique=False, nullable=False, default=None)
+    mtime = Column(DateTime, unique=False, nullable=False)
+    index_template = Column(String(255), unique=False, nullable=False)
+    settings = Column(JSON, unique=False, nullable=False)
+    mappings = Column(JSON, unique=False, nullable=False)
+    version = Column(String(255), unique=False, nullable=False)
+
+    @staticmethod
+    def create(**kwargs) -> Database.Base:
```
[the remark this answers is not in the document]
dbutenhof: Yes, it was; however I discovered later that you can quote the name as a "forward reference" (pending "real" forward references in 3.9), and I didn't think to go back and fix these. I'll do that.

lib/pbench/server/templates.py
```diff
@@ -256,13 +146,276 @@ def __init__(self, basepath, idx_prefix, logger, known_tool_handlers=None, _dbg=
...
+        if self.mappings:
+            self.mappings.load()
+            if not self.skeleton:
+                try:
+                    self.version = self.mappings.json["_meta"]["version"]
+                except KeyError:
+                    raise MappingFileError(
+                        "{} mapping missing _meta field in {}".format(
+                            self.name, self.mappings.file
+                        )
```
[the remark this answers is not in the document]
webbnh: Nevermind:  I believe that you did this in the next commit.

lib/pbench/server/templates.py
```diff
@@ -56,6 +75,18 @@ def load(self):
                 raise JsonFileError("{}: {}".format(self.file, err))
         self.json = data
 
+    def merge(self, name: str, base: "JsonFile"):
+        """
+        Base class implementation for merge. The base class has no associated
+        skeleton document to merge but this no-op implementation allows merge
+        to be called on any JsonFile.
```
webbnh, line 106: Since this code ostensibly defines the interface for this function (although, it doesn't provide an implementation), it would be nice if the description of the function were more complete.  (E.g., what is an implementation of this function expected to _do_ with the property and the base?  Not having looked at the subclass(es) implementation(s) yet, I'm having a hard time imagining how this is going to work....)

lib/pbench/server/templates.py
```diff
@@ -72,6 +103,66 @@ def add_mapping(self, name: str, body: Dict[AnyStr, Any]):
...
+            MappingFileError: Mapping file doesn't contain proper version
+
+        Returns:
+            Template file metadata version
+        """
+        try:
+            version = self.meta["version"]
+        except KeyError as e:
+            raise MappingFileError(f"mapping missing {e} in {self.file}: {self.meta}")
+        return version
```
webbnh, line 176: I kind of wish that this were refactored into a helper function which was passed the JSON blob as a parameter -- that is, it'd be up to the caller to decide whether it is in `self.meta` or `self.json["_meta"]`, and it'd up to the subroutine to extract it and deal with the error.

It's not quite as clean/effective as I'd like it to be, because the access to the "_meta" key is done outside the helper, but it would be DRYer.

lib/pbench/server/templates.py
```diff
@@ -72,6 +103,66 @@ def add_mapping(self, name: str, body: Dict[AnyStr, Any]):
...
+        Tool mapping files define a sub-document in the Elasticsearch template
+        specific to the tool (e.g., "iostat") which exists within a common
+        base skeleton. The merge of the two includes elevating the "_meta"
+        version from the sub-document into the base document.
+
+        Here we extend the JsonFile load method to remove the sub-document
+        metadata and save it to be spliced into the base document by merge.
+        """
+        super().load()
+        self.meta = self.json["_meta"]
```
webbnh: Do we need to be concerned about the `"_meta"` key not existing?

lib/pbench/server/templates.py
webbnh: I don't understand:  if this function is called, and the caller specifies a base document, the result will be a no-op...won't that come as a surprise to the caller??

I.e., if `base` is `None`, then a no-op result would probably be a good thing; but if `base` is not `None`, then it seems like this method should raise [`NotImplementedError`](https://docs.python.org/3/library/exceptions.html#NotImplementedError).

lib/pbench/server/templates.py
```diff
@@ -256,13 +146,276 @@ def __init__(self, basepath, idx_prefix, logger, known_tool_handlers=None, _dbg=
...
+            tool_mapping = copy.deepcopy(self.base.mappings.json)
+
+            # This is ugly: we get the version from the tool-specific overlay;
+            # but that becomes a subdocument of the skeleton, and we need the
+            # version there. So we delete from one and create in the other.
+            del self.mappings.json["_meta"]
+            tool_mapping["_meta"] = dict(version=self.version)
+            sub_map = self.mappings.json
+            self.mappings.json = tool_mapping
+            self.add_mapping(self.name, sub_map)
```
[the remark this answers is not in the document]
webbnh: Nevermind:  I believe that you've reworked this code in subsequent commits, so this is moot.

lib/pbench/server/templates.py
```diff
@@ -158,103 +249,90 @@ def __init__(
...
             prefix: Pbench index prefix string
             mappings: JSON document description. Defaults to None.
             settings: JSON index settings. Defaults to None.
-            base: Referenced base settings and/or mappings. Defaults to None.
             tool: Tool mapping files stitch together mapping JSON from the
-                main and base template objects using a parsed "tool" name.
-            skeleton: The JSON mappings file does not have a version.
+                main and skeleton templates using a parsed "tool" name.
+            skeleton: A versionless skeleton mappings file to merge with the tool
+                mappings.
```
webbnh: It would probably be better to have the descriptions in the same order as the function signature.

lib/pbench/server/templates.py
```diff
@@ -256,13 +146,276 @@ def __init__(self, basepath, idx_prefix, logger, known_tool_handlers=None, _dbg=
...
+        A "base" template may also be linked, from which a final template will
+        be built using shared settings and optionally base mappings.
+
+        Args:
+            prefix: Pbench index prefix string
+            mappings: JSON document description. Defaults to None.
+            settings: JSON index settings. Defaults to None.
+            base: Referenced base settings and/or mappings. Defaults to None.
+            tool: Tool mapping files stitch together mapping JSON from the
+                main and base template objects using a parsed "tool" name.
```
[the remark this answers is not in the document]
webbnh: This text is more of a declarative statement than a description of the parameter....

lib/pbench/server/templates.py
```diff
@@ -158,103 +249,90 @@ def __init__(
...
         else:
-            # If this is just a holder for a shared settings file, there's not
-            # much else we can do with it.
-            self.key = None
-            self.name = None
-            self.index_info = None
-        self.base = base
+            self.mappings = JsonFile(mappings)
+            self.name = mappings.stem
+            self.key = self.name
```
webbnh, line 328: Does this code behave well if `mappings` is `None`?

lib/pbench/server/templates.py
```diff
@@ -256,13 +146,276 @@ def __init__(self, basepath, idx_prefix, logger, known_tool_handlers=None, _dbg=
...
+        self.prefix = prefix
+        self.mappings = JsonFile(mappings) if mappings else None
+        self.settings = JsonFile(settings) if settings else None
+        if mappings:
+            if tool:
+                self.key = tool
+                m = self._fpat.match(mappings.name)
+                self.name = m.group("toolname")
+            else:
+                self.name = mappings.stem if mappings else None
```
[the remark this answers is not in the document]
webbnh: Nevermind:  the code reorg in the subsequent commit addressed this.

lib/pbench/server/templates.py
```diff
@@ -256,13 +146,276 @@ def __init__(self, basepath, idx_prefix, logger, known_tool_handlers=None, _dbg=
...
+        """
+        self.load()
+
+        # FIXME: It's ugly reaching into the JsonFile object from here; this
+        # ought to be refactored into JsonFile somehow.
+        if self.tool:
+            tool_mapping = copy.deepcopy(self.base.mappings.json)
+
+            # This is ugly: we get the version from the tool-specific overlay;
+            # but that becomes a subdocument of the skeleton, and we need the
```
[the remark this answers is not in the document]
webbnh: Nevermind:  the subsequent commit reworks this.

lib/pbench/server/templates.py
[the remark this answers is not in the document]
webbnh: A value of `None` when formatted as a string will yield `"None"` which is probably not what you want....

lib/pbench/server/templates.py
```diff
@@ -272,45 +350,48 @@ def add_mapping(self, name: str, body: Dict[AnyStr, Any]):
...
 
-            # This is ugly: we get the version from the tool-specific overlay;
-            # but that becomes a subdocument of the skeleton, and we need the
-            # version there. So we delete from one and create in the other.
-            del self.mappings.json["_meta"]
-            tool_mapping["_meta"] = dict(version=self.version)
-            sub_map = self.mappings.json
-            self.mappings.json = tool_mapping
-            self.add_mapping(self.name, sub_map)
+        Args:
```
webbnh: Having the "given" list not match the function parameter list is jarring.

Having an empty `Args:` section is worse.  ;-)

lib/pbench/server/templates.py
```diff
@@ -272,45 +350,48 @@ def add_mapping(self, name: str, body: Dict[AnyStr, Any]):
...
+            The Elasticsearch index into which the source document will be indexed.
+        """
+        version = self.version
+        idxname = self.idxname
+        index_template = self.index_info["template"]
+
+        try:
+            ts_val = source["@timestamp"]
+        except KeyError:
+            raise BadDate(f"missing @timestamp in a source document: {source!r}")
```
webbnh, line 478: It's a nit, but it would be better to fetch the missing key from the exception than to hard-code it as `@timestamp`.

lib/pbench/server/templates.py
```diff
@@ -272,45 +350,48 @@ def add_mapping(self, name: str, body: Dict[AnyStr, Any]):
...
+        except TypeError as e:
+            raise JsonFileError(
+                f"Failed to generate index name, {e}, source: {source!r}"
             )
```
webbnh: Under what circumstances would line 379 generate a `TypeError`?  (Is one of us missing something?)

lib/pbench/server/templates.py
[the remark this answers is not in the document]
webbnh: OK, with the second commit, I no longer see a call where `prefix` is unspecified; so I think you probably should make this be a required parameter and object if it comes in with a "False" value.

lib/pbench/server/templates.py
webbnh: GitHub is being weird -- it posted my review comments individually, despite giving the appearance that it was saving them up as "a review", so I didn't get a chance to correct them before they posted.

GitHub is being weird -- it posted my review comments individually, despite giving the appearance that it was saving them up as "a review", so I didn't get a chance to correct them before they posted.

> Not having looked at the subclass(es) implementation(s) yet, I'm having a hard time imagining how this is going to work...

I see now that the subclass implementation draws the rest of the information required for the merge from instance attributes, so now the interface makes more sense.  But, it would still be very helpful to have a more complete description here.

dbutenhof: In fact the final code only calls merge if there's a skeleton document to be merged, so there's no longer any need for this to be a nop on the base class. I'll add a `raise NotImplementedError` to make this clear.

lib/pbench/server/templates.py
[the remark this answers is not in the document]
dbutenhof: I think I was trying to make clear that the parameters could be referenced by keyword. In retrospect, that doesn't really help. I'm not going to add null checks everywhere, because that seems wasted effort, but I will remove the `= None` where it's not appropriate.

lib/pbench/server/templates.py
dbutenhof: Probably not; but perhaps this would be more clear if I really do reject attempts to construct a `TemplateFile` with missing parameters. Sigh.

lib/pbench/server/templates.py
```diff
@@ -158,103 +249,90 @@ def __init__(
...
-            # If this is just a holder for a shared settings file, there's not
-            # much else we can do with it.
-            self.key = None
-            self.name = None
-            self.index_info = None
-        self.base = base
+            self.mappings = JsonFile(mappings)
+            self.name = mappings.stem
+            self.key = self.name
+        self.index_info = self.index_patterns[self.key]
```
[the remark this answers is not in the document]
dbutenhof: No; the question becomes whether it's worth the effort of checking. I decided not. Like most facets of reality, you end up making me question this. ;-)

lib/pbench/server/templates.py
dbutenhof: I've always been reluctant to trust too far in the exact textual representation of an exception; since I originally coded this I've grown a bit more casual about `KeyError` values...

lib/pbench/server/templates.py
```diff
@@ -348,14 +360,58 @@ def add_mapping(self, name: str, body: Dict[AnyStr, Any]):
...
+        if not template:
+            template = Template(
+                name=self.name,
+                template_name=self.template_name,
+                file=str(self.mappings.file),
+                index_template=self.index_pattern,
+                settings=self.settings.json,
+                mappings=self.mappings.json,
+                version=self.version,
+            )
```
webbnh: You're not specifying the modification time.  I think this creates an opportunity for a pratfall:  if we do a run with an old version of a template, and then we go "oops" and try to do a run with the correct version of the template, the database won't get updated unless the correct version has a newer date than _when we did the initial run_.  So, since we have the template file here, I think we should use _its_ date.

server/bin/pbench-cull-unpacked-tarballs.py
```diff
@@ -327,6 +328,10 @@ def main(options):
 
     logger = get_pbench_logger(_NAME_, config)
 
+    # We're going to need the Postgres DB to track dataset state, so setup
+    # DB access.
+    Database.init_db(config, logger)
```
webbnh, line 333: Does this need to be modified per Nikhil's recent work?  (If not, does it need a big ol' comment explaining why it doesn't conform to the requirements of using the common entry point?)

webbnh: It feels weird that this line needs to be added here when no other modifications are required to this file.  Either this change was unnecessary, or something that this file already calls has now placed a new requirement on its callers.  If it's the latter, shouldn't that requirement be encapsulated inside the called module (e.g., via a module initializer or some other singleton implementation) and not imposed on existing callers?

Ditto for `server/bin/*.py`.

lib/pbench/test/unit/server/test_template.py
dbutenhof: I believe I use it once in the fixture and once in a test; I considered using the same constant, but I actually like having the test check against the literal rather than blindly referencing the mock, as a consistency double-check.

webbnh (commented):
Done with the third commit.

Aside from the typical nits and small suggestions, there are two things:

- Something feels wrong about having to add the database initialization to each of the tools.
- I think you need to explicitly specify the modification time when creating (or updating) the template database entry.

Continuing....

lib/pbench/server/templates.py
dbutenhof: Too much refactoring with deferred cleanup; `toolname` isn't even used at this level.

lib/pbench/server/templates.py
dbutenhof: Interesting question. This came from the original code, but I suspect became irrelevant along the way and I failed to notice.

lib/pbench/server/templates.py
```diff
@@ -283,9 +291,14 @@ def __init__(
...
             self.key = self.name
         self.index_info = self.index_patterns[self.key]
+        self.idxname = self.index_info["idxname"].format(tool=self.name)
         self.version = None
         self.tool = tool
         self.skeleton = skeleton
+        if skeleton:
+            self.modified = max(self.mappings.modified, self.skeleton.modified)
+        else:
+            self.modified = self.mappings.modified
```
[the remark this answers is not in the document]
dbutenhof: I know you dislike indentation, but I think in this case the expanded syntax is a lot more clear.

lib/pbench/server/templates.py
```diff
@@ -348,14 +360,58 @@ def add_mapping(self, name: str, body: Dict[AnyStr, Any]):
...
         """
+        Check the on-disk template mapping files' modification dates against
+        the templates DB. Only if the template's base index name doesn't exist
+        in the DB, or if the DB object is older than the on-disk template do
+        we load the mapping files from disk.
+        """
+        try:
+            template = Template.find(self.name)
+        except TemplateNotFound:
+            template = None
```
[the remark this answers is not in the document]
dbutenhof: Of the three calls, this saves me from a `None` check in the other two; this is the only case where failing to find the template in the DB is acceptable. (Because we'll then create it.)

lib/pbench/server/templates.py
```diff
@@ -256,13 +146,276 @@ def __init__(self, basepath, idx_prefix, logger, known_tool_handlers=None, _dbg=
...
+    """
+
+    def __init__(self, basepath, idx_prefix, logger, known_tool_handlers=None, _dbg=0):
+        # Determine the location of the mapping and settings files. We expect
+        # the "basepath" parameter to be the PbenchServerConfig BINDIR attr,
+        # which will be something like /opt/pbench-server/bin
+        base_dir = Path(basepath).parent / "lib"
+        mapping_dir = base_dir / "mappings"
+        # Where to find the settings
+        setting_dir = base_dir / "settings"
```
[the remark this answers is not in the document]
dbutenhof: This is difficult to parse since your line numbers don't match the code against which GitHub displays it. I think that, at the time, I thought that the `config.BASEPATH` reference was useful, and after explaining that, failing to explain the equivalent `setting_dir` felt awkward. However since this is a parameter and not a direct reference to the config object, you're probably right that it's all unnecessary.

lib/pbench/server/templates.py
webbnh: Yeah:  hmm....  I figured I would offer.

lib/pbench/server/templates.py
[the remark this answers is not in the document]
webbnh: Heh.

The problem is, the inputs originate in external files, which are subject to all sorts of corrosion and damage, so I tend to trust them less than code (which I try not to trust more than I have to...).

server/bin/pbench-cull-unpacked-tarballs.py
dbutenhof: Yes, this will need to be changed once Nikhil's change it in. (Or ideally when he merges this, if this one goes in first.) For now, this is hos it's done.

Beyond that, yes, I've often thought that our whole config setup is broken because of the dependency on the environment variable _or_ a `-C` option; basically, we can only get at it from the outer layer, which means it needs to be passed in.

And, yes, I have considered making a `PbenchServerConfig` singleton which can then be referenced from anywhere without passing it in ... I haven't quite gotten around to tackling that. (Not to mention schools of thought, which seem common in Python, that singletons are just hidden globals and therefore bad... a point of view that's not entirely incorrect.)

I'd like to be able to have inner code, for example, grab their own `Logger` ... but as our logging infrastructure requires a `PbenchServerConfig`, which requires this outer context, we end up doing things like unnaturally hanging loggers off classes.

You're welcome to tackle cleaning this up. I'm not doing it here!

lib/pbench/server/templates.py
```diff
@@ -25,17 +26,24 @@ class JsonFile:
...
+        if file:
+            self.file = file.resolve(strict=True)
+            self.modified = datetime.fromtimestamp(self.file.stat().st_mtime)
+            self.json = None
+
+    @staticmethod
+    def raw_json(json: Dict[AnyStr, Any]) -> "JsonFile":
+        new = JsonFile()
+        new.json = json
+        return new
```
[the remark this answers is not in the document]
webbnh: I see that, in the next commit, you're initializing them in the c'tor.

However, I'm still wondering if `raw_json()` should be setting the `modified` value (perhaps taking a parameter for the value, so that the caller can use `template.mtime`?).

lib/pbench/server/templates.py
```diff
@@ -174,20 +195,15 @@ def merge(self, name: str, base: JsonFile):
...
-    This doesn't load the files, but does fetch the modification timestamp of
-    the mapping file, which will be used to determine whether a version already
-    captured in the database is up to date. (If not, the new version will be
-    loaded and the DB will be updated.)
-
-    NOTE/FIXME: The type hint on the "base" link to another TemplateFile object
-    is a forward reference since the type has not been fully defined. This will
-    be a default feature of Python 3.10; in the meantime, the quoted (deferred)
-    type hint works.
+    and settings file, as a complete template. Tool templates are build from a
```
webbnh: s/build/built/

lib/pbench/server/templates.py
webbnh: Nevermind:  you fixed this in a subsequent commit.

lib/pbench/test/unit/server/test_templates.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def test_json_load(self, monkeypatch, fake_resolve, fake_mtime):
+        json = JsonFile(Path("mapping.json"))
+        with monkeypatch.context() as m:
+
+            def fake_open(self: Path, mode: str = "r"):
+                return io.StringIO('{"_meta": {"version": 6}}\n')
+
+            m.setattr(Path, "open", fake_open)
+            # m.setattr(Path, "open", lambda self, mode : io.StringIO('{"_meta": {"version": 6}}\n'))
```
webbnh: Remove this line before merging?  Ditto line 109, et al.

(A lambda seems like a nice idea, here...you couldn't get it to work?)

webbnh (commented):
More little niggly bits.

On to the next commit!

lib/pbench/server/templates.py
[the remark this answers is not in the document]
webbnh: The line numbers look OK if you read this in the "Conversation" tab.  :-)

(They would probably look OK, too, reading it in "Files changed" tab if you had the right commit selected....)

lib/pbench/server/templates.py
```diff
@@ -25,17 +26,24 @@ class JsonFile:
...
         Initialize a JSON file object by resolving the file name and reading
         the modification date.
 
         Args:
             file: JSON file path
         """
-        self.file = file.resolve(strict=True)
-        self.modified = datetime.fromtimestamp(self.file.stat().st_mtime)
-        self.json = None
+        if file:
```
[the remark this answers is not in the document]
webbnh: In the commit that I was reviewing at the time, the code was structured such that, if `file` was "False", it skipped everything and fell out of the bottom of the routine...so, inverting the test and explicitly `return`'ing seemed like a better choice.  However, in a later commit, you changed the code to initialize the attributes properly in both cases.

(I thought I responded and resolved this conversation...but perhaps GitHub played another trick on me.)

server/bin/pbench-cull-unpacked-tarballs.py
webbnh: > basically, we can only get at it from the outer layer, which means it needs to be passed in.

Ah.  OK.  If/when this code gets Click-ified, perhaps we can put the database init in a common location (on the outside) and only have to do it once.

Re singletons, they are, indeed, very much like hidden globals...but I'm not sure that that means they are necessarily bad.  :-)

lib/pbench/test/unit/server/test_templates.py
dbutenhof: Yeah, couldn't get it to work. I gave up. And the problem with having posted this with staged commits is that you keep commenting on stuff in earlier commits that don't exist anymore. I think my strategy was flawed. ;-)

server/bin/pbench-index.py
```diff
@@ -102,6 +104,15 @@ def main(options, name):
 
     idxctx = None
     try:
+        # We're going to need the Postgres DB to track dataset state, so setup
+        # DB access. We need to do this before we create the IdxContext, since
+        # that will need the template DB; so we create a PbenchServerConfig and
+        # Logger independently.
+        config = PbenchServerConfig(options.cfg_name)
+        logger = get_pbench_logger(name, config)
+        Database.init_db(config, logger)
```
portante, line 113: Why not just perform the `init_db()` at the appropriate place in the `IdxContext()` constructor?

portante (commented):
Partial review ... more to come.

server/bin/pbench-index.py
dbutenhof: I think there's already way too much stuff crammed into `IdxCtx` ... but mostly if we start doing this sort of initialization on interior classes we run the risk of having to deal with double-initialization problems like Nikhil ran into in his unit tests. I wanted to avoid that by keeping it at the outer command level, which really ought to own the configuration environment.

(I really think we should be passing the logger and config into `IdxCtx`, too, instead of having it do all that itself; but I didn't want to tackle that on top of everything else.)

portante: Yes, but this generates a second config object (I think the logger instance is the same).

This is half-way in between ...

dbutenhof: True, but neither are all that expensive. I'd rather pull them out of `IdxCtx` than bring this in, if it comes to that, but is this really so bad?

portante: See https://github.com/dbutenhof/pbench/pull/1.

lib/pbench/test/unit/server/test_templates.py
```diff
@@ -51,7 +51,7 @@ def fake_open(self: Path, mode: str = "r"):
         Return an IO stream
 
         Args:
-            file: filename
+            mode: file mode (assumed to be "r")
```
webbnh, line 54: It would probably be worth asserting that `mode` _is_ `"r"`....

webbnh (commented):
OK, I think I've finished my first pass.

lib/pbench/test/unit/server/test_templates.py
webbnh: I think your strategy was good, but, perhaps, your execution contained a few flaws.  :-)

server/bin/pbench-index.py
webbnh: OK, a PR to a PR...now _that_ was cool.  (\*mind-blown\*)

webbnh (changes requested):
Outstanding items:
- Potential ungraceful failure modes: [1](https://github.com/distributed-system-analysis/pbench/pull/2255#discussion_r625369836), [2](https://github.com/distributed-system-analysis/pbench/pull/2255#discussion_r625420212), [3](https://github.com/distributed-system-analysis/pbench/pull/2255#discussion_r625904830)
- And whatever of the other unresolved conversations you want to attend to.

lib/pbench/server/templates.py
```diff
@@ -609,29 +697,33 @@ def dump_templates(self):
...
             idxname = None
         template_names = {t.template_name: t for t in self.templates.values()}
-        names = list(template_names.keys())
-        names.sort()
         successes = retries = 0
         beg = end = None
-        for name in names:
+        for name in sorted(template_names.keys()):
             template = template_names[name]
             if idxname is not None and not target_name == template.idxname:
```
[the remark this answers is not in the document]
dbutenhof: Evolution moves through a series of forms that work. The shape of the `self.templates` dict and the `TemplateFile` object changed a number of times, and in many of the mutations I struggled to make all of these methods retain the same behavior. This indirection got left behind because some form of it was necessary at one point and when it mutated back into this useless form it didn't break anything and I didn't notice.

lib/pbench/server/database/models/template.py
```diff
@@ -144,26 +167,51 @@ def __str__(self):
...
+            Database.db_session.rollback()
+            if cause.find("UNIQUE constraint") != -1:
+                Template.logger.warning("Duplicate template {!r}: {}", self.name, cause)
+                raise TemplateDuplicate(self.name)
+            elif cause.find("NOT NULL constraint") != -1:
+                Template.logger.warning("Missing parameter {!r}, {}", self.name, cause)
+                raise TemplateMissingParameter(self.name, cause)
+            else:
+                Template.logger.exception("Other integrity error {}", cause)
+                raise
```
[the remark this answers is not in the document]
dbutenhof: I'm reluctant mostly because that adds another level of traceback to obscure the origin. At least, unless I simply return an exception to raise here, which defeats the clarity of the [re-]`raise` at the end. I decided I didn't really like it either way, although I'm also not enthusiastic about duplicate code.

lib/pbench/server/templates.py
```diff
@@ -292,9 +331,16 @@ def __init__(
...
+            tool: The key in the index_patterns dict differs from the actual
+                tool name: e.g., all "tool" documents share the "tool-data" tool
+                name.
+        """
+        # NOTE: we really don't want prefix to be None, but it is in some CLI
+        # unit tests and it's not worth the churn to fix that.
+        if mappings is None or settings is None:
+            raise TemplateError(
+                f"Missing parameters MAPPINGS={mappings}, SETTINGS={settings}"
+            )
```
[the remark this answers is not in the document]
dbutenhof: The whole `None` thing was because I started out with chained `TemplateFile` objects where you could build a "base" `TemplateFile` with just `settings`, for example, that would then be used by a series of `TemplateFile` objects that referred to it as a `base`. When I removed that obfuscation, some obfuscation droppings remained which I'd simply grown used to stepping over...

lib/pbench/server/templates.py
```diff
@@ -428,12 +487,13 @@ def resolve(self):
             template.mtime = self.modified
```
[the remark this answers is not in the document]
dbutenhof: No? Or at least, I put a stake in the ground that I'm not going to try to get that weird. `PbenchTemplates` is building these documents quite deliberately; if we change file names or introduce new ones, that `__init__` will need to be recoded, and I don't expect to ever support doing so dynamically.

lib/pbench/server/templates.py
```diff
@@ -558,6 +637,14 @@ def resolve(self):
...
+        List the registered index template patterns to the user.
+
+        NOTE: This doesn't do a simple traversal of the templates dictionary
+        in order to retain an output style identical to the previous, with the
+        generic "tool-data" description appearing only once following the last
+        tool data pattern.
+        """
         patterns = TemplateFile.index_patterns
         pattern_names = [idx for idx in patterns]
         pattern_names.sort()
```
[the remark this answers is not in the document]
dbutenhof: The tool_name loop I barely touched; I'd really tried to kill it entirely but was unable to come up with a good way to reproduce the previous behavior without it. On the other hand, I particularly hate the ugly black stacked conditional comprehension; so now that I'm no longer tearing out hair just getting it to work, I'll take that cleanup suggestion.

lib/pbench/server/database/models/template.py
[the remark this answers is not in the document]
webbnh: > that adds another level of traceback to obscure the origin

That's a fair point.

> unless I simply return an exception to raise here, which defeats the clarity of the [re-]raise at the end.

That's a capital suggestion!  If the helper function takes the original exception as an argument and returns an exception for the caller to raise, then, if it cannot come up with something better, the function can return the original exception -- if the caller raises that from within the `except` block which defined it, then I think that's the equivalent of a bare `raise` statement, which is exactly what we want!

lib/pbench/server/templates.py
```diff
@@ -348,14 +360,58 @@ def add_mapping(self, name: str, body: Dict[AnyStr, Any]):
...
+        """
+        try:
+            template = Template.find(self.name)
+        except TemplateNotFound:
+            template = None
+
+        # If we found a match that's newer, use it
+        if template and template.mtime >= self.modified:
+            self.update(template)
+            return
```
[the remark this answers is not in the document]
webbnh: This change didn't turn out to be as cool as I hoped it would be:  the problem is, if control doesn't return here, then we need to do the `self.load()` call on both the exception and non-exception paths -- if it weren't for that, then the `except` block could do the database add, and we would never have to set `template` to `None` or test it.  Oh, well....

lib/pbench/server/database/models/template.py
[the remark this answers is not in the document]
dbutenhof: I was worried that `raise e` might not be exactly that same as `raise`. It seems there's some controversy about this point but the consensus seems to be that the original Python 2 problem (overwriting the exception origin) were fixed in Python 3 and this shouldn't be a problem.

I'm tired of playing leapfrog on this PR today, but if I don't get my approvals I guess I take a whack at this tomorrow. ;-)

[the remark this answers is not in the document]
webbnh: > I don't get my approvals I guess I take a whack at this tomorrow. ;-)

That creates a perverse incentive....  :-D

webbnh (approved):
I think all of my concerns are resolved, although, if you want to do that exception decoder, I certainly won't object.

dbutenhof:
> I think all of my concerns are resolved, although, if you want to do that exception decoder, I certainly won't object.

Yes, you know how to throw down the gauntlet. ;-)

lib/pbench/server/database/models/template.py
```diff
@@ -165,6 +165,31 @@ def __str__(self) -> str:
...
+        Args:
+            exception: An IntegrityError to decode
+
+        Returns:
+            a more specific exception, or the original if decoding fails
+        """
+        # Postgres engine returns (code, message) but sqlite3 engine only
+        # returns (message); so always take the last element.
+        cause = exception.orig.args[-1]
+        Database.db_session.rollback()
```
webbnh, line 183: This line is duplicated in the callers, so it can/should be removed from here.

webbnh (changes requested):
Almost perfect!  ;-)

lib/pbench/server/database/models/template.py
dbutenhof: Well, dang. And now this really is going to wait until tomorrow, because when I changed that I lost a bunch of unit tests in ways that make absolutely no sense at all. It must be the rebase, always I was sure I'd run it before the last commit; but I'm not dealing with that now.

agent/util-scripts/gold/test-client-tool-meister/test-53.txt
```diff
@@ -1,8 +1,6 @@
 +++ Running test-53 test-client-tool-meister 
 "mpstat" tool is now registered for host "testhost.example.com" in group "default"
 "dcgm" tool is now registered for host "testhost.example.com" in group "default"
-pbench-tool-data-sink: connected to redis server localhost:17001
-pbench-tool-meister: connected to redis server localhost:17001
```
portante, line 5: Why do you have this change in your PR?

portante (changes requested):

dbutenhof:
NOTE: I didn't squash this because I'm anticipating another rebase momentarily when the pbench-results-push fix goes in. ;-)

agent/util-scripts/gold/test-client-tool-meister/test-53.txt
dbutenhof: It's not there now. Transient rebase artifact?? Weird.

webbnh (dismissed):
As far as I can tell, this is now perfect.

portante:
> As far as I can tell, this is now perfect.

With a statement like that, how could I ask for further changes in a review?

webbnh:
> With a statement like that, how could I ask for further changes in a review?

By pointing out something that I missed!

dbutenhof:
> > With a statement like that, how could I ask for further changes in a review?
> 
> By pointing out something that I missed!

Ah, the old formula of tempting fate... or is it whistling in the graveyard?

Then again, I don't see two approvals. ;-)

webbnh (dismissed):
GitHub thinks that nothing has changed since my last review, so it must still be perfect.
(Did something change other than a rebase?)

dbutenhof:
> GitHub thinks that nothing has changed since my last review, so it must still be perfect.
> (Did something change other than a rebase?)

Nope; just a rebase. And, I suppose, a subtle hint to someone or other that it needs a second approval. ;-)

webbnh (approved):
GitHub still claims I've seen it all....

portante (approved):

PR #2256, opened 2021-04-28
Replace use of become/become_user with remote_user

### LINKED ISSUE #2077: Default pbench_agent_install.yml playbook does not work for inventory files with only remote hosts and non-root users
The documentation around the pbench-agent ansible roles is a bit confusing.  Once you install the ansible galaxy role via:

```
ansible-galaxy collection install pbench.agent
```

You have to `export` the `bash` variable, `ANSIBLE_ROLES_PATH` with the installed path:

```
export ANSIBLE_ROLES_PATH=$HOME/.ansible/collections/ansible_collections/pbench/agent/roles:$ANSIBLE_ROLES_PATH
```

While the `agent/ansible/README` talks about that, I was not able to quickly find that in the regular documentation.

The `ANSIBLE_ROLES_PATH` variable setting allows the role to be found so that the following command will execute the example vanilla pbench agent installation playbook (found at [pbench_agent_install.yml](https://github.com/distributed-system-analysis/pbench/blob/master/agent/ansible/pbench_agent_install.yml)):

```
ansible-playbook -i ~/.config/Inventory/myhosts.inv ./pbench_agent_install.yml
```

However, when the inventory file only contains remote hosts, and on the local host you can't su/sudo to "root", the default playbook we provide does not work.  The two lines:

```
become: true
become_user: root
```

Should be changed to be just:

```
remote_user: root
```

And then this playbook will work for inventory files which only have remote hosts.

ndokos:
Fixes #2077

Do not use `become` or `become_user` in playbooks. Replace them
with `remote_user`.

### DIFF
--- a/agent/ansible/pbench_agent_config.yml
+++ b/agent/ansible/pbench_agent_config.yml
@@ -1,8 +1,7 @@
 ---
 - name: install pbench-agent config and key files
   hosts: servers
-  become: yes
-  become_user: root
+  remote_user: root
 
   # The default value ('production') can be overridden by cenv, a host-specific
   # inventory variable.
--- a/agent/ansible/pbench_agent_install.yml
+++ b/agent/ansible/pbench_agent_install.yml
@@ -1,8 +1,7 @@
 ---
 - name: install pbench-agent
   hosts: servers
-  become: yes
-  become_user: root
+  remote_user: root
 
   # The default value ('production') can be overridden by cenv, a host-specific
   # inventory variable.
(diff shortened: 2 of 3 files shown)


npalaska (approved):

dbutenhof (approved):
The changes look OK, but this seems to be a likely sticking point. At heart, is it "more secure" (and/or more "Linux adminish") to

1. Provide remote login to root, or,
2. Allow sudo to root

It's not hard to find controversy on the dynamics here. My concern is: are we stepping out of one sinkhole into another?

portante (approved):

PR #2263, opened 2021-04-29
Fix standalone db.init issue

npalaska:
- Creates a utility init_db method in `lib/pbench/server/api/__init__.py` that other functionalities in our code can use to initialize the standalone db access and make sure all the models are imported before the call to Database.init_db.

### DIFF
--- a/lib/pbench/server/api/__init__.py
+++ b/lib/pbench/server/api/__init__.py
@@ -18,6 +18,7 @@ from pbench.server.api.resources.graphql_api import GraphQL
 from pbench.server.api.resources.endpoint_configure import EndpointConfig
 from pbench.common.logger import get_pbench_logger
 from pbench.server.api.resources.query_apis.elasticsearch_api import Elasticsearch
+from pbench.server.database import init_db
 from pbench.server.database.database import Database
 from pbench.server.api.resources.query_apis.controllers_list import ControllersList
 from pbench.server.api.resources.query_apis.datasets_list import DatasetsList
@@ -134,7 +135,7 @@ def create_app(server_config):
     register_endpoints(api, app, server_config)
 
     try:
-        Database.init_db(server_config=server_config, logger=app.logger)
+        init_db(server_config=server_config, logger=app.logger)
     except Exception:
         app.logger.exception("Exception while initializing sqlalchemy database")
         sys.exit(1)
--- a/lib/pbench/server/api/resources/users_api.py
+++ b/lib/pbench/server/api/resources/users_api.py
@@ -284,7 +284,7 @@ class Logout(Resource):
             else:
                 self.logger.debug("User {} logged out", user.username)
         else:
-            self.logger.debug("User logout with invalid or expired token")
+            self.logger.info("User logout with invalid or expired token")
 
         return "", 200
 
--- a/lib/pbench/server/database/__init__.py
+++ b/lib/pbench/server/database/__init__.py
@@ -1,0 +1,26 @@
+"""
+For any new database model added in the models directory
+an import statement of the same is required here.
+"""
+from pbench.server.database.models.active_tokens import ActiveTokens  # noqa F401
+from pbench.server.database.models.users import User  # noqa F401
+from pbench.server.database.models.tracker import Dataset, Metadata  # noqa F401
+from pbench.server.database.models.template import Template  # noqa F401
+
+from pbench.server.database.database import Database
+
+
+def init_db(server_config, logger):
+    """
+    Utility method for initializing the database.
+
+    In order to register all the tables properly in our database, the db models
+    need to be imported before the call to the Database.init_db() function.
+
+    A lot of independent functionality in our code depends on database
+    access without creating the flask app.  Invoking init_db() from here
+    makes sure all models are imported properly; therefore, all of the code
+    where we need standalone db access should invoke this function instead
+    of directly invoking Database.init_db().
+    """
+    Database.init_db(server_config=server_config, logger=logger)
--- a/server/bin/pbench-import-datasets.py
+++ b/server/bin/pbench-import-datasets.py
@@ -33,9 +33,9 @@ from typing import Generator
 from pbench import BadConfig
 from pbench.common.logger import get_pbench_logger
 from pbench.server import PbenchServerConfig
+from pbench.server.database import init_db
 from pbench.server.database.models.tracker import Dataset, States, DatasetNotFound
 from pbench.server.database.models.users import User
-from pbench.server.database.database import Database
 
 
 _NAME_ = "pbench-import-datasets"
@@ -166,7 +166,7 @@ def main(options):
 
     # We're going to need the Postgres DB to track dataset state, so setup
     # DB access.
-    Database.init_db(config, logger)
+    init_db(config, logger)
 
     # NOTE: the importer will ignore datasets that already exist in the DB;
     # we do the ACTION links first to get set up, and then sweep other links
--- a/server/bin/pbench-state-manager.py
+++ b/server/bin/pbench-state-manager.py
@@ -19,7 +19,7 @@ from argparse import ArgumentParser
 from pbench import BadConfig
 from pbench.common.logger import get_pbench_logger
 from pbench.server import PbenchServerConfig
-from pbench.server.database.database import Database
+from pbench.server.database import init_db
 from pbench.server.database.models.tracker import Dataset, States
 
 
@@ -46,7 +46,7 @@ def main(options):
 
         # We're going to need the Postgres DB to track dataset state, so setup
         # DB access.
-        Database.init_db(config, logger)
+        init_db(config, logger)
 
         args = {}
         if options.create:
--- a/lib/pbench/cli/server/user_create.py
+++ b/lib/pbench/cli/server/user_create.py
@@ -4,7 +4,7 @@ from pbench import BadConfig
 from pbench.cli.server import pass_cli_context
 from pbench.cli.server.options import common_options
 from pbench.server import PbenchServerConfig
-from pbench.server.database.database import Database
+from pbench.server.database import init_db
 from pbench.server.database.models.users import Roles, User
 
 _NAME_ = "pbench-user-create"
@@ -20,7 +20,7 @@ class UserCreate:
         # We're going to need the Postgres DB to track dataset state, so setup
         # DB access. We execute without a logger to avoid unnecessary noise in
         # CLI gold test output from an incidental tool.
-        Database.init_db(config, None)
+        init_db(config, None)
 
         user = User(
             username=self.context.username,
--- a/server/bin/pbench-backup-tarballs.py
+++ b/server/bin/pbench-backup-tarballs.py
@@ -21,7 +21,7 @@ from pbench.server.database.models.tracker import (
     Metadata,
     DatasetError,
 )
-from pbench.server.database.database import Database
+from pbench.server.database import init_db
 
 
 _NAME_ = "pbench-backup-tarballs"
@@ -463,7 +463,7 @@ def main(cfg_name):
 
     # We're going to need the Postgres DB to track dataset state, so setup
     # DB access.
-    Database.init_db(config, logger)
+    init_db(config, logger)
 
     # Add a BACKUP and QDIR field to the config object
     config.BACKUP = config.conf.get("pbench-server", "pbench-backup-dir")
--- a/server/bin/pbench-cull-unpacked-tarballs.py
+++ b/server/bin/pbench-cull-unpacked-tarballs.py
@@ -32,7 +32,7 @@ import pbench.server
 from pbench.common.exceptions import BadConfig
 from pbench.common.logger import get_pbench_logger
 from pbench.server import PbenchServerConfig
-from pbench.server.database.database import Database
+from pbench.server.database import init_db
 from pbench.server.indexer import _STD_DATETIME_FMT
 from pbench.server.report import Report
 
@@ -330,7 +330,7 @@ def main(options):
 
     # We're going to need the Postgres DB to track dataset state, so setup
     # DB access.
-    Database.init_db(config, logger)
+    init_db(config, logger)
 
     archivepath = config.ARCHIVE
 
--- a/server/bin/pbench-index.py
+++ b/server/bin/pbench-index.py
@@ -15,7 +15,7 @@ from configparser import Error as ConfigParserError
 
 from pbench.common.logger import get_pbench_logger
 from pbench.server import PbenchServerConfig
-from pbench.server.database.database import Database
+from pbench.server.database import init_db
 from pbench.common.exceptions import (
     BadConfig,
     ConfigFileError,
@@ -110,7 +110,7 @@ def main(options, name):
         # Logger independently.
         config = PbenchServerConfig(options.cfg_name)
         logger = get_pbench_logger(name, config)
-        Database.init_db(config, logger)
+        init_db(config, logger)
 
         # Now we can initialize the index context
         idxctx = IdxContext(options, name, config, logger, _dbg=_DEBUG)
--- a/server/bin/pbench-reindex.py
+++ b/server/bin/pbench-reindex.py
@@ -30,7 +30,7 @@ from pbench.server.database.models.tracker import (
     Metadata,
     DatasetError,
 )
-from pbench.server.database.database import Database
+from pbench.server.database import init_db
 from pbench.common.logger import get_pbench_logger
 
 
@@ -184,7 +184,7 @@ def main(options):
 
     # We're going to need the Postgres DB to track dataset state, so setup
     # DB access.
-    Database.init_db(config, logger)
+    init_db(config, logger)
 
     archive_p = config.ARCHIVE
 
--- a/server/bin/pbench-report-status.py
+++ b/server/bin/pbench-report-status.py
@@ -7,7 +7,7 @@ from argparse import ArgumentParser
 from socket import gethostname
 
 from pbench.server import PbenchServerConfig
-from pbench.server.database.database import Database
+from pbench.server.database import init_db
 from pbench.common.exceptions import BadConfig
 from pbench.server.report import Report
 
@@ -74,7 +74,7 @@ except BadConfig as e:
 # We're going to need the Postgres DB to track dataset state, so setup
 # DB access. We don't pass a Logger here, because that introduces lots
 # of spurious changes in the gold CLI test output.
-Database.init_db(config, None)
+init_db(config, None)
 
 hostname = gethostname()
 pid = parsed.pid
--- a/server/bin/pbench-server-prep-shim-002.py
+++ b/server/bin/pbench-server-prep-shim-002.py
@@ -24,7 +24,7 @@ from pbench.server import PbenchServerConfig
 from pbench.server.report import Report
 from pbench.server.utils import quarantine
 from pbench.server.database.models.tracker import Dataset, States, DatasetError
-from pbench.server.database.database import Database
+from pbench.server.database import init_db
 
 
 _NAME_ = "pbench-server-prep-shim-002"
@@ -313,7 +313,7 @@ def main(cfg_name):
 
     # We're going to need the Postgres DB to track dataset state, so setup
     # DB access.
-    Database.init_db(config, logger)
+    init_db(config, logger)
 
     qdir, receive_dir = fetch_config_val(config, logger)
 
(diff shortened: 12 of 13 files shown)


lib/pbench/server/api/__init__.py
```diff
@@ -134,7 +134,7 @@ def create_app(server_config):
     register_endpoints(api, app, server_config)
 
     try:
-        Database.init_db(server_config=server_config, logger=app.logger)
+        init_db(server_config=server_config, logger=app.logger)
```
portante, line 137: How does this work?  `init_db` does not appear to accept keyword arguments, while `Database.init_db` does.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -282,9 +282,9 @@ def post(self):
                 self.logger.exception("Exception occurred deleting an auth token")
                 abort(500, message="INTERNAL ERROR")
             else:
-                self.logger.debug("User {} logged out", user.username)
+                self.logger.warning("User {} logged out", user.username)
```
portante, line 285: I'd keep this as a `debug()`, informational at best.  We don't want to be too verbose as this will happen often.

portante (changes requested):

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -59,7 +59,7 @@ def put(self, controller):
             abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
         if not request.headers.get("filename"):
-            self.logger.debug(
+            self.logger.warning(
```
dbutenhof, line 62: This conflicts with Pete's changes. At one point when we'd talked about log level someone had proposed reserving `warning` for "user errors" we wanted to note but separate from internal Pbench errors; your usage of `warning` here fits better than Pete's `info`. I would propose that Pete just back out his log level changes given this PR.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -79,26 +79,26 @@ def put(self, controller):
...
             content_length = int(request.headers.get("Content-Length"))
         except ValueError:
-            self.logger.debug(
+            self.logger.warning(
                 f"Tarfile upload: Invalid content-length header, not an integer for file {filename}"
             )
             abort(400, message="Invalid content-length header, not an integer")
         except Exception:
-            self.logger.debug(
+            self.logger.exception(
```
dbutenhof, line 95: I tend to agree with an `exception` level for a miscellaneous `Exception` ... however looking at this it seems clear that the message here is assuming that it was a `KeyError` for a missing header, and if it's really something else the messaging will be misleading. I think the code here should be an `except KeyError as e` and report the `str(e)` (probably `"Content-Length"`), with a `warning` level.

If we really think there's a reason to anticipate some other generic `Exception` here, we should add a separate `except Exception` with a 500 abort and an `exception` level log with a more generic description. But I don't think that's very likely...

lib/pbench/server/api/resources/users_api.py
```diff
@@ -282,9 +282,9 @@ def post(self):
...
         else:
-            self.logger.debug("User logout with invalid or expired token")
+            self.logger.warning("User logout with invalid or expired token")
```
dbutenhof, line 287: Tricky. A logout with an "invalid" token is clearly a user error and therefore `warning`. But is logging out with an expired token really a "user error" or just an unusual event possibly worth noting (info level)?? Hmm...

server/bin/pbench-import-datasets.py
```diff
@@ -33,9 +33,9 @@
 from pbench import BadConfig
 from pbench.common.logger import get_pbench_logger
 from pbench.server import PbenchServerConfig
+from pbench.server.api import init_db
```
dbutenhof, line 36: Again, I think this would be more logically consistent as `from pbench.server.database import init_db` since this command (and others like it) have nothing to do with the api. I get that "throwing it in there" was convenient, but I don't think it's clean design.

lib/pbench/server/api/resources/upload_api.py
portante: If fixed my changes to match.

lib/pbench/server/api/resources/users_api.py
portante: Seems like this would be best kept as an informational.

lib/pbench/server/api/__init__.py
npalaska: What do you mean? they both accept 2 parameters `server_config` and `logger`

lib/pbench/server/api/__init__.py
```diff
@@ -144,3 +144,17 @@ def shutdown_session(exception=None):
...
+    Utility method for initializing the database.
+
+    In order to register all the tables properly in our database, db models need to be imported before the
+    call to Database.init function.
+
+    Lot of independant functionality in our code depends on database access without creating the flask app.
+    Invoking init_db from here make sure all models are imported properly, all the functionalities where we
+    need a standalone db access should invoke this function instead of directly invoking Database.init_db
+    """
+    Database.init_db(server_config=server_config, logger=logger)
```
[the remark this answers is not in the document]
npalaska: Wouldn't importing all the models in `server.database.__init__.py` result in flake8 complain?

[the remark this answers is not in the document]
dbutenhof: They can be suppressed with `# noqa: F401` on the line, although that's a bit ugly.

There could also be an unused method that referenced something in each imported module, though that's also a bit ugly.

Sometimes tools can be a pain.

Even so, this method is really in the wrong place; we wouldn't have this problem for api code, and code that's dependent about the DB shouldn't be made dependent on the API... even aside from the fact that, as I commented before, it's really just coincidence that this module pulls in all the current DB models.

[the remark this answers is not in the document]
npalaska: Ah yeah there is an option to specify `# noqa` I could do something like `from pbench.server.database.models import *  # noqa` and have this method defined in /database/__init__.py file and we are almost always certain that all the models are imported before the call to init method

npalaska:
There are few conflict between this PR and Peter's https://github.com/distributed-system-analysis/pbench/pull/2264 but mostly in upload api file, where the log related issues are fixed, so I will just accept the upload api file from https://github.com/distributed-system-analysis/pbench/pull/2264 when I get to rebasing this PR on top of it

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -91,11 +91,16 @@ def put(self, controller):
                 f"Tarfile upload: Invalid content-length header, not an integer for file {filename}"
             )
             abort(400, message="Invalid content-length header, not an integer")
-        except Exception:
-            self.logger.exception(
+        except KeyError:
+            self.logger.warning(
                 f"Tarfile upload: No Content-Length header value found for file {filename}"
             )
             abort(400, message="Missing required content-length header")
```
dbutenhof, line 98: Huh. I was thinking at first that if we're bothering with `Exception` here, why not in the other header checks? Then I realized that we're using `.get` so we'll just get `None` rather than a `KeyError`; I think that would just be another `ValueError`, and the only way we'll see "no Content-Length" would be if we checked for `None` explicitly.

I wonder if all of the headers should use the optimistic pattern of

```
try:
    h = request.headers["header"]
    g = request.headers["gopher"]
except KeyError as e:
    self.logger.warning("missing required header {}", e)
    abort(400, f"Missing required {e} header")
```

`Content-Length`, with the `int()` conversion, is a special case, adding the `ValueError`, so we might not want to just put them all in a list like this, but it _would_ be good to follow the same pattern

In any case, though; the original code here was wrong about the meaning of `Exception`, because anything other than `ValueError` is a 500 Internal Error; and if we want to catch "missing header" we need to do it like the others, by checking explicitly for `None` or by using array reference style and catching `KeyError`.

lib/pbench/server/database/__init__.py
```diff
@@ -0,0 +1,16 @@
+from pbench.server.database.models import *  # noqa
```
dbutenhof, line 1: Pretty sure this won't do what you want. (Unless all of the model objects were in the `__init__.py`, which some people do recommend.)

But take a look at https://docs.pylonsproject.org/projects/pyramid_cookbook/en/latest/database/sqlalchemy.html#importing-all-sqlalchemy-models for some ideas to automatically import from the module.

npalaska: Its interesting, the pyramid package installation recommendation is a little weird though, they seem to suggest installing it like `$VENV/bin/pip install "pyramid==2.0"` instead of bin/activate, then pip3. This would however adds an additional dependency, for now I will just add the model imports in the model's `__init__` file and with a note that every new model defined we need to add an import statement here

npalaska: I guess we can use it by keeping the settings parameter default 
```
config = Configurator()
config.scan('myapp.models') 
```

lib/pbench/server/database/models/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""
+For any new database model added in the models directory
+an import statemnt of the same is required here.
```
portante, line 3: Does this really work?  We need to make sure the location where the database is supposed to be initialized does that, can't that be implemented instead in one function called in all the right places?

portante (changes requested):

lib/pbench/server/database/models/__init__.py
npalaska: So because all the models are imported here we can do `from pbench.server.database.models import *` in our database module's `__init__.py` file, that is where we invoke the `Database.init_db` method. Therefore before invoking the init_db call we indirectly import all the models as well

lib/pbench/server/api/__init__.py
webbnh: And, it's an excellent idea especially when passing boolean literals as arguments to specify them as keyword parameters.

lib/pbench/server/api/resources/users_api.py
webbnh: > is logging out with an expired token really a "user error" or just an unusual event possibly worth noting (info level)??

Logging out with an expired token shouldn't actually be an error, and it might not even be unusual, depending on how quickly tokens expire and how the Dashboard maintains them.  I can easily imagine someone staring at graphs for an hour and then logging out...which, I think, absent automatic maintenance, would use an expired token.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -79,26 +79,33 @@ def put(self, controller):
...
             )
             abort(400, message="Invalid content-length header, not an integer")
         except Exception:
-            self.logger.debug(
-                f"Tarfile upload: No Content-Length header value found for file {filename}"
+            self.logger.exception(
+                f"Tarfile upload: Exception on interpreting Content-Length header value"
             )
-            abort(400, message="Missing required content-length header")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
```
webbnh, line 105: What is our proposed practice with regard to HTTP status values:  numeric literals or symbolics?...I don't think we should be using both.  If we prefer symbolics, then we should be making an effort to replace the existing numerics, and we certainly shouldn't be adding new ones (e.g., line 92).

lib/pbench/server/database/__init__.py
webbnh: If possible, you should specify which diagnostic we want to suppress with the `# noqa` directive.  (I think Dave said it was `F401`.)

lib/pbench/server/database/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+def init_db(server_config, logger):
+    """
+    Utility method for initializing the database.
+
+    In order to register all the tables properly in our database, db models need to be imported before the
+    call to Database.init function.
+
+    Lot of independant functionality in our code depends on database access without creating the flask app.
+    Invoking init_db from here make sure all models are imported properly, all the functionalities where we
+    need a standalone db access should invoke this function instead of directly invoking Database.init_db
```
webbnh, line 14: Minor editorial corrections:
```
    In order to register all the tables properly in our database, the db models
    need to be imported before the call to the Database.init_db() function.

    A lot of independent functionality in our code depends on database
    access without creating the flask app.  Invoking init_db() from here
    makes sure all models are imported properly; therefore, all of the code
    where we need standalone db access should invoke this function instead
    of directly invoking Database.init_db().
```

lib/pbench/server/database/models/__init__.py
webbnh: s/statemnt/statement/

lib/pbench/server/database/models/__init__.py
```diff
@@ -0,0 +1,7 @@
...
+"""
+from pbench.server.database.models.active_tokens import ActiveTokens  # noqa
+from pbench.server.database.models.users import User  # noqa
+from pbench.server.database.models.tracker import Dataset, Metadata  # noqa
```
webbnh, line 7: Are the `# noqa` directives required, here?  If so, can you add the code(s) for the diagnostics that we want to suppress?

webbnh (changes requested):
Just a small handful of small items.

lib/pbench/server/database/models/__init__.py
webbnh: @dbutenhof, what about #2255, which adds a module to the `models` directory -- will that require adding an import below, or do we need to rephrase the usage of "any", above?

dbutenhof (changes requested):

lib/pbench/server/api/resources/upload_api.py
npalaska: I prefer symbolic names too but I think that can be addressed as a separate PR to map all the existing error codes from our code to their symbolic names.

lib/pbench/server/database/__init__.py
```diff
@@ -1,16 +1,18 @@
-from pbench.server.database.models import *  # noqa
+from pbench.server.database.models import *  # noqa F401
```
dbutenhof: I'm pretty sure this one isn't under F401. In any case, I think just `import pbench.server.database.models` should work without any diagnostics to awkwardly suppress. There's just the one module namespace to import from `__init__.py` including all the model imports.

Again, another alternative would be just pulling the individual module import list up to here and avoiding the indirection.

lib/pbench/server/database/__init__.py
npalaska: Yeah I thought about doing that too, and now that I think about it again I feel in the process of making it straight forward I took a long route to the destination

lib/pbench/server/database/__init__.py
npalaska: Yes, I addressed that in the rebase, this is a stale code

lib/pbench/server/api/resources/upload_api.py
portante: @npalaska, can you file an issue to track for the future?

portante (dismissed):

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -79,26 +79,33 @@ def put(self, controller):
...
                 f"Tarfile upload: Bad file extension received for file {filename}"
             )
             abort(400, message="File extension not supported. Only .xz")
 
+        content_length = request.headers.get("Content-Length")
+        if not content_length:
+            self.logger.warning(
+                f"Tarfile upload: No Content-Length header value found for file {filename}"
+            )
+            abort(400, message="Missing required content-length header")
```
dbutenhof, line 92: Or we could use "411 Length Required"

dbutenhof (dismissed):

lib/pbench/server/api/resources/upload_api.py
webbnh: Or, eh-hem, we could use `HTTPStatus.LENGTH_REQUIRED`...

lib/pbench/server/api/resources/upload_api.py
webbnh: I concur that a mass conversion from numeric literals to symbolic references should not be done in this PR.  However, any _new_ code should use symbolic references...including the line above.  :-)

webbnh (changes requested):
Just one line needs fixing, I think.

dbutenhof:
This needs a rebase; I'd like to have the db init submitted. I'm running into a dependency problem in #2291 ... I had to explicitly import `ActiveToken` into `User` in order to avoid a SQLAlchemy problem from `pbench-state-manager` now that `Dataset` depends on `User` ... even though neither of them directly references active tokens. Apparently the old `Database.db_init` path isn't enough to pull these all in.

I added an `import` in `users.py` and marked it `# noqa F401` to keep flake8 happy ... but LGTM.com is still complaining about it, so I'd like to remove it ... without breaking the code. Rebasing on top of this more general db init would probably accomplish that.

lib/pbench/server/api/resources/upload_api.py
npalaska: It is now already addressed in PR https://github.com/distributed-system-analysis/pbench/pull/2287. So this commit is just for standalone db init now

npalaska: Also @portante I filed an issue https://github.com/distributed-system-analysis/pbench/issues/2310 to track for converting numerical http status codes to their corresponding symbolic names

lib/pbench/server/database/__init__.py
```diff
@@ -0,0 +1,25 @@
+"""
+For any new database model added in the models directory
+an import statement of the same is required here.
+"""
+from pbench.server.database.models.active_tokens import ActiveTokens  # noqa F401
+from pbench.server.database.models.users import User  # noqa F401
+from pbench.server.database.models.tracker import Dataset, Metadata  # noqa F401
```
dbutenhof, line 7: You should add `Template` from `pbench.server.database.models.template` now.

server/bin/pbench-state-manager.py
```diff
@@ -19,7 +19,7 @@
 from pbench import BadConfig
 from pbench.common.logger import get_pbench_logger
 from pbench.server import PbenchServerConfig
-from pbench.server.database.database import Database
+from pbench.server.api import init_db
```
dbutenhof, line 22: `pbench.server.database`???

dbutenhof (commented):
We're also now initializing the database in:

1. pbench-backup-tarballs.py
2. pbench-cull-unpacked-tarballs.py
3. pbench-index.py
4. pbench-reindex.py
5. pbench-report-status.py
6. pbench-server-prep-shim-003.py
7. pbench-verify-backup-tarballs.py

These should be changed to the new pattern.

server/bin/pbench-state-manager.py
webbnh: @dbutenhof, what question are you asking or what sentiment are you expressing, here?

With these changes, I no longer see any reference to `pbench.server.database`, but I do see a (new) reference to `pbench.server.api:init_db`.

webbnh (dismissed):
I have no further issues with these changes, but they appear to need a rebase, now.

server/bin/pbench-state-manager.py
dbutenhof: > @dbutenhof, what question are you asking or what sentiment are you expressing, here?
> 
> With these changes, I no longer see any reference to `pbench.server.database`, but I do see a (new) reference to `pbench.server.api:init_db`.

Exactly; the other files have `from pbench.server.database import init_db` ... the new `init_db` is in `lib/pbench/server/database/__init__.py`, not in `lib/pbench/server/api/__init__.py`.

npalaska: yeah pushing the fix with other changes

dbutenhof:
Looking better, except `black` doesn't like you:

```
black --check ...
[1mwould reformat /home/pbench/lib/pbench/server/database/__init__.py[0m
[1mOh no! 💥 💔 💥[0m
[1m1 file would be reformatted[0m, 139 files would be left unchanged.[0m
FAILED: black --check .
```

Also, you might want to edit the title and description of the PR to match the reduced scope!

dbutenhof (dismissed):
I'd still prefer to see the title and description edited before it gets merged into permanent history; but the code looks good.

webbnh (commented):
Looks done...except for the rebase....

npalaska:
Yes, updated

webbnh (approved):
GTG!

dbutenhof (approved):
Yea!

portante (approved):

PR #2264, opened 2021-04-30
Add the `Content-Length` header to the PUT req

### LINKED ISSUE #2260: pbench-results-push does not provide the required `Content-Length` header
The PUT requires Content-Length header for upload

portante:
We were missing the `Content-Length` header on the PUT API request as generated by the CLI `pbench-push-results` command.

We modify the `md5sum` utility method to also return the length so that we won't have to stat the file separately.

----

Fixes #2260.

### DIFF
--- a/lib/pbench/common/utils.py
+++ b/lib/pbench/common/utils.py
@@ -11,10 +11,12 @@ def md5sum(filename):
     md5sum - return the MD5 check-sum of a given file without reading the
              entire file into memory.
 
-    Returns hex digest string of the given file.
+    Returns a tuple of the length and the hex digest string of the given file.
     """
     with open(filename, mode="rb") as f:
         d = hashlib.md5()
+        length = 0
         for buf in iter(partial(f.read, 2 ** 20), b""):
+            length += len(buf)
             d.update(buf)
-    return d.hexdigest()
+    return length, d.hexdigest()
--- a/server/bin/pbench-backup-tarballs.py
+++ b/server/bin/pbench-backup-tarballs.py
@@ -343,7 +343,7 @@ def backup_data(lb_obj, s3_obj, config, logger):
 
         # match md5sum of the tarball to its md5 file
         try:
-            archive_tar_hex_value = md5sum(tar)
+            (_, archive_tar_hex_value) = md5sum(tar)
         except Exception:
             # Could not read file.
             quarantine(qdir, logger, tb)
--- a/lib/pbench/agent/results.py
+++ b/lib/pbench/agent/results.py
@@ -175,9 +175,11 @@ class CopyResultTb:
                     authorized to make the PUT request on behalf of a
                     specific user.
         """
+        content_length, content_md5 = md5sum(self.tarball)
         headers = {
             "filename": secure_filename(str(self.tarball)),
-            "Content-MD5": md5sum(self.tarball),
+            "Content-MD5": content_md5,
+            "Content-Length": str(content_length),
             "Authorization": f"Bearer {token}",
         }
         with self.tarball.open("rb") as f:
--- a/lib/pbench/agent/tool_meister.py
+++ b/lib/pbench/agent/tool_meister.py
@@ -1314,7 +1314,7 @@ class ToolMeister:
                     failures += 1
                 else:
                     try:
-                        tar_md5 = md5sum(tar_file)
+                        (_, tar_md5) = md5sum(tar_file)
                     except Exception:
                         self.logger.exception("Failed to read tar ball, '%s'", tar_file)
                         failures += 1
--- a/lib/pbench/test/unit/common/test_utils.py
+++ b/lib/pbench/test/unit/common/test_utils.py
@@ -9,8 +9,12 @@ class TestMd5sum:
         # Reuse the existing file from the server upload fixture.
         filename = "log.tar.xz"
         test_file = Path("./lib/pbench/test/unit/server/fixtures/upload/", filename)
+        expected_length = test_file.stat().st_size
         expected_hash_md5 = open(f"{test_file}.md5", "r").read().split()[0]
-        hash_md5 = md5sum(test_file)
+        length, hash_md5 = md5sum(test_file)
+        assert (
+            length == expected_length
+        ), f"Expected length '{expected_length}', got '{length}'"
         assert (
             hash_md5 == expected_hash_md5
         ), f"Expected MD5 '{expected_hash_md5}', got '{hash_md5}'"
--- a/server/bin/pbench-server-prep-shim-002.py
+++ b/server/bin/pbench-server-prep-shim-002.py
@@ -93,7 +93,7 @@ def md5_check(tb, tbmd5, logger):
 
     # get hex value of the tarball's md5sum
     try:
-        archive_tar_hex_value = md5sum(tb)
+        (_, archive_tar_hex_value) = md5sum(tb)
     except Exception:
         archive_tar_hex_value = None
         logger.exception("Quarantine: Could not read {}", tb)
--- a/server/bin/pbench-verify-backup-tarballs.py
+++ b/server/bin/pbench-verify-backup-tarballs.py
@@ -170,7 +170,7 @@ class BackupObject:
             self.indicator_file_fail, "w"
         ) as f_fail:
             for tar in self.content_list:
-                md5_returned = md5sum(Path(self.dirname, tar.name))
+                (_, md5_returned) = md5sum(Path(self.dirname, tar.name))
                 if tar.md5 == md5_returned:
                     f_ok.write(f"{tar.name}: {'OK'}\n")
                 else:


portante (commented):
This also needs to modify the "copy TB" test to verify all required headers are provided.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -51,69 +51,76 @@ def __init__(self, config, logger):
...
+        md5sum = request.headers.get("Content-MD5")
+        if not md5sum:
+            self.logger.info(
+                "Tarfile upload: Post operation failed due to missing md5sum"
+                f" header for file {filename}"
             )
             abort(
                 400,
-                message="Missing md5sum header, POST operation requires md5sum of an uploaded file in header",
+                message="Missing md5sum header, POST operation requires md5sum"
```
dbutenhof, line 86: `PUT`

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -51,69 +51,76 @@ def __init__(self, config, logger):
...
             self.logger.exception("Tarfile upload: Exception verifying the username")
             abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
-        if not request.headers.get("filename"):
-            self.logger.debug(
-                "Tarfile upload: Post operation failed due to missing filename header"
+        filename_hdr = request.headers.get("filename")
+        if not filename_hdr:
+            self.logger.info(
+                "Tarfile upload: Post operation failed due to missing"
```
dbutenhof, line 63: Not your bug, but it's `PUT` not `POST`!

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -51,69 +51,76 @@ def __init__(self, config, logger):
...
+                " filename header"
             )
             abort(
                 400,
-                message="Missing filename header, POST operation requires a filename header to name the uploaded file",
+                message="Missing filename header, POST operation requires a"
```
dbutenhof, line 68: `PUT`

dbutenhof (dismissed):

lib/pbench/server/api/resources/upload_api.py
portante: Fixed.

lib/pbench/server/api/resources/upload_api.py
portante: Fixed.

lib/pbench/server/api/resources/upload_api.py
portante: Fixed.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -51,69 +51,75 @@ def __init__(self, config, logger):
...
         except Exception:
             self.logger.exception("Tarfile upload: Exception verifying the username")
             abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
-        if not request.headers.get("filename"):
-            self.logger.debug(
-                "Tarfile upload: Post operation failed due to missing filename header"
+        filename_hdr = request.headers.get("filename")
+        if not filename_hdr:
+            self.logger.info(
```
dbutenhof, line 62: This needs to be resolved between your PR and Nikhil's. At one point, we'd more or less decided to reserve `warning` log level for "user errors", and this qualifies. While I was never 100% comfortable with that convention, it seems at least plausible without creating our own additional log level. Whereas mixing internal informational status messages with user errors makes the log harder to parse. All in all, I think the `warning` convention is the best idea we'd seen in this area... but, more pragmatically, having you change it to `info` and then Nikhil immediately changing it to `warning` doesn't help anyone. ;-)

dbutenhof (dismissed):

lib/pbench/server/api/resources/upload_api.py
portante: Yes, good point.  Fixed.

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -96,7 +96,8 @@ def test_missing_authorization_header(client, caplog, server_config):
         )
         assert response.status_code == 401
         for record in caplog.records:
-            assert record.levelname not in ("ERROR", "CRITICAL")
+            assert record.levelname not in ("DEBUG", "ERROR", "CRITICAL")
+        assert caplog.records[-1].levelname == "WARNING"
```
dbutenhof, line 100: I've always been a bit uncomfortable with tests like this. First, the logging isn't really a part of the API contract, and white box testing is always fragile. The correctness of the API doesn't really hinge on whether we write a log message; and, furthermore, it would be as correct if we issued a `warn` level for the bad header and then an `info` or `debug` for completion of the API, possibly wrapped somewhere implicitly. (That is, I'd be a lot less unhappy with an assertion that we generated at least one `WARNING` level than with an assertion that the _last_ log was `WARNING`.)

The test may be correct as written, now; but it might fail in the future for non-obvious reasons that have nothing to do with whether the API is behaving correctly, and that makes me nervous. It seems this sort of capture is over-testing "just because we can" and we may pay for it later.

dbutenhof (approved):

npalaska (approved):

lib/pbench/test/unit/server/test_requests.py
dbutenhof: OK, as a temporary replacement for an audit log record, which I agree _is_ a part of the API, I'll accept this hack. ;-)

webbnh: I agree with Dave's concerns around testing the log contents (particularly against the notion that the test can predict what the last record will be).

I'm concerned about calcifying the contents of the log (Peter posted elsewhere about "Day 1" vs. "Day 2", and I think that the converse argument applies here -- the log should be a "developer artifact", used for diagnosis and support as well as development, and it shouldn't be part of "correct operation".)

As for "too strict", I've worked on a project which used TDD, and it became very hard to refactor the code because the "units" which the unit testing tested were too small, such that virtually any change to the code resulted in having to rewrite the unit tests as well (and, because of the way the unit testing framework worked, it wasn't easy to write anything but whitebox tests).  So, I'd like to avoid that, here:  we need to agree what the interface contract is, and test that, and try not to peek too much under the covers.

Nevertheless, until we have an audit system, the log will have to serve, I suppose.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -51,69 +51,75 @@ def __init__(self, config, logger):
...
+        if not md5sum:
+            self.logger.warning(
+                "Tarfile upload: PUT operation failed due to missing md5sum"
+                f" header for file {filename}"
             )
             abort(
                 400,
-                message="Missing md5sum header, POST operation requires md5sum of an uploaded file in header",
+                message="Missing md5sum header, PUT operation requires md5sum"
+                " of an uploaded file in header",
```
webbnh, line 86: This message seems unnecessarily wordy.  Just "Missing md5sum header" would suffice.  (Presumably the requestor knows that it was a PUT for a file upload....)

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -51,69 +51,75 @@ def __init__(self, config, logger):
...
+        if content_length > self.max_content_length:
+            self.logger.warning(
+                "Tarfile upload: Content-Length exceeded maximum upload size"
+                f" allowed fo file {filename}"
+            )
+            abort(
+                400,
+                message=f"Payload body too large, {content_length:d} bytes,"
+                " maximum size should be less than or equal to"
+                f" {humanize.naturalsize(self.max_content_length)}",
```
webbnh, line 113: This message is botched.  How about, "Upload size ({content_length:d} bytes) must be no more than {humanize.naturalsize(self.max_content_length)}."

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -170,17 +176,18 @@ def put(self, controller):
...
-                    f"Tarfile upload: Bytes received does not match with content length header value for file {filename}"
+                self.logger.warning(
+                    "Tarfile upload: Bytes received does not match with"
+                    f" content length header value for file {filename}"
                 )
                 message = (
-                    f"Bytes received ({bytes_received}) does not match with content length header"
-                    f" ({content_length}), upload failed"
+                    f"Bytes received ({bytes_received}) does not match with"
+                    f" content length header ({content_length}), upload failed"
```
webbnh, line 185: I like the response message...can we have that for the log message?

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -43,7 +43,7 @@ def test_host_info(client, pytestconfig, caplog, server_config):
         assert response.status_code == 200
         assert response.json.get("message") == expected_message
         for record in caplog.records:
-            assert record.levelname not in ("WARNING", "ERROR", "CRITICAL")
+            assert record.levelname == "INFO"
```
webbnh, line 46: You really want the test to fail if someone adds a `DEBUG` level message?

server/bin/pbench-backup-tarballs.py
```diff
@@ -343,7 +343,7 @@ def backup_data(lb_obj, s3_obj, config, logger):
 
         # match md5sum of the tarball to its md5 file
         try:
-            archive_tar_hex_value = md5sum(tar)
+            archive_tar_hex_value = md5sum(tar)[1]
```
webbnh, line 346: I would recommend the
```
            (_, archive_tar_hex_value) = md5sum(tar)
```
idiom here, too; and in `server/bin/pbench-server-prep-shim-002.py` and `server/bin/pbench-verify-backup-tarballs.py` as well.

webbnh (approved):
Just nits and other small things for your consideration.

ndokos (approved):
LGTM (but I agree with @webbnh's "nits and  quibbles").

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -51,69 +51,75 @@ def __init__(self, config, logger):
...
-                f"Tarfile upload: Bad file extension received for file {filename}"
+        content_length_hdr = request.headers.get("Content-Length")
+        if not content_length_hdr:
+            self.logger.warning(
+                "Tarfile upload: No Content-Length header value found for"
+                f" file {filename}"
             )
-            abort(400, message="File extension not supported. Only .xz")
-
+            abort(400, message="Missing required content-length header")
```
dbutenhof, line 95: Interestingly, I found by accident yesterday that there's a "411 Length Required" error defined.

There's also a "413 Payload Too Large"...

lib/pbench/test/unit/server/test_requests.py
portante: Tracked in #2281.

portante:
Okay, I could not leave well enough alone.  :(

As I reviewed some of the upload API checks from the comments on the response messages, I realized that we really should not have the "filename" in a header, but instead should be part of the URI.  That exposes it easily to most web server logging which helps us "Day 2" operationally.  Further, it removed a code path for a missing header check.

However, testing the upload API is a real pain, because the "helpful" `test_client()` object prevents us from explicitly setting a value for the `Content-Length` header.  So we can't easily test bad values, too long, or too short without removing the use of the `test_client()`.

Arguably we should keep these three commits separate.

portante:
> However, testing the upload API is a real pain, because the "helpful" `test_client()` object prevents us from explicitly setting a value for the `Content-Length` header. So we can't easily test bad values, too long, or too short without removing the use of the `test_client()`.

Tracked in #2282.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -237,15 +231,12 @@ def put(self, controller):
         response.status_code = 201
         return response
 
+    ALLOWED_EXTENSION = ".tar.xz"
+
```
dbutenhof, line 235: It might make sense to have this defined "way down here" if it were only referenced in `allowed_file` ... but it's referenced up above for the error message, and "hiding" it down here makes it hard to find. It'd be better up at the top of the class.

Also, you reference it in `allowed_file` as `Upload.ALLOWED_EXTENSION` (which is clear and generally preferred) but as `self.ALLOWED_EXTENSION` above, implying an instance attribute rather than a class attribute. Both work, but referencing as a class attribute is better and they should at least be consistent.

dbutenhof (commented):
I'm not thrilled about the endpoints hack; I don't like `users_x`, and I don't like the URI value because it requires the Javascript to find and replace the `x` instead of just glomming the username onto the end as in my simplified.

I don't like the "upload" results even more, though it matters a bit less as the dashboard code won't ever need to use that (yeah, "never say never").

I'm tempted to consider this the tipping point to do a better API schema, though that clearly doesn't belong in this PR. But I'm not entirely sure whether I'd prefer to leave it as it was or let it be completely broken until I can improve things later...

lib/pbench/test/unit/agent/task/test_copy_result_tb.py
```diff
@@ -22,9 +22,10 @@ def config_and_logger(self, valid_config):
 
     @responses.activate
     def test_copy_tar(self, valid_config):
+        tbname = os.path.basename(tarball)
         responses.add(
             responses.PUT,
-            "http://pbench.example.com/api/v1/upload/ctrl/controller",
+            f"http://pbench.example.com/api/v1/upload/ctrl/controller/{tbname}",
```
dbutenhof, line 28: Looks increasingly odd every time I see the pattern. Either we should add a "guide word" like the "ctrl" for the name, or we should remove "ctrl" and streamline things. The inconsistency bugs me.

E.g., `/api/v1/upload/ctrl/<controller>/name/<filename>` or `/api/v1/upload/<controller>/<filename>`.

lib/pbench/test/unit/server/test_endpoint_configure.py
```diff
@@ -58,9 +58,9 @@ def check_config(self, client, server_config, host, my_headers={}):
                 "register": f"{uri}/register",
                 "login": f"{uri}/login",
                 "logout": f"{uri}/logout",
-                "user": f"{uri}/user/",
+                "user_x": f"{uri}/user/x",
```
dbutenhof, line 61: Yeah, really don't like "user_x", and "x" is too generic for a client to just assume it should substitute "x" with the actual username. Again, the dashboard won't use the upload API and doesn't currently GET/DELETE/PUT users but even as a temporary lint this bugs me.

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -151,43 +153,52 @@ def test_mismatched_md5sum_header(client, caplog, server_config):
...
+                response = client.put(
+                    TestUpload.gen_uri(server_config, bad_extension),
+                    data=data_fp,
+                    headers={
+                        "Authorization": "Bearer " + auth_token,
+                        # Content-Length header set automatically
+                        "Content-MD5": "md5sum",  # Ignored
+                    },
+                )
+            assert response.status_code == 415
```
dbutenhof, line 197: 415? That doesn't sound right. "Media type" has a fairly specific meaning, and file name suffix doesn't fit very well.

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -201,38 +212,43 @@ def test_invalid_authorization_upload(client, caplog, server_config):
 
             # Upload with invalid token
             response = client.put(
-                f"{server_config.rest_uri}/upload/ctrl/{socket.gethostname()}",
+                TestUpload.gen_uri(server_config),
                 headers={
                     "Authorization": "Bearer " + auth_token,
+                    "Content-Length": "42",
```
dbutenhof, line 218: But, wait; isn't 42 __always__ the right answer???

portante: Yes, but the confounding `test_client()` ignores the right answer and does what it wants only!

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -125,28 +125,17 @@ def get(self):
...
-                # "/api/v1/foo/<string:name>/detail/<string:param>")
-                url = self.param_template.sub("", url)
-                path = url[len(self.uri_prefix) + 1 :]
-                if path.endswith("/"):
-                    path = path[:-1]
-                path = path.replace("/", "_")
+                # while "/api/v1/foo/<string:username>/detail/<int:bar>" yields:
+                #     "users_x": "/api/v1/foo/x/detail/x"
+                url = self.param_template.sub("x", url)
+                path = url[len(self.uri_prefix) + 1 :].replace("/", "_")
```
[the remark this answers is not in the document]
webbnh: I'm OK with `/upload/ctrl/${controller}/${name}`, although I would pull for replacing `ctrl` with `controller`.

I don't mind the notion that there might be different kinds of "uploads" and that this URI is for controller-based tarballs.

With that as a base, I would regard everything after the `ctrl` (or `controller`...) to be a series of parameters.  As Dave indicated, we could have specified the request as `host://api/v1/upload/controller?controller=<str>&name=<str>`, but expressing it as `host://api/v1/upload/controller/<controller:str>/<filename:str>` is, I think, equivalent.  (And, I would be happy with either one.)

If we buy into _that_, then I think we're back at where we were with the original code:  we truncate the URI at the first template parameter, assume that everything after that point is just a series of one or more parameters, and render the API name as `upload_ctrl` (or `upload_controller`...).

I agree that this PR is not the right place for reworking this stuff.  But, if y'all agree with my way of thinking here, I think we can have the filename as part of the URI without having to worry about "embedded template strings", because they aren't really "embedded" -- they are just tacked on to the end with slash-separators.  :-)

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -165,26 +161,31 @@ def put(self, controller):
...
                 )
-                abort(500, message=f"There was something wrong uploading {filename}")
+                abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
             if bytes_received != content_length:
-                self.logger.debug(
-                    f"Tarfile upload: Bytes received does not match with content length header value for file {filename}"
+                msg = (
+                    "Bytes received do not match Content-Length header"
+                    f" ({content_length} != {bytes_received})"
```
webbnh, line 174: This would probably be better expressed as 
```
                    f" (expected {content_length}; received {bytes_received})"
```

webbnh: Or, you could integrate it into the message:
```
                msg = (
                    "Bytes received ({bytes_received}) do not match Content-Length header ({content_length})"
```

lib/pbench/test/unit/agent/task/test_copy_result_tb.py
webbnh: I don't like the extra "guide word".  And, see my previous comment about `ctrl`.

lib/pbench/test/unit/agent/task/test_results_push.py
```diff
@@ -17,26 +18,20 @@ class TestResultsPush:
     URL = "http://pbench.example.com/api/v1"
 
     @staticmethod
-    def add_success_mock_response():
+    def add_http_mock_response(code: int = 200):
```
webbnh, line 21: Do we like the literal `200` here, or should we be using a symbolic reference?

webbnh: Ditto line 180.

lib/pbench/test/unit/server/test_requests.py
webbnh: What about the other test scenarios?  Shouldn't those ignore `DEBUG` level messages, too?

lib/pbench/server/api/resources/endpoint_configure.py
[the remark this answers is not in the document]
dbutenhof: > I agree that this PR is not the right place for reworking this stuff. But, if y'all agree with my way of thinking here, I think we can have the filename as part of the URI without having to worry about "embedded template strings", because they aren't really "embedded" -- they are just tacked on to the end with slash-separators. :-)

It's still just kicking the problem down the road; but that's what I did before, and I'll accept doing it again. I would definitely prefer

```
    "users": "host://api/v1/upload/"
```

Over the current under-x forms.

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -151,43 +153,52 @@ def test_mismatched_md5sum_header(client, caplog, server_config):
...
-            )
-            assert response.status_code == 400
+            expected_message = "File extension not supported, must be .tar.xz"
+            tmp_d = pytestconfig.cache.get("TMP", None)
+            datafile = Path(tmp_d, bad_extension)
+            datafile.write_text("compressed tar ball")
+            with datafile.open("rb") as data_fp:
+                response = client.put(
+                    TestUpload.gen_uri(server_config, bad_extension),
+                    data=data_fp,
```
webbnh, line 190: Do we actually need a data stream to test this?  (What happens if you omit this parameter, like the test did previously?)

lib/pbench/test/unit/server/test_requests.py
webbnh: The only plausible alternative that I could come up with was [422](https://tools.ietf.org/html/rfc4918#section-11.2); otherwise, I think we're relegated to 400, which is probably OK if the message is clear.

webbnh (approved):
I've got no major objections, other than the concern above about the URI.  (The below are just the usual collection of nits and suggestions.)

lib/pbench/server/api/resources/endpoint_configure.py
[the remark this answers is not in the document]
webbnh: It's only kicking the problem down the road if the road is eventually leading to embedded parameters; otherwise, I think we have an adequate solution so long as we only have parameters which are "appended".

portante:
Pulled the first commit regarding `tox` and positional arguments for `pytest` into PR #2285.

lib/pbench/test/unit/server/test_requests.py
portante: Moved to #2287.

lib/pbench/server/api/resources/endpoint_configure.py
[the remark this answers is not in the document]
portante: Moved to #2287.

lib/pbench/server/api/resources/upload_api.py
portante: Moved to #2287.

lib/pbench/test/unit/agent/task/test_copy_result_tb.py
portante: Moved to #2287.

lib/pbench/test/unit/server/test_endpoint_configure.py
portante: Moved to #2287.

lib/pbench/test/unit/server/test_requests.py
portante: Moved to #2287.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -50,70 +48,62 @@ def __init__(self, config, logger):
...
             abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
-        if not request.headers.get("filename"):
-            self.logger.debug(
-                "Tarfile upload: Post operation failed due to missing filename header"
+        controller = secure_filename(controller_arg)
+        filename = secure_filename(filename_arg)
+        if not self.allowed_file(filename):
+            self.logger.warning(
+                f"Unsupported file extension for {username}/{controller}/{filename}"
```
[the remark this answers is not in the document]
portante: Moved to #2287.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -50,70 +48,62 @@ def __init__(self, config, logger):
...
+                "{} for {}/{}/{}", msg, username, controller_arg, filename_arg
+            )
+            abort(HTTPStatus.LENGTH_REQUIRED, message=msg)
         try:
-            content_length = int(request.headers.get("Content-Length"))
+            content_length = int(content_length_hdr)
         except ValueError:
-            self.logger.debug(
-                f"Tarfile upload: Invalid content-length header, not an integer for file {filename}"
+            msg = "Invalid Content-Length header, not an integer"
```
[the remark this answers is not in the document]
portante: Moved to #2287.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -50,70 +48,62 @@ def __init__(self, config, logger):
...
-                self.logger.debug(
-                    f"Tarfile upload: Content-Length header value is 0 for file {filename}"
-                )
-                abort(
-                    400,
-                    message="Upload failed, Content-Length received in header is 0",
-                )
+            cl = humanize.naturalsize(content_length)
+            mcl = humanize.naturalsize(self.max_content_length)
+            abort(HTTPStatus.REQUEST_ENTITY_TOO_LARGE, message=f"{msg} ({cl} > {mcl})")
```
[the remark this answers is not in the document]
portante: Moved to #2287.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -50,70 +48,62 @@ def __init__(self, config, logger):
...
+        elif content_length == 0:
```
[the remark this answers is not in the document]
portante: Moved to #2287.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -147,10 +141,12 @@ def put(self, controller):
...
             )
 
-        self.logger.info("Uploading file {} to {}", filename, dataset)
+        self.logger.info(
+            "Uploading file {}/{}/{} to {}", username, controller, filename, dataset
+        )
 
         with tempfile.NamedTemporaryFile(mode="wb", dir=path) as ofp:
-            chunk_size = 4096
+            chunk_size = 65536
```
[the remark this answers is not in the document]
portante: Moved to #2287.

lib/pbench/server/api/resources/upload_api.py
portante: Moved to #2287.

lib/pbench/test/unit/server/test_requests.py
portante: Moved to #2287.

lib/pbench/test/unit/agent/task/test_results_push.py
portante: Moved to #2287.

portante:
Moved the rework of the upload API to PR #2287.

portante:
This commit only adds the missing `Content-Length` header to the `CopyResultTb` code.  It does not touch the upload API at all.

webbnh (approved):

riya-17 (approved):

lib/pbench/common/utils.py
```diff
@@ -11,10 +11,12 @@ def md5sum(filename):
     md5sum - return the MD5 check-sum of a given file without reading the
              entire file into memory.
 
-    Returns hex digest string of the given file.
+    Returns a tuple of the length and the hex digest string of the given file.
```
dbutenhof, line 14: Luckily you didn't change the `def` or I'd've suggested adding type hints. ;-)

dbutenhof (approved):
GO!

PR #2279, opened 2021-05-04
Generate raw statistics from archived tar balls

portante:
Helper utility to generate some raw statistics about existing tar balls.


server/bin/pbench-tarball-stats
```diff
@@ -0,0 +1,193 @@
...
+    returning information about the tar so that statistics can be generated.
+    """
+    with os.scandir(archive) as archive_scan:
+        for c_entry in archive_scan:
+            if c_entry.name.startswith(".") and c_entry.is_dir(follow_symlinks=False):
+                continue
+            if not c_entry.is_dir(follow_symlinks=False):
+                # NOTE: the pbench-audit-server should pick up and flag this
+                # unwanted condition.
+                continue
```
dbutenhof, line 53: So we skip `.*` directory files and all non-directory files. I think these two statements could be safely consolidated to

```
if c_entry.name.startswith(".") or not c_entry.is_dir(follow_symlinks=False):
    continue
````

File: server/bin/pbench-tarball-stats [hunk 2]
```diff
@@ -0,0 +1,193 @@
...
+    get_pbench_logger,
+)
+
+
+_NAME_ = "pbench-tarball-stats"
+
+tb_pat_r = (
+    r"\S+_(\d\d\d\d)[._-](\d\d)[._-](\d\d)[T_](\d\d)[._:](\d\d)[._:](\d\d)\.tar\.xz"
+)
+tb_pat = re.compile(tb_pat_r)
```
dbutenhof, line 39: Any real need to retain both the string and the `Pattern` object?

server/bin/pbench-tarball-stats
```diff
@@ -0,0 +1,193 @@
...
+
+
+def gen_tar_balls(archive):
```
dbutenhof, line 42: This scanner is almost the same as the audit generator scan, and the indexer scan, and the importer scan I hacked up a while ago ... one of these days we really need to factor out an extensible minimum base class that we can just get right once.

Or maybe we'll just keep doing this until we switch to object store and get rid of the crufty directory hierarchy "metadata" model entirely ...

server/bin/pbench-tarball-stats
```diff
@@ -0,0 +1,193 @@
...
+            by_year[year].append(tb_rec)
+            month = tb_rec[2].month
+            ym = f"{year:4d}-{month:02d}"
+            by_month[ym].append(tb_rec)
+
+    time_sec = pbench._time() - start
+
+    print(f"Took {time_sec:0.2f} seconds")
+
+    print(f"{len(invalid.keys()):18n} controllers with bad tar balls")
```
dbutenhof, line 156: How can `len(invalid.keys())` ever be different from `invalid_cnt`? If you're looking at both anyway, why not just use the `len()` SPOT?

dbutenhof (approved):
Looks good; a few minor suggestions, but I don't see anything wrong.

server/bin/pbench-tarball-stats
```diff
@@ -150,17 +157,29 @@ def main(options):
...
+    print(
+        f"{len(good.keys()):18n} controllers with good tar balls (EC2 {ec2_cnt:n}, RH {rh_cnt:n})"
+    )
     print(f"{good_cnt:18n} good tar balls, total count")
     print(f"{good_size:18n} good tar balls, total size")
 
     for year, y_list in sorted(by_year.items()):
         count = len(y_list)
-        print(f"{year}: {count:7n}")
+        ec2_cnt = 0
```
dbutenhof, line 168: Would probably be better not to re-use the same variable name in the loops. I've tripped myself up before doing stuff like this because later I want to add some other output at the end using the original total, and then have to figure out why the number is different. ;-)

Of course another option would be to keep the "global" information as instance attributes, and do each output section in a method which shares the arrays and such but have their own local variables to keep them separate... which might make the whole thing easier to read.

dbutenhof (dismissed):

server/bin/pbench-tarball-stats
portante: No.

server/bin/pbench-tarball-stats
portante: Yes, I've moved this PR to the `b0.69` branch, since we can use the data to report all this data in the `main` branch.

So for now, we'll just keep this scanner.

server/bin/pbench-tarball-stats
portante: Count of controllers with invalid tar balls vs total count of invalid tar balls.

server/bin/pbench-tarball-stats
portante: Refactored into a single function.

dbutenhof (approved):

server/bin/pbench-tarball-stats
```diff
@@ -0,0 +1,178 @@
...
+                            int(match.group(2)),
+                            int(match.group(3)),
+                            int(match.group(4)),
+                            int(match.group(5)),
+                            int(match.group(6)),
+                        )
+                    else:
+                        if entry.name.endswith(".md5"):
+                            continue
+                        tb_dt = None
```
webbnh, line 69: If we get here, it's because the file is "invalid" (so long as it's not an MD5 file), and you're signaling that to the caller by setting `tb_dt` to `None`...a comment (here and in the docstring?) explaining this would go a long way.

(Also, I would vote for using an `elif` for the MD5 check and an `else` for the `tb_dt = None`, since you really do have three co-equal cases.)

(And, I would have been tempted, when I hit an MD5 file, to go confirm that the corresponding tar file exists, and to declare the MD5 file "invalid" if the tarball is missing...but, perhaps that is beyond the scope of this tool.)

server/bin/pbench-tarball-stats
```diff
@@ -0,0 +1,178 @@
...
+
+def main(options):
+    locale.setlocale(locale.LC_ALL, "en_US.UTF-8")
+    if not options.cfg_name:
+        print(
+            f"{_NAME_}: ERROR: No config file specified; set"
+            " _PBENCH_SERVER_CONFIG env variable",
+            file=sys.stderr,
+        )
+        return 1
```
webbnh, line 92: I'm not a fan of the multiple discrete return values, except in cases where the intention is for them to provoke different responses in the caller and where they are well-documented.  Here, we should be returning `2`; ditto at lines 107 and 113 (and 98, which already does).

server/bin/pbench-tarball-stats
```diff
@@ -0,0 +1,178 @@
...
+    rh_cnt = 0
+
+    by_year = collections.defaultdict(list)
+    by_month = collections.defaultdict(list)
+
+    start = pbench._time()
+
+    gen = gen_tar_balls(archive_p)
+
+    for tb_rec in gen:
```
webbnh, line 133: It would be better if you used a `namedtuple` or if you destructured the tuple into separately-named local variables, here.

webbnh (approved):
It could use some polish, but there are no deal-breakers, especially for a developer-tool on a post-release branch.

portante (commented):
Added tests and applied all code-review feedback.

portante:
The locale changes to the `pbench-devel` container image were derived from https://stackoverflow.com/questions/58304278/how-to-fix-character-map-file-utf-8-not-found.

server/bin/pbench-tarball-stats.py
```diff
@@ -0,0 +1,196 @@
...
+_NAME_ = "pbench-tarball-stats"
+
+tb_pat = re.compile(
+    r"\S+_(\d\d\d\d)[._-](\d\d)[._-](\d\d)[T_](\d\d)[._:](\d\d)[._:](\d\d)\.tar\.xz"
+)
+
+TarBallInfo = collections.namedtuple("TarBallInfo", ["ctrl", "tb", "dt", "stat"])
+
+
+def gen_tar_balls(archive):
```
webbnh, line 38: Type hints?

server/bin/pbench-tarball-stats.py
```diff
@@ -0,0 +1,196 @@
...
+
+def report_by_bucket(buckets):
+    for bucket_name, bucket_list in sorted(buckets.items()):
+        count = len(bucket_list)
+        ec2_cnt = 0
+        for item in bucket_list:
+            if item[0].startswith("EC2::"):
+                ec2_cnt += 1
+        pct_ec2 = (ec2_cnt / count) * 100.0
+        print(f"{bucket_name}: {count:7n} (%EC2 {pct_ec2:0.02f})")
```
webbnh, line 97: "`%EC2`" is slightly cryptic...I think it would be better if the percent sign followed the number:
```
        print(f"{bucket_name}: {count:7n} (EC2 {pct_ec2:0.02f}%)")
```
or
```
        print(f"{bucket_name}: {count:7n} ({pct_ec2:0.02f}% EC2)")
```

server/bin/pbench-tarball-stats.py
```diff
@@ -0,0 +1,196 @@
...
+
+def main(options):
+    locale.setlocale(locale.LC_ALL, "en_US.UTF-8")
+    if not options.cfg_name:
+        print(
+            f"{_NAME_}: ERROR: No config file specified; set"
+            " _PBENCH_SERVER_CONFIG env variable",
+            file=sys.stderr,
+        )
+        return 1
```
webbnh, line 108: Obviously, returning any non-zero value to indicate errors will work, but I prefer the convention of returning `1` for operational errors and `2` for configuration or invocation errors.  I don't think this tool ever encounters operational errors, so all the explicit error returns would be `2`.

server/bin/pbench-tarball-stats.py
```diff
@@ -0,0 +1,196 @@
...
+
+    try:
+        config = pbench.PbenchConfig(options.cfg_name)
+    except pbench.BadConfig as e:
+        print(f"{_NAME_}: {e}", file=sys.stderr)
+        return 1
+
+    archive = config.ARCHIVE
+    archive_p = os.path.realpath(archive)
```
webbnh, line 117: I don't think the use of `archive` here is buying us enough to justify its existence (you could easily just pass `config.ARCHIVE` as the argument to `realpath()`).

And, at line 149, we pass `archive_p` to `gen_tar_balls()` where the value is renamed to `archive`, so it would be handy if you used `archive` in place of `archive_p` to hold the value in this function.

server/bin/pbench-tarball-stats.py
```diff
@@ -0,0 +1,196 @@
...
+    print(f"Took {time_sec:0.2f} seconds")
+
+    print(f"{len(invalid.keys()):18n} controllers with bad tar balls")
+    print(f"{invalid_cnt:18n} bad tar balls, total count")
+    print(f"{invalid_size:18n} bad tar balls, total size")
+
+    print(
+        f"{len(good.keys()):18n} controllers with good tar balls (EC2 {ec2_cnt:n}, RH {rh_cnt:n})"
+    )
+    print(f"{good_cnt:18n} good tar balls, total count")
```
webbnh, line 181: Perhaps the EC2/RH breakdown belongs on line 181 rather than line 179.
(See comment on `server/bin/gold/test-5.2.txt` above.)

webbnh (dismissed):
Once again, no deal-breakers, just nits and small stuff.

(But, you probably _should_ address the position of the EC2/RH breakdown.)

portante (commented):
All code review comments have been addressed.

server/bin/pbench-tarball-stats.py
```diff
@@ -175,10 +181,21 @@ def main(options):
...
-    print(f"{good_cnt:18n} good tar balls, total count")
+    print(f"{len(good.keys()):18n} controllers with good tar balls (", end="")
+    satellites_ctrl = collections.Counter()
+    for good_ctrl in good.keys():
+        satellites_ctrl[get_sat_name(good_ctrl)] += 1
+    first = True
+    for name, total in sorted(satellites_ctrl.items()):
+        comma, first = ("", False) if first else (", ", False)
+        print(f"{comma}{name} {total:n}", end="")
+    print(")")
```
webbnh, line 192: It would be nice to skip printing the parentheses (and to skip executing everything else, here) if `len(good.keys() == 0`, e.g.:
```
    print(f"{len(good.keys()):18n} controllers with good tar balls", end="")
    if good:
        print(" (", end="")
        satellites_ctrl = collections.Counter()
        for good_ctrl in good.keys():
            satellites_ctrl[get_sat_name(good_ctrl)] += 1
        first = True
        for name, total in sorted(satellites_ctrl.items()):
            comma, first = ("", False) if first else (", ", False)
            print(f"{comma}{name} {total:n}", end="")
        print(")")
```

Ditto for lines 193-198.

server/bin/pbench-tarball-stats.py
```diff
@@ -175,10 +181,21 @@ def main(options):
...
-    )
-    print(f"{good_cnt:18n} good tar balls, total count")
+    print(f"{len(good.keys()):18n} controllers with good tar balls (", end="")
+    satellites_ctrl = collections.Counter()
+    for good_ctrl in good.keys():
+        satellites_ctrl[get_sat_name(good_ctrl)] += 1
+    first = True
+    for name, total in sorted(satellites_ctrl.items()):
+        comma, first = ("", False) if first else (", ", False)
+        print(f"{comma}{name} {total:n}", end="")
```
webbnh, line 191: I think there's a more Pythonic way of doing this.  Perhaps something like,
```
    cnt_strs = [ f"{name} {total:n}" for name, total in sorted(satellites_ctrl.items())]
    print(", ".join(cnt_strs), end="")
```

Ditto for 194-197.

webbnh (dismissed):
One again, it's looking good -- no dealbreakers -- but I have a couple of tweaks for you to consider, one for the output and one for the code.

webbnh (approved):

server/bin/gold/test-5.2.txt
```diff
@@ -316,6 +316,17 @@ len(actions) = 1
 --- Finished pbench-verify-backup-tarballs (status=0)
 +++ Running pbench-satellite-cleanup
 --- Finished pbench-satellite-cleanup (status=0)
++++ Running pbench-tarball-stats
+Took 0.00 seconds
+                 0 controllers with bad tar balls
+                 0 bad tar balls, total count
+                 0 bad tar balls, total size
+                 6 controllers with good tar balls (-INT- 3, ONE 3)
```
dbutenhof, line 324: Wow, that `-INT-` and `ONE` make **no** sense at all until reading the code. (The fact that the test cases are using satellite `ONE` where we usually only see `EC2` probably doesn't help.)

I don't know if some keyword introducing the breakdown, like maybe "origin", would help. I suppose in a real installation, the satellite prefix would be commonly recognized and that might help explain the `-INT-` but something like "local" might be better. 😬

dbutenhof (approved):

server/bin/gold/test-5.2.txt
portante: > Wow, that `-INT-` and `ONE` make **no** sense at all until reading the code. (The fact that the test cases are using satellite `ONE` where we usually only see `EC2` probably doesn't help.)

Well, it is better to not hard-code a particular satellite name.

> I don't know if some keyword introducing the breakdown, like maybe "origin", would help. I suppose in a real installation, the satellite prefix would be commonly recognized and that might help explain the `-INT-` but something like "local" might be better. 😬

Naming ...

portante: What if the output looked like:

```
Took 0.00 seconds
Summary Statistics
                 0 controllers with bad tar balls
                 0 bad tar balls (total count)
                 0 bad tar balls (total size)
                 6 controllers with good tar balls
                 3 main pbench server
                 3 "ONE" pbench satellite server
                 8 good tar balls (total count)
                 5 main pbench server
                 3 "ONE" pbench satellite server
             1,764 good tar balls, total size

Tar Ball Counts broken down by Year and Month
     Total Count  % main   % ONE
1970:          8  62.50%  37.50%
1970-01:       8  62.50%  37.50%
```

portante: Or even:
```
Took 0.00 seconds
Summary Statistics
    Good Tar Balls
                 8 good tar balls total count
                 5 main pbench server
                 3 "ONE" pbench satellite server
             1,764 good tar balls total size

    Good Tar Balls - controller counts
                 6 controllers with good tar balls
                 3 main pbench server
                 3 "ONE" pbench satellite server

    Bad Tar Balls
                 0 controllers with bad tar balls
                 0 bad tar balls total count
                 0 bad tar balls total size

Tar Ball Counts broken down by Year and Month, with satellite percentages
Year/Month  Total Count  % main   % ONE
1970                  8  62.50%  37.50%
1970-01               8  62.50%  37.50%
```

webbnh: I second the request for better names and/or more context.  I like the first of the two proposals, although I would ask for the subtotals to be indented under the totals.  The second proposal is good, too, but it feels kind of big and "airy" -- I appreciate the compactness of the earlier ones (although, I do like the aligned columns for the temporal breakdowns).

riya-17 (approved):

server/bin/gold/test-5.2.txt
portante: I kind like the big and airy ...

```
Summary Statistics for Tar Balls (external data only):

    Took 0.00 seconds to find all tar balls.

    Good Tar Balls:
                 8 total count
             1,764 total size

        By Server Origin:
                 5 main pbench server
                 3 "ONE" pbench satellite server

        Controller Counts:
                 6 controllers
                 3 main pbench server
                 3 "ONE" pbench satellite server

    Bad Tar Balls:
                 0 controllers
                 0 total count
                 0 total size


Tar Ball Counts broken down by Year and Month, with satellite percentages:

Year/Month  Total Count  % main   % ONE
1970                  8  62.50%  37.50%
1970-01               8  62.50%  37.50%
```

webbnh: > I kind like the big and airy ...

Good -- let's go with that, then.

portante (commented):
Updated to use a Jinja2 template.

riya-17 (dismissed):

dbutenhof (approved):
There's a lot more logic in that Jinja2 template than I knew was possible; this is obscure but pretty cool.

dbutenhof:
Oops ... except you've got to update gold files. 👎🏻 😬

dbutenhof (approved):

PR #2287, opened 2021-05-05
Rework upload API to move filename to URI

portante:
The upload API now requires the file name, moving the controller to a required header.  This was chosen with an eye towards the future, where the "controller" value is not needed by the pbench server for processing.  At the time this code was written, the current pbench-agent and the pbench-server use the controller to organize on-disk where the files are placed.  In the future, the controller value will simply be another piece of metadata about the file.

The API is now `/api/v1/upload/<filename>`, with the controller name provided in the `controller` header.  This removes it from prominence in the URI, which is why we did not move it to a URI parameter.

We also change a number of error responses to use the appropriate error codes via `http.HTTPStatus`, and update the messages to be more succinct while improving the context given in the logs.

Further, we fix up the `upload_api.py` module to wrap long lines, and emit 400 response messages in the logs as warnings.  We also tighten down some of the tests to ensure the correct log levels are being emitted.

Finally, we:

 * bump the read & write chunk size to 64K for better through-put.
 * drop the `werkzeug` module entirely (yeah!)
 * enable DEBUG logging during server unit tests

----

Requires PR #2264 to land first.

### DIFF
--- a/lib/pbench/agent/results.py
+++ b/lib/pbench/agent/results.py
@@ -2,17 +2,17 @@ import datetime
 import errno
 import os
 import tarfile
+import urllib.parse
 from configparser import ConfigParser
 from logging import Logger
 from pathlib import Path
 from typing import IO, Iterator
 
 import requests
-from werkzeug.utils import secure_filename
 
 from pbench.agent import PbenchAgentConfig
 from pbench.common.exceptions import BadMDLogFormat
-from pbench.common.utils import md5sum
+from pbench.common.utils import md5sum, validate_hostname
 
 
 class FileUploadError(Exception):
@@ -149,23 +149,32 @@ class CopyResultTb:
     """CopyResultTb - Use the server's HTTP PUT method to upload a tarball
     """
 
-    chunk_size = 4096
+    CHUNK_SIZE = 65536
 
     def __init__(
         self, controller: str, tarball: str, config: PbenchAgentConfig, logger: Logger
     ):
+        """CopyResultTb contructor - raises FileNotFoundError if the given
+        tar ball does not exist, and a ValueError if the given controller is
+        not a valid hostname.
+        """
+        if validate_hostname(controller) != 0:
+            raise ValueError(f"Controller {controller!r} is not a valid host name")
+        self.controller = controller
         self.tarball = Path(tarball)
         if not self.tarball.exists():
             raise FileNotFoundError(f"Tarball '{self.tarball}' does not exist")
-        self.logger = logger
         server_rest_url = config.get("results", "server_rest_url")
-        self.upload_url = f"{server_rest_url}/upload/ctrl/{controller}"
+        tbname = urllib.parse.quote(self.tarball.name)
+        self.upload_url = f"{server_rest_url}/upload/{tbname}"
+        self.logger = logger
 
     def read_in_chunks(self, file_object: IO) -> Iterator[bytes]:
-        data = file_object.read(self.chunk_size)
-        while data:
+        while True:
+            data = file_object.read(self.CHUNK_SIZE)
+            if not data:
+                break
             yield data
-            data = file_object.read(self.chunk_size)
 
     def copy_result_tb(self, token: str) -> None:
         """copy_result_tb - copies tb from agent to configured server upload URL
@@ -177,10 +186,10 @@ class CopyResultTb:
         """
         content_length, content_md5 = md5sum(self.tarball)
         headers = {
-            "filename": secure_filename(str(self.tarball)),
             "Content-MD5": content_md5,
             "Content-Length": str(content_length),
             "Authorization": f"Bearer {token}",
+            "controller": self.controller,
         }
         with self.tarball.open("rb") as f:
             try:
--- a/lib/pbench/common/utils.py
+++ b/lib/pbench/common/utils.py
@@ -2,11 +2,13 @@
 Utility functions common to both agent and server.
 """
 import hashlib
+import ipaddress
+import re
 
 from functools import partial
 
 
-def md5sum(filename):
+def md5sum(filename: str) -> (int, str):
     """
     md5sum - return the MD5 check-sum of a given file without reading the
              entire file into memory.
@@ -20,3 +22,43 @@ def md5sum(filename):
             length += len(buf)
             d.update(buf)
     return length, d.hexdigest()
+
+
+# Derived from https://stackoverflow.com/questions/106179/regular-expression-to-match-dns-hostname-or-ip-address
+# with a few modifications: take advantage of ignoring case; use non-capturing
+# groups to improve efficiency; take advantage of RFC 1123 modification to RFC
+# 952 relaxing first character to be letter or digit, with the repetition moved
+# to the last group to alleviate backtracking.  Or in other words, a case-blind
+# comparison seeking a letter-or-digit, optionally followed by a sequence of 0
+# to 61 letter-or-digit-or-hyphens followed by a letter-or-digit, follwed by 0
+# or more instances of that same pattern preceded by a period.
+_allowed = re.compile(
+    r"[A-Z0-9](?:[A-Z0-9\-]{0,61}[A-Z0-9])?(?:\.[A-Z0-9](?:[A-Z0-9\-]{0,61}[A-Z0-9])?)*",
+    flags=re.IGNORECASE,
+)
+
+
+def validate_hostname(host_name: str) -> int:
+    """validate_hostname - validate the given hostname uses the proper syntax.
+
+    A host name that follows RFC 952 (amended by RFC 1123) is accepted only.
+    Host names are not resolved to IP addresses, and IP addresses are also
+    accepted.
+
+    Algorithm taken from: https://stackoverflow.com/questions/2532053/validate-a-hostname-string
+
+    Returns 0 on success, 1 on failure.
+    """
+    if not host_name or len(host_name) > 255:
+        return 1
+
+    if _allowed.fullmatch(host_name):
+        return 0
+
+    # It is not a valid host name, but could be a valid IP address.
+    try:
+        ipaddress.ip_address(host_name)
+    except ValueError:
+        return 1
+
+    return 0
--- a/lib/pbench/server/api/resources/upload_api.py
+++ b/lib/pbench/server/api/resources/upload_api.py
@@ -7,13 +7,11 @@ from pathlib import Path
 import humanize
 from flask import jsonify, request
 from flask_restful import Resource, abort
-from werkzeug.utils import secure_filename
 
+from pbench.common.utils import validate_hostname
 from pbench.server.api.auth import Auth
 from pbench.server.database.models.tracker import Dataset, DatasetDuplicate, States
 from pbench.server.utils import filesize_bytes
-
-ALLOWED_EXTENSIONS = {"xz"}
 
 
 class HostInfo(Resource):
@@ -34,12 +32,15 @@ class HostInfo(Resource):
             self.logger.exception(
                 "There was something wrong constructing the host info"
             )
-            abort(500, message="INTERNAL ERROR")
-        response.status_code = 200
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
+        response.status_code = HTTPStatus.OK
         return response
 
 
 class Upload(Resource):
+    ALLOWED_EXTENSION = ".tar.xz"
+    CHUNK_SIZE = 65536
+
     def __init__(self, config, logger):
         self.config = config
         self.logger = logger
@@ -50,70 +51,85 @@ class Upload(Resource):
         )
 
     @Auth.token_auth.login_required()
-    def put(self, controller):
-
+    def put(self, filename: str):
         try:
             username = Auth.token_auth.current_user().username
         except Exception:
-            self.logger.exception("Tarfile upload: Exception verifying the username")
+            self.logger.exception("Error verifying the username")
             abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
-        if not request.headers.get("filename"):
-            self.logger.debug(
-                "Tarfile upload: Post operation failed due to missing filename header"
+        if os.path.basename(filename) != filename:
+            msg = "File must not contain a path"
+            self.logger.warning(
+                "{} for user = {}, file = {!a}", msg, username, filename,
             )
-            abort(
-                400,
-                message="Missing filename header, POST operation requires a filename header to name the uploaded file",
-            )
-        filename = secure_filename(request.headers.get("filename"))
+            abort(HTTPStatus.BAD_REQUEST, message=msg)
 
-        if not request.headers.get("Content-MD5"):
-            self.logger.debug(
-                f"Tarfile upload: Post operation failed due to missing md5sum header for file {filename}"
+        if not self.supported_file_extension(filename):
+            msg = f"File extension not supported, must be {self.ALLOWED_EXTENSION}"
+            self.logger.warning(
+                "{} for user = {}, file = {!a}", msg, username, filename,
             )
-            abort(
-                400,
-                message="Missing md5sum header, POST operation requires md5sum of an uploaded file in header",
+            abort(HTTPStatus.BAD_REQUEST, message=msg)
+
+        controller = request.headers.get("controller")
+        if not controller:
+            msg = "Missing required controller header"
+            self.logger.warning(
+                "{} for user = {}, file = {!a}", msg, username, filename
             )
+            abort(HTTPStatus.BAD_REQUEST, message=msg)
+        if validate_hostname(controller) != 0:
+            msg = "Invalid controller header"
+            self.logger.warning(
+                "{} for user = {}, ctrl = {!a}, file = {!a}",
+                msg,
+                username,
+                controller,
+                filename,
+            )
+            abort(HTTPStatus.BAD_REQUEST, message=msg)
+
         md5sum = request.headers.get("Content-MD5")
-
-        if not self.allowed_file(filename):
-            self.logger.debug(
-                f"Tarfile upload: Bad file extension received for file {filename}"
+        if not md5sum:
+            msg = "Missing required Content-MD5 header"
+            self.logger.warning(
+                "{} for user = {}, ctrl = {!a}, file = {!a}",
+                msg,
+                username,
+                controller,
+                filename,
             )
-            abort(400, message="File extension not supported. Only .xz")
+            abort(HTTPStatus.BAD_REQUEST, message=msg)
 
+        status = HTTPStatus.OK
         try:
-            content_length = int(request.headers.get("Content-Length"))
+            content_length = int(request.headers["Content-Length"])
+        except KeyError:
+            msg = "Missing required Content-Length header"
+            status = HTTPStatus.LENGTH_REQUIRED
         except ValueError:
-            self.logger.debug(
-                f"Tarfile upload: Invalid content-length header, not an integer for file {filename}"
-            )
-            abort(400, message="Invalid content-length header, not an integer")
-        except Exception:
-            self.logger.debug(
-                f"Tarfile upload: No Content-Length header value found for file {filename}"
-            )
-            abort(400, message="Missing required content-length header")
+            msg = f"Invalid Content-Length header, not an integer ({content_length})"
+            status = HTTPStatus.BAD_REQUEST
         else:
-            if content_length > self.max_content_length:
-                self.logger.debug(
-                    f"Tarfile upload: Content-Length exceeded maximum upload size allowed. File: {filename}"
+            if not (0 < content_length <= self.max_content_length):
+                msg = "Content-Length ({}) must be greater than 0 and no greater than {}".format(
+                    content_length, humanize.naturalsize(self.max_content_length)
                 )
-                abort(
-                    400,
-                    message=f"Payload body too large, {content_length:d} bytes, maximum size should be less than "
-                    f"or equal to {humanize.naturalsize(self.max_content_length)}",
+                status = (
+                    HTTPStatus.REQUEST_ENTITY_TOO_LARGE
+                    if 0 < content_length
+                    else HTTPStatus.BAD_REQUEST
                 )
-            elif content_length == 0:
-                self.logger.debug(
-                    f"Tarfile upload: Content-Length header value is 0 for file {filename}"
-                )
-                abort(
-                    400,
-                    message="Upload failed, Content-Length received in header is 0",
-                )
+        if status != HTTPStatus.OK:
+            self.logger.warning(
+                "{} for user = {}, ctrl = {!a}, file = {!a}",
+                msg,
+                username,
+                controller,
+                filename,
+            )
+            abort(status, message=msg)
 
         path = self.upload_directory / controller
         path.mkdir(exist_ok=True)
@@ -128,35 +144,52 @@ class Upload(Resource):
             )
             dataset.add()
         except DatasetDuplicate:
-            self.logger.info("Dataset already exists {}", filename)
+            self.logger.info(
+                "Dataset already exists, user = {}, ctrl = {!a}, file = {!a}",
+                username,
+                controller,
+                filename,
+            )
             response = jsonify(dict(message="Dataset already exists"))
-            response.status_code = 200
+            response.status_code = HTTPStatus.OK
             return response
         except Exception:
-            self.logger.exception("unable to create dataset for {}", filename)
-            abort(
-                HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR",
-            )
-
-        if tar_full_path.is_file() or md5_full_path.is_file():
-            self.logger.error(
-                "Dataset or corresponding md5 file already present on the disc, {}",
+            self.logger.exception(
+                "unable to create dataset for user = {}, ctrl = {!a}, file = {!a}",
+                username,
+                controller,
                 filename,
             )
             abort(
                 HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR",
             )
 
-        self.logger.info("Uploading file {} to {}", filename, dataset)
+        if tar_full_path.is_file() or md5_full_path.is_file():
+            self.logger.error(
+                "Dataset, or corresponding md5 file, already present; tar {} ({}), md5 {} ({})",
+                tar_full_path,
+                "present" if tar_full_path.is_file() else "missing",
+                md5_full_path,
+                "present" if md5_full_path.is_file() else "missing",
+            )
+            abort(
+                HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR",
+            )
+
+        self.logger.info(
+            "Uploading file {!a} (user = {}, ctrl = {!a}) to {}",
+            filename,
+            username,
+            controller,
+            dataset,
+        )
 
         with tempfile.NamedTemporaryFile(mode="wb", dir=path) as ofp:
-            chunk_size = 4096
-            self.logger.debug("Writing chunks")
             hash_md5 = hashlib.md5()
 
             try:
                 while True:
-                    chunk = request.stream.read(chunk_size)
+                    chunk = request.stream.read(self.CHUNK_SIZE)
                     bytes_received += len(chunk)
                     if len(chunk) == 0 or bytes_received > content_length:
                         break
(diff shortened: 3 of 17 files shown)


portante (commented):
Ready for review, though we need to have PR #2264 land before we open for final review.

lib/pbench/agent/results.py
```diff
@@ -149,23 +149,25 @@ class CopyResultTb:
     """CopyResultTb - Use the server's HTTP PUT method to upload a tarball
     """
 
-    chunk_size = 4096
+    CHUNK_SIZE = 65536
 
     def __init__(
         self, controller: str, tarball: str, config: PbenchAgentConfig, logger: Logger
     ):
+        self.controller = secure_filename(controller)
```
dbutenhof, line 157: You know, I've always kinda taken this `secure_filename` thing for granted, but seeing it on what's really "just a string" made me think about it. What we really want here is an HTML-sanitized string. While `secure_filename`'s "pure ascii" signature is similar, I wonder if this is really exactly what we want? Not a big deal here, I think, in any case ...

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -40,6 +38,9 @@ def get(self):
 
 
 class Upload(Resource):
+    ALLOWED_EXTENSION = ".tar.xz"
+    CHUNK_SIZE = 65536
```
dbutenhof, line 42: Since you've tweaked the spelling of `CHUNK_SIZE` on agent and server side ... makes me wonder whether we really need to bump this into a common constants module so they're guaranteed to be in sync. (Well, "mostly guaranteed" to be in sync, since agent and server are still built separately...)

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -50,70 +51,100 @@ def __init__(self, config, logger):
...
+            msg = "Missing required controller header"
+            self.logger.warning(
+                "{} for user = {}, file = {}", msg, username, filename_arg
             )
-            abort(
-                400,
-                message="Missing md5sum header, POST operation requires md5sum of an uploaded file in header",
-            )
-        md5sum = request.headers.get("Content-MD5")
+            abort(HTTPStatus.BAD_REQUEST, message=msg)
```
dbutenhof, line 78: Repeated pattern here for all the required headers; this could be a helper method that takes the header name and returns the value.

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -88,109 +88,139 @@ def test_json_object(client, caplog, server_config):
...
+        if md5:
+            headers["Content-MD5"] = md5
+        return headers
+
+    def test_missing_authorization_header(self, client, caplog, server_config):
+        response = client.put(self.gen_uri(server_config))
         assert response.status_code == 401
         for record in caplog.records:
-            assert record.levelname not in ("ERROR", "CRITICAL")
+            assert record.levelname not in ("DEBUG", "ERROR", "CRITICAL")
```
dbutenhof, line 114: Hmm. If we know that `DEBUG` level isn't enabled for the unit tests, then including it is pointless. Otherwise... I'm not sure we should ever disallow `DEBUG` to be used freely where a developer thinks it might be useful. This seems like more of a test of the `Logging` system (we shouldn't have debug enabled) than of the interface...

lib/pbench/agent/results.py
webbnh: ["The function might return an empty filename."](https://tedboy.github.io/flask/generated/werkzeug.secure_filename.html)

I'm sure it only happens for edge cases, but...we don't check for it....

webbnh: What we probably should be doing is wrapping `secure_filename()` in a utility function:  this would allow us to do things like check it before we return it, and even completely replace the underlying mechanism, without affecting the calling code.

lib/pbench/server/api/resources/upload_api.py
webbnh: Yes, please.  However, I think it will need to take more than just the header name -- it'll presumably want the `msg`, `username`, `filename_arg`, and HTTP status code, as well.  :-/  And, I'm not sure how I feel about burying the `abort()` call inside it -- that should probably be done at the top level.  :-(

And, this call is different from the subsequent ones, as it omits the controller from the message (although, that could be finessed).

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -128,35 +159,51 @@ def put(self, controller):
...
             )
 
-        self.logger.info("Uploading file {} to {}", filename, dataset)
+        self.logger.info(
+            "Uploading file {} (user = {}, ctrl = {}) to {}",
+            username,
+            controller,
+            filename,
+            dataset,
+        )
```
webbnh, line 198: I think these parameters are mis-ordered.

lib/pbench/agent/results.py
```diff
@@ -149,23 +149,25 @@ class CopyResultTb:
...
+        self.upload_url = f"{server_rest_url}/upload/{tbname}"
+        self.logger = logger
 
     def read_in_chunks(self, file_object: IO) -> Iterator[bytes]:
-        data = file_object.read(self.chunk_size)
+        data = file_object.read(self.CHUNK_SIZE)
         while data:
             yield data
-            data = file_object.read(self.chunk_size)
+            data = file_object.read(self.CHUNK_SIZE)
```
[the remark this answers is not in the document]
webbnh: :-D  We use the first idiom above in `lib/pbench/server/api/resources/upload_api.py` at line 205.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -128,35 +159,51 @@ def put(self, controller):
...
 
         with tempfile.NamedTemporaryFile(mode="wb", dir=path) as ofp:
-            chunk_size = 4096
             self.logger.debug("Writing chunks")
             hash_md5 = hashlib.md5()
 
             try:
                 while True:
-                    chunk = request.stream.read(chunk_size)
+                    chunk = request.stream.read(self.CHUNK_SIZE)
```
webbnh, line 206: Instead of `self.CHUNK_SIZE`, should this be `Upload.CHUNK_SIZE`?  (A la the reference to `ALLOWED_EXTENSION` in `allowed_file()`....)  Ditto line 69.

lib/pbench/test/unit/agent/task/test_results_push.py
```diff
@@ -17,26 +18,18 @@ class TestResultsPush:
     URL = "http://pbench.example.com/api/v1"
 
     @staticmethod
-    def add_success_mock_response():
+    def add_http_mock_response(code: int = 200):
```
webbnh, line 21: Are we happy with the numeric literal for the HTTP status code here (and at line 173), or would we rather we used a symbolic reference?

lib/pbench/test/unit/server/test_requests.py
webbnh: I concur.  Moreover, I've been running the unit tests individually under my IDE, so I'm not aware of any reason why `DEBUG` level would be disabled in that case.  So, I second the notion that we should be permitting `DEBUG` to be used wherever we see fit and that the unit tests should not be applying back-pressure to that.

I'm not opposed to the unit tests checking the logs.  But they should only be checking for things which must or must not be there...and, generally speaking, I don't think that `"DEBUG"` should be in either list.

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -88,109 +88,139 @@ def test_json_object(client, caplog, server_config):
...
+        assert caplog.records[-1].levelname == "WARNING"
 
-    @staticmethod
```
webbnh, line 101: Just out of curiosity, why remove the `@staticmethod`?  (The only reference that I see to `self` here is the invocation of `gen_uri()` which is also a static method, so you can invoke it as `TestUpload.gen_uri()` without having to make this an instance method.)

webbnh (changes requested):
We should probably have a, hopefully brief, real-time discussion about how to pass the two parameters to the upload API.

It might also be good, now that we've started reviewing this, if you didn't do force-pushes.

lib/pbench/test/unit/server/test_requests.py
portante: I got into a battle with `pytest` on how to use instance attributes during tests.  I nearly lost, but the result of that skirmish was this change.  I think I could change it back, but am really not up for dealing with that again.

lib/pbench/server/api/resources/upload_api.py
portante: Class variables can be accessed using `self`.

lib/pbench/server/api/resources/upload_api.py
portante: Yes, along with a number of other errors with logged output!

lib/pbench/server/api/resources/upload_api.py
portante: I am not in favor of explicitly tying the agent and server to the same size.  We might want even larger buffers on the server side in the future to take advantage of available memory.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -50,70 +51,100 @@ def __init__(self, config, logger):
...
+                HTTPStatus.BAD_REQUEST,
+                message="File extension not supported, must be"
+                f" {self.ALLOWED_EXTENSION}",
             )
-        filename = secure_filename(request.headers.get("filename"))
 
-        if not request.headers.get("Content-MD5"):
-            self.logger.debug(
-                f"Tarfile upload: Post operation failed due to missing md5sum header for file {filename}"
+        controller_arg = request.headers.get("controller")
```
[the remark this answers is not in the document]
portante: I tried to explain that in the commit message.  Let me know if the updated version helps or makes it more muddled.

lib/pbench/server/api/resources/upload_api.py
portante: I agree it is a repeated pattern, but it won't save us much with all the arguments that have to be provided.

lib/pbench/agent/results.py
portante: We really don't need to use `secure_filename` on the agent side.  We only need that on the server side.  We have a secure file name on the agent side because we derive it from a file system.

We do need to `urllib.parse.quote()` it.

lib/pbench/test/unit/server/test_requests.py
portante: If we make a code change that enabled `DEBUG` logging inadvertently, then we could potentially cause a server failure in production.

For agent side code I agree, but when it comes to the server, we want to know exactly what is being logged.

If we want to add `DEBUG` log entries to the code, the unit tests should then be run with `DEBUG` enabled so that we allow those.

But I have been on the operational end of production systems going down because somebody mistakenly emitted `DEBUG` logs in the worse possible place, but was never caught in testing.

The upload API is particularly critical because we can potentially have many more clients concurrently than operations performed from the browser.

lib/pbench/agent/results.py
```diff
@@ -149,23 +149,26 @@ class CopyResultTb:
     """CopyResultTb - Use the server's HTTP PUT method to upload a tarball
     """
 
-    chunk_size = 4096
+    CHUNK_SIZE = 65536
 
     def __init__(
         self, controller: str, tarball: str, config: PbenchAgentConfig, logger: Logger
     ):
+        self.controller = controller
```
dbutenhof, line 157: Do we need to protect the character set for header values? Seems we're basically allowed to use any ASCII octets, including LWS (linear whitespace), so full HTTP encoding would be inappropriate; but a scan for non-ASCII characters in the controller name wouldn't be unreasonable. (Especially as we transition from "filename" to "metadata".)

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -50,70 +52,113 @@ def __init__(self, config, logger):
...
             username = Auth.token_auth.current_user().username
         except Exception:
-            self.logger.exception("Tarfile upload: Exception verifying the username")
+            self.logger.exception("Error verifying the username")
             abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
-        if not request.headers.get("filename"):
-            self.logger.debug(
-                "Tarfile upload: Post operation failed due to missing filename header"
+        filename = secure_filename(urllib.parse.unquote(filename_arg))
```
dbutenhof, line 62: Your addition of `urllib.parse.quote` for the filename is interesting; I'd thought requests would handle that automatically, but a bit of research suggests that you're right. However I believe that Flask _does_ handle decoding properly when extracting the templated parameter on the receiving end, so I think that you're double-decoding here. Unit testing to confirm this might not be a bad idea.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -128,35 +173,53 @@ def put(self, controller):
...
+                "present" if tar_full_path.is_file() else "missing",
+                md5_full_path,
+                "present" if md5_full_path.is_file() else "missing",
             )
             abort(
                 HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR",
             )
 
-        self.logger.info("Uploading file {} to {}", filename, dataset)
+        self.logger.info(
```
dbutenhof, line 208: Note that this is essentially the audit log. I could definitely see a unit test checking that this _particular_ INFO message is present except where a WARN or ERROR is expected ...

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -227,18 +303,13 @@ def put(self, controller):
             self.logger.exception("Unable to finalize {}", dataset)
             abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
         response = jsonify(dict(message="File successfully uploaded"))
-        response.status_code = 201
+        response.status_code = HTTPStatus.CREATED
         return response
 
     @staticmethod
-    def allowed_file(filename):
+    def allowed_file(filename: str) -> str:
```
dbutenhof, line 310: `-> bool` ???

dbutenhof (approved):

lib/pbench/agent/results.py
portante: We should be covered if we first `validate_hostname(controller)`.

lib/pbench/server/api/resources/upload_api.py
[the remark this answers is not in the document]
portante: I'd also like to be able to simple drop a header in the future then to change how to construct the URL.

lib/pbench/server/api/resources/upload_api.py
webbnh: Yes...they _can_...but that doesn't mean that they _should_....

For ones which are functionally constants, class attributes are better accessed via the class name -- this makes it clear what their scope and usage is.  (If the values are actually instance attributes which happen to be initialized at the class level, then obviously its appropriate (and presumably necessary) to access them via `self`, but that's not the use case here.)

Beyond that, it would be good to be consistent with the reference in `allowed_file()` (although, I realize that there is no choice there, since that is a static method) -- otherwise, you make the reader work extra hard to establish that the two references reach the same value.

lib/pbench/agent/results.py
webbnh: Just to confirm, "as we transition from 'filename' to 'metadata'", are we sure we want to constrain ourselves to valid hostnames for "controller" values?  (We want to outlaw colons, for instance?)

lib/pbench/test/unit/server/test_requests.py
webbnh: > I have been on the operational end of production systems going down because somebody mistakenly emitted `DEBUG` logs in the worse possible place, but was never caught in testing.

This is an excellent point.

However, is this assertion in the unit test actually going to catch that case?  E.g., if the logger used in unit test execution has `DEBUG` level disabled, then the test assertion won't detect the problem which would arise if `DEBUG` level logging were inadvertently enabled in production.

If we really want to avoid `DEBUG` level logging in production, then we probably need something like a Git hook which makes sure that it doesn't get merged into the source.  It's not clear to me that clogging up the unit tests with this gets us to where we need to be.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -271,13 +287,13 @@ def put(self, filename_arg):
                 try:
                     os.remove(md5_full_path)
```
webbnh, line 288: If we die trying to create the link file at line 285 we carefully try to remove the hash file, but we don't touch the uploaded data file or the `Dataset` DB entry.  It seems like we should be using a more structured approach here with nested operations which can be backed out in the event of problems.

webbnh (dismissed):
Most of the below are nits and small suggestions; however, I found a couple of items which are a concern...but, we could consider addressing them in a separate PR.

The concerns are centered around how the upload deals with errors and restoring external invariants, of which I think there are four:   the `Dataset` DB entry, the uploaded data file, the MD5 hash file, and a symbolic link to the data.  In the event of a problem, we remove or avoid creating the last two, but we leave one or both of the first two hanging around.  Doesn't this cause problems?  In particular, it seems like it would result in errors if the upload were re-tried.

lib/pbench/agent/results.py
portante: We already validate host names for `pbench-move-results`.

We use `validate_hostname()` here on the agent side, just before sending things off to the server, to check that we are not going to send the server a bad value (see the upcoming commit).  This gives the user feedback sooner than if it fails on the server side.

As you can imagine, the server needs to validate API parameters regardless.  In other words, this change does not in any way relieve the server of its responsibility of validating API parameters.

lib/pbench/server/api/resources/upload_api.py
portante: Thanks for pointing this out regarding Flask behavior.

I also found out that `secure_filename()` is not going to work for us.  It munges the name so that it does not match the metadata contents.  We'd have to use `secure_filename` when making the tar ball as well.

Instead, I'll just make sure the unquoted file name does not have any path elements.  If the filename is invalid, we'll get an error trying to create it further along.

lib/pbench/server/api/resources/upload_api.py
portante: I like the thinking, but for one log statement in this PR?

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -77,6 +81,16 @@ def put(self, filename_arg):
             )
             abort(HTTPStatus.BAD_REQUEST, message=msg)
         controller = secure_filename(controller_arg)
+        if not controller:
+            msg = "Invalid controller header"
+            self.logger.warning(
+                "{} for user = {}, ctrl = {}, file = {}",
```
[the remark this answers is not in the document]
portante: Good point.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -166,7 +180,7 @@ def put(self, filename_arg):
                 filename,
```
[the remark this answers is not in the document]
portante: No longer an issue.

lib/pbench/server/api/resources/upload_api.py
portante: This is a real problem, thanks for catching that.  I'll work in as another commit.

portante: Actually, not sure this is a problem.

The `with` statement _always_ removes the uploaded data file using its "named" temporary entry.  That removal will only delete the file on disk if there is no second link to it.  So once the `os.link()` is successful, the file entry is in its final resting place.

The Dataset is always a permanent record, it is never removed.  So it will live in a partially uploaded state.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -60,12 +60,12 @@ def put(self, filename: str):
...
         if os.path.basename(filename) != filename:
             msg = "File must not contain a path"
-            self.logger.warning("{} for {}, user = {}", msg, filename, username)
+            self.logger.warning("{} for {!a}, user = {}", msg, filename, username)
             abort(HTTPStatus.BAD_REQUEST, message=msg)
 
         if not self.allowed_file(filename):
             self.logger.warning(
-                "Unsupported file extension for {}, user = {}", filename, username,
+                "Unsupported file extension for {!a}, user = {}", filename, username,
```
dbutenhof, line 68: It occurs to me this message is potentially a bit fragile. The method is named `allowed_file` and one could imagine, for example, expanding the function to check for legal characters in the string in addition to just checking the suffix, while the message (which is at best loosely coupled with the function) specifically says "file extension".

Not really a big deal at this point, but generalizing the message to better match the method name (or tightening the implied scope of the method name to match the message) might possibly reduce someone's confusion in the future.

Also, since you fall into a pattern later of user, controller, filename, it'd be more consistent to use user, filename here even though we haven't yet established controller.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -224,7 +226,7 @@ def put(self, filename: str):
                     hash_md5.update(chunk)
             except Exception:
                 self.logger.exception(
-                    "Unexpected error uploading {}, user = {}, ctrl = {}",
+                    "Unexpected error uploading {!a}, user = {}, ctrl = {!a}",
```
dbutenhof, line 229: Breaking the user, ctrl, file pattern here makes a bit more sense since you're specifically calling out a problem with uploading the file... however I also kinda like the idea of having a consistent ordering through the method, and it wouldn't be too bad. You already say "Uploading file" for the pre-load INFO, so "Unexpected error uploading file" here wouldn't be out of place and putting the filename at the end would be fine.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -181,19 +195,21 @@ def put(self, filename_arg):
...
-                "Dataset or corresponding md5 file already present on the disc, {}, {}",
+                "Dataset, or corresponding md5 file, already present; tar {} ({}), md5 {} ({})",
                 tar_full_path,
+                "present" if tar_full_path.is_file() else "missing",
                 md5_full_path,
+                "present" if md5_full_path.is_file() else "missing",
             )
             abort(
                 HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR",
             )
```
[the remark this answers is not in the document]
dbutenhof: True; "worst case" error recovery is not something we've done consistently.

To what extent we're willing to roll back broken invariants ... or how we define invariants in the first place ... is a big deal.

In the case of the Dataset, we retain datasets in QUARANTINE state for rejected tarballs, but you're right that a tarball that didn't even get uploaded in the first place could be retried, and allowing that to work is a good idea. That would require either removing the Dataset, or dealing with a duplicate here in some special way. For example, we could create an UPLOAD_FAILURE state (or metadata) which would be allowed for a subsequent upload, which would just clear the bad metadata and complete. We'd have the record of the attempt in the meantime if anyone looked.

lib/pbench/server/api/resources/upload_api.py
webbnh: Perhaps not for only one statement (although, once we have the mechanism, it'll be easy to use in other places which should have it!), but is there only _one_ in these changes?  I got the impression that there were several (but maybe that assumption was misplaced...).

webbnh (commented):
Looks reasonable.  Below is the usual pile of nits and suggestions.

I think you should consider moving the creation of the `Dataset` entry during the upload to just before (or just after) where the code creates the hard-link to the successfully-uploaded tarball.  I think that will go a very long way toward ensuring that the external invariants are not violated, and we can figure out how to create the link and the DB entry in an atomic way in a subsequent PR.

Aside from that, I think it just needs a little polishing.

lib/pbench/common/utils.py
```diff
@@ -20,3 +22,43 @@ def md5sum(filename):
...
+# comparison seeking a letter-or-digit, optionally followed by a sequence of 0
+# to 61 letter-or-digit-or-hyphens followed by a letter-or-digit, follwed by 0
+# or more instances of that same pattern preceded by a period.
+_allowed = re.compile(
+    r"[A-Z0-9](?:[A-Z0-9\-]{0,61}[A-Z0-9])?(?:\.[A-Z0-9](?:[A-Z0-9\-]{0,61}[A-Z0-9])?)*",
+    flags=re.IGNORECASE,
+)
+
+
+def validate_hostname(host_name: str) -> int:
```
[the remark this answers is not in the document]
portante: Tracked in #2298.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -181,19 +195,21 @@ def put(self, filename_arg):
 
         if tar_full_path.is_file() or md5_full_path.is_file():
             self.logger.error(
-                "Dataset or corresponding md5 file already present on the disc, {}, {}",
+                "Dataset, or corresponding md5 file, already present; tar {} ({}), md5 {} ({})",
                 tar_full_path,
+                "present" if tar_full_path.is_file() else "missing",
                 md5_full_path,
+                "present" if md5_full_path.is_file() else "missing",
             )
```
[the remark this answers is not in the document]
portante: In principle I agree with reducing lines of code, the density of either of these is very high compared to what is proposed in this PR.  The readability is lost on my for those two, while I can easily understand the format and the arguments in the current form.

lib/pbench/server/api/resources/upload_api.py
portante: Tracked in #2299.

lib/pbench/server/api/resources/upload_api.py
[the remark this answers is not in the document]
portante: Tracked in #2299.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -128,35 +172,52 @@ def put(self, controller):
...
-                filename,
+                "Dataset, or corresponding md5 file, already present; tar {} ({}), md5 {} ({})",
+                tar_full_path,
+                "present" if tar_full_path.is_file() else "missing",
+                md5_full_path,
+                "present" if md5_full_path.is_file() else "missing",
             )
             abort(
                 HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR",
             )
```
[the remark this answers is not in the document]
portante: Tracked in #2299.

portante (commented):
Polished based on code-review feedback; issue #2299 tracking the file create sequence and `Dataset` updates.

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -103,61 +102,39 @@ def put(self, filename: str):
...
+                    mcl = humanize.naturalsize(self.max_content_length)
+                    msg = f"Content-Length greater than maximum allowed size ({cl} > {mcl})"
+                    status = HTTPStatus.REQUEST_ENTITY_TOO_LARGE
+                elif content_length == 0:
+                    msg = "Invalid Content-Length header, zero (0)"
+                    status = HTTPStatus.BAD_REQUEST
+                elif content_length < 0:
+                    msg = f"Invalid Content-Length header, negative ({content_length})"
+                    status = HTTPStatus.BAD_REQUEST
+        if msg:
```
dbutenhof, line 129: This is great, although if I'd coded it I think I'd've been more likely to use `if status:` ... both because that's the most important of the two, and because `None` is more "instantly intuitively self-obviously" falsey than `""`.

dbutenhof (dismissed):

dbutenhof (approved):

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -128,96 +149,124 @@ def put(self, controller):
...
+                msg = "Unexpected error encountered during file upload"
                 self.logger.exception(
-                    "Tarfile upload: There was something wrong uploading {}", filename
+                    "{} for user = {}, ctrl = {!a}, file = {!a}",
+                    username,
+                    controller,
+                    filename,
                 )
-                abort(500, message=f"There was something wrong uploading {filename}")
+                abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
```
dbutenhof, line 212: LGTM is complaining about multiple assignments to `msg` here. I'm assuming it's because it doesn't understand `abort`; but it's true that with the code as it is, this assignment to `msg` isn't used at all before the `abort`. Turns out that it's wrong but it's right anyway: you're missing the first parameter in the `self.logger.exception` here!

lib/pbench/server/api/resources/upload_api.py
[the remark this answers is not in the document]
webbnh: I concur that readability is important.  I'll also concede that readability is in the eyes of the reader.  However, I'm unclear on how you find your proposed code more readable than the first of mine:  they both use the same conditional expressions (or nearly so), but mine provides two fewer arguments and a simpler format string.

And, I believe that mine will produce a better result for the consumer of the output.

lib/pbench/server/api/resources/upload_api.py
webbnh: I'm fine with deferring the task of restoring invariants after an error; however, I was hoping that deferring the creation of the `Dataset` until after the upload (which would avoid much of the need for that restoration) would be simple enough to be included in this PR.  Is there some aspect of that which I'm missing?

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -50,70 +51,90 @@ def __init__(self, config, logger):
...
+                    status = HTTPStatus.BAD_REQUEST
+        if status:
+            self.logger.warning(
+                "{} for user = {}, ctrl = {!a}, file = {!a}",
+                msg,
+                username,
+                controller,
+                filename,
+            )
+            abort(status, message=msg)
```
webbnh, line 137: How about this?
```
        status = HTTPStatus.OK
        try:
            content_length = int(request.headers["Content-Length"])
        except KeyError:
            msg = "Missing required Content-Length header"
            status = HTTPStatus.LENGTH_REQUIRED
        except ValueError:
            msg = f"Invalid Content-Length header, not an integer ({content_length_hdr})"
            status = HTTPStatus.BAD_REQUEST
        else:
            if not (0 < content_length <= self.max_content_length):
                msg = ("Content-Length ({}) must be greater than 0 and no greater than {}"
                        .format(content_length, humanize.naturalsize(self.max_content_length))
                status = (HTTPStatus.REQUEST_ENTITY_TOO_LARGE
                        if 0 < content_length else HTTPStatus.BAD_REQUEST)
        if status != HTTPStatus.OK:
            self.logger.warning(
                "{} for user = {}, ctrl = {!a}, file = {!a}",
                msg,
                username,
                controller,
                filename,
            )
            abort(status, message=msg)
```
I don't think it's worth the time/effort/space to decode three kinds of length errors; however, since you seem to value it, this code does return different status values for files which are too large and files which have bad (non-positive) lengths.

Also, I de-humanized the request's content length:  aside from the fact that it seems like `naturalsize()` doesn't handle negative numbers well, it will truncate the value to a single decimal place, and I think it's probably more important to display the actual value.

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -74,123 +75,147 @@ def test_bad_request(client, caplog, server_config, requests_mock):
...
+        for record in caplog.records:
+            assert record.levelname not in ("DEBUG", "ERROR", "CRITICAL")
+        assert caplog.records[-1].levelname == "WARNING"
+
+    def test_missing_authorization_header(self, client, caplog, server_config):
+        response = client.put(self.gen_uri(server_config))
+        assert response.status_code == HTTPStatus.UNAUTHORIZED
+        for record in caplog.records:
+            assert record.levelname not in ("DEBUG", "ERROR", "CRITICAL")
+        assert caplog.records[-1].levelname == "WARNING"
```
webbnh, line 122: Looks like a good spot for `verify_logs()`.

webbnh (approved):
Looks good, modulo the issue that LGTM highlighted and Dave diagnosed.

However, apparently I was less clear in one of my earlier comments than I should have been, so I went ahead and proposed some alternate code below, if you're so inclined.  Also, I think found a spot that your refactoring missed.

So...when does this PR stop being a "draft"?

lib/pbench/server/api/resources/upload_api.py
portante: Filed as a separate commit as I am not convinced of this change just yet.

I have implemented what you requested above.

I think we want to record the `Dataset` in such a way as to know details of on going uploads (potentially updates the bytes received dynamically) so that users can see their uploads in process.  And so that they can see the failures from the UI.

We just need a way to know when it is okay for a new upload to over-write a previously failed one.

dbutenhof: > I think we want to record the `Dataset` in such a way as to know details of on going uploads (potentially updates the bytes received dynamically) so that users can see their uploads in process. And so that they can see the failures from the UI.

Yes, that was the idea; we start in UPLOADING and go to UPLOADED at the end of the upload. Now, there's nothing right now in the UI that's sensitive to the `Dataset` state ... but the whole state tracker was designed to allow us to add some smarts in that area. And, yes, it means we need more sophisticated error recovery than simply deferring everything to the end.

> We just need a way to know when it is okay for a new upload to over-write a previously failed one.

Indeed.

dbutenhof (dismissed):
Reluctantly approve: I'd rather get this in and finesse it later, but I don't like creating the Dataset in UPLOADED state because it reduces visibility. Ah, well; can't have everything.

riya-17 (approved):
Looks Good

dbutenhof (approved):
Again, not crazy about the deferred `Dataset` creation, but let's get it in and worry about this later.

dbutenhof (approved):

webbnh (approved):

lib/pbench/server/api/resources/upload_api.py
webbnh: Owing to the fact that I was only reviewing the deltas, I somehow missed the fact that we update the `Dataset` state after the upload completes...so, yes, my proposed change is _not_ a good idea, and Pete can (and did) roll it back.  (Vely clever of you, Peter, to keep it as a separate commit!)

I really thought that there wasn't any more code at the end of that function....  :-(

PR #2292, opened 2021-05-07
Add a role to install the EPEL repo

### LINKED ISSUE #1987: Add role to install the EPEL repo for both agent and server.
We depend on packages from EPEL, so we might as well automate the process.

ndokos:
Fixes #1987

Separately added to both server and agent roles.

Add it to the pbench_<agent|server>_install.yml playbooks.

Bump the version in galaxy.yml in preparation for a new upload
of the agent roles.

### DIFF
--- a/agent/ansible/pbench/agent/galaxy.yml
+++ b/agent/ansible/pbench/agent/galaxy.yml
@@ -9,7 +9,7 @@ namespace: pbench
 name: agent
 
 # The version of the collection. Must be compatible with semantic versioning
-version: 1.0.3
+version: 1.0.4
 
 # The path to the Markdown (.md) readme file. This path is relative to the root of the collection
 readme: README.md
--- a/agent/ansible/pbench/agent/roles/epel_repo_install/tasks/main.yml
+++ b/agent/ansible/pbench/agent/roles/epel_repo_install/tasks/main.yml
@@ -1,0 +1,7 @@
+---
+- name: install the EPEL repo
+  package:
+    name: https://dl.fedoraproject.org/pub/epel/epel-release-latest-{{ ansible_distribution_major_version }}.noarch.rpm
+    state: latest
+  when: ansible_distribution == "RedHat" or ansible_distribution == "CentOS"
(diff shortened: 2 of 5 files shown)


portante (approved):

agent/ansible/pbench/agent/roles/epel_repo_install/tasks/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: install the EPEL repo
+  package:
+    name: https://dl.fedoraproject.org/pub/epel/epel-release-latest-{{ ansible_distribution_major_version }}.noarch.rpm
+    state: latest
+  when: ansible_distribution == "RedHat" or ansible_distribution == "CentOS"
+
```
dbutenhof, line 7: I'm not fluent in Ansible Galaxy, and probably it's more trouble than it'd be worth to package a separate "collection"? But since this role (and maybe others) seem to be identical for Agent and Server sides, would it be possible/practical to publish a single role that can be referenced by both sides? (And potentially used elsewhere?)

portante: Track in #2293.

agent/ansible/pbench/agent/galaxy.yml
```diff
@@ -9,7 +9,7 @@ namespace: pbench
 name: agent
 
 # The version of the collection. Must be compatible with semantic versioning
-version: 1.0.3
+version: 1.0.4
```
webbnh, line 12: @ndokos, is there no versioning done for the server side?

ndokos: I should open an issue on that: the agent roles are on galaxy, but the server roles are not - yet. They need renaming and reorganization - and documentation - to satisfy galaxy requirements. So no versioning yet - but it's coming.

portante: Tracked in #2294.

PR #2295, opened 2021-05-10
Validate all passed in `_pbench_hostname_ip` values

portante:
Since `hostname -i` will return a space separated list of IP addresses for the given host name, validate all of them instead of assuming incorrectly that only one is returned.

I have not seen this behavior on Fedora derived systems, but on Debian, when you get a container, it auto creates an entry for the current host name in the container's view of its `/etc/hosts`, which has a value of `127.0.1.1` (yes, `.1.1`, go figure).  In this case, `hostname -i` will return the original IP address and the `127.0.1.1` address.

You can make this happen arbitrarily by modifying the `/etc/hosts` of some system to have an alternate IP for localhost or your official DHCP host name. Then `hostname -i` returns N IP addresses.

----

Added a second commit to fix the proper handling of the `hostname` output.

We now use the long options to `hostname` so that the ConfigParser behavior of lowercasing key names won't lose `-a` vs `-A`, or `-i` vs `-I`.  And while we are at it, we refactor to make sure there is one source of what we are collecting, adding a proper mock of the data collected, and add the other options for data that can be collected.

### DIFF
--- a/agent/base
+++ b/agent/base
@@ -119,11 +119,13 @@ validate_hostname "${_pbench_full_hostname}" "_pbench_full_hostname"
 if [[ -z "${_pbench_hostname_ip}" ]]; then
     export _pbench_hostname_ip=$(hostname -i)
 fi
-validate-ipaddress "${_pbench_hostname_ip}"
-if [[ ${?} -ne 0 ]]; then
-    error_log "[${script_name}] Invalid IP address: '${_pbench_hostname_ip}'"
-    exit 1
-fi
+for _ip in ${_pbench_hostname_ip}; do
+    validate-ipaddress "${_ip}"
+    if [[ ${?} -ne 0 ]]; then
+        error_log "[${script_name}] Invalid IP address: '${_ip}'"
+        exit 1
+    fi
+done
 
 if [[ -z "$ssh_opts" ]]; then
     ssh_opts=$(pbench-config ssh_opts results)
--- a/agent/base
+++ b/agent/base
@@ -109,15 +109,17 @@ function validate_hostname {
 }
 
 if [[ -z "${_pbench_hostname}" ]]; then
-    export _pbench_hostname=$(hostname -s)
+    export _pbench_hostname=$(hostname --short)
 fi
 validate_hostname "${_pbench_hostname}" "_pbench_hostname"
+
 if [[ -z "${_pbench_full_hostname}" ]]; then
-    export _pbench_full_hostname=$(hostname -f)
+    export _pbench_full_hostname=$(hostname --fqdn)
 fi
 validate_hostname "${_pbench_full_hostname}" "_pbench_full_hostname"
+
 if [[ -z "${_pbench_hostname_ip}" ]]; then
-    export _pbench_hostname_ip=$(hostname -i)
+    export _pbench_hostname_ip=$(hostname --ip-address)
 fi
 for _ip in ${_pbench_hostname_ip}; do
     validate-ipaddress "${_ip}"
--- a/agent/util-scripts/gold/test-client-tool-meister/test-53.txt
+++ b/agent/util-scripts/gold/test-client-tool-meister/test-53.txt
@@ -99,10 +99,14 @@ iterations = 0-iter-zero, 1-iter-one
 
 [controller]
 hostname = testhost.example.com
-hostname-s = agent.example.com
-hostname-f = agent.example.com
-hostname-i = agent.example.com
-hostname-a = agent.example.com
+hostname-alias = bond
+hostname-all-fqdns = agent.example.com bond.example.com
+hostname-all-ip-addresses = 192.168.0.1 172.0.0.1 127.0.0.1
+hostname-domain = example.com
+hostname-fqdn = agent.example.com
+hostname-ip-address = 172.0.0.1 127.0.0.1
+hostname-nis = hostname[mock]: Local domain name not set
+hostname-short = agent
 ssh_opts = -o StrictHostKeyChecking=no
 
 [run]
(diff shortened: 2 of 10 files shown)


agent/base
```diff
@@ -117,7 +117,7 @@ if [[ -z "${_pbench_full_hostname}" ]]; then
 fi
 validate_hostname "${_pbench_full_hostname}" "_pbench_full_hostname"
 if [[ -z "${_pbench_hostname_ip}" ]]; then
-    export _pbench_hostname_ip=$(hostname -i)
+    export _pbench_hostname_ip=$(hostname -i | awk '{print $1}')
```
dbutenhof, line 120: I had to do a `man` to see exactly what `hostname -i` is defined to do. Interestingly, it warns _Display  the  network  address(es)  of the host name. __Note that this works only if the host name can be resolved. Avoid using this option__; use hostname --all-ip-addresses instead._

This doesn't depend on DNS resolution, and I wonder if that'd be better in a cloud virtual deployment that might not have DNS?

portante: You can make this happen by modifying `/etc/hosts` to have an alternate IP for localhost or your official DHCP host name.  Then `hostname -i` returns N IP addresses.

webbnh (dismissed):
Tastes great -- less filling!

But, are you confident that there are no other places which will trip up if `_pbench_hostname_ip` contains more than one value?

dbutenhof (dismissed):
Sure; I still wonder if the non-DNS `-I` would be better in the long run, but this is at worst better than before. ;-)

webbnh (dismissed):

ndokos (changes requested):
The manpage for `-i` says:  "Avoid using this option; use hostname  --all-ip-addresses instead."
Maybe we should heed that?

ndokos:
Eh, I see that @dbutenhof already brought that up.

dbutenhof:
> The manpage for `-i` says: "Avoid using this option; use hostname --all-ip-addresses instead."
> Maybe we should heed that?

Pete's answer was that we're actually doing _both_ `hostname -i` and `hostname -I` to collect both variants; although, come to think of it, the same issues should affect both, in any case, because the difference appears to be that `-i` is subject to DNS resolution and `-I` (`--all-ip-addresses`) isn't. Do we need the loop on both?

portante:
The genesis for all this was from @hifzakh's work.  When working to analyze `uperf` runs, had to dig really deep to find all the IP addresses that were in use.  So the goal of capturing `hostname -i` AND `hostname -I` is to make sure we have both the IP addresses as they are resolved by DNS and as the current host knows them, so that the data is available to help analysis.

dbutenhof:
> The genesis for all this was from @hifzakh's work. When working to analyze `uperf` runs, had to dig really deep to find all the IP addresses that were in use. So the goal of capturing `hostname -i` AND `hostname -I` is to make sure we have both the IP addresses as they are resolved by DNS and as the current host knows them, so that the data is available to help analysis.

OK, there it is again. But this time I went looking for the `hostname -I` and I don't see it anywhere. Is that a statement of future goal? And if it (or a Python equivalent) is somewhere else I'm not finding, do we need similar validation for the returned values?

And if it's not currently anywhere ... don't we still have the potential issue that the documentation states that `hostname -i` should be retired in favor of `-I` or `--all-ip-addresses`? Or are we just closing our eyes and pretending not to see that? ;-)

webbnh (approved):
I'm still OK with this....

Nevertheless, to Dave's point, I see `-s`, `-f`, and `-i` used with `hostname`, but no `-I`.

ndokos:
The tool meister does seem to collect `hostname -I` information: https://github.com/distributed-system-analysis/pbench/blob/27d6dea306bc306ffd0970cc3fe5ac75fc94b3d3/lib/pbench/agent/tool_meister.py#L847 and https://github.com/distributed-system-analysis/pbench/blob/40c8318d443eb546e8439a7ce5badb994233fa0b/lib/pbench/agent/utils.py#L373 but I don;t know where they end up.

portante:
See https://github.com/distributed-system-analysis/pbench/blob/main/lib/pbench/agent/utils.py#L351, called by both the Tool Meister and Tool Data Sink.

portante:
And it is supposed to be here: https://github.com/distributed-system-analysis/pbench/blob/main/agent/util-scripts/gold/test-client-tool-meister/test-56.txt#L319 ...

ndokos:
Yeah, something funky is going on: I only see `hostname-i` in a couple of metadata.log files done with 0.71. And on my system, `hostname -i` returns one address, whereas `hostname -I` returns six addresses. The metadata.log entries have six addresses.

ndokos:
Seems ConfigParser folds case for option names (at least).

ndokos:
See https://stackoverflow.com/questions/19359556/configparser-reads-capital-keys-and-make-them-lower-case

portante (commented):
Okay, this looks to be a bit better now.  The original fix is in the first commit, the refactoring to address the issues with `hostname` collection in the 2nd commit.

agent/util-scripts/gold/test-client-tool-meister/test-53.txt
```diff
@@ -99,10 +99,14 @@ iterations = 0-iter-zero, 1-iter-one
...
-hostname-s = agent.example.com
-hostname-f = agent.example.com
-hostname-i = agent.example.com
-hostname-a = agent.example.com
+hostname-alias = bond
+hostname-all-fqdns = agent.example.com bond.example.com
+hostname-all-ip-addresses = 192.168.0.1 172.0.0.1 127.0.0.1
+hostname-domain = example.com
+hostname-fqdn = agent.example.com
+hostname-ip-address = 172.0.0.1 127.0.0.1
```
dbutenhof, line 107: Cool; now I finally see both!

dbutenhof (approved):

ndokos (approved):

agent/util-scripts/test-bin/hostname
```diff
@@ -1,4 +1,38 @@
...
+    print(ip_addrs)
+elif sys.argv[1] in ("-I", "--all-ip-addresses"):
+    print(all_ip_addrs)
+elif sys.argv[1] in ("-s", "--short"):
+    print(short)
+elif sys.argv[1] in ("-y", "--yp", "--nis"):
+    print("hostname[mock]: Local domain name not set")
+    exit_code = 1
+else:
+    print("hostname[mock]: unexpected hostname invocation", file=sys.stderr)
```
webbnh, line 36: You don't want to dump `sys.argv` here?  (You know it has at least one entry.)

Do you want to catch the case of more than one argument/option specified?

webbnh (dismissed):

dbutenhof (approved):

webbnh (approved):
GTG

PR #2296, opened 2021-05-10
Add support for PBENCH_ORCHESTRATE

portante:


webbnh (approved):
LGTM.

webbnh (dismissed):

Maxusmusti (changes requested):
I was curious to see what happened if an invalid value was set for PBENCH_ORCHESTRATE:
image
The arg parser does not seem to handle broken defaults gracefully

ndokos (approved):

portante (commented):
Thanks for catching that problem, @Maxusmusti!  Let's see if folks think this is a good approach to fix it.

dbutenhof (dismissed):
Weird that the environment variable option overrides `ArgumentParser` choice validation... that sounds like a bug. But, you seem to have worked around it at some cost in clarity. Sigh.

ndokos (dismissed):

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -716,6 +729,7 @@ def main(_prog: str, cli_params: Namespace) -> int:
             logger.error("required ssh command not in our PATH")
             return ReturnCode.MISSINGSSHCMD
     else:
+        assert cli_params.orchestrate == "existing"
```
webbnh, line 732: Single-parameter `assert`'s report an `AssertionError` with no other information.  I recommend using the two-parameter version, supplying a useful message (including the value of `cli_params.orchestrate`, hopefully delimited in quotes or otherwise set up to show embedded blanks).

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -995,8 +1007,8 @@ if __name__ == "__main__":
     parser.add_argument(
         "--orchestrate",
         dest="orchestrate",
-        default="create",
-        choices=("create", "existing"),
+        default=os.environ.get("PBENCH_ORCHESTRATE", _def_orchestrate_choice),
+        choices=sorted(_orchestrate_choices),
```
webbnh, line 1011: Perhaps it would be handy to ensure that `_orchestrate_choices` is already sorted, since two of the places where it is used it has to be sorted.

webbnh (dismissed):
Looks OK to me, though, of course, I have suggestions.

Maxusmusti (approved):
Tested, now exits gracefully with clearer message

dbutenhof (approved):

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -136,6 +136,10 @@ from pbench.agent.utils import (
 from pbench.common.utils import validate_hostname
 
 
+# The --orchestrate parameter default choice, and the full list of choices.
+_def_orchestrate_choice = "create"
+_orchestrate_choices = [_def_orchestrate_choice, "existing"]
```
webbnh, line 141: This doesn't necessarily need to gate the merge, but....

The value of `_orchestrate_choices` should be "complete" and not reference the default.  Instead, the default should select one of the values from the list (and there should be an assertion somewhere which ensures that the value is in the list).  The problem with the approach taken here is that the default cannot be changed without _also_ changing the list of choices (and naive attempts to do so will damage the list, by dropping a value and presumably duplicating another).

webbnh (dismissed):
You made me look again and I found something else.  :-)

portante:
Last rebase, only changed the way the "choices" are defined as requested.

ndokos (approved):

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -136,6 +136,14 @@ from pbench.agent.utils import (
...
 
 
+# The --orchestrate parameter default choice, and the full list of choices.
+_orchestrate_choices = ["create", "existing"]
+_def_orchestrate_choice = "create"
+assert _def_orchestrate_choice in _orchestrate_choices, (
+    f"Logic Bomb!  The default orchestrate choice, {_def_orchestrate_choice!r},"
+    f" is not one of the available choices, {_orchestrate_choices!r}"
+)
+
```
dbutenhof, line 146: At this point I'm not even going to make an off-hand comment that this is starting to look like an ENUM without the semantic rigor. I didn't say that out loud, did I? ;-)

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -646,6 +655,14 @@ def main(_prog: str, cli_params: Namespace) -> int:
         logger.error("invalid sysinfo option(s), '{}'", ",".join(bad_l))
         return ReturnCode.BADSYSINFO
 
+    if cli_params.orchestrate not in _orchestrate_choices:
+        logger.error(
+            "invalid --orchestrate directive, '%s', expected one of %s",
+            cli_params.orchestrate,
+            ", ".join(_orchestrate_choices),
+        )
+        return ReturnCode.INVALIDORCHESTRATE
```
dbutenhof, line 664: I wonder if this is because the "default" is immune to choice validation? And, if so, whether Click's environment variable support would avoid the need for this?

I didn't say that out loud, either ... did I?

dbutenhof (approved):

agent/util-scripts/pbench-tool-meister-start
portante: I've been resisting moving anything related to the CLI interfaces to the Tool Meister to `click` because I don't want it to maintain official CLI interfaces to the Tool Meister just yet.

I'll file a tracking issue for this, none the less.

agent/util-scripts/pbench-tool-meister-start
portante: Tracked in #2316.

agent/util-scripts/pbench-tool-meister-start
portante: Tracked in #2316.

dbutenhof: > I've been resisting moving anything related to the CLI interfaces to the Tool Meister to `click` because I don't want it to maintain official CLI interfaces to the Tool Meister just yet.

Indeed; which is why I didn't say any of that. But tracking it with an issue is good.

PR #2300, opened 2021-05-11
Add python3-psycopg2 require in the server spec file

### LINKED ISSUE #2094: Need proper psycopg2 RPMs added as pbench-server RPM dependencies
Server installation needs to install the psycopg2 module: the pbench-server service uses it to connect to the Postgres database. This can be done with a PyPI module: `pip3 install psycopg2-binary` or it could be done by installing the psycopg2 RPM (if available).

This will require a bit or research. Depending on the outcome, we either need to add the RPM as a requirement in the spec file, or add the PyPI module in requirements.txt.

ndokos:
Fixes #2094

Although `psycopg2` is in the `server/requirements.txt` file, we prefer to install it as part of the RPM installation of pbench-server. The `pip3` install phase of it will then be a no-op.

### DIFF
--- a/server/rpm/pbench-server.spec.j2
+++ b/server/rpm/pbench-server.spec.j2
@@ -15,7 +15,7 @@ Requires: python3 python3-devel gcc
 Requires:       policycoreutils
 %if 0%{?rhel} != 7
 Requires: policycoreutils-python-utils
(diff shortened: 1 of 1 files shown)


webbnh (approved):
I don't think this change makes anything worse....  :-)

PR #2304, opened 2021-05-12
Separate logger name from module name

portante:
We really don't want to have the module name "dotted" with the module name, they are separate ids, and need to be treated as such.

### DIFF
--- a/lib/pbench/common/logger.py
+++ b/lib/pbench/common/logger.py
@@ -168,7 +168,7 @@ def get_pbench_logger(caller, config):
 
         handler.setLevel(logging.DEBUG)
         if config.log_fmt is None:
-            logfmt = "{asctime} {levelname} {process} {thread} {name}.{module} {funcName} {lineno} -- {message}"
+            logfmt = "{asctime} {levelname} {process} {thread} {name} {module} {funcName} {lineno} -- {message}"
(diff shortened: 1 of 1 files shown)


dbutenhof (approved):
A small change, but it does clarify the interpretation of the path we see.

lib/pbench/common/logger.py
```diff
@@ -168,7 +168,7 @@ def get_pbench_logger(caller, config):
 
         handler.setLevel(logging.DEBUG)
         if config.log_fmt is None:
-            logfmt = "{asctime} {levelname} {process} {thread} {name}.{module} {funcName} {lineno} -- {message}"
+            logfmt = "{asctime} {levelname} {process} {thread} {name} {module} {funcName} {lineno} -- {message}"
```
webbnh, line 171: Could you try removing the parentheses in lines 160-162 and see if Blank provides a saner line wrapping?

webbnh (approved):

lib/pbench/common/logger.py
portante: See PR #2309.

webbnh: Excellent exercise of discipline!

Issue #2261, opened 2021-04-29
Server gunicorn should write access and error log


dbutenhof:
Right now when we're not running with a reverse proxy, we have no record of the access and error gunicorn server hits.

Gunicorn can configure `--access-logfile` and `--error-logfile` to write access and error logs to a file. I believe we can also configure it to use a Python logger.

portante:
Ideally we'll want to use the same logging configuration as what is specified in the `pbench-server.cfg` file.  This will let us log to `/dev/log` by default so that we don't have to worry about log file rotation in production.

dbutenhof:
> Ideally we'll want to use the same logging configuration as what is specified in the `pbench-server.cfg` file. This will let us log to `/dev/log` by default so that we don't have to worry about log file rotation in production.

Maybe; I like having the access and error logs (like httpd) easily accessible without searching through the general log. But you're right that we need rollover handling. It seems that gunicorn has hooks to work with `logrotate` that might be as easy to use as redirecting to our logger.

portante:
I agree that it is great to have all the logs in one place, but if we go forward with that, it would be an example of developer mentality impacting "Day 2" operational behaviors.  We really want to have that work the other way, "Day 2" operational considerations influencing development.

The `logrotate` is nice, but it is does not help with the proliferation of log files on a system, which is also not good for "Day 2" operational concerns.

Having logs flow through `journald` gives you the ability to filter the logs by log level or certain strings, while allowing for "off-hosting" to centralized logging solutions which provide rich capabilities.

For development, we have a mode where we send the logs to a log file, but send them to `/dev/log` (really `journald`) for production.

portante:
We should also consider changing our Apache setup to use `mod_journald`: https://httpd.apache.org/docs/trunk/mod/mod_journald.html.

portante:
Fixed by PR #2774.

PR #2307, opened 2021-05-12
Create /var/log/pbench on installation

### LINKED ISSUE #2303: gunicorn logs directory, `/var/log/pbench`, needs to be created by RPM owned by `pbench` user
The gunicorn logs directory, `/var/log/pbench`, needs to be created by the RPM, owned by `pbench` user, on install.

ndokos:
Fixes #2303


webbnh:
Given that #2303 was closed in favor of #2261, do you want to continue with this PR?

portante:
Let's put this on hold until we resolve what we are going to do for #2261.

dbutenhof:
I suppose I should apologize. I thought I was fairly clear when I hacked the gunicorn startup in `shell.py` to add logging yesterday, if only because I'd done the same thing the previous jam session and I clearly remember mumbling about trying to find my notes. Apparently that was a good deal less obvious than I thought at the time!

PR #2308, opened 2021-05-13
Ensure that tarball upload gets desired HTTP headers

webbnh:
In today's Jam Session, we ran into a curious problem with tarball upload.  As best as we could discern, the Flask server used by `gunicorn` was trying to parse the upload as though it were being sent with `Transfer-Encoding: chunked`, but the stream was clearly not formatted that way.  This lead to a great deal of sleuthing around the agent's use of the Python `requests` package.

Per [RFC 2616](http://pretty-rfc.herokuapp.com/RFC2616#message.length), a request must not contain both `Content-Length` and `Transfer-Encoding` headers; however, the server would like to receive the `Content-Length` header, but the `requests` package may opt to generate the `Transfer-Encoding` header instead, depending on how the data for the request is specified.

This commit replaces the previous generator-based data source approach (which prompts `requests` to add the `Transfer-Encoding` header) with a straightforward "file-like" data source (which prompts `requests` to instead add the `Content-Length` header).  It also adds a few assertions to the code to make ensure that the headers come out the way we expect.

The result passes the existing unit tests (which doesn't prove a _great_ deal, since the previous code passed them, too...), and it also works with the server from the Jam Session using a 29532560 byte file.

### DIFF
--- a/lib/pbench/agent/results.py
+++ b/lib/pbench/agent/results.py
@@ -194,9 +194,18 @@ class CopyResultTb:
                 # we got what we want before we send the request.  Also,
                 # confirm that the contents of the Content-Length header
                 # is what we expect.
-                assert "Transfer-Encoding" not in request.headers
-                assert "Content-Length" in request.headers
-                assert request.headers["Content-Length"] == str(content_length)
+                assert (
+                    "Transfer-Encoding" not in request.headers
+                ), "Upload request unexpectedly contains a `Transfer-Encoding` header"
+                assert (
+                    "Content-Length" in request.headers
+                ), "Upload request unexpectedly missing a `Content-Length` header"
+                assert request.headers["Content-Length"] == str(content_length), (
+                    "Upload request `Content-Length` header contains {} -- "
+                    "expected {}".format(
+                        request.headers["Content-Length"], content_length
+                    )
+                )
 
                 response = requests.Session().send(request)
                 response.raise_for_status()
--- a/lib/pbench/agent/results.py
+++ b/lib/pbench/agent/results.py
@@ -6,7 +6,6 @@ import urllib.parse
 from configparser import ConfigParser
 from logging import Logger
 from pathlib import Path
-from typing import IO, Iterator
 
 import requests
 
@@ -148,8 +147,6 @@ class MakeResultTb:
 class CopyResultTb:
     """CopyResultTb - Use the server's HTTP PUT method to upload a tarball
     """
-
-    CHUNK_SIZE = 65536
 
     def __init__(
         self, controller: str, tarball: str, config: PbenchAgentConfig, logger: Logger
--- a/lib/pbench/agent/results.py
+++ b/lib/pbench/agent/results.py
@@ -169,13 +169,6 @@ class CopyResultTb:
         self.upload_url = f"{server_rest_url}/upload/{tbname}"
         self.logger = logger
 
-    def read_in_chunks(self, file_object: IO) -> Iterator[bytes]:
-        while True:
-            data = file_object.read(self.CHUNK_SIZE)
-            if not data:
-                break
-            yield data
-
     def copy_result_tb(self, token: str) -> None:
         """copy_result_tb - copies tb from agent to configured server upload URL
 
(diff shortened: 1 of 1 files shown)


portante (approved):

portante (changes requested):
Well, looks good, but `flake8` does not think so.

dbutenhof (approved):
Cool; and with no changes on the server side.

But, does that mean we didn't try removing the explicit `Content-Length` _and_ passing `f` instead of the generator at the same time yesterday? Huh.

I like the asserts... although if this ever happens, it won't be obvious what's causing it, and logger messages and an Exception might really be a cleaner way to report it.

dbutenhof:
```
./lib/pbench/agent/results.py:9:1: F401 'typing.IO' imported but unused
./lib/pbench/agent/results.py:9:1: F401 'typing.Iterator' imported but unused
flake8.main.application   MainProcess   2523 INFO     Found a total of 186 violations and reported 2
FAILED: flake8 --verbose .
```

npalaska (dismissed):

lib/pbench/agent/results.py
```diff
@@ -169,13 +169,6 @@ def __init__(
         self.upload_url = f"{server_rest_url}/upload/{tbname}"
         self.logger = logger
 
-    def read_in_chunks(self, file_object: IO) -> Iterator[bytes]:
-        while True:
-            data = file_object.read(self.CHUNK_SIZE)
```
portante, line 174: It seems we can also remove the `CHUNK_SIZE` now.

portante (changes requested):

webbnh:
Update:

- Fixed the Flake8 issue.  (Last night I hung out until the lights turned green -- which I swear they did -- so there must be some post-green step where it actually looks at the results.  :-s )
- Removed the now-unused `CHUNK_SIZE`.
- Rebased.

> But, does that mean we didn't try removing the explicit Content-Length and passing f instead of the generator at the same time yesterday? Huh.

Well, yes.  What we tried to do was to set the request headers, but I think `requests` was overriding our choices when we used the simplified interface.  What I've done here is to use the more "advanced" interface, which gives us access to the fully-prepared request just before it is sent, which we didn't have before.

> I like the asserts... although if this ever happens, it won't be obvious what's causing it, and logger messages and an Exception might really be a cleaner way to report it.

@dbutenhof, I found your points persuasive and most compelling, until I started thinking about what the resulting code would look like.  In essence, these are "logic bombs" which should never "go off", and so spending any more code (screen real estate or execution time) on them than necessary should be silly.  The `assert` statements _will_ raise exceptions, so there's that (although, as the code currently stands, the resulting messages will be useless, so I'll do another commit to address that); I'm missing the logger message...but, as I said, I don't think it's worth the code to try to make that happen, especially as this is on the Agent side.

dbutenhof (dismissed):

webbnh:
OK, I've added messages to the assertions (I'd misplaced the fact that assertions with no messages produce no messages -- they don't even report the assertion text itself, they just raise a bare `AssertionError`...).

However, Black and I are not on speaking terms at the moment.  (Fortunately, I suppose, there's no arguing with it.)

lib/pbench/agent/results.py
```diff
@@ -194,9 +194,18 @@ def copy_result_tb(self, token: str) -> None:
...
+                    "Transfer-Encoding" not in request.headers
+                ), "Upload request unexpectedly contains a `Transfer-Encoding` header"
+                assert (
+                    "Content-Length" in request.headers
+                ), "Upload request unexpectedly missing a `Content-Length` header"
+                assert request.headers["Content-Length"] == str(content_length), (
+                    "Upload request `Content-Length` header contains {} -- "
+                    "expected {}".format(
+                        request.headers["Content-Length"], content_length
+                    )
```
dbutenhof, line 207: I'm unable to decide whether an f-string would be easier or harder to read here. The "--" at the end of the first line looks a bit odd, and "but" (or even just ",") would probably be easier to read ... although I suspect that if we ever see the output there's not going to be any doubt what it means.

dbutenhof (approved):

lib/pbench/agent/results.py
webbnh: > I'm unable to decide whether an f-string would be easier or harder to read here.

See my previous comment about Black.  (I was **amazed** at its choices for breaking and wrapping this statement!)

> The "--" at the end of the first line looks a bit odd

It only looks that way in the source code -- as you say, if the message ever appears on the screen it will look "right".

webbnh: I'll replace the `--` iff I need to make some other change.

npalaska (approved):

lib/pbench/agent/results.py
portante: If it helps, I'd be fine leaving the dash-dash there.

portante (approved):

ndokos (commented):
Better late than never...

PR #2315, opened 2021-05-17
Clean up LGTM complaints

webbnh:
Apparently LGTM only reports "new" (or newly-fixed) problems, and we now have a backlog of 40 issues.  This PR addresses many of the ones which refer to Python code.  (I did not take on the `.js` ones.)

In some cases, the changes correct the issues that LGTM raised (useless return values, unused variables, imports via both `import` and `from`, an unguarded `next()` call in a generator, a potentially confusing variable shadowing, an `else` on a `for` loop containing no `break`, unused imports).

In other cases, the changes add directives to suppress false positives.  (E.g., [unreachable code](https://github.com/github/codeql/issues/5910).)

And, I made a couple of tiny code improvements, which I couldn't resist....

Note that there are a number of cases which I did not address.  These are false-positives referring to [duplicate-import](https://github.com/github/codeql/issues/5912) and [self-import](https://github.com/github/codeql/issues/5913) cases.  I can add the suppressions to these if we like, but for whatever reason it didn't strike me as worthwhile.

(@portante, I didn't add any labels, projects, or milestones to this, as I don't know when you would prefer to see it merged, if indeed we want these changes at all -- I was just shocked to find the backlog (at least one of which is significant), so I figured I would knock a bunch of them down and put this PR up.)

### DIFF
--- a/agent/tool-scripts/datalog/haproxy-ocp-datalog
+++ b/agent/tool-scripts/datalog/haproxy-ocp-datalog
@@ -267,8 +267,6 @@ def operation(routers, ts_str):
     new_routers = curr_routers - prev_routers
     # These are all the routers that have died since the last time.
     dead_routers = prev_routers - curr_routers
-    # Remember for the next iteration of the loop.
-    prev_routers = curr_routers
 
     for pod_name in dead_routers:
         # Clear out any dead routers
@@ -399,7 +397,7 @@ while not terminate:
     next_start = time_stamp + interval
 
 # First wait on any existing sub-processes to complete.
-for router in _routers.values():
+for router in _routers.values():  # lgtm [py/unreachable-statement]
     router.wait(all=True)
 
 ts_str = datetime.utcfromtimestamp(time()).strftime("%s.%f")
--- a/lib/pbench/server/indexer.py
+++ b/lib/pbench/server/indexer.py
@@ -13,7 +13,7 @@ import re
 import socket
 import tarfile
 import errno
-from collections import Counter
+from collections import Counter, OrderedDict
 from datetime import datetime, timedelta
 from operator import itemgetter
 from random import SystemRandom
@@ -644,6 +644,13 @@ class ResultData(PbenchData):
                                 self.ptb._tbctx,
                             )
                             continue
+                except StopIteration:
+                    # The csv_reader encountered the end-of-file on the first
+                    # iteration (i.e., the file was empty), so skip it.
+                    self.ptb.idxctx.logger.warning(
+                        "CSV file {csv_name!r} is empty ({self.ptb._tbctx})"
+                    )
+                    continue
                 except OSError as exc:
                     if exc.errno == errno.ENOENT:
                         continue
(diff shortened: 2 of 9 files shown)


agent/tool-scripts/datalog/haproxy-ocp-datalog
```diff
@@ -399,7 +397,7 @@ while not terminate:
     next_start = time_stamp + interval
 
 # First wait on any existing sub-processes to complete.
-for router in _routers.values():
+for router in _routers.values():  # lgtm [py/unreachable-statement]
```
dbutenhof, line 400: So is this because LGTM can't tell that the previous loop terminates based on a global presumably set somewhere else? If so, seems like this deserves a bit of a comment to clarify, because out of the blue it looks weird. (And apparently there are a series of related changes; which one might take as emphasis on the generic philosophy that "globals are bad", I suppose... although the failure of analysis tools to deal with a language feature gracefully isn't actually quite the same as an argument that the feature is bad ...)

lib/pbench/server/indexer.py
```diff
@@ -1193,7 +1200,7 @@ def expand_template(templ, d, run=None):
                         pass
                     else:
                         s = re.sub(m, val, s)
-                if run is not None and key == "controller_host":
+                elif key == "controller_host" and run is not None:
```
dbutenhof, line 1203: OK, I like the `elif` here; but was this really an LGTM complaint? And if so, on what basis?

dbutenhof (approved):

agent/tool-scripts/datalog/haproxy-ocp-datalog
webbnh: I'm willing to cut LGTM some slack:  figuring out that the previous loop terminates due to an external, asynchronously generated signal is a bit of a tall order.  (But, I did report it to the LGTM folks, and they did agree that it was a false positive...but they aren't sure when they will be able to fix it -- they are working on security stuff, currently, so this issue went into the backlog.)

I wrestled with adding an additional comment but decided that that would only make the wart bigger.  I concluded that the presence of the `lgtm` token was enough of a clue, and, if someone really wanted to know what was going on, they could remove the directive and apply LGTM to the result.

I concur that, in general, globals are bad (and, I'll point out that the `global` statement in Python doesn't quite mean what it seems to say), but I'm not sure it would really be a win to wrap the `terminate` flag in a class with setters and getters so that it's obvious how it's being accessed...in particular, because LGTM would presumably _still_ have problems knowing when/how the setter is being called....

lib/pbench/server/indexer.py
webbnh: Um, nope -- that was one of the tiny improvements that I couldn't resist.

portante (approved):

npalaska (approved):

webbnh:
Rebased.

dbutenhof (approved):

portante (approved):

PR #2317, opened 2021-05-17
Handle `ENOSPC` and report exceptions once

portante:
We add support for returning `INSUFFICIENT_STORAGE` (`507`) when writing a tar ball to disk and the server runs out of space (the specific exception is caught and handled only while writing the tar ball contents to disk).

While we are at it, we replace the use of `logger.exception()` with `logger.error()`, since the `flask.abort()` method also emits a stack trace.  This avoids have 2 stack traces emitted in the logs for the same error.

### DIFF
--- a/lib/pbench/server/api/resources/upload_api.py
+++ b/lib/pbench/server/api/resources/upload_api.py
@@ -1,3 +1,4 @@
+import errno
 import hashlib
 import os
 import tempfile
@@ -28,9 +29,9 @@ class HostInfo(Resource):
             response = jsonify(
                 dict(message=f"{self.user}@{self.host}" f":{self.prdp}-002")
             )
-        except Exception:
-            self.logger.exception(
-                "There was something wrong constructing the host info"
+        except Exception as exc:
+            self.logger.error(
+                "There was something wrong constructing the host info: '{}'", exc
             )
             abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
         response.status_code = HTTPStatus.OK
@@ -54,8 +55,8 @@ class Upload(Resource):
     def put(self, filename: str):
         try:
             username = Auth.token_auth.current_user().username
-        except Exception:
-            self.logger.exception("Error verifying the username")
+        except Exception as exc:
+            self.logger.error("Error verifying the username: '{}'", exc)
             abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
         if os.path.basename(filename) != filename:
@@ -153,9 +154,10 @@ class Upload(Resource):
             response = jsonify(dict(message="Dataset already exists"))
             response.status_code = HTTPStatus.OK
             return response
-        except Exception:
-            self.logger.exception(
-                "unable to create dataset for user = {}, ctrl = {!a}, file = {!a}",
+        except Exception as exc:
+            self.logger.error(
+                "unable to create dataset, '{}', for user = {}, ctrl = {!a}, file = {!a}",
+                exc,
                 username,
                 controller,
                 filename,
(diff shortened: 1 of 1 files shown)


dbutenhof (dismissed):
Very nice!

dbutenhof (approved):

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -196,11 +198,34 @@ def put(self, filename: str):
...
                 msg = "Unexpected error encountered during file upload"
-                self.logger.exception(
-                    "{} for user = {}, ctrl = {!a}, file = {!a}",
+                self.logger.error(
+                    "{}, {}, for user = {}, ctrl = {!a}, file = {!a}",
                     msg,
+                    exc,
                     username,
                     controller,
                     filename,
```
webbnh, line 231: Having to repeat lines 224-233 at 213-222 is disappointing.  We should consider implementing a utility function like the following:
```
def errno_exception(errno_val: int):
    exc = sys.exc_info()[0]
    if exc is OSError and sys.exc_info()[1].errno == errno_val:
        return exc
    # Return any exception class which doesn't match.                                                                                                                                          
    return ValueError if exc is not ValueError else TypeError
```
We could then use this at line 201 to catch only the specific `OSError` that we're interested in:
```
            except errno_exception(errno.ENOSPC) as exc:
```
And any other `OSError`'s would fall through to the catch-call case.

This would presumably be useful in other places where we catch `OSError`....

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -237,39 +262,44 @@ def put(self, filename: str):
...
-                self.logger.exception(
-                    "Failed to rename tar ball '{}' to '{}'", ofp.name, md5_full_path,
+                self.logger.error(
+                    "Failed to rename tar ball '{}' to '{}': '{}'",
+                    ofp.name,
+                    md5_full_path,
+                    exc,
                 )
-                raise
+                abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
```
webbnh, line 297: What is the default behavior for an unhandled exception?  Is it `500/INTERNAL ERROR`?  If so, maybe the re-raise wasn't such a bad idea....  That is, you indicated that we already get a log message from the `abort()`...so, perhaps there is no need for the inner `try` block at _all_ and the outer `except` block doesn't need to do anything other than invoke `md5_full_path.unlink()` and `raise` (or call `abort()`)?

webbnh (approved):

lib/pbench/server/api/resources/upload_api.py
portante: We need the inner `try` block to prevent any exception from the `unlink()` from being raised at all.  We just want to "note" that something happened while handling the first exception, and then let the first exception propagate.

lib/pbench/server/api/resources/upload_api.py
portante: Tricky.

I'll have to noodle on that a bit.

Tracked in #2322.

lib/pbench/server/api/resources/upload_api.py
webbnh: I was thinking more generally:  it seems like we have a lot of places where we catch exceptions just so that we can log them and then call `abort()` with a `500` error...if it turned out that Flask's default reaction to an unhandled exception was to log it and abort with a `500` error, then we could get rid of a lot of code....

portante: Potentially.  There are ways to add generic error handling (see https://python.plainenglish.io/python-web-development-with-flask-error-handling-36349a7f82c).

This specific problem is that `abort()` does not give us a way to add color to what is logged.

But in another commit, we could re-work this code that it only raises an exception class we know about where, we provide a response code, message (optional), and logging context.

portante: Tracked in #2324.

PR #2318, opened 2021-05-17
Culmination of work from Tufts COMP 150 Jaeger tracing tool integration

portante:
A big thanks to @nikiglen, @xiaoxiong-huang, and @michaelgoldpiano for their course work to add Jaeger tracing to the pbench agent.

There are 5 commits.  The first two deal with environment changes needed to work outside of Fedora based environments.  The 3rd commit is the actually Jaeger code changes.  The 4th commit is the demo of a simple Jaeger-instrumented program.  The final commit is a generic demo script for running the pbench infrastructure using all containers.


portante:
The original code is in the `comp-150` branch.

PR #2319, opened 2021-05-18
Fix pbench-register-tool-trigger

### LINKED ISSUE #2268: `pbench-register-tool-trigger` reports no triggers found without any options
pbench-agent-0.69.7:

```
$ pbench-register-tool-trigger
No __triggers__ in /var/lib/pbench-agent/tools-default
```

ndokos:
Fixes #2268

Fix pbench-register-tool-trigger to detect missing arguments: at least
one of --start-trigger or --stop-trigger is always required.

Add unit tests for the various cases.

### DIFF
--- a/agent/util-scripts/gold/pbench-register-tool-trigger/test-35.08.txt
+++ b/agent/util-scripts/gold/pbench-register-tool-trigger/test-35.08.txt
@@ -1,0 +1,9 @@
++++ Running test-35.08 pbench-register-tool-trigger --group=default --start-trigger="START DEFAULT" --stop-trigger=
+Both --start-trigger and --stop-trigger must be specified and must be non-empty.
+usage:
+pbench-register-tool-trigger [--group=<tool group>] --start-trigger=<string to match> --stop-trigger=<string to match>
+--- Finished test-35.08 pbench-register-tool-trigger (status=1)
++++ pbench tree state
+/var/tmp/pbench-test-utils/pbench
+/var/tmp/pbench-test-utils/pbench/tmp
+--- pbench tree state
--- a/agent/util-scripts/pbench-register-tool-trigger
+++ b/agent/util-scripts/pbench-register-tool-trigger
@@ -1,4 +1,5 @@
 #!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: t; sh-basic-offset: 8; sh-indentation: 8; sh-indent-for-case-alt: + -*-
 
 script_path=`dirname $0`
 script_name=`basename $0`
@@ -19,16 +20,19 @@ stop_trigger=""
 
 function usage() {
 	printf "usage:\n"
-	printf "$script_name --group=<tool group> --start-trigger=<string to match> --stop-trigger=<string to match>"
+	printf "$script_name [--group=<tool group>] --start-trigger=<string to match> --stop-trigger=<string to match>\n"
 }
+
+if [[ "${#}" == 0 ]] ;then
+        usage >&2
+        exit 1
+fi
 
 # Process options and arguments
 opts=$(getopt -q -o h --longoptions "help,group:,start-trigger:,stop-trigger:" -n "getopt.sh" -- "$@");
 if [ $? -ne 0 ]; then
-	printf "\n"
-	printf "%s\n" $*
-	printf "$script_name: you specified an invalid option\n\n"
-	usage
+	printf "\n%s\n%s: you specified an invalid option\n\n" "${*}" "${script_name}" >&2
+	usage >&2
 	exit 1
 fi
 eval set -- "$opts";
@@ -69,12 +73,17 @@ while true; do
 	esac
 done
 
-if [ "$start_trigger" != "${start_trigger%%:*}" ]; then
-	printf "$script_name: the start trigger cannot have a colon in it: \"${start_trigger}\"\n"
+if [[ -z "$start_trigger" || -z "$stop_trigger" ]] ;then
+	printf "Both --start-trigger and --stop-trigger must be specified and must be non-empty.\n" >&2
+	usage >&2
+	exit 1
+fi
+if [[ "$start_trigger" != "${start_trigger%%:*}" ]]; then
+	printf "$script_name: the start trigger cannot have a colon in it: \"${start_trigger}\"\n" >&2
 	exit 1
 fi
 if [ "$stop_trigger" != "${stop_trigger%%:*}" ]; then
-	printf "$script_name: the stop trigger cannot have a colon in it: \"${stop_trigger}\"\n"
+	printf "$script_name: the stop trigger cannot have a colon in it: \"${stop_trigger}\"\n" >&2
 	exit 1
 fi
 
--- a/agent/util-scripts/test-bin/test-pbench-register-tool-trigger
+++ b/agent/util-scripts/test-bin/test-pbench-register-tool-trigger
@@ -1,0 +1,23 @@
+#! /bin/bash
+
+case $1 in
+    one)
+        pbench-register-tool-trigger --group=default --start-trigger="START DEFAULT" --stop-trigger="STOP DEFAULT"
+        sts=${?}
+        ;;
+    two)
+        pbench-register-tool-trigger --group=other --start-trigger="START OTHER" --stop-trigger="STOP OTHER"
+        sts=${?}
+        ;;
+    three)
+        pbench-register-tool-trigger --start-trigger="START DEFAULT" --stop-trigger="STOP DEFAULT"
+        sts=${?}
+        ;;
+    *)
+        exit 123
+esac
+
+echo "+++ tool-triggers file contents"
+cat ${_testdir}/tool-triggers
+echo "--- tool-triggers file contents"
+exit ${sts}
--- a/agent/util-scripts/unittests
+++ b/agent/util-scripts/unittests
@@ -324,7 +324,6 @@ declare -A tools=(
     [test-12.05]="pbench-list-tools"
     [test-13]="pbench-postprocess-tools"
     [test-14]="pbench-agent-config-activate"
-    [test-15]="pbench-register-tool-trigger"
     [test-16]="pbench-list-triggers"
     [test-17]="pbench-list-triggers"
     [test-18]="pbench-list-triggers"
@@ -345,8 +344,19 @@ declare -A tools=(
     [test-32]="pbench-copy-results"
     [test-33]="pbench-move-results"
     [test-34]="pbench-move-results"
-    [test-35]="pbench-register-tool-trigger"
-    [test-36]="pbench-register-tool-trigger"
+    [test-35.00]="pbench-register-tool-trigger"
+    [test-35.01]="pbench-register-tool-trigger"
+    [test-35.02]="pbench-register-tool-trigger"
+    [test-35.03]="pbench-register-tool-trigger"
+    [test-35.04]="pbench-register-tool-trigger"
+    [test-35.05]="pbench-register-tool-trigger"
+    [test-35.06]="pbench-register-tool-trigger"
+    [test-35.07]="pbench-register-tool-trigger"
+    [test-35.08]="pbench-register-tool-trigger"
+    [test-35.09]="pbench-register-tool-trigger"
+    [test-35.10]="test-pbench-register-tool-trigger"
+    [test-35.11]="test-pbench-register-tool-trigger"
+    [test-35.12]="test-pbench-register-tool-trigger"
     [test-37]="pbench-move-results"
     [test-38]="pbench-move-results"
     [test-39]="pbench-copy-results"
@@ -408,7 +418,6 @@ declare -A options=(
     [test-12.05]="--group default --name iostat"
     [test-13]="--dir=${_testdir}/mock-run/mock-iteration/mock-sample --group=default"
     [test-14]="${_testdir}/tmp/pbench-agent.cfg"
-    [test-15]="--group=default --start-trigger=\"START DEFAULT\" --stop-trigger=\"STOP DEFAULT\""
     [test-17]="--group=default"
     [test-18]="--group=foo"
     [test-20]="--dir=${_testdir}/mock-run/mock-iteration/mock-sample --group=default"
@@ -430,8 +439,19 @@ declare -A options=(
     #[test-33]=""
     # pbench-move-results
     [test-34]="--help"
-    [test-35]="--group=default --start-trigger=\"START:DEFAULT\" --stop-trigger=\"STOP DEFAULT\""
-    [test-36]="--group=default --start-trigger=\"START DEFAULT\" --stop-trigger=\"STOP:DEFAULT\""
+    [test-35.00]=""
+    [test-35.01]="--help"
+    [test-35.02]="--wrong"
+    [test-35.03]="--group=foobar"
+    [test-35.04]="--group=foobar --start-trigger=foo"
+    [test-35.05]="--group=foobar --stop-trigger=bar"
+    [test-35.06]="--group=default --start-trigger=\"START:DEFAULT\" --stop-trigger=\"STOP DEFAULT\""
+    [test-35.07]="--group=default --start-trigger=\"START DEFAULT\" --stop-trigger=\"STOP:DEFAULT\""
+    [test-35.08]="--group=default --start-trigger=\"START DEFAULT\" --stop-trigger="
+    [test-35.09]="--group=default --start-trigger= --stop-trigger=\"STOP DEFAULT\""
+    [test-35.10]="one"
+    [test-35.11]="two"
+    [test-35.12]="three"
     # pbench-move-results
     [test-37]="--prefix="
     # pbench-move-results
@@ -468,8 +488,15 @@ declare -A expected_status=(
     [test-13]=2
     [test-21.00]=1
     [test-29]=1
-    [test-35]=1
-    [test-36]=1
+    [test-35.00]=1
+    [test-35.02]=1
+    [test-35.03]=1
+    [test-35.04]=1
+    [test-35.05]=1
+    [test-35.06]=1
+    [test-35.07]=1
+    [test-35.08]=1
+    [test-35.09]=1
     [test-37]=1
     [test-38]=1
     [test-44]=1
--- a/agent/util-scripts/gold/pbench-register-tool-trigger/test-15.txt
+++ b/agent/util-scripts/gold/pbench-register-tool-trigger/test-15.txt
@@ -1,8 +1,0 @@
-+++ Running test-15 pbench-register-tool-trigger --group=default --start-trigger="START DEFAULT" --stop-trigger="STOP DEFAULT"
-tool trigger strings for start: ""START" and for stop: ""STOP" are now registered for tool group: default
---- Finished test-15 pbench-register-tool-trigger (status=0)
-+++ pbench tree state
-/var/tmp/pbench-test-utils/pbench
-/var/tmp/pbench-test-utils/pbench/tmp
-/var/tmp/pbench-test-utils/pbench/tool-triggers
---- pbench tree state
--- a/agent/util-scripts/gold/pbench-register-tool-trigger/test-35.00.txt
+++ b/agent/util-scripts/gold/pbench-register-tool-trigger/test-35.00.txt
@@ -1,0 +1,8 @@
++++ Running test-35.00 pbench-register-tool-trigger 
+usage:
+pbench-register-tool-trigger [--group=<tool group>] --start-trigger=<string to match> --stop-trigger=<string to match>
+--- Finished test-35.00 pbench-register-tool-trigger (status=1)
++++ pbench tree state
+/var/tmp/pbench-test-utils/pbench
+/var/tmp/pbench-test-utils/pbench/tmp
+--- pbench tree state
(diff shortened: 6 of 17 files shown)


agent/util-scripts/pbench-register-tool-trigger
```diff
@@ -19,17 +20,24 @@ stop_trigger=""
 
 function usage() {
 	printf "usage:\n"
-	printf "$script_name --group=<tool group> --start-trigger=<string to match> --stop-trigger=<string to match>"
+	printf "$script_name [--group=<tool group>] --start-trigger=<string to match> --stop-trigger=<string to match>\n"
 }
 
+case $# in
+    0)
+        usage
```
portante, line 28: Perhaps `usage >&2`?

agent/util-scripts/pbench-register-tool-trigger
```diff
@@ -19,17 +20,24 @@ stop_trigger=""
...
+        ;;
+esac
+
 # Process options and arguments
 opts=$(getopt -q -o h --longoptions "help,group:,start-trigger:,stop-trigger:" -n "getopt.sh" -- "$@");
 if [ $? -ne 0 ]; then
 	printf "\n"
 	printf "%s\n" $*
 	printf "$script_name: you specified an invalid option\n\n"
 	usage
```
portante, line 39: Perhaps `usage >&2`?

agent/util-scripts/pbench-register-tool-trigger
```diff
@@ -19,17 +20,24 @@ stop_trigger=""
...
-	exit 1
+	exit 2
```
portante, line 40: It is not clear to me that we want to start incrementing the exit code on the `b0.69` branch.

agent/util-scripts/pbench-register-tool-trigger
```diff
@@ -69,13 +77,18 @@ while true; do
 	esac
 done
 
-if [ "$start_trigger" != "${start_trigger%%:*}" ]; then
+if [[ -z "$start_trigger" && -z "$stop_trigger" ]] ;then
+	printf "At least one of --start-trigger or --stop-trigger must be specified.\n"
```
portante, line 81: For now we might want to require both.  This will help catch problems, and any use that does not use one or the other can set it to a value that is not generated in the script output.

portante (changes requested):
Should this be first applied to `main` before we apply it to `b0.69`?

agent/util-scripts/pbench-register-tool-trigger
webbnh: Beyond that, do we really want a different code for each individual error?  It's more conventional to use specific codes for error _types_...so all of the bad and missing option errors would be `2`'s, while operational failures (such as failure to create files or directories) would be `1`'s.

agent/util-scripts/pbench-register-tool-trigger
webbnh: What does it mean to have _no_ start trigger?  Wouldn't that mean that the tool never starts??

And, what does it mean to have no stop trigger...does the collection get stopped at the end of the iteration, or does it run longer?  (Can we end up with multiple copies of the same tool running concurrently in subsequent iterations??)

webbnh: @portante, should there be code in the Tool Meister to detect and report tools whose starts are never triggered?

agent/util-scripts/pbench-register-tool-trigger
```diff
@@ -69,13 +77,18 @@ while true; do
...
+if [[ "$start_trigger" != "${start_trigger%%:*}" ]]; then
 	printf "$script_name: the start trigger cannot have a colon in it: \"${start_trigger}\"\n"
-	exit 1
+	exit 4
 fi
 if [ "$stop_trigger" != "${stop_trigger%%:*}" ]; then
 	printf "$script_name: the stop trigger cannot have a colon in it: \"${stop_trigger}\"\n"
-	exit 1
+	exit 5
 fi
```
webbnh, line 92: Is it a problem if the start and stop triggers are the same text?  (I think not...but, if I'm wrong, then we should check for that.)

agent/util-scripts/pbench-register-tool-trigger
```diff
@@ -69,13 +77,18 @@ while true; do
...
 done
 
-if [ "$start_trigger" != "${start_trigger%%:*}" ]; then
+if [[ -z "$start_trigger" && -z "$stop_trigger" ]] ;then
+	printf "At least one of --start-trigger or --stop-trigger must be specified.\n"
+	usage
+	exit 3
+fi
+if [[ "$start_trigger" != "${start_trigger%%:*}" ]]; then
 	printf "$script_name: the start trigger cannot have a colon in it: \"${start_trigger}\"\n"
```
webbnh, line 86: We should also probably ensure that the trigger does not contain a newline, as that will screw up the `tool-triggers` file.

webbnh (dismissed):
I'm not a fan of the multiple error code values, but if that's what we want, that's fine; otherwise, these changes seem fine.  However, I do have a few questions/comments/suggestions which we might want to address before merging (or separately).  (Some of the questions are probably aimed more at Peter....)

agent/util-scripts/pbench-register-tool-trigger
portante: Shouldn't be.  That is really at the caller's discretion for the behavior they want.

agent/util-scripts/pbench-register-tool-trigger
portante: Agreed, though for the `0.69` branch I am not included to tackle this work, but instead fix this in a rewrite in Python3.

agent/util-scripts/pbench-register-tool-trigger
portante: > What does it mean to have _no_ start trigger? Wouldn't that mean that the tool never starts??

It means that what ever entity ultimately call `pbench-tool-meister-start` would have to be calling `pbench-start-tools` directly, but stopping the tools when triggered.

> And, what does it mean to have no stop trigger...does the collection get stopped at the end of the iteration, or does it run longer? (Can we end up with multiple copies of the same tool running concurrently in subsequent iterations??)

The Tool Meister code won't start more tools until the previous set are stopped.

So if no stop trigger is provided, that just means the driving entity is smart enough to stop the tools themselves.

agent/util-scripts/pbench-register-tool-trigger
webbnh: I meant that the Tool Meister code will do the right thing if the two are the same -- i.e., if it encounters the trigger, it will either perform a start (if the tool is not running) or a stop (if the tool is already running), and it won't do something silly like start it a second time or fail to stop it if appropriate....

agent/util-scripts/pbench-register-tool-trigger
webbnh: Ah.  Tricky.

In that case, I don't see a good justification for requiring _both_...which leaves the awkward question of why we require _either one_.  (I suppose that the use case where the user would want _neither_ is so far out that it's worth his/her inconvenience in order to protect all the other users...which, I suppose, is why you're suggesting requiring both...eh?  But, the trick of supplying dummy text in the rare cases to work around the guards is kind of a hack...but, perhaps those cases are so rare that it's justified.)

portante: > which, I suppose, is why you're suggesting requiring both...eh?

Yes, thinking about this interface, which is already really difficult to get right, and is only really used by `pbench-specjbb` and `pbench-trafficgen` today, by requiring both we'll flag the really odd cases where they are not both specified, to help make sure the user really knows what they are doing.

That is to say, if the user is savvy enough to only provide start or stop triggers, and knows how to call start or stop tools directly, then they are savvy enough to provide dummy strings for the option that is ignored.  Savvy?  ;)

agent/util-scripts/pbench-register-tool-trigger
ndokos: Fixed (here and elsewhere - I hope I got them all).

agent/util-scripts/pbench-register-tool-trigger
ndokos: Fixed.

agent/util-scripts/pbench-register-tool-trigger
ndokos: Reverted to a single error code (1).

agent/util-scripts/pbench-register-tool-trigger
```diff
@@ -19,16 +20,23 @@ stop_trigger=""
...
+
 # Process options and arguments
 opts=$(getopt -q -o h --longoptions "help,group:,start-trigger:,stop-trigger:" -n "getopt.sh" -- "$@");
 if [ $? -ne 0 ]; then
-	printf "\n"
-	printf "%s\n" $*
-	printf "$script_name: you specified an invalid option\n\n"
-	usage
+	printf "\n" >&2
+	printf "%s\n" $* >&2
```
portante, line 37: This is really an existing bug, but since you are making changes here ... the `$*` needs to be quoted to make sure it is expanded to one argument.

Here is a demonstration of the bad behavior:

```
$ cat /tmp/testit
#!/bin/bash

printf "%s\n" $*
printf "%s\n" "${*}"
$ /tmp/testit a b c e
a
b
c
e
a b c e
$
```

Perhaps we can collapse this to one `printf`:

```
printf "\n%s\n%s: you specified an invalid option\n\n" "${*}" "${script_name}" >&2
```

agent/util-scripts/pbench-register-tool-trigger
ndokos: Enlightening (or, to say it differently: savvy :-) ). Both are now required - more unit tests added.

portante (changes requested):

agent/util-scripts/pbench-register-tool-trigger
```diff
@@ -19,17 +20,24 @@ stop_trigger=""
 
 function usage() {
 	printf "usage:\n"
-	printf "$script_name --group=<tool group> --start-trigger=<string to match> --stop-trigger=<string to match>"
+	printf "$script_name [--group=<tool group>] --start-trigger=<string to match> --stop-trigger=<string to match>\n"
```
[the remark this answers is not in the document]
ndokos: ... and I also screwed up a couple of existing ones. In any case, fixed the old successful one (although it's renumbered 35.10) and added a new successful one (35.11).

agent/util-scripts/pbench-register-tool-trigger
```diff
@@ -19,17 +20,24 @@ stop_trigger=""
...
 }
 
+case $# in
+    0)
+        usage
+        exit 1
+        ;;
+esac
+
```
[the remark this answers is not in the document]
ndokos: The argument count turns out to be unreliable: spaces in quoted strings (e.g. `--start-trigger="foo bar"`) interfere with it. So I gave up on trying to test any kind of count other than 0 and  turned it into an `if`.

agent/util-scripts/pbench-register-tool-trigger
ndokos: I combined them into one `printf` with proper quoting (I hope): I should have done that in the first place instead of adding `>&2` umpteen times.

agent/util-scripts/pbench-register-tool-trigger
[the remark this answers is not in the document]
ndokos: Turns out I was confused: I did not screw up an existing success, I just missed it altogether: it used to be test-15, I added the same test as test-35.10 and I finally deleted the original test-15, so as to keep them all together as 35.xx. But now there is a gap at 15 :-)

webbnh (changes requested):
There is something wrong with the unit test invocation which isn't quoting arguments properly.  This doesn't have anything to do with this change per se, but it affects it nonetheless, and it should probably be fixed first, so that this PR doesn't have to submit incorrect gold files which match the incorrect behavior.  (Also, it would allow this change to [properly test the upper bound](https://github.com/distributed-system-analysis/pbench/pull/2319#discussion_r634630244) on the command's argument count.)

In the gold file for test 35.10, the tool is reporting the value of the trigger as `"START` -- notice the embedded quote and the fact that the second half of the string is missing.  The unit test invocation is apparently presenting the command line arguments with an escaped quotation mark, such that `--start-trigger="START DEFAULT"` is interpreted as two separate arguments (each containing quotation marks) broken at the blank character, instead of a single argument containing embedded quotation marks.  This results in the tool reporting the incorrect value for the triggers, and it defeated Nick's attempts to check the upper bound on the argument count (because, in what seemed like a success case, the tool was being presented with too many arguments!).

agent/util-scripts/pbench-register-tool-trigger
```diff
@@ -19,16 +20,21 @@ stop_trigger=""
...
+
 # Process options and arguments
 opts=$(getopt -q -o h --longoptions "help,group:,start-trigger:,stop-trigger:" -n "getopt.sh" -- "$@");
 if [ $? -ne 0 ]; then
-	printf "\n"
-	printf "%s\n" $*
-	printf "$script_name: you specified an invalid option\n\n"
-	usage
+	printf "\n%s\n%s: you specified an invalid option\n\n" "${*}" "${script_name}" >&2
```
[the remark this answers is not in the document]
portante: I am not too concerned about 80 columns in `bash` files, and would rather just re-write this in Python3 to make it 80 columns.

portante (dismissed):

webbnh:
Here's the issue ([`agent/util-scripts/unittests` line 112](https://github.com/distributed-system-analysis/pbench/blob/212cfff7bce8d8d320443c487caeea6ba7915a76/agent/util-scripts/unittests#L112)).

I _think_ the following (substituting `"${*}"` for `${@}`) will fix it:
```
    ${tscrpt} "${*}" &gt;&gt; ${_testout} 2&gt;&amp;1
```
...but, I really have no idea.

ndokos:
I don't think this works - consider a driver doing this:
```
#! /bin/bash

/tmp/testit.sh "${*}"
```

and the test script in /tmp/testit.sh doing this:

```
#! /bin/bash

echo "${#}"
echo "${1}"
echo "${2}"
echo "${3}"
echo "${4}"
echo "${5}"
exit 0
```

Doing `./driver.sh --b="a b" --c="c d"` results in
```
1
--b=a b --c=c d




```
a single argument is passed to the test script, no matter what. I think the most likely solution is `"${@}"` instead, although I glaze over when I read the legalese in the bash man page about it. But it seems to work OK with the above:

```$ ./driver.sh --b="a b" --c="c d"
2
--b=a b
--c=c d



```

ndokos:
Reading Kernighan and Pike, however, reassures me: http://files.catwell.info/misc/mirror/the-unix-programming-environment-kernighan-pike.pdf  seek to p.173 of the PDF (or if you have the printed book, p. 161). They are talking about Bourne shell of course, not bash, but there is hope...

I'll add the contents of the trigger file to the unit test output to make sure that I'm not fooling myself.

webbnh:
Excellent!  I tried all the combinations of `${*}` and `${@}`, with and without quotes, and I couldn't get any which worked right (I always got either `1` or `5` arguments...), so I hoped that `"${*}"` would glue them all together and the command invocation would properly divide them again.  Nevertheless, if `"${@}"` produces the correct argument count in _your_ experiments, hopefully it will work in practice...which will allow you to correct the gold files and, perhaps, to put an upper bound on the tool's argument count.

agent/util-scripts/gold/pbench-register-tool-trigger/test-35.10.txt
```diff
@@ -0,0 +1,8 @@
++++ Running test-35.10 pbench-register-tool-trigger --group=default --start-trigger="START DEFAULT" --stop-trigger="STOP DEFAULT"
+tool trigger strings for start: ""START" and for stop: ""STOP" are now registered for tool group: default
```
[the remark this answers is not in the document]
portante: I am no longer convinced this should be fixed.  @ndokos's changes here did not cause this behavior to start happen.  As such, I don't believe it is @ndokos's responsibility to correct this existing bad behavior.

If we really want to fix this on this old and out-dated support branch, we should file a separate PR to consider such a change.

We know this already fixed on the `main` branch because `pbench-register-tool-trigger` has been rewritten in Python3 and during that re-write this failure mode was removed.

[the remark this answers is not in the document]
webbnh: My concern is centered around the creation of gold files which are incorrect.  If the fix is as trivial as adding quotes to the unit test invocation, then it seems to me that it should go ahead.

However, I understand that there may be fallout among other gold files (some of which are already being modified by this PR).  If that impact is widespread, then I guess I can salve my conscience with the notion that this is (nearly) dead code.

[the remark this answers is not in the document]
webbnh: The problem in the unit test invocation code appears to still be present in the `main` branch.

I'm gonna guess that that branch isn't affected because there aren't any cases which have embedded quotes in the invocations (the unit testing for register-trigger is conducted by Python instead of directly via bash).

If fixing this in the 0.69 branch isn't too hard, it would be worthwhile as a proof-of-concept which we could then pull forward to `main` (where we cannot otherwise test it, currently, AFAICS).

agent/util-scripts/gold/pbench-register-tool-trigger/test-35.08.txt
```diff
@@ -0,0 +1,9 @@
++++ Running test-35.08 pbench-register-tool-trigger --group=default --start-trigger="START DEFAULT" --stop-trigger=
+Both --start-trigger and --stop-trigger must be specified.
```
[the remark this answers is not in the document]
portante: Given that we should not even consider this entire PR at all for the `b0.69`, can you file an issue against the `main` branch to ensure this bad behavior is covered there?

portante:
it is not clear this change should be made to the `b0.69` branch.  We only found this from a QE test, and nobody really uses the tool triggers heavily.  The tool triggers are used by `pbench-specjbb` and `pbench-trafficgen`, and some folks "might" use them with `pbench-user-benchmark`, but it is rare.

The changes this PR is bringing are an improvement over what we have today in `b0.69`, any further effort would probably be better spend on the `main` branch instead.

portante (approved):
Ship it!

agent/util-scripts/test-bin/test-pbench-register-tool-trigger
```diff
@@ -0,0 +1,21 @@
...
+        ;;
+    two)
+        pbench-register-tool-trigger --group=other --start-trigger="START OTHER" --stop-trigger="STOP OTHER"
+        sts=${?}
+        ;;
+    three)
+        pbench-register-tool-trigger --start-trigger="START DEFAULT" --stop-trigger="STOP DEFAULT"
+        sts=${?}
+        ;;
+esac
```
webbnh, line 16: No default (error-catching) case?

webbnh (approved):
OK, I suppose this horse is dead, now.  (Of course, I still found some nits which I was impelled to call out.)

(The other way to address the problem with the test command lines is to add additional levels of quoting or escaping to the original options string, but that's cumbersome, confusing, and ugly -- and it requires knowing how many layers of un-quoting are going to happen, which is unsatisfying from a maintenance perspective.)

agent/util-scripts/unittests
```diff
@@ -105,11 +105,11 @@ function _run {
...
         exit 1
     fi
-
+    local opts=${options[${tname}]}
     echo "${tname} ${tscrpt} ${@}"
-    echo "+++ Running ${tname} ${tscrpt} ${@}" >> ${_testout}
+    echo "+++ Running ${tname} ${tscrpt} ${opts}" >> ${_testout}
 
-    ${tscrpt} ${@} >> ${_testout} 2>&1
+    ${tscrpt} ${opts} >> ${_testout} 2>&1
```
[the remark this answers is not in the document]
ndokos: Ugh - I forgot to back out some experimental changes. The intent was to stay away from changing `unittests` - I'll have to do one more round.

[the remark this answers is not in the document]
ndokos: Latest commit changes the `unittests` file to back out those changes and adds a default case to the test script only.

agent/util-scripts/test-bin/test-pbench-register-tool-trigger
```diff
@@ -0,0 +1,23 @@
...
+    two)
+        pbench-register-tool-trigger --group=other --start-trigger="START OTHER" --stop-trigger="STOP OTHER"
+        sts=${?}
+        ;;
+    three)
+        pbench-register-tool-trigger --start-trigger="START DEFAULT" --stop-trigger="STOP DEFAULT"
+        sts=${?}
+        ;;
+    *)
+        exit 123
```
portante, line 17: Cute!

agent/util-scripts/test-bin/test-pbench-register-tool-trigger
```diff
@@ -0,0 +1,23 @@
...
+esac
+
+echo "+++ tool-triggers file contents"
+cat ${_testdir}/tool-triggers
+echo "--- tool-triggers file contents"
```
portante, line 22: Very nice addition.

portante (approved):

riya-17 (approved):
Looks Good!

agent/util-scripts/unittests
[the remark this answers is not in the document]
webbnh: Excellent -- I was hoping that that was the case...but I didn't want to continue beating the horse.  :-)

webbnh (approved):
Send it on its way!

PR #2326, opened 2021-05-19
Query fixes

dbutenhof:
This began as simply fixing the bad parameter type error, grew to cover a few other minor fixes, such as limiting the attempt to hide `pbench-user-create` failures in the unittests, and then to accommodate my attempt to utilize the `datasets/list` and `datasets/detail` APIs in the dashboard.

This also starts the move towards consolidating unit testing across the `ElasticBase` subclasses by moving them into a subdirectory with their own `conftest.py` fixtures.

### DIFF
--- a/lib/pbench/server/api/resources/query_apis/__init__.py
+++ b/lib/pbench/server/api/resources/query_apis/__init__.py
@@ -428,12 +428,13 @@ class ElasticBase(Resource):
         Returns:
             Postprocessed JSON body to return to client
         """
+        klasname = self.__class__.__name__
         try:
             es_request = self.assemble(json_data)
             path = es_request.get("path")
             url = urljoin(self.es_url, path)
         except Exception as e:
-            self.logger.exception("Blew it in setup: {}", type(e).__name__)
+            self.logger.exception("{} blew it in setup: {}", klasname, type(e).__name__)
             abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
         try:
@@ -441,20 +442,24 @@ class ElasticBase(Resource):
             es_response = method(url, **es_request["kwargs"])
             es_response.raise_for_status()
         except requests.exceptions.HTTPError as e:
-            self.logger.exception("HTTP error {} from Elasticsearch request", e)
+            self.logger.exception(
+                "{} HTTP error {} from Elasticsearch request", klasname, e
+            )
             abort(
                 HTTPStatus.BAD_GATEWAY,
-                message="Elasticsearch query failure {e.response.reason} ({e.response.status_code})",
+                message=f"Elasticsearch query failure {e.response.reason} ({e.response.status_code})",
             )
         except requests.exceptions.ConnectionError:
-            self.logger.exception("Connection refused during the Elasticsearch request")
+            self.logger.exception(
+                "{}: connection refused during the Elasticsearch request", klasname
+            )
             abort(
                 HTTPStatus.BAD_GATEWAY,
                 message="Network problem, could not reach Elasticsearch",
             )
         except requests.exceptions.Timeout:
             self.logger.exception(
-                "Connection timed out during the Elasticsearch request"
+                "{}: connection timed out during the Elasticsearch request", klasname
             )
             abort(
                 HTTPStatus.GATEWAY_TIMEOUT,
@@ -462,12 +467,13 @@ class ElasticBase(Resource):
             )
         except requests.exceptions.InvalidURL:
             self.logger.exception(
-                "Invalid url {} during the Elasticsearch request", url
+                "{}: invalid url {} during the Elasticsearch request", klasname, url
             )
             abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
         except Exception as e:
             self.logger.exception(
-                "Exception {} occurred during the Elasticsearch request",
+                "{}: exception {} occurred during the Elasticsearch request",
+                klasname,
                 type(e).__name__,
             )
             abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
@@ -475,18 +481,19 @@ class ElasticBase(Resource):
         try:
             return self.postprocess(es_response.json())
         except PostprocessError as e:
-            msg = f"The query postprocessor was unable to complete: {e}"
+            msg = f"{klasname}: the query postprocessor was unable to complete: {e}"
             self.logger.warning(msg)
             abort(HTTPStatus.BAD_REQUEST, message=msg)
         except KeyError as e:
-            self.logger.error("Missing Elasticsearch key {}", e)
+            self.logger.error("{}: missing Elasticsearch key {}", klasname, e)
             abort(
                 HTTPStatus.INTERNAL_SERVER_ERROR,
-                message="Missing Elasticsearch key {e}",
+                message=f"Missing Elasticsearch key {e}",
             )
         except Exception as e:
             self.logger.exception(
-                "Unexpected problem postprocessing Elasticsearch response {}: {}",
+                "{}: unexpected problem postprocessing Elasticsearch response {}: {}",
+                klasname,
                 es_response.json(),
                 e,
             )
@@ -510,10 +517,9 @@ class ElasticBase(Resource):
         try:
             new_data = self.schema.validate(json_data)
         except SchemaError as e:
-            # The format string here is important as the value we're
-            # trying to convert might contain "{}" brackets that would
-            # be interpreted as formatting commands.
-            self.logger.warning("{}", str(e))
+            self.logger.warning(
+                "{}: {} on {!r}", self.__class__.__name__, str(e), json_data
+            )
             abort(HTTPStatus.BAD_REQUEST, message=str(e))
         return self._call(requests.post, new_data)
 
--- a/lib/pbench/server/api/resources/query_apis/datasets_list.py
+++ b/lib/pbench/server/api/resources/query_apis/datasets_list.py
@@ -1,4 +1,3 @@
-from dateutil import parser
 from flask import jsonify
 from logging import Logger
 from typing import Any, AnyStr, Dict
@@ -98,7 +97,8 @@ class DatasetsList(ElasticBase):
                         }
                     },
                     "size": 5000,
-                }
+                },
+                "params": {"ignore_unavailable": "true"},
             },
         }
 
@@ -120,30 +120,28 @@ class DatasetsList(ElasticBase):
         datasets = []
         hits = es_json["hits"]["hits"]
         self.logger.info("{} controllers found", len(hits))
+        controller = None
         for dataset in hits:
             src = dataset["_source"]
             run = src["run"]
+            if not controller:
+                controller = run["controller"]
+            elif controller != run["controller"]:
+                self.logger.warning(
+                    "Expected all controllers to match: {} and {} don't match",
+                    controller,
+                    run["controller"],
+                )
             d = {
-                "key": run["name"],
-                "run.name": run["name"],
-                "run.controller": run["controller"],
-                "run.start": run["start"],
-                "run.end": run["end"],
+                "result": run["name"],
+                "controller": run["controller"],
+                "start": run["start"],
+                "end": run["end"],
                 "id": run["id"],
             }
-            try:
-                timestamp = parser.parse(run["start"]).utcfromtimestamp()
-            except Exception as e:
-                self.logger.info(
-                    "Can't parse start time {} to integer timestamp: {}",
-                    run["start"],
-                    e,
-                )
-                timestamp = dataset["sort"][0]
 
-            d["startUnixTimestamp"] = timestamp
             if "config" in run:
-                d["run.config"] = run["config"]
+                d["config"] = run["config"]
             if "prefix" in run:
                 d["run.prefix"] = run["prefix"]
             if "@metadata" in src:
@@ -153,5 +151,7 @@ class DatasetsList(ElasticBase):
                 if "satellite" in meta:
                     d["@metadata.satellite"] = meta["satellite"]
             datasets.append(d)
+        result = {controller: datasets}
+
         # construct response object
-        return jsonify(datasets)
+        return jsonify(result)
--- a/lib/pbench/test/unit/server/query_apis/conftest.py
+++ b/lib/pbench/test/unit/server/query_apis/conftest.py
@@ -1,0 +1,72 @@
+import pytest
+import requests
+from requests.exceptions import HTTPError
+
+
+def make_http_exception(status: int) -> HTTPError:
+    """
+    Helper to create a properly annotated HTTPError for testing
+
+    Args:
+        status: HTTP status code
+
+    Returns:
+        HTTPError object
+    """
+    response = requests.Response()
+    response.status_code = status
+    response.reason = "Fake reason"
+    return HTTPError(response=response)
+
+
+@pytest.fixture
+def query_api(client, server_config, requests_mock, request):
+    """
+    query_helper Help controller queries that want to interact with a mocked
+    Elasticsearch service.
+
+    This is a fixture which exposes a function of the same name that can be
+    used to set up and validate a mocked Elasticsearch query with a JSON
+    payload and an expected status.
+
+    Parameters to the mocked Elasticsearch POST are passed as keyword
+    parameters: these can be any of the parameters supported by the
+    request_mock post method. The most common are 'json' for the JSON
+    response payload, and 'exc' to throw an exception.
+
+    NOTE: the expected URI for the server query and the resulting Elasticsearch
+    query come from a test case parameter, which is passed through the Pytest
+    "mark" mechanism and accessed through the "param" attribute of the Pytest
+    request fixture. E.g.,
+
+            @pytest.mark.parametrize(
+                "query_api",
+                [
+                    {
+                        "pbench": "/controllers/list",
+                        "elastic": "/_search?ignore_unavailable=true",
+                    }
+                ],
+                indirect=True,
+            )
+
+    The first parameter must match the name of this fixture, and the second should
+    be a list (or tuple) with a single dict defining the "pbench" and "elastic"
+    URI paths.
+
+    :return: the response object for further checking
+    """
+
+    def query_api(payload, expected_index, expected_status, server_config, **kwargs):
+        host = server_config.get("elasticsearch", "host")
+        port = server_config.get("elasticsearch", "port")
+        es_url = f"http://{host}:{port}{expected_index}{request.param['elastic']}"
+        requests_mock.post(es_url, **kwargs)
+        response = client.post(
+            f"{server_config.rest_uri}{request.param['pbench']}", json=payload
+        )
+        assert requests_mock.last_request.url == es_url
+        assert response.status_code == expected_status
+        return response
+
+    return query_api
--- a/lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
+++ b/lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
@@ -1,42 +1,7 @@
 import pytest
-import re
 import requests
 from http import HTTPStatus
-
-
-@pytest.fixture
-def query_helper(client, server_config, requests_mock):
-    """
-    query_helper Help controller queries that want to interact with a mocked
-    Elasticsearch service.
-
-    This is a fixture which exposes a function of the same name that can be
-    used to set up and validate a mocked Elasticsearch query with a JSON
-    payload and an expected status.
-
-    Parameters to the mocked Elasticsearch POST are passed as keyword
-    parameters: these can be any of the parameters supported by the
-    request_mock post method. The most common are 'json' for the JSON
-    response payload, and 'exc' to throw an exception.
-
-    :return: the response object for further checking
-    """
-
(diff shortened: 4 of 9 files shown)


lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -428,65 +428,83 @@ def _call(self, method: Callable, json_data: Dict[AnyStr, Any]):
...
             Postprocessed JSON body to return to client
         """
+        klasname = self.__class__.__name__
         try:
             es_request = self.assemble(json_data)
             path = es_request.get("path")
             url = urljoin(self.es_url, path)
         except Exception as e:
-            self.logger.exception("Blew it in setup: {}", type(e).__name__)
+            self.logger.exception("{} blew it in setup: {}", klasname, type(e).__name__)
```
webbnh, line 437: Just out of curiosity, why do you like `type(e).__name__` better than `str(e)`?

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -428,65 +428,83 @@ def _call(self, method: Callable, json_data: Dict[AnyStr, Any]):
...
+            msg = f"{klasname}: the query postprocessor was unable to complete: {e}"
             self.logger.warning(msg)
             abort(HTTPStatus.BAD_REQUEST, message=msg)
         except KeyError as e:
-            self.logger.error("Missing Elasticsearch key {}", e)
+            self.logger.error("{}: missing Elasticsearch key {}", klasname, e)
             abort(
                 HTTPStatus.INTERNAL_SERVER_ERROR,
-                message="Missing Elasticsearch key {e}",
+                message=f"Missing Elasticsearch key {e}",
```
webbnh, line 502: :-D

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -510,10 +528,9 @@ def post(self):
         try:
             new_data = self.schema.validate(json_data)
         except SchemaError as e:
-            # The format string here is important as the value we're
-            # trying to convert might contain "{}" brackets that would
-            # be interpreted as formatting commands.
-            self.logger.warning("{}", str(e))
+            self.logger.warning(
+                "{}: {} on {!r}", self.__class__.__name__, str(e), json_data
```
webbnh, line 532: How is `self.__class__.__name__` different from and/or better than `klasname`, here?

server/bin/unittests
```diff
@@ -259,12 +259,13 @@ function _verify_output {
...
-    # symptoms later.
+    # NOTE: we ignore chatty stdout messages, but capture error output
+    # and fail if the command fails.
     pbench-user-create --username=${name} --email=${name}@example.com \
-        --first-name=${name^} --last-name=User --password=password >/dev/null 2>&1
+        --first-name=${name^} --last-name=User --password=password >/dev/null
+    if [[ ${?} -ne 0 ]]; then
+        exit 101
+    fi
 }
```
webbnh, line 269: Contrary to the `NOTE`, you're not actually capturing the error output (here) -- presumably it escapes and is captured by the invoking context.  If that's your intention, it would probably be good to correct/clarify the comment; otherwise...its not hard to actually capture the error output, and then, in the event of an error, do something useful with it.

lib/pbench/test/unit/server/test_datasets_list.py
```diff
@@ -158,26 +160,28 @@ def test_query(self, client, server_config, query_helper, user_ok, find_template
...
-        assert res_json[0]["@metadata.controller_dir"] == "dhcp31-187.example.com"
+        assert isinstance(res_json, dict)
+        assert len(res_json.keys()) == 1
+        data = res_json["dhcp31-187.example.com"]
+        assert isinstance(data, list)
+        assert len(data) == 1
+        # assert (
+        #     data[0]["key"]
+        #     == "fio_rhel8_kvm_perf43_preallocfull_nvme_run4_iothread_isolcpus_2020.04.29T12.49.13"
+        # )
```
webbnh, line 171: Should this code be removed?

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -428,65 +428,83 @@ def _call(self, method: Callable, json_data: Dict[AnyStr, Any]):
...
             es_response.raise_for_status()
         except requests.exceptions.HTTPError as e:
-            self.logger.exception("HTTP error {} from Elasticsearch request", e)
+            self.logger.exception(
+                "{} HTTP error {} from Elasticsearch request", klasname, e
+            )
+            # FIXME: I can't seem to figure out how to give the mock exception
+            # the proper attributes; so rather than have this blow up, or
+            # remove the information for production deployment, simply avoid
+            # trying to access them in unit tests.
```
webbnh, line 451: I think you can set these attributes via `request` and `response` keyword arguments to the `requests.exceptions.HTTPError` c'tor.  Hopefully, that will allow you to removed this stuff from here.

lib/pbench/server/api/resources/query_apis/datasets_list.py
```diff
@@ -120,30 +120,29 @@ def postprocess(self, es_json: Dict[AnyStr, Any]) -> Dict[AnyStr, Any]:
...
             run = src["run"]
+            if controller:
+                if controller != run["controller"]:
+                    self.logger.warning(
+                        "Expected all controllers to match: {} and {} don't match",
+                        controller,
+                        run["controller"],
+                    )
+            else:
+                controller = run["controller"]
```
webbnh, line 135: If I might suggest....
```
            if not controller:
                controller = run["controller"]
            elif controller != run["controller"]:
                self.logger.warning(
                    "Expected all controllers to match: {} and {} don't match",
                    controller,
                    run["controller"],
                )
```
This saves a level of indentation and a line of screen realestate.

lib/pbench/server/api/resources/query_apis/datasets_list.py
```diff
@@ -120,30 +120,29 @@ def postprocess(self, es_json: Dict[AnyStr, Any]) -> Dict[AnyStr, Any]:
...
-                    e,
-                )
-                timestamp = dataset["sort"][0]
 
-            d["startUnixTimestamp"] = timestamp
             if "config" in run:
-                d["run.config"] = run["config"]
+                d["config"] = run["config"]
             if "prefix" in run:
                 d["run.prefix"] = run["prefix"]
```
webbnh, line 147: You've removed the other instances of the `run.` key prefix; should it be removed from here, too?

webbnh (dismissed):
Looks good, although among the below there are a couple of items you might want to address.  Nevertheless, I would prefer that you remove the code accommodating unit testing from `api/resources/query_apis/__init__.py`, if my suggestion works.

lib/pbench/server/api/resources/query_apis/__init__.py
dbutenhof: Mostly, in this case, to give a concise summary without being overly redundant, since the `logger.exception` will dump everything. Sometimes it's hard to identify the actual exception from the stringification. Said stringification will appear in the exception dump anyway, so I just added the exception name.

lib/pbench/server/api/resources/query_apis/__init__.py
dbutenhof: Yes, you can; however I had trouble figuring out a way to construct a useful `Response` object in order to do that; the constructor, it turns out, doesn't allow specifying attributes. You have to create it and then set them later, which in this context is awkward. It did occur to me to try just creating a class `HTTPError` with a Response object and using that here instead of specifying the class name, although it's not clear from documentation exactly what the requests mock's assumptions are regarding this parameter; in the end I decided just to cut the mess short and get on with it. As frequently happens, you make me question that again ...

This, by the way, came out of fixing another instance of a broken f-string... we never cared before because, although I'd included the parameters, they were never interpreted! I really hate that I made a trivial single character fix, and broke all my unit tests in a way that proved exceedingly difficult to work around and harder to fix.

lib/pbench/server/api/resources/query_apis/__init__.py
dbutenhof: I think I just wrote it out twice before realizing I should capture the name once ... and then I missed fixing this one.

lib/pbench/server/api/resources/query_apis/datasets_list.py
dbutenhof: Sure.

lib/pbench/server/api/resources/query_apis/datasets_list.py
dbutenhof: Nope; this prefix is in the original dashboard code, but I'd missed that `run["config"]` was stored as just `"config"`. I don't even know where I got the timestamp stuff from; I swear I saw it in the dashboard, but when I started double-checking today I couldn't find any trace of such a field. Weird.

I was more uncertain about dropping `"key"` ... it's __not__ in the dashboard, however the dashboard has a lot of React complaints that array elements should have unique `"key"` values, and I might have added that one deliberately in an attempt to quiet them. However I'm not sure that actually helped anything, the dashboard code itself doesn't need it, and I decided to stick to the original dashboard model for this pass.

lib/pbench/test/unit/server/test_datasets_list.py
dbutenhof: Yeah; I glanced through quickly right after I posted, and noticed this as well. Again, I was on the edge about removing `"key"` so I guess you could say I kept half of it. Maybe 1/4 since it's only commented unit test code ... ;-)

server/bin/unittests
dbutenhof: The problem with writing a comment first is that it's easy to miss updating it as the code evolves.

The worst thing is that this appears to have completely broken the unit tests. I'd mentioned I was having troubles with the tox environment on my laptop; I decided to just push this PR and see what happened, but the `unittests` are failing the same way in Jenkins that they are for me, so there's something more systemic apparently triggered by this change... though I can't immediately see what it is. This, of course, was the original motivation for the branch that became this PR, so it'd be ironic if I ended up just pulling it to get the rest through...

lib/pbench/server/api/resources/query_apis/__init__.py
dbutenhof: Ah; actually, because this is a different method and `POST` has only this one use. I suppose I could pull `klasname` up to an instance variable, but that seemed somehow weird ...

server/bin/unittests
webbnh: Are you working with Nikhil's changes to move the database out of memory to a temporary file for the unit tests which need it to persist between invocations?

dbutenhof: Not relevant for the gold tests; we use an sqlite3 DB.

lib/pbench/server/api/resources/query_apis/__init__.py
webbnh: > I really hate that I made a trivial single character fix, and broke all my unit tests

:-D  Oh, I'm so sorry!  :-}

Anyway, since this is all based on the Python `requests` package, can't you use the `responses` package to mock the response from the call to ES?  That seems like it should allow you do drive the behavior of the code in question from "outside" without having to make special accommodation for it being run in a unit test.

dbutenhof: In theory, yes; except the issue here is the `exc=` kwarg we're using to mock a failure exception, and there's no special help I can find in constructing this. It's normally shown as just the class name, and we're not given a properly constructed instance of that type so it causes a mess.

The new commit uses a function to construct a properly attributed exception, but it's a bit ugly. It's more ugly because there's no real way to do this in a common fixture, so it's duplicated. I really want to find some time to refactor the query_api tests as a class hierarchy as I did for the implementation ... it's just not made it to the top of my list. (Although this PR pushed my patience a bit, I really didn't want to take the extra time that would have required before getting up the dataset changes and other fixes here.)

lib/pbench/test/unit/server/test_controllers_list.py
```diff
@@ -206,10 +223,7 @@ def test_empty_query(
     @pytest.mark.parametrize(
         "exceptions",
         (
-            {
-                "exception": requests.exceptions.HTTPError,
-                "status": HTTPStatus.BAD_GATEWAY,
-            },
+            {"exception": make_http_exception(502), "status": HTTPStatus.BAD_GATEWAY},
```
webbnh, line 226: Shouldn't that be `HTTPStatus.BAD_GATEWAY` instead of `502`?

lib/pbench/test/unit/server/test_datasets_detail.py
```diff
@@ -292,10 +309,7 @@ def test_nonunique_query(
     @pytest.mark.parametrize(
         "exceptions",
         (
-            {
-                "exception": requests.exceptions.HTTPError,
-                "status": HTTPStatus.BAD_GATEWAY,
-            },
+            {"exception": make_http_exception(502), "status": HTTPStatus.BAD_GATEWAY},
```
webbnh, line 312: Shouldn't that be `HTTPStatus.BAD_GATEWAY` instead of `502`?

lib/pbench/test/unit/server/test_datasets_list.py
```diff
@@ -186,7 +199,7 @@ def test_query(self, client, server_config, query_helper, user_ok, find_template
     @pytest.mark.parametrize(
         "exceptions",
         (
-            {"exception": requests.exceptions.HTTPError, "status": 502},
+            {"exception": make_http_exception(502), "status": 502},
```
webbnh, line 202: Shouldn't that be `HTTPStatus.BAD_GATEWAY` instead of `502`?

:-)

lib/pbench/test/unit/server/test_month_indices.py
```diff
@@ -106,7 +123,7 @@ def test_query(self, client, server_config, get_helper, find_template):
     @pytest.mark.parametrize(
         "exceptions",
         (
-            {"exception": requests.exceptions.HTTPError, "status": 502},
+            {"exception": make_http_exception(502), "status": 502},
```
webbnh, line 126: Shouldn't that be `HTTPStatus.BAD_GATEWAY` instead of `502`?

;-)

webbnh (commented):
I'm waiting to see what you think of using the `responses` package to drive the ES-based behaviors.

lib/pbench/test/unit/server/test_controllers_list.py
dbutenhof: I hadn't updated all of the subclasses with names, and I lazily copied-and-pasted the calls from one where I hadn't. Sigh.

lib/pbench/test/unit/server/test_datasets_detail.py
```diff
@@ -2,6 +2,23 @@
 import re
 import requests
 from http import HTTPStatus
+from requests.exceptions import HTTPError
+
+
+def make_http_exception(status: int) -> HTTPError:
```
portante, line 8: Seems like this should land in a common module for the tests.

portante (changes requested):
Looks pretty good, just a test refactoring to avoid that duplicate code.

lib/pbench/test/unit/server/test_datasets_detail.py
dbutenhof: I'd like to refactor all the `query_api` tests into a class hierarchy like I did on the implementation side; I just keep putting that off for a separate PR I haven't yet gotten to. Aside from that, I'm not entirely sure where to put it. It can't really be a fixture, so it doesn't really belong in `conftest.py`, but it doesn't quite belong anywhere else, either.

portante (dismissed):

lib/pbench/test/unit/server/query_apis/test_datasets_list.py
```diff
@@ -156,41 +131,58 @@ def test_query(self, client, server_config, query_helper, user_ok, find_template
...
+                "status": HTTPStatus.BAD_GATEWAY,
+            },
+            {
+                "exception": requests.exceptions.ConnectionError,
+                "status": HTTPStatus.BAD_GATEWAY,
+            },
+            {
+                "exception": requests.exceptions.Timeout,
+                "status": HTTPStatus.GATEWAY_TIMEOUT,
+            },
```
webbnh, line 171: Do we really want to report these as "gateway" issues?  In these cases, the server is a client, not a proxy, right?  So, won't this confuse things if, for instance, there is an actual proxy between the actual client and the Pbench server?

webbnh (approved):
Looks good; found a nit, hit some confusion, and I've got a pointed question for you.

lib/pbench/test/unit/server/query_apis/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+    response payload, and 'exc' to throw an exception.
+
+    :return: the response object for further checking
+    """
+
+    def query_api(payload, expected_index, expected_status, server_config, **kwargs):
+        host = server_config.get("elasticsearch", "host")
+        port = server_config.get("elasticsearch", "port")
+        es_url = f"http://{host}:{port}"
+        requests_mock.post(re.compile(f"{es_url}"), **kwargs)
```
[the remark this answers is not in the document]
dbutenhof: There was a reason for this at one point; but that time seems to have passed. Huh.

lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
```diff
@@ -118,7 +83,12 @@ def test_bad_dates(self, client, server_config, user_ok):
             == "Value '2020-19' (str) cannot be parsed as a date/time string"
         )
 
-    def test_query(self, client, server_config, query_helper, user_ok, find_template):
+    @pytest.mark.parametrize(
+        "query_api",
+        [{"pbench": "/datasets/detail", "elastic": "/_search?ignore_unavailable=true"}],
+        indirect=True,
+    )
+    def test_query(self, client, server_config, query_api, user_ok, find_template):
```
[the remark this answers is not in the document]
dbutenhof: I haven't found a better way. In fact, the last time I tried pytest magic to pass parameters to a fixture, I couldn't get it working and simply gave up. This time I seem to have achieved success; but I agree it's another little bit of obscure pytest lore. I'll add a description to the fixture, which I probably should have done anyway.

lib/pbench/test/unit/server/query_apis/test_datasets_list.py
dbutenhof: That's an issue completely separate from this PR, since it's existing code. The idea at the time was that we're reporting an external failure in a subsystem for which we're essentially acting as a proxy; but that really applies better to the pass-through Elasticsearch API than for the real server queries where the backend DB should be invisble.

Again, though, I'd rather not dilute the focus of this PR by dealing with those issues here.

webbnh: Tracked in #2327.  (GitHub makes it _very_ easy to create issues from PR comments!)

lib/pbench/test/unit/server/query_apis/conftest.py
```diff
@@ -35,20 +34,38 @@ def query_api(client, server_config, requests_mock, request):
...
+
     :return: the response object for further checking
     """
 
     def query_api(payload, expected_index, expected_status, server_config, **kwargs):
         host = server_config.get("elasticsearch", "host")
         port = server_config.get("elasticsearch", "port")
-        es_url = f"http://{host}:{port}"
-        requests_mock.post(re.compile(f"{es_url}"), **kwargs)
+        es_url = f"http://{host}:{port}" + expected_index + request.param["elastic"]
```
webbnh, line 63: The mixture of f-string and concatenation is slightly odd...I think I would recommend one or the other.  E.g., I think
```
        es_url = f"http://{host}:{port}{expected_index}{request.param['elastic']}"
```
is more compact.

lib/pbench/test/unit/server/query_apis/test_month_indices.py
```diff
@@ -27,7 +26,7 @@ def get_helper(expected_status, server_config, **kwargs):
         host = server_config.get("elasticsearch", "host")
         port = server_config.get("elasticsearch", "port")
         es_url = f"http://{host}:{port}"
-        requests_mock.get(re.compile(f"{es_url}"), **kwargs)
+        requests_mock.get(es_url + "/_aliases", **kwargs)
         response = client.get(f"{server_config.rest_uri}/controllers/months")
         assert requests_mock.last_request.url == (es_url + "/_aliases")
```
webbnh, line 31: Did you want to do here what you did in `lib/pbench/test/unit/server/query_apis/conftest.py` and put `/_aliases` into `es_url`?

webbnh (approved):
Looks good (just a couple of nits).

portante (approved):

lib/pbench/test/unit/server/query_apis/conftest.py
dbutenhof: Yeah; a result of fixing it by cutting and pasting and not paying enough attention.

lib/pbench/test/unit/server/query_apis/test_month_indices.py
dbutenhof: I noticed that right after I pushed the commit. Dang.

portante (approved):

webbnh (approved):

dbutenhof:
Squashed and ready.

webbnh (approved):
Go!

portante (approved):
Ship it!

PR #2328, opened 2021-05-20
Fio postprocess fix

portante:


agent/bench-scripts/postprocess/BenchPostprocess.pm
```diff
@@ -504,6 +505,7 @@ sub calc_aggregate_metrics {
 					# A timeseries is considered valid if it has at least 2 timestamps
 					if ((defined $$workload_ref{$metric_class}{$metric_type}[$i]{get_label('timeseries_label')}) and
 					    (scalar get_timestamps(\@{ $$workload_ref{$metric_class}{$metric_type}[$i]{get_label('timeseries_label')}}) > 1)) {
+						# FIXME - Why are we calling get_timestamps(), which creates, sorts, and returns a new array only to throw it away?
```
dbutenhof, line 508: Wait ... just to find out if the returned collection length is > 1?? Wow. Dang, I increasingly dislike Perl idioms.

OK, so `get_timestamps` just iterates through the specified array reference, which I guess is `workload_ref{$metric_class}{metric_type}[$1]{get_label('timeseries_label')}}`, pulls out the `date` field from each `{'date' => x, 'value' => y }` hash, and returns that list neatly sorted. So both the full iteration and especially the sort are wasted in this context. Instead, assuming this is truly necessary at all, just determine whether there's more than one element in the list ... either by just taking the scalar of the list itself in the first place, or by iterating until `n > 1` and dropping out with `true`.

It's trying to "efficiently" reuse an existing method, but, really?

dbutenhof (commented):
I had to skim it. It's like posting a train wreck. I haven't even attempted to completely absorb this, but I did comment on your comment...

agent/bench-scripts/postprocess/BenchPostprocess.pm
webbnh: > I increasingly dislike Perl idioms.

Heh.  You've been working in Python too long -- in Perl, you've got **_options_**!  :-D

> It's trying to "efficiently" reuse an existing method, but, really?

Yup...I think that's it.  And, as long as you never run with a sufficiently large data set (or never run interactively), you'll never know the error of your ways.  ;-)

dbutenhof: > > I increasingly dislike Perl idioms.
> 
> Heh. You've been working in Python too long -- in Perl, you've got **_options_**! :-D

Options; and `$` and `%` and `@` permutations that make Bliss dereference semantics seem simple. I get dizzy just looking at this code.

Perl starts to remind me a bit of RPG II sometimes ... if only in the sense when browsing code that I'm looking at random gibberish rather than something sensible.

(Actually, before coming here I'd mostly been writing Java for a decade, and C/C++ before that; but Instant Capacity tests were Perl, and they were particularly bad Perl... though I'm no longer certain there's any Perl that's not bad! I resisted Python for years because the space-sensitive indentation reminded me of FORTRAN, but once I switched for casual scripting I never wanted to look back. Not to mention that while Python OO isn't pretty, Perl's is full of squirrels.)

webbnh:
This PR is superseded by #2349.

PR #2329, opened 2021-05-20
Some Refactor and change Http codes to use symbolic names

npalaska:
- Attempt at refactoring few pieces of convoluted server API code
- Change Http status codes to use the respective symbolic names

### DIFF
--- a/lib/pbench/server/api/resources/users_api.py
+++ b/lib/pbench/server/api/resources/users_api.py
@@ -1,12 +1,12 @@
-import jwt
 from http import HTTPStatus
+import jwt
 from flask import request, jsonify, make_response
 from flask_restful import Resource, abort
 from flask_bcrypt import check_password_hash
 from email_validator import EmailNotValidError
 from sqlalchemy.exc import SQLAlchemyError, IntegrityError
-from sqlalchemy.orm.exc import NoResultFound
-from pbench.server.database.models.users import User, AdminUserDeletion
+from typing import NamedTuple
+from pbench.server.database.models.users import User
 from pbench.server.database.models.active_tokens import ActiveTokens
 from pbench.server.api.auth import Auth
 
@@ -49,17 +49,17 @@ class RegisterUser(Resource):
         user_data = request.get_json()
         if not user_data:
             self.logger.warning("Invalid json object: {}", request.url)
-            abort(400, message="Invalid json object in request")
+            abort(HTTPStatus.BAD_REQUEST, message="Invalid json object in request")
 
         username = user_data.get("username")
         if not username:
             self.logger.warning("Missing username field")
-            abort(400, message="Missing username field")
+            abort(HTTPStatus.BAD_REQUEST, message="Missing username field")
         username = username.lower()
         if User.is_admin_username(username):
             self.logger.warning("User tried to register with admin username")
             abort(
-                400, message="Please choose another username",
+                HTTPStatus.BAD_REQUEST, message="Please choose another username",
             )
 
         # check if provided username already exists
@@ -67,41 +67,41 @@ class RegisterUser(Resource):
             user = User.query(username=user_data.get("username"))
         except Exception:
             self.logger.exception("Exception while querying username")
-            abort(500, message="INTERNAL ERROR")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
         if user:
             self.logger.warning(
                 "A user tried to re-register. Username: {}", user.username
             )
-            abort(403, message="Provided username is already in use.")
+            abort(HTTPStatus.FORBIDDEN, message="Provided username is already in use.")
 
         password = user_data.get("password")
         if not password:
             self.logger.warning("Missing password field")
-            abort(400, message="Missing password field")
+            abort(HTTPStatus.BAD_REQUEST, message="Missing password field")
 
         email_id = user_data.get("email")
         if not email_id:
             self.logger.warning("Missing email field")
-            abort(400, message="Missing email field")
+            abort(HTTPStatus.BAD_REQUEST, message="Missing email field")
         # check if provided email already exists
         try:
             user = User.query(email=email_id)
         except Exception:
             self.logger.exception("Exception while querying user email")
-            abort(500, message="INTERNAL ERROR")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
         if user:
             self.logger.warning("A user tried to re-register. Email: {}", user.email)
-            abort(403, message="Provided email is already in use")
+            abort(HTTPStatus.FORBIDDEN, message="Provided email is already in use")
 
         first_name = user_data.get("first_name")
         if not first_name:
             self.logger.warning("Missing first_name field")
-            abort(400, message="Missing first_name field")
+            abort(HTTPStatus.BAD_REQUEST, message="Missing first_name field")
 
         last_name = user_data.get("last_name")
         if not last_name:
             self.logger.warning("Missing last_name field")
-            abort(400, message="Missing last_name field")
+            abort(HTTPStatus.BAD_REQUEST, message="Missing last_name field")
 
         try:
             user = User(
@@ -117,13 +117,13 @@ class RegisterUser(Resource):
             self.logger.info(
                 "New user registered, username: {}, email: {}", username, email_id
             )
-            return "", 201
+            return "", HTTPStatus.CREATED
         except EmailNotValidError:
             self.logger.warning("Invalid email {}", email_id)
-            abort(400, message=f"Invalid email: {email_id}")
+            abort(HTTPStatus.BAD_REQUEST, message=f"Invalid email: {email_id}")
         except Exception:
             self.logger.exception("Exception while registering a user")
-            abort(500, message="INTERNAL ERROR")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
 
 class Login(Resource):
@@ -170,35 +170,35 @@ class Login(Resource):
         post_data = request.get_json()
         if not post_data:
             self.logger.warning("Invalid json object: {}", request.url)
-            abort(400, message="Invalid json object in request")
+            abort(HTTPStatus.BAD_REQUEST, message="Invalid json object in request")
 
         username = post_data.get("username")
         if not username:
             self.logger.warning("Username not provided during the login process")
-            abort(400, message="Please provide a valid username")
+            abort(HTTPStatus.BAD_REQUEST, message="Please provide a valid username")
 
         password = post_data.get("password")
         if not password:
             self.logger.warning("Password not provided during the login process")
-            abort(400, message="Please provide a valid password")
+            abort(HTTPStatus.BAD_REQUEST, message="Please provide a valid password")
 
         try:
             # fetch the user data
             user = User.query(username=username)
         except Exception:
             self.logger.exception("Exception occurred during user login")
-            abort(500, message="INTERNAL ERROR")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
         if not user:
             self.logger.warning(
                 "No user found in the db for Username: {} while login", username
             )
-            abort(401, message="Bad login")
+            abort(HTTPStatus.UNAUTHORIZED, message="Bad login")
 
         # Validate the password
         if not check_password_hash(user.password, password):
             self.logger.warning("Wrong password for user: {} during login", username)
-            abort(401, message="Bad login")
+            abort(HTTPStatus.UNAUTHORIZED, message="Bad login")
 
         try:
             auth_token = self.auth.encode_auth_token(
@@ -214,7 +214,7 @@ class Login(Resource):
                 "Could not encode the JWT auth token for user: {} while login", username
             )
             abort(
-                500, message="INTERNAL ERROR",
+                HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR",
             )
 
         # Add the new auth token to the database for later access
@@ -228,21 +228,21 @@ class Login(Resource):
             self.logger.warning(
                 "Duplicate auth token got created, user might have tried to re-login immediately"
             )
-            abort(409, message="Retry login after some time")
+            abort(HTTPStatus.CONFLICT, message="Login collision; please wait and retry")
         except SQLAlchemyError as e:
             self.logger.error(
                 "SQLAlchemy Exception while logging in a user {}", type(e)
             )
-            abort(500, message="INTERNAL ERROR")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
         except Exception:
             self.logger.exception("Exception while logging in a user")
-            abort(500, message="INTERNAL ERROR")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
         response_object = {
             "auth_token": auth_token,
             "username": username,
         }
-        return make_response(jsonify(response_object), 200)
+        return make_response(jsonify(response_object), HTTPStatus.OK)
 
 
 class Logout(Resource):
@@ -282,13 +282,13 @@ class Logout(Resource):
                 ActiveTokens.delete(auth_token)
             except Exception:
                 self.logger.exception("Exception occurred deleting an auth token")
-                abort(500, message="INTERNAL ERROR")
+                abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
             else:
                 self.logger.debug("User {} logged out", user.username)
         else:
             self.logger.info("User logout with invalid or expired token")
 
-        return "", 200
+        return "", HTTPStatus.OK
 
 
 class UserAPI(Resource):
@@ -296,12 +296,62 @@ class UserAPI(Resource):
     Abstracted pbench API to get user data
     """
 
+    TargetUser = NamedTuple(
+        "TargetUser",
+        [("target_user", User), ("http_status", HTTPStatus), ("http_message", str)],
+    )
+
     def __init__(self, logger, auth):
         self.logger = logger
         self.auth = auth
 
+    def get_valid_target_user(
+        self, target_username: str, request_type: str
+    ) -> "UserAPI.TargetUser":
+        """
+        Helper function to determine whether the API call is permitted for the target username
+        Right now it is only permitted for an admin user and the target user itself.
+        This returns a target User on success or None on failure; in the case of failure,
+        also returns the corresponding HTTP status code and message string
+        """
+        current_user = self.auth.token_auth.current_user()
+        if current_user.username == target_username:
+            return UserAPI.TargetUser(
+                target_user=current_user, http_status=HTTPStatus.OK, http_message=""
+            )
+        if current_user.is_admin():
+            target_user = User.query(username=target_username)
+            if target_user:
+                return UserAPI.TargetUser(
+                    target_user=target_user, http_status=HTTPStatus.OK, http_message=""
+                )
+
+            self.logger.warning(
+                "User {} requested {} operation but user {} is not found.",
+                current_user.username,
+                request_type,
+                target_username,
+            )
+            return UserAPI.TargetUser(
+                target_user=None,
+                http_status=HTTPStatus.NOT_FOUND,
+                http_message=f"User {target_username} not found",
+            )
+
+        self.logger.warning(
+            "User {} is not authorized to {} user {}.",
+            current_user.username,
+            request_type,
+            target_username,
+        )
+        return UserAPI.TargetUser(
+            target_user=None,
+            http_status=HTTPStatus.FORBIDDEN,
+            http_message=f"Not authorized to access user {target_username}",
+        )
+
     @Auth.token_auth.login_required()
-    def get(self, username):
+    def get(self, target_username):
         """
         Get request for getting user data.
         This requires a Pbench auth token in the header field
@@ -325,37 +375,14 @@ class UserAPI(Resource):
                         "message": "failure message"
                     }
         """
-        try:
-            user, verified = self.auth.verify_user(username)
-        except Exception:
-            self.logger.exception("Exception occurred during verifying the user")
-            abort(500, message="INTERNAL ERROR")
-
-        # TODO: Check if the user has the right privileges
-        if verified:
-            response_object = user.get_json()
-            return make_response(jsonify(response_object), 200)
-        elif user.is_admin():
-            try:
-                target_user = User.query(username=username)
-                response_object = target_user.get_json()
-                return make_response(jsonify(response_object), 200)
-            except Exception:
-                self.logger.exception(
-                    "Exception occurred while querying the user. Username: {}", username
-                )
-                abort(500, message="INTERNAL ERROR")
-        else:
-            self.logger.warning(
-                "User {} is not authorized to get user {}.", user.username, username,
-            )
-            abort(
-                403,
-                message=f"Not authorized to access information about user {username}",
-            )
+        result = self.get_valid_target_user(target_username, "GET")
+        if not result.target_user:
+            abort(result.http_status, message=result.http_message)
+        response_object = result.target_user.get_json()
+        return make_response(jsonify(response_object), HTTPStatus.OK)
 
     @Auth.token_auth.login_required()
-    def put(self, username):
+    def put(self, target_username):
         """
         PUT request for updating user data.
         This requires a Pbench auth token in the header field
@@ -387,31 +414,19 @@ class UserAPI(Resource):
                         "message": "failure message"
                     }
         """
-        post_data = request.get_json()
-        if not post_data:
+        user_payload = request.get_json()
+        if not user_payload:
             self.logger.warning("Invalid json object: {}", request.url)
-            abort(400, message="Invalid json object in request")
+            abort(HTTPStatus.BAD_REQUEST, message="Invalid json object in request")
 
-        try:
-            user, verified = self.auth.verify_user(username)
-        except Exception:
-            self.logger.exception("Exception occurred while verifying the user")
-            abort(500, message="INTERNAL ERROR")
-
-        # TODO: Check if the user has the right privileges
-        if not verified:
-            self.logger.warning(
-                "User {} is not authorized to delete user {}.", user.username, username,
-            )
-            abort(
-                403,
-                message=f"Not authorized to update information about user {username}",
-            )
+        result = self.get_valid_target_user(target_username, "PUT")
+        if not result.target_user:
+            abort(result.http_status, message=result.http_message)
 
         # Check if the user payload contain fields that are either protected or
         # are not present in the user db. If any key in the payload does not match
         # with the column name we will abort the update request.
-        non_existent = set(post_data.keys()).difference(
+        non_existent = set(user_payload.keys()).difference(
             set(User.__table__.columns.keys())
         )
         if non_existent:
@@ -419,27 +434,40 @@ class UserAPI(Resource):
                 "User trying to update fields that are not present in the user database. Fields: {}",
                 non_existent,
             )
-            abort(400, message="Invalid fields in update request payload")
-        protected = set(post_data.keys()).intersection(set(User.get_protected()))
+            abort(
+                HTTPStatus.BAD_REQUEST,
+                message="Invalid fields in update request payload",
+            )
+        # Only admin user will be allowed to change other user's role. However,
+        # Admin users will not be able to change their admin role,
+        # This is done to prevent last admin user from de-admining him/herself
+        protected_db_fields = User.get_protected()
+        if (
+            not self.auth.token_auth.current_user().is_admin()
+            or self.auth.token_auth.current_user() == result.target_user
+        ):
+            protected_db_fields.append("role")
+
+        protected = set(user_payload.keys()).intersection(set(protected_db_fields))
         for field in protected:
-            if getattr(user, field) != post_data[field]:
+            if getattr(result.target_user, field) != user_payload[field]:
                 self.logger.warning(
                     "User trying to update the non-updatable fields. {}: {}",
                     field,
-                    post_data[field],
+                    user_payload[field],
                 )
-                abort(403, message="Invalid update request payload")
+                abort(HTTPStatus.FORBIDDEN, message="Invalid update request payload")
         try:
-            user.update(**post_data)
+            result.target_user.update(**user_payload)
         except Exception:
             self.logger.exception("Exception occurred during updating user object")
-            abort(500, message="INTERNAL ERROR")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
-        response_object = user.get_json()
-        return make_response(jsonify(response_object), 200)
+        response_object = result.target_user.get_json()
+        return make_response(jsonify(response_object), HTTPStatus.OK)
 
     @Auth.token_auth.login_required()
-    def delete(self, username):
+    def delete(self, target_username):
         """
         Delete request for deleting a user from database.
         This requires a Pbench auth token in the header field
@@ -457,45 +485,32 @@ class UserAPI(Resource):
                         "message": "failure message"
                     }
         """
-        try:
-            current_auth_user, verified = self.auth.verify_user(username)
-        except Exception:
-            self.logger.exception("Exception occurred during the getUser {}", username)
-            abort(500, message="INTERNAL ERROR")
-
-        # TODO: Check if the user has the right privileges
-        if not verified and not current_auth_user.is_admin():
+        result = self.get_valid_target_user(target_username, "DELETE")
+        if not result.target_user:
+            abort(result.http_status, message=result.http_message)
+        # Do not allow admin user to get self deleted via API
+        if (
+            result.target_user.is_admin()
+            and self.auth.token_auth.current_user() == result.target_user
+        ):
             self.logger.warning(
-                "User {} is not authorized to delete user {}.",
-                current_auth_user.username,
-                username,
+                "Admin user is not allowed to self delete via API call. Username: {}",
+                target_username,
             )
-            abort(403, message="Not authorized to delete user")
+            abort(HTTPStatus.FORBIDDEN, message="Not authorized to delete user")
 
+        # If target user is a valid and not an admin proceed to delete
         try:
-            target_user = User.query(username=username)
-            # Do not delete if the user is an admin
-            if target_user.is_admin():
-                raise AdminUserDeletion(username)
-            User.delete(username)
+            User.delete(target_username)
             self.logger.info(
                 "User entry deleted for user with username: {}, by user: {}",
-                username,
-                current_auth_user.username,
+                target_username,
+                self.auth.token_auth.current_user().username,
             )
-        except NoResultFound:
-            self.logger.info(
-                "Delete operation on non-existent user '{}'", username,
-            )
-            abort(HTTPStatus.NOT_FOUND, message="User not found")
-        except AdminUserDeletion as exc:
-            self.logger.info(str(exc))
-            abort(HTTPStatus.FORBIDDEN, message=f"User {username} can not be deleted")
         except Exception:
             self.logger.exception(
-                "Exception occurred during deleting the user entry for user '{}'",
-                username,
+                "Exception occurred while deleting a user {}", target_username,
             )
-            abort(500, message="INTERNAL ERROR")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
-        return "", 200
+        return "", HTTPStatus.OK
--- a/lib/pbench/server/api/__init__.py
+++ b/lib/pbench/server/api/__init__.py
@@ -99,7 +99,7 @@ def register_endpoints(api, app, config):
     )
     api.add_resource(
         UserAPI,
-        f"{base_uri}/user/<string:username>",
+        f"{base_uri}/user/<string:target_username>",
         resource_class_args=(logger, token_auth),
     )
 
--- a/lib/pbench/server/api/auth.py
+++ b/lib/pbench/server/api/auth.py
@@ -76,21 +76,6 @@ class Auth:
         except Exception as e:
             Auth.logger.exception(f"{__name__}: ERROR: {e.__traceback__}")
 
-    def verify_user(self, username):
-        """
-        Check if the provided username belongs to the current user by
-        querying the Usermodel with the current user
-        :param username:
-        :param logger
-        :return: User (UserModel instance), verified status (boolean)
-        """
-        user = User.query(id=self.token_auth.current_user().id)
-        # check if the current username matches with the one provided
-        verified = user is not None and user.username == username
-        Auth.logger.warning("verified status of user '{}' is '{}'", username, verified)
-
-        return user, verified
-
     def get_auth_token(self, logger):
         # get auth token
         auth_header = request.headers.get("Authorization")
--- a/lib/pbench/server/api/resources/endpoint_configure.py
+++ b/lib/pbench/server/api/resources/endpoint_configure.py
@@ -1,3 +1,4 @@
+from http import HTTPStatus
 import re
 from flask import request, jsonify
 from flask.globals import current_app
@@ -163,7 +164,7 @@ class EndpointConfig(Resource):
             response = jsonify(endpoints)
         except Exception:
             self.logger.exception("Something went wrong constructing the endpoint info")
-            abort(500, message="INTERNAL ERROR")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
         else:
-            response.status_code = 200
+            response.status_code = HTTPStatus.OK
             return response
--- a/lib/pbench/server/api/resources/graphql_api.py
+++ b/lib/pbench/server/api/resources/graphql_api.py
@@ -1,3 +1,4 @@
+from http import HTTPStatus
 import requests
 from flask_restful import Resource, abort
 from flask import request, make_response
@@ -19,7 +20,7 @@ class GraphQL(Resource):
         if not json_data:
             message = "Invalid json object"
             self.logger.warning(f"{message}: {request.url}")
-            abort(400, message=message)
+            abort(HTTPStatus.BAD_REQUEST, message=message)
 
         try:
             # query GraphQL
@@ -30,24 +31,28 @@ class GraphQL(Resource):
             abort(gql_response.status_code, message=f"HTTP error {e} from GraphQL")
         except requests.exceptions.ConnectionError:
             self.logger.exception("Connection refused during the GraphQL post request")
-            abort(502, message="Network problem, could not post to GraphQL Endpoint")
+            abort(
+                HTTPStatus.BAD_GATEWAY,
+                message="Network problem, could not post to GraphQL Endpoint",
+            )
         except requests.exceptions.Timeout:
             self.logger.exception(
                 "Connection timed out during the GraphQL post request"
             )
             abort(
-                504, message="Connection timed out, could not post to GraphQL Endpoint"
+                HTTPStatus.GATEWAY_TIMEOUT,
+                message="Connection timed out, could not post to GraphQL Endpoint",
             )
         except Exception:
             self.logger.exception("Exception occurred during the GraphQL post request")
-            abort(500, message="INTERNAL ERROR")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
         try:
             # construct response object
             response = make_response(gql_response.text)
         except Exception:
             self.logger.exception("Exception occurred GraphQL response construction")
-            abort(500, message="INTERNAL ERROR")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
         response.status_code = gql_response.status_code
         return response
--- a/lib/pbench/server/database/models/users.py
+++ b/lib/pbench/server/database/models/users.py
@@ -10,18 +10,6 @@ from sqlalchemy.orm.exc import NoResultFound
 from pbench.server.database.database import Database
 
 
-class AdminUserDeletion(Exception):
-    """
-    Attempt to delete an admin user.
-    """
-
-    def __init__(self, username: str):
-        self.username = username
-
-    def __str__(self):
-        return f"Admin User {self.username} can not be deleted"
-
-
 class Roles(enum.Enum):
     ADMIN = 1
 
--- a/lib/pbench/test/unit/server/test_user_auth.py
+++ b/lib/pbench/test/unit/server/test_user_auth.py
@@ -431,3 +431,29 @@ class TestUserAuthentication:
                 headers=dict(Authorization="Bearer " + data_login["auth_token"]),
             )
             assert response.status_code == 200
+
+    @staticmethod
+    def test_non_existent_user_delete(client, server_config):
+        with client:
+            resp_register = register_user(
+                client,
+                server_config,
+                username="username",
+                firstname="firstname",
+                lastname="lastName",
+                email="user@domain.com",
+                password="12345",
+            )
+            assert resp_register.status_code == 201
+
+            # user login
+            resp_login = login_user(client, server_config, "username", "12345")
+            data_login = resp_login.json
+            assert data_login["auth_token"]
+
+            response = client.delete(
+                f"{server_config.rest_uri}/user/username1",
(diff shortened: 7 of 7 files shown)


lib/pbench/server/api/resources/users_api.py
```diff
@@ -298,8 +299,22 @@ def __init__(self, logger, auth):
         self.logger = logger
         self.auth = auth
 
+    def operation_permitted(self, target_username: str) -> User:
```
dbutenhof, line 302: Something like `user_if_permitted` would provide a better description of what it does.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -298,8 +299,22 @@ def __init__(self, logger, auth):
...
+        Helper function to determine whether the API call is permitted for the target username
+        Right now it is only permited for an admin user and the target user itself
+        :param target_username: str
+        :return: target User object/ None
+        """
+        current_username = self.auth.token_auth.current_user().username
+        current_user = User.query(username=current_username)
+        target_user = User.query(username=target_username)
+        if current_user.is_admin() or current_username == target_username:
+            return target_user
```
dbutenhof, line 313: This check could be inverted to return `None` and moved up before doing the `query` on `target_username`, saving a small amount of time on the failure case.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -324,36 +339,26 @@ def get(self, username):
...
-            abort(
-                403,
-                message=f"Not authorized to access information about user {username}",
-            )
+                abort(
+                    HTTPStatus.FORBIDDEN,
+                    message=f"Not authorized to access information about user {target_username}",
+                )
+        except Exception:
+            self.logger.exception("Exception occurred during verifying the user")
```
dbutenhof, line 357: "during validation" or "while validating the user" ... the former is less specific, but more accurate. "user" is ambiguous as we might have two separate usernames here.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -388,23 +393,23 @@ def put(self, username):
...
             self.logger.warning("Invalid json object: {}", request.url)
-            abort(400, message="Invalid json object in request")
+            abort(HTTPStatus.BAD_REQUEST, message="Invalid json object in request")
 
         try:
-            user, verified = self.auth.verify_user(username)
+            target_user = self.operation_permitted(target_username)
+            if not target_user:
+                self.logger.warning(
+                    "User {} is not authorized to delete user {}.",
```
dbutenhof, line 402: "modify", not "delete".

lib/pbench/server/api/resources/users_api.py
```diff
@@ -456,33 +464,25 @@ def delete(self, username):
...
+                    self.auth.token_auth.current_user().username,
+                    target_username,
+                )
+                abort(HTTPStatus.FORBIDDEN, message="Not authorized to delete user")
+            if target_user.is_admin():
+                self.logger.warning(
+                    "Admin user is not allwed to self delete via API call. Username: {}",
```
dbutenhof, line 477: "allowed" ... and maybe just "the Admin user cannot be deleted via API call". But further... we don't really need the username here until we have a real distinction between those "is admin" vs "has admin" cases we'd discussed earlier; and I suspect there's no point in trying to over-generalize at this time since there'll be a lot of changes through this code when we add roles.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -417,27 +422,30 @@ def put(self, username):
                 "User trying to update fields that are not present in the user database. Fields: {}",
                 non_existent,
             )
-            abort(400, message="Invalid fields in update request payload")
+            abort(
+                HTTPStatus.BAD_REQUEST,
+                message="Invalid fields in update request payload",
+            )
         protected = set(post_data.keys()).intersection(set(User.get_protected()))
```
webbnh, line 429: If the requestor is an admin and the target is the requestor, then I think we should (at least for the interim) add the `role` field to the "protected" set, to prevent the last admin from de-admin'ing him/herself.

webbnh (commented):
Nikhil, this is great stuff!  Why is this PR a "draft"??  Let's get this merged in!!  :-)

Naturally, I have a bunch of suggestions....

lib/pbench/server/api/resources/users_api.py
webbnh: Since this is the log, I think we _do_ want the username.  (We'll want as much context as we can get if we need to investigate this retroactively.)

lib/pbench/server/api/resources/users_api.py
```diff
@@ -226,21 +227,21 @@ def post(self):
             self.logger.warning(
                 "Duplicate auth token got created, user might have tried to re-login immediately"
             )
-            abort(409, message="Retry login after some time")
+            abort(HTTPStatus.CONFLICT, message="Retry login after some time")
```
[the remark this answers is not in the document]
npalaska: That sounds good

lib/pbench/server/api/resources/users_api.py
npalaska: That is very good point @webbnh, Thanks

lib/pbench/server/api/resources/users_api.py
```diff
@@ -324,36 +339,26 @@ def get(self, username):
...
-                403,
-                message=f"Not authorized to access information about user {username}",
-            )
+                abort(
+                    HTTPStatus.FORBIDDEN,
+                    message=f"Not authorized to access information about user {target_username}",
+                )
+        except Exception:
+            self.logger.exception("Exception occurred during verifying the user")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
```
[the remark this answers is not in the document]
npalaska: if a logged-in user makes a GET request for a non-existent user, it is a `FORBIDDEN` since `get_valid_target_user` method would return `None` in this case because it couldn't find the target user. It's an interesting idea that an admin_user should get a `NOT_FOUND` instead of forbidden

lib/pbench/server/api/resources/users_api.py
webbnh: I believe that there are actually several places in the code where we pull the current user from the auth token and then use it to query the database to get some attribute of the user -- they should probably all be "improved".

Speaking of which, I have a (possibly) better code suggestion:
```
        current_user = self.auth.token_auth.current_user()
        if current_user.username == target_username:
            return current_user
        if current_user.is_admin(): 
            return User.query(username=target_username)
        return None
```
It's slightly more complex (we're back to six lines...), but it avoids a DB query when we already know the answer (and at worst it's the same number of operations).

lib/pbench/server/api/resources/users_api.py
```diff
@@ -226,21 +227,21 @@ def post(self):
             self.logger.warning(
                 "Duplicate auth token got created, user might have tried to re-login immediately"
```
portante, line 228: Why are we able to create a duplicate auth token?

portante: Let's track this in #2333.

lib/pbench/server/api/resources/users_api.py
npalaska: Okay this means we _can_ delete an admin user via API if the requester user itself is an admin until, do we want to allow that or restrict admin deletion operation via CLI only, since an admin user should have an access to the server.

dbutenhof: Right now we have no provision for the concept of "an" admin user; we only have username "admin" (and, come to think of it, we need to decide what that looks like if we adopt an "email is username" model). I'd be inclined to say, at least for now, that only the CLI can delete "admin" (whatever it's called) but anyone else can potentially be edited/deleted in the API regardless of the user's role.

webbnh: @dbutenhof, I'm pretty sure that Nikhil has implemented "admin-ness" as a **role**.  Once that's merged, we can have more than one admin user, and they don't have to have a particular name/e-mail.

If you guys are "sure", then I'm willing to accept (at least for the moment) the requirement that an API `DELETE` request is not allowed to delete _any_ user with the "admin" role.  (I concur that the CLI should be allowed to do basically _anything_.)  However, I think it's adequate to allow deletions of _other_ admin users via the API -- we just shouldn't allow an admin to delete itself.

Nevertheless, pursuant to this conversation here, the question is whether an admin should be allowed to remove the "admin" role from itself.  And, I think the answer to that (given the present discussion) is a clear "no".  (Which leaves the question of whether an admin should be allowed to remove the admin role from another user, and I'm gonna wait on asking/answering that question until I hear what the consensus is on whether admin users can delete other admin users....)

dbutenhof: > Nevertheless, pursuant to this conversation here, the question is whether an admin should be allowed to remove the "admin" role from itself. And, I think the answer to that (given the present discussion) is a clear "no". (Which leaves the question of whether an admin should be allowed to remove the admin role from another user, and I'm gonna wait on asking/answering that question until I hear what the consensus is on whether admin users can delete other admin users....)

I'm not convinced of that. I think we need a standard "admin/admin" user (although I have no idea what to call if it we're using email for usernames) to jumpstart the process. However this can only be created with the CLI, and it can be recreated, so I'm not certain it needs any particular protections. I think an ADMIN (role) user can add or remove the ADMIN role from any user; they have ADMIN because they're trusted. If they delete ADMIN from the last user to have it, they go to the CLI and create a new user with ADMIN, or add it to an existing user; no big deal.

webbnh: I'm inclined to turn your logic on itself:  if someone can just go to the CLI, then they can modify or delete any user they want, and so it doesn't matter that we impose limits on what they can do through the API.  Moreover, we don't need to have a standard "admin/admin" user (or worry about what to call it), because they can just go to the CLI and create it any way they want it.

Given that, I'm inclined to make the API "safe", so that it's hard to make big mistakes using it (and so that it's harder for hostile actors to mess things up).  I'm imagining deployments where someone with an admin role does _not_ necessarily have access to the CLI, in which case having some "blade guards" in place would be a good thing, especially since the API might be accessible from anywhere on the Internet.

dbutenhof: > I'm inclined to turn your logic on itself: if someone can just go to the CLI, then they can modify or delete any user they want, and so it doesn't matter that we impose limits on what they can do through the API. Moreover, we don't need to have a standard "admin/admin" user (or worry about what to call it), because they can just go to the CLI and create it any way they want it.

They _can_, but I think our install should provide a standard out-of-the-box admin user they can easily use for startup configuration. They can change the password, or even delete it once they've assigned ADMIN role to another user, but many products do this and it provides at very least a standard template for describing configuration. ("Log in as admin/admin and ...")

lib/pbench/server/api/resources/users_api.py
```diff
@@ -340,26 +340,35 @@ def get(self, target_username):
...
+                "User {} performed get operation but user {} is not found.",
+                self.auth.token_auth.current_user().username,
+                target_username,
+            )
+            abort(
+                HTTPStatus.NOT_FOUND, message=f"User {target_username} not found",
+            )
+        else:
+            self.logger.warning(
+                "User {} is not authorized to get user {}.",
```
dbutenhof, line 364: I'm not crazy about this message. We're not technically confirming that the user name exists, but the message implies that it does, and I'd prefer we phrase this differently. I also don't think it's necessary to identify the user encoded in the authorization token. (E.g., if they somehow stole a valid auth token, let's not tell them the username.)

lib/pbench/server/api/resources/users_api.py
```diff
@@ -473,27 +491,42 @@ def delete(self, target_username):
...
-                abort(HTTPStatus.FORBIDDEN, message="Not authorized to delete user")
+            self.logger.warning(
+                "User {} is not authorized to delete user {}.",
+                self.auth.token_auth.current_user().username,
+                target_username,
+            )
+            abort(HTTPStatus.FORBIDDEN, message="Not authorized to delete user")
+        if target_user.is_admin():
+            self.logger.warning(
+                "Admin user is not allwed to self delete via API call. Username: {}",
```
dbutenhof, line 519: "Admin user" implies "admin" account; this is more about "User with administrative access" or we could say "User with Admin role". Also, "allowed" is missing a letter. Granted, this is an internal log message, but we won't be the only ones to read system logs and we should be a little careful both about context and security. (Some customers, for example, might not want to give Red Hat their system logs for debugging issues if the logs contain internal information like hostnames or usernames.)

lib/pbench/server/api/resources/users_api.py
npalaska: But this is a log message not the abort message so I think its okay

lib/pbench/server/api/resources/users_api.py
webbnh: I'm wondering about the usability of this.  If our standard install provides a standard admin user with a well-known password, then sorta the first order of business on deployment would be to go in and change the admin user password...so why not just go with the idea that, if the person doing the deployment wants to have an admin user, they have to go manually create it -- that's pretty much the same level of effort as changing the password for it.

I'm thinking that in the overwhelming number of use-cases, an admin user isn't even required...so why create one by default?  Moreover, if we delegate that to the person doing the deployment, then we finesse all the issues of what the admin username should be, what its e-mail should be, and what it's password should be -- those are all details which the system administrator would have, so we don't need to come up with them!

lib/pbench/server/api/resources/users_api.py
webbnh: "User with ADMIN role" would certainly be clearer.  I would omit the "via API call":  that will be self-evident to the recipient of the message, and any apparent discrepancy between API policy and CLI policy is unlikely to be evident to anyone but us (and possibly the people administering the deployment).

> we won't be the only ones to read system logs and we should be a little careful both about context and security. (Some customers, for example, might not want to give Red Hat their system logs for debugging issues if the logs contain internal information like hostnames or usernames.)

This is an unsettling conundrum for me.

As a developer, I will want to know details about the context of an error or failure.  The username associated with the request and the username associated with the target of the request seem like key pieces of information.  On the server side, we will be the only ones reading the log, at least until a partner deploys their own server.  If that happens, and if they are reluctant to share their usernames (which seems...overly cautious, to me), the log is just a text file which is easily cleansed with a `sed` script...so, I would be inclined to suggest that logging the information provides much more value than it imposes in risk or cost.

Still, Dave, your point is well-taken.  (I still remember when VMS was logging failed login attempts to the system console, including the contents of the password field, which, in the days of paper logs, meant that your typos were there for anyone to see as they walked through the lab....)  But, while we shouldn't be logging security credentials, I'd like to believe that the username is an identifier and not, as such, a credential.

lib/pbench/server/api/resources/users_api.py
webbnh: I'm with Nikhil:  since this goes in the server log and not back to the user, I don't think we're leaking any secrets.  (Anyone with direct access to the log also, presumably, has access to the user database itself....)  That said, I'm not opposed to rephrasing the message to make it more accurate, but the scenario here is that a non-admin user is requesting information about someone else (who does exist), and the message seems like it captures that.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -339,23 +339,36 @@ def get(self, target_username):
...
+                self.auth.token_auth.current_user().username,
+                target_username,
+            )
             abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
+        if target_user:
+            response_object = target_user.get_json()
+            return make_response(jsonify(response_object), HTTPStatus.OK)
+        elif self.auth.token_auth.current_user().is_admin():
+            self.logger.warning(
+                "User {} performed get operation but user {} is not found.",
```
webbnh, line 355: Rather than "performed", "requested" would be better.  (Ditto at lines 423 and 504.)

lib/pbench/server/api/resources/users_api.py
```diff
@@ -390,31 +403,44 @@ def put(self, target_username):
...
-            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
+            self.logger.warning(
+                "User {} is not authorized to modify user {}.",
+                self.auth.token_auth.current_user().username,
+                target_username,
+            )
+            abort(
+                HTTPStatus.FORBIDDEN,
+                message=f"Not authorized to update information about user {target_username}",
+            )
```
webbnh: I think we have three copies of this code which are all the same except for the request type (`GET`, `PUT`, `DELETE`).  It would be nice to have them using common code, but I feel like we shouldn't be burying calls to `abort()` inside subroutines.  Does anyone else have opinions on this?

I suppose we could have a helper routine which returns a target user or a status/message pair (the details of the mechanism for that are TBD) so that this function could receive either the information that it wants or the arguments needed for the appropriate `abort()` call, but perhaps that's overdoing it.  Thoughts?

webbnh (changes requested):
Looking good.

By my reckoning, these items are still outstanding:
 - Should we permit users with the Admin role to be deleted (i.e., by other users with the Admin role) via the API?  (I believe that we agree that they should not be allowed to delete themselves...should we just apply a blanket prohibition against deleting users with the Admin role via the API?)
 - Should we permit users with the Admin role to remove the Admin role from themselves via the API?  (I believe that we should not, for the same reason as prohibiting them from deleting themselves via the API.)
 - Is this the PR in which we are going to replace the the unit tests' dependencies on the database with mocks?  (Or, are we going to defer that again to a follow-on?)

(The items below are small-fry.)

lib/pbench/server/api/resources/users_api.py
npalaska: > Some customers, for example, might not want to give Red Hat their system logs for debugging issues if the logs contain internal information like hostnames or usernames.

This is a concern that needs to be addressed separately and carefully, I think by merely changing the log message may not be sufficient to address this point. This warrants a separate discussion and a PR.

dbutenhof: Customers are often paranoid about internal identification, and I suspect this is something we'll need to deal with. It's easier to fix "from the beginning" than to retrofit later. It's also true that, especially in our initial internal deployments, we'll have a much easier time with too much information than too little; but we should be thinking about this carefully, not just throwing things out because we can.

lib/pbench/server/api/resources/users_api.py
npalaska: Good point, helper function `get_valid_target_user` could return a target_user along with the status code that can be used to proceed with a requested operation, unless some other exception occurs in which case we would be returning 50X status

npalaska:
> * Should we permit users with the Admin role to be deleted (i.e., by other users with the Admin role) via the API?  (I believe that we agree that they should not be allowed to delete themselves...should we just apply a blanket prohibition against deleting users with the Admin role via the API?)

I would vote on the latter to just not allow any admin user deletion via API, the admin users should technically have an access to the server where they can perform the delete operation. The reason I am thinking this is because what if the admin login token gets stolen or misused or by mistake someone with admin token performs the unintended request. 

> * Should we permit users with the Admin role to remove the Admin role from themselves via the API?  (I believe that we should not, for the same reason as prohibiting them from deleting themselves via the API.)

I agree we should not permit this via API call, when we have other roles defined we could allow users to change their roles via API call except the admin role.

> * Is this the PR in which we are going to replace the the unit tests' dependencies on the database with mocks?  (Or, are we going to defer that again to a follow-on?)
> 

I don't know, this will be a major change from how we do the unittests right now. Also we currently do not have a real functional test environment, where we can perform sequences of real requests against a real server. To me personally our unittests (semi-functional unittests) has helped find some edge cases and bugs since they make actual database queries, so I would say we could hold off on that until we have some functional testing environment or address it in a separate dedicated PR.

dbutenhof:
> > * Should we permit users with the Admin role to be deleted (i.e., by other users with the Admin role) via the API?  (I believe that we agree that they should not be allowed to delete themselves...should we just apply a blanket prohibition against deleting users with the Admin role via the API?)

While we're a long way away from this, I would like to supply an installation environment that allows all aspects of user management to be handled from an online admin console within the dashboard; without ever requiring that a "Pbench admin" is able to log in to the server instance.

That's why I would like to have a built-in admin user; once we start up, they can log in through the online API, from any browser, and complete the configuration remotely. They can delete the admin user when they're ready, and recovering from an error here might require local login -- if possible -- or complete re-installation in a new instance, but this shouldn't be required as part of the standard workflow.

I'm not going to tilt too hard against this windmill at this point, but I'm also pretty sure the world will conspire on my side of the argument eventually.

> > * Is this the PR in which we are going to replace the the unit tests' dependencies on the database with mocks?  (Or, are we going to defer that again to a follow-on?)
> 
> I don't know, this will be a major change from how we do the unittests right now. Also we currently do not have a real functional test environment, where we can perform sequences of real requests against a real server. To me personally our unittests (semi-functional unittests) has helped find some edge cases and bugs since they make actual database queries, so I would say we could hold off on that until we have some functional testing environment or address it in a separate dedicated PR.

While I like the idea of mocking everything and building a "pure" unit test environment, I don't want to derail/expand this PR for that purpose. That should be isolated and comprehensive.

We also need a real functional test environment (and regularly scheduled CD pipelines to stand up instances and run test scripts against them, including ideally chaos testing), sooner rather than later. A containerized server deployment will help a lot with that, but we also need to start thinking about the framework for provisioning and testing.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -299,19 +300,43 @@ def __init__(self, logger, auth):
         self.logger = logger
         self.auth = auth
 
-    def get_valid_target_user(self, target_username: str) -> User:
+    def get_valid_target_user(
+        self, target_username: str, rest_call: str
+    ) -> Tuple["User", HTTPStatus, str]:
```
dbutenhof, line 305: Anonymous tuple returns are really hard to follow; a named tuple would be better. That essentially provides a custom object with attributes that can be read by name... but can also be destructured just like a normal tuple. Seems to provide a lot of benefits in use and self-documentation with very little added complexity.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -299,19 +300,43 @@ def __init__(self, logger, auth):
...
         """
         Helper function to determine whether the API call is permitted for the target username
         Right now it is only permited for an admin user and the target user itself
         :param target_username: str
-        :return: target User object/ None
+        :param rest_call: str
```
dbutenhof, line 310: Can we have a better name than `rest_call`? That's very unspecific. Are we talking "endpoint" like `/api/v1/users` or HTTP method like `PUT`, or both? It's an obvious question when reading this code, and the name provides no hint. For that matter, giving the name and type here is pointless, because they're both provided in the method signature. The only reason to list parameters in the docstring is to explain what they mean ... but even with that, a good name is best.

As a side note, it seems this is something we'd find useful beyond the scope of this single class.

lib/pbench/server/api/resources/users_api.py
webbnh: I'm kind of fond of the clean simplicity of the current implementation of `get_valid_target_user()`, so, if you're up for this, @npalaska, rather than enhancing `get_valid_target_user()`, I would suggest implementing a wrapper for it.

lib/pbench/server/api/resources/users_api.py
webbnh: I was envisioning a `Tuple["User", Tuple[HTTPStatus, str]]`, but a [named tuple](https://docs.python.org/3/library/collections.html#collections.namedtuple) would be even better.

webbnh:
> I would vote [...] to just not allow any admin user deletion via API, the admin users should technically have an access to the server where they can perform the delete operation. The reason I am thinking this is because what if the admin login token gets stolen or misused or by mistake someone with admin token performs the unintended request.

I don't think there is any requirement that admin users have access to the CLI on the server.  I concur with Dave that, eventually, suitably privileged users should be able to administer the server (at least its users, data, and operation) remotely via the API.  That said, I think it's reasonable, in the immediate term, to prohibit admin user via the API, but I don't think it's necessary.  (As I've pointed out, as long as we prohibit an admin from deleting itself, I think we're covered.)

Concerns over an admin login token or credentials being hijacked or stolen are part of the landscape -- we need to implement a secure system and endeavor to prevent that -- we can't justify not providing deletion on that basis.  (Deleting other admins is the least of the havoc that an unsuitably privileged access could cause!)

> I agree we should not permit this via API call, when we have other roles defined we could allow users to change their roles via API call except the admin role.

I expect that modifying a user's roles should require privilege, and I think we should prohibit an admin user from removing himself from an admin role (for the same reason that we should prohibit admin users from deleting themselves).

> While we're a long way away from this, I would like to supply an installation environment that allows all aspects of user management to be handled from an online admin console within the dashboard; without ever requiring that a "Pbench admin" is able to log in to the server instance.

I support this desire, except that I think that _someone_ has to log into the server instance to do the initial installation and configuration, and I maintain that that is the point at which the initial admin user should be created, using local knowledge.  But, after that bootstrap operation, I think everything else should be able to be done remotely.  But, I think having a pre-configured privileged user creates a "window of opportunity" for mischief, and I think it's unnecessary, given the idea that someone has to install this monster first.

> I'm also pretty sure the world will conspire on my side of the argument eventually.

:-D

> > Is this the PR in which we are going to replace the the unit tests' dependencies on the database with mocks? (Or, are we going to defer that again to a follow-on?)

> I don't know, this will be a major change

and

> I don't want to derail/expand this PR for that purpose. That should be isolated and comprehensive.

I'm going to take those as "no".  :-)

> this will be a major change from how we do the unittests right now.

I'm not sure that that is true.  Mocking external dependencies is pretty easy in Python, and, once you get the hang of it, it's very powerful.  In many cases, it's actually simpler and easier than actually invoking the real dependency, and it allows the test to gain control and check assertions at points where it otherwise couldn't.

> Also we currently do not have a real functional test environment, where we can perform sequences of real requests against a real server.

True, but this isn't a reason why we shouldn't implement the unit tests properly.  Also, until we have a real functional test environment, there is no reason not to keep the existing "semi-functional" tests.  In fact, we may want to keep them in addition to a real functional test -- having a fully disposable (e.g., in-memory) database is very flexible and freeing and will allow us to do stuff that we probably wouldn't want to do in a more functional environment.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -349,25 +376,10 @@ def get(self, target_username):
...
-        elif self.auth.token_auth.current_user().is_admin():
-            self.logger.warning(
-                "User {} performed get operation but user {} is not found.",
-                self.auth.token_auth.current_user().username,
-                target_username,
-            )
-            abort(
-                HTTPStatus.NOT_FOUND, message=f"User {target_username} not found",
-            )
+            return make_response(jsonify(response_object), http_status)
```
webbnh, line 379: I wouldn't bother using `http_status` here -- it's supposed to be `OK`, right?  So, remove the obfuscation and potential pitfall.

webbnh: Ditto lines 470 and 520.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -349,25 +376,10 @@ def get(self, target_username):
...
         else:
-            self.logger.warning(
-                "User {} is not authorized to get user {}.",
-                self.auth.token_auth.current_user().username,
-                target_username,
-            )
             abort(
-                HTTPStatus.FORBIDDEN,
-                message=f"Not authorized to access information about user {target_username}",
+                http_status, message=http_message,
```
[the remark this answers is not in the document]
webbnh: Ditto line 436.

webbnh (changes requested):
Looks good, but needs some polish.

I think we've resolved that:

- Changes to mock the database accesses will be deferred to another PR.
- I _think_ we've concluded that we should not permit users to change their own roles.

However, I don't think we've yet reached closure on whether we should we permit users with the Admin role to be deleted via the API.

lib/pbench/server/api/resources/users_api.py
webbnh: OK, I see that there is an "economy of scale" to be gained by integrating the change into `get_valid_target_user()`, so nevermind -- it's good as it is now.

dbutenhof:
> > While we're a long way away from this, I would like to supply an installation environment that allows all aspects of user management to be handled from an online admin console within the dashboard; without ever requiring that a "Pbench admin" is able to log in to the server instance.
> 
> I support this desire, except that I think that _someone_ has to log into the server instance to do the initial installation and configuration, and I maintain that that is the point at which the initial admin user should be created, using local knowledge. But, after that bootstrap operation, I think everything else should be able to be done remotely. But, I think having a pre-configured privileged user creates a "window of opportunity" for mischief, and I think it's unnecessary, given the idea that someone has to install this monster first.

_Someone_ has to __install__ the server, which may well just mean a `podman` command somewhere, then giving a host/port name to the "Pbench admin" to take over... someone whose role has no need for login credentials on the server instance unless we artificially force that need. Nor should the IT admin who owns the machine/vm instance need to know anything about administering Pbench.

With that said, we can finesse this a bit with steps like exposing knobs at the installation process (whether ansible inventory variables or podman environment variables); it would definitely be desirable to be able to externally specify a password for the default "admin" user, and once we do that there's no particular reason we can't allow them to define the _name_ of the default admin user as well. Then it becomes a minor part of the installation process without requiring any Pbench-specific operations like manually logging in and running `pbench-user-create`.

> True, but this isn't a reason why we shouldn't implement the unit tests properly. Also, until we have a real functional test environment, there is no reason not to keep the existing "semi-functional" tests. In fact, we may want to keep them in addition to a real functional test -- having a fully disposable (e.g., in-memory) database is very flexible and freeing and will allow us to do stuff that we probably wouldn't want to do in a more functional environment.

We can do "anything" with pure unit tests and proper mocking. We should _try_ "everything" with real functional tests, because users will in the real world and we need to assume they're all evil idiots and try to protect against them. Maybe we'll find a use for the "halfway" __functional unit testing__ environment with a local DB, but right now what we're doing is just a crutch to compensate for the lack of real mocks, so I'm not immediately seeing a need.

webbnh:
> we can finesse this a bit with steps like exposing knobs at the installation process (whether ansible inventory variables or podman environment variables); it would definitely be desirable to be able to externally specify a password for the default "admin" user, and once we do that there's no particular reason we can't allow them to define the name of the default admin user as well.

Sounds excellent.  (With an approach like that, perhaps the world _will_ join your side of the argument.  :-) )

lib/pbench/server/api/resources/users_api.py
npalaska: Well if we come at this line the http_status is going to be `OK` but I can see the confusion.

npalaska:
Okay most of the comments are addressed except whether we should permit users with the `Admin` role to be deleted via the API call (only if the requester is an Admin as well). I am okay either way as long as we prevent admin _self_ deletion. Does anybody has a different suggestion?

lib/pbench/server/api/resources/users_api.py
webbnh: Yes, `http_status` _should_ be `OK` at this point...but, if you're reading the code you have to go kind of a long way to prove that to yourself.  And, since `OK` is what we _want_ to return here, there's no reason to create an opportunity for something else to happen by using a variable instead of a "constant".

lib/pbench/server/api/resources/users_api.py
```diff
@@ -326,36 +370,24 @@ def get(self, username):
...
-            except Exception:
-                self.logger.exception(
-                    "Exception occurred while querying the user. Username: {}", username
-                )
-                abort(500, message="INTERNAL ERROR")
-        else:
-            self.logger.warning(
-                "User {} is not authorized to get user {}.", user.username, username,
+            target_user, http_status, http_message = self.get_valid_target_user(
+                target_username, "GET"
```
webbnh, line 374: While this is, again, syntactically correct, doing it this way removes all the advantages of using a named tuple.  Instead, I think it would be better to do something like this:
```
            result = self.get_valid_target_user(target_username, "GET")
[...]
        if result.user is None:
            abort(result.http_status, message=result.http_message)
```

lib/pbench/server/api/resources/users_api.py
```diff
@@ -326,36 +370,24 @@ def get(self, username):
...
-            abort(
-                403,
-                message=f"Not authorized to access information about user {username}",
+        except Exception:
+            self.logger.exception(
+                "Exception occurred in GET request from user {!r} while validating target user {!r}",
+                self.auth.token_auth.current_user().username,
+                target_username,
             )
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
```
webbnh, line 382: Do we really need/want this `try`/`except` block here?  I don't think we expect `get_valid_target_user()` to raise or propagate any exceptions; so, if one happens, it's probably OK to just let it blow by to the catch-all handler.

lib/pbench/server/api/resources/users_api.py
webbnh: Ditto in PUT and DELETE.

lib/pbench/server/api/resources/users_api.py
webbnh: (Ditto in PUT and DELETE.)

dbutenhof:
> Okay most of the comments are addressed except whether we should permit users with the `Admin` role to be deleted via the API call (only if the requester is an Admin as well). I am okay either way as long as we prevent admin _self_ deletion. Does anybody has a different suggestion?

I'm not going to consider this a blocking issue for this PR, I guess; but I don't think we should be placing such restrictions on the API. Sure, if they mess up, they may have to find someone with CLI access to recreate an admin user or add ADMIN role somewhere, but there's always that fallback.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -387,59 +419,60 @@ def put(self, username):
...
                 "User trying to update fields that are not present in the user database. Fields: {}",
                 non_existent,
             )
-            abort(400, message="Invalid fields in update request payload")
-        protected = set(post_data.keys()).intersection(set(User.get_protected()))
+            abort(
+                HTTPStatus.BAD_REQUEST,
+                message="Invalid fields in update request payload",
+            )
+        protected = set(user_payload.keys()).intersection(set(User.get_protected()))
```
webbnh, line 456: Unless the current user `is_admin()` **and** the target user is different from the current user, then I think `role` should be added to the list of protected fields.  (That is, an admin user can change _another_ user's role; otherwise, users cannot change their own roles or the roles of others.)

webbnh: And, I think that this is the appropriate place in the code to implement this policy.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -326,36 +370,24 @@ def get(self, username):
...
+                "Exception occurred in GET request from user {!r} while validating target user {!r}",
+                self.auth.token_auth.current_user().username,
+                target_username,
             )
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
+        if target_user:
+            response_object = target_user.get_json()
+            return make_response(jsonify(response_object), HTTPStatus.OK)
+        else:
+            abort(http_status, message=http_message)
```
webbnh, line 387: This is a nit, but....  The other two HTTP methods do a negative check with an abort and otherwise perform the request...it would be nice to match that structure here:
```
        if not target_user:
            abort(http_status, message=http_message)

        response_object = target_user.get_json()
        return make_response(jsonify(response_object), HTTPStatus.OK)
```

lib/pbench/server/api/resources/users_api.py
```diff
@@ -300,8 +300,52 @@ def __init__(self, logger, auth):
...
+        Helper function to determine whether the API call is permitted for the target username
+        Right now it is only permited for an admin user and the target user itself
+        :param target_username: str
+        :param request_type: str
+        :return: NamedTuple containing target User object/ None, http_status_to_return, http_message_to_return
+        """
+        TargetUser = NamedTuple(
+            "TargetUser",
+            [("user", User), ("http_status", HTTPStatus), ("http_message", str)],
+        )
```
dbutenhof, line 316: Hmm; can you really define a named tuple from the `typing` module? I'm not sure that's a good idea, in any case.

Also, this should be defined at module level (or at least at class level) so it's accessible to callers. Plus the return type hint should be `TargetUser` rather than just a generic `NamedTuple`, which doesn't say much.

You can also define a named tuple with defaults for right-most fields; so, for example, you could default to `(HTTPStatus.OK, "")` and most of your returns would become `return TargetUser(user)`. Although it might be easier for readers to keep track of if you specified all three by keyword, e.g. `return TargetUser(user=name, http_status=HTTPStatus.OK, str="")` (should that be `message` instead of `str`?)

webbnh (changes requested):
`get_valid_target_user()` is now returning a named tuple, which is good, but the code is using it like a normal tuple...if we're going to bother with a named tuple, then we should take advantage of the support that it provides; otherwise, there's no reason to use it instead of a normal tuple.

Other issues:
 - Liberally interpreting [Dave's comment](https://github.com/distributed-system-analysis/pbench/pull/2329#issuecomment-849011004), I think we can move forward to allow admins to delete anybody except themselves.
 - Springboarding on that, I think we should prohibit anyone from modifying their own role (at least for now), but we should allow admins to modify any _other_ user's role.

lib/pbench/server/api/resources/users_api.py
webbnh: Yes, you can do this, although the documentation for it seems to be thin on the ground.  The principal advantage of this seems to be that it allows one to specify the types of the named members.

And, yes, I too thought it might be nice to set up default values for the (right-most) members, but I couldn't figure out how to get `NamedTuple` to accept the default values.  (Having the HTTP stuff disappear in the success case would actually be a feature; having the user disappear in the failure case would not necessarily be bad, either.)

> should that be message instead of str?

Dave, I think you're reading that wrong -- I think the field names are the first elements, and their types are the second elements, so `http_message` is the name of the third member and its type is `str`.

lib/pbench/server/api/resources/users_api.py
npalaska: Well there is actually a potential line of code in `get_valid_target_user` that can cause an exception where we query the target_user against a real database so I guess it's probably okay to have this catch block

webbnh: What's the line and potential exception?

npalaska: We query the target user as `target_user = User.query(username=target_username)` if the current user is an admin on line 321

lib/pbench/server/api/resources/users_api.py
npalaska: Yeah I used `typing` module because we can specify the types of the named members

lib/pbench/server/api/resources/users_api.py
```diff
@@ -296,40 +296,48 @@ class UserAPI(Resource):
...
+    )
+
     def __init__(self, logger, auth):
         self.logger = logger
         self.auth = auth
 
     def get_valid_target_user(
         self, target_username: str, request_type: str
-    ) -> NamedTuple:
+    ) -> "TargetUser":
```
dbutenhof, line 311: That's interesting. Did you have problems without the quotes? I would have thought the `TargetUser` definition would have been fully resolved by this point.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -296,40 +296,48 @@ class UserAPI(Resource):
...
         """
         Helper function to determine whether the API call is permitted for the target username
         Right now it is only permited for an admin user and the target user itself
         :param target_username: str
         :param request_type: str
-        :return: NamedTuple containing target User object/ None, http_status_to_return, http_message_to_return
+        :return: NamedTuple containing target User object/None, http_status_to_return, http_message_to_return
```
dbutenhof: Again, referencing the _types_ for parameters adds nothing, because we already have type hints. What we want here is a description of what each parameter means.

I also find this colon-delimited keyword style about the _least_ readable of the common docstring styles. I don't particularly like PEP's `name -- description` or Google's redundant type, either... but we really need to try to come up with a common style that's readable. However, for right now, I'm objecting only because you've included nothing that's not redundant with the type-hinted method declaration; you might as well have omitted the parameters entirely from the docstring.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -296,40 +296,48 @@ class UserAPI(Resource):
     Abstracted pbench API to get user data
     """
 
+    # Helper NamedTuple containing a target user, expected http status to return with message
```
dbutenhof, line 299: The word "Helper" is both redundant and perhaps a bit misleading -- the named tuple doesn't "help", it's by nature a passive container. (It's really a class that instantiates objects with the ability to act as a tuple as well as an object instance with named attributes... which is a pretty cool idea, but still just a "fancy C struct".) Also not really much point in repeating the base type `NamedTuple` in the comment.

Maybe "Fancy C struct bundling target user, suggested HTTP result status and message" would be a bit blunt, though. ;-)

lib/pbench/server/api/resources/users_api.py
webbnh: Under what circumstances would that result in an exception?  In particular, what's the likelihood of that exception occurring, and what are the costs or what do we lose if we handle it generically using the last-chance handler instead of handling it here?

This code would be simpler and much easier to read without this `try` block.  If the block isn't likely to produce sufficient value, we should remove it, since it does come with a cost.

lib/pbench/server/api/resources/users_api.py
webbnh: I'm wondering if we can get by without a comment:  ideally, code should be self-documenting -- in this case, we're talking about a single definition (i.e., it's almost just a single line) -- and the code says, "`TargetUser` is a `NamedTuple` which contains three members, `target_user` (which is a `User`), `http_status` (which is an `HTTPStatus`), and `http_message` (which is a `str`)".

Is there really anything that we need to add to that, here?

lib/pbench/server/api/resources/users_api.py
webbnh: Right...the value to add here via the comment is "Returns a target User on success or None on failure; in the case of failure, also returns the corresponding HTTP status code and corresponding message string" -- how these are packaged is obvious from the type hint.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -453,22 +456,32 @@ def put(self, target_username):
...
-        protected = set(user_payload.keys()).intersection(set(User.get_protected()))
+        # Only admin user will be allowed to change other user's role. However,
+        # Admin users will not be able to change their admin role,
+        # This is done to prevent last admin user from de-admining him/herself
+        protected_db_fields = (
+            User.get_protected()
+            if self.auth.token_auth.current_user().is_admin()
+            and self.auth.token_auth.current_user() != result.target_user
+            else User.get_protected() + ["role"]
+        )
```
webbnh, line 467: While I do love the "ternary" operator, I found this code hard to check.  I would suggest something more like the following:
```
        protected_db_fields = User.get_protected()
        if (not self.auth.token_auth.current_user().is_admin() or
            self.auth.token_auth.current_user() == result.target_user):
            protected_db_fields.append("role"])
```
(But don't bother unless you have other changes to make.)

webbnh (approved):
Looks great!  From my perspective, we're down to just small stuff now, nothing which gates the merge for me.

lib/pbench/server/api/resources/users_api.py
npalaska: I think this one is addressed, okay to resolve?

dbutenhof: In the context of this PR, yeah, I think it's resolved.

lib/pbench/server/api/resources/users_api.py
npalaska: I agree the likelihood is less but it could result in exception if the db instance is down or if we are in the middle of some migrations. By the way are you suggesting the following 
```
result = self.get_valid_target_user(target_username, "GET")
if not result.target_user:
    abort(result.http_status, message=result.http_message)
response_object = result.target_user.get_json()
return make_response(jsonify(response_object), HTTPStatus.OK)
```
or this
```
try:
  result = self.get_valid_target_user(target_username, "GET")
  if not result.target_user:
      abort(result.http_status, message=result.http_message)
  response_object = result.target_user.get_json()
  return make_response(jsonify(response_object), HTTPStatus.OK)
except Exception:
    self.logger.exception(
        "Exception occurred in GET request from user {!r} while validating target user {!r}",
        self.auth.token_auth.current_user().username,
        target_username,
    )
```

lib/pbench/server/api/resources/users_api.py
npalaska: I agree the Named tuple is self explanatory and doesn't need a comment to re-describe it

lib/pbench/server/api/resources/users_api.py
npalaska: Agree, I will remove the redundant doc string params

lib/pbench/server/api/resources/users_api.py
webbnh: I'm inclined to suggest the first code blob -- the one with no `try` block.

If there is a database error, then the user will receive an `INTERNAL_SERVER_ERROR` from the catch-all exception handler in either case, so we don't need the `try` block for that.  I'm fairly sure that the log will receive the full traceback, so we'll know exactly what happened.  The only thing that we'll miss out on is the two usernames, which, if it's a problem with accessing the database itself, presumably don't matter.

So, it seems like we're giving up nothing and getting simpler, easier-to-read code in return.

npalaska: Which catch-all exception handler are you pointing to if there is no try catch block?

webbnh: The one in the caller of `UserAPI.get()` -- Flask!  :-D

npalaska: Ah okay

webbnh (dismissed):

lib/pbench/server/api/resources/users_api.py
npalaska: Flake8 is complaining that it found an undefined name 'TargetUser' if I dont use the quote

webbnh: Presumably you need to qualify it:  `UserAPI.TargetUser`.

npalaska: weird I get `undefined name 'UserAPI'` now

dbutenhof: Ah, of course; because we _are_ still defining the class `UserAPI`, so that doesn't resolve yet. Sigh.

So maybe it should be `"UserAPI.TargetUser"??

lib/pbench/server/api/resources/users_api.py
```diff
@@ -308,13 +307,12 @@ def __init__(self, logger, auth):
 
     def get_valid_target_user(
         self, target_username: str, request_type: str
-    ) -> "TargetUser":
+    ) -> TargetUser:
```
dbutenhof: You said flake8 complained about this; but does that mean you didn't run flake8 locally before pushing the commit, or was your local flake8 happy but the Jenkins container flake8 complained?

lib/pbench/server/api/resources/users_api.py
```diff
@@ -308,13 +307,12 @@ def __init__(self, logger, auth):
...
         """
         Helper function to determine whether the API call is permitted for the target username
-        Right now it is only permited for an admin user and the target user itself
-        :param target_username: str
-        :param request_type: str
-        :return: NamedTuple containing target User object/None, http_status_to_return, http_message_to_return
+        Right now it is only permited for an admin user and the target user itself.
```
dbutenhof, line 313: "permitted"

lib/pbench/server/api/resources/users_api.py
```diff
@@ -308,13 +307,12 @@ def __init__(self, logger, auth):
...
+        This returns a target User on success or None on failure; in the case of failure,
+        also returns the corresponding HTTP status code and corresponding message string
```
dbutenhof, line 315: You could streamline this a bit to "returns the corresponding HTTP status code and message string".

dbutenhof (approved):
A few suggestions, but I don't think they're critical.

lib/pbench/server/api/resources/users_api.py
npalaska: yeah I didn't think it would be a flake8 issue and pushed the changes without running locally, just to realize it actually is an issue.

dbutenhof: Yeah, I've done the same thing with a "trivial" last-minute tweak. I was just checking whether we had a flakey flake8! ;-)

dbutenhof (approved):

npalaska:
Rebasing the commits now since I got 2 approvals

dbutenhof (approved):
image

webbnh (approved):

PR #2330, opened 2021-05-21
More query_api unit test cleanup and refactoring

dbutenhof:
Webb pointed out a few things about the query unit tests that I decided I wanted to clean up.

- I realized that the clever use of `@pytest.mark.parametrize()` to pass parameters to a fixture was unnecessary, as the fixture is returning a local method that does the heavy lifting anyway.
- By directly raising exceptions that would normally be raised through the requests package `raise_for_status()` method, we're excessively interfering with normal behavior. However, connection errors _are_ raised directly by requests, so I've split off the `HTTPError` exception behavior (`raise_for_status`) from the testing of other exceptions.
- I decided to switch to the `responses` package instead of `requests_mock`, which streamlines some code.

### DIFF
--- a/lib/pbench/test/unit/server/query_apis/conftest.py
+++ b/lib/pbench/test/unit/server/query_apis/conftest.py
@@ -1,28 +1,14 @@
+from typing import Any, Dict
 import pytest
 import requests
-from requests.exceptions import HTTPError
-
-
-def make_http_exception(status: int) -> HTTPError:
-    """
-    Helper to create a properly annotated HTTPError for testing
-
-    Args:
-        status: HTTP status code
-
-    Returns:
-        HTTPError object
-    """
-    response = requests.Response()
-    response.status_code = status
-    response.reason = "Fake reason"
-    return HTTPError(response=response)
+import responses
 
 
 @pytest.fixture
-def query_api(client, server_config, requests_mock, request):
+@responses.activate
+def query_api(client, server_config):
     """
-    query_helper Help controller queries that want to interact with a mocked
+    Help controller queries that want to interact with a mocked
     Elasticsearch service.
 
     This is a fixture which exposes a function of the same name that can be
@@ -31,41 +17,29 @@ def query_api(client, server_config, requests_mock, request):
 
     Parameters to the mocked Elasticsearch POST are passed as keyword
     parameters: these can be any of the parameters supported by the
-    request_mock post method. The most common are 'json' for the JSON
-    response payload, and 'exc' to throw an exception.
-
-    NOTE: the expected URI for the server query and the resulting Elasticsearch
-    query come from a test case parameter, which is passed through the Pytest
-    "mark" mechanism and accessed through the "param" attribute of the Pytest
-    request fixture. E.g.,
-
-            @pytest.mark.parametrize(
-                "query_api",
-                [
-                    {
-                        "pbench": "/controllers/list",
-                        "elastic": "/_search?ignore_unavailable=true",
-                    }
-                ],
-                indirect=True,
-            )
-
-    The first parameter must match the name of this fixture, and the second should
-    be a list (or tuple) with a single dict defining the "pbench" and "elastic"
-    URI paths.
+    responses mock. Exceptions are specified by providing an Exception()
+    instance as the 'body'.
 
     :return: the response object for further checking
     """
 
-    def query_api(payload, expected_index, expected_status, server_config, **kwargs):
+    def query_api(
+        pbench_uri: str,
+        es_uri: str,
+        payload: Dict[str, Any],
+        expected_index: str,
+        expected_status: str,
+        **kwargs,
+    ) -> requests.Response:
         host = server_config.get("elasticsearch", "host")
         port = server_config.get("elasticsearch", "port")
-        es_url = f"http://{host}:{port}{expected_index}{request.param['elastic']}"
-        requests_mock.post(es_url, **kwargs)
-        response = client.post(
-            f"{server_config.rest_uri}{request.param['pbench']}", json=payload
-        )
-        assert requests_mock.last_request.url == es_url
+        es_url = f"http://{host}:{port}{expected_index}{es_uri}"
+        with responses.RequestsMock() as rsp:
+            rsp.add(responses.POST, es_url, **kwargs)
+            response = client.post(
+                f"{server_config.rest_uri}{pbench_uri}", json=payload
+            )
+            assert rsp.assert_call_count(es_url, 1) is True
         assert response.status_code == expected_status
         return response
 
--- a/lib/pbench/test/unit/server/query_apis/conftest.py
+++ b/lib/pbench/test/unit/server/query_apis/conftest.py
@@ -39,7 +39,7 @@ def query_api(client, server_config):
             response = client.post(
                 f"{server_config.rest_uri}{pbench_uri}", json=payload
             )
-            assert rsp.assert_call_count(es_url, 1) is True
+            assert len(rsp.calls) == 1
         assert response.status_code == expected_status
         return response
 
--- a/lib/pbench/test/unit/server/query_apis/test_controllers_list.py
+++ b/lib/pbench/test/unit/server/query_apis/test_controllers_list.py
@@ -1,7 +1,6 @@
 import pytest
 import requests
 from http import HTTPStatus
-from pbench.test.unit.server.query_apis.conftest import make_http_exception
 
 
 class TestControllersList:
@@ -76,17 +75,7 @@ class TestControllersList:
             == "Value '2020-19' (str) cannot be parsed as a date/time string"
         )
 
-    @pytest.mark.parametrize(
-        "query_api",
-        [
-            {
-                "pbench": "/controllers/list",
-                "elastic": "/_search?ignore_unavailable=true",
-            }
-        ],
-        indirect=True,
-    )
-    def test_query(self, client, server_config, query_api, user_ok, find_template):
+    def test_query(self, server_config, query_api, user_ok, find_template):
         """
         test_query Check the construction of Elasticsearch query URI
         and filtering of the response body.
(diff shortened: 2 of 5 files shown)


lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
```diff
@@ -265,43 +259,38 @@ def test_nonunique_query(
...
+                "exception": requests.exceptions.Timeout(),
                 "status": HTTPStatus.GATEWAY_TIMEOUT,
             },
             {
-                "exception": requests.exceptions.InvalidURL,
+                "exception": requests.exceptions.InvalidURL(),
                 "status": HTTPStatus.INTERNAL_SERVER_ERROR,
             },
-            {"exception": Exception, "status": HTTPStatus.INTERNAL_SERVER_ERROR},
+            {"exception": Exception(), "status": HTTPStatus.INTERNAL_SERVER_ERROR},
```
webbnh, line 286: Wouldn't it make for a better test if you picked a concrete exception or a subclass of `Exception` which would then be caught by a match to `Exception`?

lib/pbench/test/unit/server/query_apis/test_datasets_list.py
```diff
@@ -158,36 +162,26 @@ def test_query(self, client, server_config, query_api, user_ok, find_template):
...
+                "exception": requests.exceptions.Timeout(),
                 "status": HTTPStatus.GATEWAY_TIMEOUT,
             },
             {
-                "exception": requests.exceptions.InvalidURL,
+                "exception": requests.exceptions.InvalidURL(),
                 "status": HTTPStatus.INTERNAL_SERVER_ERROR,
             },
-            {"exception": Exception, "status": HTTPStatus.INTERNAL_SERVER_ERROR},
+            {"exception": Exception(), "status": HTTPStatus.INTERNAL_SERVER_ERROR},
```
webbnh, line 176: Wouldn't it make for a better test if you picked a concrete exception or a subclass of `Exception` which would then be caught by a match to `Exception`?

lib/pbench/test/unit/server/query_apis/test_datasets_list.py
```diff
@@ -196,9 +190,32 @@ def test_http_error(
...
+            "/_search?ignore_unavailable=true",
             json,
             index,
             exceptions["status"],
-            server_config,
-            exc=exceptions["exception"],
+            body=exceptions["exception"],
+        )
+
+    @pytest.mark.parametrize("errors", (400, 500, 409))
```
webbnh, line 201: While it's true (at the moment) that, for each of these errors, the Pbench server response is the same (`BAD_GATEWAY`), this seems likely (to me...) to change...so, wouldn't it be good to have each parameter be a pair with the ES response status and the expected Pbench response status?

(Ditto for the other test(s) with this structure.)

lib/pbench/test/unit/server/query_apis/conftest.py
```diff
@@ -31,41 +17,29 @@ def query_api(client, server_config, requests_mock, request):
...
-            f"{server_config.rest_uri}{request.param['pbench']}", json=payload
-        )
-        assert requests_mock.last_request.url == es_url
+        es_url = f"http://{host}:{port}{expected_index}{es_uri}"
+        with responses.RequestsMock() as rsp:
+            rsp.add(responses.POST, es_url, **kwargs)
+            response = client.post(
+                f"{server_config.rest_uri}{pbench_uri}", json=payload
+            )
+            assert rsp.assert_call_count(es_url, 1) is True
```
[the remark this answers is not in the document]
webbnh: Ditto in `lib/pbench/test/unit/server/query_apis/test_month_indices.py`.

lib/pbench/test/unit/server/query_apis/test_month_indices.py
```diff
@@ -108,31 +110,34 @@ def test_query(self, client, server_config, get_helper, find_template):
...
+                "exception": requests.exceptions.Timeout(),
                 "status": HTTPStatus.GATEWAY_TIMEOUT,
             },
             {
-                "exception": requests.exceptions.InvalidURL,
+                "exception": requests.exceptions.InvalidURL(),
                 "status": HTTPStatus.INTERNAL_SERVER_ERROR,
             },
-            {"exception": Exception, "status": HTTPStatus.INTERNAL_SERVER_ERROR},
+            {"exception": Exception(), "status": HTTPStatus.INTERNAL_SERVER_ERROR},
```
webbnh, line 124: Wouldn't it make for a better test if you picked a concrete exception or a subclass of `Exception` which would then be caught by a match to `Exception`?

webbnh (approved):
Looks like a very nice improvement.  (See below for a few things to consider.)

lib/pbench/test/unit/server/query_apis/conftest.py
```diff
@@ -31,41 +17,29 @@ def query_api(client, server_config, requests_mock, request):
...
-    be a list (or tuple) with a single dict defining the "pbench" and "elastic"
-    URI paths.
+    responses mock. Exceptions are specified by providing an Exception()
+    instance as the 'body'.
 
     :return: the response object for further checking
     """
 
-    def query_api(payload, expected_index, expected_status, server_config, **kwargs):
+    def query_api(
```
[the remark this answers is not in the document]
dbutenhof: I'm aware of that, and considered picking another name, and decided it wasn't worthwhile; the caller is going to reference it by the fixture name in any case, and looking for code that doesn't match the name you're using to call it is always a complication. (Although granted the fixture model here is not exactly the clearest.)

lib/pbench/test/unit/server/query_apis/conftest.py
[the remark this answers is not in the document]
dbutenhof: You know... I actually copied that from the reference, and didn't think about it a whole lot at the time. If nothing else I suppose it declares explicitly that it's a boolean, but the syntax is rather unusual. Hmm.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
```diff
@@ -183,49 +168,39 @@ def test_empty_query(
...
+                "exception": requests.exceptions.Timeout(),
                 "status": HTTPStatus.GATEWAY_TIMEOUT,
             },
             {
-                "exception": requests.exceptions.InvalidURL,
+                "exception": requests.exceptions.InvalidURL(),
                 "status": HTTPStatus.INTERNAL_SERVER_ERROR,
             },
-            {"exception": Exception, "status": HTTPStatus.INTERNAL_SERVER_ERROR},
+            {"exception": Exception(), "status": HTTPStatus.INTERNAL_SERVER_ERROR},
```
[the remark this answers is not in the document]
dbutenhof: Maybe, but that wasn't a change in this PR, and any choice of a specific `Exception` instance aside from the ones we expect would be imaginary anyway. We really do intend this as a generic catch-all.

lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
dbutenhof: Again, the use of `Exception` rather than an arbitrary subclass predates this PR so I'm not too inclined to over-analyze. However we _do_ want to be sure that we test any `isinstance(e, Exception)` so I'm not sure any other test is truly appropriate short of iterating through all/many subclasses.

lib/pbench/test/unit/server/query_apis/test_datasets_list.py
dbutenhof: Yes, I briefly considered that; however _if_ we change from `BAD_GATEWAY` it's most likely still going to handle all unexpected Elasticsearch 400/500 class errors (`HTTPError`) the same, so such generality would be wasted. If it does become necessary, I'd prefer to defer that effort to the point where we know what we actually need.

dbutenhof:
Well that's interesting. Turns out that my laptop and the Jenkins runs are getting wildly different versions of the `responses` package. Jenkins has 0.10.5, while I'm getting 0.12.0. The `assert_call_count` method was added in 0.10.15, so the Jenkins tests are blowing up. The Jenkins container yum-installs the `python3-responses` RPM which has an (extremely) old version (from 2018) and the subsequent `pip install` is satisfied and isn't updating it when we check `test_requirements.txt`.

Ugh. Unfortunately, I'd tried the older way (checking the `len(rsp.calls)`), but apparently the `GET` doesn't count as "a call" and in order to make that test work I'd upgraded both instances to use `assert_call_count` instead.

I guess the joke's on me.

lib/pbench/test/unit/server/query_apis/conftest.py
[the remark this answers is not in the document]
webbnh: Um, if `assert_call_count()` returns some other "truey" value, then the test assertion will fail....

Reading it as "English", it makes lovely sense "assert <something> is True"...but, of course, that's what the keyword does all by itself...so the `is True` part can only get us into trouble....  :-D

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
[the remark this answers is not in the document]
webbnh: > We really do intend this as a generic catch-all.

Actually, no:  we expect the **_CUT_** to operate as a generic catch-all, which we would prove by raising something _else_ here.

And, it almost doesn't matter what we raise (so long as it should match) -- all we want to do is prove that the catch-all code responds reasonably to an otherwise unexpected exception.  So, while raising `Exception` certainly suffices, it's inferior, because that's what the code is _expecting_....  ;-)

lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
webbnh: Let me count the number of instances, "one...two...many".  OK, so we need to test three.  ;-)

(Actually, two is probably enough.  Really, I think one will do, so long as it's not `Exception` itself.  :-D )

webbnh (approved):
Looks good (although, I think there are two more places to add non-`Exception` exceptions).

portante (approved):

dbutenhof:
> Looks good (although, I think there are two more places to add non-`Exception` exceptions).

Not really, although it's a wee bit white-box, since the handlers we're testing are in the common superclass `_call`.

PR #2332, opened 2021-05-21
Adding pbench image test script

Maxusmusti:
Modified Peter's old demo script to serve as a pbench image test script.

Walks a user through a semi-automated image testing process with simple input prompts.

^^^ ORIGINAL GOAL ^^^

Now a full demo/test script with more options, defaults, error checking, greater clarity, and an example workflow to collect meaningful data.

### DIFF
--- a/agent/containers/images/image-test-script.sh
+++ b/agent/containers/images/image-test-script.sh
@@ -1,0 +1,349 @@
+#!/bin/bash
+
+# A demonstration of how to use pbench-tool-meister-start with an existing
+# Redis server and one or more remote hosts
+export CONTROLLER=$(hostname -f)
+
+DISTRO_DEFAULT="centos-8"
+TAG_DEFAULT="a933ae45e"
+REDIS_HOST_DEFAULT=$CONTROLLER
+REDIS_PORT_DEFAULT=17001
+PBENCH_RUN_DEFAULT=/var/tmp/test-run-dir
+
+while :; do
+    read -p "Specify distribution to test (or press 'ENTER' for default ${DISTRO_DEFAULT}): "
+    DISTRO=${REPLY:-${DISTRO_DEFAULT}}
+    read -p "Specify image tag to test (or press 'ENTER' for default ${TAG_DEFAULT}): "
+    TAG=${REPLY:-${TAG_DEFAULT}}
+
+    printf -- "Pulling tool-data-sink image to test...\n"
+    if podman pull quay.io/pbench/pbench-agent-tool-data-sink-$DISTRO:$TAG ; then
+        printf -- "Pull complete!\n\n"
+        break
+    else
+        printf -- "Distro/tag combination could not be found, please try again.\n\n"
+    fi
+done
+
+while :; do
+    read -p "Specify redis host (or press 'ENTER' for default '$REDIS_HOST_DEFAULT'): "
+    REDIS_HOST=${REPLY:-${REDIS_HOST_DEFAULT}}
+    if ping -c 1 -W 1 ${REDIS_HOST} 2>&1 1>/dev/null ; then
+        printf -- "Host pinged and validated!\n\n"
+        break
+    else
+        printf -- "Host specified either does not exist or is unreachable, please try again:\n"
+    fi
+done
+
+while :; do
+    read -p "Specify Redis port (or press 'ENTER' for default port '${REDIS_PORT_DEFAULT}'): "
+    let REDIS_PORT=${REPLY:-${REDIS_PORT_DEFAULT}}
+    if ((1 <= $REDIS_PORT <= 65535)); then
+        break
+    else
+        printf -- "Port must be a numeric value from 1 to 65535; please retry\n"
+    fi
+done
+
+export REDIS_HOST REDIS_PORT
+export PBENCH_REDIS_SERVER="${REDIS_HOST}:${REDIS_PORT}"
+
+function wait_keypress {
+    echo "Press any key to continue"
+    while :; do
+        read -t ${1} -n 1
+        if [[ ${?} = 0 ]]; then
+            return 0
+        else
+            echo "Waiting for a keypress"
+        fi
+    done
+}
+
+function yes_no_is_yes() {
+    prompt=$1
+    default=$2
+    y_n=" (y/n)"
+    read -N 1 -p "${prompt} ${y_n/${default}/[${default}]}: "
+    printf -- "\n"
+    lc_reply=$(tr '[:upper:]' '[:lower:]' <<< ${REPLY:-${default}})
+    if [ "${lc_reply}" == "y" ]; then
+        return 0
+    fi
+    return 1
+}
+
+clear
+
+printf -- "We are now ready to begin the guided pbench image test process. \n\n\n"
+
+wait_keypress 120
+
+# The pbench_run directory is the path to the directory which is used as the
+# persistent volume for the Tool Data Sink container, where the configuration
+# as well as the collected data will be stored.
+
+export pbench_run=$PBENCH_RUN_DEFAULT
+while [[ -d $pbench_run ]]; do
+    if yes_no_is_yes "Is it ok to delete the contents of '$pbench_run'?" "n"; then
+        rm -rf ${pbench_run}
+    else
+        read -p "Suggest a new alternate directory for the pbench run (will be wiped/created): " pbench_run
+    fi
+done
+export pbench_run
+mkdir -p ${pbench_run}
+printf -- "${pbench_run}\n" > ${pbench_run}/.path
+export pbench_log=${pbench_run}/pbench.log
+
+printf -- "\nThe benchmark run directory will be found inside:\n\n"
+
+ls -ld ${pbench_run}
+printf -- "\n"
+
+wait_keypress 120
+
+
+# Register tools
+
+printf -- "\nNow, let's pick hosts to collect data from, as well as which tools we wish to register for that host.\n\n"
+
+wait_keypress 120
+
+while :; do
+    read -p "Hostname (or press 'ENTER' for default '$CONTROLLER'): "
+    hostvar=${REPLY:-${CONTROLLER}}
+    if ping -c 1 -W 1 ${hostvar} 2>&1 1>/dev/null
+    then
+        echo $hostvar >> ${pbench_run}/remotes.lis
+        if ! yes_no_is_yes "Another host?" "n"; then break; fi
+    else
+        printf -- "Host specified either does not exist or is unreachable, please try again:\n"
+    fi
+done
+printf -- "\nDone selecting hosts:\n"
+cat ${pbench_run}/remotes.lis
+printf -- "\nNow to select data collection tools.\n\n"
+
+# Registered tools will be recorded in ${pbench_run}/tools-v1-default
+while :; do
+    read -p "Tool name (type 'help' to see options): " tool
+    if [ -z "$tool" ]
+    then
+        printf -- "Tool name empty. Please enter a tool name (at least one is required).\n"
+    elif [ $tool == "help" ]
+    then
+        printf -- "\nAvailable tools:\n"
+        pythonraw='
+            import sys, json;
+            meta = json.load(open(sys.argv[1]));
+            print(
+                " Transient:", *[f"\t{tool}" for tool in meta["transient"].keys()],
+                " Persistent:", *[f"\t{tool}" for tool in meta["persistent"].keys()],
+                sep="\n")'
+        # Remove indentation because Python cares
+        pythoncmd=$(sed -E -e 's/[[:space:]]{2,}//' <<<${pythonraw})
+        python3 - /opt/pbench-agent/tool-scripts/meta.json <<<${pythoncmd}
+    else
+        cmd="pbench-register-tool --name=${tool} --remotes=@${pbench_run}/remotes.lis"
+        printf -- "\n$cmd\n"
+        $cmd
+    fi
+    if [ -d $pbench_run/tools-v1-default ]; then
+        if ! yes_no_is_yes "Another tool?" "n"; then
+            break
+        fi
+    fi
+done
+printf -- "\nDone registering tools. See selections below:\n\n"
+ls -lR ${pbench_run}/tools-v1-default
+
+printf -- "\n\n"
+wait_keypress 120
+
+
+# Start Tool Meister containers on remote host
+
+printf -- "
+
+Now please start the Tool Meister container on the specified host...
+        $ podman run --name pbench-agent-tool-meister \\
+                --network host --ulimit nofile=65536:65536 --rm -d \\
+                -e REDIS_HOST=${REDIS_HOST} \\
+                -e REDIS_PORT=${REDIS_PORT} \\
+                -e PARAM_KEY=tm-default-\$(hostname -f) \\
+                -e _PBENCH_TOOL_MEISTER_LOG_LEVEL=debug \\
+                quay.io/pbench/pbench-agent-tool-meister-${DISTRO}:${TAG}
+
+"
+
+wait_keypress 120
+
+printf -- "\n\nNow we will automatically start the local Tool Data Sink locally to pull tool data into our local volume.\n\n"
+
+cmd="podman run --rm -d --name pbench-agent-tool-data-sink --network host
+    --volume ${pbench_run}:/var/lib/pbench-agent:Z --ulimit nofile=65536:65536
+    -e REDIS_HOST=${REDIS_HOST} -e REDIS_PORT=${REDIS_PORT}
+    -e PARAM_KEY=tds-default -e _PBENCH_TOOL_DATA_SINK_LOG_LEVEL=debug
+    quay.io/pbench/pbench-agent-tool-data-sink-${DISTRO}:${TAG}"
+printf -- "\n$cmd\n"
+$cmd
+
+printf -- "\n
+Optional: Now let's look at the 'podman logs' output from these containers;
+        You will notice, if we did not start a Redis server yet,
+        they'll just be waiting for it to show up.\n\n"
+
+wait_keypress 120
+
+printf -- "\n
+Please start the Redis server on ${REDIS_HOST}:${REDIS_PORT} if not currently up
+
+    $ podman run --name demo-tm-redis -p ${REDIS_PORT}:6379 --rm -d redis
+    
+    Note that the TDS and TMs notice the Redis server,
+    but now wait for their 'PARAM_KEY' to show up.\n\n"
+
+wait_keypress 120
+
+source /etc/profile.d/pbench-agent.sh
+source /opt/pbench-agent/base
+
+group="default"
+
+export script="demo"
+export config="my-demo-config-000"
+export benchmark_run_dir="${pbench_run}/${script}_${config}_${date_suffix}"
+mkdir ${benchmark_run_dir}
+
+printf -- "\n
+Typically pbench-tool-meister-start is expecting a '\${benchmark_run_dir}' to store data
+which is usually created by a pbench-user-benchmark and the like.
+We mimic the same behavior with:
+
+    benchmark_run_dir='${benchmark_run_dir}'\n\n"
+
+wait_keypress 120
+
+cat <<-__EOF__
+
+Now we will begin with data collection, but before doing so, feel free to run this on any tool-meister hosts:
+
+    $ podman run docker.io/alexeiled/stress-ng --cpu 4 --io 2 --timeout 300s --metrics-brief
+
+This will serve as a sample workload to demonstrate meaningful data collection.
+It runs for 300s by default, but feel free to alter that value (or ctrl-c for early graceful exit).
+
+__EOF__
+
+wait_keypress 120
+
+# Start the Tool Meisters, collecting system information, and start any persistent tools.
+cmd="_PBENCH_TOOL_MEISTER_START_LOG_LEVEL=debug pbench-tool-meister-start
+    --orchestrate=existing 
+    --redis-server=${REDIS_HOST}:${REDIS_PORT}
+    --tool-data-sink=${CONTROLLER}
+    --sysinfo=default ${group} 2>&1 | less -S"
+printf -- "\n$cmd\n"
+eval $cmd
+
+printf -- "\n
+The operation of pbench-tool-meister-start created the keys containing the
+operational data for the Tool Data Sink and the Tool Meisters, then issued the
+first 'sysinfo' collection, as requested, and sent the 'init' persistent tools
+command.
+
+At this point any registered persistent tools are up and running.
+Next is the handling of transient tool start/stop.\n\n"
+wait_keypress 120
+
+# Option to start grafana, where it is listening on port 3000.
+printf -- "\n
+You can also now run a live metrics visualizer for the Prometheus & PCP data
+
+    $ podman run --network host -d --rm --name pbench-viz quay.io/pbench/live-metric-visualizer
+    
+If done, open a browser to watch live metrics at: %s\n\n" "http://$(hostname -f):3000/"
+
+wait_keypress 120
+
+# start/stop/send tools
+
+printf -- "\n\nNow we will start and stop two iterations of transient tool data collection.\n"
+
+sample="sample42"
+iterations="0-iter-zero 1-iter-one"
+
+> ${benchmark_run_dir}/.iterations
+
+for iteration in ${iterations}; do
+    echo "${iteration}" >> ${benchmark_run_dir}/.iterations
+    benchmark_results_dir="${benchmark_run_dir}/${iteration}/${sample}"
+    mkdir -p ${benchmark_results_dir}
+
+    printf -- "\n\nStarting iteration '${iteration}'; when we continue the transient tools will be started.\n\n"
+    wait_keypress 120
+
+    pbench-start-tools --group="${group}" --dir="${benchmark_results_dir}"
+
+    printf -- "\n\nTransient tools have started for iteration '${iteration}'; when we continue they'll be stopped.\n\n"
+    wait_keypress 120
+
+    pbench-stop-tools --group="${group}" --dir="${benchmark_results_dir}"
+
+    printf -- "\n\nTools have stopped for iteration '${iteration}'; each Tool Meister still has the data held locally.\n\n"
+    wait_keypress 120
+done
+
+printf -- "\n
+We have completed our two iterations, and next we'll loop through those
+iterations requesting the tool data be sent back to the Tool Data Sink.\n\n"
+wait_keypress 120
+
+for iteration in ${iterations}; do
+    benchmark_results_dir="${benchmark_run_dir}/${iteration}/${sample}"
+    pbench-send-tools --group="${group}" --dir="${benchmark_results_dir}"
+done
+
+printf -- "\n
+We have gathered the transient data from our two iterations, and
+next we'll stop the Tool Meisters; this involves ending any persistent
+tools, and gathering the final 'sysinfo' collection requested.\n\n"
+wait_keypress 120
+
+cmd="_PBENCH_TOOL_MEISTER_STOP_LOG_LEVEL=debug pbench-tool-meister-stop --sysinfo=default ${group}"
+printf -- "\n$cmd\n"
+eval $cmd
+
+printf -- "\n
+At this point the Tool Data Sink has stopped, along with the Tool
+Meisters. The Redis server is still running, since the pbench-agent
+CLI commands did not start it.  Next we dump the final directory
+hierarchy of collected data to complete the test.\n\n"
+wait_keypress 120
+
+# Dump our local environment
+find ${pbench_run} -ls | less -S
+
+wait_keypress 120
+
+if [ "$(podman ps | grep pbench-viz)" ]; then
+    if yes_no_is_yes "Would you like the live-metric-visualizer to be terminated?" "n"
+    then
+        cmd="podman kill pbench-viz"
+        printf -- "Running: $cmd\n"
+        $cmd
+    fi
+fi
+
+if [ "$(podman ps | grep demo-tm-redis)" ]; then
+    if yes_no_is_yes "Would you like the redis container to be terminated?" "n"
+    then
+        cmd="podman kill demo-tm-redis"
+        printf -- "Running: $cmd\n"
+        $cmd
+    fi
+fi
+
+printf -- "\nCongratulations, you've completed the full benchmarking process!\n"


portante:
See also: https://github.com/distributed-system-analysis/pbench/pull/2318/files?file-filters%5B%5D=.sh#diff-e6ea8587fdafb54df78468e93dca74cc2c4df53ec21ec28f8d45bbda94009ad1

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,188 @@
...
+
+wait_keypress 120
+
+# If somebody wants to write their own pbench-tool-meister-start, great, then
+# that person can change the assumption on where the persistent volume is for
+# the Tool Data Sink, and how the "sysinfo", "init", and "end" commands
+# operate with respect to data storage.
+
+export pbench_run=${HOME}/test-run-dir
+rm -rf ${pbench_run}
```
dbutenhof, line 42: As a "more formal" demo script in GitHub, it'd be better to put this under `/var/tmp` and avoid arbitrarily deleting any directory path in the user's directory.

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,188 @@
...
+read -p "Specify distribution to test: " distrovar
+DISTRO=$distrovar
+
+read -p "Specify image tag to test: " tagvar
+TAG=$tagvar
+
+read -p "Specify redis host: " redishost
+read -p "Specify redis port: " redisport
+export REDIS_HOST=$redishost
+export REDIS_PORT=$redisport
```
dbutenhof, line 15: You could avoid the extra shell variables by:

```
read -p "Specify redis host: " REDIS_HOST
read -p "Specify redis port: " REDIS_PORT
export REDIS_HOST REDIS_PORT
```

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,188 @@
...
+
+wait_keypress 120
+
+
+# Start the Tool Meisters, collecting system information, and start any persistent tools.
+set -x
+_PBENCH_TOOL_MEISTER_START_LOG_LEVEL=debug pbench-tool-meister-start --sysinfo=default ${group} 2>&1 | less -S
+set +x
+
+printf -- "\n\nThe operation of pbench-tool-meister-start creates the keys containing the operational data for the Tool Data Sink and the Tool Meisters, then issued the first 'sysinfo' collection, as requested, and sent the 'init' persistent tools command.\n\nAt this point any registered persistent tools are up and running. Next is the handling of transient tool start/stop.\n\n"
```
dbutenhof, line 129: The tenses here could be cleaned up. E.g., "The pbench-tool-meister-start command first created the Redis keys to communicate with the Tool Data Sink and Tool Meisters, then issued ..."

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,188 @@
...
+
+set -x
+_PBENCH_TOOL_MEISTER_STOP_LOG_LEVEL=debug pbench-tool-meister-stop --sysinfo=default ${group}
+set +x
+
+
+printf -- "\n\nAt this point the Tool Data Sink has stopped, along with the Tool Meisters. The Redis server is still running, since the pbench-agent CLI commands did not start it.  Next we dump the final directory hierarchy of collected data to complete the test.\n\n"
+wait_keypress 120
+
+# Dump our local environment, note we are not running as root.
```
dbutenhof, line 187: We might want to make the "we are not running as root" part of one of the printf descriptions, possibly up at the front.

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,188 @@
+#!/bin/bash
+
+# A demonstration of how to use pbench-tool-meister-start with an existing
+# Redis server, and one or more remote
```
dbutenhof, line 4: The script only allows for a single remote; although that could be changed by prompting in a loop as done for tools.

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,188 @@
...
+while [ $anothertool == "y" ]
+do
+    read -p "Tool name: " tool
+    printf -- "\npbench-register-tool --name=${tool} --remotes=@${pbench_run}/remotes.lis\n"
+    pbench-register-tool --name=${tool} --remotes=@${pbench_run}/remotes.lis
+    sleep 1
+    read -p "Another tool? (y/n)" anothertool
+done
+printf -- "\nDone registering tools.\n\n"
+ls -lR ${pbench_run}/tools-v1-default
```
dbutenhof, line 76: I find the switch back and forth between `find` and `ls -lR` a bit disturbing; is there really a reason not to be consistent?

dbutenhof (commented):
Some random stream-of-consciousness comments as I read through it...

agent/containers/images/image-test-script.sh
portante: Long file name paths.

dbutenhof: Yes, the format differs; but then, that's one of the reasons that switching back and forth is disruptive. Not a big deal, so I won't dwell on it further.

dbutenhof (approved):
Ha; you "resolved" my comment on the plural hosts without "fixing it", and I wondered why for a moment before finding that you'd changed the code instead of changing the comment. Great!

This looks good. I think the only remaining meta-comment I might make is that I'd said something about possibly emphasizing the low-key "not running as root" comment, and you removed that phrase without elevating it or printing it. A good place for this might be the `printf` at 125 which mentions `benchmark_run_dir` and _could_ be expanded to explain that we "mimic the same behavior" _because_ we're not requiring `root` and need a local writable directory. Then again, the demo script doesn't have to do/explain everything in detail.

Maxusmusti:
Note: This script currently assumes the redis host and TDS host are the same during tm-start. This will be fixed and updated after the demo today.

Maxusmusti:
Another point to consider is where this script is going to be used. Should we expect it to be called from within the pbench repo/install location, or should we also be prepared for people to download it separately and/or use it from a different location.

AKA: Should the help option path be absolute or relative, and should we check if pbench-agent is installed on the controller, or assume it?

webbnh:
> Should we expect it to be called from within the pbench repo/install location, or should we also be prepared for people to download it separately and/or use it from a different location.
> 
> AKA: Should the help option path be absolute or relative, and should we check if pbench-agent is installed on the controller, or assume it?

I would suggest that we want people to be able to install the agent and run the demo...so, it certainly shouldn't be attached to or depend on having a checkout of the repo.

That said, I think it would be good if it "just worked" if I log into a system where the agent is installed and I run the demo script from an arbitrary working directory.  I don't know whether that means having absolute paths, or whether it needs to be based on an environment variable or other config parameter, or whether it should find stuff via `$PATH`, or what.

Actually, I've kind of lost track...does the agent actually need to be installed on the host where the demo is run?  What are the possibilities for making the demo script stand-alone and not even dependent on a (local) agent installation?

Maxusmusti:
> > Should we expect it to be called from within the pbench repo/install location, or should we also be prepared for people to download it separately and/or use it from a different location.
> > AKA: Should the help option path be absolute or relative, and should we check if pbench-agent is installed on the controller, or assume it?
> 
> I would suggest that we want people to be able to install the agent and run the demo...so, it certainly shouldn't be attached to or depend on having a checkout of the repo.
> 
> That said, I think it would be good if it "just worked" if I log into a system where the agent is installed and I run the demo script from an arbitrary working directory. I don't know whether that means having absolute paths, or whether it needs to be based on an environment variable or other config parameter, or whether it should find stuff via `$PATH`, or what.
> 
> Actually, I've kind of lost track...does the agent actually need to be installed on the host where the demo is run? What are the possibilities for making the demo script stand-alone and not even dependent on a (local) agent installation?

Currently, the script requires only the local agent installation (for the tool registration, tool meister start command, iterations, and tool meister stop command). The rest (tool data sink, tool meisters, redis, visualizer, etc.) is done with containers. The whole thing could probably be done entirely with containerized pbench and a volume mount, would that be preferable? Or perhaps having both options? Or maybe it's better to have the user just yum install the pbench-agent before using the script? Open to suggestions

Maxusmusti:
Once again caught up with all feeback/suggestions and ready for review!

agent/containers/images/image-test-script.sh
```diff
@@ -10,8 +10,39 @@ DISTRO=$distrovar
...
+        REDIS_HOST=$CONTROLLER
+    fi
+    ping_first=$(ping -c 1 -W 1 ${REDIS_HOST} 2>&1 | grep 'Name or service not known')
+    ping_second=$(ping -c 1 -W 1 ${REDIS_HOST} 2>&1 | grep '100% packet loss')
+    if [ -n "$ping_first" ] || [ -n "$ping_second" ]
+    then
+        printf -- "Host specified either does not exist or is unreachable, please try again:\n"
+    else
+        badhost=false
+    fi
```
dbutenhof, line 48: The whole loop is basically repeated, but this phrase in particular could easily be refactored out into a function to be reused to validate a hostname.

dbutenhof (dismissed):

webbnh:
> The whole thing could probably be done entirely with containerized pbench and a volume mount, would that be preferable? Or perhaps having both options? Or maybe it's better to have the user just yum install the pbench-agent before using the script?

While it would be very cool to be able to just pull and run a container, I think it makes good sense to have the user "install Pbench" and then run the demo -- after all, the purpose is to enable the user to use Pbench, and, unless and until we go over to a fully-containerized model, the way to do that will be to install it and then run it.

agent/containers/images/image-test-script.sh
```diff
@@ -4,14 +4,65 @@
...
-read -p "Specify distribution to test: " distrovar
-DISTRO=$distrovar
+verified=""
+while [ -z "$verified" ]
+do
+    read -p "Specify distribution to test (or press 'ENTER' for default 'centos-8'): " DISTRO
+    if [ -z "$DISTRO" ]
+    then
+        DISTRO="centos-8"
+    fi
```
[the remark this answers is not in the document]
webbnh: Also, I think there is a nicer idiom to use here:
```
DISTRO_DEFAULT='centos-8'
TAG_DEFAULT="a933ae45e"
[...]
while :; do
    read -p "Specify distribution to test (or press 'ENTER' for default ${DISTRO_DEFAULT}): "
    DISTRO=${REPLY:-${DISTRO_DEFAULT}}
    read -p "Specify image tag to test (or press 'ENTER' for default ${TAG_DEFAULT}): "
    TAG=${REPLY:-${TAG_DEFAULT}}

    if podman pull quay.io/pbench/pbench-agent-tool-data-sink-$DISTRO:$TAG
    then
        printf -- "Complete!\n\n" break
        break
    else
        printf -- "Distro/tag combination could not be found, please try again.\n\n"
    fi
done
```

agent/containers/images/image-test-script.sh
```diff
@@ -4,14 +4,65 @@
...
+    then
+        REDIS_PORT=17001
+    fi
+    if [ "$REDIS_PORT" -eq "$REDIS_PORT" 2> /dev/null ]
+    then
+        badport=false
+    else
+        printf -- "Value inputted was not numeric, please retry\n"
+    fi
+done
```
webbnh, line 65: I would recast this like:
```
REDIS_PORT_DEFAULT=17001
[...]
while :; do
    read -p "Specify Redis port (or press enter for default port '${REDIS_PORT_DEFAULT}'): "
    let REDIS_PORT=${REPLY:-${REDIS_PORT_DEFAULT}}
    if [[ $REDIS_PORT -ne 0 ]]
    then
        break
    else
        printf -- "Port must be a non-zero numeric value; please retry\n"
    fi
done
```

agent/containers/images/image-test-script.sh
```diff
@@ -60,9 +111,20 @@ wait_keypress 120
...
+    then
+        hostvar=$CONTROLLER
+    fi
+    ping_first=$(ping -c 1 -W 1 ${hostvar} 2>&1 | grep 'Name or service not known')
+    ping_second=$(ping -c 1 -W 1 ${hostvar} 2>&1 | grep '100% packet loss')
+    if [ -n "$ping_first" ] || [ -n "$ping_second" ]
+    then
+        printf -- "Host specified either does not exist or is unreachable, please try again:\n"
+    else
+        echo $hostvar >> ${pbench_run}/remotes.lis
```
webbnh, line 125: You're appending to the file, here...how do you know that the file was empty to start with?

Having it persist from run to run might be a feature...but, you'll want to display the contents to the user, get confirmation that it's acceptable, and you might want to go down the list and ping them.

Also, see previous comments regarding improving the code in this loop.

agent/containers/images/image-test-script.sh
```diff
@@ -74,15 +136,20 @@ anothertool=y
...
-        pbench-register-tool --name=${tool} --remotes=@${pbench_run}/remotes.lis
-        sleep 1
-        read -p "Another tool? (y/n)" anothertool
+        if [ $tool == "help" ]
+        then
+            printf -- "\nAvailable tools:\n"
+            python3 -c "import sys, json; meta = json.load(open(sys.argv[1])); print('  Transient:', *[f'\t{tool}' for tool in meta['transient'].keys()], '  Persistent:', *[f'\t{tool}' for tool in meta['persistent'].keys()], sep='\n')" /opt/pbench-agent/tool-scripts/meta.json
+        else
+            printf -- "\npbench-register-tool --name=${tool} --remotes=@${pbench_run}/remotes.lis\n"
+            pbench-register-tool --name=${tool} --remotes=@${pbench_run}/remotes.lis
```
webbnh, line 149: It would be better to create the command in a variable, print the variable, and then execute it -- that way what the user sees cannot be different from what gets run:
```
            cmd = "pbench-register-tool --name=${tool} --remotes=@${pbench_run}/remotes.lis"
            printf -- "\n${cmd}\n"
            $cmd
```

agent/containers/images/image-test-script.sh
```diff
@@ -74,15 +136,20 @@ anothertool=y
...
+            sleep 1
```
webbnh, line 150: Why sleep?

agent/containers/images/image-test-script.sh
```diff
@@ -133,6 +200,9 @@ printf -- "\n\nTypically pbench-tool-meister-start is expecting a '\${benchmark_
 
 wait_keypress 120
 
+printf -- "\n\nNow we will begin with data collection, but before doing so, feel free to run this on any tool-meister hosts:\n\n\t$ podman run docker.io/alexeiled/stress-ng --cpu 4 --io 2 --timeout 300s --metrics-brief\n\nThis will serve as a sample workload to demonstrate meaningful data collection.\nIt runs for 300s by default, but feel free to alter that value (or ctrl-c for early graceful exit).\n\n"
```
webbnh, line 203: Um...could we possibly break this up into readable lines?  There are several nice ways to do that...here's one:
```
cat <<-__EOF__


Now we will begin with data collection, but before doing so, feel free to run this on any tool-meister hosts:

    $ podman run docker.io/alexeiled/stress-ng --cpu 4 --io 2 --timeout 300s --metrics-brief

This will serve as a sample workload to demonstrate meaningful data collection.
It runs for 300s by default, but feel free to alter that value (or ctrl-c for early graceful exit).

__EOF__
```

webbnh (changes requested):
Looks good.  However, I have a bunch of suggestions for tweaking the code:

- You should put the default values for the prompts in variables:
  - This ensures that the contents of the prompts match the actual values used; and,
  - The defaults can be declared at the top of the script where they are easy to find and change.
- The shell has a syntax for assigning variables using default values -- you should use it -- it works nicely with the implicit target of the `read` function, `REPLY`.
- Consider using an "infinite" loop construct, with an explicit `break`, instead of adding the arbitrary "flag" variables.
- When evaluating the success of a command, generally speaking you should use its exit code (especially for `ping`).  Aside from being simpler and more reliable, you can put the command right in the `if` statement if it's not too complex.
- Your trick for checking for numeric input deserves a code comment, unless you go with my suggested alternative.
- I didn't see where you are truncating or removing `remotes.lis`....

And, there are a few smaller nits.

agent/containers/images/image-test-script.sh
Maxusmusti: In the lines above the run directory is wiped and remade, so it should be fresh for each run:
`rm -rf ${pbench_run}`
`mkdir ${pbench_run}`
But yeah the loop needs improvement

agent/containers/images/image-test-script.sh
```diff
@@ -4,65 +4,49 @@
...
+    TAG=${REPLY:-${TAG_DEFAULT}}
 
     printf -- "Pulling tool-data-sink image to test...\n"
-    verified=$(podman pull quay.io/pbench/pbench-agent-tool-data-sink-$DISTRO:$TAG 2> /dev/null)
-
-    if [ -z "$verified" ]
+    if podman pull quay.io/pbench/pbench-agent-tool-data-sink-$DISTRO:$TAG
     then
-        printf -- "Distro/tag combination could not be found, please try again.\n\n"
+        printf -- "Complete!\n\n" break
```
dbutenhof, line 21: Why are you _printing_ "break"?? Is that intentional?

Also, although the "Complete!" was there before, looking at it this way I'm not sure that's an appropriate message for validation of the first interactive parameter. We're hardly "complete" at this point. Should we instead confirm that the tag is validated and the container has been pulled?

And ... would this be a good place to use the `cmd="podman [...]" ; print ${cmd} ; ${cmd}` pattern as we do in `pbench-register-tool` at the end, so the user knows what's happening, rather than just "Pulling image"??

agent/containers/images/image-test-script.sh
```diff
@@ -4,65 +4,49 @@
...
+    if ping -c 1 -W 1 ${REDIS_HOST} 2>&1 1>/dev/null
     then
-        REDIS_HOST=$CONTROLLER
-    fi
-    ping_first=$(ping -c 1 -W 1 ${REDIS_HOST} 2>&1 | grep 'Name or service not known')
-    ping_second=$(ping -c 1 -W 1 ${REDIS_HOST} 2>&1 | grep '100% packet loss')
-    if [ -n "$ping_first" ] || [ -n "$ping_second" ]
-    then
-        printf -- "Host specified either does not exist or is unreachable, please try again:\n"
+        break
```
dbutenhof, line 33: Since we're validating successful tag/container validation above, should we also be explicitly confirming Redis host validation/reachability here? Should we announce up front that we're pinging the Redis host?

dbutenhof (approved):
No real problem, but the new structure made me think a bit more about the pattern of prompts and responses we're giving... and about whether we ought to consider making the structure a bit more consistent.

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,283 @@
...
+printf -- "\nDone registering tools.\n\n"
+ls -lR ${pbench_run}/tools-v1-default
+
+printf -- "\n\n"
+wait_keypress 120
+
+
+# Start Tool Meister containers on remote host
+
+printf -- "\n\nNow please start the Tool Meister container on the specified host...\n\n\t$ podman run --name pbench-agent-tool-meister \\ \n\t\t--network host --ulimit nofile=65536:65536 --rm -d \\ \n\t\t-e REDIS_HOST=${REDIS_HOST} \\ \n\t\t-e REDIS_PORT=${REDIS_PORT} \\ \n\t\t-e PARAM_KEY=tm-default-\$(hostname -f) \\ \n\t\t-e _PBENCH_TOOL_MEISTER_LOG_LEVEL=debug \\ \n\t\tquay.io/pbench/pbench-agent-tool-meister-${DISTRO}:${TAG}\n\n"
```
webbnh, line 147: I don't think we win any prizes for performing this in a single line -- just put it in double quotes and lay it out the way you want it to appear:
```
printf -- "

Now please start the Tool Meister container on the specified host...
        \$ podman run --name pbench-agent-tool-meister \\
                --network host --ulimit nofile=65536:65536 --rm -d \\
                -e REDIS_HOST=${REDIS_HOST} \\
                -e REDIS_PORT=${REDIS_PORT} \\
                -e PARAM_KEY=tm-default-\$(hostname -f) \\
                -e _PBENCH_TOOL_MEISTER_LOG_LEVEL=debug \\
                quay.io/pbench/pbench-agent-tool-meister-${DISTRO}:${TAG}

"
```
This could also be done using the idiom that you used on line 186.

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,283 @@
...
+
+wait_keypress 120
+
+
+# Start the Tool Data Sink locally, mapping our volume into the Tool Data Sink
+# container.
+
+printf -- "\n\nNow we will automatically start the local Tool Data Sink to pull tool data into our local volume.\n\n"
```
webbnh, line 155: In cases like this it seems odd to have _both_ code comments and print outs -- that is, the stuff in the comment seems like it should be in the print out, and then the comment is redundant.

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,283 @@
...
+
+set -x
+podman run --name pbench-agent-tool-data-sink --network host --volume ${pbench_run}:/var/lib/pbench-agent:Z --ulimit nofile=65536:65536 --rm -d -e REDIS_HOST=${REDIS_HOST} -e REDIS_PORT=${REDIS_PORT} -e PARAM_KEY=tds-default -e _PBENCH_TOOL_DATA_SINK_LOG_LEVEL=debug quay.io/pbench/pbench-agent-tool-data-sink-${DISTRO}:${TAG}
```
webbnh, line 158: Break this up:
```
podman run --rm -d --name pbench-agent-tool-data-sink --network host \
    --volume ${pbench_run}:/var/lib/pbench-agent:Z --ulimit nofile=65536:65536
    -e REDIS_HOST=${REDIS_HOST} -e REDIS_PORT=${REDIS_PORT} \
    -e PARAM_KEY=tds-default -e _PBENCH_TOOL_DATA_SINK_LOG_LEVEL=debug \
    quay.io/pbench/pbench-agent-tool-data-sink-${DISTRO}:${TAG}
```

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,283 @@
...
+set +x
+
+printf -- "\n\nOptional: Now let's look at the 'podman logs' output from these containers;\n\t You will notice if we did not start a Redis server yet,\n\tthey'll just be waiting for it to show up.\n\n"
```
webbnh, line 161: Break this (and the one at line 166) up, too:
```
printf -- "\n
Optional: Now let's look at the 'podman logs' output from these containers;
        You will notice if we did not start a Redis server yet,
        they'll just be waiting for it to show up.\n\n"
```

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,283 @@
...
+
+
+printf -- "\n\nPlease start the Redis server on ${REDIS_HOST}:${REDIS_PORT} if not currently up\n\t$ podman run --name demo-tm-redis -p ${REDIS_PORT}:6379 --rm -d redis\n\n\tNote TDS and TMs notice Redis server,\n\tbut now wait for their 'PARAM_KEY' to show up.\n\n"
+
+wait_keypress 120
+
+export date="2021-1-1T12:00:00"
+
+source /etc/profile.d/pbench-agent.sh
+source /opt/pbench-agent/base
```
webbnh, line 173: Why do we set `date` to a fixed value instead of letting `/opt/pbench-agent/base` set it to the current value?  (This would be worth an explanatory comment.)

agent/containers/images/image-test-script.sh
webbnh: And line 182, et al.

agent/containers/images/image-test-script.sh
webbnh: Ditto line 200.

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,283 @@
...
+wait_keypress 120
+
+# Option to start grafana, where it is listening on port 3000.
+printf -- "\n\nYou can also now run a live metrics visualizer for the Prometheus & PCP data\n\n\t$ podman run --network host -d --rm --name pbench-viz quay.io/pbench/live-metric-visualizer\n\n"
+
+printf -- "\n\nIf done, open a browser to watch live metrics at: %s\n\n" "http://$(hostname -f):3000/"
+wait_keypress 120
+
+
+# TBD start/stop/send tools
```
webbnh, line 214: Is this comment stale?

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,283 @@
...
+        printf -- "Tool name empty. Please write a tool name.\n"
+    else
+        if [ $tool == "help" ]
+        then
+            printf -- "\nAvailable tools:\n"
+            python3 -c "import sys, json; meta = json.load(open(sys.argv[1])); print('  Transient:', *[f'\t{tool}' for tool in meta['transient'].keys()], '  Persistent:', *[f'\t{tool}' for tool in meta['persistent'].keys()], sep='\n')" /opt/pbench-agent/tool-scripts/meta.json
+        else
+            cmd="pbench-register-tool --name=${tool} --remotes=@${pbench_run}/remotes.lis"
+            printf -- "\n$cmd\n"
+            $cmd
```
webbnh, line 129: Below, you use `set -x` to produce this effect...it would be good to use a consistent idiom.

agent/containers/images/image-test-script.sh
webbnh: Ditto line 251, et al.

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,283 @@
...
+
+printf -- "\n\nAt this point the Tool Data Sink has stopped, along with the Tool Meisters. The Redis server is still running, since the pbench-agent CLI commands did not start it.  Next we dump the final directory hierarchy of collected data to complete the test.\n\n"
+wait_keypress 120
+
+# Dump our local environment
+find ${pbench_run} -ls | less -S
+
+wait_keypress 120
+
+read -p "Would you like the live-metric-visualizer to be terminated if started locally? (y/n): " term_vis
```
webbnh, line 269: It seems like you could answer the "if started locally" programmatically and then skip this question if it's not there.

webbnh: Ditto for the Redis container at line 276.

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,283 @@
...
+fi
+
+read -p "\nWould you like the redis container to be terminated if started locally? (y/n): " term_redis
+if [ $term_redis == "y" ]
+then
+    printf -- "Running: podman kill demo-tm-redis\n"
+    podman kill demo-tm-redis
+fi
+
+printf -- "\nCongratulations, you've completed the full benchmarking process!\n"
```
webbnh, line 283: The file should end in a newline.

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,283 @@
...
+
+# Dump our local environment
+find ${pbench_run} -ls | less -S
+
+wait_keypress 120
+
+read -p "Would you like the live-metric-visualizer to be terminated if started locally? (y/n): " term_vis
+if [ $term_vis == "y" ]
+then
+    printf -- "Running: podman kill pbench-viz\n"
```
webbnh, line 272: This seems like a nice spot for the `set -x` trick.  (Ditto line 279.)

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,283 @@
...
+
+while :; do
+    read -p "Specify distribution to test (or press 'ENTER' for default ${DISTRO_DEFAULT}): "
+    DISTRO=${REPLY:-${DISTRO_DEFAULT}}
+    read -p "Specify image tag to test (or press 'ENTER' for default ${TAG_DEFAULT}): "
+    TAG=${REPLY:-${TAG_DEFAULT}}
+
+    printf -- "Pulling tool-data-sink image to test...\n"
+    if podman pull quay.io/pbench/pbench-agent-tool-data-sink-$DISTRO:$TAG
+    then
```
portante, line 20: To be consistent with `while`/`do`, you would move the `then` to the end of the `if` line, separated by a semi-colon (';').

portante (changes requested):
I am wondering if it is worth breaking this file up into two separate scripts: one to drive asking "hand-holding" type questions and execute setup, and another to drive the steps of the demo.

That way, one can use the demo driver with an alternate method.

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,283 @@
...
+            return 0
+        else
+            echo "waiting for the keypress"
+        fi
+    done
+}
+
+clear
+
+printf -- "We will now begin the guided pbench image test process. \n\n\n"
```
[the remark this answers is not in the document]
Maxusmusti: I considered all of those as part of the "test process", though I guess I could distinguish between setup and run

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,283 @@
...
+
+wait_keypress 120
+
+# If somebody wants to write their own pbench-tool-meister-start, great, then
+# that person can change the assumption on where the persistent volume is for
+# the Tool Data Sink, and how the "sysinfo", "init", and "end" commands
+# operate with respect to data storage.
```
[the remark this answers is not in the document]
Maxusmusti: Yeah this was an old, outdated comment that should have been removed months ago

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,283 @@
...
+# Redis server, and one or more remote
+export CONTROLLER=$(hostname -f)
+
+DISTRO_DEFAULT="centos-8"
+TAG_DEFAULT="a933ae45e"
+REDIS_HOST_DEFAULT=$CONTROLLER
+REDIS_PORT_DEFAULT=17001
+
+while :; do
+    read -p "Specify distribution to test (or press 'ENTER' for default ${DISTRO_DEFAULT}): "
```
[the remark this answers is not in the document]
Maxusmusti: I'll probably leave it verbose just for the case where people use this script to demo pbench. That way people viewing the demo will surely have enough information to understand if the presenter just presses enter.

dbutenhof (approved):
While I'm inclined to agree with Webb's suggestion about the prompts, we can always tweak this later. I could see some argument in a "demo" script like this for a common prompt function that takes message, default, and help-string parameters, which ensures a common formatting, supports "?" help, and so forth ... but that's arguably over-engineering at this point and I'm happy enough with what you've accomplished here.

webbnh (changes requested):
It looks great.  However:
- There are a couple of `NOTE`'s which should be either fixed or just removed.
- Places where a prompt applies a default value for an empty response should indicate what the default is in some way.
- I've got a couple of suggestions for making spots in the code a little more rigorous/resilient.

Beyond those, I've called out a bunch of small things for additional polish.

portante (dismissed):

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,333 @@
...
+    fi
+done
+
+while :; do
+    read -p "Specify Redis port (or press 'ENTER' for default port '${REDIS_PORT_DEFAULT}'): "
+    let REDIS_PORT=${REPLY:-${REDIS_PORT_DEFAULT}}
+    if ((1 <= $REDIS_PORT <= 65535)); then
+        break
+    else
+        printf -- "Port must be a non-zero numeric value; please retry\n"
```
dbutenhof, line 45: Well, technically, "must be a numeric value from 1 to 65535", which might be worth including here ... or someone could get errors where they're clearly following your instructions.

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,333 @@
...
+printf -- "We are now ready to begin the guided pbench image test process. \n\n\n"
+
+wait_keypress 120
+
+# The pbench_run directory is the path to the directory which is used as the
+# persistent volume for the Tool Data Sink container, where the configuration
+# as well as the collected data will be stored.
+
+export pbench_run=$PBENCH_RUN_DEFAULT
+while [[ -d $pbench_run ]]; do
```
dbutenhof, line 75: Hmm; if they specify an alternate `pbench_run` to your `read` which isn't a valid directory name, you'll exit the loop but then fail later trying to create it.

agent/containers/images/image-test-script.sh
```diff
@@ -0,0 +1,333 @@
...
+            printf -- "\nAvailable tools:\n"
+            pythonraw='
+                import sys, json;
+                meta = json.load(open(sys.argv[1]));
+                print(
+                    " Transient:", *[f"\t{tool}" for tool in meta["transient"].keys()],
+                    " Persistent:", *[f"\t{tool}" for tool in meta["persistent"].keys()],
+                    sep="\n")'
+            # Remove indentation because Python cares
+            pythoncmd=$(sed -E -e 's/[[:space:]]{2,}//' <<<${pythonraw})
```
dbutenhof, line 138: You could do this with `jq`, though learning the syntax is a bit tricky ...

dbutenhof (approved):
A bunch of edges could still be polished, but this looks pretty good.

agent/containers/images/image-test-script.sh
webbnh: _[Once again unresolving to start a new topic for the same line...]_

Do we really want to rely on finding the tools in `${pbench_run}/tools-v1-default`, or should we be using a command, like `pbench-list-tools`, to generate this output?

webbnh (approved):
Lookin' good.  I've got a bunch of nits and suggestions which you can consider changing, and one pointed question (which probably @portante will have to answer).

agent/containers/images/image-test-script.sh
Maxusmusti: pbench-list-tools doesn't seem to be able to recognize any tools as registered due to the non-standard registration location.

agent/containers/images/image-test-script.sh
```diff
@@ -75,14 +75,22 @@ export pbench_run=$PBENCH_RUN_DEFAULT
...
+                printf -- "Invalid dir '$pbench_run', cannot be created\n"
+            fi
+        done
     else
+        rm -rf ${pbench_run}
         break
     fi
 done
-rm -rf ${pbench_run}
 mkdir ${pbench_run}
```
webbnh: You shouldn't need the `break` statement at line 91 -- the `while` test at line 75 will let control out of the loop if the `rm` succeeded (and, if it didn't succeed, then you probably would like to loop again...).  And, therefore, I don't think you need the `else` at line 89, either.  And, I'm not sure that  the interior `while` loop is needed or does what you want.

Would something like the following suffice?
```
while [[ -d $pbench_run ]]; do
    read -p "Is it ok to delete the contents of '$pbench_run'? (y/[n]): " delete_check
    delete_check=$(echo $delete_check | tr '[:upper:]' '[:lower:]')
    if [[ "$delete_check" == 'y' || "$delete_check" == "yes" ]]; then
        rm -rf ${pbench_run}
    else
        read -p "Suggest a new alternate directory for the pbench run (will be wiped/created): " pbench_run
done
export pbench_run
mkdir ${pbench_run}
```
This doesn't handle failures to create `${pbench_run}`, but...how paranoid do we want to be?

agent/containers/images/image-test-script.sh
```diff
@@ -312,7 +332,8 @@ wait_keypress 120
 
 if [ "$(podman ps | grep pbench-viz)" ]; then
     read -p "Would you like the live-metric-visualizer to be terminated? (y/[n]): " term_vis
-    if [[ -n $term_vis && $term_vis == "y" ]]
+    term_vis=$(echo $term_vis | tr '[:upper:]' '[:lower:]')
+    if [[ "$term_vis" == "y" || "$term_vis" == "yes" ]]
```
webbnh: This looks ripe for refactoring.  These three lines (and the other y/n prompts) could be just an `if` statement invoking a function:
```
    if yes_no_is_yes("Would you like the live-metric-visualizer to be terminated?", "n")
```
which invokes something like:
```
function yes_no_is_yes() {
    prompt=$1
    default=$2
    y_n=" (y/n)"
    read -p "${prompt} ${y_n/${default}/[${default}]}"
    lc_reply=$(tr '[:upper:]' '[:lower:]' <<< ${REPLY:-${default}})
    return [[ "${lc_reply}" == "y" || "${ls_reply}" == "yes" ]]
}
```

By the way, you might want to look into some of the other switches that `read` takes.  For instance, if you used the `-n 1` switch, which limits the reply to one character, then you could simplify the subsequent comparison (by just comparing it to `y`, or to `y` or `Y` and skip the translation to lower case).  And, for the prompts for things like filenames, you should consider using `-e`, which would enable command line editing for the user.

webbnh (changes requested):
Looking good.  However, I think the `$pbench_run` deletion/creation code needs one more try.

Also, you might want to consider refactoring the y/n prompt code into a function.

agent/containers/images/image-test-script.sh
webbnh: I'm confused...are you saying that the any tools that we registered at lines 149-151 using `pbench-register-tool` cannot be detected at line 153 using `pbench-list-tools`??  How is that?

agent/containers/images/image-test-script.sh
Maxusmusti: The internal while loop is there to ensure that any specified dir for creation has an existing parent. You are right about the else and break though, and this can be cut down a bit.

Maxusmusti: Actually I've re-added the else, since I want to prompt the user again for an extra layer of protection (before deleting an existing dir)

webbnh: > to ensure that any specified dir for creation has an existing parent

This isn't strictly necessary; and, if you add the `-p` switch to the `mkdir` command, then it will automatically create the parent(s), if necessary.

> I've re-added the else, since I want to prompt the user again for an extra layer of protection

The code that I proposed also does that:  if the user responds to the "alternate directory" prompt, then control returns to the top of the loop; if the alternate directory exists, the script will prompt before attempting to delete it.

portante (changes requested):

agent/containers/images/image-test-script.sh
Maxusmusti: This is a great point. Using your suggestion, I've significantly simplified the loop, thanks!

agent/containers/images/image-test-script.sh
Maxusmusti: Yeah, it seems to be the case that despite the fact that pbench-register-tools is registering in the new pbench_run dir, pbench-list-tools is still sticking to the default dir, /var/lib/pbench-agent. Even testing it seperately, pbench-list-tools does not change behavior when $pbench_run is defined/changed. I am testing this with a 0.71.0 build that's a few months old. Not sure if it has been addressed since, but may not be a good idea to assume all users have a super recent build.

webbnh: Tracked in Issue #2473.

webbnh (approved):
Ship it!

(I had to manually start the Jenkins job, but all seems happy, now.... 😕)

dbutenhof (approved):

portante (dismissed):

dbutenhof (approved):

PR #2336, opened 2021-05-25
Various fixes to pbench-list-tools

### LINKED ISSUE #2302: pbench-list-tools (v0.71 - python version) produces a traceback  for an unknown group
It should probably catch the exception and exit with an error message. Instead it does this:

```
# pbench-list-tools --group=bar
  File "/opt/pbench-agent/lib/pbench/agent/tool_group.py", line 43, in verify_tool_group
    tg_dir = tg_dir_name.resolve(strict=True)
  File "/usr/lib64/python3.9/pathlib.py", line 1204, in resolve
    s = self._flavour.resolve(self, strict=strict)
  File "/usr/lib64/python3.9/pathlib.py", line 362, in resolve
    return _resolve(base, str(path)) or sep
  File "/usr/lib64/python3.9/pathlib.py", line 346, in _resolve
    target = accessor.readlink(newpath)
  File "/usr/lib64/python3.9/pathlib.py", line 451, in readlink
    return os.readlink(path)
FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/pbench-agent/tools-v1-bar'

During handling of the above exception, another exception occurred:

  File "/opt/pbench-agent/util-scripts/pbench-list-tools", line 10, in <module>
    sys.exit(main())
  File "/usr/lib/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python3.9/site-packages/click/decorators.py", line 73, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/usr/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/opt/pbench-agent/lib/pbench/cli/agent/commands/tools/list.py", line 135, in main
    status = ListTools(ctxt).execute()
  File "/opt/pbench-agent/lib/pbench/cli/agent/commands/tools/list.py", line 39, in execute
    for path in self.gen_tools_group_dir(group).glob("*/**"):
  File "/opt/pbench-agent/lib/pbench/agent/base.py", line 124, in gen_tools_group_dir
    return ToolGroup.verify_tool_group(group, pbench_run=self.pbench_run)
  File "/opt/pbench-agent/lib/pbench/agent/tool_group.py", line 45, in verify_tool_group
    raise BadToolGroup(
pbench.agent.tool_group.BadToolGroup: Bad tool group, 'bar': directory /var/lib/pbench-agent/tools-v1-bar does not exist
```

ndokos:
Fixes #2302.

pbench-list-tools produces tracebacks in various benign
situations. This PR add a bunch of functional tests for
various scenarios (but the set is probably not exhaustive yet).

The current set of tests just checks for exit status, so it catches
all the situations where a traceback was produced, but the tests
don't check the output yet.

THere are two fixes:

- in list.py, we catch BadToolGroup exceptions (and log them).

- in base.py, we return a Path as list.py expects, not the
corresponding name.

TBD:

- return error status appropriately and change the functional tests to
expect that: currently, we always return 0.

- the tests should check the program output, at least in some cases:
in particular, calling the command with `--with-option` prints out the
options without the corresponding tool name.

### DIFF
--- a/lib/pbench/cli/agent/commands/tools/base.py
+++ b/lib/pbench/cli/agent/commands/tools/base.py
@@ -15,7 +15,7 @@ class ToolCommand(BaseCommand):
         """List all tools in a given path"""
         return sorted(
             [
-                p.name
+                p
                 for p in path.iterdir()
                 if p.name != "__label__" and p.suffix != ".__noinstall__"
             ]
--- a/lib/pbench/cli/agent/commands/tools/list.py
+++ b/lib/pbench/cli/agent/commands/tools/list.py
@@ -13,7 +13,7 @@ import click
 from pbench.cli.agent import CliContext, pass_cli_context
 from pbench.cli.agent.commands.tools.base import ToolCommand
 from pbench.cli.agent.options import common_options
-
+from pbench.agent.tool_group import BadToolGroup
 
 class ListTools(ToolCommand):
     """ List registered Tools """
@@ -36,35 +36,38 @@ class ListTools(ToolCommand):
             host_tools = {}
             for group in groups:
                 host_tools[group] = {}
-                for path in self.gen_tools_group_dir(group).glob("*/**"):
-                    if self.context.with_option:
-                        host_tools[group][path.name] = [
-                            p.read_text() for p in self.tools(path)
-                        ]
-                    else:
-                        host_tools[group][path.name] = [p for p in self.tools(path)]
+                try:
+                    for path in self.gen_tools_group_dir(group).glob("*/**"):
+                        if self.context.with_option:
+                            host_tools[group][path.name] = [
+                                p.read_text() for p in self.tools(path)
+                            ]
+                        else:
+                            host_tools[group][path.name] = [p for p in self.tools(path)]
+                except BadToolGroup:
+                    self.logger.error("Bad tool group name: %s", group)
             if host_tools:
                 for k, v in host_tools.items():
                     for h, t in v.items():
-                        print("%s: %s %s" % (k, h, t))
+                        print(f"{k}: {h} {t}")
         else:
             # List the groups which include this tool
             group_list = []
             for group in groups:
-                tg_dir = self.gen_tools_group_dir(group)
-                if not tg_dir.exists():
-                    self.logger.error("bad or missing tool group %s", group)
+                try:
+                    tg_dir = self.gen_tools_group_dir(group)
+                    if not tg_dir.exists():
+                        self.logger.error("bad or missing tool group %s", group)
+                        continue
+                except BadToolGroup:
+                    self.logger.error("Bad Tool Group: %s", group)
                     continue
-
                 for path in tg_dir.iterdir():
                     # Check to see if the tool is in any of the hosts.
                     if self.context.name in self.tools(path):
                         group_list.append(group)
             if group_list:
-                print(
-                    "tool name: %s groups: %s"
-                    % (self.context.name, ", ".join(group_list))
-                )
+                print(f"tool name: {self.context.name} groups: {', '.join(group_list)}")
 
 
 def _group_option(f):
--- a/lib/pbench/cli/agent/commands/tools/list.py
+++ b/lib/pbench/cli/agent/commands/tools/list.py
@@ -15,6 +15,13 @@ from pbench.cli.agent.commands.tools.base import ToolCommand
 from pbench.cli.agent.options import common_options
 from pbench.agent.tool_group import BadToolGroup
 
+# FIX:
+# the set of error codes is currently
+# 0
+# 1 = Bad tool group
+# 2 = anything else
+
+
 class ListTools(ToolCommand):
     """ List registered Tools """
 
@@ -24,8 +31,10 @@ class ListTools(ToolCommand):
     def execute(self):
         if not self.pbench_run.exists():
             self.logger.warn("The %s directory does not exist", self.pbench_run)
-            return 0
+            # FIX: do we want to distinguish this from the "bad name" code below.
+            return 2
 
+        status = 0
         # list tools in one or all groups
         if self.context.group:
             groups = self.context.group
@@ -43,9 +52,10 @@ class ListTools(ToolCommand):
                                 p.read_text() for p in self.tools(path)
                             ]
                         else:
-                            host_tools[group][path.name] = [p for p in self.tools(path)]
+                            host_tools[group][path.name] = [p.name for p in self.tools(path)]
                 except BadToolGroup:
                     self.logger.error("Bad tool group name: %s", group)
+                    return 1
             if host_tools:
                 for k, v in host_tools.items():
                     for h, t in v.items():
@@ -58,16 +68,22 @@ class ListTools(ToolCommand):
                     tg_dir = self.gen_tools_group_dir(group)
                     if not tg_dir.exists():
                         self.logger.error("bad or missing tool group %s", group)
+                        status = 1
                         continue
                 except BadToolGroup:
                     self.logger.error("Bad Tool Group: %s", group)
+                    status = 1
                     continue
                 for path in tg_dir.iterdir():
                     # Check to see if the tool is in any of the hosts.
-                    if self.context.name in self.tools(path):
+                    if self.context.name in [tool.name for tool in self.tools(path)]:
                         group_list.append(group)
             if group_list:
                 print(f"tool name: {self.context.name} groups: {', '.join(group_list)}")
+            elif status == 0:
+                # the group list is empty but not because of a bad tool group, so set an "other"
+                status = 2
+        return status
 
 
 def _group_option(f):
--- a/lib/pbench/cli/agent/commands/tools/list.py
+++ b/lib/pbench/cli/agent/commands/tools/list.py
@@ -49,10 +49,12 @@ class ListTools(ToolCommand):
                     for path in self.gen_tools_group_dir(group).glob("*/**"):
                         if self.context.with_option:
                             host_tools[group][path.name] = [
-                                p.read_text() for p in self.tools(path)
+                                {p.name: p.read_text()} for p in self.tools(path)
                             ]
                         else:
-                            host_tools[group][path.name] = [p.name for p in self.tools(path)]
+                            host_tools[group][path.name] = [
+                                p.name for p in self.tools(path)
+                            ]
                 except BadToolGroup:
                     self.logger.error("Bad tool group name: %s", group)
                     return 1
@@ -77,9 +79,18 @@ class ListTools(ToolCommand):
                 for path in tg_dir.iterdir():
                     # Check to see if the tool is in any of the hosts.
                     if self.context.name in [tool.name for tool in self.tools(path)]:
-                        group_list.append(group)
+                        if self.context.with_option:
+                            p = path / self.context.name
+                            group_list.append({group: p.read_text()})
+                        else:
+                            group_list.append(group)
             if group_list:
-                print(f"tool name: {self.context.name} groups: {', '.join(group_list)}")
+                if self.context.with_option:
+                    print(
+                        f"tool name: {self.context.name} groups and options: {group_list}"
+                    )
+                else:
+                    print(f"tool name: {self.context.name} groups: {group_list}")
             elif status == 0:
                 # the group list is empty but not because of a bad tool group, so set an "other"
                 status = 2
--- a/lib/pbench/test/functional/agent/cli/commands/tools/test_tools_list.py
+++ b/lib/pbench/test/functional/agent/cli/commands/tools/test_tools_list.py
@@ -1,31 +1,159 @@
 import pytest
 
 
-def test_pbench_list_tools_help():
-    command = ["pbench-list-tools", "--help"]
-    out, err, exitcode = pytest.helpers.capture(command)
-    assert b"Usage: pbench-list-tools [OPTIONS]" in out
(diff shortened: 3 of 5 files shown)


lib/pbench/cli/agent/commands/tools/base.py
```diff
@@ -15,7 +15,7 @@ def tools(self, path):
         """List all tools in a given path"""
         return sorted(
             [
-                p.name
+                p
```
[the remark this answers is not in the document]
webbnh: For instance, is this change going to break the code at `lib/pbench/cli/agent/commands/tools/list.py` line 67?  (That is, in order for the `--name` option to match, it will need to be a full path, now, right?  I think that's a change in behavior from the user's point of view.)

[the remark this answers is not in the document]
ndokos: There is one more use in the [clear command](https://github.com/distributed-system-analysis/pbench/blob/d6ef508df6bffab1812401ddcafde5fe212172c5/lib/pbench/cli/agent/commands/tools/clear.py#L56) but the `_clear_tools` [method](https://github.com/distributed-system-analysis/pbench/blob/d6ef508df6bffab1812401ddcafde5fe212172c5/lib/pbench/cli/agent/commands/tools/clear.py#L83) concatenates paths: before this change it would concatenate a path with a (simple) name; now it concatenates two paths. Both seem to work: concatenating two paths takes care of the common prefix.

But @webbnh is correct that, *within* `list`, there are more problems: those are not found by the (currently weak) functional tests because I don't check output. So the next step is to: a) fix the expected status for (some of) the functional tests and make sure that the command returns the appropriate error status and b) take the output of the command into account. This should catch the `--with-option` problems as well as the problem that @webbnh indicated (and any others that might arise). My plan is to check the output for all the "successful" cases (as indicated by a 0 return status: if it should fail and it does fail with non-zero status, I consider the test a pass and I won't check the output, but if it succeeds and should succeed, the output should be checked as well. And of course, if it fails with a traceback, the test should fail anyway.

Does that sound reasonable?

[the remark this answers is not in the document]
webbnh: I'm a fan of confirming that we meet end-user expectations....

Since these commands are supposed to produce fairly specific output in success cases, it would be good to check, when the tool reports success, that it produced acceptable output.  Correspondingly, when the tool reports error/failure, it would be good to check that it produced appropriate output (presumably on `stderr` and not on `stdout`) in those cases, as well.

lib/pbench/cli/agent/commands/tools/list.py
```diff
@@ -36,35 +45,45 @@ def execute(self):
...
                         group_list.append(group)
             if group_list:
-                print(
-                    "tool name: %s groups: %s"
-                    % (self.context.name, ", ".join(group_list))
-                )
+                print(f"tool name: {self.context.name} groups: {', '.join(group_list)}")
+            elif status == 0:
+                # the group list is empty but not because of a bad tool group, so set an "other"
+                status = 2
```
portante, line 85: Does this case just mean the specified tool is not found in any group?  The previous code seems to exit successfully on this case.

lib/pbench/cli/agent/commands/tools/list.py
```diff
@@ -36,35 +45,45 @@ def execute(self):
...
-                if not tg_dir.exists():
-                    self.logger.error("bad or missing tool group %s", group)
+                try:
+                    tg_dir = self.gen_tools_group_dir(group)
+                    if not tg_dir.exists():
+                        self.logger.error("bad or missing tool group %s", group)
+                        status = 1
+                        continue
+                except BadToolGroup:
+                    self.logger.error("Bad Tool Group: %s", group)
```
portante, line 74: It seems like we'd never hit line 70, but this exception.  So perhaps we should move that message here, and report missing tool group there with a status code of `2`.

lib/pbench/cli/agent/commands/tools/list.py
```diff
@@ -36,35 +45,45 @@ def execute(self):
...
+                try:
+                    for path in self.gen_tools_group_dir(group).glob("*/**"):
+                        if self.context.with_option:
+                            host_tools[group][path.name] = [
+                                p.read_text() for p in self.tools(path)
+                            ]
+                        else:
+                            host_tools[group][path.name] = [p.name for p in self.tools(path)]
+                except BadToolGroup:
+                    self.logger.error("Bad tool group name: %s", group)
```
portante, line 57: This should probably be the same message as line 70 below.

lib/pbench/cli/agent/commands/tools/list.py
```diff
@@ -13,6 +13,13 @@
 from pbench.cli.agent import CliContext, pass_cli_context
 from pbench.cli.agent.commands.tools.base import ToolCommand
 from pbench.cli.agent.options import common_options
+from pbench.agent.tool_group import BadToolGroup
+
+# FIX:
```
portante, line 18: Seems like this is the `main()` method doc string.

lib/pbench/cli/agent/commands/tools/list.py
```diff
@@ -13,6 +13,13 @@
...
+# the set of error codes is currently
+# 0
+# 1 = Bad tool group
```
portante, line 21: This message should be the text emitted at line 70.

lib/pbench/cli/agent/commands/tools/list.py
```diff
@@ -13,6 +13,13 @@
...
+# 2 = anything else
```
portante, line 22: Not sure this exit code is needed.

portante (commented):
Seems like this is a good start.  It would be great to know why we need to use `p.name` instead of just `p`, describing what problem that change is solving.

lib/pbench/cli/agent/commands/tools/base.py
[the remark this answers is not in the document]
ndokos: Actually, although it is true that `_clear_tools` does not have any problem clearing the tool specified by its path, it does have a problem reporting what it did. So I will submit a fix for that in the next commit, but I have conceptual problems with its specification (and with its implementation as well), so I'll open an issue for it and address it later.

ndokos:
We need to read the file (e.g. to get the options), not just list it. We could go from the name to the path in order to open it, but it seemed much better to me to get the path instead: we can use it to get the name when that is needed and to get the contents of the file when *that* is needed.

lib/pbench/cli/agent/commands/tools/list.py
ndokos: Correct - and it's true the previous code exited successfully, but should it?

lib/pbench/cli/agent/commands/tools/list.py
ndokos: Yes, it should. I'll fix it.

lib/pbench/cli/agent/commands/tools/list.py
ndokos: I don't understand what you mean.

lib/pbench/cli/agent/commands/tools/list.py
ndokos: There are situations other than a bad tool group that seem to require an error status - no?

lib/pbench/cli/agent/commands/tools/base.py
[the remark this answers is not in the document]
ndokos: @webbnh: I'll add tests for the error text in those cases, but ATM the output (normal and error) needs work.  I am going to leave it to the last, fixing the tests first to require the "correct" output and after we are all happy with that, fixing the program to produce that output. For the time being, I'm more worried about the output text in the normal cases containing all the information that is expected (albeit in not quite final form).

lib/pbench/cli/agent/commands/tools/list.py
ndokos: Will do. I plan to change the tests so that they check for the "correct" output first and then make sure that the program produces that output in each case.

lib/pbench/cli/agent/commands/tools/base.py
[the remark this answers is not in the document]
webbnh: Sounds good.  

(You tagged me and concluded with, "Does that sound reasonable?", and I felt like a simple "yes" or "no" response lacked nuance.  :-) )

lib/pbench/test/functional/agent/cli/commands/tools/test_tools_list.py
```diff
@@ -1,31 +1,226 @@
...
-    assert exitcode == 0
-    assert b"tool name: perf groups: default" in out
+class Test_list_tool_no_tools_registered:
+    def test_help(self):
+        command = ["pbench-list-tools", "--help"]
+        out, err, exitcode = pytest.helpers.capture(command)
+        assert b"Usage: pbench-list-tools [OPTIONS]" in out
+        assert exitcode == 0
+
+    def test_no_args(self, monkeypatch, agent_config, pbench_run, pbench_cfg):
```
dbutenhof, line 11: Ugh; these fixtures are very modular and convoluted. I'm surprised there isn't one built on top of `agent_config` that provides the monkeypatched environment variable. There probably should be. And since `agent_config` is already built on top of `pbench_run` and `pbench_cfg`, there's no need to repeat them in the individual tests!

I'd suggest adding the monkeypatch to `agent_config` in the `conftest.py` and then you can simply remove all of the fixtures listed here except `agent_config`, which would simplify all the test cases and make them easier to read.

And, yeah, I'm aware that may be a lot more background rototilling than you'd really intended here, but the fixture dependency structure in these tests is pretty messy.

ndokos: Sure - that should not be a big deal. And since you volunteered, I know where to turn to for help :-)

webbnh: > these fixtures are very modular and convoluted

@dbutenhof, did you omit a "not" from that statement?  (I usually think of modularity, at least when practiced well, as an antidote to convolution....)

dbutenhof: > > these fixtures are very modular and convoluted
> 
> @dbutenhof, did you omit a "not" from that statement? (I usually think of modularity, at least when practiced well, as an antidote to convolution....)

No, they're convoluted to follow because they're excessively modular; each fixture builds on the previous with a small change, and most if not all of the intermediate fixtures are completely useless except as a building block for the next step. And yet each fixture and each test directly references all of them.

lib/pbench/cli/agent/commands/tools/list.py
ndokos: So the  exit codes should be: bad tool group = 1, missing tool group = 2? What about a bad tool name? Or a bad setup (the no pbench_run directory found case)?

lib/pbench/test/functional/agent/cli/commands/tools/test_tools_list.py
webbnh: Huh.  If each fixture references the previous one, then the test should only need to reference the last...right?

lib/pbench/cli/agent/commands/tools/list.py
webbnh: I would vote against having separate codes for each sort of error, unless we expect the invoker to take explicit actions based on the return code.

I think the typical convention is to return `2` to indicate problems invoking the command (like bad or conflicting options) and to return `1` to indicate that the intended effect of the command failed for reasons other than its invocation.

lib/pbench/test/functional/agent/cli/commands/tools/test_tools_list.py
ndokos: @dbutehnof: I left `conftest.py` mostly alone (I just added the `monkeypatch.setenv` to `agent_config`) but eliminated all the extraneous fixtures from the tests.

dbutenhof: Yes; and each fixture need only reference the one before it in the chain.

ndokos: @webbnh: that's what I did. It seems to work too (miraculously enough).

lib/pbench/cli/agent/commands/tools/base.py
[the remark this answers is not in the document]
ndokos: I like nuance :-)

dbutenhof (commented):
Nice improvement; I suspect we could collapse the fixtures more, but this simplifies your tests a lot, and doesn't risk breaking other test classes that share the `conftest.py` ... so it's a good compromise!

ndokos:
Closing this: it is replaced by #2453. Some of the tests from this PR are going to be used in the new one, but the approach is much more surgical in the new PR, trying to make `pbench-list-tools` behave the same way as the 0.69.9 version.

PR #2342, opened 2021-06-08
Ansible server legacy

portante:


server/rpm/pbench-server.spec.j2
```diff
@@ -8,30 +8,11 @@ URL:            http://github.com/distributed-systems-analysis/pbench
...
-#    yum install yum-utils
-#    yum-config-manager --enable rhel-server-rhscl-7-rpms
-#
-# we can't do that in the spec file, because of the yum lock,
-# so they have to be done beforehand.
-Requires: scl-utils, rh-python36
-%else
-Requires: python3 python3-devel gcc
-%endif
+Requires: python38 python38-six python38-urllib3
```
dbutenhof, line 11: Why is this completely different from before? Assuming we don't actually need `python3-devel` or `gcc`, where does the 0.69 server require `six` or `urllib3`??

I assume these are changes you needed to make to get an install last night, but it'd be nice to capture some of the history behind this.

server/ansible/pbench-server-install.yml
```diff
@@ -1,7 +1,9 @@
 ---
+- name: install pbench RPM repo
+  import_playbook: pbench-repo-install.yml
```
dbutenhof, line 3: That's interesting; I'd love to see some history on this. Much of Ansible still baffles me; and in this case, the use of `pbench-repo-install` as both playbook and role probably doesn't help. But, essentially, you're changing the server install playbook from using just the `pbench-repo-install` __role__ to using the entire playbook, presumably to get the "become" behavior? Huh.

server/rpm/pbench-server.spec.j2
```diff
@@ -8,30 +8,11 @@ URL:            http://github.com/distributed-systems-analysis/pbench
 Source0:        pbench-server-%{version}.tar.gz
 Buildarch:      noarch
 
-
-%if 0%{?rhel} == 7
```
dbutenhof, line 12: At this point, I don't disagree and I'm not going to complain, especially given the pressure you were under last night doing this ... I know at one point when we'd discussed the RHEL 7 addition of "native" Python we'd made a deliberate decision to leave 0.69 alone. I'm OK with changing that, but just pointing it out for posterity ... ;-0

dbutenhof (dismissed):

server/bin/unittests
```diff
@@ -270,12 +268,14 @@ function _setup_state {
     let res=res+$?
     cp -a $_tdir/{unittests,getconf.py,pbench*} $_testopt_sat/bin
     let res=res+$?
-    cp -a $_tdir/../../lib/* $_testopt/common/lib
-    let res=res+$?
     cp -a $_tdir/../lib $_testopt
     let res=res+$?
+    cp -a $_tdir/../../lib/* $_testopt/lib
```
webbnh, line 273: This creates a requirement that no file in `$_tdir/../lib/` have the same name as any file in `$_tdir/../../lib/`, or it requires the understanding that any file in `$_tdir/../lib/` will be overridden by a file in `$_tdir/../../lib/` if they have the same name.

At the very least, this strikes me as odd -- shouldn't the "more local" files in `$_tdir/../lib/` be overriding the files in `$_tdir/../../lib/`?

webbnh: Ditto for lines 275 & 277.

tox.ini
```diff
@@ -22,15 +19,19 @@ whitelist_externals =
...
 commands_pre = 
   find {toxinidir} -type f -not -path '{toxinidir}/.tox/*' -path '*/__pycache__/*' -name '*.py[c|o]' -delete
 commands =
   bash -c "./server/bin/unittests {posargs}"
 
 [testenv:agent]
-usedevelop = true
+deps =
+   -r{toxinidir}/test-requirements.txt
+   -r{toxinidir}/agent/test-requirements.txt
```
webbnh, line 34: Does the ordering of these lines matter?  (These are reversed relative to 24-25.)

webbnh (dismissed):
I see a certain amount of "general cleanup" in these changes which might not be warranted for an end-of-life branch.  (Also I think I saw only a small number of actual fixes.)  That said, will the applicable parts of these changes be merged forward into 0.71?

riya-17 (dismissed):

npalaska (dismissed):

server/ansible/roles/pbench-server-activate-httpd-setup-restart/tasks/main.yml
```diff
@@ -25,9 +25,9 @@
...
-    src: "{{ pbench_server_install_dir }}/html/static/"
-    dest: "{{ pbench_dir }}/public_html/static"
-    recursive: yes
-  delegate_to: "{{ inventory_hostname }}"
+#- name: "Copy the web-server bits to the <pbench>/public_html/static directory."
+#  synchronize:
+#    src: "{{ pbench_server_install_dir }}/html/static/"
+#    dest: "{{ pbench_dir }}/public_html/static"
+#    recursive: yes
+#  delegate_to: "{{ inventory_hostname }}"
```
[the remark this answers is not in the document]
dbutenhof: I believe Pete did that part manually for this deployment, but wanted to fix it before submit; this was getting rid of a messy dependency to make the synchronize work that apparently proved problematic on RHEL 8. (If I'm correctly capturing Pete's explanation yesterday.)

This _is_ still a draft PR, so it's in no immediate danger of merging.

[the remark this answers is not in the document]
webbnh: Yup, yup.  I was just putting in a plea for some explanation.  You and I understand the needs and forces at work.  Adding a line or two explaining why the code wasn't outright deleted and/or the criteria which should be met in order for the process to be completed (or undone) would ease the minds of reviewers and future developers, and it doesn't take much more time than adding all those `#`'s.

server/rpm/pbench-server.spec.j2
portante: > Why is this completely different from before?

Because before we used the SCL environment to gain Python3 on RHEL 7 boxes which did not provide it natively.  We may have needed to have `python3-devel` and `gcc` present in the past in order to install certain "pip" installed modules, but it is not clear if that is really the case with SCL.

Now that we are focusing the pbench production server installs only on RHEL 8, we move to Python 3.8.

> Assuming we don't actually need `python3-devel` or `gcc`

There is no assumption here.  I have verified and tested the installation on a RHEL 8 box with Python 3.8 showing that we do not need either of those two RPMs.  That is to say our production environment is running just fine without these two packages installed.

>, where does the 0.69 server require `six` or `urllib3`??

They are implicit requirements that come from the Python3 modules listed in our `requirements.txt` file.  We add these explicit dependencies here so that the local `pip3` install mechanism does not have to install them, finding them already present in the global installation of Python 3.8 `site-packages`.

> I assume these are changes you needed to make to get an install last night, but it'd be nice to capture some of the history behind this.

I can add a comment describing why we add these RPM dependencies.

server/ansible/pbench-server-install.yml
portante: No, to avoid have the `pbench-repo-install` role used in two places.

Another option is to just drop that mechanism entirely from the `pbench-server-install` playbook.

I don't believe it is helpful to have it by default because it presumes everybody wants to use the same COPR repo.

server/rpm/pbench-server.spec.j2
portante: Sure.  I did not want to have to verify this still works for RHEL 7.

Basically, these changes are declaring that we no long support RHEL 7 for the `b0.69` branch.

server/VERSION
```diff
@@ -1 +1 @@
-0.69.5
+0.69.6
```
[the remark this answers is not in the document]
portante: Historically, we have kept them separate.  The mechanism is already in place so I didn't bother to have two separate `VERSION` files with the same number.

server/ansible/roles/pbench-server-activate-httpd-setup-restart/tasks/main.yml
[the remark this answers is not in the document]
portante: I filed this change as a separate patch from the first so that I could just drop it when, or if, we fixed this problem.  But I have not invested any time to find a replacement for this under the `become_user` method.

I'll drop it from this PR, and then just fix it by hand when re-deploying.

server/bin/unittests
portante: This is already the case, because this is what the `server/Makefile` does.

This change was made so that I could feel a bit more confident that the unit tests were operating closer to what the RPM based install does.

tox.ini
portante: Sort of reversed:  The server has its own `requirements.txt` file, and then uses the "common" `test-requirements.txt` file, while the agent does not have a `requirements.txt` file, but does have its own `test-requirements.txt` file now (newly added).  So I was using the order of "common first, specific second" for the agent "test" requirements.  In the case of the server it is just the ordering of the "code" requirements followed by the "test" requirements.

The question is: "why make a change to the agent side now"?

There is no Python 3 code any where in the tree, agent or server side, except a `pprof` tool which requires it.  But that is actually a mistake for the agent to require the Python 3 `ansible` module since it is only used by a tool.  We have a rule that the `pbench-agent` RPM does list RPM dependencies (or Python 3 dependencies, for that matter).

However, because this is a Python 3 module in `requirements.txt`, and the `pbench-agent` RPM does not currently install any modules via `pipe` during RPM installation on the `b0.69` branch (it only removes old `configtools` modules), there is no harm done with these changes.

So the change here is to reflect reality: we need the `ansible` module in our test environment, but not for the operation of the agent code.

portante (commented):
Ready for final review.

dbutenhof (approved):

server/bin/pbench-audit-server.sh
```diff
@@ -772,13 +772,17 @@ fi
...
+    cat << EOF > ${index_content}
 $subj
-EOF
 
-cat ${report} >> ${index_content}
+EOF
+    cat ${report} >> ${index_content}
+else
+    printf "${subj} - Successful audit, nothing to report.\n" > ${index_content}
+fi
```
dbutenhof, line 784: So ... you broke the gold tests here, and didn't update them.

image

dbutenhof (changes requested):

server/bin/pbench-audit-server.sh
portante: Bummer, thanks for pointing that out!

dbutenhof (approved):

server/VERSION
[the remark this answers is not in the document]
webbnh: Understood.  And, there is power in having two separate version numbers, so that we have the flexibility to re-release one without the other.

However, there is also power in having the numbers synchronized when they are being released at the same time.  Yes, there is the down side of having to understand and explain that certain numbers don't exist, but the need for that is rare, and that cost is small compared to being able to say, "This version goes with that version because they have the same number".

So, I just wanted to be sure that you had the option to sync them up, if we wanted to.

server/bin/pbench-audit-server.sh
```diff
@@ -772,13 +772,17 @@ fi
 
 log_finish
 
-# send it
-subj="$PROG.$TS($PBENCH_ENV)"
-cat << EOF > ${index_content}
+subj="$PROG.$TS ($PBENCH_ENV)"
+if [[ -s ${report} ]]; then
+    cat << EOF > ${index_content}
```
webbnh, line 777: Consider using `<<-` where you have `<<` -- this allows you to indent lines 778-780, which is nicer.

webbnh: Alternatively, you could use `<<<`, which would be much nicer, still.  (You might have to use a second command to insert the blank line, but that would still be better than the "here-document".)

server/VERSION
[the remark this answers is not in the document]
portante: And another argument in favor of syncing them is that this change is impacting both lungs.

server/bin/pbench-audit-server.sh
webbnh: Um, wait...you're trying to put the contents of a string variable plus a couple of carriage returns into a file?  That sounds like a job for `echo` (or your favored `printf`)!
```
if [[ -s ${report} ]]; then
    echo -e "$subj\n" > ${index_content}
    cat ${report} >> ${index_content}
```

server/bin/pbench-audit-server.sh
webbnh: You better not fix this!  I just reviewed all those gold files!!!

webbnh (approved):
Just one suggestion (well, _three_ really, but they are all alternatives for one bit of code).

server/VERSION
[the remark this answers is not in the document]
webbnh: > impacting both lungs.

Double pneumonia?

server/bin/gold/test-27.txt
```diff
@@ -50,9 +50,9 @@ len(actions) = 1
...
             "total_chunks": 1,
-            "total_size": 1132
+            "total_size": 1134
```
ndokos, line 55: Why is the total size two bytes bigger (here and in other tests)? I've been going blind trying to discern the difference.

ndokos: Gah - GitHub does not flag it: three vs four newlines between `(unit-test)` and `start ...` on the first line

server/bin/gold/test-29.2.txt
```diff
@@ -30,9 +30,9 @@ len(actions) = 1
...
+            "text": "pbench-audit-server.run-1970-01-01T00:00:42-UTC (unit-test)\n\n\nstart-1970-01-01T00:00:42-UTC: archive hierarchy: /var/tmp/pbench-test-server/test-29.2/pbench/archive/fs-version-001\n\nBad Controllers:\n\t-rw-rw-r--          0 Wed May 27 00:52:20.0000000000 2020 bad-controller\n\nController: ctlr-out-of-range\n\t* No state directories found in this controller directory.\n\t* Unexpected files in controller directory:\n\t  ++++++++++\n\t  .ignore-me\n\t  ----------\n\nend-1970-01-01T00:00:42-UTC: archive hierarchy: /var/tmp/pbench-test-server/test-29.2/pbench/archive/fs-version-001\n",
```
ndokos, line 33: It flags the difference (two vs three newlines) here - why?

ndokos (approved):

server/bin/gold/test-27.txt
webbnh: I believe that GitHub is using Git which is using Diff which does not, by default, highlight within-a-line changes when the lines (or "difference regions") get very large.

I picked up on the pattern by looking at the smaller diffs with the highlights, and then I was able to find them in the long lines...but, yeah, it was a little crazy-making.

server/bin/pbench-audit-server.sh
portante: But this script is going to be obsolete on the `main` branch soon, and we don't need to do this level of improvement on the legacy branch, `b0.69`.

PR #2349, opened 2021-06-08
Rework benchmark results aggregate postprocessing

### LINKED ISSUE #2306: It appears `fio-postprocess` is operating O(N2) or O(N3) processing log files
Place-holder issue based on evidence seen that for a 24-hr fio job, the processing of the resulting log files is O(N2) or O(N3), with an observed 1.4GB of resident memory for the process.

webbnh:
A large fio dataset brought to light significant performance problems in the benchmark results post-processing scripts.  The algorithms they use contain a number of O(n^2) behaviors which, when the dataset becomes large, operate with punishing slowness.  This pull request replaces the ordered list (array) used to contain time-series data for aggregation with a hash-table, indexed by timestamp.  This allows individual data to be accessed in constant or O(ln(n)) time, which allows the problematic dataset to be processed in minutes instead of...days?

The central changes are in `BenchProcessing.pm`.  However, the dataset structures are created in a custom manner by the individual benchmark post-processing scripts.  As such, in addition to `fio-postprocess`, this PR includes changes to `uperf-postprocess`, `dbench-postprocess`, and `trafficgen-postprocess`, as well, since they all use common functions.  In addition, `linpack-postprocess` got caught up in the crossfire, since it references `BenchProcessing.pm` although it doesn't use any of the affected functions.

This PR is a series of commits:
- The first is a point-fix to `fio-postprocess` to avoid re-writing a file during re-post-processing which should only be written after the original run.
- The next is another tweak to `fio-postprocess`, which allows to fail gracefully when sample-sets are missing from the data.
- The next is an addition to the unit tests for fio, which exercises an error case.
- The next commit is a straightforward reformatting of `process-iteration-samples` and `trafficgen-postprocess` -- this commit is intended only to improve the readability of these modules, and it contains minimal functional changes.
- The next commit is the meat of this PR.
- The next is whitespace changes to make the indentation of `uperf-postprocess` consistent.
- The last commit is trivial changes to `linpack-postprocess`.

_You may find it easier to review the changes one commit at a time, to separate the trivial (reformatting) changes from the substantive ones.  Alternatively, it may be helpful to have GitHub ignore whitespace changes._

These changes were hand-tested by re-processing specific benchmark results and by running the Agent unit tests.  For the most part, the reprocessed results are identical to the originals.  There was one value in the `fio-0` unit test which the old code appears to have calculated incorrectly; the new code produces the correct value, which in turn results in a change to the average value which includes it, and these values show up in three of the output files.

The other changes related to testing stem from imprecision in floating point numbers combined with a feature of Perl hashes.  By default, the ordering of values in hash tables in Perl is [intentionally non-deterministic](https://perldoc.perl.org/perlsec#Algorithmic-Complexity-Attacks).  Ordinarily this is not a problem, but because of partial-bit errors in floating point calculations, the non-determinism shows up as variations in the last digit of the floating point values, which causes the gold file comparisons to fail randomly.  To address this, the unit tests now disable the Perl hash seed randomization and bucket perturbation; however, doing so meant that several of the gold files needed to be updated, even though they were otherwise unaffected.

There is one caveat:  the scripts are written in a an adaptive way, such that if there are features missing from the dataset, the scripts simply skip over them when producing the output; thus, while I've tried to test my changes exhaustively, the results of that are only as good as the datasets which I was able to find.

This PR supersedes #2328.
Fixes #2306.


agent/bench-scripts/tests/pbench-fio/test-52.txt
```diff
@@ -0,0 +1,460 @@
++++ Running test-52 pbench-fio
```
portante, line 1: Let's just punt this test entirely.

Ideally, this would be some functional test that is known to take a "long" time.

But we really want the unit tests to finish somewhat quickly.

Future work is going to move all this to the server side in Python3 anyways, so we'll want to solve such functional verification requirements other ways

agent/bench-scripts/pbench-fio
```diff
@@ -859,15 +866,16 @@ function fio_run_benchmark() {
 								fio_run_job "$iteration" "$benchmark_results_dir" "$fio_job_file" "$dev" "$clients"
 							else
 								# if we are only postprocessing, then we might have to untar an existing result
-								pushd $iteration_dir >/dev/null
-								if [ ! -e sample$sample ]; then
-									if [ -e sample$sample.tar.xz ]; then
+								if [ ! -e $iteration_dir/sample$sample ]; then
```
portante, line 869: Some sort of tabs vs spaces thing happening here.

agent/bench-scripts/postprocess/uperf-postprocess
```diff
@@ -117,23 +113,27 @@ if (opendir($dh, $tool_group_dir)) {
 		}
 	}
 } else {
-	print "$script: could not find any directories in $tool_group_dir which matched $tool_label_pattern\n";
+       print STDERR "$script: could not find any directories in $tool_group_dir which matched $tool_label_pattern\n";
```
portante, line 116: We may have a tabs vs spaces thing happening here.

agent/bench-scripts/postprocess/process-iteration-samples
```diff
@@ -277,17 +316,21 @@ foreach my $metric_type (sort keys %iteration) {
...
+					 $iteration{$metric_type}{$metric_name}[$i]{get_label('closest_sample_label')}) =
+						get_avg_stddev(
+							$samples_start_index,
+							@{$iteration{$metric_type}{$metric_name}[$i]{$samples_label}});
 					if ($iteration{$metric_type}{$metric_name}[$i]{get_label('role_label')} and
-						$iteration{$metric_type}{$metric_name}[$i]{get_label('role_label')} eq 'aggregate' and
-						$metric_type eq "throughput" and
-						$metric_name eq $primary_metric and
-						$iteration{$metric_type}{$metric_name}[$i]{get_label('stddevpct_label')} > $max_stddevpct) {
+							$iteration{$metric_type}{$metric_name}[$i]{get_label('role_label')} eq 'aggregate' and
```
portante, line 330: May have a tabs vs spaces thing here.

agent/bench-scripts/postprocess/linpack-postprocess
```diff
@@ -29,105 +30,100 @@ my %throughput;
...
 }
 
 my %client_data;
-
 foreach my $client_hostname_dir (@client_hostname_dirs) {
     $client_data{$client_hostname_dir} = {};
 
     my @client_data_files;
     if (opendir(my $clients_dh, "$dir/clients/$client_hostname_dir")) {
-	@client_data_files = grep(/^linpack\..+$/, (sort readdir($clients_dh)));
```
portante, line 48: We may have a tabs vs spaces thing going on here.

agent/bench-scripts/postprocess/BenchPostprocess.pm
```diff
@@ -10,12 +10,16 @@ use warnings;
...
 
-our @EXPORT_OK = qw(value_exists trim_series get_label create_uid get_length get_uid get_mean remove_timestamp get_timestamps write_influxdb_line_protocol get_cpubusy_series calc_ratio_series calc_sum_series div_series calc_aggregate_metrics calc_efficiency_metrics create_graph_hash get_json);
+our @EXPORT_OK = qw(
+	calc_aggregate_metrics calc_efficiency_metrics calc_ratio_series
+	calc_sum_series convert_samples_hash_to_array create_graph_hash
+	create_uid get_cpubusy_series get_json get_label get_length
+	get_mean_hash get_uid);
 
-my $script = "BenchPostprocess";
+my $script = "BenchPostprocess";  # FIXME:  this initialization doesn't seem to happen....
```
portante, line 22: Should we track this in an issue?

portante (changes requested):
Looking good.  I can't say that I have reviewed every code change in detail yet, but will dive in further when pulled out of `Draft` mode.

agent/bench-scripts/tests/pbench-fio/test-52.txt
webbnh: With the changes in this PR, this test no longer takes very long to run, but it does nearly double the length of time required to run unit tests.  On that basis, plus the fact that it's not a "unit test", I'll pull it from this PR.

However, we should keep it and run it -- it's a valid and valuable functional test.  Where should I put it and how do I hook it up?

agent/bench-scripts/pbench-fio
webbnh: What sort of thing?

As far as I can discern, the unchanged lines in this file are using hard-tabs for indentation, and the changed lines are doing the same.  (Using hard-tabs wouldn't be my preference, but my IDE handles them nicely, displaying a two-column indentation for them.)  I don't _think_ that I should be reformatting the file for this PR....

agent/bench-scripts/postprocess/uperf-postprocess
webbnh: Yep, that should be a tab, to match the surrounding code; will fix.

webbnh: Yup; fixed.  Thanks!

agent/bench-scripts/postprocess/process-iteration-samples
webbnh: Yup; fixed.

agent/bench-scripts/postprocess/linpack-postprocess
webbnh: This file was apparently using a mixed approach to indentation.  I seem to have converted it to use spaces rather than tabs.  Would you prefer that it used tabs like the other files?

agent/bench-scripts/postprocess/BenchPostprocess.pm
webbnh: That's an interesting question.

In light of other conversations that we've had today, rather than draw resources and attention away from other endeavors, I'm inclined to ignore this mystery.

FWIW, `$script` is only referenced on the error paths of `get_cpubusy_series()`, which is somewhat less used than most functions in this module.  Moreover, I'm not up on Perl's scoping rules, but it's possible that the references to it will be satisfied by definition in the calling module rather than the containing one, in which case the lack of initialization here is unimportant.

If you like, I could revert this change....  🙂

agent/bench-scripts/pbench-fio
webbnh: The "thing" happened a few lines later, and I fixed it.

webbnh:
This PR is now ready for review.  (@ndokos and @dbutenhof, I don't know if GitHub sent you a notification....)

agent/bench-scripts/postprocess/trafficgen-postprocess
```diff
@@ -174,50 +212,55 @@ while ( <TXT> ) {
...
 	if ( $trial_param eq 'output_dir' ) {
 		# The output_dir parameter is different per sample; we explicitly omit it
 		# so that when process-iteration-samples runs it will properly merge the
 		# same sample data sets together.
 		next;
 	}
-	if ( $last_trial{'trial_params'}{$trial_param} ) {
+	if (defined($last_trial{'trial_params'}{$trial_param})) {
 		$config{$trial_param} = $last_trial{'trial_params'}{$trial_param}
 	}
```
webbnh, line 227: This change is causing `test-22` to fail:  the previous code would omit any parameters with zero or empty-string values, whereas now they are included in the result.  This pulls in `"max_loss_pct"` and `"min_rate"`, which have zero values, which is probably appropriate (ditto `"trial_gap"`); however, it also pulls in `"pre_trial_cmd"`, which is an empty string, which might not be appropriate.  (The inclusion of `"max_loss_pct"` has the knock-on effect of causing the `"uid"` value to change.)

agent/bench-scripts/postprocess/trafficgen-postprocess
```diff
@@ -174,50 +212,55 @@ while ( <TXT> ) {
...
 }
-$config{get_label('uid_label')} = sprintf("%s-%s", $benchmark_name, $benchmark_version);
+
+my $uid_label = sprintf("%s-%s", $benchmark_name, $benchmark_version);
```
webbnh, line 230: This variable should be renamed to `$uid_value` or similar.  (To be consistent with other usage, a variable named `$uid_label` should contain the value returned from `get_label('uid_label')`.)

agent/bench-scripts/postprocess/trafficgen-postprocess
webbnh: Is this a bug or a feature?  (I.e., should I be reverting this change or updating the `test-22` gold file?)

agent/bench-scripts/tests/pbench-fio/test-52.txt
portante: Is PR #2365 where this landed?

agent/bench-scripts/postprocess/BenchPostprocess.pm
portante: No further changes needed for this.

agent/bench-scripts/postprocess/trafficgen-postprocess
portante: Sounds like a bug to me, but I am wondering if empty strings might be worth excluding.

agent/bench-scripts/tests/pbench-fio/test-52.txt
webbnh: > Is PR #2365 where this landed?

Yup.

portante (commented):
Interrim comments ...

agent/bench-scripts/postprocess/linpack-postprocess
webbnh: My apologies (I thought I did that, but...).

You can have GitHub turn off whitespace differences, which makes that module pretty straightforward to review.

agent/bench-scripts/postprocess/linpack-postprocess
```diff
@@ -29,105 +30,100 @@ my %throughput;
...
+                }
+            } else {
+                printf STDERR "ERROR: Could not open client %s result file %s!\n", $client_hostname_dir, $client_data_file;
+                exit 3;
+            }
+        }
     }
 }
 
 print Dumper \%client_data;
```
ndokos, line 85: Is this debugging output? Should it be conditionally executed? Or commented out?

agent/bench-scripts/postprocess/uperf-postprocess
```diff
@@ -68,7 +64,7 @@ my @cpu_busy;	# each array element contains a hash with
 		# hostname: hostname or IP
 		# role: client, server, host, kvm-host, container-host, etc.
 		# timeseries: a hash of timestamp:value key-pairs
-my %efficiency;
+# my %efficiency;
```
ndokos, line 67: Delete this line?

ndokos (commented):
FIrst pass with a few comments. Still need to review the bulk of the changes: BenchPostprocess.pm, fio-postprocess, process-iterataion-samples, trafficgen-postprocess.

ndokos (dismissed):
Minor nits here and there.

ndokos (commented):
Test-22 (trafficgen) failed and it needs a rebase.

agent/bench-scripts/pbench-fio
```diff
@@ -858,16 +865,17 @@ function fio_run_benchmark() {
 								mkdir -p $benchmark_results_dir
 								fio_run_job "$iteration" "$benchmark_results_dir" "$fio_job_file" "$dev" "$clients"
 							else
-								# if we are only postprocessing, then we might have to untar an existing result
-								pushd $iteration_dir >/dev/null
-								if [ ! -e sample$sample ]; then
-									if [ -e sample$sample.tar.xz ]; then
+								# we are only postprocessing, so we might have to untar an existing result
+								if [ ! -e $benchmark_results_dir ]; then
+									if [ -e ${iteration_dir}/sample${sample}.tar.xz ]; then
```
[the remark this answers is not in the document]
webbnh: I concur with the idea in general; however, in this case, in the next line we change the working directory to `$iteration_dir`, and in the line after that we reference `sample$sample.tar.xz`, so it makes reasonable sense to reference them both at line 870, rather than relying on the combination of them that we previously ensconced in `$benchmark_results_dir`.

agent/bench-scripts/pbench-fio
```diff
@@ -883,7 +891,7 @@ function fio_run_benchmark() {
 							fi
 							if [ $sample_failed -eq 1 ]; then
 								# we need all samples to be good, so bust out of testing samples now
-								break
+								break 2
```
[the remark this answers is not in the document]
webbnh: Yes.  (But, um, I'm afraid that you really don't get Perl -- this is Bash.  🤣)

agent/bench-scripts/postprocess/BenchPostprocess.pm
```diff
@@ -297,256 +176,297 @@ sub get_cpubusy_series {
...
+		my $value = $add_from_ref->{$add_ts_1}{'value'};
+		my $time_diff = $sum_ts - $add_ts_1;
+		if ($time_diff != 0) {
+			my $val_dif = $add_from_ref->{$add_ts_2}{'value'} - $value;
+			$value += $val_dif * $time_diff/($add_ts_2 - $add_ts_1);
+		}
+		$add_to_ref->{$sum_ts}{'value'} += $value;
 	}
-	return -1;
+	return;
```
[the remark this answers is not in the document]
webbnh: The two instances of the code appear to be consistent with each other (I presume they were cut-and-pasted from each other originally, and my changes to them didn't disrupt the original structure); however, the algorithm is very stateful, and the difference between the two instances is inside the loop.  I _could_ implement the abstraction with a callback function parameter which could be used to invoke the custom operation, but even that is a little tricky (and I'd have to figure out how to do it properly in Perl).

So, unless you insist, I'm going to decline.  (Someday, this will all be rewritten or removed, and we can address it then, hopefully with several other significant improvements.)

agent/bench-scripts/postprocess/fio-postprocess
```diff
@@ -58,31 +49,31 @@ my @cpu_busy;	# each array element contains a hash with
 		# hostname: hostname or IP
 		# role: client, server, host, kvm-host, container-host, etc.
 		# timeseries: a hash of timestamp:value key-pairs
-my %efficiency;
-my %latency;    # a hash of latency-type:array-of-latency-values,
+# my %efficiency;
```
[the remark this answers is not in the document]
webbnh: The various `*-postprocess` scripts have at their cores a similar structure, including a series of top-level data structures.  However, not all of the scripts contain code to populate all of the data structures.  As such, I've commented out the declarations which are unused (for sound reasons...), but I've left them in place to maintain the commonality (and in case we should decide to enhance the code).

So, unless you insist, I'll decline.

agent/bench-scripts/postprocess/fio-postprocess
```diff
@@ -58,31 +49,31 @@ my @cpu_busy;	# each array element contains a hash with
...
+		# role: <client or server>
+		# timeseries: a hash of timstamp:value elements
 
 my %throughput; # a hash of throughput-type:array-of-throughput-values,
 		# for example $throughput{MB_sec[0..1]
-my @MB_sec;	# each array element contains a hash with:
-	    	# hostname: <hostname or IP>
-	    	# role: <client or server>
-	    	# timeseries: a hash of timstamp:value key-pairs
+# my @MB_sec;	# each array element contains a hash with:
```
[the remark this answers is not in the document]
webbnh: See [previous comment](https://github.com/distributed-system-analysis/pbench/pull/2349/files?file-filters%5B%5D=.csv&file-filters%5B%5D=.json&file-filters%5B%5D=.opts&file-filters%5B%5D=.sts&file-filters%5B%5D=.txt&file-filters%5B%5D=No+extension#r716930094).

agent/bench-scripts/postprocess/fio-postprocess
```diff
@@ -220,101 +214,91 @@ foreach $client_hostname_dir (@client_hostname_dirs) {
...
-					my %sample;
-					my $index = value_exists(get_label('date_label'), $timestamp_ms, \@samples);
-					if ($index == -1) {
-					    %sample = ( get_label('date_label') => int $timestamp_ms,
-							'data' => { $rwtype => { 'values' => [ int $value ] } },
-							'type' => $log_type );
-					    push(@samples, \%sample);
+					if (exists($samples{$timestamp_ms})) {
+						# Behold, autovivification!
+                        push(@{$samples{$timestamp_ms}{'data'}{$rwtype}{'values'}}, int $value);
```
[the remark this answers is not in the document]
webbnh: Good catch:  fixed (along with one other, mixed tab/space indentation).

agent/bench-scripts/postprocess/fio-postprocess
```diff
@@ -349,31 +333,39 @@ if ( %parameters ) {
 
 calc_aggregate_metrics(\%workload);
 calc_efficiency_metrics(\%workload);
-		
-# pbench graphing uses a specific hash heirarchy, where the dimensions of the hash are:
+
+# Pbench graphing uses a specific hash heirarchy, where the dimensions of the hash are:
```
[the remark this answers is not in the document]
webbnh: Fixed.

agent/bench-scripts/postprocess/uperf-postprocess
webbnh: See [subsequent comment](https://github.com/distributed-system-analysis/pbench/pull/2349#discussion_r716930094).

agent/bench-scripts/postprocess/linpack-postprocess
webbnh: I'm afraid that I have no idea; and, I don't see any obvious tests for this script, so I can't really comment.  But, I will say, since I didn't add this line, I'm not really inclined to remove it....

agent/bench-scripts/postprocess/trafficgen-postprocess
webbnh: So, @portante, this issue is now the sole blocker for this PR (at least, until the next round of code review...).  Should I

- revert this change,
- accept this change and update the gold file, or
- split the difference:  report zeros but omit empty strings?

agent/bench-scripts/postprocess/linpack-postprocess
portante: As a rule, while the tool helps to deal with whitespace changes, it is not something that we should allow.

Whitespace changes should be made in separate commits, and it is easy to do by inserting a commit that only makes whitespace changes via `git rebase -i` to edit them into the series of commits.

portante: Not every tool people use give you that same way of ignoring whitespace.

And in this case, GitHub gratuitously removes the "hide whitespace changes" when switching to conversations that reference whitespace changes.

portante (changes requested):

agent/bench-scripts/postprocess/linpack-postprocess
webbnh: My apologies, again.  I did put the whitespace changes into separate commits for most of the modules which received significant amounts of whitespace change.  Somehow this module escaped that treatment.

Are you satisfied with the final state of this module?  (I.e., if so, could you mark this conversation "resolved"?)

agent/bench-scripts/pbench-fio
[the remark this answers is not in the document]
webbnh: I don't find `$(dirname ${benchmark_result_dir})` to be easier to read or execute than `$iteration_dir`; likewise I don't think that `$(basename ${benchmark_result_dir}).tar.xz` is better than `sample${sample}.tar.xz`.

Under other circumstances, we might be inclined to cache the values returned by `dirname` and `basename` in variables, so, since we already have them, why not use them?

agent/bench-scripts/postprocess/trafficgen-postprocess
portante: Filed issue #2462 to track.

agent/bench-scripts/pbench-fio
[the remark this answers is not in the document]
portante: Okay, that is fair given you are not making this any worse than it was, and hopefully we can drop the code entirely in the future.

agent/bench-scripts/postprocess/linpack-postprocess
portante: Okay, I'll be a softy and let this slide ... ;)

portante (dismissed):

agent/bench-scripts/postprocess/fio-postprocess
```diff
@@ -349,31 +332,39 @@ if ( %parameters ) {
...
+# For example, an html page which has all of the iops_sec metrics graphed may look like:
+#
+# %throughput{'uperf-throughput'}{'iops-per-second'}{client:host1-to-server-host2}{$timestamp_ms}
 #                                                   {client:host1-to-server-host3}{$timestamp_ms}
 #
-# So, for fio, we need to take parts of the data in %workload and transform it in to a new hash
-# The %workload hash has much more information about the workload than needed for graphing, 
-# and the timestamp data is stored in a different way.  So, we just need to do a minor 
-# conversion, provided by BenchPostprocess.pm
+# If we wanted another graph for transactions per second graped in a different
```
ndokos, line 345: "graped" ->"graphed"

ndokos (approved):
One typo in a comment, but otherwise LGTM. I believe I have resolved all previous conversations. Feel free to merge as is, or fix the typo and merge. I  will resolve the outstanding conversation in a little while, after you get a chance to see it.

agent/bench-scripts/postprocess/trafficgen-postprocess
webbnh: It turned out that reverting this change didn't resolve the problem.  🙁

So, I went ahead and added code to exclude fields with empty string values.  Unfortunately, there is considerably subtlety in Perl's type-less scalars, and the JSON formatter that we use disambiguates them based on how they were _last **used**_, so checking to see that a (numeric) value is a not an empty _string_ caused several of the numeric fields to then be dumped **as strings**.

So, I had to add logic to detect numeric values based on their contents, and, for things which looked numeric, to re-numerify them.  Unfortunately, JSON boolean values appear to be numeric, but we want them formatted as neither numbers nor strings, so I had to add further logic to ensure that they come through unscathed.

Nevertheless, after much wrangling, I am satisfied with the result.  It does add a few fields to the output, and the UID does change, but I think the latter is an intended result, and the former are not bad things.  (And, it aligns well with the `b0.69` results, although I'm not sure what we did in 0.71 which requires extra logic.)

webbnh:
OK, I've made all the requested changes and squashed them into their respective commits (I'm intending to merge this branch via "Rebase and merge" and keep the remaining/existing commits separate).  So, once Mr. Jenkins blesses this, it should be good to go.

Could I please get two approvals, @ndokos, @portante, @dbutenhof?

ndokos (approved):

portante (approved):

PR #2350, opened 2021-06-10
Add a re-unpack cronjob that avoids indexing

portante:


dbutenhof (dismissed):

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,244 @@
...
+                tbs_existing += 1
+                print(
+                    f"Exists {tbs_existing:d} ({tbs_restored:d} restored) of (tbs_cnt:d): {a_tb}",
+                    flush=True,
+                )
+                continue
+
+            # Create controller directory if it doesn't exist
+            ctrl_p = archive_p / ctrl
+            if not options.dry_run:
```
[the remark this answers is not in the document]
npalaska: I think on line 168 it does print the controller directory indicating implicitly that dir will be created

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,244 @@
...
+                        int(match.group(1)),
+                        int(match.group(2)),
+                        int(match.group(3)),
+                        int(match.group(4)),
+                        int(match.group(5)),
+                        int(match.group(6)),
+                    )
+                    tb = Path(entry.path[:-4])
+                    if not tb.exists():
+                        print(f"what? {entry.path}")
```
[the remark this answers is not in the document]
portante: Fixed.

server/bin/pbench-restore-and-unpack-tarballs.py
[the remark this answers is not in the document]
portante: Yes, that is true, @npalaska, but I think was Dave is pointing out is that controller directories are really only created the first time one is encountered that doesn't currently exist.  So I modified the code to implement that behavior.

server/bin/gold/test-2.txt
```diff
@@ -89,7 +89,10 @@ len(actions) = 1
 ]
 --- Finished pbench-dispatch (status=0)
 +++ Running pbench-unpack-tarballs small
-pbench-unpack-tarballs: Bad INCOMING=/var/tmp/pbench-test-server/test-2/pbench/public_html/incoming
+pbench-unpack-tarballs-small: Bad INCOMING=/var/tmp/pbench-test-server/test-2/pbench/public_html/incoming
+--- Finished pbench-unpack-tarballs (status=1)
```
dbutenhof, line 93: That's interesting: why did this change from `pbench-unpack-tarballs` to `pbench-unpack-tarballs-small`?? Or rather, it looks like it ought to have been in the first place, but why did it change? Ah; you moved the `PROG` renaming up before the config errors were detected. Cool.

server/bin/gold/test-24.txt
```diff
@@ -15,7 +15,6 @@ MAILFROM=pbench@pbench.example.com
 * * * * *  flock -n /var/tmp/pbench-test-server/test-24/opt/pbench-server/lib/locks/pbench-unpack-tarballs-medium.lock /var/tmp/pbench-test-server/test-24/opt/pbench-server/bin/pbench-unpack-tarballs medium
 * * * * *  flock -n /var/tmp/pbench-test-server/test-24/opt/pbench-server/lib/locks/pbench-unpack-tarballs-large.lock /var/tmp/pbench-test-server/test-24/opt/pbench-server/bin/pbench-unpack-tarballs large
 * * * * *  flock -n /var/tmp/pbench-test-server/test-24/opt/pbench-server/lib/locks/pbench-unpack-tarballs-huge.lock /var/tmp/pbench-test-server/test-24/opt/pbench-server/bin/pbench-unpack-tarballs huge
-* * * * *  flock -n /var/tmp/pbench-test-server/test-24/opt/pbench-server/lib/locks/pbench-unpack-tarballs.lock /var/tmp/pbench-test-server/test-24/opt/pbench-server/bin/pbench-unpack-tarballs "" re-unpack
```
dbutenhof: How is it that your latest commit is removing this entry?

dbutenhof (approved):

server/bin/gold/test-24.txt
portante: Basically it seems bad to enable the "re-unpack" cron job by default, since that is not an automated process on the system.

dbutenhof: > Basically it seems bad to enable the "re-unpack" cron job by default, since that is not an automated process on the system.

Ah: indeed, that's a very good point!

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,243 @@
...
+                to_unpack_dir.mkdir(exist_ok=True)
+            to_unpack = to_unpack_dir / tb.name
+            print(f"ln -s {a_tb} {to_unpack}", flush=True)
+            if not options.dry_run:
+                to_unpack.symlink_to(a_tb)
+
+            tbs_restored += 1
+
+            print(
+                f"\nRestored {tbs_restored:d} ({tbs_existing:d} existing) of ({tbs_cnt:d}): {a_tb}\n",
```
webbnh, line 209: I find this pair of prints somewhat confusing (the other one being line 150).  I suggest a something like
```
                "\nFile {:d} of {:d}:  {} restored\n".format(
                tbs_restored + tbs_existing, tbs_cnt, a_tb);
```

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,243 @@
...
+            cp_cmd = f"cp -a {tb} {tb}.md5 {ctrl_p}/"
+            print(f"\n{cp_cmd}", flush=True)
+            if not options.dry_run:
+                cp = subprocess.run(cp_cmd, shell=True, stderr=subprocess.STDOUT)
+                if cp.returncode != 0:
+                    print(
+                        f"FAILURE: cp command: '{cp_cmd}': {cp.returncode}, {cp.stdout!r}",
+                        file=sys.stderr,
+                    )
+                    sys.exit(1)
```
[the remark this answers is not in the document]
webbnh: Also, line 217.

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,243 @@
...
+                file=sys.stderr,
+            )
+            raise
+
+    end = pbench._time()
+    if options.dry_run:
+        print(
+            f"Restored {tbs_restored:d} tar balls from backup volume,"
+            f" {backup} (ended at: {end}, elapsed ({end - start:0.2f})"
+        )
```
webbnh, line 224: Why is this printed only for dry runs?  (And, why bother timing it?...won't t take _much_ longer if we actually do the work?).  Why note only the restored files and not the existing count or the total count?

And, given that this is only for dry runs, again, it would be better if the message made clear that it didn't actually do anything.

server/bin/gold/test-18.txt
```diff
@@ -0,0 +1,223 @@
...
+
+
+cp -a /var/tmp/pbench-test-server/test-18/pbench-local/archive.backup/ctrl-00/tarball_config-00_1970.01.01T00.00.40.tar.xz /var/tmp/pbench-test-server/test-18/pbench-local/archive.backup/ctrl-00/tarball_config-00_1970.01.01T00.00.40.tar.xz.md5 /var/tmp/pbench-test-server/test-18/pbench/archive/fs-version-001/ctrl-00/
+md5sum --check /var/tmp/pbench-test-server/test-18/pbench/archive/fs-version-001/ctrl-00/tarball_config-00_1970.01.01T00.00.40.tar.xz.md5
+ln -s /var/tmp/pbench-test-server/test-18/pbench/archive/fs-version-001/ctrl-00/tarball_config-00_1970.01.01T00.00.40.tar.xz /var/tmp/pbench-test-server/test-18/pbench/archive/fs-version-001/ctrl-00/BACKED-UP/tarball_config-00_1970.01.01T00.00.40.tar.xz
+ln -s /var/tmp/pbench-test-server/test-18/pbench/archive/fs-version-001/ctrl-00/tarball_config-00_1970.01.01T00.00.40.tar.xz /var/tmp/pbench-test-server/test-18/pbench/archive/fs-version-001/ctrl-00/TO-RE-UNPACK/tarball_config-00_1970.01.01T00.00.40.tar.xz
+
+Restored 3 (0 existing) of (3): /var/tmp/pbench-test-server/test-18/pbench/archive/fs-version-001/ctrl-00/tarball_config-00_1970.01.01T00.00.40.tar.xz
```
webbnh, line 25: Is there at test which exercises the pre-existing tarball check code?

server/bin/gold/test-18.txt
```diff
@@ -0,0 +1,223 @@
...
+            "doctype": "status",
+            "name": "pbench-audit-server",
+            "text": "pbench-audit-server.run-1970-01-01T00:00:42-UTC(unit-test)\n",
+            "total_chunks": 1,
+            "total_size": 59
+        },
+        "_type": "pbench-server-reports"
+    }
+]
+--- Finished unit test audit (status=0)
```
webbnh, line 88: What produces this report...and, more to the point, why is it being produced by a dry run?

server/bin/pbench-unpack-tarballs.sh
```diff
@@ -37,6 +37,21 @@
...
+    export PROG="pbench-re-unpack-tarballs"
+fi
+if [[ -z "${BUCKET}" || "${BUCKET}" == "none" ]]; then
+    export PROG="${PROG}"
+else
+    # We rename the PROG to include the bucket since we don't want to conflict
+    # with other unpack tar balls running using different buckets at the same
+    # time.
+    export PROG="${PROG}-${BUCKET}"
+fi
```
webbnh, line 53: What exactly does line 47 accomplish?  (I would really appreciate a code comment there....)

And, are we sure that we don't want to combine lines 45 & 46 into an `elif`?

server/bin/pbench-unpack-tarballs.sh
```diff
@@ -37,6 +37,21 @@
 # symlink from $ARCHIVE/TODO to $ARCHIVE/TO-COPY-SOS.
 
 
+BUCKET="${1}"
+PIPELINE="${2}"
+
+if [[ "${PIPELINE}" == "re-unpack" ]]; then
+    export PROG="pbench-re-unpack-tarballs"
+fi
+if [[ -z "${BUCKET}" || "${BUCKET}" == "none" ]]; then
```
webbnh, line 46: Rather than doing this compound test (which we do twice, now -- see line 68), how about handling it at line 40:
```
BUCKET=${1:-none}
```
Then lines 40 and 68 are reduced to just
```
if [[ ${BUCKET} == "none" ]]; then
```

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,244 @@
...
+
+     b. Copy tar ball to archive tree, and .md5 and run md5sum check
+
+        * Create controller directory if it doesn't exist
+
+        * Create TO-RE-UNPACK and BACKED-UP directories if they don't exist
+
+     c. Create symlink to tar ball in TO-RE-UNPACK directory
+
+  4. After processing 10 tar balls, wait for # of TO-RE-UNPACK to drop to zero
```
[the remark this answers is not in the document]
webbnh: Come to think of it, I don't see the implementation of this -- did I miss it, or did you decide not to do it?

webbnh (changes requested):
It is too much -- I will sum up:
 - In `pbench-restore-and-unpack-tarballs.py`, the use of underscore-prefixed identifiers from another module is, AFAIK, a violation; however, I don't think those functions are being put to a positive use, so perhaps you could just remove the calls.
 - Shouldn't this new command be implemented using Click for argument handling?
 - I vote against returning individual error codes for various failures without good justification, which I do not see here.  Exit codes `0`, `1`, and `2` should suffice.
 - Do you really want problems with individual files to cause the entire command to exit?  (I could see giving up after, say, 10 files or so, or, say, more than 5% of the target total, but dying on the first seems hypersensitive and possibly counterproductive (imagine that it happens 20 hours into a three-week run....)

The rest is the usual pile of suggestions, pointed questions, and nits.

server/bin/pbench-unpack-tarballs.sh
npalaska: line 46* and 68

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,243 @@
...
+            if not c_entry.is_dir(follow_symlinks=False):
+                continue
+            # We have a controller directory.
+            with os.scandir(c_entry.path) as controller_scan:
+                for entry in controller_scan:
+                    if entry.name.startswith(".") and entry.is_dir(
+                        follow_symlinks=False
+                    ):
+                        continue
+                    if entry.is_dir(follow_symlinks=False):
```
riya-17, line 65: Why are we checking entry.is_dir(follow_symlinks=False): 2 times?

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,243 @@
...
+                    # object.
+                    tb_dt = datetime(
+                        int(match.group(1)),
+                        int(match.group(2)),
+                        int(match.group(3)),
+                        int(match.group(4)),
+                        int(match.group(5)),
+                        int(match.group(6)),
+                    )
+                    tb = Path(entry.path[:-4])
```
riya-17, line 80: Path(entry.path[0 : -len(".md5")])

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,244 @@
...
+    backup_p = Path(backup).resolve(strict=True)
+
+    if not backup_p.is_dir():
+        logger.error(
+            "The configured pbench-backup-dir directory, {}, is not a valid directory",
+            backup,
+        )
+        return 6
+
+    start = pbench._time()
```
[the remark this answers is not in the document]
riya-17: Maybe, config.TS?

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,244 @@
+#!/usr/libexec/platform-python
+# -*- mode: python -*-
+
+"""Pbench Tar Balls & Unpack
+
+The goal is to find all tar balls that are in the backup tree and restore them
+to the archive tree, unpacking as appropriate (left to the unpack process to
```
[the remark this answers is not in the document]
ndokos: Yes - in the most recent incarnations, it is/was NFS-mounted from a different machine(in fact, the pbench server and the server housing the backup were not even geographically close).

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,244 @@
...
+
+  3. For each tar ball
+
+     a. Check if tar ball exists in archive tree, if so, skip
+
+     b. Copy tar ball to archive tree, and .md5 and run md5sum check
+
+        * Create controller directory if it doesn't exist
+
+        * Create TO-RE-UNPACK and BACKED-UP directories if they don't exist
```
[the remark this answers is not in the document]
ndokos: Yes, the TO-* state directories signify action states: some cron job should pick it up from there and deposit it somewhere else. Other directories signify final states: `BACKED-UP` is one of those.

server/bin/pbench-restore-and-unpack-tarballs.py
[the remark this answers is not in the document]
ndokos: That raised a flag for me as well.

server/bin/gold/test-18.txt
```diff
@@ -0,0 +1,223 @@
...
+                "hostname": "example.com",
+                "pid": 42,
+                "user_id": 44,
+                "version": ""
+            },
+            "@timestamp": "1970-01-01T00:00:42",
+            "chunk_id": 1,
+            "doctype": "status",
+            "name": "pbench-re-unpack-tarballs",
+            "text": "pbench-re-unpack-tarballs.run-1970-01-01T00:00:42-UTC(unit-test) - w/ 0 errors\nProcessed 0 result tar balls, 0 successfully, 0 warnings, 0 errors, and 0 duplicates\n\n",
```
ndokos, line 51: Didn't it process three tar balls? NM, it was a dry run... assuming IUC.

server/bin/gold/test-5.2.txt
```diff
@@ -118,6 +118,36 @@ len(actions) = 1
...
+                "hostname": "example.com",
+                "pid": 42,
+                "user_id": 44,
+                "version": ""
+            },
+            "@timestamp": "1970-01-01T00:00:42",
+            "chunk_id": 1,
+            "doctype": "status",
+            "name": "pbench-re-unpack-tarballs",
+            "text": "pbench-re-unpack-tarballs.run-1970-01-01T00:00:42-UTC(unit-test) - w/ 0 errors\nProcessed 0 result tar balls, 0 successfully, 0 warnings, 0 errors, and 0 duplicates\n\n",
```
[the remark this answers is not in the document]
ndokos: I think this is the re-unpack report: nothing was restored so nothing was processed. The previous scenario involved unpacking "new" tar balls.

This was an existing test which (because of a config file change) is now running one extra cron job: the cron job produced some output (so a change to the gold file was needed) but it didn't do anything (because there was nothing for it to do).

ndokos (changes requested):
My peeve is the early termination of the loop when a problem with a single tar ball is encountered.

server/bin/pbench-restore-and-unpack-tarballs.py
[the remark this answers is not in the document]
webbnh: Thanks, Nick.  However, I was mostly promptly Peter to update the docs here.  😁

server/bin/gold/test-18.txt
webbnh: Yes, I believe that '18 is a dry run...which puts another facet on an issue that I pointed out earlier:  the dry run should have indicated that it _would have_ processed three (or whatever) tarballs...but didn't.  And, the fact that it _didn't_ suggests that it shouldn't be making a report about it.

server/bin/gold/test-5.2.txt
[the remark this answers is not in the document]
webbnh: Nick, that all makes good sense in context, but I still have to ask, is this what was _supposed_ to have happened?  Is this this test that we _wanted_ it to do?

Are we really going to run a periodic cron job which, once we've recovered from the power shutdown, will never actually do anything (until the next time someone wants to re-unpack a tarball)?  Seems weird.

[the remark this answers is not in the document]
ndokos: I think the ratinonale (such as it is) is that  the (legacy) unittests script is supposed to test all of the cron jobs that could potentially be enabled on a real server (about a dozen `_run_allscripts` tests do that - the rest are more specific). Some tests override the  global `state/config/pbench-server.cfg` with their own, but none of them override the cron settings, so those are inherited from the "real" `pbench-server-default.cfg` file. So that set of tests runs all of the cron job scripts, including new ones that are added to the `pbench-server-default.cfg` file: I think that's a good thing. Of course, targeted tests for the new functionality have to be added as well, but at least we know (for some version of "know") that we are not going to keel over and die immediately if the script is added.

As for the second part, I think the ultimate answer is that we not going to run *any* cron jobs: the service that will replace all this is going to figure out what needs to be done (based on some currently unspecified input) and dispatch the appropriate service module instantly. If nothing is ever restored from backup, then the service module will remain unused, but even after we've recovered from the current predicament, a restore might not be such an uncommon occurrence.

server/bin/pbench-restore-and-unpack-tarballs.py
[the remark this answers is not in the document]
webbnh: > you cannot subtract the beginning and end TS
Because the value doesn't change once it's initialized?...oh, well -- there are other ways of doing timing...if we really need to.

server/bin/pbench-restore-and-unpack-tarballs.py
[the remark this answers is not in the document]
webbnh: I pointed out to him in our 1x1 today that, since we don't unpack tarballs more than a year old, this problem will eventually fix itself....  😁

server/bin/pbench-restore-and-unpack-tarballs.py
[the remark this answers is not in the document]
portante: So this is the old way of fixing "time" in unit tests.  You can see this in `pbench-cull-unpacked-tarballs` and `pbench-reindex`.

If we want to fix this, let's fix this on `main`, but I don't think we should be fixing this on `b0.69`, or in this PR.

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,244 @@
...
+    parser.add_argument(
+        "-D",
+        "--dry-run",
+        dest="dry_run",
+        action="store_true",
+        default=False,
+        help="Perform a dry-run only",
+    )
+    parser.set_defaults(cfg_name=os.environ.get("_PBENCH_SERVER_CONFIG"))
+    parsed = parser.parse_args()
```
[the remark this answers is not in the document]
portante: Not in the `b0.69` tree.

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+def main(options):
+    if not options.cfg_name:
+        print(
+            f"{_NAME_}: ERROR: No config file specified; set"
+            " _PBENCH_SERVER_CONFIG env variable",
+            file=sys.stderr,
+        )
+        return 1
```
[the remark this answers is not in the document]
portante: This was a hold-over from `pbench-cull-unpacked-tar-balls.py`.

server/bin/pbench-restore-and-unpack-tarballs.py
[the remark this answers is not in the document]
portante: Yes, this was dropped.

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,244 @@
...
+from pbench import (
+    PbenchConfig,
+    BadConfig,
+    get_pbench_logger,
+)
+
+
+_NAME_ = "pbench-restore-and-unpack-tarballs"
+
+tb_pat_r = r"\S+_(\d\d\d\d)[._-](\d\d)[._-](\d\d)[T_](\d\d)[._:](\d\d)[._:](\d\d)\.tar\.xz\.md5"
```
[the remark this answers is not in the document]
portante: For the `b0.69` branch yes, because this is the same pattern used by `pbench-cull-unpack-tar-balls.py`.

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,244 @@
...
+                continue
+            # We have a controller directory.
+            with os.scandir(c_entry.path) as controller_scan:
+                for entry in controller_scan:
+                    if entry.name.startswith(".") and entry.is_dir(
+                        follow_symlinks=False
+                    ):
+                        continue
+                    if entry.is_dir(follow_symlinks=False):
+                        continue
```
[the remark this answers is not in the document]
portante: We want to skip all `.*` files in the controller directory.

server/bin/gold/test-0.1.txt
```diff
@@ -10,6 +10,9 @@ pbench-dispatch: Bad LOGSDIR=/var/tmp/pbench-test-server/test-0.1/pbench-local/l
 +++ Running pbench-unpack-tarballs small
 pbench-unpack-tarballs: Bad LOGSDIR=/var/tmp/pbench-test-server/test-0.1/pbench-local/logs (config file /var/tmp/pbench-test-server/test-0.1/opt/pbench-server/lib/config/pbench-server.cfg)
 --- Finished pbench-unpack-tarballs (status=1)
++++ Running pbench-unpack-tarballs none re-unpack
+pbench-unpack-tarballs: Bad LOGSDIR=/var/tmp/pbench-test-server/test-0.1/pbench-local/logs (config file /var/tmp/pbench-test-server/test-0.1/opt/pbench-server/lib/config/pbench-server.cfg)
```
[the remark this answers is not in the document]
portante: Existing test infrastructure, won't change in the `b0.69` branch.

server/bin/gold/test-5.2.txt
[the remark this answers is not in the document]
portante: > Nick, that all makes good sense in context, but I still have to ask, is this what was _supposed_ to have happened? Is this the test that we _wanted_ it to do?

Yes.

> Are we really going to run a periodic cron job which, once we've recovered from the power shutdown, will never actually do anything (until the next time someone wants to re-unpack a tarball)? Seems weird.

No.  We don't have the "re-unpack" cronjob enabled by default.  So one would have to update the configuration file to add that step, and then rebuild the cronjob file to add it.

server/bin/gold/test-5.1.txt
```diff
@@ -466,6 +499,14 @@ run-1970-01-01T00:00:42-UTC: Processed 0 result tar balls, 0 successfully (0 par
 1970-01-01T00:00:42.000000 DEBUG pbench-index.pbench-index main -- pbench-index.run-1970-01-01T00:00:42-UTC: starting
 1970-01-01T00:00:42.000000 INFO pbench-index.pbench-index main -- No tar balls found that need processing
 ----- pbench-index/pbench-index.log
++++++ pbench-re-unpack-tarballs/pbench-re-unpack-tarballs.error
+----- pbench-re-unpack-tarballs/pbench-re-unpack-tarballs.error
++++++ pbench-re-unpack-tarballs/pbench-re-unpack-tarballs.log
+run-1970-01-01T00:00:42-UTC
+run-1970-01-01T00:00:42-UTC: Processed 0 tarballs
+1970-01-01T00:00:42.000000 DEBUG pbench-re-unpack-tarballs.indexer update_templates -- done templates (start ts: 1970-01-01T00:00:42-UTC, end ts: 1970-01-01T00:00:42-UTC, duration: 0.00s, successes: 1, retries: 0)
+1970-01-01T00:00:42.000000 DEBUG pbench-re-unpack-tarballs.report post_status -- posted status (start ts: 1970-01-01T00:00:42-UTC, end ts: 1970-01-01T00:00:42-UTC, duration: 0.00s, successes: 1, duplicates: 0, failures: 0, retries: 0)
```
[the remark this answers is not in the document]
portante: Agreed, but is the nature of this test: nothing happened. =)

server/bin/pbench-restore-and-unpack-tarballs.py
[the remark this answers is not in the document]
portante: Bug that is missing the `stderr` ref; added a comment for why we have to set the current working directory.

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,243 @@
...
+                    print(
+                        f"FAILURE: cp command: '{cp_cmd}': {cp.returncode}, {cp.stdout!r}",
+                        file=sys.stderr,
+                    )
+                    sys.exit(1)
+            # ... and run md5sum check against it
+            md5sum_cmd = f"md5sum --check {a_tb}.md5"
+            print(md5sum_cmd, flush=True)
+            if not options.dry_run:
+                cp = subprocess.run(md5sum_cmd, shell=True, cwd=str(ctrl_p))
```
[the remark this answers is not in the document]
portante: Bug that is missing the `stderr` ref; added a comment for why we have to set the current working directory.

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,243 @@
...
+                    print(
+                        f"FAILURE: md5sum command: '{md5sum_cmd}': {cp.returncode}, {cp.stdout!r}",
+                        file=sys.stderr,
+                    )
+                    sys.exit(1)
+
+            # Create the symlink recording the tar ball is already backed up
+            backed_up_dir = ctrl_p / "BACKED-UP"
+            if not options.dry_run:
+                backed_up_dir.mkdir(exist_ok=True)
```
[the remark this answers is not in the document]
portante: No because if we are restoring hundreds of tar balls in one directory we'll see hundreds of directory creations, and that made it a pain to wade through the output to verify it was working.

server/bin/pbench-restore-and-unpack-tarballs.py
portante: Because I wanted to see how many were acted upon separate from what was NOT acted upon.

server/bin/gold/test-18.txt
portante: This is the standard "audit server" report run by the harness after each test.

server/bin/gold/test-18.txt
portante: This is expected.  In the case of a dry-run, `pbench-restore-and-unpack-tarballs` does not take any action, so the cron job, `pbench-re-unpack-tarballs` does not process any thing.

server/bin/pbench-restore-and-unpack-tarballs.py
portante: Doesn't seem worth fixing for the `b0.69` branch, where we'll never use something other than a `.md5` file.

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,243 @@
...
+        return 1
+
+    try:
+        config = pbench.PbenchConfig(options.cfg_name)
+    except pbench.BadConfig as e:
+        print(f"{_NAME_}: {e}", file=sys.stderr)
+        return 2
+
+    archive = config.ARCHIVE
+    archive_p = Path(archive).resolve(strict=True)
```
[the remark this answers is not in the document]
portante: If this was a PR on the `main` branch that would be true.

server/bin/pbench-restore-and-unpack-tarballs.py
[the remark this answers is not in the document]
portante: Yes, as this is one of the few direct command line tools on the pbench-server that an admin would invoke directly.  So we want it to fail for the unexpected in case there is a wider problem to consider.

This was very useful as I worked through getting the server working properly, and making sure the unexpected errors were corrected before further actions were taken.

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,243 @@
...
+    backup = config.conf.get("pbench-server", "pbench-backup-dir")
+    backup_p = Path(backup).resolve(strict=True)
+
+    if not backup_p.is_dir():
+        print(
+            "The configured pbench-backup-dir directory, {backup}, is not a"
+            " valid directory",
+            file=sys.stderr,
+        )
+        return 6
```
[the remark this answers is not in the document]
portante: Only if we were on the `main` branch.

server/bin/pbench-restore-and-unpack-tarballs.py
portante: Fixed.

portante:
> * In `pbench-restore-and-unpack-tarballs.py`, the use of underscore-prefixed identifiers from another module is, AFAIK, a violation; however, I don't think those functions are being put to a positive use, so perhaps you could just remove the calls.

Legacy unit test "ism", since we're on the `b0.69` not changing.

> * Shouldn't this new command be implemented using Click for argument handling?

No `Click` in `b0.69`.

> * I vote against returning individual error codes for various failures without good justification, which I do not see here.  Exit codes `0`, `1`, and `2` should suffice.

Okay.  Fixed.

> * Do you really want problems with individual files to cause the entire command to exit?  (I could see giving up after, say, 10 files or so, or, say, more than 5% of the target total, but dying on the first seems hypersensitive and possibly counterproductive (imagine that it happens 20 hours into a three-week run....)

Yes.  The `pbench-restore-and-unpack-tarballs` command is a "@ the command line run" utility, and that behavior was very useful to me as I *carefully* restored data from backups.  This allowed me to correct situations as they came up without worry that I was running a bull in a china shop.
 
> 
> The rest is the usual pile of suggestions, pointed questions, and nits.

Thanks!

Still need to address a few more.

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,243 @@
...
+        print(f"Restoring tar balls from backup volume, {backup} (started at: {start})")
+
+    gen = gen_list(backup_p)
+    tbs_sorted = sorted(
+        [
+            (tb_dt, controller_name, tar_ball_name)
+            for tb_dt, controller_name, tar_ball_name in gen
+        ],
+        reverse=True,
+    )
```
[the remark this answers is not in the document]
portante: This code is quite specific to the `b0.69` branch.  Porting this forward to the `main` branch would only be done after we have a way to backup the ownership information of the tar ball.

As such, I'd like not to pursue too many more changes given the ability to verify this via tests.

server/bin/gold/test-18.txt
portante: No, I'll enhance 18/19 to add that.

server/bin/pbench-unpack-tarballs.sh
portante: > And, are we sure that we don't want to combine lines 45 & 46 into an `elif`?

Yes, they are orthogonal considerations.

The update to line 40 makes this easier to shrink the logic, though.

server/bin/unittests
```diff
@@ -101,8 +106,9 @@ function _run_allscripts {
     _run pbench-copy-sosreports
     _run pbench-index
     _run pbench-index --tool-data
-    # These four are independent, running periodically to accomplish
+    # This next set are independent, running periodically to accomplish
```
riya-17, line 109: s/are/is ?

riya-17 (approved):
Looks Good

server/bin/pbench-restore-and-unpack-tarballs.py
webbnh: In the updated code, the message "Exists 1 (3 restored) of (4):" doesn't read well.

It seems like you have two (or three) concepts that you're trying to present:
1. The name of the file just dealt with
2. The progress against the list
3. The running totals of dispositions

The grand totals of dispositions is presented at the end, so perhaps the running totals don't need to be presented along the way?  If you do want to see them run, that's fine, but I suggest adding a few more characters to make things clear:
```
                "\nFile {:d} of {:d} ({:d} restored, {:d} existing, so far), restored:  {}\n".format(
                tbs_restored + tbs_existing, tbs_restored, tbs_existing, tbs_cnt, a_tb);
```
This gives you the file count so far, the total count (so you know where you are), the count of restored and existing files (so you know what's been done so far), and the status and name of the current file.  (And, the poor human reader doesn't have to figure out which two quantities are the running totals and add them together himself.)

server/bin/gold/test-18.txt
```diff
@@ -6,25 +6,26 @@ md5sum --check /var/tmp/pbench-test-server/test-18/pbench/archive/fs-version-001
...
 ln -s /var/tmp/pbench-test-server/test-18/pbench/archive/fs-version-001/ctrl-00/tarball_config-00_1970.01.01T00.00.40.tar.xz /var/tmp/pbench-test-server/test-18/pbench/archive/fs-version-001/ctrl-00/TO-RE-UNPACK/tarball_config-00_1970.01.01T00.00.40.tar.xz
 
-Restored 3 (0 existing) of (3): /var/tmp/pbench-test-server/test-18/pbench/archive/fs-version-001/ctrl-00/tarball_config-00_1970.01.01T00.00.40.tar.xz
+Restored 3 (0 existing) of (4): /var/tmp/pbench-test-server/test-18/pbench/archive/fs-version-001/ctrl-00/tarball_config-00_1970.01.01T00.00.40.tar.xz
 
-Restored 3 tar balls from backup volume, /var/tmp/pbench-test-server/test-18/pbench-local/archive.backup (ended at: 42.0, elapsed (0.00)
+Exists 1 (3 restored) of (4): /var/tmp/pbench-test-server/test-18/pbench/archive/fs-version-001/ctrl-01/tarball_config-exists_1970.01.01T00.00.39.tar.xz
+Restored 3 tar balls from backup volume, /var/tmp/pbench-test-server/test-18/pbench-local/archive.backup (ended at: 42.0, elapsed: 0.00)
```
webbnh, line 28: This message suggests that it restored __three__ tarballs, but the directory listing below shows only *one* addition...is that correct?

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -17,21 +17,23 @@
...
 
-        * Create TO-RE-UNPACK and BACKED-UP directories if they don't exist
+     c. Create symlink to tar ball in BACKED-UP directory
+
+        * Creating the BACKED-UP directory if it doesn't already exist
 
      c. Create symlink to tar ball in TO-RE-UNPACK directory
 
-  4. After processing 10 tar balls, wait for # of TO-RE-UNPACK to drop to zero
+        * Creating the TO-RE-UNPACK directory if it doesn't already exist
```
webbnh: Line 28 should presumably be step "d", not a second step "c".

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -166,24 +165,36 @@ def main(options):
             cp_cmd = f"cp -a {tb} {tb}.md5 {ctrl_p}/"
             print(f"\n{cp_cmd}", flush=True)
             if not options.dry_run:
-                cp = subprocess.run(cp_cmd, shell=True, stderr=subprocess.STDOUT)
+                cp = subprocess.run(
+                    cp_cmd.replace('"', '\\"').replace("'", "\\'"),
```
webbnh: If you're going to keep the print at line 166, shouldn't the replacement at line 169 be done before the line is printed?  (You could probably append the `.replace()` invocations directly to the f-string in line 165.)

webbnh: Ditto line 181.

webbnh (approved):
My apologies re my earlier review -- I failed to note that this PR was targeting the 0.69 branch...I might have come down differently on a few things if I didn't think that this was forward-going code.

In light of that, perhaps my [comment above](https://github.com/distributed-system-analysis/pbench/pull/2350#discussion_r658216188) can be deemphasized somewhat...although, I still think you should fix it.  🙂 

Other than that and one or two nits below, it looks good.

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -107,7 +107,7 @@ def main(options):
     if not archive_p.is_dir():
         print(
             f"The configured ARCHIVE directory, {archive}, is not a valid directory",
-            file=sys.stderr
+            file=sys.stderr,
```
dbutenhof, line 110: Dang, `black` really loves those trailing list commas. It's a bit obsessive.

dbutenhof (approved):

server/bin/pbench-restore-and-unpack-tarballs.py
webbnh: IIRC from the Black documentation, the trailing comma can be _added by the developer_ to prevent Black from un-wrapping the lines; so, once Black decides that it wants the lines wrapped, I guess it adds it to ensure a consistent and deterministic result (which, I suppose, makes good sense, sadly).

server/bin/unittests
webbnh: Peter's phrase there is "The members of this next set are...", and "he members of t" has been removed for brevity.  😉

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,253 @@
...
+                        int(match.group(4)),
+                        int(match.group(5)),
+                        int(match.group(6)),
+                    )
+                    tb = Path(entry.path[:-4])
+                    if not tb.exists():
+                        raise Exception(
+                            f"Tar ball .md5, '{entry.path}', does not have"
+                            f" an associated tar ball, '{entry.path[:-4]}'"
+                        )
```
webbnh, line 85: I recommend `{tb}` instead of `{entry.path[:-4]}`.

server/bin/pbench-restore-and-unpack-tarballs.py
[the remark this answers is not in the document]
webbnh: Since this appears to be a brand new module, I'm not sure that arguments about this being an existing way or about this being on an old branch hold much water....  😏

server/bin/pbench-restore-and-unpack-tarballs.py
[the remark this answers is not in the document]
webbnh: Since `gen` already yields tuples, there is no need to destructure its return only to re-compose it.  Something like this would do exactly what you code is doing now, in a much more readable way:
```
    tbs_sorted = sorted(list(gen), reverse=True)
```

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,253 @@
...
+                    cp_cmd.replace('"', '\\"').replace("'", "\\'"),
+                    shell=True,
+                    stderr=subprocess.STDOUT,
+                )
+                if cp.returncode != 0:
+                    print(
+                        f"FAILURE: cp command: '{cp_cmd}': {cp.returncode}, {cp.stdout!r}",
+                        file=sys.stderr,
+                    )
+                    return 1
```
webbnh, line 178: The replacement at line 169 should presumably be done before the print at line 166, given that it will show up at line 175 if that appears.

webbnh: Ditto for lines 187, 181, and 194, respectively.

server/bin/pbench-restore-and-unpack-tarballs.py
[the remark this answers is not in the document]
webbnh: It's not hard for the dry-run branch to keep track of which would-create messages it has printed (you just use a `dict` keyed by the path).

webbnh (dismissed):
No deal-breakers, but there are a few things that I'd like you to consider.

And, there are also these two items from previous passes:
 - [request](https://github.com/distributed-system-analysis/pbench/pull/2350#discussion_r658216188) for a better message
 - test result [question](https://github.com/distributed-system-analysis/pbench/pull/2350#discussion_r658219945)

server/bin/pbench-restore-and-unpack-tarballs.py
portante: For the `b0.69` branch we support RHEL 7 with uses SCL (Software Collections Library) [`rh-python36`](https://github.com/distributed-system-analysis/pbench/blob/b0.69/agent/rpm/pbench-agent.spec.j2#L19).

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,253 @@
...
+                continue
+            if not c_entry.is_dir(follow_symlinks=False):
+                continue
+            # We have a controller directory.
+            with os.scandir(c_entry.path) as controller_scan:
+                for entry in controller_scan:
+                    if entry.name.startswith("."):
+                        continue
+                    if entry.is_dir(follow_symlinks=False):
+                        continue
```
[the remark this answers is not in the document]
portante: > We want the entry to be a tarball, right?

Yes.

> And, if it's a directory, then we should skip it.

Yes.

> But, if it's a symlink to a directory, then I think we want to skip it, right?

Yes.

> So, shouldn't the argument be `True`? (Do we want to consider skipping all symlinks?)

It seems this test should be "if not a file skip it."  I'll propose a fix.

server/bin/pbench-restore-and-unpack-tarballs.py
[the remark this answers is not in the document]
portante: Seems like a lot of added mechanism for little gain, since the directory creation is implied by the tar balls that would be restored in the dry-run.

portante:
All comments have been addressed.  Please resolve as you see fit.

server/bin/pbench-restore-and-unpack-tarballs.py
webbnh: _[@riya-17, I think it's on you to mark this conversation resolved, if you accept Peter's answer.]_

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,281 @@
...
+                pass
+            else:
+                if ctrl not in ctrl_created:
+                    print(f"\nCreated controller directory, '{ctrl_p}'", flush=True)
+                ctrl_created[ctrl] = True
+
+            # Copy tar ball to archive tree, along with its .md5 ...
+            cp_cmd = f"cp -a {tb} {tb}.md5 {ctrl_p}/".replace('"', '\\"').replace(
+                "'", "\\'"
+            )
```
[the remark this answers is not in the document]
webbnh: And analogously at line 199.

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,281 @@
...
+                    a_tb,
+                ),
+                flush=True,
+            )
+        except Exception as exc:
+            print(
+                f"Exception while processing ({tb_dt!r}, {ctrl!r}, {tb!r}): {exc}",
+                file=sys.stderr,
+            )
+            raise
```
webbnh, line 256: Are you sure that you want to do a `raise` here?  There is no surrounding `try` scope, so the exception will be unhandled.  This is probably OK on Linux, but it's not good on other platforms, and it's not good style in general.  I would think that the message alone would suffice, so you could replace the `raise` with a `return 1`.  (If you really want the traceback displayed, you can request that explicitly before exiting.)

webbnh (approved):
No deal-breakers, but there are a few things that you might want to consider.

server/bin/gold/test-18.txt
webbnh: Thoughts, @portante?

server/bin/pbench-restore-and-unpack-tarballs.py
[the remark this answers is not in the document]
portante: It doesn't, that is output from `pbench-unpack-tarballs.sh` running as the `pbench-re-unpack-tarballs` job.

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,281 @@
...
+                    ),
+                    flush=True,
+                )
+                continue
+
+            # Create controller directory if it doesn't exist
+            ctrl_p = archive_p / ctrl
+            try:
+                if not options.dry_run:
+                    ctrl_p.mkdir()
```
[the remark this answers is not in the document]
portante: Thanks for sharing that thought.  =)

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -0,0 +1,281 @@
...
+                    md5sum_cmd, shell=True, stderr=subprocess.STDOUT, cwd=str(ctrl_p)
+                )
+                if cp.returncode != 0:
+                    print(
+                        f"FAILURE: md5sum command: '{md5sum_cmd}': {cp.returncode}, {cp.stdout!r}",
+                        file=sys.stderr,
+                    )
+                    return 1
+
+            # Create the symlink recording the tar ball is already backed up
```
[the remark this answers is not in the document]
portante: Okay, but not going to make the change since it makes the line longer than 80 characters.

server/bin/pbench-restore-and-unpack-tarballs.py
portante: This is run at the command line by an admin, not in a cron job, and only on Linux.  Not sure about style, but this code is a stop-gap of sorts for the `0.69` release, so I'd rather have the exception for the time being.

server/bin/pbench-restore-and-unpack-tarballs.py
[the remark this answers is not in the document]
portante: And just to be clear, as I responded on June 23rd, the `PbenchServerConfig` class does not exist on the `b0.69` branch.

portante (commented):
Okay, all comments have been addressed.

server/bin/pbench-restore-and-unpack-tarballs.py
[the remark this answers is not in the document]
webbnh: Anytime....  😁

server/bin/pbench-restore-and-unpack-tarballs.py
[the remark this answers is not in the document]
webbnh: How about "Create the symlink which marks the tar ball as already backed up"?  (That fits within 80 characters....)

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -88,6 +88,25 @@ def gen_list(backup):
...
+def escape_quotes(string):
+    """Escape single and double quotes in a string suitable for use in a shell
+    command.
+
+    Returns the string with the quotes escaped.
+    """
+    return re.sub("(['\"])", r"\\\1", string)
+
+
+def report_tb(*args, **kwargs):
```
webbnh, line 100: Since you don't use `args`, you can omit it from the signature (i.e., the following is legal):
```
def report_tb(**kwargs):
```
and I think that better reflects your intent (and, it forces all arguments to be keyword arguments, which I think is good for "safety").

server/bin/pbench-restore-and-unpack-tarballs.py
```diff
@@ -103,23 +122,39 @@ def main(options):
         print(f"{_NAME_}: {e}", file=sys.stderr)
         return 2
 
-    archive = config.ARCHIVE
-    archive_p = Path(archive).resolve(strict=True)
+    try:
+        archive_p = Path(config.ARCHIVE).resolve(strict=True)
+    except Exception as e:
```
webbnh, line 127: I think you should be catching `FileNotFoundError` here, rather than `Exception` -- `resolve()` _can_ raise other exceptions, and they indicate things other than "file does not exist".  (I'm not trying to say that you need to catch additional cases, here; rather, I think that we shouldn't muddy the waters if we hit a weird case.)

webbnh: Ditto line 146.

server/bin/pbench-restore-and-unpack-tarballs.py
webbnh: However, on the topic of "safety", it really would be better if the signature listed the required arguments -- if the caller omits or misspells one of them, that's going to be a runtime error which we won't find out about until the call is actually made (at which point the resulting `KeyError` will probably seem pretty mysterious).

webbnh (changes requested):
> Okay, all comments have been addressed.

It looks pretty good, although I do have a few comments on your recent additions (see below).  And, I was hoping you could alleviate my [confusion here](https://github.com/distributed-system-analysis/pbench/pull/2350#discussion_r658219945).

(And, of course, we need to get @riya-17 and @dbutenhof to confirm that you've resolved their respective issues.)

server/bin/pbench-restore-and-unpack-tarballs.py
dbutenhof: This method is a pretty thin passthrough; it's really not adding much but the format string (which could be a module level variable) and the `flush=True` vs just calling `print` directly. Seems you could at least make it pay for its supper by calculating `combined` itself rather than making each call point do it.

dbutenhof (dismissed):
I like Webb's suggestions, plus a few more comments... some of which seem to be old but previously un-sent. (Oops.)

riya-17 (dismissed):

server/bin/unittests
portante: The goal of the change was to keep the set of changes small since this is on `b0.69` branch.  As a team we seem to be willing to apply the same level of deep introspection of a change that we apply to the `main` branch (where we want to reduce the amount of unmaintainable code) to this legacy branch.

I'd like us as a team to reduce the amount of change we request on the legacy branches to a minimum.

Having said that ...

server/bin/gold/test-18.txt
portante: That is because `test-18` is the dry-run, and the counts are what _would_ have happened, where as the directory listing below show what did happen.

portante (commented):
Applied more code-review feedback.

ndokos (approved):

riya-17 (approved):

dbutenhof (approved):
Despite the opportunity for some minor cleanups ... yeah, it's 0.69, so let's be done.

npalaska (approved):

webbnh (approved):

Issue #2310, opened 2021-05-13
Convert all the HTTP numeric literals status values to symbolic names


npalaska:
In our code there few places where we currently use numeric values to represent http status code, we should convert these to their corresponding symbolic names.

Issue #2312, opened 2021-05-14
GUnicorn needs to be configured with a sufficient timeout for the worker threads.


webbnh:
GUnicorn needs to be configured with a sufficient timeout for the worker threads.  The default is 30 seconds, which is only sufficient for perhaps 2.3Gb upload.  See [`--timeout`](https://docs.gunicorn.org/en/latest/settings.html#timeout).

PR #2359, opened 2021-06-16
Add gunicorn worker timeout and replace all http numeric status

npalaska:
Fixes issue #2310 and #2312

### DIFF
--- a/lib/pbench/cli/server/shell.py
+++ b/lib/pbench/cli/server/shell.py
@@ -3,13 +3,13 @@
 import subprocess
 import sys
 
-from configparser import NoSectionError, NoOptionError
-from sqlalchemy_utils import database_exists, create_database
+from configparser import NoOptionError, NoSectionError
+from sqlalchemy_utils import create_database, database_exists
 
 from pbench.common.exceptions import BadConfig, ConfigFileNotSpecified
+from pbench.common.logger import get_pbench_logger
 from pbench.server.api import create_app, get_server_config
 from pbench.server.database.database import Database
-from pbench.common.logger import get_pbench_logger
 
 
 def app():
--- a/lib/pbench/cli/server/shell.py
+++ b/lib/pbench/cli/server/shell.py
@@ -33,6 +33,7 @@ def main():
         port = str(server_config.get("pbench-server", "bind_port"))
         db = str(server_config.get("Postgres", "db_uri"))
         workers = str(server_config.get("pbench-server", "workers"))
+        worker_timeout = str(server_config.get("pbench-server", "worker_timeout"))
 
         # Multiple gunicorn workers will attempt to connect to the DB; rather
         # than attempt to synchronize them, detect a missing DB (from the
@@ -52,6 +53,8 @@ def main():
             "gunicorn",
             "--workers",
             workers,
+            "--timeout",
+            worker_timeout,
             "--pid",
             "/run/pbench-server/gunicorn.pid",
             "--bind",
--- a/lib/pbench/server/api/auth.py
+++ b/lib/pbench/server/api/auth.py
@@ -1,10 +1,13 @@
+import datetime
 import jwt
 import os
-import datetime
-from flask import request, abort
+
+from http import HTTPStatus
+from flask import abort, request
 from flask_httpauth import HTTPTokenAuth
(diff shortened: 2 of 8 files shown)


lib/pbench/cli/server/shell.py
```diff
@@ -2,14 +2,14 @@
 
 import subprocess
 import sys
+from configparser import NoOptionError, NoSectionError
```
webbnh, line 6: The blank line should go between the bare `import`'s and the first `from`.

lib/pbench/server/api/auth.py
```diff
@@ -1,3 +1,4 @@
+from http import HTTPStatus
```
webbnh, line 1: This line should be positioned with the other `from` imports, in alphabetical order.

And, while you're at it, could you order the bare `import`'s, and add blank lines between the bare `import`'s, the `from` `import`'s, and the project `import`'s ?

lib/pbench/test/unit/agent/task/test_copy_result_tb.py
```diff
@@ -1,3 +1,4 @@
+from http import HTTPStatus
```
webbnh, line 1: This line should be positioned after the bare `import`'s and before the project `import`'s.

lib/pbench/test/unit/server/test_user_auth.py
```diff
@@ -1,5 +1,6 @@
 import time
 import datetime
+from http import HTTPStatus
 from pbench.server.database.models.users import User
```
webbnh, line 4: We should probably have blank lines around the import that you added, and the bare `import`'s should be reordered.

server/lib/config/pbench-server-default.cfg
```diff
@@ -73,6 +73,8 @@ rest_uri = /api/v%(rest_version)s
 
 # WSGI gunicorn specific configs
 workers = 3
+# Set the gunicorn worker timeout. Setting it to 0 has the effect of infinite timeouts
+worker_timeout = 0
```
webbnh, line 77: Are we sure that we want an _infinite_ time out...as opposed to one which is merely _really big_?

Setting it to 20 minutes, or an hour, or 8 hours, or a day, or something would operate as a safety net while still being "effectively infinite", at least in human terms.

webbnh (dismissed):
Looks good, but the `import` statements ought to be tweaked.

Also, do we really want an "infinite timeout" on the worker threads?

server/lib/config/pbench-server-default.cfg
npalaska: Yes definitely. I wasn't sure on this number so just kept it 0 for now, but we can certainly change it to something like 3600 (1 hour) or some other arbitrary number.

webbnh: I recommend setting it to the longest that it would need to be.  That is, I'm pretty sure the client side would give up after 20 minutes (probably long before that), but we might want it to be longer for debugging on the server side...so, perhaps an hour would be too short...but 9000 seconds seems like a reasonable value and nice round number.

webbnh (approved):

portante (changes requested):
It looks good, though it would be much better to have 3 commits that address the changes separately:

 1. reorder / sort imports
 2. add gunicorn worker timeout option
 3. replace http numeric status values

npalaska:
@portante that's a good point

npalaska:
Actually can "reorder imports" and "replacing http numeric status values" be a part of 1 commit since they both are minor refactoring changes and (import reordering is only happening in files where we are changing the replacing the http status)

webbnh (approved):
GitHub says there's been no change since my last review, so it must still be good.

dbutenhof (approved):

dbutenhof:
Maybe a final squash and make sure there's a good git commit comment for posterity, and this can go in.

npalaska:
There are 2 separate commits in this PR for addressing 2 separate issues, so it should be good.

dbutenhof:
> There are 2 separate commits in this PR for addressing 2 separate issues, so it should be good.

You're not wrong ... although I doubt that really matters in the long run anyway.

PR #2367, opened 2021-07-01
Broken quarantine bug fix

### LINKED ISSUE #2106: Broken quarantine logic?
In `lib/pbench/server/utils.py` the `quarantine` method loops through a potential list of input files to quarantine tarballs by re-linking them into a `dest` destination directory.

It appears that the code inside the loop wants to skip certain files, because, for example, the file doesn't exist. However there's a problematic second clause:

```
        if not os.path.exists(afile) and not os.path.islink(afile):
            continue
```

As written `os.path.islink()` is redundant as a file that doesn't exist can't also be a link. Most likely, the `and` was intended to be an `or`, as the logic inside the loop doesn't create a link to a file, but rather only moved (`shutil.move`) the target... which in our normal file structure makes no sense unless the target _is_ a link.

However, as written, the loop doesn't `continue` on a file that exists but isn't a link, and would therefore move the source file; for reference, the unit tests cause `pbench-server-prep-shim-002` to call `quarantine` with files that are not links (e.g., `test-5.2`):

```
  1970-01-01T00:00:42.000000 ERROR pbench-server-prep-shim-002.pbench-server-prep-shim-002 process_tb -- run-1970-01-01T00:00:42-UTC: Duplicate: /var/tmp/pbench-test-server/test-5.2/pbench-local/pbench-move-results-receive/fs-version-002/controller-d-duplicate/tarball-duplicate_1970.01.01T00.42.00.tar.xz duplicate name
+ 1970-01-01T00:00:42.000000 INFO pbench-server-prep-shim-002.utils quarantine -- /var/tmp/pbench-test-server/test-5.2/pbench-local/pbench-move-results-receive/fs-version-002/controller-d-duplicate/tarball-duplicate_1970.01.01T00.42.00.tar.xz.md5 is not a link
  1970-01-01T00:00:42.000000 ERROR pbench-server-prep-shim-002.pbench-server-prep-shim-002 process_tb -- run-1970-01-01T00:00:42-UTC: Quarantined: /var/tmp/pbench-test-server/test-5.2/pbench-local/pbench-move-results-receive/fs-version-002/controller-e-bad-md5/tarball-bad-md5_1970.01.01T00.42.00.tar.xz failed MD5 check
  1970-01-01T00:00:42.000000 INFO pbench-server-prep-shim-002.pbench-server-prep-shim-002 process_tb -- tarball-bad-md5_1970.01.01T00.42.00.tar.xz: FAILED
  1970-01-01T00:00:42.000000 INFO pbench-server-prep-shim-002.pbench-server-prep-shim-002 process_tb -- md5sum: WARNING: 1 computed checksum did NOT match
+ 1970-01-01T00:00:42.000000 INFO pbench-server-prep-shim-002.utils quarantine -- /var/tmp/pbench-test-server/test-5.2/pbench-local/pbench-move-results-receive/fs-version-002/controller-e-bad-md5/tarball-bad-md5_1970.01.01T00.42.00.tar.xz is not a link
```

Especially in the case of the duplicate name, we probably _do_ want to move the original intake tarball directly to quarantine (do not pass go, do not collect $200) rather than moving a "state" link there as we generally would do later in the pipeline. In which case, perhaps the `and not os.path.islink(afile)` term should simply be removed.

riya-17:
Removing unneeded conditional statement

Fixes #2106
Quarantine logic must be able to deal with all the files till they exist.

### DIFF
--- a/lib/pbench/server/utils.py
+++ b/lib/pbench/server/utils.py
@@ -2,6 +2,7 @@ import os
 import sys
 import shutil
 
+from pathlib import Path
 from pbench.server.database.models.tracker import Dataset, States, DatasetNotFound
 
 
@@ -72,17 +73,18 @@ def quarantine(dest, logger, *files):
         sys.exit(101)
 
     for afile in files:
-        if not os.path.exists(afile) and not os.path.islink(afile):
+        afile = Path(afile)
+        if not afile.exists():
             continue
         try:
             # If the file we're moving is a tarball, update the dataset
             # state. (If it's the associated MD5 file, skip that.)
-            if str(afile).endswith(".tar.xz"):
+            if afile.name.endswith(".tar.xz"):
                 try:
                     Dataset.attach(path=afile, state=States.QUARANTINED)
                 except DatasetNotFound:
                     logger.exception("quarantine dataset {} not found", afile)
-            shutil.move(afile, os.path.join(dest, os.path.basename(afile)))
+            shutil.move(afile, os.path.join(dest, afile.name))
         except Exception:
             logger.exception(
                 'quarantine {} {!r}: "mv {} {}/" failed', dest, files, afile, dest


dbutenhof (approved):
Thanks, Riya; I'd like Peter to take a look before this goes in as a lot of my interpretation of this weird case was speculative.

webbnh:
While we're having @portante look at this code....

The docstring for this function says, "Errors here are fatal but we log an error message to help diagnose problems," but there is no `exit()` call after line 84, as there is in the other exception cases...should there be one?

webbnh:
Also, this branch is going to need a rebase.

lib/pbench/server/utils.py
```diff
@@ -72,7 +72,7 @@ def quarantine(dest, logger, *files):
         sys.exit(101)
 
     for afile in files:
-        if not os.path.exists(afile) and not os.path.islink(afile):
+        if not os.path.exists(afile):
```
npalaska, line 75: Since we are updating this condition, can we also update it to use `Pathlib` module something like `if not 
 Path(afile).exists()`

npalaska (commented):
Needs a rebase.

riya-17:
> The docstring for this function says, "Errors here are fatal but we log an error message to help diagnose problems," but there is no `exit()` call after line 84, as there is in the other exception cases...should there be one?

I am not sure but I think we would not want to exit if we are not able to update the dataset state as we will be saving the failure in the log and it will lose the whole of the reason of quarantine because exiting will lead to TB not being moved to quarantine directory.

webbnh:
> > The docstring for this function says, "Errors here are fatal but we log an error message to help diagnose problems," but there is no `exit()` call after line 84, as there is in the other exception cases...should there be one?
> 
> I am not sure but I think we would not want to exit if we are not able to update the dataset state as we will be saving the failure in the log and it will lose the whole of the reason of quarantine because exiting will lead to TB not being moved to quarantine directory.

I don't necessarily disagree -- mostly, I was pointing out that the code doesn't seem to be consistent with the docstring -- I would be satisfied if we changed either one.

As the code currently stands, we die if we cannot create the destination directory (which makes sense) or if we encounter a problem moving any file into it (which is OK, I guess).

If we move a tar file into the directory _without_ updating the state of its dataset, I feel like we create an inconsistency.  So, it seems like there might be a good justification for dying there, too.  (Hence my hope that @portante would chime in and opine.)

dbutenhof:
> > > The docstring for this function says, "Errors here are fatal but we log an error message to help diagnose problems," but there is no `exit()` call after line 84, as there is in the other exception cases...should there be one?
> > 
> > I am not sure but I think we would not want to exit if we are not able to update the dataset state as we will be saving the failure in the log and it will lose the whole of the reason of quarantine because exiting will lead to TB not being moved to quarantine directory.
> 
> I don't necessarily disagree -- mostly, I was pointing out that the code doesn't seem to be consistent with the docstring -- I would be satisfied if we changed either one.
> 
> As the code currently stands, we die if we cannot create the destination directory (which makes sense) or if we encounter a problem moving any file into it (which is OK, I guess).
> 
> If we move a tar file into the directory _without_ updating the state of its dataset, I feel like we create an inconsistency. So, it seems like there might be a good justification for dying there, too. (Hence my hope that @portante would chime in and opine.)

Yes; I introduced the Dataset "gently" to avoid disrupting too much, and especially because there were already a lot of odd twists to make the gold unit tests work. The original motivation was a way to carry the owner through from `PUT` to index, and I originally had used `find-or-create` operations just about everywhere, and just logging failure as I did here. I ended up solving the worst of the problems by pre-populating the DB in the `unittests` script, allowing me to slide into a more rigorous model; but by that point I was somewhat saturated, I wasn't up to dealing with error recovery mechanisms, and I only partially hardened the code.

I agree that if we can't create or update a `Dataset` at any point, that's a serious failure; gaining severity as the PostgreSQL database becomes more critical to normal server operations. E.g., publish relies on the document map created at index time on a pre-existing `Dataset`; and I'm seriously considering optimizing some of the simpler queries, especially `controllers/list`, to use the DB rather than constructing, launching, and postprocessing an Elasticsearch query with such simple constraints.

Philosophically, I really dislike the idea of a utility method deep in the bowels of the library calling `exit`; it's anti-modular. At the same time, though, I was OK with this PR being a point fix for one small problem rather than a massive re-implementation of the module!

dbutenhof:
> If we move a tar file into the directory _without_ updating the state of its dataset, I feel like we create an inconsistency. So, it seems like there might be a good justification for dying there, too. (Hence my hope that @portante would chime in and opine.)

_Putting this out as a separate reply so it doesn't get lost:_

@portante my concern with my interpretation of the problem (essentially, the inherent redundancy of the check) is in determining exactly what we want to happen when we call `quarantine` with a real tarball rather than a state link; this can happen in `put_shim` for duplicate filenames. (And, as such, the problem is perhaps a bit limited as `PUT` will simply reject bad intake requests outright and we'll never have to deal with quarantine before intake is validated and complete.)

Specifically, do we want to skip the "raw file" or do we want to move it like we do the "state" link later in the pipeline? My interpretation is that leaving a bad intake file (e.g., a duplicate name) where it is _at best_ serves no purpose, and it probably makes more sense to move it regardless of the fact that it's not a link. If this isn't what we want to do, then my commentary led Riya down a dead end.  :-)

lib/pbench/server/utils.py
```diff
@@ -72,17 +73,18 @@ def quarantine(dest, logger, *files):
...
     for afile in files:
-        if not os.path.exists(afile) and not os.path.islink(afile):
+        afile = Path(afile)
+        if not afile.exists():
             continue
         try:
             # If the file we're moving is a tarball, update the dataset
             # state. (If it's the associated MD5 file, skip that.)
-            if str(afile).endswith(".tar.xz"):
+            if afile.name.endswith(".tar.xz"):
```
portante, line 82: We don't want to do this when the file is a symbolic link, right?

portante (changes requested):

lib/pbench/server/utils.py
```diff
@@ -72,17 +73,18 @@ def quarantine(dest, logger, *files):
         sys.exit(101)
 
     for afile in files:
-        if not os.path.exists(afile) and not os.path.islink(afile):
+        afile = Path(afile)
+        if not afile.exists():
```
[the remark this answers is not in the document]
portante: Ah, sorry, that is described in the issue for this PR ...

lib/pbench/server/utils.py
```diff
@@ -72,17 +73,18 @@ def quarantine(dest, logger, *files):
         sys.exit(101)
 
     for afile in files:
-        if not os.path.exists(afile) and not os.path.islink(afile):
```
portante, line 75: How is the original logic wrong?

If you have a two symbolic links, one that references a file that does exist, and one that references a file that does not exist, `os.path.exists(afile)` will return `True` for the symbolic link reference to the existing file, and will return `False` for the symbolic link reference to the non-existent file.

With that, this test is trying to make sure that we "quarantine" a symbolic link that references a non-existent file.

So I am not sure why this code needs to be changed.

portante (changes requested):

lib/pbench/server/utils.py
dbutenhof: OK, I missed the subtlety that `os.path.exists` on a link tests for existence of the link destination; however this `continue` triggers only if `afile` isn't a link at all. So what we're actually saying here is that the path is a non-link file that doesn't exist. I still don't understand how this becomes a meaningful condition?

1. What do we want `quarantine` to do when given a path that's a real file (not a link)? This happens in put-shim, and is what drew my attention to this code.
2. What do we want `quarantine` to do when given a path that's a link to a file that exists? (We probably reasonably guess this to be a state link and rename it.)
3. What do we want `quarantine` to do when given a path that's a link to a file that doesn't exist, and how would this come about? (This might be, as your answer suggests, the intent of this code; but if so it seems it should be `and os.path.islink(afile)`.)

portante: > this `continue` triggers only if `afile` isn't a link at all.

Correct because we cannot invoke `shutil.move()` on a file that does not exist, but we can on a symbolic link that does exist, but links to a file that does not exist.

>  So what we're actually saying here is that the path is a non-link file that doesn't exist. I still don't understand how this becomes a meaningful condition?

We are just protecting the `shutil.move()` from the caller.

> 1. What do we want `quarantine` to do when given a path that's a real file (not a link)? This happens in put-shim, and is what drew my attention to this code.

Move it.

> 2. What do we want `quarantine` to do when given a path that's a link to a file that exists? (We probably reasonably guess this to be a state link and rename it.)

Move it.

> 3. What do we want `quarantine` to do when given a path that's a link to a file that doesn't exist,

Move it (in the best imitation of Sacha Baron Cohen as King Julien XIII possible).

> and how would this come about?

I don't know how, we just need to quarantine the unexpected condition.

> (This might be, as your answer suggests, the intent of this code; but if so it seems it should be `and os.path.islink(afile)`.)

Not sure I understand: We do not want to call `shutil.move()` on a file that does not exist.  So, a file *really* doesn't exist if it does not exist and it is not a symbolic link (thank you Python for being _so_ helpful!).  ;)

webbnh: Given how hard I am having to think about this, there is definitely **_something_** wrong here.  😁

This code skips `afile` if the path does not exist _and_ if the path does not refer to a symbolic link file.  So, if there were a symlink and it referred to a non-existent file, then the existence check would return `False`, and the `islink()` check would return `True`, and it would not be skipped, and it would be quarantined.

`exists()` | `islink()` | Quarantined
--- | --- | ---
False | True | Yes
True | True | Yes
True | False | Yes
False | False | No

Thus, in addition to the first case, we would also quarantine it if the symlink refers to an existent file as well as if it's an existing, real file (and not a symlink).

So, the only time we would skip the quarantine is when the file actually doesn't exist (i.e., it's not there in any form, symlink or otherwise)...which raises the question of how we know to look for it and why we're trying to prevent problems from touching it ([rather than recovering from them](https://docs.python.org/3.5/glossary.html#term-eafp)).

So, yeah, I think I concur that the original logic was appropriate.

The new logic will not quarantine existing but broken symlinks.  But, I have no idea what the ramifications of that are.

dbutenhof: OK; so, in summary, what's not clear here is the nature of the workaround to protect `shutil.move()`.  We really do want to quarantine any file, skipping only a file that __really__ doesn't exist instead of only __mostly__ doesn't exist. That is, `not os.path.exists()` isn't good enough here because we don't want to skip a broken symlink, and a symlink's `exists` is tied to the symlink's target rather than the symlink itself. The `and not os.path.islink` essentially removes the implicit symlink dereference from `exists` to avoid skipping the "mostly doesn't exist" case.

That's twisted.

Given this, I think the real resolution for the issue is to add that information to the `quarantine` function, either as part of the docstring or a comment. (I would prefer the docstring as it's generally more visible.)

webbnh: OK, nevermind...sorry for showing up late!  😆

lib/pbench/server/utils.py
dbutenhof: Why? If it follows our normal naming conventions, this theoretically works fine even if we're seeing a broken "state" link.

The `except DatasetNotFound` protects against the consequences of trying to update a dataset that doesn't actually exist; but the suffix check keeps us from even trying if it's unlikely to represent a real dataset in the first place.

dbutenhof (changes requested):
Given the discussion about the file existence check, looks like these changes really need to be backed out and replaced with a docstring note (or at least a comment) explaining the reasoning behind the odd path checks.

Although I do like the change from `os.path` to `Pathlib`...

lib/pbench/server/utils.py
webbnh: (The fact that my e-mail updates faster than the web page is a little bit...dizzying.  😝)

> (I would prefer the docstring as it's generally more visible.)

I would vote for "both":  explain in the docstring that we move all files, including symlink files; and, put a reminder in a code comment that `exists()` refers to the _target_ when the path refers to a symlink.  (In fact, I would be inclined to reverse the order of the `if` statement -- only check for (non-)existence if the path is _not_ to a symlink.)

portante: Let's close this PR, and the issue, and open a new issue for future work to address this.  If somebody wants to make a PR to do as Webb suggests, great.

Thanks all!

lib/pbench/server/utils.py
portante: Why would we want to mark a tar ball as quarantined, when there is a problem with a symlink?

I think this should be removed from this code path, and in all the places where `quarantine()` is called, have that invocation pass an additional argument to have the tar ball quarantined.

dbutenhof: After the prep-shim "pre-processing" phase, I'd expect any quarantine operation to be done using the state symlink. Ideally I'd like to capture more information about what happened in the `Dataset` object, but with this code structure we don't have any information beyond "the caller decided to quarantine". Yes, we could move it out a layer where we have more context, but that also increases the chances of missing a transition somewhere. Doing state transitions in common code is generally an advantage. This is all something to think about as we move away from "files" and "state links" to represent the data and context. For now, this feels like the best compromise.

lib/pbench/server/utils.py
riya-17: I will close this PR

riya-17:
> Given the discussion about the file existence check, looks like these changes really need to be backed out and replaced with a docstring note (or at least a comment) explaining the reasoning behind the odd path checks.
> 
> Although I do like the change from `os.path` to `Pathlib`...

We can address these changes in another PR with relevant issue

PR #2372, opened 2021-07-16
Fixes for sos-report >= 4.0 versions

### LINKED ISSUE #2363: `sos report` in Fedora 34 produces a SHA256 checksum file, not an MD5 file
That makes `pbench-sysinfo-dump` conclude that the `sos report` generation failed.

### LINKED ISSUE #2371: Use the new `sos report` (notice the space) for OSes with sos 4.0+
The new sos 4.0+ deprecates the `sosreport` command, we should change to using the `sos report` command.

ndokos:
Fixes #2363
Fixes #2371

The command name has changed in versions >= 4.0.
The generated checksum is SHA256 in versions >= 4.1.

### DIFF
--- a/agent/util-scripts/tool-meister/pbench-sysinfo-dump
+++ b/agent/util-scripts/tool-meister/pbench-sysinfo-dump
@@ -75,6 +75,10 @@ function collect_sos {
 	else
 		_quiet=" --quiet"
 	fi
+
+	_sos_report_cmd="sosreport"
+	_cksum_suffix="md5"
+
 	if [[ "${sos_ver}" -lt 3 || ( "${sos_ver}" -eq 3 && ( "${sos_ver_minor}" -lt 5 || ( "${sos_ver_minor}" -eq 5 && -z "${sos_ver_subminor}" ) ) ) ]]; then
 		# Pre-v3.5.1+
 		_modules="general lsbrelease"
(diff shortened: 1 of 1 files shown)


ndokos:
Please hold off reviewing this - I think I goofed. Where is the "Change it to Draft" button?

ndokos:
Tested on F33 (sos v4.0) and F34 (sos v4.1).

agent/util-scripts/tool-meister/pbench-sysinfo-dump
```diff
@@ -101,9 +105,19 @@ function collect_sos {
...
+
+	if [[ "${sos_ver}" -gt 4 || ( "${sos_ver}" -eq 4 && "${sos_ver_minor}" -ge 1 )]]; then
+		# sos uses sha256 sums, not md5, in 4.1+
+		_cksum_suffix="sha256"
+	fi
+
 	_name="pbench-${_pbench_full_hostname}"
 	_cmd="${dir}/sosreport-${_name}.cmd"
-	printf -- "sosreport" > ${_cmd}
+	printf -- "%s" "${_sos_report_cmd}" > ${_cmd}
```
webbnh, line 120: While this obviously works, it's more complex than it needs to be; e.g.:
```
	printf -- "${_sos_report_cmd}" > ${_cmd}
```
If you're worried about there being percent signs in the command string, then use something even simpler, like:
```
	echo ${_sos_report_cmd} > ${_cmd}
```

webbnh (approved):
It looks good except for line 130 (and the nit I called out below).

portante (dismissed):

agent/util-scripts/tool-meister/pbench-sysinfo-dump
ndokos: I am trying to make it a habit to never use a `printf` without a `%s` format spec when  printing anything but a fixed string. Feel free to whack me if I deviate from that :-)

ndokos:
Fixed the comment on l.130

agent/util-scripts/tool-meister/pbench-sysinfo-dump
webbnh: Fair enough.  Although, under the heading of "use the right tool for the job", I would suggest that, if you don't use `printf`, you won't run into that problem.  😉

webbnh (approved):
Good to go.

agent/util-scripts/tool-meister/pbench-sysinfo-dump
portante: We can't use `echo` here because it adds a newline, and we are forming a command line in the file, `${_cmd}`.

So one could use `echo -n "${_cmd}"`.

I believe the quotes around the string are required to prevent the contents of the string from being interpreted by the shell.

portante (approved):

agent/util-scripts/tool-meister/pbench-sysinfo-dump
webbnh: As you say, the `-n` switch resolves the newline problem.

I don't think the quotes are required:  the variable expansion is done during the shell's interpretation -- the contents don't get interpreted a second time.

PR #2374, opened 2021-07-17
Add Jenkins support to b0.69 branch

webbnh:
This pull request changes the CI for the `b0.69` branch from Travis CI to Jenkins by back-porting the Jenkins support from 0.71.

I've got the tests running successfully outside the Github environment; now I'm creating a PR to see if Mr. Jenkins will run the new test support automagically.

Once that's working, there is still some polishing to be done, so this PR is in draft mode for the moment.

(Note to self:  compare this PR to [2695e8e2a436cb0eaaa2099f231144dffad0c227](https://github.com/distributed-system-analysis/pbench/commit/2695e8e2a436cb0eaaa2099f231144dffad0c227) to make sure that I didn't miss anything.)


jenkins/python-setup.sh
```diff
@@ -28,3 +28,9 @@ $(head -n 1 ${_prefix}/lib*/python3.*/site-packages/pbench.egg-link 2> /dev/null
 export PYTHONPATH
 
 python3 setup.py develop --prefix="${_prefix}"
+
+set -x
+ls -l $(realpath /dev/log)
```
dbutenhof, line 33: I might've added a `file $(realpath /dev/log)`

agent/util-scripts/pbench-register-tool
```diff
@@ -232,6 +232,7 @@ debug_log "tool_opts: \"${tool_opts[@]}\""
 
 # Determine what our local hostname(s) and IP(s) are.
 # FIXME: this does not check for ipv6
+if ! command -v ip >/dev/null 2>&1 ; then echo "PATH=$PATH"; fi
```
webbnh, line 235: This is, strictly speaking, debugging code, but it was a little hard to set up (believe it or not), and it was so very illuminating when it triggered, that I'm inclined to leave it in place:  the code below it depends on having `/usr/sbin` in the User's `PATH` in order to find the `ip` command, and, if it is not there (as it is not, by default, in Mr. Jenkins's environment), it comes as a nasty surprise, which the output of this command will help diminish.  (And, otherwise, it has no effect.)

portante: Why not move this to the beginning of `jenkins/run-unittests` so that we don't have to modify exiting code already being shipped?

jenkins/Pipeline.gy
```diff
@@ -0,0 +1,11 @@
+pipeline {
+    agent { label 'pbench' }
+    stages {
+        stage('Linting & Unit Tests') {
+            steps {
+                echo 'Linting and legacy unit tests'
+                sh 'jenkins/run jenkins/tox -r --current-env -e jenkins-unittests'
```
portante, line 7: Instead of modifying `jenkins-unittests` to add explicit `black` and `flake8` invocations, why not just add the `lint` before `jenkins-unittests` tox environment target?

jenkins/tox
```diff
@@ -0,0 +1,23 @@
...
+
+export TOXWORKDIR=/var/tmp/tox
+mkdir -p ${TOXWORKDIR}
+if [[ ${?} -ne 0 ]]; then
+    exit 1
+fi
+export HOME=${TOXWORKDIR}
+export XDG_CONFIG_HOME=${TOXWORKDIR}/.config
+
+export PATH=${PATH}:/usr/sbin  # For /usr/sbin/ip, used by pbench-register-tool
```
portante, line 21: Seems like we really need to do this in `agent/profile` as well, by just adding a `_pbench_pathins /usr/sbin` before line 36.

And this would be better served in the `agent/util-scripts/unittests` since it is not necessarily a "jenkins" thing.

jenkins/development.Dockerfile
```diff
@@ -0,0 +1,113 @@
+# Fedora 33 pbench development image
```
portante, line 1: Not sure where, but we really need to document somewhere that using the `black` RPM provided by Fedora 33 and later does not work for our code base and we have to use a `pip` installed version explicitly.

server/bin/test-bin/cat
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+if [[ ! -z "${1}" && "$(basename ${1})" == "tarball-bad-prefix-file_1970.01.01T00.00.00.prefix" ]]; then
+    # The server test-20 expects a file of this name to fail when `cat` tries
+    # to read it.
+    exit 1
+fi
+/usr/bin/cat ${*}
```
portante, line 8: I removed this file from my test environment and the unit tests pass with flying colors.  Can you verify why this is needed?

portante (changes requested):

server/bin/test-bin/cat
portante: See `Draft` PR #2398 CI output.

jenkins/python-setup.sh
portante: Since this is a back-port, we might want to track such an addition on the `main` branch.

lint-requirements.txt
```diff
@@ -0,0 +1,2 @@
+black==19.10b0
+flake8==3.7.7
```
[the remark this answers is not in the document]
webbnh: Because that's what I think the `main` branch is using?

jenkins/Pipeline.gy
webbnh: Because, more or less, that's the way that `main` did it?  (I can do it whichever way we want...how do we want to do it?)

jenkins/python-setup.sh
webbnh: This was debugging code which Dave commented on during the brief interval after it was added and before it was removed.  There is nothing more to see here, now.

agent/util-scripts/pbench-register-tool
webbnh: The crux of the matter is, if the user's environment is not set up correctly, then things will go off the rails at the very next statement, regardless of how this command is invoked, and so this is the appropriate point at which to detect, diagnose, and/or report the failure.  However, this is rather cryptic way to do it, and, it would be best to prevent this problem rather than to diagnose it on the fly, so, I'll revert this change so that this PR doesn't impact any of the product code.

See also #2383.

jenkins/Pipeline.gy
webbnh: I've changed `Pipeline.gy` to invoke the `lint` "environment" instead of running Black and Flake8 in the `jenkins-unittests` script.

jenkins/tox
webbnh: I'm going to defer this; tracked in #2383.

jenkins/development.Dockerfile
webbnh: I enlarged the commentary in the Dockerfile.  Ideally, we would periodically review the test console output to ensure that it reports nothing surprising.

server/bin/test-bin/cat
webbnh: I tried removing it, and it now seems to work without it.

jenkins/Pipeline.gy
webbnh: Ayaaaand, that didn't work.  (`tox` declines to install the linters before trying to run them, and after a couple of straightforward tweaks, I'm unable to convince it to change its mind.)  So, I've restored the flow to run the linters from the script, which executes after the installation.

portante (changes requested):
Looks like we just need a rebase, and have it broken into a commit for the Jenkins work vs what it took to get the tests to pass.

webbnh:
OK, I've rebased and done some squashing.  @portante, let me know what else you want squashed.

portante (approved):
Looks good!

jenkins/python-setup.sh
```diff
@@ -0,0 +1,29 @@
...
+    exit 1
+fi
+
+pip3 install --no-cache-dir --progress-bar off --no-color --prefix="${_prefix}" \
+    -r ${progdir}/lint-requirements.txt \
+    -r ${progdir}/server/requirements.txt \
+    -r ${progdir}/agent/test-requirements.txt
+    # -r ${progdir}/agent/requirements.txt \
+    # -r ${progdir}/server/test-requirements.txt \
+    # -r ${progdir}/docs/requirements.txt \
```
dbutenhof, line 14: Do we want to retain the commented-out lines? (Default answer is "no". 😀 )

dbutenhof (approved):
One minor cleanup? But possibly not worth another commit right now?

jenkins/python-setup.sh
portante: Let's merge, we can fix in a follow-up commit.

webbnh: > Do we want to retain the commented-out lines? (Default answer is "no". 😀 )

No.  Well, yes.  But, no.  Um...maybe?  🤷

As you say, the default is "no", and I'm generally against stuff like this; however, it was just annoying enough flipping back and forth between `b0.69` and `main` that I felt like it was reasonable to have a record here of what things look(ed) like in `main`, so I left it in.  (I figured I could/would delete it if someone actually objected, but...for better or for worse, you objection was not "timely"...so it will be retained for now.)

PR #2375, opened 2021-07-17
Expect `trafficgen` to be installed already

portante:
Benchmark wrappers should not merge the concern of how software is to be installed and deployed with how to run that software.  Those are two separate concerns that create a tightly coupled set of behaviors where the benchmark wrapper has to "know" how the software should be built and installed.

For `trafficgen`, we now expect the code to be installed in the expected location, `/opt/bench-trafficgen/trafficgen`, with the installation process providing the version number of the software in the `VERSION` file.

### DIFF
--- a/agent/bench-scripts/pbench-trafficgen
+++ b/agent/bench-scripts/pbench-trafficgen
@@ -86,7 +86,6 @@ tool_group=default
 one_shot="n"
 skip_trex_install="n"
 skip_trex_server="n"
-skip_git_pull="n"
 flow_mods_def="src-ip,dst-ip,src-mac,dst-mac"
 flow_mods="${flow_mods_def}"
 all_possible_flow_mods="src-ip,dst-ip,src-mac,dst-mac,src-port,dst-port,encap-src-ip,encap-dst-ip,encap-src-mac,encap-dst-mac,protocol,none"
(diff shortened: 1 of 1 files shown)


agent/bench-scripts/pbench-trafficgen
```diff
@@ -100,7 +99,7 @@ sysinfo="none"
 tool_period="binary-search" # can also be "repeat-final-validation"
 trafficgen_version="unknown"
 if [[ -z "${trafficgen_dir}" ]]; then
-	trafficgen_dir="/opt/trafficgen"
+	trafficgen_dir="/opt/bench-trafficgen/trafficgen"
```
webbnh, line 102: Should that be `pbench-trafficgen` (with a leading `p`)?

agent/bench-scripts/pbench-trafficgen
```diff
@@ -856,39 +848,21 @@ if [ "${postprocess_only}" == "n" ]; then
...
 	else
-		if pushd /opt >/dev/null && git clone https://github.com/atheurer/trafficgen.git; then
-			trafficgen_version=$(git log --max-count=1 --pretty=%h)
-			echo "trafficgen cloned"
-			popd >/dev/null
-		else
-			error_log "could not clone latest trafficgen"
-			exit 1
-		fi
+		trafficgen_version="unknown"
```
webbnh, line 858: `trafficgen_version` is initialized to this at line 100...do we need this assignment here?

webbnh (commented):
Not sure why I'm reviewing a draft PR (other than that you requested it... 😉), but it looks good to me (but, I would add the missing `p` and remove the redundant assignment -- see below).

agent/bench-scripts/pbench-trafficgen
portante: That is apparently when the "Crucible" environment calls it.

portante:
This is a draft PR because this is only a proposal for the `bench-trafficgen` repository to consider.

portante:
Doesn't look like this support is coming anytime soon. :(

PR #2376, opened 2021-07-19
Drop Fedora 32, add Fedora 34 to builds

portante:


agent/containers/images/Makefile
```diff
@@ -198,8 +198,8 @@ centos-7-base.Dockerfile: Dockerfile.base.j2 epel-7-pbench.repo
 fedora-33-base.Dockerfile: Dockerfile.base.j2 fedora-33-pbench.repo
 	jinja2 Dockerfile.base.j2 -D pbench_repo_file=fedora-33-pbench.repo -D pkgmgr=dnf -D distro_image=fedora:33 -D distro_image_name="Fedora 33" -o $@
 
-fedora-32-base.Dockerfile: Dockerfile.base.j2 fedora-32-pbench.repo
-	jinja2 Dockerfile.base.j2 -D pbench_repo_file=fedora-32-pbench.repo -D pkgmgr=dnf -D distro_image=fedora:32 -D distro_image_name="Fedora 32" -o $@
+fedora-34-base.Dockerfile: Dockerfile.base.j2 fedora-34-pbench.repo
+	jinja2 Dockerfile.base.j2 -D pbench_repo_file=fedora-34-pbench.repo -D pkgmgr=dnf -D distro_image=fedora:34 -D distro_image_name="Fedora 34" -o $@
```
webbnh, line 202: Just a nit, but this changes the order of the Fedora targets from decreasing to increasing (while the Centos targets are still in decreasing order).

agent/containers/images/README.md
```diff
@@ -3,7 +3,7 @@
 Container image building requires both the `buildah` and `jinja2` CLI
 commands to be present.
 
-On Fedora 32 and later systems, use the following command to install the
+On Fedora 33 and later systems, use the following command to install the
```
webbnh, line 6: What about on Centos (and/or RHEL)?

webbnh (approved):
Just a couple of nits.

agent/containers/images/Makefile
portante: Was trying to reduce the amount of differences.

agent/containers/images/README.md
portante: What does the change from 32 to 33 mean for CentOS or RHEL?

dbutenhof (approved):

agent/containers/images/Makefile
webbnh: Understood, but, overall there are a small number of changes, and future readers will most likely be looking at the whole file rather than just the diffs.  And, in this case, the diffs appear (more or less) to be "this _line_ removed and this _line_ inserted", which would still be the case if you positioned them properly (and, it's not hard to see what the differences are within these two lines).

I almost wonder whether there shouldn't be a single "fedora" target using a pattern target form with the `33` or `34` just plugged in where it goes.  E.g., I think something like this would "just work":
```
fedora-%-base.Dockerfile: Dockerfile.base.j2 fedora-%-pbench.repo
	jinja2 Dockerfile.base.j2 -D pbench_repo_file=fedora-$*-pbench.repo -D pkgmgr=dnf -D distro_image=fedora:$* -D distro_image_name="Fedora $*" -o $@
```
although the value for `distro_image_name` might need some additional handling.

agent/containers/images/README.md
webbnh: The README discusses a number of points regarding Fedora.  At one spot, it also mentions CentOS.  It seems odd that it mentions CentOS there without calling it out in the other places that it mentions details for Fedora.  (I didn't notice _any_ references to RHEL, which did seem slightly odd.)

It seems worthwhile, while we were updating this document, for us to give it a full once-over, to make sure that it still says what we want it to say....

portante: Tracked in #2379.

agent/containers/images/Makefile
portante: Tracked in #2380.

PR #2377, opened 2021-07-19
Quick-n-Dirty bit-rot detection for tar balls

portante:
This is a quick capture of the current bit-rot detection that has been deployed as a stop-gap whilest backup-verification is off.

The first commit contains the originally deployed stop-gap, include a bug where empty or missing `.md5` files for `.tar.xz` files are not detected.

The second commit contains the fixes to address those problems and all the code-review feedback.

### DIFF
--- a/server/user/bitrot-detect
+++ b/server/user/bitrot-detect
@@ -18,18 +18,37 @@ mkdir -p ${WORKDIR}
 # Clean up previous run.
 rm -f ${WORKDIR}/tbs.md5-*
 
-find ${TARGETDIR}/ -maxdepth 1 -type d -printf '%P\n' | sort | while read ctrl; do
-    cat ${TARGETDIR}/${ctrl}/*.md5 2> /dev/null | while read hash file; do
-        echo "${hash} ${TARGETDIR}/${ctrl}/${file}"
+# We "find" all the controller directories and process them in alphabetical
+# order.
+find ${TARGETDIR} -maxdepth 1 -mindepth 1 -type d -printf '%P\n' | sort | while read ctrl; do
+    find ${TARGETDIR}/${ctrl} -maxdepth 1 -mindepth 1 -type f -name '*.tar.xz' | while read tb; do
+        # Detect empty tar balls
+        if [[ ! -s ${tb} ]]; then
+            printf -- "Empty tar ball found: %s\n" "${tb}" >&2
+            continue
+        fi
+        _md5=${tb}.md5
+        if [[ ! -f ${_md5} ]]; then
+            printf -- "MD5 file for tar ball missing: %s\n" "${tb}" >&2
+        elif [[ ! -s ${_md5} ]]; then
+            printf -- "Empty tar ball MD5 file found: %s\n" "${_md5}" >&2
+        else
+            # For each controller, we read all the .md5 file contents re-
+            # emitting those contents with a file name that contains the full
+            # path.
+            awk "{ print \$1, \"${TARGETDIR}/${ctrl}/\" \$2 }" ${_md5}
+        fi
     done
+    # All the output is captured into one file so that the requested
+    # concurrency can be provided below.
 done > ${WORKDIR}/tbs.md5 2> ${WORKDIR}/tbs.err
 
 cat ${WORKDIR}/tbs.err; echo ""
 
-# We use the `--number=r/N` form so that all the workers below will
-# chew through the list together in alphabetical order, and reduces
-# the possibility of one workre getting all the biggest tar balls to
-# process.
+# We use the `--number=r/N` form so that all the workers below will chew
+# through the list together in alphabetical order. This makes it convenient
+# for an observer to understand how far along the bit-rot detection process
+# has progressed.
 split --number=r/${NCPUS} ${WORKDIR}/tbs.md5 ${WORKDIR}/tbs.md5-
 
 for i in ${WORKDIR}/tbs.md5-*; do
--- a/server/user/bitrot-detect
+++ b/server/user/bitrot-detect
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+NCPUS=${1}
+if [[ -z "${NCPUS}" ]]; then
+    echo "Missing number of CPUs to use" >&2
+    exit 1
+fi
+
+TARGETDIR=${2}
+if [[ -z "${TARGETDIR}" ]]; then
+    echo "Missing target directory argument" >&2
+    exit 1
+fi
+
+WORKDIR=/var/tmp/bitrot-detect
+mkdir -p ${WORKDIR}
+
+# Clean up previous run.
+rm -f ${WORKDIR}/tbs.md5-*
+
+find ${TARGETDIR}/ -maxdepth 1 -type d -printf '%P\n' | sort | while read ctrl; do
+    cat ${TARGETDIR}/${ctrl}/*.md5 2> /dev/null | while read hash file; do
+        echo "${hash} ${TARGETDIR}/${ctrl}/${file}"
+    done
+done > ${WORKDIR}/tbs.md5 2> ${WORKDIR}/tbs.err
+
+cat ${WORKDIR}/tbs.err; echo ""
+
+# We use the `--number=r/N` form so that all the workers below will
+# chew through the list together in alphabetical order, and reduces
+# the possibility of one workre getting all the biggest tar balls to
+# process.
+split --number=r/${NCPUS} ${WORKDIR}/tbs.md5 ${WORKDIR}/tbs.md5-
+
+for i in ${WORKDIR}/tbs.md5-*; do
+    md5sum --check ${i} > ${i}.out 2> ${i}.err &
+done
+wait
+
+cat ${WORKDIR}/tbs.md5-*.err; echo ""
+
+grep -v "OK" ${WORKDIR}/tbs.md5-*.out
+
+exit 0
--- a/server/user/bitrot-detect.service
+++ b/server/user/bitrot-detect.service
@@ -1,0 +1,10 @@
+[Unit]
+Description=the bit-rot detection script
+Wants=bitrot-detect.timer
+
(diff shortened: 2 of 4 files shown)


server/user/bitrot-detect
```diff
@@ -0,0 +1,44 @@
...
+fi
+
+WORKDIR=/var/tmp/bitrot-detect
+mkdir -p ${WORKDIR}
+
+# Clean up previous run.
+rm -f ${WORKDIR}/tbs.md5-*
+
+find ${TARGETDIR}/ -maxdepth 1 -type d -printf '%P\n' | sort | while read ctrl; do
+    cat ${TARGETDIR}/${ctrl}/*.md5 2> /dev/null | while read hash file; do
```
dbutenhof, line 22: If you're collecting loop error output to `${WORKDIR}/tbs.err`, why throw away any `cat` errors here? It looks like you're really only collecting errors on the `read` and `echo` commands, which seems unlikely to be all that interesting?

server/user/bitrot-detect
```diff
@@ -0,0 +1,44 @@
...
+        echo "${hash} ${TARGETDIR}/${ctrl}/${file}"
+    done
+done > ${WORKDIR}/tbs.md5 2> ${WORKDIR}/tbs.err
+
+cat ${WORKDIR}/tbs.err; echo ""
+
+# We use the `--number=r/N` form so that all the workers below will
+# chew through the list together in alphabetical order, and reduces
+# the possibility of one workre getting all the biggest tar balls to
```
dbutenhof, line 31: `s/workre/worker`

And since you're sorting by filename, not by size, doesn't seem the order has any meaningful correlation to size: whether the splits or sequential or round-robin, the _size_ distribution is really random.

(Although I suppose that's not entirely true, since you'll be throwing the tarballs for "unnamed but well known users who typically generate enormous tarballs" across your workers... but if that's the point, maybe explaining that directly would make more sense of this.)

server/user/bitrot-detect
```diff
@@ -0,0 +1,44 @@
...
+mkdir -p ${WORKDIR}
+
+# Clean up previous run.
+rm -f ${WORKDIR}/tbs.md5-*
+
+find ${TARGETDIR}/ -maxdepth 1 -type d -printf '%P\n' | sort | while read ctrl; do
+    cat ${TARGETDIR}/${ctrl}/*.md5 2> /dev/null | while read hash file; do
+        echo "${hash} ${TARGETDIR}/${ctrl}/${file}"
+    done
+done > ${WORKDIR}/tbs.md5 2> ${WORKDIR}/tbs.err
```
webbnh, line 25: Alternatively, perhaps something like this:
```
# Find, within the list of immediate subdirectories of the target directory, all files
# ending in ".md5"; transcribe their contents, replacing the filename with the full
# path (of the MD5 file minus the ".md5" extension, to the output; sort the output
# by path.
find $(find ${TARGETDIR} -maxdepth 1 -type d) \
    -maxdepth 1 -type f -name "*.md5" \
    -exec awk '{ sub(".md5", "", FILENAME); print $1, FILENAME }' {} \+ \
    2> ${WORKDIR}/tbs.err | sort -k2,2 >${WORKDIR}/tbs.md5
```

server/user/bitrot-detect
```diff
@@ -0,0 +1,44 @@
...
+
+cat ${WORKDIR}/tbs.err; echo ""
```
webbnh, line 27: You shouldn't need the empty argument to `echo` (it will produce the newline without it).

Ditto line 40.

webbnh: (And, if you're interested, I think there's a way to incorporate the `cat` into the previous command using `tee`, but it's a little bit woo-woo.... 😉)

server/user/bitrot-detect
webbnh: We _could_ of course sort the tarballs by **size** and distribute them round-robin....

webbnh (approved):
Other than the `systemctl` descriptions (and the typo that Dave beat me to 😉), it looks OK to me.  I offered a few suggestions.

server/user/bitrot-detect
portante: Bad comment.  Reworded the comment to match the code and the intention of the code.

server/user/bitrot-detect.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=the bit-rot detection script
```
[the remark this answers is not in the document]
portante: Actually, those descriptions are emitted by systemd when you use `systemctl --user status bitrot-detect[.timer]`.  The were crafted so that output made sense.

server/user/bitrot-detect
```diff
@@ -0,0 +1,44 @@
...
+WORKDIR=/var/tmp/bitrot-detect
+mkdir -p ${WORKDIR}
+
+# Clean up previous run.
+rm -f ${WORKDIR}/tbs.md5-*
+
+find ${TARGETDIR}/ -maxdepth 1 -type d -printf '%P\n' | sort | while read ctrl; do
+    cat ${TARGETDIR}/${ctrl}/*.md5 2> /dev/null | while read hash file; do
+        echo "${hash} ${TARGETDIR}/${ctrl}/${file}"
+    done
```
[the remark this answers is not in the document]
portante: Thanks for checking, but in this case I find the more verbose logic here easier to follow than `sed` or `awk`.

server/user/bitrot-detect
portante: See the latest code which adds some additional checks.

server/user/bitrot-detect
portante: Agreed, though I left the empty string there to indicate it was intentional.

server/user/bitrot-detect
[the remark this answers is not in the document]
portante: Okay, after trying this out with the new empty file detection, I used a modified version of that `awk` command.  Thanks!

dbutenhof (dismissed):
Good addition. Makes me wonder suddenly why you did this in `bash` rather than `Python`; but `find` is pretty well optimized for the basic guts of this algorithm.

ndokos (dismissed):

server/user/bitrot-detect
webbnh: I might actually be convinced to use `printf` in this particular case:
```
cat ${WORKDIR}/tbs.err | xargs -0 printf -- "%s\n"
```
😉

webbnh:
> Good addition. Makes me wonder suddenly why you did this in `bash` rather than `Python`; but `find` is pretty well optimized for the basic guts of this algorithm.

Python offers [os.walk()](https://docs.python.org/3/library/os.html#os.walk) which should provide the `find(1)` capabilities, and then we're off to the races!  🙂

server/user/bitrot-detect
```diff
@@ -0,0 +1,66 @@
...
+WORKDIR=/var/tmp/bitrot-detect
+mkdir -p ${WORKDIR}
+
+# Clean up previous run.
+rm -f ${WORKDIR}/tbs.md5-*
+
+# We "find" all the controller directories and process them in alphabetical
+# order.
+find ${TARGETDIR}/ -maxdepth 1 -mindepth 1 -type d -printf '%P\n' | sort | while read ctrl; do
+    find ${TARGETDIR}/${ctrl} -type f -name '*.tar.xz' | while read tb; do
```
webbnh, line 24: At line 23, you don't need the trailing slash on `${TARGETDIR}/`, and providing it will ensure that `ctrl` ends in a slash, which means that the argument to `find` at line 24` will have a double slash before the final component (and, I believe that that will result in `tb` containing a doubled slash as well...I don't recall).  Not that any of this matters functionally in this case, but there are times when it's annoying, and it's easily avoided.

webbnh: Note that if `${TARGETDIR}/${ctrl}` contains any subdirectories, the `find` at line 24 will descend into them looking for tarballs...is this intentional?  If not, it would be worth adding the `-maxdepth 1` (and presumably `-mindepth 1`, as well) to the invocation.

server/user/bitrot-detect
```diff
@@ -0,0 +1,66 @@
...
+        if [[ ! -s ${tb} ]]; then
+            printf -- "Empty tar ball found: %s\n" "${tb}" >&2
+        fi
+        _md5=${tb}.md5
+        # Detect missing .md5 files
+        if [[ ! -f ${_md5} ]]; then
+            printf -- "MD5 file for tar ball missing: %s\n" "${tb}" >&2
+        else
+            # Detect empty .md5 files
+            if [[ ! -s ${_md5} ]]; then
```
webbnh, line 35: This would be a good spot for an `elif`.

webbnh (dismissed):
Looks fine.  Just two nits for polishing and a suggestion.

server/user/bitrot-detect
ndokos: I don't get trailing slashes on subdirs:

```
$ find /srv/pbench/archive/fs-version-001/  -maxdepth 1 -mindepth 1 -type d -printf '%P\n'
a
b
c
```

ndokos: There are only two levels under a controller directory,  and the second levels should not have anything but symlinks, but limiting the `find` is a good thing, particularly after what we saw last month.

ndokos (approved):

server/user/bitrot-detect
webbnh: > I don't get trailing slashes on subdirs

Ah!  That would be the `-printf` -- I'm unaccustomed to that.

> There are only two levels under a controller directory

There are only _supposed_ to be....  I'm a fan of adding resilience and robustness when it doesn't cost too much.  🙂

portante: Applied comments.

webbnh (approved):
GTG

ndokos (approved):

PR #2382, opened 2021-07-21
Documents index mappings api

npalaska:
API for returning documents mappings.
- The mapping index name need to be specified in the URI `/api/v1/mappings/<index_name>` for example to get `run-data` document mappings one can perform the GET request on `/api/v1/mappings/run`
- Internally we return the mappings by querying the `Template` table instead of making a direct ES query.

### DIFF
--- a/lib/pbench/server/api/resources/query_apis/index_mappings.py
+++ b/lib/pbench/server/api/resources/query_apis/index_mappings.py
@@ -1,0 +1,88 @@
+from http import HTTPStatus
+from logging import Logger
+from typing import Any, AnyStr, Dict
+
+from flask import jsonify
+from flask_restful import Resource, abort
+
+from pbench.server.database.models.template import Template, TemplateNotFound
+
+
+class IndexMappings(Resource):
+    """
+    Get properties of the run-data document type on which the dashboard can search.
+    This API currently only returns mapping properties of run-data documents.
+
+    TODO: In future we can tweak this API to take the document type name from the user
+    and query that type name against the template table to return the appropriate mappings
+
+    :return:
+    {
+        "@metadata":
+            [
+                "controller_dir",
+                "file-date",
+                "file-name",
+                "file-size",
+                "md5",
+                ...
+            ],
+       "host_tools_info":
+            [
+                "hostname",
+                "hostname-f",
+                ...
+            ],
+        "authorization":
+            [
+                "access",
+                "owner"
+            ]
+        "run":
+            [
+                "id",
+                "controller",
+                "user",
+                "name",
+                ...
+            ],
+        "sosreports":
+            [
+                "hostname-f",
+                "sosreport-error",
+                "hostname-s",
+                ...
+            ]
+    }
+    """
+
+    def __init__(self, logger: Logger):
+        self.logger = logger
+
+    def get(self, index_name: AnyStr) -> Dict[AnyStr, Any]:
+        """
+        Return mapping properties of the document specified by the index name in the URI.
+        For example, user can get the run document properties by making a GET request on
+        index/mappings/run. Similarly other index documents can be fetched by making a GET
+        request on appropriate index names.
+        We fetch the mapping by querying the template database. If the template is not found
+        in the database NOT_FOUND error will be raised.
+        """
+        try:
+            template = Template.find(index_name)
+            mappings = template.mappings
+
+            result = {}
+            for property in mappings["properties"]:
+                if "properties" in mappings["properties"][property]:
+                    result[property] = list(
+                        mappings["properties"][property]["properties"].keys()
+                    )
+
+            # construct response object
+            return jsonify(result)
+        except TemplateNotFound:
+            self.logger.exception(
+                "Document template {} not found in the database.", index_name
+            )
+            abort(HTTPStatus.NOT_FOUND, message="Mapping not found")
--- a/lib/pbench/test/unit/server/conftest.py
+++ b/lib/pbench/test/unit/server/conftest.py
@@ -222,17 +222,179 @@ def find_template(monkeypatch, fake_mtime):
     """
 
     def fake_find(name: str) -> Template:
-        return Template(
-            name="run",
-            idxname="run-data",
-            template_name="unit-test.v6.run-data",
-            file="run.json",
-            template_pattern="unit-test.v6.run-data.*",
-            index_template="unit-test.v6.run-data.{year}-{month}",
-            settings={"none": False},
-            mappings={"properties": None},
-            version=5,
-        )
+        if name == "run":
+            return Template(
+                name="run",
+                idxname="run-data",
+                template_name="unit-test.v6.run-data",
+                file="run.json",
+                template_pattern="unit-test.v6.run-data.*",
+                index_template="unit-test.v6.run-data.{year}-{month}",
+                settings={"none": False},
+                mappings={
+                    "_meta": {"version": "6"},
+                    "date_detection": "false",
+                    "properties": {
+                        "@generated-by": {"type": "keyword"},
+                        "@metadata": {
+                            "properties": {
+                                "controller_dir": {"type": "keyword"},
+                                "file-date": {"type": "date"},
+                                "file-name": {"type": "keyword"},
+                                "file-size": {"type": "long"},
+                                "md5": {"type": "keyword"},
+                                "pbench-agent-version": {"type": "keyword"},
+                                "raw_size": {"type": "long"},
+                                "result-prefix": {"type": "text"},
+                                "satellite": {"type": "keyword"},
+                                "tar-ball-creation-timestamp": {"type": "date"},
+                                "toc-prefix": {"type": "text"},
+                            }
+                        },
+                        "@timestamp": {"type": "date"},
+                        "authorization": {
+                            "properties": {
+                                "access": {
+                                    "type": "text",
+                                    "fields": {
+                                        "keyword": {
+                                            "type": "keyword",
+                                            "ignore_above": 256,
+                                        }
+                                    },
+                                },
+                                "owner": {
+                                    "type": "text",
+                                    "fields": {
+                                        "keyword": {
+                                            "type": "keyword",
+                                            "ignore_above": 256,
+                                        }
+                                    },
+                                },
+                            }
+                        },
+                        "host_tools_info": {
+                            "type": "nested",
+                            "properties": {
+                                "hostname": {"type": "keyword"},
+                                "hostname-f": {"type": "keyword"},
+                                "hostname-s": {"type": "keyword"},
+                                "label": {"type": "keyword"},
+                                "tools": {
+                                    "properties": {
+                                        "disk": {"type": "text"},
+                                        "haproxy-ocp": {"type": "text"},
+                                        "iostat": {"type": "text"},
+                                        "mpstat": {"type": "text"},
+                                        "oc": {"type": "text"},
+                                        "perf": {"type": "text"},
+                                        "pidstat": {"type": "text"},
+                                        "pprof": {"type": "text"},
+                                        "proc-interrupts": {"type": "text"},
+                                        "proc-vmstat": {"type": "text"},
+                                        "prometheus-metrics": {"type": "text"},
+                                        "sar": {"type": "text"},
+                                        "turbostat": {"type": "text"},
+                                        "vmstat": {"type": "text"},
+                                    }
+                                },
+                            },
+                        },
+                        "run": {
+                            "properties": {
+                                "config": {"type": "keyword"},
+                                "controller": {"type": "keyword"},
+                                "date": {"type": "date"},
+                                "end": {"type": "date"},
+                                "id": {"type": "keyword"},
+                                "iterations": {"type": "text"},
+                                "name": {"type": "keyword"},
+                                "script": {"type": "keyword"},
+                                "start": {"type": "date"},
+                                "toolsgroup": {"type": "keyword"},
+                                "user": {"type": "keyword"},
+                            }
+                        },
+                        "sosreports": {
+                            "type": "nested",
+                            "properties": {
+                                "hostname-f": {"type": "keyword"},
+                                "hostname-s": {"type": "keyword"},
+                                "inet": {
+                                    "type": "nested",
+                                    "properties": {
+                                        "ifname": {"type": "keyword"},
+                                        "ipaddr": {"type": "ip"},
+                                    },
+                                },
+                                "inet6": {
+                                    "type": "nested",
+                                    "properties": {
+                                        "ifname": {"type": "keyword"},
+                                        "ipaddr": {"type": "keyword"},
+                                    },
+                                },
+                                "md5": {"type": "keyword"},
+                                "name": {"type": "keyword"},
+                                "sosreport-error": {"type": "text"},
+                            },
+                        },
+                    },
+                },
+                version=5,
+            )
+        elif name == "result":
+            return Template(
+                name="result",
+                idxname="result-data",
+                template_name="unit-test.v5.result-data",
+                file="run.json",
+                template_pattern="unit-test.v5.result-data.*",
+                index_template="unit-test.v5.result-data.{year}-{month}",
+                settings={"none": False},
+                mappings={
+                    "_meta": {"version": "5"},
+                    "date_detection": "false",
+                    "properties": {
+                        "@timestamp": {"type": "date", "format": "dateOptionalTime"},
+                        "@timestamp_original": {"type": "keyword", "index": "false"},
+                        "result_data_sample_parent": {"type": "keyword"},
+                        "run": {
+                            "properties": {
+                                "id": {"type": "keyword"},
+                                "name": {"type": "keyword"},
+                            }
+                        },
+                        "iteration": {
+                            "properties": {
+                                "name": {"type": "keyword"},
+                                "number": {"type": "long"},
+                            }
+                        },
+                        "sample": {
+                            "properties": {
+                                "@idx": {"type": "long"},
+                                "name": {"type": "keyword"},
+                                "measurement_type": {"type": "keyword"},
+                                "measurement_idx": {"type": "long"},
+                                "measurement_title": {"type": "text"},
+                                "uid": {"type": "keyword"},
+                            }
+                        },
+                        "result": {
+                            "properties": {
+                                "@idx": {"type": "long"},
+                                "read_or_write": {"type": "long"},
+                                "value": {"type": "double"},
+                            }
+                        },
+                    },
+                },
+                version=5,
+            )
+        else:
+            return None
 
     with monkeypatch.context() as m:
         m.setattr(Template, "find", fake_find)
--- a/lib/pbench/test/unit/server/query_apis/test_index_mappings.py
+++ b/lib/pbench/test/unit/server/query_apis/test_index_mappings.py
@@ -1,0 +1,122 @@
+import pytest
+from http import HTTPStatus
+from sqlalchemy.exc import DatabaseError
+
+from pbench.server.database.models.template import Template
+
+
+@pytest.fixture()
+def database_error(monkeypatch):
+    def raise_db_error(name: str):
+        raise DatabaseError("DB Error")
+
+    with monkeypatch.context() as m:
+        m.setattr(Template, "find", raise_db_error)
+        yield
+
+
+class TestIndexMappings:
+    """
+    Unit testing for resources/IndexMappings class.
+
+    In a web service context, we access class functions mostly via the
+    Flask test client rather than trying to directly invoke the class
+    constructor and `post` service.
+    """
+
+    def test_run_template_query(self, client, server_config, find_template):
+        """
+        Check the construction of index mappings API and filtering of the
+        response body.
+        """
+        with client:
+            response = client.get(f"{server_config.rest_uri}/index/mappings/run")
+            assert response.status_code == HTTPStatus.OK
+            res_json = response.json
+            assert res_json == {
+                "@metadata": [
+                    "controller_dir",
+                    "file-date",
+                    "file-name",
+                    "file-size",
+                    "md5",
+                    "pbench-agent-version",
+                    "raw_size",
+                    "result-prefix",
+                    "satellite",
+                    "tar-ball-creation-timestamp",
+                    "toc-prefix",
+                ],
+                "authorization": ["access", "owner"],
+                "host_tools_info": [
+                    "hostname",
+                    "hostname-f",
+                    "hostname-s",
+                    "label",
+                    "tools",
+                ],
+                "run": [
+                    "config",
+                    "controller",
+                    "date",
+                    "end",
+                    "id",
+                    "iterations",
+                    "name",
+                    "script",
+                    "start",
+                    "toolsgroup",
+                    "user",
+                ],
+                "sosreports": [
+                    "hostname-f",
+                    "hostname-s",
+                    "inet",
+                    "inet6",
+                    "md5",
+                    "name",
+                    "sosreport-error",
+                ],
+            }
+
+    def test_result_template_query(self, client, server_config, find_template):
+        """
+        Check the construction of index mappings API and filtering of the
+        response body.
+        """
+        with client:
+            response = client.get(f"{server_config.rest_uri}/index/mappings/result")
+            assert response.status_code == HTTPStatus.OK
+            res_json = response.json
+            assert res_json == {
+                "iteration": ["name", "number"],
+                "result": ["@idx", "read_or_write", "value"],
+                "run": ["id", "name"],
+                "sample": [
+                    "@idx",
+                    "name",
+                    "measurement_type",
+                    "measurement_idx",
+                    "measurement_title",
+                    "uid",
+                ],
+            }
+
+    def test_with_no_index_document(self, client, server_config):
+        """
+        Check the index mappings API if there is no index document (specified by the index name in the URI)
+        present in the database.
+        """
+        with client:
+            response = client.get(f"{server_config.rest_uri}/index/mappings/run")
+            assert response.status_code == HTTPStatus.NOT_FOUND
+            assert response.json["message"] == "Mapping not found"
+
+    def test_with_db_error(self, client, server_config, database_error):
+        """
+        Check the index mappings API if there is an error connecting to sql database.
+        """
+        with client:
+            response = client.get(f"{server_config.rest_uri}/index/mappings/run")
+            assert response.status_code == HTTPStatus.INTERNAL_SERVER_ERROR
+            assert response.json["message"] == "Internal Server Error"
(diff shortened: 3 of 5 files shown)


lib/pbench/server/api/resources/query_apis/index_mappings.py
```diff
@@ -0,0 +1,45 @@
...
+        self.logger.info(mappings)
+        result = {}
+        for property in mappings["properties"]:
+            self.logger.info(property)
+            if "properties" in mappings["properties"][property]:
+                result[property] = list(
+                    mappings["properties"][property]["properties"].keys()
+                )
+
+        result.pop("authorization", None)
```
dbutenhof, line 42: Huh; that's an interesting point. The actual search will control access to datasets, though, so I'm not sure it's necessary. This introduces the idea of "Pbench metadata" fields we're more or less hiding from a client. Is that a direction we want to go? The dashboard may want to focus the search on features of the run, but possibly it'd be more appropriate for the dashboard effect to remove this if it doesn't want that to appear in the search dashboard setup pane?

dbutenhof (commented):
Good; a couple of comments, but a nice start.

Better documentation would help a lot; and of course you need unit tests.

lib/pbench/server/api/resources/query_apis/index_mappings.py
```diff
@@ -0,0 +1,45 @@
...
+        current_month = datetime.strptime(datetime.today().strftime("%Y-%m"), "%Y-%m")
+        uri_fragment = self._gen_month_range("run", current_month, current_month)
+        self.logger.info("index mapping key: {}", uri_fragment)
+        return {"path": f"/{uri_fragment}/_mapping", "kwargs": {}}
+
+    def postprocess(self, es_json: Dict[AnyStr, Any]) -> Dict[AnyStr, Any]:
+        """
+        Returns a dictionary of run documents properties.
+        """
+        index = list(es_json.keys())[0]
```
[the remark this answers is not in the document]
npalaska: Yep I need to add some comments here.

lib/pbench/server/api/resources/query_apis/index_mappings.py
npalaska: yeah I am not sure either, I thought this field is not useful to the dashboard anyway so why to return, but I also probably agree that it's a dashboard's decision whether to add this on search pane. However, even if it adds this on the pane how would that be useful because the search is anyway gonna be user access controlled.

lib/pbench/server/api/resources/query_apis/index_mappings.py
```diff
@@ -0,0 +1,45 @@
...
+        super().__init__(config, logger, Schema())
+
+    def assemble(self, json_data: Dict[AnyStr, Any]) -> Dict[AnyStr, Any]:
+        """
+        Get properties of the run-data document for the index containing the current month
+        """
+        current_month = datetime.strptime(datetime.today().strftime("%Y-%m"), "%Y-%m")
+        uri_fragment = self._gen_month_range("run", current_month, current_month)
+        self.logger.info("index mapping key: {}", uri_fragment)
+        return {"path": f"/{uri_fragment}/_mapping", "kwargs": {}}
```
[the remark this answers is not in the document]
npalaska: yep

lib/pbench/server/api/resources/query_apis/index_mappings.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    Get properties of the run-data document type on which we can search.
+    """
+
+    def __init__(self, config: PbenchServerConfig, logger: Logger):
+        super().__init__(config, logger, Schema())
+
+    def assemble(self, json_data: Dict[AnyStr, Any]) -> Dict[AnyStr, Any]:
+        """
+        Get properties of the run-data document for the index containing the current month
```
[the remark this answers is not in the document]
npalaska: Yeah but I modified the code to query against the template table so we dont need to build the indices on the server

lib/pbench/server/api/resources/query_apis/index_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+            mappings = template.mappings
+
+            result = {}
+            for property in mappings["properties"]:
+                if "properties" in mappings["properties"][property]:
+                    result[property] = list(
+                        mappings["properties"][property]["properties"].keys()
+                    )
+
+            result.pop("authorization", None)
```
dbutenhof, line 73: I'm still uncomfortable with "lying" like this. The more I think about it, the more I think we ought to return the complete "schema" and let the dashboard effect strip it out to avoid putting it on the field selector panel.

lib/pbench/test/unit/server/query_apis/test_index_mappings.py
```diff
@@ -0,0 +1,190 @@
+import pytest
+from http import HTTPStatus
+from sqlalchemy.exc import DatabaseError
+
+from pbench.server.database.models.template import Template
+
+
+@pytest.fixture()
+def run_template(monkeypatch, fake_mtime):
```
dbutenhof, line 9: Did you really need to duplicate the `find_template` fixture instead of just expanding it to include the full template body?

I'd generally prefer to avoid having multiple fixtures for "the same thing", especially because they're both mocking the run index entry.

lib/pbench/test/unit/server/query_apis/test_index_mappings.py
```diff
@@ -0,0 +1,190 @@
...
+        """
+        Check the construction of index mappings API and filtering of the
+        response body.
+        """
+        with client:
+            response = client.get(f"{server_config.rest_uri}/index/mappings")
+            assert response.status_code == HTTPStatus.OK
+            res_json = response.json
+            assert isinstance(res_json, dict)
+            assert len(res_json.keys()) == 4
```
dbutenhof, line 172: I'd like to verify the specific keys we expect, and the structure, rather than just the number of top-level keys.

dbutenhof (changes requested):
A couple of issues to think about. I dislike the duplication of `find_template`, and I'd like to see a better unit test validation of the response payload.

However, I really like the new DB-based query structure. It looks a bit odd to have it in the same directory with the Elasticsearch queries; however it replaces an Elasticsearch query, and I'm likely to do the same thing to `controllers/list` (going to the `datasets` table) at some point. In the end, I think it's better to keep them all together than to separate by implementation details that don't matter from the outside.

lib/pbench/server/api/resources/query_apis/index_mappings.py
webbnh: @dbutenhof, would it change your mind if the client were something other than the Dashboard?  That is, does this field need to be hidden from anyone other than a Dashboard User?

lib/pbench/test/unit/server/query_apis/test_index_mappings.py
```diff
@@ -0,0 +1,62 @@
...
+            res_json = response.json
+            assert isinstance(res_json, dict)
+            assert len(res_json.keys()) == 5
+            assert list(res_json.keys()) == [
+                "@metadata",
+                "authorization",
+                "host_tools_info",
+                "run",
+                "sosreports",
+            ]
```
webbnh, line 44: I'm not sure that there is any point to the assertion at line 36 -- I think it's safe to assume that `response.json` will yield a `dict`.  And, if you're going to explicitly check the keys returned, I don't think there is any point to checking the count of keys.

As for the third assertion, given that we know what the input mapping was, we should be able to predict exactly what the output will be, so I think we should be able to compare `res_json` directly to the full return value (which, presumably we would import from `conftest.py` somehow, either as a fixture or just a constant).  (If we _can't_ do that, then I think maybe there's a problem with this interface.)

lib/pbench/test/unit/server/query_apis/test_index_mappings.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def test_with_no_run_document(self, client, server_config):
+        """
+        Check the index mappings API if there is no run document present in the database.
+        """
+        with client:
+            response = client.get(f"{server_config.rest_uri}/index/mappings")
+            assert response.status_code == HTTPStatus.INTERNAL_SERVER_ERROR
+            assert response.json["message"] == "INTERNAL ERROR"
```
webbnh, line 53: This is a curious case.  Is it really true that there should never be a server deployment with no run document present in the database?  (So, part of server installation is to initialize the database with a run document?)

If so, this feels more like something which should be an assertion in the Server code or in an installation test, not a unit test.

lib/pbench/server/api/resources/query_apis/index_mappings.py
```diff
@@ -0,0 +1,84 @@
...
+                        mappings["properties"][property]["properties"].keys()
+                    )
+
+            # construct response object
+            return jsonify(result)
+        except Exception:
+            self.logger.exception(
+                "Exception occurred while querying mappings for run documents."
+            )
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
```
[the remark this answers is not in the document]
webbnh: If we do keep the `try` block here, do we really want to have just the one, catch-all case?  Or, are there specific cases that we should try to catch (like missing `"run"` document, or unable to connect to the database, et al.) which would ease debugging of issues?

lib/pbench/server/api/resources/query_apis/index_mappings.py
dbutenhof: Yes, @webbnh, that's my point; it's definitely not needed for the dashboard, but I don't feel entirely comfortable about hiding it from someone else.

For example, I've had thoughts about adding an `ELASTICSEARCH` *role* that allows privileged clients to continue using the direct `/api/v1/elasticsearch` query for more advanced (and implicitly `ADMIN` class) queries; if they want to see the document templates, we probably don't want to hide this.

But I'm not entirely sure what the use case it for allowing queries entirely outside out authorization framework (even with the special role, it sounds risky), and outside of that I'm not really sure how it would benefit anyone.

webbnh (changes requested):
How much latitude do we have around the contents of the response payload?  I think the dict-of-lists would be better if it were a dict-of-dicts, since the order of the items in the list might not be fixed (which would make life interesting on the receiving end!).  Beyond that, like Dave, I think we should validate the full payload.

It seems like our handling of errors is a little sparse.  That's OK, but in the future I suspect we'll thank ourselves if we make the code do a little more triage now.  (And, testing for the no-run-document-found case strikes me as odd.)

Other than those, I have some comments on smaller things, below.

lib/pbench/server/api/resources/query_apis/index_mappings.py
```diff
@@ -0,0 +1,84 @@
...
+    {
+        "@metadata":
+            [
+                "controller_dir",
+                "file-date",
+                "file-name",
+                "file-size",
+                "md5",
+                ...
+            ],
```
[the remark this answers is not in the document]
npalaska: This is just an example not necessarily an ordered list of keys, it lists all the features of run document index mapping keys. I am not sure if I understand it correctly but dashboard use this list on search setup pane where it uses features of the run documents. These features of run documents user can select as a part of `selected_fields` for their search query. This is how we defined in the server API documents under `IndexMappingQuery` https://docs.google.com/document/d/1c0bauJ9DM7MckFKus8XuTfkqZuLvVONnpnRDoM_TCik/edit#heading=h.5ie8tvlo5y0cqueryIndexMapping

lib/pbench/server/api/resources/query_apis/index_mappings.py
```diff
@@ -0,0 +1,84 @@
...
+                "hostname-f",
+                "sosreport-error",
+                "hostname-s",
+                ...
+            ]
+    }
+    """
+
+    def __init__(self, config: PbenchServerConfig, logger: Logger):
+        self.config = config
```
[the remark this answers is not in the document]
npalaska: Yeah this is not used, I should remove it.

lib/pbench/server/api/resources/query_apis/index_mappings.py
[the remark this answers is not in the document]
npalaska: Yeah It applies to 68 only, and the reason I kept it catch-all and return 500 is that if we encounter any exception querying a database that's automatically 500 for an end user. But you are probably right that in this case we dont need a try block to catch 500 as it will be normal result of unhandled exception

lib/pbench/test/unit/server/query_apis/test_index_mappings.py
npalaska: This is an interesting question, I also believe this to be a far extreme edge case. As far as I know when someone moves their data on the server (and considering there is no previous data on the server) the prep-shim picks up the data and the template database gets created on the server which has the run document index in it. 
@dbutenhof any thought?

lib/pbench/server/api/resources/query_apis/index_mappings.py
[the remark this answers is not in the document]
webbnh: Thanks for the documentation link.  As long as the client doesn't care about the ordering of the items in the lists, that's fine.

lib/pbench/test/unit/server/query_apis/test_index_mappings.py
dbutenhof: Right now, it *is* possible, although it's a startup edge case we'll probably want to close at some point.

The template DB table is filled in by the *first* call to the indexer, which means the first time a dataset is received via `PUT` or staged by `prep-shim`, gets unpacked, and marked for indexing. (Actually, it may be a bit better than that; server tools generate `server-report` documents even when they don't do anything, and I think the report generator code probably loads *all* of the templates; and the first tool run should happen pretty quickly after `crontab` gets started.)

Of course it's possible to connect a dashboard (or other client) and make queries on the active API before that. They won't return anything interesting, at best, but we should be robust.

One option would be to have something during install/configure run through the `PbenchTemplates` procedure before we open the server for business, so that it's populated. Another option would be for the API code to just fail with some meaningful message (sorry, waiting for inventory) until we actually index the first dataset.

portante (dismissed):
Can we consider a slightly different approach?  It is very likely that we'll be changing this API in the future both because it uses a name that belies the mechanism in which the data is stored, `index/mapping`, and because it is hard-coded to one kind of data, `run`.

As a proposal, can we make this be a "JSON schema" API, where a `GET` method returns the list of available JSON schema names on `json/schema`, and a `GET` on `json/schema/<name>` returns the JSON schema for that `name`?

Then the caller can do what it wants with the schema, and we don't have to put very specific behaviors for just returning keys in the API.

lib/pbench/server/api/resources/query_apis/index_mappings.py
dbutenhof: This feeds into Pete's issue about trying to generalize a "schema" API that would include our ES templates. I like that idea, but it doesn't necessarily have to invalidate this first attempt. I generally appreciated being "agile" and solving the problem in front of us and being prepared to refactor and rebuild later as we learn more. With that in mind, I think I'm going to resolve this conversation for now.

lib/pbench/server/api/resources/query_apis/index_mappings.py
```diff
@@ -58,12 +58,17 @@ class IndexMappings(Resource):
...
+        Return mapping properties of the document specified by the index name in the URI.
+        For example, user can get the run document properties by making a GET request on
+        index/mappings/run. Similarly other index documents can be fetched by making a GET
+        request on appropriate index names.
+        We fetch the mapping by querying the template database. If the template is not found
+        in the database NOT_FOUND error will be raised.
         """
         try:
-            template = Template.find("run")
+            template = Template.find(index_name)
```
dbutenhof, line 71: GitHub isn't letting me comment on line 84, so I'll put it here. Now that `index_name` is a parameter, your log message should include the actual index name we couldn't find, not say `Run document template`.

(I think the `abort` is OK as is, though; it's reasonable to expect them to understand that if they asked for `/index/mappings/foobar` then it means `foobar` wasn't an index template name.)

lib/pbench/test/unit/server/query_apis/test_index_mappings.py
```diff
@@ -30,7 +30,7 @@ def test_query(self, client, server_config, find_template):
         response body.
         """
         with client:
-            response = client.get(f"{server_config.rest_uri}/index/mappings")
+            response = client.get(f"{server_config.rest_uri}/index/mappings/run")
```
dbutenhof, line 33: I know, only the run `Template` is mocked, but ... it'd be good to verify that we can access a different template here. You can either duplicate the mock with a different name, or have the fixture create multiple template mocks and return them from a dict by name.

lib/pbench/server/api/resources/query_apis/index_mappings.py
npalaska: Fixed

lib/pbench/server/api/resources/query_apis/index_mappings.py
```diff
@@ -81,5 +81,5 @@ def get(self, index_name: AnyStr) -> Dict[AnyStr, Any]:
             # construct response object
             return jsonify(result)
         except TemplateNotFound:
-            self.logger.exception("Run document template not found in the database")
+            self.logger.exception("{} Run document template not found in the database", index_name)
```
dbutenhof, line 84: So, `result Run document template`? How about `"Document template {} not found in the database."`??

dbutenhof (approved):
This is looking good.

There's still a minor glitch in the log message, and I think we likely should be moving towards a more general schema API that subsumes this (without dropping all the type information... but very likely without exposing the Elasticsearch "type" keywords). I'm happy to be agile and save that more extreme change (in server and dashboard) for later when we have time to decide exactly what we want to expose and how.

dbutenhof:
Note: I would like to see this get in quickly, so we can, um, fix the dashboard.

I decided to experiment with SEARCH (particularly our repeated IP info in sosreports), and I've found that the latest dashboard code relying on the Elasticsearch passthrough `/api/v1/elasticsearch` is broken, and can't easily be fixed. Basically, Elasticsearch usually treats `GET` and `POST` as equivalent; but in the case of the `_mappings` query, there's an important difference.

With a `GET`, Elasticsearch returns the document template; with a `POST`, it returns the odd (and not particularly useful) response `{"acknowledged": "true"}`.

I tried to convert the dashboard `queryIndexMapping` back to `requests.get`, because our `ElasticBase` will pass through either; however, it appears that the Javascript `requests.get` doesn't pay attention to the `data` parameter, and it all falls down. I think that in order to solve this we'd need to add an Elasticsearch passthrough API parameter to specify the intended Elastic HTTP method, so the dashboard can `POST` to us a request for an Elasticsearch `GET`.

However, (unless we eventually want to expose and generalize the Elasticsearch passthrough), this seems kinda pointless as `queryIndexMapping` is the only API where this is an issue and we'll be changing the dashboard to use this new API instead as soon as it's submitted ... which seems like an easier path forward.

So if anyone else has any serious objections to this PR let's get them resolved quickly. (And I'd prefer that not include architecting a generalized "schema" API, because while it's a great idea I don't think we sufficiently understand the scope we'll want to support, and I really don't want to return raw Elasticsearch template syntax.)

dbutenhof:
Ah: you need to run `black` on /home/pbench/lib/pbench/server/api/resources/query_apis/index_mappings.py

Always run `black` and `flake8` before pushing a commit!

lib/pbench/server/api/resources/query_apis/index_mappings.py
npalaska: Fixed

dbutenhof (approved):

webbnh:
> with a `POST`, it returns the odd (and not particularly useful) response `{"acknowledged": "true"}`.

Um, @dbutenhof, I think that that is what ES returns when an operation (like `DELETE`) is successful...so, perhaps your `POST` did some sort of an _update_?...hopefully a no-op!...

webbnh (dismissed):
OK, I didn't see any deal-breakers, but, depending on how much hurry Dave is in, there are a couple of small things below which seem like they ought to be addressed.

lib/pbench/test/unit/server/conftest.py
```diff
@@ -222,17 +222,177 @@ def find_template(monkeypatch, fake_mtime):
...
+                            "properties": {
+                                "@idx": {"type": "long"},
+                                "read_or_write": {"type": "long"},
+                                "value": {"type": "double"},
+                            }
+                        },
+                    },
+                },
+                version=5,
+            )
```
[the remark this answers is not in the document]
dbutenhof: Hmm; it's a test fixture so it needs to be used cautiously... but, yeah, throwing an exception or failing with an `assert` would alert a future developer depending on a third index name that it's not implemented.

dbutenhof (approved):
> > with a `POST`, it returns the odd (and not particularly useful) response `{"acknowledged": "true"}`.
> 
> Um, @dbutenhof, I think that that is what ES returns when an operation (like `DELETE`) is successful...so, perhaps your `POST` did some sort of an _update_?...hopefully a no-op!...

Yes, I figured that out overnight; I commented on that in my scrum notes yesterday, but didn't get around to coming back here. 😁 

The `POST` is creating/updating a template for the index; since we're not passing any JSON body, apparently it's a "successful no-op".

lib/pbench/test/unit/server/conftest.py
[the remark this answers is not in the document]
webbnh: Actually, it's not a fixture -- it's a mock (which is installed by a fixture).  Thus, when raised, the exception would hit the CUT and appear to come from the database call during the test.

lib/pbench/test/unit/server/query_apis/test_index_mappings.py
```diff
@@ -0,0 +1,122 @@
...
+        with client:
+            response = client.get(f"{server_config.rest_uri}/index/mappings/run")
+            assert response.status_code == HTTPStatus.OK
+            res_json = response.json
+            assert res_json == {
+                "@metadata": [
+                    "controller_dir",
+                    "file-date",
+                    "file-name",
+                    "file-size",
```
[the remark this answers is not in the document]
npalaska: you mean returning RHS from the `find_template()`?

lib/pbench/test/unit/server/conftest.py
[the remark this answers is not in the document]
npalaska: fixed

lib/pbench/server/api/resources/query_apis/index_mappings.py
```diff
@@ -0,0 +1,87 @@
+from http import HTTPStatus
+from flask import jsonify
+from flask_restful import abort, Resource
+from logging import Logger
+from typing import Any, AnyStr, Dict
```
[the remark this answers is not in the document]
npalaska: fixed

dbutenhof (dismissed):

lib/pbench/test/unit/server/query_apis/test_index_mappings.py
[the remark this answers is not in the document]
webbnh: We have two values -- the one which the mock feeds to the CUT and the one which the CUT returns to the test -- and the test checks that the second is a specific transformation of the first.  Since the two values are intimately related, it seems like a bad idea to have them defined in separate modules far from each other.

Although I hesitate to suggest it, my instinct is that the second one should be dynamically generated by the test from the first one.  Yes, this probably means duplicating the code from the CUT and putting it in the test, which would seem like a violation of the DRY principal, but in this case it would be appropriate, because what we want is independent confirmation of the result.  The problem with this is, we'll be tempted (now and in the future) to cut and paste the code from the CUT into the test, which will result in the test passing even if the CUT is wrong!  Thus, this is probably not a good approach.

That leaves us with hand-crafting the output from the input (which is what you've done).  And, the only question then is, where do we put the result of those labors.  And, my answer is, put it as close to the input as practical.  For instance, rather than having the mock generate the input in-line, I would move that definition to be a class attribute or a module-global which the mock simply references.  And then I would put the output definition right next to it, so that anybody who wants to compare them can find them both easily (and, you can put a single block comment explaining what the two of them are and how they are used).

Since you mention it, it might be cool/handy to have `find_template()` return the RHS value.  Since the fixture doesn't actually return any other value (it just sets up and later removes the mock), you probably _could_ do that.  (I'm not sure -- it kind of depends on how the PyTest fixture value caching works -- but it seems plausible.)  The problem is, though, you have set up the mock with an embedded conditional whose value isn't available at test-setup, so you would probably have to split the fixture into multiple fixtures, one for each value that the mock should return.  However, this might actually be a good thing to do anyway.  (That is, the mock could validate its argument to ensure that it's the expected request and then return its value which is hardwired to correspond to that request, which would obviate the need for the `else` clause that I requested above.)

webbnh (dismissed):
Looks good to me, pending resolution of [this previous comment](https://github.com/distributed-system-analysis/pbench/pull/2382#discussion_r683052229).

(Also, @dbutenhof, you need to resolve [this conversation](https://github.com/distributed-system-analysis/pbench/pull/2382#discussion_r674001649).)

lib/pbench/test/unit/server/query_apis/test_index_mappings.py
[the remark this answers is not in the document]
dbutenhof: > Although I hesitate to suggest it, my instinct is that the second one should be dynamically generated by the test from the first one. 

I did something like this for publish; but since I was generating expected Elasticsearch responses from the document map, it wasn't directly duplicating Pbench code and that theoretical temptation doesn't exist. While "hand crafting" could be said to present a similar temptation, that is, copying and pasting the actual output of the Pbench code, the risk would be perhaps a bit smaller since it's easier to visually validate that the format is correct.

> Since you mention it, it might be cool/handy to have find_template() return the RHS value. Since the fixture doesn't actually return any other value (it just sets up and later removes the mock), you probably could do that.

Yes, you definitely *can* do that, and we've got several fixtures that do.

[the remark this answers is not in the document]
npalaska: Interesting, yeah I will fix this refactoring in another PR where I am already doing some test structure refactor.

dbutenhof (approved):

dbutenhof:
> As a proposal, can we make this be a "JSON schema" API, where a `GET` method returns the list of available JSON schema names on `json/schema`, and a `GET` on `json/schema/<name>` returns the JSON schema for that `name`?

I believe this is a great suggestion, but I don't think we understand a general use case for such an interface beyond the dashboard's current use of  the RUN document template for SEARCH. I would like to be able to expose various JSON schema documents, including Dataset templates, API request and response payloads, and possibly other things. We need to decide on a common format (e.g., standard JSON schema language, Elasticsearch template syntax, etc.)

For now, this is a great step towards separating the dashboard from direct Elasticsearch, as well as a *fix* for the current elasticsearch passthrough API `GET`/`POST` bug (which, with this PR, I believe is no longer a concern unless we eventually decide to formalize a privileged raw Elasticsearch passthrough)... so I'd like to get it in and we can step back and figure out what sort of "schema" we wish to expose and in what form then make another iteration later. (E.g., in a `/api/v2` interface layer.)

As such, I'm going to resolve this conversation and move it to an issue.

webbnh (approved):

PR #2358, opened 2021-06-16
Rename pbench-agent.cfg.example file

ndokos:
Fixes #2343.

We rename the example config file to pbench-agent.cfg. That allows
the basic operations of the pbench-agent to succeed without further
steps other than installing the RPM. However, moving results to the
server will not succeed, until a proper config file and ssh key file
are installed, so we still expect people to install using Ansible.

PR #2389, opened 2021-07-27
pbench-agent.cfg is no longer a ghost

### LINKED ISSUE #2343: pbench configuration file does not exists after installation
pbench-agent-0.69.9-24g90314ee7c.noarch

# pbench-register-tool-set
ERROR: failed to fetch tool set, default-tool-set, from the pbench-agent configuration file

Workaround:
cp pbench-agent.cfg.example pbench-agent.cfg

ndokos:
Fixes #2343 (again and finally)

The example config file was renamed in PR #2358, but there is still a problem: the %ghost line in the spec file is preventing its installation.  This PR fixes that.

### DIFF
--- a/agent/rpm/pbench-agent.spec.j2
+++ b/agent/rpm/pbench-agent.spec.j2
@@ -108,7 +108,6 @@ fi
 /%{installdir}/profile
 
 %ghost %attr(0400,pbench,pbench) /%{installdir}/id_rsa
-%ghost %attr(0664,pbench,pbench) /%{installdir}/config/pbench-agent.cfg
 
 %defattr(775,pbench,pbench,775)
 /%{installdir}/util-scripts


dbutenhof (approved):
Looks good. I had to check the documentation to be sure I understood what `%ghost` was doing, and I wonder if we should use `%config(noreplace)` instead so we don't overwrite a modified config file on upgrade... but that's a separate issue.

riya-17 (approved):

webbnh (approved):

agent/rpm/pbench-agent.spec.j2
```diff
@@ -108,7 +108,6 @@ fi
 /%{installdir}/profile
 
 %ghost %attr(0400,pbench,pbench) /%{installdir}/id_rsa
-%ghost %attr(0664,pbench,pbench) /%{installdir}/config/pbench-agent.cfg
```
portante, line 111: Shouldn't we be doing the same thing as we do as `main`?  See https://github.com/distributed-system-analysis/pbench/blob/main/agent/rpm/pbench-agent.spec.j2#L131.

portante: Tracking in #2392.

Issue #1679, opened 2020-05-22
Perf tool data with perf-archive captured can be HUGE


portante:
The `perf` tool data collects a `perf-archive` as part of the tool's `stop` operation.  This archive collection can result in a HUGE amount of repetitive data being collected.

It was observed in one run that a single `perf.data.archive.bz2` file was 147 MB in size.  With two nodes collecting `perf` tool data, across 20 iterations, with 3 samples in each iteration, the combined size was a whopping 17 GB (147 * 2 * 3 * 20 = 17,640 MB).

There are a few options we might want to consider:

 * Adding a perf tool option to not capture `perf-archive` output
   * and turning off the capturing of the `perf-archive` output by default
 * Adding a way to allow the user to remove the `perf.data.archive.bz2` files after a pbench benchmark finishes, but before the user executes `pbench-move-results`
   * This would allow a savvy user a chance to pick which iteration(s)/sample(s) they would want to keep, and for which nodes

PR #2390, opened 2021-07-27
Make the default tool set lighter

### LINKED ISSUE #1714: The "perf" tool should not be in the default tool set and should be removed, it generates a lot of data
I was wondering why pbench-move-results was taking so long for an fio run, and see that the perf tooling data takes up almost all the space utilized.

The fio run has 5 operations (read, write, randread, rand write, mixed) times 3 block sizes times 3 IO depths for a total of 45 iterations. 3 samples per iteration, makes a total of 135 samples.

Looking at the tooling data usage for 3 of the samples:

```
du -sk */tools-default/*/*
88      sample0/tools-default/worker-perf146/iostat
492     sample0/tools-default/worker-perf146/mpstat
142384  sample0/tools-default/worker-perf146/perf
3160    sample0/tools-default/worker-perf146/pidstat
4       sample0/tools-default/worker-perf146/postprocess.log
4164    sample0/tools-default/worker-perf146/proc-interrupts
184     sample0/tools-default/worker-perf146/proc-vmstat
2908    sample0/tools-default/worker-perf146/sar
4       sample0/tools-default/worker-perf146/stop.log
92      sample0/tools-default/worker-perf146/turbostat
88      sample1/tools-default/worker-perf146/iostat
492     sample1/tools-default/worker-perf146/mpstat
141444  sample1/tools-default/worker-perf146/perf
3200    sample1/tools-default/worker-perf146/pidstat
4       sample1/tools-default/worker-perf146/postprocess.log
4168    sample1/tools-default/worker-perf146/proc-interrupts
184     sample1/tools-default/worker-perf146/proc-vmstat
2908    sample1/tools-default/worker-perf146/sar
4       sample1/tools-default/worker-perf146/stop.log
88      sample1/tools-default/worker-perf146/turbostat
88      sample2/tools-default/worker-perf146/iostat
492     sample2/tools-default/worker-perf146/mpstat
142980  sample2/tools-default/worker-perf146/perf
3164    sample2/tools-default/worker-perf146/pidstat
4       sample2/tools-default/worker-perf146/postprocess.log
4164    sample2/tools-default/worker-perf146/proc-interrupts
184     sample2/tools-default/worker-perf146/proc-vmstat
2908    sample2/tools-default/worker-perf146/sar
4       sample2/tools-default/worker-perf146/stop.log
92      sample2/tools-default/worker-perf146/turbostat
```

The perf data dominates, roughly 140MB per sample, or a total of 140MB*135 = 18.9GB for an fio run. The "default" toolset should be broadly applicable to all users, and this is far too much data: users who want perf data should explicitly register it as a tool.

### LINKED ISSUE #2331: We should delete pidstat from the default tool set
To quote from chat:

```
...
I'm afraid that things like pidstat might wreak havoc on 2015's stability.  Maybe not, but ...

we have to turn that off by default
our pidstat implementation is awful, and pcp does a much better job.

I would not complain if it disappeared in the dead of night from the default toolset. ;)
```

ndokos:
Fixes #1714
Fixes #2331
Related: #1679

- Delete pidstat and perf from the default tool set.

- Define a "heavy" tool set that includes pidstat and perf.

- Define a "light" tool set that omits proc-interrupts from the
  default tool set.

Use the `--toolset` option of `pbench-register-tool-set` to choose
a non-default tool set.


agent/config/pbench-agent-default.cfg
```diff
@@ -29,7 +29,9 @@ rest_endpoint = api/v%(api_version)s
 server_rest_url = http://%(webserver)s/%(rest_endpoint)s
 
 [pbench/tools]
-default-tool-set = sar, iostat, mpstat, pidstat, proc-vmstat, proc-interrupts, turbostat, perf
+light-tool-set = sar, iostat, mpstat, proc-vmstat, turbostat
+default-tool-set = sar, iostat, mpstat, proc-vmstat, proc-interrupts, turbostat
```
dbutenhof, line 33: Do we really want to change the default? (Will that confuse or depress any existing users?)

In any case, maybe match light/heavy with a medium (even if it's the same now as the new default).

Or even set `default-tool-set = %(medium-tool-set)s` which would set up the three collections as baselines and make it really easy to customize the site default by referencing any of the three.

webbnh: Unless there is a reason to the contrary (i.e., if each set is a super-set of the previous set), I recommend defining each set in terms of the previous one, so that we don't inadvertently omit a tool.  (This also makes it more obvious to the reader what is different about each set.)

ndokos: PTAL at the latest commit: it's still a draft because of test failures, but I'll fix those and open it for final review if everything else looks OK.

webbnh (approved):
Looks good to me.

agent/config/pbench-agent-default.cfg
ndokos: > Do we really want to change the default? (Will that confuse or depress any existing users?)

Yes, that is the main point - see attached issues. We will need to let users know about it though. I added some comments to the PR comment on how to get back the heavy tool set, if required.

webbnh (approved):
GTG

dbutenhof (approved):

riya-17 (approved):
Looks Good

agent/config/pbench-agent-default.cfg
```diff
@@ -2,6 +2,10 @@
 version = 002
 pbench_web_server = pbench.example.com
 
+light-tool-set = sar, iostat, mpstat, proc-vmstat, turbostat
```
portante, line 5: I am not sure how "light" this is really, but `mpstat` and `sar` data overlap with both capturing CPU statistics; `turbostat` data is "Intel" CPUs only, and `proc-vmstat` is "a lot" of VM statistics.

Perhaps we can consider something like:

```
[pbench/tools]
legacy-tool-set = sar, iostat, mpstat, pidstat, proc-vmstat, proc-interrupts, turbostat, perf
light-tool-set = vmstat
medium-tool-set = %(light-tool-set)s, iostat, sar
default-tool-set = %(medium-tool-set)s
heavy-tool-set = %(medium-tool-set)s, mpstat, perf, pidstat, proc-interrupts, proc-vmstat, turbostat
```

The thinking is that `vmstat` gives you CPU, memory, and I/O stats (not networking per sae) for a very "light-weight" data collection, while "medium" gives you a healthy set of CPU (`vmstat`, `sar`), memory (`vmstat`, `sar`), disk (`sar`, `iostat`), networking (`sar`), with "heavy" rounding it out with the rest of the kitchen sink.

But even so, I'd rather round all this out using PCP/Prometheus mappings than tools names, but that is for another PR.

portante (changes requested):
This is a very good change.  Though I think it is worth having a discussion about what is in each tool set category.

agent/util-scripts/gold/pbench-register-tool-set/test-11.13.txt
```diff
@@ -17,6 +17,7 @@ Available tool sets from /var/tmp/pbench-test-utils/opt/pbench-agent/config/pben
 	default
 	heavy
 	light
+	medium
```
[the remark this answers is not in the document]
ndokos: NM - I see what's happening. So should the test config file just define a `foo-tool-set` and a `bar-tool-set` which will be available in addition to the newly added ones?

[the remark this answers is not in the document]
ndokos: Done.

PR #2397, opened 2021-08-03
Clean up linting requirements for `b0.69`

portante:
While reviewing PR #2374, a small clean up that can be broken out ahead of time.


dbutenhof (approved):

tox.ini
```diff
@@ -1,49 +1,56 @@
...
+    PY_COLORS
+    NO_COLORS
+    TERM
+    USER
 setenv =
-   VIRTUAL_ENV={envdir}
+    VIRTUAL_ENV = {envdir}
+    XDG_CACHE_HOME = {envdir}
+    SKIP_GENERATE_AUTHORS = 1
+    SKIP_WRITE_GIT_CHANGELOG = 1
```
webbnh, line 21: I believe that these are supposed to quiet PBR, but evidently they don't work.

tox.ini
```diff
@@ -1,49 +1,56 @@
...
 envars =
-   PATH = {env:PATH}:{toxinidir}/bin
+    PATH = {env:PATH}:{toxinidir}/bin
+commands_pre =
+    {toxinidir}/utils/detox {envdir}
```
webbnh, line 25: Aside from the most excellent name, I don't think the `detox` script is worth keeping:  it wraps a single command, and the effort of locating the script and feeding it the appropriate input is greater than the encapsulation is worth.  I would just replace this line with the command invocation and be done with it:
```
    find {envdir} -type f -path '*/__pycache__/*' -name '*.py[c|o]' -delete
```

webbnh (approved):
Let's move on.  (@portante, feel free to resolve these two conversations once you've read them.)

tox.ini
portante: I originally moved to the `detox` script so that I could run the find command myself separate from `tox` to work with and debug problems.  It was always a pain to open the `tox` file and grab the command instead of just invoking `utils/detox`.  And further, I was hoping that we could place other cleanups in there as well.

tox.ini
portante: Bummer.  Originally they did work, but maybe there is something about the `tox --current-env` that defeats them.

PR #2403, opened 2021-08-05
Tell RPM pbench-agent.cfg file is a `config` file

### LINKED ISSUE #2392: Use `%config(noreplace)` for the `pbench-agent.cfg` file
Shouldn't we be doing the same thing as we do as `main`?  See https://github.com/distributed-system-analysis/pbench/blob/main/agent/rpm/pbench-agent.spec.j2#L131.

_Originally posted by @portante in https://github.com/distributed-system-analysis/pbench/pull/2389#discussion_r679105234_

portante:
And while we are at it, update how we set all the permissions when installing.

Fixes #2392.


agent/rpm/pbench-agent.spec.j2
```diff
@@ -95,26 +95,35 @@ fi
...
 /%{installdir}/VERSION
 /%{installdir}/SEQNO
 /%{installdir}/SHA1
 /%{installdir}/profile
 
-%ghost %attr(0400,pbench,pbench) /%{installdir}/id_rsa
+%config(noreplace) %attr(0664,root,root) /%{installdir}/config/pbench-agent.cfg
 
-%defattr(775,pbench,pbench,775)
+%defattr(775,root,root,775)
```
dbutenhof, line 111: Why `775`?? Is there really anything here that's supposed to be writable?

dbutenhof (dismissed):

agent/rpm/pbench-agent.spec.j2
portante: Do you mean, why not `555` (`-r-x`)?

dbutenhof: Yes; for directories that are executable scripts, `555` would seem appropriate. I don't see why the directories need `w` either?

portante: Only historical, as far as I can tell.

agent/rpm/pbench-agent.spec.j2
```diff
@@ -95,26 +95,35 @@ fi
...
+%attr(664,root,root) /%{installdir}/util-scripts/README
 /%{installdir}/tool-scripts
+%attr(664,root,root) /%{installdir}/tool-scripts/perf.README
+%attr(664,root,root) /%{installdir}/tool-scripts/README
 /%{installdir}/bench-scripts
+%attr(664,root,root) /%{installdir}/bench-scripts/pbench-fio.md
+%attr(664,root,root) /%{installdir}/bench-scripts/pbench-uperf.md
+%attr(664,root,root) /%{installdir}/bench-scripts/postprocess/BenchPostprocess.pm
+%attr(664,root,root) /%{installdir}/bench-scripts/templates/fio.job
+%attr(664,root,root) /%{installdir}/bench-scripts/templates/fio-shared-fs.job
```
webbnh, line 122: Why set the default permissions for files to `775` and then override it to `664` on _every_ (file) entry?  (Why not just set the default to what you want and then omit the overrides?)

Also, it might be nice to segregate the files and directories, like you did at lines 98-107.

agent/rpm/pbench-agent.spec.j2
```diff
@@ -95,26 +95,35 @@ fi
...
 
 # stockpile
-%defattr(775,pbench,pbench,775)
+%defattr(664,root,root,775)
 /%{installdir}/stockpile
-
+%attr(775,root,root) /%{installdir}/stockpile/import_roles.sh
+%attr(775,root,root) /%{installdir}/stockpile/roles/openshift-cluster-topology/files/openshift_config_scraper.py
+%attr(775,root,root) /%{installdir}/stockpile/roles/openstack_common/files/openstack-config-parser.py
```
webbnh, line 129: Ditto the previous comment, except swap the permissions masks.  😀

webbnh (dismissed):
Looks OK, but the (lack of) use of the defaults is kind of weird.

agent/rpm/pbench-agent.spec.j2
dbutenhof: There are more files in `/bench-scripts`, which will be subject to the default; only the non-executable files in the directory are overridden to exclude `xxx` permissions.

Although it's kinda interesting that the Perl module doesn't need execute? I suppose it's never run directly, only read by Perl; somehow it still feels odd. 🤷🏻‍♂️

agent/rpm/pbench-agent.spec.j2
dbutenhof: Apparently, this works the other way around, with only the executable files overridden from the directory default.

I still question the `ww-` here, since I don't think we want hardly any of our files (and certainly none of our scripts) to be modified, and I'm not crazy about allowing modifying directories by default either.

agent/rpm/pbench-agent.spec.j2
webbnh: > There are more files in /bench-scripts, which will be subject to the default; only the non-executable files in the directory are overridden to exclude xxx permissions.

Ah.  Excellent.  Thank you.

> Although it's kinda interesting that the Perl module doesn't need execute? I suppose it's never run directly, only read by Perl

Yeah, it's not a "script" its an "include file".  🙂

agent/rpm/pbench-agent.spec.j2
webbnh: Yeah, I'm not a fan of granting write-access where it's not needed, but, since we're granting it only to `root` and the guys in his group, I'm telling myself that it's not worth quibbling over.

agent/rpm/pbench-agent.spec.j2
portante: What is the concern here?

agent/rpm/pbench-agent.spec.j2
portante: As above, these permissions are historical.

I've always wanted to change them.

So ... done.

portante: See also #2407.

agent/rpm/pbench-agent.spec.j2
webbnh: I guess there is none.  🙂 
(I should have marked this as resolved after Dave's excellent answer.)

webbnh (dismissed):
Looks OK, AFAICT, but I have one question:  why do we bother with an empty file (`agent/bench-scripts/templates/fio-shared-fs.job`) -- storing it in the repo and including it in the package -- when we could trivially generate it during installation?

portante:
> Question: why do we bother with an empty file (`agent/bench-scripts/templates/fio-shared-fs.job`) -- storing it in the repo and including it in the package -- when we could trivially generate it during installation?

Not sure what you mean.  The file is not empty, but this commit changes the permissions in the git repo so that it is not executable.

agent/rpm/pbench-agent.spec.j2
```diff
@@ -95,26 +95,35 @@ fi
...
-/%{installdir}/base
 
+/%{installdir}/base
 /%{installdir}/VERSION
 /%{installdir}/SEQNO
 /%{installdir}/SHA1
 /%{installdir}/profile
 
-%ghost %attr(0400,pbench,pbench) /%{installdir}/id_rsa
+%config(noreplace) %attr(0664,root,root) /%{installdir}/config/pbench-agent.cfg
```
ndokos, line 109: Do we not still want the `%ghost` on the `id_rsa file?

ndokos (changes requested):

agent/rpm/pbench-agent.spec.j2
portante: We don't have it in `main` where we still have support for the SSH operations.

dbutenhof: If this agent installation is their only copy of the server's `pbench` private key they're probably in trouble. Since we don't *install* one, we probably don't have to protect it against overwriting, either.

dbutenhof (dismissed):

agent/rpm/pbench-agent.spec.j2
portante: Right.  RPM does not delete directories that have files in them.

webbnh:
> The file is not empty

I want to believe you, but it seems like GitHub disagrees:

Screen Shot 2021-08-05 at 3 30 42 PM

Am I misunderstanding?

portante:
https://github.com/portante/pbench/blob/config-noreplace/agent/bench-scripts/templates/fio-shared-fs.job

webbnh:
So, then, what is empty is the _differences file_, not the source file...there's a UI/UX challenge for GitHub....

webbnh (approved):

dbutenhof (approved):
@ndokos has one conversation to resolve; aside from that it looks (still) good.

ndokos (approved):

PR #2404, opened 2021-08-05
Fix stand-alone use of `pbench-make-result-tb`

### LINKED ISSUE #2360: pbench-make-result-tb won't create tarball
Version: pbench-agent `0.69.8-20g90314ee7c`

`pbench-make-result-tb` won't create tarball from a target directory.

```
# pbench-make-result-tb --result-dir=/var/lib/pbench-agent/fio__2021.06.11T14.26.42/ --target-dir=./tardir/
```

Command exits with 0 exit code, does not emit any warning or traceback, but does not create tarball.

portante:
Fixes #2360.

### DIFF
--- a/agent/util-scripts/pbench-make-result-tb
+++ b/agent/util-scripts/pbench-make-result-tb
@@ -102,13 +102,14 @@ if [[ ! -d "${target_dir}" ]]; then
     exit 1
 fi
 
-controller_dir="$(basename -- "${target_dir}")"
+controller_dir=$(basename -- "${target_dir}")
+full_result_dir=$(realpath -e "${result_dir}")
+full_target_dir=$(realpath -e "${target_dir}")
 
 # We can now start building the target tarball
 
 # Move into pbench run collection directory
-full_result_dir=$(realpath -e ${result_dir})
-pushd $(dirname ${full_result_dir}) > /dev/null
+cd $(dirname ${full_result_dir}) > /dev/null
 if [[ ${?} -ne 0 ]]; then
     exit 1
 fi
@@ -177,7 +178,7 @@ printf -- "%s" "$result_size" | pbench-add-metalog-option ${mdlog} run raw_size
 ts=$(timestamp)
 printf -- "%s" "$ts" | pbench-add-metalog-option ${mdlog} pbench tar-ball-creation-timestamp
 
-tarball="${target_dir}/${pbench_run_name}.tar.xz"
+tarball="${full_target_dir}/${pbench_run_name}.tar.xz"
 if [[ "${xz_single_threaded}" == "1" ]]; then
     tar_cmd="tar --create --force-local -xz \"${pbench_run_name}\""
 else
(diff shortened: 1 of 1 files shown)


dbutenhof (dismissed):

ndokos (changes requested):
I think there is still a problem on l.199: `${tarballmd5}` includes the directory component, but after we do the `pushd`, we only want the final component.

webbnh (approved):
> I think there is still a problem on l.199: ${tarballmd5} includes the directory component, but after we do the pushd, we only want the final component.

```
197  pushd $(dirname ${tarball}) > /dev/null 2>&1
198  md5sum "$(basename ${tarball})" > "${tarballmd5}"
```

That looks mostly OK to me -- I don't think that having the full path in `tarballmd5` (which it does) hurts anything.

But, unlike at line 111-114, we're not checking that the `pushd` was successful, and we're eating `stderr`...we might want to make the line 111 change here as well.

dbutenhof (dismissed):

webbnh (dismissed):

ndokos:
Calling it with a relative pathname as is done in  #2360: `--target-dir=./tardir/` breaks l.199: it is running  *in* `./tardir` and it tries to save the output to `./tardir/foo.tar.xz.md5.check`. It's in the wrong place for that. So we either have to stipulate absolute pathnames or we got to fix `${tarballmd5}`.
`

agent/util-scripts/pbench-make-result-tb
```diff
@@ -194,7 +194,7 @@ fi
 tarballmd5="${tarball}.md5.check"
 # We need to calculate the md5 sum in the temp directory
 # in order to get the filename right.
-pushd $(dirname ${tarball}) > /dev/null 2>&1
+pushd $(dirname ${tarball}) > /dev/null
 md5sum "$(basename ${tarball})" > "${tarballmd5}"
```
ndokos, line 198: This is still broken if a relative pathname is specified for `target-dir` as is done in #2360: we are *in* `./tardir` and we try save the md5 into `.tardir/<mumble>.tar.xz.md5.check`.

ndokos (changes requested):

agent/util-scripts/pbench-make-result-tb
webbnh: Good catch!

One option would be to inflict `realpath(1)` on the input so that we know that we have an absolute, existing path with no weird components.  The other, as you've suggested, would be to use `basename(1)` to get rid of the path part.

Now...should one of us take over this PR and push it through, or do we want to wait for Peter's return?  (I suspect that we would rather not wait on getting the next 0.69 release out, and we would prefer to have this update in it.)

webbnh: OK, I think this is fixed, now.

webbnh:
I've added another commit to the branch which hopefully addresses the problem with relative paths.

I made three small changes.  The key one was:
  - add a `full_target_dir` variable which expands the input argument to an absolute path;
  - use that to generate the file paths for the tarball and the MD5 file, so that when we change directories we use an absolute path to get there (since, in the second change, we're coming from an already-changed directory and the original relative path is no longer useful); and
  - finally use only the filename part of the MD5 file, since we're now in the appropriate directory (although, now that we have the correct, absolute path, this change isn't strictly need, but I made it first, so I left it in).

The other changes were to gather the processed argument variable assignments together and to replace the `pushd` calls with `cd`, since we never `popd`.  (Oh, and I got rid of a useless set of quotes in the process.)

ndokos (approved):

dbutenhof (approved):

PR #2405, opened 2021-08-05
Fix stand-alone use of `pbench-make-result-tb`

### LINKED ISSUE #2360: pbench-make-result-tb won't create tarball

portante:
Forward port of PR #2404, which fixes #2360.

### DIFF
--- a/agent/util-scripts/pbench-make-result-tb
+++ b/agent/util-scripts/pbench-make-result-tb
@@ -84,7 +84,7 @@ while true; do
 	    break;
 	    ;;
 	*)
- 	    printf "\n%s: you specified an invalid option or an unexpected argument: \"${opt}\"\n\n" "${script_name}" >&2
+	    printf "\n%s: you specified an invalid option or an unexpected argument: \"${opt}\"\n\n" "${script_name}" >&2
 	    usage >&2
 	    exit 1
 	    ;;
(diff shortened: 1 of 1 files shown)


ndokos (changes requested):
l.199 has the same problem as in #2404 I think.

webbnh (dismissed):
As I [said](https://github.com/distributed-system-analysis/pbench/pull/2404#pullrequestreview-723715944) in #2404, I think Nick's concern is accurate but not a problem; however, I think that the change made at line 111 should presumably be made to line 198 as well.

webbnh (dismissed):

ndokos (changes requested):
As in #2404, this will not work if `target-dir` is a relative pathname.

webbnh:
I forward-ported the late-breaking changes from #2404, as well as two other items:
- I removed an errant blank character from the beginning of a line (we use hard-tabs for indentation in this file)
- I removed the quotes around an invocation of `basename` -- since we assign the results directly to a shell variable, we don't need to protect them against word-splitting.

dbutenhof (dismissed):
I think that looks as it should ...

ndokos (dismissed):

riya-17 (approved):
It looks good

dbutenhof (approved):

PR #2413, opened 2021-08-11
Remove redundant query API test code

npalaska:
This PR tries to remove some of the redundant code in our query API tests

### DIFF
--- a/lib/pbench/test/unit/server/query_apis/commons.py
+++ b/lib/pbench/test/unit/server/query_apis/commons.py
@@ -186,6 +186,7 @@ class Commons:
         for key, p in self.cls_obj.schema.parameters.items():
             # Modify date/time key in the payload to make it look invalid
             if p.type == ParamType.DATE and key in self.payload:
+                original_date_value = self.payload[key]
                 self.payload[key] = "2020-19"
                 response = client.post(
                     server_config.rest_uri + self.pbench_endpoint,
@@ -197,6 +198,7 @@ class Commons:
                     response.json.get("message")
                     == "Value '2020-19' (str) cannot be parsed as a date/time string"
                 )
+                self.payload[key] = original_date_value
 
     def test_empty_query(
         self,
--- a/lib/pbench/test/unit/server/query_apis/commons.py
+++ b/lib/pbench/test/unit/server/query_apis/commons.py
@@ -1,3 +1,4 @@
+import itertools
 import pytest
 import requests
 
@@ -17,7 +18,7 @@ class Commons:
     """
 
     # Declare the common empty search response payload that subclass can use.
-    empty_es_response_payload = {
+    EMPTY_ES_RESPONSE_PAYLOAD = {
         "took": 1,
         "timed_out": False,
         "_shards": {"total": 1, "successful": 1, "skipped": 0, "failed": 0},
@@ -59,6 +60,18 @@ class Commons:
             index += f"{idx}{d},"
         return index
 
+    def date_range(self, start: AnyStr, end: AnyStr) -> list:
+        """
+        Builds list of range of dates between start and end
+        It expects the date to look like YYYY-MM
+        """
+        y, m_start = int(start[:4]), int(start[5:])
+        y, m_end = int(end[:4]), int(end[5:])
+        date_range = []
+        for month in range(m_start, m_end + 1):
+            date_range.append(f"{y}-{month:02}")
+        return date_range
+
     def test_non_accessible_user_data(self, client, server_config, pbench_token):
         """
         Test behavior when Authorization header does not have access to other user's data
@@ -66,7 +79,7 @@ class Commons:
         # The pbench_token fixture logs in as user "drb"
         # Trying to access the data belong to the user "pp"
         if "user" not in self.cls_obj.schema.parameters.keys():
-            pytest.skip("skipping non accessible user data test")
+            pytest.skip("skipping " + self.test_non_accessible_user_data.__name__)
         self.payload["user"] = "pp"
         response = client.post(
             server_config.rest_uri + self.pbench_endpoint,
@@ -85,7 +98,9 @@ class Commons:
         Test behavior when expired Authorization header provided
         """
         if "user" not in self.cls_obj.schema.parameters.keys():
-            pytest.skip("skipping accessing user data with invalid token test")
+            pytest.skip(
+                "skipping " + self.test_accessing_user_data_with_invalid_token.__name__
+            )
         # valid token logout
         response = client.post(
             server_config.rest_uri + "/logout",
@@ -111,43 +126,72 @@ class Commons:
         assert response.status_code == HTTPStatus.BAD_REQUEST
         assert response.json.get("message") == "Invalid request payload"
 
-    def missing_keys(self, client, server_config, keys, user_ok, pbench_token):
+    def test_missing_keys(self, client, server_config, user_ok, pbench_token):
         """
         Test behavior when JSON payload does not contain all required keys.
 
         Note that Pbench will silently ignore any additional keys that are
         specified but not required.
        """
+
+        def missing_key_helper(keys):
+            response = client.post(
+                server_config.rest_uri + self.pbench_endpoint,
+                headers={"Authorization": "Bearer " + pbench_token},
+                json=keys,
+            )
+            assert response.status_code == HTTPStatus.BAD_REQUEST
+            missing = [k for k in required_keys if k not in keys]
+            assert (
+                response.json.get("message")
+                == f"Missing required parameters: {','.join(missing)}"
+            )
+
+        parameter_items = self.cls_obj.schema.parameters.items()
+
         required_keys = [
             key
             for key, parameter in self.cls_obj.schema.parameters.items()
             if parameter.required
         ]
-        response = client.post(
-            server_config.rest_uri + self.pbench_endpoint,
-            headers={"Authorization": "Bearer " + pbench_token},
-            json=keys,
-        )
-        assert response.status_code == HTTPStatus.BAD_REQUEST
-        missing = [k for k in required_keys if k not in keys]
-        assert (
-            response.json.get("message")
-            == f"Missing required parameters: {','.join(missing)}"
-        )
+
+        all_combinations = []
+        for r in range(1, len(parameter_items) + 1):
+            combinations_object = list(itertools.combinations(parameter_items, r))
+            for item in combinations_object:
+                tmp_req_keys = []
+                for element in item:
+                    if element[1].required:
+                        tmp_req_keys.append(element[0])
+                if not sorted(tmp_req_keys) == sorted(required_keys):
+                    all_combinations.append(item)
+
+        for items in all_combinations:
+            keys = {}
+            for item in items:
+                if item[1].type == ParamType.ACCESS:
+                    keys[item[0]] = "public"
+                elif item[1].type == ParamType.DATE:
+                    keys[item[0]] = "2020"
+                else:
+                    keys[item[0]] = "foobar"
+
+            missing_key_helper(keys)
+
+        # Test for random non-existent key
+        missing_key_helper({"notakey": None})
 
     def test_bad_dates(self, client, server_config, user_ok, pbench_token):
         """
         Test behavior when a bad date string is given
         """
         parameter_types = [
-            parameter.type
-            for _, parameter in self.cls_obj.schema.parameters.items()
-            if parameter.required
+            parameter.type for _, parameter in self.cls_obj.schema.parameters.items()
         ]
-        if ParamType.DATE not in parameter_types or not all(
-            k in self.payload for k in ("start", "end")
+        if ParamType.DATE not in parameter_types or any(
+            k not in self.payload for k in ("start", "end")
         ):
-            pytest.skip("skipping the bad date test")
+            pytest.skip("skipping " + self.test_bad_dates.__name__)
 
         # Modify start and end time in the payload to make it look invalid
         self.payload["start"] = "2020-12"
@@ -178,12 +222,14 @@ class Commons:
         fixture.
         """
         if not self.empty_es_response_payload or not self.elastic_endpoint:
-            pytest.skip("skipping the empty query test")
+            pytest.skip("skipping " + self.test_empty_query.__name__)
         expected_status = HTTPStatus.OK
         if build_auth_header["header_param"] != "valid":
             expected_status = HTTPStatus.FORBIDDEN
 
-        index = self.build_index(server_config, ("2020-08", "2020-09", "2020-10"))
+        index = self.build_index(
+            server_config, self.date_range(self.payload["start"], self.payload["end"])
+        )
         response = query_api(
             self.pbench_endpoint,
             self.elastic_endpoint,
@@ -230,12 +276,14 @@ class Commons:
         Check that an exception in calling Elasticsearch is reported correctly.
         """
         if not self.elastic_endpoint:
-            pytest.skip("skipping the http exception test")
+            pytest.skip("skipping " + self.test_http_exception.__name__)
 
         # Make the start and end time in payload the same to result in an exception
         self.payload["end"] = self.payload["start"]
 
-        index = self.build_index(server_config, ("2020-08",))
+        index = self.build_index(
+            server_config, self.date_range(self.payload["start"], self.payload["end"])
+        )
         query_api(
             self.pbench_endpoint,
             self.elastic_endpoint,
@@ -255,12 +303,11 @@ class Commons:
         response.raise_for_status() and Pbench handlers.
         """
         if not self.elastic_endpoint:
-            pytest.skip("skipping the http error test")
+            pytest.skip("skipping " + self.test_http_error.__name__)
 
-        # Make the start and end time in payload the same to result in an error
-        self.payload["end"] = self.payload["start"]
-
-        index = self.build_index(server_config, ("2020-08",))
+        index = self.build_index(
+            server_config, self.date_range(self.payload["start"], self.payload["end"])
+        )
         query_api(
             self.pbench_endpoint,
             self.elastic_endpoint,
--- a/lib/pbench/test/unit/server/query_apis/commons.py
+++ b/lib/pbench/test/unit/server/query_apis/commons.py
@@ -1,0 +1,233 @@
+import pytest
+import requests
+from http import HTTPStatus
+from typing import Any, AnyStr, Dict
+
+
+class Commons:
+    """
+    Unit testing for all the elasticsearch resources class.
+
+    In a web service context, we access class functions mostly via the
+    Flask test client rather than trying to directly invoke the class
+    constructor and `post` service.
+    """
+
+    def _setup(
+        self,
+        pbench_endpoint: AnyStr = None,
+        elastic_endpoint: AnyStr = None,
+        payload: [AnyStr, Any] = None,
+        bad_date_payload: [AnyStr, Any] = None,
+        error_payload: Dict[AnyStr, Any] = None,
+        empty_response_payload: Dict[AnyStr, Any] = None,
+    ):
+        self.pbench_endpoint = pbench_endpoint
+        self.elastic_endpoint = elastic_endpoint
+        self.payload = payload
+        self.bad_date_payload = bad_date_payload
+        self.error_payload = error_payload
+        self.empty_response_payload = empty_response_payload
+
+    def build_index(self, server_config, dates):
+        """
+        Build the index list for query
+
+        Args:
+            dates (iterable): list of date strings
+        """
+        idx = server_config.get("Indexing", "index_prefix") + ".v6.run-data."
+        index = "/"
+        for d in dates:
+            index += f"{idx}{d},"
+        return index
+
+    def test_non_accessible_user_data(self, client, server_config, pbench_token):
+        """
+        Test behavior when Authorization header does not have access to other user's data
+        """
+        # The pbench_token fixture logs in as user "drb"
+        # Trying to access the data belong to the user "pp"
+        if not self.payload.get("user", None):
+            pytest.skip("skipping non accessible user data test")
+        self.payload["user"] = "pp"
+        response = client.post(
+            f"{server_config.rest_uri}{self.pbench_endpoint}",
+            headers={"Authorization": "Bearer " + pbench_token},
+            json=self.payload,
+        )
+        assert response.status_code == HTTPStatus.FORBIDDEN
+
+    @pytest.mark.parametrize(
+        "user", ("drb", "pp"),
+    )
+    def test_accessing_user_data_with_invalid_token(
+        self, client, server_config, pbench_token, user
+    ):
+        """
+        Test behavior when expired Authorization header provided
+        """
+        if not self.payload.get("user", None):
+            pytest.skip("skipping accessing user data with invalid token test")
+        # valid token logout
+        response = client.post(
+            f"{server_config.rest_uri}/logout",
+            headers=dict(Authorization="Bearer " + pbench_token),
+        )
+        assert response.status_code == HTTPStatus.OK
+        self.payload["user"] = user
+        response = client.post(
+            f"{server_config.rest_uri}{self.pbench_endpoint}",
+            headers={"Authorization": "Bearer " + pbench_token},
+            json=self.payload,
+        )
+        assert response.status_code == HTTPStatus.FORBIDDEN
+
+    def test_missing_json_object(self, client, server_config, pbench_token):
+        """
+        Test behavior when no JSON payload is given
+        """
+        response = client.post(
+            f"{server_config.rest_uri}{self.pbench_endpoint}",
+            headers={"Authorization": "Bearer " + pbench_token},
+        )
+        assert response.status_code == HTTPStatus.BAD_REQUEST
+        assert response.json.get("message") == "Invalid request payload"
+
+    def missing_keys(self, client, server_config, keys, user_ok, pbench_token):
+        """
+        Test behavior when JSON payload does not contain all required keys.
+
+        Note that Pbench will silently ignore any additional keys that are
+        specified but not required.
+       """
+        response = client.post(
+            f"{server_config.rest_uri}{self.pbench_endpoint}",
+            headers={"Authorization": "Bearer " + pbench_token},
+            json=keys,
+        )
+        assert response.status_code == HTTPStatus.BAD_REQUEST
+        missing = [k for k in self.required_keys if k not in keys]
+        assert (
+            response.json.get("message")
+            == f"Missing required parameters: {','.join(missing)}"
+        )
+
+    def test_bad_dates(self, client, server_config, user_ok, pbench_token):
+        """
+        Test behavior when a bad date string is given
+        """
+        if not self.bad_date_payload:
+            pytest.skip("skipping the bad date test")
+        response = client.post(
+            f"{server_config.rest_uri}{self.pbench_endpoint}",
+            headers={"Authorization": "Bearer " + pbench_token},
+            json=self.bad_date_payload,
+        )
+        assert response.status_code == HTTPStatus.BAD_REQUEST
+        assert (
+            response.json.get("message")
+            == "Value '2020-19' (str) cannot be parsed as a date/time string"
+        )
+
+    def test_empty_query(
+        self,
+        client,
+        server_config,
+        query_api,
+        user_ok,
+        find_template,
+        build_auth_header,
+    ):
+        """
+        Check proper handling of a query resulting in no Elasticsearch matches.
+        The test will run thrice with different values of the build_auth_header
+        fixture.
+        """
+        if not self.empty_response_payload or not self.elastic_endpoint:
+            pytest.skip(
+                "skipping the empty query test since the empty response payload is not set"
+            )
+        expected_status = HTTPStatus.OK
+        if build_auth_header["header_param"] != "valid":
+            expected_status = HTTPStatus.FORBIDDEN
+
+        index = self.build_index(server_config, ("2020-08", "2020-09", "2020-10"))
+        response = query_api(
+            self.pbench_endpoint,
+            self.elastic_endpoint,
+            self.payload,
+            index,
+            expected_status,
+            headers=build_auth_header["header"],
+            status=HTTPStatus.OK,
+            json=self.empty_response_payload,
+        )
+        assert response.status_code == expected_status
+        if response.status_code == HTTPStatus.OK:
+            assert response.json == []
+
+    @pytest.mark.parametrize(
+        "exceptions",
+        (
+            {
+                "exception": requests.exceptions.ConnectionError(),
+                "status": HTTPStatus.BAD_GATEWAY,
+            },
+            {
+                "exception": requests.exceptions.Timeout(),
+                "status": HTTPStatus.GATEWAY_TIMEOUT,
+            },
+            {
+                "exception": requests.exceptions.InvalidURL(),
+                "status": HTTPStatus.INTERNAL_SERVER_ERROR,
+            },
+            {"exception": Exception(), "status": HTTPStatus.INTERNAL_SERVER_ERROR},
+            {"exception": ValueError(), "status": HTTPStatus.INTERNAL_SERVER_ERROR},
+        ),
+    )
+    def test_http_exception(
+        self,
+        server_config,
+        query_api,
+        exceptions,
+        user_ok,
+        find_template,
+        pbench_token,
+    ):
+        """
+        Check that an exception in calling Elasticsearch is reported correctly.
+        """
+        if not self.elastic_endpoint or not self.error_payload:
+            pytest.skip("skipping the http exception test")
+        index = self.build_index(server_config, ("2020-08",))
+        query_api(
+            self.pbench_endpoint,
+            self.elastic_endpoint,
+            self.error_payload,
+            index,
+            exceptions["status"],
+            body=exceptions["exception"],
+            headers={"Authorization": "Bearer " + pbench_token},
+        )
+
+    @pytest.mark.parametrize("errors", (400, 500, 409))
+    def test_http_error(
+        self, server_config, query_api, user_ok, find_template, pbench_token, errors
+    ):
+        """
+        Check that an Elasticsearch error is reported correctly through the
+        response.raise_for_status() and Pbench handlers.
+        """
+        if not self.elastic_endpoint or not self.error_payload:
+            pytest.skip("skipping the http error test")
+        index = self.build_index(server_config, ("2020-08",))
+        query_api(
+            self.pbench_endpoint,
+            self.elastic_endpoint,
+            self.error_payload,
+            index,
+            HTTPStatus.BAD_GATEWAY,
+            status=errors,
+            headers={"Authorization": "Bearer " + pbench_token},
+        )
--- a/lib/pbench/test/unit/server/query_apis/commons.py
+++ b/lib/pbench/test/unit/server/query_apis/commons.py
@@ -2,8 +2,7 @@ import itertools
 import pytest
 import requests
 
-from dateutil import parser as date_parser
-from dateutil import rrule
+from dateutil import parser as date_parser, rrule
 from http import HTTPStatus
 from typing import AnyStr, Type
 
@@ -153,15 +152,12 @@ class Commons:
         parameter_items = self.cls_obj.schema.parameters.items()
 
         required_keys = [
-            key
-            for key, parameter in self.cls_obj.schema.parameters.items()
-            if parameter.required
+            key for key, parameter in parameter_items if parameter.required
         ]
 
         all_combinations = []
         for r in range(1, len(parameter_items) + 1):
-            combinations_object = list(itertools.combinations(parameter_items, r))
-            for item in combinations_object:
+            for item in itertools.combinations(parameter_items, r):
                 tmp_req_keys = [key for key, parameter in item if parameter.required]
                 if tmp_req_keys != required_keys:
                     all_combinations.append(item)
@@ -187,14 +183,14 @@ class Commons:
         """
         Test behavior when a bad date string is given
         """
-        if all(
-            p.type != ParamType.DATE for _, p in self.cls_obj.schema.parameters.items()
-        ) or any(k not in self.payload for k in ("start", "end")):
+        skip = True
+        for key, p in self.cls_obj.schema.parameters.items():
+            # Modify date/time key in the payload to make it look invalid
+            if p.type == ParamType.DATE and key in self.payload:
+                self.payload[key] = "2020-19"
+                skip = False
+        if skip:
             pytest.skip("skipping " + self.test_bad_dates.__name__)
-
-        # Modify start and end time in the payload to make it look invalid
-        self.payload["start"] = "2020-12"
-        self.payload["end"] = "2020-19"
         response = client.post(
             server_config.rest_uri + self.pbench_endpoint,
             headers={"Authorization": "Bearer " + pbench_token},
--- a/lib/pbench/test/unit/server/query_apis/commons.py
+++ b/lib/pbench/test/unit/server/query_apis/commons.py
@@ -269,9 +269,6 @@ class Commons:
         if not self.elastic_endpoint:
             pytest.skip("skipping " + self.test_http_exception.__name__)
 
-        # Make the start and end time in payload the same to result in an exception
-        self.payload["end"] = self.payload["start"]
-
         index = self.build_index(
             server_config, self.date_range(self.payload["start"], self.payload["end"])
         )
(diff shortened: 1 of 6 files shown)


lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
```diff
@@ -12,69 +13,32 @@ class TestDatasetsDetail:
...
+                "name": "footest",
+                "start": "2020-12",
+                "end": "2020-19",
+            },
+            error_payload={"name": "foobar", "start": "2020-08", "end": "2020-08"},
+            empty_response_payload={
+                "hits": {
+                    "total": {"value": 0, "relation": "eq"},
+                    "max_score": None,
+                    "hits": [],
```
webbnh, line 38: Why is this value different from the one at line 24 in `lib/pbench/test/unit/server/query_apis/test_controllers_list.py`?

It seems like maybe it would be worth defining this value as a class attribute which could be used by code like this to initialize the instance attribute.  (That is, if the empty response payload is always the same, for those tests which specify it, then it would be good to have a common definition that those tests could use to specify it....)

lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
```diff
@@ -341,11 +269,8 @@ def test_nonunique_query(
         """
         Check the handling of a query that returns too much data.
         """
-        json = {
-            "name": "fio",
-            "start": "2020-08",
-            "end": "2020-10",
-        }
+        # We remove the user value so we can make an unauthorized query
+        del self.payload["user"]
```
webbnh, line 273: Is this current instance of this class, upon which this member function is invoked, reinstantiated for each test scenario in it?  (I'm concerned that removing the `"user"` field from the `payload` will have an impact on subsequent invocations of this scenario or others in the test-class...and, I'm wondering if this code should restore the field when it is done....)

webbnh (changes requested):
These changes seem like a very nice improvement.  And, thank you for breaking them out into a separate PR!

I have a few concerns about how PyTest invokes the code, which hopefully you can allay, below.  It would be cool if you could hoist the missing-keys test completely into the superclass.  There are a couple of small items which you should attend to (like a type hint problem, et al.).  And, then there's the usual collection of nits.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -0,0 +1,233 @@
...
+    constructor and `post` service.
+    """
+
+    def _setup(
+        self,
+        pbench_endpoint: AnyStr = None,
+        elastic_endpoint: AnyStr = None,
+        payload: [AnyStr, Any] = None,
+        bad_date_payload: [AnyStr, Any] = None,
+        error_payload: Dict[AnyStr, Any] = None,
```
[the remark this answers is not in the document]
npalaska: Fixed

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -0,0 +1,233 @@
...
+            json=keys,
+        )
+        assert response.status_code == HTTPStatus.BAD_REQUEST
+        missing = [k for k in self.required_keys if k not in keys]
+        assert (
+            response.json.get("message")
+            == f"Missing required parameters: {','.join(missing)}"
+        )
+
+    def test_bad_dates(self, client, server_config, user_ok, pbench_token):
```
dbutenhof, line 116: Instead of having a `bad_date_payload` ... how about checking the class schema for a `ParamType.DATE` parameter, and simply *modify* (a copy of) the standard subclass `self.payload` to create a bad date? Fewer parameters to worry about, more consistency. (This could even easily be made to iterate over *all* `ParamType.DATE` parameters in the schema instead of only testing one.)

dbutenhof (changes requested):
Thanks for splitting this off! It's lots easier to focus on the features of each part this way.

(Bad news, that's probably part of why I have more comments... 😈 )

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
```diff
@@ -91,33 +53,15 @@ def test_missing_keys(
...
-        assert (
-            response.json.get("message")
-            == "Value '2020-19' (str) cannot be parsed as a date/time string"
-        )
+       """
+        self.required_keys = [
+            key
+            for key, parameter in ControllersList(
+                client.config, client.logger
+            ).schema.parameters.items()
```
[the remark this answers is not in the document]
webbnh: An _instance_ of the class?...hmm...yes...could do...but...this is Python, where we can pass the class itself, and use that object to create an instance when we want it!  😀 

I dunno, maybe it would be better to use an instance...I just get so few opportunities to use class objects.  😉

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -0,0 +1,233 @@
...
+        server_config,
+        query_api,
+        user_ok,
+        find_template,
+        build_auth_header,
+    ):
+        """
+        Check proper handling of a query resulting in no Elasticsearch matches.
+        The test will run thrice with different values of the build_auth_header
+        fixture.
```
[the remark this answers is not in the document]
webbnh: The fact that the test relies on the `build_auth_header` fixture is evident from the code; what needs commenting on is the fact that that fixture is likely to cause the test to be executed repeatedly, because that might be surprising to the reader.  Naming the fixture in the comment should be sufficient to prompt the reader to go look at it if they want the details; so, all that should be need here is to point out that it will trigger multiple runs.

lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
webbnh: I thought about suggesting the copy approach, but I thought I'd first try to find out whether we even need it.  (Copying objects in Python is notoriously error-prone -- even `copy()` isn't necessarily sufficient to protect the original, depending on what needs to be done to the copy -- it might need `deepcopy()`, but, of course, even _that_ doesn't necessarily do what one wants.... 😝)

npalaska: Good point but the fixture scope of the `_setup` is "function" (its default value) and so the fixture is destroyed at the end of the each test function. And since the fixture is defined with `autouse=true` it will be available at the beginning of each test as it was defined in `_setup`.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -0,0 +1,233 @@
...
+        return index
+
+    def test_non_accessible_user_data(self, client, server_config, pbench_token):
+        """
+        Test behavior when Authorization header does not have access to other user's data
+        """
+        # The pbench_token fixture logs in as user "drb"
+        # Trying to access the data belong to the user "pp"
+        if not self.payload.get("user", None):
+            pytest.skip("skipping non accessible user data test")
```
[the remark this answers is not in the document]
npalaska: > This at least deserves an explanation in the docstring; you're skipping out if the test doesn't have `"user"` or if it's given as `None`, but then overwriting whatever value was given with `"pp"` below. So we _have_ to supply it, or the call does nothing, but if we _do_ supply it, the value we specified isn't used ... ???

Yes because as the name of the test suggest its testing the non accessible user data, the pbench_token that we use for  this test belongs to the "drb" user who will try to access the user with username "pp" (comment on line 49-50) and that's why I deliberately change the payload user value to "pp"

> I'm assuming this is to deactivate the test case for publish, which doesn't have a `user` parameter and requires `Dataset` setup to test authorization failures; and assumes that all the other subclasses will supply `user` in their default payload parameter to the super-constructor. It'd be cleaner to rely on the subclass schema, which you appear to be accessing in other contexts.

yes that good idea I will try that

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -0,0 +1,233 @@
...
+        exceptions,
+        user_ok,
+        find_template,
+        pbench_token,
+    ):
+        """
+        Check that an exception in calling Elasticsearch is reported correctly.
+        """
+        if not self.elastic_endpoint or not self.error_payload:
+            pytest.skip("skipping the http exception test")
```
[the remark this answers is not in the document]
npalaska: The payload is valid except that the `start` and `end` values are same. http_exception and error test uses a payload where start and end time are same and if they are not we dont get an expected behavior even though query_api call forces the requests mock to throw an exception and return a bad status. I guess it has to do with the fact that it somehow has to raise an exception from the main class and I get Connection error saying that the call didnt match any of the registered mock.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
```diff
@@ -12,64 +13,25 @@ class TestControllersList:
...
-            headers={"Authorization": "Bearer " + pbench_token},
+    @pytest.fixture(autouse=True)
+    def _setup(self):
+        super()._setup(
+            pbench_endpoint="/controllers/list",
+            elastic_endpoint="/_search?ignore_unavailable=true",
+            payload={"user": "drb", "start": "2020-08", "end": "2020-10"},
+            bad_date_payload={"user": "drb", "start": "2020-12", "end": "2020-19"},
+            error_payload={"start": "2020-08", "end": "2020-08"},
+            empty_response_payload={
```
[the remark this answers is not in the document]
npalaska: Okay so I did `empty_es_response_payload` not too long 😁

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -0,0 +1,233 @@
...
+        Test behavior when no JSON payload is given
+        """
+        response = client.post(
+            f"{server_config.rest_uri}{self.pbench_endpoint}",
+            headers={"Authorization": "Bearer " + pbench_token},
+        )
+        assert response.status_code == HTTPStatus.BAD_REQUEST
+        assert response.json.get("message") == "Invalid request payload"
+
+    def missing_keys(self, client, server_config, keys, user_ok, pbench_token):
```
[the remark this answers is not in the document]
webbnh: Oh, boo.

So, I still think it's worth storing the class object as an attribute of the test class (which will DRY out the code more), but, yeah, passing in the `keys` value from the PyTest parameterization is probably the best approach, rather than trying to find some way to hang that off the test class as well.  Pity.

lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
webbnh: I'm less worried about the _fixture_ than I am about the test class instance; but, if the fixture completely reinitializes the instance on each test-case (function) invocation, then I guess we're good.

lib/pbench/test/unit/server/query_apis/commons.py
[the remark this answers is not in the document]
npalaska: Yes indeed, I implemented this in the current commit

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -0,0 +1,272 @@
...
+        find_template,
+        pbench_token,
+    ):
+        """
+        Check that an exception in calling Elasticsearch is reported correctly.
+        """
+        if not self.elastic_endpoint:
+            pytest.skip("skipping the http exception test")
+
+        # Make the start and end time in payload the same to result in an exception
```
dbutenhof, line 235: Huh? We're going to force an exception on the Elasticsearch calls; we just need *valid* dates to get through the parameter parsing before we get to that point. (And even with your mock matching issue, which is reasonable enough although we could always build the expected URL to match the actual dates, the identical dates are only *enabling* not *causing* the exception, by ensuring that the mock matches the actual URL.)

In fact, this still requires coordination between setting up the mock address and construction of the query, so it would be better to do both consistently here, in which case the date values don't matter. (Although a URL with a single index name is easier/faster to construct than 15 separate index names!)

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -0,0 +1,272 @@
...
+        self, server_config, query_api, user_ok, find_template, pbench_token, errors
+    ):
+        """
+        Check that an Elasticsearch error is reported correctly through the
+        response.raise_for_status() and Pbench handlers.
+        """
+        if not self.elastic_endpoint:
+            pytest.skip("skipping the http error test")
+
+        # Make the start and end time in payload the same to result in an error
```
dbutenhof, line 260: Again, the comment is misleading as the dates don't "result in an error" but merely ensure we can get far enough for the requests mock to generate an error in the Elasticsearch call.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
```diff
@@ -12,64 +13,15 @@ class TestControllersList:
...
-            headers={"Authorization": "Bearer " + pbench_token},
-            json={"user": "pp", "start": "2020-08", "end": "2020-10"},
+    @pytest.fixture(autouse=True)
+    def _setup(self, client):
+        super()._setup(
+            cls_obj=ControllersList(client.config, client.logger),
+            pbench_endpoint="/controllers/list",
+            elastic_endpoint="/_search?ignore_unavailable=true",
+            payload={"user": "drb", "start": "2020-08", "end": "2020-10"},
+            empty_es_response_payload=Commons.empty_es_response_payload,
```
dbutenhof, line 23: Instead of having the subclasses know the superclass name and specify it here, you could use this as the default of a superclass keyword argument which the subclasses could simply omit.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
```diff
@@ -91,33 +43,8 @@ def test_missing_keys(
...
-            f"{server_config.rest_uri}/controllers/list",
-            headers={"Authorization": "Bearer " + pbench_token},
-            json={"user": "drb", "start": "2020-12", "end": "2020-19"},
-        )
-        assert response.status_code == HTTPStatus.BAD_REQUEST
-        assert (
-            response.json.get("message")
-            == "Value '2020-19' (str) cannot be parsed as a date/time string"
-        )
+       """
```
dbutenhof, line 46: It looks like the indentation of the closing triplequote has slipped?

lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
```diff
@@ -97,38 +49,8 @@ def test_missing_keys(self, client, server_config, keys, user_ok, pbench_token):
...
-                "start": "2020-12",
-                "end": "2020-19",
-            },
-        )
-        assert response.status_code == HTTPStatus.BAD_REQUEST
-        assert (
-            response.json.get("message")
-            == "Value '2020-19' (str) cannot be parsed as a date/time string"
-        )
+       """
```
dbutenhof, line 52: Here, too.

lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
dbutenhof: Ah; it seems inefficient and generally unnecessary to build a new instance for every test method, but it does eliminate the risk of cross-contamination and it's not like our tests are taking "hours" to complete. So, carry on ... 😆

dbutenhof (changes requested):
Almost there!

lib/pbench/test/unit/server/query_apis/commons.py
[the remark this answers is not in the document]
webbnh: > I guess it has to do with the fact that it somehow has to raise an exception from the main class and I get Connection error saying that the call didnt match any of the registered mock.

This sounds like something that we should understand completely before we merge this PR.

Looking at [query_api](https://github.com/distributed-system-analysis/pbench/blob/189bec13a4330c4caf582a2fd20c773172b8a27b/lib/pbench/test/unit/server/query_apis/conftest.py#L42), we're not going to register a mock if `expected_status` is `FORBIDDEN` or `UNAUTHORIZED`, so some of the test parameterizations are going to set up mocks and some of them are not...but, I don't think the difference has anything to do with the `start` and `end` values.  (And, I don't think the CUT actually looks at the `start` and `end` values -- it just passes them through to the request, which is mocked!)

lib/pbench/test/unit/server/query_apis/commons.py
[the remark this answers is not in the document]
webbnh: That would, indeed, be most cool!

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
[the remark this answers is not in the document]
webbnh: The code excerpt above instantiates an instance of `ControllersList` and then invokes a method on one of its attributes; instead, it could do the instantiation here and let the invocation be done elsewhere..._or_...it could pass the class object so that the code elsewhere could do _both_ the instantiation and the invocation.  (It's pretty close to six of one and a half dozen of the other, I think....)

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
webbnh: Or, wouldn't `self.empty_es_response_payload` work?

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -0,0 +1,272 @@
...
+        """
+        parameter_types = [
+            parameter.type
+            for _, parameter in self.cls_obj.schema.parameters.items()
+            if parameter.required
+        ]
+        if ParamType.DATE not in parameter_types or not all(
+            k in self.payload for k in ("start", "end")
+        ):
+            pytest.skip("skipping the bad date test")
```
webbnh, line 150: You can make this marginally more efficient using `any` instead of `all`:
```
        if ParamType.DATE not in parameter_types or any(
            k not in self.payload for k in ("start", "end")
        ):
            pytest.skip("skipping the bad date test")
```

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
webbnh: Hmmm, I believe that it _would_ work, but it will look extremely funny.

Regardless of how you address it, I recommend that you don't name the instance attribute the same as the class attribute which contains the default value.  While that is sort of clever, I suspect that it will just end up being confusing.  That is, as long as the instance attribute is undefined or unset, any reference to the name will resolve to the class attribute; once the instance attribute has been set, references will resolve to that.  Thus,
```
            empty_es_response_payload=self.empty_es_response_payload,
```
would do exactly what we want it do (copy the class attribute value into the instance attribute), but it will look like a pointless no-op (and, in fact, if we could avoid the assignment altogether, _that_ would do exactly what we want (in this case) it to do, as well...).

So, it would be better if the class attribute were named something like `default_empty_es_response_payload` or `EMPTY_ES_RESPONSE_PAYLOAD` (since it's a "constant").

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
```diff
@@ -91,33 +43,8 @@ def test_missing_keys(
         Note that "start", and "end" are required whereas "user" is not mandatory;
```
webbnh, line 43: There's a spurious comma here.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
dbutenhof: In principle we could just supply the JSON payload as a literal directly to the superclass `_setup` kwarg, just like `payload=` on the line above. Granted, the Elasticsearch verbiage is generally a bit more flowery, and defining it as a subclass attribute would be fine where that gets ugly... which is definitely an uppercased constant. 👍🏻

lib/pbench/test/unit/server/query_apis/test_datasets_list.py
```diff
@@ -59,13 +49,6 @@ def test_missing_keys(self, client, server_config, user_ok, pbench_token, keys):
         however, Pbench will silently ignore any additional keys that are
         specified.
        """
```
webbnh, line 51: The indentation on this line has slipped, too.

webbnh (commented):
Looking good.  Other than the rename of `Commons.empty_es_response_payload`, these are all just nits.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
webbnh: Sounds like consensus.

But, I want to clarify one thing:

> defining it as a subclass attribute would be fine

Since the value is the same across several subclasses, we're defining it as a _super_-class attribute.  (Is that what you meant?)

dbutenhof: Two separate values, if we're following my earlier suggestion of a superclass "constructor" kwarg:

1. The superclass default, when the kwarg is omitted;
2. The individual (non-default) subclass values which are specified in their superclass setup calls

Either may be an inline literal or, if that's too big, a constant in the local class (or even a local within the setup method since everyone else accesses it only through `self.empty_es_response_payload`).

webbnh: Ah, I see, now.  Yes, I was wondering how we might do something like that -- allowing instances to inherit the class attribute value as their instance attribute value while preserving the ability of subclasses to set it to something else -- and it didn't occur to me to use keyword parameters...so that's...um...quite clever.  But, how would you code that in `_setup()`?  Would you have it do key lookups (e.g., `empty_es_response_payload`) in `kwargs`, and then assign the instance attribute only if the key is present?

npalaska:
@webbnh and @dbutenhof 
Alright I fixed the `test_missing_keys` function so now its entirely part of the commons functionality, it uses schema to determine the required parameters and create keys on the fly. 

I also fixed the `test_http_error` and `test_http_exception` so they dont rely on hardcoded values for creating index, I wrote a simple helper function `date_range (start, end)` that builds list of range of dates between start and end (that we specify in our main payload) and so its consistent across all the test cases now.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
npalaska: fixed

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
npalaska: Fixed

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -59,14 +60,26 @@ def build_index(self, server_config, dates):
...
+        """
+        Builds list of range of dates between start and end
+        It expects the date to look like YYYY-MM
+        """
+        y, m_start = int(start[:4]), int(start[5:])
+        y, m_end = int(end[:4]), int(end[5:])
+        date_range = []
+        for month in range(m_start, m_end + 1):
+            date_range.append(f"{y}-{month:02}")
+        return date_range
```
dbutenhof, line 73: Um, wow. How about using `datetime` or `dateutil` to actually parse and then pick out the year and month or use the datetime f-string formatting? This code is really hard to read, and without going searching for calls to figure out the input strings I have no idea whether it's correct.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -111,43 +126,72 @@ def test_missing_json_object(self, client, server_config, pbench_token):
...
+                    keys[item[0]] = "public"
+                elif item[1].type == ParamType.DATE:
+                    keys[item[0]] = "2020"
+                else:
+                    keys[item[0]] = "foobar"
+
+            missing_key_helper(keys)
+
+        # Test for random non-existent key
+        missing_key_helper({"notakey": None})
```
dbutenhof, line 182: This seems misleading, since it's not the key name used here that's important but the fact that all of the required keys are missing because the arbitrary name you used is not one of them.

(And, technically, this test would be invalid if we had a query API with *no* required parameters, though I'm not sure that's likely.)

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -111,43 +126,72 @@ def test_missing_json_object(self, client, server_config, pbench_token):
...
-        if ParamType.DATE not in parameter_types or not all(
-            k in self.payload for k in ("start", "end")
+        if ParamType.DATE not in parameter_types or any(
+            k not in self.payload for k in ("start", "end")
         ):
-            pytest.skip("skipping the bad date test")
+            pytest.skip("skipping " + self.test_bad_dates.__name__)
 
         # Modify start and end time in the payload to make it look invalid
         self.payload["start"] = "2020-12"
```
dbutenhof, line 197: OK, well, I had in mind something more general; but I get that the silly Elasticsearch query URI kinda relies on start and end specifically. However, in that case, looking for `DATE` parameters doesn't really serve much purpose, unless you want to verify that the `start` and `end` parameters are both in the schema *and* have `DATE` type, which would be a bit safer although nobody's likely to use those names for anything else.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
```diff
@@ -20,32 +20,9 @@ def _setup(self, client):
             pbench_endpoint="/controllers/list",
             elastic_endpoint="/_search?ignore_unavailable=true",
             payload={"user": "drb", "start": "2020-08", "end": "2020-10"},
-            empty_es_response_payload=Commons.empty_es_response_payload,
+            empty_es_response_payload=Commons.EMPTY_ES_RESPONSE_PAYLOAD,
```
dbutenhof, line 23: I'd still prefer to make this an omittable kwarg on the subclass "constructor" rather than having the subclasses know the superclass attribute name.

lib/pbench/test/unit/server/query_apis/commons.py
webbnh: Also...this code can produce illegal date ranges...is that a "feature"??

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
webbnh: I'm ambivalent about the omittable kwarg approach -- it's pretty subtle.

An alternative would be to use `self` here instead of `Commons`:  if the subclass doesn't define `EMPTY_ES_RESPONSE_PAYLOAD` as a class attribute, then it will inherit the superclass definition (without referencing it by name) which is all the right things, but the subclass can still override the definition or provide another value.

webbnh (commented):
Still looks good, but there are still more nits.  🙂

lib/pbench/test/unit/server/query_apis/commons.py
npalaska: Added an assertion to check that the start_date is always less than the end_date

lib/pbench/test/unit/server/query_apis/commons.py
npalaska: Fixed, I added a condition that if required_keys are present then only we run this test.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -159,38 +162,34 @@ def missing_key_helper(keys):
         for r in range(1, len(parameter_items) + 1):
             combinations_object = list(itertools.combinations(parameter_items, r))
             for item in combinations_object:
```
webbnh: If these two lines are the only place where you reference `combinations_object`, then it would be better if you didn't instantiate the list -- that is, if you're only going to iterate over the combinations once, then iterating over the iterator returned by `combinations()` would be more efficient than collecting all the iterations into a list and then iterating over the list.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -0,0 +1,318 @@
...
+                == f"Missing required parameters: {','.join(missing)}"
+            )
+
+        parameter_items = self.cls_obj.schema.parameters.items()
+
+        required_keys = [
+            key
+            for key, parameter in self.cls_obj.schema.parameters.items()
+            if parameter.required
+        ]
```
webbnh, line 159: It would be better in several ways to reference `parameter_items` instead of `self.cls_obj.schema.parameters.items()` at line 157.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -159,38 +162,34 @@ def missing_key_helper(keys):
         for r in range(1, len(parameter_items) + 1):
```
webbnh, line 162: You should probably start the range a zero, to test the empty case:
```
        for r in range(len(parameter_items) + 1):
```

webbnh (approved):
OK, I think you've resolved all of my substantial concerns (I'll know better once Dave makes a pass over the unresolved conversations 🙂).

Below are three more nits that would be nice to have addressed, but don't feel like you have to unless you have some other reason to update the code.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -111,43 +126,72 @@ def test_missing_json_object(self, client, server_config, pbench_token):
...
+        all_combinations = []
+        for r in range(1, len(parameter_items) + 1):
+            combinations_object = list(itertools.combinations(parameter_items, r))
+            for item in combinations_object:
+                tmp_req_keys = []
+                for element in item:
+                    if element[1].required:
+                        tmp_req_keys.append(element[0])
+                if not sorted(tmp_req_keys) == sorted(required_keys):
+                    all_combinations.append(item)
```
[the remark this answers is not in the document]
webbnh: Clever of you to note that you didn't need to sort _either_ list!

dbutenhof (commented):
I'm still not entirely happy with the limitations and assumptions in the bad-date test case; it "works" for now, but it's much less robust and general than it should be, and might bite us in the future. I also don't like the misleading comments in the error and exception test cases which erroneously suggest that the dates have something to do with the error generation.

lib/pbench/test/unit/server/query_apis/commons.py
npalaska: Empty payload test is different and it behaves differently, the response would be `Invalid request payload` which already has its own test function `test_missing_json_object`, for the missing keys test we need at least one key in the payload and the response would be `Missing required parameters: <parameters>`

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
webbnh: I no longer remember clearly what I thought the wrinkles were....  However, it seemed like the approach included a superclass attribute for a default and a subclass attribute for a default, and the initialization got one behavior if it omitted the kwarg and a different behavior if it specified is as `None`, while preserving the option of specifying a non-default value either using a predefined value (e.g., one of the attributes) or a literal...and, it struck me at the time as though the obvious/straightforward code did not necessarily produce the naively-expected result.

I think I was thinking that if the kwarg were omitted, then the expected result would be `None`, which led to the idea that the subclass code would have to specify the superclass attribute to get what we were thinking would be "the default" value...and, that's not really different from what we ended up with without using a kwarg.

But, anyway, as you say, let's move on.

lib/pbench/test/unit/server/query_apis/commons.py
webbnh: I think it's different:  I think what I'm describing will send a payload which is an object which contains no fields; I think what you are describing either sends no payload or sends a payload which contains no object (I'm not sure exactly what the difference between those might be).  So, I think what I'm describing is still a valid test which is different from what you're already testing.

lib/pbench/test/unit/server/query_apis/commons.py
webbnh: OK, I'll second this request.  My motivation is that, if a request has either `"start"` **_or_** `"end"`, it should probably be considered legitimate.  That is, while we presumably don't do so currently, we _could_ support open-ended ranges.  So, having the test skip when we specify only one end of the range is probably overly restrictive.

From a practical perspective, I expect that Nikhil wants to skip the test if either `"start"` or `"end"` is missing because his code is going to overwrite those two fields.  But, it's easy enough to make the code skip any fields which don't exist...or, as Dave suggests, to have it iterate over the date fields which do exist, stomping on each one in sequence, repeating the request each time, which would be good and thorough (and general).

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -187,14 +183,14 @@ def test_bad_dates(self, client, server_config, user_ok, pbench_token):
...
-            p.type != ParamType.DATE for _, p in self.cls_obj.schema.parameters.items()
-        ) or any(k not in self.payload for k in ("start", "end")):
+        skip = True
+        for key, p in self.cls_obj.schema.parameters.items():
+            # Modify date/time key in the payload to make it look invalid
+            if p.type == ParamType.DATE and key in self.payload:
+                self.payload[key] = "2020-19"
+                skip = False
+        if skip:
             pytest.skip("skipping " + self.test_bad_dates.__name__)
```
webbnh: You shouldn't need the `skip` flag -- I believe the following will accomplish the same thing:
```
        for key, p in parameter_items:
            # Modify date/time key in the payload to make it look invalid
            if p.type == ParamType.DATE and key in self.payload:
                self.payload[key] = "2020-19"
                break
        else:
            pytest.skip("skipping " + self.test_bad_dates.__name__)
```
So, once a `DATE` field is found, it'll break out of the loop; only if it fails to break out does it execute the `else` clause.  (Note that I also replaced your use of `self.cls_obj.schema.parameters.items()` with the local `parameter_items` variable.)

However, I don't think that this will accomplish what Dave requested:  this is going to replace only the first `DATE` field that it encounters with a bad date, and then it's going to make a request with the resulting payload, and then the test will end -- so it's really only going to test _one_ of the date fields (i.e., if there is more than one).

npalaska: Technically, yes that should work since our code will abort with `BAD_REQUEST` as soon as it encounters first bad date/time field so that means we only need one bad date/time field in the payload for the test to succeed.

lib/pbench/test/unit/server/query_apis/commons.py
npalaska: you mean sending json object as `{}`? 
```
response = client.post(
                server_config.rest_uri + self.pbench_endpoint,
                headers={"Authorization": "Bearer " + pbench_token},
                json={},
            )
```
If this is what you mean then its as good as sending no payload

webbnh: "As good as" perhaps, but nevertheless _different_.  That is, a payload of `{}` has a length of `2` and a type of `application/json`, whereas "no payload" has a length of zero, and I suspect that the type is never used.  I would have to dive deep into the `requests` or Flask documentation to explore the difference, but it would probably be pretty straightforward for you to just crank up the server and try it.

It's possible that our code doesn't discriminate between the two cases.  And, it's possible that the layers above/below our code paper over the difference for us.  My knowledge isn't that detailed.  But, given that we hope to support multiple clients, it would be good to ensure that we handle both cases gracefully.

lib/pbench/test/unit/server/query_apis/commons.py
webbnh: That depends greatly on what you mean by "work" and "succeed"!  😁 

If our Server code handles bad data in `"start"` gracefully but crashes on bad data in `"end"`, then we won't know about that if our unit test only tries bad data in the `"start"` field.

Part of the idea behind unit tests is to try to get execution to follow all the paths through the CUT.  It's tricky to do this without making the unit tests know too much about the CUT, but, at a minimum, it's reasonable to test all the parameters to our interface in as many combinations as we can easily generate.

Thus, testing only the first one is not very satisfying, especially if we can put it in a loop which tests each of them.

npalaska: > If our Server code handles bad data in `"start"` gracefully but crashes on bad data in `"end"`

I highly doubt this will happen, because all the date/time fields we parse via parser from dateutil, it wont handle one field gracefully and crash on another same type of field. 

But I guess if we want to make sure that all the date/time field gets the bad date then this code wont achieve that. Alternately, I can move the code that makes the post request (with bad date payload) inside this for loop and we can test for each date/time field present in the payload for the bad date, if that make sense. @webbnh @dbutenhof

lib/pbench/test/unit/server/query_apis/commons.py
npalaska: Yes I experimented with it and we validate the json data in our post and get request (schema validation) before proceeding where it treats `{}` and `None` as same.

If you look in the class `Schema` and `validate` function line 363 in `lib/pbench/server/api/resources/query_apis/__init__.py` you will see we raise `InvalidRequestPayload()` if no json data present
```
if not json_data:
    raise InvalidRequestPayload()
```

dbutenhof: Yes, we do fail out with an empty payload. I'm not sure I can think of a "query api" that would have no JSON input, but then you've already got the index mappings where the parameter is in the URI rather than as JSON. If you'd based it on `ElasticBase`, you'd be tripping over this code.

Arguably, the `Schema` validator should fail with an empty or missing payload only if the schema has one or more required parameters. It's too bad we can't iterate over `None`, because we'd be fine if we just let it complain about missing required parameters later.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -0,0 +1,314 @@
...
+        response = client.post(
+            server_config.rest_uri + self.pbench_endpoint,
+            headers={"Authorization": "Bearer " + pbench_token},
+            json=self.payload,
+        )
+        assert response.status_code == HTTPStatus.BAD_REQUEST
+        assert (
+            response.json.get("message")
+            == "Value '2020-19' (str) cannot be parsed as a date/time string"
+        )
```
dbutenhof, line 203: It'd be *so* easy now to move the call and asserts inside the loop and continue through the entire schema.

Then we could just succeed on completion; not really any need to "skip" if the API has no `DATE` parameters to test, although it might be informational.

lib/pbench/test/unit/server/query_apis/commons.py
dbutenhof: I really don't like these unclear comments, and I wish you'd fixed them. I'm going to approve anyway with slight reluctance because I'd like to move past this. But if you make more changes for other reasons, please take a look at this again.

dbutenhof (approved):

lib/pbench/test/unit/server/query_apis/commons.py
npalaska: Fixed. Yeah, this code should't be here.

dbutenhof (dismissed):

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -183,24 +183,20 @@ def test_bad_dates(self, client, server_config, user_ok, pbench_token):
...
+                response = client.post(
+                    server_config.rest_uri + self.pbench_endpoint,
+                    headers={"Authorization": "Bearer " + pbench_token},
+                    json=self.payload,
+                )
+                assert response.status_code == HTTPStatus.BAD_REQUEST
+                assert (
+                    response.json.get("message")
+                    == "Value '2020-19' (str) cannot be parsed as a date/time string"
+                )
```
webbnh: If I understand this code, it's going to look for a `DATE` field, and, if it finds one, it's going to issue the request; if that behaves properly, then it's going to continue looking for another `DATE` field.  While it's looking, it won't issue any requests until it finds one (which is fine); when it finds one, it's going to damage the date value (which is fine -- that's how it drives the test); however, at the point when it's going to issue the second request, it's still got the damaged value from the first request in the payload!  So, the second request is unlikely to test anything different from the first one.  😿 

In general, I think I like this approach.  But, I think it needs to save and restore the original value after the assertions succeed and before it moves to the next iteration.

webbnh (changes requested):
So close!

lib/pbench/test/unit/server/query_apis/commons.py
webbnh: I've lost my grip, and I think the time for this discussion has passed, so I'm just going to resolve it.

lib/pbench/test/unit/server/query_apis/commons.py
npalaska: I understand what you mean, but the api code is gonna behave exactly same for any field that is of `DATE` type (because it passes the validation via one function, `conver_date`) so in practice really if this test works for one `Date` field we dont need to test on any other `DATE` field, it would anyway be redundant. However, I guess it wouldn't hurt to test bad date on each `Date` field separately since we are already inside a for loop

webbnh (approved):
GTG

lib/pbench/test/unit/server/query_apis/commons.py
webbnh: > the api code is gonna behave exactly same for any field that is of `DATE` type (because it passes the validation via one function, `conver_date`)

Perhaps that's true today, but the unit test code doesn't (and shouldn't!) know that, because it might not be true tomorrow.  The unit tests should verify the "interface contract", which (IMO) says that, if any `DATE` field is invalid, the API will respond in a particular way...and that means it should test them all, not just the first one (especially since it was easy to make it do so 😉).

dbutenhof (approved):

Issue #2422, opened 2021-08-19
Bumps in the road to installing a Server in a green field


webbnh:
Today, Dave, Nikhil, and I [fought our way](https://drive.google.com/file/d/1oPlvOGFG_vvGZQ1Yt3thHI9Kt7dcic1C/view) from a brand new VM to a successfully installed Pbench Server (to which we uploaded results which we were able to view in a Dashboard instance).  It was not a smooth process.  And, at the moment, it is not mechanically/automatically reproducible, although it is close.  I'm using this issue to record the bumps we hit along that road.  At the bottom is a summary of work items (although, if you read the list of bumps, you may find additional ones).  Dave has resolved some of the problems in #2423; some problems may require a more significant effort to smooth out.

NOTE: This is useful to have recorded in one place, but it ought to be a Jira story (@pravins) with a series of more focused independent *tasks* as we define them which can be reflected back into GitHub during planning.

## The bumps in the Server installation road

- [ ] Use the correct Ansible playbook....
- [ ] Ensure that the system has the correct RPM PGP key for the EPEL (and other) repos that we pull from. (Can this be automated in Ansible?)
- [ ] Enable the COPR repo selected for the Pbench install. (Ansible roles use `pbench_repo_url_prefix`, but can the `dnf`/`yum` "enable" operation be automated to enable "green field" installs or is some manual pre-install setup inevitable?)
- [ ] `firewalld` needs to be installed _and_ started; or role needs to be smart enough to skip firewall configuration if not installed.
- [x] We initially couldn't start the server because Click was not installed.  It was not listed as a direct dependency of the Server.  (We discovered later that it is normally installed as a dependency of Flask, but, because of another issue, the installations were not completing, so it wasn't being installed.)   We added it to the `requirements.txt` for the Server.
- [x] While investigating the missing Click installation, we found an error in the RPM installation "post" step indicating that `psycopg2-binary` wouldn't install via `pip3`.  This shouldn't have been an actual problem because `psycopg2` is already installed via RPM dependency.  We removed `psycopg2-binary` from the `requirements.txt`.  The weird thing is that, apparently, the ansible script proceeded despite the fact that the `pip3` installs aborted in the middle!
[...]

PR #2423, opened 2021-08-20
Server installation cleanup

dbutenhof:
1. Add missing dependencies to requirements.txt
2. Lock elasticsearch module dependency to workaround bug
3. Limit module versions to eliminate pip version matrix thrashing
4. Remove the /opt/pbench-server prefix on pip install in order to simplify PYTHONPATH (after discussion with Pete this looked like a better alternative than computing an inclusive path later).

Tested by re-installing on our new staging server.

This is a fix for some problems called out in #2422 but not sufficient to close the issue.

### DIFF
--- a/server/ansible/roles/pbench-server-activate-create-crontab/tasks/main.yml
+++ b/server/ansible/roles/pbench-server-activate-create-crontab/tasks/main.yml
@@ -1,5 +1,7 @@
 ---
 - name: create crontab
+  become: yes
+  become_user: pbench
   command: "pbench-server-activate-create-crontab {{ pbench_server_install_dir }}/lib/crontab"
   environment:
      _PBENCH_SERVER_CONFIG: "{{ pbench_server_install_dir }}/lib/config/pbench-server.cfg"
--- a/server/requirements.txt
+++ b/server/requirements.txt
@@ -1,6 +1,9 @@
+alembic>=1.6.4
+aniso8601>=9.0.1
 boto3
+click
 python-dateutil
-elasticsearch
+elasticsearch==7.13.1
 email-validator
 flask
 flask-bcrypt
@@ -14,7 +17,7 @@ gunicorn
 humanize
 pyesbulk>=2.0.1
 PyJwt
-psycopg2-binary
+psycopg2
 requests
-sqlalchemy
-sqlalchemy_utils
+sqlalchemy>=1.4.23
+sqlalchemy_utils>=0.37.6
--- a/server/rpm/pbench-server.spec.j2
+++ b/server/rpm/pbench-server.spec.j2
@@ -65,7 +65,9 @@ mv %{buildroot}/%{installdir}/%{static}/package.json %{buildroot}/%{installdir}
 # Install python dependencies for pbench user
 (su pbench -c "pip3 --no-cache-dir install --user -r /%{installdir}/requirements.txt") > /%{installdir}/pip3-install.log
 
-# Add install directory into the pbench user default PYTHONPATH
+# Create a `pbench.pth` file under the `pbench` account site library: the
+# `site` package will automatically add directories referenced in `*.pth`
+# files to the default `sys.path` when Python starts.
 su pbench -c "echo /%{installdir}/lib > \$(python3 -m site --user-site)/pbench.pth"
 
 # install node.js modules under /%{installdir}
--- a/server/rpm/pbench-server.spec.j2
+++ b/server/rpm/pbench-server.spec.j2
@@ -62,12 +62,28 @@ cp -a ./web-server/* %{buildroot}/%{installdir}/%{static}
 mv %{buildroot}/%{installdir}/%{static}/package.json %{buildroot}/%{installdir}
 
 %post
-# Install python dependencies for pbench user
+# Install python dependencies as pbench user into user's site-packages
 (su pbench -c "pip3 --no-cache-dir install --user -r /%{installdir}/requirements.txt") > /%{installdir}/pip3-install.log
 
-# Create a `pbench.pth` file under the `pbench` account site library: the
-# `site` package will automatically add directories referenced in `*.pth`
-# files to the default `sys.path` when Python starts.
+# The `site` package is Python magic; it runs automatically when Python starts,
+# and builds `sys.path`.
+#
+# For our purposes, this includes looking in the per-user site library and
+# adding the site-packages paths plus adding directories captured in *.pth
+# files it finds in the "user site" directory.
+#
+# The following line captures our install path `/${installdir}/lib` in our own
+# pbench.pth file for `site` to find when the pbench user runs Python.
+#
+# That `pbench.pth` file needs to be in the right place, and the command
+# 'python3 -m site --user-site` gives us the proper location. The `su pbench`
+# ensures that this is run as "pbench" and therefore gives us the proper
+# path where `pbench.pth` will be found, without making arbitrary assumptions
+# about `site` package behavior.
+#
+# Now, when the pbench user runs Python, without any PYTHONPATH environment
+# variable, the normal Python startup will include both the pbench user's
+# site-library paths, and /opt/pbench-server/lib.
 su pbench -c "echo /%{installdir}/lib > \$(python3 -m site --user-site)/pbench.pth"
 
 # install node.js modules under /%{installdir}
--- a/server/rpm/pbench-server.spec.j2
+++ b/server/rpm/pbench-server.spec.j2
@@ -45,6 +45,10 @@ The pbench server scripts.
 
 %build
 
+%pre
+getent group pbench > /dev/null || groupadd pbench
+getent passwd pbench > /dev/null || useradd -g pbench -d /home/pbench -c "Pbench user" pbench
+
 %install
 rm -rf %{buildroot}
 
@@ -58,8 +62,11 @@ cp -a ./web-server/* %{buildroot}/%{installdir}/%{static}
 mv %{buildroot}/%{installdir}/%{static}/package.json %{buildroot}/%{installdir}
 
 %post
-# Install python dependencies
-pip3 --no-cache-dir install -r /%{installdir}/requirements.txt > /%{installdir}/pip3-install.log
+# Install python dependencies for pbench user
+(su pbench -c "pip3 --no-cache-dir install --user -r /%{installdir}/requirements.txt") > /%{installdir}/pip3-install.log
+
+# Add install directory into the pbench user default PYTHONPATH
+su pbench -c "echo /%{installdir}/lib > \$(python3 -m site --user-site)/pbench.pth"
 
 # install node.js modules under /%{installdir}
 cd /%{installdir}
--- a/server/rpm/pbench-server.spec.j2
+++ b/server/rpm/pbench-server.spec.j2
@@ -59,7 +59,7 @@ mv %{buildroot}/%{installdir}/%{static}/package.json %{buildroot}/%{installdir}
 
 %post
 # Install python dependencies
-pip3 --no-cache-dir install --prefix=/%{installdir} -r /%{installdir}/requirements.txt > /%{installdir}/pip3-install.log
+pip3 --no-cache-dir install -r /%{installdir}/requirements.txt > /%{installdir}/pip3-install.log
 
 # install node.js modules under /%{installdir}
 cd /%{installdir}


server/requirements.txt
```diff
@@ -14,7 +17,7 @@ gunicorn
 humanize
 pyesbulk>=2.0.1
 PyJwt
-psycopg2-binary
+psycopg2
```
webbnh, line 20: Unfortunately, this change appears to have broken the Jenkins-based testing.

server/requirements.txt
```diff
@@ -1,6 +1,9 @@
+alembic>=1.6.4
+aniso8601>=9.0.1
 boto3
+click
 python-dateutil
-elasticsearch
+elasticsearch==7.13.1
```
webbnh, line 6: While this is expedient and _probably_ what we want, pinning us to a specific version seems overly restrictive, which might bite us later.

I recommend `=>7.13.0` and `<7.14.0`, which I think better expresses our requirements.

server/rpm/pbench-server.spec.j2
```diff
@@ -59,7 +59,7 @@ mv %{buildroot}/%{installdir}/%{static}/package.json %{buildroot}/%{installdir}
 
 %post
 # Install python dependencies
-pip3 --no-cache-dir install --prefix=/%{installdir} -r /%{installdir}/requirements.txt > /%{installdir}/pip3-install.log
+pip3 --no-cache-dir install -r /%{installdir}/requirements.txt > /%{installdir}/pip3-install.log
```
webbnh, line 62: I wonder how this impacts containers (vis a vis Mr. Jenkins).

webbnh (changes requested):
It appears that this change breaks the Jenkins-based testing.

(What's weird is that the breakage is not catastrophic, like it should be, it's merely completely destructive....)

server/requirements.txt
dbutenhof: The idea was simply that we knew 7.13.1 worked and didn't want to iterate through any intervening versions to determine where the package problem was introduced.

I think that in terms of general configuration management we'd be better off pinning everything to a specific version we knew meets our requirements. The downside is that we need periodic tasks to evaluate newer versions to determine when to upgrade. In general we've avoided that by letting them float except where we encounter problems. Not the best strategy in the long run, but expedient.

server/requirements.txt
dbutenhof: Yeah; because in a real install it's satisfied by RPM and we don't don't try to do this at all. The original `psycopg2-binary` apparently wasn't satisfied by the RPM; but in Jenkins where we apparently don't have that RPM, we go ahead and try to build, encountering the same problem we tried to avoid by changing this dependency. Ugh.

I'm not sure how we work around this, now, short of changing the `pbench-dev` container to supersede this dependency as we do in a real install.

server/rpm/pbench-server.spec.j2
dbutenhof: `tox` does the install with a prefix anyway, in the temporary tox venv, so that at least should have no real impact. The problems all seem to stem from our friend the psycho cop.

server/requirements.txt
webbnh: Perhaps its needless paranoia, but I'm imagining `7.13.1` being superseded by `7.13.2` and becoming unavailable, leaving our installation broken.  So, my solution to that is to lock out the version that we know doesn't work, while still affording some flexibility.

(Hmmm, I wonder if `!=7.14.0` is an available and workable option....)

server/requirements.txt
webbnh: (For anyone else following along at home, I'm working on building a new container for Mr. Jenkins which installs `psycopg2` from an RPM.)

webbnh: This will be fixed by #2428.

(I took the liberty of building that change, pushing it up to Mr. Jenkins, and having him retest this PR...and now it looks good.)

server/rpm/pbench-server.spec.j2
portante: Until we change the way the server installs code, we have to install using `--prefix`.  This causes `pip3` to install Python3 modules in the RPM installed library paths, which breaks maintenance of the server.

For some reason we dropped the setting of the `PYTHONPATH` in the `pbench-trampoline` code.

Like the agent side, we need to set the `PYTHONPATH` to include the `/opt/pbench-server` installed Python3 modules.

portante (changes requested):

riya-17 (approved):

server/rpm/pbench-server.spec.j2
webbnh: [I've un-resolved this conversation, in light of Peter's reply to it -- @portante, I think it's now on you to mark it resolved....]

Could we finesse this situation by using `--user` instead of `--prefix`?  That would prevent conflicts with the RPM-installed packages, and I believe that the user directory tree is implicitly added to the `PYTHONPATH`.  All we would have to do is to ensure that the installation is done by/for the user which is going to be running the tools.  (If necessary, perhaps we could use a `--prefix` which specifies the directory which `--user` uses.)  That might at least _reduce_ the problem-space....

dbutenhof: It's an interesting possibility; I believe installation is done under `root`, however, whereas we want to target `pbench` where the service and cron jobs will run. I'm not sure I feel entirely comfortable trying to emulate `pip --user`, although I suppose it's well documented and stable. I've never experimented with `--user` ... does `sys.path` really get implicitly adjusted?

(Well, yes, apparently, by implicit import of the `site` package, which can also be used to show the default `sys.path` a bit more straightforwardly than our experiments during the jam: `python3 -m site` will show all the standard places, including reporting that my per-user `~/.local` doesn't exist. So, indeed, perhaps `--prefix ~pbench/.local` would actually do the job?? That would implicitly take care of the cron and service files, and the "enable" script which runs as `pbench`... but maybe not other install/configure tasks from Ansible that run under root.)

Hmm.

webbnh: > I'm not sure I feel entirely comfortable trying to emulate `pip --user`

Yup, but when selecting the smaller insect (i.e., the lesser of weevils)....

Of course, perhaps we don't need to emulate it -- perhaps `su pbench pip --user` would do what we need...except for the "other install/configure tasks from Ansible`, but perhaps those could be run under `su` as well...or, maybe there is some way that Ansible can become the user that we need for certain tasks....

dbutenhof: I think I've got this working, although it's actually `su pbench -c "pip3 ..."` and even that has problems because the installation is done under `root`, and we're logging the pip results into `/opt/pbench-server/`, which we can't write as `pbench`. The `%post` actually fixes this by changing the whole tree to be owned by `pbench` at the end; however doing that earlier causes files owned by root to be laid down meaning we'd need to fix the permissions initially to write the log file and then later go ahead and sweep the directory tree. Instead, I tried running the `su pbench -c "..."` as a sub-shell with the root parent redirecting `stdout` ... and I'm not sure this is any klunkier than the alternatives although I'm not thrilled, either.

In any case, both the cron jobs and service, running as `pbench` implicitly pick up the user libraries, the `pbench-server-enable.yml` playbook already runs as `pbench`. I'm still doing some tests on the rest of the playbook (including the crontab generation, which relies on running a local Pbench server script), but I'm hopeful I'll have a new commit up today.

This all makes me wonder if the `%postun` ought to be removing the `pbench` user; but I think that's actually created by Ansible rather than by the rpm...

webbnh: Hey!  Look what I found:  there's an environment variable [`PYTHONUSERBASE`](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONUSERBASE)...which indicates where `--user` goes to...which means that you can probably skip the `su pbench`, set `PYTHONUSERBASE`, use `pip install --user`, and then let the `%post` fix everything up.  And, if you need to run any Python scripts as `root`, you can use `PYTHONUSERBASE` (instead of `PYTHONPATH`) to find our stuff.

Weird, wild, and wonderful....

dbutenhof: Yeah, I read that earlier; but I don't think there's anything "magical" about it, and you'd need to set that in order for `site` to find the relocated user site (so far as I was able to determine). Instead, I created a user configuration file causing the pbench user to automatically include `/opt/pbench-server/lib` in the PATH along with the local user site packages. This seems to work nicely, and means our Python interactive commands will run without any manual Python configuration environment variables.

Check out the new commit ...

dbutenhof:
FYI, after installation, (from root user on the server, with no `PYTHONPATH`):

```
[root@dbutenho-dev ~]# su pbench -c "python3 -m site"
sys.path = [
    '/root',
    '/usr/lib64/python36.zip',
    '/usr/lib64/python3.6',
    '/usr/lib64/python3.6/lib-dynload',
    '/home/pbench/.local/lib/python3.6/site-packages',
    '/opt/pbench-server/lib',
    '/usr/local/lib64/python3.6/site-packages',
    '/usr/local/lib/python3.6/site-packages',
    '/usr/lib64/python3.6/site-packages',
    '/usr/lib/python3.6/site-packages',
]
```

server/ansible/roles/pbench-server-activate-create-crontab/tasks/main.yml
```diff
@@ -1,5 +1,7 @@
 ---
 - name: create crontab
+  become: yes
+  become_user: pbench
```
dbutenhof, line 4: NOTE: I'm using this as the Ansible equivalent of `su pbench -c`. I ought to have been able to remove the `PYTHONPATH` definition from the `environment`, too, but I didn't think of that at the time.

webbnh (dismissed):

riya-17 (dismissed):

npalaska (dismissed):

server/rpm/pbench-server.spec.j2
```diff
@@ -58,8 +62,11 @@ cp -a ./web-server/* %{buildroot}/%{installdir}/%{static}
 mv %{buildroot}/%{installdir}/%{static}/package.json %{buildroot}/%{installdir}
 
 %post
-# Install python dependencies
-pip3 --no-cache-dir install --prefix=/%{installdir} -r /%{installdir}/requirements.txt > /%{installdir}/pip3-install.log
+# Install python dependencies for pbench user
+(su pbench -c "pip3 --no-cache-dir install --user -r /%{installdir}/requirements.txt") > /%{installdir}/pip3-install.log
+
+# Add install directory into the pbench user default PYTHONPATH
+su pbench -c "echo /%{installdir}/lib > \$(python3 -m site --user-site)/pbench.pth"
```
portante, line 69: What adds `pbench.pth` to the path?  Could this commit be missing a file change?

portante (changes requested):

server/rpm/pbench-server.spec.j2
dbutenhof: That's a feature of the `site` package, which is what sets up the initial default `sys.path`.

I can add to the comment, though, to clarify this.

portante: So by running `python3 -m site --user-site` it creates `pbench.pth`?

It looks like `python3 -m site --user-site` emits a path to which `pbench.pth` is appended, and then the `echo /${installdir}/lib` is creating that file.

But who reads `pbench.pth`?

dbutenhof: OK. Yeah, this is `site` magic, and it took me a while to work this out. Basically, the `site` package is implicit Python magic; it's run automatically when Python starts, and sets up `sys.path` on top of the external `PYTHONPATH`.

For our purposes, the `site` magic includes looking in the per-user site library and adding the `site-packages` paths *and* adding any directories captured in `*.pth` files in the root directory. This is pretty cool, if more than a bit obscure.

So the line here is taking our install path `/${installdir}/lib` and storing that path into our own little `pbench.pth` file for `site` to find when the pbench user runs Python.

That `pbench.pth` file needs to be in the right place, and the `python3 -m site --user-site` gives us the proper directory where it'll look. The `su pbench` ensures that this is run *as* "pbench" and therefore gives us the proper `/home/pbench/.local/lib/python3.6/site-packages` path where that file belongs.

Now, when the pbench user runs Python, *without* any `PYTHONPATH` environment variable, and asks for `sys.path`, it'll include the pbench user site-library paths, *plus* `/opt/pbench-server/lib`.

Maybe I over-complicated the example above, because I asked the site package to give me its nice formatting of `sys.path`. Let's try this instead:

```
[pbench@dbutenho-dev root]$ echo $PYTHONPATH

[pbench@dbutenho-dev root]$ whoami
pbench
[pbench@dbutenho-dev root]$ ls $(python3 -m site --user-site)
alembic                    botocore                    elasticsearch-7.13.1.dist-info   gunicorn-20.1.0.dist-info        s3transfer-0.5.0.dist-info
alembic-1.6.5.dist-info    botocore-1.21.28.dist-info  email_validator                  humanize                         sqlalchemy
aniso8601                  click                       email_validator-1.1.3.dist-info  humanize-3.11.0.dist-info        SQLAlchemy-1.4.23.dist-info
aniso8601-9.0.1.dist-info  click-8.0.1.dist-info       flask                            pbench.pth
boto3                      dateutil                    Flask-2.0.1.dist-info            python_dateutil-2.8.2.dist-info
boto3-1.18.28.dist-info    elasticsearch               gunicorn                         s3transfer
[pbench@dbutenho-dev root]$ cat $(python3 -m site --user-site)/pbench.pth
/opt/pbench-server/lib
[pbench@dbutenho-dev root]$ python3
Python 3.6.8 (default, Mar 18 2021, 08:58:41) 
[GCC 8.4.1 20200928 (Red Hat 8.4.1-1)] on linux
>>> import sys
>>> sys.path
['', '/usr/lib64/python36.zip', '/usr/lib64/python3.6', '/usr/lib64/python3.6/lib-dynload', '/home/pbench/.local/lib/python3.6/site-packages', '/opt/pbench-server/lib', '/usr/local/lib64/python3.6/site-packages', '/usr/local/lib/python3.6/site-packages', '/usr/lib64/python3.6/site-packages', '/usr/lib/python3.6/site-packages']
```

portante: Awesome!  That is a great explanation.  Would it be possible to place a comment in the `.spec.j2` file?

portante (changes requested):
If we could get a comment in the `.spec.j2` file describing this `pbench.pth` file that would cap this PR.

ndokos (approved):
Looks good but I too would appreciate a comment explaining the `site` magic.

server/rpm/pbench-server.spec.j2
portante: Thanks!

portante (approved):
Ship it!

webbnh (approved):
👍

webbnh:
> Blocked by PR #2439.

It doesn't _look_ blocked, except by [@portante](https://github.com/distributed-system-analysis/pbench/pull/2423#pullrequestreview-743139908).  🤣

webbnh (approved):
GitHub says there's nothing new here.

ndokos (approved):

ndokos:
There's the long comment about the `site` package in the spec file.

portante (approved):

PR #2427, opened 2021-08-20
Improve error handling for `ssh` failures from `pbench-move/copy-results`

portante:
Lately we have been servicing too many problems from incorrectly configured `pbench-agent` installations.  To help combat this a bit we improve the error handling and remove the use of `ssh -q` in order to expose the errors encountered.

We accomplish this using 2 separate commits: the first to stop use password prompts and the second which removes the `-q` usage and improves the error messages (a third fixes tabs vs spaces).


webbnh (dismissed):

portante (commented):
Once we get the latest jenkins forward ported to fix the tests, we should be good.

This will likely be back-ported to the `b0.69` branch.

riya-17 (approved):

webbnh:
@portante, I re-ran the tests for this PR and they are now passing, so this PR should be good to go.  However, it needs to be rebased, and I expect that you want it squashed.  The latter I could do with the Squash and merge" button; the former I could do with the "Update branch" button, but I'm not convinced that it will do a rebase (I suspect it will do a merge-from-`main`).  So, I'm going to leave it to you to do the rebase, and you can do the squash at the same time.  (We'll have to then re-review this PR, but I think it's worth it to ensure that we get what we want.)

dbutenhof (approved):
I like separating the failure of `ssh` itself from failure of the remote command, although the thought that the `exit` command might fail is rather bizarre ...

ndokos (approved):

portante:
> @portante, I re-ran the tests for this PR and they are now passing, so this PR should be good to go. However, it needs to be rebased, and I expect that you want it squashed. The latter I could do with the "Squash and merge" button, but that won't do the rebase; the former I could address with the "Update branch" button, but I'm not convinced that it will do a rebase (I suspect it will do a merge-from-`main`). So, I'm going to leave it to you to do the rebase, and you can do the squash at the same time. (We'll have to then re-review this PR, but I think it's worth it to ensure that we get the Git history that we want.)

I was hoping to leave the commits separated so that I can cherry pick back to `b0.69` easier without having to take all the whitespace changes, and so that the separated in-case an unforeseen problem arises.

agent/util-scripts/pbench-move-results
```diff
@@ -186,10 +186,14 @@ if [[ ! -z "${show_server}" ]] ;then
...
 
 # ssh probe test
-ssh -n -q -i ${pbench_bin}/id_rsa ${ssh_opts} ${results_repo} exit
-if [[ $? -ne 0 ]]; then
-    error_log "ERROR: results host unreachable: ${results_repo}"
-    debug_log "the following ssh command failed: \"ssh -n -q -i ${pbench_bin}/id_rsa ${ssh_opts} ${results_repo} exit\""
+ssh -n -i ${pbench_bin}/id_rsa ${ssh_opts} ${results_repo} exit
+sts=${?}
+if [[ ${sts} -eq 255 ]]; then
+    error_log "ERROR: ssh command, \"ssh -n -i ${pbench_bin}/id_rsa ${ssh_opts} ${results_repo} exit\", failed"
```
webbnh, line 192: It would be better if you captured the command in a variable which you used to execute the command and then also used to report errors with it.

webbnh (approved):
If you find a reason to make other changes, I've a suggestion below.  (If you decline to make that change, please go ahead and mark it resolved.)

dbutenhof (approved):

riya-17 (approved):

ndokos (approved):

agent/util-scripts/pbench-move-results
portante: Yes, it would be, but then we'd have to be sure we properly handle all the resolved variables.  Doesn't seem worth all that effort for this legacy code.

PR #2430, opened 2021-08-24
Add tests for admin role access

npalaska:
Adds unit tests for admin role access
1. Update user with admin token
2. Delete User with admin token
3. Admin user should not be able to delete himself/herself
4. Query APIs with admin token should pass (with payload containing another user)


lib/pbench/test/unit/server/query_apis/test_datasets_publish.py
```diff
@@ -83,16 +84,32 @@ class TestDatasetsPublish(Commons):
...
+    The reason for this might involve with the fact that we are accessing 2 different users using the
+    token (admin token) generated by another fixture that uses different database session.
+
+    In our test_query function the database session might expire after successful query (with admin token)
+    but 2 user instances might get left behind without getting properly expunged.
+
+    This often happens due to objects being in expired state, objects get expired after committing,
+    then when such expired objects are about to get used the ORM tries to refresh them, but this cannot be done
+    when objects are detached from session (e.g. because that session was closed). This behavior can be managed
+    by creating session with expire_on_commit=False param.
```
dbutenhof, line 101: Ugh.

I don't like twisting the production code `init_db` to deal with issues like this. There's got to be a better way, probably including a more thorough cleanup of the database. For example, using `yield` in the DB setup fixtures so that we can completely clean up after each test cycle.

And, again, if we can move towards a pure unit test environment that doesn't involve an actual DB, we wouldn't have to deal with this sort of issue at all ...

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
```diff
@@ -99,11 +98,11 @@ def test_query(
...
         # user, or for an invalid username.
         if (
-            not user
-            or user == "no_user"
-            or build_auth_header["header_param"] == "valid"
+            user == "no_user" or build_auth_header["header_param"] == "valid"
         ) and user != "badwolf":
             expected_status = HTTPStatus.OK
+        elif user == "drb" and build_auth_header["header_param"] == "valid_admin":
+            expected_status = HTTPStatus.OK
```
dbutenhof, line 105: These patterns show up repeatedly, and I keep wondering about trying to refactor them out into common code. I'm not quite sure that's feasible, but we're certainly creating a lot of extra work for ourselves every time we have to tweak them.

Part of this is just dealing with the bandwidth between our inflexible fixtures and the tests, and pulling some of the setup out of fixtures and into support methods with real parameters and returns would probably help.

Despite this rambling, I don't really expect you to try to solve that in this PR ... we really do, however, need to think about a massive unit test rearchitecture to clean things up. (Including dumping the live DB and Flask calls and moving to a real unit test system.)

dbutenhof (commented):
The change to the production `init_db` really bothers me. This is a small impact on the production server since the kwarg just isn't used on that side, but there's got to be a cleaner solution. Unfortunately that may not be practical without a massive amount of work, and the test case improvements here are certainly worthwhile.

lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
```diff
@@ -205,7 +207,7 @@ def test_empty_query(
         fixture.
         """
         expected_status = HTTPStatus.BAD_REQUEST
-        if build_auth_header["header_param"] != "valid":
+        if build_auth_header["header_param"] not in ["valid", "valid_admin"]:
```
webbnh, line 210: This conditional looks ripe for abstracting into a helper function -- the calling code wants to know if the header is valid, and it doesn't necessarily need to know the details of what makes it so.

lib/pbench/test/unit/server/query_apis/test_datasets_publish.py
webbnh: Obviously, for unit testing, mocking out the database methods would be the appropriate and superior approach.

Short of that, it seems like what is required is proper management of the database session -- i.e., create it at the beginning of the test, maintain it through the test, and the close it down when we're done (using a fixture with a `yield` like Dave suggests seems an excellent way to implement this).  But, yeah, it means that the production code has to be prepared for persistent sessions rather than relying on ephemeral ones.

lib/pbench/test/unit/server/test_user_auth.py
```diff
@@ -460,3 +460,53 @@ def test_non_existent_user_delete(client, server_config):
...
+                lastname="lastName",
+                email="user@domain.com",
+                password="12345",
+            )
+            assert resp_register.status_code == HTTPStatus.CREATED
+
+            # admin user login
+            resp_login = login_user(client, server_config, "test_admin", "12345")
+            data_login = resp_login.json
+            assert data_login["auth_token"]
```
webbnh, line 481: Isn't this what the `pbench_admin_token` fixture is for?

webbnh: Ditto lines 500-503.

webbnh (changes requested):
Overall it looks OK; but there are a number of small refactorings and similar bits of polishing which would be worthwhile.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
webbnh: > Including dumping the live DB and Flask calls and moving to a real unit test system.

While I'm in favor of explicitly mocking the DB calls, I think getting rid of Flask will just trade one kind of mocking for another.  (That is, if I'm not too confused, Flask provides nice facilities for mocking its `requests` package, and I don't _think_ we want to give that up.)

lib/pbench/test/unit/server/query_apis/test_datasets_publish.py
npalaska: Yeah its ugly unfortunately. However, I fixed this problem of expired orm instances by expunging before everytime we add a new user. Sqlalchemy docs says that a detached instance is an instance which corresponds, or previously corresponded, to a record in the database, but is not currently in any session and so we need to expunge them before adding.

I would also like to mention that the none of the test actually fails because of this  `DetachedInstanceError` it just prints the warning with big call stack and we probably garbage collect all the instances after the tests are finished but the warning was still annoying. Also It will take massive amount of rework in our testing architecture to get rid of database calls that we have at the moment, probably a task for another sprint.

lib/pbench/test/unit/server/conftest.py
```diff
@@ -497,15 +479,15 @@ def pbench_token(client, server_config):
     return data["auth_token"]
 
 
-class HeaderTypes(Enum):
+class HeaderTypes(IntEnum):
```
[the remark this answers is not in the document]
npalaska: Yep fixed it by having a HeaderTypes enum in a separate file

dbutenhof (approved):
One open conversation owned by Webb, which I believe you've resolved. Looks good to go!

webbnh (approved):
> One open conversation owned by Webb, which I believe you've resolved.

Nikhil had resolved the _code issue_...it should have been on me to resolve the _conversation_....

Anyway, now it's good to go.

dbutenhof:
> > One open conversation owned by Webb, which I believe you've resolved.
> 
> Nikhil had resolved the _code issue_...it should have been on me to resolve the _conversation_....
> 
> Anyway, now it's good to go.

I'm making a distinction between "resolving the comment" and "marking the conversation as resolved", but, sure. 🤷🏻‍♂️

webbnh:
And, I'm pointing out that, I think, Nikhil didn't pick up on that distinction, because he was the one who marked the conversation resolved.  😞

PR #2086, opened 2021-01-31
Fix use of date parsing in query API

portante:
The `parser.parse()` function will return a `datetime` object, filling in any missing date components missing from the input string. This means that when you call `parser.parse()` on the 29th, 30th, or 31st of a month, and you range spans months which do not have those dates, then those months are omitted.

The simple fix is to force the day to be the 1st, avoiding this problem.

Fixes #2084.

PR #2439, opened 2021-08-31
Fix unit test date_range method

dbutenhof:
Unlike the production code, the unit test Elasticsearch index URL builder fails to normalize the beginning and end dates to the beginning and end of the month. This can result in missing months and failing to match the mocked URL.

### DIFF
--- a/lib/pbench/test/unit/server/query_apis/commons.py
+++ b/lib/pbench/test/unit/server/query_apis/commons.py
@@ -3,6 +3,7 @@ import pytest
 import requests
 
 from dateutil import parser as date_parser, rrule
+from dateutil.relativedelta import relativedelta
 from http import HTTPStatus
 from typing import AnyStr, Type
 
@@ -71,7 +72,9 @@ class Commons:
         start_date = date_parser.parse(start)
         end_date = date_parser.parse(end)
         assert start_date <= end_date
-        for m in rrule.rrule(rrule.MONTHLY, dtstart=start_date, until=end_date):
+        first_month = start_date.replace(day=1)
+        last_month = end_date + relativedelta(day=31)
+        for m in rrule.rrule(rrule.MONTHLY, dtstart=first_month, until=last_month):
             date_range.append(f"{m.year:04}-{m.month:02}")
         return date_range
 


lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -71,7 +72,9 @@ def date_range(self, start: AnyStr, end: AnyStr) -> list:
         start_date = date_parser.parse(start)
         end_date = date_parser.parse(end)
         assert start_date <= end_date
-        for m in rrule.rrule(rrule.MONTHLY, dtstart=start_date, until=end_date):
+        first_month = start_date.replace(day=1)
+        last_month = end_date + relativedelta(day=31)
```
webbnh, line 76: Why add 31 days instead of 1 month?

webbnh: And, if what we're gunning for is the last day, it seems that something like this would be appropos:
```
        first_month = start_date
        last_month = end_date.replace(day=1) + relativedelta(months=1) - relativedelta(days=1)
```

webbnh: And, why are you using `relativedelta()` with `day` (_singular_) instead of `days` (plural) or instead of using `replace()` as you did for `first_month`?

webbnh (approved):
If you think this works, I'm OK with it, but it seems like it's more arcane than I would have hoped.

lib/pbench/test/unit/server/query_apis/commons.py
dbutenhof: > year, month, day, hour, minute, second, microsecond:

    Absolute information (argument is singular); adding or subtracting a
    relativedelta with absolute information does not perform an arithmetic
    operation, but rather REPLACES the corresponding value in the
    original datetime with the value(s) in relativedelta.

So the point here is to replace the day-of-month in the end `datetime` with the last day of that month. Using `days` would advance to the next month, without more complicated math like your suggestion, which we don't want. (And, yes, the parameter is maximized against the actual length of the month, so we'll get February 28 or 29, not 31, and there are relatively few months with more than 31 days ...)

dbutenhof: FYI, this change is copied and pasted from the production code equivalent. It was carefully researched and tested then, and probably still works. 😁

npalaska (approved):

portante (dismissed):
Restoring the original fix from https://github.com/distributed-system-analysis/pbench/pull/2086 works, we don't need `relativedelta`.

ndokos (changes requested):
I agree with Peter here: this looks like an elephant gun (with magic bullets) brought in to deal with the mosquito. And there is already a pattern in the code, so why use a different pattern here? If we are going to go with this pattern, then we should change the other place as well, but I vote to just change it here.

ndokos:
OK, I muddied the waters here:  the fix in #2086 seems to have gone away with the sands of time - I thought it was still around. And I see that `relativedelta` is used now. I still don't particularly like the magic, but I'll withdraw my objection.

ndokos (approved):

dbutenhof:
> OK, I muddied the waters here: the fix in #2086 seems to have gone away with the sands of time - I thought it was still around. And I see that `relativedelta` is used now. I still don't particularly like the magic, but I'll withdraw my objection.

I don't remember exactly why the end date fix-up was changed, and in particular whether it was because of another problem someone identified later. I wasn't inclined to mess with that at this point for the necessary quick fix today; it seemed to make a lot more sense to do exactly what the production `main` is doing.

ndokos:
I agree that that's the right fix (I didn't want two different patterns for the same thing in the code, so I could hardly disagree once I understood the current state), but the more I read about `relativedelta`, the less I like it: too many things going on under a single interface, with singular and plural versions doing radically different things - yecch.

Issue #2408, opened 2021-08-05
Testing strategies


dbutenhof:
### Discussed in https://github.com/distributed-system-analysis/pbench/discussions/2400

<div type='discussions-op-text'>

<sup>Originally posted by **dbutenhof** August  3, 2021</sup>
Rather than creating an issue that's primarily to begin a discussion, I figured I might as well just start it as a discussion. Throw in ideas and comments, and we can work towards defining a "testing epic" in Jira that we can boil down into discrete GitHub issues over time.

Right now, we have a fragmented testing environment including true unit tests, "fake" unit tests (e.g., server tests relying on a real database, and on the Flask package to make client API calls in order to test our `Resource` classes), and the "gold" unit tests (which are really pseudo-functional tests relying on a mixture of source code hackery and external setup), and pytest "functional tests" which drive real commands in an isolated contrived environment.

We need a true functional test environment, where we create a real server instance with backend services (PostgreSQL, Elasticsearch, Apache, NGINX, etc) which we can drive with real client interactions and additional observability hooks (including journald, database visualizer tools like Kibana and psql) to validate and diagnose the behavior of our code in an automated framework.

I think we also need to "simplify" our current testing frameworks to a straightforward pytest environment to truly isolate individual objects and methods for individual interface testing with all external interactions cleanly mocked. We shouldn't be making client POST/GET calls here; we should be calling methods of the class. We shouldn't be making DB calls; we should be mocking the results of all methods outside the specific class under test. The way we do it now is wasting time and resources testing our dependency code (SQLAlchemy, Flask, etc.) as "unit" tests. Validating all of that external integration is important, but it's not unit testing because it doesn't limit the failure domain to our own code.</div>

PR #2440, opened 2021-08-31
Iteration samples query API

npalaska:


lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,240 @@
...
+        object as CONTEXT so that the postprocess operation can proceed to get all
+        the iteration aggregations.
+        """
+        run_id = client_json.get("run_id")
+
+        # Query the dataset using the given run id
+        dataset = Dataset.query(md5=run_id)
+        if not dataset:
+            self.logger.error("Run ID {} not found", run_id)
+            abort(HTTPStatus.NOT_FOUND, message="Dataset not found")
```
dbutenhof, line 57: Should the message identify the `run_id`? Also, in the log message, it might be worth saying we can't find a Dataset for the specified run_id?

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,240 @@
...
+        if not owner:
+            self.logger.error(
+                "Dataset owner ID {} cannot be found in Users", dataset.owner_id
+            )
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="Dataset owner not found")
+
+        # For Iteration samples, we check authorization against the ownership of the
+        # dataset that was selected rather than having an explicit "user"
+        # JSON parameter. This will raise UnauthorizedAccess on failure.
+        self._check_authorization(owner.username, dataset["access"])
```
dbutenhof, line 69: `dataset` is also a SQLAlchemy model object with attributes, so `dataset.access` would be better.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,240 @@
...
+            user,
+            self.prefix,
+            start,
+            end,
+            run_id,
+        )
+
+        uri_fragment = self._gen_month_range("result-data-sample", start, end)
+        self.logger.info("fragment, {}", uri_fragment)
+        return {
```
dbutenhof, line 114: As we discussed, we need to loop on the `scroll_id` until we have all of the `[hits][hits]` elements collected.

**Also**, the dashboard `service/dashboard.js` code takes a *list* of `run_id` values, and returns all of them as a list of the raw (scrolled) Elasticsearch results.

We can decide to implement a single `run_id` query (which is easier on us but will require the dashboard to make a separate server call for each `run_id` in the list), or we can replace the entire service by taking a list of `run_id` values and searching for each one.

(I'm actually tempted to say we should stick to the single `run_id` here, but I want to point this out so we think about it deliberately.)

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,240 @@
...
+                                    }
+                                }
+                            ]
+                        }
+                    }
+                ]
+            }
+        }
+        """
+        return jsonify(es_json["aggregations"])
```
dbutenhof, line 240: And, again, we need to return the `hits` as well as the aggregations. If we're not including the model effect postprocessing, we probably should just return the raw Elasticsearch exactly as the dashboard service layer does now.

lib/pbench/server/database/models/datasets.py
```diff
@@ -632,6 +632,17 @@ def update(self):
             Database.db_session.rollback()
             raise DatasetSqlError("updating", self.controller, self.name) from e
 
+    @staticmethod
+    def query(md5: str) -> "Dataset":
+        """
+        Query dataset object based on md5 of the run document
+        """
+        try:
+            dataset = Database.db_session.query(Dataset).filter_by(md5=md5).first()
```
dbutenhof, line 641: It would be "fun" to generate an `_or` filter across a set of columns specified via `kwargs`. For some definition of "fun".

Unfortunately the default if specifying multiple columns is `_and`, which we wouldn't usually want. I think. Then again, the most likely use for multiple filters would be exclusive rather than combinatorial, and the `kwargs` just allow us to specify any one of the set, in which case `_and` vs `_or` doesn't matter ... 🙄

dbutenhof (commented):
Glad you got this started! A few comments just to be sure you don't get bored ...

lib/pbench/server/api/resources/query_apis/iteration_samples.py
webbnh: I'm not fully grokking this stuff yet, but I'm a little concerned about us collecting together the results of multiple scrolled calls and then returning it to the dashboard:  the ES results are returned piecemeal in order to avoid timeouts of the request to ES...if the Pbench Server aggregates those results, aren't we in danger of timing out the Dashboard's request to us?  And there's a similar concern regarding space:  the other reason for scrolled results is to avoid having the response grow too large; should we be concerned about a query which results in more entries than we can return in a single response?

I'm wondering if we aren't starting to solve a problem that we don't have (yet)...shouldn't we be waiting until the client has a discernible performance problem before we go implementing batching?

lib/pbench/server/api/resources/query_apis/iteration_samples.py
webbnh: We _should_ be trying to insulate the client from the details of our server implementation (e.g., our choice to use Elasticsearch)...subject to all of our other constraints (such as compatibility and ease of evolution), of course....

lib/pbench/server/api/resources/query_apis/iteration_samples.py
npalaska: Your concern is right Webb, we could end up with a query that's bigger than what we can return in single response. At the moment what the dashboard is doing is that it is going over all the iteration samples list again and aggregating the samples on top of all the aggregation returned from elasticsearch query. We could possibly do all that aggregation with elasticsearch query (by specifying more aggregation parameter) or perform the same aggregation on the server and only return the aggregated result. 
The problem is we don't fully understand the iteration samples postprocessing on the dashboard right now, so I am trying to set a meeting with Aquib and Anisha to discuss the what sort of aggregation we need to perform on the dashboard and if we can return only the aggregation that would be great.

dbutenhof: I neglected to actually *post* the earlier reply I made to this thread, but I'll add some more.

Yeah, it'd be nice if we could form a single *aggregation* that would include everything that the dashboard needs. If that's practical, we'd have a single Elasticsearch round-trip and latency wouldn't be an issue.

The original plan (represented in my architecture document) had been to limit the Elasticsearch dependency by moving both the dashboard service and model code into the server. However, as Nikhil says, the dashboard model code is complicated... and as Webb points out, moving all the Elasticsearch trips into the server poses a risk of exceeding client timeouts, and may not be a great idea. (We could consider building an asynchronous operation and an infrastructure to allow notifying the client when it's done, but that's a lot of infrastructure.)

The simplest immediate resolution would be to let the dashboard service layer continue to make repeated calls to the server, with each one fetching another scroll page of Elasticsearch data which the dashboard would consolidate as it does now. That's a bit messy, but a bit less messy than some of the other possible solutions.

npalaska: > The simplest immediate resolution would be to let the dashboard service layer continue to make repeated calls to the server, with each one fetching another scroll page of Elasticsearch data which the dashboard would consolidate as it does now. That's a bit messy, but a bit less messy than some of the other possible solutions.

So that means either sending back the elasticsearch scroll id or sending some sort of signal that indicate _there is another page of result possible would you like to scan_ and if the server receives `True` for the signal we scroll the next page (keeping the scroll_id in memory). We also have to make sure the scroll id doesnt expire in the meantime.

lib/pbench/test/unit/server/query_apis/test_iteration_samples.py
```diff
@@ -0,0 +1,328 @@
...
+            index,
+            expected_status,
+            json=response_payload,
+            status=HTTPStatus.OK,
+            headers=build_auth_header["header"],
+            scroll=True,
+        )
+        if expected_status == HTTPStatus.OK:
+            res_json = response.json
+            expected_result = {
```
webbnh, line 200: How is this value of `expected_result` different from the value of `response_payload` set at line 46?  Do we really want to maintain independent respective definitions for them?

lib/pbench/test/unit/server/query_apis/conftest.py
```diff
@@ -41,10 +42,38 @@ def query_api(
...
+    es_scroll_response = {
+        "_scroll_id": "same_random_scroll_id==",
+        "took": 3,
+        "timed_out": "false",
+        "_shards": {"total": 5, "successful": 5, "skipped": 0, "failed": 0},
+        "hits": {
+            "total": {"value": 50, "relation": "eq"},
+            "max_score": "null",
+            "hits": [],
+        },
```
webbnh, line 76: This is not going to have any discernible effect -- you need to return at least one hit so that, when this is joined together with the original response, the result is different from the original.

(That is, you should have two nearly-identical tests, one with `scroll` set to `True` and one with it `False`, and they should have different "expected results".)

webbnh (changes requested):
This stuff looks quite good.  I have lots of feedback, most of which is stuff about comments and other nits; however, there are a few slightly bigger items.  I think you're not actually testing the scroll support effectively, and you need another test for that; and there are a couple of others.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,302 @@
...
+        # JSON parameter. This will raise UnauthorizedAccess on failure.
+        self._check_authorization(owner.username, dataset.access)
+
+        # The dataset exists, and authenticated user has enough access so continue
+        # the operation with the appropriate CONTEXT.
+        return {"dataset": dataset}
+
+    def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
+        """
+        Construct a pbench search query for aggregating iteration samples of given run id
```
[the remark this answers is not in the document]
webbnh: We're really not returning any aggregations -- what we're doing is returning information about a benchmark run (like the names of the metrics collected) along with timeseries data.  The term "aggregation" is an ES artifact -- it's how we fetch the metrics list -- we don't actually aggregate any _data_....  😁

lib/pbench/server/api/resources/query_apis/iteration_samples.py
webbnh: The ES documentation offers [an alternative to using scrolling](https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html#search-after) (which only works on queries of less than 10,000 documents anyway).  We might be able to build something on that, instead.

lib/pbench/test/unit/server/conftest.py
```diff
@@ -403,17 +403,17 @@ def fake_find(name: str) -> Template:
...
+                template_name="unit-test.v6.result-data-sample",
                 file="run.json",
-                template_pattern="unit-test.v5.result-data.*",
-                index_template="unit-test.v5.result-data.{year}-{month}",
+                template_pattern="unit-test.v6.result-data.*",
+                index_template="unit-test.v6.result-data-sample.{year}-{month}",
                 settings={"none": False},
                 mappings={
-                    "_meta": {"version": "5"},
+                    "_meta": {"version": "6"},
```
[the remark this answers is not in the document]
npalaska: No its just to make the unit tests simpler by keeping the version same for all the documents, so we dont have to mange the version numbers across multiple api tests. Like run-data documents are version 6 but result-data-sample documents are version 5, so if we keep them different we have to make sure we get the right version number when dynamically building the indices. Keeping them same makes it simpler

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,302 @@
...
+                                },
+                            ]
+                        }
+                    },
+                    "aggs": {
+                        "id": {
+                            "terms": {"field": "run.id"},
+                            "aggs": {
+                                "type": {
+                                    "terms": {"field": "sample.measurement_type"},
```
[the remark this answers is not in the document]
npalaska: yeah thats true

lib/pbench/test/unit/server/query_apis/test_iteration_samples.py
```diff
@@ -0,0 +1,328 @@
...
+        response_payload = {
+            "_scroll_id": "random_scroll_id_string==",
+            "took": 14,
+            "timed_out": "false",
+            "_shards": {"total": 5, "successful": 5, "skipped": 0, "failed": 0},
+            "hits": {
+                "total": {"value": 50, "relation": "eq"},
+                "max_score": "null",
+                "hits": [
+                    {
```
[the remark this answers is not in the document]
npalaska: `hits.total.value` is always the total number of all the documents matched (across all the shards) irrespective of the `size` parameter in query. The size param would just limit the number of hits we see in the list in `hits.hits`

lib/pbench/test/unit/server/query_apis/conftest.py
npalaska: Yeah that's true but its probably little complicated to implement since we are mocking the scroll endpoint request that always return the same response, and so it would never come out of loop from inside the scroll function (since there will always be a response back from the scrolling)

lib/pbench/server/database/models/datasets.py
```diff
@@ -632,6 +632,17 @@ def update(self):
...
+    @staticmethod
+    def query(**kwargs) -> "Dataset":
+        """
+        Query dataset object based on md5 of the run document
+        """
+        try:
+            dataset = Database.db_session.query(Dataset).filter_by(**kwargs).first()
+            return dataset
+        except Exception as e:
+            raise DatasetSqlError("Querying dataset failed with {}", kwargs) from e
```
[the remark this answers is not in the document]
npalaska: or we could just raise the exception as it is right?
```
except Exception:
    raise 
```

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -266,10 +293,18 @@ def _call(self, method: Callable, json_data: JSON):
...
             es_response = method(url, **es_request["kwargs"])
+            es_response.raise_for_status()
+
+            res_json = es_response.json()
+            res_json = self._scroll(res_json)
+
             self.logger.debug(
                 "ES query response {}:{}", es_response.reason, es_response.status_code
             )
-            es_response.raise_for_status()
```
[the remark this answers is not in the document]
npalaska: Fixed

lib/pbench/test/unit/server/query_apis/test_iteration_samples.py
```diff
@@ -0,0 +1,328 @@
...
+                                "buckets": [
+                                    {
+                                        "key": "throughput",
+                                        "doc_count": 50,
+                                        "title": {
+                                            "doc_count_error_upper_bound": 0,
+                                            "sum_other_doc_count": 0,
+                                            "buckets": [
+                                                {
+                                                    "key": "Gb_sec",
```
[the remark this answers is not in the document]
npalaska: It doesnt for the purpose of the unit test, but I should fix this so things are in sync and not random

lib/pbench/test/unit/server/query_apis/conftest.py
npalaska: But anyway that should be doable with the help of reposnes callbacks https://github.com/getsentry/responses

lib/pbench/test/unit/server/conftest.py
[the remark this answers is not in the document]
webbnh: OK, I'm just not aware of what the version numbers really mean, what they are used for, and by whom, so I don't understand what the implications of changing them are.  I realize that they have to be set to _something_, and I concur that it would be best if they were set in a realistic fashion (e.g., so that the documents match the index they are supposedly coming from, etc.), but I also think that they should be meaningful, that they should serve a purpose (i.e., that we're not maintaining them for no reason), and that, therefore, they only change when they are supposed to.

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -235,23 +235,25 @@ def _scroll(self, res_json: JSON) -> JSON:
...
+            return
+        scroll_api_url = f"{self.es_url}/_search/scroll"
+        data = res_json["hits"]["hits"]
+        scroll_id = res_json["_scroll_id"]
+        while data:
+            scroll_res = requests.post(
+                scroll_api_url,
+                data=json.dumps(
+                    {"scroll": "1m", "scroll_id": scroll_id}
+                ),  # Keeps the scrolling context live for 1 min
```
webbnh, line 248: Avoid using abbreviations -- spell out "minute" (e.g., so it's not confused with "minimum").

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -235,23 +235,25 @@ def _scroll(self, res_json: JSON) -> JSON:
...
+                headers={"Content-Type": "application/json"},
+            )
+            scroll_res.raise_for_status()
+            scroll_res = scroll_res.json()
+            data = scroll_res["hits"]["hits"]
+            res_json["hits"]["hits"] += data
+            scroll_id = scroll_res["_scroll_id"]
+        return
```
webbnh, line 256: Since this function doesn't return a value, there's no need to have a `return` statement if it's the last statement in the body.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
dbutenhof: That's intriguing, although I'm not entirely sure it helps. We probably don't need a `PIT` as we don't expect the result documents to change during the search, but we'd still need to retain the "search after" across the multiple client calls, so it's probably not really any easier than using the scroll ID. (Although without the PIT, I suppose we do dispense with that awkward lifetime issue, I'm not sure that the 1 minute interval imposes a massive hardship on the dashboard and it lets *us* throw away our internal storage -- PIT or scroll ID -- so we don't inherit the same lifetime problem that motivated Elasticsearch to add the `scroll=timelimit` syntax.)

lib/pbench/server/database/models/datasets.py
```diff
@@ -560,6 +560,17 @@ def attach(path=None, controller=None, name=None, state=None) -> "Dataset":
...
+    @staticmethod
+    def query(**kwargs) -> "Dataset":
+        """
+        Query dataset object based on a given column name of the run document
+        """
+        try:
+            dataset = Database.db_session.query(Dataset).filter_by(**kwargs).first()
+            return dataset
+        except Exception:
+            raise
```
dbutenhof, line 572: Don't ever catch an exception just to raise it; that's pointless. This is exactly the same as omitting the `try` entirely, except slower and harder to read.

On the other hand, a log message here might not be bad, and generally I was passing on SQLAlchemy errors using the `DatasetSqlError` exception, which would be implicitly chained from whatever you'd caught here.

lib/pbench/test/unit/server/conftest.py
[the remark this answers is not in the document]
dbutenhof: Hmm; this is fine to some extent, however the various index templates *do* have different versions, and it was somewhat nice to be testing that this carries through to the application code. I'm not sure how I feel about this... but it bothers me just enough to unresolve Webb's conversation and make everyone think about it again.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -304,7 +309,14 @@ def test_http_exception(
...
     def test_http_error(
-        self, server_config, query_api, user_ok, find_template, pbench_token, errors
+        self,
+        server_config,
+        query_api,
+        user_ok,
+        find_template,
+        pbench_token,
+        attach_dataset,
+        errors,
```
dbutenhof, line 319: Wow, that's a lot of fixtures...

lib/pbench/test/unit/server/query_apis/conftest.py
```diff
@@ -41,10 +56,117 @@ def query_api(
...
+        "_shards": {"total": 5, "successful": 5, "skipped": 0, "failed": 0},
+        "hits": {
+            "total": {"value": 2, "relation": "eq"},
+            "max_score": "null",
+            "hits": [],
+        },
+    }
+    es_scroll_responses = {}
+    es_scroll_responses["random_scroll_id_string_1=="] = es_scroll_response
+    es_scroll_responses["random_scroll_id_string_2=="] = es_empty_response
```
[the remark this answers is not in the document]
webbnh: It's somewhat silly to dynamically construct these objects when they never change; you could do them statically, I think (although, the fact that this is `conftest.py` makes me hesitate to state anything firmly...).

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -111,10 +110,10 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
             "path": f"/{uri_fragment}/_search?scroll=1m",
             "kwargs": {
                 "json": {
-                    "size": 100,
+                    "size": 100,  # The maximum default size for scrolling is 10000
```
webbnh, line 113: I don't think that comment has the right phrasing.  The [ES documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/scroll-api.html) puts it thusly:

> We no longer recommend using the scroll API for deep pagination. If you need to preserve the index state while paging through more than 10,000 hits, use the `search_after` parameter with a point in time (PIT).

so, 10,000 isn't the "maximum default", it's more like the upper limit on what you can reach via scrolling....

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -163,23 +151,23 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
 
     def postprocess(self, es_json: JSON, context: CONTEXT) -> JSON:
         """
-        Returns all iteration samples as json containing the aggregation data.
+        Returns all iteration samples as JSON containing the aggregation data.
         The 2 top level keys are:
-        "hits": Continaing information about total number of hits and
-                list of all the iterations samples json across all the scrolled pages
+        "hits": Containing information about total number of hits and
+                list of all the iterations samples as JSON returned from the ES query.
```
webbnh, line 157: I would omit "returned from the ES query", but if you want to keep it, I would replace "query" with "query(es)" or "queries".

webbnh (changes requested):
The improvements look good.

However, I think you should rework `get_iteration_samples_scroll_payload()`.  It appears to be two functions bolted together as one which don't need to be combined; one of them is simple enough that it could be inlined; and, the other is dynamically constructing what otherwise could be static objects.  So, I suggest that you consider getting rid of it altogether, although you might want to still have a little function which returns the payload after checking that the scroll ID is appropriate (but the payload could be statically declared at the module level).

And there's a bunch of other nits and bits to be polished.

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -226,6 +227,34 @@ def postprocess(self, es_json: JSON, context: CONTEXT) -> JSON:
...
+            return
+        scroll_api_url = f"{self.es_url}/_search/scroll"
+        data = res_json["hits"]["hits"]
+        scroll_id = res_json["_scroll_id"]
+        while data:
+            scroll_res = requests.post(
+                scroll_api_url,
+                data=json.dumps(
+                    {"scroll": "1m", "scroll_id": scroll_id}
+                ),  # Keeps the scrolling context live for 1 min
```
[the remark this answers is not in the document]
webbnh: Per the [ES docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/scroll-api.html#scroll-api-query-params), 

> This value overrides the duration set by the original search API request’s scroll parameter.

So, each time, this query buys us another minute's grace.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,279 @@
...
+    """
+
+    def __init__(self, config: PbenchServerConfig, logger: Logger):
+        super().__init__(
+            config,
+            logger,
+            Schema(
+                Parameter("start", ParamType.DATE, required=True),
+                Parameter("end", ParamType.DATE, required=True),
+                Parameter("run_id", ParamType.STRING, required=True),
```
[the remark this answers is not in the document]
webbnh: From what you were saying about fio, it sounds like scrolling isn't actually viable, and we should be looking for an "after" point and maybe a "point in time" reference, as well.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,302 @@
...
+
+            "run_id": String representation of run id to match for aggregate all the iteration samples
+        """
+        start = json_data.get("start")
+        end = json_data.get("end")
+        run_id = json_data.get("run_id")
+
+        # We need to pass string dates as part of the Elasticsearch query; we
+        # use the unconverted strings passed by the caller rather than the
+        # adjusted and normalized datetime objects for this.
```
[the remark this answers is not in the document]
webbnh: True, there are other instances of this text which are your fault, Dave, but _this_ one belongs to Nikhil!  😉

lib/pbench/server/api/resources/query_apis/iteration_samples.py
webbnh: I grant that it's not much _easier_, but it's supposed to actually **work** (for large document-sets), which is plus.  😁 

If we can do without a `PIT` (which I, like you, suspect we can), then things get easier for everyone.  For starters, we're not racing against time, which is very nice.  And, if we can agree on the handshake, then we can let the client track the "search after" point and just hand it back to the server (e.g., as "opaque context"), so the server doesn't need to worry about keeping it and cleaning it up.

lib/pbench/server/database/models/datasets.py
webbnh: > generally I was passing on SQLAlchemy errors using the DatasetSqlError exception

That's what Nikhil _was_ doing prior to this last round of changes, and [apparently I talked him out of it](https://github.com/distributed-system-analysis/pbench/pull/2440#discussion_r708466285). 🙂  (In my defense, he was having trouble making that exception object fit his case, and it was upsetting poor LGTM.)

lib/pbench/server/api/resources/query_apis/iteration_samples.py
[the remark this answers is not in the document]
npalaska: @webbnh the problem is that for the result-data-sample documents there isn't a clear _after point_ unfortunately. Iteration number which we are using now for sorting can not be used as an after point, since there could be thousands of documents with the same iteration number (kibana showed almost 33,000 documents with the same iteration number for a single run.id). Therefore even if we return 10,000 documents and then use `search_after` using the iteration number we could lose 23000 documents because its gonna skip all those documents. Now even the timestamp is not unique across all the hits of the result-data-sample documents, there are multiple documents with same timestamp and we end-up in same situation where we can't use timestamp for search_after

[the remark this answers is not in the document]
webbnh: Yeah, we _really_ need to figure out **why** somebody might want _all_ of these documents.

That is, if they want a particular time-series, then we can select that and page it with `search_after` using the timestamp.  (That is, I'm not surprised that timestamps aren't unique across different sets of time-series data, but I bet within a single series they are unique.)  Alternatively, we can have ES do time-sampled aggregations of the time-series data, and return however many "buckets" as they want (within reason...)...which will get us down from tens of thousands to whatever number the server and the dashboard agree on.

But, if they aren't actually interested in the time-series data, then I bet we can have ES aggregate all the documents for us, and we can return a single result.  I seem to recall that what the dashboard really wanted was a set of column headings that it could offer to the user; I'm guessing that the user's choice would then be used to select which actual documents to fetch.  So, I'm guessing that there's a solution here which doesn't involve returning all XX,000 documents on the first pass (or ever).

lib/pbench/server/database/models/datasets.py
```diff
@@ -560,6 +560,17 @@ def attach(path=None, controller=None, name=None, state=None) -> "Dataset":
             dataset.advance(state)
         return dataset
 
+    @staticmethod
+    def query(**kwargs) -> "Dataset":
+        """
+        Query dataset object based on a given column name of the run document
+        """
+        try:
+            dataset = Database.db_session.query(Dataset).filter_by(**kwargs).first()
```
portante, line 569: Is it really worth this one line for fetching a `dataset`?  Is there a higher abstraction we should be implementing?

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,279 @@
...
+            }
+        }
+        """
+        try:
+            return jsonify(
+                {"hits": es_json["hits"], "aggregations": es_json["aggregations"]}
+            )
+        except KeyError as e:
+            raise PostprocessError(
+                HTTPStatus.BAD_REQUEST,
```
portante, line 277: Is this really a `BAD_REQUEST` or is it an `INTERNAL_ERROR`?

If the response from the Elasticsearch instance does not have `hits` or `aggregations` then is that the callers fault, or does it indicate some other error on our side?

portante (changes requested):
It seems like we could benefit from using the `elasticsearch` python module to help with the implementation.

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
```diff
@@ -0,0 +1,419 @@
...
+    aggregated term query filtered by user supplied run id.
+    """
+
+    WHITELIST_AGGS_FIELDS = ["run", "sample", "iteration", "benchmark"]
+
+    def __init__(self, config: PbenchServerConfig, logger: Logger):
+        super().__init__(
+            config,
+            logger,
+            Schema(Parameter("run_id", ParamType.STRING, required=True)),
```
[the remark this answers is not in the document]
npalaska: I think it will accept an empty string as a `run_id` but it will abort with FORBIDDEN status (since Dataset with empty run id will not be not be found)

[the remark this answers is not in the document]
webbnh: Did you mean `NOT_FOUND`, rather than `FORBIDDEN`?  (I'm looking at `RunIdBase.preprocess()` line 64 in `lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py`.)

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
```diff
@@ -236,161 +231,106 @@ def scroll_assemble(self, scroll_id: str, context: CONTEXT) -> JSON:
...
+        run_id = context["run_id"]
+        dataset = context["dataset"]
 
         self.logger.info(
-            "Return iteration sample rows for dataset {}, prefix {}, " "run id {}",
+            "Return iteration sample rows for dataset {}, prefix {}, run id {}",
             dataset,
             self.prefix,
             run_id,
         )
```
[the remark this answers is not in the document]
webbnh: The change to the message looks good...but I'm still unclear on whether the logging level should be `info` or `debug`.

webbnh (approved):
Nice cleanup.

I found one additional typographical, and I've got two lingering questions.  Otherwise, it's good by me.

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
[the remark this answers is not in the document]
webbnh: Answered [here](https://github.com/distributed-system-analysis/pbench/discussions/2509).  So, this code is fine as coded now.

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
```diff
@@ -0,0 +1,347 @@
...
+            "run_id": run id of a dataset that client wants to search for
+                    fetching the iteration samples.
+
+            "scroll_id": Server provided Elasticsearch scroll id that client
+                        recieved in the result of the original query with
+                        filters. This can be used to fetch the next page of
+                        the result.
+
+            "filters": key-value representation of query filter parameters to
+                       narrow the search results e.g. {"sample.name": "sample1"}
```
dbutenhof, line 235: These should be indented a tab stop; and, ideally, the continuation line indentation would match down the list.

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
```diff
@@ -0,0 +1,347 @@
...
+
+        EXAMPLE:
+            {
+                "run_id": "1234567"
+                "scroll_id": "cmFuZG9tX3Njcm9sbF9pZF9zdHJpbmdfMg=="
+                "filters": {"sample.name": "sample1"},
+            }
```
dbutenhof, line 242: Not really a useful example, since `scroll_id` is mutually exclusive with the other two...

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
```diff
@@ -0,0 +1,347 @@
...
+            "Return iteration sample rows {} for dataset {}, prefix {}, run id {}",
+            "next page " if scroll_id else "",
+            dataset,
+            self.prefix,
+            run_id,
+        )
+
+        scroll_id = json_data.get("scroll_id")
+        if scroll_id:
+            # Note: Scroll id expires in 1 minute
```
dbutenhof, line 258: I almost hate to throw this out there; but you're generating the `1m` scroll duration in two separate places now, and for debugging it *might* be handy to be able to extend that without searching for all the places its used. Maybe another constant is called for here? 😬

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
```diff
@@ -0,0 +1,347 @@
...
+                },
+                "params": {"ignore_unavailable": "true"},
+            },
+        }
+
+    def postprocess(self, es_json: JSON, context: CONTEXT) -> Response:
+        """
+        Returns a Flask Response containing a JSON object with keys as
+        results and possibly a scroll_id if the next page of results
+        is available.
```
dbutenhof, line 289: Technically, we don't know if another page of results is available, only that Elasticsearch returned a scroll ID; but I'm not quite sure how to make this clear without weaseling...

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
```diff
@@ -0,0 +1,347 @@
...
+        # table
+        index = self.get_index(dataset, "result-data-sample")
+
+        try:
+            template = Template.find("result-data-sample")
+        except TemplateNotFound:
+            self.logger.exception(
+                "Document template 'result-data-sample' not found in the database."
+            )
+            abort(HTTPStatus.NOT_FOUND, message="Mapping not found")
```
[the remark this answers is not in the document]
npalaska: Agree

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
npalaska: Yeah agree, there need to be 2 examples here one with filters and one with scroll_id. However, `scroll_id` is not mutually exclusive with the other two, its mutually exclusive only with `filters` parameter.

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
webbnh: Actually, I think that ES returns the scroll ID any time you ask for one...regardless of whether there are multiple "pages" or not.  😿

However, we think that we can predict whether there will be another page in all but one case; and, so, when (we think) there is no next page, we won't return a scroll ID (but, in one case, we'll return a scroll ID when the next page is going to be empty...but hopefully the client can handle that).

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
```diff
@@ -223,23 +224,27 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
...
             {
                 "run_id": "1234567"
-                "scroll_id": "cmFuZG9tX3Njcm9sbF9pZF9zdHJpbmdfMg=="
                 "filters": {"sample.name": "sample1"},
             }
+            or
+            {
+                "run_id": "1234567"
+                "scroll_id": "cmFuZG9tX3Njcm9sbF9pZF9zdHJpbmdfMg=="
+            }
```
webbnh, line 247: In the case where the request provides a `"scroll_id"`, our only use for `"run_id"` is to log the request.

I'm not saying that that is good or bad, and it doesn't seem onerous to require the requestor to include it, but, looking under the covers, it's slightly weird; however, if we were to change the underlying database in the future, we might need the run ID then, so perhaps it's good to require it even though we don't really need it.

webbnh (approved):
Looks good to me.

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
npalaska: `run_id` is not only used to log the request, it as also used to check authorization against the ownership of the dataset, since we can not get user information from `scroll_id`

webbnh: Ah, OK.  Nevermind.  🙂

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
```diff
@@ -0,0 +1,356 @@
...
+
+        If a scroll_id is specified, the query will return the next page of
+        results of the original query; otherwise we form a new query, either
+        with the specified filters or without any filters (returning up to 10000
+        documents at a time).
+
+        Args:
+            json_data:
+                "run_id": run id of a dataset that client wants to search for
+                        fetching the iteration samples.
```
dbutenhof, line 228: Ha. I'd commented before that `scroll_id` supersedes both `run_id` and `filters` ... you were correct in pointing out that's not true, but perhaps it's worth pointing out here in the description that (even though we won't use it in the Elasticsearch query) we *do* need `run_id` in order to authorize access to the dataset on subsequent scroll-id calls.

This may actually be an argument for moving to some internally-managed context across scrolls, because we could reliably ensure that the datasets match. (Not that it's a security hole, really, since in order to get a scroll ID you need to have supplied an initial dataset you can read; and the worst you could do by changing dataset in a follow-on scroll call would be to select a different dataset you *can't* read.)

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
```diff
@@ -0,0 +1,356 @@
...
+            self.prefix,
+            run_id,
+        )
+
+        scroll_id = json_data.get("scroll_id")
+        if scroll_id:
+            return {
+                "path": "/_search/scroll",
+                "kwargs": {
+                    "json": {
```
dbutenhof, line 266: Wait; shouldn't this be `"params"` rather than `"json"`??

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
dbutenhof: I was commenting on the "is available" ... you're correct that our heuristic isn't 100% reliable, but, more to the point, something like "if we think there might be more data" or "if we can't determine there is no more data".

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+    def get_index(self, dataset: Dataset, root_index_name: AnyStr) -> AnyStr:
+        """
+        Retrieve the list of ES indices from the metadata table based on a given
+        root_index_name.
+        """
+        try:
+            index_map = Metadata.getvalue(dataset=dataset, key=Metadata.INDEX_MAP)
+            index_keys = [key for key in index_map if root_index_name in key]
+        except MetadataError as e:
+            self.logger.error(f"Indices from metadata table not found {e!r}")
```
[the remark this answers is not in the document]
webbnh: Like Dave said, not a showstopper, but if you happen make other changes here, it would probably be good to move the assignment of `index_keys` out of the `try` block, since any exceptions that _it_ raises won't be caught here.  (It's a good idea for the `try` block to cover as small a range of code as reasonable and not to have extraneous statements inside it.)

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
webbnh: I concur that having that called out in a comment would be good, but I don't know that _here_ is the right place...that seems more like a schema thing.

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
webbnh: No, those "parameters" are supposed to conveyed in the payload.  (I think you can pass the ID as a parameter, but I believe that that practice is deprecated.)

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
webbnh: > "if we can't determine there is no more data"

That would be accurate...but I don't know that I like it.  (I don't mind "is available".)

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
webbnh: Because, at the moment, each of these will do the same thing:
```
GET /_search/scroll/<scroll_id>?scroll="1m"
GET /_search/scroll?scroll_id=<scroll_id>&scroll="1m"
POST /_search/scroll/<scroll_id>?scroll="1m"
POST /_search/scroll?scroll_id=<scroll_id>&scroll="1m"
POST /_search/scroll
    BODY { "scroll_id"=<scroll_id>, "scroll"="1m"}
```
but four of them are now deprecated.

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
webbnh: Actually, for just a comment, I think it would be a very good idea to include it in this PR:  it's low risk, and, if we don't do it now, when _are_ we going to do it?

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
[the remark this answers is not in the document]
npalaska: It seems like there are no outstanding conversations that needs addressing so if this is not a showstopper should I squash all the commits?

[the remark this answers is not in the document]
npalaska: Ah I still need to address this one https://github.com/distributed-system-analysis/pbench/pull/2440#discussion_r741865343 NM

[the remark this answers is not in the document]
dbutenhof: > It seems like there are no outstanding conversations that needs addressing so if this is not a showstopper should I squash all the commits?

I see 6 open conversations. A few of them were already addressed, but some aren't. The main open issue for me is the current lack of "unit" testing of the no-filter case.

lib/pbench/test/unit/server/query_apis/test_iteration_samples.py
```diff
@@ -318,3 +319,544 @@ def test_query(
...
+    @pytest.fixture(autouse=True)
+    def _setup(self, client):
+        super()._setup(
+            cls_obj=IterationSamplesRows(client.config, client.logger),
+            pbench_endpoint="/dataset/samples/rows",
+            elastic_endpoint="/_search",
+            payload={
+                "run_id": "random_md5_string1",
+                "filters": {"sample.name": "sample1"},
+            },
```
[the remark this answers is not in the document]
npalaska: Fixed

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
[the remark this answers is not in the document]
dbutenhof: This wasn't wasn't technically resolved; but since I started it, and both Webb and I called it "not a showstopper", I'm going to mark it resolved. (I still wouldn't mind if you fixed it, should you end up needing to touch this file again.)

dbutenhof (approved):
I think this leaves unresolved conversations only for @webbnh @portante ... one minor cleanup would be nice, which I marked on a resolved conversation, but otherwise (aside from the technically of those open conversations), you're ready to go.

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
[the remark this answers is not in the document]
npalaska: Ah that is correct, yes it would return `NOT_FOUND`

webbnh (dismissed):
If Dave's happy, then I'm happy.  🙂

lib/pbench/server/database/models/datasets.py
portante: This is no longer a valid comment since this code was rebased on commits which addressed this.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
portante: Was fixed.

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
+
+        if len(index_keys) == 0:
+            self.logger.error(
+                f"Found no indices matching the prefix {root_index_name}"
+                f"for a dataset {dataset!r}"
+            )
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="Found no matching indices")
+        index = "".join([f"{index}," for index in index_keys])
```
portante, line 101: Let's consider:

```
index = ",".join(index_keys)
```

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
```diff
@@ -0,0 +1,356 @@
...
+        self.logger.info(
+            "Return iteration sample namespace for dataset {}, prefix {}, " "run id {}",
+            dataset,
+            self.prefix,
+            run_id,
+        )
+
+        # Retrieve the ES indices that belong to this run_id from the metadata
+        # table
+        index = self.get_index(dataset, "result-data-sample")
```
portante, line 79: Can we rename this to `indices` to indicate the plural nature of the value?

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
```diff
@@ -0,0 +1,356 @@
...
+        # Get all the fields from result-data-sample document mapping that
+        # are of type keyword
+        result = self.get_keyword_fields(mappings)
+
+        # Build ES aggregation query for getting the iteration samples
+        # namespaces
+        aggs = {key: {"terms": {"field": key}} for key in result}
+
+        return {
+            "path": f"/{index}/_search",
```
portante, line 107: Perhaps `indices` instead of `index`?

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+            index_map = Metadata.getvalue(dataset=dataset, key=Metadata.INDEX_MAP)
+            index_keys = [key for key in index_map if root_index_name in key]
+        except MetadataError as e:
+            self.logger.error(f"Indices from metadata table not found {e!r}")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
+
+        if len(index_keys) == 0:
+            self.logger.error(
+                f"Found no indices matching the prefix {root_index_name}"
+                f"for a dataset {dataset!r}"
```
portante, line 98: Why is this an error at all?  Don't `pbench-user-benchmark` results fall in this category?

portante (changes requested):

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
portante: That is to say a "run" does not have any "result" data.

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
webbnh: Ha!  I proposed that awhile ago, but apparently we actually want the trailing comma for some reason.  😕

npalaska: Well, I think from elasticsearch point of view we don't need a trailing comma, but for our tests to have a proper mock response we need to match the way we build index in mocks vs in our APIs. So if we make a change in API then we we also need to make a same change in unit tests AFAIK.

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
npalaska: Perhaps this should also return `NOT_FOUND` instead of `SERVER_ERROR`

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
npalaska: Fixed

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
npalaska: Fixed

dbutenhof (dismissed):

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
portante: In principle we should not have our test code govern what we do in our product code.

portante (changes requested):
Looking good, just need to use `NOT_FOUND` when we don't find any result data sample documents for a given pbench run.

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
dbutenhof: Technically, this is letting the historic dashboard implementation wag our tail; I picked up the trailing comma when I first started us on this path, and it's become a sort of tradition which Nikhil maintained through this new index list constructor. Elasticsearch has no problem with it, and I just never worried about it. 🤷🏻‍♂️

webbnh (approved):
Good by me.

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+            self.logger.error(f"Indices from metadata table not found {e!r}")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
+
+        index_keys = [key for key in index_map if root_index_name in key]
+        if len(index_keys) == 0:
+            self.logger.debug(
+                f"Found no indices matching the prefix {root_index_name}"
+                f"for a dataset {dataset!r}"
+            )
+            abort(HTTPStatus.NOT_FOUND, message="Found no matching indices")
```
portante, line 100: Do we really want to `abort` in this helper function for a non-error case?  Perhaps return an list of indices and let the API handle this?

portante (changes requested):

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
npalaska: Ah I missed it in a `namespace` class

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+        object and run_id as JSON CONTEXT so that the postprocess operations
+        can use it to identify the index to be searched from document index
+        metadata.
+        """
+        run_id = client_json["run_id"]
+
+        # Query the dataset using the given run id
+        dataset = Dataset.query(md5=run_id)
+        if not dataset:
+            self.logger.error(f"Dataset with Run ID {run_id!r} not found")
```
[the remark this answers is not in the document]
npalaska: Well this log message may not be helpful in the long run (like if we think this is not adding any value for debugging purpose and just taking up a log space then we can remove it later) but until then it might help us in debugging since currently our APIs are in active development mode. But I agree that this probably should't be an error message, info/debug would be better.

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
npalaska: But `get_index` takes the `root_index_name` as a function parameter to search for that indices, returning a list of all the indices would not be a desirable outcome. Probably `get_index` can return an empty string if it doesn't find any matching indices and then `NOT_FOUND` mechanics can be handled by an API itself rather here if that make sense. 

Also returning an `abort(NOT_FOUND)` seems okay to me since for the end client it wouldn't make much difference whether the server is using an `abort` or `response with 404` internally. 
If we find the matching indices but couldn't find any data for those indices thats the case for returning 200 with an empty json instead of 404.

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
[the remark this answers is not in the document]
dbutenhof: This code depends on having a `Dataset` resource in order to identify the appropriate indices; and as our server code becomes increasingly database-driven, yes, I think it is an error. Now, if there are in fact Elasticsearch documents and a tarball with that `run_id` but it's somehow not in the `Dataset` table, that's arguably an internal error rather than a `NOT_FOUND`, but under normal circumstances it ought to be reasonable to assume that if there's no `Dataset` with the specified `run_id`, then the `run_id` was erroneous and the resource is `NOT_FOUND`.

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
dbutenhof: These helpers are closely tied to the API, so I'm not too concerned structurally about the `abort`. In some cases, I created a new exception that would be handled by the base class to `abort`; arguably a bit cleaner, but then `abort` is really just a special exception handled by `Flask`.

And, yes, a comprehensive unit test set for methods like this would be good. We started down the road of using the Flask request framework for API testing, which makes that sort of test really awkward. I started breaking away from that with full true unit tests for the query builder and authorization checker; we should be doubling down on that sort of conversion to make sure our support methods are well tested.

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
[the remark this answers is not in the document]
npalaska: The erroneous `run_id` doesn't mean its an error on our side right so the log message can be debug right?

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
npalaska: Yeah and as Dave pointed out these helper functions are closely tied to the APIs so I guess handling the abort operation here would be okay

npalaska: Yeah we have an issue to reformat the unit tests to move away from request based testing https://github.com/distributed-system-analysis/pbench/issues/2408

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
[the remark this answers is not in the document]
dbutenhof: "It's complicated" ... it's either a user-error because they made up a bad `run_id`, or it's an internal error because they had a *valid* `run_id` (probably from a previous Pbench query), but we can't find the `Dataset` resource. I suppose we really don't want to be returning an internal error on random bad input data, though; so "not found" on a very rare and unlikely internal error path might be the better of the two alternatives.

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
```diff
@@ -91,6 +91,12 @@ def get_index(self, dataset: Dataset, root_index_name: AnyStr) -> AnyStr:
             self.logger.error(f"Indices from metadata table not found {e!r}")
             abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
+        if index_map is None:
+            self.logger.error(
+                f"server.index-map not found in Metadata for a dataset {dataset!r}"
```
dbutenhof, line 96: I'm not sure you really want the `repr` of the dataset in this case; but, yeah, this extra check is a good catch

lib/pbench/test/unit/server/query_apis/test_iteration_samples.py
```diff
@@ -679,3 +683,23 @@ def test_iteration_samples_with_scroll_id(
...
+        test = Dataset.attach(controller="node", name="test")
+
+        # When server index_map is None we expect 500
+        with pytest.raises(InternalServerError) as exc:
+            self.cls_obj.get_index(test, "result-data-sample")
+        assert exc.value.code == HTTPStatus.INTERNAL_SERVER_ERROR
+
+        Metadata.setvalue(
+            dataset=test,
+            key="server.index-map",
```
dbutenhof, line 697: `Dataset.INDEX_MAP`??

lib/pbench/test/unit/server/query_apis/test_iteration_samples.py
```diff
@@ -679,3 +683,23 @@ def test_iteration_samples_with_scroll_id(
             assert response.json == {
                 "results": [hit["_source"] for hit in response_payload["hits"]["hits"]]
             }
+
+    def test_exceptions_on_get_index(self, attach_dataset):
```
dbutenhof, line 687: Now you've taken the first step: how hard would it be to add some non-failure cases now?? 😈

dbutenhof (dismissed):
I like the start on unit testing; it'd be great to see more. Since I approved before and this is an improvement, I'll approve again... but given this encouraging halfway point, I find myself slightly reluctant. (No good deed goes unpunished...?)

lib/pbench/test/unit/server/query_apis/test_iteration_samples.py
npalaska: Ah, yeah if more changes are requested, I will fix this one as well.

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
portante: Not to create a problem, but it does not seem wise to spread API handling behavior outside of the API interface method.  Raising exceptions instead of directly using `abort()` in these cases gives us the ability to see and review how the API responses are handled in one place.

lib/pbench/test/unit/server/query_apis/test_iteration_samples.py
npalaska: Ah, yeah I could have added the non-failure cases here, but they should already be indirectly tested under other tests that check the entire query api.

lib/pbench/server/api/resources/query_apis/metadata_index/iteration_samples.py
npalaska: now fixed

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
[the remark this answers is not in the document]
portante: I don't see how we can know the reason why a `run_id` is not found.  So logging this as an `error` is just going to create noise.  If the dashboard detects the inconsistency, then using a tool like Sentry, or LogRocket, the server can be notified about what is happening with the client.  But here there isn't enough information to know if a `run_id` should be found but isn't, or if the `run_id` never existed, or no longer exists.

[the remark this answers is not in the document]
npalaska: Updated to `debug`

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
npalaska: okay so if there are no matching indices then this method will return an empty indices string and I moved the empty indices error handling to `assemble`.

dbutenhof (approved):

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+        object and run_id as JSON CONTEXT so that the postprocess operations
+        can use it to identify the index to be searched from document index
+        metadata.
+        """
+        run_id = client_json["run_id"]
+
+        # Query the dataset using the given run id
+        dataset = Dataset.query(md5=run_id)
+        if not dataset:
+            self.logger.debug(f"Dataset with Run ID {run_id!r} not found")
```
portante, line 64: This really should be not using f-strings to avoid always rendering the `repr` of `run_id` even if it is not emitted.

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
```diff
@@ -91,13 +91,13 @@ def get_index(self, dataset: Dataset, root_index_name: AnyStr) -> AnyStr:
...
-        if len(index_keys) == 0:
-            self.logger.debug(
-                f"Found no indices matching the prefix {root_index_name}"
-                f"for a dataset {dataset!r}"
+        if index_map is None:
+            self.logger.error(
+                f"server.index-map not found in Metadata for a dataset {dataset!r}"
             )
-            abort(HTTPStatus.NOT_FOUND, message="Found no matching indices")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
```
webbnh: Should we be triggering on all "false-y" values of `index_map`?  (I.e., could `index_map` possibly be an empty list, rather than none?...in which case we probably want to respond the same way....)

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
portante: This happens in many places, so perhaps a follow-up change to this PR?

npalaska: Sure I will address this in timeseries query api that I am working on

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
webbnh: OK, nevermind, maybe?  It looks like the caller is coded to catch this case and report it as not-found rather than internal error.

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
portante: Can you file a tracking issue to re-factor this code later to bubble `abort()` calls to the top-level API handling function?

webbnh (approved):
Looks good...I think....

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
webbnh: Nevermind...it looks like you guys discussed this above and I just missed it.

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
npalaska: tracked in https://github.com/distributed-system-analysis/pbench/issues/2529

lib/pbench/test/unit/server/query_apis/test_iteration_samples.py
npalaska: Fixed

portante (approved):
Posting approval since this commit has been through the ringer, but we need to add a unit test for not finding any result data sample documents, not just internal error handling.

PR #2414, opened 2021-08-11
Pbench put shim server

riya-17:
Fixes #2100

put-shim server will act as an interface between v0.69 pbench-agents and v0.72 pbench-server. It will receive the tarballs using ssh transfer method from the 0.69 agents and will forward them to the 0.72 servers using the Agent's `pbench-results-push` command.

**Note**: This change includes adding a dependency in the 0.69 Server on a 0.71+ Agent. This dependency will include using the `pbench-results-push` command of 0.71+ Agent.

More Details and Discussions about this PR are available on PR #2340. This PR deals with basing the put-shim server on branch b0.69 and structural changes according to that.

PR #2441, opened 2021-09-01
Add pytests to jenkins for b0.69

riya-17:
I added some of the requirements which were obvious and visible. Would love to get inputs on it.


jenkins/run-pytests
```diff
@@ -0,0 +1,27 @@
...
+    printf -- "FAILED: flake8 --verbose .\n" >&2
+fi
+printf -- "\n"
+pytest {posargs} ./server/lib/pbench/test/unit
+pytest_sts=${?}
+if [[ ${pytest_sts} -ne 0 ]]; then
+    printf -- "FAILED: pytest ...\n" >&2
+fi
+final_sts=$((black_sts + flake_sts + pytest_sts))
+exit ${final_sts}
```
dbutenhof, line 27: This appears to be a clean copy of what's on `main` with the exception of the 0.69 pytest path, so I won't offer several possible simplifications (for example, using a `flake="flake8 --verbose ."` env variable instead of repeating the string three times for the header, command execution, and failure message).

However, I will point out that the file ought to end with a newline. 😁

jenkins/run-pytests
```diff
@@ -0,0 +1,27 @@
...
+black_sts=${?}
+if [[ ${black_sts} -ne 0 ]]; then
+    printf -- "FAILED: black --check .\n" >&2
+fi
+printf -- "\nflake8 --verbose ...\n"
+flake8 --verbose .
+flake_sts=${?}
+if [[ ${flake_sts} -ne 0 ]]; then
+    printf -- "FAILED: flake8 --verbose .\n" >&2
+fi
```
webbnh, line 19: Black and Flake8 are already being run by [`run_unittests`](https://github.com/riya-17/pbench/blob/09630364d165f4fc52b2e6d1d56fcd59c49715f1/jenkins/run-unittests#L10) on `b0.69`.  If you would rather run them from here, I'm fine with that, but you should remove them from there.

webbnh (changes requested):

jenkins/run-pytests
riya-17: if I look according to `main` branch they are in run-pytests, so, I will add them there and remove from `run-unittests`

jenkins/run-pytests
```diff
@@ -10,18 +10,22 @@ black --check .
 black_sts=${?}
 if [[ ${black_sts} -ne 0 ]]; then
     printf -- "FAILED: black --check .\n" >&2
+    (( errors++ ))
```
webbnh, line 13: You don't need to count the errors like this (and, you're missing the `typeset` declaration, which you should have if you _do_ want to do it this way), and, besides, your code is never checking the result of the count.  🙂

Instead, your code is capturing and saving the exit statuses and then adding them all together at line 30, which accomplishes basically the same thing, and that's fine.

So, either remove lines 13 and 21, or put them to some use.

webbnh (changes requested):
Needs a touch of polish.  🙂

jenkins/run-pytests
riya-17: yes I missed the typeset, I will just remove the `error`

webbnh (dismissed):
GTG.

jenkins/run-pytests
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
portante, line 1: This file needs to be executable.  The latest test runs failed with, `/usr/bin/bash: ./jenkins/run-pytests: Permission denied`.

jenkins/run-pytests
```diff
@@ -0,0 +1,29 @@
...
+
+printf -- "\nflake8 --verbose ...\n"
+flake8 --verbose .
+flake_sts=${?}
+if [[ ${flake_sts} -ne 0 ]]; then
+    printf -- "FAILED: flake8 --verbose .\n" >&2
+fi
+printf -- "\n"
+
+pytest {posargs} ./server/lib/pbench/test/unit
```
portante, line 23: Instead of modifying 4 files, and repeating infrastructure code in `run-pytests`, why not just add these four lines to `run-unittests` after the `black` and `flake8` invocations?  That way you only modify one file for this commit.

portante (changes requested):
The `run-pytests` file need to be marked executable.  If there is a good reason to modify four files on the b0.69 branch, great.  If not, let's just modify one.

riya-17:
> If there is a good reason to modify four files on the b0.69 branch, great. If not, let's just modify one.

I just thought of keeping the structure intact. if this is how we wanna proceed I will make the modifications :)

jenkins/run-pytests
riya-17: This file is removed so maybe I can resolve this review

jenkins/run-unittests
```diff
@@ -24,6 +24,13 @@ if [[ ${flake_sts} -ne 0 ]]; then
 fi
 printf -- "\n"
 
+pytest {posargs} ./server/lib/pbench/test/unit
```
portante, line 27: For the Jenkins tests, we don't need `{posargs}`, and its present does not resolve because the syntax is for `tox` not `bash`.

portante (changes requested):

jenkins/run-unittests
riya-17: I looked on Jenkins and saw that it is failing because it cannot find any file for running pytests `ERROR: file not found: ./server/lib/pbench/test/unit` which will not be available until pbench-put-shim is merged.

dbutenhof: > I looked on Jenkins and saw that it is failing because it cannot find any file for running pytests `ERROR: file not found: ./server/lib/pbench/test/unit` which will not be available until pbench-put-shim is merged.

In that case, you might want to set this PR's upstream branch to your put-shim PR; that makes the dependency obvious, and will also allow it to actually run now.

Or, given that it's now a small block in a single file, maybe integrating pytest runs with Jenkins should just be part of the PR that adds the pytest code? It doesn't necessarily make sense to break *every* change into a separate PR. 😁

webbnh (changes requested):
These changes look OK to me, but apparently they don't work...I'm assuming it's because this PR introduces an infrastructure/framework change but doesn't actually add any content (to wit, tests), and so PyTest objects...which is not unreasonable.  However, I cannot approve a change which causes the tests to report failure (and, I don't think GitHub will permit it, either).

Given how simple this change is, I suggest merging it with #2414, which is adding the first test.  (We cannot really approve that PR, either, without this PR, since it cannot run the test which it is adding...and the way out of that Catch-22 is to combine the two PRs together.)

Issue #2246, opened 2021-04-26
Add malformed authorization unit test for PUT API


npalaska:
I suppose the point is that the name refers to the authorization header, not the token, while the test has a bad token but (from strict HTTP protocol point of view) the header is correctly formatted. Both cases are worth testing. Neither case is specific to the PUT API.

_Originally posted by @dbutenhof in https://github.com/distributed-system-analysis/pbench/pull/2232#discussion_r620564957_

----

We've got

- filename
  - missing
  - bad extension
- Auth token
  - missing
  - invalidated
- MD5 Hash
  - missing

It seems like there are at least two more cases that we could cover:
- a malformed Auth token
- a mismatched MD5 hash

_Originally posted by @webbnh in https://github.com/distributed-system-analysis/pbench/pull/2232#discussion_r619370630_

PR #2442, opened 2021-09-01
Add malformed authorization unit test

npalaska:
Fixes issue https://github.com/distributed-system-analysis/pbench/issues/2246 "Add malformed authorization unit test for PUT API"

### DIFF
--- a/lib/pbench/test/unit/server/conftest.py
+++ b/lib/pbench/test/unit/server/conftest.py
@@ -11,6 +11,7 @@ from stat import ST_MTIME
 from pbench.server.api import create_app, get_server_config
 from pbench.server.api.auth import Auth
 from pbench.server.database.database import Database
+from pbench.server.database.models.datasets import Dataset
 from pbench.server.database.models.template import Template
 from pbench.server.database.models.users import User
 from pbench.test.unit.server.headertypes import HeaderTypes
@@ -489,6 +490,34 @@ def pbench_token(client, server_config):
     data = response.json
     assert data["auth_token"]
     return data["auth_token"]
+
+
+@pytest.fixture()
+def attach_dataset(pbench_token, create_user):
+    """
+    Mock a Dataset attach call to return an object. We mock the Dataset.attach
+    method to avoid DB access here, however the user authentication mechanism
+    is not yet mocked so we have to look up User data.
+
+    Args:
+        pbench_token: create a "drb" user for testing
+        create_user: create a "test" user
+    """
+    datasets = {}
+    datasets["drb"] = Dataset(
+        owner="drb",  # Created by pbench_token fixture
+        controller="node",
+        name="drb",
+        access="private",
+    )
+    datasets["drb"].add()
+    datasets["test"] = Dataset(
+        owner="test",  # Created by create_user fixture
+        controller="node",
+        name="test",
+        access="private",
+    )
+    datasets["test"].add()
 
 
 @pytest.fixture(params=[header for header in HeaderTypes])
--- a/lib/pbench/test/unit/server/conftest.py
+++ b/lib/pbench/test/unit/server/conftest.py
@@ -503,21 +503,12 @@ def attach_dataset(pbench_token, create_user):
         pbench_token: create a "drb" user for testing
         create_user: create a "test" user
     """
-    datasets = {}
-    datasets["drb"] = Dataset(
-        owner="drb",  # Created by pbench_token fixture
-        controller="node",
-        name="drb",
-        access="private",
-    )
-    datasets["drb"].add()
-    datasets["test"] = Dataset(
-        owner="test",  # Created by create_user fixture
-        controller="node",
-        name="test",
-        access="private",
-    )
-    datasets["test"].add()
+    Dataset(
+        owner="drb", controller="node", name="drb", access="private"
+    ).add()  # Created by pbench_token fixture
+    Dataset(
+        owner="test", controller="node", name="test", access="private"
+    ).add()  # Created by create_user fixture
 
 
 @pytest.fixture(params=[header for header in HeaderTypes])
--- a/lib/pbench/test/unit/server/query_apis/commons.py
+++ b/lib/pbench/test/unit/server/query_apis/commons.py
@@ -83,7 +83,7 @@ class Commons:
         "malformed_token", ("malformed", "bear token" "Bearer malformed"),
     )
     def test_malformed_authorization_header(
-        self, client, server_config, malformed_token
+        self, client, server_config, malformed_token, attach_dataset
     ):
         response = client.post(
             server_config.rest_uri + self.pbench_endpoint,
--- a/lib/pbench/test/unit/server/query_apis/commons.py
+++ b/lib/pbench/test/unit/server/query_apis/commons.py
@@ -85,7 +85,7 @@ class Commons:
             headers={"Authorization": "malformed"},
             json=self.payload,
         )
-        assert response.status_code == HTTPStatus.UNAUTHORIZED
+        assert response.status_code == HTTPStatus.FORBIDDEN
 
     def test_non_accessible_user_data(self, client, server_config, pbench_token):
         """
--- a/lib/pbench/test/unit/server/query_apis/commons.py
+++ b/lib/pbench/test/unit/server/query_apis/commons.py
@@ -80,7 +80,7 @@ class Commons:
         return date_range
 
     @pytest.mark.parametrize(
-        "malformed_token", ("malformed", "Bearer token"),
+        "malformed_token", ("malformed", "bear token" "Bearer malformed"),
     )
     def test_malformed_authorization_header(
         self, client, server_config, malformed_token
--- a/lib/pbench/test/unit/server/query_apis/commons.py
+++ b/lib/pbench/test/unit/server/query_apis/commons.py
@@ -79,6 +79,14 @@ class Commons:
             date_range.append(f"{m.year:04}-{m.month:02}")
         return date_range
 
+    def test_malformed_authorization_header(self, client, server_config):
+        response = client.post(
+            server_config.rest_uri + self.pbench_endpoint,
+            headers={"Authorization": "malformed"},
+            json=self.payload,
+        )
+        assert response.status_code == HTTPStatus.UNAUTHORIZED
+
     def test_non_accessible_user_data(self, client, server_config, pbench_token):
         """
         Test behavior when Authorization header does not have access to other user's data
--- a/lib/pbench/test/unit/server/query_apis/commons.py
+++ b/lib/pbench/test/unit/server/query_apis/commons.py
@@ -79,10 +79,15 @@ class Commons:
             date_range.append(f"{m.year:04}-{m.month:02}")
         return date_range
 
-    def test_malformed_authorization_header(self, client, server_config):
+    @pytest.mark.parametrize(
+        "malformed_token", ("malformed", "Bearer token"),
+    )
+    def test_malformed_authorization_header(
+        self, client, server_config, malformed_token
+    ):
         response = client.post(
             server_config.rest_uri + self.pbench_endpoint,
-            headers={"Authorization": "malformed"},
+            headers={"Authorization": malformed_token},
             json=self.payload,
         )
         assert response.status_code == HTTPStatus.FORBIDDEN
--- a/lib/pbench/test/unit/server/query_apis/test_datasets_publish.py
+++ b/lib/pbench/test/unit/server/query_apis/test_datasets_publish.py
@@ -83,10 +83,15 @@ class TestDatasetsPublish(Commons):
             payload={"controller": "node", "name": "drb", "access": "public"},
         )
 
+    @pytest.mark.parametrize(
+        "malformed_token", ("malformed", "Bearer token"),
+    )
     def test_malformed_authorization_header(
-        self, client, server_config, attach_dataset
+        self, client, server_config, malformed_token, attach_dataset
     ):
-        super().test_malformed_authorization_header(client, server_config)
+        super().test_malformed_authorization_header(
+            client, server_config, malformed_token
+        )
 
     @pytest.mark.parametrize(
         "owner", ("drb", "test"),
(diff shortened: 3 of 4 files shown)


dbutenhof (changes requested):
1. Add a commit message explaining what this is about.
2. Why isn't this being done in the @Auth annotations? We want to validate authentication headers for *all* authenticated APIs, not just for the ElasticBase APIs.

npalaska:
> 1. Add a commit message explaining what this is about.

So if the Authorization header is invalid (e.g. not a Bearer token or logged-out token etc) we return None from @Auth annotations but if our API does not need authorization header to be strictly specified we could still proceed with the API operation even if the Authorization header was not Bearer.

However, I did some experiment this morning and realize we dont need to actually do this since our @Auth annotation already checks for the Bearer token and other APIs are enough protected when the current user is None. For example in Dataset_publish if one specifies an access as public, it wouldnt proceed with malformed authentication because the API operation would be `UPDATE`which needs current user to be not None

> 2. Why isn't this being done in the @Auth annotations? We want to validate authentication headers for _all_ authenticated APIs, not just for the ElasticBase APIs.

Yes this should already be done in @Auth annotations, I will update the PR to just fix the issue  #2246

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -837,7 +837,7 @@ def post(self):
         # Automatically authorize the operation only if the API schema has the
         # "user" key; otherwise we assume that authorization is unnecessary, or
         # that the API-specific subclass will take care of that in preprocess.
-        if "user" in self.schema:
+        if "user" in json_data:
```
dbutenhof, line 840: It's not correct to make the authorization test only if someone specifies `"user"` in the request payload. `"user"` must be a defined Schema keyword with type USER. If you follow through my current PRs, it's evolved to that point in #2417, although main is currently more simplistic.

If the caller omits `"user"` (where it's optional), we *still* need to run `_check_authorization` against the `"access"` value; e.g., we need to fail if they specify `{"access": "private"}` and there's no authorized user or it isn't ADMIN.

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -126,6 +126,13 @@ def test_missing_authorization_header(self, client, caplog, server_config):
         self.verify_logs(caplog)
 
     def test_malformed_authorization_header(self, client, caplog, server_config):
+        response = client.put(
```
dbutenhof, line 129: Yeah; you did a good job on factoring out the common `query_api` tests into `Commons`, and we really need to think about repeating that out to the non-Elasticsearch-based resource level, akin to `ApiBase`. But not today ...

lib/pbench/server/api/resources/query_apis/__init__.py
npalaska: Ah right we need to check the access as well.

lib/pbench/test/unit/server/query_apis/test_datasets_publish.py
```diff
@@ -83,6 +83,11 @@ def _setup(self, client):
             payload={"controller": "node", "name": "drb", "access": "public"},
         )
 
+    def test_malformed_authorization_header(
+        self, client, server_config, attach_dataset
+    ):
+        super().test_malformed_authorization_header(client, server_config)
```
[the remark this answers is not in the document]
npalaska: Yeah I spent some time thinking if this is right flow of logic we have in our dataset publish where we query the dataset and then the user and then we check the authorization or should it be first authorization and then everything else. But then  to check the authorization we need a user which can only be queried from the dataset as the owner. So I kept it as it is for now and make the appropriate changes in the dataset publish test. And thats why I have to override the malformed authorization test in publish since we need a dataset before we check the authentication.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -78,6 +78,14 @@ def date_range(self, start: AnyStr, end: AnyStr) -> list:
             date_range.append(f"{m.year:04}-{m.month:02}")
         return date_range
 
+    def test_malformed_authorization_header(self, client, server_config):
```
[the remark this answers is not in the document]
npalaska: Thats a good idea but probably I will have to do in both file `commons.py` and `test_datasets_publish.py` since that overrides the `test_malformed_authorization_header` method

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -78,6 +78,19 @@ def date_range(self, start: AnyStr, end: AnyStr) -> list:
             date_range.append(f"{m.year:04}-{m.month:02}")
         return date_range
 
+    @pytest.mark.parametrize(
+        "malformed_token", ("malformed", "Bearer token"),
```
dbutenhof, line 82: OK, that's interesting; yes, "Bearer token" should fail because it's not a decodable JWT token, and that's great, but it actually wasn't what I meant. I had deliberately misspelled "bear(er)" in order to see what happened in our authorization code. 😀

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -126,6 +126,13 @@ def test_missing_authorization_header(self, client, caplog, server_config):
         self.verify_logs(caplog)
 
     def test_malformed_authorization_header(self, client, caplog, server_config):
```
dbutenhof, line 128: Why aren't we parametrizing here instead of using two separate test cases?

dbutenhof (dismissed):

lib/pbench/test/unit/server/query_apis/test_datasets_publish.py
```diff
@@ -83,6 +83,16 @@ def _setup(self, client):
...
 
+    @pytest.mark.parametrize(
+        "malformed_token", ("malformed", "bear token" "Bearer malformed"),
+    )
+    def test_malformed_authorization_header(
+        self, client, server_config, malformed_token, attach_dataset
+    ):
+        super().test_malformed_authorization_header(
+            client, server_config, malformed_token
+        )
```
webbnh, line 94: This feels wrong.

How is invoking `TestDatasetsPublish.test_malformed_authorization_header()` different from invoking `Commons.test_malformed_authorization_header()`?  They look like they have the same parameterizations (although, I have to say, I don't know what happens when one parameterized test invokes another...does the "inner" test's parameterization get ignored?); the one invokes the other, so it seems like they have the same "bodies" (do they test the same stuff?); the only difference I note is that the "outer" one is including an additional fixture (but, I can't see _why_, since it doesn't explicitly reference it).

What happens if you simply omit the definition of `TestDatasetsPublish.test_malformed_authorization_header()` and let the class _inherit_ `Commons.test_malformed_authorization_header()`?  Does it not run, or does it encounter some error due to the missing fixture?

Given how much magic PyTest uses, I have a knee-jerk reluctance to having one test function invoke another.  If simple inheritance won't work, then perhaps it would be better to make `Commons.test_malformed_authorization_header()` into a "helper" function (instead of a "test") which `TestDatasetsPublish.test_malformed_authorization_header()` could then simply call (instead of having to make a `super()` reference).

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -125,10 +125,14 @@ def test_missing_authorization_header(self, client, caplog, server_config):
         assert response.status_code == HTTPStatus.UNAUTHORIZED
         self.verify_logs(caplog)
 
-    def test_malformed_authorization_header(self, client, caplog, server_config):
+    @pytest.mark.parametrize(
+        "malformed_token", ("malformed", "bear token" "Bearer malformed"),
+    )
+    def test_malformed_authorization_header(
+        self, client, caplog, server_config, malformed_token
+    ):
```
webbnh, line 133: This appears to be intended to override `Commons.test_malformed_authorization_header()`, presumably to prevent the base class test code from running...is that correct?  (It would be nice to have the signature match as much as possible, so `caplog` should be at the end.)

So, I'm wondering how much we are benefiting from having a base class implementation (especially since we have to repeat the parameterization in the derived classes).

I'm thinking that maybe a different approach might be better.  For instance, we could have a base class test shell (i.e., named `test_malformed_authorization_header()`), which the subclasses inherit and run, and it could invoke a class helper method which the subclasses can override.  In this way, all instances of the test use the same parameterization, and subclasses like this one here could replace the request if they want to; the only question in my mind is whether fixtures can be used in helper methods, or whether that magic only works on "test" methods.

webbnh (changes requested):
I'm really not crazy about this...perhaps we should have a meeting to discuss it?

The crux of the issue seems to be that in one subclass case we need to include an extra fixture (while in another subclass case, we want to run a completely different test in exactly the same way).  I'm wondering if there isn't a different approach or mechanism that we can use to accomplish this.

lib/pbench/test/unit/server/query_apis/test_datasets_publish.py
npalaska: Yes when one parameterized test invokes another parameterized test the inner test parametrization gets ignored (because its not called with python decorator). The inner function just acts as any other function as if it didn't have the decorators in the first place. It is certainly possible to make it a helper function in the commons but then it has to be invoked in all the tests where `Commons` is getting inherited.  



> What happens if you simply omit the definition of `TestDatasetsPublish.test_malformed_authorization_header()` and let the class _inherit_ `Commons.test_malformed_authorization_header()`? Does it not run, or does it encounter some error due to the missing fixture?

Yes it encounters error due to the missing fixture because its not being called with the pytest decorator that supplies that fixture value. So we have to explicitly provide the value of the fixture from the function that overriding it.

lib/pbench/test/unit/server/test_requests.py
npalaska: > So, I'm wondering how much we are benefiting from having a base class implementation (especially since we have to repeat the parameterization in the derived classes).

This overriding is happening only for the `Dataset_publish` because for Dataset_publish we need an additional `attach_dataset` fixture. 

Right now the flow of logic we have in our dataset publish is that we query the dataset first and then the user (Since we do not require user parameter from the client) and then we check the authorization. A user can only be queried from the dataset as the owner. And thats why I have to override the malformed authorization test in publish since we need a dataset before we check the authentication. 


> I'm thinking that maybe a different approach might be better. For instance, we could have a base class test shell (i.e., named `test_malformed_authorization_header()`), which the subclasses inherit and run, and it could invoke a class helper method which the subclasses can override. In this way, all instances of the test use the same parameterization, and subclasses like this one here could replace the request if they want to; the only question in my mind is whether fixtures can be used in helper methods, or whether that magic only works on "test" methods.

If the fixtures are defined on helper function and if we are calling the helper function then we have to explicitly provide value of the fixture to that helper function just like what we are doing with this test function. Function defined with `Pytest` decorator gets called only when the pytest invokes it and pytest only invoke function with `test` prefix. If we were to call this function from another function the pytest fixture wont automatically  get called and so we have to explicitly provide that value.

lib/pbench/test/unit/server/query_apis/test_datasets_publish.py
dbutenhof: The publish API authorizes against a Dataset rather than against a JSON "user" parameter, requiring the extra fixture.

Now, the Commons version could be progressively edited to pull in every fixture required by an of the subclasses, but that's ugly, too.

While I agree that using the "base class" (undecorated) as a helper is a bit weird, but it seems perfectly legitimate, and it's a pattern to allow the subclasses to specialize their behavior while still leveraging the common code. Within the framework of pytest I don't really have any better ideas.

lib/pbench/test/unit/server/test_requests.py
webbnh: Thank you for setting me straight:  this class is _not_ derived from `Commons`, and so this is non-issue.

lib/pbench/test/unit/server/query_apis/test_datasets_publish.py
webbnh: Dave, Nikhil, and I met to discuss this.  We concluded that the least of evils will be to use just the base class method, and to endow it with all the of the fixtures that any of the subclasses will require...at least until we can come up with something better.

dbutenhof (dismissed):

lib/pbench/test/unit/server/conftest.py
```diff
@@ -491,6 +492,34 @@ def pbench_token(client, server_config):
...
+        access="private",
+    )
+    datasets["drb"].add()
+    datasets["test"] = Dataset(
+        owner="test",  # Created by create_user fixture
+        controller="node",
+        name="test",
+        access="private",
+    )
+    datasets["test"].add()
```
dbutenhof, line 520: This is ugly, since the `dict` is now completely unused. In #2417 I've reduced this to:

```
    Dataset(owner="drb", controller="node", name="drb", access="private").add()
    Dataset(owner="test", controller="node", name="test", access="private").add()
```

I'm not suggesting you have to do this now; although, if you choose to, it'll look nicer and it'll simplify my rebase. 🤞🏻

dbutenhof (dismissed):

lib/pbench/test/unit/server/conftest.py
npalaska: yep, thats a good point. pushed another commit to fix it

lib/pbench/test/unit/server/conftest.py
```diff
@@ -491,6 +492,25 @@ def pbench_token(client, server_config):
...
+    Args:
+        pbench_token: create a "drb" user for testing
+        create_user: create a "test" user
+    """
+    Dataset(
+        owner="drb", controller="node", name="drb", access="private"
+    ).add()  # Created by pbench_token fixture
+    Dataset(
+        owner="test", controller="node", name="test", access="private"
+    ).add()  # Created by create_user fixture
```
dbutenhof, line 511: The comments are really redundant with the "Args" definitions, and removing them got these to a single line each. But, sure. Let's go for it.

dbutenhof (approved):

webbnh (approved):
Okie!

PR #2444, opened 2021-09-02
Check for the `ansible` RPM for the `pprof` tool

portante:
The `pprof` tool requires both `golang` and `ansible` RPMs, but the tool was only checking for `golang`.  We add an explicit check for both now.

In principle the `pbench-agent` should not have any explicit RPM dependencies for RPMs needed by tools.  But, the `pbench-run-bencmark` interface requires the `ansible` RPM, so we happen to satisfy the `pprof` requirement indirectly.

However, we don't want to have the `ansible` Python3 module explicitly listed in the `agent/requirements.txt` file since it is not a direct dependency for any Python3 code of the agent.  So for testing purposes we move the `ansible` module requirement to `test-requirements.txt`.

### DIFF
--- a/agent/tool-scripts/base-tool
+++ b/agent/tool-scripts/base-tool
@@ -591,9 +591,18 @@ install)
 		return ${rc}
 	}
 	case "${tool}" in
-	bpftrace|haproxy-ocp|iostat|kvmtrace|mpstat|oc|perf|pidstat|pprof|prometheus-metrics|sar|turbostat)
+	bpftrace|haproxy-ocp|iostat|kvmtrace|mpstat|oc|perf|pidstat|prometheus-metrics|sar|turbostat)
 		check_required_rpm ${tool_package_name} ${tool_package_ver}
 		rc=${?}
+		;;
+	pprof)
+		check_required_rpm ${tool_package_name} ${tool_package_ver}
+		rc=${?}
+		if [[ ${rc} == 0 ]]; then
+			# For pprof, we also need ansible installed.
+			check_required_rpm "ansible"
+			rc=${?}
+		fi
 		;;
 	systemtap)
 		# We need a number of systemtap and kernel packages for the
--- a/agent/tool-scripts/base-tool
+++ b/agent/tool-scripts/base-tool
@@ -152,6 +152,7 @@ sysfs)
 systemtap)
 	script=""
 	longoptions="${longoptions},script:"
+	tool_package_name="systemtap-client"
 	;;
 tcpdump)
 	interface=""
@@ -576,10 +577,10 @@ rc=0
 case "${mode}" in
 install)
 	function check_required_rpm {
-		_tpn="${1}"
-		_tpv="${2}"
-		_installed_rpm="$(require-rpm "${_tpn}" "${_tpv}")"
-		rc=${?}
+		local _tpn="${1}"
+		local _tpv="${2}"
+		local _installed_rpm="$(require-rpm "${_tpn}" "${_tpv}")"
+		local rc=${?}
 		if [[ ${rc} != 0 ]]; then
 			if [[ ! -z "${_tpv}" ]]; then
 				_tpv="-${_tpv}"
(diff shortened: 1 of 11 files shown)


agent/tool-scripts/base-tool
```diff
@@ -595,6 +595,15 @@ install)
...
 		;;
+	pprof)
+		check_required_rpm ${tool_package_name} ${tool_package_ver}
+		rc=${?}
+		if [[ ${rc} == 0 ]]; then
+            # For pprof, we also need ansible installed.
+		    check_required_rpm "ansible"
+			rc=${?}
+		fi
+		;;
```
webbnh, line 606: Since `pprof` is in the list at line 594, does that mean that it will execute that case _instead of_ the case at line 598 or _in addition to_ it?

Given that lines 595-597 are duplicated among the lines that you've added, I would be inclined to combine the cases -- if the `if` test checks both `${rc}` and `${tool}`, then you don't need a separate case and you can avoid the other duplication.

Also, line 602 appears to be indented with spaces when the surrounding lines are indented using hard-tabs.

webbnh (changes requested):
It looks like you have redundant handling for `pprof`.

ndokos (changes requested):
I think @webbnh's objections are valid.

agent/tool-scripts/base-tool
portante: Okay, the first and last items above I have addressed in a simple code-review commit update, but the middle item actually hit a bit larger bug in the `base-tool` script where some tools were not being properly checked.  So I made that an entirely separate commit.

portante (commented):
Okay, this looks a bit better.

agent/tool-scripts/base-tool
webbnh: Do you intend to squash the three commits together when you merge?  (I hope you do....)

If not, then please remove the `pprof` reference from line 594 in the same commit that you add the new case for it.

webbnh (approved):
Looks like a nice refactor!

agent/tool-scripts/base-tool
portante: > Do you intend to squash the three commits together when you merge? (I hope you do....)

The first two, yes, but I'd like to keep the second separate since it is a more radical change, a bug fix on its own, and refactoring that is not required to make the intended behavior work.

> 
> If not, then please remove the `pprof` reference from line 594 in the same commit that you add the new case for it.

I believe by squashing the first two commits together, this concern has been addressed.

agent/tool-scripts/base-tool
```diff
@@ -590,39 +591,33 @@ install)
...
-		if [[ ${rc} == 0 ]]; then
-			check_required_rpm "kernel-debuginfo" "$(uname -r)"
-			rc=${?}
+			case "${tool}" in
+			pprof)
+				# We also need ansible installed.
+				check_required_rpm "ansible"
+				rc=${?}
+				;;
+			systemtap)
```
dbutenhof, line 604: Another solution might have been to let `tool_package_name` become a list, where you'd check each package, rather than special casing on the `tool` name.

It bothers me just a little that this special casing for the tool is inside the `! -z "${tool_package_name}"` lest anyone see the case and try to leverage it to special-case another tool without adding `tool_package_name` for that tool. It's a minor concern, and I wouldn't want it to stand in the way here ... but doing the list would be a more general solution.

(Well, OK, not quite trivial as you need to know where to use the `$(uname -r)` ... since `require-rpm` would do the wrong thing if that's supplied where not necessary. ☹️ OK, so, forget that; it's better than it was, we're good.)

dbutenhof (approved):

agent/tool-scripts/base-tool
portante: Yes, I debated this as well.

I felt that with `systemtap` being the only tool that has so many packages required, implementing a general mechanism would not solve much.  By adding `pprof` to now have two (`golang` and `ansible`), i considered that might tip it in favor of a general solution.  But then I asked myself, "do I really want to continue this solution in `bash`?"

So for now, I figured we just have 2 special cases (`pprof` and `systemtap`) and an ugly bit of indentation instead.

webbnh (approved):
GTG

ndokos (dismissed):

webbnh (approved):
GTG

ndokos (approved):

dbutenhof (approved):

Issue #2370, opened 2021-07-14
Implement full semantics for defaulting "user" and "access" in dataset queries


dbutenhof:
I liked the idea of defaulting a missing `"user"` to the authorized user; however, in terms of API expressiveness, defaulting to "all users" is probably more useful. It also has the distinct advantage of making a missing `"user"` do the same thing regardless of whether you're logged in or not.

OK, so:

`{}` means all private and public data for all users (with the proviso that unless I'm ADMIN, I can only see my own private data).

`{"user": "drb"}` means all public and private data for user "drb". (Which will return only public data for "drb" unless the authenticated user is "drb" or ADMIN.)

`{"user": "drb", "access": "private"}` means only private data for user "drb". (Must be "drb" or ADMIN to return data.)

`{"user": "drb", "access": "public"}` means only public data for user "drb".

`{"access": "private"}` means all private data, regardless of user; an unauthenticated request would return nothing, while a non-ADMIN user would see only their own private data.

`{"access": "public"}` means all public data regardless of the owner. (Which, for a non-logged-in user, is indistinguishable from `{}`, but would be different for a logged-in user, and more differenter for a logged-in ADMIN user. That is, for an authenticated user, `{}` would return the user's own private data as well, while `{"access": "public"}` won't; for an ADMIN user, `{}` would return all data.))

_Originally posted by @dbutenhof in https://github.com/distributed-system-analysis/pbench/pull/2355#discussion_r660105897_

webbnh:
> `{"user": "drb"}` means all public and private data for user "drb".

With the proviso that, unless I'm ADMIN, the request won't return "drb"'s private data, unless the request comes from user "drb"....

> `{"access": "private"}` means all private data, regardless of user; this is legal only for an ADMIN user, and of dubious real world value.

I would dispute the "legal" part:  if we do a straightforward implementation, in the typical case (i.e., a logged-in, non-Admin user), this selector will return only the private data belonging to the requestor.  It would not be illegal, and it might be a common use case (e.g., it would free the dashboard from having to track and specify the current user...).  (Obviously, an Admin user will get _all_ private data back while a non-logged in user will get nothing.)

Similarly, `{"user": "drb", "access": "private"}` should be a legal request even if the requestor is not user "drb" and even if it's not an Admin...it's just that, in that case, the request won't get any data back.

dbutenhof:
[...]

PR #2452, opened 2021-09-13
More installation cleanups

### LINKED ISSUE #2421: Remove prep-shim from 0.72 server!
The current Pbench server dataset `PUT` upload isn't a full replacement for `pbench-server-prep-shim-002`, which for some reason I'd failed to completely internalize until now.

During the server jam session, we saw a logged failure due to `pbench-server-prep-shim-002` attempting to create a `Dataset` record owned by the imaginary `pbench` user. I initially assumed this was because we really shouldn't be running `prep-shim` anymore, and that it's interfering with `PUT`.

This isn't entirely true. The current `PUT` implementation is a direct replacement for `ssh`; it creates the `Dataset` record owned by the authorized client of `PUT`, but creates the files expected by `pbench-server-prep-shim-002` and depends on it to move the tarball and md5 file and set up the state link directories to get the dataset into the server pipeline.

This is a bug: `PUT` needs to completely replace `pbench-server-prep-shim-002` so that we can get rid of it and close the door on unowned `ssh` uploads entirely.

So,

- [x] `PUT` needs to *replace* the operations of `pbench-server-prep-shim-002` rather than depending on it to bring the new dataset into the server pipeline.
- [ ] We should completely delete the pbench-server-prep-shim-002 tool from 0.72: it's useless and dangerous
- [ ] The standard `prep-shim` intake directory should be removed from the install
- [ ] The host_info web API should be removed or redesigned, dropping the intake directory from the response, which is no longer relevant. (Maybe we don't completely delete it, however, as we do want a way to query the status of the server.)

dbutenhof:
This is a small incremental change towards issue #2422.

Fix the pbench-server service file to set up a proper `PATH` to find a pbench --user installed `gunicorn` binary

The previous `Environment` statement in the `pbench-server.service` file wasn't quite doing what we thought it was, because these are not executed in a context where the `$PYTHONPATH` is substituted. (It's also not necessary since #2423.) This changes `ExecStart` to run `bash` and construct a proper `PATH` that can resolve a locally installed `gunicorn` so the server will start. (In the longer run, we should be installing the `python-gunicorn` RPM instead of resolving this from `pip install`.)

Note that the proper `PATH` won't change after installation, however the cost of the `bash` is small and this is a common `systemd` service pattern. The alternative would be to generate `pbench-server.service` at installation time from a template, using information gleaned from the `pbench` account (basically `PATH=$(su pbench -c "echo ${PATH}:\$(python3 -m site --user-base)/bin")`)

### DIFF
--- a/server/lib/config/pbench-server.cfg.example
+++ b/server/lib/config/pbench-server.cfg.example
@@ -12,7 +12,8 @@ install-dir = /opt/pbench-server
 pbench-top-dir = /srv/pbench
 # Backup dir may be remote NFS mounted
 pbench-backup-dir = /mnt/pbench.archive.backup
-roles = pbench-maintenance, pbench-results, pbench-backup
+# Add role for sync'ing with satellites
+roles = pbench-maintenance, pbench-prep, pbench-results, pbench-backup, pbench-sync-satellites
 # pbench-server rest api variables
 bind_port = 8001
 rest_version = 1
--- a/server/lib/config/pbench-server.cfg.example
+++ b/server/lib/config/pbench-server.cfg.example
@@ -12,8 +12,7 @@ install-dir = /opt/pbench-server
 pbench-top-dir = /srv/pbench
 # Backup dir may be remote NFS mounted
 pbench-backup-dir = /mnt/pbench.archive.backup
-# Add role for sync'ing with satellites
-roles = pbench-maintenance, pbench-prep, pbench-results, pbench-backup, pbench-sync-satellites
+roles = pbench-maintenance, pbench-results, pbench-backup
 # pbench-server rest api variables
 bind_port = 8001
 rest_version = 1
--- a/server/lib/systemd/pbench-server.service
+++ b/server/lib/systemd/pbench-server.service
@@ -14,16 +14,11 @@ User = pbench
 Group = pbench
 RuntimeDirectory = pbench-server
 PIDFile = /run/pbench-server/gunicorn.pid
-ExecStart = /opt/pbench-server/bin/pbench-server
+ExecStart = /bin/bash -c 'PATH=${PATH}:$(python3 -m site --user-base)/bin /opt/pbench-server/bin/pbench-server'
 ExecStop = /bin/kill -s TERM $MAINPID
 Restart = always
 StartLimitInterval = 60
 StartLimitBurst = 10
-# this is required for newer libraries
-# set appropriately for your environment
-# or use systemctl edit pbench-server to put overrides
-# in another location
-Environment="PYTHONPATH=$PYTHONPATH:/opt/pbench-server/lib"
 Environment="_PBENCH_SERVER_CONFIG=/opt/pbench-server/lib/config/pbench-server.cfg"
 
 [Install]
--- a/server/lib/systemd/pbench-server.service
+++ b/server/lib/systemd/pbench-server.service
@@ -14,12 +14,11 @@ User = pbench
 Group = pbench
 RuntimeDirectory = pbench-server
 PIDFile = /run/pbench-server/gunicorn.pid
-ExecStart = /bin/bash -c 'PATH=${PATH}:$(python3 -m site --user-base)/bin /opt/pbench-server/bin/pbench-server'
+ExecStart = /opt/pbench-server/bin/pbench-server
 ExecStop = /bin/kill -s TERM $MAINPID
 Restart = always
 StartLimitInterval = 60
 StartLimitBurst = 10
-Environment="_PBENCH_SERVER_CONFIG=/opt/pbench-server/lib/config/pbench-server.cfg"
 
 [Install]
 WantedBy = multi-user.target
--- a/lib/pbench/cli/server/shell.py
+++ b/lib/pbench/cli/server/shell.py
@@ -2,13 +2,11 @@
 from logging import Logger
 import os
 from pathlib import Path
-from posix import environ
 import site
 import subprocess
 import sys
 
 from configparser import NoOptionError, NoSectionError
-from distutils.spawn import find_executable
 from sqlalchemy_utils import create_database, database_exists
 
 from pbench.common.exceptions import BadConfig, ConfigFileNotSpecified
@@ -26,16 +24,20 @@ def app():
     return create_app(server_config)
 
 
-def find_the_unicorn(logger : Logger) -> str:
+def find_the_unicorn(logger: Logger) -> str:
     local = Path(site.getuserbase()) / "bin"
     if (local / "gunicorn").exists():
         # Use a `pip install --user` version of gunicorn
         os.environ["PATH"] = str(local) + ":" + os.environ["PATH"]
-        logger.info("Found a local unicorn: augmenting server PATH to {}", os.environ["PATH"])
+        logger.info(
+            "Found a local unicorn: augmenting server PATH to {}", os.environ["PATH"]
+        )
 
 
 def main():
-    os.environ["_PBENCH_SERVER_CONFIG"] = "/opt/pbench-server/lib/config/pbench-server.cfg"
+    os.environ[
+        "_PBENCH_SERVER_CONFIG"
+    ] = "/opt/pbench-server/lib/config/pbench-server.cfg"
     try:
         server_config = get_server_config()
     except (ConfigFileNotSpecified, BadConfig) as e:
--- a/lib/pbench/cli/server/shell.py
+++ b/lib/pbench/cli/server/shell.py
@@ -1,9 +1,14 @@
 #!/bin/env python3
-
+from logging import Logger
+import os
+from pathlib import Path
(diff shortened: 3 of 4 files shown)


server/lib/config/pbench-server.cfg.example
```diff
@@ -12,8 +12,7 @@ install-dir = /opt/pbench-server
 pbench-top-dir = /srv/pbench
 # Backup dir may be remote NFS mounted
 pbench-backup-dir = /mnt/pbench.archive.backup
-# Add role for sync'ing with satellites
-roles = pbench-maintenance, pbench-prep, pbench-results, pbench-backup, pbench-sync-satellites
+roles = pbench-maintenance, pbench-results, pbench-backup
```
portante, line 15: Can you also mention in the commit message (and perhaps also reflect it in the PR description) that you are also removing the `pbench-sync-satellites` role from the example?

server/lib/systemd/pbench-server.service
```diff
@@ -14,16 +14,11 @@ User = pbench
 Group = pbench
 RuntimeDirectory = pbench-server
 PIDFile = /run/pbench-server/gunicorn.pid
-ExecStart = /opt/pbench-server/bin/pbench-server
+ExecStart = /bin/bash -c 'PATH=${PATH}:$(python3 -m site --user-base)/bin /opt/pbench-server/bin/pbench-server'
```
portante, line 17: Seems like this is a bit delicate and nuanced.  Systemd offers a method of setting `PATH`, perhaps doing something like https://stackoverflow.com/questions/35439123/systemd-environment-directive-to-set-path?

portante (changes requested):

server/lib/config/pbench-server.cfg.example
ndokos: But why are you removing the role? Is it not needed any longer?

server/lib/systemd/pbench-server.service
ndokos: @portante: that  can only deal with a constant value - there is no mechanism for substituting the previous value of PATH or of the value of the command. Right?

portante: Sure, but does the output of `python3 -m site --user-base` change each time it is invoked?  If not, why not record that once in an environment variable file at install time?

server/lib/config/pbench-server.cfg.example
dbutenhof: The satellite code is tied to the ssh transport mechanism, and pulls tarballs with no feasible way to track ownership. Peter and I have had some conversations about possible strategies to replace it in the future with something that's authenticated (e.g., replicating user databases, relying exclusively on a shared SSO, etc), but the latest thought is that it probably makes more sense to completely can the "feature". For our purposes, the DMZ pbench will fulfill the original need. If we don't get there we'll have to do something else, but it won't look much like the current satellite polling.

I would have left it alone, but since I was changing the example file's `roles` it made sense to remove the satellite role along with `pbench-prep`.

server/lib/systemd/pbench-server.service
dbutenhof: The overhead of firing up `bash` once on server start didn't seem like a big deal vs going to the trouble of modifying the `pbench-server.service` file on installation... although I did consider that.

No, the output won't change each time, but it is tied to the creation of the `pbench` user account and the host/account's default `PATH`. E.g., I could have assumed something like `/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/home/pbench/.local/bin`, but decided I'd rather not do that.

server/lib/config/pbench-server.cfg.example
dbutenhof: I've modified #2421 after some serious frustration this morning trying to push datasets to my reconstituted server in order to start experimenting with Elasticsearch queries for user/access (#2370). My datasets weren't getting unpacked or indexed. This is because I've managed to completely overlook a flaw in our `PUT` implementation. It only replaces the actual `ssh` operation, *not* `pbench-server-prep-shim-002`. That is, it still relies on `prep-shim` to process the uploaded tarball and MD5, move them to the archive tree, and create the proper state link directories to get the new dataset "into the pipeline".

We really need to close that door completely, so I think we should fix (or "complete") `PUT` so we actually *can* drop `pbench-server-prep-shim-002` entirely. Right now, however, we can't, so I'm going to drop this change.

TBD whether I drop the satellite change, too; we don't want it on 0.72 as it has the same inherent problems as `ssh` upload, compounded by the problems of assigning ownership to satellite datasets. However, if I'm not changing the default `roles` line anyway, it might not really make sense to remove it right now.

server/lib/systemd/pbench-server.service
webbnh: > Seems like this is a bit delicate and nuanced.

Nuanced, yes; delicate, not so much.  This approach seems adequate to me, and it needs to be evaluated in the context of the `pbench` user account, and I think this accomplishes that in an unequivocal way.  (I may steal this for the containerized testing stuff that I'm working on.)

webbnh (dismissed):

server/lib/systemd/pbench-server.service
portante: As mentioned here, https://serverfault.com/questions/413397/how-to-set-environment-variable-in-systemd-service, for Fedora, the recommended why to write systemd service files is to use `EnvironmentFile`.  On our path to productization is a stop along the way where we get our code to land in Fedora.

portante: Separate from that, what is the requirement that `PATH` needs to be updated for `/opt/pbench-server/bin/pbench-server` to work?

portante: Any can't find anywhere in the code on the server side where `PATH` needs to be updated.

portante: Also filed https://github.com/dbutenhof/pbench/pull/2.

portante (changes requested):
Let's adhere to Fedora rules and find out why we need `PATH` added.

server/lib/systemd/pbench-server.service
dbutenhof: > Any can't find anywhere in the code on the server side where `PATH` needs to be updated.

It's right in the commit message: "Fix the pbench-server service file to set up a proper PATH to find a pbench --user installed gunicorn binary".

When `gunicorn` is installed via our `requirements.txt` now, it goes into `/home/pbench/.local/bin` and, because that's not on the default `PATH`, the `pbench-server` service fails to start. (Yes, the wrapper ends up actually executing the `gunicorn` binary in a subprocess, which needs the `PATH`.)

There are multiple ways around this; this seemed the most consistent with our current approach. In the longer term, especially with Fedora migration, we know we need to dump `requirements.txt` and move to RPM dependencies, which would avoid the whole `PYTHONPATH` issue as well as `PATH`. As always, the question is how much we want to do how quickly.

portante: Why not set it in `./lib/pbench/cli/server/shell.py`?

portante (changes requested):
Looks much better!

Typically you can use the `env` keyword argument, like we do at: https://github.com/distributed-system-analysis/pbench/blob/35528da202e9c73bcd174b8f75fd518b26e4171e/lib/pbench/agent/tool_meister.py#L1591

dbutenhof:
> Looks much better!
> 
> Typically you can use the `env` keyword argument, like we do at:
> 
> https://github.com/distributed-system-analysis/pbench/blob/35528da202e9c73bcd174b8f75fd518b26e4171e/lib/pbench/agent/tool_meister.py#L1591

I considered that, possibly just passing `PATH` and `_PBENCH_SERVER_CONFIG`, but decided there wasn't really any evident advantage vs updating the default environment.

webbnh (approved):
GTG

ndokos (approved):

npalaska (approved):
This is all good but we would it be worth to add `python3-gunicorn` as an rpm dependency? instead of using the pip package

dbutenhof:
> This is all good but we would it be worth to add `python3-gunicorn` as an rpm dependency? instead of using the pip package

We're going to need to do that for everything, and maybe soon, but this is a point fix for a "hole" in my change to use `pip install --user` so let's keep these streams separate.

webbnh:
> This is all good but we would it be worth to add python3-gunicorn as an rpm dependency? instead of using the pip package

I'm working on this stuff separately.  However, my intention is to do _both_:  that is, my goal is to install Pbench from the RPMs and be able to run `setup.py` and have it report "all dependencies satisfied" (in fact, I intend to put a test into place which will catch aberrations from that).

riya-17 (approved):

dbutenhof:
@portante still has unresolved conversations

portante (approved):

PR #2455, opened 2021-09-15
remove 'test_' prefix from the fixture in conftest

### LINKED ISSUE #2454: Remove "test_" prefix from fixture in conftest.py
Remove "test_" prefix from the fixture in coftest.py file, it will prevent us from the confusion of assuming it as a test.

_Originally posted by @portante in https://github.com/distributed-system-analysis/pbench/pull/2414#discussion_r707382668_

riya-17:
Fixes #2454 

Remove "test_" prefix from the `test_logger` fixture in conftest.py file.

### DIFF
--- a/server/lib/pbench/test/conftest.py
+++ b/server/lib/pbench/test/conftest.py
@@ -24,7 +24,7 @@ def setup(request, pytestconfig):
 
 
 @pytest.fixture
-def test_logger(pytestconfig):
+def logger(pytestconfig):
     tmp = Path(pytestconfig.cache.get("TMP", None))
     logger = logging.getLogger("testing")
     handler = logging.FileHandler(tmp / "test.log")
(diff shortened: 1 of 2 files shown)


webbnh (approved):
GTG

server/lib/pbench/test/conftest.py
```diff
@@ -24,7 +24,7 @@ def setup(request, pytestconfig):
 
 
 @pytest.fixture
-def test_logger(pytestconfig):
+def logger(pytestconfig):
```
dbutenhof, line 27: We haven't been doing this, and I'm not really sure we want to... but looking at the isolated diffs I had to skip back to check on the return value, and a return type hint would be handy. On the other hand, I have no idea how we'd want to document the "type" of the fixture parameters, which are resolved/overloaded pytest fixtures themselves... which we define as functions but at this point are just symbols bound to the fixture return values.

Or maybe we just ignore the parameter types for fixtures but still try to add a return value type hint?

Never mind ... not something to attack in this PR. 😀 🙄

dbutenhof (approved):
Thanks for the cleanup; even if we're pretty sure the `test_` prefix doesn't matter here, it can make the code more confusing to read.

portante (approved):

ndokos (approved):

PR #2460, opened 2021-09-23
Implementation of user & access query semantics

### LINKED ISSUE #2370: Implement full semantics for defaulting "user" and "access" in dataset queries

dbutenhof:
This allows searching for combinations of "user" and "access" parameters in Elasticsearch-based queries with much more
flexibility than previously. For example, to find all "private" datasets owned by a specific user.

The docstring for the new query builder documents all of the odd cases. I've pared `convert_username` down to just checking for existence without making any attempts to "authorize" access, because it doesn't have the full context. It does, however, check whether the client is authenticated; an unauthenticated user is not allowed to reference any username: it will always be a `ConversionError` (400 `BAD_REQUEST`) rather than a permission error, avoiding any clue about the validity of a username. (Authenticated users, however, will get `ConversionError` for a bad username and `UnauthorizedAccess` for a valid username if the authenticated user doesn't have the requested access access rights.)

I added a pure unit test to examine the various query builder cases, along with real-world testing on a deployed VM.

Closes #2370


lib/pbench/server/api/resources/query_apis/controllers_list.py
```diff
@@ -134,6 +126,9 @@ def postprocess(self, es_json: JSON, context: CONTEXT) -> JSON:
         ]
         """
         controllers = []
+        if context.get("NODATA"):
+            return jsonify(controllers)
```
webbnh, line 130: Referencing `controllers` at line 130 seems slightly obfuscating -- why not just return `jsonify([])` (similar to what you did in `lib/pbench/server/api/resources/query_apis/datasets_detail.py` and the other modules)?

(And, if you buy into that, you can do it before set up stuff, like defining `controllers`....)

lib/pbench/test/unit/server/query_apis/test_query_builder.py
```diff
@@ -0,0 +1,240 @@
+from pbench.server.api.resources import Schema
+import pytest
+from pbench.server.api.auth import Auth
```
webbnh, line 3: These should be reordered.

lib/pbench/test/unit/server/query_apis/test_query_builder.py
```diff
@@ -0,0 +1,240 @@
...
+            }
+        }
+
+    def test_access_noauth_private(
+        self, elasticbase, server_config, current_user_none, user_ok
+    ):
+        """
+        Test the query builder when "private" access is specified by an
+        unauthenticated client. This is expected to throw an exception, which
+        will be treated as if Elasticsearch returned no hits.
```
webbnh, line 104: I think the "which clause" is out of scope for this function/module.

lib/pbench/test/unit/server/query_apis/test_query_builder.py
```diff
@@ -0,0 +1,240 @@
+from pbench.server.api.resources import Schema
+import pytest
+from pbench.server.api.auth import Auth
+from pbench.server.api.resources.query_apis import ElasticBase, JSON, NoSelectedDatasets
+from pbench.server.database.models.users import User
+
+
+class TestQueryBuilder:
```
webbnh, line 8: I think the full matrix has at least 27 cases:
- The requestor is unauthorized
- The requestor is authorized but not admin
- The requestor is an admin

for which the request might target
- user is self
- user is some other user
- user is omitted

and for which the request might specify
- access public
- access private
- access omitted

(And, I don't know whether we need/want to test the difference between access/user omitted vs specified as `None`/`null`; also, I don't know whether we want to test the difference between targeting a non-existent but non-null user, as well, which potentially adds between 1 and 9 additional cases.)

I think you have 9 of the cases so far; I wonder if you can cover the rest of the matrix by parameterizing the existing scenarios.

webbnh (commented):
This looks like good stuff!  Of course, though, I have comments.

In particular, I would propose some simplifications to the ES query, and I would propose a simpler (if less generic) approach to constructing it.

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -77,45 +96,210 @@ def __init__(
...
+                    )
+            query["bool"] = {"filter": must}
         else:
-            term = {"authorization.access": Dataset.PUBLIC_ACCESS}
-        return term
+            # {"access": "xxx", "user": "xxx"}
+            must.append({"term": {"authorization.access": access}})
+            must.append({"term": {"authorization.owner": user}})
+            query["bool"] = {"filter": must}
+        return query
```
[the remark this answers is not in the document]
dbutenhof: Looking through it I don't see any obvious holes, but I also think this is one of those cases I prefer a not quite so extremely consolidated logic chain so it's easier to read the flow. But I had already intended to go back and think about whether it could be optimized; I'll feed this into the grinder.

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -258,18 +442,29 @@ def _call(self, method: Callable, json_data: JSON):
...
             abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
         try:
-            # perform the Elasticsearch query
-            es_response = method(url, **es_request["kwargs"])
-            self.logger.debug(
-                "ES query response {}:{}", es_response.reason, es_response.status_code
-            )
-            es_response.raise_for_status()
+            if not context.get("NODATA", False):
```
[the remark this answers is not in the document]
dbutenhof: My mood changes, but I often dislike relying on implicit false-y logic and prefer to make it explicit. Putting the `False` here was admittedly a momentary whim, but it wasn't entirely random.

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -258,18 +442,29 @@ def _call(self, method: Callable, json_data: JSON):
...
+                # perform the Elasticsearch query
+                es_response = method(url, **es_request["kwargs"])
```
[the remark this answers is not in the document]
dbutenhof: Well, it's a hole in the Flask-level testing. I covered all the cases in the `test_query_builder.py` unit test, but this is proof that's not quite enough since it can overflow into here. Ugh.

I'll probably close this by pulling out the `es_response.json()` here into a variable which is otherwise `None` (or maybe `{}`), which the logic outside the `try` block can use safely.

And I suppose I need to add a Flask test case (as messy as they are) to cover the `NODATA` path. ☹️

lib/pbench/server/api/resources/query_apis/controllers_list.py
dbutenhof: I nearly did that, but in this case I had a clean point to return with a known local `{}` already and decided to use it. I'm not quite sure why I went with that knowing you'd object ... maybe I was just feeling contrary that day. 😆

lib/pbench/test/unit/server/query_apis/test_query_builder.py
dbutenhof: I suspect I let vscode pull in `Schema` and forgot to go back and check the order. Ah, well. It seems to usually get them wrong; but being able to just `ctrl-.` instead of jumping to the top and typing out the import is *so* nice.

lib/pbench/test/unit/server/query_apis/test_query_builder.py
dbutenhof: Yes, I debated full black-box testing, as you suggest here, vs my semi-white-box driven by knowledge of which cases actually have differing logic.

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -77,45 +96,210 @@ def __init__(
...
+                TODO: Need to decide if this is a security issue. Do we want to
+                hide the existence of other users here? E.g., should the
+                UNAUTHORIZED case return NO DATA? And if so, should there be a
+                difference between UNAUTHORIZED and AUTHORIZED for some other
+                user? (And, NOTE, specifying a "user" parameter, either without
+                authentication or with non-ADMIN authentication where the user
+                parameter doesn't match the specified "user" parameter, will
+                result in an UnverifiedUser exception during schema validation,
+                so we won't reach this point. That is, changes must be made if
+                we want to ALLOW these requests.)
```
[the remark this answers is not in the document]
webbnh: > Is a query for `wscales/private` a 400 or a 403?

I'm inclined to muddy the waters even further:

- if the query returns (or _would return_) hits, then it's neither a `400` nor a `403` (it's a `200` or similar), so we (should) only have to visit this issue for queries which return no hits;
- if the requestor is unauthorized, then the result should be same regardless of whether the user (or data) exists or not (but, I might lean toward a `401` rather than a `400` or a `403`);
- if the requestor is an Admin, then the result should be `404`, I would think (if the user or data doesn't exist);
- otherwise, I'm not sure whether the answer should be `403` (for security) or `404` ('cuz we trust the user).

Or, we could just punt:  we could always return `200`, and whether the response contains hits or not depends on whether the requestor had access -- a malicious attack will always see "success with no data", and there is no inference to be drawn (of course, this isn't particularly friendly to authenticated users, but it's safe and consistent).

> my inclination is to get `verify_user` out of the security business [...] leaving security concerns entirely to `ApiBase._check_authorization()`

I support that:  I'm a fan of separating mechanism from policy!

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -77,45 +96,210 @@ def __init__(
...
+                                    "gte": "2021-09-01",
+                                    "lte": "2021-10-28"
+                                    }
+                                }
+                            }
+                        ]
+                    }
+                }
+            }
+        },
```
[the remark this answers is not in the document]
webbnh: > I wasn't able to find any way to construct the `OR` logic without `"constant_score"`

There's nothing magical (in that sense) about `"constant_score"` -- what got you your `OR` logic was using a nested compound query -- e.g., putting the `"bool"` inside the `"constant_score"` -- but you could have done the same thing nesting the `"bool"` inside another `"bool"`

> "dis_max" (a really ugly term that gives absolutely no clue what it does)

Yes, I don't care for the term either, but, if you're scoring, it allows you to pick the maximum score from among a disjointed set of matches...and if you're not scoring, it provides a "disjunction" operation -- i.e., an "OR".  😀

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -77,45 +96,210 @@ def __init__(
...
+                Elasticsearch (which would require an impossible combination
+                of terms resulting in no output but no errors). The _call()
+                method will handle this by calling postprocess() with an empty
+                query response that must trigger the subclass to act as if
+                there were no hits or aggregations.
 
         Returns:
-            An assembled Elasticsearch query term
+            An assembled Elasticsearch "query" mode that includes the necessary
+            user/access terms.
```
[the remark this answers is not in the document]
webbnh: The Elasticsearch documentation uses the term ["clause"](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl.html), but it uses it to refer to the components which make up a "query" (e.g., `"term"`, `"range"`, `"bool"`, `"dis_max"`, et al.).

It refers to `"query"` as a ["parameter"](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations.html#change-agg-scope) to the ["search query" request](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-your-data.html#search-your-data) (interesting the double duty that the term "query" carries...I suppose that comes from the collision between "querying" ES as a database and an HTTP request "query string").

lib/pbench/server/api/resources/query_apis/__init__.py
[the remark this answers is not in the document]
webbnh: > I often dislike relying on implicit false-y logic and prefer to make it explicit.

I respect and support that.  But, if that's the right approach here, then it seems to me that it would be the right approach in all the other locations which followed the other whim.  :wink:

lib/pbench/server/api/resources/query_apis/__init__.py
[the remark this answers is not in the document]
dbutenhof: The real question is whether we want to diagnose the client's right to ask the question. Sure, we can completely skip the CRUD model and just treat bad requests as no-op successes. Maybe that's not even entirely unreasonable for read-only operations like the Elasticsearch queries, although I'm not entirely happy with that.

We definitely don't want to distinguish between "that user doesn't exist" and "you can't ask about that user", so the 400 vs 403 discussion is important.

1. If I ask for Fred's public data, should I be able to distinguish between "Fred has no public data" and "there is no Fred"? (Maybe not.)
2. Same question if I ask for Fred's private data ... but I'm not entirely sure the answer is the same.

For example, maybe bad "public" data requests always return 200 with an empty payload, but bad "private" data requests always return 403? That is, one always has the *right* to ask for public data, though we don't want to explicitly reveal a bad username; but one does *not* have the right to ask for private data unless as admin or for your own data.

And 401 might be appropriate for an unauthenticated request that might have been OK if authenticated; e.g., if I ask for Dave's private data ... no-go unauthorized, but fine if authorized as Dave. It's not appropriate if authentication is given, but that authentication doesn't authorize the request (e.g., Dave asking for Webb's private data.) And I'm not sure I want to bother decoding 401 from 403 cases: I'm not sure it really has that much value in our context.

And I really am thinking about paring `verify_user` down to just basic validation of the `username` value, regardless of authorization. In fact, the `Auth` method should probably just be removed, and I could have `convert_username` simply do the query and either fail or return the `user.id`, leaving authorization entirely to `_check_authorization` where both the user and access can be considered. I'm going to do a commit of review comments first, but then I may try hacking around with this to see how much pain it'll cause.

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -11,23 +11,42 @@
...
+    generates an appropriate empty response (e.g., [] or {}).
+    """
+
+    def __init__(self, user: str, access: str):
+        self.user = user
+        self.access = access
+
+    def __str__(self) -> str:
+        user = f"user {self.user!r}" if self.user else "unauthorized client"
+        return f"Query from {user} for access {self.access!r} cannot produce results"
```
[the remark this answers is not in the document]
webbnh: I certainly concur that the onus is on the Pbench server to avoid returning to the user items which s/he is not entitled to access, but I would effect that by ensuring that ES doesn't return such items to the server in the first place, by constructing the ES query in such a way as to prevent that.  And, I concur that, for the cases where we can predict a priori that the query will return no data (for any of several conceivable reasons), we can skip issuing the query to ES by raising this exception instead.

Thus, raising the exception indicates that, had we issued the query to ES, there would be no results to return to the user; it indicates that the query need not and should not be issued and, indeed, that the function has therefore declined to construct the query payload.

So, from the user-visibility perspective, the request "would not" produce results; from the server perspective, the query "must not" produce results and therefore "shall not", thanks to the exception.

Anyway, I think we've probably now given this more attention than it deserved, and I'm sure that it'll be fine, so I'm going to go ahead and mark this resolved.

lib/pbench/server/api/resources/query_apis/__init__.py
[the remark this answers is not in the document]
webbnh: > If I ask for Fred's public data, should I be able to distinguish between "Fred has no public data" and "there is no Fred"? (Maybe not.)

If you ask for "Fred's public data" and the answer is "there is no public data for Fred", you cannot tell whether that's because there is no data or no Fred -- either way, the answer is correct...but, if we return something other than `200` or `404`, you'll be able to draw some sort of inference.  I don't think we want unauthorized users to be able to draw that inference; it would be nice if admin users could; and, for other users, personally I would support being able to help the user detect things like typos in usernames....

> Same question if I ask for Fred's private data ... but I'm not entirely sure the answer is the same.

Once again, if you get the answer "there is no private data for Fred", then you cannot draw any inference about the existence of Fred, unless we return something other than `200` or `404` (although in this case I think returning `401` might be an appropriate alternative if the requestor is unauthorized).  But, since we're talking about accessing something which is nominally "private", returning `403` to an authorized user might be appropriate; obviously, an admin would get `404` for an existent user and perhaps the same status with a different message for a non-existent one.

> I'm not sure I want to bother decoding 401 from 403 cases

I'm hoping that it will be obvious from context which one should be returned; I concur that it's not worth a lot of effort otherwise.  In fact, I'm at peace with returning `200`/`404` for all cases....

> I really am thinking about paring verify_user down to just basic validation of the username value, regardless of authorization.

Go you!  💪

lib/pbench/test/unit/server/query_apis/test_datasets_list.py
```diff
@@ -53,12 +44,14 @@ def test_query(
         """
         payload = {
             "user": user,
+            "access": "private",
             "controller": "dbutenho.csb",
             "start": "2020-08",
             "end": "2020-10",
         }
         if user == "no_user":
             payload.pop("user", None)
```
webbnh, line 53: In `lib/pbench/test/unit/server/query_apis/test_datasets_detail.py`, you've switched from `pop()` to `del`...did you want to do that here?

lib/pbench/test/unit/server/query_apis/test_query_builder.py
webbnh: 🎉 ❗

lib/pbench/test/unit/server/query_apis/test_query_builder.py
```diff
@@ -10,114 +12,110 @@ class TestQueryBuilder:
...
         assert query == {
             "bool": {
                 "filter": [
                     {"term": {"icecream": "vanilla"}},
+                    {"term": {"authorization.owner": id}},
                     {"term": {"authorization.access": "public"}},
-                    {"term": {"authorization.owner": "drb"}},
                 ]
             }
         }
```
webbnh: I wonder if it would be worth sorting the clauses or otherwise making this test resilient against small changes in the CUT.

lib/pbench/test/unit/server/query_apis/test_query_builder.py
```diff
@@ -10,114 +12,110 @@ class TestQueryBuilder:
...
-    ):
+    def test_access_noauth_private(self, elasticbase, server_config, current_user_none):
         """
         Test the query builder when "private" access is specified by an
-        unauthenticated client. This is expected to throw an exception, which
-        will be treated as if Elasticsearch returned no hits.
+        unauthenticated client. This is an "impossible" case as authorization
+        checks will stop the request and return UNAUTHORIZED before we get
+        here. We cover the code path anyway, looking for the AssemblyError
+        exception.
```
webbnh: I don't think that this comment should be describing code elsewhere which it isn't testing.  That is, the code here doesn't care that the setup yields an "impossible" case -- we're gonna do the set up and check that it produces the result that we want anyway.  I think you should basically delete the middle sentence and the first half of the last sentence.

lib/pbench/test/unit/server/query_apis/test_query_builder.py
```diff
@@ -135,14 +133,13 @@ def test_access_noauth_public(
             }
         }
 
-    def test_access_user_private(
-        self, elasticbase, server_config, current_user_drb, user_ok
-    ):
+    def test_access_user_private(self, elasticbase, server_config, current_user_drb):
```
webbnh, line 136: Where is the `test_access_user_public()` test?

webbnh (changes requested):
This is great stuff!  Most of my comments are nits and typos and stuff; however, I do have a couple of more substantive items.

- Need to settle on `BAD_REQUEST` vs. `NOT_FOUND` for queries against nonexistent users.
- There are a couple of spots where it looks like the code is preventing authorized users from requesting data for "all users" or "all public data".
- I _think_ there's a test missing (for authorized data requesting all public data).


(And, apparently, it needs a rebase.)

lib/pbench/server/api/resources/__init__.py
```diff
@@ -37,11 +37,19 @@ class UnauthorizedAccess(Exception):
...
     """
 
-    def __init__(self, user: str, operation: "API_OPERATION", owner: str, access: str):
+    def __init__(
+        self,
+        user: str,
+        operation: "API_OPERATION",
+        owner: str,
+        access: str,
+        status: int = HTTPStatus.FORBIDDEN,
```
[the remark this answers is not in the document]
dbutenhof: Yeah, it wasn't until I scrolled down to the references that I really noticed I'd named it differently in the other exceptions, and I didn't bother going back. You're right, of course; I should have.

lib/pbench/server/api/resources/__init__.py
```diff
@@ -229,31 +227,44 @@ def convert_username(value: Union[str, None], _) -> Union[str, None]:
...
+    here allows rejecting any "user" parameter reference by an unauthenticated
+    user.
+
+    An authenticated user *can* determine whether a username is valid, and will
+    be able to ask for another user's public data with {"user": "name",
+    "access": "public"} (or simply "{"user": "name"}, if the authenticated user
+    doesn't have rights to "name"'s private data). An authenticated user will
+    be given a validation error (BAD_REQUEST/400) if the specified username
+    isn't valid and a permission error (FORBIDDEN) if asking for another user's
+    private data (without ADMIN role).
```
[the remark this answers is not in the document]
dbutenhof: That's an interesting question. My interpretation has always been that `NOT_FOUND` means that a resource was properly identified but isn't there. So for `/api/v1/user/<name>` I'd think `NOT_FOUND` would be appropriate; that's a lot less clear when the username is just one of several parameters identifying a resource. We weren't able to even *look* for the resource, because one of your parameters was "unhelpful". That seems more like a `BAD_REQUEST` than a `NOT_FOUND` to me.

lib/pbench/server/api/resources/__init__.py
```diff
@@ -229,31 +227,44 @@ def convert_username(value: Union[str, None], _) -> Union[str, None]:
...
+        if not Auth.token_auth.current_user():
+            raise UnverifiedUser(value)
+        try:
+            user = User.query(username=value)
+        except Exception:
+            pass
+        else:
+            if user:
+                return str(user.id)
+    raise ConversionError(value, "username")
```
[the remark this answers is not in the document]
dbutenhof: I went back and forth on that a number of times. I really don't like the repeated exceptions ... but, if nothing else, your comment points out an interesting advantage in that raising from the `except` avoids losing the original `query` problem, and I do like that.

lib/pbench/server/api/resources/__init__.py
```diff
@@ -611,17 +622,18 @@ def __init__(
...
-        If "user" is None, then the request is unauthenticated. READ operations
-        may be allowed, UPDATE and DELETE operations will not be allowed.
+        If there is no current authenticated API user, only READ operations on
+        public data will be allowed.
 
-        If "access" is None, READ will assume we're looking for access only to
-        public datasets.
+        If the specified "user" context of the operation is None, we default to
+        the authenticated caller, if any; Pbench query operations will do the
+        same.
```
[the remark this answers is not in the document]
dbutenhof: Right now I've got no idea what that was intended to mean. Possibly it was a reference to the query builder, which I guess makes the most sense in this context; at best it's pretty muddy.

lib/pbench/server/api/resources/__init__.py
```diff
@@ -649,30 +661,38 @@ def _check_authorization(self, user: str, access: str):
...
             "Authorizing {} access for {} to user {} with access {}",
             self.role,
             authorized_user,
             user,
             access,
         )
+
+        # Default user to the authenticated user, if not specified
+        if authorized_user and not user:
+            user = authorized_user.username
```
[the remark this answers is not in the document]
dbutenhof: This doesn't affect what the query builder will see, only authorization. If `"user"` is missing in the query builder, and you have admin or you're looking for public data, we'll just build the `authorization.access` term. But if you're looking for private data (without a username) and don't have admin, it'll build "user == me and access == private"; so the check here is appropriate and avoids dealing with a missing `"user"` later.

Probably needs a better explanation, though. ☹️ Basically, if you don't specify user, either you're admin and we validate your undisputed right to the entire multiverse or we validate your right to only your own data because we decided to constrain the query rather than failing.

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -77,45 +106,145 @@ def __init__(
...
+                ADMIN: all owner:drb AND access:private
+                AUTHORIZED as drb: all owner:drb AND access:private
+                UNAUTHORIZED (or non-drb): Not handled (permission error)
+
+            {"user": "drb", "access": "public"}: public drb
+                All datasets owned by "drb" with "public" access
+
+                ADMIN: all owner:drb AND access:public
+                AUTHORIZED as drb: all owner:drb AND access:public
+                UNAUTHORIZED (or non-drb): all owner:drb AND access:public
```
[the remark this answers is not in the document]
dbutenhof: And then there's the fact that an unauthenticated client can't probe a username at all, so we can't actually get here in the first place. I was trying to document what the method does in isolation, but I'm increasingly less certain that's the correct strategy. (And I already referenced "permission error" above.)

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -77,45 +106,145 @@ def __init__(
...
+            if access == Dataset.PRIVATE_ACCESS:
+                raise AssemblyError(
+                    f"Internal error: can't generate query for user {user}, access {access}"
+                )
+            filter.append({"term": {"authorization.access": Dataset.PUBLIC_ACCESS}})
+            self.logger.debug("QUERY: not self public: {}", filter)
+        elif access:
+            filter.append({"term": {"authorization.access": access}})
+            if not user and not is_admin:
+                filter.append({"term": {"authorization.owner": authorized_id}})
```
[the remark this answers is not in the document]
dbutenhof: I think you may be partly correct; I'll have to investigate this further. I thought I had a test case: either it's missing or it's broken. Yes, an authorized user asking for `{"access": "public"}` should see everyone's public data.

However, an authorized user (or admin) asking for another user's public data (or not constraining user) *does* need an access filter, because Elasticsearch will happily return private data without it.

lib/pbench/test/unit/server/query_apis/test_datasets_list.py
dbutenhof: I didn't look here, since I didn't touch it. In fact, I thought I'd *restored* a `del` in the other case which I'd temporarily removed in my travels ... I forgot the original had for some reason used a `pop`. Hmm.

dbutenhof (commented):
Some light conversational responses, but I didn't get into deep code analysis today.

lib/pbench/test/unit/server/test_schema.py
```diff
@@ -44,7 +44,7 @@ def test_exceptions(self, create_user):
         s = SchemaError()
         assert str(s) == "Generic schema validation error"
         u = UnverifiedUser("you")
-        assert str(u) == "User you can not be verified"
+        assert str(u) == "Caller is unable to verify username you"
```
riya-17, line 46: is this the right assertion message? 'you' seems a bit weired

lib/pbench/server/api/resources/query_apis/__init__.py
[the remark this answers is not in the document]
webbnh: If control gets to 226, then we'll add the `"access"` filter that we need.  But, we only need to add `{"authorization.owner": authorized_id}` if we're requesting `{"access": "private"}` (f we didn't add it at line 199) and if we're not admin.

lib/pbench/server/api/auth.py
```diff
@@ -150,20 +90,22 @@ def verify_auth(auth_token):
         except jwt.ExpiredSignatureError:
             try:
                 ActiveTokens.delete(auth_token)
-            except Exception:
+            except Exception as e:
                 Auth.logger.error(
-                    "User attempted Pbench expired token but we could not delete the expired auth token from the database. token: '{}'",
+                    "User passed expired token but we could not delete the token from the database. token: {!r}: {}",
                     auth_token,
+                    str(e),
```
portante, line 97: Do you really need to explicitly convert it to a string?

lib/pbench/server/api/auth.py
```diff
@@ -150,20 +90,22 @@ def verify_auth(auth_token):
...
         except jwt.InvalidTokenError:
-            Auth.logger.warning("User attempted invalid Pbench token '{}'", auth_token)
-        except Exception:
+            Auth.logger.warning("User passed invalid token {!r}", auth_token)
+        except Exception as e:
             Auth.logger.exception(
-                "Exception occurred while verifying the auth token '{}'", auth_token
+                "Unexpected exception occurred while verifying the auth token {!r}: {}",
+                auth_token,
+                str(e),
```
portante, line 109: Do you really need to explicitly convert it to a string?

lib/pbench/server/api/auth.py
```diff
@@ -150,20 +90,22 @@ def verify_auth(auth_token):
...
                 Auth.logger.error(
-                    "User attempted Pbench expired token but we could not delete the expired auth token from the database. token: '{}'",
+                    "User passed expired token but we could not delete the token from the database. token: {!r}: {}",
                     auth_token,
+                    str(e),
                 )
-                return None
             Auth.logger.warning(
-                "User attempted Pbench expired token '{}', Token deleted from the database and no longer tracked",
+                "User passed expired token {!r}; token deleted from the database and no longer tracked",
```
portante, line 100: I realize this is not a change you made, but ... given this is a log message for consumption by the people administering the pbench server, is this really something to warn them about?  Seems like this is going to occur as a normal part of the user/server interactions.

lib/pbench/server/api/auth.py
```diff
@@ -150,20 +90,22 @@ def verify_auth(auth_token):
...
                 auth_token,
             )
         except jwt.InvalidTokenError:
-            Auth.logger.warning("User attempted invalid Pbench token '{}'", auth_token)
-        except Exception:
+            Auth.logger.warning("User passed invalid token {!r}", auth_token)
```
portante, line 104: Similar question on having this be a "warning"...

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -77,45 +106,155 @@ def __init__(
...
+                }
+            )
+            self.logger.debug("QUERY: {{}} default: {}", filter)
+            return {"bool": {"filter": filter}}
         else:
-            term = {"authorization.access": Dataset.PUBLIC_ACCESS}
-        return term
+            # Either "user" was specified and is already added to the filter,
+            # or client is ADMIN and no access terms are required.
+            pass
```
portante, line 249: The boolean logic above seems "okay" but it is hard to prove we have all the cases.  Anyway we can record how many times we get to `pass` to help understand the behavior of the queries?

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -77,45 +106,155 @@ def __init__(
...
+                    "dis_max": {
+                        "queries": [
+                            {"term": {"authorization.owner": authorized_id}},
+                            {"term": {"authorization.access": Dataset.PUBLIC_ACCESS}},
+                        ]
+                    }
+                }
+            )
+            self.logger.debug("QUERY: {{}} default: {}", filter)
+            return {"bool": {"filter": filter}}
```
portante, line 245: Can't we just use the return at line 257?  By the time we get to 245, `access_term` is still `None`, and `user_term` is must be `None` (`not user` boolean in this `elif`).

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -256,11 +259,18 @@ def _get_user_query(self, parameters: JSON, terms: List[JSON]) -> JSON:
                 }
             )
             self.logger.debug("QUERY: {{}} default: {}", filter)
+            return {"bool": {"filter": filter}}
         else:
             # Either "user" was specified and is already added to the filter,
```
webbnh, line 264: s/is already/will be/ ?

lib/pbench/server/api/resources/query_apis/__init__.py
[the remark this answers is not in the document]
webbnh: > to make sure that an unauthenticated user can never probe usernames by looking for `400` vs `401`, I've strictly outlawed any use of the `"user"` parameter for an unauthenticated user, even for public data.

I support the "minimal attack surface" approach, but I don't want us to cut off our nose to spite our face.  And, I don't know how much effort we want to put in on supporting non-logged-in users, but -- if we care -- a non-logged-in user is almost certainly going to want to search or sort or otherwise select from public data based on owner; so, if we don't provide that selection on the Server side, then we're going to force the Dashboard to fetch _all_ the public records and then filter the presentation to the end user in the browser.  That seems like a waste when we could easily do that filtering in the ES query and save a whole lot of data shuffling.

> On the whole, I don't like the idea of returning fake empty responses for disallowed queries. Again, if you're an unauthenticated client you're not even allowed to ask the question, regardless of whether the username exists or not. It's simple and consistent.

Returning "fake" empty responses is also simple and consistent; and, they aren't necessarily "fake" -- they are equivalent to what would have been returned if we had actually issued the ES query, we just took a shortcut.  My only (weak) concern about refusing to let the requestor ask the question is that I think it increases the data available from which to draw inferences...although, I believe I concur that it doesn't do so sufficiently to succeed.

lib/pbench/test/unit/server/query_apis/test_query_builder.py
```diff
@@ -31,72 +35,131 @@ def current_user(self) -> User:
...
+            (
+                {"user": USER_ID, "access": "private"},
+                {"user": USER_ID, "access": "private"},
+            ),
+            (
+                {"user": USER_ID, "access": "public"},
+                {"user": USER_ID, "access": "public"},
+            ),
+        ],
+    )
```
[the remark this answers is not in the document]
webbnh: Also, are you sure you don't want to use the same parameterization in each of the tests?  I see that you've broken out a few of the special cases, but I note that you're not testing `{"user": SELF_ID}` for the ADMIN user.  (Of course that case should work, either via the "self" path or the "is-ADMIN" path, but wouldn't it be a bummer if neither worked for some reason?!)

lib/pbench/test/unit/server/query_apis/test_query_builder.py
```diff
@@ -31,72 +35,131 @@ def current_user(self) -> User:
...
+           separate test case.
+        """
+        term = {"term": {"icecream": "ginger"}}
+        query = elasticbase._get_user_query(ask[0], [term])
+        filter = [term]
+        if "access" in ask[1]:
+            filter.append({"term": {"authorization.access": ask[1]["access"]}})
+        if "user" in ask[1]:
+            filter.append({"term": {"authorization.owner": ask[1]["user"]}})
+        assert query == {"bool": {"filter": filter}}
```
webbnh, line 105: Should this be refactored into a helper function?

lib/pbench/test/unit/server/query_apis/test_query_builder.py
```diff
@@ -31,72 +35,131 @@ def current_user(self) -> User:
...
+            ({"user": USER_ID}, {"user": USER_ID, "access": "public"}),
+            (
+                {"user": USER_ID, "access": "public"},
+                {"user": USER_ID, "access": "public"},
+            ),
+        ],
+    )
+    def test_noauth(self, elasticbase, server_config, current_user_none, ask):
+        """
+        Test the query builder when we have an authorized user.
```
webbnh, line 121: s/authorized/unauthorized/ ?

lib/pbench/test/unit/server/query_apis/test_query_builder.py
```diff
@@ -31,72 +35,131 @@ def current_user(self) -> User:
...
+                {"user": USER_ID, "access": "private"},
+            ),
+            (
+                {"user": USER_ID, "access": "public"},
+                {"user": USER_ID, "access": "public"},
+            ),
+        ],
+    )
+    def test_admin(self, elasticbase, server_config, current_user_admin, ask):
+        term = {"term": {"icecream": "ginger"}}
```
webbnh, line 56: No docstring?

lib/pbench/test/unit/server/test_authorization.py
```diff
@@ -0,0 +1,144 @@
...
+    @pytest.mark.parametrize(
+        "ask",
+        [
+            {"user": None, "access": "private", "role": API_OPERATION.UPDATE},
+            {"user": None, "access": "public", "role": API_OPERATION.UPDATE},
+            {"user": None, "access": "private", "role": API_OPERATION.DELETE},
+            {"user": None, "access": "public", "role": API_OPERATION.DELETE},
+        ],
+    )
+    def test_disallowed_admin(self, apibase, server_config, current_user_admin, ask):
```
webbnh, line 67: There's no possibility of an "update all" or a "delete all" operation?

lib/pbench/test/unit/server/test_authorization.py
```diff
@@ -0,0 +1,144 @@
...
+
+    @pytest.mark.parametrize(
+        "ask",
+        [
+            {"user": "test", "access": "private", "role": API_OPERATION.UPDATE},
+            {"user": "test", "access": "public", "role": API_OPERATION.DELETE},
+            {"user": "test", "access": "private", "role": API_OPERATION.READ},
+        ],
+    )
+    def test_disallowed_noauth(self, apibase, server_config, current_user_none, ask):
```
webbnh, line 137: Why not include the other two cases (private-delete and public-update)?

webbnh (changes requested):
_[GitHub is being weird, so I hope all (18) of my comments here actually show up...]_

Looks good; got some points for polishing.  (I'll circle back and deal with my open conversations "soon".)

lib/pbench/server/api/auth.py
dbutenhof: Indeed; that's true, but most of these are still not something that will be of any actually use in analyzing a problem, and there's no point in clogging up the log.

lib/pbench/server/api/resources/__init__.py
```diff
@@ -229,30 +228,48 @@ def convert_username(value: Union[str, None], _) -> Union[str, None]:
...
-        internal username representation or None
+        internal username representation
     """
-    if value is None:
-        return None
     if not isinstance(value, str):
-        raise ConversionError(value, str.__name__)
+        raise ConversionError(value, "username")
+    if not Auth.token_auth.current_user():
+        raise UnverifiedUser(value)
```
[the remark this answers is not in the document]
dbutenhof: We need to have `"user"` and `"access"` decoded before we can do an authentication check, so `convert_username` will always come first.

lib/pbench/server/api/resources/__init__.py
```diff
@@ -631,9 +637,13 @@ def _check_authorization(self, user: Union[str, None], access: Union[str, None])
...
         public data will be allowed.
 
-        If the specified "user" context of the operation is None, we default to
-        the authenticated caller, if any; Pbench query operations will do the
-        same.
+        If the specified "user" context of the operation is None, the context
+        is defaulted based on the authenticated user: for an ADMIN user, this
+        means implicitly "all users" (to which the ADMIN has unlimited access);
+        for an unauthenticated user, no access to another "user" is allowed at
+        all, and will be caught at convert_username(); for an authenticated
```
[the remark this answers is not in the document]
dbutenhof: Unauthenticated clients can see public data, but not by username. So `{}` shows them all public data, `{"access": "public"}` shows them all public data, but they can't probe for valid users with `{"user": "drb", "access": "public"}` (which would work if `drb` exists, possibly with an empty return if "drb" has none, and would fail otherwise).

Basically, he idea was not to allow an unauthenticated client to ask "does this username exist"?

And, yeah, the other option is to allow `{"user": "drb", "access": "public"}` and treat usernames as public. I considered this; and I'm not entirely consistent on this. For example, the `datasets/list` query can ask for `"metadata": ["dataset.owner"]` currently without restriction, so we'll reveal the owner of a public dataset. Arguably, if we can't ask for `{"user": "drb", "access": "public"}` we shouldn't reveal `dataset.owner` on unauthenticated queries, either.

I really don't like any of the alternatives, but this approach seemed to make more sense than the others. So far...

lib/pbench/server/api/resources/__init__.py
```diff
@@ -669,30 +679,49 @@ def _check_authorization(self, user: Union[str, None], access: Union[str, None])
...
         # need more checks.
+        #
+        # 1) An unauthorized client can never perform any operation on PRIVATE
+        #    data, nor any operation other than READ on PUBLIC data;
+        # 2) The "user" parameter can be omitted only for READ operations,
+        #    which will be defaulted according to the visibility of the
+        #    authenticated client.
+        # 3) An authenticated client cannot mutate data owned by a different
+        #    user, nor READ private data owned by another user.
+        # 4) An ADMIN user can do anything.
```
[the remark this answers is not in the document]
dbutenhof: I suppose I should move `4)` to `1)`; and, yeah, the terminology seems to have wandered off on its own while I wasn't looking, and needs to be herded back into the corral.

lib/pbench/server/api/resources/__init__.py
```diff
@@ -669,30 +679,49 @@ def _check_authorization(self, user: Union[str, None], access: Union[str, None])
...
+                )
+            elif self.role != API_OPERATION.READ and user is None:
+                self.logger.warning(
+                    "Unauthorized attempt by {} to {} data with defaulted user",
+                    authorized_user,
+                    self.role,
+                )
+                raise UnauthorizedAccess(
+                    authorized_user, self.role, user, access, HTTPStatus.FORBIDDEN
+                )
```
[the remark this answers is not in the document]
dbutenhof: Yes. In theory, we could support a multi-target or wildcard UPDATE/DELETE operation across multiple datasets; Nikhil's recent changes allow `Dataset.query` by user or access column, which wasn't the case before. That may be a worthwhile PR at some point, but I'm not going to try to shoehorn it into here. I'm only trying to generalize the query builder, and this PR is already rather larger than I'd hoped.

So right now only READ makes any sense without a user. For PUBLISH (the only non-READ operation), the user comes implicitly from the selected dataset, so we're always authorizing against the dataset's current user/access anyway. (And access in this case doesn't affect whether we can perform the operation.)

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -77,45 +96,210 @@ def __init__(
...
+        user's data. This method is concerned only with constructing a proper
+        Elasticsearch query to represent the authenticated restrictions.
+
+        Specific cases for user and access values:
+
+            {}: defaulting both user and access
+                All private + public regardless of owner
+
+                ADMIN: all datasets
+                AUTHORIZED: all owner:mine OR access:public
```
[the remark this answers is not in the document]
dbutenhof: I thought that would be a reasonable token that's not too long; but, OK, it's not clear. 😦

lib/pbench/server/api/resources/query_apis/__init__.py
dbutenhof: Right; because it used to be added first, but in order to ensure that user/access terms were in a consistent order for unit testing I introduced the local term variables which are added to the filter at the end. Sigh. No good (or ugly but necessary) deed goes unpunished...

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -79,34 +79,80 @@ def date_range(self, start: AnyStr, end: AnyStr) -> list:
...
+
+        Otherwise, we expect success
+        """
+        has_user = "user" in payload
+        u = payload.get("user")
+        a = payload.get("access")
+        unauthorized = not HeaderTypes.is_valid(header)
+        is_admin = header == HeaderTypes.VALID_ADMIN
+        if has_user and unauthorized:
+            expected_status = HTTPStatus.UNAUTHORIZED
```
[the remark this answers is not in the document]
dbutenhof: And ... you get your wish, because I had to re-jigger all this anyway to account for propagating the `NOT_FOUND` status.

lib/pbench/test/unit/server/query_apis/test_iteration_samples.py
```diff
@@ -263,10 +261,14 @@ def test_query(
...
 
-        if HeaderTypes.is_valid(build_auth_header["header_param"]):
-            expected_status = HTTPStatus.OK
-        else:
-            expected_status = HTTPStatus.FORBIDDEN
+        # get_expected_status() expects to read username and access from the
+        # JSON client payload, however this API acquires that information
+        # from the Dataset. Construct a fake payload corresponding to the
+        # attach_dataset fixture.
+        auth_json = {"user": "drb", "access": "private"}
```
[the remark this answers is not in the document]
dbutenhof: It's a bit more complicated than that since I'd have to look up the appropriate Dataset anyway from the set created by the fixture, which means I still need to assume at least the user. It might have been a bit more intellectually satisfying, but our tests make an enormous amount of assumptions about the users and datasets created by the fixtures and "generalizing" here just didn't seem worth the effort.

lib/pbench/test/unit/server/query_apis/test_query_builder.py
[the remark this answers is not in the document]
dbutenhof: The problem with the `{"user": <admin-id>}` is in capturing the admin user's ID in the fixture. I didn't want to hardcode it; and it just didn't seem worthwhile. Then again, I ended up with `USER_ID` and I might as well have put in an `ADMIN_ID`, as well; I suppose by the time I'd made that reluctant compromise I'd already written off using admin-owned tests.

lib/pbench/test/unit/server/query_apis/test_query_builder.py
dbutenhof: No. ? 😆

lib/pbench/test/unit/server/query_apis/test_query_builder.py
dbutenhof: Before I started, I thought about going back to refactor; when I was done I just wanted it to be done and didn't even think about it. The bit about being able to parametrize "x,y" is actually new to me, and if I do that I'll need to take a hacksaw to all of this code anyway, so I'll think about it. It's already clear this PR will *never* be done anyway since I can't even read one set of comments before another set drops!

lib/pbench/test/unit/server/query_apis/test_query_builder.py
dbutenhof: Yes, I need a smarter copy-and-paste.

lib/pbench/test/unit/server/test_authorization.py
dbutenhof: Someday, maybe; not now. Certainly not in this PR.

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -77,45 +96,210 @@ def __init__(
...
             JSON query parameters containing keys:
-                "user": Pbench internal username if present and not None
-                "access": Access category, "public" or "private"; defaults
-                    to "private" if "user" is specified, and "public" if
-                    user is not specified. (NOTE: other combinations are
-                    not currently supported.)
+                "user": Pbench internal username if present and not None; if
+                    omitted, doesn't restrict search on user.
+                "access": Access category, "public" or "private"; defaults to
+                    don't-care
```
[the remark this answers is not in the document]
webbnh: But, in terms of the code which this docstring describes, the tokens `"public"` and `"private"` never appear; instead it references `Dataset.PUBLIC_ACCESS` and `Dataset.PRIVATE_ACCESS`, whatever values those might have....  (And, if those values, which are defined elsewhere, are ever changed, what will cause this comment to be kept up to date with them?)

lib/pbench/server/api/resources/__init__.py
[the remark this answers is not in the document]
webbnh: At line 241, you have `BAD_REQUEST/400` while at line 242 you have just `FORBIDDEN` (no numerical equivalent).  It would probably be good to settle on whether to include the literal or not....

lib/pbench/test/unit/server/query_apis/test_query_builder.py
webbnh: This comment now applies to lines 140-142 in `test_auth_access_private()` and lines 165-167 in `test_noauth_access_private()`.

lib/pbench/server/api/auth.py
portante: See also: https://github.com/distributed-system-analysis/pbench/discussions/2509

lib/pbench/server/api/auth.py
portante: See also https://github.com/distributed-system-analysis/pbench/discussions/2509.

lib/pbench/server/api/resources/query_apis/__init__.py
portante: Adding a `debug` log entry would be great.

I have not reviewed the tests yet, but having a `debug` log will help us catch behavioral drift over time.

lib/pbench/server/api/resources/query_apis/__init__.py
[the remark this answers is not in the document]
webbnh: It was clear to me, but...it needs to be clear in general.  😕

lib/pbench/test/unit/server/query_apis/test_query_builder.py
webbnh: > I can't even read one set of comments before another set drops!

I apologize on behalf of GitHub for that.  I typically do reviews with two tabs open in my browser -- one to the "Conversations" view and one to the "Files" view.  Once you start a review, then every comment you enter after that becomes part of the review; if you refresh the "Conversations" view after starting a review in the "Files" view, then all of the "Reply" text boxes disappear (because it then knows that you have a review "in progress"), and you cannot enter any comments from that page; however, if you _don't_ refresh the "Conversations" view after starting a review, then you can still enter comments...but doing so submits the whole review, including any comments that you've entered in the other tab...and it does so without updating the other tab (which is why I made that cryptic comment earlier -- because I couldn't find half the comments that I thought I'd entered into my review!).  (I generally refresh my "Conversations" tab just before submitting my review because it displays all of my comments in one place (marked as "pending"), so that I can review them.)

I was halfway through my review, and I saw something in the "Conversations" view that I wanted to comment on directly, outside the context of my review, but hitting the "Comment" button committed my incomplete review...and so, once I submitted my review, you then saw a second batch of comments arrive.  That was not the presentation that I had intended.

lib/pbench/server/api/resources/__init__.py
```diff
@@ -641,38 +647,88 @@ def _check_authorization(self, user: str, access: str):
...
                 )
-                authorized = False
-        if not authorized:
-            raise UnauthorizedAccess(authorized_user, self.role, user, access)
+                raise UnauthorizedAccess(
+                    authorized_user, self.role, user, access, HTTPStatus.FORBIDDEN
+                )
+            else:
+                # We have determined that there is an authenticated user with
+                # legitimate access to the specified user and access: the
```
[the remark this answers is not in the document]
portante: I would have thought you'd want this to say:

```
We have determined that there is an authenticated user with legitimate access to the specified user and access to the data: the data is public and the operation is READ, the user owns the data, or the user has ADMIN role.
```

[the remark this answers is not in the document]
portante: But not worth holding this commit for such a change.

portante (approved):

lib/pbench/server/api/resources/__init__.py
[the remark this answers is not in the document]
dbutenhof: I was trying to avoid the apparent confusion from the reference to the dataset "user" and "access (type)" properties by simplifying the sentence. Maybe "legitimate access based on the dataset's owner and access policy" would be slightly more clear, but it's also longer and more unwieldy. I'm happy with it as is.

Issue #2025, opened 2020-12-06
Use pbench agent logger throughout the agent code base


portante:
In some parts of the pbenh-agent code, we're using libc printf `%` substitution, and `{}` substitution on the server side, with the implication that it's basically impossible to log correctly in common code.

Besides, using `%` just feels like going back to C, and we haven't been playing in that playground for a long time.

Let's convert all loggers throughout the pbench code base to use the commonly constructed logger.

_Originally posted by @dbutenhof in https://github.com/distributed-system-analysis/pbench/pull/2014#discussion_r536391000_

PR #61, opened 2015-08-16
Get the bench-scrips unit-tests working

portante:
We had removed configtools from the pbench report and moved it to its
own repository, so now we just require it to be installed and available
for the unit tests.  As a result, we just needed to remove the in-tree
build of configtools performed by the unit test script.

And fix the pbench_uperf option processing which was emitting a debug
message for "--" instead of just recognizing the termination of the
options.

Issue #10, opened 2015-07-15
Develop an "all-in-one" environment for development and testing


portante:
It is pretty clear that both for development and for simple example use / kick the tires testing we need an all-in-one environment for deploying the agent, background server tasks, web server, etc.

With such an environment we could automate builds and unit tests for integration with TravisCI, and other tools.

atheurer:
I agree.  Related to that: since using pbench currently relies on a binary
packages to provide functionality (not just providing the agent, but
providing tools and benchmarks), we need to decide if this method is
mandatory, or if there should be an alternative.  'make install' is fine
-if it actually install everything (including all benchmarks and tools).
However, we will need to account for that method of use in our current code
("package" checking and installing).

portante:
Most likely using `make install` will be preceded by a `./configure` step, and with that, other RPM packaging becomes a bit easier as well.  So this is worth doing for packaging reasons as well.

ndokos:
On 07/15/2015 11:43 PM, Peter Portante wrote:
>
> Most likely using |make install| will be preceded by a |./configure| 
> step, and with that, other RPM packaging becomes a bit easier as well. 
> So this is worth doing for packaging reasons as well.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/distributed-system-analysis/pbench/issues/10#issuecomment-121818484>.
>

This is just for the pbench agent for now, right?

portante:
Yes, and that begs the question, should we break up pbench into its subcomponents?

ndokos:
On 07/16/2015 12:04 AM, Peter Portante wrote:
>
> Yes, and that begs the question, should we break up pbench into its 
> subcomponents?
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/distributed-system-analysis/pbench/issues/10#issuecomment-121823146>.
>
I think we should.

PR #2468, opened 2021-09-29
State Signals in Pbench

Maxusmusti:
VERY MUCH IN PROGRESS - draft opened for viewable progress

Goal: To replace existing proprietary Redis solution with state-signals package.


webbnh:
@Maxusmusti, it looks like Mr. Jenkins is _really_ enjoying your changes:

>test-49 pbench-copy-results --prefix=foo/bar3% --user=ndk20%
PASS - test-49
test-51 test-start-stop-tool-meister 
🕸️ 

[That's the infinite spinning wheel of doom.]

So, the tests have been running for 90 minutes, now...we'll probably have to manually kill them.

webbnh:
Not that it's a particularly big deal for a draft PR, but Black, Flake8, and LGTM are _all_ unhappy with you....  :wink:

lib/pbench/agent/tool_meister_client.py
```diff
@@ -146,82 +114,61 @@ def publish(self, group, directory, action, args=None):
...
-                self.logger.error(
-                    "Failed to publish to the TDS, encountered %d subscribers"
-                    " on the channel",
-                    num_present,
-                )
+            if resp != 0:
+                self.logger.error(f"Missing or bad response from the TDS, {resp}")
                 ret_val = 1
+                for responder in msgs:
+                    msg = msgs[responder]
```
webbnh, line 133: How about
```
                for responder, msg in msgs.items():
```

webbnh (changes requested):
Found a bug; have a bunch of suggestions.

agent/util-scripts/gold/test-start-stop-tool-meister/test-51.txt
```diff
@@ -16,22 +16,13 @@ pbench-tool-meister-start: 5. starting tool data sink
...
-pbench-tool-meister-stop: constructing Redis() object
-pbench-tool-meister-stop: next pbench-agent-cli-to-client
-pbench-tool-meister-stop: payload from pbench-agent-cli-to-client: {'type': 'message', 'pattern': None, 'channel': b'pbench-agent-cli-to-client', 'data': b'{"action": "end", "kind": "ds", "status": "success"}'}
-pbench-tool-meister-stop: publish end on chan pbench-agent-cli-from-client
-pbench-tool-meister-stop: publish terminate on chan pbench-agent-cli-from-client
-pbench-tool-meister-stop: published pbench-agent-cli-from-client
+pbench-tool-meister-start: publish state signal for state init with metadata: {'group': 'default', 'directory': '/var/tmp/pbench-test-utils/pbench/mock-run/tools-default', 'args': None}
+pbench-tool-meister-stop: publish state signal for state end with metadata: {'group': 'default', 'directory': '/var/tmp/pbench-test-utils/pbench/mock-run/tools-default', 'args': None}
+pbench-tool-meister-stop: publish state signal for state terminate with metadata: {'group': 'default', 'directory': 'None', 'args': {'interrupt': False}}
+pbench-tool-meister-stop: publish terminate signal
```
dbutenhof, line 25: Isn't this somewhat redundant with the previous more detailed message?

dbutenhof (commented):
I have nothing substantial to add to Webb's voluminous commentary; aside from several of his excellent suggestions, this looks pretty good.

portante (commented):
Moving to `Draft` mode for now until the last round of reviews are posted.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -964,10 +964,7 @@ def main(_prog: str, cli_params: Namespace) -> int:
     # drive the following client operations ("sysinfo" [optional] and "init"
     # [required]).
     with Client(
-        redis_server=redis_client,
-        channel_prefix=cli_tm_channel_prefix,
-        to_client_chan=to_client_chan,
-        logger=logger,
+        redis_host=redis_server.host, redis_port=redis_server.port, logger=logger
```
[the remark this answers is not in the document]
Maxusmusti: state-signals currently only uses redis host and port for connection, so that information would have to be extracted at some point (either in pbench-tool-meister-start or in within the Client object creation code). This could be changed in the future with a state-signals update if desired though.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -883,6 +877,12 @@ def __enter__(self):
                 "'tm_log_capture' thread started, processing Tool Meister logs ..."
             )
 
+        # Set up signal responder for state control
+        self.sig_resp = state_signals.SignalResponder(
+            redis_host=self.redis_host, redis_port=self.redis_port
+        )
```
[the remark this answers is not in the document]
Maxusmusti: This is what I though was the desired functionality, though it may be worth discussing if there is a potential use-case for having two separate redis servers, maybe one only accessible to the agent components and accessible to the benchmark runners?

webbnh (changes requested):
These changes seem good in general, but I have a few concerns:
- there should be type hints on new and modified function signatures;
- there seem to be some questions around `api_tm_allowed_actions` vs. `tm_allowed_actions` and when to use which; and,
- some of the payloads' values for the `"directory"` field changed from `null`/`None` to the string, `"None"`, and I think we should reconsider that (but, I can't find the source code change which caused it...).

And, I have a couple of other questions.

lib/pbench/agent/tool_data_sink.py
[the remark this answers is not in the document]
webbnh: In my mental model (which, albeit, doesn't seem to match reality, necessarily 😆), the State Signals facility is a black box which is independent of how Pbench uses Redis.  (That is, the fact that State Signals uses Redis is an implementation detail which would be beyond the scope of Pbench, except that we have to give it a host and port.)

So, this code which has State Signals using the same Redis server as the rest of the TM is OK but is, in some sense, a violation of encapsulation.  Still, if State Signals isn't going to create its own Redis server and is going to rely on one provided by Pbench, then the above is a perfectly reasonable way to go.

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
portante: I am working on a change to state-signals independent of this to add support for the redis client.

agent/util-scripts/pbench-tool-meister-client
```diff
@@ -77,11 +77,8 @@ def main(argv: list) -> int:
         logger.error(str(exc))
         return exc.return_code
 
-    with Client(
-        redis_host=redis_server.host,
-        redis_port=redis_server.port,
-        channel_prefix=cli_tm_channel_prefix,
-        logger=logger,
+    with Client.create_with_redis(
```
portante, line 80: Is there a use of `Client` with out the `create_with_redis`?

portante (changes requested):
I am still digging into this code change, please pardon me for taking so long.

lib/pbench/agent/tool_meister_client.py
```diff
@@ -146,82 +101,55 @@ def publish(self, group, directory, action, args=None):
...
 
+    def publish(self, group, directory, action, args=None):
+        """publish a state signal formed from the group, directory,
+        action, and args arguments.
+
+        Returns 0 on success, 1 on failure; logs are also written for any
+        errors encountered.
+        """
+        if action not in api_tm_allowed_actions:
+            return 1
```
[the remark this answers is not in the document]
Maxusmusti: We still need a check for at least `tm_allowed_actions`, as we want to make sure the signal the user is trying to publish is in the legal event space (it would be caught by state-signals eventually, but it'd be quicker to just catch it b4 any other work). I can switch to just using `tm_allowed_actions` and allowing terminates to also be published through the standard publish method, though. I will also add logging for when we catch an illegal event.

lib/pbench/agent/tool_meister_client.py
```diff
@@ -35,108 +29,69 @@ class Client:
...
+        redis_port: str,
+        publisher_name: str = "pbench_client",
+        logger: logging.Logger = None,
+    ):
+        sig_pub = state_signals.SignalExporter(
+            publisher_name, redis_host=redis_host, redis_port=redis_port
+        )
+        sig_pub.initialize_and_wait(
+            1, list(api_tm_allowed_actions) + ["terminate"], tag="from_pbench_client",
+        )
```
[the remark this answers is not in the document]
Maxusmusti: Made the switch. Also current state-signals (0.5.0) requires the input to be a list of strings, but that can be changed in the future.

agent/util-scripts/pbench-tool-meister-client
Maxusmusti: You can also use `create_with_exporter`. Additionally, you can use the `Client`'s standard `__init__` function, and pass in an existing `SignalExporter` as well as `False` for `to_be_shutdown`.

agent/util-scripts/gold/test-start-stop-tool-meister/test-51.txt
```diff
@@ -232,9 +205,9 @@ DEBUG pbench-tool-meister _send_client_status -- publish pbench-agent-cli-from-t
...
-DEBUG pbench-tool-meister fetch_message -- channel pbench-agent-cli-to-tms payload, '{"action": "terminate", "args": {"interrupt": false}, "directory": null, "group": "default"}'
-DEBUG pbench-tool-meister wait_for_command -- testhost.example.com: msg - {'action': 'terminate', 'args': {'interrupt': False}, 'directory': None, 'group': 'default'}
+DEBUG pbench-tool-meister fetch_message -- payload from pbench-agent-cli-to-tms: {'type': 'message', 'pattern': None, 'channel': b'pbench-agent-cli-to-tms', 'data': b'{"action": "terminate", "args": {"interrupt": false}, "directory": "None", "group": "default"}'}
+DEBUG pbench-tool-meister fetch_message -- channel pbench-agent-cli-to-tms payload, '{"action": "terminate", "args": {"interrupt": false}, "directory": "None", "group": "default"}'
+DEBUG pbench-tool-meister wait_for_command -- testhost.example.com: msg - {'action': 'terminate', 'args': {'interrupt': False}, 'directory': 'None', 'group': 'default'}
```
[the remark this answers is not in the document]
Maxusmusti: It's because in the `_publish` method, the `directory` arg is cast to a string, which makes `None` become `'None'`. Will fix

agent/util-scripts/gold/test-start-stop-tool-meister/test-51.txt
```diff
@@ -16,23 +16,12 @@ pbench-tool-meister-start: 5. starting tool data sink
 pbench-tool-meister-start: 6a. starting localhost tool meister
 pbench-tool-meister-start: 7. waiting for all successfully created Tool Meister processes to show up as subscribers
 pbench-tool-meister-start: 8. Initialize persistent tools
-pbench-tool-meister-start: channel pbench-agent-cli-to-client payload, '{"action": "init", "kind": "ds", "status": "success"}'
 pbench-tool-meister-start: channel pbench-agent-cli-to-client payload, '{"action": "startup", "kind": "ds", "status": "success"}'
 pbench-tool-meister-start: next pbench-agent-cli-to-client
-pbench-tool-meister-start: next pbench-agent-cli-to-client
```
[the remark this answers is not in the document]
Maxusmusti: This is expected. "Startup" is still handled in pbench-tool-meister-start using the traditional method. The new client, however, has been switched over to state-signals, so the "init" signal is the first action now handled by state-signals rather than the traditional method. This is why we see the traditional debug logging for the action "startup" still, but no longer for "init".

portante (changes requested):
This needs a rebase.

And when you rebase it, can you update the commit message to add a separate line with "PBENCH-181" to help us link this PR to our Jira instance?

portante (approved):

portante (dismissed):

Maxusmusti:
Just rebasing

webbnh (commented):
First, I want to say thank you, @Maxusmusti, for finishing this work for us.  And, second, I want to apologize for taking so long -- this got put on the back burner and it's been hard to pull it to the front again.

The changes generally look good to me.  There is one piece of suspect code which should probably be checked and either removed or corrected.  The rest of it are various levels of nits and suggestions.

portante (changes requested):
Not sure why the legacy agent tests are failing, but they are.  We need to address that.  Let me know if you'd like help looking into the failures.

Maxusmusti:
Found the bug, at some point the TDS code changed from using `self.tool_group` to `self.params.tool_group`, and I must've missed that when handling merge conflicts.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1295,13 +1300,29 @@ def execute(self):
...
-                ret_val = self._valid_data(data)
-                if ret_val is None:
+            # Begin listening for client state signals, and react/respond
+            # accordingly upon signal receipt
+            for signal in self.sig_resp.listen():
+                client = signal.publisher_id
+                action = signal.event
+                if action == "shutdown":
+                    # This simply means that a SignalExporter/Client we
+                    # were listening to has shut down, no action needed
```
portante, line 1310: Won't this result in a hang?

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1295,13 +1300,29 @@ def execute(self):
...
+                    continue
+                if action == "initialization":
+                    self.sig_resp.srespond(signal)
                     continue
-                action, directory, args = ret_val[0], ret_val[1], ret_val[2]
-                self.execute_action(action, directory, args, data)
-            self.logger.info("Tool Data Sink terminating")
+                if not signal.metadata:
```
portante, line 1315: What does it mean to get a signal without metadata?  Perhaps we should add a `self.logger.debug()` with the contents?

lib/pbench/agent/tool_meister_client.py
```diff
@@ -41,195 +34,173 @@ class Client:
...
 
-    def terminate(self, group, interrupt=False):
+    def publish(self, group: str, directory: str, action: str, args=None) -> int:
+        """publish a state signal formed from the group, directory,
+        action, and args arguments.
+
+        Returns 0 on success, 1 on failure; logs are also written for any
+        errors encountered.
+        """
+        if action not in tm_allowed_actions:
```
portante, line 179: We have moved the bulk of the code out of this method into a helper, but we are only doing the action test here?  Is it worth making this change?

lib/pbench/agent/tool_meister_client.py
```diff
@@ -41,195 +34,173 @@ class Client:
...
-            if num_present != 1:
-                self.logger.error(
-                    "Failed to publish to the TDS, encountered %d subscribers"
-                    " on the channel",
-                    num_present,
-                )
+            if resp != 0:
+                self.logger.error("Missing or bad response from the TDS, %s", str(resp))
                 ret_val = 1
+                for responder, msg in msgs.items():
```
portante, line 163: What does it mean to get more than one responder, msg pair?

lib/pbench/agent/tool_meister_client.py
```diff
@@ -41,195 +34,173 @@ class Client:
...
         #     "args": "< arbitrary argument payload for a particular action >"
         #   }
         # The caller of tool-meister-client must be sure the directory argument
         # is accessible by the Tool Data Sink instance.
-        self.logger.debug("publish %s on chan %s", action, self.from_client_channel)
-        msg = dict(action=action, group=group, directory=str(directory), args=args)
+
+        # This check/conversion is necessary, as the pbench scripts sometimes pass
+        # in the directory as a str, and sometimes as pathlib.PosixPath
+        if directory and not isinstance(directory, str):
```
portante, line 140: Why do we have to make this check now?  What changed that is causing this problem?

portante (changes requested):

lib/pbench/agent/tool_meister_client.py
Maxusmusti: I've just formalized/cleaned-up to old "cast to string" line (before it was just `if directory: directory=str(directory)`. This is done because some pbench scripts pass a string path into the Client object, and others pass in a PosixPath object, which would result in the following error if left unconverted: `TypeError: Object of type PosixPath is not JSON serializable`

lib/pbench/agent/tool_meister_client.py
Maxusmusti: When publishing a signal, each responder can respond with an optional message (potentially including reasons why it failed or other desirable details). The user then receives a dict of all of the messages received in responses, indexed by the responders' IDs.

lib/pbench/agent/tool_meister_client.py
Maxusmusti: Yeah, I believe you had explained to me earlier in the year that it would be good practice to separate the mechanism of actually using the `SignalExporter` for publishing/response collecting into a helper, while leaving all additional functionality of the client publish in the original method. That way, if anything needs to change in the `SignalExporter`/actual publishing mechanism, the helper can be changed and the actual method remains consistent, and if we want to change Pbench Client specific checks/allowances/features, the original method can change and the mechanism helper can remain constant.

lib/pbench/agent/tool_meister_client.py
portante: Didn't the original code already handle that with `str(directory)` without the test?

lib/pbench/agent/tool_data_sink.py
portante: But will this client receive another message?

portante: We might want to add a debug log message for this.

lib/pbench/agent/tool_meister_client.py
portante: Sure that makes sense, but ... just to avoid one test it doesn't seem worth it.

lib/pbench/agent/tool_meister_client.py
portante: We were not waiting for more than one responder before, were we?

lib/pbench/agent/tool_meister_client.py
Maxusmusti: In cases where `directory=None`, simply doing `str(directory)` will convert `None` into a string of value `"None"`. Additionally I figured we should be gracefully catching the scenario where something that can't be cast to string is passed in.

lib/pbench/agent/tool_meister_client.py
Maxusmusti: Nono, the most common use-case is still expected to be only one responder (a single TDS). It's just implemented like this for robustness/extensibility (if something goes wrong somewhere else, or there are multiple tool data sinks running, or if something new wants to be subscribing/responding to clients in the future, etc.). Just to make sure we are capturing everything that could be going on in the event of a failure or preparedness for future use cases.

lib/pbench/agent/tool_meister_client.py
Maxusmusti: Should I recombine them into one then?

Maxusmusti:
Not sure what exactly is going wrong with test 61, let me know if it's obvious to you. If not, I'll look into it and fully debug/fix later tonight.

lib/pbench/agent/tool_meister_client.py
```diff
@@ -114,11 +134,20 @@ def _publish(self, group: str, directory: str, action: str, args=None) -> int:
...
+            try:
+                directory = str(directory)
+            except Exception:
+                self.logger.exception(
+                    "Publish failed: directory must be compatible with type string, was instead %s",
+                    type(directory),
+                )
         metadata = dict(group=group, directory=directory, args=args)
```
webbnh, line 148: The new comment is a huge help; however, given that this code is necessary, you should fix the function signature type hint:
```
    def _publish(self, group: str, directory: Union[str, Path, None], action: str, args=None) -> int:
```

Also, if `directory` is _not_ "true-y", then you're using its value unaltered at line 148, which seems like it might not always be what you want.  I'm wondering if structuring the code something like this might be better:
```
        if not directory:
            dir_val = None
        elif isinstance(directory, str):
            dir_val = directory
        elif isinstance(directory, Path):
            dir_val = str(directory)
        else:
            self.logger.error(
                "Publish failed: directory must be compatible with type string, was instead %s",
                type(directory),
            )
            dir_val = None
        metadata = dict(group=group, directory=dir_val, args=args)
```
This avoids overwriting the parameter, which is generally good practice.  It ensures that the directory value is `None` for all cases where it doesn't have a useful value.  If it's a `Path`, we coerce it to a string; if it's a string, we use it as is; and, otherwise, we log an error and use `None`.

Obviously, there are other ways to slice this, but this is one slice at it.

webbnh (commented):
Looks good to me, with the one caveat around the directory parameter handling.

portante (changes requested):
During review of PR [#2468](https://github.com/distributed-system-analysis/pbench/pull/2468), I found a problem in the state signals code which I filed as PR [#9 (state-signals)](https://github.com/distributed-system-analysis/state-signals/pull/9).

I also filed PR [#10 (Maxusmusti/pbench)](https://github.com/Maxusmusti/pbench/pull/10) to simplify the code a bit and add some debugging help.

Can we consider those changes first?

portante (approved):

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1295,19 +1300,45 @@ def execute(self):
...
+                    self.logger.debug("Client %s has shut down", client)
                     continue
-                action, directory, args = ret_val[0], ret_val[1], ret_val[2]
-                self.execute_action(action, directory, args, data)
-            self.logger.info("Tool Data Sink terminating")
+                if action == "initialization":
+                    self.sig_resp.srespond(signal)
+                    continue
+                if not signal.metadata:
+                    self.logger.debug(
```
dbutenhof, line 1320: Should this really be a `debug`? It means a remote is mismatched or broken, right? Unlikely as it (hopefully) is, shouldn't this be at least a `warning`?

Or better, this should probably just be removed, as it's redundant with the more complete check in `_is_valid_data` immediately below ...

lib/pbench/common/utils.py
```diff
@@ -126,11 +126,14 @@ def cleanup(self):
...
         ensure that cleanup continues as best we can.
         """
+        self.logger.debug(f"beg - {self.name}")
         try:
             self.action()
         except Exception:
             # TODO: f-string used because this is shared by agent and server
             self.logger.exception(f"Unable to {self}")
+        else:
+            self.logger.debug(f"end - {self.name}")
```
webbnh, line 137: The concern about the use of an f-string at line 133 is misplaced:  that is, I'm reasonably certain that we will always log the exception, and so there is nothing to be saved by deferring the formatting (so the use of an f-string here presents no problem).

On the other hand, at lines 129 and 136, we will (presumably) _usually_ be skipping the logging of debug messages, and so the use of an f-string there is disappointing but understandable (and worth noting with a `TODO`).

(So, if there is cause to edit things for some other reason, we should remove line 133 and place it before each of lines 129 and 136....)

lib/pbench/agent/tool_meister_client.py
```diff
@@ -87,149 +79,120 @@ def __init__(
...
+        continuing.
 
         Returns 0 on success, 1 on failure; logs are also written for any
         errors encountered.
+
         """
-        if action not in api_tm_allowed_actions:
+        if action not in tm_allowed_actions:
+            self.logger.warning(f"Attempted to publish illegal action '{action}'")
             return 1
```
webbnh, line 139: Should Pbench still be responsible for checking and enforcing this?  (Didn't we enable State Signals to do this...or what was the point of passing `tm_allowed_actions` to `SignalExporter.initialize_and_wait()` at line 116?)

lib/pbench/agent/tool_meister_client.py
```diff
@@ -87,149 +79,120 @@ def __init__(
...
+        )
+        self.logger.debug(
+            "publish state signal for state %s with metadata: %s", action, metadata
+        )
         try:
-            num_present = self.redis_server.publish(
-                self.from_client_channel, json.dumps(msg)
+            resp, msgs = self.sig_pub.publish_signal(
+                event=action, tag="from_pbench_client", metadata=metadata, timeout=100
             )
```
webbnh, line 165: That `100` looks like a "magic number".  It would be good to have it rendered as a symbolic constant, for all the usual reasons.

lib/pbench/agent/tool_meister_start.py
```diff
@@ -1267,13 +1267,14 @@ def start(_prog: str, cli_params: Namespace) -> int:
...
             )
 
-        # Setup a Client API object using our existing to_client_chan object to
-        # drive the following client operations ("sysinfo" [optional] and "init"
-        # [required]).
+        # Setup a Client API object using our existing Redis server connection
+        # to drive the following client operations: "sysinfo" [optional], and
+        # "init" [required].
+        #
+        # FIXME - What should be done with the `to_client_chan` object?
```
webbnh, line 1274: Um, we're not supposed to accept code with `FIXME`'s in it...right?

webbnh (changes requested):
The changes look nice and streamlined, now.  However, we seem to have an unresolved `FIXME`, and then there are some smaller items, like a magic number, a pointed question, and a nit about a comment.

lib/pbench/agent/tool_meister_start.py
portante: Mea culpa.  That was a comment left in the PR that was merged into this PR.  I think the answer is that we should close that channel.

lib/pbench/common/utils.py
portante: The problem here is with the formatting string itself when not using an f-string.  So the comment really applies to lines 128, 134, and 136.

You would need to use `self.logger.debug("end - {}", self.name)` when it is in the server code, and `self.logger.debug("end - %s", self.name)` when it is in the agent code.

See issue #2025.

lib/pbench/agent/tool_meister_client.py
portante: @Maxusmusti, do you know the answer to this?

lib/pbench/common/utils.py
webbnh: I concur.  But, in the case of line 134, there is no benefit to deferring the formatting, so we might as well use the f-string there, because it is more intuitive for the reader.  And, therefore, there is no point to having the code comment at line 133 (so it should be removed).

On the other hand, re #2025, we _should_ have a comment like the one at line 133 before each of the new calls to `logger.debug()`, because those calls _would_ benefit enough from deferring the formatting to justify the additional complexity...once we can.

lib/pbench/agent/tool_meister_client.py
Maxusmusti: Yeah, this is a great point, the check has become redundant, as state-signals already handles it. The original logic was to keep it in at the beginning for an even faster failure, but upon further inspection it really looks like it's doing more harm than good.

lib/pbench/agent/tool_data_sink.py
Maxusmusti: I think it's good to keep the two separate, as `_is_valid_data` simply checks the validity of a passed-in metadata (assumes it exists), so a check to make sure there is ANY data would still be required. I think you are right about changing it to a `warning`, though, as that keeps the logging in-line with how `_is_valid_data` reports malformed data.

lib/pbench/agent/tool_meister_client.py
```diff
@@ -124,7 +129,9 @@ def __exit__(self, *args):
         """On context exit, just close down the to SignalExporter object."""
         self.sig_pub.shutdown()
 
-    def publish(self, group: str, directory: str, action: str, args=None) -> int:
+    def publish(
+        self, group: str, directory: Union[str, Path, None], action: str, args=None
+    ) -> int:
```
webbnh, line 134: The last parameter is missing its type hint...something like:
```
    def publish(
        self,
        group: str,
        directory: Union[str, Path, None],
        action: str,
        args: Any = None,
    ) -> int:
```
(Sadly, Black completely wraps it....)

webbnh (approved):
Looks good to me, except for one little item; however, I can live with it if we don't want to endure another round of review.

lib/pbench/agent/tool_meister_client.py
```diff
@@ -13,25 +13,22 @@
...
-    api_tm_allowed_actions,
-    cli_tm_allowed_actions,
-    cli_tm_channel_prefix,
-    tm_channel_suffix_from_client,
-    tm_channel_suffix_to_client,
-)
-from pbench.agent.redis_utils import RedisChannelSubscriber
+from pbench.agent.constants import cli_tm_allowed_actions, tm_allowed_actions
 from pbench.agent.tool_group import ToolGroup
 from pbench.agent.utils import RedisServerCommon
```
webbnh, line 26: I'm not sure why, but apparently `isort` is objecting to this ordering:
```
ERROR: /home/jenkins/pbench/lib/pbench/agent/tool_meister_client.py Imports are incorrectly sorted and/or formatted.
```

webbnh (approved):
👍

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1295,19 +1300,45 @@ def execute(self):
...
+                    continue
+                if not self._is_valid_data(action, signal.metadata):
+                    continue
+                data = signal.metadata
+                data["action"] = action
+                self.execute_action(client, data)
+                if action == "terminate":
+                    self.logger.info("Tool Data Sink terminating")
+                    self._send_client_status(client, action, "success")
+                    break
```
portante, line 1333: What changed with state signals that we need to explicitly use a `break` statement here?

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1295,19 +1300,45 @@ def execute(self):
...
+            for signal in self.sig_resp.listen():
+                client = signal.publisher_id
+                action = signal.event
+                self.logger.debug(
+                    "client %r, action %r, metadata %r", client, action, signal.metadata
+                )
+                if action == "shutdown":
+                    # This simply means that a SignalExporter/Client we
+                    # were listening to has shut down, no action needed
+                    self.logger.debug("Client %s has shut down", client)
```
portante, line 1314: I think this debug statement could be removed since the same data is logged at line 1308.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1295,19 +1300,45 @@ def execute(self):
...
                     continue
-                action, directory, args = ret_val[0], ret_val[1], ret_val[2]
-                self.execute_action(action, directory, args, data)
-            self.logger.info("Tool Data Sink terminating")
+                if action == "initialization":
+                    self.sig_resp.srespond(signal)
```
portante, line 1317: Can we add a comment here that the state signals responder requires this `srespond()` call?

portante: The next question is should the tool data sink respond to any client?  Or just a pbench client?  Meaning, should we be looking at the client ID as set by the API call to know it is a publisher we are interested in?

portante (changes requested):
Looking good, still have some questions and comments though just on the Tool Data Sink portion.

portante:
@Maxusmusti and the tests are failing again.

Maxusmusti:
> @Maxusmusti and the tests are failing again.

Yeah if it's #61 then that one has been failing for a while now...not sure why. Could you take a look and see if the reason is obvious to you @portante

lib/pbench/agent/tool_data_sink.py
webbnh: Doesn't the publisher know how many subscribers there are?  If so, wouldn't it be "rude" not to respond, since the publisher could conceivably be waiting for ack's from _all_ subscribers...no?  (More to the point, though, how would we end up subscribed to a publisher that we're _not interested in_??)

portante: I believe the `initialization` step is the way the publisher comes to know how many subscribers are out there.  If that is not the case, we need a better understanding of the protocol.

portante:
@Maxusmusti, the failure of `agent/util-scripts/test-61` is due to `pbench-start-tools` not being able to send the `terminate` message to the Tool Data Sink as expected.  I had thought the `listen()` method change would fix that, but something else is going on.  I'll see if I can debug it further tomorrow.

portante:
Posted https://github.com/Maxusmusti/pbench/pull/11 to correct the failing `test-61`.

lib/pbench/agent/tool_data_sink.py
Maxusmusti: The tool data sink is already only responding to pbench clients. The responder is locked onto a specific tag `"from_pbench_client"`, and only receives/responds to signals with those tags (see line 911 in tds). The pbench clients are expecting one subscriber (the tds), and the response to `initialization` is the means by which the tds actually acknowledges the client and subscribes

portante:
@dbutenhof, do you know why this server test would fail with an agent side change?

```
___________________ TestUpload.test_mismatched_md5sum_header ___________________
[gw8] linux -- Python 3.9.13 /var/tmp/jenkins/tox/py39/bin/python

self = <pbench.test.unit.server.test_requests.TestUpload object at 0x7efc8b82a4c0>
client = <FlaskClient <Flask 'api-server'>>
caplog = <_pytest.logging.LogCaptureFixture object at 0x7efc8a6f7e50>
server_config = <pbench.server.PbenchServerConfig object at 0x7efc8a70a490>
setup_ctrl = None
pbench_token = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpYXQiOjE2NjQzNzcxODksImV4cCI6MTY2NDM4MDc4OSwic3ViIjozfQ.Wpzuwy-p70SlAztY_MPuMHzLQ0YqICcTWoExh_oKWI0'

    def test_mismatched_md5sum_header(
        self, client, caplog, server_config, setup_ctrl, pbench_token
    ):
        filename = "log.tar.xz"
        datafile = Path("./lib/pbench/test/unit/server/fixtures/upload/", filename)
    
        with datafile.open("rb") as data_fp:
            response = client.put(
                self.gen_uri(server_config, filename),
                data=data_fp,
                # Content-Length header set automatically
                headers=self.gen_headers(pbench_token, "md5sum"),
            )
        assert (
            response.json.get("message")
            == "MD5 checksum 9d5a479f6f75fa9b3bab27ef79ad5b29 does not match expected md5sum"
        )
>       self.verify_logs(caplog)

/var/tmp/jenkins/tox/py39/lib/python3.9/site-packages/pbench/test/unit/server/test_requests.py:271: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

caplog = <_pytest.logging.LogCaptureFixture object at 0x7efc8a6f7e50>

    @staticmethod
    def verify_logs(caplog):
        for record in caplog.records:
            assert record.levelname not in ("ERROR", "CRITICAL")
>       assert caplog.records[-1].levelname == "WARNING"
E       AssertionError: assert 'DEBUG' == 'WARNING'
E         - WARNING
E         + DEBUG

/var/tmp/jenkins/tox/py39/lib/python3.9/site-packages/pbench/test/unit/server/test_requests.py:115: AssertionError
------------------------------ Captured log setup ------------------------------
INFO     pbench.server.api:__init__.py:57 Registering service endpoints with base URI /api/v1
INFO     pbench.server.api:users_api.py:238 New auth token registered for user drb@example.com
------------------------------ Captured log call -------------------------------
INFO     pbench.server.api:upload_api.py:59 Configured PUT temporary directory as /var/tmp/jenkins/tox/py39/tmp/popen-gw8/server-tmp0/srv/pbench/archive/fs-version-001/UPLOAD
INFO     pbench.server.api:upload_api.py:89 Uploading log.tar.xz on controller 2c96a1480aaa
INFO     pbench.server.api:upload_api.py:145 PUT uploading 2c96a1480aaa:log.tar.xz for user drb to /var/tmp/jenkins/tox/py39/tmp/popen-gw8/server-tmp0/srv/pbench/archive/fs-version-001/UPLOAD/log.tar.xz
INFO     pbench.server.api:upload_api.py:193 Uploading file 'log.tar.xz' (user = drb, ctrl = '2c96a1480aaa') to drb(3)|log
WARNING  pbench.server.api:upload_api.py:369 drb:2c96a1480aaa:log.tar.xz error MD5 checksum 9d5a479f6f75fa9b3bab27ef79ad5b29 does not match expected md5sum (None)
DEBUG    pbench.server.api:utils.py:130 beg - <bound method Path.unlink of PosixPath('/var/tmp/jenkins/tox/py39/tmp/popen-gw8/server-tmp0/srv/pbench/archive/fs-version-001/UPLOAD/log.tar.xz')>
DEBUG    pbench.server.api:utils.py:137 end - <bound method Path.unlink of PosixPath('/var/tmp/jenkins/tox/py39/tmp/popen-gw8/server-tmp0/srv/pbench/archive/fs-version-001/UPLOAD/log.tar.xz')>
DEBUG    pbench.server.api:utils.py:130 beg - <bound method Dataset.delete of <pbench.server.database.models.datasets.Dataset object at 0x7efc8a7987c0>>
DEBUG    pbench.server.api:utils.py:137 end - <bound method Dataset.delete of <pbench.server.database.models.datasets.Dataset object at 0x7efc8a7987c0>>
```

portante:
FWIW, I tried running the server unit tests locally on Fedora 36 and the passed on `main`, and then tried running both the agent and server unit tests at the same time and they passed.  So if this is a flaky test it will be fun to find.

portante:
Trying it locally via the Jenkins `./build.sh` to see if I can repeat the failure...

portante:
Filed PR #3022 with the fix (thanks @webbnh!!!).

lib/pbench/agent/tool_data_sink.py
Maxusmusti: Used to exit the listen() for loop

portante: Right but the previous `for` loop did not need a break because in `execute_action()` the "terminate" action closed the `_from_client_chan`.  Can the same be done so that the generator returned by `.listen()` will exit gracefully?

lib/pbench/agent/tool_data_sink.py
portante: The jist of the above response needs to be added to the code comment above, thanks!

portante (changes requested):
Looking good.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1530,6 +1528,8 @@ def execute_action(self, client: str, data: Dict[str, Any]):
 
             self._from_tms_chan.close()
             self._to_logging_chan.unsubscribe()
+            self.logger.info("Tool Data Sink terminating")
```
portante, line 1531: Bugger all, this log point results in a different function name in the test output which causes them to fail.  Just need a few tests updated.

portante (changes requested):

webbnh (approved):
👍

lib/pbench/agent/tool_data_sink.py
portante: See https://github.com/Maxusmusti/pbench/pull/12.

lib/pbench/agent/tool_data_sink.py
portante: See https://github.com/Maxusmusti/pbench/pull/12.

portante (approved):

lib/pbench/agent/tool_meister_client.py
```diff
@@ -87,149 +83,126 @@ def __init__(
...
         # is accessible by the Tool Data Sink instance.
-        self.logger.debug("publish %s on chan %s", action, self.from_client_channel)
-        msg = dict(action=action, group=group, directory=str(directory), args=args)
+
+        metadata = dict(
+            group=group,
+            # This check/conversion is necessary, as the pbench scripts
+            # sometimes pass in the directory as a str, and sometimes as
+            # pathlib.PosixPath
+            directory=None if directory is None else str(directory),
```
dbutenhof, line 163: The "check" is presumably necessary because sometimes `directory is None` and we don't want to pass `str(None)`; the "conversion" is because we're passing a "path-like object" which might be `str` or `Path`. (I.e., the comment claims to explain "check/conversion" but actually only explains "conversion".)

lib/pbench/agent/tool_meister_client.py
```diff
@@ -87,149 +83,126 @@ def __init__(
...
+        """On context exit, just close down the to SignalExporter object."""
+        self.sig_pub.shutdown()
 
-    def publish(self, group, directory, action, args=None):
-        """publish - marshal a JSON document formed from the group, directory,
-        action, and args arguments.
+    def publish(
+        self,
+        group: str,
+        directory: Union[str, Path, None],
```
dbutenhof, line 134: That's OK, although I would have said `Optional[Union[str, Path]]` which I think is a bit more clear.

lib/pbench/agent/tool_meister_client.py
```diff
@@ -87,149 +83,126 @@ def __init__(
...
+        action: str,
+        args: Any = None,
```
dbutenhof, line 136: I suppose that `Any` includes `None` but somehow this still looks odd and maybe `Optional[Any]` would be more clear?

dbutenhof (dismissed):
I made a few comments but I don't know that any are really "problems"; still, I logged just for consideration.

lib/pbench/agent/tool_meister_client.py
portante: See if you like https://github.com/Maxusmusti/pbench/pull/13.

lib/pbench/agent/tool_meister_client.py
portante: https://github.com/Maxusmusti/pbench/pull/13

lib/pbench/agent/tool_meister_client.py
dbutenhof: Yeah, I like Peter's version a bit better, but I'm fine with this having been marked resolved (although we generally prefer to leave that step to the author of the comment).

dbutenhof (approved):

webbnh (commented):
👍 

(Can we merge this, now?  🤣)

webbnh (approved):
👍 

(I was having so much trouble spelling rofl that I forgot to tick "Approve"....)

dbutenhof (approved):
💥

PR #2474, opened 2021-10-05
Moar jenkins container stuff

webbnh:
These changes represent the next step along the path to producing a comprehensive containerized testing environment:
- Organize the container definition into separate steps for the Agent, Server, lint, testing, and other dependencies, which will hopefully allow them to be maintained more easily.  The sections for the Agent and Server can be generated from the `dnf repoquery --requires` command, so that they can easily be kept in sync with the requirements of the Agent and Server RPMs.
 - Update the Server's RPM spec file template, to add a number of dependencies which were missing, based on the contents of the Server's `requirements.txt` file.
- Update a couple of the dependencies in the Servers `requirements.txt` file which have specific version requirements.

The development flow here is to run the container, install the Agent and Server RPMs, and then do a `pip` install of the various `requirements.txt` files.  The expectation is that the RPM installs will find all of their dependencies already installed in the container.  The goal is for the `pip` install to find that all of _its_ dependencies are also already installed in the container.

If the Agent or Server RPM install finds a dependency missing, this means that it was added to the RPM spec file template without also adding it to the container definition, and the installation list in the container definition should be regenerated from the up-to-date RPM.

If the `pip` install finds a dependency missing, it represents one of several issues:
- The dependency was added to the `requirements.txt` file without also being added to the RPM spec file template, in which case the spec file template (and the container definition) should be updated accordingly.
- The dependency is not available in RPM form, in which case we should give careful consideration to the addition, as, at least for the Agent, we will ultimately need to have all dependencies packaged as RPMs, one way or another.
- The dependency is on a version which does not match the version in the distro on which the container was built:
  - If the dependency is on an older version, as is the case with Black and Flake8, we'll live with the `pip`-installed version, although we should consider retooling to use the current version of the dependency (e.g., PyTests).
  - If the dependency is on a newer version, then we have little choice; however, this problem should eventually go away as we migrate to newer versions of the distro, although we should be cognizant of our compatibility story and avoid reliance on software versions which are not available in our targeted distros.

Ultimately, our testing framework should check the RPM and `pip` installs to ensure that there are no dependencies unexpectedly missing.  (This is future work.)

Reviewers:  please feel free to point out any dependencies which are questionable or obsolete.
@portante, @ndokos, @dbutenhof


server/requirements.txt
```diff
@@ -4,10 +4,10 @@ boto3
 click>=8.0
 python-dateutil
 elasticsearch==7.13.1
```
webbnh, line 6: I think that this should read
```
elasticsearch>=7.13.1, !=7.14.0
```
Anyone (@dbutenhof?) concur?

dbutenhof: It's certainly a valid experiment. 7.13.1 was selected simply because it was the version successfully running on my development VM, and the 7.14.0 our jam session install pulled clearly was broken. It's likely but not guaranteed that any 7.14.x or 7.15.x will fix the problem, and at least reasonable to determine whether higher versions of 7.13.x work. I just didn't want to deal with that at the time!

dbutenhof (commented):
Quick comment, but I'll look through the rest of this after my hike!

portante (dismissed):
Looks good, just a note that multiple `RUN` lead to large images for some reason.

jenkins/development.Dockerfile
```diff
@@ -60,60 +65,131 @@ RUN \
...
         python3-responses \
-        python3-s3transfer \
-        python3-sh \
-        python3-smmap \
-        python3-sphinx \
-        python3-tox \
-        python3-tox-current-env \
-        python3-typed_ast \
+        \
+        python3-atomicwrites \
```
dbutenhof, line 158: Should this be in alphabetical order? Or is the "almost blank line" intended to separate this into sections? And, if so, can you document the significance of these sections?

dbutenhof (approved):
I'm not entirely comfortable with the idea of manually/statically trying to do `dnf`'s job by tracking indirect dependencies, but I'll roll with it if that's the way people want to go (even knowing it'll probably bite us in the future). Aside from that philosophical issue, a few minor comments and questions ...

server/requirements.txt
webbnh: Yeah, I tried this and it pulled in 7.15...  I'm hoping that we can determine whether that's viable at tomorrow's Jam session.

dbutenhof (approved):
Although not 100% convinced on several of these tradeoffs, I think it's worth giving it a shot.

PR #2475, opened 2021-10-05
More performance improvements

portante:


portante (commented):
Lots of change but starts to speed things up a bit.

PR #2476, opened 2021-10-06
Remove explicit removals from RPM removal

### LINKED ISSUE #2042: Uninstalling the pbench server deletes the whole /opt/pbench-server directory
`yum erase pbench-server` removes the whole `/opt/pbench-server` directory, including things that are not put there by installing the RPM (e.g. the `pbench-server.cfg` file). The RPM should be restructured not to touch those.

webbnh:
This change touches up the RPM spec file template for the Server to enable `dnf remove` to work properly.  Previously, we used the "big hammer" approach of simply (manually/explicitly) removing the entire `%{installdir}` during the `%postun` stage, but this has the undesirable side effect of wiping out any user files which might be in `/opt/pbench-server/` (such as `pbench-server.cfg`).  Instead, we now use better discipline, removing during the `%preun` stage the files which we created during the `%post` stage, so that, during the "uninstall" stage, the directory tree looks like what RPM expects it to look like, which means that all of the entries in the `%files` list are removed implicitly for us (but, things like user files are left alone).

This change also declares the entire `%{installdir}` to belong to this package, so, if there are no user files in it, then the entire directory tree is removed on uninstall.  And, that means that a number of files no longer need to be explicitly listed in the `%files` section (because they are already in the tree).  However,  the RPM build now issues more `File listed twice` warnings, because the files from entire `bin` subdirectory are also listed under the top-level directory list (but with different attributes).  (There doesn't seem to be a good way to avoid double listings -- the recommended approach is to dynamically generate the files list, but I concluded that ignoring the warnings was a better way to go.)

This PR also contains several smaller fixes:
- It looked like the `%doc` directive was being mis-used, so I guessed at how we wanted it to be applied.  However, I now cannot find where RPM is putting those files, but maybe that's a feature?  I used the `%license` directive for the license files, and they behave reasonably, so I think RPM may be configured to drop documentation on my test platform.
- I added a file to suppress the NPM message about the `package-lock.json` file.  Since we're not distributing it, and since we're not doing development on this platform or repeated NPM deployments, there's no reason to generate the file or to complain about its absence.
- I added a switch to the `pip` install to suppress the `PATH` warning, since it's not germane in this context.
- And, I fixed a typo in the Agent spec file template.  (I was going to fix it up like I fixed up the Server one, but it took so much effort to do the Server one that I've kind of lost interest....)

Fixes #2042.

Reviewers:  any thoughts about what to do about the two items under the `FIXME` comment?
@portante, @ndokos, @dbutenhof

### DIFF
--- a/server/rpm/pbench-server.spec.j2
+++ b/server/rpm/pbench-server.spec.j2
@@ -63,7 +63,7 @@ mv %{buildroot}/%{installdir}/%{static}/package.json %{buildroot}/%{installdir}
 
 %post
 # Install python dependencies as pbench user into user's site-packages
-(su pbench -c "pip3 --no-cache-dir install --user -r /%{installdir}/requirements.txt") > /%{installdir}/pip3-install.log
+(su pbench -c "pip3 --no-cache-dir install --user --no-warn-script-location -r /%{installdir}/requirements.txt") > /%{installdir}/pip3-install.log
 
 # The `site` package is Python magic; it runs automatically when Python starts,
 # and builds `sys.path`.
@@ -89,6 +89,7 @@ su pbench -c "echo /%{installdir}/lib > \$(python3 -m site --user-site)/pbench.p
 # install node.js modules under /%{installdir}
 cd /%{installdir}
 rm -rf node_modules
+echo 'package-lock=false' >> .npmrc
 npm install
 
 # this only handles v0.3
@@ -112,6 +113,16 @@ chown -R pbench.pbench /%{installdir}
 
 %preun
 
+rm -rf \
+    "$(python3 -m site --user-site)/pbench.pth"       \
+    /%{installdir}/node_modules                       \
+    /%{installdir}/.npmrc                             \
+    /%{installdir}/package-lock.json                  \
+    /%{installdir}/pip3-install.log                   \
+    /%{installdir}/%{static}/js/v0.3/d3.min.js        \
+    /%{installdir}/%{static}/js/v0.3/d3-queue.min.js  \
+    /%{installdir}/%{static}/js/v0.3/saveSvgAsPng.js
+
 %postun
 # if uninstalling, rather than updating, remove everything
 if [ $1 -eq 0 ] ;then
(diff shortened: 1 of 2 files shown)


server/rpm/pbench-server.spec.j2
```diff
@@ -112,39 +113,42 @@ chown -R pbench.pbench /%{installdir}
 
 %preun
 
+# FIXME:  what about these?
+#   - the results of the `pip3 install --user`
+#   - the pbench user account
```
dbutenhof, line 118: Interesting question. Historically, the `.spec` file *depended* on the `pbench` user account existing, but it was actually created by an Ansible role. However, because we did depend on it, I added `.spec` code to create it to be sure it was there; and maybe that means we should also *delete* it on uninstall? And, if so, that would take with it all of the `--user` Python modules.

We could extend this cleanup discussion to the archive/result directory trees; arguably the archive tree implicitly belongs to the user, and contains potentially valuable and irreplacable user data (the raw tarballs), but the unpacked results and links are arguably ours.

I think for right now I'd be inclined to say that we should drop the pbench user/group but leave the archive/result trees alone ... except they're *owned* by pbench, which means we'd probably need to `chown` them all first to ... `root`??

Maybe: can we determine any standard Fedora practice for installers that create accounts or groups? Is it written down in the packaging rules?

dbutenhof (approved):
OK, this all looks plausible, aside from your FIX-ME, which we probably could ignore and tackle as a separate pass.

This is where I wish we had a functional test chain, though; wouldn't it be great to have Mr Jenkins stand up a server using our standard installer toolchain (ansible, dnf), validate the installed filesystem and server state, then uninstall and validate again?

ndokos (approved):
AFAIAC, you can resolve the conversation. I don't think that this PR needs to do any of that. Any subsequent actions can be captured in a new issue.

server/rpm/pbench-server.spec.j2
webbnh: Tracked in #2481.

dbutenhof (approved):

Issue #2470, opened 2021-09-30
Server API: queryIterationSampleRows


dbutenhof:
We've made a decision to try to break up the current `queryIterationSamples` into two simpler APIs.

The data is current used for two purposes:

1. An aggregated query against a run.id (or potentially a list) to compile the filter header which shows "every" field in the "benchmark" and "sample" subdocuments with all the actual values for each;
2. A query on the run.id leveraging selected filter values from (1) to generate a set of aggregated rows displaying data.

This query will implement (2); given a run document id value, and specific filter values chosen from the fields and values represented in (1), run an aggregated Elasticsearch query reporting the data represented by row in the Summary table.

PR #2478, opened 2021-10-06
Unique iteration sample values for each sub-document field in result-data-sample documents

npalaska:
Addresses issue https://github.com/distributed-system-analysis/pbench/issues/2470

Returns a list of aggregated unique values for keyword fields in the `benchmark`, `sample`, `iterations` and other sub-documents of result-data-sample index.

1. Get all the fields from `result-data-sample` index that are of type keyword
2. Builds the Elasticsearch `term_aggregation` query based on generated list of fields of type `keyword` 
3. Returns a dictionary where the keys are dot separated fully qualified field names and values are a list of unique values of keys from the entire document across all the ES shards. 

Example:
```
{
               "authorization.access":[],
               "authorization.owner":[],
               "benchmark.bs":[],
               "benchmark.iodepth":[],
               "benchmark.ioengine":[],
               "benchmark.name":["uperf"],
               "benchmark.negative_packet_loss_mode":[],
               "benchmark.numjobs":[],
               "benchmark.primary_metric":["Gb_sec"],
               "benchmark.protocol":["tcp"],
               "benchmark.uid":[
                  "benchmark_name:uperf-controller_host:hostname.com"
               ],
               "benchmark.uid_tmpl":[
                  "benchmark_name:%benchmark_name%-controller_host:%controller_host%"
               ],
               "iteration.name":[
                  "1-tcp_stream-131072B-2i",
                  "1-tcp_stream-131072B-2i-fail1",
                  "1-tcp_stream-131072B-2i-fail2",
                  "1-tcp_stream-131072B-2i-fail3",
                  "1-tcp_stream-131072B-2i-fail4"
               ],
               "run.config":["run-config"],
               "run.controller":["controller-name.com"],
               "run.id":["f3a37c9891a78886639e3bc00e3c5c4e"],
               "run.name":["uperf"],
               "run.script":["uperf"],
               "run.toolsgroup":[],
               "sample.client_hostname":[
                  "127.0.0.1",
                  "all"
               ],
               "sample.field":[],
               "sample.group":[],
               "sample.measurement_title.raw":["Gb_sec"],
               "sample.measurement_type":["throughput"],
               "sample.name":[
                  "sample1",
                  "sample2",
                  "sample3",
                  "sample4",
                  "sample5"
               ],
               "sample.pgid":[],
               "sample.server_hostname":[
                  "127.0.0.1",
                  "all"
               ],
               "sample.uid":[
                  "client_hostname:127.0.0.1-server_hostname:127.0.0.1-server_port:20010",
                  "client_hostname:all-server_hostname:all-server_port:all"
               ],
               "sample.uid_tmpl":[
                  "client_hostname:%client_hostname%-server_hostname:%server_hostname%-server_port:%server_port%"
               ]
            }
```


lib/pbench/server/api/__init__.py
```diff
@@ -82,6 +83,11 @@ def register_endpoints(api, app, config):
     api.add_resource(
         IndexSearch, f"{base_uri}/index/search", resource_class_args=(config, logger),
     )
+    api.add_resource(
+        IterationSampleRows,
+        f"{base_uri}/iterations/unique",
```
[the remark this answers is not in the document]
npalaska: Yes, fixed

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,271 @@
...
+               ],
+               "sample.uid_tmpl":[
+                  "client_hostname:%client_hostname%-server_hostname:%server_hostname%-server_port:%server_port%"
+               ]
+            }
+        """
+        try:
+            es_json_aggs = es_json["aggregations"]
+            new_json = {}
+            for key in as_json_aggs:
```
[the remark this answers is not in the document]
npalaska: But do we want to suppress empty entries? I suppose no right, since these keys can be returned by the client for further aggregation.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,271 @@
...
+        {
+            "start": "start-time",
+            "end": "end-time",
+            "run_id": "run id string"
+        }
+        json_data: JSON dictionary of type-normalized parameters
+            "start" and "end" are string representation of datetime objects to search within.
+            "run_id": String representation of run id
+        """
+        start = json_data.get("start")
```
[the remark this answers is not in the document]
npalaska: It can be in the metadata table since it has a foreign key that points to the dataset entry. So if we can get the dataset from run_id, we can find the metadata related to that dataset and get the index stored at `server.index-map`

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,271 @@
...
+        if not owner:
+            self.logger.error(
+                "Dataset owner ID {} cannot be found in Users", dataset.owner_id
+            )
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="Dataset owner not found")
+
+        # For Iteration samples, we check authorization against the ownership of the
+        # dataset that was selected rather than having an explicit "user"
+        # JSON parameter. This will raise UnauthorizedAccess on failure.
+        self._check_authorization(owner.username, dataset.access)
```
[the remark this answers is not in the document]
dbutenhof: `ElasticBase` understands that `preprocess()` might raise `UnauthorizedAccess`, and handles it appropriately.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,271 @@
...
+
+    def postprocess(self, es_json: JSON, context: CONTEXT) -> JSON:
+        """
+        Returns a list of aggregated unique values for keyword fields in the
+        'benchmark', 'sample', 'iterations' and other subdocuments of result-data-sample index.
+
+        Example:
+            {
+               "authorization.access":[],
+               "authorization.owner":[],
```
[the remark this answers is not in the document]
dbutenhof: I would definitely prefer not to return the `authorization` subdocument or any other "bookkeeping" stuff. I'm OK with a blacklist or a whitelist; but I can see some advantage in paring it down to a clear whitelist of the things we actually want a client to be able to query in this context.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -67,17 +68,33 @@ def build_index(self, server_config, dates):
...
-        return index
+        if not index:
+            drb = Dataset.attach(controller="node", name="drb")
+            index_map = Metadata.getvalue(dataset=drb, key="server.index-map")
+            for key, values in index_map.items():
+                if self.payload["run_id"] in values:
+                    index = f"{key}"
+        return f"/{index}"
+
+    def build_index_from_metadata(self, find_metadata):
```
webbnh, line 82: I don't see a reference to `find_metadata` in this function; can it/should it be removed?

I see only one call to `build_index_from_metadata()`, and it's neglecting to specify a second parameter....

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -67,17 +68,33 @@ def build_index(self, server_config, dates):
...
+        drb = Dataset.attach(controller="node", name="drb")
+        index_map = Metadata.getvalue(dataset=drb, key="server.index-map")
+        for key, values in index_map.items():
+            if self.payload["run_id"] in values:
+                return "/" + key
```
webbnh, line 87: This looks like half the guts of `build_index()`...should it be calling this?

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -67,17 +68,33 @@ def build_index(self, server_config, dates):
...
+        return None
 
     def date_range(self, start: AnyStr, end: AnyStr) -> list:
         """
         Builds list of range of dates between start and end
         It expects the date to look like YYYY-MM
         """
         date_range = []
+        if not start or not end:
+            return date_range
```
webbnh, line 97: If what you want to do here is to return an empty list if either `start` or `end` is blank or omitted, then you should just do that.  Returning the value of `data_range` accomplishes the same thing, but it's slightly obfuscating for the reader, and doing the test first saves Python from having to set up `date-range` if you're going to bail out early.
```
        if not start or not end:
            return []
        date_range = []
```

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -248,7 +264,8 @@ def test_empty_query(
             expected_status = HTTPStatus.FORBIDDEN
 
         index = self.build_index(
-            server_config, self.date_range(self.payload["start"], self.payload["end"])
+            server_config,
+            dates=self.date_range(self.payload.get("start"), self.payload.get("end")),
         )
```
webbnh, line 269: If you're going to use a keyword for one of the arguments here, you might as well do it for both.  However, I would just get rid of the parameter definition that you're not using (or, at least, move it to the end of the signature) and then omit the keywords.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -300,7 +317,9 @@ def test_http_exception(
             pytest.skip("skipping " + self.test_http_exception.__name__)
 
         index = self.build_index(
-            server_config, self.date_range(self.payload["start"], self.payload["end"])
+            server_config,
+            provide_metadata,
```
webbnh, line 321: As I said above, I don't think you need to supply `provide_metadata` here as an argument.

Ditto line 354.

lib/pbench/test/unit/server/query_apis/test_controllers_list.py
```diff
@@ -91,7 +91,8 @@ def test_query(
         }
 
         index = self.build_index(
-            server_config, self.date_range(self.payload["start"], self.payload["end"])
+            server_config,
+            dates=self.date_range(self.payload["start"], self.payload["end"]),
         )
```
webbnh, line 96: See previous comment.

lib/pbench/test/unit/server/query_apis/test_datasets_detail.py
```diff
@@ -120,7 +120,8 @@ def test_query(
         }
 
         index = self.build_index(
-            server_config, self.date_range(self.payload["start"], self.payload["end"])
+            server_config,
+            dates=self.date_range(self.payload["start"], self.payload["end"]),
         )
```
webbnh, line 125: See previous comment.

Ditto for the other calls to `build_index` in this module.

lib/pbench/test/unit/server/query_apis/test_datasets_list.py
```diff
@@ -92,7 +92,8 @@ def test_query(
         }
 
         index = self.build_index(
-            server_config, self.date_range(self.payload["start"], self.payload["end"])
+            server_config,
+            dates=self.date_range(self.payload["start"], self.payload["end"]),
         )
```
webbnh, line 97: See previous comment.

Ditto for the other call to build_index in this module.

lib/pbench/test/unit/server/query_apis/test_search.py
```diff
@@ -123,7 +123,7 @@ def test_query(
             },
         }
         index = self.build_index(
-            server_config, self.date_range(payload["start"], payload["end"])
+            server_config, dates=self.date_range(payload["start"], payload["end"])
         )
```
webbnh, line 127: See previous comment.

webbnh (changes requested):
More code, more comments.  🙂

- I suspect that the code which fetches the index from the run metadata can be made much more straightforward
- The code attendant to it looks like it has a small bug
- The `build_index()` and `build_index_from_metadata()` test helper functions look like they need some tweaking and possibly refactoring

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -89,88 +85,77 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
...
 
-        uri_fragment = self._gen_month_range("result-data-sample", start, end)
-        self.logger.info("fragment, {}", uri_fragment)
+        # Retrieve the ES indix that belongs to this run_id from the metadata table
+        try:
+            index_map = Metadata.getvalue(dataset=dataset, key="server.index-map")
+            for key, values in index_map.items():
+                if run_id in values:
+                    index = key
+                    self.logger.debug("Iteration samples index, {}", index)
```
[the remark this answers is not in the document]
npalaska: Yes I need a break, we don't need to keep looping if we already find the index. The server value of the metadata looks like this:
```
server = {
   "reindex":false,
   "index-map":{
      "pbench-unittests.v6.run-data.2018-02":[
         "4689ac905cf35910d5d110ea6724d5f3"
      ],
      "pbench-unittests.v6.run-toc.2018-02":[
         "43ea6a5f442f6daa1b3431488e4b68ef",
         "7a03e52ff28e8296d28df67fc9fde313"
      ],
      "pbench-unittests.v5.result-data-sample.2018-02-02":[
         "b2e3445bf15bf9fdb7342d4f974a95b9"
      ]
   }
}
```
index-map is a key value pair where key is an index name and value is a list of run_ids that belong to that index.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -50,26 +53,48 @@ def _setup(
         self.bad_date_payload = bad_date_payload
         self.error_payload = error_payload
         self.empty_es_response_payload = empty_es_response_payload
+        self.index_prefix = index_prefix
+        self.index_version = index_version
 
-    def build_index(self, server_config, dates):
+    def build_index(self, server_config=None, find_metadata=None, dates=[]):
```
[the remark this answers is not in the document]
npalaska: Yeah I did this to build the index in one place, instead of having 2 separate functions. Otherwise we would have to modify other test files by adding an `if-else` statements to check if the index need to built using the metadata or the client provided dates. e.g some of our common test cases like `test_http_error` or `test_http_exception` need to determine if they should get the index from metadata or from dates as our previous API still depends on start and end dates.

lib/pbench/test/unit/server/query_apis/commons.py
npalaska: Fixed

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -67,17 +68,33 @@ def build_index(self, server_config, dates):
...
+        index = ""
         for d in dates:
             index += f"{idx}{d},"
-        return index
+        if not index:
+            drb = Dataset.attach(controller="node", name="drb")
+            index_map = Metadata.getvalue(dataset=drb, key="server.index-map")
+            for key, values in index_map.items():
+                if self.payload["run_id"] in values:
+                    index = f"{key}"
```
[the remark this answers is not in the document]
npalaska: I refactored it, so its no longer used

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -89,88 +85,77 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
...
+        except MetadataError as e:
+            abort(HTTPStatus.BAD_REQUEST, message=str(e))
+
+        if index_map is None:
+            abort(
+                HTTPStatus.BAD_REQUEST,
+                message=f"Dataset with run_id: {run_id} not found",
+            )
```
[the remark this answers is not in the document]
npalaska: Right, I should check the `index` value instead of `index_map`

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -50,26 +53,48 @@ def _setup(
         self.bad_date_payload = bad_date_payload
         self.error_payload = error_payload
         self.empty_es_response_payload = empty_es_response_payload
+        self.index_prefix = index_prefix
+        self.index_version = index_version
```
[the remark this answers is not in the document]
npalaska: Yeah I can totally remove these 2 since they no longer are useful as we extracting index from the metadata and all other APIs currently only use `run-data` prefix.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -50,26 +53,48 @@ def _setup(
...
+        return None
 
     def date_range(self, start: AnyStr, end: AnyStr) -> list:
         """
         Builds list of range of dates between start and end
         It expects the date to look like YYYY-MM
         """
         date_range = []
+        if not start or not end:
+            return date_range
```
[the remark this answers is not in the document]
npalaska: Fixed

lib/pbench/test/unit/server/query_apis/commons.py
webbnh: I get that we need to specify a reference to the `provide_metadata` fixture in the function signature (e.g., at line 311), so that the fixture will be executed and set up the metadata, but I don't see how providing it here (at line 321) accomplishes anything.

What we require is a side-effect of the fixture's execution; the fixture doesn't yield any explicit value, so referencing it inside the function presumably just produces a `None` value (or whatever the last statement in the the `provide_metadata()` function evaluates to).

lib/pbench/test/unit/server/conftest.py
```diff
@@ -289,6 +295,11 @@ def provide_metadata(attach_dataset):
     test = Dataset.attach(controller="node", name="test")
     Metadata.setvalue(dataset=drb, key="user.contact", value="me@example.com")
     Metadata.setvalue(dataset=drb, key=Metadata.DELETION, value="2022-12-25")
+    Metadata.setvalue(
+        dataset=drb,
+        key="server.index-map",
+        value={"unit-test.v5.result-data-sample.2020-08": ["random_md5_string1"]},
```
dbutenhof, line 301: Again, for some semblance of reality, the "random_md5_string1" shouldn't be the `run_id` MD5. That value goes under the `unit-test.v6.run-data.2020-08` index.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -8,6 +8,7 @@
 from typing import AnyStr, Type
 
 from pbench.server.api.resources import JSON, ParamType
+from pbench.server.database.models.datasets import Dataset, Metadata
```
dbutenhof, line 11: Out of order!

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -59,10 +60,17 @@ def build_index(self, server_config, dates):
...
+        index = "".join([f"{idx}{d}," for d in dates])
+
+        return f"/{index}"
+
+    def build_index_from_metadata(self, provide_metadata):
+        drb = Dataset.attach(controller="node", name="drb")
+        index_map = Metadata.getvalue(dataset=drb, key="server.index-map")
+        for key, values in index_map.items():
+            if self.payload["run_id"] in values:
+                return "/" + key
```
dbutenhof, line 72: Again, no! 😬

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -282,16 +291,22 @@ def test_http_exception(
...
         """
         Check that an exception in calling Elasticsearch is reported correctly.
         """
         if not self.elastic_endpoint:
             pytest.skip("skipping " + self.test_http_exception.__name__)
 
-        index = self.build_index(
-            server_config, self.date_range(self.payload["start"], self.payload["end"])
-        )
+        if self.payload.get("start") and self.payload.get("end"):
```
dbutenhof, line 302: Although it may not matter in the cases we've seen, I think we should be more specific than this. Maybe it should be based on the class *schema* containing start/end? Or maybe it should be explicit in your test class (e.g., as a constructor parameter).

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -282,16 +291,22 @@ def test_http_exception(
...
+            index = self.build_index(
+                server_config,
+                self.date_range(self.payload["start"], self.payload["end"]),
+            )
+        else:
+            index = self.build_index_from_metadata(provide_metadata)
```
dbutenhof, line 308: I still don't get what you're accomplishing by passing the value of the fixture to your helper method. That's not a fixture, so referencing a fixture accomplishes nothing, and the code doesn't use it as a parameter. I don't think it should be here at all.

dbutenhof (changes requested):
More assorted comments. There are several serious issues that aren't tracking here, starting with the apparent misunderstanding of the document map metadata structure.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,240 @@
...
+                },
+                "params": {"ignore_unavailable": "true"},
+            },
+        }
+
+    def postprocess(self, es_json: JSON, context: CONTEXT) -> JSON:
+        """
+        Returns a non-empty list of aggregated unique values for keyword fields
+        in the 'benchmark', 'sample', 'iterations' and 'run' subdocuments of
+        result-data-sample index. The output only contains fields that has at
```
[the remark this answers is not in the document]
webbnh: Given that this API takes a run ID, it seems reasonable to me that it return results which are custom to that run.  I don't see how returning headers which have no values is going to be useful.  If the client really wants _that_, perhaps they should be querying the index mapping instead of a particular dataset.

lib/pbench/test/unit/server/query_apis/commons.py
webbnh: (I.e., you need to swap lines 11 and 12.... 😁)

lib/pbench/test/unit/server/query_apis/commons.py
webbnh: Right:  as I understand it, referencing a PyTest "fixture" is only significant if it is referenced in the function signature of a "test" (i.e., a function whose name begins with `test_`) or another fixture; inside a test or fixture which references it in its signature, referencing a fixture produces the cached value which was originally returned (or `yield`ed) by the fixture.

Since `provide_metadata` doesn't yield or return a value, the reference in the cited code will produce `None` (or whatever value was produced by the last statement in the fixture), and it won't have any other effect.  So, there's no point in referencing it here.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,240 @@
...
+        )
+
+        # Retrieve the ES index that belongs to this run_id from the metadata
+        # table
+        try:
+            index_map = Metadata.getvalue(dataset=dataset, key="server.index-map")
+            for key, values in index_map.items():
+                if run_id in values:
+                    index = key
+                    self.logger.debug(f"Iteration samples index, {index!r}")
```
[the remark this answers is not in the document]
dbutenhof: > Okay so we only need to match keys for this pattern `*v5.result-data-sample*` and if we find multiple such matches we concatenate them as a comma separated indices for our search just the way we were doing with start and end argument.

The date pattern of an index is locked in by a single timestamp from the top-level run document, and the version number of the currently registered document template; neither can vary across indices for a given dataset.

That is, there can only be a single `result-data-sample` root index in any dataset document map. So *the* first-level key matching `result-data-sample` is the one you want.

lib/pbench/test/unit/server/conftest.py
dbutenhof: To reiterate: it's the `<prefix>.<version>.result-data-sample.<date>` pattern of the index you need to match, not the values in its list of documents.

lib/pbench/test/unit/server/query_apis/test_iteration_samples.py
```diff
@@ -0,0 +1,317 @@
...
+        build_auth_header,
+        find_template,
+        provide_metadata,
+    ):
+        """
+        Check the construction of Elasticsearch query URI and filtering of the
+        response body. Note that the mock set up by the attach_dataset fixture
+        matches the dataset name to the dataset's owner.
+        """
+        response_payload = {
```
dbutenhof, line 40: The way we've got this structured, playing with different ES payloads in the unit tests is awkward. I haven't really come up with a better way. In any case, there was a discussion in another context about dropping fields (that appear in the document template) which have no values in this specific dataset ... are we testing that case? 😁

lib/pbench/test/unit/server/query_apis/commons.py
webbnh: Yeah...this feels weird:  if we have exactly one of start or end, we build the index from metadata?  It's not obvious how that follows logically...which makes me inclined to think that this test is wrong.  (Should the test be if we have _neither_ of them, then we use the metadata?)

lib/pbench/test/unit/server/query_apis/commons.py
webbnh: (And, @dbutenhof, don't look too closely at `lib/pbench/server/api/__init__.py`.... 🤣)

webbnh (changes requested):
Coming along nicely...but, of course, I have comments.  The more substantive of them are:
- I think that the code which fetches the index from the run metadata still has a small bug
- LGTM agrees with Dave that you should fix the existing instantiations of `DatasetSqlError`
- You still need to get the `provide_metadata` usages straightened out

lib/pbench/server/api/resources/query_apis/iteration_samples.py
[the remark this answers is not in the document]
webbnh: So....
```
            index_keys = [key for key in index_map if "result-data-sample" in key]
            assert len(index_keys) == 1
            index = index_keys[0]
            self.logger.debug(f"Iteration samples index, {index!r}")
```

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,240 @@
...
+    ParamType,
+    PostprocessError,
+)
+from pbench.server.api.resources.query_apis import CONTEXT, ElasticBase
+from pbench.server.database.models.datasets import Dataset, Metadata, MetadataError
+from pbench.server.database.models.users import User
+from pbench.server.database.models.template import Template, TemplateNotFound
+
+
+class IterationSamples(ElasticBase):
```
dbutenhof, line 22: It occurred to me while I was hiking ("the trees told me") that this pattern of "authorize using a dataset based on a run_id" will be repeated also for TOC and timeseries data, and isn't that different from the publish pattern (albeit with controller/name instead of run_id).

I'm back to thinking this should be "hoisted" into a common place and given a better name and perhaps even a more general structure to accommodate them.

npalaska: Interesting point, the universe is telling us to do that I guess. I believe we could refactor the parent `preprocess` method from `ElasticBase` and do the all the stuff related authorization from dataset there (since it would be common to all)

lib/pbench/test/unit/server/query_apis/commons.py
npalaska: Right, so I could use class schema to determine the if the start and end are part of required keys and then determine which function to use based on that, like following:
```
parameter_items = self.cls_obj.schema.parameters.items()
required_keys = [
    key for key, parameter in parameter_items if parameter.required
]
if all(x in required_keys for x in ["start", "end"]):
    index = self.build_index(
        server_config,
        self.date_range(self.payload["start"], self.payload["end"]),
    )
elif any(x in required_keys for x in ["start", "end"]):
    pytest.skip("skipping " + self.test_http_error.__name__)
else:
    index = self.build_index_from_metadata()
```
Or I could just use another parameter in the constructor like this:
```
use_index_from_metadata: bool = False
```
and if a class wants to use the index from metadata it can set it to true. So I guess 2nd option seems simpler to me (Unless you guys think first is more dynamic)

lib/pbench/test/unit/server/query_apis/test_iteration_samples.py
npalaska: Yes the mappings contain other properties like `timestamp_original` that has a type keyword but they are not included in the aggregation only `"run", "sample", "iteration", "benchmark"` fields are used

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,245 @@
...
+        )
+
+    def preprocess(self, client_json: JSON) -> CONTEXT:
+        """
+        Query the Dataset associated with this run id, and determine whether the
+        request is authorized for this dataset.
+
+        If the user has authorization to read the dataset, return the Dataset
+        object as CONTEXT so that the postprocess operation can proceed to get all
+        the iteration aggregations.
```
dbutenhof, line 43: That's not a very clear explanation of why/how we use `dataset` here. We're using the document index metadata to identify the index to be searched...

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,245 @@
...
+        # dataset that was selected rather than having an explicit "user"
+        # JSON parameter. This will raise UnauthorizedAccess on failure.
+        self._check_authorization(owner.username, dataset.access)
+
+        # The dataset exists, and authenticated user has enough access so continue
+        # the operation with the appropriate CONTEXT.
+        return {"dataset": dataset, "run_id": run_id}
+
+
+class IterationSampleNamespaces(IterationSamples):
```
dbutenhof, line 70: I think this a single "namespace" rather than "namespaces"...

lib/pbench/server/api/resources/query_apis/iteration_samples.py
dbutenhof: Almost makes me wonder if we might want to do something like moving these classes (plus TOC and timeseries queries) into another sub-module under `query_apis` with this in the `__init__.py` ... like maybe `run_id` ... 😁

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+    Iteration samples API that returns only unique namespaces for each
```
dbutenhof, line 72: I'm not really sure where the "unique" comes from here. We're returning all of the aggregatable (keyword) *fields* in the whitelisted subdocuments, each with a list of all actual values. I think the "unique" is at best redundant, and may imply some additional postprocessing that would not be at all obvious.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,245 @@
...
+    whitelisted result-data-sample subdocuments.
+
+    This class inherits the common IterationSamples class and builds an
+    aggregated query against a user supplied run id to compile the filter header
+    which shows every unique field in the 'benchmark', 'sample', 'iterations'
```
dbutenhof, line 77: "filter header" is a reference to one intended use by the dashboard; but the API is more generic than that, so let's not imply limitations. And, again, I don't like "unique" -- a natural reaction would be why is the list of fields "not unique" and what do we do to that list to *make* it "unique"?

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,245 @@
...
+
+        def recurse(prefix, mappings, result):
+            if "properties" in mappings:
+                for p, m in mappings["properties"].items():
+                    recurse(f"{prefix}{p}.", m, result)
+            elif mappings.get("type") == "keyword":
+                result.append(prefix[:-1])  # Remove the trailing dot, if any
+            elif mappings.get("type") == "text":
+                for f, v in mappings.get("fields", {}).items():
+                    recurse(f"{prefix}{f}.", v, result)
```
dbutenhof, line 95: Should this just be an `else`? I wonder if it's possible to have other "primary" field types with `keyword` "fields" underneath? Like `date`, `ip`, etc.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,245 @@
...
+        """
+        Construct a pbench search query for aggregating unique values for keyword
+        fields in the 'benchmark', 'sample', 'iterations' and 'run'
+        sub-documents of result-data-sample index document that belong to the
+        given run id.
+        {
+            "run_id": "run id string"
+        }
+        json_data: JSON dictionary of type-normalized parameters
+            "run_id": String representation of run id
```
dbutenhof, line 110: This could benefit from some blank lines to separate the sections.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,245 @@
...
+        """
+        run_id = context.get("run_id")
+        dataset = context.get("dataset")
+
+        self.logger.info(
+            "Return sample iteration namespaces for user {}, prefix {}: for "
+            "run id: {}",
+            Auth.token_auth.current_user().username,
```
dbutenhof, line 118: I'm not sure specifying the authenticated user here is what you really want. Or at least the "for user {}" is misleading in this context, as I think that looks like what we might otherwise report for the `"user"` JSON parameter. I suspect that identifying the *dataset* would be more useful.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -102,48 +118,39 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
...
             for key, values in index_map.items():
                 if run_id in values:
                     index = key
-                    self.logger.debug("Iteration samples index, {}", index)
+                    self.logger.debug(f"Iteration samples index, {index!r}")
         except MetadataError as e:
             abort(HTTPStatus.BAD_REQUEST, message=str(e))
 
-        if index_map is None:
+        if index is None:
```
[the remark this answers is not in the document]
dbutenhof: How would it skip over the assignment? A `MetadataError` aborts the API call, another exception will propagate, and there are no conditionals ...

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,245 @@
...
+            self.prefix,
+            run_id,
+        )
+
+        # Retrieve the ES index that belongs to this run_id from the metadata
+        # table
+        try:
+            index_map = Metadata.getvalue(dataset=dataset, key="server.index-map")
+            index_keys = [key for key in index_map if "result-data-sample" in key]
+            assert len(index_keys) == 1
```
dbutenhof, line 128: I don't really like the `assert` in server code, because it'll take down the server process. The way gunicorn works I think our server instances are actually single-threaded, but I'm not sure the gunicorn instance would be restarted unless the root gunicorn process crashes, so I think we've just reduced our parallelism permanently. This should be an `abort` with an internal error and a log to tell us what happened.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -59,10 +62,16 @@ def build_index(self, server_config, dates):
...
-        return index
+        index = "".join([f"{idx}{d}," for d in dates])
+
+        return f"/{index}"
+
+    def build_index_from_metadata(self):
+        drb = Dataset.attach(controller="node", name="drb")
+        index_map = Metadata.getvalue(dataset=drb, key="server.index-map")
+        index_keys = [key for key in index_map if "result-data-sample" in key]
+        assert len(index_keys) == 1
```
dbutenhof, line 73: Different problem than in the production server code, but this `assert` is also ... awkward ... since it means that the mocked document index is set up incorrectly. If that's a case we want to cover in a unit test, we'd be looking for the API to fail with an internal server error (not an `assert`) and otherwise this isn't going to be really helpful to someone who hits it in testing. This might be a case where the optional assert message would help: this means that the document index map has a duplicate root index name!

lib/pbench/test/unit/server/query_apis/test_iteration_samples.py
```diff
@@ -0,0 +1,320 @@
...
+from pbench.server.api.resources.query_apis.iteration_samples import (
+    IterationSampleNamespaces,
+)
+from pbench.test.unit.server.headertypes import HeaderTypes
+from pbench.test.unit.server.query_apis.commons import Commons
+
+
+class TestIterationSamplesRows(Commons):
+    """
+    Unit testing for IterationSamplesRows class.
```
dbutenhof, line 12: Class name and comment are wrong; we're testing the "namespace" API, not the "rows" API!

lib/pbench/test/unit/server/query_apis/test_iteration_samples.py
dbutenhof: OK, that wasn't helpful; but it seems that your sample Elasticsearch response does show empty buckets, and your response comparison would fail if they were present, so I think this is good.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
npalaska: As you said we are returning aggregatable fields in the whitelisted subdocuments, each with a list of all actual values, but these actual values are the bucket keys from the ES query result and so they are ought to be unique. So yeah the unique word could be just redundant but I mentioned to in the docstring if someone doesnt know that these values represent (bucket keys from ES)

lib/pbench/server/api/resources/query_apis/iteration_samples.py
npalaska: That is an interesting point and I think that make sense. We could make primary fields like IP a keyword by adding a `fields` underneath it.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
dbutenhof: The word "unique" is used so often and so casually as to have become meaningless, and it is simply not helpful in this context.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -67,20 +66,19 @@ def preprocess(self, client_json: JSON) -> CONTEXT:
         return {"dataset": dataset, "run_id": run_id}
 
 
-class IterationSampleRows(IterationSamples):
+class IterationSampleNamespace(IterationSamples):
     """
-    Iteration samples API that returns only unique rows for each subdocument.
+    Iteration samples API that returns namespaces for each whitelisted
+    result-data-sample subdocuments.
```
webbnh, line 72: I think we're thinking that the "namespace" contains the list of what the Dashboard is going to use a column headers and filter keys (rather than thinking of each key as a namespace for which we are returning the available values).  So, we're returning the names from the "space" as well as lists of available values for each name.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -67,20 +66,19 @@ def preprocess(self, client_json: JSON) -> CONTEXT:
...
     """
 
     def __init__(self, config: PbenchServerConfig, logger: Logger):
         super().__init__(config, logger)
 
-    def get_keyword_fields(self, mappings):
+    def get_keyword_fields(self, mappings: dict) -> list:
         fields = []
 
         def recurse(prefix, mappings, result):
```
webbnh, line 84: You should presumably have type hints here, too.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -89,31 +87,35 @@ def recurse(prefix, mappings, result):
...
         recurse("", mappings, fields)
         return fields
 
     def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
         """
-        Construct a pbench search query for aggregating unique values for keyword
-        fields in the 'benchmark', 'sample', 'iterations' and other
+        Construct a pbench search query for aggregating a list of values for
+        keyword fields in the 'benchmark', 'sample', 'iterations' and 'run'
         sub-documents of result-data-sample index document that belong to the
```
webbnh: Rather than listing the sub-documents explicitly, you should probably just say "the white-listed sub-documents" (or "sub-documents in WHITELIST_AGGS_FIELDS"), since the definition of the list is elsewhere, and this text is unlike to be updated if the contents of the list are changed.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -89,31 +87,35 @@ def recurse(prefix, mappings, result):
...
         json_data: JSON dictionary of type-normalized parameters
             "run_id": String representation of run id
         """
         run_id = context.get("run_id")
         dataset = context.get("dataset")
 
         self.logger.info(
-            "Return sample iteration rows for user {}, prefix {}: for run id: {}",
-            Auth.token_auth.current_user().username,
+            "Return iteration sample namespace for dataset {}|{}, prefix {}: "
```
webbnh, line 115: I would use a comma instead of a colon, and I would omit the `for` from line 116.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -122,13 +124,22 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
...
             index_map = Metadata.getvalue(dataset=dataset, key="server.index-map")
-            for key, values in index_map.items():
-                if run_id in values:
-                    index = key
-                    self.logger.debug(f"Iteration samples index, {index!r}")
+            index_keys = [key for key in index_map if "result-data-sample" in key]
         except MetadataError as e:
             abort(HTTPStatus.BAD_REQUEST, message=str(e))
 
+        if not len(index_keys) == 1:
```
webbnh, line 131: I think this would be better expressed as 
```
        if len(index_keys) != 1:
```

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -122,13 +124,22 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
...
+            self.logger.error(
+                f"Found irregular result-data-sample indices for a dataset "
+                f"{dataset.controller!r}|{dataset.name!r}"
+            )
```
webbnh: If you saw this error in the log, wouldn't you want to know what the irregular values _were_??

I think you should include the list in the message.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -125,12 +125,21 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
...
-            assert len(index_keys) == 1
-            index = index_keys[0]
-            self.logger.debug(f"Iteration samples index, {index!r}")
         except MetadataError as e:
             abort(HTTPStatus.BAD_REQUEST, message=str(e))
 
+        if not len(index_keys) == 1:
+            self.logger.error(
+                f"Found irregular result-data-sample indices for a dataset "
+                f"{dataset.controller!r}|{dataset.name!r}"
```
dbutenhof, line 134: Might be useful to have the value of `index_keys` to diagnose ...

lib/pbench/test/unit/server/query_apis/test_iteration_samples.py
```diff
@@ -1,13 +1,13 @@
...
+    IterationSampleNamespace,
 )
 from pbench.test.unit.server.headertypes import HeaderTypes
 from pbench.test.unit.server.query_apis.commons import Commons
 
 
-class TestIterationSamplesRows(Commons):
+class TestIterationSamplesNamespace(Commons):
     """
     Unit testing for IterationSamplesRows class.
```
dbutenhof, line 12: You forgot to fix the docstring!

lib/pbench/server/api/resources/query_apis/iteration_samples.py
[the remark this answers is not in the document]
webbnh: There used to be a way; then Nikhil updated the code, and, somehow, this comment didn't become "outdated".  🤷

However, the converse point has now arisen:  I don't think there is any way that `index` can be `None` at line 143.  So, I think we can delete lines 143-147, now.  (If the dataset is not found, that will be reported at line 49.)

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -0,0 +1,271 @@
...
+                            == "keyword"
+                        ):
+                            result.append(f"{property}.{sub_property}")
+
+            # Special case: sample.measurement_title is of type `text` which is not available for aggregation.
+            # However, we have added a "keyword" overlay to allow it to be aggregated such as:
+            # "measurement_title": {"type": text", "fields": {"raw": {"type": "keyword"}}}
+            result.append("sample.measurement_title.raw")
+
+        except TemplateNotFound:
```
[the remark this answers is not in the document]
webbnh: I think only line 150 needs to be in the body of the `try` block.

dbutenhof (changes requested):

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -171,10 +185,10 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
 
     def postprocess(self, es_json: JSON, context: CONTEXT) -> JSON:
         """
-        Returns a non-empty list of aggregated unique values for keyword fields
-        in the 'benchmark', 'sample', 'iterations' and 'run' subdocuments of
-        result-data-sample index. The output only contains fields that has at
-        least 1 unique value.
+        Returns a stringified JSON object containing keyword/value pairs where
```
webbnh: Actually, this function returns a Flask `Response`, I think (see [jsonify()](https://flask.palletsprojects.com/en/2.0.x/api/#flask.json.jsonify)), so both the comment and the type hint could use some adjustment.

lib/pbench/server/api/resources/query_apis/iteration_samples.py
```diff
@@ -171,10 +185,10 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
...
+        each key is the fully qualified dot-separated name of a keyword
+        (sub-)field and each value is a non-empty list of Elasticsearch
+        bucket keys which appear in that field.
```
webbnh, line 191: It's not obvious to me that the caller of this function knows or should care that the list contents are Elasticsearch bucket keys.  The point is that the elements in those lists are _values_ associated with the corresponding keys in documents in the database, such that those key/value pairs could be used to select or filter the document sets in a subsequent DB query.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -299,13 +299,13 @@ def test_http_exception(
...
 
-        if self.payload.get("start") and self.payload.get("end"):
+        if not self.use_index_from_metadata:
             index = self.build_index(
                 server_config,
                 self.date_range(self.payload["start"], self.payload["end"]),
             )
         else:
-            index = self.build_index_from_metadata(provide_metadata)
+            index = self.build_index_from_metadata()
```
webbnh: This code would be prettier if you reversed the clauses:
```
        if self.use_index_from_metadata:
            index = self.build_index_from_metadata()
        else:
            index = self.build_index(
                server_config,
                self.date_range(self.payload["start"], self.payload["end"]),
            )
```

webbnh: Ditto lines 338-344.

I kind of wonder whether you should refactor this code into a helper function, or even put it back inside `build_index()`, now that you have a nice way of handling the build-from-metadata case.

webbnh (approved):
Lookin' good!

Down to just small stuff, now.

webbnh (dismissed):
👍

lib/pbench/server/api/resources/query_apis/iteration_samples.py
npalaska: Yeah and that's true for most of our APIs where the postprocess type-hint returns are JSON which are actually JSONified flask response object

PR #2486, opened 2021-10-07
Changes to spec file and reqs (from jam session 2021-10-07)

ndokos:
In the spec file:
- make a couple of package Require:s in the spec file conditional, since the packages don't exist on RHEL8.4
- make sure that pip3 is at the latest version

In requirements.txt:
- require >=0.3.9 of flask-restful
- require >=7.13 .1, <7.14 of elasticsearch

### DIFF
--- a/server/requirements.txt
+++ b/server/requirements.txt
@@ -3,12 +3,12 @@ aniso8601>=9.0.1
 boto3
 click>=8.0
 python-dateutil
-elasticsearch>=7.13.1, !=7.14.0
+elasticsearch>=7.13.1, <7.14
 email-validator>1.1.1
 flask
 flask-bcrypt
 Flask-HTTPAuth>=4.0.0
-flask-restful
+flask-restful>=0.3.9
 flask_cors
 flask-jwt-extended
 flask-sqlalchemy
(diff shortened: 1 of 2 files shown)


server/requirements.txt
```diff
@@ -8,7 +8,7 @@ email-validator>1.1.1
 flask
 flask-bcrypt
 Flask-HTTPAuth>=4.0.0
-flask-restful
+flask-restful>=0.3.9
```
webbnh, line 11: We also need the change to the Elasticsearch version requirement.

webbnh (dismissed):

server/rpm/pbench-server.spec.j2
```diff
@@ -25,12 +25,19 @@ Requires: npm
 
 Requires: python3-alembic python3-aniso8601 python3-boto3 python3-click
 Requires: python3-dateutil python3-elasticsearch python3-email-validator
-Requires: python3-flask python3-flask-cors python3-flask-httpauth
+Requires: python3-flask python3-flask-cors
+%if 0%{?rhel} != 8
+Requires:  python3-flask-httpauth
```
dbutenhof, line 30: I'm wondering if a comment to the effect that `# Not available on RHEL8` might avoid some future confusion? That is, explicitly recording that it's not because we don't need it but because it ain't there to need?

dbutenhof (approved):
Sadly, this is only the tip of the jamberg, but it's a great first step.

webbnh (approved):
Good to go.

Issue #2000, opened 2020-11-19
Add password recovery mechanism for user api


npalaska:
We need to add a mechanism to retrieve the user password or let user reset the password after they register.
- Add an api endpoint to reset the password

PR #2491, opened 2021-10-11
Password Recovery mechanism

aquibbaig:
This is a draft PR which tracks #2000

I am yet to work on how to generate token for email recovery and send email after that

### DIFF
--- a/lib/pbench/server/api/__init__.py
+++ b/lib/pbench/server/api/__init__.py
@@ -31,7 +31,13 @@ from pbench.server.api.resources.query_apis.metadata_index.namespace_and_rows im
 )
 from pbench.server.api.resources.query_apis.month_indices import MonthIndices
 from pbench.server.api.resources.upload_api import HostInfo, Upload
-from pbench.server.api.resources.users_api import Login, Logout, RegisterUser, UserAPI, ForgotPassword
+from pbench.server.api.resources.users_api import (
+    Login,
+    Logout,
+    RegisterUser,
+    UserAPI,
+    ForgotPassword,
+)
 from pbench.server.database import init_db
 from pbench.server.database.database import Database
 
@@ -121,7 +127,9 @@ def register_endpoints(api, app, config):
         Login, f"{base_uri}/login", resource_class_args=(config, logger, token_auth),
     )
     api.add_resource(
-        ForgotPassword, f"{base_uri}/forgot_password", resource_class_args=(config, logger, token_auth),
+        ForgotPassword,
+        f"{base_uri}/forgot_password",
+        resource_class_args=(config, logger, token_auth),
     )
     api.add_resource(
         Logout, f"{base_uri}/logout", resource_class_args=(config, logger, token_auth),
--- a/lib/pbench/server/api/__init__.py
+++ b/lib/pbench/server/api/__init__.py
@@ -31,7 +31,7 @@ from pbench.server.api.resources.query_apis.metadata_index.namespace_and_rows im
 )
 from pbench.server.api.resources.query_apis.month_indices import MonthIndices
 from pbench.server.api.resources.upload_api import HostInfo, Upload
-from pbench.server.api.resources.users_api import Login, Logout, RegisterUser, UserAPI
+from pbench.server.api.resources.users_api import Login, Logout, RegisterUser, UserAPI, ForgotPassword
 from pbench.server.database import init_db
 from pbench.server.database.database import Database
 
@@ -119,6 +119,9 @@ def register_endpoints(api, app, config):
     )
     api.add_resource(
         Login, f"{base_uri}/login", resource_class_args=(config, logger, token_auth),
+    )
+    api.add_resource(
+        ForgotPassword, f"{base_uri}/forgot_password", resource_class_args=(config, logger, token_auth),
     )
     api.add_resource(
         Logout, f"{base_uri}/logout", resource_class_args=(config, logger, token_auth),
--- a/lib/pbench/server/api/resources/users_api.py
+++ b/lib/pbench/server/api/resources/users_api.py
@@ -18,12 +18,13 @@ from pbench.server.database.models.active_tokens import ActiveTokens
 from pbench.server.api.auth import Auth
 
 
-class ForgotPassword(Schema(Parameter("email", ParamType.STRING, required=True))):
+class ForgotPassword(Resource):
     def __init__(self, config, logger, auth):
         self.auth = auth
         self.config = config
         self.logger = logger
         self.mail = Mail(flask_app)
+        Schema(Parameter("email", ParamType.STRING, required=True),),
 
     def post(self):
         """
@@ -69,10 +70,12 @@ class ForgotPassword(Schema(Parameter("email", ParamType.STRING, required=True))
 
         # generate email and send to the user
         # Ideally the message should be a template
-        # rendered using render_template
+        # rendered using render_template and the hostname
+        # needs to be a variable that points opens up
+        # pbench-dashboard
         msg = Message(
             subject="Reset your password!",
-            sender=self.config.get("pbench-server", "email_username"),
+            sender=self.config.get("pbench-server", "mailfrom"),
             recipients=[email],
             body=(
                 "Please reset your password at http://localhost:8000/reset_password/"
@@ -80,10 +83,7 @@ class ForgotPassword(Schema(Parameter("email", ParamType.STRING, required=True))
             ),
         )
 
-        # This function needs to be asynchronous
         self.mail.send(msg)
-
-        return "", HTTPStatus.OK
 
 
 class RegisterUser(Resource):
--- a/lib/pbench/server/api/resources/users_api.py
+++ b/lib/pbench/server/api/resources/users_api.py
@@ -1,14 +1,76 @@
 from http import HTTPStatus
 import jwt
-from flask import request, jsonify, make_response
+from flask import request, jsonify, make_response, app
 from flask_restful import Resource, abort
 from flask_bcrypt import check_password_hash
+from flask_mail import Mail, Message
 from email_validator import EmailNotValidError
 from sqlalchemy.exc import SQLAlchemyError, IntegrityError
 from typing import NamedTuple
 from pbench.server.database.models.users import User
 from pbench.server.database.models.active_tokens import ActiveTokens
 from pbench.server.api.auth import Auth
+
+
+class ForgotPassword(Resource):
+    def __init__(self, logger):
+        self.logger = logger
+        self.mail = Mail(app)
+
+    """
+        /forgot_password
+    """
+    def post(self):
+        """
+        Post request for initiating password recovery.
+        This requires a JSON data with required user fields
+        {
+            "email": "user@domain.com"
+        }
+
+        Required headers include
+
+            Content-Type:   application/json
+            Accept:         application/json
+
+        :return:
+            Success: 200 with password reset link sent to email
+            Failure: <status_Code>,
+                    response_object = {
+                        "message": "failure message"
+                    }
+        """
+        # get the post data
+        user_data = request.get_json()
+        if not user_data:
+            self.logger.warning("Invalid json object: {}", request.url)
+            abort(HTTPStatus.BAD_REQUEST, message="Invalid json object in request")
+
+        email = user_data.get("email")
+        if not email:
+            self.logger.warning("Missing email field")
+            abort(HTTPStatus.BAD_REQUEST, message="Missing email field")
+
+        # Check if email exists in the Db, else abort
+        try:
+            user = User.query(email=email)
+        except Exception:
+            self.logger.exception("Exception while querying email")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
+
+        if not user:
+            self.logger.exception("No such user exists with the given email ID")
+            abort(HTTPStatus.BAD_REQUEST, message="No such user exists with the given email ID")
+
+        # generate email and send to the user
+        msg = Message(subject="Hello World!",
+                      sender=self.config.get("pbench-server", "email_username"),
+                      recipients=[email],
+                      body="Hello World"
+        )
+        self.mail.send(msg)
+
+        return "", HTTPStatus.OK
 
 
 class RegisterUser(Resource):
--- a/lib/pbench/server/api/resources/users_api.py
+++ b/lib/pbench/server/api/resources/users_api.py
@@ -1,27 +1,29 @@
 from http import HTTPStatus
 import jwt
-from flask import request, jsonify, make_response, current_app as app
+from flask import request, jsonify, make_response, current_app as flask_app
 from flask_restful import Resource, abort
 from flask_bcrypt import check_password_hash
 from flask_mail import Mail, Message
 from email_validator import EmailNotValidError
 from sqlalchemy.exc import SQLAlchemyError, IntegrityError
 from typing import NamedTuple
+
+from pbench.server.api.resources import (
+    Schema,
+    Parameter,
+    ParamType,
+)
 from pbench.server.database.models.users import User
 from pbench.server.database.models.active_tokens import ActiveTokens
 from pbench.server.api.auth import Auth
 
 
-class ForgotPassword(Resource):
+class ForgotPassword(Schema(Parameter("email", ParamType.STRING, required=True))):
     def __init__(self, config, logger, auth):
-        self.logger = logger
-        self.mail = Mail(app)
         self.auth = auth
         self.config = config
-
-    """
-        /forgot_password
-    """
+        self.logger = logger
+        self.mail = Mail(flask_app)
 
     def post(self):
         """
@@ -37,8 +39,8 @@ class ForgotPassword(Resource):
             Accept:         application/json
 
         :return:
-            Success: 200 with password reset link sent to email
-            Failure: Success with a logger warning
+            Success: OK status with password reset link sent to email
+            Failure: OK status, with a logger warning -- no email sent
         """
         # get the post data
         user_data = request.get_json()
@@ -59,7 +61,7 @@ class ForgotPassword(Resource):
             abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
         if not user:
-            self.logger.warning("No such user exists with the given email ID")
+            self.logger.warning("Invalid email for password recovery: {}", email)
             return "", HTTPStatus.OK
 
         # generate auth token for the valid user
--- a/lib/pbench/server/api/resources/users_api.py
+++ b/lib/pbench/server/api/resources/users_api.py
@@ -1,6 +1,6 @@
 from http import HTTPStatus
 import jwt
-from flask import request, jsonify, make_response, app
+from flask import request, jsonify, make_response, current_app as app
 from flask_restful import Resource, abort
 from flask_bcrypt import check_password_hash
 from flask_mail import Mail, Message
@@ -13,13 +13,16 @@ from pbench.server.api.auth import Auth
 
 
 class ForgotPassword(Resource):
-    def __init__(self, logger):
+    def __init__(self, config, logger, auth):
         self.logger = logger
         self.mail = Mail(app)
+        self.auth = auth
+        self.config = config
 
     """
         /forgot_password
     """
+
     def post(self):
         """
         Post request for initiating password recovery.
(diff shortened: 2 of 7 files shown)


lib/pbench/server/api/__init__.py
```diff
@@ -148,8 +153,9 @@ def create_app(server_config):
     app.config["TESTING"] = False
 
     api = Api(app)
+    mail = Mail(app)
```
dbutenhof, line 156: Since this isn't used outside your new class, I suggest removing it from this module entirely; import the package and construct the instance inside your `ForgotMail.__init__()` instead.

lib/pbench/server/api/__init__.py
```diff
@@ -111,6 +113,9 @@ def register_endpoints(api, app, config):
     api.add_resource(
         Login, f"{base_uri}/login", resource_class_args=(config, logger, token_auth),
     )
+    api.add_resource(
+        ForgotPassword, f"{base_uri}/forgot_password", resource_class_args=(config, mail, logger, token_auth),
```
dbutenhof, line 117: Pass `app` instead of `mailer` here; that'll keep the `Mail` package encapsulated where it's used. And the only change you'd have to make to this file would be your `api.add_resource()` call since `app` is already available inside `register_endpoints`. If possible, I'd prefer not to have this registration mechanism get special cased for various APIs...

lib/pbench/server/api/resources/users_api.py
```diff
@@ -11,6 +11,57 @@
 from pbench.server.api.auth import Auth
 
 
+class PasswordRecovery(Resource):
+    def __init__(self, logger, auth, mail):
+        self.logger = logger
+        self.auth = auth
+        self.mail = mail
```
dbutenhof, line 18: This could just as easily be `self.mail = Mail(app)` if you replace your `mail` instance with `app` in the constructor parameter list.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -11,6 +11,57 @@
...
+            abort(HTTPStatus.BAD_REQUEST, message="Invalid json object in request")
+
+        email = user_data.get("email")
+        if not email:
+            self.logger.warning("Missing email field")
+            abort(HTTPStatus.BAD_REQUEST, message="Missing email field")
+
+        # Check if email exists in the Db, else abort
+        try:
+            user = User.query(email=user_data.get("email"))
```
dbutenhof, line 53: You already fetched `user_data.get("email")` into local `email` so you should use that here instead of repeating the `get`.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -11,6 +11,57 @@
 from pbench.server.api.auth import Auth
 
 
+class PasswordRecovery(Resource):
```
dbutenhof, line 14: You're only using one string JSON parameter, so this isn't a big deal, but if you extended our `ApiBase` class with a schema (`Schema(Parameter("email", ParamType.STRING, required=True))`), the validation (that it exists and isn't null) would be done for you. The main advantage is that this would keep messaging of errors consistent across our APIs.

dbutenhof (commented):
Good start, Aquib! I have some comments for you to consider as you work on filling this out.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -11,6 +11,57 @@
...
+
+            Content-Type:   application/json
+            Accept:         application/json
+
+        :return:
+            Success: 201 with password reset link sent to email
+            Failure: <status_Code>,
+                    response_object = {
+                        "message": "failure message"
+                    }
```
webbnh, line 38: It looks like this function actually returns `OK`/`200` on success, not `201` (`CREATED`).

Also, I'm not sure that documenting the failure like that is particularly helpful.  (That's _really_ generic, and in fact the code actually calls the Flask `abort()` function on failure, so if someone wants to know the details they should presumably consult the documentation for _that_.)

lib/pbench/server/api/resources/users_api.py
```diff
@@ -11,6 +11,57 @@
...
+        # Check if email exists in the Db, else abort
+        try:
+            user = User.query(email=user_data.get("email"))
+        except Exception:
+            self.logger.exception("Exception while querying email")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
+
+        if not user:
+            self.logger.exception("No such user exists with the given email ID")
+            abort(HTTPStatus.BAD_REQUEST, message="No such user exists with the given email ID")
```
webbnh, line 60: I like reporting when the email address is unrecognized -- I think it produces a superior user experience -- but I also think it constitutes an information leak, so we probably shouldn't do that.  That is, if we can validate the email address, we'll send the message; otherwise we won't; and, either way, we should return success.  🙁

aquibbaig: Good point! But in dashboard's perspective if the user has a typo in his email address and he gets a success, now he is waiting for the email and no such email arrives. I think a generic message such as "Invalid email" can be useful here

lib/pbench/server/api/__init__.py
```diff
@@ -144,6 +148,17 @@ def create_app(server_config):
 
     app.logger = get_pbench_logger(__name__, server_config)
 
+    # sets up email configuration
+    mail_settings = {
```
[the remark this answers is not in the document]
dbutenhof: I would like to see the send-mail configuration completely confined either to your new `ForgotPassword` API class or to a new separate class rather than accumulating all sort of unrelated stuff to the Flask app configuration class here.

lib/pbench/server/api/__init__.py
```diff
@@ -144,6 +148,17 @@ def create_app(server_config):
...
+        "MAIL_SERVER": "smtp.gmail.com",
+        "MAIL_PORT": 465,
+        "MAIL_USE_TLS": False,
+        "MAIL_USE_SSL": True,
+        "MAIL_USERNAME": server_config.get("pbench-server", "email_username"),
+        "MAIL_PASSWORD": server_config.get("pbench-server", "email_password")
```
webbnh, line 158: It would be good to name the configuration option more specifically than `"email_username"`.  I can imagine that there might be several email addresses that we would want to configure.  This one is the one used to send mail _from_.  (Another might be one which the server uses to send periodic status reports _to_; another might be one for sending "alerts" to; etc....once you endow the server with the ability to send email, we may find that there are lots of things it wants to send!)

lib/pbench/server/api/resources/users_api.py
```diff
@@ -1,6 +1,6 @@
 from http import HTTPStatus
 import jwt
-from flask import request, jsonify, make_response, app
+from flask import request, jsonify, make_response, current_app as app
```
webbnh, line 3: @dbutenhof, @npalaska do either of you feel like `app` is too generic a name?  I would vote for `flask_app`, myself.

(That is, by the time I get to line 18, it's not obvious to me that `app` refers to Flask's object...now, that could just be my limited experience with Flask (I generally do my best to ignore the fact that it is up there somewhere in my call tree...), so I figure you guys would know better.)

lib/pbench/server/api/resources/users_api.py
```diff
@@ -13,13 +13,16 @@
 
 
 class ForgotPassword(Resource):
-    def __init__(self, logger):
+    def __init__(self, config, logger, auth):
         self.logger = logger
         self.mail = Mail(app)
+        self.auth = auth
+        self.config = config
```
webbnh: It's kind of trivial (or worse), but it would be nice if these statements were ordered, somehow.  Alphabetical by attribute name would be nice; or, failing that, in the same order as the parameters in the signature.  (The latter is more "functional"; the former provides a better "reference" for the reader.)

lib/pbench/server/api/resources/users_api.py
```diff
@@ -35,10 +38,7 @@ def post(self):
 
         :return:
             Success: 200 with password reset link sent to email
-            Failure: <status_Code>,
-                    response_object = {
-                        "message": "failure message"
-                    }
+            Failure: Success with a logger warning
```
webbnh: It would be good if these two were clearer, or at least more consistent with each other.  And, I think we generally prefer symbolic references to literals.  So, something like this would be better:
```
        :return:
            Success: OK status with password reset link sent to email
            Failure: OK status, with a logger warning -- no email sent
```

lib/pbench/server/api/resources/users_api.py
```diff
@@ -59,15 +59,22 @@ def post(self):
             abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
 
         if not user:
-            self.logger.exception("No such user exists with the given email ID")
-            abort(HTTPStatus.BAD_REQUEST, message="No such user exists with the given email ID")
+            self.logger.warning("No such user exists with the given email ID")
```
webbnh, line 62: Will we have enough context when we look at the log to know that this message comes from a "forgot password" request?  If not, you should include that in the text.  (Ditto, perhaps, for the other `logger` calls.)

lib/pbench/server/api/resources/users_api.py
```diff
@@ -59,15 +59,22 @@ def post(self):
...
         # generate email and send to the user
-        msg = Message(subject="Hello World!",
+        # Ideally the message should be a template
+        # rendered using render_template
+        msg = Message(subject="Reset your password!",
                       sender=self.config.get("pbench-server", "email_username"),
                       recipients=[email],
-                      body="Hello World"
-        )
+                      body=("Please reset your password at http://localhost:8000/reset_password/" + token)
```
webbnh, line 74: `http://localhost:8000` is just a "stub", right?  I presume that the URL is available via `self.config.get()`.

lib/pbench/server/api/__init__.py
```diff
@@ -113,7 +113,7 @@ def register_endpoints(api, app, config):
         Login, f"{base_uri}/login", resource_class_args=(config, logger, token_auth),
     )
     api.add_resource(
-        ForgotPassword, f"{base_uri}/forgot_password", resource_class_args=(config, logger),
+        ForgotPassword, f"{base_uri}/forgot_password", resource_class_args=(config, logger, token_auth),
```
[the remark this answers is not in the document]
webbnh: We will actually have two API endpoints, right?  One for requesting a reset-token (e.g., `/forgot_password`) and another for using the token to request a reset (e.g., `/reset_password`), right?  (I.e., this code defines `/forgot_password` while the code at line 74 of `lib/pbench/server/api/resources/users_api.py` is referencing `/reset_password`, but I don't see the definition of it here.)

`/forgot_password` should not require an authorization token; `/reset_password` _should_ require a token, but perhaps it will be passed as a path parameter instead of in the header?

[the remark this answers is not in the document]
webbnh: Could/should these two endpoints instead be implemented as different methods on the _same_ (or related) endpoint(s)?  (E.g., `GET` for "forgot" (with the email as a query parameter) and `PUT` for reset, or `POST` to `/forgot_password` for requesting the token and `PUT` to `/forgot_password/<token>` for reset?)

lib/pbench/server/api/resources/users_api.py
aquibbaig: That's right

lib/pbench/test/unit/server/test_user_auth.py
```diff
@@ -388,6 +393,44 @@ def test_invalid_logout(client, server_config):
...
+                username="username",
+                firstname="firstname",
+                lastname="lastName",
+                email="user@domain.com",
+                password="12345",
+            )
+            assert resp_register.status_code == HTTPStatus.CREATED
+
+        resp_forgot_password = forgot_password(
+            client, server_config, "user1@domain.com"
```
webbnh, line 430: You should probably mock out the `flask_mail` `send()` method for this test too -- the mock should cause the test to fail if it _is_ called.

webbnh (commented):
It looks like it's coming along, Aquib!  I've commented on a few things that you might want to fix sooner rather than later.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -1,16 +1,91 @@
...
 from pbench.server.database.models.active_tokens import ActiveTokens
 from pbench.server.api.auth import Auth
 
 
+class ForgotPassword(Resource):
+    def __init__(self, config, logger, auth):
+        self.auth = auth
+        self.config = config
+        self.logger = logger
+        self.mail = Mail(flask_app)
```
dbutenhof, line 26: Another option would have been to pass the app as a constructor parameter along with `config, logger, auth`; but I guess this works.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -18,12 +18,13 @@
...
 
 
-class ForgotPassword(Schema(Parameter("email", ParamType.STRING, required=True))):
+class ForgotPassword(Resource):
     def __init__(self, config, logger, auth):
         self.auth = auth
         self.config = config
         self.logger = logger
         self.mail = Mail(flask_app)
+        Schema(Parameter("email", ParamType.STRING, required=True),),
```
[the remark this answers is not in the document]
dbutenhof: Yes, leveraging the basic schema validation from `ApiBase` would be a good idea. And certainly declaring a `Schema` without assigning it anywhere doesn't help much; good catch.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -1,16 +1,91 @@
...
 
+class ForgotPassword(Resource):
+    def __init__(self, config, logger, auth):
+        self.auth = auth
+        self.config = config
+        self.logger = logger
+        self.mail = Mail(flask_app)
+        Schema(Parameter("email", ParamType.STRING, required=True),),
+
+    def post(self):
```
dbutenhof, line 29: Switching to `ApiBase` would mean re-casting this as `def _post(self, json_data: JSON, request: Request) -> Response:` ... the parameter validation will happen in the `ApiBase` wrappers and you get the processed JSON payload as `json_data` instead of fetching it yourself out of the request.

Look at `datasets_metadata.py` for an example of a direct `ApiBase` subclass.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -69,17 +70,20 @@ def post(self):
...
+        # needs to be a variable that points opens up
+        # pbench-dashboard
+        msg = Message(
+            subject="Reset your password!",
+            sender=self.config.get("pbench-server", "mailfrom"),
+            recipients=[email],
+            body=(
+                "Please reset your password at http://localhost:8000/reset_password/"
+                + token
+            ),
```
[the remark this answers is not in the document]
dbutenhof: Right, this will need to leverage the header processing I built into `endpoint_configure.py` to point back at the desired hostname and port (based on either direct or reverse-proxy access). And I suspect it should be `/dashboard/#reset_password` or something like that so that the browser will load the dashboard `index.html` and bootstrap the UMI minified Javascript that will route it to your reset page?

lib/pbench/server/api/resources/users_api.py
```diff
@@ -69,17 +70,20 @@ def post(self):
...
+        )
 
-        return "", HTTPStatus.OK
+        self.mail.send(msg)
```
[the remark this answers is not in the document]
dbutenhof: I think we just want to basically `return HTTPStatus.OK`, right?, or maybe even something like `return HTTPStatus.OK, "Password reset email sent"`??

lib/pbench/server/api/resources/users_api.py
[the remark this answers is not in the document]
aquibbaig: Yes...

lib/pbench/test/unit/server/test_user_auth.py
```diff
@@ -388,6 +393,44 @@ def test_invalid_logout(client, server_config):
...
+                username="username",
+                firstname="firstname",
+                lastname="lastName",
+                email="user@domain.com",
+                password="12345",
+            )
+            assert resp_register.status_code == HTTPStatus.CREATED
+
+        resp_forgot_password = forgot_password(client, server_config, "user@domain.com")
+        assert resp_forgot_password.status_code == HTTPStatus.OK
```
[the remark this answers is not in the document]
aquibbaig: The official documentation for flask_mail suggests that if `TESTING=True` in the flask app config, then emails will actually be suppressed and will not be sent. https://pythonhosted.org/Flask-Mail/#unit-tests-and-suppressing-emails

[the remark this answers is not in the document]
aquibbaig: Anyways, I have mocked the `send` method for now

lib/pbench/server/api/resources/users_api.py
```diff
@@ -18,12 +18,13 @@
 from pbench.server.api.auth import Auth
 
 
-class ForgotPassword(Schema(Parameter("email", ParamType.STRING, required=True))):
+class ForgotPassword(Resource):
```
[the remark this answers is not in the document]
aquibbaig: Did that

lib/pbench/server/api/resources/users_api.py
```diff
@@ -68,23 +75,21 @@ def post(self):
...
         msg = Message(
             subject="Reset your password!",
-            sender=self.config.get("pbench-server", "mailfrom"),
+            sender=self.mailfrom,
             recipients=[email],
-            body=(
-                "Please reset your password at http://localhost:8000/reset_password/"
-                + token
-            ),
+            body=("Please reset your password at " + self.host + "/" + token),
```
dbutenhof, line 85: > `body=("Please reset your password at " + self.host + "/" + token),`

`self.host` isn't ideal here; if the dashboard has been reaching the server through a reverse proxy, we ought to use that same path. This is OK for now, however; I'll convert this comment into an issue, which should be addressed by factoring the reverse proxy header logic out of `/api/v1/endpoints` so that it can be shared here. (Or anywhere else we want to report a URI to a client.)

lib/pbench/test/unit/server/test_user_auth.py
```diff
@@ -408,11 +410,16 @@ def test_forgot_password_valid_email(client, server_config):
             )
             assert resp_register.status_code == HTTPStatus.CREATED
 
+        def ok(self, msg: str):
+            return
+
+        monkeypatch.setattr(Mail, "send", ok)
```
dbutenhof, line 416: Just as a comment, you could have avoided duplicating this code by creating a `@pytest.fixture` doing this, and referencing the fixture in both tests. That's not a big deal, however.

lib/pbench/server/api/resources/users_api.py
```diff
@@ -7,25 +7,33 @@
 from email_validator import EmailNotValidError
 from sqlalchemy.exc import SQLAlchemyError, IntegrityError
 from typing import NamedTuple
+from logging import Logger
+from pbench.server import PbenchServerConfig
 
 from pbench.server.api.resources import (
```
webbnh, line 13: Line 11 belongs with the "pbench" group of imports.

Ideally, line 10 would be inserted into its group in sorted order...but that would require sorting the group....  (And, the "pbench group" could use sorting, as well....)

lib/pbench/test/unit/server/test_user_auth.py
```diff
@@ -6,7 +6,14 @@
...
 from pbench.server.database.database import Database
-from pbench.test.unit.server.conftest import register_user, login_user, admin_username
+from pbench.test.unit.server.conftest import (
+    admin_username,
+    forgot_password,
+    login_user,
+    register_user,
+)
+
+from flask_mail import Mail
```
webbnh, line 16: This should be in the group at line 4.

lib/pbench/test/unit/server/test_user_auth.py
webbnh: The mock function should be setting a flag defined in the test so that the test can check whether the mock was called:
```
    def test_forgot_password_valid_email(client, monkeypatch, server_config):
        mail_sent = False
        with client:
            ...

        def send_mock(self, msg: str):
            mail_sent = True

        monkeypatch.setattr(Mail, "send", send_mock)

        resp_forgot_password = forgot_password(client, server_config, "user@domain.com")
        assert resp_forgot_password.status_code == HTTPStatus.OK
        assert mail_sent
```
(This is a little awkward to do if the mock is set up in a fixture.)

You should use the same approach in `test_forgot_password_invalid_email()`, except that the additional assertion checks that the mail was _not_ sent.

webbnh (changes requested):
A couple of nits and a suggestion for better use of the `Mail.send()` mocks.

pravins:
Moving to Keycloak !!

PR #2492, opened 2021-10-12
Convert to Elasticsearch streaming_bulk helper

### LINKED ISSUE #2490: A large dataset cannot be marked as "published"
The recent [Greenfield Installation Exercise](https://docs.google.com/document/d/1DQP0BL3CChQmkxUxuD5-iuYRakcmCjeikeW5iJQHRfw) revealed a problem with publishing a large dataset.

Publishing a large dataset involves updating each of the Elasticsearch "documents" which comprise the dataset.  This is currently done in a single, bulk request.  Elasticsearch has a configurable maximum size for such requests which, by default, is "only" 100 Mb.  For a large dataset, this is likely to be too small to allow all the documents to be modified in a single request.  Attempting to do so currently produces a somewhat cryptic error from ES which is translated into a rather opaque message from the API.

dbutenhof:
This creates a new `ElasticBulkBase` class providing support for Elasticsearch bulk operations including publish and delete, and which currently shares the query_apis `__init__.py` with the `ElasticBase` class.

Resolves #2490


lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -1,14 +1,16 @@
 from datetime import datetime
 from http import HTTPStatus
-from logging import Logger
-from typing import Any, Callable, Dict
+from logging import Logger, error
```
dbutenhof, line 3: Weird; vscode added the `error`. I don't know why. 🤷🏻

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -1,14 +1,16 @@
...
+from typing import Any, Callable, Dict, Iterator
```
dbutenhof, line 4: Note that type hinting a generator with `Generator[...]` is ... weird. PEP says that `Iterator[root-type]` is a valid equivalent, and much easier to read.

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -350,3 +353,201 @@ def _get(self, json_data: JSON, _) -> Response:
...
+            json_data: JSON dictionary of type-normalized key-value pairs
+                controller: the controller that generated the dataset
+                name: name of the dataset to publish
+                access: The desired access level of the dataset
+
+            error_count: The number of errors encountered during the bulk
+                operation. We want to support idempotency: the Dataset should
+                not be altered unless the error_count is 0.
+        """
+        raise NotImplementedError()
```
dbutenhof, line 432: I suppose in this case we could make the abstract superclass behavior a no-op so that a subclass wouldn't need to override it if there's no completion action. (On the other hand, publish does need a completer, and delete's will be vastly more complex, so I'm not sure there's a real use case for that idea.)

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -350,3 +353,201 @@ def _get(self, json_data: JSON, _) -> Response:
...
+                message="Network problem, could not reach Elasticsearch",
+            )
+        except elasticsearch.UnsupportedProductError as e:
+            self.logger.exception(
+                "{}: Python elasticsearch package is not compatible with Elasticsearch server: {}", klasname, str(e)
+            )
+            abort(
+                HTTPStatus.INTERNAL_SERVER_ERROR,
+                message="Elasticsearch server version is incompatible",
+            )
```
dbutenhof, line 496: I don't know how much of this might actually be necessary given the bulk helper and `raise_on_error=False`; I need to investigate that. For now I left it all in (copied with minor tweaks from `ElasticBase`).

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -350,3 +352,185 @@ def _get(self, json_data: JSON, _) -> Response:
...
+        except UnauthorizedAccess as e:
+            abort(HTTPStatus.FORBIDDEN, message=str(e))
+
+        # Build an Elasticsearch instance to manage the bulk update
+        elastic = elasticsearch.Elasticsearch([self.es_url])
+
+        try:
+            # Pass the bulk command generator to the helper
+            results = elasticsearch.helpers.streaming_bulk(
+                elastic,
```
npalaska, line 478: Didn't we use something similar in our py-es-bulk package that implements `streaming_bulk`? Can't we just import and use that package?

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -350,3 +352,185 @@ def _get(self, json_data: JSON, _) -> Response:
...
+
+        self.logger.info(
+            "Update access for dataset {}: {} successful document updates and {} failures",
+            dataset,
+            count - error_count,
+            error_count,
+        )
+
+        # Let the subclass complete the operation
+        self.complete(dataset, json_data, error_count)
```
npalaska, line 521: I feel there are lot of different nomenclature that we use in our query APIs like preprocess, assemble, postprocess, now complete and generate documents. In theory I feel their end-goal is similar like here `complete` is somewhat similar to `postprocess` isn't it? Do you think we need to streamline these naming conventions

lib/pbench/server/api/resources/query_apis/__init__.py
dbutenhof: Check out issue [#19](https://github.com/distributed-system-analysis/py-es-bulk/issues/19) and [#20](https://github.com/distributed-system-analysis/py-es-bulk/issues/20) in the py-es-bulk project.

The problem is that pyesbulk is very narrowly designed for *indexing*, and was hardcoded to expect "create" operation codes... it'll fail if the operation wasn't "create". That won't be too hard to fix (aside from the testing), but the way it handles reporting errors also bothers me because it forces a human-readable output format (pretty-printed JSON) that's harder to consume electronically... I really want to fix that. I don't want to get any of that entangled with an immediate fix for "publish", so I'll handle that separately some other time.

lib/pbench/server/api/resources/query_apis/__init__.py
dbutenhof: No, `complete` isn't like `postprocess` because this is focused on setting a final `Dataset` SQL state (in this case changing the `access` column) whereas "postprocess" was specifically targeted to handling the results of an Elasticsearch query ... which, in this case, is handled in common superclass code.

Really, I chose each set of names to correspond to the "missing" operations in the superclass that needed to be implemented by the subclass. For general Elasticsearch queries we needed to "assemble" the Elasticsearch query payload and "postprocess" the Elasticsearch response. (And then I complicated it by adding "preprocess", which didn't really fit but I needed a name for setup that ran before "asssemble"...)

I did initially retrofit the new publish on top of `ElasticBase` and just made `assemble` a generator and consumed the `streaming_bulk` response generator inside `postprocess`: but that results in a somewhat different signature and substantially different semantic, which is potentially confusing. Furthermore, I don't have any use for most of `ElasticBase`. It was easier to split the class tree, and reusing the same names (with different signature and somewhat subtly different semantics) also risks confusion and seemed to have no real upside.

npalaska (approved):
Looks good to me

webbnh (changes requested):
You should update the PR description (e.g., remove the references to it being a draft), and the branch needs a rebase...but, this is great stuff!

Of course, I have the usual pile of comments.  The principal theme running through them is that this code is evidently evolutionary:  there are vestiges of the original code in the new code, and there are vestiges of the subclass in the superclass; I've pointed out the ones that I found.  Most of my concerns are centered around documentation and comments, particularly around the `json_data` parameter, which turns out to be a bit of a swiss-army knife.  (It's really more like a "context" parameter.)

So, really, it just needs a little polish and then it will be good to go!

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -350,3 +352,185 @@ def _get(self, json_data: JSON, _) -> Response:
...
+
+        try:
+            # Pass the bulk command generator to the helper
+            results = elasticsearch.helpers.streaming_bulk(
+                elastic,
+                self.generate_documents(json_data, dataset),
+                raise_on_exception=False,
+                raise_on_error=False,
+            )
+        except Exception as e:
```
portante, line 483: Keep in mind that line 477 returns a generator object.  If an exception is raised while creating the generator object, this try catch will be useful.  But that is why we have `py-es-bulk`.  =)  The exception handling needs to wrap the `for` loop at line 494 below.

portante (changes requested):
While it looks good, the details of the bulk handling is being repeated from `py-es-bulk`.  By the time we are done here, there won't be much difference.

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -350,3 +352,185 @@ def _get(self, json_data: JSON, _) -> Response:
...
+                e = u["error"]
+                status = e["type"]
+                self.logger.debug(
+                    "{} ({}: {}) for id {} in index {}",
+                    u["status"],
+                    status,
+                    e["reason"],
+                    u["_id"],
+                    u["_index"],
+                )
```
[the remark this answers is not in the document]
dbutenhof: `report` is the client response; I don't want Elasticsearch details there. I carried through the ES "status" keyword to provide some loose categorization of errors, although I wasn't entirely sure about that. I definitely don't want to return document IDs and index names; but in case of failure, I *did* want them visible somewhere.

Although since I decided to avoid logging them at `info` level this is to some extent moot as we won't see it in a real server deployment.

lib/pbench/test/unit/server/conftest.py
```diff
@@ -293,6 +303,33 @@ def provide_metadata(attach_dataset):
     Metadata.setvalue(dataset=test, key=Metadata.DELETION, value="2023-01-25")
 
 
+@pytest.fixture()
+def get_document_map(monkeypatch, attach_dataset):
```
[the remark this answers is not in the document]
dbutenhof: It's an interesting question; I have no idea what you'd call it, partly because the type changes: in the method signature, it's definitely a `Callable` with no parameters ... but within the body the same name becomes just a symbol bound to nothing or to whatever the referenced fixture returns. This is Python at its twistiest.

lib/pbench/server/api/resources/query_apis/__init__.py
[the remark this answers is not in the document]
webbnh: > `report` is the client response; I don't want Elasticsearch details there.

Good point.

> we won't see it in a real server deployment.

Perhaps not...but have some pity on the guy debugging this stuff!  😉

lib/pbench/test/unit/server/conftest.py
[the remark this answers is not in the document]
webbnh: Yeah, I have no idea what type a fixture is (I suppose I could code up a quick example or two and ask Python to tell me...), but I was thinking more along the lines of the fact that type hints aren't actually validated in anyway currently, so we could put basically anything we wanted there!  😈

lib/pbench/test/unit/server/query_apis/test_datasets_publish.py
```diff
@@ -40,210 +16,180 @@ class TestDatasetsPublish(Commons):
...
-        server_config,
-        query_api,
-        exceptions,
-        user_ok,
-        find_template,
-        get_document_map,
-        pbench_token,
+        # Verify the report and status
+        assert response.status_code == HTTPStatus.INTERNAL_SERVER_ERROR
+        assert response.json["data"] == {"ok": 28, "KIDDING": 3}
```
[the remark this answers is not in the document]
webbnh: I think that you were right to try to maintain some abstraction from ES.  And, in this particular case, it's probable that the recipient of the failure won't be able to do anything about it; so the important thing will be to have good diagnostics in the log.  So, perhaps, yeah, going with just "OK" and "failed" for the response would be appropriate.  😕

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -18,18 +18,33 @@
...
+    The subclass schema is missing the required "controller" or dataset "name"
+    parameters required to locate a Dataset.
+    """
+
+    def __init__(self, subclass_name: str):
+        super().__init__()
+        self.subclass_name = subclass_name
+
+    def __str__(self) -> str:
+        return f"ElasticBulkBase subclass {self.subclass_name} is missing required schema parameters"
```
webbnh, line 45: Does it matter that `self.subclass_name` is an `ElasticBulkBase` subclass?  I would think that the class name would be enough to get our diagnostician started; however, if that's me, I would be awfully curious about _which_ "required schema parameters" are missing....

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -377,59 +392,57 @@ def __init__(
...
         Args:
             config: server configuration
             logger: logger object
             schema: API schema: for example,
                     Schema(
-                        Parameter("user", ParamType.USER, required=True),
-                        Parameter("start", ParamType.DATE)
+                        Parameter("controller", ParamType.STRING, required=True),
+                        Parameter("name", ParamType.STRING, required=True)
                     )
```
webbnh: I see that it says "for example", but having just `controller` and `name` doesn't seem viable -- it'll need at least one additional filed (e.g., `access`) -- so I suggest adding an another field or an ellipsis after line 405.

lib/pbench/test/unit/server/query_apis/test_datasets_publish.py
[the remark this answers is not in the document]
dbutenhof: I'm OK with just summarizing "ok" and "failure"... the problem is that I think this makes logging the actual errors more critical since the client is totally insulated; and while I agree that logging each error response could be hard to read, I don't see another practical way to make failure diagnosable. I'm thinking the errors should be `info` or even `warning` instead of `debug` so we'll actually see them. It's not like we expect a lot in real operation, and it indicates something wrong with Elasticsearch that we need to diagnose.

lib/pbench/server/api/resources/query_apis/__init__.py
dbutenhof: I really didn't want to over-engineer this. I don't expect it to happen; and if it does, it's a major developer problem that'll be thrown the first time they try to stand up a unit test using their new subclass. How much effort is it really worth to diagnose exactly what (I) did wrong? 😀

lib/pbench/test/unit/server/query_apis/test_datasets_publish.py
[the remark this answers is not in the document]
webbnh: > It's not like we expect a lot in real operation, and it indicates something wrong with Elasticsearch that we need to diagnose.

It's true -- we don't _expect_ them -- but, when they show up, it's generally spectacular.  And, that's why I'm strongly recommending that we try to digest them intelligently, rather than logging them individually.

> I'm thinking the errors should be info or even warning instead of debug so we'll actually see them.

I concur.  (I would use whichever of `warn` or `error` is not associated with user actions -- I don't think there is (or should be) any way that a user could provoke these failures, so, when they show up, it should be a "big deal".)

> I think this makes logging the actual errors more critical

I concur.

> I don't see another practical way to make failure diagnosable.

In my previous life, we dealt with this by taking the message that we wanted to log (e.g., the detailed message provided by ES), and using it as a key into a `dict` storing a count as the value (I didn't know about `defaultdict(int)`, then).  The messages tended to be specific but not unique, so when we had catastrophe, it boiled down to relatively few messages (with really large incidence counts...).

lib/pbench/server/api/resources/query_apis/__init__.py
dbutenhof: In fact right now I have one other use in mind for this base class, which is DELETE; and it won't need any additional parameters. An ellipsis perhaps clarifies that you *can* but also could be read to imply that you *must*, which isn't true. I'll throw it in anyway if it makes you happy, since I won't be confused by it when I write delete. 😆

webbnh (dismissed):
It's a thing of beauty!  I just have one question....

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -507,48 +542,47 @@ def _post(self, json_data: JSON, _) -> Response:
...
                 message=f"{error_count:d} of {count:d} Elasticsearch document UPDATE operations failed",
-                data=report,
+                data=summary,
             )
 
-        return report
+        self.logger.info(
+            "{}:dataset {}: {} successful document updates", klasname, dataset, count
+        )
+        return summary
```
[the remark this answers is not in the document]
dbutenhof: It's pretty flexible in what it accepts. You can return a full Response object, or various formats the define one implicitly. Our existing code is a mixed bag of formats; a lot are using an explicit `jsonify(...)`, which is fine, but returning a dict will be implicitly jsonified, and the default status is 200. You can return a tuple with (status, data), too.

I didn't change the format of the return here, so I'm not going to worry about this; but if we want to think about forcing some sort of consistency in the future, arguably we should go for a full `Response` object on each API since that's the most flexible. (And Flask transformers will end up building one anyway.)

portante (dismissed):

lib/pbench/server/api/resources/query_apis/datasets_publish.py
```diff
@@ -34,218 +32,46 @@ def __init__(self, config: PbenchServerConfig, logger: Logger):
...
-
-        self.logger.info(
-            "Update access for dataset {} for user {} to {}, prefix {}",
-            name,
-            user,
-            access,
-            self.prefix,
-        )
-
         map = Metadata.getvalue(dataset=dataset, key=Metadata.INDEX_MAP)
```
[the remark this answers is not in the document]
portante: Tracked in issue #2505.

dbutenhof:
@webbnh still has an unresolved conversation.

PR #2493, opened 2021-10-13
pbench-list-tools: new output format and error handling

### LINKED ISSUE #2479: Proposed output for pbench-list-tools
The current output of `pbench-list-tools` does not include all the information the user might need. In particular, when using `--name=<tool>`, the output does not distinguish between different instances of the tool registered to different remotes (with possibly different options).

Without any options (or with just the `--with-option` option), the proposed output is a sequence of lines of the following form:

```
group: <group_name>; host: <host_name>; tools: <tool_name_1>, <tool_name_2>, ...
```
or (with `--with-option`):
```
group: <group_name>; host: <host_name>; tools: <tool_name_1> <option_1> ... <option_N>, <tool_name_2> <option_1>...<option_N>, ...
```
There are semicolons between the categories and each category is of the form `keyword: value`. The value for `tools` is a comma-separated list of tools (with or without a space-separated list of options, depending on `--with-option`).

The lines are sorted by group, then by host, and the tool list within a line is sorted by tool name. Currently, the options are not sorted (the order depends on how the tool was registered originally).

Using `--group=<group_name>` restricts the output above to just the lines corresponding to that group name, with no other changes.

Using `--tool=<tool_name>` changes the output to a sequence of lines of this form"

```
tool: <tool_name>; group: <group_name_1>; host: <host_name_1> [; options: <option_1>, ... <option_N>]
```
one line per group/host combination. The lines are again sorted by group name and then by host name. If present, the option list within a line is not sorted.

Comments? Assuming we come to an agreement, I'd like to implement this output format in the upcoming `pbench-list-tools` PR.

ndokos:
FIxes #2479.

Add a bunch of functional tests for `pbench-list-tools` and make the output of the tool more predictable and regular. The output format is described and discussed in issue #2479. Some additional considerations that came up during review of this PR are captured in issue #2501.

### DIFF
--- a/lib/pbench/cli/agent/commands/tools/list.py
+++ b/lib/pbench/cli/agent/commands/tools/list.py
@@ -24,15 +24,16 @@ class ListTools(ToolCommand):
 
     @staticmethod
     def print_results(toolinfo: dict, tool: str, with_option: bool):
-        for group in sorted(toolinfo.keys()):
-            for host in sorted(toolinfo[group].keys()):
-                tools = toolinfo[group][host]
+        for group, gval in sorted(toolinfo.items()):
+            for host, tools in sorted(gval.items()):
                 if tools:
                     if not with_option:
                         s = ", ".join(sorted(tools.keys()))
                     else:
-                        tools_with_options = [f"{t} {' '.join(tools[t])}" for t in sorted(tools.keys())]
-                        s = ", ".join( tools_with_options)
+                        tools_with_options = [
+                            f"{t} {' '.join(v)}" for t, v in sorted(tools.items())
+                        ]
+                        s = ", ".join(tools_with_options)
                     print(f"group: {group}; host: {host}; tools: {s}")
 
     def execute(self):
@@ -53,14 +54,21 @@ class ListTools(ToolCommand):
             for group in groups:
                 tool_info[group] = {}
                 try:
-                    for path in self.gen_tools_group_dir(group).glob("*/**"):
-                        host = path.name
-                        tool_info[group][host] = {}
-                        for tool in sorted(self.tools(path)):
-                            tool_info[group][host][tool] = sorted((path / tool).read_text().rstrip('\n').split('\n')) if opts else ""
+                    tg_dir = self.gen_tools_group_dir(group).glob("*/**")
                 except BadToolGroup:
                     self.logger.error("Bad tool group: %s", group)
                     return 1
+
+                for path in tg_dir:
+                    host = path.name
+                    tool_info[group][host] = {}
+                    for tool in sorted(self.tools(path)):
+                        tool_info[group][host][tool] = (
+                            sorted((path / tool).read_text().rstrip("\n").split("\n"))
+                            if opts
+                            else ""
+                        )
+
             if tool_info:
                 self.print_results(tool_info, None, self.context.with_option)
         else:
@@ -84,10 +92,15 @@ class ListTools(ToolCommand):
                     tool_info[group][host] = {}
                     # Check to see if the tool is in any of the hosts.
                     if tool in self.tools(path):
-                        tool_info[group][host][tool] = sorted((path / tool).read_text().rstrip("\n").split('\n')) if opts else ""
+                        tool_info[group][host][tool] = (
+                            sorted((path / tool).read_text().rstrip("\n").split("\n"))
+                            if opts
+                            else ""
+                        )
                         found = True
             if found:
                 self.print_results(tool_info, tool, self.context.with_option)
+                return 0
             else:
                 self.logger.error(
                     "Tool does not exist in any group: %s", self.context.name
--- a/lib/pbench/test/functional/agent/cli/commands/tools/test_tools_list.py
+++ b/lib/pbench/test/functional/agent/cli/commands/tools/test_tools_list.py
@@ -1,12 +1,12 @@
 import pytest
 
 
-EMPTY         = b""
-USAGE         = b"Usage: pbench-list-tools [OPTIONS]"
-TRACEBACK     = b"Traceback (most recent call last):\n"
+EMPTY = b""
+USAGE = b"Usage: pbench-list-tools [OPTIONS]"
+TRACEBACK = b"Traceback (most recent call last):\n"
 
 BAD_GROUP_ERR = b"Bad tool group: "
-TOOL_ERR      = b"Tool does not exist in any group: "
+TOOL_ERR = b"Tool does not exist in any group: "
 
 
 class Test_list_tools_no_tools_registered:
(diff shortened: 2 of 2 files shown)


dbutenhof (dismissed):
While I'd love to see a neater and more easily readable indented hierarchical format ... as a cleanup and "culturally compatible" improvement to the old code, this looks good.

lib/pbench/cli/agent/commands/tools/list.py
```diff
@@ -22,6 +22,19 @@ class ListTools(ToolCommand):
...
+        for group in sorted(toolinfo.keys()):
+            for host in sorted(toolinfo[group].keys()):
+                tools = toolinfo[group][host]
+                if tools:
+                    if not with_option:
+                        s = ", ".join(sorted(tools.keys()))
+                    else:
+                        tools_with_options = [f"{t} {' '.join(tools[t])}" for t in sorted(tools.keys())]
+                        s = ", ".join( tools_with_options)
+                    print(f"group: {group}; host: {host}; tools: {s}")
```
webbnh, line 36: If I might suggest...
```
        for group, gval in sorted(toolinfo.items()):
            for host, tools in sorted(gval.items()):
                if tools:
                    if with_option:
                        tools_with_options = [f"{t} {' '.join(v)}" for t, v in sorted(tools.items())]
                        s = ", ".join(tools_with_options)
                    else:
                        s = ", ".join(sorted(tools.keys()))
                    print(f"group: {group}; host: {host}; tools: {s}")
```

lib/pbench/cli/agent/commands/tools/list.py
```diff
@@ -33,47 +46,53 @@ def execute(self):
...
-                    self.logger.error("bad or missing tool group %s", group)
-                    continue
+                try:
+                    tg_dir = self.gen_tools_group_dir(group)
+                except BadToolGroup:
+                    self.logger.error("Bad tool group: %s", group)
+                    return 1
 
+                tool_info[group] = {}
                 for path in tg_dir.iterdir():
```
webbnh, line 78: I kind of like this idiom (with the small body of the `try` block); please consider using it for the previous call to `get_tools_group_dir()`.

lib/pbench/cli/agent/commands/tools/list.py
```diff
@@ -33,47 +46,53 @@ def execute(self):
...
+                    if tool in self.tools(path):
+                        tool_info[group][host][tool] = sorted((path / tool).read_text().rstrip("\n").split('\n')) if opts else ""
+                        found = True
+            if found:
+                self.print_results(tool_info, tool, self.context.with_option)
+            else:
+                self.logger.error(
+                    "Tool does not exist in any group: %s", self.context.name
                 )
+                return 1
```
webbnh, line 95: It looks like you're missing the return statement for the case where control executes line 90.  (It probably defaults to `None` which is probably a lot like `0`, and so we don't notice, but we should be returning `0` explicitly for the success case.)

lib/pbench/test/functional/agent/cli/commands/tools/test_tools_list.py
```diff
@@ -41,6 +92,17 @@ def tool_with_random_file(self, pbench_run):
...
+    @pytest.fixture
+    def tools_on_multiple_hosts(self, pbench_run):
+        p = pbench_run / "tools-v1-default" / "testhost.example.com"
+        p.mkdir(parents=True)
+        for tool in ["perf", "mpstat"]:
+            (p / tool).touch()
+        p = pbench_run / "tools-v1-default" / "testhost2.example.com"
+        p.mkdir(parents=True)
+        for tool in ["iostat", "sar"]:
+            (p / tool).touch()
```
webbnh, line 104: Would it make for a more interesting test if there were a partial overlap between the hosts' tool-sets?

lib/pbench/test/functional/agent/cli/commands/tools/test_tools_list.py
```diff
@@ -54,15 +116,15 @@ def test_no_args(self, tool, agent_config):
         out, err, exitcode = pytest.helpers.capture(command)
         assert TRACEBACK not in err
         assert EMPTY == err
-        assert b"default: testhost.example.com ['perf']" in out
+        assert b"group: default; host: testhost.example.com; tools: perf\n" in out
```
webbnh, line 119: Shouldn't `mpstat` be here, too?

lib/pbench/test/functional/agent/cli/commands/tools/test_tools_list.py
```diff
@@ -54,15 +116,15 @@ def test_no_args(self, tool, agent_config):
...
         assert exitcode == 0
 
     def test_name(self, tool, agent_config):
         command = ["pbench-list-tools", "-n", "perf"]
         out, err, exitcode = pytest.helpers.capture(command)
         assert TRACEBACK not in err
         assert EMPTY == err
-        assert b"tool name: perf groups: default" in out
+        assert b"group: default; host: testhost.example.com; tools: perf" in out
```
webbnh, line 127: What about `mpstat` in this one?  (It's supposed to be _absent_, right?  Shouldn't we assert that?)

webbnh: Ditto line 136, 156, et al.

lib/pbench/test/functional/agent/cli/commands/tools/test_tools_list.py
webbnh: Ditto line 150, et al.

webbnh (dismissed):
Looks good.  There's one missing return statement; otherwise I just have nits and small suggestions.

ndokos:
I'm not sure why all checks passed, but there is a typo  on l.170 of `test_tools_list.py` that should have blown up the tests.

lib/pbench/cli/agent/commands/tools/list.py
ndokos: I did that except for the inversion of the tests: I like the longer body to go with the `else`.

lib/pbench/cli/agent/commands/tools/list.py
ndokos: Done.

lib/pbench/cli/agent/commands/tools/list.py
ndokos: Done.

lib/pbench/test/functional/agent/cli/commands/tools/test_tools_list.py
ndokos: Why should `mpstat` be here (or on l.150)? The `tool` fixture just creates a `perf` tool.

lib/pbench/test/functional/agent/cli/commands/tools/test_tools_list.py
ndokos: Yes, I changed all of them to require equality, rather than inclusion.

lib/pbench/test/functional/agent/cli/commands/tools/test_tools_list.py
webbnh: I must have been confused about which fixture was used to set up the tools list.

PR #2494, opened 2021-10-14
Add support for performance testing of the Agent postprocessing tools

webbnh:
This PR is a rework of #2365 adding a test for the postprocessing of a big Fio dataset.  Instead of checking the results of the postprocessing for correctness, with the changes in this PR the results are ignored an the PASS/FAIL criterion is whether the test completed execution in under 5 minutes.

The branch is currently comprised of four commits.  The first two are deep refactoring of `agent/tool-scripts/postprocessing/unittests`, which boils each of the test execution functions down to its essentials, moving the common code to subroutines which are invoked from the higher-level code.  The third commit introduces the performance testing framework, and the last adds the new performance test.  You will probably want to review each commit separately, and the last commit is just input files for the new test, so it's not particularly interesting.  (When it's time to merge this PR, I'll probably squash the first two commits together and leave the last two separate.)

This PR also deletes two files from the gold tree.  These probably shouldn't have been there to begin with, and, with the revamped test framework, they cause the diff's to fail because there is nothing in the result tree to compare them against.  (One is a symlink to a file which doesn't exist, IIRC; the other is a directory full of _input_ files which have no business mixed in with the gold _output_ files.)

Reviewer invitations:  @portante, @ndokos, @dbutenhof


agent/tool-scripts/postprocess/unittests
```diff
@@ -2,6 +2,7 @@
 export LANG=C.UTF-8
 export LC_ALL=C.UTF-8
 export PERL_HASH_SEED=0
+let MAX_PERFTEST_SECONDS=300  # 5 minutes
```
dbutenhof, line 5: Do we really want a single fixed performance threshold for all performance tests? Another option would be to define a new setup file (like the server's `state` directory protocol) that allows per-test configuration. E.g., for tests requiring database pre-loads, I created a `${test}.db` file that would be sourced. Something like a `${test}.perf` could not only set per-test timeouts, but could also be used to identify a performance test instead of requiring a specific name prefix.

Then again, if you think a global fixed constant is good enough for now, there's enough in this PR and I'm not going to object.

webbnh: Yeah, the problem is, where to draw the line....  I think we're all hoping that this entire framework is going to go away and be replaced by something better (e.g., something written in Python).  Moreover, if a performance test fails to complete by the deadline, it should be killed and the rest of the tests allowed to continue, but I deemed that to be beyond the scope of this change, as well.  And, ideally, the threshold would be not only per-test but also adaptive -- "I, Mr. Jenkins, think this test failed because the current performance is out of line with recent runs of this test."

So, your suggestion is an excellent one, but it's just a little more than I'm inclined to take on for this PR.

Issue #2447, opened 2021-09-07
`pbench-fio --help` says `-d`/`--targets` option accepts a directory, but a file is needed


jhutar:
Help says it accepts directory:

```
[root@... ~]# pbench-fio -h | grep -A 1 -- '-d str'
	-d str[,str] --targets=str[,str]
		one or more directories or block devices (default is /tmp/fio)
```

But it does not work that way:

```
[root@... ~]# pbench-fio -d /var/lib/kafka/ --samples=1
[...]
running fio job: /var/lib/pbench-agent/fio__2021.09.07T08.47.37/1-read-4KiB/fio.job (sample1)
fio: pid=0, err=21/file:filesetup.c:150, func=unlink, error=Is a directory
[warn][2021-09-07T08:48:06.490405156] log_hist_msec specified in job file but failed to find any log files (.../clients/localhost/fio_clat_hist.*.log*) to process for histograms
[warn][2021-09-07T08:48:06.491953332] log_hist_msec specified in job file but failed to find any log files (.../clients/*/fio_clat_hist.*.log*) to process for histograms
fio job complete
```

When I created the file and reran with that, it seems to be running:

```
[root@... ~]# touch /var/lib/kafka/fio
[root@... ~]# pbench-fio -d /var/lib/kafka/fio --samples=1
```

IMHO help string needs to be corrected.

portante:
The behavior is correct, the help message should read something like:

```
        -d str[,str] --targets=str[,str]
               one or more files, block devices, or directories (default is /tmp/fio)
               persistent names for block devices highly recommended; to use a directory,
               use the /opt/pbench-agent/bench-scripts/templates/fio-shared-fs.job file or specify a
               job file with --job-file=<path> which uses the 'directory = $target' syntax
```

PR #2517, opened 2021-10-30
Enhance `pbench-fio --targets` help message

portante:
The help message for `-d`/`--targets` directs the user to use the `directory = $targets` option in a fio job file they provide or gives them the location of the `templates/fio-shared-fs.job` file which uses it.


dbutenhof (dismissed):
My only comment might be the potential vagueness of the "persistent names for devices highly recommended," but that comes from the original text and I suppose even "beginner" `fio` users would need to have a fairly strong grasp of the fundamentals of device management...

agent/bench-scripts/pbench-fio
```diff
@@ -115,8 +115,10 @@ function usage {
 	printf "\t\tfile sizes in MiB\n"
 	printf "\n"
 	printf -- "\t-d str[,str] --targets=str[,str]\n"
-	printf "\t\tone or more directories or block devices (default is $targets)\n"
-	printf "\t\t(persistent names for devices highly recommended)\n"
+	printf "\t\tone or more files, block devices, or directories (default is $targets)\n"
+	printf "\t\tpersistent names for block devices highly recommended; to use a directory,\n"
+	printf "\t\tuse the ${script_path}/templates/fio-shared-fs.job file or specify a\n"
+	printf "\t\tjob file with --job-file=<path> which uses the 'directory = \$target' syntax\n"
```
webbnh, line 121: The pseudo BNF (i.e., `[,str]`) doesn't make it clear that more than two items can be specified.  Perhaps adding an elipsis would suffice ((i.e., `[,str...]` or `[,str]...` or `[[,str]...]]`).

The phrase `persistent names for block devices highly recommended` is sort of floating in space.  It should be preceded by a comma, em-dash, or semicolon, and it might be good if it were placed inside the parenthetical.

The clause starting with `to use a directory` is confusing:  it seems to contradict the preceding verbiage which explains how to specify a directory on the command line.  Perhaps we could add some words which make it clear that these are alternative means for specifying the directory.

agent/bench-scripts/tests/pbench-fio/test-14.txt
```diff
@@ -27,8 +27,10 @@ The following options are available:
 		file sizes in MiB
 
 	-d str[,str] --targets=str[,str]
-		one or more directories or block devices (default is /tmp/fio)
-		(persistent names for devices highly recommended)
+		one or more files, block devices, or directories (default is /tmp/fio)
+		persistent names for block devices highly recommended; to use a directory,
+		use the /var/tmp/pbench-test-bench/opt/pbench-agent/bench-scripts/templates/fio-shared-fs.job file or specify a
+		job file with --job-file=<path> which uses the 'directory = $target' syntax
```
webbnh, line 33: Since we can guess that the value of `${script_path}/templates/fio-shared-fs.job` is going to be a long path, perhaps we could set up the text so that it lands on a line by itself?

webbnh (changes requested):
Based on reading #2447, it looks like Jan was unable to reference a directory using `-d`.  However, the proposed change to the help text still suggests (to me, at least), that a directory is a valid value for `-d`.

portante:
> Based on reading #2447, it looks like Jan was unable to reference a directory using `-d`. However, the proposed change to the help text still suggests (to me, at least), that a directory is a valid value for `-d`.

It is, you just need to use a "fio" job file which takes advantage of that, which is what the new help text says.

agent/bench-scripts/pbench-fio
portante: I addressed the elipsis change as a separate commit since it really is not part of this fix.

Hopefully the revised text is clear.

dbutenhof (approved):

agent/bench-scripts/pbench-fio
```diff
@@ -108,13 +108,13 @@ function usage {
 	printf -- "\t--ramptime=int\n"
 	printf "\t\ttime in seconds to warm up test before taking measurements (default is $ramptime)\n"
 	printf "\n"
-	printf -- "\t-b int[,int] --block-sizes=str[,str] (default is $block_sizes)\n"
+	printf -- "\t-b int[,int...] --block-sizes=int[,int...] (default is $block_sizes)\n"
 	printf "\t\tone or more block sizes in KiB\n"
 	printf "\n"
-	printf -- "\t-s int[,int] --file-size=str[,str] (no default)\n"
+	printf -- "\t-s int[,int...] --file-size=int[,int...] (no default)\n"
```
webbnh, line 114: 😀

webbnh (approved):
The current reword is excellent.  Thank you.

PR #2518, opened 2021-10-30
Ensure `pbench-fio` exits w/ non-zero exit status

portante:
If `pbench-fio` fails for some reason, either the maximum number of sample failures per iteration was reached in an iteration, or if any of the `fio` commands exited with an error, or if the benchmark post-processing failed, or if the final benchmark summary generation failed, the `pbench-fio` command now exits with a non-zero exit status code.


agent/bench-scripts/test-bin/bm
```diff
@@ -5,6 +5,8 @@
 
 echo "$0 $*" >> $_testlog
 
+_exit_sts=${_BM_EXIT_STS:-0}
```
dbutenhof, line 8: This could have been collapsed into the `exit` command since `_BM_EXIT_STS` appears to be a McGuffin created exclusively for your new legacy test case, and `_exit_sts` isn't assigned anywhere else. I think that assigning it separately causes the eyes to try to follow all the intervening code to see where it might be changed... which turns out to be a waste of time. 😁

agent/bench-scripts/pbench-fio
```diff
@@ -758,7 +759,12 @@ function fio_run_job() {
 		fi
 	fi
 
-	echo "fio job complete"
+	if [[ ${_fio_exit_code} -eq 0 ]]; then
+		echo "fio job complete"
+	else
+		echo "fio job failed, status code ${_fio_exit_code}"
+	fi
+	return ${_fio_exit_code}
```
dbutenhof, line 767: Interesting that you went to the trouble above of changing an `echo` to `prinf --` and then added two new instances of `echo` down here. Is there a reason?

dbutenhof (approved):

webbnh (dismissed):
Seems like a worthy change, overall.  However, there's a little bit of code churn here which I think we can do without, and I'm curious as to why you felt it was necessary to add `exit` commands to the one-line scripts.  And, there's a nit.  And, I'm interested to hear what you have to say about the loop exit that I called out.

agent/bench-scripts/pbench-fio
```diff
@@ -902,14 +919,18 @@ function fio_run_benchmark() {
 					if [ $fail -ne  0 ]; then
 						((failures++))
 					fi
+					if [ $failures -ge $max_failures ]; then
+						((iteration_failures++))
+					fi
 					if [ $fail -eq 0 -o $failures -ge $max_failures ]; then
 						break 1
 					fi
 				done
```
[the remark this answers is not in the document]
portante: This code is really convoluted.  I have undertaken a bit of refactoring to make it clearer.

agent/bench-scripts/pbench-fio
```diff
@@ -941,9 +962,12 @@ if [[ "$postprocess_only" == "n" ]]; then
 fi
 
 fio_run_benchmark
-echo "$script_path/postprocess/generate-benchmark-summary \"$benchmark\" \"$orig_cmd\" \"$benchmark_run_dir\"" >"$benchmark_run_dir/generate-benchmark-summary.cmd"
+_frb_sts=${?}
+
+printf -- "$script_path/postprocess/generate-benchmark-summary \"$benchmark\" \"$orig_cmd\" \"$benchmark_run_dir\"\nexit \${?}" >"$benchmark_run_dir/generate-benchmark-summary.cmd"
```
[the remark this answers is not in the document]
portante: Paranoia on my part.  I'll revert.

agent/bench-scripts/pbench-fio
```diff
@@ -916,14 +916,23 @@ function fio_run_benchmark() {
...
+							# We consider this iteration to be a failure since too
+							# many sample failures were encountered.
+							((iteration_failures++))
+							repeat_iteration=0
+						fi
+						if [ "$postprocess_only" = "n" ]; then
+							# Do not repeat the iteration in post-processing only mode.
+							repeat_iteration=0
+						fi
 					fi
```
webbnh: If I might suggest:
```
					if [ $fail -eq 0 -o "$postprocess_only" = "n" ]; then
						# This sample was successful, 
						# or we're in post-processing
						# only mode, so break out of
						# the retry loop.
						break
					fi
					# The sample was not within the max stddev.
					((failures++))
					if [ $failures -ge $max_failures ]; then
						# We consider this iteration to
						# be a failure since too many
						# sample failures were encountered.
						((iteration_failures++))
						break
					fi
```
This would allow us to get rid of the `repeat_iteration` variable by changing line 821 to an "infinite" loop.  (This approach won't count any failures during postprocess-only operation, but I don't think that changes overall behavior.)

agent/bench-scripts/pbench-fio
```diff
@@ -957,3 +990,10 @@ if [[ "$postprocess_only" == "n" ]]; then
 	fi
 	rmdir $benchmark_run_dir/.running
 fi
+
+if [[ ${_frb_sts} -ne 0 || ${_gbs_sts} -ne 0 ]]; then
+	_exit_sts=1
+else
+	_exit_sts=0
+fi
+exit ${_exit_sts}
```
webbnh, line 999: Alternatively:
```
[[ ${_frb_sts} -eq 0 && ${_gbs_sts} -eq 0 ]]
exit
```
(and, you can omit the `exit`....)

webbnh (approved):
This looks very good, now.

I have a couple of coding suggestions.

agent/bench-scripts/pbench-fio
portante: Wow that is subtle and tricky.  I'll do it, but I have added a comment since it is so sly.

agent/bench-scripts/pbench-fio
portante: Ick, I really try to avoid explicit infinite loop constructs, but okay.

portante (commented):
Since I had to rebase I also squashed it to one commit.

agent/bench-scripts/pbench-fio
webbnh: LOL

`[[` is a builtin shelll _command_!  😁 It does what it says on the tin....  (But, commenting things which are "surprising" is good.)

agent/bench-scripts/pbench-fio
webbnh: I figure, if you buy into using `break` statements, then all you've really done is moved the exit test from the top of the loop to a more convenient point...so, it's not actually an "infinite loop", it's a loop with a "deferred test".  😁

webbnh: Oh, and, when one gets to the point where s/he is creating control variables just to make the loop exit test work right, it's probably time to switch to a "deferred test" loop.  😉

webbnh (dismissed):
Looks glorious.  (I'm not sure what you did to `test-55.txt`, but I'm taking it on faith that it's good now, too.)

agent/bench-scripts/pbench-fio
webbnh: OK, so it's not so much the presence of the `[[` "command" as it is the absence of the explicit `exit` command...so, yeah, a comment there is instructive.

agent/bench-scripts/pbench-fio
```diff
@@ -957,3 +985,7 @@ if [[ "$postprocess_only" == "n" ]]; then
 	fi
 	rmdir $benchmark_run_dir/.running
 fi
+
+# FYI this tricky using the exit status of `[[` itself, where we only exit with
+# zero if both statuses are 0, otherwise we exit with 1 indicating an error.
```
dbutenhof, line 990: Hmm; the sentence structure here is a bit weird. Maybe you meant "this is tricky" or "this is a trick"? (bash style trick or treat?) But probably not worth another commit to clarify since I assume any reader will likely figure it out.

agent/bench-scripts/pbench-fio
```diff
@@ -801,21 +807,21 @@ function fio_run_benchmark() {
...
 				if [[ "$job_mode" = "serial" ]]; then
 					dev_short_name="$(basename $dev)"
 					# easier to identify what job used what device when having 1 job per device
 					iteration="${iteration}-${dev_short_name}"
 				fi
 				record_iteration ${count} ${test_type} ${block_size} ${dev} ${iteration}
-				iteration_dir=$benchmark_run_dir/$iteration
-				result_stddevpct=$maxstddevpct # this test case will get a "do-over" if the stddev is not low enough
-				failures=0
-				while [[ $(echo "if (${result_stddevpct} >= ${maxstddevpct}) 1 else 0" | bc) -eq 1 ]]; do
```
dbutenhof, line 818: I don't quite see what happened to the "do-over if stddev is not low enough" logic ... but then I must have missed where `result_stddevpct` was set in the loop, because I don't see it. Somewhere I'd expect we're *not* exiting even if everything was successful if the stddev is too high? I see this same pattern repeated across the various wrappers, but nowhere under the `agent` tree do I see anything that *sets* `result_stddevpct` except to initialize it to `$maxstddevpct`. Weird.

dbutenhof (approved):

agent/bench-scripts/pbench-fio
webbnh: I think it was being initialized so that the loop would run, and then the script which should/would have updated the value is instead returning a failure status if the value is out of bounds...it was just a fancily-named control variable.  What Peter has in there now is much better, I think.

agent/bench-scripts/pbench-fio
portante: Fixed.

agent/bench-scripts/pbench-fio
portante: > I don't quite see what happened to the "do-over if stddev is not low enough" logic ... but then I must have missed where `result_stddevpct` was set in the loop, because I don't see it. Somewhere I'd expect we're _not_ exiting even if everything was successful if the stddev is too high? I see this same pattern repeated across the various wrappers, but nowhere under the `agent` tree do I see anything that _sets_ `result_stddevpct` except to initialize it to `$maxstddevpct`. Weird.

I agree, this code is weird.  Ideally, the inner loop should be a separate method that returns a value to be tested for the control loop, but fixing these nested loops is a level of refactoring best left to a re-write in Python3 or shedding in favor of just using https://github.com/cloud-bulldozer/benchmark-wrapper.

For many years I have read this code where `result_stddevpct` was being initialized but never changed, and just left this weirdness alone for fear of breaking something given no bandwidth to address it.  @webbnh's comments pushed me to make the change so we can more easily verify the logic is correct.

dbutenhof: > I think it was being initialized so that the loop would run, and then the script which should/would have updated the value is instead returning a failure status if the value is out of bounds...it was just a fancily-named control variable. What Peter has in there now is much better, I think.

While I suspect that, on the surface, all of that is true, the point of this variable was clearly to *continue running even on success* if the standard deviation was above a threshhold; and an error exit deeper in the code can't accomplish that. Maybe there's something deeper that's looping without dropping out to this code, but I didn't find it on a quick scan. It seems likely that this code isn't doing what some developer once thought they wanted it to do; what it *is* doing (in this respect), or *how* that's happening, isn't at all clear... or whether that represents a problem we ought to consider worrying about. 😀

dbutenhof: > For many years I have read this code where `result_stddevpct` was being initialized but never changed, and just left this weirdness alone for fear of breaking something given no bandwidth to address it. @webbnh's comments pushed me to make the change so we can more easily verify the logic is correct.

Ha; yeah, I know that feeling. So basically you have no more insight into this bizarre piece of historical logic than I do, we know that it's not doing anything sensible (and is in addition "not doing it" in a completely bizarre way); so any problem we have after removing it, we probably already had, and we're better off for losing the cruft. 😆

dbutenhof (approved):

webbnh (approved):
GTG

PR #2522, opened 2021-11-04
Fix specjbb mdlog

portante:
Correct gross errors in `pbench-specjbb2005`

Past modification to add iteration records never defined the `mdlog` variable. The `mdlog` variable is now properly defined for use.  See commit 93391fb9e6ac5848e79bda7111618ea8391456f5 (PR #1049).

When the Tool Meister first landed, an undefined variable, `sample`, was being passed to fourth paramater to `pbench-tool-trigger`. The fourth is now properly passed as `1` since only one sample directory is created per iteration.  See commit 23cc097e2857e8069c3a4c6ab485a27ed6c19d73 (PR #1248)

This commit also removes the unused `peak_warehouses` variable.


agent/bench-scripts/tests/pbench-specjbb2005/test-56.pre
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/bash
+
+_dir="${_testtmp}/specjbb2005"
+mkdir ${_dir} || exit 1
+touch ${_dir}/SPECjbb.props || exit 1
+printf "input.jvm_instances=_replace_\n" >> ${_dir}/SPECjbb.props
+printf "input.measurement_seconds=_replace_\n" >> ${_dir}/SPECjbb.props
+printf "input.starting_number_warehouses=_replace_\n" >> ${_dir}/SPECjbb.props
+printf "input.increment_number_warehouses=_replace_\n" >> ${_dir}/SPECjbb.props
+printf "input.ending_number_warehouses=_replace_\n" >> ${_dir}/SPECjbb.props
```
dbutenhof, line 10: Argh; it took me too much time to realize you're processing a template using a `sed` pattern where the value after the `=` doesn't matter, and this caused some confusion. I don't really have a specific alternative since clearly this `sed` pattern has been accepted for a while. Maybe even just a `# benchmark wrapper replaces everything after =` to clarify what's really happening here?

dbutenhof (approved):

agent/bench-scripts/pbench-specjbb2005
```diff
@@ -325,7 +324,7 @@ specjbb_opts="-cp ${specjbb_cp} -Xms${heap_size} -Xmx${heap_size} ${java_opts} s
 for i in `seq $nr_jvms -1 1`; do
 	if [ $i -eq 1 ]; then
 		# the last JVM gets output so we can trigger the tools starting at iteration 1
-		java ${specjbb_opts} -id $i | tee ./jvm-$i-worker.txt | pbench-tool-trigger 1 "${benchmark_run_dir}" "${tool_group}" "${sample}"
+		java ${specjbb_opts} -id $i | tee ./jvm-$i-worker.txt | pbench-tool-trigger 1 "${benchmark_run_dir}" "${tool_group}" 1
```
webbnh, line 327: Arguably, we should replace `${sample}` with `$i`.  Replacing it with `1` instead is OK, because we know that, at this point, `$i` _is_ `1`, but, on that basis, should we replace the other `$i` (or `$i`'s) with `1`, as well?

(It also occurs to me that perhaps line 327 should be pulled out of the loop and executed afterwards, with the loop ending at `2` instead of `1`:  this would remove the need for the conditional, and it would better justify using a constant value instead of `$i`.)

agent/bench-scripts/test-bin/java
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python3
```
webbnh, line 1: Shouldn't we have a nice block comment here explaining what this blob of elegance is for and what it does?

agent/bench-scripts/tests/pbench-specjbb2005/test-56.post
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/bash
+
+_dir="${_testtmp}/specjbb2005"
+rm -rf ${_dir}/*
+rmdir ${_dir}
```
webbnh, line 5: Why do this in two steps, instead of just one?
```
rm -rf ${_dir}
```

agent/bench-scripts/tests/pbench-specjbb2005/test-56.pre
webbnh, line 10: I missed it:  where does the `_replace_` token get replaced?

agent/bench-scripts/tests/pbench-specjbb2005/test-56.txt
```diff
@@ -0,0 +1,183 @@
...
+/var/tmp/pbench-test-bench/pbench-agent/specjbb2005_test-56_1900.01.01T00.00.00/7-warehouses:7/sample1
+/var/tmp/pbench-test-bench/pbench-agent/specjbb2005_test-56_1900.01.01T00.00.00/7-warehouses:7/sample1/SPECjbb.raw
+/var/tmp/pbench-test-bench/pbench-agent/specjbb2005_test-56_1900.01.01T00.00.00/7-warehouses:7/sample1/tools-default
+/var/tmp/pbench-test-bench/pbench-agent/specjbb2005_test-56_1900.01.01T00.00.00/8-warehouses:8
+/var/tmp/pbench-test-bench/pbench-agent/specjbb2005_test-56_1900.01.01T00.00.00/8-warehouses:8/reference-result
+/var/tmp/pbench-test-bench/pbench-agent/specjbb2005_test-56_1900.01.01T00.00.00/8-warehouses:8/sample1
+/var/tmp/pbench-test-bench/pbench-agent/specjbb2005_test-56_1900.01.01T00.00.00/8-warehouses:8/sample1/SPECjbb.raw
+/var/tmp/pbench-test-bench/pbench-agent/specjbb2005_test-56_1900.01.01T00.00.00/8-warehouses:8/sample1/tools-default
+/var/tmp/pbench-test-bench/pbench-agent/specjbb2005_test-56_1900.01.01T00.00.00/SPECjbb.props
+/var/tmp/pbench-test-bench/pbench-agent/specjbb2005_test-56_1900.01.01T00.00.00/SPECjbb.props.sed
```
webbnh, line 84: Should this file have been removed (or never created)?

agent/bench-scripts/tests/pbench-specjbb2005/test-57.txt
```diff
@@ -0,0 +1,96 @@
...
+
+[iterations/2-warehouses:18]
+iteration_number = 2
+iteration_name = 2-warehouses:18
+user_script = specjbb2005
+
+[iterations/3-warehouses:23]
+iteration_number = 3
+iteration_name = 3-warehouses:23
+user_script = specjbb2005
```
webbnh, line 94: So...when going from 18 to 23 by fives, the "23" is inclusive?  (Should we have a test like 18-_22_ by fives?)

webbnh (dismissed):
This looks generally good, but it's a tad long on magic and short on explanation.  Also, I have a few comments.

agent/bench-scripts/tests/pbench-specjbb2005/test-56.pre
dbutenhof: Line 279 to 285 in `pbench-specjbb2005` ... it generates the `.sed` file you commented on elsewhere to patch command line parameters into the template `.props` file. This is an ... interesting ... implementation.

agent/bench-scripts/test-bin/java
```diff
@@ -0,0 +1,35 @@
...
+if sys.argv[-2] != "-id":
+    sys.exit(0)
+
+jvm_id = int(sys.argv[-1])
+
+nr_jvms = None
+start = None
+inc = None
+stop = None
+with open("SPECjbb.props", "r") as fp:
```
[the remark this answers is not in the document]
portante: We could change the name of the script, but how would `pbench-specjbb2005` be able to invoke it?  We'd have to create a symlink in the test `.pre` scripts to create a `java` symlink to the script name.  If you think it is worth doing for this commit, that seems fine to me.

agent/bench-scripts/tests/pbench-specjbb2005/test-56.pre
portante: Sure, I'll add a comment.

agent/bench-scripts/test-bin/java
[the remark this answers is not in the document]
dbutenhof: Oh, right; that's a really good point, and let's not worry about it for now. If the `java` mock needs to handle another Java script, we can worry about making it smarter then. Which probably won't happen. 😁

agent/bench-scripts/tests/pbench-specjbb2005/test-56.pre
webbnh: Thanks, Dave...so, the mechanism has long been there, we're just adding a new engagement of it?  OK, that's cool.

agent/bench-scripts/tests/pbench-specjbb2005/test-56.txt
webbnh: Nevermind:  I thought this was a backup of the original file produced before `sed` modified it (i.e., the result of the `-i` switch, which, apparently, we don't actually use...).  So, this is the actual `sed` script itself....  OK.

agent/bench-scripts/pbench-specjbb2005
portante: Pulling it out of the loop and executing it afterwards is a good idea.

But the use of `${i}` instead of `1` is actually incorrect from the meaning of the value of `i` verses the meaning of the 4th argument of `pbench-tool-trigger`.

The value of `i` is the ID of the worker.  The 4th argument is "which sample number of the given iteration should the tool trigger record under."

For `pbench-specjbb2005`, the `specjbb2005` benchmark is invoked once, and the last JVM's output is processed to determine when the benchmark changes between iterations.  So the fourth parameter should be dropped entirely since it does not need to be provided.  This is also how `pbench-user-benchmark` leverages `pbench-tool-trigger`.

In contrast, `pbench-trafficgen` invokes the benchmark (`trafficgen`) once per sample of an iteration, where it uses `pbench-tool-trigger` to determine where in the potentially long benchmark invocation should performance metrics be captured.  In that case, the iteration number and sample number need to be provided to `pbench-tool-trigger` to have the data recorded appropriately.

agent/bench-scripts/tests/pbench-specjbb2005/test-56.post
portante: I am always nervous about `rm -rf`.

agent/bench-scripts/test-bin/java
portante: Yes, I should write that it is a mock of the expected behavior of `specjbb2005`.

agent/bench-scripts/tests/pbench-specjbb2005/test-57.txt
portante: I don't know exactly how `specjbb2005` handles the parameters, but what I wanted for this test was to have the number of JVMs be not 1 so that the logic of  single vs multiple JVMs in `pbench-specjbb2005` would be exercised.  I then used the start/stop/inc arguments to use a different sequence than the previous test.

portante (commented):
Applied code review feedback, have at it!  Thanks!

dbutenhof (approved):

portante (commented):
Rebased so should be ready for a review.

dbutenhof (approved):

npalaska (approved):

agent/bench-scripts/pbench-specjbb2005
webbnh: Y'see, Peter...this is why we are so desperate to keep you around!

agent/bench-scripts/tests/pbench-specjbb2005/test-56.post
webbnh: You do realize that if `${_dir}` evaluates to an empty string, your approach is actually _more dangerous_ than mine, right?  😉

agent/bench-scripts/tests/pbench-specjbb2005/test-57.txt
webbnh: Yup, I understand and support that.  I'm just wondering whether the mock should be running twice instead of thrice.

PR #2525, opened 2021-11-04
Fix `/api/v1/endpoints` in container environment

dbutenhof:
It may be that we need to ensure that `pbench-server.cfg` has a valid `host` value for other reasons, but `/api/v1/endpoints` is all about making sure an API client (A) knows about all the endpoints we support and (B) can call them reliably using the recommended path (e.g., through an NGINX reverse proxy).

We had advertised endpoints (in the absence of reverse proxy forwarding headers) relative to the static `pbench-server.cfg`
`host` value; this however is unnecessary as the incoming HTTP Requests object includes the actual access path, which makes a much better fallback. This PR changes the endpoints API to use that value, which has the advantage of working within a container regardless of the pod virtual networking configuration.

### DIFF
--- a/lib/pbench/server/api/resources/endpoint_configure.py
+++ b/lib/pbench/server/api/resources/endpoint_configure.py
@@ -31,13 +31,20 @@ class EndpointConfig(Resource):
         Report the server configuration to a web client. By default, the Pbench
         server ansible script sets up a local Apache reverse proxy routing
         through the HTTP port (80); an external reverse-proxy can be configured
-        without the knowledge of the server, and this API will use reverse-proxy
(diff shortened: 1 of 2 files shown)


webbnh (approved):
Seems good.  (I had one comment, but I talked myself out of entering it....)

dbutenhof:
> Seems good. (I had one comment, but I talked myself out of entering it....)

How unusual! Now my curiosity is piqued...

webbnh:
> > Seems good. (I had one comment, but I talked myself out of entering it....)
> 
> How unusual! Now my curiosity is piqued...

Nope!  Not going to go there!  Have other things to do...let's get this merged.  😁

npalaska (approved):

PR #2526, opened 2021-11-05
Ensure fio usage msg sent to stderr as necessary

portante:
Originally part of PR #2524, we correct the usage output to be emitted on `stderr` in error conditions, and to `stdout` for the one non-error condition, `--help`.

Further we remove the annoying trailing semi-colons, and remove the extra `shift` statement while processing options.

Lastly, unrecognized parameters now cause an error.

### DIFF
--- a/agent/bench-scripts/pbench-fio
+++ b/agent/bench-scripts/pbench-fio
@@ -179,143 +179,130 @@ function usage {
 }
 
 function fio_process_options() {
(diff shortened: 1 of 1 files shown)


dbutenhof (approved):
Removing the trailing `;` characters is fine, but not really part of "usage msg send to stderr as necessary" and this particular slice would have been quicker to review without that extraneous cleanup. 😆

portante (commented):
I debated about making separate PRs for each of this changes, but since they are non-structural, no blocks of code moved, just mostly changes on a line or a removal of a line, where the diffs were easy to see side-by-side, it seemed the overhead of more PRs was not worth all the mechanism involved.

webbnh (commented):
This is a good change, but it was way harder to review than it should have been.

For that many semicolon removals, it would have been worthwhile to have them in their own review:  since they shouldn't have any functional impact on the code, we could have just blessed them with a trivial pass; however, having them intermingled with functional changes made it much harder to account for the places where things were actually (functionally) changed.

webbnh (approved):
(Apparently I neglected to tick the "Approve" box on my previous.)

PR #2533, opened 2021-11-08
Refactor `pbench-uperf`

portante:
This is a general refactoring of `pbench-uperf` to address a number of changes which have been found from other PR reviews, but which make it easier to see how changes in `pbench-uperf` line up with other changes in it's sister bench-scripts.


agent/bench-scripts/pbench-uperf
```diff
@@ -200,220 +201,220 @@ if [ $? -ne 0 ]; then
...
-		shift
-		if [ -n "$1" ]; then
-			shift
-		fi
-		warn_log "The --client-label option is deprecated, please use --tool-label-pattern"
+			if [ -n "$1" ]; then
+				shift
+			fi
+			warn_log "The --client-label option is deprecated, please use --tool-label-pattern"
 		;;
```
dbutenhof, line 225: With allowance that you only indented this code, so I won't withhold approval on this basis ...

1. The "deprecated" label seems inappropriate when we're not just "warning that it's going to stop working in the future and recommending that they switch over" ... in fact `--client-label` is **ignored** which, in my mind, is beyond "deprecated". More like "obsolete". E.g., "The --client-label option is obsolete and ignored, please use --tool-label-pattern instead".
2. How long as "deprecated" been in the code? Maybe it's time to just remove this entirely?

agent/bench-scripts/pbench-uperf
```diff
@@ -425,36 +426,39 @@ if [ ! -z "$clients" -a "$postprocess_only" != "y" ]; then
...
-fi
-# before we run a test, verify clients and servers are accessible and install
-# uperf if necessary
-if [ "$postprocess_only" != "y" ]; then
-	# Always install locally
-	# FIXME: why is this necessary?
-	check_install_rpm "${benchmark_rpm}" "${ver}" "${match}"
+
+	if [[ ${check_local} -gt 0 ]]; then
+		check_install_rpm "${benchmark_rpm}" "${ver}" "${match}"
```
portante, line 431: This results in only checking for the installed RPM on the local host when we are actually going to use the local host as a client or server.

agent/bench-scripts/pbench-uperf
```diff
@@ -200,220 +201,220 @@ if [ $? -ne 0 ]; then
...
+fi
+
+if [[ -z "${servers}" ]]; then
+	servers=${servers_def}
+fi
+
+# before we run a test, verify clients and servers are accessible and install
+# uperf if necessary
+if [[ "${postprocess_only}" != "y" ]]; then
+	# Let's count how many localhost clients are servers are specified.
```
dbutenhof, line 381: s/are/and/

agent/bench-scripts/pbench-uperf
```diff
@@ -200,220 +201,220 @@ if [ $? -ne 0 ]; then
...
+if [[ -z "${clients}" ]]; then
+	clients=${clients_def}
+fi
+
+if [[ -z "${servers}" ]]; then
+	servers=${servers_def}
+fi
+
+# before we run a test, verify clients and servers are accessible and install
+# uperf if necessary
```
dbutenhof, line 379: "Install"? Isn't this just the standard `check_install_rpm` check? I like the change you made in fio from `install` to `pre-check`...

agent/bench-scripts/pbench-uperf
```diff
@@ -200,220 +201,220 @@ if [ $? -ne 0 ]; then
...
+if [[ "${postprocess_only}" != "y" ]]; then
+	# Let's count how many localhost clients are servers are specified.
+	let check_local=0
+
+	# Verify the number of clients, servers, and option vsock servers match.
```
dbutenhof, line 384: "Verify that"??

agent/bench-scripts/pbench-uperf
```diff
@@ -200,220 +201,220 @@ if [ $? -ne 0 ]; then
...
+	for client in ${clients//,/ }; do
+		(( c_cnt++ ))
+		if [[ "${client}" == "127.0.0.1" || "${client}" == "localhost" ]]; then
+			(( check_local++ ))
+		fi
 	done
+	if [[ ${c_cnt} -eq 0 ]]; then
+		# No clients on the command line, default to localhost.
+		c_cnt=1
+	fi
```
dbutenhof, line 395: Is this necessary? If `clients` was empty, you used `clients_def`, which should ensure it can never be empty here, so the loop should have always counted at least one.

agent/bench-scripts/pbench-uperf
```diff
@@ -402,36 +426,39 @@ if [ ! -z "$clients" -a "$postprocess_only" != "y" ]; then
...
-	# FIXME: why is this necessary?
-	check_install_rpm "${benchmark_rpm}" "${ver}" "${match}"
+
+	if [[ ${check_local} -gt 0 ]]; then
+		check_install_rpm "${benchmark_rpm}" "${ver}" "${match}"
+	fi
+
 	let err_cnt=0
 	err_clients=""
+	_clients=${clients:-localhost}
```
dbutenhof, line 436: 1. Isn't this redundant since `clients` is already set to `clients_def` (`127.0.0.1`) if it wasn't specified?
2. You don't actually seem to be using `_clients` anyway. Did I miss it?

agent/bench-scripts/pbench-uperf
```diff
@@ -402,36 +426,39 @@ if [ ! -z "$clients" -a "$postprocess_only" != "y" ]; then
...
 	for client in ${clients//,/ }; do
-		debug_log "checking for uperf on client $client"
-		ssh $ssh_opts $client pbench-uperf --install
-		if [ $? -ne 0 ]; then
-			let err_cnt=err_cnt+1
-			err_clients="$err_clients $client"
+		if [[ "${client}" != "127.0.0.1" && "${client}" != "localhost" ]]; then
```
dbutenhof, line 438: You do this so often; and while it's relatively simple, I wonder if a `is_nonlocal` function would make the code just a bit easier to read and maintain? (If nothing else, it eliminates the remote chance of a typo in some obscure path that we'll miss.)

dbutenhof (approved):
Some nits and possible improvements, but functionally this looks OK.

agent/bench-scripts/pbench-uperf
webbnh: That sounds great, but if it was obscure and important enough to warrant a comment here, perhaps it should (have) had one in the code?

agent/bench-scripts/pbench-uperf
webbnh: And, are there other definitions of "local host" which we might need or want to consider?  (I think my container had almost six....)

agent/bench-scripts/pbench-uperf
```diff
@@ -574,11 +578,8 @@ for protocol in `echo $protocols | sed -e s/,/" "/g`; do
 								mkdir -p $benchmark_results_dir
 								server_nr=1
 								# the following loop does all of the pre-benchmark execution work
-								for server in `echo $servers | sed -e s/,/" "/g`; do
-									client=`echo $clients | cut -d, -f$server_nr`
-									if [ -z "$client" ]; then
-										client="127.0.0.1"
-									fi
+								for server in ${servers//,/ }; do
+									client=`echo ${clients} | cut -d, -f$server_nr`
```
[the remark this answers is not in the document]
webbnh: And, as I think about it, we should be able to get rid of the subcommand altogether and do this directly in the shell:
```
									declare -a client_list=(${clients//,/ })
									client=${client_list[$server_nr]}
```
Note that the array is zero-based, while `cut`'s `-f` is one-based, so if you switch to the code here you'll need to change line 579 to initialize `server_nr` to zero instead of `1`.  And, of course, there's no need to put the assignment of `client_list` inside the loop -- it should be done above in some common location.

In fact, instead of keeping the client list as a comma-separated string and iterating over it using `${clients//,/ }`, we could instead keep it as an array and iterate over it using `${clients[@]}`.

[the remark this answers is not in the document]
webbnh: Ditto for `vsock_servers` at line 583 and `server_node` at line 603; and also for `client_nodes` at line 620...and many other places...OK, nevermind...it was a nice thought.

webbnh (dismissed):
I don't think I found anything to actually object to, but I did find a couple of bits of commentary which could be improved, and I had a question or two, and I found lots of fun additional ways to improve the code which you can consider.

However, about halfway through, my eyes began to glaze over and I got distracted, so I fear the quality of my review may have dropped off after that point....

agent/bench-scripts/pbench-uperf
portante: That is coming in PR #2528.

agent/bench-scripts/pbench-uperf
portante: It appears to be unused.  Removed.

agent/bench-scripts/pbench-uperf
```diff
@@ -200,220 +201,220 @@ if [ $? -ne 0 ]; then
...
-	case "$1" in
+	arg="${1}"
+	shift
+	case "${arg}" in
 		--install)
-		shift
-		check_install_rpm "${benchmark_rpm}" "${ver}" "${match}"
-		exit 0
+			check_install_rpm "${benchmark_rpm}" "${ver}" "${match}"
+			exit 0
```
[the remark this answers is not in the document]
portante: Correct.

agent/bench-scripts/pbench-uperf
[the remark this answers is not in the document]
portante: This is the kind of change that I should make in a separate commit in this PR, kept separate, or in a separate PR altogether, but it would be a good change to make, as I can't stand the gratuitous use of pipes to manipulate strings.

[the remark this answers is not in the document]
portante: Or we just re-write in Python3.

agent/bench-scripts/pbench-uperf
```diff
@@ -181,280 +182,295 @@ function gen_xml {
...
+	local server="${1}"
+	local server_port="${2}"
+	local message="${3}"
+
+	local ss_output
+	if is_remote ${server}; then
+		ss_output=$(ssh ${ssh_opts} ${server} ss -tlnp --vsock)
+	else
+		ss_output=$(ss -tlnp --vsock)
+	fi
```
dbutenhof, line 208: I *almost* hate to even suggest this tweak, but ... yeah, I'm going to anyway.

It seems that most of your `is_remote` usage is for exactly this purpose: execute a command, either locally or remotely via `ssh`. So ... *that* could be encapsulated into a function, too, e.g., `execute`, taking a command line string and either executing it directly or via `ssh`.

The result would be much more streamlined and easier to read.

agent/bench-scripts/pbench-uperf
```diff
@@ -181,280 +182,295 @@ function gen_xml {
...
+	for client in ${clients//,/ }; do
+		(( c_cnt++ ))
+		if ! is_remote "${client}"; then
+			check_local=1
+		fi
 	done
+	if [[ ${c_cnt} -eq 0 ]]; then
+		error_log "[${script_name}] INTERNAL ERROR - no clients found"
+		exit 1
+	fi
```
dbutenhof, line 401: It's hard to track the motion of comments across rebases; was this the same redundant check I pointed out before, or a different one? Since `clients` and `servers` are both defaulted to `127.0.0.1`, `c_cnt` can't be 0. Right?

agent/bench-scripts/pbench-uperf
```diff
@@ -181,280 +182,295 @@ function gen_xml {
...
+		(( s_cnt++ ))
+		if ! is_remote "${server}"; then
+			check_local=1
+		fi
 	done
-	if [ $c_cnt -ne $s_cnt ]; then
+	if [[ ${c_cnt} -eq 0 ]]; then
+		error_log "[${script_name}] INTERNAL ERROR - no servers found"
+		exit 1
+	fi
```
dbutenhof, line 412: While this (third? second?) occurrence is not only "redundant", but wrong ... I assume you wanted `s_cnt` here rather than checking `c_cnt` again ... but it shouldn't be possible either way. 😆

agent/bench-scripts/pbench-uperf
```diff
@@ -181,280 +182,295 @@ function gen_xml {
...
-if [ ! -z "$clients" -a "$postprocess_only" != "y" ]; then
+clients=${clients:-${clients_def}}
+servers=${servers:-${servers_def}}
+
+if [[ "${postprocess_only}" != "y" ]]; then
+	# Boolean value recording if local clients or servers are specified.
+	let check_local=0
+
+	# Verify that the number of clients, servers, and optional vsock servers
+	#  match.
```
dbutenhof, line 390: Did you mean to indent `match` two spaces instead of one? It looks inconsistent. 😉

agent/bench-scripts/pbench-uperf
dbutenhof: The recent changes resolve *my* comment, but not necessarily Webb's. Do we want to consider handling other definitions of "localhost" aside from `localhost` and `127.0.0.1`? E.g., `ip addr`, or `::`?? (The latter being especially difficult as there are an almost vast number of valid textual aliases that reduce to `::`, and `ip addr` for example reports it as the fully resolved `00:00:00:00:00:00`. Though IPv6 support is surely at best a PR for another day...)

dbutenhof (commented):
Maybe a few issues still to resolve? Maybe a few unnecessary extra suggestions? 👀

agent/bench-scripts/pbench-uperf
portante: Sort of.  As part of other work, I had noticed this needed to be changed.  You can see it via a `git blame` on the file from the PR branch: https://github.com/portante/pbench/blame/refactor-pbench-uperf/agent/bench-scripts/pbench-uperf

I had fixed this as part of the `is_remote` commit: https://github.com/portante/pbench/commit/39148a0aa9e0e1b8425d835a29c29bb81ea202ef

agent/bench-scripts/pbench-uperf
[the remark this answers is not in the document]
webbnh: A separate, subsequent PR would be excellent.

agent/bench-scripts/pbench-uperf
webbnh: Peter [already suggested basically that](https://github.com/distributed-system-analysis/pbench/pull/2533#discussion_r745576765).  😁

webbnh: But...the encapsulation could and should use the `is_remote` function!

agent/bench-scripts/pbench-uperf
```diff
@@ -416,47 +426,49 @@ if [[ "${postprocess_only}" != "y" ]]; then
...
-		error_log "You must provide vsock-servers when the vsock protocol is specified"
-		exit 1
-	fi
-	if [[ ${v_cnt} -gt 0 && ${v_cnt} -ne ${s_cnt} ]]; then
-		error_log "Number of servers and vsock-servers specified on command line must match"
-		error_log "    servers(${s_cnt}): ${servers}"
-		error_log "    vsock-servers(${v_cnt}): ${vsock_servers}"
+	if [[ ${has_vsock} -eq 1 ]]; then
+		if [[ ${v_cnt} -ne ${s_cnt} ]]; then
+			error_log "When using the vsock protocol, the number of servers and vsock-servers specified on command line must also match"
```
webbnh, line 431: I would drop "also".  And, given the context, I think we can drop "When using the vsock protocol", as well.  (And, we can probably dispense with "on command line", too.)  And, the result will fit in an 80-column line:
```
12345678901234567890123456789012345678901234567890123456789012345678901234567890
The number of servers and vsock-servers specified must match.
```
🙂

webbnh (approved):
Nice improvements!

A few more suggestions below.

agent/bench-scripts/pbench-uperf
portante: Yes, this is an assertion in `bash`.

agent/bench-scripts/pbench-uperf
portante: Yes, there are, but I'd rather not deal with that in this PR as I'd want to leverage a common function for that shared across various consumers in the agent code.

agent/bench-scripts/pbench-uperf
portante: I like the sentiment of this further encapsulation, but it becomes more difficult in `bash` to ensure the command makes its way through the layers of encapsulation to land on the remote host correctly in the general case.

Solving this in Python3 is much cleaner but out of the scope of this PR or of this code base as it stands in general.

agent/bench-scripts/pbench-uperf
```diff
@@ -182,21 +182,45 @@ function gen_xml {
...
+	local _remote
+	if [[ "${arg}" == "127.0.0.1" || "${arg}" == "localhost" ]]; then
+		# False, non-zero return code
+		_remote=1
+	else
+		# True, zero return code
+		_remote=0
+	fi
+	return ${_remote}
+}
```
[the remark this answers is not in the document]
portante: Oh the trickiness of `bash`, ick.

agent/bench-scripts/pbench-uperf
```diff
@@ -367,39 +379,37 @@ while true; do
...
+		if ! is_remote "${client}"; then
+			check_local=1
 		fi
 	done
 	if [[ ${c_cnt} -eq 0 ]]; then
-		# No clients on the command line, default to localhost.
-		c_cnt=1
+		error_log "[${script_name}] INTERNAL ERROR - no clients found"
+		exit 1
 	fi
```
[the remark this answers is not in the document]
portante: The intention is to use this as an assertion.

portante (commented):
Ready for next round of review.

agent/bench-scripts/pbench-uperf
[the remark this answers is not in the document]
webbnh: Actually, that's not just `bash` -- you can do equivalent things in Python, Bliss, and probably C -- whatever is left in the "return value"/"result"/etc. register gets used.  And, `[[` sets it just like any other "command".

dbutenhof (approved):

webbnh (approved):
GTG (once the tests pass)

dbutenhof (approved):

PR #2534, opened 2021-11-08
TimeSeries document query API

npalaska:
- Updates the existing iteration samples API endpoint to serve both iteration (result-data-sample) and timeseries (result-data) documents.
- Two main APIs are as follows:
    - Namespace API serving on `api/v1/dataset/namespace/<iteration/timeseries>`
    - Values API serving on `api/v1/dataset/values/<iteration/timeseries>`

Namespace API:
- Returns the list of available fields in the dataset as well as lists of values available for each of them. These can be used by UI clients to allow users to select documents from the data sent by selecting for certain fields or certain values within those fields.
Example: 
```
{
"benchmark.name":["uperf"],
"benchmark.primary_metric":["Gb_sec"],
"benchmark.protocol":["tcp"],
"iteration.name":[
    "1-tcp_stream-131072B-2i",
    "1-tcp_stream-131072B-2i-fail1"
],
"run.config":["run-config"],
"run.controller":["controller-name.com"],
"run.id":["f3a37c9891a78886639e3bc00e3c5c4e"],
"sample.client_hostname":[
    "127.0.0.1",
    "all"
],
"sample.measurement_title.raw":["Gb_sec"],
"sample.measurement_type":["throughput"],
"sample.name":[
    "sample1",
    "sample2"
],
"sample.uid":[
    "client_hostname:127.0.0.1-server_hostname:127.0.0.1-server_port:20010",
    "client_hostname:all-server_hostname:all-server_port:all"
]
}
```
Values API:
- Returns client supplied document sample values after applying client specified filters.
- Response contains a JSON object with keys as `results` and possibly a `scroll_id` if the next page of results is available.
Example: 

Example:
```
            {
                "scroll_id": "Scroll_id_string", # conditional
                "results": [
                    {
                        "@timestamp": "2020-09-03T01:58:58.712889",
                        "@timestamp_original": "1626277442035",
                        "result_data_sample_parent": "d8ec0b9d6affc5b638169db8b4cbbda2",
                        "run": {...},
                        "iteration": {...},
                        "sample": {...},
                        "result": {...}
                    },
                    {
                        "@timestamp": "2021-03-03T01:58:58.712889",
                        "@timestamp_original": "1626277442038",
                        "result_data_sample_parent": "d8ec0b9d6affc5b638169db8b4cbbda2",
                        "run": {...},
                        "iteration": {...},
                        "sample": {...},
                        "result": {...}
                    },
                    ...
                ]
            }
```


npalaska:
If we feel TimeSeries API class is very similar to the `IterationSamples` class, we could consider merging them together. So we can have `namespace` and `rows` API extended for `result-data` (TimeSeries) documents as well. We could have URI encoded endpoints such as `api/v1/dataset/samples/namespace/<timeseries>` for fetching namespaces for `result-data` document and  `api/v1/dataset/samples/namespace/<iterations>` for fetching namespaces for `result-data-sample` documents and similarly for the `rows` API. 
Let me know your thoughts @dbutenhof @webbnh @portante

dbutenhof:
> If we feel TimeSeries API class is very similar to the `IterationSamples` class, we could consider merging them together. So we can have `namespace` and `rows` API extended for `result-data` (TimeSeries) documents as well. We could have URI encoded endpoints such as `api/v1/dataset/samples/namespace/<timeseries>` for fetching namespaces for `result-data` document and `api/v1/dataset/samples/namespace/<iterations>` for fetching namespaces for `result-data-sample` documents and similarly for the `rows` API. Let me know your thoughts @dbutenhof @webbnh @portante

Yeah, more or less. One of my first comments (and I still haven't gotten through the whole thing, even though you've already hidden some of my comments behind your second commit 😆 ) was that this new `get_mappings` looks very familiar and ought to be generalized.

Again, we'd discussed before that "namespace" is a very general concept, spanning any schema. We could have some sort of general "schema" API that would return document namespaces, API namespaces. etc. We just need to figure out how to structure it, which is the big issue.

If you want to take a first stab at just consolidating the ones you have, I'd recommend something better named than `/datasets/samples/namespace` ... maybe just `/index/namespace/<index>`, possibly even replacing your `/index/mappings`, where the `<index>` parameter may not be a direct Elasticsearch root index name, but possibly a key like `search`, `samples`, `timeseries`. Something to think about anyway.

lib/pbench/server/api/resources/query_apis/metadata_index/timeseries_samples.py
```diff
@@ -0,0 +1,232 @@
...
+        # properties we will not apply that filter in the ES query.
+        for filter, value in json_data.get("filters", {}).items():
+            if filter in mappings.keys() and mappings[filter] == "keyword":
+                # Get all the keyword filters to apply
+                es_filter.append({"match": {filter: value}})
+            elif filter in mappings.keys() and mappings[filter] != "keyword":
+                # Get all the text filters to apply
+                # Note: There is only one text field sample.measurement_title
+                # in result-data documents and if we can re-index it as a
+                # keyword we can get rid of this loop.
```
dbutenhof, line 147: Hmm. Of course in the `result-data-sample` mappings, I'd overlaid `measurement_title.raw` as a keyword; I found that out from working with real Dashboard queries, but getting to the `result-data` index data is a bit more complex and I guess I didn't realize at the time that the field was duplicated.

We could set up the same mapping there, of course, because we're still going to be re-indexing from scratch on our deployments for a while. However, presuming a "namespace" query, it probably wouldn't even be returning a non-keyword field as filterable, so you shouldn't be able to get here. 🤔

lib/pbench/server/api/resources/query_apis/metadata_index/timeseries_samples.py
```diff
@@ -0,0 +1,232 @@
...
+    Parameter,
+    ParamType,
+    Schema,
+)
+from pbench.server.api.resources.query_apis import CONTEXT, PostprocessError
+from pbench.server.api.resources.query_apis.metadata_index import RunIdBase
+from pbench.server.database.models.template import Template, TemplateNotFound
+
+
+class TimeSeriesSamples(RunIdBase):
```
dbutenhof, line 20: Without comparing line-by-line ... it doesn't appear this is much different from the sample data query, except with a different index and document time. Given that the filters are constructed from a namespace query (or from out-of-band knowledge of the schema), could these be consolidated with, for example, a base class `__init__` that takes the root index name to use?

webbnh (commented):
More tomorrow; however...no tests?

lib/pbench/server/api/resources/query_apis/metadata_index/timeseries_samples.py
```diff
@@ -0,0 +1,232 @@
...
+        except (TemplateNotFound, KeyError) as e:
+            self.logger.exception(f"Exception while getting 'result-data' template {e}")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
+
+        es_filter = [{"match": {"run.id": run_id}}]
+        query_strings = []
+
+        # Validate each user provided filters against the result-data document
+        # mappings. If the filter is not present in result-data mappings
+        # properties we will not apply that filter in the ES query.
```
webbnh, line 138: If the filter is not present, then we "fail" silently...I'm not sure I consider this a feature.  I suggest that either
- we validate the filters and return an error for any which aren't in the mappings; or
- we don't validate the filters and let ES deal with it (e.g., by returning an empty result...which I know that Dave won't support 😉).

lib/pbench/server/api/resources/query_apis/metadata_index/timeseries_samples.py
```diff
@@ -0,0 +1,232 @@
...
+                query_strings.append(
+                    {"query_string": {"fields": f"{filter}", "query": f"{value}"}}
+                )
+
+        return {
+            "path": f"/{indices}/_search?scroll={TimeSeriesSamples.SCROLL_EXPIRY}",
+            "kwargs": {
+                "json": {
+                    "size": TimeSeriesSamples.DOCUMENT_SIZE,
+                    "query": {"bool": {"filter": es_filter, "must": query_strings}},
```
webbnh, line 157: Why do you use a `"must"` clause?  Why not include the `query_strings` in the `"filter"` clause?  (Using just `"filter"` avoids scoring.)

lib/pbench/server/api/resources/query_apis/metadata_index/timeseries_samples.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def postprocess(self, es_json: JSON, context: CONTEXT) -> Response:
+        """
+        Returns a Flask Response containing a JSON object with keys as
+        results and possibly a scroll_id if the next page of results
+        is available.
+
+        If there are more than 10,000 documents available for the given filters, then
+        we return the first 10,000 documents along with a scroll id which the client
+        can use to request the next 10,000 documents.
```
webbnh, line 174: It would be nicer if this text were wrapped to 80 columns.

Also, since the number of documents returned is configurable (via `DOCUMENT_SIZE`), it would be better to avoid naming the limit in comments.  You can say something like, "If there is more than one page of documents available [...], we return the first page along with a scroll ID [...]."

lib/pbench/server/api/resources/query_apis/metadata_index/timeseries_samples.py
```diff
@@ -0,0 +1,232 @@
...
+                        "@timestamp_original": "1626277442035",
+                        "result_data_sample_parent": "d8ec0b9d6affc5b638169db8b4cbbda2",
+                        "run": {...},
+                        "iteration": {...},
+                        "sample": {...},
+                        "result": {...}
+                    },
+                    {
+                        "@timestamp": "2021-03-03T01:58:58.712889",
+                        "@timestamp_original": "1626277442038",
```
webbnh, line 194: I realize that this is just sample data, but these values are rather amusing:  we have two timeseries values (in the `@timestamp` fields) which differ, down to the _microsecond_, by exactly six **months**...that was a very long and very precise run!  😁  Also, the values in the `@timestamp_original` fields, which look more reasonable (i.e., they differ from each other by three milliseconds), bear no resemblance to the values in the `@timestamp` fields.  🙂

webbnh (commented):
> If we feel `TimeSeries` API class is very similar to the `IterationSamples` class, we could consider merging them together.

I'm failing to engage with the details adequately to say either yeah or nay to this, but it does seem like there is significant commonality which we should consider hoisting into a common superclass (creating a new one if none of the existing ones seem suitable).

I have a few concerns about the filtering details.  In particular, I expect that the requests for timeseries data are going to want numerical values, so limiting filtering to `keyword` and text fields is going to be counterproductive (and silently ignoring requests for numerical fields is probably unfriendly).

The rest of my comments are mostly editorial and similar nits.

lib/pbench/server/api/resources/query_apis/metadata_index/timeseries_samples.py
```diff
@@ -0,0 +1,232 @@
...
+        """
+        run_id = context["run_id"]
+        dataset = context["dataset"]
+        scroll_id = json_data.get("scroll_id")
+
+        self.logger.info(
+            "Return TimeSeries sample rows {} for dataset {}, prefix {}, "
+            "run id {} next page "
+            if scroll_id
+            else "",
```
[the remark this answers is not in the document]
dbutenhof: Regardless of the exact wording, you're certainly correct that a comma after the format string would help a lot. 😆

lib/pbench/server/api/resources/query_apis/metadata_index/timeseries_samples.py
dbutenhof: But `"filters"` defines which timeseries data points to return, not which items from each datapoint are revealed in the `"_source"` document. Yes, we could *also* allow specifying which non-keyword fields to return to the caller rather than returning all of them, but that's got nothing to do with `"filters"`.

webbnh: Oh, dear...of course you are right:  there are _two_ issues here -- what _documents_ do we want to select, and what _fields_ do we **fetch** from those documents.  And, they both face similar questions.

In the case of document selection, we're (probably) principally interested in `keyword`/text fields (although, if the data fields aren't in the document, it might be handy to avoid selecting it!), and if the request references fields which are not in the mapping, we should probably report an error; however, in the case of selecting fields to return, we're probably interested in the numerical fields rather than the text ones, and it's probably OK if they aren't present in the selected documents.

Nevertheless, it probably would be valuable to be able to stipulate which data fields get returned (i.e., to avoid returning lots of stuff that the client doesn't want).  I dunno...are the different series indexed in separate documents?  (I.e., if I ask for latency data, I'm not going to inadvertently get throughput fields as well?)

npalaska: Yeah if we can update the `measurement_title` mapping to keyword with our new deployment that would be great.

lib/pbench/server/api/resources/query_apis/metadata_index/timeseries_samples.py
```diff
@@ -0,0 +1,232 @@
...
+        template = Template.find(index_name)
+        mappings = template.mappings
+        result = {}
+        for p, m in mappings["properties"].items():
+            if m.get("type"):
+                result[f"{p}"] = m["type"]
+            elif m.get("properties"):
+                for p_inner, m_inner in m["properties"].items():
+                    result[f"{p}.{inner_p}"] = m_inner["type"]
+        return result
```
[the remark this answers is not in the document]
npalaska: Updated, we can actually re-use the `get_keyword_fields` method here instead of `get_mappings` but anyway this file is not tracked anymore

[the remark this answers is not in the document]
npalaska: Removed the function since we can actually re-use the `get_keyword_fields` method

lib/pbench/server/api/resources/query_apis/metadata_index/namespace_and_rows.py
```diff
@@ -66,29 +48,34 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
         """
         run_id = context["run_id"]
         dataset = context["dataset"]
+        document = self.ES_INTERNAL_INDEX_NAMES.get(json_data["target_document"])
```
dbutenhof, line 51: This should be validated, rather than just taking a `KeyError` if they specify a bad value in the URI.

As we discussed, adding a `target_document` KEYWORD Parameter to the class Schema would be a good way to accomplish that.

lib/pbench/server/api/__init__.py
```diff
@@ -85,18 +82,13 @@ def register_endpoints(api, app, config):
...
-        f"{base_uri}/dataset/samples/rows",
+        SampleNamespace,
+        f"{base_uri}/index/namespace/<string:target_document>",
         resource_class_args=(config, logger),
     )
     api.add_resource(
-        TimeSeriesSamples,
-        f"{base_uri}/dataset/samples/timeseries",
+        SampleRows,
+        f"{base_uri}/index/rows/<string:target_document>",
```
dbutenhof, line 91: I think we want this API, at least, to remain under the `/dataset` resource hierarchy. Arguably, though, I think I'm realizing that your `SampleNamespace` should be, too, since while you're returning keywords based on the document mappings, the *values* for each are specific to a dataset. However, I'm having a bit of trouble right now coming up with a good naming hierarchy here.

I do also think that the parameter name `target_document` is misleading: that implies that the dataset name (or run_id) is in the URI. And maybe it *should be* (in a pure REST model), but *this* parameter is really specifying a particular view of the dataset, which is really the granularity of the view (either `result-data-sample` or `result-data`, which, in retrospect, are not particularly well named indices 😆 )

I don't want to tie this to "index", though. Maybe `<string:view>`??

lib/pbench/test/unit/server/query_apis/test_namespace_and_rows.py
```diff
@@ -22,13 +23,65 @@ class TestIterationSamplesNamespace(Commons):
...
+            "iteration.number",
+            "sample.@idx",
+            "sample.name",
+            "sample.measurement_type",
+            "sample.measurement_idx",
+            "sample.uid",
+            "result.@idx",
+            "result.read_or_write",
+            "result.value",
+        ]
```
webbnh, line 83: I feel like this is too literal or realistic to be a good unit test.  E.g., `get_aggregatable_fields()` is completely orthogonal to `WHITELIST_AGGS_FIELDS`, so there should be no need to reference the latter in this test.  The whitelist is only being used because you're starting with a mapping from `Template.find("result-data")`, which is way more ornate than I think this test needs.  And, because the mapping comes from "far away" it's hard for me as a reviewer to tell whether the assertion which checks `result` is correct.

I think this test has only few cases to cover:
 - nested mappings (ideally, at least three levels deep)
 - mappings with `"field"` entries (ideally, at all three levels)
 - mappings which have type `"text"` (ideally, at all three levels)
 - one or more mappings which have a type other than `"text"` (ideally, at all three levels)

If you hand-craft an appropriate mapping dict here in this function, then the assertion should be able to cover everything with a single result, and it will be straightforward for the reviewer to match up the pieces.

lib/pbench/test/unit/server/query_apis/test_namespace_and_rows.py
```diff
@@ -22,13 +23,65 @@ class TestIterationSamplesNamespace(Commons):
...
             payload={"run_id": "random_md5_string1"},
             use_index_from_metadata=True,
         )
 
+    def test_with_no_index_document(self, client, server_config):
+        """
+        Check the Namespace API when no index name is provided
+        """
+        with client:
+            response = client.get(f"{server_config.rest_uri}/dataset/samples/namespace")
```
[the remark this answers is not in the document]
webbnh: Ideally, the query should be sent to `self.pbench_endpoint` with the last component removed.

webbnh (changes requested):
I stumbled a bunch over naming and the conceptual model here.  I think we need to reconsider them and consider renaming some of the URI path components and the variables which represent them.

There are a couple of places where the docstrings need some work.  We need to be cognizant of which "direction" they face:  when they are describing a class or interface, they should avoid implementation details and focus on a functional description; it's only when they are describing a function that they should be talking about implementation.  And, there are a couple of places where they need tweaking so that they actually describe the components of the interface accurately.

And, I might have found a bug or two.

lib/pbench/server/api/resources/query_apis/metadata_index/namespace_and_rows.py
webbnh: Dave, your "this" is ambiguous:  are you talking about checking whether `json_data` has a `"target_document"` key?  (Because Nikhil does handle the case where an existing `"target_document"` key corresponds to a bad value....)

I'm certainly not saying that having a better Schema wouldn't be better, but I think that, by virtue of it being a path parameter, `json_data` will always have a `"target_document"` key here (at least, it will, if Nikhil corrects the code `ApiBase._dispatch()`)....but I could be wrong.

lib/pbench/server/api/resources/__init__.py
```diff
@@ -801,6 +803,8 @@ def _dispatch(self, method: Callable, request: Request) -> Response:
             abort(HTTPStatus.BAD_REQUEST, message="Invalid request payload")
 
         try:
+            if json_data:
+                json_data.update(uri_encoded)
```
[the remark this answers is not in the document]
npalaska: Yeah I think It should be 
```
            if json_data:
                json_data.update(uri_encoded)
            else:
                json_data = uri_encoded
```

lib/pbench/server/api/__init__.py
```diff
@@ -82,13 +82,13 @@ def register_endpoints(api, app, config):
         IndexSearch, f"{base_uri}/index/search", resource_class_args=(config, logger),
     )
     api.add_resource(
-        IterationSampleNamespace,
-        f"{base_uri}/dataset/samples/namespace",
+        SampleNamespace,
+        f"{base_uri}/index/namespace/<string:target_document>",
```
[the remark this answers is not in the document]
npalaska: Yeah I was just trying few thing with naming and gave up at one point I guess, I will add the dataset back to the URI

lib/pbench/server/api/resources/query_apis/metadata_index/namespace_and_rows.py
```diff
@@ -17,43 +16,26 @@
 from pbench.server.database.models.template import Template, TemplateNotFound
 
 
-class IterationSampleNamespace(RunIdBase):
+class SampleNamespace(RunIdBase):
     """
-    Iteration samples API that returns namespace for each whitelisted
-    result-data-sample subdocuments as well as lists of available values for
-    each name.
+    Pbench ES query API that returns namespace for each whitelisted user
```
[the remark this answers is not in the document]
npalaska: Okay your choice of words are much better :smile: but does removing the UI from wording make sense? like These can be used by the clients to select documents from the data they sent by selecting for certain fields or certain values within those fields.

lib/pbench/server/api/resources/query_apis/metadata_index/namespace_and_rows.py
```diff
@@ -17,43 +16,26 @@
...
     """
 
-    WHITELIST_AGGS_FIELDS = ["run", "sample", "iteration", "benchmark"]
-
     def __init__(self, config: PbenchServerConfig, logger: Logger):
         super().__init__(
             config,
             logger,
-            Schema(Parameter("run_id", ParamType.STRING, required=True)),
+            Schema(Parameter("run_id", ParamType.STRING, required=True),),
```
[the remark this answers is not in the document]
npalaska: Black adds it back

lib/pbench/server/api/__init__.py
npalaska: Yeah I will add the `dataset` back in the URI, and `<string:view>` sounds okay but its not really a view its a particular index that we will be searching isn't it? So maybe `<string:type>` (for the type of data). Idk, view is also fine.

lib/pbench/server/api/__init__.py
```diff
@@ -82,13 +82,13 @@ def register_endpoints(api, app, config):
...
         resource_class_args=(config, logger),
     )
     api.add_resource(
-        IterationSamplesRows,
-        f"{base_uri}/dataset/samples/rows",
+        SampleRows,
+        f"{base_uri}/index/rows/<string:target_document>",
```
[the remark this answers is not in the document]
npalaska: Yeah rows is very relational db term but I carried this term forward because we started the discussion on iteration samples API about naming them as `namespace` and `rows`. So thats where this is coming from. I agree we could convert it to `values` since its little more intuitive that rows if we all think that's better.

lib/pbench/server/api/resources/__init__.py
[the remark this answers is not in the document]
webbnh: As long as `json_data` is never `None` and is `{}` when it is empty/unspecified, then
```
            json_data.update(uri_encoded)
```
unconditionally would be nice and simple.

If it can be `None`, then what you suggest would be better.

lib/pbench/server/api/resources/query_apis/metadata_index/namespace_and_rows.py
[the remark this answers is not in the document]
webbnh: My thinking on this is that the "user" (the human interacting with a (G)UI) doesn't interface with the Pbench Server directly; so, I think of the "client" (a piece of software) or a "request"/"requestor" as having that role.  Given that, what is the purpose of this API and/or how would it be used by a client?  The answer is, it returns a list of the fields which the client on behalf of the user can access in the data, as well as a list of possible values, for a given dataset, which can be found in each field.  The exact details of how the client might use those in its interactions with the user are really beyond the scope of this docstring, but we can suggest that they might be used to drive the next API (which is what we intended to do with the Pbench Dashboard).  Although, of course, that API can be used without using this one, if the client already knows what it wants to fetch.

lib/pbench/server/api/__init__.py
webbnh: Yeah, "view" doesn't resonate with me, either.  It really is a "data object type" (e.g., result data, or timeseries data), so I would be inclined toward "type", if that's not a too-overloaded term.

```
/dataset/values/<string:type>
```
?

lib/pbench/server/api/__init__.py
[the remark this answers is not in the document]
webbnh: At the time, we were looking at the Dashboard display which was quite definitely tabular, but that was when we were listing results, IIRC.  Now that we're talking about producing timeseries data, which will probably be plotted in a visualization, "rows" no longer seems appealing.

lib/pbench/server/api/__init__.py
dbutenhof: Our representation of a dataset is spread across multiple Elasticsearch indices for our own internal convenience; but really it's all one dataset, and we're merely offering various "slices" of that dataset as a summary (run), detailed list of captured files (toc), iteration summaries, or low-level timeseries details. In a REST model I have been inclined to consider representing this all as a `GET /api/v2/dataset/<name>?view=timeseries` ... which was what inspired the "view" name here.

We definitely don't want "index" because we want to move away from directly passing on the Elasticsearch model; "type" seems a bit too generic. The fact that this particular aspect of the dataset is represented in a distinct Elasticsearch "document type" (and index) is an implementation detail we'd rather not expose. The REST resource type is "dataset" ... we're just presenting a particular aspect of the dataset here. I'm not tied to "view"  in this context, and it doesn't really fit the original inspiration of using the semi-standard "view" query parameter; but I struggled to find a better term for a moment and then gave up.

We could consider it "level of detail", but that wouldn't be a clear representation, either. On the whole, I think I prefer "view" over "type", or "index" and right now I don't have a better suggestion. But, fine, you guys like "type"; in the context of the whole v1 API, which is a fairly ugly first-pass expedient hack based on the old dashboard rather than a consistent or meaningful REST model, I'll go along with the majority here.

lib/pbench/server/api/__init__.py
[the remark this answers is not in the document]
dbutenhof: Yes, a "namespace" query and a follow-on "values" or "data" query based on that namespace would work.

lib/pbench/server/api/resources/query_apis/metadata_index/namespace_and_rows.py
```diff
@@ -66,29 +52,36 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
         """
         run_id = context["run_id"]
         dataset = context["dataset"]
+        document = self.ES_INTERNAL_INDEX_NAMES.get(json_data["type"])
+        if not document:
+            self.logger.debug("Illegal document name {}", json_data["type"])
+            abort(HTTPStatus.NOT_FOUND, message="Namespace not found")
```
dbutenhof, line 58: This isn't like looking for a user-visible resource, like a Dataset name or a User; this is the client giving an unrecognized keyword. I think this should be `BAD_REQUEST`.

I would have defined your `type` Parameter as a `KEYWORD` with `list(RunIdBase.ES_INTERNAL_INDEX_NAMES.keys())` as the allowed values; which would have already failed with a `BAD_REQUEST` before reaching this code, with an error message consistent with other keyword validation failures.

lib/pbench/server/api/resources/query_apis/metadata_index/namespace_and_rows.py
```diff
@@ -233,13 +225,14 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
                 "run_id": run id of a dataset that client wants to search for
                         fetching the iteration samples.
 
-                "scroll_id": Server provided Elasticsearch scroll id that client
-                            recieved in the result of the original query with
-                            filters. This can be used to fetch the next page of
-                            the result.
+                "scroll_id": Optional Server-provided Elasticsearch scroll id
```
dbutenhof, line 228: I don't think `Server-provided` means anything here; or at least the context is so convoluted that it doesn't help. Simply saying "Elasticsearch scroll id" would be better.

lib/pbench/server/api/resources/query_apis/metadata_index/namespace_and_rows.py
```diff
@@ -270,33 +270,60 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
...
+            )
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="Mapping not found")
+
+        mappings = {
+            "properties": {
+                key: value
+                for key, value in template.mappings["properties"].items()
+                if key in document["whitelist"]
+            }
+        }
```
dbutenhof, line 302: This is common code between your two APIs, which could be abstracted into the base class.

lib/pbench/server/api/resources/query_apis/metadata_index/namespace_and_rows.py
npalaska: Ah good point

dbutenhof (dismissed):
OK, not as much common code moved as I'd hoped ... but I think it's probably a good idea that you left the exception handling in the API method, so I'm satisfied.

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
```diff
@@ -136,3 +137,15 @@ def get_aggregatable_fields(
             for f, v in mappings.get("fields", {}).items():
                 self.get_aggregatable_fields(v, f"{prefix}{f}.", result)
         return result
+
+    def get_mappings(self, document):
```
webbnh, line 141: Type hints?  And, a docstring would be nice, since `document` is not exactly self-describing, and knowing that it's a `dict` doesn't help much either.

lib/pbench/server/api/resources/query_apis/metadata_index/namespace_and_rows.py
```diff
@@ -225,10 +225,10 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
                 "run_id": run id of a dataset that client wants to search for
                         fetching the iteration samples.
 
-                "scroll_id": Optional Server-provided Elasticsearch scroll id
-                            that the client recieved in the result of the
-                            original query. This will, if specified, be used to
-                            fetch the next page of the result.
+                "scroll_id": Optional Elasticsearch scroll id that the client
+                             recieved in the result of the original query.
```
webbnh, line 229: s/recieved/received/

webbnh (dismissed):
So close, but I think it needs one more round of polishing.
[Sorry to be slow...I started this yesterday, got interrupted, and failed to return to it.]

lib/pbench/server/api/resources/__init__.py
```diff
@@ -789,7 +791,7 @@ def _dispatch(
         # We don't accept or process a request payload for GET, or if no
         # parameter schema is defined
         if not self.schema or method == self._get:
-            return method(uri_encoded, request)
+            return method(uri_parameters, request)
 
         try:
             json_data = request.get_json()
```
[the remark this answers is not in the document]
npalaska: So in `_dispatch` the `json_data` immediately goes through the validation (self.schema.validate(json_data)) and if it is `None` or empty `{}`, we raise `InvalidRequestPayload`. Check line 535 of `.../api/resources/__init__.py`. So I dont think it would be a cause of concern, since if the payload is missing and `uri_parameter` is also empty `{}` it will be a `BAD_REQUEST`, and if the payload is missing but  `uri_parameter` is not empty `{"x":"y"}` then it is possible that is might raise `BAD_REQUEST` because of missing parameter in the payload (if there is any required parameter in our class definition, and if not we will proceed with the request).

lib/pbench/server/api/resources/query_apis/metadata_index/namespace_and_rows.py
```diff
@@ -242,11 +248,16 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
         run_id = context["run_id"]
         dataset = context["dataset"]
         scroll_id = json_data.get("scroll_id")
-        document = self.ES_INTERNAL_INDEX_NAMES.get(json_data["target_document"])
+        document = self.ES_INTERNAL_INDEX_NAMES.get(json_data["type"])
+        if not document:
+            self.logger.debug(f"Illegal document name {json_data['type']!r}")
+            abort(HTTPStatus.NOT_FOUND, message="Namespace not found")
```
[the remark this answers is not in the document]
dbutenhof: I thought this had gone away in favor of the keyword check on `"type"` ... and, in any case, it should be `BAD_REQUEST`, not `NOT_FOUND` ... why did I think that had changed last night? Hmm.

lib/pbench/server/api/resources/__init__.py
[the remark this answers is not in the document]
webbnh: What if the request includes a body which has a MIME type of `application/octet-stream`?  My understanding is, even if the body is valid JSON, the call to `request.get_json()` will return `None`.  We won't detect this until the call to `self.schema.validate()` which happens some time later; and, if there are URL parameters which are present, we won't reject the request unless the schema specifies required parameters in the body.  

We should probably specify`force=true` in the `request.get_json()` call if we're not going to reject the request based on the payload MIME type.

My instinct is that `request.get_json()` should return `None` (or raise an exception) on error; otherwise, it should return `{}` or the contents of the payload.  So, I feel like we should be catching and reporting the error (i.e., the `None`) case before proceeding.  I think we should be treating `uri_parameters` the same way.

And, having done that, by the time we get to the following `try` block, we can just add the contents of `uri_parameters` to `json_data` without any conditional and proceed on our way.

lib/pbench/server/api/resources/query_apis/metadata_index/namespace_and_rows.py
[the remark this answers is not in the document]
webbnh: Because there are two of them?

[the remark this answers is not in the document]
npalaska: Yeah I removed it from the Namespace class but not from the Values class, fixing it now

lib/pbench/server/api/resources/__init__.py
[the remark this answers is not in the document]
npalaska: > What if the request includes a body which has a MIME type of `application/octet-stream`?

Yeah as you said we can probably specify `force=true` in the request.get_json() to restrict to `application/json`

But the thing about the `uri_parameter` is that it is either going to be something or the user is going to get `NOT_FOUND` so `uri_parameter` can either be `{}` or `{"x":"y"}` but never `None`. Even with the current code we do check the `request.get_json()` for None and unless all the parameters in the json are validated according to the class schema definition we don't proceed. Its just that we do consume the json as it is and then validate it by our implementation and not Flask's framework.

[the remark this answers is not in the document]
webbnh: > Yeah as you said we can probably specify `force=true` in the `request.get_json()` to restrict to `application/json`

By default, `request.get_json()` [will return `None`](https://flask.palletsprojects.com/en/latest/api/#flask.Request.get_json) if the body MIME type is not `application/json` (i.e., `force` is `false` by default); we would specify `force=true` if we wanted to be able to ignore the MIME type.

Being able to depend on the `uri_parameter` being a `dict` in our code is handy.  I would like to have the same assurance for `json_data`, not only for convenience but also because I think the cases in which it is not a `dict` represent errors which we want to respond to and not ignore or cover over.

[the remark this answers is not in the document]
dbutenhof: Yeah, I've thought several times about enforcing `Content-Type` and even `Accepts` since we *return* JSON. It'd actually be cool if Flask would transparently handle `dict` <-> (`application/json` | `application|xml`), but I don't think it does and it's not really a big deal. But we probably shouldn't be just silently assuming JSON everywhere, and I never really got motivated to try to do anything about it.

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
```diff
@@ -138,7 +138,17 @@ def get_aggregatable_fields(
                 self.get_aggregatable_fields(v, f"{prefix}{f}.", result)
         return result
 
-    def get_mappings(self, document):
+    def get_mappings(self, document: JSON):
+        """
+        Utility function to return ES mappings by querying the Template
+        database against a given index.
+
+        :Args
```
webbnh, line 146: Your colon ended up on the wrong end of "Args".  :-)

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
```diff
@@ -138,7 +138,17 @@ def get_aggregatable_fields(
                 self.get_aggregatable_fields(v, f"{prefix}{f}.", result)
         return result
 
-    def get_mappings(self, document):
+    def get_mappings(self, document: JSON):
```
webbnh, line 141: Type hit for the return value?  (I think you can use `-> JSON`, although I'm not sure that it's not returning a `dict`...).

lib/pbench/server/api/resources/query_apis/metadata_index/namespace_and_rows.py
```diff
@@ -249,9 +249,6 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
         dataset = context["dataset"]
         scroll_id = json_data.get("scroll_id")
         document = self.ES_INTERNAL_INDEX_NAMES.get(json_data["type"])
-        if not document:
-            self.logger.debug(f"Illegal document name {json_data['type']!r}")
-            abort(HTTPStatus.NOT_FOUND, message="Namespace not found")
 
         document_index = document["index"]
```
webbnh: Since you're no longer checking the value returned by `get()` to see if it is `None`, you should replace `get()` with `[]`-access:  that way, if something goes wrong, you get a nice, clear exception at the point of error, instead of getting a confusing exception from the next line.

webbnh (approved):
Two superficial problems and one nearly-superficial one.

It's up to you (and Dave? 😉) whether to fix or skip.

lib/pbench/server/api/resources/query_apis/metadata_index/__init__.py
npalaska: Ah, yeah

webbnh:
@npalaska, there's [one lingering comment](https://github.com/distributed-system-analysis/pbench/pull/2534/files/f76dd8caa28f86593381b7b6c8af7c0d516533cc#diff-d987ce8ebac0ab4f0756443fd944f6271d48e0635975c29f6d50eb8abfb3ad99) way back which, if you change other things, you should consider addressing.

And, @dbutenhof, I'm interested in your thoughts on [another comment above](https://github.com/distributed-system-analysis/pbench/pull/2534#discussion_r752654594), although, perhaps, at this point, it's beyond the scope of this PR.

(And, somehow we need to come to resolution on [this one](https://github.com/distributed-system-analysis/pbench/pull/2534#discussion_r752646934).)

lib/pbench/server/api/resources/__init__.py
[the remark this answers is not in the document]
npalaska: Yeah I agree, but at this time its not catastrophic if the `get_json` returns `None` because if the conditional statement on line 813, but yes we should be enforcing `Content-Type` to be `application/json`. I will work on it separately since I think it might need little bit experimentation

lib/pbench/server/api/resources/query_apis/metadata_index/namespace_and_rows.py
```diff
@@ -18,24 +18,28 @@
...
 
     def __init__(self, config: PbenchServerConfig, logger: Logger):
         super().__init__(
             config,
             logger,
-            Schema(Parameter("run_id", ParamType.STRING, required=True),),
+            Schema(
+                Parameter("run_id", ParamType.STRING, required=True),
+                Parameter("type", ParamType.STRING, required=True, uri_parameters=True),
+            ),
```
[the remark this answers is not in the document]
dbutenhof: We don't have a resource-oriented REST API model at all. I struggled to create one, but at the time it was a matter of how much of the dashboard was going to be massively restructured ... and implementing functional APIs mapping (even roughly) to the dashboard structure made a lot more sense at the time.

If we're really ready to toss the current dashboard, I'd love to revisit that and come up with a clean resource model for which we can build a real REST API, and that includes consistent resource IDs that are part of the URI; and possibly query parameters for most of the other stuff. (For the large filter constructs, yeah, maybe we continue with a JSON payload, but the POST/GET/PUT should be to a REST URI not a generic endpoint.)

I also don't really like that we strayed from "dataset name" to "run id", because they're functionally equivalent and the inconsistency is an annoyance. Again, I left this issue alone because we were basing the "v1" API on the current dashboard. If we were starting this API now, I'd probably have strongly urged dropping the run ID and doing the query by dataset name... but there are worse inconsistencies, and I decided not to derail things mid-stream by bringing it up at all. (Until now, apparently 😆 )

I consider the v1 API an agile PoC, a throwaway in the long run; again, I would have pushed for more architecture if we hadn't been worrying mostly about a short path to user-authenticated queries through the existing dashboard. At this point, I'm not worried.

npalaska:
> @npalaska, there's [one lingering comment](https://github.com/distributed-system-analysis/pbench/pull/2534/files/f76dd8caa28f86593381b7b6c8af7c0d516533cc#diff-d987ce8ebac0ab4f0756443fd944f6271d48e0635975c29f6d50eb8abfb3ad99) way back which, if you change other things, you should consider addressing.

I can not open it for some reason, github says it couldn’t find those commits

dbutenhof (approved):
Lots of open conversations by that Webb guy, but it's looking good.

webbnh (changes requested):
Looks good to go.

(The lingering conversation was [here](https://github.com/distributed-system-analysis/pbench/pull/2534#discussion_r746724325)...GitHub has two URLs for open conversations, and somehow I pasted in the useless one...it's such a small thing that it's not worth bothering with, now.)

webbnh (approved):
And, this time, click on the **correct** radio button.

PR #2536, opened 2021-11-09
Increase Prom retention in visualizers

jeniferh:
Increase Prom retention to 1yr. 

This is especially useful for the "post" PromGrafVisualizer option where you may be looking at older data. Without an increase in retention, when running the visualizer image Prom can delete any original data (provided by the volume mount) older than the default 15d.

### DIFF
--- a/agent/containers/images/visualizers/run.py
+++ b/agent/containers/images/visualizers/run.py
@@ -64,6 +64,7 @@ elif metric_type == "prom":
         "./prometheus",
         "--config.file=/data/prometheus.yml",
         "--storage.tsdb.path=/data",
+        "--storage.tsdb.retention.time=1y",
     ]
     collector = subprocess.Popen(cmd)
 


jeniferh:
@Maxusmusti thoughts on this?

Maxusmusti (commented):
LGTM, makes sense. I'll build and test to make sure it's all still working as expected, but it seems like it should be fine.

jeniferh:
Thanks! I tested locally so hopefully no surprises.

portante (approved):
This is a good short-term fix, but the visualizer should find a way to ensure when looking at older data sets they are not modified in any way.  Having something automatic, which would not require a hard-coded setting would be more supportable.

webbnh (approved):
I hope that 1 year is long enough...is there any reason not to make it longer?  (Is there an "infinity" or "no-limit" value?)

jeniferh:
I will poke around to see if there are other ways to extend Prom retention.. but I figured if the data is older than 1yr is it even that relevant anyway? ;)

jeniferh:
I did try a readonly volume mount but that won't work:
`caller=query_logger.go:87 component=activeQueryTracker msg="Error opening query log file" file=/data/queries.active err="open /data/queries.active: read-only file system"`

portante:
@Maxusmusti said he was going to review.  I'll merge following that review.

Maxusmusti (changes requested):
I've built and pushed updated images, everything should be good to go (after a rebase, branch out of date). I've also had to change PCP version for pcp-graf-visualizer, so you could go into DockerfilePostPCP and change to `ARG PCP_VERSION="5.3.4-2.fc33"` (otherwise I'll do it in a seperate PR after this is merged).

portante:
> I've also had to change PCP version for pcp-graf-visualizer, so you could go into DockerfilePostPCP and change to `ARG PCP_VERSION="5.3.4-2.fc33"` (otherwise I'll do it in a seperate PR after this is merged).

@Maxusmusti, could we consider approving this now and work that in a separate PR?

Maxusmusti:
> > I've also had to change PCP version for pcp-graf-visualizer, so you could go into DockerfilePostPCP and change to `ARG PCP_VERSION="5.3.4-2.fc33"` (otherwise I'll do it in a seperate PR after this is merged).
> 
> @Maxusmusti, could we consider approving this now and work that in a separate PR?

Sounds good, just the rebase then 👍

PR #2542, opened 2021-11-13
Canonically determine local vs remote hosts/ips

portante:
We need to properly determine whether a given address for a host, IP or name, refers to the local host or to a remote host.

In all the case where it is used (tool registration, Tool Meister sub-system, bench-scripts (`pbench-fio` and `pbench-uperf`, primarily), we need to record at the time a bench-script is run which address is local vs remote.

We also need to make sure that we don't register tools for the same local host under different names.

We also need to ensure bench-scripts, i.e. `pbench-uperf` and `pbench-fio`, know the difference between local and remote hosts to avoid `ssh` operations where possible.


lib/pbench/agent/utils.py
```diff
@@ -390,3 +392,32 @@ def collect_local_info(pbench_bin):
         hostdata[arg] = cp.stdout.strip() if cp.stdout is not None else ""
 
     return (version, seqno, sha1, hostdata)
+
+
+class LocalRemoteHost:
+    """Dynamically build a list of all the local IP addresses so that a one of
```
dbutenhof, line 398: "so that one of two" not "a one of two" 😸  Then again, one could actually make *both* queries, if one wanted to, so maybe just "supporting two simple queries"?

I'm also wondering how many times the caller really knows (or should have to care) whether whatever string it's got is a raw ip address vs a hostname. Granted, there's some optimization here when it *is* known to be an ip, but what's the real benefit of the complication? (Given that accidentally passing a hostname to the ip form can return a false negative.)

lib/pbench/agent/utils.py
```diff
@@ -390,3 +392,32 @@ def collect_local_info(pbench_bin):
...
+    two simple queries can be made: is the given host name remote?  is the
+    given IP address remote?
+    """
+
+    def __init__(self):
+        """Build up a list of local IP addresses from all the interfaces.
+        """
+        local_ips_l = []
+        for ifaceName in interfaces():
+            for ifaddr in ifaddresses(ifaceName).setdefault(AF_INET, [{"addr": None}]):
```
dbutenhof, line 408: I hate to say this ... but although we pretty much use only IPv4 internally (which is honestly weird), I'm not sure we want to think about releasing an agent (especially as a "product") without IPv6 support.

(Although I also recognize that's a vastly bigger issue than just this code, I'd feel vaguely better if we didn't write additional code that delberately precluded IPv6 support.)

dbutenhof (commented):
Just a minor comment comment and a major philosophical rabbit hole ... that's all. 🤣

lib/pbench/agent/utils.py
portante: This doesn't. I just creates a structure which returns `{"addr": None}` when the interface does not have any IP address configured.

portante (commented):
Keep in mind this is still a `Draft` PR, so the interface may change as I continue to work at how we can leverage it throughout our code.

lib/pbench/agent/utils.py
webbnh: So, the `setdefault()` call is intended to avoid a `KeyError` at line 410 and trigger the `continue` at line 412?

This seems a little heavy-handed, not to mention sort of a violation encapsulation:  `setdefault()` actually modifies the dict to which it is applied.  Also, it doesn't necessarily make sense for some interfaces to even have an `AF_INET` entry.

Why not just do the `if "addr" not in ifaddr: continue` test and skip it that way?

lib/pbench/agent/utils.py
```diff
@@ -390,3 +393,36 @@ def collect_local_info(pbench_bin):
...
+                ip_addr = ifaddr["addr"]
+                if ip_addr is None:
+                    continue
+                local_ips_l.append(ip_addr)
+        self.local_ips = frozenset(local_ips_l)
+
+    def is_remote(self, host_name):
+        """Returns True if the given host name (or IP address) is remote else
+        returns False.
+        """
```
webbnh, line 419: It's a bit pedantic, but I suggest that this function returns `True` if the given host name or IP address is not recognized as referring to the local host.  (We cannot really tell whether it is remote or not without requesting the routing information, and, even then, it might be routed back to the current host...so, all that we can tell is that we don't think it's us.)

lib/pbench/agent/utils.py
```diff
@@ -390,3 +393,36 @@ def collect_local_info(pbench_bin):
...
+        try:
+            ip_addr = socket.gethostbyname(host_name)
+        except socket.gaierror:
```
webbnh, line 422: I don't know how expensive `gethostbyname()` is, but it might be worth a simple pre-filter:  if `host_name` doesn't start with an letter ([a-zA-Z]), then I don't think it can be a hostname; if it doesn't start with a hex-digit (or a colon or a left-square-bracket), then I don't think it can be an IP address; so we might be able to detect most of the error cases locally up front, and save the `try` blocks for the truly exceptional cases.

(You might also want to expand the function docstring to point out that errors result in `True` being returned.  Although, if we use a pre-filter to differentiate between names and addressses, we could just let the exceptions propagate and omit the `try` blocks.)

lib/pbench/agent/utils.py
webbnh: Also, I second Dave's concern; however, I think this will support IPv6 out of the box, but it would be good if the test code proved that.

webbnh (commented):
Looks like a good start.

My principal reservation is that the `netifaces` package is not a "pure-Python" package (it of necessity contains platform-specific code) and does not currently have a maintainer, not to mention the fact that it's not available (as far as we know) as an RPM.  So, while it's a very cool package, we need to ensure that our dependency on it is highly encapsulated, so that we can rip it out if needed.  (E.g., if we have to, we could replace it with a small set of heuristics.)

lib/pbench/agent/utils.py
```diff
@@ -390,3 +393,36 @@ def collect_local_info(pbench_bin):
...
+                    continue
+                local_ips_l.append(ip_addr)
+        self.local_ips = frozenset(local_ips_l)
+
+    def is_remote(self, host_name):
+        """Returns True if the given host name (or IP address) is remote else
+        returns False.
+        """
+        try:
+            ip_addr = socket.gethostbyname(host_name)
```
dbutenhof, line 421: `socket.gethostbyname` returns a *single* IPv4 address, according to the documentation. To get IPv6 as well, or multiple addresses, you apparently need to use `socket.getaddrinfo` instead.

I'm also trying to figure out what's documented, if anything, about the format of an IPv6 address reported here; the documentation examples I've seen are all in full 8-group format without an shortening, but I'm pretty sure from other contexts that it can return shortened names (e.g., `::1` for `localhost`). And, if so, we need to be careful about tests like `ip_address in self.local_ips`, because `0:0:0:0:0:0:0:1` and `0000:0000:0000:0000:0000:0000:0000:0001` need to match `::1` as well as `::0000:0:1` and other forms.

Python's `ipaddress` package looks a lot like the Java version I've used before: `ipaddress.ip_address(address)` is a factory creating either an `IPv4Address` object or an `IPv6Address` object depending on the value, and the objects have comparison operators so `x == a` works regardless of the original format. It's possible that your `netifaces` can do that, too, or tie into it. Anyway, the simple solution would be to feed each IP address into `ipaddress.ip_address`, store the list of resulting objects (`IPv4Address` and `IPv6Address`), the convert the input parameter as well and compare those rather than the raw text.

What you've got here will work only if you get lucky, and someone supplies the exact same form you got back from `netifaces`. If you want to explicitly say "we're not going to worry about IPv6 addresses for this spin", I won't object; but we'll definitely have to deal with that at some point.

lib/pbench/agent/utils.py
dbutenhof: The after-the-fact filter to fill in empties is ... interesting, and definitely not obvious. In any case, this definitely won't (fully or even usefully) support IPv6 "out of the box" (even if the format returned by `netifaces` is guaranteed, the input value you're trying to compare definitely isn't) -- see comment below.

But if this `AF_INET` is just a filter to fill in for interfaces without any value ... um, weird ... but I guess irrelevant. 👀

lib/pbench/agent/utils.py
webbnh: Hmmm...what I think we're trying to determine is whether sending a packet to `host_name` will use local transport or whether it will go out through a NIC; and, to do that we are trying to match "the address" for `host_name` to one of the addresses for the local host.  It's not clear to me whether the fact that there might be multiple address translations for `host_name` makes that harder or not -- it seems likely to me that if any of the translations are non-local then they'll all be non-local, so any one of them is as good for our purposes as the others.  (Of course, I could just be confused....)

However, the problem of comparing IPv6 addresses is very real:  they need to be cast into a canonical format before they can be compared.  If `ipaddress` would handle that for us, that would be good.

lib/pbench/agent/utils.py
webbnh: The "after-the-fact filter" just keeps the code from barfing when it hits an interface which has no `"addr"` field; it doesn't help or hurt things along the IPv6 dimension.  If the caller happens to specify an IPv6 address in the same format that `netifaces` uses, then I think the code will work as intended.  (And, as I said below, if we enhance the comparison to use `ipaddress` as you suggested, then I think we're covered.)

lib/pbench/agent/utils.py
dbutenhof: In an ideal universe (is this one of those?), all addresses for the local host will be local. However, in an IPv6-only network, `gethostbyname` is probably useless.

agent/requirements.txt
```diff
@@ -1,6 +1,7 @@
 bottle
 cffi
 click>=8.0
+hiredis
```
dbutenhof, line 4: OK ... why *do* we need this package we apparently don't reference directly? I see that it's some sort of Redis wrapper or extension, but with no references elsewhere ... what *is* going on here?

lib/pbench/agent/utils.py
webbnh: But, in this usage, we're applying `gethostbyname()` to the _target_ address, not the local host.  (We created the list of local addresses in the c'tor.)  So, assuming that the target's addresses are either all local or all remote, it should work to arbitrarily use one of them to compare against the list of local addresses (modulo, obviously, address format issues).

dbutenhof: As long as the target host *has* an IPv4 address, sure; but eventually we'll have to deal with an environment where it doesn't, and `getaddrinfo` avoids that issue by supporting both from the beginning. You're correct that we should be able to take any one of the addresses it returns since we have all `localhost` aliases.

webbnh: > As long as the target host has an IPv4 address

OK, I'd neglected the fact that `gethostbyname()` [only returns IPv4 addresses](https://docs.python.org/3/library/socket.html#socket.gethostbyname).  _[Le sigh]_  So, yeah, @portante, you should consider using [`getaddrinfo()`](https://docs.python.org/3/library/socket.html#socket.getaddrinfo), instead.  (And, we need to confirm that what `ifaddresses()` returns for the local host matches what `getaddrinfo()` returns for it if it has IPv6 configured, or we have to convert the formats.)

agent/requirements.txt
portante: Ugh: https://github.com/redis/redis-py/commit/16cfcc7fced84d2b53edf95af1c40b230b30fc3d

portante: I'll file a separate PR for this fix.

portante: See PR #2547.

lib/pbench/agent/utils.py
portante: This will be addressed in the new revision.

lib/pbench/agent/utils.py
```diff
@@ -390,3 +393,36 @@ def collect_local_info(pbench_bin):
...
+    def __init__(self):
+        """Build up a list of local IP addresses from all the interfaces.
+        """
+        local_ips_l = []
+        for ifaceName in interfaces():
+            for ifaddr in ifaddresses(ifaceName).setdefault(AF_INET, [{"addr": None}]):
+                ip_addr = ifaddr["addr"]
+                if ip_addr is None:
+                    continue
+                local_ips_l.append(ip_addr)
```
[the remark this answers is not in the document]
portante: Addressed via the new use of the `ifaddr` module.

lib/pbench/agent/utils.py
portante: Addressed in the new implmentation.

lib/pbench/agent/utils.py
```diff
@@ -390,3 +393,60 @@ def collect_local_info(pbench_bin):
...
+                    "testhost.example.com",
+                    "testhost",
+                ]
+            )
+
+        def is_remote(self, host_name):
+            return host_name not in self._local_names
+
+
+else:
```
dbutenhof, line 419: Ugh. But, yeah; there's no way in the legacy test structure to mock external Python calls. ☹️ 

Even so, I'd prefer a smaller hack "mocking" `ifaddr.get_adapters` and `socket.getaddrinfo` so that we're testing most of the logic here.

Even if it's just something like

```
if os.environ.get("_PBENCH_UNIT_TESTS"):
    get_adapters = ifaddr.get_adapters
    getaddrinfo = socket.getaddrinfo
else:
    get_adapters = my_local_hack1
    getaddrinfo = lambda a, b: return a
```

lib/pbench/agent/utils.py
```diff
@@ -390,3 +393,60 @@ def collect_local_info(pbench_bin):
...
+
+        def is_remote(self, host_name):
+            """Returns True if the given host name (or IP address) is remote else
+            returns False.
+            """
+            ips_l = []
+            try:
+                ip_addrs = socket.getaddrinfo(host_name, None)
+                for addrinfo in ip_addrs:
+                    ips_l.append(addrinfo[4][0])
```
webbnh, line 444: FWIW, I don't think we really want _all_ the addresses...we just want the one that `ssh` is going to use.  I'm not sure how to tell which one that is, but my suspicion is that if we work with more than one address we're likely to produce the wrong result, so I would just grab the first one, I think.

lib/pbench/agent/utils.py
```diff
@@ -390,3 +393,60 @@ def collect_local_info(pbench_bin):
...
+            except socket.gaierror:
+                try:
+                    ipaddress.ip_address(host_name)
+                except ValueError:
+                    return True
+                ips_l.append(host_name)
+            ips = frozenset(ips_l)
+            return self.local_ips.isdisjoint(ips)
```
webbnh, line 452: I think that we want `is_remote()` to return `True` unless we're sure that `ssh` would be hitting the local host.  So, if _any_ of the translations aren't in the local set, then I think we should probably return `True`.  So, `isdisjoint()` isn't the right operation -- we need something more like `issubset()`.

(Of course, if we can limit our comparison to a single target address, then we can just use `in` and dispense with producing `frozenset`'s here....)

lib/pbench/agent/utils.py
webbnh: If you're going to stick with generating the whole set, a comprehension would be nicer here:
```
                ips = frozenset(addrinfo[4][0] for addrinfo in ip_addrs)
```

lib/pbench/agent/utils.py
```diff
@@ -390,3 +393,60 @@ def collect_local_info(pbench_bin):
...
+                ip_addrs = socket.getaddrinfo(host_name, None)
+                for addrinfo in ip_addrs:
+                    ips_l.append(addrinfo[4][0])
+            except socket.gaierror:
+                try:
+                    ipaddress.ip_address(host_name)
+                except ValueError:
+                    return True
+                ips_l.append(host_name)
+            ips = frozenset(ips_l)
```
webbnh, line 451: There's no need to do an append operation -- we can just assign a new one to `ips_l`, and while we're at it, we can skip the middleman and make it a frozenset right off the bat and skip the assignment at line 451 if we do the analogous thing at line 443/444:
```
                ips = frozenset({host_name})
```

lib/pbench/agent/utils.py
```diff
@@ -390,3 +393,60 @@ def collect_local_info(pbench_bin):
...
+        def is_remote(self, host_name):
+            """Returns True if the given host name (or IP address) is remote else
+            returns False.
+            """
+            ips_l = []
+            try:
+                ip_addrs = socket.getaddrinfo(host_name, None)
+                for addrinfo in ip_addrs:
+                    ips_l.append(addrinfo[4][0])
+            except socket.gaierror:
```
webbnh, line 445: A quick look at the docs suggests that, other than `EAI_AGAIN`, these errors are either catastrophic or in the "should never happen" category for our usage, so it doesn't seem worthwhile to maintain `ip_address` as a fall-back (especially since it will only work if the input was numeric).  More to the point, it strikes me that this `except` block is going to eat errors that we would otherwise want to know about.

So, if you really want to maintain the fallback, I suggest that we should be be using it only if the `gaierror` is reporting an `EAI_AGAIN`.

webbnh (changes requested):
- [x] The mocking for `LocalRemoteHost` should be done in some other way.
- [x] `pbench-is-remote` should have a proper docstring.
- [x] I think we should reconsider whether/how we work with _all_ of the target addresses (when there is more than one); and, I think we should give due consideration to what the function should return in the "failure case" (i.e., if we get an ambiguous result).
- [x] I think I might have found a missing `else`.
- [x] Consider renaming things from "remote" to (not) "local".
- [ ] Consider DRYing out a command invocation.
- [x] Why use symlinks for the test completion status files?  (If one changes, they aren't all going to change...are they?...so they shouldn't be coupled to each other.)

Beyond those, there are a dozen smaller items.

lib/pbench/agent/utils.py
dbutenhof: Just to throw this out here for "posterity" (whomever that may turn out to be), I needed this PR's basic logic but, partially by way of fixing some of the weaknesses here (especially in IPv6), and partially just because I felt like it, I put my own spin on it... including reversing the test. (Mostly because the usage in the code ends up being "is local" vs "not is remote".)

I haven't done unit testing, nor tried to accommodate the legacy test environment variable, but I think there are two important aspects to contribute to this PR:

1. Proper handling of IPv6 comparison using `ipaddress` rather than string matches;
2. Not trying to do some sort of full set match but rather looking only for proof that (at least) one translation of the parameter hostname is a match for a `localhost` representation. (Webb may have a point about making sure that *all* host representations are in `localhost` but I'm not at all convinced the sort of split routing that implies is feasible and I didn't try to deal with it...)

__NOTE__ that I'm ignoring all the translations that aren't INET or INET6 family as that's all we care about. I'm also ignoring anything with a non-zero IPv6 scope_id on the principle that they're probably not interesting or useful, as well as the pragmatic aspect that these will include the `%scope` suffix on the text IPv6 representation, which `ipaddress.ip_address()` can't parse. It's not entirely impossible that we could have a captive subnet on one interface that's the only path to our remote nodes, and I'm not certain whether that's realistic; but for now I decided not to bother with it.

In short, I'm throwing this out as a follow-up comment, as sample, and as a prototype, not as the answer to life, the universe, and everything.

```
class HostComparator:
    """
    Determine whether a given hostname or IP is an alias for localhost.
    """

    def __init__(self):
        """
        Build a list of local IP addresses for all interfaces
        """
        ips = []
        for adapter in ifaddr.get_adapters():
            for ip in adapter.ips:
                if isinstance(ip.ip, tuple):
                    if len(ip.ip) > 2 and ip.ip[2] == 0:
                        # This is an IPv6 address: use it if the scope ID is 0
                        # (That is, it's routable through any interface)
                        ips.append(ipaddress.ip_address(ip.ip[0]))
                else:
                    ips.append(ipaddress.ip_address(ip.ip))
        self.aliases = frozenset(ips)

    def is_local(self, host: str) -> bool:
        """
        Determine whether a given hostname is an alias for the local host; in
        other words, whether it matches a canonical representation of one of
        the routable addresses reported by `ifaddr`.

        Args:
            host: A hostname or IP address to check

        Returns:
            True if `host` matches a routable local host address
        """
        try:
            infos = socket.getaddrinfo(host, None)

            # getaddrinfo returns 5-tuples of names for the host in the format
            # (family, type, proto, name, (address, port[, flowinfo, scopeid]))
            # where flowinfo and scopeid are present only for IPv6. We want to
            # capture all IPv4 addresses plus all IPv6 addresses without a
            # scope ID. If any of the resulting addresses match against the
            # "localhost" alias set, return true.
            for info in infos:
                proto = info[0]
                addr = info[4]
                if proto == socket.AF_INET or (proto == socket.AF_INET6 and addr[3] == 0):
                    ip = ipaddress.ip_address(addr[0])
                    if ip in self.aliases:
                        return True
        except socket.gaierror:
            # If address lookup fails, check whether the hostname is legal
            try:
                addr = ipaddress.ip_address(host)
                return addr in self.aliases
            except ValueError:
                return False
```

agent/bench-scripts/pbench-fio
```diff
@@ -588,11 +599,15 @@ function fio_run_job() {
 	echo "running fio job: $fio_job_file ($(basename $benchmark_results_dir))"
 
 	if [ -n "${clients}" ]; then
```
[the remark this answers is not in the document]
dbutenhof: That's a really good question. I'd like to have @portante think about this. I *think* that's correct, because the documented behavior is to run locally if there are no clients specified. As is, I suspect we're skipping the `do_local_client` actions in that case, and we probably shouldn't.

I'd love to have confirmation before fixing that and updating more gold files (which currently fail due to the extra debug messages).

lib/pbench/agent/utils.py
dbutenhof: OK, I thought about this but I don't remember whether I forgot or just decided to be more conservative about my changes.

dbutenhof (commented):
I've rebased and updated this to handle the subjectively "biggest" issues:

- [ ] IPv6 support (non-text comparisons);
- [ ] Changing the sense of the test to explicitly support the fact that we can't/don't really test for "remote" but whether a host matches a known local address.

There are a few other issues worth considering here; but I'd also like to get this in because my non-DNS tool meister changes require it. (Which is why I "stole" it from Peter after it's been languishing in the GitHub closet since November...)

dbutenhof:
Argh; bug revealed by Jenkins IPv6 configuration, which is interesting. I think I need to try to mock `ifaddr.get_adapters()` ... which I'd lazily tried to avoid. ☹️

webbnh (changes requested):
The new changes appear to have resolved most of my previous concerns.  I have a few new comments below.  And I still have some conversations from earlier passes:

- [ ] I think the mocking should be done in a different way.  (@dbutenhof, and I discussed some possibilities -- my current favorite is to have the class to define both the real methods and mocks for them, and to bind the appropriate definition to the name at load-time.)
- [ ] There's the question about the missing else in the fio script
- [ ] My concern about using symlinks for the test completion status files

agent/bench-scripts/pbench-fio
```diff
@@ -908,9 +929,16 @@ if [ "$postprocess_only" = "n" ]; then
...
+			if ! pbench-is-remote ${client}; then
+				(( local_client_count++ ))
+			fi
 			client_names+=("${client_name}")
 			client_ports["${client_name}"]=${client_port}
 		done
+		if [[ ${local_client_count} -gt 1 ]]; then
+			error_log "${script_name}: more than one client refers to the local host"
+			exit 1
+		fi
```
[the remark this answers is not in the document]
webbnh: I believe this comment refers to line 922, now, which is the first statement in the `for` loop body.

agent/bench-scripts/tests/pbench-fio/test-49.sts
```diff
@@ -1 +0,0 @@
-1
```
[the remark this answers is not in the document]
dbutenhof: I doubt it; but there are 13 soft links in this directory, 9 of them linking various test `.sts` files to this same `test-14.sts`, not just the three here. I'm going to close my eyes, make funny noises, and pretend I didn't "hear" you point this out because I'm not inclined to rework this apparent legacy "tradition".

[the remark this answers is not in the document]
webbnh: _Le sigh._

lib/pbench/test/unit/agent/test_utils.py
```diff
@@ -216,12 +218,42 @@ class TestLocalRemoteHost:
...
         lrh = LocalRemoteHost()
-        assert not lrh.is_local(
-            "testhost.example.com"
-        ), "'testhost.example.com' (which is not a real name) should still be remote"
+
+        # An unresolvable host isn't usable for benchmarking; we expect name
+        # resolution to fail, and the exception should propagate to the caller.
+        with pytest.raises(socket.gaierror):
+            lrh.is_local("testhost.example.com")
         assert not lrh.is_local("example.com"), "'example.com' should be remote"
```
webbnh, line 252: Why does `testhost.example.com` cause an exception to be raised...or, rather, why _doesn't_ the same happen for `example.com`??

webbnh (approved):
This PR is nearly done from my perspective...it's just down to the question of the missing `else`, some old nits (a spot which could use DRYing, some possibly missing kill messages, and a trivial code-positioning thing) all of which could be left alone, a new nit (an include ordering problem), and a small question.

So, if we can get @portante's feedback, we can be good to go.

lib/pbench/test/unit/agent/test_utils.py
```diff
@@ -1,12 +1,14 @@
...
 """
-
 import os
 import pathlib
+import socket
 import pytest
 import signal
 import time
 
+import ifaddr
```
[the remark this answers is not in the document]
dbutenhof: Yes; I *knew* you'd notice that, but *I* didn't notice until after I'd pushed the commit. I have no idea why `vscode` can't get this right, but it doesn't, and sometimes I forget to check on it. (Le sigh.)

lib/pbench/test/unit/agent/test_utils.py
dbutenhof: Because `example.com` really exists, but `test.example.com` doesn't. I stopped handling the error internally because I really don't think we want to. We're going to fail at some point if they specify a bad remote; why fudge here?

```
dbutenho 16:26 pbench-fio:is-remote ping example.com
PING example.com (93.184.216.34) 56(84) bytes of data.
64 bytes from 93.184.216.34 (93.184.216.34): icmp_seq=1 ttl=57 time=28.4 ms

dbutenho 17:18 pbench-fio:is-remote ping test.example.com
ping: test.example.com: Name or service not known
```

lib/pbench/test/unit/agent/test_utils.py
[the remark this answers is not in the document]
webbnh: I figured I would point it out, just in case you ended up going back in to fix something else (which I'm hoping will happen once Peter looks at the missing `else`).

lib/pbench/test/unit/agent/test_utils.py
webbnh: I concur with and support your choice.  I just didn't realize that `example.com` actually existed.

agent/bench-scripts/pbench-fio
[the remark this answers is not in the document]
portante: Yes, this is correct.

The semantics of the `pbench-fio` bench-scripts `--clients` and `--client-file` are defined by fio itself.  By requesting fio separate the client and server portion of fio, we really trying to determine if a `fio --server` command was run remotely or locally.  The `do_local_client` flag is set, it means we have a local `fio --server` process running somewhere and we don't need to use `ssh` to do anything about it.

A possible fix would be to hoist the code at line 658 up to replace line 607.  But I would not advocate we do that in this commit.

See [Section 1.15 "Client/Server"](https://fio.readthedocs.io/en/latest/fio_doc.html#client-server) of the [fio](https://fio.readthedocs.io/en/latest/fio_doc.html) documentation for a more thorough review of the behavior.

agent/bench-scripts/pbench-fio
```diff
@@ -640,6 +655,11 @@ function fio_run_job() {
 	chmod +x $benchmark_results_dir/fio.cmd
 	debug_log "$benchmark: Going to run [$benchmark_bin $bench_opts $client_opts]"
 	pushd $benchmark_results_dir >/dev/null
+	if [[ ${do_local_client} -eq 1 ]]; then
+		debug_log "killing any old local fio processes and starting new a fio process for the local client"
+		killall ${benchmark} > /dev/null 2>&1
+		screen -dmS fio-server bash -c '${benchmark_bin} --server=,${client_ports[${client}]} > client-result.txt 2> client-result.err'
```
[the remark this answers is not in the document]
portante: As mentioned in https://github.com/distributed-system-analysis/pbench/pull/2542/files#r808592022, this code could be moved to replace line 605.

agent/bench-scripts/tests/pbench-fio/test-61.txt
```diff
@@ -0,0 +1,256 @@
...
+[debug][1900-01-01T00:00:00.000000] killing any old local fio processes and starting new a fio process for the local client
+[warn][1900-01-01T00:00:00.000000] log_hist_msec specified in job file but failed to find any log files (.../clients/foo/fio_clat_hist.*.log*) to process for histograms
+[warn][1900-01-01T00:00:00.000000] log_hist_msec specified in job file but failed to find any log files (.../clients/bar/fio_clat_hist.*.log*) to process for histograms
+[warn][1900-01-01T00:00:00.000000] log_hist_msec specified in job file but failed to find any log files (.../clients/testhost/fio_clat_hist.*.log*) to process for histograms
+[warn][1900-01-01T00:00:00.000000] log_hist_msec specified in job file but failed to find any log files (.../clients/*/fio_clat_hist.*.log*) to process for histograms
+[debug][1900-01-01T00:00:00.000000] post-processing tool data
+[debug][1900-01-01T00:00:00.000000] post-processing fio result
+[debug][1900-01-01T00:00:00.000000] killing any old fio processes, creating directories, and starting new a fio process on the remote clients
```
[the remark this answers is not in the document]
portante: Let's address in a separate PR.  That is an existing behavior where we only try to clean up remote processes from before when we start new ones.

lib/pbench/cli/agent/commands/is_local.py
```diff
@@ -0,0 +1,29 @@
...
+@click.argument("host")
+def main(host):
+    """
+    Determine whether host is a representation of `localhost`.
+
+    If this logic cannot absolutely identify the host as an alias for the local
+    system, then it will report "remote".
+
+    Returns 0 for "local", 1 for "remote", or 2 for a parameter error.
+    \f
```
portante, line 16: What does `\f` do?

portante (commented):
Technically I cannot "Approve" or "Request changes" on my PR.  But this looks good -- somebody else must approve.

agent/bench-scripts/pbench-fio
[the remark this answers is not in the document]
dbutenhof: > Yes, this is correct.

Hmm. I find that a bit ambiguous. 👀 That is, "the original code is correct" or "Webb's concern is correct"??

However, given that without `${clients}` we're presumably running "monolithic" rather than client/server, and the cleanup action (local or remote) involves restarting `fio-server`, I'm assuming you meant (1) ... the code is correct as is, and the cleanup isn't necessary or meaningful without client configuration.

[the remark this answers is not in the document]
webbnh: > I find that a bit ambiguous.

Me, too!  😆  (I so hate pronouns.)

If the code is correct, perhaps we should have a comment explaining the logic and ramifications (i.e., that, if there are no "clients", then we running "monolithic", and there is nothing else to do)?

agent/bench-scripts/tests/pbench-fio/test-61.txt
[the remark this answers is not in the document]
webbnh: Tracked in #2629.

webbnh:
> Technically I cannot "Approve" or "Request changes" on my PR. But this looks good -- somebody else must approve.

Less "technically" and more like "actually" -- GitHub grays-out those options for the author...how...intriguing.

agent/bench-scripts/pbench-fio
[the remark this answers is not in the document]
webbnh: Dave and I discussed this and we determined that there are some fairly fundamental problems with this script which pre-date this PR and that these changes do not make things materially worse.  Dave is going to open one or more issues describing what we found, and we can take up the discussion in the context of those issues or the PR(s) that they result in.

webbnh (changes requested):
So, Dave and I found a bug.

webbnh (approved):
👍

portante (commented):
FWIW, this looks good still.

webbnh (approved):
GitHub says there's no change since my last review, so it must still be good!

dbutenhof:
> GitHub says there's no change since my last review, so it must still be good!

Yeah, just needed to be dusted off (rebased); of course now it's out of date again!

webbnh:
Given that we can do a Squash-and-merge after we get everything approved, I'm no longer inclined to see the author do a rebase if it's "just to bring the branch up to date with main".

Granted, there are exceptions to that:  if the reason for rebasing is another commit to a dependent or related area, we should have the author rebase so that we get the two new changes tested together before they are merged; and, obviously, if there are merge conflicts, those are probably best resolved by the author.  But, if the reason for the rebase is otherwise unrelated changes, particularly if the intention of the PR is to merge a single commit, then I think it produces less overhead if we do the Squash-and-merge at the end, rather than having the author handle it and undergoing an additional round of approvals.  (Likewise, if the contents of the PR are to be merged as a series of commits, then the author needs to do the squashes, but if the branch is already properly arranged, we can do a Rebase-and-merge without requiring an additional round of approvals.)

dbutenhof:
> Given that we can do a Squash-and-merge after we get everything approved, I'm no longer inclined to see the author do a rebase if it's "just to bring the branch up to date with main".

You're not wrong; however,

1. Relying on GitHub to do the rebase always makes me nervous even when it seems there "can't" be a conflict. 🍤 
2. I rebased before reading this reply;
3. Since I rebased on top of my `jenkins/run` change, this is the first test of the merged code (even though it's already run in my PR branch, and there should be no difference...)

webbnh (approved):
It _still_ looks good to me.  🙃

npalaska (approved):

Issue #1921, opened 2020-10-28
Agent `v0.71.0` Release Notes


portante:
Intended as a parking area for release notes.

- [ ] ~~Layout the tech-preview for the new "pbench" command~~
- [x] Layout the tech-preview for the new HTTP PUT method of posting tar balls
- [x] Latent support for the new pbench-server PUT API
- [ ] ~~Tool Meister now collects system information in addition to tool data~~
- [ ] ~~Addition of the `pbench-move-results --controller=<name>` option~~
- [x] Removal of deprecated `pbench-fio --remote-only` (#2543)
- [x] Change to check command versions instead of RPM version for `pbench-fio` and `pbench-uperf` (#2528)
- [x] Require the `--user` option for `pbench-move-results` (PR #2544)

portante:
The Alpha release notes will be created via PR #2792.

portante:
All of the items listed in the description have been added to PR #2792.

portante:
Release notes finalized in PR #2792.

PR #2543, opened 2021-11-16
Remove deprecated `--remote-only` pbench-fio option

portante:


dbutenhof (approved):
Interesting. So this never appeared in any legacy test gold output?

webbnh (approved):
This change looks fine.

Do we have any policy about how long a feature needs to remain present after being deprecated (or any other boilerplate about announcing deprecation and removal)?  (E.g., should this PR include any documentation changes...either to remove this from documentation or to add a release note for its departure?)

portante:
> Do we have any policy about how long a feature needs to remain present after being deprecated (or any other boilerplate about announcing deprecation and removal)?

I don't think we have a well established policy, other than we typically give folks at least one release before its removal.  In this case, it was marked this way years ago.

> (E.g., should this PR include any documentation changes...either to remove this from documentation or to add a release note for its departure?)

I don't see it referenced in our existing documentation (either in branch `main` or `gh-pages`), and I have added it to the release notes issue (#1921) for the `v0.71` release target.

PR #2544, opened 2021-11-16
Require `--user` for `pbench-move/copy-results`

portante:
Ahead of the work coming for the pbench server being endowed with the notion of a user, we require results being moved to provide the `--user` option for `pbench-move-results`.

### DIFF
--- a/agent/util-scripts/pbench-move-results
+++ b/agent/util-scripts/pbench-move-results
@@ -8,8 +8,24 @@ pbench_bin="$(realpath -e ${script_path}/..)"
 . "${pbench_bin}"/base
 
 function usage() {
-    printf "usage:\n"
-    printf "${script_name} [--help] [--controller=<controller>] [--user=<user>] [--prefix=<path>] [--xz-single-threaded] [--show-server]\n"
+    printf -- "usage:\n%s [--help] --user=<user> [--controller=<controller>]" "${script_name}"
+    printf -- " [--prefix=<path>] [--xz-single-threaded] [--show-server]\n"
+    printf -- "\nThe '--user' option value is required, and can also be provided\n"
+    printf -- "by the 'PBENCH_USER' environment variable.\n"
+    printf -- "\nThe '--controller' option is also required, but defaults to\n"
+    printf -- "the 'hostname -f' value, or the 'PBENCH_CONTROLLER' environment\n"
+    printf -- "variable.\n"
+    printf -- "\nThe '--prefix' option allows the user to specify an optional\n"
+    printf -- "directory path hierarchy to be used when displaying the result\n"
+    printf -- "tar balls on the pbench server.\n"
+    local op="move"
+    if [[ "${script_name}" == "pbench-copy-results" ]]; then
+	op="copy"
+    fi
+    printf -- "\nThe '--show-server' will not ${op} any results, but resolve and\n"
+    printf -- "then display the pbench server destination for results.\n"
+    printf -- "\nThe '--xz-single-threaded' will force the use of a single\n"
+    printf -- "thread for locally compressing the result tar balls.\n"
 }
 
 function missing_arg() {
@@ -83,11 +99,16 @@ done
 if [[ -z "${controller}" ]]; then
     controller=${_pbench_full_hostname}
     if [[ -z "${controller}" ]]; then
-        error_log "Missing controller name (should be \"hostname -f\" value)"
+        error_log "Missing required controller name (should be \"hostname -f\" value)"
         exit 1
     fi
 fi
 validate_hostname "${controller}" "controller"
+
+if [[ -z "${user}" ]]; then
+    error_log "Missing required '--user' option (not found via 'PBENCH_USER' environment variable, either)"
+    exit 1
+fi
 
 # Move into pbench run collection directory
 cd ${pbench_run} >/dev/null
@@ -135,7 +156,7 @@ if [[ -z "${rel}" ]]; then
     rel="unknown"
 fi
 # User-Agent HTTP header: <pbench-agent-ver-rel>:<FQDN>:<$USER>:<full path of this script>""
-user_agent="pbench-agent-${ver}-${rel}:${_pbench_full_hostname}:${USER}:${script_name}"
+user_agent="pbench-agent-${ver}-${rel}:${_pbench_full_hostname}:${USER}:${user}:${script_name}"
 
 results_host_info_url=$(pbench-config host_info_url results)
 results_host_info=$(curl -s -A "${user_agent}" -L "${results_host_info_url}")
@@ -216,11 +237,7 @@ while read -r dir; do
     # contain the tarball and the md5 file (as ${tb}.tar.xz.md5.check); pass
     # that information on to pbench-make-result-tb.
     result_dir=$(basename ${dir})
-    if [[ -n "${user}" ]]; then
-        user_arg="--user ${user}"
-    else
-        user_arg=""
-    fi
+    user_arg="--user ${user}"
     if [[ -n "${prefix}" ]]; then
         prefix_arg="--prefix ${prefix}"
     else
@@ -247,7 +264,7 @@ while read -r dir; do
         continue
     fi
 
-    if [[ "$script_name" == "pbench-move-results" ]]; then
+    if [[ "${script_name}" == "pbench-move-results" ]]; then
         rm -r ${result_dir}
         if [[ ${?} -ne 0 ]]; then
             error_log "UNEXPECTED ERROR: rm failed to remove the ${result_dir} directory hierarchy"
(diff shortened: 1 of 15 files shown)


portante:
I realize this might be a bit controversial, but wanted to get this publicly available for us to consider.

pravins:
+1 for this change.

agent/util-scripts/pbench-move-results
```diff
@@ -83,12 +99,17 @@ done
...
     if [[ -z "${controller}" ]]; then
-        error_log "Missing controller name (should be \"hostname -f\" value)"
+        error_log "Missing required controller name (should be \"hostname -f\" value)"
         exit 1
     fi
 fi
 validate_hostname "${controller}" "controller"
 
+if [[ -z "${user}" ]]; then
+    error_log "Missing required '--user' option (not found via 'PBENCH_USER' environment variable, either)"
```
dbutenhof, line 109: Maybe `Missing required 'user' value (provided either by '--user' option or 'PBENCH_USER' environment variable)`??

dbutenhof (dismissed):
This looks good on its own. It raises a few strategic questions about 0.72, however, with the concept of user (and prefix) as *result path* structuring rather than simply "--user as a hack placeholder for dataset owner".

`pbench-results-move` doesn't support either `--prefix` or `--user`. Should it? I've been working on my class structure to comprehend and mutate the on-disk structure without considering these (and therefore without needing to read the `metadata.log` in order to discover the `RESULTS` tree structure ... and completely ignoring a `USERS` tree as it doesn't exist).

Right now, the classes aren't doing unpack (because I'm not touching the current unpack code); but if we want this sort of access path customization, we need to add them into `pbench-results-move` and I'll need to discover the prefix and user paths in order to properly delete them.

agent/util-scripts/pbench-move-results
```diff
@@ -8,8 +8,24 @@ pbench_bin="$(realpath -e ${script_path}/..)"
 . "${pbench_bin}"/base
 
 function usage() {
-    printf "usage:\n"
-    printf "${script_name} [--help] [--controller=<controller>] [--user=<user>] [--prefix=<path>] [--xz-single-threaded] [--show-server]\n"
+    printf -- "usage:\n%s [--help] --user=<user> [--controller=<controller>]" "${script_name}"
+    printf -- " [--prefix=<path>] [--xz-single-threaded] [--show-server]\n"
+    printf -- "\nThe '--user' option value is required, and can also be provided\n"
+    printf -- "by the 'PBENCH_USER' environment variable.\n"
```
webbnh, line 14: "The '--user' option value is required if not provided by the 'PBENCH_USER' environment variable; otherwise, the value provided on the command line will override any value provided by the environment."

agent/util-scripts/pbench-move-results
```diff
@@ -8,8 +8,24 @@ pbench_bin="$(realpath -e ${script_path}/..)"
...
+    printf -- "\nThe '--controller' option is also required, but defaults to\n"
+    printf -- "the 'hostname -f' value, or the 'PBENCH_CONTROLLER' environment\n"
+    printf -- "variable.\n"
```
webbnh, line 17: You cannot have a switch which is required but which provides a default!  😀

"The '--controller' option may be used to override the value provided by the 'PBENCH_CONTROLLER' environment variable; if neither value is available, the result of 'hostname -f' is used.  (If no value is available, the command will exit with an error.)"

agent/util-scripts/pbench-move-results
```diff
@@ -8,8 +8,24 @@ pbench_bin="$(realpath -e ${script_path}/..)"
...
+    printf -- "\nThe '--prefix' option allows the user to specify an optional\n"
+    printf -- "directory path hierarchy to be used when displaying the result\n"
+    printf -- "tar balls on the pbench server.\n"
+    local op="move"
+    if [[ "${script_name}" == "pbench-copy-results" ]]; then
+	op="copy"
+    fi
+    printf -- "\nThe '--show-server' will not ${op} any results, but resolve and\n"
+    printf -- "then display the pbench server destination for results.\n"
```
webbnh, line 26: `op` would be potentially useful in two other places as well (lines 267 and 285), so I suggest doing this calculation near line 4 and putting the result in a global variable for use in all three places.

Also, I suggest doing it with parameter expansion:
```
op=${script_name#pbench-}
op=${op%-results}
```
(If you're going to stick with exact matches, I would recommend being consistent about the RHS (e.g., compare against `"pbench-move-results"` in all cases), and possibly add some code to verify that `$script_name` has only expected values.)

agent/util-scripts/pbench-move-results
```diff
@@ -83,12 +99,17 @@ done
 if [[ -z "${controller}" ]]; then
     controller=${_pbench_full_hostname}
     if [[ -z "${controller}" ]]; then
-        error_log "Missing controller name (should be \"hostname -f\" value)"
+        error_log "Missing required controller name (should be \"hostname -f\" value)"
```
webbnh, line 102: Actually, there is no reason why the controller name "should" be the hostname....  Moreover, what we're trying to report here is that we have _no_ value for `controller` because `hostname -f` produced a blank, so calling for it to be set to that value isn't helpful, either.

How about something like "no hostname configured, please specify --controller" or "--controller is required, because no hostname is configured".

agent/util-scripts/pbench-move-results
portante: Fixed.

agent/util-scripts/pbench-move-results
```diff
@@ -135,7 +156,7 @@ if [[ -z "${rel}" ]]; then
     rel="unknown"
 fi
 # User-Agent HTTP header: <pbench-agent-ver-rel>:<FQDN>:<$USER>:<full path of this script>""
-user_agent="pbench-agent-${ver}-${rel}:${_pbench_full_hostname}:${USER}:${script_name}"
+user_agent="pbench-agent-${ver}-${rel}:${_pbench_full_hostname}:${USER}:${user}:${script_name}"
```
webbnh, line 159: What are we doing, here?  (An explanatory code comment would be useful....)  In particular, why is it important to include _both_ `$USER` and `$user`, without including `$controller` or `$prefix`?

(This change resulted in lots of churn in the gold files...is it really worth it?)

agent/util-scripts/pbench-move-results
```diff
@@ -216,11 +237,7 @@ while read -r dir; do
     # contain the tarball and the md5 file (as ${tb}.tar.xz.md5.check); pass
     # that information on to pbench-make-result-tb.
     result_dir=$(basename ${dir})
-    if [[ -n "${user}" ]]; then
-        user_arg="--user ${user}"
-    else
-        user_arg=""
-    fi
+    user_arg="--user ${user}"
```
webbnh, line 240: Given that `--user` is now required, I would in-line this expression and dispense with the variable.

agent/util-scripts/pbench-move-results
```diff
@@ -247,7 +264,7 @@ while read -r dir; do
         continue
     fi
 
-    if [[ "$script_name" == "pbench-move-results" ]]; then
+    if [[ "${script_name}" == "pbench-move-results" ]]; then
```
webbnh, line 267: Why make this change without also changing line 285?  (This looks like churn.)

agent/util-scripts/pbench-move-results
```diff
@@ -83,12 +99,17 @@ done
...
+        error_log "Missing required controller name (should be \"hostname -f\" value)"
         exit 1
     fi
 fi
 validate_hostname "${controller}" "controller"
 
+if [[ -z "${user}" ]]; then
+    error_log "Missing required 'user' value (provided either by the '--user' option or the 'PBENCH_USER' environment variable)"
+    exit 1
+fi
```
webbnh, line 111: How about, "No 'PBENCH_USER' environment variable found; please specify '--user''.

webbnh (changes requested):
Looks generally good, but could use some polish.

dbutenhof (approved):

agent/util-scripts/pbench-move-results
portante: Does @dbutenhof's suggestion work?

dbutenhof (approved):

agent/util-scripts/pbench-move-results
portante: Change was originally made to bring the variable name reference in line with others.  But going forward based on other comments, this will be removed entirely.

agent/util-scripts/pbench-move-results
webbnh: Yes...it's just long and less "actionable".

webbnh (approved):
👍

dbutenhof (approved):
🎆

PR #2545, opened 2021-11-16
Post deprecation notice for unsupported benchmark scripts

portante:
There are six benchmark scripts languishing in the pbench-agent with little or no support for them:

 1. `pbench-cyclictest`
 2. `pbench-dbench`
 3. `pbench-iozone`
 4. `pbench-migrate`
 5. `pbench-netperf`
 6. `pbench-run-benchmark`

The goal is to have these deprecated in the `v0.71` release so that we can remove them entirely in the `v0.72` release.

### DIFF
--- a/agent/bench-scripts/pbench-run-benchmark.pl
+++ b/agent/bench-scripts/pbench-run-benchmark.pl
@@ -59,7 +59,7 @@ if (! exists($benchmarks_hash{$benchmark})) {
     exit 1;
 }
 
-print "WARNING: pbench-run-benchmark is deprecated and will be removed in the next release.\n";
+print "WARNING: pbench-run-benchmark is deprecated and will be removed in the next release\n";
 
 # The rest of the parameters are --arg=val, most of which we just pass to other scripts,
 my %params = get_params(@ARGV);
--- a/agent/bench-scripts/pbench-run-benchmark.pl
+++ b/agent/bench-scripts/pbench-run-benchmark.pl
@@ -59,6 +59,8 @@ if (! exists($benchmarks_hash{$benchmark})) {
     exit 1;
 }
 
+print "WARNING: pbench-run-benchmark is deprecated and will be removed in the next release.\n";
+
 # The rest of the parameters are --arg=val, most of which we just pass to other scripts,
 my %params = get_params(@ARGV);
 
--- a/agent/bench-scripts/pbench-cyclictest
+++ b/agent/bench-scripts/pbench-cyclictest
@@ -114,6 +114,8 @@ while true; do
 done
 verify_common_bench_script_options $tool_group $sysinfo
 
+warn_log "${script_name} is deprecated and will be removed in the next release"
+
 ## Ensure the right version of the benchmark is installed
 check_install_rpm "${benchmark_rpm}" "${ver}"
 
--- a/agent/bench-scripts/pbench-cyclictest
+++ b/agent/bench-scripts/pbench-cyclictest
@@ -12,6 +12,8 @@ pbench_bin="`cd ${script_path}/..; /bin/pwd`"
 # source the base script
 . "$pbench_bin"/base
 
+warn_log "${script_name} is deprecated and will be removed in the next release"
+
 export benchmark="cyclictest"
 benchmark_rpm=rt-tests
 ver=0.87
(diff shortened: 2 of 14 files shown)


dbutenhof (approved):
These are all visible only when the tool is used, interactively. I guess this change will also appear in the change log of release notes, which is probably somewhat (in theory) more visible. A general documentation update to list deprecated features would be good, too; although I'm not sure how likely that is to be seen. 😄  Maybe `docs/guides/UserGuide.rst`?? Or ... ???

agent/bench-scripts/pbench-run-benchmark.pl
```diff
@@ -59,6 +59,8 @@
     exit 1;
 }
 
+print "WARNING: pbench-run-benchmark is deprecated and will be removed in the next release.\n";
```
webbnh, line 62: You don't want to offer guidance on a replacement path?

webbnh (approved):
Looks OK to me.

webbnh:
What about references in `agent/bench-scripts/tests/test-benchmark-clis/test-CL.txt` and `docs/guides/UserGuide.rst`?

agent/bench-scripts/pbench-run-benchmark.pl
portante: I don't know what guidance to offer at this point.  Ideally we'd have a path via https://github.com/cloud-bulldozer/benchmark-wrapper, but we are not there yet.

For this PR I think it is enough to make changes to issue the `WARNING`, and then work via the release notes and other documentation later when a path becomes available.

portante:
> What about references in `agent/bench-scripts/tests/test-benchmark-clis/test-CL.txt`

I think they are okay for now because we test all the help messages.  But the `--help` messages should be updated as well.

> and `docs/guides/UserGuide.rst`?

I'd like to approach the documentation change as a separate PR since I'll also have to update the `gh-pages` branch as well.

portante (commented):
Looks ready for review.  I'll make the documentation changes in a separate PR.

portante:
Filed PR #2550 for the documentation changes.

dbutenhof (approved):
I vaguely question *removing* them from `UserGuide.rst` when we're still supplying them (and they ought to "work" as they did before aside from the new warning). The counterweight is that by removing them we won't tempt anyone else...

npalaska (approved):

agent/bench-scripts/pbench-run-benchmark.pl
webbnh: Removing this doesn't reduce our capability for the user, right?  All they have to do is run their choice of benchmark directly, right?  So, a little suggestion here would go a long way.

portante: Well, sort of, but not really.  We have `pbench-user-benchmark`, but `pbench-run-benchmark` is a "generalized" `pbench-fio` combined with `pbench-uperf`.  It has the ability to run `fio`, `uperf`, `trafficgen`, etc. but from one script.  We really don't have an alternative for that, other than the existing scripts.

webbnh: Right, but if someone was using `pbench-run-benchmark` to run fio, they can recover from the deprecation by running `pbench-fio`, right?  (And, yeah, for the "gap" cases, they can use `pbench-user-benchmark`.)  So, it would be helpful for the message to point that out.

(If you really think that no one is using `pbench-run-benchmark`, then obviously there's little point to enhancing this, but if someone uses this to run a benchmark that we have support for in another script, it seems like we could point them to that script, and otherwise we could suggest `pbench-user-benchmark`.)

portante: And the `0.69` release doesn't even work.  So maybe we should just get rid of it entirely:

```
# pbench-run-benchmark
Empty compile time value given to use lib at /opt/pbench-agent/bench-scripts/pbench-run-benchmark line 19.
Use of uninitialized value in require at /opt/pbench-agent/bench-scripts/pbench-run-benchmark line 23.
Can't locate PbenchCDM.pm in @INC (@INC contains:  /usr/local/lib64/perl5 /usr/local/share/perl5 /usr/lib64/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib64/perl5 /usr/share/perl5 .) at /opt/pbench-agent/bench-scripts/pbench-run-benchmark line 23.
BEGIN failed--compilation aborted at /opt/pbench-agent/bench-scripts/pbench-run-benchmark line 23.
```

portante: Let's just merge this commit as is, and I'll take that removal up in another commit.

webbnh: 👍

PR #2548, opened 2021-11-16
pbench-register-tool: treat an empty remotes file as an error

### LINKED ISSUE #2538: Should we check that the remotes file in `pbench-register-tool[-set]` is non-empty?
The implementation does not currently raise an error if in specifying the  `--remotes=@foo` option, the file `foo` is empty. The `{pbench_run}/tools-v1-<toolset>` directory is created if necessary, but no tools are registered and the exit status is 0 (from either registration program).

Arguably, however, that is a potential error that the user might want to know about.

ndokos:
Fixes #2538

An empty remotes file specified with `--remotes=@/path/to/file` now
makes `pbench-register-tool` (and also `pbench-register-tool-set`) log
an error message and exit with an error status.

### DIFF
--- a/agent/util-scripts/pbench-register-tool
+++ b/agent/util-scripts/pbench-register-tool
@@ -250,6 +250,12 @@ if [[ ${remotes_arg::1} == "@" ]]; then
 		exit 1
 	fi
 
+	if [[ ! -s ${remotes_file} ]]; then
+		error_log "--remotes=@${remotes_file} specifies an empty file"
+		usage >&2
+		exit 1
+	fi
+
 	declare -a remotes_A
 	declare -A labels_A
 	idx=0
--- a/agent/util-scripts/unittests
+++ b/agent/util-scripts/unittests
@@ -371,6 +371,7 @@ declare -A tools=(
     [test-11.14]="pbench-register-tool-set"
     [test-11.15]="pbench-register-tool-set"
     [test-11.16]="pbench-register-tool-set"
+    [test-11.17]="pbench-register-tool-set"
     [test-17]="test-tm-start-bad-group"
     [test-18]="test-tm-stop-bad-group"
     [test-19]="test-tool-trigger"
@@ -454,6 +455,7 @@ declare -A options=(
     [test-11.13]="--help"
     [test-11.15]="--remotes=@non-existent-file"
     [test-11.16]="--remotes=@${_testdir}/tmp/good-remote-file"
+    [test-11.17]="--remotes=@${_testdir}/tmp/empty-remote-file"
 
     # pbench-move-results
     [test-20]="--help"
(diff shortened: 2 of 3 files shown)


agent/util-scripts/unittests
```diff
@@ -531,6 +534,7 @@ declare -A pre_hooks=(
     [test-06]='ln -s mock-pbench-tool-meister-client ${_testopt}/unittest-scripts/pbench-tool-meister-client; mkdir -p ${_testdir}/42-iter/sample42/tools-default/testhost.example.com/iostat; touch ${_testdir}/42-iter/sample42/tools-default/testhost.example.com/iostat/iostat-stdout.txt'
     [test-07]='mkdir -p ${_testdir}/42-iter/sample42/tools-foobar/testhost.example.com/iostat; touch ${_testdir}/42-iter/sample42/tools-foobar/testhost.example.com/iostat/iostat-stdout.txt'
     [test-11.16]='mkdir -p ${_testdir}/tmp; (echo foo; echo bar) > ${_testdir}/tmp/good-remote-file'
+    [test-11.17]='mkdir -p ${_testdir}/tmp; touch  ${_testdir}/tmp/empty-remote-file'
```
webbnh, line 537: Shouldn't there be a "post hook" which removes this file again?  (Aside from being tidy, this prevents it from becoming non-empty in the future....)

webbnh (approved):
Looks good...but shouldn't the test clean up?

agent/util-scripts/unittests
ndokos: `_reset_state` deletes the whole `_testdir`  after each test (which is why the file was created in `${_testdir}/tmp`). The post hooks are meant for things like sorting output to make it reproducible.

agent/util-scripts/pbench-register-tool
```diff
@@ -250,6 +250,12 @@ if [[ ${remotes_arg::1} == "@" ]]; then
 		exit 1
 	fi
 
+	if [[ ! -s ${remotes_file} ]]; then
+		error_log "--remotes=@${remotes_file} specifies an empty file"
+		usage >&2
```
portante, line 255: Seems like issuing a usage message here will cloud the error.

portante (approved):
Ship it!

agent/util-scripts/unittests
webbnh: Cool.  (A recent PR from @portante, included a post-hook clean-up, so I was keying off that.)

agent/util-scripts/pbench-register-tool
ndokos: I followed the current "standard practice": both `pbench-register-tool` and `pbench-register-tool-set` print out their usage message on any error (although when `pbench-register-tool` is used in checking mode by `pbench-register-tool-set`, its usage message is suppressed: only the `pbench-register-tool-set`  usage is printed out).

Opened a discussion about this: #2552

dbutenhof (approved):

PR #2550, opened 2021-11-17
Remove deprecated benchmark scripts

portante:
Documentation changes for PR #2545.

### DIFF
--- a/doc.html
+++ b/doc.html
@@ -181,12 +181,14 @@
                                 separately as well with a benchmark
                                 that is not built-in to Pbench, or a Pbench script can be written for
                                 the benchmark. Such contributions are
-                                more than welcome!</p>
+                                more than welcome!
+                            </p>
                         </div>
                         <div id="how" class="section_cardbody">
                             <p class="cardHeader">TL;DR - How to set up Pbench and run a benchmark </p>
-                            <p class="info_text" style="text-align:justify">Prerequisite: Somebody has
-                                already done the server setup.</p>
+                            <p class="info_text" style="text-align:justify">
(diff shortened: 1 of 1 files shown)


doc.html
```diff
@@ -1,1005 +1,1003 @@
 <!DOCTYPE html>
 <html>
+    <head>
+        <meta charset="utf-8">
+        <title></title>
+        <meta name="author" content="">
+        <meta name="description" content="">
+        <meta name="viewport" content="width=device-width, initial-scale=1">
+        <link href="css/style.css" rel="stylesheet">
+        <link href="css/fontawesome.min.css">
```
dbutenhof, line 10: So, not *quite* just massive whitespace changes in the first commit. 😆

dbutenhof (approved):
Do we want to remove all mention entirely, or would it maybe be better to add them in a *list* of deprecated scripts that will be removed. It wouldn't need any additional details, since obviously we don't want to encourage anyone to use them ... but it might be more noticeable, to anyone who bothers to skim the document, that "something is happening", than simply having them drop out.

doc.html
portante: Refactored into a separate commit, thanks for pointing out!

portante:
> Do we want to remove all mention entirely ...

I don't think it is worth keeping mention of the deprecated scripts in there because there is 0 evidence of use for years, the `pbench-tpcc` does not actually exist, and we are missing `pbench-cyclictest`, `pbench-iozone`, and `pbench-netperf`.

Best to let sleeping dogs lie.

PR #2551, opened 2021-11-17
Revamp how we use tox to reduce envs

portante:
In summary, by default tox runs all the tests, while accepting arguments to request various sub-sets to be run.  See the `README.md` file for a description of the interface.

We also simplify the `jenkins/run` script to remove the `CONTAINER_HOME` options and just build out `/home/developer` inside the container. There is no need for an external mounted `/home`.

The default jenkins job will use the default installed python, RPM provided python modules, and pip install the rest as normal.  By default now Fedora 34 using Python 3.9.


agent/tox-unittests
```diff
@@ -4,6 +4,6 @@ prog="$(basename "${0}")"
 progdir="$(realpath -e $(dirname "${0}"))"
 
 # List of tox environments to run.
-tox_envs="py3-server server"
+tox_envs="agent"
```
dbutenhof, line 7: This is a side comment on GitHub: it decided that you *moved* `server/run-unittests` to `agent/tox-unittests` despite the fact that almost half the lines differ substantially. That's ... intriguing.

tox.ini
```diff
@@ -36,75 +30,23 @@ whitelist_externals =
     env
 
 [testenv:server]
-description = Runs all non-Python3-based server unit/functional tests
+basepython = python3.8
```
dbutenhof, line 33: I wonder what this will do on RHEL 8.4, with python 3.6.8?

dbutenhof (commented):
Interesting. Since it's still draft, I do more than a few comments.

In general, my biggest wish for our test environment (which, some day, I might get around to working on) would be to break up the pytest stage so running "server" pytests doesn't also run the agent "functional" tests.

tox.ini
portante: Oh, I thought RHEL 8 had python 3.8 as it's base.  Let's use 3.6 then.

portante:
> In general, my biggest wish for our test environment (which, some day, I might get around to working on) would be to break up the pytest stage so running "server" pytests doesn't also run the agent "functional" tests.

Tracking in issue #2553.

dbutenhof (dismissed):

tox.ini
```diff
@@ -36,75 +31,21 @@ whitelist_externals =
...
 [testenv:server]
-description = Runs all non-Python3-based server unit/functional tests
+description = Runs all server unit/functional tests
 passenv = PBENCH_UNITTEST_SERVER_MODE
 deps =
     -r{toxinidir}/server/requirements.txt
     -r{toxinidir}/server/test-requirements.txt
 commands =
-    bash -c "./server/bin/unittests {posargs}"
+    bash -c "./server/exec-unittests {envdir} {posargs}"
```
webbnh, line 40: You like invoking `exec-unittests` on a relative path instead of on `{toxinidir}/server/exec-unittests` (like line 38...or does tox not expand the references inside double quotes?)?

Ditto line 48.

webbnh (dismissed):
This generally looks fine, other than the problem with `jenkins/tox` in the `README.md`.  However, I'm unexcited by the little `exec-unittest` wrappers, but I'm hoping you'll educate me as to why they are necessary.

My foremost concern is that there doesn't seem to be any provision for making Mr. Jenkins use any of this, and I think it is a bad idea to have a testing regimen for developers which is separate form the one used by the CI.

ndokos (approved):

portante:
> This generally looks fine, other than the problem with `jenkins/tox` in the `README.md`.

I'll address that shortly.

> However, I'm unexcited by the little `exec-unittest` wrappers, but I'm hoping you'll educate me as to why they are necessary.

If you put the commands directly in `tox.ini`, then all the commands have to be run via `tox -e <envi>`.  By using these scripts, arguments can then be passed to those scripts to only run sub-sets of the tests.

For example, instead of running all the `util-scripts` and `tool-scripts` tests when making changes to `pbench-uperf`, for example, we can just run `tox -e agent -- bench-scripts` (we used to have that in a separate environment, `tox -e bench-scripts`).  Further, we can ask to run a specific test via `tox -e agent -- bench-scripts test-CL`.

> My foremost concern is that there doesn't seem to be any provision for making Mr. Jenkins use any of this, and I think it is a bad idea to have a testing regimen for developers which is separate form the one used by the CI.

Is this concern introduced via this commit?  Isn't it the case that we have this concern before this commit?  If this was a concern before, then this commit is not changing the status quo.  Let's address that concern in another commit.

tox.ini
portante: Did I make a change to invoke on a relative path?  It seems like that it how it was before.  I believe the commands are always run in the directory in which the `tox.ini` file is found.

agent/exec-unittests
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
+# Test execution help for the "agent" tox environment.  It is not intended
+# to be executed outside of a tox environment.
+
+_envdir="${1}"
+_subtst="${2}"
+shift; shift
```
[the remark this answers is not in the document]
portante: I tried it, but it doesn't work, only `shift; shift` works.

riya-17 (commented):
looks good

agent/exec-unittests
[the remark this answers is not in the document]
dbutenhof: Really? How odd...

```
dbutenho 07:55 ansible:filetree cat >test
echo ${@}
shift 2
echo ${@}
dbutenho 07:56 ansible:filetree chmod +x test
dbutenho 07:56 ansible:filetree ./test 1 2 3 4 5 6
1 2 3 4 5 6
3 4 5 6
```

tox.ini
webbnh: As one can see from the diff's above, your change did not introduce the relative path; however, as a reviewer, I find it jarring that we use a relative path at line 40 when we use what seems like an absolute path (or a potentially _different_ relative path) at line 38 (et al.).  I'm concerned that inconsistencies like these open cracks which can admit surprising results.  (I'm aware of the power of position-independent code, but I'm also aware of the destructiveness of misplaced assumptions.)

webbnh:
> > This generally looks fine, other than the problem with `jenkins/tox` in the `README.md`.
> 
> I'll address that shortly.
> 
> [...]
> 
> > My foremost concern is that there doesn't seem to be any provision for making Mr. Jenkins use any of this, and I think it is a bad idea to have a testing regimen for developers which is separate form the one used by the CI.
> 
> Is this concern introduced via this commit? Isn't it the case that we have this concern before this commit? If this was a concern before, then this commit is not changing the status quo. Let's address that concern in another commit.

When you and I previously considered how we run tests and whether we should use Tox, we decided that we should move away from Tox.  I made the requisite changes to the Jenkins processes, but, out of an abundance of caution, I deferred removing the Tox-related files at that time.  Now, instead of completing that effort, this PR starts to reverse it, by dusting off the project Tox support, but it doesn't carry that through to restoring its use by the CI.  I regard this as problematic:  while I don't disagree with returning to Tox, and while it's wonderful that the revamped Tox support here provides nice flexibility to us as developers, it shouldn't be creating a test system which is separate from and parallel to what we use in the CI -- that's only going to breed confusion and possibly code-rot.

And, along those lines, your change to the `README.md` file is an additional step in the wrong direction.  The `jenkins/run jenkins/tox` command-line _was_ the way that developers could run the tests in exactly the same way as the CI ran them (back when the CI was using Tox...).  Having this PR effectively reintroduce Tox to our development environment while at the same time removing access to it in our CI environment is unhelpful.


> >I'm unexcited by the little `exec-unittest` wrappers, but I'm hoping you'll educate me as to why they are necessary.

My apologies, I referenced the wrong file:  it is `agent/tox-unittests` and `server/tox-unittests` to which I'm objecting.  These seem like clutter, given that one can just invoke `./tox-unittests agent`, or I presume, `tox -e agent`.


FWIW, it looks like [Tox is happy to run PyTest](https://tox.wiki/en/3.24.4/example/pytest.html#pytest-and-tox), and [PyTest is happy to run individual tests, to run them with special configurations and on multiple CPUs, without external support](https://tox.wiki/en/3.24.4/example/pytest.html#using-multiple-cpus-for-test-runs), but I guess we need fancy support to facilitate running the legacy tests in portions and/or in parallel.

README.md
```diff
@@ -79,17 +79,7 @@ $ cd pbench
...
 ### Running the Unit Tests
 
-To simply run the unit tests quickly from within the checked out source tree,
-execute:
-
- * `jenkins/run jenkins/tox -r --current-env -e jenkins-pytests`
- * `jenkins/run jenkins/tox -r --current-env -e jenkins-unittests`
-
-The above commands run the tests in a Fedora-base container with all the
-proper packages installed.
```
webbnh, line 89: I think it's really important for developers to be able to run the tests exactly like the CI will run them, and I think it would be appropriate for the CI to run tests using the same mechanism that the developers will use.  So, I think simply deleting this text (and removing `jenkins/tox`) is not moving things in the right direction.

webbnh (changes requested):
I think that the CI should be invoking the tests in the same way that the developers do (and that it should be trivial for developers to be able to invoke the tests in the same way that the CI does), so having two separate approaches to this is inappropriate.  I think that closing that gap is within the purview of this PR.

agent/exec-unittests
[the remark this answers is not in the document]
portante: If only one argument is provided, `shift 2` doesn't work.

portante:
> it shouldn't be creating a test system which is separate from and parallel to what we use in the CI -- that's only going to breed confusion and possibly code-rot.

It's not doing that.  The state of the CI prior to this PR already affords two different ways tests are run.

Compare https://github.com/distributed-system-analysis/pbench/blob/e5c4dbde1ae55427f483105d6fc0d31b69f6ab3c/tox.ini and https://github.com/distributed-system-analysis/pbench/blob/29f2e04832e075c967d56bcf84cbe0e4322d923c/jenkins/Pipeline.gy.

I'd rather address the reunification of developer / CI test invocation in a separate PR.

> 
> And, along those lines, your change to the `README.md` file is an additional step in the wrong direction. The `jenkins/run jenkins/tox` command-line _was_ the way that developers could run the tests in exactly the same way as the CI ran them (back when the CI was using Tox...).

I am not seeing how this was true.  This was a _third_ way tests could be run, where the developer method of tests could be run in the development container _used_ by Jenkins.  Notice that in https://github.com/distributed-system-analysis/pbench/blob/29f2e04832e075c967d56bcf84cbe0e4322d923c/jenkins/Pipeline.gy#L14 `jenkins/tox` is not used.

> Having this PR effectively reintroduce Tox to our development environment while at the same time removing access to it in our CI environment is unhelpful.

I am not seeing how this is a re-introduction.  If anything it simply removes the alternative (and confusing) 3rd way to run tests, while ensuring that for developers a much reduced test environment exists.

> FWIW, it looks like [Tox is happy to run PyTest](https://tox.wiki/en/3.24.4/example/pytest.html#pytest-and-tox), and [PyTest is happy to run individual tests, to run them with special configurations and on multiple CPUs, without external support](https://tox.wiki/en/3.24.4/example/pytest.html#using-multiple-cpus-for-test-runs),

Very cool.  The future PR to reunify developer/CI will take advantage of that.

> but I guess we need fancy support to facilitate running the legacy tests in portions and/or in parallel.

Unfortunately, yes, this is a requirement for the convenience of developers.

README.md
portante: See https://github.com/distributed-system-analysis/pbench/blob/29f2e04832e075c967d56bcf84cbe0e4322d923c/jenkins/Pipeline.gy#L14 -- the CI and developer mode of running tests are very different.  This is yet a 3rd way to run them.

agent/exec-unittests
```diff
@@ -0,0 +1,52 @@
...
+        printf -- "\nAgent pytest command succeeded\n\n"
+    fi
+fi
+
+function run_legacy {
+    printf -- "\n\n\nRunning agent/${1} legacy unit tests\n\n"
+    agent/${1}/unittests ${_posargs}
+    local _rc=${?}
+    if [[ ${_rc} -ne 0 ]]; then
+        printf -- "\nAgent ${1} legacy tests failed with '%s'\n\n" "${_rc}"
```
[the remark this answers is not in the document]
portante: Why should the error reporting have knowledge of what to expect?  Can it really hurt to keep it?

server/tox-unittests
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+prog="$(basename "${0}")"
+progdir="$(realpath -e $(dirname "${0}"))"
+
+# List of tox environments to run.
+tox_envs="server"
+
+${progdir}/../tox-unittests ${tox_envs}
```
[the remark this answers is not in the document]
portante: Now that we reduced this the number of tox enviroments, it isn't anymore.

agent/tox-unittests
portante: This file is no longer really useful.  Removing.

agent/exec-unittests
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
+# Test execution help for the "agent" tox environment.  It is not intended
+# to be executed outside of a tox environment.
+
+_envdir="${1}"
+_subtst="${2}"
+shift; shift
+_posargs="${@}"
```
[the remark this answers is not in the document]
portante: It started out with me taking the syntax from the `tox.ini` file for the command execution and to be sure I was using them correctly, I added the leading underscore.  And then never switched back.

agent/exec-unittests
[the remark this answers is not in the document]
portante: The `unittests` might emit a different value.

server/exec-unittests
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+# Test execution help for the "server" tox environment.  It is not intended
+# to be executed outside of a tox environment.
```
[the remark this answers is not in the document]
portante: Yes, i have unified these two `exec-unittests` at the top level.

agent/exec-unittests
```diff
@@ -0,0 +1,52 @@
...
+
+rc=0
+if [[ -z "${_subtst}" || "${_subtst}" == "python" ]]; then
+    printf -- "\n\n\nRunning agent python3-base unit tests via pytest\n\n"
+    pytest --cov --cov-append --cov-report html:${_envdir}/cov/html ${_posargs} \
+        ./lib/pbench/test/unit/common \
+        ./lib/pbench/test/unit/agent \
+        ./lib/pbench/test/functional || rc=1
+    if [[ ${rc} -ne 0 ]]; then
+        printf -- "\nAgent pytest command failed with '%s'\n\n" "${rc}"
```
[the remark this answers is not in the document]
portante: Why should the error reporting have knowledge of what to expect? Can it really hurt to keep it?

portante (commented):
I believe this now addresses all review comments.

npalaska (dismissed):

riya-17:
Can't we keep a common `test-requirements.txt` file? They are almost same.

riya-17 (dismissed):

agent/exec-unittests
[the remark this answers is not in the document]
webbnh: Aaaynd, apparently that's a documented "feature" -- `shift` should return an error status in that case.  😝

My apologies for the bad lead.

webbnh:
> This was a third way tests could be run

Actually, I consider it to be an omission in my PR that updated the CI:  that PR _should_ have provided a more turn-key way for developers to invoke the tests in the container used by the CI, so that developers can run the tests exactly the way the CI does, and that PR should have updated the `README.md` accordingly.

Having this PR delete that section from the `README.md` instead of, say, correcting it to the current state of affairs (or putting a reference there to how the CI works) is a step away from where I think we should be heading.

Moreover, given that this PR now provides a clean, common way of running the tests, it should be a very small addition to make the CI use it.  (Is that assertion incorrect??)

The CI is a gatekeeper to being able to merge; ideally, prior to posting a PR, developers would "run the tests", such that when the CI encounters the change there are no failures; however, if the CI runs the tests differently from the way that the developers run them, then, even though they've "run the tests" (i.e., via the fruits of this PR), the developers have no assurances that the CI won't hit any problems.

So, we should make it easy for the developers to run the tests _in the same way_ that the CI runs them.  And, with this PR, I think that's nearly true.  With this PR, I think all a developer should have to do to run the tests is
```
jenkins/run tox -e lint,agent,server
```
The only things which are missing from this PR are (1) pointing this out in the `README.md` and (2) changing the `jenkins/Pipeline.gy` to do that.

However, if it turns out that that is _not_ enough to make it work, then I think that that is something that we should know _before_ this PR is merged:  if the CI cannot run the tests like the developers do, then I think that indicates that there is a gap in our coverage, and I think that should be addressed before we get too far out on this branch.

README.md
webbnh: Your statement that they "are" very different makes me regret my failure to delete all traces of Tox when I "fixed" the CI -- had I done so, then there would be no argument here....

And, given that my previous PR left us in a transitional state which was never completed, I would prefer that this PR not produce the analogous result.

webbnh:
> Can't we keep a common `test-requirements.txt` file? They are almost same.

That's a very interesting question.  Why _do_ the two have differing requirements for **testing**?  Does this indicate that one has gaps which the other has covered?  Or, does it indicate that they are testing the same things using differing methodologies (which might be undesirable)?

Does the Agent testing really use `ansible`?  How is it that the _testing_ is dependent upon Ansible when the Agent (according to its `requirements.txt`) is **_not_**?

How does the server use `freezegun`?  Is this something that the Agent should be doing?

Finally, the Server apparently uses `requests_mock` which the Agent doesn't; however, the Agent does make requests of (the Server's) RESTful interface...should the Agent be making use of `requests_mock`?  Or, if it can get by without it, should the Server be doing the same?

In any case, it does seem like maybe we should consider combining them.  The contents of these files _should_ be just the dependencies of the unit test code itself and only those dependencies which aren't already satisfied for the requirements of the CUT (which is where I would expect there to be differences between the Agent and the Server).  So, it seems like, assuming that we're using common methodologies, the testing should have common requirements.

webbnh (changes requested):
The changes look good.  I just have two outstanding issues.

The small one is, if the second argument to `exec-unittests` is blank, I think we should be insisting that the `_posargs` be empty, since its meaning is derived from the context given in the second argument and if that argument is missing then anything in the subsequent arguments is likely to be misinterpreted and result in confusion deep inside some sub-script.

The larger issue is I think this PR should include switching the CI to use the same command which this PR provides for developers' use, and the `README.md` file should describe how developers can run the tests like the CI does.  I don't think that this is a big ask, and, if it _is_ a big ask, then I think we need to understand that before this PR moves forward.

webbnh:
> the Server apparently uses `requests_mock` which the Agent doesn't

For the most part, both the Server and the Agent testing use the `responses` package.  What does the `requests_mock` package do that `responses` doesn't?  (@dbutenhof?)

portante:
Converted to `Draft` mode while I perform a significant rebase and update.

README.md
portante: We are going back the other way.  From what I can tell, `tox` gives us the best way for us to run tests in multiple python environments.

README.md
```diff
@@ -79,31 +79,65 @@ $ cd pbench
...
+
+  * `tox -e py36`    -- run all tests in a Python 3.6 environment (our default)
+  * `tox -e py39`    -- run all tests in a Python 3.9 environment
+  * `tox -e py310`   -- run all tests in a Python 3.10 environment
+  * `tox -e pypy3`   -- run all tests in a PyPy 3 environment
+  * `tox -e pypy3.8` -- run all tests in a PyPy 3.8 environment
+
+See https://tox.wiki/en/latest/example/basic.html#a-simple-tox-ini-default-environments.
+
+Each time tests are run, the linting steps (`block` and `flake8`) are run first.
```
ndokos, line 99: `black`?

exec-unittests
```diff
@@ -0,0 +1,90 @@
...
+        if [[ -z "${subtst}" || "${subtst}" == "util-scripts" ]]; then
+            run_legacy ${_major} util-scripts ${posargs} || rc=1
+        fi
+        if [[ -z "${subtst}" || "${subtst}" == "bench-scripts" ]]; then
+            run_legacy ${_major} bench-scripts ${posargs} || rc=1
+        fi
+    elif [[ "${_major}" == "server" ]]; then
+        if [[ -z "${subtst}" || "${subtst}" == "legacy" ]]; then
+            run_legacy ${_major} bin ${posargs} || rc=1
+        fi
```
webbnh, line 92: Since every one of the `run_legacy` command lines is unique anyway, we might as well replace the `${major}` with the value that it's going to have -- that will be clearer for the reader and it will prevent oopsies if the code gets disturbed.

(I wish there were a nice way to catch typos in the `${subtst}` value, but you can't do `elif`'s here because you want them _all_ to run by default.  🙁)

exec-unittests
```diff
@@ -0,0 +1,90 @@
...
+
+function run_legacy {
+    printf -- "\n\n\nRunning %s/%s legacy unit tests\n\n" "${1}" "${2}"
+    ${1}/${2}/unittests ${posargs}
+    local _rc=${?}
+    if [[ ${_rc} -ne 0 ]]; then
+        printf -- "\n%s %s legacy tests failed with '%s'\n\n" "${1^}" "${2}" "${_rc}"
+    else
+        printf -- "\n%s %s legacy tests succeeded\n\n" "${1^}" "${2}"
+    fi
```
webbnh, line 40: What does the `^` do?

webbnh (changes requested):
Can we defer the changes relating to building the Jenkins container to another PR?  I think there's some stuff to discuss there, and I don't want that to hold up getting the Tox changes in.

I think that the addition of the templated Dockerfiles and the `.gitignore` file and the removal of the GPG key files should be deferred.  (Does the addition of the `--security-opt` to `buildah` need to be in this PR?)  That will make it easier to move the Tox changes forward.

jenkins/run-tox
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Why do we need to do this?  Why not just invoke tox directly?  Because tox
+# does not reap any zombie processes.  During the course of our tests, the
+# agent/util-scripts tests for the Tool Meister system create "zombie" processes
+# as a natural consequence of how that sub-system has to work.  By running tox
+# via bash, bash takes care of reaping those zombie processes.
+
+tox ${@}
```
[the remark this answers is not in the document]
portante: I had thought of doing that, so sure.

exec-unittests
```diff
@@ -0,0 +1,90 @@
...
+shift
+if [[ -n "${major}" ]]; then
+    if [[ "${major}" != "agent" && "${major}" != "server" ]]; then
+        printf -- "Expected major sub-system to be either 'agent' or 'server', got '%s'\n" "${major}" >&2
+        exit 2
+    fi
+    major_list="${major}"
+else
+    major_list="agent server"
+fi
```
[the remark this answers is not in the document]
portante: `major_list` is set, `major` is not changed.  More re-work of this script happens in PR #2649.

exec-unittests
portante: Uppercase the argument.

exec-unittests
```diff
@@ -0,0 +1,90 @@
...
+if [[ -z "${subtst}" || "${subtst}" == "python" ]]; then
+    _pytest_majors="./lib/pbench/test/unit/common"
+    for _major in ${major_list}; do
+        _pytest_majors="${_pytest_majors} ./lib/pbench/test/unit/${_major}"
+        if [[ "${_major}" == "agent" ]]; then
+            _pytest_majors="${_pytest_majors} ./lib/pbench/test/functional"
+        fi
+    done
+
+    printf -- "\n\n\nRunning %s python3-based unit tests via pytest\n\n" "${major_list}"
```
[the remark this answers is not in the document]
portante: Sure, is it worth polishing?

exec-unittests
```diff
@@ -0,0 +1,90 @@
...
+    pytest \
+        --cov --cov-append --cov-report html:cov/html \
+        ${posargs} ${_pytest_majors} || rc=1
+    if [[ ${rc} -ne 0 ]]; then
+        printf -- "\n%s pytest command failed with '%s'\n\n" "${_major^}" "${rc}"
+    else
+        printf -- "\n%s pytest command succeeded\n\n" "${_major^}"
+    fi
```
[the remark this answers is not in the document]
portante: Uppercases the first letter, I believe.

exec-unittests
portante: Fixing.

portante (commented):
Applied code review feedback.

portante (commented):
Fixed the coverage report problem.

jenkins/run
```diff
@@ -33,13 +30,8 @@ if [[ -z "${IMAGE_REPO}" ]]; then
...
 
-IHOME=/container/home
-
-if [[ -z "${CONTAINER_HOME}" ]]; then
-    HOME_VOLUME=${IHOME}
-else
-    HOME_VOLUME="${CONTAINER_HOME}:${IHOME}:${CONTAINER_HOME_OPTION:="z"}"
-fi
+USER_NAME=developer
+HOME_DIR=/home/${USER_NAME}
```
dbutenhof, line 34: I thought I'd commented on this before, but I can't find it.

Mostly, it's amusing rather than distressing that you undid the ability to map the transient container HOME into a host path; that was something Webb suggested as a "one plus" and I thought might possibly be useful. I don't bemoan the loss, although I wonder if we'd find a use for it.

I don't see any advantage of `/home/${USER_NAME}` over `/container/home` or something else equally arbitrary since the container is really single-user; but it doesn't hurt. It makes the test containers a bit less consistent, but I doubt that matters in any practical way.

dbutenhof (approved):

jenkins/run
```diff
@@ -50,6 +50,7 @@ podman run \
     --volume ${HOME_DIR} \
     ${GIT_BASE_VOLUME} \
     -w ${HOME_DIR}/pbench \
+    --env-host \
```
dbutenhof, line 53: Interesting. I'd considered using this, and wondered if we'd risk container pollution this way, particularly in a local container test run as opposed to the relatively "pristine" Jenkins environment. What specifically was behind this decision?

dbutenhof (approved):

README.md
```diff
@@ -79,31 +79,68 @@ $ cd pbench
...
-Once tox is installed you can run the unit tests (use `tox --listenvs` to see
-the full list); e.g.:
+Once tox is installed you can run the unit tests against different versions of
+python using the python environment short-hands:
+
+  * `tox -e py36`    -- run all tests in a Python 3.6 environment (our default)
+  * `tox -e py39`    -- run all tests in a Python 3.9 environment
+  * `tox -e py310`   -- run all tests in a Python 3.10 environment
+  * `tox -e pypy3`   -- run all tests in a PyPy 3 environment
+  * `tox -e pypy3.8` -- run all tests in a PyPy 3.8 environment
```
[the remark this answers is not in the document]
portante: Yes, I have tried, but we have some module dependencies which we need to fix before we can use it.

jenkins/run
dbutenhof: Would it be better to transfer `COV_REPORT_LOC` specifically rather than throwing the entire pile over the fence? I'm not sure this is likely to be a real problem; just thinking in the abstract that keeping some isolation in the container is probably a good thing.

dbutenhof (approved):

dbutenhof (changes requested):

jenkins/Pipeline.gy
```diff
@@ -2,7 +2,7 @@ pipeline {
     agent { label 'pbench' }
     environment {
         COV_REPORT_LOC="cov.${BUILD_NUMBER}.xml"
-        EXTRA_PODMAN_SWITHCES="--env COV_REPORT_LOC=${COV_REPORT_LOC}"
+        EXTRA_PODMAN_SWITHCES="--env COV_REPORT_LOC=cov.${BUILD_NUMBER}.xml"
```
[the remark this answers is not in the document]
portante: Yeah, bugger all.

portante (commented):
Okay, this is looking good, but we don't want to merge it until the 1st commit is removed and I can build the development container images for `main`.

dbutenhof (dismissed):

jenkins/python-setup.sh
```diff
@@ -17,10 +17,22 @@
 # chosen over any installed ones, and it adds `/usr/sbin` to the end of the
 # path for the `ip` command, used by pbench-register-tool.
 
-PATH=$(python3 -m site --user-base)/bin:${PATH}:/usr/sbin
+function _pbench_pathins() {
+    if [[ ":${PATH:-}:" != *":${1}:"* ]]; then
```
webbnh, line 21: Does the `:-` add anything here?  (If `PATH` is empty or undefined, then `${PATH}` will expand vacuously which I think is the same result as what you have there.)

jenkins/run-unittests
```diff
@@ -1,41 +1,10 @@
...
-else
-    printf -- "\n\nTests succeeded.\n"
-    exit 0
-fi
+# For the CI jobs, we run unit tests expecting most of the environment to be
+# provided by the container image, and only a few modules not installable via
+# RPM need to be added (see jenkins/python-setup.sh).  We then instruct tox to
+# use the current environment (not setting up a virtualenv), and to run the
+# tests using whatever version of the python interpreter the install provides
+# (the `py` directs to use whatever it finds).
```
webbnh, line 8: This is an awesome explanation.

s/directs to/directs Tox to/

tox.ini
```diff
@@ -3,123 +3,40 @@ minversion = 3.5.0
...
-[testenv:jenkins-pytests]
-description = Runs all external tests from one tox environment
-commands =
-    bash -c "./jenkins/run-pytests {envdir}"
+    bash -c "{toxinidir}/exec-unittests {envdir} {posargs}"
+whitelist_externals =
+    bash
+    rm
+    find
+    env
```
webbnh, line 42: By the way, apparently `whitelist_externals` is [now deprecated](https://tox.wiki/en/latest/config.html#conf-allowlist_externals) -- it's been respelled `allowlist_externals`.

Apparently this is supposed to be a list of commands that we use in the `commands` section that we don't want to be warned about...I'm thinking that `rm`, `find`, and `env` aren't in that set anymore (and, I'm wondering if `detox` doesn't need to be added)....

webbnh (commented):
I didn't quite finish my review (sorry!  I've got one file to go...), but here's what I've got so far.  I'll try to finish tomorrow.

jenkins/python-setup.sh
portante: Yes, this is tried and true code from `agent/profile`.

tox.ini
portante: Not sure what is going on, but when I use `allowlist_externals`, I can the traceback as see in http://jenkins.intlab.perf-infra.lab.eng.rdu2.redhat.com/job/pbench-ci-multi/job/PR-2551/41/console.

portante (commented):
Hopefully we can move this forward soon!

These CI jobs take too long to run.  We really need to get to PR #2649.

dbutenhof (changes requested):
No good deed goes unpunished...

exec-unittests
```diff
@@ -59,7 +59,7 @@ if [[ -z "${subtst}" || "${subtst}" == "python" ]]; then
         fi
     done
 
-    printf -- "\n\n\nRunning %s python3-based unit tests via pytest\n\n" "${major_list}"
+    printf -- "\n\n\nRunning %s python3-based unit tests via pytest\n\n" "${major_list/ /, /}"
```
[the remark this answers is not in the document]
webbnh: Yeah, that.  (Sorry, Peter!)

jenkins/run
```diff
@@ -54,12 +46,12 @@ fi
...
-    --volume ${HOME_VOLUME} \
+    --volume $(pwd):${HOME_DIR}/pbench:z \
+    --volume ${HOME_DIR} \
     ${GIT_BASE_VOLUME} \
-    -w /src/pbench \
-    --env-host \
-    --env HOME=${IHOME} \
+    -w ${HOME_DIR}/pbench \
+    --env HOME=${HOME_DIR} \
+    --env USER=${USER_NAME} \
```
webbnh, line 54: Are you sure that we want to override the value of `USER`?  I think this will mean that the `uid` (e.g., as returned by `id -nu`) will disagree with the value in `USER`, which seems like it will cause confusion.  Does this accomplish anything other than to match `USER` to the last directory in `HOME` (which might be nice but hardly seems important...)?

jenkins/python-setup.sh
webbnh: What does it add (other than complication and confusion for the reader)?

jenkins/python-setup.sh
```diff
@@ -17,10 +17,22 @@
...
+_pbench_pathins "/usr/bin"
+_pbench_pathins "$(python3 -m site --user-base)/bin"
+unset -f _pbench_pathins
+export PATH
+
 unset PYTHONPATH
-pip3 install --user -r lint-requirements.txt -r docs/requirements.txt -r agent/requirements.txt -r server/requirements.txt -r agent/test-requirements.txt -r server/test-requirements.txt
+pip3 install --user --no-cache-dir -r lint-requirements.txt -r docs/requirements.txt -r agent/requirements.txt -r server/requirements.txt -r agent/test-requirements.txt -r server/test-requirements.txt
 python3 setup.py develop --user
+export PYTHONPATH=$(python3 -c "import sys; print(':'.join(x for x in sys.path if x))")
```
webbnh, line 35: What is exporting `PYTHONPATH` supposed to accomplish here?

From `python --help`:
> PYTHONPATH   : ':'-separated list of directories prefixed to the default module search path.  The result is sys.path.

So, I think what this would do is double all of the non-blank entries in `sys.path`, except that it looks like Python suppresses the duplication, such that this has no effect at all.

jenkins/python-setup.sh
```diff
@@ -17,10 +17,22 @@
...
 
 # We delete any python compiler caches to avoid conflicts with the python compiler
 # in the container.
```
webbnh, line 38: The `find` command on line 39 is ill-timed:  when this script is run, we're typically already inside the container, and we've already run Python several of times (lines 28, 33, 34, and 35), so we've already started creating cached Python compilations.

We really should do this earlier, and we should do it again after the end of the test run.  I suggest moving this to `jenkins/run-unittests` and doing it before we run `python-setup.sh` and again after `tox` finishes.

webbnh: Also, this seems to be intended to perform the same function as `utils/detox`.  Perhaps we should be using that script instead of reproducing the code here (note that the two are using slightly different criteria...).  Also, we're already running that script (via Tox) before invoking `python-setup.sh`, so perhaps we don't need this code here at all (although, I _do_ think it would be a good idea to clear these caches after the test run before we exit the container).

webbnh: Tox appears to offer a [`commands_post`](https://tox.wiki/en/latest/config.html#conf-commands_post) configurator.  Perhaps that would be an appropriate way to invoke `detox` (rather than doing it from either `python-setup.sh` or `jenkins/run-unittests`).  (It would be nicely symmetrical with the existing use of `commands_pre`.)

webbnh (dismissed):
Nearly there.  I just have a couple of things here to consider.

jenkins/run
portante: `USER` is no longer passed in.

jenkins/Pipeline.gy
```diff
@@ -2,6 +2,7 @@ pipeline {
     agent { label 'pbench' }
     environment {
         COV_REPORT_LOC="cov.${BUILD_NUMBER}.xml"
+        EXTRA_PODMAN_SWITCHES="--env COV_REPORT_LOC=cov.${BUILD_NUMBER}.xml"
```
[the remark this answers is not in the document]
portante: I'll try it.

jenkins/run
webbnh: What does line 54 do?  It looks to me like it defines the `USER` environment variable inside the container.

jenkins/python-setup.sh
webbnh: For any Python script run _after_ line 35 is executed, the value of `PYTHONPATH` is added to the front of the list of directories in `sys.path`; so, if we set `PYTHONPATH` to the contents of `sys.path`, then that would produce a list of those directories twice...except that, apparently, Python is clever enough to de-duplicate the list somehow.

As an experiment, I executed line 35 over and over, and always got the same result, which seemed weird, since I would have expected each subsequent execution to include the previous execution's result in the value of `sys.path`...but, each time, I just got the normal list (which, by the way, includes `~/.local`).

jenkins/python-setup.sh
webbnh: Well, then, I propose that we omit it and go with
```{
    if [[ ":${PATH}:" != *":${1}:"* ]]; then
```

jenkins/Pipeline.gy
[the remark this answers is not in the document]
portante: Doesn't work, the command line arguments are not re-evaluated.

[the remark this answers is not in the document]
webbnh: 😢  Thanks for trying!

npalaska (dismissed):

exec-unittests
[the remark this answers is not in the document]
webbnh: Because the result is prettier, or more human-compatible?

portante:
Converted to `Draft` mode until we can land PR #2663, which will help us address some problems with the unit tests.

jenkins/run
portante: See future commits.

jenkins/python-setup.sh
portante: Have you tried that?  Could you propose that change for `agent/profile` and ensure it works before we try another experiment here?  It would really help the whole team to move on this work.

The reason for using the existing `agent/profile` method is to make sure we are properly adding the paths without losing the existing `PATH` found in the container and without adding an unintentionally occluded path entry.

jenkins/python-setup.sh
portante: Without this defined we get:

```
py run-test: commands[0] | black --check .
  File "/home/developer/.local/bin/black", line 5, in <module>
    from black import patched_main
ModuleNotFoundError: No module named 'black'
```

jenkins/python-setup.sh
portante: Okay, so this comment prompted a very deep rabbit hole investigation.

We found during this investigation that `utils/detox` is not actually doing anything substantially useful prior to this PR.  The reason is that if you look at the `tox.ini` file, and the CI job's `jenkins/run-pytests` (prior to this PR) you'll notice that we run the `pytests` specifying files in the source tree.  So all tests then run in the source tree, creating `__pycache__` directories all over it.  While there are `__pycache__` directories under the tox virtualenv directories, they are not for our code.

The newly added commit (is/will be) titled, "detox tox without detox", addresses all this by doing the following:

 1. Properly installing our source into the virtualenv
 2. Running the tests by the module path reference so that they execute out of the installed sources
 3. Working to ensure no other files are created in the source tree

This removes the need for `utils/detox` at all.

jenkins/run
portante: So this commit is landed by GitHub across multiple commits.  So I am not entirely sure what version this comment is made against.

Considering the code compared against `main`, by dropping `--env-host`, which leaks all kinds of stuff into the container environment, we need a `USER` for certain programs and sub-systems to work.

portante (commented):
I have set this back to `Draft` mode until we can get the python test code reviewed and merged in PR #2663.

README.md
portante: @ndokos, fixed.

README.md
```diff
@@ -79,31 +79,68 @@ $ cd pbench
...
+  * `tox -- server python`  -- run server python tests (via `pytest`)
+
+For any of the test sub-sets on either the agent or server sides of the tree,
+one can pass additional arguments to run a specific test:
 
-  * `tox -e util-scripts`  -- for agent/util-scripts tests
-  * `tox -e server`  -- for server tests
-  * `tox -e lint`  -- to run the linting and code style checks
+  * `tox -- agent bench-scripts test-CL` -- run bench-scripts' test-CL
+  * `tox -- server legacy test-28`       -- run server legacy test-28
```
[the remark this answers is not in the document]
portante: Fixed.

portante (commented):
Okay, should be ready for review now.

dbutenhof (approved):

jenkins/run
```diff
@@ -59,12 +51,12 @@ fi
...
```
webbnh, line 59: Since, for good reasons, we are no longer using `--env-host`, we need to define certain environment variables inside the container.  The definition of `HOME` is pretty obvious and straightforward, since I don't think that there are very many constraints on it, and it is dictated by the setup we've done with the mounted volumes.  However, I think that setting up `USER` is a little bit trickier.  In particular, I think we should be wary of setting it to an arbitrary value.

Under normal circumstances, I believe that `USER` is set to a value which comes from `/etc/passwd` or its distributed equivalent.  And, I think the value matches what is returned by `id -nu`.  However, with the proposed change, inside the container, the value of `USER` _won't_ match the `id` value.  I suspect that the result of that will be Bad, e.g.:
```
# echo $USER
developer
# id $USER
id: 'developer': no such user
# whoami
root
```
I'm concerned that this is going to lead to some sort of identity crisis.

jenkins/python-setup.sh
webbnh: Is this still true now that the virtual environment stuff is working?

.gitignore
```diff
@@ -1,5 +1,4 @@
 doc/*.html
-__pycache__
```
webbnh, line 2: Are you sure that you want to remove this?  (Won't this adversely affect human developers?)

Makefile
```diff
@@ -0,0 +1,9 @@
+GARBAGE = build cov AUTHORS ChangeLog .eggs lib/pbench.egg-info/ /var/tmp/${USER}
+
+all:
+	ls -ld ${GARBAGE}
+	utils/detox . -print
+
+clean:
+	rm -rf ${GARBAGE}
+	utils/detox .
```
webbnh, line 9: These targets are a little weird.

Normally `clean` removes targets and intermediates which were _created by this Makefile_....  And, normally, `make all` does something more than print stuff out....

Is there any particular reason why these two function were written in "make" instead of in, say, bash?

Is this "tool" of sufficient general use to be included in the repo?  And, do we really want it at the _top level_?  (Shouldn't it be in `utils` or similar?)

jenkins/run-unittests
```diff
@@ -1,41 +1,10 @@
...
-    exit 0
-fi
+# For the CI jobs, we run unit tests expecting most of the environment to be
+# provided by the container image, and only a few modules not installable via
+# RPM need to be added (see jenkins/python-setup.sh).  We then instruct tox to
+# use the current environment (not setting up a virtualenv), and to run the
+# tests using whatever version of the python interpreter the install provides
+# (the `py` directs Tox to use whatever it finds).
+source jenkins/python-setup.sh
+tox --current-env -e py -- ${@}
```
webbnh, line 10: How does this interface with the whole "detox doesn't work because it targets the wrong environment" question?

webbnh (changes requested):
This stuff all looks generally good, but I've got some lingering questions and concerns.

We no longer detoxify (i.e., clear out all of the cached Python compilations) because they now get put in an ephemeral file system inside the container, and so they disappear when the container is torn down?  If so, that's wonderful...but it makes me think that there is some stuff in this PR that we no longer need.

I have a concern about setting `USER` in the container to an arbitrary value.

And, a couple of smaller things....

.gitignore
portante: For whatever reason, it seems like `git` ignores them without adding it explicitly.  At least, with this in place I have not seen those "untracked" files.

jenkins/python-setup.sh
portante: The container job does not use virtualenvs, `--current-env` turns that off so that all the pre-installed RPMs of the container are used.  Basically, the container IS the virtualenv.

jenkins/run-unittests
portante: There is now only one environment in the CI job, that which is provided by the container.  The key is that `pytest` is no  longer being told to run tests by file name from the source tree, but from the `pbench` installed `test` module hierarchy.

Makefile
portante: I can take it or leave it, but found it very useful to cleanup the tree.

jenkins/run
portante: Not sure if I share the concerns, but let's see if a simple change does the trick.

portante (commented):
Good?

.gitignore
dbutenhof: That's interesting; a search for .gitignore and __pycache__ shows lots of recommended `.gitignore` templates that include `__pycache__`. I don't see anything indicating that it should be ignored automatically. I wonder what's going on?

dbutenhof (dismissed):
I'm not sure the `USER` change really makes much difference; but while I do wonder if we'll ever run into issues I find that I can't muster much real concern right now. 😄

jenkins/python-setup.sh
webbnh: > The container job does not use virtualenvs

So that's what is stuck in my craw, here:  if we're running natively and not in a virtualenv, then setting the `PYTHONPATH` to `sys.path` should have no net effect!

If setting it like this is actually necessary in order to get something like Black to work, I'd like to understand why!...

portante: Certainly, but can we not have this hold up this work?

.gitignore
portante: So, so why it doesn't show up is because empty directories don't show up.  But if `.pyc` files get created in the tree that DO show up.  Which is what we want, because that should not happen.

portante: I certainly agree with looking for files in the tree use `git status --ignored`, etc.

Nobody should be using `git add .`, or `git add *`, or without being really sure.  There are lots of files that we _could_ to `.gitignore` to avoid things, but the `__pycache__` files are not one of them.

The creation of these files does not come from normal development tasks, or otherwise.  They come from somebody doing something wrong (at least with this commit) in how the python code was executed in the tree.

So _if_ they get created they should not be hidden, but should be front and center so that folks don't continue to use them.

Perhaps a `pre-add` git hook would help here.

jenkins/python-setup.sh
webbnh: OK, so I'm trying to close the loop on this....  The plan is to install our code into a virtualenv, run the tests out of there, and then dispose of the virtualenv when we're done...which all sounds great, when we have a virtualenv.  

However, if we're running in the Jenkins container, then, as you've said elsewhere, "the container _is_ the virtualenv", but...in that environment I think we're running the tests out of the source tree, and the source tree is a mapped volume such that any Python cached compilation files end up being persistent...so we need something like detox running at the end of the test...I think....  Help?

jenkins/python-setup.sh
webbnh: > Have you tried that?

Interactively, sure...and it works as documented and expected.

> The reason for using the existing `agent/profile` method is [...]

I understand what the code is endeavoring to do; I'm objecting to a detail of how it is doing it.  Moreover, if we're so intent on doing the same thing here that we do in `agent/profile`, then we should be sharing that code, not copying it...then we can fix it in one place....

webbnh (commented):
Looks good.  Just my lingering concerns to clear up:

- [x] The unnecessary `:-` expansion in `jenkins/python-setup.sh`
- [ ] The mysterious setting of `PYTHONPATH` in `jenkins/python-setup.sh`
- [ ] The removal of the `__pycache__` entry from `.gitignore`
- [ ] The questions around whether we are really free of the need for detoxification
- [x] The top-level `Makefile`

jenkins/python-setup.sh
portante: If you can propose a way to share the code, perhaps a PR would be helpful?

I have noodled about this, and I don't know how to share the `agent/profile` for those functions in the this file without going through contortions that would affect the `agent` code.  So this seemed like a good case to violate the `DRY` principle.

As for dropping the `:-`, if that will allow this PR to get merged, I'll make the change.

.gitignore
webbnh: > _if_ they get created they should not be hidden

They aren't hidden:  they are visible via `ls`, `find`, et al.  `git status`, particularly without the `--ignored` switch, is simply the wrong tool for that job.

And, sure, a Git hook would help, but we would be setting it up to do the job that the `.gitignore` file is supposed to do up front.

> So, so why it doesn't show up is because empty directories don't show up. But if `.pyc` files get created in the tree that DO show up.

I'm not quite sure what you're trying to say, here.  If you want to propose a change to the `.gitignore` file to instead ignore `.pyc` files, I'm good with that.  If we need to add a trailing slash to the existing `__pycache__` entry to allow it to be properly treated as a directory, I'm good with that, too.  But, I really think we should have Git ignore the Python cache files.

jenkins/python-setup.sh
webbnh: I don't know how to answer, that, Peter.

Based on your experiments, we've found it necessary to add code which executes a statement which AFAICT should have no effect.  That sounds dangerously close to voodoo to me.

At the very least, we should have a nice fat comment explaining either why we are doing that or pointing out the fact that we need it and don't know why.  Otherwise, someone is going to come along later and spend a lot of time trying to delete it (or, at least, wondering why we're doing it).

jenkins/python-setup.sh
webbnh: I believe in Bash the `source` command can function like `import` does in Python, allowing you to pull in function definitions from external files...so we could build a "library" of useful Bash functions and pull it in to our favorite scripts (sort of like how we execute `agent/base`, now).

Alternatively, we could contemplate executing `agent/profile` (perhaps with some parameterization) from this script, since, in both cases, we seem to be trying to accomplish the same things.

But, as you say, that's for another PR.

portante:
Are these concerns worth holding up this commit or do they need to be resolved before we can merge it?

I believe the value of this work far out-weighs both the mystery of the `PYTHONPATH` setting requirement and the detox removal.

I personally don't know how to answer those questions right now, so if you can help drive an answer, great.

Please propose a separate PR if you think those changes are not needed and work to verify in both the Jenkins CI, running the container locally, and running `tox` outside the PR.

jenkins/run-unittests
portante: `tox` doesn't look for tests.  Tox bundles everything after the `--` as `posargs`.  We then pass those `posargs` to the new `exec-unittests`.  In turn, that tells `pytest` to find tests in the `PYTHONPATH` using the module names, and then the `posargs` are added as args for `pytest`.

As I said above, now that `pytest` is directed to run tests by the installed modules, it does not litter the source tree with `__pycache__` directories.

webbnh:
> Are these concerns worth holding up this commit or do they need to be resolved before we can merge it?

My concerns are that neither you nor I understand the changes that you're proposing; I'm concerned that the result of the changes will therefore be fragile or subject to unexpected behavior.  Rushing to adding parallel execution when we don't understand the serial case seems unwise.

And, yes, I think it's worth holding up this merge...otherwise I wouldn't be doing so....

> I believe the value of this work far out-weighs both the mystery of the `PYTHONPATH` setting requirement and the detox removal.

I don't dispute the value of this work -- I support making these changes -- but I don't think we should accept them without understanding why that pointless setting of `PYTHONPATH` appears to matter and without understanding whether and why we don't need to get rid of the cached Python compilations.

> I personally don't know how to answer those questions right now, so if you can help drive an answer, great.

I'll try to pull your branch and take a look.

portante:
I don't know how to make further progress.  I would welcome any help folks can offer.  Thanks.

jenkins/python-setup.sh
webbnh: So, I've purchased our first clue.

Without an explicit `PYTHONPATH`, Black works fine in the container just before we invoke Tox, but it doesn't work inside the Tox invocation.  The difference is, outside of Tox, `sys.path` contains 
```
/usr/lib64/python39.zip:/usr/lib64/python3.9:/usr/lib64/python3.9/lib-dynload:/home/root/.local/lib/python3.9/site-packages:/usr/lib64/python3.9/site-packages:/usr/lib/python3.9/site-packages
```
but inside of Tox it contains
```
/usr/lib64/python39.zip:/usr/lib64/python3.9:/usr/lib64/python3.9/lib-dynload:/usr/lib64/python3.9/site-packages:/usr/lib/python3.9/site-packages
```
So, it's missing `/home/root/.local/lib/python3.9/site-packages` which is where the guts of Black have been installed.

Thus, by capturing the native `sys.path` in `PYTHONPATH`, we allow it to be inherited and used inside Tox, and that allows execution inside Tox to access stuff which wasn't installed there....

So, I think our Tox setup is missing a step.

dbutenhof (approved):
Webb has some valid concerns, and I'd like to get those resolved; but, on the other hand, I see this as valuable work and it's at worst a step forward. (Although I'm happy to see `__pytest__` restored to `.gitignore` as it preserves the ability to run tests locally and directly while protecting against PRs that inadvertently contain build results.)

jenkins/python-setup.sh
dbutenhof: Hmm. Inside the container our local Python artifacts would be in the transient `/home/${USER}/.local`, and the in-container `python3 -m site` ought to see and report that path. This seems to be saying that's not happening; but `black` is actually there? That's weird... and I wonder if that's connected to the discussion about mucking with the container `${USER}` value. ☹️

portante:
No one is saying @webbnh's concerns are not valid.  The question is, "are the concerns worth holding up this PR?"  I don't see how they are.  They can be solved by follow up work.  But the whole team is slowed down by the lack of parallel test runs, and it seems like a good balance to trade of solving these concerns for getting this merged to then allow the work on the parallel tests.

But if the team believes it is worth waiting, great.  I just need help debugging the concerns to try to get this to merge as fast as possible.

jenkins/python-setup.sh
webbnh: The sequence is:
- we enter the container
- we run, among other things, `python setup.py` which creates the `~/.local` installation (and Black works in this environment)
- we run `tox --current-env -e py`...and now things get fuzzier.  We're supposed to use the "current environment", but that no longer seems to include the `~/.local` installation.

We inherit the `PATH` value, which includes the Black executable/wrapper, but it cannot find its guts, because `sys.path` no longer includes `~/.local`.

`USER` is defined, and it has a reasonable value (`root`).

The problem is something to do with the Tox setup.  (I'm assuming that either we told it to do the wrong thing, or we didn't tell it everything to do.)  I think that defining `PYTHONPATH` outside and letting it be inherited is papering over the misconfiguration.

webbnh: The mystery has been unveiled.

Python adds the `--user` install directory to the `sys.path` any time that it exists (and/or is accessible).  However, it appears to locate the directory via the `HOME` environment variable.  And, it turns out that our `tox.ini` configuration was _not_ making that variable available to the test environment, so Python was not using there.  Once I fixed that, I was able to run Black (and Flake and PyTest) with this definition of `PYTHONPATH` commented out.

So, we can now remove this line from here, once we tweak `tox.ini`.

tox.ini
```diff
@@ -1,125 +1,35 @@
...
+description = Lints code and runs all agent and server unit/functional tests
 install_command = pip install --no-cache-dir --progress-bar off --prefix={envdir} {opts} {packages}
 passenv =
     PY_COLORS
     NO_COLORS
     TERM
     USER
+    PBENCH_UNITTEST_SERVER_MODE
+    PYTHONPATH
+    COV_REPORT_XML
```
webbnh, line 18: We need to add `HOME` to this list, to enable Python in the test environment to find the `--user` installation without using an inherited `PYTHONPATH`.

Also, I think we should remove `PYTHONPATH` from this list:  I don't think we want to have it inherited explicitly, since the test environment is supposed to be self-contained and we don't want it polluted by an outside definition.  Note that, if `PYTHONPATH` is explicitly defined in the user's environment, [Tox will pass it in _anyway_](https://tox.wiki/en/latest/config.html#conf-passenv), but if we don't have it listed here, then Tox should issue a warning, which I think is a good idea.

jenkins/run-unittests
webbnh: My apologies for conflating Tox and PyTest....

Anyway, as you helped me demonstrate, by virtue of invoking the `setup.py` with `--user`, the cached compilation files are all placed under the `${HOME}/.local` directory and not in the source tree, which is excellent, as, when running in a container, those are in an ephemeral file system by default.

However, unfortunately, the packaging step, which uses PBR, does leave some litter in the source tree (which in a container is in a mapped file system).  But, that we can address in a subsequent PR.

webbnh (approved):
With the addition of `HOME` to the `tox.ini` file and the removal of `PYTHONPATH` from the `python-setup.sh` file, this PR should be good to go.  (I suppose I should have "requested changes", but...I'd rather be "approving"....)

jenkins/python-setup.sh
webbnh: See this [other comment](https://github.com/distributed-system-analysis/pbench/pull/2551/files#r824224828).

npalaska (approved):

jenkins/branch.name
```diff
@@ -1 +1 @@
-main
+tox
```
dbutenhof, line 1: In an overabundance of caution, don't forget to fix this. 👍🏻

dbutenhof (changes requested):

portante:
Unfortunately, exposing `HOME` causes tests 51, 52, 53, 56, & 57 to fail in my local environment because I have a `.ssh/config` file, which is now found a copied.  Before it was hidden because `tox` never exposed `HOME` to the test sub-system.

So we'll have to work on this landing next week.

portante (commented):
The new `HOME` environment variable exposure through `tox` has caused a small problem for local testing.

If you happen to have a `${HOME}/.ssh/config` file, the unit tests (specifically, `agent/util-scripts` tests 51, 52, 53, 56, & 57) collect that file as part of their operation causing the resulting test gold files not to match the test execution output.  To prevent that, we have created a fake `HOME` directory for the `agent/util-scripts` tests.  However, that caused the tests to fail when run via the container CI because the "--user" installed pbench could not be found at `${HOME}/.local`.  To fix that, we also symlink a user's real `${HOME}/.local` to `${_testhome}/.local` (regardless of its existence) in order to still allow the tests to be invoked.

As part of the changes to create the "fake" `HOME` directory for the tests, we also clean up how the `agent/util-scripts` create the test environment, where the initial `rm -rf` would not report an errors, and any `mkdir -p` was changed to drop the `-p` so that we can detect bad setup state.

dbutenhof (approved):

webbnh:
> Unfortunately, exposing `HOME` causes tests 51, 52, 53, 56, & 57 to fail in my local environment because I have a `.ssh/config` file, which is now found

Gah!

So...functional tests need to be run in a known/properly-isolated environment...which Tox is supposed to provide for us...but when we run Tox in a container, we are disabling that (i.e., we're explicitly borrowing the "host" environment), and we need to provide `HOME` there, because that is used to find the `--user` install which we did outside the Tox environment...but that's a "bad" thing to do when we're _not_ using `--current_env`.

So, maybe we should not be passing `HOME` through, and we should be deferring the `--user` install until we're inside the Tox environment.  However, it seems weird to be running with an undefined `HOME`, but perhaps we can set it to point somewhere inside the Tox environment.  (Yeah, I know, when we're running Tox with `--current-env`, we mostly don't have a "Tox environment", but we do have some level of isolation...so, if we set up `HOME` to point at or below the Tox working directory or the Tox temporary directory, we'll maintain that isolation.)  With that in place, then we can do our install, and perhaps the pieces will fall into place naturally?

webbnh (changes requested):
I realize that we're feeling a great deal of pressure to get this PR merged, but I have to question whether this latest round of changes is a good choice.

At one point in the recent past, we had things "working" by allowing part of the external environment to leak into the test environment via a `PYTHONPATH` definition.  Then (at my instigation, I feely admit), we replaced that leak with what turned out to be another leak, via the `HOME` definition.  With this latest round of changes, we are plugging the `HOME` leak but only in  a couple of restricted cases, and we're undermining that plug with a bit of magic in the form of a symlink for `${HOME}/.local` buried in the test infrastructure where, to the casual reader, it's going to seem like magic.

In my opinion, the partial fix for `HOME` and the symlink are a worse workaround than the previous state (now that we understand it) where we were inheriting `PYTHONPATH` and leaving `HOME` undefined.

If we cannot move forward to establishing a properly-isolated `HOME` inside the Tox working area, then I recommend that we go back to passing in `PYTHONPATH` (with a nice explanatory comment of what that does and why it helps).

agent/util-scripts/test-bin/test-client-tool-meister
```diff
@@ -79,7 +79,7 @@ source ${_script_path}/common-tm-cleanup
 export PBENCH_REDIS_SERVER="localhost:17001;localhost:17001"
 export PBENCH_TOOL_DATA_SINK="localhost;localhost"
 
-_timeout pbench-tool-meister-start --sysinfo="${sysinfo}" "${group}"
+HOME=${_testhome} _timeout pbench-tool-meister-start --sysinfo="${sysinfo}" "${group}"
```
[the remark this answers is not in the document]
portante: `_testhome` is always defined by `agent/unittests`.

tox.ini
```diff
@@ -1,125 +1,35 @@
...
 
 [testenv]
-usedevelop = True
+description = Lints code and runs all agent and server unit/functional tests
 install_command = pip install --no-cache-dir --progress-bar off --prefix={envdir} {opts} {packages}
 passenv =
     PY_COLORS
     NO_COLORS
     TERM
+    HOME
```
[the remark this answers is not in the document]
portante: Unfortunately, that does not work because the way we have constructed the CI environment requires HOME to have a `.local` hierarchy with pbench installed in it, along with all of it's dependencies.

agent/util-scripts/unittests
```diff
@@ -63,6 +66,12 @@ let res=res+${?}
 cp -rL ${_tdir}/test-bin/* ${_testopt}/unittest-scripts/
 let res=res+${?}
 
+export _testhome=${_testroot}/home
+mkdir ${_testhome}
+let res=res+${?}
+ln -s ${HOME}/.local ${_testhome}/.local
```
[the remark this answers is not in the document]
portante: I believe this is worth moving forward with by filing an issue to remove it later, accepting this PR as is, moving on to focus on PR #2649, so that we can gain the benefits of the new `tox` "envs" and the parallelism while we work on solving the test infrastructure independently.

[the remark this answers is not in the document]
portante: FWIW, the run time of tests in the CI under PR #2649 is roughly 8 minutes, while the run time under this PR is 32 minutes (roughly).  The run time of `main` CI jobs is 29 minutes.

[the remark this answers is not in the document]
webbnh: Peter has assured me that this change limits and contains the Badness better than his previous change, that he's going to open an issue to note the existence of this Badness, and that it will soon be removed as we continue in our evolution toward better use of Tox.

tox.ini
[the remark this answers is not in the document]
webbnh: Peter assures me that, in a subsequent PR, we're going to wean the CI off of `--current-env`, at which point we won't need `HOME` here, and so this won't be a problem.

webbnh (approved):
It's gonna be great...eventually.

agent/util-scripts/unittests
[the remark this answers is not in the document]
portante: Filed tracking issue #2680.

tox.ini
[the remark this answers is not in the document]
portante: Tracked in #2681.

PR #2556, opened 2021-11-18
Spec file and ansible fixes for RHEL9

ndokos:
- Replace `ansible` dependency by `ansible-core` (more for documentation
than any other reason: `ansible-core` is installed by default on RHEL9)

- Add some python dependencies similar to RHEL8, although I had to
leave out `python3-click` which is not available (yet?)

- Fix the name of the COPR repo for RHEL9 in the ansible roles.

- The second commit adds an ansible role to punch holes in the firewall
for tool meister use

### DIFF
--- a/agent/ansible/pbench/agent/roles/pbench_agent_config/tasks/main.yml
+++ b/agent/ansible/pbench/agent/roles/pbench_agent_config/tasks/main.yml
@@ -17,3 +17,8 @@
     dest: "{{ pbench_key_dest }}"
     mode: "0600"
     files: id_rsa
+
+- name: "pbench agent configuration - open firewall ports for tool meister"
+  include_role:
+    name: pbench_firewall_open_ports
+
--- a/agent/ansible/pbench/agent/roles/pbench_firewall_open_ports/tasks/main.yml
+++ b/agent/ansible/pbench/agent/roles/pbench_firewall_open_ports/tasks/main.yml
@@ -1,0 +1,12 @@
+---
+# pbench agent configuration
+- name: "pbench agent configuration - open firewall ports for tool meister"
+  ansible.posix.firewalld:
+    port: "{{ item }}/tcp"
+    permanent: yes
+    offline: yes
+    state: enabled
+  with_items:
+    - "{{ pbench_redis_port }}"
+    - "{{ pbench_wsgi_port }}"
+
--- a/agent/rpm/pbench-agent.spec.j2
+++ b/agent/rpm/pbench-agent.spec.j2
@@ -8,6 +8,12 @@ URL:            https://github.com/distributed-system-analysis/pbench
 Source0:        pbench-agent-%{version}.tar.gz
 Buildarch:      noarch
 
+%if 0%{?rhel} < 9
+Requires:  ansible
+%else
+Requires:  ansible-core
+%endif
+
 %if 0%{?rhel} == 7
 Requires:  python3, python3-pip
 # For RHEL boxen we need the python development environment in order for the
@@ -19,6 +25,14 @@ Requires:  gcc python3-devel
 Requires:  python36, python3-pip
 # RPMs for modules in requirements.txt
 Requires:  python3-cffi, python3-click, python3-requests
+# RPMs for module dependencies
+Requires:  python3-docutils, python3-psutil
+%endif
+
+%if 0%{?rhel} == 9
+Requires:  python3-pip
+# RPMs for modules in requirements.txt
+Requires:  python3-cffi, python3-requests
 # RPMs for module dependencies
 Requires:  python3-docutils, python3-psutil
 %endif
@@ -38,7 +52,7 @@ Requires:  python3-psutil
 Requires:  perl, perl-Data-UUID, perl-JSON, %{?prefixjsonxs}perl-JSON-XS
 Requires:  perl-Time-HiRes
 
-Requires:  ansible, bc, bzip2, hostname, iproute, iputils, net-tools
+Requires:  bc, bzip2, hostname, iproute, iputils, net-tools
 Requires:  openssh-clients, openssh-server, procps-ng, psmisc, redis
 Requires:  rpmdevtools, rsync, screen, sos, tar, xz
 
--- a/agent/ansible/pbench/agent/roles/pbench_firewall_open_ports/README.md
+++ b/agent/ansible/pbench/agent/roles/pbench_firewall_open_ports/README.md
@@ -1,0 +1,40 @@
+pbench_firewall_open_ports
+===================
+
+This role punches a hole through the firewall for various ports that the tool meister
+processes need to use.
+
+Requirements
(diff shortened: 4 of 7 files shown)


webbnh (dismissed):
I have no problem with this change per se, but what about just dropping the requirement for ansible altogether?

I'm expecting us to drop `pbench-run-benchmark` which is half of the need; what uses `stockpile`?  Can we drop or decouple that?

ndokos:
We are probably going to go that way, but it's going to take a bit longer. This is a stopgap measure so that  we can be building pbench-agent RPMs for RHEL9 *now*.

webbnh:
As a stop-gap, why bother pulling in `ansible-core`?  Why not just conditionalize-out `ansible` and leave it at that?

ndokos:
I could do that. But since I cannot find a way to test the minor  version, I'm going to unconditionally jettison the dependency and see what @portante has to say :-)

webbnh (dismissed):
👍

portante (changes requested):
Let's not drop support partially.  If we can drop the `ansible`/`ansible-core` dependency entirely, great.  If we ship code that relies on it (`stockpile`) then we need to keep it.

Currently, we have not removed `pbench-run-benchmark`, which is the other dependency that needs it.  It is simple enough to add `ansible-core` for RHEL 9 as a conditional and then address the larger dependency issue in a more comprehensive follow-up.

riya-17 (approved):

agent/ansible/pbench/agent/roles/pbench_agent_config/tasks/main.yml
```diff
@@ -17,3 +17,8 @@
     dest: "{{ pbench_key_dest }}"
     mode: "0600"
     files: id_rsa
+
+- name: "pbench agent configuration - open firewall ports for tool meister"
```
portante, line 21: It is not clear to me that we want to unconditionally open firewall ports.  I think we need to offer the role in the documentation, and show how it can be enabled in the playbook (perhaps with a variable?), but this is something the admin needs to be aware of and deliberately perform themselves.

portante (approved):
Good for now, but we really need to move ownership of firewalld changes to the admin.

riya-17 (dismissed):
LGTM

agent/ansible/pbench/agent/roles/pbench_agent_config/tasks/main.yml
dbutenhof: What we need is a big warning that the software they're installing *won't work* without these ports open; we can open them, they can open them, or they might as well just quit out of the install right now and forget it.

Implicitly punching holes in their firewall is bad; it definitely needs to be a conscious and tracked decision of the admin. On the other hand, it shouldn't be "oh, the port's not open? no big deal" -- because it is a big deal. I have no idea how we balance this in the ansible; maybe we should give the installer a prompt, and a warning and a paragraph of instruction if they say "no"???

agent/ansible/pbench/agent/roles/pbench_firewall_open_ports/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+# pbench agent configuration
+- name: "pbench agent configuration - open firewall ports for tool meister"
+  ansible.posix.firewalld:
```
dbutenhof, line 4: Is this sensitive to whether `firewalld` is installed or running, or is extra logic required? (Probably not an issue if we require someone to run this manually; but definitely something we should think about if it's run implicitly from the main install playbook.)

agent/rpm/pbench-agent.spec.j2
```diff
@@ -23,6 +29,14 @@ Requires:  python3-cffi, python3-click, python3-requests
 Requires:  python3-docutils, python3-psutil
 %endif
 
+%if 0%{?rhel} == 9
+Requires:  python3-pip
+# RPMs for modules in requirements.txt
+Requires:  python3-cffi, python3-requests
+# RPMs for module dependencies
```
dbutenhof, line 36: What does "module dependencies" mean? Are we explicitly requiring indirect requirements? If so, it might be nice (if possible) to mention which modules place these requirements to make it easier in the future to track them.

dbutenhof (approved):
Some minor concerns and issues that can be resolved later...

agent/ansible/pbench/agent/roles/pbench_firewall_open_ports/tasks/main.yml
ndokos: This is not sensitive to whether the firewall is running: it will unconditionally modify the configuration (that's problem 1). It will *not* restart `firewalld` however (that's problem 2 - or not, depending on your POV), so the settings will *not* take effect until it is restarted. I could make the settings `immediate`, but then the playbook would fail if the firewall is not running.

So the current state is:
- the settings will be unconditionally applied, but will take effect if `firewalld` is (re)started.
- the role is invoked by the `pbench_agent_config` role automatically: that's for testing convenience, but it's not going to fly for actual use.

We could  put the onus on the user: if they are running `firewalld`, we ask them to run the (separate) playbook, which not only opens the ports, but also restarts `firewalld` (or opens the ports `immediate`ly). OTOH, if they are not running `firewalld`, they don't have to run the playbook at all.

agent/rpm/pbench-agent.spec.j2
ndokos: I'm actually not sure and I did not stop to investigate who uses the `docutils` and `psutil` modules. This was cut-and-pasted from RHEL8. I'll open an issue.

ndokos: Opened #2583.

agent/ansible/pbench/agent/roles/pbench_agent_config/tasks/main.yml
ndokos: Maybe decouple the firewall role from pbench_agent_config completely? It can then be included (or not) in the main playbook, as the user desires. The documentation would include an example playbook including it, but with a comment describing the situation.

The other thing that's missing is the inverse role that undoes everything.

agent/rpm/pbench-agent.spec.j2
webbnh: [Blame](https://github.com/distributed-system-analysis/pbench/blame/13c20cbe3dfefed66fd8ddb32498c204233c3461/agent/rpm/pbench-agent.spec.j2) suggests that some guy named @portante  [put that code comment](https://github.com/distributed-system-analysis/pbench/commit/7f614fe041886a0fea352710cacd82d272ad2d75) in there as a replacement for the phrase "derived dependencies".

webbnh (approved):
👍

dbutenhof (approved):

webbnh (approved):
👍

PR #2561, opened 2021-11-30
Move index_mappings api to appropriate directory

npalaska:
- Move index_mappings api under lib/pbench/server/api/resources/query_apis/metadata_index directory.
- Replace index name in the index_mappings get URI where <index> parameter is not a direct Elasticsearch root index name, but a key like search, samples, timeseries.

### DIFF
--- a/lib/pbench/server/api/__init__.py
+++ b/lib/pbench/server/api/__init__.py
@@ -78,7 +78,7 @@ def register_endpoints(api, app, config):
     api.add_resource(
         IndexMappings,
         f"{base_uri}/index/mappings/<string:index_key>",
-        resource_class_args=(logger,),
+        resource_class_args=(config, logger),
     )
     api.add_resource(
         IndexSearch, f"{base_uri}/index/search", resource_class_args=(config, logger),
--- a/lib/pbench/server/api/__init__.py
+++ b/lib/pbench/server/api/__init__.py
@@ -23,7 +23,9 @@ from pbench.server.api.resources.query_apis.datasets_detail import DatasetsDetai
 from pbench.server.api.resources.query_apis.datasets_list import DatasetsList
 from pbench.server.api.resources.query_apis.datasets_publish import DatasetsPublish
 from pbench.server.api.resources.query_apis.elasticsearch_api import Elasticsearch
-from pbench.server.api.resources.query_apis.index_mappings import IndexMappings
+from pbench.server.api.resources.query_apis.metadata_index.index_mappings import (
+    IndexMappings,
+)
 from pbench.server.api.resources.query_apis.index_search import IndexSearch
 from pbench.server.api.resources.query_apis.metadata_index.namespace_and_rows import (
     SampleNamespace,
@@ -75,7 +77,7 @@ def register_endpoints(api, app, config):
     )
     api.add_resource(
         IndexMappings,
-        f"{base_uri}/index/mappings/<string:index_name>",
+        f"{base_uri}/index/mappings/<string:index_key>",
         resource_class_args=(logger,),
     )
     api.add_resource(
--- a/lib/pbench/server/api/resources/query_apis/metadata_index/index_mappings.py
+++ b/lib/pbench/server/api/resources/query_apis/metadata_index/index_mappings.py
@@ -89,8 +89,7 @@ class IndexMappings(ApiBase):
         request on index/mappings/search. Similarly other index documents can be
         fetched by making a GET request on appropriate index names.
 
-        We fetch the mapping by querying the template database. If the
-        template is not found in the database NOT_FOUND error will be raised.
+        The mappings are retrieved by querying the template database.
         """
         # Normalize and validate the index keys we got via URI string. These
         # don't go through JSON schema validation, so we have
--- a/lib/pbench/server/api/resources/query_apis/metadata_index/index_mappings.py
+++ b/lib/pbench/server/api/resources/query_apis/metadata_index/index_mappings.py
@@ -3,13 +3,23 @@ from logging import Logger
 from typing import Any, AnyStr, Dict
 
 from flask import jsonify
-from flask_restful import Resource, abort
+from flask.wrappers import Request
+from flask_restful import abort
+
+from pbench.server import PbenchServerConfig
+from pbench.server.api.resources import (
+    ApiBase,
+    Schema,
+    Parameter,
+    ParamType,
+    SchemaError,
+)
 
 from pbench.server.api.resources.query_apis.metadata_index import RunIdBase
 from pbench.server.database.models.template import Template, TemplateNotFound
 
 
-class IndexMappings(Resource):
+class IndexMappings(ApiBase):
     """
     Get properties of a document on which the client can search.
     This API currently only supports mapping properties of documents specified
@@ -56,23 +66,41 @@ class IndexMappings(Resource):
     }
     """
 
-    def __init__(self, logger: Logger):
-        self.logger = logger
+    def __init__(self, config: PbenchServerConfig, logger: Logger):
+        super().__init__(
+            config,
+            logger,
+            Schema(
+                Parameter(
+                    "index_key",
+                    ParamType.KEYWORD,
+                    required=True,
+                    keywords=list(RunIdBase.ES_INTERNAL_INDEX_NAMES.keys()),
+                    uri_parameter=True,
+                )
+            ),
+        )
 
-    def get(self, index_key: AnyStr) -> Dict[AnyStr, Any]:
+    def _get(self, index_key: AnyStr, request: Request) -> Dict[AnyStr, Any]:
         """
-        Return mapping properties of the document specified by the index name
-        in the URI. For example, user can get the run document properties by
-        making a GET request on index/mappings/search. Similarly other index
-        documents can be fetched by making a GET request on appropriate index names.
+        Return mapping properties of the document specified by the index key
+        in the URI (supported index keys are defined in RunIdBase.ES_INTERNAL_INDEX_NAMES).
+        For example, user can get the run document properties by making a GET
+        request on index/mappings/search. Similarly other index documents can be
+        fetched by making a GET request on appropriate index names.
 
         We fetch the mapping by querying the template database. If the
         template is not found in the database NOT_FOUND error will be raised.
         """
-        index = RunIdBase.ES_INTERNAL_INDEX_NAMES.get(index_key)
-        if not index:
-            self.logger.debug(f"Document index key is not valid {index_key!r}")
-            abort(HTTPStatus.NOT_FOUND, message="Mapping not found")
+        # Normalize and validate the index keys we got via URI string. These
+        # don't go through JSON schema validation, so we have
+        # to do it here.
+        try:
+            self.schema.validate(index_key)
+        except SchemaError as e:
+            abort(HTTPStatus.BAD_REQUEST, message=str(e))
+
+        index = RunIdBase.ES_INTERNAL_INDEX_NAMES[index_key["index_key"]]
         try:
             index_name = index["index"]
             template = Template.find(index_name)
--- a/lib/pbench/server/api/resources/query_apis/metadata_index/index_mappings.py
+++ b/lib/pbench/server/api/resources/query_apis/metadata_index/index_mappings.py
@@ -5,17 +5,17 @@ from typing import Any, AnyStr, Dict
 from flask import jsonify
 from flask_restful import Resource, abort
 
+from pbench.server.api.resources.query_apis.metadata_index import RunIdBase
 from pbench.server.database.models.template import Template, TemplateNotFound
 
 
 class IndexMappings(Resource):
     """
-    Get properties of the run-data document type on which the dashboard can search.
-    This API currently only returns mapping properties of run-data documents.
+    Get properties of a document on which the client can search.
+    This API currently only supports mapping properties of documents specified
+    in RunIdBase.ES_INTERNAL_INDEX_NAMES.
 
-    TODO: In future we can tweak this API to take the document type name from the user
-    and query that type name against the template table to return the appropriate mappings
-
+    Example: get /api/v1/index/mappings/search
     :return:
     {
         "@metadata":
@@ -59,16 +59,22 @@ class IndexMappings(Resource):
     def __init__(self, logger: Logger):
         self.logger = logger
 
-    def get(self, index_name: AnyStr) -> Dict[AnyStr, Any]:
+    def get(self, index_key: AnyStr) -> Dict[AnyStr, Any]:
         """
-        Return mapping properties of the document specified by the index name in the URI.
-        For example, user can get the run document properties by making a GET request on
-        index/mappings/run. Similarly other index documents can be fetched by making a GET
-        request on appropriate index names.
-        We fetch the mapping by querying the template database. If the template is not found
-        in the database NOT_FOUND error will be raised.
+        Return mapping properties of the document specified by the index name
+        in the URI. For example, user can get the run document properties by
+        making a GET request on index/mappings/search. Similarly other index
+        documents can be fetched by making a GET request on appropriate index names.
+
+        We fetch the mapping by querying the template database. If the
+        template is not found in the database NOT_FOUND error will be raised.
         """
+        index = RunIdBase.ES_INTERNAL_INDEX_NAMES.get(index_key)
+        if not index:
+            self.logger.debug(f"Document index key is not valid {index_key!r}")
+            abort(HTTPStatus.NOT_FOUND, message="Mapping not found")
         try:
+            index_name = index["index"]
             template = Template.find(index_name)
             mappings = template.mappings
 
@@ -85,4 +91,4 @@ class IndexMappings(Resource):
             self.logger.exception(
                 "Document template {} not found in the database.", index_name
             )
-            abort(HTTPStatus.NOT_FOUND, message="Mapping not found")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
--- a/lib/pbench/test/unit/server/query_apis/test_index_mappings.py
+++ b/lib/pbench/test/unit/server/query_apis/test_index_mappings.py
@@ -30,7 +30,7 @@ class TestIndexMappings:
         response body.
         """
         with client:
-            response = client.get(f"{server_config.rest_uri}/index/mappings/run")
+            response = client.get(f"{server_config.rest_uri}/index/mappings/search")
             assert response.status_code == HTTPStatus.OK
             res_json = response.json
             assert res_json == {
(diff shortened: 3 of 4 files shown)


lib/pbench/server/api/__init__.py
```diff
@@ -23,7 +23,9 @@
 from pbench.server.api.resources.query_apis.datasets_list import DatasetsList
 from pbench.server.api.resources.query_apis.datasets_publish import DatasetsPublish
 from pbench.server.api.resources.query_apis.elasticsearch_api import Elasticsearch
-from pbench.server.api.resources.query_apis.index_mappings import IndexMappings
+from pbench.server.api.resources.query_apis.metadata_index.index_mappings import (
+    IndexMappings,
+)
```
dbutenhof, line 28: Ugh; our nested paths are getting too long when `black` won't even let us import a single name from a module!

Not really a serious comment; and yet, `metadata_index` is relatively long and not all that meaningful... nor, for that matter, is `query_apis`. Sigh ... but let's not worry about it for this PR. (Having brought it up out of nowhere...)

lib/pbench/server/api/resources/query_apis/metadata_index/index_mappings.py
```diff
@@ -59,16 +59,22 @@ class IndexMappings(Resource):
...
-        request on appropriate index names.
-        We fetch the mapping by querying the template database. If the template is not found
-        in the database NOT_FOUND error will be raised.
+        Return mapping properties of the document specified by the index name
+        in the URI. For example, user can get the run document properties by
+        making a GET request on index/mappings/search. Similarly other index
+        documents can be fetched by making a GET request on appropriate index names.
+
+        We fetch the mapping by querying the template database. If the
+        template is not found in the database NOT_FOUND error will be raised.
```
dbutenhof, line 70: The user input parameter no longer directly corresponds to a "template" but rather to one of the keywords you've defined in `ES_INTERNAL_INDEX_NAMES`; so this is really "If the keyword isn't valid". I'm not entirely sure I like `NOT_FOUND` anymore, either, since we're really not representing a resource directly... I think I'd rather have it look like a keyword validation error (which it is) as we'd get from a schema validation.

In fact, you could do this by defining a class `Parameter` object using your `ES_INTERNAL_INDEX_NAMES` as the keyword list, and validate against that. This is more or less what I did in the `get` for `datasets_metadata.py` to validate URI parameter keywords. (Look at the definition and use of `GET_SCHEMA`.)

lib/pbench/server/api/resources/query_apis/metadata_index/index_mappings.py
```diff
@@ -85,4 +91,4 @@ def get(self, index_name: AnyStr) -> Dict[AnyStr, Any]:
             self.logger.exception(
                 "Document template {} not found in the database.", index_name
             )
-            abort(HTTPStatus.NOT_FOUND, message="Mapping not found")
+            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
```
dbutenhof, line 94: I've been wondering whether we should change all of these (on `Template` lookup failures) to 503/service-unavailable since they'll be resolved when we index our first tarball. (Though that's also my main motivation for thinking we ought to pre-load the mapping templates at startup rather than waiting for the first index operation, which might make this irrelevant.)

lib/pbench/test/unit/server/query_apis/test_index_mappings.py
```diff
@@ -30,7 +30,7 @@ def test_run_template_query(self, client, server_config, find_template):
         response body.
         """
         with client:
-            response = client.get(f"{server_config.rest_uri}/index/mappings/run")
+            response = client.get(f"{server_config.rest_uri}/index/mappings/search")
```
dbutenhof, line 33: The more I see the pattern, the more I think that the `index` in this URI has become meaningless. It's really too much of an Elasticsearch term. We're more ... I don't know ... getting the mappings for "views" of a dataset, whether the run summary, iteration data, TOC, or the more detailed timeseries data.

A related observation ... we currently tie "search" to the run index, but there's no real reason for this except that's the way the old dashboard search page worked. Given the flexible mappings/search mechanism we've got in the server a client could just as easily search any "view" of the data.

In the particular context of this PR, I think that using the keyword "run" instead of "search" for the current "mapping" of the dashboard search function probably makes sense, even if we don't emphasize the fact that you could just as easily use any of the keywords. 😄

dbutenhof (approved):
Several comments, though I'm not sure we need to (or should) resolve them for this PR if we consider it an incremental step forward...

lib/pbench/server/api/resources/query_apis/metadata_index/index_mappings.py
webbnh: I concur that the server shouldn't fail due to the template mapping being unavailable, and so, if it does, then it's not a matter of "not found" but rather an internal error (which I might be convinced should be reported as a 503 although I'm not "there" yet).

As for whether the server code should depend on the database being pre-populated during installation or not, I feel like that's going to be inconvenient for development (I think it would be better if the server code just "fixed" the situation when it encounters it).  But, perhaps this isn't the place for that discussion.

lib/pbench/test/unit/server/query_apis/test_index_mappings.py
webbnh: So @dbutenhof, should (perhaps not in this PR) "index" be replaced with "dataset"?  That might enable a URI of `/dataset/mappings/<keyword>`....

lib/pbench/server/api/__init__.py
webbnh: Perhaps we don't need "index" in _both_ `metadata_index` **and** `index_mappings`?

lib/pbench/server/api/resources/query_apis/metadata_index/index_mappings.py
```diff
@@ -59,16 +59,22 @@ class IndexMappings(Resource):
...
-        Return mapping properties of the document specified by the index name in the URI.
-        For example, user can get the run document properties by making a GET request on
-        index/mappings/run. Similarly other index documents can be fetched by making a GET
-        request on appropriate index names.
-        We fetch the mapping by querying the template database. If the template is not found
-        in the database NOT_FOUND error will be raised.
+        Return mapping properties of the document specified by the index name
+        in the URI. For example, user can get the run document properties by
+        making a GET request on index/mappings/search. Similarly other index
+        documents can be fetched by making a GET request on appropriate index names.
```
webbnh, line 67: We should probably change "index name(s)" to something like "document selector keyword" (😝), since, on the outside, we really shouldn't be admitting that indices exist (that's an implementation detail).  What we're returning here is basically the list of fields in the document type, right?  And, that has "nothing" to do with the index (at least from the requestor's point of view).

lib/pbench/test/unit/server/query_apis/test_index_mappings.py
dbutenhof: I hadn't made that observation with any particular name in mind, but I like that idea a lot. The pair of `/api/v1/datasets/mappings/<type>` and `/api/v1/datasets/search/<type>` would seem very clear.

lib/pbench/test/unit/server/query_apis/test_index_mappings.py
```diff
@@ -110,7 +108,7 @@ def test_with_no_index_document(self, client, server_config):
         present in the database.
         """
         with client:
-            response = client.get(f"{server_config.rest_uri}/index/mappings/run")
+            response = client.get(f"{server_config.rest_uri}/index/mappings/test")
```
webbnh, line 111: Since this is intended to fail, I would use a string which is more obviously bad than "test"...like "bad_data_object_type" or similar.

webbnh (approved):
Here are a few comments, but, some of them are nits and the others don't necessarily need to be addressed in this PR.

lib/pbench/server/api/__init__.py
npalaska: Webb, I think Metadata is a very broad term in our server context and it can mean multiple things, like metadata_table/user_metadata/datasets_metadata etc but I also concur that `metadata_index` is also not very useful name

lib/pbench/test/unit/server/query_apis/test_index_mappings.py
npalaska: I concur `/dataset/mappings/<keyword>` looks better, I will create an issue to track for replacing `index` with `dataset`

lib/pbench/server/api/resources/query_apis/metadata_index/index_mappings.py
npalaska: And I should also probably use the ApiBase to inherit instead of direct flask Resource class

dbutenhof (dismissed):
I guess I'm OK with deferring the renaming we're talking about in modules and URIs, so with that, this is probably fine.

lib/pbench/server/api/resources/query_apis/metadata_index/index_mappings.py
```diff
@@ -56,23 +66,41 @@ class IndexMappings(Resource):
...
-        making a GET request on index/mappings/search. Similarly other index
-        documents can be fetched by making a GET request on appropriate index names.
+        Return mapping properties of the document specified by the index key
+        in the URI (supported index keys are defined in RunIdBase.ES_INTERNAL_INDEX_NAMES).
+        For example, user can get the run document properties by making a GET
+        request on index/mappings/search. Similarly other index documents can be
+        fetched by making a GET request on appropriate index names.
 
         We fetch the mapping by querying the template database. If the
         template is not found in the database NOT_FOUND error will be raised.
```
webbnh: If the _key_ is not found, it's a malformed request; if the corresponding template is not found, then it's some sort of internal configuration error.  (I think I would just delete the second sentence here and let the code speak for itself.)

lib/pbench/server/api/resources/query_apis/metadata_index/index_mappings.py
```diff
@@ -56,23 +66,41 @@ class IndexMappings(Resource):
...
-        if not index:
-            self.logger.debug(f"Document index key is not valid {index_key!r}")
-            abort(HTTPStatus.NOT_FOUND, message="Mapping not found")
+        # Normalize and validate the index keys we got via URI string. These
+        # don't go through JSON schema validation, so we have
+        # to do it here.
+        try:
+            self.schema.validate(index_key)
+        except SchemaError as e:
+            abort(HTTPStatus.BAD_REQUEST, message=str(e))
```
webbnh, line 101: I'm confused..."JSON schema validation" validates only the _body_ of the request?

lib/pbench/test/unit/server/query_apis/test_index_mappings.py
```diff
@@ -108,9 +108,14 @@ def test_with_no_index_document(self, client, server_config):
...
-            response = client.get(f"{server_config.rest_uri}/index/mappings/test")
-            assert response.status_code == HTTPStatus.NOT_FOUND
-            assert response.json["message"] == "Mapping not found"
+            response = client.get(
+                f"{server_config.rest_uri}/index/mappings/bad_data_object_type"
+            )
+            assert response.status_code == HTTPStatus.BAD_REQUEST
+            assert (
+                response.json["message"]
+                == "Unrecognized keyword ['bad_data_object_type'] given for parameter index_key; allowed keywords are ['iterations', 'search', 'timeseries']"
```
webbnh, line 117: This is a nit, but do we want to name the parameter "index_key"?  To the extent that this name is externally visible and subject to upward compatibility constraints, shouldn't we name it something that isn't tied to the particular database implementation that we happen to be using at the moment?

webbnh (approved):
I've got a question and a couple of nits.  (If the question doesn't generate a change, whether you want to address the nits is up to you.)

Also, looks like you need a rebase....

lib/pbench/server/api/resources/query_apis/metadata_index/index_mappings.py
npalaska: Yeah I will need to update the comment here.

lib/pbench/server/api/resources/query_apis/metadata_index/index_mappings.py
npalaska: `index_key` is a dictionary represented as `{"index_key": "<index_key_name_from_uri>"}` this is how the flask will pass the object to our `_get` method and since I am using the the same name for the `_get` parameter it becomes `index_key = {"index_key": "<index_key_name_from_uri>"}`.

And we validate this dictionary against the class schema that is defined as:
```
            Schema(
                Parameter(
                    "index_key",
                    ParamType.KEYWORD,
                    required=True,
                    keywords=list(RunIdBase.ES_INTERNAL_INDEX_NAMES.keys()),
                    uri_parameter=True,
                )
            )
```

lib/pbench/test/unit/server/query_apis/test_index_mappings.py
npalaska: Yeah that's a valid point, but this is raised because the `KeywordError` is calculating all the valid keywords for the key in json where it failed. And we validate the URI key by making it a json schema, to change the behavior I guess we will have to update the way we do the validation of the URI parameter.

lib/pbench/server/api/resources/query_apis/metadata_index/index_mappings.py
webbnh: The answer to my question is, `ApiBase._dispatch()` calls `self.schema.validate()` but only for the body of the request, and only if the request is not a `GET`.

So, my follow-on question is, instead of calling `validate()` here, should we be enhancing `ApiBase._dispatch()` to handle that implicitly for us?  (@dbutenhof?)

lib/pbench/test/unit/server/query_apis/test_index_mappings.py
webbnh: My concern is not about what this message represents on the Server side of the interface but rather what the client (particularly the human) will see on the its side of the interface.

Most of the text of that message is fine ("Unrecognized keyword" is good; listing the allowable keywords is excellent; displaying the offending keyword is good too); what I'm objecting to is just "index_key" -- I think we should find something other than "index" to use in that token, because "index" is too Elasticsearch-y.

Something like "object_type_key" (although, that's a little long) would be good.  ("Data type"?  "Table type"? Something in that vein.)

webbnh (dismissed):
Looks good.

We can defer my lingering comments to another PR, if you don't want to address them in this one.

lib/pbench/server/api/resources/query_apis/metadata_index/index_mappings.py
dbutenhof: > So, my follow-on question is, instead of calling validate() here, should we be enhancing ApiBase._dispatch() to handle that implicitly for us? (@dbutenhof?)

I briefly considered wrapping query parameter validation into the framework when I added them for `GET /datasets/metadata` ... at the time that seemed like unnecessary complication. Now we've got payload body parameters, mapped Flask URI parameters, and query parameters ... and it'd be nice to have a coordinated standard mechanism to validate all of them. Hmm.

But, for the URI embedded parameters like this, we've apparently all forgotten that Nikhil did exactly that:

```
        try:
            if json_data:
                json_data.update(uri_parameters)
            else:
                json_data = uri_parameters
```

This updates the `Request.get_json()` dict with the `kwargs` query parameters supplied to `PUT`/`POST`, and validates them *all* at once. So simply adding `index_key` (and I agree with Webb that the name is perhaps a bit too "elastic-y") to the API's `Schema` should work just fine.

However ... due to the way Flask works, the parameters really aren't interchangeable; the Flask API mapping won't work if URI parameters are omitted, and I'm not sure I really like just lumping them all together.

I'm becoming tempted to enhance `_dispatch` to *separately* validate the `Request.get_json()` request body, the `kwargs` URI parameters, and the `Request.args` query parameters ... with separate `Schema` rather than lumping them all together.

But let's not blow up this PR, and I'll think about that later. For right now ... the `_dispatch` implementation is inconsistent with this subclass validation code, but it doesn't really conflict since the `_dispatch` validation will *ignore* the unknown `"index_key"` parameter rather than complaining...

lib/pbench/test/unit/server/query_apis/test_index_mappings.py
dbutenhof: We're talking about a "view", or a "slice", of the dataset, not really even a "data type". I agree I'd prefer to avoid referring to this as an "index" since our Elasticsearch document schema should be a backend implementation detail.

I think that in an ideal REST model, where "dataset" is a resource, we might have something more like

* `GET /datasets/<id>?view=overview` for the current `/datasets/detail`, (or perhaps `/datasets/<id>/overview`?),
* `GET /datasets/<id>?view=iterations`,
* `GET /datasets/<id>?view=samples`

If I hadn't faced the issue of trying to fit this into the dashboard, I probably would have just gone that way instead of the muddle we've got now, which grew out of the current dashboard service/model structure. Right now, I'm content to get a working set of comprehensive APIs that we can use and analyze and then throw out in favor of something better...

But for right now, I think I like "view" better than any other alternative here.

dbutenhof (approved):
I'm not happy with the `index_key` name, and the subclass "kwargs" validation is inconsistent with the changes previously made to the base class -- but neither is what I think I'd really like to have, and we can defer that cleanup. (Though I wouldn't mind another commit to change the name to something with less ... "elasticity"...)

lib/pbench/server/api/resources/query_apis/metadata_index/index_mappings.py
webbnh: > for the URI embedded parameters like this, we've apparently all forgotten that Nikhil did exactly that

@dbutenhof, I don't think the code you referenced gets executed for a `GET` request, which is what I think we're talking about here.

> let's not blow up this PR, and I'll think about that later.

Fair enough.  (Should one of us create an issue to track this?)

lib/pbench/test/unit/server/query_apis/test_index_mappings.py
webbnh: OK, given that the currently proposed endpoint is `/index/mappings/<string:index_key>`, I guess that having the parameter name include "index" is OK, since the base component of the URI is also "index".

I thought we had decided to change the base to "dataset", but apparently I'm confused.  So, if we're not changing the resource name then we don't need to change the parameter name.  🤷‍♂️

lib/pbench/server/api/resources/query_apis/metadata_index/index_mappings.py
dbutenhof: Yes, technically `GET` with URI parameters represents a bug in the prior PR because it misses the integrated validation step. Since I'd rather solve it a bit differently, it's OK not to fix that here. 😁 

I'll fork my response as an issue.

webbnh (approved):
👍

dbutenhof (approved):
I guess this was just a rebase?

PR #2566, opened 2021-12-02
Adding ssh for creating directory on Host machine

riya-17:
RHEL 9: scp -r no longer works the same as it does on RHEL 8 and earlier.

Adding ssh to create a directory on the Host machine and then performing `scp` operation.


dbutenhof (approved):
This looks good; although I can't help wondering whether this should be reported as a bug instead of working around it, I wasn't able to find anything that definitively says `scp` *should* be creating the root directory. (I just can't really imagine how anyone would have thought it was useful to *not* do that... ☹️ )

agent/util-scripts/pbench-copy-result-tb
```diff
@@ -65,6 +65,8 @@ fi
 # foo.tar.xz.md5.check file to foo.tar.xz.md5. That's the signal that the
 # agent has finished with this tarball.
 
+controller=$(basename ${controller_dir})
+ssh -n -i ${pbench_bin}/id_rsa ${ssh_opts} ${results_repo} "mkdir -p ${results_path_prefix}/${controller}"
```
webbnh, line 69: This is probably beyond the scope of this PR, but we should really consider using the `-F` switch to `ssh`, which would likely allow us to avoid having to specify things like `-i` and the other `${ssh_opts}` here.  This would make the scripts' command lines simpler as well as allowing us to do things like adding options (e.g., for persistent sessions) in a single, centralized place, rather than having to go and change each individual `ssh` invocation.

webbnh (approved):
> I just can't really imagine how anyone would have thought it was useful to not do that... ☹️

And, I can't imaging how anyone would have made such a change in behavior without hanging a big flashing light on it....

agent/util-scripts/pbench-copy-result-tb
webbnh: Tracked in #2567.

npalaska (approved):

portante (approved):

PR #2571, opened 2021-12-03
wip

portante:


npalaska (approved):
:+1:

webbnh (approved):
👍

webbnh (changes requested):
There should probably be attendant changes made to `server/rpm/pbench-server.spec.j2` and `jenkins/development.Dockerfile`, each of which are requiring/installing packages intended to support `flask-bcrypt`.

Now, it's possible that `Bcrypt-Flask` has the same dependencies, in which case there is no need to change those files.  It's also possible that `Bcrypt-Flask` is available as an RPM, in which case those files should be updated to install it.  Either way, I think there's a box to be checked, here....

dbutenhof (dismissed):
I don't see anything that looks like an RPM for `Bcrypt-Flask` (nor for `flask-bcrypt`).

It does appear that `Bcrypt-Flask` doesn't depend on `python3-bcrypt`. (The `__init__.py` seems to import its own `_bcrypt.abi3.so`.) That suggests we might be able to drop that dependency as Webb suggests; however while it'd be cleanest to do that now, I'm OK with getting this fix in quickly and doing cleanup later.

webbnh:
Given that this change doesn't affect correctness, per se, that it is in large part an effort to clean up the log contents, I think we should be able to take the time to touch up `server/rpm/pbench-server.spec.j2` and `jenkins/development.Dockerfile` now; otherwise, we won't remember why that stuff is there or be aware that we can remove it, and there won't be any impetus to do so.

portante (commented):
Just updated some comments, no much else to change.

dbutenhof (approved):

webbnh (approved):
👍

PR #2572, opened 2021-12-03
Dynamic `root_index_name`

riya-17:
`root_index_name` in the `commons.py` file is hard coded to `result-data-sample` where it needs to be script-based. Like for `namespaces_and_rows` PR we need `result-data-sample` as a `root_index_name` for finding the appropriate indices, while we need `run-toc` as `root_index_name` for finding indices related to that

### DIFF
--- a/lib/pbench/test/unit/server/query_apis/commons.py
+++ b/lib/pbench/test/unit/server/query_apis/commons.py
@@ -69,14 +69,16 @@ class Commons:
 
     def build_index_from_metadata(self, root_index_name: str) -> str:
         """
-            Retrieve the list of ES indices from the dataset index map metadata based on a given root index name.
+            Retrieve the list of ES indices from the dataset index
+            map metadata based on a given root index name.
 
             Args:
                 root_index_name: root index name
 
             Returns:
-                An Elasticsearch query URL string listing the set of indices containing documents for the specified
-                root index name.
+                An Elasticsearch query URL string listing the set of
+                indices containing documents for the specified root
+                index name.
         """
         drb = Dataset.attach(controller="node", name="drb")
         index_map = Metadata.getvalue(dataset=drb, key="server.index-map")
--- a/lib/pbench/test/unit/server/query_apis/commons.py
+++ b/lib/pbench/test/unit/server/query_apis/commons.py
@@ -33,6 +33,7 @@ class Commons:
             "hits": [],
         },
     }
+    RESULT_DATA_SAMPLE = "result-data-sample"
 
     def _setup(
         self,
@@ -66,7 +67,11 @@ class Commons:
 
         return f"/{index}"
 
-    def build_index_from_metadata(self, root_index_name):
+    def build_index_from_metadata(self, root_index_name:str) -> str:
+        """Args
+            - root_index_name: Retrieve the list of ES indices from
+                the metadata table based on a given root_index_name.
+        """
         drb = Dataset.attach(controller="node", name="drb")
         index_map = Metadata.getvalue(dataset=drb, key="server.index-map")
         index_keys = [key for key in index_map if root_index_name in key]
@@ -365,7 +370,7 @@ class Commons:
             pytest.skip("skipping " + self.test_http_exception.__name__)
 
         if self.use_index_from_metadata:
-            index = self.build_index_from_metadata("result-data-sample")
+            index = self.build_index_from_metadata(Commons.RESULT_DATA_SAMPLE)
         else:
             index = self.build_index(
                 server_config,
@@ -400,7 +405,7 @@ class Commons:
             pytest.skip("skipping " + self.test_http_error.__name__)
 
         if self.use_index_from_metadata:
-            index = self.build_index_from_metadata("result-data-sample")
+            index = self.build_index_from_metadata(Commons.RESULT_DATA_SAMPLE)
         else:
             index = self.build_index(
                 server_config,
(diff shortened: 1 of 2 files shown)


lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -66,10 +66,10 @@ def build_index(self, server_config, dates):
 
         return f"/{index}"
 
-    def build_index_from_metadata(self):
+    def build_index_from_metadata(self, root_index_name):
```
dbutenhof, line 69: You didn't add a type hint for the new parameter. However, while you're at it, it'd be nice to add the return type and a proper docstring as well! ("No good deed goes unpunished" as they say...)

```
def build_index_from_metadata(self, root_index_name: str) -> str:
```

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -365,7 +365,7 @@ def test_http_exception(
             pytest.skip("skipping " + self.test_http_exception.__name__)
 
         if self.use_index_from_metadata:
-            index = self.build_index_from_metadata()
+            index = self.build_index_from_metadata("result-data-sample")
```
dbutenhof, line 368: Having repeated literal strings makes it easy for someone to miss a typo; using symbols is better, as the language will detect typos for you. For example, a `Commons.RESULT_DATA_SAMPLE` ... or maybe mapping our root index names as an ENUM isn't a bad idea.

dbutenhof (changes requested):
It'd be good to avoid the repeated string literals, but I mostly want to see the type hints. 😄

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -365,7 +370,7 @@ def test_http_exception(
             pytest.skip("skipping " + self.test_http_exception.__name__)
 
         if self.use_index_from_metadata:
-            index = self.build_index_from_metadata()
+            index = self.build_index_from_metadata(Commons.RESULT_DATA_SAMPLE)
```
dbutenhof, line 373: You can reference this as `self.RESULT_DATA_SAMPLE` everywhere, regardless of whether you're in the base class or a subclass context. I also have a tendency to reference class symbols using a class name; but as Webb pointed out in a recent review, using `self` means the reference is a bit more flexible across refactoring. (At least, when you know that the current class is derived from the class defining the symbol!)

lib/pbench/test/unit/server/query_apis/test_namespace_and_rows.py
```diff
@@ -327,7 +327,7 @@ def test_query(
                 },
             },
         }
-        index = self.build_index_from_metadata()
+        index = self.build_index_from_metadata("result-data-sample")
```
dbutenhof, line 330: You missed this one.

lib/pbench/test/unit/server/query_apis/test_namespace_and_rows.py
```diff
@@ -397,6 +397,7 @@ class TestSampleValues(Commons):
     """
 
     SCROLL_ID = "random_scroll_id_string_1=="
+    RESULT_DATA_SAMPLE = "result-data-sample"
```
dbutenhof, line 400: Since this class extends `Commons`, you don't need to duplicate the symbol; and you can still reference it here as `self.RESULT_DATA_SAMPLE`.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -66,10 +67,14 @@ def build_index(self, server_config, dates):
 
         return f"/{index}"
 
-    def build_index_from_metadata(self):
+    def build_index_from_metadata(self, root_index_name:str) -> str:
+        """Args
+            - root_index_name: Retrieve the list of ES indices from
+                the metadata table based on a given root_index_name.
```
dbutenhof, line 73: The description here should really be the description of the *method* rather than of the parameter. Maybe:

```
    """
    Retrieve the list of ES indices from the dataset index map metadata based on a given root index name.

    Args:
        root_index_name: root index name

    Returns:
        An Elasticsearch query URL string listing the set of indices containing documents for the specified
        root index name.
    """
```

dbutenhof (changes requested):
Nice; but you don't need to duplicate the `RESULT_DATA_SAMPLE` symbol, and the docstring could use a bit of work.

lib/pbench/test/unit/server/query_apis/test_namespace_and_rows.py
riya-17: yeah. I thought the class is different.

lib/pbench/test/unit/server/query_apis/test_namespace_and_rows.py
riya-17: that was in a different class and only one, so I avoided making changes there.

dbutenhof (approved):

webbnh (approved):
This looks like a very good code improvement, but there is no indication (e.g., in the PR description) of _why_ this change is being made.  I assume that something else you are working on prompted you to make this change (in which case you deserve huge kudos for making this change separately from that effort!), but it would be good to have a hint here as to what that is (for people looking back from the future, as well as for those of us here in the present who are curious).

Otherwise, change just for the sake of change is generally a bad idea.

PR #2573, opened 2021-12-04
Add missing 0.69 agent generated metadata

portante:
Add the missing pbench-agent v0.69 generated metadata to the run index mapping, updating `test-7.0.1` to reflect the additions, and adding an adding an additional test which contains the `0.69` metadata.

This represents the minimal support needed for handling data from v0.69 clients going forward.


portante (commented):
Ready for review.

I kept this separate from the pbench-agent v0.71 metadata changes we need so that it is a commit that can be used to backport to the `b0.69` branch so that we can update production sooner than later.

dbutenhof (approved):
Interestingly, looking at `main`, I see `hostname_f` and `hostname_s` being indexed, but not `controller_orig`. This PR however clearly shows that it *was* indexed in 0.69.

portante:
> Interestingly, looking at `main`, I see `hostname_f` and `hostname_s` being indexed, but not `controller_orig`.

Can you point out where in the code this is being done explicitly?  I see only that these fields are being attached via https://github.com/distributed-system-analysis/pbench/blob/104d308ffbb173607b4903bcdd30b83558918249/lib/pbench/server/indexer.py#L3346.

> This PR however clearly shows that it _was_ indexed in 0.69.

I am not sure what you mean by "this PR clearly shows," since these fields don't have mappings in either the `main` or the `b0.69` branch.

I believe the fields are being indexed without an explicit mapping, which means they are being created with the default mapping type the Java detects.

So this particular PR provides the proper mappings and demonstrates that with a legacy "unit test".

dbutenhof:
> Can you point out where in the code this is being done explicitly? I see only that these fields are being attached via

No, actually; I'd convinced myself on Friday that I found them, but I can't locate them now and perhaps I imagined it. 😆 

> I am not sure what you mean by "this PR clearly shows," since these fields don't have mappings in either the `main` or the `b0.69` branch.
> 
> I believe the fields are being indexed without an explicit mapping, which means they are being created with the default mapping type the Java detects.
> 
> So this particular PR provides the proper mappings and demonstrates that with a legacy "unit test".

Yes, that's what I meant: your test clearly shows those fields *are* indexed even though we didn't have them in the mapping file.

I think, mostly, I was too lazy to look at 0.69 and wanted to point out that there might be a difference. Adding them to the mapping file is clearly a good change, and I approved despite the commentary.

webbnh (approved):
👍

PR #2576, opened 2021-12-08
Add a quick-n-dirty tar ball indexed verifier

portante:


server/bin/pbench-verify-indexed.py
```diff
@@ -0,0 +1,187 @@
+#!/usr/libexec/platform-python
+# -*- mode: python -*-
+
+"""Pbench Re-Index
```
dbutenhof, line 4: I guess we can tell where this code started ... but maybe you should update the docstring. 😆

server/bin/pbench-verify-indexed.py
```diff
@@ -0,0 +1,187 @@
...
+def gen_tb_list(archive):
+    """gen_tb_list - yield all controller/tarball names
+    """
+    with os.scandir(archive) as archive_scan:
+        # We are scanning the archive directory for all controller
+        # sub-directories.
+        for c_entry in archive_scan:
+            if c_entry.name.startswith(".") and c_entry.is_dir(follow_symlinks=False):
+                # Ignore the ".", "..", and any other ".*" subdirectories.
+                continue
```
dbutenhof, line 46: The comment is a bit misleading, since `os.scandir` doesn't return `.` or `..` entries in the first place; but that doesn't really change the code.

server/bin/pbench-verify-indexed.py
```diff
@@ -0,0 +1,187 @@
...
+
+    es = Elasticsearch(
+        [
+            {
+                "host": "elasticsearch.perf.lab.eng.bos.redhat.com",
+                "port": "9280",
+                "timeout": 600,
+            },
+        ]
+    )
```
dbutenhof, line 112: OK, it's hacked to a constant for a particular purpose; but as you've already got the config file parsed, which means you can't run this somewhere *without* a valid configuration file, I don't see any reason not to use the real configuration?

server/bin/pbench-verify-indexed.py
```diff
@@ -0,0 +1,187 @@
...
+    body = {
+        "query": {
+            "filtered": {
+                "query": {"query_string": {"analyze_wildcard": True, "query": "*"}}
+            }
+        }
+    }
+    res = es.search(
+        index="dsa-pbench.v4.run.*",
```
dbutenhof, line 121: Again, I don't see the advantage in hardcoding this.

server/bin/pbench-verify-indexed.py
```diff
@@ -0,0 +1,187 @@
...
+        fields=["@metadata.file-name"],
+        size=100000,
+    )
+    indexed = {}
+    indexed_cnt = 0
+    for hit in res["hits"]["hits"]:
+        tb_file = hit["fields"]["@metadata.file-name"][0]
+        assert tb_file not in indexed, f"Whoa, found a duplicate, f{tb_file}"
+        if tb_file.startswith("/pbench/archive"):
+            tb_file = f"/srv{tb_file}"
```
dbutenhof, line 134: Really? Huh.

server/bin/pbench-verify-indexed.py
```diff
@@ -0,0 +1,187 @@
...
+        indexed[tb_file] = False
```
dbutenhof, line 135: This is actually a bit weird; if you find the tarball in the *index*, you're marking it "not indexed" and then if you also find it on disk you're marking it "indexed". The logic would probably be easier to follow with some renaming?

dbutenhof (commented):
I think this could use some documentation/name tweaks; and if it's going in the repo, driving it from the config file rather than hardcoding ES and paths would be better.

Aside from that ... what's the actual report? 😆

webbnh (approved):
What Dave said:  if this is going into the repo, it would be good if it were less "dirty".  (E.g., this is a prime example of existing code being copied...thus, it would be good if this were "clean" so that when _it_ gets copied, the dirt doesn't propagate.)

server/bin/pbench-verify-indexed.py
```diff
@@ -0,0 +1,187 @@
...
+
+import pbench
+
+
+_NAME_ = "pbench-reindex"
+
+tb_pat_r = (
+    r"\S+_(\d\d\d\d)[._-](\d\d)[._-](\d\d)[T_](\d\d)[._:](\d\d)[._:](\d\d)\.tar\.xz"
+)
+tb_pat = re.compile(tb_pat_r)
```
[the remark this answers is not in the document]
portante: At some point we want to compile this once.  Why not here?

If we inline this code, then it will be compiled each time through the loop.

Seems worth one line of code.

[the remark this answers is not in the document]
dbutenhof: I think the point is that if we never use `tb_pat_r` after compilation, this might as well be `tb_pat = re.compile(r"...")` 😁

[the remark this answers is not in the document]
portante: Doh!

server/bin/pbench-verify-indexed.py
```diff
@@ -0,0 +1,187 @@
...
+        expand_wildcards="open",
+        body=body,
+        fields=["@metadata.file-name"],
+        size=100000,
+    )
+    indexed = {}
+    indexed_cnt = 0
+    for hit in res["hits"]["hits"]:
+        tb_file = hit["fields"]["@metadata.file-name"][0]
+        assert tb_file not in indexed, f"Whoa, found a duplicate, f{tb_file}"
```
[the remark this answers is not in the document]
portante: Agreed, just needs to be reported, not asserted.

portante (commented):
Ready for another round of review.

server/bin/pbench-verify-indexed.py
portante: Took a different approach.

server/bin/pbench-verify-indexed.py
portante: Took a different approach.

server/bin/pbench-verify-indexed.py
```diff
@@ -1,52 +1,42 @@
...
+from elasticsearch1 import Elasticsearch, helpers
 
+_NAME_ = "pbench-verify-indexed"
 
-_NAME_ = "pbench-reindex"
-
-tb_pat_r = (
+tb_pat = re.compile(
     r"\S+_(\d\d\d\d)[._-](\d\d)[._-](\d\d)[T_](\d\d)[._:](\d\d)[._:](\d\d)\.tar\.xz"
 )
```
webbnh, line 24: There's nowhere that we can import this from (like `pbench.indexer`) so that we don't have to keep redefining it?

server/bin/pbench-verify-indexed.py
```diff
@@ -61,105 +51,114 @@ def gen_tb_list(archive):
...
+    scanner = helpers.scan(
+        es_client,
+        index=f"{index_prefix}.v4.run.*",
         doc_type="pbench-run",
+        query=query,
         expand_wildcards="open",
-        body=body,
         fields=["@metadata.file-name"],
-        size=100000,
+        size=1000,
```
webbnh, line 116: This is (still) a magic number which should be declared and referenced symbolically.  (And, I suspect that 1000 records is pretty small, depending on how big each record is.)

server/bin/pbench-verify-indexed.py
```diff
@@ -61,105 +51,114 @@ def gen_tb_list(archive):
...
+    d_end = pbench._time()
+
     not_on_disk = []
-    for key, val in indexed.items():
-        if not val:
-            not_on_disk_cnt += 1
-            not_on_disk.append(key)
+    for tb_file in indexed.keys():
+        if tb_file not in on_disk:
+            not_on_disk.append(tb_file)
```
webbnh, line 148: Alternately,
```
    not_on_disk = [f for f in indexed.keys() if f not in on_disk]
```

server/bin/pbench-verify-indexed.py
```diff
@@ -175,13 +174,23 @@ def main(options):
...
 if __name__ == "__main__":
     parser = ArgumentParser(f"Usage: {_NAME_} [--config <path-to-config-file>]")
+    cfg_name_def = os.environ.get("_PBENCH_SERVER_CONFIG")
     parser.add_argument(
         "-C",
         "--config",
         dest="cfg_name",
-        default=os.environ.get("_PBENCH_SERVER_CONFIG"),
+        default=cfg_name_def,
         help="Specify config file",
```
webbnh, line 183: The local `cfg_name_def` variable doesn't seem like it's worth creating.

webbnh (commented):
Looks pretty good.  A quick pass turned up a few nits.  I'll do a proper review on Monday.

server/bin/pbench-verify-indexed.py
portante: Sure, there is a way to do it.  But to make the change for just one file system silly.  We'd have to use it in the other modules to make it worth doing.  I'll file separate commits here with that change.

server/bin/pbench-verify-indexed.py
portante: Yes, definitely a magic number.  It is going to be a magic number whether or not it is symbolically referenced.  I have made that change, but it seems rather "empty" to do on the `b0.69` branch.

server/bin/pbench-verify-indexed.py
portante: Left over from a failed experiment.

portante (commented):
Okay, now it is ready for review.

server/bin/pbench-verify-indexed.py
```diff
@@ -19,9 +19,10 @@
 
 _NAME_ = "pbench-verify-indexed"
 
-tb_pat = re.compile(
-    r"\S+_(\d\d\d\d)[._-](\d\d)[._-](\d\d)[T_](\d\d)[._:](\d\d)[._:](\d\d)\.tar\.xz"
-)
+tb_pat = re.compile(pbench.TAR_BALL_NAME_W_TAR_PAT_S)
+
+# Number of JSON documents to fetch in one batch for the scan operation.
+batch_size = 10000
```
dbutenhof, line 25: By normal conventions, that really ought to be `BATCH_SIZE` as a constant. 😁 

Although if we're not already drowning in one-pluses, you could promote this to a CLI parameter as `--batch` with a default value of 1000.

dbutenhof (approved):
I haven't looked at this in a while, but it's matured nicely. One minor comment that's either "consistency" or "one-plus" depending on which clause you want to consider. 😆

dbutenhof (approved):

server/bin/pbench-verify-indexed.py
webbnh: Cool!  You saw the two that I found and even found one more!  😀

server/bin/pbench-verify-indexed.py
webbnh: The point is that it's a "configuration parameter" or a "tunable", etc.  Those should be declared at the top of the class or module, or in a definitions file, so that they can be easily found and modified, rather than having to dive into the code to make adjustments.  Moreover, it's nice to have the semantic context which a symbolic reference provides.

Yeah, given that it's coming late to the game in the `b0.69`, it's a less satisfying effort, but it doing it is good practice and the next time we're tempted to encode a magic number we'll be more likely to do it symbolically.

portante (commented):
Just a rebase.

server/bin/pbench-verify-indexed.py
```diff
@@ -0,0 +1,192 @@
...
+    on_disk_indexed_cnt = 0
+    not_indexed = []
+    for ctrl, tb_name in gen_tb_list(archive_p):
+        tb_file = f"{ctrl}/{tb_name}"
+        on_disk[tb_file] += 1
+        on_disk_cnt += 1
+        if tb_file in indexed:
+            on_disk_indexed_cnt += 1
+        else:
+            not_indexed.append(tb_file)
```
webbnh, line 140: If `indexed` and `on_disk` were `set`'s, then you could do the `intersection()` and `difference()` to produce the `on_disk_indexed_cnt` and `not_indexed` (set) values.

Likewise for `not_on_disk`.

server/bin/pbench-verify-indexed.py
```diff
@@ -0,0 +1,192 @@
...
+    print(f"Indexed: {indexed_cnt:n}, not on disk {not_on_disk_cnt:n}")
+    print(
+        f"Run-time: {(end - start):.2f}s, query time {(q_end - start):.2f}s,"
+        f" disk time {(d_end - q_end):.2f}s, final pass {(end - d_end):.2f}s"
+    )
+
+    if not_on_disk:
+        print(f"Not on disk... ({not_on_disk_cnt:n})")
+        for tb_file in sorted(not_on_disk):
+            print(f"\t{tb_file}")
```
webbnh, line 163: Consider replacing this with
```
        print("\t".join(sorted(not_on_disk)))
```
Ditto for lines 166-167.

webbnh (approved):
It looks OK to me.  I did find a typo and another editorial that you might want to address.  And, I have a couple of minor code suggestions (e.g., using `str.join()` instead of explicit loops).  The biggest suggestion is to use `set`'s for the lists of indexed and on-disk files:  then you can use set operations to trivially determine which files are missing, and you can use `len()` to get counts.

But, if you aren't interested in making any further changes, let me know and I'll resolve the conversations.

server/bin/pbench-verify-indexed.py
portante: First entry wouldn't have a leading tab.  Since buffered I/O is used by default anyway it doesn't really matter.

server/bin/pbench-verify-indexed.py
```diff
@@ -0,0 +1,192 @@
...
+        expand_wildcards="open",
+        fields=["@metadata.file-name"],
+        size=BATCH_SIZE,
+    )
+    indexed = defaultdict(int)
+    for hit in scanner:
+        tb_file_p = Path(hit["fields"]["@metadata.file-name"][0])
+        tb_file = f"{tb_file_p.parent.name}/{tb_file_p.name}"
+        indexed[tb_file] += 1
+    indexed_cnt = len(indexed)
```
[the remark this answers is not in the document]
portante: I meant to add back the duplicate detection and failed to.

server/bin/pbench-verify-indexed.py
portante: I wanted the duplicate detection and it doesn't seem worth to iterate the results and not just make the checks that way.

server/bin/pbench-verify-indexed.py
webbnh: > First entry wouldn't have a leading tab.

Oh, sorry:
```
        print("\t" + "\t".join(sorted(not_on_disk)))
```
😁

> Since buffered I/O is used by default anyway it doesn't really matter.

I have no idea whether it matters or not at execution time -- we would have to do some sort of trial.  But, in terms of the source code, there is a difference...although, whether it's an improvement or not is perhaps in the eye of the beholder.

server/bin/pbench-verify-indexed.py
webbnh: I'm not sure what you mean by "duplicate detection"?  (Was that something you wanted during development but subsequently removed?)

It looks to me like your code does a search of the indexed "set" for each entry in the on-disk "set"...which seems like an O(n^2) approach.  I have no idea how the `set` operations are implemented, but I would like to think that `intersection()` and `difference()` are cleverer than that.

webbnh: > I'm not sure what you mean by "duplicate detection"?

Nevermind, I see, now.

I'll just point out that you can still do duplicate detection if you use a `set`-based approach, but it does take some of the beauty out of it.

webbnh (approved):
Okie.

I still think that a `set`-based approach would have been a thing of beauty, but...ah, well....

server/bin/pbench-verify-indexed.py
```diff
@@ -123,6 +123,17 @@ def main(options):
         indexed[tb_file] += 1
     indexed_cnt = len(indexed)
 
+    # Look for duplicates in the returned Elasticsearch data to help detect
+    # problems with the indexer.
+    duplicates = []
+    for tb_file, val in indexed.items():
+        if val > 1:
+            duplicates.append(tb_file)
```
dbutenhof, line 131: `duplicates = [n for n, v in indexed.items() if v > 1]` ???

The comprehension is not only more compact, I think it's easier to read. (Even if you use `name, val`, although I think single-character variables are usually pretty effective for small blocks like this.)

dbutenhof (approved):
It looks fine, but one-plus-ing is the game, right? 😆

dbutenhof (approved):

webbnh (approved):
I like the latest change.  😀

PR #2577, opened 2021-12-08
Nomenclature update index->dataset

npalaska:


lib/pbench/server/api/__init__.py
```diff
@@ -76,12 +76,14 @@ def register_endpoints(api, app, config):
...
-        IndexMappings,
-        f"{base_uri}/index/mappings/<string:index_key>",
+        DatasetMappings,
+        f"{base_uri}/dataset/mappings/<string:dataset_view>",
         resource_class_args=(config, logger),
     )
     api.add_resource(
-        IndexSearch, f"{base_uri}/index/search", resource_class_args=(config, logger),
+        DatasetSummary,
+        f"{base_uri}/dataset/summary",
```
dbutenhof, line 85: Hmm.

In terms of "view" mapping, yes, I'd thought maybe "run-data" is something like a "summary" vs the various details of "result-data-sample" and "result-data".

However, I think the basic "summary" view is `datasets/detail`, which shows an entire "run-data" document. And we have two separate "search" APIs for "run-data" -- `datasets/list` (which currently selects by user, time range, and controller but is moving to "metadata" constraint) and `index/search` which selects by document fields.

Maybe `datasets/search` would be good. And I think that would merge the old `index/search` and your new `datasets/values`, because they're all returning a set of documents based on document field value matches (derived from the namespace queries, which could also maybe be merged).

lib/pbench/server/api/resources/query_apis/datasets/namespace_and_rows.py
```diff
@@ -209,7 +209,7 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
         from a selected set of documents that belong to the given run id in
         the specified index.
```
dbutenhof, line 210: Here's an off-hand comment I'd considered earlier but rejected since we were following the guideline of supporting the current dashboard model... I was never entirely comfortable with changing our access criteria from "controller"/"name" at the "higher" levels to "run_id" at these "lower" levels. But that's how the dashboard worked. I also didn't really want to spread the "controller" out to more APIS.

But now, especially since I'm in the process of removing "controller" from the APIs ... I like the consistency of moving towards using dataset name for identification of a dataset, throughout. (And if that's a bit more work for the current dashboard code, "so what"? 🤷🏻 )

dbutenhof (changes requested):
A couple of "philosophical" issues that maybe we probably don't address here; the important criticism in this context is that I don't think we want "search" to become "summary".

The current `datasets/detail` might really be better named "summary" as it returns the "overview" `run-data` document information and "detail" might imply, well, *detail*. And the idea of consolidating the "search" based on the "view" keyword across all three indices (and possibly additional tool data indices as well) is appealing, giving `/datasets/search/<view>` where `<view>` could point at any "slice" of the dataset. E.g., "summary" for the run-data, sample, iterations, prometheus, uperf, etc ... (if we decide to extend it ... and right now we're not indexing tool data anyway so it doesn't matter).

lib/pbench/server/api/__init__.py
webbnh: > Maybe `datasets/search` would be good.

We have an existing branch in the URI "tree" labeled "dataset", singular.  I recommend against creating another branch labeled "datasets" (plural), as I don't think there is enough distinction between the two labels.

lib/pbench/server/api/__init__.py
```diff
@@ -23,11 +23,11 @@
...
+from pbench.server.api.resources.query_apis.datasets.dataset_mappings import (
+    DatasetMappings,
 )
-from pbench.server.api.resources.query_apis.index_search import IndexSearch
-from pbench.server.api.resources.query_apis.metadata_index.namespace_and_rows import (
+from pbench.server.api.resources.query_apis.dataset_summary import DatasetSummary
+from pbench.server.api.resources.query_apis.datasets.namespace_and_rows import (
     SampleNamespace,
     SampleValues,
 )
```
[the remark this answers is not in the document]
webbnh: Similarly, once we're settled on the names, it would probably be worth re-ordering the `api.add_resource()` calls, as well.  (Perhaps you could merge two commits:  one with the name changes, and a separate one with the reorderings.)

lib/pbench/server/api/resources/query_apis/datasets/dataset_mappings.py
```diff
@@ -95,11 +95,11 @@ def _get(self, index_key: AnyStr, request: Request) -> Dict[AnyStr, Any]:
         # don't go through JSON schema validation, so we have
         # to do it here.
         try:
-            self.schema.validate(index_key)
+            self.schema.validate(dataset_view)
         except SchemaError as e:
             abort(HTTPStatus.BAD_REQUEST, message=str(e))
 
-        index = RunIdBase.ES_INTERNAL_INDEX_NAMES[index_key["index_key"]]
+        index = RunIdBase.ES_INTERNAL_INDEX_NAMES[dataset_view["dataset_view"]]
```
webbnh, line 102: I'm confused...what is going on in this statement?  According to the type hint at line 84, `dataset_view` is a string...but from the access, it looks like it's a dictionary.  (And, having the name of the dict and the key value be the same looks quite weird to me.)

Perhaps the dict should be renamed to "url_parameters" or similar?  (And, the type hint should be corrected!)

lib/pbench/test/unit/server/test_endpoint_configure.py
```diff
@@ -51,12 +51,12 @@ def check_config(self, client, server_config, host, my_headers={}):
...
                 "elasticsearch": f"{uri}/elasticsearch",
                 "endpoints": f"{uri}/endpoints",
-                "index_mappings": f"{uri}/index/mappings/",
-                "index_search": f"{uri}/index/search",
                 "graphql": f"{uri}/graphql",
                 "controllers_list": f"{uri}/controllers/list",
+                "dataset_mappings": f"{uri}/dataset/mappings/",
                 "controllers_months": f"{uri}/controllers/months",
                 "datasets_namespace": f"{uri}/datasets/namespace/",
+                "dataset_summary": f"{uri}/dataset/summary",
```
webbnh, line 59: It would be good if this whole list were ordered, as well, either by the keys or by the values.

webbnh (dismissed):
I'm afraid that I don't have anything particularly cogent to add to the resource hierarchy or naming discussion, other than a few small detail bits.  (Find those and a few other comments, below.)

However, I would, once the new names are settled, like to see a few of the various lists put into order (perhaps in a separate commit).

In addition, Mr. Jenkins says that you missed an update to the imports in `test_namespace_and_rows.py`.

lib/pbench/server/api/resources/query_apis/datasets/namespace_and_rows.py
npalaska: Right using the dataset name instead of run id to identify the dataset is something we would want to do anyway.

lib/pbench/server/api/resources/query_apis/datasets/dataset_mappings.py
```diff
@@ -81,7 +81,7 @@ def __init__(self, config: PbenchServerConfig, logger: Logger):
             ),
         )
 
-    def _get(self, index_key: AnyStr, request: Request) -> Dict[AnyStr, Any]:
+    def _get(self, dataset_view: AnyStr, request: Request) -> Dict[AnyStr, Any]:
         """
         Return mapping properties of the document specified by the index key
         in the URI (supported index keys are defined in RunIdBase.ES_INTERNAL_INDEX_NAMES).
```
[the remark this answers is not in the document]
npalaska: I will just use the dataset view

lib/pbench/server/api/__init__.py
dbutenhof: Technically, we have an existing "branch" labeled "datasets" ... it's the addition of the singular that's inconsistent; but thanks for noticing that. 😄  Nikhil should fix that!

lib/pbench/server/api/resources/query_apis/datasets/dataset_mappings.py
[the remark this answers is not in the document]
dbutenhof: I think we probably ought to be consistent in naming from the abstract declaration in the superclass to here, and the generic isn't specific to your view parameter, datasets or otherwise. These are generically the JSON parameters to the call, although in the specific case of `_get` that's only the accumulated URI parameters as we don't support a JSON payload.

The abstract just uses "json_data" so let's stick to that. The docstring can point out the meaning of the specific keys we expect (in this context) to find withing `json_data`.

dbutenhof (changes requested):

lib/pbench/server/api/resources/query_apis/datasets/dataset_mappings.py
npalaska: Yeah The type hint needs to be updated, it is a dict

lib/pbench/server/api/__init__.py
```diff
@@ -76,12 +76,14 @@ def register_endpoints(api, app, config):
         resource_class_args=(config, logger),
     )
     api.add_resource(
-        IndexMappings,
-        f"{base_uri}/index/mappings/<string:index_key>",
+        DatasetMappings,
+        f"{base_uri}/dataset/mappings/<string:dataset_view>",
```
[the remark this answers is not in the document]
dbutenhof: This is still singularly singular `dataset/mappings` and inconsistent with all the others!

dbutenhof (changes requested):

dbutenhof:
> This pull request **introduces 1 alert** when merging [1447bbc](https://github.com/distributed-system-analysis/pbench/commit/1447bbcde70d94aff8ab6fb13604ab0b350438e4) into [13c20cb](https://github.com/distributed-system-analysis/pbench/commit/13c20cbe3dfefed66fd8ddb32498c204233c3461) - [view on LGTM.com](https://lgtm.com/projects/g/distributed-system-analysis/pbench/rev/pr-a29d92130a4ba93744141e1cbc60032bab7c3e98)
> 
> **new alerts:**
> 
> * 1 for Unused import

You should also always run `black .` and `flake8 .` from the root of your branch before pushing a commit; when we don't remember, we give LGTM an excuse to yell at us. 😆

riya-17 (approved):
Looks Good

dbutenhof (dismissed):
Looks good!

lib/pbench/server/api/resources/query_apis/datasets/datasets_mappings.py
```diff
@@ -19,13 +19,13 @@
...
 
-class DatasetMappings(ApiBase):
+class DatasetsMappings(ApiBase):
     """
     Get properties of a document on which the client can search.
     This API currently only supports mapping properties of documents specified
     in RunIdBase.ES_INTERNAL_INDEX_NAMES.
 
-    Example: get /api/v1/index/mappings/search
+    Example: get /api/v1/dataset/mappings/search
```
webbnh, line 28: s/dataset/datasets/

lib/pbench/server/api/resources/query_apis/datasets/datasets_mappings.py
```diff
@@ -81,25 +81,28 @@ def __init__(self, config: PbenchServerConfig, logger: Logger):
             ),
         )
 
-    def _get(self, dataset_view: AnyStr, request: Request) -> Dict[AnyStr, Any]:
+    def _get(self, json_data: JSON, request: Request) -> Response:
         """
-        Return mapping properties of the document specified by the index key
-        in the URI (supported index keys are defined in RunIdBase.ES_INTERNAL_INDEX_NAMES).
+        Return mapping properties of the document specified by the dataset view
+        in the URI (supported dataset views are defined in
```
webbnh, line 87: I suggest adding a word or two here, e.g., "by the dataset view _parameter_ in the URI" or " by the dataset view _specified_ in the URI _parameter_".

lib/pbench/server/api/resources/query_apis/datasets/datasets_mappings.py
```diff
@@ -81,25 +81,28 @@ def __init__(self, config: PbenchServerConfig, logger: Logger):
...
+        RunIdBase.ES_INTERNAL_INDEX_NAMES).
+
         For example, user can get the run document properties by making a GET
-        request on index/mappings/search. Similarly other index documents can be
-        fetched by making a GET request on appropriate index names.
+        request on dataset/mappings/search. Similarly other index documents
```
webbnh, line 91: s/dataset/datasets/

lib/pbench/server/api/resources/query_apis/datasets_search.py
```diff
@@ -17,9 +17,10 @@
 )
 
 
-class DatasetSummary(ElasticBase):
+class DatasetsSearch(ElasticBase):
     """
-    Create a search request based on given query parameter.
+    Pbench ES query API that returns run-data document sample rows after
+    applying client specified search term withing specified start and end time.
```
webbnh, line 23: s/withing/within/ ?

webbnh (approved):
Looks good.  There are just a few typos left.

webbnh:
@npalaska, at this point, I think we're ready for final review.  So, after you make any final edits, please squash this branch down to the commits that you intend to merge (and rebase it as necessary).  I recommend two commits:  one with the nomenclature changes and one with the reorderings, if you can pull that off.

I think you can do the following:
- use `git reset HEAD^` to pop the last commit;
- then use `git stash` to stash it; 
- then make your final edit and commit that; 
- then squash everything down to a single commit (using `git rebase -i origin/main`, while, while you're at it, will move your branch to the end of `main`); 
- then pop the stash and commit it.

That should result in two commits with your branch on the end of `main` and ready to merge.

npalaska:
Thanks @webbnh. I did the following:
1. git rebase -i upstream/main
2. `e` (edit) the commit I wanted to amend
3. git add <changed_files>
4. git commit --amend
5. git rebase --continue
6. git pull upstream main --rebase
7. squashed the necessary commits

To reviewers: The first (nomenclature update) commit is changed to address the typos Webb pointed out and no other new changes are present. The PR now has 2 commits, one to update the nomenclature and one to re-order the imports

dbutenhof (approved):

webbnh (approved):
Your approach works, too.  (GTG)

PR #2581, opened 2021-12-09
fix hardcoded index name in commons.py

npalaska:


lib/pbench/test/unit/server/query_apis/test_namespace_and_rows.py
```diff
@@ -26,7 +26,7 @@ def _setup(self, client):
             pbench_endpoint="/datasets/namespace/iterations",
             elastic_endpoint="/_search",
             payload={"run_id": "random_md5_string1"},
-            use_index_from_metadata=True,
+            index_from_metadata="result-data-sample",
```
riya-17, line 29: Do we really want to provide `index_name` explicitly? I thought we are trying to find `index_name` through `run_id` and `root_index_name`

riya-17 (approved):
This looks good :)

lib/pbench/test/unit/server/query_apis/test_namespace_and_rows.py
dbutenhof: We search the dataset's index map (discovered via run_id, although we should change to use dataset name) to find indices with the expected root index name.

This only changes how the test system gets that root index name.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -377,8 +373,8 @@ def test_http_exception(
         if not self.elastic_endpoint:
             pytest.skip("skipping " + self.test_http_exception.__name__)
 
-        if self.use_index_from_metadata:
-            index = self.build_index_from_metadata(self.RESULT_DATA_SAMPLE)
+        if self.index_from_metadata:
+            index = self.build_index_from_metadata()
         else:
             index = self.build_index(
                 server_config,
```
dbutenhof, line 380: This `if`/`else` is repeated several times, and that stands out in the pattern of your changes. It could really be refactored into a new superclass method, which would streamline the test cases a bit.

dbutenhof (approved):
Looks good; one minor "I couldn't resist" one-plus that doesn't need to be done now. (If you want to, great; if not, let's create a separate issue to think about it later.)

lib/pbench/test/unit/server/query_apis/commons.py
npalaska: Okay created an issue to track https://github.com/distributed-system-analysis/pbench/issues/2582

PR #2584, opened 2021-12-10
Miscellaneous fixes in support of a Containerized Pbench Server

webbnh:
This PR is for a set of miscellaneous fixes which I made leading up to providing a containerized Pbench Server (see #2585).

- The first change works around a bug in the RHEL 8 `libselinux` implementation, which results in a SEGV when the `restorecon` function is invoked with SELinux disabled (as it is, for whatever reason, in the RHEL 8 container).  The fix is simple and of general applicability, which is why it's included in this PR:  only call `restorecon` if SELinux is enabled.  However, it turns out that the Server's functional test suite mocks the `selinux` module, but it does so incompletely, so I had to add `is_selinux_enabled()` to the mock-module.
- The second change is centered on the `pbench-server.service` file which is used by `systemd` to run the Pbench Server.  Part of it is to correct the file permissions to remove the eXecutable bit, which `systemd` complains about.  The more substantive change is to establish a cooldown interval between failures:  if the Pbench Server crashes for some reason, it will wait 30 seconds before being restarted.  And, I've set it to allow 20 retries.  In the case of the containerized Pbench Server, on startup, the Server process attempts to connect to the PostgreSQL database before the database is ready to accept connections, which causes the Pbench Server to crash.  Without increasing the retry interval, the Pbench Server repeatedly crashes, exhausting its retries before the database becomes available, and so it never starts.  With the increased interval, the Pbench Server stabilizes after only a few attempts.  This change should have little impact in other deployment cases -- at the worst, it will increase the downtime for an unexpected failure to half a minute.
- The next change acknowledges the fact that the Pbench Server depends on `cron` by adding the `cronie` package to the list of required packages in the Server's RPM spec file.  Typically, on systems where the Server is installed, `cron` is already available; however, in the containerized environment, this is not the case, so we need to request it explicitly.
- The next change is basically a housekeeping update, adding `cronie` to the Jenkins development container recipe.  The intention is that the container should already include all the Pbench requirements, and this brings it up to date with the previous change in this PR.  This change also adds the `copr-cli` package, to enable building Server RPMs in the container.
- The last change adds `selinux` to the Server `requirements.txt` file since it was missing.

[Review invitations:  @portante, @dbutenhof, @ndokos, and anyone else who would like to chime in]

### DIFF
--- a/jenkins/development.Dockerfile
+++ b/jenkins/development.Dockerfile
@@ -92,6 +92,7 @@ RUN \
         `# using the Server RPM.  Then, if a 'pip3 install -r requirements.txt'` \
         `# shows any missing dependencies they should be added to the RPM spec` \
         `# file and this list should be regenerated.` \
+        cronie \
         npm \
         policycoreutils \
         policycoreutils-python-utils \
@@ -171,6 +172,7 @@ RUN \
         python3-sphinx \
         `#` \
         `# Install utilities for building RPMs, evaluating test results, etc.` \
+        copr-cli \
         diffutils \
         git \
         less \
--- a/server/bin/pbench-server-prep-shim-002.py
+++ b/server/bin/pbench-server-prep-shim-002.py
@@ -241,13 +241,16 @@ def process_tb(config, logger, receive_dir, qdir_md5, duplicates, errors):
             nerrs += 1
             continue
 
-        # Restore the SELinux context properly
-        try:
-            selinux.restorecon(dest / tb.name)
-            selinux.restorecon(dest / tbmd5.name)
-        except Exception as e:
-            # log it but do not abort
-            logger.error("{}: Error: 'restorecon {}', {}", config.TS, dest / tb.name, e)
+        if selinux.is_selinux_enabled():
+            # Restore the SELinux context properly
+            try:
+                selinux.restorecon(dest / tb.name)
+                selinux.restorecon(dest / tbmd5.name)
+            except Exception as e:
+                # log it but do not abort
+                logger.error(
+                    "{}: Error: 'restorecon {}', {}", config.TS, dest / tb.name, e
+                )
 
         # Now that we have successfully moved the tar ball and its .md5 to the
         # destination, we can remove the original .md5 file.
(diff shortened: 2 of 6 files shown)


server/bin/pbench-server-prep-shim-002.py
```diff
@@ -241,13 +241,16 @@ def process_tb(config, logger, receive_dir, qdir_md5, duplicates, errors):
...
+        if selinux.is_selinux_enabled():
+            # Restore the SELinux context properly
+            try:
+                selinux.restorecon(dest / tb.name)
+                selinux.restorecon(dest / tbmd5.name)
+            except Exception as e:
+                # log it but do not abort
+                logger.error(
+                    "{}: Error: 'restorecon {}', {}", config.TS, dest / tb.name, e
+                )
```
dbutenhof, line 253: Ha; you've broken the legacy gold files by removing these errors from the log.

jenkins/development.Dockerfile
```diff
@@ -171,6 +172,7 @@ RUN \
         python3-sphinx \
         `#` \
         `# Install utilities for building RPMs, evaluating test results, etc.` \
+        copr-cli \
```
dbutenhof, line 175: I suppose it will eventually become clear why we need this?

dbutenhof (dismissed):
Although you need to repair the gold files now that you've fixed a traditional problem...

jenkins/development.Dockerfile
webbnh: Nope, it won't...unless you try to build the Server RPM...at which point you'll need to install this package on the platform on which you're trying to run the build...but, if you're building it in the development container, then, because it will already have been installed by this change, it won't be a problem, and you'll never know!  😀

(The Server's RPM build `Makefile` uses the COPR CLI to initiate a build on COPR, so it relies on having this package installed.  To the extent that you've been building the Server RPMs, you've probably already installed this and long since forgotten.)

dbutenhof: The real question is whether we really want/need to prepare for building an RPM inside the container.

However, on reflection, it's not a bad idea since "Mr. Jenkins" eventually ought to be doing that for us automatically as part of the test toolchain.

server/bin/pbench-server-prep-shim-002.py
webbnh: The update to the mock `selinux` module fixed this.  😁

dbutenhof (approved):

npalaska (approved):

PR #2591, opened 2022-01-11
Remove implicit use of "quarantiner" user

dbutenhof:
The original "light dataset" implementation provided for dynamically creating datasets. While most of that was removed, it remained in the bash base utilities `quarantine` function because that could be called from `pbench-server-prep-shim-002` before a dataset was created. However, in normal use the "quarantiner" user doesn't exist anyway.

This removes that code, so that `quarantine` depends on the dataset already existing. This includes an adjustment to test-5.2 (which is the only legacy test case exercising this code through pbench-dispatch) to pre-create the expected datasets. (Using the same "quarantiner" user, so that the gold file doesn't change!) Because we collect the usernames specified in the `.db` file, we no longer need to hardcode `quarantiner` in the list of users to create.

### DIFF
--- a/server/bin/pbench-base.sh
+++ b/server/bin/pbench-base.sh
@@ -137,9 +137,7 @@ function quarantine () {
         fi
         # Create/update state record if we're not moving a .md5 file
         if [ ${afile} != ${afile%.tar.xz} ] ;then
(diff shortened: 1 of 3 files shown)


portante (approved):

webbnh (approved):
👍

Issue #2383, opened 2021-07-22
Address Shell command PATH requirements in shell script Agent tools


webbnh:
User commands which are implemented as Bash scripts, particularly on the Agent side, rely on the shell's `PATH` environment variable in order to locate external commands.  In most cases, the referenced commands are typical user commands which are reachable via a conventional `PATH`, so this is not a problem, and, in fact, it's a _feature_ because it allows the unit test framework to intercept the invocations and redirect them to mock implementations by modifying the `PATH`.

However, in some cases, such as `pbench-register-tools`, the scripts reference system utilities, such as `/usr/sbin/ip`.  This is not a problem, _provided_ that the User has `/usr/sbin` in his/her `PATH`.  Having system directories in the `PATH` is not unusual for sysadmins, power users, and `root`, but it's not the case for more basic users, such as the one used to run the CI, which results in failures.

This leaves us with a choice.  Either we have to place a requirement on the User (and document it!) that s/he include certain directories (such as `/usr/sbin`) in his/her `PATH`, or we have to provide a mechanism which inserts them automatically.  We already have `agent/profile`, which adds the Agent's tool directories to the user's path.  This would be a logical place to add our other `PATH` requirements, such as `/usr/sbin` and any which we develop in the future.

However, just to note, the `agent/profile` is not currently executed in the the unit tests setup.  In part, this is because the unit tests provide their own environment, which includes a custom installation-like setup as well as mocks, so running `agent/profile` as it currently stands would be largely superfluous.  Nevertheless, it has exposed this hidden requirement.  For the moment, this has been addressed by having `tox/run` add `/usr/sbin` to the `PATH`, but it seems like having `agent/profile` provide the proper path (and having the unit tests execute it) prior to setting up the modified path might be a better solution.

Also note that this problem exists only while the Pbench tools are implemented as shell scripts:  once they are all recast into Python, this requirement will go away.

PR #2592, opened 2022-01-14
Make sure /usr/sbin is present in system PATH

npalaska:
Addresses issue https://github.com/distributed-system-analysis/pbench/issues/2383

### DIFF
--- a/agent/profile
+++ b/agent/profile
@@ -3,6 +3,8 @@
 function _pbench_pathins() {
     if [[ -d "${1}" ]] && [[ ":${PATH:-}:" != *":${1}:"* ]]; then
         PATH="${1}${PATH:+:${PATH}}"
+        if [[$PATH != *"/usr/sbin"*]]; then
+            PATH="$PATH:/usr/sbin" # For /usr/sbin/ip, used by pbench-register-tool
     fi
 }
 
--- a/agent/profile
+++ b/agent/profile
@@ -6,12 +6,8 @@ function _pbench_pathins() {
     fi
 }
 
-_pbench_pathins "/usr/local/sbin"
-_pbench_pathins "/usr/local/bin"
 _pbench_pathins "/usr/sbin"
 _pbench_pathins "/usr/bin"
-_pbench_pathins "/root/bin"
-_pbench_pathins "/sbin"
 
 function _pbench_setup() {
     local prefix=/opt/pbench-agent
--- a/agent/profile
+++ b/agent/profile
@@ -3,10 +3,15 @@
 function _pbench_pathins() {
     if [[ -d "${1}" ]] && [[ ":${PATH:-}:" != *":${1}:"* ]]; then
         PATH="${1}${PATH:+:${PATH}}"
-        if [[$PATH != *"/usr/sbin"*]]; then
-            PATH="$PATH:/usr/sbin" # For /usr/sbin/ip, used by pbench-register-tool
     fi
 }
+
+_pbench_pathins "/usr/local/sbin"
+_pbench_pathins "/usr/local/bin"
+_pbench_pathins "/usr/sbin"
+_pbench_pathins "/usr/bin"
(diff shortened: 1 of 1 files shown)


dbutenhof (changes requested):

agent/profile
```diff
@@ -3,6 +3,8 @@
 function _pbench_pathins() {
     if [[ -d "${1}" ]] && [[ ":${PATH:-}:" != *":${1}:"* ]]; then
         PATH="${1}${PATH:+:${PATH}}"
+        if [[$PATH != *"/usr/sbin"*]]; then
```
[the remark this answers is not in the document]
npalaska: > The `_pbench_pathins` function is designed to do exactly what you're adding here: conditionally add a directory that's not already in `PATH`. It's called repeatedly (e.g., line 38/39) so you're unnecessarily duplicating this check; plus doing the check in a manner inconsistent with the existing check at line 4.

Ah yes thanks for pointing out.


> And while `/usr/sbin` happens to satisfy this particular requirement, I wonder if we should have any others, like maybe `/sbin`, which might not also be in a standard `PATH`. I can't think of any really good automated way to figure out what paths we might actually need beyond a normal user `PATH`, but maybe setting `export PATH=/opt/pbench-agent/util-scripts:/opt/pbench-agent/bench-scripts` without anything else and then making sure `pbench-user-benchmark` and `pbench-fio` work would at least be a start.

Yes I did some tests around this like running a pbench-* commands with one of different versions of PATH (removing everything except `/opt/pbench-agent/util-scripts:/opt/pbench-agent/bench-scripts` and then manually adding paths to system PATH) to check which command needs what but then this creates a lot of permutations. I was running it as a root so I expected that default PATH would contain everything we need based on webb's comment on issue 2383. 

So maybe we could do 
_pbench_pathins /usr/sbin
_pbench_pathins /usr/local/sbin
_pbench_pathins /usr/local/bin
_pbench_pathins /usr/sbin
_pbench_pathins /usr/bin
_pbench_pathins /root/bin

to make sure everything is present in system PATH. And I will do some testing with `/sbin`

portante (changes requested):
Like the suggestion of using "pathin" on the list of known paths we need.

agent/profile
```diff
@@ -6,6 +6,13 @@ function _pbench_pathins() {
     fi
 }
 
+_pbench_pathins "/usr/local/sbin"
+_pbench_pathins "/usr/local/bin"
+_pbench_pathins "/usr/sbin"
+_pbench_pathins "/usr/bin"
+_pbench_pathins "/root/bin"
+_pbench_pathins "/sbin"
```
portante, line 14: These last two I am not sure of now that I see them.  I think it opens an unnecessary set of problems.

agent/profile
```diff
@@ -6,6 +6,13 @@ function _pbench_pathins() {
     fi
 }
 
+_pbench_pathins "/usr/local/sbin"
+_pbench_pathins "/usr/local/bin"
```
portante, line 10: I don't these these two should be included.  Does the pbench-agent RPM install anything in these directories?  If not, then we should not include them.  It is up to the administrator of the system to add them to the path explicitly.

portante (changes requested):

agent/profile
npalaska: Yes the pbench-agent rpm doesnt seem to install anything in these 2 directories, I will remove them

agent/profile
npalaska: Yes I did some tests with pbench-agent commands and `/sbin` doesn't seem to be required in the path (but I wasn't fully sure and its possible that I missed something that actually need this directory in the path so I added it thinking we dont want to come across an issue where something is missing from the system PATH)

portante (approved):

webbnh (approved):
👍

dbutenhof (approved):

PR #2594, opened 2022-01-17
Cleanup of agent ansible roles

ndokos:
Two commits:

- the first one deals with a couple of changes that are needed to satisfy Galaxy: some whitespace changes, a missing role README and a `requires_ansible` directive.
- the second one decouples the firewall role from the `pbench_agent_config` role and makes it standalone.

New version of the pbench.agent collection has been uploaded to Galaxy and tested with RHEL9.

### DIFF
--- a/agent/ansible/Makefile
+++ b/agent/ansible/Makefile
@@ -18,6 +18,7 @@ test:   build
 	ansible-galaxy collection install ${tb} -p ./collections
 	git submodule update --init --recursive
 	cd ${colldir}; ansible-test sanity
+	# test for README.md in *every* role
 
 publish: clean build
 	ansible-galaxy collection publish -vvv -s "${apiserver}"  --api-key "${apikey}" ${tb}
--- a/agent/ansible/pbench/agent/roles/epel_repo_install/README.md
+++ b/agent/ansible/pbench/agent/roles/epel_repo_install/README.md
@@ -1,24 +1,16 @@
-pbench_firewall_open_ports
-==========================
+epel_repo_install
+=================
 
-This role punches a hole through the firewall for various ports that the tool meister
-processes need to use.
+This role installs a repo file for EPEL.
 
 Requirements
 ------------
-None.
 
 Role Variables
 --------------
-These are the variables that are defined by default and do not
-generally need to be modified unless there is a conflict:
-
-- pbench_redis_port: 17001
-- pbench_wsgi_port: 8080
 
 Dependencies
 ------------
-None
 
 Example Playbook
 ----------------
@@ -33,8 +25,10 @@ or you can use the raw link to wget/curl the file:
 
 License
 -------
+
 GPL-3.0-or-later.
 
 Author Information
 ------------------
+
 The Pbench team.
--- a/agent/ansible/pbench/agent/galaxy.yml
+++ b/agent/ansible/pbench/agent/galaxy.yml
@@ -9,7 +9,7 @@ namespace: pbench
 name: agent
 
 # The version of the collection. Must be compatible with semantic versioning
-version: 1.0.4
+version: 1.0.5
 
 # The path to the Markdown (.md) readme file. This path is relative to the root of the collection
 readme: README.md
--- a/agent/ansible/pbench/agent/meta/runtime.yml
+++ b/agent/ansible/pbench/agent/meta/runtime.yml
@@ -1,0 +1,2 @@
+---
+requires_ansible: ">=2.5"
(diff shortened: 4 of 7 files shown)


agent/ansible/Makefile
```diff
@@ -18,6 +18,7 @@ test:   build
 	ansible-galaxy collection install ${tb} -p ./collections
 	git submodule update --init --recursive
 	cd ${colldir}; ansible-test sanity
+	# test for README.md in *every* role
```
dbutenhof, line 21: Sounds like a `FIXME`? Is this still necessary?

agent/ansible/pbench/agent/roles/epel_repo_install/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+Example Playbook
+----------------
+An example playbook can be obtained from
+
+    https://github.com/distributed-system-analysis/pbench/blob/master/agent/ansible/pbench_agent_install.yml
+
+or you can use the raw link to wget/curl the file:
+
+    https://raw.githubusercontent.com/distributed-system-analysis/pbench/master/agent/ansible/pbench_agent_install.yml
```
dbutenhof, line 23: That's interesting: I've never seen this `raw.githubusercontent.com` path. What's the difference, and why is it worth including both?

agent/ansible/pbench/agent/roles/epel_repo_install/README.md
```diff
@@ -0,0 +1,34 @@
...
+This role installs a repo file for EPEL.
+
+Requirements
+------------
+
+Role Variables
+--------------
+
+Dependencies
+------------
```
dbutenhof, line 13: Do we really need all this boilerplate if it's just taking up space? Or is there something useful we could put in these sections. The `pbench_firewall_open_ports` version seems like an improvement.

dbutenhof (approved):
OK, but a couple of comments/nits ...

portante (approved):

agent/ansible/Makefile
ndokos: Yes, it is a FIXME. The problem is that Galaxy imposes rules and regulations, but `ansible-test sanity` is unaware of them, so it does not flag them. They are found when the collection is uploaded and Galaxy goes through its steps, and fails (with messages that sometimes require tea leaf reading abilities). So I would like to replace the comment with a real test, but I didn't want to do it while running on empty. I can probably take a stab at implementing the test now, but I don't want to delay merging the PR for this. I'll open a (decidedly minor) issue and fix it later.

agent/ansible/pbench/agent/roles/epel_repo_install/README.md
ndokos: It's where you go if you click on the "Raw" link of a file on Github. CLI addicts tend to prefer that kind of link, so they don't have to deal with HTML junk.

agent/ansible/pbench/agent/roles/epel_repo_install/README.md
ndokos: Some roles use some/all of these (e.g. the firewall role). These files are used by Galaxy to provide its views and given how picky it is, I wasn't about to experiment too much: I don't remember if you can omit them when empty. Having them in the file doesn't bother me much: I find it helpful to just copy a file from one role to another and having all the fields there (misspelling "Requirements" would probably cause a Galaxy conniption and those tend to be dangerous :-)  - have you seen my [wallpaper](https://apod.nasa.gov/apod/ap100507.html)? But I suppose that we can add "None" in all of the empty places. I'll experiment a bit at some point and try to figure out the Galaxy rules.

agent/ansible/Makefile
ndokos: See issue #2595

agent/ansible/pbench/agent/roles/epel_repo_install/README.md
ndokos: See issue #2595

npalaska (approved):

agent/ansible/Makefile
webbnh: This comment is problematic:  it's describing a function or code which _ought_ to be here but isn't.  This would be made clear if it were prefixed with `FIXME` or `TODO`, but I believe we have a preference for removing those prior to merging.  So, if we're going to retain this comment in the code, it really should be enhanced (e.g., replace "test" with "This code *should* be testing").

webbnh (approved):
Looks OK to me (not that you needed yet another approval...), but I added a comment to Dave's (resolved) "conversation" about the "fix me" comment in `agent/ansible/Makefile`.

PR #2604, opened 2022-01-19
Adjust dependencies

dbutenhof:
I've run into a weird problem in a VM deployment showing what appears to be a circular `import` dependency. The `filetree` module imports `JSON` and `convert_date` from `pbench.server.api.resources`, and (judging from the stack trace) Python is somehow importing `pbench.server.api`, which imports `datasets_delete`, which imports `filetree`. I don't understand the linkage from `pbench.server.api.resources` to `bench.server.api` that seems to be represented in the stack trace; however in thinking about this I decided that I'd prefer to push the `JSON` type definitions outside the "API resources" tree anyway, so I'm experimentally pushing this PR as a cleanup.

Along the way, I also replaced the reference to `convert_date` with a direct call to `dateutil.parser.parse` (which is basically what `convert_date` does anyway, without requiring a reference to another Pbench module).

Both are minor changes, although the first touches a fair number of files.

These are the only "functional" changes in this PR, although I also tried to improve the sorting of import statements in a few of the modules, where I happened to notice bad ordering.

### DIFF
--- a/lib/pbench/server/__init__.py
+++ b/lib/pbench/server/__init__.py
@@ -6,9 +6,20 @@ from pathlib import Path
 from datetime import datetime, tzinfo, timedelta
 from time import time as _time
 from configparser import NoSectionError, NoOptionError
+from typing import Dict, List, Union
 
 from pbench import PbenchConfig, _STD_DATETIME_FMT
 from pbench.common.exceptions import BadConfig
+
+
+# A type defined to conform to the semantic definition of a JSON structure
+# with Python syntax.
+JSONSTRING = str
+JSONNUMBER = Union[int, float]
+JSONVALUE = Union["JSONOBJECT", "JSONARRAY", JSONSTRING, JSONNUMBER, bool, None]
+JSONARRAY = List[JSONVALUE]
+JSONOBJECT = Dict[JSONSTRING, JSONVALUE]
+JSON = JSONVALUE
 
 
 class simple_utc(tzinfo):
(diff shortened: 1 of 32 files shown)


webbnh (dismissed):
Good enough.

(I am not going to point out that `requests` and `responses` are third-party libraries.)

webbnh (approved):
> I've never liked the idea of trying to separate the standard vs third party imports anyway; some are "obviously" third party, but I'll probably never have my head stuffed full of a complete list of standard package names.

I was solidly in your camp...and, then I spent a bunch of time with `requirements.txt` and the RPM `.spec` files, and now I'm inclined to see things somewhat differently....  🦐

(Nevertheless, I'm certainly not going to point out that the arrangement in `lib/pbench/test/unit/server/query_apis/conftest.py` is wrong (:wink:), so don't change it unless you find something else to change, first.)

dbutenhof:
> > I've never liked the idea of trying to separate the standard vs third party imports anyway; some are "obviously" third party, but I'll probably never have my head stuffed full of a complete list of standard package names.
> 
> I was solidly in your camp...and, then I spent a bunch of time with `requirements.txt` and the RPM `.spec` files, and now I'm inclined to see things somewhat differently.... shrimp

Unless you know of a reliable automatic way to extract the `requirements.txt` list from our `.py` files that's unavoidably dependent on the *order* of our `import` statements, I'd say that configuration management issue is pretty much a red herring in this context. 🍤 

> (Nevertheless, I'm certainly not going to point out that the arrangement in `lib/pbench/test/unit/server/query_apis/conftest.py` is wrong (wink), so don't change it unless you find something else to change, first.)

And, luckily, I didn't *read* the comment that you didn't make, so we're all good here. 🥇

npalaska (approved):

Issue #2603, opened 2022-01-18
When no tools are registered, TM should not try to connect to the Redis server


npalaska:
When there are no tools registered on the machine, the tool meister still tries to connect to the redis server (which wouldn't have any active channels because no tools are registered). This creates a loop where the TM continuously tries to connect to redis at a regular interval. This results in the following error:

```
Traceback (most recent call last):
  File "/opt/pbench-agent/lib/python3.6/site-packages/redis/connection.py", line 607, in connect
    sock = self._connect()
  File "/opt/pbench-agent/lib/python3.6/site-packages/redis/connection.py", line 671, in _connect
    raise err
  File "/opt/pbench-agent/lib/python3.6/site-packages/redis/connection.py", line 659, in _connect
    sock.connect(socket_address)
ConnectionRefusedError: [Errno 111] Connection refused

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/pbench-agent/util-scripts/pbench-tool-meister-client", line 92, in <module>
    status = main(sys.argv)
  File "/opt/pbench-agent/util-scripts/pbench-tool-meister-client", line 84, in main
    logger=logger,
  File "/opt/pbench-agent/lib/pbench/agent/tool_meister_client.py", line 121, in __enter__
    self.redis_server, self.to_client_channel
  File "/opt/pbench-agent/lib/pbench/agent/redis.py", line 50, in __init__
    self._pubsub.subscribe(channel_name)
  File "/opt/pbench-agent/lib/python3.6/site-packages/redis/client.py", line 1527, in subscribe
    ret_val = self.execute_command("SUBSCRIBE", *new_channels.keys())
  File "/opt/pbench-agent/lib/python3.6/site-packages/redis/client.py", line 1369, in execute_command
    "pubsub", self.shard_hint
  File "/opt/pbench-agent/lib/python3.6/site-packages/redis/connection.py", line 1311, in get_connection
    connection.connect()
  File "/opt/pbench-agent/lib/python3.6/site-packages/redis/connection.py", line 611, in connect
    raise ConnectionError(self._error_message(e))
redis.exceptions.ConnectionError: Error 111 connecting to localhost:17001. Connection refused.
test sample 1 of 5
client[127.0.0.1]protocol[tcp]test[stream]instances[2]size[131072] <-> server[127.0.0.1]
```
steps to reproduce: 
```
pbench-clear-tools
[...]

Issue #2433, opened 2021-08-26
pbench-agent rpm does not have dependency on pbench-sysstat


aokuliar-rh:
pbench-agent-0.69.9-1ga12a9aa0f 

After installation of `pbench-agent-0.69.9-1ga12a9aa0f`  rpm `pbench-sysstat` is not installed as a dependency. I assume that this is a bug, because pbench-agent cannot be reasonably used without `pbench-sysstat`.

portante:
This problem is tied up with the default tool set.  Ideally the default tool set would not include any `pbench-sysstat` tools so that we don't have to have a dependency on a particular RPM for tools.

But that might not be reasonable.

We really want to be able to use the `pcp` tool by default, which would be another dependency.  However, if we have the right messaging, we should properly warn the user when no tools are found that were registered.

ndokos:
Although there are deeper issues as @portante mentions, we are not going to impose a dependency of the `pbench-agent` RPM  on `pbench-sysstat`: currently, the problem is "solved" by having the Ansible playbook install *both* the `pbench-agent` RPM and the `pbench-sysstat` RPM, but of course, that does not work if the Ansible playbook is *not* used. The ultimate resolution is that we are going to get rid of `pbench-sysstat`, but for the time being, the user is expected to either install using Ansible or install `pbench-sysstat` manually.

Closing this, but feel free to reopen if you object.

PR #2605, opened 2022-01-19
Add a warning message when no tools are registered

npalaska:
When no tools are registered on the host, we can at least put a warning about it for the user to consider.

### DIFF
--- a/agent/base
+++ b/agent/base
@@ -209,8 +209,8 @@ function verify_tool_group {
         printf -- "\t${script_name}: invalid ${param_name} option (\"${group}\"), directory not found: ${tools_group_dir}\n" >&2
         return 1
     fi
-    if [[ -z "$(ls -A ${tools_group_dir})" ]]; then
-        printf -- "\tWARNING: No tools are registered\n" >&2
+    if [[ -z "$(ls ${tools_group_dir})" ]]; then
+        printf -- "${script_name}: WARNING: No tools are registered\n" >&2
     fi
     printf -- "${tools_group_dir}"
     return 0
--- a/agent/base
+++ b/agent/base
@@ -209,6 +209,9 @@ function verify_tool_group {
         printf -- "\t${script_name}: invalid ${param_name} option (\"${group}\"), directory not found: ${tools_group_dir}\n" >&2
         return 1
     fi
+    if [[ -z "$(ls -A ${tools_group_dir})" ]]; then
+        printf -- "\tWARNING: No tools are registered\n" >&2
+    fi
     printf -- "${tools_group_dir}"
     return 0
 }


agent/base
```diff
@@ -209,6 +209,9 @@ function verify_tool_group {
         printf -- "\t${script_name}: invalid ${param_name} option (\"${group}\"), directory not found: ${tools_group_dir}\n" >&2
         return 1
     fi
+    if [[ -z "$(ls -A ${tools_group_dir})" ]]; then
+        printf -- "\tWARNING: No tools are registered\n" >&2
```
dbutenhof, line 213: The similar `printf` at 209 uses `${script_name}:` and doesn't have an explicit severity. Including the script name seems to be a common convention that might be worth following.

Additionally, is a simple warning `printf` that continues operation sufficient here? It may give an indirect clue of why something bad happens later, but won't serve to prevent it.

npalaska: > The similar `printf` at 209 uses `${script_name}:` and doesn't have an explicit severity. Including the script name seems to be a common convention that might be worth following.

Yes I should have added `${script_name}:` here as well.

> Additionally, is a simple warning `printf` that continues operation sufficient here? It may give an indirect clue of why something bad happens later, but won't serve to prevent it.

We should't exit if no tools are registered, it is a valid use case to run user benchmark with no tools registered. However, I felt it ought to give some indication or warning when starting the benchmark that there are currently no tools registered on this host (and if something bad happens this may be the reason although its not in the message and I dont know how I should add that message). Also if you look at the comment from peter on the issue https://github.com/distributed-system-analysis/pbench/issues/2433 he seem to suggest that we should warn the user when no tools are found. 

Just a warning is definitely not sufficient and we need to handle exceptions more robustly when no tools are registered but that's another issue.

webbnh: The `-A` switch is an interesting choice:  if someone puts a "dot file" in the tools directory, that will silence this warning without "being a tool", right?  Is that a bug or a feature?

Not knowing the history or culture behind this setup, I would be inclined to omit the `-A` switch an let "dot files" stay hidden and ignored.

As for the greater question, I think we should allow users to run workloads without data collection tools (i.e., the workload will probably report its own results), but we _should_ warn them, because running a benchmark might be an expensive operation and we don't want them to unknowingly fail to collect all of their data.

webbnh (approved):
👍

agent/base
dbutenhof: > Just a warning is definitely not sufficient and we need to handle exceptions more robustly when no tools are registered but that's another issue.

Fine; I was reading this in the context of your recent #2603 ... however you're both right we don't necessarily want to prevent using a benchmark script without explicit tools and a warning is likely appropriate. I'd like to see the message made more consistent, and I suspect Webb's comment about dropping the `-A` is a good one.

dbutenhof (approved):

webbnh (approved):
👍

Issue #2473, opened 2021-10-04
pbench-list-tools should be consistent with pbench-register-tools


webbnh:
it seems to be the case that despite the fact that `pbench-register-tools` is registering in the new `pbench_run` dir, `pbench-list-tools` is still sticking to the default dir, `/var/lib/pbench-agent`. Even testing it seperately, `pbench-list-tools` does not change behavior when `$pbench_run` is defined/changed. I am testing this with a 0.71.0 build that's a few months old. Not sure if it has been addressed since

_Originally posted by @Maxusmusti in https://github.com/distributed-system-analysis/pbench/pull/2332#discussion_r721064474_

PR #2606, opened 2022-01-20
pbench-list-tools and pbench-register-tools inconsistency with env and config values

riya-17:
Fixes Issue #2473 

Checking and adding environment value before config file value for pbench-list-tools.

If an environment value for `pbench_run` is present then `pbench-register tool` considers env value and doesn't care about config file value, `pbench-list-tool` on-the-other hand only considers environment value of `pbench_run` if `pbench_run`'s value is None and take config file's value as default.

### DIFF
--- a/lib/pbench/agent/base.py
+++ b/lib/pbench/agent/base.py
@@ -21,12 +21,17 @@ class BaseCommand(metaclass=abc.ABCMeta):
         self.config = PbenchAgentConfig(self.context.config)
         self.name = os.path.basename(sys.argv[0])
 
-        self.pbench_run = self.config.pbench_run
+        env_pbench_run = os.environ.get("pbench_run")
+        self.pbench_run = pathlib.Path(
+            env_pbench_run if env_pbench_run else self.config.pbench_run
+        )
+
         if not self.pbench_run.exists():
-            click.secho(
-                f"[ERROR] the provided pbench run directory, {self.pbench_run}, does not exist"
+            click.echo(
+                f"[ERROR] pbench run directory does not exist, {self.pbench_run}",
+                err=True,
             )
-            sys.exit(1)
+            click.get_current_context().exit(1)
 
         # the pbench temporary directory is always relative to the $pbench_run
         # directory
--- a/lib/pbench/test/functional/agent/cli/commands/conftest.py
+++ b/lib/pbench/test/functional/agent/cli/commands/conftest.py
@@ -22,15 +22,6 @@ def pbench_cfg(tmp_path, opt_pbench):
 
 
 @pytest.fixture
-def pbench_run(monkeypatch, tmp_path):
-    monkeypatch.delenv("pbench_run")
-    pbench_agent_d = tmp_path / "var" / "lib" / "pbench-agent"
-    pbench_agent_d.mkdir(parents=True, exist_ok=True)
-
-    yield pbench_agent_d
-
-
-@pytest.fixture
 def agent_config(monkeypatch, tmp_path, opt_pbench, pbench_cfg, pbench_run):
     shutil.copyfile("./agent/config/pbench-agent-default.cfg", pbench_cfg)
 
--- a/lib/pbench/test/functional/agent/cli/commands/conftest.py
+++ b/lib/pbench/test/functional/agent/cli/commands/conftest.py
@@ -22,7 +22,8 @@ def pbench_cfg(tmp_path, opt_pbench):
 
 
 @pytest.fixture
-def pbench_run(tmp_path):
+def pbench_run(monkeypatch, tmp_path):
+    monkeypatch.delenv("pbench_run")
     pbench_agent_d = tmp_path / "var" / "lib" / "pbench-agent"
     pbench_agent_d.mkdir(parents=True, exist_ok=True)
 
--- a/lib/pbench/test/functional/agent/cli/commands/test_config.py
+++ b/lib/pbench/test/functional/agent/cli/commands/test_config.py
@@ -1,0 +1,40 @@
+import os
+import pytest
+
+
+def test_pbench_run(monkeypatch, tmpdir, agent_config, pbench_run):  # noqa F811
+    """Use pbench_run value from pbench-agent.cfg file"""
+    assert os.environ.get("pbench_run") is None
+    command = ["pbench-list-tools"]
+    out, err, exitcode = pytest.helpers.capture(command)
+    assert exitcode == 0
+
+
+def test_pbench_run_dir_existence(monkeypatch, agent_config, pbench_run):
+    """Verify that pbench_list_tools fail when pbench_run is defined
+    to a directory that doesn't exist"""
+    monkeypatch.setenv("pbench_run", f"{pbench_run}/test")
+    command = ["pbench-list-tools"]
+    out, err, exitcode = pytest.helpers.capture(command)
+    assert exitcode == 1
+    assert b"" == out
+    assert b"[ERROR] pbench run directory does not exist," in err
+
+
+def test_pbench_dir_existence(monkeypatch, pbench_run, agent_config):
+    """Verify the existence of existing pbench_run directory
+    and successful completion of command"""
+    pbench_run_d = pbench_run / "test_dir"
+    pbench_run_d.mkdir(parents=True, exist_ok=True)
+    monkeypatch.setenv("pbench_run", str(pbench_run_d))
+    p = pbench_run_d / "tools-v1-default" / "th1.example.com"
+    p.mkdir(parents=True)
+    tool = p / "iostat"
+    tool.write_text("--interval=30")
+    tool = p / "mpstat"
+    tool.write_text("--interval=300")
+    command = ["pbench-list-tools"]
+    out, err, exitcode = pytest.helpers.capture(command)
+    assert exitcode == 0
+    assert b"group: default; host: th1.example.com; tools: iostat, mpstat\n" in out
+    assert b"" == err
--- a/lib/pbench/test/functional/conftest.py
+++ b/lib/pbench/test/functional/conftest.py
@@ -1,7 +1,5 @@
-import os
-import subprocess
-
 import pytest
+import subprocess
 
 from pbench.test.unit.agent.conftest import base_setup
 
@@ -19,7 +17,9 @@ def capture(command):
 
 
 @pytest.fixture
-def pbench_run(tmpdir):
-    pbench_run = tmpdir / "test/var/lib/pbench-agent"
-    os.makedirs(pbench_run)
-    yield pbench_run
+def pbench_run(monkeypatch, tmp_path):
+    monkeypatch.delenv("pbench_run")
+    pbench_agent_d = tmp_path / "var" / "lib" / "pbench-agent"
+    pbench_agent_d.mkdir(parents=True, exist_ok=True)
+
+    yield pbench_agent_d


lib/pbench/agent/base.py
```diff
@@ -21,7 +21,10 @@ def __init__(self, context):
         self.config = PbenchAgentConfig(self.context.config)
         self.name = os.path.basename(sys.argv[0])
 
-        self.pbench_run = self.config.pbench_run
+        env_pbench_run = os.environ.get("pbench_run")
+        self.pbench_run = pathlib.Path(
+            env_pbench_run if env_pbench_run else self.config.pbench_run
+        )
```
dbutenhof, line 27: This mostly matches the code in `agent/base`, which is used by `pbench-register-tool`. The only real difference is that `base`, interestingly, verifies existence of the directory only if it was provided by a `pbench_run` environment variable.

My concern is exactly where we expect the default `/var/lib/pbench-agent/` to be created. The Python code expects it to already exist, but the following bash `base` statement is a `mkdir -p "${pbench_run}/tmp" which would create it if it doesn't already exist ... the Python code will instead fail on the following `if not self.pbench_run.exists()`.

Then again, since this is mostly pre-existing code I assume we'd have run into this problem already if there were realistic cases where `/var/lib/pbench-agent` *didn't* exist, and it probably doesn't matter.

Adding a click CLI unit test case for your new code, however, would be a good idea.

riya-17: So, what should we do? Should I create the directory if it is not present? or let it throw error? 

if [[ -z "$pbench_run" ]]; then                                                                                                       1   ┊ pbench_run=$(pbench-config pbench_run pbench-agent)
  2   ┊ if [[ -z "$pbench_run" ]]; then
  3   ┊ ┊ ┊ pbench_run=/var/lib/pbench-agent
  4   ┊ fi
  5   ┊ export pbench_run
  6 else
  7   ┊ if [[ ! -d $pbench_run ]]; then
  8   ┊ ┊ ┊ echo "[ERROR] the provided pbench run directory, ${pbench_run}, does not exist" >&2
  9   ┊ ┊ ┊ exit 1
 10   ┊ fi
 11 fi


In `agent/base` it does check if the directory exists and if it doesn't then it exits and doesn't continue.

dbutenhof: Yes; the difference is that the bash version only exits with error if `pbench_run` is defined to a non-blank value and the directory doesn't exist, whereas the Python makes the check even if `pbench_run` wasn't defined.

Whether this matters depends on whether there's supposed to be a guarantee that the `pbench-agent.cfg` directory already exists; and whether the bash `mkdir -p` that immediately follows is a deliberate attempt to dynamically create a "default" `pbench_run` directory or an accident. (That is, `mkdir ${pbench_run}/tmp` would have worked just fine if we were sure that `${pbench_run}` already existed, but the `-p` will create the entire path if necessary.)

Also, if this is a bug, it wasn't added by your changes, so the issue is somewhat separate from your PR even though it's adjacent code. (You added the environment variable lookup, but the config file default path isn't affected.) Maybe @portante or @ndokos can clarify the historical intent without requiring someone to search through the code to figure out whether a missing default `pbench_run` directory at this point ought to be an error.

dbutenhof (changes requested):
Your change also broke a bunch of the click unit tests, so you have some work there in addition to making sure there's specific test coverage of your change.

lib/pbench/agent/base.py
portante: First, I believe the Bash code is wrong, it should not issue an `mkdir -p ${pbench_run}/tmp`.

The check for existence of `${pbench_run}` should be removed entirely, and a `mkdir ${pbench_run}/tmp` should be used, where on failure we report why it failed.

That is probably best addressed in a separate PR.

Once that is addressed, then I believe we can resolve cleanly the differences between the Python3 code and the Bash code.

webbnh: > That is probably best addressed in a separate PR.

By which, @portante, means please go do that one first, and then come back and build this PR on that one.

dbutenhof: Under the circumstances, we definitely need an issue/Jira to track the correct behavior and get the bash and Python versions completely aligned... however I think it's also perfectly reasonable to add the `_pbench_run` check here and *then* fix the check/create paths in both afterwards. (This change doesn't make the disconnect worse.)

webbnh: > we definitely need an issue/Jira to track the correct behavior and get the bash and Python versions completely aligned...

Not to go out on a tangent or anything, but I believe the plan is to remove the Bash code and replace it with Python, at which point they presumably will be completely aligned (and there _is_ a Jira ticket (an epic, I think) for that...).  🙂

dbutenhof: Tangent? Us? When have we *ever* gone on a tangent about any discussion. Name one example. 👀 OK, name 1,000 examples... 

But yes, that's certainly true; we also explicitly kicked the remaining "Pythonization" out of sprint 8 and Agent 0.71 so we want to do what we can prior to that.

I fussed just a bit yesterday afternoon over how to organize something like an epic, perhaps, for "0.71 completers" which might or might not include defects... but didn't really strike any inspiration.

riya-17:
Do I need to add tests for this PR? As it is already using the `Env` value of `pbench_run` for all tests.

webbnh:
> Do I need to add tests for this PR? As it is already using the `Env` value of `pbench_run` for all tests.

Just looking quickly at the lines that you've "added" in the diff, in conjunction with tests which "operate normally", I'd say that we should have tests which cover the following:

- [ ] `pbench_run` is defined by an environment variable
- [ ] `pbench_run` is defined by the configuration file
- [ ] `pbench_run` refers to a directory which does not exist

If we already have tests which exercise those paths, then, no, you don't need to add any new tests.

However, I suspect that all of our tests run the same way, and so perhaps we cover only the first one...I don't know -- you'd have to check.

riya-17:
For the first test, I tried putting `None` as value for the `pbench_run` environment variable and then letting the `base.py` script decide further, but `monkeypatch.setenv` is unable to set the `None` value as constant for `pbench_run` but as a string value.

dbutenhof:
> For the first test, I tried putting `None` as value for the `pbench_run` environment variable and then letting the `base.py` script decide further, but `monkeypatch.setenv` is unable to set the `None` value as constant for `pbench_run` but as a string value.

Try `monkeypatch.delenv("pbench_run")` instead.

lib/pbench/test/functional/agent/cli/commands/test_config.py
```diff
@@ -14,6 +15,7 @@ def test_pbench_run(monkeypatch, tmpdir, agent_config, pbench_run):  # noqa F811
 
 
 def test_pbench_run_dir_existence(monkeypatch, agent_config, pbench_run):
+    """Verify the existence of pbench_run directory"""
```
dbutenhof, line 18: We're always expecting the command to "verify the existence" ... I think the important distinction of this test is that we've created our own non-standard `pbench_run` environment variable and we expect a failure because it doesn't exist.

That is, I think the docstring here is a bit too generic; but the test looks good.

lib/pbench/test/functional/agent/cli/commands/test_config.py
```diff
@@ -4,6 +4,7 @@
 
 
 def test_pbench_run(monkeypatch, tmpdir, agent_config, pbench_run):  # noqa F811
+    """Use pbench_run value from pbench-config"""
```
dbutenhof, line 7: Maybe "from the `pbench-agent.cfg` file? (Or even just "from the config file"?) In this context we're not using the `pbench-config` command so this could be a bit confusing.

dbutenhof (dismissed):
Minor comments on the docstrings, which could be improved; but it's looking good.

lib/pbench/agent/base.py
```diff
@@ -21,10 +21,14 @@ def __init__(self, context):
         self.config = PbenchAgentConfig(self.context.config)
         self.name = os.path.basename(sys.argv[0])
 
-        self.pbench_run = self.config.pbench_run
+        env_pbench_run = os.environ.get("pbench_run")
+        self.pbench_run = pathlib.Path(
+            self.config.pbench_run if env_pbench_run is None else env_pbench_run
+        )
```
webbnh, line 27: Rather than testing to see if `env_pbench_run` is `None` (which would mean that it was not defined), I suggest testing it for "truthiness", e.g.:
```
        self.pbench_run = pathlib.Path(
            env_pbench_run if env_pbench_run else self.config.pbench_run
        )
```
In this way, if it exists but has a null or empty value, we will also ignore it.

lib/pbench/agent/base.py
```diff
@@ -21,10 +21,14 @@ def __init__(self, context):
...
+
         if not self.pbench_run.exists():
             click.secho(
-                f"[ERROR] the provided pbench run directory, {self.pbench_run}, does not exist"
+                f"[ERROR] pbench run directory does not exist, {self.pbench_run}"
             )
```
webbnh, line 32: Since you're not providing a "style", I suggest using [`echo()`](https://click.palletsprojects.com/en/8.0.x/api/#click.echo) instead of `secho()`.  Also, since this is intended to be an error message, I suggest sending it to `stderr` instead of the default `stdout` by providing the `err=true` parameter.)

lib/pbench/test/functional/agent/cli/commands/test_config.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def test_pbench_run(monkeypatch, tmpdir, agent_config, pbench_run):  # noqa F811
+    """Use pbench_run value from pbench-agent.cfg file"""
+    monkeypatch.delenv("pbench_run")
+    env_pbench_run = os.environ.get("pbench_run")
+    command = ["pbench-list-tools"]
+    out, err, exitcode = pytest.helpers.capture(command)
+    assert exitcode == 0
+    assert env_pbench_run is None
```
webbnh, line 13: This is fine as-is, but I would have put the `env_pbench_run` assertion immediately after its assignment:  we want to ensure that the _input_ to the command is an undefined value for `pbench_run`; whereas, with this arrangement what we are checking is that the value is undefined when the command _returns_.  (I suppose, for complete paranoia, we should do the checks in _both_ places....)

lib/pbench/test/functional/agent/cli/commands/test_config.py
```diff
@@ -0,0 +1,25 @@
...
+
+def test_pbench_run_dir_existence(monkeypatch, agent_config, pbench_run):
+    """Verify the existence of a non-existing pbench_run directory
+    which results in expected failure of the test"""
+    monkeypatch.setenv("pbench_run", f"{pbench_run}/test")
+    command = ["pbench-list-tools"]
+    out, err, exitcode = pytest.helpers.capture(command)
+    assert exitcode == 1
+    assert b"[ERROR] pbench run directory does not exist," in out
+    assert b"" == err
```
webbnh, line 25: It strikes me that we should have a third test which is just like these two tests combined:  it should pull in the `pbench_run` fixture and then use `monkeypatch` to override the environment variable, setting to a _valid_ directory *other than* the one from `pbench_run` (e.g., doing what `test_pbench_run_dir_existence` does but then _creating_ it), and then make sure that value of `BaseCommand.pbench_run` ends up matching the test's choice and not the config file's choice...although, the exact details of how to do that escape me....  😈

lib/pbench/agent/base.py
```diff
@@ -21,10 +21,14 @@ def __init__(self, context):
...
             sys.exit(1)
```
webbnh, line 33: Oh, and, as long as you're "in the area", this should be invoking `click.get_current_context().exit(1)` instead of `sys.exit(1)`.

webbnh (approved):
It looks good, although, if you're amenable, I have a couple of suggestions for the code and a couple for the tests.

lib/pbench/test/functional/agent/cli/commands/test_config.py
riya-17: The input to the command is a `pbench_run` value from `config` file. As the value for environment value is None it seeks the value from the config file and works with that and as it has been able to use the `pbench_run` value from the config file the command is running without any errors.

I must be unable to understand the context here

dbutenhof: The `assert` is at the end, but it's based on the environment value captured at the *beginning*, so I don't really understand Webb's take on this. Maybe it would be a bit more clear if the `assert` was up front. Then again ... ultimately this really is a test of the `monkeypatch.delenv`; because failure of that call is the only way the `assert` would fail. Generally, we focus on testing our code, not the testing infrastructure, so this is perhaps excessively paranoid.

I'm not sure if we want to test whether `pbench-list-tools` *creates* a `"pbench_run"` environment variable ... or whether there's any reason to care.

lib/pbench/test/functional/agent/cli/commands/test_config.py
```diff
@@ -0,0 +1,40 @@
...
+
+def test_pbench_dir_existence(monkeypatch, pbench_run, agent_config):
+    """Verify the existence of existing pbench_run directory
+    and successful completion of command"""
+    pbench_run_d = pbench_run / "test_dir"
+    pbench_run_d.mkdir(parents=True, exist_ok=True)
+    monkeypatch.setenv("pbench_run", str(pbench_run_d))
+    command = ["pbench-list-tools"]
+    out, err, exitcode = pytest.helpers.capture(command)
+    pbench_run_val = BaseCommand.pbench_run
```
npalaska, line 38: BaseCommand doesn't have a class attribute `pbench_run`, its an object property. You may have to create BaseCommand object to refer to the pbench_run attribute

npalaska (commented):
Looks good, just 1 comment about referencing a class object attribute. And thats what the jenkins is also complaining.

lib/pbench/test/functional/agent/cli/commands/test_config.py
riya-17: That is an abstract class due to which direct call or object creation is not possible. Calling it indirectly also require `context` will look for some other way

lib/pbench/test/functional/agent/cli/commands/test_config.py
riya-17: Hey, @webbnh can you take a look at the implemented 3rd test. I believe there is some requirement for `Context` value, which I am feeling would be pretty difficult to retrieve or duplicate. Is it the right direction I am moving in?

lib/pbench/test/functional/agent/cli/commands/test_config.py
dbutenhof: Yeah, this is a consequence of the way we run our `click` commands in test: the `capture` helper fires up the whole command in a `Popen` subprocess and captures the results... so we have no access to any of the internal command context to check. We can only rely on `stderr`, `stdout`, `proc.returncode`, and the state of the `tmp` filesystem tree. 

So in this test we can only verify that we saw the tools we expected and whether the command succeeded or failed. We have to pretty much trust that it connected inputs and outputs the way we hoped.

That's what makes this a "functional" test, but like the legacy gold tests it's a hacked environment so it's neither a unit test nor a true functional test. I'd love to get rid of this whole category: both for agent and server. ☹️

lib/pbench/test/functional/agent/cli/commands/test_config.py
webbnh: > The input to the command is a `pbench_run` value from `config` file. As the value for environment value is `None` it seeks the value from the config file and works with that and as it has been able to use the `pbench_run` value from the config file the command is running without any errors.

My apologies, @riya-17.  Let me expand.

The code-under-test checks the environment as well as the configuration file, so, in that sense, there are two inputs which the unit test must control.  That is, this test should ensure that the environment variable is not defined, in order to force the CUT to obtain its value from the configuration file, and then it should assert that the value that the CUT used matches the one in the configuration file.

In total, the unit tests should drive the CUT through all three behaviors -- using the value from the `pbench_run` environment variable, using the value from the configuration file, and using the fall-back value -- and it should set up the pre-conditions for each test so that it can detect whether the CUT is producing the desired result.

lib/pbench/test/functional/agent/cli/commands/test_config.py
webbnh: As Dave points out, the test style in use here is a "functional test" model, where we actually invoke a real tool and try to make sure that it functions correctly under various circumstances.

The "unit test" model would be to invoke only the code which we want to test, here that would be `BaseCommand.__init__()`.  We _could_ do that directly, but that's a bit complicated, since it's an object constructor.  Typically, we would test those by instantiating the object, but this is an abstract base class, so we cannot create an instance of it.  So, to address that, we would have the unit test code declare a concrete subclass of it and instantiate that.

The challenge is that the unit test will need to come up with a Click context object to pass to the instantiation...but maybe that's not too hard.  This being Python, I think that all we need is a generic object which has a field named `config` with an appropriate value in it.  So, maybe something like this would work:
```
    class TestCommand(BaseCommand):
        class FakeContext:
            config = "/path/to/config_file"

        def __init__(self):
            super(TestCommand, self).__init__(FakeContext())

    pbench_run_val = TestCommand().pbench_run
```

If that works well, you might consider using it in the other tests to make them more unit-ish and less functinal-ish.  (To make them properly unit-ish, you'd probably want to mock `PbenchAgentConfig.__init__()` and maybe a few other things, but this would be a good start.)

lib/pbench/test/functional/agent/cli/commands/test_config.py
webbnh: See my [comment](https://github.com/distributed-system-analysis/pbench/pull/2606#discussion_r803063752) below.

lib/pbench/test/functional/agent/cli/commands/test_config.py
dbutenhof: I think I'd rather not put "starting to move towards unit testing" under *this* PR, and simply conform to the current pseudo-functional testing style for now. (For one thing, let's not contemplate moving these from `test/functional` to `test/unit`.)

That means we can't measure internal aspects of the environment that aren't externally observable. That's OK; at this level, if there's no externally observable effect then it simply doesn't matter.

(However, it ought to be observable: if we re-direct `pbench_run` we shouldn't find the tools registered under the (temp) `/var/lib/pbench_agent` and should instead find tools registered under our `pbench_run` directory. Right?)

lib/pbench/test/functional/agent/cli/commands/test_config.py
```diff
@@ -2,15 +2,17 @@
...
 def test_pbench_run(monkeypatch, tmpdir, agent_config, pbench_run):  # noqa F811
     """Use pbench_run value from pbench-agent.cfg file"""
     monkeypatch.delenv("pbench_run")
     env_pbench_run = os.environ.get("pbench_run")
+    assert env_pbench_run is None
     command = ["pbench-list-tools"]
     out, err, exitcode = pytest.helpers.capture(command)
     assert exitcode == 0
-    assert env_pbench_run is None
     assert pathlib.Path(f"{tmpdir}/var/lib/pbench-agent") == pbench_run
```
webbnh, line 16: Does this assertion (at line 16) actually test anything relevant to this test?  It appears to be checking that the value the `pbench_run` _fixture_ is what we expect it to be...which has nothing to do with the execution of `pbench-list-tools` nor with the changes made in this PR.  (This is the problem of _functional_ testing.)

If you're able to set this up as a unit test (see earlier [comment](https://github.com/distributed-system-analysis/pbench/pull/2606#discussion_r803063752)), then I think you can make this assertion test something meaningful.

lib/pbench/test/functional/agent/cli/commands/test_config.py
webbnh: By the way, in case it's not obvious...if you switch to a unit testing model (where you get the value of `pbench_run` directly from your own instantiation of a subclass of `BaseCommand`), then you don't need to invoke `capture()` or to run `pbench-list-tools` at all, because you'll be exercising the CUT directly.

lib/pbench/test/functional/agent/cli/commands/conftest.py
```diff
@@ -22,9 +22,10 @@ def pbench_cfg(tmp_path, opt_pbench):
 
 
 @pytest.fixture
-def pbench_run(tmp_path):
+def pbench_run(monkeypatch, tmp_path):
     pbench_agent_d = tmp_path / "var" / "lib" / "pbench-agent"
     pbench_agent_d.mkdir(parents=True, exist_ok=True)
+    monkeypatch.setenv("pbench_run", str(pbench_agent_d))
```
webbnh, line 28: This change may have wider effects than we really intend.  If the "normal" operation of our code is to pull the `pbench_run` value from the configuration file, then this change will prevent the unit tests from ever testing that configuration, because it will cause the code to almost always get the value from the environment variable.  Now, perhaps that doesn't matter (because your change will ensure that we test it at least once in each configuration), but my instinct is that the default testing configuration should match our expectation for the default operational configuration, and so we should be setting up the `pbench_run` environment variable only in the few tests which want to check that it works.

Thoughts anyone?

webbnh (approved):
I think this will be good to go once we get a few last details sorted.

The new test case will only work if we can convert it from a "functional test" style to more of a "unit test" style.  And, if we can nail that conversion, I would like to see it applied to the other two new tests, too.

The other detail to consider is whether we want to be defining a `pbench_run` environment variable for _all_ tests which use the `pbench_run` fixture, or whether we should move that step out of the fixture and into only the tests which want to test the environment variable.

So, other than those items, once you get the new test working, this should be GTG.

lib/pbench/test/functional/agent/cli/commands/test_config.py
webbnh: > I think I'd rather not put "starting to move towards unit testing" under this PR, and simply conform to the current pseudo-functional testing style for now. (For one thing, let's not contemplate moving these from test/functional to test/unit.)

Well, given that these tests are brand new code, they don't have to be put "here", they could be put wherever its appropriate (from the perspective of the `main` branch, that's a "different addition", not a "move").

Nevertheless, I don't disagree that it would be best to have them conform to the approach in the surrounding code/modules...so, if they are going to stay in the functional area, it would probably be best if they stayed "functional".  And, toward that end, I concur that their results should be observable.  So, rather than running (just) `pbench-list-tools` we should have the tests run `pbench-register-tools` and then check to see that the tools end up in the proper location.

But, I still think it would be better to convert these to a more unit-ish model and then put them in the appropriate place (e.g., add them to the existing `lib/pbench/test/unit/agent/test_config.py`).

riya-17:
Maybe what I can do is, Open another issue to move away from Functional tests and leave this as it is. In the third test, I can register the tools in the given location with `pbench-register-tool` and check whether the tools are registered in the given location with the help of `pbench-list-tools`.

lib/pbench/test/functional/agent/cli/commands/test_config.py
riya-17: Maybe what I can do is, Open another issue, (I think there are issues available for this) to move away from Functional tests and leave this as it is. In the third test, I can register the tools in the given location with `pbench-register-tool` and check whether the tools are registered in the given location with the help of `pbench-list-tools`.

lib/pbench/test/functional/agent/cli/commands/conftest.py
dbutenhof: That's a good point; I like the idea of a fixture to set this up, but maybe these should be layered, with the original `pbench_run` fixture to create the directory and a layered fixture to use the value of that fixture and establish the environment variable.

The new fixture would be used in the few targeted test cases that need it, while the remainder would continue to operate as they did before.

(And, for what it's worth, as constructed the `yield` is pretty silly; might as well be a `return` unless we're cleaning something up *after* the test ... which I don't believe is necessary since I think the scope of `tmp_path` is per-test.)

dbutenhof (commented):
> Maybe what I can do is, Open another issue to move away from Functional tests and leave this as it is. In the third test, I can register the tools in the given location with pbench-register-tool and check whether the tools are registered in the given location with the help of pbench-list-tools.

It doesn't hurt to start doing unit tests for new changes; I do like Webb's suggestion of starting that here by creating new tests in the proper place in the tree. I also think it's fine to just build your test cases using the current pseudo-functional test style; but we do have to stop doing that at some point.

webbnh (approved):
It would be nice if this PR added unit tests in addition to or instead of the functional tests.

It would be nice if this PR did not change things to picking up the `pbench_run` value from the environment variable instead of the config file _by default_.

And, it would be nice if the last test didn't have those silly one-iteration `for` loops in it.

But, it would also be nice to have this PR merged this week.

So, I guess it's good to go, once we have all the outstanding conversations reconciled.

lib/pbench/test/functional/agent/cli/commands/test_config.py
```diff
@@ -0,0 +1,44 @@
...
+    pbench_run_d.mkdir(parents=True, exist_ok=True)
+    monkeypatch.setenv("pbench_run", str(pbench_run_d))
+    for group in ["default"]:
+        for host in ["th1.example.com"]:
+            p = pbench_run_d / f"tools-v1-{group}" / host
+            p.mkdir(parents=True)
+            tool = p / "iostat"
+            tool.write_text("--interval=30")
+            tool = p / "mpstat"
+            tool.write_text("--interval=300")
```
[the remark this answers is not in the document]
riya-17: Thank you @webbnh I will make this change and update the PR :)

lib/pbench/test/functional/agent/cli/commands/conftest.py
dbutenhof: Of the various open issues, this is the only one that really bothers me; we've implicitly changed the behavior of all the existing tests using `pbench_run`. I'd much prefer to see a new fixture layered on top of `pbench_run` that adds the environment variable, leaving the old behavior unchanged and using this new fixture only for the new test cases specifically intended to check the environment variable behavior.

dbutenhof (approved):
I'm giving conditional approval; I'm not thrilled about the fixture change and I'd prefer that it be re-worked. However if Webb's willing to approve and resolve all his conversations, I'll go along.

webbnh (dismissed):
We seem to be down to just two unresolved "conversations":  whether we should be adding functional, unit, or both kinds of testing for this change; and, whether we should be making a change to the test system environment to access the `pbench_run` directory via the environment variable instead of the configuration file value.

The first one we can defer.  Although, in thinking about it, testing how we access the `pbench_run` value is not really appropriate for a functional test.  That is, that piece of functionality is present in nearly every one of the Agent tools, and so, from a functional test perspective, we should be testing that the functionality works _for nearly every one of the Agent tools_...not just `pbench-list-tools`.  Once again, we're trying to create a unit test out of a functional test, rather than just creating a proper unit test.

The second one is more annoying, especially in light of Dave's insight that the change affects every test **_except_** for the newly added ones.  😲  That pretty much puts that change out of scope for this PR!

So, I'm of like mind with Dave:  if he's willing to go along, I'm willing to go along, but I would prefer that the change to `conftest` be rescinded, and I'd like to see a follow-on PR real soon now which adds the unit tests.

riya-17:
> So, I'm of like mind with Dave: if he's willing to go along, I'm willing to go along, but I would prefer that the change to conftest be rescinded, and I'd like to see a follow-on PR real soon now which adds the unit tests.

I will work on adding a follow-up PR with unit tests. Maybe we can Merge this PR after adding that :)

riya-17:
> I think if you revert the change to lib/pbench/test/functional/agent/cli/commands/conftest.py then the tests included in this PR will pass as they are, and Dave and I will be comfortable having this merged.

@webbnh do we need to remove the change from conftest.py or test_config.py?

dbutenhof:
> @webbnh do we need to remove the change from conftest.py or test_config.py?

Reverting the fixture change in `conftest.py` to avoid forcing a `pbench_run` definition for every test should have *no* effect on the `test_config.py` as currently structured as none of your new tests depend on it. (They either re-define it or undefine it.) All the other tests should continue to work as they used to.

So, specifically, just remove the new `pbench_run` code in the `conftest.py` fixture and your `pytest` run should succeed.

riya-17:
> > @webbnh do we need to remove the change from conftest.py or test_config.py?
> 
> Reverting the fixture change in `conftest.py` to avoid forcing a `pbench_run` definition for every test should have _no_ effect on the `test_config.py` as currently structured as none of your new tests depend on it. (They either re-define it or undefine it.) All the other tests should continue to work as they used to.
> 
> So, specifically, just remove the new `pbench_run` code in the `conftest.py` fixture and your `pytest` run should succeed.

Reverting the fixture change does not have any effect on tests_config.py but it does have an effect on other tests.

Earlier `pbench_run` used to use the config value whether there is any environment value of `pbench_run` is available there or not.
Now after we have changed that to accepting the environment value of `pbench_run`, things changed. 

The change of considering environmental value first there is a change in tests behavior. Due to the weird setup of functional tests, there is already an environment value of `pbench_run` is present in functional tests and after this change, the tests are taking that as a `pbench_run` value instead of the one present in the config file. 
Due to this, I have to make the desired `pbench_run` value to environment value. So, tests can work well with the right `pbench_run` value. Either I will have to `Monkeypatch.del` the existing `pbench_run` environment value or I will have to `Monkeypatch.setenv` the `pbench_run` config file value.

webbnh:
> Where is this `pbench_run` defined?

That's an interesting question.  The answer seems to be it's defined "everywhere"...I just don't know know what the _scope_ of the definitions are.

That is, in the "base" scripts (`agent/base` and `agent/base.py`), if all goes "normally", we pull the value out of the config file and define it in the environment.  This is important because certain things (like `ToolGroup.verify_tool_group()`) may pull the value _from_ the environment for their operation.  But, it would appear to mean that any time we run any tool, `pbench_run` ends up in the environment, one way or another.

riya-17:
> That's an interesting question. The answer seems to be it's defined "everywhere"...I just don't know know what the _scope_ of the definitions are.
> 
> That is, in the "base" scripts (`agent/base` and `agent/base.py`), if all goes "normally", we pull the value out of the config file and define it in the environment. This is important because certain things (like `ToolGroup.verify_tool_group()`) may pull the value _from_ the environment for their operation. But, it would appear to mean that any time we run any tool, `pbench_run` ends up in the environment, one way or another.

yes, both `base` as well as `base.py` are exporting the `pbench_run` value

lib/pbench/test/functional/agent/cli/commands/conftest.py
```diff
@@ -23,9 +23,9 @@ def pbench_cfg(tmp_path, opt_pbench):
 
 @pytest.fixture
 def pbench_run(monkeypatch, tmp_path):
+    monkeypatch.delenv("pbench_run")
```
dbutenhof, line 26: Sure; that makes sense. We're creating a new `var/lib/pbench-agent` under some transient temporary, and I like the idea of making sure there's no existing `pbench_run` environment variable pointing somewhere else.

lib/pbench/test/functional/agent/cli/commands/test_config.py
```diff
@@ -4,7 +4,6 @@
 
 def test_pbench_run(monkeypatch, tmpdir, agent_config, pbench_run):  # noqa F811
     """Use pbench_run value from pbench-agent.cfg file"""
-    monkeypatch.delenv("pbench_run")
     env_pbench_run = os.environ.get("pbench_run")
     assert env_pbench_run is None
```
dbutenhof, line 8: This would be a bit more clear as a single line, since the local variable opens the question of whether there's another use of the variable:

```
assert os.environ.get("pbench_run") is None
```

or maybe even

```
assert "pbench_run" not in os.environ
```

At least, something to think about if you make another pass. 😆

dbutenhof (approved):

lib/pbench/test/functional/agent/cli/commands/test_config.py
riya-17: sure, I will also merge both commits :)

dbutenhof (approved):

lib/pbench/test/functional/agent/cli/commands/conftest.py
```diff
@@ -22,7 +22,8 @@ def pbench_cfg(tmp_path, opt_pbench):
 
 
 @pytest.fixture
-def pbench_run(tmp_path):
+def pbench_run(monkeypatch, tmp_path):
+    monkeypatch.delenv("pbench_run")
     pbench_agent_d = tmp_path / "var" / "lib" / "pbench-agent"
     pbench_agent_d.mkdir(parents=True, exist_ok=True)
```
webbnh, line 28: It turns out that we have two definitions of the `pbench_run` fixture -- one here (the "inner" one) and one in `lib/pbench/test/functional/conftest.py` (the "outer" one, which is superseded by the inner one for tests in `agent/cil/commands` and below):
```
@pytest.fixture
def pbench_run(tmpdir):
    pbench_run = tmpdir / "test/var/lib/pbench-agent"
    os.makedirs(pbench_run)
    yield pbench_run
```

[`tmpdir`](https://docs.pytest.org/en/6.2.x/reference.html#tmpdir) here is a built-in fixture provided by PyTest.  It has [been replaced](https://docs.pytest.org/en/6.2.x/fixture.html#:~:text=replaced%20by%20tmp_path.) by [`tmp_path`](https://docs.pytest.org/en/6.2.x/reference.html#std-fixture-tmp_path) which does the same thing.

So, prior to this PR, the two `pbench_run()` fixtures did substantially the same thing, although the "outer" one inserts an extra level of subdirectory (`test`) above the `/var/lib`.  With this PR, the "inner" one will also remove the `pbench_run` definition from the environment.

I think it's a bad idea to have two fixtures with the same name which are nearly but not entirely identical, and I think it's a worse idea to let them diverge further in behavior.  So, I propose that we delete this "inner" one (which is currently being modified in this PR) and use its code to replace the "outer" one -- existing code will then use the "outer" one.  (We can debate whether the `test` level should be retained or removed...I would remove it, but maybe it's a good safeguard against `tmp_path` somehow returning a null value.)

webbnh (changes requested):
I really want to see this PR merged, but, now that I've looked closer, there's another problem with the change to the `pbench_run` fixture.  😞

lib/pbench/test/functional/agent/cli/commands/test_config.py
npalaska: Hey @riya-17, I'll mark this as resolved, but can you please mention one of those issues here for our tracking.

webbnh (approved):
👍 (🎉)

npalaska (approved):

PR #2607, opened 2022-01-20
Rework Agent containers build

webbnh:
The current build procedure for the Agent containers no longer works.  The first problem is that it insists upon cross-checking all of its default build targets, including Fedora 32 (even if the requested build doesn't include F32...), which doesn't work now that we've removed our support for Fedora 32 from COPR and COPR itself has deprecated Fedora 33 support.  The second problem is that its build targets are specific to the respective target platforms, and so adding a new platform requires cookie-cutter copying the existing targets to create targets for the new platform.

This PR makes the following changes:
- Replace the old cookie-cutter approach with generic templates and pattern rules.  With this change, the Makefile no longer refers to specific platforms or targets except in the default list of targets to build.
- Switch from `docker.io` to `quay.io` as the source for base images.  Since the repositories have different names, and since the tags have different formats, this required some changes to the plumbing.
- Update the default list of targets to Fedora 34, Fedora 35, and CentOS 8 -- dropping Fedora 32, Fedora 33, and CentOS 7.

Note that there is nothing intrinsic to these changes which would prevent Fedora 32 or Fedora 36 from working, but we don't currently provide the required stuff on COPR for these targets.  Likewise, there is nothing in these changes which prevents CentOS 7 or CentOS 9 from working -- the CentOS 7 build had stopped working prior to these changes (and its build now fails in exactly the same way with these changes), and `packagecloud.io`, from which we source the Prometheus RPMs, does not yet support CentOS 9.

[I've tagged @portante as a reviewer for, oh, so many reasons; @ndokos, I would appreciate it if you could be my second reviewer, and anyone else is welcome to step up.]

### DIFF
--- a/agent/containers/images/Makefile
+++ b/agent/containers/images/Makefile
@@ -295,7 +295,7 @@ centos-7-base.Dockerfile: _PKGMGR = yum
 centos-%-base.Dockerfile: Dockerfile.base.j2
 	jinja2 Dockerfile.base.j2 -D pbench_repo_file=centos-$*-pbench.repo \
         -D pkgmgr=${_PKGMGR} -D distro_name="CentOS $*" -o $@ \
-        -D image_repo=quay.io/centos -D image_name=centos -D image_tag=centos$*
+        -D image_repo=quay.io/centos -D image_name=centos -D image_tag=stream$*
 
 fedora-%-base.Dockerfile: Dockerfile.base.j2
 	jinja2 Dockerfile.base.j2 -D pbench_repo_file=fedora-$*-pbench.repo \
--- a/agent/containers/images/Makefile
+++ b/agent/containers/images/Makefile
@@ -295,12 +295,12 @@ centos-7-base.Dockerfile: _PKGMGR = yum
 centos-%-base.Dockerfile: Dockerfile.base.j2
 	jinja2 Dockerfile.base.j2 -D pbench_repo_file=centos-$*-pbench.repo \
         -D pkgmgr=${_PKGMGR} -D distro_name="CentOS $*" -o $@ \
-        -D image_repo=quay.io/centos -D image_name=centos -D image_tag=stream$*
+        -D image_repo=quay.io/centos -D image_name=centos -D image_rev=$*
 
 fedora-%-base.Dockerfile: Dockerfile.base.j2
 	jinja2 Dockerfile.base.j2 -D pbench_repo_file=fedora-$*-pbench.repo \
         -D pkgmgr=dnf -D distro_name="Fedora $*" -o $@ \
-        -D image_repo=quay.io/fedora -D image_name=fedora -D image_tag=$*
+        -D image_repo=quay.io/fedora -D image_name=fedora -D image_rev=$*
 
 # Helper target to build each distro's ".repo" and ".Dockerfile"
 all-dockerfiles: \
--- a/agent/containers/images/Makefile
+++ b/agent/containers/images/Makefile
@@ -1,16 +1,46 @@
 # Base Makefile for building all images and tagging them
 
+# List of external Make targets:
+#
+#   FIXME:  Are there any missing from this list?  Are there any here which can/should be removed (e.g., the last two)??
+#
+#   These act on all default platforms' containers:
+#
+#     all (default):  make all images for all default platforms
+#     tds:  make Tool Data Sink images for all default platforms
+#     tm:  make Tool Meister images for all default platforms
+#     tag-<TYPE> (e.g., "tag-latest"):  apply specified tag to all images for all default platforms
+#     push:  push images for all default platforms with "<git commit hash>" and "v<full RPM version>" tags
+#     push-<TYPE> (e.g., "push-latest"):  push images for all default platforms with specified tag
+#
+#   These act on the indicated distribution's containers:
+#
+#     <DISTRO> (e.g., "fedora-34"):  make all images for the requested platform
+#     <DISTRO>-tds (e.g., "fedora-34-tds"):  make the TDS image for the requested platform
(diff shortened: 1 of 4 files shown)


agent/containers/images/Makefile
```diff
@@ -66,105 +98,92 @@ _WORKLOAD_RPMS = fio uperf
...
-tag-beta: $(foreach distro, ${_DISTROS}, ${distro}-tag-beta)
-
-# Add the "alpha" tag to the local images.
-tag-alpha: $(foreach distro, ${_DISTROS}, ${distro}-tag-alpha)
-
-# Add the "v<Major>-latest" tag to the local images.
-tag-major: $(foreach distro, ${_DISTROS}, ${distro}-tag-major)
+#
+# Available tags values are "latest", "alpha", "beta", "major", and "major-minor"
+# "major" produces a tag of "v<Major>-latest", and "major-minor" produces a tag
```
dbutenhof, line 150: Does Major Minor hang out with General Specialist?

agent/containers/images/Makefile
```diff
@@ -66,105 +98,92 @@ _WORKLOAD_RPMS = fio uperf
 # Not intended to be overridden with an environment variable.
 _ALL_RPMS = ${_TOOL_RPMS} ${_WORKLOAD_RPMS}
 
+# Compose lists of all conceivable distro-version strings (e.g.,
+# `_ALL_centos_VERSIONS` including "centos-7", "centos-8", "centos-9",
+# `_ALL_fedora_VERSIONS` including "fedora-32", "fedora-33", and "fedora-34",
+# and `_DISTROS` containing the contents of all the lists) which will be used
+# to drive pattern matching on distro-based target rules.
+_DIGITS := 0 1 2 3 4 5 6 7 8 9
+_NUMBERS := $(patsubst 0%,%,$(foreach t,${_DIGITS},$(foreach o,${_DIGITS},${t}${o})))
```
dbutenhof, line 107: Wow. I get it, I think, but that's rather dense.

agent/containers/images/Makefile
webbnh: 😁

I didn't write that...I just moved it there.  😉

agent/containers/images/Makefile
webbnh: I'm open to suggestions....  🙂

dbutenhof: That was more an expression of cautious awe than a comment. I think it takes a bit of work to unpack this nested loop (maybe it's more like a "make comprehension"), but I'm not sure the complexity justifies the paragraphs of description that would be needed to break it down in a comment.

agent/containers/images/Makefile
dbutenhof: Not a comment: just a random brain emission as I read the conjunction. 😆

dbutenhof (commented):
I'm impressed. It's still "Draft" so I won't bother to approve, but it seems to be looking rather comprehensive assuming testing proves it all out in practice.

agent/containers/images/Makefile
webbnh: I changed the inner loop indices from `t` and `o` to `tens` and `ones`...perhaps that'll be enough of a clue.

I did decide that a detailed "how it works" comment would be over-doing it, and so I went with just a "what it's doing" description instead, figuring that, as long as it works, that's the important part.  🦐

agent/containers/images/Makefile
```diff
@@ -1,16 +1,46 @@
 # Base Makefile for building all images and tagging them
 
+# List of external Make targets:
+#
+#   FIXME:  Are there any missing from this list?  Are there any here which can/should be removed (e.g., the last two)??
```
dbutenhof, line 5: Are you referring to the "all-tags" and "all-dockerfiles"? Yeah, I don't really know why we'd need these; perhaps we could generate them for validation, but is that really worthwhile?

dbutenhof (approved):
The `make` magic is kinda blinding, but it looks good.

agent/containers/images/Makefile
ndokos: I guess the idea is to be able to constuct pieces that one could use in a run "by hand", e.g. for debugging purposes. I vote we keep them.

ndokos (commented):
A tour-de-force of makefileology! Opaque but impressive. 

A couple of questions:
- Is it possible to produce the Makefile after all the automatic expansions have happened? That might be a good thing to have if one needs to debug it.
-  I wonder if some of the opaqueness could be alleviated by using a more expressive language, e.g. by having a program that would write a makefile for some distro (e.g centos-8) that the main Makefile could then run.

In general, I like programs that write programs, but I'm not sure how I feel about  Makefiles that create Makefiles (particularly if they don't write them out) :-)

Nevertheless, once the CentOS problems are resolved, I'm ready to approve.

ndokos (dismissed):
LGTM

agent/containers/images/Makefile
```diff
@@ -66,105 +110,93 @@ _WORKLOAD_RPMS = fio uperf
 # Not intended to be overridden with an environment variable.
 _ALL_RPMS = ${_TOOL_RPMS} ${_WORKLOAD_RPMS}
 
-# By default we only build images for the following distributions:
-_DISTROS = centos-8 centos-7 fedora-33 fedora-32
+# Compose lists of all conceivable distro-version strings (e.g.,
```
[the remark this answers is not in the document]
dbutenhof: Still not thrilled about the "all conceivable", but the extravagant claim in a comment is a relatively minor issue; I'll approve, and if you don't feel like fixing it I'll resolve the conversation.

dbutenhof (approved):

ndokos (approved):

dbutenhof (approved):

portante (approved):

PR #2611, opened 2022-01-26
Remove `pbench-server-prep-shim-002`

### LINKED ISSUE #2421: Remove prep-shim from 0.72 server!

dbutenhof:
This removes the `pbench-server-prep-shim-002` tool, which no longer serves a purpose as Server 1.0 will not support `scp` dataset upload. This inevitably affects many legacy test gold files simply because the name `pbench-server-prep-shim-002` is going away, but the set of dataset objects, users, and the "shape" of the filesystem will remain largely identical.

The biggest legacy test change is in test-5.2 where the test tarball `test-5.2.tar.xz` contained a set of "uploaded" files in the receive directory to test `pbench-server-prep-shim-002` processing. I've removed the entire directory, and adjusted the gold file appropriately. Note that most of these were intended to have been quarantined (duplicate, bad MD5), which is completely irrelevant in the new server as such datasets would have failed at `PUT` and would be left on the agent.

Resolves #2421.


lib/pbench/test/unit/server/query_apis/test_datasets_publish.py
```diff
@@ -5,7 +5,7 @@
 import elasticsearch
 import pytest
 
-from pbench.server.api.resources import JSON
+from pbench.server import JSON
```
webbnh, line 8: How fascinating.  Good catch...I guess.  🤯

webbnh (approved):
Seems good.  I only reviewed one of the tests, so there are 80-some files that I skipped over, but I've only got so much juice.  I noticed a few changes which seem out of scope, but I guess they are OK.  I found a few small things to mention (I "pre-resolved" most of them, you'll either do them or you won't).

portante (changes requested):
The commit message has the Jira issue ID as its title.  Can you change that to have a real summary title (<= 50 characters) and move that Jira issue ID to the first line of the commit message body?

dbutenhof:
> The commit message has the Jira issue ID as its title. Can you change that to have a real summary title (<= 50 characters) and move that Jira issue ID to the first line of the commit message body?

That's weird; pretty sure I didn't type it that way. Ah, well; I need to rebase anyway.

lib/pbench/cli/server/tree_manage.py
```diff
@@ -37,13 +37,22 @@ def print_tree(tree: FileTree):
...
 @click.option(
     "--display", default=False, is_flag=True, help="Display the full tree on completion"
 )
-@common_options
-def tree_manage(context: object, display: bool):
+@click.option(
+    "--full-discovery",
+    default=False,
+    is_flag=True,
+    help="Fully discovery file tree before performing requested operation",
```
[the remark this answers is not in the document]
dbutenhof: Ha; when I dropped the "create" logic (which was the only real reason for touching this file again), I very nearly just reverted it entirely. I like the new subcommand format, but it doesn't really make a lot of sense with just one. 😦

lib/pbench/server/filetree.py
```diff
@@ -231,33 +234,37 @@ def extract(self, path: str) -> str:
...
-            )
+            return tarfile.open(tarball, "r:*").extractfile(path).read().decode()
         except Exception as exc:
-            raise MetadataError(self.tarball_path, exc)
+            raise MetadataError(tarball, exc)
 
-    def get_metadata(self) -> Dict[str, Any]:
+    # FIXME: FileTree.create(tarball) should open the metadata.log and get the
+    # controller implicitly; and the metadata object should be stored on the
+    # Tarball object.
```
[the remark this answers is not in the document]
dbutenhof: Probably should have been a `TODO`; this is really referring to #2878, but I didn't do it quite the way I'd imagined when I started this several sprints ago.

lib/pbench/server/filetree.py
```diff
@@ -231,33 +234,37 @@ def extract(self, path: str) -> str:
...
-            "controller": {"hostname": metadata.get("controller", "hostname")},
             "pbench": {
-                "config": metadata.get("pbench", "config"),
-                "date": created.utc_time,
-                "script": metadata.get("pbench", "script"),
-                "version": metadata.get("pbench", "rpm-version"),
+                "config": metadata.get("pbench", "config", fallback=""),
+                "date": UtcTimeHelper.from_string(
+                    metadata.get("pbench", "date")
+                ).utc_time,
```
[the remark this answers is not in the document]
dbutenhof: And #2878 stops trying to convert the data on this end anyway in favor of returning the raw metadata from `metadata.log` as I can't easily type-infer every parameter recorded and I didn't want to try to special case. Yeah, this is ugly; but once #2878 goes in this won't even exist.

webbnh (approved):
👍

webbnh (approved):
👍

webbnh (approved):
👍

portante:
FYI, the previous commit resulting in a hung CI job.  I restarted the build to see if it was transitory.

server/rpm/pbench-server.spec.j2
```diff
@@ -22,8 +22,6 @@ Requires: python3-libselinux
 %endif
 
 Requires: npm
-Requires: pbench-dashboard
```
portante, line 25: If you are going to remove `pbench-dashboard`, you might as well remove `npm` too.  But why not move this to a separate commit?  This commit already has 100+ files in it.

server/bin/unittests
```diff
@@ -225,11 +225,6 @@ declare -A cmds=(
     # non-trivial results: mail
     [test-11]="_run pbench-sync-satellite satellite-one"
 
-    # shim - failure, no quarantine directory
-    [test-13]="_run pbench-server-prep-shim-002"
```
portante, line 229: Should the gold files also be removed?

server/ansible/pbench-server-disable.yml
```diff
@@ -8,5 +8,4 @@
 
   roles:
     - pbench-server-vars
-    - { role: pbench-server-set-state, pbench_server_state: off}
```
portante, line 11: Can we just make a separate PR entirely to remove the server ansible code?

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -90,7 +90,7 @@ class TestUpload:
     filetree_create_path = None
     tarball_deleted = None
 
-    @pytest.fixture
```
portante, line 93: Can we move this kind of change to a separate PR where we apply it everywhere else it is needed?

lib/pbench/test/unit/server/test_file_tree.py
```diff
@@ -174,6 +174,7 @@ def test_duplicate(self, selinux_disabled, server_config, make_logger, tarball):
         controller = tree.archive_root / "ABC"
         controller.mkdir()
         (controller / source_tarball.name).write_text("Send in the clones")
+        (controller / f"{source_tarball.name}.md5").write_text("notanmd5 notmyname")
```
portante, line 177: Why does this patch set require this addition?

lib/pbench/server/filetree.py
```diff
@@ -116,7 +116,8 @@ def __init__(self, path: Path, controller: "Controller"):
         # Record the path of the companion MD5 file
         self.md5_path: Path = path.with_suffix(".xz.md5")
 
-        # Record the name of the containing controller
+        # Record the resource ID and the name of the containing controller
+        self.md5 = self.md5_path.read_text().split()[0]
```
portante, line 120: What about this patch set caused this addition?

lib/pbench/cli/server/tree_manage.py
```diff
@@ -37,13 +37,27 @@ def print_tree(tree: FileTree):
                     print(f"        States: {', '.join(states)}")
 
 
-@click.command(name="pbench-tree-manager")
+@click.group("pbench-tree-manage")
```
portante, line 40: What about the removal of `pbench-server-prep-shim-002` required these changes below? Could this be "leaked" changes from another PR?

lib/pbench/test/unit/server/conftest.py
```diff
@@ -789,7 +789,12 @@ def tarball(tmp_path):
     datafile = tmp_path / filename
     metadata = ConfigParser()
     metadata.add_section("pbench")
+    metadata.add_section("run")
```
portante, line 792: What about the removal of `pbench-server-prep-shim-002` required changes to this test configuration?

lib/pbench/test/unit/server/query_apis/test_datasets_publish.py
portante, line 8: What about the removal of `pbench-server-prep-shim-002` required this change?

portante (changes requested):

server/rpm/pbench-server.spec.j2
webbnh: > If you are going to remove `pbench-dashboard`, you might as well remove `npm` too.

That's not quite true (I think).  That is, the Dashboard sources are now in the same repo as the Server, and the Dashboard will (quite soon, I hope) be installed by the same RPM as the Server, so, to the extent that the running Dashboard depends on `npm`, this spec file will need a `Requires` for it.

Thus, if you tear it out in this PR, I'll be adding it right back in in an upcoming PR.  🙂

webbnh (dismissed):
👍 

(Although, isn't Peter correct that we can delete a gold file?)

server/rpm/pbench-server.spec.j2
portante: Where does the dashboard code depend on `npm` at run time?

webbnh: That's a fair question.  I believe the answer is, "it doesn't".

However, other than "run-time", there's also "build-time" and "install-time".  Someone smarter than me needs to figure out whether we want to "build" the Dashboard when we build the RPM (and then do a straightforward copy-type installation of it), or whether we want to build the Dashboard when we "install" (which will likely make our RPM _much_ smaller, but it will require an Internet connection at install-time, and it's likely to produce less deterministic results).

Given that we intend to deploy the Server (and, therefore, the Dashboard) as a container, we'll control the "install-time" environment, so it may be reasonable to run `npm` at that point instead of during the RPM build.  (For the container deployment, it'll probably be "six of one vs. a half dozen of the other", but the difference will include whether the spec file requires `npm` or not.)

riya-17 (dismissed):

server/rpm/pbench-server.spec.j2
webbnh: > The dashboard `build` script relies on `npm`, but a dashboard deployment doesn't.

That statement is probably too difficult to unpack here.  (I.e., "replies" implies some "present tense", and I'm operating in the "near future"; and, how/when/where the dashboard gets "built" vs. "deployed" is actually open for discussion -- I'm contemplating doing virtually nothing for the Dashboard during the RPM build and then running `npm` when the RPM is installed, so, yeah, `npm` will need to be one of the `Requires`....)

>  I'm happy to remove that, too.

You are welcome to remove it, if it will please Peter and/or improve life for you.  However, I fully expect to put it right back in there, so it'll only be "gone" for a little while.  😀

webbnh (approved):
👍

lib/pbench/server/filetree.py
```diff
@@ -257,7 +257,7 @@ def create(cls, tarball: Path, controller: "Controller") -> "Tarball":
 
     def extract(self, path: str) -> str:
         """
-        Extract a file from the tarball and return it as a string
```
portante, line 260: Wouldn't we want to document this as "return it's contents as a string."?

And are we sure we want to alway decode the contents?

server/rpm/pbench-server.spec.j2
portante: Don't worry, it won't be coming back.  =)  We can have a discussion about how we'll handle deployments in the future.  The browser never makes a reference to files in the npm tree, so we don't need the dependency in the RPM.

portante (approved):

portante (changes requested):
I am no longer in favor of "approving" a PR while still having out-standing comments.

lib/pbench/server/filetree.py
webbnh: Adding "contents" is probably good; adding the apostrophe is not.  🙂

As for decoding, that's an interesting question.  I'm going to defer to Dave on this one, as I started digging into this and the hole got too deep.  Depending on what `extractfile()` returns, `read()` is likely to return an "encoded" string which is possibly _not_ what we want to return to the requestor, and therefore we probably should always decode it.  However, if we were to specify the encoding (or, perhaps, read the file in binary mode?), then we wouldn't need the decode step because it wouldn't have been encoded....

dbutenhof: This was not intended to be a full general artifact access method. It's also not new code, and I only brought attention to it because for some bizarre reason I tweaked the docstring. I don't intend to generalize it further in this context.

It exists so that the `PUT` code could extract `metadata.log` before we unpacked the tarball. I expect this will become a call into the cache manager, which will definitely need to handle binary files (e.g., the tarball itself) as well as text files. We may or may not need to support different encodings, or specify `utf-8` and be done with it (if we know everything we care about is ASCII or Unicode).

None of this, luckily, is relevant to this PR except wordsmithing the docstring... and I'm beginning to think I should just revert the change and pretend I never applied the can opener to this particular can of worms ...

server/rpm/pbench-server.spec.j2
dbutenhof: Right; there's no RPM dependency. If there's a Dockerfile dependency, that's a completely different matter and shouldn't be resolved by burdening the RPM.

webbnh: Well...the RPM can run commands when it is installed, and, for my changes, I was contemplating having it run `npm` then, in which case there _would_ be an RPM dependency.  However, Peter pointed out that there is another approach which is probably better, so I won't be needing this.

webbnh (dismissed):
👍

server/rpm/pbench-server.spec.j2
```diff
@@ -21,9 +21,6 @@ Requires: policycoreutils-python
 Requires: python3-libselinux
 %endif
 
-Requires: npm
```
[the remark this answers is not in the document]
dbutenhof: Ouch; good catch. I didn't see this code, but basically the vague suspicion that our dependency on `npm` didn't start with the old dashboard is why I didn't originally remove `npm`. The server's static `web-server` package, which has a `package.json`, depends on a number of `npm` packages to support display of the agent-generated `.html` files.

The generated `<script>` tags should probably refer to an external CDN instead of loading them locally (e.g., `<script src="/static/js/v0.3/d3.min.js"></script>`. We want to get rid of the agent-side generated html, but until then this actually means we can't drop Apache file sharing; even if the UI uses `datasets/inventory` to load the generated HTML into a frame, that code will still reference the Javascript packages from the server, and they're outside the dataset tarball.

So, yeah; big wrench, expertly lobbed right in the center of the works. We need our `npm` back, until the generated agent html is fixed or removed... which isn't going to happen for 0.69 or 0.71.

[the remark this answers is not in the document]
dbutenhof: > So, yeah; big wrench, expertly lobbed right in the center of the works. We need our `npm` back, until the generated agent html is fixed or removed... which isn't going to happen for 0.69 or 0.71.

Oh yeah; and if we had a CI pipeline that built and deployed an RPM, we'd have known this by now. 👍🏻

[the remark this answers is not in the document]
webbnh: Yeah...about that...Peter just talked me out of including the Dashboard in the Server RPM, so the pipeline may have gotten a little longer....

portante (approved):

Issue #2337, opened 2021-05-26
pbench-clear-tools potential potholes


ndokos:
I did some experiments with the current (0.71) version of `pbench-clear-tools` and was surprised with many of the outcomes.

The setup for every test in the following table was the same:
```
pbench-register-tool --group foo --name iostat
pbench-register-tool --group foo --name mpstat
pbench-register-tool-set 
```
so we always start with two tool groups (`default` with the default tool set and `foo` with `iostat` and `mpstat`.

Here's the result of calling `pbench-clear-results` with the indicated options:

| option                                    | result                                            | my expectation                   | comments                     | 
| -------------------------------------- | ------------------------------------------- | -------------------------------- | ---------------------------- | 
|                                                | clear default                                  | clear all groups                 |                              |
| –group=foo                            | clear foo                                        | clear foo                  | X: All tools removed message | 
| –group=default –group=foo  | clear foo                                         | clear both groups                | X: last one wins             | 
| –name=mpstat                      | clear mpstat from default                | clear mpstat from all groups     |                              | 
| –group=default –name=mpstat            | clear mpstat from default | clear mpstat from default        |                              | 
| –group=default –group=foo –name=mpstat | clear mpstat from foo     | clear mpstat from both groups    | X: last one wins (group)     |
| –name=mpstat –name=iostat              | clear iostat from default     | clear both tools from all groups | X: last one wins (name)      |
| --group=foo –name=mpstat –name=iostat   | clear iostat from foo     | clear both tools from foo        | X: last one wins (name)      |

There are no tracebacks thankfully, but there is one bad message: "All tools removed" when all the tools in a group are removed even though there are tools in another group. The CLI (here and probably in other cases) seems to accept multiple entries for an option but uses a "last one wins" policy, so the only way to remove all tools from every group is to loop over every group.

All of this is probably OK if the user has a single group named 'default`, but for any more complicated situation, I find the handling confusing.

- I think the bad message should probably be fixed to indicate the group that was cleared. 

- Should we signal an error for multiple entries of a single option (IOW, do not put ourselves in the position of having to enforce "last one wins" or similar)?

- Should we provide a way to clear a tool from all groups?

- Should we provide a way to clear all tools from all groups?

dbutenhof:
If there's a case where we think we ought to support multiple instances of
an option, like --group, Click can do that
<https://click.palletsprojects.com/en/7.x/arguments/#variadic-arguments>.
[...]

PR #2612, opened 2022-01-26
Small improvements to `pbench-list-tools` and `pbench-clear-tools`

### LINKED ISSUE #2458: pbench-list-tools issues (brought forward from a now-closed issue)
These were comments by @webbnh in the now-closed issue #1504 "pbench-list-tools not finding the group members" (which was resolved by PR #2301):

> Initial investigation

>This command works OK for me (although the command produces no output...) using quay.io/pbench/pbench-agent-all-fedora-32:v0.71.0-130, so, presumably the issue here is related to something in the configuration.

>However, if I mess around with the -C switch or the _PBENCH__AGENT_CONFIG environment variable, the code responds "ungracefully" (but doesn't produce the result above).

>It would be helpful if there were command options which listed the tool names and group names which were available to be supplied to the -n and -g switches, and if those switches validated their options. (I will look into adding these.)

>(And, there's a typo in the help text for the -C switch: envrionment ==> environment.)

I'm pretty sure the typo in the 4th paragraph is fixed. The others may still be open.

webbnh:
This PR is a once-over on `pbench-list-tools` and `pbench-clear-tools`:
- Fixed up docstrings.
- Tweaked and corrected the help text to match the actual functionality.
- Finish by calling the Click `exit()` function instead of `sys.exit()` (and fixed up the `import`'s accordingly).
- Picked lint (fixed import ordering, added an underscore prefix to unused parameters, added type hints).
- If a bad tool group is specified (e.g., to `pbench-clear-tools`, we print a message (with a slightly different formatting) and we no longer print the usage information.
- Added diagnostic messages to most cases where the tools previously exited silently; changed some of the existing messages to make them more useful to end-users.
- Reworked the `is_empty()` function.
- Only changed the unit tests to reflect the changes to the messages.

Fixes #2458.

### DIFF
--- a/lib/pbench/agent/base.py
+++ b/lib/pbench/agent/base.py
@@ -112,7 +112,9 @@ class BaseCommand(metaclass=abc.ABCMeta):
         try:
             self.tool_group_dir = self.gen_tools_group_dir(group)
         except BadToolGroup as exc:
-            click.echo(f'{self.name}: invalid --group option "{group}" ({exc})')
+            click.echo(
+                f'{self.name}: invalid --group option "{group}" ({exc})', err=True
+            )
             ret_code = 1
         else:
             ret_code = 0
--- a/lib/pbench/agent/base.py
+++ b/lib/pbench/agent/base.py
@@ -112,9 +112,7 @@ class BaseCommand(metaclass=abc.ABCMeta):
         try:
             self.tool_group_dir = self.gen_tools_group_dir(group)
         except BadToolGroup as exc:
-            click.echo(f'\t{self.name}: invalid --group option ("{group}"), {exc}')
-            ctxt = click.get_current_context()
-            click.echo(ctxt.get_help())
+            click.echo(f'{self.name}: invalid --group option "{group}" ({exc})')
             ret_code = 1
         else:
             ret_code = 0
--- a/lib/pbench/cli/agent/commands/tools/clear.py
+++ b/lib/pbench/cli/agent/commands/tools/clear.py
@@ -44,7 +44,7 @@ class ClearTools(ToolCommand):
                 self.logger.error(f'No such group "{self.context.group}".')
                 return 1
 
-        tool_not_found = bool(self.context.name)  # Can't not find if not specified
+        tool_not_found = True  # Not found, yet
         for remote in remotes:
             tg_dir_r = self.tool_group_dir / remote
             if not tg_dir_r.exists():
@@ -96,7 +96,7 @@ class ClearTools(ToolCommand):
                     self.logger.error("Failed to remove remote directory %s", tg_dir_r)
                     errors = 1
 
-        if tool_not_found:
+        if self.context.name and tool_not_found:
             self.logger.warn(f'Tool "{self.context.name}" not found')
 
         return 0 if errors == 0 else 1
--- a/lib/pbench/cli/agent/commands/tools/clear.py
+++ b/lib/pbench/cli/agent/commands/tools/clear.py
@@ -35,6 +35,12 @@ class ClearTools(ToolCommand):
             # from the tools group directory.
             remotes = self.remote(self.tool_group_dir)
             if not remotes:
+                # FIXME:  This is a weird case -- the group directory exists,
+                #  but contains no remotes. This is the result of clearing the
+                #  tools from a group which has already been cleared, because
+                #  the first clear does not remove the group directory.  This
+                #  behavior might be desirable for the `default` directory, but
+                #  it seems wrong for custom groups.
                 self.logger.error(f'No such group "{self.context.group}".')
                 return 1
 
@@ -53,6 +59,8 @@ class ClearTools(ToolCommand):
                 # Discover all the tools registered for this remote
                 names = self.tools(tg_dir_r)
                 if not names:
+                    # FIXME:  this is another odd case -- the remote subdirectory
+                    #  exists, but it's empty.  (We'll remove it below.)
                     self.logger.warn(
                         'No tools in group "%s" on host "%s".',
                         self.context.group,
--- a/lib/pbench/cli/agent/commands/tools/clear.py
+++ b/lib/pbench/cli/agent/commands/tools/clear.py
@@ -8,7 +8,6 @@ Specifying a tool name and/or remote host will limit the scope of the removal.
 
 import pathlib
 import shutil
-import sys
 
 import click
 
@@ -23,7 +22,7 @@ class ClearTools(ToolCommand):
     def __init__(self, context):
         super().__init__(context)
 
-    def execute(self):
+    def execute(self) -> int:
         errors = 0
 
         if self.verify_tool_group(self.context.group) != 0:
@@ -35,27 +34,37 @@ class ClearTools(ToolCommand):
             # We were not given any remotes on the command line, build the list
             # from the tools group directory.
             remotes = self.remote(self.tool_group_dir)
+            if not remotes:
+                self.logger.error(f'No such group "{self.context.group}".')
+                return 1
 
+        tool_not_found = bool(self.context.name)  # Can't not find if not specified
         for remote in remotes:
             tg_dir_r = self.tool_group_dir / remote
             if not tg_dir_r.exists():
                 self.logger.warn(
-                    'The given remote host, "%s", is not a directory in %s.',
-                    remote,
-                    self.tool_group_dir,
+                    'No remote host "%s" in group %s.', remote, self.context.group,
                 )
                 continue
 
             if self.context.name:
-                names = self.context.name
+                names = [self.context.name]
             else:
                 # Discover all the tools registered for this remote
                 names = self.tools(tg_dir_r)
+                if not names:
+                    self.logger.warn(
+                        'No tools in group "%s" on host "%s".',
+                        self.context.group,
+                        remote,
+                    )
 
             for name in names:
                 status = self._clear_tools(name, remote)
-                if status != 0:
-                    errors += 1
+                if status > 0:
+                    errors = 1
+                elif status == 0:
+                    tool_not_found = False
 
             tool_files = [p.name for p in tg_dir_r.iterdir()]
 
@@ -65,26 +74,37 @@ class ClearTools(ToolCommand):
                     label.unlink()
                 except Exception:
                     self.logger.error("Failed to remove label for remote %s", tg_dir_r)
-                    errors += 1
+                    errors = 1
 
             if self.is_empty(tg_dir_r):
-                self.logger.info('All tools removed from host, "%s"', tg_dir_r.name)
+                self.logger.info(
+                    'All tools removed from group "%s" on host "%s"',
+                    self.context.group,
+                    tg_dir_r.name,
+                )
                 try:
                     shutil.rmtree(tg_dir_r)
                 except OSError:
                     self.logger.error("Failed to remove remote directory %s", tg_dir_r)
-                    errors += 1
+                    errors = 1
 
-        return errors
+        if tool_not_found:
+            self.logger.warn(f'Tool "{self.context.name}" not found')
 
-    def _clear_tools(self, name, remote):
-        """Remove specified tool and associated files"""
+        return 0 if errors == 0 else 1
+
+    def _clear_tools(self, name, remote) -> int:
+        """
+        Remove specified tool and associated files
+
+        Returns zero for success, -1 for name-not-found, and 1 for failure.
+        """
         tpath = self.tool_group_dir / remote / name
         try:
             tpath.unlink()
         except FileNotFoundError:
             self.logger.debug('Tool "%s" not registered for remote "%s"', name, remote)
-            return 0
+            return -1
         except Exception as exc:
             self.logger.error("Failed to remove %s: %s", tpath, exc)
             ret_val = 1
@@ -102,18 +122,17 @@ class ClearTools(ToolCommand):
 
         if ret_val == 0:
             self.logger.info(
-                'Removed "%s" from host, "%s", in tools group, "%s"',
+                'Removed "%s" from host "%s" in tools group "%s"',
                 name,
                 remote,
                 self.context.group,
             )
         return ret_val
 
-    def is_empty(self, path):
-        """Determine if directory is empty or not"""
-        for dirent in path.iterdir():
-            return False
-        return True
+    @staticmethod
+    def is_empty(path):
+        """Determine if directory is empty"""
+        return not any(path.iterdir())
 
 
 def _group_option(f):
@@ -128,10 +147,12 @@ def _group_option(f):
         "-g",
         "--group",
         default="default",
-        required=True,
         expose_value=False,
         callback=callback,
-        help="list the tools used in this <group-name>",
+        help=(
+            "Clear the tools in the <group-name> group.  "
+            "If no group is specified, the 'default' group is assumed."
+        ),
     )(f)
 
 
@@ -140,9 +161,9 @@ def _name_option(f):
 
     def callback(ctxt, _param, value):
         clictxt = ctxt.ensure_object(CliContext)
-        clictxt.name = []
+        clictxt.name = None
         if value:
-            clictxt.name.append(value)
+            clictxt.name = value
         return value
 
     return click.option(
@@ -150,10 +171,7 @@ def _name_option(f):
         "--name",
         expose_value=False,
         callback=callback,
-        help=(
-            "list the tool groups in which <tool-name> is used.\n"
-            "Not allowed with the --group option"
-        ),
+        help="Clear only the <tool-name> tool.",
     )(f)
 
 
@@ -172,8 +190,9 @@ def _remote_option(f):
         expose_value=False,
         callback=callback,
         help=(
-            "a specific remote on which tools needs to be cleared.\n"
-            "If no remote is specified, all the tools on all remotes are removed"
+            "Clear the tool(s) only on the specified remote(s).  "
+            "Multiple remotes may be specified as a comma-separated list.  "
+            "If no remote is specified, all remotes are cleared."
         ),
     )(f)
 
@@ -186,4 +205,4 @@ def _remote_option(f):
 @pass_cli_context
 def main(ctxt):
     status = ClearTools(ctxt).execute()
-    sys.exit(status)
+    click.get_current_context().exit(status)
--- a/lib/pbench/cli/agent/commands/tools/clear.py
+++ b/lib/pbench/cli/agent/commands/tools/clear.py
@@ -1,15 +1,14 @@
 """
 pbench-clear-tools
 
-This script will remove tools that have been registered.  If no options are
-used, then all tools from the "default" tool group are removed.  Specifying
-a tool name and/or remote host will limit the scope of the removal.
-
-
+This script will remove tools that have been registered for a particular group.
+If no options are used, then all tools from the "default" tool group are removed.
+Specifying a tool name and/or remote host will limit the scope of the removal.
 """
+
+import pathlib
 import shutil
 import sys
-import pathlib
 
 import click
 
@@ -41,7 +40,7 @@ class ClearTools(ToolCommand):
             tg_dir_r = self.tool_group_dir / remote
             if not tg_dir_r.exists():
                 self.logger.warn(
-                    'The given remote host, "%s", is not a directory in' " %s.",
+                    'The given remote host, "%s", is not a directory in %s.',
                     remote,
                     self.tool_group_dir,
                 )
@@ -120,7 +119,7 @@ class ClearTools(ToolCommand):
 def _group_option(f):
     """Group name option"""
 
-    def callback(ctxt, param, value):
+    def callback(ctxt, _param, value):
         clictxt = ctxt.ensure_object(CliContext)
         clictxt.group = value
         return value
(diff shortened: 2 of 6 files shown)


lib/pbench/cli/agent/commands/tools/clear.py
```diff
@@ -129,17 +139,19 @@ def callback(ctxt, param, value):
...
     )(f)
 
 
 def _name_option(f):
     """Tool name to use"""
 
-    def callback(ctxt, param, value):
+    def callback(ctxt, _param, value):
         clictxt = ctxt.ensure_object(CliContext)
         clictxt.name = []
```
npalaska, line 156: Why do we use the list here?, we could just do (like we do with group option)
```
clictxt = ctxt.ensure_object(CliContext)
clictxt.name = value
return value
```
and then on line 52 we can do `names = self.context.name.split(",")` that way we support the comma separated `name` option

lib/pbench/cli/agent/commands/tools/clear.py
```diff
@@ -151,17 +163,14 @@ def callback(ctxt, param, value):
         "--name",
         expose_value=False,
         callback=callback,
-        help=(
-            "list the tool groups in which <tool-name> is used.\n"
-            "Not allowed with the --group option"
-        ),
```
npalaska, line 157: Nice! we do need to update the help message in pbench-clear-tool (as this messages are from the list tools command) :+1:

lib/pbench/cli/agent/commands/tools/clear.py
```diff
@@ -36,12 +34,15 @@ def execute(self):
             # We were not given any remotes on the command line, build the list
             # from the tools group directory.
             remotes = self.remote(self.tool_group_dir)
+            if not remotes:
+                self.logger.error(f'No such group "{self.context.group}".')
+                return 1
 
         for remote in remotes:
```
npalaska, line 41: I guess we also need to do 
```
groups = self.context.group.split(",")
for group in groups:
```
to support multiple comma separated groups

lib/pbench/cli/agent/commands/tools/clear.py
```diff
@@ -66,17 +73,21 @@ def execute(self):
...
                     shutil.rmtree(tg_dir_r)
                 except OSError:
                     self.logger.error("Failed to remove remote directory %s", tg_dir_r)
-                    errors += 1
+                    errors = 1
 
-        return errors
+        return 0 if errors == 0 else 1
 
     def _clear_tools(self, name, remote):
```
npalaska, line 92: Should we add Type-hints here?

npalaska (commented):
Its still a draft but I have some thoughts

lib/pbench/cli/agent/commands/tools/clear.py
webbnh: Thanks.  (I'm a little fuzzy on the second part...is this all good?)

lib/pbench/cli/agent/commands/tools/clear.py
webbnh: Maybe in the next PR.  😉

lib/pbench/cli/agent/commands/tools/clear.py
webbnh: We could, but I'm disinclined:  IMO, we should add type hints to new code and to existing function definitions when the signatures are changed or at any time substantial modifications are made to its body; however, adding type hints to functions which aren't otherwise being changed is a gray area -- it amounts to "change for change sake", which comes with costs and risks which are probably best avoided.

lib/pbench/agent/base.py
```diff
@@ -112,9 +112,7 @@ def verify_tool_group(self, group):
         try:
             self.tool_group_dir = self.gen_tools_group_dir(group)
         except BadToolGroup as exc:
-            click.echo(f'\t{self.name}: invalid --group option ("{group}"), {exc}')
-            ctxt = click.get_current_context()
-            click.echo(ctxt.get_help())
+            click.echo(f'{self.name}: invalid --group option ("{group}") -- {exc}')
```
dbutenhof, line 115: I'm not sure I like this `-- {exc}` style, if only because it clashes with the earlier `--group option` phrase. I don't particularly love the `,` either, but `:` might be better here.

lib/pbench/cli/agent/commands/tools/clear.py
```diff
@@ -24,7 +22,7 @@ class ClearTools(ToolCommand):
     def __init__(self, context):
         super().__init__(context)
 
-    def execute(self):
+    def execute(self) -> int:
```
dbutenhof, line 25: Is this "change for change sake?" Asking for a friend... 😆

lib/pbench/cli/agent/commands/tools/clear.py
```diff
@@ -36,12 +34,15 @@ def execute(self):
...
+            if not remotes:
+                self.logger.error(f'No such group "{self.context.group}".')
+                return 1
 
         for remote in remotes:
             tg_dir_r = self.tool_group_dir / remote
             if not tg_dir_r.exists():
                 self.logger.warn(
-                    'The given remote host, "%s", is not a directory in' " %s.",
+                    'The given remote host, "%s", is not a directory in %s.',
```
dbutenhof, line 45: Normally as a "log" message I'd say this is OK, but since these seem to be the only user output rather than internal logs, complaining about the internal representation of "remote" as a subdirectory seems wonky. Shouldn't we say something more like "The remote host isn't in tool group 'group'"?

lib/pbench/cli/agent/commands/tools/list.py
```diff
@@ -70,7 +76,19 @@ def execute(self):
...
             if tool_info:
-                self.print_results(tool_info, None, self.context.with_option)
+                found = self.print_results(tool_info, self.context.with_option)
+                if not found:
+                    self.logger.warn(
+                        "No tools found%s",
+                        f' in group "{self.context.group[0]}"'
+                        if self.context.group
+                        else "",
+                    )
```
dbutenhof, line 86: The combination of logger with f-string syntax always bugs me. The lazy evaluation isn't an issue here since we presumably don't expect `WARN` to be disabled, but it still looks odd. (Even aside from the blackened ternary.)

lib/pbench/cli/agent/commands/tools/list.py
```diff
@@ -98,8 +116,10 @@ def execute(self):
                             else ""
                         )
                         found = True
+
             if found:
-                self.print_results(tool_info, tool, self.context.with_option)
+                found = self.print_results(tool_info, self.context.with_option)
+                assert found, "Tool found but not printed!"
```
dbutenhof, line 122: I don't see how this `assert` (or being able to write it) actually helps anyone. We didn't print anything and we're about to exit after finding something we think should have been printed; we abort with "found is False because we found something but didn't print it" ... it doesn't strike me as being useful diagnostic context because it only re-states the obvious in what seems an unnecessarily fatal manner for a transient CLI command.

lib/pbench/agent/base.py
webbnh: When I saw it printed out, the comma looked funny.  I was reaching for the colon, but then I noticed that there was already one in the line, and I feel that they don't nest.  So, I ended up with an "em dash".

I could change it to a single hyphen (or I could pull out the code point for a real unicode em dash 😉) or I could be convinced to go with a colon.

lib/pbench/cli/agent/commands/tools/clear.py
webbnh: No...I added a `return` statement (and changed the basis for another), and I wasn't sure whether it should return an `int` or a `bool`, so I figured I should record my conclusion.

(Tell your friend, "It's OK."  😉)

lib/pbench/cli/agent/commands/tools/clear.py
webbnh: Probably.  My intention with this change was just to remove the string concatenation (and to make the quotation marks uniform).  Of course, in this context, the arguments actually are directory names/paths, so the message makes sense...although, I concur, this doesn't necessarily help the end-user much.

Reading over these two scripts, I was struck by how not-particularly-user-friendly it is.  But, I didn't really want to scrap it completely, so I tried to touch up only the most egregious bits and pass over the rest.

I really wish there were a "style guide" for diagnostic messages.  I feel like we had something like that back in the VMS days...and it kinda got lost with the move to Unix and now it's completely gone on Linux....

lib/pbench/cli/agent/commands/tools/list.py
webbnh: Yeah, when I saw what Black did to it, I almost ripped it back out.  (In an earlier iteration, the ternary all fit on one line, so it was reasonable, but then the indentation level got increased or something, and then Black wrecked it.)

I'll try building the string before the logger invocation.

webbnh: OK, I like that better:
```
                    msg = "No tools found"
                    if self.context.group:
                        msg += f' in group "{self.context.group[0]}"'
                    self.logger.warn(msg)
```
(I think I'm going to give up using the ternary in Python. 😢)

lib/pbench/cli/agent/commands/tools/list.py
webbnh: One of the things I was trying to accomplish with these changes is to remove the cases where the command completes silently.  I'm of the opinion (which might, of course, be _wrong_...) that this command should either issue the requested information on `stdout` or a diagnostic on `stderr` -- it shouldn't complete without a comment, one way or the other.  (I.e., either the user gets what they asked for, or they get told why they were wrong to ask!  😀)

Eventually, I hit that spot in the code, where it _should_ have printed something, and `found` _shouldn't_ ever be false, there, so I figured I would put that assertion in for testing purposes.

I'll take it out (along with the reassignment of `found`) before I open this PR for review.

lib/pbench/agent/base.py
webbnh: We could remove the parens around the group name (it's already in quotation marks) and put them around the exception message...would that be good?

dbutenhof: Yeah, I like that a better ... parenthesizing the group name here seems weird anyway.

lib/pbench/cli/agent/commands/tools/clear.py
webbnh: OK, I ended up changing what this function returns for values, so I added the type hint to the signature.  🙂

lib/pbench/agent/base.py
webbnh: Now that I've seen the message that the new formatting produces, I kinda like it.  🙂

lib/pbench/cli/agent/commands/tools/list.py
webbnh: It's gone now.  (I never saw it trigger during testing, as I'd hoped would be the case....)

lib/pbench/cli/agent/commands/tools/clear.py
webbnh: OK, you convinced me...I replaced the list with a direct assignment.  😀

dbutenhof (approved):

npalaska (dismissed):

riya-17 (approved):

lib/pbench/cli/agent/commands/tools/clear.py
```diff
@@ -36,27 +34,45 @@ def execute(self):
...
+                # FIXME:  This is a weird case -- the group directory exists,
+                #  but contains no remotes. This is the result of clearing the
+                #  tools from a group which has already been cleared, because
+                #  the first clear does not remove the group directory.  This
+                #  behavior might be desirable for the `default` directory, but
+                #  it seems wrong for custom groups.
+                self.logger.error(f'No such group "{self.context.group}".')
+                return 1
+
+        tool_not_found = bool(self.context.name)  # Can't not find if not specified
```
dbutenhof, line 47: OK, I find your double negative vaguely disturbing; I also think this comment doesn't help much in understanding the logic. Perhaps something more like "if tool name is specified, assume it won't be found (until proven otherwise)"...

lib/pbench/cli/agent/commands/tools/clear.py
```diff
@@ -36,27 +34,45 @@ def execute(self):
...
                 # Discover all the tools registered for this remote
                 names = self.tools(tg_dir_r)
+                if not names:
+                    # FIXME:  this is another odd case -- the remote subdirectory
+                    #  exists, but it's empty.  (We'll remove it below.)
+                    self.logger.warn(
+                        'No tools in group "%s" on host "%s".',
+                        self.context.group,
+                        remote,
+                    )
```
dbutenhof, line 68: Is this a case the *user* really even needs to know about? As a "real" `logger.warn` I might be in favor, but as these commands (ridiculously) use a file-streamed logger as user feedback, this seems excessive. It's almost certainly an internal error, it's not harmful, and we're about to clean it up.

lib/pbench/test/functional/agent/cli/commands/tools/test_clear_tools.py
```diff
@@ -70,6 +77,8 @@ def test_clear_tools_test65(monkeypatch, agent_config, pbench_run, pbench_cfg):
     assert pidstat_tool.exists() is False
     assert turbostat_tool.exists() is False
     assert mpstat_tool.exists() is False
+    # FIXME:  Why do we want to retain the group directory after deleting everything out of it??
+    # FIXME:  Contrary to the test docstring, this test doesn't create or assert that the default group is unmolested.
```
dbutenhof, line 81: These be good questions, matie. Can we fix them? 👍🏻

dbutenhof (dismissed):

lib/pbench/cli/agent/commands/tools/clear.py
webbnh: I think it's actually a _triple_ negative.  (Negative boolean variables are always slightly tricky to describe succinctly.)

The problem is that I'm not sure that this one line is worth a block comment (if you think it is, I will happily provide one!), so it was a challenge to "say it all" in 30 or so characters, and that's what I came up with.  😆

I'm willing to remove the comment if you think it provides, um, negative value.  😉

lib/pbench/cli/agent/commands/tools/clear.py
webbnh: Unfortunately, I'm building an edifice on a rickety foundation, here.

It should be that this case never arises, and, therefore, there should be no need to report it to anybody.  Yet, the code (silently, prior to my change) defends itself against the situation.  And, in the case where there is only one host (or where all the hosts are in the same state), the command will exit silently, which is something that I'm trying to ensure never happens:  I want it to either report success or report errors -- or both -- but it should never report nothing.

So, assuming that the user remembers (or at least _did_) register tools for that host, I think it's worth having the tool note that there are no tools for that host, now (i.e., that the "clear" operation was aware of the host, and tried to clear the requested tools from it, but determined that they weren't there).

[But...I was hoping that we could discuss this, which is why I put the `FIXME` there....  🙂]

lib/pbench/test/functional/agent/cli/commands/tools/test_clear_tools.py
webbnh: Before I can fix the first one, I need to know whether it is intended behavior...and perhaps _why_....

As for fixing the second one...well, originally, my intention was to make as few changes to the unit tests as possible, to "show" that my changes to the CUT were "good" (i.e., if the new code works with the old test, then it must be OK, right??), so I didn't want to step too far out on that slippery slope...but, y'know, I'm sure that changing comments like docstrings would be _harmless_...of course.  😆

npalaska: About the first one I don't think it is an intended behavior I think it's a bug (I dont see how it can be an intended behavior since we can always create a same directory again if it doesn't exist). We should always create tools directory on demand and shouldn't leave any garbage on disk when a user clears those tools.

lib/pbench/cli/agent/commands/tools/clear.py
dbutenhof: Actually, I find the logic of pre-supposing an error and canceling that error to be awkward. I'd much rather use a positive logic, although I understand the complication with trying to factor in the state of the option.

Still, your initial `bool(self.context.name)` isn't "tool not found" but "tool specified", whereas it *becomes* "tool found" at line 75. Which makes the final test really "if tool specified and not found"; but the implementation makes this relationship fuzzy and poorly documented.

It could be improved with just a name change, maybe, but I'd prefer to have the code more explicitly model the logic.

In any case, if you have trouble writing a short line comment describing something, it's probably time for (a) a block comment, or (b) cleaner logic that's easier to describe. 😆

webbnh: Speaking most generally, I think code should be self-documenting...so, if I need a **block** comment to explain it, it better be real important for it to be that way, or it should be re-written.  (I felt like that line would benefit from some esplainin', and then I was amused by the phrasing that the lack of space imposed on me, so I left it there to see if it would amuse you.)

But, clearly, this change is inappropriate, so I shall amend it.  🙂

(Although, I'm not coming up with a way to do it which uses "positive logic".  And, looking at the obvious name-change, I think I'll rework the logic instead.  😆)

lib/pbench/cli/agent/commands/tools/clear.py
webbnh: That basic UNIX command philosophy is targeted at pipeline components, not at human users.

Given that the command in question doesn't take input from `stdin` nor produce output which is designed to be passed on to a subsequent pipeline stage, I'd say this command falls into the "human interface" bucket, where "silence is ambiguous".

I would be willing to put this output under a `-v` flag if we had one, but, given that we don't, I would prefer to just put it in the user's view in the rare case where it's actually appropriate.

What I'm more interested in pondering here is whether we should be contemplating this case at all.  E.g., should we just be skipping this `remote` completely like we do in the previous `if`-test, if the directory _doesn't_ exist?  The fact that this test triggered indicates that the subdirectory is in a "bad" state...should we clean it up or ignore it?  (Perhaps we should tell the user to clean it up?)

lib/pbench/test/functional/agent/cli/commands/tools/test_clear_tools.py
webbnh: What I find curious is how awkwardly the tools behave when the `default` group doesn't exist.

In point of fact, the `default` group is no different from any other group -- it's just the group which gets used when no other group is specified -- but there is a difference in behavior between dealing with a group which has never existed (this is a "Bad tool group") and a group which once existed but has since been cleared (which is silently ignored).  So, if you've never had a `default` group, the first time you run these commands, you'll get something like this:
```
# pbench-list-tools
No tool groups found
# pbench-clear-tools 
pbench-clear-tools: invalid --group option "default" (Bad tool group, 'default': directory /var/lib/pbench-agent/tools-v1-default does not exist)
# pbench-register-tool-set 
"vmstat" tool is now registered for host "401c3456f30f" in group "default"
"iostat" tool is now registered for host "401c3456f30f" in group "default"
"sar" tool is now registered for host "401c3456f30f" in group "default"
# pbench-clear-tools 
Removed "iostat" from host "401c3456f30f" in tools group "default"
Removed "sar" from host "401c3456f30f" in tools group "default"
Removed "vmstat" from host "401c3456f30f" in tools group "default"
All tools removed from group "default" on host "401c3456f30f"
# :
# pbench-list-tools
No tools found
# pbench-clear-tools 
No such group "default".
```
Note the difference between the first invocation of `pbench-list-tools` and the last, and between the first invocation of `pbench-clear-tools` and the last.

I'm inclined to defer changing the behavior with respect to removing the tool directories to another PR (perhaps the one addressing #2337?  😁), but I agree that I think we should be removing them...but I think we should handle the situation when `default` is missing a little more gracefully, even if it requires some special-case code.  In the contest of that change, it might be appropriate to add some assertions to the unit tests.

webbnh:
This is ready for review again.

The foremost question is, what do we want to do with the `FIXME`'s?
- We can leave them as they are, with the expectation that the PR for #2337 will be along shortly and will address and remove them.
- I can remove them and convert them to Issues.
- I can remove them and add fixes for them (but I'd rather see that done in another PR and the one addressing #2337 would be an obvious choice...).

There are four:
- two which deal with left-over cleared-out tool group directories
- one which deals with tool remote subdirectories which are empty
- one which deals with an inappropos docstring in a test which should probably be addressed by changing the code rather than the docstring, once we make the change to remove the cleared-out tool group directories.

lib/pbench/test/functional/agent/cli/commands/tools/test_clear_tools.py
npalaska: Yes It can be a part of #2337 since you pointed this bug under that issue. :)

npalaska:
> This is ready for review again.
> 
> The foremost question is, what do we want to do with the `FIXME`'s?

I stand with your 1st point. I guess we can leave them here and when someone opens a PR to fix this bug, the `FIXME`s can be removed.

dbutenhof (approved):
I'm not thrilled about leaving the FIXME comments in place, nor some of the bizarre behaviors you've tweaked and/or documented. Then again, in the spirit of "small improvements" (and extended triage), it probably doesn't make sense to boil the entire ocean right here, either; so, fine, let's get this done and move on.

portante (approved):

PR #2613, opened 2022-01-27
Remove parent directory creation from Agent side

riya-17:


agent/base
```diff
@@ -12,24 +12,20 @@ set -o pipefail
 # very first thing to do is figure out which pbench we are
 if [[ -z "$pbench_run" ]]; then
     pbench_run=$(pbench-config pbench_run pbench-agent)
-    if [[ -z "$pbench_run" ]]; then
-        pbench_run=/var/lib/pbench-agent
-    fi
```
dbutenhof, line 17: This is unrelated to the change we'd discussed. In fact, Python has similar code, where the `PbenchAgentConfig` constructor supplies a hardcoded default if the `pbench_run` setting isn't defined in the config file. If we're working to improve consistency, we want this to stay.

And, now that I look at the Python more carefully ... `PbenchAgentConfig` goes on to `self.pbench_tmp.mkdir(parents=True, exist_ok=True)` so any Python agent command using `PbenchAgentConfig` is guaranteed to have the full `/var/lib/pbench_agent/tmp` created if it didn't already exist.

The `PbenchAgentConfig` however doesn't do anything about a user-supplied `pbench_run` override; the real difference was that, in *that* case, I believe the Python command code we looked at before would fail rather than creating it.

agent/base
```diff
@@ -12,24 +12,20 @@ set -o pipefail
...
-mkdir -p $pbench_tmp
+mkdir $pbench_tmp
 if [[ ! -d $pbench_tmp ]]; then
+  if [[ ! -d $pbench_run ]]; then
+    echo "[ERROR] the provided pbench run directory, ${pbench_run}, does not exist" >&2
+  else
     echo "[ERROR] unable to create TMP dir, ${pbench_tmp}" >&2
-    exit 1
+  fi
+  exit 1
```
dbutenhof, line 28: I think it's a more common pattern to check `${?}` after a command to see whether it worked. In this case, since we're requiring and assuming that the `pbench_run` directory already exists, I think we'd just want something simple like

```
mkdir ${pbench_tmp}
if [[ ${?} -ne 0 ]]; then
    echo "[ERROR] unable to create TMP directory, ${pbench_tmp}" >&2
    exit 1
fi
```

dbutenhof (changes requested):

agent/base
riya-17: I wanted to keep it simple. but if an environment path for `pbench_run` is passed and it is not already created then this will help us to know the reason behind that. But if we wanna rely on this msg then I will remove that check :)

agent/base
riya-17: Do we want to pass a hardcoded `pbench_run` path? I feel providing a hardcoded path might create problem in the future.

agent/base
dbutenhof: Fair enough; there's nothing wrong with this, really, and it probably is good to distinguish between the two cases. Using the `-d` check instead of testing `${?}` just looks a bit odd. I suppose you could combine the two by checking `pbench_run` existence only if the `mkdir` failed, but it's probably not important.

riya-17: Ok

npalaska (dismissed):

webbnh (approved):
I'm OK with removing the `-p` from the creation of the temporary directory (for all the reasons that Peter illuminated).  And, I concur that doing so removes the _need_ for the existence check on the `${pbench_run}` value.  But, I don't feel like relying on the error from `mkdir` is going to produce the best result when the run directory doesn't exist.

Hopefully, having a non-existent run directory is a very unusual case (like it "never happens"...), and hopefully the failure to create the temporary directory will deliver enough information to allow the user to address the problem when and if it does.

webbnh:
Houston, we have a problem, here:  it looks like the tests were run on one of the new executors (or agents, or whatever they are called) that Yinchuan set up, so the container never actually got started.  (@riya-17, this is not a problem with your code, but it is blocking the merge at the moment.)

riya-17:
Thank you @webbnh I was concerned about those failures :)

webbnh:
So, Yinchuan got the test running, but there are a lot of failures:   removing the `-p` not only means that we don't create the parents, but it also means that we now fail if the target directory already exists!  😀

So...why does the temp-dir already exist in those tests?  And, is this actually the "normal state", which our code needs to defend against?

(And, this branch needs to be rebased, again.)

webbnh:
It looks like the first time I run a `pbench-` command, the `tmp` directory gets created, and so it's "always there" on subsequent commands.  So, if we want to remove the `-p` we'll need to test for existence before trying to create it.

I think perhaps we should test for the existence of the parent directory and then do `mkdir -p` for the `tmp` directory:  that will enable us to report errors much better, e.g., when the _parent_ doesn't exist (and we won't care whether the `tmp` directory already exists).

dbutenhof (approved):

webbnh (approved):
👍

PR #2615, opened 2022-01-28
Correct the plumbing of stderr and stdout in certain unit tests

webbnh:
Several of our unit tests use a custom PyTest plug-in (aka a "[helper](https://pypi.org/project/pytest-helpers-namespace/)") called [`capture()`](https://github.com/distributed-system-analysis/pbench/blob/18b3814d82be22105caadab414fa92ea6d0f3981/lib/pbench/test/functional/conftest.py#L15).  This widget runs a command in a subprocess and returns the contents of `stdout` and `stderr` and the process exit code as a tuple.  As a brilliant example of why anonymous tuples are a bad idea, two of our unit tests contain invocations of the helper where the tuple is unpacked incorrectly and the identities of `stdout` and `stderr` are _reversed_.

This PR simply corrects the ordering of the unpacking, and swaps the references to the resulting streams, such that there is no functional change -- it's a simple swap of names.

In a subsequent PR (e.g,. #2612) we can debate whether and how any behavioral changes should be made.

### DIFF
--- a/lib/pbench/test/functional/agent/cli/commands/test_cleanup.py
+++ b/lib/pbench/test/functional/agent/cli/commands/test_cleanup.py
@@ -25,10 +25,10 @@ def test_cleanup(monkeypatch, agent_config, pbench_run, pbench_cfg):
 
     # test-64
     command = ["pbench-cleanup"]
-    err, out, exitcode = pytest.helpers.capture(command)
+    out, err, exitcode = pytest.helpers.capture(command)
     assert (
         b"pbench-cleanup deprecated, will be removed in future release in favor of pbench-clear-results\n"
-        in out
+        in err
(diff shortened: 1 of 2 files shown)


dbutenhof (approved):
Interesting; so some tests worked correctly with the incorrectly ordered returns, looking for stderr output in the `out` return. Head spinning. 🍤

npalaska (approved):

webbnh:
> so some tests worked correctly with the incorrectly ordered returns, looking for stderr output in the `out` return.

Yes:  they worked because they were looking for what they wanted in the wrong place and finding it there....

> Head spinning. 🍤

Welcome to where I spent much of the afternoon!  🍤

PR #2620, opened 2022-02-08
Fixes footer content overlap issue

vishalvvr:
Footer text content is overlapping the documentation text content.
To resolve it, just add margin-bottom equivalent to footer height

Old rendering:
Screenshot from 2022-02-08 09-15-09

New rendering:
Screenshot from 2022-02-08 09-15-37
.

### DIFF
--- a/doc.html
+++ b/doc.html
@@ -1012,5 +1012,8 @@
                 $(".searchBtn").show();
             }
         }
+
+        // add margin between body content and footer to avoid overlap
+        $('.section_card').css('margin-bottom', $('.footer').height());
     </script>
 </html>


webbnh (approved):
Not that I would know if this were bad or wrong, but, lacking other reviewers, it seems OK to me.  (@dbutenhof?)

dbutenhof (approved):
> Not that I would know if this were bad or wrong, but, lacking other reviewers, it seems OK to me. (@dbutenhof?)

Ha; I'd looked at it the other day, but without any particular indication that it was "ready to go" I didn't have any feedback. You're right, though; as small as this is, the sample output looks like an improvement and I don't see any harm.

doc.html
```diff
@@ -1012,5 +1012,8 @@
                 $(".searchBtn").show();
             }
         }
+
+        // add margin between body content and footer to avoid overlap
+        $('.section_card').css('margin-bottom', $('.footer').height());
```
npalaska, line 1017: Correct me if I am wrong and I don't really have a good understanding here but I have a naive question. So you added margin-bottom that is equal to the height of the footer but what if the document text content is bigger than that of the footer height would that result in overlap again?

vishalvvr: So current situation is exactly the same as you have asked in the question.
Footer is a sticky footer( [css](https://developer.mozilla.org/en-US/docs/Web/CSS/position) property) it won't move even if you scroll and document text(last few lines) is getting overlapped due to this behavior.

What we need is doc text should be display before the footer starts, so i have added margin-bottom (wide space) equal to footer height(this will add wide space after the document text ends). Also in future if footer content increases and result in increasing the footer size(height), this fix would adapt and adjust the rendering.

npalaska (approved):

PR #2621, opened 2022-02-09
Add option to allow chroot specs for COPR builds (b0.69)

ndokos:
Add a CHROOTS variable to the agent/rpm/Makefile and an option
to the `copr-cli build' command in rpm.mk, to allow
the specification of chroots to include or exclude in a build.

The default setting for the b0.69 branch excludes all RHEL9 builds:
the 0.69 agent does not work on RHEL9 because of an `scp -r' bug.

### DIFF
--- a/agent/rpm/Makefile
+++ b/agent/rpm/Makefile
@@ -8,6 +8,13 @@
 # 7. Optionally generate a local RPM.
 # 8. Clean up the temp directory
 
+# Do not use the RHEL9 chroots - v0.69 does not work on RHEL9
+CHROOTS = --exclude-chroot centos-stream-9-aarch64 \
+	  --exclude-chroot centos-stream-9-x86_64 \
+	  --exclude-chroot epel-9-aarch64 \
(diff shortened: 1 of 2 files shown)


webbnh (approved):
Looks OK to me (although, I would have preferred that we use `$CHROOTS` to specify the platforms that we **do** support rather than to list the ones which we *don't* support (at this moment), but I guess that this will work).

However, there is a failing test, which looks like it might be due to a non-deterministic output ordering?  (Either that, or someone slipped in a prior change without a gold file update....)  I'm not sure what we should be doing about this....

dbutenhof (approved):
Given it's a "Make-do" (pardon the pun) for 0.69, this is OK. I think the label `CHROOTS` is a bit misleading for a value that needs to contain the `copr-cli` options to select/exclude chroots (and in fact can contain any `copr-cli` options)... it's not perfect, (or elegant), but it's probably good enough for now. 👍🏻

npalaska (approved):

portante (approved):

PR #2622, opened 2022-02-09
Add option to allow chroot specs for COPR builds

ndokos:
Add a CHROOTS variable to the agent/rpm/Makefile and an option
to the `copr-cli build' command in ../../utils/rpm.mk, to allow
the specification of chroots to include or exclude in a build.

The default setting for the main branch builds all chroots configured
in the COPR repo.

### DIFF
--- a/agent/rpm/Makefile
+++ b/agent/rpm/Makefile
@@ -1,6 +1,25 @@
 # Makefile for generating a source RPM and optional local RPM
 # for the Pbench agent.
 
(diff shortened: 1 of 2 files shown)


webbnh (approved):
I'm OK with this change, but it neglects `server/rpm/Makefile`...but, perhaps that's OK -- I believe that by default Make will expand undefined variables vacuously, so the Server will build everywhere, just like the client does, only without explanation.  😉

portante (approved):

PR #2631, opened 2022-02-17
Use --all-ip-addresses instead of --ip-address

### LINKED ISSUE #2608: `hostname --ip-address` is problematic
```
       -i, --ip-address
              Display  the  network  address(es)  of the host name. Note that this works only if the host name can be resolved. Avoid using this option; use
              hostname --all-ip-addresses instead.

       -I, --all-ip-addresses
              Display all network addresses of the host. This option enumerates all configured addresses on all network interfaces. The  loopback  interface
              and  IPv6 link-local addresses are omitted. Contrary to option -i, this option does not depend on name resolution. Do not make any assumptions
              about the order of the output.
```

Attempting to fire up a remote tool session from `pbench-user-benchmark` from my laptop to an agent on my RDU2 server VM fails in `/opt/pbench-agent/base` validation:

```
root 14:28 ~: pbench-user-benchmark -C multi --iteration-list=iterations1.data -- sleep 10
[error][2022-01-24T19:28:14.679631799] [] Invalid IP address: 'fe80::0:c8f0%eth0'
pbench-tool-meister-start: failed to start tool meister on remote host '10.1.1.1' (pid 2856574), exit status: 1
pbench-tool-meister-start: terminating tool meister startup due to failures
[error][2022-01-24T19:28:21.252016312] [pbench-user-benchmark]: failed to start the tool meisters.
```

This is because on the lab VM, `hostname --ip-address` returns

```
[root@dbutenho-dev ~]# hostname --ip-address
fe80::0:c8f0%eth0 2620:0:f1de:aa45 10.1.1.1
```

The `validate_ipaddress` tool, which uses the Python `ipaddress` package to parse each item in the returned list, cannot support the `%eth0` scope ID, so it fails the Tool Meister remote setup. (Without very good messaging.)

We should `hostname --all-ip-addresses` instead, because (as noted above in the man page excerpt) the link-local address is inhibited. (Though that's only one use for a scope ID; there may be real addresses that are meaningful only on a specific NIC, and dealing with *that* can get very complicated since we'd need to do pings to find an address that's reachable on a subnet the controller can reach.) However, for now, and for most normal cases, this simple substitution would solve the problem.

```
[root@dbutenho-dev ~]# hostname --all-ip-addresses
10.1.1.1 2620:0:aa45 
```

(Note that I munged all the IP addresses for safety.)

dbutenhof:
The `hostname --ip-addresses` returns loopback and link-local addresses, which are not useful for agent communication and break the `ipaddress.ip_address()` parser. Instead, use `--all-ip-addresses` which provides similar results but corrects these weaknesses and does not depend on name resolution.

```
       -i, --ip-address
              Display  the  network  address(es)  of the host name. Note that this works only if the host name can be resolved. Avoid using this option; use
              hostname --all-ip-addresses instead.

       -I, --all-ip-addresses
              Display all network addresses of the host. This option enumerates all configured addresses on all network interfaces. The  loopback  interface
              and  IPv6 link-local addresses are omitted. Contrary to option -i, this option does not depend on name resolution. Do not make any assumptions
              about the order of the output.
```

Resolves #2608

### DIFF
--- a/agent/base
+++ b/agent/base
@@ -119,7 +119,7 @@ fi
 validate_hostname "${_pbench_full_hostname}" "_pbench_full_hostname"
 
 if [[ -z "${_pbench_hostname_ip}" ]]; then
-    export _pbench_hostname_ip=$(hostname --ip-address)
+    export _pbench_hostname_ip=$(hostname --all-ip-addresses)
 fi
 for _ip in ${_pbench_hostname_ip}; do
     validate-ipaddress "${_ip}"


webbnh (approved):
👍

portante (approved):

PR #2637, opened 2022-02-22
Sort all items in the tarball stats report

### LINKED ISSUE #2623: The b0.69 branch is failing its tests
With the recent merge (which has nothing to do with the current problem), the Jenkins tests are [now failing](http://jenkins.intlab.perf-infra.lab.eng.rdu2.redhat.com/job/pbench-ci-multi/job/b0.69/18/console).

Peter indicated that the root cause is a `find(1)` command in the audit tool whose output should be sorted in order to produce a deterministic result.

webbnh:
Currently, the v0.69 branch testing is failing due to a non-determinism in the `server/bin/pbench-tarball-stats.py` output.  For parts of the report, the lists of servers are sorted; and, for other parts they are not.  This change adds the missing sort directives to the output template.

Fixes #2623.

### DIFF
--- a/server/bin/gold/test-5.2.txt
+++ b/server/bin/gold/test-5.2.txt
@@ -356,13 +356,13 @@ Summary Statistics for Tar Balls on 1970-01-01 (external data only):
                     1.8 kB size
 
         By Server Origin:
-                         5 main pbench server
                          3 "ONE" pbench satellite server
+                         5 main pbench server
 
         Controller Counts:
                          6 controllers
-                         3 main pbench server
                          3 "ONE" pbench satellite server
+                         3 main pbench server
 
 Tar Ball Counts broken down by weeks for most recent 2 months (with satellite percentages):
 
--- a/server/bin/pbench-tarball-stats.py
+++ b/server/bin/pbench-tarball-stats.py
@@ -173,13 +173,13 @@ report_tmpl = """Summary Statistics for Tar Balls on {{ now.strftime("%Y-%m-%d")
         {{ "{:>18s}".format(good.size|humanize_naturalsize) }} size
 
         By Server Origin:
-        {% for name, value in server_origin.items() %}
+        {% for name, value in server_origin.items()|sort %}
         {{ "{:18n}".format(value) }} {% if name == "main" %}{{ name }} pbench server{% else %}"{{ name }}" pbench satellite server{% endif +%}
         {% endfor %}
 
         Controller Counts:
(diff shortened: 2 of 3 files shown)


dbutenhof (dismissed):

dbutenhof (dismissed):
My first reaction is, "oh, no, another revlock" ... however, realistically, we need real configuration management which means *everything* is revlocked to not just a range but a specific version, preferably from a supported RPM. So, yeah. Cool. Third time's a charm?

dbutenhof:
Oops: Or,

```
ERROR: Invalid requirement: 'humanize>=3.11 <4.0.0' (from line 3 of /home/pbench/server/requirements.txt)
```

webbnh:
> "oh, no, another revlock"

Yeah, well, adding the rev-lock actually has nothing to do with the problem I'm fixing...it's just that when I went to test my change, it suddenly failed in a **_new_** way...because, apparently, the build picked up a new version of the `humanize` package which had recently rolled from v3.11 to v4.0 and put in [a breaking change](https://github.com/jmoiron/humanize/commit/2482af79f3367d6ddbfb68edd918a9e358f09cee).  And, of course, the breakage isn't in _our_ code -- it's in `jinja2-humanize-extension` which we depend upon -- but we inherit the problem.  (I [reported the issue](https://github.com/metwork-framework/jinja2_humanize_extension/issues/4), but I'm dubious as to whether it will be fixed -- `jinja2-humanize-extension` seems to live in a very quiet repo.)  So, I added the rev-lock (and, as you noticed, it took two tries...).

And this enabled me to see that the resulting sort comes out with the opposite order from what I expected, so I have yet another update with the fixed gold file.  😛

ndokos (approved):

thebaptiste:
The repo jinja2-humanize-extension is not so quiet...
We have fixed the issue requiring humanize<4.0.0 in release 0.2.1
We will try to phase with humanize 4.0.0 later
Thanks for the issue

dbutenhof (approved):

webbnh:
Hi @thebaptiste!  Thanks for commenting here!!

I apologize for my earlier comment -- clearly my doubts were misplaced:  the fact that my issue had a single-digit number must be an indication of the quality of the work on `jinja2-humanize-extension` rather than any lack of attentiveness from its maintainers or users.  😀

(I'm quite amused by the fact that my referencing it in this repo added a mention on the Issue in your repo...it didn't occur to me that GitHub provides that level of integration.  😄)

PR #2639, opened 2022-02-23
Improve Jenkins container run mechanism

dbutenhof:
There are several small related issues here; one is that we currently require `--user=root` to run reliably, but we really don't actually "need" to be running our tests as root. This turns out to be mostly required because of the second issue, and with this change is no longer  required.

The main issue is that the container inherits the `HOME` environment variable; for example, in Jenkins, typically `/home/jenkins`. This value is used for `pip install --user` and to install our package with PBR for testing. Because `/home/jenkins` doesn't exist in the container, we require `root` to create the path. Instead, we'll now override the definition of `HOME` in order to redirect the `--user` to `/tmp/jenkins` which we can create without privilege.

In conjunction, this removes the default definition of `EXTRA_PODMAN_SWITCHES` from Pipeline.gy.

There's another "convenience" feature in here useful for local manual testing using the container from git worktrees; because a worktree is a mapping of a root branch, we need to map both the worktree directory and the referenced branch directory inside the container. This uses the `git rev-parse` command to figure out the environment and add a second
podman `--volume` option if necessary.

### DIFF
--- a/jenkins/python-setup.sh
+++ b/jenkins/python-setup.sh
@@ -17,6 +17,12 @@
 # chosen over any installed ones, and it adds `/usr/sbin` to the end of the
 # path for the `ip` command, used by pbench-register-tool.
 
+# Override the inherited HOME (which is a path generally not mapped inside the
+# container) so that pip and setup.py will install to a directory that exists
+# and doesn't require root access (e.g., to create /home/jenkins/.local/)
+export HOME=/tmp/jenkins
+mkdir -p ${HOME}
+
 PATH=$(python3 -m site --user-base)/bin:${PATH}:/usr/sbin
 unset PYTHONPATH
 pip3 install --user -r lint-requirements.txt -r docs/requirements.txt -r agent/requirements.txt -r server/requirements.txt -r agent/test-requirements.txt -r server/test-requirements.txt
--- a/jenkins/python-setup.sh
+++ b/jenkins/python-setup.sh
@@ -17,12 +17,6 @@
 # chosen over any installed ones, and it adds `/usr/sbin` to the end of the
 # path for the `ip` command, used by pbench-register-tool.
 
-# Override the inherited HOME (which is a path generally not mapped inside the
-# container) so that pip and setup.py will install to a directory that exists
-# and doesn't require root access (e.g., to create /home/jenkins/.local/)
-export HOME=/tmp/jenkins
-mkdir -p ${HOME}
-
 PATH=$(python3 -m site --user-base)/bin:${PATH}:/usr/sbin
 unset PYTHONPATH
 pip3 install --user -r lint-requirements.txt -r docs/requirements.txt -r agent/requirements.txt -r server/requirements.txt -r agent/test-requirements.txt -r server/test-requirements.txt
--- a/jenkins/run
+++ b/jenkins/run
@@ -33,6 +33,7 @@ podman run \
     ${GIT_BASE_VOLUME} \
     -w /src/pbench \
     --env-host \
+    --env HOME=/tmp/jenkins \
     --ulimit nofile=65536:65536 \
     --rm \
     ${EXTRA_PODMAN_SWITCHES} \
--- a/jenkins/run
+++ b/jenkins/run
@@ -1,5 +1,22 @@
 #!/bin/bash
 
+# Run commands within the pbench-devel container image.
+#
+# Generally you want to run a shell, and `source jenkins/python-setup.py` to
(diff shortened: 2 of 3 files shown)


jenkins/python-setup.sh
```diff
@@ -17,6 +17,12 @@
 # chosen over any installed ones, and it adds `/usr/sbin` to the end of the
 # path for the `ip` command, used by pbench-register-tool.
 
+# Override the inherited HOME (which is a path generally not mapped inside the
+# container) so that pip and setup.py will install to a directory that exists
+# and doesn't require root access (e.g., to create /home/jenkins/.local/)
+export HOME=/tmp/jenkins
+mkdir -p ${HOME}
+
```
webbnh, line 25: To use Peter's phrase, I thinking that adding this here violates the "separation of concerns":  this script is intended to run `python setup` and related stuff, and it could be used outside the Jenkins environment...changing `HOME` inside it limits its applicability (or, is likely to surprise someone who uses it directly).

It is easy enough to set `HOME` on the `podman` command in `jenkins/run`.  Creating the directory is trickier, but we could add it to the container recipe so it's always there.  (I have one other point change that I'd like to add to the recipe, so we can easily do this if you're amenable.)

webbnh (changes requested):
I would prefer that we use a different approach to setting and creating the `HOME` directory.  Can we collaborate on that?

webbnh (approved):
Looks good.  It'll be even better when you add the volume-mapping for `$HOME`.

jenkins/run
```diff
@@ -16,12 +33,33 @@ if [[ -z "${IMAGE_REPO}" ]]; then
     IMAGE_REPO="quay.io/pbench"
 fi
 
+IHOME=/container/home
+
+if [[ -z "${CONTAINER_HOME}" ]]; then
+    HOME_VOLUME=${IHOME}
+else
+    HOME_VOLUME="${CONTAINER_HOME}:${IHOME}:${CONTAINER_HOME_OPTION:="z"}"
```
webbnh, line 41: I would use `:-` rather than `:=` for `CONTAINER_HOME_OPTION`, but I guess either will work, here.

webbnh (approved):
Nice!!

jenkins/run
dbutenhof: Well, shoot, I didn't think of that.

Technically someone could have `export CONTAINER_HOME_OPTION=` and now it'd come back as "z" instead of just affecting the expansion within `run`. Not a particularly realistic scenario, but I guess you're right.

webbnh: > someone could have `export CONTAINER_HOME_OPTION=` and now it'd come back as "z" instead of just affecting the expansion within run

Assuming that `jenkins/run` is invoked as a command, it will be run in a subshell, with its own environment.  If line 41 is executed, then after that point, `CONTAINER_HOME_OPTION` will be defined (regardless of whether it was before or not), and it will either still have its original value or it will have the value `"z"` if it was previously null or unset.  But, as we're in a subprocess where we control the code, probably no one will notice or care.

Now, inside the container, which we've set to inherit the parent process environment, `CONTAINER_HOME_OPTION` will be set (to `"z"`, if nothing else) any time `CONTAINER_HOME` is defined.  I don't know if that's useful or not, but there it is.  Switching from `:=` to `:-` would mean that `CONTAINER_HOME_OPTION` would only be defined inside the container when it was explicitly defined outside the container...I don't know if that's better or not.

BTW, I don't think that setting `CONTAINER_HOME_OPTION` to an empty value makes the `:=` expansion behave differently from the way it does when the variable is unset -- either way I think the variable ends up set to `"z"`...and perhaps that's a "feature".

ndokos (approved):

jenkins/run
dbutenhof: Yes, you're correct that the value won't be propagated back out of the script. I got carried away, there. 😆 

For the last paragraph; right, but my point was to contrast `-` and `:-` with respect to a variable that's *defined* but blank. (That is, `:-` will set `CONTAINER_HOME_OPTION=''` to `z` while `-` won't. Similarly for `:=` and `=`.)

webbnh: I never noticed before the sentence at the end of the paragraph in the man page which says, "omitting the colon results in a test only for a parameter that is unset."  That's good to know....

> That is, `:-` will set `CONTAINER_HOME_OPTION=''` to `z` while `-` won't.

I agree iff you change "set" to "evaluate" (but, I concur with your assertion about `:=` and `=`).

PR #2642, opened 2022-02-24
Fix `user-tool stop`

### LINKED ISSUE #2614: The `user-tool stop` is incorrect
```
[root@perf144 user-tool-tooltest]# diff -cw /opt/pbench-agent/tool-scripts/base-tool.orig /opt/pbench-agent/tool-scripts/base-tool
*** /opt/pbench-agent/tool-scripts/base-tool.orig       2022-01-27 21:43:26.826737966 -0500
--- /opt/pbench-agent/tool-scripts/base-tool    2022-01-27 22:17:24.725907434 -0500
***************
*** 178,184 ****
        postprocess_script=""
        start_script=""
        stop_script=""
!       tool_name=""
        longoptions="${longoptions},file-to-capture:,postprocess-script:,start-script:,stop-script:,tool-name:"
        ;;
  virsh-migrate)
--- 178,184 ----
        postprocess_script=""
        start_script=""
        stop_script=""
!       tool_name_arg=""
        longoptions="${longoptions},file-to-capture:,postprocess-script:,start-script:,stop-script:,tool-name:"
        ;;
  virsh-migrate)
***************
*** 463,469 ****
                ;;
        --tool-name)
                if [[ -n "${1}" ]]; then
!                       tool_name="${1}"
                        shift
                fi
                ;;
--- 463,469 ----
                ;;
        --tool-name)
                if [[ -n "${1}" ]]; then
!                       tool_name_arg="${1}"
                        shift
                fi
                ;;
***************
*** 496,502 ****
  fi
 
  if [[ "${tool}" == "user-tool" ]]; then
!       if [[ -z "${tool_name}" ]]; then
                printf -- "%s: You must provide a value for --tool-name\n" "${tool}" >&2
                usage >&2
                exit 1
--- 496,502 ----
  fi
 
  if [[ "${tool}" == "user-tool" ]]; then
!       if [[ -z "${tool_name_arg}" ]]; then
                printf -- "%s: You must provide a value for --tool-name\n" "${tool}" >&2
                usage >&2
                exit 1
***************
*** 506,512 ****
                usage >&2
                exit 1
        fi
!       tool_name="${tool}-${tool_name}"
  else
        tool_name="${tool}"
  fi
--- 506,512 ----
                usage >&2
                exit 1
        fi
!       tool_name="${tool}-${tool_name_arg}"
  else
        tool_name="${tool}"
  fi
***************
*** 771,777 ****
        if [[ -s "${tool_pid_file}" ]]; then
                pid="$(cat ${tool_pid_file})"
                if [[ ! -z "${pid}" ]] ;then
!                       case "${tool_name}" in
                        kvmtrace)
                                # kill the sleep process if it still exists
                                sleep_pid="$(ps --ppid ${pid} | pgrep sleep)"
--- 771,777 ----
        if [[ -s "${tool_pid_file}" ]]; then
                pid="$(cat ${tool_pid_file})"
                if [[ ! -z "${pid}" ]] ;then
!                       case "${tool}" in
                        kvmtrace)
                                # kill the sleep process if it still exists
                                sleep_pid="$(ps --ppid ${pid} | pgrep sleep)"
***************
*** 788,794 ****
                                rc=0
                                ;;
                        perf)
!                               safe_kill "${tool_name}" "${pid}" "INT"
                                # Wait for perf to finish recording.  if you
                                # do not wait, 'perf report' will not be
                                # correct.  perf is not a child process, so we
--- 788,794 ----
                                rc=0
                                ;;
                        perf)
!                               safe_kill "${tool}" "${pid}" "INT"
                                # Wait for perf to finish recording.  if you
                                # do not wait, 'perf report' will not be
                                # correct.  perf is not a child process, so we
***************
*** 809,815 ****
                                rc=${?}
                                ;;
                        blktrace|bpftrace|iostat|mpstat|pcp|sar|systemtap|tcpdump|vmstat)
!                               safe_kill "${tool_name}" "${pid}"
                                rc=${?}
                                ;;
                        proc-interrupts|proc-sched_debug|proc-vmstat)
--- 809,815 ----
                                rc=${?}
                                ;;
                        blktrace|bpftrace|iostat|mpstat|pcp|sar|systemtap|tcpdump|vmstat)
!                               safe_kill "${tool}" "${pid}"
                                rc=${?}
                                ;;
                        proc-interrupts|proc-sched_debug|proc-vmstat)
***************
*** 817,823 ****
                                rc=${?}
                                ;;
                        *)
!                               safe_kill "${tool_name}-datalog" "${pid}"
                                rc=${?}
                                ;;
                        esac
--- 817,823 ----
                                rc=${?}
                                ;;
                        *)
!                               safe_kill "${tool}-datalog" "${pid}"
                                rc=${?}
                                ;;
                        esac
```

webbnh:
Correct references to `$tool` and `$tool_name` in the `base_tool` script, which has the side effect of fixing `user-tool stop`.

In most cases, the changes in this commit have no effect, since `$tool_name` typically has the same value as `$tool`; however, this is not the case when `$tool` is `user-tool`, and this breaks the `case` dispatcher in the `stop` subcommand.  Changing the reference from `$tool_name` (which may include the user tool name as a suffix) to `$tool` (which is always the name of the tool) corrects this.

The rest of the changes have little effect, other than making the code use the "right" variable.

Fixes #2614.

### DIFF
--- a/agent/tool-scripts/base-tool
+++ b/agent/tool-scripts/base-tool
@@ -176,7 +176,7 @@ user-tool)
 	postprocess_script=""
 	start_script=""
 	stop_script=""
-	tool_name=""
+	tool_name_arg=""
 	longoptions="${longoptions},file-to-capture:,postprocess-script:,start-script:,stop-script:,tool-name:"
 	;;
 virsh-migrate)
@@ -464,7 +464,7 @@ while true; do
 		;;
 	--tool-name)
 		if [[ -n "${1}" ]]; then
-			tool_name="${1}"
+			tool_name_arg="${1}"
 			shift
 		fi
 		;;
@@ -497,7 +497,7 @@ if [[ "${mode}" != "install" && ( -z "${tool_dir}" || ! -d "${tool_dir}" ) ]]; t
 fi
 
 if [[ "${tool}" == "user-tool" ]]; then
-	if [[ -z "${tool_name}" ]]; then
+	if [[ -z "${tool_name_arg}" ]]; then
 		printf -- "%s: You must provide a value for --tool-name\n" "${tool}" >&2
 		usage >&2
 		exit 1
@@ -507,7 +507,7 @@ if [[ "${tool}" == "user-tool" ]]; then
 		usage >&2
 		exit 1
 	fi
-	tool_name="${tool}-${tool_name}"
+	tool_name="${tool}-${tool_name_arg}"
 else
 	tool_name="${tool}"
 fi
@@ -708,7 +708,7 @@ start)
 	mkdir -p "${tool_output_dir}"
 	pushd "${tool_output_dir}" >/dev/null
 	if [[ ${?} -ne 0 ]]; then
-		printf -- "%s: failed to create tool output directory, %s\n" "${tool_name}" "${tool_output_dir}" >&2
+		printf -- "%s: failed to create tool output directory, %s\n" "${tool}" "${tool_output_dir}" >&2
 		exit 1
 	fi
 
@@ -720,14 +720,14 @@ start)
 	tool_cmd_file="./${tool_name}.cmd"
 	if [[ -z "${gopath}" ]]; then
 		if [[ "${tool_package_name}" == "golang" ]]; then
-			printf -- "%s: golang based tools require the --gopath option\n" "${tool_name}" >&2
+			printf -- "%s: golang based tools require the --gopath option\n" "${tool}" >&2
 			popd > /dev/null
 			exit 1
 		fi
 		_gopath=""
 	else
 		if [[ ! -d "${gopath}/bin" ]]; then
-			printf -- "%s: invalid GOPATH directory specified, '%s'\n" "${tool_name}" "${gopath}" >&2
+			printf -- "%s: invalid GOPATH directory specified, '%s'\n" "${tool}" "${gopath}" >&2
 			popd > /dev/null
 			exit 1
 		fi
@@ -738,7 +738,7 @@ start)
 		_goroot=""
 	else
 		if [[ ! -x "${goroot}/bin/go" ]]; then
-			printf -- "%s: invalid GOROOT directory specified, '%s'\n" "${tool_name}" "${goroot}" >&2
+			printf -- "%s: invalid GOROOT directory specified, '%s'\n" "${tool}" "${goroot}" >&2
 			popd > /dev/null
 			exit 1
 		fi
@@ -759,7 +759,7 @@ start)
 	# Force LANG=C for all tools
 	printf -- "#!/bin/bash\n# base-tool generated command file\n\nLANG=C PYTHONUNBUFFERED=True%s%s%s exec %s\n" "${_goroot}" "${_gopath}" "${_newpath}" "${tool_cmd}" > "${tool_cmd_file}"
 	chmod +x "${tool_cmd_file}"
-	printf -- "%s: running \"%s\"\n" "${tool_name}" "${tool_cmd}"
+	printf -- "%s: running \"%s\"\n" "${tool}" "${tool_cmd}"
 	tool_stdout_file="./${tool_name}-stdout.txt"
 	tool_stderr_file="./${tool_name}-stderr.txt"
 	${tool_cmd_file} > "${tool_stdout_file}" 2> "${tool_stderr_file}" & echo ${!} > "./${tool_name}.pid"
@@ -769,16 +769,16 @@ start)
 	;;
 stop)
 	if [[ ! -d "${tool_output_dir}" ]]; then
-		printf -- "%s: WARNING - stop - tool output directory, '%s', missing\n" "${tool_name}" "${tool_output_dir}" >&2
+		printf -- "%s: WARNING - stop - tool output directory, '%s', missing\n" "${tool}" "${tool_output_dir}" >&2
 		exit 1
 	fi
 	enable -n kill
-	printf -- "%s: stopping\n" "${tool_name}"
+	printf -- "%s: stopping\n" "${tool}"
 	tool_pid_file="${tool_output_dir}/${tool_name}.pid"
 	if [[ -s "${tool_pid_file}" ]]; then
 		pid="$(cat ${tool_pid_file})"
 		if [[ ! -z "${pid}" ]] ;then
-			case "${tool_name}" in
+			case "${tool}" in
 			kvmtrace)
 				# kill the sleep process if it still exists
 				sleep_pid="$(ps --ppid ${pid} | pgrep sleep)"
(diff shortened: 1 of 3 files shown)


agent/tool-scripts/base-tool
```diff
@@ -759,7 +759,7 @@ start)
 	# Force LANG=C for all tools
 	printf -- "#!/bin/bash\n# base-tool generated command file\n\nLANG=C PYTHONUNBUFFERED=True%s%s%s exec %s\n" "${_goroot}" "${_gopath}" "${_newpath}" "${tool_cmd}" > "${tool_cmd_file}"
 	chmod +x "${tool_cmd_file}"
-	printf -- "%s: running \"%s\"\n" "${tool_name}" "${tool_cmd}"
+	printf -- "%s: running \"%s\"\n" "${tool}" "${tool_cmd}"
```
dbutenhof, line 762: Hmm; you missed some gold files. At the same time, I wonder about the change in the prefix; do we really want to be identifying the "root" `${tool}`, which is generic, vs the full `${tool_name}` here?

```
*** 1,4 ****
! user-tool-foo42: running "/var/tmp/pbench-test-tool-scripts/user-tool/tmp/foo42.start /var/tmp/pbench-test-tool-scripts/user-tool/tools-group/user-tool-foo42 10"
! user-tool-foo42: stopping
! user-tool-foo42: post-processing following stop
! user-tool-foo42: post-processing data
--- 1,4 ----
! user-tool: running "/var/tmp/pbench-test-tool-scripts/user-tool/tmp/foo42.start /var/tmp/pbench-test-tool-scripts/user-tool/tools-group/user-tool-foo42 10"
! user-tool: stopping
! user-tool: post-processing following stop
! user-tool: post-processing data
```

webbnh: (Sorry about the missing gold files!)

>I wonder about the change in the prefix

While this is certainly open for discussion, this change was intentional:  the "tool" which the user invoked is `user-tool`, and so it seems like it should identify itself as such.  Now, it's true that `user-tool` is actually used to run another tool, so...it's not crazy to use the combined prefix, but it _is_ a little weird.  (In particular, it's not quite right in the first line of the output that you quoted, since that's describing `user-tool`'s invocation of the user's tool; and, if we go with the straightforward prefix for that one, then the others look strange, especially since they are describing actions by `user-tool` itself -- if the sub-tool reports something, it can tag it with it's own name....)

Now, if someone points out that a user could have a single Pbench run with multiple tools being run by separate invocations of `user-tool`, I might be swayed against this change, but I think the outputs go to separate log files, in which case any ambiguity created by the common prefix should be easy to dispel.

(I suppose, we might compromise by using a different format for the prefix (e.g. `user-tool(foo42): `) which would be clearer (i.e., there _is_ no `user-tool-foo42` command or script, but there _was_ an invocation of `user-tool` with the **argument** `foo42`).)

portante: Since `tool_name` is used to form the "tool output directory", and all remaining uses of `tool_name` after this change refer to files in the `tool_output_directory`, it seems we can just drop all other uses of it entirely.

portante (changes requested):
Could we have one commit with the original diff, and the additional changes on top?  The original patch posted in issue #2614 has been verified on a live system and in use currently.

agent/tool-scripts/base-tool
dbutenhof: I'm not sure `user-tool` alone is actually of much use to anyone (it can't be used and has no meaning without `--tool-name`)... none of the other tools linked to this script have that problem.

I like the `user-tool(foo42)` idea, though, instead of implying a tool path that doesn't exist.

webbnh:
> Could we have one commit with the original diff, and the additional changes on top?

I think you mean, could I please start over?...

Are you expecting the two commits to be merged as separate commits to `main`, or are they going to end up squashed anyway...because if separating them is just for review purposes, this review is not that hard, and I'm pretty sure that separating them will be more trouble than it's worth.

> The original patch posted in issue #2614 has been verified on a live system and in use currently.

In that case, @portante, why did you post it as an Issue instead of as a PR??  (And, why did you post it as a big ol' blob of text instead of as a patch file?)

agent/tool-scripts/base-tool
webbnh: > Since `tool_name` is used to form the "tool output directory", and all remaining uses of `tool_name` after this change refer to files in the `tool_output_directory`, it seems we can just drop all other uses of it entirely.

I'm not sure what you're suggesting, here, @portante....  With my change, the only references to `tool_name` are in the context of constructing various filenames; however, not all of them are (obviously) in `tool_output_directory` (so I cannot, e.g., simply replace uses of `tool_name` with `tool_output_directory`), so I think we still need to use `tool_name` in the places where I haven't replaced it with `tool`.  Or, are you saying something else?

webbnh: > I like the `user-tool(foo42)` idea, though, instead of implying a tool path that doesn't exist.

The principal problem there is that there are points in the script where we know we're running `user-tool` but don't yet know that we're using it to run `foo42`...so, if I go with the decorated approach, many of the messages will have the decoration but more than a quarter of them won't...which feels weird.

webbnh:
> Could we have one commit with the original diff

The original diff appears to me to show three sets of changes:
- there's the bugfix (which I think is a one-liner)
- there's the clean-up which employs `tool_name_arg` as a replacement for early uses of `tool_name`
- there's later clean-up which replaces uses of `tool_name` with `tool` in contexts where the two contain the same value (and there's also one other change which falls into this bucket once the bugfix is in place).

So, the changes that I added are just more of the same stuff.

portante:
> > Could we have one commit with the original diff, and the additional changes on top?
> 
> I think you mean, could I please start over?...

Ah, no.

You can insert a commit before an existing one to isolate changes.

> 
> Are you expecting the two commits to be merged as separate commits to `main`, or are they going to end up squashed anyway...because if separating them is just for review purposes, this review is not that hard, and I'm pretty sure that separating them will be more trouble than it's worth.

Perhaps.  But the key is the bug fix does not require all the changes being made here.  So having the bug fix separate from the rest of the changes being made seems worth-while to do.

> 
> > The original patch posted in issue #2614 has been verified on a live system and in use currently.
> 
> In that case, @portante, why did you post it as an Issue instead of as a PR?? (And, why did you post it as a big ol' blob of text instead of as a patch file?)

For a number of reasons: 1. at the time I was working on a higher priority set of work when this came up, 2. the person who reported the problem had showed me the beginnings of a fix, but it was not correct and reviewing it resulted in that patch, 3. the work it would take to make a proper PR for review would have had taken more time to do than was available at that time.

The issue report was quick-n-dirty for sure.  Sorry.

portante:
Here is an example branch with the simple tool fix applied before all the other changes, and then the full removal of the `tool_name` variable since it is unnecessary with all the output in a unique directory.

https://github.com/portante/pbench/commits/user-tool

webbnh:
> Here is an example branch with the simple tool fix applied before all the other changes, and then the full removal of the `tool_name` variable since it is unnecessary with all the output in a unique directory.
> 
> https://github.com/portante/pbench/commits/user-tool

If I might suggest...why don't you put up a PR for that?  Then I can close this PR, and we can go approve that one instead.  (If I incorporate your work here that would potentially disqualify you as a reviewer....  😁)

portante:
Replaced by #2650.

PR #2643, opened 2022-02-25
Support non-DNS controller hosts

### LINKED ISSUE #2609: Agent python installation failure
Although it's hard to be sure exactly why, the current Agent spec file's Python installation fails on my RHEL 8.4 server VM, resulting in no Python modules being installed and no useful messaging in `/opt/pbench-agent/pip3-install.log`. 

I experimentally replaced the `pip3` install command in the spec file with something more akin to what I put in the server spec file:

```
python3 -m pip install --upgrade pip 2>&1 > /%{installdir}/pip3-upgrade.log
python3 -m pip --no-cache-dir install --prefix=/%{installdir} -r /%{installdir}/requirements.txt > /%{installdir}/pip3-install.log 2>&1
```

... and this seems to work just fine.

The agent also has some "pip3 bootstrap" code in front of this for RHEL 7, which might possibly be unnecessary assuming that the RHEL 7 Python3 at least has a functional `pip` module even if `pip3` doesn't work for us.

### LINKED ISSUE #2610: Tool Meister on RHEL CSB laptop
This is low priority, and probably not worth worrying about for 0.71 (if ever), but I'm recording it so it doesn't get lost. (Although I'll stick it in the "backlog" project, which is "next to lost".)

I've been trying multi-node pbench agent collection, and I tried to go from my laptop (which already had an agent installed) to my RDU2 pbench server VM. Because CSB networking is weird (the word "weird" doesn't do it justice, but we'll go with that for now), the Redis server ends up listening for a connection to "dbutenho@bos.csb", an address that nobody can talk to.

Interestingly, `hostname --all-ip-addresses` on my RHEL 8.4 CSB laptop returns three IPv4 addresses, only one of which is actually reachable. It happens to be the last of the three, though I have no idea whether that's architecturally guaranteed or sheer coincidence, in which case pragmatically we might need to try remotely pinging all the addresses to find if one responds.

We have the same problem on RDU2 VMs, where the configured hostnames are also non-routable (dbutenho-dev, rhel8-pbenchoctjam).

Peter and I did some experimentation and it seems there's a relatively straightforward way around these problems during TM / TDS setup by tracking the origin address exposed by an `ssh` to each remote system.

dbutenhof:
By default, TDS and Redis are started listening on (and remote hosts are set to connect to) the controller node's `hostname`. If this is not a routable DNS name, Tool Meister start hangs.

Instead, identify a routable IP address for the host by determining the origin of ssh connections to the remote hosts. We need to establish that the remotes are reachable anyway; this allows us to do so early in startup before creating the local TDS and Redis instances.

If "orchestration=create" mode isn't selected, or if `PBENCH_REDIS_SERVER` (`--redis-server`) or `PBENCH_TOOL_DATA_SINK` (`--too-data-sink`) configuration options are specified, these values are used regardless, trusting the external input.

Resolves #2610

### DIFF
--- a/agent/rpm/pbench-agent.spec.j2
+++ b/agent/rpm/pbench-agent.spec.j2
@@ -88,24 +88,8 @@ cp -a agent/* %{?buildroot}/%{installdir}/
 # configtools installed: zap it.
 if pip3 show configtools > /dev/null 2>&1 ;then pip3 uninstall -y configtools ;fi
 
-%post
-%if 0%{?rhel} == 7
-# This is a very ugly work-around for RHEL 7 / CentOS 7 environments. The
-# version of `pip3` that is available in those environments no longer works
-# to install the latest Python3 modules, which we need. So we have to fetch
-# our own copy of `pip3`, as described at:
-#   https://pip.pypa.io/en/latest/installing/#installing-with-get-pip-py 
-# We then install it separately from the existing version and set `PATH` and
-# `PYTHONPATH` so that we use this updated version when we install our module
-# dependencies below.
-curl -X GET -o /%{installdir}/get-pip.py https://bootstrap.pypa.io/get-pip.py > /%{installdir}/pip3-update.log 2>&1
-python3 /%{installdir}/get-pip.py --prefix=/%{installdir} >> /%{installdir}/pip3-update.log 2>&1
-export PYTHONPATH=/%{installdir}/lib/python3.6/site-packages:${PYTHONPATH}
-export PATH=/%{installdir}/bin:${PATH}
-%endif
-
-# Install python dependencies
-pip3 --no-cache-dir install --prefix=/%{installdir} -r /%{installdir}/requirements.txt > /%{installdir}/pip3-install.log
+python3 -m pip install --upgrade pip 2>&1 > /%{installdir}/pip3-upgrade.log
+python3 -m pip --no-cache-dir install --prefix=/%{installdir} -r /%{installdir}/requirements.txt > /%{installdir}/pip3-install.log 2>&1
 
 # link the pbench profile, so it'll automatically be sourced on login
 ln -sf /%{installdir}/profile /etc/profile.d/pbench-agent.sh
--- a/agent/rpm/pbench-agent.spec.j2
+++ b/agent/rpm/pbench-agent.spec.j2
@@ -88,6 +88,9 @@ cp -a agent/* %{?buildroot}/%{installdir}/
 # configtools installed: zap it.
 if pip3 show configtools > /dev/null 2>&1 ;then pip3 uninstall -y configtools ;fi
 
+%post
+
+# Install python dependencies
 python3 -m pip install --upgrade pip 2>&1 > /%{installdir}/pip3-upgrade.log
 python3 -m pip --no-cache-dir install --prefix=/%{installdir} -r /%{installdir}/requirements.txt > /%{installdir}/pip3-install.log 2>&1
 
--- a/agent/rpm/pbench-agent.spec.j2
+++ b/agent/rpm/pbench-agent.spec.j2
@@ -89,10 +89,23 @@ cp -a agent/* %{?buildroot}/%{installdir}/
 if pip3 show configtools > /dev/null 2>&1 ;then pip3 uninstall -y configtools ;fi
 
 %post
+%if 0%{?rhel} == 7
+# This is a very ugly work-around for RHEL 7 / CentOS 7 environments. The
+# version of `pip3` that is available in those environments no longer works
+# to install the latest Python3 modules, which we need. So we have to fetch
+# our own copy of `pip3`, as described at:
+#   https://pip.pypa.io/en/latest/installing/#installing-with-get-pip-py 
+# We then install it separately from the existing version and set `PATH` and
+# `PYTHONPATH` so that we use this updated version when we install our module
+# dependencies below.
+curl -X GET -o /%{installdir}/get-pip.py https://bootstrap.pypa.io/get-pip.py > /%{installdir}/pip3-update.log 2>&1
+python3 /%{installdir}/get-pip.py --prefix=/%{installdir} >> /%{installdir}/pip3-update.log 2>&1
+export PYTHONPATH=/%{installdir}/lib/python3.6/site-packages:${PYTHONPATH}
+export PATH=/%{installdir}/bin:${PATH}
+%endif
 
 # Install python dependencies
-python3 -m pip install --upgrade pip 2>&1 > /%{installdir}/pip3-upgrade.log
-python3 -m pip --no-cache-dir install --prefix=/%{installdir} -r /%{installdir}/requirements.txt > /%{installdir}/pip3-install.log 2>&1
+pip3 --no-cache-dir install --prefix=/%{installdir} -r /%{installdir}/requirements.txt > /%{installdir}/pip3-install.log
 
 # link the pbench profile, so it'll automatically be sourced on login
 ln -sf /%{installdir}/profile /etc/profile.d/pbench-agent.sh
--- a/agent/util-scripts/pbench-tool-meister-start
+++ b/agent/util-scripts/pbench-tool-meister-start
@@ -753,7 +753,7 @@ def main(_prog: str, cli_params: Namespace) -> int:
         # us connecting, and we'll use that (by default) as the server address.
         localhost = LocalRemoteHost()
         origin_ip = set()
-        template = TemplateSsh(ssh_cmd, shlex.split(ssh_opts), ['echo', '${SSH_CONNECTION}'])
+        template = TemplateSsh(ssh_cmd, shlex.split(ssh_opts), "echo ${SSH_CONNECTION}")
         for host in tool_group.hostnames.keys():
             if not localhost.is_local(host):
                 template.start(host)
@@ -762,14 +762,14 @@ def main(_prog: str, cli_params: Namespace) -> int:
             if not localhost.is_local(host):
                 connection = template.wait(host)
                 logger.debug("Host %s reports connection `%s`", host, connection)
-                if connection[0] == 0 and connection[1]:
+                if connection.status == 0 and connection.stdout:
                     # The SSH_CONNECTION value is the full `stdout` and we
                     # don't expect a stderr on success; the format is
                     #   "origin_node origin_port local_node local_port"
                     # we only need the origin node because we know that's an
                     # IP for our local TDS and Redis host that the remote can
-                    # understand.
-                    origin = connection[1].split()[0]
+                    # reach.
+                    origin = connection.stdout.split()[0]
                     origin_ip.add(origin)
 
                     # Store the origin reported by this remote in its parameter
@@ -777,12 +777,13 @@ def main(_prog: str, cli_params: Namespace) -> int:
                     # we send the init handshake to that remote later.
                     params["origin_host"] = origin
                 else:
-                    # If `ssh` fails, there's no point in continuing!
+                    # If `ssh` fails, we can't orchestrate anything on this
+                    # remote, so terminate with an error.
                     logger.error("Host %s reports %s", host, connection)
                     return ReturnCode.REMOTENOTREACHABLE
 
-        # Convert the de-duplicating set into an indexable list. Ideally there
-        # will be only a single entry here, since the current BaseServer init
+        # Process the collected origin addresses from our remotes. Ideally we
+        # have only a single entry here, since the current BaseServer init
         # (and the Bottle WSGI server) doesn't support listening on multiple
         # addresses; however we're going to just take the first returned origin
         # address and hope they can all connect.
@@ -791,21 +792,15 @@ def main(_prog: str, cli_params: Namespace) -> int:
         # connection spec) to always listen on '0.0.0.0', but this eliminates
         # IPv6, and we currently lack the mechanism to listen on both '0.0.0.0'
         # and '::' (the IPv6 equivalent, aka '0:0:0:0:0:0:0:0').
-        origin_list = list(origin_ip)
-
+        #
         # NOTE: If the caller/human supplied explicit PBENCH_REDIS_SERVER or
         # PBENCH_TOOL_DATA_SINK (--redis-server or --tool-data-sink), we'll use
         # those, but by default we'll use the ssh connection origin instead of
         # the local `hostname` which may not be reachable by the remotes (or
         # necessarily routable at all).
-        #
-        # If we aren't able to contact any remotes, or if there were no remotes
-        # in the first place, then the origin_list will be empty, and we'll
-        # proceed with the default local hostname. If we have SSH origin IPs,
-        # however, we'll use the first of them.
-        if len(origin_list) > 0:
-            logger.debug("Our connection hosts are %s", origin_list)
-            ip = ipaddress.ip_address(origin_list[0])
+        if len(origin_ip) > 0:
+            logger.debug("Our connection hosts are %s", origin_ip)
+            ip = ipaddress.ip_address(iter(origin_ip).next())
             if isinstance(ip, ipaddress.IPv6Address):
                 origin = f"[{str(ip)}]"
             else:
--- a/agent/util-scripts/pbench-tool-meister-start
+++ b/agent/util-scripts/pbench-tool-meister-start
@@ -802,12 +802,12 @@ def main(_prog: str, cli_params: Namespace) -> int:
         # necessarily routable at all).
         if len(origin_ip) != 1 and any_remote:
             logger.warning(
-                "Remote hosts don't report a single controller "
-                "origin IP: this may be a problem: origins are %s",
-                origin_ip
+                "Remote hosts don't agree on a single controller "
+                "origin IP, which may indicate a problem: origin(s) %s",
+                ",".join(origin_ip) if origin_ip else "<none>"
             )
         if len(origin_ip) > 0:
-            logger.debug("Our connection hosts are %s", origin_ip)
+            logger.debug("Our connection host(s): %s", ",".join(origin_ip))
             ip = ipaddress.ip_address(next(iter(origin_ip)))
             if isinstance(ip, ipaddress.IPv6Address):
                 origin = f"[{str(ip)}]"
--- a/agent/util-scripts/pbench-tool-meister-start
+++ b/agent/util-scripts/pbench-tool-meister-start
@@ -97,6 +97,7 @@ the benchmark execution environment:
   - ssh-opts
 """
 
+import ipaddress
 import json
 import logging
 import os
@@ -108,7 +109,6 @@ import time
 from argparse import ArgumentParser, Namespace
 from distutils.spawn import find_executable
 from pathlib import Path
-
 import redis
 
 from pbench.agent.constants import (
@@ -133,6 +133,7 @@ from pbench.agent.utils import (
     info_log,
     LocalRemoteHost,
     RedisServerCommon,
+    TemplateSsh
 )
 from pbench.common.utils import validate_hostname
 
@@ -194,6 +195,7 @@ class ReturnCode(BaseReturnCode):
     BADFULLHOSTNAME = 37
     BADHOSTNAME = 38
     INVALIDORCHESTRATE = 39
+    REMOTENOTREACHABLE = 40
 
 
 def _waitpid(pid: int) -> int:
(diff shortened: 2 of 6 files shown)


lib/pbench/agent/utils.py
```diff
@@ -396,6 +398,58 @@ def collect_local_info(pbench_bin):
...
+        self.base_args = [ssh_cmd] + ssh_args
+
+    def start(self, host: str):
+        """
+        Begin an asynchronous ssh command on the specified remote host
+
+        Args:
+            host: hostname or IP
+        """
+        args = self.base_args + [host] + self.command
```
webbnh, line 427: Should we be [quoting](https://docs.python.org/3/library/shlex.html#shlex.quote) (or joining) `self.command`?  (Or, accepting it as a single string, instead of a list, and expecting the caller to do it?)

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -731,13 +733,86 @@ def main(_prog: str, cli_params: Namespace) -> int:
...
+                    origin = connection[1].split()[0]
+                    origin_ip.add(origin)
+
+                    # Store the origin reported by this remote in its parameter
+                    # block: we'll program this as the connection address when
+                    # we send the init handshake to that remote later.
+                    params["origin_host"] = origin
+                else:
+                    # If `ssh` fails, there's no point in continuing!
+                    logger.error("Host %s reports %s", host, connection)
```
webbnh, line 780: This is going to log a stringized tuple...is that really what you want?  (Wouldn't it be clearer to unpack and annotate it a bit?)

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -731,13 +733,86 @@ def main(_prog: str, cli_params: Namespace) -> int:
...
+        origin_list = list(origin_ip)
+
+        # NOTE: If the caller/human supplied explicit PBENCH_REDIS_SERVER or
+        # PBENCH_TOOL_DATA_SINK (--redis-server or --tool-data-sink), we'll use
+        # those, but by default we'll use the ssh connection origin instead of
+        # the local `hostname` which may not be reachable by the remotes (or
+        # necessarily routable at all).
+        #
+        # If we aren't able to contact any remotes, or if there were no remotes
+        # in the first place, then the origin_list will be empty, and we'll
```
webbnh, line 802: Actually, if there were any remotes which we were unable to contact, then control would have returned by now and `origin_list` would never have been created...right?

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -731,13 +733,86 @@ def main(_prog: str, cli_params: Namespace) -> int:
...
+        # proceed with the default local hostname. If we have SSH origin IPs,
+        # however, we'll use the first of them.
+        if len(origin_list) > 0:
+            logger.debug("Our connection hosts are %s", origin_list)
+            ip = ipaddress.ip_address(origin_list[0])
```
webbnh, line 807: I don't see any obvious reason to convert `origin_ip` from a `set` to a `list` (e.g., `len()` works on `set`'s).

If the problem is that you just want to pick an arbitrary item from it, you can `pop()` from a set.  (Or, if you don't want to modify it, you can say `next(x for x in origin_ip)`.)

webbnh (commented):
I realize that this is just a draft, but I've got some things you might want to consider as you work on this.

lib/pbench/agent/utils.py
```diff
@@ -396,6 +398,58 @@ def collect_local_info(pbench_bin):
...
+        Wait for an asynchronous ssh command to complete, returning the
+        completion status, stdout and stderr streams as strings.
+
+        Args:
+            host: Remote host
+
+        Returns:
+            Tuple of completion status, stdout, stderr
+        """
+        popen: subprocess.Popen = self.procs[host]
```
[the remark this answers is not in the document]
dbutenhof: > Or, we could just go with the KeyError, which is probably sufficient....

That was my thought...

dbutenhof (commented):
I mostly pushed as a draft so it would run tests before I opened it, but I also somewhat counted on you taking an early look. I'll make some changes and push another commit before I consider "publishing" it.

agent/rpm/pbench-agent.spec.j2
```diff
@@ -89,23 +89,10 @@ cp -a agent/* %{?buildroot}/%{installdir}/
...
-curl -X GET -o /%{installdir}/get-pip.py https://bootstrap.pypa.io/get-pip.py > /%{installdir}/pip3-update.log 2>&1
-python3 /%{installdir}/get-pip.py --prefix=/%{installdir} >> /%{installdir}/pip3-update.log 2>&1
-export PYTHONPATH=/%{installdir}/lib/python3.6/site-packages:${PYTHONPATH}
-export PATH=/%{installdir}/bin:${PATH}
-%endif
 
 # Install python dependencies
-pip3 --no-cache-dir install --prefix=/%{installdir} -r /%{installdir}/requirements.txt > /%{installdir}/pip3-install.log
+python3 -m pip install --upgrade pip 2>&1 > /%{installdir}/pip3-upgrade.log
+python3 -m pip --no-cache-dir install --prefix=/%{installdir} -r /%{installdir}/requirements.txt > /%{installdir}/pip3-install.log 2>&1
```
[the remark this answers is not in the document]
webbnh: Yeah, I remember...which is why it's so mystifying that `python3 -m pip` is somehow better....  (And, I'm not suggesting going back to `pip3`....)

lib/pbench/agent/utils.py
webbnh: I looked at what Peter did and ultimately decided that I liked yours better (because I couldn't come up with a really good way to "remember" which element was the hostname...the best I could come up with was to use a `dict` instead of a `list`, which the code could access by key-name and then reference the `values()` of it for the command invocation, but what you've got seemed better (although, now that I think of it, a `namedtuple` would work well, since you can access the components by name but also iterate over the components like a list, so you might be able to pass it directly to `Popen()` (or worst-case, apply the `list()` constructor to it))).

But, what I'm getting at with my comment here is that `base_args` + `host` are the arguments to the `ssh` command, and it takes **_one_** more argument, which is the remote command to be run.  It shouldn't take a list of components for that command, because (I think) we don't want to feed additional components to the `ssh` command.  (I was reading about `shlex.split()` which led to `shlex.quote()` whose code-injection example led me to ponder your command argument to `ssh`....)

agent/util-scripts/pbench-tool-meister-start
webbnh: I was mostly reacting to the wall of text here.  (It's all good and important stuff, so I'm not saying we should get rid of it, but, since it starts out by pointing out that we're converting a set to a list (which is about the least interesting aspect, here), I figured maybe we could get rid of that one facet so it wouldn't be a distraction from the rest of it.)  I think it's wonderfully clever to use a `set` here.  I just wish there were a nicer way to grab an element from it.

lib/pbench/agent/utils.py
ndokos: > `ssh root@host ls /tmp` and `ssh root@host 'ls /tmp'` are exactly the same.

So it is guaranteed that no substitutions are made? How? Do we have to examine the callers?

lib/pbench/agent/utils.py
```diff
@@ -396,6 +398,58 @@ def collect_local_info(pbench_bin):
...
+    def wait(self, host: str) -> Tuple[int, str, str]:
+        """
+        Wait for an asynchronous ssh command to complete, returning the
+        completion status, stdout and stderr streams as strings.
+
+        Args:
+            host: Remote host
+
+        Returns:
+            Tuple of completion status, stdout, stderr
```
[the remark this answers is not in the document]
ndokos: "Noone expects the Spanish Inquisition..."

ndokos (changes requested):

lib/pbench/agent/utils.py
dbutenhof: Substitution depends on whether you're running the command through a shell or "bare", as well as on the tokens you use (regardless of argv breakdown vs string) are quoted. I'm not running the `subprocess.Popen` with `shell=True` so there's no local shell in any case.

There's a single caller for the new class (although I've toyed with the idea of converting the `pbench-tool-meister-start` method `start_tms_via_ssh` because I think this framework would be cleaner ... if I expand the templating mechanism to allow altering the command as well as the host). The single caller just wants to pass `echo ${SSH_CONNECTION}` to be expanded on the remote so we get the SSH origin hostname, which provides our guarantee of external routability.

agent/rpm/pbench-agent.spec.j2
```diff
@@ -89,23 +89,10 @@ cp -a agent/* %{?buildroot}/%{installdir}/
...
-# our own copy of `pip3`, as described at:
-#   https://pip.pypa.io/en/latest/installing/#installing-with-get-pip-py 
-# We then install it separately from the existing version and set `PATH` and
-# `PYTHONPATH` so that we use this updated version when we install our module
-# dependencies below.
-curl -X GET -o /%{installdir}/get-pip.py https://bootstrap.pypa.io/get-pip.py > /%{installdir}/pip3-update.log 2>&1
-python3 /%{installdir}/get-pip.py --prefix=/%{installdir} >> /%{installdir}/pip3-update.log 2>&1
-export PYTHONPATH=/%{installdir}/lib/python3.6/site-packages:${PYTHONPATH}
-export PATH=/%{installdir}/bin:${PATH}
-%endif
```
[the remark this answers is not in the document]
dbutenhof: I've encountered situations (of unknown cause) on both server and agent installs where this doesn't work; we still get a "no pip command" error and nothing gets installed.

When we butted our heads against this during server jam sessions, we ended up using the new code I've pasted here, which just uses `python3 -m pip` instead of expecting a `pip` command. I really need to experiment with this on some bare untouched systems to be sure it works on them all; which is only one of the reasons this is still a draft. But if it *doesn't* work, then we've got some holes that may be hard to fix.

lib/pbench/agent/utils.py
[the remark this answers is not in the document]
dbutenhof: There was a guy at Digital who went to Spain and brought back an ornate box just so he could refuse any request to examine it by chortling madly and intoning "Nobody inspects the Spanish acquisition".

lib/pbench/agent/utils.py
webbnh: > I'm not running the subprocess.Popen with shell=True so there's no local shell in any case.

That's a "feature".  (That also allays most of my concern...and, it also means that there's no confusion about where `${SSH_CONNECTION}` will be evaluated.)

lib/pbench/agent/utils.py
```diff
@@ -447,7 +454,7 @@ def wait(self, host: str) -> Tuple[int, str, str]:
             out, err = popen.communicate()
         status = popen.returncode
         del self.procs[host]
-        return (status, out, err)
+        return self.Return(status, out, err)
```
webbnh, line 457: It would be nicer if you used the keywords:
```
        return Return(status=status, stdout=out, stderr=err)
```

webbnh (commented):
Looks good...for a draft.... 😉

agent/rpm/pbench-agent.spec.j2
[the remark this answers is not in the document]
dbutenhof: I decided to do what I resisted in the first place, pushing this out of the PR to test separately with the new VM pool. (Once I get through all the `subscription-manager` setup muddle across the releases.) So no longer relevant here...

webbnh (commented):
👍

(Although, I'm sad to see the "simple" `python -m pip` reverted to the `curl`-based approach....)

dbutenhof:
> +1
> 
> (Although, I'm sad to see the "simple" `python -m pip` reverted to the `curl`-based approach....)

Yeah; but Nick brought up a valid point about testing, and it'll take me a bit of time to wrangle all of the new VM instances he set up for that purpose so I'll do that separately and not bog this one down.

Although this still needs a rebase, and I'll probably wait until tomorrow to let Pete get in a few more of his zoo before I mark this "ready for review".

agent/rpm/pbench-agent.spec.j2
```diff
@@ -89,10 +89,23 @@ cp -a agent/* %{?buildroot}/%{installdir}/
...
+# dependencies below.
+curl -X GET -o /%{installdir}/get-pip.py https://bootstrap.pypa.io/get-pip.py > /%{installdir}/pip3-update.log 2>&1
+python3 /%{installdir}/get-pip.py --prefix=/%{installdir} >> /%{installdir}/pip3-update.log 2>&1
+export PYTHONPATH=/%{installdir}/lib/python3.6/site-packages:${PYTHONPATH}
+export PATH=/%{installdir}/bin:${PATH}
+%endif
 
 # Install python dependencies
-python3 -m pip install --upgrade pip 2>&1 > /%{installdir}/pip3-upgrade.log
-python3 -m pip --no-cache-dir install --prefix=/%{installdir} -r /%{installdir}/requirements.txt > /%{installdir}/pip3-install.log 2>&1
```
portante, line 95: Looks like a bad merge?

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -732,13 +734,81 @@ def main(_prog: str, cli_params: Namespace) -> int:
...
+        # IPv6, and we currently lack the mechanism to listen on both '0.0.0.0'
+        # and '::' (the IPv6 equivalent, aka '0:0:0:0:0:0:0:0').
+        #
+        # NOTE: If the caller/human supplied explicit PBENCH_REDIS_SERVER or
+        # PBENCH_TOOL_DATA_SINK (--redis-server or --tool-data-sink), we'll use
+        # those, but by default we'll use the ssh connection origin instead of
+        # the local `hostname` which may not be reachable by the remotes (or
+        # necessarily routable at all).
+        if len(origin_ip) > 0:
+            logger.debug("Our connection hosts are %s", origin_ip)
```
portante, line 802: Perhaps make this an informational message when the length of `origin_ip` is greater than one?  Having this is a `debug` message might get lost.

portante (changes requested):
Looks good!

But ... the tests are not working ...

agent/rpm/pbench-agent.spec.j2
dbutenhof: Weird. 😦

agent/util-scripts/pbench-tool-meister-start
dbutenhof: I briefly debated a `warning` since it may indicate that some remotes might not be able to reach Redis/TDS. (Or at least, if they can't, this message may provide a clue as to why.) I'm not quite sure why I just left it at `debug`.

dbutenhof: I'm going to issue a warning if the number of origin IPs isn't exactly 1, because either 0 or > 1 might indicate a routing problem (that is, it's possible that not all remotes will be able to reach our TDS/Redis). I'm going to leave the debug for > 0 just as incidental information on the discovery process.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -732,13 +734,87 @@ def main(_prog: str, cli_params: Namespace) -> int:
...
+        # those, but by default we'll use the ssh connection origin instead of
+        # the local `hostname` which may not be reachable by the remotes (or
+        # necessarily routable at all).
+        if len(origin_ip) != 1:
+            logger.warning(
+                "Remote hosts don't report a single controller "
+                "origin IP: this may be a problem: origins are %s",
+                origin_ip
+            )
+        if len(origin_ip) > 0:
```
portante, line 807: What happens if the length is zero?

portante (changes requested):

portante (changes requested):
Looks like the test output needs updating ...

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -732,13 +734,87 @@ def main(_prog: str, cli_params: Namespace) -> int:
...
+            logger.debug("Our connection hosts are %s", origin_ip)
+            ip = ipaddress.ip_address(iter(origin_ip).next())
+            if isinstance(ip, ipaddress.IPv6Address):
+                origin = f"[{str(ip)}]"
+            else:
+                origin = str(ip)
+            if not tds_server_spec:
+                tds_server_spec = origin
+            if not redis_server_spec:
+                redis_server_spec = origin
```
webbnh, line 817: I misread this code on the first try (and I don't like the "origins are set()" message when the set is empty), so I recommend a slight restructuring:
```
        if len(origin_ip) < 1:
            # This means that all hosts are local, right?  So, there _should_ be
            # nothing to do in this case, because any errors would have caused
            # control to return from this function above.
            pass
        else:
            if len(origin_ip) > 1:
                logger.warning(
                    "Remote hosts don't report a single controller "
                    "origin IP: this may be a problem: origins are %s",
                     origin_ip
                )
            logger.debug("Our connection hosts are %s", origin_ip)
            ip = ipaddress.ip_address(iter(origin_ip).next())
            ...
```
Obviously, the first `if` condition could be reversed to get rid of the `pass` body, but then we would need to put the explanatory comment before it.

webbnh (approved):
Looks good, although I have a small suggestion to help address Peter's and my respective stumbles.

agent/util-scripts/pbench-tool-meister-start
webbnh: That should be the case where all of the "remote" are actually local (but otherwise valid).

dbutenhof:
> Looks like the test output needs updating ...

That's weird ... with my mock `ssh` changes I didn't think we could hit this case (although I should have looked, and it's nice to have it covered); but more importantly I ran a full local test set and didn't see any failures. Sigh; it's always something.

agent/util-scripts/pbench-tool-meister-start
dbutenhof: It could also happen if the remotes don't get a definition of `SSL_CONNECTION` for some reason, I think (i.e., success, but `stdout=""`); I don't really want to fail in that case, but the remote isn't helping us to figure out a common origin. My idea was that we just go ahead an follow the old behavior, trying `hostname`; if it's not routable, there's not much more we can do, but I didn't want to force a failure when we can't be sure ahead of time. We should have a timeout on the initialization protocol to catch this case, but that doesn't belong in this PR.

However, Webb's comment makes me realize I do need another tweak here; `len(origin_ip) == 0` could mean we're strictly local, and the `warning` is inappropriate in that case.

agent/util-scripts/pbench-tool-meister-start
dbutenhof: Again `len(origin_ip) == 0` ideally means we're local; but if not, then the `ssh` remotes didn't get a definition for `SSL_CONNECTION` and are not giving us a clue. I need to separate those two cases and skip the warning only when there are no remotes.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -732,13 +734,89 @@ def main(_prog: str, cli_params: Namespace) -> int:
...
+        # NOTE: If the caller/human supplied explicit PBENCH_REDIS_SERVER or
+        # PBENCH_TOOL_DATA_SINK (--redis-server or --tool-data-sink), we'll use
+        # those, but by default we'll use the ssh connection origin instead of
+        # the local `hostname` which may not be reachable by the remotes (or
+        # necessarily routable at all).
+        if len(origin_ip) != 1 and any_remote:
+            logger.warning(
+                "Remote hosts don't report a single controller "
+                "origin IP: this may be a problem: origins are %s",
+                origin_ip
```
webbnh, line 807: I suggest replacing line 807 with something like
```
                origin_ip if origin_ip else "<none>"
```
(I really don't like the `origins are set()` output.)

webbnh (approved):
👍

agent/util-scripts/pbench-tool-meister-start
dbutenhof: These are log messages, and it seems silly to waste the effort to make them more "pretty" than necessary to convey the debugging information. However, you have a point in that here, as in much of the agent code, we seem to be using Python's logger to generate end-user messages. I really don't like that, but I suppose I'm stuck with it here. 😬

npalaska (approved):
Looks good

portante (dismissed):

portante (approved):

npalaska (approved):

agent/util-scripts/pbench-tool-meister-start
webbnh: > it seems silly to waste the effort to make them more "pretty" than necessary

I agree; however, given that I might be the poor slob who has to read them, future-me would likely appreciate it if past-me (or past-you) made them "pretty enough".  (Without a certain amount of context, I think that `set()` is likely to be confusing, whereas `{}` or `<none>` is pretty clear.)

webbnh (approved):
Oooh, `",".join()` -- that's even nicer!  😁

PR #2644, opened 2022-02-26
Sort python requirements files

portante:
Broken out from PR #2551.

PR #2646, opened 2022-02-26
Stop using `selinux` virtualenv wrapper

### LINKED ISSUE #2645: Stop using `selinux` as an explicit required python module
This is a real mess.

Basically, there is no real `selinux` python module provided via `pip` install.  All that is out there is a "shim" (https://github.com/pycontribs/selinux), which is really broken.  On module load, it forks an invocation of the predominate `/usr/bin/python3` program (see https://github.com/pycontribs/selinux/blob/master/selinux/__init__.py#L76) to get it's list of site packages, and replaces itself with a module load of those.

The goal of the system is to allow a virtualenv using the same python version as the real `selinux` module is installed on the system to work "seemlessly."

But if your virtualenv is for a python version different from what is installed on the system, then the `selinux` import will fail.

For our testing environment, we should not use the `selinux` pip installed module, but instead provide our own selinux module which checks the existence of the real `selinux` module and provides a `is_selinux_enabled()` method which returns True and mocks out the `restorecon()` method.

For non-testing environments, our selinux wrapper should just check the existence of the real `selinux` module and provide a `is_selinux_enabled()` method which returns False when the real module does not exist (and mock out the `restorecon()` methods just in case).

portante:
Fixes #2645.

Basically, there is no real `selinux` python module provided via `pip` install. All that is out there is a "shim" [1], which is really broken.  On module load, it forks an invocation of the predominate `/usr/bin/python3` program [2] to get it's list of site packages, and replaces itself with a module load of the `selinux` module it finds among those site package locations.

The goal of that virtualenv wrapper is to allow a virtualenv using the same python version as the real installed `selinux` module to work "seamlessly."

But if your virtualenv is for a python version different from what is installed on the system, then the `selinux` import will fail with an error trying to load a module that doesn't work with the version of python in the virtualenv.

We no longer use the `selinux` pip installed module.  Instead we provide our own wrapper which provides pass-through methods for `is_selinux_enabled()` and `restorecon()` when a `selinux` module is found, and mocks for those same methods where the `is_selinux_enabled()` method returns `False`, and the `restorecon()` method raises an exception to help detect logic errors.

For our testing environments, we already provide proper mocks of the `selinux` module to get the desired behavior (now updated to work with the new wrapper).

This is foundational work for PR #2551.

[1] https://github.com/pycontribs/selinux
[2] https://github.com/pycontribs/selinux/blob/master/selinux/__init__.py#L76

### DIFF
--- a/lib/pbench/common/selinux.py
+++ b/lib/pbench/common/selinux.py
@@ -1,0 +1,34 @@
+"""SELinux wrapper module.
+
+There is no "pip" installable `selinux` python module, just a shim, of sorts [1],
+which provides a way to add the installed (via RPM) system's `selinux` module to
+a virtualenv.  But that does not work when the virtualenv uses a version of
+python different from the python version where the real `selinux` module is
+installed.
+
+This wrapper acts as a pass-through to the real `selinux` module for the methods
+we need, when that module is present.  When that module is not present, we
+provide module methods to mimic the behaviors we need.
+
+[1] https://github.com/pycontribs/selinux
+"""
+try:
+    import selinux
+except ImportError:
+    # The import failed; provide stub implementations.
+
+    def is_selinux_enabled():
+        """Always indicates "disabled"."""
+        return 0
+
+    def restorecon(path):
+        """Always raises an exception, to identify logic problems."""
+        raise NotImplementedError(
+            "Logic bomb!  selinux.restorecon() called when selinux is not enabled."
+        )
+
+
+else:
+    # The import succeeded; forward our functions to the real implementations.
+    is_selinux_enabled = selinux.is_selinux_enabled
+    restorecon = selinux.restorecon
--- a/lib/pbench/server/filetree.py
+++ b/lib/pbench/server/filetree.py
@@ -6,9 +6,9 @@ import shutil
 from typing import Dict, List, Union
 
 from dateutil import parser as date_parser
-import selinux
 import tarfile
 
+from pbench.common import selinux
 from pbench.server import PbenchServerConfig, JSON
 from pbench.server.database.models.datasets import Dataset
 
--- a/lib/pbench/test/unit/server/test_file_tree.py
+++ b/lib/pbench/test/unit/server/test_file_tree.py
@@ -51,7 +51,7 @@ def selinux_disabled(monkeypatch):
     """
     Pretend that selinux is disabled so restorecon isn't called
     """
-    monkeypatch.setattr("selinux.is_selinux_enabled", lambda: 0)
+    monkeypatch.setattr("pbench.common.selinux.is_selinux_enabled", lambda: 0)
 
 
 @pytest.fixture()
@@ -59,8 +59,8 @@ def selinux_enabled(monkeypatch):
     """
     Pretend that selinux is enabled but make restorecon exit with no action
     """
-    monkeypatch.setattr("selinux.is_selinux_enabled", lambda: 1)
-    monkeypatch.setattr("selinux.restorecon", lambda a: None)
+    monkeypatch.setattr("pbench.common.selinux.is_selinux_enabled", lambda: 1)
+    monkeypatch.setattr("pbench.common.selinux.restorecon", lambda a: None)
 
 
 class TestFileTree:
--- a/server/bin/pbench-server-prep-shim-002.py
+++ b/server/bin/pbench-server-prep-shim-002.py
@@ -12,11 +12,11 @@ import os
 import sys
 import glob
 import shutil
-import selinux
 import tempfile
 
 from pathlib import Path
 
+from pbench.common import selinux
 from pbench.common.exceptions import BadConfig
 from pbench.common.logger import get_pbench_logger
 from pbench.common.utils import md5sum
--- a/server/requirements.txt
+++ b/server/requirements.txt
@@ -19,6 +19,5 @@ pyesbulk>=2.0.1
 PyJwt
 psycopg2
 requests
-selinux
 sqlalchemy>=1.4.23
 sqlalchemy_utils>=0.37.6


lib/pbench/common/selinux.py
```diff
@@ -0,0 +1,45 @@
...
+    import selinux
+except ImportError:
+
+    class selinux:
+        """Pbench SELinux wrapper."""
+
+        @staticmethod
+        def is_selinux_enabled():
+            """Always returns False."""
+            return False
```
dbutenhof, line 30: Oh, the layered PRs ... so this comment belongs here rather than in #2644 ...

I found while working out my selinux mocks for file tree testing that selinux returns integer `0` or `1`, not `True` or `False`, and we should do the same here for consistency.

dbutenhof (changes requested):

lib/pbench/common/selinux.py
portante: Fixed.

portante (commented):
All set, just need another reviewer.

lib/pbench/common/selinux.py
```diff
@@ -0,0 +1,45 @@
+"""SELinux wrapper module.
+
+There is no "pip" installable `selinux` python module, just a shim, of sorts [1],
+which provides a way to add the installed system's `selinux` module to a
+virtualenv.  But that does not work when the virtualenv uses a version of
+python different from the python version where the real `selinux` module is
+installed.
+
+The only time we need a fully working `selinux` module is when we are not
+running tests.
```
webbnh, line 10: I wish that that were obviously true.

In an ideal world, we would test both with SELinux enabled and with it disabled, and we would do that for each of the versions of Python that we support (on each of the platforms that we support).

Assuming that our code will work with SELinux enabled after we've carefully avoided testing that seems a bit of a stretch.  🙁

lib/pbench/common/selinux.py
```diff
@@ -0,0 +1,45 @@
...
+        @staticmethod
+        def restorecon(path):
+            """Always raises an exception to identify logic problems."""
+            raise Exception(
+                "Logic bomb!  selinux.restorecon() called when selinux is not enabled."
+            )
+
+
+def is_selinux_enabled():
+    return selinux.is_selinux_enabled()
```
webbnh, line 41: It bothers me slightly that sometimes here `selinux` will be a package and sometimes it will be a class.  It's an elegant hack that Python doesn't care, but I wonder if maybe we shouldn't do better.

For instance, it wouldn't be hard to set up functions/wrappers/pointers in each of the `try` and `except` branches at lines 20-22, where we know what's what.

lib/pbench/common/selinux.py
```diff
@@ -0,0 +1,45 @@
...
+        def is_selinux_enabled():
+            """Always returns 0."""
+            return 0
+
+        @staticmethod
+        def restorecon(path):
+            """Always raises an exception to identify logic problems."""
+            raise Exception(
+                "Logic bomb!  selinux.restorecon() called when selinux is not enabled."
+            )
```
webbnh, line 37: Rather than a generic exception, wouldn't it be better to raise `NotImplementedError`?  (You can still provide a custom message for it.)  Or, perhaps, `NameError`, `AssertionError `, or `ImportError`?

webbnh (approved):
These changes are OK, but I have some reservations.

First, should we have unit tests for this new module?  (Yes, I realize that the module is fairly trivial, but that just means the tests would be easy to write....)

I'm generally opposed to using generic `Exception` objects for anything...we should try to do better.

And, the conflation of package and class seems like a hack.

lib/pbench/common/selinux.py
dbutenhof: The file tree unit test mocks selinux enabled and disabled; the legacy test system is too fake to be real. We should have a real functional test environment, and, yes one of the variants we should be able to test is selinux on or off ... definitely a point to keep in mind.

dbutenhof (dismissed):

lib/pbench/common/selinux.py
webbnh: For instance, I think something like this would be better:
```
try:
    import selinux

    # Forward  our functions to the real implementations.                                                                                                                                                                             
    is_selinux_enabled = selinux.is_selinux_enabled
    restorecon = selinux.restorecon
except ImportError:
    # The import failed; provide stub implementations.

    def is_selinux_enabled():
        """Always indicates "disabled"."""
        return 0

    def restorecon(path):
        """Always raises an exception, to identify logic problems."""
        raise NotImplementedError(
            "Logic bomb!  selinux.restorecon() called when selinux is not enabled."
        )
```
or
```
try:
    import selinux
except ImportError:
    # The import failed; provide stub implementations.

    def is_selinux_enabled():
        """Always indicates "disabled"."""
        return 0

    def restorecon(path):
        """Always raises an exception, to identify logic problems."""
        raise NotImplementedError(
            "Logic bomb!  selinux.restorecon() called when selinux is not enabled."
        )

else:
    # The import succeeded; forward our functions to the real implementations.
    is_selinux_enabled = selinux.is_selinux_enabled
    restorecon = selinux.restorecon
```

portante: I took the second suggestion.

lib/pbench/common/selinux.py
portante: Removed, not worth having it.

lib/pbench/common/selinux.py
```diff
@@ -0,0 +1,36 @@
+"""SELinux wrapper module.
+
+There is no "pip" installable `selinux` python module, just a shim, of sorts [1],
+which provides a way to add the installed system's `selinux` module to a
+virtualenv.  But that does not work when the virtualenv uses a version of
+python different from the python version where the real `selinux` module is
+installed.
```
webbnh, line 7: If you edit this module again, it would be nice to clarify that the "installed system's" module got there (e.g.) via RPM (because, typically, Pip is how Python modules get "installed"...and without the RPM prompt, these statements are likely to be confusing).

lib/pbench/common/selinux.py
```diff
@@ -0,0 +1,36 @@
...
+
+This wrapper acts as a pass-through to the real `selinux` module for the methods
+we need, when that module is present.  When that module is not present, we
+create a class to mimic the behaviors we need.  That created class declares that
```
webbnh, line 11: "[...]  When that module is not present, we provide stub routines which declare that [...]"

webbnh (dismissed):
I'll approve, but the comment (which refers to a "class" that we no longer provide) should presumably be fixed.

lib/pbench/common/selinux.py
webbnh: Ha!  You and your affinity for `else` clauses on `try` blocks.  😀

webbnh (dismissed):
I totally did not see that missing `d`.

portante (commented):
Okay, fixed!

dbutenhof (approved):

webbnh (approved):
👍

PR #2649, opened 2022-02-26
Parallelize `tox` invocations

portante:
Parallelize the python & legacy unit tests

We add the use of `pytest-xdist`, which allows us to pick the level of parallelism via the `-n` switch.  By default we use `-n auto` where it determines the parallelism to use based on the number of CPUs.

Use `parallel` to run all the legacy tests.  Each agent legacy sub-test is run as a separate parallel job, and then all the server tests are individually run through GNU `parallel`.

To facilitate the use of `parallel` for the server legacy tests, we break out the core server test harness out into a separate script. This helps to isolate the way tests are run from how the parallelism is achieved.

We also take the opportunity to use `local` in the various `bash` scripts, and make variable references consistent.

Finally, we address `mkdir -p` usage to remove its use, where possible, to ensure setup was performed without any hidden, or unexpected, directory creation, while fixing a problem in the server unit tests where the `common` directory was created but not used.

### DIFF
--- a/exec-unittests
+++ b/exec-unittests
@@ -6,7 +6,7 @@
 # For now we ignore the tox environment directory argument.
 _toxenvdir="${1}"
 if [[ -z "${_toxenvdir}" ]]; then
-    printf -- "Missing required tox environment directory." >&2
+    printf -- "Missing required tox environment directory.\n" >&2
     exit 2
 fi
 shift
@@ -31,25 +31,56 @@ fi
 subtst="${1}"
 shift
 
-# Remaining positional arguments past along to whatever test is run.
+# Remaining positional arguments passed along to whatever test is run.
 posargs="${@}"
+
+parallel=${PBENCH_UNITTEST_PARALLEL:-auto}
+if [[ "${parallel}" == "serial" ]]; then
+    para_jobs_arg="-j 1"
+    pytest_jobs_arg="-n 1"
+elif [[ "${parallel}" == "auto" ]]; then
+    para_jobs_arg=""
+    pytest_jobs_arg="-n auto"
+else
+    printf -- "Unrecognized PBENCH_UNITTEST_PARALLEL environment variable value, '%s'\n" "${parallel}" >&2
+    exit 2
+fi
 
 function _time {
     /usr/bin/time --format="\n\nCommand: '%C'\nExit status: %x\nTimings: user %Us, system %Ss, elapsed %es (%E, %P)\nMemory: max RSS %Mk, minor pf: %R, major pf: %F, swaps %W\nContext switches: inv %c, vol %w, signals %k\nI/O: fs in %I, fs out %O, socket in %r, socket out %s\n" ${@}
 }
 
 function run_legacy {
-    printf -- "\n\n\nRunning %s/%s legacy unit tests\n\n" "${1}" "${2}"
     local _rc
-    _time ${1}/${2}/unittests ${posargs}
+    local _major
+    local _subtst
+
+    _major=${1}
+    shift
+    _subtst=${1}
+    shift
+
+    printf -- "\n\n\nRunning %s/%s legacy unit tests\n\n" "${_major}" "${_subtst}"
+    _time ${_major}/${_subtst}/unittests ${@}
     _rc=${?}
     if [[ ${_rc} -ne 0 ]]; then
-        printf -- "\n%s %s legacy unit tests failed with '%s'\n\n" "${1^}" "${2}" "${_rc}"
+        printf -- "\n%s %s legacy unit tests failed with '%s'\n\n" "${_major^}" "${_subtst}" "${_rc}"
     else
-        printf -- "\n%s %s legacy unit tests succeeded\n\n" "${1^}" "${2}"
+        printf -- "\n%s %s legacy unit tests succeeded\n\n" "${_major^}" "${_subtst}"
     fi
     return ${_rc}
 }
+
+function para_run_legacy {
+    # When we want to use the `run_legacy` function with `parallel`, we have a
+    # problem because the command line arguments are passed as one long
+    # string.  So this jump function breaks up the arguments and invokes
+    # `run_legacy` as expected.
+    run_legacy ${1}
+}
+
+# Export functions for use with `parallel` below.
+export -f para_run_legacy run_legacy _time
 
 rc=0
 
@@ -74,6 +105,7 @@ if [[ -z "${subtst}" || "${subtst}" == "python" ]]; then
     printf -- "\n\n\nRunning %s python3-based unit tests via pytest\n\n" "${major_list// /,}"
     _pbench_sources=$(python3 -c 'import inspect, pathlib, pbench; print(pathlib.Path(inspect.getsourcefile(pbench)).parent.parent)')
     _PBENCH_COV_DIR="${_toxenvdir}/cov" _time pytest \
+        ${pytest_jobs_arg} \
         --basetemp="${_toxenvdir}/tmp" \
         --cov=${_pbench_sources} \
         --cov-branch \
@@ -91,28 +123,41 @@ fi
 
 _subtst_list="tool-scripts/datalog tool-scripts/postprocess tool-scripts util-scripts bench-scripts"
 
+_para_jobs_file="${_toxenvdir}/agent-legacy-jobs.lis"
+trap "rm -f ${_para_jobs_file}" EXIT INT TERM
+
 let count=0
 for _major in ${major_list}; do
     if [[ "${_major}" == "agent" ]]; then
+        # The parallel program is really cool.  The usage of `parallel` is
+        # internal and automated; only test code depends on this tool, and we,
+        # as developers, have viewed the citation and are justified in
+        # suppressing future displays of it in our development processes (use of
+        # --will-cite below).
+
         for _subtst in ${_subtst_list}; do
-            if [[ -z "${subtst}" || "${subtst}" == "$(basename ${_subtst})" ]]; then
+            if [[ "${subtst:-legacy}" == "legacy" || "${subtst}" == "$(basename ${_subtst})" ]]; then
+                echo "${_major} ${_subtst} ${posargs}"
                 (( count++ ))
-                run_legacy ${_major} ${_subtst} ${posargs} || rc=1
             fi
-        done
+        done > ${_para_jobs_file}
+        parallel --will-cite -k --lb -a ${_para_jobs_file} ${para_jobs_arg} para_run_legacy
+        if [[ ${?} -ne 0 ]]; then
+            rc=1
+        fi
     elif [[ "${_major}" == "server" ]]; then
         if [[ -z "${subtst}" || "${subtst}" == "legacy" ]]; then
             (( count++ ))
             run_legacy ${_major} bin ${posargs} || rc=1
         fi
     else
-        printf -- "Logic bomb!  Unrecognized major test sub-set, '%s'" "${_major}" >&2
+        printf -- "Logic bomb!  Unrecognized major test sub-set, '%s'\n" "${_major}" >&2
         rc=1
     fi
 done
 
 if [[ ${count} -eq 0 && "${subtst}" != "python" ]]; then
-    printf -- "Error - unrecognized sub-test, '%s'" "${subtst}" >&2
+    printf -- "Error - unrecognized sub-test, '%s'\n" "${subtst}" >&2
     rc=1
 fi
 
--- a/server/bin/unittests
+++ b/server/bin/unittests
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# Main unit/functional test driver for server legacy tests.  All the actual
+# test running code is located in `utils/run-unittest`.
+
 # We force the umask to a common value so that unit test output matches on all
 # platforms (well, mostly matches on all platforms).
 umask 0002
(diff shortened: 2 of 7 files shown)


dbutenhof (commented):
A lot to unpack here, but I'll drop some initial comments and look again later...

portante (commented):
Updated with a rebase on PR #2551, which in turn, is based on PR #2663 now.  So Most of the files being changed here are from those PRs.  Once those two PRs land, and this PR is rebase directly on `main`, we'll be ready for a review.

The PR will still have multiple commits in order to facilitate making the reviews "easier".

portante:
Now that PR #2551 and #2663 were merged, we have one other PR we need merged first, PR #2682.

portante (commented):
One #2682 lands we'll be ready for this PR to be reviewed.

portante (commented):
Ready for review!

exec-unittests
```diff
@@ -102,25 +132,45 @@ _subtst_list="tool-scripts/datalog tool-scripts/postprocess tool-scripts util-sc
 let count=0
 for _major in ${major_list}; do
     if [[ "${_major}" == "agent" ]]; then
+        # The parallel program is really cool, but what a pain that it posts this
+        # citation request.
```
dbutenhof, line 136: Yikes!

> --citation
> Print the citation notice and BibTeX entry for GNU parallel, silence citation notice for all future runs, and exit. It will not run any commands.
> 
> If it is impossible for you to run --citation you can instead use --will-cite, which will run commands, but which will only silence the citation notice for this single run.
> 
> If you use --will-cite in scripts to be run by others you are making it harder for others to see the citation notice. The development of GNU parallel is indirectly financed through citations, so if your users do not know they should cite then you are making it harder to finance development. However, if you pay 10000 EUR, you have done your part to finance future development and should feel free to use --will-cite in scripts.
> 
> If you do not want to help financing future development by letting other users see the citation notice or by paying, then please consider using another tool instead of GNU parallel. You can find some of the alternatives in man parallel_alternatives.
> 

Can we use `--will-cite`??

dbutenhof (approved):
The massive code relocations are difficult to review, but I'm assuming it's really just moved and I didn't review those commits line by line.

The GNU parallel "citation" adware is annoying, but that's kinda GNU; would the documented `--will-cite` be cleaner than manually touching a file you presumably expect it to use to track this?

exec-unittests
```diff
@@ -39,24 +39,54 @@ if [[ -n ${PBENCH_UNITTEST_PARALLEL} ]]; then
...
 
 function _time {
     /usr/bin/time --format="\n\nCommand: '%C'\nExit status: %x\nTimings: user %Us, system %Ss, elapsed %es (%E, %P)\nMemory: max RSS %Mk, minor pf: %R, major pf: %F, swaps %W\nContext switches: inv %c, vol %w, signals %k\nI/O: fs in %I, fs out %O, socket in %r, socket out %s\n" ${@}
 }
 
 function run_legacy {
-    printf -- "\n\n\nRunning %s/%s legacy unit tests\n\n" "${1}" "${2}"
     local _rc
-    _time ${1}/${2}/unittests ${posargs}
+    local _posargs
```
webbnh, line 59: Is `_posargs` being used?  (Should it be?  🙂)

exec-unittests
```diff
@@ -102,25 +132,45 @@ _subtst_list="tool-scripts/datalog tool-scripts/postprocess tool-scripts util-sc
...
             fi
         done
+        for _subtst in ${_subtst_list}; do
+            if [[ -z "${subtst}" || "${subtst}" == "$(basename ${_subtst})" ]]; then
+                echo ${_major} ${_subtst} ${posargs}
+            fi
+        done | parallel -k ${para_jobs_arg} para_run_legacy
+        if [[ ${?} -ne 0 ]]; then
+            rc=${?}
+        fi
```
webbnh, line 160: I think we should be setting `rc` to `1`, here -- aside from the fact that we probably don't care _how many_ jobs failed, the fact that it might be overwritten with a different, subsequent failure value means that we cannot count on it being representative of more than a single bit of resolution.

server/bin/unittests
```diff
@@ -327,34 +331,16 @@ else
...
-
-trap "kill -KILL -${$}" INT TERM QUIT
+# The parallel program is really cool, but what a pain that it posts this
+# citation request.
+if [[ -z "${HOME}" ]]; then
+    # If `HOME` is not set, the `parallel` program looks here. Don't ask.
+    _para_dir=/tmp/.parallel
+else
+    _para_dir=${HOME}/.parallel
+fi
```
webbnh, line 341: Perhaps
```
_para_dir=${HOME:-/tmp/}.parallel
```

server/bin/unittests
```diff
@@ -365,28 +351,18 @@ for testname in ${tests}; do
...
-        report_test_result ${testname}
-        if [[ ${?} -ne 0 ]]; then
-            let failures=${failures}+1
+        cmd=${cmds[${testname}]}
+        if [[ -n "${cmd}" ]]; then
+            echo "${testname} ${cmd}"
         fi
-    done
+    done | parallel -k ${jobs_arg} ${_tdir}/utils/run-unittest
+    failures=${?}
```
webbnh, line 365: I think that, if `$?` is non-zero, we should set `failures` to `1` rather than copying the exit value of `parallel`.

server/bin/utils/run-unittest
```diff
@@ -580,8 +581,29 @@ function _run_test {
     _dump_logs
     _verify_output ${testname}
     _reset_state ${testname}
-    return 0
+
+    pass_or_fail=$(cat ${_testres})
```
webbnh, line 585: I think `$(<` is preferable to `$(cat`.

Ditto for lines 587 and 601.

server/bin/utils/run-unittest
```diff
@@ -580,8 +581,29 @@ function _run_test {
...
+    if [[ ${pass_or_fail} == "PASS" ]]; then
+        echo "${pass_or_fail} - ${testname} ($(cat ${_testdur}))"
+        rm ${_testres}
+        rm -f ${_testdur}
+        rm -f ${_test_db}
+        rmdir ${_testroot}
```
webbnh, line 591: Do you really want all of these in separate commands?  Given that this is the `PASS` case, wouldn't `rm -rf ${_testroot}` do what we want?  If not, can't we combine the first three into a single command?...why does the first omit the `-f`?

server/bin/utils/run-unittest
```diff
@@ -580,8 +581,29 @@ function _run_test {
...
+        if [[ -e ${_testdiff} ]]; then
+            cat ${_testdiff}
+        else
+            echo "+++"
+            echo "*** ${testname} failed with no output"
+            echo "---"
+        fi
+        echo "FAIL - ${testname} ($(cat ${_testdur}))"
+        res=1
+        rm -f ${_testdur}
```
webbnh, line 603: What about `_testres` and/or `_test_db`?

webbnh (dismissed):
Just small stuff.

exec-unittests
webbnh: I feel like we should expand this code comment to point out that this usage of `parallel` is internal and automated; that we are not distributing product code which depends on this tool; and that we, as developers, have viewed the citation and therefore are justified suppressing future displays of it in our development processes.

portante:
FWIW, I kicked off a rebuild of the Jenkins job and it completed and updated the job status.

server/bin/unittests
```diff
@@ -15,605 +18,31 @@ unset _PBENCH_AGENT_CONFIG
...
-    rm -rf $_testroot/opt
-    if [[ -d $_testroot/opt ]]; then
-        echo "ERROR: unable to remove opt hierarchy" >&2
-        exit 1
-    fi
+    echo ${WORK}
 }
+PATH=$(remove_path /opt/pbench-server/bin ${PATH})
+PATH=$(remove_path /opt/pbench-agent/bench-scripts ${PATH})
+export PATH=$(remove_path /opt/pbench-agent/util-scripts ${PATH})
```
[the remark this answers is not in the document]
portante: This is existing code, but will add a comment.

server/bin/utils/run-unittest
```diff
@@ -0,0 +1,587 @@
...
+export _tlib=${_tdir}/../lib
+
+function _run {
+    tname=${1}
+    shift
+    if [[ -z "${@}" ]]; then
+        echo "+++ Running ${tname}" >> ${_testout}
+    else
+        echo "+++ Running ${tname} ${@}" >> ${_testout}
+    fi
```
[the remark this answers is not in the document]
portante: I was trying to avoid disturbing the gold files which fail when the extra space is added.

server/bin/utils/run-unittest
```diff
@@ -583,5 +531,57 @@ function _reset_state {
...
+    mkdir -p $_testroot
+    if [[ ! -d $_testroot ]]; then
+        echo "ERROR: failed to create test root directory, \"$_testroot\"" >&2
+        exit 1
+    fi
+    rm -rf $_testroot/*
+    if [[ $? -gt 0 ]]; then
+        echo "ERROR: failed to empty test root directory, \"$_testroot\"" >&2
+        exit 1
+    fi
```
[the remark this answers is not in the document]
portante: Fixed in a later commit.

exec-unittests
```diff
@@ -39,24 +39,54 @@ if [[ -n ${PBENCH_UNITTEST_PARALLEL} ]]; then
...
     local _rc
-    _time ${1}/${2}/unittests ${posargs}
+    local _posargs
+    local _major
+    local _subtst
+
+    _major=${1}
+    shift
+    _subtst=${1}
+    shift
```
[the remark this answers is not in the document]
portante: We ran into this in the past, I believe.  If you have `shift 2` but don't have two arguments then the behavior is odd.

exec-unittests
```diff
@@ -102,25 +132,45 @@ _subtst_list="tool-scripts/datalog tool-scripts/postprocess tool-scripts util-sc
...
             if [[ -z "${subtst}" || "${subtst}" == "$(basename ${_subtst})" ]]; then
                 (( count++ ))
-                run_legacy ${_major} ${_subtst} ${posargs} || rc=1
             fi
         done
+        for _subtst in ${_subtst_list}; do
+            if [[ -z "${subtst}" || "${subtst}" == "$(basename ${_subtst})" ]]; then
+                echo ${_major} ${_subtst} ${posargs}
+            fi
+        done | parallel -k ${para_jobs_arg} para_run_legacy
```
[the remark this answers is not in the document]
portante: You can't, which is what the comment refers to.  If you combine the loops the count is lost because it is executed in a sub-shell.

server/bin/utils/run-unittest
portante: We wanted to see when things were left around or something was missing.  In particular, the `rmdir` should be cleaning up an empty directory.

server/bin/utils/run-unittest
portante: Yes, so that one can go back and look at the data related to the failure.

exec-unittests
```diff
@@ -74,7 +80,9 @@ if [[ -z "${subtst}" || "${subtst}" == "python" ]]; then
     printf -- "\n\n\nRunning %s python3-based unit tests via pytest\n\n" "${major_list// /,}"
     _pbench_sources=$(python3 -c 'import inspect, pathlib, pbench; print(pathlib.Path(inspect.getsourcefile(pbench)).parent.parent)')
     _PBENCH_COV_DIR="${_toxenvdir}/cov" _time pytest \
+        -n ${parallel} \
         --basetemp="${_toxenvdir}/tmp" \
+        --confcutdir=".." \
```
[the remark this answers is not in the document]
portante: Removing this did not change test behavior and they all passed.

server/bin/utils/run-unittest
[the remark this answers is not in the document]
portante: Tests failed with white-space mismatch.  Would rather not fix that in this commit since this code is just a move of the existing behavior.

exec-unittests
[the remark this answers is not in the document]
dbutenhof: > n must be a non-negative number less than or equal to $#.  If n is 0, no parameters are changed.  If n is not given, it is assumed  to  be  1.   _If  n  is greater than $#, the positional parameters are not changed._

exec-unittests
```diff
@@ -94,25 +126,45 @@ _subtst_list="tool-scripts/datalog tool-scripts/postprocess tool-scripts util-sc
...
+        # The parallel program is really cool.  The usage of `parallel` is
+        # internal and automated; we are not disturbing product code to depend
+        # on this tool, and we, as developers, have viewed the citation and are
+        # justified in suppressing future displays of it in our development
+        # processes.
+        #
+        # If `HOME` is not set, the `parallel` program looks here. Don't ask.
+        _para_dir=${HOME:-/tmp}/.parallel
+        mkdir -p ${_para_dir}
+        touch ${_para_dir}/will-cite
```
dbutenhof, line 138: So did `--will-cite` not work? Is there documentation somewhere about creating this `will-cite` marker file? (I didn't find any, although `parallel` seems to try very hard to be opaque about suppressing the nagware.) Otherwise, it seems something like `alias parallel='parallel --will-cite'` would have been cleaner.

server/bin/unittests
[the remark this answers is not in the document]
dbutenhof: Really? I'd looked through the old code but didn't find it. In any case, thanks. 👍🏻

dbutenhof (dismissed):

server/bin/unittests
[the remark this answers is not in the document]
portante: See https://github.com/distributed-system-analysis/pbench/blob/main/server/bin/unittests#L44.

exec-unittests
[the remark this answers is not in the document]
webbnh: > If n is greater than $#, the positional parameters are not changed.

Oh, urk.

So, then...
```
    _major=${1}
    _subtst=${2}
    shift; shift
```
🤪

dbutenhof (approved):
Nice!

exec-unittests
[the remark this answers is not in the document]
webbnh: Well, gosh!

OK, I propose more specific wording for the comment, then.  E.g., something like
```
        # Count what we _will_ do since the `for` command which drives the
        # `parallel` command below will be run in a sub-shell (because it is
        # part of a "pipeline") which will make changes to `count` unavailable.
```

Alternatively, we could build the list and count it in a single loop, and then invoke `parallel` separately:
```
        let count=0
        test_list=""
        for _subtst in ${_subtst_list}; do
            test_list+="${_major} ${_subtst} ${posargs}"$'\n'
            (( count++ ))
        done
        parallel --lb -k ${para_jobs_arg} para_run_legacy <<< ${test_list}
```

exec-unittests
```diff
@@ -132,31 +126,26 @@ _subtst_list="tool-scripts/datalog tool-scripts/postprocess tool-scripts util-sc
...
-        else
-            _para_dir=${HOME}/.parallel
-        fi
-        mkdir -p ${_para_dir}
-        touch ${_para_dir}/will-cite
+        # The parallel program is really cool.  The usage of `parallel` is
+        # internal and automated; we are not disturbing product code to depend
+        # on this tool, and we, as developers, have viewed the citation and are
+        # justified in suppressing future displays of it in our development
+        # processes (use of --will-cite below).
```
webbnh, line 133: s/disturbing/distributing/  😁

Also, I would change "to depend" to "which depends".

server/bin/utils/run-unittest
```diff
@@ -0,0 +1,607 @@
...
+        echo "ERROR: unable to remove opt hierarchy" >&2
+        exit 1
+    fi
+}
+
+function _run_test {
+    # What it takes to run one test
+    local testname=${1}
+    local cmd=${2}
+    shift 2
```
webbnh, line 542: As noted elsewhere, the `shift 2` is potentially problematic...should we replace it with `shift; shift`?

server/bin/utils/run-unittest
```diff
@@ -0,0 +1,607 @@
...
+
+    export _testroot="${_testbase}/${testname}"
+    rm -rf ${_testroot}
+    mkdir -p ${_testroot}
```
webbnh, line 546: How are we feeling today about the `-p`?  (I have no problem with it, but lately we've been removing it elsewhere.)

webbnh (approved):
Looks good; there are just nits remaining:
- in a couple of places, `shift 2` vs. `shift; shift` or separate shifts;
- some typos and other editorials in code comments; and,
- some other questions/suggestions.

server/bin/utils/run-unittest
portante: This is case it is not because `server/bin/unittests` always/must invoke `run-unittest` with the test name and the command to run, the additional arguments are optional.

server/bin/utils/run-unittest
portante: Okay, so because of this prompt, I looked at the other uses of `mkdir -p` and found that we were unnecessarily creating a `common` directory.  I've added a separate commit that addresses that along with removing the remaining uses of `mkdir -p` that hide errors.

dbutenhof (dismissed):

dbutenhof (dismissed):

webbnh (dismissed):
I think you may have removed a `-p` which we would be better to keep, and I think the wording in the `parallel` comment should have a tweak...but, we can negotiate on those.  Also, I'm still wondering what your thoughts are on a simplification that I offered in the last round.

exec-unittests
```diff
@@ -132,31 +126,26 @@ _subtst_list="tool-scripts/datalog tool-scripts/postprocess tool-scripts util-sc
...
 
         # Count what we will do since an implicit sub-shell is used when
         # piping `parallel` below.
         for _subtst in ${_subtst_list}; do
-            if [[ -z "${subtst}" || "${subtst}" == "$(basename ${_subtst})" ]]; then
+            if [[ -z "${subtst}" || "${subtst}" == "legacy" || "${subtst}" == "$(basename ${_subtst})" ]]; then
```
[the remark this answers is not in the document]
webbnh: Sounds good; so what do you think of the simplification I offered above?

server/bin/utils/run-unittest
```diff
@@ -243,47 +243,47 @@ function _create_user {
 function _setup_state {
     local rc=0
     local res=0
-    mkdir -p ${_testopt}/unittest-scripts/
+    mkdir ${_testroot}/opt
     let res=res+${?}
-    mkdir -p ${_testopt_sat}/unittest-scripts/
+    mkdir ${_testopt} ${_testopt_sat}
```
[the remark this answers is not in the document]
portante: I believe the `-p` will hide unexpected directory creations that should have been properly setup.

And yes one can argue that the relationship is setup elsewhere, but it is really this script, and so I'll create a proper variable to encapsulate that.

dbutenhof (dismissed):

dbutenhof (dismissed):

portante (commented):
Just removed the commits used to verify Jenkins integration was working.

dbutenhof (dismissed):
Maybe for the final time? 😆

jenkins/branch.name
```diff
@@ -1 +1 @@
-main
+tox-parallel
```
dbutenhof, line 1: OK, I guess we're *not* done yet...

webbnh (approved):
Good to go...except for the "DO NOT MERGE" commit.

webbnh:
Oh, and, can we have a better description for the PR than "TBD"?  😀

webbnh (approved):
👍

dbutenhof (approved):

PR #2652, opened 2022-02-28
Remove `BaseCommand` environment manipulations

### LINKED ISSUE #2647: The `py3-server` tox environment does not work
When using `tox -e py3-server`, the tests fail with hundreds of errors like the following:

```
    @pytest.fixture
    def pbench_run(monkeypatch, tmp_path):
>       monkeypatch.delenv("pbench_run")
E       KeyError: 'pbench_run'

lib/pbench/test/functional/conftest.py:21: KeyError
```

portante:
There was some thinking that the Tool Meister sub-system needed various environment variables set in the `BaseCommand` class.  Because of how the `pytest` unit tests are run, where the `os.environ` dictionary is shared across tests, these variables would show up for any test that ran subsequent to a test that invoked a `BaseCommand` sub-class.

We experimented with removing them entirely, and could not find any test that failed as a result.  This also resolves the inadvertent problem introduced by PR #2606 which caused `tox -e py3-server` runs to fail.

Fixes #2647.


webbnh (approved):
I'm in favor of this change...I just hope that we can get it working without too much pain:
- for instance, [the tool group commands](https://github.com/distributed-system-analysis/pbench/blob/0dabb2392cacb46103fed7757426bc49319b18f5/lib/pbench/agent/tool_group.py#L34) expect `pbench_run` to be defined in the environment if it isn't passed into `verify_tool_group()`; and
- the [Tool Data Sync](https://github.com/distributed-system-analysis/pbench/blob/0dabb2392cacb46103fed7757426bc49319b18f5/lib/pbench/agent/tool_data_sink.py#L2022) expects to be able to find `pbench_run` in the environment, too.

As an aside, it seems like the `BaseCommand` constructor could use some cleanup:  it's maintaining its own definitions `pbench_run`, `pbench_tmp`, `pbench_bspp_dir`, `pbench_lib_dir` (et al.?) when it should probably be copying them from `PbenchAgentConfig`'s; and, if we really want the ability to override the location of things like `pbench_run`, somebody should probably tell `PbenchAgentConfig` that....

Issue #1754, opened 2020-08-14
`tool-scripts/datalog` unit tests not cleaning up processes correctly


portante:
It seems that after running the `agent/tool-scripts/datalog/unittests` we have lingering processes left over from the tests.

```
$ ps jx
   PPID     PID    PGID     SID TTY        TPGID STAT   UID   TIME COMMAND
      1 2925492 2916425 1968408 pts/2    2939865 S     1000   0:00 /bin/bash ${pbench_dir}/agent/tool-scripts/datalog/mock-bin/oc get --all-namespaces ep -o wide -w
      1 2925494 2916425 1968408 pts/2    2939865 S     1000   0:00 /usr/bin/perl ${pbench_dir}/agent/tool-scripts/datalog/mock-bin/pbench-log-timestamp
      1 2925497 2916425 1968408 pts/2    2939865 S     1000   0:00 /bin/bash ${pbench_dir}/agent/tool-scripts/datalog/mock-bin/oc get --all-namespaces pods -o wide -w
      1 2925499 2916425 1968408 pts/2    2939865 S     1000   0:00 /usr/bin/perl ${pbench_dir}/agent/tool-scripts/datalog/mock-bin/pbench-log-timestamp
2925497 2925502 2916425 1968408 pts/2    2939865 S     1000   0:00 /bin/sleep 120
2925492 2925506 2916425 1968408 pts/2    2939865 S     1000   0:00 /bin/sleep 120
      1 2925507 2916425 1968408 pts/2    2939865 S     1000   0:00 /bin/sleep 120
```

PR #2653, opened 2022-03-01
Fix `oc-datalog` to properly reap subprocesses

portante:
The `oc-datalog` was not properly reaping its sub-processes, which left lingering processes after `tox -e datalog` test runs.  Since they would eventually timeout, it wasn't really noticeable.

 We also fix the `mock-bin/oc` command to exec the `sleep 120` process so that it is properly reaped (the real `oc` handles its own sub-process clean up).


dbutenhof (approved):

webbnh (approved):
Good catch!  😀

portante:
Fixes #1754.

Issue #2347, opened 2021-06-08
pbench-uperf complains about missing tool output directory


aokuliar-rh:
Version: `pbench-agent-0.71.0-185g5527f23f1.noarch`

```
# pbench-uperf -t bidirec -S vales2.slevarna.tpb.lab.eng.brq.redhat.com
.
.
.
iostat: WARNING - postprocess - tool output directory, '/var/lib/pbench-agent/uperf__2021.06.08T12.46.44/3-tcp_bidirec-1B-64i/sample2/tools-default/vales1.slevarna.tpb.lab.eng.brq.redhat.com/iostat', missing
iostat: WARNING - postprocess - tool output directory, '/var/lib/pbench-agent/uperf__2021.06.08T12.46.44/3-tcp_bidirec-1B-64i/sample2/tools-default/vales1.slevarna.tpb.lab.eng.brq.redhat.com/iostat', missing
mpstat: WARNING - postprocess - tool output directory, '/var/lib/pbench-agent/uperf__2021.06.08T12.46.44/3-tcp_bidirec-1B-64i/sample2/tools-default/vales1.slevarna.tpb.lab.eng.brq.redhat.com/mpstat', missing
iostat: WARNING - postprocess - tool output directory, '/var/lib/pbench-agent/uperf__2021.06.08T12.46.44/3-tcp_bidirec-1B-64i/sample2/tools-default/vales1.slevarna.tpb.lab.eng.brq.redhat.com/iostat', missing
mpstat: WARNING - postprocess - tool output directory, '/var/lib/pbench-agent/uperf__2021.06.08T12.46.44/3-tcp_bidirec-1B-64i/sample2/tools-default/vales1.slevarna.tpb.lab.eng.brq.redhat.com/mpstat', missing
perf: WARNING - postprocess - tool output directory, '/var/lib/pbench-agent/uperf__2021.06.08T12.46.44/3-tcp_bidirec-1B-64i/sample2/tools-default/vales1.slevarna.tpb.lab.eng.brq.redhat.com/perf', missing
iostat: WARNING - postprocess - tool output directory, '/var/lib/pbench-agent/uperf__2021.06.08T12.46.44/3-tcp_bidirec-1B-64i/sample2/tools-default/vales1.slevarna.tpb.lab.eng.brq.redhat.com/iostat', missing
[...]

PR #2659, opened 2022-03-02
Terminating benchmark run if tool rpm does not exist

riya-17:
Fixes Issue #2347 

Stop the Tool Meister if the required RPMs are not installed on the hosts, and output the logs around the Failure.

### DIFF
--- a/agent/util-scripts/gold/test-client-tool-meister/test-61.txt
+++ b/agent/util-scripts/gold/test-client-tool-meister/test-61.txt
@@ -1,0 +1,289 @@
++++ Running test-61 test-client-tool-meister lite with-remotes
+"mpstat" tool is now registered for host "testhost.example.com" in group "lite"
+"dcgm" tool is now registered for host "testhost.example.com" in group "lite"
+"mpstat" tool is now registered for host "remote-a.example.com" in group "lite"
+"mpstat" tool is now registered for host "remote-b.example.com", with label "blue", in group "lite"
+"node-exporter" tool is now registered for host "remote-b.example.com", with label "blue", in group "lite"
+"dcgm" tool is now registered for host "remote-c.example.com", with label "red", in group "lite"
+"pcp" tool is now registered for host "remote-c.example.com", with label "red", in group "lite"
+pbench-tool-meister-start: Error: host: remote-a.example.com; tool: mpstat; Run-Failure-Info: mpstat: pbench-sysstat-12.0.3-1 is not installed
+pbench-tool-meister-start: Error: host: remote-b.example.com; tool: mpstat; Run-Failure-Info: mpstat: pbench-sysstat-12.0.3-1 is not installed
+pbench-tool-meister-start: Error: host: testhost.example.com; tool: mpstat; Run-Failure-Info: mpstat: pbench-sysstat-12.0.3-1 is not installed
+pbench-tool-meister-start: error TDS didn't confirm init sequence completion (None)
+ERROR - "pbench-tool-meister-start --sysinfo='block,security_mitigations,sos' 'lite'" failed to execute successfully (exit code: 21)
+--- Finished test-61 test-client-tool-meister (status=1)
++++ pbench tree state
+/var/tmp/pbench-test-utils/pbench
+/var/tmp/pbench-test-utils/pbench/mock-run
+/var/tmp/pbench-test-utils/pbench/mock-run/metadata.log
+/var/tmp/pbench-test-utils/pbench/mock-run/ssh_config
+/var/tmp/pbench-test-utils/pbench/mock-run/ssh_config.d
+/var/tmp/pbench-test-utils/pbench/mock-run/tm
+/var/tmp/pbench-test-utils/pbench/mock-run/tm/pbench-tool-data-sink.err
+/var/tmp/pbench-test-utils/pbench/mock-run/tm/pbench-tool-data-sink.out
+/var/tmp/pbench-test-utils/pbench/mock-run/tm/redis.conf
+/var/tmp/pbench-test-utils/pbench/mock-run/tm/redis.log
+/var/tmp/pbench-test-utils/pbench/mock-run/tm/tm-lite-testhost.example.com.err
+/var/tmp/pbench-test-utils/pbench/mock-run/tm/tm-lite-testhost.example.com.out
+/var/tmp/pbench-test-utils/pbench/mock-run/tm/tm.logs
+/var/tmp/pbench-test-utils/pbench/pbench.log
+/var/tmp/pbench-test-utils/pbench/tmp
+/var/tmp/pbench-test-utils/pbench/tmp/tm-lite-remote-a.example.com.err
+/var/tmp/pbench-test-utils/pbench/tmp/tm-lite-remote-a.example.com.out
+/var/tmp/pbench-test-utils/pbench/tmp/tm-lite-remote-b.example.com.err
+/var/tmp/pbench-test-utils/pbench/tmp/tm-lite-remote-b.example.com.out
+/var/tmp/pbench-test-utils/pbench/tmp/tm-lite-remote-c.example.com.err
+/var/tmp/pbench-test-utils/pbench/tmp/tm-lite-remote-c.example.com.out
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-a.example.com
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-a.example.com/mpstat
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-b.example.com
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-b.example.com/__label__
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-b.example.com/mpstat
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-b.example.com/node-exporter
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-c.example.com
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-c.example.com/__label__
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-c.example.com/dcgm
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-c.example.com/pcp
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite/testhost.example.com
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite/testhost.example.com/dcgm
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite/testhost.example.com/mpstat
+=== /var/tmp/pbench-test-utils/pbench/tmp/tm-lite-remote-a.example.com.err:
+ERROR pbench-tool-meister fetch_message -- lost connection to redis server on channel pbench-agent-cli-to-tms
+INFO pbench-tool-meister __exit__ -- remote-a.example.com: terminating
+ERROR pbench-tool-meister _send_client_status -- Failed to publish client status message, '{"hostname": "remote-a.example.com", "kind": "tm", "status": "terminated"}': Error 111 connecting to localhost:17001. Connection refused.
+WARNING pbench-tool-meister driver -- RedisHandler redis_errors: 0, errors: 3, dropped: 0
+=== /var/tmp/pbench-test-utils/pbench/tmp/tm-lite-remote-a.example.com.out:
+=== /var/tmp/pbench-test-utils/pbench/tmp/tm-lite-remote-b.example.com.err:
+ERROR pbench-tool-meister fetch_message -- lost connection to redis server on channel pbench-agent-cli-to-tms
+INFO pbench-tool-meister __exit__ -- remote-b.example.com: terminating
+ERROR pbench-tool-meister _send_client_status -- Failed to publish client status message, '{"hostname": "remote-b.example.com", "kind": "tm", "status": "terminated"}': Error 111 connecting to localhost:17001. Connection refused.
+WARNING pbench-tool-meister driver -- RedisHandler redis_errors: 0, errors: 3, dropped: 0
+=== /var/tmp/pbench-test-utils/pbench/tmp/tm-lite-remote-b.example.com.out:
+=== /var/tmp/pbench-test-utils/pbench/tmp/tm-lite-remote-c.example.com.err:
+ERROR pbench-tool-meister fetch_message -- lost connection to redis server on channel pbench-agent-cli-to-tms
+INFO pbench-tool-meister __exit__ -- remote-c.example.com: terminating
+ERROR pbench-tool-meister _send_client_status -- Failed to publish client status message, '{"hostname": "remote-c.example.com", "kind": "tm", "status": "terminated"}': Error 111 connecting to localhost:17001. Connection refused.
+WARNING pbench-tool-meister driver -- RedisHandler redis_errors: 0, errors: 3, dropped: 0
+=== /var/tmp/pbench-test-utils/pbench/tmp/tm-lite-remote-c.example.com.out:
+=== /var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-a.example.com/mpstat:
+--interval=42
+--options=forty-two
+=== /var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-b.example.com/__label__:
+blue
+=== /var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-b.example.com/mpstat:
+--interval=42
+--options=forty-two
+=== /var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-b.example.com/node-exporter:
+--inst=/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts
+=== /var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-c.example.com/__label__:
+red
+=== /var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-c.example.com/dcgm:
+--inst=/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts
+=== /var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-c.example.com/pcp:
+--interval=42
+--options=forty-two
+=== /var/tmp/pbench-test-utils/pbench/tools-v1-lite/testhost.example.com/dcgm:
+--inst=/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts
+=== /var/tmp/pbench-test-utils/pbench/tools-v1-lite/testhost.example.com/mpstat:
+--interval=42
+--options=forty-two
+--- pbench tree state
++++ pbench.log file contents
+[debug][1900-01-01T00:00:00.000000] tool_opts: "--interval=42 --options=forty-two"
+[info][1900-01-01T00:00:00.000000] "mpstat" tool is now registered for host "testhost.example.com" in group "lite"
+[debug][1900-01-01T00:00:00.000000] tool_opts: "--inst=/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts"
+[info][1900-01-01T00:00:00.000000] "dcgm" tool is now registered for host "testhost.example.com" in group "lite"
+[debug][1900-01-01T00:00:00.000000] tool_opts: "--interval=42 --options=forty-two"
+[info][1900-01-01T00:00:00.000000] "mpstat" tool is now registered for host "remote-a.example.com" in group "lite"
+[debug][1900-01-01T00:00:00.000000] tool_opts: "--interval=42 --options=forty-two"
+[info][1900-01-01T00:00:00.000000] "mpstat" tool is now registered for host "remote-b.example.com", with label "blue", in group "lite"
+[debug][1900-01-01T00:00:00.000000] tool_opts: "--inst=/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts"
+[info][1900-01-01T00:00:00.000000] "node-exporter" tool is now registered for host "remote-b.example.com", with label "blue", in group "lite"
+[debug][1900-01-01T00:00:00.000000] tool_opts: "--inst=/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts"
+[info][1900-01-01T00:00:00.000000] "dcgm" tool is now registered for host "remote-c.example.com", with label "red", in group "lite"
+[debug][1900-01-01T00:00:00.000000] tool_opts: "--interval=42 --options=forty-two"
+[info][1900-01-01T00:00:00.000000] "pcp" tool is now registered for host "remote-c.example.com", with label "red", in group "lite"
+--- pbench.log file contents
++++ mock-run/metadata.log file contents
+[pbench]
+config = 
+date = 1900-01-01T00:00:00
+name = mock-run
+rpm-version = v(unknown)-g(unknown)
+script = fake-bm
+
+[controller]
+hostname = testhost.example.com
+hostname-alias = bond
+hostname-all-fqdns = agent.example.com bond.example.com
+hostname-all-ip-addresses = 192.168.0.1 172.0.0.1 127.0.0.1
+hostname-domain = example.com
+hostname-fqdn = agent.example.com
+hostname-ip-address = 172.0.0.1 127.0.0.1
+hostname-nis = hostname[mock]: Local domain name not set
+hostname-short = agent
+ssh_opts = -o BatchMode=yes -o StrictHostKeyChecking=no
+
+[run]
+controller = testhost.example.com
+start_run = 1900-01-01T00:00:00.000042
+
+[tools]
+hosts = remote-a.example.com remote-b.example.com remote-c.example.com testhost.example.com
+group = lite
+trigger = None
+
+[tools/remote-a.example.com]
+label = 
+tools = mpstat
+hostname-alias = bond
+hostname-all-fqdns = agent.example.com bond.example.com
+hostname-all-ip-addresses = 192.168.0.1 172.0.0.1 127.0.0.1
+hostname-domain = example.com
+hostname-fqdn = agent.example.com
+hostname-ip-address = 172.0.0.1 127.0.0.1
+hostname-nis = hostname[mock]: Local domain name not set
+hostname-short = agent
+rpm-version = v(unknown)-g(unknown)
+mpstat = --interval=42 --options=forty-two
+
+[tools/remote-a.example.com/mpstat]
+options = --interval=42 --options=forty-two
+install_check_status_code = 1
+install_check_output = mpstat: pbench-sysstat-12.0.3-1 is not installed
+
+[tools/remote-b.example.com]
+label = blue
+tools = mpstat,node-exporter
+hostname-alias = bond
+hostname-all-fqdns = agent.example.com bond.example.com
+hostname-all-ip-addresses = 192.168.0.1 172.0.0.1 127.0.0.1
+hostname-domain = example.com
+hostname-fqdn = agent.example.com
+hostname-ip-address = 172.0.0.1 127.0.0.1
+hostname-nis = hostname[mock]: Local domain name not set
+hostname-short = agent
+rpm-version = v(unknown)-g(unknown)
+mpstat = --interval=42 --options=forty-two
+node-exporter = --inst=/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts
+
+[tools/remote-b.example.com/mpstat]
+options = --interval=42 --options=forty-two
+install_check_status_code = 1
+install_check_output = mpstat: pbench-sysstat-12.0.3-1 is not installed
+
+[tools/remote-b.example.com/node-exporter]
+options = --inst=/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts
+install_check_status_code = 0
+install_check_output = node_exporter tool properly installed
+
+[tools/remote-c.example.com]
+label = red
+tools = dcgm,pcp
+hostname-alias = bond
+hostname-all-fqdns = agent.example.com bond.example.com
+hostname-all-ip-addresses = 192.168.0.1 172.0.0.1 127.0.0.1
+hostname-domain = example.com
+hostname-fqdn = agent.example.com
+hostname-ip-address = 172.0.0.1 127.0.0.1
+hostname-nis = hostname[mock]: Local domain name not set
+hostname-short = agent
+rpm-version = v(unknown)-g(unknown)
+dcgm = --inst=/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts
+pcp = --interval=42 --options=forty-two
+
+[tools/remote-c.example.com/dcgm]
+options = --inst=/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts
+install_check_status_code = 0
+install_check_output = dcgm tool (dcgm-exporter) properly installed
+
+[tools/remote-c.example.com/pcp]
+options = --interval=42 --options=forty-two
+install_check_status_code = 0
+install_check_output = pcp tool (pmcd) properly installed
+
+[tools/testhost.example.com]
+label = 
+tools = dcgm,mpstat
+hostname-alias = bond
+hostname-all-fqdns = agent.example.com bond.example.com
+hostname-all-ip-addresses = 192.168.0.1 172.0.0.1 127.0.0.1
+hostname-domain = example.com
+hostname-fqdn = agent.example.com
+hostname-ip-address = 172.0.0.1 127.0.0.1
+hostname-nis = hostname[mock]: Local domain name not set
+hostname-short = agent
+rpm-version = v(unknown)-g(unknown)
+dcgm = --inst=/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts
+mpstat = --interval=42 --options=forty-two
+
+[tools/testhost.example.com/dcgm]
+options = --inst=/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts
+install_check_status_code = 0
+install_check_output = dcgm tool (dcgm-exporter) properly installed
+
+[tools/testhost.example.com/mpstat]
+options = --interval=42 --options=forty-two
+install_check_status_code = 1
+install_check_output = mpstat: pbench-sysstat-12.0.3-1 is not installed
+
+--- mock-run/metadata.log file contents
++++ mock-run/tm/pbench-tool-data-sink.err file contents
+
+Bottle v#.##.## server starting up (using DataSinkWsgiServer(handler_class=<class 'pbench.agent.tool_data_sink.DataSinkWsgiServer.__init__.<locals>.DataSinkWsgiRequestHandler'>))...
+ERROR pbench-tool-data-sink fetch_message -- lost connection to redis server on channel pbench-agent-cli-from-client
+Hit Ctrl-C to quit.
+INFO pbench-tool-data-sink execute -- Tool Data Sink terminating
+INFO pbench-tool-data-sink tm_log_capture -- Running Tool Meister log capture ...
+INFO pbench-tool-data-sink web_server_run -- Bottle web server exited
+INFO pbench-tool-data-sink web_server_run -- Running Bottle web server ...
+Listening on http://localhost:8080/
+Terminating on signal 15
+lost connection to redis server on channel pbench-agent-cli-to-logging
+--- mock-run/tm/pbench-tool-data-sink.err file contents
++++ mock-run/tm/pbench-tool-data-sink.out file contents
+--- mock-run/tm/pbench-tool-data-sink.out file contents
++++ mock-run/tm/redis.conf file contents
+bind localhost
+daemonize yes
+dir /var/tmp/pbench-test-utils/pbench/mock-run/tm
+save ""
+appendonly no
+dbfilename pbench-redis.rdb
+logfile /var/tmp/pbench-test-utils/pbench/mock-run/tm/redis.log
+loglevel notice
+pidfile /var/tmp/pbench-test-utils/pbench/mock-run/tm/redis_17001.pid
+port 17001
+--- mock-run/tm/redis.conf file contents
++++ mock-run/tm/redis.log file contents
+# Configuration loaded
+# Server initialized
+* Ready to accept connections
+# User requested shutdown...
+* Removing the pid file.
+# Redis is now ready to exit, bye bye...
+--- mock-run/tm/redis.log file contents
++++ mock-run/tm/tm-lite-testhost.example.com.err file contents
+ERROR pbench-tool-meister fetch_message -- lost connection to redis server on channel pbench-agent-cli-to-tms
+INFO pbench-tool-meister __exit__ -- testhost.example.com: terminating
+ERROR pbench-tool-meister _send_client_status -- Failed to publish client status message, '{"hostname": "testhost.example.com", "kind": "tm", "status": "terminated"}': Error 111 connecting to localhost:17001. Connection refused.
+WARNING pbench-tool-meister driver -- RedisHandler redis_errors: 0, errors: 3, dropped: 0
+--- mock-run/tm/tm-lite-testhost.example.com.err file contents
++++ mock-run/tm/tm-lite-testhost.example.com.out file contents
+--- mock-run/tm/tm-lite-testhost.example.com.out file contents
++++ mock-run/tm/tm.logs file contents
+pbench-tool-meister-start - verify logging channel up
+--- mock-run/tm/tm.logs file contents
++++ test-execution.log file contents
+/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/cp -rL /etc/ssh/ssh_config.d /var/tmp/pbench-test-utils/pbench/mock-run/
+/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/rpm --query --queryformat=%{EVR}\n pbench-sysstat
+/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/rpm --query --queryformat=%{EVR}\n pbench-sysstat
+/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/rpm --query --queryformat=%{EVR}\n pbench-sysstat
+/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/ssh -o BatchMode=yes -o StrictHostKeyChecking=no remote-a.example.com /var/tmp/pbench-test-utils/opt/pbench-agent/util-scripts/tool-meister/pbench-tool-meister-remote localhost 17001 tm-lite-remote-a.example.com yes
+/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/ssh -o BatchMode=yes -o StrictHostKeyChecking=no remote-a.example.com echo ${SSH_CONNECTION}
+/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/ssh -o BatchMode=yes -o StrictHostKeyChecking=no remote-b.example.com /var/tmp/pbench-test-utils/opt/pbench-agent/util-scripts/tool-meister/pbench-tool-meister-remote localhost 17001 tm-lite-remote-b.example.com yes
+/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/ssh -o BatchMode=yes -o StrictHostKeyChecking=no remote-b.example.com echo ${SSH_CONNECTION}
+/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/ssh -o BatchMode=yes -o StrictHostKeyChecking=no remote-c.example.com /var/tmp/pbench-test-utils/opt/pbench-agent/util-scripts/tool-meister/pbench-tool-meister-remote localhost 17001 tm-lite-remote-c.example.com yes
+/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/ssh -o BatchMode=yes -o StrictHostKeyChecking=no remote-c.example.com echo ${SSH_CONNECTION}
+--- test-execution.log file contents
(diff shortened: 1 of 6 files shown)


lib/pbench/agent/tool_data_sink.py
```diff
@@ -1259,9 +1267,10 @@ def execute(self):
             self._tm_tracking = self.record_tms(tms)
             self._num_tms = len(self._tm_tracking.keys())
 
+            success = self.check_tms(tms)
```
dbutenhof, line 1270: `status` would be a better name, since the value may be "success" or "failure".

agent/tool-scripts/base-tool
```diff
@@ -579,8 +579,10 @@ install)
 	function check_required_rpm {
 		local _tpn="${1}"
 		local _tpv="${2}"
-		local _installed_rpm="$(require-rpm "${_tpn}" "${_tpv}")"
-		local rc=${?}
+		local _installed_rpm
+		local rc
+    _installed_rpm="$(require-rpm "${_tpn}" "${_tpv}")"
+    rc=${?}
```
dbutenhof, line 585: These lines should be indented to align with the other function statements.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1242,6 +1242,14 @@ def _valid_data(self, data):
             else:
                 return (action, directory, args)
 
+    @staticmethod
+    def check_tms(tms):
+
+        for host, tm in tms.items():
+            if len(tm["failed_tools"]) > 0:
+                return "failure"
+        return "success"
```
dbutenhof, line 1251: `check_tms` is a fairly generic name. A couple of comments:

1. The name should suggest something closer to what it does: looking for failed tool initialization checks.
2. Add a docstring
3. Type hints in the declaration
4. While it's convenient that this returns the strings we're passing over Redis, the method would be more general purpose if it returned `True`/`False` with the caller translating into the string required for our TDS protocol.

dbutenhof (commented):
Great start on finding and fixing some of the basic problems here.

Are you planning to make changes on the TDS and `pbench-tool-meister-start` sides to report the tool init failure in detail (which I had the impression isn't happening now), or would that be deferred to a separate PR?

agent/tool-scripts/base-tool
```diff
@@ -579,8 +579,10 @@ install)
 	function check_required_rpm {
 		local _tpn="${1}"
 		local _tpv="${2}"
-		local _installed_rpm="$(require-rpm "${_tpn}" "${_tpv}")"
-		local rc=${?}
+		local _installed_rpm
+		local rc
```
webbnh, line 583: Since we seem to have the convention of marking local variables with a leading underscore, it would probably be best to apply that to `rc` as well.  (I realize that you're not responsible for this lack, but, since you're here....)

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1259,9 +1267,10 @@ def execute(self):
             self._tm_tracking = self.record_tms(tms)
             self._num_tms = len(self._tm_tracking.keys())
 
+            success = self.check_tms(tms)
             # Tell the entity that started us who we are, indicating we're
             # ready, and all the TMs are ready.
-            started_msg = dict(kind="ds", action="startup", status="success")
+            started_msg = dict(kind="ds", action="startup", status=success)
```
webbnh, line 1273: Now that `status` might be something other than `"success"`, don't we need some additional code here (for the case where it's not `"success"`)?  (And, I think the code comment needs some related tweaking....)

webbnh (commented):
I second Dave's comments:  this is good start, but I think we need a little bit more, here.

ndokos:
The main question left is what to do about failures. My initial reaction would be for the TDS to include the `tms` dict  in the message to `pbench-tool-meister-start` , but maybe only in the failure case. It would be the responsibility of `pbench-tool-meister-start` to check in the case of failure that

- the `tms` dict is part of the message
- trawl through the dict to format an error message to present to the user.

This assumes that `tms` has *all* the information necessary.

webbnh:
> include the `tms` dict in the message

Of course, this is not quite trivial.  In order to pass a dict as a string, it needs to be "marshaled" somehow -- e.g., converted to stringized JSON -- and the other end has to know to "unmarshal" it.  So, we're going to need to develop a bit of a protocol for this.

ndokos:
IIUC, nothing is needed: the current code already serializes  `start_msg` which is a (very simple) dict, using `json.dumps()`. Adding a nested dict as another item should not make any difference. The other side fetches the message and deserializes it.

riya-17:
> Great start on finding and fixing some of the basic problems here.
> 
> Are you planning to make changes on the TDS and `pbench-tool-meister-start` sides to report the tool init failure in detail (which I had the impression isn't happening now), or would that be deferred to a separate PR?

I was planning, those but there were many questions around how we should go around them, so, preferred to put up a PR and then discuss around them.

agent/tool-scripts/base-tool
riya-17: I couldn't see any indentation issues in my `vim`, Maybe due to some other issues. will make the change :)

webbnh: The adjacent lines appear to be using (two) hard-tab characters for indentation, while these lines are using (4) blank space characters.

The file does have the
```
# -*- mode: shell-script; indent-tabs-mode: t; sh-basic-offset: 8; sh-indentation: 8; tab-width: 8 -*-
```
annotation which supports the tab-based indentation.

riya-17: Thank you. I will take a look :)

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1259,9 +1267,16 @@ def execute(self):
...
+            status = self.check_failed_tool_init(tms)
+            # Tell the entity that started us who we are, indicating if we
+            # are ready or not based on the status collected from TMs and
+            # proceed further accordingly.
+            if not status:
+                started_msg = dict(
+                    kind="ds", action="startup", tool_meister_info=tms, status="failed"
+                )
+            else:
+                started_msg = dict(kind="ds", action="startup", status="success")
```
dbutenhof, line 1279: This is probably OK, since the redundancy is fairly simple; but it might be a bit easier to read if you factored out the redundancy. For example,

```
    started_msg = dict(kind="ds", action="startup", status="success" if status else "failed")
    if not status:
        started_msg["tool_meister_info"] = tms
```

This is also kinda interesting in that we have tool meister code looking specifically for "success", as an inline literal, but no convention for the opposite. I searched, because I'd expect the opposite of "success" to be "failure" rather than "failed"... however it seems that there's only "success" and "something other than success". Still, I'd prefer "failure" here.

I wish we had a standard constant definition (possibly an enum, or if there's really only to be "success" and "not success", a simple boolean) to define status values that would be less subject to typos ... but arguably that's beyond the scope of your PR and I won't say what I just said. 😄

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -447,6 +447,21 @@ class ToolDataSink(BaseServer):
...
                 status = data["status"]
+                if status != "success":
+                    tool_meister_info = data["tool_meister_info"]
+                    for host in tool_meister_info:
+                        if len(tool_meister_info[host]["failed_tools"]) == 0:
+                            continue
+                        hostname = tool_meister_info[host]["hostname"]
+                        tool = tool_meister_info[host]["installs"]
+                        for t in tool:
+                            tool_info = tool[t][1].split(": ")[1]
```
dbutenhof, line 458: Almost any time you iterate through the keys of a dict and then reference the *value*, you're better off with `.items()` to avoid the extra lookup; and it can make the code easier to read, as well, by eliminating the indexing (`host_info["failed_tools"]` instead of `tool_meister_info[host]["failed_tools"]`)

Finally, I think expressions like `tool[t][1].split(": ")[1]` deserve a bit of explanation about what's expected; because finding the code that constructs this payload isn't trivial and it's impossible to really analyze this without knowing what's there.

dbutenhof (commented):
This looks like you're moving forward nicely, but I think the code could use a bit of restructuring (and some more docstring and comment explanations) to clarify the code -- which is often not very obvious in isolation.

Also, I know this is mostly (or exclusively) tested through the legacy "unit" tests, which are not straightforward to mock... but I wonder if anyone can think up clever ways to exercise the bad tool init paths you're creating. If not ... we're probably overdue in building an actual tool meister infrastructure unit test suite, though I hesitate to require that for this PR. ☹️

agent/tool-scripts/base-tool
```diff
@@ -580,18 +580,18 @@ install)
...
 		local _tpv="${2}"
 		local _installed_rpm
-		local rc
-    _installed_rpm="$(require-rpm "${_tpn}" "${_tpv}")"
-    rc=${?} 
-		if [[ ${rc} != 0 ]]; then
+		local _rc
+        _installed_rpm="$(require-rpm "${_tpn}" "${_tpv}")"
+        _rc=${?} 
+		if [[ ${_rc} != 0 ]]; then
```
webbnh, line 586: Lines 583 and 586 are indented two levels, each indicated by a hard-tab character.  Lines 584 and 585 appear now to be indented two levels, but each level is signified by four blank-space characters.  If you cannot get your editor to use tabs for indentation in this module, you'll need to insert the characters "manually".  (You can cut-and-paste them, if you can't come up with a better way.)

agent/util-scripts/pbench-tool-meister-start
webbnh: > you're better off with `.items()`

I second this suggestion.  And, if you embrace it, then you don't need the `tool_meister_info` variable, either.

However, as I look at this, we don't actually need the keys!  So, something like this might suffice:
```
                    for entry in data["tool_meister_info"].values():
                        if len(entry["failed_tools"]) == 0:
                            continue
                        hostname = entry["hostname"]
                        tool = entry["installs"]
                        ...
```

> I think expressions like `tool[t][1].split(": ")[1]` deserve a bit of explanation

I agree.  And, I think that the `[1]` deserves to be a named reference (i.e., `tool[t]` should be a `namedtuple` or a `dict`) instead, but that's presumably beyond the scope of this PR.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1243,12 +1243,12 @@ def _valid_data(self, data):
...
-    def check_tms(tms):
-
+    def check_failed_tool_init(tms: dict) -> bool:
+        """checks if there are any tool failures in the Tool-Meisters"""
         for host, tm in tms.items():
             if len(tm["failed_tools"]) > 0:
-                return "failure"
-        return "success"
+                return False
+        return True
```
webbnh, line 1251: Since you never reference `host`, you might as well omit it:
```
        for _, tm in tms.items():
```
or, better still
```
        for tm in tms.values():
```

lib/pbench/agent/tool_data_sink.py
webbnh: I concur with Dave.

Also, I would recommend using a dictionary comprehension instead of the constructor:
```
            started_msg = {
                kind: "ds",
                action: "startup",
                status: "success" if status else "failure",
            }
            if not status:
                started_msg["tool_meister_info"] = tms
```
as I think that that is easier to comprehend.  😉

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1242,6 +1242,14 @@ def _valid_data(self, data):
             else:
                 return (action, directory, args)
 
+    @staticmethod
+    def check_failed_tool_init(tms: dict) -> bool:
```
webbnh, line 1246: In terms of naming this function, using "check" describes what the function _does_, but it doesn't provide much semantic value where it is _used_.  That is, this function is going to return a value, and "check" doesn't really convey what that value is going to mean.

I would recommend something more like `is_failed_tool_init()`.  That results in code which could look like
```
            # Tell the entity that started us who we are, indicating if we
            # are ready or not based on the status collected from TMs.
            started_msg = {kind:"ds", action:"startup"}
            if self.is_failed_tool_init(tms):
                started_msg["status"]="failed"
                started_msg["tool_meister_info"]=tms
            else:
                started_msg["status"]="success"
```
which is a little more expressive than what "check" provides.

webbnh (approved):
This looks good to me.  There's one indentation nit to fix, and I have a few code quality suggestions.

npalaska (dismissed):
This looks good but I do second some code refactoring changes suggested in the comments

agent/util-scripts/pbench-tool-meister-start
ndokos: One additional suggestion is to hide these manipulations inside a reporting function, rather than have them in-line. IMO, that makes it much easier to read the code: the function name says what it is doing, but its innards can be studied later - otherwise, the reader comes to a screeching halt, going from high level considerations to extremely low-level ones.

ndokos (commented):
LGTM in general. Just waiting until the comments are addressed.

agent/tool-scripts/base-tool
riya-17: i will try with a different editor than `vim`

lib/pbench/agent/tool_data_sink.py
portante, line 1246: Seems like this method, `check_failed_tool_init` should be shared with what looks like the same code in `pbench-tool-meister-start`.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1259,9 +1267,16 @@ def execute(self):
...
-            # Tell the entity that started us who we are, indicating we're
-            # ready, and all the TMs are ready.
-            started_msg = dict(kind="ds", action="startup", status="success")
+            status = self.check_failed_tool_init(tms)
+            # Tell the entity that started us who we are, indicating if we
+            # are ready or not based on the status collected from TMs and
+            # proceed further accordingly.
+            if not status:
+                started_msg = dict(
+                    kind="ds", action="startup", tool_meister_info=tms, status="failed"
```
portante, line 1276: I think this data is already stored as a key in Redis.

portante (changes requested):
We need a few tweaks to this:

 1. Let's leverage the key where the results for tools are already stored in Redis
 2. Let's be sure we share code between the Tool Data Sink and Tool Meister start code

lib/pbench/agent/tool_data_sink.py
riya-17: I will check that :)

riya-17: I couldn't find a similar looking code in `pbench-tool-meister-start`. Are you referring to `record_tms` function in `tool_data_sink.py`?

lib/pbench/agent/tool_data_sink.py
portante: When I first implemented the Tool Meister I was storing additional keys for each registered Tool Meister in Redis.  But that was removed mid-way through the development, if I recall correctly.

Having the full Tool Meister dictionary passed back through the response message works, but it might be better to have `record_tms()` also store the JSON blob in Redis as a key / value pair JSON dictionary entry so that it is also available for debugging.

lib/pbench/agent/tool_data_sink.py
portante: See https://github.com/distributed-system-analysis/pbench/pull/2659/files#diff-28fc2747e424d2014f9e3b3dd3e834315fa6db5a6079f62185988250ed197f69R453.

agent/util-scripts/test-bin/rpm
```diff
@@ -19,8 +19,13 @@ fi
 
 _rpm="${3}"
 if [[ "${_rpm}" == "pbench-sysstat" ]]; then
-    echo "12.0.3-1"
-    _rc=0
+    if [[ $PBENCH_UNIT_TEST_NAME == "test-61" ]]; then
+	echo "pbench-sysstat-12.0.3-1 is not installed"
```
webbnh, line 23: I realize that this PR is still marked "draft"...but, I'm not seeing this "not installed" message in the output from `test-61`...so I'm not sure that things are "working", here.

webbnh (commented):
It looks like the indentation in the bash script is still posing a challenge.

agent/tool-scripts/base-tool
```diff
@@ -581,8 +581,8 @@ install)
 		local _tpv="${2}"
 		local _installed_rpm
 		local _rc
-        _installed_rpm="$(require-rpm "${_tpn}" "${_tpv}")"
-        _rc=${?} 
+	        _installed_rpm="$(require-rpm "${_tpn}" "${_tpv}")"
+        	_rc=${?}
```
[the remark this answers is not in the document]
riya-17: I'm sorry about this. I don't know why my test-editors are unable to record tabs. I will make sure next time :(

agent/util-scripts/test-bin/rpm
riya-17: I must have forgotten to add the difference. Sorry about that.

agent/tool-scripts/base-tool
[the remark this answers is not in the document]
riya-17: I hope it got corrected this time :)

riya-17:
The PR needs some more updates and tweaks

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -451,6 +456,9 @@ class ToolDataSink(BaseServer):
                 kind = data["kind"]
                 action = data["action"]
                 status = data["status"]
+                if status != "success":
+                    tms = json.loads(redis_client.get(tms_tool_info))
+                    log_failed_tool_init(logger, tms)
```
dbutenhof, line 461: Given that we're succeeding or failing based on `status == "success"` at the end of the loop, and that we went to the trouble of declaring `status = ""` before the loop, I think we shouldn't be overwriting `status` from each iteration. Instead, maybe just start with `all_ok = True` and on each iteration, `if status != "success` set `all_ok` to False. Then at the end return 0 if `all_ok` else 1. (Or even `final_status = 0` and set it to 1 if any TM reported failure, in which case you can just return `final_status` at the end.)

Right now, `wait` (and the startup orchestration) is failing only if the *last* remote reported failure.

(This isn't new with your change, but this is probably a good time to fix it since you're extending the definition of "failed".)

agent/util-scripts/test-bin/rpm
```diff
@@ -19,8 +19,13 @@ fi
 
 _rpm="${3}"
 if [[ "${_rpm}" == "pbench-sysstat" ]]; then
-    echo "12.0.3-1"
-    _rc=0
+    if [[ ${_PBENCH_UNIT_TEST_NAME} == "test-61" ]]; then
```
dbutenhof, line 22: While I'm not thrilled about this, maybe it's not worth working out a better hack.

dbutenhof (changes requested):
You drew my attention to one existing bug in the code that I think becomes more important to fix with your changes. That is, we want `wait` to fail if *any* TM has a tool initialization failure, not only if the *last* TM has a failure.

agent/tool-scripts/base-tool
portante: FWIW, I pulled this change out into its own PR keeping authorship as @riya-17, see https://github.com/distributed-system-analysis/pbench/pull/2723.

agent/util-scripts/unittests
```diff
@@ -118,13 +118,15 @@ function _run {
         exit 1
     fi
 
+    export _PBENCH_UNIT_TEST_NAME=${tname}
     local opts=${options[${tname}]}
     echo "${tname} ${tscrpt} ${opts}"
     echo "+++ Running ${tname} ${tscrpt} ${opts}" >> ${_testout}
 
     _runout=${_testroot}/run.out
     eval ${tscrpt} ${opts} > ${_runout} 2>&1
```
portante, line 127: Rather than adding both an `export` a line 121, and an `unset` at line 129, can we consider using an inline environment variable, e.g. `_PBENCH_UNIT_TEST_NAME=${tname} eval ${tscrpt} ${opts} > ${_runout} 2>&1`?

lib/pbench/agent/utils.py
```diff
@@ -210,6 +210,41 @@ class RedisServerCommon(BaseServer):
     local_host = None
 
 
+def check_failed_tool_init(tms: dict) -> str:
```
portante, line 213: Can we consider adding python unit tests for this new method?

lib/pbench/agent/utils.py
```diff
@@ -210,6 +210,41 @@ class RedisServerCommon(BaseServer):
...
+
+    returns "success" when there are no failed_tools else "failure".
+    """
+    for _, tm in tms.items():
+        if len(tm["failed_tools"]) > 0:
+            return "failed"
+    return "success"
+
+
+def log_failed_tool_init(logger: logging.Logger, tms: dict):
```
portante, line 227: Can we consider adding python unit tests for this new method?

lib/pbench/agent/constants.py
```diff
@@ -21,6 +21,8 @@
 tm_channel_suffix_from_tms = "from-tms"
 # Channel suffix for the Tool Meister logging channel
 tm_channel_suffix_to_logging = "to-logging"
+# Tool-Meisters info key
+tms_tool_info = "tms_tool_info"
```
portante, line 25: Can we consider some name changes here?

Perhaps one of:

 * `tm_data_key = "tm-data-key"`
 * `tm_data_key = "tool-meister-data-key"`
 * `tool_meister_data_key = "tool-meister-data-key"`

agent/util-scripts/gold/test-client-tool-meister/test-61.txt
```diff
@@ -0,0 +1,291 @@
++++ Running test-61 test-client-tool-meister lite with-remotes
+"mpstat" tool is now registered for host "testhost.example.com" in group "lite"
+"dcgm" tool is now registered for host "testhost.example.com" in group "lite"
+"mpstat" tool is now registered for host "remote-a.example.com" in group "lite"
+"mpstat" tool is now registered for host "remote-b.example.com", with label "blue", in group "lite"
+"node-exporter" tool is now registered for host "remote-b.example.com", with label "blue", in group "lite"
+"dcgm" tool is now registered for host "remote-c.example.com", with label "red", in group "lite"
+"pcp" tool is now registered for host "remote-c.example.com", with label "red", in group "lite"
+/var/tmp/pbench-test-utils/opt/pbench-agent/util-scripts/pbench-tool-meister-start:110: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
```
portante, line 9: This error message is coming from using Python 3.11.  It does not show up when using Python 3.9.

We need to find a suitable replacement for `find_executable`.

Filed issue https://github.com/distributed-system-analysis/pbench/issues/2724 to track.

agent/util-scripts/gold/test-client-tool-meister/test-61.txt
```diff
@@ -0,0 +1,291 @@
...
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-b.example.com/node-exporter
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-c.example.com
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-c.example.com/__label__
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-c.example.com/dcgm
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite/remote-c.example.com/pcp
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite/testhost.example.com
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite/testhost.example.com/dcgm
+/var/tmp/pbench-test-utils/pbench/tools-v1-lite/testhost.example.com/mpstat
+=== /var/tmp/pbench-test-utils/pbench/tmp/tm-lite-remote-a.example.com.err:
+ERROR pbench-tool-meister fetch_message -- lost connection to redis server on channel pbench-agent-cli-to-tms
```
portante, line 54: This is not supposed to happen here, we should have a clean shutdown, but I don't think this code is responsible for this behavior, though it is hard to tell.

portante (changes requested):
Looking good.  We just need to figure out a few failure modes.

agent/util-scripts/pbench-tool-meister-start
riya-17: Sure :)

agent/util-scripts/gold/test-client-tool-meister/test-61.txt
riya-17: Maybe we can look at my test environment as my agent tests are not passing completely.

riya-17:
@portante @dbutenhof I am unable to understand the reason behind the Failure in tests. Can you please help me out?

dbutenhof:
> @portante @dbutenhof I am unable to understand the reason behind the Failure in tests. Can you please help me out?

Fascinating. Something's raising `SIGTERM` in the middle. Very likely, we're doing this in `test-start-stop-tool-meister`, which uses the `timeout` command to send a SIGTERM after 10 seconds:

```
function _timeout {
    timeout --kill-after=10 --signal TERM 20 $*
}
```

The problem is going to be figuring out exactly which tool meister operation hung for more than 10 seconds, because this certainly isn't normal. (It's likely something related to your changes, but not *directly* caused by them.)

Unfortunately, all the details logs will be under the `/var/tmp` unit test directory; I don't remember now exactly where we point that in the Jenkins runs. In any case, Jenkins "helpfully" packages up the test workspace, but that doesn't include anything under `/var/tmp` and therefore doesn't include the detailed logs.

@portante For triage, I wonder if we should direct Jenkins unit tests to a `JENKINS_TMP` directory we create **inside** the workspace, so the artifacts are all packaged up inside the [workspace zip](http://jenkins.intlab.perf-infra.lab.eng.rdu2.redhat.com/job/pbench-ci-multi/job/PR-2659/24/execution/node/3/ws/)?

Pragmatically, perhaps the immediate question is whether you can reproduce this with a simple `tox agent util-scripts` on your system, where your `/var/tmp/${USER}/` will be available?

lib/pbench/agent/utils.py
```diff
@@ -210,6 +210,44 @@ class RedisServerCommon(BaseServer):
...
+    """logs the tool failures in the Tool-Meisters
+
+    Args:
+        tms: Tool Meister data and metadata extracted from the redis.
+    """
+    failure_count = 0
+    for entry in tms.values():
+        if len(entry["failed_tools"]) == 0:
+            continue
```
webbnh, line 236: Per [PEP-8](https://peps.python.org/pep-0008/#programming-recommendations), this would be better expressed as
```
        if not entry["failed_tools"]:
            continue
```

portante (changes requested):
Reviewing the test failures from the CI job, and looking at how the tests run locally using `tox -e py -- agent`, I believe we need to take a slightly different approach to the sequence.

First, I think the Tool Data Sink should still return an unconditional "success" at [line 1265](https://github.com/distributed-system-analysis/pbench/pull/2659/files?show-viewed-files=true&file-filters%5B%5D=#diff-f186c241cb95e4c5972607dfea8d6669f4cb48c20a0def4276f58ad08808214dR1265), and not invoke `check_failed_tool_init(tms) earler at line 1261.

Next, `pbench-tool-meister-start` should unconditionally load the "tms" data from Redis, and check for failed tools, but ... instead of invoking the "cleanup" handlers, it should do what `pbench-tool-meister-stop` does :

 1. Send unconditionally send a "terminate" message (like at [line 246](https://github.com/distributed-system-analysis/pbench/blob/main/agent/util-scripts/pbench-tool-meister-stop#L246))
 2. Invoke graceful_shutdown() if locally managed (like at [line 265](https://github.com/distributed-system-analysis/pbench/blob/main/agent/util-scripts/pbench-tool-meister-stop#L265))

lib/pbench/test/unit/agent/test_utils.py
```diff
@@ -253,3 +260,286 @@ def test_methods(monkeypatch):
...
+
+    def test_log_failed_tool_init_single_host(self):
+        logger = logging.getLogger("test_log_methods")
+        failure_count = log_failed_tool_init(logger, self.mock_tms_data_single_host)
+        assert failure_count == 1
+
+    def test_log_success_tool_init(self):
+        logger = logging.getLogger("test_log_methods")
+        failure_count = log_failed_tool_init(logger, self.mock_tms_log_data)
+        assert failure_count == 0
```
webbnh, line 545: Arguably, we should test `log_failed_tool_init()` with the same ten scenarios that I proposed for `check_failed_tool_init()`.  However, for each remote where there is a tool failure, we really should test an analogous ten sub-scenarios (i.e., one, two, or three tools; one, two, or three failures among them; with the failures occurring in each possible position).

I think there is a nice, programmatic way to generate the cases, so we could use `@pytest.mark.parametrize()` with an argument which is a generator function instead of a constant list.  But, I think I'll leave coding that up to tomorrow. 😴

webbnh (changes requested):
There's a lot of good stuff, here.

I have a question around the "utility functions" -- they don't seem like they are "general use", and they seem like they are using specialized knowledge which belongs in the modules which call them, so it seems like they should be moved to those modules.

The unit tests for them look like a good start.  I would recommend making them more exhaustive.  (I.e., there are a bunch of "edge" cases which we should test, even though we know that the CUT can handle them now, because things might be different in the future.)  Also, the "mock data" for them doesn't need to be "complete" -- it only needs to include the fields which the CUT is going to reference.

Other than those, and the nits and suggestions, I do have a concern about a potential problem in `ToolDataSink.wait()`.

lib/pbench/test/unit/agent/test_utils.py
```diff
@@ -253,3 +260,286 @@ def test_methods(monkeypatch):
...
+                "dcgm": "--inst=/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts",
+                "mpstat": "--interval=42 --options=forty-two",
+            },
+            "noop_tools": [],
+            "persistent_tools": ["dcgm"],
+            "transient_tools": [],
+            "failed_tools": ["mpstat"],
+            "posted": None,
+        }
+    }
```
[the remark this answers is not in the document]
riya-17: Sure, I wasn't sure whether I should trim these down or not, So, I added all data.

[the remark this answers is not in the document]
webbnh: The way to think about unit testing is that we want the unit tests to exercise the code-under-test, driving it down every possible path (or, at least, specific paths) and ensuring that it always produces the appropriate behavior.  The mechanism required to do that doesn't always have match realistic operation.  For instance, we generally mock out all subroutine calls, both to enable us to execute the CUT in isolation as well as to allow us to control its operation and to check its behavior.  Similarly, in terms of the data supplied to the CUT, we don't need to provide realistic inputs (that's something that _functional_ testing will do); instead we only need to provide the specific inputs required to drive the intended execution of the CUT.

One problem with unit testing is that it can be difficult to understand (and even more difficult to "proofread").  So, keeping definitions small, defining nothing more than we need to, and keeping the definitions in logical, easy-to-find places (like at the top of each test scenario) helps a lot.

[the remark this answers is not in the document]
webbnh: The pared down data makes it much easier to see how the three test cases are different from each other and what they are testing -- thanks!

lib/pbench/test/unit/agent/test_utils.py
riya-17: If I move module `log_failed_tool_init` back to `pbench-tool-meister-start` then, I believe I won't be able to add pytests for this module.

lib/pbench/agent/utils.py
```diff
@@ -210,6 +210,44 @@ class RedisServerCommon(BaseServer):
...
 
+def check_failed_tool_init(tms: dict) -> str:
+    """checks if there are any tool failures in the Tool-Meisters
+
+    Args:
+        tms: Tool Meister data and metadata published by redis.
+
+    returns "success" when there are no failed_tools else "failure".
+    """
+    for _, tm in tms.items():
```
dbutenhof, line 221: 1. In `log_failed_tool_init` you iterate through `tms.values()` because you don't want the key; here, you use `.items()` but throw away the key with `_`; these should be consistent since they're similar in function (and especially since they're currently adjacent)
2. These two really do the same thing, with the exception that this one doesn't log the failures. It'd be better to use a common definition of "failed tool" that encapsulates the root logic (even though it's currently just `tm["failed_tools"]` (which would be equivalent to your `len(tm["failed_tools"]) > 0` and arguably more "Pythonic").)
3. It might even make sense to combine them with a `log_failures: bool = False` parameter; returning `True` or `False` would make more sense at this level... and `check_failed_tool_init` is a better name than `log_failed_tool_init` given that the primary return value (not even mentioned, incidentally, in the `log` variant) is a value indicating whether there were failures. (Although I think "check" is even a bit ambiguous here; something specific like `if tool_init_failed()` makes a much more readable statement.)

lib/pbench/agent/utils.py
dbutenhof: Neither the function signature nor the docstring mentions the critical return value! As I've said elsewhere, I don't think this is a good name in the first place, and I'd prefer to see this and the non-reporting variant combined. (Which would also address some of Webb's unit test comments as there'd be no need for a duplicate set of tests for the two alternatives.)

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -502,6 +507,9 @@ class ToolDataSink(BaseServer):
                 kind = data["kind"]
                 action = data["action"]
                 status = data["status"]
+                if status != "success":
+                    tms = json.loads(redis_client.get(tm_data_key))
+                    failure_count = log_failed_tool_init(logger, tms)
```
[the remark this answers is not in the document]
dbutenhof: Then again ... thinking more about Webb's comment on `kind` and `action` ... the "failed tool init" really only applies to the TDS confirmation of the init message, whereas this loop is processing a sequence of TDS messages, potentially including others?

I'm not really sure what we should be doing here. Are there any other messages aside from ds/startup we really expect to handle here? What should we do about them? I don't think we should be "logging tool failure" for any other sort of message, if there are any. Do we log them and discard them? Should we be prepared to take some additional action?

At the very least we should be checking what we're getting.

I really think the original expectation here was that we'd get a *single* message confirming startup through TDS, and maybe the initial `status = ""` was simply intended to handle the case where the generator doesn't yield any results. (And there's no exit from this loop; it exits when the Redis pubsub listener quits or throws.

In fact, I realize that I initially completely misunderstood the context and purpose of this loop, and now I'm not sure I understand how it works at all. What makes the pubsub channel "stop" here so that we terminate the generator and this loop? Clearly that happens somehow, and the structure now isn't significantly different than before.

So, yeah, the old `status` returned would be the status for the final *message* received through the pubsub channel before the generator stopped; and **if** that was the startup confirmation (which is clearly implicitly assumed by the loop), exiting with that status is fine: any previous messages are presumably irrelevant. (What would they be? Does this actually happen?)

dbutenhof (changes requested):
I'm curious whether rebasing on my startup fix will resolve your Jenkins problems ... although this commit didn't ever even finish, apparently.

I've also revised my opinion regarding the `wait` loop as I think about the context a bit more. I think this needs more checking and some additional understanding of exactly how the Redis pubsub channel and our intervening generators are going to behave here.

I'd also like to see some renaming and refactoring of the new utility methods; because they're really almost the same and neither name or signature really captures the critical semantics.

lib/pbench/test/unit/agent/test_utils.py
dbutenhof: Yes, that's true, and it's a valid point. (I'm going to have some nasty rebases, I expect, but I'm working on refactoring `pbench-tool-meister-{start|stop}` into click-like trampoline files with all the logic in testable modules; and this will likely go into a shared based class rather than this generic utilities module.)

I think that if you take my suggestions about merging these twin functions, this will be less of an issue since the utility is used by both TDS and start. 😃

lib/pbench/test/unit/agent/test_utils.py
```diff
@@ -253,3 +260,286 @@ def test_methods(monkeypatch):
...
+        },
+    }
+
+    def test_check_failed_tool_init(self):
+        status = check_failed_tool_init(self.mock_tms_failed_data)
+        assert status == "failed"
+
+    def test_check_success_tool_init(self):
+        status = check_failed_tool_init(self.mock_tms_success_data)
+        assert status == "success"
```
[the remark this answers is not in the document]
portante: Why do we need to be that exhaustive?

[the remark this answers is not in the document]
portante: I am just wondering what code paths are in our control that this is exercising?  We just need to have one case that has an entry with failed tools and one without in the same loop, no?

lib/pbench/test/unit/agent/test_utils.py
portante: It needs to be removed from TDS entirely.

There is no need for TDS to use either of these functions, they are entirely for the start mechanism to use because they are about messaging the user.

And the plain logging is not what we want, we want to use `error_log`.  The logging methods only dump to the pbench.log file, I believe.

portante: I'll post a PR with my thinking.

dbutenhof: I don't think that's quite true; the "check" variety is used by TDS in processing the TM confirmations to determine whether the message back to the CLI is "success" or "failure".

Now I suppose this could be simplified, because the individual TM clients should be reporting "failure" if there are any problems, which really means TDS only needs to report failure if at least one of the clients did; it arguably shouldn't have any need to re-analyze the data.

portante: I know that is this code is proposing to do, but it is not necessary.  I'll be posting a PR shortly.

portante (changes requested):
Please rebase on `main` before making any other changes.

Second, please consider merging https://github.com/riya-17/pbench/pull/5, this represents a working state of the behavior.

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
riya-17: > Then again ... thinking more about Webb's comment on `kind` and `action` ... the "failed tool init" really only applies to the TDS confirmation of the init message, whereas this loop is processing a sequence of TDS messages, potentially including others?
> 
> I'm not really sure what we should be doing here. Are there any other messages aside from ds/startup we really expect to handle here? What should we do about them? I don't think we should be "logging tool failure" for any other sort of message, if there are any. Do we log them and discard them? Should we be prepared to take some additional action?
> 
> At the very least we should be checking what we're getting.
> 
> I really think the original expectation here was that we'd get a _single_ message confirming startup through TDS, and maybe the initial `status = ""` was simply intended to handle the case where the generator doesn't yield any results. (And there's no exit from this loop; it exits when the Redis pubsub listener quits or throws.
> 
> In fact, I realize that I initially completely misunderstood the context and purpose of this loop, and now I'm not sure I understand how it works at all. What makes the pubsub channel "stop" here so that we terminate the generator and this loop? Clearly that happens somehow, and the structure now isn't significantly different than before.
> 
> So, yeah, the old `status` returned would be the status for the final _message_ received through the pubsub channel before the generator stopped; and **if** that was the startup confirmation (which is clearly implicitly assumed by the loop), exiting with that status is fine: any previous messages are presumably irrelevant. (What would they be? Does this actually happen?)

I did a benchmark run, where all the requirements were met and it was successful without emitting anything other than these messages and also on failures. So, I am also not sure if there might be some other conditions that might provoke it to emit a message which I might not have seen.

lib/pbench/test/unit/agent/test_utils.py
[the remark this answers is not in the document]
webbnh: This is an interesting philosophical question.  I would propose that unit testing should do two things:

1. define and enforce the "interface contract" (aka "test as specification")
2. ensure coverage

The former suggests a sort of arms-length/black-box approach.  We know what the CUT should do given a particular set of inputs, and the tests should provide those inputs and check the outputs.

The latter requires a certain amount of "privileged" or "internal" knowledge of the CUT.  And, while I do think that it is important that we have excellent test coverage of our code, too much white-box testing is likely to interfere with our ability to change the CUT.  So, we need to take care when implementing unit tests that we are not relying on internals knowledge any more than we have to in order to drive coverage.

With that said, in my experience debugging software, when we have a "list of things", problems with the list generally occur at the beginning, middle, or end of the list; thus, we should have test cases where we have only one item (i.e., it's all three), where we have two items (where there is no middle), and where we have three items (where the three are distinct).  Having more than three is unlikely to produce value relative to three, but omitting any of those first three configurations is likely to leave something unchecked.  Moreover, if one of the entries is special (e.g., we have three results, each of which might have two values), it's important to check for the special entries in each of the three positions and in each of the various combinations.

Now, if we look inside the CUT, we can see that this level of testing isn't "needed".  But, the CUT might change in the future, and we would prefer that the existing unit tests be able to validate future revisions of the CUT.  So, to the extent that exhaustive testing is easy to implement and not too expensive to execute, I think it is worthwhile.

The short of it is that we could have a failure in the first item, the last item, or in any item in between, and we could have more than one failure, and so we want to make sure that we get the result that we expect for each case, and there are (I think) 11 of them...which means that I missed 4 of them 😉, but they are all in scenarios with shorter lists, so they are probably OK  😁.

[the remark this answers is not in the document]
portante: Also, I have proposed removing this entirely.  Why do we need to pursue this?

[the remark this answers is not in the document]
webbnh: [Pronouns can be so challenging....]

My post here was answering your question; it was not a comment on whether or how to go forward with this part of the change which this "conversation" references.

On that topic, I looked at your PR to Riya's branch, and it looks like it includes this same functionality (albeit in a different form)...and so there really _should_ be unit testing for this code (however/wherever it lands), and so we will need to decide how exhaustive that testing should be....

lib/pbench/agent/tool_data_sink.py
```diff
@@ -21,12 +21,18 @@
...
+from daemon import DaemonContext
+from jinja2 import Environment, FileSystemLoader
+from pathlib import Path
 from threading import Thread, Lock, Condition
 from typing import Any, Dict, List, NamedTuple, Tuple
 
 from bottle import Bottle, ServerAdapter, request, abort
 from configparser import DuplicateSectionError
 from daemon import DaemonContext
+from distutils.spawn import find_executable
```
webbnh, line 35: LGTM points out that the code is not using this `import`...which is a good thing, since we're switching away from this function....

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1254,9 +1263,11 @@ def execute(self):
             self._tm_tracking = self.record_tms(tms)
             self._num_tms = len(self._tm_tracking.keys())
 
-            # Tell the entity that started us who we are, indicating we're
-            # ready, and all the TMs are ready.
-            started_msg = dict(kind="ds", action="startup", status="success")
+            status = check_failed_tool_init(tms)
```
portante, line 1266: As per https://github.com/riya-17/pbench/pull/5, we do not want to have the Tool Data Sink applying a policy based on failed tools.  That policy should be applied by `pbench-tool-meister-start` where it can direct what the Tool Data Sink should do as a result.

lib/pbench/agent/tool_data_sink.py
portante: The merge process must have missed this change.

lib/pbench/agent/utils.py
portante: Based on https://github.com/riya-17/pbench/pull/5 we no longer need these methods add to the `utils` module in order for them to be shared between the Tool Data Sink code (which should not be applying policy) and `pbench-tool-meister-start`.

There is a refactoring PR for `pbench-tool-meister-start` in PR #2741 where the suggested behavior from https://github.com/riya-17/pbench/pull/5 would also be moved there, but I'd rather see this PR land with as few refactorings as possible.

Let's move this kind of improvement to another PR that layers on top of PR #2741.

lib/pbench/test/unit/agent/test_utils.py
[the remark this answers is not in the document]
portante: I agree there should be, but I am really pushing hard to get the v0.71 release out the door without more refactorings.

Let's defer this kind of testing to post v0.71 in a PR that layers on top of PR #2741.

portante (changes requested):
Unless there is a bug introduced by https://github.com/riya-17/pbench/pull/5, let's get that merged and drive this to completion so we can work at releasing v0.71.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -21,12 +21,18 @@
 import subprocess
 import sys
 import tempfile
+
+from configparser import DuplicateSectionError
+from daemon import DaemonContext
+from jinja2 import Environment, FileSystemLoader
+from pathlib import Path
```
portante, line 28: This looks like a bad merge as these imports appear to be repeated below.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1216,6 +1223,8 @@ def record_tms(self, tms):
         with (mdlog_name).open("w") as fp:
             mdlog.write(fp)
 
+        self.redis_server.set(tm_data_key, json.dumps(tms, sort_keys=True))
```
portante, line 1226: FYI, I've proposed this change as a separate PR in https://github.com/distributed-system-analysis/pbench/pull/2742.

lib/pbench/agent/tool_data_sink.py
riya-17: Ohh,, I must have missed that

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -549,6 +554,40 @@ class ToolDataSink(BaseServer):
         if self.wait_for_pid(self.get_pid()):
             self.kill(status)
 
+    @staticmethod
```
portante, line 557: This looks like the results of a bad merge.

lib/pbench/agent/tool_data_sink.py
portante: PR #2742 was merged, so this code can be dropped entirely.

portante (changes requested):

portante (changes requested):
Looking good, just need to drop the extra `.set(tm_data_key, ...)` in `tool_data_sink.py` as noted.

ndokos (approved):

portante (approved):

npalaska (approved):
LGTM :+1:

webbnh (approved):
👍

Issue #2085, opened 2021-01-31
Tool Meister working directory on remote hosts not being cleaned up


portante:
When the Tool Meister is run on remote hosts, its working directory is created in the `pbench_tmp` directory as specified by the `pbench-agent.cfg` file.

When the Tool Meister exits, it should remove its working directory, but that does not appear to be happening.

We should also consider `tar`ing up that directory and sending its contents back to the controller to be stored with the results before it is cleaned up and removed.

npalaska:
I have few observations and questions on this issue:

Case 1 (When we have only transient tools):
1. `init_tools` of the tool meister sets up the temporary directory for the persistent tools to store its data.
2. However, the transient tools dont use this temporary directory created by the `init_tools` instead they use another temporary directory that is created by the `start_tools` which actually starts the transient tools.
3. If we don't have any persistent tools registered then tmp directory created by the `init_tools` would be empty.
4. The tmp directory created by the tool meister is only removed when the `send_tools` is called which internally calls `_send_directory`.
4. Now if we only have transient tools registered the above empty tmp directory (created by `inti_tools`) doesn't get cleaned up after the `send_tools`. (Because `send_tools` is invoked after the `stop_tools` which only gets invoked for transient tools, for the persistent tools there is no start and stop, instead we have init and end)
5. Therefore if we only have transient tools then at any given time on remote there are 2 tmp directories gets created by the tool meister (one by the init_tools and one by the start_tools). The one created by the `start_tools` (the correct tmp directory for the transient tools) gets cleaned up after the `stop tools` is called and we are left with an one empty directory (created by `init_tools`)

Mitigation: Only create temporary directory in init_tools when we have at least one persistent tool registered otherwise don't create it in the `init_tools` let `start_tools` handle it.

Case 2 ( When we have both transient and persistent tools):
1. If we have both persistent tools and transient tools then we need to create a tmp directory in the `init_tools` (Because that's when we start the persistent tools data collection)
2. This is the tmp directory that persistent tool uses to store std_err and std_out files of its execution on remote host such as `tm-node-exporter-start.err` and `tm-node-exporter-start.out`.
3. However, when the `end_tools` is called we are not cleaning up the directory that was created by the `init_tools`.
4. For the persistent tools we dont send any _tool data_ per se like transient tools but we send scraped data every certain interval over the `tool-data-sink` to the controller, where `tool-data-sink` store all persistent tool data as a tar file.
[...]

PR #2661, opened 2022-03-03
Remove persistent tmp directory on remote hosts

npalaska:
Fixes issue #2085 
- Deletes persistent tool tmp directory on remote hosts when end_tools is called 
- Instead of writing persistent tool local logs on a remote file system, send them to TM via tool data sink on redis channel.

### DIFF
--- a/agent/util-scripts/gold/test-client-tool-meister/test-53.txt
+++ b/agent/util-scripts/gold/test-client-tool-meister/test-53.txt
@@ -63,11 +63,6 @@ Collecting system information
 /var/tmp/pbench-test-utils/pbench/mock-run/tm/tm.logs
 /var/tmp/pbench-test-utils/pbench/mock-run/tools-default
 /var/tmp/pbench-test-utils/pbench/mock-run/tools-default/prometheus_data.tar.xz
-/var/tmp/pbench-test-utils/pbench/mock-run/tools-default/testhost.example.com
-/var/tmp/pbench-test-utils/pbench/mock-run/tools-default/testhost.example.com/dcgm
-/var/tmp/pbench-test-utils/pbench/mock-run/tools-default/testhost.example.com/dcgm/dcgm-exporter.file
-/var/tmp/pbench-test-utils/pbench/mock-run/tools-default/testhost.example.com/dcgm/tm-dcgm-start.err
-/var/tmp/pbench-test-utils/pbench/mock-run/tools-default/testhost.example.com/dcgm/tm-dcgm-start.out
 /var/tmp/pbench-test-utils/pbench/pbench.log
 /var/tmp/pbench-test-utils/pbench/tmp
 /var/tmp/pbench-test-utils/pbench/tools-v1-default
@@ -180,6 +175,7 @@ port 17001
 +++ mock-run/tm/tm-default-testhost.example.com.err file contents
 INFO pbench-tool-meister sysinfo -- pbench-sysinfo-dump -- /var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/pbench-sysinfo-dump /var/tmp/pbench-test-utils/pbench/mock-run/sysinfo/beg/testhost.example.com block,security_mitigations,sos parallel
 INFO pbench-tool-meister sysinfo -- testhost.example.com: sysinfo send (no-op) default /var/tmp/pbench-test-utils/pbench/mock-run/sysinfo/beg/testhost.example.com
+INFO pbench-tool-meister start -- Starting persistent tool dcgm, args ['/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/dcgm-exporter']
 INFO pbench-tool-meister start -- Started persistent tool dcgm, ['/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/dcgm-exporter']
 INFO pbench-tool-meister start -- mpstat: start_tool -- /var/tmp/pbench-test-utils/opt/pbench-agent/tool-scripts/mpstat --start --dir=/var/tmp/pbench-test-utils/pbench/mock-run/0-iter-zero/sample42/tools-default/testhost.example.com --interval=42 --options=forty-two
 INFO pbench-tool-meister stop -- mpstat: stop_tool -- /var/tmp/pbench-test-utils/opt/pbench-agent/tool-scripts/mpstat --stop --dir=/var/tmp/pbench-test-utils/pbench/mock-run/0-iter-zero/sample42/tools-default/testhost.example.com --interval=42 --options=forty-two
@@ -193,6 +189,7 @@ INFO pbench-tool-meister wait -- Waiting for transient tool mpstat start process
 INFO pbench-tool-meister send_tools -- testhost.example.com: send_tools (no-op) default /var/tmp/pbench-test-utils/pbench/mock-run/1-iter-one/sample42/tools-default/testhost.example.com
 INFO pbench-tool-meister stop -- Terminate issued for persistent tool dcgm
 INFO pbench-tool-meister wait -- Stopped persistent tool dcgm
+INFO pbench-tool-meister end_tools -- testhost.example.com: deleting persistent tool tmp directory default /var/tmp/pbench-test-utils/pbench/mock-run/tools-default/testhost.example.com
 INFO pbench-tool-meister sysinfo -- pbench-sysinfo-dump -- /var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/pbench-sysinfo-dump /var/tmp/pbench-test-utils/pbench/mock-run/sysinfo/end/testhost.example.com block,security_mitigations,sos parallel
 INFO pbench-tool-meister sysinfo -- testhost.example.com: sysinfo send (no-op) default /var/tmp/pbench-test-utils/pbench/mock-run/sysinfo/end/testhost.example.com
 INFO pbench-tool-meister __exit__ -- testhost.example.com: terminating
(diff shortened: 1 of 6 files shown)


agent/util-scripts/gold/test-client-tool-meister/test-53.txt
```diff
@@ -180,6 +175,7 @@ port 17001
 +++ mock-run/tm/tm-default-testhost.example.com.err file contents
 INFO pbench-tool-meister sysinfo -- pbench-sysinfo-dump -- /var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/pbench-sysinfo-dump /var/tmp/pbench-test-utils/pbench/mock-run/sysinfo/beg/testhost.example.com block,security_mitigations,sos parallel
 INFO pbench-tool-meister sysinfo -- testhost.example.com: sysinfo send (no-op) default /var/tmp/pbench-test-utils/pbench/mock-run/sysinfo/beg/testhost.example.com
+INFO pbench-tool-meister start -- Starting persistent tool dcgm, args ['/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/dcgm-exporter']
 INFO pbench-tool-meister start -- Started persistent tool dcgm, ['/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/dcgm-exporter']
```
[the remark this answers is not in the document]
npalaska: Yeah this should really be a `debug` message, `INFO` doesn't add much value. Also this message is coming from `if-else` blocks where the `if` block of the code has an identical message with a `debug` log so I need to consistent with it

lib/pbench/agent/tool_meister.py
```diff
@@ -422,39 +423,49 @@ def __init__(self, name, tool_opts, logger=None, **kwargs):
...
+                "Starting persistent tool %s, args %r", self.name, self.args
+            )
+            self.process = subprocess.Popen(
+                self.args,
+                cwd=self.tool_dir,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
+            )
+        child_reader = threading.Thread(
+            target=self.log_subprocess_output, args=(self.process.stdout,)
```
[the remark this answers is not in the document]
npalaska: The `threading` documentations says that the `args` is the argument tuple for the target invocation. Since its a tupple i had to insert a comma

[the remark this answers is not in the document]
dbutenhof: Ah; indeed. I should have thought of that. Python is weird, and sometimes ugly. ☹️

lib/pbench/agent/tool_meister.py
```diff
@@ -1518,6 +1533,39 @@ def end_tools(self, data):
...
+            try:
+                shutil.rmtree(directory_to_delete)
+            except Exception:
+                self.logger.exception(
+                    f"Failed to remove tmp persistent tool data hierarchy, "
+                    f"{directory_to_delete}"
+                )
+                failures += 1
+        else:
+            self.logger.exception(
```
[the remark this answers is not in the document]
npalaska: I agree, `self.logger.warning` would be better.

lib/pbench/agent/tool_meister.py
```diff
@@ -1518,6 +1533,39 @@ def end_tools(self, data):
...
                 failures += 1
+
+        # Remove persistent tool temporary working directory on remote host
+        directory = data["directory"]
+        tool_dir = self.directories[directory]
+
+        if self._hostname == self._controller:
+            directory_to_delete = tool_dir
+        else:
+            directory_to_delete = tool_dir.parent
```
webbnh, line 1544: This seems like a bad idea:  doing it this way spreads specialize knowledge across multiple areas of the code when it could (and should) be encapsulated in one.  We should be recording the directory that we want deleted in `init_tools()` so that we _don't_ have to be making adjustments like this, here.

lib/pbench/agent/tool_meister.py
```diff
@@ -1518,6 +1533,39 @@ def end_tools(self, data):
...
+        self.logger.debug(
+            "%s: deleting persistent tool tmp directory %s %s",
+            self._hostname,
+            self._group,
+            directory_to_delete,
+        )
+        unexpected_files = [x for x in directory_to_delete.iterdir() if x.is_file()]
+        if not unexpected_files:
+            try:
+                shutil.rmtree(directory_to_delete)
```
webbnh, line 1555: It seems like we should be complaining if there are any _files_ at **any** level of the hierarchy.  That is, we can afford to ignore subdirectories if they are empty or if they contain no subdirectories which contain any files, but we need to do a recursive file tree walk ([e.g.](https://docs.python.org/3/library/os.html#os.walk)) to find that out.

And, I wonder if, after issuing a warning, we should to go ahead and delete the files anyway.  If the files actually were valuable, then I think requiring the user to make another run is probably a lesser evil than leaving behind a bunch of useless litter (in the cases where the files are _not_ valuable).

lib/pbench/agent/tool_meister.py
```diff
@@ -1518,6 +1533,39 @@ def end_tools(self, data):
...
+                failures += 1
+        else:
+            self.logger.exception(
+                f"Unexpected files found at tmp persistent tool directory "
+                f"hierarchy, {directory_to_delete}. Files: {unexpected_files}"
+            )
+            failures += 1
+        del self.directories[directory]
         if failures > 0:
             msg = f"{failures} of {tool_cnt} failed stopping persistent tools"
```
webbnh, line 1570: Are you sure that we want to count the presence of unexpected files or the inability to delete a temporary directory as a "failure"?  They don't result in us failing to obtain the data, so they seem like things that we should just warn the user about.

webbnh (approved):
These changes are generally OK, but I've noted some improvements that I think you should consider.

lib/pbench/agent/tool_meister.py
```diff
@@ -422,39 +423,49 @@ def __init__(self, name, tool_opts, logger=None, **kwargs):
         self.args = None
         self.process = None
 
+    def log_subprocess_output(self, pipe):
+        for line in iter(pipe.readline, b""):
```
webbnh, line 427: Just for my own edification, why is this idiom better than just
```
        for line in pipe.readline:
```
?

lib/pbench/agent/tool_meister.py
```diff
@@ -422,39 +423,49 @@ def __init__(self, name, tool_opts, logger=None, **kwargs):
...
+            self.logger.info(f"{self.__class__.__name__}: {line}")
```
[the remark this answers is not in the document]
webbnh: Also, `line` will likely contain a `\n` at the end, do we need to strip that, or does the logger do that for us?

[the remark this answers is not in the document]
webbnh: Regarding my first comment, we might want to do something like:
```
    def log_subprocess_output(self, pipe):
        logger = self.logger.getChild(self.__class__.__name__)
        for line in pipe.readlines():
           logger.info(line)
```
[doc](https://docs.python.org/3/library/logging.html#logging.Logger.getChild)

ndokos (commented):
LGTM. I;'ll wait until the comments are addressed.

lib/pbench/agent/tool_meister.py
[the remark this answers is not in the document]
npalaska: This is how the log message looks like without explicitly putting class name 
```
hostname.com 0016 2022-03-04 16:17:39,789 INFO 20555 139737573009152 pbench-tool-meister log_subprocess_output 428 -- b'time="2022-03-04T16:17:39Z" level=info msg=" - ipvs" source="node_exporter.go:97"\n'
```
and after putting class name
```
hostname.com 0016 2022-03-04 16:17:39,789 INFO 20555 139737573009152 pbench-tool-meister log_subprocess_output 428 -- NodeExporterTool: b'time="2022-03-04T16:17:39Z" level=info msg=" - ipvs" source="node_exporter.go:97"\n'
```
So I guess it is not adding the class name by default, but it is adding the def name and a very top level file name where it was called from (and that's the case with all of our logs at this point). Changing the logger configuration to add class name would be good but I think that should be captured in another commit. Also creating another logger just for this case would be an overkill isn't it?

lib/pbench/agent/tool_meister.py
```diff
@@ -1045,6 +1056,10 @@ def init_tools(self, data):
                 else:
                     self._persistent_tools[name] = persistent_tool
                     self.logger.debug("NAME: " + name + "  TOOL OPTS: " + tool_opts)
+
+        # Remember this persistent tmp tool directory so that we can delete it
+        # when requested.
+        self.directories[data["directory"]] = _tool_dir
```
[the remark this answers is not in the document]
npalaska: > And, why is it using `data["directory"]` as a key? Is that value meaningful when `self._controller` is not the same as `self._hostname`?

`data["directory"]` is a value where the persistent tool data is going to be stored on the controller. The key is always (in case of persistent tools) `/var/lib/pbench-agent/<benchmark-name>/tools-<group_name>` irrespective of whether hostname is same as controller. The value for this key is either a remote tmp directory in case there is a remote or same as key in case hostname is the controller. The key just points to where the tmp tools data is stored.

lib/pbench/agent/tool_meister.py
npalaska: > And, I wonder if, after issuing a warning, we should to go ahead and delete the files anyway. If the files actually were valuable, then I think requiring the user to make another run is probably a lesser evil than leaving behind a bunch of useless litter (in the cases where the files are _not_ valuable).

That's a fair point, these tmp directory is created by us and only we should have control over this directory and if we are expected to delete this directory we should delete all the files in it as well (even though we dont expect any files in there)

lib/pbench/agent/tool_meister.py
```diff
@@ -1518,6 +1533,39 @@ def end_tools(self, data):
...
+            except Exception:
+                self.logger.exception(
+                    f"Failed to remove tmp persistent tool data hierarchy, "
+                    f"{directory_to_delete}"
+                )
+                failures += 1
+        else:
+            self.logger.exception(
+                f"Unexpected files found at tmp persistent tool directory "
+                f"hierarchy, {directory_to_delete}. Files: {unexpected_files}"
```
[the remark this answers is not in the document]
npalaska: > `iterdir()` is only going to report regular files at the first level anyway.

Yep, I should have been careful, fixed the bug in the latest commit.

lib/pbench/agent/tool_meister.py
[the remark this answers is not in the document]
npalaska: Fixed it

lib/pbench/agent/tool_meister.py
npalaska: Yep I concur that we should be recording the directory that we want deleted in init_tools() itself.

lib/pbench/agent/tool_meister.py
npalaska: hmm I dont know, `iter` is just something that occurred to me first I guess. But yeah I think `readlines()` is a simpler implementation to read lines.

lib/pbench/agent/tool_meister.py
```diff
@@ -1534,37 +1540,27 @@ def end_tools(self, data):
...
-            try:
-                shutil.rmtree(directory_to_delete)
-            except Exception:
-                self.logger.exception(
-                    f"Failed to remove tmp persistent tool data hierarchy, "
-                    f"{directory_to_delete}"
+        for dirpath, _, files in os.walk(tool_dir):
+            if files:
+                self.logger.warning(
+                    f"Found regular files {files} at {dirpath} and will be deleted later"
```
dbutenhof, line 1556: "found [this] and will be deleted later" isn't grammatical. I also don't find the reference to "regular files" or the "will be deleted later" to be of much real value in what could turn out to be a very long message. (Also, "later" is really "immediately following this message", which means the distinction isn't very useful.)

Including the full directory path only helps in that it includes a hostname, which would otherwise be ambiguous; but I'd kinda prefer that the remote generating the message identify itself explicitly (and more compactly) instead. Listing the files, if we're implicitly deleting them, probably makes sense, but let's drop at least some of the boilerplate around the list, and put the list last.

Perhaps something like `"{hostname} unexpected temp files [{','.join(files)}]`???

Also, I think showing a separate `warning` message, from the same place, for each directory in the tree, is confusing. (Granted, we don't expect this to happen a lot.) It's only a little more work to aggregate a list that's dumped in a single message. (For bonus points, produce relative paths from the root temp directory with `Path.relative_to(*other)`.)

dbutenhof (commented):
Mostly good; I'm just quibbling about the wording of the dangling file message. 😁

lib/pbench/agent/tool_meister.py
npalaska: Yes I knew you would have something better to offer in terms of messaging here, so I didnt give a lot of thought to format this message :stuck_out_tongue_closed_eyes:

> Perhaps something like `"{hostname} unexpected temp files [{','.join(files)}]`???
> 
> Also, I think showing a separate `warning` message, from the same place, for each directory in the tree, is confusing. (Granted, we don't expect this to happen a lot.) It's only a little more work to aggregate a list that's dumped in a single message. (For bonus points, produce relative paths from the root temp directory with `Path.relative_to(*other)`.)

Yes I concur.

lib/pbench/agent/tool_meister.py
```diff
@@ -1518,6 +1539,41 @@ def end_tools(self, data):
...
+                            files,
+                        )
+                    )
+                }
+                unexpected_files += relative_file_path
+
+        if unexpected_files:
+            self.logger.warning(
+                f"{self._hostname}: unexpected temp files {unexpected_files}"
+            )
```
dbutenhof, line 1569: Much better, although I don't get the degenerate dict comprehension for `relative_file_path`. I also think that this list of joined lists representation is a bit weird. Since you're capturing the path of each relative to the common base directory, I would have just extended the `unexpected_files` array and done a single join at the end rather than have a list of comma-lists (assuming there's more than one directory with more than one file each, which doesn't seem to happen in the gold files I examined).

```
         unexpected_files: List[str] = []
￼        for dir, _, files in os.walk(tool_dir):
             dirpath = Path(dir).relative_to(tool_dir)
￼            if files:
￼                relative_file_path = map(
￼                    lambda x: f"{dirpath}/{x}",
￼                    files,
￼                )
￼                unexpected_files += relative_file_path
￼
￼        if unexpected_files:
￼            self.logger.warning(
￼                f"{self._hostname}: unexpected temp files {','.join(unexpected_files)}"
￼            )
```

npalaska: Ah! Thanks I found a bug in my approach, multiple `join` inside a for loop would result in forming a single string in a list, e.g. `["/var/1,/var/2,/var/3"]` instead of `["/var/1","/var/2","/var/3"]` if there are multiple files at the same directory, and we want the latter output.

dbutenhof (approved):

riya-17 (approved):
Looks Good

lib/pbench/agent/tool_meister.py
```diff
@@ -1518,6 +1539,35 @@ def end_tools(self, data):
...
+        unexpected_files = []
+        for dir, _, files in os.walk(tool_dir):
+            dirpath = Path(dir).relative_to(tool_dir)
+            if files:
+                relative_file_path = map(lambda x: f"{dirpath}/{x}", files)
+                unexpected_files += relative_file_path
+
+        if unexpected_files:
+            self.logger.warning(
+                f"{self._hostname}: unexpected temp files {','.join(unexpected_files)}"
```
portante, line 1562: Here we want to use something like:

```
self.logger.warning(
    "%s: unexpected temp files %s", self._hostname, ','.join(unexpected_files)
)
```

lib/pbench/agent/tool_meister.py
```diff
@@ -1518,6 +1539,35 @@ def end_tools(self, data):
...
+            )
+        try:
+            shutil.rmtree(tool_dir)
+        except Exception:
+            self.logger.exception(
+                f"Failed to remove persistent tool data tmp directory: " f"{tool_dir}"
```
portante, line 1568: And again ... `"Failed to remove persistent tool data tmp directory: %s", tool_dir`.

portante (changes requested):
Looks good, just need to fix the log statements.

lib/pbench/agent/tool_meister.py
```diff
@@ -1518,6 +1539,36 @@ def end_tools(self, data):
...
+        for dir, _, files in os.walk(tool_dir):
+            dirpath = Path(dir).relative_to(tool_dir)
+            if files:
+                unexpected_files += map(lambda x: f"{dirpath}/{x}", files)
+
+        if unexpected_files:
+            self.logger.warning(
+                "%s: unexpected temp files %s",
+                self._hostname,
+                ",".join(unexpected_files),
```
dbutenhof, line 1563: While Peter's comment was technically valid, pragmatically we're unlikely to ever want to disable `warning` output ... and even if we did, we're still computing the `",".join` here anyway, which isn't so easy to avoid (if we really care, this would have to be transformed into an `if self.logger.isEnabledFor(logging.WARNING)` block, in which case the f-string would probably be at least as good). Still, this looks a bit more consistent than the f-string form, so that's good.

dbutenhof (approved):

lib/pbench/agent/tool_meister.py
portante: The typical reason I make this comments is more for the discipline, since `.debug` and `.info` statements can become quite costly performance wise as a result of a lapse.

portante (approved):
Ship it!

lib/pbench/agent/tool_meister.py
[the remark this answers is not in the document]
ndokos: Did you mean to change this to an error or warning? AFAICT, it's still `exception`.

ndokos (changes requested):
I flagged a possible oversight.

lib/pbench/agent/tool_meister.py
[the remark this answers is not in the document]
npalaska: I am not sure but this `else` block is no longer there, I dont know why Github did not put the `outdated` label on this conversation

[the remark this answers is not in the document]
npalaska: Oh! I know how this can be confusing, this comment was made when `self.logger.exception` was under the `else` block in my first commit, but it is no longer inside the else block, Idk why githuub UI didnt detect that change here. This is what on line 1563 now `",".join(unexpected_files),`

riya-17 (approved):

lib/pbench/agent/tool_meister.py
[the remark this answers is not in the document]
dbutenhof: Yeah; GitHub is attaching this conversation to the wrong place as if it were current and unchanged. That's a bit disturbing. Also, I thought I'd resolved it ... did someone unresolve it?

(Although, for the line to which it's now attached, this really *doesn't* need to be `self.logger.exception` as the exception traceback isn't particularly important and, given it would all be inside `shutil`, is more likely to be confusing than useful for debugging. *This* might as well be `self.logger.error`... or maybe even another `self.logger.warning` as we continue regardless and it doesn't affect the primary outcome of the agent activity.)

Nevertheless, I'm going to (re-)resolve this as I don't think it's critical for this PR.

ndokos (approved):

lib/pbench/agent/tool_meister.py
```diff
@@ -1534,37 +1540,34 @@ def end_tools(self, data):
...
-                    f"Failed to remove tmp persistent tool data hierarchy, "
-                    f"{directory_to_delete}"
-                )
-                failures += 1
-        else:
+        unexpected_files = []
+        for dir, _, files in os.walk(tool_dir):
+            dirpath = Path(dir).relative_to(tool_dir)
+            if files:
+                unexpected_files += map(lambda x: f"{dirpath}/{x}", files)
```
webbnh, line 1557: FWIW, you don't have to resort to using a `lambda` here -- `str.format()` is a callable which you can use `map()` to apply
```
                unexpected_files += map(f"{dirpath}/{{}}".format, files)
```
or you can just use a list comprehension:
```
                unexpected_files += [f"{dirpath}/{x}" for x in files]
```

lib/pbench/agent/tool_meister.py
```diff
@@ -1534,37 +1540,34 @@ def end_tools(self, data):
...
+
+        if unexpected_files:
+            self.logger.warning(
+                "%s: unexpected temp files %s",
+                self._hostname,
+                ",".join(unexpected_files),
+            )
```
webbnh, line 1564: I would use the following (but, I'm just **that** anal-retentive 😆):
```
        if unexpected_files:
            self.logger.warning(
                "%s: removing unexpected temp file%s: %s",
                self._hostname,
                "s" if len(unexpected_files) > 1 else "",
                ",".join(unexpected_files),
            )
```
(That is, pluralize `file` only if there is more than one file, and add a colon after it; also, add "removing".)

webbnh (approved):
Looks good to me.  Just a couple of aside comments...please mark them as resolved if you decide not to act on them.

lib/pbench/agent/tool_meister.py
npalaska: I liked the suggestion but this case almost will never happen (at least we hope so) and if there are number of files getting written under this directory then either we have a bug somewhere or some other program is using the directory created by us. So I don't want to make change here after I got all the approvals :stuck_out_tongue_closed_eyes:

lib/pbench/agent/tool_meister.py
npalaska: Ah! yeah I should have just used the list comprehension with f-strings. But I dont know if this change worth collapsing all the approvals.

PR #2663, opened 2022-03-05
Scrub the pytests

portante:
Lots of little changes towards a test suite which can be parallelized.

The first change is to remove the use of the session scoped `pytestconfig`.  Session scoped fixtures make it difficult to share them across test sub-systems (one has to be aware of how different sub-systems will interact, which is prone to error).

As a result of dropping `pytestconfig`, `tmp_path` is used instead to give us a proper temporary directory using a fixture that can handle the coming parallelism.

In order to share fixtures between the agent unit and functional tests, a separate on-disk config method allows both fixtures to leverage the disk setup which is fixed and common.  The environment variable monkey-patching is moved to non-session scoped fixtures.

A similar discipline is applied to the server-side on-disk configuration as well.

The `invalid_config` fixture is removed in favor of having the one test the used to handling the monkeypatching, and then make its sister fixture, `valid_config`, be marked `autouse`.

We also take the opportunity to remove any unnecessary fixtures.

This PR is a base enabler for PR #2551.

### DIFF
--- a/lib/pbench/test/__init__.py
+++ b/lib/pbench/test/__init__.py
@@ -1,0 +1,40 @@
+"""Common test support functions."""
+import json
+
+from filelock import FileLock
+from pathlib import Path
+from typing import Callable
+
+
+def on_disk_config(
+    tmp_path_factory, prefix: str, setup_func: Callable[[Path], Path]
+) -> dict:
+    """Base on-disk configuration recorder.  It returns a dictionary with two
+    items, "tmp" and "cfg_dir", two Path objects for the temporary directory
+    in which the configuration directory is created.
+
+    The first argument is the temporary directory creator factory to be used.
+
+    The second argument is a prefix to use for the JSON marker file in which
+    will be store the dictionary containing the temporary directory and the
+    configuration directory created by the setup_func.
+
+    The third argument is the setup function itself.
+
+    The setup_func is expected to take one argument, a Path object
+    representing the created temporary directory, which this method invokes
+    before it is called.
+    """
+    root_tmp_dir = tmp_path_factory.getbasetemp()
+    marker = root_tmp_dir / f"{prefix}-marker.json"
+    with FileLock(f"{marker}.lock"):
+        if marker.is_file():
+            the_setup = json.loads(marker.read_text())
+            the_setup["tmp"] = Path(the_setup["tmp"])
+            the_setup["cfg_dir"] = Path(the_setup["cfg_dir"])
+        else:
+            tmp_d = tmp_path_factory.mktemp(f"{prefix}-tmp")
+            cfg_dir = setup_func(tmp_d)
+            marker.write_text(json.dumps(dict(tmp=str(tmp_d), cfg_dir=str(cfg_dir))))
+            the_setup = dict(tmp=tmp_d, cfg_dir=cfg_dir)
+    return the_setup
--- a/lib/pbench/test/__init__.py
+++ b/lib/pbench/test/__init__.py
@@ -8,7 +8,7 @@ from typing import Callable
 
 def on_disk_config(
     tmp_path_factory, prefix: str, setup_func: Callable[[Path], Path]
-) -> dict[Path, Path]:
+) -> dict[str, Path]:
     """Base on-disk configuration recorder.
 
     Callers use this method to ensure an on-disk configuration is only created
@@ -16,18 +16,19 @@ def on_disk_config(
 
     Args:
         tmp_path_factory:  the temporary directory creator factory to be used
-        prefix:            a value to be prefixed to the name of the JSON marker
-                           file in which will be store the dictionary containing
-                           the temporary directory and the configuration
-                           directory created by the setup_func.
+        prefix:            a value to be prefixed to the name of the JSON
+                           marker file in which will be stored the dictionary
+                           containing the temporary directory and the
+                           configuration directory created by the setup_func
         setup_func:        the setup function provided by the caller, which
                            expects to be called with a Path argument for the
-                           temporary directory to use, and returns a Path object
-                           for the configuration directory created.
+                           temporary directory to use, and returns a Path
+                           object for the configuration directory created
 
     Returns:
         a dictionary with two items, "tmp" and "cfg_dir", two Path objects for
-        the temporary directory in which the configuration directory is created
+        the temporary directory in which the configuration directory is
+        created
     """
     root_tmp_dir = tmp_path_factory.getbasetemp()
     marker = root_tmp_dir / f"{prefix}-marker.json"
--- a/lib/pbench/test/__init__.py
+++ b/lib/pbench/test/__init__.py
@@ -8,27 +8,31 @@ from typing import Callable
 
 def on_disk_config(
     tmp_path_factory, prefix: str, setup_func: Callable[[Path], Path]
-) -> dict:
-    """Base on-disk configuration recorder.  It returns a dictionary with two
-    items, "tmp" and "cfg_dir", two Path objects for the temporary directory
-    in which the configuration directory is created.
+) -> dict[Path, Path]:
+    """Base on-disk configuration recorder.
 
-    The first argument is the temporary directory creator factory to be used.
+    Callers use this method to ensure an on-disk configuration is only created
+    once for a given prefix.
 
-    The second argument is a prefix to use for the JSON marker file in which
-    will be store the dictionary containing the temporary directory and the
-    configuration directory created by the setup_func.
+    Args:
+        tmp_path_factory:  the temporary directory creator factory to be used
+        prefix:            a value to be prefixed to the name of the JSON marker
+                           file in which will be store the dictionary containing
+                           the temporary directory and the configuration
+                           directory created by the setup_func.
+        setup_func:        the setup function provided by the caller, which
+                           expects to be called with a Path argument for the
+                           temporary directory to use, and returns a Path object
+                           for the configuration directory created.
 
-    The third argument is the setup function itself.
-
-    The setup_func is expected to take one argument, a Path object
-    representing the created temporary directory, which this method invokes
-    before it is called.
+    Returns:
+        a dictionary with two items, "tmp" and "cfg_dir", two Path objects for
+        the temporary directory in which the configuration directory is created
     """
     root_tmp_dir = tmp_path_factory.getbasetemp()
     marker = root_tmp_dir / f"{prefix}-marker.json"
     with FileLock(f"{marker}.lock"):
-        if marker.is_file():
+        if marker.exists():
             the_setup = json.loads(marker.read_text())
             the_setup["tmp"] = Path(the_setup["tmp"])
             the_setup["cfg_dir"] = Path(the_setup["cfg_dir"])
--- a/lib/pbench/test/functional/agent/conftest.py
+++ b/lib/pbench/test/functional/agent/conftest.py
@@ -2,13 +2,14 @@ import os
 import pytest
 import subprocess
 
-from pbench.test.unit.agent.conftest import on_disk_agent_config
+from pbench.test import on_disk_config
+from pbench.test.unit.agent.conftest import do_setup
 
 
 @pytest.fixture(scope="session", autouse=True)
 def func_setup(tmp_path_factory):
     """Test package setup for functional tests"""
-    return on_disk_agent_config(tmp_path_factory)
+    return on_disk_config(tmp_path_factory, "agent", do_setup)
 
 
 @pytest.helpers.register
@@ -19,7 +20,7 @@ def capture(command):
 
 
 @pytest.fixture
-def pbench_run(monkeypatch, tmp_path):
+def pbench_run(tmp_path):
     assert "pbench_run" not in os.environ
     pbench_agent_d = tmp_path / "var" / "lib" / "pbench-agent"
     pbench_agent_d.mkdir(parents=True, exist_ok=True)
--- a/lib/pbench/test/functional/agent/conftest.py
+++ b/lib/pbench/test/functional/agent/conftest.py
@@ -2,12 +2,13 @@ import os
 import pytest
 import subprocess
 
-from pbench.test.unit.agent.conftest import base_setup
+from pbench.test.unit.agent.conftest import on_disk_agent_config
 
 
 @pytest.fixture(scope="session", autouse=True)
-def setup(request, pytestconfig):
-    return base_setup(request, pytestconfig)
+def func_setup(tmp_path_factory):
+    """Test package setup for functional tests"""
+    return on_disk_agent_config(tmp_path_factory)
 
 
 @pytest.helpers.register
--- a/lib/pbench/test/functional/agent/conftest.py
+++ b/lib/pbench/test/functional/agent/conftest.py
@@ -1,4 +1,5 @@
 import os
+from pathlib import Path
 import pytest
 import subprocess
 
@@ -7,7 +8,7 @@ from pbench.test.unit.agent.conftest import do_setup
 
 
 @pytest.fixture(scope="session", autouse=True)
-def func_setup(tmp_path_factory):
+def func_setup(tmp_path_factory) -> dict[str, Path]:
     """Test package setup for functional tests"""
     return on_disk_config(tmp_path_factory, "agent", do_setup)
 
--- a/lib/pbench/test/unit/agent/conftest.py
+++ b/lib/pbench/test/unit/agent/conftest.py
@@ -1,9 +1,7 @@
-import json
 import pytest
 import shutil
 
-from filelock import FileLock
-from pathlib import Path
+from pbench.test import on_disk_config
 
 
 agent_cfg_tmpl = """[DEFAULT]
@@ -20,12 +18,15 @@ files = pbench-agent-default.cfg
 """
 
 
-def do_setup(tmp_path_factory):
-    """Perform on disk agent config setup."""
-    # Create a single temporary directory for the "/opt/pbench-agent" and
-    # "/var/lib/pbench-agent" directories.
-    tmp_d = tmp_path_factory.mktemp("agent-tmp")
+def do_setup(tmp_d):
+    """Perform on disk agent config setup.
 
+    Accept a single temporary directory created by its caller, creating a
+    proper pbench directory hierarchy in it, with appropriately constructed
+    configuration files.
+
+    Returns a Path object for the created pbench configuration directory.
+    """
     opt_pbench = tmp_d / "opt" / "pbench-agent"
     pbench_cfg = opt_pbench / "config"
     pbench_cfg.mkdir(parents=True, exist_ok=True)
@@ -39,39 +40,19 @@ def do_setup(tmp_path_factory):
     with open(cfg_file, "w") as fp:
         fp.write(agent_cfg_tmpl.format(TMP=tmp_d))
 
-    return dict(tmp=tmp_d, cfg_dir=pbench_cfg)
-
-
-def on_disk_agent_config(tmp_path_factory):
-    """Base setup function shared between the agent and functional tests.
-
-    See https://github.com/pytest-dev/pytest-xdist.
-    """
-    root_tmp_dir = tmp_path_factory.getbasetemp()
-    marker = root_tmp_dir / "agent-marker.json"
-    with FileLock(f"{marker}.lock"):
-        if marker.is_file():
-            the_setup = json.loads(marker.read_text())
-            the_setup["tmp"] = Path(the_setup["tmp"])
-            the_setup["cfg_dir"] = Path(the_setup["cfg_dir"])
-        else:
-            the_setup = do_setup(tmp_path_factory)
-            data = dict(tmp=str(the_setup["tmp"]), cfg_dir=str(the_setup["cfg_dir"]))
-            marker.write_text(json.dumps(data))
-    return the_setup
+    return pbench_cfg
 
 
 @pytest.fixture(scope="session", autouse=True)
 def agent_setup(tmp_path_factory):
     """Test package setup for agent unit tests"""
-    on_disk_agent_config(tmp_path_factory)
+    return on_disk_config(tmp_path_factory, "agent", do_setup)
 
 
 @pytest.fixture(autouse=True)
-def setup(tmp_path_factory, monkeypatch):
+def setup(tmp_path_factory, agent_setup, monkeypatch):
     """Test package setup for pbench-agent"""
-    _setup = on_disk_agent_config(tmp_path_factory)
-    var = _setup["tmp"] / "var" / "lib" / "pbench-agent"
+    var = agent_setup["tmp"] / "var" / "lib" / "pbench-agent"
     try:
         if var.exists():
             assert var.is_dir()
@@ -80,6 +61,6 @@ def setup(tmp_path_factory, monkeypatch):
         print(exc)
     var.mkdir(parents=True, exist_ok=True)
     monkeypatch.setenv(
-        "_PBENCH_AGENT_CONFIG", str(_setup["cfg_dir"] / "pbench-agent.cfg")
+        "_PBENCH_AGENT_CONFIG", str(agent_setup["cfg_dir"] / "pbench-agent.cfg")
     )
-    return _setup
+    return agent_setup
--- a/lib/pbench/test/unit/agent/conftest.py
+++ b/lib/pbench/test/unit/agent/conftest.py
@@ -1,44 +1,9 @@
-import shutil
-import tempfile
+import json
 import pytest
+import shutil
 
+from filelock import FileLock
 from pathlib import Path
-
-
-def base_setup(request, pytestconfig):
-    """Test package setup for pbench-agent"""
-
-    # Create a single temporary directory for the "/opt/pbench-agent" and
-    # "/var/lib/pbench-agent" directories.
-    TMP = tempfile.TemporaryDirectory(suffix=".d", prefix="pbench-agent-unit-tests.")
-    tmp_d = Path(TMP.name)
-
-    opt_pbench = tmp_d / "opt" / "pbench-agent"
-    pbench_cfg = opt_pbench / "config"
-    pbench_cfg.mkdir(parents=True, exist_ok=True)
-
-    var = tmp_d / "var" / "lib" / "pbench_agent"
-    var.mkdir(parents=True, exist_ok=True)
-
-    shutil.copyfile(
-        "./agent/config/pbench-agent-default.cfg",
-        str(pbench_cfg / "pbench-agent-default.cfg"),
-    )
-
-    pytestconfig.cache.set("TMP", TMP.name)
-    cfg_file = pbench_cfg / "pbench-agent.cfg"
-    pytestconfig.cache.set("_PBENCH_AGENT_CONFIG", str(cfg_file))
-
-    def teardown():
-        """Test package teardown for pbench-agent"""
-        TMP.cleanup()
-
-    request.addfinalizer(teardown)
-
-
-@pytest.fixture(scope="session", autouse=True)
-def setup(request, pytestconfig):
-    return base_setup(request, pytestconfig)
 
 
 agent_cfg_tmpl = """[DEFAULT]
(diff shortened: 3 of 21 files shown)


portante (commented):
Ready for review.

lib/pbench/test/unit/server/conftest.py
```diff
@@ -96,31 +93,43 @@ def setup(request, pytestconfig):
...
+
+    return dict(tmp=tmp_d, cfg_file=str(cfg_file))
 
-    pytestconfig.cache.set("TMP", TMP.name)
-    pytestconfig.cache.set("_PBENCH_SERVER_CONFIG", str(cfg_file))
 
-    def teardown():
-        """Test package teardown for pbench-server"""
-        TMP.cleanup()
+@pytest.fixture(scope="session", autouse=True)
```
dbutenhof, line 101: Is there a point to the `autouse=True` if it's got a *value* that's required by dependent fixtures? To get the value you have to reference it in the fixture signature anyway. Or are there fixtures I missed that simply expect to find the file on disk? (And, if so, why don't they all work that way instead?)

Or to put it another way, if a JSON file falls in the test case and nobody's there to read it, does it make a sound?

dbutenhof (dismissed):
Looks like a nice simplification and cleanup of some of the low-level cruft; cool.

lib/pbench/test/unit/agent/conftest.py
```diff
@@ -1,77 +1,85 @@
...
-@pytest.fixture
-def invalid_config(pytestconfig):
-    cfg_file = pytestconfig.cache.get("_PBENCH_AGENT_CONFIG", None)
-    shutil.copyfile(
-        "./lib/pbench/test/unit/agent/config/pbench-agent.invalid.cfg", cfg_file
-    )
+@pytest.fixture(scope="session", autouse=True)
+def agent_setup(tmp_path_factory):
+    """Test package setup for agent unit tests"""
+    on_disk_agent_config(tmp_path_factory)
```
ndokos, line 67: Is there a `return` missing here?

ndokos (changes requested):
Partial review. I'll get back to it later.

lib/pbench/test/unit/agent/conftest.py
dbutenhof: It's marked `autouse=True`, so it should be pulled in on every test session automagically.

portante: So missing return.  The `agent_setup` is an `autouse` fixture, see line 64 above.   It is automatically run before each test session begins.

lib/pbench/test/unit/server/conftest.py
portante: Since `on_disk_server_config` is a "session" scoped fixture, it is run once before a test thread starts executing tests.  But, other fixtures, see `server_config` below, rely on its data structure to find the `cfg_file` that is setup for tests.

portante (commented):
Seems like this could be merged now.

lib/pbench/test/unit/server/conftest.py
dbutenhof: Sure; but the root question is that if the *value* is important to tests that depend on this fixture, then it has to be referenced specifically by that test or some fixture used by the test; so what is the `autouse` buying us? Is the setup here important for tests that don't reference this fixture or any fixture that references it? If so, then a comment might be in order. If not, then it would seem that the `autouse` is serving no purpose. (Though I approved on the basis that it probably doesn't hurt, it may somewhat obfuscate the actual scope and dependencies.)

portante: Not sure `autouse` matters here since `server_config` is pretty much used all over the place, and removing the `autouse` parameter did nothing to affect the unit tests.

I think the `server_config` could be marked `autouse`, and its references removed except for the tests that need the value.

But that seems like a separate piece of work.

dbutenhof (approved):

npalaska (approved):

npalaska (dismissed):

riya-17 (dismissed):

dbutenhof (approved):

dbutenhof:
Hmm; the only open conversation is Nick's fixture question, which was answered, so I suspect it'd be reasonable to resolve it and merge, but that feels awkward. 😆

portante:
@webbnh has expressed an interest in reviewing this PR, so I am going to hold off merging.

lib/pbench/test/unit/agent/test_tool_data_sink.py
```diff
@@ -26,10 +25,9 @@ class TestBenchmarkRunDir:
...
-    def cleanup_tmp(self, pytestconfig):
-        TMP = Path(pytestconfig.cache.get("TMP", None))
-        self.int_pb_run = TMP / "pbench-run-int"
-        self.ext_pb_run = TMP / "pbench-run-ext"
+    def cleanup_tmp(self, tmp_path):
+        self.int_pb_run = tmp_path / "pbench-run-int"
+        self.ext_pb_run = tmp_path / "pbench-run-ext"
         yield
         try:
             shutil.rmtree(self.int_pb_run)
```
webbnh, line 33: If `rmtree()` fails, here, do we get the behavior that we desire?  (I.e., I think line 35 will probably result in any exception being silently ignored.)  Ditto for line 39.

(Since we expect the removals to succeed, and since we take the same actions for each failure, we might as well do them in the same `try` block, no?)

webbnh (dismissed):
There are some nice improvements here!  However, I've got the usual pile of nits, suggestions, and questions.

In particularly, I'm unclear on why `on_disk_agent_config()` on the Agent side is a helper function (instead of a fixture), when it is defined as a fixture for the Server side.  If it can't be a fixture on the Agent side, then perhaps it would be worthwhile to name it differently -- I think there is some value to having consistency in the respective infrastructure for the Agent and Server.

lib/pbench/test/functional/agent/conftest.py
```diff
@@ -2,12 +2,13 @@
...
-from pbench.test.unit.agent.conftest import base_setup
+from pbench.test.unit.agent.conftest import on_disk_agent_config
 
 
 @pytest.fixture(scope="session", autouse=True)
-def setup(request, pytestconfig):
-    return base_setup(request, pytestconfig)
+def func_setup(tmp_path_factory):
+    """Test package setup for functional tests"""
+    return on_disk_agent_config(tmp_path_factory)
```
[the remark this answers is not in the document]
portante: Because we don't want global change to the `os.environ` for every test.

lib/pbench/test/functional/agent/conftest.py
```diff
@@ -2,12 +2,13 @@
 import pytest
 import subprocess
 
-from pbench.test.unit.agent.conftest import base_setup
+from pbench.test.unit.agent.conftest import on_disk_agent_config
```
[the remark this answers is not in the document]
portante: Sort of.  I think we want to share the base behavior of the on-disk setup, AND the agent configuration, which is not common to the server.  I'll propose something shortly.

lib/pbench/test/functional/agent/conftest.py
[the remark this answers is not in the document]
portante: I have refactored things a bit, perhaps that will make it clearer?

lib/pbench/test/unit/agent/test_config.py
```diff
@@ -1,30 +1,38 @@
...
+def test_invalid_config(setup, monkeypatch):
+    shutil.copyfile(
+        "./lib/pbench/test/unit/agent/config/pbench-agent.invalid.cfg",
+        str(setup["cfg_dir"] / "pbench-agent.invalid.cfg"),
+    )
+    monkeypatch.setenv(
+        "_PBENCH_AGENT_CONFIG", str(setup["cfg_dir"] / "pbench-agent.invalid.cfg")
+    )
     with pytest.raises(BadConfig):
         PbenchAgentConfig(os.environ["_PBENCH_AGENT_CONFIG"])
```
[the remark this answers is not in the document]
portante: Yes, this is odd.

lib/pbench/test/unit/agent/test_tool_data_sink.py
portante: Not sure, but I think the goal was to squash the exception. Would have to look at the history of this test to be sure.

portante: History doesn't help.  This code landed with the containerized Tool Meister work in PR #2076, see https://github.com/distributed-system-analysis/pbench/pull/2076/files#diff-a43ec41cd50cf299ee7d47c0bc84490551777c991833971e49984fe54cf5b0c5R35.

lib/pbench/test/unit/agent/conftest.py
```diff
@@ -1,77 +1,85 @@
...
-        TMP.cleanup()
-
-    request.addfinalizer(teardown)
-
-
-@pytest.fixture(scope="session", autouse=True)
-def setup(request, pytestconfig):
-    return base_setup(request, pytestconfig)
+    with open(cfg_file, "w") as fp:
+        fp.write(agent_cfg_tmpl.format(TMP=tmp_d))
```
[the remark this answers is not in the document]
portante: Fixed.

lib/pbench/test/unit/agent/test_config.py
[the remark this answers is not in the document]
portante: Fixed.

lib/pbench/test/unit/agent/conftest.py
```diff
@@ -1,77 +1,85 @@
...
+def setup(tmp_path_factory, monkeypatch):
+    """Test package setup for pbench-agent"""
+    _setup = on_disk_agent_config(tmp_path_factory)
+    var = _setup["tmp"] / "var" / "lib" / "pbench-agent"
+    try:
+        if var.exists():
+            assert var.is_dir()
+            shutil.rmtree(str(var))
+    except Exception as exc:
+        print(exc)
```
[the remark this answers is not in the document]
portante: Fixed.

lib/pbench/test/unit/agent/test_tool_data_sink.py
portante: Fixed.

lib/pbench/test/unit/agent/conftest.py
```diff
@@ -1,77 +1,85 @@
...
 
+from filelock import FileLock
 from pathlib import Path
 
 
-def base_setup(request, pytestconfig):
-    """Test package setup for pbench-agent"""
+agent_cfg_tmpl = """[DEFAULT]
+pbench_install_dir = {TMP}/opt/pbench-agent
+pbench_results_redirector = pbench.results.example.com
```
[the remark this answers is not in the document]
portante: The configuration file itself uses it.

portante (commented):
I believe I have addressed all the code-review feedback.

riya-17 (approved):

lib/pbench/test/unit/agent/test_tool_data_sink.py
```diff
@@ -31,12 +31,16 @@ def cleanup_tmp(self, tmp_path):
...
-        except Exception as exc:
-            print(exc)
+        except FileNotFoundError:
+            pass
+        except NotADirectoryError:
+            self.int_pb_run.unlink()
```
dbutenhof, line 37: Hmm; this is cleanup code after the `yield`, so here we're *accepting* a test run that somehow creates a file where we expect either nothing (is it actually OK if the associated test cases don't even create the directory?) or a directory. Doesn't it sound like an error in the tests that we might detect here?

And the previous code did the `print(exc)` (which is rather anemic) if the directory didn't exist ... should it actually be an error if the test case didn't create the directory?

lib/pbench/test/unit/agent/conftest.py
```diff
@@ -39,39 +40,19 @@ def do_setup(tmp_path_factory):
...
 @pytest.fixture(scope="session", autouse=True)
 def agent_setup(tmp_path_factory):
     """Test package setup for agent unit tests"""
-    on_disk_agent_config(tmp_path_factory)
+    return on_disk_config(tmp_path_factory, "agent", do_setup)
 
 
 @pytest.fixture(autouse=True)
-def setup(tmp_path_factory, monkeypatch):
+def setup(tmp_path_factory, agent_setup, monkeypatch):
```
webbnh, line 52: Just an observation:  since `setup()` is an auto-use fixture which references the `agent_setup()` fixture, that means that`agent_setup()` will always be used, regardless of whether it is marked as `autouse` or not.

lib/pbench/test/unit/server/conftest.py
```diff
@@ -64,12 +63,10 @@
 generic_password = "12345"
 
 
-def do_setup(tmp_path_factory) -> dict:
+def do_setup(tmp_d: Path) -> dict:
```
webbnh, line 66: I think the return type is now `Path`, not `dict`.

lib/pbench/test/unit/agent/test_tool_data_sink.py
portante: What do you think this should be?

I don't know of another way to do this, other than have the individual tests not use a fixture, but create and destroy what is appropriate for their test case all within the test.

We can use `shutil.rmtree()` on a non-existent file, and the behavior of the test determines whether or not it gets created.

We can use `Path.unlink()` because it might be a directory, and the directory hierarchy won't get removed.

So this seemed like the best approach.

If you have another way to do this without re-writing unit tests (which would be a whole 'nother level of effort) I am all ears.

lib/pbench/test/unit/server/conftest.py
```diff
@@ -96,31 +94,43 @@ def setup(request, pytestconfig):
...
+    marker = root_tmp_dir / "server-marker.json"
+    with FileLock(f"{marker}.lock"):
+        if marker.is_file():
+            the_setup = json.loads(marker.read_text())
+            the_setup["tmp"] = Path(the_setup["tmp"])
+        else:
+            the_setup = do_setup(tmp_path_factory)
+            data = dict(tmp=str(the_setup["tmp"]), cfg_file=the_setup["cfg_file"])
+            marker.write_text(json.dumps(data))
+    return the_setup
```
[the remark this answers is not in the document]
portante: Please let me know if the new refactoring addresses your comment.

lib/pbench/test/unit/agent/conftest.py
```diff
@@ -1,77 +1,85 @@
...
-[pbench-agent]
-debug_unittest = 1
-pbench_run = {TMP}/var/lib/pbench-agent
-
-[config]
-path = %(pbench_install_dir)s/config
-files = pbench-agent-default.cfg
-"""
+def on_disk_agent_config(tmp_path_factory):
+    """Base setup function shared between the agent and functional tests.
```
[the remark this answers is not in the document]
portante: I believe I have addressed this to the best of my ability, but type hints and pytest fixtures don't always mix.

lib/pbench/test/unit/agent/conftest.py
```diff
@@ -1,77 +1,85 @@
...
 
+    See https://github.com/pytest-dev/pytest-xdist.
```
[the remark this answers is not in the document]
portante: I just removed it.

lib/pbench/test/unit/agent/conftest.py
```diff
@@ -1,77 +1,85 @@
...
+pbench_run = {TMP}/var/lib/pbench-agent
+
+[config]
+path = %(pbench_install_dir)s/config
+files = pbench-agent-default.cfg
+"""
+
 
+def do_setup(tmp_path_factory):
+    """Perform on disk agent config setup."""
```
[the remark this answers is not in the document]
portante: Please see if the refactoring has addressed this sufficiently.

lib/pbench/test/unit/agent/conftest.py
```diff
@@ -1,77 +1,85 @@
...
-"""
+def on_disk_agent_config(tmp_path_factory):
+    """Base setup function shared between the agent and functional tests.
 
+    See https://github.com/pytest-dev/pytest-xdist.
+    """
+    root_tmp_dir = tmp_path_factory.getbasetemp()
+    marker = root_tmp_dir / "agent-marker.json"
+    with FileLock(f"{marker}.lock"):
+        if marker.is_file():
```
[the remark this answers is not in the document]
portante: Okay, I'll change it to `exists()`.

lib/pbench/test/unit/agent/test_tool_data_sink.py
webbnh: First, I think that this change is better than it was.

Beyond that, I agree that it would be better still if this code detected incorrect behavior and caused the test to fail.

However, in this context (without adding a little superstructure), I don't think we can tell whether the file/directory should have existed or not.  (That is, we might have tests which use this fixture which, in their "success scenario", don't create both files...in which case we don't want this fixture to generate a failure for that.)

lib/pbench/test/functional/agent/conftest.py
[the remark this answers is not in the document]
webbnh: It is much nicer, now!  (I have another comment, but I'll start a new "conversation" for it.)

lib/pbench/test/unit/agent/conftest.py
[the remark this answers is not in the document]
webbnh: I must be blind...I see where it _sets_ it, but I don't see any references to it....  Help?

lib/pbench/test/unit/agent/conftest.py
```diff
@@ -55,23 +18,48 @@ def setup(request, pytestconfig):
...
-        fp.write(agent_cfg_tmpl.format(TMP=TMP))
+def do_setup(tmp_d):
+    """Perform on disk agent config setup.
+
+    Accept a single temporary directory created by its caller, creating a
+    proper pbench directory hierarchy in it, with appropriately constructed
+    configuration files.
 
+    Returns a Path object for the created pbench configuration directory.
+    """
```
webbnh, line 29: Type hints?

lib/pbench/test/unit/agent/conftest.py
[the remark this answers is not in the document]
webbnh: It still needs type hints.  🙂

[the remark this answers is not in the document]
webbnh: But, the new description is much better!

lib/pbench/test/unit/agent/conftest.py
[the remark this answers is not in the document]
webbnh: There's a great deal of chatter in the PyTest Issues about them.  😀

lib/pbench/test/unit/server/conftest.py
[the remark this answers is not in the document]
webbnh: It's a thing of beauty.  Thanks for doing it!

webbnh (approved):
I think this looks good, now...modulo a few type hints.

lib/pbench/test/unit/agent/conftest.py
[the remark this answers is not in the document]
portante: https://github.com/distributed-system-analysis/pbench/blob/main/agent/config/pbench-agent-default.cfg#L24

lib/pbench/test/unit/agent/test_tool_data_sink.py
portante: So can we resolve this?

lib/pbench/test/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+from pathlib import Path
+from typing import Callable
+
+
+def on_disk_config(
+    tmp_path_factory, prefix: str, setup_func: Callable[[Path], Path]
+) -> dict:
+    """Base on-disk configuration recorder.  It returns a dictionary with two
+    items, "tmp" and "cfg_dir", two Path objects for the temporary directory
+    in which the configuration directory is created.
```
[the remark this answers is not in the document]
portante: It is `dict[Path, Path]` I believe, but I am not sure how valuable that is for the type hint.

lib/pbench/test/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+
+    The first argument is the temporary directory creator factory to be used.
+
+    The second argument is a prefix to use for the JSON marker file in which
+    will be store the dictionary containing the temporary directory and the
+    configuration directory created by the setup_func.
+
+    The third argument is the setup function itself.
```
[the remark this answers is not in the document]
portante: The reason I kept it that way is that the parameter class is not exposed anywhere that I could find to import.  So keeping it the same name as the fixture seemed the best course of action.

I can change the comment use the parameter names.

npalaska (dismissed):

portante (commented):
Okay ready for perhaps the last round of reviews?

lib/pbench/test/unit/agent/conftest.py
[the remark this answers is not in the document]
webbnh: Ah...um...well...no.  😮 

That reference is to `pbench_result_redirector` (singluar), whereas the definition is for `pbench_results_redirector` (plural)!

So...does that mean we haven't been using the default value for `host_path`, or is there some clever fallback somewhere??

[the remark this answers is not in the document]
webbnh: (I don't see any references to `host_path`....)

lib/pbench/test/unit/agent/test_tool_data_sink.py
webbnh: I'm good with it the way it is currently proposed to be.

lib/pbench/test/__init__.py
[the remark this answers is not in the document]
webbnh: Changing the comment would be good, regardless.

I don't see how the lack of an import drives the name of the parameter.  Given that this is Python, we don't actually have to know the type of the parameters until call-time (and, even then, we can be flexible...).  Keeping it the same as the fixture seems likely to create confusion:  that is, receiving the value of a fixture is different from referencing one -- I think it would be good to be clear that this function is not referencing the fixture (i.e., it is not causing the fixture to be included/invoked), rather it is making use of the value that the fixture provides to this function's caller.

(And, yeah, I'm willing to take a pass on the type hint for it...although, I s'pose we could use `Any`, but that's probably no better than just omitting it.)

lib/pbench/test/__init__.py
```diff
@@ -8,27 +8,31 @@
 
 def on_disk_config(
     tmp_path_factory, prefix: str, setup_func: Callable[[Path], Path]
-) -> dict:
-    """Base on-disk configuration recorder.  It returns a dictionary with two
-    items, "tmp" and "cfg_dir", two Path objects for the temporary directory
-    in which the configuration directory is created.
+) -> dict[Path, Path]:
```
webbnh, line 11: I think the return type is `dict[str, Path]`.  (Yeah, there are two `Path`'s in the `dict`, but their keys are `str`'s.)

lib/pbench/test/__init__.py
```diff
@@ -8,27 +8,31 @@
...
+    Callers use this method to ensure an on-disk configuration is only created
+    once for a given prefix.
 
-    The second argument is a prefix to use for the JSON marker file in which
-    will be store the dictionary containing the temporary directory and the
-    configuration directory created by the setup_func.
+    Args:
+        tmp_path_factory:  the temporary directory creator factory to be used
+        prefix:            a value to be prefixed to the name of the JSON marker
+                           file in which will be store the dictionary containing
```
webbnh, line 20: s/store/stored/

lib/pbench/test/__init__.py
webbnh: Note that `on_disk_config()`'s caller's return types (and one of their callers) have the same problem.

npalaska (approved):

lib/pbench/test/__init__.py
dbutenhof, line 11: Technically, this should be `Dict[str, Path]` because you're mapping a string key to a `Path` value ...

webbnh (approved):
There's still a problem with the type hints.

The new `on_disk_config()` docstring is wonderful, but there's a typo in it, and it doesn't mention parallel execution.  And, there's the lingering question of how to spell its first parameter's name.

lib/pbench/test/unit/agent/conftest.py
```diff
@@ -42,14 +43,14 @@ def do_setup(tmp_d):
     return pbench_cfg
 
 
-@pytest.fixture(scope="session", autouse=True)
-def agent_setup(tmp_path_factory):
+@pytest.fixture(scope="session")
+def agent_setup(tmp_path_factory) -> dict[Path, Path]:
```
dbutenhof, line 47: Same here: `Dict[str, Path]`

lib/pbench/test/unit/agent/conftest.py
```diff
@@ -42,14 +43,14 @@ def do_setup(tmp_d):
...
     """Test package setup for agent unit tests"""
     return on_disk_config(tmp_path_factory, "agent", do_setup)
 
 
 @pytest.fixture(autouse=True)
-def setup(tmp_path_factory, agent_setup, monkeypatch):
+def setup(tmp_path_factory, agent_setup, monkeypatch) -> dict[Path, Path]:
```
dbutenhof, line 53: `Dict[str, Path]`

lib/pbench/test/unit/server/conftest.py
```diff
@@ -96,7 +96,7 @@ def do_setup(tmp_d: Path) -> dict:
 
 
 @pytest.fixture(scope="session")
-def on_disk_server_config(tmp_path_factory) -> dict:
+def on_disk_server_config(tmp_path_factory) -> dict[Path, Path]:
```
dbutenhof, line 99: `Dict[str, Path]`

lib/pbench/test/unit/agent/test_tool_data_sink.py
dbutenhof: Yeah; if we have tests that really do create regular files here, then Webb's correct that the mechanism to manage failure here probably isn't justified. Ah, well.

lib/pbench/test/__init__.py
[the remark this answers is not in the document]
portante: First, a non-test function physically can't reference a fixture.  It is a parameter argument, and so it is always the name of the first parameter.  I'd rather keep the argument to be the same name as the fixture for two reasons:

 1. Naming it something else is likely to cause confusion because we can't type hint it, and so unless you really understand the nature of the callers and their fixture, it won't be clear what to pass here
 2. I don't know what to name it that would help make it any clearer that we really want to use a `TempPathFactory` class only available from a fixture of the name `tmp_path_factory`

lib/pbench/test/__init__.py
dbutenhof: BTW:

```
*: Note: If you are using Python 3.7 or newer, you can use dict as a generic type if you start your module with [from __future__ import annotations](https://www.python.org/dev/peps/pep-0563/), and as of Python 3.9, dict (as well as other standard containers) [supports being used as generic type even without that directive](https://www.python.org/dev/peps/pep-0585/).
```

We still have Python 3.6 platforms (RHEL 8), so until we're "all 9 all the time" we should stick with using the `typing` symbols for collections rather than try to use the raw type as a generic.

portante: I am sorry, I really don't see the value of Python type hints.  They serve more to obfuscate code that add clarity, and they just cause additional imports to be added to modules just to reference the types, and don't help during execution.  I'll fix this, but we really need to review how much value they add.

lib/pbench/test/__init__.py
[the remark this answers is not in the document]
webbnh: > a non-test function physically can't reference a fixture

Right(*)...which is a big part of why it is so confusing to see one look like it's doing exactly that.

> Naming it something else is likely to cause confusion because we can't type hint it

Aside from the fact that we _can_ apply a hint to it (if nothing else, I believe that we can put the hint in double quotes to escape definition problems, which means we can label it anything that we think will help the reader)...the name only has to be meaningful within the context of the function (once we understand that it's not a fixture...).  So, I think we can find a spelling which is expressive without being confusing.

> I don't know what to name it

We could call it `path_factory`, since, in this context, I don't think we care whether it is temporary or not....

Anyway, I think we have given this issue due consideration, and I think we have bigger fish to fry, so I'm going to mark this resolved, now.


---------
(*)Ignoring the fact that other fixtures are also non-test functions which _can_ reference other fixtures.

lib/pbench/test/__init__.py
dbutenhof: I think the *documentation* is useful, both for the developer and for subsequent readers. And we could be running a static analysis tool that actually checks this, which would probably be a good idea.

But you're right that the Python compiler doesn't care, so I won't complain that you didn't use `Dict[str, Path]`... and `dict[str, Path]` is wrong until Python 3.9. I'm not sure what Python we're running in the Jenkins container, and if it's 3.9 I hope this won't cause problems on earlier versions. However, for now I'm going to approve and resolve this conversation.

dbutenhof (approved):

webbnh (approved):
👍

lib/pbench/test/unit/agent/conftest.py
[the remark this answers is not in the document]
webbnh: Tracked in #2669.

lib/pbench/test/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def on_disk_config(
+    tmp_path_factory, prefix: str, setup_func: Callable[[Path], Path]
+) -> dict:
+    """Base on-disk configuration recorder.  It returns a dictionary with two
+    items, "tmp" and "cfg_dir", two Path objects for the temporary directory
+    in which the configuration directory is created.
+
+    The first argument is the temporary directory creator factory to be used.
```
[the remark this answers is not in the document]
webbnh: 🦐

npalaska (approved):

PR #2664, opened 2022-03-07
Add `pbench-run-benchmark.pl` to the Agent's installed files list

webbnh:
`pbench-run-benchmark` is a trampline bash script which sets up the environment and then transfers control to a Perl script, `pbench-run-benchmark.pl`.  (This seems to be a unique case in Pbench at the moment....)  Unfortunately, the target of the transfer was never added to the list of files to be installed.  This PR remedies that omission.

### DIFF
--- a/agent/Makefile
+++ b/agent/Makefile
@@ -82,6 +82,7 @@ bench-scripts = \
 	pbench-migrate \
 	pbench-netperf \
 	pbench-run-benchmark \
+	pbench-run-benchmark.pl \
 	pbench-run-benchmark-sample \
 	pbench-specjbb2005 \
 	pbench-trafficgen \


ndokos (approved):

ndokos:
I think it's the only example of a Perl script that is called directly by a user, so the trampoline probably sets up env variable and such.

dbutenhof (approved):

portante (approved):

webbnh:
> the trampoline probably sets up env variable and such.

Ah, yes -- it `source`'s the Agent's `base` script to set up all of our favorite context -- I've edited the PR description to reflect that knowledge.

However, while this PR does address the symptom which Lukas reported, the results don't seem to actually work for me.  Nevertheless, since I don't think this change makes anything _worse_, I'm going to go ahead and merge it (but I'll leave the #2624 open for the moment).

PR #2670, opened 2022-03-09
Remove references to the `pbench_results_redirector` configuration setting

portante:
Remove references to the `pbench_results_redirector` configuration setting.

The configuration setting was not used, and it was incorrectly spelled in the default agent config file.  As a result of removing that, we also remove unnecessary `webserver` and `host_path` variables.

Found while reviewing PR #2663.

### DIFF
--- a/agent/config/pbench-agent-default.cfg
+++ b/agent/config/pbench-agent-default.cfg
@@ -21,10 +21,8 @@ logger_type = file
 
 [results]
 user = pbench
-host_path = http://%(pbench_result_redirector)s/pbench-archive-host
-webserver = %(pbench_web_server)s
 host_info_uri = pbench-results-host-info.versioned/pbench-results-host-info.URL002
-host_info_url = http://%(webserver)s/%(host_info_uri)s
+host_info_url = http://%(pbench_web_server)s/%(host_info_uri)s
 dir = /srv/pbench/public_html/incoming
 scp_opts = -o BatchMode=yes -o StrictHostKeyChecking=no
 ssh_opts = -o BatchMode=yes -o StrictHostKeyChecking=no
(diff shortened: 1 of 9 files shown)


dbutenhof (dismissed):

agent/config/pbench-agent-default.cfg
```diff
@@ -21,7 +21,7 @@ logger_type = file
 
 [results]
 user = pbench
-host_path = http://%(pbench_result_redirector)s/pbench-archive-host
+host_path = http://%(pbench_results_redirector)s/pbench-archive-host
```
webbnh, line 24: Where is `host_path` used?

dbutenhof: That, actually, seems to be a really good question. I find no trace of the string host_path anywhere in the source tree. And, furthermore, since `pbench_result_redirector` ought to be defined to the server hostname, the path `http://{host}/pbench-archive-host` (e.g., `/var/www/html/pbench-archive-host`) doesn't actually exist on a Pbench server.

This line was last changed about 3 years ago by Pete, but the PR says nothing about it so I assume (I didn't look) that it was a minor refactoring or rearrangement.

dbutenhof: And in fact, since the spelling error goes back at least to 2019 and nobody's complained, we might even consider that proof that nobody's used it?

portante: Okay, went "whole-hog."

portante (commented):
Fixed the test problem.

dbutenhof (approved):
Ha; you're right, that weird `pbench_results_redirector` wasn't used anywhere else, either. Nice cleanup!

npalaska (approved):

agent/config/pbench-agent-default.cfg
portante: @webbnh, do we need more discussion on this change?

webbnh (approved):
This is great!  Thanks for following through on this!

PR #2671, opened 2022-03-10
Ensure agent legacy unit tests pass

portante:
The recent PR #2661 (commit 487f4ba) added reporting of left-over files in the Tool Meister's temporary directory.  But if the file-system discovery operation happens to return them in a different order than the unit tests are expecting test 56 and 57 of the agent 'util-scripts' tests will fail.


dbutenhof (approved):
Ah, yes; bitten by the *expected* unexpected files. 😆

npalaska (approved):
Ah! good catch, Thank you for fixing this.

PR #2679, opened 2022-03-11
Add `util/tree-cleanup`

portante:


utils/tree-cleanup
```diff
@@ -0,0 +1,50 @@
...
+    if [[ -z "${_file}" ]]; then
+        continue
+    fi
+    if [[ "${_verbose}" == "yes" || "${_mode}" == "dry-run" ]]; then
+        echo "rm -rf '${_file}'"
+    fi
+    if [[ "${_mode}" == "execute" ]]; then
+        rm -rf ${_file}
+    fi
+done <<< "$(git status --short --ignored | grep -E '^!!')"
```
dbutenhof, line 50: I think using `--porcelain` might be a bit more canonical than `--short`, although (aside from colorization on the latter) there's not much pragmatic difference.

This does present a few potential gotchas, although I'm not sure I know of any good way around them. For example, Webb mentioned the other day that he has a `copr` config file he keeps in his repo so that he can access it inside a container. I have an ignored `.env` file to help `vscode` find the Pbench modules. In both cases, we wouldn't want those deleted, and that pragmatically makes this otherwise intriguing cleanup script dangerous.

However maybe we just have to beware of that, because I don't see any straightforward automated solution. And, regardless of the fact that it won't be useful for me, I think the idea of this mechanism is a good one.

I guess I'll just stick with `find . -name __pycache_ -exec rm -rf {} \+` 😆

portante:
I've decided I am not going to pursue this based on the feedback received.

utils/tree-cleanup
webbnh: I think the kernel of this idea is an excellent one, but, like Dave, I think the gotchas are likely to prevent me from using it as presented; however, perhaps we could move forward with it based on a slightly different approach.

Basing this on Git is probably the wrong foundation -- the `.gitignore` file contains a list of things to avoid committing to the repo, not a list of garbage files -- however, we could build a list of "cleanable" files, and make this tool consume that, instead.  Yes, the two lists are likely to look very similar, and, yes, when we come up with a new piece of garbage we would need to add it to both lists, but I don't think that necessarily makes it a bad approach:  the two lists serve different purposes which _happen_ to have a lot of overlap.  (I _suppose_ we could apply a filter to the `.gitignore` file, but I'm inclined to think that the risk of that going awry outweigh the benefits from the commonality.)

FWIW, I'd be willing to continue [our earlier discussion](https://github.com/distributed-system-analysis/pbench/pull/2551#discussion_r823828853) which got crushed under the weight of other concerns.  I think that having a `clean` target in a top level `Makefile` is not an unreasonable way to access this functionality (although, I think I like your `utils` script approach better).

webbnh: > Webb mentioned the other day that he has a `copr` config file he keeps in his repo

FWIW, I keep it in my check-out...I can't commit it to the repo because it's got my credentials in it...but, it's also not an ignored file, so it wouldn't have been hit by this tool.

PR #2682, opened 2022-03-12
Remove references to logger method of base class

portante:
When the server unit tests run in parallel, there is something about how the Database.base class is modified to have a "logger" that is not working.

It seems a bad idea in general to do this kind of logging in a database class, and we should have the logging at the consumers of the database object.  Given the unit tests are not affected by this change, seems like it is a good opportunity to make this kind of a change.

We also fix a wayward server test.

### DIFF
--- a/lib/pbench/server/database/models/template.py
+++ b/lib/pbench/server/database/models/template.py
@@ -25,12 +25,13 @@ class TemplateSqlError(TemplateError):
     original SQLAlchemy exception.
     """
 
-    def __init__(self, operation: str, name: str):
+    def __init__(self, operation: str, name: str, cause: str):
         self.operation = operation
         self.name = name
+        self.cause = cause
 
     def __str__(self) -> str:
-        return f"Error {self.operation} index {self.name!r}"
+        return f"Error {self.operation} index {self.name!r}: {self.cause}"
 
 
 class TemplateFileMissing(TemplateError):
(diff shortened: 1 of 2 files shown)


npalaska (approved):
This does seem like a good cleaning, but we are still missing the logs from the consumer objects  as you pointed out. Do we want to add `try-except` blocks to the consumer of these methods (so we can log the exceptions) in this PR or should that be another PR?

dbutenhof (approved):
> It seems a bad idea in general to do this kind of logging in a database class, and we should have the logging at the consumers of the database object. Given the unit tests are not affected by this change, seems like it is a good opportunity to make this kind of a change.

I disagree, in general; we should be able to have observability where we need it. While I agree that assigning a logger into the `Base` class after initialization is odd, it's valid Python... and this isn't a unique logging pattern so if this is the only problem you're hitting than it may be a symptom of something else.

We should probably be building a subclass logger ourselves either in each model class or in a common parent class between them and the SQLAlchemy `declarative_base` class, however I guess we can continue to defer dealing with all that.

Since the logging you removed was mostly useful during initial debugging, I guess this is fine.

portante:
The principle is that logging should not be "shoe-horned" into a class, it should be part of the class, and the logging should be performed at a level that is useful.

Both of these principles were violated with this code.

The changes here add all the information that was in the logs to the exceptions, so that a higher-level method which isn't in violation of the principles of "no shoe-horning" and is "useful" can properly emit the logs where they can be consumed at a rate that is usable.

PR #2684, opened 2022-03-15
Remove `ansible` from agent `test-requirements.txt`

portante:


portante:
Layered on top of PR #2649 which needs to land first before we can open for review.

agent/tool-scripts/datalog/mock-lib/ansible/parsing/dataloader/__init__.py
```diff
@@ -0,0 +1,3 @@
+class DataLoader:
+    def __init__(self):
+        self._empty_ = True
```
webbnh, line 3: What reads `_empty_`?

webbnh (commented):
Now that we're using Python3, do we still need the empty `__init__.py` files?  (I thought that that requirement went away with Python2.)

webbnh (approved):
👍

agent/tool-scripts/datalog/mock-lib/ansible/parsing/dataloader/__init__.py
portante: Is this resolved?

agent/tool-scripts/datalog/mock-lib/ansible/inventory/manager/__init__.py
```diff
@@ -0,0 +1,9 @@
+class InventoryManager:
+    def __init__(self, loader, sources):
+        self.loader = loader
+        self.sources = sources
+
+    def get_groups_dict(self):
+        masters = ["masterA.example.com"]
+        nodes = ["pprofA.example.com"]
+        return {"masters": masters, "pprof": nodes}
```
dbutenhof, line 9: That's fine, although I would have just done `return {"masters": ["masterA.example.com"], "pprof": ["pprofA.example.com"]}`, which would clarify that these are hardcoded literals and make it just a bit easier to read.

dbutenhof (approved):
One minor comment, but I'm going to just resolve it regardless of any response.

agent/tool-scripts/datalog/mock-lib/ansible/inventory/manager/__init__.py
webbnh: Me, too...but I figured we would rather spend time on other stuff.  😏

PR #2689, opened 2022-03-16
Rev-lock `pbench-trafficgen` to our fork of the trafficgen repo

### LINKED ISSUE #2223: Rev-lock `pbench-trafficgen` to a particular trafficgen Git commit hash
Until we are able to remove the mechanisms for running benchmark workloads from Pbench (i.e., replacing it with Benchmark-Wrapper or something similar), we need to rev-lock Pbench to the versions of the workloads for which it contains support.

Recent changes to trafficgen have broken Pbench's ability to run it "out of the box", and so we cannot have `pbench-trafficgen` blindly pulling the end of the `master` branch.  Instead, if `pbench-trafficgen` must pull trafficgen, then it should pull a specific commit which is known to work with Pbench; and, otherwise, we should have `pbench-trafficgen` to check if the installed version of trafficgen is compatible.

Otherwise, we risk problems like #2265.

### LINKED ISSUE #2676: `pbench-trafficgen` requires `iommu_intel=on` even on AMD machines
pbench-agent-0.71.0-14g4cbc95439.noarch
pbench-trafficgen commit `47d4fe5575c94b949c4db8c5cf8d6b5c148c56bc` (HEAD -> master, origin/master, origin/HEAD)
trex v2.87
```
pbench-trafficgen --config=Test --devices 81:00.0,81:00.1  --traffic-generator=trex-txrx
[error][2022-03-04T17:33:12.676816505] could not find IOMMU option, intel_iommu=on, in /proc/cmdline, exiting
```
`intel_iommu=on` is completely redundant on AMD machines. AMD requires `iommu=pt` instead. 
This is a cosmetic problem, when adding `iommu_intel=on` and `iommu=pt` to the kernel command line, everything works. It is just a little bit confusing.

webbnh:
Pbench has been having problems with version skew between `pbench-trafficgen` and trafficgen.  Until we can implement a permanent, go-forward solution for running benchmark workloads, we are going to fork trafficgen and freeze the version of it which `pbench-trafficgen` uses.

This change modifies `pbench-trafficgen` to pull trafficgen from the fork in the `distributed-system-analysis` organization and to pull a specific tag, rather than fetching the tip of the main branch.  This will establish a stable version of trafficgen which `pbench-trafficgen` can support.  (Fixes #2223.)

In addition to tweaking how we use Git, this change also changes the IOMMU check from a fatal error to a warning.  (Fixes #2676.)

Note that this PR is still a draft, pending better testing; once that's in place, the last commit from this branch, which contains debugging hacks, will be removed.

### DIFF
--- a/agent/bench-scripts/pbench-trafficgen
+++ b/agent/bench-scripts/pbench-trafficgen
@@ -40,6 +40,8 @@ orig_cmd="${*}"
 
 benchmark_name="trafficgen"
 export benchmark=${benchmark_name}
+trafficgen_repo="https://github.com/distributed-system-analysis/trafficgen.git"
+trafficgen_tag="v0.71"
 
 # General pbench defaults
 export benchmark_run_dir=""
@@ -410,7 +412,8 @@ function usage {
 	printf -- "  running already)\n"
 	printf -- "\n"
 	printf -- "--skip-git-pull\n"
-	printf -- "  Do not call git pull on the trafficgen repo if it already exists\n"
+	printf -- "  By default, this script will maintain a compatible version of trafficgen\n"
+	printf -- "  in ${trafficgen_dir}; use this switch to disable updates.\n"
 	printf -- "\n\n"
 	printf -- "options common in most pbench benchmark scripts\n"
 	printf -- "-----------------------------------------------\n"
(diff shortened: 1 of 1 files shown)


portante (changes requested):
This is looking good, but we have suspended work related to TrafficGen until we can confirm we want to invest in its support.

If we were to continue this work, it would be great to have the `git` commit ID discovered even when we don't issue a `git clone` or `git fetch/checkout`.

portante:
TrafficGen is no longer supported starting in v0.71 and later.

PR #2690, opened 2022-03-16
Converting `test_config.py` functional test to unit test

riya-17:
https://github.com/distributed-system-analysis/pbench/pull/2606#discussion_r803063752  As quoted in 2606 PR comment, 

The test style in use here is a "functional test" model, where we actually invoke a real tool and try to make sure that it functions correctly under various circumstances.

The "unit test" model would be to invoke only the code which we want to test, here that would be `BaseCommand.__init__()`.  We _could_ do that directly, but that's a bit complicated, since it's an object constructor.  Typically, we would test those by instantiating the object, but this is an abstract base class, so we cannot create an instance of it.  So, to address that, we would have the unit test code declare a concrete subclass of it and instantiate that.

The challenge is that the unit test will need to come up with a Click context object to pass to the instantiation...but maybe that's not too hard.  This being Python, I think that all we need is a generic object which has a field named `config` with an appropriate value in it.  So, maybe something like this would work:
```
    class TestCommand(BaseCommand):
        class FakeContext:
            config = "/path/to/config_file"

        def __init__(self):
            super(TestCommand, self).__init__(FakeContext())

    pbench_run_val = TestCommand().pbench_run
```

If that works well, you might consider using it in the other tests to make them more unit-ish and less functinal-ish.  (To make them properly unit-ish, you'd probably want to mock `PbenchAgentConfig.__init__()` and maybe a few other things, but this would be a good start.)

_Originally posted by @webbnh in https://github.com/distributed-system-analysis/pbench/pull/2606#discussion_r803063752_

Fix to be Done

There is one fix yet to be done. Line. 34 `click.get_current_context().exit(1)` in `agent/base.py` need to be mocked and exit gracefully and also gather the `[ERROR] pbench run directory does not exist` from terminal output to assert the error message.


lib/pbench/test/unit/agent/test_config.py
```diff
@@ -34,3 +38,72 @@ def test_results_attr():
...
+        self.pbench_run = "/test/pbench-run"
+        self.pbench_log = "logger"
+        self.pbench_install_dir = Path("/test/install_dir")
+        self.ssh_opts = "/ssh_opts"
+        self.scp_opts = "/scp_opts"
+
+
+def mock_exist(path: Path):
+    """mock method for Path.exist"""
+    return True
```
webbnh, line 59: This is a slightly weak approach.  While, it's good enough for `test_pbench_run_path()`, in general, it would be better if it confirmed that `path` it receives matches the expected value (e.g., the value of `pbench_install_dir`).  (But, I think it's OK as is.)

lib/pbench/test/unit/agent/test_config.py
```diff
@@ -34,3 +38,72 @@ def test_results_attr():
...
+    monkeypatch.setenv("pbench_run", "/test/pbench-run/environ")
+    monkeypatch.setattr(PbenchAgentConfig, "__init__", MockPbenchAgentConfig.__init__)
+    monkeypatch.setattr(Path, "exists", mock_exist)
+    pbench_run_val = TestCommand().pbench_run
+    assert str(pbench_run_val) == "/test/pbench-run/environ"
+
+
+def test_pbench_run_dir_existence(monkeypatch):
+    """Verify that pbench_list_tools fail when pbench_run is defined
+    to a directory that doesn't exist"""
```
webbnh, line 104: This docstring looks like it needs updating (i.e., I don't see `pbench_list_tools` used in this test).

lib/pbench/test/unit/agent/test_config.py
```diff
@@ -34,3 +38,72 @@ def test_results_attr():
...
+    monkeypatch.setattr(PbenchAgentConfig, "__init__", MockPbenchAgentConfig.__init__)
+    monkeypatch.setattr(click, "get_current_context", mock_current_context)
```
webbnh, line 106: I suspect that we want a slightly different approach, here.  We _could_ mock `get_current_context()`, but we would need the mock function to actually return a usable `click.Context` object (although, we could probably mock the object fairly easily).  Instead, I think you can mock just the `exit()` function on that class.

I would recommend having the mock `exit()` function raise a custom exception and have the test catch it.  You can include the exit status as an argument to the exception and check it when it's caught.

lib/pbench/test/unit/agent/test_config.py
```diff
@@ -34,3 +38,72 @@ def test_results_attr():
...
+    pbench_run = TestCommand().pbench_run
+    assert pbench_run
```
webbnh, line 108: This seems like a _very_ weak test -- can't we predict what value it should have and confirm it?

lib/pbench/test/unit/agent/test_config.py
```diff
@@ -34,3 +38,72 @@ def test_results_attr():
...
+    # assert pbench_run.startswith("[ERROR] pbench run directory does not exist")
```
webbnh, line 109: We should be mocking the `exists()` function to return `False` for this test (rather than counting on make a file system access and having it fail).

As for checking the contents of `stderr`, look at the Click [Testing](https://click.palletsprojects.com/en/7.x/api/#testing) support; [here's an example](https://github.com/riya-17/pbench/blob/0b3c2d4029aaccbbcb2b5fc3b855c48591551a25/lib/pbench/test/unit/agent/task/test_generate_token.py#L138) of how we use it.  (I think it will work with your mock-`TestCommand()`.)

Once the `exists()` call returns `False`, the error message will be sent to `stderr`, which the test runner will capture in the `Result` object, and you can reference that in an assertion, once you get control back after catching the exception from the `exit()` mock.

(Let me know if you have any questions or need any more explanation or help with this.)

lib/pbench/test/unit/agent/test_config.py
```diff
@@ -34,3 +38,72 @@ def test_results_attr():
...
+    pbench_run_val = TestCommand().pbench_run
+    assert str(pbench_run_val) == "/test/pbench-run"
+
+
+def test_pbench_run_as_env(monkeypatch):
+    """Use pbench_run value, passed manually through
+    `pbench-run` environment variable"""
+    monkeypatch.setenv("pbench_run", "/test/pbench-run/environ")
+    monkeypatch.setattr(PbenchAgentConfig, "__init__", MockPbenchAgentConfig.__init__)
+    monkeypatch.setattr(Path, "exists", mock_exist)
```
webbnh, line 97: Since all you want to do is return `True` (or, in the case of `test_pbench_run_dir_existence()`, `False`), you might want to consider using a lambda here instead of referencing a standalone function, since the expression is so simple:
```
    monkeypatch.setattr(Path, "exists", lambda path: True)
```
This saves the reader from having to refer elsewhere to determine what the mock does (and, it lets you choose locally whether to return `True` or `False`).

webbnh (changes requested):
It looks like a good start.

Hopefully, I've explained what you need to do to address the test failures.

portante (changes requested):
First, the commit needs a proper commit message to describe what is being changed, and why.

Is there an issue for this work?  Doesn't have to be, but it would be nice to know where this originates.

Second, the original functional CLI test invoked `pbench-list-tools`, but the test was named, `test_config`.  Is this PR trying to fix that?

Third, do we know what the original three tests were doing exactly?

From what I can tell:

 * the first test, `test_pbench_run`, was relying on the behavior of `pbench-list-tools` to somehow execute successfully, but no attempt was made to verify its behavior.
 * the second test, `test_pbench_run_dir_existence`, used `pbench-list-tools` with `pbench_run` defined to a non-existent directory and expected `pbench-list-tools` to report an error; that seems like a test of `pbench-list-tools` behavior, and not directly related to "config" testing
 * the third test, `test_pbench_dir_existence`, used `pbench-list-tools` to emit output about registered tools without mentioning anything about the `pbench_run` environment variable set to an existing directory

All three of those tests are really testing aspects of the behavior of `pbench-list-tools`, and that behavior is incorrect, as `pbench-list-tools` should not care about the definition of the `pbench_run` directory.

The new tests need to be testing the behavior of the pbench code for the agent configuration, but I am not following how this is being done.  It seems like we are trying to test the behavior of the PbenchAgentConfig class.

I think we want to approach the config testing by:

 * First test the `PbenchConfig` base class by mocking the behavior of `configtools` and `ConfigParser`, by checking fields are created as expected, and a `BadConfig` exception is raised when `hostport` logger type does not have the `logger_host` and/or `logger_port` exceptions
 * Next test the `PbenchAgentConfig` class and its behaviors for field values, and exceptions raised

And we should file an issue to stop having the `PbenchAgentConfig` constructor from creating the `pbench_tmp` directory.

riya-17:
> First, the commit needs a proper commit message to describe what is being changed, and why.
> 
> Is there an issue for this work? Doesn't have to be, but it would be nice to know where this originates.
> 
Sure I have added that, where it originated from and why

> Second, the original functional CLI test invoked `pbench-list-tools`, but the test was named, `test_config`. Is this PR trying to fix that?
> 
No, it is not trying to fix any CLI test. it is just trying to test whether the `pbench_run` value that is being propagated is right. 
> Third, do we know what the original three tests were doing exactly?
> 
1st was trying to check whether the `pbench_run` value being taken from config file and no environment variable for `pbench_run` is present.
2nd is trying to confirm that is an environment value is passed than that is being used as `pbench_run` value not the one present in config_file.py
3rd is checking whether the `pbench_run` path that is being passed exists or not.
> From what I can tell:
> 
> * the first test, `test_pbench_run`, was relying on the behavior of `pbench-list-tools` to somehow execute successfully, but no attempt was made to verify its behavior.
> * the second test, `test_pbench_run_dir_existence`, used `pbench-list-tools` with `pbench_run` defined to a non-existent directory and expected `pbench-list-tools` to report an error; that seems like a test of `pbench-list-tools` behavior, and not directly related to "config" testing
> * the third test, `test_pbench_dir_existence`, used `pbench-list-tools` to emit output about registered tools without mentioning anything about the `pbench_run` environment variable set to an existing directory
> 
> All three of those tests are really testing aspects of the behavior of `pbench-list-tools`, and that behavior is incorrect, as `pbench-list-tools` should not care about the definition of the `pbench_run` directory.
> 
`pbench_list_tools` and `pbench_register_tool` were only taken as a medium to test `pbench_run` value in functinal tests.
> The new tests need to be testing the behavior of the pbench code for the agent configuration, but I am not following how this is being done. It seems like we are trying to test the behavior of the PbenchAgentConfig class.
> 
> I think we want to approach the config testing by:
> 
> * First test the `PbenchConfig` base class by mocking the behavior of `configtools` and `ConfigParser`, by checking fields are created as expected, and a `BadConfig` exception is raised when `hostport` logger type does not have the `logger_host` and/or `logger_port` exceptions
> * Next test the `PbenchAgentConfig` class and its behaviors for field values, and exceptions raised
> 
> And we should file an issue to stop having the `PbenchAgentConfig` constructor from creating the `pbench_tmp` directory.

No, I am not aiming to test the PbenchConfig

lib/pbench/test/unit/agent/test_config.py
```diff
@@ -34,3 +38,72 @@ def test_results_attr():
...
+    None
+
+
+class MockContext:
+    """Mocking Context"""
+
+    config = "/test/pbench-config-path"
+
+
+class TestCommand(BaseCommand):
```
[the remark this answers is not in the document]
riya-17: yes, sure

lib/pbench/test/unit/agent/test_config.py
riya-17: yes, I can understand, I was facing some issues here. I also tried to mock the exit(), which module does the `exit()` belong to? Is it `sys`? I tried with `click`& I also tried with `sys` but none of them is working.

lib/pbench/test/unit/agent/test_config.py
riya-17: yes, It was still under work. Sorry, I should have mentioned it in the test itself.

lib/pbench/test/unit/agent/test_config.py
riya-17: Thank you this is what I was looking for

lib/pbench/test/unit/agent/test_config.py
dbutenhof: Yeah, I had similar thoughts. The scope of the test is limited, but but the methodology bothered me anyway and I ended up undecided about commenting. (Which is why I hadn't posted comments yesterday!)

dbutenhof (commented):
Peter has a great point that this really *is* testing the behavior of the `PbenchAgentConfig` class with respect to `pbench_run`, and testing through CLI commands somewhat muddies the focus. The test path (`unit/agent/test_config.py`) is pretty much what it does (we could add additional `PbenchAgentConfig` unit test cases here), and as a *unit* test there's no obvious value (or point) in dragging in the click commands at all since the desired artifacts are observable in the `PbenchAgentConfig` instance itself.

lib/pbench/test/unit/agent/test_config.py
webbnh: Thinking about this this morning, if you use the Click test runner, then you probably won't need to mock `exit()`.  So, I would try using the test runner first and defer mocking `exit()` until it proves to be needed.

lib/pbench/test/unit/agent/test_config.py
webbnh: _If_ you need to mock it (which I don't think you do, if you use the Click test runner), it would be something close to
```
    monkeypatch.setattr(click.Context, "exit", lambda s: raise MyExitException(s))
```
where `MyExitException` is a class that you define which captures the exit status and makes it available to the code which catches the exception.

riya-17:
Closing in Favor of a new One, Mainly around Pbench Agent Test Coverage.

PR #2693, opened 2022-03-17
Use an explicit cache directory for pip installs

portante:
This does not help the Jenkins CI job much, since we are using an ephermal mount for the virtualenvs.  It does help a little bit for local development environments.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -7,7 +7,7 @@ envlist = py
 
 [testenv]
 description = Lints code and runs all agent and server unit/functional tests
-install_command = pip install --no-cache-dir --progress-bar off --prefix={envdir} {opts} {packages}
+install_command = pip install --cache-dir={toxworkdir}/cache --progress-bar off --prefix={envdir} {opts} {packages}
 passenv =
     PY_COLORS
     NO_COLORS


tox.ini
```diff
@@ -7,7 +7,7 @@ envlist = py
 
 [testenv]
 description = Lints code and runs all agent and server unit/functional tests
-install_command = pip install --no-cache-dir --progress-bar off --prefix={envdir} {opts} {packages}
+install_command = pip install --cache-dir={toxworkdir}/cache --progress-bar off --prefix={envdir} {opts} {packages}
```
dbutenhof, line 10: Hmm; so we're using two separate cache directories: `/var/tmp/USER/cache` and `/var/tmp/USER/tox/cache`. Does this matter? I think I've lost track of the contexts. At worst, we're duplicating some cache and it's still probably better than before (and pragmatically no different for Jenkins).

dbutenhof (approved):

webbnh (approved):
What Dave said.

tox.ini
portante: The cache directory use in `jenkins/python-setup.sh` is really of no use because it is for an ephemeral drive.  So best to just remove it.

portante (commented):
Okay, let's try this simpler version.

tox.ini
webbnh: How much trouble would we get into if we mapped the cache directory from a persistent volume?  (How good is pip's cache management -- would it be able to determine when an entry is stale or bad and discard it?)

webbnh (approved):
👍

tox.ini
portante: It is supposed to be able to do that.  The problem is where to put the volume.  I'd rather approach that in a separate effort since it will involve environmental factors.  We might want to look at a cache that works for all Jenkins instances as well.

dbutenhof (approved):

PR #2694, opened 2022-03-17
Drop use of `--current-env` in CI jobs

portante:
We now only use `tox` created virtualenvs for running our tests.  This means that CI jobs will pay the cost of setting up a full virtualenv with all the dependent modules installed each time tests are run.

This means we can remove the external environment leakage the agent util-scripts tests create to work properly.


dbutenhof (commented):
Wow; no more `python-setup.sh` ... that's a biggie.

But you've been stopped by the psycopg:

> Error: pg_config executable not found.

May not be anything directly related to your changes; this happens to me every once in a while, and then goes away all by itself after repeating the run (albeit sometimes requiring multiple retries and environment reset...)

jenkins/run-unittests
```diff
@@ -1,10 +1,3 @@
 #!/bin/bash
 
-# For the CI jobs, we run unit tests expecting most of the environment to be
-# provided by the container image, and only a few modules not installable via
-# RPM need to be added (see jenkins/python-setup.sh).  We then instruct tox to
-# use the current environment (not setting up a virtualenv), and to run the
-# tests using whatever version of the python interpreter the install provides
-# (the `py` directs Tox to use whatever it finds).
```
webbnh, line 8: While I grant that the old comment is no longer needed, I think we still need a comment here:  there is a reason that we wrap this command in a script, and we need to point that out, so that no one like me is tempted to try to move this command into the pipeline file and get rid of this "silly little script".

webbnh (changes requested):
We need to explain why `jenkins/run-unittests` exists, and I'm a little concerned about the (lack of) `PATH` setup, but otherwise I'm OK with this.

webbnh:
> Error: pg_config executable not found.

IIRC, there are two ways to acquire `psycopg2`:

1. Pip-install the `psycopg2` package, which pulls the source and attempts to build it during the installation, which requires a compiler tool-chain, etc.
2. Install the `psycopg2-binary` package, which comes pre-built.

The problem was, we _wanted_ to install it via RPM and then have the pip-install decide that it was already present.  In order to make that happen, we had to _require_ the `psycopg2` package, so that it would match up with with the RPM provided; however, if the match failed, then what we _needed_ to require was `psycopg2-binary` (but, requiring that _never_ matched with any RPM, so that didn't suit our purposes).

Now that our "requirements" are different, we presumably need to change our `requirements.txt` file.

portante:
Perhaps a clue: `libpg-devel` provides `pg_config`, so by rebuilding the development image with that package installed might allow the CI job to pass without needing `--current-env`.

jenkins/run-unittests
portante: No need for this script anymore.

portante (commented):
This is now working.  I'll file a commit shortly to remove the no longer needed `jenkins/run-unittests`.

portante:
This does increase the time the CI jobs take, but we can reduce that time but tackling the slowness in the individual unit tests.

jenkins/python-setup.sh
```diff
@@ -1,40 +0,0 @@
...
-
-function _pbench_pathins() {
-    if [[ ":${PATH}:" != *":${1}:"* ]]; then
-        PATH="${1}${PATH:+:${PATH}}"
-    fi
-}
-
-_pbench_pathins "/usr/sbin"
-_pbench_pathins "/usr/bin"
-_pbench_pathins "$(python3 -m site --user-base)/bin"
```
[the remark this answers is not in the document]
portante: All the tests pass, wasn't necessary apparently.

jenkins/branch.name
```diff
@@ -1 +1 @@
-main
+no-current-env
```
webbnh, line 1: I imagine that you don't want this change merged...is that correct?

jenkins/python-setup.sh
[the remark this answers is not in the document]
webbnh: 😬

jenkins/run-unittests
webbnh: Because it has been superseded and replaced by `jenkins/runner`?...

jenkins/runner
```diff
@@ -0,0 +1,12 @@
...
+# This script exists to ensure that `bash` is the process 1 in the container
+# executed by jenkins/run.  If `bash` is not process 1, then any zombie
+# processes won't be reaped until the program running as process 1 explicitly
+# reaps zombie processes.
+
+if [[ ${#} -ne 1 ]]; then
+    printf -- "${0}: Unexpected number of arguments: ${#}\n" >&2
+    exit 1
+fi
+eval "${1}"
```
webbnh, line 12: Why is this script restricted to a single-element command?  (Why not just `eval "${*}"` or the equivalent?)  This greatly limits what we can do with `jenkins/run`!

With this change, I think I can still do things like `jenkins/run bash`, but I won't be able to do things like `jenkins/run make srpm`.  😢

webbnh (changes requested):
Can we have the restriction on the `jenkins/run` command line lifted?

Also, I think this PR contains a change that you don't actually want merged.

jenkins/branch.name
portante: Correct, I didn't want to pollute the current image on quay.io.  I have added a "DO NOT MERGE" label.

jenkins/runner
portante: Because it is married to `jenkins/run` which passes all its arguments as 1 argument.

jenkins/branch.name
portante: And I have pulled the change out into a separate commit to make it easy to isolate and remove.

jenkins/runner
webbnh: So, the intention is that the entire command line will be passed as a single (quoted) argument, this script counts on that, and so it performs the check (which, under those circumstances, is a Good Thing).

However, wouldn't it work just as well to use `"${*}"` in the `eval` statement?  Then we wouldn't need to do the check, and there would be no confusion.

jenkins/run
```diff
@@ -1,20 +1,27 @@
...
-# initialize a Pbench testing environment.
+# Run commands within the pbench-${IMAGE_ROLE}-${IMAGE_KIND} container image.
 #
 # Several customizing inputs can be given via environment variables:
 #
 # EXTRA_PODMAN_SWITCHES []
 # IMAGE_REPO [quay.io/pbench]
+# IMAGE_ROLE [devel] (see Makefile for options)
 # IMAGE_KIND [fedora] (see Makefile for options)
+# IMAGE [image spec you want, overrides all of the above]
```
webbnh, line 11: This is starting to look like a general-purpose tool.  Should it be moved out of the `jenkins` directory and into, e.g., `utils`?

jenkins/run
```diff
@@ -27,13 +34,9 @@ fi
...
-if [[ -z "${IMAGE_REPO}" ]]; then
-    IMAGE_REPO="quay.io/pbench"
-fi
-
-if [[ -z "${IMAGE_KIND}" ]]; then
-    IMAGE_KIND="fedora"
-fi
+_image_repo=${IMAGE_REPO:-quay.io/pbench}
+_image_role=${IMAGE_ROLE:-devel}
+_image_kind=${IMAGE_KIND:-fedora}
```
webbnh, line 39: I wish we could back away from this practice of putting leading underscores on everything.  Yes, there are situations in which doing so is warranted, but I don't think that shell variables at the top level of a script are one of them, and thus they are just  unproductive visual clutter.

jenkins/Pipeline.gy
```diff
@@ -1,7 +1,8 @@
 pipeline {
     agent { label 'pbench' }
     environment {
-        EXTRA_PODMAN_SWITCHES="-e COV_REPORT_XML=yes"
+        IMAGE_ROLE="ci"
```
webbnh, line 4: Do we really want to let `IMAGE_KIND` (and `IMAGE_REPO`) fall back to the default?

webbnh (approved):
Looks like some good stuff.  I just have a bunch of small comments.

jenkins/Makefile
```diff
@@ -3,7 +3,13 @@
 
 BRANCH := $(shell cat ./branch.name)
```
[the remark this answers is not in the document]
portante: Two different files use it, `jenkins/Makefile` and `jenkins/run`.  So having a file seems reasonable, and ... it allowed me to make an isolated change when testing to ensure the PR does not disrupt `main`.

jenkins/development.Dockerfile.j2
```diff
@@ -20,7 +20,7 @@
 # You can use this image to run our development environment from a command
 # line, e.g.:
 #
-#  $ podman run -it --rm pbench-devel:<branch name> /bin/bash
+#  $ podman run -it --rm pbench-devel-*:<branch name> /bin/bash
```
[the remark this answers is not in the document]
portante: I was trying to escape making further changes while not polluting the know of the names available since that is entirely separate.

jenkins/development.Dockerfile.j2
```diff
@@ -110,6 +109,7 @@ RUN \
         `# file and this list should be regenerated.` \
         `#` \
         cronie \
+        libpq-devel \
```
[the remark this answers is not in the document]
portante: Something else pulls it in.

Neither file lists the full dependency tree, since each RPM sub-dependency contributes as well.

jenkins/run
portante: The typical use I have encountered is to separate internal from external.  Does this pose a serious support problem?

jenkins/Pipeline.gy
portante: For the CI job there are no other alternatives, so defaults are fine.

jenkins/runner
portante: I think I have a way to make this a bit better.  Will post shortly.

jenkins/Pipeline.gy
portante: I figured it was best not to duplicate anything since all of this IS for the CI jobs.  Since you asked, we now have them specified in two places.  =)

jenkins/run
webbnh: In this context, there is no "external" (or, for that matter, "internal").

The underscore adds a useless and unneeded character to each name; when several of these names appear together on the line, it uselessly extends the line.  Underscores are normally used on the interior of symbols, so adding them on the exterior makes the symbol harder to parse, visually.  Shell scripts already suppress most use of horizontal space, so adding extra underscores compounds the problem.

In short, they make the code harder to read.

For cases where we want to make it explicit that a variable has limited scope (e.g., if it is declared `local` inside a defined shell function), the cost is potentially worthwhile.  But, these variables have normal scope, so there is nothing to be gained by trying to claim that they are "local".  And, they lead to this (which doesn't fit on an 80-column line):
```
_image=${IMAGE:-${_image_repo}/pbench-${_image_role}-${_image_kind}:${_branch_name}}
```
which should look more like this:
```
image=${IMAGE:-${image_repo}/pbench-${image_role}-${image_kind}:${branch_name}}
```

jenkins/development.Dockerfile.j2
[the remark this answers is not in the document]
webbnh: We're including `libpq-devel` here because we know that the pip-install of `psycopg2` will need it, so it seems like we ought to be (explicitly) listing `gcc` here as well, since we know that the pip-install will need that, too.  The fact that something else (implicitly) pulls it in for us is fine...as long as we keep pulling that thing in and as long as its behavior doesn't change.  I mean, we can go with the "it seems to work" definition of success, but it seems like it would be better for us to state our requirements (to the extent that we know them...) and let `dnf` decide whether they are already satisfied.

jenkins/Makefile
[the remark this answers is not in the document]
webbnh: Apparently, there's a _third_ usage:  `server/pbenchinacan/container-build.sh`  😀

In any case, I appreciate what this mechanism enables; the part that is unreasonable is that it forces you to mark the PR as "DO NOT MERGE".  I don't really want to be reviewing a submission which cannot be "accepted".  And, it's annoying to know that, once the PR gets all of its approvals, the branch will have to be modified (and probably re-approved) before it can be merged.  So, I'm wondering if there isn't an alternative approach.

webbnh (dismissed):
👍

jenkins/Pipeline.gy
webbnh: One of the places is a "request"; the other is a "default" used in the absence of a request -- I think they are not duplicates because they aren't definitions.  (E.g., you could delete either one, and, barring changes elsewhere, things would continue to work.)

I think it's good to have defaults, because, when this stuff is used interactively (by humans), well-chosen defaults make life easier.  But, in the case of the CI, it doesn't need to depend on the defaults, and it costs very little to specify the values explicitly, so there's very little reason not to do so.

Before you went and made things so nicely general, there was only one choice, so we didn't need things like "requests" and "defaults".  But, now that we have options, I think the CI should ask for what it wants, and we can leave the defaults for people with fingers who have to type enough as it is.  :wink:

dbutenhof (approved):

webbnh (approved):
👍

portante (commented):
Rebased to remove the `branch.name` file change so that I could rebuild the images and push to quay.io (both CI and develop).  This will allow the CI job to run as it will on `main` before we merge.

dbutenhof (approved):

webbnh (approved):
👍

PR #2695, opened 2022-03-17
Remove refs to non-existent persistent tool files

portante:
With commit 487f4baf1 (PR #2661), the persistent tool sub-process no longer uses on-disk tool output files.

Since those output files are no longer created, we have to remove the remaining references to them when logging a warning message regarding failed persistent tools.  Without this change, the Tool Meister would crash with an exception being raised for missing files.

### DIFF
--- a/lib/pbench/agent/tool_meister.py
+++ b/lib/pbench/agent/tool_meister.py
@@ -424,8 +424,7 @@ class PersistentTool(Tool):
         self.process = None
 
     def log_subprocess_output(self, pipe: subprocess.PIPE):
-        """
-        Utility function to log outputs from a given pipe.
+        """Thread start function to log outputs from a given pipe.
         """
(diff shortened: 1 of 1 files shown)


dbutenhof (approved):
I think the PR title could be tweaked.

npalaska (approved):
Ah! Yes we dont need to write log files while we wait for the persistent tools to exit. I guess I didnt encounter this error before.

webbnh (approved):
And I think it would be good to have a description for the PR, as well.  🙂 
(E.g., what are the implications of this change?  We're just not dumping the outputs into the log?  Does this also prevent a hang?)

portante (commented):
Please re-review, thanks!

npalaska (approved):

dbutenhof (approved):

webbnh (approved):
👍

Issue #2692, opened 2022-03-16
Tool Meister should retry tar ball creation on first failure ignoring warnings


portante:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

If the first tar ball fails on creation, the Tool Meister issues a log message indicating the error, but does not complete the protocol with the Tool Data Sink, leaving the Tool Data Sync to hang waiting for ever.

### To Reproduce

There is no known way to make this happen in a reproducible fashion.

### Actual Results

```
192.168.122.192 0280 2022-03-14 21:38:02,387 ERROR 26848 140432394401600 pbench-tool-meister _send_directory 1307 -- Failed to create tar ball; return code: 1
192.168.122.192 0281 2022-03-14 21:38:02,389 WARNING 26848 140432394401600 pbench-tool-meister driver 1721 -- 1 failures encountered for action, <bound method ToolMeister.send_tools of <pbench.agent.tool_meister.ToolMeister object at 0x7fb8f5ac8dc0>>, on data, {'action': 'send', 'args': None, 'directory': '/var/lib/pbench-agent/fio__2022.03.15>
```

### Expected Results

Tool Meister properly sends a tar ball completing the protocol sequence.

### Additional information

Ideally, the Tool Data Sync should timeout if it doesn't hear from a Tool Meister.

The Tool Meister should be able to re-try the `tar` ball creation ignoring all warnings to gather whatever data it can.

The output of the tar command should be captured and logged remotely.

We could consider having all Transient tool `stderr` logged remotely.

portante:
Here are the contents of a Tool Meister work directory where we can see all the `tm-*.out` and `tm-*.err` files which arguably should be just captured and sent remotely instead of living in files.

```
tm.default.26848.ul9har60/192.168.122.192/tm-iostat-start.out
tm.default.26848.ul9har60/192.168.122.192/tm-iostat-start.err
tm.default.26848.ul9har60/192.168.122.192/tm-mpstat-start.out
tm.default.26848.ul9har60/192.168.122.192/tm-mpstat-start.err
tm.default.26848.ul9har60/192.168.122.192/tm-proc-interrupts-start.out
tm.default.26848.ul9har60/192.168.122.192/tm-proc-interrupts-start.err
tm.default.26848.ul9har60/192.168.122.192/tm-proc-vmstat-start.out
tm.default.26848.ul9har60/192.168.122.192/tm-proc-vmstat-start.err
[...]

PR #2696, opened 2022-03-17
Correct Tool Meister sub-process termination

portante:
If a sub-process that the Tool Meister creates does not terminate correctly, it can cause problems.  One such issue, described in issue #2692, was the result of the `subprocess.kill()` method working but not taking effect immediately.

The algorithm used now is:

 1. `subprocess.terminate()`
 2. wait with timeout
 3. `subprocess.kill()`
 4. wait with timeout
 5. close process file descriptors

As part of this work, and to help facilitate the proper closing of the sub-process file descriptors, we setup read threads for logging all the stout/stderr data coming from the sub-proceses.

Partial fix for #2691.

### DIFF
--- a/agent/util-scripts/gold/test-client-tool-meister/test-53.txt
+++ b/agent/util-scripts/gold/test-client-tool-meister/test-53.txt
@@ -197,25 +197,25 @@ INFO pbench-tool-meister __exit__ -- testhost.example.com: terminating
 --- mock-run/tm/tm-default-testhost.example.com.out file contents
 +++ mock-run/tm/tm.logs file contents
 pbench-tool-meister-start - verify logging channel up
-testhost.example.com 0000 INFO pbench-tool-meister sysinfo -- pbench-sysinfo-dump -- /var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/pbench-sysinfo-dump /var/tmp/pbench-test-utils/pbench/mock-run/sysinfo/beg/testhost.example.com block,security_mitigations,sos parallel
-testhost.example.com 0001 INFO pbench-tool-meister sysinfo -- testhost.example.com: sysinfo send (no-op) default /var/tmp/pbench-test-utils/pbench/mock-run/sysinfo/beg/testhost.example.com
-testhost.example.com 0002 INFO pbench-tool-meister start -- Started persistent tool dcgm, ['/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/dcgm-exporter']
-testhost.example.com 0003 INFO pbench-tool-meister start -- mpstat: start_tool -- /var/tmp/pbench-test-utils/opt/pbench-agent/tool-scripts/mpstat --start --dir=/var/tmp/pbench-test-utils/pbench/mock-run/0-iter-zero/sample42/tools-default/testhost.example.com --interval=42 --options=forty-two
-testhost.example.com 0004 INFO pbench-tool-meister stop -- mpstat: stop_tool -- /var/tmp/pbench-test-utils/opt/pbench-agent/tool-scripts/mpstat --stop --dir=/var/tmp/pbench-test-utils/pbench/mock-run/0-iter-zero/sample42/tools-default/testhost.example.com --interval=42 --options=forty-two
-testhost.example.com 0005 INFO pbench-tool-meister wait -- Waiting for transient tool mpstat stop process
-testhost.example.com 0006 INFO pbench-tool-meister wait -- Waiting for transient tool mpstat start process
-testhost.example.com 0007 INFO pbench-tool-meister send_tools -- testhost.example.com: send_tools (no-op) default /var/tmp/pbench-test-utils/pbench/mock-run/0-iter-zero/sample42/tools-default/testhost.example.com
-testhost.example.com 0008 INFO pbench-tool-meister start -- mpstat: start_tool -- /var/tmp/pbench-test-utils/opt/pbench-agent/tool-scripts/mpstat --start --dir=/var/tmp/pbench-test-utils/pbench/mock-run/1-iter-one/sample42/tools-default/testhost.example.com --interval=42 --options=forty-two
-testhost.example.com 0009 INFO pbench-tool-meister stop -- mpstat: stop_tool -- /var/tmp/pbench-test-utils/opt/pbench-agent/tool-scripts/mpstat --stop --dir=/var/tmp/pbench-test-utils/pbench/mock-run/1-iter-one/sample42/tools-default/testhost.example.com --interval=42 --options=forty-two
-testhost.example.com 0010 INFO pbench-tool-meister wait -- Waiting for transient tool mpstat stop process
-testhost.example.com 0011 INFO pbench-tool-meister wait -- Waiting for transient tool mpstat start process
-testhost.example.com 0012 INFO pbench-tool-meister send_tools -- testhost.example.com: send_tools (no-op) default /var/tmp/pbench-test-utils/pbench/mock-run/1-iter-one/sample42/tools-default/testhost.example.com
-testhost.example.com 0013 INFO pbench-tool-meister stop -- Terminate issued for persistent tool dcgm
-testhost.example.com 0014 INFO pbench-tool-meister wait -- Stopped persistent tool dcgm
-testhost.example.com 0015 WARNING pbench-tool-meister end_tools -- testhost.example.com: unexpected temp files dcgm/dcgm-exporter.file
-testhost.example.com 0016 INFO pbench-tool-meister sysinfo -- pbench-sysinfo-dump -- /var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/pbench-sysinfo-dump /var/tmp/pbench-test-utils/pbench/mock-run/sysinfo/end/testhost.example.com block,security_mitigations,sos parallel
-testhost.example.com 0017 INFO pbench-tool-meister sysinfo -- testhost.example.com: sysinfo send (no-op) default /var/tmp/pbench-test-utils/pbench/mock-run/sysinfo/end/testhost.example.com
-testhost.example.com 0018 INFO pbench-tool-meister __exit__ -- testhost.example.com: terminating
+testhost.example.com INFO pbench-tool-meister sysinfo -- pbench-sysinfo-dump -- /var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/pbench-sysinfo-dump /var/tmp/pbench-test-utils/pbench/mock-run/sysinfo/beg/testhost.example.com block,security_mitigations,sos parallel
+testhost.example.com INFO pbench-tool-meister sysinfo -- testhost.example.com: sysinfo send (no-op) default /var/tmp/pbench-test-utils/pbench/mock-run/sysinfo/beg/testhost.example.com
+testhost.example.com INFO pbench-tool-meister start -- Started persistent tool dcgm, ['/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/dcgm-exporter']
+testhost.example.com INFO pbench-tool-meister start -- mpstat: start_tool -- /var/tmp/pbench-test-utils/opt/pbench-agent/tool-scripts/mpstat --start --dir=/var/tmp/pbench-test-utils/pbench/mock-run/0-iter-zero/sample42/tools-default/testhost.example.com --interval=42 --options=forty-two
+testhost.example.com INFO pbench-tool-meister stop -- mpstat: stop_tool -- /var/tmp/pbench-test-utils/opt/pbench-agent/tool-scripts/mpstat --stop --dir=/var/tmp/pbench-test-utils/pbench/mock-run/0-iter-zero/sample42/tools-default/testhost.example.com --interval=42 --options=forty-two
+testhost.example.com INFO pbench-tool-meister wait -- Waiting for transient tool mpstat stop process
+testhost.example.com INFO pbench-tool-meister wait -- Waiting for transient tool mpstat start process
+testhost.example.com INFO pbench-tool-meister send_tools -- testhost.example.com: send_tools (no-op) default /var/tmp/pbench-test-utils/pbench/mock-run/0-iter-zero/sample42/tools-default/testhost.example.com
+testhost.example.com INFO pbench-tool-meister start -- mpstat: start_tool -- /var/tmp/pbench-test-utils/opt/pbench-agent/tool-scripts/mpstat --start --dir=/var/tmp/pbench-test-utils/pbench/mock-run/1-iter-one/sample42/tools-default/testhost.example.com --interval=42 --options=forty-two
+testhost.example.com INFO pbench-tool-meister stop -- mpstat: stop_tool -- /var/tmp/pbench-test-utils/opt/pbench-agent/tool-scripts/mpstat --stop --dir=/var/tmp/pbench-test-utils/pbench/mock-run/1-iter-one/sample42/tools-default/testhost.example.com --interval=42 --options=forty-two
+testhost.example.com INFO pbench-tool-meister wait -- Waiting for transient tool mpstat stop process
+testhost.example.com INFO pbench-tool-meister wait -- Waiting for transient tool mpstat start process
+testhost.example.com INFO pbench-tool-meister send_tools -- testhost.example.com: send_tools (no-op) default /var/tmp/pbench-test-utils/pbench/mock-run/1-iter-one/sample42/tools-default/testhost.example.com
+testhost.example.com INFO pbench-tool-meister stop -- Terminate issued for persistent tool dcgm
+testhost.example.com INFO pbench-tool-meister wait -- Stopped persistent tool dcgm
+testhost.example.com WARNING pbench-tool-meister end_tools -- testhost.example.com: unexpected temp files dcgm/dcgm-exporter.file
+testhost.example.com INFO pbench-tool-meister sysinfo -- pbench-sysinfo-dump -- /var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/pbench-sysinfo-dump /var/tmp/pbench-test-utils/pbench/mock-run/sysinfo/end/testhost.example.com block,security_mitigations,sos parallel
+testhost.example.com INFO pbench-tool-meister sysinfo -- testhost.example.com: sysinfo send (no-op) default /var/tmp/pbench-test-utils/pbench/mock-run/sysinfo/end/testhost.example.com
+testhost.example.com INFO pbench-tool-meister __exit__ -- testhost.example.com: terminating
 --- mock-run/tm/tm.logs file contents
 +++ tools-default/prometheus/prom.log file contents
 --- tools-default/prometheus/prom.log file contents
--- a/agent/util-scripts/gold/test-client-tool-meister/test-53.txt
+++ b/agent/util-scripts/gold/test-client-tool-meister/test-53.txt
@@ -18,10 +18,6 @@ Collecting system information
 /var/tmp/pbench-test-utils/pbench/mock-run/0-iter-zero/sample42/tools-default/testhost.example.com/mpstat/mpstat-stop-postprocess.out
 /var/tmp/pbench-test-utils/pbench/mock-run/0-iter-zero/sample42/tools-default/testhost.example.com/mpstat/mpstat.cmd
 /var/tmp/pbench-test-utils/pbench/mock-run/0-iter-zero/sample42/tools-default/testhost.example.com/mpstat/online-cpus.txt
-/var/tmp/pbench-test-utils/pbench/mock-run/0-iter-zero/sample42/tools-default/testhost.example.com/tm-mpstat-start.err
-/var/tmp/pbench-test-utils/pbench/mock-run/0-iter-zero/sample42/tools-default/testhost.example.com/tm-mpstat-start.out
-/var/tmp/pbench-test-utils/pbench/mock-run/0-iter-zero/sample42/tools-default/testhost.example.com/tm-mpstat-stop.err
-/var/tmp/pbench-test-utils/pbench/mock-run/0-iter-zero/sample42/tools-default/testhost.example.com/tm-mpstat-stop.out
 /var/tmp/pbench-test-utils/pbench/mock-run/1-iter-one
 /var/tmp/pbench-test-utils/pbench/mock-run/1-iter-one/sample42
 /var/tmp/pbench-test-utils/pbench/mock-run/1-iter-one/sample42/tools-default
@@ -33,10 +29,6 @@ Collecting system information
 /var/tmp/pbench-test-utils/pbench/mock-run/1-iter-one/sample42/tools-default/testhost.example.com/mpstat/mpstat-stop-postprocess.out
 /var/tmp/pbench-test-utils/pbench/mock-run/1-iter-one/sample42/tools-default/testhost.example.com/mpstat/mpstat.cmd
 /var/tmp/pbench-test-utils/pbench/mock-run/1-iter-one/sample42/tools-default/testhost.example.com/mpstat/online-cpus.txt
-/var/tmp/pbench-test-utils/pbench/mock-run/1-iter-one/sample42/tools-default/testhost.example.com/tm-mpstat-start.err
-/var/tmp/pbench-test-utils/pbench/mock-run/1-iter-one/sample42/tools-default/testhost.example.com/tm-mpstat-start.out
-/var/tmp/pbench-test-utils/pbench/mock-run/1-iter-one/sample42/tools-default/testhost.example.com/tm-mpstat-stop.err
-/var/tmp/pbench-test-utils/pbench/mock-run/1-iter-one/sample42/tools-default/testhost.example.com/tm-mpstat-stop.out
 /var/tmp/pbench-test-utils/pbench/mock-run/metadata.log
 /var/tmp/pbench-test-utils/pbench/mock-run/ssh_config
 /var/tmp/pbench-test-utils/pbench/mock-run/ssh_config.d
@@ -132,7 +124,7 @@ mpstat = --interval=42 --options=forty-two
 [tools/testhost.example.com/dcgm]
 options = --inst=/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts
 install_check_status_code = 0
-install_check_output = dcgm tool properly installed
+install_check_output = dcgm tool (dcgm-exporter) properly installed
 
 [tools/testhost.example.com/mpstat]
 options = --interval=42 --options=forty-two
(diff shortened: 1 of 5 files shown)


lib/pbench/agent/tool_meister.py
```diff
@@ -110,20 +127,112 @@ def __init__(
...
+                if count > 0:
+                    self.logger.debug(
+                        "%s tool %s has not stopped after %d seconds",
+                        self._tool_type,
+                        self.name,
+                        5 * (6 - count),
+                    )
+        return sts
+
+    def _wait_for_process_w_kill(self, process, ctx_name):
```
webbnh, line 182: Type hints?  (And, since you go to the trouble of including an explicit `return` statement and call it out in the docstring, the type hint should probably include `-> None`.)

lib/pbench/agent/tool_meister.py
```diff
@@ -110,20 +127,112 @@ def __init__(
...
+            "Waiting for %s tool %s %s process", self._tool_type, self.name, ctx_name
+        )
+        # We wait for the {ctx_name} process to finish first ...
+        sts = self._wait_for_process(process)
+        if sts is None:
+            # The {ctx_name} process did not terminate gracefully after 30 seconds,
+            # so we bring out the big guns ...
+            process.kill()
+            self.logger.error(
+                "Killed un-responsive %s tool %s %s process after" " 30 seconds",
```
webbnh, line 199: This message should omit the "after 30 seconds", since this code doesn't know how long `_wait_for_process()` waited.  (Also, this gets rid of the embedded `" "` break.)

(Of course, having `_wait_for_process()` take the overall timeout as a parameter might be a good thing....)

lib/pbench/agent/tool_meister.py
```diff
@@ -110,20 +127,113 @@ def __init__(
...
+                count -= 1
+                if count > 0:
+                    self.logger.debug(
+                        "%s tool %s%s has not stopped after %d seconds",
+                        self._tool_type,
+                        self.name,
+                        f" {ctx_name}" if ctx_name else "",
+                        5 * (6 - count),
+                    )
+        return sts
```
webbnh, line 180: I think I would have written this something like:
```
        intervals = 6
        timeout = 5
        for count in range(0, timeout * intervals, timeout):
            try:
                return process.wait(timeout=timeout)
            except subprocess.TimeoutExpired:
                self.logger.debug(
                        "%s tool %s%s has not stopped after %d seconds",
                        self._tool_type,
                        self.name,
                        f" {ctx_name}" if ctx_name else "",
                    count + timeout
                )
        return None
```

lib/pbench/agent/tool_meister.py
```diff
@@ -248,162 +348,133 @@ def wait(self):
...
-            if sts != 0 and sts != -(signal.SIGTERM):
-                self.logger.warning(
-                    "Transient tool %s start process failed with %d", self.name, sts
-                )
-        self.start_process = None
+        # First wait for the "stop" process to do it's job ...
+        self.stop_process = self._wait_for_process_w_kill(self.stop_process, "stop")
+        # ... then we wait for the start process to finish; in either case,
+        # we'll only wait a short time before killing them.
+        self.start_process = self._wait_for_process_w_kill(self.start_process, "start")
```
webbnh, line 355: Can we have the `ith` in `self._wait_for_process_with_kill()`?  (This line is already too long, so adding three more characters to make the name clear won't really hurt.)

lib/pbench/agent/tool_meister.py
```diff
@@ -248,162 +348,133 @@ def wait(self):
...
-                "3s",
-                "-c",
-                f"{self.pbench_install_dir}/templates/pmlogger.conf",
-                "--host=localhost:55677",
-                f"{self.tool_dir}/%Y%m%d.%H.%M",
-            ]
-        else:
-            self.pmlogger_args = None
+            # FIXME - Shouldn't this be provided by the environment?
+            os.environ["PATH"] += ":/usr/libexec/pcp/bin"
```
webbnh, line 369: > Shouldn't this be provided by the environment?

This is just providing a fall-back -- that is, this won't be used unless `find_executable()` couldn't find it in any of the directories in the existing `PATH`.

Ideally, we would use `find_executable()` on the existing `PATH` and then, only if that came up empty, we would see if there was a suitable file at this fallback location, and so forth.  The code here does basically the same thing only in a more straightforward way.  (To be honest, since it won't hurt if the directory is in the `PATH` more than once, I wouldn't bother with the `if` statement -- just append it and be done with it.)

portante:
Moving back to `Draft` mode, need to fix problems.

lib/pbench/agent/tool_meister.py
```diff
@@ -248,162 +348,133 @@ def wait(self):
...
+            "--report",
+            "-t",
+            "3s",
+            "-c",
+            f"{self.pbench_install_dir}/templates/pmlogger.conf",
+            "--host=localhost:55677",
+            f"{self.tool_inst_dir}/%Y%m%d.%H.%M",
+        ]
+
         return (0, "pcp tool (pmcd and pmlogger properly installed")
```
webbnh, line 402: Mismatched parentheses.

lib/pbench/agent/tool_meister.py
```diff
@@ -248,162 +348,133 @@ def wait(self):
...
+            )
+        if self.pmlogger_process is not None:
             raise ToolException(
-                f"Tool({self.name}) has an unexpected process still running"
+                f"Tool({self.name}) has an unexpected pmlogger process running"
             )
 
         self.logger.info(
-            "%s: start_tool -- %s -- %s",
+            "%s: start_tool -- '%s' && '%s'",
```
webbnh, line 415: As long as you're getting rid of `--`'s, I would get rid of them both.  E.g.,
```
        self.logger.info(
            "%s: [start_tool] '%s' && '%s'",
```

lib/pbench/agent/tool_meister.py
```diff
@@ -248,162 +348,133 @@ def wait(self):
...
+            self.pmlogger_args,
+            cwd=self.tool_inst_dir,
+            stdin=subprocess.DEVNULL,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+        )
+        pmlogger_reader = threading.Thread(
+            target=log_subprocess_output,
+            args=(self.pmlogger_process.stdout, self.logger, self.__class__.__name__),
+        )
```
webbnh, line 443: This is going to send the PM logger output to the same log stream as the PMCD output...is that "OK"?

lib/pbench/agent/tool_meister.py
```diff
@@ -248,162 +348,133 @@ def wait(self):
...
-            self.pmlogger_args = None
+            # FIXME - Shouldn't this be provided by the environment?
+            os.environ["PATH"] += ":/usr/libexec/pcp/bin"
+        self.pmcd_args = None
         self.pmcd_process = None
+        self.pmlogger_args = None
         self.pmlogger_process = None
-        if self.tool_dir:
-            try:
-                self.tool_dir.mkdir()
```
webbnh, line 333: Where/how does the tool directory get created now?

lib/pbench/agent/tool_meister.py
```diff
@@ -481,21 +545,24 @@ def start(self, env=None):
...
 
         This method does not wait for the process to actually exit. The caller
         should issue a wait() for that.
         """
         if self.process is None:
-            self.logger.error("Nothing to terminate")
+            self.logger.error("No process to stop")
             return
 
         # First try to gracefully terminate.
```
webbnh, line 557: "First"?  This is the only thing that this function does...we can probably do without this comment.

lib/pbench/agent/tool_meister.py
```diff
@@ -504,24 +571,9 @@ def wait(self):
...
-        elif sts != 0 and sts != -(signal.SIGTERM):
-            self.logger.warning(
-                "Persistent tool %s failed to return success, %d",
-                self.name,
-                self.process.returncode,
-            )
-        else:
-            self.logger.info("Stopped persistent tool %s", self.name)
-        self.process = None
+        self.process = self._wait_for_process_w_kill(self.process)
```
webbnh, line 576: I think this call is missing an argument.

Also, the last paragraph should be removed from the docstring for this function -- this function doesn't call `wait()`, it doesn't emit periodic log messages, and it doesn't know that the interval will be 30 seconds.

webbnh (commented):
OK, draft or not, here's some feedback.  😁

I found one problem (a missing argument), the rest are nits of various calibre.

lib/pbench/agent/tool_meister.py
```diff
@@ -80,6 +80,16 @@
 fmtstr = "%(asctime)s %(levelname)s %(process)s %(thread)s %(name)s %(funcName)s %(lineno)d -- %(message)s"
 
 
+def log_subprocess_output(
+    pipe: subprocess.PIPE, logger: logging.Logger, name: str
+) -> logging.Logger:
+    """Thread start function to log outputs from a given pipe.
+    """
+    inst_logger = logger.getChild(name)
```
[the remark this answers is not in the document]
portante: Yeah, let's see if a helper method for creating the process makes this any better.

lib/pbench/agent/tool_meister.py
```diff
@@ -80,6 +80,16 @@
 fmtstr = "%(asctime)s %(levelname)s %(process)s %(thread)s %(name)s %(funcName)s %(lineno)d -- %(message)s"
 
 
+def log_subprocess_output(
+    pipe: subprocess.PIPE, logger: logging.Logger, name: str
+) -> logging.Logger:
+    """Thread start function to log outputs from a given pipe.
+    """
```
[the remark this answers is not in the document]
portante: Sure, but why add support for the general case when we don't need it?

lib/pbench/agent/tool_meister.py
```diff
@@ -110,20 +127,112 @@ def __init__(
...
-        self.stop_process = None
 
-    def _check_no_processes(self):
-        if self.start_process is not None:
-            raise ToolException(
-                f"Tool({self.name}) has an unexpected start process running"
+    def install(self):
+        raise NotImplementedError(
+            f"{self.__class__.name} does not implement the install method"
+        )
```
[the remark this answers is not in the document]
portante: Those ABCs have so much mechanism, not sure it is really worth it for what we need.

lib/pbench/agent/tool_meister.py
portante: That would make them really different from all the others.

lib/pbench/agent/tool_meister.py
portante: Addressed in a new helper method.

lib/pbench/agent/tool_meister.py
portante: Argument is deliberately omitted.  Will fix the comment.

lib/pbench/agent/tool_meister.py
npalaska: This is a `PcpTransTool` so I believe the `tool_dir` creation should happen in `ToolMeister.start` call

npalaska (commented):
This looks nice, and this will bring lot of updates to our gold files where we would have to remove most of `tm-*.out/err` files coming from the transient tools. :smile:

npalaska (commented):
Another place where we are writing logs in a separate file is when we create tar balls, we can use `log_subprocess_output` mechanism to log the `tar` command errors as well. 

Also do we want use something like `--ignore-failed-read` on our `tar` creation to get around this error of `file changed as we read it` that you mentioned in one of the GChat conversation?

lib/pbench/agent/tool_meister.py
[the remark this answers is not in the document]
webbnh: Because the only change is to the type hint?

(I.e., thanks to the magic which is Python, we already **have** the support...we just have to "let" callers use it.)

lib/pbench/agent/tool_meister.py
webbnh: Unlike `_wait_for_process()`, `_wait_for_process_w_kill()` doesn't supply a default value for the second (explicit) argument (at least in the sources presented here)...so I would expect Python to report an error if this line gets executed (and I think LGTM agrees...).

So, sure, a comment might be nice, but, if you're not going to supply the last argument here, then you also need to fix the function definition....  🙂

lib/pbench/agent/tool_meister.py
webbnh: @npalaska, I'm not finding a `ToolMeister.start()` function.  The class has a `start_tools()` method -- did you mean that?  That does create a directory which it passes to the tool that it's starting.

@portante, so, the `mkdir()` call which was removed here was redundant (and presumably resulting in errors)?

FWIW, to my eye, the directory-creation code in `ToolMeister.init_tools()` and `ToolMeister.start_tools()` looks tragically similar (although `start_tools()` has better documentation)...should it be refactored into a helper function?  Also, I think the code recurs in `ToolMeister.sysinfo()`.  Having that level of duplication is really discomfiting.

lib/pbench/agent/tool_meister.py
portante: Allow me to address this in another commit coming.

lib/pbench/agent/tool_meister.py
portante: Yes, this is a bit ugly.  Let me take a stab at addressing this in another commit.  At least this fixes it to work as expected, `:` vs `os.pathsep`.

lib/pbench/agent/tool_meister.py
portante: Regarding @webbnh's question, yes.  We just don't have an existing test for it.

lib/pbench/agent/tool_meister.py
```diff
@@ -418,25 +496,15 @@ class PersistentTool(Tool):
...
-            logger.info(line)
-
     def start(self, env=None):
         assert self.args is not None, "Logic bomb!  {self.name} install had failed!"
-        assert (
-            self.tool_dir is not None
-        ), "Logic bomb!  No persistent tool directory provided!"
-        self.tool_dir = self.tool_dir / self.name
-        self.tool_dir.mkdir()
+        tool_dir = self.tool_dir / self.name
```
dbutenhof, line 506: The `TransientTool` asserts that `self.tool_dir` isn't `None`, but the persistent tool class just uses it.

Since both branches require it, seems this should really be checked once in the `Tool` constructor rather than deferring it until later when a problem will be harder to track down.

lib/pbench/agent/tool_meister.py
```diff
@@ -94,37 +103,170 @@ class Tool:
...
+    This base class provides for constructing an object with the required
+    parameters, ensuring the pbench installation directory and tool directory
+    exist.
+
+    The four abstract methods, install, start, stop, and wait, are defined, and
+    two helper methods are provided for waiting on processes.
     """
 
-    _tool_type = "Transient"
+    _tool_type = "None"
```
dbutenhof, line 114: Why `"None"` instead of `None`??

dbutenhof (approved):

lib/pbench/agent/tool_meister.py
```diff
@@ -200,210 +347,141 @@ def stop(self):
...
-                self.name,
-            )
-        else:
-            if sts != 0 and sts != -(signal.SIGTERM):
-                self.logger.warning(
-                    "Transient tool %s start process failed with %d", self.name, sts
-                )
-        self.start_process = None
+        # First wait for the "stop" process to do it's job ...
+        self.stop_process = self._wait_for_process_with_kill(self.stop_process, "stop")
```
webbnh, line 373: This seems like a bad idea to me:  `_wait_for_process_with_kill()` doesn't contain an explicit `return` statement, so we're counting Python's default/implicit behavior to generate the `None` value that we want to assign to `stop_process`.  I think it would be better to treat `_wait_for_process_with_kill()` as a subroutine rather than a function, and have a separate, explicit assignment statement for `stop_process`:
```
        self._wait_for_process_with_kill(self.stop_process, "stop")
        self.stop_process = None
```
This relieves the reader from having to consult `self._wait_for_process_with_kill()` to see what value(s) `stop_process` might end up with.

Ditto for lines 376-378.

npalaska (approved):
Looks good, I tested the code with recent changes and it seem to be working as expected (at least for the stuff that I can test), 
1. All the logs from TM related to persistent tools and transient tools landed in `tm.log` file
2. There are no `tm-*.out/err` files tracked in `<sampleX>/tools-<group>/<hostname>/` directory, only transient tools folder with its own tool data that it collected.

lib/pbench/agent/tool_meister.py
```diff
@@ -200,210 +347,141 @@ def stop(self):
...
-                "-c",
-                f"{self.pbench_install_dir}/templates/pmlogger.conf",
-                "--host=localhost:55677",
-                f"{self.tool_dir}/%Y%m%d.%H.%M",
-            ]
-        else:
-            self.pmlogger_args = None
+    def __init__(self, name, tool_opts, **kwargs):
+        super().__init__(name, tool_opts, **kwargs)
+        self.real_name = self.name.replace("-transient", "")
```
webbnh, line 389: `real_name` is an odd name for this field....especially since this field is almost never used, relative to `name`.

Since it's only used in one place, to generate a subdirectory name, it strikes me that maybe the directory name should be generated here, instead.  Or, at least, this field should be named something more like `subdir_name`.

lib/pbench/agent/tool_meister.py
dbutenhof: Not exactly what I'd intended, but since it appears you're going to change all this anyway in the next PR... "whatever" ...

dbutenhof (approved):
It's probably going to have a pain rebasing against all your `tool_meister.py` changes; but we're making progress, even if it's a bit messy. 😬

lib/pbench/agent/tool_meister.py
```diff
@@ -200,210 +347,141 @@ def stop(self):
...
-            except Exception as exc:
-                self.logger.error(
-                    "Pmlogger run process failed: '%s', %r", exc, self.pmlogger_args
-                )
+        self.pmcd_process = self._create_process_with_logger(
+            self.pmcd_args, tool_dir, "pmcd"
+        )
+        self.pmlogger_process = self._create_process_with_logger(
+            self.pmlogger_args, tool_dir, "pmlogger"
+        )
```
webbnh, line 455: Should we be checking that the processes were created successfully?

lib/pbench/agent/tool_meister.py
```diff
@@ -200,210 +347,141 @@ def stop(self):
...
-        return (0, "pcp tool (pmcd and pmlogger properly installed")
+        return (0, "pcp tool (pmcd and pmlogger) properly installed")
 
     def start(self):
-        assert self.tool_dir is not None, "Logic bomb!  no tool directory provided!"
-        if self.pmcd_process or self.pmlogger_process:
+        if not self.pmcd_path:
+            raise RuntimeError("Path to pmcd not provided")
+        if not self.pmlogger_path:
+            raise RuntimeError("Path to pmlogger not provided")
```
webbnh, line 414: Are these checks necessary?  (They look redundant with`install()`....)

portante:
Well this change introduces gold file instability.  Bummer.  Not sure what to do about the order in which the sub-process loggers work.

lib/pbench/agent/tool_meister.py
```diff
@@ -446,29 +514,15 @@ def start(self, env=None):
...
-                cwd=self.tool_dir,
-                stdout=subprocess.PIPE,
-                stderr=subprocess.STDOUT,
-                env=self.env,
-                shell=True,
-            )
         else:
             self.logger.debug(
                 "Starting persistent tool %s, args %r", self.name, self.args
             )
```
webbnh, line 526: Consider collapsing the two branches of this `if` statement into one:
```
            self.logger.debug(
                "Starting persistent tool %s%s, args %r",
                self.name,
                f", env PYTHONPATH={env["PYTHONPATH"]}" if env else "",
                self.args,
            )
```
(I sure hope `PYTHONPATH` actually _is_ in the environment....)

webbnh: And/or, perhaps, move the log call inside `_create_process_with_logger()`?

lib/pbench/agent/tool_meister.py
```diff
@@ -820,7 +857,7 @@ def __enter__(self):
             try:
                 tklass = self._tool_name_class_mappings[name]
             except KeyError:
-                tklass = Tool
+                tklass = TransientTool
```
webbnh, line 866: Shouldn't this `try` block be just a `get()` call??
```
            tklass = self._tool_name_class_mappings.get(name, TransientTool)
```

Ditto the one at 1193.

webbnh (commented):
This is just for the first commit (but I want to get it "in").  I'll move on to the next commit in a few minutes.

These are mostly nits and pointed questions....

Speaking of which...no unit tests?...not even for the new functions?

lib/pbench/agent/tool_meister.py
```diff
@@ -94,37 +103,170 @@ class Tool:
...
+
+    def _create_process_with_logger(
+        self, args: list, cwd: Path, ctx: str = None, env: dict = None
+    ) -> subprocess.Popen:
+        """Generic method of creating a sub-process with a thread to capture
+        stdout/stderr and log it.
+        """
+        if env:
+            process = subprocess.Popen(
+                " ".join(args),
```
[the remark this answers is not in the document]
portante: Not sure the history here, but I believe I had to do the construction myself in the past.  FWIW, this is removed in #2697.

lib/pbench/agent/tool_meister.py
```diff
@@ -94,37 +103,170 @@ class Tool:
...
-            raise ToolException(
-                f"Tool({self.name}) has an unexpected stop process running"
+        else:
+            process = subprocess.Popen(
+                args,
+                cwd=cwd,
+                stdin=subprocess.DEVNULL,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
+            )
```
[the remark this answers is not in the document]
portante: Moving code along, this is not new code, addressed in PR #2697.

lib/pbench/agent/tool_meister.py
portante: Let's take care of this in PR #2697.

webbnh: What was wrong with using `os.pathsep`?  (Not that we need it, but that seems like it would be more portable....)

portante: My bad, mistaked it for `os.path.sep`.

agent/util-scripts/gold/test-client-tool-meister/test-53.txt
```diff
@@ -177,17 +169,23 @@ INFO pbench-tool-meister sysinfo -- pbench-sysinfo-dump -- /var/tmp/pbench-test-
...
 INFO pbench-tool-meister start -- Started persistent tool dcgm, ['/var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/dcgm-exporter']
 INFO pbench-tool-meister start -- mpstat: start_tool -- /var/tmp/pbench-test-utils/opt/pbench-agent/tool-scripts/mpstat --start --dir=/var/tmp/pbench-test-utils/pbench/mock-run/0-iter-zero/sample42/tools-default/testhost.example.com --interval=42 --options=forty-two
+INFO pbench-tool-meister.logger-start log_subprocess_output -- mpstat: running "/var/tmp/pbench-test-utils/opt/pbench-agent/tool-scripts/datalog/mpstat-datalog 42 --options=forty-two "
 INFO pbench-tool-meister stop -- mpstat: stop_tool -- /var/tmp/pbench-test-utils/opt/pbench-agent/tool-scripts/mpstat --stop --dir=/var/tmp/pbench-test-utils/pbench/mock-run/0-iter-zero/sample42/tools-default/testhost.example.com --interval=42 --options=forty-two
-INFO pbench-tool-meister wait -- Waiting for transient tool mpstat stop process
-INFO pbench-tool-meister wait -- Waiting for transient tool mpstat start process
+INFO pbench-tool-meister _wait_for_process_with_kill -- Waiting for Transient tool mpstat stop process
```
webbnh, line 174: Do you like the now-capital `T` in `Transient` better than the previous lower-case one?

lib/pbench/agent/tool_meister.py
```diff
@@ -147,29 +289,34 @@ def start(self):
...
+        if not self.tool_dir.is_dir():
+            raise RuntimeError(f"tool directory does not exist: {self.tool_dir}")
+        if self.start_process is not None:
+            raise ToolException(
+                f"Tool({self.name}) has an unexpected start process running"
+            )
+        if self.stop_process is not None:
+            raise ToolException(
+                f"Tool({self.name}) has an unexpected stop process running"
+            )
```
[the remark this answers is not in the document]
portante: Can we address this in another commit?

lib/pbench/agent/tool_meister.py
```diff
@@ -200,210 +347,141 @@ def stop(self):
...
         completed.
 
         Waits for the tool's "stop" process to complete, if started, then
-        waits for the tool's start process to complete.
+        waits for the tool's start process to complete (since the "stop"
+        process is supposed stop the "start" process).
         """
+        assert self.tool_dir is not None, "Logic bomb!  no tool directory provided!"
+        if not self.tool_dir.is_dir():
+            raise RuntimeError(f"tool directory does not exist: {self.tool_dir}")
```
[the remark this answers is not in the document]
portante: Because until PR #2697 lands, `self.tool_dir` must be defined.  Meaning, `.wait()` should only be called in a context where `self.tool_dir` is present.

lib/pbench/agent/tool_meister.py
portante: The dangers of reviewing multiple commits.

agent/util-scripts/gold/test-client-tool-meister/test-53.txt
```diff
@@ -201,21 +199,27 @@ testhost.example.com 0000 INFO pbench-tool-meister sysinfo -- pbench-sysinfo-dum
...
-testhost.example.com 0013 INFO pbench-tool-meister stop -- Terminate issued for persistent tool dcgm
-testhost.example.com 0014 INFO pbench-tool-meister wait -- Stopped persistent tool dcgm
-testhost.example.com 0015 WARNING pbench-tool-meister end_tools -- testhost.example.com: unexpected temp files dcgm/dcgm-exporter.file
-testhost.example.com 0016 INFO pbench-tool-meister sysinfo -- pbench-sysinfo-dump -- /var/tmp/pbench-test-utils/opt/pbench-agent/unittest-scripts/pbench-sysinfo-dump /var/tmp/pbench-test-utils/pbench/mock-run/sysinfo/end/testhost.example.com block,security_mitigations,sos parallel
-testhost.example.com 0017 INFO pbench-tool-meister sysinfo -- testhost.example.com: sysinfo send (no-op) default /var/tmp/pbench-test-utils/pbench/mock-run/sysinfo/end/testhost.example.com
-testhost.example.com 0018 INFO pbench-tool-meister __exit__ -- testhost.example.com: terminating
+testhost.example.com 0004 INFO pbench-tool-meister.logger-start log_subprocess_output -- mpstat: running "/var/tmp/pbench-test-utils/opt/pbench-agent/tool-scripts/datalog/mpstat-datalog 42 --options=forty-two "
```
webbnh, line 202: We should consider masking the "timestamps" (like `0004`, here):  they make the differences much harder to review, because inserting or removing a single line in the sequence causes all subsequent "timestamps" to be different, which makes otherwise unchanged lines appear to be different, and that makes it hard to see which lines were actually changed.

webbnh (approved):
We just need to figure out what lingering bits get addressed in this PR and what gets addressed in #2697.

lib/pbench/agent/tool_meister.py
portante: That'll happen during the `.wait()`.

lib/pbench/agent/tool_meister.py
portante: Yes, all this is fixed in PR #2697.

lib/pbench/agent/tool_meister.py
portante: See PR #2697.

agent/util-scripts/gold/test-client-tool-meister/test-53.txt
portante: Those are not time stamps. those are the sequence order of logs.  We could consider dropping that for unit tests, but this highlights that out-of-order execution can and will occur (see failed test in PR #2697).

agent/util-scripts/gold/test-client-tool-meister/test-53.txt
portante: That comes from `self._tool_type`.

lib/pbench/agent/tool_meister.py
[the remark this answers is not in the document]
webbnh: By "another commit" I presume you mean "another PR"....  🙂

lib/pbench/agent/tool_meister.py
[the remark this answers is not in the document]
webbnh: This probably isn't worth arguing over, but I fail to see how it matters to _this function_ whether that attribute has a value, since this function doesn't reference it.  It seems like, if we're trying to protect a subroutine or otherwise verify an invariant, that check should either be in the code which needs protection or at some level above this one.

(I totally get the idea of sprinkling the flow with debugging statements to try to figure out where something goes south, but this is a PR, so the code should have been debugged by now, and, to the extent that we need to retain these checks, they should be properly positioned...and I don't yet see why this is "the right place" for these.)

Put another way, is a `RuntimeError` from here really better than an `AttributeError` from somewhere deeper in the stack, particularly for a case that we don't expect to occur?

lib/pbench/agent/tool_meister.py
webbnh: I think (and hope) that that is the wrong answer.  I think the correct answer is, "no, we don't need to check, because any failure would have been reported by an exception."  Am I right?

The problem is, after we return from this function -- and before we call `wait()` -- I think we report back that "all processes were started successfully", so, if we defer the check until we call `wait()`, then I think we end up having lied.

agent/util-scripts/gold/test-client-tool-meister/test-53.txt
webbnh: Yes...and so I'm asking whether you are satisfied with how the code is now setting `self._tool_type`....  😆

agent/util-scripts/gold/test-client-tool-meister/test-53.txt
webbnh: I could have been confused at the time, but I'm pretty sure that the code that generates those thinks it's producing timestamps (I believe it's a mock), and it goes to some trouble to make sure that they are all different and monotonically increasing....

webbnh (approved):
👍

lib/pbench/agent/tool_meister.py
webbnh: Can you comment on my suggestion to refactor the directory-creation code?  (I.e., tell me that it's already taken care of in #2697!  😆)

portante (commented):
Still some variability but its is taking me a while to reproduce it ...

lib/pbench/agent/tool_meister.py
portante: Yes, I believe it is taken care of in PR #2697.

agent/util-scripts/gold/test-client-tool-meister/test-53.txt
portante: See https://github.com/distributed-system-analysis/pbench/blob/main/lib/pbench/agent/redis.py#L212.

webbnh: Awesome!  You removed it!  Thanks!!

webbnh: Of course...since the numbers are still there in the gold file, it doesn't help so much on _this_ review...but the _next_ one will be easier.  😝

webbnh: > See https://github.com/distributed-system-analysis/pbench/blob/main/lib/pbench/agent/redis.py#L212.

Thanks for the reference!  That wasn't the one that I was thinking of (but, I cannot find the one that I was thinking of, now, so perhaps it was all a dream).

agent/util-scripts/unittests
```diff
@@ -574,15 +574,20 @@ function sort_tdslog {
 }
 
 function sort_tmlogs {
-    sort_log_file ${_testdir}/mock-run/tm/tm.logs
+    # Sort the tool-meister logs created by the tool data sink leveraging
+    # the sequence number ordering, but remove the sequence the number.
+    sort ${_testdir}/mock-run/tm/tm.logs \
+        | sed -e 's;^\([a-z\.0-9-]*\) \([0-9][0-9]*\) \(.*\)$;\1 \3;g' \
```
webbnh, line 579: This is OK for the moment, but there are a couple of things.
- I don't think you need the third subexpression:  if you match on the first two (with a trailing space) and replace them with the first one (plus a trailing space), then I think you've got it.  (The match doesn't need to be anchored at the end of the line.)
- The fact that either of the two expressions could be zero-length is a weakness, especially when paired with the third expression which is a total wildcard.  (E.g., it would match any line which began with two spaces.)

If you have time, you can use the `-E` switch to make `sed` use extended regular expressions, which will allow you to use `+` instead of `*` which tighten things up considerably.  Also, I think we can take advantage of the fact that the sequence number is always four digits:
```
        | sed -E -e 's;^([^[:space:]]+) ([0-9]{4}) ;\1 ;g'  \
```
(Replace one or more non-blank-space characters followed by a space followed by four digits followed by a space, anchored at the beginning of the line, with the first group followed by a space.)

webbnh (approved):
👍

lib/pbench/agent/tool_meister.py
webbnh: > The dangers of reviewing multiple commits.

Does this mean you're inclined to fix this, or should I hold my nose, or is this also fixed in #2697?

portante (commented):
Okay, it took a little bit to get right, but I think the gold files are now stable.

I added an initial commit to change the gold files to drop the sequence number, so that you can look at the diffs of the individual commits to see that the gold files changes make sense.

lib/pbench/agent/tool_meister.py
portante: Sorry, yes, I'll fix this.

webbnh (approved):
GTG

(Although, I did have [one more comment](https://github.com/distributed-system-analysis/pbench/pull/2696#pullrequestreview-914912995)...but depending on how you get notifications and view things, you may not have seen it because I "pre-reconciled" it.... 😏)

npalaska (approved):

dbutenhof (approved):
> I added an initial commit to change the gold files to drop the sequence number

On the whole I think I would have preferred "normalizing" the sequence number rather than dropping it entirely, e.g. something like "XXXX" or "0000" or even "[SEQ]".

Your changes, for example, won't fail if the sequence numbers were to disappear. Maybe that's OK, however, as they're not essential...

In any case, while this makes me vaguely uncomfortable, I won't object.

PR #2698, opened 2022-03-18
Use debug-level environment variables

dbutenhof:
The remote Tool Meister has a defined environment variable to override the default "INFO" debug level; however it was not propagated to the remote over ssh. This adds a command line parameter to directly pass the value to the remote.

The same logic was replicated in the Tool Data Sink; although "create" orchestration always forks this on the same host, this allows a simple and consistent command line hook to enable debug log messages when a client launches either command directly via "existing" orchestration.

This code isn't as elegant as I'd like, but in the name of minimizing the size and mess, I'm going to stop here. I'm hoping to refactor a class out of the `start` methods in both `tool_meister.py` and `tool_data_sink.py`, probably with a common base class to encapsulate some commonality. I considered doing that here in a second commit, but so much is flying around in these files doing it piecemeal should help to minimize the merge conflicts.

This refactoring would also make it easier to write unit tests.

### DIFF
--- a/agent/util-scripts/pbench-tool-meister-start
+++ b/agent/util-scripts/pbench-tool-meister-start
@@ -299,7 +299,7 @@ def start_tms_via_ssh(
                 )
                 tms[host] = {"status": "failed"}
             else:
-                # Record the child pid to wait below.
+                # Record the the host command has spawned
                 tms[host] = {"status": "spawned"}
 
     for host, tm_proc in tms.items():
--- a/agent/util-scripts/pbench-tool-meister-start
+++ b/agent/util-scripts/pbench-tool-meister-start
@@ -1062,7 +1062,7 @@ def main(_prog: str, cli_params: Namespace) -> int:
             try:
                 sysinfo_path.mkdir(parents=True)
             except Exception as e:
-                logger.debug("Unable to create sysinfo-dump directory {}: {}", sysinfo_path, e)
+                logger.debug("Unable to create sysinfo-dump directory %s: %s", sysinfo_path, e)
                 error_log(
                     f"Unable to create sysinfo-dump directory base path: {sysinfo_path}"
                 )
--- a/agent/util-scripts/pbench-tool-meister-start
+++ b/agent/util-scripts/pbench-tool-meister-start
@@ -244,16 +244,15 @@ def start_tms_via_ssh(
     failures = 0
     successes = 0
     tool_meister_cmd = exec_dir / "tool-meister" / "pbench-tool-meister"
-    base_args = [ssh_cmd]
-    base_args.extend(shlex.split(ssh_opts))
-    args = [
-        "<host replace me>",
-        f"{tool_meister_cmd}-remote",
-        redis_server.host,
-        str(redis_server.port),
-        "<tm param key>",
-        "yes",  # Yes, request the tool meister daemonize itself
-    ]
+    debug_level = os.environ.get("_PBENCH_TOOL_MEISTER_LOG_LEVEL")
+    cmd = f"{tool_meister_cmd}-remote {redis_server.host} {redis_server.port} {{tm_param_key}} yes"
+    if debug_level:
+        cmd += f" {debug_level}"
+    template = TemplateSsh(
+        ssh_cmd,
+        shlex.split(ssh_opts),
+        cmd
+    )
     tms = dict()
     tm_count = 0
     for host in tool_group.hostnames.keys():
@@ -276,6 +275,7 @@ def start_tms_via_ssh(
                             str(redis_server.port),
                             tm_param_key,
                             "yes",  # Yes, daemonize yourself TM ...
+                            debug_level
                         ]
                     )
                     sys.exit(status)
@@ -290,16 +290,9 @@ def start_tms_via_ssh(
                 # Record the child pid to wait below.
                 tms[host] = {"pid": pid, "status": "forked"}
         else:
-            args[0] = host
-            args[4] = tm_param_key
-            ssh_args = base_args + args
-            logger.debug(
-                "6b. starting remote tool meister, ssh_path=%r ssh_args=%r",
-                ssh_path,
-                ssh_args,
-            )
+            logger.debug("6b. starting remote tool meister on %s", host)
             try:
-                pid = os.spawnv(os.P_NOWAIT, ssh_path, ssh_args)
+                template.start(host, tm_param_key=tm_param_key)
             except Exception:
                 logger.exception(
                     "failed to create a tool meister instance for host %s", host
@@ -307,28 +300,40 @@ def start_tms_via_ssh(
                 tms[host] = {"status": "failed"}
             else:
                 # Record the child pid to wait below.
-                tms[host] = {"pid": pid, "status": "spawned"}
+                tms[host] = {"status": "spawned"}
 
     for host, tm_proc in tms.items():
         if tm_proc["status"] == "failed":
             failures += 1
             continue
-        pid = tm_proc["pid"]
-        try:
-            exit_status = _waitpid(pid)
-        except Exception:
-            failures += 1
-            logger.exception(
-                "failed to create a tool meister instance for host %s", host
-            )
-        else:
-            if exit_status != 0:
+        elif tm_proc["status"] == "forked":
+            pid = tm_proc["pid"]
+            try:
+                exit_status = _waitpid(pid)
+            except Exception:
+                failures += 1
+                logger.exception(
+                    "failed to create a tool meister instance for host %s", host
+                )
+            else:
+                if exit_status != 0:
+                    failures += 1
+                    logger.error(
+                        "failed to start tool meister on remote host '%s'"
+                        " (pid %d), exit status: %d",
+                        host,
+                        pid,
+                        exit_status,
+                    )
+                else:
+                    successes += 1
+        elif tm_proc["status"] == "spawned":
+            status = template.wait(host)
+            if status.status != 0:
                 failures += 1
                 logger.error(
-                    "failed to start tool meister on remote host '%s'"
-                    " (pid %d), exit status: %d",
+                    "failed to start tool meister on remote host '%s', exit status: %d",
                     host,
-                    pid,
                     exit_status,
                 )
             else:
@@ -390,6 +395,7 @@ class ToolDataSink(BaseServer):
                         str(redis_server.port),
                         tds_param_key,
                         "yes",  # Request tool-data-sink daemonize itself
+                        os.environ.get("_PBENCH_TOOL_DATA_SINK_LOG_LEVEL", "info")
                     ]
                 )
                 sys.exit(status)
@@ -1055,7 +1061,8 @@ def main(_prog: str, cli_params: Namespace) -> int:
             sysinfo_path = benchmark_run_dir / "sysinfo" / "beg"
             try:
                 sysinfo_path.mkdir(parents=True)
-            except Exception:
+            except Exception as e:
+                logger.debug("Unable to create sysinfo-dump directory {}: {}", sysinfo_path, e)
                 error_log(
                     f"Unable to create sysinfo-dump directory base path: {sysinfo_path}"
                 )
--- a/agent/util-scripts/pbench-tool-meister-start
+++ b/agent/util-scripts/pbench-tool-meister-start
@@ -299,7 +299,7 @@ def start_tms_via_ssh(
                 )
                 tms[host] = {"status": "failed"}
             else:
-                # Record the the host command has spawned
+                # Record that the host command has spawned
                 tms[host] = {"status": "spawned"}
 
     for host, tm_proc in tms.items():
--- a/lib/pbench/agent/tool_data_sink.py
+++ b/lib/pbench/agent/tool_data_sink.py
@@ -9,31 +9,29 @@
 #   sudo dnf install python3-bottle python3-daemon
 #   sudo pip3 install python-pidfile
 
+from datetime import datetime
 import errno
 import hashlib
+from http import HTTPStatus
 import json
 import logging
 import os
+from pathlib import Path
 import shutil
 import subprocess
 import sys
 import tempfile
-
-from configparser import DuplicateSectionError
-from datetime import datetime
-from distutils.spawn import find_executable
-from http import HTTPStatus
-from jinja2 import Environment, FileSystemLoader
-from pathlib import Path
 from threading import Thread, Lock, Condition
-from typing import Tuple
-from wsgiref.simple_server import WSGIRequestHandler, make_server
-
-import pidfile
-import redis
+from typing import Any, Dict, NamedTuple, Tuple
 
 from bottle import Bottle, ServerAdapter, request, abort
+from configparser import DuplicateSectionError
 from daemon import DaemonContext
+from distutils.spawn import find_executable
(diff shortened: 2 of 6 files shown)


agent/util-scripts/pbench-tool-meister-start
```diff
@@ -290,45 +290,50 @@ def start_tms_via_ssh(
...
+                template.start(host, tm_param_key=tm_param_key)
             except Exception:
                 logger.exception(
                     "failed to create a tool meister instance for host %s", host
                 )
                 tms[host] = {"status": "failed"}
             else:
                 # Record the child pid to wait below.
-                tms[host] = {"pid": pid, "status": "spawned"}
+                tms[host] = {"status": "spawned"}
```
npalaska, line 303: Maybe we also need to update the comment, since we are not recording the child pid here

lib/pbench/common/logger.py
```diff
@@ -8,6 +9,28 @@
...
+        hostname value.
+
+        Args:
+            hostname: Hostname to report. Defaults to platform.node().
+        """
+        self.hostname = hostname if hostname else platform.node()
+
+    def filter(self, record: logging.LogRecord) -> bool:
+        record.hostname = self.hostname
+        return True
```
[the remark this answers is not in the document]
dbutenhof: No, because in principle we could have more than one filter, and any filter returning `False` stops the chain.

Filtering is a bit weird, actually; the `logging.Filter` "base class" is actually an *example* of filtering by logger name. LGTM wants me to call the superclass `__init__`, but that just adds instance properties we don't want (and won't use in our `filter` overload). It's a great concept, but I get the sense maybe it's not used a lot. 😆

dbutenhof (commented):
I switched this to Draft; LGTM bugged me about some silly stuff `flake8` didn't care about, and when I pushed an update I managed to upset `black`. I still haven't seen LGTM's revised complaints, but I'll open it again when the peanut gallery is happy.

portante (commented):
Liking where this is going!  Let's get this merged!

lib/pbench/common/logger.py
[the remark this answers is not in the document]
dbutenhof: I should add here that filters can do either (or both, I suppose) of two things: they can supply record attributes (like I do here) which are substituted via the format string (`%(hostname)`), or they can *reject* a log record based on some attribute (which is what the example `logging.Filter` base class does). In the latter case, you'd return `False` to reject the record: the log record will be dropped without any additional filters being run. In the former case, we don't want to drop the record, and we want to allow additional filters (of either variety), so we always return `True`.

lib/pbench/agent/tool_meister.py
```diff
@@ -73,11 +71,13 @@
...
 from pbench.agent.toolmetadata import ToolMetadata
 from pbench.agent.utils import collect_local_info
+from pbench.common.logger import HostnameFilter
+from pbench.common.utils import md5sum
 
 
 # Logging format string for unit tests
 fmtstr_ut = "%(levelname)s %(name)s %(funcName)s -- %(message)s"
-fmtstr = "%(asctime)s %(levelname)s %(process)s %(thread)s %(name)s %(funcName)s %(lineno)d -- %(message)s"
+fmtstr = "%(asctime)s %(levelname)s %(hostname)s %(process)s %(thread)s %(name)s %(funcName)s %(lineno)d -- %(message)s"
```
portante, line 80: Why do we need to add the host name when we have it here (https://github.com/distributed-system-analysis/pbench/blob/main/lib/pbench/agent/redis.py#L212)?

lib/pbench/common/logger.py
```diff
@@ -8,6 +9,35 @@
...
+        logger "name". Calling the superclass constructor adds two instance
+        properties our `filter` will ignore, but LGTM.com complains if it's
+        omitted. In theory we can suppress the warning with
+        `# lgtm[py/missing-call-to-init]` but that doesn't seem to work.
+
+        Args:
+            hostname: Hostname to report. Defaults to platform.node().
+        """
+        super().__init__()
+        self.hostname = hostname if hostname else platform.node()
```
portante, line 34: In a production environment, the hostname is already in the logs added by `journald` (at least for our target RHEL environments).

While our test environment doesn't use `syslog`, production does and host name is added.

When remote ToolMeisters send logs to the Tool Data Sink, the host name is added there, see https://github.com/distributed-system-analysis/pbench/blob/main/lib/pbench/agent/redis.py#L212.

What is the case where we were missing a host name in the log output?

portante (changes requested):
Is there a way to demonstrate what results in the logs?  I believe production logs already have the host name added by `journald`.

lib/pbench/agent/tool_meister.py
dbutenhof: This was originally part of some experimentation to cover the gaps before we get the `RedisHandler` installed. What I probably really need is a log listener thread on `stderr` pipes for the `ssh` commands... which could be reported locally with the proper annotation, so maybe the filter wouldn't have been necessary anyway.

In any case, when I decided to scale back and just do the log level handoff, I didn't remove this. And you're right that as things are it doesn't help; some stuff remains stranded locally where the context is implicit, and once we get the `RedisHandler` it annotates the records.

lib/pbench/common/logger.py
dbutenhof: I'm not quite sure what you mean here; but I assume you mean when we're collecting the distributed syslogs into a common place. Where messages do currently show up in local `journalctl` streams (which, actually, none appear to do right now), they wouldn't be annotated ... but of course in that case the context is implicit so we don't need to add anything.

Ah, well; it was a fun experiment.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1966,47 +1968,28 @@ def daemon(
...
 
-def main(argv):
-    _prog = Path(argv[0])
-    PROG = _prog.name
+def start(prog: Path, parsed: Arguments):
+    PROG = prog.name
     # The Tool Data Sink executable is in:
     #   ${pbench_bin}/util-scripts/tool-meister/pbench-tool-data-sink
     # So .parent at each level is:
     #   _prog       ${pbench_bin}/util-scripts/tool-meister/pbench-tool-data-sink
```
webbnh, line 1988: s/_prog/prog/

lib/pbench/agent/tool_data_sink.py
```diff
@@ -2059,26 +2042,62 @@ def main(argv):
...
+        parsed = Arguments(
+            host=argv[1],
+            port=int(argv[2]),
+            key=argv[3],
+            daemonize=argv[4] == "yes" if len(argv) > 4 else False,
+            level=argv[5] if len(argv) > 5 else "info",
+        )
+    except (ValueError, IndexError) as e:
+        print(f"{PROG}: Invalid arguments, {argv!r}: {e}", file=sys.stderr)
+        return 1
```
webbnh, line 2094: Since we're at the outer-most level, I suspect that we should be catching all `Exception`'s -- I think that dying with a traceback should be avoided (logging it would be great; dumping it at the command line would be less so).

lib/pbench/agent/tool_meister.py
```diff
@@ -1700,7 +1699,7 @@ def sysinfo(self, data):
         return failures
 
 
-def get_logger(PROG: str, daemon: bool = False) -> logging.Logger:
+def get_logger(PROG, daemon: bool = False, level: str = "info") -> logging.Logger:
```
webbnh, line 1702: Type hint for `PROG`?

lib/pbench/agent/tool_meister.py
```diff
@@ -1728,22 +1727,31 @@ def get_logger(PROG: str, daemon: bool = False) -> logging.Logger:
     return logger
 
 
+class Arguments(NamedTuple):
+    host: str
+    port: int
+    key: str
+    daemonize: bool
+    level: str
```
webbnh, line 1735: This is the same definition as `tool_data_sync.Arguments`...should it be shared?

lib/pbench/agent/tool_meister.py
```diff
@@ -1728,22 +1727,31 @@ def get_logger(PROG: str, daemon: bool = False) -> logging.Logger:
...
+    tmp_dir: Path,
+    parsed: Arguments,
+    params: Dict[str, Any],
+    redis_server: redis.Redis,
+    logger: logging.Logger = None,
 ):
-    """driver - responsible for creating and driving operation of the Tool
+    """
+    responsible for creating and driving operation of the Tool
     Meister instance
```
webbnh, line 1751: "responsible for" doesn't deliver much value; how about replacing this this with something like
```
    Create and drive the operation of the Tool Meister instance
```
which fits comfortably on one line.

Similarly, line 1823.

lib/pbench/agent/tool_meister.py
```diff
@@ -1947,7 +1925,7 @@ def main(argv):
             break
     else:
         _sep = "" if not _path else ":"
-        os.environ["PATH"] = f"{_path}{_sep}{_prog.parent}"
+        os.environ["PATH"] = f"{_path}{_sep}{prog.parent}"
```
webbnh, line 1928: Consider using [`os.pathsep`](https://docs.python.org/3/library/os.html#os.pathsep) instead of `":"`.

lib/pbench/agent/tool_meister.py
```diff
@@ -1978,41 +1956,75 @@ def main(argv):
...
+        parsed = Arguments(
+            host=argv[1],
+            port=int(argv[2]),
+            key=argv[3],
+            daemonize=argv[4] == "yes" if len(argv) > 4 else False,
+            level=argv[5] if len(argv) > 5 else "info",
+        )
+    except (ValueError, IndexError) as e:
+        print(f"{PROG}: Invalid arguments, {argv!r}: {e}", file=sys.stderr)
+        return 1
```
webbnh, line 2021: As I mentioned above, we should have a catch-all clause which logs the traceback and prints only the exception message, rather than dying and dumping the traceback to the terminal.

webbnh (commented):
Nice improvements.  Most of my comments are just nits and typos.  However, ~I don't think we should ignore LGTM's concern,~ I do think it would be best if we avoided dumping tracebacks to the terminal, and I'm wondering if we should improve the support for logging granularity.

(This review is from before the update you just pushed; I'll circle back to that, now.)

lib/pbench/common/logger.py
```diff
@@ -8,6 +9,35 @@
 from pbench.common.exceptions import BadConfig
 
 
+class HostnameFilter(logging.Filter):
```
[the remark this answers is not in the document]
webbnh: Well, nevermind.  😁

webbnh (dismissed):
Looks good, once the wrinkles are smoothed out.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -1055,7 +1061,8 @@ def main(_prog: str, cli_params: Namespace) -> int:
             sysinfo_path = benchmark_run_dir / "sysinfo" / "beg"
             try:
                 sysinfo_path.mkdir(parents=True)
-            except Exception:
+            except Exception as e:
+                logger.debug("Unable to create sysinfo-dump directory {}: {}", sysinfo_path, e)
```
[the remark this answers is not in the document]
dbutenhof: Ouch; no, we don't, and I've been dinged before on doing it where not necessary. However, I wrote this just because there was another nearby `except` that wrote to the debug log in addition to the stderr/file `error_log` and it seemed to "make sense" to be consistent ... but looking at it now I realize that I used *server* syntax `{}` rather than *agent* syntax `%s` and this won't actually work. 😦

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1831,16 +1830,16 @@ def put_document(self, data_ctx, hostname):
             abort(500, "INTERNAL ERROR")
 
 
-def get_logger(PROG, daemon=False):
-    """get_logger - construct a logger for a Tool Meister instance.
+def get_logger(PROG: str, daemon: bool = False, level: bool = "info") -> logging.Logger:
```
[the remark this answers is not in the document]
dbutenhof: Obviously it's *not* a `bool`, and I have no idea why my editor put that here. Bad editor.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1831,16 +1830,16 @@ def put_document(self, data_ctx, hostname):
...
     If in non-unit test environment:
        If daemonized, log to syslog and log back to Redis.
        If not daemonized, log to console AND log back to Redis
     """
     logger = logging.getLogger(PROG)
-    if os.environ.get("_PBENCH_TOOL_DATA_SINK_LOG_LEVEL") == "debug":
+    if level == "debug":
         log_level = logging.DEBUG
     else:
         log_level = logging.INFO
```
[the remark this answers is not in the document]
dbutenhof: Yeah; this was all based on simply transporting Pete's original logging protocol across state (er, network) lines. The default is "info" and you have the option of overriding that to "debug". I could have generalized this, and chose not to. If you want to make that argument, open an issue and we can deal with it separately.

And, yes, I suppose I could have used the `int` value at this level, but this encapsulates the interpretation of the environment variable value in `get_logger`, which I liked.

lib/pbench/agent/tool_data_sink.py
dbutenhof: I really wasn't trying to completely rearchitect the file. (Although I kinda started out in that direction, seeing Pete already doing multiple PRs overlaying this code dissuaded me from being very radical. I peeled of `main` from `start` because I was going to use `ArgumentParser` and options, but I chickened out and even backed off on that.)

agent/util-scripts/pbench-tool-meister-start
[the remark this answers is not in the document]
webbnh: Hurray for code review!  (Nevermind that I found the wrong bug, I found a bug!!  🤣)

lib/pbench/agent/tool_meister.py
dbutenhof: These aren't CLI commands; they're only run via `ssh` from `pbench-tool-meister-start` and any output (except through the `logger` once we hook up the `RedisHandler`) is completely lost anyway. I'd like to work on that, and originally intended to at least start addressing it in this PR, but as mentioned I scaled it back to minimize conflicts. (And, yes, I didn't undo my main/start split so I wasn't entirely consistent in following through on that decision, but under the circumstances I don't feel inclined to add another level of `try`/`except` for this comment, in this context.)

lib/pbench/agent/tool_data_sink.py
portante: I am not touching this code in other PRs (except for the `black` work in PR #2703).

portante (approved):
Looks good!

lib/pbench/agent/tool_meister.py
webbnh: Ha!  Since the traceback will otherwise be lost, as you say, it's important then to catch the exception and properly log it....  😆

Seriously, though, I think that all I'm suggesting here is one additional `except` clause which catches all the other exceptions and logs them.  But, if there's another PR coming for this stuff, I can retract this comment....

webbnh (approved):
👍

Issue #2665, opened 2022-03-08
Whither `/var/lib/pbench-agent`?


webbnh:
In a fresh, RPM-based installation of the Pbench Agent (such as one might find in the `quay.io/pbench/pbench-agent-all-centos-8:beta` container), the default configuration points `pbench_run` at `/var/lib/pbench-agent`...but that directory doesn't exist, yet:
```
[root@c63f47495b69 /]# pbench-register-tool --name=vmstat
[ERROR] pbench run directory, /var/lib/pbench-agent, does not exist
```
The question is, is this a bug or a feature?

The problem is, ostensibly, the user should be allowed to put the "run directory" anywhere s/he wants and then point to it via the `pbench_run` environment variable or by configuring `pbench-agent:pbench_run` in their config file.  It is only if no other provision is made that the value ends up being `/var/lib/pbench-agent`.  So, there's probably a reasonable argument to be made that _installation_ should not create the directory and, in fact, that the directory shouldn't be created until some Pbench command is run...and, possibly, not even then!

But, a user running a Pbench command (as above) and having it fail because they didn't manually create the run directory creates a poor user experience (particularly, if the default is in a system directory which is not normally user writable).

Still, we don't really want to put Pbench in the position of creating the directory on the fly:
- if we think that the directory doesn't exist because the user made a typo when specifying the location, then we'll create a new one in the "wrong place";
- if we create a directory in any location other than the default, then we won't know to remove it if pbench is uninstalled; and,
- if the invocation happens to be privileged, we don't want pbench to become a vector for a security attack.

So, what to do??

But, most of all, our behavior should be consistent.  E.g., given the above, this is a problem:
```
[root@c63f47495b69 /]# ls -ld /var/lib/pbench-agent; pbench-list-tools; ls -ld /var/lib/pbench-agent
ls: cannot access '/var/lib/pbench-agent': No such file or directory
No tool groups found
drwxr-xr-x. 3 root root 35 Mar  8 00:11 /var/lib/pbench-agent
```
(Yeah:  `pbench-list-tools`, unlike `pbench-register-tool`, when run as root, will create `/var/lib/pbench-agent` if it doesn't exist.)

npalaska:
[...]

PR #2699, opened 2022-03-21
Create pbench_run directory on remote if it doesn't exists

npalaska:
Issue #2665 
If the `pbench_run` directory specified in pbench_agent config on remote host does not exist, we need to create it. 

We create it in our `lib/pbench/agent/base.BaseCommand` class if the directory does not exists already, however, if the same happens when running `agent/base` file we exit. I think we need to be consistent with our `pbench_run` directory creation.

If any script that runs `agent/base` file and need `pbench_run` directory, the base file will create it (if it does not exists already) when it creates the `tmp` directory (`"$pbench_run/tmp"`)

### DIFF
--- a/agent/base
+++ b/agent/base
@@ -16,10 +16,11 @@ if [[ -z "$pbench_run" ]]; then
         pbench_run=/var/lib/pbench-agent
     fi
     export pbench_run
-fi
-
-if [[ ! -d $pbench_run ]]; then
-    echo "[Warning] pbench run directory, ${pbench_run}, does not exist and will be created" >&2
+else
+    if [[ ! -d $pbench_run ]]; then
+        echo "[ERROR] the provided pbench run directory, ${pbench_run}, does not exist" >&2
+        exit 1
+    fi
 fi
 
 # the pbench temporary directory is always relative to the $pbench_run
--- a/agent/base
+++ b/agent/base
@@ -19,8 +19,7 @@ if [[ -z "$pbench_run" ]]; then
 fi
 
 if [[ ! -d $pbench_run ]]; then
-    echo "[ERROR] pbench run directory, ${pbench_run}, does not exist" >&2
-    exit 1
+    echo "[Warning] pbench run directory, ${pbench_run}, does not exist and will be created" >&2
 fi
 
 # the pbench temporary directory is always relative to the $pbench_run
--- a/lib/pbench/agent/base.py
+++ b/lib/pbench/agent/base.py
@@ -26,7 +26,7 @@ class BaseCommand(metaclass=abc.ABCMeta):
             self.pbench_run = pathlib.Path(env_pbench_run)
             if not self.pbench_run.exists():
                 click.echo(
-                    f"[ERROR] pbench run directory does not exist, {env_pbench_run}",
+                    f"[ERROR] the provided pbench run directory, {env_pbench_run}, does not exist",
                     err=True,
                 )
                 click.get_current_context().exit(1)
--- a/lib/pbench/agent/base.py
+++ b/lib/pbench/agent/base.py
@@ -23,13 +23,14 @@ class BaseCommand(metaclass=abc.ABCMeta):
 
         env_pbench_run = os.environ.get("pbench_run")
         if env_pbench_run:
-            self.pbench_run = pathlib.Path(env_pbench_run)
-            if not self.pbench_run.exists():
+            pbench_run = pathlib.Path(env_pbench_run)
+            if not pbench_run.is_dir():
                 click.echo(
                     f"[ERROR] the provided pbench run directory, {env_pbench_run}, does not exist",
                     err=True,
                 )
                 click.get_current_context().exit(1)
+            self.pbench_run = pbench_run
         else:
             self.pbench_run = pathlib.Path(self.config.pbench_run)
             if not self.pbench_run:
--- a/lib/pbench/agent/base.py
+++ b/lib/pbench/agent/base.py
@@ -22,16 +22,18 @@ class BaseCommand(metaclass=abc.ABCMeta):
         self.name = os.path.basename(sys.argv[0])
 
         env_pbench_run = os.environ.get("pbench_run")
-        self.pbench_run = pathlib.Path(
-            env_pbench_run if env_pbench_run else self.config.pbench_run
-        )
-
-        if not self.pbench_run.exists():
-            click.echo(
-                f"[ERROR] pbench run directory does not exist, {self.pbench_run}",
-                err=True,
-            )
-            click.get_current_context().exit(1)
+        if env_pbench_run:
+            self.pbench_run = pathlib.Path(env_pbench_run)
+            if not self.pbench_run.exists():
+                click.echo(
+                    f"[ERROR] pbench run directory does not exist, {env_pbench_run}",
+                    err=True,
+                )
+                click.get_current_context().exit(1)
+        else:
+            self.pbench_run = pathlib.Path(self.config.pbench_run)
+            if not self.pbench_run:
+                self.pbench_run = pathlib.Path("/var/lib/pbench-agent")
 
         # the pbench temporary directory is always relative to the $pbench_run
         # directory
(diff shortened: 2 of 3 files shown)


portante:
PR #2613 removed it for a reason, which does not look valid then.  But adding it back this way is not right either, I don't think.  Please take a look at the code before PR #2613.

This really needs some discussion and understanding of the behavior.

Originally, the `base` script was responsible for only creating the `${pbench_run}` directory IFF it was not specified as an explicit environment variable (i.e. only specified in the `pbench-agent.cfg` file).

It was changed in PR #2613 to not create it at all.  That seems to be the wrong thing to do.

dbutenhof:
> It was changed in PR #2613 to not create it at all. That seems to be the wrong thing to do.

The primary motivation of #2613 was to make the `bash` and `python` consistent. This was discussed a lot at the time, and "we all" (for some definition of "we" and "all") thought this looked like a reasonable course.

If not, we should at least make sure that we don't diverge the bash and python behaviors.

npalaska:
Yeah our click commands using `PbenchAgentConfig` create this directory (whether coming from an environment variable or the agent config file) if it doesnt exist already even though we have this check in place (`if not self.pbench_run.exists()`). I am not sure how the check returns False even if we dont have the directory (for example `pbench-list-tools` would create it if it does not exists but the TM which depends on bash does not).

We either need to document that the user need to make sure the `pbench_run` directory exists or we should explicitly try to create it because we depend on it.

portante (changes requested):
I am not in favor of this change as currently coded.

I believe we need to go back to pre-PR #2613 and fix the `PbenchAgentConfig` code to only create the directory if it is from the configuration file and not from the `pbench_run` environment variable.

npalaska:
> I believe we need to go back to pre-PR #2613 and fix the `PbenchAgentConfig` code to only create the directory if it is from the configuration file and not from the `pbench_run` environment variable.

Yes that's another parallel way to achieve the consistent behavior from bash and python base script

webbnh:
> fix the `PbenchAgentConfig` code to only create the directory if it is from the configuration file and not from the `pbench_run` environment variable.

That seems like a pretty good compromise.

webbnh:
I took the liberty of marking this PR as a draft, pending the rework.

npalaska:
Thanks @webbnh, I will open it when it ready.

lib/pbench/agent/base.py
```diff
@@ -22,16 +22,18 @@ def __init__(self, context):
...
-            click.echo(
-                f"[ERROR] pbench run directory does not exist, {self.pbench_run}",
-                err=True,
-            )
-            click.get_current_context().exit(1)
+        if env_pbench_run:
+            self.pbench_run = pathlib.Path(env_pbench_run)
+            if not self.pbench_run.exists():
+                click.echo(
+                    f"[ERROR] pbench run directory does not exist, {env_pbench_run}",
```
portante, line 29: Could you make the error messages between `base` and `base.py` look the same?

portante (approved):
Looks good to me!  Consider making the error messages the same.

portante (approved):

lib/pbench/agent/base.py
```diff
@@ -22,16 +22,18 @@ def __init__(self, context):
...
+                    err=True,
+                )
+                click.get_current_context().exit(1)
+        else:
+            self.pbench_run = pathlib.Path(self.config.pbench_run)
+            if not self.pbench_run:
+                self.pbench_run = pathlib.Path("/var/lib/pbench-agent")
 
         # the pbench temporary directory is always relative to the $pbench_run
         # directory
```
webbnh, line 39: If `$pbench_run/tmp` already exists, the Bash code will proceed while the Python code will fail.  Is this OK/desirable?  (We can easily make the Python code match the Bash code by adding `exist_ok=True` to the `os.makedirs()` call at line 43.)

lib/pbench/test/functional/agent/cli/commands/test_config.py
```diff
@@ -18,7 +18,10 @@ def test_pbench_run_dir_existence(monkeypatch, agent_config, pbench_run):
     out, err, exitcode = pytest.helpers.capture(command)
     assert exitcode == 1
     assert b"" == out
-    assert b"[ERROR] pbench run directory does not exist," in err
+    assert (
+        f"[ERROR] the provided pbench run directory, {os.getenv('pbench_run')}, does not exist\n".encode()
```
webbnh, line 22: We know that the value of `pbench_run` should be, since we set it at line 16; therefore, we should be checking that the error message contains _that_, rather than checking that it contains what is _currently_ in the environment.

That is, the code should look something like this:
```
    my_pbench_run = f"{pbench_run}/test"
    monkeypatch.setenv("pbench_run", my_pbench_run)
...
    assert (
        f"[ERROR] the provided pbench run directory, {my_pbench_run}, does not exist\n".encode()
        == err
    )
```

webbnh (approved):
Looks good!

However, I think there's still a difference in behavior between Bash and Python; and, I propose a tweak to the unit test.

lib/pbench/agent/base.py
portante: We should make them behave the same.

npalaska: > If `$pbench_run/tmp` already exists, the Bash code will proceed while the Python code will fail

WDYM? We have this check in base.py `if not self.pbench_tmp.exists():` which should prevent python to fail if the tmp directory already exist.

I just noticed that there is one difference we need to fix between bash and python and that is `base.py` has this check `if not self.pbench_tmp.exists():` and if it returns True we create the directory using `os.makedirs(self.pbench_tmp)` which will also create the parent directories. Bash does not have this check it just proceed to create the directory with `mkdir -p` even if it exists. 

I suppose we can do 
```
try:
    os.makedirs(self.pbench_tmp, exists_ok=True)
except OSError:
    ...
``` 
instead of 
```
if not self.pbench_tmp.exists():
    try:
        os.makedirs(self.pbench_tmp)
    except OSError:
        ...
```
Is this what you mean?

portante: I agree that they appear to be the same.  @webbnh, can you point out where you see the difference?

lib/pbench/agent/base.py
```diff
@@ -22,16 +22,18 @@ def __init__(self, context):
...
-
-        if not self.pbench_run.exists():
-            click.echo(
-                f"[ERROR] pbench run directory does not exist, {self.pbench_run}",
-                err=True,
-            )
-            click.get_current_context().exit(1)
+        if env_pbench_run:
+            self.pbench_run = pathlib.Path(env_pbench_run)
+            if not self.pbench_run.exists():
```
portante, line 27: To match the `base` script this needs to be `if not self.pbench_run.is_dir()`.

dbutenhof: Or 

```
    pbench_run = pathlib.Path(env_pbench_run)
    if not pbench_run.is_dir():
        ...
    self.pbench_run = pbench_run
```

agent/base
```diff
@@ -16,11 +16,11 @@ if [[ -z "$pbench_run" ]]; then
         pbench_run=/var/lib/pbench-agent
     fi
     export pbench_run
-fi
-
-if [[ ! -d $pbench_run ]]; then
-    echo "[ERROR] pbench run directory, ${pbench_run}, does not exist" >&2
-    exit 1
+else
```
riya-17, line 19: What if we change the value of pbench_run` in the config file? Don't we wanna check then, whether the directory exists or not?

dbutenhof: Pbench will implicitly create the default run directory (from the config file or fallback default) if it doesn't already exist (`mkdir -p` at line 29 in `base` or `os.makedirs` at line 42 in `base.py`). However we make sure that only takes effect when the value of `pbench_run` does come from the config file; if it's specified by an environment variable, and doesn't exist, we fail.

That is, regardless of the value in the config file, we'll make sure it exists the first time we reference it. (And if something deletes it later, a subsequent tool will fail.) If the user overrides it with an environment variable, however, they're responsible for creating it first: we don't own it, so we make the existence check.

npalaska: Riya, `mkdir -p` at line 29 in `base` should take care of creating a new pbench_run directory if the value in config file is changed. However, if user overrides it with an environment variable as Dave mentioned they are responsible for the directory creation as well.

lib/pbench/agent/base.py
webbnh: Nope -- my bad:  I somehow missed or misread the `if not self.pbench_tmp.exists()` test.

My apologies!

portante (approved):

webbnh (approved):
👍

dbutenhof (approved):

agent/base
riya-17: Thanks

riya-17 (approved):

PR #2697, opened 2022-03-18
Refactor `ToolMeister` to construct Tools once

portante:
We were running the constructor for each Tool object once to run the `install()` method, and then again for each "init" and "start" operation.

Now the Tool object constructor runs once, and all code used to determine if a tool is installed correctly results in the `install()` method.

PR #2706, opened 2022-03-22
Stop using `Exception` everywhere

portante:
Base on PR #2697.

### DIFF
--- a/lib/pbench/agent/tool_meister.py
+++ b/lib/pbench/agent/tool_meister.py
@@ -111,23 +111,28 @@ class Tool:
 
     _tool_type = None
 
-    def __init__(self, name, tool_opts, pbench_install_dir, logger, tool_dir=None):
+    def __init__(
+        self,
+        name,
+        tool_opts,
+        pbench_install_dir,
+        logger,
+    ):
         self.name = name
         self.tool_opts = tool_opts
         if not pbench_install_dir.is_dir():
-            raise RuntimeError(
+            raise ToolException(
                 f"pbench installation directory does not exist: {pbench_install_dir}"
             )
         self.pbench_install_dir = pbench_install_dir
         self.logger = logger
-        self.tool_dir = tool_dir
 
     def install(self):
         raise NotImplementedError(
             f"{self.__class__.__name__} does not implement the install method"
         )
 
-    def start(self):
+    def start(self, tool_dir: Path):
         raise NotImplementedError(
             f"{self.__class__.__name__} does not implement the start method"
         )
@@ -143,29 +148,18 @@ class Tool:
         )
 
     def _create_process_with_logger(
-        self, args: list, cwd: Path, ctx: str = None, env: dict = None
+        self, args: list, cwd: Path, ctx: str = None
     ) -> subprocess.Popen:
         """Generic method of creating a sub-process with a thread to capture
         stdout/stderr and log it.
         """
-        if env:
-            process = subprocess.Popen(
-                " ".join(args),
-                cwd=cwd,
-                stdin=subprocess.DEVNULL,
-                stdout=subprocess.PIPE,
-                stderr=subprocess.STDOUT,
-                env=env,
-                shell=True,
-            )
-        else:
-            process = subprocess.Popen(
-                args,
-                cwd=cwd,
-                stdin=subprocess.DEVNULL,
-                stdout=subprocess.PIPE,
-                stderr=subprocess.STDOUT,
-            )
+        process = subprocess.Popen(
+            args,
+            cwd=cwd,
+            stdin=subprocess.DEVNULL,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+        )
         _ctx = f"-{ctx}" if ctx else ""
         process_logger = threading.Thread(
             target=log_subprocess_output,
@@ -254,15 +248,17 @@ class TransientTool(Tool):
 
     def __init__(self, name, tool_opts, **kwargs):
         super().__init__(name, tool_opts, **kwargs)
+        self.tool_script = f"{self.pbench_install_dir}/tool-scripts/{self.name}"
         self.start_process = None
         self.stop_process = None
+        self.tool_dir = None
 
     def install(self):
         """Synchronously runs the tool --install mode capturing the return code and
         output, returning them as a tuple to the caller.
         """
         args = [
-            f"{self.pbench_install_dir}/tool-scripts/{self.name}",
+            self.tool_script,
             "--install",
             self.tool_opts,
         ]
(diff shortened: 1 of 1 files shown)


dbutenhof (commented):
Just some rambling impressions first thing in the morning ...

While I appreciate the intent to make localized exception handling more specific, I question whether the broad categories of `Warning`, `RuntimeError`, and `OSError` are really better than `Exception`.

I have my doubts about handling `Warning` at all; do we really expect to be running our code in `-X dev` mode? And, if so, we wouldn't do this in production, only in internal testing; I think this would be a case where we'd want a clean termination rather than having to search the log output.

Philosophically, the `except Exception` says "if anything goes wrong in this section of code, trap the exception locally, report as a localized operation failure". Being more specific is great, but I'm pretty hazy about a middle ground of "a few generic exceptions" as a replacement.

I feel like the main objective here is really just to exclude `AssertionError`; but moving to including a seemingly somewhat arbitrary subset of its siblings doesn't feel like an improvement to me. And if that's really all we care about, I'd almost prefer the (admittedly somewhat more ugly) alternatives of

```
    except AssertionError:
        raise
    except Exception:
        ...
```

or even

```
    except Exception as e:
        if isinstance(e, AssertionError):
            raise
```

... either of which expresses that specific intent more clearly.

And as much as I love exceptions, if the point of `assert` is to help diagnose "impossible" conditions, I prefer the C convention of "print a message and dump core right now" vs trying to propagate an exception through <n> levels of code which may disturb the process state with arbitrary cleanups. Using `assert` as "just another exception" never felt right to me.

webbnh:
There are basically three things that code can do in response to an exception (other than just letting it propagate past):
- log it
- clean-up/shutdown
- eat it and recover

It's perfectly reasonable for code to catch an exception, log the result, and re-raise it.  And, for that purpose, it's OK to use very broad `except` clauses (so long as the exception doesn't indicate a failure in the logging facility...).  Analogously, it is reasonable to structure code such that, in the event of an exception, it restores its external invariants and then re-raises the exception.  And, again, for that sort of purpose, it's generally OK (and probably critical) to use very broad `except` clauses (again, so long as the exception doesn't indicate that there will be problems with the cleanup).  In each of these cases, the original exception typically continues to propagate, and the current function's operation is aborted.

The last response is to _recover_ from the exception.  However, in order to be safe and effective, the calling code has to understand the semantics associated with the exception that it caught; otherwise, there's a risk that the code's response will be inappropriate or ineffective, or, at worst, it will compound the problem and result in a subsequent, harder-to-diagnose failure.  Thus, an `except Exception` clause in these cases should raise a red flag -- how can the catching code know that it is doing the right thing if it doesn't know what it caught?

My framework for thinking about exceptions is that, like input parameters and return value, they are part of the interface for the the function.  Thus, in an ideal world, the caller of a function should know what exceptions can be raised by the function and the semantics associated with each one.  (Unfortunately, it's not always easy for a function to document all the exceptions that might emanate from it, as some of them might come from subroutines that it calls and maintaining the transitive closure of the set is an overwhelming challenge...in my experience, Python modules don't even try.)

So, this leaves us trying to come up with a good way to figure out which exceptions to catch and which ones to allow to propagate.

My recommendation is to first determine which of the three cases we're dealing with.  If it's one of the first two, then I'm OK with using `except Exception` (at least initially).  However, if the code is attempting to recover from an exception, then I think it bears close scrutiny.

(A quick tangent:  in most cases, I don't think that we should be allowing exceptions to propagate out of the `main()` function.  For a CLI program, I think we should have a "catch-all" handler in the outer-most block which prints the exception message to the terminal and then exits with an error status.  It would be useful to log the traceback, but I don't think we should be displaying it to the user.  So, this would be an "exception" to the "don't try to recover" rule, but it's not really a recovery, per se, since the action is to exit the program.)

To address the other cases, we could implement a common function which encapsulates the policy of which exceptions we catch and which we ignore.  In that way, we can easily revisit our choices, without having to make wide-scale changes to the code.  It could look something like this:

```
def pbench_catchable_exceptions():
    return (OSError, RuntimeError)
```
and it would be used like this:
```
try:
    dangerous_function()
except pbench_catchable_exception():
    do_recovery()
```
which looks to me like pretty reasonable code.

dbutenhof:
> However, in order to be safe and effective, the calling code has to understand the semantics associated with the exception that it caught; otherwise, there's a risk that the code's response will be inappropriate or ineffective, or, at worst, it will compound the problem and result in a subsequent, harder-to-diagnose failure.

I think your `pbench_catchable_exception()` suggestion violates this principle, because it implies a global unconditional knowledge of the semantics of this externally (to the module) defined exception tuple with regard to every block of code being protected; an invariant that (at best) will be broken the first time this encapsulated specification changes.

I doubt there are many low-level blocks of code that ought to be finalizing `RuntimeError` or even `OSError` -- the former because it's so loosely defined and the latter because it's so broad (essentially covering all of standard UNIX `errno` values).

If we're unlinking a file and it's OK if the file doesn't already exist, for example, we should be catching and finalizing `FileNotFoundError`, not `OSError`. If other OSError cases aren't expected, we shouldn't be finalizing them.

A lot of our `except Exception` cases are in your first category; they're intended to improve diagnosability by clarifying a subsystem: we don't need the `os.unlink` traceback, but we want to clarify that a cleanup function tried to remove a file that didn't exist by logging it as "foobar file should have existed but didn't" ... with or without propagating an exception (whether the original or a new exception; and in this case maybe even dropping the cause with the `from None` clause since it probably doesn't help diagnosis). This is certainly in line with your request for `main` behaviors... though many of these cases could also be described as "eat and recover".

webbnh:
> > However, in order to be safe and effective, the calling code has to understand the semantics associated with the exception that it caught; otherwise, there's a risk that the code's response will be inappropriate or ineffective, or, at worst, it will compound the problem and result in a subsequent, harder-to-diagnose failure.
> 
> I think your `pbench_catchable_exception()` suggestion violates this principle, because it implies a global unconditional knowledge of the semantics of this externally (to the module) defined exception tuple with regard to every block of code being protected; an invariant that (at best) will be broken the first time this encapsulated specification changes.

You are, of course, correct.  But, as a replacement for `except Exception`, I think it's an improvement.  And, I think using it is better than replacing `Exception` with `(Warning, RuntimeError, OSError)` all over the place, if only because it makes it easy to revisit our choices.  But, you're absolutely correct that it's not a proper solution...it's only a small improvement (one which addresses our issue with `AssertionError`).

> I doubt there are many low-level blocks of code that ought to be finalizing `RuntimeError` or even `OSError` -- the former because it's so loosely defined and the latter because it's so broad (essentially covering all of standard UNIX `errno` values).
> 
> If we're unlinking a file and it's OK if the file doesn't already exist, for example, we should be catching and finalizing `FileNotFoundError`, not `OSError`. If other OSError cases aren't expected, we shouldn't be finalizing them.

Before we disagree on this, we should probably look at some concrete cases.  I think there are a number (like the unlink example that you cited), where we are intent on all but ignoring the result of the operation.  So, we need a way to separate the exceptions into "infrastructure" (like `AssertError`) which we should allow to propagate, "failures" (like `FileNotFoundError` and most other file system errors) which we just eat, and "others" which we should probably just allow to propagate...but we might want to eat...I don't know.

portante:
Not going to pursue this at this time.

PR #2708, opened 2022-03-23
Use `black` on some additional files

portante:
We also apply `black` and `flake8` to them.

### DIFF
--- a/agent/bench-scripts/postprocess/user-benchmark-wrapper
+++ b/agent/bench-scripts/postprocess/user-benchmark-wrapper
@@ -1,13 +1,16 @@
 #!/usr/bin/env python3
 
-import sys, json, os.path, logging, shutil
+import json
+import logging
+import os.path
+import shutil
+import sys
 
 benchmark_run_dir = sys.argv[1]
 benchmark_duration = int(sys.argv[2])
 
-summary_file_name = os.path.join(benchmark_run_dir,
-        "user-benchmark-summary.json")
-with open(summary_file_name, 'a+') as data_file:
+summary_file_name = os.path.join(benchmark_run_dir, "user-benchmark-summary.json")
+with open(summary_file_name, "a+") as data_file:
     # We use the "a+" mode on open to ensure the file is created if it does not
     # already exist. But we have to seek to the beginning to make sure we read
     # it properly if it already exists.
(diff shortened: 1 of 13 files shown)


portante (commented):
Is this change worth making?  How do we find all the files `black .` should check?

This also leverages blacks default use of `.gitignore`, since it seems silly to repeat ignored files and sub-directories where we have to use a regex syntax to do so.

agent/bench-scripts/test-bin/mock-postprocess/dbench-postprocess
```diff
@@ -1,28 +1,29 @@
...
-#echo "dir=$1"
-#echo "iteration=$2"
-#echo "toollabelprefix=$3"
-#echo "toolgroup=$4"
-#"${count}-${protocol}_${test_type}-${message_size}B-${instance}i"
+# echo "dir=$1"
+# echo "iteration=$2"
+# echo "toollabelprefix=$3"
+# echo "toolgroup=$4"
+# "${count}-${protocol}_${test_type}-${message_size}B-${instance}i"
```
dbutenhof, line 18: I know you're just using `black` to reformat, but it's awfully tempting to ask if we really want to retain these commented-out statements.

I'm struggling to resist. 😆

pyproject.toml
```diff
@@ -7,16 +7,44 @@ build-backend="setuptools.build_meta"
...
 [tool.black]
 skip-string-normalization = false
-exclude = '''
+include = '''
+\.pyi?$
+| ^/agent/bench-scripts/postprocess/user-benchmark-wrapper$
+| ^/agent/bench-scripts/test-bin/mock-postprocess/dbench-postprocess$
+| ^/agent/bench-scripts/test-bin/mock-postprocess/uperf-postprocess$
+| ^/agent/bench-scripts/test-bin/java$
+| ^/agent/tool-scripts/datalog/haproxy-ocp-datalog$
```
dbutenhof, line 16: Well, dang; this makes the convention of `xxx.py` and softlink `xxx` look really good, because I hate to see all these special cases. I wish we could at least filter on something like `$(file ${name}|grep 'Python script')`

dbutenhof (approved):
I hate the special case patterns, but I like the ultimate effect.

pyproject.toml
portante: All ears if somebody can figure out how to get `black` to do that.

dbutenhof (approved):

dbutenhof (dismissed):

pyproject.toml
webbnh: We could consider consolidating this list by including directories, rather than specific files.

Or, we can wrap Black in a script which generates the value for the `--include` switch, using `find`, `file`, and `grep`.

And/or...we can reorganize our source tree so that certain directories only contain Python files (regardless of how they are named), and let Black loose on those without restrictions.

webbnh (commented):
I'll come back to this tomorrow....

npalaska (approved):

agent/bench-scripts/postprocess/user-benchmark-wrapper
```diff
@@ -18,31 +21,39 @@ with open(summary_file_name, 'a+') as data_file:
...
         # Either there was nothing in the file, or they didn't create the file.
-        logging.info("pbench-user-benchmark run didn't create a %s or the"
-                " file is empty, creating %s containing info about the run",
-                summary_file_name, summary_file_name)
+        logging.info(
+            "pbench-user-benchmark run didn't create a %s or the"
+            " file is empty, creating %s containing info about the run",
+            summary_file_name,
+            summary_file_name,
+        )
```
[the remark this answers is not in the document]
portante: I'd have to dive into the history, seems best for another PR (the dangers of taking on `black`).

pyproject.toml
portante: Seems best to address that level of work in a future PR.  Are you interested in pursuing that?

dbutenhof (approved):

dbutenhof (approved):

agent/tool-scripts/pcp
```diff
@@ -7,11 +7,16 @@ import os
...
-    print(f"{PROG}: This script is deprecated, please run it with --help for info on registering the tool.", file=sys.stderr)
-    print(f"{PROG}: Run /opt/pbench-agent/tool-scripts/pcp --help for more info.", file=sys.stderr)
+    print(
+        f"{PROG}: This script is deprecated, please run it with --help for info on registering the tool.",
+        file=sys.stderr,
+    )
+    print(
+        f"{PROG}: Run /opt/pbench-agent/tool-scripts/pcp --help for more info.",
+        file=sys.stderr,
+    )
```
webbnh, line 17: These look redundant...I think we would be better off with only one of them (probably the first).

webbnh (approved):
Looks good (enough) to me.

There's one line you might consider deleting...but perhaps that's outside the scope of this PR.  (I'll just go ahead an pre-resolve the comment.... 😝)

agent/tool-scripts/pcp
portante: And it is not deprecated, this is just a place-holder to fit the new "pcp" into the legacy framework.

portante: I'd rather address this is in future to refactor all this stuff.

PR #2711, opened 2022-03-25
Bring `BaseCommand` behaviors inline with `base`

portante:
In `lib/pbench/agent/base.py`, the `ssh_opts` and `scp_opts` fields were not being set correctly, and the `date_suffix` had the wrong formatting; we also create `pbench_tmp` using `pathlib`

### DIFF
--- a/lib/pbench/agent/base.py
+++ b/lib/pbench/agent/base.py
@@ -39,12 +39,13 @@ class BaseCommand(metaclass=abc.ABCMeta):
         # the pbench temporary directory is always relative to the $pbench_run
         # directory
         self.pbench_tmp = self.pbench_run / "tmp"
-        if not self.pbench_tmp.exists():
-            try:
-                os.makedirs(self.pbench_tmp)
-            except OSError:
-                click.secho(f"[ERROR] unable to create TMP dir, {self.pbench_tmp}")
-                sys.exit(1)
+        try:
+            self.pbench_tmp.mkdir(parents=True, exist_ok=True)
+        except Exception as exc:
+            click.secho(
+                f"[ERROR] unable to create TMP dir, '{self.pbench_tmp}': '{exc}'"
+            )
+            sys.exit(1)
 
         # log file - N.B. not a directory
         self.pbench_log = self.config.pbench_log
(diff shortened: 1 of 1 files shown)


agent/base
```diff
@@ -273,28 +274,28 @@ function vercmp {
...
+	error_log "unrecognized 'match' argument, '${_match}'"
+	return 1
+    fi
+    if [[ ${?} -ne 0 ]]; then
+	error_log "wrong version installed: found ${_package}-${_ins_ver}, expected ${_package}-${_exp_ver} (${_match})"
+	return 1
+    fi
+
+    debug_log "found ${_package}-${_ins_ver}"
     return 0
```
dbutenhof, line 299: You only "fixed" the control expression lines: this whole function was indented using tabs, and the lines in between are now just one tab instead of two; which, in "correct" GitHub rendering makes the visual indentation the same as the control lines.

And why *did* you change this file? So far as I can see, there's only the indentation changes, bracketed variables (which I like but I believe isn't essential in any of these contexts), and the addition of `local` declarations ... again, nice but not functional. If there are any "real" changes I'm afraid I lost them in the drizzle of red and green splotches. 😆 Not objecting (except that you should be consistent about the indentation)... but if I did miss anything "functional", please point it out!

portante: The changes here are posted in a separate commit.

The changes are all cosmetic, though the use of local declarations is arguably not cosmetic.

I'll post this as a separate PR.

portante: Pulled out in PR #2712.

dbutenhof (dismissed):

dbutenhof (dismissed):

lib/pbench/agent/base.py
```diff
@@ -65,16 +66,24 @@ def __init__(self, context):
...
             self.hostname = "testhost"
             self.full_hostname = "testhost.example.com"
         else:
-            self.date = datetime.datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%s")
-            self.date_suffix = datetime.datetime.utcnow().strftime("%Y-%m-%dT%H.%M.%s")
+            now = datetime.datetime.utcnow()
+            self.date = now.strftime("%Y-%m-%dT%H:%M:%s")
+            self.date_suffix = now.strftime("%Y.%m.%dT%H.%M.%s")
             self.hostname = socket.gethostname()
             self.full_hostname = socket.getfqdn()
```
webbnh, line 88: I don't see any [documentation](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes) for the `%s` format specifier; moreover, when I try to use it, I get junk.  Shouldn't that be `%S`??

Also, if I might, I would suggest:
```
        if os.environ.get("_PBENCH_UNIT_TESTS"):
            self.hostname = "testhost"
            self.full_hostname = "testhost.example.com"
            now = datetime.datetime(1900, 1, 1, tzinfo=datetime.timezone.utc)
        else:
            self.hostname = socket.gethostname()
            self.full_hostname = socket.getfqdn()
            now = datetime.datetime.utcnow()

        self.date = now.strftime("%Y-%m-%dT%H:%M:%s")
        self.date_suffix = now.strftime("%Y.%m.%dT%H.%M.%s")
```
or, possibly even something like
```
        ut = os.environ.get("_PBENCH_UNIT_TESTS")
        self.hostname = "testhost" if ut else socket.gethostname()
        self.full_hostname = "testhost.example.com" if ut else socket.getfqdn()
        now = datetime.datetime.utcnow()
        if ut:
            now = datetime.datetime(1900, 1, 1, tzinfo=datetime.timezone.utc)
        self.date = now.strftime("%Y-%m-%dT%H:%M:%s")
        self.date_suffix = now.strftime("%Y.%m.%dT%H.%M.%s")
```
(The fact is, I like single assignments, and, therefore, I like ternaries, even though Black is making it hard.)

webbnh (changes requested):
Looks OK, but I _think_ there's a problem with the `%s` format specifier in the `strftime()` call.  (I realize that it was not introduced by this change, but it looks like a bug which should be fixed and this PR is small enough that we could stand to fix it here.)

I also (of course) have some other suggestions.

lib/pbench/agent/base.py
webbnh: It looks like [`%s` is](https://man7.org/linux/man-pages/man3/strftime.3.html)
```
The number of seconds since the Epoch, 1970-01-01 00:00:00 +0000 (UTC).
```
which I'm pretty sure is _not_ what we want there.

portante: Nice catch.

dbutenhof (approved):

webbnh (approved):
👍

PR #2716, opened 2022-03-29
Initial Dashboard Code Base

MVarshini:
Merge pbench-dasboard repo code into `pbench/dashboard`.


dashboard/README.md
```diff
@@ -0,0 +1,70 @@
+# Getting Started with Create React App
```
dbutenhof, line 1: While all of this is useful, it's a README for the react npm application template generator rather than for the Pbench dashboard. Can we rename this so we can work towards a dashboard `README.md`?

dbutenhof (approved):
So this is strictly a "dashboard framework" with the toast alert mechanism and a main page to contain it. That's a perfectly fine starting point, although I'm curious why there's an empty `LoginComponent` rather than omitting it. That doesn't really matter, though, since I assume that Shubham will be replacing it soon with his code.

On the whole, I'm in favor of getting the dashboard framework into GitHub and refining it later. We should consider having a session where you walk the team through how this is structured and what the parts mean, possibly at the dashboard sync meeting.

Thanks!

dashboard/README.md
MVarshini: :100:  agree with you on this.
I'll start looking into updating this on further releases.
Creating a story to track this effort real quick ...

MVarshini: Here it is..
https://issues.redhat.com/browse/PBENCH-96

webbnh (approved):
As far as I can tell, this is all fine, with two possible exceptions:
- Any files (particularly code files) which are not going to be automatically (re-)generated in the future should end in a newline.  (There are quite a few missing.)
- What is the function or purpose of `dashboard/src/modules/containers/MainLayout/index.less`?...it's empty.  Should it be removed from the PR?

portante (approved):
I'd like to see this merged as is and fixes mentioned applied later in follow-up PRs.

PR #2719, opened 2022-03-31
New Dashboard codebase

shubham-html-css-js:
Code for New Dashboard


dashboard/.env
```diff
@@ -0,0 +1,7 @@
+REACT_APP_DATASETLIST=http://10.1.170.201/api/v1/datasets/list
```
dbutenhof, line 1: Although it's great to get something in, and we can refine it later, I'd prefer that you move quickly towards using `/api/v1/endpoints` rather than hardcoding access to my test server. Even if you're not ready to figure out how to leverage that off the initial page load (e.g., essentially `window.endpoints = fetch('/api/v1/endpoints')` in `index.html` as I'd had in the old dashboard), at least something more like `REACT_APP_SERVER=http://10.1.170.201` with (perhaps in data store initialization or at the `App.js` top level?), even if it's initially something simple like `endpoints = {REACT_APP_SERVER}/api/v1/endpoints`. Then the components would be accessing `endpoints.api.register`, `endpoints.api.login`, and so forth.

dbutenhof (commented):
I haven't gone through your redux and components in detail, but I'm going to throw out a first high-level comment on the configuration before I do that.

portante (changes requested):
Let's merge this after #2716 is merged.

portante:
Converted to `Draft` mode until it can be rebased on top of PR #2716.

portante:
No longer needed since most of the changes here have already been implemented.  It is still available for reference as a closed PR.

PR #2720, opened 2022-04-01
Turn on CodeQL analysis

portante:
This file is auto-generated by the GitHub button.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,70 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main, b0.69, b0.70 ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main ]
+  schedule:
+    - cron: '39 19 * * 5'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript', 'python' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
+        # Learn more about CodeQL language support at https://git.io/codeql-language-support
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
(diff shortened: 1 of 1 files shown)


dbutenhof (approved):
Intriguing that GitHub apparently shows the results of the scan in the Files Changed view rather than on the Conversation view...

portante (commented):
Seems like this is worth while to add now that we have moved the dashboard code back into this repo.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,70 @@
...
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main, b0.69, b0.70 ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main ]
+  schedule:
+    - cron: '39 19 * * 5'
```
webbnh, line 21: Every Friday at 3:39 pm EDT (Saturday at 1:09 am IST)?  That seems like an odd time.

I suppose that Friday/Saturday is good, but it seems like we should wait until after the work day is over.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,70 @@
...
+        languages: ${{ matrix.language }}
+        # If you wish to specify custom queries, you can do so here or in a config file.
+        # By default, queries listed here will override any specified in a config file.
+        # Prefix the list here with "+" to use these queries and those in the config file.
+        # queries: ./path/to/local/query, your-org/your-repo/queries@main
+
+    # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
+    # If this step fails, then you should remove it and run the build manually (see below)
+    - name: Autobuild
+      uses: github/codeql-action/autobuild@v2
```
webbnh, line 56: Do we want to comment this out, since we have no code in those languages, or is it sufficient to let Autobuild decide that for itself?

webbnh (dismissed):
Looks OK to me, although like Dave I'm struck by the presence of `web-server/v0.2/lineWithFocusChart.html` in this review....  I'm assuming that subsequent alerts will show up under the security tab, rather than being sprinkled throughout our PRs....

portante:
> Looks OK to me, although like Dave I'm struck by the presence of `web-server/v0.2/lineWithFocusChart.html` in this review...

Not sure what to think about this, other than that is very old code and reference support that should be removed in a separate PR.

> I'm assuming that subsequent alerts will show up under the security tab, rather than being sprinkled throughout our PRs....

Not sure what to think about this tool yet.  I figured we try and see how it works out.

.github/workflows/codeql-analysis.yml
portante: All this was auto-generated by "pressing the button" that generated this PR.

.github/workflows/codeql-analysis.yml
portante: Nice catch, we should comment this out.

portante (commented):
Applied code review comments.

dbutenhof (approved):

.github/workflows/codeql-analysis.yml
webbnh: What time was it when you pressed it?  😆

webbnh (approved):
GTG

PR #2722, opened 2022-04-02
Move to `perftool-incubator/bench-trafficgen`

portante:
The old TrafficGen repo has moved:

 * From: https://github.com/atheurer/trafficgen.git
 * To: https://github.com/perftool-incubator/bench-trafficgen.git

The new installed location is now `/opt/bench-trafficgen/trafficgen`.

### DIFF
--- a/agent/bench-scripts/pbench-trafficgen
+++ b/agent/bench-scripts/pbench-trafficgen
@@ -98,7 +98,7 @@ traffic_generator_def="trex-txrx" # can be either trex-txrx or moongen-txrx
 traffic_generator="${traffic_generator_def}"
 traffic_profiles=""
 if [[ -z "${trafficgen_dir}" ]]; then
-	trafficgen_dir="/opt/trafficgen"
+	trafficgen_dir="/opt/bench-trafficgen/trafficgen"
 fi
 trafficgen_version="unknown"
 trex_use_ht="n"
@@ -886,7 +886,7 @@ if [ "${postprocess_only}" == "n" ]; then
 			fi
 		fi
 	else
-		if pushd /opt >/dev/null && git clone https://github.com/atheurer/trafficgen.git; then
+		if pushd /opt >/dev/null && git clone https://github.com/perftool-incubator/bench-trafficgen.git; then
 			trafficgen_version=$(git log --max-count=1 --pretty=%h)
 			echo "trafficgen cloned"
 			popd >/dev/null


portante:
FYI @pradiptapks.

pradiptapks:
@portante @webbnh 

I have verified the latest `pbench-trafficgen.patch` successfully in `pbench-agent-0.69.9` rpm.
For reference, please find the configuration and results in detail.


**Environment:**
```
Red Hat Enterprise Linux release 8.4 (Ootpa)
Kernel: 4.18.0-305.40.2.el8_4.x86_64
```

**pbench-agent installed version**
```
# rpm -qa |grep pbench
pbench-perl-Types-Serialiser-1.0-1.noarch
pbench-agent-0.69.9-24g90314ee7c.noarch
pbench-perl-common-sense-3.74-1.x86_64
pbench-perl-JSON-XS-4.03-1.x86_64
pbench-sysstat-12.0.3-1.x86_64
```

**DPDK Version:**
```
# rpm -qa *dpdk*
dpdk-tools-20.11-3.el8.x86_64
dpdk-20.11-3.el8.x86_64
```

**Set Alias for DPDK Devbind:**
Note: From RHEL8.x, the `dpdk-devbind` command changed to `dpdk-devbind.py`. So manually updated the command details in `pbench-trafficgen`.

```
# grep dpdk-devbind /opt/pbench-agent/bench-scripts/pbench-trafficgen
                                if dpdk-devbind.py -u $dev; then
                                        dpdk-devbind.py -b vfio-pci $dev
                                dpdk-devbind.py -b vfio-pci $dev
```

**Install JSON schema**
`dnf install python3-jsonschema`

**Applying pbench-trafficgen patch:**
```
# cd /opt/pbench-agent/bench-scripts
# patch < /root/pbench-trafficgen.patch                                                                                                                                                                    
patching file pbench-trafficgen

# grep bench-trafficgen pbench-trafficgen
        trafficgen_dir="/opt/bench-trafficgen/trafficgen"
                if pushd /opt >/dev/null && git clone https://github.com/perftool-incubator/bench-trafficgen.git; then
```

**pbench-trafficgen command:**
`pbench-trafficgen --traffic-generator=trex-txrx --devices=0000:86:00.0,0000:86:00.1 --samples=1 --max-loss-pct=0.002 --config=Test --frame-sizes=64 --num-flows=128 --traffic-directions=bidirectional --flow-mods=src-ip --tool-period=binary-search --skip-git-pull --search-runtime=30 --validation-runtime=120 --rate=100 --rate-unit=% -- --send-teaching-measurement --send-teaching-warmup --teaching-measurement-packet-type=generic --teaching-warmup-packet-type=generic --disable-flow-cache`

webbnh (approved):
Thanks @pradiptapks!

For us Pbench folks,

> Set Alias for DPDK Devbind

Unfortunately, I don't think that DPDK gets installed implicitly or automatically...but, however it gets installed, there should probably be a symlink set up for `dpdk-devbind` so that the script doesn't have to be modified (although, I suppose we could augment the script to make it smart enough to figure out how to find the `.py` file if the naked command isn't there).  I'm surprised that the DPDK RPM doesn't do this already...it seems like changing the name of the command would be an incompatible change.

> `dnf install python3-jsonschema`

I've seen what happens if this is missing, but I haven't learned where the dependency for this is.  Is this something that the user just has to do (like installing DPDK), or should this be a dependency of the Pbench Agent RPM?  (Given that, apparently, nothing but the trafficgen support uses it, I would prefer that it not be added as a general dependency, but....)  If it's not going to be installed implicitly, we need to figure out how to fail more gracefully when it is missing!

npalaska (approved):

PR #2727, opened 2022-04-05
Ensure `${pbench_run}` directory exists

portante:
Fixes issue #2665.

We have to unconditionally create the `${pbench_run}` directory in `agent/base` (and `lib/pbench/agent/base.py` to maintain parallel behavior) because too many scripts depend on its existence for their correct function.

In particular, there are bench-scripts which try to report errors but if the `${pbench_run}` directory does not exist already, the `${pbench_log}` file location won't exist and the actual error is lost.

Rather than trying to fix the myriad number of places with this assumption, it is best to just create the directory to meet the assumed state.

### DIFF
--- a/agent/base
+++ b/agent/base
@@ -15,6 +15,13 @@ if [[ -z "${pbench_run}" ]]; then
     if [[ -z "${pbench_run}" ]]; then
         pbench_run=/var/lib/pbench-agent
     fi
+    if [[ ! -d ${pbench_run} ]]; then
+        mkdir ${pbench_run}
+        if [[ ${?} -ne 0 ]]; then
+            echo "[ERROR] unable to create pbench_run directory, ${pbench_run}" >&2
+            exit 1
+        fi
+    fi
     export pbench_run
 else
     if [[ ! -d ${pbench_run} ]]; then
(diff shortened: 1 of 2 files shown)


agent/base
```diff
@@ -15,6 +15,13 @@ if [[ -z "${pbench_run}" ]]; then
     if [[ -z "${pbench_run}" ]]; then
         pbench_run=/var/lib/pbench-agent
     fi
+    if [[ ! -d ${pbench_run} ]]; then
+        mkdir ${pbench_run}
```
dbutenhof, line 19: I think the asymmetry in the bash/python logic is unfortunate, but it doesn't appear there's any way to model Python's `exist_ok` in `mkdir` without `-p`, which also creates parents. We can probably count on `/var/lib` existing, so in the default case the distinction isn't important.

Note however that, prior to Riya's PR, we only created `${pbench_run}` when the environment variable wasn't previously defined, so that a user specifying their own run directory was explicitly on their own. Your new code doesn't do this; semantically, this means that if I `export pbench_run=~/my/fake/run`, our contract is now that I'm responsible for creating `~/my/fake` but Pbench will be happy to create `run`. That seems a bit weird: is that what we want to support?

Anyway, I'm not complaining, just commenting: I'm having a hard time thinking it's a big deal either way.

dbutenhof (approved):

agent/base
portante: > I think the asymmetry in the bash/python logic is unfortunate, but it doesn't appear there's any way to model Python's `exist_ok` in `mkdir` without `-p`, which also creates parents. We can probably count on `/var/lib` existing, so in the default case the distinction isn't important.

We could create a small python wrapper to expose the same semantics.

> Note however that, prior to Riya's PR, we only created `${pbench_run}` when the environment variable wasn't previously defined, so that a user specifying their own run directory was explicitly on their own. Your new code doesn't do this; semantically, this means that if I `export pbench_run=~/my/fake/run`, our contract is now that I'm responsible for creating `~/my/fake` but Pbench will be happy to create `run`. That seems a bit weird: is that what we want to support?

When `agent/base` is run in `pbench-list-tools` at the beginning, won't line 13 above see that `pbench_run` is defined and move to line 27 instead?

dbutenhof: Thanks; you're right, I didn't open the previous lines to check the context and I should have! This is all under the "we don't have an external definition of `pbench_run`", so we're good.

npalaska (approved):

webbnh (approved):
👍

PR #2728, opened 2022-04-05
Remove `perf-fio-0` test directory when successful

portante:


agent/tool-scripts/postprocess/unittests
```diff
@@ -117,6 +117,7 @@ function evaluate_performance_results {
             <<< "Duration: $(((end_time - start_time))) seconds"
         return 1
     else
+        rmdir ${_testroot}/$1
```
dbutenhof, line 120: Since we know the directory is empty at this point, what's the reason for not simply doing `rm -rf ${_testroot}` up at line 112? There would seem to be no value in retaining this empty directory for triage on error??

portante: Because in the failure case, we create a `result.txt` with the failure, so the `rm -rf` above keeps the parent directory.

dbutenhof (approved):

webbnh (approved):
👍

npalaska (approved):

PR #2730, opened 2022-04-06
Move abort() call to top-level API handling

### LINKED ISSUE #2529: Move `abort()` calls to the top-level API handling function
It does not seem wise to spread API handling behavior outside of the API interface method.  Raising exceptions instead of directly using `abort()` in these cases gives us the ability to see and review how the API responses are handled in one place.

_Originally posted by @portante in https://github.com/distributed-system-analysis/pbench/pull/2440#discussion_r743901867_

siddardh-ra:
As mentioned in [#2529](https://github.com/distributed-system-analysis/pbench/issues/2529) comments,
Implemented fix  for this `/datasets/contents` API request :

- Moved `abort()` to the top-level API handling function
- Implemented an `InvalidAPIRequest` Exception.
- Created multiple unit tests for this API based on various edge cases with run_id

Resolves #2529

### DIFF
--- a/lib/pbench/server/api/resources/__init__.py
+++ b/lib/pbench/server/api/resources/__init__.py
@@ -221,6 +221,19 @@ class ListElementError(SchemaError):
         return f"Unrecognized list value{'s' if len(self.bad) > 1 else ''} {self.bad!r} given for parameter {self.parameter.name}; expected {expected}"
 
 
+class APIAbort(Exception):
+    """
+    Used to report an error and abort if there is a failure in processing of API request.
+    """
+
+    def __init__(self, http_status: int, message: str = None):
+        self.http_status = http_status
+        self.message = message if message else HTTPStatus(http_status).phrase
+
+    def __str__(self) -> str:
+        return f"API error {self.http_status} : {self.message}"
+
+
 def convert_date(value: str, _) -> datetime:
     """
     Convert a date/time string to a datetime.datetime object.
@@ -1041,6 +1054,15 @@ class ApiBase(Resource):
             except SchemaError as e:
                 self.logger.exception("{}: SchemaError {}", api_name, e)
                 abort(e.http_status, message=str(e))
+            except APIAbort as e:
+                self.logger.exception("{} {}", self.__class__.__name__, e)
+                abort(e.http_status, message=e.message)
+            except Exception as e:
+                self.logger.exception("{} API error: {}", self.__class__.__name__, e)
+                abort(
+                    HTTPStatus.INTERNAL_SERVER_ERROR,
+                    message=HTTPStatus.INTERNAL_SERVER_ERROR.phrase,
+                )
 
         try:
             json_data = request.get_json()
@@ -1095,6 +1117,15 @@ class ApiBase(Resource):
         except SchemaError as e:
             self.logger.exception("{}: SchemaError {}", api_name, e)
             abort(e.http_status, message=str(e))
+        except APIAbort as e:
+            self.logger.exception("{} {}", self.__class__.__name__, e)
+            abort(e.http_status, message=e.message)
+        except Exception as e:
+            self.logger.exception("{} API error: {}", self.__class__.__name__, e)
+            abort(
+                HTTPStatus.INTERNAL_SERVER_ERROR,
+                message=HTTPStatus.INTERNAL_SERVER_ERROR.phrase,
+            )
 
     def _get(self, json_data: JSON, request: Request) -> Response:
         """
--- a/lib/pbench/server/api/resources/datasets_list.py
+++ b/lib/pbench/server/api/resources/datasets_list.py
@@ -5,10 +5,10 @@ from flask.wrappers import Request, Response
 
 from pbench.server import PbenchServerConfig, JSON
 from pbench.server.api.resources import (
-    ApiBase,
     API_OPERATION,
-    Parameter,
+    ApiBase,
     ParamType,
+    Parameter,
     Schema,
 )
 from pbench.server.database.database import Database
--- a/lib/pbench/server/api/resources/query_apis/__init__.py
+++ b/lib/pbench/server/api/resources/query_apis/__init__.py
@@ -11,12 +11,12 @@ from dateutil import rrule
 from dateutil.relativedelta import relativedelta
 from elasticsearch import Elasticsearch, helpers, VERSION
 from flask.wrappers import Response
-from flask_restful import abort
 import requests
 
 from pbench.server import PbenchServerConfig, JSON
 from pbench.server.api.auth import Auth
 from pbench.server.api.resources import (
+    APIAbort,
     API_OPERATION,
     ApiBase,
     Schema,
@@ -389,14 +389,10 @@ class ElasticBase(ApiBase):
                 return "", HTTPStatus.NO_CONTENT
         except UnauthorizedAccess as e:
             self.logger.warning("{}", e)
-            abort(e.http_status, message=str(e))
+            raise APIAbort(e.http_status, str(e))
         except KeyError as e:
             self.logger.exception("{} problem in preprocess, missing {}", klasname, e)
-            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
-        except Exception as e:
-            self.logger.exception("{} preprocess failed: {}", klasname, e)
-            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
-
+            raise APIAbort(HTTPStatus.INTERNAL_SERVER_ERROR)
         try:
             # prepare payload for Elasticsearch query
             es_request = self.assemble(json_data, context)
@@ -409,7 +405,7 @@ class ElasticBase(ApiBase):
             )
         except Exception as e:
             self.logger.exception("{} assembly failed: {}", klasname, e)
-            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
+            raise APIAbort(HTTPStatus.INTERNAL_SERVER_ERROR)
 
         try:
             # perform the Elasticsearch query
@@ -428,38 +424,37 @@ class ElasticBase(ApiBase):
                 e,
                 es_request,
             )
-            abort(
+            raise APIAbort(
                 HTTPStatus.BAD_GATEWAY,
-                message=f"Elasticsearch query failure {e.response.reason} ({e.response.status_code})",
+                f"Elasticsearch query failure {e.response.reason} ({e.response.status_code})",
             )
         except requests.exceptions.ConnectionError:
             self.logger.exception(
                 "{}: connection refused during the Elasticsearch request", klasname
             )
-            abort(
-                HTTPStatus.BAD_GATEWAY,
-                message="Network problem, could not reach Elasticsearch",
+            raise APIAbort(
+                HTTPStatus.BAD_GATEWAY, "Network problem, could not reach Elasticsearch"
             )
         except requests.exceptions.Timeout:
             self.logger.exception(
                 "{}: connection timed out during the Elasticsearch request", klasname
             )
-            abort(
+            raise APIAbort(
                 HTTPStatus.GATEWAY_TIMEOUT,
-                message="Connection timed out, could reach Elasticsearch",
+                "Connection timed out, could reach Elasticsearch",
             )
         except requests.exceptions.InvalidURL:
             self.logger.exception(
                 "{}: invalid url {} during the Elasticsearch request", klasname, url
             )
-            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
+            raise APIAbort(HTTPStatus.INTERNAL_SERVER_ERROR)
         except Exception as e:
             self.logger.exception(
                 "{}: exception {} occurred during the Elasticsearch request",
                 klasname,
                 type(e).__name__,
             )
-            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
+            raise APIAbort(HTTPStatus.INTERNAL_SERVER_ERROR)
 
         try:
             # postprocess Elasticsearch response
@@ -467,12 +462,11 @@ class ElasticBase(ApiBase):
         except PostprocessError as e:
             msg = f"{klasname}: the query postprocessor was unable to complete: {e}"
             self.logger.warning("{}", msg)
-            abort(e.status, message=msg, data=e.data)
+            raise APIAbort(e.status, str(e.message))
         except KeyError as e:
             self.logger.error("{}: missing Elasticsearch key {}", klasname, e)
-            abort(
-                HTTPStatus.INTERNAL_SERVER_ERROR,
-                message=f"Missing Elasticsearch key {e}",
+            raise APIAbort(
+                HTTPStatus.INTERNAL_SERVER_ERROR, f"Missing Elasticsearch key {e}"
             )
         except Exception as e:
             self.logger.exception(
@@ -481,7 +475,7 @@ class ElasticBase(ApiBase):
                 json_response,
                 e,
             )
-            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
+            raise APIAbort(HTTPStatus.INTERNAL_SERVER_ERROR)
 
     def _post(self, json_data: JSON, _) -> Response:
         """
@@ -627,14 +621,14 @@ class ElasticBulkBase(ApiBase):
         try:
             dataset = Dataset.query(name=json_data["name"])
         except DatasetNotFound as e:
-            abort(HTTPStatus.NOT_FOUND, message=str(e))
+            raise APIAbort(HTTPStatus.NOT_FOUND, str(e))
 
         owner = User.query(id=dataset.owner_id)
         if not owner:
             self.logger.error(
                 "Dataset owner ID {} cannot be found in Users", dataset.owner_id
             )
-            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="Dataset owner not found")
+            raise APIAbort(HTTPStatus.INTERNAL_SERVER_ERROR)
 
         # For bulk Elasticsearch operations, we check authorization against the
         # ownership of a designated dataset rather than having an explicit
@@ -646,7 +640,7 @@ class ElasticBulkBase(ApiBase):
         try:
             self._check_authorization(owner.username, dataset.access)
         except UnauthorizedAccess as e:
-            abort(e.http_status, message=str(e))
+            raise APIAbort(e.http_status, str(e))
 
         # Build an Elasticsearch instance to manage the bulk update
         elastic = Elasticsearch(self.elastic_uri)
@@ -734,7 +728,7 @@ class ElasticBulkBase(ApiBase):
                 type(e).__name__,
                 report,
             )
-            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
+            raise APIAbort(HTTPStatus.INTERNAL_SERVER_ERROR)
 
         summary = {"ok": count - error_count, "failure": error_count}
 
@@ -747,7 +741,7 @@ class ElasticBulkBase(ApiBase):
                 klasname,
                 type(e).__name__,
             )
-            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
+            raise APIAbort(HTTPStatus.INTERNAL_SERVER_ERROR)
 
         # Return the summary document as the success response, or abort with an
         # internal error if we weren't 100% successful. Some elasticsearch
@@ -766,12 +760,7 @@ class ElasticBulkBase(ApiBase):
                 error_count,
                 json.dumps(report),
             )
-            abort(
-                HTTPStatus.INTERNAL_SERVER_ERROR,
-                message=f"{error_count:d} of {count:d} Elasticsearch document actions failed",
-                data=summary,
-            )
-
+            raise APIAbort(HTTPStatus.INTERNAL_SERVER_ERROR, summary)
         self.logger.info(
             "{}:dataset {}: {} successful document actions", klasname, dataset, count
         )
--- a/lib/pbench/server/api/resources/query_apis/datasets/__init__.py
+++ b/lib/pbench/server/api/resources/query_apis/datasets/__init__.py
@@ -2,10 +2,9 @@ from http import HTTPStatus
 from logging import Logger
 from typing import AnyStr, Union, List
 
-from flask_restful import abort
-
 from pbench.server import PbenchServerConfig, JSON
-from pbench.server.api.resources import Schema, SchemaError
+from pbench.server.api.resources import APIAbort, Schema, SchemaError
+
 from pbench.server.api.resources.query_apis import CONTEXT, ElasticBase
 from pbench.server.database.models.datasets import (
     Dataset,
@@ -92,6 +91,9 @@ class RunIdBase(ElasticBase):
         object and run_id as JSON CONTEXT so that the postprocess operations
         can use it to identify the index to be searched from document index
         metadata.
+
+        Raises:
+        APIAbort: input can't be validated or normalized
         """
         run_id = client_json["run_id"]
 
@@ -99,16 +101,15 @@ class RunIdBase(ElasticBase):
         try:
             dataset = Dataset.query(md5=run_id)
         except DatasetNotFound:
-            self.logger.debug(f"Dataset with Run ID {run_id!r} not found")
-            abort(HTTPStatus.NOT_FOUND, message="Dataset not found")
-
+            raise APIAbort(
+                HTTPStatus.NOT_FOUND, f"No datasets with Run ID '{run_id!r}' found."
+            )
         owner = User.query(id=dataset.owner_id)
         if not owner:
             self.logger.error(
                 f"Dataset owner ID { dataset.owner_id!r} cannot be found in Users"
             )
-            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="Dataset owner not found")
-
+            raise APIAbort(HTTPStatus.INTERNAL_SERVER_ERROR)
         # We check authorization against the ownership of the dataset that
         # was selected rather than having an explicit "user"
         # JSON parameter. This will raise UnauthorizedAccess on failure.
@@ -125,15 +126,13 @@ class RunIdBase(ElasticBase):
         """
         try:
             index_map = Metadata.getvalue(dataset=dataset, key=Metadata.INDEX_MAP)
-        except MetadataError as e:
-            self.logger.error(f"Indices from metadata table not found {e!r}")
-            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
+        except MetadataError as exc:
+            self.logger.error("{}", str(exc))
+            raise APIAbort(HTTPStatus.INTERNAL_SERVER_ERROR)
 
         if index_map is None:
-            self.logger.error(
-                f"server.index-map not found in Metadata for a dataset {dataset!r}"
-            )
-            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
+            self.logger.error("Index map metadata has no value")
+            raise APIAbort(HTTPStatus.INTERNAL_SERVER_ERROR)
 
         index_keys = [key for key in index_map if root_index_name in key]
         indices = ",".join(index_keys)
--- a/lib/pbench/server/api/resources/query_apis/datasets/datasets_contents.py
+++ b/lib/pbench/server/api/resources/query_apis/datasets/datasets_contents.py
@@ -1,8 +1,6 @@
 from http import HTTPStatus
 from logging import Logger
 
-from flask_restful import abort
-
 from pbench.server import PbenchServerConfig
 from pbench.server.api.resources import (
     JSON,
@@ -65,14 +63,7 @@ class DatasetsContents(RunIdBase):
 
         # Retrieve the ES indices that belong to this run_id from the metadata
         # table
-
         indices = self.get_index(dataset, "run-toc")
-        if not indices:
-            self.logger.debug(
-                f"Found no indices matching the prefix run-toc"
-                f" for a dataset {dataset!r}"
-            )
-            abort(HTTPStatus.NOT_FOUND, message="Found no matching indices")
 
         return {
             "path": f"/{indices}/_search",
--- a/lib/pbench/server/api/resources/query_apis/datasets_detail.py
+++ b/lib/pbench/server/api/resources/query_apis/datasets_detail.py
@@ -1,10 +1,9 @@
 from http import HTTPStatus
 from flask import jsonify
-from flask_restful import abort
 from logging import Logger
 
 from pbench.server import PbenchServerConfig, JSON
-from pbench.server.api.resources import Schema, Parameter, ParamType
+from pbench.server.api.resources import APIAbort, ParamType, Parameter, Schema
 from pbench.server.api.resources.query_apis import (
     CONTEXT,
     ElasticBase,
@@ -158,12 +157,11 @@ class DatasetsDetail(ElasticBase):
         try:
             m = self._get_metadata(src["run"]["name"], context["metadata"])
         except DatasetNotFound:
-            abort(
-                HTTPStatus.BAD_REQUEST,
-                message=f"Dataset {src['run']['name']} not found",
+            raise APIAbort(
+                HTTPStatus.BAD_REQUEST, f"Dataset {src['run']['name']} not found"
             )
         except MetadataError as e:
-            abort(HTTPStatus.BAD_REQUEST, message=str(e))
+            raise APIAbort(HTTPStatus.BAD_REQUEST, str(e))
 
         if m:
             result["serverMetadata"] = m
--- a/lib/pbench/test/unit/server/query_apis/test_datasets_contents.py
+++ b/lib/pbench/test/unit/server/query_apis/test_datasets_contents.py
@@ -472,3 +472,18 @@ class TestDatasetsContents(Commons):
         drb = Dataset.query(name="drb")
         indices = self.cls_obj.get_index(drb, self.index_from_metadata)
         assert indices == "unit-test.v6.run-toc.2020-05"
+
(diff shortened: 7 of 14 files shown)


lib/pbench/server/api/resources/query_apis/datasets/__init__.py
```diff
@@ -127,14 +134,19 @@ def get_index(self, dataset: Dataset, root_index_name: AnyStr) -> AnyStr:
...
         if index_map is None:
             self.logger.error(
                 f"server.index-map not found in Metadata for a dataset {dataset!r}"
             )
-            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
-
+            raise InvalidAPIRequest(
+                HTTPStatus.INTERNAL_SERVER_ERROR,
+                f"server.index-map not found in Metadata for a dataset {dataset!r}",
+            )
```
dbutenhof, line 149: Not your fault since you're using the existing text here; but, really, the `MetadataError` handler up above is really "server.index-map key not found"; *this* case is "the key was found but has no useful value".

And while the text of these `logger.error` messages is OK because they're for our own internal consumption when analyzing a server problem, we shouldn't be confusing the external messages with this sort of internal detail.

Our convention has been that for "internal server errors" (where we really don't understand the exact cause in most cases), we've just said "INTERNAL ERROR" (leaving it to a developer or server admin to analyze the log journal); this message is too detailed to be useful, because most users won't have any idea what that means. This means that somehow either our database has become corrupt, or something went wrong when we indexed the dataset and we didn't catch it at the time; it's not a problem with the client's API query.

I think that the `self.logger.error` messages for these two cases (here and line 136) could be reversed, which would be a more accurate description; and the `InvalidAPIRequest` exceptions should go back to "INTERNAL ERROR".

lib/pbench/server/api/resources/query_apis/datasets/datasets_contents.py
```diff
@@ -65,14 +67,20 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
 
         # Retrieve the ES indices that belong to this run_id from the metadata
         # table
+        try:
+            indices = self.get_index(dataset, "run-toc")
+        except InvalidAPIRequest:
+            raise InvalidAPIRequest(
+                HTTPStatus.INTERNAL_SERVER_ERROR,
+                "Indices from metadata table not found",
+            )
```
dbutenhof, line 76: Let's not do this. You're *catching* one `InvalidAPIRequest` that was raised where the problem was discovered and raising a *new* `InvalidAPIRequest` further from the problem. That's counter-productive. We should just let the original exception propagate out to the external handler unless we really need to

1. Add logger information to help triage,
2. Raise a different exception to better frame the problem at this level,
3. Perform local cleanup before letting the exception propagate

None of these are true here.

lib/pbench/server/api/resources/query_apis/datasets/datasets_contents.py
```diff
@@ -65,14 +67,20 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
...
 
-        indices = self.get_index(dataset, "run-toc")
         if not indices:
             self.logger.debug(
                 f"Found no indices matching the prefix run-toc"
                 f" for a dataset {dataset!r}"
             )
-            abort(HTTPStatus.NOT_FOUND, message="Found no matching indices")
+            raise InvalidAPIRequest(HTTPStatus.NOT_FOUND, f"No matching indices found")
```
dbutenhof, line 83: Again, not really your problem since it already existed; but since `get_index` already raises `InvalidAPIRequest` if the metadata value is "falsey" there's no need for this block at all. (In fact, it should be impossible to get here, and I wonder why our tools aren't flagging that.)

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -390,6 +403,9 @@ def _call(self, method: Callable, json_data: JSON):
         except UnauthorizedAccess as e:
             self.logger.warning("{}", e)
             abort(e.http_status, message=str(e))
+        except InvalidAPIRequest as e:
+            self.logger.warning("{}", e)
+            abort(e.status, message=str(e))
```
dbutenhof, line 408: I think that this catch (and the definition of the exception) should be a layer further out, in `ApiBase._dispatch` (`lib/pbench/server/api/resources/__init__.py`). The various subclasses of that class (including this one) shouldn't need to provide handlers unless they need to clean up locally, which I don't think is the case here.

dbutenhof (changes requested):
Really good start, Siddardh! I'd like to see your fix cover *all* APIs, not just the "Elasticsearch" family under `query_apis`, and I think you've got some unnecessary catch/re-throw... but you're on the right track.

lib/pbench/server/api/resources/query_apis/datasets/datasets_contents.py
portante: So we can just remove the entire block from line 78 to 83?

Sounds good to me.

lib/pbench/server/api/resources/query_apis/datasets/__init__.py
```diff
@@ -100,15 +105,17 @@ def preprocess(self, client_json: JSON) -> CONTEXT:
             dataset = Dataset.query(md5=run_id)
         except DatasetNotFound:
             self.logger.debug(f"Dataset with Run ID {run_id!r} not found")
```
portante, line 107: Now that we are raising an exception here, should we drop the debug statement?  The same information is in the exception raised.

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -46,6 +46,19 @@ def __str__(self) -> str:
...
+    """
+    Used to report an Invalid API Request
+    """
+
+    def __init__(self, status: int, message: str, data: JSON = None):
+        self.status = status
+        self.message = message
+
+    def __str__(self) -> str:
+        return f"Invalid API request error returning {self.status} "
```
portante, line 59: Should this include the message value somehow?

portante (changes requested):
Looks good, just need to adjust the error handling as mentioned.

lib/pbench/server/api/resources/query_apis/__init__.py
dbutenhof: That's a really good point; I almost made a similar comment, but refrained. However, when we *catch* this, we do `abort(e.status, message=str(e))` ... which means the *message* is a repeat of the status along with the generic "Invalid API request" boilerplate. That's not useful at all.

We don't want either the boilerplate or the status in the `abort` `message=` key; we want the actual description, the `self.message` of the exception. First, we should probably be doing `abort(e.status, message=e.message)` instead of just string-ifying the exception at the catch point. However, the "Invalid API request" boilerplate is really a problem itself; it's not always an "invalid API request" ... it's a problem discovered in the API implementation. We don't need to expose those details to the client, regardless of the names or internal representation: we just want to report the status and message determined at the point where we raised the exception.

I'm not sure we need an `__str__` at all, but it's generally a good idea to have one. In this case, I'd suggest something more generic like f"API error '{self.message}': {self.status}".

And while I'm on the subject, I nearly suggested changing the exception name to `ApiError`... and I might as well throw that out since, again, "Invalid API request" suggests that the client did something wrong, which isn't necessarily the case. Or even `ApiAbort` since this is really our own replacement for Flask's (actually Werkzeug's) `abort` mechanism so we don't have to deal with their implicit semi-hidden exception outside the outer API implementation layer.

lib/pbench/server/api/resources/query_apis/datasets/__init__.py
```diff
@@ -100,15 +105,17 @@ def preprocess(self, client_json: JSON) -> CONTEXT:
...
+            )
         owner = User.query(id=dataset.owner_id)
         if not owner:
             self.logger.error(
                 f"Dataset owner ID { dataset.owner_id!r} cannot be found in Users"
             )
-            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="Dataset owner not found")
-
+            raise InvalidAPIRequest(
+                HTTPStatus.INTERNAL_SERVER_ERROR, "Dataset owner not found"
```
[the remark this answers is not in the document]
siddardh-ra: Yes, this seems to be a good point. I will update those changes in the next commit.

lib/pbench/server/api/resources/query_apis/__init__.py
siddardh-ra: Since there is a generic exception handler written for preprocess stage:

> `        except Exception as e:
>             self.logger.exception("{} preprocess failed: {}", klasname, e)
>             abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")`

Therefore all the exceptions that occurred during `preprocess()`, get handled here itself. So only I updated here instead of `ApiBase._dispatch`. 
But I get the context here, I will update the suggested changes accordingly

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -46,6 +46,19 @@ def __str__(self) -> str:
         return f"API {self.subclass_name} is missing schema parameters controller and/or name"
 
 
+class InvalidAPIRequest(Exception):
+    """
+    Used to report an Invalid API Request
+    """
+
+    def __init__(self, status: int, message: str, data: JSON = None):
+        self.status = status
```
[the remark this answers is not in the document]
dbutenhof: That's not a bad idea.

lib/pbench/server/api/resources/query_apis/__init__.py
dbutenhof: It's not a `SchemaError` ... let's not reuse base classes just because it's operationally convenient, unless they're really related.

If we want to do that, though, it could be a great idea to create a new base class which all API exceptions extend, building the concept of a return `http_status` into all of them by inheritance. That could cover this new exception along with `SchemaError`, `UnauthorizedAccess`, and `PostprocessError`.

lib/pbench/server/api/resources/query_apis/datasets/datasets_contents.py
webbnh: If we don't remove this block, then we Flake8 would like us to fix the f-string which contains no format parameters....

dbutenhof (changes requested):
Great; this is coming along nicely. But I'd like to see a couple more improvements. 😄

lib/pbench/server/api/resources/__init__.py
```diff
@@ -991,7 +1004,11 @@ def _dispatch(
             except UnauthorizedAccess as e:
                 self.logger.warning("{}", e)
                 abort(e.http_status, message=str(e))
-        return method(new_data, request)
+        try:
+            response_data = method(new_data, request)
+            return response_data
+        except APIAbort as e:
+            abort(e.http_status, message=str(e.message))
```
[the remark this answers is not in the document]
siddardh-ra: Whenever the `APIAbort` exception is raised, there is already `self.logger.exception(..)` triggered. In that case, It'll be a repetitive one. If needed we can add `self.logger.exception(..)` in the `APIAbort` as well .

lib/pbench/server/api/resources/query_apis/datasets/__init__.py
```diff
@@ -92,23 +94,25 @@ def preprocess(self, client_json: JSON) -> CONTEXT:
...
         run_id = client_json["run_id"]
 
         # Query the dataset using the given run id
         try:
             dataset = Dataset.query(md5=run_id)
         except DatasetNotFound:
-            self.logger.debug(f"Dataset with Run ID {run_id!r} not found")
-            abort(HTTPStatus.NOT_FOUND, message="Dataset not found")
-
+            raise APIAbort(
```
dbutenhof, line 107: FYI, this is one of the places where we don't have any `logger` message associated with *raising* `APIAbort`. I'd rather not depend on this; it's nice to have that handled in common at the `except APIAbort`. The only problem with that is the `INTERNAL ERROR` messages, which don't add a lot.

lib/pbench/server/api/resources/query_apis/datasets/__init__.py
```diff
@@ -126,15 +130,20 @@ def get_index(self, dataset: Dataset, root_index_name: AnyStr) -> AnyStr:
...
         except MetadataError as e:
-            self.logger.error(f"Indices from metadata table not found {e!r}")
-            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
-
-        if index_map is None:
             self.logger.error(
                 f"server.index-map not found in Metadata for a dataset {dataset!r}"
             )
-            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
+            raise APIAbort(
```
dbutenhof, line 136: And here we're doing a `logger.error` instead of `logger.exception` so we don't have the original cause. (Which could be a SQL error or something "interesting".)

lib/pbench/server/api/resources/__init__.py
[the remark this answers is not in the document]
dbutenhof: That's not really true, as I've pointed out in several other examples.

lib/pbench/server/api/resources/query_apis/datasets/__init__.py
webbnh: > Our convention has been that for "internal server errors" (where we really don't understand the exact cause in most cases), we've just said "INTERNAL ERROR" (leaving it to a developer or server admin to analyze the log journal)

I'd like to push back on this just a little bit.

Our convention for returning vague messages (e.g., `"INTERNAL ERROR"`) has mostly applied to security-related failures.

I have no problems with us returning detailed messages to the requestor (provided the contents are not security-sensitive), so long as it's under a `500`/`INTERNAL_ERROR` status.  That status indicates that the fault lies on the server side and that the requestor is not expected to be able to address it.  **_But_**, if we give a good message, then the requestor can relay that to us if it happens, and we might not _need_ to dive into the logs...which I think is worthwhile.

I don't think that any message we give under a `500` status will make the situation worse.  I don't see how it would  make the message "confusing" for the requestor or user.

lib/pbench/server/api/resources/__init__.py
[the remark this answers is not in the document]
webbnh: Per the [Flask docs](https://flask.palletsprojects.com/en/2.1.x/errorhandling/#handling):
> When Flask catches an exception while handling a request, it is first looked up by code. If no handler is registered for the code, Flask looks up the error by its class hierarchy; the most specific handler is chosen. If no handler is registered, [HTTPException](https://werkzeug.palletsprojects.com/en/2.1.x/exceptions/#werkzeug.exceptions.HTTPException) subclasses show a generic message about their code, while other exceptions are converted to a generic “500 Internal Server Error”.

So, other than for logging and clean-up requirements, we don't need to handle "internal errors" at all.

lib/pbench/server/api/resources/__init__.py
```diff
@@ -195,6 +195,19 @@ def __str__(self):
...
 
 
+class APIAbort(Exception):
+    """
+    Used to report an Invalid API Request
+    """
+
+    def __init__(self, http_status: int, message: str):
+        self.http_status = http_status
+        self.message = message
```
webbnh, line 205: As a suggestion, we could supply a default value for the message string based on the status value:
```
    def __init__(self, http_status: int, message: str = None):
        self.http_status = http_status
        self.message = message if message else HTTPStatus(http_status).phrase
```
In this way, we can declare an `APIAbort(500)` and it will automatically have a message of `'Internal Server Error'` without us having to spell it out.  (That is, when we don't have anything interesting to say, we don't have to say anything.)

lib/pbench/server/api/resources/query_apis/datasets/datasets_contents.py
```diff
@@ -1,16 +1,17 @@
...
     JSON,
     Schema,
     Parameter,
     ParamType,
 )
-from pbench.server.api.resources.query_apis import CONTEXT, PostprocessError
+from pbench.server.api.resources.query_apis import (
+    CONTEXT,
+    PostprocessError,
+)
```
webbnh, line 14: @siddardh-ra, could you try removing the trailing comma here as well?

lib/pbench/server/api/resources/query_apis/datasets/datasets_contents.py
```diff
@@ -65,14 +66,7 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
...
         indices = self.get_index(dataset, "run-toc")
-        if not indices:
-            self.logger.debug(
-                f"Found no indices matching the prefix run-toc"
-                f" for a dataset {dataset!r}"
-            )
-            abort(HTTPStatus.NOT_FOUND, message="Found no matching indices")
 
         return {
             "path": f"/{indices}/_search",
```
webbnh, line 72: Why don't we need to protect against an empty return from `get_index()`?

lib/pbench/test/unit/server/query_apis/test_datasets_contents.py
```diff
@@ -472,3 +472,39 @@ def test_get_index(self, attach_dataset, provide_metadata):
...
+        """
+        Check the Contents API when no run_id is provided
+        """
+        del self.payload["run_id"]
+        response = client.post(
+            f"{server_config.rest_uri}{self.pbench_endpoint}",
+            headers={"Authorization": "Bearer " + pbench_token},
+            json=self.payload,
+        )
+        assert response.status_code == HTTPStatus.BAD_REQUEST
```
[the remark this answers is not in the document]
webbnh: That's not a bad idea, @npalaska, but you code needs a small tweak, because both `""` and `None` are false-y, and we only want to delete the mapping if `run_id` is actually `None`.

webbnh (approved):
I have a couple of ideas for making `APIAbort` simpler to use and more useful.  Also, Flask can probably help us out in terms of logging and aborting on unhandled exceptions.

You have a couple of Flake8 complaints to deal with, and there are a couple of Black-isms which I think you can undo.  And, there are a couple of other small items (e.g., can we really get by without catching empty indices lists?).

lib/pbench/test/unit/server/query_apis/test_datasets_contents.py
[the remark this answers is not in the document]
npalaska: You are right, should be:
```
    if run_id is None:
        del self.payload["run_id"]
    else:
        self.payload["run_id"] = run_id
```

lib/pbench/server/api/resources/__init__.py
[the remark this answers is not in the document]
dbutenhof: Yeah, looks like it uses a `StreamHandler` to `stderr` by default, with `%s` formatting; so we'd need to do some default configuration work (e.g., with `dictConfig` in the documentation page you linked) in order to get Flask's logging into shape. All of that is way too much for this PR.

Going back to my initial comment here, we don't need the explicit `str` conversion on the `e.message`, and a `self.logger.exception` on an `Exception` that gets to here would probably be handy rather than completely losing the cause of the unexpected failure.

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -393,9 +393,6 @@ def _call(self, method: Callable, json_data: JSON):
         except KeyError as e:
             self.logger.exception("{} problem in preprocess, missing {}", klasname, e)
             abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
-        except Exception as e:
-            self.logger.exception("{} preprocess failed: {}", klasname, e)
-            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
```
dbutenhof, line 398: Basically, since we're removing this from the inner API layers, we should have it at the outer layer.

lib/pbench/server/api/resources/query_apis/__init__.py
```diff
@@ -393,9 +393,6 @@ def _call(self, method: Callable, json_data: JSON):
         except KeyError as e:
             self.logger.exception("{} problem in preprocess, missing {}", klasname, e)
             abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
```
dbutenhof, line 395: These `abort` calls should `raise APIAbort(HTTPStatus.INTERNAL_SERVER_ERROR)`.

lib/pbench/server/api/resources/query_apis/datasets/__init__.py
dbutenhof: I suppose that we could either special case `HTTPStatus.INTERNAL_SERVER_ERROR` or just generalize the exception to include separate `debug_message` and `user_message` properties, with `debug_message` defaulting to `user_message` if only one is specified. (And `user_message` defaulting to `"INTERNAL ERROR"` for `HTTPStatus.INTERNAL_SERVER_ERROR`.) The main question being whether we really would ever want separate messages *except* in that case. 👀

lib/pbench/server/api/resources/query_apis/datasets/datasets_contents.py
dbutenhof: Because `get_index` will throw `APIAbort` in that case, and we just let it propagate out to `ApiBase._dispatch`.

dbutenhof (commented):
A few comments, but I'd like to see you finish the remaining API branches so we can get this merged. This is coming along well.

dbutenhof (changes requested):
This looks good. I still think we should have a fallback `except Exception` around the `method` call in `ApiBase._dispatch` to be sure we get a `logger.exception` since we're not entirely sure what Flask will do with it on its own. I'm pretty sure we already have an open issue to investigate how we can better use Flask in this area; but that's outside the scope of this PR and I'd rather not leave that hole.

lib/pbench/server/api/resources/__init__.py
```diff
@@ -991,7 +1010,11 @@ def _dispatch(
             except UnauthorizedAccess as e:
                 self.logger.warning("{}", e)
                 abort(e.http_status, message=str(e))
-        return method(new_data, request)
+        try:
+            return method(new_data, request)
+        except APIAbort as e:
+            self.logger.exception("{} API call failed: {}", self.__class__.__name__, e)
```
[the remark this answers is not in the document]
siddardh-ra: @dbutenhof .Updated except Exception in ApiBase._dispatch and also formatted logger. exception too in the latest commit

dbutenhof (dismissed):
I've resolved all of my open conversations; I think many of the others will be resolved as @webbnh, @portante, and @npalaska review again. Good work!

lib/pbench/server/api/resources/query_apis/datasets_detail.py
```diff
@@ -158,12 +157,11 @@ def postprocess(self, es_json: JSON, context: CONTEXT) -> JSON:
...
         except DatasetNotFound:
-            abort(
-                HTTPStatus.BAD_REQUEST,
-                message=f"Dataset {src['run']['name']} not found",
+            raise APIAbort(
+                HTTPStatus.NOT_FOUND, f"Dataset {src['run']['name']} not found"
             )
         except MetadataError as e:
-            abort(HTTPStatus.BAD_REQUEST, message=str(e))
+            raise APIAbort(HTTPStatus.BAD_REQUEST, str(e))
```
webbnh, line 164: Why did we change from `BAD_REQUEST` to `NOT_FOUND` just for this one case?  (The other failure cases in this function return `BAD_REQUEST`....)  And, it's not that the _dataset_ is not found, right?  The issue is that the server-side metadata for this dataset is not found...isn't that weirder than `NOT_FOUND`?

lib/pbench/test/unit/server/query_apis/test_datasets_mappings.py
```diff
@@ -127,4 +127,4 @@ def test_with_db_error(self, client, server_config, database_error):
         with client:
             response = client.get(f"{server_config.rest_uri}/datasets/mappings/summary")
             assert response.status_code == HTTPStatus.INTERNAL_SERVER_ERROR
-            assert response.json["message"] == "Internal Server Error"
+            assert response.json["message"] == "INTERNAL ERROR"
```
webbnh, line 130: Should we be comparing against `HTTPStatus.INTERNAL_SERVER_ERROR.phrase`, like we do in the other test modules?

lib/pbench/server/api/resources/query_apis/datasets/__init__.py
```diff
@@ -125,15 +126,15 @@ def get_index(self, dataset: Dataset, root_index_name: AnyStr) -> AnyStr:
...
-            self.logger.error(f"Indices from metadata table not found {e!r}")
-            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
-
-        if index_map is None:
+        except MetadataError:
             self.logger.error(
                 f"server.index-map not found in Metadata for a dataset {dataset!r}"
             )
-            abort(HTTPStatus.INTERNAL_SERVER_ERROR, message="INTERNAL ERROR")
+            raise APIAbort(HTTPStatus.INTERNAL_SERVER_ERROR)
```
webbnh, line 133: We're not logging the exception, here...is there no information in the exception object or the traceback that we want to collect?  (And, in the message that we _are_ logging, we're **assuming** that the failure is for `server.index-map`, rather than using the actual input key (`Metadata.INDEX_MAP`) or the reported failure value (`MetadataError.key`)...which, presumably, _are_ `server.index-map`, but there's no guarantee that that will always be the case.)

webbnh (approved):
Looks good.

It would be good to keep the lists of imported references in alphabetical order -- this minimizes confusion and disruptions with future changes.

I'm sorry to see places where we are reporting less information, and I found one which looks like it should be fixed.

Other than those points, I have a couple of nits and a couple of questions, but overall the changes look really nice.

lib/pbench/server/api/resources/query_apis/datasets/__init__.py
dbutenhof: I wouldn't worry about the fact that we're doing a `logger.error` here, since the original `MetadataError` will be implicitly chained to the `APIAbort`, and `_dispatch` *does* `logger.exception`. I suppose we could almost argue that the `logger.error` here is redundant, although it does mention the dataset name which might not otherwise be clear.

In principle, yeah, it would have been cleaner to format `Metadata.INDEX_MAP` as a parameter rather than as a literal. I suppose that's my fault, along with the original bad language of the messages. 😄 

Do we need to improve this now?

lib/pbench/server/api/resources/__init__.py
```diff
@@ -954,7 +954,14 @@ def _dispatch(
         # We don't accept or process a request payload for GET, or if no
         # parameter schema is defined
         if not self.schema or method == self._get:
-            return method(uri_parameters, request)
+            try:
+                return method(uri_parameters, request)
+            except APIAbort as e:
+                self.logger.exception("{} {}", self.__class__.__name__, e)
+                abort(e.http_status, message=e.message)
```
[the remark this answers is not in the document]
dbutenhof: Oh, my, now you're getting really "meta". Really, the point is not some theoretical technicality about "most outerest outer level" (which doesn't matter to Flask), but only to get the `abort` calls out of low-level helper methods where the Werkzeug `HTTPException` is likely to be caught by an intermediate handler and result in an unintended internal server error.

`_dispatch` is the outermost layer with "meat", and the real "API orchestrator". Moving out to the `get`, `post`, `put`, `delete` methods would add redundancy to no actual purpose; I think this is right where it needs to be.

lib/pbench/server/api/resources/query_apis/datasets_detail.py
dbutenhof: You're right, and maybe this really should be an `HTTPStatus.INTERNAL_SERVER_ERROR` because I don't think it implies any problem at all in the client's inputs.

There are several more `BAD_REQUEST` errors up above (138, 143) in the same category. We presumably already validated the dataset name, made the Elasticsearch request, and we're complaining about the state or lack of Elastic data by blaming it on the client when that's clearly an internal problem.

So I'll agree that the change in this PR from `BAD_REQUEST` to `NOT_FOUND` is probably not what we want; however reverting that change isn't the answer, and I think the correct fix is that most of these should be changed to `INTERNAL_SERVER_ERROR`. I think that ought to be handled as a separate bug Jira rather than stretch out this PR.

npalaska (approved):
Looks like a pretty good cleanup now

lib/pbench/server/api/resources/query_apis/datasets_detail.py
webbnh: OK, let's not (unduly) stretch out this PR.  But, are we happy with changing this one `BAD_REQUEST` to `NOT_FOUND` or should it be reverted for this PR?

lib/pbench/server/api/resources/query_apis/datasets/__init__.py
webbnh: If `_dispatch` is going to log the entire exception chain, then clearly we don't need to log the `MetadataError` exception traceback here.  😅

My principal concern is that the stringized `MetadataError` exception object reports the dataset and key names (and sometimes other stuff), and I think we've lost that information with this change.  (The traceback doesn't include the exception object string, does it?).

So, I think we should either log the `MetadataError` exception string (e.g., instead of the current log message at line 131), or we should use that string in the `APIAbort` (and omit the `logger.error()` call here).

siddardh-ra:
> Looks good.
> 
> It would be good to keep the lists of imported references in alphabetical order -- this minimizes confusion and disruptions with future changes.
> 
> I'm sorry to see places where we are reporting less information, and I found one which looks like it should be fixed.
> 
> Other than those points, I have a couple of nits and a couple of questions, but overall the changes look really nice.

@webbnh .In the latest commit, I've addressed all the improvements and suggestions

lib/pbench/server/api/resources/datasets_list.py
```diff
@@ -7,12 +7,12 @@
 from pbench.server import PbenchServerConfig, JSON
 from pbench.server.api.resources import (
     ApiBase,
+    APIAbort,
```
dbutenhof, line 10: Hmmm... normally "P" is sorted before "p", but we already did it the other way here so you're "in fashion" if not necessarily "right". 😆

lib/pbench/server/api/resources/query_apis/datasets/__init__.py
```diff
@@ -128,7 +128,7 @@ def get_index(self, dataset: Dataset, root_index_name: AnyStr) -> AnyStr:
             index_map = Metadata.getvalue(dataset=dataset, key=Metadata.INDEX_MAP)
         except MetadataError:
             self.logger.error(
-                f"server.index-map not found in Metadata for a dataset {dataset!r}"
+                f"server.index-map not found in Metadata for a dataset {dataset!r} with key {Metadata.INDEX_MAP!r}"
```
dbutenhof, line 131: Now we're showing "server.index-map" *twice*; this should really be `f"Metadata key {Metadata.INDEX_MAP!r} not found for dataset {dataset!r}".

And really we shouldn't be pre-formatting f-strings with the `Logger`, so technically it should be

```
    self.logger.error("Metadata {} not found for dataset {}", Metadata.INDEX_MAP, dataset)
```

(The difference with `!r` here, in either form is that `str.__repr__` will add quotes around the key name, and the default `Dataset.__repr__` will be used instead of `Dataset.__str__` ... which I'm not actually sure I like, and I don't think the quotes are meaningful here.)

dbutenhof (approved):
I have a minor complaint about the changed `logger.error` message, but that can be fixed later if you get another approval.

lib/pbench/server/api/resources/datasets_list.py
webbnh: Moreover, normally "A" is sorted before "B", so on a case-blind basis these two lines should be reversed, as well.  🙂

lib/pbench/server/api/resources/query_apis/datasets_detail.py
```diff
@@ -158,7 +158,7 @@ def postprocess(self, es_json: JSON, context: CONTEXT) -> JSON:
             m = self._get_metadata(src["run"]["name"], context["metadata"])
         except DatasetNotFound:
             raise APIAbort(
-                HTTPStatus.NOT_FOUND, f"Dataset {src['run']['name']} not found"
+                HTTPStatus.BAD_REQUEST, f"Dataset {src['run']['name']} not found"
```
webbnh, line 161: Interesting that we can make this change without breaking any tests....  😏

webbnh (approved):
Looks good.

I am not inclined to insist on correcting the imports order further, unless you're going to make another change (like tweaking that pesky log message...).

lib/pbench/server/api/resources/query_apis/datasets_detail.py
```diff
@@ -158,12 +157,11 @@ def postprocess(self, es_json: JSON, context: CONTEXT) -> JSON:
         try:
             m = self._get_metadata(src["run"]["name"], context["metadata"])
         except DatasetNotFound:
-            abort(
-                HTTPStatus.BAD_REQUEST,
-                message=f"Dataset {src['run']['name']} not found",
+            raise APIAbort(
+                HTTPStatus.BAD_REQUEST, f"Dataset {src['run']['name']} not found"
```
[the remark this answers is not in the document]
dbutenhof: The `_get_metadata` method takes the dataset name and tries to look it up, so `DatasetNotFound` means the *dataset* isn't there. That's "impossible" at this point since we're in postprocess and we already found it in preprocess; but if it happens, this is now an internal error because something beyond the user's control happened. (Although, technically, maybe it's HTTPStatus.GONE (410), because it was there but someone else poached it with a DELETE in the middle of your query.)

A metadata access error would be `MetadataError` (handled below).

And, since this code was written, I ended up splitting `_get_metadata` as a wrapper (which looks up the dataset) around `_get_dataset_metadata` which takes the `Dataset` object instead of looking it up again. I believe we have that in the `context` object here, so we could really be using that ... not a big deal for this PR, however.

lib/pbench/server/api/resources/datasets_list.py
siddardh-ra: Oh okay. I thought the other way around

webbnh (approved):
👍

dbutenhof (approved):

npalaska (approved):

portante (approved):
Looks really good.  Just needs a rebase and conflict resolution.

dbutenhof (approved):

PR #2734, opened 2022-04-07
PBENCH-158 Wait for TDS to terminate gracefully

dbutenhof:
Instead of lobbing a "terminate" message and then almost immediately killing the TDS and Redis, wait (for up to 60 seconds) for the TDS to terminate on its own, giving it time to contact the remote clients and shut down gracefully.

PBENCH-158 #in-progress #time 3h

### DIFF
--- a/agent/util-scripts/pbench-tool-meister-start
+++ b/agent/util-scripts/pbench-tool-meister-start
@@ -172,8 +172,9 @@ assert _def_orchestrate_choice in _orchestrate_choices, (
 )
 
 # Wait at most 60 seconds for the Tool Data Sink to start listening on its
-# logging sink channel.
+# logging sink channel and then to exit after terminate message is sent.
 _TDS_STARTUP_TIMEOUT = 60
+_TDS_TERMINATE_TIMEOUT = 60
 
 
 class ReturnCode(BaseReturnCode):
@@ -513,6 +514,40 @@ class ToolDataSink(BaseServer):
                     continue
                 break
         return 0 if status == "success" else 1
+
+    @staticmethod
+    def is_running(pid: int) -> bool:
+        """Is the given PID running?
+
+        See https://stackoverflow.com/questions/7653178/wait-until-a-certain-process-knowing-the-pid-end
+
+        Return True if a PID is running, else False if not.
+        """
+        try:
+            os.kill(pid, 0)
+        except ProcessLookupError:
+            return False
+        return True
+
+    @staticmethod
+    def wait_for_pid(pid: int) -> bool:
+        """wait for a process to actually stop running, but eventually give up
+        to avoid a hang."""
+        running = True
+        timeout = time.time() + _TDS_STARTUP_TIMEOUT
+        while running:
+            running = __class__.is_running(pid)
+            if running:
+                if time.time() >= timeout:
+                    break
+                time.sleep(0.1)
+        return running
+
+    def shutdown_tds(self, status: int) -> None:
+        """Make sure TDS is shut down; wait for it to stop running on its own
+        and kill it if the wait times out."""
+        if self.wait_for_pid(self.get_pid()):
+            self.kill(status)
 
 
 class RedisServer(RedisServerCommon):
@@ -1095,7 +1130,8 @@ def main(_prog: str, cli_params: Namespace) -> int:
                 )
 
             recovery.add(
-                lambda: tool_data_sink.kill(ReturnCode.KEYBOARDINTERRUPT), "stop TDS"
+                lambda: tool_data_sink.shutdown_tds(ReturnCode.KEYBOARDINTERRUPT),
+                "stop TDS",
             )
 
         # We haven't yet started any remote clients; however, if we
--- a/lib/pbench/agent/utils.py
+++ b/lib/pbench/agent/utils.py
@@ -140,11 +140,18 @@ class BaseServer:
             self._repr = f"{self.name} - {self.host}:{self.port}"
 
         self.pid_file = None
+        self.pid = None
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return self._repr
 
-    def kill(self, ret_val: int):
+    def get_pid(self) -> int:
+        if not self.pid:
+            raw_pid = self.pid_file.read_text()
+            self.pid = int(raw_pid)
+        return self.pid
+
+    def kill(self, ret_val: int) -> int:
         """kill - attempt to KILL the running Redis server.
 
         This method is a no-op if the server instance isn't managed by us.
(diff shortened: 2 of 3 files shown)


portante (approved):

riya-17 (approved):

portante:
Bummer: `terminate_no_wait` has been missing an argument, `args`, for a long time: https://github.com/dbutenhof/pbench/blob/4bee10e7a8cdcebfa149d2afb9ce05a1b6f1810a/agent/util-scripts/pbench-tool-meister-start#L711

I tried this code with PR #2659 and it doesn't work.

dbutenhof:
> Bummer: `terminate_no_wait` has been missing an argument, `args`, for a long time: https://github.com/dbutenhof/pbench/blob/4bee10e7a8cdcebfa149d2afb9ce05a1b6f1810a/agent/util-scripts/pbench-tool-meister-start#L711
>
> I tried this code with PR #2659 and it doesn't work.

Dang. Because the missing `args: {"interrupt": ...}` actually breaks the terminate, or for some other reason? ☹️

portante:
The TDS is explicitly looking for it: https://github.com/dbutenhof/pbench/blob/4bee10e7a8cdcebfa149d2afb9ce05a1b6f1810a/lib/pbench/agent/tool_data_sink.py#L1226, and https://github.com/dbutenhof/pbench/blob/4bee10e7a8cdcebfa149d2afb9ce05a1b6f1810a/lib/pbench/agent/tool_data_sink.py#L1448.

dbutenhof:
> The TDS is explicitly looking for it: https://github.com/dbutenhof/pbench/blob/4bee10e7a8cdcebfa149d2afb9ce05a1b6f1810a/lib/pbench/agent/tool_data_sink.py#L1226, and https://github.com/dbutenhof/pbench/blob/4bee10e7a8cdcebfa149d2afb9ce05a1b6f1810a/lib/pbench/agent/tool_data_sink.py#L1448.

Indeed; but it's just logging a warning and continuing as if nothing happened, so we never noticed. Too much log volume! Which means I screwed up in testing, apparently not noticing that my nice timeout on the termination wait was firing. (There was no logger available at that level, so I didn't log it ... I knew I should have! Sigh.)

dbutenhof:
I guess this is worth fixing separately, so I'll post another PR and try to be more careful. Bummer.

portante:
First of all, this is not on you, this is on me, since I am responsible for this mess.

Second, until we get all this code move to the library side of the house and add unit tests, this kind of thing will keep happening.

So this is not on you at all.

dbutenhof:
The original mistake might have been yours, but that happens; it's complicated code. I was not just the last one to *use*/*touch* this code, but I broadened the use from one error exit to my generalized cleanup stack, which vastly increased the exposure.

Granted, the design of the code makes it easy to overlook the symptoms, since it's just a `warning` in the TDS error log; still, I won't accept any less responsibility than you. 😀 

I just started looking at this ... the "args" is easy to add, figuring out what value to use for "interrupt" is a bit messier since I need to look at the TDS to figure out exactly what this means (i.e., are any of the exit cases in `pbench-tool-meister-start` qualified as "interrupt", or not?)

But if you'd like to fix you, I'll hand the new Jira ticket over to you... just let's not both do it!

portante:
I have a fix coming ...

dbutenhof:
And, for the record, the `args["interrupt"]` value is logged into `metadata.log` as `run_interrupted: true`.

So I guess ideally `KeyboardInterrupt` would set a variable accessible to the `lambda` call that would be passed through... although if we interrupt `pbench-tool-meister-start` I'm not sure the `metadata.log` actually matters much!

dbutenhof:
BTW: My reference to the Jira ticket in this PR seems to have worked. Try putting PBENCH-163 in your git log header, and see if that's enough. (I also used a `PBENCH-158 #time 3h` command in the body of the log message, which also seemed to work; but I'm curious whether just the Jira ticket tag reference in the header is enough to make the connection.)

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -514,6 +515,40 @@ class ToolDataSink(BaseServer):
...
+        to avoid a hang."""
+        running = True
+        timeout = time.time() + _TDS_STARTUP_TIMEOUT
+        while running:
+            running = __class__.is_running(pid)
+            if running:
+                if time.time() >= timeout:
+                    break
+                time.sleep(0.1)
+        return running
```
webbnh, line 544: This code tests `running`, then overwrites it with a real value, and then tests that; if the process has already stopped running, then we test `running` one more time before we return.

I think we can improve this a bit (in case anyone who is working on this code _now_ is so inclined):
```
        timeout = time.time() + _TDS_STARTUP_TIMEOUT
        while time.time() < timeout:
            if not __class__.is_running(pid):
                return False
            time.sleep(0.1)
        return True
```
In exchange for front-loading the timeout test (and embracing a nested `return` statement), we can get rid of the `running` variable and only test its value once (in the case where the process has already exited).  (It's too bad that Python doesn't offer post-tested loops....)

lib/pbench/agent/utils.py
```diff
@@ -140,11 +140,18 @@ def __init__(self, spec: str, def_host_name: str):
...
-    def __repr__(self):
+    def __repr__(self) -> str:
         return self._repr
 
-    def kill(self, ret_val: int):
+    def get_pid(self) -> int:
+        if not self.pid:
+            raw_pid = self.pid_file.read_text()
+            self.pid = int(raw_pid)
+        return self.pid
```
webbnh, line 152: Since there's only one caller (and a single call?) to this function, is the caching of the return value really worthwhile?  (I like the encapsulation, though.)

lib/pbench/test/unit/agent/test_utils.py
```diff
@@ -97,17 +97,21 @@ def test_kill(self, tmp_path, monkeypatch):
...
             bs.kill(1)
 
-        bs = OurServer("localhost", "localhost")
         pidfile = tmp_path / "test.pid"
+
+        bs = OurServer("localhost", "localhost")
         pidfile.write_text("12345")
         bs.pid_file = pidfile
         ret = bs.kill(42)
         assert ret == 42
```
webbnh, line 106: Shouldn't we be mocking `OurServer.get_pid()` (or `BaseServer.get_pid()`), `os.kill()`, and possibly the `time` functions?

webbnh (commented):
@dbutenhof, I realize that I'm late to this game, but I have some comments nonetheless.

Since there is more work going on in these modules, these bear some consideration.

agent/util-scripts/pbench-tool-meister-start
dbutenhof: This is not a bad idea, and I intend to consolidate the nearly duplicate code from the start and stop sides anyway. Note, this started as just a direct lift from `pbench-tool-meister-stop`, except I was trying to merge with the old `self.kill()` call which meant either dropping the kill (since we'd never get there without a graceful exit) or adding a timeout. I decided to add a simple timeout, but I didn't worry much about the elegance since this was a patch to plug the current roof leak before I replace the whole roof, and I wanted a really quick turnaround because Riya's PR needed it.

lib/pbench/agent/utils.py
dbutenhof: There are actually two places we call `kill'; though the caching arguably *is* a bit excessive as they don't occur in the same process. It just seemed like the right thing to do. 🍤

lib/pbench/test/unit/agent/test_utils.py
dbutenhof: Yeah, well, I wouldn't even have touched this file except that I added that excessive pid caching, which broke the test as it re-used the `OurServer` object instance while changing `pid_file`. For some reason I just couldn't make myself go back to drop the `pid` caching, though, so I had to either add `bs.pid = None` (which I didn't like) or use a fresh `OurServer` for each case.

But you're right; mocking `os.kill` would have been good. I don't think I would want to mock `BaseServer.get_pid()`, however mocking the `Path.read_text` would avoid using a real file.

All things to consider in refactoring and improving the unit tests....

webbnh: > I wouldn't even have touched this file except that I added that excessive pid caching
 😀

> I don't think I would want to mock `BaseServer.get_pid()`

That depends on how you define the "unit" and how you want to inject the PID value into the CUT (i.e., you _could_ use a `Path.read_text()` mock for that, or you could mock out the _need_ for a `Path.read_text()` mock...).

> All things to consider in refactoring and improving the unit tests....

Yay!  Thank you....

PR #2741, opened 2022-04-11
Move logic out of pbench-tool-meister-* files

dbutenhof:
PBENCH-120 #time 2d

This is a first step in a process; we could call it a "task" against story PBENCH-120. My intention is to factor out some clean classes with methods that are easier to test; however to minimize confusion and complication in merges, this PR simply lifts the entire logic aside from the CLI trampoline into a separate module for each of the three Tool Meister control CLI commands, `pbench-tool-meister-start`, `pbench-tool-meister-client`, and `pbench-tool-meister-stop`.

(My intention is to do further refactoring and unit tests in a follow-up, rather than try to do unit tests to the current non-structure and then have to massively refactor them as well. In the meantime, the code continues to be covered by the legacy test suite as before.)

There are a few additional changes, as, having gotten my vscode into a virtual environment where it's no longer complaining about missing dependencies, it's suddenly taken to pointing out type errors that are apparently beneath the radar of flake8 and LGTM.com. I've only fixed a few here, most notably an error path in `Client.__enter__` which would have returned an integer status code rather than `self`. (Most of the others are more cosmetic, though intriguing, like "mismatches" between arrays declared as 'list[str]` vs `List[str]`.)


lib/pbench/agent/tool_meister_start.py
```diff
@@ -0,0 +1,1296 @@
...
+            " (NOTE: binding is not used with --orchestrate=existing);"
+        ),
+    )
+    parser.add_argument(
+        "tool_group",
+        help="The tool group name of tools to be run by the Tool Meisters.",
+    )
+    parsed = parser.parse_args()
+    status = start(sys.argv[0], parsed)
+    sys.exit(status)
```
webbnh, line 1296: This all (except for the call to `start()`, I suppose) seems like it should stay in the CLI script -- the entry to this module should have an API-style interface.

lib/pbench/agent/tool_meister_stop.py
```diff
@@ -0,0 +1,307 @@
...
+            " as well."
+        ),
+    )
+    parser.add_argument(
+        "tool_group",
+        help="The tool group name of tools being run in the Tool Meisters.",
+    )
+    parsed = parser.parse_args()
+    status = start(sys.argv[0], parsed)
+    sys.exit(status)
```
webbnh, line 307: This, too, should stay behind.

lib/pbench/agent/utils.py
```diff
@@ -254,7 +254,8 @@ def run_command(args, env=None, name=None, logger=None):
         return output
     except subprocess.CalledProcessError as e:
         message = "%s failed: %s" % (name, e.output)
-        logger.error(message)
+        if logger:
+            logger.error(message)
```
webbnh, line 258: I see why we need this test; my question is, why do we let `logger` be `None` instead of assigning it a default logger?

webbnh (approved):
I didn't give this a close review, as it's supposed to be a big lift-and-shift, but I did stumble across a few things.

First, the LGTM bot has a legitimate complaint as well as two other items which are, well, OK, legitimate.

Second, I think you shifted a little too much -- I think that the library modules should have API/Library-style (e.g., object-oriented) interfaces, leaving the CLI stuff in the `util-scripts` modules.

Finally, I have a pointed question about a bug that you fixed.

portante (changes requested):
@webbnh's comments about `sys.exit()` calls seem worth addressing.

agent/util-scripts/pbench-tool-meister-start
```diff
@@ -1,1329 +1,13 @@
...
-        ),
-    )
-    parser.add_argument(
-        "tool_group",
-        help="The tool group name of tools to be run by the Tool Meisters.",
-    )
-    parsed = parser.parse_args()
-    status = main(sys.argv[0], parsed)
-    sys.exit(status)
+sys.exit(main())
```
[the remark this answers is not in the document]
dbutenhof: I wonder if there's some way to run the LGTM checks before pushing a commit to GitHub; because while I now have a bunch of `vscode` analysis checks that *aren't* caught by `flake8` or by LGTM, this also shows LGTM checks not found by either of the others. This is annoying.

lib/pbench/agent/tool_meister_client.py
```diff
@@ -224,3 +228,76 @@ def terminate(self, group, interrupt=False):
...
+                "Unrecognized action, '{}', allowed actions are: {}",
+                action,
+                cli_tm_allowed_actions,
+            )
+            return 1
+        elif action == "kill":
+            # FIXME: we need to implement the gritty method of killing all the
+            # tool meisters, locally and remotely, and ensuring they are all
+            # properly shut down.
+            return 0
```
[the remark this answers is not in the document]
dbutenhof: Although these commands aren't (at least, yet) using `click`, this is a very click-like division, with the very simple CLI wrapper and all logic in a library module.

I like this logic, as the `pbench-tool-meister-start`/etc. modules aren't pytest-able. I suppose that whether we really want/need to test the command line parsing setup is debatable; but we could. In any case, if we do switch these to click, this is how it would look, and that makes some sense to me.

lib/pbench/agent/tool_meister_start.py
```diff
@@ -0,0 +1,1296 @@
...
+    for host, tm_proc in tms.items():
+        if tm_proc["status"] == "failed":
+            failures += 1
+            continue
+        elif tm_proc["status"] == "forked":
+            pid = tm_proc["pid"]
+            try:
+                exit_status = _waitpid(pid)
+            except Exception:
+                exit_status = 1
```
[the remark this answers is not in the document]
dbutenhof: I don't remember making this change. I wonder if it was a `vscode` complaint. It's very strict about "might be unassigned" references; but if it made such a complaint in this context, it seems that logic is flawed. Hmm.

lib/pbench/agent/utils.py
dbutenhof: I was trying to resolve the error (which probably doesn't happen for real) without a lot of coding work. I fixed another analyzer error in this module and this one was easy to "fix". I don't remember how many callers I found, but one was in a click module where this was called without referencing the logger it already has; I nearly changed that. It's possible we could change the signature here to require providing a logger without breaking anything; this change satisfied the scanner with a minimal footprint.

If you object to touching it without a more comprehensive fix, I can just drop this change; but I don't want to spread out further to "fix" it.

agent/util-scripts/pbench-tool-meister-stop
```diff
@@ -1,306 +1,13 @@
...
-        ),
-    )
-    parser.add_argument(
-        "tool_group",
-        help="The tool group name of tools being run in the Tool Meisters.",
-    )
-    parsed = parser.parse_args()
-    status = main(sys.argv[0], parsed)
-    sys.exit(status)
+sys.exit(main())
```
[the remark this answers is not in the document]
dbutenhof: This is odd; clearly I messed something up which neither vscode, flake8, nor unit tests discovered. (Which suggests that the legacy tests aren't provoking errors here, though that's hardly surprising.)

lib/pbench/agent/tool_meister_stop.py
dbutenhof: Actually, it *did* stay behind ... the problem is that it came along, too. 🍤 

Attack of the Clones.

lib/pbench/agent/tool_meister_start.py
dbutenhof: While I stopped short of converting these utility CLIs to click, the current division of labor is very click-like. I mean to refactor the real logic into a reusable class hierarchy, which will make the division between `main` and the library logic more evident.

lib/pbench/agent/tool_meister_client.py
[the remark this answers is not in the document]
portante: We could replace the logger calls with "interface" exceptions which the caller of main can just catch with a `try`/`except` block and do the honors of how to notify the CLI context.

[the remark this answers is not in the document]
dbutenhof: For the `.error` and `.exception` CLI log points, that might be a great idea. A lot of the tool meister logging is `.debug`, with some `.info` and `.warning`, none of which are terminal, so an exception model doesn't really work there.

We do need to move the agent away from using the logger for user feedback; though I'd rather not tackle that here.

agent/util-scripts/pbench-tool-meister-stop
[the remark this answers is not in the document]
webbnh: I wouldn't necessarily lay the blame on you.  IIRC, as the code was at the time, it wasn't so much that `main()` didn't _return a value_ as that `main()` _didn't return **at all**_....  😁  So, whether the above code was actually a problem devolves into a discussion about the noises that trees make when there is no one around to hear them.  😉

Anyway, it's fixed now.  🙂

lib/pbench/agent/utils.py
webbnh: Well, I don't like the code having the option of just dropping log/diagnostic output.  (And, I suspect, you might be sympathetic to that viewpoint....)  I _think_ I'd rather see an error/crash than have the output just dropped.

I support your desire to avoid spreading change...but, if `logger` is `None`, can't we just "get" a logger to use?  (I don't recall the details, but getting a logger in general is easy...perhaps I don't care too much about which one we get?)

lib/pbench/agent/tool_meister_client.py
```diff
@@ -224,3 +229,76 @@ def terminate(self, group, interrupt=False):
...
```
webbnh, line 287: Do we want to retain this?  Peter is in the process of implementing a `pbench-tools-kill` command which seems like it replaces this.

webbnh (dismissed):
Just one little question....

lib/pbench/agent/utils.py
portante: Another consideration is logging in utility functions.  If this is too low-level we may want this returned to the call to deal with as required.

portante: But for another PR.  =)

webbnh: > Another consideration is logging in utility functions.

This is a very good point...and it feeds into my growing ambivalence about moving away from the Python logger.  The Python logger is so versatile, and I'm kind of inclined to make it do what we want rather than moving away from it altogether...we just need some way to reconcile our "logging needs" with our "user interface requirements".  (That is, for CLI tools, we want the output to go to `stderr`, not a log somewhere, and we want it to be user-relevant, not "internal diagnostics"...unless, of course, the "user" is a Pbench developer....)

As you say, @portante, that's all for another PR.  For this PR, I guess I would vote for rescinding this change.  (I don't want us left in a state in which we quietly (and, perhaps, unexpectedly) eat diagnostics.)

portante: > (That is, for CLI tools, we want the output to go to `stderr`, not a log somewhere, and we want it to be user-relevant, not "internal diagnostics"...unless, of course, the "user" is a Pbench developer....)

Except that in many places the pbench-agent code emits logs to `stdout` or `stderr` AND to the `${pbench_log}` file.

portante (dismissed):
Let's get this merged for `v0.71` so that PR #2743 can leverage it.

lib/pbench/agent/utils.py
webbnh: Yeah...and, I wonder if that's really what we want (in the general case).  That is, typically, CLI tools send their output to `stdout` or `stderr`, and if the user wants to capture that s/he can redirect it (or use `tee` or `screen` or something).  If we want to produce log output in addition to that, we can certainly do so, but that's more like a developer function than a user function, so I wouldn't expect it to be enabled by default...and, for the most part, I wouldn't expect a single message to be delivered to _both_ places...but, if that's really what we want, the Python logger can be configured to do that sort of thing automagically, without having to set up a special utility function to make it happen.

npalaska (approved):
LGTM :+1:

lib/pbench/agent/utils.py
portante: Perhaps, but we don't want to introduce partial behavioral changes.  So we need to (in a separate PR) either move the code base to only `stdout` / `stderr` or ensure we have the correct loggers in place to do both.

Regardless, we need to over-haul the pbench-agent side logging.

webbnh: That will resolve this conversation.

dbutenhof: Yes, at the expense of yet another merge conflict when I still haven't figured out what went wrong with the last merge conflict. (Sigh.)

lib/pbench/agent/tool_meister_client.py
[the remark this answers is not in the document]
dbutenhof: Yes, with click the trivial launch wrapper is generated; but all of the parameter handling is inside the `main` called by that function, just as it is here. While my intent was to move towards a `main` that can just instantiate an object and tickle a method or two, the specific intent here was to minimize the code flow so it can be reviewed as a lift-and-shift. (And that's messy enough to deal with.)

lib/pbench/agent/tool_meister_start.py
```diff
@@ -0,0 +1,1366 @@
...
+   5. [orchestrate] Starting the local Tool Data Sink process
+   6. [orchestrate] Starting all the local and remote Tool Meisters
+   7. Waiting for the TDS to send a message reporting that it, and all the TMs,
+      started
+      - The TDS knows all the TMs that were started from the registered tools
+        data structure argument given to it
+   8. Verify all the requested Tool Meisters have reported back and that their
+      tool installation checks were successful
+
+There is a specific flow of data between these various components.  This		￼
```
portante, line 44: Looks like an extra line was added.

portante (changes requested):
Looks good, just an extra line added in a comment.

lib/pbench/agent/tool_meister_start.py
dbutenhof: Of course, there had to be something. Weird, though.

dbutenhof: This line, incidentally, came along with trailing whitespace that nobody complained about. I thought black didn't like that?

portante (approved):

npalaska (approved):

lib/pbench/agent/tool_meister_start.py
webbnh: Most of the tools are pretty tolerant of whitespace (particularly trailing) inside block comments.

PR #2743, opened 2022-04-12
Add the new `pbench-tools-kill` command

portante:
PBENCH-702

A new comprehensive command to hunt down and kill any and all Tool Meister sub-system components running locally or remotely, based on the recorded pbench "result" data directories found in the configured `${pbench_run}` directory hierarchy.

For example, if there are 5 pbench result directories found in `${pbench_run}`, and two of those results have left-over processes still running locally or remotely, this command will find and kill them all without prejudice.

The new `pbench.agent.tool_group.gen_tool_groups` helper module method is added to encapsulate searching for tool group data in result directories.

### DIFF
--- a/lib/pbench/agent/tool_group.py
+++ b/lib/pbench/agent/tool_group.py
@@ -2,7 +2,7 @@ import os
 from pathlib import Path
 import re
 import shutil
-from typing import Iterable
+from typing import Iterable, Optional
 
 
 class BadToolGroup(Exception):
@@ -20,7 +20,7 @@ class ToolGroup:
     TOOL_GROUP_PREFIX = "tools-v1"
 
     @staticmethod
-    def verify_tool_group(name: str, pbench_run: str = None) -> Path:
+    def verify_tool_group(name: str, pbench_run: Optional[str] = None) -> Path:
         """verify_tool_group - given a tool group name, verify it exists in the
         ${pbench_run} directory as a properly prefixed tool group directory
         name.
@@ -57,7 +57,7 @@ class ToolGroup:
             else:
                 return tg_dir
 
-    def __init__(self, name: str, pbench_run: str = None):
+    def __init__(self, name: str, pbench_run: Optional[str] = None):
         """Construct a ToolGroup object from the on-disk data of the given
         tool group.
 
@@ -182,6 +182,5 @@ def gen_tool_groups(pbench_run: str) -> Iterable[ToolGroup]:
     """
     for tg_dir in Path(pbench_run).glob(f"{ToolGroup.TOOL_GROUP_PREFIX}-*"):
         # All on-disk tool group directories will have names that look like
-        # above, where the prefix also contains a '-'.  To strip the prefix,
-        # we split the string and use the last component of the split.
-        yield ToolGroup(tg_dir.name.split("-", 2)[2], pbench_run)
+        # above.
+        yield ToolGroup(tg_dir.name[len(ToolGroup.TOOL_GROUP_PREFIX) + 1 :], pbench_run)
--- a/lib/pbench/agent/tool_group.py
+++ b/lib/pbench/agent/tool_group.py
@@ -76,7 +76,7 @@ class ToolGroup:
 
         Raises BadToolGroup via the verify_tool_group() method on error.
         """
-        self.tg_dir = ToolGroup.verify_tool_group(name, pbench_run)
+        self.tg_dir = self.verify_tool_group(name, pbench_run)
         self.name = name
 
         # __trigger__
--- a/lib/pbench/agent/tool_group.py
+++ b/lib/pbench/agent/tool_group.py
@@ -19,7 +19,7 @@ class ToolGroup:
     TOOL_GROUP_PREFIX = "tools-v1"
 
     @staticmethod
-    def verify_tool_group(name, pbench_run=None):
+    def verify_tool_group(name: str, pbench_run: str = None) -> Path:
         """verify_tool_group - given a tool group name, verify it exists in the
         ${pbench_run} directory as a properly prefixed tool group directory
         name.
@@ -56,7 +56,7 @@ class ToolGroup:
             else:
                 return tg_dir
 
-    def __init__(self, name):
+    def __init__(self, name: str, pbench_run: str = None):
         """Construct a ToolGroup object from the on-disk data of the given
         tool group.
 
@@ -75,7 +75,7 @@ class ToolGroup:
 
         Raises BadToolGroup via the verify_tool_group() method on error.
         """
-        self.tg_dir = self.verify_tool_group(name)
+        self.tg_dir = ToolGroup.verify_tool_group(name, pbench_run)
         self.name = name
 
         # __trigger__
--- a/lib/pbench/agent/tool_group.py
+++ b/lib/pbench/agent/tool_group.py
@@ -2,6 +2,7 @@ import os
 from pathlib import Path
 import re
 import shutil
+from typing import Iterable
 
 
 class BadToolGroup(Exception):
@@ -173,3 +174,14 @@ class ToolGroup:
             0
         """
         shutil.copytree(str(self.tg_dir), target_dir / self.tg_dir.name, symlinks=False)
+
+
+def gen_tool_groups(pbench_run: str) -> Iterable[ToolGroup]:
+    """Generate a series of ToolGroup objects for each on-disk tool group
+    found in the given pbench run directory.
+    """
+    for tg_dir in Path(pbench_run).glob(f"{ToolGroup.TOOL_GROUP_PREFIX}-*"):
+        # All on-disk tool group directories will have names that look like
+        # above, where the prefix also contains a '-'.  To strip the prefix,
+        # we split the string and use the last component of the split.
+        yield ToolGroup(tg_dir.name.split("-", 2)[2], pbench_run)
--- a/lib/pbench/cli/agent/commands/tools/kill.py
+++ b/lib/pbench/cli/agent/commands/tools/kill.py
@@ -1,0 +1,249 @@
+# -*- mode: python -*-
+
+"""Tool Meister "Kill"
+
+Module responsible for hunting down and stopping all Tool Meisters, the Tool
+Data Sink, and the Redis server orchestrated by pbench-tool-meister-start.
+
+This is a "big hammer" approach that is offered to users when they find the
+distributed system state of the Pbench Agent not working correctly.
+
+The algorithm is fairly straight-forward:
+
+  For each pbench run directory in ${pbench_run}
+  1. If the run was NOT orchestrated by pbench-tool-meister-start, ignore
+  2. Find the recorded pids for the Redis Server, Tool Data Sink, and local
+     Tool Meister in their respective pid files, and stop those processes
+     from running
+  3. Determine all the remote hosts used for that run
+  4. For each remote host:
+     a. `ssh` to that remote host
+     b. Stop the Tool Meister running on that host
+
+The pbench-tool-meister-start uses a UUID in the command line of all the
+remote Tool Meister processes. Any Tool Meister process with that UUID in its
+command line string will be stopped via `kill -KILL`, along with all of its
+child processes.
+"""
+
+from collections import defaultdict
+import pathlib
+import shlex
+from typing import Callable, List, Tuple
+
+import click
+import psutil
+
+from pbench.agent.base import BaseCommand
+from pbench.agent.tool_group import ToolGroup
+from pbench.agent.utils import LocalRemoteHost, TemplateSsh
+from pbench.cli.agent import CliContext, pass_cli_context
+from pbench.cli.agent.options import common_options
+
+
+def kill_family(proc: psutil.Process):
+    """Kill a parent process and all its children."""
+    try:
+        # Get the list of children of the parent before killing it.
+        children = list(proc.children())
+    except psutil.NoSuchProcess:
+        return
+    try:
+        proc.kill()
+    except psutil.NoSuchProcess:
+        pass
+    for child in children:
+        try:
+            child.kill()
+        except psutil.NoSuchProcess:
+            pass
+
+
+class PidSource:
+    """For a given PID file name keep track of discovered PIDs and UUIDs as
+    Tool Meister directories are `load()`ed.  The `killem()` method is invoked
+    by the caller at their discretion.
+
+    The `killem()` method clears out all accumlated data.
+    """
+
+    def __init__(self, file_name: str, display_name: str):
+        self.file_name = file_name
+        self.display_name = display_name
+        self.pids_by_uuid = {}
+        self.uuid_to_tmdir = {}
+
+    def load(self, tm_dir: pathlib.Path, uuid: str) -> bool:
+        """Load a PID from the given directory associated with the given UUID.
+
+        Records the loaded PID if it has a live process associated with it and
+        returns True, otherwise returns False.
+        """
+        try:
+            pid = (tm_dir / self.file_name).read_text()
+        except FileNotFoundError:
+            return False
+        try:
+            self.pids_by_uuid[uuid] = psutil.Process(pid)
+        except psutil.NoSuchProcess:
+            return False
+        self.uuid_to_tmdir[uuid] = tm_dir
+        return True
+
+    def killem(self, echo: Callable[[str], None]) -> None:
+        """Kill all PIDs found, and their children."""
+        if not self.pids_by_uuid:
+            return
+        echo(f"Killing {self.display_name} PIDs ...")
+        pids_by_uuid, self.pids_by_uuid = self.pids_by_uuid, {}
+        uuid_to_tmdir, self.uuid_to_tmdir = self.uuid_to_tmdir, {}
+        for uuid, proc in pids_by_uuid.items():
+            pid = proc.pid
+            echo(f"\tKilling {pid} (from {uuid_to_tmdir[uuid]})")
+            try:
+                kill_family(proc)
+            except Exception as exc:
+                echo(f"\t\terror killing {pid}: {exc}", err=True)
+
+
+def gen_run_directories(pbench_run: pathlib.Path) -> Tuple[pathlib.Path, str]:
+    """Generate the list of run directories available under ${pbench_run},
+    yielding a Path object for that directory, along with its recorded
+    UUID.
+
+    Yields a tuple of the run directory Path object and associated UUID.
+    """
+    for entry in pbench_run.iterdir():
+        if not entry.is_dir():
+            continue
+        tm_dir = entry / "tm"
+        try:
+            uuid = (tm_dir / ".uuid").read_text()
+        except FileNotFoundError:
+            # This is either not a pbench run directory, or the Tool Meister
+            # sub-system was not orchestrated by pbench-tool-meister-start
+            # for this run.
+            continue
+        yield tm_dir, uuid
+
+
+def gen_host_names(tm_dir: pathlib.Path) -> str:
+    """Read the registered tool data saved for this run and return the list
+    of remote hosts.
+    """
+    run_dir = tm_dir.parent
+    tool_group_dirs = list(run_dir.glob(f"{ToolGroup.TOOL_GROUP_PREFIX}-*"))
+    if not tool_group_dirs:
+        return
+
+    lrh = LocalRemoteHost()
+
+    for tool_group_dir in tool_group_dirs:
+        tg_name = tool_group_dir.name.split("-", 2)[2]
+        tg = ToolGroup(tg_name, run_dir)
+        for host_name_k in tg.hostnames.keys():
+            host_name = str(host_name_k)
+            if lrh.is_local(host_name):
+                continue
+            yield host_name
+
+
+class KillTools(BaseCommand):
+    """Find and stop all orchestrated Tool Meister instances."""
+
+    def execute(self, uuids: List[str]) -> int:
+        """Execute the tools kill operation.
+
+        If any UUIDs are passed as arguments, we only want to look for, and
+        locally kill, processes having those UUIDs.
+
+        Without command line arguemnts, kill all the local PIDs from all the
+        discovered run directories.
+
+        All the Redis server PIDs are killed first, then the Tool Data Sinks,
+        and finally the local Tool Meisters.
+
+        We then remotely kill (via `ssh`) all the Tool Meisters by invoking
+        this same command on a remote host with the list of UUIDs found across
+        all runs involving that host.
+        """
+        if uuids:
+            # We have a list of UUIDs to kill, implying that we search locally
+            # by UUID and only kill those PIDs found with the UUID in their
+            # registered command line.
+            for proc in psutil.process_iter():
+                # Consider each command line element.
+                for el in proc.cmdline():
+                    for uuid in uuids:
+                        if uuid in el:
+                            pid = proc.pid
+                            click.echo(f"\tKilling {pid} with UUID '{uuid}'")
+                            try:
+                                kill_family(proc)
+                            except Exception as exc:
+                                click.echo(f"\t\terror killing {pid}: {exc}", err=True)
+            return 0
+
+        # All three dictionaries for PID files that might be found, in the
+        # order in which we'll kill their PIDs.
+        all_pids = [
+            PidSource("redis.pid", "redis server"),
+            PidSource("pbench-tool-data-sink.pid", "tool data sink"),
+            PidSource("tm.pid", "local tool meister"),
+        ]
+        local_pids = False
+        remote_tms = defaultdict(list)
+        for tm_dir, uuid in gen_run_directories(self.pbench_run):
+            # If a run directory has any dangling components of the Tool
+            # Meister sub-system active, that is PID files for any local
+            # component, record those components.
+            for pidsrc in all_pids:
+                found = pidsrc.load(tm_dir, uuid)
+                if found:
+                    local_pids = True
+            # Find all the remotes for this run that need to be tracked down.
+            for host in gen_host_names(tm_dir.parent):
+                remote_tms[host].append(uuid)
+
+        if not local_pids and not remote_tms:
+            # No local or remote pids found, nothing to do.
+            return 0
+
+        # Kill all the local PIDs (and their children).
+
+        # We kill all the Redis Servers first in case killing them causes all
+        # the other processes to just exit on their own.  Then we kill all the
+        # Tool Data Sinks, and their children.  Then we kill all the (local)
+        # Tool Meisters, and their children.
+        for pidsrc in all_pids:
+            pidsrc.killem(click.echo)
+
+        # Kill all the remote Tool Meisters.
+
+        cmd = "pbench-tools-kill {{uuids}}"
+        template = TemplateSsh("ssh", shlex.split(self.ssh_opts), cmd)
+
+        # First fire off a number of background ssh processes, one per remote
+        # host.
+        remotes = []
+        for host, uuids in remote_tms.items():
+            click.echo(f"Killing all Tool Meister processes on remote host {host}...")
+            template.start(host, uuids=" ".join(uuids))
+            remotes.append(host)
+
+        # Wait for them all to complete, oldest to youngest.
+        for host in remotes:
+            template.wait(host)
+
+        return 0
+
+
+@click.command(
+    help="Ensure all instances of running tools are stopped locally or remotely"
+)
+@common_options
+@click.argument("uuids", nargs=-1)
+@pass_cli_context
+def main(ctxt: CliContext, uuids: List[str]):
+    status = KillTools(ctxt).execute(uuids)
+    click.get_current_context().exit(status)
--- a/lib/pbench/cli/agent/commands/tools/kill.py
+++ b/lib/pbench/cli/agent/commands/tools/kill.py
@@ -10,12 +10,12 @@ distributed system state of the Pbench Agent not working correctly.
 
 The algorithm is fairly straight-forward:
 
-  For each pbench run directory in ${pbench_run}
-  1. If the run was NOT orchestrated by pbench-tool-meister-start, ignore
+  For each pbench result directory in ${pbench_run}
+  1. If the result was NOT orchestrated by pbench-tool-meister-start, ignore
   2. Find the recorded pids for the Redis Server, Tool Data Sink, and local
      Tool Meister in their respective pid files, and stop those processes
      from running
-  3. Determine all the remote hosts used for that run
+  3. Determine all the remote hosts used for that result
   4. For each remote host:
      a. `ssh` to that remote host
      b. Stop the Tool Meister running on that host
@@ -107,12 +107,13 @@ class PidSource:
                 echo(f"\t\terror killing {pid}: {exc}", err=True)
 
 
-def gen_run_directories(pbench_run: pathlib.Path) -> Iterable[Tuple[pathlib.Path, str]]:
-    """Generate the list of run directories available under ${pbench_run},
-    yielding a Path object for that directory, along with its recorded
-    UUID.
+def gen_result_tm_directories(
+    pbench_run: pathlib.Path,
+) -> Iterable[Tuple[pathlib.Path, str]]:
+    """Generate the list of result directories available under ${pbench_run},
+    yielding a Path object for that directory, along with its recorded UUID.
 
-    Yields a tuple of the run directory Path object and associated UUID.
+    Yields a tuple of the result directory Path object and associated UUID.
     """
     for entry in pbench_run.iterdir():
         if not entry.is_dir():
@@ -121,19 +122,18 @@ def gen_run_directories(pbench_run: pathlib.Path) -> Iterable[Tuple[pathlib.Path
         try:
             uuid = (tm_dir / ".uuid").read_text()
         except FileNotFoundError:
-            # This is either not a pbench run directory, or the Tool Meister
-            # sub-system was not orchestrated by pbench-tool-meister-start
-            # for this run.
+            # This is either not a pbench result directory, or the Tool
+            # Meister sub-system was not orchestrated by
+            # pbench-tool-meister-start for this result.
             continue
         yield tm_dir, uuid
 
 
-def gen_host_names(tm_dir: pathlib.Path) -> Iterable[str]:
-    """Read the registered tool data saved for this run and return the list
-    of remote hosts.
+def gen_host_names(result_dir: pathlib.Path) -> Iterable[str]:
+    """Read the registered tool data saved for this result directory and
+    return the list of remote hosts.
     """
-    run_dir = tm_dir.parent
-    tool_groups = list(gen_tool_groups(run_dir))
+    tool_groups = list(gen_tool_groups(result_dir))
     if not tool_groups:
         return
 
@@ -156,7 +156,7 @@ class KillTools(BaseCommand):
         locally kill, processes having those UUIDs.
 
         Without command line arguments, kill all the local PIDs from all the
-        discovered run directories.
+        discovered result directories.
 
         All the Redis server PIDs are killed first, then the Tool Data Sinks,
         and finally the local Tool Meisters.  We kill all the Redis Servers
@@ -167,7 +167,7 @@ class KillTools(BaseCommand):
 
         We then remotely kill (via `ssh`) all the Tool Meisters by invoking
         this same command on a remote host with the list of UUIDs found across
-        all runs involving that host.
+        all results involving that host.
         """
         if uuids:
             # We have a list of UUIDs to kill, implying that we search locally
@@ -195,14 +195,15 @@ class KillTools(BaseCommand):
         ]
         local_pids = False
         remote_tms = defaultdict(list)
-        for tm_dir, uuid in gen_run_directories(self.pbench_run):
-            # If a run directory has any dangling components of the Tool
+        for tm_dir, uuid in gen_result_tm_directories(self.pbench_run):
+            # If a result directory has any dangling components of the Tool
             # Meister sub-system active, that is PID files for any local
             # component, record those components.  NOTE: we use a list
             # comprehension here to ensure that the .load() method is invoked
             # on each PidSource object.
             local_pids |= any([pidsrc.load(tm_dir, uuid) for pidsrc in all_pids])
-            # Find all the remotes for this run that need to be tracked down.
+            # Find all the remotes for this result that need to be tracked
+            # down.
             for host in gen_host_names(tm_dir.parent):
                 remote_tms[host].append(uuid)
 
--- a/lib/pbench/cli/agent/commands/tools/kill.py
+++ b/lib/pbench/cli/agent/commands/tools/kill.py
@@ -29,7 +29,7 @@ will be stopped via `kill -KILL`, along with all of its child processes.
 from collections import defaultdict
 import pathlib
 import shlex
-from typing import Callable, Dict, List, Tuple
+from typing import Callable, Dict, Iterable, List, Tuple
 
 import click
 import psutil
@@ -107,7 +107,7 @@ class PidSource:
                 echo(f"\t\terror killing {pid}: {exc}", err=True)
 
 
-def gen_run_directories(pbench_run: pathlib.Path) -> Tuple[pathlib.Path, str]:
+def gen_run_directories(pbench_run: pathlib.Path) -> Iterable[Tuple[pathlib.Path, str]]:
     """Generate the list of run directories available under ${pbench_run},
     yielding a Path object for that directory, along with its recorded
     UUID.
@@ -128,7 +128,7 @@ def gen_run_directories(pbench_run: pathlib.Path) -> Tuple[pathlib.Path, str]:
         yield tm_dir, uuid
 
 
-def gen_host_names(tm_dir: pathlib.Path) -> str:
+def gen_host_names(tm_dir: pathlib.Path) -> Iterable[str]:
     """Read the registered tool data saved for this run and return the list
     of remote hosts.
     """
--- a/lib/pbench/cli/agent/commands/tools/kill.py
+++ b/lib/pbench/cli/agent/commands/tools/kill.py
@@ -198,7 +198,9 @@ class KillTools(BaseCommand):
         for tm_dir, uuid in gen_run_directories(self.pbench_run):
             # If a run directory has any dangling components of the Tool
             # Meister sub-system active, that is PID files for any local
-            # component, record those components.
+            # component, record those components.  NOTE: we use a list
+            # comprehension here to ensure that the .load() method is invoked
+            # on each PidSource object.
             local_pids |= any([pidsrc.load(tm_dir, uuid) for pidsrc in all_pids])
             # Find all the remotes for this run that need to be tracked down.
             for host in gen_host_names(tm_dir.parent):
--- a/lib/pbench/cli/agent/commands/tools/kill.py
+++ b/lib/pbench/cli/agent/commands/tools/kill.py
@@ -159,7 +159,11 @@ class KillTools(BaseCommand):
         discovered run directories.
 
         All the Redis server PIDs are killed first, then the Tool Data Sinks,
-        and finally the local Tool Meisters.
+        and finally the local Tool Meisters.  We kill all the Redis Servers
+        first in case killing them causes all the other processes to just exit
+        on their own.  Then we kill all the Tool Data Sinks, and their
+        children.  Then we kill all the (local) Tool Meisters, and their
+        children.
 
         We then remotely kill (via `ssh`) all the Tool Meisters by invoking
         this same command on a remote host with the list of UUIDs found across
@@ -206,10 +210,6 @@ class KillTools(BaseCommand):
 
         # Kill all the local PIDs (and their children).
 
-        # We kill all the Redis Servers first in case killing them causes all
-        # the other processes to just exit on their own.  Then we kill all the
-        # Tool Data Sinks, and their children.  Then we kill all the (local)
-        # Tool Meisters, and their children.
         for pidsrc in all_pids:
             pidsrc.killem(click.echo)
 
--- a/lib/pbench/cli/agent/commands/tools/kill.py
+++ b/lib/pbench/cli/agent/commands/tools/kill.py
@@ -20,22 +20,22 @@ The algorithm is fairly straight-forward:
      a. `ssh` to that remote host
      b. Stop the Tool Meister running on that host
 
-The pbench-tool-meister-start uses a UUID in the command line of all the
-remote Tool Meister processes. Any Tool Meister process with that UUID in its
-command line string will be stopped via `kill -KILL`, along with all of its
-child processes.
+The pbench-tool-meister-start generates a UUID for the entire session and
+inserts that value into each command line of spawned remote Tool Meister
+processes. Any Tool Meister process with that UUID in its command line string
+will be stopped via `kill -KILL`, along with all of its child processes.
 """
 
 from collections import defaultdict
 import pathlib
 import shlex
-from typing import Callable, List, Tuple
+from typing import Callable, Dict, List, Tuple
 
 import click
 import psutil
 
 from pbench.agent.base import BaseCommand
-from pbench.agent.tool_group import ToolGroup
+from pbench.agent.tool_group import gen_tool_groups
 from pbench.agent.utils import LocalRemoteHost, TemplateSsh
 from pbench.cli.agent import CliContext, pass_cli_context
 from pbench.cli.agent.options import common_options
@@ -60,9 +60,9 @@ def kill_family(proc: psutil.Process):
 
 
 class PidSource:
-    """For a given PID file name keep track of discovered PIDs and UUIDs as
+    """For a given PID file name keep track of discovered Process-es and UUIDs as
     Tool Meister directories are `load()`ed.  The `killem()` method is invoked
-    by the caller at their discretion.
+    by the caller at its discretion.
 
     The `killem()` method clears out all accumlated data.
     """
@@ -70,8 +70,8 @@ class PidSource:
     def __init__(self, file_name: str, display_name: str):
         self.file_name = file_name
         self.display_name = display_name
-        self.pids_by_uuid = {}
-        self.uuid_to_tmdir = {}
+        self.procs_by_uuid: Dict[str, psutil.Process] = {}
+        self.uuid_to_tmdir: Dict[str, pathlib.Path] = {}
 
     def load(self, tm_dir: pathlib.Path, uuid: str) -> bool:
         """Load a PID from the given directory associated with the given UUID.
@@ -84,7 +84,7 @@ class PidSource:
         except FileNotFoundError:
             return False
         try:
-            self.pids_by_uuid[uuid] = psutil.Process(pid)
+            self.procs_by_uuid[uuid] = psutil.Process(pid)
         except psutil.NoSuchProcess:
             return False
         self.uuid_to_tmdir[uuid] = tm_dir
@@ -92,12 +92,13 @@ class PidSource:
 
     def killem(self, echo: Callable[[str], None]) -> None:
         """Kill all PIDs found, and their children."""
-        if not self.pids_by_uuid:
+        if not self.procs_by_uuid:
             return
         echo(f"Killing {self.display_name} PIDs ...")
-        pids_by_uuid, self.pids_by_uuid = self.pids_by_uuid, {}
+        # Clear out the stored data ahead of the killings.
+        procs_by_uuid, self.procs_by_uuid = self.procs_by_uuid, {}
         uuid_to_tmdir, self.uuid_to_tmdir = self.uuid_to_tmdir, {}
-        for uuid, proc in pids_by_uuid.items():
+        for uuid, proc in procs_by_uuid.items():
             pid = proc.pid
             echo(f"\tKilling {pid} (from {uuid_to_tmdir[uuid]})")
             try:
@@ -132,17 +133,14 @@ def gen_host_names(tm_dir: pathlib.Path) -> str:
     of remote hosts.
     """
     run_dir = tm_dir.parent
-    tool_group_dirs = list(run_dir.glob(f"{ToolGroup.TOOL_GROUP_PREFIX}-*"))
-    if not tool_group_dirs:
+    tool_groups = list(gen_tool_groups(run_dir))
+    if not tool_groups:
         return
 
     lrh = LocalRemoteHost()
 
-    for tool_group_dir in tool_group_dirs:
-        tg_name = tool_group_dir.name.split("-", 2)[2]
-        tg = ToolGroup(tg_name, run_dir)
-        for host_name_k in tg.hostnames.keys():
-            host_name = str(host_name_k)
+    for tg in tool_groups:
+        for host_name in tg.hostnames.keys():
             if lrh.is_local(host_name):
                 continue
             yield host_name
@@ -157,7 +155,7 @@ class KillTools(BaseCommand):
         If any UUIDs are passed as arguments, we only want to look for, and
         locally kill, processes having those UUIDs.
 
-        Without command line arguemnts, kill all the local PIDs from all the
+        Without command line arguments, kill all the local PIDs from all the
         discovered run directories.
 
         All the Redis server PIDs are killed first, then the Tool Data Sinks,
--- a/lib/pbench/cli/agent/commands/tools/kill.py
+++ b/lib/pbench/cli/agent/commands/tools/kill.py
@@ -195,10 +195,7 @@ class KillTools(BaseCommand):
             # If a run directory has any dangling components of the Tool
             # Meister sub-system active, that is PID files for any local
             # component, record those components.
-            for pidsrc in all_pids:
-                found = pidsrc.load(tm_dir, uuid)
-                if found:
-                    local_pids = True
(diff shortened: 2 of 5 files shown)


agent/util-scripts/unittests
```diff
@@ -150,7 +153,7 @@ function _save_tree {
 		  	  -e 's;\(tm\.[a-z-]*\)\.[0-9][0-9]*\.[a-z0-9_][a-z0-9_]*;\1.NNNNN.nnnnnnnn;g' \
         | sort >> ${_testout}
 
-    for x in $(find "${_testdir}/tools"-* "${_testdir}/tmp" -type f 2> /dev/null | sort); do
+    for x in $(find "${_testdir}/tools"-* "${_testdir}/tmp" "${_testdir}/run"-* -type f 2> /dev/null | sort); do
```
portante, line 156: Either remove or add this earlier.

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -0,0 +1,141 @@
+# -*- mode: python -*-
+
+"""Tool Meister "Kill"
+
+Module responsible for hunting down and stopping all Tool Meisters, the Tool
+Data Sink, and the Redis server orchestrated by pbench-tool-meister-start.
+
+This is a "big hammer" approach that is offered to users when they find the
+distributed system state of the Pbench Agent is not working correctly.
```
portante, line 9: `s/not//`

portante (commented):
Rough algorithm in place.

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -0,0 +1,174 @@
...
+Module responsible for hunting down and stopping all Tool Meisters, the Tool
+Data Sink, and the Redis server orchestrated by pbench-tool-meister-start.
+
+This is a "big hammer" approach that is offered to users when they find the
+distributed system state of the Pbench Agent not working correctly.
+
+The algorithm is fairly straight-forward:
+
+  For each pbench run directory in ${pbench_run}
+  1. If the run was NOT orchestrated pbench-tool-meister start, ignore
```
dbutenhof, line 14: "not orchestrated *by* pbench-tool-meister-start"?

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -0,0 +1,174 @@
...
+            except FileNotFoundError:
+                pass
+            else:
+                redis_pids[uuid] = psutil.Process(redis_pid)
+
+            if any(uuid in _pids for _pids in [tds_pids, tm_pids, redis_pids]):
+                self.uuid_to_rundir[uuid] = tm_dir
+
+        # Kill all the Redis Servers.  We do this first in case this causes
+        # all the other processes to just exit on their own.
```
dbutenhof, line 113: Maybe reword the comments to describe what we're doing here rather than what we expect the consumer to do with the values we yield?

portante (commented):
While this is a still a work in progress, needing unit tests and other PRs to land first, the logic and structure is essentially in place.

If you feel like reviewing, have at it, but will be making more changes as unit test implementation proceeds and as live functional tests are run.

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -0,0 +1,199 @@
...
+        lrh = LocalRemoteHost()
+
+        for tool_group_dir in tool_group_dirs:
+            tg_name = tool_group_dir.name.split("-", 2)
+            tg = ToolGroup(tg_name, run_dir)
+            for host_name_k in tg.hostnames.keys():
+                host_name = str(host_name_k)
+                if lrh.is_local(host_name):
+                    continue
+                yield (host_name)
```
portante, line 152: Parens not needed.

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -0,0 +1,199 @@
...
+        ]
+
+        for tm_dir, uuid in self.gen_run_directories():
+            # If a run directory has any dangling components of the Tool
+            # Meister sub-system active, that is PID files for any local
+            # component, record those components.
+            for pidsrc in all_pids:
+                pidsrc.load(tm_dir, uuid)
+
+        # Kill all the Redis Servers.  We do this first in case this killing
```
portante, line 127: `s/this//`

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -0,0 +1,199 @@
...
+        # them causes all the other processes to just exit on their own.  Then
+        # kill all the Tool Data Sinks, and their children.  Then kill all the
+        # (local) Tool Meisters, and their children.
+        for pidsrc in all_pids:
+            pidsrc.killem()
+
+    def gen_host_names(self, tm_dir: pathlib.Path):
+        """Read the registered tool data save for this run and return the list
```
portante, line 135: `s/save/saved/`

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -0,0 +1,199 @@
...
+        """
+        run_dir = tm_dir.parent
+        tool_group_dirs = list(run_dir.glob(f"{ToolGroup.TOOL_GROUP_PREFIX}-*"))
+        if not tool_group_dirs:
+            return
+
+        lrh = LocalRemoteHost()
+
+        for tool_group_dir in tool_group_dirs:
+            tg_name = tool_group_dir.name.split("-", 2)
```
portante, line 146: Need to index at `[2]` as well.

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -0,0 +1,199 @@
...
+            yield host, uuids
+
+    def kill_remote_pids(self):
+        """Remotely kill all the Tool Meisters on a given host bearing at least
+        one of the UUIDs in the given list.
+
+        We "ssh" to that remote host, find all the processes with one of the
+        UUIDs, and kill them.
+        """
+        ssh_opts = os.environ.get("ssh_opts", "")
```
portante, line 172: Remove, this should be `self.ssh_opts`.

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -0,0 +1,199 @@
...
+            for host in self.gen_host_names(tm_dir.parent):
+                remote_tms[host].append(uuid)
+        for host, uuids in remote_tms.items():
+            yield host, uuids
+
+    def kill_remote_pids(self):
+        """Remotely kill all the Tool Meisters on a given host bearing at least
+        one of the UUIDs in the given list.
+
+        We "ssh" to that remote host, find all the processes with one of the
```
portante, line 169: `s/with one/with any`

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -0,0 +1,199 @@
...
+        return True
+
+    def killem(self):
+        """Kill all PIDs found, and their children."""
+        if not self.pids_by_uuid:
+            return
+        print(f"Killing {self.display_name} PIDs ...")
+        for uuid, proc in self.pids_by_uuid:
+            print(f"\tKilling {proc.pid} (from {self.uuid_to_tmdir[uuid]})")
+            # First get the list of children so that we can kill the parent
```
portante, line 75: Or maybe "... so we can then kill the parent first to prevent it from creating any other children."

lib/pbench/test/unit/agent/test_tool_group.py
```diff
@@ -108,10 +108,11 @@ def test_target_dir(self, monkeypatch):
     class Test_ToolGroup:
         """Verify ToolGroup class"""
 
-        def mock_verify_tool_group(name: str, pbench_run: Path):
+        @staticmethod
```
portante, line 111: This corrects an existing bug.

portante:
Bugger all, two of the tests generate `AttributeError: __enter__` under Python 3.6.

lib/pbench/test/unit/agent/task/test_tools_kill.py
```diff
@@ -0,0 +1,440 @@
...
+                self.parent = MockPathGlob(parent)
+
+        class MockLocalRemoteHost:
+            """A simple mock for the LocalRemoteHost behaviors."""
+
+            def is_local(self, host_name: str):
+                return host_name.startswith("localhost")
+
+        # To verify the case where there are no tool groups created, we make a
+        # special Path object which will behavior like it has no tool groups.
```
ndokos, line 187: "... behave as if..."?

lib/pbench/test/unit/agent/task/test_tools_kill.py
```diff
@@ -0,0 +1,440 @@
...
+            assert hosts == MockToolGroup._expected_hosts
+
+    @staticmethod
+    def test_with_uuids():
+        """Verify UUID arguments work correctly.
+
+        The scenario we construct artificially starts with an invocation using
+        3 UUIDs, one won't be found.  The mocked-out `psutil.process_iter()`
+        will generate 3 PIDs, one of which will have child processes, which
+        will all operate normally with no expections.  Of the other two, one
```
ndokos, line 209: "exceptions"

ndokos (approved):
LGTM -  the tests pass on my laptop, but not in Jenkins.  I found a few typos and made a suggestion for a renaming.

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -0,0 +1,249 @@
...
+        self.uuid_to_tmdir[uuid] = tm_dir
+        return True
+
+    def killem(self, echo: Callable[[str], None]) -> None:
+        """Kill all PIDs found, and their children."""
+        if not self.pids_by_uuid:
+            return
+        echo(f"Killing {self.display_name} PIDs ...")
+        pids_by_uuid, self.pids_by_uuid = self.pids_by_uuid, {}
+        uuid_to_tmdir, self.uuid_to_tmdir = self.uuid_to_tmdir, {}
```
dbutenhof, line 99: This "fetch and clear" idiom is cute but a bit obscure: I'd rather see these broken up, I think!

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -0,0 +1,249 @@
...
+    """
+    run_dir = tm_dir.parent
+    tool_group_dirs = list(run_dir.glob(f"{ToolGroup.TOOL_GROUP_PREFIX}-*"))
+    if not tool_group_dirs:
+        return
+
+    lrh = LocalRemoteHost()
+
+    for tool_group_dir in tool_group_dirs:
+        tg_name = tool_group_dir.name.split("-", 2)[2]
```
dbutenhof, line 142: A comment describing the semantics of your `<x>-<y>-<tool-group>` pattern would be a help here! `tools-v1-<name>` may not be obvious to casual browsers here...

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -0,0 +1,249 @@
...
+        tg = ToolGroup(tg_name, run_dir)
+        for host_name_k in tg.hostnames.keys():
+            host_name = str(host_name_k)
```
dbutenhof, line 145: I had to do some digging to figure out why we might need to `str`-ify the "host name". Apparently the type is `byte` *iff* the path name you give `os.listdir` is a byte string. Otherwise it's `str` ... is there any particular reason this might be a byte in our case, or to be that paranoid? If so, a comment would be helpful.

(And, in fact, if we really do need to worry, but it's OK to stringify as we do here, then using `str(name)` as the key in the first place when constructing the dict would eliminate any need to worry about it elsewhere.)

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -0,0 +1,249 @@
...
+class KillTools(BaseCommand):
+    """Find and stop all orchestrated Tool Meister instances."""
+
+    def execute(self, uuids: List[str]) -> int:
+        """Execute the tools kill operation.
+
+        If any UUIDs are passed as arguments, we only want to look for, and
+        locally kill, processes having those UUIDs.
+
+        Without command line arguemnts, kill all the local PIDs from all the
```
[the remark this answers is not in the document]
dbutenhof: Why do we argue with our command lines, anyway?

lib/pbench/test/unit/agent/task/test_tools_kill.py
dbutenhof: Or "will expose no tool groups"? The "behave as if" somehow implies it does have tool groups, but lies. 👀

lib/pbench/test/unit/agent/task/test_tools_kill.py
```diff
@@ -0,0 +1,440 @@
...
+            result = runner.invoke(kill.main)
+        assert (
+            result.exit_code == 0
+        ), f"stdout={result.stdout!r}, stderr={result.stderr!r}"
+        assert (
+            result.stdout == ""
+        ), f"stdout={result.stdout!r}, stderr={result.stderr!r}"
+
+    @staticmethod
+    def test_without_uuids_wi_action():
```
dbutenhof, line 328: `with_action`?

dbutenhof (changes requested):
I know it's been a long time since we saw this before, but it would have been nice to see the changes as a separate commit for context.

lib/pbench/cli/agent/commands/tools/kill.py
portante: Perhaps it would be better to have a `staticmethod` of the ToolsGroup class?

lib/pbench/cli/agent/commands/tools/kill.py
portante: This is such a common python code sequence that breaking it up would be an afront to the language.  =)  I certainly can add a comment.

lib/pbench/cli/agent/commands/tools/kill.py
portante: The use of `str` is not needed.

lib/pbench/cli/agent/commands/tools/kill.py
dbutenhof: Sure, the syntax is common, and the semantics here are clever but I think unnecessarily obscured by the common syntax. But whatever: it's a small section, and it works.

lib/pbench/test/unit/agent/test_tool_group.py
```diff
@@ -269,3 +270,52 @@ def mock_read_text(path):
...
+
+def test_gen_tool_groups():
+    """Verify the tool group directory generator."""
+
+    class MockPath:
+        """A simple ToolGroup mock to generate host names from different
+        tool group directories.
+        """
+
+        _groups = set(("heavy", "light", "medium"))
```
ndokos, line 283: "_groups" -->  "_sets" (or perhaps "_toolsets" to avoid ambiguity)?

ndokos (dismissed):
One renaming suggestion.

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -132,17 +133,14 @@ def gen_host_names(tm_dir: pathlib.Path) -> str:
     of remote hosts.
     """
     run_dir = tm_dir.parent
-    tool_group_dirs = list(run_dir.glob(f"{ToolGroup.TOOL_GROUP_PREFIX}-*"))
-    if not tool_group_dirs:
+    tool_groups = list(gen_tool_groups(run_dir))
```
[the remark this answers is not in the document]
portante: One should not care about the other, right?  I don't want to influence the interface by this one use case.

The other use case was elusive to find (see https://github.com/distributed-system-analysis/pbench/blob/main/lib/pbench/cli/agent/commands/tools/base.py#L28) and that is a property.  It might be better to remove `gen_tool_groups` and fix the property to do it correctly (not hard-coding the prefix).

And thanks for opening this can of worms.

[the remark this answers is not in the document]
dbutenhof: Oh, wow, and, 

```
            [p.name.split("tools-v1-")[1] for p in self.pbench_run.glob("tools-v1-*")]
```

Using the entire tools/version string as the split separator?

And, yes; another use case, but also immediately list-ifying the results, so not a demonstration of a "real" generator.

While I agree that the method API shouldn't depend on a single caller pattern, we also don't want a design that essentially forces a common caller "mitigation" pattern everyone now "needs to know about" like list-ifying. And here, for example, we sort in addition: should the "generator" be sorting the output at the source?

lib/pbench/agent/tool_group.py
```diff
@@ -173,3 +174,14 @@ def archive(self, target_dir: Path):
...
+
+def gen_tool_groups(pbench_run: str) -> Iterable[ToolGroup]:
+    """Generate a series of ToolGroup objects for each on-disk tool group
+    found in the given pbench run directory.
+    """
+    for tg_dir in Path(pbench_run).glob(f"{ToolGroup.TOOL_GROUP_PREFIX}-*"):
+        # All on-disk tool group directories will have names that look like
+        # above, where the prefix also contains a '-'.  To strip the prefix,
+        # we split the string and use the last component of the split.
+        yield ToolGroup(tg_dir.name.split("-", 2)[2], pbench_run)
```
webbnh, line 187: If what we want is the "last component", then maybe the index should be `-1` instead of `2`.

However, it would be better to just remove the prefix (e.g., via [`str.removeprefix()`](https://docs.python.org/3/library/stdtypes.html#str.removeprefix)...unfortunately, that's new in 3.9, but you could use something like `string[len(ToolGroup.TOOL_GROUP_PREFIX) + 1:]`).

And, I would reword the comment to read something like
> Extract the tool group name from the directory name by removing the group prefix.

lib/pbench/test/unit/agent/test_tool_group.py
```diff
@@ -1,9 +1,10 @@
 import os
 from pathlib import Path
+from unittest import mock
```
webbnh, line 3: Why are we adding `unittest.mock` instead of using PyTest's `monkeypatch` fixture?

(`unittest` looks very cool, but I'm concerned about having more than one way to do things sewing confusion.)

lib/pbench/test/unit/agent/test_tool_group.py
```diff
@@ -268,3 +270,52 @@ def mock_read_text(path):
...
+
+        def glob(self, prefix: str):
+            if self._name == "no-tool-group-dirs":
+                # The special name which for a directory which "does not
+                # have tool groups".
+                return
+            for n in self._groups:
+                # Return the 3 tools groups that are found with the prefix
+                # attached.
+                yield Path(f"{prefix[:-2]}-{n}")
```
webbnh, line 296: Why does this remove the last character (i.e., the `-`) and then add a `-`?  Why doesn't it just
```
                yield Path(prefix + n)
```

lib/pbench/test/unit/agent/test_tool_group.py
```diff
@@ -268,3 +270,52 @@ def mock_read_text(path):
...
+
+        TOOL_GROUP_PREFIX = "tests-vX"
+
+        def __init__(self, name: str, pbench_run: str):
+            self.name = name
+            self.pbench_run = pbench_run
+
+    with mock.patch("pbench.agent.tool_group.Path", MockPath), mock.patch(
+        "pbench.agent.tool_group.ToolGroup", MockToolGroup
+    ):
```
webbnh, line 311: If you add parens, Black will format this better:
```
    with (
        mock.patch("pbench.agent.tool_group.Path", MockPath),
        mock.patch("pbench.agent.tool_group.ToolGroup", MockToolGroup),
    ):
```
and I think it will work the same.

lib/pbench/test/unit/agent/test_tool_group.py
```diff
@@ -268,3 +270,52 @@ def mock_read_text(path):
...
+        # To verify the case where there are no tool groups created, we make a
+        # special Path object which will expose no tool groups.
+        tgs = list(gen_tool_groups("no-tool-group-dirs"))
+        assert len(tgs) == 0
+
+        # Using the special Path object which does have tool groups, we verify
+        # that the expected generated list of Tool Group objects matches the
+        # list of names we expect.
+        names = set([tg.name for tg in gen_tool_groups("have-tool-group-dirs")])
```
webbnh, line 320: `"have-tool-group-dirs"` is not a "special Path object".

It would be better if there were a nice symbolic definition for `"no-tool-group-dirs"`, e.g., `MockPath.NOSUBDIRS" -- then we would be able to get by without the comments.

webbnh (commented):
I didn't review `lib/pbench/cli/agent/commands/tools/kill.py` or its tests; I'll review them after once I have the canned Pbench Server running again, unless this PR gets merged, first....

lib/pbench/agent/tool_group.py
```diff
@@ -75,7 +76,7 @@ def __init__(self, name):
 
         Raises BadToolGroup via the verify_tool_group() method on error.
         """
-        self.tg_dir = self.verify_tool_group(name)
+        self.tg_dir = ToolGroup.verify_tool_group(name, pbench_run)
```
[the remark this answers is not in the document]
portante: I'll make the change.

Moving this change to another PR (forthcoming).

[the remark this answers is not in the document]
portante: Well, it doesn't really work that way.  By invoking the static method with `self.verify_tool_group` an extra "self" argument is passed to the static method.  Only when I invoke it via the class name does it work correctly.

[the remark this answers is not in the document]
webbnh: I think you're likely conflating two separate issues.

In the CUT, there should be no difference between `ToolGroup.verify_tool_group()` and `self.verify_tool_group()`; however, when you're setting up a mock for the function in a unit test, you need to be very careful of how you define and reference the target (mock) function.

Since the function in the CUT is a `@staticmethod`, you need to make sure that the mock function either is a plain function or also is marked as a `@staticmethod`.  You can also make it explicit using the `staticmethod()` constructor.  If the mock is a regular member function, then it will end up being called without the `self` argument if you reference it with the class when you set the mock; if you reference it using an instance of a class when you set up the mock, then it will receive that class (not the class that it is invoked on!) as `self` when it is called.  (Python's binding rules never cease to surprise me.)

lib/pbench/cli/agent/commands/tools/kill.py
portante: Perhaps it is best to just keep it this way for now, and rely on PR #3038 to correct all this?

lib/pbench/agent/tool_group.py
```diff
@@ -173,3 +174,14 @@ def archive(self, target_dir: Path):
             0
         """
         shutil.copytree(str(self.tg_dir), target_dir / self.tg_dir.name, symlinks=False)
+
+
+def gen_tool_groups(pbench_run: str) -> Iterable[ToolGroup]:
```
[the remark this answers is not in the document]
portante: See PR #3038.

lib/pbench/cli/agent/commands/tools/kill.py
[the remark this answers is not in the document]
portante: Can we defer all this to PR #3038?

lib/pbench/agent/tool_group.py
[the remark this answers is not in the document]
portante: Okay, the only way to get this to work is to use `staticmethod()` around the mock function.

lib/pbench/test/unit/agent/test_tool_group.py
portante: For some reason I keep forgetting about `monkeypatch`.  I'll see about re-working that.

lib/pbench/test/unit/agent/test_tool_group.py
portante: I'll try it.

lib/pbench/test/unit/agent/test_tool_group.py
portante: Looks like a difference between upper case and lower case references.  Is that really a change that is necessary?

portante (commented):
The changes to the `tool_group` module have been added to a separate PR, #3038.

I am not sure we want to gate this work on that PR, but I'd like to find an interim solution to get this merged ahead of PR #3038.

The big problem with this PR is that for some reason the unit tests fail under Python 3.6, which may mean it is using some python 3.7+ feature not available in 3.6.

I'll move this to `Draft` mode for now until I can resolve that test failure.

dbutenhof (commented):
Although you've marked this draft while working on the test failure, and you have other issues I believe you plan to change in response to some of Webb's comments, I've resolved all my conversations and I'm tagging this with an "Approve" mostly as a way to say I accept the proposal to defer the `ToolGroup` cleanup issues to your new PR. 👍🏻

lib/pbench/test/unit/agent/test_tool_group.py
webbnh: This means that you are a Good Programmer -- it means that you're writing new code instead of just stealing old code.

("Good Programmers write Good Code; Great Programmers steal Good Code." 🤣)

lib/pbench/test/unit/agent/test_tool_group.py
webbnh: The difference is that `"no-tool-group-dirs"` is a literal string which has no scope or clue to its usage, while `MockPath.NOSUBDIRS` is an encapsulated value with a specific scope and reference to context.

That is, because `NOSUBDIRS` is qualified by a path (i.e., a class), the reader can refer to the class definition for context and potentially documentation.  No such (explicit) indicator is available for the literal string.  Also, because `NOSUBDIRS` encapsulates the value, we don't need to concern ourselves with what the actual "special Path object" _is_, and we can trivially change it if needed because it is specified in only one place and merely referenced in all the others.

The fact that this code snippet requires a block comment to explain what the string is and does (and the fact that the comment _here_ constrains or explains implementation **elsewhere**) suggests that we should probably be doing something different.  Using a symbolic constant instead of a literal value should address that and alleviate the need for the explanation of the _mechanism_ (if we still need to explain the intention, that's another story), and choosing a good name for the symbol might relieve the need for the rest of the comment (i.e., "code as documentation").

webbnh:
> I am not sure we want to gate this work on that PR, but I'd like to find an interim solution to get this merged ahead of PR #3038.

I'm OK with that.  However, I do think that there are some issues in this PR which should be addressed before it is merged (although, I'm not yet clear on which are specific to this PR and which will be swept up in #3038).

> The big problem with this PR is that for some reason the unit tests fail under Python 3.6, which may mean it is using some python 3.7+ feature not available in 3.6.

Yeah, I cannot tell from the context which PyTest presents exactly where the issue lies.  My advice is to replace `unittest.mock` with PyTest's `monkeypatch` and see if that helps.  (We have ample experience with `monkeypatch` and not so much with `unittest`.)  At the very least, it might give a different perspective on the problem which allows us to see what it is.

lib/pbench/test/unit/agent/test_tool_group.py
portante: They are not tool sets but names of tool groups; perhaps `_names`?

lib/pbench/test/unit/agent/test_tool_group.py
portante: Because in this case it is accepting the values passed to it from `get_tool_groups`, which is `<something>-v1-*`, and we want to return `<something>-v1-heavy`, etc.

lib/pbench/test/unit/agent/test_tool_group.py
portante: I'll make a change, but I just don't see how this helps.  The symbols are used only within the scope of the test function so that it results in just upper case letters and underscores vs lowercase letters and dashes.  The comments are still useful, so this ends up just being more code.

lib/pbench/test/unit/agent/test_tool_group.py
webbnh: Sorry, I was "off by one".  How about?
```
                yield Path(prefix[:-1] + n)
```
😁

portante:
The unit test failures under Python 3.6 are because multiple `with` contexts don't work with commas, you have to use multiple nested `with` statements.

portante (commented):
Ready for another round of reviews.

ndokos (dismissed):

dbutenhof (approved):

lib/pbench/test/unit/agent/task/test_tools_kill.py
```diff
@@ -281,17 +281,18 @@ def __init__(self, file_name: str, display_name: str):
...
+            "pbench.cli.agent.commands.tools.kill.PidSource", MockPidSource
+        ):
+            with mock.patch(
                 "pbench.cli.agent.commands.tools.kill.gen_run_directories",
                 mock_gen_run_directories,
-            ),
-        ):
-            result = runner.invoke(kill.main)
+            ):
+                result = runner.invoke(kill.main)
```
webbnh, line 295: I thought you were going to replace `mock` with `monkeypatch`.

I believe that that fixture can be used as a (single) context, which would allow you to do multiple patches without nested `with` statements.

Ditto especially lines 359-370.  (I think Riya's recent PR has examples of how to do this.)

lib/pbench/test/unit/agent/task/test_tools_kill.py
```diff
@@ -330,15 +331,17 @@ def killem(self, echo: Callable[[str], None]) -> None:
...
             pbench_run_dir: pathlib.Path,
         ) -> Tuple[pathlib.Path, str]:
-            return [
+            for run_dir in [
                 (pathlib.Path("run0/tm"), "uuid1abc"),
                 (pathlib.Path("run1/tm"), "uuid2def"),
                 (pathlib.Path("run2/tm"), "uuid3ghi"),
-            ]
+            ]:
+                yield run_dir
```
webbnh, line 339: I think the return type hint here should be an iterator over a list of tuples.

lib/pbench/test/unit/agent/test_tool_group.py
```diff
@@ -280,20 +290,24 @@ class MockPath:
         tool group directories.
         """
 
-        _groups = set(("heavy", "light", "medium"))
+        # The path name determines the behavior.
+        NO_TOOL_GROUP_DIRS = "ntgd"
+        # Tool group names that will be found by the `.glob()` method.
+        _names = set(("heavy", "light", "medium"))
```
webbnh, line 296: Alternately,
```
        _names = {"heavy", "light", "medium"}
```

lib/pbench/test/unit/agent/test_tool_group.py
```diff
@@ -306,16 +320,15 @@ def __init__(self, name: str, pbench_run: str):
...
+        with mock.patch("pbench.agent.tool_group.ToolGroup", MockToolGroup):
+            # To verify the case where there are no tool groups created, we make
+            # a special Path object which will expose no tool groups.
+            tgs = list(gen_tool_groups(MockPath.NO_TOOL_GROUP_DIRS))
+            assert len(tgs) == 0
+
+            # Using a Path object which does have tool groups, we verify that
+            # the expected generated list of Tool Group objects matches the list
+            # of names we expect.
+            names = set([tg.name for tg in gen_tool_groups("some-pbench-run-dir")])
```
webbnh, line 333: Alternately,
```.
            names = {tg.name for tg in gen_tool_groups("some-pbench-run-dir")}
```

webbnh (commented):
There is a spot where I think you are trying to create a (mock) generator which returns no values; this should have a comment explaining it (otherwise, it looks like a bug).

Also, I thought you were going to replace the uses of `unittest.mock` with `pytest.monkeypatch`.  Please consider following though on that, as I think it will make the code simpler and easier to read.

And, I have a couple of coding suggestions.

lib/pbench/test/unit/agent/task/test_tools_kill.py
portante: I can't get rid of the `mock.patch` decorator use because `monkeypatch` does not offer that functionality, but everything else has been removed.

lib/pbench/test/unit/agent/task/test_tools_kill.py
```diff
@@ -80,15 +80,15 @@ def test_noop_killem(mock_kf):
 
     @staticmethod
     @mock.patch("pbench.cli.agent.commands.tools.kill.kill_family")
-    def test_killem(mock_kf):
+    def test_killem(mock_kf, monkeypatch):
```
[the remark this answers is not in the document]
portante: Why would we spend the effort to drop the functionality of `@mock.patch()`, as coded here in two places, with an explicit implementation?  What is the benefit?  Just to not also import `unittest`?

`monkeypatch` offered by `pytest` is quite convenient and powerful, but it does not pitch itself as a complete solution.  The `unittest.MagicMock` is very powerful and avoids a lot of busy-work, which is why it exists in the first place.

FWIW, there is also another use of `unittest.mock` in the tests for the Tool Meister sub-system.

Certainly we can discuss as a team what we want to do going forward.

portante (commented):
Nasty merge conflict with PR #3044, should be okay now.

lib/pbench/test/unit/agent/task/test_tools_kill.py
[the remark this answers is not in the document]
webbnh: If we are going to start using `unittest.mock`, then I think we should stop using `pytest.monkeypatch`.  So, before we add `unittest` here, I think that we should have a plan for removing other uses of`monkeypatch` and that we as a team should commit to that plan.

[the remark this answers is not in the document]
portante: Why do we have to be so rigid?  Why can't we use the feature of a module which provides the right tool for the right job?

dbutenhof (approved):

ndokos (dismissed):

portante (commented):
One more change made to match the `mock_verify_tool_group()` signature to the real thing.

dbutenhof (approved):

portante:
@webbnh, the CI job seems a bit flaky.

portante:
CI jobs need to be fixed first, moved this PR to draft mode until PR #3048 is merged.

lib/pbench/test/unit/agent/test_tool_group.py
```diff
@@ -255,3 +267,56 @@ def mock_read_text(path):
...
+
+    class MockPath:
+        """A simple ToolGroup mock to generate host names from different
+        tool group directories.
+        """
+
+        # The path name determines the behavior.
+        NO_TOOL_GROUP_DIRS = "ntgd"
+        # Tool group names that will be found by the `.glob()` method.
+        _names = {"heavy", "light", "medium"}
```
ndokos, line 283: It's a minor point but I believe it is still the case that the config file uses `light`, `medium` and `heavy` as tool-set names, so using them as tool group names here might be a bit confusing.  It is likely of course that the tool set names (as well as the config file) will fall off by the wayside, so who cares? But I just wanted to point it out. It's definitely not something to delay the merge of this PR, so I'll resolve the convo preemptively.

ndokos (approved):

dbutenhof (approved):

dbutenhof (dismissed):

lib/pbench/test/unit/agent/task/test_tools_kill.py
```diff
@@ -0,0 +1,424 @@
...
+            [
+                ("abc.file", None),  # top level file
+                ("filenotf", None),  # directory with no .uuid file
+                ("run1", "abcdef"),  # directory with a file
+                ("run2", "ghijkl"),  # directory with a file
+            ]
+        )
+        pairs = list(kill.gen_run_directories(pr))
+        assert pairs[0][0].name == "run1/tm" and pairs[0][1] == "abcdef"
+        assert pairs[1][0].name == "run2/tm" and pairs[1][1] == "ghijkl"
```
webbnh, line 127: If `pairs[0][0]` is a `pathlib.Path`, then it's really unlikely that `.name` would have a slash in it.  (I mean, it _could_ if the slash were escaped when the file was created, but it seems more likely to me that, something weird is going on with the code.)

webbnh (commented):
The change to use `MagicMock` looks good.

However, I stumbled over this...and I have to run out on an errand, so I'll have to get back to this later.

webbnh (commented):
Here's some more stuff I found.

I'm trying to get through this, but I didn't; I'll try to finish tomorrow.

lib/pbench/test/unit/agent/test_tool_group.py
```diff
@@ -123,7 +125,9 @@ def mock_read_text(self: Path):
             """Mocked directory check"""
             raise FileNotFoundError("Mock Path.resolve()")
 
-        monkeypatch.setattr(ToolGroup, "verify_tool_group", self.mock_verify_tool_group)
+        monkeypatch.setattr(
+            ToolGroup, "verify_tool_group", staticmethod(self.mock_verify_tool_group)
+        )
```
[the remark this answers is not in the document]
portante: When we simply remove the `staticmethod()` constructor, this fails.  If we replace it with `Test_ToolGroup.mock_verify_tool_group` instead of `self.mock_verify_tool_group` it also fails.

lib/pbench/test/unit/agent/task/test_tools_kill.py
portante: The current version of `AnotherPath` as reviewed here does.  I have updated the class to fix that.

lib/pbench/test/unit/agent/task/test_tools_kill.py
```diff
@@ -0,0 +1,424 @@
...
+    def is_dir(self):
+        return not self.name.endswith((".uuid", ".pid", ".file"))
+
+    def read_text(self):
+        if self.val is None:
+            raise FileNotFoundError(self.name)
+        return self.val
+
+    def __truediv__(self, arg: str):
+        return AnotherPath(f"{self.name}/{arg}", val=self.val)
```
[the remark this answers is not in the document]
portante: I am not sure what you are getting at here, but perhaps by making `AnotherPath` work closer to `Path` this address whatever the concern is here?

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -0,0 +1,247 @@
...
+from pbench.agent.utils import LocalRemoteHost, TemplateSsh
+from pbench.cli.agent import CliContext, pass_cli_context
+from pbench.cli.agent.options import common_options
+
+
+def kill_family(proc: psutil.Process):
+    """Kill a parent process and all its children."""
+    try:
+        # Get the list of children of the parent before killing it.
+        children = list(proc.children())
```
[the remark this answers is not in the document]
portante: Sure that was an oversight to not use `recursive=True`.

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -0,0 +1,247 @@
...
+            self.procs_by_uuid[uuid] = psutil.Process(pid)
+        except psutil.NoSuchProcess:
+            return False
+        self.uuid_to_tmdir[uuid] = tm_dir
+        return True
+
+    def killem(self, echo: Callable[[str], None]) -> None:
+        """Kill all PIDs found, and their children."""
+        if not self.procs_by_uuid:
+            return
```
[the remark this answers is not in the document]
portante: I only want to emit output if actions are going to be taken.

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -0,0 +1,247 @@
...
+    if not tool_groups:
+        return
+
+    lrh = LocalRemoteHost()
+
+    for tg in tool_groups:
+        for host_name in tg.hostnames.keys():
+            if lrh.is_local(host_name):
+                continue
+            yield host_name
```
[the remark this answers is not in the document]
portante: Its really expensive.

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -0,0 +1,247 @@
...
+        If any UUIDs are passed as arguments, we only want to look for, and
+        locally kill, processes having those UUIDs.
+
+        Without command line arguments, kill all the local PIDs from all the
+        discovered run directories.
+
+        All the Redis server PIDs are killed first, then the Tool Data Sinks,
+        and finally the local Tool Meisters.
+
+        We then remotely kill (via `ssh`) all the Tool Meisters by invoking
```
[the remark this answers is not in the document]
portante: That is the explicit goal.  Kill the Redis servers, that "should" cause all the TMs and the TDS to exit on their own, gracefully.  Anything left over really needs to be killed.

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -0,0 +1,247 @@
...
+        local_pids = False
+        remote_tms = defaultdict(list)
+        for tm_dir, uuid in gen_run_directories(self.pbench_run):
+            # If a run directory has any dangling components of the Tool
+            # Meister sub-system active, that is PID files for any local
+            # component, record those components.
+            for pidsrc in all_pids:
+                found = pidsrc.load(tm_dir, uuid)
+                if found:
+                    local_pids = True
```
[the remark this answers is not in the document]
portante: One modification to add a proper comprehension:
```
local_pids |= any([pidsrc.load(tm_dir, uuid) for pidsrc in all_pids])
```

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -0,0 +1,247 @@
...
+        # We kill all the Redis Servers first in case killing them causes all
+        # the other processes to just exit on their own.  Then we kill all the
+        # Tool Data Sinks, and their children.  Then we kill all the (local)
+        # Tool Meisters, and their children.
+        for pidsrc in all_pids:
+            pidsrc.killem(click.echo)
+
+        # Kill all the remote Tool Meisters.
+
+        cmd = "pbench-tools-kill {{uuids}}"
```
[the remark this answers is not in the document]
portante: Originally I was going to write a special interface hidden from the user for the UUID case, so it was hard-coded that way.

Unfortunately, I can't find a method or attribute that will be the right thing in the various contexts in which this code needs it.

I am open to ideas if you have something in mind.

lib/pbench/test/unit/agent/task/test_tools_kill.py
```diff
@@ -20,7 +20,14 @@ class AnotherPath:
     """
 
     def __init__(self, name: str, val: Optional[str] = None):
-        self.name = name
+        """Approximate mocked behavior of a Path object where the"""
```
dbutenhof, line 23: "where the ..."???

dbutenhof (dismissed):

lib/pbench/cli/agent/commands/tools/kill.py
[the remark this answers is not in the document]
portante: We are not hoping they terminate on their own.  We are killing the Redis server first so that they "can't" continue on their own, and to nudge them along to exiting while we work on the remotes.

I fail to see why this order is bad and worth changing.

lib/pbench/cli/agent/commands/tools/kill.py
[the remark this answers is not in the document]
webbnh: Why do you think we need a list comprehension here?  The `XXX for Y in Z` should create an iterator which should be acceptable as an argument to `any()`.  (It worked in my testing....)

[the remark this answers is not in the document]
portante: It failed until I made a list.

lib/pbench/test/unit/agent/task/test_tools_kill.py
portante: Ugh.

lib/pbench/cli/agent/commands/tools/kill.py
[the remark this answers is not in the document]
webbnh: Sorry -- I misspoke -- it's not an _iterator_ it's a **generator**:
```
>>> f = (is_even(i) for i in my_ints)
>>> type(f)
<class 'generator'>
```
🥴

lib/pbench/cli/agent/commands/tools/kill.py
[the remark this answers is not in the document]
webbnh: OK, but that's actually less explicit:  that is, the first output confirms the action and then the lack of subsequent output confirms the absence of effect; without the first output, we have less information as to why nothing happened.

lib/pbench/test/unit/agent/task/test_tools_kill.py
```diff
@@ -0,0 +1,440 @@
...
+
+    @staticmethod
+    def test_with_uuids(monkeypatch):
+        """Verify UUID arguments work correctly.
+
+        The scenario we construct artificially starts with an invocation using
+        3 UUIDs, one won't be found.  The mocked-out `psutil.process_iter()`
+        will generate 3 PIDs, one of which will have child processes, which
+        will all operate normally with no exceptions.  Of the other two, one
+        will raise an expected exception, and the other will not be found.
```
webbnh, line 204: s/expected/unexpected/ ?

lib/pbench/test/unit/agent/task/test_tools_kill.py
```diff
@@ -0,0 +1,440 @@
...
+        result = runner.invoke(kill.main, uuid_list)
+        assert result.exit_code == 0
+        expected_action_list = [
+            ("kill", 12345, "uuid1abc"),
+            ("kill", 12346, None),
+            ("kill", 12347, None),
+            ("kill-nosuch", 12348, None),
+            ("kill-except", 34567, "uuid2def"),
+            ("kill-nosuch", 45678, "uuid3ghi"),
+        ]
```
webbnh, line 280: Shouldn't the third value in the tuples at lines 274, 278, and 279 be referencing `uuid_list` (a la lines 258, 266, and 267)?

lib/pbench/test/unit/agent/task/test_tools_kill.py
```diff
@@ -0,0 +1,440 @@
...
+        """Verify UUID arguments work correctly.
+
+        The scenario we construct artificially starts with an invocation using
+        3 UUIDs, one won't be found.  The mocked-out `psutil.process_iter()`
+        will generate 3 PIDs, one of which will have child processes, which
+        will all operate normally with no exceptions.  Of the other two, one
+        will raise an expected exception, and the other will not be found.
+        """
+        action_list = []
+        uuid_list = ["uuid1abc", "uuid2def", "uuid3ghi"]
```
webbnh, line 207: Consider using a(n immutable) tuple here instead of a list.  (It shouldn't affect the code which references `uuid_list`, but it will prevent `uuid_list` from being inadvertently altered.)

lib/pbench/test/unit/agent/task/test_tools_kill.py
```diff
@@ -0,0 +1,440 @@
...
+        """Exercise the code path where no run directories are found, so there
+        is nothing to kill.
+        """
+
+        class MockPidSource:
+            """No-op mock for PidSource."""
+
+            def __init__(self, file_name: str, display_name: str):
+                self.file_name = file_name
+                self.display_name = display_name
```
webbnh, line 299: Do we need lines 298 & 299?  (Consider using [unittest.mock.Mock](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock) instead of rolling your own mock-class.)

lib/pbench/test/unit/agent/task/test_tools_kill.py
```diff
@@ -0,0 +1,440 @@
...
+            ("load", "tm.pid", "run0/tm", "uuid1abc"),
+            ("load", "redis.pid", "run1/tm", "uuid2def"),
+            ("load", "pbench-tool-data-sink.pid", "run1/tm", "uuid2def"),
+            ("load", "tm.pid", "run1/tm", "uuid2def"),
+            ("load", "redis.pid", "run2/tm", "uuid3ghi"),
+            ("load", "pbench-tool-data-sink.pid", "run2/tm", "uuid3ghi"),
+            ("load", "tm.pid", "run2/tm", "uuid3ghi"),
+            ("killem", "redis server"),
+            ("killem", "tool data sink"),
+            ("killem", "local tool meister"),
```
webbnh, line 411: Why do we load nine and kill only three?

webbnh (commented):
OK, I finished:  down to just small stuff except for a spot where I think we're missing a mock and some confusion on loading vs. killing.

lib/pbench/cli/agent/commands/tools/kill.py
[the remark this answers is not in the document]
portante: The reason for the failure is that when `any` is given an iterable, and the iterable is a generator, the generator will stop being used once a `True` value is encountered:

```
$ cat show-any.py
def gen(second):
   print("1st")
   yield False
   print("2nd")
   yield second
   print("3rd")
   yield False

if not any(gen(False)):
    print("All False")
if any(gen(True)):
    print("Second True")
$
$
$ python3 show-any.py
1st
2nd
3rd
All False
1st
2nd
Second True
$
```

lib/pbench/test/unit/agent/task/test_tools_kill.py
portante: Because `killem()` is `kill()`ing all the PIDs for each of the PID sources, Redis, TDS, and TM.  We have 3 run directories, with 3 PID sources, for a total of 9 loads.

portante (commented):
I believe all code-review feedback given up to this point has been addressed.  Thanks for the reviews!

dbutenhof (approved):

lib/pbench/test/unit/agent/task/test_tools_kill.py
```diff
@@ -191,8 +192,31 @@ def iterdir(self):
...
+    def test_gen_host_names_no_tool_groups(monkeypatch):
+        """Verify kill.gen_host_names() is a no-op when we have no tool groups
+        present.
+        """
+
+        def mock_gen_tool_groups(pbench_run: str) -> Iterable[ToolGroup]:
+            """Override kill.gen_tool_groups definition to yield no ToolGroup
+            objects.
+            """
+            for hosts in []:
```
ndokos, line 205: nit: "host"

In addition, `ToolGroup(hosts)` is confusing: the constructor takes the name of a tool group as argument, not a host. Even though it is not called at all, it should still make sense.

ndokos (changes requested):
Some more items for your consideration.

lib/pbench/cli/agent/commands/tools/kill.py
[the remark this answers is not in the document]
webbnh: Right, and the problem is that the code is doing _two_ things:  in addition to looking for a `True` result, we also want it to call `load()` for **all** of the `pidsrc`/`tm_dir` combinations.  I had failed to note that second part when I made my suggestion.  🥴

lib/pbench/test/unit/agent/task/test_tools_kill.py
```diff
@@ -79,29 +96,63 @@ def test_load_found(monkeypatch):
...
+            ps.load(AnotherPath(vals["dirname"], vals["pid"]), uuid)
         ps.killem(ourecho)
         assert len(mock_kf.mock_calls) == 3
-        assert mock_kf.mock_calls[0][1][0].pid == 123
-        assert mock_kf.mock_calls[1][1][0].pid == 456
-        assert mock_kf.mock_calls[2][1][0].pid == 789
+        idx = 0
+        for uuid in pid_data.keys():
+            assert mock_kf.mock_calls[idx][1][0].pid == int(pid_data[uuid]["pid"])
+            idx += 1
```
webbnh, line 146: I'm not convinced that this table-driven approach is significantly "better" than just listing out the three assertions explicitly, but...if you're going to do it this way, consider using [enumerate()](https://docs.python.org/3/library/functions.html#enumerate)...something like this:
```
        for idx, uuid in enumerate(pid_data.keys()):
            assert mock_kf.mock_calls[idx][1][0].pid == int(pid_data[uuid]["pid"])
```
which saves you from having to set up and manually increment the index variable.

lib/pbench/test/unit/agent/task/test_tools_kill.py
```diff
@@ -79,29 +96,63 @@ def test_load_found(monkeypatch):
...
+        # Verify a second call to killem() is a no-op.
+        ps.killem(ourecho)
+        assert len(mock_kf.mock_calls) == len(pid_data)
```
webbnh, line 149: Line 149 is a nice touch (as long as we're using the table-driven approach)...shouldn't line 142 be the same?

webbnh, line 149: Wouldn't it be better to [reset `mock_kf`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.reset_mock), and then `mock_kf.assert_not_called()`?

lib/pbench/test/unit/agent/task/test_tools_kill.py
```diff
@@ -141,8 +192,31 @@ def iterdir(self):
...
+            objects.
+            """
+            for hosts in []:
+                yield ToolGroup(hosts)
+
+        monkeypatch.setattr(
+            "pbench.cli.agent.commands.tools.kill.gen_tool_groups", mock_gen_tool_groups
+        )
+        hosts = list(kill.gen_host_names(pathlib.Path("no-tool-group-dirs")))
+        assert len(hosts) == 0
```
webbnh, line 212: Supposedly, it's "not Pythonic" to check if the length of a list is zero.  Instead, we're supposed to look at the list itself in a boolean context, e.g.:
```
        assert not list(kill.gen_host_names(pathlib.Path("no-tool-group-dirs")))
```
However, that's not very satisfying.  How about this, instead:
```
        assert not any(kill.gen_host_names(pathlib.Path("no-tool-group-dirs")))
```

webbnh (dismissed):
Looks good; I just have a few items from your latest changes for your consideration.

BTW, GitHub seemed to have lost track of Mr. Jenkins, so I manually kicked off a new build; they seem to be talking, now.

portante (commented):
After having a conversation with @ndokos about these code changes, I thought I had implemented this incorrectly.  Upon review, it was not implemented incorrectly, but the names of variables and functions, and references in comments, we're not using the terminology we typically use.

lib/pbench/cli/agent/commands/tools/kill.py
```diff
@@ -107,12 +107,13 @@ def killem(self, echo: Callable[[str], None]) -> None:
                 echo(f"\t\terror killing {pid}: {exc}", err=True)
 
 
-def gen_run_directories(pbench_run: pathlib.Path) -> Iterable[Tuple[pathlib.Path, str]]:
-    """Generate the list of run directories available under ${pbench_run},
-    yielding a Path object for that directory, along with its recorded
-    UUID.
+def gen_result_tm_directories(
+    pbench_run: pathlib.Path,
+) -> Iterable[Tuple[pathlib.Path, str]]:
```
dbutenhof, line 112: Sometimes I really hate `black`. I'm tempted to suggest a `NamedTuple` or even an `@dataclass` to give your path+uuid a real type, which could make it easier to read and just might "incidentally" shorten this declaration enough to fit on a line❓ 

Anyway, it's a simple tuple and I'm not going to hold the PR over this, so I'll pre-resolve this and approve: but something to consider if you make another pass for something else.

lib/pbench/test/unit/agent/task/test_tools_kill.py
```diff
@@ -42,8 +42,9 @@ def is_dir(self) -> bool:
         return not self.name.endswith((".uuid", ".pid", ".file"))
 
     def read_text(self) -> str:
-        """The read text value is return that was provided when the mock'd
-        object was created, otherwise a FileNotFoundError is raised.
+        """The read text value is returned using the value that was provided
+        when the mock'd object was created, otherwise a FileNotFoundError is
```
dbutenhof, line 46: s/mock'd/mocked/ ??

dbutenhof (dismissed):

lib/pbench/test/unit/agent/task/test_tools_kill.py
portante: Are you [mocking](https://youtu.be/OLWFcva2AYQ) me?

dbutenhof (approved):

lib/pbench/test/unit/agent/task/test_tools_kill.py
```diff
@@ -176,18 +176,18 @@ def iterdir(self):
...
                 ("abc.file", None),  # top level file
                 ("filenotf", None),  # directory with no .uuid file
-                ("run1", "abcdef"),  # directory with a file
-                ("run2", "ghijkl"),  # directory with a file
+                ("res1", "abcdef"),  # directory with a file
+                ("res2", "ghijkl"),  # directory with a file
             ]
         )
-        pairs = list(kill.gen_run_directories(pr))
+        pairs = list(kill.gen_result_tm_directories(pr))
```
webbnh, line 183: It would probably be good to confirm that the correct number of pairs is returned.

webbnh (dismissed):
Looks good to me.  However, in order to exact retribution on Peter for sending me to YouTube in the middle of a work day, I will point out one more thing...but I'll pre-resolve it.  😉

dbutenhof (approved):

ndokos (approved):

webbnh (approved):
👍

PR #2744, opened 2022-04-12
Dev dashboard init

MVarshini:
The PR intends to mock the workflow of the user authentication process (example: Login, Signup) by imitating the user notion.

The wireframe of the implementation:[ wireframe](https://marvelapp.com/prototype/dh1cd91/screen/66676167)


MVarshini:
@dbutenhof CI is failed becasue of the error in email_validator/__init__.py
Found in the [logs](http://jenkins.intlab.perf-infra.lab.eng.rdu2.redhat.com/job/pbench-ci-multi/job/PR-2744/13/console)

dashboard/src/modules/components/AuthComponent/signupFormData.js
```diff
@@ -0,0 +1,70 @@
...
+    name: "passwordSpecialChars",
+  },
+  {
+    label: "1 number",
+    name: "passwordContainsNumber",
+  },
+  {
+    label: "1 uppercase letter",
+    name: "passwordBlockLetter",
+  }
```
dbutenhof, line 69: This generally seems to be compatible with the structure of Red Hat SSO; however instead of a fixed requirement, it allows "domains" to be structured with specific password policy settings, including a "number of character classes" and "minimum length". The "character classes" would include the special character, digits, uppercase, and lowercase; it doesn't appear you can require specific classes, but you can specify how many, with a default of 0. (That is, if you configure your domain with "2 classes", a valid password might have uppercase and lowercase, or digit and special character; we might need to have a server API for the dashboard to discover the confirmation set by the SSO domain.)

You're also hardcoding a minimum length of 8, and I know sites that require 12; again, Keycloak/Red Hat SSO allows setting that by domain, so eventually we'll need something more dynamic.

Right now, this is probably fine, although I've become a fan of the "random hyphenated words" password style we're not supporting here, which I suppose would qualify as a "2 character class" (hyphen and lowercase). Not worth worrying about right now, but the configuration is something to keep in mind.

dashboard/src/utils/apiConstants.js
```diff
@@ -1,2 +1,4 @@
-export const LOCAL_BASE_URL = "/";
+export const LOCAL_BASE_URL = "http://10.1.170.201/";
```
dbutenhof, line 1: We need to get the `endpoints` change in. 😁

That will not only remove this constant, but this entire file. (In theory you could have symbols referencing the various endpoint key names, but in JSON format they're reasonably clear symbols already, like `api.register`, which we can reference, if we keep my `index.html` assignment to `window.origin` as `window.origin.api.register`...)

dashboard/src/actions/authActions.js
```diff
@@ -0,0 +1,122 @@
+import * as types from "./types";
+import API from "../utils/api";
+import * as API_ROUTES from "../utils/apiConstants";
```
dbutenhof, line 3: Some style consistency might be nice; here you're importing the `apiConstants` using an all-caps `API_ROUTES`, but in `App.js` you imported the similar set of `routeConstants` as a camel-case `AppRoutes` and `types` is imported using lowercase `types`. Can we choose one style?

dbutenhof (changes requested):
> @dbutenhof CI is failed becasue of the error in email_validator/init.py

Fascinating. Yes; I had another Jenkins run fail the same way this morning.

Apparently we've been "upgraded" to a new version of the Python email validation package that now explicitly disallows use of the `example.com` domain! ☹️ 

We'll have to get that fixed, either by changing the tests to use "real" emails, or locking to an older version of the validator for now.

For now, most of this looks good; I would like to see some consistency around the imported names used for constant collections, and a few other minor changes. (And we need to fix the tests or requirements before we can merge anything, anyway.)

dashboard/src/modules/components/AuthComponent/SignupForm.jsx
```diff
@@ -0,0 +1,190 @@
...
+      });
+      // edge case where user deliberately
+      // edits the password field, even when
+      // confirm password is not empty.
+      checkPasswordError(value, userDetails.confirmPassword);
+    } else if (fieldName === "passwordConfirm") {
+      checkPasswordError(userDetails.password, value);
+    }
+  };
+  const sendForRegisteration = () => {
```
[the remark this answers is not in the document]
MVarshini: Updated! :grin:

dashboard/src/actions/authActions.js
MVarshini: As suggested, changed the imports to all-caps format.

dbutenhof (approved):
We'll need to run the Jenkins tests again after #2776 is merged, (and a manual rebase wouldn't hurt, although there's no overlap); but this is looking good.

portante (changes requested):
Needs a rebase in order to pick up the fixed test infrastructure.

dbutenhof (approved):

portante (approved):

dbutenhof:
This is interesting: we had a failure in tool meister test-53, which didn't reproduce when I restarted the Jenkins job. I just created Jira [PBENCH-633](https://github.com/distributed-system-analysis/pbench/pull/2744) to track this.

```
test-53 test-client-tool-meister 
*** /home/jenkins/pbench/agent/util-scripts/gold/test-client-tool-meister/test-53.txt	2022-04-27 12:32:06.030368654 +0000
--- /var/tmp/pbench-test-utils/output.txt	2022-04-27 12:37:51.451674512 +0000
***************
*** 1,8 ****
  +++ Running test-53 test-client-tool-meister 
  "mpstat" tool is now registered for host "testhost.example.com" in group "default"
  "dcgm" tool is now registered for host "testhost.example.com" in group "default"
! Collecting system information
! --- Finished test-53 test-client-tool-meister (status=0)
  +++ pbench tree state
  /var/tmp/pbench-test-utils/pbench
  /var/tmp/pbench-test-utils/pbench/mock-run
--- 1,10 ----
  +++ Running test-53 test-client-tool-meister 
  "mpstat" tool is now registered for host "testhost.example.com" in group "default"
  "dcgm" tool is now registered for host "testhost.example.com" in group "default"
! timeout: sending signal TERM to command ‘pbench-tool-meister-stop’
! ERROR - "pbench-tool-meister-stop --sysinfo='block,security_mitigations,sos' 'default'" failed to execute successfully (exit code: 124)
! killing /var/tmp/pbench-test-utils/pbench/mock-run/tm/redis*.pid (20490)
! --- Finished test-53 test-client-tool-meister (status=1)
  +++ pbench tree state
  /var/tmp/pbench-test-utils/pbench
  /var/tmp/pbench-test-utils/pbench/mock-run
FAIL - test-53
```

PR #2747, opened 2022-04-12
Remove `run_command` utility function

portante:


webbnh (approved):
It's truly a thing of beauty.

dbutenhof (dismissed):
Yeah, that works.

#2741 is going to be a moment ... the merge seems to have broken our favorite test-61 and I need to figure out why... which of course is complicated by the fact that I moved all the code and had to do the rebase cleanup piece by piece.

lib/pbench/cli/agent/commands/cleanup.py
```diff
@@ -19,7 +18,7 @@ def execute(self):
             "%s deprecated, will be removed in future release in favor of pbench-clear-results",
             self.name,
         )
-        run_command("pbench-clear-results")
+        os.execvp("pbench-clear-results")
```
dbutenhof, line 21: So, `os.execlp("pbench-clear-results")` would have been better since you're not using an argv.

portante (commented):
Okay, this now works.  Was missing an argument.  Added the output to the assertion to help debug.

lib/pbench/cli/agent/commands/cleanup.py
webbnh: Dang it, Peter!...I told you to change that!  🤪

webbnh (approved):
👍

lib/pbench/cli/agent/commands/cleanup.py
```diff
@@ -19,7 +18,7 @@ def execute(self):
             "%s deprecated, will be removed in future release in favor of pbench-clear-results",
             self.name,
         )
-        run_command("pbench-clear-results")
+        os.execlp("pbench-clear-results", "pbench-clear-results")
```
dbutenhof, line 21: OK, that's uglier than I hoped; but, yeah "path" and "arguments" and we need `argv[0]` to be set manually? Blech. 😬

dbutenhof (approved):

lib/pbench/cli/agent/commands/cleanup.py
portante: Still better than `run_command`!!!

webbnh: Yep, I had both of those thoughts, too.

PR #2748, opened 2022-04-12
PBENCH-201 Experimental Jira linkage

dbutenhof:
New PR with linkage to new Jira ticket.

### DIFF
--- a/lib/pbench/agent/utils.py
+++ b/lib/pbench/agent/utils.py
@@ -245,7 +245,7 @@ def setup_logging(debug, logfile):
     return rootLogger
 
 
-def run_command(args, env=None, name=None, logger=None):
+def run_command(args, env=None, name=None):
     """Run the command defined by args and return its output"""
     try:
         output = subprocess.check_output(args=args, stderr=subprocess.STDOUT, env=env)
@@ -253,9 +253,7 @@ def run_command(args, env=None, name=None, logger=None):
             output = output.decode("utf-8")
         return output
     except subprocess.CalledProcessError as e:
-        message = "%s failed: %s" % (name, e.output)
-        logger.error(message)
-        raise RuntimeError(message)
+        raise RuntimeError("%s failed: %s" % (name, e.output))
 
 
 def _log_date():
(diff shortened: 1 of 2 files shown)


webbnh:
I'm assuming that this PR does not require review (as it's been superseded by #2747...) and that you're just exploring the Jira-GitHub integration.  Let me know if you need me to do anything for this PR.

dbutenhof:
> I'm assuming that this PR does not require review (as it's been superseded by #2747...) and that you're just exploring the Jira-GitHub integration. Let me know if you need me to do anything for this PR.

Right, which is why it was marked DONT-MERGE. I suppose I could have been more explicit, though!

dbutenhof:
The Jira linkage appears to be working now. The Jira ticket was automatically updated to IN PROGRESS status and now shows the linkage to the OPEN pull request.

Out of curiosity, I'm going to try adding some time in this comment.

PBENCH-201 #time 2h

dbutenhof:
> Out of curiosity, I'm going to try adding some time in this comment.

Which didn't work. It seems that although Jira uses the jira ticket names to *connect* the PR with the Jira, it doesn't process `#time` until the merge. (E.g., in #2741 I put several `#time` commands in the git commit logs, which were ignored; but I edited a final `#time` into the GitHub dialog when I did the squash-and-merge, and *that* one was processed.)

webbnh:
> It seems that although Jira uses the jira ticket names to _connect_ the PR with the Jira, it doesn't process `#time` until the merge.

This makes sense to me.  Admittedly, it means that we cannot dynamically track time-on-task _during_ the task execution, it does mean that the time can be accounted for retrospectively.

The problem that I see Jira facing in trying to take `#time` commands from bits of the Git dialogs is that, until they are merged into the main line, they are mutable.  If Jira were to look at all the commits on the to-be-merged branch (e.g., when the pull request is created or updated) and add each one with a `#time` tag, it would have to be able to determine which tags were new, so that it could avoid duplicating existing log entries; and, it would also have to account for any log entries which came from Git commits which have since been squashed or deleted.  This seems like a really big ask!

Having Jira do the accounting only once, when the PR is merged into the main line, makes much more sense (because, in most cases, the commits and their `#time` tags aren't going to change after that).  (That is, it makes more sense for the implementation of the Jira tool...less so for our desire to see how much time is currently left in outstanding sub-tasks.)

PR #2749, opened 2022-04-12
Get rid of the `pbench-perl-JSON-XS` dependency

ndokos:
All distros of interest now make perl-JSON-XS available either in
EPEL (for RHEL7) or in the CRB repo (for RHEL8 and above). The Fedoras
always had it available.

Delete obsolete targets in rpm.mk.

### DIFF
--- a/agent/rpm/pbench-agent.spec.j2
+++ b/agent/rpm/pbench-agent.spec.j2
@@ -23,16 +23,8 @@ Requires:  scl-utils, rh-python36
 Requires:  python3-pip
 %endif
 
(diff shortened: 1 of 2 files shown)


portante (approved):

dbutenhof (approved):

webbnh:
This change was made for `b0.69`...is there a similar change due for `main`?

ndokos:
I didn't see a problem when testing with a 0.71 RPM but I must have missed RHEL7 which still uses `pbench-perl-JSON-XS`. So yes, there will be a similar change for `main`.

PR #2754, opened 2022-04-13
Remove `pbench-sysinfo-dump` reliance on `base`

portante:
There are two environment variables provided by the `agent/base` script which `pbench-sysinfo-dump` was relying on to have in place.  One was explicitly set by the Tool Meister in the environment used to run the script, `pbench_install_dir`, the other was inherited, `_pbench_full_hostname`.

We change `pbench-sysinfo-dump` to use `sysinfo_install_dir` and `sysinfo_full_hostname` instead, with the Tool Meister setting those environment variables explicitly.

We also take the opportunity to remove a useless `_PBENCH_UNIT_TESTS` reference which is not used due to the existence of the mock, `agent/util-scripts/test-bin/pbench-sysinfo-dump`.

### DIFF
--- a/agent/util-scripts/tool-meister/pbench-sysinfo-dump
+++ b/agent/util-scripts/tool-meister/pbench-sysinfo-dump
@@ -1,6 +1,24 @@
 #!/bin/bash
 # -*- mode: shell-script; indent-tabs-mode: t; sh-basic-offset: 8; sh-indentation: 8; tab-width: 8 -*-
 
+# Installation directory.
+install_dir="$(realpath -e ${sysinfo_install_dir} 2> /dev/null)"
+if [[ ${?} -ne 0 ]]; then
+	printf -- "You must provide an installation directory that exists, \"%s\" is not a real path\n" "${sysinfo_install_dir}" >&2
+	exit 1
+fi
+if [[ ! -d "${install_dir}" ]]; then
+	printf -- "You must provide an installation directory that exists, \"%s\" does not\n" "${install_dir}" >&2
+	exit 1
+fi
+
+# Ensure we have a host name to use.
+if [[ -z "${sysinfo_full_hostname}" ]]; then
+	printf -- "You must provide a FQDN host name via 'sysinfo_full_hostname'\n" >&2
+	exit 1
+fi
+
+# Target directory where data should be written.
 dir="$(realpath -e ${1} 2> /dev/null)"
 if [[ ${?} -ne 0 ]]; then
 	printf -- "You must provide a directory that exists, \"%s\" is not a real path\n" "${1}" >&2
@@ -11,13 +29,15 @@ if [[ ! -d "${dir}" ]]; then
 	exit 1
 fi
 
+# Which pieces of system information should be collected. The options
+# available are the suffixes to all the collect_* functions.
 sysinfo="${2}"
 
-if [[ -z "${_PBENCH_UNIT_TESTS}" ]]; then
-	mode="${3}"
-else
-	# Force sequential mode for unit tests.
-	mode=""
+# Collection mode, serial or parallel.
+mode="${3}"
+if [[ "${mode}" != "serial" && "${mode}" != "parallel" ]]; then
+	printf -- "Invalid collection mode, \"%s\"\n" "${mode}" >&2
+	exit 1
 fi
 
 function collect_kernel_config {
@@ -115,7 +135,7 @@ function collect_sos {
 		_cksum_suffix="sha256"
 	fi
 
-	_name="pbench-${_pbench_full_hostname}"
+	_name="pbench-${sysinfo_full_hostname}"
 	_cmd="${dir}/sosreport-${_name}.cmd"
 	printf -- "%s" "${_sos_report_cmd}" > ${_cmd}
 	for mod in ${_modules}; do
@@ -146,8 +166,8 @@ function collect_sos {
 function collect_ara {
 	# collect data only when ara is installed
 	if python -c "import ara" &> /dev/null; then
-		ara_pb="${pbench_install_dir}/ansible/ara/ara.yml"
-		if [[ ! -d "${pbench_install_dir}" || ! -f ${ara_pb} ]]; then
+		ara_pb="${install_dir}/ansible/ara/ara.yml"
+		if [[ ! -d "${install_dir}" || ! -f ${ara_pb} ]]; then
 			printf -- "ERROR: ara playbook not found, '%s'\n" "${ara_pb}" >&2
 			return 1
 		fi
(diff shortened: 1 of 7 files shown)


agent/util-scripts/tool-meister/pbench-sysinfo-dump
```diff
@@ -11,13 +29,15 @@ if [[ ! -d "${dir}" ]]; then
...
-if [[ -z "${_PBENCH_UNIT_TESTS}" ]]; then
-	mode="${3}"
-else
-	# Force sequential mode for unit tests.
-	mode=""
+# Collection mode, serial or parallel.
+mode="${3}"
+if [[ "${mode}" != "serial" && "${mode}" != "parallel" ]]; then
+	printf -- "Invalid collection mode, \"%s\"\n" "${mode}" >&2
+	exit 1
```
ndokos, line 40: So there is no reason to force sequential mode for unit tests? Won't the various collections step on each other? Could it produce a random order in the output file? Or is that sorted? Or is it invoked in sequential mode?

ndokos (dismissed):
First pass looks good to me. I have one question about sequential collection for unit tests, but I'm probably missing something and it's too late to chase it down tonight.

agent/util-scripts/tool-meister/pbench-sysinfo-dump
portante: Yes, because it is mocked out.  See https://github.com/distributed-system-analysis/pbench/blob/main/agent/util-scripts/test-bin/pbench-sysinfo-dump.

portante (commented):
I added the fix from PR #2758 to help with tests.  We should merge that PR first.

agent/util-scripts/test-bin/common-tm-cleanup
```diff
@@ -10,12 +10,26 @@ function _kill_pid {
     fi
 }
 
+function _destroy_redis {
+    redis_pids=$(ps uxww | grep -E "[r]edis-server" | awk '{print $2}')
```
dbutenhof, line 14: What is `"[r]edis-server"`?? This is matching a required "set" consisting of exclusively "r". If there's some subtle RE distinction from just "redis-server", I'm missing it.

lib/pbench/agent/tool_meister.py
```diff
@@ -1725,7 +1725,8 @@ def sysinfo(self, data: Dict[str, str]) -> int:
         try:
             with o_file.open("w") as ofp, e_file.open("w") as efp:
                 my_env = os.environ.copy()
-                my_env["pbench_install_dir"] = self.pbench_install_dir
+                my_env["sysinfo_install_dir"] = self.pbench_install_dir
+                my_env["sysinfo_full_hostname"] = self._params.hostname
```
dbutenhof, line 1729: Not really complaining; however, since we control both sides of the interface, and you don't want to use the "standard" environment variables here, why not simply add these values to the `pbench-sysinfo-dump` argv instead of creating/reading new environment variables?

dbutenhof (approved):

agent/util-scripts/test-bin/common-tm-cleanup
portante: Prevents matching the `grep` process itself.

lib/pbench/agent/tool_meister.py
portante: I thought about that, but wanted to reduce the changes as much as possible to the non-`pbench-sysinfo-dump` code.

agent/util-scripts/test-bin/common-tm-cleanup
dbutenhof: Fascinating. I can see that it does. I can't see why it would. Technically, in strict RE definition, "r" is exactly the same as `[r]`... an RE that matches exactly the single character "r". There's clearly some semantic distinction I'm missing. ☹️

portante: The `[r]edis-server` != `redis-server`.  The regex will match `redis-server` but not the text string `[r]edis-server`.

portante: This clever little trick is one I remember from the great @agriffis.

dbutenhof: Ha; of course. Tunnel vision. 🙄

portante (commented):
Correction applied from a self-review.

webbnh (approved):
Looks OK to me, but I am concerned about the spelling of the new environment variables.  (And, then there's a nit and a coding suggestion...but I held it to just those three!)

lib/pbench/agent/tool_meister.py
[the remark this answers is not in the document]
portante: I didn't want to use names of environment variables that are provided by `agent/base`.  That is the reason for the change.

`pbench-sysinfo-dump` is now "standalone".

agent/util-scripts/tool-meister/pbench-sysinfo-dump
```diff
@@ -222,9 +242,9 @@ for item in ${sysinfo//,/ }; do
 	elif [[ "${item}" == "stockpile" ]]; then
 		:
 	elif [[ "${item}" == "insights" ]]; then
-                :
+		:
 	else
```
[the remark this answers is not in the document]
portante: I am trying not to change much code, this is a whitespace change that came long with the editor, and I can remove it if necessary, but I'd rather not make another change.

lib/pbench/agent/tool_meister.py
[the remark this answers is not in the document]
portante: Yes, the Tool Meister is the only entity that invokes `pbench-sysinfo-dump`.

webbnh (approved):
👍

agent/util-scripts/tool-meister/pbench-sysinfo-dump
```diff
@@ -1,6 +1,24 @@
 #!/bin/bash
 # -*- mode: shell-script; indent-tabs-mode: t; sh-basic-offset: 8; sh-indentation: 8; tab-width: 8 -*-
 
+# Installation directory.
+install_dir="$(realpath -e ${sysinfo_install_dir} 2> /dev/null)"
```
npalaska, line 5: Is there a specific reason to use `install_dir` instead of `pbench_install_dir`? I thought since this is pbench specific install directory, we would have pbench in its name

npalaska (approved):
Looks good, just one question

agent/util-scripts/tool-meister/pbench-sysinfo-dump
dbutenhof: It doesn't matter here since it's local to the bash script; it's not `export`ed anywhere. Our code would be hard to read if we used `_pbench_` prefixes for *every* local variable. (We actually did that in threads, which was a nightmare, and I really enjoyed getting everyone to tear that out when we went "pthreads native".)

lib/pbench/agent/tool_meister.py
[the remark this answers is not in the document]
dbutenhof: Note that we're only putting this in the environment of the new forked child process, which only runs the sysinfo command; so these environment variables are actually fairly isolated. (Though making them command line arguments would probably be cleaner, I don't think it's a big deal in this context.)

dbutenhof (approved):

agent/util-scripts/tool-meister/pbench-sysinfo-dump
npalaska: > It doesn't matter here since it's local to the bash script; it's not `export`ed anywhere

That make sense

PR #2755, opened 2022-04-14
Ensure `agent/base` is not used by Tool Meisters

portante:
Remove the use of `agent/base` for remote Tool Meisters, and ensure that the test environment is "cleaned" for the "fake" remote Tool Meisters.

Layered on top of PR #2754.


agent/util-scripts/test-bin/ssh
```diff
@@ -21,16 +21,45 @@ done
 remote=${1}
 shift
 
+function mkdirp {
+    if [[ -d ${1}/${2} ]]; then
+        return 0
+    fi
+    (cd ${1}; mkdir -p ${2})
```
webbnh, line 28: This purpose of this is to use `cd` to test for the existence of `${1}` before creating `${2}`?

Shouldn't that `;` be a `&&`??

(It would be nice to have a block comment explaining the motivation for using `mkdirp` instead of `mkdir -p`....)

agent/util-scripts/test-bin/ssh
```diff
@@ -21,16 +21,45 @@ done
...
     exit_code=0
-elif [[ "$(basename -- "${1%% *}")" == "pbench-tool-meister-remote" ]]; then
+elif [[ "$(basename -- "${1%% *}")" == "pbench-tool-meister" ]]; then
     _dir=$(dirname ${0})
-    _pbench_full_hostname="${remote}" _pbench_hostname="${remote}" _tool_bin=${_dir}/mpstat ${1} localhost "${3}" "${4}" "${5}"
-    exit_code=$?
+    # To simulate running a Tool Meister instance "remotely" we reset the
+    # environment presented to the pbench-tool-meister instance. We keep the
+    # test environment variables, PATH, HOME, etc., but we also keep the
+    # pbench_tmp environment variable.  This is convenient for the test system
```
webbnh, line 43: s/also keep/change/ ?

agent/util-scripts/test-bin/ssh
```diff
@@ -21,16 +21,45 @@ done
...
+    # to avoid tracking down arbitrary /var/tmp usage.
+    _remote_dir="remote/${remote}/tmp"
+    mkdirp ${_testdir} ${_remote_dir}
+    _pbench_tmp=${_testdir}/${_remote_dir}
+    env -i \
+        PATH=${PATH} \
+        HOME=${HOME} \
+        PYTHONHASHSEED=${PYTHONHASHSEED} \
+        PARALLEL_TMP=${PARALLEL_TMP} \
```
webbnh, line 52: Where does `PARALLEL_TMP` come from?  What is it used for?  And, why doesn't it have `${_remote_dir}` in it?

agent/util-scripts/test-bin/ssh
```diff
@@ -21,16 +21,45 @@ done
...
+        _PBENCH_UNIT_TESTS=1 \
+        _PBENCH_UNIT_TEST_NAME=${_PBENCH_UNIT_TEST_NAME} \
+        _testdir=${_testdir} \
+        _testhome=${_testhome} \
+        _testlog=${_testlog} \
+        _testout=${_testout} \
+        _real_cp=${_real_cp} \
+        pbench_tmp=${_pbench_tmp} \
+        _tool_bin=${_dir}/mpstat ${1} localhost "${3}" "${4}" "${5}"
```
webbnh, line 61: I feel like we're missing one or more items related to SSH...isn't there a "here's the host I came from" and/or "here's the host I connected to" that we use sometimes?

webbnh (dismissed):
Found a `;` which should probably be a `&&`; other than that, just a few questions.

dbutenhof (approved):

agent/util-scripts/test-bin/ssh
webbnh: Um, OK.  I'd kinda misplaced the fact that we're in the test support stuff here, so it's probably not an issue.

agent/util-scripts/test-bin/ssh
portante: From `parallel`, just passing it along.

portante (commented):
Ready for review!

dbutenhof (approved):

webbnh (approved):
👍

PR #2756, opened 2022-04-14
Employ a `NamedTuple` for TDS parameters

portante:
Like the Tool Meister module, employ a `NamedTuple` for the Tool Data Sink parameters.

Layered on PR #2755.

PR #2757, opened 2022-04-14
Refactor Tool Meister start to consolidate logic

portante:
We apply a small refactoring to the Tool Meister start logic to bring  together the required environment variables in one place, and only  create the local Tool Meister working directory when we are asked to  orchestrate the sub-system setup.

Layered on PR #2756.

### DIFF
--- a/lib/pbench/agent/tool_meister_start.py
+++ b/lib/pbench/agent/tool_meister_start.py
@@ -187,7 +187,7 @@ class ReturnCode(BaseReturnCode):
     """
 
     BADTOOLGROUP = 1
-    BADAGENTCONFIG = 2
+    # Removed BADAGENTCONFIG = 2
     MISSINGBENCHRUNDIR = 3
     MISSINGINSTALLDIR = 4
     EXCINSTALLDIR = 5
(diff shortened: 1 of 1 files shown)


dbutenhof (approved):
Sure, that's fine. This sorta stuff definitely inspires me to take my attempts at refactoring all this code more slowly and maybe just concentrate on Jira for a while.

lib/pbench/agent/tool_meister_start.py
```diff
@@ -187,7 +187,6 @@ class ReturnCode(BaseReturnCode):
     """
 
     BADTOOLGROUP = 1
-    BADAGENTCONFIG = 2
```
webbnh, line 190: If I might request it, instead of just deleting this line, please replace it with a placeholder -- e.g., a comment indicating that the value is currently unused.  This will save future readers from wondering if it was inadvertently omitted, and it will help future developers to find the hole if they want to put a new status in it.

lib/pbench/agent/tool_meister_start.py
```diff
@@ -801,15 +800,43 @@ def start(_prog: str, cli_params: Namespace) -> int:
             ", ".join(_orchestrate_choices),
         )
         return ReturnCode.INVALIDORCHESTRATE
+    if cli_params.orchestrate == "create":
+        orchestrate = True
+    else:
+        assert (
+            cli_params.orchestrate == "existing"
+        ), f"Unexpected --orchestrate parameter, {cli_params.orchestrate!r}"
```
webbnh, line 808: Given the test at line 796, isn't this assertion overkill?  (Alternately, if you're really concerned about the value of not-`"create"`, perhaps the test at line 803 should be checking for not-`"existing"` instead?)

lib/pbench/agent/tool_meister_start.py
```diff
@@ -801,15 +800,43 @@ def start(_prog: str, cli_params: Namespace) -> int:
...
+        if cli_params.redis_server is None or cli_params.tool_data_sink is None:
```
webbnh, line 809: Perhaps
```
        if not cli_params.redis_server or not cli_params.tool_data_sink:
```
because any kind of false-y value there would be bad.

lib/pbench/agent/tool_meister_start.py
```diff
@@ -840,10 +867,12 @@ def start(_prog: str, cli_params: Namespace) -> int:
             exc,
         )
         return ReturnCode.EXCBENCHRUNDIR
-    else:
+    if orchestrate:
         tm_dir = benchmark_run_dir / "tm"
         try:
             tm_dir.mkdir()
+            # All orchestration occurs using the newly created Tool Meister
+            # directory as the current working.
```
webbnh, line 875: s/current working/current working directory/

webbnh (dismissed):
This is good stuff.  I just have a nit, a small request, and a couple of suggestions.

lib/pbench/agent/tool_meister_start.py
portante: Just moved existing code.  PRs to follow up with a change welcome!

lib/pbench/agent/tool_meister_start.py
portante: Just moved existing code.  PRs to follow up with a change welcome!

dbutenhof (approved):

npalaska (approved):

lib/pbench/agent/tool_meister_start.py
webbnh: 👃 🤏 

You "just" moved the existing code to a place where the assertion is no longer quite so valuable....

webbnh (approved):
I'm holding my nose....  😉

portante:
@webbnh, could the automation have moved this out of "Done" to "Reviewer approved" because of your comment?  That seems broken.

webbnh:
> could the automation have moved this out of "Done" to "Reviewer approved" because of your comment?

I suspect that it was because it was actually a "review" (an "approval") not just a comment.  (Let's see if _this_ comment changes its state...I expect it _won't_.)

> That seems broken.

Agreed.

portante:
It looks like a race condition only.

webbnh:
I like that explanation better.

PR #2758, opened 2022-04-14
Add a big-hammer redis server cleanup

portante:


portante (commented):
Hopefully this helps in situations where a redis server is left around from some test failure.

riya-17 (approved):

agent/util-scripts/test-bin/common-tm-cleanup
```diff
@@ -10,12 +10,26 @@ function _kill_pid {
     fi
 }
 
+function _destroy_redis {
+    redis_pids=$(ps uxww | grep -E "[r]edis-server" | awk '{print $2}')
```
dbutenhof, line 14: OK, since I apparently made this comment in a stacked PR without realizing it... 😄 

What does the `[r]` represent here? A mandatory set with a single element should be exactly the same as the single character, right? Why isn't this just `-F redis-server`?

agent/util-scripts/test-bin/common-tm-cleanup
```diff
@@ -10,12 +10,26 @@ function _kill_pid {
...
+    for pid in ${redis_pids}; do
+        openfiles=$(lsof -p ${pid} 2> /dev/null | grep -F ${_testdir})
+        if [[ -z "${openfiles}" ]]; then
```
dbutenhof, line 17: This, by the way, is a rather clever way to identify a Redis we own. 👏🏻

dbutenhof (approved):

agent/util-scripts/test-bin/common-tm-cleanup
portante: It prevents the `grep` from matching itself.

PR #2759, opened 2022-04-14
Simplify PID file names used by TM sub-system

portante:
In anticipation of the "big hammer" Tool Meister sub-system work, we simplify the PID file names used to make them easy to determine.

Layered on top of PR #2757.

### DIFF
--- a/lib/pbench/agent/tool_meister.py
+++ b/lib/pbench/agent/tool_meister.py
@@ -1901,10 +1901,9 @@ def daemon(
         working_dir = tmp_dir
     else:
         working_dir = Path(".")
-    param_key = parsed.key
-    pidfile_name = working_dir / f"{param_key}.pid"
-    d_out = working_dir / f"{param_key}.out"
-    d_err = working_dir / f"{param_key}.err"
+    pidfile_name = working_dir / "tm.pid"
+    d_out = working_dir / "tm.out"
+    d_err = working_dir / "tm.err"
     pfctx = pidfile.PIDFile(pidfile_name)
     with d_out.open("w") as sofp, d_err.open("w") as sefp, DaemonContext(
         stdout=sofp,
--- a/lib/pbench/agent/tool_meister_start.py
+++ b/lib/pbench/agent/tool_meister_start.py
@@ -661,7 +661,7 @@ port {redis_port:d}
         bind_host_names = " ".join(bind_hostnames_l)
 
         # Create the Redis server pbench-specific configuration file
-        self.pid_file = tm_dir / f"redis_{self.bind_port:d}.pid"
+        self.pid_file = tm_dir / "redis.pid"
         redis_conf = tm_dir / "redis.conf"
         params = {
             "bind_host_names": bind_host_names,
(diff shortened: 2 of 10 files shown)


dbutenhof (dismissed):

lib/pbench/agent/tool_meister.py
```diff
@@ -1901,10 +1901,9 @@ def daemon(
         working_dir = tmp_dir
     else:
         working_dir = Path(".")
-    param_key = parsed.key
-    pidfile_name = working_dir / f"{param_key}.pid"
-    d_out = working_dir / f"{param_key}.out"
-    d_err = working_dir / f"{param_key}.err"
+    pidfile_name = working_dir / "tm.pid"
```
webbnh, line 1904: I think that this means that a user cannot have two tool meisters running on the same system using the same (e.g., the default) configuration...is that a problem?  (Should we be trying to detect collisions rather than just overwriting the PID file??)

lib/pbench/agent/tool_meister_start.py
```diff
@@ -660,7 +660,7 @@ def start(self, tm_dir: Path, full_hostname: str, logger: logging.Logger) -> Non
         bind_host_names = " ".join(bind_hostnames_l)
 
         # Create the Redis server pbench-specific configuration file
-        self.pid_file = tm_dir / f"redis_{self.bind_port:d}.pid"
+        self.pid_file = tm_dir / "redis.pid"
```
webbnh, line 663: I think that this means that a user cannot have TM-orchestrated Redis servers running on the same system, even if they are using different ports.  Isn't that a problem? (Again, should we be trying to detect collisions rather than just overwriting the PID file??)

webbnh (approved):
I like the simplification, but I'm concerned that it might pose a problem.

lib/pbench/agent/tool_meister.py
portante: We always ensure the working directory used by the "create" orchestration path is unique.

lib/pbench/agent/tool_meister_start.py
portante: We always ensure the working directory used by the "create" orchestration path is unique.

webbnh: OK, so it's based on `benchmark_run_dir`, and anyone who runs two workloads out of the same directory at the same time is already in trouble.

Nevertheless, the PID file should not exist before we create it (right??), so it might be worth adding that check.

lib/pbench/agent/tool_meister.py
webbnh: OK, so it's based on `benchmark_run_dir`, and anyone who runs two workloads out of the same directory at the same time is already in trouble.

Nevertheless, the PID file should not exist before we create it (right??), so it might be worth adding that check.

webbnh (dismissed):
👍

lib/pbench/agent/tool_meister.py
portante: Thankfully that is handled by the `pidfile` module.

lib/pbench/agent/tool_meister_start.py
portante: Redis owns this behavior already.

portante (commented):
Ready for review!

webbnh (commented):
👍

dbutenhof (approved):

webbnh (approved):
👍

PR #2760, opened 2022-04-14
Add UUIDs to command line of running Tool Meisters

portante:
In order to facilitate a "big-hammer" approach to cleaning up wayward Tool Meister sub-systems, we add UUIDs to the command line invocations of the remote Tool Meister processes.  We don't need to do that for the local Redis server, Tool Data Sink, and Tool Meister process because we have their PID files already.

We do take this opportunity, though, to verify the UUID given on the command line against the running parameters stored in Redis.

----

Layered on PR #2797.


portante (commented):
Ready for review!

lib/pbench/agent/tool_data_sink.py
```diff
@@ -762,10 +763,16 @@ def fetch_params(params, pbench_run):
...
+                instance_uuid=params["instance_uuid"],
             )
         except KeyError as exc:
             raise ToolDataSinkError(f"Invalid parameter block, missing key {exc}")
         else:
+            if instance_uuid != tdsp.instance_uuid:
+                raise ToolDataSinkError(
+                    f"Parameter block has unexpected UUID '{tdsp.instance_uuid}',"
+                    f" expected '{instance_uuid}'"
+                )
```
webbnh, line 775: Why is this code here instead of in the caller?  In order to have it here, we have to add a parameter to the function signature.  It would be simpler (and less coupled) to just check in the caller.

lib/pbench/agent/tool_meister_start.py
```diff
@@ -882,6 +888,17 @@ def start(_prog: str, cli_params: Namespace) -> int:
                 exc,
             )
             return ReturnCode.EXCCREATETMDIR
+        uuid_file = tm_dir / ".uuid"
+        instance_uuid = str(uuid.uuid4())
+        try:
+            uuid_file.write_text(instance_uuid)
```
npalaska, line 894: Should we make this file immutable for other users by running `uuid_file.chmod(644)`? or is it not necessary?

npalaska (approved):
Looks good, just one question

lib/pbench/agent/tool_meister_start.py
dbutenhof: There are plenty of ways for people to mess with us if they really want to; that seems like excessive paranoia to me.

dbutenhof (approved):

lib/pbench/agent/tool_data_sink.py
```diff
@@ -748,7 +749,7 @@ class ToolDataSink(Bottle):
     _data_actions = frozenset(("send", "sysinfo"))
 
     @staticmethod
-    def fetch_params(params, pbench_run):
+    def fetch_params(params, pbench_run, instance_uuid):
```
webbnh, line 752: Type hints?

lib/pbench/agent/tool_data_sink.py
```diff
@@ -2060,7 +2073,7 @@ def start(prog: Path, parsed: Arguments):
         # E.g. params = '{ "channel_prefix": "some-prefix",
         #                  "benchmark_run_dir": "/loo/goo" }'
         params = json.loads(params_str)
-        ToolDataSink.fetch_params(params, pbench_run)
+        ToolDataSink.fetch_params(params, pbench_run, parsed.instance_uuid)
```
webbnh, line 2076: What does this line accomplish?  (Doesn't `fetch_params()` return a value...which we're ignoring here...so what side-effect are we depending on?)

webbnh (changes requested):
I have to step out for a moment, but I wanted to get these points out there.

lib/pbench/agent/tool_data_sink.py
portante: Validates the parameters are what we expect before we go through the hassle of daemonizing and instantiating a full blown instance of the Tool Data Sink.  This is existing code behavior.  PRs to suggest a change to the way this work are welcome!

lib/pbench/agent/tool_data_sink.py
portante: But we there are two callers and we wanted the knowledge of valid parameters in one place.

lib/pbench/agent/tool_data_sink.py
webbnh: It seems that we've blended some concerns which perhaps should be separated:
- here, we want to be able to validate the `params` mapping
- in the other call location, we want to allocate and initialize a `ToolDataSinkParams` (perhaps after validation).

The fact that much of the validation is done simply by accessing the `params` mapping doesn't change this.  And, that quirk of the implementation is not a reason to add a bunch of stuff to the allocation function which should be packaged separately.  (Arguably, the allocation function should be taking a `BenchmarkRunDir` instead of `pbench_run`; or, the caller should be adding it to the returned object.)

Would it be better if we converted the `params` mapping to a `ToolDataSinkParams` object up front, and then passed the object along, so it doesn't have to be (re)created later?

portante: But is this PR the place to do that kind of refactoring?

These changes are very simple to follow and understand, albeit not ideal.  Perhaps a follow-on commit could correct that?  Posted PRs are always welcome!

npalaska (approved):

ndokos (approved):

dbutenhof (dismissed):

dbutenhof (dismissed):

npalaska (approved):

dbutenhof (commented):
I've got a review request pending, but Webb's spawned off that last commit into a separate PR, and as it stands this doesn't pass the tests anyway, so for now I'm just clearing the review request and acknowledging that I looked it it. 🍤

webbnh (changes requested):
I'd like to see this rebased on #2766, once that is merged.  Hopefully, once that is done, we can keep the UUID checks in the caller instead of placing them in the fetch-params functions, and then this will be good to go.

npalaska (approved):

portante (commented):
I backed out @webbnh's changes because he has filed PR #2766 separately.

lib/pbench/agent/tool_meister_start.py
```diff
@@ -229,6 +230,7 @@ class ReturnCode(BaseReturnCode):
     KEYBOARDINTERRUPT = 41
     INVALIDTMDATA = 42
     TOOLINSTALLFAILURES = 43
+    EXCCREATEUUID = 44
```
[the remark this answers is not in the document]
dbutenhof: I'd rather not develop a convention of "going back and filling in" ... to the extent that these numbers have any real value in the field, we're depending on them being unique. Ideally, they're not so version-dependent that we need to do a lot of research to figure out what an exit status means.

dbutenhof (approved):

webbnh (changes requested):
@portante, since you are declining to base this PR on #2766, I believe that that work should then be done in this PR, instead.  If we "hold our noses and git-r-done", then we're just creating more technical debt, and we would be doing so in order to meet an artificial deadline.  (We would be following the old software development adage once again that "there's never enough time to do it right, but there's always enough time to do it over.")

The changes to `tool_data_sink.py` and `tool_meister.py` proposed in this PR involve adding a parameter to every function in the call chain from `start()` down to `fetch_params()` so that `fetch_params()` can validate the instance UUID against the other parameter that it receives.  This is a staggering amount of change, considering that it could be avoided by doing that validation at the top level where both values are already available, and where it would support the doctrine of "fail early".

Delegating the check of the UUID value to `fetch_params()` also leads to the idiom of calling the function twice and throwing away the result of the first call, in order to avoid duplicating the validation.  This looks silly, too.

There's a similar problem with delegating the instantiation of the `BenchmarkRunDir` to `fetch_params()`.  Granted, that wasn't introduced by this PR, but the solution to that problem is substantially the same as the one for validating the UUID value, so it is not inappropriate to address it at the same time, since doing so touches all the same parts of the code.  But, we can quibble over whether it is appropriate to make that change at this time.

You proposed a modification to my changes for this.  Perhaps that should be incorporated here (although, I would recommend against including `PROG` in `ExternalEnvironment` and I don't see the point of adding the `optional_md` attribute to the `ToolDataSink` object -- that results in significant churn for only the tiniest gain).

portante (commented):
Ready for review.

webbnh (dismissed):
Looks good.  I just have one confirmation and one suggestion.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -2069,6 +2072,14 @@ def start(prog: Path, parsed: Arguments):
...
         )
         return 6
+    else:
+        if parsed.instance_uuid != tdsp.instance_uuid:
+            print(
+                f"Parameter block has unexpected UUID '{tdsp.instance_uuid}',"
+                f" expected '{parsed.instance_uuid}'",
+                file=sys.stderr,
+            )
+            return 6
```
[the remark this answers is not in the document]
portante: I had thought about using a different number here, but it seemed close enough to the meaning of the previous use that I kept it.

dbutenhof (dismissed):

webbnh (approved):
👍

PR #2761, opened 2022-04-14
Small changes deferred from recent PRs

webbnh:


portante (approved):

npalaska (approved):

dbutenhof (approved):

lib/pbench/agent/tool_meister_start.py
```diff
@@ -801,20 +801,17 @@ def start(_prog: str, cli_params: Namespace) -> int:
...
+        if not cli_params.redis_server or not cli_params.tool_data_sink:
             logger.error(
                 "both --redis-server and --tool-data-sink must be specified"
                 " if --orchestrate=%s is used",
                 cli_params.orchestrate,
             )
             return ReturnCode.MISSINGPARAMS
         orchestrate = False
+    else:
+        orchestrate = True
```
ndokos, line 814: With the `assert` gone, we've lost the checking of the `orchestrate` option - does that matter?

I actually prefer a structure like this:
```
     if create:
            orchestrate = True
     elif  existing:
            if redis and tds:
                  orchestrate = False
            else:
                   log error(both must be given)
      else:
             log bad option error
``` 
avoiding the negations (which always give me a headache) and checking for a bad option (OTOH, this may be done elsewhere in which case the final `else` is redundant - but it doesn't hurt to leave it in). I find that checking for "normal cases"  first is easier to understand and leaves errors to be handled in the else's. My $0.02...

ndokos (approved):

lib/pbench/agent/tool_meister_start.py
ndokos: Feel free to resolve this conversation without waiting for me to resolve it if you disagree.

portante: I think we solve that with https://github.com/distributed-system-analysis/pbench/blob/948564b5698bec1e2e6da886d108583d653c6dc4/lib/pbench/agent/tool_meister_start.py#L1315.

PR #2762, opened 2022-04-14
PBENCH-199 access server endpoints API, with local dev mode nodeJS mirror

dbutenhof:
This includes two pieces:

1. We introduce an early call to the server's /api/v1/endpoints API to identify the full URL for each functional API endpoint;
2. We introduce a nodeJS "mirror" that will run under local npm development mode to gather the endpoint configuration of a remote Pbench server.

For now, we demonstrate that we successfully fetched the endpoint definitions by adding the server's identification string to the "Pbench Dashboard" text. Essentially, to call `datasets/list` you'd now reference the API map as `window.endpoints.api.datasets_list`.

The idea here is that `.env` defines a `PBENCH_SERVER` environment variable to point to the actual Pbench server when running the dashboard in local dev mode. The `index.html` was "hacked" to read the server endpoints definition from `/api/v1/endpoints`. In a production mode, where the dashboard `index.html` and Javascript are loaded from the server, this will resolve directly to the server endpoints definitions. In dev mode (`npm run dev`) we'll start up a small nodeJS express "mirror" that simply listens on `localhost:3001/api/v1/endpoints`, makes the call to the endpoints API on the configured Pbench server, and returns the results. The `index.html` stores this data on the Javascript window object as `window.endpoints` for use anywhere in the Pbench dashboard application. (This was just the way it worked out in the old dashboard, from which I lifted the `index.html` code; if you have a better idea, let's do it.)

This adds a `"proxy"` definition to the `package.json`, which tells npm to proxy access to port 3001 (the express app) through the primary dashboard port (usually 3000) so the app will respond to the request on `window.origin`. This also adds `nvm` as a dependency, which (in the process of floundering to run this with apparently a very old npm) I found to ensure I was getting the right npm code.

### DIFF
--- a/dashboard/public/index.html
+++ b/dashboard/public/index.html
@@ -30,6 +30,16 @@
   </head>
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
+    <!-- runtime endpoints config -->
+    <script type="text/javascript">
+      fetch("/api/v1/endpoints").then((response) => {
+        if (response.ok) {
+          response.text().then((data) => {
+            window.endpoints = JSON.parse(data);
+          });
+        }
+      });
+    </script>
     <div id="root"></div>
     <!--
       This HTML file is a template.
--- a/dashboard/server/index.js
+++ b/dashboard/server/index.js
@@ -1,0 +1,21 @@
+const express = require('express');
+const axios = require('axios');
+require('dotenv').config()
+
+
+const app = express()
+app.get('/api/v1/endpoints', (req, res) => {
+    console.log('Mirror engaged: ', process.env.PBENCH_SERVER)
+    axios.get(
+            `${process.env.PBENCH_SERVER}/api/v1/endpoints`,
+            { headers: { 'Accept': 'application/json' } })
+        .then(endpoints => {
+            res.setHeader('Content-Type', 'application/json');
+            res.send(endpoints.data);
+        })
+        .catch(err => {
+            console.log('Error: ', err.message);
+        })
+})
+
+app.listen(3001, () => console.log('Mirror server running on localhost:3001'));
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -9,6 +9,7 @@
     "@testing-library/react": "^12.1.4",
     "@testing-library/user-event": "^13.5.0",
     "axios": "^0.26.1",
+    "nvm": "0.0.4",
     "patternfly": "^3.59.5",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
@@ -21,9 +22,12 @@
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
+    "server": "node server",
+    "dev": "npm run server & npm run start",
     "test": "react-scripts test",
     "eject": "react-scripts eject"
   },
+  "proxy": "http://localhost:3001",
   "eslintConfig": {
     "extends": [
       "react-app",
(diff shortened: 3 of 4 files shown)


dashboard/server/index.js
```diff
@@ -0,0 +1,21 @@
+const express = require('express');
+const axios = require('axios');
+require('dotenv').config()
+
+
+const app = express()
+app.get('/api/v1/endpoints', (req, res) => {
+    console.log('Mirror engaged: ', process.env.PBENCH_SERVER)
+    axios.get(
+            `${process.env.PBENCH_SERVER}/api/v1/endpoints`,
```
shubham-html-css-js, line 10: Looks good,I have just one question
Where are we going to deploy our env variables?

dbutenhof: This is only for local dev mode, so there's nothing to "deploy" ... you've got your own `.env` in your local development tree, which takes effect while you're testing locally against a remote server. But when we deploy on a real server, we don't need the nodeJS reflector (or the `.env`) at all; the `index.html` "hack" here just loads the endpoints from `window.origin` by default, and all is good with the world.

portante (approved):

dashboard/public/index.html
```diff
@@ -30,6 +30,16 @@
...
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
+    <!-- runtime endpoints config -->
+    <script type="text/javascript">
+      fetch("/api/v1/endpoints").then((response) => {
+        if (response.ok) {
+          response.text().then((data) => {
+            window.endpoints = JSON.parse(data);
+          });
+        }
```
webbnh, line 40: Should we be doing something if the response is _not_ OK?  (I don't want to leave our developers wondering what happened, if, for example, the express server didn't come up or something.)

webbnh (approved):
Looks good to me (as far as I can tell... 😉).  Just one pointed question.

portante (approved):

dashboard/public/index.html
dbutenhof: The nodeJS reflector isn't really the issue; yes, that's one *additional* failure mode in local npm developer mode, but, in theory, this could fail on a real server if the HTTP server (or reverse proxy) is down. Hmm. (And if it fails in dev mode because the nodeJS reflector wasn't started, it'll look just like an unreachable host anyway, because that's what it is.)

This code was copied directly from the old dashboard, and my changes *there* were directly descended from the original code to load the server configuration JSON file. However you're correct that there's no error handling here. (Which actually surprised me, because I'd thought there was. Huh.) I think this means we never had any error handling if the config file was missing or otherwise unavailable, I didn't notice and nobody pointed that out when I changed it. Intriguing.

I'm not quite sure what we could do except splash up an error message.

@MVarshini mentioned that she was interested in integrating the endpoints load into the Redux level rather than putting it out here in `index.html` (which has always made me a bit uncomfortable). If we can't load our UI app code, there's no point in worrying about the endpoints; and, if we can load the code and get it started, but can't load endpoints, we may be in a position at that level to at least generate a better looking error message. 😄 

If @MVarshini wants to take in this PR more or less "as is" instead of reworking it, I'll add an `else` that at least renders the HTTP response on the screen.

webbnh: I'm happy to defer to either you or @MVarshini.

I'm just imagining myself trying to run the dashboard locally...and, when I've screwed it up, I'm hoping that there will be helpful feedback....  😕

portante (changes requested):
Seems like we need to resolve the concerns with this API for the long term before moving forward.

portante:
@dbutenhof, why was this PR closed?  Did #2871 replace it?

portante:
> @dbutenhof, why was this PR closed? Did #2871 replace it?

Seems so.

PR #2763, opened 2022-04-14
Archive the input tool group data during start

portante:
We archive the valid tool group data when we start the Tool Meister sub-system so that we have a record of the inputs used during orchestration.  We don't want to rely on the Tool Data Sink to save that for us.


webbnh (dismissed):
I assume that these changes just foundational and that the _uses_ of the archive will come in a subsequent PR.

portante:
> I assume that these changes just foundational and that the _uses_ of the archive will come in a subsequent PR.

In part.

Having the data archived means we don't have to rely on the Tool Data Sink getting to the point where it saved the tool data in the `metadata.log` file when a problem comes up.

portante (commented):
Ready for review.

dbutenhof (approved):

riya-17 (approved):

PR #2765, opened 2022-04-20
Clarify that md5sum() takes a Path not a string

webbnh:
It turns out that _almost_ all callers of `md5sum()` pass in a `Path`, despite the fact that we declare it to accept a `str`.  Python's "duck typing" allows this to work, because what we actually require is a "path-like", and both `str`'s and `Path`'s qualify...but it's unsatisfying.  This PR makes it clear that `md5sum()` accepts a `Path` (and fixes the two inconsistent usages).

### DIFF
--- a/lib/pbench/common/utils.py
+++ b/lib/pbench/common/utils.py
@@ -5,6 +5,7 @@ from collections import deque
 import hashlib
 import ipaddress
 from logging import Logger
+from pathlib import Path
 import re
 from typing import Callable, Deque, NamedTuple
 
@@ -16,7 +17,7 @@ class Md5Result(NamedTuple):
     md5_hash: str
 
 
-def md5sum(filename: str) -> Md5Result:
+def md5sum(filename: Path) -> Md5Result:
     """
     Return the MD5 check-sum of a given file without reading the entire file
     into memory.
--- a/lib/pbench/cli/agent/commands/results/push.py
+++ b/lib/pbench/cli/agent/commands/results/push.py
@@ -1,3 +1,5 @@
+from pathlib import Path
+
 import click
 
 from pbench.agent.base import BaseCommand
@@ -37,7 +39,14 @@ class ResultsPush(BaseCommand):
 @click.argument("controller")
 @click.argument(
     "result_tb_name",
-    type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True),
+    type=click.Path(
+        path_type=Path,
+        exists=True,
+        file_okay=True,
+        dir_okay=False,
+        readable=True,
+        resolve_path=True,
+    ),
 )
 @pass_cli_context
 def main(
--- a/lib/pbench/test/unit/agent/task/test_make_result_tb.py
+++ b/lib/pbench/test/unit/agent/task/test_make_result_tb.py
@@ -119,7 +119,7 @@ class TestMakeResultTb:
         assert tarball.samefile(expected_tb), f"{tarball} {expected_tb}"
         assert tarball.exists()
         assert tarball.stat().st_size == tarball_len and tarball_len > 0
-        calc_len, calc_md5 = md5sum(str(tarball))
+        calc_len, calc_md5 = md5sum(tarball)
         assert tarball_len == calc_len
         assert calc_md5 == tarball_md5
         with tarfile.open(str(tarball), "r:xz") as tf:


npalaska (dismissed):

lib/pbench/common/utils.py
```diff
@@ -16,7 +17,7 @@ class Md5Result(NamedTuple):
     md5_hash: str
 
 
-def md5sum(filename: str) -> Md5Result:
+def md5sum(filename: Path) -> Md5Result:
```
dbutenhof, line 20: This is interesting because the function is actually using `filename` as a string in `open(filename, mode="rb") as f`. If we're continuing with that implementation, `filename: str` would be more appropriate, or at least an explicit `open(str(filename)...)` (Possibly with `filename: Union[str,Path]` to document explicitly that we'll take either.)

When we have a `Path`, I prefer using the path methods: and the intent would be better represented here `filename.open(mode="rb")`.

dbutenhof (changes requested):

lib/pbench/common/utils.py
webbnh: > the function is actually using `filename` as a string in `open(filename, mode="rb")`

According to the docs, the `file` parameter to [`open()`](https://docs.python.org/3/library/functions.html#open) takes a [path-like object](https://docs.python.org/3/glossary.html#term-path-like-object).  And `Path` objects, oddly enough, [implement](https://docs.python.org/3/library/pathlib.html#pure-paths) the pathlike interface. 🙂

So, I don't think we need to have `filename` be a `str`.

> When we have a Path, I prefer using the path methods

I'm generally in agreement with that, so I'm willing to change it if you still want me to after reading the first part of this comment, but I don't think it's needed.  (E.g., I bet that `Path.open()` calls the `open()` builtin, so the code as it probably cuts out a step.)

dbutenhof: Of the 8 direct calls I found to `md5sum`, at least one (`cli.agent.commands.results.push.ResultsPush.execute`) still uses `str` (although it *is* possible to have click instantiate a `click.Path` as a Python `Path` via `path_type=Path`, it's not the default), along with the one you fixed.

So if you'd rather have `md5sum` require a `Path` but accept the "pathlike" `os.open`, you should fix the `ResultsPush` use as well. Or you could just change the type hint to declare that `filename` is pathlike and leave the calls alone. (Which, incidentally, turns out to be `Union[str, bytes, os.PathLike]`.) As currently coded, it'll work but the type hint is wrong. (Or at least one of the callers is wrong, depending on how you choose to look at it.)

If we're requiring `Path`, I'd prefer to use `Path.open` instead of `os.open`; but you're right that it's not a big deal and I won't follow up on that as long as the type hint and usage are consistent.

portante (approved):

riya-17 (dismissed):

lib/pbench/common/utils.py
portante: This does not seem worth changing.  Why can't we declare our `md5sum()` as requiring a `Path` and just be done?

dbutenhof: While that argument isn't invalid, if we're really going to take the approach that "it's wrong but it works so it doesn't matter" then there's no reason for this PR at all; why not just leave it "str"?

portante: Or perhaps we should sub-class `Path` to add an `md5sum()` method.

webbnh: > Of the 8 direct calls I found to `md5sum`, at least one (`cli.agent.commands.results.push.ResultsPush.execute`) still uses `str` (although it _is_ possible to have click instantiate a `click.Path` as a Python `Path` via `path_type=Path`, it's not the default)

@dbutenhof, could you confirm this?  I think you're referring to [this line](https://github.com/distributed-system-analysis/pbench/blob/bceae2c295b78cf2b0240d087d0466c94c74df47/lib/pbench/cli/agent/commands/results/push.py#L15), which is referencing `self.context.result_tb_name` which I think is created by [this definition](https://github.com/distributed-system-analysis/pbench/blob/bceae2c295b78cf2b0240d087d0466c94c74df47/lib/pbench/cli/agent/commands/results/push.py#L38) which (I think...) results in a type of `Path`.

So, I _think_, other than the one that I changed, all the values passed were already `Path`'s.

I'm open to making our `md5sum()` take a path-like.  (Subclassing `Path` is clever, but it seems like too much.)  But, with the changes in this PR, I think I have all the code consistent.  (Nevertheless, if you want, I can certainly change `open()` to `Path.open()`, but, as I said, I don't think that's needed and I'm not sure that it's better.)

dbutenhof: No; as I said, `click.Path` makes click verify that the string value represents a file system path, but the result is a Python `str`, not a Python `Path`, unless you also specify the `path_type=Path` parameter. (Assuming `from pathlib import Path`, of course.)

I ran into this in #2611 ... [lib/pbench/cli/server/tree_manage.py](https://github.com/distributed-system-analysis/pbench/blob/fbb0ad2f94e34d9f9e901398ece6b71c2ebce727/lib/pbench/cli/server/tree_manage.py)

webbnh: Yeah, this is the sort of thing which makes me yearn for C++ and strict (static) typing.  I mean, I love Python and all, but why would anyone write anything _real_ in it??

I'll see if I can fix `ResultsPush`....  (Thanks!...)

portante: Actually, the language du jour is Rust now ...

portante (approved):

dbutenhof (approved):
👍🏻

lib/pbench/common/utils.py
webbnh: Does it do static/build-time typing?

portante: I believe so, but I am a bit rusty at this point.  ;)

riya-17 (approved):

PR #2766, opened 2022-04-20
Streamline parameter passing and front-load parameter object creation

webbnh:
This PR is a re-do of [portante/pbench #19](https://github.com/portante/pbench/pull/19), expanded to include the TM as well as the TDS, and excluding the UUID changes.  The intention is that #2760 should be layered on top of this PR.

This PR front-loads the creations of the `ToolDataSinkParams` and `ToolMeisterParams` objects.  This allows errors in their constructions to be detected up-front without having to undergo the constructions twice; the constructed object is then passed down to the subroutines instead of passing the parameter-block et al. which would otherwise have been used to construct and validate them.


lib/pbench/agent/tool_meister.py
```diff
@@ -1324,9 +1341,9 @@ def _create_tar(
         Returns the CompletedProcess object returned by subprocess.run.
         """
 
-        def tar(tar_args: List):
+        def tar(tar_args_: List):
```
dbutenhof, line 1344: Trailing underscore? Why?

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1889,7 +1889,15 @@ def driver(
...
 
-    logger.debug("params_key (%s): %r", parsed.key, tdsp)
+    logger.debug(
+        "params_key (%s): %r",
+        parsed.key,
+        {
+            k: v
+            for k, v in sorted(tdsp._asdict().items())
+            if isinstance(v, (dict, list, tuple, str, int, float, True, False, None))
+        },
```
dbutenhof, line 1899: Why not just put an `__str__` on the `NamedTuple` and do this once rather than dragging it out here twice?

lib/pbench/agent/tool_meister.py
webbnh: `tar_args` is a variable defined in the outer scope, which is passed into this inner function as a parameter.  One option would be to remove the parameter from this function and just reference the outer scope variable; however, since we do change its value between calls, I think it's better as an explicit parameter.  But, that creates potential confusion, since the reader has to realize that the inner reference is to the parameter and not the outer scope variable.  So, the trailing underscore disambiguates the two (and highlights for the reader that something interesting is going on here).

(The real reason is, my IDE objected, and I couldn't come up with a better name for the parameter, since it really does mean the same thing as the value declared in the outer scope; and, since PEP-8 provides for [using a trailing underscore](https://peps.python.org/pep-0008/#function-and-method-arguments) in a similar situation (see where I did it to `dir`), I decided to do it here, too.)

lib/pbench/agent/tool_data_sink.py
webbnh: That's a very, very good question.  😁

webbnh: OK, I did it.  I'm not completely happy with the result, but it's probably much better than doing it in-line.  I ran into a challenge over the dichotomy between `__str__()` and `__repr__()`, but, owing to the difference in their respective semantics, I decided to stick with `__str__()` and change the call site to invoke it instead of the other one.

Oh, and, FWIW, we don't get to "do this once" (unless we refactor my change and put it into a utility function (or a superclass), which we may well want to do...), because the two references go to two different classes, so I need two separate `__str__()` implementations....

lib/pbench/agent/tool_meister.py
dbutenhof: Neither is a Python keyword, so the PEP-8 rule doesn't really apply, and the odd convention immediately makes the reader (me) stop and wonder "what's going on"? Your actual conflict is by no means obvious, and the mystery is counter-productive to understanding. 😦 An explanation (perhaps in a docstring) would be a better solution.

> single_trailing_underscore_: used by convention to avoid conflicts with Python keyword

lib/pbench/agent/tool_data_sink.py
```diff
@@ -738,6 +739,21 @@ class ToolDataSinkParams(NamedTuple):
...
+        filter out any object references, because their addresses vary from run
+        to run.
+        """
+        return str(
+            {
+                k: v
+                for k, v in sorted(self._asdict().items())
+                if isinstance(v, (dict, list, tuple, str, int, float, bool, type(None)))
+            }
+        )
```
dbutenhof, line 755: In Java there's a standard utility class that builds a generic string representation of an object by introspection for precisely this purpose (it's inefficient, but often convenient); I wonder if Python has such a thing? (Note, a quick search didn't reveal such a beast, unless it's nested within `inspect`.) If not, perhaps you could factor your two loops out into our own "generic string" helper?

The `NamedTuple` "class" is actually a weird meta-class, and I'm not sure you could get away with building a common base class for these two. (I remember looking at one point, and documentation suggested that was a "no".) At least, I think it would require more than the normal annotation.

webbnh: Python is happy to produce a string representation, via either `__str__()` or `__repr__()`, depending on "details"...but they tend to end up with addresses in them, which is what I'm trying to avoid.  (I _could_ take their output and apply a regex substitution to them, but that seems...heavy-handed?)

I'll see how my next attempt looks.

lib/pbench/agent/tool_meister.py
webbnh: I'm happy to add an explanation, but, if the names aren't different, then the linter complains, so we need _something_ to address that.

My reading of PEP-8 doesn't conflict with what you quote there, but it doesn't seem to be limited to only that case.  They indicate that appending an underscore is preferable to using abbreviated or "corrupted" spelling.  Given that what I'm trying to do here is avoid a name collision with an outer scope, it seemed like the PEP-8 guidance applied.  (But, if you want to offer an alternate spelling which isn't abbreviated or corrupted, we can go that way.)

Peter likes leading underscores, but I don't...and that's already associated with a convention, one which doesn't apply here any better than using a trailing one.

agent/util-scripts/gold/test-start-stop-tool-meister/test-51.txt
```diff
@@ -125,7 +125,7 @@ install_check_output = perf: perf is installed
...
...: 'pbench-agent-cli', 'port': 8080, 'tool_group': 'default', 'tool_metadata': "{'persistent': {'dcgm': {'collector': 'prometheus', 'port': '9400'}, 'node-exporter': {'collector': 'prometheus', 'port': '9100'}, 'pcp': {'collector': 'pcp', 'port': '44321'}}, 'transient': {'blktrace': None, 'bpftrace': None, 'cpuacct': None, 'disk': None, 'dm-cache': None, 'docker': None, 'docker-info': None, 'external-data-source': None, 'haproxy-ocp': None, 'iostat': None, 'jmap': None, 'jstack': None, 'kvm-spinlock': None, 'kvmstat': None, 'kvmtrace': None, 'lockstat': None, 'mpstat': None, 'numastat': None, 'oc': None, 'openvswitch': None, 'pcp-transient': None, 'perf': None, 'pidstat': None, 'pprof': None, 'proc-interrupts': None, 'proc-sched_debug': None, 'proc-vmstat': None, 'prometheus-metrics': None, 'qemu-migrate': None, 'rabbit': None, 'sar': None, 'strace': None, 'sysfs': None, 'systemtap': None, 'tcpdump': None, 'turbostat': None, 'user-tool': None, 'virsh-migrate': None, 'vmstat': None}}", 'tool_trigger': None, 'tools': {'testhost.example.com': {'mpstat': '', 'perf': '--record-opts="-a -freq=100 -g --event=branch-misses --event=cache-misses --event=instructions" --report-opts="-I -g"'}}}
```
webbnh, line 128: In case anyone is wondering and silly enough to take my word for it, the differences here are as follows:
- the `group` key is now `tool_group` (its value is unchanged)
- the `optional_md` key (and dictionary value) are no longer present
- the `port` key now appears earlier in the list (because the `group`/`tool_group` key now appears after it)
- the value of the `tool_metadata` key is now a stringized dictionary instead of just a dict.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -2060,7 +2062,9 @@ def start(prog: Path, parsed: Arguments):
         # E.g. params = '{ "channel_prefix": "some-prefix",
         #                  "benchmark_run_dir": "/loo/goo" }'
         params = json.loads(params_str)
-        ToolDataSink.fetch_params(params, pbench_run)
+
+        benchmark_run_dir = BenchmarkRunDir(params["benchmark_run_dir"], pbench_run)
```
webbnh, line 2066: I don't want to gloss over one "rough edge" here relative to the way Peter had structured them: if `"benchmark_run_dir"` is not present in the JSON object, then we'll get a `KeyError` raised here rather than a `ToolDataSinkError`.

The error will be handled with reasonable grace either way, and I think the slight inconsistency is worth the increased simplicity in the code. @portante, if you disagree, I can add either a try block or an explicit key-check before accessing params. However, I think we can just ensure that the key will be there via testing.

lib/pbench/agent/tool_meister.py
webbnh: I've added the explanation...but it appears in the diffs _after_ this comment...just FYI so you don't miss it.  😉

webbnh (commented):
Notes for reviewers.

lib/pbench/agent/tool_data_sink.py
webbnh: OK, what I've got is, I believe, adequate for the immediate need.  It's nicely encapsulated, so we can expand upon or replace it easily in the future.

lib/pbench/agent/tool_data_sink.py
webbnh: OK, we now have two stub `__str__()` functions which call a utility function to do the real work, and it seems adequate for the moment.

lib/pbench/agent/tool_data_sink.py
portante: Why not just have the `ToolDataSinkParams` field be a string, and then let the `ToolDataSink` constructor convert it to a `BenchmarkRunDir` object?

lib/pbench/agent/tool_meister.py
portante: What about this being named `args`?  Since the context of being the `args` of the `tar` method is sufficient to communicate intent while disambiguating.

While we are here, why does the `tar` method inherit `directory` but not `tar_args`?

It would seem that this is a great opportunity to move the `tar` method to a utility function to avoid all these naming problems along with the inheritance issue.

lib/pbench/agent/tool_meister.py
```diff
@@ -1628,8 +1654,8 @@ def end_tools(self, data: Dict[str, str]) -> int:
             tool_dir,
         )
         unexpected_files = []
-        for dir, _, files in os.walk(tool_dir):
-            dirpath = Path(dir).relative_to(tool_dir)
+        for dir_, _, files in os.walk(tool_dir):
+            dirpath = Path(dir_).relative_to(tool_dir)
```
portante, line 1658: Why is this trailing underscore needed?  Could we use `dirent` instead?

portante (changes requested):
We need to discuss a bit more what is happening here.

If one person's IDE has a linter that is not available to everybody else, should we be making changes based on that?

If our `flake8` and `black` are not sufficient, then we should be able to add a linter in the same way so that all code passes the same test.

Can we consider getting that change in first before we start making so many changes based on an obscure linter?

There is a lot of change here before the end of the release, and none of it solves some bug.  All of these changes are good and desirable, but given how close we are to the release are they really worth taking in now?  It seems we really need to raise the bar higher for changes  so that we reduce the risk of destabilizing the release.

We have been working on reducing the size of commits as well.  It might be worth to break this up into:

 1. Add a linter run as part of the pre-commit check and `tox` along side `flake8` and `black`, fixing everything the new linter finds
 2. Streamline parameter passing and front-load parameter object creation

lib/pbench/agent/tool_meister.py
dbutenhof: > Peter likes leading underscores, but I don't...and that's already associated with a convention, one which doesn't apply here any better than using a trailing one.

I don't like leading *or* trailing underscores, (at best, leading underscores are a hack because Python lacks true `private` properties and methods.) but I like trailing underscores much less because they're even easier to miss and the convention is very obscure.

And I like Peter's suggestion much better. 💯

dbutenhof (dismissed):
I'm not thrilled about the `canonicalize` name, but I've never seen another name for this sort of thing that has struck me as particularly useful; `ReflectionToStringBuilder` and `GsonBuilder` don't exactly roll off the tongue either, though I like the standard Java's explicit reference to "reflection".

`named_tuple_string_builder` is pretty ugly, too. Ah, well.

I get Peter's comments about the hyper linter you're using; it wouldn't be a bad idea to at least identify that somewhere. (My vscode has started exposing dust bunnies that LGTM and `flake8` don't see, but frankly I'm not even sure what it's using.) I like the idea of fixing them where they're obviously real issues regardless of whether anyone else would have seen them; at worst, it does no harm.

He does have a point that if there are better lint filters it'd be nice to have the build using them... if we can figure out how to make that happen.

lib/pbench/agent/tool_meister.py
webbnh: > What about this being named `args`? Since the context of being the `args` of the `tar` method is sufficient to communicate intent while disambiguating.

`args` it is!  (See..._this_ is why we do reviews!  😁)

> While we are here, why does the `tar` method inherit `directory` but not `tar_args`?

Both values _could_ be accessed the same way; however, since `directory` doesn't change between calls but `tar_args` _does_, it seemed clearer to have the latter passed as a parameter.

> It would seem that this is a great opportunity to move the `tar` method to a utility function to avoid all these naming problems along with the inheritance issue.

I think this is where I'm supposed to say, "Perhaps a follow-on commit could correct that? Posted PRs are always welcome!"  😉  Seriously, though, unless there is a need from other code, I don't think moving it from here would be an improvement.

lib/pbench/agent/tool_data_sink.py
webbnh: @portante, I'm not sure what you're asking.

I think that "`ToolDataSinkParams` as a string" is the value in `params` here; passing that around creates the problem that I was trying to solve in this PR:  we want to validate it here, and the best way to validate it is to construct the `ToolDataSinkParams` object; if we defer that construction, then we have to come up with a separate way to validate it here (and constructing it twice seems silly for a couple of reasons).

Also, the instantiation of the `BenchmarkRunDir` object requires the value of `pbench_run`; if we do that here, then we don't have to pass that value around, either (and we find out up front if there's a problem with it).

So, I think this is the best location for instantiating both the `BenchmarkRunDir` and the `ToolDataSinkParams` objects.  The lingering question is, if the instantiation of the `BenchmarkRunDir` fails, how do we want to see that reported?...does it need to be a `ToolDataSinkError` or is some other error (such as `KeyError`) acceptable?

lib/pbench/agent/tool_meister.py
webbnh: The trailing underscore is not, strictly speaking _needed_, but it is _advised_ as `dir` is a Python reserved word.

I added it under the advice of PEP-8, but I suppose we could use `dirent` instead.  (That's less Pythonic, but it's consistent with other Python code in the project, which looks like Bash inheritance....)

portante: `dir` is a built-in, not a `keyword`, so the spirit is good, but trailing single underscores are a terrible idea in PEP-8, especially when other names could be used.

Now leading single-underscores is a different story ... ;)

webbnh:
> If one person's IDE has a linter that is not available to everybody else, should we be making changes based on that?

Let's pose the inverse of that question:  if one person's IDE has a linter that is not available to everybody else, should we be rejecting changes based on that?  That seems silly if, on the face of them, there is nothing wrong with the proposed changes.  So, pretend that I didn't mention the motivations for the changes, and just look and see if they make the code better.

> If our flake8 and black are not sufficient, then we should be able to add a linter in the same way so that all code passes the same test.

I don't think we really want to do that.  The reason why we call it "lint", instead of, say, "bugs", is that the items fall at various points along a continuum between "good" and "almost bad", and many of them aren't worth addressing.  (My PR fixes only a small number of the complaints, some of them (like the use of `PROG` as a variable or parameter (because it's solid-caps) I don't expect us to change...).

> It seems we really need to raise the bar higher for changes so that we reduce the risk of destabilizing the release.

This is a good point.  I suggest that we apply that bar uniformly to all PRs (including the ones from yesterday...).  But, also, I think it's reasonable to look at the content of the changes to assess the risk.  (I think all of the changes I made to appease the linter are "low risk".)

> We have been working on reducing the size of commits as well. It might be worth to break this up

As PRs go, this is not a large one by any means (and, much of the changes are in the gold files 😝).  Nevertheless, I'll split the PR.

> Add a linter run as part of the pre-commit check and tox along side flake8 and black, fixing everything the new linter finds

That would be a titanic PR.  🙃  I'm not about to even try that.

portante:
Certainly the view points on the use of linters is debatable.  And the process of how to structure PRs to facilitate change being made smoothly is helped by reduced technical debt, increased test coverage, and a reliable functional test suite (which we don't have).  And this PR is another step towards paying down our technical debt.

No one is rejecting this PR.

The thing is, we are proposing making this PR gate PR #2760.  That no longer makes sense to me given the nature and direction of this PR.

The end goal of PR #2760 is to get an implementation of the `pbench-tools-kill` command posted (delivered in PR #2743).  But these changes, while desirable and worth-while, are beyond the scope of end-game for the release.

I want to emphasize again, that I believe we should be making these changes.

I just don't believe we should be making this changes ahead of the PRs #2760, #2763, and #2743 and gate the v0.71 release.

Let's consider this PR independently and merge it after we get v0.71 out the door.

webbnh:
> I'm not thrilled about the `canonicalize` name

@dbutenhof, I'm not in love with it either, but it seemed reasonable.  And, I'm sorta hoping that we expand it beyond just `NamedTuple`'s into something fully generalized...but there's no call for that in this PR.

> I get Peter's comments about the hyper linter you're using; it wouldn't be a bad idea to at least identify that somewhere. (My vscode has started exposing dust bunnies that LGTM and flake8 don't see, but frankly I'm not even sure what it's using.) I like the idea of fixing them where they're obviously real issues regardless of whether anyone else would have seen them; at worst, it does no harm.

I have to laugh at the pejorative qualifiers that you and Peter apply to the linter!  It's just a linter...haven't you guys used one before?  (The `man` page for the original lint, under the "Bugs" section used to say, "There are some things that you just can't get `lint` to shut up about.")  It appears to be built into or encapsulated inside PyCharm.  I can't seem to find any obvious, separable identification, and it's not clear that there is only one linter involved.

Like you, I think we should address the items that it reports when they are "real issues" (whatever that actually means); and, if the resulting changes pass code review, I don't see the harm (that is, if the changes are obviously good enough to justify the code churn and the review cost, then I think we should accept them).

> He does have a point that if there are better lint filters it'd be nice to have the build using them... if we can figure out how to make that happen.

I certainly agree.  However, in the case of Black (and Flake8, in general), the assessments are non-negotiable.  If we had some sort of static code analyzer, any bugs it reported would be on a similar level.  But, with lint, a lot of it comes down to what level of orthodoxy we want to follow.  (E.g., PEP-8 is _supposed_ to be advisory...but if we require our code to be lint-free it will become mandatory, and I don't think we'll like that.)  So, at least for now, I'm inclined to propose changes based on the more reasonable of my IDE's comments, and do my best to shield us from the rest of them.  (E.g., do we really want to consider replacing all instances of `except Exception`??...'cause my IDE objects to every one of them as being "too general".... 😬)

webbnh:
> Let's consider this PR independently and merge it after we get v0.71 out the door.

No...let's split off the objectionable portions to a separate PR which we can evaluate on its own merits and leave the rest of them to support the present v0.71 effort.  And, I'm just about to do that.

portante:
> > He does have a point that if there are better lint filters it'd be nice to have the build using them... if we can figure out how to make that happen.
> 
> I certainly agree. However, in the case of Black (and Flake8, in general), the assessments are non-negotiable. If we had some sort of static code analyzer, any bugs it reported would be on a similar level. But, with lint, a lot of it comes down to what level of orthodoxy we want to follow. (E.g., PEP-8 is _supposed_ to be advisory...but if we require our code to be lint-free it will become mandatory, and I don't think we'll like that.) So, at least for now, I'm inclined to propose changes based on the more reasonable of my IDE's comments, and do my best to shield us from the rest of them. (E.g., do we really want to consider replacing all instances of `except Exception`??...'cause my IDE objects to every one of them as being "too general".... 😬)

Great, this is a great argument for addressing these kinds of changes after we release v0.71.  =)

lib/pbench/agent/tool_data_sink.py
webbnh: Let me get this PR split, first.  Then I can look at that.  🥴

webbnh:
> Great, this is a great argument for addressing these kinds of changes after we release v0.71. =)

Um, only if "these kinds of changes" refers to adding an additional linter to our CI tool-chain.  If "these kinds of changes" refers to improving the quality of our code, then not so much -- those should be looked at on an individual basis.

In any case, I've split the "linting" changes off into a separate branch, so they are no longer part of this PR.

lib/pbench/agent/tool_meister.py
webbnh: This is addressed in #2769.

lib/pbench/agent/tool_meister.py
webbnh: This is addressed in #2769.

dbutenhof (approved):

portante (changes requested):
Even with the PR posted to fix the promotion of `BenchmarkRunDir` object outside of the `ToolDataSink` class, it still feels like this work is best suited to land after `v0.71` where we can work out the changes over time and refine this.

I don't feel this should block PR #2760.

We have already delayed work by two days now towards landing PR #2743, and we have other work for `v0.71` that needs to land.

Let's hold our noses and git-r-done.

webbnh:
Rather than do this work here, let's do it in #2760, which is where it belongs.

PR #2768, opened 2022-04-21
Improved `linpack` benchmark wrapper

portante:
This PR supersedes PR #2752.

----

(Proposed "squashed" commit comment.)

Improved `linpack` benchmark wrapper

Instead of a complicated method of writing a script file using `echo`
statements, we create a driver script for linpack which takes care of
transforming a set of command line parameters into the form LINPACK
expects.

We also provide a new `linpack-wait` to hide the method of determining
when LINPACK completes.  The RPM packaging has been updated to deliver
these new files.

These changes remove the need for a per-iteration generated script,
which eliminates the need to copy extra files for each iteration and/or
sample.  Each execution is now per-sample, which allows us to drop the
logic of sharing setup between samples of a given iteration.

Since the `pbench-run-benchmark` script is deprecated, and this commit
completes the implementation of `pbench-linpack` to replace its
functionality, we take a number of steps to:

 * Remove `linpack` support from `pbench-run-benchmark`
 * Rename `linpack-postprocess` to `...-sample`
 * Rename `linpack-postprocess-cdm` to `linpack-postprocess`

We also:

 * Correct `pbench-linpack -C` short option processing

   The `-C` short option was not being handled at all.  We also fix up a
   bit of the formatting for option processing while we're at it.

 * Max header length 254 for `linpack.input` file

   The header and sub-header lines in a `linpack.input` file can only be
   at most 254 characters (plus a new-line character).  So the headers
   are truncated unconditionally.

   Since the idea was to place the final command line used in the header,
   we now write the final command to a file, `linpack.cmd`, in the output
   directory in case the header is truncated.

 * Remove use of `mkdir -p` where possible

 * Start using "last sample" parameter


agent/bench-scripts/driver/linpack
```diff
@@ -0,0 +1,277 @@
...
+# -*- mode: shell-script; indent-tabs-mode: t; sh-basic-offset: 8; sh-indentation: 8; tab-width: 8 -*-
+#
+# "linpack" driver script
+#
+# Responsible for transforming pbench-linpack command line inputs into the
+# the linpack invocation requirements.
+#
+# The linpack binary expects an input file with 7 lines in it:
+#
+#     Line 1: Header (disgarded)
```
ndokos, line 11: "disregarded"?  "discarded"?

agent/bench-scripts/driver/linpack
```diff
@@ -0,0 +1,277 @@
...
+use_omp="y"
+threads_arg=""
+kmp_affinity_args="nowarnings,compact,1,0,granularity=fine"
+numactl_args=""
+
+# Output directory required.
+output_dir=""
+
+function usage() {
+	printf -- "TBD -- how to use the pbench linpack driver ...\n"
```
ndokos, line 77: Are you planning to change this before merging?

ndokos, line 77: Are you planning to change this before merging?

agent/bench-scripts/driver/linpack
```diff
@@ -0,0 +1,277 @@
...
+#
+# For lines 4-7, if there are N tests to run, each line should have N values
+# describing the parameters for those tests. For example:
+#
+#     Line 1: Header
+#     Line 2: Sub-Header
+#     Line 3: 3
+#     line 4: 10000 20000 30000
+#     line 5: 10016 20016 30016
+#     line 6:     2     2     2
```
ndokos, line 27: Are the leading dimensions always equal to `number-of-equations-to-solve + 16`? If so, can we get rid of them?

ndokos, line 27: Are the leading dimensions always equal to `number-of-equations-to-solve + 16`? If so, can we get rid of them?

agent/bench-scripts/postprocess/linpack-postprocess-sample
```diff
@@ -0,0 +1,163 @@
...
+        printf STDERR "ERROR: Could not open client %s results directory!\n", $client_hostname_dir;
+        exit 2;
+    }
+
+    foreach my $client_data_file (@client_data_files) {
+        if ($client_data_file =~ /^linpack\.(out|meta)$/) {
+            if (open(CLIENT_OUTPUT, "$dir/clients/$client_hostname_dir/$client_data_file")) {
+                while (<CLIENT_OUTPUT>) {
+                    if ($client_data_file =~ /meta$/) {
+                        if ($_ =~ /^(binary|use_omp|threads|kmp_affinity|numactl_cmd|problem_sizes|leading_dimensions|alignment_values)/) {
```
ndokos, line 59: This is basically a rename of `linpack-postprocess` to `linpack-postprocess-sample` with l.59 being the only difference, correct?

ndokos (approved):
Minor typos, a few questions, but  LGTM.

agent/bench-scripts/postprocess/linpack-postprocess-sample
ndokos: I went through the individual commits and figured out that this is indeed the case.

agent/bench-scripts/driver/linpack
```diff
@@ -0,0 +1,277 @@
...
+printf -- "binary=%s\nuse_omp=%s\nthreads=%s\nkmp_affinity=%s\nnumactl_cmd=%s\nproblem_sizes=%s\nleading_dimensions=%s\nalignment_values=%s\n" \
+	"${binary}" \
+	"${use_omp}" \
+	"${threads_arg}" \
+	"${kmp_affinity_args}" \
+	"${numactl_cmd}" \
+	"${problem_sizes}" \
+	"${leading_dimensions}" \
+	"${alignment_values}" \
+	> ${linpack_metadata_file}
```
webbnh, line 275: I don't really like the long format string (it wraps...) nor the separation of the arguments from the format -- it makes the code hard to review.  I think it would be better if we let the shell expand the variables in-line, e.g.:
```
cat > ${linpack_metadata_file} <<-__EOF__
	binary=${binary}
	use_omp=${use_omp}
	threads=${threads_arg}
	kmp_affinity=${kmp_affinity}
	numactl_cmd=${numactl_cmd}
	problem_sizes=${problem_sizes}
	leading_dimensions=${leading_dimensions}
	alignment_values=${alignment_values}
	__EOF__
```
Something similar would be good at lines 249-257, although the need there isn't quite so great, because the lines are mostly just ordered, they aren't trying to match keys to values.

webbnh (commented):
Here's the first installment of comments (more later today, perhaps).  No dealbreakers.  One or two small issues, some nits, and a bunch of code suggestions.

agent/bench-scripts/driver/linpack
portante: This is not up to us, this is dictated by the linpack program.

agent/bench-scripts/driver/linpack
portante: I wasn't because it is not part of the pbench-agent command line interface, and was taking the easy way out since this is just a refactor of the previous incarnation of this script, `linpack-prepare-input-file`.

I'll take a stab at a usage message.

agent/bench-scripts/driver/linpack
```diff
@@ -0,0 +1,277 @@
...
+
+# Output directory required.
+output_dir=""
+
+function usage() {
+	printf -- "TBD -- how to use the pbench linpack driver ...\n"
+}
+
+# Process options and arguments
+opts=$(getopt -q -o h --longoptions "binary:,header:,subheader:,problem-sizes:,leading-dimensions:,alignment-values:,run-trials:,threads:,use-omp:,kmp-affinity:,numactl-args:,output-dir:,help" -n "getopt.sh" -- "${@}")
```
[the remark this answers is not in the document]
portante: Ah ... no.  Thanks.

agent/bench-scripts/driver/linpack
```diff
@@ -0,0 +1,277 @@
...
+exit_code=${?}
+
+# Create the linpack.meta file containing all the metadata used for invoking
+# LINPACK (does not include "header" or "subheader").  This serves as the marker
+# file for its completion.
+printf -- "binary=%s\nuse_omp=%s\nthreads=%s\nkmp_affinity=%s\nnumactl_cmd=%s\nproblem_sizes=%s\nleading_dimensions=%s\nalignment_values=%s\n" \
+	"${binary}" \
+	"${use_omp}" \
+	"${threads_arg}" \
+	"${kmp_affinity_args}" \
```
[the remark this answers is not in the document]
portante: Why do you believe that?  Doesn't the existing code use the value of what `KMP_AFFINITY` will be set to?

Prior to this commit:

 * The [value](https://github.com/distributed-system-analysis/pbench/blob/main/agent/bench-scripts/postprocess/linpack-prepare-input-file#L218) set is the contents of the `KMP_AFFINITY` environment variable
 * It is consumed [here](h]ttps://github.com/distributed-system-analysis/pbench/blob/main/agent/bench-scripts/postprocess/linpack-postprocess#L59), where the stored key is `kmp_affinity` and the value is anything after the `=`

agent/bench-scripts/driver/linpack
```diff
@@ -0,0 +1,277 @@
...
+	# N.B. - Trailing whitespace is required.
+	kmp_affinity="KMP_AFFINITY=${kmp_affinity_args} "
+else
+	# When "use OpenMP" is false, we are not going to use the "threads_arg" or "kmp_affinity_args"
+	threads_arg=""
+	kmp_affinity_args=""
+
+	# No environment variables to add to the command line.
+	omp_num_threads=""
+	kmp_affinity=""
```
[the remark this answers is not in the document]
portante: I kept them here next to their use in `final_cmd` since that is their only purpose, they are not part of the CLI interface.

I don't think the reduction is code is helpful here because we want to call the attention of the reader to the semantic that when `use_omp` is not `y`, we don't use either of those values.

agent/bench-scripts/pbench-linpack
```diff
@@ -266,7 +282,8 @@ for thread in ${threads//,/ }; do
 
 		# Post-process the benchmark results after the tools so that the
 		# post-processing has that data available to it.
-		postprocess-iteration "${sample_dir}" "${benchmark_run_dir}"
+		[[ ${sample} -ne ${nr_samples} ]]; last=${?}
```
webbnh, line 285: Come to find out, Bash has a ternary operator!  So this can be either
```
let last=(sample == nr_samples ? 1 : 0)
```
or
```
(( last = (sample == nr_samples ? 1 : 0) ))
```
🙂

agent/bench-scripts/postprocess/linpack-postprocess-sample
```diff
@@ -0,0 +1,163 @@
...
+                            chomp($_);
+                            my @fields = split(/=/, $_, 2);
+                            $client_data{$client_hostname_dir}{$fields[0]} = $fields[1];
+                        }
```
webbnh, line 63: Are you sure we want to ignore any keys which aren't in the list on line 59?  Is there a downside or reason not to capture them all?

webbnh (commented):
First pass done; no dealbreakers.  I have a couple more coding suggestions (one which I think you'll really like 😁), and a pointed question.

Stand by for round two.

agent/bench-scripts/postprocess/linpack-postprocess-sample
portante: This is existing code which I did not want to change.

agent/bench-scripts/driver/linpack
```diff
@@ -27,8 +27,8 @@
 #     line 6:     2     2     2
 #     line 7:     4     4     4
 #
-# This script provides optional input parameters for lines 1, 2, 4 - 7, and
-# then calculates the number of tests to be run from those inputs to fill in
+# This script offers input parameters for lines 1, 2, 4 - 7, with default values
+# and then calculates the number of tests to be run from those inputs to fill in
 # line 3 of the LINPACK input file.
```
webbnh: Perhaps something like, "This script provides optional command line switches to override the default values for lines 1, 2, 4-7; the script calculates...".

agent/bench-scripts/driver/linpack
```diff
@@ -234,27 +235,27 @@ else
...
 
-if [[ -z "${header}" ]]; then
-	# Line 1 default
-	header="Pbench - LINPACK Benchmark data file (${linpack_input_file})"
-fi
-if [[ -z "${subheader}" ]]; then
-	# Line 2 default header
-	subheader="Pbench LINPACK Benchmark data -- ${final_cmd}"
-fi
+# Line 1, prividing a default header
```
webbnh, line 242: s/prividing/providing/

webbnh (approved):
Looking good, modulo a typo and a bit of wordsmithing.

Once we resolve the `kmp_affinity_args` discussion, it'll be GTG.

agent/bench-scripts/driver/linpack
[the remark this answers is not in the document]
webbnh: OK, I concur:  in the metadata we are capturing what will [have] be[en] the contents/value of the `KMP_AFFINITY` environment variable and associating them with the `kmp_affinity` key, so, yes, this should be `kmp_affinity_args`, here...so we have to manage two parallel variables...bleh.

agent/bench-scripts/driver/linpack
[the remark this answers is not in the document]
webbnh: OK, given my [recent revelation](https://github.com/distributed-system-analysis/pbench/pull/2768#discussion_r856733848), I need to amend my statement above:  in the current code structure, we _would_ need to clear `kmp_affinity_args` here.  😝  So, let me start over....

There are two obvious (to me...) approaches to structuring code like this.

The first declares a set of variables and initializes them using a coherent set of defaults; it then collects the command line arguments, overriding the variables' values as specified; finally, it uses the variables for the job at hand.

The second declares a set of more-or-less constant values indicating what the default values are; it then collects the command line arguments in a different set of variables; it then uses the variables for the job at hand, filling in the defaults for any variables which aren't set by the command line.

Unfortunately, the code here is doing something sort of in between.

That is, it's mostly following the first approach above, and we're sort of set up for `use_omp == "y"` by default.  If that setup were complete, then we wouldn't need to change any options unless we find that `use_omp == "n"`.  However, we didn't initialize `threads_arg` properly, and we defer setting `omp_num_threads` and `kmp_affinity` until we're done with the command line processing.  (The latter two are more or less necessary, but I think it would be better if we put the `/proc/cpuinfo` count up at line 72 so that we don't need to touch it at line 223 and we can just use it as is to set up `omp_num_threads`.)

The weird thing is, in the `use_omp == "n"` case, we're still putting the `threads` and `kmp_affinity` keys in the metadata, even though their values are _blank_.  It seems like we should be omitting them, in which case we wouldn't have to worry about clearing their values in the code referenced here.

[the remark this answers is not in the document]
portante: > That is, it's mostly following the first approach above, and we're sort of set up for `use_omp == "y"` by default. If that setup were complete, then we wouldn't need to change any options unless we find that `use_omp == "n"`. However, we didn't initialize `threads_arg` properly, and we defer setting `omp_num_threads` and `kmp_affinity` until we're done with the command line processing. (The latter two are more or less necessary, but I think it would be better if we put the `/proc/cpuinfo` count up at line 72 so that we don't need to touch it at line 223 and we can just use it as is to set up `omp_num_threads`.)

The previous behavior of `linpack-prepare-input-files` only fetched the CPU count from `/proc/cpuinfo` when `${use_omp} == "y"` and no explicit `threads` value was provided.  So this code maintains that behavior.

I added use of `_def` values to make it clearer when we end up using defaults.

> 
> The weird thing is, in the `use_omp == "n"` case, we're still putting the `threads` and `kmp_affinity` keys in the metadata, even though their values are _blank_. It seems like we should be omitting them, in which case we wouldn't have to worry about clearing their values in the code referenced here.

The previous behavior of `linpack-prepare-input-files` did that, so I kept [that behavior](https://github.com/distributed-system-analysis/pbench/blob/main/agent/bench-scripts/postprocess/linpack-prepare-input-file#L220).

agent/bench-scripts/driver/linpack
```diff
@@ -184,50 +219,62 @@ while true; do
...
 av_list=(${alignment_values})
 if [[ ${#ps_list[*]} -ne ${#ls_list[*]} || \
       ${#ls_list[*]} -ne ${#rt_list[*]} || \
       ${#rt_list[*]} -ne ${#av_list[*]} ]]; then
-    printf -- "[%s] ERROR: You must have the same number of elements for --problem-sizes, --leading-dimensions, --run-trials, and --alignment-values\n" >&2
-    exit 2
+	printf -- "[%s] ERROR: You must have the same number of elements for --problem-sizes, --leading-dimensions, --run-trials, and --alignment-values\n" >&2
+	exit 2
 fi
 number_of_tests=${#ps_list[*]}
```
webbnh, line 257: This code might be clearer if we reversed the order, here:
```
number_of_tests=${#rt_list[*]}
if [[ ${#ps_list[*]} -ne ${number_of_tests} || \
	${#ls_list[*]} -ne ${number_of_tests} || \
	${#av_list[*]} -ne ${number_of_tests} ]]
then
	printf -- "[%s] ERROR: You must have the same number of elements for --problem-sizes, --leading-dimensions, --run-trials, and --alignment-values\n" >&2
	exit 2
fi
```

webbnh (dismissed):
Looks good.

I have a few thoughts, comments, and questions though.  The biggest ones center around the help text (I don't think you want the de-indenting "here document" operator, and I'm not sure we want to include _all_ the help text in the the usage message...).

Other that those, there are just nits and suggestions which you can consider or ignore.

agent/bench-scripts/driver/linpack
```diff
@@ -1,87 +1,128 @@
...
+kmp_affinity_args_def="nowarnings,compact,1,0,granularity=fine"
+kmp_affinity_args=""
 numactl_args=""
 
 # Output directory required.
 output_dir=""
 
 function usage() {
-	printf -- "TBD -- how to use the pbench linpack driver ...\n"
+	cat <<-__EOF__
```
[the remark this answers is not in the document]
portante: It does not appear to be stripping the leading whitespace.

agent/bench-scripts/driver/linpack
```diff
@@ -184,50 +219,62 @@ while true; do
...
 ps_list=(${problem_sizes})
+leading_dimensions=${leading_dimensions:-${leading_dimensions_def}}
 ls_list=(${leading_dimensions})
+run_trials=${run_trials:-${run_trials_def}}
 rt_list=(${run_trials})
+alignment_values=${alignment_values:-${alignment_values_def}}
 av_list=(${alignment_values})
 if [[ ${#ps_list[*]} -ne ${#ls_list[*]} || \
       ${#ls_list[*]} -ne ${#rt_list[*]} || \
       ${#rt_list[*]} -ne ${#av_list[*]} ]]; then
```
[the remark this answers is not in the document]
portante: Not if we want them to still line up.

agent/bench-scripts/driver/linpack
```diff
@@ -1,87 +1,128 @@
...
 output_dir=""
 
 function usage() {
-	printf -- "TBD -- how to use the pbench linpack driver ...\n"
+	cat <<-__EOF__
+Pbench ${script_name} linpack driver
+
+Usage: ${script_name} [--problem-sizes=#[[,#]...]] [--leading-dimenions=#[[,#]...]]
+        [--run-trials=#[[,#]]...] [--alignment-values=#[[,#]...]] [--use-omp={[y]|n}]
+        [--threads=#] [--kmp-affinity=<options,>] [--numactl-args=<args,>]
```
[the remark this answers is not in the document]
portante: I have removed the `,` from both because this script cannot dictate the format of those arguments, it should only be acting as a pass-thru.

agent/bench-scripts/driver/linpack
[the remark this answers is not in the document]
webbnh: Huh.  OK.  Nevermind then.  Hrmm....

webbnh (approved):
Looks OK, unless I'm right about that exit status thing....

agent/bench-scripts/driver/linpack
```diff
@@ -248,13 +258,13 @@ run_trials=${run_trials:-${run_trials_def}}
 rt_list=(${run_trials})
 alignment_values=${alignment_values:-${alignment_values_def}}
 av_list=(${alignment_values})
-if [[ ${#ps_list[*]} -ne ${#ls_list[*]} || \
-      ${#ls_list[*]} -ne ${#rt_list[*]} || \
-      ${#rt_list[*]} -ne ${#av_list[*]} ]]; then
+number_of_tests=${#ps_list[*]}
+if [[ ${#ls_list[*]} -ne ${number_of_tests} || \
+      ${#rt_list[*]} -ne ${number_of_tests} || \
+      ${#av_list[*]} -ne ${number_of_tests} ]]; then
```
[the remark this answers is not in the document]
webbnh: Ah, so they are indented this way on purpose, to align all the dollar signs and things?  Well, OK....  😁

webbnh (approved):
Looks good.  Just down to nits, now.

agent/bench-scripts/driver/linpack
```diff
@@ -0,0 +1,333 @@
...
+
+# Line 1, providing a default header (truncated to 254 characters below)
+header=${header:-"Pbench - LINPACK Benchmark data file (${linpack_input_file})"}
+# Line 2, providing a default sub-header (truncated to 254 characters below)
+subheader=${subheader:-"Pbench - LINPACK Command: '${display_cmd}'"}
+
+# Create the LINPACK input file.  Note the comment lines are redundant but
+# provided in a spirit of consistency with the example input files provided by
+# linpack RPM.
+cat > ${linpack_input_file} <<-__EOF__
```
dbutenhof, line 305: Just as here, the lines *are* tab-indented so the `<<-` dedents the text... cool.

webbnh (approved):
👍

dbutenhof (approved):
Nothing substantial...

ndokos (dismissed):

dbutenhof (approved):

agent/bench-scripts/driver/linpack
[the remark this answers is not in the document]
webbnh: Yeah, I rechecked the docs earlier and noted that it was specifically tabs.  But, that effect is applied to the whole "document", not just the closing delimiter...but, as you say, there are no tabs in the document itself in this case, so it doesn't matter.

webbnh (approved):
👍

ndokos (approved):

PR #2780, opened 2022-04-26
Deprecate unused benchmark convenience scripts

portante:


dbutenhof (approved):
Very specific, "next release" rather than "future release"; I like that. (Though we'll still have to work out "0.71.1" vs "0.72" vs "1.0".)

npalaska (approved):

PR #2785, opened 2022-04-26
Merge the `doc` directory into `docs`

portante:
This is a whole-sale merge of the `doc` directory into the `docs` directory, with the two `Makefile`s combined.


npalaska (dismissed):
Interesting, btw why did we create 2 doc(s) directory to begin with?

dbutenhof (dismissed):

docs/Makefile
```diff
@@ -1,4 +1,42 @@
-# Minimal makefile for Sphinx documentation
+all:  pbench-agent.html test-rpms.html
+
+everything: all pbench-agent.pdf help
+
+%.html:	%.org
```
webbnh, line 5: Shouldn't this target also depend on `batch-export.el`?  (Ditto line 11.)

docs/Makefile
```diff
@@ -1,4 +1,42 @@
...
+push-results:	results.html
+	rsync -ave ssh $< ${dest}/$<
+
+push-test-rpms:	test-rpms.html
+	rsync -ave ssh $< ${dest}/test-repo/$<
+
+push-relnotes: RELEASE-NOTES.html
+	rsync -ave ssh $< ${dest}/doc/$<
+
+.PHONY: clean veryclean
```
webbnh, line 31: I'm pretty sure that `everything` and each of the `push-*` targets are phony, too.

docs/Makefile
```diff
@@ -1,4 +1,42 @@
...
+clean:
+	rm -f *~ *.fdb_latexmk *.fls *.toc *.aux *.log *.out
+
+veryclean:	clean
+	rm -f *.html *.pdf
```
webbnh, line 36: I don't see `*.tex` in either of these lists...should it be there?

docs/Makefile
```diff
@@ -7,13 +45,6 @@ SPHINXBUILD   = sphinx-build
 SOURCEDIR     = .
 BUILDDIR      = _build
 
-# Put it first so that "make" without argument is like "make help".
+.PHONY: help Makefile
```
webbnh, line 48: I realize that you didn't change this, but I'll just observe that it seems weird to declare `Makefile` as a phony target.  For one thing, I don't see a definition for a `Makefile` target, but, more importantly, if we _were_ to produce a makefile, then we probably don't want to reproduce it unless it is stale, which is part-and-parcel to Make's handling of (non-phony) targets.

webbnh (dismissed):
Looks fine.

Just a few items for your consideration.

docs/Makefile
portante: I'll add it, but I'll also defer to @ndokos for the final word on this.

portante: And `*.xdv` should be there as well.

docs/Makefile
```diff
@@ -1,4 +1,42 @@
-# Minimal makefile for Sphinx documentation
+all:  pbench-agent.html test-rpms.html
+
+everything: all pbench-agent.pdf help
+
+%.html:	%.org
+	emacs -batch -l batch-export.el --eval "(batch-org-export-as 'html)" $<
+
+%.pdf:	%.tex
+	latexmk --shell-escape -pdf -xelatex $<
```
portante, line 9: @ndokos, this command fails for me with missing `*.sty` files that `latexmk` prompts me for during the build, interactively.

portante (commented):
@ndokos, gonna need you to me verify these changes.

docs/Makefile
ndokos: I'll check.

docs/Makefile
ndokos: `*.tex` should probably be in the `veryclean` action. For debugging purposes, it is useful to do a `clean` and then process the `tex` file again to see where it breaks (e.g. the missing `.sty` files).

OTOH, we've never made PDFs available, only HTML, so one solution is to get rid of the `tex/pdf` generation path altogether.

docs/Makefile
ndokos: I presume that this came with the Sphinx stuff, but I have no recollection of doing that. Maybe some git archaeology is warranted to find out when/who/how (if it comes back to me, I'll take the Reagan defence).

ndokos (commented):
Some comments/replies for now, until I try it out.

docs/Makefile
ndokos: The Org mode LaTeX exporter adds a bunch of packages to the export which need to be installed on whatever box you are doing the LaTeX processing.  The question is whether we want to worry about producing PDF.

ndokos (approved):
The changes look OK to me.

docs/Makefile
```diff
@@ -45,6 +47,16 @@ SPHINXBUILD   = sphinx-build
 SOURCEDIR     = .
 BUILDDIR      = _build
 
-.PHONY: help Makefile
-help:
-	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+# Catch-all target: route all unknown targets to Sphinx using the new
+# "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
+singlehtml:
```
webbnh, line 52: I suspect that `singlehtml` is a phony target.

docs/Makefile
```diff
@@ -45,6 +47,16 @@ SPHINXBUILD   = sphinx-build
...
+	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
```
webbnh, line 53: Where does `$(O)` get defined?  (And, btw, that's an "oh", not a zero...not that it matters, because we're not in a context where `$(0)` would be defined, either.)

webbnh (dismissed):
Looks good...just two more nits for your consideration.

docs/Makefile
webbnh: Not to tattle or anything, but `*.tex` is currently in the `clean` target, not the `veryclean` target....

portante (commented):
All comments have been addressed!

docs/Makefile
```diff
@@ -7,13 +47,17 @@ SPHINXBUILD   = sphinx-build
...
-help:
-	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+# Build the "singlehtml" Sphinx target.  There are other supported targets
+# one could use; see `$(SPHINXBUILD) -M help` for a complete list.
+.PHONY: singlehtml
+singlehtml:
+	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS)
 
-.PHONY: help Makefile
+.PHONY: clean veryclean
```
dbutenhof, line 56: Do we need `veryclean` here when it's also phony-ed below where the target is declared?

docs/Makefile
dbutenhof: So ... does this mean we should just remove the `%.pdf` target? Write some instructions about installing the packages to actually generate it?

portante: Yes, but I'd rather do that once we decide how we'll be creating documentation in the future.

docs/Makefile
```diff
@@ -7,13 +47,18 @@ SPHINXBUILD   = sphinx-build
...
-.PHONY: help Makefile
+.PHONY: clean
+clean:
+	rm -f *~ *.fdb_latexmk *.fls *.toc *.aux *.log *.out *.xdv
+	rm -rf $(BUILDDIR)
 
-# Catch-all target: route all unknown targets to Sphinx using the new
-# "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
-%: Makefile
-	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
```
webbnh, line 19: I'm feeling kind of sad that we're losing the ability to pass unknown targets on to `$SPHINXBUILD`.

If you change the `html:` target to `%:`, does that really cause a problem?  (According to the docs, it shouldn't interfere with the `%.html`, `%.tex`, and `%.pdf` targets because they come before it in the makefile, and it shouldn't be selected over the `push-*` targets because they are more specific.)

(I can imagine how it could cause a problem for the `clean` or `veryclean` targets (although it _shouldn't_), but that would be easy to address by placing their definitions before the catch-all one.)

webbnh (approved):
Looks good to me...but I wish we didn't have to ditch the catch-all target.

docs/Makefile
portante: Yes, you get whacky output try to build the other existing targets.  We'll resolve this later once we decide what documentation build system we should settle on for the future.

I suspect it will be `Sphinx`, but I did not want to do all sorts of whole-sale changes in this PR.  This PR is trying to just bring directories together.

PR #2787, opened 2022-04-27
Fix CI build image

portante:
Not sure how this was working up to now, except that the existing image was not built from this latest sources.


portante (commented):
Verified after pushing to Quay.

webbnh (approved):

ndokos (approved):

PR #2789, opened 2022-04-27
Agent v0.72, Server v1.0

portante:
Movin' on!  Next Agent targeted as v0.72, while we are going to prepare for v1.0 for the Pbench Server.


portante (commented):
Certainly this is debatable.

I am personally not in favor of tying the agent and server code bases in lock-step releases, since the code changes are going to move forward at different rates, and with different goals and constraints (agent code has more legacy support required, while server code will only be deployed via controlled container versions).

npalaska (approved):

dbutenhof:
> Certainly this is debatable.
> 
> I am personally not in favor of tying the agent and server code bases in lock-step releases, since the code changes are going to move forward at different rates, and with different goals and constraints (agent code has more legacy support required, while server code will only be deployed via controlled container versions).

I think there are (at least) two "reasonable" ways to look at this:

1. We have a "Pbench" deliverable, with a version number. Some releases might be more packed with Agent changes than with Server changes, or Dashboard changes, or any combination, but it's a single "product".
2. We have a "Pbench Agent" deliverable and a "Pbench Server" deliverable, tied loosely only by the interface specifications (the `PUT` and the tarball format); they have different versions and potentially different schedules, and we only need to coordinate changes in those interface constraints.

While I don't see any inherent problem with the second alternative, I also don't particularly want to go that direction at this point, so I would argue for a single release plan covering both.

ndokos:
Stirring the pot a bit: if we are going to have separate release streams for the agent and the server, shouldn't we also have different repos for them?  And possibly more: e.g. utility libraries that might be useful to both should go in their own repo. That should also simplify the directory structure.

That would allow existing release mechanisms provided by github to deal with the independent streams easily. 

As it stands, the next release of either is going to be an RPM or a container image or something, with a version number attached, so there is no confusion for users, but we will have to explicitly name the release with either an `agent` or a `server` tag (something which we have done reluctantly in the past IIRC).

I guess the question is do the different numbers mean anything other than that they are independent streams, and if they are independent streams (which I believe they are), should we split them apart into multiple repos? After all, isn't that one of the reasons that the dashboard is a separate repo?

webbnh:
My instinct follows Nick's -- if the Agent and the Server (and the Dashboard) are to be only loosely coupled, why are they being delivered out of the same repo?  However, I have to admit, to the extent that we work on all three as a team, I _like_ having them all in the same repo.  🙂 

Nevertheless, let me stir the pot some more:  I think the only thing that we ever (for at least the medium term) **_release_** is the Agent, and, therefore, that's the only thing for which version numbers are really significant.  I think the Server will be _deployed_ -- that is, there will only be one of them which users will use at a time, and its version will be entirely under our control.  And, the Dashboard will be _delivered_...again, the user won't be choosing what version of the Dashboard they run -- they'll get whatever the Server sends them.

Yes, I know, we'll need ways of referring to the version of the Server and the Dashboard in things like bug reports, but that could be a Git hash, or a date, or any of a number of things other than a classic version number.

So, I guess I'm of the opinion that we don't need to keep the Server version in sync with the Agent version (as much as that would be my preference), because I don't think it's really going to matter.

webbnh (commented):
Movin' on.

webbnh (approved):
Eh-hem.  _**Approving**_...and movin' on.

dbutenhof:
> Stirring the pot a bit: if we are going to have separate release streams for the agent and the server, shouldn't we also have different repos for them?

No, no, no, please, no! If keeping them in the same repo is an argument for keeping a single release stream, I'm all in. Separating them would be a massive pain, and has absolutely no value. We brought the dashboard back *into* the common repo for the benefits of keeping the code synchronized!

pravins:
This is a good move. agent will be versioned with v0.7* and Server with v1.0.0. 
AFAI think at the moment, we do not have any urgent need to do any separation. We can definitely think for that in future, when will have more users and significant maintenance activities at both side agent and server.

PR #2790, opened 2022-04-27
Allow a list of ?metadata tags for simplicity

dbutenhof:
PBENCH-637

This is a minor improvement to "ease of use" allowing multiple metadata keys on a single `metadata` query parameter. It essentially allows a `ParamType.LIST` to be a comma-embedded string rather than a JSON list. It also refactors the two places that processed query parameters into an `ApiBase` helper.

(I'm also experimenting with adding the Jira tag to the GitHub PR description but *not* the git log message.)


portante (dismissed):

dbutenhof:
> (I'm also experimenting with adding the Jira tag to the GitHub PR description but _not_ the git log message.)

It's been three hours and this PR hasn't been linked to the Jira ticket. I'm going to treat that essentially as proof that referencing a Jira ticket in the GitHub PR description **is not** sufficient; we need the reference to be in the git log message.

dbutenhof:
I've rebased, but the rebased commit is more in the nature of another Jira experiment. I added the Jira reference to the git log message to see whether Jira will pick it up after the fact. (I suspect not, from previous experience, though this is slightly different from what I've tried before.)

portante (dismissed):

lib/pbench/server/api/resources/__init__.py
```diff
@@ -337,27 +351,36 @@ def convert_keyword(value: str, parameter: "Parameter") -> str:
     raise KeywordError(parameter, "keyword", [value])
 
 
-def convert_list(value: List[Any], parameter: "Parameter") -> List[Any]:
+def convert_list(value: JSONVALUE, parameter: "Parameter") -> List[Any]:
```
webbnh, line 354: It seems like the type hint for `value` should be `Union[JSONARRAY, JSONSTRING]` or `Union[List[Any], str]` -- this function won't accept any other type of `JSONVALUE`.

lib/pbench/server/api/resources/__init__.py
```diff
@@ -412,7 +435,7 @@ class ParamType(Enum):
     STRING = ("String", convert_string)
     USER = ("User", convert_username)
 
-    def __init__(self, name: AnyStr, convert: Callable[[AnyStr], Any]):
+    def __init__(self, name: AnyStr, convert: Callable[[Any, "Parameter"], Any]):
```
webbnh, line 438: As long as we're changing this line, do you really like `AnyStr`, or should we go with `str`?

lib/pbench/server/api/resources/__init__.py
```diff
@@ -618,6 +641,40 @@ def __init__(
...
+        """
+        json = {}
+        badkey = []
+        for key in request.args.keys():
+            if key in schema:
+                values = request.args.getlist(key)
+                if len(values) > 1 and schema[key].type == ParamType.LIST:
+                    json[key] = values
+                else:
+                    json[key] = values[0]
```
webbnh, line 666: Why use the first one?  (Why not the last one, or why not reject the key, if it's a list when it's not supposed to be?)

lib/pbench/test/unit/server/test_datasets_list.py
```diff
@@ -167,7 +167,7 @@ def test_get_bad_keys(self, query_as):
             query_as: Query helper fixture
         """
         response = query_as(
-            {"metadata": ["xyzzy", "plugh", "dataset.owner", "dataset.access"]},
+            {"metadata": "xyzzy,plugh,dataset.owner,dataset.access"},
```
webbnh, line 170: You've removed the explicit `list` and replaced it with the stringized list...is there any testing of an explicit `list`, now?

lib/pbench/test/unit/server/test_datasets_metadata.py
```diff
@@ -76,6 +67,18 @@ def test_get2(self, client, server_config, provide_metadata):
...
+        response = client.get(
+            f"{server_config.rest_uri}/datasets/metadata",
+            query_string={
+                "name": "drb",
+                "controller": "foobar",
+                "metadata": "dashboard.seen,server.deletion,dataset.access,user",
+            },
+        )
+        assert response.status_code == HTTPStatus.BAD_REQUEST
+        assert response.json == {"message": "Unknown URL query keys: controller"}
```
webbnh, line 80: Is there a test which checks multiple unknown keys?

webbnh, line 80: When/where did `"controller"` get removed from the schema?

webbnh (approved):
These are nice changes!  (Although, I'm not sure that shrinking `dateutil.parser` to `date_parser` was worth the effort, but....)

I've got a couple of nits and a couple of questions...foremost among them being, did you create a testing deficit?

lib/pbench/server/api/resources/__init__.py
dbutenhof: Very possibly. ☹️

lib/pbench/server/api/resources/__init__.py
dbutenhof: Most of the changes outside the code I refactored were driven by continuing annoyance at the red from `vscode`'s mysterious linter. I wasn't trying to go out of my way to pick lint, but I did (arguably inappropriately) give in a make a few easy fixes. But it didn't complain about the `name: AnyStr`, and I didn't pay any attention to it. We probably should expunge all use of `AnyStr` because we rarely use byte strings and certainly not in any of this code.

lib/pbench/server/api/resources/__init__.py
dbutenhof: The code before used `getlist` for a `LIST`-type parameter and `get` for a non-list parameter. The actual behavior of the Werkzeug `Multidict.get` is to return the first element of the list and I just translated the idiom.

Basically it means that `?name=foo&name=bar` will ignore the extra; and, you're right, maybe we shouldn't do that, but it wasn't a nuance I considered at the time.

Hmm.

lib/pbench/test/unit/server/test_datasets_list.py
dbutenhof: The `test_dataset_list` uses a list, albeit with just one element, which doesn't really matter. The point was to make sure at least one test used the string, and I preferred the case that used multiple elements. Nor did it seem necessary to add another test case. The `test_schema.py` cases cover all the variations pretty thoroughly.

lib/pbench/test/unit/server/test_datasets_metadata.py
dbutenhof: Hmm; no, there isn't. This was just a result of discovering that the tests still used the obsolete `controller` key, and I figured that was worth testing.

dbutenhof: To (mis)quote George Lucas, "Moderately long ago, in a galaxy not directly adjacent".

It went bye-bye when I re-wrote `datasets/list` to use SQL instead of Elastic.

lib/pbench/server/api/resources/__init__.py
webbnh: I'm right there with you on the IDE linter thing -- the sidebar is so full of "look at me!" marks that I can't always find the marks that I really care about...so, I think picking lint is a good thing...but we need to take some care that it doesn't distract from the change at hand (i.e., it would be good to segregate lint-picking to its own commit).

As for purgining `AnyStr`'s, I agree completely, but I'm inclined to attack the problem piecemeal, and this one looked like it could be changed with little impact.  😉

lib/pbench/test/unit/server/test_datasets_metadata.py
webbnh: Should I be concerned about the fact that the tests didn't object...or...are these not the droids I'm looking for?  (I can go about my business...I'll just move along, now....)

dbutenhof: No, those droids went "thataway" ...

I was previously ignoring unrecognized query parameters. When I was writing the new common handler it occurred to me that it would be easy to misspell one, and it'd be silently ignored, so I added the error checking as in our `KEYWORD` schema type.

And of course immediately tests failed because of these obsolete (and previously ignored) `controller` params.

npalaska (approved):
Indeed a good improvement :+1:

lib/pbench/server/api/resources/__init__.py
```diff
@@ -107,14 +107,42 @@ class MissingParameters(SchemaError):
...
+        super().__init__()
+        self.keys = sorted(keys)
+
+    def __str__(self):
+        return f"Unknown URL query keys: {','.join(self.keys)}"
+
+
+class RepeatedQueryParam(SchemaError):
+    """
+    A unique URI query parameter key for which Pbench doesn't support multiple
```
portante, line 134: Perhaps two commas might be helpful?  `... key, for ... values, was ...`.

portante (approved):

portante (approved):

lib/pbench/server/api/resources/__init__.py
```diff
@@ -648,10 +660,20 @@ def _validate_query_params(self, request: Request, schema: Schema) -> JSONOBJECT
         parameters into a JSON object and validates them against a specified
         schema.
 
+        Note that a multi-valued query parameter can be specified *either* by
+        a comma-separated single string value *or* by a list of individual
+        values, not both.
```
webbnh, line 665: So, if the schema defines `listparm` as a list parameter, and the request provides
```
?listparm=a,b,c&listparm=d`
```
I can imagine any of the following as being "right" answers:
- `["a", "b", "c"]` (first value wins)
- `["d"]` (last values wins)
- `["a", "b", "c", "d"]` (concatenated lists)
- `RepeatedQueryParam` (...well, yeah...)

But, what I think we get is `["a,b,c", "d"]` which I think is legal (for some twisted interpretation of that) but is unlikely to be what the user wanted.

Do we just ignore this case, with the expectation that the resulting key error for `"a,b,c"` will offer enough of a clue about what the requestor did wrong?

lib/pbench/server/api/resources/__init__.py
```diff
@@ -663,6 +685,8 @@ def _validate_query_params(self, request: Request, schema: Schema) -> JSONOBJECT
                 if len(values) > 1 and schema[key].type == ParamType.LIST:
                     json[key] = values
                 else:
+                    if len(values) > 1:
+                        raise RepeatedQueryParam(key)
                     json[key] = values[0]
```
webbnh, line 690: It seems a shame to ask the length question twice...so how about:
```
                if len(values) > 1:
                    if schema[key].type == ParamType.LIST:
                        json[key] = values
                    else:
                        raise RepeatedQueryParam(key)
                else:
                    json[key] = values[0]
```
or, this, which is prettier:
```
                if len(values) == 1:
                    json[key] = values[0]
                elif schema[key].type == ParamType.LIST:
                    json[key] = values
                else:
                    raise RepeatedQueryParam(key)
```
but, unless there's a really good reason for avoiding a "list of one", this strikes me as best:
```
                if schema[key].type == ParamType.LIST:
                    json[key] = values
                elif len(values) == 1:
                    json[key] = values[0]
                else:
                    raise RepeatedQueryParam(key)
```

webbnh (approved):
Just down to a nit or two...I think.

npalaska (dismissed):

lib/pbench/server/api/resources/__init__.py
dbutenhof: > Do we just ignore this case, with the expectation that the resulting key error for "a,b,c" will offer enough of a clue about what the requestor did wrong?

Yeah, that was more or less my intent, just to keep this simple. But when you repeat my words back to me like that, I don't like what I'm hearing at all, and now you're going to make me go back and massively refactor this.

I think that the concatenated list makes the most sense, but my string to list converter needs to be more general and therefore more controlled. I talked myself out of doing this, barely, and once again you've pushed me over the edge.

I'm actually not sure whether to thank you, though as usual I suspect I'll be happier with the results. 😒

lib/pbench/server/api/resources/__init__.py
dbutenhof: There was actually a reason to avoid the "list of one", because that would break my comma-list-string logic. However, now that you've twisted my arm and forced me to generalize that, "list of one" should be fine and I do prefer the simplified code here. Maybe that's another reason to re-work this whole mess. (Sigh.)

ndokos (dismissed):

npalaska:
Is this ready for review?

dbutenhof:
> Is this ready for review?

Oops; forgot to toggle it back to Open. Yes.

portante (dismissed):

npalaska (approved):

ndokos (approved):

ndokos (approved):

lib/pbench/server/api/resources/__init__.py
```diff
@@ -384,17 +384,22 @@ def convert_list(value: Union[str, List[str]], parameter: "Parameter") -> List[A
...
-    etype: "ParamType" = parameter.element_type
+    etype: Union["ParamType", None] = parameter.element_type
     retlist = []
     errlist = []
     for v in values:
         try:
-            retlist.append(etype.convert(v, parameter))
+            retlist.append(etype.convert(v, parameter) if etype else v)
         except SchemaError:
             errlist.append(v)
```
webbnh, line 404: It probably won't make a huge difference (since we don't expect the values list to be very long), but consider recasting this code as:
```
    if etype:
        for v in values:
            try:
                retlist.append(etype.convert(v, parameter) if etype else v)
            except SchemaError:
                errlist.append(v)
    else:
        retlist = values
```
That is, if there's no `etype`, then there's no point in going through the loop -- we can just return `values`.

webbnh (approved):
👍

(Just one observation.)

lib/pbench/server/api/resources/__init__.py
webbnh: I note that you went with option #2 which avoids the list-of-one result....

dbutenhof: And once again you've got me second-guessing myself.

I thought "why create a list of one element when it's not meaningful" ... however it occurs to me after the fact that this isn't quite correct. Now that I've made the list-ization of a string conditional on defining a separator, I'm back to having a real difference; if the subclass calling this common function didn't choose to define a separator for a LIST-type query parameter, the non-list single entry will be rejected.

Right now it doesn't matter, because "metadata" is the only LIST type query parameter, and it has separators. However, if we had a LIST with no separator defined, it'd be important to return a single value as a List[]. So I think something like your 3rd option is necessary, after all.

portante (approved):
The unit tests failures are due to `perf-fio-0` timing out, which has nothing to do with these changes.

webbnh (approved):
👍

webbnh:
Apparently @portante kicked it.

I think we have an intermittent failure running around, but I think it's in the infrastructure, not in our tests.

PR #2794, opened 2022-04-28
Use HTTPS for D3 loads

portante:


webbnh (approved):
Hurray for fixing LGTM alerts!

PR #2797, opened 2022-04-28
Isolate parameter transforms to ToolDataSink class

portante:
We streamline parameter passing and front-load parameter object creation.


portante (commented):
Ready for review.

webbnh (dismissed):
Looks OK to me.  I have two thoughts, though.

lib/pbench/agent/tool_data_sink.py
```diff
@@ -1878,45 +1896,34 @@ class Arguments(NamedTuple):
 
 
 def driver(
-    PROG: str,
+    ext_env: ExternalEnvironment,
```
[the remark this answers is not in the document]
dbutenhof: Yeah, it's always bothered me a little that `PROG` was defined "as a literal"; I've always told myself that was more churn than justified for whatever I was doing at the time, and that's probably still true.

dbutenhof (approved):

lib/pbench/agent/tool_data_sink.py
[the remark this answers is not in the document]
portante: I'll take a stab at this nonetheless.

lib/pbench/agent/tool_meister.py
```diff
@@ -1834,19 +1834,19 @@ class Arguments(NamedTuple):
...
     tar_path: str,
     sysinfo_dump: str,
     pbench_install_dir: Path,
     tmp_dir: Path,
     parsed: Arguments,
     tm_params: ToolMeisterParams,
     redis_server: redis.Redis,
+    logger_name: str,
     logger: logging.Logger = None,
 ):
```
webbnh, line 1846: Passing _both_ the `logger_name` and the `logger` produces a weird interface.  Since the only purpose of the `logger_name` is for use in creating the logger, iff it wasn't specified, then we really do not need both (or, at least, we don't need to require `logger_name` in the case when `logger` is specified).  But, rather than making `logger_name` optional and all that cruft, it would be cleaner to just have the caller provide the logger.  Now that creating the logger has been nicely encapsulated in `get_logger()`, I think that this is easy to do.

Execution flows from `start()` to `driver()` with an optional side-trip through `daemon()`.  We can easily create the logger in `start()`, where we have access to all of `get_logger()`'s parameter values for either path; and we can define both `daemon()` and `driver()` to expect to receive a logger (and not a logger name), with `daemon()` passing the logger through to `driver()`.

This yields a cleaner interface, without all the conditionalization.  (The only conditionalization is in `start()`'s call to get logger, and that's simple:  pass `get_logger(prog.name, is_daemon=parsed.daemonize, level=parsed.level)` for the logger value.)

webbnh (dismissed):
Looks great!

However, I have one suggestion which I hope you will entertain.

lib/pbench/agent/tool_meister.py
webbnh: Oh, yeah, 'cause we're letting the daemonization close _all_ the file descriptors...right.  (I wonder how the daemon manages to create a new logger....  🤯)

Bleh.  👃 🤏

portante: Bleh is right!  If you can find a better way, I'll all for it!

webbnh: Well, in a perfect world, we would do all the process setup and then daemonize the process _keeping_ the setup.  But, I'm not sure how we locate (e.g.) the open file descriptors used by the `Redis` server object and the logger, so that we can tell the `DaemonizeContext` not to close them...but, if we could do that, then we wouldn't have to recreate those things after we daemonize, which might simplify things.

But, I think we have other fish to fry.

dbutenhof (dismissed):
Looks good. Sadly, you have a merge conflict to resolve. 😄

webbnh (approved):
👍

dbutenhof (approved):

portante (commented):
Rebased, ready to be merged.

Should we squash and merge or just rebase keeping this as two separate commits?

portante:
FWIW, each commit independently passes all our tests.

dbutenhof (approved):
I'd just go for the squash-and-merge unless you plan to cherry-pick just one of them back to 0.71??

webbnh (approved):
I vote for squash and merge (I don't think these are going back to v0.71).

npalaska (approved):

PR #2799, opened 2022-04-28
Delete the spurious pbench3-devel RPM dependency

### LINKED ISSUE #2798: [BUG] Installation of pbench-agent RPM fails on RHEL7.9 because of unsatisfied dependency

### Description

Currently the spec file for the pbench-agent requires `pbench3-devel` in the RHEL7 case. However, this package does not exist in RHEL7.9 and the installation of the pbench-agent fails. 

### To Reproduce

Build an RPM for the pbench-agent and install it.

### Actual Results

The installation fails:

```
Resolving Dependencies
--> Running transaction check
---> Package pbench-agent.noarch 0:0.71.0-13g2c579a21e will be installed
--> Processing Dependency: python3-devel for package: pbench-agent-0.71.0-13g2c579a21e.noarch
--> Finished Dependency Resolution
Error: Package: pbench-agent-0.71.0-13g2c579a21e.noarch (copr-pbench-test-ndokos)
           Requires: python3-devel
```

### Expected Results

The installation should succeed.

### Additional information

_No response_

ndokos:
Fixes #2798

AFAICT, the python3-devel dependency (which only exists for RHEL7) is
spurious and in fact harmful: there seems to be no python3-devel
package for RHEL7.9, so installation fails with an unsatisfied
dependency, whereas an RPM built without it installs without
problems.

### DIFF
--- a/agent/rpm/pbench-agent.spec.j2
+++ b/agent/rpm/pbench-agent.spec.j2
@@ -16,9 +16,6 @@ Requires:  ansible-core
 
 %if 0%{?rhel} == 7
 Requires:  python3, python3-pip
-# For RHEL boxen we need the python development environment in order for the
-# pip3 installs for some modules to be successful.
-Requires:  gcc python3-devel
 %endif
 
 %if 0%{?rhel} == 8


webbnh (dismissed):
Do we need `gcc`?  😁

portante (changes requested):
Looks good.  We should double check, but I don't think we need `gcc` as a result of removing this dependency.

dbutenhof (dismissed):
Interesting...

portante (approved):

dbutenhof (approved):

webbnh (approved):
🎉

PR #2804, opened 2022-04-30
Convert server APIs to ISO 8601 UTC dates

dbutenhof:
PBENCH-638

SQLAlchemy stores SQL dates in "naive UTC" by default; add logic to make this explicit and to add adjustment of aware dates on input and creation of aware UTC dates on retrieval.

Adopt ISO 8601 formatting for dates reported through the API. (This does not include a rigorous analysis of Elasticsearch data, although at least some already appear in ISO 8601 UTC.)

### DIFF
--- a/lib/pbench/server/api/resources/upload_api.py
+++ b/lib/pbench/server/api/resources/upload_api.py
@@ -6,18 +6,18 @@ import os
 
 import humanize
 from flask import jsonify, request
-from flask_restful import Resource, abort
+from flask_restful import abort, Resource
 
-from pbench.common.utils import validate_hostname, Cleanup
-from pbench.server.filetree import DatasetNotFound, FileTree, Tarball
+from pbench.common.utils import Cleanup, validate_hostname
 from pbench.server.api.auth import Auth
 from pbench.server.database.models.datasets import (
     Dataset,
     DatasetDuplicate,
-    States,
     Metadata,
+    States,
 )
-from pbench.server.utils import filesize_bytes
+from pbench.server.filetree import DatasetNotFound, FileTree, Tarball
+from pbench.server.utils import IsoTimeHelper, filesize_bytes
 
 
 class CleanupTime(Exception):
@@ -296,7 +296,8 @@ class Upload(Resource):
             # won't be committed to the database until the "advance" operation
             # at the end.
             try:
-                dataset.created = tarball.get_metadata()["pbench"]["date"]
+                metadata = tarball.get_metadata()
+                dataset.created = metadata["pbench"]["date"]
             except Exception as exc:
                 raise CleanupTime(
                     HTTPStatus.BAD_REQUEST,
@@ -330,7 +331,9 @@ class Upload(Resource):
                     value=str(tarball.tarball_path),
                 )
                 Metadata.setvalue(
-                    dataset=dataset, key=Metadata.DELETION, value=f"{deletion:%Y-%m-%d}"
+                    dataset=dataset,
+                    key=Metadata.DELETION,
+                    value=IsoTimeHelper(deletion).iso(),
                 )
             except Exception as e:
                 raise CleanupTime(
--- a/lib/pbench/server/api/resources/upload_api.py
+++ b/lib/pbench/server/api/resources/upload_api.py
@@ -17,7 +17,7 @@ from pbench.server.database.models.datasets import (
     States,
 )
 from pbench.server.filetree import DatasetNotFound, FileTree, Tarball
-from pbench.server.utils import IsoTimeHelper, filesize_bytes
+from pbench.server.utils import UtcTimeHelper, filesize_bytes
 
 
 class CleanupTime(Exception):
@@ -333,7 +333,7 @@ class Upload(Resource):
                 Metadata.setvalue(
                     dataset=dataset,
                     key=Metadata.DELETION,
-                    value=IsoTimeHelper(deletion).iso(),
+                    value=UtcTimeHelper(deletion).to_iso_string(),
                 )
             except Exception as e:
                 raise CleanupTime(
--- a/lib/pbench/server/api/resources/users_api.py
+++ b/lib/pbench/server/api/resources/users_api.py
@@ -1,16 +1,14 @@
 from http import HTTPStatus
-from typing import NamedTuple
-
-from email_validator import EmailNotValidError
-from flask import jsonify, make_response, request
-from flask_restful import abort, Resource
-from flask_bcrypt import check_password_hash
 import jwt
-from sqlalchemy.exc import IntegrityError, SQLAlchemyError
-
-from pbench.server.api.auth import Auth
-from pbench.server.database.models.active_tokens import ActiveTokens
+from flask import request, jsonify, make_response
+from flask_restful import Resource, abort
+from flask_bcrypt import check_password_hash
+from email_validator import EmailNotValidError
+from sqlalchemy.exc import SQLAlchemyError, IntegrityError
+from typing import NamedTuple
 from pbench.server.database.models.users import User
+from pbench.server.database.models.active_tokens import ActiveTokens
+from pbench.server.api.auth import Auth
 
 
 class RegisterUser(Resource):
--- a/lib/pbench/server/api/resources/users_api.py
+++ b/lib/pbench/server/api/resources/users_api.py
@@ -1,14 +1,16 @@
 from http import HTTPStatus
-import jwt
-from flask import request, jsonify, make_response
-from flask_restful import Resource, abort
-from flask_bcrypt import check_password_hash
-from email_validator import EmailNotValidError
-from sqlalchemy.exc import SQLAlchemyError, IntegrityError
 from typing import NamedTuple
-from pbench.server.database.models.users import User
-from pbench.server.database.models.active_tokens import ActiveTokens
+
+from email_validator import EmailNotValidError
+from flask import jsonify, make_response, request
+from flask_restful import abort, Resource
+from flask_bcrypt import check_password_hash
+import jwt
+from sqlalchemy.exc import IntegrityError, SQLAlchemyError
+
 from pbench.server.api.auth import Auth
+from pbench.server.database.models.active_tokens import ActiveTokens
+from pbench.server.database.models.users import User
 
 
 class RegisterUser(Resource):
--- a/lib/pbench/server/database/models/datasets.py
+++ b/lib/pbench/server/database/models/datasets.py
@@ -314,6 +314,10 @@ def current_time() -> datetime.datetime:
     """
     Return the current time in UTC.
 
+    This provides a Callable that can be specified in the SQLAlchemy Column
+    to generate an appropriate (aware UTC) datetime object when a Dataset
+    object is created.
+
     Returns:
         Current UTC timestamp
     """
@@ -337,7 +341,7 @@ class TZDateTime(TypeDecorator):
         objects are converted to UTC and made "naive" by replacing the TZ
         for SQL storage.
         """
-        if value is not None and not value.tzinfo:
+        if value is not None and value.utcoffset() is not None:
             value = value.astimezone(datetime.timezone.utc).replace(tzinfo=None)
         return value
 
--- a/lib/pbench/server/database/models/datasets.py
+++ b/lib/pbench/server/database/models/datasets.py
@@ -2,13 +2,14 @@ import copy
 import datetime
 import enum
 import os
-import re
 from pathlib import Path
+import re
 from typing import Any, List, Tuple, Union
 
-from sqlalchemy import event, Column, DateTime, Enum, ForeignKey, Integer, JSON, String
+from sqlalchemy import Column, DateTime, Enum, event, ForeignKey, Integer, JSON, String
+from sqlalchemy.exc import IntegrityError, SQLAlchemyError
 from sqlalchemy.orm import relationship, validates
-from sqlalchemy.exc import SQLAlchemyError, IntegrityError
+from sqlalchemy.types import TypeDecorator
 
 from pbench.server.database.database import Database
 from pbench.server.database.models.users import User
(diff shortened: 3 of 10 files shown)


portante (dismissed):
Any way we can get the IDEs to resist sorting all the imports along with the other changes being made?  Perhaps having the sorting done in a separate commit first?

dbutenhof:
> Any way we can get the IDEs to resist sorting all the imports along with the other changes being made? Perhaps having the sorting done in a separate commit first?

It's partly the IDEs that make it easy to result in out of order imports. Yes, we could target a PR (or a set) just to clean this up, but I also think that cleanup and refactoring is a normal part of development and (if it's not extreme and overly distracting) *should* be done routinely in the course of normal work.

And out-of-order imports always gets a mention in reviews anyway; it does make some sense to be proactive. There's a fine balance between throwing a few extra isolated diffs into a review vs the overhead of managing a second review, and we can easily disagree on the threshold. In a case like this, it takes maybe an extra second or two to scan the minor reordering; seems like a good tradeoff to me.

dbutenhof:
> PR #2805 took 25 minutes to break out, fixes a few sorting problems, shrinks the 2nd commit by one file, and makes the 2nd commit a bit easier to review.
> 
> Seems worth it to break cleanup refactoring out from other code changes.

Wow.

I glad you had fun spending "25 minutes" of your weekend tearing up my PR.

Frankly, while I'll admit to some incremental value in separating out incidental refactoring, this sort of micro-fiddling with commits just to isolate already distinct small changes is not something I would ever judge to be worth that amount of extra effort.

I would very much like to make continuous incremental improvements part of the normal workflow, and that means making it easy... without having to go back and second-guess how the cleanup is presented much less fiddling with reorganizing lines of change across git commits.

If we're going to follow a rule that "incidental" cleanup can't occur "inline" then I know that I for one will be doing a lot less of it, and I think that would be sad; in particular I would not have bothered fixing any imports in this PR. Is that really how we want to do things?

I look at this as a matter of scale; the refactoring is isolated from the code, and it's just imports. I don't see it as any harder to review in one integrated commit than in separate commits. It wasn't a massive rewrite of half a module. There has to be a point where we draw the line; but if we're going to set the bar this low, I find the level of "makework" implied to be demotivating.

portante:
> > PR #2805 took 25 minutes to break out, fixes a few sorting problems, shrinks the 2nd commit by one file, and makes the 2nd commit a bit easier to review.
> > Seems worth it to break cleanup refactoring out from other code changes.
> 
> Wow.

Wow is right.  Oh well.  We'll see how this plays out.

lib/pbench/server/filetree.py
```diff
@@ -274,11 +274,13 @@ def get_metadata(self) -> JSON:
...
         metadata.read_string(data)
 
+        created = IsoTimeHelper.from_string(metadata.get("pbench", "date"))
+
         return {
             "controller": {"hostname": metadata.get("controller", "hostname")},
             "pbench": {
                 "config": metadata.get("pbench", "config"),
-                "date": date_parser.parse(metadata.get("pbench", "date"), None),
+                "date": created.utc_time,
```
webbnh, line 283: Perhaps I'm biased (or even a bigot), but shouldn't this be a functional interface rather than a data interface, i.e., shouldn't we be calling a getter function rather than accessing an attribute directly?  (I realize that, in Python, there's not much difference between the two, but for the sake of form?)

lib/pbench/server/utils.py
```diff
@@ -88,3 +91,23 @@ def quarantine(dest, logger, *files):
...
             sys.exit(102)
+
+
+class IsoTimeHelper:
+    def __init__(self, time: datetime.datetime):
+        self.utc_time = time
+        if (
+            self.utc_time.tzinfo is None
+            or self.utc_time.tzinfo.utcoffset(self.utc_time) is None
+        ):
```
webbnh, line 102: Can this be [simplified](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcoffset) to just?
```
        if (self.utc_time.utcoffset() is None):
```

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -130,7 +131,9 @@ def delete(self):
                 TestUpload.tarball_deleted = self.name
 
             def get_metadata(self):
-                return {"pbench": {"date": dateutil.parser.parse("2002-05-16")}}
+                return {
+                    "pbench": {"date": IsoTimeHelper.from_string("2002-05-16").utc_time}
```
webbnh, line 135: Do you really want to use `IsoTimeHelper` here, as opposed to `datetime.datetime.isoformat()`, like you did in `lib/pbench/test/unit/server/test_datasets_list.py`?  (Should the two be using the same approach?)

lib/pbench/test/unit/server/test_requests.py
```diff
@@ -406,10 +409,12 @@ def test_upload(
...
         assert dataset.state == States.UPLOADED
-        assert dataset.created == dateutil.parser.parse("2002-05-16")
-        assert dataset.uploaded == dateutil.parser.parse("1970-01-01")
+        assert dataset.created == date_parser.parse("2002-05-16T00:00:00+00:00")
+        assert dataset.uploaded == date_parser.parse("1970-01-01T00:00:00+00:00")
         assert Metadata.getvalue(dataset, "dashboard") is None
-        assert Metadata.getvalue(dataset, Metadata.DELETION) == "1972-01-01"
+        assert (
+            Metadata.getvalue(dataset, Metadata.DELETION) == "1972-01-01T00:00:00+00:00"
+        )
```
webbnh, line 417: Is there a reason why the `dataset` `created` and `uploaded` timestamps are compared as `datetime` objects while the `DELETION` value is compared as a string?  (E.g., should they all be compared in the same way?  Might there be an advantage or disadvantage to comparing them one way instead of the other?)

webbnh (approved):
Looks OK to me.  I have the usual handful of items to consider.

Also, should there be unit tests for the `IsoTimeHelper` class?

portante (approved):

lib/pbench/server/api/resources/users_api.py
```diff
@@ -1,14 +1,16 @@
...
+from email_validator import EmailNotValidError
+from flask import jsonify, make_response, request
+from flask_restful import abort, Resource
+from flask_bcrypt import check_password_hash
+import jwt
+from sqlalchemy.exc import IntegrityError, SQLAlchemyError
+
 from pbench.server.api.auth import Auth
+from pbench.server.database.models.active_tokens import ActiveTokens
+from pbench.server.database.models.users import User
```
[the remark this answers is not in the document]
webbnh: I didn't mean to imply that this change should now be removed from this PR...I meant to indicate that, a priori, I wouldn't have included it.  🍤

lib/pbench/test/unit/server/test_requests.py
webbnh: The compelling (er, to the extent that it is...) reason would be to minimize the amount of project code used to test each unit (to avoid unanticipated interactions between disparate components).  That is, there is no compelling reason to use `IsoTimeHelper` when the standard library function will suffice.  (But, if you're not inclined to change this, then I'm not feeling compelled to argue.)

lib/pbench/test/unit/server/test_requests.py
webbnh: I guess what I'm wondering is whether the type of the `created` and `uploaded` timestamps is important, or whether we are ultimately most concerned with their values, particularly their values as they will be rendered at the interface.  So, it struck me that we would prefer to do the checks as strings...but, it's not a big deal.  (As far as `DELETION` goes, if it is already stored as a string, then there's no question, there.)

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -17,7 +17,7 @@
     States,
 )
 from pbench.server.filetree import DatasetNotFound, FileTree, Tarball
-from pbench.server.utils import IsoTimeHelper, filesize_bytes
+from pbench.server.utils import UtcTimeHelper, filesize_bytes
```
webbnh, line 20: Are we OK with the order of the imports on line 20?  🙃

lib/pbench/server/database/models/datasets.py
```diff
@@ -311,17 +312,44 @@ def __str__(self) -> str:
 
 def current_time() -> datetime.datetime:
     """
-    Return the current time.
-
-    NOTE: we use this intermediary rather than binding the Column default
-    directly to datetime.datetime.now because that load-time binding defeats
-    unit test mocking. Binding it instead to this method allows a mock to
-    take effect.
```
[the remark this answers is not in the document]
webbnh: When I didn't see any reference, I suspect I was doing my search with a trailing open parenthesis...because searching without it finds the two you mentioned with no problem.  🥴 

> I could have done that with a simple `datetime.datetime.now`, but not with the `datetime.datetime.now(datetime.timezone.utc)` I need now.

Would [`datetime.datetime.utcnow()`](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow) work?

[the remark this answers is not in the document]
webbnh: Or...if you really wanted...you could use a `lambda`:
```
12345678901234567890123456789012345678901234567890123456789012345678901234567890
    uploaded = Column(
        TZDateTime,
        nullable=False,
        default=lambda: datetime.datetime.now(datetime.timezone.utc))
```
🙃

lib/pbench/server/utils.py
```diff
@@ -93,21 +93,47 @@ def quarantine(dest, logger, *files):
...
-class IsoTimeHelper:
+class UtcTimeHelper:
+    """
+    A helper class to work with UTC "aware" datetime objects. A "naive" object
+    (without timezone offset) will be set to UTC timezone.
+    """
+
     def __init__(self, time: datetime.datetime):
+        """
+        Capture a datetime object and; if it's "naive", set it to be UTC; if
```
webbnh, line 104: s/and;/and,/

lib/pbench/server/utils.py
```diff
@@ -93,21 +93,47 @@ def quarantine(dest, logger, *files):
...
+        """
         self.utc_time = time
-        if (
-            self.utc_time.tzinfo is None
-            or self.utc_time.tzinfo.utcoffset(self.utc_time) is None
-        ):
+        if self.utc_time.utcoffset() is None:
             self.utc_time = self.utc_time.replace(tzinfo=datetime.timezone.utc)
+        else:
+            self.utc_time = self.utc_time.astimezone(datetime.timezone.utc)
```
webbnh, line 114: If `self.utc_time` is already "aware", we're going to set it to UTC, even if it is already there....

I wonder if the code would be better if we changed line 111 to just
```
        if !self.utc_time.utcoffset():
```
so that both "naive" and UTC times are treated the same way, and only "aware", non-UTC times get converted.

lib/pbench/server/api/resources/upload_api.py
dbutenhof: The bulk name change didn't stop to consider this, and for a case-sensitive sort it's arguably still correct. Are we going to define case-blind sorting for imports? That's always been ambiguous, and I wouldn't really be inclined to worry about that right now.

lib/pbench/server/database/models/datasets.py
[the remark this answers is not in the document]
dbutenhof: No, `.utcnow()` returns the current naive time in UTC; I want aware time. And I don't see any value in blowing up the Column definition with a long lambda.

lib/pbench/test/unit/server/test_requests.py
dbutenhof: > reason would be to minimize the amount of project code used to test each unit

Except that this is a mock masquerading as product code, so the cut is a bit more subtle than that. I liked the idea of applying the same mechanism.

lib/pbench/test/unit/server/test_utils.py
```diff
@@ -35,3 +38,23 @@ def test_filesize_bytes():
             ), f"string '{size}', improperly converted to {res}, expected {exp_val}"
         with pytest.raises(Exception):
             res = filesize_bytes("bad")
+
+
+class TestUtcTimeHelper:
+    @freeze_time("1970-01-01")
+    def test_naive(self):
+        d = UtcTimeHelper(datetime.datetime.now())
+        assert d.to_iso_string() == "1970-01-01T00:00:00+00:00"
```
webbnh, line 47: Rather than relying on `freeze_time`, wouldn't it be better to just use the standard library?
```
    def test_naive(self):
        the_date = "1970-01-01T00:00:00"
        d = UtcTimeHelper(datetime.datetime.fromisoformat(the_date))
        assert d.to_iso_string() == the_date + "+00:00"

    def test_aware_utc(self):
        the_date = "2002-05-16T10:23:00+00:00"
        d = UtcTimeHelper(datetime.datetime.fromisoformat(the_date))
        assert d.to_iso_string() == the_date
```
In fact, it looks like this could be parameterized by providing parameters for input and output strings.

lib/pbench/server/api/resources/upload_api.py
webbnh: Yeah, I was wondering the same thing...but perhaps we can kick this can further down the road.

lib/pbench/server/database/models/datasets.py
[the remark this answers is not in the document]
webbnh: Ah, sorry -- I was confused:  I thought this was the context where we wanted a naive time.

webbnh (approved):
Looks good; I just have some parting thoughts.

lib/pbench/server/api/resources/upload_api.py
webbnh: (Personally, I would prefer the case-blind approach.)

lib/pbench/server/utils.py
dbutenhof: I was trying to separate the two comma clauses: admittedly, I did this incorrectly as I should have used ":" instead of "and;" to set it up. I must have been tired.

lib/pbench/test/unit/server/test_utils.py
dbutenhof: This felt more natural at the time; arguably it doesn't really matter since the class isn't creating the timestamp.

And, FYI, `datetime.datetime.fromisoformat` is new in Python 3.7; we're still supporting Python 3.6 (default on RHEL 8.4), so this required `dateutil.parser`. Ah well...

lib/pbench/server/utils.py
dbutenhof: I suppose technically we don't need to do *either* to an aware UTC time, but I'm not sure I really want to add an extra case. I think that I vaguely feel slightly better about "adjusting" a UTC TZ to UTC than "replacing" the TZ... though I can't entirely justify why.

portante (approved):

lib/pbench/server/utils.py
webbnh: I didn't really want to add an extra case, either, which is why I suggested collapsing the "naive" and UTC cases into the same one (and, because I'm guessing that replacing the `tzinfo` is cheaper than applying the timezone offset, so we "win" if we shunt more flows down that path).

Nevertheless, it seems like it should be the case that we shouldn't have to do anything to "aware" values until we try to render them into strings, at which point we should just stipulate which timezone they should come out in and not worry about what timezone they went in in.  But, I confess, I don't grok what the requirements are for SQLalchemy...perhaps that requires actually setting the timezone to UTC for values that we are going to feed into it.

webbnh (approved):
Looks good!

PR #2807, opened 2022-05-02
Remove controllers/list API

dbutenhof:
PBENCH-640

The Elastic-based controllers/list API is obsolete as we move to focus on USERS and DATASETS instead of relying on the old concept of the "controller" (which was a standin for USER when testers could rely on identifying their own dedicated systems).


portante (dismissed):

webbnh (approved):
👍

webbnh (approved):
No change; must still be good.

portante (approved):

PR #2810, opened 2022-05-02
Remove controllers/months API

dbutenhof:
PBENCH-639

The range of months (YYYY-MM) containing Elasticsearch run documents for a specific controller name is no longer of any interest, so remove this obsolete API.

(TBD whether this should be replaced with a new API returning a date range for datasets available perhaps with a specific owner and access filter, e.g. for purposes of pre-loading a date picker?)


portante (dismissed):

webbnh (dismissed):
Deleting code is so much fun!

> should be replaced with a new API returning a date range for datasets available perhaps with a specific owner and access filter, e.g. for purposes of pre-loading a date picker?

My knee-jerk reaction is "yes", but how hard would it be for the Dashboard to glean that information from the existing APIs?  (If there are a thousand responses datasets, does that change the answer?)

dbutenhof:
> My knee-jerk reaction is "yes", but how hard would it be for the Dashboard to glean that information from the existing APIs? (If there are a thousand responses datasets, does that change the answer?)

The real issue is whether we want to drive the date picker default by the actual range of dates. Shubham's not doing this now. We *could* default to include all accessible datasets; but if the list goes back years with thousands of entries, does that make sense?

To scale, we're going to need to stop just fetching *everything*, because it'll end up being a lot. The date picker allows narrowing that down; an advantage that of course goes away if we default to a range that includes "everything". (Even though a new dataset-oriented query would report only datasets that are accessible to the client, and potentially match owner and access filters.)

Anyway, if we don't fetch everything, and we want to know the oldest and newest with reasonable scaling, the SQL "select min(create),max(created) from datasets" would be relatively efficient, and possibly worth coding. (In fact, I've coded it, but stopped to think before trying to test it, and now I'm undecided whether we actually need/want this.)

portante:
> > My knee-jerk reaction is "yes", but how hard would it be for the Dashboard to glean that information from the existing APIs? (If there are a thousand responses datasets, does that change the answer?)
> 
> The real issue is whether we want to drive the date picker default (in the Pbench Dashboard) by the actual range of dates. 

Seems like something we should drive in a separate, future, PR.

webbnh:
> The real issue is whether we want to drive the date picker default by the actual range of dates.

Right...and, personally, I think that it would be a cool thing if we did.  For the logged-in user, it's a nice prompt or bit of history; for the non-logged-in user, it's kind of like putting "In business since..." on our "sign".  In both cases, its a fairly small and rather subtle thing, but, if we can do it efficiently, I think it's worthwhile...in another PR....

dbutenhof:
> > The real issue is whether we want to drive the date picker default by the actual range of dates.
> 
> Right...and, personally, I think that it would be a cool thing if we did. For the logged-in user, it's a nice prompt or bit of history; for the non-logged-in user, it's kind of like putting "In business since..." on our "sign". In both cases, its a fairly small and rather subtle thing, but, if we can do it efficiently, I think it's worthwhile...in another PR....

Heh; I mentioned it in this PR as a "TBD" not as a direct question, but the discussion is useful all the same. I put in a Jira subtask this morning when I started this series, and maybe tomorrow I'll pick up the query again.

portante (approved):

webbnh (approved):
👍

PR #2817, opened 2022-05-04
User Profile Page

MVarshini:
User profile displays information associated with a user. It contains information such as First name, last name, user name, email and account created date.

This page allows the user to update the information like `first name`, `last name` and `email`


dashboard/src/actions/profileActions.js
```diff
@@ -0,0 +1,92 @@
...
+    if (response.status === 200 && Object.keys(response.data).length > 0) {
+      dispatch({
+        type: TYPES.GET_USER_DETAILS,
+        payload: response?.data,
+      });
+    } else {
+      let toast = {
+        variant: "danger",
+        title: response?.data?.message,
+      };
```
dbutenhof, line 22: Why do the `authActions` methods assume that API errors go on the `catch(error)` path while here the API error handling is on an `else` and the `catch` seems to assume total `HTTP` "network error" failure? Is this `API.get` vs `API.post`, or is there something else going on that I'm missing?

dashboard/src/actions/profileActions.js
```diff
@@ -0,0 +1,92 @@
...
+    let userDetails = { ...getState().userProfile.userDetails };
+    userDetails[fieldName] = value;
+
+    dispatch({
+      type: TYPES.UPDATE_USER_DETAILS,
+      payload: { ...userDetails },
+    });
+  };
+
+export const sendForUpdation = () => async (dispatch, getState) => {
```
dbutenhof, line 46: I kinda the sound like "updation" as a verb, but just `sendForUpdate` would probably be better. 😄

dashboard/src/reducers/userProfileReducer.js
```diff
@@ -0,0 +1,29 @@
...
+    case GET_USER_DETAILS:
+      return {
+        ...state,
+        userDetails: Object.assign({}, payload),
+        userDetails_copy: Object.assign({}, payload),
+      };
+    case UPDATE_USER_DETAILS:
+      return {
+        ...state,
+       // userDetails: { ...setValue(state.userDetails, action) },
```
portante, line 20: Is this a reminder?

dashboard/src/actions/profileActions.js
```diff
@@ -0,0 +1,92 @@
...
+          type: TYPES.SHOW_TOAST,
+          payload: { ...toast },
+        });
+      }
+      dispatch({ type: TYPES.COMPLETED });
+      return true;
+    }
+  } catch (error) {
+    dispatch({ type: TYPES.COMPLETED });
+  }
```
dbutenhof, line 79: So on user update we're just ignoring errors completely?

dashboard/src/modules/components/ProfileComponent/index.jsx
```diff
@@ -0,0 +1,243 @@
...
+    const registeredDate = new Date(registered);
+    let mon = registeredDate?.getMonth() + 1;
+    mon = mon < 10 ? "0" + mon : mon;
+    return (
+      mon +
+      "/" +
+      registeredDate?.getDate() +
+      "/" +
+      registeredDate?.getFullYear()
+    );
```
dbutenhof, line 61: Would something like `Date.prototype.toLocaleTimeString()` be better here? I don't really like the idea of a fixed hardcoded date format. Even though this is "US standard" it's going to look odd to a lot of people... and I have always preferred the YYYY-MM-DD format myself because it sorts naturally. If we use the built-in localization mechanisms, that's one less detail to worry about.

dashboard/src/actions/profileActions.js
```diff
@@ -0,0 +1,92 @@
...
+      type: TYPES.UPDATE_USER_DETAILS,
+      payload: { ...userDetails },
+    });
+  };
+
+export const sendForUpdation = () => async (dispatch, getState) => {
+  try {
+    dispatch({ type: TYPES.LOADING });
+    const userDetails = { ...getState().userProfile.userDetails };
+    const userDetails_copy = { ...getState().userProfile.userDetails_copy };
```
portante, line 50: Can we consider inverting the names?  Instead of using `_copy` to mark the values that are the actual, perhaps we could consider `userDetails` to be the original copy, and add `userDetails_updates` where it is empty until somebody makes changes, and then when the final button is pressed to commit them, the values are moved from `_updates` to `userDetails`?

portante (commented):
Looking good!

dbutenhof (commented):
Just a few comments on a first scan through this ...

dashboard/src/actions/profileActions.js
MVarshini: Done! :grinning:

dashboard/src/actions/profileActions.js
MVarshini: The `_copy` is needed for the `reset` functionality. Because only with `userDetails` the UI is rendered. On changing the input values and for that to reflect in the UI, need to update the `userDetails` object.

Code changes:
-  Modified the code with `_updates` where only the updated fields get inside it and sent for update
-  `Save` button is disabled until the user updates any of the fields
-  `Cancel` and `Reset to default` both the buttons are maintained. If not, the `Settings` card would have lot of empty space

**Working of `Cancel` button**
Reverts back to the `normal` view from the `edit` view while retaining the last saved state

**Working of `Reset` button**
Reverts to the last saved state but maintains the `edit` view

dbutenhof (dismissed):
This is looking good!

My main comment would be that we're adding a lot of code to the dashboard without unit tests, even though you'd created a basic unit test framework and "test" command.

In the agent and server we try to maintain a rule that anything new or changed needs to be tested. We're definitely not perfect, and we don't have real functional testing nor anywhere near complete unit test coverage. Still, during this early rapid evolution of the new dashboard we can either build a solid testing framework or build up a substantial technical debt to resolve later.

At this point, I'm just "throwing this out" for consideration; but maybe after this and Shubham's browsing code goes in, it would be a good idea to spend some time writing unit tests for the existing code before going further?

dashboard/src/modules/components/ProfileComponent/index.jsx
```diff
@@ -217,7 +215,7 @@ const ProfileComponent = () => {
                     <TextContent>
                       <span>Account creation Date</span>
                       <Text component={TextVariants.h4}>
-                        {formatData(user?.registered_on)}
+                        {formatDate(user?.registered_on)}
```
portante, line 218: Do you have a way to automatically clean trailing whitespace in you editor?

dashboard/src/actions/profileActions.js
```diff
@@ -15,18 +15,17 @@ export const getProfileDetails = () => async (dispatch) => {
...
-      let toast = {
-        variant: "danger",
-        title: response?.data?.message,
-      };
-      dispatch({
-        type: TYPES.SHOW_TOAST,
-        payload: { ...toast },
-      });
-    }
+    }
```
portante, line 18: Are we sure we won't get a status of 200 or a response length of zero without an error being thrown?

dashboard/src/actions/profileActions.js
```diff
@@ -35,31 +34,29 @@ export const getProfileDetails = () => async (dispatch) => {
...
-        params[key] = userDetails[key];
-      }
-    }
+    const updatedUserDetails = { ...getState().userProfile.updatedUserDetails };
+        
     const username = getState().userAuth.loginDetails.username;
     if (username) {
       const response = await API.put(`${API_ROUTES.GET_USER}/${username}`, {
-        ...params,
+        ...updatedUserDetails,
```
portante, line 59: We also need to check that `updatedUserDetails` is not empty.

dashboard/src/reducers/userProfileReducer.js
```diff
@@ -1,8 +1,15 @@
...
+  UPDATE_USER_DETAILS,
+  RESET_DATA,
+  SET_USER_DETAILS,
+} from "../actions/types";
 
 const initialState = {
   userDetails: {},
   userDetails_copy: {},
+  updatedUserDetails: {},
+  isUserDetailsUpdated: false,
```
portante, line 12: Where is this used?

portante (changes requested):

dashboard/src/reducers/userProfileReducer.js
MVarshini: It is used in ProfileComponent/index.jsx to disable/enable the `Save` button.
By checking `updatedUserDetails` is empty the same behaviour can be achieved but it would be a additional load.

dashboard/src/actions/profileActions.js
MVarshini: Only if `updatedUserDetails` not empty, `Save` button will be enabled and request will be sent. Else-clause should be handled

dashboard/src/modules/components/ProfileComponent/index.jsx
MVarshini: Yes, white space is removed.

dashboard/src/utils/helper.js
```diff
@@ -22,3 +22,6 @@ export const validateEmail = (email) => {
   };
 };
 
+export const isValidDate = (d) => {
+  return d instanceof Date && !isNaN(d);
+}
```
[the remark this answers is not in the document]
MVarshini: I was unaware of the `isValidDate` method in `@patternfly/react_core`. 

"Invalid date object" (new Date('foo'))  is an `instanceof Date` so checking for `!isNaN` :grinning:

MVarshini:
> This is looking good!
> 
> My main comment would be that we're adding a lot of code to the dashboard without unit tests, even though you'd created a basic unit test framework and "test" command.
> 
> In the agent and server we try to maintain a rule that anything new or changed needs to be tested. We're definitely not perfect, and we don't have real functional testing nor anywhere near complete unit test coverage. Still, during this early rapid evolution of the new dashboard we can either build a solid testing framework or build up a substantial technical debt to resolve later.
> 
> At this point, I'm just "throwing this out" for consideration; but maybe after this and Shubham's browsing code goes in, it would be a good idea to spend some time writing unit tests for the existing code before going further?

Sure, let's start writing the unit tests!

dashboard/src/actions/profileActions.js
```diff
@@ -0,0 +1,107 @@
+import * as TYPES from "./types";
+import API from "../utils/axiosInstance";
+import * as API_ROUTES from "../utils/apiConstants";
+import Cookies from "js-cookie";
+import { showFailureToast } from "./toastActions";
+
+export const getProfileDetails = () => async (dispatch) => {
+  try {
+    dispatch({ type: TYPES.LOADING });
+    const username = Cookies.get("username") ? Cookies.get("username") : null;
```
dbutenhof, line 10: Is it necessary to call `Cookies.get` twice here? According to the documentation, it seems to just return `novalue` if the cookie doesn't exist.

Also, I suspect that if we don't have a username here, the attempt to show a profile should just fail; the `/api/v1/user/null` isn't going to do anything useful anyway.

dashboard/src/utils/helper.js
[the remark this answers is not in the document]
dbutenhof: `new Date('foo')` is a `Date` object that's also a NaN?? Wow. I wonder what the React `isValidDate` actually does?

dbutenhof (commented):
I'm a little concerned about the behavior of `getProfileDetails`, though the "bad cookie" behavior may be an edge case that we're unlikely to hit. Aside from that, this is looking really good.

dashboard/src/reducers/userProfileReducer.js
```diff
@@ -0,0 +1,47 @@
...
+};
+
+const UserProfileReducer = (state = initialState, action = {}) => {
+  const { type, payload } = action;
+  switch (type) {
+    case GET_USER_DETAILS:
+      return {
+        ...state,
+        userDetails: Object.assign({}, payload),
+        userDetails_copy: Object.assign({}, payload),
```
webbnh, line 22: What is the purpose of `userDetails_copy`?  (When would it ever be different from `userDetails`?  Or, is it intended for (ab)use by the caller?)

dashboard/src/utils/axiosInstance.js
```diff
@@ -0,0 +1,41 @@
...
+    Authorization: `Bearer ${token}`,
+  },
+});
+
+axiosInstance.interceptors.request.use(async (req) => {
+  token = Cookies.get("token") ? Cookies.get("token") : null;
+  req.headers.Authorization = `Bearer ${token}`;
+
+  return req;
+});
```
webbnh, line 24: Once again, if `Cookies.get("token")` returns `undefined`, what should `token` (or `req.headers.Authorization`) be?  `null` seems like an odd choice, given that it will stringize as `'null'`.  It seems like, if there's no token value, then we should be omitting the `Authorization` property altogether.  No?

dashboard/src/utils/axiosInstance.js
```diff
@@ -0,0 +1,41 @@
...
+/** Intercept any unauthorized request.
+ * dispatch logout action accordingly **/
+const UNAUTHORIZED = 401;
```
webbnh, line 27: It seems weird that we would roll our own definition for this...isn't there a standard one that we could import from somewhere? (I found [http-status-codes](https://www.npmjs.com/package/http-status-codes)...I have no idea whether that's a good source or not....)

dashboard/src/utils/axiosInstance.js
```diff
@@ -0,0 +1,41 @@
...
+axiosInstance.interceptors.response.use(
+  (response) => response,
+  (error) => {
+    if (error?.response) {
+      const { status } = error.response;
+      if (status === UNAUTHORIZED) {
+        dispatch(showSessionExpired());
+      }
```
webbnh, line 35: What do/should we do if the status is something other than `UNAUTHORIZED`?  Really nothing?

dashboard/src/utils/axiosInstance.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+    return Promise.reject(error);
```
webbnh, line 37: OK, we reject the `Promise` if there is an error...we don't have to do anything (explicit) if there is no error?

webbnh (approved):
Looks good.

I have a pile of nits, suggestions, and pointed questions, but there's a lot of code here to poke at.  😇

webbnh:
Also, now that Shubham's PR is in, you have need to rebase your branch on the end of `main` and resolve the conflicts.

dashboard/src/App.js
```diff
@@ -25,7 +25,7 @@ function App() {
         <Routes>
           <Route path="/" element={<MainLayout />}>
             <Route path="user-profile" element={<ProfileComponent />} />
-          </Route>           
+          </Route>
```
webbnh, line 28: You should remove this trailing whitespace.

dashboard/src/modules/containers/MainLayout/index.jsx
```diff
@@ -4,6 +4,10 @@ import ToastComponent from "modules/components/ToastNotificationComponent";
 import { useSelector } from "react-redux";
 import { Spinner, Page, PageSection } from "@patternfly/react-core";
 import HeaderComponent from "modules/components/HeaderComponent";
+import NavbarDrawer from '../../components/NavbarDrawerComponent';
+import Sidebar from '../../components/SidebarComponent';
+import { TableWithFavorite } from 'modules/components/TableComponent';
```
webbnh, line 9: Does Prettier make the choice of quotation marks on the `import` statements uniform?

Also, do we still need the `../../` in lines 7 & 8, or can we start the path with `modules/`, like we do in lines 6 & 9?

dashboard/src/modules/containers/MainLayout/index.jsx
```diff
@@ -1,28 +1,30 @@
-import React from 'react';
+import React from "react";
+import { Outlet } from "react-router-dom";
 import ToastComponent from "modules/components/ToastNotificationComponent";
 import { useSelector } from "react-redux";
-import { Page, Spinner } from '@patternfly/react-core';
+import { Spinner, Page, PageSection } from "@patternfly/react-core";
+import HeaderComponent from "modules/components/HeaderComponent";
 import NavbarDrawer from '../../components/NavbarDrawerComponent';
```
webbnh, line 7: The LGTM bot says you aren't using the `NavbarDrawer`, here.  (Did something get dropped during the merge?)

webbnh (commented):
A few quick comments on your latest commit.

dashboard/src/modules/containers/MainLayout/index.jsx
MVarshini: Done!

dashboard/src/modules/containers/MainLayout/index.jsx
MVarshini: `NavbarDrawer` has been changed to `HeaderComponent`

dashboard/src/actions/toastActions.js
```diff
@@ -25,3 +25,11 @@ export const showFailureToast = () => async (dispatch) => {
     payload: { ...toast },
   });
 }
+
+export const constructToast = (variant, title, message ="") => async(dispatch) => {
+  let toast = { variant, title, message };
+  dispatch({
+    type: TYPES.SHOW_TOAST,
+    payload: { ...toast },
+  });
```
webbnh, line 34: Why bother creating `toast` and then unpacking it again?...why not just use the values in-line:
```
  dispatch({
    type: TYPES.SHOW_TOAST,
    payload: { variant, title, message },
  });
```

dashboard/src/actions/profileActions.js
```diff
@@ -62,32 +54,22 @@ export const sendForUpdate = () => async (dispatch, getState) => {
         ...updatedUserDetails,
       });
       if (response.status === 200) {
-        let toast = {
-          variant: "success",
-          title: "Updated!",
-        };
+        dispatch(constructToast("success", "Updated!"));
```
webbnh, line 57: What does it mean to invoke `dispatch()` on the result of `constructToast()` which itself invokes `dispatch()`?  (That is, why do we call `dispatch()` here _and_ inside `constructToast()`?  And...is the call inside `constructToast()` asynchronous?...in which case, what value are we passing to `dispatch()` here?)

dashboard/src/modules/components/HeaderComponent/index.jsx
```diff
@@ -29,14 +29,16 @@ import {
...
 const HeaderToolbar = () => {
   const dispatch = useDispatch();
   const loginDetails = useSelector((state) => state.userAuth.loginDetails);
+  const [isDropdownOpen, setIsDropdownOpen] = useState(false);
   const navigate = useNavigate();
-  let { pathname } = useLocation();
+  const { pathname } = useLocation();
   const isLoggedIn = Cookies.get("isLoggedIn");
-
+
```
webbnh, line 41: Spurious blank characters.

dashboard/src/utils/axiosInstance.js
```diff
@@ -17,7 +17,7 @@ const axiosInstance = axios.create({
 });
 
 axiosInstance.interceptors.request.use(async (req) => {
-  token = Cookies.get("token") ? Cookies.get("token") : null;
+  token = Cookies.get("token");
   req.headers.Authorization = `Bearer ${token}`;
```
webbnh, line 21: If `token` is false-y (e.g., `undefined`, `null`, or `""`), shouldn't we be omitting the request `Authorization` header?  (Here and at line 15?)

webbnh (commented):
Some questions/comments on your update.

dashboard/src/actions/profileActions.js
```diff
@@ -0,0 +1,86 @@
...
+
+export const resetUserDetails = () => async (dispatch, getState) => {
+  const userDetails_copy = { ...getState().userProfile.userDetails_copy };
+
+  dispatch({
+    type: TYPES.SET_USER_DETAILS,
+    payload: { ...userDetails_copy },
+  });
+  dispatch({ type: TYPES.RESET_DATA });
+};
```
webbnh, line 86: `userDetails_copy` is a copy of the object returned by `getState().userProfile.userDetails_copy` (which is itself, a copy, I think); we then make a copy of that copy for the `payload` attribute of the object that we pass the first call to `dispatch()`.  Do we really need to make all these copies?  That is, would this work?
```
export const resetUserDetails = () => async (dispatch, getState) => {
 dispatch({
    type: TYPES.SET_USER_DETAILS,
    payload: getState().userProfile.userDetails_copy,
  });
  dispatch({ type: TYPES.RESET_DATA });
};
```
That is, instead of making `payload` refer to a temporary object, make it refer to the original object.

dashboard/src/utils/axiosInstance.js
MVarshini: If there is no error, response is being returned.

dashboard/src/utils/axiosInstance.js
MVarshini: Added the check

dashboard/src/actions/profileActions.js
MVarshini: Done!

dashboard/src/utils/axiosInstance.js
MVarshini: Promise is rejected with error and it gets handled in the corresponding functions `catch` block

dashboard/src/actions/authActions.js
```diff
@@ -65,58 +91,83 @@ export const setUserLoggedInState = (value) => async (dispatch) => {
...
+export const logout = () => async (dispatch) => {
+  dispatch({ type: TYPES.LOADING });
+  let keys = ['username', 'token', 'isLoggedIn'];
+  for(let key of keys) {
+    Cookies.remove(key);
+  }
+  dispatch({ type: TYPES.COMPLETED });
+  setTimeout(() => {
+    window.location.href = "login";
+  }, 2000);
```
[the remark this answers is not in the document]
MVarshini: Done!

dashboard/src/actions/authActions.js
```diff
@@ -1,53 +1,79 @@
...
+          navigate("/");
+          let toast = {
+            variant: "success",
+            title: "Logged in successfully",
+          };
+          dispatch({
+            type: TYPES.SHOW_TOAST,
+            payload: { ...toast },
+          });
+        }
```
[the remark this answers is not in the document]
MVarshini: `await` to set the login details and proceed.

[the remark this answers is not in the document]
MVarshini: `await` to set the login details and proceed.

dashboard/src/actions/authActions.js
```diff
@@ -1,53 +1,79 @@
...
+      dispatch({
+        type: TYPES.USER_NOTION_ALERTS,
+        payload: [],
+      });
+      const response = await API.post(API_ROUTES.LOGIN, {
+        ...details,
+      });
+      if (response.status === 200 && Object.keys(response.data).length > 0) {
+        let keepUser = getState().userAuth.keepLoggedIn;
+        let expiryTime = keepUser ? 7 : 0.5;
```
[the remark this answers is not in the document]
MVarshini: Done!

dbutenhof (changes requested):
Looking good. You should do a `git rebase -i` and edit your first git log message to the standard format

```
User Profile Page

PBENCH-197

Add user profile edit page and authorization infrastructure.
```

(Or something to that effect; the body of the log message should give future readers of the git history some reasonable idea what you did.)

dashboard/src/modules/components/HeaderComponent/index.jsx
```diff
@@ -41,17 +41,15 @@ const HeaderToolbar = () => {
...
-    setIsDropdownOpen(false);
+    const menuOptions = {
+      profile: function () {
+        return navigate("/user-profile");
+      },
+      logout: function () {
+        return dispatch(logout());
+      },
+    };
+    return menuOptions[type]() ?? setIsDropdownOpen(false);
```
webbnh, line 52: Using a mapping instead of a `switch` statement is clever, but, ultimately, I think it's making things more complicated.

For instance, if `type` doesn't match one of the entries, then `menuOptions[type]` will be `undefined` and then the code will try to _call_ that (which I think will produce an exception).  It looks like you've attempted to address that via the `??` operator, but it would need to be applied before the `()` operator.  (And, I cannot tell whether you mean to invoke `setIsDropdownOpen()` in all cases or in only the situation where there is no hit in the mapping.)

Stylistically, I'm curious why you opted to use the `function` keyword instead of the "arrow" to produce a callable value.  I would think we would code this something like:
```
    const menuOptions = {
      profile: () => navigate("/user-profile"),
      logout: () => dispatch(logout()),
    };
    action = menuOptions[type];
    if (action) {
      action();
    }
    setIsDropdownOpen(false);
```

dashboard/src/utils/axiosInstance.js
```diff
@@ -11,15 +11,13 @@ let token = Cookies.get("token");
 const axiosInstance = axios.create({
   baseURL: LOCAL_BASE_URL,
   responseType: "json",
-  headers: {
-    Authorization: `Bearer ${token}`,
-  },
```
webbnh: With this removed, do you still want to be fetching the token at line 9, or should that be removed, too?

dashboard/src/utils/axiosInstance.js
```diff
@@ -30,7 +28,8 @@ axiosInstance.interceptors.response.use(
   (error) => {
     if (error?.response) {
       const { status } = error.response;
-      if (status === UNAUTHORIZED) {
+      const { responseURL } = error.request;
+      if (status === UNAUTHORIZED && !responseURL.includes("login")) {
         dispatch(showSessionExpired());
       }
```
webbnh, line 34: So, _any_ response of "unauthorized" from a URI which doesn't include `login` will be deemed to indicate an expired session?  That might be true, if the Dashboard is highly disciplined, but I'm having trouble imagining how we'll cover all the use cases without ever violating that idea.

That is, I expect that there will be cases where a given user can request a particular result or view, and, for various reasons, the server will respond with `UNAUTHORIZED`...not because the user's session has expired but rather because the requesting user is specifically not authorized to access the target resource (e.g., because the data belongs to another user who has not published it).

So, I think we're going to find that this criterion is insufficient.

webbnh (changes requested):
This PR now has merge conflicts with `main`.  Also,
- It looks like the code needs to be submitted to Prettier before it is merged.
- Could you please confirm the Patternfly version dependency?
- I've made another attempt at explaining my concern around the `constructToast()` construction.
- Replacing the `switch` statement with a mapping doesn't look quite right.
- The detection of a session expiration looks like it's not going to serve us well.
- And, another odd nit or two.

dashboard/src/utils/axiosInstance.js
webbnh: As clarified in today's Dashboard Sync meeting, we expect that `FORBIDDEN` will be returned for access failures, while `UNAUTHORIZED` will be reserved for token problems, so this code should be OK as is (however, Dave suggested replacing the URL test with a different criterion tied to whether the request had specified a token).

webbnh (dismissed):
@MVarshini, it looks like you've resolved the merge conflicts with `main`, but, unfortunately, you did so by merging `main` into your branch, rather than by rebasing your branch to the end of `main`.  When all is said and done, I'm hopeful that this won't matter, but, in general, we prefer to maintain a linear history in Git, and so `rebase` is generally preferred over `merge`.  (Also, since you had to rebase anyway to amend the Git log for the initial commit, you could have killed two birds with that one stone.)

Speaking of the initial commit log message, while it now has the stuff in it that Dave requested, it also has a bunch of other stuff in it.  (It looks like about 10 commits got squashed into one and all of their (rather uninteresting) commit logs got rolled up into it; and, there are a couple of stray quotation marks.)  Don't worry about it -- we can fix it when we merge this PR into `main` -- but just be cognizant of the result for the next time you need to amend a log message.

In any case, I think we're nearly done here.  There are two open conversations from @portante, which can probably be resolved now, and there are two from me.  There's the `constructToast()` thunk, which is probably just a misapprehension on my part (especially if it's actually working!  😁), and there's the `UNAUTHORIZED` test, which I'd like confirmation from @dbutenhof on whether to fix in this PR or another one.

dashboard/src/utils/axiosInstance.js
webbnh: @dbutenhof, I'm going to leave it to you to resolve this conversation:  you mentioned that the criterion in the code should be changed; I don't know whether you were hoping to see that in this PR or not.

portante (dismissed):

dbutenhof (dismissed):
One question/conversation from Webb to close up, and we're ready to go.

dbutenhof (dismissed):

dashboard/src/utils/axiosInstance.js
```diff
@@ -0,0 +1,36 @@
+import axios from "axios";
+import Cookies from "js-cookie";
+import { showSessionExpired } from "actions/toastActions";
+import store from "store/store";
+
+const { dispatch } = store;
+
+const axiosInstance = axios.create({
+  responseType: "json",
+});
```
webbnh, line 10: If you remove the comma from the end of line 9, will Prettier put this all on one line for us?
```
const axiosInstance = axios.create({ responseType: "json" });
```

webbnh (approved):
I'm not sure what happened, but you recently force-pushed your branch, and now I've lost the sequence of changes since the start of this review...so, I just reviewed it all again...and, naturally, I found stuff.

For the most part, it's small things -- unneeded object copies, lists of things which it would be nice to have ordered, a function with inconsistent return values, and some places where we could tighten up the code a bit.  None of these need to gate this PR (although the function return values thing is annoying...), but it would be good to have them touched up "soon".  If you want to do them in this PR, that would be fine with me (although, I think Dave is itching to have this merged); if you want to put up another PR with them, that would be OK (especially if that PR included unit tests 😁); I just don't want them to get lost....

So, let me know whether you want to address these here, and, if not, I'll mark them resolved.

pravins:
Hi,
  Already 1+ month for this PR, Any chance can we open new issue with unresolved comments and merge this PR?

webbnh:
> Any chance can we open new issue with unresolved comments and merge this PR?

Hi, @pravins, does my last review summary address your question?  I said:

>If you want to do them in this PR, that would be fine with me (although, I think Dave is itching to have this merged); if you want to put up another PR with them, that would be OK (especially if that PR included unit tests 😁); I just don't want them to get lost....
>
> So, let me know whether you want to address these here, and, if not, I'll mark them resolved.

pravins:
@webbnh Thanks, Yes that comment helps. :+1:  
@MVarshini i think, you can go ahead. Lets have layered PR's :) Please open new issue with open comments.

dashboard/src/modules/containers/MainLayout/index.jsx
```diff
@@ -1,20 +1,48 @@
...
 import ToastComponent from "modules/components/ToastNotificationComponent";
 import { useSelector } from "react-redux";
-import { Page, Spinner } from "@patternfly/react-core";
-import NavbarDrawer from "modules/components/NavbarDrawerComponent";
+import HeaderComponent from "modules/components/HeaderComponent";
+import {
+  Spinner,
+  Page,
+  PageSection,
+  PageSectionVariants,
```
portante, line 10: If we are going to alphabetize in other modules, we should alphabetize here.  I'd suggest making such an alphabetization change in a separate PR.

dashboard/src/modules/containers/MainLayout/index.jsx
```diff
@@ -1,20 +1,48 @@
...
+} from "@patternfly/react-core";
 import Sidebar from "modules/components/SidebarComponent";
 import TableWithFavorite from "modules/components/TableComponent";
+import "./index.less";
+
+const LoginView = () => {
+  const { pathname } = useLocation();
+  return (
+    <>
+      {pathname === "/" ? (
```
portante, line 20: Can you help me understand what `useLocation()` returns here?  It seems this might be a bit fragile, and we should have a "is logged in" function that should be used instead of the return value from `useLocation()`.

portante (changes requested):
Sorry for the late review.  I am just curious about the `useLocation()` to determine the "logged in" state, when the `Cookies` is consulted in other locations.

dashboard/src/modules/components/HeaderComponent/index.jsx
```diff
@@ -0,0 +1,154 @@
...
+    };
+    const action = menuOptions[type];
+    if (action) {
+      action();
+    }
+    setIsDropdownOpen(false);
+  };
+  const onDropdownToggle = () => {
+    setIsDropdownOpen(!isDropdownOpen);
+  };
```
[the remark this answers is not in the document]
MVarshini: I don't think it's necessary.

dashboard/src/actions/profileActions.js
```diff
@@ -0,0 +1,86 @@
...
+  try {
+    dispatch({ type: TYPES.LOADING });
+
+    const username = getState().userAuth.loginDetails.username;
+    const endpoints = getState().apiEndpoint.endpoints;
+
+    if (username) {
+      const response = await API.put(`${endpoints?.api?.user}/${username}`, {
+        ...getState().userProfile.updatedUserDetails,
+      });
```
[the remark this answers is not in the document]
MVarshini: If not, it will be an error here

dashboard/src/modules/containers/MainLayout/index.jsx
MVarshini: The `useLocation ` hook is a function that returns the location object that contains information about the current URL. Whenever the URL changes, a new location object will be returned.

Here, if the user is not logged in  and the URL is "/" it is a browsing page for other URLs login prompt message is shown.

At line:44, `is logged in` is checked and `LoginView` is rendered

dashboard/src/modules/containers/MainLayout/index.jsx
MVarshini: Let's take it separately

dashboard/src/utils/axiosInstance.js
MVarshini: Done!

dashboard/src/modules/components/TableComponent/index.jsx
```diff
@@ -180,54 +183,68 @@ const TableWithFavorite = () => {
...
-          page={page}
-          setPage={setPage}
-          perPage={perPage}
-          setPerPage={setPerPage}
-        />
-      </PageSection>
+        <div
+          style={{
+            height: "200px",
+          }}
```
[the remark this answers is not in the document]
MVarshini: Done

dashboard/src/modules/components/ProfileComponent/index.jsx
```diff
@@ -0,0 +1,232 @@
...
+  TextVariants,
+  Grid,
+  GridItem,
+  Card,
+  Level,
+  LevelItem,
+  Button,
+  TextInput,
+  CardBody,
+  isValidDate,
```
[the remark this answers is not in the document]
MVarshini: Done. For other files, I'll be working and raising different PR

dashboard/src/actions/profileActions.js
```diff
@@ -0,0 +1,86 @@
...
+  }
+};
+
+export const resetUserDetails = () => async (dispatch, getState) => {
+  dispatch({
+    type: TYPES.SET_USER_DETAILS,
+    payload: getState().userProfile.userDetails_copy,
+  });
+  dispatch({ type: TYPES.RESET_DATA });
+};
```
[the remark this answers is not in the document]
MVarshini: Previously, I was dispatching `SHOW_TOAST` with the message. As it increases the number of lines I wrote `constructToast`

dashboard/src/actions/types.js
```diff
@@ -24,3 +25,10 @@ export const NAVBAR_CLOSE = "NAVBAR_CLOSE";
 export const GET_PUBLIC_DATASETS = "GET_PUBLIC_DATASETS";
 export const FAVORITED_DATASETS = "GET_FAVORITE_DATASETS";
 export const UPDATE_PUBLIC_DATASETS = "UPDATE_PUBLIC_DATASETS";
+export const SET_LOGIN_DETAILS = "SET_LOGIN_DETAILS";
+
+/* USER DETAILS*/
```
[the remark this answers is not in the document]
MVarshini: Done!

dashboard/src/actions/toastActions.js
```diff
@@ -0,0 +1,36 @@
...
+
+export const showSessionExpired = () => async (dispatch) => {
+  const toast = {
+    variant: "danger",
+    title: "Session Expired",
+    message: "Please login to continue",
+  };
+  dispatch({
+    type: TYPES.SHOW_TOAST,
+    payload: { ...toast },
```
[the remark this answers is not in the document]
MVarshini: Done!

webbnh (approved):
GTG!

dbutenhof (approved):
@portante still has open conversations, but I think we're one review short of a merge here.

dashboard/src/modules/components/TableComponent/index.jsx
```diff
@@ -134,113 +135,119 @@ const TableWithFavorite = () => {
...
-                        <Td
-                          favorites={{
-                            isFavorited: isRepoFavorited(repo),
-                            onFavorite: (_event, isFavoriting) => {
-                              markRepoFavorited(repo, isFavoriting);
-                            },
-                            rowIndex,
-                          }}
-                        />
+      {!loginDetails?.isLoggedIn && (
```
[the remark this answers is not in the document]
MVarshini: If the user is logged in, Overview page should be rendered which is under development.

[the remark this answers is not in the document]
dbutenhof: Shouldn't that be handled by conditional *routing* to a different component rather than conditionalizing the `TableComponent`?

In general, we want *page views* to work regardless of whether you're logged in; it's just that if you're not, you can only see "public" datasets, and can't change anything. The "Overview" page (and as you implemented elsewhere the Profile view) is an exception as they have no meaning if you're not logged in. (Although a user who hasn't logged in should probably be able to select the overview page and get side-tracked to login before showing anything.)

In particular, a user who logs in should still be able to browse other accessible data, not just "curate" their own data, and there's nothing about this page that shouldn't work in that environment, right?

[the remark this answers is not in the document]
MVarshini: Let's have the separate route for the "public" datasets which the user can view irrespective of being logged in. 
Let's take that separately. 

Tracked in [PBENCH-757](https://issues.redhat.com/browse/PBENCH-757)

dbutenhof (approved):
Even though I've written a few extended comments, I'll resolve those conversations and I'd be happy to get this in and move on. (Particularly since it fixes the dashboard for the `/datasets/list` pagination change... which, at this point, I won't even suggest ought to have been a separate PR that could have gone in quickly. 👀 😆)

dashboard/src/modules/components/TableComponent/index.jsx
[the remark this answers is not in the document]
webbnh: I suspect that this added conditional is not accomplishing anything.  (That is, if the user is not logged in, then I think we get redirected at the app level and never end up doing much here.)  So, I'm not sure that it's worth adding the extra complication here.

Nevertheless, this function is really, really hard to read.  I'm not saying that we should fix it in this PR, but this function should be broken up with each of the conditional pieces relegated to a subroutine or other external definition, so that the structure is much more clear -- "if this condition is true, then we load that; otherwise, we load this other" -- having the tags in-line hides the structure.  (Ternaries are good when it's easy to see the alternate values, e.g., when they fit on one line or so; at this scale, they are evil. 😈)

webbnh (approved):
👍

dashboard/src/modules/components/TableComponent/index.jsx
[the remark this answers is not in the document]
dbutenhof: I'm not entirely comfortable with this, but I'm willing to defer it for another PR. @webbnh are you ready for me to resolve this, or is this something you'd like to push further? 😄

dashboard/src/modules/containers/MainLayout/index.jsx
portante: See https://github.com/distributed-system-analysis/pbench/pull/2817#pullrequestreview-1007673357.

portante (approved):

PR #2820, opened 2022-05-04
New API to fetch date range of accessible datasets

dbutenhof:
PBENCH-643

Add a new `datasets/daterange` API to return the earliest and latest creation date of datasets readable by the authenticated user and with specified owner and access filters.

### DIFF
--- a/lib/pbench/server/api/resources/datasets_daterange.py
+++ b/lib/pbench/server/api/resources/datasets_daterange.py
@@ -1,0 +1,83 @@
+from logging import Logger
+import logging
+
+from flask.json import jsonify
+from flask.wrappers import Request, Response
+from sqlalchemy import func
+
+from pbench.server import PbenchServerConfig
+from pbench.server.api.resources import (
+    ApiBase,
+    API_OPERATION,
+    Parameter,
+    ParamType,
+    Schema,
+)
+from pbench.server.database.database import Database
+from pbench.server.database.models.datasets import Dataset
+
+
+class DatasetsDaterange(ApiBase):
+    """
+    API class to retrieve the available date range of accessible datasets.
+    """
+
+    GET_SCHEMA = Schema(
+        Parameter("owner", ParamType.USER, required=False),
+        Parameter("access", ParamType.ACCESS, required=False),
+    )
+
+    def __init__(self, config: PbenchServerConfig, logger: Logger):
+        super().__init__(
+            config,
+            logger,
+            None,
+            role=API_OPERATION.READ,
+        )
+
+    def _get(self, _, request: Request) -> Response:
+        """
+        Get the date range for which datasets are available to the client based
+        on authentication plus optional dataset owner and access criteria.
+
+        Args:
+            json_data: For a GET, this contains only the Flask URI template
+                values, which aren't used here
+            request: The original Request object containing query parameters
+
+
+        GET /api/v1/datasets/daterange?owner=user&access=public
+        """
+
+        json = self._collect_query_params(request, self.GET_SCHEMA)
+        new_json = self.GET_SCHEMA.validate(json) if json else json
+
+        access = new_json.get("access")
+        owner = json.get("owner")
+        self.logger.info("Getting date range for user {!r}, access {!r}", owner, access)
+
+        # Validate the authenticated user's authorization for the combination
+        # of "owner" and "access".
+        self._check_authorization(owner, access)
+
+        # Build a SQLAlchemy Query object expressing all of our constraints
+        query = Database.db_session.query(
+            func.min(Dataset.created), func.max(Dataset.created)
+        )
+        query = self._build_sql_query(new_json, query)
+
+        # Useful for debugging, but verbose: this displays the fully expanded
+        # SQL `SELECT` statement.
+        if self.logger.isEnabledFor(logging.DEBUG):
+            self.logger.debug(
+                "QUERY {}",
+                query.statement.compile(compile_kwargs={"literal_binds": True}),
+            )
+
+        # Execute the query, returning a tuple of the 'min' date and the
+        # 'max' date.
+        results = query.first()
+
+        self.logger.info("Results: {!r}", results)
+
+        return jsonify({"from": results[0].isoformat(), "to": results[1].isoformat()})
--- a/lib/pbench/server/api/resources/datasets_daterange.py
+++ b/lib/pbench/server/api/resources/datasets_daterange.py
@@ -3,7 +3,7 @@ from flask.wrappers import Request, Response
 from logging import Logger
 from sqlalchemy import func
 
-from pbench.server import PbenchServerConfig
+from pbench.server import JSON, PbenchServerConfig
 from pbench.server.api.resources import (
     ApiBase,
     API_OPERATION,
@@ -33,12 +33,13 @@ class DatasetsDateRange(ApiBase):
             role=API_OPERATION.READ,
         )
 
-    def _get(self, _, request: Request) -> Response:
+    def _get(self, json_data: JSON, request: Request) -> Response:
         """
         Get the date range for which datasets are available to the client based
         on authentication plus optional dataset owner and access criteria.
 
         Args:
+            json_data: Ignored because GET has no JSON payload
             request: The original Request object containing query parameters
 
         GET /api/v1/datasets/daterange?owner=user&access=public
--- a/lib/pbench/server/api/resources/datasets_daterange.py
+++ b/lib/pbench/server/api/resources/datasets_daterange.py
@@ -1,8 +1,6 @@
-from logging import Logger
-import logging
-
 from flask.json import jsonify
 from flask.wrappers import Request, Response
+from logging import Logger
 from sqlalchemy import func
 
 from pbench.server import PbenchServerConfig
@@ -17,7 +15,7 @@ from pbench.server.database.database import Database
 from pbench.server.database.models.datasets import Dataset
 
 
-class DatasetsDaterange(ApiBase):
+class DatasetsDateRange(ApiBase):
     """
     API class to retrieve the available date range of accessible datasets.
     """
@@ -41,20 +39,14 @@ class DatasetsDaterange(ApiBase):
         on authentication plus optional dataset owner and access criteria.
 
         Args:
-            json_data: For a GET, this contains only the Flask URI template
-                values, which aren't used here
             request: The original Request object containing query parameters
-
 
         GET /api/v1/datasets/daterange?owner=user&access=public
         """
 
-        json = self._collect_query_params(request, self.GET_SCHEMA)
-        new_json = self.GET_SCHEMA.validate(json) if json else json
-
-        access = new_json.get("access")
+        json = self._validate_query_params(request, self.GET_SCHEMA)
+        access = json.get("access")
         owner = json.get("owner")
-        self.logger.info("Getting date range for user {!r}, access {!r}", owner, access)
 
         # Validate the authenticated user's authorization for the combination
         # of "owner" and "access".
@@ -64,20 +56,10 @@ class DatasetsDaterange(ApiBase):
         query = Database.db_session.query(
             func.min(Dataset.created), func.max(Dataset.created)
         )
-        query = self._build_sql_query(new_json, query)
-
-        # Useful for debugging, but verbose: this displays the fully expanded
-        # SQL `SELECT` statement.
-        if self.logger.isEnabledFor(logging.DEBUG):
-            self.logger.debug(
-                "QUERY {}",
-                query.statement.compile(compile_kwargs={"literal_binds": True}),
-            )
+        query = self._build_sql_query(json, query)
 
         # Execute the query, returning a tuple of the 'min' date and the
         # 'max' date.
         results = query.first()
-
-        self.logger.info("Results: {!r}", results)
 
         return jsonify({"from": results[0].isoformat(), "to": results[1].isoformat()})
--- a/lib/pbench/server/api/resources/datasets_list.py
+++ b/lib/pbench/server/api/resources/datasets_list.py
@@ -53,8 +53,6 @@ class DatasetsList(ApiBase):
         desired metadata keys; instead we rely on URI query parameters.
 
         Args:
-            json_data: For a GET, this contains only the Flask URI template
-                values, which aren't used here
             request: The original Request object containing query parameters
 
         GET /api/v1/datasets/list?start=1970-01-01&end=2040-12-31&owner=fred&metadata=dashboard.seen,server.deletion
@@ -63,33 +61,32 @@ class DatasetsList(ApiBase):
         # We missed automatic schema validation due to the lack of a JSON body;
         # construct an equivalent JSON body now so we can run it through the
         # validator.
-        json = self._collect_query_params(request, self.schema)
-        new_json = self.schema.validate(json) if json else json
+        json = self._validate_query_params(request, self.schema)
 
         # Validate the authenticated user's authorization for the combination
         # of "owner" and "access".
-        self._check_authorization(json.get("owner"), new_json.get("access"))
+        self._check_authorization(json.get("owner"), json.get("access"))
 
         # Build a SQLAlchemy Query object expressing all of our constraints
         query = Database.db_session.query(Dataset)
-        if "start" in new_json and "end" in new_json:
+        if "start" in json and "end" in json:
             self.logger.info("Adding start / end query")
             query = query.filter(
-                Dataset.created.between(new_json["start"], new_json["end"])
+                Dataset.created.between(json["start"], json["end"])
             )
-        elif "start" in new_json:
+        elif "start" in json:
             self.logger.info("Adding start query")
-            query = query.filter(Dataset.created >= new_json["start"])
-        elif "end" in new_json:
+            query = query.filter(Dataset.created >= json["start"])
+        elif "end" in json:
             self.logger.info("Adding end query")
-            query = query.filter(Dataset.created <= new_json["end"])
-        if "name" in new_json:
+            query = query.filter(Dataset.created <= json["end"])
+        if "name" in json:
             self.logger.info("Adding name query")
-            query = query.filter(Dataset.name.contains(new_json["name"]))
-        if "controller" in new_json:
+            query = query.filter(Dataset.name.contains(json["name"]))
+        if "controller" in json:
             self.logger.info("Adding controller query")
-            query = query.filter(Dataset.controller == new_json["controller"])
-        query = self._build_sql_query(new_json, query)
+            query = query.filter(Dataset.controller == json["controller"])
+        query = self._build_sql_query(json, query)
 
         # Useful for debugging, but verbose: this displays the fully expanded
         # SQL `SELECT` statement.
@@ -103,7 +100,7 @@ class DatasetsList(ApiBase):
         # consistent and reproducible output to compare.
         results = query.order_by(Dataset.name).all()
 
-        keys = new_json.get("metadata")
+        keys = json.get("metadata")
 
         response = []
         for dataset in results:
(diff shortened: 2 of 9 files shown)


lib/pbench/server/api/resources/datasets_daterange.py
```diff
@@ -0,0 +1,83 @@
+from logging import Logger
+import logging
```
dbutenhof, line 2: Oops.

lib/pbench/server/api/resources/datasets_daterange.py
```diff
@@ -0,0 +1,83 @@
...
+        )
+        query = self._build_sql_query(new_json, query)
+
+        # Useful for debugging, but verbose: this displays the fully expanded
+        # SQL `SELECT` statement.
+        if self.logger.isEnabledFor(logging.DEBUG):
+            self.logger.debug(
+                "QUERY {}",
+                query.statement.compile(compile_kwargs={"literal_binds": True}),
+            )
```
dbutenhof, line 75: I originally left this in `datasets/list` because I didn't want to lose the incantation. I duplicated it only because I needed to check after it initially didn't seem to be working. But I should either delete it here or move it to a base class method. Not sure either is worth doing in this PR, though.

lib/pbench/server/api/resources/datasets_daterange.py
```diff
@@ -0,0 +1,83 @@
...
+
+        # Execute the query, returning a tuple of the 'min' date and the
+        # 'max' date.
+        results = query.first()
+
+        self.logger.info("Results: {!r}", results)
```
dbutenhof, line 81: This was debugging code and should be removed.

lib/pbench/test/unit/server/test_datasets_daterange.py
```diff
@@ -0,0 +1,132 @@
...
+        [
+            ("drb", {"owner": "drb"}, ["drb", "fio_1"]),
+            ("drb", {"access": "public"}, ["drb", "fio_2"]),
+            ("test_admin", {"owner": "drb"}, ["drb"]),
+            ("drb", {}, ["drb", "fio_1", "fio_2"]),
+            ("test", {}, ["test", "fio_1", "fio_2"]),
+            ("test_admin", {}, ["drb", "test", "fio_1", "fio_2"]),
+        ],
+    )
+    def test_dataset_list(self, query_as, login, query, results):
```
dbutenhof, line 86: I should probably rename this!

lib/pbench/test/unit/server/test_datasets_daterange.py
```diff
@@ -0,0 +1,132 @@
...
+        response = query_as(
+            {"plugh": "xyzzy", "passages": "twisty"},
+            "drb",
+            HTTPStatus.BAD_REQUEST,
+        )
+        assert response.json == {"message": "Unknown URL query keys: passages,plugh"}
+
+    def test_get_repeat_keys(self, query_as):
+        """
+        Test case requesting repeated single-value metadata keys.
```
dbutenhof, line 117: Different comment...

lib/pbench/test/unit/server/test_datasets_daterange.py
```diff
@@ -0,0 +1,132 @@
...
+            ("drb", {"access": "public"}, ["drb", "fio_2"]),
+            ("test_admin", {"owner": "drb"}, ["drb"]),
+            ("drb", {}, ["drb", "fio_1", "fio_2"]),
+            ("test", {}, ["test", "fio_1", "fio_2"]),
+            ("test_admin", {}, ["drb", "test", "fio_1", "fio_2"]),
+        ],
+    )
+    def test_dataset_list(self, query_as, login, query, results):
+        """
+        Test the operation of `datasets/list` against our set of test
```
dbutenhof, line 88: ... and fix the comment. 🐑

lib/pbench/server/api/resources/datasets_daterange.py
webbnh: 🤣  At least they were right next to each other....

lib/pbench/server/api/resources/datasets_daterange.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        Get the date range for which datasets are available to the client based
+        on authentication plus optional dataset owner and access criteria.
+
+        Args:
+            json_data: For a GET, this contains only the Flask URI template
+                values, which aren't used here
+            request: The original Request object containing query parameters
+
+
```
portante, line 47: Is the extra blank line significant?

lib/pbench/server/api/resources/datasets_daterange.py
```diff
@@ -0,0 +1,72 @@
...
+        GET /api/v1/datasets/daterange?owner=user&access=public
+        """
+
+        json = self._collect_query_params(request, self.GET_SCHEMA)
+        new_json = self.GET_SCHEMA.validate(json) if json else json
+
+        access = new_json.get("access")
+        owner = json.get("owner")
+        self.logger.info("Getting date range for user {!r}, access {!r}", owner, access)
```
portante, line 56: Could this be considered more of a "debug" message?

lib/pbench/server/api/resources/datasets_list.py
```diff
@@ -45,24 +45,30 @@ def __init__(self, config: PbenchServerConfig, logger: logging.Logger):
...
-    def _get(self, json_data: JSON, request: Request) -> Response:
+    def _get(self, _, request: Request) -> Response:
         """
         Get a list of datasets matching a set of criteria.
 
         NOTE: This does not rely on a JSON payload to identify the dataset and
         desired metadata keys; instead we rely on URI query parameters.
 
+        Args:
+            json_data: For a GET, this contains only the Flask URI template
```
portante, line 56: Why are we documenting `json_data` here, but have a `_` in the function definition?

lib/pbench/server/api/resources/datasets_metadata.py
```diff
@@ -66,10 +66,16 @@ def _get(self, _, request: Request) -> Response:
         GET. In this case, we're going to experiment with an alternative, using
         query parameters.
 
+        Args:
+            json_data: For a GET, this contains only the Flask URI template
```
portante, line 70: Not sure why we are documenting a `_`.

lib/pbench/test/unit/server/conftest.py
```diff
@@ -323,6 +323,49 @@ def attach_dataset(create_drb_user, create_user):
...
+    Supplement the conftest.py "attach_dataset" fixture with a few more
+    datasets so we can practice various queries. In combination with
+    attach_dataset, the resulting datasets are:
+
+        Owner   Access  Date        Name
+        ------- ------- ----------- ---------
+        drb     private 2020-02-15  drb
+        test    private 2002-05-16  test
+        drb     public  2020-02-15  fio_1
+        test    public  2002-05-16  fio_2
```
portante, line 340: How does this result in 4 things when there are 2 `Dataset().add()` calls below?

lib/pbench/server/api/resources/datasets_daterange.py
```diff
@@ -0,0 +1,72 @@
...
+    API_OPERATION,
+    Parameter,
+    ParamType,
+    Schema,
+)
+from pbench.server.database.database import Database
+from pbench.server.database.models.datasets import Dataset
+
+
+class DatasetsDaterange(ApiBase):
```
portante, line 19: Can we consider naming this class, `DatasetDateRange`?

portante (commented):
Looks good.

lib/pbench/test/unit/server/conftest.py
dbutenhof: Yes, I'd considered breaking it up by fixture, but having the list in one place is more useful. Maybe I should have put a *subset* of this list in the `attach_dataset` fixture, however, to complete the set. (I even thought of just collapsing them into one, but I didn't want to bother dealing with any fallout from that if it affects "older" test cases.)

npalaska (approved):

lib/pbench/server/api/resources/datasets_daterange.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+        GET /api/v1/datasets/daterange?owner=user&access=public
+        """
+
+        json = self._collect_query_params(request, self.GET_SCHEMA)
+        new_json = self.GET_SCHEMA.validate(json) if json else json
+
+        access = new_json.get("access")
+        owner = json.get("owner")
```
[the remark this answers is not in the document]
dbutenhof: Because `json` is the text user input; `new_json` is the internal representation. In particular, `new_json["owner"]` is the **user ID** rather than the **username**, and for authorization we need the latter.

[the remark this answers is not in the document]
portante: Why are we not using the user ID for authorization?  Shouldn't we be translating the name to an ID before hand?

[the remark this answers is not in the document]
portante: Not that you have to fix that in this PR, but this is a good question to consider.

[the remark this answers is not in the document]
dbutenhof: That's an old decision I'm definitely *not* going to re-open in this PR.

I'm going to suggest we defer thinking about this until we figure out exactly how we're integrating with Keycloak which may have some opinions on the subject. 👀

portante (changes requested):

webbnh:
> Whoa, why does `flake8` failure kill the entire CI job?

Because someone coded it that way?  😆

lib/pbench/server/api/resources/datasets_list.py
```diff
@@ -45,24 +45,28 @@ def __init__(self, config: PbenchServerConfig, logger: logging.Logger):
...
+
         GET /api/v1/datasets/list?start=1970-01-01&end=2040-12-31&owner=fred&metadata=dashboard.seen,server.deletion
         """
 
         # We missed automatic schema validation due to the lack of a JSON body;
         # construct an equivalent JSON body now so we can run it through the
         # validator.
-        new_json = self._validate_query_params(request, self.schema)
+        json = self._collect_query_params(request, self.schema)
+        new_json = self.schema.validate(json) if json else json
```
webbnh, line 65: This is sort of an obfuscation (or else it's much more subtle than it might otherwise be...).  I think it would be clearer if it were styled like:
```
        new_json = self.schema.validate(json) if json else {}
```
That is, if `json` is empty, lets set `new_json` to something which is obvious-to-the-reader empty, instead of this already empty thing which we happen to have lying around.

Ditto in `lib/pbench/server/api/resources/datasets_metadata.py` and `lib/pbench/server/api/resources/datasets_daterange.py`.

lib/pbench/test/unit/server/conftest.py
```diff
@@ -323,6 +330,49 @@ def attach_dataset(create_drb_user, create_user):
...
+
+    Args:
+        client: Provide a Flask API client
+        server_config: Provide a Pbench server configuration
+        create_drb_user: Create the "drb" user
+        create_admin_user: Create the "test_admin" user
+        attach_dataset: Provide some datasets
```
webbnh, line 354: I don't think we need to reference `create_drb_user` here, because I think we "inherit" it via `attach_dataset` (and we don't reference that fixture explicitly in this function, just its side-effect).

Likewise, I don't think we need to reference `create_admin_user`, since we don't reference it (or its user) in this function...unless you had intended to....  👀  (We reference the `"test"` `owner`, which I think we get via the inherited `create_user` fixture.)

lib/pbench/test/unit/server/test_datasets_daterange.py
```diff
@@ -0,0 +1,134 @@
+import datetime
+from http import HTTPStatus
+import requests
+from typing import Dict, List
+
+import pytest
+
+from pbench.server import PbenchServerConfig, JSON
```
webbnh, line 8: `J` before `P`?  😁

lib/pbench/test/unit/server/test_datasets_daterange.py
```diff
@@ -0,0 +1,134 @@
...
+        Use a list of "expected results" to determine the earliest and the
+        latest creation date of the set of datasets.
+
+        Args:
+            name_list: List of dataset names
+
+        Returns:
+            {"from": first_date, "to": last_date}
+        """
+        from_time = datetime.datetime.now(datetime.timezone.utc)
```
webbnh, line 67: Given our propensity for playing with freezeguns, I would suggest setting this to a known value, e.g.:
```
        to_time = datetime.datetime(
            year=9999, month=12, day=31, tzinfo=datetime.timezone.utc
        )
```

lib/pbench/server/api/resources/datasets_daterange.py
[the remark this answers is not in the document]
webbnh: It does look odd to be pulling `owner` from `json` and `access` from `new_json`.

(If this really is the right thing to do, perhaps move the assignment of `owner` up above the assignment of `new_json`?)

webbnh (approved):
`lib/pbench/test/unit/server/test_datasets_daterange.py` looks suspiciously like _functional_ testing.  😉  But, what's really cool (and a testament to the infrastructure that you've created!) is that there is basically nothing in `DatasetsDateRange._get()` for a unit test to test!!  😆  (That is, it's basically a straight-line sequence of subroutine calls, so if the subroutines work, then it should work...all that we could test would be that the outputs of the previous (mock) subroutine appear as inputs to the next (mock) subroutine, which is a little too white-box and a little short on value; it would be good to have a single, small test which verifies that the function returns a dict with two ISO formatted strings, but it would require at least six mocks to do that, so perhaps it's more effort than it's worth.)

Anyway, this looks very good.  I just have a few items for your consideration.

lib/pbench/server/api/resources/datasets_list.py
dbutenhof: If !json, then json already is {}... but `validate` will fail with an empty payload and I didn't want to change that. I suppose if I have to change this I'll do what I avoided doing initially:

```
    new_json = json
    if json:
        new_json = self.schema.validate(json)
```

... which avoids all ambiguity at the expense of a few extra pixels.

lib/pbench/test/unit/server/conftest.py
dbutenhof: I only moved this fixture, and I didn't really feel the need to re-evaluate the dependencies. You may be right, but I'm not sure it's relevant to this PR.

lib/pbench/test/unit/server/test_datasets_daterange.py
dbutenhof: Why, because, in the future, we might create a dataset in the future? 🤣 

If we are going to worry about that, I'm not sure there's any good fixed upper bound. I mean, we might want to make sure that Pbench still works in the year 10000... we'll probably have a managed service by then.

(Sigh.) I suppose I'll just go back to my first rejected gut instinct, which was to start them at `None` and assign `to_time` and `from_time` from the first dataset in the list, then work from there. I thought this was simpler.

lib/pbench/test/unit/server/test_datasets_daterange.py
dbutenhof: I just wish those tokens would stop moving on me...

lib/pbench/server/api/resources/datasets_daterange.py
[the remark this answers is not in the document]
dbutenhof: I don't even have a response for all this. I'm just accommodating an existing design, and there are other calls.

But I guess I'm going to put this into draft and go off and spend a day converting `_check_authorization` and all the callers to use ID so all this cruft becomes irrelevant. Ugh.

lib/pbench/server/api/resources/datasets_list.py
webbnh: > If `!json`, then json already is `{}`

Agreed....

> ... but `validate` will fail with an empty payload

...agreed...but `validate()` won't be called, so that's not a problem, right?

I don't think your proposed revision is better than using the ternary.  (And, no, you don't _have_ to change this!)

If we call `validate()`, it will return a new dictionary; so, all I was saying is that it would be better to assign `new_json` to a fresh empty dictionary than to bind it to the value that `json` has, because this is clearer.

lib/pbench/test/unit/server/conftest.py
webbnh: The question of relevance is hard to evaluate:  it appears to be new code (but, as you say, it was deleted elsewhere and reinstantiated here), which makes this a rather philosophical question.

Obviously, the code functions as is.  Mostly, I was trying to make sure that you hadn't mistyped something or left something out which was supposed to make use of `create_admin_user`.  If that's not the case, then we're good here (unless you _want_ to remove the unneeded fixture references...which I would encourage 🙂).

lib/pbench/test/unit/server/test_datasets_daterange.py
webbnh: My concern is that, given the way we use freezegun, `now()` might evaluate to an inconvenient value (but, perhaps it will be no newer than the datasets' timestamps 😆).

lib/pbench/server/api/resources/datasets_daterange.py
```diff
@@ -0,0 +1,65 @@
+from flask.json import jsonify
+from flask.wrappers import Request, Response
+from logging import Logger
+from sqlalchemy import func
```
webbnh, line 4: I am not going to point out that three of these imports are third-party, while one of them is from the standard Python library.  😉

webbnh (approved):
Looks even nicer!

lib/pbench/server/api/resources/datasets_daterange.py
```diff
@@ -0,0 +1,65 @@
...
+            None,
+            role=API_OPERATION.READ,
+        )
+
+    def _get(self, _, request: Request) -> Response:
+        """
+        Get the date range for which datasets are available to the client based
+        on authentication plus optional dataset owner and access criteria.
+
+        Args:
```
portante, line 41: Can we add a note to document why the first argument is named `_`?

lib/pbench/server/api/resources/datasets_list.py
```diff
@@ -45,45 +45,46 @@ def __init__(self, config: PbenchServerConfig, logger: logging.Logger):
...
 
-    def _get(self, json_data: JSON, request: Request) -> Response:
+    def _get(self, _, request: Request) -> Response:
         """
         Get a list of datasets matching a set of criteria.
 
         NOTE: This does not rely on a JSON payload to identify the dataset and
         desired metadata keys; instead we rely on URI query parameters.
 
+        Args:
```
portante, line 55: Can we move move the above `NOTE` to something like:

```
Args:
    _: The JSON data payload of the GET request, ignored
    request: The original Request object containing query parameters
             identifying the dataset and desired metadata keys
```

portante (changes requested):
Looks really good, thanks!

lib/pbench/server/api/resources/datasets_daterange.py
webbnh: We _really_ shouldn't need to do that:  it is common practice in Python (see the end of the box [here](https://docs.python.org/3/reference/lexical_analysis.html#reserved-classes-of-identifiers)) to indicate unused parameters by naming them `_`.

(But, I see that Dave has already addressed this...so...nevermind?)

lib/pbench/server/api/resources/datasets_list.py
webbnh: The JSON payload is not part of the signature of this function (which is what the note was explaining).  Yeah, there's a parameter there (which is why the `_` is there), but this function doesn't use it.  So, we shouldn't (have) need(ed) to document it....

portante (approved):

npalaska (approved):

PR #2821, opened 2022-05-05
Quick v0.71 fixes

webbnh:
This PR consists of two commits:  the first is a bunch of lint-pickings; the second is some quick fixes for problems identified during the v0.71 Test Day (and subsequent code review):
- a mis-handling of the return status in the `_wait_pid()` function in `tool_meister_start.py`;
- a change to raise an exception if the TDS daemon-child process fails before detaching itself; and,
- a change to raise an exception if the TDS PID file is missing after start up.

I recommend reviewing the commits individually.  The second (and only the second) commit should be backported to v0.71.

(Several of the changes originally in this PR landed in #2797.)

### DIFF
--- a/lib/pbench/agent/tool_meister_start.py
+++ b/lib/pbench/agent/tool_meister_start.py
@@ -133,6 +133,7 @@ import shutil
 import socket
 import sys
 import time
+from typing import Dict, Union
 
 from argparse import ArgumentParser, Namespace
 from distutils.spawn import find_executable
@@ -271,11 +272,9 @@ class StartTmsErr(ReturnCode.Err):
 def start_tms_via_ssh(
     exec_dir: Path,
     ssh_cmd: str,
-    ssh_path: Path,
     tool_group: ToolGroup,
     ssh_opts: str,
     redis_server: RedisServerCommon,
-    redis_client: redis.Redis,
     logger: logging.Logger,
 ) -> None:
     """Orchestrate the creation of local and remote Tool Meister instances using
@@ -296,8 +295,8 @@ def start_tms_via_ssh(
     cmd = f"{tool_meister_cmd} {redis_server.host} {redis_server.port} {{tm_param_key}} yes"
     if debug_level:
         cmd += f" {debug_level}"
-    template = TemplateSsh(ssh_cmd, shlex.split(ssh_opts), cmd)
-    tms = dict()
+    template = TemplateSsh(Path(ssh_cmd), shlex.split(ssh_opts), cmd)
+    tms: Dict[str, Union[str, int, Dict[str, str]]] = {}
     tm_count = 0
     for host in tool_group.hostnames.keys():
         tm_count += 1
@@ -412,11 +411,9 @@ class ToolDataSink(BaseServer):
     def start(
         self,
         exec_dir: Path,
-        full_hostname: str,
         tds_param_key: str,
         redis_server: RedisServerCommon,
         redis_client: redis.Redis,
-        logger: logging.Logger,
     ) -> None:
         assert (
             self.host is not None
@@ -581,7 +578,7 @@ port {redis_port:d}
         super().__init__(spec, def_host_name)
         self.pid_file = None
 
-    def start(self, tm_dir: Path, full_hostname: str, logger: logging.Logger) -> None:
+    def start(self, tm_dir: Path) -> None:
         """start_redis - configure and start a Redis server.
 
         Raises a BaseServer.Err exception if an error is encountered.
@@ -705,11 +702,11 @@ def terminate_no_wait(
     and we only check whether the message was sent.
 
     TODO: Ideally, we'd wait some reasonable time for a response from TDS and
-    then quit; we don't have that mechanism, but this means we may kill a
-    managed Redis instance before the requests propagate.
+        then quit; we don't have that mechanism, but this means we may kill a
+        managed Redis instance before the requests propagate.
 
     Args:
-        group: The tool group we're trying to terminate
+        tool_group_name: The tool group we're trying to terminate
         logger: Python Logger
         redis_client: Redis client
         key: TDS Redis pubsub key
@@ -767,6 +764,8 @@ def start(_prog: str, cli_params: Namespace) -> int:
     shf = logging.Formatter(f"{prog.name}: %(message)s")
     sh.setFormatter(shf)
     logger.addHandler(sh)
+    tm_dir = None
+    ssh_cmd = None
 
     # +
     # Step 1. - Load the tool group data for the requested tool group
@@ -1013,7 +1012,7 @@ def start(_prog: str, cli_params: Namespace) -> int:
         if orchestrate:
             logger.debug("2. starting redis server")
             try:
-                redis_server.start(tm_dir, full_hostname, logger)
+                redis_server.start(tm_dir)
             except redis_server.Err as exc:
                 raise CleanupTime(
                     exc.return_code, f"Failed to start a local Redis server: '{exc}'"
@@ -1122,12 +1121,7 @@ def start(_prog: str, cli_params: Namespace) -> int:
             logger.debug("5. starting tool data sink")
             try:
                 tool_data_sink.start(
-                    prog.parent,
-                    full_hostname,
-                    tds_param_key,
-                    redis_server,
-                    redis_client,
-                    logger,
+                    prog.parent, tds_param_key, redis_server, redis_client
                 )
             except tool_data_sink.Err as exc:
                 raise CleanupTime(
@@ -1164,15 +1158,9 @@ def start(_prog: str, cli_params: Namespace) -> int:
 
         if orchestrate:
             try:
+                # FIXME:  Should we use ssh_path instead of ssh_cmd?
                 start_tms_via_ssh(
-                    prog.parent,
-                    ssh_cmd,
-                    Path(ssh_path),
-                    tool_group,
-                    ssh_opts,
-                    redis_server,
-                    redis_client,
-                    logger,
+                    prog.parent, ssh_cmd, tool_group, ssh_opts, redis_server, logger
                 )
             except StartTmsErr as exc:
                 raise CleanupTime(
--- a/lib/pbench/agent/tool_meister_start.py
+++ b/lib/pbench/agent/tool_meister_start.py
@@ -304,7 +304,7 @@ def start_tms_via_ssh(
     cmd = f"{tool_meister_cmd} {redis_server.host} {redis_server.port} {{tm_param_key}} yes"
     if debug_level:
         cmd += f" {debug_level}"
-    template = TemplateSsh(Path(ssh_cmd), shlex.split(ssh_opts), cmd)
+    template = TemplateSsh(ssh_cmd, shlex.split(ssh_opts), cmd)
     tms: Dict[str, Union[str, int, Dict[str, str]]] = {}
     tm_count = 0
     for host in tool_group.hostnames.keys():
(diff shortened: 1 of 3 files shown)


dbutenhof (dismissed):

npalaska (dismissed):

lib/pbench/agent/tool_meister_start.py
```diff
@@ -488,7 +493,9 @@ def start(
         if pid_file.exists():
             self.pid_file = pid_file
         else:
-            logger.error("TDS daemonization didn't create %s", pid_file)
+            raise self.Err(
```
portante, line 496: Seems like line 455 should be changed to raise an error as well.

lib/pbench/agent/tool_meister_start.py
```diff
@@ -1164,15 +1169,9 @@ def start(_prog: str, cli_params: Namespace) -> int:
...
-                    prog.parent,
-                    ssh_cmd,
-                    Path(ssh_path),
-                    tool_group,
-                    ssh_opts,
-                    redis_server,
-                    redis_client,
-                    logger,
+                    prog.parent, ssh_cmd, tool_group, ssh_opts, redis_server, logger
                 )
```
webbnh, line 1175: Thoughts, @portante?

webbnh (commented):
OK, this is ready for review, now.  😁

lib/pbench/agent/tool_meister_start.py
```diff
@@ -412,11 +411,9 @@ class ToolDataSink(BaseServer):
     def start(
         self,
         exec_dir: Path,
-        full_hostname: str,
         tds_param_key: str,
         redis_server: RedisServerCommon,
         redis_client: redis.Redis,
-        logger: logging.Logger,
```
portante, line 419: If we are going to squash the two commits in this PR together when we merge this PR then this is fine.  But if we are going to keep these two commits separate, then it seems we can't remove the `logger` argument unless also remove it from lines 449 and 490.  Perhaps just moving this change to the following PR would be sufficient?

lib/pbench/agent/tool_meister_start.py
```diff
@@ -296,8 +295,8 @@ def start_tms_via_ssh(
     cmd = f"{tool_meister_cmd} {redis_server.host} {redis_server.port} {{tm_param_key}} yes"
     if debug_level:
         cmd += f" {debug_level}"
-    template = TemplateSsh(ssh_cmd, shlex.split(ssh_opts), cmd)
-    tms = dict()
+    template = TemplateSsh(Path(ssh_cmd), shlex.split(ssh_opts), cmd)
```
portante, line 307: While it is true that the first argument to the `TemplateSsh()` constructor is typed as a `Path` [1], it really seems that it should be a string (`str`) instead.  If you agree, then this should not be changed, but instead we change `TemplateSsh` to accept a `str`.

[1] https://github.com/distributed-system-analysis/pbench/blob/main/lib/pbench/agent/utils.py#L400

lib/pbench/agent/tool_meister_start.py
```diff
@@ -296,8 +295,8 @@ def start_tms_via_ssh(
...
+    tms: Dict[str, Union[str, int, Dict[str, str]]] = {}
```
portante, line 308: Does this type hint really help?  This is so hard to parse that it's value is dubious.

lib/pbench/agent/tool_meister_start.py
portante: I think `ssh_cmd` should be dropped in favor of `ssh_path`, which should not be a `Path` object, but just the return value from `shutil.which("ssh")`.  Then as mentioned elsewhere, fix the `TempalteSsh()` constructor to accept a `str`.

portante (changes requested):

lib/pbench/agent/tool_meister_start.py
dbutenhof: I guess you're right. I was sure I'd started that code depending on a `Path`; but looking at `main` it's not even a "path-like object" because I'm pulling it into the `argv` array without a `str(ssh_cmd)`. However, although the `subprocess.Popen` documentation doesn't seem to say anything to suggest this, it clearly does actually *work* as a path-like object, because it's being called both ways.

One however (from `start`) explicitly creates a `Path` from the `shutil.which`, while the other just passes a raw "ssh" string. It'd probably be better if both passed `shutil.which` output, and it'd probably be clearer to make it a string under the circumstances.

lib/pbench/agent/tool_meister_start.py
dbutenhof: I agree some of these are getting a bit gnarly for human readers, but I think part of this is motivated by the fact that our tools are getting smarter; I (and I believe Webb) use IDEs that have started to diagnose potential problems with the use of variables due to our lax type hints. (E.g., this may be a type that doesn't have `.get` or can't be indexed or whatever.) The result is that we're seeing a lot of red representing false problems just because we haven't properly informed the IDE of our intent.

I think our code base is still a ways from being able to put a strong type-hint aware lint checker into our Jenkins pipeline and have any chance of success; however I'd love to be able to work towards doing that. "Picking lint" like this along the way moves us towards better (testable) self-documentation, and I look at that as a good thing.

We could probably help to make it more readable with well-chosen typedefs as Webb and I tried to do with the `JSON` family. (Unfortunately this isn't *quite* a `JSONOBJECT` since we're tying this one to `Dict[str, str]`.)

dbutenhof (changes requested):
I suppose it's "in scope" for this scattered cleanup to resolve the `TemplateSsh` ambiguity around the command parameter. Aside from that, this looks good.

lib/pbench/agent/tool_meister_start.py
webbnh: My intention was to post (just) the second commit as a PR against the `b0.71` branch and then squash-and-merge this branch on to `main`...but, I'm open to other suggestions.

lib/pbench/agent/tool_meister_start.py
portante: I thought the whole point of a language was to make it so that a human being can write in it reasonably, and the computer does the hard work of transforming what is written into what it can execute.

That is one reason why I am desperately trying to drop Perl!

These Python3 Type Hints seem to be taking Python the other direction: you need a IDE to derive value for what is written?

I am just not buying it.  I'll resolve this issue, but it is just not the right direction for the language.

dbutenhof: A language should be readable, and it should be semantically robust enough to support technologies that help prevent or diagnose coding errors at build time rather than at run time. It's why we run `flake8` and `LGTM` ... although the degree to which they can help us is significantly limited by Python's flexibility in typing and name binding.

Perl and Python are both originally "scripting" languages, not "software development" languages: they were designed to be easy and quick to write and hack, without much concentration on supporting readability, maintainability, or important technologies like static analysis. The fact that a name can bind to anything is a great convenience for throwaway code; it's a massive complication for long-term maintenance and growth. Type hints allow unambiguous documentation both for the reader and for tools of what the developer *intended*, instead of requiring a reader to study the internal code and call points to figure out exactly how to call a function.

The fact that Python itself doesn't care can be characterized as a throwback to the "throwaway scripting language" mentality, although, to be fair, there's also a substantial backwards compatibility issue.

Python 3.9 will make type hinting just a bit neater and easier because we don't have to worry about importing the `typing` generics; we can use `dict[str, str]` directly.

I find it a lot easier to read code with type hints than without; I think we'll have fewer bugs the more we can rely on accurate type hints and linting tools. But, yeah; that's at least partly my opinion and instinct. There's probably data out there to be found if we want to bother to look.

lib/pbench/agent/tool_meister_start.py
webbnh: I'm guessing that `subprocess.Popen()` wants a "command" for this argument, and, while that would typically be a _path_, it certainly doesn't have to be -- in appropriate circumstances I believe it could be a shell alias or a shell function.  So, if we want to be fully general, we should be requiring a string, not a `Path`.  (As for it "working"...Python's differentiation between `str` and `Path` is pretty thin, since they are both "path-like" and it seems to be trivial to convert between them...but, I'd like to see if we can do more than just "working" -- I would aspire toward making it "right".  😆)

dbutenhof: The thing is that `subprocess.Popen` takes an `argv` list; the first element in that list can apparently be "path-like" while the others are just strings. There must be code somewhere that's treating the first element specially... or maybe it's all just buried down in C code with some interface magic. Weird, in any case, and I don't particularly like that this happens transparently ... or that I managed to get the type hint wrong but it doesn't matter. 😵‍💫

lib/pbench/agent/tool_meister_start.py
webbnh: The challenge here is that we're using a Python dictionary as a generic "container", so the values could be strings, integers, or...other dictionaries (which are themselves string-string mappings)...so, that necessitates a somewhat more arcane description.  The parsing difficulty is partly that we're not used to it, and it's partly that the code's usage of this variable is simply that complex.

As for the value, well, that expression describes what we actually do.  I could use a cop-out like this:
```
    tms: Dict[str, Any] = {}
```
but, I feel like that provides less value.  (That is, I don't think the complicated type hint particularly hurts anything, so I appreciate the specificity.)  We could define a custom type for this, but I'm not convinced that that would improve the situation (especially since I don't think this type is used anywhere else).

webbnh: _[I didn't see the previous two posts when I posted my reply...the web page seems to be selective as to what it auto-updates.  🙂]_

I agree with Dave's points:  Python seems to be a superb language for prototyping, but it seems highly questionable that anyone would write anything "real" in it.  🤣  The fact that I can't tell whether the code will work until I run it (because we don't know what the types are until the code executes...and, possibly, not even then!) seems nigh on to silly.  Using type hints (followed by a code analyzer, such as Dave and I have in our respective IDEs) goes a long way toward alleviating those problems, and the more specific they are the better they work.  I'll echo Dave's comment that type hints, while not necessarily themselves making the declarations easier to read, do make the rest of the code much easier to understand (e.g., because you don't have to look at the **call sites** to know what types the parameters are supposed to have).  I think that, as we use them and get used to them, we'll start to value them more (and, at some point, we'll enable static analysis of our code, which will help even more).

lib/pbench/agent/tool_meister_start.py
webbnh: Given the confusion here (i.e., in general, around our use of "path-like" values), I would prefer not to use the name `ssh_path` to refer to something which is not a `Path` object.  So, would it be OK to stick with a variable named `ssh_cmd` which is a `str`, and get rid of the `Path` object?  (Or, could you propose another name?)

portante: `ssh_cmd` is fine.

lib/pbench/agent/utils.py
```diff
@@ -397,12 +396,12 @@ class Return(NamedTuple):
...
         stderr: str
 
-    def __init__(self, ssh_cmd: Path, ssh_args: List[str], cmd: str):
+    def __init__(self, ssh_cmd: str, ssh_args: List[str], cmd: str):
         """
         Create an SSH template object
 
         Args:
-            ssh_cmd: Path to the ssh command
+            ssh_cmd: file path of the ssh command
```
dbutenhof, line 404: `Path`/"path" ... sigh. Could be misleading, but hard to clarify, and at least the type hint is now consistent with what we expect. 🍤

dbutenhof (approved):

lib/pbench/agent/tool_meister_start.py
portante: So squash and merge it is.

portante (approved):
Please use a "squash and merge" operation.  Thanks!

PR #2825, opened 2022-05-05
Convert authorization to use ID rather than string

### LINKED ISSUE #2824: Why are we not using the user's ID for authorization?
Why are we not using the user ID for authorization?  Shouldn't we be translating the name to an ID before hand?

_Originally posted by @portante in https://github.com/distributed-system-analysis/pbench/pull/2820#discussion_r866009319_

dbutenhof:
PBENCH-646

User authorization via ApiBase._check_authorization has always used the "raw" specified username string from the client. It makes more sense to use the validated internal form (the stringified DB row ID), and this in particular will simplify handling of a username that appears in a URL query parameter.

Resolves #2824

### DIFF
--- a/lib/pbench/server/api/resources/__init__.py
+++ b/lib/pbench/server/api/resources/__init__.py
@@ -34,8 +34,8 @@ class UnauthorizedAccess(Exception):
         self,
         user: Union[User, None],
         operation: "API_OPERATION",
-        owner: str,
-        access: str,
+        owner: Union[str, None],
+        access: Union[str, None],
         http_status: int = HTTPStatus.FORBIDDEN,
     ):
         self.user = user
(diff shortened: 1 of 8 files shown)


portante (commented):
Looks good, but what's up with the test failures?

portante:
> Looks good, but what's up with the test failures?

Bugger all, looks like a flaky Tool Meister test.

portante:
Restarted the tests to see if it'll pass.

portante (approved):

dbutenhof:
> Restarted the tests to see if it'll pass.

And ... test-63 failed. Same one? In any case, same pattern I've seen with several tool meister tests. Maybe Webb's right and this is environmental rather than an inherently flakey test, but it's annoying.

I restarted again...

npalaska (approved):
Looks good

PR #2826, opened 2022-05-06
New user metadata namespace for Overview page

dbutenhof:
PBENCH-645

This change re-purposes the `user.*` Dataset Metadata namespace as a per-user key space, where each authenticated client with READ access to a given dataset will be able to write independent metadata attached to that dataset. (Only the owner has UPDATE access to change the "global" metadata keys in the "dashboard.*" namespace.)

At the Metadata internal class level, methods now allow specifying a `User` to associate each item of metadata. At the API level, the user is always the authenticated user (or None for only global metadata).

An unauthenticated client can ask for `user.*` keys, but the value will always be None as authenticated clients will never create global values for the `user.*` namespace and unauthenticated clients can't ever set metadata values.

### DIFF
--- a/lib/pbench/server/api/resources/__init__.py
+++ b/lib/pbench/server/api/resources/__init__.py
@@ -1022,10 +1022,9 @@ class ApiBase(Resource):
                 metadata[i] = UtcTimeHelper(dataset.uploaded).to_iso_string()
             elif Metadata.is_key_path(i, Metadata.USER_METADATA):
                 native_key = Metadata.get_native_key(i)
-                user = None
+                user: Optional[User] = None
                 if native_key == Metadata.USER_NATIVE_KEY:
-                    authorized_user: Optional[User] = Auth.token_auth.current_user()
-                    user = authorized_user
+                    user = Auth.token_auth.current_user()
                 try:
                     metadata[i] = Metadata.getvalue(dataset=dataset, key=i, user=user)
                 except MetadataNotFound:
--- a/lib/pbench/server/api/resources/__init__.py
+++ b/lib/pbench/server/api/resources/__init__.py
@@ -4,7 +4,7 @@ from http import HTTPStatus
 import json
 from json.decoder import JSONDecodeError
 from logging import Logger
-from typing import Any, Callable, List, Union
+from typing import Any, Callable, List, Optional, Union
 
 from dateutil import parser as date_parser
 from flask import request
@@ -1021,8 +1021,14 @@ class ApiBase(Resource):
             elif i == Dataset.UPLOADED:
                 metadata[i] = UtcTimeHelper(dataset.uploaded).to_iso_string()
             elif Metadata.is_key_path(i, Metadata.USER_METADATA):
+                native_key = Metadata.get_native_key(i)
+                user = None
+                if native_key == Metadata.USER_NATIVE_KEY:
+                    authorized_user: Optional[User] = Auth.token_auth.current_user()
+                    user = authorized_user
                 try:
-                    metadata[i] = Metadata.getvalue(dataset=dataset, key=i)
+                    self.logger.info("Accessing {!r} ({!r}) as {}", i, native_key, user)
+                    metadata[i] = Metadata.getvalue(dataset=dataset, key=i, user=user)
                 except MetadataNotFound:
                     metadata[i] = None
             else:
--- a/lib/pbench/server/api/resources/__init__.py
+++ b/lib/pbench/server/api/resources/__init__.py
@@ -1027,7 +1027,6 @@ class ApiBase(Resource):
                     authorized_user: Optional[User] = Auth.token_auth.current_user()
                     user = authorized_user
                 try:
-                    self.logger.info("Accessing {!r} ({!r}) as {}", i, native_key, user)
                     metadata[i] = Metadata.getvalue(dataset=dataset, key=i, user=user)
                 except MetadataNotFound:
                     metadata[i] = None
--- a/lib/pbench/server/api/resources/datasets_metadata.py
+++ b/lib/pbench/server/api/resources/datasets_metadata.py
@@ -97,10 +97,7 @@ class DatasetsMetadata(ApiBase):
             "name": "datasetname",
             "metadata": [
                 "dashboard.seen": True,
-                "user": {
-                    "cloud": "AWS",
-                    "contact": "john.carter@mars.org"
-                }
+                "user": {"favorite": True}
             ]
         }
 
@@ -114,7 +111,7 @@ class DatasetsMetadata(ApiBase):
 
         [
             "dashboard.seen": True,
-            "user": {"cloud": "AWS", "contact": "json.carter@mars.org}
+            "user": {"favorite": False}
         ]
         """
         name = json_data["dataset"]
@@ -129,10 +126,20 @@ class DatasetsMetadata(ApiBase):
         # Validate the authenticated user's authorization for the combination
         # of "owner" and "access".
         #
-        # Note that we allow authenticated clients to set "user.*" namespace
-        # metadata with READ access to the dataset because it doesn't affect
-        # the Dataset itself and isn't visible to other users, but we require
-        # UPDATE access for any keys outside the "user" namespace.
+        # The "unusual" part here is that we make a special case for
+        # authenticated that are not the owner of the data: we want to allow
+        # them UPDATE access to PUBLIC datasets (to which they naturally have
+        # READ access) as long as they're only trying to modify a "user."
+        # metadata key:
+        #
+        # * We want to validate authorized non-owners for READ access if
+        #   they're only trying to modify "user." keys;
+        # * We want to validate unauthorized users for UPDATE because they have
+        #   READ access to "public" datasets but still can't write even "user."
+        #   metadata;
+        # * We want to validate authorized users for UPDATE if they're trying
+        #   to set anything other than a "user." key because only the owner can
+        #   do that...
         role = API_OPERATION.READ
         if not Auth.token_auth.current_user():
             role = API_OPERATION.UPDATE
@@ -147,10 +154,9 @@ class DatasetsMetadata(ApiBase):
         failures = []
         for k, v in metadata.items():
             native_key = Metadata.get_native_key(k)
-            user = None
+            user: Optional[User] = None
             if native_key == Metadata.USER_NATIVE_KEY:
-                authorized_user: Optional[User] = Auth.token_auth.current_user()
-                user = authorized_user
+                user = Auth.token_auth.current_user()
             try:
                 Metadata.setvalue(key=k, value=v, dataset=dataset, user=user)
             except MetadataError as e:
--- a/lib/pbench/server/api/resources/datasets_metadata.py
+++ b/lib/pbench/server/api/resources/datasets_metadata.py
@@ -1,10 +1,12 @@
 from http import HTTPStatus
 from logging import Logger
+from typing import Optional
 
 from flask.json import jsonify
 from flask.wrappers import Request, Response
 
 from pbench.server import JSON, JSONOBJECT, PbenchServerConfig
+from pbench.server.api.auth import Auth
 from pbench.server.api.resources import (
     APIAbort,
     API_OPERATION,
@@ -19,6 +21,7 @@ from pbench.server.database.models.datasets import (
     Metadata,
     MetadataError,
 )
+from pbench.server.database.models.users import User
 
 
 class DatasetsMetadata(ApiBase):
@@ -125,16 +128,35 @@ class DatasetsMetadata(ApiBase):
 
         # Validate the authenticated user's authorization for the combination
         # of "owner" and "access".
-        self._check_authorization(str(dataset.owner_id), dataset.access)
+        #
+        # Note that we allow authenticated clients to set "user.*" namespace
+        # metadata with READ access to the dataset because it doesn't affect
+        # the Dataset itself and isn't visible to other users, but we require
+        # UPDATE access for any keys outside the "user" namespace.
+        role = API_OPERATION.READ
+        if not Auth.token_auth.current_user():
+            role = API_OPERATION.UPDATE
+        else:
+            for k in metadata.keys():
+                if Metadata.get_native_key(k) != Metadata.USER_NATIVE_KEY:
+                    role = API_OPERATION.UPDATE
+        self._check_authorization(
+            str(dataset.owner_id), dataset.access, check_role=role
+        )
 
         failures = []
         for k, v in metadata.items():
+            native_key = Metadata.get_native_key(k)
+            user = None
+            if native_key == Metadata.USER_NATIVE_KEY:
+                authorized_user: Optional[User] = Auth.token_auth.current_user()
+                user = authorized_user
             try:
-                Metadata.setvalue(dataset, k, v)
+                Metadata.setvalue(key=k, value=v, dataset=dataset, user=user)
             except MetadataError as e:
                 self.logger.warning("Unable to update key {} = {!r}: {}", k, v, str(e))
                 failures.append(k)
         if failures:
             raise APIAbort(HTTPStatus.INTERNAL_SERVER_ERROR)
-        results = self._get_metadata(name, list(metadata.keys()))
+        results = self._get_dataset_metadata(dataset, list(metadata.keys()))
         return jsonify(results)
--- a/lib/pbench/server/database/models/datasets.py
+++ b/lib/pbench/server/database/models/datasets.py
@@ -821,51 +821,34 @@ class Metadata(Database.Base):
     # the "server" namespace, and are strictly controlled by keyword path:
     # e.g., "server.deleted", "server.archived";
     #
-    # Metadata keys within the "user" namespace allow exposure of user-specific
-    # configuration data through the Dataset metadata API; for example, a
-    # client can ask whether a specific dataset has been "favorited" by the
-    # authenticated user (TBD) and the allowable access settings (currently
-    # private and public).
-    #
-    # Metadata keys intended for use by the dashboard client are in the
-    # "dashboard" namespace. While these can be modified by any API client, the
-    # separate namespace provides some protection against accidental
-    # modifications that might break dashboard semantics. This is an "open"
-    # namespace, allowing the dashboard to define and manage any keys it needs
-    # within the "dashboard.*" hierarchy.
+    # The "dashboard" and "user" namespaces can be written by an authenticated
+    # client to track external metadata. The difference is that "dashboard" key
+    # values are visible to all clients with READ access to the dataset, while
+    # the "user" namespace is visible only to clients authenticated to the user
+    # that wrote the data. That is, "dashboard.seen" is a global dashboard
+    # metadata property visible to all users, while "user.favorite" is visible
+    # only to the specific user that wrote the value; each authenticated user
+    # may have its own unique "user.favorite" value.
     #
     # The following class constants define the set of currently available
     # metadata keys, where the "open" namespaces are represented by the
-    # syntax "user.*".
+    # syntax "dashboard.*" and "user.*" which allow clients to control the
+    # key names using a "dotted path" notation like "dashboard.seen" or
+    # "dashboard.contact.email".
 
-    # DASHBOARD is arbitrary data saved on behalf of the dashboard client; the
-    # reserved namespace differs from "user" only in that reserving a primary
-    # key for the "well known" dashboard client offers some protection against
-    # key name collisions.
+    # DASHBOARD is arbitrary data saved on behalf of the dashboard client as a
+    # JSON document. Writing these keys requires ownership of the referenced
+    # dataset, and the data is visible to all clients with READ access to the
+    # dataset.
     #
     # {"dashboard.seen": True}
     DASHBOARD = "dashboard.*"
 
     # USER is arbitrary data saved with the dataset on behalf of an
-    # authenticated user, as a JSON document.
-    #
-    # Note that the hierarchical key management in the getvalue and setvalue
-    # static methods (used consistently by the API layer) allow client code
-    # to interact with these either as a complete JSON document ("user") or
-    # as a full dotted key path ("user.contact.name.first") or at any JSON
-    # layer between.
-    #
-    # API keyword validation uses the trailing ".*" here to indicate that only
-    # the first element of the path should be validated, allowing the client
-    # a completely uninterpreted namespace below that.
-    #
-    # Metadata in the "user" namespace is user-specific; it will always be set
-    # and accessed using the authenticated user ID. That is, different users
-    # will see their own "user" namespace keys for any given dataset.
-    #
-    # TODO: ?? It occurs to me that we could allow "user" metadata rows without
-    # a user_id, which could act as a "default" for all users examining the
-    # dataset -- I'm not entirely sure this is a good model.
+    # authenticated user, as a JSON document. Writing these keys requires READ
+    # access to the referenced dataset, and are visible only to clients that
+    # are authenticated as the user which set them. Each user can have its own
+    # unique value for these keys, for example "user.favorite".
     #
     # {"user.favorite": True}
     USER_NATIVE_KEY = "user"
@@ -997,7 +980,7 @@ class Metadata(Database.Base):
         Returns:
             native SQL key name ("user")
         """
-        return key.lower().split(".").pop(0)
+        return key.lower().split(".")[0]
 
     @staticmethod
     def is_key_path(key: str, valid: List[str]) -> bool:
@@ -1007,7 +990,7 @@ class Metadata(Database.Base):
         valid. If the key is a dotted path and the first element plus a
         trailing ".*" is in the list, then this is an open key namespace where
         any subsequent path is acceptable: e.g., "user.*" allows "user", or
-        "user.contact", "user.contact.name", etc.
+        "user.favorite", "user.notes.status", etc.
 
         Args:
             key: metadata key path
@@ -1040,16 +1023,28 @@ class Metadata(Database.Base):
         level Metadata object. The full JSON value of a top level key can be
         acquired directly using `Metadata.get(dataset, key)`
 
-        E.g., for "user.contact.name" with the dataset's Metadata value for the
-        "user" key as {"contact": {"name": "Dave", "email": "d@example.com"}},
-        this would return "Dave", whereas Metadata.get(dataset, "user") would
-        return the entire user key JSON, such as
-        {"user" {"contact": {"name": "Dave", "email": "d@example.com}}}
+        For example, if the metadata database has
+
+            "dashboard": {
+                    "contact": {
+                        "name": "dave",
+                        "email": "d@example.com"
+                    }
+                }
+
+        then Metadata.get(dataset, "dashboard.contact.name") would return
+
+            "Dave"
+
+        whereas Metadata.get(dataset, "dashboard") would return the entire user
+        key JSON, such as
+
+            {"dashboard" {"contact": {"name": "Dave", "email": "d@example.com}}}
 
         Args:
             dataset: associated dataset
             key: hierarchical key path to fetch
-            user:
+            user: User-specific key value (used only for "user." namespace)
 
         Returns:
             Value of the key path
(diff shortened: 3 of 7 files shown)


portante (approved):

portante (approved):

lib/pbench/server/database/models/datasets.py
```diff
@@ -857,21 +846,44 @@ class Metadata(Database.Base):
...
-    # API keyword validation uses the trailing ".*" here to indicate that only
-    # the first element of the path should be validated, allowing the client
-    # a completely uninterpreted namespace below that.
+    # {"user.access": ["private", "public"]}
+    USER_ACCESS = "user.access"
+
+    # USER_THRESHOLD allows direct access (e.g., in /datasets/list) to the user
+    # preference setting controlling advance notice of expiring datasets in the
+    # Overview page. This will be mapped to user profile metadata when that's
+    # available.
```
webbnh, line 867: Once again, I'm not understanding how this is supposed to work.  That is, I can imagine the user (or the Dashboard on the user's behalf) wanting to list all of the datasets which are going to expire within the user's default threshold.  To do that, the dashboard would do a `/datasets/list` query with an "expiring before" qualifier.  To come up with a value for that qualifier, the dashboard would request the user setting for the threshold.  I can see how this value might be stored under a `user.threshold` key, but I don't see any particular reason to architect that.  The dashboard is likely to have a number of pieces of context that it wants to store and retrieve on behalf of the user...why single this one out?

If the server needs special access to it, then that would be a valid reason, but I would advise against that level of coupling.  (I don't think it is overburdensome to expect that the Dashboard would handle this.)

webbnh (changes requested):
I feel like this is a considerable departure from the previous vision of how the metadata namespaces would be used.  The old model seemed pretty solid and comprehensive...whereas this seems like the new keys don't fit well and the change seems to remove support for third-party clients...am I misunderstanding this change?

lib/pbench/server/database/models/datasets.py
```diff
@@ -857,21 +846,44 @@ class Metadata(Database.Base):
...
+    # user-specific "favorite" metadata. NOTE: for now, this is hardcoded as
+    # False until we have a framework for user-specific metadata.
     #
-    # Note that the hierarchical key management in the getvalue and setvalue
-    # static methods (used consistently by the API layer) allow client code
-    # to interact with these either as a complete JSON document ("user") or
-    # as a full dotted key path ("user.contact.name.first") or at any JSON
-    # layer between.
+    # {"user.favorite": False}
+    USER_FAVORITE = "user.favorite"
```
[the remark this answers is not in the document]
dbutenhof: This is cross-cutting: each user can "favorite" a distinct set of datasets. This returns the specific user-specific favorite setting for this dataset instead of requiring a separate user metadata call (once we have such a thing). I thought that was described in the comment. 😦

lib/pbench/server/database/models/datasets.py
```diff
@@ -857,21 +846,44 @@ class Metadata(Database.Base):
...
+    # USER_ACCESS allows direct access (e.g., in /datasets/list) to the set of
+    # access policies the authenticated user is allowed to set. Initially this
+    # is always just private and public; in the future this can be extended to
+    # support groups.
     #
-    # API keyword validation uses the trailing ".*" here to indicate that only
-    # the first element of the path should be validated, allowing the client
-    # a completely uninterpreted namespace below that.
+    # {"user.access": ["private", "public"]}
+    USER_ACCESS = "user.access"
```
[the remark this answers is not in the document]
dbutenhof: It's the set of "access" values the specific authenticated user is allowed to set. So, yeah, it's a server IAM setting specific to the authenticated user. As such I like it under a "user" namespace, whereas under the "server" namespace it'd need a more specific name. Or maybe it needs a more specific name here, anyway. Code is easy: names are hard.

Sigh; but, then again, this won't vary among the datasets returned as `user.favorite` will. I put this here because we *have* a mechanism for dataset metadata, and I like the idea of making user favorite data visible when querying a list of datasets instead of requiring a separate user metadata call for each dataset. (Plus of course the fact that we don't currently have any such mechanism.)

But you're probably right: I shouldn't cram this in here, even though it was convenient.

lib/pbench/server/database/models/datasets.py
dbutenhof: Again, valid points, and putting it here just seemed like a nice convenience; and an expedience since we have the mechanism, and they're calling it, and we have no user profile to deal with this.

We need some server support to persist user preference settings like this; and while it's possible they'll only be used at the dashboard, I think there's probably a use case for something like an email to warn about expiring datasets rather than count on users logging in. So a server supported user preference seemed "preferable" to having the dashboard chuck it somewhere under the "dashboard" key namespace like "saved" and "seen".

However, like the IAM access values, this isn't something that differs between datasets, and in fact isn't directly related to any of them. Dang. So it should be either part of the user profile, or part of an extended user metadata model which we don't have.

And while not as excessive as returning the access list for every dataset, it's still a bit inefficient.

lib/pbench/server/database/models/datasets.py
```diff
@@ -906,10 +918,15 @@ class Metadata(Database.Base):
     # --- Standard Metadata keys
 
     # Metadata keys that clients can update
-    USER_UPDATEABLE_METADATA = [DASHBOARD, USER]
+    USER_UPDATEABLE_METADATA = [DASHBOARD]
```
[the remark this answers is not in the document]
dbutenhof: "user" was never defined, and as we discussed at the time it's not even necessarily clear that it makes sense. I liked this for favorites, because mapping dataset-specific user context into the dataset queries makes a lot of sense.

We never addressed any real model for how "clients" might share a single key namespace, which means there really isn't much advantage in user and dashboard over just dashboard.

We can add a new "client" namespace that behaves like "dashboard" if we want, and there's a Jira to make that configurable in a server instance (e.g., via an admin API or config file)... which would let a server admin build a namespace for each client that wants one. Or we can just build a "client" and document some "good neighbor" sharing rules ... or maybe nobody will even care.

lib/pbench/server/database/models/datasets.py
```diff
@@ -857,21 +846,44 @@ class Metadata(Database.Base):
...
+    # Temporal.Duration.from("P2W")), as does the Python pandas library (for
+    # example, pandas.Timedelta("P2W")). For now, let's use ISO 8601 format,
+    # but restrict it to "days", which is easy to parse and produce in both
+    # languages.
+    #
+    # {"user.threshold": "P14D"}
+    USER_THRESHOLD = "user.threshold"
+
+    # DELETION timestamp for dataset based on user settings and system
+    # settings at time the dataset is created.
```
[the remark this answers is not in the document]
dbutenhof: This was just moved, not changed; but, yeah.

dbutenhof (commented):
Apparently I need to do some more noodling here. (Sigh.)

lib/pbench/server/database/models/datasets.py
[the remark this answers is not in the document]
webbnh: I'm grasping for a framework in which to understand this.  I get the idea that the Dashboard wants an easy way to fetch the list of "favorite" datasets.  And, I totally understand that there's a matrix problem here -- "user favorite dataset" lies at the intersection of "dataset" and "user".  But, I don't think that "dataset X is a favorite of user Y" should be an attribute of a given dataset (although, structuring it that way would likely make it easy to pull the set of results out of Elasticsearch), because that requires us to modify the records when the user changes his/her mind.

I think that leaves us storing the favorite-ness outside Elasticsearch...which I think means that there is no advantage to attaching it to the dataset.

My instinct is that the list of favorite datasets should be an attribute of the user.  The Dashboard can fetch it and then fetch the individual datasets.  In real terms, the number of favorites is unlikely to be large, and, if we need to, we can provide a mechanism which allows the server to return multiple datasets (e.g., via a list of IDs) in a single query.  (Or, we can create a special request which is implicitly driven by the favorite list, but I don't like that because it probably requires tight coupling and/or architecture.)

The advantage of associating it with the user is that adding or removing users or individual favorites does not affect datasets (or require touching them in any way).  There is the issue of what happens when a user-favorited dataset (e.g., one owned by someone else) is deleted, but I presume that we can handle that when the request is made to fetch the dataset (rather than trying to update all users' metadata when the dataset is removed).

Making it user metadata also means that it can be implemented on the client side, without requiring mapping the key or otherwise special handling in the `user` namespace.

lib/pbench/server/database/models/datasets.py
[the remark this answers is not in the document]
webbnh: Right:  the value of this key is not associated with any given dataset...but, then again, in the previous model, nothing in the `user` namespace was associated with a given dataset.  I would put it under the `server` namespace, for the reason that you point out; and, there, it would potentially return the server's value associated with the requestor's login.

As for mechanism, I though Nikhil implemented most of what we needed...it just didn't have sufficient priority at the time to be integrated...perhaps that has finally changed!

lib/pbench/server/database/models/datasets.py
webbnh: So, I'm thinking about how this value would be used.

What I _think_ we want to do is have the server do a single, complex database query to return all of the datasets which are "about to expire".  If each dataset had an expiration date and a threshold, then perhaps we could make a query for all entries whose expiration minus its threshold is before today.  In order for this to be possible, the threshold has to be an attribute of the dataset, stored with the dataset.

What I'm pretty sure that we don't want to do is to iterate over each dataset, locate its user, look up the threshold associated with that user, and then check the expiration date.  That sounds crushingly slow.

It would be slightly better to iterate over each user, grab its threshold, add it to the current date, and then do a database lookup for all datasets owned by that user with expirations less than the target date.  That's still not as nice as doing it in a single (albeit, complex) query.  But, this would allow us to associate the threshold with the user instead of with the dataset.

Nevertheless, I think I would recommend having the threshold be per-dataset attribute which is set (on upload) from a default which is user attribute (unless it is specified explicitly).  (The user can update the dataset's threshold later with a dataset-metadata change.)  I think this will ultimately be more flexible and easier to administer, even if it does take up more space (presuming that there are many more datasets than there are users).

In any case, I don't think the threshold value belongs in the `user` namespace.  There should probably be one in the `dataset` namespace for each dataset, and a default value (if it's settable on a per-user basis) should be stored in the user profile along with the user's email address.

lib/pbench/server/database/models/datasets.py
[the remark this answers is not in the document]
webbnh: I don't like the idea of needing a server admin to configure a per-client namespace.  I feel like, if a user can log into the server, s/he should be able to create whatever metadata his client needs (up to some size limit...).  If someone wants to bring their own client, I feel like that should be OK (but, as you say, we should outline some "good neighbor" rules).  I'm willing to support having a custom namespace for the Dashboard, because we like them best, but I think we should be making provisions for other clients (if for no other reason than to dissuade them from invading the `dashboard` namespace).

I think that once we have Nikhil's mechanism hooked up to the `user` namespace, it'll all make good sense.

(Maybe the mistake was naming that namespace "user" instead of "client", since user has other connotations in the domain.)

lib/pbench/server/database/models/datasets.py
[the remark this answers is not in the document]
dbutenhof: Elasticsearch has nothing to do with this.

We have a piece of data that's both user- and dataset-specific; the future user metadata infrastructure will provide a way to get all dataset favorites through the user profile, but this provides a way to get the authenticated user's setting for a specific dataset. This provides an easy and straightforward way for the dashboard to get the information where it matters rather than doing a separate user query and then threading the two lists together.

lib/pbench/server/database/models/datasets.py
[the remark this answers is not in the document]
dbutenhof: This is an IAM configuration that should be accessible in some form through the user profile. Maybe it even makes sense to add that now, as a read-only `User` object property, which current would be hardcoded to the `["public", "private"]` list and can be extended in the future assuming the entire user DB model doesn't change with Keycloak.

My intent here was to provide a shortcut where the dashboard might need the information, as part of querying a dataset, to populate a PUBLISH pulldown. The repetition of having the same user/IAM-defined values for each dataset in a list does bug me a bit, although it's convenient.

I think it's conceivable that we could have dataset-specific lists. E.g., if a dataset was owned by a group, an authorized user might be able to make the data public, or choose another related group, but maybe couldn't make it "private". If we went that way (which sounds too confusing to be viable), putting it here makes sense. Since that's unlikely, though, I guess adding it to the `User` is probably a better choice.

lib/pbench/server/database/models/datasets.py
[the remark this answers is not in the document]
dbutenhof: Yeah, "user" wasn't the best name in the first place, and I like `user.favorite` in the new model regardless of contextual dispute about the other two. 😄 

I could easily create a "client" namespace now or later. We could easily allow dynamically creating namespaces with an API to register/list/delete dataset metadata namespaces. Right now I think that's probably overkill.

lib/pbench/server/database/models/datasets.py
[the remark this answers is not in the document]
webbnh: I guess I don't understand what the specific requirements or intended design problem is.  There are two which leap to mind immediately (and I suspect that, at different times, the Dashboard will want to do each of these things):

1. support the Dashboard marking specific results within a larger display to show which ones have been marked as the user's favorites
1. support the Dashboard fetching/displaying a list of results where only the user's favorites are listed

For the first, we could have the `/datasets/list` "decorate" the returned values to indicate which are favorited.  I don't know that that is any easier to do on the server side than it would be to do on the client side (once the client has the list of favorites).  Either way, this seems like a relatively straightforward problem to solve, without adding to the metadata on the dataset side.

The second seems like a more challenging problem.  I'm not immediately coming up with a solution other than fetching the list of favorites from the user and then fetching those results.  If we maintained as part of the dataset the list of users who had favorited it, obviously it would be pretty easy, but I don't think we want to burden individual datasets with an unbounded list of the users which had marked it as their favorite.  Of course, if the client fetched the available datasets, it could filter out any which weren't on the user's list, but that's a pretty unsatisfying solution.

lib/pbench/server/database/models/datasets.py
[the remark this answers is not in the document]
webbnh: My instinct is that the list of operations (e.g., "publish") and the list of permissions (e.g, "read", "write", "delete") should be basically orthogonal -- when the user makes a request, the server checks their intersection, and if there's not a "true" there, then the user gets a `403`.  I don't think that we need to bother trying to populate pull downs or to gray out options on a per-dataset basis.  That seems like it will be a lot of effort (or data, or bandwidth, or...) for very little return.  (That is, is it really better to say "you cannot do that" (by removing or graying out the option) than it is to let the user try and _then_ say "you cannot do that"?)

lib/pbench/server/database/models/datasets.py
webbnh: > My intent with `datasets/list` was that it be metadata-driven, and in theory you should be able to select based on whatever metadata you can show.

That sounds great, but "metadata" in this context probably has to be limited to _dataset_ metadata; user/dashboard/client and server metadata wouldn't be applicable, right?

> But we still need a persisted user preference for the threshold. That could be `dashboard.threshold`, completely under the dashboard's control; but if so I wouldn't want the server to understand it.

I concur.

> If we want to entertain the option of, for example, sending warning emails, (and I kinda like that idea), I'd prefer to have it in an architected user metadata; whether as a first-class User property or as a UserMetadata key analogous to my dataset metadata.

I'm with you on the warning emails.  So, either the threshold needs to be (per-)dataset metadata or I think we take on some serious labor (unless the value is the same for all users and/or all datasets).

I can easily imagine a desire for having different datasets to have different expiration dates (and to have expiration dates which are independent of their creation/upload dates).  I'm having more trouble imagining that different datasets (owned by the same user) would need different warning thresholds.  To be honest, having different thresholds for different users seems like a capability that we probably don't need.  I think having a single threshold set in the server configuration for all users (perhaps with multiple warnings at "soon", "very soon", and "now") would be perfectly adequate.

lib/pbench/server/database/models/datasets.py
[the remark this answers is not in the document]
webbnh: I suppose we could create a new `user` namespace (with the understanding that the old one is changing its name to something to be determined when we actually implement it...).

Having an API to register/etc. new namespaces sounds cool, but I'm not sure it buys us anything over having clients just using their own node under `client`.

You said "dataset metadata namespaces"...I was thinking in terms of the values in the `client` namespace being tied to the logged in user and not to any specific dataset.  I'm not prepared to contemplate the possibilities of per-user/per-client metadata values being associated with individual datasets.  🤯  Is that where you are coming from?

portante (approved):

lib/pbench/server/api/resources/datasets_metadata.py
```diff
@@ -134,16 +137,35 @@ def _put(self, json_data: JSONOBJECT, _) -> Response:
...
         # of "owner" and "access".
-        self._check_authorization(str(dataset.owner_id), dataset.access)
+        #
+        # Note that we allow authenticated clients to set "user.*" namespace
+        # metadata with READ access to the dataset because it doesn't affect
+        # the Dataset itself and isn't visible to other users, but we require
+        # UPDATE access for any keys outside the "user" namespace.
+        role = API_OPERATION.READ
+        if not Auth.token_auth.current_user():
+            role = API_OPERATION.UPDATE
```
npalaska, line 147: Why can't the user who is authenticated have UPDATE` access?

npalaska (commented):
I have one question

lib/pbench/server/api/resources/datasets_metadata.py
dbutenhof: This is a bit complicated, and somewhat hacky. I could have tried to solve this by rewriting `_check_authorization` with more complicate contextual rules, but that seemed overkill and unjustified at this point. Basically the "hack" is that we're manipulating `role` here more to prohibit than to allow.

With the user-specific key mappings, as I tried to describe above, any *authenticated* user can create their own private `user.` metadata for any dataset they can `READ`. The authenticated *owner* of the dataset can write non-`user.` metadata as well. An unauthenticated user can do neither.

So:
* We want to validate authorized non-owners for `READ` access *if* they're *only* trying to modify `user.` keys;
* We want to validate unauthorized users for `UPDATE` because they *have* `READ` access to "public" datasets but still can't write even `user.` metadata;
* We want to validate authorized users for `UPDATE` if they're trying to set anything other than a `user.` key because only the owner can do that...

lib/pbench/server/database/models/datasets.py
```diff
@@ -980,6 +986,19 @@ def create(**kwargs) -> "Metadata":
...
+        """
+        Extract the root key name
+
+        Args:
+            key:    Key path (e.g., "user.tag")
+
+        Returns:
+            native SQL key name ("user")
+        """
+        return key.lower().split(".").pop(0)
```
webbnh, line 1000: Why `pop()`?  I think the result of `split()` is a temporary list...there's no need to modify it, is there?  I.e., why not just:
```
        return key.lower().split(".")[0]
```

lib/pbench/server/database/models/datasets.py
```diff
@@ -1030,6 +1049,7 @@ def getvalue(dataset: Dataset, key: str) -> JSON:
         Args:
             dataset: associated dataset
             key: hierarchical key path to fetch
+            user:
```
webbnh, line 1052: That's a rather limited description...but, really, need we say more?  😆

lib/pbench/server/database/models/datasets.py
```diff
@@ -1140,21 +1170,17 @@ def get(dataset: Dataset, key: str) -> "Metadata":
...
-                Database.db_session.query(Metadata)
-                .filter_by(dataset=dataset, key=key)
-                .first()
-            )
-        except SQLAlchemyError as e:
+            meta = __class__._query(dataset, key, user).first()
+        except SQLAlchemyError:
             Metadata.logger.exception("Can't get {}>>{} from DB", dataset, key)
-            raise MetadataSqlError("getting", dataset, key) from e
+            raise MetadataSqlError("getting", dataset, key)
```
webbnh, line 1176: Peter pointed out that there is a subtle difference in the resulting output of the traceback which justifies including the `from` keyword.  I'm not sure that it's worth adding the `as` just so that we can have the `from`, but I felt obliged to wave that flag (since he and I spent a lot of mana wrangling over it).

webbnh (approved):
My brain is filling up, so I'm going to go ahead and post this now.

If I hit any issues with the tests, I'll post a follow-up.

lib/pbench/test/unit/server/database/test_datasets_metadata_db.py
```diff
@@ -211,3 +257,23 @@ def test_delete_with_metadata(self, db_session, create_user):
...
+        Metadata.setvalue(
+            dataset=ds, key="user.contact", value="Fred", user=create_user
+        )
+        Metadata.setvalue(
+            dataset=ds, key="user.contact", value="Wilma", user=create_drb_user
+        )
+
+        assert Metadata.getvalue(dataset=ds, user=None, key="user") == {
+            "contact": "Barney"
+        }
```
webbnh, line 273: This "works" only because we're testing the primitives, here, right?  Normally, if `user` were `None` execution would never reach the `setvalue()` call, and so there would never be a value there for the `getvalue()` call to retrieve, right?  (It _might_ be worth a comment to that effect, in here somewhere.)

lib/pbench/test/unit/server/database/test_datasets_metadata_db.py
```diff
@@ -85,104 +85,150 @@ def test_metadata(self, db_session, create_user):
...
+        Metadata.remove(key="user", dataset=ds, user=create_user)
+        assert create_user.dataset_metadata == []
+        assert ds.metadatas == []
+
+        # Peek under the carpet to look for orphaned metadata objects linked
+        # to the Dataset
+        metadata = (
+            Database.db_session.query(Metadata).filter_by(dataset_ref=ds.id).first()
+        )
+        assert metadata is None
```
webbnh, line 165: Should we be doing the analogous checks of the users?

lib/pbench/server/database/models/datasets.py
```diff
@@ -857,8 +846,8 @@ class Metadata(Database.Base):
     # {"dashboard.seen": True}
     DASHBOARD = "dashboard.*"
 
-    # USER is arbitrary data saved on behalf of the owning user, as a JSON
-    # document.
+    # USER is arbitrary data saved with the dataset on behalf of an
+    # authenticated user, as a JSON document.
```
webbnh, line 850: You should probably update the example path at line 855 (and maybe 180, 181, 1010, 1043, 1044, and 1047).

lib/pbench/server/api/resources/datasets_metadata.py
```diff
@@ -134,16 +137,35 @@ def _put(self, json_data: JSONOBJECT, _) -> Response:
 
         # Validate the authenticated user's authorization for the combination
         # of "owner" and "access".
-        self._check_authorization(str(dataset.owner_id), dataset.access)
+        #
+        # Note that we allow authenticated clients to set "user.*" namespace
+        # metadata with READ access to the dataset because it doesn't affect
```
webbnh, line 142: You might want to update the example at lines 109-112 and 126.

webbnh (commented):
I found a few more things that you might want to consider (but, I also may have cast my net too wide...).

portante (approved):

riya-17 (approved):

lib/pbench/server/api/resources/__init__.py
```diff
@@ -1021,8 +1021,13 @@ def _get_dataset_metadata(
             elif i == Dataset.UPLOADED:
                 metadata[i] = UtcTimeHelper(dataset.uploaded).to_iso_string()
             elif Metadata.is_key_path(i, Metadata.USER_METADATA):
+                native_key = Metadata.get_native_key(i)
+                user = None
+                if native_key == Metadata.USER_NATIVE_KEY:
+                    authorized_user: Optional[User] = Auth.token_auth.current_user()
+                    user = authorized_user
```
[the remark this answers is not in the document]
dbutenhof: Except the `Optional` here was a type hint (which you dropped) clarifying that `Auth.token_auth.current_user()` could be `None`, which is still true ... or is that why you already resolved the conversation? 😁 

(Then again, it's a bit weird that I type-hinted `authorized_user` here but not `user`... or, in fact, even introduced `authorized_user` in the first place as I just unconditionally copy it to `user` anyway. 😕 Most likely there was some additional logic that got simplified away...)

lib/pbench/server/api/resources/datasets_metadata.py
dbutenhof: The example will still work just fine... although I suppose "contact information" isn't perhaps the best example for user-specific non-owner metadata. 😆

lib/pbench/server/database/models/datasets.py
```diff
@@ -821,34 +821,23 @@ class Metadata(Database.Base):
     # the "server" namespace, and are strictly controlled by keyword path:
     # e.g., "server.deleted", "server.archived";
     #
+    # Metadata keys within the "user" namespace allow exposure of user-specific
+    # configuration data through the Dataset metadata API; for example, a
+    # client can ask whether a specific dataset has been "favorited" by the
+    # authenticated user (TBD) and the allowable access settings (currently
+    # private and public).
```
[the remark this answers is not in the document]
dbutenhof: You mean that dangling final phrase probably referring to the `user.access` property that I dropped? Um... probably not?? 🦐

lib/pbench/server/database/models/datasets.py
```diff
@@ -870,8 +859,23 @@ class Metadata(Database.Base):
...
     #
-    # {"user.cloud": "AWS", "user.mood": "CLOUDY"}}
-    USER = "user.*"
+    # Metadata in the "user" namespace is user-specific; it will always be set
+    # and accessed using the authenticated user ID. That is, different users
+    # will see their own "user" namespace keys for any given dataset.
+    #
+    # TODO: ?? It occurs to me that we could allow "user" metadata rows without
+    # a user_id, which could act as a "default" for all users examining the
+    # dataset -- I'm not entirely sure this is a good model.
```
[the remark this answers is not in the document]
dbutenhof: Yeah, I should just delete that ...

lib/pbench/server/database/models/datasets.py
dbutenhof: Nothing semantically wrong with the statement at 855; although, as mentioned before, "contact information" probably isn't the most relevant example anymore, and maybe I'll change them to "user.favorite".

lib/pbench/server/database/models/datasets.py
dbutenhof: I moved the code; in the original context there was a reason, but the historical artifact has just been dragged along...

lib/pbench/server/database/models/datasets.py
dbutenhof: That's interesting since I don't believe I was a party to that discussion, and I know that my habit of dropping the `from` (designating, I see, an explicitly chained exception rather than an unintentional secondary exception during handling of a primary) came from your advocacy a while back.

But, yeah; I just (re-)read [PEP-3134](https://legacy.python.org/dev/peps/pep-3134/): I think I see the point, and I agree that `__cause__` would be a bit better than letting it default to `__context__` which implies to the traceback reader that the chain was unintended.

So, basically, we want to put back most of the `from` clauses ... 😬

lib/pbench/test/unit/server/database/test_datasets_metadata_db.py
dbutenhof: Yes, these are unit tests working below the level of user authentication; but that's kinda the point of unit tests. Testing of this interface has nothing to do with the semantics of higher-level interfaces, and do we really want a convention of documenting where they differ? Does it matter?

portante (approved):

lib/pbench/server/api/resources/__init__.py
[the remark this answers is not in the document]
webbnh: I pre-resolved the comment, in case you got another approval and decided to keep the extra local variables.  🙂

Once I'd done that, I opted to drop the type hint, since it didn't seem to provide much value that wasn't available from context.  🍤

lib/pbench/server/database/models/datasets.py
webbnh: Yeah.  Sorta.  Not really, but...yeah.  😏

webbnh (approved):
👍

PR #2828, opened 2022-05-10
Add authentication to /datasets/metadata

dbutenhof:
PBENCH-649

When I fixed the exception propagation for `/datasets/list`, I noticed that `/datasets/metadata` wasn't authorizing access at all, either for `GET` or `PUT`. This adds (and tests) authorization checks, which requires a new mechanism in `_check_authorization` to override the class default API role.

### DIFF
--- a/lib/pbench/server/api/resources/datasets_metadata.py
+++ b/lib/pbench/server/api/resources/datasets_metadata.py
@@ -16,7 +16,6 @@ from pbench.server.api.resources import (
 from pbench.server.database.models.datasets import (
     Dataset,
     DatasetError,
-    DatasetNotFound,
     Metadata,
     MetadataError,
 )
@@ -79,8 +78,7 @@ class DatasetsMetadata(ApiBase):
 
         try:
             dataset = Dataset.query(name=name)
-        except DatasetError as e:
-            self.logger.warning("Dataset {!r} not found: {}", name, str(e))
+        except DatasetError:
             raise APIAbort(HTTPStatus.BAD_REQUEST, f"Dataset {name!r} not found")
 
         # Validate the authenticated user's authorization for the combination
@@ -91,8 +89,6 @@ class DatasetsMetadata(ApiBase):
 
         try:
             metadata = self._get_dataset_metadata(dataset, keys)
-        except DatasetNotFound:
-            raise APIAbort(HTTPStatus.BAD_REQUEST, f"Dataset {name} not found")
         except MetadataError as e:
             raise APIAbort(HTTPStatus.BAD_REQUEST, str(e))
 


portante (dismissed):

npalaska (dismissed):
Interesting!

lib/pbench/server/api/resources/datasets_metadata.py
```diff
@@ -77,9 +77,20 @@ def _get(self, json_data: JSON, request: Request) -> Response:
...
+        # Validate the authenticated user's authorization for the combination
+        # of "owner" and "access".
+        self._check_authorization(
+            str(dataset.owner_id), dataset.access, check_role=API_OPERATION.READ
+        )
+
+        try:
+            metadata = self._get_dataset_metadata(dataset, keys)
         except DatasetNotFound:
             raise APIAbort(HTTPStatus.BAD_REQUEST, f"Dataset {name} not found")
```
webbnh, line 95: It would be nice if this message were (more) distinct from the message produced at line 84.  (Given that we should have aborted at line 84 if the dataset were not found, should this one be an internal error?

lib/pbench/server/api/resources/datasets_metadata.py
```diff
@@ -77,9 +77,20 @@ def _get(self, json_data: JSON, request: Request) -> Response:
         name = json.get("name")
         keys = json.get("metadata")
 
-        self.logger.info("GET metadata {} for {}", keys, name)
         try:
-            metadata = self._get_metadata(name, keys)
+            dataset = Dataset.query(name=name)
+        except DatasetError as e:
+            self.logger.warning("Dataset {!r} not found: {}", name, str(e))
+            raise APIAbort(HTTPStatus.BAD_REQUEST, f"Dataset {name!r} not found")
```
webbnh, line 84: Since this a 400-level error (i.e., a "user error"), are you sure we want to make a log entry for it?  If so, is there any other information that we want to log for context (like some sort of requestor ID information)?  (I note that we don't log anything anywhere else in this function.)

Ditto line 136.

webbnh (approved):
Looks good, but I have a coupe of questions.

lib/pbench/server/api/resources/datasets_metadata.py
dbutenhof: I think I cut and pasted that from somewhere else; we have logged a lot of client errors as "warning". It's a fair question to what extent we want to do that; it's mostly useful during debugging.

lib/pbench/server/api/resources/datasets_metadata.py
dbutenhof: That's a "cut and paste" error, more or less. The previous version called the "wrapper" metadata function which looked up the dataset. Now that it doesn't, I should probably just let this exception fly. OK, one more revision...

lib/pbench/server/api/resources/datasets_metadata.py
webbnh: > it's mostly useful during debugging.

Right...but, in that case, they should probably be logged as debug-level messages -- that way we can shut them off or easily delete them from the code.  (I would rather avoid having to do a wholesale pass through the code later, looking for log messages which should be removed for production.)

portante (approved):

dbutenhof:
Thanks, all; I was just about to call out for another review since I'm at the point where I need to merge this into my new `user.*` metadata API...

PR #2830, opened 2022-05-11
`pbench-specjbb2005` now handles commented props

portante:
If the `SPECjbb.props` file has a commented out line we now uncomment it to ensure the requested behaviors take effect.

### DIFF
--- a/agent/bench-scripts/pbench-specjbb2005
+++ b/agent/bench-scripts/pbench-specjbb2005
@@ -277,12 +277,12 @@ trap "interrupt" INT QUIT TERM
 
 # edit the jbb properties file to match user's options
 > ${benchmark_run_dir}/SPECjbb.props.sed
-printf "s/input.jvm_instances=.*/input.jvm_instances=$nr_jvms/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
-printf "s/input.measurement_seconds=.*/input.measurement_seconds=$runtime/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
-printf "s/input.starting_number_warehouses=.*/input.starting_number_warehouses=$start_warehouses/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
-printf "s/input.increment_number_warehouses=.*/input.increment_number_warehouses=$inc_warehouses/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
-printf "s/input.ending_number_warehouses=.*/input.ending_number_warehouses=$stop_warehouses/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
-sed -f ${benchmark_run_dir}/SPECjbb.props.sed ${specjbb2005_dir}/SPECjbb.props > ${benchmark_run_dir}/SPECjbb.props
+printf -- "s/#?input.jvm_instances=.*/input.jvm_instances=$nr_jvms/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
+printf -- "s/#?input.measurement_seconds=.*/input.measurement_seconds=$runtime/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
+printf -- "s/#?input.starting_number_warehouses=.*/input.starting_number_warehouses=$start_warehouses/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
+printf -- "s/#?input.increment_number_warehouses=.*/input.increment_number_warehouses=$inc_warehouses/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
+printf -- "s/#?input.ending_number_warehouses=.*/input.ending_number_warehouses=$stop_warehouses/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
+sed -E -f ${benchmark_run_dir}/SPECjbb.props.sed ${specjbb2005_dir}/SPECjbb.props > ${benchmark_run_dir}/SPECjbb.props
 if [ $? -ne 0 ]; then
 	printf "\t${benchmark}: unable to create the SPECjbb.props file (\"${benchmark_run_dir}/SPECjbb.props\")\n\n"
 	exit 1
--- a/agent/bench-scripts/pbench-specjbb2005
+++ b/agent/bench-scripts/pbench-specjbb2005
@@ -276,17 +276,16 @@ fi
 trap "interrupt" INT QUIT TERM
 
 # edit the jbb properties file to match user's options
-> ${benchmark_run_dir}/SPECjbb.props.sed
-printf -- "s/#?input.jvm_instances=.*/input.jvm_instances=$nr_jvms/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
-printf -- "s/#?input.measurement_seconds=.*/input.measurement_seconds=$runtime/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
-printf -- "s/#?input.starting_number_warehouses=.*/input.starting_number_warehouses=$start_warehouses/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
-printf -- "s/#?input.increment_number_warehouses=.*/input.increment_number_warehouses=$inc_warehouses/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
-printf -- "s/#?input.ending_number_warehouses=.*/input.ending_number_warehouses=$stop_warehouses/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
-sed -E -f ${benchmark_run_dir}/SPECjbb.props.sed ${specjbb2005_dir}/SPECjbb.props > ${benchmark_run_dir}/SPECjbb.props
+cp ${specjbb2005_dir}/SPECjbb.props ${benchmark_run_dir}/SPECjbb.props
 if [ $? -ne 0 ]; then
 	printf "\t${benchmark}: unable to create the SPECjbb.props file (\"${benchmark_run_dir}/SPECjbb.props\")\n\n"
 	exit 1
 fi
+printf "input.jvm_instances=$nr_jvms\n" >> ${benchmark_run_dir}/SPECjbb.props
+printf "input.measurement_seconds=$runtime\n" >> ${benchmark_run_dir}/SPECjbb.props
+printf "input.starting_number_warehouses=$start_warehouses\n" >> ${benchmark_run_dir}/SPECjbb.props
+printf "input.increment_number_warehouses=$inc_warehouses\n" >> ${benchmark_run_dir}/SPECjbb.props
+printf "input.ending_number_warehouses=$stop_warehouses\n" >> ${benchmark_run_dir}/SPECjbb.props
 cp ${specjbb2005_dir}/SPECjbb_config.props ${benchmark_run_dir}/
 if [ $? -ne 0 ]; then
 	printf "\t${benchmark}: unable to copy SPECjbb_config.props file from \"${specjbb2005_dir}\"\n\n"
--- a/agent/bench-scripts/unittests
+++ b/agent/bench-scripts/unittests
@@ -160,6 +160,7 @@ function _dump_logs {
     if [[ ! -z "${rundir}" ]]; then
         _dump_file ${rundir} metadata.log
         _dump_file ${rundir} user-benchmark-summary.json
+        _dump_file ${rundir} SPECjbb.props
         (
         if cd ${_testdir}/${rundir} > /dev/null 2>&1 ;then
             for f in $(find . -name '*.cmd' 2> /dev/null | sort); do
--- a/agent/bench-scripts/test-bin/java
+++ b/agent/bench-scripts/test-bin/java
@@ -22,16 +22,25 @@ nr_jvms = None
 start = None
 inc = None
 stop = None
+runtime = None
 with open("SPECjbb.props", "r") as fp:
     for line in fp:
+        if not line.startswith("input."):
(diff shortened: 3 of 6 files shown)


agent/bench-scripts/pbench-specjbb2005
```diff
@@ -277,12 +277,12 @@ trap "interrupt" INT QUIT TERM
 
 # edit the jbb properties file to match user's options
 > ${benchmark_run_dir}/SPECjbb.props.sed
-printf "s/input.jvm_instances=.*/input.jvm_instances=$nr_jvms/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
-printf "s/input.measurement_seconds=.*/input.measurement_seconds=$runtime/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
-printf "s/input.starting_number_warehouses=.*/input.starting_number_warehouses=$start_warehouses/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
-printf "s/input.increment_number_warehouses=.*/input.increment_number_warehouses=$inc_warehouses/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
-printf "s/input.ending_number_warehouses=.*/input.ending_number_warehouses=$stop_warehouses/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
-sed -f ${benchmark_run_dir}/SPECjbb.props.sed ${specjbb2005_dir}/SPECjbb.props > ${benchmark_run_dir}/SPECjbb.props
+printf -- "s/#?input.jvm_instances=.*/input.jvm_instances=$nr_jvms/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
```
dbutenhof, line 280: OK, what's the logic behind "if [the file] has a commented out line we now uncomment it"? Doesn't that make it pointless to comment a props line in the first place? Why would someone do that, and why do we want to ignore it if they do?

Granted, handling of this template props file is extremely opinionated, e.g., the `.*` instead of `_replace_`; but this particular transform seems a bit odd, and I'd love to see some rationale recorded in a comment or at least in the git log.

agent/bench-scripts/unittests
```diff
@@ -160,6 +160,7 @@ function _dump_logs {
     if [[ ! -z "${rundir}" ]]; then
         _dump_file ${rundir} metadata.log
         _dump_file ${rundir} user-benchmark-summary.json
+        _dump_file ${rundir} SPECjbb.props
```
dbutenhof, line 163: My first thought was "gee, why are we dumping the `props` file in `_dump_logs`"? My second was ... um, we're dumping *one* log file, a json file, an html file, and a set of cmd files, now plus props ... it's really that the function is misnamed. 😆

agent/bench-scripts/pbench-specjbb2005
webbnh: This only handles the case of a comment-introducer followed by no whitespace or any other character (e.g., it won't entirely uncomment a line prefixed with `##`)...is this a bug or a feature?

In any case, my reaction is similar to Dave's:  this idea seems highly questionable.

The clincher for me is this:  if there is _no entry_ for `input.jvm_instances` in the file, we **don't add one**...so why would we uncomment one that we find there?  ("Commented-out" should be the same as "not there" in functional terms, so the two cases **_should_** be handled similarly.)

webbnh (commented):
It seems like, if the target configurator is missing or commented-out, we should exit with an error rather than trying to uncomment it (which seems like it is fragile).  (This seems more in harmony with our recent efforts to avoid things like software installs and firewall re-configurations.)

(I'm neither "approving" nor "requesting changes"....)

agent/bench-scripts/pbench-specjbb2005
portante: The first question we might want to answer before this line of discussion is: what is the purpose of the `pbench-specjbb2005` bench script?  For that matter, what is the purpose of any of the CLI interfaces located in `bench-scripts`?

agent/bench-scripts/unittests
portante: Agreed.  Great task for another PR to solve!

portante:
> It seems like, if the target configurator is missing or commented-out, we should exit with an error rather than trying to uncomment it (which seems like it is fragile).

Certainly this method of uncommenting is fragile and could stand some improvements to make it robust.

> (This seems more in harmony with our recent efforts to avoid things like software installs and firewall re-configurations.)

Very different beasts entirely.  In the case of software installs and firewall re-configuration, the gratuitous change is made external to the responsibilities that the pbench-agent takes on (running benchmarks and collecting data).

The `sed` script is not changing an external file.  It is using the external file as input to create a file that is used to accomplish that which the person who invoked `pbench-specjbb2005` asked for by virtue of invoking the script.

> 
> (I'm neither "approving" nor "requesting changes"....)

Great!

agent/bench-scripts/pbench-specjbb2005
webbnh: It's a fair question, but take care that it doesn't lead us into a minefield.  😆 

I think your contention is that, if the user specified something like the number of JVMs on the command line, then that should override whatever they (or someone) put in the configuration file.  I don't disagree with that position, but, if that's the case, then `pbench-specjbb2005` should have a provision for _adding_ the `input.jvm_instances=` line to the file if it's not there.  (And, if we are going to do something like that, perhaps we can just append all the options that we want to the bottom and they will supersede any specifications earlier in the file, and we won't need to do stuff like trying to uncomment or otherwise edit the existing lines.)

But, if the purpose of the script is to allow the user to run the workload despite how s/he has configured it, what's so bad about it poking holes in the firewall?  Doing that is just "enabling the workload to run", as evidenced by the fact that it won't run without doing it, and, surely the user _wants_ us to do that, because he told us to run the workload, no?  🤣

portante: > It's a fair question, but take care that it doesn't lead us into a minefield. 😆

Exactly.  Which is why this is a small step and we can take more steps later in a proper benchmark workload runner script.

> I think your contention is that, if the user specified something like the number of JVMs on the command line, then that should override whatever they (or someone) put in the configuration file. I don't disagree with that position, but, if that's the case, then `pbench-specjbb2005` should have a provision for _adding_ the `input.jvm_instances=` line to the file if it's not there. (And, if we are going to do something like that, perhaps we can just append all the options that we want to the bottom and they will supersede any specifications earlier in the file, and we won't need to do stuff like trying to uncomment or otherwise edit the existing lines.)

Yes, but I am trying not to make huge changes.  This change represents work that was tried out on a broken system that fixed the problem and worked.

Future PRs can either make an improvement to this code, or move the support to a more proper benchmark workload runner script.

> But, if the purpose of the script is to allow the user to run the workload despite how s/he has configured it, what's so bad about it poking holes in the firewall?

It is not quite that.  The user is asking the script to make that configuration happen.

Poking holes in a firewall is a much bigger problem with wider implications that are not on the same level as how a benchmark is run.

Gratuitous software installs and firewall hole poking is not a role for the Pbench Agent and must be handled deliberately by the person or people responsible for that server's configuration.

> Doing that is just "enabling the workload to run", as evidenced by the fact that it won't run without doing it, and, surely the user _wants_ us to do that, because he told us to run the workload, no? 🤣

Ah ... no.  That is a contortion of reality.  =)

portante: Sure, that all sounds like good future work.

This is a really simple change that measurably improves this, does not cause a problem, and we can still iterate going forward.

I really want to avoid trying to get a single PR "perfect", but consider iterating across multiple PRs over time.

webbnh: It causes a problem because it only addresses the issue of commented-out config lines _some_ of the time.

This isn't about trying to get the PR to be perfect -- it's about deciding whether accepting the PR is better than not accepting it.

At the moment, we can state that commenting out lines in the config file causes problems and that the user can address those problems by un-commenting-out those lines.  If we accept this change, then we'll be in a position of saying that, if the user has commented out the lines in just the right way, our script will uncomment them, but otherwise things are just as broken as they ever were.

The value of that amount of improvement doesn't seem, to me, to offset the potential for resulting confusion...especially if there are better solutions within reach.

Can you point me to instructions on how to run SpecJBB (in particular, what I need to install and set up)?  I'd like to try adding overrides to the end of the config file to see if that will "just work".

portante: > It causes a problem because it only addresses the issue of commented-out config lines _some_ of the time.

How is that causing a problem?  The current code in 0.69 and in 0.71 does the same thing.  So this change, while is not complete, is not causing a bigger problem.  It is just not a complete solution.  It is incremental.  It is useful as demonstrated by unblocking a user of the interface.

Why would we not take the incremental change and focus on fixing the long term in the "next framework" instead of investing in this framework?

> This isn't about trying to get the PR to be perfect -- it's about deciding whether accepting the PR is better than not accepting it.

What is the argument that this helps the user by not accepting it?  There is nothing else in development that does this fix, and not likely to happen for a while given all the other priorities the project has.

> At the moment, we can state that commenting out lines in the config file causes problems and that the user can address those problems by un-commenting-out those lines. If we accept this change, then we'll be in a position of saying that, if the user has commented out the lines in just the right way, our script will uncomment them, but otherwise things are just as broken as they ever were.

Sure, but it does solve one case and it was demonstrated to be useful.  What is wrong with a small incremental change?

> The value of that amount of improvement doesn't seem, to me, to offset the potential for resulting confusion...especially if there are better solutions within reach.

Where is the confusion?  This patch when applied "live" solved a problem and allowed the user to move forward.

Are we really focusing on a real problem or one that just might happen in the future?  Let's invest our effort in a longer term solution for workloads rather than trying to polish this interface further.

We are trying to get 0.71 out the door.  This whole discussion has now potentially delayed the release 4 days.

> Can you point me to instructions on how to run SpecJBB (in particular, what I need to install and set up)? I'd like to try adding overrides to the end of the config file to see if that will "just work".

I don't know where they live in the wild.

webbnh: Please see [portante/pbench #20](https://github.com/portante/pbench/pull/20) for a better solution.

portante: Thanks for the alternative solution!  Along with working unit tests, I hope it works.

portante (commented):
Ready for review.

agent/bench-scripts/pbench-specjbb2005
```diff
@@ -276,17 +276,16 @@ fi
...
-printf -- "s/#?input.starting_number_warehouses=.*/input.starting_number_warehouses=$start_warehouses/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
-printf -- "s/#?input.increment_number_warehouses=.*/input.increment_number_warehouses=$inc_warehouses/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
-printf -- "s/#?input.ending_number_warehouses=.*/input.ending_number_warehouses=$stop_warehouses/\n" >> ${benchmark_run_dir}/SPECjbb.props.sed
-sed -E -f ${benchmark_run_dir}/SPECjbb.props.sed ${specjbb2005_dir}/SPECjbb.props > ${benchmark_run_dir}/SPECjbb.props
+cp ${specjbb2005_dir}/SPECjbb.props ${benchmark_run_dir}/SPECjbb.props
 if [ $? -ne 0 ]; then
 	printf "\t${benchmark}: unable to create the SPECjbb.props file (\"${benchmark_run_dir}/SPECjbb.props\")\n\n"
 	exit 1
 fi
+printf "input.jvm_instances=$nr_jvms\n" >> ${benchmark_run_dir}/SPECjbb.props
```
dbutenhof, line 284: So I assume "we're confident" that leaving current (uncommented) lines in the default alone and overriding them with new lines at the end actually works correctly? Our test scripts use `_replace_` or `_bad_` and I guess I never really thought about what was in the real ones as the replacement was pretty generic.

dbutenhof (approved):

agent/bench-scripts/pbench-specjbb2005
portante: This change came from https://github.com/portante/pbench/pull/20.  I did not personally verify this change, so I don't have confidence in it.  But if somebody can provide the confidence, great.

agent/bench-scripts/pbench-specjbb2005
webbnh: I'm not sure how we would unit test this, given that our existing unit tests stub-out the thing which actually reads the configuration file....  Got any thoughts on that?

agent/bench-scripts/pbench-specjbb2005
webbnh: I personally verified this change, by installing SpecJBB2005 and running it with `pbench-specjbb2005`, using the default (and problematical) config file, with several sets of command line switches, and the config lines which the script appends to the config file produce the desired effects, including overriding each of commented-out and non-commented-out config parameters.

webbnh (approved):
Looks good to me!  (Even if I did "write" half of it.  😉)

agent/bench-scripts/unittests
dbutenhof: Then again, we want to "dump" the legacy tests entirely, and "this, too, shall pass" so maybe it's not worth worrying about too much. It was, however, an amusing observation I couldn't help sharing ... but will now quietly mark resolved. 😁

agent/bench-scripts/pbench-specjbb2005
webbnh: Let me amend...do we need anything beyond the improvements to the unit testing which this PR delivers?

PR #2831, opened 2022-05-11
Correct `pbench-specjbb2005` use of `-send-tools`

portante:
The `pbench-send-tools` command must use the directory that was used for starting/stopping tools.


webbnh (approved):
I suspect that there are some subtleties here that I'm not appreciating, but I guess I have no problems with this change.

portante:
> I suspect that there are some subtleties here that I'm not appreciating

The Tool Meister sub-system requires that the same target directory for `start`, `stop`, and `send` be used.

For SpecJBB2005, we use tool triggers, where the target directories for the start and stop operations are generated by `pbench-tool-trigger`. That script uses a simple [monotonically incrementing number in an iteration directory](https://github.com/distributed-system-analysis/pbench/blob/main/agent/bench-scripts/pbench-specjbb2005#L328), `%d-default`, where it [appends](https://github.com/distributed-system-analysis/pbench/blob/main/agent/util-scripts/pbench-tool-trigger#L64) the "sample" directory with the requested sample number [1].  For `pbench-specjbb2005`, it is always `1`.

This means that the `start` and `stop` operations issued by `pbench-tool-trigger` use one directory name, and `pbench-specjbb2005` invokes the `send` operation using a "renamed" directory.

This fix moves the `pbench-send-tools` call up and changes the directory name to be the expected `%d-default/sample1`.

webbnh:
> I don't know, off-hand, if the post-processed tool results need to work off of the final name or is independent.

I don't either, which is why I was gently inquiring....  😁

> I know that we don't want to record the generic tool trigger iteration names

I was concerned about that, as well.

> why do we need to consider changing the rest of the existing logic?

"Continuous improvement?"  😇

portante (commented):
Ready for review.

dbutenhof (approved):

ndokos (approved):

webbnh (approved):
Still looks good.

PR #2833, opened 2022-05-12
Move towards embedded URL parameters: metadata

dbutenhof:
PBENCH-669

We started out with all parameters in a JSON request payload to `POST` operations following Elasticsearch conventions. While some new APIs have used both query parameters and embedded Flask URL parameters, there are older APIs that should be retrofitted to that model.

In this PR, I'm addressing `/datasets/metadata` (which seems like a good place to start as the Overview page is on the cusp of using it).

In particular, we'll now `GET`|`PUT` `http://<server>/api/v1/datasets/metadata/<name>`. rather than specifying the dataset name with `?name=<name>`.

### DIFF
--- a/lib/pbench/test/unit/server/test_datasets_metadata.py
+++ b/lib/pbench/test/unit/server/test_datasets_metadata.py
@@ -218,7 +218,9 @@ class TestDatasetsMetadata:
         Note that Pbench will silently ignore any additional keys that are
         specified but not required.
         """
-        response = client.put(f"{server_config.rest_uri}/datasets/metadata/drb")
+        response = client.put(
+            f"{server_config.rest_uri}/datasets/metadata/drb", json={}
+        )
         assert response.status_code == HTTPStatus.BAD_REQUEST
         assert response.json.get("message") == "Missing required parameters: metadata"
 
--- a/lib/pbench/test/unit/server/test_datasets_metadata.py
+++ b/lib/pbench/test/unit/server/test_datasets_metadata.py
@@ -203,12 +203,14 @@ class TestDatasetsMetadata:
         )
         assert response.json == {"message": "Unknown URL query keys: controller,plugh"}
 
-    def test_put_missing_json_object(self, client, server_config, pbench_token):
+    @pytest.mark.parametrize(
+        "uri", ("/datasets/metadata/", "/datasets/metadata"))
+    def test_put_missing_uri_param(self, client, server_config, pbench_token, uri):
         """
         Test behavior when no dataset name is given on the URI. (NOTE that
         Flask automatically handles this with a NOT_FOUND response.)
         """
-        response = client.put(f"{server_config.rest_uri}/datasets/metadata/")
+        response = client.put(f"{server_config.rest_uri}{uri}")
         assert response.status_code == HTTPStatus.NOT_FOUND
 
     def test_put_missing_key(self, client, server_config, pbench_token):
--- a/lib/pbench/test/unit/server/test_datasets_metadata.py
+++ b/lib/pbench/test/unit/server/test_datasets_metadata.py
@@ -203,8 +203,7 @@ class TestDatasetsMetadata:
         )
         assert response.json == {"message": "Unknown URL query keys: controller,plugh"}
 
-    @pytest.mark.parametrize(
-        "uri", ("/datasets/metadata/", "/datasets/metadata"))
+    @pytest.mark.parametrize("uri", ("/datasets/metadata/", "/datasets/metadata"))
     def test_put_missing_uri_param(self, client, server_config, pbench_token, uri):
         """
         Test behavior when no dataset name is given on the URI. (NOTE that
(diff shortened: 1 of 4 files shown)


portante (approved):

lib/pbench/test/unit/server/test_datasets_metadata.py
```diff
@@ -214,94 +205,38 @@ def test_get_bad_query_2(self, query_get_as):
...
-
-        # Test in case all of the required keys are missing and some
-        # random non-existent key is present in the payload
-        if required_keys:
-            missing_key_helper({"notakey": None})
+        response = client.put(
+            f"{server_config.rest_uri}/datasets/metadata/drb", json={}
+        )
+        assert response.status_code == HTTPStatus.BAD_REQUEST
+        assert response.json.get("message") == "Missing required parameters: metadata"
```
webbnh, line 225: Does that message ("Missing required parameters: metadata") make sense to you in this context?  The request provided the metadata _resource_ ("drb"), but it didn't provide any metadata _key/value pairs_...so the token "metadata" seems, ambiguous at best (but, I don't know what to suggest, so I hope it's good enough as is...).

And, it's it really a bad request to ask to change nothing?

lib/pbench/test/unit/server/test_datasets_metadata.py
```diff
@@ -214,94 +205,38 @@ def test_get_bad_query_2(self, query_get_as):
...
-        ]
-
-        all_combinations = []
-        for r in range(1, len(parameter_items) + 1):
-            for item in itertools.combinations(parameter_items, r):
-                tmp_req_keys = [key for key, parameter in item if parameter.required]
-                if tmp_req_keys != required_keys:
-                    all_combinations.append(item)
-
-        for items in all_combinations:
```
webbnh, line 261: Is this test no longer applicable or relevant?

webbnh (dismissed):
Looks nice; I've just got a few questions.

lib/pbench/test/unit/server/test_datasets_metadata.py
```diff
@@ -214,94 +205,38 @@ def test_get_bad_query_2(self, query_get_as):
...
     def test_put_missing_json_object(self, client, server_config, pbench_token):
         """
-        Test behavior when no JSON payload is given
+        Test behavior when no dataset name is given on the URI. (NOTE that
+        Flask automatically handles this with a NOT_FOUND response.)
         """
-        response = client.put(f"{server_config.rest_uri}/datasets/metadata")
-        assert response.status_code == HTTPStatus.BAD_REQUEST
-        assert response.json.get("message") == "Invalid request payload"
+        response = client.put(f"{server_config.rest_uri}/datasets/metadata/")
```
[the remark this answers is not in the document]
dbutenhof: I guess that's a good question, since I had to experiment to answer it; Flask's mapping can be quirky.

In this case, however, it seems to have exactly the same behavior: Flask throws a NOT_FOUND before we get control, as in, "that URL doesn't connect to anything".

I guess I might as well push my new test case...

lib/pbench/test/unit/server/test_datasets_metadata.py
dbutenhof: Hmm. The URL is saying we're getting metadata for Dataset `drb`. So the "dataset" parameter is present, but the "metadata" parameter (which we expect to find in the JSON body) is missing.

There are probably several ways we could in theory be more specific. I'm working right now on generalizing and encapsulating the schema validation to understand the differences between `GET` and `PUT`, and between "URI template parameters" (dataset in this case), "URI query parameters" (`?metadata` on the `GET`) and "JSON body keys" (e.g., `metadata` here).

But I'm not sure I want to even try to parameterize the low-level schema validators enough to clarify the context of each error... it'd mean catching and re-writing the low-level validation errors at the higher level. It'd be more clear; is it worth the cost and complexity? My instinct is no since this is "developer" interfaces not "end user" interfaces... but the line can be thin.

In any case, I definitely don't want to attack that morass in *this* PR, so for now I'm good with this and I assume your approval means that you can live with it, too.

lib/pbench/test/unit/server/test_datasets_metadata.py
dbutenhof: It was very complicated and generalized schema validation copied from `Common`; and while it made only a tiny bit of sense before when "dataset" and "metadata" were in the same space, I decided to simplify and just do them separately.

I.e., I added the `/datasets/metadata/` (and now `/datasets/metadata`) above, and this now just checks the missing `metadata` key.

webbnh (dismissed):
👍

lib/pbench/test/unit/server/test_datasets_metadata.py
webbnh: As long as we haven't lost any coverage....

webbnh (approved):
👍

portante (approved):

PR #2834, opened 2022-05-12
Remove minor redundancy in datasets mappings

npalaska:
Remove bit of redundancy in `dataset_mappings` for getting ES mapping properties.

### DIFF
--- a/lib/pbench/server/api/resources/query_apis/datasets/datasets_mappings.py
+++ b/lib/pbench/server/api/resources/query_apis/datasets/datasets_mappings.py
@@ -115,6 +115,7 @@ class DatasetsMappings(ApiBase):
             return jsonify(result)
         except TemplateNotFound:
             self.logger.exception(
-                "Document template {} not found in the database.", index["index"]
+                "Document template for index {} not found in the database",
+                index["index"],
             )
             raise APIAbort(HTTPStatus.INTERNAL_SERVER_ERROR)
--- a/lib/pbench/server/api/resources/query_apis/datasets/datasets_mappings.py
+++ b/lib/pbench/server/api/resources/query_apis/datasets/datasets_mappings.py
@@ -113,9 +113,6 @@ class DatasetsMappings(ApiBase):
 
             # construct response object
             return jsonify(result)
-        except TemplateNotFound:
-            self.logger.exception(
-                "Document template for index {} not found in the database",
-                index["index"],
-            )
+        except TemplateNotFound as e:
+            self.logger.exception(str(e))
             raise APIAbort(HTTPStatus.INTERNAL_SERVER_ERROR)
--- a/lib/pbench/server/api/resources/query_apis/datasets/datasets_mappings.py
+++ b/lib/pbench/server/api/resources/query_apis/datasets/datasets_mappings.py
@@ -15,7 +15,7 @@ from pbench.server.api.resources import (
 )
 
 from pbench.server.api.resources.query_apis.datasets import IndexMapBase
-from pbench.server.database.models.template import Template, TemplateNotFound
+from pbench.server.database.models.template import TemplateNotFound
 
 
 class DatasetsMappings(ApiBase):
@@ -103,10 +103,7 @@ class DatasetsMappings(ApiBase):
 
         index = IndexMapBase.ES_INTERNAL_INDEX_NAMES[json_data["dataset_view"]]
         try:
-            index_name = index["index"]
-            template = Template.find(index_name)
-            mappings = template.mappings
-
+            mappings = IndexMapBase.get_mappings(index)
             result = {}
             for property in mappings["properties"]:
                 if "properties" in mappings["properties"][property]:
@@ -118,6 +115,6 @@ class DatasetsMappings(ApiBase):
             return jsonify(result)
         except TemplateNotFound:
             self.logger.exception(
-                "Document template {} not found in the database.", index_name
+                "Document template {} not found in the database.", index["index"]
             )
             raise APIAbort(HTTPStatus.INTERNAL_SERVER_ERROR)
--- a/lib/pbench/test/unit/server/query_apis/test_datasets_mappings.py
+++ b/lib/pbench/test/unit/server/query_apis/test_datasets_mappings.py
@@ -1,20 +1,5 @@
 from http import HTTPStatus
 
-import pytest
-from sqlalchemy.exc import DatabaseError
-
-from pbench.server.database.models.template import Template
-
-
-@pytest.fixture()
-def database_error(monkeypatch):
-    def raise_db_error(name: str):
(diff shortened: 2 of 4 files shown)


lib/pbench/server/api/resources/query_apis/datasets/datasets_mappings.py
```diff
@@ -103,10 +103,7 @@ def _get(self, json_data: JSON, request: Request) -> Response:
 
         index = IndexMapBase.ES_INTERNAL_INDEX_NAMES[json_data["dataset_view"]]
         try:
-            index_name = index["index"]
-            template = Template.find(index_name)
-            mappings = template.mappings
-
+            mappings = IndexMapBase.get_mappings(index)
```
dbutenhof, line 106: Why did you make this `@staticmethod`? Granted, there's no reason it couldn't be. But either way you should be able to `self.get_mappings(index)`.

dbutenhof (approved):

lib/pbench/server/api/resources/query_apis/datasets/datasets_mappings.py
npalaska: That's because `DatasetsMappings` does not inherit from `IndexMapBase` so we can not use `self` here. `DatasetMappings` does not have a `name` as a required schema parameter which is why it can not inherit from `IndexMapBase`

lib/pbench/server/api/resources/query_apis/datasets/datasets_mappings.py
```diff
@@ -118,6 +115,6 @@ def _get(self, json_data: JSON, request: Request) -> Response:
             return jsonify(result)
         except TemplateNotFound:
             self.logger.exception(
-                "Document template {} not found in the database.", index_name
+                "Document template {} not found in the database.", index["index"]
```
portante, line 118: Since we are making a change here, can we perhaps update the exception string to add clarity?  Perhaps something like, "Document template for index {} not found in the database"?

portante (dismissed):

dbutenhof (dismissed):

lib/pbench/server/api/resources/query_apis/datasets/datasets_mappings.py
webbnh: My IDE seems to prefer static methods wherever a method doesn't reference `self`.  I don't know whether they are really "better", or whether it's some sort of "best practice", or whether it's just suggesting it because it can....

lib/pbench/server/api/resources/query_apis/datasets/datasets_mappings.py
```diff
@@ -118,6 +115,7 @@ def _get(self, json_data: JSON, request: Request) -> Response:
             return jsonify(result)
         except TemplateNotFound:
             self.logger.exception(
-                "Document template {} not found in the database.", index_name
+                "Document template for index {} not found in the database",
+                index["index"],
             )
```
webbnh, line 120: It's too bad that we have to compose our own message text here instead of just using the `str()` of the `TemplateNotFound` exception object....

lib/pbench/test/unit/server/query_apis/test_datasets_mappings.py
```diff
@@ -1,20 +1,5 @@
...
-
-
-@pytest.fixture()
-def database_error(monkeypatch):
-    def raise_db_error(name: str):
-        raise DatabaseError("DB Error")
-
-    with monkeypatch.context() as m:
-        m.setattr(Template, "find", raise_db_error)
-        yield
```
webbnh, line 16: Why isn't this mock needed anymore?

webbnh (changes requested):
The change to the code seems like a good catch; but I don't see why it enables us to remove a mock from the unit test.

lib/pbench/test/unit/server/query_apis/test_datasets_mappings.py
npalaska: It is because if the `find` can not talk to the database its anyway gonna result in some sort of sqlalchemy database error and we catch it with broad `SQLAlchemyError` so we don't have to raise it explicitly

lib/pbench/server/api/resources/query_apis/datasets/datasets_mappings.py
npalaska: Yeah that's a good point actually.

dbutenhof (approved):

lib/pbench/test/unit/server/query_apis/test_datasets_mappings.py
webbnh: So...we're counting on the environment, by default, aligning so that the database happens to be inaccessible?

Wouldn't it be better to establish the behavior that we want the CUT to encounter?  That way, there is no question what `find()` should see during the test, and therefore no question how the CUT should behave.

webbnh (approved):
Looks good, but I think we should be using a mock to ensure the behavior that we want, rather than depending on the default behavior of the underlying database component.

lib/pbench/test/unit/server/query_apis/test_datasets_mappings.py
npalaska: Yeah I am little confused now tbh, my thinking behind this was to make sure we catch any exception that can be raised (since we are catching the broad `SQLAlchemyError` exception) and, by default, its raising the database error so no need for this fixture. But you have a point, this fixture would establish the fixed behavior that our CUT will encounter.

webbnh: > I am little confused now tbh

My bad.  I'm sorry about that.

Where we should be (based on the changes you've made to the CUT) is that we should have a unit test for `DatasetsMappings._get()` which uses a mocked `IndexMapBase.get_mappings()`, which does or does not (depending on the scenario) raise `TemplateNotFound`.  And, separately, we should have a unit test for `IndexMapBase.get_mappings()`, which uses a mocked `Template.find()` to drive its behaviors.  (And, somewhere in there, we should probably have an assertion that checks that the stringization of `TemplateNotFound` is satisfactory.)

However, this part of the code has typically been tested more functionally, rather than in discrete units.  So, I'm guessing that that is where our respective mindsets may not be fitting together very well.

Previous to your change, it looks like there were two tests for the `/datasets/mappings/summary` URI, one with a `find()` mock which returns lots of data, and one with a `find()` mock which raised an exception.  Your change removes the latter mock, depending instead on the exception being raised more or less by default or coincidence.

Your assertion that we don't need this latter mock is technically correct, but it leads to a fragile and somewhat confusing test regimen.  So, we really should have this mock, so that we can control the behavior that the CUT encounters, so that we can have confidence that the test produces the appropriate result.

portante (approved):

Issue #2601, opened 2022-01-18
Consider kicking `controller` out of `Dataset` SQL table


dbutenhof:
Do we want to kick the "controller" column out of the Dataset record and into metadata? I'm considering collecting all of the `metadata.log` values into metadata for easy reference, and perhaps controller is just a part  of that now? The main issue is that figuring out how to integrate the  JSON Metadata table into a Dataset query (via a join or sub-select?) will be somewhat complicated.

Ideally, we'd be able to select from any `Metadata` for queries (especially `/datasets/list` which may eventually become a `GET /datasets` collection query in API V2). That probably requires a join between `Datasets` and `Metadata`, which is complicated by figuring out JSON hierarchy queries through SQLAlchemy. (PostgreSQL can do that; it's less clear that SQLAlchemy can.)

dbutenhof:
Resolved by #2835

PR #2835, opened 2022-05-13
Drop `controller` from primary Dataset SQL table

dbutenhof:
PBENCH-445

This completely removes `controller` from the `Dataset` DB schema.

I've been talking about this a long time; after pushing PR #2826 which will require rebuilding the DB, I figured I might as well try to take this additional step at the same time rather than doing it twice. Not that I expect this will be the last DB schema change...

There are a lot of changed files; most of the changes (including the legacy test gold files) are small and straightforward. I was able to fix all but one gold file -- where the loss of "controller" changed the `psql` dump sort order -- with a simple `sed`. (Beware of some of the gold file diffs, though; a few have enormous `index-map` metadata.)

FYI: note that I moved `Tarball.stem` to `Dataset.stem` to avoid an import circularity. I suspect that the `utils.quarantine` function ought to be a `FileTree` method if we can't get rid of it entirely, but I stopped short of attempting that in this PR.


lib/pbench/server/utils.py
```diff
@@ -82,7 +83,8 @@ def quarantine(dest, logger, *files):
             # state. (If it's the associated MD5 file, skip that.)
             if str(afile).endswith(".tar.xz"):
                 try:
-                    Dataset.attach(path=afile, state=States.QUARANTINED)
+                    ds_name = os.path.basename(afile)[: -len(".tar.xz")]
+                    Dataset.attach(name=ds_name, state=States.QUARANTINED)
```
portante, line 87: Would this be a good place to use `Dataset.stem()`?

portante (commented):
Initial comment while waiting for more changes to be posted.  Should this be a `Draft` PR for now?

lib/pbench/server/utils.py
dbutenhof: This was actually where I started dealing with the circular dependency problems that caused me to move it from `Tarball` under `filetree.py` into `Dataset`. And since we're already using `Dataset` here that'd work ... but apparently I missed circling back here.

portante: Why was the circularity solved by moving `.stem()` to `Dataset` instead of moving `TARBALL_SUFFIX` to `filetree.py`?

dbutenhof: The problems come from the fact that `FileTree` is relatively "high level" whereas `Dataset` is "lower" level and in particular is already a requirement for any code dealing with datasets.

dbutenhof: There's a `Tarball` object, and that's why I'd put the suffix handling in `file_tree.py` when I created it. Indeed, at some point it may make sense to pull `Tarball` out into a separate file, which would simplify some of the dependencies. I'm not sure whether that alone would be sufficient to resolve the immediate problem, and I decided not to try within this PR.

portante: Why not make that change independently in a separate PR, get it merged, and then land this one?

dbutenhof: Because I don't see much value in refactoring the file tree until we know exactly what we want for object store, and it's completely incidental and nearly completely orthogonal to this PR?

portante: Except that why does the `Dataset` class have knowledge about the on-disk of the tar ball when the `Tarball` class doesn't.

portante: PRs are getting hammered about technical debt of late.  Isn't this technical debt?

dbutenhof: This isn't "on-disk" structure, though; it's the iconic "agent API convention" of our "data bundle" naming. That's independently enforced at both the agent and server end and retained through artifact access, currently through Apache file sharing and, presumably, by our future internally hosted artifact access when we compute the response header along with shipping off the tarball byte stream.

If we want to change the tarball naming conventions in the future, that's unrelated to the planned transition from `/srv/pbench/archive/<controller>` directory hierarchy to an S3 object URI. So, no, I don't see this as "technical debt" even if in the future we devise a new plan to change this.

It's true that the `Dataset` class is mostly unconcerned with the naming and structural conventions, which was why I'd moved this out to `Tarball` when I created the file tree. I'd (at least slightly) prefer it there, but I don't see it as a big enough architectural issue at this point to mangle the file tree code around before we get serious about object store and know where the refactoring arrow will point. (Since I'm hoping we'll get to that "soon" and it will change a lot, I'd view partial refactoring now as a distraction.)

portante: That is certainly one point of view.  It is not worth arguing about.  We'll just change this later.

portante (dismissed):
Looks okay to move forward, even with the `Dataset.stem()` oddness.  I did not review the large gold file diffs, however.

riya-17 (dismissed):
Looks Good :)

lib/pbench/server/api/resources/upload_api.py
```diff
@@ -168,7 +166,6 @@ def put(self, filename: str):
                     self.logger.error(
                         "Duplicate dataset {}:{}:{} not found",
```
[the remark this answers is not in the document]
dbutenhof: Ouch; missed that entirely! Thanks. I'm working on rebasing now, so I'll fix that...

portante (approved):

lib/pbench/server/database/models/datasets.py
```diff
@@ -24,13 +23,25 @@ class DatasetError(Exception):
...
 class DatasetSqlError(DatasetError):
     """
     SQLAlchemy errors reported through Dataset operations.
 
-    The exception will identify the controller and name of the target dataset,
-    along with the operation being attempted; the __cause__ will specify the
-    original SQLAlchemy exception.
+    The exception will identify the name of the target dataset, along with the
+    operation being attempted; the __cause__ will specify the original
+    SQLAlchemy exception.
```
webbnh, line 44: While I believe that what you say here is true, I don't see how the code in this class causes that to happen.  (That is, I think nearly all of that is effected by the code which instantiates this object, not by the methods of the object itself.)

_[I understand that all that this PR does is remove two words and re-wrap the rest, so if you're not inclined to address this, we can move along.]_

lib/pbench/server/database/models/datasets.py
```diff
@@ -467,6 +473,31 @@ def is_tarball(path: Union[Path, str]) -> bool:
...
+        this just checks for the expected 7 character suffix and strips it.
+
+        If the path does not end in ".tar.xz" then the full path.name is
+        returned.
+
+        Args:
+            path: A file path that might be a Pbench tarball
+
+        Raises:
+            BadFilename: the path name does not end in TARBALL_SUFFIX
```
webbnh, line 490: s/BadFilename/DatasetBadName/

lib/pbench/server/indexing_tarballs.py
```diff
@@ -329,7 +329,7 @@ def sighup_handler(*args):
 
                             try:
                                 dataset = Dataset.attach(
-                                    path=path, state=States.INDEXING
+                                    name=Dataset.stem(path), state=States.INDEXING
```
webbnh, line 332: I'm having trouble connecting all the dots...prior to this change, I _think_ the "name" of a dataset was only the stem of the filename; but I think that `Dataset.stem(path)` removes only the suffix and not the parent directory path...is that a problem?

webbnh (approved):
Looks generally good.  I found one "typo" to fix, and I have one concern (but maybe I'm just confused), and there are a couple of other items for your consideration.

lib/pbench/server/utils.py
webbnh: Since most of the infrastructure for `stem()` was already in the `Dataset` class, moving the function there (at least temporarily) makes reasonable sense.  However, it doesn't look to me like `stem()`, or `is_tarball()`, or `TARBALL_SUFFIX` has any real relationship to that class, so perhaps they should (eventually) be hosted somewhere else (and, so, `DatasetBadName` should be disentangled from it, as well).

While these four pieces fit together, I don't see any particular requirement for them to be packaged together in a class.  They could just as well be module-level definitions in some suitable utility module or other place.  I don't know what the details of the circularity problem was, but prying them out of any class altogether would likely help.

npalaska (approved):
I think this looks good.

riya-17 (approved):

lib/pbench/server/api/resources/datasets_list.py
```diff
@@ -105,7 +101,6 @@ def _get(self, json_data: JSON, request: Request) -> Response:
         for dataset in results:
             d = {
                 "name": dataset.name,
-                "controller": dataset.controller,
                 "run_id": dataset.md5,
             }
             try:
```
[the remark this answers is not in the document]
dbutenhof: This is not a definitive permanent statement of what the base object returned from `/datasets/list` ought to be, (there's a separate Jira ticket for that), but rather just a statement that we don't want `controller` here.

I don't think `run_id` makes much sense here, either; especially now that it's no longer the identifying key for the secondary index queries. Perhaps all we want is name and requested metadata ... which can include owner, access, creation date, and (once I get around to it after this) anything that's in `metadata.log`.

lib/pbench/server/database/models/datasets.py
dbutenhof: True; we expect it to be raised in the context of a SQLAlchemy exception, but there's nothing this exception class can do about that. It's not without value, though it probably doesn't really belong here. Given the weight of approvals I'd just as soon "move along".

lib/pbench/server/indexing_tarballs.py
dbutenhof: The `Dataset` name is the `Path`-speak "stem", which is `.name` - `.suffix`. But the `Path.stem` only removes a single suffix, and our standard tarball format has two.

So `Dataset.stem` removes the `.tar.xz` suffix from the `.name` of the tarball `Path`... e.g., "xyzzy" from `/var/tmp/xyzzy.tar.xz`.

lib/pbench/server/database/models/datasets.py
dbutenhof: That's up to you; if you don't resolve the conversation soon, I guess I will, and I can rebase manually and such as well. I'd simply preferred not to wait for another round of approvals based on this minor issue.

lib/pbench/server/indexing_tarballs.py
webbnh: OK, nevermind:  I finally tumbled to the fact that in `stem()` we're _also_ using `.name`.  We're all good here.

portante:
@dbutenhof, does this close #2601?

PR #2851, opened 2022-05-16
Unit test for browsing page

shubham-html-css-js:
Configured Package.json to support unit testing without interfering with Patternfly icon import and wrote unit tests for some browsing page components

### DIFF
--- a/dashboard/src/modules/components/DatePickerComponent/DatePicker.test.js
+++ b/dashboard/src/modules/components/DatePickerComponent/DatePicker.test.js
@@ -21,10 +21,10 @@ jest.mock("utils/api", () => {
 test("data is filtered based on date range selected from date picker", async () => {
   render(<AppWrapper />);
   await screen.findByText("pbench_user_benchmark1");
-  const datePickerInput = screen.getAllByPlaceholderText(/yyyy-mm-dd/i);
+  const datePickerInput = screen.getAllByPlaceholderText('YYYY-MM-DD');
   fireEvent.change(datePickerInput[0], { target: { value: "2022-02-16" } });
   fireEvent.change(datePickerInput[1], { target: { value: "2022-02-20" } });
-  const updateBtn = screen.getByRole("button", { name: /update/i });
+  const updateBtn = screen.getByRole("button", { name: 'Update'});
   fireEvent.click(updateBtn);
   const datasetNameOne = screen.queryByText("pbench_user_benchmark1");
   const datasetNameTwo = screen.queryByText("pbench_user_benchmark2");
--- a/dashboard/src/modules/components/TableComponent/Table.test.js
+++ b/dashboard/src/modules/components/TableComponent/Table.test.js
@@ -21,7 +21,7 @@ jest.mock("utils/api", () => {
 test("Page heading is displayed on initial load", async () => {
   render(<AppWrapper />);
   await screen.findByText("pbench_user_benchmark1");
-  const heading = screen.getByRole("heading", { name: /results/i });
+  const heading = screen.getByRole("heading", { name: 'Results' });
   expect(heading).toBeInTheDocument();
 });
 test("data from API is displayed on initial load", async () => {
@@ -43,12 +43,12 @@ test("row is favorited after clicking on favorite icon", async () => {
   render(<AppWrapper />);
   await screen.findByText("pbench_user_benchmark1");
   const starBtn = screen.getAllByRole("button", {
-    name: /not starred/i,
+    name: 'Not starred',
   });
   fireEvent.click(starBtn[0]);
   fireEvent.click(starBtn[1]);
   const favoriteBtn = screen.getByRole("button", {
-    name: /see favorites button/i,
+    name: 'see favorites button',
   });
   fireEvent.click(favoriteBtn);
   const datasetNameOne = screen.queryByText("pbench_user_benchmark1");
@@ -65,10 +65,10 @@ test("row is favorited after clicking on favorite icon", async () => {
 test("data is filtered based on value in search box", async () => {
   render(<AppWrapper />);
   await screen.findByText("pbench_user_benchmark1");
-  const searchBox = screen.getByPlaceholderText(/search dataset/i);
+  const searchBox = screen.getByPlaceholderText('Search Dataset');
   fireEvent.change(searchBox, { target: { value: "pbench_user_benchmark2" } });
   const searchBtn = screen.getByRole("button", {
-    name: /searchButton/i,
+    name: 'searchButton',
   });
   fireEvent.click(searchBtn);
   const datasetNameTwo = screen.queryByText("pbench_user_benchmark2");
--- a/dashboard/src/modules/components/AlertComponent/Alert.test.js
+++ b/dashboard/src/modules/components/AlertComponent/Alert.test.js
@@ -11,13 +11,13 @@ const AppWrapper = () => {
 };
 test("Alert message is displayed on initial load", () => {
   render(<AppWrapper />);
-  const alert = screen.getByText(/want to see your own data/i);
+  const alert = screen.getByText('Want to see your own data?');
   expect(alert).toBeInTheDocument();
 });
 
 test("Alert message is closed on clicking close button", () => {
   render(<AppWrapper />);
-  const alert = screen.getByText(/want to see your own data/i);
+  const alert = screen.getByText('Want to see your own data?');
   const closeButton = screen.getByRole("button", {
     name: /close info alert/i,
   });
--- a/dashboard/src/modules/components/TableComponent/common-components.jsx
+++ b/dashboard/src/modules/components/TableComponent/common-components.jsx
@@ -79,10 +79,10 @@ export const SearchBox = ({
     <InputGroup className="searchInputGroup">
       <TextInput
         type="text"
-        placeholder="Search"
+        placeholder="Search Dataset"
         onChange={(e) => setSearchKey(e)}
       ></TextInput>
-      <Button variant="control" onClick={search}>
+      <Button variant="control" onClick={search} aria-label="searchButton">
         <SearchIcon />
       </Button>
     </InputGroup>


shubham-html-css-js:
I don't know why it still says I want to push 30 commits even though all other commits have been merged(I am still learning github),but we have previous commits merged,so you just have to review the latest commit related to unit test

portante (changes requested):
30 commits for an initial PR might indicate that the starting point was not from the already merged code base in branch `main`.

Merge conflicts are being reported for this PR, so perhaps we could have it rebased?

dbutenhof (changes requested):
Great start, Shubham! Thanks for working on the unit tests; it's not always as much fun as new features, but it's important!

It sounds like you pushed this PR from a local branch where you'd already been working, without rebasing against `main`. The rebase should make them go away, so long as they really were merged.

I usually create a new branch from `upstream main` when starting a PR instead of re-using one; if you reuse, be sure you update your main and rebase first.

webbnh:
> I don't know why it still says I want to push 30 commits even though all other commits have been merged(I am still learning github),but we have previous commits merged,so you just have to review the latest commit related to unit test

Based on the GitHub ["Network" graph on the "Insights" page](https://github.com/distributed-system-analysis/pbench/network), it would appear that you created your `unit_test_browsing_page` branch from the end of your `browsing_page_dashboard_dev` branch, and that has about 30 commits on it since it branched from `main`.  So, your PR, which requests merging the `unit_test_browsing_page` branch into `main`, has 30 commits to merge.  Now, hopefully, the contents of all but the last of them would have already been present, so the merge would be a no-op, but that's not evident until the merge is attempted (and, even if it would be a no-op in the _end_, the individual merges might have been painful).  Now, you have since merged `main` into your `unit_test_browsing_page` branch, which _should_ make the subsequent merge into `main` fairly painless, but this isn't the usual route that we take.

When you start new work, you should start a new branch, and generally that branch should start from the current end of the `main` branch.  (So, before checking out and creating your new branch, you should issue a command like `git remote update` (and, I like to add the `--prune` switch to remove local tracking branches for upstream branches which have been deleted).  And, then, you should make sure that it's the `main` branch that you're checking out from.  (You _can_ reuse local development branches, but there's generally no need to, and it doesn't generally fit well with our preference for a "linear history".)

Second, you generally should _not_ merge **from** `main` to your development branch.  If there is something on `main` that you need, or if your development branch is otherwise out of date, you should _rebase_ your development branch, in effect "moving" the root of it to the new end of the main branch, as thought it had been checked out from there initially.  (This process has effects similar to merging from the main branch (including the possibility of having to manually resolve merge conflicts), but it produces a different, linear history, and it leaves your branch in a state where it can be rebased again, as needed.)

As for your present situation, there are several ways that you could recover.  One would be to reset your branch back prior to the merge from main, and then rebase it, changing the root from `browsing_page_dashboard_dev` branch to the end of `main` (this is a fairly advanced but fairly straightforward operation).  But, since you've only got one commit that you're interested in, the simplest thing would be to create a new branch rooted at the end of `main` and "cherry-pick" the one commit from your old branch on to it.

If you rename your old branch first, and then create the new branch with the old name, I believe that you'll be able to force-push it to your fork and it will replace the branch there...and the PR will be updated with the result.

dashboard/src/modules/components/SearchComponent/Search.test.js
```diff
@@ -0,0 +1,33 @@
...
+  const searchBox = screen.getByPlaceholderText(/search controllers/i);
+  fireEvent.change(searchBox, { target: { value: "dhcp2" } });
+  const searchBtn = screen.getByRole("button", {
+    name: /searchButton/i,
+  });
+  fireEvent.click(searchBtn);
+  const controllerTwo = screen.queryByText("dhcp2");
+  const controllerThree = screen.queryByText("dhcp3");
+  expect(controllerTwo).toBeInTheDocument();
+  expect(controllerThree).not.toBeInTheDocument();
```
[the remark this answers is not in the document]
webbnh: Also, "InTheDocument" seems like a pretty wide net -- shouldn't they be "in the table"?

dashboard/src/modules/components/DatePickerComponent/DatePicker.test.js
```diff
@@ -0,0 +1,30 @@
...
+  return {
+    get: () => ({
+      data: MOCK_DATA,
+    }),
+  };
+});
+test("data is filtered based on date range selected from date picker", async () => {
+  render(<AppWrapper />);
+  await screen.findByText("dhcp1");
+  const datePickerInput = screen.getAllByPlaceholderText(/yyyy-mm-dd/i);
```
webbnh, line 23: Can you target them by `aria-label` value?  That seems like a more robust discriminator.

dashboard/src/modules/components/DatePickerComponent/DatePicker.test.js
```diff
@@ -0,0 +1,30 @@
...
+  fireEvent.change(datePickerInput[0], { target: { value: "2022-02-16" } });
```
webbnh, line 24: Shouldn't we be checking that the UI has updated from `YYYY-MM-DD` to `2022-02-16`?

Also, can the code here look at the values in the React-store?  I think those would be good things to check, as well.

dashboard/src/modules/components/HeadingComponent/Heading.test.js
```diff
@@ -0,0 +1,16 @@
...
+    <Provider store={store}>
+      <App />
+    </Provider>
+  );
+};
+test("Page heading is displayed on initial load", () => {
+  render(<AppWrapper />);
+  const heading = screen.getByRole("heading", { name: /controllers/i });
+  expect(heading).toBeInTheDocument();
+});
```
webbnh, line 16: I think the `HeadingComponent` has been removed...in favor of the `NavbarCpomponent`?  So, this test should be renamed/removed/retargeted?

webbnh (changes requested):
Shubham, this is a great start!

However, I'm thinking that this PR should be split:  there are a number of pre-cursor changes which are not _directly_ related to unit testing, although they provide features which the unit testing depends on; they should be pulled out into a separate PR which this PR should be based on.

You do a lot of searching for UI elements by various bits of text.  These searches seem to be done via case-blind regular expressions containing literal strings.  It seems like we should know the exact text, so regular expressions (particularly the case-mapping) shouldn't be necessary.  And, it also seems like there should be a more rigorous way of addressing these elements than by picking bits of UI text...but, since that's what the user actually sees, perhaps that's just the thing, but...it seems odd to me.

And, I've got a few other nits, comments, suggestions, and questions.

dashboard/src/utils/mockData.js
```diff
@@ -0,0 +1,32 @@
+export const MOCK_DATA = [
```
dbutenhof, line 1: I'm still a bit concerned about the generic name `MOCK_DATA` ... this mocks the response of `/datasets/list`, but we'll certainly want to have other mocks in the future to test additional parts of the dashboard.

dbutenhof (commented):
Thanks, Shubham! I can resolve some of my previous conversations, but Webb has some good points/questions and I'd prefer to see a better name for the mocked data.

dbutenhof:
You need to `git fetch upstream main` and `git rebase` (or equivalent) to merge your `package.json` changes with the latest on `main`, then `git push --force origin unit_test_browsing_page` (that's a long branch name!)

webbnh:
> You need to `git fetch upstream main`

I've come to prefer `git remote update --prune`, but I guess `git fetch` works too, if you only want to update the single branch.  (One of the things that `--prune` does is remind me to delete my development branches after they've been merged into `main`.)

> that's a long branch name!

I don't know about you, but I've got Git's command line "completer" configured, so I just type a couple of characters and then tab-complete the branch name...so it doesn't matter how long it is.  😉

dashboard/src/modules/components/TableComponent/Table.test.js
```diff
@@ -19,30 +14,65 @@ jest.mock("utils/api", () => {
...
+  expect(datasetNameTwo).toBeInTheDocument();
+  expect(datasetNameThree).not.toBeInTheDocument();
+  expect(datasetNameFour).not.toBeInTheDocument();
+  expect(datasetNameFive).not.toBeInTheDocument();
+});
+test("data is filtered based on value in search box", async () => {
+  render(<AppWrapper />);
+  await screen.findByText("pbench_user_benchmark1");
+  const searchBox = screen.getByPlaceholderText('Search Dataset');
+  fireEvent.change(searchBox, { target: { value: "pbench_user_benchmark2" } });
```
webbnh, line 69: Just as a general thought (that is, I'm not necessarily requesting that you make a change), since the first element that the test wants to interact with is `searchBox`, it would make better sense for the `await` statement at line 67 to wait for _that_ element to appear.  This would also make the test more resilient to changes in the mock dataset.

dashboard/src/modules/components/TableComponent/index.jsx
```diff
@@ -51,7 +51,7 @@ const TableWithFavorite = () => {
     dispatch(getFavoritedDatasets());
   }, [dispatch]);
 
-  const { publicData, favoriteRepoNames } = useSelector(
+  const { publicData, favoriteRepoNames,tableData } = useSelector(
```
webbnh, line 54: Looks like you need to run Prettier again.  🙃

It would be good to develop a habit of running Prettier before each commit.

dashboard/package.json
```diff
@@ -37,6 +37,13 @@
     "test": "react-scripts test",
     "eject": "react-scripts eject"
   },
+  "jest":{
+    "transform": {
+      "^.+\\.[t|j]sx?$": "babel-jest"
```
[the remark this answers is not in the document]
webbnh: No; the first one is an escape for the second one; the resulting single backslash is then an escape for the subsequent dot, so that it is a literal period instead of a wildcard character.

dashboard/public/index.html
```diff
@@ -30,7 +30,7 @@
       Notice the use of %PUBLIC_URL% in the tags above.
       It will be replaced with the URL of the `public` folder during the build.
       Only files inside the `public` folder can be referenced from the HTML.
-
+
```
webbnh, line 33: You've got some superfluous blank characters, here.

webbnh (commented):
It looks like you got the extraneous commits off your branch (yay!), but apparently your branch has conflicts in `dashboard/package.json`, so it cannot be merged, and the CI tests are refusing to run; so you should address that as soon as possible.  (E.g., update your local tracking branch for `main`, rebase your branch on the end of `main` again, and resolve the conflicts during the rebase.)

Otherwise, it's just small stuff, but there are some previous comments of mine which you haven't addressed, yet.

dashboard/src/modules/components/SearchComponent/index.jsx
```diff
@@ -1,32 +0,0 @@
-import React,{useState} from "react";
```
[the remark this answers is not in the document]
webbnh: Nevermind:  this was an artifact of looking at partial diffs -- looking at the entire branch, this seems to be fine.

portante:
Please consider https://opensource.com/article/20/11/multiple-git-repositories.  This gives us all a common way to organize and work together using the same names.

> You need to `git fetch upstream main`

While it is nice to target one branch, it is more often helpful to keep track of everything in `upstream` (when defined as recommended in the blog above), since that avoid missing the creation of remote branches for shared projects or releases.

> and `git rebase` (or equivalent) to merge your `package.json` changes with the latest on `main`, then `git push --force origin unit_test_browsing_page` (that's a long branch name!)

You can also use `git push origin +unit_test_browsing_page` where the leading `+` means `--force`.

dbutenhof:
> While it is nice to target one branch, it is more often helpful to keep track of everything in upstream (when defined as recommended in the blog above), since that avoid missing the creation of remote branches for shared projects or releases.

Yup; but I don't use most of those branches routinely, and keeping `main` up to date is important. I tend to update remotes with `--prune` once in a while (maybe slightly more often than `git gc`), while I use `git fetch upstream main` regularly.

> You can also use git push origin +unit_test_browsing_page where the leading + means --force.

Now *that* I did not know, and that's useful...

portante (changes requested):
Please update commit log message with Jira issue ID "PBENCH-708".

dashboard/src/modules/components/TableComponent/Table.test.js
shubham-html-css-js: The rendering of `searchBox` on the  web page is more of a synchronous task with no practical waiting time involved.So using await for that does not make much sense.But the data fetched from the API is asynchronous,so we have to wait for that data to appear completely before we interact with it, hence use of await

dashboard/src/utils/mockData.js
dbutenhof: The `/datasets/list` response format has changed. There's now a "results" field in the JSON as well as a "next_url" for paging. Also, the dashboard now relies on the `resource_id`.

```
{
  "next_url": "",
  "results": [
    {
      "metadata": {
        "dataset.created": "2022-07-27T15:39:06+00:00"
      },
      "name": "pbench-user-benchmark_xyz.22_2022.07.27T15.39.06",
      "resource_id": "c54ab5929dc6230937402b682751c1ac"
    },
    {
      "metadata": {
        "dataset.created": "2022-07-27T15:01:57+00:00"
      },
      "name": "Short timer",
      "resource_id": "8c19617fc80580eaf6bb6808a1f8c46f"
    },
    {
      "metadata": {
        "dataset.created": "2022-07-27T13:45:02+00:00"
      },
      "name": "Soon FOUR",
      "resource_id": "4e28f76ccc9572f7e6a27898678db39d"
    },
    {
      "metadata": {
        "dataset.created": "2022-07-27T13:45:11+00:00"
      },
      "name": "Soonish",
      "resource_id": "892e506d3302ebaa466b9bc4c02f1780"
    },
    {
      "metadata": {
        "dataset.created": "2022-07-20T16:28:32+00:00"
      },
      "name": "Soon ONE",
      "resource_id": "6a4b55f64f8dd5f997be0da70a5e0d2c"
    },
    {
      "metadata": {
        "dataset.created": "2022-07-27T13:44:53+00:00"
      },
      "name": "Soon THREE",
      "resource_id": "e368132b16f8a45407c77c0a54a427fb"
    }
  ],
  "total": 6
}
```

Issue #2841, opened 2022-05-14
`pbench-uperf -h` fails with an error not being able to find the `uperf` command on `PATH`


portante:
When somebody invokes `pbench-uperf -h` they should not be given an error about not finding the `uperf` command on the `PATH`, but they should see the help text for `pbench-uperf`.

portante:
Fixed by #2860.

Issue #2842, opened 2022-05-14
When the `uperf` command fails for some reason `pbench-uperf` does not report a failure


portante:
When the `uperf` command fails for some reason `pbench-uperf` does not report a failure.

E.g., in the following cases if `root` does not have ssh keys setup on the client ...
```
pbench-uperf -r 10 -t stream -m 1 -p tcp -i 1 --samples 1 --clients 192.168.122.1
# reports no failures ($? == 0), the “results.json” file is generated and contains only metadata, no results.
# In “client_output.txt” I can see “** TCP: Cannot connect to 127.0.0.1:20010 Connection refused”

pbench-uperf -r 10 -t stream -m 1 -p tcp -i 1 --samples 1 --clients 192.168.122.1 --servers 0.0.0.0
# no error, $? == 0, but results.json contains only metadata.
# in sample output I can see “** TCP: Cannot connect to 0.0.0.0:20010 Connection refused”
```

PR #2712, opened 2022-03-25
Normalize variable references in `agent/base`

portante:
And fix a tab vs spaces problem and local variable declarations.

PR #2852, opened 2022-05-18
Correct `pbench-fio` and `pbench-uperf` handling of benchmark existence and version checks

portante:
Fixes issues #2841 and #2842.

The commits in this PR are separated into 3 preparatory commits followed by 2 separate commits with the actual fixes.

The "Correct `fio` and `uperf` existence checks" commit addresses issue #2841, while the "Correct local assignment then return code check" commit addresses issue #2842.  A similar breakage was posted previously in PR #2712.

The intention is to squash and merge all commits into one at the end; the separation is thought to help the review process.

One might ask, "why did you need to post commit 'Remove use of `benchmark_bin` environment variable' as part of this set?

There is an ordering problem with the old `benchmark_bin` variable initial value and where it is used in the rest of the script (both for `pbench-fio` and `pbench-uperf`).  The existence check is not always performed locally (no clients or servers specified which are local), but the commands run remotely required `benchmark_bin` to be set during creation of the commands for remote execution.  By only checking for the existence of the benchmark binary when performing the version check, and allowing the shell to resolve the location of the benchmark binary at run time, we avoid the interdependency altogether.


portante (commented):
Ready for review!

Remember -- the individual commits might make it easier to wade through all the changes.

portante:
No joke, `which` is a command provided by the `which` RPM, and was not part of CI container or set as an RPM dependency.

portante:
Still not quite right as the `which` output should not be captured in the gold files.

portante (commented):
Fixed the "which" problem.

dbutenhof:
> Fixed the "which" problem.

Wait, what? Who fixed *which* problem?

dbutenhof (approved):
You could use `whereis` instead, which is already on the container (and I think built into linux at a much more central level, though still technically an RPM):

```
EXTRA_PODMAN_SWITCHES=-it jenkins/run bash
bash-5.1$ which ls
bash: which: command not found
bash-5.1$ whereis -b ls
ls: /usr/bin/ls
bash-5.1$ dnf provides whereis
[...]
util-linux-2.36.2-1.fc34.i686 : A collection of basic system utilities
Repo        : fedora
Matched from:
Filename    : /usr/bin/whereis
```

agent/bench-scripts/tests/pbench-user-benchmark/test-67.txt
```diff
@@ -33,7 +33,7 @@ pbench-postprocess-tools: WARNING: No tools are registered
 [info][1900-01-01T00:00:00.000000] [pbench-user-benchmark] Running bm for iteration 1-default
 --- pbench.log file contents
 +++ test-execution.log file contents
-/var/tmp/pbench-test-bench/opt/pbench-agent/unittest-scripts/bm 
+/var/tmp/pbench-test-bench/opt/pbench-agent/unittest-scripts/bm
```
ndokos, line 36: Wasn't `bm` deleted? Why is it still here?

ndokos (approved):

agent/bench-scripts/tests/pbench-user-benchmark/test-67.txt
ndokos: NM - it was not deleted.

agent/bench-scripts/test-bin/which
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
webbnh, line 1: It would be worthwhile having a header comment explaining what the purpose of this script is.  (I presume it is intended to direct an invocation of `which` to the `/usr/bin/which` command instead of letting it be intercepted by a shell function or alias, and it also ignores any command line switches.)

agent/bench-scripts/unittests
```diff
@@ -169,6 +171,9 @@ function _dump_logs {
             for f in $(find . -name results.html 2> /dev/null | sort); do
                 _dump_file ${rundir} ${f:2}
             done
+            for f in $(find . \( -name '*--client_start.sh' -o -name '*--server_start.sh' \) 2> /dev/null | sort); do
```
webbnh, line 174: You don't need the `\(` and `\)`.

agent/rpm/pbench-agent.spec.j2
```diff
@@ -48,7 +48,7 @@ Requires:  perl-Time-HiRes
 
 Requires:  bc, bzip2, hostname, iproute, iputils, net-tools
 Requires:  openssh-clients, openssh-server, procps-ng, psmisc, redis
-Requires:  rpmdevtools, rsync, screen, sos, tar, xz
+Requires:  rpmdevtools, rsync, screen, sos, tar, which, xz
```
webbnh, line 51: Adding a dependency, even one as  common as `which`, is disappointing and seems like something that we should resist.

Why can't we use `command -v`?  (We already use it in 30 places in our code.)

agent/bench-scripts/tests/pbench-fio/test-55.env
```diff
@@ -1 +1 @@
-export _BM_EXIT_STS=42
+export _BM_FIO_STS=42
```
webbnh, line 1: I'm not against this change per se, but I think it was better having a generic "interface", rather than having differently-named environment variables which do the same things for each workload.

agent/bench-scripts/pbench-uperf
```diff
@@ -204,7 +197,13 @@ function local_pre_check {
...
 	local _match="${2}"
 
-	local ver_installed=$(${benchmark_bin} -V | awk 'NR==1 {print $3}')
+    resolve_benchmark_bin_err="$(resolve_benchmark_bin "${benchmark}" 2>&1)"
+    if [[ ${?} -ne 0 ]]; then
+        error_log "${script_name}: ${benchmark} executable not found, '${resolve_benchmark_bin_err}'"
+        exit 1
+    fi
+
+	local ver_installed=$(${benchmark} -V | awk 'NR==1 {print $3}')
```
webbnh, line 206: There seems to be a tabs vs. spaces indentation problem.

webbnh (approved):
Looks OK to me, but, naturally, I have a few comments and suggestions, some of which you might want to incorporate.

Also, there's a bug (not introduced in this PR, but it might be worth fixing here) in `pbench_fio`'s `fio_pre_check()`:  if, after a remote pre-check fails, we do a local pre-check which succeeds, we will overwrite the failure status with success and won't report the problem.  (Consider using the `|| rc=${?}` pipeline idiom to write the status only on failure.)

portante:
> You could use `whereis` instead, which is already on the container (and I think built into linux at a much more central level, though still technically an RPM):
> 
> ```
> EXTRA_PODMAN_SWITCHES=-it jenkins/run bash
> bash-5.1$ which ls
> bash: which: command not found
> bash-5.1$ whereis -b ls
> ls: /usr/bin/ls
> bash-5.1$ dnf provides whereis
> [...]
> util-linux-2.36.2-1.fc34.i686 : A collection of basic system utilities
> Repo        : fedora
> Matched from:
> Filename    : /usr/bin/whereis
> ```

I am not sure what to think looking at the history, the original request was to support aliases, see https://github.com/distributed-system-analysis/pbench/pull/1719#discussion_r459223742.  But what was implemented does not (it explicitly uses `--skip-aliases --skip-functions`).  So it is probably best to just use `command -v` like we do elsewhere.

Introducing `whereis` is still another dependency that needs to be added to the RPM even if it is already "magically" satisfied.

portante:
> > Fixed the "which" problem.
> 
> Wait, what? Who fixed _which_ problem?

I am not sure, _which_ one are you talking about?

agent/bench-scripts/tests/pbench-fio/test-55.env
portante: It seems safer to be sure the behavior request for `fio` does not impact `uperf` and vice versa.  A mistake elsewhere can lead to a leak across tests which this prevents.

portante (commented):
Ready for review without replacing `which` with `command -v`.  I'll do that in a follow-up commit in this PR so we can evaluate that separately.

portante (commented):
Updated with a separate commit replacing `which` with `command -v`.

ndokos (dismissed):

dbutenhof (dismissed):

agent/base
```diff
@@ -293,8 +293,10 @@ function generate_inventory {
 
 function resolve_benchmark_bin {
     # Encapsulation of method for resolving the actual benchmark
-    # binary.
-    which --skip-alias --skip-functions "${1}"
+    # binary ignoring `bash` aliases and functions.
+    unalias "${1}" > /dev/null 2>&1
+    unset -f "${1}" > /dev/null 2>&1
+    command -v "${1}"
```
webbnh, line 299: I'm slightly uncomfortable with the issuance of the `unalias` and `unset` commands, here, because they affect the environment of the calling shell.

If `${1}` is an alias, `command` will return `alias` as the command:
```
$ command -v ls
alias ls='ls --color=auto'
```
while if `${1}` is a shell function or built-in, `command` will return the unadorned value of `${1}`:
```
$ command -v which
which
$ command -v echo
echo
```
If `${1}` is anything else (i.e., an executable file), `command` returns a path (containing a slash):
```
$ command -v /bin/echo
/bin/echo
$ command -v exec-unittests 
./exec-unittests
$ command -v whereis
/usr/bin/whereis
```

So, it seems like, rather than modifying the environment (and, perhaps, surprising the user by _silently_ removing an intended redirect), we could instead detect and report the problem.  (I.e., if the value returned by `command -v` starts with `"alias ${1}"` or doesn't contain a slash, then reject it.)  Thoughts?

webbnh (dismissed):
Looks good; just one item to consider.

agent/base
webbnh: I think you misinterpreted what I meant by "calling shell".

In your example, the `./this` script is being executed in its own shell; therefore, the `unset -f` does not affect the environment in the shell which issued the `source demo` command.

However, in `agent/base`, the `unset -f` is issued inside a shell function, not a subshell, so it affects the environment of the shell which invoked the function...which, as far as I can tell, is very much the intend effect.  So, no, it doesn't affect the parent shell of the `pbench-fio` invocation, but it affects execution within that invocation.

My concern is that it is overriding the user's apparent intention, and it is doing so silently.  That is, if someone tries to set up an alias (or a shell function) for `fio`, it won't work, and there will be no indication of why (and no suggestion that what they needed to do was to change their `PATH`).  And, worse, if there is some other `fio` executable on their path, we'll try to run that instead of the one that the user wanted.

portante: But, that is how it worked before.  Aliases did not work before and don't work now. Functions were ignored, and the command was constructed using the resolved path to the actual binary.

This is now only checking for the actual binary, with the side effect of ignoring the function so that the resolved binary is still used.  We _could_ (see below) just document that we explicitly ignore functions.

I am beginning to think we are constructing too fragile a requirement on the "existence" of the executable, and should instead focus on the version check: if it exists, we'll go on to see if it returns a version we support, we're done.  If it doesn't exist, or the version does not match what we expect, we return an error.

webbnh: > But, that is how it worked before.

OK; on that basis, I'm going to go ahead and resolve this conversation.

> I am beginning to think we [...] should instead focus on the version check

I think that that would be great, but IIRC we went this other way because LINPACK didn't have a way to report its version....

portante: > I think that that would be great, but IIRC we went this other way because LINPACK didn't have a way to report its version....

`pbench-linpack` doesn't use this method as of today.

webbnh: > `pbench-linpack` doesn't use this method as of today.

I'm not sure which "this method" that is.

My understanding is that `pbench-linpack` doesn't do any version check today, and, in no small part, that's because there is nothing in the LINPACK benchmark workload which produces a version that it can check.  (So, we're trying to manufacture something, by looking at the path to the executable.)  Am I confused?

portante: I am a bit confused by what is being discussed.

Checking for the existence of an executable file should be removed (I'll post an update to do that late Monday or Tuesday).

Checking that when the `fio` or `uperf` commands are invoked with the appropriate option to report a version of the program, where that version is acceptable to us, should be the only thing we need to do.  If the invocation to generate a version fails, we should exit with an error.  If the invocation reports a version that we don't support, we should fail.  If it reports a version we do support, we should succeed.

What we do in `pbench-linpack` should not be discussed here since it is handled in an entirely different way and has not bearing on this discussion since the `resolve_benchmark_bin` function provided by `agent/base` is not used there.

portante (commented):
This seems like a much better solution where we don't have to resolve the binary location at all.

agent/bench-scripts/test-bin/fio
```diff
@@ -18,6 +18,7 @@ if [[ "${1%%=*}" == "--client" ]]; then
 	done
 elif [[ "${1}" == "--version" ]]; then 
 	echo "fio-${_BM_FIO_VER:-3.42}"
+	exit 0
 fi
 
 exit ${_BM_FIO_STS:-0}
```
webbnh, line 24: Are you sure you want to add the nested invocation of `exit`?  Without it, we'll still perform the invocation at line 24, and that one offers more interesting possibilities for testing.

agent/bench-scripts/tests/pbench-fio/test-22.txt
```diff
@@ -1,5 +1,5 @@
 +++ Running test-22 pbench-fio
-[error][1900-01-01T00:00:00.000000] [pbench-fio] fio executable not found
+[error][1900-01-01T00:00:00.000000] [pbench-fio] fio failed reporting its version: '/var/tmp/pbench-test-bench/opt/pbench-agent/bench-scripts/pbench-fio: line 476: fio: command not found'
```
webbnh, line 2: We like having the "script name" including the full directory path?  (I think I'm OK with it, but I don't think it's uniform practice.)

agent/bench-scripts/test-bin/fio
portante: The new behavior introduced by removing the executable file check and executing the `fio -V` is that we check the return status of that command.  We now need to ensure `fio -V` is successful, while the non-version check is not.

ndokos (approved):

PR #2855, opened 2022-05-19
Add dataset list pagination

npalaska:
Adds a server side pagination for dataset list api.

`/dataset/list` API will return paginated JSON object like following
```
{
   "next_url":"<url containing next dataset offset number>",
   "results":[
      {
         <requested dataset properties>
      },
      {
         <requested dataset properties>
      }
   ],
   "total":<total result count>
}
```

### DIFF
--- a/lib/pbench/server/api/resources/__init__.py
+++ b/lib/pbench/server/api/resources/__init__.py
@@ -351,10 +351,10 @@ def convert_string(value: str, _) -> str:
     return value
 
 
-def convert_int(value: int, _) -> int:
+def convert_int(value: Union[int, str], _) -> int:
     """
-    Verify that the parameter value is a int (e.g., not a JSON dict, or a
-    string), and return it.
+    Verify that the parameter value is either int or string and
+    if string then convert it to an int.
 
     Args:
         value: parameter value
@@ -366,7 +366,12 @@ def convert_int(value: int, _) -> int:
     Returns:
         the input value
     """
-    if type(value) is not int:
+    if type(value) in [str, int]:
+        try:
+            value = int(value)
+        except ValueError:
+            raise ConversionError(value, int.__name__)
+    else:
         raise ConversionError(value, int.__name__)
     return value
 
@@ -486,11 +491,11 @@ class ParamType(Enum):
 
     ACCESS = ("Access", convert_access)
     DATE = ("Date", convert_date)
+    INTEGER = ("Int", convert_int)
     JSON = ("Json", convert_json)
     KEYWORD = ("Keyword", convert_keyword)
     LIST = ("List", convert_list)
     STRING = ("String", convert_string)
-    INT = ("Int", convert_int)
     USER = ("User", convert_username)
 
     def __init__(self, name: str, convert: Callable[[Any, "Parameter"], Any]):
@@ -735,10 +740,7 @@ class ApiBase(Resource):
                 if schema[key].type == ParamType.LIST:
                     json[key] = values
                 elif len(values) == 1:
-                    if schema[key].type == ParamType.INT:
-                        json[key] = int(values[0])
-                    else:
-                        json[key] = values[0]
+                    json[key] = values[0]
                 else:
                     raise RepeatedQueryParam(key)
             else:
--- a/lib/pbench/server/api/resources/__init__.py
+++ b/lib/pbench/server/api/resources/__init__.py
@@ -351,6 +351,26 @@ def convert_string(value: str, _) -> str:
     return value
 
 
+def convert_int(value: int, _) -> int:
+    """
+    Verify that the parameter value is a int (e.g., not a JSON dict, or a
+    string), and return it.
+
+    Args:
+        value: parameter value
+        _: The Parameter definition (not used)
+
+    Raises:
+        ConversionError: input can't be validated or normalized
+
+    Returns:
+        the input value
+    """
+    if type(value) is not int:
+        raise ConversionError(value, int.__name__)
+    return value
+
+
 def convert_keyword(value: str, parameter: "Parameter") -> str:
     """
     Verify that the parameter value is a string and a member of the
@@ -470,6 +490,7 @@ class ParamType(Enum):
     KEYWORD = ("Keyword", convert_keyword)
     LIST = ("List", convert_list)
     STRING = ("String", convert_string)
+    INT = ("Int", convert_int)
     USER = ("User", convert_username)
 
     def __init__(self, name: str, convert: Callable[[Any, "Parameter"], Any]):
--- a/lib/pbench/server/api/resources/__init__.py
+++ b/lib/pbench/server/api/resources/__init__.py
@@ -366,14 +366,14 @@ def convert_int(value: Union[int, str], _) -> int:
     Returns:
         the input value
     """
-    if type(value) in [str, int]:
+    if type(value) is int:
+        return value
+    elif type(value) is str:
         try:
-            value = int(value)
+            return int(value)
         except ValueError:
-            raise ConversionError(value, int.__name__)
-    else:
-        raise ConversionError(value, int.__name__)
-    return value
+            pass
+    raise ConversionError(value, int.__name__)
 
 
 def convert_keyword(value: str, parameter: "Parameter") -> str:
--- a/lib/pbench/server/api/resources/__init__.py
+++ b/lib/pbench/server/api/resources/__init__.py
@@ -699,6 +699,7 @@ class ApiBase(Resource):
         and describe a distinct Schema for each HTTP method.
         """
         super().__init__()
+        self.config = config
         self.logger = logger
         self.schema = schema
         self.role = role
@@ -734,7 +735,10 @@ class ApiBase(Resource):
                 if schema[key].type == ParamType.LIST:
                     json[key] = values
                 elif len(values) == 1:
-                    json[key] = values[0]
+                    if schema[key].type == ParamType.INT:
+                        json[key] = int(values[0])
+                    else:
+                        json[key] = values[0]
                 else:
                     raise RepeatedQueryParam(key)
             else:
--- a/lib/pbench/server/api/resources/__init__.py
+++ b/lib/pbench/server/api/resources/__init__.py
@@ -491,7 +491,7 @@ class ParamType(Enum):
 
     ACCESS = ("Access", convert_access)
     DATE = ("Date", convert_date)
-    INTEGER = ("Int", convert_int)
+    INT = ("Int", convert_int)
     JSON = ("Json", convert_json)
     KEYWORD = ("Keyword", convert_keyword)
     LIST = ("List", convert_list)
--- a/lib/pbench/server/api/resources/datasets_list.py
+++ b/lib/pbench/server/api/resources/datasets_list.py
@@ -1,4 +1,5 @@
 import logging
+from typing import Dict, List, Tuple
 from urllib.parse import urlencode, urlparse
 
 from flask.json import jsonify
@@ -35,8 +36,8 @@ class DatasetsList(ApiBase):
                 Parameter("access", ParamType.ACCESS),
                 Parameter("start", ParamType.DATE),
                 Parameter("end", ParamType.DATE),
-                Parameter("page_no", ParamType.INT),
-                Parameter("page_limit", ParamType.INT),
+                Parameter("offset", ParamType.INTEGER),
+                Parameter("limit", ParamType.INTEGER),
                 Parameter(
                     "metadata",
                     ParamType.LIST,
@@ -48,35 +49,43 @@ class DatasetsList(ApiBase):
             role=API_OPERATION.READ,
         )
 
-    def get_paginated_obj(self, query: Query, json: JSON, url: str):
+    def get_paginated_obj(
+        self, query: Query, json: JSON, url: str
+    ) -> Tuple[List, Dict[str, str]]:
         """
-        Helper function to return a paginated object containing slice of dataset
-        (constructed according to the page limit and a page number), next page url,
-        and total items count.
-        TODO: There are multiple ways to optimize the functionality
-            1. Use of caching mechanism to temporarily store the queried datasets
-            instead of querying the database with every page request.
-            2. Use of unique pointers to store the last returned row and then use
-            this pointer in subsequent page request instead of initial start to
+        Helper function to return a slice of datasets (constructed according to the
+        user specified limit and an offset number) and a paginated object containing next page
+        url and total items count.
+        E.g. specifying the following limit and offset values will result in the corresponding
+        dataset slice:
+        "limit": 10, "offset": 20 -> dataset[20: 30]
+        "limit": 10 -> dataset[0: 10]
+        "offset": 20 -> dataset[20: total_items_count]
+        TODO: We may need to optimize the pagination
+            e.g Use of unique pointers to record the last returned row and then use
+            this pointer in subsequent page request instead of an initial start to
             narrow down the result.
         """
         paginated_result = {}
         total_count = query.count()
-        page_limit = json.get(
-            "page_limit", int(self.config.server_config.get("page_limit"))
-        )
-        page_no = json.get("page_no", 1)
-        items = (
-            query.order_by(Dataset.name)
-            .limit(page_limit)
-            .offset((page_no - 1) * page_limit)
-            .all()
-        )
-        json["page_no"] = page_no + 1
-        parsed_url = urlparse(url)
-        if len(items) < page_limit:
+        query = query.order_by(Dataset.name)
+
+        # Get the user specified limit, otherwise return all the items
+        limit = json.get("limit", total_count + 1)
+        query = query.limit(limit)
+
+        # Shift the query search by user specified offset value,
+        # otherwise return the batch of result starting from the
+        # first queried item.
+        offset = json.get("offset", 0)
+        items = query.offset(offset).all()
+
+        n = len(items)
+        if n < limit:
             next_url = ""
         else:
+            json["offset"] = offset + n
+            parsed_url = urlparse(url)
             next_url = (
                 parsed_url.scheme
                 + "://"
--- a/lib/pbench/server/api/resources/datasets_list.py
+++ b/lib/pbench/server/api/resources/datasets_list.py
@@ -1,7 +1,9 @@
 import logging
+from urllib.parse import urlencode
 
 from flask.json import jsonify
 from flask.wrappers import Request, Response
+from sqlalchemy.orm import Query
 
 from pbench.server import JSON, PbenchServerConfig
 from pbench.server.api.resources import (
@@ -23,6 +25,8 @@ class DatasetsList(ApiBase):
     API class to list datasets based on PostgreSQL metadata
     """
 
+    PAGE_LIMIT = 20
+
     def __init__(self, config: PbenchServerConfig, logger: logging.Logger):
         super().__init__(
             config,
@@ -33,6 +37,7 @@ class DatasetsList(ApiBase):
                 Parameter("access", ParamType.ACCESS),
                 Parameter("start", ParamType.DATE),
                 Parameter("end", ParamType.DATE),
+                Parameter("page_no", ParamType.INT),
                 Parameter(
                     "metadata",
                     ParamType.LIST,
@@ -43,6 +48,38 @@ class DatasetsList(ApiBase):
             ),
             role=API_OPERATION.READ,
         )
+
+    def get_paginated_obj(self, query: Query, json: JSON, url: str, page_limit: int):
+        """
+        Helper function to return a slice of dataset according to the page limit and
+        a page number and a json object containing next page url.
+        TODO: There are multiple ways to optimize the functionality
+            1. Use of caching mechanism to temporarily store the queried datasets
+            instead of querying the database with every page request.
+            2. Use of unique pointers to store the last returned row and then use
+            this pointer in subsequent page request instead of initial start to
+            narrow down the result.
+        """
+        paginated_result = {}
+        total_count = query.count()
+        if total_count > page_limit:
+            if "page_no" in json:
+                items = (
+                    query.order_by(Dataset.name)
+                    .limit(page_limit)
+                    .offset((json["page_no"] - 1) * page_limit)
+                    .all()
+                )
+                json["page_no"] += 1
+            else:
+                # This is the first page
+                items = query.limit(page_limit).all()
+                json["page_no"] = 1
+            paginated_result["next_url"] = url.rsplit("?", 1)[0] + "?" + urlencode(json)
+            paginated_result["total"] = total_count
+        else:
+            items = query.order_by(Dataset.name).all()
+        return items, paginated_result
 
     def _get(self, json_data: JSON, request: Request) -> Response:
         """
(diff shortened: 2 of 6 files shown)


lib/pbench/server/api/resources/datasets_list.py
```diff
@@ -23,6 +25,8 @@ class DatasetsList(ApiBase):
     API class to list datasets based on PostgreSQL metadata
     """
 
+    PAGE_LIMIT = 20
```
dbutenhof, line 28: 20 may be handy for debugging, but it's awfully low to go into `main`. I'm tempted to say this should be a config parameter, in `pbench-server.cfg` and/or in my imaginary future server configuration DB/API... but the default should be a good deal higher than 20.

dbutenhof (commented):
Cool start!

lib/pbench/server/api/resources/datasets_list.py
```diff
@@ -45,6 +50,38 @@ def __init__(self, config: PbenchServerConfig, logger: logging.Logger):
...
+                    query.order_by(Dataset.name)
+                    .limit(page_limit)
+                    .offset((json["page_no"] - 1) * page_limit)
+                    .all()
+                )
+                json["page_no"] += 1
+            else:
+                # This is the first page
+                items = query.limit(page_limit).all()
+                json["page_no"] = 1
```
webbnh, line 78: My instinct is that the only difference between the two values for `"page_no" in json` would be the value specified to `.offset()`.  That is, if the test is false, then the offset should be zero; but that doesn't mean that the query couldn't have provided an explicit zero-offset; analogously, is there any difference between not supplying the `.offset()` qualifier and supplying it with a default (presuably, zero) value?

So, this code should probably look more like this:
```
            offset_page = json.get("page_no", 1)
            items = (
                query.order_by(Dataset.name)
                .limit(page_limit)
                .offset((offset_page - 1) * page_limit)
                .all()
            )
            json["page_no"] = offset_page + 1
```

webbnh (commented):
I agree with Dave:  this looks like a good start.  I have a few thoughts below, the biggest one is to have the selection done by lines instead of pages, leaving the details to the client.

lib/pbench/server/api/resources/datasets_list.py
```diff
@@ -45,6 +50,38 @@ def __init__(self, config: PbenchServerConfig, logger: logging.Logger):
...
+            paginated_result["next_url"] = url.rsplit("?", 1)[0] + "?" + urlencode(json)
+            paginated_result["total"] = total_count
```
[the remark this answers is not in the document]
npalaska: I think that's a really good suggestion. I see how the `total` can be confusing as to whether it is total pages or total items. having a user-provided limit and offset value would give users control over how many datasets they want to see on their page.

lib/pbench/server/api/resources/datasets_list.py
```diff
@@ -45,6 +50,38 @@ def __init__(self, config: PbenchServerConfig, logger: logging.Logger):
...
+    def get_paginated_obj(self, query: Query, json: JSON, url: str, page_limit: int):
+        """
+        Helper function to return a slice of dataset according to the page limit and
+        a page number and a json object containing next page url.
+        TODO: There are multiple ways to optimize the functionality
+            1. Use of caching mechanism to temporarily store the queried datasets
+            instead of querying the database with every page request.
+            2. Use of unique pointers to store the last returned row and then use
+            this pointer in subsequent page request instead of initial start to
+            narrow down the result.
```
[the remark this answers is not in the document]
webbnh: > I thought Redis would be able to handle that.

I didn't think that Redis was involved, here...did you mean Postgres or SQLAlchemy?

> Implementing this approach means sorting the dataset on a unique column preferably a created timestamp

We need to determine exactly what our requirements are.  For starters, as Dave pointed out, eventually we will presumably want the client to be able to specify the ordering.  Then we need to determine whether we're allowed to return "fuzzy" results (i.e., is it OK if we return fewer items than requested?...is it OK if we return more?).  Finally, we need to decide what sort of help the database query can provide, and how expensive those various options are.

For instance, if we allow the client to specify the sorting order but restrict it to a single column, then we can come up with a pair of values which bound the items returned (i.e., every item returned has a value in the key column which is greater than or equal to `MIN` and less than `MAX`), and we return the `MAX` value with the results so that it becomes the `MIN` for the next query.  The problem with this approach is that for a given range of `MIN..MAX`, the number of items may not be exactly equal to the client's requested page size...if there is more than one value, then it's straightforward to return fewer rows than what was requested, but if all the selected rows have the same value then we could easily end up needing to return way more than what was requested.  If this is OK, then we're golden...but, if we want something more conformant or deterministic, then we need to work harder (which is where your desire for unique key values comes in...).  (We can do better if the client specifies multiple columns, but then we have to track the `MAX` for each of the columns, and there's still no guarantee of uniqueness.)

If the database query provides a stable sort, then we can use `offset()` and `limit()` instead of first doing a min/max selection.  If this is reasonably efficient (e.g., if sorting and offsetting is comparable in cost to doing a selection first and sorting the result of that), then this approach is easy; however, if this gets painful as the number of datasets gets large, then we'll want to look at another approach.

So, this is a rather long-winded way of saying that what I think you're proposing here is an excellent initial approach.  We'll presumably want to tweak it to allow the client to specify the basis for the sort, but, until the database operation proves to be too expensive, I don't think we need to worry about alternatives.

lib/pbench/server/api/resources/datasets_list.py
```diff
@@ -62,23 +62,31 @@ def get_paginated_obj(self, query: Query, json: JSON, url: str, page_limit: int)
...
+        items = (
+            query.order_by(Dataset.name)
+            .limit(page_limit)
+            .offset((page_no - 1) * page_limit)
+            .all()
+        )
+        json["page_no"] = page_no + 1
+        parsed_url = urlparse(url)
+        if len(items) < page_limit:
+            next_url = ""
```
webbnh, line 78: Are you sure that you want the server-side of the request to work in "pages"?  It has a couple of downsides:
- we (apparently) have to have a server configuration option to set the default page size;
- we have to convert between "page numbers" and item offsets; and,
- it produces unpleasant results if the page size varies between requests.

If we tracked things by items/lines, the code might look more like this:
```
        offset = json.get("item_offset", 0)
        item_count = json.get("item_count", 0)
        q = query.order_by(Dataset.name).offset(offset)
        if item_count:
            q = q.limit(item_count)
        items = q.all()
        n = len(items)
        json["item_offset"] = offset + n
        parsed_url = urlparse(url)
        if n < item_count:
            next_url = ""
```
- With this approach, we don't impose a limit on the number of items returned (unless the request does); thus, we don't need a server configuration option for that value.  (Ultimately, we will probably want to have some sort of a limit on the maximum size of _any_ response, but it will probably be in _bytes_ rather than **rows**.)
- Instead of passing back a page number, we pass back a row number.  This means that there is no conversion on the server side between items and pages, and the server doesn't need to know how big a page is, at all (and, the offset doesn't fall victim to changes in the page size).

lib/pbench/server/api/resources/datasets_list.py
```diff
@@ -62,23 +62,31 @@ def get_paginated_obj(self, query: Query, json: JSON, url: str, page_limit: int)
...
-            items = query.order_by(Dataset.name).all()
+            next_url = (
+                parsed_url.scheme
+                + "://"
+                + parsed_url.netloc
+                + parsed_url.path
+                + "?"
+                + urlencode(json)
+            )
+        paginated_result["next_url"] = next_url
```
webbnh, line 88: If there is no "next URL" (e.g., because we set it to an empty string), would you rather just omit it from the JSON object?  (That's easy to do....)

lib/pbench/server/api/resources/datasets_list.py
```diff
@@ -62,23 +62,31 @@ def get_paginated_obj(self, query: Query, json: JSON, url: str, page_limit: int)
...
+        paginated_result["total"] = total_count
```
webbnh, line 89: I still don't like "total"...how about "count" or "items" or "item_count"?

lib/pbench/test/unit/server/test_datasets_list.py
```diff
@@ -93,12 +103,25 @@ def get_results(self, name_list: List[str]) -> JSON:
...
-        return {"results": list}
+        query["page_no"] = page_no + 1
+        if len(paginated_name_list) < page_limit:
+            next_url = ""
+        else:
+            next_url = (
+                f"http://localhost{server_config.rest_uri}/datasets/list"
+                + "?"
+                + urlencode(query)
+            )
```
webbnh, line 114: Given that you're already using an f-string, you might as well include the `?` in line 111, rather than using an extra concatenation.

lib/pbench/server/api/resources/__init__.py
```diff
@@ -470,6 +490,7 @@ class ParamType(Enum):
     KEYWORD = ("Keyword", convert_keyword)
     LIST = ("List", convert_list)
     STRING = ("String", convert_string)
+    INT = ("Int", convert_int)
     USER = ("User", convert_username)
```
webbnh, line 494: Anybody up for maintaining the alphabetical order here?

webbnh (approved):
This looks good, but I still think we should leave the notion of pagination entirely to the client.  (This change would be much simpler if we had the server deal just in items/rows instead of pages.)

Beyond that, I've got the usual nits and suggestions/questions.

lib/pbench/server/api/resources/datasets_list.py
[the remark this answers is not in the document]
webbnh: Do we need to retain this `TODO` comment in this change?

lib/pbench/server/api/resources/datasets_list.py
```diff
@@ -44,6 +48,47 @@ def __init__(self, config: PbenchServerConfig, logger: logging.Logger):
             role=API_OPERATION.READ,
         )
 
+    def get_paginated_obj(self, query: Query, json: JSON, url: str):
+        """
+        Helper function to return a paginated object containing slice of dataset
+        (constructed according to the page limit and a page number), next page url,
+        and total items count.
```
webbnh, line 55: This function signature is missing its return type -- which should be something like `Tuple[List, Dict[str, str]]` -- which would make the text description somewhat more clear (since the text implies that it returns a single object containing three things, when in fact it returns two things, a list and an object containing two things).

webbnh (commented):
One more item I hit when trying to resolve my other comments.

lib/pbench/server/api/resources/datasets_list.py
```diff
@@ -62,23 +62,31 @@ def get_paginated_obj(self, query: Query, json: JSON, url: str, page_limit: int)
...
+        page_no = json.get("page_no", 1)
+        items = (
+            query.order_by(Dataset.name)
+            .limit(page_limit)
+            .offset((page_no - 1) * page_limit)
+            .all()
+        )
+        json["page_no"] = page_no + 1
+        parsed_url = urlparse(url)
+        if len(items) < page_limit:
```
webbnh, line 77: There's no need to call `urlparse()` unless there are more items to return, so you might as well move this call into the `else` branch.

webbnh (commented):
And, here's another one....

lib/pbench/server/api/resources/datasets_list.py
dbutenhof: "Page" is a pretty common term in this sort of model, so there's some risk in arbitrarily avoiding it. "Rows" sounds too DB-focused. We could do "items" or something like that, but it seems a bit artificial.

npalaska: Yeah, as Dave said these terminologies are very common in the community when the pagination is used. My reason behind using "pages", "next_url", and "total" is to sort of mimic the Elasticsearch response which many people are familiar with. Returning row numbers could expose some internals about our database which I am not sure is necessarily a bad thing though.

lib/pbench/server/api/resources/datasets_list.py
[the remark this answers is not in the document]
npalaska: Well, we would still need to update the algorithm (or optimize it) if we need to so that's there for a probable TODO. So I will just update the docstring a little bit.

lib/pbench/server/api/resources/datasets_list.py
dbutenhof: > Returning row numbers could expose some internals about our database which I am not sure is necessarily a bad thing though.

I think that returning our actual DB row values *would* be a bad thing, partly because we should be able to reconstruct our DB at need without affecting external APIs.

I think that in this context, however, "row" would refer to an ordinal displacement within the set of datasets selected by the query, *not* absolute PostgreSQL `id` column values. That is, the query covers 1,000 datasets, hypothetically enumerated 1 to 1000; we're now returning the 21st through the 40th, for example.

npalaska: Ah! That's right. So do we want to use lines/items or should we continue using the pages and count?

lib/pbench/server/api/resources/datasets_list.py
```diff
@@ -48,35 +49,43 @@ def __init__(self, config: PbenchServerConfig, logger: logging.Logger):
...
-            .all()
-        )
-        json["page_no"] = page_no + 1
-        parsed_url = urlparse(url)
-        if len(items) < page_limit:
+        query = query.order_by(Dataset.name)
+
+        # Get the user specified limit, otherwise return all the items
+        limit = json.get("limit", total_count + 1)
+        query = query.limit(limit)
```
webbnh, line 75: This is OK, but wouldn't it be better to just skip the `limit()` call if the user specifies no limit?

lib/pbench/server/api/resources/datasets_list.py
```diff
@@ -48,35 +49,43 @@ def __init__(self, config: PbenchServerConfig, logger: logging.Logger):
...
+
+        # Shift the query search by user specified offset value,
+        # otherwise return the batch of result starting from the
```
webbnh, line 78: s/result/results/

lib/pbench/test/unit/server/test_datasets_list.py
```diff
@@ -85,13 +85,9 @@ def get_results(self, name_list: List[str], query: JSON, server_config) -> JSON:
...
-        page_limit = query.get(
-            "page_limit", int(server_config.server_config.get("page_limit"))
-        )
-        page_no = query.get("page_no", 1)
-        paginated_name_list = name_list[
-            (page_no - 1) * page_limit : page_no * page_limit
-        ]
+        limit = query.get("limit", len(name_list) + 1)
+        offset = query.get("offset", 0)
+        paginated_name_list = name_list[offset : (offset + 1) * limit]
```
webbnh, line 90: I may be confused, but shouldn't that be
```
        paginated_name_list = name_list[offset : offset + limit]
```
That is, shouldn't the size of the slice be the same regardless of the offset?

lib/pbench/test/unit/server/test_datasets_list.py
```diff
@@ -112,7 +131,7 @@ def get_results(self, name_list: List[str]) -> JSON:
             ("drb", {"end": "1970-09-01"}, []),
         ],
     )
-    def test_dataset_list(self, query_as, login, query, results):
+    def test_dataset_list(self, query_as, login, query, results, server_config):
```
webbnh, line 134: You no longer need the `server_config` fixture here, right?  Ditto line 148.

lib/pbench/server/api/resources/datasets_list.py
```diff
@@ -44,6 +49,55 @@ def __init__(self, config: PbenchServerConfig, logger: logging.Logger):
...
+            json["offset"] = offset + n
+            parsed_url = urlparse(url)
+            next_url = (
+                parsed_url.scheme
+                + "://"
+                + parsed_url.netloc
+                + parsed_url.path
+                + "?"
+                + urlencode(json)
+            )
```
dbutenhof, line 96: Don't we want to retain the previous query parameters?

You could use `parsed_url._replace(query=urlencode(json))` to avoid picking apart all of the pieces. And you can use `urllib.parse.parse_qs` to turn the raw parsed query string into a dict which you could then `update` with your `json`. Maybe something like ...

```
    parsed_url = urlparse(url)
    next_url = parsed_url._replace(
        query=urlencode(parse_qs(parsed_url.query) | json)
    )
```

lib/pbench/test/unit/server/test_schema.py
```diff
@@ -128,15 +130,16 @@ def ok(username: str) -> User:
...
+            (ParamType.ACCESS, ["foobar"]),  # ACCESS is "public" or "private"
+            (ParamType.ACCESS, 0),  # ACCESS must be a string
             (ParamType.DATE, "2021-06-45"),  # few months have 45 days
             (ParamType.DATE, "notadate"),  # not valid date string
             (ParamType.DATE, 1),  # not a string representing a date
+            (ParamType.INT, "a"),  # can not convert to an int
+            (ParamType.JSON, (1, False)),  # tuple is not JSON
+            (ParamType.STRING, {"not": "string"}),  # dict is not string
             (ParamType.USER, False),  # not a user string
             (ParamType.USER, "xyzzy"),  # not a defined username
```
webbnh, line 142: There's no case for either `ParamType.KEYWORD` or `ParamType.LIST`...should there be?

lib/pbench/test/unit/server/test_schema.py
```diff
@@ -95,13 +95,15 @@ def test_enum(self):
...
-            (ParamType.USER, None, "drb", "1"),
             (ParamType.ACCESS, None, "PRIVATE", "private"),
+            (ParamType.DATE, None, "2021-06-29", date_parser.parse("2021-06-29")),
+            (ParamType.INT, None, "1", 1),
+            (ParamType.INT, None, 1, 1),
+            (ParamType.JSON, None, {"key": "value"}, {"key": "value"}),
             (ParamType.JSON, ["key"], {"key": "value"}, {"key": "value"}),
             (ParamType.KEYWORD, ["Llave"], "llave", "llave"),
+            (ParamType.STRING, None, "x", "x"),
+            (ParamType.USER, None, "drb", "1"),
```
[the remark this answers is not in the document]
dbutenhof: `LIST` and `KEYWORD` have separate constraints (e.g., the type of a list, and the set of keywords), and they have their own tests.

On the other hand ... it'd probably be good to have a test case for good/bad `INT` elements of a `LIST` in `test_list_normalization` and `test_invalid_list`...

lib/pbench/server/api/resources/datasets_list.py
webbnh: That's annoying:  the `_replace()` method looks like a really useful widget, but the leading underscore indicates that it is internal to its defining class and not for general use; consistent with this, there is no documentation for the method in the obvious place; and, yet, the documentation for [`urlparse()`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) mentions it and shows examples of what it does and how to use it.  😝

My apologies @dbutenhof, but could you explain what the `| json` does, there?

lib/pbench/test/unit/server/test_datasets_list.py
webbnh: Actually, I think I like [Dave's approach](https://github.com/distributed-system-analysis/pbench/pull/2855#discussion_r882030796) better, although it is sort of hard to immediately comprehend.

lib/pbench/test/unit/server/test_schema.py
webbnh: [No.](https://github.com/distributed-system-analysis/pbench/pull/2855#discussion_r882046133)

lib/pbench/server/api/resources/datasets_list.py
dbutenhof: Note that `|` also *replaces* keys in the left-hand dict with new values from the right-hand dict: which means merging a URL that already has a `&offset=old` will *update* it to `&offset=new`.

npalaska: hmm `_replace` is an interesting choice, I didn't come across it when I read the documentation. Thanks, @dbutenhof and @webbnh for pointing out the use of the method but note that this method is not documented for python version 3.8 and earlier so I wonder whether it is added on python version 3.9 onwards, and will this be a problem for us? 

> Don't we want to retain the previous query parameters?

@dbutenhof, the `json` that gets passed to this method is the original query parameters and we just update the offset value of the json and everything else remains the same. So we do retain all the original query parameters.

lib/pbench/server/api/resources/datasets_list.py
npalaska: In effect, they are the same thing but yeah we could avoid using the limit, this approach just avoids the conditional statement that comes when avoiding the limit.

lib/pbench/test/unit/server/test_datasets_list.py
npalaska: Ah! you are right it should be `+` not `*`

lib/pbench/test/unit/server/test_datasets_list.py
npalaska: Yes I do, to construct the next url

lib/pbench/server/api/resources/datasets_list.py
dbutenhof: The documentation is minimal; it's mentioned, with an example, on https://docs.python.org/3/library/urllib.parse.html#url-parsing, and it works fine on my Python 3.6 installation.

(`dict1 | dict2`, however, doesn't work, and appears to be new in 3.9; unfortunately, `dict.update()` doesn't return a value, so the code is a bit messier than the sample I'd written.)

lib/pbench/server/api/resources/__init__.py
```diff
@@ -470,6 +490,7 @@ class ParamType(Enum):
     KEYWORD = ("Keyword", convert_keyword)
     LIST = ("List", convert_list)
     STRING = ("String", convert_string)
+    INT = ("Int", convert_int)
```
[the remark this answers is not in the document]
dbutenhof: Fine; I'd prefer `INTEGER = ("Integer", convert_integer)` since none of the others are abbreviated. (Though JSON is an acronym...) But as you've resisted this for several iterations I guess I'll give up.

lib/pbench/test/unit/server/test_schema.py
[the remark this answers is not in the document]
dbutenhof: I do think you should have test cases (good and bad) in `LIST` context...

dbutenhof (commented):
A few conversations left to resolve with Webb, and I'd like to see a few extra test cases.

lib/pbench/server/api/resources/datasets_list.py
webbnh: Yeah, it depends on which is more expensive, the call to `query.limit()` or the conditional statement (where the "cost" is not only in terms of execution time but also code complexity, testing, etc.).

Also, if we don't make the call to `limit()` then we don't have to worry about whether the default should be `total_count` or `total_count + 1`.  🙃  (I'm _pretty_ sure that the `+1` is superfluous and a waste....)

webbnh (commented):
Almost there!  I have a couple of code suggestions for you, and there are a couple of previous comments which are still open.

lib/pbench/test/unit/server/test_schema.py
```diff
@@ -354,23 +354,37 @@ def test_invalid_keyword(self, test):
...
+                [1, 2, 3],
+                ",",
+            ),
+            (
+                ParamType.INT,
+                None,
+                ["1", "2", "3"],
+                [1, 2, 3],
+                None,
+            ),
```
dbutenhof, line 378: Not going to try one where the incoming is already `int`??

dbutenhof (approved):

webbnh (approved):
What Dave said.  😀

lib/pbench/test/unit/server/test_schema.py
dbutenhof: Given that Webb and I both approved despite this loose end ... I guess I'll resolve and move on.

Issue #2843, opened 2022-05-14
`pbench-linpack` hangs indefinitely if `linpack` not installed on remote clients


portante:
`pbench-linpack` hangs indefinitely if `linpack` not installed on remote clients.

In the example below, the expected linpack installation is not present on node `192.168.122.1`:

```
pbench-linpack --samples=1 --threads=1 -c 192.168.122.1
for NAME in pbench-tool-meister-start redis-server gpg-agent scdaemon fio uperf linpack; do killall -9 $NAME; done
```

Hangs for indefinitely: “Waiting for all clients to finish pbench-linpack jobs.”

Perhaps we need to detect when the main process finishes to prevent such hangs?

Note that there were no failures in the output, nor anything useful in the files on `localhost` nor on the `192.168.122.1`.

PR #2856, opened 2022-05-19
Add local and remote pre-check for linpack

portante:
Fixes issues #2843.


portante:
Proposal to fix #2843.

agent/bench-scripts/driver/linpack
```diff
@@ -225,16 +231,22 @@ while true; do
...
-binary=${1}
-if [[ -z "${binary}" ]]; then
-	printf -- "[%s] ERROR: You must specify the location of the LINPACK binary\n\n" "${script_name}" >&2
+ver=${1}
+if [[ -z "${ver}" ]]; then
+	printf -- "[%s] ERROR: You must specify the version of the LINPACK binary\n\n" "${script_name}" >&2
 	usage >&2
 	exit 2
 fi
+binary="/usr/local/pbench-linpack-${ver}/benchmarks/linpack/xlinpack_xeon64"
```
webbnh, line 240: This is highly opinionated!  How does this file get created, or, how does the user know to put this file here?

webbnh, line 240: Just a thought....  How about changing line 14 to
```
binary_template="/usr/local/pbench-linpack-%s/benchmarks/linpack/xlinpack_xeon64"
```
and changing this line to
```
binary=$(printf -- ${binary_template} ${ver})
```
(since I know that you like `printf` 😉).

Alternately, we could set the template like
```
binary_template="/usr/local/pbench-linpack-\${ver}/benchmarks/linpack/xlinpack_xeon64"
```
and change this line to
```
eval binary=${binary_template}
```
but that feels like it might be a little too arcane (and slightly fragile), to me.

agent/bench-scripts/pbench-linpack
```diff
@@ -113,24 +117,42 @@ while true; do
...
-fi
-if [[ -z "${linpack_dir}" ]]; then
-	linpack_dir="/usr/local/${script_name}-${ver}/benchmarks/linpack"
-	linpack_dir_kind="default"
-else
-	linpack_dir_kind="provided"
+function pre_check {
+	${pbench_bin}/bench-scripts/driver/linpack --pre-check ${ver}
+	return ${?}
+}
```
webbnh, line 123: Where is `${ver}` defined?  Better yet, should it be a positional argument (i.e., `${1}`) to this function?

Should we be checking the argument (like, making sure there is one...) before passing it on to the driver?

(FWIW, I don't think that the `return ${?}` actually accomplishes anything -- I think that we get the same result if you omit the argument or the entire statement.)

agent/bench-scripts/pbench-linpack
```diff
@@ -113,24 +117,42 @@ while true; do
...
+
+if [[ ${do_pre_check} -ne 0 ]]; then
+	# We have been invoked remotely to check for the expected version of
+	# linpack installed.
+	pre_check ${@}
+	exit ${?}
```
webbnh, line 129: You can omit the `${?}` from the `exit` command -- that's the default.

agent/bench-scripts/pbench-linpack
```diff
@@ -113,24 +117,42 @@ while true; do
...
-	linpack_dir_kind="provided"
+function pre_check {
+	${pbench_bin}/bench-scripts/driver/linpack --pre-check ${ver}
+	return ${?}
+}
+
+if [[ ${do_pre_check} -ne 0 ]]; then
+	# We have been invoked remotely to check for the expected version of
+	# linpack installed.
+	pre_check ${@}
```
webbnh, line 128: What is in `${@}` at this point?  (I think it's just the non-switch command line arguments....)  There's nothing in the help text which mentions command line arguments.

webbnh (approved):
This looks good.  I think there's probably a small problem with the `pre_check()` function argument, but that's easily dealt with.  My only real question centers around how the user is supposed to know where to put the linpack binary and how to name the directory path.

portante (commented):
@webbnh, I have not had a chance to review your comments yet.  Sorry.  I have been working on a better fix for this problem and will work on addressing your comments next week.  Thanks for the review!

webbnh:
No worries...I hope you don't see them before Monday.  😉

agent/bench-scripts/tests/pbench-linpack/test-65.pre
```diff
@@ -1,7 +1,11 @@
 #!/bin/bash
 
 # Setup a fake install of linpack but with a stub for the expected executable.
-_linpack_dir=${_testtmp}/linpack
+_linpack_dir=${_testtmp}/pbench-linpack-11.1.3
+mkdir ${_linpack_dir} || exit 1
+_linpack_dir+="/benchmarks"
+mkdir ${_linpack_dir} || exit 1
+_linpack_dir+="/linpack"
 mkdir ${_linpack_dir} || exit 1
```
webbnh, line 9: Making `_linpack_dir` "cumulative" is clever, but I think the following is shorter and clearer code:
```
_linpack_dir=${_testtmp}/pbench-linpack-11.1.3
mkdir ${_linpack_dir} || exit 1
mkdir ${_linpack_dir}/benchmarks || exit 1
mkdir ${_linpack_dir}/benchmarks/linpack || exit 1
```
and, actually, this might actually be a good and appropriate place for `mkdir -p`...
```
_linpack_dir=${_testtmp}/pbench-linpack-11.1.3
mkdir -p ${_linpack_dir}/benchmarks/linpack || exit 1
```
Given that we've already created `${_testtmp}`, we really only need one command to create all of the Linpack directories.

webbnh (commented):
Looks like you addressed a couple of my comments without seeing them.  😀

Here's one more.

agent/bench-scripts/driver/linpack
webbnh: Your subsequent change improved this.

webbnh: With the addition of the install-directory parameter, this is largely obviated.

agent/bench-scripts/pbench-linpack
webbnh: Other than the nit about the `return` statement, you've largely addressed these.

agent/bench-scripts/pbench-linpack
portante: It won't ever be described in the help text, this is an internal hand shake of invoking the `pbench-linpack` remotely with `--pre-check` to cause the same behavior to be taken on those hosts.

webbnh (dismissed):
Looks good.

agent/bench-scripts/pbench-linpack
```diff
@@ -113,24 +117,49 @@ while true; do
...
-else
-	linpack_dir_kind="provided"
+function pre_check {
+	# Invoke the linpack driver to perform a pre-check that it will be able to
+	# execute the benchmark. The expected version is the first argument, and the
+	# second is the expected directory prefix for the linpack installation.
+	local ver=${1}
+	local install_prefix_arg=${2}
+
+	${pbench_bin}/bench-scripts/driver/linpack --pre-check ${ver} ${install_prefix_arg}
```
dbutenhof, line 127: The driver option is `--pre-check-only`; `getopt` allows abbreviation, but is that really what we want here?

dbutenhof (dismissed):
Misleading comment and possibly confusing long option abbreviation... but I'll leave these as optional suggestions since it ought to work as-is.

agent/bench-scripts/driver/linpack
```diff
@@ -38,15 +39,18 @@ kmp_affinity_args_def="nowarnings,compact,1,0,granularity=fine"
 kmp_affinity_args=""
 numactl_args=""
 
-# Output directory required.
+# Output directory, defaults to the current working directory if not provided.
 output_dir=""
 
+# By default we do not perform pre-checks unless requested explicitly.
+pre_check_only=0
```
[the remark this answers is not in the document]
portante: > The comment appears to be incorrect. In fact, we _always_ do pre-checks, but by default continue upon successful completion, unless we're asked to exit by setting "pre-check _only_".

Nice catch, thanks!

> (This appears to describe the `pbench-linpack` wrapper behavior, which wasn't described.)

Does this (`pbench-linpack` [line 132](https://github.com/distributed-system-analysis/pbench/pull/2856/files/bcf9e7a8df368b5677152d35a58085869a6b8c20#diff-c4d94e914ef64c2311d6e5f0922bc41b92ad5076e3c0adf5d377366222d628d6R132)) help?

agent/bench-scripts/pbench-linpack
portante: Nice catch, this is a bug.

portante (commented):
Ready for final review after correcting the "pre-check" handling.

webbnh (approved):
👍

dbutenhof (approved):
More nits, but relatively minor.

dbutenhof (approved):

ndokos (approved):

Issue #2582, opened 2021-12-10
Streamline the `if`/`else` cases in tests a bit.


npalaska:
This `if`/`else` is repeated several times, and that stands out in the pattern of your changes. It could really be refactored into a new superclass method, which would streamline the test cases a bit.

_Originally posted by @dbutenhof in https://github.com/distributed-system-analysis/pbench/pull/2581#discussion_r766644049_

PR #2860, opened 2022-05-24
Stop resolving benchmark binary location

portante:
Both `pbench-fio` and `pbench-uperf`:

 * No longer allow for the benchmark binary to be overriden by an environment variable -- this was an out-dated way for the unit tests to mock the respective benchmark behavior

 * No longer resolve and check that the benchmark binary is executable

There was an ordering problem with the old `benchmark_bin` variable initial value and where it is used in the rest of the script (both for `pbench-fio` and `pbench-uperf`). The existence check was not always performed locally (no clients or servers specified which are local), but the commands run remotely required `benchmark_bin` to be set during creation of the commands for remote execution. By only checking for the existence of the benchmark binary when performing the version check, and allowing the shell to resolve the location of the benchmark binary at run time, we avoid the interdependency altogether.

Mock commands for `fio` and `uperf` are provided on the `PATH` for tests.  For the CLI tests, those mocks are removed so that we can verify that help and usage text is emitted before the checks for the particular command existence (which is shown by the failing `test-CL`).

We also add failing tests for `uperf` and `fio` behaviors:

 * `pbench-fio`
   * `test-22` -- missing `fio` command
   * `test-50` -- `fio -V` reports a bad version
 * `pbench-uperf`
   * `test-02` -- missing `uperf` command
   * `test-51` -- `uperf -V` reports a bad version

The existence check of the `fio` and `uperf` benchmark commands now occurs after any help requests or command usage errors. This fixes issue #2841 [1].

Finally, we correct the way `pbench-uperf` checked the status of the `uperf` command execution of the benchmark by making sure the `local` declaration is performed before the assigment, so that the return code check is not overridden by the "status" of the `local` declaration.  This fixes issue #2842 [2].

[1] https://github.com/distributed-system-analysis/pbench/issues/2841
[2] https://github.com/distributed-system-analysis/pbench/issues/2842

----

This is a respin of PR #2582 to correct the mistaken merge without squashing.


webbnh (approved):
GTG

dbutenhof (approved):

PR #2862, opened 2022-05-24
Handle response payloads which are not JSON in generate_token.py

### LINKED ISSUE #2839: Improve error message for `pbench-generate-token` when server is unreachable
The `pbench-generate-token` command has a bad failure mode if server is unreachable:

    Expecting value: line 1 column 1 (char 0)

webbnh:
While the server responds to requests with a JSON payload, even for errors, not all responses come from the server.  (E.g., in the event that the server is unresponsive (e.g., because it's not running), the proxy will respond with a `408/Request Timeout`.)  This PR adds a little bit of defense to `generate_token.py` where we decode the response payload.

@dbutenhof, can you give this a try?

Fixes #2839.

### DIFF
--- a/lib/pbench/cli/agent/commands/generate_token.py
+++ b/lib/pbench/cli/agent/commands/generate_token.py
@@ -1,5 +1,7 @@
 """pbench-generate-token"""
 
+from json import JSONDecodeError
+
 import click
 import requests
 
@@ -30,7 +32,11 @@ class GenerateToken(BaseCommand):
         except requests.exceptions.ConnectionError as exc:
             raise RuntimeError(f"Cannot connect to '{uri}'") from exc
 
-        payload = response.json()
+        try:
+            payload = response.json()
+        except JSONDecodeError:
+            payload = {"message": response.text}
+
         if response.ok:
             click.echo(payload["auth_token"])
             return 0
--- a/lib/pbench/cli/agent/commands/generate_token.py
+++ b/lib/pbench/cli/agent/commands/generate_token.py
@@ -37,7 +37,7 @@ class GenerateToken(BaseCommand):
         except JSONDecodeError:
             payload = {"message": response.text}
 
-        if response.ok:
+        if response.ok and "auth_token" in payload:
             click.echo(payload["auth_token"])
             return 0
 


lib/pbench/cli/agent/commands/generate_token.py
```diff
@@ -30,7 +32,11 @@ def execute(self):
...
             raise RuntimeError(f"Cannot connect to '{uri}'") from exc
 
-        payload = response.json()
+        try:
+            payload = response.json()
+        except JSONDecodeError:
+            payload = {"message": response.text}
+
         if response.ok:
             click.echo(payload["auth_token"])
```
dbutenhof, line 41: This is interesting, because if we get the `JSONDecodeError` when `response.ok`, we'll get a key error accessing `"auth_token"`, which probably isn't ideal.

However, for my case, with the server stopped, this is a distinct improvement, although the rendering of `response.text` (which is apparently an Apache error html page) is pretty ugly:

```
[root@dhcp31-136 ~]# pbench-generate-token --username roger --password roger
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>503 Service Unavailable</title>
</head><body>
<h1>Service Unavailable</h1>
<p>The server is temporarily unable to service your
request due to maintenance downtime or capacity
problems. Please try again later.</p>
</body></html>
```

I wonder if a `try` around the `payload["auth_token"]` to cleanly handle an error there might not be a bad idea while you're at it?

dbutenhof (dismissed):
Definitely solves the immediate problem, though maybe could be a bit more robust against other potential issues???

lib/pbench/cli/agent/commands/generate_token.py
webbnh: Instead of `response.text` I could use `response.reason` (which, in the case, above, I presume would be `Service Unavailable`), do you think that would be better?

(I'll see what I can do for the `KeyError` problem.)

webbnh: OK, in the unlikely event that the response is "OK" but doesn't return an `"auth_token"`, we will now exit with an error after reporting that things are "OK".  😝

Let me know if this is not Good-er.

portante (approved):

dbutenhof (approved):
Yeah, that extra level of protection is probably good; however I kinda wonder, if that happens, what `response.reason` is going to be at line 45. OK?? 😆 

This seems to cover the most likely cases, anyway, and maybe is "good enough".

BTW, yes, `response.reason` is "Service unavailable". While the HTML is longer, it's not really any more useful and it's a lot uglier; so maybe the `response.reason` *would* be better. (The proxy service seems to be guessing *why* the service is unavailable, but it really doesn't know so maybe simplerer is betterer.)

webbnh:
> I kinda wonder, if that happens, what `response.reason` is going to be at line 45. OK?? 😆

If we're very lucky, it will be `'Request fulfilled, document follows'`; otherwise, yeah, it'll be just `'OK'`; either way, the user is going to be left wondering what happened...but, I guess it's better than the `JSONDecodeError` text.

(I'm betting that this will never happen:  who in the pipeline other than the server is going to return `200`?...and hopefully the server will always supply the token when it returns that status.)

npalaska (approved):

Issue #2861, opened 2022-05-24
`pbench-tool-meister-stop` uses the wrong IP address to communicate with the Redis server


portante:
A simple 2-node `uperf` run fails as follows:

```
# pbench-uperf \
    --config=portante-test-3 \
    --server=10.10.10.156 \
    --test-types=stream \
    --protocols=tcp \
    --message-sizes=8192 \
    --instances=8 \
    --runtime=10 \
    --samples=1 \
    --server-node=1 \
    --client-node=1
Starting iteration  (1 of 1)
test sample 1 of 1
client[127.0.0.1]node[1]protocol[tcp]test[stream]instances[8]size[8192] <-> server[10.10.10.156]node[1]
Iteration 1-tcp_stream-8192B-8i complete (1 of 1), with 1 pass and 0 failures
Traceback (most recent call last):
  File "/opt/pbench-agent/lib/python3.9/site-packages/redis/connection.py", line 614, in connect
    sock = self.retry.call_with_retry(
  File "/opt/pbench-agent/lib/python3.9/site-packages/redis/retry.py", line 46, in call_with_retry
    return do()
  File "/opt/pbench-agent/lib/python3.9/site-packages/redis/connection.py", line 615, in <lambda>
    lambda: self._connect(), lambda error: self.disconnect(error)
  File "/opt/pbench-agent/lib/python3.9/site-packages/redis/connection.py", line 680, in _connect
    raise err
  File "/opt/pbench-agent/lib/python3.9/site-packages/redis/connection.py", line 668, in _connect
    sock.connect(socket_address)
ConnectionRefusedError: [Errno 111] Connection refused

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/pbench-agent/util-scripts/pbench-tool-meister-stop", line 14, in <module>
    sys.exit(main())
  File "/opt/pbench-agent/lib/pbench/agent/tool_meister_stop.py", line 301, in main
    return start(sys.argv[0], parsed)
  File "/opt/pbench-agent/lib/pbench/agent/tool_meister_stop.py", line 204, in start
    with Client(
  File "/opt/pbench-agent/lib/pbench/agent/tool_meister_client.py", line 125, in __enter__
    self.to_client_chan = RedisChannelSubscriber(
  File "/opt/pbench-agent/lib/pbench/agent/redis_utils.py", line 50, in __init__
    self._pubsub.subscribe(channel_name)
[...]

PR #2863, opened 2022-05-24
Ensure use of `localhost` when stopping Redis

portante:
When `pbench-tool-meister-stop` is invoked where the Redis server was created locally by `pbench-tool-meister-start`, we always want to use the `localhost` IP address to talk to it.

Fixes issue #2861.

### DIFF
--- a/lib/pbench/agent/tool_meister_stop.py
+++ b/lib/pbench/agent/tool_meister_stop.py
@@ -65,6 +65,10 @@ class RedisServer(RedisServerCommon):
             )
         except FileNotFoundError:
             pass
+        else:
+            # Since this Redis server is locally managed, communication to it
+            # will always be through the "localhost" interface.
+            self.host = "localhost"
 
     def locally_managed(self) -> bool:
         return self.pid_file is not None
--- a/lib/pbench/agent/tool_meister_stop.py
+++ b/lib/pbench/agent/tool_meister_stop.py
@@ -66,8 +66,9 @@ class RedisServer(RedisServerCommon):
         except FileNotFoundError:
             pass
         else:
-            # Since this Redis server is locally managed, communication to it
-            # will always be through the "localhost" interface.
+            # The redis.pid file exists in the "tm" directory, which means this
+            # Redis server is locally managed.  Communication with it will
+            # always be through the "localhost" interface.
             self.host = "localhost"
 
     def locally_managed(self) -> bool:
@@ -290,7 +291,7 @@ def main():
     parser.add_argument(
         "--redis-server",
         dest="redis_server",
-        default=os.environ.get("PBENCH_REDIS_SERVER", None),
+        default=os.environ.get("PBENCH_REDIS_SERVER", ""),
         help=(
             "Use an existing Redis server specified by <hostname>:<port>;"
             " implies the use of an existing Tool Data Sink and Tool Meisters"
--- a/lib/pbench/test/unit/agent/test_tool_meister_stop.py
+++ b/lib/pbench/test/unit/agent/test_tool_meister_stop.py
@@ -20,7 +20,7 @@ class TestRedisServer:
         rs = RedisServer("", rundir, "notme.example.com")
         assert (
             rs.locally_managed()
-        ), "Populated run directory did not result in a 'locally managed' RedisServer instance"
+        ), "RedisServer incorrectly inferred a non-locally managed instance from a run directory with a 'tm/redis.pid' file"
         assert (
             rs.host == "localhost"
         ), f"Expected 'RedisServer.host' to be 'localhost', got '{rs.host}'"
@@ -34,7 +34,7 @@ class TestRedisServer:
         rs = RedisServer(f"{rs_host}:4343", rundir, "notme.example.com")
         assert (
             not rs.locally_managed()
-        ), "Empty run directory still resulting in a 'locally managed' RedisServer instance"
+        ), "RedisServer incorrectly inferred a locally managed instance from an empty run directory"
         assert (
             rs.host == "redis.example.com"
         ), f"Expected 'RedisServer.host' to be '{rs_host}', got '{rs.host}'"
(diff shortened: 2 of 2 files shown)


portante (commented):
Still need to add a unit test.

lib/pbench/agent/tool_meister_stop.py
```diff
@@ -65,6 +65,10 @@ def __init__(self, spec: str, benchmark_run_dir: Path, def_host_name: str):
             )
         except FileNotFoundError:
             pass
+        else:
+            # Since this Redis server is locally managed, communication to it
+            # will always be through the "localhost" interface.
```
dbutenhof, line 70: We're assuming that since we have a `redis.pid` in the run directory, we created and manage the local Redis. Which I suppose is a fairly solid assumption, but saying that directly might be better. (That is, justifying the assumption rather than simply stating it.) This is, in any case, consistent with `locally_managed`, which triggers off the lack of a `pid_file` attribute.

dbutenhof (approved):

portante (commented):
Ready for review!

lib/pbench/agent/tool_meister_stop.py
```diff
@@ -65,6 +65,10 @@ def __init__(self, spec: str, benchmark_run_dir: Path, def_host_name: str):
...
+            self.host = "localhost"
```
ndokos, line 72: Is this enough? What bothers me is the asymmetry between `tool_meister_start` and `tool_meister_stop`. `start` seems to worry about a bind host name that is *not* `localhost` but is mapped to 127.0.0.1 (see  the comment in l.649 of `tool_meister_start`- not sure what the config file will look like in that case). But the above assumes that `localhost` will *always* be the proper name if redis is locally managed. Why the difference? Why not force the bind host name to be `localhost` when starting? Am I right to worry or am I confused?

dbutenhof: For "bind" address we generally want to make sure we have something that's externally reachable; but for some reason we *also* tell Redis to listen on `localhost`. The code at line 649 is avoiding a failure if `localhost` is already specified (in some alias), which Redis doesn't like. (That is, we don't *add* `localhost` if it's already there.)

In this case, where we know that we want to connect to the local Redis, `localhost` is a convenience.

lib/pbench/test/unit/agent/test_tool_meister_stop.py
```diff
@@ -0,0 +1,40 @@
...
+        tmdir = rundir / "tm"
+        tmdir.mkdir()
+        # ... containing a "redis.pid" file.
+        pidfile = tmdir / "redis.pid"
+        pidfile.write_text("12345")
+
+        rs = RedisServer("", rundir, "notme.example.com")
+        assert (
+            rs.locally_managed()
+        ), "Populated run directory did not result in a 'locally managed' RedisServer instance"
```
dbutenhof, line 23: "Populated"? Since we actually care about `redis.pid` this seems unnecessarily indirect.

lib/pbench/test/unit/agent/test_tool_meister_stop.py
```diff
@@ -0,0 +1,40 @@
...
+    def test_not_locally_managed(self, tmp_path):
+        # Empty benchmark run directory indicates not locally managed.
+        rundir = tmp_path / "empty-run-dir"
+        rundir.mkdir()
+
+        rs_host = "redis.example.com"
+        rs = RedisServer(f"{rs_host}:4343", rundir, "notme.example.com")
+        assert (
+            not rs.locally_managed()
+        ), "Empty run directory still resulting in a 'locally managed' RedisServer instance"
```
dbutenhof, line 37: "still resulting in" vs "did not result in" above seem unnecessarily divergent in tense. Maybe "RedisServer incorrectly inferred a locally managed instance from an empty run directory"??

dbutenhof (dismissed):

lib/pbench/agent/tool_meister_stop.py
portante: > For "bind" address we generally want to make sure we have something that's externally reachable; but for some reason we _also_ tell Redis to listen on `localhost`.

It is not "for some reason," it is for _this_ reason: when `pbench-tool-meister-start` is orchestrating the stand-up of the Tool Meister sub-system, we use `localhost` to talk to the Redis server so that the disconnected `pbench-tool-meister-stop` does not have to guess at where the Redis server is listening.  And we don't have the Redis server listen on `0.0.0.0` to avoid any potential accidental conflicts.

> The code at line 649 is avoiding a failure if `localhost` is already specified (in some alias), which Redis doesn't like. (That is, we don't _add_ `localhost` if it's already there.)

This points out a change that should be made to `pbench-tool-meister-start` in another PR, namely, that the addition of `localhost` should only be performed if it orchestrated the sub-system creation.

> In this case, where we know that we want to connect to the local Redis, `localhost` is a convenience.

It is more than a convenience because communicating from `pbench-tool-meister-start` to `pbench-tool-meister-stop` the IP address that should be used would involve a lot more work (don't just parse redis.conf because we might not be able to reach it from the other IP address) and locally managed instances using `localhost` is very natural and efficient.

lib/pbench/test/unit/agent/test_tool_meister_stop.py
portante: We do care about `redis.pid`, but that must be in a sub-directory, `tm`, and so with the two things that were created I used populated.

lib/pbench/agent/tool_meister_stop.py
webbnh: I have to ask...do we take steps to ensure that there is no `redis.pid` file when we are _not_ orchestrating the Redis server?  (That is, when we start running, do we, e.g., clear out the run directory?...or, could there possibly be a `redis.pid` left over from a previous run?)

[Unresolving.]

ndokos (approved):

lib/pbench/agent/tool_meister_stop.py
portante: Each run directory is unique, so there is no previous run to consider.  The `pbench-tool-meister-start` creates the `tm` directory, but Redis is the responsible party for creating the `redis.pid` file.  Only when `pbench-tool-meister-start` is responsible for orchestrating Redis is it directed in the `redis.conf` file to create it at `${benchmark_run_dir}/tm/redis.pid`.

So unless somebody is very clever and creates the expected run directory, `tm` sub-directory, and directs their Redis instance to create the pid file there, we are good.

PR #2865, opened 2022-05-25
First pass at ToC (Table of Contents) API documentation

portante:
Let's collaborate on documenting the API we want for the Table of Contents APIs going forward in this PR.


docs/API.md
```diff
@@ -0,0 +1,26 @@
...
+ * {{*path*}} (optional): E.g.{{{}/1-none-bidirectional-64B-128flows-0.002pct_drop/sample2/tools-default/perf188{}}}
+ ** Note that the {{*path*}} must contain a leading slash ("{{{}/{}}}"), which represents the "root" directory of the tar ball contents
+ * {{*HEAD*}} or {{*GET*}} on API URI {{/api/v1/inventory/<{*}dataset{*}>}} (no path specified)
+ ** If dataset does not exist, returns 404
+ ** If dataset is not in inventory, but exists (i.e. the tar ball exists, but is not unpacked)
+ *** returns 307 to indicate dataset is being loaded, URI returned is a dataset load operation object (TBD) where HEAD on that object returns a 201 when it is loaded, and 307 to itself indicating still loading
+ ** If dataset is in inventory, returns 200
+ * {{*HEAD*}} on API URI {{/api/v1/inventory/<{*}dataset{*}><{*}path{*}>}} (path _*IS*_ specified)
+ ** Returns a 400 if the {{*dataset*}} does not exist
+ ** Returns a 404 if the {{*path*}} does not exist
```
webbnh, line 18: Is this distinction useful?  Why not just return `404` in both cases, with a different message text?

(It seems weird to me that `HEAD` without a path returns `404` if the dataset does not exit, but it returns `400` for the _same_ failure if you add the path.)

docs/API.md
```diff
@@ -0,0 +1,26 @@
...
+ ** When {{*path*}} exists, returns a status 200, headers contain metadata about the object
```
webbnh, line 19: I don't have the experience to appreciate the difference, but do we really want to return the metadata in headers instead of in a nice JSON object in the payload?

docs/API.md
```diff
@@ -0,0 +1,26 @@
...
+ * {{*GET*}} on API URI {{/api/v1/inventory/<{*}dataset{*}><{*}path{*}>}} (path _*IS*_ specified)
+ ** Returns a 400 if the {{*dataset*}} does not exist
+ ** Returns a 404 if the {{*path*}} does not exist
+ ** When {{*path*}} exists, headers contain metadata about the object
+ *** If symlink, status 302 with payload a URL to symlinked object
```
webbnh, line 24: Is it really important to inform the client that the directory entry is a symbolic link?  Will anybody care?  Or, will this just generate an extra step which someone has to follow to get to the contents of the linked-to file?

That is, perhaps we should return enough information with the directory contents list for the client to determine whether a given entry is a subdirectory, a file, or a symlink.  If we do that, and the client then does a `GET` on the symlink, we can return the linked-to contents, without having to do the `302` dance.

(The best reason for getting rid of the `302` return, is that I cannot figure out whether `302` is the right value or whether it should be one of the other 300-level statuses....)

docs/API.md
```diff
@@ -0,0 +1,26 @@
+Here are two URLs representing the "perf188" directory object of an unpacked tar ball.  The "old" is the existing v0.69 Pbench Server way of exposing the unpacked tar ball contents, the "new" is an example the proposal.  See below for more concrete examples.
+
+{*}Old{*}: {{/incoming/perf102.perf.lab.eng.bos.redhat.com/trafficgen_config0_2021-12-01T07:01:45/1-none-bidirectional-64B-128flows-0.002pct_drop/sample2/tools-default/perf188}}
+
+{*}New{*}: {{/api/vi/inventory/trafficgen_config0_2021-12-01T07:01:45/1-none-bidirectional-64B-128flows-0.002pct_drop/sample2/tools-default/perf188}}
+
+Components of new API URI {{{}/api/v1/inventory/<{*}dataset{*}>{}}}[ {{<{*}path{*}>}} ]:
+ * {{{}*dataset*{}}}: E.g. {{trafficgen_config0_2021-12-01T07:01:45}}
+ * {{*path*}} (optional): E.g.{{{}/1-none-bidirectional-64B-128flows-0.002pct_drop/sample2/tools-default/perf188{}}}
+ ** Note that the {{*path*}} must contain a leading slash ("{{{}/{}}}"), which represents the "root" directory of the tar ball contents
```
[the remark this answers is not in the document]
dbutenhof: In theory, that's great; in practice it means figuring out what we want and then mapping the actual Elasticsearch TOC (and physical directory structure) into that model.

Once the TOC includes URIs for artifact access, this all becomes completely self describing and we can change our storage representation in the future without changing the API.

In terms of documentation, however, can we consider saying something more like "root of the dataset TOC"?? That is, each *file* reported by `/datasets/contents/<dataset>` has a URL which, when passed to this API, will result in a byte stream representing that dataset artifact.

docs/API.md
```diff
@@ -0,0 +1,26 @@
+Here are two URLs representing the "perf188" directory object of an unpacked tar ball.  The "old" is the existing v0.69 Pbench Server way of exposing the unpacked tar ball contents, the "new" is an example the proposal.  See below for more concrete examples.
+
+{*}Old{*}: {{/incoming/perf102.perf.lab.eng.bos.redhat.com/trafficgen_config0_2021-12-01T07:01:45/1-none-bidirectional-64B-128flows-0.002pct_drop/sample2/tools-default/perf188}}
+
+{*}New{*}: {{/api/vi/inventory/trafficgen_config0_2021-12-01T07:01:45/1-none-bidirectional-64B-128flows-0.002pct_drop/sample2/tools-default/perf188}}
+
+Components of new API URI {{{}/api/v1/inventory/<{*}dataset{*}>{}}}[ {{<{*}path{*}>}} ]:
+ * {{{}*dataset*{}}}: E.g. {{trafficgen_config0_2021-12-01T07:01:45}}
+ * {{*path*}} (optional): E.g.{{{}/1-none-bidirectional-64B-128flows-0.002pct_drop/sample2/tools-default/perf188{}}}
```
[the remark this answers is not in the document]
dbutenhof: If we need to do urlencoding when generating the URLs in the TOC API, we can. It's a point: we don't really restrict the character set in the several components of the path that are directly user-specified; so either we need to do that, or we need to be prepared for weird stuff.

docs/API.md
```diff
@@ -0,0 +1,26 @@
...
+ ** Note that the {{*path*}} must contain a leading slash ("{{{}/{}}}"), which represents the "root" directory of the tar ball contents
+ * {{*HEAD*}} or {{*GET*}} on API URI {{/api/v1/inventory/<{*}dataset{*}>}} (no path specified)
+ ** If dataset does not exist, returns 404
+ ** If dataset is not in inventory, but exists (i.e. the tar ball exists, but is not unpacked)
+ *** returns 307 to indicate dataset is being loaded, URI returned is a dataset load operation object (TBD) where HEAD on that object returns a 201 when it is loaded, and 307 to itself indicating still loading
```
dbutenhof, line 14: I really don't like this 307 stuff; it means inventing the "status" object query as well, which is outside the scope of the artifact access. (I don't really think that "temporary redirect" is the right mechanism for asynchronous status, either, though that's a different issue.)

docs/API.md
dbutenhof: Both are "missing resources"; 400 isn't appropriate for either. The problem is that the HTTP status code space is woefully shallow. We can help with good message text.

docs/API.md
```diff
@@ -0,0 +1,26 @@
...
+ * {{*HEAD*}} on API URI {{/api/v1/inventory/<{*}dataset{*}><{*}path{*}>}} (path _*IS*_ specified)
+ ** Returns a 400 if the {{*dataset*}} does not exist
+ ** Returns a 404 if the {{*path*}} does not exist
+ ** When {{*path*}} exists, returns a status 200, headers contain metadata about the object
+ * {{*GET*}} on API URI {{/api/v1/inventory/<{*}dataset{*}><{*}path{*}>}} (path _*IS*_ specified)
+ ** Returns a 400 if the {{*dataset*}} does not exist
+ ** Returns a 404 if the {{*path*}} does not exist
+ ** When {{*path*}} exists, headers contain metadata about the object
+ *** If symlink, status 302 with payload a URL to symlinked object
+ *** If directory, status 200, returns payload of directory contents as a JSON object
```
dbutenhof, line 25: Again; you get the contents of a directory through `/datasets/contents`; this API is for returning the byte stream of a leaf file, not for metadata. And `/datasets/contents` should return a URI link for each file/directory, either pointing through `/datasets/contents` (which we need to refactor to `/datasets/contents/<dataset>/<dir-path>`) or to this API for the final byte stream.

portante:
Replaced by #2891.

PR #2866, opened 2022-05-25
Backport container build fixes and generate-token tweaks

webbnh:
This PR contains two commits which backport the changes from #2858 and #2862, respectively.


dbutenhof (approved):
I won't object to including the `pbench-generate-token` change, however in practice it's unlikely anyone outside our team will have the opportunity to use 0.71 against a 1.0 server so deferring this fix to 0.72 would probably have been fine.

webbnh:
Peter pointed out that v0.71 is likely to have a long life in the field which will overlap the life of the v1.0 server, so having this fix in place would be helpful (and, its small and low-risk...so, why not?).

portante (approved):

PR #2867, opened 2022-05-25
More backports for `b0.71`

portante:
Provides back-ports of PRs #2863, #2856, #2860, and #2870.

----

Note that with the backport of #2860, where the `resolve_benchmark_bin` function was removed from `agent/base`, we still have three deprecated commands, `pbench-dbench`, `pbench-iozone`, and `pbench-netperf`, which were referencing that function.  Since we don't want to introduce the `which` RPM as a dependency, we just use `command -v` in each of those commands instead.

### DIFF
--- a/lib/pbench/agent/tool_meister_stop.py
+++ b/lib/pbench/agent/tool_meister_stop.py
@@ -65,6 +65,11 @@ class RedisServer(RedisServerCommon):
             )
         except FileNotFoundError:
             pass
+        else:
+            # The redis.pid file exists in the "tm" directory, which means this
+            # Redis server is locally managed.  Communication with it will
+            # always be through the "localhost" interface.
+            self.host = "localhost"
 
     def locally_managed(self) -> bool:
         return self.pid_file is not None
@@ -286,7 +291,7 @@ def main():
     parser.add_argument(
         "--redis-server",
         dest="redis_server",
-        default=os.environ.get("PBENCH_REDIS_SERVER", None),
+        default=os.environ.get("PBENCH_REDIS_SERVER", ""),
         help=(
             "Use an existing Redis server specified by <hostname>:<port>;"
             " implies the use of an existing Tool Data Sink and Tool Meisters"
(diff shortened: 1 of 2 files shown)


agent/bench-scripts/driver/linpack
```diff
@@ -38,15 +39,19 @@ kmp_affinity_args_def="nowarnings,compact,1,0,granularity=fine"
 kmp_affinity_args=""
 numactl_args=""
 
-# Output directory required.
+# Output directory, defaults to the current working directory if not provided.
 output_dir=""
 
+# The linpack driver always perform pre-checks. When requested, the driver will
```
ndokos, line 45: "perrforms"

ndokos (dismissed):
Typo in a comment. There is also a typo in the commit message of the #2863 backport (mdoule instead of module), but I certainly won't insist that either of those be fixed here.

dbutenhof (approved):

agent/bench-scripts/test-bin/uperf
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+echo "$0 $*" >> $_testlog
+
+# Mock uperf command behaviors.
+if [[ "${1}" == "-V" ]]; then 
+	echo "Uperf Version ${_BM_UPERF_VER:-1.0.42}"
+	echo "... garbage to be ignored ..."
+fi
```
webbnh, line 9: Based on what you said about the analogous portion of `agent/bench-scripts/test-bin/fio`, we should be invoking `exit 0` at the end of the `if` statement, here, instead of falling through to the common exit point which might not exit "successfully".

webbnh (dismissed):
Looks OK to me (although, I skipped the unit tests...).

I have one comment which is probably better applied to `main`, but I only noticed it here.

agent/bench-scripts/driver/linpack
webbnh: But...with only one `r`.  🤣

dbutenhof (approved):

ndokos (approved):

PR #2868, opened 2022-05-25
Documentation for helper methods of Pbench_dashboard

shubham-html-css-js:
Documented the helper methods using jsDoc.A new run script added.The user can now run `npm run jsdoc` to generate the HTML version of documentation.
The documentation format looks like this:-

Description of the Function
Datatype of the parameters passed into the function along with a short description of each parameter
Return type of the function along with a small description

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -42,7 +42,7 @@
     "build": "react-app-rewired build",
     "test": "react-app-rewired test",
     "eject": "react-scripts eject",
-    "doc":"jsdoc -c jsdoc.json"
+    "doc": "jsdoc -c jsdoc.json"
   },
   "eslintConfig": {
     "extends": [
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -13,12 +13,12 @@
     "@testing-library/react": "^12.1.4",
     "@testing-library/user-event": "^13.5.0",
     "axios": "^0.26.1",
+    "babel-jest": "^27.5.1",
     "classnames": "^2.3.1",
     "customize-cra": "^1.0.0",
-    "js-cookie": "^3.0.1",
-    "babel-jest": "^27.5.1",
     "gulp": "^4.0.2",
     "jest": "^27.5.1",
+    "js-cookie": "^3.0.1",
     "less-watch-compiler": "^1.16.3",
     "patternfly": "^3.9.0",
     "react": "^17.0.2",
@@ -41,7 +41,8 @@
     "start": "react-app-rewired start",
     "build": "react-app-rewired build",
     "test": "react-app-rewired test",
-    "eject": "react-scripts eject"
+    "eject": "react-scripts eject",
+    "doc":"jsdoc -c jsdoc.json"
   },
   "eslintConfig": {
     "extends": [
@@ -63,6 +64,7 @@
   },
   "devDependencies": {
     "customize-cra-less-loader": "^2.0.0",
+    "jsdoc": "^3.6.10",
     "less": "^4.1.2",
     "less-loader": "^11.0.0"
   }
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -38,11 +38,11 @@
     "web-vitals": "^2.1.4"
   },
   "scripts": {
-    "start": "react-app-rewired start",
     "build": "react-app-rewired build",
-    "test": "react-app-rewired test",
+    "doc": "jsdoc -c jsdoc.json",
     "eject": "react-scripts eject",
-    "doc": "jsdoc -c jsdoc.json"
+    "start": "react-app-rewired start",
+    "test": "react-app-rewired test"
   },
   "eslintConfig": {
     "extends": [
--- a/dashboard/src/utils/dateFunctions.js
+++ b/dashboard/src/utils/dateFunctions.js
@@ -1,11 +1,29 @@
+/**
+ * Bump Todate ahead by specified number of days
+ * @function
+ * @param {Date} date - Date to be bumped
+ * @param {number} numberOfDays - Number of days the date is to be bumped
+ * @returns {Date} - New Bumped date
+ */
 export const bumpToDate = (date, numberOfDays) => {
   return date.setUTCDate(date.getUTCDate() + numberOfDays);
 };
 
+/**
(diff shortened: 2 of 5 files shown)


dashboard/package.json
```diff
@@ -41,7 +41,8 @@
     "start": "react-app-rewired start",
     "build": "react-app-rewired build",
     "test": "react-app-rewired test",
-    "eject": "react-scripts eject"
+    "eject": "react-scripts eject",
+    "doc":"jsdoc -c jsdoc.json"
```
dbutenhof, line 45: I like that you sorted the library dependencies. Should we sort the script names, too? 😆 

Also, you should include the space after `"doc": ` like all the other script targets, for consistency.

dashboard/src/utils/dateFunctions.js
```diff
@@ -1,11 +1,29 @@
...
+ * @returns {Date} - New Bumped date
+ */
 export const bumpToDate = (date, numberOfDays) => {
   return date.setUTCDate(date.getUTCDate() + numberOfDays);
 };
 
+/**
+ * Takes a string date and convert it into UTC Date object
+ * @function
+ * @param {string} strDate - Date in string format
```
dbutenhof, line 15: You should say something here about the expected format of the string; for example, we're depending on it *not* having any timezone information, or the `+ "Z"` will result in an error.

dashboard/src/utils/dateFunctions.js
```diff
@@ -1,11 +1,29 @@
...
+ * @returns {Date} - UTC Date object
+ */
 export const dateFromUTCString = (strDate) => {
   return new Date(strDate + "Z");
 };
 
+/**
+ * Set today's date to Midnight UTC
```
dbutenhof, line 23: Returns a `Date` object set to midnight UTC. "Today's date" doesn't change. 😆

dashboard/src/utils/filterDataset.js
```diff
@@ -1,3 +1,12 @@
+/**
+ * Filter the Dataset based on Date Range or Dataset Name or both
```
dbutenhof, line 2: "Filter a list of datasets"

dashboard/src/utils/helper.js
```diff
@@ -5,6 +5,13 @@ export const uid = () => {
   return head + tail;
 };
 
+/**
+ * Check if password entered by user is valid or not
+ * @function
+ * @param {string} password - Entered email
+ * @param {number} passwordLength - Length of Password
+ * @returns {boolean} - Password valid or not
```
dbutenhof, line 13: For boolean returns, it's good to be more specific about what the values mean. "`true` if the password is valid", for example.

dbutenhof (commented):
Great! I think some of the descriptions need a bit of work, but thanks for adding this.

dashboard/package.json
shubham-html-css-js: I guess the scripts are also sorted ,in this case not alphabetically,but priority wise 😆 ,don't you think so?

dashboard/src/utils/helper.js
```diff
@@ -10,7 +10,7 @@ export const uid = () => {
  * @function
  * @param {string} password - Entered email
  * @param {number} passwordLength - Length of Password
- * @returns {boolean} - Password valid or not
+ * @returns {boolean} - Password valid or not,true if the password is valid,false otherwise
```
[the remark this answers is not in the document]
shubham-html-css-js: done

dbutenhof (dismissed):

dashboard/src/utils/dateFunctions.js
```diff
@@ -1,11 +1,30 @@
...
+ * @function
+ * @param {string} strDate - Date in string format
+ * @returns {Date} - UTC Date object
+ */
 export const dateFromUTCString = (strDate) => {
   return new Date(strDate + "Z");
 };
 
+/**
+ * Return Date object set to midnight UTC
```
webbnh, line 24: Perhaps something more like, "Return a Date object representing the current day, UTC, with the time set to midnight."  (Your description doesn't mention _which_ date is returned.)

Likewise you might want to add the word "current" to the return value description at line 26.

dashboard/src/utils/filterDataset.js
```diff
@@ -1,3 +1,12 @@
+/**
+ * Filter the List of Datasets based on Date Range or Dataset Name or both
```
webbnh, line 2: Looking at the body of the function, I'm not seeing how it could be one or the other -- it looks like the selection criteria is _both_ contains `searchKey` _and_ date within the specified bounds.  I'm not seeing any provision for having parameters omitted or otherwise unspecified.  So, you should probably tighten up this description.

dashboard/src/utils/helper.js
```diff
@@ -5,6 +5,13 @@ export const uid = () => {
   return head + tail;
 };
 
+/**
+ * Check if password entered by user is valid or not
+ * @function
+ * @param {string} password - Entered email
```
webbnh, line 11: Which is it, the "password" or the "email"?

dashboard/src/utils/helper.js
```diff
@@ -5,6 +5,13 @@ export const uid = () => {
   return head + tail;
 };
 
+/**
+ * Check if password entered by user is valid or not
```
webbnh, line 9: It would be good to provide a least a little more here.  For instance, "Determine whether the specified password meets the requirements for a valid password, including length, presence of upper-case letters, digits, and special characters."

dashboard/src/utils/helper.js
```diff
@@ -5,6 +5,13 @@ export const uid = () => {
...
+ * @function
+ * @param {string} password - Entered email
+ * @param {number} passwordLength - Length of Password
```
webbnh, line 12: The parameter is badly named -- it should be `minimumLength` -- so your description needs to make up for that!

webbnh (approved):
This is great stuff, Shubham!  Thanks for adding this.

There are a couple of bits which need correcting, and I have some other suggestions.  (English is hard!)

webbnh (approved):
👍

dbutenhof (approved):

PR #2869, opened 2022-05-25
Dataset name in the URI for /datasets/details endpoint

npalaska:
URI parameterization of the dataset name in the `dataset/details` endpoint

### DIFF
--- a/lib/pbench/server/api/__init__.py
+++ b/lib/pbench/server/api/__init__.py
@@ -71,7 +71,7 @@ def register_endpoints(api, app, config):
     )
     api.add_resource(
         DatasetsDetail,
-        f"{base_uri}/datasets/detail",
+        f"{base_uri}/datasets/detail/<string:name>",
         resource_class_args=(config, logger),
     )
     api.add_resource(
--- a/lib/pbench/server/api/__init__.py
+++ b/lib/pbench/server/api/__init__.py
@@ -71,7 +71,7 @@ def register_endpoints(api, app, config):
     )
     api.add_resource(
         DatasetsDetail,
-        f"{base_uri}/datasets/detail/<string:name>",
+        f"{base_uri}/datasets/detail/<string:dataset>",
         resource_class_args=(config, logger),
     )
     api.add_resource(
--- a/lib/pbench/server/api/resources/query_apis/datasets_detail.py
+++ b/lib/pbench/server/api/resources/query_apis/datasets_detail.py
@@ -43,7 +43,7 @@ class DatasetsDetail(ElasticBase):
 
         {
             "user": "username",
-            "dataset_name": "dataset-name",
+            "name": "dataset-name",
             "start": "start-time",
             "end": "end-time",
             "metadata": ["seen", "saved"]
@@ -56,7 +56,7 @@ class DatasetsDetail(ElasticBase):
                 data. If "user": None is specified, then only public datasets
                 will be returned.
 
-            "dataset_name" is the name of a Pbench agent dataset (tarball).
+            "name" is the name of a Pbench agent dataset (tarball).
 
             "start" and "end" are time strings representing a set of Elasticsearch
                 run document indices in which the dataset will be found.
@@ -69,7 +69,7 @@ class DatasetsDetail(ElasticBase):
             requested set of metadata to the postprocess method.
         """
         user = json_data.get("user")
-        dataset_name = json_data.get("dataset_name")
+        dataset_name = json_data.get("name")
         start = json_data.get("start")
         end = json_data.get("end")
 
--- a/lib/pbench/server/api/resources/query_apis/datasets_detail.py
+++ b/lib/pbench/server/api/resources/query_apis/datasets_detail.py
@@ -24,7 +24,9 @@ class DatasetsDetail(ElasticBase):
             Schema(
                 Parameter("user", ParamType.USER, required=False),
                 Parameter("access", ParamType.ACCESS, required=False),
-                Parameter("name", ParamType.STRING, required=True, uri_parameter=True),
+                Parameter(
+                    "dataset", ParamType.STRING, required=True, uri_parameter=True
+                ),
                 Parameter("start", ParamType.DATE, required=True),
                 Parameter("end", ParamType.DATE, required=True),
                 Parameter(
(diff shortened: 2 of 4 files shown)


lib/pbench/server/api/__init__.py
```diff
@@ -71,7 +71,7 @@ def register_endpoints(api, app, config):
     )
     api.add_resource(
         DatasetsDetail,
-        f"{base_uri}/datasets/detail",
+        f"{base_uri}/datasets/detail/<string:dataset_name>",
```
dbutenhof, line 74: I'd like to keep these names consistent. I already used `dataset` for `/datasets/metadata`, and I don't see any value in the longer name.

lib/pbench/server/api/resources/query_apis/datasets_detail.py
```diff
@@ -24,7 +24,6 @@ def __init__(self, config: PbenchServerConfig, logger: Logger):
             Schema(
                 Parameter("user", ParamType.USER, required=False),
                 Parameter("access", ParamType.ACCESS, required=False),
-                Parameter("name", ParamType.STRING, required=True),
```
dbutenhof, line 27: Although with my new schema changes the `dataset` name will move to the `uri_schema`, right now `_dispatch` merges them, and I don't really like removing it from the schema entirely. (Even though we don't drop unrecognized parameters, so it's not rejected, and Flask will reject the URI mapping if it's not provided.) I'd rather keep it in the schema if only for documentation purposes!

lib/pbench/server/api/resources/query_apis/datasets_detail.py
```diff
@@ -69,15 +68,15 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
             requested set of metadata to the postprocess method.
         """
         user = json_data.get("user")
-        name = json_data.get("name")
+        dataset_name = json_data.get("dataset_name")
         start = json_data.get("start")
         end = json_data.get("end")
```
dbutenhof, line 73: I definitely do like the idea of converting all of these to use SQL validation and the index map; we'll drop three of these four primary parameters!

dbutenhof (changes requested):

webbnh (approved):
Looks OK to me.

dbutenhof:
By the way: please let's try to develop the habit of referencing the JIRA ticket in the git log message. E.g.,

```
Dataset name in the URI for /datasets/details endpoint

PBENCH-671

Add the dataset resource name to the URI as /datasets/details/<dataset>.
```

lib/pbench/server/api/resources/query_apis/datasets_detail.py
npalaska: Yeah, the problem right now is if keep the `name` schema for the documentation purpose then we also have to keep the `required` parameter as `True`. But then the `test_missing_keys` test fails since we are not actually providing the `name` in the json payload (for something that is required), it is coming from the URI parameterization. Also, `name` parameter validation will always be true in this case because without having a `name` parameter, flask will return 404 on our `dataset/details/` endpoint.

npalaska: Well, we have `uri_parameter` as one of the parameters in the schema that we can set to True. So Nevermind :smiley:

dbutenhof: Yeah; and with my schema change (TBD whether you merge or I do 😆 ) the uri parameters move to their own schema object so it'll be even more clear. (At some point I'd love to rewrite `Common` to take better advantage of all this.)

lib/pbench/server/api/resources/query_apis/datasets_detail.py
```diff
@@ -43,7 +43,7 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
 
         {
             "user": "username",
-            "name": "dataset-name",
+            "dataset_name": "dataset-name",
```
dbutenhof, line 46: `s/_name//`

lib/pbench/server/api/resources/query_apis/datasets_detail.py
```diff
@@ -56,7 +56,7 @@ def assemble(self, json_data: JSON, context: CONTEXT) -> JSON:
                 data. If "user": None is specified, then only public datasets
                 will be returned.
 
-            "name" is the name of a Pbench agent dataset (tarball).
+            "dataset_name" is the name of a Pbench agent dataset (tarball).
```
dbutenhof, line 59: `s/_name//`

dbutenhof (approved):
Your API parameter docstring is now inconsistent ...

npalaska:
> Your API parameter docstring is now inconsistent ...

Ah yes incoming another commit

lib/pbench/server/api/__init__.py
```diff
@@ -71,7 +71,7 @@ def register_endpoints(api, app, config):
     )
     api.add_resource(
         DatasetsDetail,
-        f"{base_uri}/datasets/detail",
+        f"{base_uri}/datasets/detail/<string:name>",
```
dbutenhof, line 74: Wait; was I imagining it, or did you just rewrite history? Your first commit had been `dataset_name` and I suggested `dataset` which is what I'd used for metadata, for consistency. Now you have two commits in the PR and both use "name", which is confusing. I'd prefer that we make the URI mapping consistent across the APIs even though the template tag is more or less private to the API class; given that "dataset" is already in the repo, I'd prefer sticking with that. (It seemed more expressive than "name" at the time, and I don't have a better suggestion.) Besides, with my schema changes we can make this `DATASET` *type*, and the converted value will actually *be* the `Dataset` reference, not the "name" (which feeds into the idea of using the index map instead of the date range, along with authorizing against the dataset... none of which, granted, is or should be part of this PR...)

npalaska: Oh! Okay I thought you are suggesting to use name instead of dataset_name. I made the change and squashed all the changes and that rewrote the history (I guess) but yeah using dataset make sense instead of name.

dbutenhof (approved):

lib/pbench/server/api/resources/query_apis/datasets_detail.py
dbutenhof: Oh, is this what was confusing? That good old `ex` command structure could be easily misinterpreted if you look at it quickly! That's "replace the string `_name` by nothing"; in other words, leave just `dataset`. Ah, well. Sorry for the brevity!

webbnh (approved):
👍

PR #2871, opened 2022-05-26
Fetch endpoints API

MVarshini:
[PBENCH-198](https://issues.redhat.com/browse/PBENCH-198) 
During startup, we access the `/api/v1/endpoints` API from the associated Pbench server, caching to results where all dashboard code can use it. (For example, on `window.endpoints`.)

In every service/action the dashboard instead of constructing a URL directly like `http://<server>/api/datasets/list` would find the URL through the `api` subobject in the endpoints configuration: e.g., `window.endpoints.api.datasets_list`.


dashboard/src/actions/authActions.js
```diff
@@ -12,7 +11,8 @@ export const makeLoginRequest = (details, navigate) => async (dispatch, getState
       type: TYPES.USER_NOTION_ALERTS,
       payload: [],
     });
-    const response = await API.post(API_ROUTES.LOGIN, {
+    const endpoints = {...getState().apiEndpoint.endpoints};
+    const response = await API.post(endpoints?.api?.login, {
```
dbutenhof, line 15: I'm not sure about the `?.` usage here. What do we *want* to happen if we somehow get here without having loaded the endpoints? I suspect that should be a failure, rather than trying the `API.post` with `undefined` as a URL.

I understand that this is a "safe" general pattern for navigating an object, and doesn't hurt if all the expected references are there; but this makes me think about what happens if we can't load. The App component is throwing up a TOAST message; but maybe we should try to go further and completely lock up the UI with the warning message until they refresh (and it loads successfully)?

dashboard/src/actions/endpointAction.js
```diff
@@ -0,0 +1,25 @@
...
+
+    if (!response.ok) {
+      throw new Error(`HTTP error! status: ${response.status}`);
+    }
+    const data = await response.json();
+    window.endpoints = data;
+    dispatch({
+      type: TYPES.SET_ENDPOINTS,
+      payload: { ...data },
+    });
```
dbutenhof, line 15: So we're saving the endpoints object both at `window.endpoints` and at `apiEndpoings` in Redux. Since your components are using Redux to access the configuration, I don't think we need the `window.endpoints` copy.

dbutenhof (dismissed):
I have a couple of minor comments, including the fact that you're still storing the endpoints in `window.endpoints` even though you've changed all the references to your new Redux store. But I don't see anything that would keep it from working, so I'm going to approve. I'm anxious to get this in.

dashboard/src/actions/endpointAction.js
webbnh, line 15: Perhaps orthogonal to Dave's question, I'll just note that what we are saving in the `window` object is a _reference_ to the endpoints object, while what we are passing to `dispatch()` is a _copy_ of it.  It seems like a reference should work in both instances.

dashboard/src/actions/endpointAction.js
```diff
@@ -0,0 +1,25 @@
...
+  } catch (error) {
+    dispatch({
+      type: TYPES.SHOW_TOAST,
+      payload: {
+        variant: "danger",
+        title: "Something went wrong, try again later",
+      },
+    });
```
webbnh, line 23: "Something went wrong", but we're not going to give any hint as to what??  (Can't we string-ize `error` and include it in the message?  That would be much more useful than the suggestion to "try again later"....)

dashboard/src/modules/components/TableComponent/index.jsx
```diff
@@ -47,9 +48,11 @@ const TableWithFavorite = () => {
...
 
   useEffect(() => {
-    dispatch(fetchPublicDatasets());
-    dispatch(getFavoritedDatasets());
-  }, [dispatch]);
+    if (Object.keys(endpoints).length > 0) {
+      dispatch(fetchPublicDatasets());
+      dispatch(getFavoritedDatasets());
+    }
+  }, [dispatch, endpoints]);
```
webbnh, line 55: Do we need to check for the existence of the endpoint list here?  That is, can we actually get this far if we cannot load the endpoints?

(If we can and we do, then aren't there a bunch of other places which need this protection, too?  This is the first one that I've noted in this PR.)

dashboard/src/reducers/endpointReducer.js
```diff
@@ -0,0 +1,20 @@
...
+};
+
+const EndpointReducer = (state = initialState, action = {}) => {
+  const { type, payload } = action;
+  switch (type) {
+    case SET_ENDPOINTS:
+      return {
+        ...state,
+        endpoints: Object.assign({}, payload),
+      };
```
webbnh, line 14: Just out of curiosity, why did you opt to use `Object.assign()` here instead of a simple, shallow copy (which I _think_ does the same thing):
```:
      return {
        ...state,
        endpoints: {...payload},
      };
```
and, because it's my current hobby horse (😀), I have to ask:  would it work to make `endpoints` be a _reference_ to `payload` instead of a copy of it?

dashboard/src/reducers/endpointReducer.js
```diff
@@ -0,0 +1,20 @@
...
+    default:
+      return state;
+  }
+};
```
webbnh, line 18: Are you expecting to add more cases to this `switch` statement?  Otherwise, I think we would be better served by an `if` statement, or maybe even a ternary expression.

webbnh (dismissed):
These changes seem good, however, I'm continuing my crusade against unnecessary object copying.  😁  And, there are a few other items for your consideration.

dashboard/public/index.html
```diff
@@ -10,16 +10,6 @@
...
-    <link
-      rel="stylesheet"
-      type="text/css"
-      href="/node_modules/patternfly/dist/css/patternfly.css"
-    />
-    <link
-      rel="stylesheet"
-      type="text/css"
-      href="/node_modules/patternfly/dist/css/patternfly-additions.css"
-    />
```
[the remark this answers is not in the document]
MVarshini: These stylesheets are being loaded in App.js. These are duplicates so removed them.

dashboard/src/App.js
```diff
@@ -6,12 +6,18 @@ import * as APP_ROUTES from "./utils/routeConstants";
 import AuthComponent from './modules/components/AuthComponent';
 import '@patternfly/patternfly/patternfly.css';
 import MainLayout  from 'modules/containers/MainLayout';
+import { useDispatch,  } from 'react-redux';
```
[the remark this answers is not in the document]
MVarshini: Yes, it does.

dashboard/src/reducers/endpointReducer.js
MVarshini: If the new state is different, the reducer must create new object, and making a `copy ` is a way to describe the unchanged part.  When there are only a few properties to keep, we can specify unchanged properties instead of copying.

dashboard/src/actions/endpointAction.js
MVarshini: Yes, it works!

dashboard/package.json
```diff
@@ -37,13 +40,16 @@
...
     "build": "react-app-rewired build",
-    "doc": "jsdoc -c jsdoc.json",
+    "server": "node server",
+    "dev": "npm run server & npm start",
+    "test": "react-app-rewired test",
     "eject": "react-scripts eject",
-    "start": "react-app-rewired start",
-    "test": "react-app-rewired test"
+    "doc": "jsdoc -c jsdoc.json",
   },
```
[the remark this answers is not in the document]
MVarshini: Followed  alphabetical ordering

dashboard/src/actions/endpointAction.js
MVarshini: Removed the `window.endpoints` copy

dashboard/package.json
```diff
@@ -16,10 +16,16 @@
     "babel-jest": "^27.5.1",
     "classnames": "^2.3.1",
     "customize-cra": "^1.0.0",
+<<<<<<< HEAD
+=======
+    "crypto-js": "^4.1.1",
+    "dotenv": "^16.0.1",
+>>>>>>> 67cefb6df (Fetch endpoints)
```
dbutenhof, line 23: You have an unresolved merge conflict here!

dashboard/package.json
```diff
@@ -40,10 +46,13 @@
   "scripts": {
     "build": "react-app-rewired build",
     "doc": "jsdoc -c jsdoc.json",
+    "dev": "npm run server & npm start",
```
dbutenhof, line 49: Alphabetically, `doc` and `dev` are reversed.

dashboard/src/actions/authActions.js
MVarshini: If `fetch endpoints` request is successful, _login_ or _register_ buttons would be enabled.

dashboard/package.json
MVarshini: Updated!

dashboard/package.json
MVarshini: Updated the order :smile:

dashboard/package.json
```diff
@@ -65,6 +74,8 @@
   "devDependencies": {
     "customize-cra-less-loader": "^2.0.0",
     "jsdoc": "^3.6.10",
+    "css-loader": "^6.7.1",
```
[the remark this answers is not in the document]
MVarshini: Done :+1:

dbutenhof (dismissed):
Unless my page cache is misbehaving (again), you may still have a few of the import list entries out of order; at this point I'm going to approve anyway since I think any remaining issues are trivial and I'd like to get this in.

dashboard/package.json
```diff
@@ -39,11 +42,14 @@
   },
   "scripts": {
     "build": "react-app-rewired build",
+    "dev": "npm run server & npm start",
     "doc": "jsdoc -c jsdoc.json",
     "eject": "react-scripts eject",
+    "server": "node server",
     "start": "react-app-rewired start",
-    "test": "react-app-rewired test"
+    "test": "react-app-rewired test"
```
webbnh, line 50: You have trailing whitespace on this line.

Prettier would probably take care of that...does that tool do the `package.json` file, or is it only applied to Javascript files?

Also, you can probably configure your IDE or your editor to strip trailing whitespace automatically when it saves your changes.

webbnh (approved):
Looks good, Varshini.

There's some trailing whitespace, but we don't have to hold this merge up for that.  However, I do have a question on whether `fetchEndpoints` should be a callable; but, we don't necessarily have to gate the merge on that, either.

dashboard/src/actions/endpointAction.js
```diff
@@ -0,0 +1,24 @@
+import * as TYPES from "./types";
+
+export const fetchEndpoints = () => async (dispatch) => {
```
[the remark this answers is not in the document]
MVarshini: Modified the code so that nested call is avoided.

dashboard/src/actions/authActions.js
```diff
@@ -65,58 +72,59 @@ export const setUserLoggedInState = (value) => async (dispatch) => {
...
-      type: TYPES.USER_NOTION_ALERTS,
-      payload: alerts,
-    });
-    dispatch({ type: TYPES.COMPLETED });
-  }
-};
+  };
 
-export const toggleSignupBtn = ( isDisbaled ) => async( dispatch ) => {
+export const toggleSignupBtn = (isDisbaled) => async (dispatch) => {
```
dbutenhof, line 118: `s/isDisbaled/isDisabled`?

dbutenhof (commented):
Plus editing the initial git log message with `git rebase -i` to include the `PBENCH-200` Jira reference, and a description of the change. 😉

dashboard/src/modules/components/AuthComponent/common-components.jsx
```diff
@@ -129,18 +129,16 @@ export const PasswordConstraints = (props) => {
...
               >
-                {
-                  constraint.name === "passwordLength" &&
-                  passwordLength
-                }{" "}{constraint.label}
+                {constraint.name === "passwordLength" && passwordLength}{" "}
+                {constraint.label}
               </HelperTextItem>
             </HelperText>
           );
```
webbnh, line 144: At line 137, should we be using the local variable, `variant`, that we defined at line 132 (and which we use at line 138)?

Also, what do lines 140-141 actually do?  They look to me like they produce a boolean value (possibly as a string), followed by a space character, followed by a string...what does the `<HelperTextItem>` tag do with that text?

webbnh (dismissed):
The changes look good, although I found a few more places where we seem to have a callable which returns a callable, where we could cut out the extra layer of call, and there's couple of other spots where I have coding suggestions.  And, I have a question.  But, none of these necessarily need to gate the merge.

However, before this PR is merged, someone should resolve the question posed in the PR description and update the description accordingly.  😀

dbutenhof:
> Let's resolve which Jira issue this should be tracked by, PBENCH-198 or PBENCH-200.

PBENCH-200 is a sub-task under PBENCH-198.

Technically, though, this PR encompasses *both* sub-tasks:
* PBENCH-199 is the nodeJS reflector I originally created in a separate PR, but which has been merged into this one.
* PBENCH-200 is the code to load and use `/api/v1/endpoints` instead of hardcoding the server address and URI schema.

So we could break this up, but at this point I think it makes more sense for me to just close out #2762 (which I'd planned to to when this merged) and handle both PBENCH-199 and PBENCH-200 here. (Or, alternatively, just to confuse things, we could dump/ignore the sub-tasks and simply link this to the PBENCH-198 story. 😏 )

dashboard/src/modules/components/AuthComponent/common-components.jsx
MVarshini: Lines 140-141:
This is with the UI. 

All other constraints are hard-coded except for the password length. It gets the value from redux store. It provides the `space` between `constraint.name` and `constraint.label`

dashboard/src/actions/datasetListActions.js
```diff
@@ -20,12 +21,12 @@ export const fetchPublicDatasets = () => async (dispatch) => {
...
-  fav_datasets = fav_datasets_text ? JSON.parse(fav_datasets_text) : [];
+  let favDatasets = [];
+  const favDatasetsText = localStorage.getItem("favorite_datasets");
+  favDatasets = favDatasetsText ? JSON.parse(favDatasetsText) : [];
   dispatch({
     type: TYPES.FAVORITED_DATASETS,
-    payload: [...fav_datasets],
+    payload: [...favDatasets],
   });
 };
```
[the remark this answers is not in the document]
MVarshini: Updated

dashboard/src/actions/authActions.js
```diff
@@ -1,56 +1,63 @@
...
-    });
-    if (response.status === 200 && Object.keys(response.data).length > 0) {
-      let keepUser = getState().userAuth.keepLoggedIn;
-      let expiryTime = keepUser ? 7 : 0.5;
-      Cookies.set("isLoggedIn", true, { expires: expiryTime, secure:true });
-      Cookies.set("token", response.data?.auth_token, { expires: expiryTime, secure:true });
-      Cookies.set("username", response.data?.username, { expires: expiryTime, secure:true });
-      navigate("/");
+export const makeLoginRequest =
+  (details, navigate) => async (dispatch, getState) => {
```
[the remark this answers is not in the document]
MVarshini: I think this is the standard way to dispatch an action. In case any changes needed, let's dig into it later.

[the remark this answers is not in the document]
dbutenhof: I think this is my biggest source of discomfort with the Javascript convention of binding names to lambdas instead of just declaring functions: it causes patterns like this which are awkward to follow visually.

And, yeah; binding a lambda to another lambda just seems odd.

portante (approved):

dashboard/src/actions/authActions.js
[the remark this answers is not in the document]
webbnh: @dbutenhof, Varshini tossed me a clue in her other PR:  it's a _thunk_.  Apparently, it's a pretty common technique in Javascript used to defer the construction of values:  instead of constructing them in-line, you create a function which provides the value and then you reference the function when you want the value.  The usefulness of this is obvious if the value is produced asynchronously, but it's also useful to avoid producing complex values when they aren't actually used.  So, I understand and appreciate it as a concept.

However, like any tool, I would recommend only using this idiom when it's actually needed or otherwise beneficial.  I would recommend against using it just because it has become a habit.  It is intended to be a tradeoff between making obtaining the value more expensive (by wrapping it inside a function call) against avoiding producing the value initially (while rendering the page) or at all (if it ends up never being needed, because the user never clicked in the right spot).  The problem is, if producing the value is less expensive than the function call required to get it, then we should avoid using a thunk!

As for the distinction between using the `function` keyword and the `=>` thingie (apparently, it is _not_ an "operator"...does that make it a "keyword"??), there doesn't seem to be a huge difference, and so I guess I can get used to what Dave is calling "binding names to lambdas".  But, in a situation like the one here, where what we are doing is defining a named function, I would definitely lean toward using the `function` keyword; I would reserve using `=>` for unnamed functions or similar in-line definitions.

But, if this is all actually working, I won't push back too hard....

dbutenhof (approved):
The commit message looks good now.

webbnh (approved):

PR #2872, opened 2022-05-27
Correct installation path for Server RPM build

riya-17:
`python3 setup.py install --prefix=...` adds `local` to the installation path.

I believe this issue is only in Fedora-36.


dbutenhof:
You even mentioned in the description that this was only on Fedora-36, although your change is global, and COPR builds on other platforms are clearly working. Your tests failed because clearly this change is *not* applicable to our Jenkins environment:

> cp: cannot stat '/var/tmp/jenkins/tox/py/src/server/python3/local/bin/pbench-config': No such file or directory

If this is a fix for *only* Fedora-36, you'll need to figure out a way to make it conditional. (Or figure out a way to make Fedora-36 "behave". 😁 )

portante (changes requested):
We need a Jira issue for this.

webbnh:
We are going to adopt a different approach to this problem.

PR #2875, opened 2022-05-27
Added notion of aria-label to elements of browsing page that are not easily targetable

shubham-html-css-js:
While unit testing, i encountered a problem that few elements, which are either having no particular label or elements who are having non-unique labels are very difficult to target.So to make them easy to target(and therefore, easy to test),I have given all such elements an aria-label attribute, by which we can target those elements easily.

### DIFF
--- a/dashboard/src/modules/components/TableComponent/index.jsx
+++ b/dashboard/src/modules/components/TableComponent/index.jsx
@@ -152,12 +152,14 @@ const TableWithFavorite = () => {
             startDate={startDate}
             endDate={endDate}
             setDatasetName={setDatasetName}
+            aria-label="search box"
           />
           <DatePickerWidget
             dataArray={dataArray}
             setPublicData={setPublicData}
             datasetName={datasetName}
             setDateRange={setDateRange}
+            aria-label="date picker"
           />
         </div>
         <ToggleGroup aria-label="Result Selection Options">
@@ -167,6 +169,7 @@ const TableWithFavorite = () => {
             isSelected={isSelected === "datasetListButton"}
             onChange={handleButtonClick}
             className="datasetListButton"
+            aria-label="see dataset button"
           />
           <ToggleGroupItem
             text={`Favorites(${favoriteRepoNames?.length})`}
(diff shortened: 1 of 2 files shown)


dashboard/src/modules/components/TableComponent/index.jsx
```diff
@@ -167,13 +169,15 @@ const TableWithFavorite = () => {
...
             className="datasetListButton"
+            aria-label="see dataset button"
           />
           <ToggleGroupItem
             text={`Favorites(${favoriteRepoNames?.length})`}
             buttonId="favoriteListButton"
             isSelected={isSelected === "favoriteListButton"}
             onChange={handleButtonClick}
             className="favoriteListButton"
+            aria-label="see favorites button"
```
dbutenhof, line 180: While this is clever, and including the accessibility labels is probably good anyway, I'm wondering if most or all of this targeting couldn't be done with the more common `buttonId` tags??

dbutenhof (approved):

webbnh (approved):
👍

dashboard/src/modules/components/TableComponent/index.jsx
webbnh: > that failed

How did it fail?  Did it find no element, or did it find more than one?  (Or, was there some other objection?)

shubham-html-css-js: `TestingLibraryElementError: Unable to find an accessible element with the role "button" and name "favoriteListButton"`

webbnh: A brief bit of research suggests that `getByRole()` is probably the right choice of function, and specifying `"button"` is appropriate, and that the remaining options target a number of characteristics which do _not_ include the ID.  The targeting seems to have a much more user-focused perspective, so they key off items which are user-accessible, including the ARIA values.  So, I think Shubham's on the right track, here.

portante (changes requested):
We need a Jira ticket for this.

dbutenhof:
> We need a Jira ticket for this.

PBENCH-709 (sub-task under PBENCH-211, dashboard unit test coverage)

Shubham: edit the git log history (git rebase -i) and modify the top git log message to a format like:

TITLE (e.g., "Add aria label to UI elements for testing)
<blank line>
PBENCH-709
<blank line>
Description of the change (keep the lines under 72 columns)

MVarshini (approved):

webbnh (approved):
👍

portante (approved):

PR #2878, opened 2022-06-02
Expose tarball metadata.log values through API

### LINKED ISSUE #2600: Expose `metadata.log` as Python `Metadata`
Both the file tree `Tarball` object and the indexer have code to process the tarball `metadata.log` into a JSON object to use within Python. This needs to be completed and refactored to avoid duplication. Ideally, we'll expose the full contents of `metadata.log` (at least everything that might actually be useful) into one or more `Metadata` JSON documents. Once we've done this during `PUT`, the indexer will be able to leverage the metadata without processing the file again.

dbutenhof:
PBENCH-682

The filetree code was already parsing metadata.log in order to obtain the "created" timestamp. This PR formalizes that a bit, converting the entire config hierarchy (sections/settings) as a JSONOBJECT, which is persisted in SQL as a "metalog" key value for the dataset.

The "metalog" key value is exposed as a key in the "dataset" namespace, so that `/datasets/list?metadata=dataset` will return it as part of the JSON representation of the primary Dataset SQL row. As with any other metadata namespace, segments can be addressed directly, e.g., `dataset.metalog.run.controller`.

NOTE: I'm not deeply in love with the name "metalog"; however I didn't want to use the obvious "metadata" as it'll all appear nested inside the outer "metadata" key. I'm open to alternate suggestions!

Resolves #2600

### DIFF
--- a/lib/pbench/server/filetree.py
+++ b/lib/pbench/server/filetree.py
@@ -237,7 +237,7 @@ class Tarball:
         except Exception as exc:
             raise MetadataError(self.tarball_path, exc)
 
-    def get_metadata(self) -> Optional[JSONOBJECT]:
+    def get_metadata(self) -> JSONOBJECT:
         """
         Fetch the values in metadata.log from the tarball, and return a JSON
         document organizing the metadata by section.
@@ -245,18 +245,13 @@ class Tarball:
         The information is unpacked and processed once, and cached.
 
         Returns:
-            A parsed ConfigParser of `metadata.log`
+            A JSON representation of the dataset `metadata.log`
         """
         if not self.metadata:
-            try:
-                data = self.extract(f"{self.name}/metadata.log")
-                metadata = ConfigParser()
-                metadata.read_string(data)
-                self.metadata = {
-                    s: dict(metadata.items(s)) for s in metadata.sections()
-                }
-            except Exception as e:
-                self.logger.exception("{} metadata extraction: {}", self.name, e)
+            data = self.extract(f"{self.name}/metadata.log")
+            metadata = ConfigParser()
+            metadata.read_string(data)
+            self.metadata = {s: dict(metadata.items(s)) for s in metadata.sections()}
         return self.metadata
 
     def unpack(self, incoming: Path, results: Path):
--- a/lib/pbench/server/filetree.py
+++ b/lib/pbench/server/filetree.py
@@ -3,14 +3,13 @@ from logging import Logger
 from pathlib import Path
 import re
 import shutil
-from typing import Any, Dict, List, Union
+from typing import Dict, List, Optional, Union
 
 import tarfile
 
 from pbench.common import selinux
-from pbench.server import PbenchServerConfig
+from pbench.server import JSONOBJECT, PbenchServerConfig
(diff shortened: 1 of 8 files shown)


npalaska (dismissed):
Nice!

lib/pbench/test/unit/server/conftest.py
```diff
@@ -394,6 +394,32 @@ def provide_metadata(attach_dataset):
...
+                "date": "2020-02-15T00:00:00",
+                "config": "test1",
+                "script": "unit-test",
+                "name": "drb",
+            },
+            "run": {"controller": "node1.example.com"},
+        },
+    )
+    Metadata.create(
+        dataset=test,
```
webbnh, line 411: The various steps of the instantiations of the `drb` and `test` datasets are kind of arbitrarily intertwined -- it would be nicer if they were either separated or performed in alternating sequence (e.g., by key).

webbnh (approved):
Looks good, but I have a quibble and a suggestion.

lib/pbench/server/filetree.py
```diff
@@ -237,30 +237,27 @@ def extract(self, path: str) -> str:
         except Exception as exc:
             raise MetadataError(self.tarball_path, exc)
 
-    def get_metadata(self) -> Dict[str, Any]:
+    def get_metadata(self) -> Optional[JSONOBJECT]:
```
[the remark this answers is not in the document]
dbutenhof: The `try` block in upload will catch a missing key, in case we can't extract metadata. The apparent willingness to set the metadata key to value `None` is misleading since obviously we'll never get that far if we can't excavate metadata.

You're probably right that it's better to let that outer `try` catch an exception during the process of acquiring metadata; I think I'll just remove this `try` block entirely, which does probably make the `Optional`, um, optional...

lib/pbench/test/unit/server/conftest.py
dbutenhof: Basically these new ones were unavoidably multi-line and at the time I decided it was "neater" looking to segregate those at the end. However, yeah; I put in an `index-map` for `drb` (hey, my sample datasets are ahead of their time: freeform "metadata" name instead of canonical `<script>_<config>_<date>`; [PBENCH-744](https://issues.redhat.com/browse/PBENCH-744)). I'll move the `drb` metalog up with it.

lib/pbench/test/unit/server/conftest.py
```diff
@@ -407,6 +405,9 @@ def provide_metadata(attach_dataset):
             "run": {"controller": "node1.example.com"},
         },
     )
+
+    Metadata.setvalue(dataset=test, key="dashboard.contact", value="you@example.com")
```
webbnh, line 409: Should line 383 (the instantiation of `test`) have been moved down to here?

webbnh (dismissed):
Nice.

npalaska (dismissed):

lib/pbench/test/unit/server/conftest.py
dbutenhof: Oh, fine...

webbnh (approved):
👍

portante (approved):

PR #2879, opened 2022-06-03
Add `numactl` RPM requirement

portante:
Both the `pbench-uperf` and `pbench-linpack` commands offer interfaces that rely on `numactl`.

### DIFF
--- a/agent/rpm/pbench-agent.spec.j2
+++ b/agent/rpm/pbench-agent.spec.j2
@@ -46,7 +46,7 @@ Requires:  python3-psutil
 Requires:  perl, perl-Data-UUID, perl-JSON, perl-JSON-XS
 Requires:  perl-Time-HiRes
 
-Requires:  bc, bzip2, hostname, iproute, iputils, net-tools
+Requires:  bc, bzip2, hostname, iproute, iputils, net-tools, numactl
 Requires:  openssh-clients, openssh-server, procps-ng, psmisc, redis
 Requires:  rpmdevtools, rsync, screen, sos, tar, xz
 


agent/Makefile
```diff
@@ -254,7 +254,7 @@ install-python3-setup: install-util-scripts install-lib
 	${COPY} requirements.txt ${DESTDIR}
 	mkdir -p ${DESTDIR}/python3
 	(cd ..; SKIP_GENERATE_AUTHORS=1 SKIP_WRITE_GIT_CHANGELOG=1 python3 setup.py install --prefix=${DESTDIR}/python3)
-	${COPY} $(addprefix ${DESTDIR}/python3/bin/, ${click-scripts}) ${UTILDIR}/
+	${COPY} $(addprefix ${DESTDIR}/python3/local/bin/, ${click-scripts}) ${UTILDIR}/
```
dbutenhof, line 257: If I'm not mistaken, this was exactly what Riya tried to do for Fedora-36, but it doesn't work elsewhere, including on our Jenkins VMs:

```
cp: cannot stat '/var/tmp/jenkins/tox/py/src/agent/python3/local/bin/pbench-generate-token': No such file or directory
cp: cannot stat '/var/tmp/jenkins/tox/py/src/agent/python3/local/bin/pbench-cleanup': No such file or directory
```

portante: Yeah, this is why I posted this in a separate commit.  I needed to ensure an RPM built before submitting this.

portante:
@ndokos asked elsewhere: "Why not modify `pbench-uperf` and `pbench-linpack` to complain and abort the run as with tools?"

The answer is that those benchmark convenience scripts are not tools.  They are part of the interface provided by those scripts, and all the dependencies for the interfaces provided by the pbench-agent should be direct RPM dependencies.

If we take the logic that we are applying to tools, that is that the dependencies for tools are not applied to the pbench-agent RPM, then the pbench-agent RPM would have _NO_ dependencies at all.

Further, `numactl` does not fall in the category of a "tool", since it does not gather metrics or other data about a program or the system.  Rather, it offers controls over the system.

Its sibling, the `numastat` command, is a tool, though it is provided by the `numactl` RPM.

portante:
FYI, the `numactl` RPM was removed in PR #1887 (commit 73ccffc0534edffbf5c8b916244d7befbbc378f6).

portante (commented):
Remove the Fedora 36 changes from this PR.

ndokos (approved):

ndokos:
But one could argue that  the `pbench-uperf` interface also depends on `uperf`, but we don't make `pbench-agent` depend on `uperf`.

npalaska (approved):

dbutenhof (approved):

dbutenhof:
BTW: While using the branch name to link to Jira clearly works, it also means the final `main` git history won't include the Jira number, while putting it in the message will. I like having it in the history for future reference. (Granted, we can always edit the log on merge; but if we put it there in the first place, we don't have to.)

PR #2882, opened 2022-06-04
Correct CI / unit tests to work with python 3.10

portante:
In order to get our the environment to work with Python 3.10 and later, we need to changes:

 1. Correct CI / unit tests to work with python 3.10

    This also corrects the version of `black` used in the commit hook to work with Python 3.10.

    The version of pytest needs to be at least 6.2.5 in order to work with Python 3.10.  As a result of this change, we need to correct one test with its use of `ExceptionInfo` objects.

    Unfortunately, the "usage" message created by the `argparse` module changed to use "options:" instead of "optional arguments:". We ensure that only for the two legacy tests that use a python command the output emitted by older versions of Python3 look like Python 3.10 and later.

 2. Stop using deprecated setuptools method of install

    The new way to install our Python3 interfaces is:

        /usr/bin/python3 -m pip install --prefix=${dir} -e .

    See https://docs.python.org/3.6/installing/index.html#installing-index.

    The new interface allows for the `agent/Makefile` and `server/Makefile` to create the python CLI interfaces on environments running Python 3.10 and later.

    Why do we use `/usr/bin/python3`?  If you invoke pip inside a virtualenv to install a package, it automatically uninstalls the previous one from where it was installed, and installs it again in the `--prefix` path. That is wrong on so many levels. So, to work around this terrible behavior we have to invoke the non-virtualenv `/usr/bin/python3` so that the installation works without touching the virtualenv.


portante (commented):
Oy -- getting this to work was a pain.  If you invoke `pip` inside a virtualenv to install a package, it automatically uninstalls the previous one from where it was installed, and installs it again in the `--prefix`.  That is wrong on so many levels.  So, to work around this we have to invoke the non-virtualenv `/usr/bin/python3` so that the installation works without touching the virtualenv.

agent/Makefile
```diff
@@ -250,10 +250,13 @@ install-lib:
 	${INSTALL} ${INSTALLOPTS} ${LIBDIR}
 	${COPY} lib ${DESTDIR}
 
+# NOTE: we deliberately use `/usr/bin/python3` to make sure that when this is
+# invoked inside a python virtualenv the pip installation won't uninstall the
+# existing pbench module, breaking the environment.
```
dbutenhof, line 255: My head hurts in sympathy for your pain ... 😵‍💫 

But more practically ... is this going to be an issue if we try to expand our `tox` to work with multiple Python environments, since we're always forcing the install through the "default" Python?

agent/util-scripts/unittests
```diff
@@ -615,6 +615,8 @@ declare -A post_hooks=(
     [test-07]='cat ${_testdir}/42-iter/sample42/tools-foobar/testhost.example.com/postprocess.log >> ${_testout} 2>&1'
     [test-19]='rm ${_testopt}/unittest-scripts/pbench-tool-meister-client'
     [test-53]='sort_testlog; sort_tmlogs; filter_tmerrs'
+    [test-54]='sed -i "s/optional arguments:/options:/" ${_testout}'
+    [test-55]='sed -i "s/optional arguments:/options:/" ${_testout}'
```
dbutenhof, line 619: It might be (a bit) safer to anchor the search string with "^", though an incidental match is unlikely.

lib/pbench/test/unit/agent/task/test_copy_result_tb.py
```diff
@@ -48,13 +48,13 @@ def test_bad_tar(self, caplog):
...
+        expected_error_message = f"Tar ball '{bad_tarball}' does not exist"
         caplog.set_level(logging.ERROR, logger=self.logger.name)
-        with pytest.raises(FileNotFoundError) as e:
+        with pytest.raises(FileNotFoundError) as excinfo:
             crt = CopyResultTb(
                 "controller", bad_tarball, 0, "ignoremd5", self.config, self.logger
             )
             crt.copy_result_tb("token")
-        assert str(e).endswith(expected_error_message)
+        assert str(excinfo.value).endswith(
```
dbutenhof, line 58: Is this because the message actually changed? 😕

dbutenhof (approved):
Comments and questions, but basically looks good.

agent/Makefile
portante: I don't know.  I suspect we need to run that install by either having it detect that it is in a CI job and using the artifacts of the virtualenv of the CI job, and otherwise using the available `python3` interpreter out side our CI environment.

lib/pbench/test/unit/agent/task/test_copy_result_tb.py
portante: The way `pytest.raises` works in the later versions of `pytest` (6.2.5 and later) that is compatible with Python 3.10.

dbutenhof: Actually, I guess I highlighted the wrong line; I understand the `ExcInfo` change, but I meant to call attention to dropping the `FileNotFoundError: ` prefix on the message. I did that badly, didn't I? 😆

agent/Makefile
dbutenhof: In other words, we don't know, but this is a step forward. Got it. We really ought to test in multiple environments unless we can actually standardize support on a single environment (which I doubt will be practical any time soon); but boiling the ocean to make a cup of tea is probably a bad idea so "carry on".

dbutenhof (approved):

lib/pbench/test/unit/agent/task/test_copy_result_tb.py
dbutenhof: Wait; you're saying that under Python 3.10 `str(FileNotFoundError)` doesn't include the "FileNotFoundError: " prefix anymore? Wow.

dbutenhof (approved):

lib/pbench/test/unit/agent/task/test_copy_result_tb.py
portante: No, I think worse, the new version of Pytest seems to strip it, because this works under Python 3.6 thru 3.10.

agent/Makefile
webbnh, line 255: Is it ever the case that, inside a virtual environment, the version of Python matches `/usr/bin/python3`, such that referencing `/usr/bin/python3` below would _not_ avoid this effect?

webbnh (approved):
Looks good, just one "small" question.

portante:
See also #2882.

portante:
Sorry mistakenly closed the PR adding a comment.

webbnh:
> See also #2882.

Every time I click on this link I find another "See also" reference....  😉

PR #2883, opened 2022-06-07
Remove the unused `check_install_rpm` function

portante:
We also remove in a separate commit the contributed `pbench-bzt` and `pbench-mpt` scripts.

The first script, `pbench-bzt` is not really a Pbench Agent benchmark workload convenience script, it is a wrapper around the "bzt" workload.  So it really doesn't belong here.

The second script, `pbench-mpt` is woefully out of date and unmaintained.


webbnh (approved):
👍

portante (commented):
Just rebased to double check no conflicts.

webbnh (approved):
👍

dbutenhof (approved):

PR #2890, opened 2022-06-10
Switch to tarball MD5 as Dataset resource_id

dbutenhof:
PBENCH-744

Change the Dataset table to make resource_id the primary key. This must be unique due to Elasticsearch constraints, and the change allows us to "demote" dataset name (which is just script+config+date) to metdata that we can allow the user to set arbitrarily.

This change ripples through the code, resulting in a lot of diffs. Most of these are fairly straightforward.

Note that this removes legacy test-5.2; PR #2611 also removes this as it finally drops pbench-server-prep-shim-002 entirely, but the fairly twisted things test-5.2 does with duplicate and MD5 errors simply don't apply to the 1.0 server and it's not worth trying to fix it.


server/bin/pbench-state-manager.py
```diff
@@ -49,15 +50,27 @@ def main(options):
...
-        elif options.name:
+            if not options.md5:
+                args["resource_id"] = md5sum(options.path).md5_hash
+            if options.create and not options.name:
+                args["name"] = Dataset.stem(options.path)
+        elif options.name and options.create:
             args["name"] = options.name
         if options.md5:
-            args["md5"] = options.md5
+            args["resource_id"] = options.md5
```
webbnh, line 72: Alternately,
```
        if options.md5:
            args["resource_id"] = options.md5
        elif options.path:
            args["resource_id"] = md5sum(options.path).md5_hash

        if options.create:
            if options.name:
                args["name"] = options.name
            elif options.path:
                args["name"] = Dataset.stem(options.path)
            else:
                I_think_we_need_to_fail_here()
```
I think this approach is simpler, and it addresses the case when `create`, `path`, and `name` are all true-y, which I think your code misses.  (I don't know whether we need/want to handle the failure at the last line, but it seems like a good idea and that seems like a good place to do it.)

server/bin/pbench-state-manager.py
```diff
@@ -69,21 +82,21 @@ def main(options):
...
         # Stringify any exception and report it; then fail
-        logger.exception("Failed")
+        logger.exception(
+            "Failed to {} {}", "create" if options.create else "attach", args
+        )
+        try:
+            dup = Dataset.query(resource_id=args["resource_id"])
+            logger.error("Found dup for {} : {}", args, dup.as_dict())
+        except Exception as d:
+            logger.error("Checking {}: {}", args["resource_id"], str(d))
```
webbnh, line 99: Do you want to ignore the dataset-not-found case, as that probably isn't an interesting error to report, here?  (That is, we should report a duplicate, and we should report access errors, but if the only problem is that that the dataset that we were trying to create is not here...well...um...duh?)

server/bin/utils/run-unittest
```diff
@@ -463,7 +463,13 @@ function _setup_state {
         local state
         local user
         while read -r controller name state user; do
-            pbench-state-manager --create=${user} --name=${name} --state=${state}
+            tarball=$(find ${_testdir} -type f -name ${name}.tar.xz)
```
webbnh, line 466: Using `find()` here seems like sort of a big hammer.  If we cannot accurately predict the location of the file, perhaps a straightforward glob expression would do?  Something like:
```
            tarball=${_testdir}/*/*/${name}.tar.xz)
```
or whatever is appropriate.

(If you decide to keep the `find()` command, consider reversing the order of the criteria -- I suspect that type-file will be true more often than a name match.)

server/bin/utils/run-unittest
```diff
@@ -463,7 +463,13 @@ function _setup_state {
...
+            pbench-state-manager --create=${user} --path=${tarball} --state=${state}
+            if [[ ${?} -gt 0 ]]; then
+                echo "ERROR: unable to create ${tarball} dataset" >&2
+                find ${_testdir} -type f -name ${name}.tar.xz >&2
```
webbnh, line 470: Repeating the `find` command here seems...unreliable.  Any diagnostics from the `find` command at line 466 have already been sent to `stderr`, and the regular output of that command was captured in `tarball` and reported at line 469; so, at best, the output of line 470 will be redundant, and, at worst, it'll be _different_ from the previous invocation (due to external changes to the environment) which is liable to sow confusion.

(Hopefully, we can replace the `find` command at line 466, in which case there will be no cause to invoke it here....)

webbnh (approved):
Looks good, other than a few typos, but I have a few pointed questions for you (including a possible bug) and a couple of other things for you to consider.

lib/pbench/server/api/resources/datasets_list.py
```diff
@@ -154,7 +154,7 @@ def _get(self, params: ApiParams, request: Request) -> Response:
         for dataset in datasets:
             d = {
                 "name": dataset.name,
-                "run_id": dataset.md5,
+                "resource_id": dataset.resource_id,
```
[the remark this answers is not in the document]
dbutenhof: Resources have an informational name as well as the formal resource ID; nobody wants to display the resource ID, and the name (whatever it's been set to) provides a more friendly display.

The default `dataset.name` will be `dataset.metalog.pbench.script` + `dataset.metalog.pbench.config` + `dataset.metalog.pbench.date` just as it is now, but I plan to make a change to `PUT /datasets/metadata` to allow changing it now that it's not essential internally. ([PBENCH-749](https://issues.redhat.com/browse/PBENCH-749))

So, yes, the default response is the resource ID and the resource name.

lib/pbench/server/database/models/datasets.py
```diff
@@ -413,7 +413,7 @@ class Dataset(Database.Base):
     id = Column(Integer, primary_key=True, autoincrement=True)
 
     # Dataset name
-    name = Column(String(255), unique=True, nullable=False)
+    name = Column(String(255), unique=False, nullable=False)
```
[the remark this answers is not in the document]
dbutenhof: I could in theory move the resource name somewhere else; but I like having it, I think it's useful, and I don't see any point to move it out of the primary resource object.

Could we allow someone to set the resource name to ""? Sure, but I'd rather not.

lib/pbench/server/utils.py
```diff
@@ -60,6 +63,25 @@ def filesize_bytes(size):
...
+    result in an MD5 file. This fallback covers several legacy testing cases
+    that are otherwise problematic.
+
+    Args:
+        tarball: Path or string filepath to a tarball file
+    """
+    md5_file = Path(f"{str(tarball)}.md5")
+    if md5_file.is_file():
+        return md5_file.read_text().split()[0]
+    return md5sum(tarball).md5_hash
```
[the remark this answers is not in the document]
dbutenhof: Yeah, I almost "fixed" `md5sum` to correctly document a path-like object and decided I didn't really need to change that code at this time.

lib/pbench/test/unit/server/database/test_datasets_metadata_db.py
```diff
@@ -174,51 +175,55 @@ def test_server_keys(self, provide_metadata):
...
+        assert Metadata.get(key="user", dataset=ds, user=user2).value is False
 
-        Metadata.remove(key="user", dataset=ds, user=create_drb_user)
-        assert create_drb_user.dataset_metadata == []
-        assert create_user.dataset_metadata == [t]
-        assert ds.metadatas == [t, g]
+        Metadata.remove(key="user", dataset=ds, user=user1)
+        assert user1.dataset_metadata == []
+        assert user2.dataset_metadata == [d]
+        assert ds.metadatas == [d, g]
```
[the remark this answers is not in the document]
dbutenhof: Making the change to use the standard fixture datasets seemed like a good idea because I didn't have to add resource IDs in a lot of places; but it also led to some messy unraveling as the references changed here. Perhaps I could have done it differently. Perhaps that would have removed a few diffs. I didn't think it was worth any additional effort. I still don't.

server/bin/utils/run-unittest
dbutenhof: I didn't want to do any more massive legacy test reconstruction than essential. I wish I hadn't had to do this much. Long ago I'd listed the dataset names, and created the `.db` mechanism to pre-load `Dataset` objects... but that didn't include the path. I don't want to go back and analyze all of those test cases to expand the `.db` file with a path; I really don't. The `find` was an expedient fix that could be done once in the script, and I like that.

server/bin/utils/run-unittest
dbutenhof: I'm not concerned about diagnostics; I was trying to figure out why my original attempt didn't work even though I knew a file was there. It was a fairly special case of failure, though, and the second `find` definitely outlived its usefulness. Ah, well.

server/bin/pbench-state-manager.py
dbutenhof: This was more debugging code, which I don't think really serves a purpose. I just forgot to forget it.

npalaska (approved):

lib/pbench/server/api/resources/datasets_list.py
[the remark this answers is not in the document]
webbnh: This might be out of scope for this PR, but what are we thinking the Dashboard is going to do with this?  

My previous understanding was that we were expecting the Dashboard to display the results list with the script, config, and date in three separate columns, with the user able to sort on any of the columns.  If the dataset name is still a thing, then we have four columns, and the others are kind of redundant with the name (up until the user changes the name).

Having the default display be the name, and having the default name be `script`+`config`+`date` seems like we haven't improved anything, here.  On the other hand, having the Dashboard display `script`,`config`, and `date` (and _not_ `name`) seems like a nice improvement...right up until we allow the user to change the name, at which point we want to be able to display the name...but then that screws up what we display by default (I think).

So, I'm questioning whether we really want a name field at all.  We know what we would put in it by default if we had it, but that just makes it redundant with other stuff that we want to do.  We do want to enable users to do certain things, and one way to do that is to allow them to set the name, but I feel like there are other ways to enable that support which don't involve a _name_ field, per se.  I'm proposing that we provide "labels" or "annotations" and that we get rid of the `name` field altogether.  I believe that we've already got ways to display our equivalent of the "name" (via the three components) and that, if the user wants to create a custom handle for the result, they should do it via a label or annotation, and that delivers us from the question of whether/when to display the name versus the `script`/`config`/`date` trio.

server/bin/pbench-state-manager.py
```diff
@@ -61,15 +61,23 @@ def main(options):
...
         if options.md5:
             args["resource_id"] = options.md5
+        elif options.path:
+            args["resource_id"] = md5sum(options.path).md5_hash
+
+        if options.create:
+            if options.name:
+                args["name"] = options.name
+            else:
+                args["name"] = Dataset.stem(options.path)
```
webbnh, line 80: Do you want to combine the bodies of the `if` statements at lines 61 and 76, since they have the same conditional?  (That would then allow you to combine the `if` statement at line 64 with the `else` at line 79 to produce a simpler `elif` test.)

server/bin/utils/run-unittest
webbnh: OK.  (I think the glob expression is likely to be more efficient than invoking `find`, but I don't know what constraints this code is facing.)

Given that we're sticking with `find`, do you want to check the `-name` first and then the file `-type`?  I'm figuring that the name will rarely match, which will save a bunch of `stat()` calls trying to decide whether the type matches.

lib/pbench/server/utils.py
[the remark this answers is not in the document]
webbnh: Well, if we're not going to "fix" `md5sum()`, then it seems like we should be passing `Path(tarball)` to it here, no?

webbnh (approved):
Looks good, but I still have a few items for you to consider.

npalaska (approved):

lib/pbench/server/api/resources/datasets_list.py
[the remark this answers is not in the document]
webbnh: I guess what is stuck in my craw about this is that the default behavior doesn't land the user anywhere near where we want to be.  That is, we think that having a name for run results would be The Thing, but in order to get there, the user has to explicitly specify an option (that we don't support, yet).  Instead, the typical case is going to be the run results get the default name (with the `script`/`config`/`date` trio) which is the opposite of what we want (we want those three to be in separate columns).  And, yeah, the user can reconfigure his view to omit the `name` column and present the other three, but, since we think the run name is The Thing, that's probably not the view that we're going to offer by default.

Since the user has to take an explicit step to use the `--name` option, it seems like the user should have to take an explicit step to get the "name" column in the UI.  Then the default-taking user would have the `script`/`config`/`date` trio display by default, which I think would result in a much better experience.  There's nothing which prevents a user from getting The Thing, but s/he has to take explicit steps to get there; otherwise, the user gets the trio by default.

lib/pbench/server/utils.py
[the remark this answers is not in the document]
webbnh: Well, if we're going to rely on being able to pass a string to `md5sum`, then I think we ought to fix its signature...but, I guess that's beyond the scope of this PR.

webbnh (approved):
👍

portante (commented):
So I was not quick enough to review this PR.

What is the plan for replacing what was tested in `test-5.2`?  The still exists `test-5.1` which runs everything but without any data.  We need something that runs everything still, no?

Further, I think we need to document our nomenclature regarding the data being up-loaded by the clients.

What is a run?

What is a result?

What is an iteration?

What is a sample?

That nomenclature is going to be important with regard to the UI and the Pbench Agent CLI interfaces.

With this change, `resource_id` is more of a generic server-side term.  We need to document clearly how the UI and CLI nomenclature maps to the internal server use.

I had hoped not to have to keep to separate nomenclatures, but maybe that is worth maintaining?

webbnh:
> I was not quick enough to review this PR.

My apologies, Peter!  😞   (I _hate_ it when that happens to me!!)

dbutenhof:
> So I was not quick enough to review this PR.
> 
> What is the plan for replacing what was tested in `test-5.2`? The still exists `test-5.1` which runs everything but without any data. We need something that runs everything still, no?

Can you please enumerate what you think was dropped? There may well be subtleties I missed that remain relevant.

Basically, what test-5.2 seemed to be testing (beyond what other tests already cover) as `pbench-prep-shim-002` boundary cases, especially with quarantine. None of those cases are relevant for the new server, because `PUT` will reject duplicates, bad MD5 values, etc, and the concept of a "missing" MD5 file is meaningless because there is none until we create it from the MD5 HTTP header (and if that header is missing or incorrect, again, we'd have rejected the `PUT`). In any of these cases, the tarball remains in control of the agent, and we don't need to worry about it.

Additionally, I believe it tested some aspect of the `--prefix` handling; I'm not sure if there are other tests that use `--prefix`, but in the server 1.0 schema we only care about validating that the metadata is picked up correctly; even if we lose coverage in the legacy tests (until we have our functional test system, which hopefully will be soon), I'm not that worried about this aspect.

Is there more?

> Further, I think we need to document our nomenclature regarding the data being up-loaded by the clients.
> 
> What is a run?
> 
> What is a result?
> 
> What is an iteration?
> 
> What is a sample?
> 
> That nomenclature is going to be important with regard to the UI and the Pbench Agent CLI interfaces.

Which is all great, but I don't see how any of that has anything to do with test-5.2 or even this PR. How does this fit?

> With this change, `resource_id` is more of a generic server-side term. We need to document clearly how the UI and CLI nomenclature maps to the internal server use.
> 
> I had hoped not to have to keep to separate nomenclatures, but maybe that is worth maintaining?

I suppose we'll have to work this out at some point by mapping CLI/options through to directory/tarball artifacts and then to server data and APIs.

In this specific case, `resource_id` is a REST API concept that ensures a unique key for each object. In terms of the data provided, the behavior of the APIs, and how the data is used, it's more or less incidental (and probably should be opaque) that we're using the tarball's MD5 hash. All that really matters in the context of this PR and the use in our APIs is that it's a unique handle on this particular object. The client doesn't directly expose the concept of the MD5, and it's not relevant to the benchmark or metric data collected, so I don't see how this presents any problem. What am I missing?

PR #2891, opened 2022-06-10
Prototype API documentation

dbutenhof:
PBENCH-754

Begin building Pbench server API documentation in GitHub Markdown, beginning with some overview material and a few detailed (maybe not complete) V1 API examples.

We can discuss and refine the limited examples in this PR to get the ball rolling, and continue to build out APIs in subsequent PRs.

(Helpful hint: selecting "View file" from the top right "..." pulldown on GitHub's diff view will show the *formatted* documentation page.)


docs/API/V1/inventory.md
```diff
@@ -0,0 +1,39 @@
...
+The client did not provide an authentication token and there is no public dataset with the name `<dataset>`.
+
+`403`   **FORBIDDEN** \
+The named `<dataset>` is not public and the authenticated user lacks authorization to read it.
+
+`404`   **NOT FOUND** \
+Either the `<dataset>` or the relative `<path>` within the dataset does not exist
+
+`415` **UNSUPPORTED MEDIA TYPE** \
+The `<path>` refers to a directory. Use `/api/v1/dataset/contents/<dataset><path>` to acquire a JSON response document describing the directory contents.
```
webbnh, line 35: s/acquire/request/

docs/API/V1/inventory.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Response body
+
+The raw byte stream of the regular file will be returned.
```
webbnh, line 39: Perhaps, "The contents of the file will be returned as a raw byte stream"?

We should mention that it's the contents of the file, and we should omit "regular" since we'll also handle symlinks, right?  (If we want to discuss what happens if the target is a symlink, this would be the spot to do it, I think...but I would just remain silent on the topic.)

docs/API/V1/list.md
```diff
@@ -0,0 +1,125 @@
+# `GET /api/v1/datasets/list`
+
+This API returns an `application/json` document describing the set of datasets accessible to the authenticated client. (Or, for an unauthenticated client, all datasets with access "public".)
```
webbnh, line 3: I would omit the word "authenticated" here -- the statement would remain true and accurate without it, and the parenthetical provides the needed "color" (although I would drop the "Or", as well).

You already talk more about authentication under `Request headers`.

webbnh (commented):
This looks great so far!  I didn't quite finish, but I'm posting what I've got.

I have some concerns about the handling of symlinks.  Other that that, my comments are mostly typographical and similar.

> Helpful hint: selecting "View file" from the top right "..." pulldown on GitHub's diff view will show the formatted documentation page.

Alternatively, on the "Files Changed" page, you can select "rich diff" (via the "document" icon just to the left of the "Viewed" tickbox), which will let you view the formatted text in the page.

docs/API/V1/README.md
```diff
@@ -0,0 +1,86 @@
+# Pbench server V1 API documentation
+
+The documents in this set describe the APIs supported by the Pbench 1.0 server.
+
+## Discovering the Pbench server API
+
+Once you know the hostname of a Pbench server, you can ask for the API configuration using the [endpoints](endpoints.md) API. This will report the server's version and a list of all API end points (URI roots) supported by the server.
```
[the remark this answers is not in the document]
dbutenhof: Yes, I tried to build a framework for an API overview, without writing every piece of API documentation. So what are you advocating? Don't even mention APIs not yet documented in the overview? Remove the links (which will just need to be restored later)? Or not bother publishing anything to the repo until it's done?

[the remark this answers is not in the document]
dbutenhof: The "root" was an oblique reference to the fact that the `endpoints` output omits the URI parameters. Note taken that the term is not very clear. 😆

docs/API/V1/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+While logged in, you can retrieve (`GET`) and modify (`PUT`) your user profile through the [user](user.md) API.
+
+## Dataset metadata
+
+You can read a more complete specification of Pbench server metadata at [metadata](../metadata.md).
+
+When a dataset is first processed, the Pbench server will populate basic metadata, including the creation timestamp, the owner of the dataset (the user authorization token used to authenticate the `PUT` from the agent via `pbench-results-move`), and the full contents of the dataset's `metadata.log` file inside the dataset tarball. The Pbench server will also calculate a default deletion date for the dataset based on the owner's retention policy and the server administrator's retention policy.
+
+Clients can also set arbitrary metadata through the "dashboard" and "user" metadata namespaces. The "dashboard" namespace can only be modified by the owner of the dataset, and is visible to anyone with read access to the dataset. The "user" namespace is private to each authenticated user, and even if you don't own a dataset you can set your own private "user" metadata to help you categorize that dataset and to find it again.
```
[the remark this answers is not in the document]
dbutenhof: In this case I wanted to set them off but I didn't think the fixed case "code sample" sense was really useful/meaningful. But, yeah, perhaps I should just do that consistently.

docs/API/V1/README.md
```diff
@@ -0,0 +1,86 @@
...
+        print(f"{'  '*level}{name}")
+        for d in ls.directories:
+            directory(request, level + 1, d.name, d.url)
+        for f in ls.files:
+            print(f"{'  '*(level+1)}{f.name})
+            bytes = request.get(f.url)
+            # display byte stream:
+            # inline on terminal doesn't really make sense
+
+    directory(request, 0, "/", "http://host.example.com/api/v1/contents/<dataset>/")
```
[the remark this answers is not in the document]
dbutenhof: This was a pseudo-code sketch, not an attempt to build a full example. I suppose I shouldn't have included the default at all; and having made that fatal error, I guess I have no excuse not to listen to your critique. 😆

docs/API/V1/README.md
```diff
@@ -0,0 +1,86 @@
...
+>
+> One potential solution would be to put a symlink to a regular file in the `files` list and a symlink to a directory in the `directories` list. Note that the symlink "file object" includes the `"linkpath"` of a symlink; however the API will have to figure out whether that's a directory or a regular file, either with another Elasticsearch query or via a `FileTree` check.
+>
+>__NOTE__: that the most common symlink in current archives seems to be a `reference-result` link to the `sample1` directory in the same parent directory, which would be easy to check.
+>
+>__ALTERNATIVE__: symlinks to directories would duplicate a directory tree in a straightforward contents traversal; e.g., both `reference-result` and `sample1` trees, which are identical. Another option is to treat a symlink specially, and not produce a URL; the target will either appear elsewhere in the tree and be handled normally, or it's a broken symlink that can't be followed anyway.
```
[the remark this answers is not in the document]
dbutenhof: The "file object" currently defined in Elasticsearch, and preserved in the `contents` API, does identify a file as a symlink. The big question would be how we manage moving directory symlinks into the `directories` list, which is currently just a list of strings. I'm already proposing to expand that to an object at least containing the string and URL; however to make these into a full "file object" like the `files` list we'd need to come up with the `mtime`/`size`/`mode` data from a filesystem since that information's not indexed. Or maybe we ignore those and just go with `name`, `url`, and `type` which would be (for a "directory") either `sym` or `dir`.

docs/API/V1/contents.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Response headers
+
+`content-type: application/json` \
+The return is a serialized JSON object with information about the named file.
+
+## Response status
+
+`401`   **UNAUTHORIZED** \
+The client did not provide an authentication token and there is no public dataset with the name `<dataset>`.
```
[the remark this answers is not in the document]
dbutenhof: This means there *is* no authentication; disambiguating the expiration case depends on the client knowing whether it supplied authentication. If not, then it's an unauthenticated client looking for data that requires authorization; if it did, then it's reasonably safe to infer that the token has expired.

Even the API implementation doesn't really know, because the Flask auth decorators have detected the expiration and invalidated/removed the token before we get to our API code.

I suppose I could add that to the documentation, though; it's a useful detail, but also potentially a bit confounding. 😆

docs/API/V1/inventory.md
dbutenhof: Technically, the proposal is to effectively hide the symlinks by reporting them as if they were files/directories (albeit with with `"type": "sym"`), and if we have a symlink in the `files` list, and you use that URL, you're getting the contents of the "regular file" to which that symlink points, *not* the contents of the symlink file.

I kinda hoped that "regular file" would bridge that gap in one concise reference rather than writing out a paragraph. Perhaps not. 😦

docs/API/V1/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+The primary dataset `resource_id` is the computed MD5 of the dataset tarball. This is generated by the agent, and checked on the server side to ensure data consistency.
+
+## Discovering accessible datasets
+
+The datasets accessible by a client are limited by the dataset access controls and the client's authorization. Any client can read all "public" datasets, but only the owning user can access "private" datasets.
+
+You can determine the datasets you're allowed to view using the [list](list.md) API. This will return the resource name, the formal resource ID, and selected metadata for each dataset the authenticated user is allowed to read. You can filter the datasets you want listed by date range, owning user, or access policy.
+
+It may sometimes be convenient to know the date range of those datasets, without retrieving the entire list and processing the `dataset.created` timestamp for each. You can do this by calling the [daterange](daterange.md) API; for example, to establish a date picker control.
```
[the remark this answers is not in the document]
webbnh: I guess I kind of keyed on the "It may sometimes be convenient to know the YYY, without retrieving the entire XXX", and that sounded very much like the description of a `HEAD` operation, to me.  If you really want it to be a `GET` without a body, then that's fine -- just add the pertinent headers to _both_ responses.  🙃

Or, we can just provide the extra API, but it seemed like there might be a savings here.

docs/API/V1/README.md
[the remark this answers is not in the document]
webbnh: I'm not proposing lying about symlinks in the lists (or even about omitting them from the lists).  I just think we should minimize the impact of them (in terms of ease of use as well as documentation).  That is, the URL that we return for them should access their referent, the directory or the file, so that if the client hits it, it will act very much like hitting a symlink in a filesystem.  Nevertheless, I don't think we should hide them completely -- I think the client should, if it _really_ cares, be able to tell whether a given entry is a symlink or not.  As I said elsewhere, I doubt that this will matter much in the case of files (although, we would have to do something for broken links), but it could be critical for directories, to allow clients to avoid infinite file walks due to symlink-produced loops.

docs/API/V1/inventory.md
webbnh: "Regular" file typically refers to not-directory, not-block-special, not-char-special, not-FIFO, not-device, not-symlink, etc.  😛  So, yeah, if someone applies this API to a symlink, by that rubric we won't return the contents of the symlink...we'll presumably return the contents of its referent.  But, since we care, it would probably be worth a short paragraph pointing that out.

Btw, when we generate the URL, we're going to need to handle symlinks to symlinks....  😝

And, we need to decide what a broken symlink will return (I _think_ I prefer a `404`, but someone might wish for the contents of the symlink file itself...perhaps we can return that in the body or message).

webbnh (changes requested):
Here's my next installment.  (I think I've completed the first pass; now I have to look at your revision...tomorrow?)  I found one significant issue.

docs/API/V1/README.md
```diff
@@ -20,9 +20,9 @@ While logged in, you can retrieve (`GET`) and modify (`PUT`) your user profile t
 
 You can read a more complete specification of Pbench server metadata at [metadata](../metadata.md).
 
-When a dataset is first processed, the Pbench server will populate basic metadata, including the creation timestamp, the owner of the dataset (the user authorization token used to authenticate the `PUT` from the agent via `pbench-results-move`), and the full contents of the dataset's `metadata.log` file inside the dataset tarball. The Pbench server will also calculate a default deletion date for the dataset based on the owner's retention policy and the server administrator's retention policy.
+When a dataset is first processed, the Pbench server will populate basic metadata, including the creation timestamp, the owner of the dataset (the user associated with the authorization token given to the agent `pbench-results-move` command), and the full contents of the dataset's `metadata.log` file inside the dataset tarball. The Pbench server will also calculate a default deletion date for the dataset based on the owner's retention policy and the server administrator's retention policy.
```
webbnh, line 23: s/the Pbench server/the Pbench Server/
s/the agent/the Pbench Agent/

Also at line 27.

docs/API/V1/contents.md
```diff
@@ -1,14 +1,14 @@
...
+This API returns an `application/json` document describing a `<path>` within the `<dataset>` tarball representation.
 
 ## URI parameters
 
 `<dataset>` string \
 The resource ID of a Pbench dataset on the server.
 
 `<path>`    string \
-The relative resource path of an item in the dataset inventory, as captured by the Pbench agent packaging; for example, `/` for the root, or `/1-default/` for the default first iteration directory.
+The path of an item in the dataset inventory, as captured by the Pbench agent packaging; for example, `/` for the root, or `/1-default/` for the default first iteration directory.
```
webbnh, line 11: s/the agent/the Pbench Agent/

docs/API/V1/README.md
```diff
@@ -40,28 +40,28 @@ To get the details of the dataset's run configuration, in addition to specified
...
 
 The [contents](contents.md) API exposes the directory hierarchy within the tarball, from the root directory `/` through the leaf files.
 
 The [inventory](inventory.md) API returns the full byte stream of any regular file within the directory hierarchy, including log files, postprocessed JSON files, and benchmark result text files.
 
-Starting from the root directory of the dataset tarball, the "contents" API allows the client to discover the entire file system structure, including all directories, regular files, and symbolic links.
+Starting from the root directory of the dataset tarball, the [contents](contents.md) API allows the client to discover all directories and files within the dataset directory structure.
 
-The "inventory" API allows the client to retrieve any regular file for further analysis.
+The [inventory](inventory.md) API allows the client to retrieve the contents any regular file for further analysis.
```
webbnh, line 59: s/contents any/contents of any/

docs/API/V1/contents.md
```diff
@@ -1,14 +1,14 @@
 # `GET /api/v1/datasets/contents/<dataset><path>`
 
-This API returns an `application/json` document describing the contents of the referenced relative `<path>` within the `<dataset>` tarball representation.
+This API returns an `application/json` document describing a `<path>` within the `<dataset>` tarball representation.
```
webbnh, line 3: I'm not quite sure what it means to "describe a path"...perhaps this API describes an "item" found at the location indicated by `<path>`?  This comment applies to line 36, as well.

docs/API/V1/contents.md
```diff
@@ -33,25 +33,23 @@ Either the `<dataset>` or the relative `<path>` within the dataset does not exis
...
 
-When the target is a directory, the response object contains a list of subdirectories and a list of files; if the target is a file (either a regular file or a symlink) it returns the file object for that path.
-
-When the target directory contains symlinks, the symlinks will be reported either in the `"file"` list (if the target of the symlink is a file) or in the `"directory"` list (if the target of the symlink is a directory).
+When the target is a directory, the response object contains a list of subdirectories and a list of files; if the target is a file it returns the file object for that path.
 
 ### Directory
 
-When the `<path>` refers to a directory within the dataset representation, Pbench returns a JSON object with a list of subdirectory objects and a list of file objects.
+When the `<path>` refers to a directory within the dataset representation, Pbench returns a JSON object containing a list of subdirectory objects (`directories`) and a list of file objects (`files`).
```
webbnh, line 42: `directories` and `files` are keys, right?  So, should the be in quotes?  E.g., `"directories"`?

docs/API/V1/contents.md
```diff
@@ -33,25 +33,23 @@ Either the `<dataset>` or the relative `<path>` within the dataset does not exis
...
 
 #### Directory object
 
-The directory object gives the name of the directory, and a URI that can be used with a subsequent `GET` operation to get directory data about that nested path.
+The directory object gives the name of the directory, and a URI that can be used with a subsequent `GET` operation to get directory data for that nested path.
```
webbnh, line 46: s/, and/ and/  (The second clause contains no verb....)

docs/API/V1/list.md
```diff
@@ -0,0 +1,125 @@
...
+
+#### Directory object
+
+The directory object gives the name of the directory, and a URI that can be used with a subsequent `GET` operation to get directory data about that nested path.
+
+#### File object
+
+The file object gives the name of the file and a URI that can be used with a subsequent `GET` operation to return the buffered byte stream of that file.
+
+Note that soft links within the dataset representation will result in a URI returning a linked file's byte stream or a linked directory's [directory object](Directory object).
```
webbnh, line 71: s/soft links/symlinks/

webbnh (commented):
I've resolved all the conversations from my initial post, but in the process I found a few more things....  In any case, I'm now caught up until your next  update.

docs/API/metadata.md
```diff
@@ -0,0 +1,43 @@
...
+
+The Pbench server will also calculate a default deletion date for the dataset based on the owner's retention policy and the server administrator's retention policy along with some other internal management context. These are all accessible under the `server` metadata key namespace.
+
+Clients can also set arbitrary metadata through the "dashboard" and "user" metadata namespaces:
+* The "dashboard" namespace can only be modified by the owner of the dataset, and is visible to anyone with read access to the dataset.
+* The "user" namespace is private to each authenticated user, and even if you don't own a dataset you can set your own private "user" metadata to help you categorize that dataset and to find it again.
+
+Metadata namespaces are hierarchical, and are exposed as nested JSON objects. You can address an entire namespace, e.g., `dashboard` or `dataset` and retrieve the entire JSON object, or you can address nested objects or values using a dotted metadata key path like `dashboard.contact.email` or `dataset.metalog.pbench.script`.
```
portante, line 15: Is it worth noting what `dataset.metalog.pbench` would return?  A JSON object of all the sub-fields?

In a user JSON context, it would be the dictionary value.  But in other schemes, like what OpenShift offers for labels and annotations, you can have a value for `dataset.metalog.pbench` separate from the sub-objects.  We might want to document that we don't support that.

docs/API/metadata.md
```diff
@@ -0,0 +1,43 @@
...
+
+There are currently three metadata namespaces.
+
+The `dataset` and `server` namespaces are defined and managed by Pbench. The `dashboard` namespace allows an authenticated client to define an arbitrary nested set of JSON objects. The `user` namespace is similar; except that where metadata in the `dashboard` namespace can only be modified by the owner of the dataset, and is visible to all clients with read access to the dataset, any authenticated user can set arbitrary values in the `user` namespace and those values are visible only to the same authenticated user.
+
+### Dataset namespace
+
+This defines the dataset resource, and contains metadata from the dataset tarball recieved from the Pbench agent, including the full contents of a `metadata.log` file which was generated when the agent packaged up the result data into a tarball.
+
+This namespace includes the resource name, which can be modified by the owner of the dataset, but all other metadata in this namespace is controlled by the server.
```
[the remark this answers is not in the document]
portante: Yes, it might be worth declaring the `metalog` namespace immutable.

docs/API/metadata.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Server namespace
+
+This defines internal server state related to a dataset that's not inherent to the representation of the data. Most of this is not of much use to external clients, but can be observed.
+
+Only `server.deletion` can be modified by an authenticated client on behalf of the owner of the dataset. When a dataset is uploaded, the server computes an expected time representing the user's default data retention period based on the time the dataset was uploaded. The owner of the dataset is allowed to modify this later, as long as the new timestamp is within the maximum allowed server data retention period.
```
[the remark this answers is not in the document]
portante: Similar to @webbnh's observations, when I read the first couple of sentences it seems like we were heading in the direction of exposing relevant server settings as immutable metadata.  With the notion of the owner allowed to modify it, this would seem to fit better as metadata for the dataset.

docs/API/metadata.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Dashboard namespace
+
+The server will never modify or directly interpret values in this namespace. An authenticated client representing the owner of a dataset can set any keys within this namespace to any desired values for retrieval later. All clients with read access to the dataset will see the same values.
```
portante, line 37: We need to be careful with "desired values".  Could the APIs be used to built up a huge set of keys and data the consume the database?  We might want to consider some kind of limits on this, configured by the server (total # of keys, total size of values, etc.).

docs/API/metadata.md
```diff
@@ -0,0 +1,43 @@
...
+
+__NOTE__: The server will in the future be able to use these values to filter the selected datasets for [datasets/list](V1/list.md).
+
+### User namespace
+
+This is treated almost exactly like the `dashboard` namespace, except that the values stored here are specific to the authenticated user. Each authenticated client may set distinct values for the same keys, and will be able to retrieve those values later. A client authenticated for some other user will see a completely different set of values. An unauthenticated client can neither set nor retrieve any `user` namespace values; such a client will always see the `user` namespace as empty.
```
portante, line 43: Similare comment here with the `Dashboard` namespace, we might want to consider limits on keys and values in some way.

portante (commented):
First pass set of comments on the outer `README.md` and the `metadata.md`.

docs/API/metadata.md
webbnh: Yeah, I assume that there will be a number of places where we want to institute limits to prevent abuse (number of keys, bytes per key, total bytes, etc.).  But, for the moment, I think we're relying on the fact that you have to be a registered user in order to create anything, so that should protect us from the main part of the onslaught.

portante: We are not just protecting against the malicious but the inadvertent.

docs/API/V1/README.md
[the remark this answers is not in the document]
webbnh: > I want to add the reference URL, and pointing a symlink at the `contents` API to get the file object again doesn't make much sense. (And leaves the caller floundering to figure out the proper URL.)

I'm not sure what you're saying here....  (In no case should we leave the caller floundering.)  For a symlink, I think we should return either the `contents` or `inventory` URL for the referent, depending on whether the referent is a directory or a file, respectively.  For a broken link, I would return either one -- I'm assuming that it'll generate a `NOT_FOUND`, either way.  If we've already reported the contents of the symlink in the file object, then the caller can already deal with the link if they want to.  And as long as the caller can consult the `type` before following the URL, I think everything is covered.

docs/API/V1/list.md
```diff
@@ -0,0 +1,125 @@
+# `GET /api/v1/datasets/list`
+
+This API returns an `application/json` document describing the set of datasets accessible to the authenticated client. (Or, for an unauthenticated client, all datasets with access "public".)
+
+The list of datasets may be further filtered by "owner", "access", "name" substring, or by creation date range using the query parameters.
+
+## Query parameters
+
+`access`    string \
+Select whether only `private` or only `public` access datasets will be included in the list. By default, all datasets readable by the authenticated user are included. For example, without constraints `/datasets/list` for an authenticated user will include all `public` datasets plus all datasets owned by the authenticated user; specifying `private` will show only the authenticated user's private datasets, while specifying `public` will show only `public` datasets (regardless of ownership).
```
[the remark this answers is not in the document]
webbnh: Right; my bad.

docs/API/metadata.md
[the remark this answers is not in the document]
webbnh: OK, I keep loosing my grip on this stuff (which is a "me problem", mostly).  So, we have "four" namespaces (at the moment, and ignoring "server configuration" and "user profile"), and they _all_ are attached to a dataset, such that the same keywords potentially have different values when applied to different datasets.  Yes?

If that's the case, then I'm way less concerned about `server.deletion` and the resource name.  I mean, I do think it would be nice if there weren't namespaces which were immutable except for one key, but that's much less of an issue than having namespaces in which some keys are tied to datasets and some keys aren't.

docs/API/V1/contents.md
webbnh: Since the text says that Pbench returns a JSON object, I would refer to [this definition](https://www.json.org/json-en.html), which says that object members are a "string" followed by a colon followed by a "value" (with optional whitespace sprinkled in there), and a "string" is "characters" enclosed in double-quotes.

But, yeah, if someone is implementing code around this, they are probably using Javascript or Python, in which case there is additional or different syntax involved...but, then we would be talking about Javascript or Python and not about JSON.  😉

docs/API/V1/contents.md
```diff
@@ -88,11 +91,44 @@ Note that symlinks within the dataset representation will result in a URI return
...
+
+```json
+{
+    "name": "reference-result",
+    "type": "sym",
+    "uri": "http://hostname/api/v1/datasets/contents/<id>/sample1"
+}
+```
+
+#### File object
```
webbnh, line 116: Should this have three `#`'s to match line 52 ("Directory Object")?  (Lines 49 & 50 reference that and this, respectively, so they seem like they should be peers (although, I can see how you want this to match line 94 ("Sub[d]irectory Object").)

docs/API/V1/contents.md
```diff
@@ -23,35 +28,33 @@ The return is a serialized JSON object with information about the named file.
...
-Note that symlinks within the dataset representation will result in a URI returning the linked file's byte stream or the linked directory's [directory object](#directory-object).
-
-#### Example
+When the `<path>` refers to a directory within the dataset representation,
+Pbench returns a JSON object with two list fields:
+* `directories` is a list of [subdirectory objects](#subirectory-object), and
+* `files` is a list of [file objects](#file-object)
 
 ```json
 {
```
webbnh, line 60: The objects in the `"directories"` list here (at lines 62 and 67) are missing their `"type"` members.

docs/API/V1/contents.md
```diff
@@ -0,0 +1,142 @@
...
+    "directories": [
+        {
+            "name": "1-iter1",
+            "uri": "http://hostname/api/v1/datasets/contents/<id><path>/1-iter1"
+        },
+        {
+            "sysinfo",
+            "uri": "http://hostname/api/v1/datasets/contents/<id><path>/sysinfo"
+        }
+        [...]
```
webbnh, line 70: Line 69 should end in a comma.

As for line 70, I would keep the elipsis but omit the square brackets (since in this context they would indicate a list).  Ditto line 89, and elsewhere.

docs/API/V1/contents.md
```diff
@@ -88,11 +91,44 @@ Note that symlinks within the dataset representation will result in a URI return
...
+The file system information includes:
+* `mtime`: the file's last modification time,
+* `size`: the size of the file,
+* `mode`: the file permissions (as an octal "mode" string), and
+* `type`: the file's type. The type values are:
+  * `reg`: Regular UNIX file
+  * `sym`: Symbolic link
+
+Note that symlinks to files within the dataset representation will result in a
+URI returning the linked file's byte stream.
```
webbnh, line 131: If the file is a symlink, do we get `mtime`, `size`, and `mode`, and, if so, do they refer to the symlink file itself or to its referent?  (I thought we were going to omit those three fields for symlinks.)

docs/API/V1/list.md
```diff
@@ -1,125 +1,118 @@
...
 
 `404`   **NOT FOUND** \
-Either the `<dataset>` or the relative `<path>` within the dataset does not exist
+Either the `<dataset>` or the relative `<path>` within the dataset does not
+exist.
 
 ## Response body
 
-This returns an `application/json` response body describing the target of `<path>`, in terms of "file" and "directory" JSON objects.
+This returns an `application/json` response body containing a list of objects
```
webbnh, line 89: s/This returns/This API returns/

(Or, perhaps, more properly, instead of "API", use "method"?  Although, if you make that change here, you should do the same in the analogous places in the other methods, as well....)

docs/API/V1/list.md
```diff
@@ -1,125 +1,118 @@
...
+which describe the datasets selected by the specified query criteria.
 
-When the target is a directory, the response object contains a list of subdirectories and a list of files; if the target is a file (either a regular file or a symlink) it returns the file object for that path.
+Each of these objects contains the following fields:
+* `resource_id`: The internal unique ID of the dataset within the Pbench Server.
+This value will be used to reference the dataset in most other APIs.
+* `name`: The resource name given to the dataset. While this has an initial
+default value related to the benchmark script, date, and user configuration
+parameters, this value can be changed by the owner of the dataset and is for
+display purposes and must not be assumed unique or definitive.
```
webbnh, line 98: s/assumed unique/assumed to be unique/

docs/API/metadata.md
```diff
@@ -1,43 +1,112 @@
...
+* The `dataset` and `server` namespaces are defined and managed by Pbench.
+* The `dashboard` namespace allows an authenticated client to define an
+arbitrary nested set of JSON objects associated with a specific dataset
+owned by the authenticated user.
+* The `user` namespace is similar; except that where metadata in the
+`dashboard` namespace can only be modified by the owner of the dataset, and is
+visible to all clients with read access to the dataset, any authenticated user
+can set arbitrary values in the `user` namespace and those values are visible
+only to the same authenticated user. Other users may set different values for
+the same `user` namespace keys on the dataset, or completely different keys.
```
webbnh, line 69: If I might rework the punctuation as well as the final clause:
```
* The `user` namespace is similar, except that, where metadata in the
`dashboard` namespace can only be modified by the owner of the dataset and is
visible to all clients with read access to the dataset, any authenticated user
can set arbitrary values in the `user` namespace, and those values are visible
only to the same authenticated user. Other users may set different values for
the same `user` namespace keys on the dataset, or they may set values for
completely different keys.
```

docs/API/metadata.md
```diff
@@ -1,43 +1,112 @@
...
 
 ### Dataset namespace
 
-This defines the dataset resource, and contains metadata from the dataset tarball recieved from the Pbench agent, including the full contents of a `metadata.log` file which was generated when the agent packaged up the result data into a tarball.
+This defines the dataset resource, and contains metadata recieved from the
```
webbnh, line 73: s/recieved/received/

webbnh, line 73: s/, and/ and/

docs/API/metadata.md
```diff
@@ -1,43 +1,112 @@
...
+Pbench Agent, including the full contents of a `metadata.log` file created
+while gathering results and during dataset packaging.
 
-This namespace includes the resource name, which can be modified by the owner of the dataset, but all other metadata in this namespace is controlled by the server.
+This namespace includes the resource name, which can be modified by the owner
+of the dataset. All other data in this namespace is controlled by the server
```
webbnh, line 78: s/All other data/All other values/ ?

docs/API/metadata.md
```diff
@@ -1,43 +1,112 @@
...
 
 ### Server namespace
 
-This defines internal server state related to a dataset that's not inherent to the representation of the data. Most of this is not of much use to external clients, but can be observed.
+This defines internal Pbench Server management state related to a dataset
+that's not inherent to the representation of the user's performance metrics.
+Most of this is not of much use to external clients, but can be observed.
 
-Only `server.deletion` can be modified by an authenticated client on behalf of the owner of the dataset. When a dataset is uploaded, the server computes an expected time representing the user's default data retention period based on the time the dataset was uploaded. The owner of the dataset is allowed to modify this later, as long as the new timestamp is within the maximum allowed server data retention period.
+The exception is `server.deletion`, which is a time after which the Pbench
```
webbnh, line 87: s/time/[date|point in time]/ ?

docs/API/metadata.md
```diff
@@ -1,43 +1,112 @@
...
+Server may choose to delete the dataset. This is computed when a dataset is
+received based on user profile preferances and server configuration; but it can
```
webbnh, line 89: s/preferances/preferences/

docs/API/metadata.md
```diff
@@ -1,43 +1,112 @@
...
+be modified by the owner of the dataset, as long as the new timestamp remains
+within the maximum allowed server data retention period.
 
 ### Dashboard namespace
 
-The server will never modify or directly interpret values in this namespace. An authenticated client representing the owner of a dataset can set any keys within this namespace to any desired values for retrieval later. All clients with read access to the dataset will see the same values.
+The server will never modify or directly interpret values in this namespace. An
+authenticated client representing the owner of a dataset can set any keys
+within this namespace to any valid JSON values (string, integer, boolean, list,
+or nested objects) for retrieval later. All clients with read access to the
```
webbnh, line 98: s/integer/number/  (I.e., we're not going to prohibit floating point numbers, are we?)

docs/API/metadata.md
```diff
@@ -1,43 +1,112 @@
...
+dataset will see the same values.
 
-__NOTE__: The server will in the future be able to use these values to filter the selected datasets for [datasets/list](V1/list.md).
+__NOTE__: The server will in the future be able to use these values to filter
+the selected datasets for [datasets/list](V1/list.md).
 
 ### User namespace
 
-This is treated almost exactly like the `dashboard` namespace, except that the values stored here are specific to the authenticated user. Each authenticated client may set distinct values for the same keys, and will be able to retrieve those values later. A client authenticated for some other user will see a completely different set of values. An unauthenticated client can neither set nor retrieve any `user` namespace values; such a client will always see the `user` namespace as empty.
+This is treated almost exactly like the `dashboard` namespace, except that the
```
webbnh, line 106: s/This is/This namespace is/

docs/API/metadata.md
```diff
@@ -1,43 +1,112 @@
...
+values stored here are specific to the authenticated user. Each authenticated
+client may set distinct values for the same keys, and will be able to retrieve
```
webbnh, line 108: Remove the comma.

webbnh (approved):
It would have been nice if you had done the line wrapping in a separate commit from the other changes (since they make pretty much every line appear changed in the source-diff view).  Fortunately, in the rich diff, GitHub is clever enough to note that the line breaks don't produce a difference.

In any case, this is coming together quite nicely.

Below, find my next installment of comments.

docs/API/metadata.md
```diff
@@ -0,0 +1,121 @@
...
+policy along with some other internal management context. The expected deletion
+date is accessible under the `server` metadata key namespace.
+
+Clients can also set arbitrary metadata through the "dashboard" and "user"
+metadata namespaces:
+* The "dashboard" namespace can only be modified by the owner of the dataset,
+and is visible to anyone with read access to the dataset.
+* The "user" namespace is private to each authenticated user, and even if you
+don't own a dataset you can set your own private "user" metadata to help you
+categorize that dataset and to find it again.
```
portante, line 32: When somebody deletes a dataset, I would assume we would remove that user metadata?  Or would we keep it around noting that the associated dataset has been deleted?

portante (changes requested):
Wonderful!

I am still not convinced of the separate APIs of "contents" and "inventory".  Why doesn't `GET` on a directory return the information about the files in the directory, with headers containing information about the directory, and a `GET` on the file return the contents, with headers that return information about the file?

docs/API/metadata.md
webbnh: Yeah, it would probably be worth mentioning that, if the dataset is deleted, other users' associated metadata is deleted automatically.  (This is done by the database engine -- we don't currently have any control over this...i.e., it's a _feature_!  😆)

webbnh (dismissed):
@dbutenhof, this is looking good.  However, I've got a couple of new (or forwarded) comments, and there are a half-dozen old ones which are still up for consideration.

> Why doesn't GET on a directory return the information about the files in the directory, with headers containing information about the directory, and a GET on the file return the contents, with headers that return information about the file?

@portante, the way I would like to look at this API is that the `contents` method provides support for a directory tree "walk":  it returns JSON objects containing a name, a type, and a URL, and, if the object refers a directory (or a symlink to a directory), then the returned URL is another `contents` method.  This is simple, straightforward, and self-consistent.

If the object refers to a file, then the method returns a URL for an `inventory` method.  The client knows from the object type that the URL will return a byte stream and not a JSON object.  But, the power here (in my opinion) is that the URL itself doesn't _have_ to contain a recognizable file path.  It could instead refer to an S3 ID or anything else that we find convenient.  The salient point is that the client knows that it can make a request to that URL and receive a byte stream response.

The fundamental question (in my mind) is whether we should support the client making requests for file contents without using the `contents` method.  That is, I would recommend that the `inventory` parameter be "opaque", and that access to it be provided through the `contents` method.

I understand that historically, people have been very cognizant of the process of collecting results, packaging them as a tarball, and using that tarball to replicate the client's directory structure on the server.  However, my instinct is that, going forward, that is _not_ going to be important.  I expect that much of the "results tree" is going to disappear, with the data placed inside a single PCP archive file.  And, I expect that the artifacts which we extract from the tarball are going to be stored in an object store, not a file system, and they are going to be indexed and served to the client not as files but as data streams and visualizations.

So, I really don't think we are going to need or even want a client-side-directory-path-based `inventory` method, and the more that we can do to hide that and abstract that away the better off we will be.

Thus, I think that having separate `contents` and `inventory` methods provides a valuable decoupling which will improve the flexibility of the Server and the clarity of the client-side of the interface.

portante:
> @dbutenhof, this is looking good. However, I've got a couple of new (or forwarded) comments, and there are a half-dozen old ones which are still up for consideration.
> 
> > Why doesn't GET on a directory return the information about the files in the directory, with headers containing information about the directory, and a GET on the file return the contents, with headers that return information about the file?
> 
> @portante, the way I would like to look at this API is that the `contents` method provides support for a directory tree "walk": it returns JSON objects containing a name, a type, and a URL, and, if the object refers a directory (or a symlink to a directory), then the returned URL is another `contents` method. This is simple, straightforward, and self-consistent.
> 
> If the object refers to a file, then the method returns a URL for an `inventory` method. The client knows from the object type that the URL will return a byte stream and not a JSON object. But, the power here (in my opinion) is that the URL itself doesn't _have_ to contain a recognizable file path. It could instead refer to an S3 ID or anything else that we find convenient. The salient point is that the client knows that it can make a request to that URL and receive a byte stream response.

Do we not have a tar ball that is a capture of the file system state at the time of pbench-move-results?  Why are we spending so much time to turn this into something it isn't?

Do we hide the knowledge that it is a tar ball from the pbench agent interfaces?  I don't believe so.  The `benchmark_run_dir` and `benchmark_results_dir` variables are exported so that user benchmark scripts can refer to them, and other script interfaces in the other benchmark convenience scrsipts also have access.  This is part of the interface.

If we had a tar ball structure that was opaque then it would make sense to provide APIs to it that afford access to the abstraction that makes sense.

Let's make APIs for what it is, and then we can work to change the pbench agent to hide its data collection methods in the future so that we have the freedom to be as abstract as we want.

In summary, I don't support this interface direction.

But, if the team wants to choose to follow this path, that is fine, but somebody has to own this and make it happen.

webbnh:
> Let's make APIs for what it is, and then we can work to change the pbench agent to hide its data collection methods in the future so that we have the freedom to be as abstract as we want.

In my experience, users value stable interfaces and backward compatibility...so, the design of this interface is something that we're likely to be stuck with.  As such, I would like to see it be as flexible, in terms of the Server implementation, as possible.  Introducing abstractions in the future likely won't mean much if we specify too much concrete stuff in the present.

> Do we not have a tar ball that is a capture of the file system state at the time of pbench-move-results? Why are we spending so much time to turn this into something it isn't?

We aren't.  We're spending time designing an interface method which allows the client to walk the directory structure of the captured file system; and, we're providing another interface method which allows the client to fetch the contents of files in the captured file system.  What I'm trying to do is to make sure that those two interface methods don't get too tied together (or even collapsed into a single one), and I'm trying to keep the method which fetches the contents as decoupled from the directory structure as I can, so that it will retain the maximum flexibility moving forward.

> Do we hide the knowledge that it is a tar ball from the pbench agent interfaces?

I think we're talking about the Server, here, not the Agent.  And, I think the heart of the question is, once the data has been transferred to the Server, how do we organize and present it to the client...and, at that point, how important is the original, client-side filesystem organization?  I'm trying to explore the idea that it isn't important...or, at least, it _won't_ be important in the near future.

> In summary, I don't support this interface direction.
> 
> But, if the team wants to choose to follow this path, that is fine, but somebody has to own this and make it happen.

The good news is that, if having two, separate interfaces turns out to be the wrong choice, it's fairly easy to evolve toward a single, more complicated interface like the one that you're suggesting.  I'm not sure that the reverse is true.  In any case, I think Dave is owning this and he and Siddardh are making it happen.

docs/API/V1/README.md
```diff
@@ -0,0 +1,119 @@
...
+retention policy and the server administrator's retention policy.
+
+Clients can also set arbitrary metadata through the `dashboard` and `user`
+metadata namespaces. The `dashboard` namespace can only be modified by the
+owner of the dataset, and is visible to anyone with read access to the dataset.
+The `user` namespace is private to each authenticated user, and even if you
+don't own a dataset you can set your own private `user` metadata to help you
+categorize that dataset and to find it again.
+
+The primary dataset `resource_id` is the computed MD5 of the dataset tarball.
```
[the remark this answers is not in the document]
webbnh: I stumbled across the answer:  FIPS 140 compliance precludes the use of MD5 even for non-cryptographic purposes.  🤯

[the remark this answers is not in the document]
dbutenhof: As I'm sure I pointed out somewhere, we have [PBENCH-324](https://issues.redhat.com/browse/PBENCH-324) recording the intent to drop MD5. I recall at the time we first discussed that, there was concern about the fact that AWS relied on MD5 checksums for S3. Out of curiosity I just checked AWS and found a Feb 2022 article saying that they also support SHA-1 and SHA-256 as well as CRC-32 and CRD-32C. (But apparently not SHA-512?) I wasn't able to find anything about that in Ceph, however, so they may be lagging (or just harder to locate...)

Anyway, yeah, we should consider that, might even want to make it a MVP for whatever we call our "downstream" hardened release. For now, this is just trying to document what we have now. (Despite a few liberties with the definition of "now" like documenting Siddardh's inventory API which at the time was still in flight.)

docs/API/V1/list.md
```diff
@@ -0,0 +1,118 @@
...
+[
+    {
+        "name": "pbench-fio_config_2022-06-29:00:00:00",
+        "resource_id": "07f0a9cb817e258a54dbf3444abcd3aa",
+        "metadata": {"user.favorite": true}
+    },
+    {
+        "name": "the dataset I created for fun",
+        "resource_id": "8322d8043755ccd33dc6d7091d1f9ff9",
+        "metadata": {"user.favorite": null}
```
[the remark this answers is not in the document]
dbutenhof: Any undefined key in the `user` or `dashboard` namespaces will be reported as `null`. They don't exist until someone (presumably the dashboard Javascript) sets them. In this case, one of the datasets has been given the value `true`, the other hasn't been set.

docs/API/metadata.md
dbutenhof: It's not true that we don't have any control; rather, we've exercised that control by telling SQLAlchemy (and therefore PostgreSQL) that they're related and that deletion of the "parent" should delete all the "children".

I suppose I could say explicitly that it goes away, but that feels silly as it isn't visible to the user anyway... there's no way to get/set metadata without a dataset handle.

docs/API/V1/contents.md
webbnh: I'm thinking, broadly, of my experiences with symlinks in the filesystem.  Generally speaking, when accessing a symlink, tools instead access the file system object to which it refers.  This is obviously not true of `ls`, but to my surprise it was true of `chown` (unless you include the `-h` switch).

And, per the `chmod` man page:
> `chmod` never changes the permissions of symbolic links; the `chmod` system call cannot change their permissions.  This is not a problem since the permissions of symbolic links are never used.  However, for each symbolic link listed on the command line, `chmod` changes the permissions of the pointed-to file.

So, I'm (generally) thinking that the `mtime`, `size`, and `mode` of the symlink file itself aren't worth bothering about.  (Obviously, the `mode` is useless -- `ls` reports it as `lrwxrwxrwx` -- and, as you said, the `size` is of limited use; the `mtime` could conceivably be significant, but it's probably not as important as the `mtime` of the file-object to which the link refers.)

So, I would be inclined to do whatever is easiest for symlinks; and, I suspect that that is to omit the three fields altogether!  And, I would do so for both file objects and directory objects....

docs/API/metadata.md
webbnh: I do think it's worth pointing out that, if a user creates private "user" metadata associated with a dataset owned by some _other_ user, when that dataset is deleted, **all** private "user" metadata is also deleted.  (I don't know what a user would store in "user" metadata which would be valuable after the associated dataset had been deleted, but it's probably prudent to point out this effect.)

docs/API/V1/README.md
```diff
@@ -1,8 +1,8 @@
-# Pbench server V1 API documentation
+# Pbench Server V1 API documentation
 
 The documents in this set describe the APIs supported by the Pbench 1.0 server.
```
webbnh, line 3: s/server/Server/

docs/API/V1/list.md
```diff
@@ -0,0 +1,118 @@
...
+
+`offset` integer \
+"Paginate" the selected datasets by skipping the first `offset` datasets that
+would have been selected by the other query terms. This can be used with
+`limit` to progress through the full list in smaller chunks either to assist
+with a paginated display or to limit data transfer requirements.
+
+`owner` string \
+The `username` property of a registered user. Only datasets owned by that user
+will be included in the list.
```
webbnh, line 51: The descriptions of the other parameters begin with a verb. Perhaps something like

> Select only datasets owned by the specified username.

docs/API/V1/list.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Response status
+
+`401`   **UNAUTHORIZED** \
+The client did not provide an authentication token and there is no public
+dataset with the name `<dataset>`.
+
+`403`   **FORBIDDEN** \
+The named `<dataset>` is not public and the authenticated user lacks
+authorization to read it.
```
webbnh, line 81: There presumably should be commas before these `and`'s.

docs/API/V1/list.md
```diff
@@ -0,0 +1,118 @@
...
+
+`404`   **NOT FOUND** \
+Either the `<dataset>` or the relative `<path>` within the dataset does not
+exist.
+
+## Response body
+
+This returns an `application/json` response body containing a list of objects
+which describe the datasets selected by the specified query criteria.
```
webbnh, line 90: s/This returns/This API returns/

(Or, perhaps, more properly, instead of "API", use "method"? Although, if you make that change here, you should do the same in the analogous places in the other methods, as well....)

docs/API/metadata.md
```diff
@@ -0,0 +1,121 @@
...
+`dataset.metalog.pbench`.
+
+### Server namespace
+
+This defines internal Pbench Server management state related to a dataset
+that's not inherent to the representation of the user's performance metrics.
+Most of this is not of much use to external clients, but can be observed.
+
+The exception is `server.deletion`, which is a time after which the Pbench
+Server may choose to delete the dataset. This is computed when a dataset is
```
webbnh, line 94: s/time/date/ ?

webbnh (approved):
I found one more uncapitalized "server", and I've brought forward my unresolved comments from three weeks ago, since the originals are getting hard to find.

docs/API/metadata.md
dbutenhof: Again, I don't see how this is relevant at this level of description. It'd be a server bug (dangling resources) if we didn't delete them; but from the point of view of the caller, it doesn't matter as there's no possible way to access them without the dataset ID.

webbnh: Speaking most technically, which is to say, from the _interface_ perspective, I suppose it's true that there is no way for a requestor to know what happens when the dataset is deleted.  It is possible that the requestor still possesses the dataset ID, but the server will respond that the dataset does not exist, and, therefore, the fate of the metadata will be unknowable.

But, the people who are providing the UX for the end-users are going to be working based on this document.  And so, in support of their users, it would be good to make it clear that the lifetime of private "user" metadata is not entirely under the control of the user which created it, and that it will end when the dataset that it's associated with is deleted.

docs/API/V1/list.md
```diff
@@ -47,8 +47,8 @@ would have been selected by the other query terms. This can be used with
 with a paginated display or to limit data transfer requirements.
 
 `owner` string \
-The `username` property of a registered user. Only datasets owned by that user
-will be included in the list.
+Select only datasets owned by the specified username. Unless the username
+matches the authenticated user, only "public" datasets can be selected.
```
webbnh, line 51: > Unless the username matches the authenticated user

...or the user has admin access?

docs/API/V1/user.md
```diff
@@ -1,6 +1,7 @@
 # `GET/PUT/DELETE /api/v1/user/<username>`
 
-Describe the Pbench Server V1 API to `GET` (retrieve), `PUT` (modify), or
-`DELETE` (remove) a registered user's profile.
+This API allows viewing, changing, or deleting the Pbench Server profile for
+a designated user. The authenticated user must be the user being accessed, or
+must hold an administrator role.
```
webbnh, line 5: No comma required....

webbnh (approved):
Looks great!

(If you are inclined to decline the first of my comments below, then you can ignore the second.)

docs/API/metadata.md
webbnh: I see that you've added the comment about the namespace going away down lower (at line 72); hopefully, that's sufficient.

webbnh (commented):
(These failed to get attached to my previous review.)

docs/API/V1/list.md
webbnh: As long as there's a reasonable way to find all the places when it's time to update them.  (Putting a token like "admin" there might make them easy to find....  😉)

docs/API/V1/list.md
```diff
@@ -70,19 +70,22 @@ and only `public` access datasets will be selected.
...
+
+See [Access model](../access_model.md)
+
 ## Response status
 
 `401`   **UNAUTHORIZED** \
-The client did not provide an authentication token, and there is no public
-dataset with the name `<dataset>`.
+The client did not provide an authentication token but asked to filter datasets
+by `owner` or `access=private`.
```
webbnh, line 84: I'm confused:  if I'm not logged in, I cannot request public datasets which are owned by a particular user?

docs/API/metadata.md
```diff
@@ -120,5 +120,11 @@ distinct values for the same keys, or use completely different keys, and can
 retrieve those values later. A client authenticated for another user has
 its own comletely unique `user` namespace.
 
+The `user` metadata namespace behaves as a user-specific sub-resource under the
+dataset. Any authenticated client has UPDATE and DELETE access to this private
+sub-resource as long as the client has READ access to the dataset. See
```
webbnh, line 125: There's an interesting foible here:  if a user un-publishes a dataset to which a second user has attached `user` metadata, then the second user will lose access to his/her metadata even though it presumably will continue to _exist_ as long as the dataset does.  (And, if the dataset is re-published, then the second user's metadata will become available again.)

I'm not saying that this is good or bad, I'm just pointing out what will likely fall out of this.  🙃

docs/API/access_model.md
```diff
@@ -0,0 +1,36 @@
...
+user accounts on the server. These users will be granted full CRUD access to
+all server data, including
+
+- All datasets
+- All user profiles
+- Server configuration settings
+
+The `ADMIN` role can be assigned to a user by another user with the `ADMIN`
+role through the [user](V1/user.md) API. The server management CLI (described
+elsewhere) must be used to assign the `ADMIN` role to the first user.
```
webbnh, line 32: The last sentence here feels like it needs some wordsmithing.  Perhaps something like

> The server management CLI (described elsewhere) may also be used to assign the `ADMIN` role to a user.  (Note that, initially, there are no users with the `ADMIN` role.)

I don't know...it's just that the "first user" doesn't necessarily have to be an `ADMIN`....  🦐

webbnh (approved):
Looks good.

(We should probably start thinking about actually merging this and putting additional stuff in the next PR....)

docs/API/V1/contents.md
```diff
@@ -28,15 +28,20 @@ E.g., `authorization: bearer <token>`
 `content-type: application/json` \
 The return is a serialized JSON object with information about the named file.
 
+## Resource access
+
+* `READ` access to the `<dataset>` resource
```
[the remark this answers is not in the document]
dbutenhof: Done

docs/API/metadata.md
dbutenhof: Yeah, it's true, and I have no idea whether it's "good" or "bad". If the dataset transitions to PRIVATE (and the owner of the `user` metadata doesn't have `ADMIN` role), and it *remains* private permanently, then the data is invisible. But I suspect that "unpublish" is probably a rare event? And maybe it's a useful feature that "publish" will *"restore"* the data?

Don't know; but I added a note in the new resource section.

dbutenhof (commented):
> (We should probably start thinking about actually merging this and putting additional stuff in the next PR....)

Then "we" should probably stop posting interesting comments. 😆 

(Or maybe "we" should just stop reading them, but "whatever"... 🍤 )

docs/API/V1/README.md
```diff
@@ -8,6 +8,57 @@ Once you know the hostname of a Pbench Server, you can ask for the API
...
+
+### Users
+
+User resources are identified by a `username` property, which must be unique
+among all users registered on the Pbench Server. The user resource has a
+"user profile" that includes first and last name, and a contact email.
+
+A user resource is the "owner" of each [dataset](#datasets) managed by the
+Pbench Server. If a user is deleted, then any datasets owned by that user
+become orphaned; datasets with PUBLIC access are still accessible to other
```
portante, line 38: Do we cover somewhere what it means to be "orphaned"?

It might be good to reference another section here that talks about the behaviors of orphaned datasets.  An administrator of the Pbench Server would need to know what the impact will be on the server operation.

docs/API/V1/list.md
```diff
@@ -0,0 +1,121 @@
...
+datasets.
+
+## Resource access
+
+* Only `<dataset>` resources selected by the filter to which the authenticated
+user has READ access will be returned.
+
+See [Access model](../access_model.md)
+
+## Response status
```
portante, line 80: Should we document the success case explicitly?

docs/API/metadata.md
```diff
@@ -0,0 +1,130 @@
...
+* `project` would return `["OCP", "customer"]`
+* `user.tracker.examined` would return `"2022-05-15"`
+* `user.analysis` would return `{"cpu": "high", "memory": "nominal"}`
+
+## Metadata namespaces
+
+There are currently four metadata namespaces.
+
+* The `dataset` and `server` namespaces are defined and managed by Pbench.
+* The `dashboard` namespace allows an authenticated client to define an
```
portante, line 61: Perhaps we should consider a name change from `dashboard` to ... something else?  Perhaps "owner"?

It feels to me like the name "dashboard" is tied to the UI, but this is an API which can be used independent of the Pbench Dashboard.

portante (approved):
Approval, with reservations.

I am approving to get the existing state of this API documentation merged because it is so valuable.

I am also approving _most_ of the contents as I understand it.

But I am not approving the "contents" and "inventory" APIs.

portante (changes requested):
Sorry meant to "approve" will my comments, as I am no longer in favor of "approvals" with outstanding comments.

docs/API/V1/README.md
dbutenhof: I actually realized as I wrote this section that I don't believe I wrote a unit test to *verify* what happens when you delete a user. We probably need an admin mode API to change ownership of a dataset, and I wonder if maybe we should just refuse to delete a user if the user owns datasets rather than dealing with all this. That is, if you own datasets, then they need to be re-assigned *before* deleting the user.

docs/API/V1/list.md
dbutenhof: Probably! Good idea.

docs/API/metadata.md
dbutenhof: I've been thinking of this since forever, and especially when I changed the `user` namespace. I just haven't gotten around to doing it.

I'd been thinking `client`, but never entirely convinced myself since that's somewhat of an alias for "user" in this context. So `owner` might make sense. Or we could go for something more "meta" like, um, `meta`. Or `global` (which breaks `owner`'s symmetry with `user` to avoid perhaps implying that the values are only visible to the owner?)

docs/API/V1/list.md
dbutenhof: We went back and forth on this a long time ago; but if we want unauthenticated clients to be able to look up usernames, we'd need to change the validator and the messaging.

docs/API/metadata.md
portante: Naming is hard.

portante:
FWIW, I'd be happy to resolve all comments by moving them to issues to be tracked in another PR just to get this merged as is.

dbutenhof:
> FWIW, I'd be happy to resolve all comments by moving them to issues to be tracked in another PR just to get this merged as is.

Yeah, I'd be happy to finish the framework and move on to specifics (and writing additional pages). The bugs and loose ends can be tucked under [PBENCH-753](https://issues.redhat.com/browse/PBENCH-753) as stories and/or bugs. I can do that this afternoon (after my hike), if you don't get to it first.

I believe that Webb approved the last round despite his outstanding conversations, but resolving his comments for him doesn't feel right. (And also makes me lean towards your new policy of not approving with open conversations. 👍🏻 )

dbutenhof:
> But I am not approving the "contents" and "inventory" APIs.

Yes, I know, you still prefer a single combined API, although I still don't comprehend why you think that's an actual advantage. (And also, as something of a technicality, we're talking about *documentation* of the current implementation strategy, which isn't really the place to be "not approving" based on potential alternate implementations.)

In any case, while out hiking a potential compromise occurred to me. My principal objection (and I believe Webb's) is that I want the expected response to be under the control of the caller, not a surprise based on the URL's trailing path. I also don't like returning metadata as non-standard response headers, as that makes a much "muddier" API specification.

But what if we select the "JSON metadata" (current `/datasets/contents`) or "file byte stream" (current `/datasets/inventory`) based on direct client provided input, instead of changing it internally based on the target?

This could be either simply based on the `accept` header (`application/json` for a JSON metadata or `application/octet-stream` for the file byte stream), or an explicit query parameter to be more explicit (`?view=metadata` or `?view=stream` for example).

If you ask for "metadata", that works regardless of the path because we can always report metadata. If you ask for "stream" and point at a directory, you get a type mismatch error.

That would allow traversing the entire tree through a single API, without the "surprise" factor.

The reason I like the `?view` query parameter is for ease of creating "url" links in the metadata objects returned: they'd be self-contained rather than having to know which datatype to request. Otherwise the client needs to figure it out, which I'd rather avoid.

(My idea of including the `"url"` is, at least in part, a prejudice formed from OneView, where our architectural rule was that URLs are opaque and not to be constructed or deconstructed by the client... which means that you can't get a sub-resource unless the URL is reported explicitly by the resource. And in general I think there's a lot of benefit in that rule, though I also appreciate that it can sometimes be taken too far...)

portante (approved):
Resolved my comments and added them to PBENCH-753 as sub-tasks.

webbnh (approved):
> My principal objection (and I believe Webb's) is that I want the expected response to be under the control of the caller, not a surprise based on the URL's trailing path. I also don't like returning metadata as non-standard response headers, as that makes a much "muddier" API specification.

Those are pretty much my feelings as well:  the requestor should know in advance what the body of the request will be (although, I suppose there is some merit to using the `accept` header to control that, but I would prefer that it be determined by the URI).  I also don't care for the non-standard response headers.

Nevertheless, having the result determined by a query parameter seems reasonable:  that's an explicit action taken by the requestor, which is nearly as specific as selecting a different API method.  (In code, we would call that a "control parameter", and we would look at it suspiciously and ask if it shouldn't be implemented via a separate function, but using a separate method is not _always_ the right answer in those cases....)

> My idea of including the "url" is, at least in part, a prejudice formed from OneView, where our architectural rule was that URLs are opaque and not to be constructed or deconstructed by the client.

There are times when a RESTful API should be composable by the client, but there are also situations in which an opaque URL is appropriate; since our API is returning a URL for the client to use, I think it's good and appropriate for it to be opaque -- I think it will be good to have the flexibility which that affords as we swap out storage for these artifacts.

Anyway, I found a couple of missing words, one of which I noted below.  If we don't care, I'll just resolve the conversation.

webbnh:
> I believe that Webb approved the last round despite his outstanding conversations, but resolving his comments for him doesn't feel right.

I applaud your instinct not to resolve my conversations for me...but looking now I don't see any that were unresolved at the time....  🍤

So, I'm going to go ahead and pre-resolve the new one that I opened...I'll let you decide whether to address it in this PR or another one.

webbnh:
> looking now I don't see any that were unresolved at the time...

And that would be because I resolved them before doing the review.  🤯

docs/API/V1/README.md
```diff
@@ -8,6 +8,57 @@ Once you know the hostname of a Pbench Server, you can ask for the API
...
+
+### Datasets
+
+The dataset resource encompasses everything the Pbench Server knows about a
+performance run captured by the Pbench Agent through a "dataset tarball". This
+includes the physical files transferred from the agent along with all backend
+databases maintained by the server.
+
+The API identifies a dataset using the dataset resource ID, which is a hex
+representing a checksum hash of the dataset tarball. This ID can't change during
```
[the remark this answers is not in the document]
dbutenhof: OK, now I want to go create a new Pbench Server resource called a "hex". I'm not sure what it'll do, but the documentation should probably make it sound scary...

dbutenhof (commented):
> And that would be because I resolved them before doing the review. exploding_head

It was the hex, right?

webbnh:
> > And that would be because I resolved them before doing the review. :exploding_head:
> 
> It was the hex, right?

It must have been!

PR #2892, opened 2022-06-13
Dependency fixes

ndokos:
Centos-Stream-8/9: change the 'ansible' dependency to 'ansible-core'.
Everybody else (at least for now) can get its hands on an 'ansible'
package (EPEL packages it for RHEL: it has a dependency on
'ansible-core' and includes some curated roles).


portante (commented):
I see only the change for CentOS.  Perhaps we can make separate PRs for these changes so that we can very the others while waiting for the resolution on the CentOS issue.

portante (changes requested):
Let's first fix this on `main`, and then create a backport.

PR #2894, opened 2022-06-14
Add dependency for Fedora 36

ndokos:
`pbench-verify-sysinfo-options` requires the python ifaddr module. Fedora 36 does not install it by default.

`ifaddr` is in `requirements.txt`, but we prefer a distro package if available.

### DIFF
--- a/agent/rpm/pbench-agent.spec.j2
+++ b/agent/rpm/pbench-agent.spec.j2
@@ -39,6 +39,11 @@ Requires:  python3, python3-pip
 # RPMs for modules in requirements.txt
 Requires:  python3-bottle, python3-cffi, python3-click, python3-daemon
 Requires:  python3-jinja2, python3-redis, python3-requests, python3-sh
+
+%if 0%{?fedora} >= 36
+Requires: python3-ifaddr
+%endif
+
 # RPMs for module dependencies
 Requires:  python3-psutil
 %endif


webbnh (approved):
Consider removing the conditionalization.

dbutenhof (approved):
It's probably premature to guess whether it's `== 36` or `>= 36` but that's good enough for now. 😁

agent/rpm/pbench-agent.spec.j2
```diff
@@ -39,6 +39,11 @@ Requires:  python3, python3-pip
 # RPMs for modules in requirements.txt
 Requires:  python3-bottle, python3-cffi, python3-click, python3-daemon
 Requires:  python3-jinja2, python3-redis, python3-requests, python3-sh
+
+%if 0%{?fedora} >= 36
```
portante, line 43: If that module is available for CentOS and RHEL, we should drop the conditional entirely.

portante (changes requested):

agent/rpm/pbench-agent.spec.j2
```diff
@@ -39,6 +39,11 @@ Requires:  python3, python3-pip
...
+Requires: python3-ifaddr
+%endif
```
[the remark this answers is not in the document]
ndokos: No - it's only available as an RPM package on F35 and F36. Everybody else does a `pip install` to get it. That's what F36 tried to do as well, but that failed because of the PYTHONPATH problems. So I think it has to be conditional.

agent/rpm/pbench-agent.spec.j2
ndokos: It is not: it's only available on Fedora.

PR #2896, opened 2022-06-14
CentOS-Stream-8/9 dependency fixes - backport to b0.71

ndokos:
Centos-Stream-8/9: change the 'ansible' dependency to 'ansible-core'.
Everybody else (at least for now) can get its hands on an 'ansible'
package (EPEL packages it for RHEL: it has a dependency on
'ansible-core' and includes some curated roles).

### DIFF
--- a/agent/rpm/pbench-agent.spec.j2
+++ b/agent/rpm/pbench-agent.spec.j2
@@ -19,11 +19,11 @@ Requires:  ansible
 %endif
 
 
-%if 0%{?rhel} == 7 
+%if 0%{?rhel} == 7
 Requires:  python3, python3-pip
 %endif
 
-# this applies to CentOS-Stream-8 as well
+# This condition will be true on a CentOS-Stream-8 system as well
 %if 0%{?rhel} == 8
 Requires:  python36, python3-pip
 # RPMs for modules in requirements.txt
@@ -32,7 +32,7 @@ Requires:  python3-cffi, python3-click, python3-requests
 Requires:  python3-docutils, python3-psutil
 %endif
 
-# this applies to CentOS-Stream-9 as well
+# This condition will be true on a CentOS-Stream-9 system as well
 %if 0%{?rhel} == 9
 Requires:  python3-pip
 # RPMs for modules in requirements.txt
--- a/agent/rpm/pbench-agent.spec.j2
+++ b/agent/rpm/pbench-agent.spec.j2
@@ -8,16 +8,22 @@ URL:            https://github.com/distributed-system-analysis/pbench
 Source0:        pbench-agent-%{version}.tar.gz
 Buildarch:      noarch
 
-%if 0%{?rhel} < 9
-Requires:  ansible
-%else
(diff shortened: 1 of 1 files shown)


agent/rpm/pbench-agent.spec.j2
```diff
@@ -8,16 +8,22 @@ URL:            https://github.com/distributed-system-analysis/pbench
...
+
+%if 0%{?centos} >= 8
 Requires:  ansible-core
+%else
+Requires:  ansible
 %endif
 
-%if 0%{?rhel} == 7
+
+%if 0%{?rhel} == 7
```
webbnh, line 22: There's a trailing space on this line (dunno whether we should care...but that's what makes it show up in the diffs).

agent/rpm/pbench-agent.spec.j2
```diff
@@ -8,16 +8,22 @@ URL:            https://github.com/distributed-system-analysis/pbench
...
 
-%if 0%{?rhel} == 7
+
+%if 0%{?rhel} == 7 
 Requires:  python3, python3-pip
 %endif
 
+# this applies to CentOS-Stream-8 as well
 %if 0%{?rhel} == 8
 Requires:  python36, python3-pip
```
webbnh, line 28: I'm not quite sure what this comment is trying to tell me...is it saying that, on CentOS Stream 8, `{rhel}` will be `8`, so that this requirement will be added?  If so, I'm glad you added the comment, because that's rather counter-intuitive...and I'd be inclined to make the comment a little more expressive.  Perhaps, "This condition will be true on CentOS-Stream-8 as well" (or even `Note:  {?rhel} will evaluate to '8' on CentOS-Stream-8`).

Likewise for line 35.

agent/rpm/pbench-agent.spec.j2
```diff
@@ -34,6 +41,7 @@ Requires:  python3-cffi, python3-requests
 Requires:  python3-docutils, python3-psutil
 %endif
 
+
```
webbnh, line 44: Did you intend to add an extra blank line here?

webbnh (approved):
Looks good -- I'm glad the mysteries have been solved!

(I found a few nits, and perhaps you could strengthen a couple of the comments.)

agent/rpm/pbench-agent.spec.j2
ndokos: That's exactly what I'm saying.  I'll change the comments.

agent/rpm/pbench-agent.spec.j2
ndokos: I think it's mostly the extra empty line that I added: I was trying to separate sections more visibly.  I'll get rid of the trailing space.

agent/rpm/pbench-agent.spec.j2
ndokos: I wanted a larger break between RHEL and Fedora stuff. I find it helpful, but YMMV.

PR #2898, opened 2022-06-15
Fix rpmlint breakage

dbutenhof:
PBENCH-765

Changes were made to conform to Fedora 36 rpmlint changes to use `--rpmlintrc` instead of `--file`, but this breaks the other platforms.

Revert that change. Supporting Python3.10 and Fedora 36 is an ongoing challenge, but breaking existing platforms doesn't move us forward.


portante (changes requested):
Let's make a wrapper.

webbnh (approved):
Given that this is blocking Dave (and will presumably soon be blocking me), I think reverting now and developing a wrapper (or just conditionalization around the switch) in a separate PR is probably best.

dbutenhof:
If Peter's workaround in #2900 makes both versions happy, let's go with that. Besides, "2900" is a nice round number, and that's got to be good. 😆

PR #2899, opened 2022-06-15
Simple utility script to reset Server state

dbutenhof:
PBENCH-766

This script clears out an existing server installation, reading the configuration from `pbench-server.cfg` (via `pbench-config`) and clearing the PostgreSQL database, the Elasticsearch indices and templates, and the ARCHIVE, INCOMING, and RESULTS file systems.

NOTE: Run `systemctl stop pbench-server` before running this, as root, *then* run this
script under `su --login pbench`, and then exit the `su` and restart the server. The server will initialize the new SQL schema, create the Elasticsearch templates, and so forth, as usual.

It might be noted that while running a self-contained development pod with containerized private server, Elasticsearch, and PostgreSQL would mitigate some of this, we will still be faced with redeployments referencing external backend components that aren't reset simply by restarting the pod.

### DIFF
--- a/server/bin/pbench-server-erase.sh
+++ b/server/bin/pbench-server-erase.sh
@@ -9,7 +9,14 @@
 # 3. Delete the ARCHIVE, INCOMING, and RESULTS file system trees
 #
 # NOTE: The Pbench server should be *stopped* previously as root, and this
-# script should be run under `su --login pbench`.
+# script should be run under `su --login pbench`. For example:
+#
+# systemctl stop pbench-server
+# su --login pbench
+# export _PBENCH_SERVER_CONFIG=/opt/pbench-server/lib/config/pbench-server.cfg
+# pbench-server-erase.sh
+# exit  # (back to root account)
+# systemctl start pbench-server
 
 es_host=$(pbench-config host elasticsearch)
 es_port=$(pbench-config port elasticsearch)
--- a/server/bin/pbench-server-erase.sh
+++ b/server/bin/pbench-server-erase.sh
@@ -1,0 +1,69 @@
+#! /bin/bash
+
+# Erase Pbench Server 1.0 state completely. This is a simple and "brute force"
+# sequence with few "blade guards", so be careful! (You have been warned!)
+#
+# 1. Delete the Elasticsearch indices and templates using the configured
+#    prefix value;
+# 2. Delete the configured PostgreSQL database;
+# 3. Delete the ARCHIVE, INCOMING, and RESULTS file system trees
+#
+# NOTE: The Pbench server should be *stopped* previously as root, and this
+# script should be run under `su --login pbench`.
+
+es_host=$(pbench-config host elasticsearch)
+es_port=$(pbench-config port elasticsearch)
+
+if [[ -z ${es_host} || -z ${es_port} ]] ;then
+    echo "Host is missing Elasticsearch connection data" >&2
+    exit 1
+fi
+
+elastic="http://${es_host}:${es_port}"
+
+es_prefix=$(pbench-config index_prefix Indexing)
+
+if [[ -z ${es_prefix} ]] ;then
+    echo "Host is missing Elasticsearch prefix" >&2
+    exit 1
+fi
+
+# Clear out the associated Elasticsearch indices and templates
+
+echo "Clearing Elasticsearch ${elastic}, prefix '${es_prefix}'"
+
+curl -X DELETE "${elastic}/${es_prefix}.*"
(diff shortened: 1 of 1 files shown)


server/bin/pbench-server-erase.sh
```diff
@@ -0,0 +1,69 @@
...
+psql ${ps_server} -c "drop database ${ps_db}"
+
+archive=$(pbench-config pbench-archive-dir pbench-server)
+incoming=$(pbench-config pbench-incoming-dir pbench-server)
+results=$(dirname ${incoming})/results
+
+if [[ ! -d ${archive} || ! -d ${incoming} || ! -d ${results} ]] ;then
+    echo "Host archive (${archive}), incoming (${incoming}) or results (${results}) is missing" >&2
+    exit 1
+fi
```
webbnh, line 60: Are you sure you want to exit after issuing this diagnostic?  (Are you sure that you want to make this check at all?...the `rm -rf` command doesn't care....)

webbnh (dismissed):
Looks OK to me; I just have two nits for you.

Beyond those, regarding the statement in the description:
> It might be noted that while running a self-contained development pod with containerized private server, Elasticsearch, and PostgreSQL would mitigate some of this, we will still be faced with redeployments referencing external backend components that aren't reset simply by restarting the pod.

My intent with the canned Server was to have everything contained inside the pod.  The script here cleans up Elasticsearch and PostgreSQL, each of which are inside containers in the pod; it also cleans up the archived and upacked tarballs, which are stored in the pod (volatile) file system.  So, I don't see what this script covers which isn't covered by restarting the pod...what did I miss?

npalaska (approved):

server/bin/pbench-server-erase.sh
```diff
@@ -0,0 +1,69 @@
...
+# Erase Pbench Server 1.0 state completely. This is a simple and "brute force"
+# sequence with few "blade guards", so be careful! (You have been warned!)
+#
+# 1. Delete the Elasticsearch indices and templates using the configured
+#    prefix value;
+# 2. Delete the configured PostgreSQL database;
+# 3. Delete the ARCHIVE, INCOMING, and RESULTS file system trees
+#
+# NOTE: The Pbench server should be *stopped* previously as root, and this
+# script should be run under `su --login pbench`.
```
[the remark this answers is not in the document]
dbutenhof: The real point was intended to be that, because of the way we have the Pbench Server installation set up, you should run *this* script under the `pbench` account, which means I can't realistically automate the server shutdown/restart as *that* needs to be under `root`. Clearly I missed the target here.

server/bin/pbench-server-erase.sh
dbutenhof: I was trying to be excessively paranoid in order to protect against either bad values in the config file, or a messed up environment that breaks `pbench-config` (which is actually pretty easy to do with bad `PYTHONPATH`). If I get values that don't look like directories, I don't want to screw things up.

dbutenhof:
> My intent with the canned Server was to have everything contained inside the pod. The script here cleans up Elasticsearch and PostgreSQL, each of which are inside containers in the pod; it also cleans up the archived and upacked tarballs, which are stored in the pod (volatile) file system. So, I don't see what this script covers which isn't covered by restarting the pod...what did I miss?

Except sometimes in early developer testing we'll never be running fully self-contained canned Pbench. We'll be deploying against *managed* PostgreSQL, Elasticsearch, and eventually Ceph object store instances. I expect that we'll want our long-rumored and only occasionally attempted "staging server" to work in this mode, too; and we're going to be resetting it periodically until we get to the point of supporting a real upgrade process.

Your canned server is a great idea for throwaway developer testing, and possibly even for functional testing (presuming all of the pod pieces don't overwhelm the Jenkins executor contexts, or can be hosted elsewhere through scripting); honestly I've been too lazy to work my way through all the twists and turns of the scripts and COPR/Quay infrastructure to try it for myself. One of these days. 😆

webbnh (approved):
👍

server/bin/pbench-server-erase.sh
webbnh: I think you almost succeeded...but, for top marks, you need to do the check and exit before touching Elasticsearch and PostgreSQL....

I don't think you need to be paranoid here:  if these directories don't exist, then `rm` will skip over them...yeah, the check enables you to detect a bad or partially-damaged configuration, but, given that we're trying to clean things out here, I don't think aborting in that case is a feature.  (If the configuration is actually bad, then the `rm` is unlikely to find anything to delete.)

portante (approved):

PR #2900, opened 2022-06-15
Work around for rpmlint incompatibilities

portante:
Starting with `rpmlint` 2.0.0, the `--file` option was changed to `--rpmlintrc`.  The first encounter with this version of `rpmlint` came with Fedora 36, but will likely show up soon in other distributions.

To work around this change, we take advantage of two behaviors of `rpmlint`, one from the pre-2.0.0 version and one from the new version.

For pre-2.0.0, we set the `XDG_CONFIG_HOME` to `${PBENCH_TOP}/utils` so that our `rpmlint` file containing the configuration we want is found.

For 2.0.0 and later, we copy our `rpmlint` file to a `pbench-common.rpmlintrc` along side the build location of the .spec file.

### DIFF
--- a/utils/rpm.mk
+++ b/utils/rpm.mk
@@ -42,7 +42,8 @@ srpm: spec patches tarball
 spec: rpm-dirs ${prog}.spec.j2
 	if [ -e ./seqno ] ;then expr ${seqno} + 1 > ./seqno ;fi
 	jinja2 ${prog}.spec.j2 -D version=${VERSION} -D gdist=g${sha1} -D seqno=${seqno} > ${RPMSPEC}/${prog}.spec
-	rpmlint --rpmlintrc ${PBENCHTOP}/utils/rpmlint ${RPMSPEC}/${prog}.spec
+	cp ${PBENCHTOP}/utils/rpmlint ${RPMSPEC}/pbench-common.rpmlintrc
+	XDG_CONFIG_HOME=${PBENCHTOP}/utils rpmlint ${RPMSPEC}/${prog}.spec
 
 .PHONY: patches
 patches: rpm-dirs


agent/rpm/pbench-agent.rpmlintrc
```diff
@@ -0,0 +1 @@
+../../utils/rpmlint
```
dbutenhof, line 1: Are these really necessary if `rpmlint` is looking for the config file at `${PBENCHTOP}/utils` already?? Or is there something weird(er)(er) about this `XDG_CONFIG_HOME`???

dbutenhof (dismissed):
Cool; if this makes both work, we're back in operation on non-F36 and maybe F36 still works, too?

agent/rpm/pbench-agent.rpmlintrc
portante: The new version of `rpmlint` looks for `${XDG_CONFIG_HOME}/rpmlint/*toml`, which it won't find because we have `${XDG_CONFIG_HOME}/rpmlint` as the configuration file for the previous version.

So this symlinks are auto-discovered by `rpmlint` in the directory in which it is invoked.

webbnh: In the [old version](https://linux.die.net/man/1/rpmlint), it looks for the config file at `$XDG_CONFIG_HOME/rpmlint`.

In the [new version](https://github.com/rpm-software-management/rpmlint#configuration) the config file can come from a number of places, one being any `*.rpmlintrc` or `*-rpmlintrc` located in the same directory as the specfile. (The best practice apparently is to use the name `$PACKAGE_NAME.rpmlintrc`.)  (This file uses syntax compatible with the old version.)

Given that the old version doesn't read TOML format, I think Peter's approach is probably as good as it gets.

webbnh (approved):
👍

portante:
See also https://github.com/rpm-software-management/rpmlint/issues/887.

portante (commented):
Okay, after reviewing the behavior, I have simplified this to a one file change commit.

dbutenhof (approved):

webbnh (approved):
👍

ndokos (approved):
I haven't tried it but looks reasonable to me.

PR #2901, opened 2022-06-17
Fix agent ansible and  python deps

ndokos:
This effectively reverts PR #2893 (and its backport to b0.71 #2896).

Simplify the python dependencies as much as possible:

- RHEL7, RHEL9, CentOS-Stream-9 and the Fedoras have a `python3'
package available from the standard distro repos. The actual version
varies by distro: 3.6 for RHEL7, 3.9.10 for RHEL9/CentOS-Stream-9,
3.10.0 for Fedora 35, 3.10.4 for Fedora 36.

- RHEL8 and CentOS-Stream-8 have python36, python38 and python39
packages available through the module system.

We use the generic `python3` package on most distros, and standardize
on `python39` for RHEL8/CentOS-Stream-8. That is mostly treated using
a jinja2 loop and a (new) `target.json` file that generates the separate
conditionals for RHEL/CentOS Stream 8 on one hand and the rest on the
other.

Unfortunately, I had to treat RHEL7 separately from the rest because
it is missing a python cffi package that all the others have. I haven't
found a clean way to deal with that in jinja2, so it is an explicit
RPM conditional in the spec file.

Also, the Fedoras are treated separately, at least for now.

I am testing by setting the release in subscription-manager now
(wherever possible: there is no 7.9 release which I don't understand -
this needs investigation - and there is no 9.1 release yet which is
understandable).

However, that means that RHEL 8.4 and 8.5 have an `ansible` package
and 8.6 has an `ansible-core` package available. I've hard-coded
`ansible-core` in the spec file (except on RHEL7) which breaks the
installation on 8.4 and 8.5. I don't know how to deal with this problem.

Modify the base dockerfile to enable python39.  Since we standardized on
python39 for CentOS-Stream-8, we enable that module here and stop
worrying about python36.

### DIFF
--- a/agent/rpm/pbench-agent.spec.j2
+++ b/agent/rpm/pbench-agent.spec.j2
@@ -8,60 +8,92 @@ URL:            https://github.com/distributed-system-analysis/pbench
 Source0:        pbench-agent-%{version}.tar.gz
 Buildarch:      noarch
 
-# EPEL provides ansible (a curated set of roles with a dependency on ansible-core)
-# on RHEL 8 and RHEL9. Only CentOS-Stream seems to not have an ansible package
-# available.
 
-%if 0%{?centos} >= 8
-Requires:  ansible-core
+# RHEL7, RHEL9 (and CentOS-Stream-9) and the Fedoras provide a python3 package
+# through the standard distro repos.
+# RHEL8 (and CentOS-Stream-8) provide python36, python38 and python39 as modules.
+# We stick with python39 for them for now.
+# N.B. The condition catches both RHEL8 and CentOS-Stream-8)
+%if 0%{?rhel} == 8
+
+Requires:  python39
+# This is used by the shebang processor
+%define __python3 /usr/bin/python3.9
+
 %else
-Requires:  ansible
+
+Requires:  python3
+
 %endif
 
+
+# RHEL7 is a special case: it does not conform to the (slightly) more
+# general pattern of the rest
 
 %if 0%{?rhel} == 7
-Requires:  python3, python3-pip
-%endif
 
-# This condition will be true on a CentOS-Stream-8 system as well
-%if 0%{?rhel} == 8
-Requires:  python36, python3-pip
 # RPMs for modules in requirements.txt
-Requires:  python3-cffi, python3-click, python3-requests
-# RPMs for module dependencies
-Requires:  python3-docutils, python3-psutil
-%endif
+Requires: python3-pip, python3-requests
+#RPMS for module dependencies
+Requires: python3-psutil
 
-# This condition will be true on a CentOS-Stream-9 system as well
-%if 0%{?rhel} == 9
-Requires:  python3-pip
+# RHEL7 also does not define __python3 - we need it for installing
+# requirements through pip so we define it here
+%define __python3 /usr/bin/python3
+
+%else
+
+{# Jinja2 template for python dependencies - see the targets.json file for the input data #}
+{%for tname, tcond in targets.items() -%}
+
+%if {{ tcond }}
 # RPMs for modules in requirements.txt
-Requires:  python3-cffi, python3-requests
-# RPMs for module dependencies
-Requires:  python3-docutils, python3-psutil
+Requires: {{ tname }}-pip, {{ tname }}-cffi, {{ tname }}-requests
+#RPMS for module dependencies
+Requires: {{ tname }}-psutil
 %endif
 
+{% endfor %}
+{# end of template #}
 
+%endif
+
+# docutils is not available for RHEL7 - it is also *only* available as
+# `python3-docutil's on everything else, which is not good for
+# RHEL8/CentOS Stream 8 where we would want `python39-docutils', but
+# this does not exist. We need to handle it specially: check for
+# RHEL9/CentOS Stream 9 OR Fedora
+
+%if 0%{?rhel} > 8 || 0%{?fedora} != 0
+Requires: python3-docutils
+%endif
+
+# additional packages that Fedora builds but the RHELs don't
 %if 0%{?fedora} != 0
-Requires:  python3, python3-pip
 # RPMs for modules in requirements.txt
-Requires:  python3-bottle, python3-cffi, python3-click, python3-daemon
-Requires:  python3-jinja2, python3-redis, python3-requests, python3-sh
+Requires:  python3-bottle, python3-click, python3-daemon
+Requires:  python3-jinja2, python3-redis, python3-sh
+%endif
 
 %if 0%{?fedora} >= 36
 Requires: python3-ifaddr
 %endif
 
-# RPMs for module dependencies
-Requires:  python3-psutil
-%endif
-
+# Common requirements
 Requires:  perl, perl-Data-UUID, perl-JSON, perl-JSON-XS
 Requires:  perl-Time-HiRes
 
 Requires:  bc, bzip2, hostname, iproute, iputils, net-tools, numactl
 Requires:  openssh-clients, openssh-server, procps-ng, psmisc, redis
 Requires:  rpmdevtools, rsync, screen, sos, tar, xz
+
+# The condition here should be 7 OR 8.4 OR 8.5 but we don't have a way to deal with minor versions.
+# The resulting RPM will NOT work on RHEL8.[45].
+%if 0%{?rhel} == 7
+Requires:  ansible
+%else
+Requires:  ansible-core
+%endif
 
 Obsoletes: pbench <= 0.34
 Conflicts: pbench <= 0.34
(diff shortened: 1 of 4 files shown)


dbutenhof (approved):

agent/rpm/pbench-agent.spec.j2
```diff
@@ -8,42 +8,39 @@ URL:            https://github.com/distributed-system-analysis/pbench
...
 
 %if 0%{?rhel} == 7
-Requires:  python3, python3-pip
+Requires:  ansible
+%else
+Requires:  ansible-core
 %endif
 
-# This condition will be true on a CentOS-Stream-8 system as well
+# RHEL7, RHEL9 (and CentOS-Stream-9) and the Fedoras provide a python3 package
```
portante, line 18: We don't want to work with anything else by Python 3.8, with the exception of Python 3.whatever that is available on RHEL 7.

portante (changes requested):

agent/rpm/pbench-agent.spec.j2
ndokos: On F35 and F36, doing `dnf module list` shows various versions of  perl, postgresql etc. but there is no mention of a different python.

webbnh (approved):
👍

agent/rpm/pbench-agent.spec.j2
portante: On my Fedora 36 box I see:

```
python3.11.x86_64                                         3.11.0~b1-2.fc36                      @updates
python3.11-devel.x86_64                                   3.11.0~b1-2.fc36                      @updates
python3.11-libs.x86_64                                    3.11.0~b1-2.fc36                      @updates
python3.11-tkinter.x86_64                                 3.11.0~b1-2.fc36                      @updates
python3.5.x86_64                                          3.5.10-7.fc34                         @updates
python3.6.x86_64                                          3.6.15-8.fc36                         @fedora
python3.7.x86_64                                          3.7.13-1.fc36                         @fedora
python3.8.x86_64                                          3.8.13-2.fc36                         @updates
python3.9.x86_64                                          3.9.13-1.fc36                         @updates
```

webbnh: Right, but were all of those there with F36 "out of the box", or did you install them at various times (perhaps pre-F36), e.g., when you were playing with Tox environments?

portante: Why does that matter for the Fedoras?  If they are available, but not installed, and a `pbench-agent` RPM has a `Requires: python3.8` on it, wouldn't they be installed along side the `pbench-agent` RPM?

webbnh: Yes..._if it is available_, then, yeah, I would expect `python3.8` to be installed as a precursor.

So, why does it seem to be available to you but not to Nick?

ndokos: What does your `/etc/yum.repos.d` look like? I have a fedora-modular.repo that is enabled, amd I just did a `dnf update` just in case, but I still can't see *any* python modules: `dnf module list` shows a bunch of other stuff, but nothing about python.  I cannot understand why the two setups are so different.

portante: That is because they are not modules.

```
$ sudo dnf list python3.\*.x86_64 | less -S
Last metadata expiration check: 1:22:15 ago on Sat 18 Jun 2022 06:06:11 AM EDT.
Installed Packages
python3.11.x86_64                       3.11.0~b1-2.fc36                @updates
python3.11-devel.x86_64                 3.11.0~b1-2.fc36                @updates
python3.11-libs.x86_64                  3.11.0~b1-2.fc36                @updates
python3.11-tkinter.x86_64               3.11.0~b1-2.fc36                @updates
python3.5.x86_64                        3.5.10-7.fc34                   @updates
python3.6.x86_64                        3.6.15-8.fc36                   @fedora
python3.7.x86_64                        3.7.13-1.fc36                   @fedora
python3.8.x86_64                        3.8.13-2.fc36                   @updates
python3.9.x86_64                        3.9.13-1.fc36                   @updates
Available Packages
python3.11.x86_64                       3.11.0~b3-1.fc36                updates
python3.11-debug.x86_64                 3.11.0~b3-1.fc36                updates
python3.11-devel.x86_64                 3.11.0~b3-1.fc36                updates
python3.11-idle.x86_64                  3.11.0~b3-1.fc36                updates
python3.11-libs.x86_64                  3.11.0~b3-1.fc36                updates
python3.11-test.x86_64                  3.11.0~b3-1.fc36                updates
python3.11-tkinter.x86_64               3.11.0~b3-1.fc36                updates
```

portante: Further, the existing `python3*` RPMs which provide our various python module dependency are only built for Python 3.10.  Notice that in the previous list `python3.10.*` does not exist.

If we install `python3.8`, and use that python in our invocations (that is another problem), then we'll still have to `pip3` install all our module dependencies.

We are going to have to support the distribution's default python version if we are going to avoid as many pip installs as possible.

agent/rpm/pbench-agent.spec.j2
```diff
@@ -8,61 +8,78 @@ URL:            https://github.com/distributed-system-analysis/pbench
...
+# Common requirements
 Requires:  perl, perl-Data-UUID, perl-JSON, perl-JSON-XS
 Requires:  perl-Time-HiRes
 
 Requires:  bc, bzip2, hostname, iproute, iputils, net-tools, numactl
 Requires:  openssh-clients, openssh-server, procps-ng, psmisc, redis
 Requires:  rpmdevtools, rsync, screen, sos, tar, xz
 
+# The condition here should be 7 OR 8.4 OR 8.5 but we don't have a way to deal with minor versions.
+# The resulting RPM will NOT work on RHEL8.[45].
```
dbutenhof, line 76: Sounds like we need to remember to call this out in the release notes. Hopefully it's OK to require 8.6 now? 😒

agent/rpm/targets.json
```diff
@@ -0,0 +1,11 @@
...
+    "targets": {
+        "python3": {
+            "condition": "0%{?rhel} != 8"
+        },
+        "python39": {
+            "condition": "0%{?rhel} == 8"
+        }
+    }
+}
+
```
dbutenhof, line 11: Is there an intention to add more than `"condition"` in each object, or should this be simplified to `"python3": "0%{?rhel} != 0"` & etc?

agent/rpm/pbench-agent.spec.j2
ndokos: I hope so. No 8.[45], no F36, no 9.1 - the distros are dropping like flies :-)

agent/rpm/targets.json
ndokos: At this point, I don't think we need to add anything, but I don't know for sure.

ndokos: I simplified it as you suggested. We can always complicate it later if necessary.

dbutenhof (approved):

portante (commented):
Posting some previously unposted comments.

agent/rpm/pbench-agent.spec.j2
portante: Resolving since we are now moving to Python 3.9 for RHEL 8 to line up with Python 3.9 being used in all other distributions except RHEL 7 (3.6) and Fedora 36 (3.10 default, but 3.9 available).

agent/rpm/pbench-agent.spec.j2
```diff
@@ -61,7 +74,7 @@ Requires:  perl-Time-HiRes
 
 Requires:  bc, bzip2, hostname, iproute, iputils, net-tools, numactl
 Requires:  openssh-clients, openssh-server, procps-ng, psmisc, redis
-Requires:  rpmdevtools, rsync, screen, sos, tar, xz
```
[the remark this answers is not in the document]
portante: The most recent PR update resolved this issue.

portante (approved):

agent/rpm/pbench-agent.spec.j2
```diff
@@ -98,8 +111,8 @@ if pip3 show configtools > /dev/null 2>&1 ;then pip3 uninstall -y configtools ;f
 %post
 
 # Install python dependencies
-python3 -m pip install --upgrade pip 2>&1 > /%{installdir}/pip3-upgrade.log
-python3 -m pip --no-cache-dir install --prefix=/%{installdir} -r /%{installdir}/requirements.txt > /%{installdir}/pip3-install.log 2>&1
+%{__python3} -m pip install --upgrade pip 2>&1 > /%{installdir}/pip3-upgrade.log
```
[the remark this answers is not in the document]
portante: This was resolved in the most recent commit updates.

agent/rpm/targets.json
webbnh: @ndokos, using a loop to iterate over mutually-exclusive targets seems like a fairly heavy mechanism for what it accomplishes.  Unless we have some future expansion in mind, I think we would be better served by a straightforward conditional definition of `tname` to either `python3`/`python39`.

I assume I'll be modifying this file soon, would you be OK with me making that change?

webbnh: Looking closer, I think this definition could be made along side the definition of `__python` which would be pretty simple.

PR #2905, opened 2022-06-17
Fix pbench-register-tool-set argument handling - backport of #2904

### LINKED ISSUE #2903: [BUG] pbench-register-tool-set does not handle a named toolset argument

### Description

`pbench-register-tool-set` does not handle a named toolset argument at all. It issues a deprecation warning and proceeds to register the default tool set (with or without a `--group` option):

```
# pbench-register-tool-set light
[warn][2022-06-17T04:01:59.291939863] pbench-register-tool-set: the 'default' tool set is deprecated and will be removed in the Pbench Agent v1.0 future release, please use a named tool set argument from one of: heavy, legacy, light, medium
"iostat" tool is now registered for host "dhcp31-1.perf.lab.eng.bos.redhat.com" in group "default"
"mpstat" tool is now registered for host "dhcp31-1.perf.lab.eng.bos.redhat.com" in group "default"
"pidstat" tool is now registered for host "dhcp31-1.perf.lab.eng.bos.redhat.com" in group "default"
"proc-interrupts" tool is now registered for host "dhcp31-1.perf.lab.eng.bos.redhat.com" in group "default"
"proc-vmstat" tool is now registered for host "dhcp31-1.perf.lab.eng.bos.redhat.com" in group "default"
"sar" tool is now registered for host "dhcp31-1.perf.lab.eng.bos.redhat.com" in group "default"
"turbostat" tool is now registered for host "dhcp31-1.perf.lab.eng.bos.redhat.com" in group "default"
"perf" tool is now registered for host "dhcp31-1.perf.lab.eng.bos.redhat.com" in group "default"
```

@webbnh and I observed that behavior during the RPM/container testing session.

### To Reproduce

Install the current v0.71 RPM and run ` pbench-register-tool-set light`.

### Actual Results

It issues a deprecation warning about the default tool set, ignores the argument and registers the tools of the default tool set.

### Expected Results

No deprecation warning in this case and registration of the tool set that is named in the argument.

### Additional information

_No response_

ndokos:
Fixes #2903.

Back-port of #2904 (6ba02ca28).

Invoking pbench-register-tool-set with a named toolset argument, e.g.

    pbench-register-tool-set light

ignores the named argument, and substitutes the default tool set, resulting in a deprecation warning and registration of the default tool set.

The option-handling code does one extra shift of the argument list falling off the end. The fix just eliminates the extra shift.

Also add a simple unit test for this case.

### DIFF
--- a/agent/util-scripts/gold/pbench-register-tool-set/test-11.18.txt
+++ b/agent/util-scripts/gold/pbench-register-tool-set/test-11.18.txt
@@ -1,0 +1,17 @@
++++ Running test-11.18 pbench-register-tool-set light
+"vmstat" tool is now registered for host "testhost.example.com" in group "default"
+--- Finished test-11.18 pbench-register-tool-set (status=0)
++++ pbench tree state
+/var/tmp/pbench-test-utils/pbench
+/var/tmp/pbench-test-utils/pbench/pbench.log
+/var/tmp/pbench-test-utils/pbench/tmp
+/var/tmp/pbench-test-utils/pbench/tools-v1-default
+/var/tmp/pbench-test-utils/pbench/tools-v1-default/testhost.example.com
+/var/tmp/pbench-test-utils/pbench/tools-v1-default/testhost.example.com/vmstat
+=== /var/tmp/pbench-test-utils/pbench/tools-v1-default/testhost.example.com/vmstat:
+--interval=3
+--- pbench tree state
++++ pbench.log file contents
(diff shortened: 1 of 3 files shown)


dbutenhof (approved):
Although they behave exactly the same (since `getopt` always returns `--` before the first non-option value, which I hadn't previously known), I wonder if we should also be explicitly testing `pbench-register-tool-set -- light`? But ... probably not a big deal.

portante:
Set to `Draft` mode until #2904 is resolved and merged.

portante (approved):

PR #2906, opened 2022-06-17
Clarify "tarball" vs "dataset" in FileTree

dbutenhof:
PBENCH-769

With the recent change to make tarball MD5 the authoritative "resource id" for datasets, the ambiguity in `FileTree` classes between "tarball" and "dataset" becomes more important. A planned change to allow altering the dataset resource name for display will compound the problem since tarball names are immutable but will no longer match the dataset name.

This change attempts to more clearly separate the two concepts, while redefining the primary access into the file tree to use the resource ID of a dataset, which is the tarball MD5 value and therefore easily accessible in both models.


webbnh (approved):
👍

npalaska (approved):
Looks good

PR #2909, opened 2022-06-17
Improve endpoints API for multiple URI params

dbutenhof:
PBENCH-764

I always wanted to expand `/api/v1/endpoints` into a server API schema, and with the introduction of multiple URI parameters in the inventory API, I've decided to make an attempt.

Previously I stripped all Flask URI parameter tags from the URI, leaving a "/" character, and substituting "_" for "/" in the API name. A trailing "/" works for a single parameter, but "//" isn't very useful for two, nor is "datasets_inventory_" a great name.

The familiar "api" dict values now give only a single trailing "/"; but I'm also adding a new "uri" dict which ties the API names to a "schema" which defines a substitute-able template URI and also provides parameter definitions with name and type. For example,

```
"datasets_inventory": {
    "template": "http://host/api/v1/datasets/inventory/{dataset}/{path}",
    "params": [
        {"name": "dataset", "type": "string"},
        {"name": "path", "type": "path"}
    ]
}
```

It's worth noting that for current APIs, `endpoints?.api?.datasets_publish` would look just as it did before; however, `endpoints?.uri?.datasets_publish` will show a template of `http://host/api/v1/datasets/publish/{dataset}` with a param of `{"name": "dataset", "type": "string"}`.

(And, FYI, it *is* possible to define new Flask converters, so, in theory, we could define a `dataset` converter and a `username` converter, which would at least provide a bit more detailed documentation even if we only validated them as strings; that's way out of scope for this PR.)


lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -114,35 +148,33 @@ def get(self):
...
-                # adjustment later if we add any. (E.g., something like
-                # "/api/v1/foo/<string:name>/detail/<string:param>")
+                simplified = self.param_template.sub(r"/{\g<name>}", url)
+                template: Dict[str, Any] = {"template": urljoin(host, simplified)}
+                params = []
+                for match in self.param_template.finditer(url):
+                    params.append(
+                        {"name": match.group("name"), "type": match.group("type")}
+                    )
+                template["params"] = params
```
webbnh, line 158: Alternately:
```
                params = []
                for match in self.param_template.finditer(url):
                    params.append(
                        {"name": match.group("name"), "type": match.group("type")}
                    )
                template: Dict[str, Any] = {"template": urljoin(host, simplified), "params": params}
```
or, even,
```
                template: Dict[str, Any] = {
                    "template": urljoin(host, simplified),
                    "params": [
                        {"name": match.group("name"), "type": match.group("type")}
                        for match in self.param_template.finditer(url):
                    ]
                }
```
Although, I would pull for something more like this:
```
                template: Dict[str, Any] = {
                    "template": urljoin(host, simplified),
                    "params": {
                        match.group("name"): {"type": match.group("type")}
                        for match in self.param_template.finditer(url):
                    }
                }
```
🙂

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -114,35 +148,33 @@ def get(self):
...
-                    path = path[:-1]
                 path = path.replace("/", "_")
+
+                # Terminate the base URL with a slash if there were parameters
+                # removed. The client can just append a single parameter value.
+                # If there are two parameters, it'll have to add the separating
+                # "/" or use the template string: for example, with
+                # template.replace('{name1}', value1)
+                if params:
+                    url += "/"
```
webbnh, line 169: Is this of sufficient value to justify the complication?  That is, instead of doing this, why not just state that, if the client adds one or more parameters to the URI, it has to add the separating slashes, including for the first parameter?

Or, are we using the presence of the trailing slash for more than just making it easy to append the first parameter?

lib/pbench/test/unit/server/test_endpoint_configure.py
```diff
@@ -41,10 +41,6 @@ def check_config(self, client, server_config, host, my_headers={}):
         expected_results = {
             "identification": f"Pbench server {server_config.COMMIT_ID}",
             "api": {
-                # API endpoints with trailing Flask parameters are marked with
-                # a trailing "/" here; for example, /datasets/mappings/
-                # corresponds to /datasets/mappings/<string:dataset_view>";
-                # see endpoint_configure.py for more detail.
```
webbnh, line 47: You really want to remove this comment?  It seems like a useful thing to point out (and, it's still true after this change, right?).

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -114,35 +148,33 @@ def get(self):
...
-                # E.g, "/api/v1/controllers/list" yields:
-                #     "controllers_list": "/api/v1/controllers/list"
-                #
-                # while "/api/v1/users/<string:username>" yields:
-                #     "users": "/api/v1/users/"
-                #
-                # TODO: This won't work right with embedded template strings,
-                # which we're not currently using anywhere; but it'll require
-                # adjustment later if we add any. (E.g., something like
-                # "/api/v1/foo/<string:name>/detail/<string:param>")
```
webbnh, line 134: It seems premature to remove this warning: the proposed approach still doesn't work if there is an embedded parameter in the middle of the URI (as in the example at old-line 134).

Either that, or we need an approach which truncates the API at the first parameter, instead of just removing the parameters.

webbnh (dismissed):
It seems like the parameters would be better represented by a mapping:
```
"datasets_inventory": {
    "template": "http://host/api/v1/datasets/inventory/{dataset}/{path}",
    "params": {
        "dataset": {"type": "string"},
        "path": {"type": "path"}
    ]
}
```
That is, each parameter has a name, so we can use that as a key, instead of including it as a field in the objects in the list, and then the consumer can either iterate over the keys (instead of iterating over the list) or it can look up the entries directly (instead of searching for them).

lib/pbench/server/api/resources/endpoint_configure.py
dbutenhof: I put that back in at the last minute for compatibility. It does have the small advantage of being able to do `endpoints.api.users + username` instead of worrying about the slash. I'd prefer to get rid of the `"api"` object entirely and just use the templated `"uri"` object, but I wanted the alternatives side by side to get feedback from @MVarshini @shubham-html-css-js.

lib/pbench/test/unit/server/test_endpoint_configure.py
dbutenhof: Yeah, I pulled the trailing `/` back in at the last moment, after having justified the removal of this comment... and I didn't think to restore it. 😆

lib/pbench/server/api/resources/endpoint_configure.py
```diff
@@ -71,17 +71,17 @@ def get(self):
         underscores.
 
         The "api" object contains a key for each API name, where the value is a
-        simplified URI omitting URI parameters. We append a single trailing "/"
-        to signal that the API includes URI parameters that must be provided.
-        With a single URI parameter, the value can be appended to this string.
+        simplified URI omitting URI parameters. The client must either know the
+        require parameters and order, and connect them to the "api" value
```
webbnh, line 75: s/require/required/

webbnh (approved):
👍

MVarshini (approved):
:+1: 
It will be great if we don't have any `/` at the last. While making the request in the dashboard it can be addded

PR #2910, opened 2022-06-17
Ensure `fio` arguments are used with `bash -c`

portante:
The execution of a fio "client" was failing on the local host because "bash -c ..." receives multiple, separate arguments.  Since `bash -c` only considers its first argument, this results in an execution of only `fio` without the remaining arguments.

The solution is to create a `fio-client` driver script which accepts two arguments, the directory to execute `fio` from and the port to use with `--server`.  The invocation of `bash -c` can be properly quoted in the new context.


portante:
Supersedes PR #2907.

webbnh (approved):
👍

ndokos (approved):

ldoktor (approved):
Thanks, this is works as well (tested local execution only). It increases the learning curve when trying to find out why things don't work, but in the end is clearer than evals... Would you please rebuild the test rpm packages once this gets merged?

ldoktor (approved):
Forgot to tick the Approve...

PR #2911, opened 2022-06-20
PbenchConfig unit-test Test Coverage

riya-17:
PBENCH-213

Increases the test coverage for the PbenchConfig class.
Removed unnecessary file usage along the script.

### DIFF
--- a/lib/pbench/test/unit/common/test_config.py
+++ b/lib/pbench/test/unit/common/test_config.py
@@ -87,11 +87,17 @@ class TestPbenchConfig:
         assert config.TZ == "UTC"
 
     @pytest.mark.parametrize(
-        "config_value, expected_error_msg",
+        "config_value, error_type, expected_error_msg",
         [
             (
                 {"logging": {"logger_type": "hostport"}},
+                BadConfig,
                 "No option 'logger_host' in section: 'logging'",
+            ),
+            (
+                {"logging": {"logger_type": "hostport", "logger_host": ""}},
+                ValueError,
+                "ValueError('logger_host', 'logging')",
             ),
             (
                 {
@@ -100,27 +106,8 @@ class TestPbenchConfig:
                         "logger_host": "logger.example.com",
                     }
                 },
+                BadConfig,
                 "No option 'logger_port' in section: 'logging'",
-            ),
-        ],
-    )
-    def test_logger_hostport_BadConfig(
-        self, config_value, expected_error_msg, monkeypatch
-    ):
-        """Show that PbenchConfig raises 'BadConfig' when the
-        configuration does not contain required option."""
-
-        self.logger_test_setup(monkeypatch, config_value)
-        with pytest.raises(BadConfig) as exc:
-            PbenchConfig("pbench.cfg")
-        assert expected_error_msg in str(exc.value)
-
-    @pytest.mark.parametrize(
-        "config_value, expected_error_msg",
-        [
-            (
-                {"logging": {"logger_type": "hostport", "logger_host": ""}},
-                "ValueError('logger_host', 'logging')",
             ),
             (
                 {
@@ -130,18 +117,20 @@ class TestPbenchConfig:
                         "logger_port": "",
                     }
                 },
+                ValueError,
                 "ValueError('logger_port', 'logging')",
             ),
         ],
     )
-    def test_logger_hostport_ValueError(
-        self, config_value, expected_error_msg, monkeypatch
+    def test_bad_logger_hostport(
+        self, config_value, error_type, expected_error_msg, monkeypatch
     ):
-        """Show that PbenchConfig raises 'ValueError' when the
-        option contains an empty value."""
+        """Shows that PbenchConfig raises 'BadConfig' and 'ValueError' when
+        the configuration does not contain required option or an empty
+        value respectively."""
 
         self.logger_test_setup(monkeypatch, config_value)
-        with pytest.raises(ValueError) as exc:
+        with pytest.raises(error_type) as exc:
             PbenchConfig("pbench.cfg")
         assert expected_error_msg in str(exc)
 
--- a/lib/pbench/test/unit/common/test_config.py
+++ b/lib/pbench/test/unit/common/test_config.py
@@ -1,68 +1,173 @@
 """Test PbenchConfig class and objects
 """
-
 import pytest
 
-from pathlib import Path
+from configparser import ConfigParser, NoOptionError, NoSectionError
+from typing import List, Dict, Any
 
 from pbench import PbenchConfig
+from pbench.common import configtools
 from pbench.common.exceptions import BadConfig
 
 
-_config_path_prefix = Path("lib/pbench/test/unit/common/config")
+class MockedConfigParser:
+    """Mocked ConfigParser"""
+
+    def __init__(self):
+        self._sections = {}
+
+    def get(self, section: str, option: str, fallback=None) -> str:
+        """returns value of the given `section` and `option`"""
+
+        if section not in self._sections:
+            raise NoSectionError(section)
+        if option not in self._sections[section]:
+            raise NoOptionError(option, section)
+        value = self._sections[section][option]
+        if not value:
+            if not fallback:
+                raise ValueError(option, section)
+            else:
+                return fallback
+        return value
 
 
 class TestPbenchConfig:
-    def test_empty_config(self):
-        config = PbenchConfig(_config_path_prefix / "pbench.cfg")
-        assert config.TZ == "UTC", f"Unexpected TZ value, {config.TZ!r}"
-        assert (
-            config.log_fmt is None
-        ), f"Unexpected log format value, {config.log_fmt!r}"
-        assert (
-            config.default_logging_level == "INFO"
-        ), f"Unexpected default logging level, {config.default_logging_level!r}"
-        assert (
-            config.log_using_caller_directory is False
-        ), f"Unexpected 'log using caller directory' boolean, {config.log_using_caller_directory!r}"
-        assert config.log_dir is None, f"Unexpected log directory, {config.log_dir!r}"
-        assert (
-            config.logger_type == "devlog"
-        ), f"Unexpected logger type, {config.logger_type!r}"
-        with pytest.raises(AttributeError):
-            print(f"{config.logger_host!r}")
-        with pytest.raises(AttributeError):
-            print(f"{config.logger_port!r}")
-        assert "42" == config.get(
-            "other", "foobar"
-        ), "Failed to fetch 'foobar' from 'DEFAULT' section"
-        assert "43" == config.get(
-            "other", "barfoo"
-        ), "Failed to fetch 'barfoo' from 'other' section"
-        assert isinstance(
-            config.files, list
-        ), f"Unexpected object class for 'files', {config.files.__class__!r}"
+    @staticmethod
+    def logger_test_setup(monkeypatch, sections: Dict[str, Any]):
+        def mock_ConfigParser_init(self):
+            self._sections = sections
 
-    def test_log_dir_provided(self):
-        config = PbenchConfig(_config_path_prefix / "logdir.cfg")
-        assert (
-            config.log_dir == "/srv/log/directory"
-        ), f"Unexpected log directory, {config.log_dir!r}"
+        monkeypatch.setattr(configtools, "file_list", TestPbenchConfig.mock_file_list)
+        monkeypatch.setattr(ConfigParser, "__init__", mock_ConfigParser_init)
+        monkeypatch.setattr(ConfigParser, "get", MockedConfigParser.get)
 
-    def test_logger_type_provided(self):
-        config = PbenchConfig(_config_path_prefix / "hostport.cfg")
-        assert (
-            config.logger_type == "hostport"
-        ), f"Unexpected logger type, {config.logger_type!r}"
-        assert (
-            config.logger_host == "logger.example.com"
-        ), f"Unexpected logger host value, {config.logger_host!r}"
-        assert (
-            config.logger_port == "42"
-        ), f"Unexpected logger port value, {config.logger_port!r}"
+    def mock_file_list(cfg_name: str) -> List[str]:
+        """file_list mock module"""
+        return [cfg_name]
 
-    def test_logger_type_hostport_missing(self):
-        with pytest.raises(BadConfig):
-            PbenchConfig(_config_path_prefix / "hostport-missing.cfg")
-        with pytest.raises(BadConfig):
-            PbenchConfig(_config_path_prefix / "hostport-missing-port.cfg")
+    def test_default_logger(self, monkeypatch):
+        """Show that PbenchConfig uses default values when NO
+        configuration values are provided for `logging`"""
+
+        self.logger_test_setup(monkeypatch, {})
+        config = PbenchConfig("pbench.cfg")
+        assert config.logger_type == "devlog"
+        assert config.log_dir is None
+        assert config.log_using_caller_directory is False
+        assert config.default_logging_level == "INFO"
+        assert config.log_fmt is None
+        assert config.TZ == "UTC"
+
+    def test_custom_logger(self, monkeypatch):
+        """Show that when configuration values are provided,
+        PbenchConfig uses those values instead of default values"""
+
+        logger_type = "file"
+        log_dir = "/mock/log_dir"
+        logging_level = "Devlog"
+        log_format = "1970-01-01T00:00:42.000000 -- 'message'"
+        self.logger_test_setup(
+            monkeypatch,
+            {
+                "logging": {
+                    "logger_type": logger_type,
+                    "log_dir": log_dir,
+                    "logging_level": logging_level,
+                    "log_format": log_format,
+                }
+            },
+        )
+        config = PbenchConfig("pbench.cfg")
+        assert config.logger_type == logger_type
+        assert config.log_dir == log_dir
+        assert config.log_using_caller_directory is False
+        assert config.default_logging_level == logging_level
+        assert config.log_fmt == log_format
+        assert config.TZ == "UTC"
+
+    @pytest.mark.parametrize(
+        "config_value, expected_error_msg",
+        [
+            (
+                {"logging": {"logger_type": "hostport"}},
+                "No option 'logger_host' in section: 'logging'",
+            ),
+            (
+                {
+                    "logging": {
+                        "logger_type": "hostport",
+                        "logger_host": "logger.example.com",
+                    }
+                },
+                "No option 'logger_port' in section: 'logging'",
+            ),
+        ],
+    )
+    def test_logger_hostport_BadConfig(
+        self, config_value, expected_error_msg, monkeypatch
+    ):
+        """Show that PbenchConfig raises 'BadConfig' when the
+        configuration does not contain required option."""
+
+        self.logger_test_setup(monkeypatch, config_value)
+        with pytest.raises(BadConfig) as exc:
+            PbenchConfig("pbench.cfg")
+        assert expected_error_msg in str(exc.value)
+
+    @pytest.mark.parametrize(
+        "config_value, expected_error_msg",
+        [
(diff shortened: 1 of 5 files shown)


lib/pbench/test/unit/common/test_config.py
```diff
@@ -1,68 +1,252 @@
...
+        monkeypatch.setattr(configtools, "file_list", self.mock_file_list)
+        monkeypatch.setattr(ConfigParser, "__init__", mock_MockedConfigParser.__init__)
+        monkeypatch.setattr(ConfigParser, "get", MockedConfigParser.get)
+        with pytest.raises(BadConfig) as exc:
+            config = PbenchConfig("pbench.cfg")
+            assert config.logger_type == "hostport"
+            assert exc == "No option 'logger_host' in section: 'logging'"
+
+    def test_logger_type_host_value(self, monkeypatch):
+        """verify `host` value of logger_type `hostport`"""
```
dbutenhof, line 76: This should be based more on the docstring from `test_logger_type_hostport` distinguishing that in this case the key is present but has an empty value.

(The test names could make this distinction more clear, as well, although it's best not to make them too long!)

dbutenhof (dismissed):
I think the test names and docstrings could clarify what's being tested better. However, Cobertura says 100% coverage of `PbenchConfig`, so that's a big success, and I generally like your improvements here.

lib/pbench/test/unit/common/test_config.py
```diff
@@ -1,68 +1,252 @@
...
+
+        class mock_MockedConfigParser(MockedConfigParser):
+            def __init__(self):
+                self._sections = {"logging": {"logger_type": "file"}}
+
+        monkeypatch.setattr(configtools, "file_list", self.mock_file_list)
+        monkeypatch.setattr(ConfigParser, "__init__", mock_MockedConfigParser.__init__)
+        monkeypatch.setattr(ConfigParser, "get", MockedConfigParser.get)
+        config = PbenchConfig("pbench.cfg")
+        assert config.logger_type == "file"
```
webbnh, line 58: I invite you to refactor this function like this:
```
    @staticmethod
    def logger_test_setup(sections: Dict[str, Any]):
        def mock_ConfigParser_init(self):
            self._sections = sections

        monkeypatch.setattr(configtools, "file_list", TestPbenchConfig.mock_file_list)
        monkeypatch.setattr(ConfigParser, "__init__", mock_ConfigParser_init)
        monkeypatch.setattr(ConfigParser, "get", MockedConfigParser.get)

    def test_logger(self, monkeypatch):
        """verify logger_type value from config file"""
        logger_type = "file"
        logger_test_setup({"logging": {"logger_type": logger_type}})
        config = PbenchConfig("pbench.cfg")
        assert config.logger_type == logger_type
```
Then, you can refactor `test_default_logger` similarly:
```
    def test_default_logger(self, monkeypatch):
        logger_test_setup({})
        config = PbenchConfig("pbench.cfg")
        assert config.logger_type == "devlog"
```
And then `test_logger_type_host_value`:
```
    def test_logger_type_host_value(self, monkeypatch):
        """verify `host` value of logger_type `hostport`"""
        logger_test_setup({"logging": {"logger_type": "hostport"}}}
        with pytest.raises(BadConfig) as exc:
            config = PbenchConfig("pbench.cfg")
            assert config.logger_type == "hostport"
            assert exc == "No option 'logger_host' in section: 'logging'"
```
And so on....

lib/pbench/test/unit/common/test_config.py
```diff
@@ -1,68 +1,252 @@
...
+            def __init__(self):
+                self._sections = {"logging": {"logger_type": "hostport"}}
+
+        monkeypatch.setattr(configtools, "file_list", self.mock_file_list)
+        monkeypatch.setattr(ConfigParser, "__init__", mock_MockedConfigParser.__init__)
+        monkeypatch.setattr(ConfigParser, "get", MockedConfigParser.get)
+        with pytest.raises(BadConfig) as exc:
+            config = PbenchConfig("pbench.cfg")
+            assert config.logger_type == "hostport"
+            assert exc == "No option 'logger_host' in section: 'logging'"
```
webbnh, line 73: There's a problem here:  the call to `PbenchConfig()` is going to raise the `BadConfig` exception, which means that execution is going to skip over the assertions at lines 72 & 73.

Line 73 needs to be moved out of the `with` block (and, `exc` is an exception-info object, so I think you have to extract the message from it before you compare it).

Line 72 should be removed: since the initialization of the configuration results in an exception, there's no reason to think that its contents have any particular values, so there's no point to checking them.

An analogous comment applies to each of the other `with` blocks.

lib/pbench/test/unit/common/test_config.py
webbnh: As a general rule of thumb, the docstring for a test should say something similar to "Show that when XXXX happens, function YYYY responds with ZZZZ".  So, for this test, it should be something like, "Show that PbenchConfig raises ValueError when the configuration contains an empty string for 'logger_host'."

lib/pbench/test/unit/common/test_config.py
```diff
@@ -1,68 +1,252 @@
...
+                }
+
+        monkeypatch.setattr(configtools, "file_list", self.mock_file_list)
+        monkeypatch.setattr(ConfigParser, "__init__", mock_MockedConfigParser.__init__)
+        monkeypatch.setattr(ConfigParser, "get", MockedConfigParser.get)
+        with pytest.raises(ValueError) as exc:
+            config = PbenchConfig("pbench.cfg")
+            assert config.logger_type == "hostport"
+            assert config.logger_host == "logger.example.com"
+            assert exc == "No option 'logger_port' in section: 'logging'"
```
webbnh, line 130: It strikes me that all of the "hostport" failure test scenarios could be combined into a single, parametrize'd test, with input parameters for the sections, the expected exception, and the expected exception message.

lib/pbench/test/unit/common/test_config.py
```diff
@@ -1,68 +1,252 @@
...
+                    "logging": {
+                        "logger_type": "file",
+                    }
+                }
+
+        monkeypatch.setattr(configtools, "file_list", self.mock_file_list)
+        monkeypatch.setattr(ConfigParser, "__init__", mock_MockedConfigParser.__init__)
+        monkeypatch.setattr(ConfigParser, "get", MockedConfigParser.get)
+        config = PbenchConfig("pbench.cfg")
+        assert config.log_dir is None
```
webbnh, line 168: This test looks to me to be identical to `test_logger()` except for the assertion at line 168.  I recommend that you move the assertion from here to that test and omit this test.

lib/pbench/test/unit/common/test_config.py
```diff
@@ -1,68 +1,252 @@
...
+            def __init__(self):
+                self._sections = {
+                    "logging": {"logger_type": "file", "log_dir": "/mock/log_dir"}
+                }
+
+        monkeypatch.setattr(configtools, "file_list", self.mock_file_list)
+        monkeypatch.setattr(ConfigParser, "__init__", mock_MockedConfigParser.__init__)
+        monkeypatch.setattr(ConfigParser, "get", MockedConfigParser.get)
+        config = PbenchConfig("pbench.cfg")
+        assert config.log_using_caller_directory is False
```
webbnh, line 183: You don't have an assertion in any of the test cases where this value is expected to be `True`...should there be one?  (In `test_logger()`, perhaps?)

lib/pbench/test/unit/common/test_config.py
```diff
@@ -1,68 +1,252 @@
...
+                    "logging": {
+                        "logger_type": "file",
+                    }
+                }
+
+        monkeypatch.setattr(configtools, "file_list", self.mock_file_list)
+        monkeypatch.setattr(ConfigParser, "__init__", mock_MockedConfigParser.__init__)
+        monkeypatch.setattr(ConfigParser, "get", MockedConfigParser.get)
+        config = PbenchConfig("pbench.cfg")
+        assert config.default_logging_level == "INFO"
```
webbnh, line 201: This test looks to me to be identical to `test_logger()` except for the assertion at line 201. I recommend that you move the assertion from here to that test and omit this test.

lib/pbench/test/unit/common/test_config.py
```diff
@@ -1,68 +1,252 @@
...
+                    "logging": {
+                        "logger_type": "file",
+                    }
+                }
+
+        monkeypatch.setattr(configtools, "file_list", self.mock_file_list)
+        monkeypatch.setattr(ConfigParser, "__init__", mock_MockedConfigParser.__init__)
+        monkeypatch.setattr(ConfigParser, "get", MockedConfigParser.get)
+        config = PbenchConfig("pbench.cfg")
+        assert config.log_fmt is None
```
webbnh, line 233: This test looks to me to be identical to `test_logger()` except for the assertion at line 233. I recommend that you move the assertion from here to that test and omit this test.

webbnh (changes requested):
This is great, Riya!

However, there are a couple of `assert`'s which aren't being executed, so that should be fixed.

Also, you have several test scenarios which are identical except for their assertions -- it would be better to combine them and do all the assertions after a single set-up.

Finally, all of the tests seem to have a nearly-identical setup -- it would be better you refactored the setup into a subroutine which all the functions can share.

And, I have a couple of other suggestions, as well.

lib/pbench/test/unit/common/test_config.py
dbutenhof: Good catch ... however even though I somehow missed the missed asserts here, I *did* notice the use of the `ExcInfo` object, and that's actually OK because it stringifies to the embedded exception text exactly as expected here.

(Although, yes, something like `str(exc.value)` would make that more explicit.)

lib/pbench/test/unit/common/test_config.py
riya-17: yeah, maybe

riya-17: this is the default value of this opton and it is hard coded.

lib/pbench/test/unit/common/test_config.py
```diff
@@ -1,68 +1,252 @@
...
+            raise NoSectionError(section)
+        if option not in self._sections[section]:
+            raise NoOptionError(option, section)
+        value = self._sections[section][option]
+        if not value:
+            if not fallback:
+                raise ValueError(option, section)
+            else:
+                return fallback
+        return self._sections[section][option]
```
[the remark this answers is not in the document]
riya-17: yes

lib/pbench/test/unit/common/test_config.py
```diff
@@ -1,68 +1,252 @@
...
-        ), f"Unexpected logger port value, {config.logger_port!r}"
-
-    def test_logger_type_hostport_missing(self):
-        with pytest.raises(BadConfig):
-            PbenchConfig(_config_path_prefix / "hostport-missing.cfg")
-        with pytest.raises(BadConfig):
-            PbenchConfig(_config_path_prefix / "hostport-missing-port.cfg")
+    def mock_file_list(self, cfg_name: str) -> List[str]:
+        """file_list mock module"""
+        return [cfg_name]
```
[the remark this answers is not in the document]
riya-17: make sense

lib/pbench/test/unit/common/test_config.py
riya-17: right

lib/pbench/test/unit/common/test_config.py
riya-17: it seems a little cluttery but sure

lib/pbench/test/unit/common/test_config.py
riya-17: sure

lib/pbench/test/unit/common/test_config.py
riya-17: yeah

lib/pbench/test/unit/common/test_config.py
riya-17: yes

lib/pbench/test/unit/common/test_config.py
webbnh: OK, so `log_using_caller_directory` is only set to `True` by a subclass of `PbenchConfig`...so, in the tests in this, module it _should_ always be `False`.

But, it would probably be worth checking that in more than just this one test (e.g., it would be good to add this assertion to the others collected in `test_logger()`...which you've already done!  🎉).

lib/pbench/test/unit/common/test_config.py
```diff
@@ -29,224 +29,119 @@ def get(self, section: str, option: str, fallback=None) -> str:
...
-            def __init__(self):
-                self._sections = {
-                    "logging": {"logger_type": "hostport", "logger_host": "localhost"}
-                }
-
-        monkeypatch.setattr(configtools, "file_list", self.mock_file_list)
-        monkeypatch.setattr(ConfigParser, "__init__", mock_MockedConfigParser.__init__)
-        monkeypatch.setattr(ConfigParser, "get", MockedConfigParser.get)
+            PbenchConfig("pbench.cfg")
+        expected_error_msg = "ValueError('logger_host', 'logging')"
```
webbnh, line 102: The sub-cases in this scenario aren't really sharing much in the way of setup.  So, if you aren't going to go with the parameterized approach, then I would split these sub-cases into separate cases.

Having each test exercising a single scenario is helpful...if for no other reason that it will help highlight it when you miss an assertion.  (E.g., at line 102 you set up `expected_error_msg`, but there's no subsequent `assert` which checks it!  😈)

lib/pbench/test/unit/common/test_config.py
```diff
@@ -29,224 +29,119 @@ def get(self, section: str, option: str, fallback=None) -> str:
...
+                    "log_format": log_format,
+                }
+            },
+        )
         config = PbenchConfig("pbench.cfg")
-        assert config.logger_type == "file"
+        assert config.logger_type == logger_type
+        assert config.log_dir == log_dir
+        assert config.default_logging_level == logging_level
+        assert config.log_fmt == log_format
```
webbnh, line 85: This is a "success" case:  should we be testing the whole "interface", here?  (That is, we have four assertions here, whereas at line 55 we have six...should we be checking `log_using_caller_directory` and `TZ` here as well?)

Correspondingly, at line 145, should we be making the other five assertions?

webbnh (changes requested):
This is great, and almost "there"...however, you dropped an assertion.

lib/pbench/test/unit/common/test_config.py
webbnh: Unfortunately, @dbutenhof, I'm not sure that `str(exc)` produces the same result as `str(exc.value)` (or, at least, they don't produce the result that I as a human expect...); e.g., note the expected message in this snippet from this PR:
```
        with pytest.raises(ValueError) as exc:
            PbenchConfig("pbench.cfg")
        expected_error_msg = "ValueError('logger_port', 'logging')"
        assert expected_error_msg in str(exc)
```
That message looks like the `repr(exc.value)` value.  `str(exc.value)` would be just the values (in this rather weird case, it's a tuple).  Me, I might have expected the message to look something like `ValueError: ('logger_port', 'logging')`...but I'm thinking that `ValueError` (and our (mis)use of it) might be a bad example to be quibbling over.  😬

dbutenhof (approved):

lib/pbench/test/unit/common/test_config.py
```diff
@@ -1,68 +1,147 @@
...
+                    "logger_type": logger_type,
+                    "logger_host": logger_host,
+                    "logger_port": "",
+                }
+            },
+        )
+        with pytest.raises(ValueError) as exc:
+            PbenchConfig("pbench.cfg")
+        expected_error_msg = "ValueError('logger_port', 'logging')"
+        assert expected_error_msg in str(exc)
```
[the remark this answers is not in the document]
webbnh: > I don't know how ConfigParser names the ValueError attributes being reported here.

I presume that it goes like this:
```
    assert exc.value.args[1] == "logger_port"
    assert exc.value.args[0] == "logging"
```
...but I'm not excited about actually using this approach....

lib/pbench/test/unit/common/test_config.py
riya-17: Sure. we can add them, but that won't make any difference, as they are hardcoded.

lib/pbench/test/unit/common/test_config.py
riya-17: I didn't realize that I missed an assertion. What is the parametrized approach?

portante (approved):

lib/pbench/test/unit/common/test_config.py
webbnh: > that won't make any difference, as they are hardcoded.

😆 That's true...until someone changes the code!!

Seriously, the reason why the unit test code is separate from the CUT is so that the unit tests can maintain the notion of "truth" regardless of the implementation of the CUT or of subsequent changes to it.

To our best ability, the unit tests should be "black box" tests -- that is, they should be implemented and operate with as little knowledge of how the CUT is implemented as is feasible, while still generating full coverage of the CUT's operation.  (So, yeah, when we write the unit tests, we "peek" inside the CUT, but what the unit tests actually check should be the "external interface" of the "unit", not its internals.)  Thus, the fact that this aspect of the interface happens to be hard-coded today is immaterial:  given that it is part of the interface, it should be checked by the unit tests, and it is appropriate to check it so that a future change to the CUT doesn't break the "interface contract".

lib/pbench/test/unit/common/test_config.py
```diff
@@ -81,34 +81,41 @@ def test_custom_logger(self, monkeypatch):
...
             monkeypatch, {"logging": {"logger_type": logger_type, "logger_host": ""}}
         )
         with pytest.raises(ValueError) as exc:
             PbenchConfig("pbench.cfg")
-        expected_error_msg = "ValueError('logger_host', 'logging')"
+        assert exc.value.args[0] == "logger_host"
+        assert exc.value.args[1] == "logging"
+
         logger_host = "logger.example.com"
+        expected_error_msg = "No option 'logger_port' in section: 'logging'"
```
webbnh, line 110: We're not actually using the value assigned at line 100.  So, we should either omit that line or rework lines 106-107 to be more like line 98 (or 117).  (That is, unless there is some reason to do it differently here, I would make the code here match what it does in those other places.)

Ditto for line 129 (and 132-133).

webbnh (dismissed):
Looks good.

However, it would be better if `test_bad_logger_hostport()` used a consistent approach to checking the exception that it catches (and, I recommend the `str(exc.value)` approach)...unless there is some reason for the currently proposed alternation of approaches.

lib/pbench/test/unit/common/test_config.py
riya-17: I went with the parametrize approach.. I hope it looks good :D

lib/pbench/test/unit/common/test_config.py
```diff
@@ -1,68 +1,173 @@
...
+    def test_logger_hostport_ValueError(
+        self, config_value, expected_error_msg, monkeypatch
+    ):
+        """Show that PbenchConfig raises 'ValueError' when the
+        option contains an empty value."""
+
+        self.logger_test_setup(monkeypatch, config_value)
+        with pytest.raises(ValueError) as exc:
+            PbenchConfig("pbench.cfg")
+        assert expected_error_msg in str(exc)
```
webbnh, line 146: Since the only difference (that I can spot) between `test_logger_hostport_BadConfig()` and `test_logger_hostport_ValueError()` is the exception passed to the `with` statement, I would add that to the parametrize-ation and combine the two test cases together.

webbnh (approved):
This looks good, Riya!

> I went with the parametrize approach.. I hope it looks good :D

It looks pretty ugly, but I think it's the better approach.  🙂

In fact, I would take it one step further -- see below.

webbnh (approved):
🎉

PR #2912, opened 2022-06-21
Fix FileTree.find_dataset to link unpacked data

dbutenhof:
PBENCH-774

The logic to discover and link INCOMING directories and RESULTS links was too far out, and weren't being used by `find_dataset`. Refactor this into object construction so it doesn't get missed.

### DIFF
--- a/lib/pbench/server/filetree.py
+++ b/lib/pbench/server/filetree.py
@@ -100,23 +100,39 @@ class Tarball:
         self.controller_name: str = controller.name
         self.metadata: Optional[JSONOBJECT] = None
 
-    def record_unpacked(self, directory: Path):
+    def check_unpacked(self, incoming: Path) -> bool:
         """
-        Record that an unpacked tarball directory was found for this dataset.
+        Determine whether a tarball in the ARCHIVE tree has been unpacked into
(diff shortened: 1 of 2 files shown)


lib/pbench/server/filetree.py
```diff
@@ -698,7 +693,7 @@ def __contains__(self, dataset: str) -> bool:
         """
         return dataset in self.datasets
 
-    def __getitem__(self, dataset: str) -> Dataset:
+    def __getitem__(self, dataset: str) -> Tarball:
```
webbnh, line 696: This change seems like it's in the wrong PR.

webbnh (approved):
Looks OK, but there's one change which maybe belongs in your other PR?

lib/pbench/server/filetree.py
dbutenhof: No; this is just fixing a day-1 bug because I typed the wrong type. `FileTree` never returns a `Dataset` object, and what #2906 attempts is to clarify the distinction better. Interestingly, I completely missed this bug in that context.

npalaska (approved):

siddardh-ra (approved):
Looks good!

PR #2913, opened 2022-06-22
Default to Python 3.9 for unit tests

portante:
Test Python 3.6 as well for the agent side only.

### DIFF
--- a/jenkins/Pipeline.gy
+++ b/jenkins/Pipeline.gy
@@ -16,6 +16,12 @@ pipeline {
                 sh 'jenkins/run tox'
             }
         }
+        stage('Agent Python3.6 Check') {
+            steps {
+                echo 'Verify agent side works with Python 3.6'
+                sh 'jenkins/run tox -e py36 -- agent'
+            }
+        }
     }
     post {
         success {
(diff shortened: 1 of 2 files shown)


jenkins/Pipeline.gy
```diff
@@ -16,6 +16,12 @@ pipeline {
                 sh 'jenkins/run tox'
             }
         }
+        stage('Agent Python3.6 Check') {
+            steps {
+                echo 'Verify agent side works with Python 3.6'
+                sh 'jenkins/run tox -e py36 -- agent'
+            }
+        }
```
dbutenhof, line 24: This is interesting in two dimensions:
1. We're testing with Python 3.6 on the Pbench dev container, which isn't RHEL 7; so while the 3.6 test is independently valuable, it's not clear we're really confirming anything about the intended installation environment. 😆 
2. At the moment, the server is running across a variety of Python versions on various platforms ... it would be just as valuable to run 3.6 & 3.9 (and maybe 3.8) here, even if, eventually, we'll be dropping back to a single containerized version. Why *not* just test everything here?

dbutenhof (approved):

webbnh (approved):
👍

PR #2915, opened 2022-06-24
Allow changing the dataset resource name and deletion date

dbutenhof:
PBENCH-749

Having changed the API to use the unique tarball MD5 as the formal resource ID, we're free to allow users to select more meaningfull resource names for filtering and display.

This PR adds a layer allowing the owner of a dataset to use the metadata PUT API to alter the dataset.name and server.deletion metadata keys with basic validity checking:
- The name value must be a UTF-8 string from 1 to 32 characters
- The deletion value must be a valid date/time string that does not exceed the server's maximum dataset lifetime from the dataset's upload timestamp.

### DIFF
--- a/lib/pbench/server/__init__.py
+++ b/lib/pbench/server/__init__.py
@@ -45,6 +45,10 @@ class PbenchServerConfig(PbenchConfig):
     for the pbench server environment.
     """
 
+    # Define a fallback default for dataset maximum retention, which we expect
+    # to be defined in pbench-server-default.cfg.
+    MAXIMUM_RETENTION_DAYS = 3650
+
     def __init__(self, cfg_name):
         super().__init__(cfg_name)
 
@@ -160,6 +164,24 @@ class PbenchServerConfig(PbenchConfig):
     @property
     def rest_uri(self):
         return self._get_conf("pbench-server", "rest_uri")
+
+    @property
+    def max_retention_period(self) -> timedelta:
+        """
+        Produce a timedelta representing the number of days the server allows
+        a dataset to be retained.
+
+        Returns
+            delta time representing retention period
+        """
+        try:
+            retention_days = int(
+                self._get_conf("pbench-server", "maximum-dataset-retention-days")
+            )
+        except (NoOptionError, NoSectionError):
+            retention_days = self.MAXIMUM_RETENTION_DAYS
+
+        return timedelta(days=retention_days)
 
     def _get_conf(self, section, option):
         """
--- a/lib/pbench/server/database/database.py
+++ b/lib/pbench/server/database/database.py
@@ -26,9 +26,11 @@ class Database:
 
     @staticmethod
     def init_db(server_config, logger):
-        # Attach the logger to the base class for models to find
+        # Attach the logger and server config object to the base class for
+        # database models to find
         if logger and not hasattr(Database.Base, "logger"):
             Database.Base.logger = logger
+            Database.Base.config = server_config
 
         # WARNING:
         # Make sure all the models are imported before this function gets called
--- a/lib/pbench/server/database/models/datasets.py
+++ b/lib/pbench/server/database/models/datasets.py
@@ -1124,7 +1124,9 @@ class Metadata(Database.Base):
                 or len(value) > __class__.MAX_NAME_LEN
                 or len(value) < __class__.MIN_NAME_LEN
             ):
-                raise MetadataBadValue(dataset, key, value, "UTF-8 string")
+                raise MetadataBadValue(
+                    dataset, key, value, "UTF-8 string of 1 to 32 characters"
+                )
 
             try:
                 value.encode("utf-8", errors="strict")
--- a/lib/pbench/server/database/models/datasets.py
+++ b/lib/pbench/server/database/models/datasets.py
@@ -1125,7 +1125,10 @@ class Metadata(Database.Base):
                 or len(value) < __class__.MIN_NAME_LEN
             ):
                 raise MetadataBadValue(
-                    dataset, key, value, "UTF-8 string of 1 to 32 characters"
+                    dataset,
+                    key,
+                    value,
+                    f"UTF-8 string of {__class__.MIN_NAME_LEN} to {__class__.MAX_NAME_LEN} characters",
                 )
 
             try:
--- a/lib/pbench/server/database/models/datasets.py
+++ b/lib/pbench/server/database/models/datasets.py
@@ -899,6 +899,13 @@ class Metadata(Database.Base):
     # by a period.
     _valid_key_charset = re.compile(r"[a-z0-9_.-]+")
 
+    # Dataset name size limits
+    MIN_NAME_LEN = 1
+    MAX_NAME_LEN = 32
+
+    # 1 day as a delta time
+    ONE_DAY = datetime.timedelta(days=1)
+
     id = Column(Integer, primary_key=True, autoincrement=True)
     key = Column(String(255), unique=False, nullable=False, index=True)
     value = Column(JSON, unique=False, nullable=True)
@@ -1083,12 +1090,12 @@ class Metadata(Database.Base):
 
         We handle two keys specially:
 
-            dataset.name:   This is the primary resource name of a dataset, and
-                can be modified by the dataset's owner
-            server.deletion: This is the expected automatic deletion date of
-                the dataset, and can be modified so long as the new value is a
-                valid date and falls within the maximum server retention period
-                from the dataset's upload timestamp.
+            dataset.name: The primary resource name of a dataset, which can be
+                modified by the dataset's owner.
+            server.deletion: The expected automatic deletion date of the
+                dataset. It can be modified by the owner so long as the new
+                value is a valid date that falls within the maximum server
+                retention period from the dataset's upload timestamp.
 
         Args:
             dataset: Associated Dataset
@@ -1103,30 +1110,43 @@ class Metadata(Database.Base):
 
         # Setting the dataset name is a direct modification to the Dataset SQL
         # column, so do that first and exit without touching the Metadata
-        # table.
+        # table. We require a UTF-8 encoded string of 1 to 32 characters.
         #
         # If we're setting the reserved "server.deletion" key, then we must
-        # be able to parse the string as a date/time, and we fail otherwise.
+        # be able to parse the string as a date/time, and we fail otherwise. We
+        # store only the UTC date, as we don't guarantee that deletion will
+        # occur at any specific time of day.
+        #
         # For any other key value, there's no required format.
         if key == __class__.DATASET_NAME:
-            if value:
-                dataset.name = str(value)
-                dataset.update()
-            else:
-                raise MetadataBadValue(dataset, key, value, "string")
+            if (
+                type(value) is not str
+                or len(value) > __class__.MAX_NAME_LEN
+                or len(value) < __class__.MIN_NAME_LEN
+            ):
+                raise MetadataBadValue(dataset, key, value, "UTF-8 string")
+
+            try:
+                value.encode("utf-8", errors="strict")
+            except UnicodeDecodeError as u:
+                raise MetadataBadValue(dataset, key, value, "UTF-8 string") from u
+
+            dataset.name = value
+            dataset.update()
             return
         elif key == __class__.DELETION:
             try:
                 target = date_parser.parse(value).astimezone(datetime.timezone.utc)
-                v = target.isoformat()
-                maximum = dataset.uploaded + __class__.config.max_retention_period
-            except date_parser.ParserError:
-                raise MetadataBadValue(dataset, key, value, "date/time")
+            except date_parser.ParserError as p:
+                raise MetadataBadValue(dataset, key, value, "date/time") from p
 
+            maximum = dataset.uploaded + __class__.config.max_retention_period
             if target > maximum:
                 raise MetadataBadValue(
                     dataset, key, value, f"date/time before {maximum:%Y-%m-%d}"
                 )
+            target += __class__.ONE_DAY
+            v = f"{target:%Y-%m-%d}"
         else:
             v = value
 
--- a/lib/pbench/server/database/models/datasets.py
+++ b/lib/pbench/server/database/models/datasets.py
@@ -5,6 +5,7 @@ from pathlib import Path
 import re
 from typing import Any, Dict, List, Optional, Union
 
+from dateutil import parser as date_parser
 from sqlalchemy import Column, DateTime, Enum, event, ForeignKey, Integer, JSON, String
 from sqlalchemy.exc import IntegrityError, SQLAlchemyError
 from sqlalchemy.orm import relationship, validates, Query
@@ -234,6 +235,24 @@ class MetadataBadKey(MetadataKeyError):
 
     def __str__(self) -> str:
         return f"Metadata key {self.key!r} is not supported"
+
+
+class MetadataBadValue(MetadataError):
+    """
+    An unsupported value was specified for a special metadata key
+
+    The error text will identify the metadata key that was specified and the
+    actual and expected value type.
+    """
+
+    def __init__(self, dataset: "Dataset", key: str, value: str, expected: str):
+        self.name = dataset.name
+        self.key = key
+        self.value = value
+        self.expected = expected
+
+    def __str__(self) -> str:
+        return f"Metadata key {self.key!r} value {self.value!r} for dataset {self.name!r} must be a {self.expected}"
 
 
 class MetadataProtectedKey(MetadataKeyError):
@@ -795,6 +814,11 @@ class Metadata(Database.Base):
     # {"dataset.created": "3000-03-30T03:30:30.303030+00:00"}
     DATASET = "dataset"
 
+    # The Dataset name column can be modified by the owner
+    #
+    # {"dataset.name": "my name string"}
+    DATASET_NAME = f"{DATASET}.name"
+
     # SERVER is an internally maintained key namespace for additional metadata
     # relating to the server's management of datasets. The information here is
     # accessible to callers, but can't be changed.
@@ -858,10 +882,13 @@ class Metadata(Database.Base):
     # --- Standard Metadata keys
 
     # Metadata keys that clients can update
-    USER_UPDATEABLE_METADATA = [DASHBOARD, USER]
+    #
+    # NOTE: the entire "dashboard" and "user" namespaces are writable, but only
+    # specific keys in the "dataset" and "server" namespaces can be modified.
+    USER_UPDATEABLE_METADATA = [DASHBOARD, DATASET_NAME, DELETION, USER]
 
-    # Metadata keys that are accessible to clients
-    METADATA_KEYS = sorted(USER_UPDATEABLE_METADATA + [DATASET, SERVER])
+    # Metadata keys that clients can read
+    METADATA_KEYS = sorted([DASHBOARD, DATASET, SERVER, USER])
 
     # NOTE: the ECMA JSON specification allows JSON "names" (keys) to be any
     # string, though most implementations and schemas enforce or recommend
@@ -1041,9 +1068,7 @@ class Metadata(Database.Base):
         return value
 
     @staticmethod
-    def setvalue(
-        dataset: Dataset, key: str, value: Any, user: Optional[User] = None
-    ) -> "Metadata":
+    def setvalue(dataset: Dataset, key: str, value: Any, user: Optional[User] = None):
         """
         Create or modify an existing metadata value. This method supports
         hierarchical dotted paths like "dashboard.seen" and should be used in
@@ -1054,20 +1079,55 @@ class Metadata(Database.Base):
         "a.b.c" a value of "bar", and then query "a", you'll get back
         "a": {"b": {"c": "bar"}}}
 
+        We handle two keys specially:
+
+            dataset.name:   This is the primary resource name of a dataset, and
+                can be modified by the dataset's owner
+            server.deletion: This is the expected automatic deletion date of
+                the dataset, and can be modified so long as the new value is a
+                valid date and falls within the maximum server retention period
+                from the dataset's upload timestamp.
+
         Args:
             dataset: Associated Dataset
             key: Lookup key (including hierarchical dotted paths)
             value: Value to be assigned to the specified key
-
-        Returns:
-            A Metadata object representing the new database row with the
-            properly assigned value.
         """
         if not Metadata.is_key_path(key, Metadata.METADATA_KEYS):
             raise MetadataBadKey(key)
         keys = key.lower().split(".")
         native_key = keys.pop(0)
         found = True
+
+        # Setting the dataset name is a direct modification to the Dataset SQL
+        # column, so do that first and exit without touching the Metadata
+        # table.
+        #
+        # If we're setting the reserved "server.deletion" key, then we must
+        # be able to parse the string as a date/time, and we fail otherwise.
+        # For any other key value, there's no required format.
+        if key == __class__.DATASET_NAME:
+            if value:
+                dataset.name = str(value)
+                dataset.update()
+            else:
+                raise MetadataBadValue(dataset, key, value, "string")
+            return
+        elif key == __class__.DELETION:
+            try:
+                target = date_parser.parse(value).astimezone(datetime.timezone.utc)
+                v = target.isoformat()
+                maximum = dataset.uploaded + __class__.config.max_retention_period
+            except date_parser.ParserError:
+                raise MetadataBadValue(dataset, key, value, "date/time")
+
+            if target > maximum:
+                raise MetadataBadValue(
+                    dataset, key, value, f"date/time before {maximum:%Y-%m-%d}"
+                )
+        else:
+            v = value
+
         try:
             meta = Metadata.get(dataset, native_key, user)
 
@@ -1080,7 +1140,7 @@ class Metadata(Database.Base):
             meta_value = {}
 
         if not keys:
-            meta_value = value
+            meta_value = v
         else:
             walker = meta_value
             name = native_key
@@ -1095,7 +1155,7 @@ class Metadata(Database.Base):
                 # If we've reached the final key, assign the value in this
                 # dict; otherwise continue descending.
                 if i == len(keys) - 1:
-                    walker[inner_key] = value
+                    walker[inner_key] = v
                 else:
                     if inner_key not in walker:
                         walker[inner_key] = {}
@@ -1109,7 +1169,6 @@ class Metadata(Database.Base):
             meta = Metadata.create(
                 dataset=dataset, key=native_key, value=meta_value, user=user
             )
-        return meta
 
     @staticmethod
     def _query(dataset: Dataset, key: str, user: Optional[User]) -> Query:
--- a/lib/pbench/server/database/models/datasets.py
+++ b/lib/pbench/server/database/models/datasets.py
@@ -203,6 +203,26 @@ class MetadataBadStructure(MetadataError):
         return f"Key {self.element!r} value for {self.key!r} in {self.dataset} is not a JSON object"
 
 
+class MetadataBadValue(MetadataError):
+    """
+    An unsupported value was specified for a special metadata key
+
+    The error text will identify the metadata key that was specified and the
+    actual and expected value type.
+    """
+
+    def __init__(self, dataset: "Dataset", key: str, value: str, expected: str):
+        super().__init__(dataset, key)
+        self.value = value
+        self.expected = expected
+
+    def __str__(self) -> str:
+        return (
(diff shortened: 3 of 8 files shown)


lib/pbench/server/database/models/datasets.py
```diff
@@ -1054,20 +1081,55 @@ def setvalue(
         "a.b.c" a value of "bar", and then query "a", you'll get back
         "a": {"b": {"c": "bar"}}}
 
+        We handle two keys specially:
+
+            dataset.name:   This is the primary resource name of a dataset, and
```
portante, line 1086: Can we consider making sure the original name is always saved so that it can be restored via the UI?

portante (commented):
> - Should we limit the character set and/or length of the resource name to make display more manageable? I was thinking perhaps something like `r"[\w-]{1,32}"` for example.

UTF-8, maximum 256 characters?

> - I coded this to allow any truthy value (`if value:`), then realized this technically allowed non-string values and whimsically added an explicit `str(value)` and tests for non-string cases. This is probably ridiculous... but should we require `type(value) is str`?

Yes, UTF-8.

> - I'm recording the full `datetime.isoformat()` in `server.deletion`, whereas before this change I only set a date. I really doubt we'll ever want to proactively schedule automatic deletion at specific times; more likely we'll simply sweep once or twice a day, perhaps even less. So should I just record `{date:%Y-%m-%d}`?

Could we consider this to always be an integer number of days from upload date, so that it becomes a TTL?  But that would mean we'd need a field for when it was uploaded.  And then we'd probably want to do round-to-next-day(upload_date + TTL) and document that as the behavior.

lib/pbench/server/database/models/datasets.py
dbutenhof: I considered that, decided not to waste extra space because it can be reconstructed from `dataset.metalog.{script + config + date}` just as it was created in the first place.

I also considered that we could add a `--name` option on the wrapper scripts on the agent side; but that would have to be communicated to the server via a new `metadata.log` value (e.g., `pbench/name`) so it'd still be recorded immutably (and available to clients as metadata).

dbutenhof (commented):
> UTF-8, maximum 256 characters?

256 characters is big for an accessible UI display. That's a paragraph; my preference would be to keep the *name* relatively short and accessible, while also allowing arbitrary additional metadata for sorting and searching.

For similar reasons I'd be tempted to restrict whitespace and punctuation.

This, however, is an excellent start to a discussion. 👍🏻 

> Could we consider this to always be an integer number of days from upload date, so that it becomes a TTL? But that would mean we'd need a field for when it was uploaded. And then we'd probably want to do round-to-next-day(upload_date + TTL) and document that as the behavior.

We certainly could do it that way; I think a date is easier to read and process, however, instead of making them guess.

lib/pbench/server/__init__.py
```diff
@@ -45,6 +45,10 @@ class PbenchServerConfig(PbenchConfig):
     for the pbench server environment.
     """
 
+    # Define a fallback default for dataset maximum retention, which we expect
+    # to be defined in pbench-server-default.cfg.
+    MAXIMUM_RETENTION_DAYS = 3650
```
npalaska, line 50: If we expect it to come from pbench-server-default why do we define it here?

lib/pbench/server/__init__.py
```diff
@@ -161,6 +165,24 @@ def _ref_datetime(self):
...
+
+        Returns
+            delta time representing retention period
+        """
+        try:
+            retention_days = int(
+                self._get_conf("pbench-server", "maximum-dataset-retention-days")
+            )
+        except (NoOptionError, NoSectionError):
+            retention_days = self.MAXIMUM_RETENTION_DAYS
```
npalaska, line 182: If we expect the "default" config file to have this value defined, should we fail if it is not there (since clearly, it's changed at this point by some other means)

npalaska (dismissed):
Looks good, just couple questions

webbnh:
> > * Should we limit the character set and/or length of the resource name to make display more manageable? I was thinking perhaps something like `r"[\w-]{1,32}"` for example.
> 
> UTF-8, maximum 256 characters?

UTF-8 seems reasonable; 256 characters seems large but not unreasonably so.  (Regardless, anything over about 30 is going to be challenging for the UI....)

> > * I coded this to allow any truthy value (`if value:`), then realized this technically allowed non-string values and whimsically added an explicit `str(value)` and tests for non-string cases. This is probably ridiculous... but should we require `type(value) is str`?
> 
> Yes, UTF-8.

I presume that the Server is going to receive this value as a JSON value or similar (I haven't looked at the code, yet).  Given that, I would just treat it as a string (or string-like thing) and be done with it.  (That is, if someone gives us a number, the stringized version of that becomes the name; if someone gives us a JS "array" or "object", then the stringized version of _that_ becomes the name; we don't need to care -- it's the GIGO philosophy -- so long as what they give us doesn't break the UI (e.g., we should probably filter or escape control characters, etc.).)

> > * I'm recording the full `datetime.isoformat()` in `server.deletion`, whereas before this change I only set a date. I really doubt we'll ever want to proactively schedule automatic deletion at specific times; more likely we'll simply sweep once or twice a day, perhaps even less. So should I just record `{date:%Y-%m-%d}`?
> 
> Could we consider this to always be an integer number of days from upload date, so that it becomes a TTL? But that would mean we'd need a field for when it was uploaded. And then we'd probably want to do round-to-next-day(upload_date + TTL) and document that as the behavior.

On the Server side, I'm not convinced that it matters.  Whether it is TTL or an expiration date or time, the Server is going to have to decide, when it does a sweep, whether the dataset has expired or not.  It seems simplest to me to record it as a full date-time, but recording it as a UTC date seems like it will work just as well.  TTL will require either producing an expiration date from the origin (presumably the upload date-time) or producing a comparable interval (by subtracting the origin from the current time), so that strikes me as being more work than just recording an expiration date/date-time...but, the Server and User "expiration options" will presumably be TTL values, so using TTL will make things easier for the upload operation at the expense of the sweep operation.

But, regardless of what we choose, the UI is going to have to present "expiration time" to the User in some reasonable way.  If we're doing TTL, then either the UI is going to have to convert it to a date or it's going to need to adjust it and present it as "time remaining".  And, if we're doing a date/date-time, then the UI needs to present it in a way which doesn't give the wrong impression or is otherwise confusing...so, claiming that we'll delete it at a particular second of a particular day probably isn't useful either.  (And, if it's a date/date-time, then the UI is probably going to need to convert it to a TTL value to decide whether to warn the User about the dataset's imminent demise, although I suppose it could compute a threshold _date_ value and do a date-to-date comparison instead.)

So, I really don't think it matters.  My instinct is to go with an "expiration _date_" (no time), as I think that that is the most straightforward and generally-easiest representation to work with.  (I don't think that having a time makes anything better, and being _required_ to convert between TTL and dates may be unhelpful.)

webbnh (approved):
Looks generally good, although I've identified a couple of rough edges below, which you should consider addressing.

However, I would like to register some opposition to allowing the owner to change the name of the dataset.  I think that we should be providing labels and annotations which make this unnecessary.  Moreover, once the UI provides the ability to share URLs, the _name_ of the dataset won't really be important, because the dataset will be accessible directly by URL or indirectly by label.  (Given that we place such importance on the client-side filesystem view and the name of the tarball, it seems weird that we're willing to accept a change to the dataset name in other contexts.)

Also, I think this sets us up for a further debate over what the UI default view should be:  we want it to be the script/config/date triplet (in separate columns) rather than the default name, but, if we can set the name then we presumably would rather it be the name...unless the user hasn't set names, in which case we would rather it be the script/config/date triplet....  If the user cannot set the name, then the default is (I think) obvious -- it's the triplet -- and the user can add and sort on labels if s/he wishes.

And, then there's the fact that allowing the owner to change the name means having to special-case the key in what is otherwise an immutable namespace.  This stands in juxtaposition to adding support for labels, which presumably would go in the `dashboard` namespace where they make good sense (and can be added without changes to the Server code...).

riya-17 (dismissed):

lib/pbench/server/__init__.py
dbutenhof: This is a pattern followed in many other places. We've discussed before the idea of doing config file validation on startup; there's probably even an issue on that, although I won't bother to look it up right now.

lib/pbench/server/__init__.py
dbutenhof: OK, I did look it up: [PBENCH-520](https://issues.redhat.com/browse/PBENCH-520) covers startup validation of config options. If we did that, we could fail on server startup for missing options rather than having fallback defaults ... if we really want to go that way.

lib/pbench/server/database/database.py
```diff
@@ -26,9 +26,11 @@ def get_engine_uri(config, logger):
 
     @staticmethod
     def init_db(server_config, logger):
-        # Attach the logger to the base class for models to find
+        # Attach the logger and server config object to the base class for
+        # database models to find
         if logger and not hasattr(Database.Base, "logger"):
             Database.Base.logger = logger
+            Database.Base.config = server_config
```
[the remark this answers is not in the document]
dbutenhof: Our whole logging infrastructure needs some work; we generally create a new logger with non-default options, and that's impossible to recreate without access to the config object, which we can really only get at the CLI level (e.g., it might be `_PBENCH_SERVER_CONFIG` or `-C` ... the former we could access anywhere, but not the latter). I suppose we should be modifying the default logger so that we can just `logging.getlogger()` in any arbitrary python module (which is how it's supposed to work).

Right now, having the base classes wired is expedient.

And, yeah, I probably shouldn't have nested the config setup under the logger test, although I'm not sure that the check isn't silly in the first place. Sigh.

lib/pbench/server/database/models/datasets.py
```diff
@@ -1054,20 +1081,55 @@ def setvalue(
...
+            else:
+                raise MetadataBadValue(dataset, key, value, "string")
+            return
+        elif key == __class__.DELETION:
+            try:
+                target = date_parser.parse(value).astimezone(datetime.timezone.utc)
+                v = target.isoformat()
+                maximum = dataset.uploaded + __class__.config.max_retention_period
+            except date_parser.ParserError:
+                raise MetadataBadValue(dataset, key, value, "date/time")
```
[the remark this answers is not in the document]
dbutenhof: Life was simpler when you had everyone convinced that the `from` was redundant! As Bob Seger said, "I wish I didn't know now what I didn't know then."

lib/pbench/test/unit/server/test_datasets_metadata.py
```diff
@@ -296,7 +296,7 @@ def test_put_bad_keys(self, client, server_config, attach_dataset):
         )
         assert response.status_code == HTTPStatus.BAD_REQUEST
         assert response.json == {
-            "message": "Unrecognized JSON keys ['what', 'xyzzy'] given for parameter metadata; allowed namespaces are ['dashboard', 'user']"
+            "message": "Unrecognized JSON keys ['what', 'xyzzy'] given for parameter metadata; allowed namespaces are ['dashboard', 'dataset.name', 'server.deletion', 'user']"
```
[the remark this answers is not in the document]
dbutenhof: Only those specific key paths are allowed by the keyword check, which is sensitive to key namespace. (That is, we allow `user.xyzzy` because the `user` namespace is allowed for `PUT`; but only `dataset.name` and `server.deletion` are listed for those namespaces.)

dbutenhof (commented):
> Also, I think this sets us up for a further debate over what the UI default view should be: we want it to be the script/config/date triplet (in separate columns) rather than the default name, but, if we can set the name then we presumably would rather it be the name...unless the user hasn't set names, in which case we would rather it be the script/config/date triplet.... If the user cannot set the name, then the default is (I think) obvious -- it's the triplet -- and the user can add and sort on labels if s/he wishes.

We want to help users find what they want. We want a lot more metadata flexibility in searching, sorting, and display; and while being able to filter on (and display) the specific benchmark script you used, or the date range, and certainly the "config" string, is useful, having a "name" with a pre-defined and universal semantic but under user control seems like a good idea.  Unlike `dashboard.name` or whatever, it's always defined, and has some potentially useful default (whether we continue defaulting to `<script>_<config>_<date>` or change to something else we, or test users, think might be more useful).

They should be able to disable any column or add columns for any metadata, to view and filter and sort by whatever they want. But a "name" is a very familiar and nearly universal concept. And this resolves the problem of the "name" being an aggregation of data only vaguely under the user's control and of questionable distinction.

lib/pbench/server/database/models/datasets.py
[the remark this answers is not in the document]
webbnh: Yeah.  And, then Peter had to go point out that the difference was two trivial words buried in the middle of the output which changed the whole sense if its meaning.  (And, yeah, "We're older now but we're still running against the wind.")

portante:
> > > * I'm recording the full `datetime.isoformat()` in `server.deletion`, whereas before this change I only set a date. I really doubt we'll ever want to proactively schedule automatic deletion at specific times; more likely we'll simply sweep once or twice a day, perhaps even less. So should I just record `{date:%Y-%m-%d}`?
> > 
> > 
> > Could we consider this to always be an integer number of days from upload date, so that it becomes a TTL? But that would mean we'd need a field for when it was uploaded. And then we'd probably want to do round-to-next-day(upload_date + TTL) and document that as the behavior.
> 
> On the Server side, I'm not convinced that it matters.

Oh good, then if it doesn't matter, why are not just use a simple "relative" offset of "days" from it creation date (read, "an integer") instead of having to parse dates and interpret time zones?

Isn't this wading into the waters of over-engineering a solution?

portante:
Giving a name to something is a very powerful concept from antiquity.  Let's not lose that nature of a name vs a label.  If you want to think of the name as the "primary label" that the owner gives to an entity for others to consider, that is fine, too.

dbutenhof:
> > > > * I'm recording the full `datetime.isoformat()` in `server.deletion`, whereas before this change I only set a date. I really doubt we'll ever want to proactively schedule automatic deletion at specific times; more likely we'll simply sweep once or twice a day, perhaps even less. So should I just record `{date:%Y-%m-%d}`?
> > > 
> > > 
> > > Could we consider this to always be an integer number of days from upload date, so that it becomes a TTL? But that would mean we'd need a field for when it was uploaded. And then we'd probably want to do round-to-next-day(upload_date + TTL) and document that as the behavior.
> > 
> > 
> > On the Server side, I'm not convinced that it matters.
> 
> Oh good, then if it doesn't matter, why are not just use a simple "relative" offset of "days" from it creation date (read, "an integer") instead of having to parse dates and interpret time zones?
> 
> Isn't this wading into the waters of over-engineering a solution?

The expiration date is a bit easier to manage; we only compute it once when it's set, and then it's a date comparison. It's also easier for a human to read and understand, I think. We also need to repeatedly read the TTL, convert to a delta-time, add to the upload date, and then compare it against the current date. I simply don't see any advantage in storing as a TTL vs just keeping the date we actually need.

lib/pbench/server/database/models/datasets.py
```diff
@@ -1054,20 +1081,55 @@ def setvalue(
...
         "a": {"b": {"c": "bar"}}}
 
+        We handle two keys specially:
+
+            dataset.name:   This is the primary resource name of a dataset, and
+                can be modified by the dataset's owner
+            server.deletion: This is the expected automatic deletion date of
+                the dataset, and can be modified so long as the new value is a
+                valid date and falls within the maximum server retention period
+                from the dataset's upload timestamp.
```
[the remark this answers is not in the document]
webbnh: My foremost concern is to prevent the documentation from misleading the reader.  Since the code uses symbolic references, it creates a risk if the surrounding comments try to specify the contents of the indirection.  In particular, when/if the values of the symbols are ever changed, how can we be sure that the documentation will be properly updated?  Is this a debt that we should take on?  If it's so important that the documentation mention the real value, then perhaps the code should use the real value, as well!  (I think we have very good reasons to avoid using literals in the code, and it seems a stretch (to me) to say that those reasons don't apply to the documentation; it's true that "bugs in the documentation" don't affect the operation of the code, but they can affect those who read the code...and, if we train the readers to ignore the documentation then we might as well omit it and save time and space.)

lib/pbench/test/unit/server/database/test_datasets_metadata_db.py
```diff
@@ -363,3 +364,61 @@ def test_setgetvalue_user(self, attach_dataset):
...
+        Test the reaction to a "bad" (empty) string name.
+        """
+        ds = Dataset.query(name="drb")
+        name = ds.name
+        with pytest.raises(MetadataBadValue) as exc:
+            Metadata.setvalue(ds, "dataset.name", "")
+        assert (
+            str(exc.value)
+            == "Metadata key 'dataset.name' value '' for dataset drb(3)|drb must be a string"
+        )
```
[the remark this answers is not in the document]
webbnh: I'm inclined to follow the GIGO philosophy:  if the user can set it to _almost_ anything that s/he wants, why not just remove the blade guards altogether?  If the UI cannot display the result, well...the user did that to her-/himself...it's not the Server's or UI's fault.

lib/pbench/test/unit/server/database/test_datasets_metadata_db.py
```diff
@@ -365,26 +365,33 @@ def test_setgetvalue_user(self, attach_dataset):
...
         """
         ds = Dataset.query(name="drb")
         Metadata.setvalue(ds, "dataset.name", value)
-        assert Metadata.getvalue(ds, "dataset")["name"] == str(value)
+        assert Metadata.getvalue(ds, "dataset")["name"] == value
 
-    def test_mutable_dataset_bad(self, attach_dataset):
+    @pytest.mark.parametrize(
+        "value",
+        ["", b"\xe8", "This is a really long name that exceeds the maximum length"],
```
webbnh, line 382: The `b"\xe8"` fails, not because of its bit pattern but because of its type (`bytes`), right?  So, I don't think this quite tests what we want it to.

webbnh (commented):
The code looks good.  However, I still think that allowing the owner to change the name is a bad idea.  I prefer Peter's concept of having the concept of a "primary label" as a means for a user-applied name.

lib/pbench/test/unit/server/database/test_datasets_metadata_db.py
dbutenhof: True, but as I got more into playing with test cases late Friday and this morning I've discovered that I can't actually create bad unicode strings anyway since code that's merged needs to pass basic syntax checks and an invalid u"" string or "\uxxx" code point is a Python syntax error detected by flake8. Nor can I `decode` a bad byte string here in the unit test, because the `.decode` would fail.

Right now I'm not seeing any way around this. Because utf-8 can encode any Unicode code point, the only way to provoke a failure is with an illegal Unicode code point, which I can't represent anyway.

I'm afraid I went into this thinking that UTF-8 was a subset of UTF-16, and that a "valid UTF-16" would fail the `encode`, (and that test case wasn't quite right anyway, despite the hex representation)... but that's not the case, so it really doesn't matter anyway. 😦

dbutenhof (commented):
> Giving a name to something is a very powerful concept from antiquity. Let's not lose that nature of a name vs a label. If you want to think of the name as the "primary label" that the owner gives to an entity for others to consider, that is fine, too.

I'm really not sure what you're arguing for here, nor against what. I guess this was a response to earlier comments by Webb, but the "conversation" is extremely fragmented and difficult to follow.

While we want a flexible plethora of metadata users can apply to their resources and by which they can filter their views, as well as fully configurable display columns in the UI, I'm quite strongly attached to the idea of each resource having a "name" -- it's a virtually universal concept -- and that the name should be under user control.

We artificially limited that before with the agent's pre-generated name, and then using that as the resource ID as well as a name. One of the big advantages of shifting to the MD5 as the resource ID was that it allows us to let the user choose a meaningful name.

This is a universal inherent "metadata" property of the resource, one that everyone knows and understands, without needing to agree on a particular "dashboard.name" or other metadata key.

Our current `script`/`config`/`date` has multiple limitations; the date is pretty much arbitrary, and the benchmark script name is generally uninteresting except for `fio` and `uperf` which have specific semantics. We use `pbench-user-benchmark` to run an arbitrary benchmark script, *which isn't reflected in the name*, or as an orchestrator to measure a non-synthetic workload, which also isn't reflected here.

(Note that `dataset.metalog.iterations/1-default.user_script` has, for example, `"sleep, 10"` showing that I generated the dataset over 10 seconds of ambient server activity, but we don't record that anywhere else.)

lib/pbench/test/unit/server/database/test_datasets_metadata_db.py
webbnh: It seems ironic that Python, with its "duck typing", makes it prohibitively hard to create a "bad" string (or, perhaps it's _because_ of "duck typing"...that is, it cannot be a duck if it contains un-duck-like stuff!).  (However, if it's only Flake8 which is standing in the way, here, then this seems like an appropriate spot for a localized "disable checks" annotation, if that would enable the test.)

Anyway, if you end up making another change, I think it would be appropriate to replace the `bytes` object with a `list` or `dict` or something more obviously of the "wrong type", since, as things stand, it's not the "string contents" which provoke the failure.

dbutenhof: Yeah, but it's not *just* `flake8`, or LGTM: it's a Python ***syntax error***, so Python blows up parsing the test file even if the `flake8` warning was suppressed.

webbnh:
> > Giving a name to something is a very powerful concept from antiquity. Let's not lose that nature of a name vs a label. If you want to think of the name as the "primary label" that the owner gives to an entity for others to consider, that is fine, too.
> 
> I'm really not sure what you're arguing for here, nor against what. I guess this was a response to earlier comments by Webb, but the "conversation" is extremely fragmented and difficult to follow.

I concur...I wish that at least we were conducting it _in_ a "conversation".

> I'm quite strongly attached to the idea of each resource having a "name" -- it's a virtually universal concept -- and that the name should be under user control.

OK, I relent.

> This is a universal inherent "metadata" property of the resource, one that everyone knows and understands, without needing to agree on a particular "dashboard.name" or other metadata key.

Yeah, 'cause, instead of us saying that the name is in `"dashboard.name"` we're telling people that it's in `"server.name"`.  😆  This has has the advantage that we maintain the principal that the `"dashboard"` namespace is basically untouched by the server, but it comes at the cost of making part of the `"server"` namespace mutable by the user...which I'm not convinced is the appropriate trade-off.  (But, if we really do end up with additional client namespaces, perhaps it is.)

> Our current `script`/`config`/`date` has multiple limitations; the date is pretty much arbitrary, and the benchmark script name is generally uninteresting except for `fio` and `uperf` which have specific semantics.

I'm just stuck on the fact that Peter indicated (prior to your effort to make the name settable; and I agree with him) that the default display should be the triplet in three separate columns, because having them glued together to form a name was unpleasant.  With this change, the result name is going to become the primary column, and, by default, it's going to contain the triplet glued together...which is exactly what we said we didn't want!  🙃

I think it would be good if the UI (by default) used the triplet as the primary columns, to which the user can add a name column if he or she sees fit.  But, this will relegate the name to "just another piece of metadata" -- it will be indistinguishable from a "label", if we decide to add that feature -- in which case, it doesn't need or deserve anything like the first-class treatment which this PR gives it.

So, I think we should be providing a label capability, and get rid of the concept of a dataset "name" altogether.  (I mean, really, what's the difference between a "name" and a "label"?  Why stick with this concept from antiquity when we can have something much more general and flexible?  (1/2-🙂) )

lib/pbench/server/database/models/datasets.py
```diff
@@ -1124,7 +1124,9 @@ def setvalue(dataset: Dataset, key: str, value: Any, user: Optional[User] = None
                 or len(value) > __class__.MAX_NAME_LEN
                 or len(value) < __class__.MIN_NAME_LEN
             ):
-                raise MetadataBadValue(dataset, key, value, "UTF-8 string")
+                raise MetadataBadValue(
+                    dataset, key, value, "UTF-8 string of 1 to 32 characters"
+                )
```
webbnh, line 1129: Shouldn't this be
```
                raise MetadataBadValue(
                    dataset, key, value,
                    f"UTF-8 string of {__class__.MIN_NAME_LEN} to {__class__.MAX_NAME_LEN} characters"
                )
```

webbnh (approved):

lib/pbench/server/database/models/datasets.py
dbutenhof: You're not wrong, but, "ugh".

dbutenhof (commented):
> OK, I relent.

I can sense that you need some practice with this whole "relent" thing... 😆 

> So, I think we should be providing a label capability, and get rid of the concept of a dataset "name" altogether. (I mean, really, what's the difference between a "name" and a "label"? Why stick with this concept from antiquity when we can have something much more general and flexible? (1/2-slightly_smiling_face) )

Why complicate things by labeling the "name" as a "label", just to avoid calling it a name, when everyone knows what "name" means and "label" is very generic? We have support for arbitrary "label" metadata. I don't see how eliminating a "name" helps anyone.

Remember that while the name is accessible as `dataset.name`, it's also in the fundamental JSON object returned by `/datasets/list`, which is `{"resource_id": "<md5>", "name": "<name>"}`. Obviously we could demote it to the `"metadata": {}` object, requiring everyone to know how to reference it (and of course allowing them to omit it), but, again, I don't see how this is a benefit to anyone.

My idea was that `dataset` namespace is the identity of the user's data, while the `server` namespace is ancillary server context that we use to manage it but isn't inherent to the data. In my world, "name" clearly belongs with the identity even though we're allowing it to be changed.

webbnh (approved):
>> OK, I relent.
>
> I can sense that you need some practice with this whole "relent" thing... 😆

I did change my reviews from "Comment" to "Approve"....

> Why complicate things by labeling the "name" as a "label", just to avoid calling it a name, when everyone knows what "name" means and "label" is very generic? We have support for arbitrary "label" metadata. I don't see how eliminating a "name" helps anyone.

Well, because it's not obvious to me how a dataset "name" is different from a "label".  (In fact, I have the impression that names in general are basically just labels for things.)  With this change, it will be that we used to tag datasets with a name/label containing particular contents, and now we'll allow users to tag them using their own choice of contents.  The point of the name/label is to make it easier for a human to locate and recognize a particular dataset.  That seems like a fairly generic thing to do.  (And, it's not like datasets are people or pets -- we're not requiring the name to be _unique_, so -- it's really more of a label than a name.)

> Remember that while the name is accessible as dataset.name, it's also in the fundamental JSON object returned by `/datasets/list`, which is `{"resource_id": "<md5>", "name": "<name>"}`.

Yeah, and I'm pretty sure I raised the question in a recent PR wondering if we should remove it from there...since, with the resource ID being the proper way to refer to a dataset, the name doesn't really have much status above the rest of the metadata.  🙂

lib/pbench/server/database/models/datasets.py
webbnh: I just call 'em as I see 'em....  🍤

riya-17 (approved):

lib/pbench/server/database/models/datasets.py
portante: That presumes we have control over the behavior of the agent, which may not be the case.  I think we should preserve the original values either via an explicit field, or via mechanism where we can re-read from the original `metadata.log`.

portante: Tracking this in a Jira issue as future work.

portante (approved):
Approved after tracking comments in a separate Jira issue for the future.

lib/pbench/server/database/models/datasets.py
dbutenhof: I really don't understand your context here; if you really want to track some possible future work, the Jira issue needs a lot more information about your concern.

Are you trying to say that a tarball night not have a `metadata.log` with a `pbench` section containing `script`, `config` and `date` values? There's also already a `name` property, recording the original agent-side name. If we're supposed to operate without any of the `metadata.log` properties, then we need a fair amount of changes in the intake processing (e.g., the creation date comes from here, and we use it in indexing).

Or do you have something else in mind entirely that I'm not grasping?

npalaska (approved):
Looks good!

lib/pbench/server/database/models/datasets.py
portante: What I mean is that ...

> it can be reconstructed from `dataset.metalog.{script + config + date}` just as it was created in the first place

... implies that every pbench agent does it that way.  The server can't assume anything about the value of name since it does not have "control" over how it is set.  The value is controlled by the Pbench Agent code.

"We" as developers "control" the behavior, but enforcing that behavior will never be possible on the server side, unless the server ignores `name` and creates it from those fields.

So I'd rather see the server offer a way to either "save" the original name value in an immutable field, or offer a mechanism to restore it from the original `metadata.log` file.

PR #2917, opened 2022-06-25
Remove wayward "shbangs" in Python 3 library files

portante:
And we also remove the unnecessary `cli/getconf.py` file.

This work was identified while reviewing PR #2901.


ndokos (approved):

npalaska (approved):

dbutenhof (approved):

webbnh (approved):
I guess I have no problem with this, but a quick search suggests that we have 69 such files...why do we address only five of them here?

webbnh:
And, FWIW, here are the shbangs that I found:
```
   1 #!/bin/env python
   5 #!/usr/bin/python
   1 #! /usr/bin/env python
  55 #!/usr/bin/env python
```

ndokos:
The ones under `/opt/pbench-agent/{bin,tool-scripts,util-scripts}` are properly "mangled" when the RPM is created: the mangling tool seems to handle `bin`-type directories, but it leaves `/opt/pbench-agent/lib` alone: it doesn't touch those. That is not to say that we should not be consistent of course.

PR #2918, opened 2022-06-25
Add support for SHA256 sosreports

portante:


dbutenhof (approved):
That simple, eh? Convenient that we have `md5sum` and `sha256sum` for consistency.

npalaska (approved):

npalaska:
Should we also think about supporting `sha256sum` for the dataset upload operation?

server/bin/pbench-copy-sosreports.sh
```diff
@@ -109,31 +109,35 @@ for result in $list ;do
...
-        chmod 444 $x.md5
-        md5=$(md5sum $x)
-        md5=${md5%% *}
-        if [ "$md5" != $(cat $x.md5) ] ;then
-            log_error "$TS: FAILED: MD5 does not match for  $x"
-            nerrs=$nerrs+1
+        chmod 444 ${x}.${suffix}
+        val=$(${suffix}sum ${x})
+        val=${val%% *}
+        if [[ "${val}" != $(cat ${x}.${suffix}) ]]; then
```
webbnh, line 127: `md5sum` and `sha256sum` will do the check for you if you like, but, if we're going to be paranoid, we need to check that the checksum file we've got matches, so we could do something like this:
```
        if ! grep -q ${x} ${x}.${suffix} || ! ${suffix}sum --status -c ${x}.${suffix} ; then
```

Otherwise, you can use `$(< ${x}.${suffix})` instead of `$(cat ${x}.${suffix})`

server/bin/pbench-copy-sosreports.sh
```diff
@@ -109,31 +109,35 @@ for result in $list ;do
...
+            nerrs=${nerrs}+1
             continue
         fi
         # copy them to a temp dir
-        cp "$x" "$x".md5 $TMPDIR/
-        sts=$?
-        if [[ $sts -ne 0 ]] ;then
-            log_error "$TS: FAILED: cp "$x" "$x".md5 $TMPDIR/ - code $sts"
-            nerrs=$nerrs+1
+        cp "${x}" "${x}".${suffix} ${TMPDIR}/
```
webbnh, line 133: If we're worried about embedded whitespace (et al.) in the filename here, then we should worry about it at lines 114 and 124-127, as well.

server/bin/pbench-copy-sosreports.sh
```diff
@@ -109,31 +109,35 @@ for result in $list ;do
...
             continue
         fi
-        chmod 444 $x.md5
-        md5=$(md5sum $x)
-        md5=${md5%% *}
-        if [ "$md5" != $(cat $x.md5) ] ;then
-            log_error "$TS: FAILED: MD5 does not match for  $x"
-            nerrs=$nerrs+1
+        chmod 444 ${x}.${suffix}
+        val=$(${suffix}sum ${x})
```
webbnh, line 125: Composing the command from the suffix is cute, but it would probably be better to assign the command to variable when we assign `suffix`.

webbnh (approved):
I found a few nits, but, since Peter isn't around to address them, I'll "pre-resolve" the conversations, and he can deal with them later.

webbnh:
I just noticed that this PR is targeting the `b0.69` branch...shouldn't this change be applied first to the `main` branch and then backported to v0.69 (and v0.71)??

riya-17 (approved):

portante:
> I just noticed that this PR is targeting the `b0.69` branch...shouldn't this change be applied first to the `main` branch and then backported to v0.69 (and v0.71)??

I'd rather remote the separate sosreport processing and replace it with what PR #2933 is heading towards with @rmohnani's work.  This is so that I can facilitate updating our existing pbench server deployment a bit easier.

webbnh:
So, v0.71 doesn't require a fix for this any time soon?  ([PBENCH-787](https://issues.redhat.com/browse/PBENCH-787), which is still open, BTW, doesn't reference anything beyond v0.69.

PR #2920, opened 2022-06-26
Style changes and Elimination of timeout errors on execution

rmohnani:
Convert multiple try-except blocks to 1 try-except block in a loop to avoid code duplication
Change mode of all to_csv method calls to 'w' so files are overwritten each execution - eliminating the problem of non-uniqueness that was arising after multiple executions.

[23-06-2022.md](https://github.com/distributed-system-analysis/pbench/files/8987479/23-06-2022.md)
@portante

### DIFF
--- a/contrib/analysis/merge.py
+++ b/contrib/analysis/merge.py
@@ -39,10 +39,10 @@ def main(args):
     thr_df = df[df["sample.measurement_type"] == "throughput"]
 
     # Covert dataframes to csv files
-    slat_df.to_csv(r"latency_slat.csv", sep=";", mode="a")
-    clat_df.to_csv(r"latency_clat.csv", sep=";", mode="a")
-    lat_df.to_csv(r"latency_lat.csv", sep=";", mode="a")
-    thr_df.to_csv(r"throughput_iops_sec.csv", sep=";", mode="a")
+    slat_df.to_csv(r"latency_slat.csv", sep=";", mode="w")
+    clat_df.to_csv(r"latency_clat.csv", sep=";", mode="w")
+    lat_df.to_csv(r"latency_lat.csv", sep=";", mode="w")
+    thr_df.to_csv(r"throughput_iops_sec.csv", sep=";", mode="w")
 
     return 0
 
--- a/contrib/analysis/merge_sos_and_perf_parallel.py
+++ b/contrib/analysis/merge_sos_and_perf_parallel.py
@@ -55,7 +55,7 @@ def es_data_gen(es, index, doc_type):
         index=index,
         doc_type=doc_type,
         scroll="1d",
-        request_timeout = 3600, # to prevent timeout errors (3600 is arbitrary)
+        request_timeout=3600,  # to prevent timeout errors (3600 is arbitrary)
     ):
         yield doc
 
@@ -178,7 +178,7 @@ def extract_clients(results_meta, es):
         index=run_index,
         doc_type="pbench-run-toc-entry",
         scroll="1m",
-        request_timeout = 3600, # to prevent timeout errors (3600 is arbitrary)
+        request_timeout=3600,  # to prevent timeout errors (3600 is arbitrary)
     ):
         src = doc["_source"]
         if src["parent"] == parent_dir_name:
@@ -333,7 +333,9 @@ def transform_result(source, pbench_runs, results_seen, stats):
 
     # optional workload parameters accounting for defaults if not found
 
-    result["benchmark.filename"] = sentence_setify(benchmark.get("filename", "/tmp/fio"))
+    result["benchmark.filename"] = sentence_setify(
+        benchmark.get("filename", "/tmp/fio")
+    )
     result["benchmark.iodepth"] = benchmark.get("iodepth", "32")
     result["benchmark.size"] = sentence_setify(benchmark.get("size", "4096M"))
     result["benchmark.numjobs"] = sentence_setify(benchmark.get("numjobs", "1"))
@@ -344,7 +346,8 @@ def transform_result(source, pbench_runs, results_seen, stats):
 
     return result
 
-def sentence_setify(sentence : str) -> str:
+
+def sentence_setify(sentence: str) -> str:
     """Splits input by ", " gets rid of duplicates and rejoins unique
     items into original format. Effectively removes duplicates in input.
     """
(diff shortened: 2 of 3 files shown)


contrib/analysis/merge_sos_and_perf_parallel.py
```diff
@@ -329,44 +331,39 @@ def transform_result(source, pbench_runs, results_seen, stats):
...
+    for opt_param in optional_workload_params_defaults.keys():
+        try:
+            if opt_param in params_to_setify:
+                result[f"benchmark.{opt_param}"] = sentence_setify(benchmark[opt_param])
+            else:
+                result[f"benchmark.{opt_param}"] = benchmark[opt_param]
+        except KeyError:
+            result[f"benchmark.{opt_param}"] = optional_workload_params_defaults[opt_param]
 
     return result
```
webbnh, line 357: I agree that this code could use some tightening, but I'm not crazy about using a loop for that, particularly with the extra conditional inside it.  How about this, instead:
```
    result["benchmark.filename"] = sentence_setify(benchmark.get("filename", "/tmp/fio"))
    result["benchmark.iodepth"] = benchmark.get("iodepth", "32")
    result["benchmark.numjobs"] = sentence_setify(benchmark.get("numjobs", "1"))
    result["benchmark.ramp_time"] = benchmark.get("ramp_time", "none")
    result["benchmark.runtime"] = benchmark.get("runtime", "none")
    result["benchmark.size"] = sentence_setify(benchmark.get("size", "4096M"))
    result["benchmark.sync"] = benchmark.get("sync", "none")
    result["benchmark.time_based"] = benchmark.get("time_based", "none")
```
You could probably do better that that where the calls to `sentence_setify()` are made on default values, but you would have to make the code more complex.

contrib/analysis/merge_sos_and_perf_parallel.py
```diff
@@ -329,44 +331,39 @@ def transform_result(source, pbench_runs, results_seen, stats):
...
 
+def sentence_setify(sentence):
+    '''
+    input: string
+    Splits input by ", " gets rid of duplicates and rejoins unique
+    items into original format. Effectively removes duplicates in input.
+    '''
```
webbnh, line 364: See PEP-8 for recommendations on how to do [function docstrings](https://peps.python.org/pep-0008/#documentation-strings).  This should be more like:

```
    """Splits input by ", " gets rid of duplicates and rejoins unique
    items into original format. Effectively removes duplicates in input.

    input: string
    """
```
It would be good to include a "returns" line as well.  However, given that the input and output of this function is so straightforward, if you add type hints to the function definition, you can omit the "input" and "returns" lines from the docstring.  (That is, if the docstring doesn't say anything more than what the type hints do, then just omit those bits from the docstring.)

contrib/analysis/merge_sos_and_perf_parallel.py
```diff
@@ -329,44 +331,39 @@ def transform_result(source, pbench_runs, results_seen, stats):
...
+            if opt_param in params_to_setify:
+                result[f"benchmark.{opt_param}"] = sentence_setify(benchmark[opt_param])
+            else:
+                result[f"benchmark.{opt_param}"] = benchmark[opt_param]
+        except KeyError:
+            result[f"benchmark.{opt_param}"] = optional_workload_params_defaults[opt_param]
 
     return result
 
+def sentence_setify(sentence):
```
webbnh, line 359: It's good to get in the habit of using "type hints" on function definitions.  (It's not necessary to convert existing code, but when you add new code or make substantial changes to existing code, it's good to add them.)  E.g.,
```
def sentence_setify(sentence: str) -> str:
```

webbnh (approved):
Looks good, Rohit!

I have an alternate suggestion for getting rid of the `try` blocks, though.  And, I have a couple of comments on documenting functions for you.

contrib/analysis/merge_sos_and_perf_parallel.py
rmohnani: Thanks for the tip, I've made the changes and will definitely be sure to have them in the future!

contrib/analysis/merge_sos_and_perf_parallel.py
rmohnani: Thanks for the reference to documentation conventions, I changed the format to match as you suggested, and will be following it more closely from here!

contrib/analysis/merge_sos_and_perf_parallel.py
rmohnani: I initially thought that the code looked like it was being copy pasted with the same format but just the workload param changing, so figured a loop would be better to condense it. But I see now, that trying to do a loop or probably anything else fancier forces the use of more conditionals which are probably unnecessary and consume more time. I made the changes you suggested

webbnh (dismissed):
👍

portante (dismissed):
Looks good, just need to run `black` on everything.

contrib/analysis/merge.py
```diff
@@ -39,10 +39,10 @@ def main(args):
...
 
     # Covert dataframes to csv files
-    slat_df.to_csv(r"latency_slat.csv", sep=";", mode="a")
-    clat_df.to_csv(r"latency_clat.csv", sep=";", mode="a")
-    lat_df.to_csv(r"latency_lat.csv", sep=";", mode="a")
-    thr_df.to_csv(r"throughput_iops_sec.csv", sep=";", mode="a")
+    slat_df.to_csv(r"latency_slat.csv", sep=";", mode="w")
+    clat_df.to_csv(r"latency_clat.csv", sep=";", mode="w")
+    lat_df.to_csv(r"latency_lat.csv", sep=";", mode="w")
+    thr_df.to_csv(r"throughput_iops_sec.csv", sep=";", mode="w")
```
npalaska, line 45: I don't have a context but are you sure that these files only get written once and if they do get overwritten, we don't care about the first copy of it?

rmohnani: They should only get written once - after all the processing is done these are the final output files. The error I was getting was when performing multiple executions of the script, the files were being appended to causing duplicates which was a problem. So having it be overwritten is desirable at least at this stage.

webbnh (approved):
Looks good to me.

portante (approved):

PR #2934, opened 2022-07-06
Correct CentOS 7 container builds, update defaults

portante:
We have restored the ability to build CentOS 7-based containers.

The default container distributions supported are now CentOS 7 (restored), CentOS 8 (no change), CentOS 9 (added), and Fedora 35 (no change).  We removed Fedora 34 which recently was declared EOL.

### DIFF
--- a/agent/containers/images/Dockerfile.base.j2
+++ b/agent/containers/images/Dockerfile.base.j2
@@ -1,6 +1,6 @@
 {% set is_centos_8 = true if image_name == 'centos' and image_rev == '8' else false %}
 {% set is_centos_9 = true if image_name == 'centos' and image_rev == '9' else false %}
-{% set image_tag = 'stream' + image_rev if image_name == 'centos' else image_rev %}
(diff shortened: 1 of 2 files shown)


agent/containers/images/Dockerfile.base.j2
```diff
@@ -1,6 +1,6 @@
 {% set is_centos_8 = true if image_name == 'centos' and image_rev == '8' else false %}
 {% set is_centos_9 = true if image_name == 'centos' and image_rev == '9' else false %}
-{% set image_tag = 'stream' + image_rev if image_name == 'centos' else image_rev %}
+{% set image_tag = 'stream' + image_rev if image_name == 'centos' and image_rev != '7' else image_rev %}
```
dbutenhof, line 3: So `image_tag` is `stream8` for CentOS 8 and `stream9` for CentOS 9, but just `7` for CentOS 7?? Or am I missing some magic here? 😦

webbnh (approved):
👍

ndokos (approved):

PR #2935, opened 2022-07-07
Enhancements to the CI pipeline, part 1

webbnh:
This PR is the first installment in a series of enhancements to the Pbench CI pipeline.  The goal is to have the CI perform lint and style checks, run unit tests, build RPMs, build containers, and run functional tests for each of the Agent, Server, and Dashboard.  This PR provides the first several of those, up through building RPMs for the Agent and Server (the Dashboard is expected to be deployed using a different mechanism).  It also enables the Cobertura report to find and display the source lines!

For convenience, this PR consists of a series of commits, but they should be squashed into a single commit when it is merged (NOTE:  this PR is now based on #2938 and #2939, so their respective commits will appear in this PR until they are merged):
- enhancements to the CI pipeline
- tweaks to the CI dockerfile, adding dependencies of the extended pipeline features
- tweaks to the Dashboard environment for running ESLint
- tweaks to the Server RPM build to limit its default build to RHEL-9
- tweaks to the Dashboard code to make it pass ESlint

Highlights of the changes:
- Add a new top-level file, `build.sh` which executes the steps outlined above.  The CI executes this script in the CI container via the `jenkins/run` command.  The intention is that developers should be able to run the script as well, either natively or via a `jenkins/run` command in their own environment.  The script does the following:
  - extract credentials from environment variables and from a local COPR config file
  - install the Python and Node.js requirements for running the Agent, Server, and Dashboard linters and the Dashboard unit tests
  - run the various linters
  - run the various unit tests
  - use COPR (either the internal or the external one) to build the RPMs for the Agent and the Server
  - declare victory (there is latent code for building and pushing the Agent container images and the Pbench Server-in-a-Can container images, but it doesn't work inside the CI container, currently)
- Add support for a "CI" tag for building Agent containers, to keep them separate from the normally published containers.
- Add some configuration files for ESLint, the Dashboard lint program.
- Make some tweaks to the Dashboard files to enable them to pass the lint checks and unit tests.  (I'm hoping that these will come in via another PR from the Dashboard folks...but, if that doesn't happen, then they are here, so that the pipeline won't fail.)
- Tweak the Jenkins pipeline definition:
  - add environment variables/secrets for the credentials for accessing COPR and the container registry
  - alphabetize the environment variable list, now that it is long
  - change the invocation from running `tox` directly to running the new `build.sh` script
  - add (and remove) a symlink which enables the Cobertura report to find the sources properly
- Remove the responsibility for running the Python linters from Tox, now that it is handled by `build.sh`, and add to the dependences for the linters in the CI container image definition.
- Add the dependencies for building the RPMs and container images to the CI container image definition.
- Modify the `jenkins/run` script to add a mapping for the COPR config file from the Jenkins secret file (or the user's file) to the conventional file inside the container.  Also, add a volume for the `${HOME}/.config` directory (so the `npm` install inside the container can write it), and reorder the volume options in the container invocation from top to bottom (i.e., `${HOME}` first, then `${HOME}/.config`, and so on).
- Tweak the Server RPM makefile to set the default build to (only) RHEL-9.
- Change the common RPM makefile to explicitly specify a COPR config file for credentials, and add a `copr-ci` target.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -60,8 +60,8 @@ export PATH=${HOME}/.local/bin:${PATH}
 python3 -m pip install --user -r lint-requirements.txt
 
 # If this script is run in a container and the user in the container doesn't
-# match the owner of the Git checkout, then Git objects; these config settings
-# prevent the complaints.
+# match the owner of the Git checkout, then Git issues an error; these config
+# settings avoid the problem.
 GITTOP=$(git rev-parse --show-toplevel 2>&1 | head -n 1)
 if [[ ${GITTOP} =~ "fatal: unsafe repository ('/home/root/pbench'" ]] ; then
 	git config --global --add safe.directory /home/root/pbench
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,112 @@
+#!/bin/bash -e
+
+# This script drives the various tasks involved in testing and building the
+# various artifacts for the Pbench product.  It is intended to be run from the
+# root directory of a Git branch checkout.
+
+# The following variables will normally be defined as credentials in the Jenkins
+# environment.  For developer use of this script, they should be defined in the
+# shell environment; otherwise, they will be given default values, here.
+#
+#  COPR_CONFIG - the path to the configuration file for the COPR account
+#  PODMAN_USR - the username for the container registry account
+#  PODMAN_PSW - the password for the container registry account
+#
+PODMAN_USR=${PODMAN_USR-"pbench+mr_jenkins"}
+PODMAN_PSW=${PODMAN_PSW-${DEFAULT_PODMAN_PASSWORD}}
+COPR_CONFIG=${COPR_CONFIG:-"${HOME}/.config/copr"}
+
+
+# Function to read an initialization/configuration file and print the value of
+# the specified key in the specified section.  Borrowed from
+# https://stackoverflow.com/questions/49399984/parsing-ini-file-in-bash/49400353
+function getIni() {
+  local f=${1}  # Initialization/configuration file name
+  local s=${2}  # Section name
+  local o=${3}  # Option name
+
+  if [[ ! -r ${f} ]]; then
+    echo "Missing or bad COPR config file:  ${f}" >&2
+    exit 1
+  fi
+
+  local rv=$(awk '/^\[.*\]$/{obj=$0}/=/{print obj $0}' ${f} | \
+    sed -En "/${s}]${o}/s/^.*=[[:space:]]*//p" | \
+    tail -n 1)
+
+  if [[ -z ${rv} ]]; then
+    echo "Missing '${o}' in COPR config file, '${f}'" >&2
+    exit 1
+  fi
+
+  echo ${rv}
+}
+
+
+COPR_USER=$(getIni ${COPR_CONFIG} copr-cli username)
+COPR_URL=$(getIni ${COPR_CONFIG} copr-cli copr_url)
+
+if [[ ${COPR_URL} =~ "copr.fedorainfracloud.org" ]]; then
+  URL_PREFIX=https://download.copr.fedorainfracloud.org/results/${COPR_USER}
+elif [[ ${COPR_URL} =~ "copr.devel.redhat.com" ]]; then
+  URL_PREFIX=http://coprbe.devel.redhat.com/results/${COPR_USER}
+else
+  echo "Unexpected COPR_URL: '${COPR_URL}'" >&2
+  exit 1
+fi
+
+# Install the linter requirements and add them to the PATH.
+export PATH=${HOME}/.local/bin:${PATH}
+python3 -m pip install --user -r lint-requirements.txt
+
+# If this script is run in a container and the user in the container doesn't
+# match the owner of the Git checkout, then Git objects; these config settings
+# prevent the complaints.
+GITTOP=$(git rev-parse --show-toplevel 2>&1 | head -n 1)
+if [[ ${GITTOP} =~ "fatal: unsafe repository ('/home/root/pbench'" ]] ; then
+	git config --global --add safe.directory /home/root/pbench
+	git config --global --add safe.directory /home/root/pbench/agent/stockpile
+	GITTOP=$(git rev-parse --show-toplevel)
+fi
+
+# Install the Dashboard dependencies, including the linter's dependencies and
+# the unit test dependencies.
+( cd dashboard && npm install )
+
+# Test for code style and lint
+black --check .
+flake8 .
+( cd dashboard && npx eslint "src/**" --max-warnings 0 )
+
+# Run unit tests
+tox                                     # Agent and Server unit tests and legacy tests
+( cd dashboard && CI=true npm test )    # Dashboard unit tests
+
+# Build Agent and Server RPMs (Server RPM includes the Dashboard) on COPR
+( cd agent/rpm && make COPR_USER=${COPR_USER} COPR_CONFIG=${COPR_CONFIG} copr-ci )
+( cd server/rpm && make COPR_USER=${COPR_USER} COPR_CONFIG=${COPR_CONFIG} copr-ci )
+
+# The following steps don't work inside the CI container, so just declare
+# victory now.  (Building a container inside a container requires a
+# specially-configured and/or -invoked container, and we don't have that
+# arranged, yet.)
+exit 0
+
+################################################################################
+#  Future extensions
+################################################################################
+
+# Need to log into the registry to access the Server base image, etc.
+podman login -u="${PODMAN_USR}" -p="${PODMAN_PSW}" quay.io
+
+# Build the Agent container images and the Pbench-in-a-Can Server image
+( cd agent/containers/images && \
+  make COPR_USER=${COPR_USER} URL_PREFIX=${URL_PREFIX} TEST=ci \
+  )
+server/pbenchinacan/container-build.sh
+
+# Push the Agent and Server container images to the container registry.
+( cd agent/containers/images && \
+  make COPR_USER=${COPR_USER} URL_PREFIX=${URL_PREFIX} TEST=ci tag-ci push-ci \
+  )
+podman push quay.io/pbench/pbenchinacan-pbenchserver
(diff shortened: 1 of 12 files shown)


build.sh
```diff
@@ -0,0 +1,130 @@
...
+  make COPR_USER=${COPR_USER} URL_PREFIX=${URL_PREFIX} TEST=ci tag-ci push-ci \
+  )
+podman push quay.io/pbench/pbenchinacan-pbenchserver
+
+# FIXME:  We need a consistent set of container images for the Pbench Server,
+# PostgreSQL, and Elasticsearch, in terms of the data that they contain.  If
+# we're going to go with an "empty" Server, then we need to ensure that the
+# images on Quay.io for the PostgreSQL and Elasticsearch containers are empty,
+# as well, but we don't need to push them every time we rebuild the Server;
+# otherwise, we need to commit the seeded images for all three and push them up.
```
dbutenhof, line 127: Yeah; maybe start out with empty backend DBs. We can unpack a few of the legacy test tarballs (or just make a few new ones) and build a simple script to `PUT` them to the new server, let them be indexed, and then do some queries.

build.sh
```diff
@@ -0,0 +1,130 @@
...
+flake8 .
+( cd dashboard && npx eslint "src/**" --max-warnings 0 )
+
+# Run unit tests
+tox                                     # Agent and Server unit tests and legacy tests
+( cd dashboard && CI=true npm test )    # Dashboard unit tests
+
+# Build Agent and Server RPMs (Server RPM includes the Dashboard) on COPR
+#
+# TODO:  We should set up or maintain a ./seqno file.
```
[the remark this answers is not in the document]
webbnh: To be honest, I haven't fully contemplated how this thing is going to work on the COPR end.  At this point, concurrent PRs will get mixed up with each other.  And, the builds are going to "build up" on COPR, too, as we have no mechanism for clearing out old ones, presently (although, I think we can make them self-delete with a short expiration time).

The fact that the workspace is not recreated would potentially help out.  That would mean that each subsequent rerun of a particular PR would get a new sequence number, which would help considerably once we start pulling the RPMs.  And, it _might_ mean that the `main` would be properly sequenced without us having to deal with having a `seqno` file in the repo, which would be very cool.

So, yeah, there is definitely a couple of TODO items here.  Fortunately, they don't need to gate this PR.

build.sh
webbnh: I've already dealt with the problem of pre-loading data into the canned Server.  I just have to enshrine the process in code here.  Obviously it would be simpler to spin up an empty Server, as doing so would allow us to skip the pre-load steps, but the preloading isn't hard to do.  What we should do depends on what we want the functional testing to look like and at what point we want it to start from.

Given that the functional testing is presumably going to load data, my thinking was that there's probably no point (in this context) to pre-loading data....

build.sh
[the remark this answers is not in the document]
webbnh: I've added a `seqno` file, and it seems to work.

I looked on COPR, and there is a way to mark a project for auto-deletion, but I don't think we want that:  we want to delete individual builds, not the whole project.  😢  And, I don't see a way to do that...so this may need some attention in the future.

build.sh
```diff
@@ -0,0 +1,130 @@
...
+# Need to log into the registry to access the Server base image, etc.
+podman login -u="${PODMAN_USR}" -p="${PODMAN_PSW}" quay.io
+
+# Build the Agent container images and the Pbench-in-a-Can Server image
+( cd agent/containers/images && \
+  make COPR_USER=${COPR_USER} URL_PREFIX=${URL_PREFIX} TEST=ci \
+  )
+server/pbenchinacan/container-build.sh
+
+# TODO:  do we need/want to pre-load sample data into the canned Server?
```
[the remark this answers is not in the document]
webbnh: Nevermind, your subsequent comment convinced me:  we should _not_ have pre-populated containers for the CI's use -- once we have the containers, the next step will be to launch the functional tests which, among other things, will load data into them, so spending time pre-loading them will just be a waste.

webbnh (commented):
OK, it seems to be working fairly well, so I think it's ready for review!  Thanks!!  (And, see you Monday....)

jenkins/Pipeline.gy
```diff
@@ -25,6 +39,14 @@ pipeline {
     }
     post {
         success {
+            // This symlink somehow allows the Cobertura plug-in to find the
+            // sources referenced in the coverage report.  However, the presence
+            // of this link inside the container seems to confuse the coverage
+            // data collection, so we create it here and then remove it after
+            // generating the report. (We use the --force option just in case
```
dbutenhof, line 46: You probably shouldn't *say* you're using `--force` when you're *actually* using `-f` ... or perhaps you should actually use `--symbolic --no-target-directory --force`?

build.sh
```diff
@@ -0,0 +1,112 @@
...
+# the unit test dependencies.
+( cd dashboard && npm install )
+
+# Test for code style and lint
+black --check .
+flake8 .
+( cd dashboard && npx eslint "src/**" --max-warnings 0 )
+
+# Run unit tests
+tox                                     # Agent and Server unit tests and legacy tests
```
dbutenhof, line 82: I'm not sure what's going on here, but somehow _you're only running the "agent" tests_. I glanced through the Cobertura report and noticed 0% coverage for all the server code, and the console log confirms this.

Cobertura is also voluminously complaining about all the (pypi) module source files it can't find, which is annoying. I don't recall seeing these before when it couldn't find *any* of the source files, so I'm not sure what changed.

dbutenhof (changes requested):

build.sh
```diff
@@ -0,0 +1,112 @@
...
+
+# Build Agent and Server RPMs (Server RPM includes the Dashboard) on COPR
+( cd agent/rpm && make COPR_USER=${COPR_USER} COPR_CONFIG=${COPR_CONFIG} copr-ci )
+( cd server/rpm && make COPR_USER=${COPR_USER} COPR_CONFIG=${COPR_CONFIG} copr-ci )
+
+# The following steps don't work inside the CI container, so just declare
+# victory now.  (Building a container inside a container requires a
+# specially-configured and/or -invoked container, and we don't have that
+# arranged, yet.)
+exit 0
```
portante, line 93: Why do we need everything performed in one container?

Why can't the pipeline be broken up into steps, some containerized, some not?

We can also have Jenkins start a particular container environment for us instead of running one monolithic one.

build.sh
```diff
@@ -0,0 +1,112 @@
...
+fi
+
+# Install the Dashboard dependencies, including the linter's dependencies and
+# the unit test dependencies.
+( cd dashboard && npm install )
+
+# Test for code style and lint
+black --check .
+flake8 .
+( cd dashboard && npx eslint "src/**" --max-warnings 0 )
```
portante, line 79: This change means we don't get these checks in local tox runs.  Isn't that bad?

jenkins/ci.fedora.Dockerfile
```diff
@@ -71,7 +71,27 @@ RUN \
...
+        `#` \
+        black \
+        npm \
+        python3-flake8 \
+        `#` \
+        `# Required for building RPMs` \
+        `#` \
+        python3-jinja2-cli \
+        rpmlint \
+        http://hdn.corp.redhat.com/rhel8-csb/RPMS/noarch/redhat-internal-cert-install-0.1-25.el7.noarch.rpm \
```
portante, line 85: Why is this in the public repo?

build.sh
```diff
@@ -0,0 +1,112 @@
...
+
+# Run unit tests
+tox                                     # Agent and Server unit tests and legacy tests
+( cd dashboard && CI=true npm test )    # Dashboard unit tests
```
portante, line 83: Why not add this to tox?

portante (changes requested):
Let's move the coverage fix to a separate PR.

build.sh
webbnh: I'll look into what's going on with the Server tests.  However, take a look at the log for [PR-2891](http://jenkins.intlab.perf-infra.lab.eng.rdu2.redhat.com/job/pbench-ci-multi/view/change-requests/job/PR-2891/17/console) (which I think would be familiar to you 😉) -- isn't it full of the same Cobertura complaints?

I agree that the diagnostics in the log are annoying; even worse are all of the non-project lines in the coverage report.  I assume that there is some way to get Cobertura to report only on _our_ code, but I haven't looked into that.  (Just getting the source working in the report was enough of an achievement for the nonce!)

build.sh
webbnh: I believe that we should have everything performed by one _command_.  Now whether that command runs a single script which invokes a series of containers (or whatnot), or whether than command invokes a single container which executes a series of actions is certainly open for discussion.  But, what I think we want to avoid is putting lots of stages in the `Pipeline.gy` file, because then it won't be feasible for developers to run "the CI tests" locally.

The advantage of running it all in a single container (or, perhaps, running _each_ of "The Things" in containers) is that we (collectively and individually) don't have to be concerned about what the respective environments look like.  And, if we _could_ run it all in one container, then that would greatly reduce the amount of administration required to run the test pipeline.

But, no, it doesn't have to be that way.  It just seemed, at the outset, like a good idea.  I would be happy to hear your thoughts on the matter.

build.sh
webbnh: Why should it be?  See my [previous](https://github.com/distributed-system-analysis/pbench/pull/2935#discussion_r918053542).

jenkins/Pipeline.gy
webbnh: I addressed this in #2938.

jenkins/ci.fedora.Dockerfile
webbnh: Well...I guess that's a good question.

The logic goes like, "The CI container needs those certs, so they are installed as part of the container recipe, which we currently store in the public repo...."  But, yes, that's an internal network name, reachable only through the tunnel.

So, how should we proceed?

webbnh:
This PR has been superseded by PRs #2946, #2982, and follow-ons.

PR #2936, opened 2022-07-08
Remove ansible roles and playbooks for server installation

portante:
The Pbench Server will not be using ansible playbooks for it setup and installation.  It is containerized.


dbutenhof (approved):
We do eventually need to cut off non-containerized server deployments, especially as we move towards a component model.

Since I haven't actually tried to follow the workflow to build/deploy my own development server as a container, the abruptness of this cutoff is a bit frightening ... but apparently it's working for Webb so it'll maybe work for me, too. 🙏🏻 🤞🏻 ⚔️ 🚸

dbutenhof (approved):

ndokos (approved):

webbnh (approved):
> We do eventually need to cut off non-containerized server deployments, especially as we move towards a component model.

Hurray for deleting code (or files...)!  🎆

(Dave, you can still do a manual, RPM-based install.... 😉  And, failing that, just write down the Git hash, and you can get these files back....  😆)

> Since I haven't actually tried to follow the workflow to build/deploy my own development server as a container, the abruptness of this cutoff is a bit frightening ... but apparently it's working for Webb so it'll maybe work for me, too. 🙏🏻 🤞🏻 ⚔️ 🚸

It _was_ working for Webb...but, y'know, if nobody uses it, the code rots....  But, if you want to see if it still works, here are the [instructions](https://docs.google.com/document/d/1k4zrnGNa8BdjcVvMWmFKpp26wGnXHpMJq6EA8kXOsnk/edit#heading=h.o3givyjig091).  😁

PR #2937, opened 2022-07-09
Use `major.minor` version in COPR repo name

portante:
PBENCH-823

The COPR rules are such that you can only have one RPM for a given chroot per repository.  So you can't have a `pbench-agent` `0.69.*` and a `pbench-agent` `0.71.*` RPM available at the same time.

### DIFF
--- a/agent/containers/images/Makefile
+++ b/agent/containers/images/Makefile
@@ -28,8 +28,10 @@
 #     all-dockerfiles:  build all default distro ".repo" and ".Dockerfile" files
 #
 
+_PBENCH_TOP := $(shell git rev-parse --show-toplevel)
+
 # Include definition of _DISTROS
-include ../utils/utils.mk
+include ${_PBENCH_TOP}/utils/utils.mk
 
 # By default we only build images for x86_64 architectures.
 _ARCH = x86_64
--- a/agent/containers/images/Makefile
+++ b/agent/containers/images/Makefile
@@ -50,7 +50,7 @@ URL_PREFIX = https://copr-be.cloud.fedoraproject.org/results/${COPR_USER}
 # the final repo name would be "pbench-test".
 TEST =
 _TEST_SUFFIX = $(if $(TEST),-$(TEST),"")
-_PBENCH_REPO_NAME = pbench${_TEST_SUFFIX}
+_PBENCH_REPO_NAME = pbench-$(shell grep -oE '[0-9]+\.[0-9]+' ${_PBENCH_TOP}/agent/VERSION)${_TEST_SUFFIX}
 
 # By default we use the pbench Quay.io organization for the image
 # repository.  You can override this default using an environment
--- a/utils/rpm.mk
+++ b/utils/rpm.mk
@@ -105,7 +105,7 @@ submodules:
 ${subcomps}:
 	make -C ${PBENCHTOP}/$@ DESTDIR=${TBDIR}/$@ install
 
-$(RPMSRPM)/$(prog)-$(version)-$(seqno)g$(sha1).src.rpm: srpm
+$(RPMSRPM)/$(prog)-$(VERSION)-$(seqno)g$(sha1).src.rpm: srpm
 
 ifdef COPR_USER
 _copr_user = ${COPR_USER}
@@ -115,7 +115,7 @@ endif
 
 COPR_TARGETS = copr copr-test
 .PHONY: ${COPR_TARGETS}
-${COPR_TARGETS}: $(RPMSRPM)/$(prog)-$(version)-$(seqno)g$(sha1).src.rpm
+${COPR_TARGETS}: $(RPMSRPM)/$(prog)-$(VERSION)-$(seqno)g$(sha1).src.rpm
 	copr-cli build ${CHROOTS} $(_copr_user)/$(subst copr,pbench,$@) $(RPMSRPM)/$(prog)-$(VERSION)-$(seqno)g$(sha1).src.rpm
 
 # Determine the present working directory relative to ${PBENCHTOP} so that we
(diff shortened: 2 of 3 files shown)


portante (commented):
This works.  The failure had to do with something in the COPR environment itself.

npalaska (dismissed):

utils/rpm.mk
```diff
@@ -68,7 +69,7 @@ submodules:
 ${subcomps}:
 	make -C ${PBENCHTOP}/$@ DESTDIR=${TBDIR}/$@ install
 
-$(RPMSRPM)/$(prog)-$(version)-$(seqno)$(sha1).src.rpm: srpm
+$(RPMSRPM)/$(prog)-$(version)-$(seqno)g$(sha1).src.rpm: srpm
```
webbnh, line 72: Make variable names are case-sensitive...so where does `$(version)` get defined?  (Should that be `$(VERSION)`, here and at line 82?  Note that the references in lines 82 and 83 don't match.)

webbnh (changes requested):
It looks to me like the lines where you added the missing `g` (good catch!) _also_ need to have their references to `$(version)` corrected.

ndokos (commented):
What do we do about the default value of `pbench_repo_name` in the `pbench_repo_install` role?

utils/rpm.mk
ndokos: There is a comment in the rpm.mk on b0.69, saying "version and sha1 have to be provided by the including Makefile" - although the b0.69 Makefile does not do that either.

ndokos: All I can say is that since the same mistake was made in two different places (the "g" as well as the undefined version), it all came out right in the end :-) Just a bit of unnecessary rebuilding...

webbnh: OK, so the targets `copr` and `copr-test` (and, soon, `copr-ci`) depend on `$(RPMSRPM)/$(prog)-$(version)-$(seqno)$(sha1).src.rpm` (to which Peter is adding a `g`) which in turn depends on the `srpm` target which, after removing `${RPMSRPM}/$(prog)-*.src.rpm` (which is probably too broad a wildcard...), invokes `rpmbuild -bs ${RPMSPEC}/${prog}.spec` which apparently produces `$(RPMSRPM)/$(prog)-$(VERSION)-$(seqno)g$(sha1).src.rpm` (which already has a `g` and a `$(VERSION)` component in its name).

So, without Peter's change, it all works because the `copr-cli` command finds the SRPM which `rpmbuild` produced despite the lack of a `$(version)` definition (and the missing `g`).  Peter's change (plus the up-casing of the `$(version)` reference) might allow Make to skip over rebuilding the SRPM if Make thinks that it's up-to-date; without those changes, I imagine that the process will continue to work as it has done, rebuilding the SRPM every time because Make can never find the one that we specified.  😈

(So, I'm not convinced that just adding the `g` is sufficient to actually change anything.)

ndokos: I believe you are correct.

portante: @webbnh, I believe this is now fixed.

webbnh (approved):
👍

dbutenhof (dismissed):

portante:
> What do we do about the default value of `pbench_repo_name` in the `pbench_repo_install` role? Delete it and require that it be set explicitly in the inventory?

Nice catch, @ndokos!  Turns out this will end up breaking the container builds.  A proposal to fix this is coming.

webbnh:
> Turns out this will end up breaking the container builds. A proposal to fix this is coming.

Can (should!) this additional fix be made under a separate PR??  (Does the fix in this PR break something (something which currently somehow works) without that fix?)

(This PR potentially affects stuff that I'm working on (namely, container builds...), and I'd like to get it integrated...and I don't see how ansible stuff comes into play with, well, what I'm working on, at least....)

portante:
> > Turns out this will end up breaking the container builds. A proposal to fix this is coming.
> 
> Can (should!) this additional fix be made under a separate PR?? (Does the fix in this PR break something (something which currently somehow works) without that fix?)

This set of changes will break container builds as it currently stands.

> 
> (This PR potentially affects stuff that I'm working on (namely, container builds...), and I'd like to get it integrated...and I don't see how ansible stuff comes into play with, well, what I'm working on, at least....)

Ansible stuff is not really the issue, it is the container build method.

portante:
Toying with a possible direction we can take as drafted in #2992.

ndokos (dismissed):

agent/containers/images/Makefile
```diff
@@ -48,7 +50,7 @@ URL_PREFIX = https://copr-be.cloud.fedoraproject.org/results/${COPR_USER}
 # the final repo name would be "pbench-test".
 TEST =
 _TEST_SUFFIX = $(if $(TEST),-$(TEST),"")
-_PBENCH_REPO_NAME = pbench${_TEST_SUFFIX}
+_PBENCH_REPO_NAME = pbench-$(shell grep -oE '[0-9]+\.[0-9]+' ${_PBENCH_TOP}/agent/VERSION)${_TEST_SUFFIX}
```
portante, line 53: I don't like the duplication here with `rpm.mk`, but neither could I figure out a way to resolve it gracefully.

portante (commented):
Ready for review.

agent/containers/images/Makefile
dbutenhof: I suppose you could unbundle it into a script you `$(shell script {path})` in both places, but I don't think it's that big a deal.

dbutenhof (approved):

agent/containers/images/Makefile
webbnh: I think it is probably best as is.

webbnh (dismissed):
👍

npalaska (approved):

ndokos (dismissed):

portante (commented):
No changes, had to re-order the commits to avoid a sequencing problem.

dbutenhof (approved):

ndokos (approved):

PR #2947, opened 2022-07-19
Delete the epel role from pbench_agent_install.yml

ndokos:
It should not be part of the installation: it is part of the preparation of the hosts and therefore belongs in a different playbook.


webbnh (approved):
🦐

npalaska (approved):
Yey!

PR #2948, opened 2022-07-19
Use the "epel" chroot for "Rocky Linux"

ndokos:
Set up a map with the `ansible_distribution' value  as key and the
COPR chroot name as value. Use the map to calculate the final
component of the baseurl field in the repo file (except for CentOS 7
which is one-of-a-kind).

The map can be extended easily for other distros, RHEL clones or
otherwise.

### DIFF
--- a/agent/ansible/pbench/agent/roles/pbench_repo_install/vars/main.yml
+++ b/agent/ansible/pbench/agent/roles/pbench_repo_install/vars/main.yml
@@ -1,9 +1,13 @@
 ---
-distro: "{{ 'centos-stream' if ansible_distribution == 'CentOS' and ansible_distribution_major_version > '7'
-          else
-            'epel' if ansible_distribution == 'RedHat' or ansible_distribution == 'CentOS'
-          else
-            'fedora' if ansible_distribution == 'Fedora'
-          else
-            '' }}"
-distrodir: "{{ distro }}-{{ ansible_distribution_major_version }}-$basearch"
+# Map distribution name to COPR chroot
+copr_chroots:
+  RedHat: epel
+  Fedora: fedora
+  CentOS: centos-stream
+  Rocky: epel
+
+# CentOS 7 is an exception
+copr_distro: "{{ 'epel' if ansible_distribution == 'CentOS' and ansible_distribution_major_version == '7'
+                 else
+                     copr_chroots[ansible_distribution] }}"
+distrodir: "{{ copr_distro }}-{{ ansible_distribution_major_version }}-$basearch"


dbutenhof (dismissed):
I thought YAML could handle quoted keys? (Though the documentation is slightly maddening.) Or is this a quirk specific to Ansible?

I.e., can't you avoid the special case with

```
  "sui generis": epel
```

agent/ansible/pbench/agent/roles/pbench_repo_install/vars/main.yml
```diff
@@ -1,9 +1,13 @@
...
-distrodir: "{{ distro }}-{{ ansible_distribution_major_version }}-$basearch"
+# Map distribution name to COPR chroot
+copr_chroots:
+  RedHat: epel
+  Fedora: fedora
+  CentOS: centos-stream
+  Rocky: epel
+
+# CentOS 7 is an exception
+copr_distro: "{{ 'epel' if ansible_distribution == 'CentOS' and ansible_distribution == '7'
```
webbnh, line 10: > `if ansible_distribution == 'CentOS' and ansible_distribution == '7'`

My eyes might be challenged, but I don't think `ansible_distribution` can ever be _both_ `CentOS` and `7`.

ndokos: Ugh.

ndokos:
> I thought YAML could handle quoted keys? (Though the documentation is slightly maddening.) Or is this a quirk specific to Ansible?
> 
> I.e., can't you avoid the special case with
> 
> ```
>   "sui generis": epel
> ```



> I thought YAML could handle quoted keys? (Though the documentation is slightly maddening.) Or is this a quirk specific to Ansible?
> 
> I.e., can't you avoid the special case with
> 
> ```
>   "sui generis": epel
> ```

I think you misunderstood: CentOS 7 just doesn't fit the pattern; it reports `CentOS` as the value of `ansible_distribution' (just as CentOS-Stream 8/9 do) but that key has `centos-stream' as its value in order to accommodate the later versions.

webbnh (approved):
Looks OK to me, but it looks like you've got a change from #2947 in here, so I'm not sure how the eventual merge(s) will proceed.

The branch for this PR should probably be rebased on the branch for that one, and then we should approve and merge them in sequence.

webbnh:
I've merged #2947, so my concerns above should be alleviated once this is rebased on `main`.

ndokos:
Crud - I'm not sure how that snuck in. I'll redo it against main: they can then go in independently.

dbutenhof (approved):

webbnh (approved):
👍

PR #2950, opened 2022-07-20
Improve handling of key validation errors

dbutenhof:
PBENCH-837

Setting the "dataset.name" and "server.deletion" keys can produce validation errors which the `PUT /datasets/metadata` API wasn't properly handling, so that the attempt fails with `INTERNAL_SERVER_ERROR` instead of `BAD_REQUEST`.

Add logic to pre-validate so that on validation failure we can fail before modifying any metadata values.

(FYI: I also experimented with the alternative of wrapping the key setting loop in a nested SQLAlchemy transaction that could be cleanly unrolled on error. I didn't manage to get that to work, but it's probably something we should consider trying again later.)

### DIFF
--- a/lib/pbench/server/api/resources/datasets_metadata.py
+++ b/lib/pbench/server/api/resources/datasets_metadata.py
@@ -165,7 +165,7 @@ class DatasetsMetadata(ApiBase):
         # Now update the metadata, which may occur in multiple SQL operations
         # across namespaces. Make a best attempt to update all even if we
         # encounter an unexpected error.
-        fail_count = 0
+        fail = False
         for k, v in metadata.items():
             native_key = Metadata.get_native_key(k)
             user: Optional[User] = None
@@ -175,9 +175,9 @@ class DatasetsMetadata(ApiBase):
                 Metadata.setvalue(key=k, value=v, dataset=dataset, user=user)
             except MetadataError as e:
                 self.logger.warning("Unable to update key {} = {!r}: {}", k, v, str(e))
-                fail_count += 1
+                fail = True
 
-        if fail_count:
+        if fail:
             raise APIAbort(HTTPStatus.INTERNAL_SERVER_ERROR)
         results = self._get_dataset_metadata(dataset, list(metadata.keys()))
         return jsonify(results)
--- a/lib/pbench/server/api/resources/datasets_metadata.py
+++ b/lib/pbench/server/api/resources/datasets_metadata.py
@@ -21,6 +21,7 @@ from pbench.server.api.resources import (
 )
 from pbench.server.database.models.datasets import (
     Metadata,
+    MetadataBadValue,
     MetadataError,
 )
 from pbench.server.database.models.users import User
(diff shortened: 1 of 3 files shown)


lib/pbench/server/api/resources/datasets_metadata.py
```diff
@@ -149,7 +150,22 @@ def _put(self, params: ApiParams, _) -> Response:
...
         self._check_authorization(str(dataset.owner_id), dataset.access, role)
 
+        # Validate the metadata key values in a separate pass so that we can
+        # fail before committing any changes to the database.
         failures = []
+        for k, v in metadata.items():
+            try:
+                Metadata.validate(dataset=dataset, key=k, value=v)
+            except MetadataBadValue as e:
+                failures.append(str(e))
```
webbnh, line 160: Should we be decorating the exception message with the key value (and, perhaps, the corresponding value value) before appending it to the list?  (That is, if we get failures on only a subset of the keys, the requestor would probably appreciate knowing which ones went bad.)

Also, it might not be worth it in this context, but in general we should consider de-duplicating the failure list in code like this.

Also, also, for the sake of testing, we might want to consider sorting the failures prior to returning them, so that they appear in a deterministic order.

lib/pbench/server/api/resources/datasets_metadata.py
```diff
@@ -159,8 +175,9 @@ def _put(self, params: ApiParams, _) -> Response:
                 Metadata.setvalue(key=k, value=v, dataset=dataset, user=user)
             except MetadataError as e:
                 self.logger.warning("Unable to update key {} = {!r}: {}", k, v, str(e))
-                failures.append(k)
-        if failures:
+                fail_count += 1
+
+        if fail_count:
```
webbnh, line 180: A counter is better than a list, but why not a simple boolean?

lib/pbench/server/database/models/datasets.py
```diff
@@ -1151,9 +1128,50 @@ def setvalue(dataset: Dataset, key: str, value: Any, user: Optional[User] = None
...
+            raise MetadataBadKey(key)
+        keys = key.lower().split(".")
+        native_key = keys.pop(0)
+        found = True
+
+        # Setting the dataset name is a direct modification to the Dataset SQL
+        # column, so do that first and exit without touching the Metadata
+        # table.
+        v = __class__.validate(dataset, key, value)
+        if key == __class__.DATASET_NAME:
```
webbnh, line 1171: It looks to me like line 1170 should precede the block comment, since the comment applies to the following line and 1170 applies to all cases.

lib/pbench/server/database/models/datasets.py
```diff
@@ -1151,9 +1128,50 @@ def setvalue(dataset: Dataset, key: str, value: Any, user: Optional[User] = None
...
+        "a": {"b": {"c": "bar"}}}
+
+        We handle two keys specially:
+
+            dataset.name: The primary resource name of a dataset, which can be
+                modified by the dataset's owner.
+            server.deletion: The expected automatic deletion date of the
+                dataset. It can be modified by the owner so long as the new
+                value is a valid date that falls within the maximum server
+                retention period from the dataset's upload timestamp.
```
webbnh, line 1154: `server.deletion` is no longer handled specially by this function; that has been delegated to `validate()`, so we probably shouldn't mention it here.

And, given that we have only one special case (which already has a nice block comment in the code), I would be inclined to remove this whole section of the docstring.

webbnh (commented):
Just a couple of nits and suggestions.

However, shouldn't we have unit tests for `Metadata.validate()`?  I realize that the higher level more-functional tests cover it, but I think we want to be moving away from that approach, and since `validate()` is new code and relatively self-contained, it would be a good candidate for unit testing.

lib/pbench/server/api/resources/datasets_metadata.py
dbutenhof: Note that `MetadataBadValue` *does* have those things, and this code appends the exception string for each error in the loop.

But, yeah; in theory the `.items()` doesn't have a fixed order, so sorting the list before doing the `join` might be a good idea.

lib/pbench/server/api/resources/datasets_metadata.py
dbutenhof: Because it goes to 11?

lib/pbench/server/database/models/datasets.py
dbutenhof: I knew you were going to say that. But I didn't know it until I'd already committed the code and I didn't feel like fixing it then.

webbnh (approved):
👍

riya-17 (approved):

PR #2951, opened 2022-07-22
Generalize the "dashboard" namespace

dbutenhof:
PBENCH-814

Originally  we had the "dashboard" namespace for global dataset metadata controlled (albeit only by convention) by the dashboard, and the "user" namespace for general use by "other clients".  It's not entirely clear that this separation (as it was only a loose convention) made much sense, and the semantic of "user" was misleading.

After repurposing the "user" namespace for a new user-specific metadata concept, it makes sense to change "dashboard" to something more general.

The API documentation will recommend that each client assign a key under the namespace; e.g., I think the dashboard should be using something like "global.dashboard.seen" rather than just "global.seen".

### DIFF
--- a/docs/API/metadata.md
+++ b/docs/API/metadata.md
@@ -112,9 +112,9 @@ within this namespace to any valid JSON values (string, number, boolean, list,
 or nested objects) for retrieval later. All clients with read access to the
 dataset will see the same values.
 
-The recommended best practice is to assign a project sub-key that will be unique
-and avoid collisions between distinct clients. The Pbench Dashboard project,
-for example, will store all client metadata under the `global.dashboard`
+The recommended best practice is to select a project sub-key that will be unique
+and minimize the risk of collisions between various clients. The Pbench Dashboard
+project, for example, will store all client metadata under the `global.dashboard`
 sub-namespace, for example `global.dashboard.seen`. A hypothetical client named
 "clienta" might use `global.clienta`, for example `global.clienta.configuration`.
 
@@ -137,9 +137,9 @@ sub-resource as long as the client has READ access to the dataset. See
 [Access model](./access_model.md) for general information about the Pbench
 Server access controls.
 
-The recommended best practice is to assign a project sub-key that will be unique
-and avoid collisions between distinct clients. The Pbench Dashboard project,
-for example, will store all client metadata under the `user.dashboard`
+The recommended best practice is to select a project sub-key that will be unique
+to minimize the risk of collisions between various clients. The Pbench Dashboard
+project, for example, will store all client metadata under the `user.dashboard`
 sub-namespace, for example `user.dashboard.favorite`. A hypothetical client
 named "clienta" might use `user.clienta`, for example `user.clienta.configuration`.
 
--- a/docs/API/metadata.md
+++ b/docs/API/metadata.md
@@ -23,19 +23,24 @@ based on the owner's retention policy and the server administrator's retention
 policy along with some other internal management context. The expected deletion
 date is accessible under the `server` metadata key namespace.
 
-Clients can also set arbitrary metadata through the "dashboard" and "user"
+Clients can also set arbitrary metadata through the "global" and "user"
 metadata namespaces:
-* The "dashboard" namespace can only be modified by the owner of the dataset,
+* The "global" namespace can only be modified by the owner of the dataset,
 and is visible to anyone with read access to the dataset.
 * The "user" namespace is private to each authenticated user, and even if you
 don't own a dataset you can set your own private "user" metadata to help you
 categorize that dataset and to find it again.
 
 Metadata namespaces are hierarchical, and are exposed as nested JSON objects.
-You can address an entire namespace, e.g., `dashboard` or `dataset` and
+You can address an entire namespace, e.g., `global` or `dataset` and
 retrieve the entire JSON object, or you can address nested objects or values
-using a dotted metadata key path like `dashboard.contact.email` or
+using a dotted metadata key path like `global.contact.email` or
 `dataset.metalog.pbench.script`.
+
+By convention, a Pbench Server client should create a sub-namespace to minimize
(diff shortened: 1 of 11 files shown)


docs/API/metadata.md
```diff
@@ -23,20 +23,25 @@ based on the owner's retention policy and the server administrator's retention
...
+You can address an entire namespace, e.g., `global` or `dataset` and
 retrieve the entire JSON object, or you can address nested objects or values
-using a dotted metadata key path like `dashboard.contact.email` or
+using a dotted metadata key path like `global.contact.email` or
 `dataset.metalog.pbench.script`.
 
+By convention, a Pbench Server client should create a sub-namespace to minimize
+the risk of key collisions within the `global` and `user` namespaces. The
+Pbench Dashboard client, for example, uses `global.dashboard.seen` and
+`user.dashboard.favorite`.
```
webbnh, line 43: I wonder if we should use a funkier spelling, like `pbdashboard` -- if all goes well, ours might not be the only "dashboard" client.

lib/pbench/server/api/resources/datasets_metadata.py
```diff
@@ -80,7 +80,7 @@ def _get(self, params: ApiParams, request: Request) -> Response:
             params: API parameters
             request: The original Request object containing query parameters
 
-        GET /api/v1/datasets/metadata?name=dname&metadata=dashboard.seen,server.deletion
+        GET /api/v1/datasets/metadata?name=dname&metadata=global.seen,server.deletion
```
webbnh, line 83: Given the comment [here](https://github.com/distributed-system-analysis/pbench/blob/ea1cc24f0ed2ac4c119702fb9ae22e846984b1d7/docs/API/metadata.md?plain=1#L42), this should presumably be `global.dashboard.seen`.  Ditto lines 104 and 118.

lib/pbench/server/database/models/datasets.py
```diff
@@ -792,22 +792,23 @@ class Metadata(Database.Base):
...
-
-    # DASHBOARD is arbitrary data saved on behalf of the dashboard client as a
-    # JSON document. Writing these keys requires ownership of the referenced
-    # dataset, and the data is visible to all clients with READ access to the
-    # dataset.
+    # that wrote the data. Both are represented as arbitrarily nested JSON
+    # objects accessible at any level by a dotted key path. For example, for a
+    # value of "global": {"favorite": true, "mine": {"contact": "me"}}, the
+    # value of key path "global" is the entire object, "global.favorite" is
+    # true, "global.mine" is {"contact": "me"}, and "global.mind.contact" is
```
webbnh, line 803: s/mind/mine/

lib/pbench/server/database/models/datasets.py
```diff
@@ -939,8 +942,8 @@ def create(**kwargs) -> "Metadata":
         for the referenced Dataset.
 
         NOTE: use this only for raw first-level Metadata keys, not dotted
-        paths like "dashboard.seen", which will fail key validation. Instead,
-        use the higher-level `set` helper.
+        paths like "global.seen", which will fail key validation. Instead,
```
webbnh, line 945: s/global.seen/global.dashboard.seen/ ?

or

s/global.seen/global.contact.name/ ?

Ditto line 1086.

lib/pbench/test/unit/server/test_datasets_metadata.py
```diff
@@ -105,7 +105,7 @@ def token(self, client, config: PbenchServerConfig, user: str) -> str:
     def test_get_no_dataset(self, query_get_as):
         response = query_get_as(
             "foobar",
-            {"metadata": ["dashboard.seen", "dashboard.saved"]},
+            {"metadata": ["global.seen", "global.saved"]},
```
webbnh, line 108: s/global/global.dashboard/g ?

Ditto for the rest of this module?

webbnh (commented):
Looks good -- I like the way this design has turned out.

However, we should probably be slightly more rigorous about what's in the `global` space and what's in the `global.dashboard` subspace.  (And, before we do that, we might want to decide how to name "the Pbench Dashboard subspace".)

lib/pbench/server/database/models/datasets.py
webbnh: The reason why I called these out is that I get a certain dissonance from code comments which reflect something other than current practice.  I fully appreciate that everything after the first dot is just theoretical or for example, but there is or will be existing practice in which `seen` is a key which is not at the top level, and so it will make the documentation look "funny".  And since the [documentation](https://github.com/distributed-system-analysis/pbench/blob/ea1cc24f0ed2ac4c119702fb9ae22e846984b1d7/docs/API/metadata.md?plain=1#L42) explicitly points out `global.dashboard.seen`, it just seems like using `global.seen` elsewhere isn't the best idea.

But, I concede, in the abstract, it's perfectly fine.  So, I'm ready to move on.  🙂

webbnh (approved):
👍

docs/API/metadata.md
```diff
@@ -104,14 +104,20 @@ received based on user profile preferences and server configuration; but it can
...
+### Global namespace
 
 The server will never modify or directly interpret values in this namespace. An
 authenticated client representing the owner of a dataset can set any keys
 within this namespace to any valid JSON values (string, number, boolean, list,
 or nested objects) for retrieval later. All clients with read access to the
 dataset will see the same values.
 
+The recommended best practice is to assign a project sub-key that will be unique
+and avoid collisions between distinct clients. The Pbench Dashboard project,
```
webbnh, line 116: I don't think "assign" is the right word.  We don't have or envision any sort of registrar who would be responsible for giving out such assignments.  I suggest using "select" (or "use", or similar) in place of "assign", here and at line 140.

webbnh (approved):

webbnh (approved):
🥳

npalaska (approved):

PR #2954, opened 2022-07-27
Api dataset contents

siddardh-ra:
PBENCH-742

Refactored `/datasets/contents/<dataset>/<path>` API to GET method from POST.

- Fetch the relative path from the URI parameter `<path>` instead of passing through the payload.

- An added new fixture in unit test for GET method and modified test cases accordingly.


lib/pbench/server/api/resources/query_apis/datasets/__init__.py
```diff
@@ -91,19 +90,20 @@ def __init__(self, config: PbenchServerConfig, logger: Logger, *schemas: ApiSche
         super().__init__(config, logger, *schemas)
 
         api_name = self.__class__.__name__
+        self.method = schemas[0].method
```
dbutenhof, line 93: We don't have any `ElasticBase` APIs that have two schemas, supporting two different HTTP methods, and I'm not entirely sure we'll need to do that... so I'm just pointing out here that your code won't support that.

I don't think it's worth complicating the code in this PR to handle that elegantly, however it might be a good idea to add a check for `len(schemas) > 1` and fail the constructor as done below on line 95 if no schema is supplied at all. (In fact, if that condition ever actually occurred, you'd be throwing an exception at line 93, so reversing these might not be a bad idea.)

(While `MissingDatasetNameParameter` has apparently already been somewhat twisted by this "no schema provided", one option here would be to simply change the `if not schemas` to `if len(schemas) != 1` with a slight tweak to the text.)

dbutenhof (commented):
Very nice, Siddardh! I think the Flask route mapping needs some work; this is a problem that I've also just discovered in dealing with my config option PR, and I *think* I've got a solution.

(I may have missed something, since I'm multitasking during class while the instructor talks about POSIX realtime priority scheduling. I understand that pretty well, but my attention is still divided!)

lib/pbench/server/api/__init__.py
```diff
@@ -57,7 +57,8 @@ def register_endpoints(api, app, config):
 
     api.add_resource(
         DatasetsContents,
-        f"{base_uri}/datasets/contents/<string:dataset>",
+        f"{base_uri}/datasets/contents/<string:dataset>/",
```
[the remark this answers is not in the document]
dbutenhof: I retract this... maybe... for now. This works great on my laptop venv with direct pytest, but doesn't seem to work under tox or in Jenkins. Something about the routing is still messed up and I haven't figured it out yet.

lib/pbench/test/unit/server/query_apis/conftest.py
```diff
@@ -60,3 +60,61 @@ def query_api(
...
+@pytest.fixture
+@responses.activate
+def query_api_get(client, server_config, provide_metadata):
+    """
+    Help controller queries that want to interact with a mocked
+    Elasticsearch service.
+
+    This is a fixture which exposes a function of the same name that can be
+    used to set up and validate a mocked Elasticsearch query with a JSON
+    payload and an expected status.
```
webbnh, line 74: > which exposes a function of the same name

That's not actually what it does or how it works.  In a test, a reference to a fixture yields the (cached) value which the function of that name returned.  So, "This is a fixture which returns a function that can be...."

lib/pbench/test/unit/server/query_apis/conftest.py
```diff
@@ -60,3 +60,61 @@ def query_api(
...
+        port = server_config.get("elasticsearch", "port")
+        es_url = f"http://{host}:{port}{expected_index}{es_uri}"
+        with responses.RequestsMock() as rsp:
+            """
+            We need to set up mocks for the Server's call to Elasticsearch,
+            which will only be made if we don't expect Pbench to fail before
+            making the call. We never expect an Elasticsearch call when the
+            expected status is FORBIDDEN or UNAUTHORIZED; or when we give the
+            canonical "bad username" (badwolf) and are expecting NOT_FOUND.
+            """
```
webbnh, line 103: I'm finding this text confusing.  For one thing, I think we only set up a single mock, for a request to `es_url`.  (I suppose we set up multiple mocks across multiple calls to this function, but I don't think that perspective helps understanding here.)  It's true that we only set the mock under the circumstances which the text outlines, but whether the CUT calls ES or not has nothing to do with our expectations here.  That is, we don't set up the mock when we expect the CUT to issue a failure before invoking the mock (which is not really what the text says).

Moreover, if we _don't_ expect the Server to call ES, then we should have a mock for it in place which fails the test if it is invoked....

webbnh (commented):
This looks generally good...I just found a bunch of nits and niggly stuff, and I have a couple of suggestions.

lib/pbench/server/api/__init__.py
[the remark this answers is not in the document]
webbnh: Does you laptop have a newer/different version of Flask?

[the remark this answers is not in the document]
dbutenhof: Not flask version, or Python version. With a relatively low bandwidth during class, I didn't try to isolate every module difference. Instead I did some research and experimentation, and worked out a different approach. For one thing, my filtering of the URI pattern to generate the API key name (e.g., `datasets_contents`) falls prey to this quirk, and I've changed the `add_resource` calls to specify the desired tag name using the `endpoint` parameter. Then I made the endpoints generator smarter so it handles these alternatives... although I'm not confident of the *order* I'll see the `Rule` objects so I tried to make it resilient. Siddardh's case is "fancier" than mine, so I realize that I need another tweak, which I should probably do as a separate PR anyway...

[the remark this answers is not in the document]
webbnh: @dbutenhof, I saw that your tests are now passing!  🎉 

> I've changed the `add_resource` calls

Yeah, I saw; I posted questions about that in your PR.

> Siddardh's case is "fancier" than mine, so I realize that I need another tweak, which I should probably do as a separate PR anyway...

Maybe do it as a PR to this one (assuming that your PR merges before this one), and Siddardh can merge it in here and it can get into `main` that way?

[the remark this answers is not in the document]
dbutenhof: That's why I separated out #2956 ... ideally that'll go through quickly, and we'll both merge against it for the win.

lib/pbench/test/unit/server/query_apis/commons_get.py
```diff
@@ -0,0 +1,390 @@
+from http import HTTPStatus
+from typing import AnyStr
+
+import pytest
+import requests
+from dateutil import parser as date_parser
+from dateutil import rrule
```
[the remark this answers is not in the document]
dbutenhof: I suspect this is the same way in `commons.py`, which Siddardh just cloned with minor changes to replace a lot of `post` calls with `get`. Not that this invalidates style comments; but ideally I'd like to see these two merged or separated more cleanly in the future, in which case non-essential divergence will just complicate someone's job at that time.

lib/pbench/test/unit/server/query_apis/commons_get.py
```diff
@@ -0,0 +1,390 @@
...
+        self.cls_obj = cls_obj
+        self.pbench_endpoint = pbench_endpoint
+        self.elastic_endpoint = elastic_endpoint
+        self.payload = payload
+        self.bad_date_payload = bad_date_payload
+        self.error_payload = error_payload
+        self.empty_es_response_payload = empty_es_response_payload
+        self.index_from_metadata = index_from_metadata
+
+    def build_index(self, server_config, dates):
```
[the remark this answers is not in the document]
dbutenhof: Again, I think you'll find this is the same in `commons.py`. While improving this is great, I'm not sure I want them to diverge more than necessary at this point...

lib/pbench/test/unit/server/query_apis/commons_get.py
```diff
@@ -0,0 +1,390 @@
...
+        actually hitting now, which appears to fail as expected but actually
+        fails because we're asking for {"access": "private"} data, which is not
+        allowed for an unauthenticated client call.
+        """
+        response = client.get(
+            server_config.rest_uri + self.pbench_endpoint,
+            headers={"Authorization": malformed_token},
+            json=self.payload,
+        )
+        assert response.status_code == HTTPStatus.UNAUTHORIZED
```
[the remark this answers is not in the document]
dbutenhof: Can I? Maybe. Will I? More complicated... and I'd rather not.

This history is messy. There are several "authorization validation" tests originally written that turned out to have been "mostly meaningless" due to the way the Flask authorization package works. (E.g., bad authorization format is the same as none at all, as documented here.) I added the comments partly thinking we might want to write our own replacement for `Flask-HTTPAuth` to "do it right"... if we actually care... and as the time between then and now increases I'm less sure that we'll ever have any reason to care.

In any case, again, this is the same in `commons.py` and the new `commons_get.py` and I'd prefer not to have them diverge unnecessarily under this PR. (That is, I'd like the ability to `diff` them and see the `post` -> `get` transformation, which may or may not help with unifying them or layering a better model to avoid the duplication in the future.)

lib/pbench/server/api/__init__.py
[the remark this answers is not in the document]
webbnh: Yes, that was a good move.  (It would have been better if you'd done it even sooner.... 😉)

lib/pbench/test/unit/server/query_apis/commons_get.py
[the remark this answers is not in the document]
webbnh: OK.  On that basis, should I be insta-resolving the rest of my comments on `commons_get.py`?  (I'm kind of at a loss to tell what should be fixed in this PR and what should be deferred for the ease of merging.)

lib/pbench/server/api/__init__.py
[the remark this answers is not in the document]
dbutenhof: "Sooner" is a relative term... I worked out the solution in #2956 yesterday during class, so I couldn't possibly have separated it out from #2949 much sooner than yesterday afternoon! 😆

lib/pbench/test/unit/server/query_apis/commons_get.py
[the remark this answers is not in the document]
dbutenhof: Ha. The `POST` -> `GET` changes were required by Siddardh's change. We still have `ElasticBase` classes that use `POST` (and perhaps always will), and I directed Siddardh to start with a "brute force" clone that made just the changed he needed. In theory that makes it possible to compare them and evolve in parallel if necessary. I've always wanted to refactor `Commons` just because a lot has changed (and been learned, such as the broken validation tests) since it was written. Realistically, this might sit on the back burner until we build a real unit test infrastructure for the API classes instead of using the mocked requests/Flask... but who knows?

Mostly, I was pointing out that most if not all of your comments here could be against `commons.py` as well and if we're taking action on them we should do it in both at the same time (if they can't ultimately be unified) and that suggests at least deferring to another PR. 🍤

lib/pbench/server/api/__init__.py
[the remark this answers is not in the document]
webbnh: In this case, "sooner" would have been "before you added it to #2949".  🙂

lib/pbench/test/unit/server/query_apis/commons_get.py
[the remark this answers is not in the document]
webbnh: Alright, I resolved a bunch of them, mostly the semi-substantial ones.  The trivial ones should probably be fixed in this PR, and then, when someone diffs the two modules, they will show up as small things to fix in the other one.  I did leave two of them unresolved which call for more substantive changes -- there's a redundant assertion (which should be trivial to get rid of and easy to merge) and there's the save-and-restore thing which I think should be changed to a copy-and-destroy approach.

lib/pbench/server/api/__init__.py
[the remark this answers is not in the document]
dbutenhof: I glanced through this again and, just to be thorough, after #2949, the pattern here should resemble what I have for `ServerConfiguration`:

```
    api.add_resource(
        DatasetsContents,
        f"{base_uri}/datasets/contents/<string:dataset>",
        f"{base_uri}/datasets/contents/<string:dataset>/<path:path>",
        endpoint="datasets_contents",
    )
```

You should be able to add a third pattern for `f"{base_uri}/datasets/contents/<string:dataset>/"` and have them all merged by the endpoint name.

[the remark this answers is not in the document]
webbnh: @dbutenhof, this conversation is currently marked "resolved"...if you're making a specific request, should you "unresolve" it?  (Will @siddardh-ra see and note your comment if the conversation is resolved?)

Also, you said, "after #2949" -- did you mean #2956?  (This PR doesn't need to wait until #2949 has been merged, does it?)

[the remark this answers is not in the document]
siddardh-ra: Yeah @dbutenhof, noted. I've already updated the code once https://github.com/distributed-system-analysis/pbench/pull/2956 is merged. I will push my changes shortly with those updated changes :+1:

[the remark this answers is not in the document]
dbutenhof: Perhaps that was a bit ambiguous. #2956 changes the `Rule` parsing to enable both #2954 and #2949 to express these multiple endpoints without causing confusion. The intent of the note here was that #2954 should follow the example of (not yet merged) #2949 to leverage the change in #2956 rather than the earlier advice I'd given here. 😁

webbnh (approved):
Good by me.

dbutenhof (approved):

npalaska (approved):

lib/pbench/server/api/__init__.py
```diff
@@ -57,7 +57,8 @@ def register_endpoints(api, app, config):
...
+        f"{base_uri}/datasets/contents/<string:dataset>/<path:path>",
```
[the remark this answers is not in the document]
portante: If we are going to change from `parent` to `path`, why are we not also changing the `context` variable to `path` and the table of contents documents to `path`.  ToC entries have a keyword, `parent`, and a analyzed path, `directory`, to facilitate different ways to search.  I'd rather not introduce yet another name to add more confusion as to what goes where.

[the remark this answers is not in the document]
dbutenhof: I'm not concerned about the Elasticsearch mapping: we're just using ES as a hidden backend database, so (in theory) nobody outside the server needs to know how we organize data there. While "parent" may be, in that chained document structure, a convenient and even meaningful name for the "chain" field, I don't think it's a good representation of the "target" of `/datasets/contents`; our return payload is a JSON document that consolidates multiple ES documents. The caller is specifying a relative path within the tarball tree, and getting data about it.

Even for "contents" I felt that the name "parent" was ambiguous and a bit loose, but it's totally off the mark for "inventory", and I really wanted the two to be "twinsies". Whereas "path" seems like a good fit for both. We could say "target" but that seems less clear.

And, sure, changing the `CONTEXT` object key to be consistent is a good idea.

lib/pbench/server/api/resources/query_apis/datasets/datasets_contents.py
```diff
@@ -30,13 +30,11 @@ def __init__(self, config: PbenchServerConfig, logger: Logger):
...
-                API_METHOD.POST,
+                API_METHOD.GET,
                 API_OPERATION.READ,
                 uri_schema=Schema(
-                    Parameter("dataset", ParamType.DATASET, required=True)
-                ),
-                body_schema=Schema(
-                    Parameter("parent", ParamType.STRING, required=True),
+                    Parameter("dataset", ParamType.DATASET, required=True),
+                    Parameter("path", ParamType.STRING, required=False),
```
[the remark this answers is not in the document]
webbnh: See [discussion below](https://github.com/distributed-system-analysis/pbench/pull/2954#discussion_r943685309).

lib/pbench/server/api/resources/query_apis/datasets/datasets_contents.py
```diff
@@ -50,16 +48,15 @@ def assemble(self, params: ApiParams, context: CONTEXT) -> JSON:
...
-
-        EXAMPLE:
-        {
-            "parent": '/1-default'
-        }
         """
         # Copy parent directory metadata to CONTEXT for postprocessor
-        parent = context["parent"] = params.body.get("parent")
+        try:
+            parent = context["parent"] = "/" + params.uri["path"]
```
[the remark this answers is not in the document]
webbnh: Just FYI, @dbutenhof, posting a comment like this does not revoke your previous approval (you would need to "request changes" to do that, I think), although the open conversation _will_ have a similar effect.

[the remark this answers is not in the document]
dbutenhof: That was deliberate, although I did briefly consider revoking the approval I'm really not unhappy as it is. 🍤

lib/pbench/server/api/__init__.py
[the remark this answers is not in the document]
portante: Being consistent is good, except where we allow an internal representation to impact the external API without considering the consumer of the API.

In this case, I don't believe that is happening, though it seems we might be leaning towards having the API impact the internal representation.

When I read, `path`, I find that I ask myself: is `path` the full path including the `name` or is it the `path` to the parent?  `path` seems to be a name that indicates the form the contents take, but does not convey the semantic meaning.  I deliberately chose `parent` in the past because it conveys the meaning of the field, though I did not try to encode in the name the "type" of the value for that field (a directory path).

For the API, could we consider using `object` or `target` instead of `path`?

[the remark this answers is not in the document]
dbutenhof: While I see what you mean by "parent", that feels very unnatural to me; yes, `/datasets/contents` returns the contents or "children" of the named `path`/`target`, but if I saw "parent" in the documentation that's not what it implies to me. I guess because I'm pointing at the directory I want, not at the parent of the directory I want.

But `target` seems OK if that'll make everyone happy. Or, although I liked the idea of consistency between the two sibling APIs, we could be more explicit and call this one `directory` and the other `file` ...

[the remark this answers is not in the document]
webbnh: What's a "directory" (and "file")?  My storage system's API and GUI seems to have "folders" (and "objects").  😝

Anyway, 'target' is OK with me.

lib/pbench/server/api/resources/query_apis/datasets/datasets_contents.py
```diff
@@ -48,21 +48,22 @@ def assemble(self, params: ApiParams, context: CONTEXT) -> JSON:
...
-        try:
-            parent = context["parent"] = "/" + params.uri["path"]
-        except KeyError:
-            parent = context["parent"] = "/"
+        # Copy target directory metadata to CONTEXT for postprocessor
+        get_target = params.uri.get("target")
+
+        k = "/" if get_target is None else f"/{get_target}"
+
+        target = context["target"] = k
```
webbnh, line 59: If I might suggest:
```
        # Copy target directory metadata to CONTEXT for postprocessor
        target = "/" + params.uri.get("target", "")
        context["target"] = target
```

webbnh (dismissed):
Looks good, although I have a suggestion for a slightly more graceful bit of code.

dbutenhof (dismissed):
Unfortunately, Webb made a minor suggestion and while he approved regardless, I'm philosophically reluctant to resolve the conversation on his behalf. 😦

lib/pbench/test/unit/server/query_apis/commons_get.py
```diff
@@ -0,0 +1,387 @@
+from http import HTTPStatus
```
portante, line 1: This seems like a bad idea to clone 387 lines of one file just to add support for GET.

lib/pbench/test/unit/server/query_apis/conftest.py
```diff
@@ -60,3 +60,53 @@ def query_api(
...
+            This set up mocks the Server's call to Elasticsearch,
+            unless when the expected status is either FORBIDDEN or UNAUTHORIZED
+            """
+            if expected_status not in [
+                HTTPStatus.FORBIDDEN,
+                HTTPStatus.UNAUTHORIZED,
+            ]:
+                rsp.add(responses.GET, es_url, **kwargs)
+            response = client.get(
```
portante, line 105: Why are we hiding the actual test in a fixture?

portante (changes requested):
I am not in favor of cloning a whole test module.

lib/pbench/test/unit/server/query_apis/conftest.py
dbutenhof: Technically this isn't a fixture; it's a customized helper method created using the magic of fixtures. The pattern simplifies a lot of test cases by pushing common fixture dependencies (e.g., `client, server_config, provide_metadata`) out of the individual test case declarations and returns the constructed `query_api` to be called where needed.

Of course it could have been done other ways, but this pattern goes way back to the start of our Flask "unit" tests and it wasn't invented here.

lib/pbench/test/unit/server/query_apis/commons_get.py
dbutenhof: The `Commons` base class is already pretty complicated and messy. Some of it is already driven by introspecting a constructed resource object; and extending that would make it "straightforward" if not necessarily trivial to use the resource `SchemaSet` to figure out which HTTP methods are supported by the resource, as well as which parameters go with each.

The question was whether it was worth trying to accomplish that *before* getting the `/dataset/contents` API fixed. At the time, we decided not to complicate this work or create a new dependency.

On one hand, I'd like to see `Commons` cleaned up, generalized, extended, and probably broken down to be used at higher and lower levels in the hierarchy.

On another hand, I'd rather devote that effort to stripping out the flask and request mocks and making the API suite true unit tests.

And all of this seemed extra baggage for this specific PR.

portante: There has to be some kind of middle ground here.

If one were to `diff` the new `commons_get.py` and `commons.py`, what changed?  Can't we just pass a new parameter to use `GET` instead?

webbnh:
> Unfortunately, Webb made a minor suggestion and while he approved regardless, I'm philosophically reluctant to resolve the conversation on his behalf. 😦

Whether it was unfortunate or not is debatable:  since Peter already had an outstanding "change requested", I figured I had the "headroom" to make a suggestion, and, given that he's since posted another, I don't think my comment is causing any misfortune. 😀

But, thank you for your reluctance.  😁

lib/pbench/test/unit/server/query_apis/test_datasets_contents.py
```diff
@@ -14,25 +15,25 @@ class TestDatasetsContents(Commons):
...
+            api_method=API_METHOD.GET,
         )
 
     def test_with_no_index_document(self, client, server_config):
         """
         Check the DatasetsContents API when no index name is provided
         """
-        # remove the last component of the pbench_endpoint
-        incorrect_endpoint = "/".join(self.pbench_endpoint.split("/")[:-1])
+        # remove the last two component of the pbench_endpoint
```
portante, line 35: `s/component/components`

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -129,6 +131,24 @@ def get_expected_status(self, payload: JSON, header: HeaderTypes) -> int:
             expected_status = HTTPStatus.OK
         return expected_status
 
+    def make_request_call(
+        self, client, url, header: JSON, json: JSON = None, data=None
+    ):
+        if self.api_method == API_METHOD.GET:
```
portante, line 137: You could consider:

```
if self.api_method not in (API_METHOD.GET, API_METHOD.POST):
    raise Exception(f"Unexpected API method, {self.api_method}")
_func = client.get if self.api_method == API_METHOD.GET else client.post
return _func(url, headers=header, json=json, data=data)
```

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -254,10 +279,11 @@ def test_missing_keys(self, client, server_config, pbench_token):
         """
 
         def missing_key_helper(keys):
-            response = client.post(
+            response = self.make_request_call(
```
portante, line 282: Is this method defined inline only used for `POST` operations?  It appears so.  We should only run this test for `POST` operations.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -399,21 +440,32 @@ def test_http_exception(
...
+                index,
+                exceptions["status"],
+                body=exceptions["exception"],
+                headers={"Authorization": "Bearer " + pbench_token},
+            )
+        elif self.api_method == API_METHOD.POST:
+            query_api(
+                self.pbench_endpoint,
+                self.elastic_endpoint,
+                self.payload,
```
portante, line 456: Can we make this an optional parameter, so we don't have to have this duplicate code?

portante (changes requested):

lib/pbench/test/unit/server/query_apis/commons.py
dbutenhof: Yeah; at least, it's more complicated for `GET` since "missing parameters" are URI or query parameters rather than JSON body. That's not currently handled. For example, if there's a required dataset ID in the URL, we could drive a test case without it (e.g., `GET /datasets/delete` where the code requires `/datasets/delete/<resource_id>`). Or maybe that test should just be special cased (which I think it is), and we just skip this test case for `GET` APIs.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -290,28 +321,37 @@ def missing_key_helper(keys):
                 else:
                     keys[key] = "foobar"
 
-            missing_key_helper(keys)
+            if self.api_method == API_METHOD.POST:
+                missing_key_helper(keys)
```
dbutenhof, line 326: This should be done at the top level of the method, not in an inner loop, because the answer won't change!

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -266,9 +292,14 @@ def missing_key_helper(keys):
...
-            API_METHOD.POST
-        ].body_schema.parameters.items()
+        if self.api_method == API_METHOD.GET:
+            parameter_items = self.cls_obj.schemas[
+                self.api_method
+            ].uri_schema.parameters.items()
+        elif self.api_method == API_METHOD.POST:
+            parameter_items = self.cls_obj.schemas[
+                self.api_method
+            ].body_schema.parameters.items()
```
webbnh, line 302: I think Dave mentioned this elsewhere, but here's the code for it:  this is assuming that `POST` requests have parameters only in the body and none in the URI nor query parameters.

Also, should we have more branches to this `if` statement, e.g., for `PUT` and/or `DELETE`?  (If not, should we have an "else fail"??)

lib/pbench/test/unit/server/query_apis/commons.py
webbnh: It's not the best time of day for me to be making this assertion, but I'm _pretty_ sure that the answer **will** change, because...isn't the value of `keys` different on every call?  (Also, this call is _after_ the innermost loop....)

lib/pbench/test/unit/server/query_apis/commons.py
webbnh: Peter, is the addition of `query_api_get()` the duplication that you're trying to avoid?

Yeah, it looks like that can be done with an optional parameter, but I think it's going to make a complicated piece of logic even more complex (I'm too foggy right now to figure out how much).

lib/pbench/test/unit/server/query_apis/commons.py
dbutenhof: No, the point was that `self.api_method == API_METHOD.POST` won't change!

And, yeah, I wasn't entirely clear. I mean, if we're not going to actually deal with checking for missing keys in the URI or QUERY schemas, then we should skip this test when `self.api_method != API_METHOD.POST` up at the beginning rather than going through this whole thing doing nothing...

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -128,6 +130,24 @@ def get_expected_status(self, payload: JSON, header: HeaderTypes) -> int:
...
         return expected_status
 
+    def make_request_call(
+        self, client, url, header: JSON, json: JSON = None, data=None
+    ):
+        if self.api_method == API_METHOD.GET:
+            return client.get(
+                url,
+                headers=header,
+                json=json,
```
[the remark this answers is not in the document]
dbutenhof: `json` is a `Dict` representing a JSON request payload; it's equivalent to using `data=json.dumps(json)`. So we should never be specifying *both*.

It's usually probably better to keep both as `kwargs` so the caller can specify one or the other... except that I don't think we have *any* test cases where we actually want to specify raw data, so we probably shouldn't bother with `data` at all.

lib/pbench/test/unit/server/query_apis/commons.py
webbnh: Ah, right:  we could move the `if` test to just before the `for` statement at line 314 so that the outer loop is skipped if the method isn't `POST` (and then we could remove the `if` test from inside the loop).

@siddardh-ra, if you do that, then you should pull the `if` statement at line 329 into the same block as the `for` statement at line 314 (and remove the method test from it).

lib/pbench/test/unit/server/query_apis/commons.py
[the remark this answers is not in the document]
webbnh: And we should probably have a check in there somewhere that no call specifies values for _both_ of them.

lib/pbench/test/unit/server/query_apis/commons.py
dbutenhof: Note that `self.payload` on your new `GET` API classes will be `None`, so a single pattern ought to work with my suggested consolidated `query_api`. (Though "some assembly may be required" 😆)

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -255,20 +251,21 @@ def test_malformed_payload(self, client, server_config, pbench_token):
...
+                    "Authorization": "Bearer " + pbench_token,
+                    "Content-Type": "application/json",
+                },
+                data='{"x":0]',
+            )
+            assert response.status_code == HTTPStatus.BAD_REQUEST
+            assert (
+                response.json.get("message")
+                == "Invalid request payload: '400 Bad Request: The browser (or proxy) sent a request that this server could not understand.'"
+            )
```
webbnh, line 268: I think Dave also mentioned this, but this test now does _nothing_ if the method is not `POST`.  I concur with Dave -- this should be recast this as
```
        if self.api_method != API_METHOD.POST:
            pytest.skip("skipping " + self.test_malformed_payload.__name__)

        response = self.make_request_call(
            client,
            ...
```
Likewise for the other functions like this.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -382,17 +373,27 @@ def test_empty_query(
...
+            response = query_api_get(
+                self.pbench_endpoint,
+                self.elastic_endpoint,
+                index,
+                expected_status,
+                headers=build_auth_header["header"],
+                status=HTTPStatus.OK,
+                json=self.empty_es_response_payload,
+            )
+        elif self.api_method == API_METHOD.POST:
```
webbnh, line 386: I'm not really happy with the use of `elif`, here:  if we're reasonably sure that `self.api_method` is either `GET` or `POST`, then this `elif` should be a simple `else`; however, if we're making allowances for `self.api_method` to be something else (now or in the future), then there should be an `else` clause after line 396 which does something intelligent (like failing or skipping the test).

(You could also consider expanding the conditional at line 368.)

webbnh (changes requested):
Just some niggly stuff, but I think they should be fixed.

lib/pbench/server/api/resources/query_apis/datasets/__init__.py
```diff
@@ -112,8 +114,9 @@ def preprocess(self, params: ApiParams) -> CONTEXT:
         using the Dataset ownership/access, so validation and authorization has
         already taken place.
         """
+
         _, dataset = self.schemas.get_param_by_type(
-            API_METHOD.POST, ParamType.DATASET, params
+            self.schema.method, ParamType.DATASET, params
```
dbutenhof, line 119: Or `self.schema.get_param_by_type(ParamType.DATASET, params)` since we *know* which method/schema we want to check.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -55,6 +55,13 @@ def _setup(
         self.error_payload = error_payload
         self.empty_es_response_payload = empty_es_response_payload
         self.index_from_metadata = index_from_metadata
+        if API_METHOD.GET in self.cls_obj.schemas:
+            self.api_method = API_METHOD.GET
+        elif API_METHOD.POST in self.cls_obj.schemas:
+            self.api_method = API_METHOD.POST
+        else:
+            self.api_method = None
+            raise Exception("Unexpected API method")
```
dbutenhof, line 64: Hmm. My first thought was that you don't really need to set `self.api_method` here since you're aborting the constructor with an exception. Except ... in this weird pytest hackery, it's *not* really a constructor. In any case, I don't think that setting `self.api_method` to `None` is really going to help much if the test continues, so that's probably not the best solution.

Another option is to `assert` that the class has either `API_METHOD.GET` or `API_METHOD.POST` so that the test fails immediately. _Or_ else we could `pytest.skip` because the common test class isn't applicable.

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -129,6 +136,14 @@ def get_expected_status(self, payload: JSON, header: HeaderTypes) -> int:
             expected_status = HTTPStatus.OK
         return expected_status
 
+    def make_request_call(
+        self, client, url, header: JSON, json: JSON = None, data=None
+    ):
+        assert self.api_method != API_METHOD.GET or json is data is None
+
+        func = client.get if self.api_method == API_METHOD.GET else client.post
```
dbutenhof, line 144: Since you have two tests for `API_METHOD.GET`, this might be better as an `if/else` instead of two expressions:

(NOTE, I put two separate assert statements because that makes slightly better messaging if one condition fails; even though that's probably not a big deal in this context.)

```python
    assert self.api_method in (API_METHOD.GET, API_METHOD.POST)
    if self.api_method == API_METHOD.GET:
        assert json is None
        assert data is None
        func = client.get
    else:
        func = client.post
```

lib/pbench/server/api/resources/query_apis/datasets/__init__.py
```diff
@@ -87,18 +86,21 @@ def __init__(self, config: PbenchServerConfig, logger: Logger, *schemas: ApiSche
...
-        if not schemas:
-            raise MissingDatasetNameParameter(api_name, "no schema provided")
-        dset = self.schemas.get_param_by_type(
-            API_METHOD.POST,
+        if len(schemas) != 1:
+            raise MissingDatasetNameParameter(
+                api_name, f"exactly one schema required: found {len(schemas)}"
+            )
+
+        self.schema = schemas[0]
```
dbutenhof, line 94: I don't think we should store this as an instance property; it's not (and probably shouldn't be) used outside the constructor for basic consistency checking. I think just a local `schema` would be better.

dbutenhof (commented):
Much cleaner without the "near duplicate" helpers! A few more minor comments...

lib/pbench/test/unit/server/query_apis/test_datasets_contents.py
```diff
@@ -142,14 +145,16 @@ def test_query(
             auth_json, build_auth_header["header_param"]
         )
 
-        response = query_api_get(
+        response = query_api(
             self.pbench_endpoint,
             self.elastic_endpoint,
+            self.payload,
```
webbnh, line 151: `self.payload` is always `None`, here, right?  If that's the case, I recommend changing this line to
```
           payload=None,
```
Ditto line 263, et al.

webbnh (approved):
This looks good, now.  I have just a few bits of polishing that you should consider while you're making other changes.

lib/pbench/server/api/resources/query_apis/datasets/__init__.py
dbutenhof: Indeed; I was skimming too quickly this morning and failed to notice that there was a gap between the diffs -- it's outside the constructor.

Since we do know there's only a single `Schema` it could still use `self.schemas[0]`; but let's just keep the `self.schema`. That seems a bit cleaner outside the constructor.

siddardh-ra: Yeah, Dave.So only I thought of using this `self.schema.` Then I'll leave this code as it is :smile:

lib/pbench/test/unit/server/query_apis/commons.py
```diff
@@ -251,21 +250,22 @@ def test_malformed_payload(self, client, server_config, pbench_token):
...
-                    "Content-Type": "application/json",
-                },
-                data='{"x":0]',
-            )
-            assert response.status_code == HTTPStatus.BAD_REQUEST
-            assert (
-                response.json.get("message")
-                == "Invalid request payload: '400 Bad Request: The browser (or proxy) sent a request that this server could not understand.'"
-            )
+        if self.api_method == API_METHOD.GET:
```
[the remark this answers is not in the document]
siddardh-ra: Yeah agreed :100:

webbnh (approved):
Looks good to me.

webbnh (approved):
Looks good to me.

dbutenhof (approved):
I have one minor comment still open, but it's not a big deal. If Peter closes his conversations without Siddardh needing to make additional changes, I'll just close it.

portante (changes requested):

lib/pbench/test/unit/server/query_apis/commons.py
siddardh-ra: > Is there a reason not to consider this change?

Actually, I tried using the above code block as well, but while asserting and validating `data` and `json` only for the `GET` operation. This seemed to be a more simple way

dbutenhof: I like the `assert` rather than explicitly raising `Exception` since this is test code. Though it's definitely "debatable" since this is test infrastructure diagnosing a problem in the unit test code rather than a problem in the product code we're testing...

portante (approved):

